PR #197093, opened 2026-05-12
[RISCV][P-ext] Add initial 64-bit support for RV32.

topperc:
Most operations are set to expand. A few operations that were easy to support using isel patterns have been added. concat_vectors and extract_subvector are supported in order to allow type legalization to split 64-bit vectors into 32-bit vectors around the supported operations.

Loads and stores are custom split into two i32 scalars or two v4i8/v2i16 vectors.

I've added new opcodes to build and split vectors into 2 GPRs at function arguments and returns. These are similar to BuildPairF64 and SplitF64 nodes we use for RV32D soft float. Long term we might want to use concat_vectors/build_vector and
extract_subvector/extract_vectorelt.

### DIFF
--- a/llvm/lib/Target/RISCV/RISCVCallingConv.cpp
+++ b/llvm/lib/Target/RISCV/RISCVCallingConv.cpp
@@ -405,6 +405,10 @@ static bool CC_RISCV_Impl(unsigned ValNo, MVT ValVT, MVT LocVT,
       ValNo > 1)
     return true;
 
+  if (Subtarget.isPExtPackedType(LocVT) && LocVT.getSizeInBits() > XLen &&
+      IsRet && ValNo > 0)
+    return true;
+
   // AllowFPRForF16_F32 if targeting an FLEN>=32 ABI and the argument isn't
   // variadic.
   bool AllowFPRForF16_F32 = false;
@@ -525,8 +529,10 @@ static bool CC_RISCV_Impl(unsigned ValNo, MVT ValVT, MVT LocVT,
 
   // Handle passing f64 on RV32D with a soft float ABI or when floating point
   // registers are exhausted. Or 64-bit P extension vectors on RV32.
-  if (XLen == 32 && (LocVT == MVT::f64 || (Subtarget.isPExtPackedType(LocVT) &&
-                                           LocVT.getSizeInBits() == 64))) {
+  if (XLen == 32 &&
+      (LocVT == MVT::f64 ||
+       (Subtarget.isPExtPackedType(LocVT) && LocVT.getSizeInBits() == 64 &&
+        !ArgFlags.isSplit() && PendingLocs.empty()))) {
     assert(PendingLocs.empty() &&
            "Can't lower f64 or P extension vector if it is split");
     // Depending on available argument GPRS, f64 may be passed in a pair of
@@ -556,7 +562,8 @@ static bool CC_RISCV_Impl(unsigned ValNo, MVT ValVT, MVT LocVT,
 
   // If the split argument only had two elements, it should be passed directly
   // in registers or on the stack.
-  if ((LocVT.isScalarInteger() || Subtarget.isPExtPackedType(LocVT)) &&
+  if ((LocVT.isScalarInteger() ||
+       (Subtarget.isPExtPackedType(LocVT) && LocVT.getSizeInBits() == XLen)) &&
       ArgFlags.isSplitEnd() && PendingLocs.size() <= 1) {
     assert(PendingLocs.size() == 1 && "Unexpected PendingLocs.size()");
     // Apply the normal calling convention rules to the first half of the
@@ -632,7 +639,7 @@ static bool CC_RISCV_Impl(unsigned ValNo, MVT ValVT, MVT LocVT,
   // end of a split argument that must be passed indirectly.
   if (!PendingLocs.empty()) {
     assert(ArgFlags.isSplitEnd() && "Expected ArgFlags.isSplitEnd()");
-    assert(PendingLocs.size() > 2 && "Unexpected PendingLocs.size()");
+    assert(PendingLocs.size() > 1 && "Unexpected PendingLocs.size()");
 
     for (auto &It : PendingLocs) {
       if (Reg)
--- a/llvm/lib/Target/RISCV/RISCVCallingConv.cpp
+++ b/llvm/lib/Target/RISCV/RISCVCallingConv.cpp
@@ -405,8 +405,8 @@ static bool CC_RISCV_Impl(unsigned ValNo, MVT ValVT, MVT LocVT,
       ValNo > 1)
     return true;
 
-  if (Subtarget.isPExtPackedType(LocVT) && LocVT.getSizeInBits() > XLen &&
-      IsRet && ValNo > 0)
+  // Double wide packed types require 2 GPRs so we can only return 1 of them.
+  if (Subtarget.isPExtPackedDoubleType(LocVT) && IsRet && ValNo > 0)
     return true;
 
   // AllowFPRForF16_F32 if targeting an FLEN>=32 ABI and the argument isn't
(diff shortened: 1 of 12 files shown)


llvm/lib/Target/RISCV/RISCVCallingConv.cpp
```diff
@@ -524,9 +524,11 @@ static bool CC_RISCV_Impl(unsigned ValNo, MVT ValVT, MVT LocVT,
          "PendingLocs and PendingArgFlags out of sync");
 
   // Handle passing f64 on RV32D with a soft float ABI or when floating point
-  // registers are exhausted.
-  if (XLen == 32 && LocVT == MVT::f64) {
-    assert(PendingLocs.empty() && "Can't lower f64 if it is split");
+  // registers are exhausted. Or 64-bit P extension vectors on RV32.
+  if (XLen == 32 && (LocVT == MVT::f64 || (Subtarget.isPExtPackedType(LocVT) &&
+                                           LocVT.getSizeInBits() == 64))) {
```
lenary: Can we put the check over `LocVT` into a named helper?

Something like `static bool needs64BitSplit(MVT LocVT, STI) {...}`?

llvm/lib/Target/RISCV/RISCVInstrInfoP.td
```diff
@@ -1840,6 +1840,11 @@ let Predicates = [HasStdExtP, IsRV32] in {
 // Codegen patterns
 //===----------------------------------------------------------------------===//
 
+// Pattern class for GPRPair operations
+class PatGprPairGprPair<SDPatternOperator OpNode, RVInst Inst, ValueType vt>
+    : Pat<(vt (OpNode (vt GPRPair:$rs1), (vt GPRPair:$rs2))),
+          (Inst GPR:$rs1, GPR:$rs2)>;
```
lenary: I don't quite understand why the `Inst` side of this takes two GPRs - shouldn't it be taking two GPRPairs?

llvm/test/CodeGen/RISCV/calling-conv-p-ext-vector.ll
```diff
@@ -103,22 +86,8 @@ define <16 x i8> @test_cc_v16i8(<16 x i8> %a, <16 x i8> %b) {
 define <8 x i16> @test_cc_v8i16(<8 x i16> %a, <8 x i16> %b) {
 ; RV32-LABEL: test_cc_v8i16:
 ; RV32:       # %bb.0:
-; RV32-NEXT:    lw a3, 0(a2)
```
lenary, line 106: The calling convention here is completely different - were we really getting it so wrong before?

llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -602,8 +586,10 @@ define <8 x i8> @test_paaddu_b(<8 x i8> %a, <8 x i8> %b) {
 define <2 x i32> @test_pabs_w(<2 x i32> %a) {
 ; RV32-LABEL: test_pabs_w:
 ; RV32:       # %bb.0:
-; RV32-NEXT:    abs a0, a0
-; RV32-NEXT:    abs a1, a1
+; RV32-NEXT:    li a2, 0
```
lenary, line 1025: This needs revisiting in a follow up. Same for `test_pabs_h` below. I don't think they're wrong, just the original implementations were probably better.

llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -700,8 +688,9 @@ define <2 x i32> @test_pdifu_w(<2 x i32> %a, <2 x i32> %b) {
 define <4 x i16> @test_pdif_h(<4 x i16> %a, <4 x i16> %b) {
 ; RV32-LABEL: test_pdif_h:
 ; RV32:       # %bb.0:
-; RV32-NEXT:    pabd.h a0, a0, a2
-; RV32-NEXT:    pabd.h a1, a1, a3
+; RV32-NEXT:    pmin.dh a4, a0, a2
```
lenary, line 1127: Same applies here, worth a follow up as the original sequence was probably better but this one works too.

llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -980,8 +843,10 @@ define <8 x i8> @test_pasubu_b(<8 x i8> %a, <8 x i8> %b) {
 define <4 x i16> @test_pli_h() {
 ; RV32-LABEL: test_pli_h:
 ; RV32:       # %bb.0:
-; RV32-NEXT:    pli.h a0, 100
-; RV32-NEXT:    pli.h a1, 100
+; RV32-NEXT:    lui a1, %hi(.LCPI50_0)
+; RV32-NEXT:    lw a0, %lo(.LCPI50_0)(a1)
+; RV32-NEXT:    addi a1, a1, %lo(.LCPI50_0)
+; RV32-NEXT:    lw a1, 4(a1)
```
lenary, line 1285: The pli tests are also worth a revisit.

llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -4287,14 +5607,49 @@ define <2 x i32> @test_umax_w(<2 x i32> %a, <2 x i32> %b) {
 define <4 x i16> @test_select_v4i16(i1 %cond, <4 x i16> %a, <4 x i16> %b) {
 ; RV32-LABEL: test_select_v4i16:
 ; RV32:       # %bb.0:
-; RV32-NEXT:    andi a5, a0, 1
-; RV32-NEXT:    mv a0, a1
-; RV32-NEXT:    bnez a5, .LBB205_2
+; RV32-NEXT:    addi sp, sp, -32
```
lenary, line 6046: Selects also need follow-up work.

lenary (commented):
Broadly looking right.

Lots of opportunities to do better, most noted below but I also think our bitcasts are going via the stack? Maybe some of the other vector construction/destruction operations are too. All these opportunities can be addressed in a follow-up -- this is already a big patch.

llvm/test/CodeGen/RISCV/calling-conv-p-ext-vector.ll
topperc: I think we were hitting the indirect memory path due to 4 splits. Maybe I should only handle 64-bit vectors that haven't already been split?

llvm/lib/Target/RISCV/RISCVInstrInfoP.td
topperc: copy/paste mistake and I guess tablegen didn't care. I'll fix it.

llvm/test/CodeGen/RISCV/calling-conv-p-ext-vector.ll
lenary: Looking closer, yeah. `8xi16` is more than 2xlen so needs to go on the stack. The convention in the earlier function was correct.

topperc: Without P we probably use a different calling convention. Every element will be converted to i32 before going to the calling convention code.

llvm/test/CodeGen/RISCV/calling-conv-p-ext-vector.ll
```diff
@@ -77,18 +74,18 @@ define <16 x i8> @test_cc_v16i8(<16 x i8> %a, <16 x i8> %b) {
...
-; RV32-NEXT:    padd.b a1, a1, a2
-; RV32-NEXT:    sw a3, 0(a0)
+; RV32-NEXT:    lw a6, 12(a1)
+; RV32-NEXT:    lw a7, 8(a1)
+; RV32-NEXT:    lw t0, 4(a1)
+; RV32-NEXT:    lw a1, 0(a1)
+; RV32-NEXT:    padd.b a2, a6, a2
+; RV32-NEXT:    padd.b a5, a7, a5
+; RV32-NEXT:    padd.b a4, t0, a4
+; RV32-NEXT:    padd.b a1, a1, a3
```
lenary, line 84: For some reason we're not using `padd.db` here, worth adding to your list of things to follow-up on. :)

lenary (approved):
LGTM.

llvm/test/CodeGen/RISCV/calling-conv-p-ext-vector.ll
topperc: This one is because DAGCombine believes that (extract_subvectors (add (concat_vectors), (concat_vectors)) should be split.

And for AVX that's probably true, but we don't have a way to disable it for RISC-V

PR #197094, opened 2026-05-12
[CIR] Lower aligned operator new in CIRGen (and matching cleanup-during-new)

adams381:
Two related `errorNYI`s in CIRGen for C++17 aligned `operator new` / `delete`:

- `emitCXXNewExpr` `errorNYI`'d on `"emitCXXNewExpr: pass alignment"` when the allocator is the `align_val_t`-overloaded form (called whenever you `new` an over-aligned type).
- The cleanup-during-new path (the `CallDeleteDuringNew` cleanup that fires if a ctor throws after `operator new` returns) `errorNYI`'d on the matching `"CallDeleteDuringNew: aligned allocation"`.

The non-cleanup delete path was already wired up; this just plumbs the matching new side.

Both fixes are direct ports of `clang/lib/CodeGen/CGExprCXX.cpp`.  In `emitCXXNewExpr` we look up the alignment parameter type from the allocator prototype (or fall back to `size_t` for the variadic corner case classic also handles), build a constant of that type, and append it after the size argument.  The cleanup uses `getSizeType()` for the alignment param type per the existing `// FIXME` comment that's also in classic.

Hit this trying to build `706.stockfish_r` from SPEC CPU 2026 with `-fclangir` — stockfish allocates over-aligned NNUE state via plain `new`.  The minimum repro:

```c++
struct alignas(64) Big { char data[64]; };
void f() { Big *p = new Big; delete p; }
```

The new test cross-checks against OGCG so any drift between the two backends shows up — both emit the same `_ZnwmSt11align_val_t(64, 64)` and `_ZdlPvmSt11align_val_t(p, 64, 64)` calls.

A few honest caveats worth flagging:

- The `CallDeleteDuringNew` fix isn't directly exercised by the test.  Triggering it needs `-fcxx-exceptions` plus a throwing ctor, and CIR's EH path has its own NYIs that would muddy the test.  The cleanup is a 1:1 mirror of the regular-delete-side fix that already landed and uses the same `getSizeType()` fallback as classic.
- The variadic `operator new(size_t, ...)` corner case is also untested.  The `if (allocatorType->getNumParams() > paramsToSkip)` branch covers the common case; the else branch falls back to `sizeType` (mirroring classic, which guards the equivalent with `assert(allocator->isVariadic())` and likewise has no test).
- The cleanup path uses `getSizeType()` for the alignment param type rather than the actual `align_val_t` enum, mirroring the existing `// FIXME` in classic.  ABI-equivalent — `align_val_t`'s underlying type is `size_t` — but the IR type label differs.

Both CIR test suites pass cleanly (`check-clang-cir-codegen` 325/326 + 1 unsupported, `check-clang-cir` 718/729 + 1 XFAIL + 10 unsupported).

### DIFF
--- a/clang/include/clang/CIR/Dialect/Builder/CIRBaseBuilder.h
+++ b/clang/include/clang/CIR/Dialect/Builder/CIRBaseBuilder.h
@@ -800,6 +800,14 @@ public:
     return alignment ? getI64IntegerAttr(alignment) : mlir::IntegerAttr();
   }
 
+  // Materialize an alignment value as a CIR integer constant of the given
+  // integer type.  Used to pass std::align_val_t (or its size_t fallback) to
+  // aligned-allocation operator new / operator delete calls.
+  cir::ConstantOp getAlignment(mlir::Location loc, mlir::Type t,
+                               clang::CharUnits alignment) {
+    return getConstantInt(loc, t, alignment.getQuantity());
+  }
+
   mlir::IntegerAttr getSizeFromCharUnits(clang::CharUnits size) {
     return getI64IntegerAttr(size.getQuantity());
   }
--- a/clang/include/clang/CIR/Dialect/Builder/CIRBaseBuilder.h
+++ b/clang/include/clang/CIR/Dialect/Builder/CIRBaseBuilder.h
@@ -801,8 +801,7 @@ public:
   }
 
   // Materialize an alignment value as a CIR integer constant of the given
-  // integer type.  Used to pass std::align_val_t (or its size_t fallback) to
-  // aligned-allocation operator new / operator delete calls.
+  // integer type.
   cir::ConstantOp getAlignment(mlir::Location loc, mlir::Type t,
                                clang::CharUnits alignment) {
     return getConstantInt(loc, t, alignment.getQuantity());
--- a/clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
@@ -897,8 +897,12 @@ public:
     // is an enum whose underlying type is std::size_t.
     // FIXME: Use the right type as the parameter type. Note that in a call
     // to operator delete(size_t, ...), we may not have it available.
-    if (isAlignedAllocation(params.Alignment))
-      cgf.cgm.errorNYI("CallDeleteDuringNew: aligned allocation");
+    if (isAlignedAllocation(params.Alignment)) {
+      QualType sizeType = cgf.getContext().getSizeType();
+      cir::ConstantOp align = cgf.getBuilder().getAlignment(
+          *cgf.currSrcLoc, cgf.convertType(sizeType), allocAlign);
+      deleteArgs.add(RValue::get(align), sizeType);
+    }
 
     // Pass the rest of the arguments, which must match exactly.
     for (unsigned i = 0; i != numPlacementArgs; ++i) {
@@ -1613,7 +1617,19 @@ mlir::Value CIRGenFunction::emitCXXNewExpr(const CXXNewExpr *e) {
 
     // The allocation alignment may be passed as the second argument.
     if (e->passAlignment()) {
-      cgm.errorNYI(e->getSourceRange(), "emitCXXNewExpr: pass alignment");
+      // The alignment is always the second positional argument to a
+      // C++17 aligned allocation function -- right after the size.
+      constexpr unsigned indexOfAlignArg = 1;
+      // The corresponding parameter type, if the allocator declares one;
+      // otherwise fall back to size_t (the underlying type of
+      // std::align_val_t).
+      QualType alignValType = sizeType;
+      if (allocatorType->getNumParams() > indexOfAlignArg)
+        alignValType = allocatorType->getParamType(indexOfAlignArg);
+      cir::ConstantOp align = builder.getAlignment(
+          *currSrcLoc, convertType(alignValType), allocAlign);
+      allocatorArgs.add(RValue::get(align), alignValType);
+      ++paramsToSkip;
     }
 
     // FIXME: Why do we not pass a CalleeDecl here?
@@ -1821,8 +1837,8 @@ void CIRGenFunction::emitDeleteCall(const FunctionDecl *deleteFD,
     CharUnits deleteTypeAlign =
         getContext().toCharUnitsFromBits(getContext().getTypeAlignIfKnown(
             deleteTy, /*NeedsPreferredAlignment=*/true));
-    cir::ConstantOp align = builder.getConstInt(
-        *currSrcLoc, convertType(alignValType), deleteTypeAlign.getQuantity());
+    cir::ConstantOp align = builder.getAlignment(
+        *currSrcLoc, convertType(alignValType), deleteTypeAlign);
     deleteArgs.add(RValue::get(align), alignValType);
   }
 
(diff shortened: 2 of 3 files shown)


clang/test/CIR/CodeGen/aligned-allocation.cpp
```diff
@@ -0,0 +1,43 @@
+// RUN: %clang_cc1 -std=c++17 -triple x86_64-unknown-linux-gnu -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck --check-prefix=CIR --input-file=%t.cir %s
+// RUN: %clang_cc1 -std=c++17 -triple x86_64-unknown-linux-gnu -fclangir -emit-llvm %s -o %t-cir.ll
+// RUN: FileCheck --check-prefix=LLVM --input-file=%t-cir.ll %s
+// RUN: %clang_cc1 -std=c++17 -triple x86_64-unknown-linux-gnu -emit-llvm %s -o %t.ll
+// RUN: FileCheck --check-prefix=OGCG --input-file=%t.ll %s
+
+// C++17 added aligned-allocation overloads of operator new / delete:
```
erichkeane: Please don't do this.  These are useless as soon as the review is done.  Please stop doing these.

clang/test/CIR/CodeGen/aligned-allocation.cpp
```diff
@@ -0,0 +1,43 @@
...
+  Big *p = new Big;
+  delete p;
+}
+
+// Inside f(), the call to operator new passes (size=64, align=64) and
+// operator delete passes (ptr, size=64, align=64).  We use the
+// std::align_val_t-overloaded mangled names of the allocation functions.
+
+// CIR-LABEL: cir.func{{.*}}@_Z1fv
+// CIR:         %{{.+}} = cir.const #cir.int<64> : !u64i
```
erichkeane: These check-lines should be more detailed and show where in the call these are actually going, as that is important.

clang/test/CIR/CodeGen/aligned-allocation.cpp
```diff
@@ -0,0 +1,43 @@
...
+// C++17 added aligned-allocation overloads of operator new / delete:
+//   void* operator new   (std::size_t, std::align_val_t);
+//   void  operator delete(void*, std::size_t, std::align_val_t);
+// These are called when allocating an over-aligned type.  CIRGen
+// previously errorNYI'd the new-side path and the cleanup-during-new
+// path (the delete-side path was already implemented).  This test pins
+// the alignment passing on both sides and cross-checks against OGCG.
+
+struct alignas(64) Big {
+  char data[64];
```
erichkeane: Can we choose different values than 64, so that they show up meaningfully in the check-lines?

clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
```diff
@@ -897,8 +897,12 @@ class CallDeleteDuringNew final
     // is an enum whose underlying type is std::size_t.
     // FIXME: Use the right type as the parameter type. Note that in a call
     // to operator delete(size_t, ...), we may not have it available.
-    if (isAlignedAllocation(params.Alignment))
-      cgf.cgm.errorNYI("CallDeleteDuringNew: aligned allocation");
+    if (isAlignedAllocation(params.Alignment)) {
+      QualType sizeType = cgf.getContext().getSizeType();
+      cir::ConstantOp align = cgf.getBuilder().getConstInt(
```
erichkeane: We do this often enough, CIRBaseBuilder should probably have a `getAlignment` (like it does for `getAlignmentAttr`) so that we cna just do args.add alignment/alignment.getType.

clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
```diff
@@ -1613,7 +1617,17 @@ mlir::Value CIRGenFunction::emitCXXNewExpr(const CXXNewExpr *e) {
 
     // The allocation alignment may be passed as the second argument.
     if (e->passAlignment()) {
-      cgm.errorNYI(e->getSourceRange(), "emitCXXNewExpr: pass alignment");
+      // The corresponding parameter type, if the allocator declares one;
+      // otherwise fall back to size_t (the underlying type of
+      // std::align_val_t).  Mirrors the IndexOfAlignArg / corner-case
+      // handling in classic CodeGen at CGExprCXX.cpp.
```
andykaylor:
```suggestion
      // std::align_val_t). 
```

clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
```diff
@@ -1613,7 +1617,17 @@ mlir::Value CIRGenFunction::emitCXXNewExpr(const CXXNewExpr *e) {
...
+      QualType alignValType = sizeType;
+      if (allocatorType->getNumParams() > paramsToSkip)
+        alignValType = allocatorType->getParamType(paramsToSkip);
```
andykaylor: While this is equivalent to the IndexOfAlignArg handling in classic codegen, it feels really strange that we're using `paramsToSkip` as the index of the align argument. Also, given that these are different values that happen to have the same value in classic codegen, it's easy to imagine some change being introduced that desynchronizes them. I think it might be better to add `indexOfAlignArg` explicitly.

clang/test/CIR/CodeGen/aligned-allocation.cpp
adams381: Done in da85fec — dropped the preamble block.

clang/test/CIR/CodeGen/aligned-allocation.cpp
adams381: Done in da85fec — captured the size and alignment SSA names with FileCheck variables so it is visible which constant lands in each argument slot of _ZnwmSt11align_val_t / _ZdlPvmSt11align_val_t.

clang/test/CIR/CodeGen/aligned-allocation.cpp
adams381: Done in da85fec — changed to alignas(32) char[80] so size 96 and align 32 show up as distinct values in the check lines. (alignas(16) does not actually trigger aligned allocation on x86_64 since __STDCPP_DEFAULT_NEW_ALIGNMENT__ is 16, so I bumped to 32.)

clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
adams381: Done in da85fec — added cir::CIRBaseBuilderTy::getAlignment(loc, mlir::Type, clang::CharUnits) next to getAlignmentAttr. All three callsites in CIRGenExprCXX.cpp now use it: the two new sites this PR introduces plus the existing emitDeleteCall site (the latter is NFC, folded in for consistency now that the helper exists).

clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
adams381: Done in da85fec — applied the suggestion.

clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
adams381: Done in da85fec — introduced constexpr unsigned indexOfAlignArg = 1 with a one-line comment explaining why. Both lookups (getNumParams() > indexOfAlignArg and getParamType(indexOfAlignArg)) use it; paramsToSkip continues as the running counter for the downstream emitCallArgs call.

clang/include/clang/CIR/Dialect/Builder/CIRBaseBuilder.h
```diff
@@ -800,6 +800,14 @@ class CIRBaseBuilderTy : public mlir::OpBuilder {
     return alignment ? getI64IntegerAttr(alignment) : mlir::IntegerAttr();
   }
 
+  // Materialize an alignment value as a CIR integer constant of the given
+  // integer type.  Used to pass std::align_val_t (or its size_t fallback) to
```
erichkeane: 2nd sentence of the comment here isn't valuable/is misleading, the intent is to use this anywhere we need an alignment constant.

erichkeane (approved):

clang/include/clang/CIR/Dialect/Builder/CIRBaseBuilder.h
adams381: Dropped the second sentence.

PR #197096, opened 2026-05-12
[X86][AVX10.2] Add BF16 to (U/S)8 saturating FP to int lowering 

akhilgoe:
This PR adds BF16 to I8 saturating FP to int convert custom lowering.


phoebewang (approved):
LGTM.

PR #197097, opened 2026-05-12
[DAGCombiner] Use KnownBits in `combineFMulOrFDivWithIntPow2`

el-ev:
There's a TODO for this.

Use `computeKnownBits` to tighten the high bit width bound via `countMaxActiveBits()`, which accounts for known leading zeros.

### DIFF
--- a/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
+++ b/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
@@ -19029,14 +19029,17 @@ SDValue DAGCombiner::combineFMulOrFDivWithIntPow2(SDNode *N) {
 
     ConstOp = peekThroughBitcasts(N->getOperand(ConstOpIdx));
     Pow2Op = N->getOperand(1 - ConstOpIdx);
-    if (Pow2Op.getOpcode() != ISD::UINT_TO_FP &&
-        (Pow2Op.getOpcode() != ISD::SINT_TO_FP ||
-         !DAG.computeKnownBits(Pow2Op).isNonNegative()))
+    unsigned Pow2Opc = Pow2Op.getOpcode();
+    if (Pow2Opc != ISD::UINT_TO_FP && Pow2Opc != ISD::SINT_TO_FP)
       return false;
 
     Pow2Op = Pow2Op.getOperand(0);
 
-    int MaxExpChange = DAG.computeKnownBits(Pow2Op).countMaxActiveBits();
+    KnownBits Pow2OpKnownBits = DAG.computeKnownBits(Pow2Op);
+    if (Pow2Opc == ISD::SINT_TO_FP && !Pow2OpKnownBits.isNonNegative())
+      return false;
+
+    int MaxExpChange = Pow2OpKnownBits.countMaxActiveBits();
 
     auto IsFPConstValid = [N, MaxExpChange, &Mantissa](ConstantFPSDNode *CFP) {
       if (CFP == nullptr)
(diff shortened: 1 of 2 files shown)


llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
```diff
@@ -19036,9 +19034,7 @@ SDValue DAGCombiner::combineFMulOrFDivWithIntPow2(SDNode *N) {
 
     Pow2Op = Pow2Op.getOperand(0);
 
-    // `Log2(Pow2Op) < Pow2Op.getScalarSizeInBits()`.
-    // TODO: We could use knownbits to make this bound more precise.
-    int MaxExpChange = Pow2Op.getValueType().getScalarSizeInBits();
+    int MaxExpChange = DAG.computeKnownBits(Pow2Op).countMaxActiveBits();
```
jayfoad: Should avoid recalculating this if it was already calculated on line 19032

jayfoad (commented):
Not related to your patch, but where does this function check that Pow2Op really is a power of two?

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
el-ev: They don't really operate on the same node, fixed it though. The code here is a bit messy.

el-ev:
```
  SDValue Log2 = BuildLogBase2(Pow2Op, DL, DAG.isKnownNeverZero(Pow2Op),
                               /*InexpensiveOnly*/ true, NewIntVT);
```

This call guarantees power of two.

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
```diff
@@ -19029,16 +19030,17 @@ SDValue DAGCombiner::combineFMulOrFDivWithIntPow2(SDNode *N) {
...
+    unsigned Pow2Opc = Pow2Op.getOpcode();
+    if (Pow2Opc != ISD::UINT_TO_FP && Pow2Opc != ISD::SINT_TO_FP)
       return false;
 
     Pow2Op = Pow2Op.getOperand(0);
 
-    // `Log2(Pow2Op) < Pow2Op.getScalarSizeInBits()`.
-    // TODO: We could use knownbits to make this bound more precise.
-    int MaxExpChange = Pow2Op.getValueType().getScalarSizeInBits();
+    Pow2OpKnownBits = DAG.computeKnownBits(Pow2Op);
```
RKSimon: Why split off the declaration?
```suggestion
    KnownBits Pow2OpKnownBits = DAG.computeKnownBits(Pow2Op);
```

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
```diff
@@ -19089,7 +19091,7 @@ SDValue DAGCombiner::combineFMulOrFDivWithIntPow2(SDNode *N) {
       *DAG.getContext(),
       EVT::getIntegerVT(*DAG.getContext(), VT.getScalarSizeInBits()));
 
-  SDValue Log2 = BuildLogBase2(Pow2Op, DL, DAG.isKnownNeverZero(Pow2Op),
+  SDValue Log2 = BuildLogBase2(Pow2Op, DL, Pow2OpKnownBits.isNonZero(),
```
RKSimon: Hopefully this is enough, but isKnownNeverZero is a lot more powerful than knownbits

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
el-ev: edited: updated

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
el-ev: it is indeed regressing, reverted this part

arsenm (approved):

RKSimon (approved):
LGTM

PR #197099, opened 2026-05-12
[GlobalISel] Use KnownBits to determine zeros and ones in CombineHelper.cpp

el-ev:
There are two TODOs for this.

Use `GISelValueTracking::getKnownBits()` in `tryFoldSelectOfConstants` and `tryFoldBoolSelectToLogic` to detect zeroes and all-ones operands that aren't constants.


jayfoad (changes requested):
This just feels wrong to me. If KnownBits can determine that some value is a constant then it should already have been combined into an actual `G_CONSTANT` or similar.

To put it another way, we don't want every place that checks for a constant to have to do an explicit KnownBits check in addition to calling `getIConstantVRegValWithLookThrough`.

Issue #169768, opened 2025-11-27
[Modules] Clang modules require header files to be present even when using precompiled modules


atetubou:
This is derived from https://issues.chromium.org/423815877

We see include not found error in clang modules build when we don't send header files to remote workers even if the header is given as Clang modules and with `-fmodules-embed-all-files`.
So I'm wondering why clang still requires header files to be there even if those files are included in given module file?

This reproduces with the following script.

```sh
cat > module.modulemap <<<"""
module A {
  header \"A.h\"
}
"""

touch A.h

cat > A.c <<<"""
#include \"A.h\"
"""

clang -cc1 -emit-module -o A.pcm -fmodules module.modulemap -fmodules-embed-all-files -fmodule-name=A
clang -cc1 -emit-obj A.c -fmodules -fmodule-map-file=module.modulemap -fmodule-file=A.pcm

rm A.h

# This compile failed with: 'A.h' file not found
clang -cc1 -emit-obj A.c -fmodules -fmodule-map-file=module.modulemap -fmodule-file=A.pcm
```

PR #197102, opened 2026-05-12
[Modules] Fix lazy-loading buffer inconsistency in createFileID under…

atetubou:
… home-is-cwd

This change is a preparatory prerequisite for fixing issue #169768 (missing header error with embedded files under lazy-loading). It resolves an internal compiler inconsistency that occurs when using precompiled modules (PCM) with embedded files under `-fmodule-map-file-home-is-cwd`.

### Root Causes:
Under some circumstances (e.g., home-is-cwd), virtual files are registered in FileManager beforehand due to Bitstream eagerness (e.g. deserializing pragma diagnostic mappings), but their corresponding SLocEntries (source location records containing the embedded buffer Blob) remain lazy-loaded. When Preprocessor creates a new FileID and attempts to read the buffer, it bypasses PCM's embedded buffer (since SLocEntry is not yet loaded) and falls back to reading from the physical disk. Because the physical header is missing, this triggers a fatal `cannot open file` error (assertion/I/O failure) and crashes compilation.

### Fixes:
1. **Force SLocEntry Deserialization (Buffer Fill)**:
   - Added `FileIDLookup` map in `SourceManager` to cache FileID by FileEntryRef in O(1).
   - Updated `SourceManager::createFileID` to check if the target file already has an existing Loaded FileID. If so, it forces SLocEntry deserialization (`getSLocEntry`) to fill the MemoryBuffer cache from PCM before creating the new FileID, avoiding incorrect fallbacks to disk.
   - **Assertion Crash Fix**: Checked `MainFileID.isValid()` before forcing deserialization to prevent assertion crashes inside `ASTReader::getImportLocation` during the early `Chained Includes` setup phase when no main file is active.
2. **Test Coverage**:
   - Updated `clang/test/Modules/home-is-cwd-search-paths.c` expectations as missing embedded headers are now successfully resolved from PCM instead of triggering a fatal inconsistency error.

### DIFF
--- a/clang/include/clang/Basic/SourceManager.h
+++ b/clang/include/clang/Basic/SourceManager.h
@@ -714,6 +714,10 @@ class SourceManager : public RefCountedBase<SourceManager> {
   /// as they do not refer to a file.
   std::vector<SrcMgr::ContentCache*> MemBufferInfos;
 
+  /// A cache to speed up FileID lookup by FileEntryRef, avoiding expensive
+  /// linear scans through LoadedSLocEntryTable and LocalSLocEntryTable.
+  mutable llvm::DenseMap<const FileEntry *, FileID> FileIDLookup;
+
   /// The table of SLocEntries that are local to this module.
   ///
   /// Positive FileIDs are indexes into this table. Entry 0 indicates an invalid
--- a/clang/lib/Basic/SourceManager.cpp
+++ b/clang/lib/Basic/SourceManager.cpp
@@ -339,6 +339,7 @@ void SourceManager::clearIDTables() {
   LastLookupStartOffset = LastLookupEndOffset = 0;
 
   IncludedLocMap.clear();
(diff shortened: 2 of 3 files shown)


zmodem:
Both the PR and its description look AI generated. Please see https://llvm.org/docs/AIToolPolicy.html for how to disclose the tool usage.

I'm having trouble parsing this change. Maybe it's because I'm not too familiar with modules, but it would be great if the PR description could be made clearer.

I'm also wondering about the updated test file: it's changing the expectation from an error to no-diagnostic. Is that really right? I.e. was the test just broken before?

atetubou:
Thank you for taking a look. And yes, I used AI to generate this PR. I'll check the policy again and update description.

atetubou:
After some more discussion with AI, I changed the approach for this issue to https://github.com/llvm/llvm-project/pull/197627.

PR #197103, opened 2026-05-12
[WebAssembly][GlobalISel] Implement conversion/cast ops between integer and floating-point

QuantumSegfault:
Adds legalization and tests for operations converting between int and float (`G_BITCAST`, `G_FPTOSI`, `G_FPTOUI`, `G_SITOFP` and `G_UITOFP`).

Saturating float => int ops will be finished later once comparison operators are in place (which for this, `G_SELECT`, `G_FCMP`, `G_UMIN`, and `G_SMIN` are needed at least).


llvm/lib/Target/WebAssembly/GISel/WebAssemblyLegalizerInfo.cpp
```diff
@@ -103,6 +103,34 @@ WebAssemblyLegalizerInfo::WebAssemblyLegalizerInfo(
       .clampScalar(0, s32, s32)
       .clampScalar(1, s64, s64);
 
+  getActionDefinitionsBuilder(G_BITCAST)
+      .legalFor({{i32, f32}, {f32, i32}, {i64, f64}, {f64, i64}})
```
arsenm, line 107: I thought we had a form of this for the cross product of types, legalForCartesianProduct

QuantumSegfault: True. But this is not a full Cartesian product, but rather a specific subset of only f32<=>i32 and f64<=>i64 (no bitcast to its own type, or across sizes)

arsenm: You could have one for each size, I suppose

arsenm (approved):

llvm/lib/Target/WebAssembly/GISel/WebAssemblyLegalizerInfo.cpp
QuantumSegfault: No, cause then `.legalForCartesianProduct({i32, f32}, {i32, f32})` would allow bitcasting i32 to i32, and f32 to f32 (which isn't valid).

arsenm: It would be `.legalForCartesianProduct({i32, f32})`

Also nop bit casts should be legal. The IR allows them

QuantumSegfault: https://github.com/llvm/llvm-project/blob/db436f826bb97a1a36d300eaa8f8f5317fb66686/llvm/lib/CodeGen/GlobalISel/IRTranslator.cpp#L1584-L1598

IRTranslator emits `COPY` instead in such cases.

QuantumSegfault:
Closing of in favor of the otherwise identical #197255 (so it's part of the Graphite stack)

PR #197104, opened 2026-05-12
[libc++] Fix filesystem::remove_all bug with read-only nested directory

### LINKED ISSUE #197100: [libc++] filesystem::remove_all returns wrong error code on a read-only nested directory
Originally reported here: https://github.com/nodejs/node/issues/62790

Test case:
```C++
#include <iostream>
#include <filesystem>
#include <system_error>
#include <sys/stat.h>

int main() {
    std::filesystem::create_directories("parent/child");
    
    // Make parent read-only to trigger permission error
    chmod("parent", 0555);
    
    std::error_code error;
    std::filesystem::remove_all("parent", error);
    
    std::cout << "error.value() = " << error.value() << std::endl;
    std::cout << "error.message() = " << error.message() << std::endl;
    
    chmod("parent", 0755);
    std::filesystem::remove_all("parent");

    return 0;
}
```
AIX Output:
```
error.value() = 17
error.message() = File exists
```
MacOS Output:
```
error.value() = 66
error.message() = Directory not empty
```

It's expected for `remove_all` to fail with a permission error rather than the above errors.

jakeegan:
Consider a directory `parent/child`, where `parent` has read-only permissions. Calling `remove_all` on `parent` is expected to fail with a permission error, however it fails with a different error depending on the platform (such as "error file exists" on AIX). 

During the recursive removal, `unlinkat` is first called on the child and fails with a permission error. Rather than returning, the function continues and tries `unlinkat` on the parent and fails with a different error depending on the platform. This error is returned instead of the permission error. This PR makes the first encountered error be returned instead. 

According to the standard [fs.op.remove.all]:
> Effects: Recursively deletes the contents of p if it exists, then deletes file p itself, as if by POSIX remove.

The word "then" implies an ordering where deleting the contents of p should be successful before deleting file p itself. 

Fixes #197100.


hubert-reinterpretcast (commented):
Please add a test.

jakeegan:
Do you have an idea what that test would look like? I didn't add a test because I'm not sure there's a way to verify this change without checking the specific error code that is returned, which can vary by platform. This change is ensuring the first encountered error is returned, but I don't think the standard mandates specifically what that error should be. I could add a test that checks any error is returned in this scenario, but it wouldn't verify this functional change.

hubert-reinterpretcast:
> Do you have an idea what that test would look like?

Perhaps we can create a `libcxx` (not `std`) test marked as unsupported for Windows and make use of POSIX APIs and semantics?

We can expect `std::errc::permission_denied` to be associated with removing the child if the parent directory is readable and writable but not searchable (i.e., `0600`).

jakeegan:
Remaining CI failures don't appear to be related

libcxx/test/libcxx/input.output/filesystems/fs.op.funcs/fs.op.remove_all/bad_perms_parent.pass.cpp
```diff
@@ -0,0 +1,41 @@
...
+// when the parent directory has insufficient permissions.
+
+// <filesystem>
+
+#include <filesystem>
+
+#include "test_macros.h"
+#include "filesystem_test_helper.h"
+namespace fs = std::filesystem;
+using namespace fs;
```
philnik777: Please avoid `using namespace`.

libcxx/test/std/input.output/filesystems/fs.op.funcs/fs.op.remove_all/bad_perms_parent.pass.cpp
philnik777, line 1: I don't think this should be libc++-specific. IIUC you're claiming this is required by the standard. We may not be able to test this on all platforms, but that doesn't make it libc++-specific.

hubert-reinterpretcast (approved):
LGTM!

jakeegan:
Ping (libc++ reviewers)

jakeegan:
Ping

ldionne (approved):

PR #197106, opened 2026-05-12
[OpenMP][Flang][Clang][Offload] Prevent shadow pointer write-back to null host pointers

agozillon:
This patch attempts to solve an issue encountered in the UMT Fortran benchmark, where setting up a derived (record) type map and then updating it to host can cause some unideal map backs of other elements of the derived type due to shadow pointer registration.

In this particular reproducing example we end up re-allocating prior deallocated data in a subsequent iteration of a for loop that updates a different variable, which seems to be triggered by the shadow pointer system.

### DIFF
--- a/offload/test/offloading/fortran/no-null-write-back.f90
+++ b/offload/test/offloading/fortran/no-null-write-back.f90
@@ -1,0 +1,70 @@
+! OpenMP offloading regression test that checks we do not write-back data
+! in cases where the host descriptor and data have been deallocated during the
+! time the data was on device. This predominantly crops up as an issue in the
+! UMT benchmark.
+!
+! The key component of this test is that we map back elements of an array inside
+! of a derived type, and then want to deallocate them on the host, but subsequent
+! maps back from device of other elements re-allocate/re-map back the old data we
+! just went deallocated on host.
+! REQUIRES: flang, amdgpu
+
+! RUN: %libomptarget-compile-fortran-run-and-check-generic
+program prog
+   implicit none
+
+   type :: dtype_a
+      real(8),      pointer, contiguous :: array2d(:,:) => null()
+      integer :: scalar_int
+   end type dtype_a
+
+   type :: dtype_b
+      type(dtype_a),      pointer     :: array1d(:) => null()
+   end type dtype_b
+
+   integer :: i, j = 4
+   type(dtype_b), pointer :: alloca_dtype => null()
+
+   allocate(alloca_dtype)
+   allocate(alloca_dtype%array1d(j))
+   do i = 1, j
+      allocate(alloca_dtype%array1d(i)%array2d(10,10))
+   enddo
+
+    !$omp target enter data map(to:alloca_dtype)
+    !$omp target enter data map(always,to:alloca_dtype%array1d)
+   do i = 1, j
+    !$omp target enter data map(always,to:alloca_dtype%array1d(i)%array2d)
+   end do
+
+    ! Check everything is all still associated as we'd expect.
+    do i = 1, j
+        if (associated(alloca_dtype%array1d(i)%array2d)) then
+          write (*,*) "iteration", i, "is associated"
+       endif
+    enddo
+
+! In this loop, we map back scalar_int and array2d of the corresponding
+! array1d iteration, however, subsequent map backs of scalar_int, will
+! re-allocate and re-associate the previous iteration of array2d.
+    do i = 1, j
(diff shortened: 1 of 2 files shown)


agozillon:
So, this might not be the best fix for this issue, I am not the most runtime savvy, so I'm more than happy to take suggestions for other adjustments to solve this problem :-) Perhaps the UMT benchmark is even currently doing something not so legal, by breaking the maps of a larger object up like this (as I am fairly certain there's some wording to that affect in the OpenMP specification, but I'm unsure if it's applicable in this scenario). I will mention that this adjustment to the runtime has been part of our downstream branch for about a year now and it's not been flagged up as a problem by any of our testing infrastructure or client issues.

Please do feel free to add other reviewers if you think there's someone that would be good as an additional reviewer!

offload/test/offloading/fortran/no-null-write-back.f90
```diff
@@ -0,0 +1,70 @@
...
+          write (*,*) "iteration", i, "is associated"
+       endif
+    enddo
+
+! In this loop, we map back scalar_int and array2d of the corresponding
+! array1d iteration, however, subsequent map backs of scalar_int, will
+! re-allocate and re-associate the previous iteration of array2d.
+    do i = 1, j
+!$omp target update from(alloca_dtype%array1d(i)%scalar_int)
+!$omp target update from(alloca_dtype%array1d(i)%array2d)
```
abhinavgaba, line 52: It's a test-case error. This line should have been:

```f90
!$omp target exit data map(delete, from: alloca_dtype%array1d(i)%array2d)

```

With that, the test works as expected with flang on x86_64:

```
❯ flang-new -O0 -fopenmp -fopenmp-targets=x86_64 map_struct_with_dv_then_deallocate.f90 -fopenmp-version=61 && ./a.out
warning: The specification for OpenMP version 61 is still under development; the syntax and semantics of new features may be subject to change
warning: The specification for OpenMP version 61 is still under development; the syntax and semantics of new features may be subject to change
warning: The specification for OpenMP version 61 is still under development; the syntax and semantics of new features may be subject to change
 iteration 1 is associated
 iteration 2 is associated
 iteration 3 is associated
 iteration 4 is associated
 iteration 1 is unassociated
 iteration 2 is unassociated
 iteration 3 is unassociated
 iteration 4 is unassociated
```

An attached pointer stays attached for its lifetime on device, and if a pointer is attached on device, it's illegal to nullify/deallocate it on host. OpenMP 6.1 will add support for force-re-attachment using `attach(always)`, but detaching is not supported in 6.1 either.

As per OpenMP 6.0, it is not legal to even re-associate an attached-pointer to a different (non-null) target:

>OpenMP 6.0 page 288
>• If a list item of a map clause is an allocatable variable or is the subobject of an allocatable variable, the original list item must not be allocated, deallocated or reshaped while the corresponding list item has allocated storage.
>• A pointer that has a corresponding pointer that is an attached pointer and is associated with a given pointer target must not become associated with a different pointer target for the duration of the lifetime of the list item to which the corresponding pointer is attached in the device data environment

agozillon: Thank you Abhinav, I forgot about that rule! That might be an error in UMT then, or this test doesn't quite fully capture what they're trying to do. I'll double check, in the near-ish future and update.

abhinavgaba: If they cannot use OpenMP 6.1, the 6.0 version would be to first do an `update from`, then `delete`:

```f90
!$omp target update from(alloca_dtype%array1d(i)%array2d)
!$omp target exit data map(delete:alloca_dtype%array1d(i)%array2d)

```

agozillon: Thank you very much for the tip!

I imagine it's fine for them to use OpenMP 6.0/6.1, Flang doesn't have particularly strict versioning at the moment, at least from recollection, but Krzsyztof's did a chunk of work in that area a while back so it's perhaps a now restricted case :-)

I'll need to re-verify what it is exactly they were intending to do, as it's been around a year since I looked into it unfortunately, and also test whatever there latest version is to verify it's still a present issue (could also just be a bug we had that was unrelated to this fix that's now fixed inadvertently or perhaps they altered the bit of code themselves)!

But in either case it's not a huge rush, was just something we had downstream that I wanted to get some clarity on so we can resolve it in one way or another, having the exact rule is incredibly helpful, so thank you :-)

agozillon:
Going to close this as it's been addressed now external to LLVM, thank you very much Abhinav for your help!

PR #196769, opened 2026-05-09
workflows/release-doxygen: Add some security checks and input validation

tstellar:
We now ensure the job was started by a release manager before granting the contents: write permissions and we also validate the input to ensure it is a proper release string and not something malicious.

PR #197107, opened 2026-05-12
[github] Fix invalid local action invocation in release-doxygen workflow

kwk:
Fix the `Validate Input` step in `.github/workflows/release-doxygen.yml` to use a valid local action invocation.

Before:
```yaml
- name: Validate Input
  ./.github/workflows/validate-release-version
  with:
    release-version: ${{ inputs.release-version }}
```

After:
```yaml
- name: Validate Input
  uses: ./.github/workflows/validate-release-version
  with:
    release-version: ${{ inputs.release-version }}
```

GitHub Actions steps must use `uses:` or `run:`. The current form is invalid.

This appears to have been introduced in #196769.

- Verified the workflow syntax change is minimal and correct.
- No functional behavior intended beyond fixing the invalid step declaration.


c-rhodes (approved):
LGTM thanks for fixing

PR #197108, opened 2026-05-12
[Docs] Project governance documentation

petrhosek:
This page largely covers the same information as proposal LP-0004 and also includes information on current area team and project council composition.


llvm/docs/ProjectGovernance.rst
rnk, line 185: The formatting of the various lists below didn't come out well:

Image

rnk: We need to expand on this section, we need to define the voting body (all LLVM org members). I think it's worth including the specific dates for when area team candidate applications are open, when elections are open, when terms begin, etc. Those specific details are what people ask about the most, they should be highly searchable on the web.

rnk (commented):
Can we iterate on this collaboratively, rather than me sending comments? There are a few ways we can do it. We could use user branches on the main llvm/llvm-project repo, or you can add me as a collaborator on your petrhosek/llvm-project github fork and I could just push changes to this branch there.

Another thing we should document: The process for forming a new area team, people ask about this a lot.

rnk:
I asked for LLM feedback, and this was the [complete feedback](https://gist.github.com/rnk/24a799a1524e66e246f506c2c98c7229). I think the only parts that matter are:
* Election process details
* New team formation creation

The rest are nits / hallucinations.

rnk (approved):
We should merge this as is, and iterate in tree. I got distracted, and wasn't able to find my way back to this PR to help move it forward.

llvm/docs/ProjectGovernance.rst
petrhosek: Done, I included the section on elections.

petrhosek: This should be resolved in the latest revision and I verified locally that it renders correctly.

petrhosek:
> We should merge this as is, and iterate in tree. I got distracted, and wasn't able to find my way back to this PR to help move it forward.

SGTM I was going to ask for the same. I addressed your comments in the latest revision, but for further improvements, I think it's easier to iterate in tree.

PR #194662, opened 2026-04-28
[CodeGen] Fold load into LZCNT/TZCNT/POPCNT after compare elimination in peephole optimizer

Michael-Chen-NJU:
When a loaded value feeds both a bit-count instruction (LZCNT/TZCNT/POPCNT) and an `icmp eq zero` / `select` sequence, ISel emits the register form (`rr`) because the loaded value has two uses at that point. The peephole optimizer later eliminates the redundant compare (recognizing that LZCNT/TZCNT/POPCNT already set the necessary EFLAGS), but failed to retry folding the load into the now single-use bit-count instruction — because that instruction had already been
scanned.

The fix pre-saves the compared register before `optimizeCmpInstr` erases the compare. After successful elimination, if the EFLAGS producer and its defining load are both already-scanned and now have a single use, fold immediately via `TII->optimizeLoadInstr`.

Before:
```
movq  (%rdi), %rax
lzcntq %rax, %rax
cmovbq %rsi, %rax
```

After:
```
lzcntq (%rdi), %rax
cmovbq %rsi, %rax
```

Fixes #164782

PR #197110, opened 2026-05-12
[CodeGen][NFC] Extract foldLoadInto helper in PeepholeOptimizer

Michael-Chen-NJU:
Pull the load-folding bookkeeping out of `PeepholeOptimizer::run` into a new `PeepholeOptimizer::foldLoadInto` helper.

No functional change intended. 

This is a preliminary NFC split out from #194662 per @RKSimon's review suggestion:

  > Still think this is worth pulling out as its own NFC PR

  The follow-up patch (#194662) adds a second call site that folds a load into an EFLAGS producer after `optimizeCmpInstr` erases the compare, and will reuse this helper instead of duplicating the bookkeeping.

### DIFF
--- a/llvm/lib/CodeGen/PeepholeOptimizer.cpp
+++ b/llvm/lib/CodeGen/PeepholeOptimizer.cpp
@@ -497,7 +497,8 @@ private:
   /// Try to fold the load defined by \p FoldReg into \p MI using
   /// TII->optimizeLoadInstr. On success, updates \p LocalMIs, erases the old
   /// instructions, and returns the replacement; returns nullptr otherwise.
-  MachineInstr *foldLoadInto(MachineInstr &MI, Register FoldReg,
+  MachineInstr *foldLoadInto(MachineFunction &MF, MachineInstr &MI,
+                             Register FoldReg,
                              SmallPtrSet<MachineInstr *, 16> &LocalMIs);
 
   /// Check whether \p MI is understood by the register coalescer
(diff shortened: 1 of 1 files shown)


llvm/lib/CodeGen/PeepholeOptimizer.cpp
```diff
@@ -1392,6 +1398,31 @@ bool PeepholeOptimizer::isLoadFoldable(
...
+PeepholeOptimizer::foldLoadInto(MachineInstr &MI, Register FoldReg,
+                                SmallPtrSet<MachineInstr *, 16> &LocalMIs) {
+  Register Reg = FoldReg;
+  MachineInstr *DefMI = nullptr;
+  MachineInstr *CopyMI = nullptr;
+  MachineInstr *FoldMI = TII->optimizeLoadInstr(MI, MRI, Reg, DefMI, CopyMI);
+  if (!FoldMI)
+    return nullptr;
+  LLVM_DEBUG(dbgs() << "Replacing: " << MI);
+  LLVM_DEBUG(dbgs() << "     With: " << *FoldMI);
```
arsenm:
```suggestion
  LLVM_DEBUG(dbgs() << "Replacing: " << MI
   << "     With: " << *FoldMI);
```

llvm/lib/CodeGen/PeepholeOptimizer.cpp
```diff
@@ -1392,6 +1398,31 @@ bool PeepholeOptimizer::isLoadFoldable(
...
+  LocalMIs.erase(&MI);
+  LocalMIs.erase(DefMI);
+  LocalMIs.insert(FoldMI);
+  if (CopyMI)
+    LocalMIs.insert(CopyMI);
+  if (MI.shouldUpdateAdditionalCallInfo())
+    MI.getMF()->moveAdditionalCallInfo(&MI, FoldMI);
```
arsenm: Shouldn't find the function from the instruction

arsenm (approved):

arsenm (approved):

PR #197111, opened 2026-05-12
[AArch64][GlobalISel] Treat vector select as legal

davemgreen:
AArch64 has a BSP psuedo that lowers to BIF/BIT/BSL and acts like a `or(and(m, x), and(not(m), y))`, to selects between x and y depending on the mask. For whatever reason vector selects have always been expanded to and/or in AArch64, being recombined into a BSP later during selection. This can be a little unreliable, with many places where we do not manage to turn them back. Plus it makes adding selection patterns for selects more difficult.

This change that to treat selects as legal, adding some combines that optime the places where and / or is more beneficial.

Currently includes #200131


llvm/test/CodeGen/AArch64/GlobalISel/combine-select-zero.mir
```diff
@@ -0,0 +1,73 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 6
+# RUN: llc -mtriple aarch64 -run-pass=aarch64-prelegalizer-combiner -verify-machineinstrs %s -o - | FileCheck %s
```
arsenm, line 2:
```suggestion
# RUN: llc -mtriple aarch64 -run-pass=aarch64-prelegalizer-combiner %s -o - | FileCheck %s
```

llvm/lib/Target/AArch64/AArch64InstrInfo.td
```diff
@@ -6260,6 +6260,24 @@ def : Pat<(AArch64bsp (v4i32 V128:$Rd), V128:$Rn, V128:$Rm),
 def : Pat<(AArch64bsp (v2i64 V128:$Rd), V128:$Rn, V128:$Rm),
           (BSPv16i8 V128:$Rd, V128:$Rn, V128:$Rm)>;
 
+def : Pat<(v8i8 (select (v8i8 V64:$Rd), V64:$Rn, V64:$Rm)),
```
arsenm, line 6263: These can also trigger for SDAG, so why weren't they already here?

llvm/lib/Target/AArch64/GISel/AArch64LegalizerInfo.cpp
```diff
@@ -1068,13 +1067,39 @@ AArch64LegalizerInfo::AArch64LegalizerInfo(const AArch64Subtarget &ST)
...
+      .maxScalarEltSameAsIf(isVector(0), 1, 0)
+      .clampNumElements(0, v8s8, v16s8)
+      .clampNumElements(0, v4s16, v8s16)
+      .clampNumElements(0, v2s32, v4s32)
+      .clampMaxNumElements(0, s64, 2)
+      .clampMaxNumElements(0, p0, 2)
+      .bitcastIf(isPointerVector(0), [=](const LegalityQuery &Query) {
+        // Bitcast pointers vector to i64.
+        const LLT DstTy = Query.Types[0];
+        return std::pair(0, DstTy.changeElementType(LLT::integer(64)));
```
arsenm, line 1101:
```suggestion
        return {0, DstTy.changeElementType(LLT::integer(64))}
```

PR #193174, opened 2026-04-21
RFC PR: Refactor AArch64 Linux native register context

DavidSpickett:
I've been asked to do this several times, and it's time I actually showed something. This PR is to show reviewers the end point of what will have to be a number of incremenetal PRs for review purposes.

So the code shown here is not perfect but it shows the ways it will eventually simplify what we have, and proves that it will eventually look neat and work as before.

The current class handles registers like this:
* For each set there is:
  * A class member representing the validity of the current data.
  * A class member storing the data for that register set.
  * A set of methods GetFooBuffer(), GetFooSize(), ReadFoo(), WriteFoo(), IsFoo(reg).
* To read a register set we call ReadFoo() which uses ptrace to fill in the buffer got by GetFooBuffer(). The validity bool is set according to the success or not of that operation.
* Same thing for writing, more calls to these GetFoo... methods.

This style could be debated:
1. Why not return data from these readers instead of writing into class global buffers?
2. Why not track validity using optional instead of extra bools?

First time I tackled this, I tried to address those and it became a mess very quickly. So while I don't think the current style is perfect, I'm not trying to change it here.

The problem I'm addressing is that the current design is sprawling and hard to add new sets to. What I am trying to get away from is having to explicitly add GetFoo... N times per new register set.

So the new design is centered around the RegisterKind enum. We had this for save and restore but I have expanded it to account for all register sets (aside from the SME pseudos, but they are unusual).

This means instead of having ReadFooRegisters() you can call a generic method and pass in the kind you want to read. So we implement that method only once.

Adding a new register set will usually be a case of adding an enum value, and cases to various switch statements.

If the set is "special", like the entanglement of FP/SVE/SME is, then we will need special handling but at least it will be clear that these sets are unusual. There's no way to avoid that.

For example, ReadRegister can now attempt to read the set as a "simple set" and only if that fails, go into the nest of FP/SVE/SME handling. So it's very clear where the complexity lies.

At present register details are described by a few functions that switch on the kind: GetPtraceSet, GetSetSize, GetSetBuffer and so on.

This is a sort of "struct of arrays" approach, relying on compiler warnings and asserts to check we haven't missed anything.

In theory it could be done as a "array of struct" where each entry was the complete set of register information for that one set. I haven't tried to implement that though. Maybe it would make adding new sets easier but I don't think by much.

Given that the standard advice for adding a new thing is: take an existing set that behaves the same way and copy everything it does.

That advice works just as well for this pile of switch statements.

I have tested this on the following configurations:
[...]

PR #197113, opened 2026-05-12
[lldb][AArch64][Linux] Track all register cache validity in one place

DavidSpickett:
In this change I've expanded the existing RegisterSetType enum
to be used as part of a singular cache tracking object which replaces
all the separate booleans. These booleans were hard to track and 
set/reset in the right places.

This new validity object can be reset by default constructing,
and uses a member initialiser so everything starts out as invalid.

RegisterSetType is now a bitmask enum for easier storage.
It's 32-bit now which is enough for now. Just in case we expand it. 
(and I previously updated all sizeof calls to use RegisterSetType
instead of the raw type)

The SME and SME 2 entries have been renamed to ZA and ZT
as this is what ptrace calls them. Note that SME's pseudo registers
are still managed in their own unique way, I hope to address that
later.

Over time, RegisterSetType will become the key for all the 
operations in the register context. You can see a preview of
that "vision" in #193174.

For instance getting the buffer location and size, looking up
the ptrace set number, finding out what a set write invalidates,
and so on.

I did consider using optional for the register storage. I think this
could work if we had done it from the beginning but starting from
this point it has too many issues with it.

* We cannot assume that sizeof(buffer) == sizeof(buffer_type). Which leads to some places where you have to write sizeof(some_generic_type). Which is not ideal. Being able to use the member variable name is clearer.
* We have to make the buffer "valid" before trying to read, then make it invalid if it fails. Rather than read and if it succeeds mark it valid. 3 steps vs. 2.
* Some things like SVE resize the buffer when we get the new vector length, then at some point later assume the buffer can hold the result of a read. We'd have to track the expected size separately, and use that to resize it just before read. Which is not difficult, but it's a different strategy and I don't want to disturb all the SVE/SME handling at this stage.
* Though our buffer + bool is in theory an optional, the way the class is designed where one thing fills the buffer and another reads from it later, the flexibility of when to set the bool does help us. Versus the automatic handling optional does.

Once this whole class has been simplified and I can see the "unusual" code paths more easily, I might revisit the optional idea.

### DIFF
--- a/lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.cpp
+++ b/lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.cpp
@@ -235,18 +235,6 @@ NativeRegisterContextLinux_arm64::NativeRegisterContextLinux_arm64(
 
   m_refresh_hwdebug_info = true;
 
-  m_gpr_is_valid = false;
-  m_fpu_is_valid = false;
-  m_sve_buffer_is_valid = false;
-  m_sve_header_is_valid = false;
-  m_pac_mask_is_valid = false;
-  m_mte_ctrl_is_valid = false;
-  m_tls_is_valid = false;
-  m_zt_buffer_is_valid = false;
-  m_fpmr_is_valid = false;
-  m_gcs_is_valid = false;
-  m_poe_is_valid = false;
-
   // SME adds the tpidr2 register
   m_tls_size = GetRegisterInfo().IsSSVEPresent() ? sizeof(m_tls_regs)
                                                  : sizeof(m_tls_regs.tpidr_reg);
@@ -678,7 +666,8 @@ Status NativeRegisterContextLinux_arm64::WriteRegister(
         uint64_t vg_value = reg_value.GetAsUInt64();
 
         if (sve::vl_valid(vg_value * 8)) {
-          if (m_sve_header_is_valid && vg_value == GetSVERegVG())
+          if (m_validity.IsValid(RegisterSetType::SVE_HEADER) &&
+              vg_value == GetSVERegVG())
             return error;
 
           SetSVERegVG(vg_value);
@@ -687,11 +676,12 @@ Status NativeRegisterContextLinux_arm64::WriteRegister(
           if (error.Success()) {
             // Changing VG during streaming mode also changes the size of ZA.
             if (m_sve_state == SVEState::Streaming)
-              m_za_header_is_valid = false;
+              m_validity.Invalidate(RegisterSetType::ZA_HEADER);
             ConfigureRegisterContext();
           }
 
-          if (m_sve_header_is_valid && vg_value == GetSVERegVG())
+          if (m_validity.IsValid(RegisterSetType::SVE_HEADER) &&
+              vg_value == GetSVERegVG())
             return error;
         }
 
@@ -828,25 +818,11 @@ Status NativeRegisterContextLinux_arm64::WriteRegister(
   return Status::FromErrorString("Failed to write register value");
 }
 
-enum RegisterSetType : uint32_t {
-  GPR, // General purpose registers.
-  SVE, // Used for SVE registers in streaming or non-streaming mode.
-  FPR, // When there is no SVE, or SVE in FPSIMD mode, or streaming only SVE
-       // that is in non-streaming mode.
-  // Pointer authentication registers are read only, so not included here.
-  MTE,  // Memory tagging control registers.
-  TLS,  // Thread local storage registers.
-  SME,  // ZA only, because SVCR and SVG are pseudo registers.
-  SME2, // ZT only.
-  FPMR, // Floating point mode control registers.
-  GCS,  // Guarded Control Stack registers.
-  POE,  // Permission Overlay registers.
-};
-
-static uint8_t *AddRegisterSetType(uint8_t *dst,
-                                   RegisterSetType register_set_type) {
-  *(reinterpret_cast<uint32_t *>(dst)) = register_set_type;
-  return dst + sizeof(uint32_t);
+uint8_t *NativeRegisterContextLinux_arm64::AddRegisterSetType(
+    uint8_t *dst,
+    NativeRegisterContextLinux_arm64::RegisterSetType register_set_type) {
+  *(reinterpret_cast<RegisterSetType *>(dst)) = register_set_type;
+  return dst + sizeof(RegisterSetType);
 }
 
 static uint8_t *AddSavedRegistersData(uint8_t *dst, void *src, size_t size) {
@@ -854,9 +830,10 @@ static uint8_t *AddSavedRegistersData(uint8_t *dst, void *src, size_t size) {
   return dst + size;
 }
 
-static uint8_t *AddSavedRegisters(uint8_t *dst,
-                                  enum RegisterSetType register_set_type,
-                                  void *src, size_t size) {
+uint8_t *NativeRegisterContextLinux_arm64::AddSavedRegisters(
+    uint8_t *dst,
+    NativeRegisterContextLinux_arm64::RegisterSetType register_set_type,
+    void *src, size_t size) {
   dst = AddRegisterSetType(dst, register_set_type);
   return AddSavedRegistersData(dst, src, size);
 }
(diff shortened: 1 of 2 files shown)


lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.cpp
```diff
@@ -1010,7 +987,7 @@ Status NativeRegisterContextLinux_arm64::ReadAllRegisterValues(
     // Use the header size not the buffer size, as we may be using the buffer
     // for fake data, which we do not want to write out.
     assert(m_za_header.size <= GetZABufferSize());
-    dst = AddSavedRegisters(dst, RegisterSetType::SME, GetZABuffer(),
+    dst = AddSavedRegisters(dst, RegisterSetType::ZA, GetZABuffer(),
```
DavidSpickett, line 930: RegisterSetType::SME always meant ZA really.

SME registers from the user POV are:
* ZA - a real register with it's own ptrace set
* svcr / svl - pseudo registers we derive from the streaming sve set
* ZT - a real register provided by sme2.

So I've changed SME -> ZA to be more honest about what it tracks. The names won't always line up with ptrace's names exactly but generally we should try to.

lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.cpp
```diff
@@ -1209,14 +1187,21 @@ Status NativeRegisterContextLinux_arm64::WriteAllRegisterValues(
         // exiting streaming mode.
         error = WriteRegisterSet(&ioVec, sve_fpsimd_data.size(), NT_ARM_SVE);
 
+        // Wrote FPU, and SVE overlaps FPU.
+        m_validity.Invalidate(RegisterSetType::FPR, RegisterSetType::SVE_HEADER,
+                              RegisterSetType::SVE);
+
+        m_sve_state = SVEState::Unknown;
+        ConfigureRegisterContext();
+
```
DavidSpickett, line 1130: This is now happening twice, need to check which to keep.

lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.cpp
```diff
@@ -1746,26 +1721,27 @@ Status NativeRegisterContextLinux_arm64::WriteZA() {
...
   ioVec.iov_len = GetZABufferSize();
 
-  m_za_buffer_is_valid = false;
-  m_za_header_is_valid = false;
-  // Writing to ZA may enable ZA, which means ZT0 may change too.
-  m_zt_buffer_is_valid = false;
+  m_validity.Invalidate(
+      RegisterSetType::ZA_HEADER, RegisterSetType::ZA,
+      // Writing to ZA may enable ZA, which means ZT0 may change too.
+      RegisterSetType::ZT);
```
DavidSpickett: You will be seeing some of these invalidate calls repeat, and later I plan to add a register set -> what it invalidates function.

For now they are written out each time.

lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
```diff
@@ -79,21 +79,66 @@ class NativeRegisterContextLinux_arm64
...
+
+    void MakeValid(RegisterSetType set) {
+      m_valid_flags |= static_cast<Storage>(set);
+    }
+    bool IsValid(RegisterSetType set) {
+      return (m_valid_flags & static_cast<Storage>(set)) != 0;
+    }
+  } m_validity;
+
+  static uint8_t *AddRegisterSetType(uint8_t *dst,
```
DavidSpickett: For the moment, this and AddSavedRegisters could be static functions in the cpp file, but making them members allows me to keep RegisterSetType private.

It also makes future refactoring easier, as they need to access other things in the register context later.

lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
```diff
@@ -79,21 +79,66 @@ class NativeRegisterContextLinux_arm64
...
+
+  static uint8_t *AddSavedRegisters(uint8_t *dst,
+                                    RegisterSetType register_set_type,
+                                    void *src, size_t size);
+
+  Status RestoreRegisters(void *buffer, const uint8_t **src, size_t len,
+                          const RegisterSetType set,
+                          std::function<Status()> writer);
+
+  size_t m_tls_size = 0;
```
DavidSpickett, line 135: Note that m_tls_size is not new, it has just moved, and all the bools around it were deleted. In reality it just moved down.

DavidSpickett:
This is the beginning of a refactor @omjavaid has been asking for for a while now, but I welcome anyone reviewing it.

Both because Omair is very busy, and because these changes should be purely mechanical. So if they don't make sense without AArch64 knowledge, then it's on me to explain it or make them better.

By the end of all this the idea is that it'll be eas(ier) to expand and update even if you aren't knee deep in this stuff.

DavidSpickett:
> In this change I've expanded the existing RegisterSetType enum to be used as part of a singular cache tracking object which replaces all the separate booleans. These booleans were hard to track and set/reset in the right places.

An alternative to this would be to use `std::optional<whatever>` for all of the storage. This changes the order of operations to:
* storage.emplace() to make the thing valid
* ptrace(read whatever ....)
* if failure, reset the optional

I still think you would have some MakeValid type function that knows what to emplace, and some things (SME's ZA) would need an explicit validity bool, but it might work and if we could get free asserts from optional, that'd be nice.

I'm going to give that a go as an alternative approach.

DavidSpickett:
Got part way into the optional idea and though I think it would be a good one if we had started with it, it has some usability problems.

* We cannot assume that `sizeof(buffer) == sizeof(buffer_type)`. Which leads to some places where you have to write `sizeof(some_generic_type)`. Which is not ideal. Being able to use the member variable name is clearer.
* We have to make the buffer "valid" before trying to read, then make it invalid if it fails. Rather than read and if it succeeds mark it valid. 3 steps vs. 2.
* Some things like SVE resize the buffer when we get the new vector length, then at some point later assume the buffer can hold the result of a read. We'd have to track the expected size separately, and use that to resize it just before read. Which is not difficult, but it's a different strategy and I don't want to disturb all the SVE/SME handling at this stage.
* Though our buffer + bool is in theory an optional, the way the class is designed where one thing fills the buffer and another reads from it later, the flexibility of when to set the bool does help us. Versus the automatic handling optional does.

Once this whole class has been simplified and I can see the "unusual" code paths more easily, I might revisit the idea.

For now, I'm sticking with the bag of booleans strategy.

DavidSpickett:
ping!

Anyone who isn't familiar with this code you can still review because the point of this work is to make the class easier to work with. So it *should* be an NFC refactor where the result is easier to work with.

lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
```diff
@@ -79,36 +81,45 @@ class NativeRegisterContextLinux_arm64
...
-  bool m_sve_header_is_valid;
-  bool m_za_buffer_is_valid;
-  bool m_za_header_is_valid;
-  bool m_pac_mask_is_valid;
-  bool m_tls_is_valid;
+  RegisterSetType m_validity = static_cast<RegisterSetType>(0);
+
+  void MakeValid(RegisterSetType set) { m_validity |= set; }
+
+  [[nodiscard]] bool IsValid(RegisterSetType set) {
```
Teemperor: Could be `const`

lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
```diff
@@ -79,36 +81,45 @@ class NativeRegisterContextLinux_arm64
...
+    PAC = 1 << 4,        // Pointer authentication mask registers.
+    MTE = 1 << 5,        // Memory tagging control registers.
+    TLS = 1 << 6,        // Thread local storage registers.
+    ZA = 1 << 7,         // ZA only, because SVCR and SVG are pseudo registers.
+    ZA_HEADER = 1 << 8,  // Only the ptrace header for ZA.
+    ZT = 1 << 9,         // ZT only.
+    FPMR = 1 << 10,      // Floating point mode control registers.
+    GCS = 1 << 11,       // Guarded Control Stack registers.
+    POE = 1 << 12,       // Permission Overlay registers.
+    LLVM_MARK_AS_BITMASK_ENUM(POE),
```
Teemperor, line 115: <PersonalOpinion> I like having the comments before each enum case as the inline-comments become hard to read once they spawn several rows or the identifier gets too long.</PersonalOpinion>

Teemperor (approved):
LGTR with a nit.

lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
DavidSpickett: Agreed. I also make people put comments on the line above in general, I should follow my own advice.

PR #197114, opened 2026-05-12
[LV] Add support for partial reduction chains with fsubs.

sdesmalen-arm:
The cost-model prevented this from happening, but the LV would otherwise generate incorrect code (i.e. without the fneg).


llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
```diff
@@ -6119,13 +6124,18 @@ InstructionCost AArch64TTIImpl::getPartialReductionCost(
         llvm::is_contained({MVT::i8, MVT::i16, MVT::i32}, InVT.SimpleTy))
       return Cost * 2;
 
-    // SVE2 fmlalb/t and NEON fmlal(2)
+    // SVE2 fml[as]lb/t and NEON fml[as]l(2)
     if (IsSupported(ST->hasSVE2(), ST->hasFP16FML()) && InVT == MVT::f16)
       return Cost * 2;
 
+    // SVE2p1 bfmlslb/t
+    if (IsSupported(ST->hasSVE2p1() && ST->hasBF16(), false) &&
```
MacDue: It looks like we could support this without +bf16, but the ISEL is currently linked to the legality of the `bfmlalb/t`. I don't think anything needs doing (I expect having +bf16 is pretty likely), but maybe leave a note.

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -6001,14 +6012,23 @@ static void transformToPartialReduction(const VPPartialReductionChain &Chain,
...
-    auto *Zero = Plan.getZero(ElemTy);
-    auto *NegRecipe =
-        new VPWidenRecipe(Instruction::Sub, {Zero, ExtendedOp}, VPIRFlags(),
-                          VPIRMetadata(), DebugLoc::getUnknown());
+    VPWidenRecipe *NegRecipe;
+    if (WidenRecipe->getOpcode() == Instruction::FSub)
+      NegRecipe =
+          new VPWidenRecipe(Instruction::FNeg, {ExtendedOp}, VPIRFlags(),
+                            VPIRMetadata(), DebugLoc::getUnknown());
+    else {
```
MacDue: nit: 
```suggestion
    if (WidenRecipe->getOpcode() == Instruction::FSub) {
      NegRecipe =
          new VPWidenRecipe(Instruction::FNeg, {ExtendedOp}, VPIRFlags(),
                            VPIRMetadata(), DebugLoc::getUnknown());
    } else {
```

llvm/test/Transforms/LoopVectorize/AArch64/partial-reduce-costs.ll
```diff
@@ -664,3 +664,127 @@ exit:
 !33 = distinct !{!33, !34, !35}
 !34 = !{!"llvm.loop.interleave.count", i32 1}
 !35 = !{!"llvm.loop.vectorize.width", i32 4}
+
+; fsub(float, fpext(half)->float * fpext(half)->float)
+define float @sub_reduction_float_fpext_half_fpext_half(ptr %src1, ptr %src2, ptr %src3, float %init, i32 %n) {
+; NEON-LABEL: 'sub_reduction_float_fpext_half_fpext_half'
+; SVE-LABEL: 'sub_reduction_float_fpext_half_fpext_half'
```
MacDue, line 727: Are these deliberately ignored as they don't produce partial reductions? If so, maybe a comment/CHECK-NOT would be good.

llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
```diff
@@ -6119,13 +6124,18 @@ InstructionCost AArch64TTIImpl::getPartialReductionCost(
         llvm::is_contained({MVT::i8, MVT::i16, MVT::i32}, InVT.SimpleTy))
       return Cost * 2;
 
-    // SVE2 fmlalb/t and NEON fmlal(2)
+    // SVE2 fml[as]lb/t and NEON fml[as]l(2)
```
MacDue, line 6125: Not changed here, but I don't think we have any test defending the NEON case with `+fp16fml`.

MacDue (commented):
Generally, LGTM, just a few comments:

llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
sdesmalen-arm: The documentation for [BFMLALB/T](https://developer.arm.com/documentation/ddi0602/2026-03/SVE-Instructions/BFMLALB--vectors---BFloat16-multiply-add-to-single-precision--bottom--?lang=en) says that if FEAT_BF16 is not enabled the instruction is undefined (see "Decode for this encoding").

llvm/test/Transforms/LoopVectorize/AArch64/partial-reduce-costs.ll
sdesmalen-arm: Yes, that's right. I'll add a comment (I prefer that over manually adding a CHECK-NOT line, because then we can keep using the update_analyze_test_checks.py script)

llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
sdesmalen-arm: I've added it to the RUN line so that we now see the partial reduction.

llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
MacDue: This is bfmlslb/t (not bfmlalb/t) though. It appears to only require SME2 or SVE2p1: https://developer.arm.com/documentation/ddi0602/2025-12/SVE-Instructions/BFMLSLB--vectors---BFloat16-multiply-subtract-from-single-precision--bottom--

sdesmalen-arm: Ah, I'll finish my morning coffee first then :) I've fixed the costs and added a RUN line for SME2.

MacDue (approved):
LGTM, thanks 👍

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -6031,21 +6042,27 @@ static void transformToPartialReduction(const VPPartialReductionChain &Chain,
...
     Type *ElemTy = TypeInfo.inferScalarType(ExtendedOp);
-    auto *Zero = Plan.getZero(ElemTy);
-    auto *NegRecipe =
-        new VPWidenRecipe(Instruction::Sub, {Zero, ExtendedOp}, VPIRFlags(),
-                          VPIRMetadata(), DebugLoc::getUnknown());
+    VPWidenRecipe *NegRecipe;
+    if (WidenRecipe->getOpcode() == Instruction::FSub) {
+      NegRecipe =
+          new VPWidenRecipe(Instruction::FNeg, {ExtendedOp}, VPIRFlags(),
+                            VPIRMetadata(), DebugLoc::getUnknown());
```
fhahn, line 6131: Something to maybe investigate as follow-up: could we preserve same of the FMFs from the sub?

llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
```diff
@@ -6057,13 +6057,18 @@ InstructionCost AArch64TTIImpl::getPartialReductionCost(
...
   InstructionCost Cost = InputLT.first * TTI::TCC_Basic;
   // Integer partial sub-reductions that don't map to a specific instruction,
   // carry an extra cost for implementing a double negation:
   //      partial_reduce_umls  acc, lhs, rhs
   // <=> -partial_reduce_umla -acc, lhs, rhs
   InstructionCost INegCost = IsSub ? 2 * InputLT.first * TTI::TCC_Basic : 0;
+  // FP partial sub-reductions that don't map to a specific instruction,
+  // carry an extra cost for implementing an extra negation:
+  //      partial_reduce_umls  acc, lhs, rhs
+  // <=>  partial_reduce_umla  acc, lhs, -rhs
```
fhahn: Should this be fmls/fmla?

llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
```diff
@@ -6057,13 +6057,18 @@ InstructionCost AArch64TTIImpl::getPartialReductionCost(
...
+  InstructionCost FNegCost = IsSub ? InputLT.first * TTI::TCC_Basic : 0;
```
fhahn: could sink to only use

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -3375,17 +3375,23 @@ class VPExpressionRecipe : public VPSingleDefRecipe {
...
     assert(getNumOperands() >= 3 && "Expected at least three operands");
-    [[maybe_unused]] auto *SubConst = dyn_cast<VPConstantInt>(getOperand(2));
-    assert(SubConst && SubConst->isZero() &&
-           Sub->getOpcode() == Instruction::Sub && "Expected a negating sub");
+    if (Neg->getOpcode() == Instruction::Sub) {
+      [[maybe_unused]] auto *SubConst = dyn_cast<VPConstantInt>(getOperand(2));
+      assert(SubConst && SubConst->isZero() &&
+             Neg->getOpcode() == Instruction::Sub && "Expected a negating sub");
+    } else
+      assert(Neg->getOpcode() == Instruction::FNeg && "Unexpected opcode");
```
fhahn, line 3500: the constructors are becoming a bit unwieldy, might be good to try to simplify as follow-up. I think at most (all?) callers we should already know the ExpressionType, so it might be simpler to just have a simple constructor that takes ExpressionType + operands.

sdesmalen-arm: Yes, I agree this is becoming a bit unwieldy. I'll try to simplify that in a follow-up.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -3097,8 +3097,16 @@ InstructionCost VPExpressionRecipe::computeCost(ElementCount VF,
                                           Ctx.CostKind);
 
   case ExpressionTypes::ExtNegatedMulAccReduction:
-    assert(Opcode == Instruction::Add && "Unexpected opcode");
-    Opcode = Instruction::Sub;
+    switch (Opcode) {
```
MacDue, line 3179: (not here) I think the `printRecipe` for `ExtNegatedMulAccReduction` needs updating to (correctly) print the `fneg` case (it'll print it as `sub 0, mul(...))` right now, I think).

fhahn (approved):
LGTM, thanks

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
MacDue: (note: feel free to do this as a follow-up or ignore this, as it's a minor printing suggestion)

PR #197115, opened 2026-05-12
[AArch64] Apply generational deltas for Apple tuning features

tomershafir:
This patch refactors how we manage tuning features for AArch64 Apple CPUs. Instead of duplicating feature lists for each generation, we now form a chain such that every generation only adds or removes features from the immediate predecessor.

This creates a much more compact representation, enables to view generational changes at a glance, and can reduce potential copy-paste errors by establishing a more structured dataflow.

However, not without pitfalls: to gather a feature list for a specific gen we need to traverse a chain, and feature addition and removal becomes more subtle as it propagates across the chain. The TableGen test of complete feature lists per generation may alleviate these issues a bit.


llvm/test/TableGen/aarch64-apple-tuning-features.td
```diff
@@ -13,16 +13,16 @@
 // CHECK-LABEL: def TuneAppleA13 {
 // CHECK: list<SubtargetFeature> Implies = [FeatureAlternateSExtLoadCVTF32Pattern, FeatureArithmeticBccFusion, FeatureArithmeticCbzFusion, FeatureDisableLatencySchedHeuristic, FeatureFuseAES, FeatureFuseCryptoEOR, FeatureStorePairSuppress, FeatureZCRegMoveGPR64, FeatureZCRegMoveFPR128, FeatureZCZeroingGPR32, FeatureZCZeroingGPR64, FeatureNoZCZeroingFPR64, FeatureZCZeroingFPR128];
 // CHECK-LABEL: def TuneAppleA14 {
-// CHECK: list<SubtargetFeature> Implies = [FeatureAggressiveFMA, FeatureAlignCmpCSelPairs, FeatureAlternateSExtLoadCVTF32Pattern, FeatureArithmeticBccFusion, FeatureArithmeticCbzFusion, FeatureDisableLatencySchedHeuristic, FeatureFastLD1Single, FeatureFuseAddress, FeatureFuseAES, FeatureFuseArithmeticLogic, FeatureFuseCmpCSel, FeatureFuseCryptoEOR, FeatureFuseLiterals, FeatureStorePairSuppress, FeatureZCRegMoveGPR64, FeatureZCRegMoveFPR128, FeatureZCZeroingGPR32, FeatureZCZeroingGPR64, FeatureNoZCZeroingFPR64, FeatureZCZeroingFPR128, FeatureMaxInterleaveFactor4];
```
jroelofs: I was hoping that by splitting this and the previous patch, that there would be no change here. Maybe we need to arrange for them to be sorted by name?

tomershafir: I dont think there is currently a sort function in TableGen that we could use to sort complete lists.

I think we also cannot sort them by hand as the order would be different before and after the patch, as the new generational boundaries would produce partially-sorted lists.

We can probably add a new `!listsort`, but I think in this case we may not wanna block this patch

jroelofs: https://github.com/llvm/llvm-project/pull/197303
https://github.com/llvm/llvm-project/pull/197302

jroelofs (approved):
LGTM, thanks!

PR #197116, opened 2026-05-12
[flang][cuda] Place box value kernel args in managed memory

khaki3:
Example:
```fortran
type deviceArray
  integer, allocatable, dimension(:,:), device :: Arr
end type deviceArray
type(deviceArray), allocatable, dimension(:) :: DA

allocate(DA(2))
allocate(DA(1)%Arr(32,32))
call mykernel<<<1,32>>>(DA(1)%Arr, 32)  ! cudaErrorIllegalAddress
```

In this code, `DA(1)%Arr` is a device allocatable component inside a managed derived type. The compiler loads the descriptor, reboxes it on the host stack, and passes it to the kernel. Since `!fir.box` is lowered to a pointer in LLVM IR, the kernel receives a host-stack pointer it cannot dereference — causing `cudaErrorIllegalAddress`.

The existing `isDeviceAllocation` check is def-based: it traces the value's origin to decide whether to use managed memory. It works for global device descriptors and `cuf.data_attr`-annotated arguments, but not for this case — the rebox input is a plain host alloca, and only the *use* as a kernel argument requires managed memory.

**Fix:** Add a use-based `isUsedByGPULaunchFunc` check alongside `isDeviceAllocation` in the embox, rebox, and load conversions in CodeGen. When a box descriptor is passed directly to `gpu.launch_func`, its storage is allocated in managed memory via `_FortranACUFAllocDescriptor` instead of a stack alloca.

### DIFF
--- a/flang/lib/Optimizer/CodeGen/CodeGen.cpp
+++ b/flang/lib/Optimizer/CodeGen/CodeGen.cpp
@@ -2099,6 +2099,15 @@ struct EmboxOpConversion : public EmboxCommonConversion<fir::EmboxOp> {
   }
 };
 
+/// Return true if any user of \p val is a gpu.launch_func operation,
+/// indicating the descriptor must be in device-accessible memory.
+static bool isUsedByGPULaunchFunc(mlir::Value val) {
+  for (auto *user : val.getUsers())
+    if (mlir::isa<mlir::gpu::LaunchFuncOp>(user))
+      return true;
+  return false;
+}
+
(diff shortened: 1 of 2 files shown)


clementval:
There is already some logic in CodeGen.cpp to reallocate boxes in managed memory when needed. Can you check if you can extend it for this case?

wangzpgi (approved):

wangzpgi (changes requested):
Clicked "Approve" by mistake, not sure how to revert. I will let @clementval approve.

clementval (approved):

wangzpgi (approved):

PR #197117, opened 2026-05-12
[AArch64][GlobalISel] Legalize pointer vector icmp through bitcast

davemgreen:
This fixes the lowering of pointer-vector icmp by bitcasting to an integer type, similar to what we do for other operations like vector shuffles (and in the future selects).


arsenm (commented):
It's kind of better to just directly handle these. Why did these fail before? Is this just mishandled pointers by tablegen?

PR #197120, opened 2026-05-12
[RegisterScavenging] Respect early-clobber defs when scavenging registers

dominik-steenken:
When scavenging registers backwards for virtual registers introduced during frame index elimination, the register scavenger was ignoring early-clobber constraints on the instruction using the scavenged register. This could lead to assigning a virtual register to a physical register marked as early-clobber output, violating the constraint that early-clobber outputs cannot overlap with inputs.

This change inspects `RestoreAfter` to determine if the scavenged register will be used by the instruction pointed at by MBBI, and if so, remove any such registers from the scavengeable set.

This also adds a test to check if such EC defs are indeed respected whne they otherwise wouldn't be.

co-authored-by: @uweigand


llvm/lib/CodeGen/RegisterScavenging.cpp
```diff
@@ -141,6 +141,15 @@ findSurvivorBackwards(const MachineRegisterInfo &MRI,
          "Target instruction is in other than current basic block, use "
          "enterBasicBlockEnd first");
 
+  // If RestoreAfter is set, the scavenged register is needed at
+  // std::next(From), so we need to take into account any possible early-clobber
+  // def regs defined there.
+  if (RestoreAfter)
```
arsenm: Braces

llvm/lib/CodeGen/RegisterScavenging.cpp
```diff
@@ -141,6 +141,15 @@ findSurvivorBackwards(const MachineRegisterInfo &MRI,
...
+    for (const MachineOperand &MOP : std::next(From)->operands())
+      if (MOP.isReg() && MOP.getReg().isPhysical() && MOP.isDef() &&
+          MOP.isEarlyClobber())
```
arsenm:
```suggestion
    for (const MachineOperand &MOP : std::next(From)->all_defs())
      if (MOP.getReg().isPhysical() && MOP.isEarlyClobber())
```

llvm/lib/CodeGen/RegisterScavenging.cpp
dominik-steenken: I added braces on the outer if and the for loop.

llvm/test/CodeGen/SystemZ/scavenge-clobbered-reg.mir
```diff
@@ -0,0 +1,53 @@
...
+alignment:       16
+
+tracksRegLiveness: true
+registers:       []
+liveins:
+  - { reg: '$r2d', virtual-reg: '' }
+  - { reg: '$r3d', virtual-reg: '' }
+frameInfo:
+  stackSize:       0
+  maxAlignment:    1
```
arsenm:
```suggestion
tracksRegLiveness: true
```

llvm/test/CodeGen/SystemZ/scavenge-clobbered-reg.mir
```diff
@@ -0,0 +1,53 @@
...
+stack:
+  - { id: 0, name: '', type: default, offset: 0, size: 16384, alignment: 1,
+      stack-id: default, callee-saved-register: '', callee-saved-restored: true,
+      debug-info-variable: '', debug-info-expression: '', debug-info-location: '' }
```
arsenm:
```suggestion
  - { id: 0, offset: 0, size: 16384, callee-saved-restored: true }
```

llvm/lib/CodeGen/RegisterScavenging.cpp
```diff
@@ -141,6 +141,16 @@ findSurvivorBackwards(const MachineRegisterInfo &MRI,
          "Target instruction is in other than current basic block, use "
          "enterBasicBlockEnd first");
 
+  // If RestoreAfter is set, the scavenged register is needed at
```
arsenm, line 144: Why isn't this caught by the accumulate call for Restore after down on line 173?

uweigand: The problematic case is where we don't even reach line 173 due to the early exit in line 164, i.e. when a free register is available.   It is OK to not exclude defs in general here (so we don't want to move that accumulate call earlier), but *earlyclobber* defs must be excluded.

arsenm (approved):

PR #197122, opened 2026-05-12
[lldb][AArch64][Linux] Use member initialisers

DavidSpickett:
Member initialise a bunch of things in the register context instead of setting them all in the constructor with memsets.

The only things I've left are related to hardware breakpoints, and need changes to non-AArch64 classes so I'll try that separately.

I have not changed the validity bools because those will be removed by #197113.

### DIFF
--- a/lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
+++ b/lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
@@ -91,63 +91,56 @@ private:
   bool m_za_header_is_valid;
   bool m_pac_mask_is_valid;
   bool m_tls_is_valid;
-  size_t m_tls_size;
+  size_t m_tls_size = 0;
   bool m_gcs_is_valid;
   bool m_poe_is_valid;
 
-  struct user_pt_regs m_gpr_arm64; // 64-bit general purpose registers.
+  // 64-bit general purpose registers.
+  struct user_pt_regs m_gpr_arm64{};
 
-  RegisterInfoPOSIX_arm64::FPU
-      m_fpr; // floating-point registers including extended register sets.
+  // Floating-point registers including extended register sets.
+  RegisterInfoPOSIX_arm64::FPU m_fpr{};
 
   SVEState m_sve_state = SVEState::Unknown;
-  struct sve::user_sve_header m_sve_header;
+  struct sve::user_sve_header m_sve_header{};
   std::vector<uint8_t> m_sve_ptrace_payload;
 
   sve::user_za_header m_za_header;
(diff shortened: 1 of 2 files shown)


lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
```diff
@@ -91,63 +91,56 @@ class NativeRegisterContextLinux_arm64
...
-  bool m_refresh_hwdebug_info;
+  bool m_refresh_hwdebug_info = true;
 
   struct user_pac_mask {
-    uint64_t data_mask;
-    uint64_t insn_mask;
-  };
+    uint64_t data_mask = 0;
+    uint64_t insn_mask = 0;
+  } m_pac_mask;
```
DavidSpickett, line 116: I realise now that I could `{}` instead of `= 0` but I'll leave it as is and see what reviewers think.

DavidSpickett:
Like https://github.com/llvm/llvm-project/pull/197113, this is for AArch64 Linux but I welcome reviews from anyone because it's just c++ refactoring.

DavidSpickett:
Split out of https://github.com/llvm/llvm-project/pull/193174.

lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
```diff
@@ -91,63 +91,56 @@ class NativeRegisterContextLinux_arm64
   bool m_za_header_is_valid;
   bool m_pac_mask_is_valid;
   bool m_tls_is_valid;
-  size_t m_tls_size;
+  size_t m_tls_size = 0;
   bool m_gcs_is_valid;
   bool m_poe_is_valid;
 
-  struct user_pt_regs m_gpr_arm64; // 64-bit general purpose registers.
+  // 64-bit general purpose registers.
```
JDevlieghere: These should be `///` but I'm sure you did this for consistency. If there aren't that many, maybe it's worth fixing all of them?
```suggestion
  /// 64-bit general purpose registers.
```

lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
JDevlieghere: I personally prefer the ` = 0` 👍

JDevlieghere (approved):

lldb/source/Plugins/Process/Linux/NativeRegisterContextLinux_arm64.h
DavidSpickett: Fixed the ones changed by this PR.

I'll deal with the others later.

PR #174380, opened 2026-01-05
[InstCombine] Shrink added constant using LHS known zeros

ParkHanbum:
Previously, `SimplifyDemandedUseBits` for `add` instructions only
used known zeros from the RHS to simplify the LHS. It failed to
handle the symmetric case where the LHS has known zeros and the
result does not demand the low bits.

This patch implements this missing optimization, allowing the RHS
constant to be shrunk when the LHS low bits are known zero and unused.

Proof: https://alive2.llvm.org/ce/z/6v9iFY
Fixed: https://github.com/llvm/llvm-project/issues/135411

PR #197126, opened 2026-05-12
[SCEV] Canonicalise round-up idiom when some bits known

ostannard:
Since #174380, instcombine can clear some set bits in the added constant in expressions like this, when A has some known-clear low order bits.
  (A + 15) & ~15

This transformation is valid, but can make it harder for later passes to recognise this idiom for rounding up to a power of 2. This is causing the ARM MVE tail predication pass to fail on loops with a trip count which is a multiple of a small power of 2.

The fix is to reverse the transformation when building SCEV expressions, canonicalising to always use the largest valid value for the added constant.

Alive proofs:
https://alive2.llvm.org/ce/z/hhndoW
https://alive2.llvm.org/ce/z/_JYVat


llvm/lib/Analysis/ScalarEvolution.cpp
```diff
@@ -3704,6 +3704,25 @@ const SCEV *ScalarEvolution::getUDivExpr(SCEVUse LHS, SCEVUse RHS) {
         }
       }
 
+      // ((N - M) + (M * A)) / N --> ((N - 1) + (M * A)) / N
+      // This is an idiom for rounding A up to the next multiple of N, where A
+      // is aready known to be a multiple of M. In this case, instcomine can
```
nasherm: Nit: typo `instcomine` -> `instcombine`

llvm/test/Analysis/ScalarEvolution/rounding-demanded-bits.ll
nikic, line 1: Please use update_analyze_test_checks.

nikic (commented):
Can you please add an alive2 proof to the PR description?

llvm/lib/Analysis/ScalarEvolution.cpp
```diff
@@ -3704,6 +3704,25 @@ const SCEV *ScalarEvolution::getUDivExpr(SCEVUse LHS, SCEVUse RHS) {
...
+      // see that some low bits of the added constant are unused, so can clear
+      // them, but we want to canonicalise to set the low bits. This makes the
+      // pattern easier to match, without needing to check for known bits in
+      // A*M.
+      const APInt &N = RHSC->getAPInt();
+      const APInt *NMinusM, *M;
+      const SCEV *A;
+      if (match(LHS, m_scev_Add(m_scev_APInt(NMinusM),
+                                m_scev_Mul(m_scev_APInt(M), m_SCEV(A))))) {
+        if (N.urem(*M).isZero() && M->slt(N) && *NMinusM == N - *M) {
```
ParkHanbum: Do we need to restrict this to cases where `M` is a power of two?

llvm/lib/Analysis/ScalarEvolution.cpp
```diff
@@ -3704,6 +3704,26 @@ const SCEV *ScalarEvolution::getUDivExpr(SCEVUse LHS, SCEVUse RHS) {
...
+      // see that some low bits of the added constant are unused, so can clear
+      // them, but we want to canonicalise to set the low bits. This makes the
+      // pattern easier to match, without needing to check for known bits in
+      // A*M.
+      const APInt &N = RHSC->getAPInt();
+      const APInt *NMinusM, *M;
+      const SCEV *A;
+      if (match(LHS, m_scev_Add(m_scev_APInt(NMinusM),
+                                m_scev_Mul(m_scev_APInt(M), m_SCEV(A))))) {
+        if (N.isPowerOf2() && M->isPowerOf2() && M->slt(N) &&
```
nikic:
```suggestion
        if (N.isPowerOf2() && M->isPowerOf2() && M->ult(N) &&
```
https://alive2.llvm.org/ce/z/iBLxW6

nikic (approved):
LGTM

PR #197127, opened 2026-05-12
[lldb] Member initialise hardware breakpoint structures

DavidSpickett:
These are delcared in NativeRegisterContextDBReg so we should zero-init them there rather than have everyone memset them later.

ppc64le has its own equivalent that I've made the same change to.


mchoo7 (approved):
LGTM.

PR #191186, opened 2026-04-09
[LV] Handle FSub Partial Reductions

jacob-crawley:
Introduces a new RecurKind value 'FSub' in order to handle partial reductions of floating point values.

This is done by following the existing method for integer partial reductions, doing a positive accumulation followed by a final subtraction in the middle block.

PR #194594, opened 2026-04-28
[AArch64] Reflect cost of integer sub-reductions.

sdesmalen-arm:
The cost of sub-reductions is either the cost of *mlslb + *mlslt, or the cost of a dot operation with 2 negations:
```
       partial_reduce_umls acc, lhs, rhs
  <=> -partial_reduce_umla -acc, lhs, rhs
```
(codegen for this was added by #186809)

The cost-model was previously a bit of a hack, since sub-reductions were expanded and therefore expensive, although we made the expansion cost artifically cheaper so that it would still be a candidate for cdot instructions.

PR #197134, opened 2026-05-12
[LV] Handle FSub Partial Reductions

jacob-crawley:
Reland #191186 after fixing up test failures 

Introduces a new RecurKind value 'FSub' in order to handle partial reductions of floating point values.

This is done by following the existing method for integer partial reductions, doing a positive accumulation followed by a final subtraction in the middle block.


llvm/test/Transforms/LoopVectorize/AArch64/partial-reduce-sub.ll
```diff
@@ -412,5 +412,7 @@ for.exit:
 !8 = !{!"llvm.loop.mustprogress"}
 !9 = !{!"llvm.loop.vectorize.predicate.enable", i1 true}
 !10 = !{!"llvm.loop.vectorize.enable", i1 true}
+
 attributes #0 = { vscale_range(1,16) "target-features"="+sve" }
 attributes #1 = { "target-cpu"="apple-m1" }
+attributes #2 = { vscale_range(1,16) "target-features"="+sve2" }
```
MacDue, line 418: I guess requiring SVE2 was a consequence of https://github.com/llvm/llvm-project/pull/194594 removing the lower cost for sub expansions? So we now need `fmlalb/t` from SVE2?

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -6104,8 +6104,9 @@ getPartialReductionLinkCost(VPCostContext &CostCtx,
   if (RdxType->isFloatingPointTy())
     Flags = Link.ReductionBinOp->getFastMathFlags();
 
-  unsigned Opcode = Link.RK == RecurKind::Sub
-                        ? (unsigned)Instruction::Add
+  unsigned Opcode = Link.RK == RecurKind::Sub ? (unsigned)Instruction::Add
+                    : Link.RK == RecurKind::FSub
+                        ? (unsigned)Instruction::FAdd
```
MacDue: nit: It's more verbose, but I find a `switch` more readable than a nested ternary:
```c++
  auto GetLinkOpcode = [&Link]() -> unsigned {
    switch (Link.RK) {
      case RecurKind::Sub:
        return Instruction::Add;
      case RecurKind::FSub:
        return Instruction::FAdd;
      default:
        return Link.ReductionBinOp->getOpcode();
    }
  };

  return CostCtx.TTI.getPartialReductionCost(
      GetLinkOpcode(), ExtendedOp.ExtendA.SrcType, ExtendedOp.ExtendB.SrcType, RdxType,
      VF, ExtendedOp.ExtendA.Kind, ExtendedOp.ExtendB.Kind, BinOpc,
      CostCtx.CostKind, Flags);
```

MacDue:
Could you include the original description of https://github.com/llvm/llvm-project/pull/191186 in the summary?

llvm/test/Transforms/LoopVectorize/AArch64/partial-reduce-sub.ll
jacob-crawley: Thats right, the fsub case wouldn't generate a partial reduction otherwise.

sdesmalen-arm (approved):

MacDue (approved):
LGTM

PR #197135, opened 2026-05-12
[NVPTX] Promote non-standard integer global widths to .u{8,16,32,64}

### LINKED ISSUE #154337: NVPTX assembly emited by LLVM NVPTX assembly failed to compile due to `u2` type in global
Example LLVM IR:

```
source_filename = "test"
target datalayout = "e-i64:64-i128:128-v16:16-v32:32-n16:32:64"
target triple = "nvptx64-nvidia-cuda11.0.1-unknown"

@"test_global" = internal unnamed_addr addrspace(1) constant i2 1, align 1

attributes #0 = { noredzone nounwind "frame-pointer"="all" "target-cpu"="sm_80" "target-features"="+ptx70,+sm_80,-ptx32,-ptx40,-ptx41,-ptx42,-ptx43,-ptx50,-ptx60,-ptx61,-ptx62,-ptx63,-ptx64,-ptx65,-ptx71,-ptx72,-ptx73,-ptx74,-ptx75,-ptx76,-ptx77,-ptx78,-ptx80,-ptx81,-ptx82,-ptx83,-ptx84,-ptx85,-ptx86,-ptx87,-sm_100,-sm_100a,-sm_101,-sm_101a,-sm_120,-sm_120a,-sm_20,-sm_21,-sm_30,-sm_32,-sm_35,-sm_37,-sm_50,-sm_52,-sm_53,-sm_60,-sm_61,-sm_62,-sm_70,-sm_72,-sm_75,-sm_86,-sm_87,-sm_89,-sm_90,-sm_90a" }

!llvm.module.flags = !{}
```

Used to generate assembly with `llc -march=nvptx64 -mcpu=sm_80 bug.ll` emits the following NVPTX assembly:
```
//
// Generated by LLVM NVPTX Back-End
//

.version 7.0
.target sm_80
.address_size 64

.global .align 1 .u2 test_global = 1;
```

`ptxas` fails to assembly it:

```shell
$ ptxas --gpu-name=sm_80 bug.s -o bug.cubin
ptxas bug.s, line 9; fatal   : Parsing error near '.u2': syntax error
ptxas fatal   : Ptx assembly aborted due to errors
```

DingchenZhu:
## Summary

NVPTX globals whose IR type is a non-standard integer width (anything other
than `i1`, `i8`, `i16`, `i32`, `i64`) currently emit invalid PTX. For example
`@g = addrspace(1) constant i2 1` lowers to:

```
.global .align 1 .u2 g = 1;
```

which `ptxas` rejects with `Parsing error near '.u2': syntax error`. PTX only
defines `.u{8,16,32,64}` as fundamental integer storage types.

This PR fixes the asm printer to round such widths up to the next supported
PTX storage type, and to zero-extend the initializer literal to match. Both
sides have to move together — a type-only fix would turn the current
compile-time error into a silent miscompile for any value whose top bit was
set in the pre-promotion width.

Fixes #154337.

## Two commits

1. **`[NVPTX] Promote non-standard integer global widths to .u{8,16,32,64}`**
   Introduces a file-local helper `getPromotedPtxStorageBits` that encodes the
   "round up to next supported PTX storage type" rule in a single place, and
   uses it from `getPTXFundamentalTypeStr` for the type specifier. `i1` still
   maps to `.pred` (other callers of this helper depend on it); the two storage
   paths that would otherwise see `i1` short-circuit it to `.u8` before the
   call.

2. **`[NVPTX] Zero-extend literal when integer global is width-promoted`**
   Reuses the same helper from `printScalarConstant`. The old code printed the
   `ConstantInt` via `APInt`'s default signed formatter, so any value whose
   top bit was set in the original width became a negative literal in the
   promoted storage:

   | Input             | Before fix                    | After fix                       |
   |-------------------|-------------------------------|---------------------------------|
   | `i12 -1` (0xFFF)  | `.u16 g = -1;` (cubin 0xFFFF) | `.u16 g = 4095;` (cubin 0x0FFF) |
   | `i40 -1`          | `.u64 g = -1;` (cubin all 1s) | `.u64 g = 1099511627775;` (low 40 bits set) |

   The fix zero-extends the `APInt` to the promoted width and prints unsigned.
   For natively supported widths (`i8`/`i16`/`i32`/`i64`) the cubin byte
   pattern is unchanged — `ptxas` reinterprets `-1` and `2^N - 1` to the same
   target-typed unsigned representation.

The two commits can be squashed at merge time if reviewers prefer.

## `i1` behaviour change (disclosure)

`i1 true` globals previously emitted `.u8 g = -1` (cubin byte `0xFF`). With
this PR they emit `.u8 g = 1` (cubin byte `0x01`), matching LLVM IR's
canonical zero-extension of `i1 true`. Both encodings are truthy, so this has
no functional impact on standard boolean uses (conditional branches, `zext`
to a wider integer, etc.). The change only affects byte-level inspection of
the storage, such as `__builtin_memcpy(dst, &g_i1, 1)` followed by reading
the destination byte. No such pattern observed in the LLVM or MLIR test
suites.

A grep confirms no in-tree test locked in the prior `0xFF` byte pattern:

```bash
grep -rE '\.u(8|16|32|64) +\S+ *= *-[0-9]' llvm/test/ mlir/test/   # no matches
```

## Test coverage

New test: `llvm/test/CodeGen/NVPTX/int-globals-promotion.ll`. Each case pins
down an independent dimension so future regressions surface immediately:

| Case               | Pins                                                              |
|--------------------|-------------------------------------------------------------------|
| `i1 true`          | `i1`-zero-extension lock-in                                        |
| `i2 1`             | original #154337 repro, lower edge of `.u8` bucket                 |
| `i7 5`             | upper edge of `.u8` bucket                                         |
| `i12 -1`           | `.u16` bucket + value zero-extension                               |
| `i24 -1`           | `.u32` bucket + value zero-extension                               |
| `i34 8589934591`   | `.u64` bucket pinned from the value side, value > 2^32 (positive — sign-bit pinning is covered by `i40`) |
| `i40 -1`           | `.u64` bucket + value zero-extension                               |

The test runs `%ptxas-verify` when available so the emitted PTX is also
end-to-end checked against the real assembler.

## Risk & impact

Low. The two changed functions are NVPTX AsmPrinter helpers used only
within `llvm/lib/Target/NVPTX/`:

- `getPTXFundamentalTypeStr` — 3 callers, all in `NVPTXAsmPrinter.cpp`
  (global definition, extern global declaration, kernel scalar param).
- `printScalarConstant` — 1 caller, in `NVPTXAsmPrinter.cpp` (global
  initializer emission).

No IR semantics, no DAG/SDAG logic, no MIR. Compile-time cost is a handful
of additional integer comparisons per emitted global or kernel scalar param,
unmeasurable.

## Local verification

Built with `clang;lld;mlir` projects, `X86;NVPTX` targets, `RelWithDebInfo`
+ assertions on Ubuntu 24.04 with CUDA 12.6.20 (`ptxas` from CUDA 12.6
toolkit, PTX ISA 8.5):

- `ninja check-llvm-codegen-nvptx`: **500 / 500 pass** (baseline on
  unmodified `origin/main`: 499 / 499 — the new
  `int-globals-promotion.ll` is the 500th test, added in this PR).
- `ninja check-mlir`: **3583 pass / 0 unexpected fail / 591 unsupported /
  1 XFAIL** (pre-existing on main, verified against baseline before
  applying the patch).
- `git clang-format HEAD~2`: no reformatting.
- `%ptxas-verify` step inside the new test successfully assembles all 7
  emitted PTX globals.

## Out of scope

This PR targets the global-initializer emission path (`EmitVariable` →
`printScalarConstant`). Two related paths are intentionally left alone:

- **`bufferLEByte`** (used by array / struct globals) operates byte-by-byte
  via `getTypeStoreSize`, never goes through `printScalarConstant`. Sub-byte
  array elements (e.g. `[N x i2]`) are a separate codegen question not
  addressed here.
- **`emitFunctionParamList`** (kernel scalar params) emits only the type
  specifier — already routes through `getPTXFundamentalTypeStr` and so
  picks up the type-side fix automatically. There is no parameter-side
  initializer literal to print.

## AI tool use disclosure

Per the [LLVM AI Tool Use Policy](https://llvm.org/docs/AIToolPolicy.html):
this PR was developed in collaboration with Claude (Anthropic). The tool
assisted with root-cause investigation in `NVPTXAsmPrinter`, helped
enumerate edge cases (including the sign-bit promotion hazard that
motivated the second commit and the latent `i1` byte-pattern bug),
suggested the `getPromotedPtxStorageBits` helper-extraction shape, and
drafted prose for the commit messages and this description.

All design decisions, code, test-case selection, branch structuring, and
CI verification (`check-llvm-codegen-nvptx` + `check-mlir`) were reviewed
and approved by the contributor. The contributor will answer review
questions directly.


DingchenZhu:
@Artem-B @AlexMaclean — first-time contributor here, would appreciate a review when convenient. Pure NVPTX AsmPrinter fix for #154337, with an i1 behaviour disclosure noted in the description. The two commits are logically separable and can be squashed at merge time per LLVM workflow.

DingchenZhu:
Confirmed — I have read the LLVM AI Tool Use Policy, the LLVM Code-Review Policy and Practices, and the LLVM Developer Policy. AI tool usage on this PR is noted in the updated PR description under "AI tool use disclosure". Happy to answer review questions directly.

PR #197138, opened 2026-05-12
[LLVM] Add FastMathFlags operand to simplifySelectInst.

paulwalker-arm:
This removes the potentially bogus use of SimplifyQuery.CxtI, whose FastMathFlags are not necessarily relevant to the simplification.

Most uses don't care about fast-math-flags so I followed FoldBinaryIntrinsic's lead and made FMF optional for the folder interface.


nikic (approved):
LGTM

PR #197139, opened 2026-05-12
[clangd] Fix signature help for static call operator 196153 (C++23) 

Ippo47:
Parameter hints were not appearing for static operator because the object was always passed as the first argument, making static candidates invalid. Fix by checking each operator candidate individually if it is static, add it without the object argument.


clang/lib/Sema/SemaCodeComplete.cpp
```diff
@@ -185,7 +185,7 @@ class ResultBuilder {
 
   /// Overloaded C++ member functions found by SemaLookup.
   /// Used to determine when one overload is dominated by another.
-  llvm::DenseMap<std::pair<DeclContext *, /*Name*/uintptr_t>, ShadowMapEntry>
+  llvm::DenseMap<std::pair<DeclContext *, /*Name*/ uintptr_t>, ShadowMapEntry>
```
timon-ul: I see a lot of whitespace changes in this file, note that this is not desired as you did not change any logic in these lines. You most likely got a complaint by the checkers about formatting, but these checkers only check your changed lines for failed formatting and thus you should only fix it in those lines, not the whole file. These formatting rules have grown over time and there is no desire to just reformat the whole codebase, since this will create more effort in tracking who actually touched code logic last. So please undo these formattings in the unrelated sections.

timon-ul (changes requested):

clang/lib/Sema/SemaCodeComplete.cpp
Ippo47: fixed the the formatting please re review

clang/lib/Sema/SemaCodeComplete.cpp
```diff
@@ -6472,14 +6472,37 @@ SemaCodeCompletion::ProduceCallSignatureHelp(Expr *Fn, ArrayRef<Expr *> Args,
...
-                        ArgsWithoutDependentTypes.end());
-        SemaRef.AddFunctionCandidates(R.asUnresolvedSet(), ArgExprs,
-                                      CandidateSet,
-                                      /*ExplicitArgs=*/nullptr,
-                                      /*SuppressUserConversions=*/false,
-                                      /*PartialOverloading=*/true);
+        for (NamedDecl *D : R) {
+          NamedDecl *UD = D->getUnderlyingDecl();
+          if (auto *FD = dyn_cast<FunctionDecl>(UD)) {
+            if (FD->isStatic()) {
```
cor3ntin, line 6476:
```suggestion
          if (auto *FD = dyn_cast<FunctionDecl>(UD); FD && FD-> isStatic()) {
```

clang/lib/Sema/SemaCodeComplete.cpp
```diff
@@ -6472,14 +6472,37 @@ SemaCodeCompletion::ProduceCallSignatureHelp(Expr *Fn, ArrayRef<Expr *> Args,
...
+              SemaRef.AddOverloadCandidate(
+                  FD, DeclAccessPair::make(FD, FD->getAccess()),
+                  ArgsWithoutDependentTypes, CandidateSet,
+                  /*SuppressUserConversions=*/false,
+                  /*PartialOverloading=*/true);
+              continue;
+            }
+          } else if (auto *FTD = dyn_cast<FunctionTemplateDecl>(UD)) {
+            if (FTD->getTemplatedDecl()->isStatic()) {
```
cor3ntin, line 6485: Ditto

clang/lib/Sema/SemaCodeComplete.cpp
```diff
@@ -6470,14 +6470,37 @@ SemaCodeCompletion::ProduceCallSignatureHelp(Expr *Fn, ArrayRef<Expr *> Args,
...
+              SemaRef.AddTemplateOverloadCandidate(
+                  FTD, DeclAccessPair::make(FTD, FTD->getAccess()), nullptr,
+                  ArgsWithoutDependentTypes, CandidateSet,
+                  /*SuppressUserConversions=*/false,
+                  /*PartialOverloading=*/true);
+              continue;
+            }
+          }
+          SmallVector<Expr *, 12> ArgExprs(1, NakedFn);
```
timon-ul, line 6494: Any reason this code block has to be in the `for` loop? Seems to me like instead we could just use the old code block outside of the `for` loop since it is the same logic, just without having to construct a set and do a function call every loop.

clang/lib/Sema/SemaCodeComplete.cpp
```diff
@@ -6470,14 +6470,37 @@ SemaCodeCompletion::ProduceCallSignatureHelp(Expr *Fn, ArrayRef<Expr *> Args,
...
         R.suppressDiagnostics();
-        SmallVector<Expr *, 12> ArgExprs(1, NakedFn);
-        ArgExprs.append(ArgsWithoutDependentTypes.begin(),
-                        ArgsWithoutDependentTypes.end());
-        SemaRef.AddFunctionCandidates(R.asUnresolvedSet(), ArgExprs,
-                                      CandidateSet,
-                                      /*ExplicitArgs=*/nullptr,
-                                      /*SuppressUserConversions=*/false,
-                                      /*PartialOverloading=*/true);
+        for (NamedDecl *D : R) {
```
timon-ul, line 6473: No strong opinion on this, but why did you decide to iterate here over the declarations? Technically in the condensed `AddFunctionCandidates` call it already iterates over them and to me right now the only difference there is that it does not look at the underlying decl (and I assume that is what is currently failing?), so you could have also checked for it there. Just trying to understand if there is any difference.

clang-tools-extra/clangd/unittests/CodeCompleteTests.cpp
```diff
@@ -3455,6 +3455,93 @@ TEST(SignatureHelpTest, SkipExplicitObjectParameter) {
...
+    auto TU2 = TestTU::withCode(TemplateCode.code());
+    TU2.ExtraArgs = {"-std=c++23"};
+    MockFS FS2;
+    auto Inputs2 = TU2.inputs(FS2);
+    auto Preamble2 = TU2.preamble();
+    ASSERT_TRUE(Preamble2);
+    const auto Result =
+        signatureHelp(testPath(TU2.Filename), TemplateCode.point("c3"),
+                      *Preamble2, Inputs2, MarkupKind::PlainText);
+    EXPECT_EQ(1U, Result.signatures.size());
```
timon-ul, line 3516: You missed the `EXPECT_THAT` matcher here.

timon-ul (changes requested):

clang/lib/Sema/SemaCodeComplete.cpp
Ippo47: yes really good point, just updated the code

clang-tools-extra/clangd/unittests/CodeCompleteTests.cpp
Ippo47: fixed this

PR #194307, opened 2026-04-27
Update GitHub PR Greeter

banach-space:
Following these two discussions:
* https://discourse.llvm.org/t/rfc-mention-our-ai-policy-in-the-greeting-message-for-first-time-contributors/,
* https://discourse.llvm.org/t/concerns-about-influx-of-ai-generated-bug-fixes/,

add a reference to the LLVM AI policy in the GH greeter. 

In addition:
* Update the message to include links to other relevant policies as
  well, since these are often shared during PR review.
* Add FAQ section and move some of the original content there.
* Include a request for people to confirm that they have familiarised themselves with
  the policies.
* Add `Hello @{self.author} 👋` to make the greeting more personal.

PR #197140, opened 2026-05-12
Fix "author" handling in GitHub PR Greeter

banach-space:
This is a follow-up to #194307 and fixes the issue reported in:
  * https://github.com/llvm/llvm-project/pull/194307#issuecomment-4426270256

Use the same author-detection logic in `PRGreeter` as in
`PRBuildbotInformation`, so both components handle PR authors consistently.


DavidSpickett (approved):
LGTM

DavidSpickett:
If you're changing this stuff again in future, having the source code of PyGitHub to hand is a good idea. The type hints help a lot.

PR #197142, opened 2026-05-12
[GlobalISel][MatchTable] Fix RTTI of Imm/ImmPredicate classes

Pierre-vh:

PR #197143, opened 2026-05-12
[GlobalISel] Recursively Optimise MatchTable Matchers

Pierre-vh:
The core of this change is the additional call to `Matcher::optimize()` in the `optimizeRules` function,
which enables the match table optimization logic to recurse on the children of every GroupMatcher, forming
additional groups (which hoist more common predicates into a shared group).

To enable that, I had to update the `getFirstConditionAsRootType` implementation to support `GroupMatcher`.
I also included a small refactoring of the match table optimization pipeline that was identical between the
GlobalISel and GlobalISelCombiner emitters.

The results of this change are up to a 25% size reduction for GlobalISel match tables.
There is a tiny increase (a few bytes) in a combiner table because we now create new groups
(which need up to 3 additional opcodes because of the new `Try` and `Reject` required) to hoist one predicate for only 2 rules, which
result in a small net negative change (one or two more ops).

I used a small bash script to compare all relevant files, this is the before/after:
```
FILE                                          OLD      NEW    DIFF%    SAME?
----                                      -------  -------    -----    -----
AArch64GenGlobalISel.inc                   202434   192045    -5,1%       no
AArch64GenPostLegalizeGICombiner.inc         4428     4432     0,1%       no
AArch64GenPreLegalizeGICombiner.inc          8974     8977     0,0%       no
AMDGPUGenGlobalISel.inc                    789379   590082   -25,2%       no
AMDGPUGenPostLegalizeGICombiner.inc          9203     9207     0,0%       no
AMDGPUGenPreLegalizeGICombiner.inc           8950     8954     0,0%       no
AMDGPUGenRegBankGICombiner.inc               1981     1981     0,0%      yes
```


Pierre-vh:
* **#197143** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197143?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197143?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#197142** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197142?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

llvm/test/CodeGen/AArch64/GlobalISel/select-intrinsic-aarch64-sdiv.mir
```diff
@@ -1,6 +1,24 @@
 # NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
 # RUN: llc -mtriple=aarch64-- -run-pass=instruction-select -verify-machineinstrs %s -o - | FileCheck %s
 
+# XFAIL: *
```
davemgreen: Can you change the s32 to i32 in the test? That should be how the IR is generated nowadays.

llvm/test/CodeGen/AArch64/GlobalISel/preselect-process-phis.mir
```diff
@@ -1,5 +1,8 @@
 # NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
 # RUN: llc -verify-machineinstrs -mtriple aarch64--- --run-pass=instruction-select -global-isel-abort=1 %s -o - | FileCheck %s
+
+# XFAIL: *
```
davemgreen: Is this the same issue as below too?

Pierre-vh: Yes, thanks they were both the same problem.

llvm/test/CodeGen/AArch64/GlobalISel/select-intrinsic-aarch64-sdiv.mir
Pierre-vh: Thanks it fixes it

llvm/utils/TableGen/GlobalISelEmitter.cpp
```diff
@@ -2309,44 +2309,15 @@ Expected<RuleMatcher> GlobalISelEmitter::runOnPattern(const PatternToMatch &P) {
...
                                    bool Optimize, bool WithCoverage) {
-  std::vector<Matcher *> InputRules;
-  for (Matcher &Rule : Rules)
-    InputRules.push_back(&Rule);
-
-  if (!Optimize)
+  if (!Optimize) {
+    std::vector<Matcher *> InputRules;
+    for (Matcher &Rule : Rules)
+      InputRules.push_back(&Rule);
```
arsenm:
```suggestion
    std::vector<Matcher *> InputRules(Rules.begin(), Rules.end());
```

jayfoad: You'd probably need something like `make_pointer_range` to make that work?

arsenm (approved):

Pierre-vh:
@davemgreen @HolyMolyCowMan I am curious if this helps with compilation times on AArch64 GlobalIsel? Is this something you are tracking?

davemgreen:
@c-rhodes has been looking at -O0 recently.

There are some results for that in https://llvm-compile-time-tracker.com/compare.php?from=992df0aed5dda0a644c8939daa81b028f198651b&to=06d50ac2b7ba29646c0e871fbef928889abf4182&stat=instructions:u, so yeah it looks like it helps.

PR #156864, opened 2025-09-04
[VPlan] Don't use the legacy cost model for loop conditions

john-brawn-arm:
The current behaviour of using the legacy cost model for instructions that compute the loop exit condition gets the wrong result when the loop has been transformed to use a different exit condition, e.g. when have tail-folded predicated vectorization the exit condition is based on the predicate vector.

Fix this by adding cost computation for BranchOnCount and removing the restriction on computing the cost for scalar ICmp/FCmp, and removing the use of the legacy cost model for loop exit conditions.

This causes quite a lot of changes to expected output in tests. Some of these are just changes to the -debug output, others are choosing a different VF due to previously over or under-estimating the cost, and in others the minimum trip count has changed as we now compute the cost for compares in the middle block.

PR #197146, opened 2026-05-12
[VPlan] Compute the cost for scalar cmp outside the vector region

john-brawn-arm:
Currently we don't compute the cost of any scalar compares. Change this to only avoid computing the cost if it's inside the vector region, as compares that are used in the loop exit condition are handled by the legacy cost model and this is the simplest way to avoid double-counting those instructions.

This mainly affects the compare in the middle block, and accounting for the cost of that can change the requred minimum trip count.


llvm/test/Transforms/LoopVectorize/X86/CostModel/vpinstruction-cost.ll
```diff
@@ -43,7 +43,7 @@ define void @wide_or_replaced_with_add_vpinstruction(ptr %src, ptr noalias %dst)
 ; CHECK:  Cost of 0 for VF 2: IR %l = load i64, ptr %g.src, align 8
 ; CHECK:  Cost of 0 for VF 2: IR %iv.4 = add nuw nsw i64 %iv, 4
 ; CHECK:  Cost of 0 for VF 2: IR %c = icmp ule i64 %l, 128
-; CHECK:  Cost of 0 for VF 2: EMIT vp<%cmp.n> = icmp eq ir<32>, vp<[[VP2]]>
+; CHECK:  Cost of 1 for VF 2: EMIT vp<%cmp.n> = icmp eq ir<32>, vp<[[VP2]]>
```
david-arm, line 46: It looks like this is being used as the exit condition?

john-brawn-arm: This compare is in the middle block. The full vplan here is
```
VPlan 'Initial VPlan for VF={2,4},UF>=1' {
Live-in vp<%0> = VF
Live-in vp<%1> = VF * UF
Live-in vp<%2> = vector-trip-count
Live-in ir<32> = original trip-count

ir-bb<entry>:
Successor(s): scalar.ph, vector.ph

vector.ph:
Successor(s): vector loop

<x1> vector loop: {
vp<%3> = CANONICAL-IV

  vector.body:
    ir<%iv> = WIDEN-INDUCTION nuw nsw ir<0>, ir<1>, vp<%0>
    vp<%4> = SCALAR-STEPS vp<%3>, ir<1>, vp<%0>
    CLONE ir<%g.src> = getelementptr inbounds ir<%src>, vp<%4>
    vp<%5> = vector-pointer inbounds ir<%g.src>
    WIDEN ir<%l> = load vp<%5>
    WIDEN ir<%iv.4> = add ir<%iv>, ir<4>
    WIDEN ir<%c> = icmp ule ir<%l>, ir<128>
    EMIT ir<%or> = add ir<%iv.4>, ir<1>
    CLONE ir<%g.dst> = getelementptr ir<%dst>, ir<%or>
    vp<%6> = vector-pointer ir<%g.dst>
    WIDEN store vp<%6>, ir<%iv.4>, ir<%c>
    EMIT vp<%index.next> = add nuw vp<%3>, vp<%1>
    EMIT branch-on-count vp<%index.next>, vp<%2>
  No successors
}
Successor(s): middle.block

middle.block:
  EMIT vp<%cmp.n> = icmp eq ir<32>, vp<%2>
  EMIT branch-on-cond vp<%cmp.n>
Successor(s): ir-bb<exit>, scalar.ph

ir-bb<exit>:
No successors

scalar.ph:
  EMIT-SCALAR vp<%bc.resume.val> = phi [ vp<%2>, middle.block ], [ ir<0>, ir-bb<entry> ]
Successor(s): ir-bb<loop.header>

ir-bb<loop.header>:
  IR   %iv = phi i64 [ 0, %entry ], [ %iv.next, %loop.latch ] (extra operand: vp<%bc.resume.val> from scalar.ph)
  IR   %g.src = getelementptr inbounds i64, ptr %src, i64 %iv
  IR   %l = load i64, ptr %g.src, align 8
  IR   %iv.4 = add nuw nsw i64 %iv, 4
  IR   %c = icmp ule i64 %l, 128
No successors
}
```

john-brawn-arm:
I've changed this to only count compares that are outside of the vector loop. This because in the test added by #197504 we have a compare that's used in the loop condition but it goes through an "or" before the branch-on-cond:
```
<x1> vector loop: {
vp<%3> = CANONICAL-IV

  vector.body:
    ir<%iv> = WIDEN-INDUCTION nuw nsw ir<0>, ir<1>, vp<%0>
    WIDEN-REDUCTION-PHI ir<%max> = phi ir<-1.000000e+07>, ir<%max.next>
    WIDEN-CAST ir<%iv.f> = sitofp ir<%iv> to float
    WIDEN ir<%v> = fadd ir<%base>, ir<%iv.f>
    WIDEN-INTRINSIC ir<%max.next> = call llvm.maxnum(ir<%max>, ir<%v>)
    EMIT vp<%index.next> = add nuw vp<%3>, vp<%1>
    EMIT vp<%4> = fcmp uno ir<%v>, ir<%v>
    EMIT vp<%5> = any-of vp<%4>
    EMIT vp<%6> = icmp eq vp<%index.next>, vp<%2>
    EMIT vp<%7> = or vp<%5>, vp<%6>
    EMIT branch-on-cond vp<%7>
  No successors
}
Successor(s): middle.block
```
I don't know of any situation where we have a scalar cmp inside the vector loop that isn't used for the loop exit condition, so I think this is the simplest way to handle this. And this check will be going away anyway in #156864

llvm/test/Transforms/LoopVectorize/RISCV/tail-folding-div.ll
```diff
@@ -404,7 +408,8 @@ define void @test_sdiv_divisor_invariant_safeimm(ptr noalias %a, ptr noalias %c)
 ; NO-VP-NEXT:  [[LOOP_PREHEADER:.*]]:
 ; NO-VP-NEXT:    [[TMP0:%.*]] = call i64 @llvm.vscale.i64()
 ; NO-VP-NEXT:    [[TMP1:%.*]] = shl nuw i64 [[TMP0]], 1
-; NO-VP-NEXT:    [[MIN_ITERS_CHECK:%.*]] = icmp ult i64 1024, [[TMP1]]
+; NO-VP-NEXT:    [[UMAX:%.*]] = call i64 @llvm.umax.i64(i64 [[TMP1]], i64 4)
+; NO-VP-NEXT:    [[MIN_ITERS_CHECK:%.*]] = icmp ult i64 1024, [[UMAX]]
```
lukel97, line 412: Not necessarily relevant to this PR but are these umaxes redundant? If VF (TMP1) was less than 4 and we're now always assuming it's least 4, it would still fail the MIN_ITERS_CHECK

lukel97: I don't think this is blocking btw, it looks like these only show up on the non-tail folded loops on RISC-V so it shouldn't be too important

david-arm (approved):
LGTM!

fhahn (commented):
I think this needs a rebase + update to newly added llvm/test/Transforms/LoopVectorize/AArch64/cmp_cost.ll

PR #197154, opened 2026-05-12
Updating tests for usage of CLANG_RESOURCE_DIR (clang/test/Driver/driverkit-path.c) (1/4)

dyung:
We would like to enable the clang CMake option CLANG_RESOURCE_DIR on our build bots, but found that a few tests that need updating since they make assumptions about compiler paths that are modified when using CLANG_RESOURCE_DIR. This is change 1 of 4 to update these tests individually.

Test: clang/test/Driver/driverkit-path.c

The issue is with the regex on line 36, it assumes the path to the include directories does not change, which when using CLANG_RESOURCE_DIR, it can.

For example, the RUN line on line 27 is `%clang %s -target x86_64-apple-driverkit19.0 -isysroot %S/Inputs/DriverKit19.0.sdk -x c++ -###`, which when not using CLANG_RESOURCE_DIR, it generates the following output (edited for brevity and clarity):
```
"-isysroot" "/Users/dyung/src/git/llvm-project/clang/test/Driver/Inputs/DriverKit19.0.sdk"
"-internal-isystem" "/Users/dyung/src/git/llvm-project/clang/test/Driver/Inputs/DriverKit19.0.sdk/System/DriverKit/usr/local/include"
"-internal-isystem" "/Users/dyung/src/git/llvm-project/21776-baseline/lib/clang/23/include"
"-internal-externc-isystem" "/Users/dyung/src/git/llvm-project/clang/test/Driver/Inputs/DriverKit19.0.sdk/System/DriverKit/usr/include"
```
Note in the second instance of `-internal-isystem`, the path to the include directory includes `lib/clang/23`.

Now consider the case we would like to use where the compiler is built with `-DCLANG_RESOURCE_DIR=../lib/clang`. The output of the previous command is now the following output (edited for brevity and clarity):
```
"-isysroot" "/Users/dyung/src/git/llvm-project/clang/test/Driver/Inputs/DriverKit19.0.sdk"
"-internal-isystem" "/Users/dyung/src/git/llvm-project/clang/test/Driver/Inputs/DriverKit19.0.sdk/System/DriverKit/usr/local/include"
"-internal-isystem" "/Users/dyung/src/git/llvm-project/21776-relative/bin/../lib/clang/include"
"-internal-externc-isystem" "/Users/dyung/src/git/llvm-project/clang/test/Driver/Inputs/DriverKit19.0.sdk/System/DriverKit/usr/include"
```
Note here in the second instance of `-internal-isystem`, the path to the include directory is now `bin/../lib/clang`. In particular, the clang version number is no longer present in the path.

My proposed fix here is to update the regex so that it works both with and without the clang version number in the path to the include directory. While it won't necessarily work for any arbitrary value of CLANG_RESOURCE_DIR that someone might try to use, this is the value we would like to use and so we wish to update the test so that we can enable it on our build bots.

### DIFF
--- a/clang/test/Driver/driverkit-path.c
+++ b/clang/test/Driver/driverkit-path.c
@@ -1,10 +1,6 @@
 // UNSUPPORTED: system-windows
 //   Windows is unsupported because we use the Unix path separator `\`.
 
-// Determine the resource directory used by the compiler so we do not have to guess at it
-// RUN: rm -rf %t && mkdir %t
-// RUN: %clang -print-resource-dir | tr -d '\n' > %t/resource-dir
-
 // RUN: %clang %s -target x86_64-apple-driverkit19.0 -mlinker-version=0 \
 // RUN:   -isysroot %S/Inputs/DriverKit19.0.sdk -### 2>&1               \
 // RUN: | FileCheck %s --check-prefix=LD64-OLD
@@ -29,11 +25,11 @@ int main() { return 0; }
 
 
 // RUN: %clang %s -target x86_64-apple-driverkit19.0 -isysroot %S/Inputs/DriverKit19.0.sdk -x c++ -### 2>&1 \
-// RUN: | FileCheck %s -DSDKROOT=%S/Inputs/DriverKit19.0.sdk -DRESOURCE_DIR="%{readfile:%t/resource-dir}" --check-prefix=INC
+// RUN: | FileCheck %s -DSDKROOT=%S/Inputs/DriverKit19.0.sdk -DRESOURCE_DIR=%clang-resource-dir --check-prefix=INC
 // RUN: %clang %s -target x86_64-apple-driverkit21.0.1 -isysroot %S/Inputs/DriverKit21.0.1.sdk -x c++ -### 2>&1 \
-// RUN: | FileCheck %s -DSDKROOT=%S/Inputs/DriverKit21.0.1.sdk -DRESOURCE_DIR="%{readfile:%t/resource-dir}" --check-prefix=INC
+// RUN: | FileCheck %s -DSDKROOT=%S/Inputs/DriverKit21.0.1.sdk -DRESOURCE_DIR=%clang-resource-dir --check-prefix=INC
 // RUN: %clang %s -target x86_64-apple-driverkit23.0 -isysroot %S/Inputs/DriverKit23.0.sdk -x c++ -### 2>&1 \
-// RUN: | FileCheck %s -DSDKROOT=%S/Inputs/DriverKit23.0.sdk -DRESOURCE_DIR="%{readfile:%t/resource-dir}" --check-prefix=INC
+// RUN: | FileCheck %s -DSDKROOT=%S/Inputs/DriverKit23.0.sdk -DRESOURCE_DIR=%clang-resource-dir --check-prefix=INC
 //
 // INC: "-isysroot" "[[SDKROOT]]"
 // INC: "-internal-isystem" "[[SDKROOT]]/System/DriverKit/usr/local/include"
--- a/clang/test/Driver/driverkit-path.c
+++ b/clang/test/Driver/driverkit-path.c
@@ -33,7 +33,7 @@ int main() { return 0; }
 //
 // INC: "-isysroot" "[[SDKROOT]]"
 // INC: "-internal-isystem" "[[SDKROOT]]/System/DriverKit/usr/local/include"
-// INC: "-internal-isystem" "{{.+}}/lib{{(64)?}}/clang/{{[^/ ]+}}/include"
+// INC: "-internal-isystem" "{{.+}}/lib{{(64)?}}/clang{{/?[^"]*}}/include"
 // INC: "-internal-externc-isystem" "[[SDKROOT]]/System/DriverKit/usr/include"
 // INC: "-internal-iframework" "[[SDKROOT]]/System/DriverKit/System/Library/Frameworks"
 // INC: "-internal-iframework" "[[SDKROOT]]/System/DriverKit/System/Library/SubFrameworks"
--- a/clang/test/Driver/driverkit-path.c
+++ b/clang/test/Driver/driverkit-path.c
@@ -1,6 +1,10 @@
 // UNSUPPORTED: system-windows
 //   Windows is unsupported because we use the Unix path separator `\`.
 
+// Determine the resource directory used by the compiler so we do not have to guess at it
+// RUN: rm -rf %t && mkdir %t
+// RUN: %clang -print-resource-dir | tr -d '\n' > %t/resource-dir
+
 // RUN: %clang %s -target x86_64-apple-driverkit19.0 -mlinker-version=0 \
 // RUN:   -isysroot %S/Inputs/DriverKit19.0.sdk -### 2>&1               \
 // RUN: | FileCheck %s --check-prefix=LD64-OLD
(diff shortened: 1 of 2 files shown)


ian-twilightcoder:
Should we use `clang -print-resource-dir` so that we can be specific in the test instead of try to guess with a regex?

dyung:
> Should we use `clang -print-resource-dir` so that we can be specific in the test instead of try to guess with a regex?

I'm not familiar with that option, so let me take a look and see if that might be a suitable alternative.

dyung:
@ian-twilightcoder, thanks for the suggestion! I've updated the test to use `-print-resource-dir` to determine the correct directory and the test now searches for that instead of trying to use a regex to guess at it.

clang/test/Driver/driverkit-path.c
```diff
@@ -1,6 +1,10 @@
 // UNSUPPORTED: system-windows
 //   Windows is unsupported because we use the Unix path separator `\`.
 
+// Determine the resource directory used by the compiler so we do not have to guess at it
+// RUN: rm -rf %t && mkdir %t
+// RUN: %clang -print-resource-dir | tr -d '\n' > %t/resource-dir
```
ian-twilightcoder: Does it work to do this?
```
// DEFINE: %{resourcedir} = $(%clang -print-resource-dir)
...
... -DRESOURCE_DIR=%resourcedir
```
Looking at "Test-specific substitutions" at https://llvm.org/docs/TestingGuide.html

ian-twilightcoder: Or maybe we should just add clang_resource_dir to lit.cfg.py, that might be a lot easier.

ian-twilightcoder: Does it work to add `config.clang_resource_dir = @CLANG_RESOURCE_DIR@` to lit.site.cfg.py?

dyung: > Does it work to add `config.clang_resource_dir = @CLANG_RESOURCE_DIR@` to lit.site.cfg.py?

I tried this, but for some weird reason, it prepends "../../../" making the result seemingly really inaccurate.

dyung: > Or maybe we should just add clang_resource_dir to lit.cfg.py, that might be a lot easier.

This I found to be much easier and I can just query clang directly for the value so this is what I decided to use.

dyung:
@ian-twilightcoder when you get a chance, could you review the latest changes I've made?

clang/test/lit.cfg.py
```diff
@@ -412,6 +412,20 @@ def calculate_arch_features(arch_string):
 if config.enable_threads:
     config.available_features.add("thread_support")
 
+# Add clang resource directory as a substitution
+if config.clang:
+    try:
```
ian-twilightcoder: This seems like a really painful way to do this. Why not this instead?
```
clang_resource_dir = subprocess.run([config.clang, "-print-resource-dir"], stdout=subprocess.PIPE, text=True).stdout.rstrip()
```

dyung: I was modeling it after the function `run_clang_repl` which does something similar, but I suppose we can simplify to what you suggest. Let me give it a try and if it looks good, I'll update the change.

dyung: It worked well and passed my tests, so I've simplified how it is run to what you suggest. Thanks for the suggestion!

ian-twilightcoder (approved):
Looks great, thank you!

clang/test/lit.cfg.py
```diff
@@ -412,6 +412,11 @@ def calculate_arch_features(arch_string):
 if config.enable_threads:
     config.available_features.add("thread_support")
 
+# Add clang resource directory as a substitution
```
ian-twilightcoder, line 415: I would add a note that the obvious `config.clang_resource_dir = @CLANG_RESOURCE_DIR@` doesn't work.

ian-twilightcoder: Actually, does that work if you use `os.path.realpath` on it?

carlocab: The difficulty might be is that, IIRC, `CLANG_RESOURCE_DIR` can be a path relative to the `clang` binary. In that case checking `clang -print-resource-dir` might be simplest, or we can reconstruct it from where we expect `clang` to be.

ian-twilightcoder: Does it work if you do `config.clang_resource_dir = path(r"@CLANG_RESOURCE_DIR@")`?

ian-twilightcoder: Or maybe `config.clang_resource_dir = lit_config.substitute(path(r"@CLANG_RESOURCE_DIR@"))`?

dyung: > Does it work if you do `config.clang_resource_dir = path(r"@CLANG_RESOURCE_DIR@")`?

This is what I originally tried, but it resolved to a non-existent path, not sure why exactly though. I'll give your suggestions another try though just to reverify.

dyung: So I'm not sure why your suggestions didn't work as-is (I might be missing something), but I went back to my original implementations to try what you suggested. This is the code that was added to `clang/test/lit.site.cfg.py.in`:
```
config.substitutions.append(("%alfa-clang-resource-dir", "@CLANG_RESOURCE_DIR@"))
config.substitutions.append(("%bravo-clang-resource-dir", path(r"@CLANG_RESOURCE_DIR@")))
config.charlie_clang_resource_dir = path(r"@CLANG_RESOURCE_DIR@")
config.delta_clang_resource_dir = lit_config.substitute(path(r"@CLANG_RESOURCE_DIR@"))
```
Of the changes here, alfa and bravo worked, charlie and delta did not.

When building with CLANG_RESOURCE_DIR=../lib/clang, It generated the following values in the lit.cfg.py:
```
config.substitutions.append(("%clang-resource-dir-2", "../../../../lib/clang"))
config.substitutions.append(("%clang-resource-dir-3", path(r"../../../../lib/clang")))
config.clang_resource_dir_4 = path(r"../../../../lib/clang")
config.clang_reousrce_dir_5 = lit_config.substitute(path(r"../../../../lib/clang"))
```
I added the following run lines to a test to examine the output:
```
// RUN: echo clang-resource-dir = %clang-resource-dir
// RUN: echo alfa-clang-resource-dir = %alfa-clang-resource-dir
// RUN: echo bravo-clang-resource-dir = %bravo-clang-resource-dir
// RUN: echo charlie_clang_resource_dir = %charlie_clang_resource_dir
// RUN: echo delta_clang_resource_dir = %delta_clang_resource_dir
// RUN: false
```
Which output the following (the first line just prints the output of `clang -print-resource-dir`:
```
# RUN: at line 4
echo clang-resource-dir = /Users/dyung/src/git/llvm-project/21776-relative/bin/../lib/clang
# executed command: echo clang-resource-dir = /Users/dyung/src/git/llvm-project/21776-relative/bin/../lib/clang
# .---command stdout------------
# | clang-resource-dir = /Users/dyung/src/git/llvm-project/21776-relative/bin/../lib/clang
# RUN: at line 5
echo alfa-clang-resource-dir = ../../../../lib/clang
# executed command: echo alfa-clang-resource-dir = ../../../../lib/clang
# .---command stdout------------
# | alfa-clang-resource-dir = ../../../../lib/clang
# RUN: at line 6
echo bravo-clang-resource-dir = /Users/dyung/src/git/llvm-project/lib/clang
# executed command: echo bravo-clang-resource-dir = /Users/dyung/src/git/llvm-project/lib/clang
# .---command stdout------------
# | bravo-clang-resource-dir = /Users/dyung/src/git/llvm-project/lib/clang
# RUN: at line 7
echo charlie_clang_resource_dir = %charlie_clang_resource_dir
# executed command: echo charlie_clang_resource_dir = %charlie_clang_resource_dir
# .---command stdout------------
# | charlie_clang_resource_dir = %charlie_clang_resource_dir
# RUN: at line 8
echo delta_clang_resource_dir = %delta_clang_resource_dir
# executed command: echo delta_clang_resource_dir = %delta_clang_resource_dir
# .---command stdout------------
# | delta_clang_resource_dir = %delta_clang_resource_dir
```
Note that for alfa and bravo, it ultimately resolves to a path `/Users/dyung/src/git/llvm-project/lib/clang` which doesn't actually exist.

The more normal case where CLANG_RESOURCE_DIR is not defined ends up being worse because the paths end up being empty strings and so we don't actually get any path out of this which would be bad for trying to match it with a regex in the test:
```
# RUN: at line 4
echo clang-resource-dir = /Users/dyung/src/git/llvm-project/21776-baseline/lib/clang/23
# executed command: echo clang-resource-dir = /Users/dyung/src/git/llvm-project/21776-baseline/lib/clang/23
# .---command stdout------------
# | clang-resource-dir = /Users/dyung/src/git/llvm-project/21776-baseline/lib/clang/23
# RUN: at line 5
echo alfa-clang-resource-dir =
# executed command: echo alfa-clang-resource-dir =
# .---command stdout------------
# | alfa-clang-resource-dir =
# RUN: at line 6
echo bravo-clang-resource-dir =
# executed command: echo bravo-clang-resource-dir =
# .---command stdout------------
# | bravo-clang-resource-dir =
# RUN: at line 7
echo charlie_clang_resource_dir = %charlie_clang_resource_dir
# executed command: echo charlie_clang_resource_dir = %charlie_clang_resource_dir
# .---command stdout------------
# | charlie_clang_resource_dir = %charlie_clang_resource_dir
# RUN: at line 8
echo delta_clang_resource_dir = %delta_clang_resource_dir
# executed command: echo delta_clang_resource_dir = %delta_clang_resource_dir
# .---command stdout------------
# | delta_clang_resource_dir = %delta_clang_resource_dir
```
Because of this, I think the approach I currently have of using the output of `clang -print-resource-dir` is probably the best approach. Any to just committing what I currently have in this PR?

ian-twilightcoder: Yeah, looking at `Driver::GetResourcesPath` that all makes sense now, I think what you have is best. I hadn't read it close enough I had thought/hoped that CLANG_RESOURCE_DIR was like CLANG_BINARY_DIR.

ian-twilightcoder (approved):

clang/test/lit.cfg.py
dyung: > Yeah, looking at `Driver::GetResourcesPath` that all makes sense now, I think what you have is best. I hadn't read it close enough I had thought/hoped that CLANG_RESOURCE_DIR was like CLANG_BINARY_DIR.

Yeah, that would have been a lot easier, but sadly it was not the case. Thanks for reviewing the change!

PR #197157, opened 2026-05-12
[AA] Consider read-only provenance capture for synchronization effects

nikic:
If only read-only provenance is captured, this means that another thread may only read the object, not write to it. As such, we can also model synchronizing operations as only reading the location (and thus allow reordering of reads, but not writes, across the synchronization).


antoniofrighetto (approved):
LGTM. For a moment, I thought we would now be saying Ref for `%a` if we were also to return it, but that's indeed not the case...
```llvm
define ptr @malloc_escape_after_readonly_and_return(ptr %x, i64 %sz) {
  %a = call ptr @malloc(i64 %sz)
  ; atomic ops
  call void @escape(ptr captures(read_provenance) %a)
  ret ptr %a
}
```

PR #196616, opened 2026-05-08
[mlir][dataflow] IntRange: Replace yield-based widening with per-state lattice budget

Hardcode84:
IntegerRangeAnalysis can hang on `scf.while` loops with dynamic bounds: a
loop-carried range ratchets [0,0]->[0,1]->[0,2]->... by one per worklist
visit, requiring up to 2^31 iterations on i32. The new
`int-range-analysis-convergence.mlir` test reproduces this.

The ratchet lives at framework merge sites (region successors, callable
args) where the solver joins lattices via virtual
`Lattice::join(const AbstractSparseLattice &)`. The pre-existing
`isYieldedResult`/`isYieldedValue` heuristic in
`IntegerRangeAnalysis::visitOperation` doesn't help: it runs in the
transfer-function callback for inferrable-op results used by a terminator,
not on the merge path. It is also harmful where it fires - slams to
maxRange on the *second* visit (after, say, [1,1]->[1,2]), so naturally
bounded accumulators (e.g. `arith.minsi`-clamped iter args) widen to
[INT_MIN, INT_MAX].

Replace it with a per-state widening budget on `IntegerValueRangeLattice`:
the lattice counts merge-site joins and forces the range to its max once
the count hits `kIntegerRangeWideningBudget` (128). Only the virtual
overload is overridden, so transfer-function joins via the non-virtual
`join(const ValueT &)` are unaffected. The new `int-range-loop-iter-args.mlir`
test pins the tighter bounds; the convergence test verifies termination.

PR #197159, opened 2026-05-12
[mlir][dataflow] IntRange: Bump int-range-analysis widening budget.

Hardcode84:
Downstream testing (Triton) found budget is too low on realistic workloads. I'm going to switch to pass option probably, but for now, just bump the budget to unblock things.

See https://github.com/llvm/llvm-project/pull/196616#issuecomment-4430041952


boomanaiden154 (commented):
We were able to resolve the triton regressions by changing the line in https://github.com/triton-lang/triton/blob/36394d4c65a44eec9371012e2d3133107971eae3/third_party/amd/lib/Analysis/RangeAnalysis.cpp#L775

to `changed = argLat->join(operLat.getValue());`.

Hardcode84:
> We were able to resolve the triton regressions by changing the line in https://github.com/triton-lang/triton/blob/36394d4c65a44eec9371012e2d3133107971eae3/third_party/amd/lib/Analysis/RangeAnalysis.cpp#L775
> 
> to `changed = argLat->join(operLat.getValue());`.

Cool, thanks. I will keep this PR open and probably switch to pass option later anyway.

PR #197164, opened 2026-05-12
[InstCombine] Fix one operator precedence

el-ev:
It is `!(A) || B` while it is intended to be `!(A || B)`.


nikic (commented):
The change looks right. Is this testable?

el-ev:
Seems not... The second branch is unreachable as `icmp ne X, 0` + `select` is uusally canonicalized to something else

nikic (approved):
LGTM

PR #197166, opened 2026-05-12
[X86] Manage atomic store of fp -> int promotion in DAG

jofrn:
When lowering `atomic store <1 x T>` vector types with floats (i.e. during scalarization in the selection DAG), selection can fail since
this pattern is unsupported. To support this, floats can be casted to
an integer type of the same size.

Store-side counterpart to #148895. Stacked on top of #197165; and below of #197618.


llvm/test/CodeGen/X86/atomic-load-store.ll
```diff
@@ -222,6 +222,140 @@ define void @store_atomic_vec1_ptr(ptr %x, <1 x ptr> %v) nounwind {
   ret void
 }
 
+define void @store_atomic_vec1_bfloat(ptr %x, <1 x bfloat> %v) {
```
arsenm: The tests look more like they are for vector scalarization. This also isn't removing any of the IR lowering TargetLowering hooks, so I doubt this patch is doing anything as it is

jofrn: They are for vector scalarization. It casts the types instead of removing hooks. Please check out how it was done for the loads in https://github.com/llvm/llvm-project/pull/148895.

RKSimon:
```suggestion
define void @store_atomic_vec1_bfloat(ptr %x, <1 x bfloat> %v) nounwind {
```
same for others to remove cfi noise

arsenm: This doesn't make sense, you only changed the rules for the scalar types. This also will not be reached since AtomicExpand will have replaced the float atomics unless you do something to shouldExpandAtomicStoreInIR / shouldCastAtomicStoreInIR

jofrn: That makes sense. The hooks weren't modified, so we don't need these.

arsenm: I think the hooks should be modified (and by modified, I mean fully deleted).

I tried doing this a long time ago, I think load and store shouldn't be too much trouble to move the casting into codegen. The tricky one is atomicrmw, since when expanding into cmpxchg loops, putting the cast inside the loop is bad for some targets including x86

jofrn: The hook in x86 isel lowering (i.e. the base class) only handles scalar floating point types.
`virtual AtomicExpansionKind shouldCastAtomicStoreInIR(StoreInst *SI) const {
    if (SI->getValueOperand()->getType()->isFloatingPointTy())
      return AtomicExpansionKind::CastToInteger;
    return AtomicExpansionKind::None;
}`
Therefore, we do need these. I understand what you are saying ; it will be much more difficult to modify the base class though. This statement was accurate: https://github.com/llvm/llvm-project/pull/197166#discussion_r3226321068

jofrn: Should it be there for those that don't require it? i.e. those that do not have CFI noise currently.

RKSimon: I tend to blanket add it to most tests - but if that will cause large diffs, just add it to cases that need it

arsenm: OK, this is a convoluted set of circumstances. shouldCastAtomicStoreInIR is buggy, and missed the 1 x vector edge case (it ought to be checking isFPOrFPVectorTy). This patch exclusively fixes the handling of the 1x vector, which is not obvious from the title and description. I'd prefer to fix this by first fixing the buggy cast logic to handle the vector case, and then proceed with the migration to the DAG based promotion

RKSimon (approved):
LGTM

llvm/test/CodeGen/X86/atomic-load-store.ll
jofrn: Ok, sure. We can try specializing in it x86.

arsenm: No, not in x86. In the base implementation (but this is still just pre-work to deleting it)

jofrn: Right, work prior to the final state is redundant. will leave this as it is then until we get to the final state. thank you.

PR #187589, opened 2026-03-19
[VPlan] Expand DerivedIV into executable recipes

artagnon:
This allows us to strip DerivedIVRecipe::execute, and remove the dependency on emitTransformedIndex. It allows us to benefit from existing simplifications in VPlan.

PR #197169, opened 2026-05-12
[LV] Generate compact IR in VPScalarIVStepsRecipe::execute

david-arm:
While trying to remove the instcombine pass from the RUN
line of

Transforms/LoopVectorize/AArch64/sve-gather-scatter.ll

I realised that we frequently generate poor code for the
VPScalarIVStepsRecipe recipe. I know that a later
instcombine pass cleans up the code, but I didn't see any
harm in tidying up the output for this one recipe. I've
used the InstSimplifyFolder here and the impact on the IR
is minimal. An alternative approach would be to break up
the recipe into smaller ones that can be simplified with
simplifyRecipes. However, I'm not sure if this is the
preferred route.

I think it's also easier to see the true cost of the recipe
with this PR. It's possible to use the folder more
generally for all recipes. I tried that but it leads to
some very significant changes in IR, including rewrites of
PHIs and terminators which may be confusing.

While in this area I also noticed that VPScalarIVStepsRecipe
is missing a filled-in computeCost, which I'm happy to look
at in a follow-on PR.


artagnon (commented):
Not sure if we want to InstSimplify-fold the execute of SIVSteps: would be best to expand SIVSteps into executable recipes, similar to https://github.com/llvm/llvm-project/pull/187589?

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2580,21 +2581,27 @@ void VPDerivedIVRecipe::printRecipe(raw_ostream &O, const Twine &Indent,
...
-  IRBuilderBase &Builder = State.Builder;
 
   // Ensure step has the same type as that of scalar IV.
   Type *BaseIVTy = BaseIV->getType()->getScalarType();
   assert(BaseIVTy == Step->getType() && "Types of BaseIV and Step must match!");
 
+  // Use the instsimplify folder to clean up any inefficient IR due to
+  // add of zero, or multiply by one, etc.
+  InstSimplifyFolder Folder(State.Builder.GetInsertBlock()->getDataLayout());
+  IRBuilder<InstSimplifyFolder> Builder(BaseIV->getType()->getContext(),
```
fhahn: IIRC using InstSimplifyFolder here is not safe in general; the main issue is that the IR is in an incomplete state (e.g. header phis won't have all their incoming values set correctly), which can cause miscompiles/crashes

david-arm: Ah ok I can believe that. I could go down the route of expanding ScalarIVSteps, although I don't think I currently have time for that right now. Well, for now I can either abandon this PR or manually avoiding doing the add or mul in the execute function, which is what I originally started doing.

david-arm:
I've put up a new version that applies the folds within the execute function. However, if you think there is no real value in this then I'm happy to abandon the PR. Thanks!

PR #197172, opened 2026-05-12
[InstCombine] Fix incorect `foldPowiReassoc` on signed overflow

el-ev:
Reproducer: 

```
#include <math.h>
#include <stdio.h>

__attribute__((noinline))
double f(double x) {
    return __builtin_powi(x, 1073741824) * __builtin_powi(x, 1073741824);
}

int main(void) {
    double r = f(2.0);
    printf("%f\n", r);
    return r == 0.0; // 0 = correct, 1 = miscompile
}
```

https://llvm.godbolt.org/z/sjK1EsGhx


llvm/test/Transforms/InstCombine/powi-mul-overflow.ll
nikic, line 1: Add these tests to the existing test file please.

llvm/lib/Transforms/InstCombine/InstCombineMulDivRem.cpp
```diff
@@ -633,9 +633,17 @@ Instruction *InstCombinerImpl::foldFPSignBitOps(BinaryOperator &I) {
 
 Instruction *InstCombinerImpl::foldPowiReassoc(BinaryOperator &I) {
   auto createPowiExpr = [](BinaryOperator &I, InstCombinerImpl &IC, Value *X,
-                           Value *Y, Value *Z) {
+                           Value *Y, Value *Z) -> Instruction * {
+    auto *CY = dyn_cast<ConstantInt>(Y);
+    auto *CZ = dyn_cast<ConstantInt>(Z);
+    if (CY && CZ) {
```
nikic: But doesn't this mean the problem still exists if the arguments don't happen to be constants? I think you need to either reject non-constants entirely or you need to use willNotOverflowAdd() and hope known bits can prove it for variables.

el-ev: fixed both. thanks for suggestions.

nikic (approved):
LGTM

Issue #197171, opened 2026-05-12
[Flang][Semantics] Flang incorrectly allows referencing same named entity from different modules.


ShashwathiNavada:
Flang incorrectly treats distinct submodule procedures imported from different modules as equivalent during use association. As a result, ambiguous procedure references are accepted instead of producing diagnostics.

Example:

```fortran
module submodules_03_one
   integer :: one_i
   interface
      subroutine inside_one()
      end subroutine
   end interface
 end module

 submodule (submodules_03_one) submodules_03_sub_one
 contains
   subroutine inside_one()
   one_i = 6
   end subroutine
 end submodule

 module submodules_03_two
   integer :: two_i
   interface
      subroutine inside_one()
      end subroutine
   end interface
 end module

 submodule (submodules_03_two) sub_one
   contains
   subroutine inside_one()
   two_i = 6
   end subroutine
 end submodule

 program p
 use submodules_03_one
 use submodules_03_two
 call inside_one()
 end program
```

eugeneepshteyn:
@ShashwathiNavada , here's the problem with the reproducer: the declarations in the interfaces are not marked as `module`, like this:

```fortran
  interface
    module subroutine inside_one()
    end subroutine
  end interface
```

Without `module` the `inside_one()` is considered some external subroutine and its declarations are merged.

Adding `module` results in the following correct diagnostics:
```
$ flang -c repro-3-experiment.f90 
error: Semantic errors in repro-3-experiment.f90
repro-3-experiment.f90:34:8: error: Reference to 'inside_one' is ambiguous
    call inside_one()
         ^^^^^^^^^^
repro-3-experiment.f90:32:7: 'inside_one' was use-associated from module 'm_one'
    use m_one
        ^^^^^
repro-3-experiment.f90:33:7: 'inside_one' was use-associated from module 'm_two'
    use m_two
        ^^^^^
```
With the original reproducer (without `module subroutine` in the interface), I think it may be useful to warn about duplicate interface defined in different USE-associated modules, but it shouldn't be an error.

ronGreenNV:
@eugeneepshteyn my belief is that it should be an error.  The other compilers I've tried all diagnose this as ERROR.  A warning may be ignored and behavior for the resolution of inside_one is left to what, the linker?   I believe Flang is doing the right thing here by making this diagnostic an error.  We could consider a more detailed diagnostic.  Of the other compilers, intel ifx may have the best diagnostic:

```
[...]

PR #197173, opened 2026-05-12
[Flang][Semantics] Diagnose missing MODULE prefix in submodule procedures

ShashwathiNavada:
Fixes #https://github.com/llvm/llvm-project/issues/197171

Add a portability warning when a submodule procedure is missing the MODULE prefix required by a MODULE PROCEDURE interface in its parent module.


eugeneepshteyn (changes requested):
This may be a correct fix for a wrong problem ;-)

I'm currently looking at the issue and will respond there with my investigation.

I'm going to set this PR to "request changes" to make it obvious that it shouldn't go in until I respond to the issue.

ShashwathiNavada:
Ping @eugeneepshteyn

ShashwathiNavada:
Gentle ping @eugeneepshteyn

ShashwathiNavada:
@eugeneepshteyn @jeanPerier
Could you please take a look at the PR when you have some time? I've updated it to add the warning as suggested. Please let me know if there's anything else that should be changed. 
Thanks!

flang/test/Semantics/resolve132.f90
eugeneepshteyn, line 1: What about this case:
```
module parent_mod_9
   interface
      module subroutine inside_one()
      end subroutine
   end interface
 end module

 submodule (parent_mod_9) sub_9
 contains
   subroutine inside_one()
   end subroutine
 end submodule
```
Here I think it's an opposite problem: the interface for `inside_one()` has `module`, but the submodule doesn't, so submodule's implementation is not in fact implementation of the interface. I think a similar warning should be also generated in this case.

eugeneepshteyn: Also consider the following. If you compile these files separately with `-pedantic`, there's no warning:

First file:
```
module parent_mod_11
   interface
      subroutine inside_one()
      end subroutine
   end interface
 end module
```

Second file:
```
submodule (parent_mod_11) sub_11
 contains
   subroutine inside_one()
   end subroutine
 end submodule
```

On the other hand, if you combine them, you get a warning:
```
$ flang -c -pedantic p14combined.f90 
p14combined.f90:14:15: portability: Subprogram 'inside_one' in this submodule hides an external interface from its parent module; did you mean 'MODULE SUBROUTINE'? [-Wportability]
     subroutine inside_one()
                ^^^^^^^^^^
```
I think module check needs to be adjusted to allow for this case.

ShashwathiNavada: @eugeneepshteyn Done! Thank you so much for pointing that out. I have updated the PR.

ShashwathiNavada:
Gentle ping ! @eugeneepshteyn.
I have addressed the comments. If there is anything else, please do let me know.
Thank you.

ShashwathiNavada:
@eugeneepshteyn Gentle Ping!

ShashwathiNavada:
@eugeneepshteyn Gentle Ping.

ShashwathiNavada:
@eugeneepshteyn Gentle Ping!

flang/lib/Semantics/resolve-names.cpp
```diff
@@ -5771,6 +5771,24 @@ const Symbol *SubprogramVisitor::CheckExtantProc(
...
     Symbol::Flag subpFlag, const parser::LanguageBindingSpec *bindingSpec,
     bool hasModulePrefix) {
+  if (!inInterfaceBlock() && currScope().IsSubmodule() && !hasModulePrefix) {
+    const Scope &parent{currScope().parent()};
+    if (parent.IsModule() || parent.IsSubmodule()) {
+      if (const Symbol *host{parent.FindSymbol(name.source)}) {
+        const Symbol &hostUlt{host->GetUltimate()};
+        const auto *hostSubp{hostUlt.detailsIf<SubprogramDetails>()};
+        if (hostSubp && hostSubp->isInterface() &&
+            hostUlt.attrs().test(Attr::MODULE)) {
```
eugeneepshteyn: Please look into `IsSeparateModuleProcedureInterface()`

eugeneepshteyn:
Let's consider the case of compiling the module and the submodule separately:

```
$ cat p17a.f90
module parent_mod_17
  interface
    module subroutine inside_one()
    end subroutine
  end interface
end module
$ cat p17b.f90
submodule (parent_mod_17) sub_17
contains
  subroutine inside_one()
  end subroutine
end submodule
$ flang -c -pedantic p17a.f90
$ flang -c -pedantic p17b.f90        # no warning
$ flang -c -pedantic p17a.f90 p17b.f90   # also no warning
$ cat p17a.f90 p17b.f90 > combined.f90
$ flang -fsyntax-only -pedantic combined.f90
combined.f90:10:14: portability: Subprogram 'inside_one' in this submodule is missing the MODULE prefix ... # only this warns
```

The reason is that `SemanticsContext::Warn` suppresses any warning whose location's scope ancestry contains a module-file scope (`IsInModuleFile`, Semantics/semantics.cpp). When the submodule is compiled separately, its parent module scope comes from the `.mod` file, so every warning emitted through `context().Warn(...)` anchored inside the submodule is silently dropped. The message is created, it just never gets out.

The fix is to emit through the lower-level overload with the resolve-names visitor's own `InModuleFile()` gate (which is true only while actually resolving module-file content), the way `fold-integer.cpp` already does to bypass the same check:

```cpp
context().messages().Warn(/*isInModuleFile=*/InModuleFile(),
    context().languageFeatures(), common::UsageWarning::Portability,
    name.source, "..."_port_en_US, ...);
```

With the fix, please add two tests:

* a two-file test: compile the module first (producing the `.mod`), then the submodule, and expect the warning;
* a three-file negative test: module + submodule + a descendant submodule, expecting _no_ warning when compiling the descendant (this pins the `.smod` re-read case).

eugeneepshteyn:
Also, I recommend checking that the PR description matches the latest code.

ShashwathiNavada:
@eugeneepshteyn Hey! I have done the said changes. Would be great if you look at the PR in your free time. Thank you!

ShashwathiNavada:
@eugeneepshteyn Gentle ping!

flang/lib/Semantics/resolve-names.cpp
```diff
@@ -5976,6 +5976,26 @@ const Symbol *SubprogramVisitor::CheckExtantProc(
...
+          context().messages().Warn(/*isInModuleFile=*/InModuleFile(),
+              context().languageFeatures(), common::UsageWarning::Portability,
+              name.source,
+              "Subprogram '%s' in this submodule is missing the MODULE prefix "
+              "to implement the module procedure interface from its parent; "
+              "did you mean 'MODULE %s'?"_port_en_US,
+              name.source,
+              subpFlag == Symbol::Flag::Subroutine ? "SUBROUTINE" : "FUNCTION");
```
eugeneepshteyn: Consider the following case:
```fortran
module m2
  interface
    module subroutine sub()
    end subroutine
  end interface
end module

submodule (m2) s2
contains
  integer function sub()
    sub = 2
  end function
end submodule
```
Here the interface declares `sub` as subroutine, but `sub` in a submodule is a function.

Compiling it results in the following message:
```
$ flang -fsyntax-only -pedantic ex-kind.f90
ex-kind.f90:10:20: portability: Subprogram 'sub' in this submodule is missing the MODULE prefix to implement the module procedure interface from its parent; did you mean 'MODULE FUNCTION'? [-Wportability]
```
This is not bad, but if user follows instructions and changes it to `module function`, it still cannot implement subroutine's interface. I'm wondering if we change the check to `subpFlag == hostSubp->isFunction() ? "FUNCTION" : "SUBROUTINE"`, maybe this error message would be more faithful to the reality.

This is not a blocking issue, just putting it out for your consideration.

flang/lib/Semantics/resolve-names.cpp
```diff
@@ -5976,6 +5976,26 @@ const Symbol *SubprogramVisitor::CheckExtantProc(
...
     bool hasModulePrefix) {
+  if (!inInterfaceBlock() && currScope().IsSubmodule() && !hasModulePrefix) {
+    const Scope &parent{currScope().parent()};
+    if (parent.IsModule() || parent.IsSubmodule()) {
+      if (const Symbol *host{parent.FindSymbol(name.source)}) {
+        const Symbol &hostUlt{host->GetUltimate()};
+        const auto *hostSubp{hostUlt.detailsIf<SubprogramDetails>()};
+        if (hostSubp && hostSubp->isInterface() &&
+            hostUlt.attrs().test(Attr::MODULE)) {
+          context().messages().Warn(/*isInModuleFile=*/InModuleFile(),
```
eugeneepshteyn, line 5989: A short comment on the context().messages().Warn(...) call explaining why it bypasses SemanticsContext::Warn (module-file scope-ancestry suppression) would protect it from a future "simplification" back to context().Warn — fold-integer.cpp carries such a comment at its analogous call.

eugeneepshteyn (approved):
Sorry for the delay. 

LGTM, with some additional suggestions for improvements.

flang/lib/Semantics/resolve-names.cpp
eugeneepshteyn: Since you took my suggestion, please add a test for it :-)

ShashwathiNavada: Done!

ShashwathiNavada:
@eugeneepshteyn Thank you so much for all the suggestions, now that all of them are addressed, Can I go ahead and merge this PR?

PR #197174, opened 2026-05-12
[clang][bytecode] Fix a diagnostic difference in bitcasts

tbaederr:
Don't immediately return failure and let it be handled by later checks.


clang/test/AST/ByteCode/cxx11.cpp
```diff
@@ -404,13 +404,12 @@ constexpr int PassByValue(Counter c) { return c.copies; }
...
 
 namespace PointerCast {
-  /// The two interpreters disagree here.
   struct S { int x, y; } s;
   constexpr S* sptr = &s;
   struct U {};
   struct Str {
-    int e : (Str*)(sptr) == (Str*)(sptr); // expected-error {{not an integral constant expression}} \
-                                          // expected-note {{cast that performs the conversions of a reinterpret_cast}}
+    int e : (Str*)(sptr) == (Str*)(sptr);
```
shafik, line 411: So we say these are gnu constant folding but gcc nor edg in gcc mode agree: https://gcc.godbolt.org/z/h5YrEMshE

tbaederr: 🤷‍♂️ I'm trying to match clang, not other compilers.

ojhunt: I'd be willing to accept changing the clang behavior to match other compilers - if not now, maybe a followup PR and have a note about where the logic has to be? Or even a - sigh - compatibility flag

ojhunt (approved):
I'm ok with the mismatch from gcc as the primary goal at the moment is for the byte code to produce identical behavior.

I would like a follow up issue filed on the behavior - especially since we warn on it being a gnu extension that gcc no longer seems to have.

PR #195886, opened 2026-05-05
Disable pre-link LTO vectorisation and loop unroll for at O2 and O3 optimisation levels

momchil-velikov:
[[RFC] Disable pre-LTO loop vectorisation](https://discourse.llvm.org/t/rfc-disable-pre-lto-loop-vectorisation)

Patches:

* A small refactor split out to: https://github.com/llvm/llvm-project/pull/196290

* **Disable pre-LTO `LoopVectorizePass` and `SLPVectorizerPass`**
In this PR

* **Disable pre-LTO `LoopUnrollPass`**
A number of loops end up as runtime-unrolled where in the original compiler they are vectorised+unrolled; the epilogue part of the runtime-unrolled loops persists through optimisations and is left in the final binary.
In this PR

* **Adjust `FunctionSpecialization` parameters**
Address performance regressions; because of the differences in loop unrolling, resp. function sizes,  the `FunctionSpecialization` starts missing some profitable specialisations.
Split out to: https://github.com/llvm/llvm-project/pull/196570

* **Move `LoopFullUnrollPass` after `LoopVectorizePass` in LTO pipeline**
Some functions were vectorised in the pre-LTO pipeline in the original compiler and unrolled/not-vectorised by `LoopFullUnrollPass` during LTO in the modified compiler.
Split out to: https://github.com/llvm/llvm-project/pull/197176

* **Add extra `SimplifyCFG`, `LICM`, and `EarlyCSE` to the late passes**
After the patches above there are still scattered opportunities for LICM and CSE, which show up as regressions.
(A variant of this change used GVN with marginally better results, however GVN PRE of loads combined with LICM memory value promotion resulted in massive regression due to excessive spills/reloads).
Split out to:https://github.com/llvm/llvm-project/pull/197678

PR #197176, opened 2026-05-12
[PassBuilder] Remove LoopFullUnrollPass from before LoopVectorize in the FullLTO pipeline

momchil-velikov:
This patch is in preparation for disabling the pre-LTO vectorisation in the FullLTO pipeline.

It removes  the `LoopFullUnrollPass` run before `LoopVectorizePass` in the full LTO pipeline.

Running the `LoopFullUnrollPass` before `LoopVectorizePass` causes functions to be unrolled before the vectoriser has had a chance to vectorise them (in the modified pipeline where vectorisation does not happen in the pre-link stage).


nikic (commented):
Maybe I'm misunderstanding the change, but doesn't this mean that we will no longer perform runtime unrolling during LTO at all?

momchil-velikov:
> Maybe I'm misunderstanding the change, but doesn't this mean that we will no longer perform runtime unrolling during LTO at all?

I guess so, yes, but really it does not sound right.

I'll investigate only removing the `LoopFullUnrollPass` in front of `LoopVectorize`,  but keep this `LoopUnrollPass` after `LoopVectorize`.

llvm/test/Transforms/PhaseOrdering/loop-vectorize-bfi.ll
```diff
@@ -1,6 +1,6 @@
 ; NOTE: Assertions have been autogenerated by utils/update_test_checks.py UTC_ARGS: --check-globals none --version 6
 ; REQUIRES: riscv-registered-target
-; RUN: opt -p 'lto<O3>' -mtriple riscv64 -mattr=+v -S < %s | FileCheck %s
+; RUN: opt -p 'loop-unroll,loop-vectorize' -mtriple riscv64 -mattr=+v -S < %s | FileCheck %s
```
fhahn: Why did this change?

momchil-velikov: I think that way the test keeps testing what it was supposed to test.

I meant to ask @lukel97 if it is so?

lukel97: IIRC this needs the LTO pipeline to trigger the original BFI invalidation issue. Does this test change if we keep the original `-p 'lto<O3>'`?

momchil-velikov: Yes, the test changes, the vectorisation does not happen in this case. I've added the `loop-unroll` to trigger the invalidation.

lukel97: Can you replace the test with this? This manually unrolls the loop which still allows the lto pipeline to be tested

```llvm
; NOTE: Assertions have been autogenerated by utils/update_test_checks.py UTC_ARGS: --check-globals none --version 6
; REQUIRES: riscv-registered-target
; RUN: opt -p 'lto<O3>' -mtriple riscv64 -mattr=+v -S < %s | FileCheck %s

; Test that BlockFrequencyInfo is invalidated after loop passes, so it's
; accurate whenever LoopVectorize uses it. LoopVectorizer requires that
; innermost loop headers have a greater than or equal to frequency than any
; block it dominates.

define void @f(i1 %x) !prof !0 {
; CHECK-LABEL: define void @f(
; CHECK-SAME: i1 [[X:%.*]]) local_unnamed_addr #[[ATTR0:[0-9]+]] {{.*}}{
; CHECK-NEXT:  [[ENTRY:.*]]:
; CHECK-NEXT:    [[DOTSCALAR:%.*]] = xor i1 [[X]], true
; CHECK-NEXT:    [[TMP0:%.*]] = insertelement <vscale x 2 x i1> poison, i1 [[DOTSCALAR]], i64 0
; CHECK-NEXT:    [[TMP1:%.*]] = shufflevector <vscale x 2 x i1> [[TMP0]], <vscale x 2 x i1> poison, <vscale x 2 x i32> zeroinitializer
; CHECK-NEXT:    br label %[[VECTOR_BODY:.*]]
; CHECK:       [[VECTOR_BODY]]:
; CHECK-NEXT:    [[INDEX:%.*]] = phi i64 [ 0, %[[ENTRY]] ], [ [[CURRENT_ITERATION_NEXT:%.*]], %[[VECTOR_BODY]] ]
; CHECK-NEXT:    [[AVL:%.*]] = phi i64 [ 65, %[[ENTRY]] ], [ [[AVL_NEXT:%.*]], %[[VECTOR_BODY]] ]
; CHECK-NEXT:    [[TMP2:%.*]] = call i32 @llvm.experimental.get.vector.length.i64(i64 [[AVL]], i32 2, i1 true)
; CHECK-NEXT:    [[TMP3:%.*]] = getelementptr [8 x i8], ptr null, i64 [[INDEX]]
; CHECK-NEXT:    call void @llvm.vp.store.nxv2i64.p0(<vscale x 2 x i64> poison, ptr align 8 [[TMP3]], <vscale x 2 x i1> [[TMP1]], i32 [[TMP2]])
; CHECK-NEXT:    [[TMP4:%.*]] = zext nneg i32 [[TMP2]] to i64
; CHECK-NEXT:    [[CURRENT_ITERATION_NEXT]] = add nuw i64 [[INDEX]], [[TMP4]]
; CHECK-NEXT:    [[AVL_NEXT]] = sub nuw nsw i64 [[AVL]], [[TMP4]]
; CHECK-NEXT:    [[TMP5:%.*]] = icmp eq i64 [[AVL_NEXT]], 0
; CHECK-NEXT:    br i1 [[TMP5]], label %[[EXIT:.*]], label %[[VECTOR_BODY]], !llvm.loop [[LOOP1:![0-9]+]]
; CHECK:       [[EXIT]]:
; CHECK-NEXT:    ret void
;
entry:
  br label %loop

loop:
  %iv = phi i64 [ %iv.next, %baz.1 ], [ 0, %entry ]
  %gep = getelementptr i64, ptr null, i64 %iv
  br label %foo

foo:
  br i1 %x, label %baz, label %bar

bar:
  store i64 0, ptr %gep
  br label %baz

baz:
  br i1 %x, label %baz.1, label %bar.1

bar.1:
  store i64 0, ptr %gep
  br label %baz.1

baz.1:
  %iv.next = add i64 %iv, 1
  %ec = icmp eq i64 %iv, 64
  br i1 %ec, label %exit, label %loop

exit:
  ret void
}

!0 = !{!"function_entry_count", i64 1}
```

momchil-velikov: Done.

lukel97:
Can you get some benchmark numbers on this for the Graviton4? Just to check there's no really big regressions. I've kicked off a run on some RISC-V hardware, I should have results tomorrow

llvm/lib/Passes/PassBuilderPipelines.cpp
```diff
@@ -2226,12 +2226,10 @@ PassBuilder::buildLTODefaultPipeline(OptimizationLevel Level,
...
   // FIXME: Add loop interchange.
 
-  // Unroll small loops and perform peeling.
-  LPM.addPass(LoopFullUnrollPass(Level.getSpeedupLevel(),
-                                 /* OnlyWhenForced= */ !PTO.LoopUnrolling,
-                                 PTO.ForgetAllSCEVInLoopUnroll));
   // The loop passes in LPM (LoopFullUnrollPass) do not preserve MemorySSA.
   // *All* loop passes must preserve it, in order to be able to use it.
+  // TODO: Explore enabling MemorySSA for loop passes after we remove the
+  // LoopFullUnrollPass.
```
lukel97: LoopFullUnrollPass is removed in this PR right? Should this comment say something more like "now that LoopFullUnrollPass is removed"

momchil-velikov:
> Can you get some benchmark numbers on this for the Graviton4? Just to check there's no really big regressions. I've kicked off a run on some RISC-V hardware, I should have results tomorrow

I don't have benchmarks on individual patches, just on everything together, results [here](https://github.com/llvm/llvm-project/pull/195886#issuecomment-4830998464)

lukel97:
> > Can you get some benchmark numbers on this for the Graviton4? Just to check there's no really big regressions. I've kicked off a run on some RISC-V hardware, I should have results tomorrow
> 
> I don't have benchmarks on individual patches, just on everything together, results [here](https://github.com/llvm/llvm-project/pull/195886#issuecomment-4830998464)

Is it possible to get a benchmark run on this individual patch? IIUC this patch needs to land first before https://github.com/llvm/llvm-project/pull/195886 so we need to make sure we don't regress in the meantime

momchil-velikov:
> > > Can you get some benchmark numbers on this for the Graviton4? Just to check there's no really big regressions. I've kicked off a run on some RISC-V hardware, I should have results tomorrow
> > 
> > 
> > I don't have benchmarks on individual patches, just on everything together, results [here](https://github.com/llvm/llvm-project/pull/195886#issuecomment-4830998464)
> 
> Is it possible to get a benchmark run on this individual patch? IIUC this patch needs to land first before #195886 so we need to make sure we don't regress in the meantime

Yeah, it's not a problem technically (I'll make a run and post the results later), but really these patches are not supposed to exist individually.

momchil-velikov:
Here are SPEC2026 results with this PR only on top of main, between neutral to positive (lower/negative is better)

| Benchmark        |      Value |
| ---------------- | ---------: |
| 706.stockfish_r  |      0.17% |
| 707.ntest_r      |      0.05% |
| 708.sqlite_r     |     -0.15% |
| 710.omnetpp_r    |     -0.34% |
| 714.cpython_r    |      0.03% |
| 721.gcc_r        |     -0.51% |
| 723.llvm_r       |     -0.08% |
| 727.cppcheck_r   |      0.07% |
| 729.abc_r        |     -0.09% |
| 734.vpr_r        |     -0.04% |
| 735.gem5_r       |      0.12% |
| 750.sealcrypto_r |      0.03% |
| 753.ns3_r        |     -0.11% |
| 777.zstd_r       |     -0.16% |
| **Geomean**      | **-0.07%** |

| Benchmark        |      Value |
| ---------------- | ---------: |
| 709.cactus_r     |     -0.04% |
| 722.palm_r       |     -1.49% |
| 731.astcenc_r    |     -0.06% |
| 736.ocio_r       |      0.06% |
| 737.gmsh_r       |      0.13% |
| 748.flightdm_r   |      0.20% |
| 749.fotonik3d_r  |     -0.01% |
| 765.roms_r       |      0.01% |
| 766.femflow_r    |     -0.07% |
| 767.nest_r       |     -0.15% |
| 772.marian_r     |     -0.01% |
| 782.lbm_r        |     -0.04% |
| **Geomean**      | **-0.12%** |

llvm/lib/Passes/PassBuilderPipelines.cpp
lukel97: I think this still needs updated

lukel97 (approved):
RISC-V performance looks similar, https://lnt.lukelau.me/db_default/v4/nts/1620?compare_to=1621

LGTM w/ that comment updated

llvm/lib/Passes/PassBuilderPipelines.cpp
momchil-velikov: Done

lukel97:
Can you go ahead and land this independently? It would be good to see all the pre-requisite PRs in tree so we can evaluate https://github.com/llvm/llvm-project/pull/195886 on its own before it lands.

nikic:
(Do not land before release branch)

nikic:
Is it possible to add a phase ordering test that shows a situation where this is beneficial? This will help regressing on future pipeline changes.

momchil-velikov:
> Is it possible to add a phase ordering test that shows a situation where this is beneficial? This will help regressing on future pipeline changes.

Luckily there's already the quant_4x4 test, I just added an LTO RUN line here: https://github.com/llvm/llvm-project/pull/210088

That test passes on main. It also passes with this PR (vectorisation happens in pre-link stage).

It would fail with **only** https://github.com/llvm/llvm-project/pull/195886 applied, but it will pass (with a FileCheck variable rename) when **both** this and https://github.com/llvm/llvm-project/pull/195886 are applied.

PR #197180, opened 2026-05-12
[BasicAA] Remove special malloc handling

nikic:
We currently assume that allocation functions don't ModRef other memory. However, this is something that should be controlled by the `memory` attribute on the allocator, which is typically inaccessiblememonly for things like malloc.

This code path specifically only affected hardcoded allocation functions from MemoryBuiltins (not those using allocator attributes). Nowadays, these are only the `operator new` family functions.

For those functions, we should not unconditionally assume that they don't access other memory: They are replaceable global allocators, which in principle can have arbitrary memory effects (they can just be pair-wise elided, but must be respected if not elided.)

https://github.com/llvm/llvm-project/pull/217652 changed clang to emit `memory(inaccessiblemem: readwrite, errnomem: write)` for `operator new` if `-fassume-sane-operator-new` is used (which is the default). We should not make any additional assumptions in BasicAA if the attribute is not present.

### DIFF
--- a/llvm/include/llvm/Analysis/MemoryBuiltins.h
+++ b/llvm/include/llvm/Analysis/MemoryBuiltins.h
@@ -59,11 +59,6 @@ isAllocationFn(const Value *V,
                function_ref<const TargetLibraryInfo &(Function &)> GetTLI);
 
 /// Tests if a value is a call or invoke to a library function that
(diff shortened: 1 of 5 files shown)


fhahn (approved):
LGTM, thanks!

antoniofrighetto (approved):
LGTM

PR #197184, opened 2026-05-12
[Support] Add a function to print the debug log

BStott6:
With `EnableDebugBuffering`, the debug log is stored in a circular buffer and printed, with a nice banner, on program termination - this is achieved via a signal handler. For in-process tool execution, such as for running the regression tests using daemon versions of the tools, we need to be able to trigger the printing/flushing of the debug log from the process itself. This PR just adds a small function `printDebugLog` which checks if debug output and debug log buffering are enabled and, if so, prints the debug log.

The code for printing the debug log in the signal handler is moved to a new function `printDebugLogImpl` which is called by the signal handler and `printDebugLog` - the reason this is separate from `printDebugLog` is to avoid running the option check in the signal handler implementation, in case options were reset before the signal handler is called, as this would be an unintentional behavioral change.

### DIFF
--- a/llvm/include/llvm/Support/Debug.h
+++ b/llvm/include/llvm/Support/Debug.h
@@ -104,6 +104,11 @@ LLVM_ABI extern bool EnableDebugBuffering;
 /// like: dbgs() << "foo" << "bar";
 LLVM_ABI raw_ostream &dbgs();
 
+/// If EnableDebugBuffering is true, this flushes the debug stream with
+/// the banner displayed, the same way it is printed automatically on
+/// program termination.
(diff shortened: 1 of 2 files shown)


OCHyams (approved):
LGTM for upstream

BStott6:
> LGTM for upstream

Cool, I actually just posted it upstream already as I judged it was a small enough change to do 'on the way upstream'

OCHyams:
> LGTM for upstream

Oh whoops, we are upstream already! IMO this seems fine, but let's wait to see if anyone else has any comments

nikic (approved):

Issue #167001, opened 2025-11-07
Support `import lldb` in a Python interpreter that's different from the one we link against


JDevlieghere:
When building LLDB agains the Python Limited C ABI [1], it should be possible to do `import lldb` in a Python interpreter that isn't the same version as the one we built and link against. 

- [x] [Make LLDB compatible with the Python Limited C API](https://github.com/llvm/llvm-project/issues/151617)
- [x] [Support a Stable ABI LLDB_PYTHON_EXT_SUFFIX](https://github.com/llvm/llvm-project/pull/166269)
- [ ] [Fix crash in `_PyObject_Malloc`](https://github.com/llvm/llvm-project/issues/167002)

[1] https://github.com/llvm/llvm-project/issues/151617

patryk4815:
I think we should remove the `libpython.so` dependency from `/site-packages/lldb/_lldb.cpython-312-x86_64-linux-gnu.so`, since it’s not actually needed there. `_lldb.so` is always loaded by an already running Python interpreter, so `libpython` is already present in the process.
`libpython.so` should only be required directly by the `lldb` binary, not by the Python extension module itself.

PR #197185, opened 2026-05-12
[lldb][Python] Include limited API library directory

Nerixyz:
On Windows, when compiling with the limited API in non-debug mode, Python will instruct the linker to link `python3.lib` [here](https://github.com/python/cpython/blob/f5fb491341e566bbaf17d9bf3e4ec3af4a56bb3f/PC/pyconfig.h#L344-L345).

Currently, we don't directly specify the library in CMake. We specify `python3xx.lib`. When the linker encounters the `python3.lib` requirement, it errors out, because it doesn't know where that is. With this PR, we add the directory of `python3xx.lib`. This is the same one as `python3.lib`. So the linker can find the library.

This is more of a temporary fix. As commented in the code, we should use `Python3_SABI_LIBRARY_DIRS` once we require CMake 3.26. Furthermore, I'm not fully sure why we link to `python3xx.lib` when the limited API is requested (presumably because of #167001?).

### DIFF
--- a/lldb/source/Plugins/ScriptInterpreter/Python/CMakeLists.txt
+++ b/lldb/source/Plugins/ScriptInterpreter/Python/CMakeLists.txt
@@ -37,6 +37,12 @@ set(python_plugin_sources
   Interfaces/ScriptedThreadPythonInterface.cpp
 )
 
+if (WIN32)
+  # Add the library directories to allow Python to use #pragma comment(lib, "python3.lib").
+  # FIXME: We should use Python3_SABI_LIBRARY_DIRS here, but this requires CMake 3.26.
+  set(PYTHON_SABI_LIBRARY_DIRS ${Python3_LIBRARY_DIRS})
+endif()
+
 if (LLDB_ENABLE_DYNAMIC_SCRIPTINTERPRETERS)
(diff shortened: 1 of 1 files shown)


charles-zablit (approved):
LGTM, let's wait for @JDevlieghere before merging.

charles-zablit:
Do we have a plan to enforce CMake > 3.26 on Windows?

DavidSpickett:
Last proposal was https://discourse.llvm.org/t/rfc-2026-upgrading-llvm-s-minimum-required-cmake-version/89829/17.

JDevlieghere (approved):
LGTM. If we want to avoid duplicating the comment, we can hoist this out of the if to have something like:

```
if (WIN32)
  # Comment
  set(PYTHON_SBIN_LIBRARY_DIR ${Python3_LIBRARY_DIRS})
endif()
```

and then unconditionally add this:

```
target_link_directories(lldbPluginScriptInterpreterPython PUBLIC ${PYTHON_SBIN_LIBRARY_DIR}
```

Not sure if that works if the variable is empty. If not we can just wrap it in an `if()` again (which makes this suggestion slightly less appealing)

PR #197187, opened 2026-05-12
[libc] Use correct include path for in_port_t.h

boomanaiden154:
llvm-libc-types/stdint-macros.h does not exist. Not sure why this was passing the CMake build, but this causes the bazel build to fail.


akuegel (approved):

labath (approved):
Huh. Thanks for the fix.

labath:
> Not sure why this was passing the CMake build

I suspect that's because nothing actually includes this file. I guess I should really add a test for it, even if it's just a bunch of types.

PR #196918, opened 2026-05-11
[LV] Complete VPWidenCastRecipe cost model

david-arm:
Now the legacy/vplan cost model assert has gone we can finish off the cost model for VPWidenCastRecipe to account for recipes with no underlying value.

PR #197188, opened 2026-05-12
[LV] Fix the cost model for freeze instructions

david-arm:
While working on a PR to add a cost model for VPDerivedIV recipes I noticed that a loop in or_reduction_with_freeze:

test/Transforms/LoopVectorize/AArch64/reduction-cost.ll

stopped vectorising because the cost model decided it was no longer worth it. However, the main cause of this was the incredibly high cost (14) of freeze for VF=2. We were using the cost of a vector mul instruction as a proxy for the freeze cost, which is incredibly bad for an AArch64 target without SVE since the operation needs scalarising.

As far as I understand, the freeze instruction does not lead to any actual code being generated and acts merely as a barrier to potentially unsafe optimisations. As such, I've updated the cost model to return 0 instead.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
+++ b/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
@@ -5524,9 +5524,7 @@ LoopVectorizationCostModel::getInstructionCost(Instruction *I,
       return InstructionCost::getInvalid();
     return TTI.getArithmeticInstrCost(Instruction::Mul, RetTy, Config.CostKind);
   case Instruction::Freeze:
-    // This will never lead to any real code being generated. It acts purely
-    // as a blocker for optimizations.
-    return 0;
+    return TTI.getInstructionCost(I, Config.CostKind);
   default:
     // This opcode is unknown. Assume that it is the same as 'mul'.
     return TTI.getArithmeticInstrCost(Instruction::Mul, VectorTy,
--- a/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
+++ b/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
@@ -5524,7 +5524,7 @@ LoopVectorizationCostModel::getInstructionCost(Instruction *I,
       return InstructionCost::getInvalid();
     return TTI.getArithmeticInstrCost(Instruction::Mul, RetTy, Config.CostKind);
   case Instruction::Freeze:
-    return TTI.getInstructionCost(I, Config.CostKind);
+    return TTI::TCC_Free;
   default:
     // This opcode is unknown. Assume that it is the same as 'mul'.
     return TTI.getArithmeticInstrCost(Instruction::Mul, VectorTy,
--- a/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
+++ b/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
@@ -5523,6 +5523,10 @@ LoopVectorizationCostModel::getInstructionCost(Instruction *I,
     if (VF.isScalable())
       return InstructionCost::getInvalid();
     return TTI.getArithmeticInstrCost(Instruction::Mul, RetTy, Config.CostKind);
+  case Instruction::Freeze:
+    // This will never lead to any real code being generated. It acts purely
+    // as a blocker for optimizations.
+    return 0;
   default:
     // This opcode is unknown. Assume that it is the same as 'mul'.
     return TTI.getArithmeticInstrCost(Instruction::Mul, VectorTy,
--- a/llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
+++ b/llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
@@ -1000,9 +1000,9 @@ InstructionCost VPRecipeWithIRFlags::getCostForRecipeWithOpcode(
         RHSInfo, Operands, CtxI, &Ctx.TLI);
   }
   case Instruction::Freeze:
-    // This opcode is unknown. Assume that it is the same as 'mul'.
-    return Ctx.TTI.getArithmeticInstrCost(Instruction::Mul, ResultTy,
-                                          Ctx.CostKind);
+    // This will never lead to any real code being generated. It acts purely
+    // as a blocker for optimizations.
+    return 0;
   case Instruction::ExtractValue:
     return Ctx.TTI.getInsertExtractValueCost(Instruction::ExtractValue,
                                              Ctx.CostKind);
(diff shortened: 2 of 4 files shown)


paulwalker-arm (approved):

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -992,9 +992,9 @@ InstructionCost VPRecipeWithIRFlags::getCostForRecipeWithOpcode(
         RHSInfo, Operands, CtxI, &Ctx.TLI);
   }
   case Instruction::Freeze:
-    // This opcode is unknown. Assume that it is the same as 'mul'.
-    return Ctx.TTI.getArithmeticInstrCost(Instruction::Mul, ResultTy,
-                                          Ctx.CostKind);
+    // This will never lead to any real code being generated. It acts purely
+    // as a blocker for optimizations.
+    return 0;
```
fhahn: Is there a reason to not call TTI? I'd expect that to just return 0 cost

llvm/test/Transforms/LoopVectorize/AArch64/reduction-cost.ll
huntergr-arm: Is this test still relevant given the cost is always 0 now?

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
david-arm: I didn't realise there was a TTI hook specifically for freeze. I can have a look.

david-arm: OK I see. getInstructionCost in llvm/Analysis/TargetTransformInfoImpl.h has

```
    case Instruction::Freeze:
      return TTI::TCC_Free;
```

Happy to use that then

llvm/test/Transforms/LoopVectorize/AArch64/reduction-cost.ll
david-arm: I personally think there is still value in having a test that uses freeze to demonstrate we can vectorise it sensibly. However, the comment about the function probably needs removing. Good spot!

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
david-arm: The only problem with getInstructionCost is that you have to pass in the exact operands. I think this is easier to do for the original scalar version.

huntergr-arm (approved):
LGTM

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
fhahn: Hmm, would be nice if we could keep both uses in sync. I guess from the current implementation, passing an empty operands array ref would be fine. Alternatively we could pass the underlying value of the operand, which we do in a number of other TTI calls already

david-arm: I'm not keen on passing in the underlying value, which could be null. That would mean we have yet another special case in the cost model where we just bail out and fall back on some other randomly incorrect cost. This is something I'm currently trying to fix up - see https://github.com/llvm/llvm-project/pull/196918 for an example. It feels like a lot of effort to basically return a value we know will be 0.

In my opinion, if we want to keep them in sync, the absolute best way to guarantee this is to either:

1. Explicitly return TCC_Free in both cases, or
2. Introduce a brand new getFreezeCost that only accepts a single input type as argument (because we don't have any operands available to pass in) and set the default to be TCC_Free.

All of the alternatives feel inferior to me. Currently the only publicly available interfaces available are:

```
  /// The returned cost is defined in terms of \c TargetCostConstants, see its
  /// comments for a detailed explanation of the cost values.
  LLVM_ABI InstructionCost getInstructionCost(const User *U,
                                              ArrayRef<const Value *> Operands,
                                              TargetCostKind CostKind) const;

  /// This is a helper function which calls the three-argument
  /// getInstructionCost with \p Operands which are the current operands U has.
  InstructionCost getInstructionCost(const User *U,
                                     TargetCostKind CostKind) const {
    SmallVector<const Value *, 4> Operands(U->operand_values());
    return getInstructionCost(U, Operands, CostKind);
  }
```

and since we don't actually have an instruction or indeed any operands whilst computing the vector cost, it's therefore impossible to have the costs in sync between the scalar and vector versions without creating a synthetic vector freeze in computeCost that we delete after invoking the hook.

fhahn: But we could still just pass an empty ArrayRef, assuming no TTI hook ever would look at the operand of the freeze, right? If it would, this would mean it may not be free on the target (otherwise why look at the operand?). This would at least surface that the assumption was violated.

david-arm: How? The TTI hook requires an instruction. I think it's a bad idea to pass in a null User.

david-arm: Also, passing in an empty ArrayRef is making the dangerous assumption that the TTI hook for all targets will never attempt to (quite validly) look at the operands. I think it would be a bug on the caller's side to pass in an empty set of operands, rather than a bug in the TTI override function.

david-arm: If this *must* be done via a TTI hook in both instances then I think it requires a new hook that is dedicated to freeze in the same way we have dedicated hooks for arithmetic instructions, etc. If that's what's required then I'm happy to do that, but otherwise I'd like to merge the PR in the next day or so to unblock other work.

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -5527,6 +5529,8 @@ LoopVectorizationCostModel::getInstructionCost(Instruction *I,
     if (VF.isScalable())
       return InstructionCost::getInvalid();
     return TTI.getArithmeticInstrCost(Instruction::Mul, RetTy, Config.CostKind);
+  case Instruction::Freeze:
+    return TTI.getInstructionCost(I, Config.CostKind);
```
fhahn: can we either assert that this returns `TTI::TCC_FREE`  or directly return `TTI::TCC_FREE` to keep both call sites in sync

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
fhahn: Ah right, I initially thought there would be a way to query with opcode. It's a shame there's not currently, but I don't think it's worth adding one just for Freeze. It would generally be useful to have a unified TTI function to get cost for opcode + specific types/args, but that would be something completely separate

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
david-arm: Done. Thanks!

PR #196903, opened 2026-05-11
[libc] Create socketcall helper for Linux syscall wrappers

labath:
This patch implements a generic socketcall variadic wrapper template under linux/syscall_wrappers/socketcall.h and refactors socket functions to use it.

Assisted by Gemini.

PR #197189, opened 2026-05-12
[libc] Remove legacy SYS_socketcall fallbacks

labath:
This patch removes legacy SYS_socketcall fallback paths from all Linux socket entry points. The individual syscall entry points were [added](https://github.com/torvalds/linux/commit/9dea5dc921b5f4045a18c63eb92e84dc274d17eb) in linux 4.3 (on x86, other architectures have had them even sooner). Our policy is to support the lowest kernel version on https://kernel.org/, which is 5.10 as of this writing.

This is motivated by the problems in
[testing](https://github.com/llvm/llvm-project/pull/196903#pullrequestreview-4263553670) the fallback paths -- to make sure this even builds, one needs to get a hold of very old kernel headers, or otherwise hack its build to force it to select the fallback path.

New ABIs don't have the accept syscall (only accept4), so I've added an accept->accept4 fallback.

Assisted by Gemini.


libc/src/sys/socket/linux/recv.cpp
kaladron: Can this be removed?  It's still in a few files:

1.  libc/src/sys/socket/linux/recv.cpp
2.  libc/src/sys/socket/linux/recvfrom.cpp
3.  libc/src/sys/socket/linux/recvmsg.cpp
4.  libc/src/sys/socket/linux/send.cpp
5.  libc/src/sys/socket/linux/sendmsg.cpp
6.  libc/src/sys/socket/linux/sendto.cpp

kaladron (approved):
One question within, thanks!

libc/src/sys/socket/linux/recv.cpp
labath: Yes it can. Done.

PR #197190, opened 2026-05-12
[llvm-shlib] Fix parallel build failure in gen-msvc-exports.py on Windows

jsji:
The script was failing with OSError [Errno 22] Invalid argument when
opening a temp file during parallel Windows builds. The root cause is
that mkstemp() creates and closes a file descriptor, then the script
re-opens it by path in a loop. On Windows, between the close and
re-open, antivirus software or filesystem contention from parallel
build processes can briefly lock the file, causing the subsequent
open() to fail with EINVAL.

Fix by replacing the temp-file-based approach with
subprocess.check_output(), which captures nm's stdout directly in
memory. This eliminates the temp file entirely, removing the race
condition and simplifying the code (removing the unused mkstemp,
contextmanager, and os imports along with the helper functions).

### DIFF
--- a/llvm/tools/llvm-shlib/gen-msvc-exports.py
+++ b/llvm/tools/llvm-shlib/gen-msvc-exports.py
@@ -23,11 +23,8 @@
 # You can use the --output flag to set the name of the export file.
 #
 # ===------------------------------------------------------------------------===#
-from tempfile import mkstemp
-from contextlib import contextmanager
-from subprocess import check_call
+from subprocess import check_output
 import argparse
-import os
 import re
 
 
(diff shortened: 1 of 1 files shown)


jsji:
Ping... If no objections, I will merge this. Thanks.

e-kud (approved):
I wasn't able to find anything about necessity to have the file in https://reviews.llvm.org/D35077. Tagging @Wallbraker just in case. 

I don't see anything critical. Maybe the only concern is about quite long `nm` output. But I don't think that it is a real problem.

LGTM

jsji:
> I wasn't able to find anything about necessity to have the file in https://reviews.llvm.org/D35077. Tagging @Wallbraker just in case.
> 
> I don't see anything critical. Maybe the only concern is about quite long `nm` output. But I don't think that it is a real problem.
> 
> LGTM

Thanks @e-kud .

PR #197191, opened 2026-05-12
[libomp] OpenMP 6.0 Device-Scope Environment Variables 

loopacino:
The patch introduces `_ALL` `_DEV` `_DEV_<id>` env var suffixes to the LLVM's OpenMP host runtime `libomp`.  The eligible-list env var currently supports `OMP_NUM_THREADS`. 

Assisted by Claude Opus 4.7 / Claude 4.6 Sonnet

### DIFF
--- a/openmp/runtime/src/exports_so.txt
+++ b/openmp/runtime/src/exports_so.txt
@@ -53,6 +53,8 @@ VERSION {
         __kmp_thread_pool;
 
 	__kmp_reset_stats;
+        __kmp_resolve_host_env;
+        __kmp_resolve_device_env;
 
 #if USE_ITT_BUILD
         #
--- a/openmp/runtime/src/kmp_device_env.cpp
+++ b/openmp/runtime/src/kmp_device_env.cpp
@@ -7,8 +7,7 @@
 //===----------------------------------------------------------------------===//
 //
 // OpenMP 6.0 device-scope env-var registry. See kmp_device_env.h for the
-// public-internal contract; see openmp/docs/OpenMP_6_0_DeviceScope_EnvVars_
-// Walkthrough.md for the design notes and examples.
+// public-internal contract;
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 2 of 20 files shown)


loopacino:
ping

loopacino:
ping

openmp/runtime/src/exports_so.txt
hansangbae, line 57: Exporting symbols for testing purposes does not seem ideal.

I wonder if it is possible to
1) Use only standard API to test host env-vars and remove `__kmp_resolve_host_env`. Whether an env-var is correctly initializing ICV can be tested with existing OpenMP API routines in my opinion.
2) Rename `__kmp_resolve_device_env` to `__kmpc_get_device_env` (for example) and remove it from the exported symbols (all `__kmpc_*` symbols are already exported). I believe we need a runtime entry like this for the offload runtime. It needs to be listed in `dllexports` for Windows platform.

openmp/runtime/src/kmp_device_env.cpp
hansangbae, line 27: Do we expect additional table entries later?

hansangbae, line 63: We can just return `sizeof(__kmp_device_env_eligible_names) / sizeof(const char *) - 1`.

hansangbae, line 150: It looks like `000000000001` is rejected.

openmp/runtime/src/exports_so.txt
loopacino: Thanks for the review.

Removed `__kmp_resolve_host_env` and now the host behaviour is addressed by standard API (like `omp_get_max_threads()`). Renamed device lookup to `__kmpc_get_device_env` (also listed in `dllexports`) and removed extra entries from `exports_so.txt` as suggested.

openmp/runtime/src/kmp_device_env.cpp
loopacino: Yes, and might require its own review as an addition in the eligible list.

loopacino: Done.

loopacino: Hmm yeah. The length check is removed.

hansangbae (approved):
LGTM

PR #195518, opened 2026-05-03
[VPlan] Move call widening decision to VPlan. (NFCI)

fhahn:
This patch adds a new makeCallWideningDecisions transform which converts Call VPInstructions to VPWidenCallRecipe/VPWidenIntrinsicRecipe/VPReplicateRecipe depending on their costs.

To compute the costs, static helpers are introduced to re-use the existing VPlan cost model logic:
 * VPWidenIntrinsicRecipe::computeCallCost
 * VPReplicateRecipe::computeCallCost

The cost-model logic is still retained; we assert that the decisions match to make sure we do not miss any edge cases. The legacy logic will be removed in a follow-up.

PR #197192, opened 2026-05-12
[NFCI][LV] Remove unsupported variable mask position for vector variants.

huntergr-arm:
We only support masks as the last argument of a vector function variant.
Make it official and remove the interface to query the position of the
mask.

### DIFF
--- a/llvm/include/llvm/IR/VFABIDemangler.h
+++ b/llvm/include/llvm/IR/VFABIDemangler.h
@@ -128,31 +128,23 @@ struct VFInfo {
   std::string VectorName; /// Vector Function Name associated to this VFInfo.
   VFISAKind ISA;          /// Instruction Set Architecture.
 
-  /// Returns the index of the first parameter with the kind 'GlobalPredicate',
-  /// if any exist.
-  std::optional<unsigned> getParamIndexForOptionalMask() const {
-    unsigned ParamCount = Shape.Parameters.size();
-
+  /// Returns true if the last operand to the vectorized function has the
+  /// kind 'GlobalPredicate'.
+  bool isMasked() const {
+    // There should be at most one GlobalPredicate parameter.
 #ifndef NDEBUG
     unsigned NumMaskParams =
         llvm::count_if(Shape.Parameters, [](const VFParameter &I) {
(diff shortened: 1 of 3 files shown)


huntergr-arm:
Should  alleviate some of the confusion in #195518

llvm/include/llvm/IR/VFABIDemangler.h
```diff
@@ -128,31 +128,23 @@ struct VFInfo {
...
 
-  /// Returns the index of the first parameter with the kind 'GlobalPredicate',
-  /// if any exist.
-  std::optional<unsigned> getParamIndexForOptionalMask() const {
-    unsigned ParamCount = Shape.Parameters.size();
-
+  /// Returns true if the last operand to the vectorized function has the
+  /// kind 'GlobalPredicate'.
+  bool isMasked() const {
+    // There should be at most one GlobalPredicate parameter.
```
artagnon:
```suggestion
```

Maybe improve the assert message instead?

artagnon (approved):
LGTM, thanks!

llvm/include/llvm/IR/VFABIDemangler.h
huntergr-arm: done, thanks.

david-arm (approved):
LGTM!

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -994,24 +994,23 @@ class LoopVectorizationCostModel {
     InstWidening Kind;
     Function *Variant;
     Intrinsic::ID IID;
-    std::optional<unsigned> MaskPos;
+    bool RequiresMask;
```
fhahn: I think we could also get rid of that field, by comparing the number of arguments the variant expect vs the arguments for the call, as in the latest version of https://github.com/llvm/llvm-project/pull/195518?

huntergr-arm: done, thanks.

fhahn (approved):
LGTM, thanks

PR #197193, opened 2026-05-12
[VPlan] Pass ArrayRef<VPValue *> instead ArrayRef<const VPValue *> (NFC)

fhahn:
Operands are stored in SmallVector<VPValue *>. This means we cannot automatically convert this to ArrayRef<const VPValue *>. Update to pass ArrayRef<VPValue *> to some helpers, to avoid creating temporary SmallVectors.

I'm not 100% sure this is actually worthwhile.


artagnon (commented):
I personally think this is worthwhile: can you inline the ArrayRefs?

artagnon (commented):
Should close?

fhahn:
Yep, this is no longer needed

PR #197194, opened 2026-05-12
[NFC][LLVM][VPlan] Fix "parameter ‘P’ set but not used" warning.

paulwalker-arm:
For Is... = {} the fold expression short-circuits to true and does not evaluate P.


fhahn (approved):
LGTM with formatting fixed, thanks

david-arm (approved):
Stunning!

PR #197195, opened 2026-05-12
[DAG] canCreateUndefOrPoison - fmaxnum/fminnum/fmaximum/fminimum/fmaximumnum/fminimumnum don't create poison

RKSimon:
Test coverage is proving tricky due to lack of folds that work with these - I'm open to suggestions if we don't want to just eyeball this.


nikic (approved):
LGTM

arsenm (approved):

PR #164916, opened 2025-10-24
[LTO][LLD] Prevent invalid LTO libfunc transforms

mysterymath:
In LTO, part of LLVM's middle-end runs after linking has finished. LTO's semantics depend on the complete set of extracted bitcode files being known at this time. If the middle-end inserts new calls to library functions (libfuncs) that are implemented in bitcode, this could extract new bitcode object files into the link. These cannot be compiled, leading to undefined symbol references.

Additionally, the middle-end in LTO may reason that such library functions have no references, and it may internalize them, then manipulate their API or even delete them. Afterwards, it may emit a call to them, again producing undefined symbol references.

This patch resolves the former issue by ensuring that the middle end emits no new references to symbols defined in bitcode, and it resolves the latter issue by ensuring that extracted bitcode for libfuncs is considered external, since new calls may be emitted to them at any time.

The new semantics are not yet established for MachO LLD, which does not yet appear to have any special handling for libcalls in LTO. It also does not yet support distributed ThinLTO; doing so would require additional (de)serialization work.

This is the patch referenced in @ilovepi's and my talk at the last LLVM devmeeting: "LT-Uh-Oh"

Gemini 3.1 was used in porting to COFF and WASM LLDs.

PR #197199, opened 2026-05-12
[Clang] Mark new as inaccessiblememonly if sane

nikic:
If `-fassume-sane-operator-new` (the default), assume that `operator new` does not read or write accessible memory.

Currently, this assumption already exists due to special treatment in BasicAA. I'd like to remove this special treatment (see https://github.com/llvm/llvm-project/pull/197180), and instead rely only on the `memory` attribute.

It's worth noting that this is consistent with GCC's interpretation of the flag (where it is also enabled by default):

> [...] With -fassume-sane-operators-new-delete option GCC may assume that calls to the replaceable global operators from new or delete expressions or from __builtin_operator_new or __builtin_operator_delete calls don’t read or modify any global variables or variables whose address could escape to the operators (global state; except for errno for the new and new[] operators). [...]

### DIFF
--- a/clang/lib/CodeGen/CGCall.cpp
+++ b/clang/lib/CodeGen/CGCall.cpp
@@ -2615,11 +2615,14 @@ void CodeGenModule::ConstructAttributeList(StringRef Name,
       AddAttributesFromFunctionProtoType(
           getContext(), FuncAttrs, Fn->getType()->getAs<FunctionProtoType>());
       if (AttrOnCallSite && Fn->isReplaceableGlobalAllocationFunction()) {
-        // A sane operator new returns a non-aliasing pointer.
+        // A sane operator new returns a non-aliasing pointer and does not
+        // read or write accessible memory.
         auto Kind = Fn->getDeclName().getCXXOverloadedOperator();
         if (getCodeGenOpts().AssumeSaneOperatorNew &&
-            (Kind == OO_New || Kind == OO_Array_New))
+            (Kind == OO_New || Kind == OO_Array_New)) {
           RetAttrs.addAttribute(llvm::Attribute::NoAlias);
+          FuncAttrs.addMemoryAttr(llvm::MemoryEffects::inaccessibleMemOnly());
+        }
       }
       const CXXMethodDecl *MD = dyn_cast<CXXMethodDecl>(Fn);
       const bool IsVirtualCall = MD && MD->isVirtual();
--- a/clang/lib/CodeGen/CGCall.cpp
+++ b/clang/lib/CodeGen/CGCall.cpp
@@ -2621,6 +2621,8 @@ void CodeGenModule::ConstructAttributeList(StringRef Name,
         if (getCodeGenOpts().AssumeSaneOperatorNew &&
             (Kind == OO_New || Kind == OO_Array_New)) {
           RetAttrs.addAttribute(llvm::Attribute::NoAlias);
+          // FIXME: inaccessiblemem could cause issues if LTO makes the
+          // previously inaccessible memory accessible after linking.
           FuncAttrs.addMemoryAttr(
               llvm::MemoryEffects::inaccessibleOrErrnoMemOnly(
                   llvm::ModRefInfo::ModRef, llvm::ModRefInfo::Mod));
--- a/clang/lib/CodeGen/CGCall.cpp
+++ b/clang/lib/CodeGen/CGCall.cpp
@@ -2621,7 +2621,9 @@ void CodeGenModule::ConstructAttributeList(StringRef Name,
         if (getCodeGenOpts().AssumeSaneOperatorNew &&
             (Kind == OO_New || Kind == OO_Array_New)) {
           RetAttrs.addAttribute(llvm::Attribute::NoAlias);
-          FuncAttrs.addMemoryAttr(llvm::MemoryEffects::inaccessibleMemOnly());
+          FuncAttrs.addMemoryAttr(
+              llvm::MemoryEffects::inaccessibleOrErrnoMemOnly(
+                  llvm::ModRefInfo::ModRef, llvm::ModRefInfo::Mod));
         }
       }
       const CXXMethodDecl *MD = dyn_cast<CXXMethodDecl>(Fn);
--- a/clang/lib/CodeGen/CGCall.cpp
+++ b/clang/lib/CodeGen/CGCall.cpp
@@ -2617,9 +2617,8 @@ void CodeGenModule::ConstructAttributeList(StringRef Name,
       if (AttrOnCallSite && Fn->isReplaceableGlobalAllocationFunction()) {
         // A sane operator new returns a non-aliasing pointer and does not
         // read or write accessible memory.
-        auto Kind = Fn->getDeclName().getCXXOverloadedOperator();
         if (getCodeGenOpts().AssumeSaneOperatorNew &&
-            (Kind == OO_New || Kind == OO_Array_New)) {
+            Fn->getDeclName().isAnyOperatorNew()) {
           RetAttrs.addAttribute(llvm::Attribute::NoAlias);
           // FIXME: inaccessiblemem could cause issues if LTO makes the
           // previously inaccessible memory accessible after linking.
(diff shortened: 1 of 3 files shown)


nikic:
Comparison of current behavior with and without sane operator new: https://clang.godbolt.org/z/oc1Gz4n6e

nikic:
It's worth noting that GCC also applies this to operator delete (https://clang.godbolt.org/z/3bqoq1e8e). I think it would make sense for us to do this as well, but I didn't include it in this PR, as that part would be a change in behavior.

nikic:
As a second commit, I've pushed a variant that also adds an `errnomem: write` effect, in order to match GCC documentation more closely. TBH I find the special provision for errno in this context somewhat odd (because as far as I know, `new` does not specify any specific behavior wrt errno, so this is just malloc behavior leaking through), but it's probably a good idea to match GCC semantics for the flag.

antoniofrighetto:
I wonder if this might ever somehow regress for type-aware overloads of new, which IIRC do not rely on malloc, possibly not a concern for the time being though, as I don't think they are encompassed in isReplaceableGlobalAllocationFunction()? (maybe cc/ @ojhunt)

efriedma-quic (commented):
For errno, it's "just" state leaking through... but it's hard to prevent in practice if you're relying on malloc.  In theory, your new operator which delegates to malloc could save/restore errno, but nobody does.

For inaccessiblememonly, I'm not quite confident it's safe.  Suppose the user defines new using a standard implementation... but they build it with -flto.  It's a problem if the implementation ends up in the same module as uses.  Consider, for example, what happens if we inline the implementation into some uses.  I think we need to model this somehow.  (See also discussion of malloc in #164916, which is a similar issue in a slightly different context.)

nikic:
> For inaccessiblememonly, I'm not quite confident it's safe. Suppose the user defines new using a standard implementation... but they build it with -flto. It's a problem if the implementation ends up in the same module as uses. Consider, for example, what happens if we inline the implementation into some uses. I think we need to model this somehow. (See also discussion of malloc in #164916, which is a similar issue in a slightly different context.)

Yes, I agree that this is a problem, but I believe it's a pre-existing one, because it also applies to the current special case in BasicAA.

Something we discussed in the past is to adjust inaccessiblemem effects during IR linking similar to how we drop nocallback. That would also cover the operator new case after this change.

efriedma-quic:
If we think inconsistencies involving inaccessiblememonly are unlikely to cause problems in practice, fine, but please leave a comment describing it as a known issue.

nikic:
> If we think inconsistencies involving inaccessiblememonly are unlikely to cause problems in practice, fine, but please leave a comment describing it as a known issue.

Done.

I've also done a quick experiment with adjusting inaccessiblemem in IRLinker (https://github.com/llvm/llvm-project/commit/c24a4a1902bc4697bd3462fbbb9106ff938a3f63) and the impact of that doesn't seem to be too bad, so this may be a viable direction.

ojhunt:
> I wonder if this might ever somehow regress for type-aware overloads of new, which IIRC do not rely on malloc, possibly not a concern for the time being though, as I don't think they are encompassed in isReplaceableGlobalAllocationFunction()? (maybe cc/ @ojhunt)

They're only considered replaceable in constexpr evaluation, but I'll look through,

ojhunt (commented):
I don't see any problem from the type aware allocators pov - to be replaceable it would take a non-template version (something like `void *operator new<std::type_identity<int>,....)`), which is a custom one, and so not going to be replaceable.

philnik777:
> As a second commit, I've pushed a variant that also adds an `errnomem: write` effect, in order to match GCC documentation more closely. TBH I find the special provision for errno in this context somewhat odd (because as far as I know, `new` does not specify any specific behavior wrt errno, so this is just malloc behavior leaking through), but it's probably a good idea to match GCC semantics for the flag.

It's originally Clang's extension AFAIK, so I don't think we need to restrict ourselves to what GCC documents.

clang/lib/CodeGen/CGCall.cpp
```diff
@@ -2615,11 +2615,18 @@ void CodeGenModule::ConstructAttributeList(StringRef Name,
       AddAttributesFromFunctionProtoType(
           getContext(), FuncAttrs, Fn->getType()->getAs<FunctionProtoType>());
       if (AttrOnCallSite && Fn->isReplaceableGlobalAllocationFunction()) {
-        // A sane operator new returns a non-aliasing pointer.
+        // A sane operator new returns a non-aliasing pointer and does not
+        // read or write accessible memory.
         auto Kind = Fn->getDeclName().getCXXOverloadedOperator();
         if (getCodeGenOpts().AssumeSaneOperatorNew &&
-            (Kind == OO_New || Kind == OO_Array_New))
+            (Kind == OO_New || Kind == OO_Array_New)) {
```
ojhunt: You can use `Fn->getDeclName().isAnyOperatorNew()` rather than manually through Kind (this came into existence with the type aware new/delete so it's a relatively recent wrapper for the above logic)

I should probably go through all the existing versions of this code at some point

ojhunt (approved):
One trivial code change suggestion but otherwise LGTM

nikic:
@efriedma-quic Are you fine with this as well?

efriedma-quic (approved):
LGTM

aengelke:
Is this intended to [severely regress compile time](https://llvm-compile-time-tracker.com/compare.php?from=b835d5a98f48cb9dcae3ec915a0790826eb7d2a2&to=01c93f511b4b5f734ac420709266036497610014&stat=instructions:u) and for 7zip also code size?

nikic:
I believe this is a representative example of the issue:
```llvm
@g = external global ptr

define i32 @test(ptr dereferenceable(16) %this) {
  %_length.i = getelementptr inbounds nuw i8, ptr %this, i64 8
  store i32 0, ptr %_length.i
  %call.i = call noalias ptr @_Znam(i64 8) #14
  store ptr %call.i, ptr @g
  %res = load i32, ptr %_length.i, !tbaa !20
  ret i32 %res
}

declare ptr @_Znam(i64)

attributes #14 = { builtin allocsize(0) memory(inaccessiblemem: readwrite, errnomem: write) }

!llvm.errno.tbaa = !{!4}

!4 = !{!5, !5, i64 0}
!5 = !{!"int", !6, i64 0}
!6 = !{!"omnipotent char", !7, i64 0}
!7 = !{!"Simple C++ TBAA"}
!9 = !{!"_ZTS11CStringBaseIwE", !10, i64 0, !5, i64 8, !5, i64 12}
!10 = !{!"p1 wchar_t", !11, i64 0}
!11 = !{!"any pointer", !6, i64 0}
!20 = !{!9, !5, i64 8}
```

We no longer perform store to load forwarding.

The issue is the `errnomem: write`. We consider the load here to alias errno. Previously this worked because we fell through to the special allocator handling (that I'm trying to remove).

GCC considers these to not alias: https://gcc.godbolt.org/z/ejT6doxv8

I think it would make sense to say that anything using struct TBAA cannot alias errno? Even if the member happens to be an int. (As usual, this may be slightly iffy when it comes to libc LTO, where errno might be part of a struct.)

ojhunt:
> I think it would make sense to say that anything using struct TBAA cannot alias errno? Even if the member happens to be an int. (As usual, this may be slightly iffy when it comes to libc LTO, where errno might be part of a struct.)

That seems reasonable, even though most operator new impls use malloc under the hood, it's unlikely anyone calling new is ever going to caring about the errno impact :D

nikic:
Opened https://github.com/llvm/llvm-project/pull/200367 to make use of the struct TBAA

> That seems reasonable, even though most operator new impls use malloc under the hood, it's unlikely anyone calling new is ever going to caring about the errno impact :D

Dropping the `errnomem: write` would be another option, which would avoid this issue entirely. We did recently add this to malloc though, so we should probably at least try to respect it for new, though the case is weaker there...

ojhunt:
> Opened #200367 to make use of the struct TBAA
> 
> > That seems reasonable, even though most operator new impls use malloc under the hood, it's unlikely anyone calling new is ever going to caring about the errno impact :D
> 
> Dropping the `errnomem: write` would be another option, which would avoid this issue entirely. We did recently add this to malloc though, so we should probably at least try to respect it for new, though the case is weaker there...

I think it's more reasonable to that anything with struct based TBAA cannot alias errno - I'd want to be even stricter and say anything other than `decltype(errno)*` isn't aliasing errno, which I recognize is incorrect w.r.t to the standard type aliasing rules, but it feels like passing pointers to errno via the wrong type must surely be uncommon? (I haven't ever seen code the passes errno around by address in general, but unlike an llm I didn't need to learn to program by copying the entire body of all code that has ever been written :D)

**I realized the dumbness: the actually a macro in most implementations isn't it, and turns into something like `*the_tls_storage_of_errno`

ojhunt:
TBAA non-aliasing of errno might be generally useful as well

PR #197202, opened 2026-05-12
[clang][NFC] Mark CWG941 as implemented and add a test

localspook:
[CWG941](https://wg21.link/cwg941) allowed specializing deleted function templates. Clang accepted this between 2.7 and 2.9, regressed and started emitting redefinition errors between 3.0 and 3.8, then went back to accepting in 3.9: https://godbolt.org/z/GKnf9je7j. I've marked it as implemented since 3.9.


cor3ntin (approved):
Thanks!

Endilll (commented):
Can you add a "member function template of a class template" case?

Endilll (approved):

PR #197208, opened 2026-05-12
[Hexagon] Support reserving R16-R28 registers via -ffixed-rN

quic-santdas:
Extend register reservation from R19-only to R16-R28. This allows users to reserve callee-saved registers (R16-R27) and R28 via command-line flags -ffixed-r16 through -ffixed-r28. The single bool ReservedR19 is replaced with an array-based approach (ReservedR[32]) to scale cleanly across all supported registers.


quic-k (commented):
nit: probably squash the formatting changes

clang/test/Driver/hexagon-toolchain-elf.c
```diff
@@ -544,14 +544,52 @@
 // CHECK360:      {{hexagon-link|ld}}
 
 // -----------------------------------------------------------------------------
-// ffixed-r19
-// -----------------------------------------------------------------------------
+// ffixed-r16 through ffixed-r28
+// -----------------------------------------------------------------------------
+// RUN: %clang -### --target=hexagon-unknown-elf -ffixed-r16 %s 2>&1 \
+// RUN:        | FileCheck --check-prefix=CHECK-R16 %s
+// CHECK-R16: "-target-feature" "+reserved-r16"
```
quic-k, line 551: should we have a codegen test as well?

clang/test/Driver/hexagon-toolchain-elf.c
```diff
@@ -544,14 +544,52 @@
...
+// RUN:        | FileCheck --check-prefix=CHECK-R27 %s
+// CHECK-R27: "-target-feature" "+reserved-r27"
+// RUN: %clang -### --target=hexagon-unknown-elf -ffixed-r28 %s 2>&1 \
+// RUN:        | FileCheck --check-prefix=CHECK-R28 %s
+// CHECK-R28: "-target-feature" "+reserved-r28"
 // RUN: %clang -### --target=hexagon-unknown-elf %s 2>&1 \
 // RUN:        | FileCheck --check-prefix=CHECK371 %s
+// CHECK371-NOT: "+reserved-r16"
 // CHECK371-NOT: "+reserved-r19"
+// CHECK371-NOT: "+reserved-r28"
```
quic-k: a regex to cover all cases would be better
// CHECK371-NOT: "+reserved-r([0-9]|[1-2][0-9]|3[0-1])"

llvm/lib/Target/Hexagon/HexagonSubtarget.h
```diff
@@ -286,7 +286,7 @@ class HexagonSubtarget : public HexagonGenSubtargetInfo {
   bool useHVX64BOps() const { return useHVXOps() && UseHVX64BOps; }
 
   bool hasMemNoShuf() const { return HasMemNoShuf; }
-  bool hasReservedR19() const { return ReservedR19; }
+  bool isRRegReserved(unsigned i) const { return ReservedR[i]; }
```
quic-k: unbounded array access, no out-of-range access right now, but an assert(16 <= i <= 28) would be safe

quic-santdas: Thanks! I will add the assertion.

clang/test/Driver/hexagon-toolchain-elf.c
quic-santdas: Thanks for the suggestion.

llvm/lib/Target/Hexagon/HexagonSubtarget.h
```diff
@@ -63,7 +63,7 @@ class HexagonSubtarget : public HexagonGenSubtargetInfo {
   bool HasPreV65 = false;
   bool HasMemNoShuf = false;
   bool EnableDuplex = false;
-  bool ReservedR19 = false;
+  bool ReservedR[32] = {};
```
quic-asaravan: RISC-V uses "std::bitset<Hexagon::NUM_TARGET_REGS> UserReservedRegister;" (RISCVSubtarget.h:112) indexed by the actual register enum value which seems to be correct by construction rather than a raw integer that implicitly assumes register numbers match array positions. This also enables overriding "TargetSubtargetInfo::isRegisterReservedByUser" instead of a custom isRRegReserved(), making the reservation querable by other passes. It's worth considering how RISC-V handles user reserved registers (RISCVRegisterInfo.cpp:128)

quic-santdas: Thanks for the comment, will push a patch.

quic-santdas:
Hi @quic-akaryaki , do you think the patch is good to go?

llvm/lib/Target/Hexagon/HexagonRegisterInfo.cpp
```diff
@@ -205,8 +205,13 @@ BitVector HexagonRegisterInfo::getReservedRegs(const MachineFunction &MF)
   for (auto Reg : Hexagon_MC::GetVectRegRev())
     Reserved.set(Reg);
 
-  if (MF.getSubtarget<HexagonSubtarget>().hasReservedR19())
-    Reserved.set(Hexagon::R19);
+  static const MCPhysReg RRegs[] = {
```
quic-akaryaki, line 208: You can use Register here.

quic-akaryaki (approved):
Only minor suggestion.

llvm/lib/Target/Hexagon/HexagonRegisterInfo.cpp
quic-santdas: MCPhysReg has lesser storage requirements vs Register class. The LLVM convention is that for static tables with known physical registers, we use MCPhysReg. Functionally they should be same.

quic-asaravan:
LGTM

PR #197214, opened 2026-05-12
[CUDA][HIP] Defer device diagnostics in implicit H+D explicit instantiations

yxsamliu:
When clang explicitly instantiates a class template, it must emit device-side
bodies for the implicit `__host__ __device__` members so the vtable and
instantiation symbols resolve. Some of those members chain into host-only
calls (for example libstdc++ destructors that eventually call ::operator
delete). If no device code actually uses the class, the user still sees
errors about calling a `__host__` function from device code, even though they
wrote no device code that touches it. Overload ambiguity in the same context
behaves the same way.

This patch defers device-side errors in implicit `__host__ __device__`
functions reached only via an explicit template instantiation. At the end of
the translation unit, clang checks whether a real device caller exists. If
one does, the deferred errors are surfaced with the usual call-stack notes.
If not, the diagnostics are dropped and the function gets a trap body, so the
vtable and instantiation symbols stay well-formed and the failure only fires
if device code ever calls the function. This works without -fgpu-defer-diag.


efriedma-quic (commented):
The basic problem is that the translation unit where the destruction happens isn't necessarily the same to the translation unit where the destructor is defined/instantiated.  So this is basically just guessing.

I'm not sure there's a great solution here... but it would be nice if we could at least cause a linker error if the heuristic goes wrong; crashing with a null pointer dereference on device is a difficult failure mode.

Artem-B:
I'm out of my depth with mechanics of vtable construction, so it's just a drive-by comment.

The issue looks somewhat similar to an issue with virtual functions. When they are one-side only, I believe we end up with a NULL vtable entry. I wonder if we could populate such ghost entries with a stub that aborts with a visible or easier to diagnose error. E.g. crashing on a distinct trap instruction to distinguish it from random null pointer crashes, or literally calling abort with a message.

That said, this class of problems existed since the beginning, but, AFAICT, it very rarely shows up in practice. It may not be worth complicating the error path handling here. Crash on null vtable entry is reasonably distinct from a generic null pointer crash to point investigation in the right direction.

In general, I'm OK with the patch approach. In this case runtime failure on NULL in the code path that is not intended to or unlikely to be executed is preferable to the failure to compile the code that would most likely work otherwise. Standard library support on the GPU has always been a best effort affair, so some sharp corners are expected.

yxsamliu:
> The basic problem is that the translation unit where the destruction happens isn't necessarily the same to the translation unit where the destructor is defined/instantiated. So this is basically just guessing.
> 
> I'm not sure there's a great solution here... but it would be nice if we could at least cause a linker error if the heuristic goes wrong; crashing with a null pointer dereference on device is a difficult failure mode.

Thanks for the review. I tried to build a case where another TU dispatches through a slot we filled and ends up unresolved, but extern template still re-instantiates the dtor as linkonce_odr on the device pass, and an out-of-line dtor decl in a header gets host-only inference so Sema rejects device use before the heuristic runs. I could not find a real cross-TU break path, though I agree the heuristic is local and could in principle be wrong. To address the diagnosis side, I just pushed an update that fills the slot with a tiny per-dtor internal trap stub instead of NULL — the stub name embeds the dtor's mangled name, so if the heuristic ever misses a case, the crash backtrace points right at the offending class rather than a bare null deref. Happy to iterate if you have a scenario in mind I missed.

yxsamliu:
> I'm out of my depth with mechanics of vtable construction, so it's just a drive-by comment.
> 
> The issue looks somewhat similar to an issue with virtual functions. When they are one-side only, I believe we end up with a NULL vtable entry. I wonder if we could populate such ghost entries with a stub that aborts with a visible or easier to diagnose error. E.g. crashing on a distinct trap instruction to distinguish it from random null pointer crashes, or literally calling abort with a message.
> 
Thanks. I took the trap-stub idea and applied it to this case in the latest update — the slot now points at a tiny internal function whose body is llvm.trap + unreachable, and its name embeds the dtor's mangled name so a backtrace identifies the offending class. I kept the change scoped to this heuristic and did not extend it to the long-standing NULL fill for one-side-only HD methods, since that's a broader behavior change worth its own discussion. Happy to revisit that separately if you think it's worthwhile.

Artem-B (approved):
LGTM for the trap as the safeguard. I'll defer to @efriedma-quic for the overall approval.

clang/lib/CodeGen/CGVTables.cpp
```diff
@@ -811,6 +813,72 @@ void CodeGenVTables::addVTableComponent(ConstantArrayBuilder &builder,
...
+                                          ? "__clang_hip_unreachable_dtor."
+                                          : "__clang_cuda_unreachable_dtor.");
+            StubName += CGM.getMangledName(GD);
+            llvm::Module &M = CGM.getModule();
+            llvm::Function *Stub = M.getFunction(StubName);
+            if (!Stub) {
+              llvm::FunctionType *FT = llvm::FunctionType::get(
+                  CGM.VoidTy, {CGM.GlobalsInt8PtrTy}, /*isVarArg=*/false);
+              Stub = llvm::Function::Create(
+                  FT, llvm::GlobalValue::InternalLinkage, StubName, &M);
```
efriedma-quic: SetInternalFunctionAttributes() here?

clang/lib/CodeGen/CGVTables.cpp
```diff
@@ -811,6 +813,72 @@ void CodeGenVTables::addVTableComponent(ConstantArrayBuilder &builder,
...
+          bool IsExplicitInst =
+              Spec && (Spec->getTemplateSpecializationKind() ==
+                           TSK_ExplicitInstantiationDeclaration ||
+                       Spec->getTemplateSpecializationKind() ==
+                           TSK_ExplicitInstantiationDefinition);
+          bool ClassUsedOnDevice = CGM.GetGlobalValue(CGM.getMangledName(GD));
+          if (IsImplicitHD && IsExplicitInst && !ClassUsedOnDevice) {
+            for (const auto *Ctor : Spec->ctors()) {
+              if (CGM.GetGlobalValue(
+                      CGM.getMangledName(GlobalDecl(Ctor, Ctor_Complete))) ||
```
efriedma-quic: Do we have any reason to expect these will be emitted before the vtable?

clang/lib/CodeGen/CGVTables.cpp
```diff
@@ -811,6 +813,72 @@ void CodeGenVTables::addVTableComponent(ConstantArrayBuilder &builder,
         return builder.add(
             llvm::ConstantExpr::getNullValue(CGM.GlobalsInt8PtrTy));
       }
+      // For an implicit H+D virtual dtor of an explicit template
+      // instantiation, force-emitting the dtor body on device can pull
+      // host-only callees (e.g. via libstdc++ destructor chains). Skip
+      // unless device code has already referenced the dtor or any ctor
+      // of the class (a ctor implies an instance, hence possible
+      // polymorphic delete through this vtable).
```
efriedma-quic: Please add an explanation here of the known gaps in this approach (specifically, the fact that we can't do the "ctor" check across translation units).

yxsamliu:
The previous iteration of this patch emitted a trap body for the implicit H+D destructor whenever the current TU did not reference it through normal code. The gap: the destructor may actually be a valid device function. Its body has no host-only calls. No real device errors. But the trap body still gets emitted. The vtable for the explicit instantiation in this TU then points at the trap. If another TU instantiates the same class on device with the real destructor, the two vtables compete under COMDAT linkage. The trap version can win. Device code that calls the destructor through the vtable then traps at runtime, even though a perfectly valid definition exists.

The new design classifies by what happens when clang tries to emit the destructor body.

1. The body causes no errors on the device side. The function gets a real definition. The vtable gets a real pointer. No trap.

2. The body produces device errors, and the function has a real device caller in this TU. The deferred errors are surfaced with the usual call-stack notes. The user sees the real chain.

3. The body produces device errors, and no real device caller exists. The diagnostics are dropped. The function gets a trap body. The vtable stays well-formed. The failure only fires if device code ever calls the function at runtime.

Valid destructors stay valid. Real errors get reported with real call chains. Unused destructors no longer break the vtable just because something deep in their chain is not device-callable.

clang/lib/CodeGen/CGVTables.cpp
yxsamliu: The redesign removes this inline stub. The trap body is now emitted into the existing llvm::Function for the dtor itself. Therefore no longer need this.

clang/lib/CodeGen/CGVTables.cpp
yxsamliu: New design: this inspection is gone entirely. CGVTables always writes the real function pointer into the slot. The defer-then-classify walk at end of Sema runs after all parsing/instantiation, so it sees the full call graph and decides per function whether to trap. The ordering concern doesn't apply because the decision is no longer made during vtable construction.

clang/lib/CodeGen/CGVTables.cpp
yxsamliu: The cross-TU ctor heuristic is gone in the new design. CGVTables always writes the real function pointer; the trap decision moves to per-function emit time. Added a brief comment at CGCXX.cpp `tryEmitCUDADeviceInvalidFunctionBody` explaining when the trap can actually fire: only when a host-allocated object with virtual members is destroyed on the device, which HIP documents as invalid.

yxsamliu:
gentle ping

clang/lib/Sema/Sema.cpp
```diff
@@ -2103,13 +2103,43 @@ void Sema::emitDeferredDiags() {
...
+      for (const PartialDiagnosticAt &PDAt : Pair.second) {
+        if (getDiagnostics().getDiagnosticLevel(PDAt.second.getDiagID(),
+                                                PDAt.first) >=
+            DiagnosticsEngine::Error) {
+          HasError = true;
+          break;
+        }
+      }
+      if (!HasError)
+        continue;
```
Artem-B: Nit: This begs for `HasError = any_of(Pair.second, [](){ return getDiagnostics().getDiagnosticLevel(...) })`

Artem-B (approved):
Nice.

clang/lib/Sema/Sema.cpp
yxsamliu: done

PR #197215, opened 2026-05-12
[Clang] Evaluate concepts in their declaration context.

### LINKED ISSUE #115838: [clang] C++20 Concept in Template Specialization Should Not See Protected Member.
This bug exists in all released clang versions with C++20 support up to trunk.

The following code defined classes `Publ`/`Priv`/`Prot` with different visibility of member `x`, and inherit into a templated class `D` specialized by its accessibility using C++20 concept.
Boolean var `has` shows the specialization route.

In theory, only public member is visible, while protected/private are not.
These are checked by `static_assert`.

Clang mistakenly thinks protected member from `class Prot` is accessible, and failed to compile this code at the last static_assert, while any versions of gcc and msvc works.

Even more confusing, if we do the exact same thing again, with another class `Same` identical to `Prot`, but query the concept directly before template instantiation, clang will give opposite answer.

```cc
template<typename T> concept has_x = requires(T t){{ t.x };};

class Publ { public:    int x = 0; };
class Priv { private:   int x = 0; };
class Prot { protected: int x = 0; };
class Same { protected: int x = 0; };

template<typename T> class D;
template<typename T> requires ( has_x<T>) class D<T>: public T { public: static constexpr bool has = 1; };
template<typename T> requires (!has_x<T>) class D<T>: public T { public: static constexpr bool has = 0; };

// "Same" is identical to "Prot" but queried before used.
static_assert(!has_x<Same>,  "Protected should be invisible.");
static_assert(!D<Same>::has, "Protected should be invisible.");

static_assert( D<Publ>::has, "Public should be visible.");
static_assert(!D<Priv>::has, "Private should be invisible.");
static_assert(!D<Prot>::has, "Protected should be invisible.");  // clang failed here.

int main() { return 0; }
```

- GCC 14.2 works.
    - https://godbolt.org/z/Pb9j5x3q8
- clang 19.1.0 failed.
    - https://godbolt.org/z/3E9qnxKbW
- MSVC 19.40 x64 works.
    - https://godbolt.org/z/Tf14M4dK4
- clang-cl 18.1.0 failed.
    - https://godbolt.org/z/rjPoE1ve9

### LINKED ISSUE #194803: static member function is not checcked correclty in requirement.
Clang in the trunk version on compiler explorer fails to check a requirement properly. The requirement in line 21 and the static_assert do not fail when instantiated through line 83. The static_assert in line 81 produces the correct error.  

The error should be in all cases that the method in line 48 is not marked as static.

https://godbolt.org/z/ExWY6dTYx

```c++
#include <utility>

template<typename T>
struct MyComplex {
};

namespace adTool {
  template<typename Base, typename... Args>
  concept has_expression_constructor = requires(Args&&... args) { Base::construct(std::forward<Args>(args)...); };

  template<typename T_ClassValue, typename T_Impl>
  struct MemberOperations {};
  template<typename T_Value, typename T_Impl>
  struct ExpressionInterface : MemberOperations<T_Value, T_Impl> {};

  template<typename T_Value, typename T_Identifier, typename T_Tape>
  struct ActiveValue : ExpressionInterface<T_Value, ActiveValue<T_Value, T_Identifier, T_Tape>> {
      using Value = T_Value;
      template<typename... Args>
        // Requirement is ok but should fail
        requires has_expression_constructor<MemberOperations<Value, ActiveValue>, Args...>
      explicit ActiveValue(Args&&... args) {
        // Static assert is ok but should fail
        static_assert(has_expression_constructor<MemberOperations<Value, ActiveValue>, Args...>);
      }
      explicit ActiveValue() {}
  };
}

// ------------------------------------------------------------------
// Double specialization
// ------------------------------------------------------------------
namespace adTool {
  struct PrimalValueTape {};
  using Active_double = ActiveValue<double, int, PrimalValueTape>;
}

// ------------------------------------------------------------------
// Complex specialization
// ------------------------------------------------------------------

// Member operations definition
template<typename T, typename Impl>
struct MyComplexMemberOperationsImpl {

    // Definition of expression constructor
    template<typename re_expr_t, typename im_expr_t>
    decltype(auto) construct(
            ::adTool::ExpressionInterface<T, re_expr_t> const& re_expr,
            ::adTool::ExpressionInterface<T, im_expr_t> const& im_expr
    )  {
      return;
    }
};

// Specialization of member operations for MyComplex types.
template<typename T, typename Impl>
struct adTool::MemberOperations<::MyComplex<T>, Impl> :
    public ::MyComplexMemberOperationsImpl<T, Impl> {
};

// Specialization of std::MyComplex<ActiveValue>>
template<typename T, typename Identifier, typename Tape>
struct MyComplex<adTool::ActiveValue<T, Identifier, Tape>> : public adTool::ActiveValue<MyComplex<T>, Identifier, Tape> {
    using Base = adTool::ActiveValue<MyComplex<T>, Identifier, Tape>;
    using Base::Base;
};

namespace adTool {
  template<typename T>
  using Active_complex = MyComplex<ActiveValue<T, int, PrimalValueTape>>;
}

// ------------------------------------------------------------------
// The test
// ------------------------------------------------------------------
using Real = adTool::Active_double;
using Complex = adTool::Active_complex<double>;
int main() {
  // Static assert fails for direct instantiation
  //static_assert(adTool::has_expression_constructor<MyComplexMemberOperationsImpl<double, Complex>, Real, Real>);
  Real a, b;
  Complex temp1(a, b);
  (void) temp1;

  return 0;
}
```

cor3ntin:
Concepts appearing in a constraint expression of a class member had access to both `this` and the private member of the class.

This changes fixes that by setting the concext to that of the context before evaluation of its constraint expression.

This is done after we have substituted the template argument.

Code in `Sema::isThisOutsideMemberFunctionBody` that no longer seems useful is renoved as it was interefering with this change.

This is not an implementation of CWG2589 - at least not a complete one, as we still check access when doing substitution in the parameter mapping.

Fixes #115838
Fixes #194803

### DIFF
--- a/clang/lib/Sema/SemaConcept.cpp
+++ b/clang/lib/Sema/SemaConcept.cpp
@@ -482,6 +482,9 @@ class ConstraintSatisfactionChecker {
   ConstraintSatisfaction &Satisfaction;
   bool BuildExpression;
 
+  // The closest concept declaration when evaluating atomic constraints.
+  ConceptDecl *ParentConcept = nullptr;
+
   // This is for TemplateInstantiator to not instantiate the same template
   // parameter mapping many times, in order to improve substitution performance.
   llvm::DenseMap<llvm::FoldingSetNodeID, TemplateArgumentLoc>
@@ -724,6 +727,13 @@ ExprResult ConstraintSatisfactionChecker::EvaluateSlow(
     Satisfaction.IsSatisfied = false;
     return ExprEmpty();
   }
+
+  // Make sure that concepts are not evaluated in the context they are used,
+  // i.e they should not have access to the current class object or its
+  // non-public members.
+  std::optional<Sema::ContextRAII> ConceptContext;
+  if (ParentConcept)
+    ConceptContext.emplace(S, ParentConcept->getDeclContext());
 
   Sema::ArgPackSubstIndexRAII SubstIndex(S, PackSubstitutionIndex);
   ExprResult SubstitutedAtomicExpr = EvaluateAtomicConstraint(
@@ -1034,6 +1044,9 @@ ExprResult ConstraintSatisfactionChecker::Evaluate(
     return ExprError();
 
   unsigned Size = Satisfaction.Details.size();
+
+  llvm::SaveAndRestore PushConceptDecl(
+      ParentConcept, cast<ConceptDecl>(ConceptId->getNamedConcept()));
 
   ExprResult E = Evaluate(Constraint.getNormalizedConstraint(), MLTAL);
 
--- a/clang/docs/ReleaseNotes.rst
+++ b/clang/docs/ReleaseNotes.rst
@@ -566,6 +566,8 @@ Bug Fixes to C++ Support
 - Clang now rejects constant template parameters with block pointer types, since these are not implemented anyway and would lead to crashes. (#GH189247)
 - Fixed a crash on error recovery when dealing with invalid templates. (#GH183075)
 - Fixed a crash when instantiating ``requires`` expressions involving substitution failures in C++ concepts. (#GH176402)
+- Concepts appearing in the require-clause of a member function no longer have access to non-public members of that class,
+  or to a current class object. (#GH115838) (#GH194803)
 - We no longer caches invalid variable specializations. (#GH132592)
 - Fixed an incorrect template argument deduction when matching packs of template
   template parameters when one of its parameters is also a pack. (#GH181166)
--- a/clang/lib/Sema/SemaExprCXX.cpp
+++ b/clang/lib/Sema/SemaExprCXX.cpp
@@ -1485,11 +1485,6 @@ void Sema::MarkThisReferenced(CXXThisExpr *This) {
 }
 
 bool Sema::isThisOutsideMemberFunctionBody(QualType BaseType) {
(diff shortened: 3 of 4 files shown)


clang/lib/Sema/SemaConcept.cpp
```diff
@@ -725,6 +728,13 @@ ExprResult ConstraintSatisfactionChecker::EvaluateSlow(
     return ExprEmpty();
   }
 
+  // Make sure that concepts are not evaluated in the context they are used,
+  // i.e they should not have access to the current class object or its
+  // non-public members.
+  std::optional<Sema::ContextRAII> ConceptContext;
+  if (ParentConcept)
+    ConceptContext.emplace(S, ParentConcept->getDeclContext());
```
erichkeane, line 736: This ends up setting the 'this' override, do we mean that?

erichkeane (approved):

clang/lib/Sema/SemaConcept.cpp
erichkeane: Well huh, the 'NewThisContext' boolean variable apparently doesn't CREATE a new `This` context, it CLEARS the `This` context.  That isn't confusing at all.

erichkeane (approved):

PR #196309, opened 2026-05-07
[LLVM][InstSimplify] Refactor simplifyBinaryIntrinsic to remove Call operand.

paulwalker-arm:
I picked this up from https://github.com/llvm/llvm-project/pull/194825#issuecomment-4343487394 because I have a low priority thread where I'm trying to simplify/reduce the uses of vscale_range.

PR #197216, opened 2026-05-12
[LLVM] Refactor FoldBinaryIntrinsic to remove Instruction* dependency.

paulwalker-arm:
As suggested by https://github.com/llvm/llvm-project/pull/196309#discussion_r3203960471.

All existing uses pass nullptr for FMSource thus making the operand redundant. Rather than remove it I have replaced the operand with explicit fast-math flags and updated IRBuilder to use it.

NOTE: An existing use did pass FMSource to ConstantFoldBinaryIntrinsic which in turn passed it to ConstantFoldIntrinsicCall2 as its "call" operand. This looks invalid and with no live uses it seemed easiest to remove the operand.


nikic (approved):
LGTM

PR #197218, opened 2026-05-12
[lldb][windows] add sibling-import fallback to LoadImage helper

charles-zablit:
`PlatformWindows::DoLoadImage` injects `__lldb_LoadLibraryHelper` which registers each caller-supplied search path via `AddDllDirectory()` and calls `LoadLibraryExW(name, nullptr, LOAD_LIBRARY_SEARCH_DEFAULT_DIRS)`. In practice however, this flow fails for a common layout where a DLL lives next to its sibling imports. LoadLibraryExW returns ERROR_CANT_RESOLVE_FILENAME and the load is reported as unrecoverable, even though every required DLL is present in one of the search paths.

This patch adds a fallback after the first `LoadLibraryExW(name, nullptr, LOAD_LIBRARY_SEARCH_DEFAULT_DIRS)` call: if it fails, iterate the caller-provided search paths and try `LoadLibraryExW(path, nullptr, LOAD_WITH_ALTERED_SEARCH_PATH)`. `LOAD_WITH_ALTERED_SEARCH_PATH` instructs the Windows loader to add the DLL's own directory to the dependency search.

See https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibraryexa.

### DIFF
--- a/lldb/source/Plugins/Platform/Windows/PlatformWindows.cpp
+++ b/lldb/source/Plugins/Platform/Windows/PlatformWindows.cpp
@@ -645,6 +645,11 @@ extern "C" {
 // application should include in its DLL search path.
 #define LOAD_LIBRARY_SEARCH_DEFAULT_DIRS 0x00001000
 
+// If this value is used, and lpFileName specifies an absolute path, the system
+// uses the alternate file search strategy to find associated executable
+// modules.
+#define LOAD_WITH_ALTERED_SEARCH_PATH 0x00000008
+
 // WINBASEAPI DWORD WINAPI GetLastError(VOID);
 /* __declspec(dllimport) */ uint32_t __stdcall GetLastError();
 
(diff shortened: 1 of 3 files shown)


lldb/source/Plugins/Platform/Windows/PlatformWindows.cpp
```diff
@@ -688,6 +693,32 @@ void * __lldb_LoadLibraryHelper(const wchar_t *name, const wchar_t *paths,
...
+      size_t plen = wcslen(path);
+      size_t nlen = wcslen(name);
+      // Need room for: path + '\\' + name + '\0'
+      if (plen + 1 + nlen + 1 > 4096)
+        continue;
+      wchar_t *p = full;
+      for (size_t i = 0; i < plen; ++i)
+        *p++ = path[i];
+      *p++ = L'\\';
+      for (size_t i = 0; i <= nlen; ++i) // copy name including trailing '\0'
```
Nerixyz:
```suggestion
      for (size_t i = 0; i <= nlen; ++i) // Copy name including trailing '\0'.
```

Nerixyz (commented):
This looks reasonable.

> This enables 2 tests and fixes an issue when running tests on Windows in swiftlang lldb.

I don't understand why these tests are fixed with this - they don't explicitly load anything. Which library did they attempt to load and why?

charles-zablit:
> I don't understand why these tests are fixed with this - they don't explicitly load anything. Which library did they attempt to load and why?

I was originally working on this fix in my swiftlang checkout. This fixes tests in the swiftlang fork. The 2 other tests enablements are not related, I split them up in a separate PR: https://github.com/llvm/llvm-project/pull/197251/changes

Nerixyz (approved):

PR #197223, opened 2026-05-12
[MLIR][XeGPU] Improve deinterleave/interleave/dpas_mx ops handling

akroviakov:
This PR adds some of the necessary bits to enable the end-to-end _distribution_ of `dpas_mx` examples.

Some minor changes include:
- Assembly is changed in dpas_mx definition because starting multiple optional groups with the same character is ambiguous.
- Use layout info when possible during setting result layouts of interleave/bitcast, instead of hardcoding subgroup size.
- Add canonicalization and cse after the linearization pass in the master pipeline (interleaves/deinterleaves/transpose will be linearized, optimize foldable ops, and remove remaining shape casts).
- Make `convert_layout` strictly illegal in sg-to-wi.
- Enable shape cast layout inference to handle trailing unit dims.

### DIFF
--- a/mlir/lib/Dialect/GPU/Pipelines/GPUToXeVMPipeline.cpp
+++ b/mlir/lib/Dialect/GPU/Pipelines/GPUToXeVMPipeline.cpp
@@ -96,6 +96,8 @@ void buildGPUPassPipeline(OpPassManager &pm,
     pm.addNestedPass<gpu::GPUModuleOp>(createLoopInvariantCodeMotionPass());
     pm.addNestedPass<gpu::GPUModuleOp>(createCSEPass());
     pm.addNestedPass<gpu::GPUModuleOp>(xegpu::createXeGPUVectorLinearize());
+    pm.addNestedPass<gpu::GPUModuleOp>(createCanonicalizerPass());
+    pm.addNestedPass<gpu::GPUModuleOp>(createCSEPass());
   }
   pm.addNestedPass<gpu::GPUModuleOp>(createConvertMathToXeVM());
   ConvertXeGPUToXeVMPassOptions xegpuToXeVMOptions;
--- a/mlir/lib/Dialect/XeGPU/Transforms/XeGPUSgToWiDistributeExperimental.cpp
+++ b/mlir/lib/Dialect/XeGPU/Transforms/XeGPUSgToWiDistributeExperimental.cpp
@@ -1535,6 +1535,152 @@ struct SgToWiConvertLayout
   }
 };
 
+// Trivially distribute `vector.interleave`
+struct SgToWiVectorInterleave
+    : public OpConversionPattern<vector::InterleaveOp> {
+  using OpConversionPattern<vector::InterleaveOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(vector::InterleaveOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    auto newOp = vector::InterleaveOp::create(
+        rewriter, op.getLoc(), adaptor.getLhs(), adaptor.getRhs());
+    rewriter.replaceOp(op, newOp.getResult());
+    return success();
+  }
+};
+
+// Trivially distribute `vector.deinterleave`
+struct SgToWiVectorDeinterleave
+    : public OpConversionPattern<vector::DeinterleaveOp> {
+  using OpConversionPattern<vector::DeinterleaveOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(vector::DeinterleaveOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    auto newOp = vector::DeinterleaveOp::create(rewriter, op.getLoc(),
+                                                adaptor.getSource());
+    rewriter.replaceOp(op, newOp.getResults());
+    return success();
+  }
+};
+
+struct SgToWiDpasMx : public OpConversionPattern<xegpu::DpasMxOp> {
+  using OpConversionPattern<xegpu::DpasMxOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(xegpu::DpasMxOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // Check if the op has A, B and CD layouts attached.
+    auto layoutA = cast<xegpu::LayoutAttr>(op.getLayoutAAttr());
+    auto layoutB = cast<xegpu::LayoutAttr>(op.getLayoutBAttr());
+    auto layoutCd = cast<xegpu::LayoutAttr>(op.getLayoutCdAttr());
+    if (!layoutA || !layoutB || !layoutCd)
+      return rewriter.notifyMatchFailure(
+          op, "missing required layout attributes for DpasMxOp distribution");
+    auto layoutScaleA = cast<xegpu::LayoutAttr>(op.getLayoutAScaleAttr());
+    auto layoutScaleB = cast<xegpu::LayoutAttr>(op.getLayoutBScaleAttr());
+
+    // Retrieve expected types, according to anchor layouts.
+    auto expected1DTypeResult =
+        xegpu::getDistributedVectorType(op.getType(), layoutCd);
+    auto expected1DTypeA =
+        xegpu::getDistributedVectorType(op.getA().getType(), layoutA);
+    auto expected1DTypeB =
+        xegpu::getDistributedVectorType(op.getB().getType(), layoutB);
+
+    VectorType expected1DTypeScaleA, expected1DTypeScaleB;
+    if (op.getScaleA()) {
+      auto expected1DTypeScaleAOrFailure = xegpu::getDistributedVectorType(
+          cast<VectorType>(op.getScaleA().getType()), layoutScaleA);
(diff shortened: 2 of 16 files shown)


Garra1980:
cc @silee2

mlir/lib/Dialect/XeGPU/Transforms/XeGPUSgToWiDistributeExperimental.cpp
```diff
@@ -1535,6 +1535,152 @@ struct SgToWiConvertLayout
...
+  }
+};
+
+struct SgToWiDpasMx : public OpConversionPattern<xegpu::DpasMxOp> {
+  using OpConversionPattern<xegpu::DpasMxOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(xegpu::DpasMxOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // Check if the op has A, B and CD layouts attached.
```
silee2, line 1583: I think doing a uArch check upfront would be better than doing it later:
dpas_mx support is tied to SubgroupScaledMatrixMultiplyAcc
Maybe a check like?
```
    const uArch *uArch = getUArch(xegpu::getChipStr(op).value_or(""));
     if (!uArch)
         return failure();
     if (!uArch->isSupportedInstruction(SubgroupScaledMatrixMultiplyAcc))
         return failure();
```

mlir/lib/Dialect/XeGPU/Transforms/XeGPUSgToWiDistributeExperimental.cpp
```diff
@@ -1535,6 +1535,152 @@ struct SgToWiConvertLayout
...
+    auto layoutA = cast<xegpu::LayoutAttr>(op.getLayoutAAttr());
+    auto layoutB = cast<xegpu::LayoutAttr>(op.getLayoutBAttr());
+    auto layoutCd = cast<xegpu::LayoutAttr>(op.getLayoutCdAttr());
+    if (!layoutA || !layoutB || !layoutCd)
+      return rewriter.notifyMatchFailure(
+          op, "missing required layout attributes for DpasMxOp distribution");
+    auto layoutScaleA = cast<xegpu::LayoutAttr>(op.getLayoutAScaleAttr());
+    auto layoutScaleB = cast<xegpu::LayoutAttr>(op.getLayoutBScaleAttr());
```
silee2: These values are only used if op.getScaleA() and op.getScaleB() is not empty.
Better to move it inside `if` the statements below.
And should add check if the values are missing.

silee2:
Since this PR updates dpas_mx op print format,
dont' forget to update `mlir/test/Integration/Dialect/XeGPU/WG/simple_mxfp_gemm.mlir`
That test is currently marked as XFAIL so you won't notice the parser failure with the print format update.

mlir/lib/Dialect/XeGPU/Transforms/XeGPUSgToWiDistributeExperimental.cpp
akroviakov: Done

mlir/lib/Dialect/XeGPU/Transforms/XeGPUSgToWiDistributeExperimental.cpp
akroviakov: Done

silee2 (approved):

mlir/lib/Dialect/GPU/Pipelines/GPUToXeVMPipeline.cpp
```diff
@@ -96,6 +96,8 @@ void buildGPUPassPipeline(OpPassManager &pm,
     pm.addNestedPass<gpu::GPUModuleOp>(createLoopInvariantCodeMotionPass());
     pm.addNestedPass<gpu::GPUModuleOp>(createCSEPass());
     pm.addNestedPass<gpu::GPUModuleOp>(xegpu::createXeGPUVectorLinearize());
+    pm.addNestedPass<gpu::GPUModuleOp>(createCanonicalizerPass());
```
Garra1980, line 99: do we really need so many CSE and canonicalizer passes here - one after SG to WI distribution, one after linearize. Can we combine those perhaps and call in just one place?

mlir/test/Dialect/XeGPU/sg-to-wi-experimental-unit.mlir
```diff
@@ -1244,3 +1245,78 @@ gpu.func @vector_multi_reduction_1d_to_scalar() {
...
+  %A = arith.constant dense<3.> : vector<8x16xf8E4M3FN>
+  %B = arith.constant dense<3.> : vector<16x16xf8E4M3FN>
+  %scale_A = arith.constant dense<1.> : vector<8x2xf8E8M0FNU>
+  %scale_B = arith.constant dense<0.5> : vector<2x16xf8E8M0FNU>
+
+  %4 = xegpu.dpas_mx %A, %B scale_a = %scale_A scale_b = %scale_B
+  {layout_a = #xegpu.layout<lane_layout = [1, 16], lane_data = [1, 1]>,
+    layout_b = #xegpu.layout<lane_layout = [1, 16], lane_data = [1, 1]>,
+    layout_cd = #xegpu.layout<lane_layout = [1, 16], lane_data = [1, 1]>,
+    layout_a_scale = #xegpu.layout<lane_layout = [1, 2], lane_data = [1, 1]>,
```
Jianhui-Li: The layout_a_scale should be [8, 1]?

mlir/lib/Dialect/XeGPU/Transforms/XeGPUSgToWiDistributeExperimental.cpp
```diff
@@ -1535,6 +1535,153 @@ struct SgToWiConvertLayout
...
+
+    auto expectedNDTypeResult =
+        xegpu::getDistVecTypeBasedOnLaneLayout(layoutCd, op.getType());
+    if (failed(expected1DTypeResult) || failed(expected1DTypeA) ||
+        failed(expected1DTypeB))
+      return rewriter.notifyMatchFailure(
+          op,
+          "failed to calculate supported workitem 1D vector types for DpasOp "
+          "from layouts");
+    if (failed(expectedNDTypeResult))
```
Jianhui-Li, line 1627: why we need this check?  Isn't it the same as expected1DTypeResult?

mlir/lib/Dialect/XeGPU/Transforms/XeGPUSgToWiDistributeExperimental.cpp
```diff
@@ -1535,6 +1535,153 @@ struct SgToWiConvertLayout
...
+      auto expected1DTypeScaleAOrFailure = xegpu::getDistributedVectorType(
+          cast<VectorType>(op.getScaleA().getType()), layoutScaleA);
+      if (failed(expected1DTypeScaleAOrFailure))
+        return rewriter.notifyMatchFailure(
+            op, "failed to calculate expected 1D vector type for scale A");
+      expected1DTypeScaleA = expected1DTypeScaleAOrFailure.value();
+    }
+    if (op.getScaleB()) {
+      auto layoutScaleB = cast<xegpu::LayoutAttr>(op.getLayoutBScaleAttr());
+      auto expected1DTypeScaleBOrFailure = xegpu::getDistributedVectorType(
```
Jianhui-Li, line 1611: Can we use getDistVecTypeBasedOnLaneLayout() here instead of getDistributedVectorType()? I think it is better to restrict the usage of getDistributedVectorType() just for TensorDescriptor type.

mlir/test/Dialect/XeGPU/sg-to-wi-experimental-unit.mlir
akroviakov: Fixed

mlir/lib/Dialect/XeGPU/Transforms/XeGPUSgToWiDistributeExperimental.cpp
akroviakov: `getDistributedVectorType()` is aligned with what a/b/c operands use and what the non-mx dpas distribution uses. I suppose this is an issue of a broader nD `getDistVecTypeBasedOnLaneLayout()` support in lane-level XeGPU. I raised this question a week ago with @charithaintc and I was told this is an effort separate from this PR.

mlir/lib/Dialect/XeGPU/Transforms/XeGPUSgToWiDistributeExperimental.cpp
akroviakov: Follows the existing dpas distribution. A separate PR could outline the distribution of both and cleanup the `getDistVecTypeBasedOnLaneLayout()` vs `getDistributedVectorType()` usage

Jianhui-Li (approved):
LGTM

mlir/lib/Dialect/GPU/Pipelines/GPUToXeVMPipeline.cpp
akroviakov: The new ones are needed functionally. The `cse` and `canonicalize` before `LoopInvariantCodeMotionPass` can eliminate multiple thousands lines of code in certain examples. Not sure how much of a performance hit it is, though

PR #197224, opened 2026-05-12
[TableGen] Fix assertion when initializing a bit from !cond or !foldl

### LINKED ISSUE #197217: "Illegal VarBitInit expression!" assertion failed with !cond
With this example:
```
class C<bit val> {
  bits<32> word;
  bit flag;
  let word{0} = !cond(val : flag);
}
class D<bit val> : C<val>;
```
I get:
```
$ llvm-tblgen crash.td
llvm-tblgen: include/llvm/TableGen/Record.h:1268: llvm::VarBitInit::VarBitInit(const TypedInit *, unsigned int): Assertion `T->getType() && (isa<IntRecTy>(T->getType()) || (isa<BitsRecTy>(T->getType()) && cast<BitsRecTy>(T->getType())->getNumBits() > B)) && "Illegal VarBitInit expression!"' failed.
```
(This test case was originally added to #84910, but the fix for that bug did not fix it.)

jayfoad:
TableGen hit an assertion when trying to initialize a bit in a bits
field from an unfolded !cond expression like this:

    let word{0} = !cond(val : flag);

Fixed by changing CondOpInit::getBits to check for the case that it
already has a bit type, matching the pattern used in OpInit::getBit and
elsewhere.

Fixes #197217


jayfoad:
Claude Code helped to write the patch.

lenary:
Do we need to do this for:
- `FoldOpInit::getBit`
- `IsAOpInit::getBit`
- `ExistsOpInit::getBit`
- `InstancesOpInit::getBit`

Which all have implementations like the original implementation of `CondOpInit::getBit`?

llvm/lib/TableGen/Record.cpp
```diff
@@ -2732,6 +2732,8 @@ std::string CondOpInit::getAsString() const {
 }
 
 const Init *CondOpInit::getBit(unsigned Bit) const {
+  if (getType() == BitRecTy::get(getRecordKeeper()))
```
wangpc-pp: You may use `isa<BitRecTy>`?

jayfoad: #197450 does this for existing code.

jayfoad: Done

jayfoad:
> Do we need to do this for:
> 
> * `FoldOpInit::getBit`
> * `IsAOpInit::getBit`
> * `ExistsOpInit::getBit`
> * `InstancesOpInit::getBit`
> 
> Which all have implementations like the original implementation of `CondOpInit::getBit`?

Good catch. Yes at least some of these can provoke the same assertion failure. I'll work on them.

jayfoad:
> > Do we need to do this for:
> > 
> > * `FoldOpInit::getBit`
> > * `IsAOpInit::getBit`
> > * `ExistsOpInit::getBit`
> > * `InstancesOpInit::getBit`
> > 
> > Which all have implementations like the original implementation of `CondOpInit::getBit`?
> 
> Good catch. Yes at least some of these can provoke the same assertion failure. I'll work on them.

`IsAOpInit` and `ExistsOpInit` always have type `int` so they don't need the check. `InstancesOpInit` always has a list type so it doesn't need the check. Only `FoldOpInit` can have `bit` type.

lenary (approved):
LGTM

lenary:
> > > Do we need to do this for:
> > > 
> > > * `FoldOpInit::getBit`
> > > * `IsAOpInit::getBit`
> > > * `ExistsOpInit::getBit`
> > > * `InstancesOpInit::getBit`
> > > 
> > > Which all have implementations like the original implementation of `CondOpInit::getBit`?
> > 
> > 
> > Good catch. Yes at least some of these can provoke the same assertion failure. I'll work on them.
> 
> `IsAOpInit` and `ExistsOpInit` always have type `int` so they don't need the check. `InstancesOpInit` always has a list type so it doesn't need the check. Only `FoldOpInit` can have `bit` type.

Sorry, I did a LGTM, but I kept looking into what `getBit` is used for, and how it is implemented. Some of the `getBit` implementations use `llvm_unreachable` when the type is not of int/bit type - which I think applies to `InstancesOpInit`.  See for instance `ListInit`.

jayfoad:
> > > > Do we need to do this for:
> > > > 
> > > > * `FoldOpInit::getBit`
> > > > * `IsAOpInit::getBit`
> > > > * `ExistsOpInit::getBit`
> > > > * `InstancesOpInit::getBit`
> > > > 
> > > > Which all have implementations like the original implementation of `CondOpInit::getBit`?
> > > 
> > > 
> > > Good catch. Yes at least some of these can provoke the same assertion failure. I'll work on them.
> > 
> > 
> > `IsAOpInit` and `ExistsOpInit` always have type `int` so they don't need the check. `InstancesOpInit` always has a list type so it doesn't need the check. Only `FoldOpInit` can have `bit` type.
> 
> Sorry, I did a LGTM, but I kept looking into what `getBit` is used for, and how it is implemented. Some of the `getBit` implementations use `llvm_unreachable` when the type is not of int/bit type - which I think applies to `InstancesOpInit`. See for instance `ListInit`.

Agreed. Done.

lenary (approved):
LGTM

jayfoad:
> Congratulations on having your first Pull Request (PR) merged into the LLVM Project!

Thanks! Feeling so proud of myself.

PR #197228, opened 2026-05-12
[ARM] Reject unencodable Thumb2 LDRD/STRD post-index offsets

MarkMurrayARM:
Thumb2 post-index LDRD/STRD immediates are encoded as scaled-by-4
values. Without the Imm8s4 parser class, plain assembly can accept byte
offsets that cannot be represented exactly.

Reject those operands during asm matching so invalid input does not
reach later MC paths where debug builds assert and release builds can
silently encode the rounded-down value.

The t2am_imm8s4_offset_asmoperand parser is updated to accommodate
constructs such as #-0 which preserve the negative sentinel on the zero.


llvm/lib/Target/ARM/AsmParser/ARMAsmParser.cpp
```diff
@@ -1179,7 +1179,8 @@ class ARMOperand : public MCParsedAsmOperand {
     const MCConstantExpr *CE = dyn_cast<MCConstantExpr>(getImm());
     if (!CE) return false;
     int64_t Value = CE->getValue();
-    return ((Value & 3) == 0) && Value >= N && Value <= M;
+    return (((Value & 3) == 0) && Value >= N && Value <= M) ||
+           Value == std::numeric_limits<int32_t>::min();
```
davemgreen, line 1187: Can you add a comment that std::numeric_limits<int32_t>::min() is the encoding to -0? It was quite surprising to read.

davemgreen (approved):
Thanks LGTM.

PR #197230, opened 2026-05-12
[MLIR] Add getter for the action handler of MLIR context

akigyosi:
Current implementation of MLIRContext's action handling, requires the user to hold the stored memory for any observer or breakpoint manager added to the execution context. Using a getter for the registered action handler, permits the user to store the observers and breakpoints into MLIRContext and modify their state later, by retrieving the action handler and invoking the functor's target. Mainly for attaching new observers later in the compilation pipeline, after the execution context got registered.


mlir/include/mlir/IR/MLIRContext.h
joker-eph: Why not return by const ref?


```suggestion
  const HandlerTy &getActionHandler();
```

mlir/unittests/Debug/ActionHandlerTest.cpp
andrey-golubev: i suggest to use "the API" that would be normally called during pass execution, etc.:

```cpp
    ctx.executeAction<DummyAction>(workFn);
```

mlir/include/mlir/IR/MLIRContext.h
andrey-golubev: +1. the user can copy the object if desired.

andrey-golubev (commented):
Please do not forget to add description. I'd also update the title to something like "Add getter for an action handler of MLIR context".

Then we need an approval from Mehdi (at least?) to merge it.

mlir/unittests/Debug/ActionHandlerTest.cpp
joker-eph: Honestly I would skip adding an entire new C++ unit-test just for exposing a getter.

andrey-golubev: we were thinking that a test can be:

a) protection against deletion - someone decides to delete the getter, they'd need to delete the test, but they'd probably look at it first and see that downstream project uses the getter (can be stated in the test)
b) if handler's type changes (e.g. someone decides to switch from std::function to a class), these tests could be the first place to see how the new API would work (from set / get / implicit use perspectives)

but i don't have a strong preference overall personally

andrey-golubev (approved):
Please add the PR description! :)

mlir/unittests/Debug/ActionHandlerTest.cpp
joker-eph: I understand that this can be the general point of tests, but this still seems overkill: C++ unit-tests are the exception already, and a whole new TU to build for a getter is unprecedented.

For example for 1) I would not keep an API because there is a test, in general "downstream" isn't my main concern for upstream APIs and a test wouldn't guarantee anything.

joker-eph:
The MR is still "Draft" by the way

joker-eph (approved):

PR #197234, opened 2026-05-12
[llvm][tools] Use temp dir for offload-binary unbundling test

boomanaiden154:
Certain environments will leave some of the test dirs read-only for immutability purposes. Create a new temporary directory so that llvm-offload-binary has a writable directory to unbundle the image into. With this method we can also delete the temporary directory preventing breakage of the failure from still passing tests due to leftover files.

### DIFF
--- a/llvm/test/tools/llvm-offload-binary/llvm-offload-binary.ll
+++ b/llvm/test/tools/llvm-offload-binary/llvm-offload-binary.ll
@@ -17,6 +17,8 @@
 ; RUN: diff %s %t4
 
 ; Test extracting all images without specifying --image filters.
+; RUN: rm -rf %.dir && mkdir %t.dir
+; RUN: cd %t.dir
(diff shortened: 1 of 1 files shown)


llvm/test/tools/llvm-offload-binary/llvm-offload-binary.ll
```diff
@@ -17,6 +17,8 @@
 ; RUN: diff %s %t4
 
 ; Test extracting all images without specifying --image filters.
+; RUN: rm -rf %.dir && mkdir %t.dir
```
pranavk: %t ?

boomanaiden154: Yes. Fixed.

sarnex (approved):

pranavk (approved):

jhuber6 (approved):

adurang (approved):
thanks for fixing this!

PR #197235, opened 2026-05-12
[libclc] Apply hidden visibility to amdgpu / nvptx builds

jhuber6:
Summary:
    This is not currently used because we force `--internalize` for
    llvm-link, but if you ever want to link this normally we'd need hidden
    visibility. SPIR-V does not currently handle hidden visibility, as it is
    an extension still under discussion, so it is omitted for now.

### DIFF
--- a/libclc/CMakeLists.txt
+++ b/libclc/CMakeLists.txt
@@ -23,9 +23,10 @@ option(
 )
 
 # List of all supported architectures.
-set( LIBCLC_ARCHS_ALL amdgpu amdgcn nvptx64 )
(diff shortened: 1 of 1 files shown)


arsenm (commented):
Shouldn't clang set this as the default?

jhuber6:
> Shouldn't clang set this as the default?

Not for NVPTX and I figured it'd be better to be explicit about it.

arsenm (approved):

wenju-he (approved):
LGTM

PR #91961, opened 2024-05-13
[SCEV] Introduce SCEVUse, use it instead of const SCEV * (NFCI).

fhahn:
This patch introduces SCEVUse, which is a tagged pointer containing the
used const SCEV *, plus extra bits to store NUW/NSW flags that are only
valid at the specific use.

This was suggested by @nikic as an alternative
to https://github.com/llvm/llvm-project/pull/90742.

This patch just updates most SCEV infrastructure to operate on SCEVUse
instead of const SCEV *. It does not introduce any code that makes use
of the use-specific flags yet which I'll share as follow-ups.

Note that this should be NFC, but currently there's at least one case
where it is not (turn-to-invariant.ll), which I'll investigate once we
agree on the overall direction.

This PR at the moment also contains a commit that updates various SCEV
clients to use `const SCEV *` instead of `const auto *`, to prepare for
this patch. This reduces the number of changes needed, as SCEVUse will
automatically convert to `const SCEV *`. This is a safe default, as it
just drops the use-specific flags for the expression (it will not drop
any use-specific flags for any of its operands though).

SCEVUse could probably also be used to address mis-compiles due to
equivalent AddRecs modulo flags result in an AddRec with incorrect flags
for some uses of some phis, e.g. the one
https://github.com/llvm/llvm-project/pull/80430 attempted to fix

Compile-time impact:
stage1-O3: +0.06%
stage1-ReleaseThinLTO: +0.07%
stage1-ReleaseLTO-g: +0.07%
stage2-O3: +0.11%

https://llvm-compile-time-tracker.com/compare.php?from=ce055843e2be9643bd58764783a7bb69f6db8c9a&to=8c7f4e9e154ebc4862c4e2716cedc3c688352d7c&stat=instructions:u

PR #197236, opened 2026-05-12
[SCEV] Propagate nuw/nsw flags from or disjoint

vmustya:
The Scalar Evolution analysis converts the `or disjoint` instruction
into a synthetic add with the `IsNSW` and `IsNUW` flags set, but the
real instruction reference set to `nullptr`. The flags propagation logic
only reads the real instruction flags, when the instruction reference is
present. This behavior silently ignores the flags for the synthetic add,
which can lead to missed optimization opportunities.

This patch inserts the pointer to the actual instruction into the
`BinaryOp` object created for the synthetic add. It also adjust the
flags handling logic to threat the `disjoint` flag as both `nuw` and
`nsw` as if they were set on the `add` instruction.

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/llvm/lib/Analysis/ScalarEvolution.cpp
+++ b/llvm/lib/Analysis/ScalarEvolution.cpp
@@ -5418,9 +5418,14 @@ static std::optional<BinaryOp> MatchBinaryOp(Value *V, const DataLayout &DL,
 
   case Instruction::Or: {
     // Convert or disjoint into add nuw nsw.
-    if (cast<PossiblyDisjointInst>(Op)->isDisjoint())
-      return BinaryOp(Instruction::Add, Op->getOperand(0), Op->getOperand(1),
-                      /*IsNSW=*/true, /*IsNUW=*/true);
+    if (cast<PossiblyDisjointInst>(Op)->isDisjoint()) {
+      BinaryOp BinOp(Instruction::Add, Op->getOperand(0), Op->getOperand(1),
+                     /*IsNSW=*/true, /*IsNUW=*/true);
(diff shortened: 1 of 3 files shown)


efriedma-quic (dismissed):
This is not a valid transform.  SCEV nsw/nuw flags need to be valid in all contexts.  (See also #91961 etc.)

vmustya:
@efriedma-quic, could you please clarify, why the transformation isn't valid?

The existing code creates a `BinaryOp` object fron the `or disjoint` instruction, and keeps the `BinaryOp::Op` initialized to the null pointer:
https://github.com/llvm/llvm-project/blob/033c9e65cdef46a374c3622633777fbf0f7ad35b/llvm/lib/Analysis/ScalarEvolution.cpp#L5419-L5426

So, the `IsNSW` and `IsNUW` flags are semantically valid, and can be used in the patch context. Do I miss anything?

nikic:
You can use the flags, but you need to prove that a poison result implies UB. The preceding code path uses getNoWrapFlagsFromUB(), which calls isSCEVExprNeverPoison(). That's what you need here.

vmustya:
I've updated the patch. Now, instead of the special handling, it inserts the `or disjoint` instruction pointer to the `BinaryOp` object. Also, the `getNoWrapFlagsFromUB` handles the `disjoint` flag as both `nsw` and `nuw`.

@nikic, @efriedma-quic, could you please review again? Thanks in advance!

nikic (approved):
LGTM

PR #197237, opened 2026-05-12
[lldb] Remove .noindex suffix from test output directory for non-Darwin

speednoisemovement:
.noindex is only needed for macOS Spotlight. Given that some platforms limit path length, there's a benefit to omitting it except for Darwin.

### DIFF
--- a/lldb/packages/Python/lldbsuite/test/dotest_args.py
+++ b/lldb/packages/Python/lldbsuite/test/dotest_args.py
@@ -235,7 +235,7 @@ def create_parser():
         dest="test_build_dir",
         metavar="Test build directory",
         default=(
-            "lldb-test-build" if sys.platform == "win32" else "lldb-test-build.noindex"
+            "lldb-test-build.noindex" if sys.platform == "darwin" else "lldb-test-build"
         ),
         help="The root build directory for the tests. It will be removed before running.",
     )
--- a/lldb/packages/Python/lldbsuite/test/dotest_args.py
+++ b/lldb/packages/Python/lldbsuite/test/dotest_args.py
@@ -234,7 +234,9 @@ def create_parser():
         "--build-dir",
         dest="test_build_dir",
         metavar="Test build directory",
-        default="lldb-test-build.noindex",
+        default=(
+            "lldb-test-build" if sys.platform == "win32" else "lldb-test-build.noindex"
+        ),
         help="The root build directory for the tests. It will be removed before running.",
     )
     group.add_argument(
--- a/lldb/test/API/sanity/TestModuleCacheSanity.py
+++ b/lldb/test/API/sanity/TestModuleCacheSanity.py
@@ -16,9 +16,7 @@ class ModuleCacheSanityTestCase(TestBase):
 
     def test(self):
         build_dir_name = (
-            "lldb-test-build"
-            if sys.platform == "win32"
-            else "lldb-test-build.noindex"
+            "lldb-test-build" if sys.platform == "win32" else "lldb-test-build.noindex"
         )
         self.expect(
             "settings show symbols.clang-modules-cache-path",
--- a/lldb/test/API/sanity/TestModuleCacheSanity.py
+++ b/lldb/test/API/sanity/TestModuleCacheSanity.py
@@ -16,7 +16,7 @@ class ModuleCacheSanityTestCase(TestBase):
 
     def test(self):
         build_dir_name = (
-            "lldb-test-build" if sys.platform == "win32" else "lldb-test-build.noindex"
+            "lldb-test-build.noindex" if sys.platform == "darwin" else "lldb-test-build"
         )
         self.expect(
             "settings show symbols.clang-modules-cache-path",
--- a/lldb/test/API/sanity/TestModuleCacheSanity.py
+++ b/lldb/test/API/sanity/TestModuleCacheSanity.py
@@ -4,6 +4,8 @@ correctly and points inside the default test build directory.
 """
 
 
+import sys
+
 import lldb
 import lldbsuite.test.lldbutil as lldbutil
 from lldbsuite.test.lldbtest import *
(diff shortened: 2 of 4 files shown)


lldb/packages/Python/lldbsuite/test/dotest_args.py
```diff
@@ -234,7 +234,9 @@ def create_parser():
         "--build-dir",
         dest="test_build_dir",
         metavar="Test build directory",
-        default="lldb-test-build.noindex",
+        default=(
+            "lldb-test-build" if sys.platform == "win32" else "lldb-test-build.noindex"
```
adrian-prantl:
```suggestion
            "lldb-test-build" if sys.platform != "darwin" else "lldb-test-build.noindex"
```

adrian-prantl: The no index suffix really only exists for macOS Spotlight, so we could make the default on all non-Darwin platforms.

speednoisemovement: Updated

JDevlieghere (commented):
This seems like a lot of complexity for saving 8 characters. Can we save them some other way, for example by trimming `lldb-test-build` to `test.noindex` or something?

speednoisemovement:
> This seems like a lot of complexity for saving 8 characters. Can we save them some other way, for example by trimming `lldb-test-build` to `test.noindex` or something?

This was definitely motivated by saving characters, but I think it makes sense on its own since the `.noindex` is only meaningful on Darwin, and the complexity is limited to two ternaries (plus test changes). Probably shortening elsewhere is also good, but the only thing special about this path is the `.noindex`

lldb/test/API/sanity/TestModuleCacheSanity.py
```diff
@@ -13,7 +15,10 @@ class ModuleCacheSanityTestCase(TestBase):
     NO_DEBUG_INFO_TESTCASE = True
 
     def test(self):
+        build_dir_name = (
+            "lldb-test-build.noindex" if sys.platform == "darwin" else "lldb-test-build"
+        )
         self.expect(
             "settings show symbols.clang-modules-cache-path",
-            substrs=["lldb-test-build.noindex", "module-cache-lldb"],
+            substrs=[build_dir_name, "module-cache-lldb"],
```
kastiglione: I think it would be fine to simplify this to:

```py
substrs=["lldb-test-build", "module-cache-lldb"]
```

This works whether the name is "lldb-test-build.noindex" or "lldb-test-build", and makes this file a 8 character change.

lldb/test/Shell/Settings/TestModuleCacheSanity.test
```diff
@@ -1,4 +1,4 @@
 # This is a sanity check that verifies that the module cache path is set
 # correctly and points inside the default test build directory.
 RUN: %lldb -o 'settings show symbols.clang-modules-cache-path' | FileCheck  %s
-CHECK: lldb-test-build.noindex{{.*}}module-cache-lldb
+CHECK: lldb-test-build{{(\.noindex)?}}{{.*}}module-cache-lldb
```
kastiglione: this could also be simplified

```
CHECK: lldb-test-build{{.*}}module-cache-lldb
```

lldb/test/CMakeLists.txt
```diff
@@ -42,12 +42,16 @@ if(LLDB_BUILT_STANDALONE)
...
+if(APPLE)
+  # The .noindex suffix is a marker for Spotlight to never index the
+  # build directory.  LLDB queries Spotlight to locate .dSYM bundles
+  # based on the UUID embedded in a binary, and because the UUID is a
+  # hash of filename and .text section, there *will* be conflicts inside
+  # the build directory.
+  set(LLDB_TEST_BUILD_DIRECTORY "${PROJECT_BINARY_DIR}/lldb-test-build.noindex" CACHE PATH "The build root for building tests.")
+else()
+  set(LLDB_TEST_BUILD_DIRECTORY "${PROJECT_BINARY_DIR}/lldb-test-build" CACHE PATH "The build root for building tests.")
+endif()
```
kastiglione:
```
if(APPLE)
  # The .noindex suffix is a marker for Spotlight to never index the
  # build directory.  LLDB queries Spotlight to locate .dSYM bundles
  # based on the UUID embedded in a binary, and because the UUID is a
  # hash of filename and .text section, there *will* be conflicts inside
  # the build directory.
  set(LLDB_TEST_BUILD_DIR_SUFFIX ".noindex") 
endif()
set(LLDB_TEST_BUILD_DIRECTORY "${PROJECT_BINARY_DIR}/lldb-test-build$(LLDB_TEST_BUILD_DIR_SUFFIX)" CACHE PATH "The build root for building tests.")
```

lldb/test/CMakeLists.txt
```diff
@@ -42,12 +42,16 @@ if(LLDB_BUILT_STANDALONE)
...
-# based on the UUID embedded in a binary, and because the UUID is a
-# hash of filename and .text section, there *will* be conflicts inside
-# the build directory.
-set(LLDB_TEST_BUILD_DIRECTORY "${PROJECT_BINARY_DIR}/lldb-test-build.noindex" CACHE PATH "The build root for building tests.")
+if(APPLE)
+  # The .noindex suffix is a marker for Spotlight to never index the
+  # build directory.  LLDB queries Spotlight to locate .dSYM bundles
+  # based on the UUID embedded in a binary, and because the UUID is a
+  # hash of filename and .text section, there *will* be conflicts inside
+  # the build directory.
```
kastiglione, line 50: @adrian-prantl @JDevlieghere What if, instead of using a .noindex suffix to solve the problem described in this comment, we instead add `-random_uuid` as a linker flag on macOS/darwin?

kastiglione: for consideration https://github.com/llvm/llvm-project/pull/199385

bulbazord: Avoid dSYM conflicts is definitely one reason to leave Spotlight out of the picture, but do we want Spotlight indexing the test binaries by default anyway? Seems like a waste of compute for something so ephemeral.

adrian-prantl: Yeah it would unnecessarily use extra OS and compute resources if we let spotlight index them. And potentially take up space in the Spotlight database that I don't know when it will be reclaimed.

adrian-prantl (approved):

lldb/test/CMakeLists.txt
kastiglione: I use "build.noindex" as the name of my top level build directory for these reasons. I don't want and don't spotlight to concern itself with anything in the build.

lldb/test/API/sanity/TestModuleCacheSanity.py
speednoisemovement: Thanks! This and below are done

PR #197238, opened 2026-05-12
[AArch64][GlobalISel] Add always legal action builders.

davemgreen:
This defined some always legal actions, removing our dependency on the Legacy ruleset in aarch64.


llvm/lib/Target/AArch64/GISel/AArch64LegalizerInfo.cpp
```diff
@@ -1509,6 +1509,11 @@ AArch64LegalizerInfo::AArch64LegalizerInfo(const AArch64Subtarget &ST)
 
   getActionDefinitionsBuilder({G_SCMP, G_UCMP}).lower();
 
+  getActionDefinitionsBuilder({G_INTRINSIC, G_INTRINSIC_W_SIDE_EFFECTS})
+      .alwaysLegal();
```
arsenm, line 1513: I don't think this is necessary, amdgpu doesn't have it

davemgreen: AMDGPU will need it too if we remove the legacy rules (and don't special case it). It is special cased for actual legalisation (through the LI.legalizeIntrinsic call), but the intrinsics still need to be Legal for queries to prevent errors.

AMDGPU is just missing G_BR too. G_TRUNC seems to be a common one to be relying on legacy rules.

llvm/lib/Target/AArch64/GISel/AArch64LegalizerInfo.cpp
```diff
@@ -1509,6 +1509,11 @@ AArch64LegalizerInfo::AArch64LegalizerInfo(const AArch64Subtarget &ST)
...
+  getActionDefinitionsBuilder(G_FENCE).alwaysLegal();
+  getActionDefinitionsBuilder(G_INVOKE_REGION_START).alwaysLegal();
```
arsenm, line 1515: These could all be one common getActionDefinitionsBuilder group call

arsenm (approved):

llvm/lib/Target/AArch64/GISel/AArch64LegalizerInfo.cpp
davemgreen: Yep - I was trying to keep them together in related groups.

aemerson (approved):

PR #197243, opened 2026-05-12
[LSV] Handle or-disjoint as an add-like operation for vectorization.

vmustya:
LoadStoreVectorizer can safely handle `or disjoint` instructions as
`add nuw nsw`, when they are used for offset computations. This patch
enables vectorization of such patterns.

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>


michalpaszkowski (approved):

llvm/test/Transforms/LoadStoreVectorizer/X86/or-disjoint-zext.ll
```diff
@@ -0,0 +1,105 @@
+; RUN: opt %s -mtriple=x86_64-unknown-linux-gnu -passes=load-store-vectorizer -S -o - | FileCheck %s
```
dakersnar, line 2: please auto generate checks for this test

vmustya: Done: https://github.com/llvm/llvm-project/pull/197243/commits/399c5143e13141a8d60338777935cb8073d3f16b

llvm/lib/Transforms/Vectorize/LoadStoreVectorizer.cpp
```diff
@@ -1360,11 +1360,30 @@ bool Vectorizer::isSafeToMove(
 }
 
 static bool checkNoWrapFlags(Instruction *I, bool Signed) {
+  // or disjoint is equivalent to add nuw nsw, so it never wraps.
+  if (auto *PDI = dyn_cast<PossiblyDisjointInst>(I); PDI && PDI->isDisjoint())
+    return true;
```
dakersnar, line 1365: Nit: can you add a test that hits this case? I think your existing test hits an earlier case within checkIfSafeAddSequence that doens't call checkNoWrapFlags.

dakersnar (approved):
LGTM, thanks for the change!

llvm/lib/Transforms/Vectorize/LoadStoreVectorizer.cpp
vmustya: Done: https://github.com/llvm/llvm-project/pull/197243/commits/3e5b4a6a5151ec3ea18984304adb5d589691a7ed

PR #176963, opened 2026-01-20
[AMDGPU][True16] Legalize S16 for stores & loads in GlobalISel when True16 is enabled

saxlungs:
Upstream enabling selection of the FLAT, GLOBAL, and SCRATCH True16 D16 load stores

PR #197245, opened 2026-05-12
[AMDGPU][True16] Add test for various s16 G_LOAD/G_STORE cases

saxlungs:
This test will be updated by https://github.com/llvm/llvm-project/pull/176963 and should show differences between True and Fake 16 in all cases. Currently it does not for some cases


llvm/test/CodeGen/AMDGPU/GlobalISel/legalize-load-store-s16-true16.mir
```diff
@@ -0,0 +1,481 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+# RUN: llc -mtriple=amdgcn-amd-amdhsa -mcpu=gfx1100 -mattr=+real-true16 -O0 -run-pass=legalizer -global-isel-abort=0 %s -o - | FileCheck -check-prefix=GFX11,GFX11-TRUE16 %s
+# RUN: llc -mtriple=amdgcn-amd-amdhsa -mcpu=gfx1100 -mattr=-real-true16 -O0 -run-pass=legalizer -global-isel-abort=0 %s -o - | FileCheck -check-prefix=GFX11,GFX11-FAKE16 %s
+# RUN: llc -mtriple=amdgcn-amd-amdhsa -mcpu=gfx1200 -mattr=+real-true16 -O0 -run-pass=legalizer -global-isel-abort=0 %s -o - | FileCheck -check-prefix=GFX12,GFX12-TRUE16 %s
+# RUN: llc -mtriple=amdgcn-amd-amdhsa -mcpu=gfx1200 -mattr=-real-true16 -O0 -run-pass=legalizer -global-isel-abort=0 %s -o - | FileCheck -check-prefix=GFX12,GFX12-FAKE16 %s
```
arsenm: Don't need -O0, it shouldn't really do anything when using -run-pass + legalizer

saxlungs: Removed `-O0` and I also removed the `-global-isel-abort=0` part because it didn't seem to affect it either

arsenm (approved):
End to end would be better

PR #197246, opened 2026-05-12
[PseudoProbe] Preserve pseudoprobe for tailcall pseudo instrs

mustartt:
Preserve probes for tail call pseudo instructions. 

On AArch64, X86, and other platforms, the lowering of tail calls requires a pseudo instruction like `TCRETURNdi`. It is often the case that `TCRETURNdi` is within its own MBB, and the direct call probe associated with that tail call is considered dangling and removed. This patch tries to preserve that.

### DIFF
--- a/llvm/lib/CodeGen/PseudoProbeInserter.cpp
+++ b/llvm/lib/CodeGen/PseudoProbeInserter.cpp
@@ -28,6 +28,14 @@
 using namespace llvm;
 
 namespace {
+
+// A real instruction is a non-meta, non-pseudo instruction. Calls (including
+// call pseudos like tail-call returns) are also treated as real because they
+// expand to a branch and any preceding call probe must be preserved.
+static bool isCallOrRealInstruction(const MachineInstr &MI) {
+  return MI.isCall() || (!MI.isPseudo() && !MI.isMetaInstruction());
+}
+
 class PseudoProbeInserter : public MachineFunctionPass {
 public:
   static char ID;
@@ -56,7 +64,7 @@ public:
       for (MachineInstr &MI : MBB) {
         // Pseudo instructions like TCRETURNdi results in a branch instruction
         // and the call probe for that tail call should be preserved.
-        if ((!MI.isPseudo() && !MI.isMetaInstruction()) || MI.isCall())
+        if (isCallOrRealInstruction(MI))
           FirstInstr = &MI;
         if (MI.isCall()) {
           if (DILocation *DL = MI.getDebugLoc()) {
@@ -92,8 +100,10 @@ public:
         auto MII = MBB.rbegin();
         while (MII != MBB.rend()) {
           // Skip all pseudo probes followed by a real instruction since they
-          // are not dangling.
-          if (!MII->isPseudo())
+          // are not dangling. Treat call pseudos (e.g. tail-call returns)
+          // as real instructions to keep this consistent with the forward
+          // scan above.
+          if (isCallOrRealInstruction(*MII))
             break;
           auto Cur = MII++;
           if (Cur->getOpcode() != TargetOpcode::PSEUDO_PROBE)
--- a/llvm/lib/CodeGen/PseudoProbeInserter.cpp
+++ b/llvm/lib/CodeGen/PseudoProbeInserter.cpp
@@ -54,7 +54,7 @@ public:
     for (MachineBasicBlock &MBB : MF) {
       MachineInstr *FirstInstr = nullptr;
       for (MachineInstr &MI : MBB) {
-        // Pseudo instructions like TCRETURNbi results in a branch instruction
+        // Pseudo instructions like TCRETURNdi results in a branch instruction
         // and the call probe for that tail call should be preserved.
         if ((!MI.isPseudo() && !MI.isMetaInstruction()) || MI.isCall())
           FirstInstr = &MI;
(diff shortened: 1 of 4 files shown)


aaupov (approved):
LGTM but let's give a couple of days to other reviewers

llvm/test/Transforms/SampleProfile/pseudo-probe-tailcall.ll
apolloww: Since this affects x86 as well, could you drop this or include x86 as well?

Alternatively, you can move the test to `CodeGen/AArch64` and then have a similar one for `X86`.

llvm/lib/CodeGen/PseudoProbeInserter.cpp
```diff
@@ -54,7 +54,9 @@ class PseudoProbeInserter : public MachineFunctionPass {
     for (MachineBasicBlock &MBB : MF) {
       MachineInstr *FirstInstr = nullptr;
       for (MachineInstr &MI : MBB) {
-        if (!MI.isPseudo())
+        // Pseudo instructions like TCRETURNbi results in a branch instruction
```
apolloww: nit: `TCRETURNdi` insead of `TCRETURNbi`

llvm/lib/CodeGen/PseudoProbeInserter.cpp
```diff
@@ -54,7 +54,9 @@ class PseudoProbeInserter : public MachineFunctionPass {
...
+        // and the call probe for that tail call should be preserved.
+        if ((!MI.isPseudo() && !MI.isMetaInstruction()) || MI.isCall())
```
apolloww: The PR tried to preserve probes for tail call, but there are other pseudo instructions also meet the condition of `isCall()`, such as `PATCHPOINT`, `STATEPOINT`, `STACKMAP`(this creates a label and NOPs), etc. Would that be an issue?

llvm/test/Transforms/SampleProfile/pseudo-probe-tailcall.ll
apolloww: nit: foo_tail

llvm/lib/CodeGen/PseudoProbeInserter.cpp
mustartt: I don't believe this is a problem. `PATCHPOINT, STATEPOINT, STACKMAP` creates NOPs, which results in an instruction in the final output, even if it's a NOP. 
This is arguably more correct, as the profiler can land on these ranges that contains NOPs and still get range samples out of the MBB, instead of having the pseudo probes removed. 

All the other pseudo instructions that are calls all lower to an actual call or NOPs.

llvm/lib/CodeGen/PseudoProbeInserter.cpp
```diff
@@ -54,7 +54,9 @@ class PseudoProbeInserter : public MachineFunctionPass {
     for (MachineBasicBlock &MBB : MF) {
       MachineInstr *FirstInstr = nullptr;
       for (MachineInstr &MI : MBB) {
-        if (!MI.isPseudo())
+        // Pseudo instructions like TCRETURNdi results in a branch instruction
+        // and the call probe for that tail call should be preserved.
+        if ((!MI.isPseudo() && !MI.isMetaInstruction()) || MI.isCall())
```
HighW4y2H3ll: Is this change only intended to handle `TCRETURNdi` pseudo instruction? Looks like they will just be replace to a branch instruction in the backend, so it's not really a call in the binary.
https://github.com/llvm/llvm-project/blob/849e5c5fc623bc8713d3a51e1747778faee51610/llvm/lib/Target/X86/X86InstrControl.td#L285
https://github.com/llvm/llvm-project/blob/ccc903863af2335e1ff7bb0dc65ebc1fc0ec6e57/llvm/lib/Target/ARM/ARMInstrInfo.td#L2805
https://github.com/llvm/llvm-project/blob/ccc903863af2335e1ff7bb0dc65ebc1fc0ec6e57/llvm/lib/Target/ARM/ARMExpandPseudoInsts.cpp#L2275-L2277

Could you share what issue you are running into that has to emit a call probe for tail call? because in the llvm-profgen, there are specifically code to handle tail calls and infer missing frames when generating CS profile:
https://reviews.llvm.org/D139367

tobias-stadler: We are losing the call target samples for these tail calls, even though we can extract this information from the LBRs. Without a probe on the tail call, there is no place to attribute the CalledTargetSamples to. This issue also affects the non-CS profiles.

mustartt:
Gentle ping

HighW4y2H3ll:
Sorry for getting back late. Just trying to be careful here since this change might be affecting a lot of places.. Looks like the check happens to mirror the `isCallOrRealInstruction` from X86AvoidTrailingCall.cpp which i think makes sense. https://github.com/llvm/llvm-project/blob/2d9407bb56b79645a6d6700d90e1626ae894b5ac/llvm/lib/Target/X86/X86AvoidTrailingCall.cpp#L77 but might be good to get a second opinion from the other reviewers since I'm not very familiar with the different pseudos and meta instructions..

It also looks like right now the order of block probe and call probe is reversed... Maybe you'd also want to update the condition here to be consistent?
https://github.com/llvm/llvm-project/blob/0df8d4730a29d76098db1b450bf68016ec261880/llvm/lib/CodeGen/PseudoProbeInserter.cpp#L94

mustartt:
> It also looks like right now the order of block probe and call probe is reversed... Maybe you'd also want to update the condition here to be consistent?
> 
> https://github.com/llvm/llvm-project/blob/0df8d4730a29d76098db1b450bf68016ec261880/llvm/lib/CodeGen/PseudoProbeInserter.cpp#L94

Thanks for taking a look @HighW4y2H3ll. I updated the second call site that also requires it, and the order looks right now.

HighW4y2H3ll (approved):

PR #197247, opened 2026-05-12
[libclc] Create a static `.a` file in addition to the `.bc` file

jhuber6:
Summary:
This changes the libraries to be object libraries instead of static
libraries. We can then just link these into final static libraries that
contain everything we need.

The desire here is that we'd need static libraries if we wanted to move
away from `-mlink-builtin-bitcode` appraoches.

Effectively we'll have `libclc.a` next to `libclc.bc` and the idea is that we could alternatively link it and let the target linker handle it.

### DIFF
--- a/libclc/cmake/modules/AddLibclc.cmake
+++ b/libclc/cmake/modules/AddLibclc.cmake
@@ -52,7 +52,7 @@ function(libclc_set_source_options option)
   )
 endfunction()
 
-# Creates a static library target for libclc builtins and configures its
+# Creates an object library target for libclc builtins and configures its
 # compile options, include directories, and definitions. Subdirectories
 # populate sources via libclc_add_sources() after this call.
 function(libclc_add_builtin_library target_name)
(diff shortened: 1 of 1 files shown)


arsenm (approved):

libclc/cmake/modules/AddLibclc.cmake
```diff
@@ -63,53 +63,76 @@ function(libclc_add_builtin_library target_name)
...
 
   set(library_dir ${LIBCLC_OUTPUT_LIBRARY_DIR}/${ARG_TARGET_TRIPLE})
   file(MAKE_DIRECTORY ${library_dir})
 
-  set(linked_bc ${CMAKE_CURRENT_BINARY_DIR}/${target_name}.linked.bc)
+  # Create a combined static archive from all object libraries for installation.
+  set(archive_target ${target_name}.a)
+  add_library(${archive_target} STATIC)
+  target_link_libraries(${archive_target} PRIVATE
+    ${ARG_LIBRARIES} ${ARG_INTERNALIZE_LIBRARIES}
```
wenju-he, line 100: create optimized .a library? This would avoid repeated intra-libclc optimization when compiling user code.
Could be addressed later.

wenju-he (approved):
LGTM

libclc/cmake/modules/AddLibclc.cmake
jhuber6: Like merging them into a single file? I prefer individual files because it lets the linker semantics do something similar to as-needed.

wenju-he: yes, it is like link+optimize. It looks like clang can do it after extracting objects from ${ARG_LIBRARIES} and link in ${ARG_INTERNALIZE_LIBRARIES}?

jhuber6: I'd prefer not, means the linker would need to pull in everything from the merged library (like 8 MiB) rather than just scanning the undefined functions. Though I have observed in practice this can make a performance difference when messing with the OpenMP GPU runtime. Though, I am not sure if the libclc functions are as complex or dependent on IPO as those are.

wenju-he: but we can still create .a library with bag of bitcodes, something like following steps:
1. llvm-ar extract bitcode from ${ARG_LIBRARIES}
2. for each bitcode, use clang to link in ${ARG_INTERNALIZE_LIBRARIES} and optimize
3. pack bitcodes into .a library.

PR #187623, opened 2026-03-20
[clangd] Add InsertReplaceEdit for code completion

argothiel:
Handle new insertReplaceSupport capability (defined in LSP 3.16). Add the new option to the protocol layer and pass it around to the code completion logic. Update CompletionItem::textEdit to become the union type as per the LSP specification.

Add a new helper function to the Lexer public API to find the end of an identifier with full context lexing, to avoid duplicating the logic. Use the helper both in the Sema flow and in the comment completion flow. Use a simpler ASCII-only scan in no-Sema mode.

Add LIT tests to verify auto-triggered completions, mid-word replacement, Unicode, and snippets. Add unit tests to verify insert/replace ranges with and without Sema, including comments and the feature-off case.

Update the release notes to document the new capability.

Fixes https://github.com/clangd/clangd/issues/2190

PR #197249, opened 2026-05-12
[clangd] Fix parens suppression in mid-identifier code-completion

argothiel:
When completing in the middle of an existing identifier (e.g. `fo^o<int>(42)`), the next-token check lexes the character immediately after the cursor, which prevents parens suppression to kick in.

After the fix, we go to the end of the current identifier first and only then we start lexing for the next token, which handles redundant parens even when the cursor is mid-identifier.

This also fixes the parens suppression in the replace mode which by design is used mid-identifier.

### DIFF
--- a/clang-tools-extra/clangd/CodeComplete.cpp
+++ b/clang-tools-extra/clangd/CodeComplete.cpp
@@ -1536,46 +1536,6 @@ FuzzyFindRequest speculativeFuzzyFindRequestForCompletion(
   return CachedReq;
 }
 
-// This function is similar to Lexer::findNextToken(), but assumes
-// that the input SourceLocation is the completion point (which is
-// a case findNextToken() does not handle).
-std::optional<Token>
-findTokenAfterCompletionPoint(SourceLocation CompletionPoint,
-                              const SourceManager &SM,
-                              const LangOptions &LangOpts) {
-  SourceLocation Loc = CompletionPoint;
-  if (Loc.isMacroID()) {
-    if (!Lexer::isAtEndOfMacroExpansion(Loc, SM, LangOpts, &Loc))
-      return std::nullopt;
-  }
-
-  // Advance to the next SourceLocation after the completion point.
-  // Lexer::findNextToken() would call MeasureTokenLength() here,
-  // which does not handle the completion point (and can't, because
-  // the Lexer instance it constructs internally doesn't have a
-  // Preprocessor and so doesn't know about the completion point).
-  Loc = Loc.getLocWithOffset(1);
-
-  // Break down the source location.
-  std::pair<FileID, unsigned> LocInfo = SM.getDecomposedLoc(Loc);
-
-  // Try to load the file buffer.
-  bool InvalidTemp = false;
-  StringRef File = SM.getBufferData(LocInfo.first, &InvalidTemp);
-  if (InvalidTemp)
-    return std::nullopt;
-
-  const char *TokenBegin = File.data() + LocInfo.second;
-
-  // Lex from the start of the given location.
-  Lexer TheLexer(SM.getLocForStartOfFile(LocInfo.first), LangOpts, File.begin(),
-                 TokenBegin, File.end());
-  // Find the token.
-  Token Tok;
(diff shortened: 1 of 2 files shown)


timon-ul (commented):
First of all, this looks already really good and I am almost happy to merge this, that being said I have some things that I want to discuss beforehand

1. As fun as smileys are, I just learned that they are not allowed UTF8 in C++ (anymore). For some unicode characters there is some unofficial support, but I think this requires a compiler  flag. Point is that per default my vscode complains when I use a smiley the way we use it in a test, so I am a bit unsure if we should keep it in our test since we are kinda relying on unofficial behaviour and maybe we should instead use an officially support UTF8 character.

2. There is this one odd case that annoys me a little bit. When taking the test code
```
template <typename T> struct Contariner {
Container(int Size) {}
};
atuo z = Con^ner<typename T>();
```
I have 2 possible completion items, `Container<typename T>` and `Container<typename T>(int Size)`, so far so good, one for the type and one for the actual constructor. The problem is that only the constructor works as we expect of a replacement item, meanwhile the type will always produce an additional `<typename T>`, this is especially annoying since it is always the first completion item.

3. I struggle a bit with what to do with the `insert` behaviour. Right now the behaviour is you insert the full identifier, but braces and brackets you get only if they are not already present after the suffix. This makes me wonder what even the use case for this is. I personally always prefer the `replace` over the `insert` so I am also not sure what a user of this feature would really want behaviour to be, but I could imagine that maybe they always want their braces and brackets. This would mean that an alternative approach, where braces and prackets are part of the `replace` range so we can properly insert them with `insert` while not creating duplicates with `replace`. Again, I just feel a bit lost here, but mabe users of this feature are mainly people who do not realize there is a `replace` feature, given `insert` is the default in vscode.

argothiel:
Hi, @timon-ul,

Thank you for the review. These are all valid issues, however I think they go a bit outside of this PR's scope.

Ad. 1.
Smileys issue has been introduced by #187623. The current change doesn't introduce any tests with smileys, it doesn't even touch any such tests. This PR is focused on fixing parens suppression in mid-identifier code-completion and this is independent from the encoding.

Ad. 2.
This is a separate issue, unrelated to the mid-identifiers issue. It's broken in general, also in the insert mode, and it's even independent from the `()` parens. Even things like `std::ve^<int>` trigger it, not suppressing correctly the triangular brackets.

Ad. 3.
Unfortunately, I also don't have intuition for the `insert` mode. For me, it's like _legacy broken mode_, and all we care about is to keep it more-or-less backwards compatible. I think many LSP servers were trying to implement some of the replace logic in the insert mode, before the replace logic was officially added to LSP 3.16.0 in 2020. But I think it's too late to change it now, given that users got used to its behavior, unless until we get some real bug reports or requests.

I'll be happy to work on these issues, but as separate changes, not extending the scope of this PR.

timon-ul:
Thank you for your comments, regarding (1) I agree, this is not scope of this PR and not a holdup, I just kinda messed that up in the last PR and wanted to raise it as I only now tested it as part of this PR (oops). The same is kinda true for (2), it's just me noticing the behaviour as part of this PR, again no hold up for this. 

Regarding the last point, I personally also just see it as an inferior mode, but fact is that it is still default behaviour in for example vscode and in [this](https://github.com/clangd/clangd/issues/1302) issue the old maintainers discussed their thoughts on it, which also turned out inconclusive. So I agree to keep the current behaviour and wait for someone with an actual interest in the `insert` behaviour to file a report so we can get a better understanding. Also I have to admit that before your PR's I was not even aware of the replace mode existing (ofc it was not functional but that is besides the point, I just never thought of it as a feature).

argothiel:
> fact is that it is still default behaviour in for example vscode

@timon-ul I think that's the backward compatibility lock-in. Given the `replace` mode is new and usually implemented after the `insert` mode, the editor developers don't want to surprise users with a change of behavior. I have the same issue with Neovim now. I want to implement the `replace` mode in Neovim and I would gladly change the default to `replace` when implemented, but I don't think that's warranted. Changing the mode could be a breaking change for some users, so probably `insert` will remain the default there too, just like it is in VS Code.

PR #197252, opened 2026-05-12
[CIR] Lower builtin_launder

erichkeane:
This patch introduces a cir.launder operation to represent a call to __builtin_launder, which is an optimization barrier. This cir.launder lowers directly to the LLVM-IR launder.invariant.group.

This patch also moves the existing check to see if a type needs laundering to QualType from classic-codegen so it can be shared, however the CodeGenOpt check is still left in codegen/duplicated between the two implementations.


andykaylor (approved):
lgtm

clang/include/clang/CIR/Dialect/IR/CIROps.td
```diff
@@ -8178,4 +8178,30 @@ def CIR_MemChrOp : CIR_Op<"libc.memchr"> {
...
+    %1 = cir.load align(8) %1 : !cir.ptr<!cir.ptr<!rec_S>>, !cir.ptr<!rec_S>
+    %2 = cir.launder(%1) : !cir.ptr<!rec_S>
+  ```
+  }];
+
+  let arguments = (ins CIR_PointerType:$arg);
+  let results = (outs CIR_PointerType:$result);
+
+  let assemblyFormat = [{
+    `(` $arg `)` `:` qualified(type($result)) attr-dict
```
xlauko: nit: I would simplify format a bit
```suggestion
    $arg `:` qualified(type($result)) attr-dict
```

xlauko (approved):
lgtm % nit

clang/lib/AST/Type.cpp
```diff
@@ -5558,6 +5558,38 @@ QualType::DestructionKind QualType::isDestructedTypeImpl(QualType type) {
...
+  if (!Seen.insert(Record).second)
+    return false;
+
+  assert(Record->hasDefinition() &&
+         "Incomplete types should already be diagnosed");
+
+  if (Record->isDynamicClass())
+    return true;
+
+  for (FieldDecl *F : Record->fields()) {
```
AaronBallman, line 5584: Do we have to look at base classes and their fields too?

clang/lib/AST/Type.cpp
```diff
@@ -5558,6 +5558,38 @@ QualType::DestructionKind QualType::isDestructedTypeImpl(QualType type) {
   return DK_none;
 }
 
+static bool
+requiresBuiltinLaunderImpl(const ASTContext &Context, QualType Ty,
+                           llvm::SmallPtrSetImpl<const Decl *> &Seen) {
+  if (const auto *Arr = Context.getAsArrayType(Ty))
```
AaronBallman, line 5564: Do we have to strip other type details like `AttributedType` or `AtomicType`? I presume things like type attributes or wrapping the type with `_Atomic` (in C++) don't remove the need to launder but they would return null from `getAsCXXRecordDecl()`, I believe.

Also, do we need to do any adjustments for reference types?

clang/lib/AST/Type.cpp
```diff
@@ -5558,6 +5558,38 @@ QualType::DestructionKind QualType::isDestructedTypeImpl(QualType type) {
...
+
+  const auto *Record = Ty->getAsCXXRecordDecl();
+  if (!Record)
+    return false;
+
+  // We've already checked this type, or are in the process of checking it.
+  if (!Seen.insert(Record).second)
+    return false;
+
+  assert(Record->hasDefinition() &&
```
AaronBallman: Do we want a similar assert for calling this on a dependent type? Also, should this be using `isIncompleteType()` on the type rather than asking the record if it has a definition?

AaronBallman (commented):
No concerns with the direction of moving things into `QualType` for better code sharing, but I did have some questions about moved code that may point out missed test coverage or bugs (don't have to solve any of it as part of this PR; perfectly fine to file issues for follow-up work).

clang/lib/AST/Type.cpp
erichkeane: Atomic requires trivially copyable, so it can't have any type that IS, so I don't think that is a problem (as you can't form an Atomic to a dynamic type).

I can't find any version of 'attributed-type' where the modified type is a record type, but i can add the stripping-off anyway.

clang/lib/AST/Type.cpp
erichkeane: Yeah, i think a 'better' assert is a good idea.

clang/lib/AST/Type.cpp
erichkeane: Seemingly no.  Best I can tell is the `isDynamicClass` test considers dynamic bases as well: 
https://godbolt.org/z/6rTdaqr66

AaronBallman (approved):
LGTM!

PR #157161, opened 2025-09-05
[WebAssembly] Implement GlobalISel

QuantumSegfault:
A WIP attempt to implement the new GlobalISel system for the WebAssembly backend.

It may not end up being worth the effort in the short term, but hopefully if it works out, it will open the doors to improved code maintainability and future GlobalISel-exclusive optimizations.

Based largely on the AArch GlobalISel, combined with the old SelectionDAG-based (WASM) code.

-----

This PR isn't to be merged. This is just a development branch, and will be closed once the smaller PRs have been merged and main has been brought up to speed.

Partial PRs:

- #178796
- #180263
- #190247
- #193102
- #183694
- #190234
- #197255
- #197256
- #197257
- #197258
- #197259
- #206885
- #218857
- _more to come..._

PR #197255, opened 2026-05-12
[WebAssembly][GlobalISel] Implement conversion/cast ops between integer and floating-point

QuantumSegfault:
Adds legalization and tests for operations converting between int and float (`G_BITCAST`, `G_FPTOSI`, `G_FPTOUI`, `G_SITOFP` and `G_UITOFP`).

Saturating float => int ops will be finished later once comparison operators are in place (which for this, `G_SELECT`, `G_FCMP`, `G_UMIN`, and `G_SMIN` are needed at least).

Split from #157161

PR #197256, opened 2026-05-12
[WebAssembly][GlobalISel] Implement `COPY`

QuantumSegfault:
Adds instruction select handling and tests for `COPY`. While `COPY` gets dissolved in the end, these changes are required to keep the selector from crashing when trying to assign a regclass to its operands.

Split from #157161


QuantumSegfault:
* **#197259** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197259?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#197258** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#197257** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#197256** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197256?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197256?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#197255** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197255?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

llvm/lib/Target/WebAssembly/GISel/WebAssemblyInstructionSelector.cpp
```diff
@@ -83,13 +85,82 @@ WebAssemblyInstructionSelector::WebAssemblyInstructionSelector(
...
+  if (DstReg.isPhysical()) {
+    switch (DstReg.id()) {
+    case WebAssembly::SP32:
+      DstRC = &WebAssembly::I32RegClass;
+      break;
+    case WebAssembly::SP64:
+      DstRC = &WebAssembly::I64RegClass;
+      break;
+    default:
+      reportFatalInternalError("Copy to physical register other than SP32 or SP64?");
```
arsenm: These fatal errors should probably be llvm_unreachable

llvm/lib/Target/WebAssembly/GISel/WebAssemblyInstructionSelector.cpp
```diff
@@ -83,13 +85,82 @@ WebAssemblyInstructionSelector::WebAssemblyInstructionSelector(
 {
 }
 
+bool WebAssemblyInstructionSelector::selectCopy(
+    MachineInstr &I, MachineRegisterInfo &MRI) const {
+  Register DstReg = I.getOperand(0).getReg();
+  Register SrcReg = I.getOperand(1).getReg();
+
+  const TargetRegisterClass *DstRC;
```
arsenm: Use getConstrainedRegClassForOperand? (these APIs need work, there's no reason they should have a MachineOperand involved)

llvm/lib/Target/WebAssembly/GISel/WebAssemblyInstructionSelector.cpp
arsenm: alternatively return false to fallback

llvm/lib/Target/WebAssembly/GISel/WebAssemblyInstructionSelector.cpp
QuantumSegfault: Oh, yeah, that might be simpler.

Yeah, I remember...just haven't gotten around to maybe trying to fix that.

-----

And sorry about the mess. I did not expect Graphite to immedietly post these PRs. I though I'd have a chance to clean them up first (so what's the "publish" button for then!?!?!?).

I guess if doing a stack, the commits should have the PR title and description from the start.

llvm/lib/Target/WebAssembly/GISel/WebAssemblyInstructionSelector.cpp
```diff
@@ -83,13 +84,39 @@ WebAssemblyInstructionSelector::WebAssemblyInstructionSelector(
...
 
+bool WebAssemblyInstructionSelector::selectCopy(
+    MachineInstr &I, MachineRegisterInfo &MRI) const {
+  const TargetRegisterClass *DstRC =
+      TRI.getConstrainedRegClassForOperand(I.getOperand(0), MRI);
+  if (!DstRC)
+    return false;
+
+  const TargetRegisterClass *SrcRC =
+      TRI.getConstrainedRegClassForOperand(I.getOperand(0), MRI);
```
dschuff: should this use operand 1 instead of operand 0?

llvm/lib/Target/WebAssembly/WebAssemblyRegisterInfo.cpp
```diff
@@ -184,8 +184,21 @@ WebAssemblyRegisterInfo::getConstrainedRegClassForOperand(
     const MachineOperand &MO, const MachineRegisterInfo &MRI) const {
   assert(MO.isReg());
 
-  const RegClassOrRegBank &RegClassOrBank =
-      MRI.getRegClassOrRegBank(MO.getReg());
+  Register Reg = MO.getReg();
+
+  if (Reg.isPhysical()) {
+    switch (Reg.id()) {
+    case WebAssembly::SP32:
```
dschuff, line 191: I guess we should also handle FP32/FP64 here.

llvm/lib/Target/WebAssembly/GISel/WebAssemblyInstructionSelector.cpp
QuantumSegfault: Oh, yeah, good catch

llvm/lib/Target/WebAssembly/WebAssemblyRegisterInfo.cpp
QuantumSegfault: I hadn't personally run into it when I was doing the initial draft implementation and tests, but yeah, I can add that in as well.

llvm/lib/Target/WebAssembly/GISel/WebAssemblyInstructionSelector.cpp
```diff
@@ -83,13 +84,39 @@ WebAssemblyInstructionSelector::WebAssemblyInstructionSelector(
...
+  if (!DstRC)
+    return false;
+
+  const TargetRegisterClass *SrcRC =
+      TRI.getConstrainedRegClassForOperand(I.getOperand(1), MRI);
+  if (!SrcRC)
+    return false;
+
+  if (DstRC != SrcRC)
+    return false;
```
arsenm: This check doesn't make sense; COPY between classes is one of the primary reasons to use a copy?

llvm/lib/Target/WebAssembly/GISel/WebAssemblyInstructionSelector.cpp
```diff
@@ -83,13 +84,39 @@ WebAssemblyInstructionSelector::WebAssemblyInstructionSelector(
...
+
+  if (I.getOperand(0).getReg().isVirtual())
+    MRI.setRegClass(I.getOperand(0).getReg(), DstRC);
+  if (I.getOperand(1).getReg().isVirtual())
+    MRI.setRegClass(I.getOperand(1).getReg(), SrcRC);
```
arsenm: RBI.constrainGenericRegister(DstReg, *RC, *MRI)?

llvm/lib/Target/WebAssembly/GISel/WebAssemblyInstructionSelector.cpp
QuantumSegfault: Maybe it used to be, and maybe it still is for other targets. But now that LLT `i32` and `f32` are distinct, there is no overlap. Each legal LLT is assigned it's own unique class, so what used to be `s32 => s32` copy across two regbanks and regclasses is now simply a separate `G_BITCAST`. I'm leaving that condition there.

Can always be revised later if we end up running into it down the line.

arsenm: This may have been a already selected COPY with mismatched classes , this will fail I the nop-select case

QuantumSegfault: Nop-select?

Like `G_SELECT` between the same value?

How could they have mismatching classes? When would legalization ever create a *copy* between two different regclasses, when there is a 1 : 1 mapping between regclasses and valid LLT. COPY is explicitly enforced to only be copy between the same type. And regclass corresponds to one type (`LLT::integer(32)` => `i32regbank` => `i32` (class)) And I'm doing everything I can to avoid `s32`, which WOULD cause this problem.

arsenm: No, like you already had a selected machine copy between register classes. i.e. `%0:some_rc = COPY %1:other_rc`

QuantumSegfault: Okay...maybe. Do you have any particular examples of why/when this would happen? When would the pipeline skip to a fully selected COPY between potentially mismatched classes?

arsenm: There isn't a firm boundary between gMIR and MIR. Fully selected target instructions are allowed to be introduced at any point. COPY is just a special case because it's reused between generic vregs and normal vregs. You can test this by just writing a ordinary machine copy in a mir test

QuantumSegfault: Okay, I've extended the COPY to support transforming it into an appropriate reinterpret. Basically, we allow `COPY` to act as a `G_BITCAST`. However this still does not support COPY across different sized reg classes, nor copies to/from "physical" registers with mismatching classes.

Is this enough?

QuantumSegfault:
Ping

@arsenm @dschuff

dschuff (approved):

QuantumSegfault:
Ping @arsenm

QuantumSegfault:
Okay, I'm merging.

If it needs to be fleshed out to support more cases, that can be done later.

QuantumSegfault:
### Merge activity

* **Jun 30, 7:22 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/llvm/llvm-project/197256).
* **Jun 30, 7:31 AM UTC**: The [Graphite](https://app.graphite.com/github/pr/llvm/llvm-project/197256) merge of this pull request was cancelled.

PR #197258, opened 2026-05-12
[WebAssembly][GlobalISel] Implement floating-point comparisons

QuantumSegfault:
Adds legalization and tests for various floating point comparison operations.

Custom legalization is done for `G_FCMP` to lower certain predicates that Wasm doesn't support into ones that it does.

Custom lowering is also done for `G_FMINNUM`, `G_FMAXNUM`, `G_FMINIMUMNUM`, and `G_FMAXIMUMNUM` to rewrite them in terms of the Wasm-legal `G_FMINIMUM`/`G_FMAXIMUM`.

`G_IS_FPCLASS` is required for `G_FMINIMUMNUM`/`G_FMAXIMUMNUM` lowering so it is also included (default `lower`).

Split from #157161

PR #197257, opened 2026-05-12
[WebAssembly][GlobalISel] Implement integer comparisons and `G_SELECT`

QuantumSegfault:
Adds legalization and tests for various integer comparison operations (namely `G_ICMP` is legal, but also enable `lower` for some other ones), as well as `G_SELECT`.

Split from #157161

PR #96297, opened 2024-06-21
[GlobalISel][AArch64] Add G_FPTOSI_SAT/G_FPTOUI_SAT

davemgreen:
This is an implementation of the saturating fp to int conversions for GlobalISel. On AArch64 the converstion instrctions work this way, producing saturating results.  LegalizerHelper::lowerFPTOINT_SAT is ported from SDAG.

AArch64 has a lot of existing tests for fptosi_sat, covering a wide range of types. I have tried to make most of them work all at once, but a few fall back due to other missing features such as f128 handling for min/max.

PR #197259, opened 2026-05-12
[WebAssembly][GlobalISel] Implement saturating floating-point to int

QuantumSegfault:
Adds legalization and tests for `G_FPTOSI_SAT` and `G_FPTOUI_SAT`.

Split from #157161


QuantumSegfault:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197259?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#197259** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197259?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197259?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#197258** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#197257** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197257?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#197256** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197256?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#197255** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197255?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

llvm/lib/Target/WebAssembly/GISel/WebAssemblyLegalizerInfo.cpp
```diff
@@ -141,7 +141,7 @@ WebAssemblyLegalizerInfo::WebAssemblyLegalizerInfo(
       .minScalar(1, s32);
 
   // TODO: once comparison ops are in place
-  /*if (ST.hasNontrappingFPToInt()) {
+  if (ST.hasNontrappingFPToInt()) {
```
arsenm: Shouldn't this only apply to the .legalFor part? The rest should just hit lower?

llvm/lib/Target/WebAssembly/WebAssemblyGISel.td
```diff
@@ -15,3 +15,16 @@
 
 include "WebAssembly.td"
 include "WebAssemblyCombine.td"
+
+//===----------------------------------------------------------------------===//
+// Miscallenous patterns
```
arsenm: Typo

llvm/lib/Target/WebAssembly/WebAssemblyGISel.td
```diff
@@ -15,3 +15,16 @@
...
+//===----------------------------------------------------------------------===//
+
+def : Pat<(i32 (fp_to_sint_sat_gi F32:$src)), (I32_TRUNC_S_SAT_F32 F32:$src)>;
```
arsenm, line 23: Why don't these patterns exist for sdag as it is?

QuantumSegfault: I'm exactly sure what the difference is, but `fp_to_sint_sat_gi` != `fp_to_sint_sat` (the latter of which does exist already).

llvm/lib/Target/WebAssembly/GISel/WebAssemblyLegalizerInfo.cpp
QuantumSegfault: I did it this way to be explicit that it's only legal or lower for these particular types. Don't support anything else.

But it probably wouldn't hurt to just stick `lower` at the end, and only call `legalForCartesianProduct` conditionally (actually, at that point I'll probably just `legalFor`, manually enumerate the type combo, so I can use the inline boolean Pred form (`legalForCartesianProduct` doesn't allow that)

llvm/lib/Target/WebAssembly/WebAssemblyGISel.td
arsenm: Should fix that, this shouldn't require pattern duplication

QuantumSegfault: How would you suggest fixing this properly? This has been a problem since the beginning:

https://github.com/llvm/llvm-project/pull/96297#discussion_r2952095484

arsenm: The type profile of FP_TO_UINT_SAT should be the same either way, with a GINodeEquiv defining the relation. This is how every other node works, this one is not special

QuantumSegfault: There already is a `GINodeEquiv` in place, but for the distinct `fp_to_sint_sat_gi` DAG node added especially for GISel...but yeah, seems like that `SDTFPToIntOp` and `SDTFPToIntSatOp` need to be reconciled somehow.

Another thing to investigate after FP comparisons.

arsenm: SDTFPToIntOp and SDTFPToIntSatOp are different, SDTFPToIntSatOp carries an additional operand. It looks like the globalisel support was added incorrectly, and just doesn't include the necessary operand. It will require a little bit of a special case in tablegen to encode the type but we have similar cases already for sext_inreg and compares

PR #197262, opened 2026-05-12
[CIR] Lower __builtin_assume_dereferenceable

adams381:
`__builtin_assume_dereferenceable` was reported NYI by ClangIR.  It's heavily exercised by libcxx via `__memory/valid_range.h`, so most of `<algorithm>` currently fails to compile under `-fclangir`.

Adds a `cir.assume_dereferenceable` op (pointer + pointer-sized integer), emits it from CIRGenBuiltin, and lowers it to

```
call void @llvm.assume(i1 true) [ "dereferenceable"(ptr, i64) ]
```

which is what classic Clang's `Builder.CreateDereferenceableAssumption` produces.

The size operand is widened/narrowed to `intptr_t` at the CIRGen layer to match classic CodeGen's `IntPtrTy` conversion.

Tests cover the dynamic-size, constant-size, and narrow-integer-size cases against CIR, CIR->LLVM, and OGCG.

### DIFF
--- a/clang/include/clang/CIR/Dialect/IR/CIROps.td
+++ b/clang/include/clang/CIR/Dialect/IR/CIROps.td
@@ -6157,7 +6157,7 @@ def CIR_AssumeOp : CIR_Op<"assume"> {
   let summary = "Tell the optimizer that a boolean value is true";
   let description = [{
     The `cir.assume` operation takes a single boolean predicate as its first
-    argument and does not have any results.  The operation tells the optimizer
+    argument and does not have any results. The operation tells the optimizer
     that the predicate is always true.
 
     An optional operand bundle carries additional assumption metadata,
--- a/clang/include/clang/CIR/Dialect/IR/CIROps.td
+++ b/clang/include/clang/CIR/Dialect/IR/CIROps.td
@@ -6144,6 +6144,15 @@ def CIR_IsFPClassOp : CIR_Op<"is_fp_class", [Pure]> {
 // Assume Operations
 //===----------------------------------------------------------------------===//
 
+def CIR_AssumeBundleKind : CIR_I32EnumAttr<
+  "AssumeBundleKind", "kind of cir.assume operand bundle", [
+  I32EnumAttrCase<"None", 0>,
+  I32EnumAttrCase<"Align", 1, "align">,
+  I32EnumAttrCase<"SeparateStorage", 2, "separate_storage">,
+  I32EnumAttrCase<"Dereferenceable", 3, "dereferenceable">
+]> {
+}
+
 def CIR_AssumeOp : CIR_Op<"assume"> {
   let summary = "Tell the optimizer that a boolean value is true";
   let description = [{
@@ -6151,49 +6160,35 @@ def CIR_AssumeOp : CIR_Op<"assume"> {
     argument and does not have any results.  The operation tells the optimizer
     that the predicate is always true.
 
-    Optional operand bundles can be attached to carry additional assumption
-    metadata, mirroring MLIR LLVM dialect's `llvm.assume` operand bundles
-    (`mlir/include/mlir/Dialect/LLVMIR/LLVMIntrinsicOps.td`).  Each bundle
-    has a string tag and a (possibly empty) list of operands.  For example,
+    An optional operand bundle carries additional assumption metadata,
+    mirroring MLIR LLVM dialect's `llvm.assume` operand bundles
+    (`mlir/include/mlir/Dialect/LLVMIR/LLVMIntrinsicOps.td`).  The bundle
+    kind is a `AssumeBundleKind` enum; bundle operands follow the kind
+    keyword in the assembly form.  For example,
     `__builtin_assume_dereferenceable(p, n)` lowers to
-    `cir.assume %true ["dereferenceable"(%p, %n : !cir.ptr<!void>, !u64i)]`,
-    which in turn lowers to
+    `cir.assume %true dereferenceable(%p, %n : !cir.ptr<!void>, !u64i)
+    : !cir.bool`, which in turn lowers to
     `call void @llvm.assume(i1 true) [ "dereferenceable"(ptr, i64) ]`.
 
     This operation corresponds to the `__assume` / `__builtin_assume`
     builtins (no bundle), as well as `__builtin_assume_aligned`
-    (`"align"` bundle), `__builtin_assume_separate_storage`
-    (`"separate_storage"` bundle), and `__builtin_assume_dereferenceable`
-    (`"dereferenceable"` bundle).
+    (`align` bundle), `__builtin_assume_separate_storage`
+    (`separate_storage` bundle), and `__builtin_assume_dereferenceable`
+    (`dereferenceable` bundle).
   }];
 
   let arguments = (ins
     CIR_BoolType:$predicate,
-    VariadicOfVariadic<CIR_AnyType, "op_bundle_sizes">:$op_bundle_operands,
-    DenseI32ArrayAttr:$op_bundle_sizes,
-    OptionalAttr<ArrayAttr>:$op_bundle_tags
+    DefaultValuedAttr<CIR_AssumeBundleKind,
+                      "::cir::AssumeBundleKind::None">:$bundle_kind,
+    Variadic<CIR_AnyType>:$bundle_args
   );
 
   let assemblyFormat = [{
-    $predicate
-    custom<OpBundles>($op_bundle_operands, type($op_bundle_operands),
-                      $op_bundle_tags)
+    $predicate custom<AssumeBundle>($bundle_kind, $bundle_args,
+                                    type($bundle_args))
     `:` type($predicate) attr-dict
   }];
-
-  let builders = [
-    OpBuilder<(ins "mlir::Value":$predicate), [{
-      build($_builder, $_state, predicate,
-            /*op_bundle_operands=*/llvm::ArrayRef<mlir::ValueRange>{},
-            /*op_bundle_tags=*/mlir::ArrayAttr{});
-    }]>,
-    OpBuilder<(ins "mlir::Value":$predicate, "llvm::StringRef":$bundleTag,
-                   "mlir::ValueRange":$bundleArgs), [{
-      build($_builder, $_state, predicate,
-            llvm::ArrayRef<mlir::ValueRange>(bundleArgs),
-            $_builder.getStrArrayAttr(bundleTag));
-    }]>
-  ];
 
   let hasVerifier = 1;
 }
--- a/clang/include/clang/CIR/Dialect/IR/CIROps.td
+++ b/clang/include/clang/CIR/Dialect/IR/CIROps.td
@@ -6147,19 +6147,47 @@ def CIR_IsFPClassOp : CIR_Op<"is_fp_class", [Pure]> {
 def CIR_AssumeOp : CIR_Op<"assume"> {
   let summary = "Tell the optimizer that a boolean value is true";
   let description = [{
-    The `cir.assume` operation takes a single boolean prediate as its only
+    The `cir.assume` operation takes a single boolean predicate as its first
     argument and does not have any results. The operation tells the optimizer
     that the predicate is always true.
 
+    An optional operand bundle (tag + variadic args) can be attached to carry
+    additional assumption metadata, mirroring LLVM dialect's `llvm.assume`
+    bundles.  For example, `__builtin_assume_dereferenceable(p, n)` lowers
+    to `cir.assume %true ["dereferenceable"(%p, %n : !cir.ptr<!void>, !u64i)]`,
+    which in turn lowers to
+    `call void @llvm.assume(i1 true) [ "dereferenceable"(ptr, i64) ]`.
+
     This operation corresponds to the `__assume` and the `__builtin_assume`
-    builtin functions.
+    builtin functions, and (with a bundle) to other `__builtin_assume_*`
+    variants such as `__builtin_assume_dereferenceable`.
   }];
 
-  let arguments = (ins CIR_BoolType:$predicate);
+  let arguments = (ins
+    CIR_BoolType:$predicate,
+    OptionalAttr<StrAttr>:$bundle_tag,
+    Variadic<CIR_AnyType>:$bundle_args
+  );
 
   let assemblyFormat = [{
-    $predicate `:` type($predicate) attr-dict
+    $predicate
+    ( ` ` `[` $bundle_tag^ `(` $bundle_args `:` type($bundle_args) `)` `]` )?
+    `:` type($predicate) attr-dict
   }];
+
+  let builders = [
+    OpBuilder<(ins "mlir::Value":$predicate), [{
+      build($_builder, $_state, predicate, /*bundle_tag=*/mlir::StringAttr{},
+            /*bundle_args=*/mlir::ValueRange{});
+    }]>,
+    OpBuilder<(ins "mlir::Value":$predicate, "llvm::StringRef":$bundleTag,
+                   "mlir::ValueRange":$bundleArgs), [{
+      build($_builder, $_state, predicate,
+            $_builder.getStringAttr(bundleTag), bundleArgs);
+    }]>
+  ];
+
+  let hasVerifier = 1;
 }
 
 def CIR_AssumeAlignedOp : CIR_Op<"assume_aligned", [
@@ -6230,35 +6258,6 @@ def CIR_AssumeSepStorageOp : CIR_Op<"assume_separate_storage", [
 
   let assemblyFormat = [{
     $ptr1 `,` $ptr2 `:` qualified(type($ptr1)) attr-dict
-  }];
-}
-
-def CIR_AssumeDereferenceableOp : CIR_Op<"assume_dereferenceable"> {
-  let summary = "Tell the optimizer that a pointer is dereferenceable";
-  let description = [{
-    The `cir.assume_dereferenceable` operation takes a pointer and a size in
-    bytes.  It tells the optimizer that loading from `pointer` is well-defined
-    for any number of bytes up to `size`, i.e. the pointer is known to refer
-    to a region of at least `size` allocated bytes.
-
-    The size argument must be a pointer-sized integer (the lowering converts
-    smaller integer types to the target pointer width before emitting the
-    operand bundle).
-
-    This operation corresponds to the `__builtin_assume_dereferenceable`
-    builtin function.
-
-    Example:
-
-    ```
-    cir.assume_dereferenceable %p, %n : !cir.ptr<!void>, !u64i
-    ```
-  }];
-
-  let arguments = (ins CIR_PointerType:$pointer, CIR_IntType:$size);
-
-  let assemblyFormat = [{
-    $pointer `,` $size `:` qualified(type($pointer)) `,` type($size) attr-dict
   }];
 }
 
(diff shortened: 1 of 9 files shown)


clang/lib/CIR/CodeGen/CIRGenBuiltin.cpp
```diff
@@ -1098,6 +1098,18 @@ RValue CIRGenFunction::emitBuiltinExpr(const GlobalDecl &gd, unsigned builtinID,
     return RValue::get(nullptr);
   }
 
+  case Builtin::BI__builtin_assume_dereferenceable: {
+    mlir::Value ptrValue = emitScalarExpr(e->getArg(0));
+    mlir::Value sizeValue = emitScalarExpr(e->getArg(1));
+    // The operand bundle expects a pointer-sized integer; widen/narrow to
+    // intptr_t to match classic CodeGen.
+    mlir::Type intPtrTy = convertType(getContext().getIntPtrType());
```
erichkeane: Why int-ptr isntead of uintptr?

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -859,6 +859,18 @@ mlir::LogicalResult CIRToLLVMAssumeSepStorageOpLowering::matchAndRewrite(
   return mlir::success();
 }
 
+mlir::LogicalResult CIRToLLVMAssumeDereferenceableOpLowering::matchAndRewrite(
+    cir::AssumeDereferenceableOp op, OpAdaptor adaptor,
+    mlir::ConversionPatternRewriter &rewriter) const {
+  auto cond = mlir::LLVM::ConstantOp::create(rewriter, op.getLoc(),
+                                             rewriter.getI1Type(), 1);
+  mlir::LLVM::AssumeOp::create(
```
erichkeane: Why doesn't this use `replaceOpWithNewOp` like above?  Is there something special here that we can't use that for?

clang/include/clang/CIR/Dialect/IR/CIROps.td
```diff
@@ -6230,6 +6230,35 @@ def CIR_AssumeSepStorageOp : CIR_Op<"assume_separate_storage", [
   }];
 }
 
+def CIR_AssumeDereferenceableOp : CIR_Op<"assume_dereferenceable"> {
```
andykaylor: Do we need a dedicated op for this, as opposed to handling it with the existing CIR_AssumeOp and OpBundles like the LLVM dialect does?

clang/lib/CIR/CodeGen/CIRGenBuiltin.cpp
adams381: Switched to `getUIntPtrType()`.  Also moved this codepath into the bundle redesign -- see Andy's thread below.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Gone -- the redesign routes through the existing `CIRToLLVMAssumeOpLowering`, which uses `replaceOpWithNewOp`.

clang/include/clang/CIR/Dialect/IR/CIROps.td
adams381: Agreed.  Deleted the dedicated op and taught `cir.assume` to optionally carry one operand bundle; builtin now emits `cir.assume %true ["dereferenceable"(%p, %n)]`.  Left `assume_aligned` and `assume_separate_storage` alone for now -- happy to convert them in a follow-up if you'd like.

clang/include/clang/CIR/Dialect/IR/CIROps.td
```diff
@@ -6144,19 +6144,47 @@ def CIR_IsFPClassOp : CIR_Op<"is_fp_class", [Pure]> {
...
     This operation corresponds to the `__assume` and the `__builtin_assume`
-    builtin functions.
+    builtin functions, and (with a bundle) to other `__builtin_assume_*`
+    variants such as `__builtin_assume_dereferenceable`.
   }];
 
-  let arguments = (ins CIR_BoolType:$predicate);
+  let arguments = (ins
+    CIR_BoolType:$predicate,
+    OptionalAttr<StrAttr>:$bundle_tag,
```
erichkeane: Its a little odd that the tag is a string and not an enum.  Also, since these two have to stay in sync, there should probably be a pair-type for the tag/args pair.

That would also remove the need for the 'verifier'.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -820,8 +820,13 @@ mlir::LogicalResult CIRToLLVMSignBitOpLowering::matchAndRewrite(
...
     cir::AssumeOp op, OpAdaptor adaptor,
     mlir::ConversionPatternRewriter &rewriter) const {
-  auto cond = adaptor.getPredicate();
-  rewriter.replaceOpWithNewOp<mlir::LLVM::AssumeOp>(op, cond);
+  mlir::Value cond = adaptor.getPredicate();
+  if (std::optional<llvm::StringRef> tag = op.getBundleTag()) {
+    rewriter.replaceOpWithNewOp<mlir::LLVM::AssumeOp>(op, cond, *tag,
+                                                      adaptor.getBundleArgs());
+  } else {
+    rewriter.replaceOpWithNewOp<mlir::LLVM::AssumeOp>(op, cond);
```
erichkeane, line 825: What happens if we just do away with the branches here, and pass `op.getBundleTag().value_or(...)` and `getBundleArgs` directly to the function here?

clang/include/clang/CIR/Dialect/IR/CIROps.td
adams381: Moved to the parallel-array shape `llvm.assume` itself uses -- `VariadicOfVariadic` of operand groups + `DenseI32ArrayAttr` of segment sizes + `OptionalAttr<ArrayAttr>` of string tags (LLVMOpBase.td:336, LLVMIntrinsicOps.td:633).  The framework keeps operand groups and segment sizes in sync, so the CIR-side verifier collapses to the canonical `llvm.assume` check (tags are `StringAttr`, `numTags == numBundles`).  Also unified `cir.assume_aligned` and `cir.assume_separate_storage` onto the same bundle form -- `cir.assume` is now the single op for assume-with-metadata.

On the enum sub-ask: `llvm.assume`'s tag is intentionally a string because LLVM IR operand-bundle tags are an open universe.  The typed-tag wrappers there (`AssumeAlignTag`, `AssumeSeparateStorageTag` at LLVMDialect.h:88-89) live only at the C++ builder layer; the IR underneath stays string.  Happy to add analogous CIR builder wrappers as cheap follow-up if a fourth bundle kind comes along.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Done -- under the new parallel-array bundle shape the lowering is a single `replaceOpWithNewOp` forward of `(predicate, materialized SmallVector<ValueRange>, op_bundle_tags ArrayAttr)`.  Couldn't literally use `value_or("")` because `llvm.assume`'s `(cond, StringRef tag, ValueRange args)` builder always wraps into a one-bundle form (LLVMDialect.cpp:4321-4324), so an empty tag would mean "one bundle with an empty tag," not "no bundle."  Forwarding the three parallel fields sidesteps that.

erichkeane (commented):
This has honestly gotten even more complicated, and unnecessarily so it seems to me.  

I like that the existing uses of assume were simplified / unified, but the storage of this across the operands, sizes, and tags seems too much.  We should probably abstract out each of the 'bundle' kinds/whatever into a wrapper type that is string + operands-list, and let it handle parsing/printing/verification automatically, rather than having custom verification at all..

adams381:
The parallel-array machinery is over-fitted for a multi-bundle world CIR doesn't actually use.  Every Clang builtin that lowers to `llvm.assume` with bundles emits exactly one (`align`, `separate_storage`, `dereferenceable`), so I'd propose a single-bundle shape with a kind enum:

```tablegen
def CIR_AssumeBundleKindAttr : I32EnumAttr<"AssumeBundleKind",
    "kind of cir.assume operand bundle", [
  I32EnumAttrCase<"None",            0>,
  I32EnumAttrCase<"Align",           1, "align">,
  I32EnumAttrCase<"SeparateStorage", 2, "separate_storage">,
  I32EnumAttrCase<"Dereferenceable", 3, "dereferenceable">
]>;

def CIR_AssumeOp : CIR_Op<"assume"> {
  let arguments = (ins
    CIR_BoolType:$predicate,
    DefaultValuedAttr<CIR_AssumeBundleKindAttr,
                      "::cir::AssumeBundleKind::None">:$bundle_kind,
    Variadic<CIR_AnyType>:$bundle_args
  );
  // cir.assume %true align(%p, %a : !cir.ptr<!void>, !u64i) : !cir.bool
  // cir.assume %true : !cir.bool   (no bundle; kind defaults to None)
  let hasVerifier = 1;
}
```

Gone: `VariadicOfVariadic`, the segment-sizes array, the ~85 lines of custom `printOpBundles` / `parseOpBundles` in `CIRDialect.cpp`.  The (kind absent, args present) invalid state is structurally gone -- kind is always present, defaults to `None`.  The verifier collapses to a per-kind arity check, which is the "abstract out each bundle kind" explicit per-kind shape you flagged.

Lowering goes back to a two-branch shape (`LLVM::AssumeOp::build(cond)` for `None`, `build(cond, stringifyAssumeBundleKind(kind), args)` otherwise) -- not the branchless parallel-array forward we had, but structurally honest for the single-bundle case (the two forms are genuinely distinct entry points in the LLVM dialect builder).

Is this the shape you had in mind, or something else?

xlauko:
> ```
> def CIR_AssumeBundleKindAttr : I32EnumAttr<"AssumeBundleKind",
>     "kind of cir.assume operand bundle", [
>   I32EnumAttrCase<"None",            0>,
>   I32EnumAttrCase<"Align",           1, "align">,
>   I32EnumAttrCase<"SeparateStorage", 2, "separate_storage">,
>   I32EnumAttrCase<"Dereferenceable", 3, "dereferenceable">
> ]>;


nit: you should use `CIR_I32EnumAttr` here

erichkeane:
> The parallel-array machinery is over-fitted for a multi-bundle world CIR doesn't actually use. Every Clang builtin that lowers to `llvm.assume` with bundles emits exactly one (`align`, `separate_storage`, `dereferenceable`), so I'd propose a single-bundle shape with a kind enum:
> 
> ```
> def CIR_AssumeBundleKindAttr : I32EnumAttr<"AssumeBundleKind",
>     "kind of cir.assume operand bundle", [
>   I32EnumAttrCase<"None",            0>,
>   I32EnumAttrCase<"Align",           1, "align">,
>   I32EnumAttrCase<"SeparateStorage", 2, "separate_storage">,
>   I32EnumAttrCase<"Dereferenceable", 3, "dereferenceable">
> ]>;
> 
> def CIR_AssumeOp : CIR_Op<"assume"> {
>   let arguments = (ins
>     CIR_BoolType:$predicate,
>     DefaultValuedAttr<CIR_AssumeBundleKindAttr,
>                       "::cir::AssumeBundleKind::None">:$bundle_kind,
>     Variadic<CIR_AnyType>:$bundle_args
>   );
>   // cir.assume %true align(%p, %a : !cir.ptr<!void>, !u64i) : !cir.bool
>   // cir.assume %true : !cir.bool   (no bundle; kind defaults to None)
>   let hasVerifier = 1;
> }
> ```
> 
> Gone: `VariadicOfVariadic`, the segment-sizes array, the ~85 lines of custom `printOpBundles` / `parseOpBundles` in `CIRDialect.cpp`. The (kind absent, args present) invalid state is structurally gone -- kind is always present, defaults to `None`. The verifier collapses to a per-kind arity check, which is the "abstract out each bundle kind" explicit per-kind shape you flagged.
> 
> Lowering goes back to a two-branch shape (`LLVM::AssumeOp::build(cond)` for `None`, `build(cond, stringifyAssumeBundleKind(kind), args)` otherwise) -- not the branchless parallel-array forward we had, but structurally honest for the single-bundle case (the two forms are genuinely distinct entry points in the LLVM dialect builder).
> 
> Is this the shape you had in mind, or something else?

I mean, I was happy with/requested a assume-has-a-single-bundle-kind earlier, but you'd added the additional complexity in your last patch for seemingly no reason, all I'd requested is to have the two packaged together, to keep arity correct.

Given the above, why do you still need a verifier? 

Side note: Is the above LLM generated text?  I have quite a few comments/reviews that i believe you've done that with, and it makes it VERY difficult to actually review your code/ideas.  You're generating a substantial amount of work for reviewers.

clang/include/clang/CIR/Dialect/IR/CIROps.td
adams381: Implemented the single-bundle enum shape: AssumeBundleKind (None, align, separate_storage, dereferenceable) with flat bundle_args.  The invalid (kind absent, args present) state is gone; the verifier is just per-kind arity.  Lowering uses replaceOpWithNewOp with stringifyAssumeBundleKind when kind != None.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: With the enum reshape, lowering has a None vs bundle branch again (plain llvm.assume(cond) when kind is None).  The bundle path is still one replaceOpWithNewOp forwarding the tag and args.

clang/include/clang/CIR/Dialect/IR/CIROps.td
```diff
@@ -6144,93 +6144,53 @@ def CIR_IsFPClassOp : CIR_Op<"is_fp_class", [Pure]> {
...
+]> {
+}
+
 def CIR_AssumeOp : CIR_Op<"assume"> {
   let summary = "Tell the optimizer that a boolean value is true";
   let description = [{
-    The `cir.assume` operation takes a single boolean prediate as its only
-    argument and does not have any results. The operation tells the optimizer
+    The `cir.assume` operation takes a single boolean predicate as its first
+    argument and does not have any results.  The operation tells the optimizer
```
erichkeane: Why this double-space change?

clang/lib/CIR/Dialect/IR/CIRDialect.cpp
```diff
@@ -372,6 +372,91 @@ LogicalResult cir::DeleteArrayOp::verify() {
...
+      return emitOpError("unexpected bundle operands for kind 'none'");
+    return success();
+  }
+
+  if (numArgs == 0)
+    return emitOpError("expected bundle operands for kind '")
+           << cir::stringifyAssumeBundleKind(kind) << "'";
+
+  switch (kind) {
+  case cir::AssumeBundleKind::Align:
```
erichkeane, line 442: Do we intend this list to grow much more?  If so, we probably want to do some sort of tablegen for these.

clang/test/CIR/IR/invalid-assume.cir
```diff
@@ -0,0 +1,25 @@
...
+
+// -----
+
+!void = !cir.void
+
+module {
+  cir.func @assume_deref_wrong_arity(%arg0 : !cir.ptr<!void>) {
+    %0 = cir.const #cir.bool<true> : !cir.bool
+    // expected-error@+1 {{dereferenceable bundle expects 2 operands}}
+    cir.assume %0 dereferenceable(%arg0 : !cir.ptr<!void>) : !cir.bool
```
erichkeane, line 22: we should probably have one of hte roundtrip tests to.

adams381: Added clang/test/CIR/IR/assume.cir roundtrip coverage on the branch.

clang/include/clang/CIR/Dialect/IR/CIROps.td
adams381: Reverted the double-space in the description — accidental formatting slip.

clang/test/CIR/IR/invalid-assume.cir
adams381: Added `@assume_align` roundtrip in `assume.cir`; plain/deref/separate_storage were already covered.

clang/lib/CIR/Dialect/IR/CIRDialect.cpp
adams381: Bundle tags are already TableGen (`CIR_AssumeBundleKind` in CIROps.td). The switch in `verify()` is per-kind operand-count checks only — same pattern as other CIR ops with enum kinds and arity rules.

clang/lib/CIR/Dialect/IR/CIRDialect.cpp
```diff
@@ -372,6 +372,91 @@ LogicalResult cir::DeleteArrayOp::verify() {
...
+    if (numArgs != 2 && numArgs != 3)
+      return emitOpError("align bundle expects 2 or 3 operands");
```
erichkeane, line 444: probably should test all of these too.

clang/lib/CIR/Dialect/IR/CIRDialect.cpp
```diff
@@ -372,6 +372,91 @@ LogicalResult cir::DeleteArrayOp::verify() {
...
+  size_t numArgs = getBundleArgs().size();
+
+  if (kind == cir::AssumeBundleKind::None) {
+    if (numArgs != 0)
+      return emitOpError("unexpected bundle operands for kind 'none'");
+    return success();
+  }
+
+  if (numArgs == 0)
+    return emitOpError("expected bundle operands for kind '")
```
erichkeane, line 438: needs to be tested.

clang/lib/CIR/Dialect/IR/CIRDialect.cpp
```diff
@@ -372,6 +372,91 @@ LogicalResult cir::DeleteArrayOp::verify() {
...
+  if (failed(p.parseOptionalKeyword(&keyword))) {
+    bundleKindAttr = cir::AssumeBundleKindAttr::get(
+        p.getContext(), cir::AssumeBundleKind::None);
+    return success();
+  }
+
+  std::optional<cir::AssumeBundleKind> parsedKind =
+      cir::symbolizeAssumeBundleKind(keyword);
+  if (!parsedKind)
+    return p.emitError(loc, "unknown assume bundle kind '") << keyword << "'";
```
erichkeane, line 413: We should have a test for this too.

erichkeane (commented):
Still missing some invald tests, else this looks good at the moment

clang/lib/CIR/Dialect/IR/CIRDialect.cpp
adams381: Added invalid-assume.cir cases for unknown bundle kind parsing and align / separate_storage / dereferenceable operand-count verification.

clang/lib/CIR/Dialect/IR/CIRDialect.cpp
adams381: Added invalid-assume.cir cases for unknown bundle kind parsing and align / separate_storage / dereferenceable operand-count verification.

clang/lib/CIR/Dialect/IR/CIRDialect.cpp
adams381: Added invalid-assume.cir cases for unknown bundle kind parsing and align / separate_storage / dereferenceable operand-count verification.

erichkeane (approved):

PR #197267, opened 2026-05-12
[AMDGPU] Widen MUBUF/MTBUF source-vgpr WAR hazard on gfx940-family to SGPR soffset

panditsa:
createsVALUHazard previously gated the MUBUF/MTBUF source-vgpr WAR hazard to fire only when SOFFSET was a literal or absent. On gfx940-family subtargets that gate is too narrow: the hazard also fires when SOFFSET is sourced from an SGPR.

Concretely, on gfx950 a sequence of the form

```
  buffer_store_dwordx4 v[X:X+3], voff, descr, sN offen
  v_pk_mul_f32 v[X:X+1], <src>, <src>           # next VALU cycle
```

deterministically commits the post-pk_mul value of v[X+1] to memory for the second dword of the store; the other three dwords store correctly.

The wait-state window depends on the SOFFSET shape:

  - sgpr-sourced SOFFSET: 1 wait state  (S_NOP 0)
  - literal/absent SOFFSET: 2 wait states (S_NOP 1)
  - pre-gfx940 literal SOFFSET (unchanged): 1 wait state
  - pre-gfx940 sgpr SOFFSET (unchanged): no hazard

Implementation widens createsVALUHazard's trigger on gfx940-family and splits checkVALUHazardsHelper's single 2-wait-state scan into a per-MI window: a 1-cycle scan and a 2-cycle scan, taking the max. FLAT-wide stores keep the existing 2-wait-state behavior on gfx940 family.

Empirically reproduced and characterised on gfx950 by Triton's fused-attention backward kernel; the threshold sweep in repro/minimal_hazard pins the sgpr-SOFFSET window at exactly 1 cycle (a single S_NOP 0 clears the corruption). gfx942 (MI300) was not validated on hardware; the same gate is applied there because the existing recognizer rule treats the gfx940 family uniformly.

New MIR test covers literal and SGPR soffset on gfx900 (older), gfx942 and gfx950, plus negative cases (dwordx2 store, non-overlapping write).

Assisted-by: Cursor <cursoragent@cursor.com>


llvm/lib/Target/AMDGPU/GCNHazardRecognizer.cpp
```diff
@@ -949,25 +952,51 @@ int GCNHazardRecognizer::createsVALUHazard(const MachineInstr &MI) const {
...
-    return DataIdx >= 0 &&
-           TRI->regsOverlap(MI.getOperand(DataIdx).getReg(), Reg);
+  const Register Reg = Def.getReg();
+
+  const int MaxWaitStates = ST.hasGFX940Insts() ? 2 : 1;
+
+  auto WindowFor = [this, TII](const MachineInstr &MI) -> int {
+    if (!ST.hasGFX940Insts())
+      return 1;
+    if (TII->isMUBUF(MI) || TII->isMTBUF(MI)) {
```
arsenm:
```suggestion
    if (TII->isBUF(MI)) {
```

llvm/test/CodeGen/AMDGPU/buffer-store-dwordx4-vpk-mul-war-hazard-gfx942.mir
```diff
@@ -0,0 +1,115 @@
+# RUN: llc -mtriple=amdgcn -mcpu=gfx950 -verify-machineinstrs \
+# RUN:   -run-pass=post-RA-hazard-rec %s -o - | FileCheck -check-prefix=GFX950 %s
+# RUN: llc -mtriple=amdgcn -mcpu=gfx942 -verify-machineinstrs \
+# RUN:   -run-pass=post-RA-hazard-rec %s -o - | FileCheck -check-prefix=GFX942 %s
+# RUN: llc -mtriple=amdgcn -mcpu=gfx900 -verify-machineinstrs \
+# RUN:   -run-pass=post-RA-hazard-rec %s -o - | FileCheck -check-prefix=GFX900 %s
```
arsenm:
```suggestion
# RUN: llc -mtriple=amdgcn -mcpu=gfx950 \
# RUN:   -run-pass=post-RA-hazard-rec %s -o - | FileCheck -check-prefix=GFX950 %s
# RUN: llc -mtriple=amdgcn -mcpu=gfx942 \
# RUN:   -run-pass=post-RA-hazard-rec %s -o - | FileCheck -check-prefix=GFX942 %s
# RUN: llc -mtriple=amdgcn -mcpu=gfx900 \
# RUN:   -run-pass=post-RA-hazard-rec %s -o - | FileCheck -check-prefix=GFX900 %s
```

llvm/test/CodeGen/AMDGPU/buffer-store-dwordx4-vpk-mul-war-hazard-gfx942.mir
```diff
@@ -0,0 +1,115 @@
...
+# GFX950-NEXT: $vgpr0_vgpr1 = nofpexcept V_PK_MUL_F32
+#
+# GFX942-LABEL: name: buffer_store_dwordx4_sgpr_soffset_then_vpk_mul
+# GFX942: BUFFER_STORE_DWORDX4_OFFEN_exact
+# GFX942-NEXT: S_NOP 0
+# GFX942-NEXT: $vgpr0_vgpr1 = nofpexcept V_PK_MUL_F32
+#
+# GFX900-LABEL: name: buffer_store_dwordx4_sgpr_soffset_then_vpk_mul
+# GFX900: BUFFER_STORE_DWORDX4_OFFEN_exact
+# GFX900-NEXT: $vgpr0_vgpr1 = nofpexcept V_PK_MUL_F32
```
arsenm: Use update_mir_test_checks with a common prefix

llvm/lib/Target/AMDGPU/GCNHazardRecognizer.cpp
```diff
@@ -913,15 +913,18 @@ int GCNHazardRecognizer::createsVALUHazard(const MachineInstr &MI) const {
...
-    // If we have no soffset operand, then assume this field has been
-    // hardcoded to zero.
-    if (AMDGPU::getRegBitWidth(VDataRCID) > 64 &&
-        (!SOffset || !SOffset->isReg()))
-      return VDataIdx;
+    if (AMDGPU::getRegBitWidth(VDataRCID) > 64) {
+      // On gfx940-family the BUFFER_STORE source-vgpr WAR hazard exists for
+      // every SOFFSET shape; the wait-state count differs by SOFFSET, and is
+      // computed in checkVALUHazardsHelper. Pre-gfx940 keeps the original
+      // narrow gate.
```
jayfoad: Nit: the comment should not refer to how the code used to be:
```suggestion
      // computed in checkVALUHazardsHelper. Pre-gfx940 the hazard only exists if soffset is not an SGPR.
```

llvm/lib/Target/AMDGPU/GCNHazardRecognizer.cpp
```diff
@@ -949,25 +952,51 @@ int GCNHazardRecognizer::createsVALUHazard(const MachineInstr &MI) const {
...
+      if (SOffset && SOffset->isReg())
+        return 1;
+    }
+    return 2;
   };
 
-  int WaitStatesNeededForDef =
-    VALUWaitStates - getWaitStatesSince(IsHazardFn, VALUWaitStates);
-  WaitStatesNeeded = std::max(WaitStatesNeeded, WaitStatesNeededForDef);
+  for (int Window = 1; Window <= MaxWaitStates; ++Window) {
```
jayfoad: I don't understand this part. Why do you need to call getWaitStatesSince multiple times in a loop? That seems pretty expensive.

panditsa: The loop splits producers into two buckets: SOFFSET=literal/absent (2 wait states) and SOFFSET=SGPR (1 wait state), and scans each separately. The motivation is the case where two wide BUFFER_STOREs sharing source VGPRs one with literal SOFFSET, one with SGPR SOFFSET sit before a VALU write:

   ```
   cycle -2:  buffer_store_dwordx4 ..., 0 offen     window=2
   cycle -1:  buffer_store_dwordx4 ..., s1 offen    window=1
   cycle  0:  consumer
   ```

`getWaitStatesSince` terminates at the first matching producer it finds, so a single merged predicate returns one distance with no way to know which producer's window to pair it with so you can either over- or under-pad. Bucketing by window makes each scan's returned distance meaningful against its own limit.

However, one scan can cover the same case. I'll update the logic here.

jayfoad: > The loop splits producers into two buckets: SOFFSET=literal/absent (2 wait states) and SOFFSET=SGPR (1 wait state), and scans each separately. The motivation is the case where two wide BUFFER_STOREs sharing source VGPRs one with literal SOFFSET, one with SGPR SOFFSET sit before a VALU write:
> 
> ```
> cycle -2:  buffer_store_dwordx4 ..., 0 offen     window=2
> cycle -1:  buffer_store_dwordx4 ..., s1 offen    window=1
> cycle  0:  consumer
> ```
> 
> `getWaitStatesSince` terminates at the first matching producer it finds, so a single merged predicate returns one distance with no way to know which producer's window to pair it with so you can either over- or under-pad. Bucketing by window makes each scan's returned distance meaningful against its own limit.

Well, since the max wait (2 cycles) is only one more than the min wait (1 cycle), you could actually stop at the first producer and you would never get an incorrect result.

But the way you have written it looks OK too.

llvm/lib/Target/AMDGPU/GCNHazardRecognizer.cpp
```diff
@@ -949,25 +952,57 @@ int GCNHazardRecognizer::createsVALUHazard(const MachineInstr &MI) const {
...
+  // needed using that producer's own window. The predicate always returns
+  // false so the walk runs to MaxWaitStates.
+  int Distance = 0;
+  auto Counter = [&](const MachineInstr &MI) {
+    int DataIdx = createsVALUHazard(MI);
+    if (DataIdx >= 0 &&
+        TRI->regsOverlap(MI.getOperand(DataIdx).getReg(), Reg)) {
+      int Need = WindowFor(MI) - Distance;
+      if (Need > WaitStatesNeeded)
+        WaitStatesNeeded = Need;
```
jayfoad: Nit: use `WaitStatesNeeded = std::max(...)` ?

jayfoad (approved):
LGTM

ronlieb:
Hi @panditsa  we are seeing numerical failures in our downstream testing of MIOpen and rocBLAS, reverting this upstream. Please work with Larry Meadows on possible fix and relanding this one.

ronlieb:
holding off on landing the revert , Larry will provide additional details here shortly.

lfmeadow:
Here are two patches (against ROCm/llvm-project but applicable to upstream). First is a code fix, second is a new regression test. 
Blurb: The checkVALUHazardsHelper rewrite accumulated wait-state distance in a
  single counter shared across recursively-visited predecessor blocks, so
   on branchy control flow it over-counts and drops the protective S_NOP;
   the fix restores per-window getWaitStatesSince min-over-paths
  accounting while preserving #197267's sgpr-soffset windows. The new
  test fails without the fix and passes with it on gfx942/gfx950.

[add_regression_test.patch](https://github.com/user-attachments/files/28915074/add_regression_test.patch)
[fix_GCNHazardRecognizer.patch](https://github.com/user-attachments/files/28915075/fix_GCNHazardRecognizer.patch)

The patch fixes the two internal regressions as well as maintaining the Triton fix. The key missing piece of this PR was handling multi-block hazards correctly.

Ideally a new PR would include the functionality of this patch as well as the new multi-block regression test.

PR #197268, opened 2026-05-12
[clang-format] Add BreakBeforeReturnType option

dlebed:
In certain codebases (e.g. embedded) — function declarations could accumulate a long prefix of specifiers and attributes (`static`, `inline`, `__attribute__((...))`, project-specific `AttributeMacros`, etc.) before the return type, which buries the core prototype and pushes parameters past the column limit.

This patch adds a `BreakBeforeReturnType` style option that places that prefix on its own line(s):

```cpp
__attribute__((always_inline)) static inline
int do_thing(int a, int b, int c);
```

The recognized prefix tokens are function/storage specifiers (`static`, `extern`, `inline`, `virtual`, `constexpr`, `consteval`, `friend`, `export`, `_Noreturn`, `__forceinline`), C++11 attribute groups `[[...]]`, GNU/MSVC attribute groups `__attribute__((...))` / `__declspec(...)`, and identifiers configured via `AttributeMacros`.

The new `BreakBeforeReturnTypeStyle` enum has values `None`, `All`, `TopLevel`, `AllDefinitions`, and `TopLevelDefinitions`. The default is `None`, preserving previous behavior. Constructors and destructors are not affected. The option composes with `BreakAfterReturnType`, `BreakAfterAttributes`, and `BreakTemplateDeclarations`.

`ContinuationIndenter::getNewLineColumn` is adjusted so the wrapped return type is dedented to the line's base indent when the preceding token is a function/storage specifier keyword, matching the behavior already used after attribute groups.

Adds tests in `FormatTest.cpp`.

Assisted-by: Claude (claude-opus-4-7, Claude Code)

### DIFF
--- a/clang/docs/ClangFormatStyleOptions.rst
+++ b/clang/docs/ClangFormatStyleOptions.rst
@@ -3694,6 +3694,37 @@ the configuration (without a prefix: ``Auto``).
 
 
 
+.. _BreakBeforeReturnType:
+
+**BreakBeforeReturnType** (``BreakBeforeReturnTypeStyle``) :versionbadge:`clang-format 23` :ref:`¶ <BreakBeforeReturnType>`
+  The function declaration/definition return type breaking style to use.
+  To have identifier macros (e.g. ``__always_inline``) treated as
+  specifiers, add them to ``AttributeMacros``.
+
+  Possible values:
+
+  * ``BBRTS_None`` (in configuration: ``None``)
+    Do not force a break before the return type.
+
+  * ``BBRTS_All`` (in configuration: ``All``)
+    Always break before the return type.
+
+    .. code-block:: c++
+
+      static inline
+      void f();
+
+  * ``BBRTS_TopLevel`` (in configuration: ``TopLevel``)
+    Break before the return type of top-level functions only.
+
+  * ``BBRTS_AllDefinitions`` (in configuration: ``AllDefinitions``)
+    Break before the return type of function definitions only.
+
+  * ``BBRTS_TopLevelDefinitions`` (in configuration: ``TopLevelDefinitions``)
+    Break before the return type of top-level definitions only.
+
+
+
 .. _BreakBeforeTemplateCloser:
 
 **BreakBeforeTemplateCloser** (``Boolean``) :versionbadge:`clang-format 21` :ref:`¶ <BreakBeforeTemplateCloser>`
--- a/clang/docs/ClangFormatStyleOptions.rst
+++ b/clang/docs/ClangFormatStyleOptions.rst
@@ -3698,8 +3698,9 @@ the configuration (without a prefix: ``Auto``).
 
 **BreakBeforeReturnType** (``BreakBeforeReturnTypeStyle``) :versionbadge:`clang-format 23` :ref:`¶ <BreakBeforeReturnType>`
   The function declaration/definition return type breaking style to use.
-  To have identifier macros (e.g. ``__always_inline``) treated as
-  specifiers, add them to ``AttributeMacros``.
+  Trailing return types (``auto f() -> T``) are not affected. To have
+  identifier macros (e.g. ``__always_inline``) treated as specifiers,
+  add them to ``AttributeMacros``.
 
   Possible values:
 
--- a/clang/include/clang/Format/Format.h
+++ b/clang/include/clang/Format/Format.h
@@ -2481,8 +2481,9 @@ struct FormatStyle {
   };
 
   /// The function declaration/definition return type breaking style to use.
-  /// To have identifier macros (e.g. ``__always_inline``) treated as
-  /// specifiers, add them to ``AttributeMacros``.
+  /// Trailing return types (``auto f() -> T``) are not affected. To have
+  /// identifier macros (e.g. ``__always_inline``) treated as specifiers,
+  /// add them to ``AttributeMacros``.
   /// \version 23
   BreakBeforeReturnTypeStyle BreakBeforeReturnType;
 
--- a/clang/include/clang/Format/Format.h
+++ b/clang/include/clang/Format/Format.h
@@ -2462,6 +2462,30 @@ struct FormatStyle {
   /// \version 16
   BreakBeforeInlineASMColonStyle BreakBeforeInlineASMColon;
 
+  /// Different ways to break before the function return type.
+  enum BreakBeforeReturnTypeStyle : int8_t {
+    /// Do not force a break before the return type.
+    BBRTS_None,
+    /// Always break before the return type.
+    /// \code
+    ///   static inline
+    ///   void f();
+    /// \endcode
+    BBRTS_All,
+    /// Break before the return type of top-level functions only.
+    BBRTS_TopLevel,
+    /// Break before the return type of function definitions only.
+    BBRTS_AllDefinitions,
+    /// Break before the return type of top-level definitions only.
+    BBRTS_TopLevelDefinitions,
+  };
+
+  /// The function declaration/definition return type breaking style to use.
+  /// To have identifier macros (e.g. ``__always_inline``) treated as
+  /// specifiers, add them to ``AttributeMacros``.
+  /// \version 23
+  BreakBeforeReturnTypeStyle BreakBeforeReturnType;
+
   /// If ``true``, break before a template closing bracket (``>``) when there is
   /// a line break after the matching opening bracket (``<``).
   /// \code
@@ -6092,6 +6116,7 @@ struct FormatStyle {
            BreakBeforeCloseBracketSwitch == R.BreakBeforeCloseBracketSwitch &&
            BreakBeforeConceptDeclarations == R.BreakBeforeConceptDeclarations &&
            BreakBeforeInlineASMColon == R.BreakBeforeInlineASMColon &&
+           BreakBeforeReturnType == R.BreakBeforeReturnType &&
            BreakBeforeTemplateCloser == R.BreakBeforeTemplateCloser &&
            BreakBeforeTernaryOperators == R.BreakBeforeTernaryOperators &&
            BreakBinaryOperations == R.BreakBinaryOperations &&
--- a/clang/lib/Format/TokenAnnotator.cpp
+++ b/clang/lib/Format/TokenAnnotator.cpp
@@ -4062,6 +4062,60 @@ bool TokenAnnotator::mustBreakForReturnType(const AnnotatedLine &Line) const {
   return false;
 }
 
+bool TokenAnnotator::mustBreakBeforeReturnType(
+    const AnnotatedLine &Line) const {
+  assert(Line.MightBeFunctionDecl);
+
+  switch (Style.BreakBeforeReturnType) {
+  case FormatStyle::BBRTS_None:
+    return false;
+  case FormatStyle::BBRTS_All:
+    return true;
+  case FormatStyle::BBRTS_TopLevel:
+    return Line.Level == 0;
+  case FormatStyle::BBRTS_AllDefinitions:
+    return Line.mightBeFunctionDefinition();
+  case FormatStyle::BBRTS_TopLevelDefinitions:
+    return Line.Level == 0 && Line.mightBeFunctionDefinition();
+  }
+
+  return false;
+}
+
+static FormatToken *findReturnTypeStart(const AnnotatedLine &Line) {
+  auto *Tok = Line.getFirstNonComment();
+  if (!Tok)
+    return nullptr;
+
+  if (Tok->is(tok::kw_template)) {
+    auto *Opener = Tok->Next;
+    while (Opener && Opener->isNot(TT_TemplateOpener))
+      Opener = Opener->Next;
+    if (!Opener || !Opener->MatchingParen)
+      return nullptr;
+    Tok = Opener->MatchingParen->Next;
+  }
+
+  while (Tok) {
+    if (isReturnTypePrefixSpecifier(*Tok) ||
+        Tok->isOneOf(tok::kw___attribute, tok::kw___declspec) ||
+        Tok->is(TT_AttributeMacro)) {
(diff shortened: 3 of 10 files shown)


clang/unittests/Format/FormatTest.cpp
```diff
@@ -10594,6 +10594,229 @@ TEST_F(FormatTest, ReturnTypeBreakingStyle) {
...
+  FormatStyle Style = getLLVMStyle();
+
+  // 1. None (default): unchanged.
+  EXPECT_EQ(Style.BreakBeforeReturnType, FormatStyle::RTBS_None);
+  verifyFormat("static inline void myfun(void);", Style);
+
+  Style.BreakBeforeReturnType = FormatStyle::RTBS_All;
+
+  // 2. Basic C-style "static inline" declaration (motivating example).
+  verifyFormat("static inline\n"
```
HazardyKnusperkeks, line 10601: I think all of these can be done with the 2 argument version.

clang/unittests/Format/FormatTest.cpp
```diff
@@ -10594,6 +10594,229 @@ TEST_F(FormatTest, ReturnTypeBreakingStyle) {
...
+  // 9. Return type with multiple qualifiers stays whole.
+  verifyFormat("static\n"
+               "const struct foo *g(void);",
+               "static const struct foo *g(void);", Style);
+
+  // 10. Pointer-to-function return.
+  // Cross-check: pre-existing BreakAfterReturnType=All has the same
+  // behavior on this construct (clang-format's function-declaration
+  // detection is the gate). Validate via the existing option here.
+  {
```
HazardyKnusperkeks: Drop the block.

clang/unittests/Format/FormatTest.cpp
```diff
@@ -10594,6 +10594,229 @@ TEST_F(FormatTest, ReturnTypeBreakingStyle) {
...
+    FormatStyle AfterStyle = getLLVMStyle();
```
HazardyKnusperkeks: Why a new style object?

clang/unittests/Format/FormatTest.cpp
```diff
@@ -10594,6 +10594,229 @@ TEST_F(FormatTest, ReturnTypeBreakingStyle) {
...
+  // 11. friend inside class.
+  verifyFormat("class A {\n"
+               "  friend\n"
+               "  int f();\n"
+               "};",
+               "class A { friend int f(); };", Style);
+
+  // 12. alignas: cross-check that pre-existing BreakAfterReturnType=All
+  // also leaves `alignas(N) ...` declarations untouched — clang-format's
+  // function-declaration detection is the gate, not our new option.
```
HazardyKnusperkeks: Words like `new` do not age very well...

Most of the comments can and should be dropped.

clang/unittests/Format/FormatTest.cpp
```diff
@@ -10594,6 +10594,229 @@ TEST_F(FormatTest, ReturnTypeBreakingStyle) {
...
+  verifyFormat("static\n"
+               "auto f() -> int;",
+               "static auto f() -> int;", Style);
+
+  // 15. Parameter reflow: after the prefix is moved up, the return-type
+  // / parameter line gets the full ColumnLimit budget, so parameters
+  // that previously wrapped may now fit on a single line. The original
+  // line is ~88 chars and would wrap parameters under the default LLVM
+  // ColumnLimit=80; with the option, line 2 is ~48 chars and fits.
+  verifyFormat(
```
HazardyKnusperkeks: Can you do this just with ColumnLimit 50 or so, that there's no break in the string literal?

clang/unittests/Format/FormatTest.cpp
```diff
@@ -10594,6 +10594,229 @@ TEST_F(FormatTest, ReturnTypeBreakingStyle) {
...
+  // 27. Empty-definition attribute macro: observable behaviour identical
+  // to #23 — clang-format only sees the identifier regardless of
+  // expansion.
+  Style.AttributeMacros = {"LIBC_INLINE"};
+  verifyFormat("LIBC_INLINE static\n"
+               "int small(void);",
+               "LIBC_INLINE static int small(void);", Style);
+
+  // 28. Combined mix: C++ attr + GNU attr + macro + specifier.
+  Style.AttributeMacros = {"LIBC_INLINE"};
```
HazardyKnusperkeks, line 10757: Already set.

clang/lib/Format/TokenAnnotator.cpp
```diff
@@ -4053,6 +4053,106 @@ bool TokenAnnotator::mustBreakForReturnType(const AnnotatedLine &Line) const {
...
+    // Function/storage specifier keywords.
+    if (Tok->isOneOf(tok::kw_static, tok::kw_extern, tok::kw_inline,
+                     tok::kw_virtual, tok::kw_explicit, tok::kw_constexpr,
+                     tok::kw_consteval, tok::kw_constinit, tok::kw_friend,
+                     tok::kw_thread_local, tok::kw_mutable,
+                     tok::kw__Noreturn)) {
+      Tok = Tok->Next;
+      continue;
+    }
+    // alignas(...) / _Alignas(...).
```
HazardyKnusperkeks: quite obvious... Please drop this kind of comments.

clang/lib/Format/TokenAnnotator.cpp
```diff
@@ -4053,6 +4053,106 @@ bool TokenAnnotator::mustBreakForReturnType(const AnnotatedLine &Line) const {
...
+    while (Opener && Opener->isNot(TT_TemplateOpener))
+      Opener = Opener->Next;
+    if (!Opener || !Opener->MatchingParen)
+      return nullptr;
+    Tok = Opener->MatchingParen->Next;
+  }
+
+  while (Tok) {
+    // Function/storage specifier keywords.
+    if (Tok->isOneOf(tok::kw_static, tok::kw_extern, tok::kw_inline,
```
HazardyKnusperkeks: I don't think one can return thread_local. mutable also isn't something you encounter for return types.

clang/docs/ClangFormatStyleOptions.rst
```diff
@@ -3659,6 +3659,173 @@ the configuration (without a prefix: ``Auto``).
...
+    static inline int g();
+    static inline
+    int h() {
+      return 1;
+    }
+
+  Possible values:
+
+  * ``RTBS_None`` (in configuration: ``None``)
+    This is **deprecated**. See ``Automatic`` below.
```
HazardyKnusperkeks: You are introducing a deprecated option? That's wild.

For the love of god, please review your AI generated code: https://llvm.org/docs/AIToolPolicy.html

clang/docs/ClangFormatStyleOptions.rst
```diff
@@ -3659,6 +3659,173 @@ the configuration (without a prefix: ``Auto``).
...
+      class A {
+        int f() { return 0; };
+      };
+      int f();
+      int f() { return 1; }
+      int
+      LongName::AnotherLongName();
+
+  * ``RTBS_ExceptShortType`` (in configuration: ``ExceptShortType``)
+    Same as ``Automatic`` above, except that there is no break after short
```
HazardyKnusperkeks: Except there's no exception. This is exactly as `Automatic`.

clang/docs/ClangFormatStyleOptions.rst
```diff
@@ -3659,6 +3659,173 @@ the configuration (without a prefix: ``Auto``).
...
+  i.e. break before the return type so that any preceding tokens
+  (function/storage specifiers such as ``static``, ``inline``,
+  ``extern``, ``virtual``, ``explicit``, ``constexpr``, ``friend``;
+  C++11 attribute groups; ``__attribute__((...))`` and ``__declspec(...)``
+  groups; ``alignas(...)``; and identifiers configured via
+  ``AttributeMacros``) are placed on the preceding line(s). The
+  remainder of the line layout is unaffected.
+  Constructors and destructors have no return type and are not affected.
+  Reuses the values of ``ReturnTypeBreakingStyle``; ``Automatic`` and
+  ``ExceptShortType`` behave the same as ``None`` for this option.
```
HazardyKnusperkeks: The first sentence is obvious. And do you really think the remainder has to be in the user facing documentation?

Add your own enum.

clang/include/clang/Format/Format.h
```diff
@@ -2456,6 +2456,65 @@ struct FormatStyle {
   /// \version 16
   BreakBeforeInlineASMColonStyle BreakBeforeInlineASMColon;
 
+  /// The function declaration/definition return type breaking style to use,
```
HazardyKnusperkeks: That's a bit verbose. If such a simple option needs that much text, there's something wrong.

HazardyKnusperkeks (changes requested):

clang/docs/ClangFormatStyleOptions.rst
dlebed: My bad, missed this. I'll address all the comments.

clang/lib/Format/TokenAnnotator.cpp
dlebed: Indeed, I've cleaned it up, also added reference to AttributeMacros for custom macros with attributes which is very common.

dlebed:
@HazardyKnusperkeks, thanks for your comments. I've polished the PR, now it should be in better shape.

clang/lib/Format/TokenAnnotator.cpp
```diff
@@ -4062,6 +4062,60 @@ bool TokenAnnotator::mustBreakForReturnType(const AnnotatedLine &Line) const {
...
+      Opener = Opener->Next;
+    if (!Opener || !Opener->MatchingParen)
+      return nullptr;
+    Tok = Opener->MatchingParen->Next;
+  }
+
+  while (Tok) {
+    if (isReturnTypePrefixSpecifier(*Tok) ||
+        Tok->isOneOf(tok::kw___attribute, tok::kw___declspec) ||
+        Tok->is(TT_AttributeMacro)) {
```
HazardyKnusperkeks:
```suggestion
        Tok->isOneOf(tok::kw___attribute, tok::kw___declspec, TT_AttributeMacro)) {
```

clang/lib/Format/TokenAnnotator.cpp
```diff
@@ -4062,6 +4062,60 @@ bool TokenAnnotator::mustBreakForReturnType(const AnnotatedLine &Line) const {
...
+
+  return false;
+}
+
+static FormatToken *findReturnTypeStart(const AnnotatedLine &Line) {
+  auto *Tok = Line.getFirstNonComment();
+  if (!Tok)
+    return nullptr;
+
+  if (Tok->is(tok::kw_template)) {
```
HazardyKnusperkeks, line 4090: You similar have to skip requires clauses.

clang/lib/Format/TokenAnnotator.cpp
```diff
@@ -4180,6 +4234,18 @@ void TokenAnnotator::calculateFormattingInformation(AnnotatedLine &Line) const {
     }
   }
 
+  if (Line.MightBeFunctionDecl && LineIsFunctionDeclaration &&
+      mustBreakBeforeReturnType(Line)) {
+    if (auto *ReturnTypeStart = findReturnTypeStart(Line);
+        ReturnTypeStart && ReturnTypeStart != FirstNonComment &&
+        ReturnTypeStart->isNot(TT_FunctionDeclarationName) &&
+        ReturnTypeStart->isNot(TT_CtorDtorDeclName) &&
+        ReturnTypeStart->isNot(tok::tilde)) {
```
HazardyKnusperkeks:
```suggestion
        ReturnTypeStart->isNoneOf(TT_FunctionDeclarationName, TT_CtorDtorDeclName, tok::tilde)) {
```

clang/unittests/Format/FormatTest.cpp
```diff
@@ -10594,6 +10594,173 @@ TEST_F(FormatTest, ReturnTypeBreakingStyle) {
   verifyFormat("void foo (int a, int b);", Style);
 }
 
+TEST_F(FormatTest, BreakBeforeReturnType) {
+  FormatStyle Style = getLLVMStyle();
+
+  EXPECT_EQ(Style.BreakBeforeReturnType, FormatStyle::BBRTS_None);
+  verifyFormat("static inline void myfun(void);", Style);
```
HazardyKnusperkeks: That can actually be dropped. The setting of `None` is tested in every other test case.

clang/unittests/Format/FormatTest.cpp
```diff
@@ -10594,6 +10594,173 @@ TEST_F(FormatTest, ReturnTypeBreakingStyle) {
...
+               "  friend\n"
+               "  int f();\n"
+               "};",
+               Style);
+
+  verifyFormat("static int x = 0;", Style);
+  verifyFormat("static const char *msg;", Style);
+
+  verifyFormat("static\n"
+               "auto f() -> int;",
```
HazardyKnusperkeks, line 10647: The documentation should make it clear, that trailing return types are not affected.

clang/unittests/Format/FormatTest.cpp
```diff
@@ -10594,6 +10594,173 @@ TEST_F(FormatTest, ReturnTypeBreakingStyle) {
...
+               Style);
+
+  Style.ColumnLimit = 50;
+  verifyFormat("__attribute__((always_inline)) static inline\n"
```
HazardyKnusperkeks, line 10651: Here too should the two argument variant be the right choice.

HazardyKnusperkeks (commented):
Please add changes as single commits, they will be squashed on merge. This allows me to review a diff, otherwise is the github UI not capable of that.

HazardyKnusperkeks (approved):

PR #197269, opened 2026-05-12
[CIR] Allow implicit truncation in CharacterLiteral codegen

adams381:
`VisitCharacterLiteral` is calling the `IntAttr::get(Type, int64_t)`
builder, which builds the underlying `APInt` with
`isSigned=type.isSigned()` and no implicit truncation. That asserts the
moment a high-bit character literal like `'\xFF'` (which the AST stores
as `CharacterLiteral 'int' 4294967295`) comes through with `int` as the
destination type:

```
APInt.h:121: Assertion `llvm::isIntN(BitWidth, val) &&
                       "Value is not an N-bit signed value"' failed.
```

Classic CodeGen passes `IsSigned=false, ImplicitTrunc=true` here, with a
comment that says character literals are stored unsigned even for signed
char. Just mirror that: build the `APInt` explicitly with those flags
and hand it to `IntAttr::get`. After this the CIR const folder produces
`#cir.int<-1> : !s8i` for `signed char c = '\xFF';` and the lowered
LLVM is identical to OGCG.

Found while sweeping SPEC CPU 2026 with `-fclangir` -- hits cpython
(`Modules/_decimal/_decimal.c`) and omnetpp (yxml + 8 other TUs).
One-liner repro: `void f(void) { signed char c = '\xFF'; }`.

Test covers `\xFF`, `\x80`, `'A'`, and the unsigned-char destination,
all with the CIR + LLVM + OGCG triple.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenExprScalar.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExprScalar.cpp
@@ -247,8 +247,7 @@ public:
   mlir::Value VisitCharacterLiteral(const CharacterLiteral *e) {
     mlir::Type ty = cgf.convertType(e->getType());
     // Character literals are always stored in an unsigned (even for signed
-    // char), so allow implicit truncation here.  Mirrors classic CodeGen
-    // in CGExprScalar.cpp's VisitCharacterLiteral.
+    // char), so allow implicit truncation here.
     auto intTy = mlir::cast<cir::IntTypeInterface>(ty);
     llvm::APInt apValue(intTy.getWidth(), e->getValue(),
                         /*isSigned=*/false, /*implicitTrunc=*/true);
--- a/clang/lib/CIR/CodeGen/CIRGenExprScalar.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExprScalar.cpp
@@ -246,8 +246,14 @@ public:
 
   mlir::Value VisitCharacterLiteral(const CharacterLiteral *e) {
     mlir::Type ty = cgf.convertType(e->getType());
-    auto init = cir::IntAttr::get(ty, e->getValue());
-    return cir::ConstantOp::create(builder, cgf.getLoc(e->getExprLoc()), init);
+    // Character literals are always stored in an unsigned (even for signed
+    // char), so allow implicit truncation here.  Mirrors classic CodeGen
+    // in CGExprScalar.cpp's VisitCharacterLiteral.
+    auto intTy = mlir::cast<cir::IntTypeInterface>(ty);
+    llvm::APInt apValue(intTy.getWidth(), e->getValue(),
+                        /*isSigned=*/false, /*implicitTrunc=*/true);
+    return cir::ConstantOp::create(builder, cgf.getLoc(e->getExprLoc()),
+                                   cir::IntAttr::get(ty, apValue));
   }
 
   mlir::Value VisitCXXBoolLiteralExpr(const CXXBoolLiteralExpr *e) {
--- a/clang/test/CIR/CodeGen/character-literal.c
+++ b/clang/test/CIR/CodeGen/character-literal.c
@@ -3,7 +3,7 @@
 // RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-llvm %s -o %t-cir.ll
 // RUN: FileCheck --check-prefix=LLVM --input-file=%t-cir.ll %s
 // RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -emit-llvm %s -o %t.ll
-// RUN: FileCheck --check-prefix=OGCG --input-file=%t.ll %s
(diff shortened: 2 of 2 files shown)


clang/lib/CIR/CodeGen/CIRGenExprScalar.cpp
```diff
@@ -246,8 +246,14 @@ class ScalarExprEmitter : public StmtVisitor<ScalarExprEmitter, mlir::Value> {
 
   mlir::Value VisitCharacterLiteral(const CharacterLiteral *e) {
     mlir::Type ty = cgf.convertType(e->getType());
-    auto init = cir::IntAttr::get(ty, e->getValue());
-    return cir::ConstantOp::create(builder, cgf.getLoc(e->getExprLoc()), init);
+    // Character literals are always stored in an unsigned (even for signed
+    // char), so allow implicit truncation here.  Mirrors classic CodeGen
```
erichkeane: Remove the 'mirrors' comment, thats only useful for review.

clang/test/CIR/CodeGen/character-literal.c
```diff
@@ -0,0 +1,74 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck --input-file=%t.cir %s --check-prefix=CIR
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-llvm %s -o %t-cir.ll
+// RUN: FileCheck --input-file=%t-cir.ll %s --check-prefix=LLVM
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -emit-llvm %s -o %t.ll
+// RUN: FileCheck --input-file=%t.ll %s --check-prefix=OGCG
```
erichkeane:
```suggestion
// RUN: FileCheck --input-file=%t.ll %s --check-prefix=LLVM
```

See previous comments on past reviews.

clang/test/CIR/CodeGen/character-literal.c
```diff
@@ -0,0 +1,74 @@
...
+
+// A CharacterLiteral '\xFF' has AST type `int` and value 4294967295
```
erichkeane: Same here, we don't do these comments typically.  Please stop generating these.

clang/lib/CIR/CodeGen/CIRGenExprScalar.cpp
adams381: Dropped the "Mirrors classic CodeGen" lines.  Kept the two preceding lines explaining why `isSigned=false, implicitTrunc=true` is needed.

clang/test/CIR/CodeGen/character-literal.c
adams381: Rewrote all four FileCheck lines to the canonical `--check-prefix=NAME --input-file=PATH %s` order.  Read the literal `suggestion` as a typo (`LLVM` instead of `OGCG` on line 6) since applying it verbatim would have left the test without OGCG coverage.

clang/test/CIR/CodeGen/character-literal.c
adams381: Preamble removed.

clang/test/CIR/CodeGen/character-literal.c
```diff
@@ -0,0 +1,66 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck --check-prefix=CIR --input-file=%t.cir %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-llvm %s -o %t-cir.ll
+// RUN: FileCheck --check-prefix=LLVM --input-file=%t-cir.ll %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -emit-llvm %s -o %t.ll
+// RUN: FileCheck --check-prefix=OGCG --input-file=%t.ll %s
```
erichkeane:
```suggestion
// RUN: FileCheck --check-prefix=LLVM --input-file=%t.ll %s
```

adams381: Done; collapsed the now-duplicate OGCG check blocks into the shared LLVM prefix on the same amend.

erichkeane: Did you forget to push? This change hasn't been made above.

erichkeane (changes requested):

clang/test/CIR/CodeGen/character-literal.c
adams381: Pushed now — line 6 uses the shared LLVM prefix on `%t.ll` and the duplicate OGCG-* blocks are gone.

erichkeane (approved):

PR #197270, opened 2026-05-12
[clang][deps] Move logic out of the PP callback

jansvoboda11:
This PR moves the main scanner logic from `ModuleDepCollectorPP::EndOfMainFile()` to the new `ModuleDepCollector::run()`. In a follow-up PR, this will allow invoking the main logic with different `DependencyConsumer` objects and reusing the `ModuleDepCollector` state, skipping some work on subsequent calls.


Bigcheese (approved):
lgtm

PR #197272, opened 2026-05-12
[Bazel]: Pull from Bazel-Central-Registry for third party deps.

UebelAndre:
The majority of these dependencies are available in the [Bazel-Central-Registry](https://github.com/bazelbuild/bazel-central-registry) (BCR) and to improve build performance for bzlmod users, llvm-project should pull from the BCR to consolidate targets.


keith (approved):
change looks good but im interested to hear from others on if this will be disruptive for them

UebelAndre:
I'm so sorry, I accidentally closed this PR. A new one has been opened here: https://github.com/llvm/llvm-project/pull/197316

PR #192876, opened 2026-04-19
[VPlan] Move IV predicate handling to VPlan.

fhahn:
Update induction SCEV predicate handling to track the SCEV predicates required for a induction in IVDescriptor directly. This allows moving the AllowedExit check to VPlan and tracking predicates per induction allows for more accurate decisions in follow-up patches.

For example, in some cases, the same phi could either be a fixed-order recurrence or an induction with predicates. This will allow more aggressively converting phis to fixed-order recurrences in VPlan, while still being able to gracefully handle the fall back when predicated inductions are needed.

Note that this also moves checking the SCEV predicate thresholds. This catches additional cases we missed before, because predicates were added after we performed the check in Legal.

PR #197276, opened 2026-05-12
[VPlan] Construct VPlan1 once, share across buildVPlans calls.

fhahn:
Extract the VF-independent VPlan1 setup pipeline (header phis, simplification, early-exit handling, middle check, loop regions, tail folding, mask introduction) into a new helper tryToBuildVPlan1().

Construct the initial Vlan1 once, and pass to repeated buildVPlans calls.

Note that this means we need to move collectInLoopReductions up. We not may construct VPlan1 on code paths where we did not before, because we failed UserVF validation/selection, but I think that should be fine as this makes the overall code simpler and the UserVF code paths are for testing.


fhahn (commented):
ping

llvm/lib/Transforms/Vectorize/LoopVectorizationPlanner.h
```diff
@@ -896,7 +896,10 @@ class LoopVectorizationPlanner {
   /// Build VPlans for power-of-2 VF's between \p MinVF and \p MaxVF inclusive,
   /// according to the information gathered by Legal when it checked if it is
   /// legal to vectorize the loop.
-  void buildVPlans(ElementCount MinVF, ElementCount MaxVF);
+  void buildVPlans(VPlan &VPlan0, ElementCount MinVF, ElementCount MaxVF);
+
+  /// Build base VPlan0 with scalar transformations applied.
+  VPlanPtr buildVPlan0();
```
ayalz: Better place first, prefix with tryTo, and call this VPlan1?
```suggestion
  /// Build an initial VPlan, with HCFG wrapping the original scalar loop and scalar transformations applied. Returns null if an initial VPlan cannot be built.
  VPlanPtr tryToBuildVPlan1();
  
  /// Build a VPlan using VPRecipes according to the information gathered by
  /// Legal and VPlan-based analysis. For outer loops, performs basic recipe
  /// conversion only. For inner loops, \p Range's largest included VF is
  /// restricted to the maximum VF the returned VPlan is valid for. If no VPlan
  /// can be built for the input range, set the largest included VF to the
  /// maximum VF for which no plan could be built. Each VPlan is built starting
  /// from a copy of \p InitialPlan, which is a plain CFG VPlan wrapping the
  /// original scalar loop.
  VPlanPtr tryToBuildVPlan(VPlanPtr InitialPlan, VFRange &Range);

  /// Build VPlans for power-of-2 VF's between \p MinVF and \p MaxVF inclusive,
  /// according to the information gathered by Legal when it checked if it is
  /// legal to vectorize the loop.
  void buildVPlans(VPlan &VPlan0, ElementCount MinVF, ElementCount MaxVF);
```

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -5640,18 +5646,19 @@ void LoopVectorizationPlanner::plan(ElementCount UserVF, unsigned UserIC) {
     } else {
       assert(isPowerOf2_32(UserVF.getKnownMinValue()) &&
              "VF needs to be a power of two");
-      // Collect the instructions (and their associated costs) that will be more
-      // profitable to scalarize.
```
ayalz, line 5791: Worth retaining the comment as it applies to collectNonVectorizedAndSetWideningDecisions() below?

llvm/test/Transforms/LoopVectorize/AArch64/store-costs-sve.ll
```diff
@@ -158,7 +158,7 @@ define void @trunc_store(ptr %dst, ptr %src, i16 %x) #1 {
 ; DEFAULT-NEXT:    [[VEC_EPILOG_RESUME_VAL:%.*]] = phi i64 [ [[N_VEC]], [[VEC_EPILOG_ITER_CHECK]] ], [ 0, [[VECTOR_MAIN_LOOP_ITER_CHECK]] ]
 ; DEFAULT-NEXT:    [[BROADCAST_SPLATINSERT4:%.*]] = insertelement <8 x i16> poison, i16 [[X]], i64 0
 ; DEFAULT-NEXT:    [[BROADCAST_SPLAT5:%.*]] = shufflevector <8 x i16> [[BROADCAST_SPLATINSERT4]], <8 x i16> poison, <8 x i32> zeroinitializer
-; DEFAULT-NEXT:    [[TMP8:%.*]] = load i64, ptr [[SRC]], align 8, !alias.scope [[META13:![0-9]+]]
+; DEFAULT-NEXT:    [[TMP8:%.*]] = load i64, ptr [[SRC]], align 8, !alias.scope [[META6]]
```
ayalz, line 160: Change in metadata intentional?

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
ayalz, line 6801: There's now a slight discrepancy between `LVP::buildVPlan0()` and `VPlanTransform::buildVPlan0()`. Should the former be renamed to produce `InitialVPlan` or `VPlan1`, keeping `VPlan0` to be the plain CFG wrapping of the original loop?

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -5599,12 +5599,17 @@ void LoopVectorizationPlanner::plan(ElementCount UserVF, unsigned UserIC) {
   if (!MaxFactors) // Cases that should not to be vectorized nor interleaved.
     return;
 
+  Config.collectInLoopReductions();
+
   if (!OrigLoop->isInnermost()) {
     // For outer loops, computeMaxVF returns a single non-scalar VF; build a
     // plan for only that VF.
```
ayalz: Independent nit:
```suggestion
    // plan for that VF only.
```

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -5632,6 +5637,7 @@ void LoopVectorizationPlanner::plan(ElementCount UserVF, unsigned UserIC) {
 
   ElementCount MaxUserVF =
       UserVF.isScalable() ? MaxFactors.ScalableVF : MaxFactors.FixedVF;
+
```
ayalz: Can add empty line independently.

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -5640,18 +5646,19 @@ void LoopVectorizationPlanner::plan(ElementCount UserVF, unsigned UserIC) {
...
-      Config.collectInLoopReductions();
       CM.collectNonVectorizedAndSetWideningDecisions(UserVF);
       ElementCount EpilogueUserVF =
           ElementCount::getFixed(EpilogueVectorizationForceVF);
-      if (EpilogueUserVF.isVector() &&
-          ElementCount::isKnownLT(EpilogueUserVF, UserVF)) {
+      bool UseEpilogueUserVF = EpilogueUserVF.isVector() &&
+                               ElementCount::isKnownLT(EpilogueUserVF, UserVF);
+      if (UseEpilogueUserVF)
         CM.collectNonVectorizedAndSetWideningDecisions(EpilogueUserVF);
```
ayalz, line 5802: Could this (continue to) be placed below under `if (UseEpilogueUserVF)` next to `buildVPlans(*VPlan0, EpilogueUserVF, EpilogueUserVF);`?

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -5599,12 +5599,17 @@ void LoopVectorizationPlanner::plan(ElementCount UserVF, unsigned UserIC) {
   if (!MaxFactors) // Cases that should not to be vectorized nor interleaved.
     return;
 
+  Config.collectInLoopReductions();
+
```
ayalz, line 5753: Try to build VPlan0 (or rather VPlan1) once, here at the outset after collecting in-loop reductions, and bail if failed?
```suggestion
  auto VPlan1 = tryTobuildVPlan1();
  if (!VPlan1)
    return;

```

fhahn: Unfortunately there's a subtle dependency  that means `collectNonVectorizedAndSetWideningDecisions` can impact the result of `tryTobuildVPlan1`: it may add predicates (e.g. %var == 1), which will get materialized as part of `tryTobuildVPlan1`

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: Fixed here, thanks

llvm/lib/Transforms/Vectorize/LoopVectorizationPlanner.h
fhahn: reordered, thanks!

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: dropped thanks

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: Yep retained, thanks

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: Similar to  https://github.com/llvm/llvm-project/pull/197276/changes#r3281497406, `collectNonVectorizedAndSetWideningDecisions` may add predicates which are included in VPlan1.

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: Yep, updated/renamed to VPlan1, thanks!

llvm/test/Transforms/LoopVectorize/AArch64/store-costs-sve.ll
fhahn: Yes, previously we would construct different noalias metadata for both calls of `buildVPlans`, whereas now we just create the medata once, and share if for all plans across `buildVPlans` calls

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
ayalz: Can the creation of these VF-agnostic predicates, which affect VPlan1 (and VPlan0), be factored out of collectNVASWD(VF) - which applies to non-scalar VF's only, possibly into tryToBuildVPlan1()?

fhahn: We could do that - they are coming from the VF-independent `isConsecutivePtr` calls used. We would need to  walk all memory instructions in the loop and call `isConsecutivePtr` on their pointers to populate the predicates. Not sure if that's preferred?

ayalz: It's arguably better to be explicit and clear about such implicit subtle dependencies, separating the VF-agnostic decisions that take part in building a common VPlan1 (or rather VPlan0) from those taken later per VF>1 range? The compile-time concern associated with multiple scans is worth analyzing and addressing separately, e.g., by perhaps fusing multiple consecutive transforms together, rather than fusing together distinctly staged decisions. Sounds reasonable?

ayalz: Furthermore, per-VF decisions could potentially migrate to take place based on VPlan1 rather than on original IR, in contrast to initial VF-agnostic decisions that affect VPlan0 construction?

fhahn: I added the code to populate the predicates up front for now. Migrating to be based on VPlan would require a bit more extra work.

Note that the order we populate the predicates can impact vectorization decisions, doing up-front surfaced an additional improvement

llvm/test/Transforms/LoopVectorize/version-stride-with-integer-casts.ll
```diff
@@ -320,29 +320,16 @@ define void @test_versioned_with_non_ex_use(i32 %offset, ptr noalias %dst.1, ptr
...
-; CHECK-NEXT:    [[TMP14:%.*]] = getelementptr i8, ptr [[DST_1]], i32 [[TMP13]]
-; CHECK-NEXT:    [[TMP16:%.*]] = getelementptr i8, ptr [[DST_1]], i32 [[TMP15]]
+; CHECK-NEXT:    [[TMP9:%.*]] = trunc i64 [[INDEX]] to i32
+; CHECK-NEXT:    [[TMP17:%.*]] = mul i32 [[TMP9]], [[ADD]]
 ; CHECK-NEXT:    [[TMP18:%.*]] = getelementptr i8, ptr [[DST_1]], i32 [[TMP17]]
-; CHECK-NEXT:    store i32 0, ptr [[TMP12]], align 8
-; CHECK-NEXT:    store i32 0, ptr [[TMP14]], align 8
-; CHECK-NEXT:    store i32 0, ptr [[TMP16]], align 8
-; CHECK-NEXT:    store i32 0, ptr [[TMP18]], align 8
+; CHECK-NEXT:    store <4 x i32> zeroinitializer, ptr [[TMP18]], align 8
```
ayalz, line 329: Nice cleanup!

llvm/lib/Transforms/Vectorize/LoopVectorizationPlanner.h
```diff
@@ -938,7 +943,7 @@ class LoopVectorizationPlanner {
   /// Build VPlans for power-of-2 VF's between \p MinVF and \p MaxVF inclusive,
   /// according to the information gathered by Legal when it checked if it is
   /// legal to vectorize the loop.
```
ayalz:
```suggestion
/// based on \p VPlan1 and according to the information gathered by Legal
/// when it checked if it is legal to vectorize the loop.
```

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -5614,12 +5614,17 @@ void LoopVectorizationPlanner::plan(ElementCount UserVF, unsigned UserIC) {
...
   if (!OrigLoop->isInnermost()) {
     // For outer loops, computeMaxVF returns a single non-scalar VF; build a
-    // plan for only that VF.
+    // plan for that VF only.
+    auto VPlan1 = tryToBuildVPlan1();
+    if (!VPlan1)
+      return;
```
ayalz: Predicates for consecutive accesses are not added (yet) to outer-loop vectorization, but better factor that into allowRuntimeSCEVChecks(), or under collectUnitStridePredicates() - see below.

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -5645,6 +5650,19 @@ void LoopVectorizationPlanner::plan(ElementCount UserVF, unsigned UserIC) {
...
 
+  // Make sure predicates for consecutive accesses are added to PSE up-front.
+  if (Legal->allowRuntimeSCEVChecks() &&
+      (MaxFactors.FixedVF.isVector() || MaxFactors.ScalableVF.isVector())) {
+    for (BasicBlock *BB : OrigLoop->blocks())
+      for (Instruction &I : *BB)
+        if (Value *Ptr = getLoadStorePointerOperand(&I))
+          Legal->isConsecutivePtr(getLoadStoreType(&I), Ptr);
+  }
+
```
ayalz: Better wrapped in, say,
```
  // Cases that may be vectorized may be optimized by unit stride predicates.
  if (MaxFactors.FixedVF.isVector() || MaxFactors.ScalableVF.isVector())
    Legal->collectUnitStridePredicates();
```
which internally checks if runtime SCEV checks are allowed, if loop vectorized is innermost, and takes care of collecting the predicates by calling isConsecutivePtr(). Could this then take place upfront, along with creating VPlan1 and bailing out if it fails, as suggested earlier.

This raises a (follow-up/TODO) point: if we end up unrolling only, could these predicates be dropped, or should adding of these predicated clone VPlan1 into one serving all VF>1 cases (VPlan1.1?) and the other serving VF=1 only (original VPlan1, or VPlan1.0)?

llvm/lib/Transforms/Vectorize/LoopVectorizationPlanner.h
fhahn: Updated thanks

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: Thanks, moved to `collectUnitStridePredicates`. I left the place where it is called as-is for now, as otherwise we would add unneeded predicates when forcing VF = 1 (or pass `MaxFactors` through). 

I added a TODO regarding dropping predicates when not used (somewhat similar to https://github.com/llvm/llvm-project/pull/192876)

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: moved into `collectUnitStridePredicates`, thanks

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
ayalz: The `if (MaxFactors.FixedVF.isVector() || MaxFactors.ScalableVF.isVector())` guard should also take care of forcing VF = 1, rather than pass `MaxFactors` through?

ayalz (approved):
This LGTM, thanks!

The MaxFactors.*.isVector() guard should catch all unvectorized cases, including UserVF=1, permitting building VPlan1 once, at the outset.

collectUnitStridePredicates() could be introduced separately, independent of introducing VPlan1, for more gradual refactoring.

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: Yep, it takes care of the cases were VF is forced to 1.

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -5654,6 +5659,10 @@ void LoopVectorizationPlanner::plan(ElementCount UserVF, unsigned UserIC) {
   if (MaxFactors.FixedVF.isVector() || MaxFactors.ScalableVF.isVector())
     Legal->collectUnitStridePredicates();
 
+  auto VPlan1 = tryToBuildVPlan1();
+  if (!VPlan1)
+    return;
+
```
ayalz: So could the creation of VPlan1 (along with the conditional collection of unit stride predicates) take place once, at the outset, after collecting in loop reductions and checking that MaxFactors exist? It's fine to move as a separate follow up if preferred.

ayalz (approved):
This LGTM, thanks!

PR #197279, opened 2026-05-12
[SLP] Skip no-return blocks in spill cost analysis

alexey-bataev:
Treat basic blocks containing a non-terminator no-return call as
dead-end paths in getSpillCost. Execution doesn't exit such blocks,
so their non-vec calls don't keep loop-invariant vector values live.
Terminator no-return calls (invoke/callbr) are excluded because
their unwind/indirect successors remain reachable.


bababuck (approved):
LGTM

RKSimon (approved):

PR #197280, opened 2026-05-12
[mlir][AMDGPU] Canonicalize masks on global_load_async_to_lds

krzysz00:
If the mask is always true, remove the mask operand (there are patterns that key off the presence of the lack of a mask operand to know when they can be more aggressive). If the mask is always false, just go ahead and delete the op as it won't write anythig.

AI: I described the patterns, Codex 5.5 wrote them

### DIFF
--- a/mlir/lib/Dialect/AMDGPU/IR/AMDGPUOps.cpp
+++ b/mlir/lib/Dialect/AMDGPU/IR/AMDGPUOps.cpp
@@ -1065,6 +1065,31 @@ LogicalResult GlobalLoadAsyncToLDSOp::verify() {
   return success();
 }
 
+static LogicalResult
+foldGlobalLoadAsyncToLDSConstantMask(GlobalLoadAsyncToLDSOp op,
+                                     PatternRewriter &rewriter) {
+  Value mask = op.getMask();
+  if (!mask)
+    return failure();
+
+  APInt maskValue;
+  if (!matchPattern(mask, m_ConstantInt(&maskValue)))
+    return failure();
+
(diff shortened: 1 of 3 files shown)


mlir/lib/Dialect/AMDGPU/IR/AMDGPUOps.cpp
```diff
@@ -1065,6 +1065,31 @@ LogicalResult GlobalLoadAsyncToLDSOp::verify() {
...
+  }
+
+  rewriter.modifyOpInPlace(op, [&]() { op.getMaskMutable().clear(); });
+  return success();
+}
+
+void GlobalLoadAsyncToLDSOp::getCanonicalizationPatterns(
+    RewritePatternSet &results, MLIRContext *context) {
+  results.add(foldGlobalLoadAsyncToLDSConstantMask);
+}
```
amd-eochoalo, line 1091: I would split this into two patterns, one for all-zeros in canonicalization and one for all-ones in a fold.

amd-eochoalo (approved):

mlir/lib/Dialect/AMDGPU/IR/AMDGPUOps.cpp
krzysz00: I figure the fold is unlikely enough that I don't want to be checking it outside of a `-canonicalize` but I could go either way here

amd-eochoalo: Your call :)

efric (approved):

PR #197281, opened 2026-05-12
[MLIR][XeGPU][VectorToXeGPU] Fixed transposed transfer_read/write

AndreyPavlenko:
The problem was that adjustStridesForPermutation() permutes strides, but they are required to compute base offset from transfer read indices. Added optional argument to computeMemrefMeta() to preserve the original strides.

### DIFF
--- a/mlir/lib/Conversion/VectorToXeGPU/VectorToXeGPU.cpp
+++ b/mlir/lib/Conversion/VectorToXeGPU/VectorToXeGPU.cpp
@@ -256,13 +256,8 @@ computeMemrefMeta(OpType xferOp, PatternRewriter &rewriter) {
       offsetVal = meta.getOffset();
   }
 
-  if constexpr (llvm::is_one_of<std::decay_t<OpType>, vector::TransferReadOp,
-                                vector::TransferWriteOp>::value) {
-    AffineMap permMap = xferOp.getPermutationMap();
-    // Adjust strides according to the permutation map (e.g., for transpose)
-    adjustStridesForPermutation(permMap, strides);
-  }
-
+  // Strides are returned in original memref order; permutation is applied in
+  // computeOffsets only where offsets are indexed in vector order.
   return {strides, offsetVal};
 }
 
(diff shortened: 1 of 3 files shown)


Garra1980:
cc @dchigarev

mlir/lib/Conversion/VectorToXeGPU/VectorToXeGPU.cpp
```diff
@@ -288,7 +292,7 @@ computeMemrefMeta(OpType xferOp, PatternRewriter &rewriter) {
 //   %offsets =  memref_offset + orig_offset + local_offsets
 static Value computeOffsets(VectorTransferOpInterface xferOp,
```
charithaintc, line 292: why this has two stride arguments? can't we only use the original strides everywhere.

AndreyPavlenko: We need the permuted strides when reading a transposed vector. It's required for the vector.steps calculation.

mlir/lib/Conversion/VectorToXeGPU/VectorToXeGPU.cpp
```diff
@@ -195,7 +195,8 @@ template <
         std::decay_t<OpType>, vector::TransferReadOp, vector::TransferWriteOp,
         vector::GatherOp, vector::ScatterOp>::value>>
 static std::pair<SmallVector<Value>, Value>
-computeMemrefMeta(OpType xferOp, PatternRewriter &rewriter) {
+computeMemrefMeta(OpType xferOp, PatternRewriter &rewriter,
```
Garra1980: I think adding new argument deserves mentioning in the function comment

mlir/lib/Conversion/VectorToXeGPU/VectorToXeGPU.cpp
Jianhui-Li: Instead of change two function signature and pass two strides to computeOffsets(), you could do the following: 
  1. computeMemrefMeta() no longer permutes
  2. Inside computeOffsets(), it builds a local permutedStrides copy (via
  adjustStridesForPermutation) used only for the local-offset step-vector loop; the base-offset loop keeps the original memref-order strides.

mlir/test/Conversion/VectorToXeGPU/transpose-to-xegpu.mlir
```diff
@@ -0,0 +1,46 @@
+// RUN: mlir-opt %s -convert-vector-to-xegpu -split-input-file | FileCheck %s
+
+// -----
+gpu.module @xevm_module {
```
Jianhui-Li: please combine this test to the existing test file: transfer-read-to-xegpu.mlir

Jianhui-Li (commented):
Confirmed the issue, thanks! Please address the comments

mlir/test/Conversion/VectorToXeGPU/transpose-to-xegpu.mlir
AndreyPavlenko: Done.

Jianhui-Li (approved):
LGTM

charithaintc (approved):

PR #195392, opened 2026-05-02
Add a breakpoint override resolver feature to lldb

jimingham:
This is a general mechanism that allows a custom breakpoint resolver to "override" the resolver lldb would have used.  It is implemented by making a scripted breakpoint resolver class, implementing an extra "overrides_resolver" method.  When lldb has decided what resolver it is going to use for the breakpoint, if there are any override resolvers registered, it will pass the serialized form of the resolver to the `overrides_resolver` method of the registered resolvers.  If any of the override resolvers return true from the overrides_resolver method, that resolver will be used in place of the one lldb would have used.

The original motivation for this feature is to make the support of "debugging a scripting language from the implementation side".  We added the notion of "facade breakpoints" so that a location given in the context of the scripting language can be handled by the emulation code.  But when you are using lldb in some IDE, when "stopped" in the scripting language, the IDE will show the scripting language source.  It is natural in that case for the user to expect the UI gesture to set a file and line breakpoint should work the same way.

To implement that, the implementor would make their facade breakpoint resolver also an override resolver that checks to see if the incoming resolver was a file and line breakpoint for a source file of the scripting language type.  If it is, it returns true, and lldb will use that resolver instead.  Note, this will also work when breakpoints are set on the command line, so you can transparently support "scripting language" breakpoints from the command line as well.

PR #197285, opened 2026-05-12
[lldb][windows] mark test_overrides_resolver_resolver_cmd as XFAIL

charles-zablit:
Follow up to https://github.com/llvm/llvm-project/pull/195392 to mark `test_overrides_resolver_resolver_cmd` as XFAIL on Windows, like `test_overrides_resolver_resolver_python`.


jimingham (approved):
LGTM for now.

Note, the "list" part of the test that is failing is actually common in the API version (that seems to always succeed) and the command line version that is somehow flakey...

It seems like the `breakpoint override add` command must be failing, though it claims to have succeeded (we do check that in the test).

This never fails on macOS - either on my desktop or the bots.  It fails sometimes on Ubuntu, so I'm spinning up an Ubuntu VM to see if I can repro there.

jimingham:
This is affecting the bots, so I'm going to push this now.

PR #197286, opened 2026-05-12
MachOObjectFile: Don't limit `getSectionSize` to object's file size

mlechu:
`MachOObjectFile::getSectionSize` currently bounds its return value based on the size of the file.  This breaks with separate-debuginfo `MH_DSYM` files, where the file size is not necessarily larger than the sections described within it.

This check was added in 2015 (https://github.com/llvm/llvm-project/commit/46e642f8c56022f4218699a7071996e7e0863ba2) to fix a crash in llvm-objdump with malformed files. Better parse-time checking was added later (https://github.com/llvm/llvm-project/commit/c614d283b7aecbc9bee0b7bdb7902d2796b04a8c), so the original failure should not reproduce.

This PR just removes the check, which makes it consistent with other
implementations of `getSectionSize`.  Note the unit test was written with AI.

### DIFF
--- a/llvm/lib/Object/MachOObjectFile.cpp
+++ b/llvm/lib/Object/MachOObjectFile.cpp
@@ -1950,32 +1950,9 @@ uint64_t MachOObjectFile::getSectionIndex(DataRefImpl Sec) const {
 }
 
 uint64_t MachOObjectFile::getSectionSize(DataRefImpl Sec) const {
-  // In the case if a malformed Mach-O file where the section offset is past
-  // the end of the file or some part of the section size is past the end of
-  // the file return a size of zero or a size that covers the rest of the file
-  // but does not extend past the end of the file.
-  uint32_t SectOffset, SectType;
-  uint64_t SectSize;
-
-  if (is64Bit()) {
-    MachO::section_64 Sect = getSection64(Sec);
-    SectOffset = Sect.offset;
-    SectSize = Sect.size;
-    SectType = Sect.flags & MachO::SECTION_TYPE;
-  } else {
-    MachO::section Sect = getSection(Sec);
-    SectOffset = Sect.offset;
-    SectSize = Sect.size;
-    SectType = Sect.flags & MachO::SECTION_TYPE;
-  }
-  if (SectType == MachO::S_ZEROFILL || SectType == MachO::S_GB_ZEROFILL)
-    return SectSize;
-  uint64_t FileSize = getData().size();
-  if (SectOffset > FileSize)
-    return 0;
-  if (FileSize - SectOffset < SectSize)
-    return FileSize - SectOffset;
-  return SectSize;
+  if (is64Bit())
+    return getSection64(Sec).size;
+  return getSection(Sec).size;
 }
 
 ArrayRef<uint8_t> MachOObjectFile::getSectionContents(uint64_t Offset,
--- a/llvm/test/Object/macho-dsym-section-size.test
+++ b/llvm/test/Object/macho-dsym-section-size.test
@@ -1,0 +1,47 @@
+## Show that for MH_DSYM files, MachOObjectFile::getSectionSize returns the
+## size from the section header even when the section's offset is past the
+## end of the file (as the DSYM may be smaller than the original object file).
+
+# RUN: yaml2obj %s -o %t.dsym
+# RUN: %python -c "import sys; f = open(sys.argv[1], 'r+b'); f.truncate(1024); f.close()" %t.dsym
+# RUN: llvm-objdump --section-headers %t.dsym | FileCheck %s
+
+# CHECK:      Sections:
+# CHECK-NEXT: Idx Name   Size     VMA              Type
+# CHECK-NEXT:   0 __text 00000100 0000000000000000 TEXT
+
+--- !mach-o
+FileHeader:
+  magic:           0xFEEDFACF
+  cputype:         0x01000007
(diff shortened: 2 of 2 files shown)


JDevlieghere:
Rather than removing the check, it seems to me like we should:

1. Use the on-disk file size rather than the file size as reported by the Mach header.
2. Update `dsymutil` to patch the header to the new on-disk file size, which presumably is the source of this issue?

mlechu:
I'm not sure I understand.  Currently, `FileSize = getData().size()` is already the on-disk size, just of the wrong file (of the .dSYM rather than the file whose sections it describes).

vtjnash (approved):
LGTM. I guess there isn't a simple way to check if the section is really in this file, and what the real section in the real file looks like (since the file might not exist anymore)?

vtjnash:
@llvm/issue-subscribers-julialang

JDevlieghere (changes requested):
> I'm not sure I understand. Currently, `FileSize = getData().size()` is already the on-disk size, just of the wrong file (of the .dSYM rather than the file whose sections it describes).

I misunderstood the problem, but the problem you're describing here sounds even odder. Why are we reading the section size of the main binary form the dSYM's mach header? What's the code path where those two get commingled? In its current state, it seems like this PR will allow us to read past the end of the file.

mlechu:
I was worried julia was doing something crazy, but I took a look and it's just calling `createObjectFile(path_to_dsym)` and looping through its sections to find one containing a certain address (the loop part actually being copied from llvm's `SymbolizableObjectFile::getModuleSectionIndexForAddress`).  `llvm-objdump --section-headers` also reaches the same code, so the same issue presents itself (this file's size is `0x79f5d3`):

```
% /usr/bin/objdump --version
Apple LLVM version 16.0.0
   (clang-1600.0.26.6)Optimized build.

% /usr/bin/objdump -m --section-headers ./sys.dylib.dSYM/Contents/Resources/DWARF/sys.dylib

Sections:
Idx Name             Size     VMA              Type
  0 __text           0079f5d3 0000000000000814 TEXT
[...]
```

where GNU objdump has a different answer:

```
% objdump --version
GNU objdump (GNU Binutils) 2.44
Copyright (C) 2025 Free Software Foundation, Inc.

% objdump --section-headers ./sys.dylib.dSYM/Contents/Resources/DWARF/sys.dylib

./sys.dylib.dSYM/Contents/Resources/DWARF/sys.dylib:     file format mach-o-arm64

Sections:
Idx Name          Size      VMA               LMA               File off  Algn
  0 .text         00c6f0fc  0000000000000814  0000000000000814  00000000  2**2
                  ALLOC, LOAD, CODE
[...]
```

> it seems like this PR will allow us to read past the end of the file.

I don't see why; shouldn't the caller be responsible for knowing what `getSize` may be used for?  That's already the case for `S_ZEROFILL` and `S_GB_ZEROFILL`. It also appears to be the case for `ELFObjectFile::getSectionSize`.  COFF appears to check for other reasons (file-data size is rounded up from the useful size).

Having the invariant that every `ObjectFile` contain its `Section`s sounds OK to me, it just doesn't seem to be true now.  If this is indeed what we want, then the check I've deleted should be an assertion or error depending on whether we trust parsing, and `createObjectFile` should not produce the `Section`s we're seeing for `MH_DSYM`.  What is the invariant you were thinking of?

If you like, I can just add `MH_DSYM` and `MH_DYLIB_STUB` cases to the `S_ZEROFILL` check and make other greater-than-file-size values throw an error, which seems consistent enough to me.

mlechu:
Polite bump

vtjnash:
As Em points out, this was apparently originally added to avoid reader mistakes elsewhere but it would be the wrong place to implement that since it causes llvm-objdump to print incorrect information and causes split debug info to be described incorrectly. The other formats implement this check more carefullly in getSectionContents, but it looks like MachO ignores the Expected interface there and instead hard-codes true. That can't be fixed without first merging this PR.

https://llvm.org/doxygen/ELFObjectFile_8h_source.html#l00930
https://llvm.org/doxygen/classllvm_1_1object_1_1XCOFFObjectFile.html#a9800b5c811f6cb665c2c118d1ceccb54

It looks like we could fix MachO to match the checks done by the other formats, so that correctly reported lengths don't cause implicit section truncation in `StringRef.substr` instead of the explicit truncation here (currently this PR appears to be a no-op for most readers--with <4GB or ordered sections--because the subsequent implicit behavior matches the current explicit behavior):
https://llvm.org/doxygen/MachOObjectFile_8cpp_source.html#l01981

mlechu:
Bump again; @JDevlieghere I'm still interested in fixing this.  Let me know if you need anything from me.

JDevlieghere:
Ok, I think I finally understand the problem you're trying to solve: a .dSYM deliberately keeps section metadata while dropping much of the section content, so its file is smaller than the sections it describes. And the problem you're trying to solve is that currently we conflate "how big is the section" with "how many bites are in it". I'm supportive of fixing that of course.

My remaining concern is that this now removes the check that the original change was defending against. The parse-time "section extends past the end of the file" check you're relying on is explicitly skipped for MH_DSYM and MH_DYLIB_STUB. After this change getSize() returns that raw size, while getContents() still clamps via substr, and various callers rely on getSize while going through the getContents buffer. 

So what I'd like to see is (1) a test that exercises this (i.e. a crash or trips up with Asan) and (2) a fix for that.

PR #197288, opened 2026-05-12
[RISCV][XCV] Fix incorrect llvm.ctlz lowering to cv.fl1

### LINKED ISSUE #197006: [RISCV][XCV] Incorrect lowering of llvm.ctlz to cv.fl1 (XCVbitmanip)
### Summary

The XCVbitmanip pattern set in `llvm/lib/Target/RISCV/RISCVInstrInfoXCV.td` lowers `llvm.ctlz.i32` directly to the `cv.fl1` instruction:
```
def : PatGpr<ctlz, CV_FL1, i32>;
```

This is semantically incorrect: `cv.fl1` and `llvm.ctlz` compute different quantities.

### Semantics mismatch

Per the [OpenHW CORE-V ISA specification](https://docs.openhwgroup.org/projects/cv32e41p-user-manual/instruction_set_extensions.html#bit-manipulation-operations) for XCVbitmanip:
- `cv.fl1 rD, rs1` returns the **bit position** of the most-significant '1' in rs1, in [0..31] for non-zero rs1, and 32 when rs1 == 0.

Per the LLVM LangRef:
- `llvm.ctlz.i32(%x, i1 is_zero_poison)` returns the **count** of leading zero bits, in [0..32], and 32 when %x == 0 (when is_zero_poison is false).

The two quantities are related by
```
ctlz(x) = (x == 0) ? 32 : 31 - fl1(x)
```

They coincide on x == 0 (both yield 32) but disagree on every non-zero input by up to 31. The current pattern silently produces wrong results for every non-zero input.

### Reproducer

```c
// clang --target=riscv32 -march=rv32i_xcvbitmanip -O2 -c test.c
unsigned my_clz(unsigned x) {
    return __builtin_clz(x);   // is_zero_poison=true (UB on x==0)
}
```

Current output (clang main):
```assembly
my_clz:
  cv.fl1 a0, a0
  ret
```

Expected output (should compute 31 - fl1(x), e.g.):
```assembly
my_clz:
  cv.fl1 a0, a0
  xori   a0, a0, 31
  ret
```

For example, `__builtin_clz(0x80000000)` should return `0` but currently returns `31`; `__builtin_clz(1)` should return `31` but currently returns `0`.

### Real-world impact

This miscompiles every call to `__builtin_clz` / `__builtin_clzl` in the program, including the implementations inside compiler-rt.
In particular, `compiler-rt/lib/builtins/udivmoddi4.c` uses `__builtin_clz` to normalise the divisor in the Knuth-D long division shift loop:
```c
sr = (32 + __builtin_clz(d.s.low)) - __builtin_clz(n.s.low);
```

With the wrong values, `sr` is off by up to 62, and `__udivdi3` returns wrong quotients silently. Any user code that performs a 64-bit unsigned division will receive incorrect results.

I observed this in an RTL simulation on a CV32E40P core: the firmware was performing
```c
uint64_t nco = 20000000ULL / ((uint64_t)115200 << 4);  // expected: 2
```

and the actual returned value was 0, because the compiler-rt's __udivdi3 produced 0 instead of 2. The simulation trace shows the two cv.fl1 instructions inside __udivdi3 returning bit positions (0x14 and 0x16) where the algorithm expected leading-zero counts (0x0b and 0x09).

### Symmetric pattern

The companion pattern
```
def : PatGpr<cttz, CV_FF1, i32>;
```
happens to be correct: `cv.ff1(0) == 32 == cttz(0, false)`, and they coincide for non-zero inputs as well (both return the trailing-zero count). No action is needed there, though it could also be gated on `cttz_zero_undef` for symmetry.

### Proposed fix

Replace
```
def : PatGpr<ctlz, CV_FL1, i32>;
```

with a pattern that only matches `is_zero_poison=true` and emits
```assembly
xori (cv.fl1 rs1), 31
```

For `is_zero_poison=false` (i.e. defined behaviour on x == 0), fall through to the generic RISC-V lowering, which produces a correct sequence handling the zero case.

I have a patch with the fix and a regression test ready. Happy to send it as a PR if this analysis is confirmed.

### Environment

  - LLVM commit: 4f41bd4d04fdb2c48fdd2f7fa134bd6ee0a9a73d
  - Target: `riscv32-unknown-elf`
  - `-march=rv32i_xcvbitmanip`

VittorioBurani:
Fixes #197006.

The XCVbitmanip pattern `def : PatGpr<ctlz, CV_FL1, i32>;` is
semantically incorrect: `cv.fl1` returns a bit position while
`llvm.ctlz` returns a leading-zero count. They differ by up to 31 on
every non-zero input, miscompiling every use of `__builtin_clz`,
including compiler-rt's `__udivdi3`.

This PR:
- Replaces the pattern with one matching `llvm.ctlz` with
  `is_zero_poison=true`, lowered to `xori (cv.fl1 rs1), 31`.
- Adds explicit `setOperationAction` calls in `RISCVISelLowering.cpp`
  to route `ISD::CTLZ` to generic expansion and keep
  `ISD::CTLZ_ZERO_*` matched by the `.td` pattern.

The companion `def : PatGpr<cttz, CV_FF1, i32>;` is correct as-is
(both return 32 on zero, coincide on non-zero), and is left
untouched.

See the linked issue for the full analysis including a CV32E40P sim
trace showing miscompilation of `__udivdi3` in compiler-rt.

### DIFF
--- a/llvm/lib/Target/RISCV/RISCVISelLowering.cpp
+++ b/llvm/lib/Target/RISCV/RISCVISelLowering.cpp
@@ -454,6 +454,8 @@ RISCVTargetLowering::RISCVTargetLowering(const TargetMachine &TM,
         (Subtarget.hasStdExtZbb() || Subtarget.hasStdExtP()))
       setOperationAction({ISD::CTLZ, ISD::CTLZ_ZERO_POISON}, MVT::i32, Custom);
   } else {
+    if (Subtarget.hasVendorXCVbitmanip() && !Subtarget.is64Bit())
+      setOperationAction(ISD::CTLZ_ZERO_POISON, XLenVT, Legal);
     setOperationAction(ISD::CTLZ, XLenVT, Expand);
   }
 
--- a/llvm/lib/Target/RISCV/RISCVInstrInfoXCV.td
+++ b/llvm/lib/Target/RISCV/RISCVInstrInfoXCV.td
@@ -734,7 +734,10 @@ let Predicates = [HasVendorXCVbitmanip, IsRV32] in {
                                           (CV_LO5 cv_uimm10:$imm))>;
 
   def : PatGpr<cttz, CV_FF1, i32>;
-  def : PatGpr<ctlz, CV_FL1, i32>;
+
+  // ctlz(x) = 31 - fl1(x), for x != 0. Not intercepting the x == 0 case.
+  def : Pat<(i32 (ctlz_zero_poison GPR:$rs1)), (XORI (CV_FL1 GPR:$rs1), 31)>;
+
   def : PatGpr<int_riscv_cv_bitmanip_clb, CV_CLB>;
   def : PatGpr<ctpop, CV_CNT, i32>;
 
--- a/llvm/lib/Target/RISCV/RISCVSubtarget.h
+++ b/llvm/lib/Target/RISCV/RISCVSubtarget.h
@@ -191,10 +191,7 @@ public:
     return HasStdExtZfhmin || HasStdExtZfbfmin;
   }
 
-  bool hasCLZLike() const {
-    return HasStdExtZbb || HasVendorXTHeadBb ||
-           (HasVendorXCVbitmanip && !IsRV64);
-  }
+  bool hasCLZLike() const { return HasStdExtZbb || HasVendorXTHeadBb; }
   bool hasCTZLike() const {
     return HasStdExtZbb || (HasVendorXCVbitmanip && !IsRV64);
   }
--- a/llvm/test/CodeGen/RISCV/xcvbitmanip.ll
+++ b/llvm/test/CodeGen/RISCV/xcvbitmanip.ll
@@ -116,7 +116,48 @@ define i32 @test.cv.fl1(i32 %a) {
 ; CHECK-LABEL: test.cv.fl1:
 ; CHECK:       # %bb.0:
 ; CHECK-NEXT:    cv.fl1 a0, a0
+; CHECK-NEXT:    xori a0, a0, 31
 ; CHECK-NEXT:    ret
+  %1 = call i32 @llvm.ctlz.i32(i32 %a, i1 1)
+  ret i32 %1
+}
+
+; Verifies that ctlz with defined-on-zero behavior is lowered
+; correctly: a zero-check guard branches around the cv.fl1 + xori 31
+; sequence and returns 32 directly when the input is zero. This
+; relies on ISD::CTLZ being Expand and ISD::CTLZ_ZERO_POISON being
+; Legal for XCVbitmanip; the LegalizeDAG framework synthesises the
+; branch-guarded form automatically.
+define i32 @test.ctlz.zero.defined(i32 %a) {
+; CHECK-O0-LABEL: test.ctlz.zero.defined:
+; CHECK-O0:       # %bb.0:
+; CHECK-O0-NEXT:    addi sp, sp, -16
+; CHECK-O0-NEXT:    .cfi_def_cfa_offset 16
+; CHECK-O0-NEXT:    cv.fl1 a1, a0
+; CHECK-O0-NEXT:    xori a1, a1, 31
+; CHECK-O0-NEXT:    li a2, 32
+; CHECK-O0-NEXT:    sw a2, 8(sp) # 4-byte Folded Spill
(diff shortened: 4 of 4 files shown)


llvm/test/CodeGen/RISCV/xcvbitmanip-clz.ll
```diff
@@ -0,0 +1,59 @@
+; NOTE: Do not use update_llc_test_checks.py on this file.
```
lenary: Why not? It's much easier to actually use update_llc_test_checks.py than to have a custom test file. If you need notes about a specific function in the file, leave them above the function declaration, and they will be preserved.

lenary:
I think you should probably update the definition of `RISCVSubtarget::hasCLZLike()` as well - given this influences some other places used in codegen.

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
```diff
@@ -399,6 +399,8 @@ RISCVTargetLowering::RISCVTargetLowering(const TargetMachine &TM,
     setOperationAction({ISD::ROTL, ISD::ROTR}, XLenVT, Custom);
   } else if (Subtarget.hasVendorXCVbitmanip() && !Subtarget.is64Bit()) {
     setOperationAction(ISD::ROTL, XLenVT, Expand);
+    setOperationAction(ISD::CTLZ_ZERO_POISON, XLenVT, Legal);
```
topperc: Please keep these lines with other CTLZ/CTTZ code. This section is sorted by operation, not by features.

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
```diff
@@ -399,6 +399,8 @@ RISCVTargetLowering::RISCVTargetLowering(const TargetMachine &TM,
...
+    setOperationAction(ISD::CTLZ, XLenVT, Expand);
```
lenary: If you update `hasCLZLike` to return false for xcvbitmanip, then you only need the ctlz_zero_poison, line, which would be better placed near the block around line 445 which deals with CTLZ and related operations.

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
topperc: We definitely need to update `hasCLZLike` in RISCVSubtarget.h to remove XCVBitmanip

llvm/test/CodeGen/RISCV/xcvbitmanip-clz.ll
topperc: I'm not sure why we need a separate test file. Just put the tests in xcvbitmanip.ll

VittorioBurani: It was for the CHECK-NOT in the new file. I'll drop the new one and merge the files.

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
VittorioBurani: i'll fix it

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
VittorioBurani: i'll fix it

VittorioBurani:
Done @topperc @lenary. Moved `CTLZ_ZERO_POISON Legal` to the block at line ~445 (the one for `!hasCLZLike()`). Removed `HasVendorXCVbitmanip` from `RISCVSubtarget::hasCLZLike()`. Consolidated tests into `xcvbitmanip.ll`, regenerated with update_llc_test_checks.py. Now LegalizeDAG synthesises a branch-guarded form for `ISD::CTLZ`.

topperc (approved):
LGTM

lenary (approved):
LGTM

wangpc-pp (commented):
The test is failing.

lenary:
Sorry, I didn't see this had been re-updated. I've just approved the CI checks to run, so we can see if all the tests pass.

lenary:
@VittorioBurani please can you set your github email to public (https://llvm.org/docs/DeveloperPolicy.html#email-addresses) ? Then I can land this.

VittorioBurani:
@lenary done ✅

PR #197289, opened 2026-05-12
[SandboxIR][Tracker] Implement accept(/*AcceptAll*/) and revert(/*RevertAll*/)

vporpo:
In the context of nested checkpoints the tracker's API was somewhat inconsistent. Tracker::revert() would revert to the last checkpoint but accept() would accept all changes.

This patch fixes this, and introduces `accept(bool AcceptAll)` and `revert(bool RevertAll)`.


gandhi56:
Hey @vporpo, could you explain what a snapshot is exactly? My guess is it represents a state of a region but I am not really sure how an unsigned int represents a state.

gandhi56 (approved):

vporpo:
Discussed this offline, but for reference the snapshot is an index into the Tracker::Changes vector which is the vector that holds the tracked objects. The index + the changes vector is all the information we need to point to a specific IR state.

PR #187110, opened 2026-03-17
[z/OS] Add z/OS archive reading support

uyoyo0:
Add support for reading `z/OS` archives, which use EBCDIC-encoded header fields and an EBCDIC magic string. The `z/OS` archive format shares the same structural layout as traditional Unix archives but all text fields (member names, timestamps, permissions, and symbol names) are in EBCDIC. This patch adds:

 - `K_ZOS` archive kind 
 - `ZOSArchiveMemberHeader`: converts EBCDIC header fields to ASCII on read 
 - `ZOSArchive`: parses the __.SYMDEF symbol table, converting EBCDIC symbol names to ASCII 
 - Updates to symbol table traversal for `K_ZOS`, which uses big-endian 4-byte offsets paired with 4-byte attribute words per symbol 
 
 This is part 2 of a patch series adding `z/OS` archive support to LLVM.
 
 Part 1: #186854

PR #197290, opened 2026-05-12
Fix z/OS archive test failure on macOS

uyoyo0:
Fixes failures introduced by #187110.
- https://lab.llvm.org/buildbot/#/builders/190/builds/42544
- https://lab.llvm.org/buildbot/#/builders/23/builds/19989
- https://logs.chromium.org/logs/fuchsia/buildbucket/cr-buildbucket/8682002400978909393/+/u/clang/test/stdout

The original test hit a "file to small" error on macOS before hitting the expected "truncated or malformed archive" error. This patch updates the test to generate a valid archive then truncates it to 28 bytes so that its large enough to pass the initial size check but too small for the 60-byte member header, so it correctly hits the expected failure on all platforms.


ilovepi (approved):

smeenai:
I merged to fix buildbots

PR #197292, opened 2026-05-12
[AMDGPU][SCEV] Tighten max backedge-taken count for shift recurrence loops.

carlobertolli:
Application code on AMDGPUs often uses the following pattern:

for (int i = blockDim.x / 2; i >= 1; i >>= 1)
  if (threadIdx.x < i) {
    <STMT based on threadIdx.x>
  }
  __syncthreads();
}

but this is currently not unrolled by LLVM for two reasons:
1. We overestimate the loop count with the number of bits in the iteration variable (32).
2. Runtime unrolling is not applicable because of convergent instructions.

For the specific pattern above, we can compute a tighter bound by using range information, which for blockDim.x is (0,1025) and already emitted by clang, resulting in 10 as upper bound as opposed to 32.
The next step in unrolling the loop is to adjust the threshold used to control full loop unrolling with convergent instructions.

Assisted-by: Cursor

### DIFF
--- a/llvm/lib/Analysis/ScalarEvolution.cpp
+++ b/llvm/lib/Analysis/ScalarEvolution.cpp
@@ -9699,10 +9699,11 @@ ScalarEvolution::ExitLimit ScalarEvolution::computeShiftCompareExitLimit(
     return getCouldNotCompute();
 
   // Return true if V is of the form "LHS `shift_op` <positive constant>".
-  // Return LHS in OutLHS and shift_opt in OutOpCode.
-  auto MatchPositiveShift =
-      [](Value *V, Value *&OutLHS, Instruction::BinaryOps &OutOpCode) {
-
+  // Return LHS in OutLHS, shift_op in OutOpCode, and the shift amount in
+  // OutShiftAmt.
+  auto MatchPositiveShift = [](Value *V, Value *&OutLHS,
+                               Instruction::BinaryOps &OutOpCode,
+                               unsigned &OutShiftAmt) {
     using namespace PatternMatch;
 
     ConstantInt *ShiftAmt;
@@ -9715,7 +9716,11 @@ ScalarEvolution::ExitLimit ScalarEvolution::computeShiftCompareExitLimit(
     else
       return false;
 
-    return ShiftAmt->getValue().isStrictlyPositive();
+    uint64_t Amt = ShiftAmt->getValue().getLimitedValue();
+    if (Amt == 0 || Amt >= OutLHS->getType()->getScalarSizeInBits())
+      return false;
+    OutShiftAmt = Amt;
+    return true;
   };
 
   // Recognize a "shift recurrence" either of the form %iv or of %iv.shifted in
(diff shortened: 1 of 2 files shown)


llvm/lib/Analysis/ScalarEvolution.cpp
```diff
@@ -9687,6 +9688,7 @@ ScalarEvolution::ExitLimit ScalarEvolution::computeShiftCompareExitLimit(
     else
       return false;
 
+    OutShiftAmt = ShiftAmt->getZExtValue();
```
efriedma-quic: `ShiftAmt->getZExtValue()` will assert if the shift amount doesn't fit into a uint64_t.

Probably we should check that the shift amount is less than the bit width.

carlobertolli: done, thanks for reporting this.

llvm/lib/Analysis/ScalarEvolution.cpp
```diff
@@ -9687,7 +9688,9 @@ ScalarEvolution::ExitLimit ScalarEvolution::computeShiftCompareExitLimit(
     else
       return false;
 
-    return ShiftAmt->getValue().isStrictlyPositive();
+    OutShiftAmt = ShiftAmt->getValue().getLimitedValue();
+    return OutShiftAmt > 0 &&
+           OutShiftAmt < OutLHS->getType()->getScalarSizeInBits();
```
adelejjeh: The semantics of the return value is now different, we will now return false for shiftamnt > bitdwidth. While it is sound, this seems orthogonal to the intended optimization and changes existing behavior. If the change is intentional, maybe we should incorporate a test that executes this path?

llvm/lib/Analysis/ScalarEvolution.cpp
```diff
@@ -9687,7 +9688,9 @@ ScalarEvolution::ExitLimit ScalarEvolution::computeShiftCompareExitLimit(
     else
       return false;
 
-    return ShiftAmt->getValue().isStrictlyPositive();
+    OutShiftAmt = ShiftAmt->getValue().getLimitedValue();
```
adelejjeh: This is doing a truncation from uint64 to unsigned. Do we need OutshiftAmt to be unsigned?

adelejjeh (commented):
Left some comments. Also, the PR description needs reformating I think, your bullet 2. keeps extending with the description of what you are trying to do in your approach. The description of the change can maybe be more elaborate as well.

llvm/lib/Analysis/ScalarEvolution.cpp
carlobertolli: This was added when responding to an earlier review. It does make sense as per LangRef lshr:

> If op2 is (statically or dynamically) equal to or larger than the number of bits in op1, this instruction returns a poison value.

I added a test.

llvm/lib/Analysis/ScalarEvolution.cpp
carlobertolli: fixed with a local variable + check.

carlobertolli:
ping

adelejjeh (approved):
Approve with Nit: the change fixes the bug that can arrise if Amt gets truncated from uint64_t to unsigned, however it may be good to have a test that shows a large shift ammount that gets truncated doesn't cause problems.

nikic (approved):
LGTM

carlobertolli:
patch is screwed up by local merge. converting to draft to fix it.

carlobertolli:
fixed PR, back to proper PR, ready to merge

carlobertolli:
added a test for a shift value that is truncated from uint64_t to unsigned and triggers the check.

PR #197293, opened 2026-05-12
[ELF] Fix location counter used for output section address with MEMOR…

### LINKED ISSUE #112919: [lld] location counter alignment not affecting . when assigning an output section address and a MEMORY region.
This example is derived from an internal linker script used with ld.bfd that brought up an unexpected warning with LLD. There are ways to work around it. Raising to log the discrepancy.

When assigning an OutputSection an address an explicit address and a MEMORY region, my reading of the ld.bfd manual is that the explicit address takes precedence over the location in the MEMORY region (https://sourceware.org/binutils/docs/ld/Output-Section-Address.html)

It looks like when assigning the OutputSection the value of the location counter as the address, lld uses the location counter from the MEMORY region and not the global location counter. For example:
```
        .section s01, "ax", %progbits
        .byte 1

        .section s02, "ax", %progbits
        .balign 4
        .word 1
```
With linker script:
```
MEMORY
{
  FLASH (rx) : ORIGIN = 0x1000, LENGTH = 0x100
  RAM (rw) : ORIGIN = 0x2000, LENGTH = 0x1000
}

SECTIONS
{
 /* s01 is 1 byte */
  s01 : { *(s01) } > FLASH
 /* . == 1 */
  . = ALIGN(4);
 /* . == 4 expectation that s02 gets 4-byte aligned address */
  s02 . : { *(s02) } > FLASH
}
```
However we get 
```
ld.lld: warning: address (0x1001) of section s02 is not a multiple of alignment (4)
```
With Map file:
```
             VMA              LMA     Size Align Out     In      Symbol
            1000             1000        1     1 s01
            1000             1000        1     1         align.o:(s01)
            1001             1001        3     1 . = ALIGN(4)
            1001             1001        5     4 s02
            1004             1004        2     4         align.o:(s02)
```
Whereas with ld.bfd we get
```
s01             0x0000000000001000        0x1
 *(s01)
 s01            0x0000000000001000        0x1 align.o
                0x0000000000001004                . = ALIGN (0x4)

s02             0x0000000000001004        0x2
 *(s02)
 s02            0x0000000000001004        0x2 align.o
```
If I move the ALIGN to inside the s01 OutputSection `{ *(s01); . = ALIGN(4); }`, we get similar results
```
            1000             1000        4     1 s01
            1000             1000        1     1         align.o:(s01)
            1001             1001        3     1         . = ALIGN(4)
            1004             1004        2     4 s02
            1004             1004        2     4         align.o:(s02)
```

My suspicion is that when using `.` in the context of Memory regions is using the location counter from the MEMORY region and not the global location counter.

deepakshirkem:
When a section uses '.' as its address expression inside a MEMORY region (e.g. 's02 . : { ... } > FLASH'), lld was incorrectly evaluating '.' against the memory region's current position instead of the global location counter. Save the global dot before the memRegion override and restore it when an explicit addrExpr is present.

Fixes #112919

@maskray @smithp35 could you please review this patch?

### DIFF
--- a/lld/ELF/LinkerScript.cpp
+++ b/lld/ELF/LinkerScript.cpp
@@ -1193,7 +1193,9 @@ bool LinkerScript::assignOffsets(OutputSection *sec) {
     else
       dot = state->tbssAddr;
   } else {
-    if (state->memRegion)
+    // If there is an explicit address expression this takes precedence over
+    // the memory region address.
+    if (state->memRegion && !(hasSectionsCommand && sec->addrExpr))
       dot = state->memRegion->curPos;
     if (sec->addrExpr)
       setDot(sec->addrExpr, sec->location, false);
--- a/lld/test/ELF/linkerscript/memory-loc-counter-dot-addr.s
+++ b/lld/test/ELF/linkerscript/memory-loc-counter-dot-addr.s
@@ -1,0 +1,41 @@
+# REQUIRES: x86
+# RUN: rm -rf %t && split-file %s %t && cd %t
+# RUN: llvm-mc -filetype=obj -triple=x86_64 asm.s -o t.o
+# RUN: ld.lld -T lds t.o -o t.out
+# RUN: llvm-readelf -S t.out | FileCheck %s
+
+## Test that when a section uses '.' as its address expression inside a MEMORY
+## region, the global location counter is used, not the memory region position.
+## GNU ld manual: explicit address takes precedence over memory region position.
+## s01 placed at FLASH origin 0x1000
+# CHECK: s01 PROGBITS 0000000000001000
+## s02 uses '.' as explicit address — global dot after ALIGN(4) = 0x1004
+# CHECK: s02 PROGBITS 0000000000001004
+## s03 has no explicit address — follows memRegion->curPos after s02
+## second ALIGN(4) does not affect s03 as it has no explicit address
+# CHECK: s03 PROGBITS 0000000000001006
+
+#--- asm.s
+.section s01, "a", @progbits
+.byte 1
+
+.section s02, "a", @progbits
+.short 1
+
+.section s03, "a", @progbits
+.short 1
+
+#--- lds
+MEMORY
+{
+  FLASH (rx) : ORIGIN = 0x1000, LENGTH = 0x100
+}
+
+SECTIONS
+{
+  s01 : { *(s01) } > FLASH
+  . = ALIGN(4);
+  s02 . : { *(s02) } > FLASH
+  . = ALIGN(4);
+  s03 : { *(s03) } > FLASH
+}
--- a/lld/docs/ReleaseNotes.rst
+++ b/lld/docs/ReleaseNotes.rst
@@ -48,6 +48,12 @@ Breaking changes
   feature saw no adoption beyond a Chromium experiment that has since been
   retired.
 
+* An OutputSection that has an address expression, and is also assigned
+  to a MEMORY region, will now use the address expression in preference
+  to the next available location in the MEMORY region. This brings LLD
+  in line with GNU ld, but is a change in behavior from previous LLD
+  releases.
+  
 COFF Improvements
 -----------------
 
--- a/lld/test/ELF/linkerscript/end-overflow-check.test
+++ b/lld/test/ELF/linkerscript/end-overflow-check.test
@@ -35,7 +35,7 @@ MEMORY
 
 SECTIONS
 {
-    _abss ALIGN(REGION1__PRE_ALIGNMENT) :
+    _abss : ALIGN(REGION1__PRE_ALIGNMENT)
     {
         REGION1__BEGIN = .; REGION1__ALIGNED_BEGIN = .; REGION1_ALIGNED_BEGIN = .;
         *(.a.bss)
@@ -58,7 +58,7 @@ MEMORY
 
 SECTIONS
 {
-    _abss ALIGN(REGION1__PRE_ALIGNMENT) :
+    _abss : ALIGN(REGION1__PRE_ALIGNMENT) 
     {
         REGION1__BEGIN = .; REGION1__ALIGNED_BEGIN = .; REGION1_ALIGNED_BEGIN = .;
         *(.a.bss)


smithp35:
I'll take a look. I noticed this myself in https://github.com/llvm/llvm-project/issues/112919 hopefully we can add fixes #112919 to the description to close that.

I'd like to check over how GNU ld handles this case to check we've got the right model.

lld/test/ELF/linkerscript/memory-loc-counter-dot-addr.test
```diff
@@ -0,0 +1,23 @@
+# REQUIRES: aarch64
```
smithp35: LLD convention is for all non-architecture specific tests to use x86

lld/test/ELF/linkerscript/memory-loc-counter-dot-addr.test
```diff
@@ -0,0 +1,23 @@
...
+# RUN: echo '.section s01,"a"; .byte 1; .section s02,"a"; .balign 4; .short 1' \
```
smithp35: New tests now use split-file to combine the assembler and linker script into one file. For an example of how to use split-file see `edata-etext.s`

lld/test/ELF/linkerscript/memory-loc-counter-dot-addr.test
```diff
@@ -0,0 +1,23 @@
...
+{
+  FLASH (rx) : ORIGIN = 0x1000, LENGTH = 0x100
+}
+
+SECTIONS
+{
+  s01 : { *(s01) } > FLASH
+  . = ALIGN(4);
+  s02 . : { *(s02) } > FLASH
+}
```
smithp35: Would you be able to expand the test to include a s03 that does not use the location counter. For example:
```
MEMORY
{
  FLASH (rx) : ORIGIN = 0x1000, LENGTH = 0x100
}

SECTIONS
{
  s01 : { *(s01) } > FLASH
  . = ALIGN(4);
  s02 . : { *(s02) } > FLASH
  . = ALIGN(4);
  s03 : { *(s03) } > FLASH
}
```
We want to show that the `.=ALIGN(4)` has no effect on `s03`. which should follow on from `s02`.

lld/ELF/LinkerScript.cpp
```diff
@@ -1197,10 +1197,13 @@ bool LinkerScript::assignOffsets(OutputSection *sec) {
     else
       dot = state->tbssAddr;
   } else {
+    const uint64_t savedDot = dot;
```
smithp35: Instead of using a variable (that shadows an existing savedDot variable), can we do something like:
```
// If no address specified and section is assigned
// to a memory region use next available memory
// address.
if (!sec->addrExpr && state->memRegion)
  dot = state->memRegion->curPos;
if (sec->addrExpr)
  setDot(sec->addrExpr, sec->location, false);
```
That matches the docs more closely https://sourceware.org/binutils/docs/ld/MEMORY.html

smithp35 (commented):
I've checked GNU ld source, and testing as well as the GNU ld docs https://sourceware.org/binutils/docs/ld/MEMORY.html which say:

> If no address was specified for the output section, the linker will set the address to the next available address within the memory region.
 
I've left a few comments on the test and implementation.

lld/test/ELF/linkerscript/memory-loc-counter-dot-addr.test
deepakshirkem: Yes, Will implement that

deepakshirkem:
Hi @smithp35, Thank you for your review. WIll try to resolve those comments one by one.

MaskRay:
> Thank you for your review. WIll try to resolve those comments one by one.

You don't have to send such a comment. You can just resolve comments, update patch, then click "re-request review".

lld/test/ELF/linkerscript/memory-loc-counter-dot-addr.s
```diff
@@ -0,0 +1,44 @@
+# REQUIRES: x86
+# RUN: split-file %s %t
```
smithp35: There's a pattern you can use to simplify the test.
```
# RUN: rm -rf %t && split-file %s %t && cd %t
```
Now you don't need to prefix all the file commands with %t

lld/test/ELF/linkerscript/memory-loc-counter-dot-addr.s
```diff
@@ -0,0 +1,44 @@
...
+
+## s01 placed at FLASH origin 0x1000
+# CHECK: s01 PROGBITS 0000000000001000
+## s02 uses '.' as explicit address — global dot after ALIGN(4) = 0x1004
+# CHECK: s02 PROGBITS 0000000000001004
+## s03 has no explicit address — follows memRegion->curPos after s02 = 0x1006
+## second ALIGN(4) does not affect s03 as it has no explicit address
+# CHECK: s03 PROGBITS 0000000000001006
+
+#--- asm
```
smithp35: Could this be `asm.s` as it makes it easier to tell that the input file is assembler in the command-line?

lld/test/ELF/linkerscript/memory-loc-counter-dot-addr.s
```diff
@@ -0,0 +1,44 @@
...
+# RUN: split-file %s %t
+# RUN: llvm-mc -filetype=obj -triple=x86_64 %t/asm -o %t.o
+# RUN: ld.lld -T %t/lds %t.o -o %t.out
+# RUN: llvm-readelf -S %t.out | FileCheck %s
+
+## Test that when a section uses '.' as its address expression inside a MEMORY
+## region, the global location counter is used, not the memory region position.
+## GNU ld manual: explicit address takes precedence over memory region position.
+## https://sourceware.org/binutils/docs/ld/Output-Section-Address.html
+## See https://github.com/llvm/llvm-project/issues/112919
```
smithp35: I wouldn't bother mentioning that here, it can be derived from the git log/PR if anyone is particularly curious.

smithp35 (commented):
Thanks for the updates. I've got a couple of small comments on the test, but other than that it is looking good.

Could you add `fixes #112919` to the commit message so it will close that issue when merged?

smithp35:
There's a couple of test failures (I authorized the CI) that look related:
```
lld :: ELF/linkerscript/end-overflow-check.test
lld :: ELF/linkerscript/memory-no-sections.test
```
The end-overflow-check looks like this is this change behaving as expected. I think the test may need updating to move the alignment after the : https://github.com/llvm/llvm-project/blob/main/lld/test/ELF/linkerscript/end-overflow-check.test 
`_abss ALIGN(REGION1__PRE_ALIGNMENT) :` -> `_abss : ALIGN(REGION1__PRE_ALIGNMENT)` as that shouldn't override the base address of the output section.

Not sure about the memory-no-sections.text . I expect that test is assigning output sections to default Memory regions but there shouldn't be any address expressions. That may take some debugging.

The first test failure implies that we should release note the change in https://github.com/llvm/llvm-project/blob/main/lld/docs/ReleaseNotes.rst#breaking-changes 

For example:
```
* An OutputSection that has an address expression, that is also assigned to a MEMORY region, will use the address expression in preference to the next available location in the MEMORY region. This brings LLD in line with GNU ld, but is a change in behavior from previous LLD releases.
```

deepakshirkem:
Hi @smithp35, Updated the fix, and it now passes both previously failing tests. Please take another look at the changes when you get a chance.

smithp35:
> Hi @smithp35, Updated the fix, and it now passes both previously failing tests. Please take another look at the changes when you get a chance.

Thanks for the update. I'll take a look this week. It would help me out if you can give an explanation for why those two tests failed, and why the new fixes work.

MaskRay:
There are multiple problems when you force-push a branch after a rebase, so as a guideline we avoid force pushes and append new commits instead. https://maskray.me/blog/2026-01-22-maintaining-shadow-branches-for-github-prs

You can use tools  like spr and pr-shadow to avoid force pushes.

deepakshirkem:
Hi @smithp35,  those two test failures were caused by our initial fix not correctly handling the case where internally sets `fixSectionAlignments()` while the user also provides a dot address. As you suggested, I used gdb to dubug the  `memory-no-sections.test` failure. The breakpoint at LinkerScript.cpp:1200 showed:

```
  sec->name         = ".text"
  dot               = 344  (ELF header size)
  memRegion->curPos = 4096 (0x1000 — correct flash origin)
  hasSectionsCommand = false
  sec->addrExpr     = Writer::fixSectionAlignments()::lambda
```
This confirmed that our initial fix was skipping  `dot = memRegion->curPos` whenever `addrExpr` existed even when it  was set internally by `fixSectionAlignments()`. As a result, the synthetic `addrExpr` was evaluating with the wrong dot value (344)  instead of the correct memory region position (0x1000).

To fix this, I added the condition `hasSectionsCommand && savedDot > dot` which ensures we only restore the global dot when: - `hasSectionsCommand` is true

smithp35:
> Hi @smithp35, those two test failures were caused by our initial fix not correctly handling the case where internally sets `fixSectionAlignments()` while the user also provides a dot address. As you suggested, I used gdb to dubug the `memory-no-sections.test` failure. The breakpoint at LinkerScript.cpp:1200 showed:
> 
> ```
>   sec->name         = ".text"
>   dot               = 344  (ELF header size)
>   memRegion->curPos = 4096 (0x1000 — correct flash origin)
>   hasSectionsCommand = false
>   sec->addrExpr     = Writer::fixSectionAlignments()::lambda
> ```
> 
> This confirmed that our initial fix was skipping `dot = memRegion->curPos` whenever `addrExpr` existed even when it was set internally by `fixSectionAlignments()`. As a result, the synthetic `addrExpr` was evaluating with the wrong dot value (344) instead of the correct memory region position (0x1000).
> 
> To fix this, I added the condition `hasSectionsCommand && savedDot > dot` which ensures we only restore the global dot when: - `hasSectionsCommand` is true

Thanks for the explanation. I need a bit more time to think about this. It seems like the fix is in pseudo code:
* Move the dot to the location counter in the memory region curpos.
* If the global dot happens to be greater than the memory region curpos, then reset the dot to global dot.

While that will work in practice for `. = ALIGN(x)` expressions in between assignments to the same memory region, I'm not sure yet whether that's going to work in all possible cases.

smithp35:
My apologies I ran out of time this week. I'll be out for the next 4 days, back on Tuesday 26th. Will pick this back up then.

lld/ELF/LinkerScript.cpp
```diff
@@ -1199,9 +1199,11 @@ bool LinkerScript::assignOffsets(OutputSection *sec) {
   } else {
     if (state->memRegion)
       dot = state->memRegion->curPos;
-    if (sec->addrExpr)
+    if (sec->addrExpr) {
+      if (hasSectionsCommand && savedDot > dot)
```
smithp35: I don't think the `savedDot > dot` is correct. There's nothing in the linker script documentation that prevents the address expression from having a smaller value.

For example:
```
MEMORY
{
  FLASH (rx) : ORIGIN = 0x2000, LENGTH = 0x100
}

SECTIONS {

  s01 : { *(s01) } > FLASH1
  . = 0x1f00;
  s02 . + 0x108 : { *(s02) } > FLASH1
}
```
With inputs
```
.section s01, "a", @progbits
.byte 1

.section s02, "a", @progbits
.short 1
```

My savedDot is lower than the memory region location counter but after we've added the 0x108 then we get a value within the bounds of the memory region.

With this change lld ignores my move of the location counter, whereas GNU ld does not.

I think you added the `>` to make the `end-overflow-check.test` pass.

I think the way that test is written is flawed.

```
MEMORY
{
    SRAM0 (rw) : ORIGIN = 0x20000400, LENGTH = 4K
}

SECTIONS
{
    _abss ALIGN(REGION1__PRE_ALIGNMENT) :
    {
        REGION1__BEGIN = .; REGION1__ALIGNED_BEGIN = .; REGION1_ALIGNED_BEGIN = .;
        *(.a.bss)
        REGION1__END = .; . = ALIGN(REGION1__POST_ALIGNMENT); REGION1_ALIGNED_END = .;
    } > SRAM0
}
```

The address of _abss should be `.` aligned to `REGION1__PRE_ALIGNMENT`.

The orphan sections like .text are default assigned to the `SRAM0` memory region. In GNU ld the .text happens to be placed prior to _abss, which sets the location counter after .text to `0x20000401` which is then aligned up to `0x20000800`. In lld the orphan .text section is placed after non-orphans so the value of the location counter is 0x0 on entering `_abss`.

If I remove the `savedDot > dot` but write the test so the location counter is 0x20000400 then the test will pass. I can do this in a number of ways:
* Put the ALIGN to the right hand side of the colon. _abss : ALIGN(REGION1__PRE_ALIGNMENT)
* Make the location counter 0x20000400 prior to _abss. We could just do `. = 0x20000400` or we place another 0 sized .bss section in SRAM0. Such as `.anchor.bss : { *(.anchor.bss) } > SRAM0`

smithp35 (commented):
I've had a chance to check over GNU ld. Apologies for the delay; I couldn't find any logical reason for `savedDot > Dot` based on the GNU ld documentation, but at the time I didn't have a counter-example for when this could go wrong. I think I've found some now.

smithp35 (commented):
Thanks for the updates. I've got one refactoring suggestion, but other than that this is looking good.

deepakshirkem:
> Thanks for the updates. I've got one refactoring suggestion, but other than that this is looking good.

Please share your refactoring suggestion when you get a chance? I will try. Thank You ::))

lld/ELF/LinkerScript.cpp
```diff
@@ -1195,9 +1195,11 @@ bool LinkerScript::assignOffsets(OutputSection *sec) {
   } else {
     if (state->memRegion)
       dot = state->memRegion->curPos;
-    if (sec->addrExpr)
+    if (sec->addrExpr) {
```
smithp35: Thanks for updating. 

I think the following is logically equivalent, yet doesn't involve savedDot. I've added a comment to explain.
```
    // If there is an explicit address expression this takes precedence over
    // the memory region address.
    if (state->memRegion &&
        !(hasSectionsCommand && sec->addrExpr))
      dot = state->memRegion->curPos;
    if (sec->addrExpr)
      setDot(sec->addrExpr, sec->location, false);

```

smithp35 (commented):
My apologies, I had a comment with the suggestion, but it looks like it is still pending, I hate the GitHub UI! 

Hopefully should be there now.

lld/ELF/LinkerScript.cpp
deepakshirkem: Thank you ::)). This makes sense. I tested with our existing tests and they all pass with this patch.

smithp35 (approved):
Thanks for the update LGTM. As a new contributor I expect you won't have merge permission. Assuming @MaskRay has no objections or further comments I'll merge this on Friday (to give time for other comments).

smithp35:
> Thanks for the update LGTM. As a new contributor I expect you won't have merge permission. Assuming @MaskRay has no objections or further comments I'll merge this on Friday (to give time for other comments).

Now Friday: merging.

lld/ELF/LinkerScript.cpp
```diff
@@ -1193,7 +1193,9 @@ bool LinkerScript::assignOffsets(OutputSection *sec) {
     else
       dot = state->tbssAddr;
   } else {
-    if (state->memRegion)
+    // If there is an explicit address expression this takes precedence over
+    // the memory region address.
+    if (state->memRegion && !(hasSectionsCommand && sec->addrExpr))
```
MaskRay, line 1198: FTR, Writer::fixSectionAlignments() sets `addrExpr` when hasSectionsCommand is false.

 Without the guard, in a MEMORY-without-SECTIONS script, the synthetic page-alignment addrExpr would suppress the dot = memRegion->curPos placement and break

MaskRay:
Note: emails are not supposed to be hidden https://discourse.llvm.org/t/github-hidden-email-and-automated-checker-for-pull-requests/90629

PR #197294, opened 2026-05-12
[BOLT][rewrite] warn about functions without CFG before binary analyses.

devnexen:


bolt/lib/Rewrite/RewriteInstance.cpp
```diff
@@ -3888,8 +3888,24 @@ void RewriteInstance::runBinaryAnalyses() {
   NamedRegionTimer T("runBinaryAnalyses", "run binary analysis passes",
                      TimerGroupName, TimerGroupDesc, opts::TimeRewrite);
   BinaryFunctionPassManager Manager(*BC);
-  // FIXME: add a pass that warns about which functions do not have CFG,
-  // and therefore, analysis is most likely to be less accurate.
+
+  // Warn about functions without a reconstructed CFG: binary analyses degrade
+  // in accuracy on those and may produce false negatives.
```
atrosinenko: Probably better not to mention "false negatives" specifically. For example, LLVM may have to emit multi-instruction checker sequences under some conditions for PAuth code to be safe. These sequences are of the form

```
        ; x0 was authenticated, but the CPU may not implement FEAT_FPAC
        ; <check if x0 is a valid address>
        ; <branch to .Lon_success if the check passes>
        brk     0xc472 ; or a few other error codes
.Lon_success:
        ; Now x0 is known to be authenticated *successfully*.
```

Presently, PAuth gadget scanner only detects such sequences if CFG is available ([example](https://github.com/llvm/llvm-project/blob/ec60d62118bf17328b4d63c67eb981f9c8347d56/bolt/test/binary-analysis/AArch64/gs-pauth-signing-oracles.s#L53)), meaning it would pessimistically ignore this safety measure if CFG is missing. Thus, false positives are possible, too.

bolt/lib/Rewrite/RewriteInstance.cpp
```diff
@@ -3888,8 +3888,24 @@ void RewriteInstance::runBinaryAnalyses() {
...
+  uint64_t NoCFGCount = 0;
```
atrosinenko: [nit] Possibly, plain `unsigned` would be a bit less surprising :)

bolt/lib/Rewrite/RewriteInstance.cpp
```diff
@@ -3888,8 +3888,24 @@ void RewriteInstance::runBinaryAnalyses() {
...
+  for (const auto &BFI : BC->getBinaryFunctions()) {
+    const BinaryFunction &BF = BFI.second;
+    if (BF.isIgnored() || BF.hasCFG())
```
atrosinenko: Hmm, I wonder why `PAuthGadgetScanner.cpp` does not check `isIgnored()` as well - was it by design or should we skip `ignored` functions there too? Maybe @kbeyls remembers something on this decision.

bolt/lib/Rewrite/RewriteInstance.cpp
```diff
@@ -3888,8 +3888,24 @@ void RewriteInstance::runBinaryAnalyses() {
...
+      continue;
+    ++NoCFGCount;
+    if (opts::Verbosity >= 1)
+      BC->errs() << "BOLT-WARNING: no CFG for " << BF
+                 << "; binary analyses may be imprecise\n";
+  }
+  if (NoCFGCount)
+    BC->errs() << "BOLT-WARNING: " << NoCFGCount
+               << " function(s) lack CFG; binary-analysis results may be"
+                  " incomplete. Re-run with -v=1 to list them.\n";
```
atrosinenko: [nit] `s/list them/list these functions`?

atrosinenko (commented):
@devnexen Thank you for the patch, I left a few nit-picky comments, but the code change looks good. Could you please add some test cases?

bolt/lib/Rewrite/RewriteInstance.cpp
```diff
@@ -3888,8 +3888,24 @@ void RewriteInstance::runBinaryAnalyses() {
   NamedRegionTimer T("runBinaryAnalyses", "run binary analysis passes",
                      TimerGroupName, TimerGroupDesc, opts::TimeRewrite);
   BinaryFunctionPassManager Manager(*BC);
-  // FIXME: add a pass that warns about which functions do not have CFG,
-  // and therefore, analysis is most likely to be less accurate.
+
+  // Warn about functions for which BOLT could not reconstruct the CFG: binary
+  // analyses are les
```
atrosinenko: Half-written comment.

bolt/lib/Rewrite/RewriteInstance.cpp
atrosinenko: Thinking a bit more, I'm not really sure if we should skip `ignored` functions for the analysis purpose. I would rather only check for `if (BF.hasCFG())` so far and postpone handling of ignored functions in the analyzer until we actually need to.

As far as I can see, an example of `ignored` functions are those marked as `pseudo` functions ([link](https://github.com/llvm/llvm-project/blob/f63d8d67a3ff11d06607f2f0cfedfd88d877e9de/bolt/lib/Rewrite/RewriteInstance.cpp#L3496)). An [example](https://github.com/llvm/llvm-project/blob/b160d1d6e14c74a18b4a3043a210872cd27b6842/bolt/include/bolt/Core/BinaryFunction.h#L1322) of a pseudo function is PLT stub. I guess it can even make sense to analyze PLT stubs sometimes...

bolt/test/binary-analysis/AArch64/cfg-warning.s
```diff
@@ -0,0 +1,29 @@
...
+## reconstruct the CFG, since analysis results are less reliable for them.
+
+// RUN: %clang %cflags %s %p/../../Inputs/asm_main.c -o %t.exe
+// RUN: llvm-bolt-binary-analysis --scanners=ptrauth-pac-ret %t.exe 2>&1 \
+// RUN:   | FileCheck --check-prefix=SUMMARY %s
+// RUN: llvm-bolt-binary-analysis --scanners=ptrauth-pac-ret -v=1 %t.exe 2>&1 \
+// RUN:   | FileCheck --check-prefix=VERBOSE %s
+
+        .text
+
```
atrosinenko, line 11: I would rather add a "good" example not triggering any reports, something along the lines
```
        .globl  f_good
        .type   f_good,@function
f_good:
        ret
.size f_good, .-f_good
// SUMMARY-NOT: BOLT-WARNING:{{.*}}f_good
// VERBOSE-NOT: BOLT-WARNING:{{.*}}f_good
```

atrosinenko (commented):
Thank you for the updates, I added a few more comments, but the patch looks good overall.

bolt/lib/Rewrite/RewriteInstance.cpp
devnexen: hehe yes I did wrong vim commands :)

bolt/lib/Rewrite/RewriteInstance.cpp
atrosinenko: Oh, I'm really sorry, I tested the latest version locally once more and it turned out PLT stubs are indeed reported as lacking CFG...

Well, maybe checking `BF.isIgnored()` here is not the best idea (and thus might still deserve a comment), but it seems that there is not much we can check for an `ignored` function anyway :) On ELF targets CFG reconstruction is not even [attempted](https://github.com/llvm/llvm-project/blob/cafd16fc6e5e1556fd9962c8b2e4b4a6719dc68a/bolt/lib/Rewrite/RewriteInstance.cpp#L3837) if [`shouldDisassemble(BF)`](https://github.com/llvm/llvm-project/blob/main/bolt/lib/Rewrite/RewriteInstance.cpp#L502) returns `false`, which is generally the case for `pseudo` functions, except for PLT stubs in a **BTI-enabled** binary (which is not necessary true even for a binary that is expected to be fed to PAuth gadget scanner).

atrosinenko (approved):
Looks good, thank you!

PR #197048, opened 2026-05-11
Move lldb-python.h to be the first included as its comment demands.

jimingham:
The comment was preserved even though the header was almost never actually placed first.  For the most part that didn't matter, but you can through bad luck uncover the conflict between Python typedefs (in this case of pid_t) and ones from include/lldb/Host/windows/PosixAPI.h.

So this patch just moves this header to the front in all its appearances.

PR #197298, opened 2026-05-12
[lldb][windows] remove mandatory ordering of the lldb-python.h header

charles-zablit:
`PosixApi.h` typedef'd `pid_t` as `uint32_t`, while Python's `pyconfig.h` on Windows  typedef's it as `int`. C++ forbids redeclaring a typedef with a different type, so the two headers cannot coexist. The `NO_PID_T` macro in `lldb-python.h` suppressed LLDB's typedef, but only if `lldb-python.h` got included before `PosixApi.h`.

`pid_t` on Windows was originally defined in d87fc157d2b7. At this time, there was no Python support for LLDB on Windows and `uint32_t` matches the `DWORD` type used by the Win32 API for process IDs.

This patch matches the Python type in `PosixApi.h`, removing the need for the include ordering.

This is a follow up to https://github.com/llvm/llvm-project/pull/197048.

### DIFF
--- a/lldb/include/lldb/Host/windows/PosixApi.h
+++ b/lldb/include/lldb/Host/windows/PosixApi.h
@@ -21,6 +21,8 @@
 // time_t, timespec, etc.
 #include <ctime>
 
+#include <sys/types.h>
+
 #ifndef PATH_MAX
 #define PATH_MAX 32768
 #endif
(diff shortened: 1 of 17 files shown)


jimingham (approved):
LGTM, thanks!

PR #197302, opened 2026-05-12
[llvm] Sort the Subtarget feature implies list by name

jroelofs:


jroelofs:
Motivation for this came from: https://github.com/llvm/llvm-project/pull/197115/changes#r3228102644

tomershafir (approved):

jroelofs:
Oops. Accidentally merged that from the UI, instead of through SPR, so it landed in the wrong branch.

https://github.com/llvm/llvm-project/pull/197700

PR #197303, opened 2026-05-12
[llvm] Add a tablegen !sort operator

jroelofs:
This operator creates a new ``list`` containing the same elements as *list*
but in sorted order. To determine the order, TableGen binds the variable
*var* to each element and evaluates the *key* expression, which presumably
refers to *var*. The key must produce a ``string`` or integer value
(``bit``, ``bits``, or ``int``); all keys must be of the same type. Elements
with equal keys preserve their original relative order, resulting in a stable
sort.

For example, to sort a list of records by their ``Name`` field::

`  list<Thing> sorted = !sort(t, Things, t.Name);`

### DIFF
--- a/llvm/docs/TableGen/ProgRef.rst
+++ b/llvm/docs/TableGen/ProgRef.rst
@@ -2009,7 +2009,6 @@ and non-0 as true.
     The size of a DAG is the number of arguments; the operator does not count.
 
 ``!sort(``\ *var*\ ``,`` *list*\ ``,`` *key*\ ``)``
-
     This operator creates a new ``list`` containing the same elements as *list*
     but in sorted order. To determine the order, TableGen binds the variable
     *var* to each element and evaluates the *key* expression, which presumably
--- a/llvm/docs/TableGen/ProgRef.rst
+++ b/llvm/docs/TableGen/ProgRef.rst
@@ -2008,6 +2008,20 @@ and non-0 as true.
     This operator produces the size of the string, list, or dag *a*.
     The size of a DAG is the number of arguments; the operator does not count.
 
+``!sort(``\ *var*\ ``,`` *list*\ ``,`` *key*\ ``)``
+
+    This operator creates a new ``list`` containing the same elements as *list*
+    but in sorted order. To determine the order, TableGen binds the variable
+    *var* to each element and evaluates the *key* expression, which presumably
+    refers to *var*. The key must produce a ``string`` or integer value
+    (``bit``, ``bits``, or ``int``); all keys must be of the same type. Elements
+    with equal keys preserve their original relative order, resulting in a
+    stable sort.
+
+    For example, to sort a list of records by their ``Name`` field::
+
+      list<Thing> sorted = !sort(t, Things, t.Name);
+
 ``!sra(``\ *a*\ ``,`` *count*\ ``)``
     This operator shifts *a* right arithmetically by *count* bits and produces the resulting
     value. The operation is performed on a 64-bit integer; the result
--- a/llvm/docs/TableGen/ProgRef.rst
+++ b/llvm/docs/TableGen/ProgRef.rst
@@ -2019,6 +2019,8 @@ and non-0 as true.
 
     For example, to sort a list of records by their ``Name`` field::
 
+    .. code-block:: text
+
       list<Thing> sorted = !sort(t, Things, t.Name);
 
 ``!sra(``\ *a*\ ``,`` *count*\ ``)``
--- a/llvm/lib/TableGen/Record.cpp
+++ b/llvm/lib/TableGen/Record.cpp
@@ -1809,11 +1809,13 @@ static const Init *SortHelper(const Init *LHS, const Init *MHS, const Init *RHS,
     return ListInit::get({}, cast<ListRecTy>(Type)->getElementType());
 
   // Determine key type from the first element; all keys must agree.
-  bool UseInt = dyn_cast_or_null<IntInit>(
-                    KeyedList[0].first->convertInitializerTo(IntRecTy::get(RK))) != nullptr;
+  bool UseInt =
+      dyn_cast_or_null<IntInit>(KeyedList[0].first->convertInitializerTo(
+          IntRecTy::get(RK))) != nullptr;
   for (auto &[Key, Item] : KeyedList) {
     if (UseInt) {
-      if (!dyn_cast_or_null<IntInit>(Key->convertInitializerTo(IntRecTy::get(RK))))
+      if (!dyn_cast_or_null<IntInit>(
+              Key->convertInitializerTo(IntRecTy::get(RK))))
         return nullptr;
     } else {
       if (!isa<StringInit>(Key))
@@ -1823,9 +1825,12 @@ static const Init *SortHelper(const Init *LHS, const Init *MHS, const Init *RHS,
 
   llvm::stable_sort(KeyedList, [&RK, UseInt](const KV &A, const KV &B) {
     if (UseInt)
-      return cast<IntInit>(A.first->convertInitializerTo(IntRecTy::get(RK)))->getValue() <
-             cast<IntInit>(B.first->convertInitializerTo(IntRecTy::get(RK)))->getValue();
-    return cast<StringInit>(A.first)->getValue() < cast<StringInit>(B.first)->getValue();
+      return cast<IntInit>(A.first->convertInitializerTo(IntRecTy::get(RK)))
+                 ->getValue() <
+             cast<IntInit>(B.first->convertInitializerTo(IntRecTy::get(RK)))
+                 ->getValue();
+    return cast<StringInit>(A.first)->getValue() <
+           cast<StringInit>(B.first)->getValue();
   });
 
   SmallVector<const Init *, 8> Result;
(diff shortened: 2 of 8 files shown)


jroelofs:
Motivation for this came from: https://github.com/llvm/llvm-project/pull/197115/changes#r3228102644

llvm/docs/TableGen/ProgRef.rst
```diff
@@ -2008,6 +2008,20 @@ and non-0 as true.
     This operator produces the size of the string, list, or dag *a*.
     The size of a DAG is the number of arguments; the operator does not count.
 
+``!sort(``\ *var*\ ``,`` *list*\ ``,`` *key*\ ``)``
+
```
tomershafir:
```suggestion
```

llvm/lib/TableGen/Record.cpp
```diff
@@ -1788,6 +1788,57 @@ static const Init *FilterHelper(const Init *LHS, const Init *MHS,
...
+  }
+
+  if (KeyedList.empty())
+    return ListInit::get({}, cast<ListRecTy>(Type)->getElementType());
+
+  // Determine key type from the first element; all keys must agree.
+  bool UseInt =
+      dyn_cast_or_null<IntInit>(KeyedList[0].first->convertInitializerTo(
+          IntRecTy::get(RK))) != nullptr;
+  for (auto &[Key, Item] : KeyedList) {
```
tomershafir, line 1815:
```suggestion
  for (auto &[Key, Item] : llvm::drop_begin(KeyedList)) {
```

llvm/lib/TableGen/TGParser.cpp
```diff
@@ -2571,10 +2572,11 @@ const Init *TGParser::ParseOperationFind(Record *CurRec,
...
 }
 
-/// Parse the !foreach and !filter operations. Return null on error.
+/// Parse the !foreach, !filter, and !sort operations. Return null on error.
 ///
 /// ForEach ::= !foreach(ID, list-or-dag, expr) => list<expr type>
-/// Filter  ::= !foreach(ID, list, predicate) ==> list<list type>
+/// Filter  ::= !filter(ID, list, predicate) ==> list<list type>
+/// Sort    ::= !sort(ID, list, key-expr) ==> list<list type>
 const Init *TGParser::ParseOperationForEachFilter(Record *CurRec,
```
tomershafir:
```suggestion
const Init *TGParser::ParseOperationForEachFilterSort(Record *CurRec,
```

llvm/test/TableGen/sort.td
```diff
@@ -0,0 +1,84 @@
...
+// RUN: not llvm-tblgen -DERROR_KEYTYPE %s 2>&1 | FileCheck --check-prefix=ERROR_KEYTYPE %s
+
+defvar EmptyStrings = []<string>;
+defvar EmptyInts = []<int>;
+defvar Creatures = ["cephalopod", "axolotl", "dragonfly", "barracuda"];
+defvar Nums = [5, 2, 8, 1, 3];
+
+// Stable sort: equal keys preserve original relative order.
+// CHECK-LABEL: def duplicates
+// CHECK:   stable = ["a", "a", "b", "b", "c"];
```
tomershafir: I think this check cannot really distinguish between duplicates, their relative order may not have been preserved. I think a more reliable way could be something like:

```
// Stable sort: equal keys preserve original relative order.
// CHECK-LABEL: def duplicates
// CHECK:   stable = ["b", "a"];
defvar DuplicateWords = ["b", "a"];
def duplicates {
  list<string> stable = !sort(item, DuplicateWords, 1);
}
```

llvm/lib/TableGen/TGParser.cpp
```diff
@@ -2688,14 +2695,16 @@ const Init *TGParser::ParseOperationForEachFilter(Record *CurRec,
...
-  } else if (Operation == tgtok::XFilter) {
+  } else if (Operation == tgtok::XFilter || Operation == tgtok::XSort) {
     OutType = InEltType->getListTy();
   }
 
-  return (TernOpInit::get((Operation == tgtok::XForEach) ? TernOpInit::FOREACH
-                                                         : TernOpInit::FILTER,
-                          LHS, MHS, RHS, OutType))
-      ->Fold(CurRec);
+  TernOpInit::TernaryOp Opc = TernOpInit::FOREACH;
```
tomershafir: I think we may not want to speculatively assign `FOREACH` in any case, but uniformly guard all of the operations, unless we have a reason to bias towards it

llvm/test/TableGen/sort.td
```diff
@@ -0,0 +1,84 @@
...
+}
+def thing_c : Thing<"charlie", 30>;
+def thing_a : Thing<"alpha",   10>;
+def thing_b : Thing<"beta",    20>;
+
+defvar Things = [thing_c, thing_a, thing_b];
+
+// CHECK-LABEL: def key_expr
+// CHECK:   by_name  = [thing_a, thing_b, thing_c];
+// CHECK:   by_value = [thing_a, thing_b, thing_c];
```
tomershafir: I think the 2 sort orders by int or string should not match, to make sure we dont accidentally sort by the other field. Maybe swap 2 names or values

llvm/test/TableGen/sort.td
```diff
@@ -0,0 +1,84 @@
...
+  string Name = N;
+  int Value = V;
+}
+def thing_c : Thing<"charlie", 30>;
+def thing_a : Thing<"alpha",   10>;
+def thing_b : Thing<"beta",    20>;
+
+defvar Things = [thing_c, thing_a, thing_b];
+
+// CHECK-LABEL: def key_expr
```
tomershafir: Maybe hoist this check block a bit, so that we dont get interleaved check lines and TableGen code for this test case?

llvm/test/TableGen/sort.td
```diff
@@ -0,0 +1,84 @@
...
+  list<string> one = !sort(item, ["only"], item);
+}
+
+// CHECK-LABEL: def sorted
+// CHECK:   sorted_strings = ["axolotl", "barracuda", "cephalopod", "dragonfly"];
+// CHECK:   sorted_ints = [1, 2, 3, 5, 8];
+// CHECK:   sorted_empty_strings = [];
+// CHECK:   sorted_empty_ints = [];
+
+def sorted {
```
tomershafir, line 48: Can we sink the following defs over here, next to where they are used: EmptyStrings, EmptyInts, Creatures, Nums

llvm/lib/TableGen/Record.cpp
jroelofs: I don't think that's an improvement, since if we drop the first like this, we lose the `int ^ string` check on the key type for single-element lists.

llvm/lib/TableGen/TGParser.cpp
```diff
@@ -2628,9 +2630,11 @@ const Init *TGParser::ParseOperationForEachFilter(Record *CurRec,
     InEltType = InListTy->getElementType();
     if (ItemType) {
       if (const auto *OutListTy = dyn_cast<ListRecTy>(ItemType)) {
-        ExprEltType = (Operation == tgtok::XForEach)
-                          ? OutListTy->getElementType()
-                          : IntRecTy::get(Records);
+        ExprEltType =
```
fpetrogalli: nit: maybe a switch statement is more readable?

llvm/test/TableGen/sort.td
```diff
@@ -0,0 +1,84 @@
...
+def myop;
+defvar mydag = (myop);
+def err_nonlist {
+  list<string> bad = !sort(x, mydag, !cast<string>(x));
+}
+#endif
+
+#ifdef ERROR_KEYTYPE
+// Key that cannot be resolved to int or string leaves the op unfolded.
+// ERROR_KEYTYPE: could not be fully resolved
```
fpetrogalli: There is no Loc that points at what cannot be resolved? I think you should check that too if there is one

llvm/test/TableGen/sort.td
fpetrogalli: I think that best would be to sort instances of classes that have different data in it, if not for the field that is used to sort.

llvm/lib/TableGen/TGParser.cpp
fpetrogalli: Or something that makes it generic enough to avoid having to rename it if we end up adding a 4th operator that has the same List->List signature, say `ParseOperationListToList`

llvm/test/TableGen/sort.td
```diff
@@ -0,0 +1,70 @@
...
+// CHECK:   by_value = [thing_a10, thing_a20, thing_b20, thing_c30];
+class Thing<string N, int V> {
+  string Name = N;
+  int Value = V;
+}
+def thing_a10 : Thing<"alpha",   10>;
+def thing_a20 : Thing<"alpha",   20>;
+def thing_b20 : Thing<"beta",    20>;
+def thing_c30 : Thing<"charlie", 30>;
+def thing_d10 : Thing<"delta",   10>;
```
tomershafir, line 25: dead definition?

llvm/test/TableGen/sort.td
```diff
@@ -0,0 +1,70 @@
...
+// RUN: not llvm-tblgen -DERROR_KEYTYPE %s 2>&1 | FileCheck --check-prefix=ERROR_KEYTYPE %s
+
+// Sort an already-sorted list — should be a no-op.
+// CHECK-LABEL: def idempotent
+// CHECK:   already = ["a", "b", "c"];
+def idempotent {
+  list<string> already = !sort(item, ["a", "b", "c"], item);
+}
+
+// Sort records by a string field or an int field.
```
tomershafir: I think that the ad-hoc sort stability check was useful (or at least add a comment that we test stability ion this test case implicitly as well)

tomershafir (approved):
LGTM with a few minor comments

PR #197304, opened 2026-05-12
[flang][cuda] Add support for -gpu=pinned

clementval:
Add a new language feature to implicitly treat dynamic allocation as pinned allocation.

### DIFF
--- a/flang/lib/Semantics/resolve-names.cpp
+++ b/flang/lib/Semantics/resolve-names.cpp
@@ -10173,18 +10173,27 @@ void ResolveNamesVisitor::FinishSpecificationPart(
         SetBindNameOn(symbol);
       }
     }
-    // Implicitly treat allocatable arrays as managed when feature is enabled.
-    // This is done after all explicit CUDA attributes have been processed.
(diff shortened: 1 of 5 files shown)


flang/lib/Semantics/resolve-names.cpp
```diff
@@ -10183,18 +10173,27 @@ void ResolveNamesVisitor::FinishSpecificationPart(
...
+        // Fortran managed descriptor pipeline.
+        if (context().languageFeatures().IsEnabled(
+                common::LanguageFeature::CudaManaged) &&
+            context().languageFeatures().IsEnabled(
+                common::LanguageFeature::CUDA))
           object->set_cudaDataAttr(common::CUDADataAttr::Managed);
+        // Implicitly treat allocatable arrays as pinned when feature is
+        // enabled.
+        else if (context().languageFeatures().IsEnabled(
+                     common::LanguageFeature::CudaPinned))
```
wangzpgi, line 10193: Should we also guard this with `-cuda`?

clementval: No because the reference compiler also enable this with acc or stdpar.

wangzpgi (approved):

PR #197305, opened 2026-05-12
[clang-tidy] modernize-use-nullptr: Add new libstdc++ compare type

pobrn:
For gcc 16, the libstdc++ type `__unspec` has been renamed to `__literal_zero`, so update the default ignore list to take that into account.

Link: https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=d9a4c7158a70896c5d4281f42310e88c987acb3e

### DIFF
--- a/clang-tools-extra/clang-tidy/modernize/UseNullptrCheck.cpp
+++ b/clang-tools-extra/clang-tidy/modernize/UseNullptrCheck.cpp
@@ -490,13 +490,16 @@ private:
   bool PruneSubtree = false;
 };
 
+static constexpr char DefaultIgnoredTypes[] = "_CmpUnspecifiedParam;"
+                                              "^std::__cmp_cat::__unspec;";
+
 } // namespace
 
 UseNullptrCheck::UseNullptrCheck(StringRef Name, ClangTidyContext *Context)
     : ClangTidyCheck(Name, Context),
       NullMacrosStr(Options.get("NullMacros", "NULL")),
-      IgnoredTypes(utils::options::parseStringList(Options.get(
-          "IgnoredTypes", "_CmpUnspecifiedParam;^std::__cmp_cat::__unspec"))) {
+      IgnoredTypes(utils::options::parseStringList(
+          Options.get("IgnoredTypes", DefaultIgnoredTypes))) {
   NullMacrosStr.split(NullMacros, ",");
 }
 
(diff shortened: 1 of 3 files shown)


clang-tools-extra/clang-tidy/modernize/UseNullptrCheck.cpp
```diff
@@ -490,13 +490,18 @@ class CastSequenceVisitor : public RecursiveASTVisitor<CastSequenceVisitor> {
   bool PruneSubtree = false;
 };
 
+static constexpr char DefaultIgnoredTypes[] =
+    "_CmpUnspecifiedParam;"
+    "^std::__cmp_cat::__unspec;"
+    "^std::__cmp_cat::__literal_zero;";
```
pobrn, line 496: Should this have `$` at the end? The other libstdc++ does not have since its introduction, but I'm not sure why.

clang-tools-extra/test/clang-tidy/checkers/modernize/use-nullptr-cxx20.cpp
```diff
@@ -1,29 +1,25 @@
...
-// RUN: %check_clang_tidy -std=c++20-or-later %s modernize-use-nullptr %t -- -- -DCLANG
+// RUN: %check_clang_tidy -std=c++20-or-later %s modernize-use-nullptr %t -- -- -D UNSPECIFIED_TYPE=_CmpUnspecifiedParam
+// RUN: %check_clang_tidy -std=c++20-or-later %s modernize-use-nullptr %t -- -- -D UNSPECIFIED_TYPE=__cmp_cat::__unspec
+// RUN: %check_clang_tidy -std=c++20-or-later %s modernize-use-nullptr %t -- -- -D UNSPECIFIED_TYPE=__cmp_cat::__literal_zero
 
 namespace std {
 class strong_ordering;
 
 // Mock how STD defined unspecified parameters for the operators below.
-#ifdef CLANG
```
unterumarmung, line 8: why is it removed?

pobrn: I considered two approaches to extending the test:

1. add a new `#ifdef GCC_NEW` (or so) block that duplicates essentially the `#ifdef GCC` block;
2. remove all conditional blocks, always define all types, and set `UNSPECIFIED_TYPE` on the compiler command line.

Option (2) seemed better to me, so I went with that.

PR #197306, opened 2026-05-12
workflows: Require hashed dependencies when installing docs dependencies

tstellar:


.github/workflows/docs.yml
```diff
@@ -100,9 +100,9 @@ jobs:
         with:
           python-version: '3.14'
           cache: 'pip'
-          cache-dependency-path: 'llvm/docs/requirements-hashed.txt'
+          cache-dependency-path: 'llvm/docs/requirements.txt'
       - name: Install python dependencies
-        run: pip install -r llvm/docs/requirements-hashed.txt
+        run: pip install -r llvm/docs/requirements.txt
```
boomanaiden154: `--require-hashes` here too?

boomanaiden154 (approved):
LGTM.

PR #197308, opened 2026-05-12
[GlobalISel] Remove LegacyLegalizerInfo

davemgreen:
After #197238, #197370, #197371, #197374, #197375, #197377, #197378 and #197379, this removes the legacy rules from global isel.

The above patches attempt to make all _tested_ operations legal, but have only gone through the operations that have llvm tests. If more fallbacks are now found to be happening, the other operations from #197238 might be needed.


arsenm (approved):

e-kud:
I see up to 17% fallback to SDAG rate increase on X86. Let me understand whether it is the wrong stat collected or the real problem.

davemgreen:
Hi - thanks for checking. It sounds like some other node doesn't have a action builder, but isn't testing in any of the tests. Maybe one of the ones from https://github.com/llvm/llvm-project/pull/197238/changes?

e-kud:
> Hi - thanks for checking. It sounds like some other node doesn't have a action builder, but isn't testing in any of the tests. Maybe one of the ones from https://github.com/llvm/llvm-project/pull/197238/changes?

Yes, you are completely right! `G_INVOKE_REGION_START` is the one causing the regression. Once added `.alwaysLegal`, everything returned back.

e-kud (approved):
LGTM

PR #197310, opened 2026-05-12
[mlir][AMDGPU] Move memory access op folding to memref interfaces

krzysz00:
This PR implements IndexedAccessOpInterface and
IndexedMemCopyOpInterface for relevant ops in the AMDGPU dialect, removing the custom folding pass we used to have now that there's interfaces for this sort of thing.

As a result:

- The in-bonuds semantics of various AMDGPU ops have been clarified
- Interface methods to enable oob checks on DMA operations have been added (to prevent accidental `disjoint`ing and the like)
- Said memref rewrite patterns have been hardened to allow for mixed tensor/memref semantics.
- Helpers for detecting memory spaces were factored out of `AMDGPUOps.cpp` so that they could be re-used in the interface implementations.

# Breaking changes / migration

The custom `amdgpu-fold-memrefs-ops` pass and its patterns are removed. Users should move to the general FoldMemRefAliasOps.cpp pass on the memref dialect instead, as these special-case patterns are no longer needed.


mlir/lib/Dialect/AMDGPU/Transforms/MemoryAccessOpInterfacesImpl.cpp
```diff
@@ -0,0 +1,258 @@
...
+                         ValueRange newIndices) const {
+    auto accessOp = cast<OpTy>(op);
+    rewriter.modifyOpInPlace(accessOp, [&]() {
+      accessOp.getSrcMutable().assign(newMemref);
+      accessOp.getSrcIndicesMutable().assign(newIndices);
+    });
+    return std::nullopt;
+  }
+
+  bool hasInboundsIndices(Operation *) const { return true; }
```
lialan, line 53: By default we should be conservative and set it to false?

mlir/lib/Dialect/AMDGPU/Transforms/MemoryAccessOpInterfacesImpl.cpp
```diff
@@ -0,0 +1,258 @@
...
+                                              OpTy> {
+  TypedValue<MemRefType> getAccessedMemref(Operation *op) const {
+    return cast<TypedValue<MemRefType>>(cast<OpTy>(op).getBase());
+  }
+
+  Operation::operand_range getIndices(Operation *op) const {
+    return cast<OpTy>(op).getIndices();
+  }
+
+  SmallVector<int64_t> getAccessedShape(Operation *) const { return {}; }
```
lialan, line 68: Should we instead get the memref's shape here?

mlir/lib/Dialect/AMDGPU/Transforms/MemoryAccessOpInterfacesImpl.cpp
```diff
@@ -0,0 +1,258 @@
...
+    if (!memref)
+      return {};
+    return cast<TypedValue<MemRefType>>(memref);
+  }
+
+  Operation::operand_range getIndices(Operation *op) const {
+    return cast<OpTy>(op).getAtomicBarrierIndices();
+  }
+
+  SmallVector<int64_t> getAccessedShape(Operation *) const { return {}; }
```
lialan, line 99: same here.

mlir/include/mlir/Dialect/AMDGPU/AMDGPUIRUtils.h
```diff
@@ -0,0 +1,61 @@
...
+#include "llvm/Support/Compiler.h"
+
+namespace mlir::amdgpu {
+
+// Integer memory-space attributes are deprecated, but still accepted here for
+// compatibility with existing IR.
+
+inline bool isGlobalMemorySpace(Attribute memorySpace) {
+  if (!memorySpace)
+    return true;
```
lialan:
```suggestion
    return false;
```

mlir/include/mlir/Dialect/AMDGPU/AMDGPUIRUtils.h
```diff
@@ -0,0 +1,61 @@
+//===- AMDGPUIRUtils.h - AMDGPU dialect IR utilities -----------*- C++ -*-===//
```
lialan: Should go into `./Utils` subfolder.

mlir/include/mlir/Dialect/AMDGPU/AMDGPUIRUtils.h
```diff
@@ -0,0 +1,61 @@
...
+
+// Integer memory-space attributes are deprecated, but still accepted here for
+// compatibility with existing IR.
+
+inline bool isGlobalMemorySpace(Attribute memorySpace) {
+  if (!memorySpace)
+    return true;
+  if (auto gpuMemorySpace = dyn_cast<gpu::AddressSpaceAttr>(memorySpace))
+    return gpuMemorySpace.getValue() == gpu::AddressSpace::Global;
+  if (LLVM_UNLIKELY(isa<IntegerAttr>(memorySpace))) {
```
lialan:
```suggestion
  if (LLVM_UNLIKELY(dyn_cast<IntegerAttr>(memorySpace))) {
```

lialan: same thing a few lines down.

mlir/include/mlir/Dialect/AMDGPU/AMDGPUIRUtils.h
krzysz00: Yeah, fair, though it's not a header for Utils/ but for IR/

mlir/include/mlir/Dialect/AMDGPU/AMDGPUIRUtils.h
krzysz00: We historically and currently assume that the flat memory space is global.

However, that probably should be a knob on the utility - I'll fix that

mlir/lib/Dialect/AMDGPU/Transforms/MemoryAccessOpInterfacesImpl.cpp
krzysz00: No, this is an instruction that points at a specific in-bounds address and traps on failure. That is, it has `memref.load` semantics

mlir/lib/Dialect/AMDGPU/Transforms/MemoryAccessOpInterfacesImpl.cpp
krzysz00: No - "accessed shape" means "the shape of the values that have to be kept together when folding in an aliasing operation". These operations access a particular point in memory, not all the memory in that memref.

mlir/lib/Dialect/AMDGPU/Transforms/MemoryAccessOpInterfacesImpl.cpp
krzysz00: This one's very clear-cut - it accesses exactly one element, which means `{}` shape. That is, you can fold a strided subview in no problem.

mlir/include/mlir/Dialect/AMDGPU/Transforms/Passes.td
```diff
@@ -71,15 +71,4 @@ def AmdgpuMaskedloadToLoadPass : Pass<"amdgpu-maskedload-to-load"> {
   ];
 }
 
-def AmdgpuFoldMemRefOpsPass : Pass<"amdgpu-fold-memrefs-ops"> {
```
lialan, line 74: I mean, this is just ... gone? Let's at least give a migration plan/idea in the PR description.

lialan (approved):
+1, but please update the PR description a little bit more.

mlir/lib/Dialect/AMDGPU/IR/AMDGPUOps.cpp
```diff
@@ -13,6 +13,7 @@
 
 #include "mlir/Dialect/AMDGPU/IR/AMDGPUDialect.h"
 
+#include "mlir/Dialect/AMDGPU/Utils/MemorySpaceUtils.h"
```
rupprecht, line 16: fyi this seems to be a [layering violation](https://llvm.org/docs/CodingStandards.html#library-layering) because now there is a IR<->Utils cycle

krzysz00: You're welcome to followup by moving it back or moving it to `IR/Utils` ro something

krzysz00: ... or to move it to `mlir/lib/Dialect/AMDGPU` since it's internal de-duplication ... even though these are useful functions elsewhere

PR #197311, opened 2026-05-12
[lldb] Increase availability of ValueObject::GetParent

kastiglione:
While working formatter bytecode, one of the C++ formatters needs `GetParent`. While adding `GetParent` support in the bytecode, I noticed the SB API also does not expose `GetParent`. This remedies that.

During review of this PR, it was pointed out that `GetParent` does work with synthetic value objects. This PR also addresses that shortcoming.

Assisted-by: claude


jimingham:
It would be good to add a test with one of the standard synthetic child providers (like std::vector) and see what the parent of that is.  It should be the std::vector SBValue.  I'm not 100% sure the synthetic child providers always do this right.

kastiglione:
good call, the test fails with std::vector. I will investigate.

kastiglione:
@jimingham Updated to make `GetParent` work for children of synthetic value objects.

jimingham (approved):
LGTM

DavidSpickett:
The new test is failing on AArch64 Linux using libstdc++.

```
FAIL: test (TestSBValueGetParent.TestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/davspi01/llvm-project/lldb/test/API/python_api/sbvalue_get_parent/TestSBValueGetParent.py", line 20, in test
    self._do_test(frame, "vec", 0)
  File "/home/davspi01/llvm-project/lldb/test/API/python_api/sbvalue_get_parent/TestSBValueGetParent.py", line 40, in _do_test
    self.assertEqual(child_parent.name, parent_name)
AssertionError: '_M_start' != 'vec'
- _M_start
+ vec
```
My guess is that indexing goes through more layers than it did wherever you were running the test:
```
(Pdb) parent
(std::vector<int>) vec = size=3 {
  [0] = 10
  [1] = 20
  [2] = 30
}
(Pdb) child = parent.GetChildAtIndex(0)
(Pdb) child
(int) [0] = 10
(Pdb) child_parent = child.GetParent()
(Pdb) child_parent
(std::_Vector_base<int, std::allocator<int> >::pointer) _M_start = 0x0000c9a17c695eb0
```
Though x86 Linux CI likely would have complained too, I can't confirm it ran there but I don't see a reason that it wouldn't.

DavidSpickett:
To get back to `vec` I have to get the 5th parent:
```
(Pdb) child.GetParent().GetParent().GetParent().GetParent().GetParent()
(std::vector<int>) vec = size=3 {
  [0] = 10
  [1] = 20
  [2] = 30
}
```
If I load the program file manually I see:
```
>>> parent = lldb.frame.FindVariable('vec')
>>> parent
(std::vector<int>) vec = size=3 {
  [0] = 10
  [1] = 20
  [2] = 30
}
>>> parent.GetChildAtIndex(0)
(int) [0] = 10
>>> child = parent.GetChildAtIndex(0)
>>> child.GetParent()
(std::_Vector_base<int, std::allocator<int> >::pointer) _M_start = 0x0000aaaaaaac5eb0
```
The provider is presumably going through all those levels for us, but when we try to go up one level we go up one level in the raw type.

If I disable cplusplus formatters and try again (in the same session):
```
>>> parent
(std::vector<int>) vec = {
  std::_Vector_base<int, std::allocator<int> > = {
    _M_impl = {
      std::_Vector_base<int, std::allocator<int> >::_Vector_impl_data = {
        _M_start = 0x0000aaaaaaac5eb0
        _M_finish = 0x0000aaaaaaac5ebc
        _M_end_of_storage = 0x0000aaaaaaac5ebc
      }
    }
  }
}
>>> parent.GetChildAtIndex(0)
(std::_Vector_base<int, std::allocator<int> >) std::_Vector_base<int, std::allocator<int> > = {
  _M_impl = {
    std::_Vector_base<int, std::allocator<int> >::_Vector_impl_data = {
      _M_start = 0x0000aaaaaaac5eb0
      _M_finish = 0x0000aaaaaaac5ebc
      _M_end_of_storage = 0x0000aaaaaaac5ebc
    }
  }
}
>>> parent.GetChildAtIndex(0).GetParent()
(std::vector<int>) vec = {
  std::_Vector_base<int, std::allocator<int> > = {
    _M_impl = {
      std::_Vector_base<int, std::allocator<int> >::_Vector_impl_data = {
        _M_start = 0x0000aaaaaaac5eb0
        _M_finish = 0x0000aaaaaaac5ebc
        _M_end_of_storage = 0x0000aaaaaaac5ebc
      }
    }
  }
}
```
Which would pass the test in theory, but probably is not what we want because I think you want this to work for synthetic objects as well.

(also I couldn't find the right place to put `type category disable cplusplus` anyway, it seems to work in the terminal but not in a test case)

kastiglione:
thanks for the report and investigation, I will follow up later this morning.

kastiglione:
https://github.com/llvm/llvm-project/pull/198859

PR #197312, opened 2026-05-12
[lldb][bytecode] Add GetParent and Clone selectors

kastiglione:
`GetParent` and `Clone` are needed to implement a `std::optional<T>` data formatter for libc++.


kastiglione:
Partially depends on #197311

PR #197314, opened 2026-05-12
[CIR] Lower materialize-global-temporaries that self-reference

erichkeane:
This showed up in a test suite, but cases where a materialized global temporary references itself, we emitted an NYI.  This patch implements it by creating a 'throw-away' global op that can be referenced by name, then erased/thrown away.  Since it is referenced by name in global-view, all of the uses I could reproduce didn't require a 'replaceAllUsesWith', but the call is left in case some sort of initialization causes a reference directly to the global.

Also, as a drive-by, `createGlobalOp` was a static member function that took the CIRGenModule as an argument for some reason, so this patch just makes it a member function.


andykaylor (approved):
lgtm -- my only concern is that maybe the changes to make createGlobalOp non-static should be a separate commit to keep the revision history clean, but I don't have strong feelings about that.

erichkeane:
> lgtm -- my only concern is that maybe the changes to make createGlobalOp non-static should be a separate commit to keep the revision history clean, but I don't have strong feelings about that.

Split that off into a separate commit, so going to merge this when it passes CI.

PR #197316, opened 2026-05-12
[Bazel]: Pull from Bazel-Central-Registry for third party deps.

UebelAndre:
The majority of these dependencies are available in the [Bazel-Central-Registry](https://github.com/bazelbuild/bazel-central-registry) (BCR) and to improve build performance for bzlmod users, llvm-project should pull from the BCR to consolidate targets.


keith:
LGTM, @rupprecht @aaronmondal any issues with this?

michaelrj-google:
Seems fine to me, CC @lntue

rupprecht (approved):
Yeah, sorry for the delay -- I don't see any problems w/ this, it should be an improvement. Thanks!

PR #197318, opened 2026-05-12
[AMDGPU] Optimize fcanonicalize/fneg/fsub with packed bf16 math ops

changpeng:
This work makes fcanonicalize v2bf16  'Legal' and implements the selection pattern for it with
v_pk_max_num_bf16.

  We also make fneg and fabs 'Legal' in this patch. With this change, packed fadd can be selected
for vector fsub with bf16. Also, the vector fneg can be successfully folded into the operand in the
packed bf16 math ops.


llvm/test/CodeGen/AMDGPU/bf16.ll
```diff
@@ -23797,6 +23814,37 @@ define <32 x bfloat> @v_minnum_v32bf16(<32 x bfloat> %a, <32 x bfloat> %b) {
 ; GFX1250-NEXT:    s_wait_loadcnt_dscnt 0x0
 ; GFX1250-NEXT:    s_wait_kmcnt 0x0
 ; GFX1250-NEXT:    scratch_load_b32 v31, off, s32
+; GFX1250-NEXT:    v_pk_max_num_bf16 v16, v16, v16
```
shiltian: Why do we have these many extra instructions compared to before?

changpeng: > Why do we have these many extra instructions compared to before?

These are to canonicalize the operands to handle NaNs. If the call is with 'nnan" flag, we won't need canonicalization.

arsenm (commented):
We do need some work here, but I think this is breaking things in the wrong direction. I'm 90% sure that gfx12 just rug pulled the behavior of the min and max instructions. Pre-gfx12, we had the IEEE 754 2008 behavior instructions. With the IEEE mode bit enabled, we need the additional quieting instructions to match the behavior of minimumnum/maximumnum. Post-gfx12, I believe the instructions have the IEEE 754 2019 behavior (which exactly match minimumnum/maximumnum) already. 

We also need to switch to directly selecting llvm.minnum/maxnum to the instructions, as per the recent definition change for the nondeterministic signaling nan behavior. We do however have to either fix the OpenCL implementation to use minimumnum/maximumnum, or fix the conformance test to actually test the nondeterministic snan behavior the spec wording suggests should be legal

arsenm (commented):
Also, this should really touch min and max in a fully separate change from any other FP operation

changpeng:
> Also, this should really touch min and max in a fully separate change from any other FP operation

Agreed! We should work on these min/max to eliminate unnecessary canonicalizations with recent hardwares.
So, for this case of packed bfloat, I am keeping FMINNUM/FMAX 'Legal; as it was (before the patch. And the remaining optimizations are
1) make fcanonocalize 'Legal", and use v_max_num_nf16 to implement the selection pattern.
2) make FNEG legal to 
    a) Allow fneg to be folded into the packed bf16 ops
    b) Select v_pk_add_bf16 for fsub bf16 vectors

Need to investigate min/max further as separate work.

llvm/lib/Target/AMDGPU/SIInstructions.td
```diff
@@ -3733,6 +3733,12 @@ multiclass SelectCanonicalizeAsMax<
     // FIXME: Should have VOP3P subtarget predicate
     let OtherPredicates = f16_preds;
   }
+
+let SubtargetPredicate = HasBF16PackedInsts in
+  def : GCNPat<
+    (fcanonicalize (v2bf16 (VOP3PMods v2bf16:$src, i32:$src_mods))),
+    (V_PK_MAX_NUM_BF16 $src_mods, $src, $src_mods, $src, DSTCLAMP.NONE)
```
arsenm: This would probably be better encoded as `v_pk_mul_bf16 dst, 1.0, x` to avoid the second register use

llvm/test/CodeGen/AMDGPU/packed-fneg-fsub-bf16.ll
```diff
@@ -0,0 +1,589 @@
...
+; GFX1310-NEXT:    s_wait_kmcnt 0x0
+; GFX1310-NEXT:    v_pk_max_num_bf16 v0, v0, v4 neg_lo:[0,1] neg_hi:[0,1]
+; GFX1310-NEXT:    v_pk_max_num_bf16 v1, v1, v5 neg_lo:[0,1] neg_hi:[0,1]
+; GFX1310-NEXT:    v_pk_max_num_bf16 v2, v2, v6 neg_lo:[0,1] neg_hi:[0,1]
+; GFX1310-NEXT:    v_pk_max_num_bf16 v3, v3, v7 neg_lo:[0,1] neg_hi:[0,1]
+; GFX1310-NEXT:    s_set_pc_i64 s[30:31]
+  %neg = fneg <8 x bfloat> %b
+  %max = call <8 x bfloat> @llvm.maxnum.v8bf16( <8 x bfloat> %a, <8 x bfloat> %neg)
+  ret <8 x bfloat> %max
+}
```
arsenm: We test minnum/maxnum, but not minimumnum/maximumnum and minimum/maximum?

changpeng: > We test minnum/maxnum, but not minimumnum/maximumnum and minimum/maximum?

We are working on v_pk_max_num_bf16 and  v_pk_min_num_bf16. I am not sure whether minimum/maximum
should generate these instructions, but we definitely need some work for minimumnum/maximumnum.

In short, we need to clarify the exact difference among these three set of intrinsics. But yes, we mainly test  minnum/maxnum for most cases.

llvm/lib/Target/AMDGPU/SIInstructions.td
changpeng: > This would probably be better encoded as `v_pk_mul_bf16 dst, 1.0, x` to avoid the second register use

I though we prefer v_pk_max_num_* when they are legal? 

>// Prefer selecting to max when legal, but using mul is always valid.
> let AddedComplexity = -5 in {

llvm/test/CodeGen/AMDGPU/packed-fneg-fsub-bf16.ll
arsenm: The difference is quite clear, all 3 need to be supported

changpeng: > The difference is quite clear, all 3 need to be supported

You are right. However I think we need proper lowering of these intrinsics because we can do any tests.
I would restrict this work to optimization with fneg and fcanonicalize.

llvm/lib/Target/AMDGPU/SIInstructions.td
arsenm: It doesn't really matter, the judgement should be based on power which I have no data on

llvm/test/CodeGen/AMDGPU/packed-fneg-fsub-bf16.ll
arsenm: You are changing the legalization actions for minnum and maxnum in this PR. This should not be in this PR

arsenm (changes requested):
Patch content is still going beyond what the title says. You should separate out the min and max changes

changpeng:
Okay. I am going to remove the legalization rule change for ISD::FMAXNUM and ISD::FMINNUM.
Then I can not test fneg with llvm.minnum and llvm.maxnum. Will do in a following patch. Thanks.

changpeng:
Now this patch just makes fcanonicalize/fneg/fabs legal, and does corresponding optimizations and instruction selections.

llvm/lib/Target/AMDGPU/SIInstructions.td
changpeng: > It doesn't really matter, the judgement should be based on power which I have no data on

Done with v_pk_mul_bf16

llvm/test/CodeGen/AMDGPU/packed-fneg-fsub-bf16.ll
changpeng: Keep the legalization as it is for minnum and maxnum.

arsenm (approved):

PR #196833, opened 2026-05-10
[tsan][go]: add a128 store, load, compare_exchange

mauri870:
This commit adds a128 load, store and compare_exchange to the Go sanitizer functions. This is needed for a new Go atomic type that operate on 16-byte values. 

The SANITIZER_GO macro was gating a128 and is now lifted. Based on the __TSAN_HAS_INT128 macro definition, the only architecture excluded by it that intersects with [Go race support ](https://go.dev/doc/articles/race_detector#Requirements) is s390x.

For https://github.com/golang/go/issues/61236

PR #197319, opened 2026-05-12
[tsan] Enable __TSAN_HAS_INT128 on s390x and mips64

mauri870:
The s390x exclusion was added in https://github.com/llvm/llvm-project/commit/b17673816d7f65e07015489993b22049e36b04db (https://reviews.llvm.org/D105629)
citing lack of hardware 128-bit atomics. s390x does have the CDSG instruction,
but `alignof(__int128) == 8` on the s390x psABI while CDSG requires 16-byte
alignment, so Clang does not define `__GCC_HAVE_SYNC_COMPARE_AND_SWAP_16` for
this target. The mips64 exclusion has been present since `__TSAN_HAS_INT128` was
introduced in https://github.com/llvm/llvm-project/commit/06bbca9ec4bbc5a103c94c4973797ea4ddf944b5 (https://reviews.llvm.org/D18543) with no documented
rationale. Clang similarly does not define `__GCC_HAVE_SYNC_COMPARE_AND_SWAP_16`
for mips64.

Both exclusions are unnecessary. When `__GCC_HAVE_SYNC_COMPARE_AND_SWAP_16` is
not defined, all a128 operations use the SpinMutex-based emulation path, which
serializes 128-bit accesses under a global lock without requiring any hardware
CAS16 support.

Verified with Clang (targets: s390x-linux-gnu, mips64-linux-gnuabi64,
mips64el-linux-gnuabi64): all three produce the full set of 12
`__tsan_atomic128_*` symbols with zero `__sync_*_16 / __atomic_*_16`
libcall references.

Runtime testing has not been performed as I lack access to these architectures.

Relates https://github.com/llvm/llvm-project/pull/196833
For https://github.com/golang/go/issues/61236


mauri870:
@vitalybuka Continuing with the 128-bit atomics topic, I'd also like to get this merged. s390x and mips64 already have an emulation path when no CAS16 hardware instruction is available, so there is no reason to prevent 128-bit atomics from being used on those architectures. This would also simplify the work on the Go runtime, since we wouldn't need to come up with awkward workarounds for those platforms when integrating the race detector changes into Go.

vitalybuka (approved):

PR #197323, opened 2026-05-12
[Bazel]: Replace deprecated use of `@bazel_tools//`

UebelAndre:
The platform configs in `@bazel_tools` are deprecated. This change introduces local configs and refers to [bazelbuild/platforms](https://github.com/bazelbuild/platforms) for the constraints.


keith (approved):

keith:
real ci failures

UebelAndre:
> real ci failures

The root cause was a missing config_setting. I had applied a patch from an older version and since 17.0.3 a new constraint was added. This is now fixed.

PR #197326, opened 2026-05-12
[Bazel] Pull in common incompatibility flags from BCR testing

UebelAndre:
This change brings in flags which disable legacy behavior to discourage incompatibilities being introduced for future versions of Bazel.


keith (approved):

utils/bazel/.bazelrc
```diff
@@ -46,6 +46,24 @@ build --incompatible_no_implicit_file_export
...
+common --incompatible_disable_starlark_host_transitions
+
+# Avoid the use of native Bazel rules that are planned to be exported into
+# independent repositories.
+common --incompatible_disable_native_repo_rules
+common --incompatible_autoload_externally=
+
+# Avoid reliance on the legacy `external` directory in runfiles directories
+build --nolegacy_external_runfiles
+
```
keith:
```suggestion
# Avoid reliance on the legacy `external` directory in runfiles directories
build --nolegacy_external_runfiles
```

this one actually breaks all lit tests, there is a TODO towards the top about this

UebelAndre: Ah, I see the "TODO", fixed

keith (approved):

PR #197328, opened 2026-05-12
[libc] Temporarily disable iscanonical, isnan, issignaling in shared/math.h

lntue:
These clashes with same-name macros defined in <math.h>.

### DIFF
--- a/libc/shared/math.h
+++ b/libc/shared/math.h
@@ -281,16 +281,19 @@
 #include "math/ilogbf128.h"
 #include "math/ilogbf16.h"
 #include "math/ilogbl.h"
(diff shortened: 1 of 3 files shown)


bassiounix (approved):

bassiounix:
I think we need to remove the tests too, not just the includes

lntue:
> I think we need to remove the tests too, not just the includes

Done.

ilovepi:
The build is running now and should be done fairly soon: https://ci.chromium.org/ui/p/fuchsia/builders/toolchain.ci/clang-host-linux-x64/b8681964940787253745/overview

PR #128943, opened 2025-02-26
[lldb-dap] Add process picker command to VS Code extension

matthewbastien:
Adds a process picker command to the LLDB DAP extension that will prompt the user to select a process running on their machine. It is hidden from the command palette, but can be used in an `"attach"` debug configuration to select a process at the start of a debug session. I've also added a debug configuration snippet for this called `"LLDB: Attach to Process"` that will fill in the appropriate variable substitution. e.g:

```json
{
    "type": "lldb-dap",
    "request": "attach",
    "name": "Attach to Process",
    "pid": "${command:pickProcess}"
}
```

The logic is largely the same as the process picker in the `vscode-js-debug` extension created by Microsoft. It will use available executables based on the current platform to find the list of available processes:
- **Linux**: uses the `ps` executable to list processes
- **macOS**: nearly identical to Linux except that the command line options passed to `ps` are different
- **Windows**: uses the PowerShell cmdlet`Get-CimInstance` to query for processes

I manually tested this on a MacBook Pro running macOS Sequoia, a Windows 11 VM, and an Ubuntu 22.04 VM.

Fixes #96279

PR #197330, opened 2026-05-12
[lldb-dap] Enumerate platform processes via --list-processes

JDevlieghere:
In #128943, we decided that the VS Code extension should use LLDB's platform abstraction to power the process picker. The benefits of this approach are that it's cross platform and can support listing remote processes when connected to a platform server.

In preparation for that, this PR adds a `--list-processes` flag. The host platform is used by default. The `--platform` and `--platform-url` flags allow specifying and connecting to a (remote) platform.

The result is a JSON array printed to stdout. Here's what that looks like:

```
  [ { "pid": number, "name": string, "triple": string,
      "executable": string, "user": number }, ... ]
```

Keys other than `pid` are optional. A follow-up will teach the VS Code extension to use this as input for the process picker.

### DIFF
--- a/lldb/tools/lldb-dap/tool/lldb-dap.cpp
+++ b/lldb/tools/lldb-dap/tool/lldb-dap.cpp
@@ -810,7 +810,7 @@ int main(int argc, char *argv[]) {
     }
     llvm::outs() << llvm::formatv("{0}\n",
                                   llvm::json::Value(std::move(*arr_or_err)));
-    return EXIT_FAILURE;
+    return EXIT_SUCCESS;
   }
 
   ReplMode default_repl_mode = ReplMode::Auto;
--- a/lldb/tools/lldb-dap/tool/lldb-dap.cpp
+++ b/lldb/tools/lldb-dap/tool/lldb-dap.cpp
@@ -196,7 +196,8 @@ static llvm::Error LaunchClient(const llvm::opt::InputArgList &args) {
 ///       "user": number, "executable": string }, ... ]
 ///
 /// Fields other than `pid` are omitted if not available.
-static llvm::Error ListProcesses(const llvm::opt::InputArgList &args) {
+static llvm::Expected<llvm::json::Array>
+GetProcessArray(const llvm::opt::InputArgList &args) {
   llvm::StringRef platform_name = args.getLastArgValue(OPT_platform_name);
   llvm::StringRef platform_url = args.getLastArgValue(OPT_platform_url);
 
@@ -274,8 +275,7 @@ static llvm::Error ListProcesses(const llvm::opt::InputArgList &args) {
     array.push_back(std::move(entry));
   }
 
-  llvm::outs() << llvm::formatv("{0}\n", llvm::json::Value(std::move(array)));
-  return llvm::Error::success();
+  return array;
 }
 
 llvm::Error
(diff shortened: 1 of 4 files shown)


lldb/tools/lldb-dap/tool/lldb-dap.cpp
```diff
@@ -182,6 +187,97 @@ static llvm::Error LaunchClient(const llvm::opt::InputArgList &args) {
...
+        llvm::formatv("failed to list processes: {0}", error.GetCString()));
+  }
+
+  llvm::json::Array array;
+  for (uint32_t i = 0, n = processes.GetSize(); i < n; ++i) {
+    lldb::SBProcessInfo info;
+    if (!processes.GetProcessInfoAtIndex(i, info))
+      continue;
+
+    llvm::json::Object entry;
```
DrSergei, line 259: I think it would be nice to add process arguments to the output

lldb/tools/lldb-dap/tool/lldb-dap.cpp
```diff
@@ -182,6 +187,97 @@ static llvm::Error LaunchClient(const llvm::opt::InputArgList &args) {
...
+/// Handles `--list-processes`: print the processes visible to the selected
+/// platform as a JSON array on stdout, then exit.
+///
+/// The JSON shape is intentionally minimal and stable:
+///
+///   [ { "pid": number, "name": string, "triple": string,
+///       "user": number, "executable": string }, ... ]
+///
+/// Fields other than `pid` are omitted if not available.
+static llvm::Error ListProcesses(const llvm::opt::InputArgList &args) {
```
da-viper: Nit prefer we return an `llvm::Expected<json::Array>` and print the value outside of this.

lldb/tools/lldb-dap/tool/lldb-dap.cpp
da-viper: +1 could be in a different PR as `lldb:SBProcessInfo` does not expose this currently.

JDevlieghere: I filed https://github.com/llvm/llvm-project/issues/197509

lldb/tools/lldb-dap/tool/lldb-dap.cpp
```diff
@@ -182,6 +187,97 @@ static llvm::Error LaunchClient(const llvm::opt::InputArgList &args) {
...
+    if (info.UserIDIsValid())
+      entry["user"] = info.GetUserID();
+
+    lldb::SBFileSpec exe = info.GetExecutableFile();
+    if (exe.IsValid()) {
+      std::string path = lldb_dap::GetSBFileSpecPath(exe);
+      if (!path.empty())
+        entry["executable"] = path;
+    }
+
```
da-viper, line 274: Let's add  empty `arguments` field in the json array for now so  this is not dependent on the feature request. It prevents us from checking if `X` lldb-dap version has an argument field.

JDevlieghere: Besides the pid, all the fields are meant to be optional. So there should be no need to check the version of the tool, the extension should be resilient against the key not being there. That also allows us to add more to that in the future. I think that's better than providing an empty array.

DrSergei (approved):

da-viper (approved):

PR #197334, opened 2026-05-13
[CIR] Make collectUnreachable more robust

andykaylor:
The CXXABILowering and LowerToLLVM passes both use a `collectUnreachable` function to find unreachable blocks and add them to a SmallVector that will be passed to `applyPartialConversion` to avoid leaving unconverted operations in the dead blocks. We can't simply call `eraseUnreachableBlocks` because we don't yet protect indirect branch and goto targets from being erased.

Application testing revealed that our `collectUnreachable` function was missing blocks in SCCs, because they were connected to each other. This change revises `collectUnreachable` to instead build a list of reachable blocks and then walk all blocks to find the ones that aren't in the reachable list (which is what `eraseUnreachableBlocks` does).

### DIFF
--- a/clang/include/clang/CIR/Dialect/Transforms/CIRTransformUtils.h
+++ b/clang/include/clang/CIR/Dialect/Transforms/CIRTransformUtils.h
@@ -6,12 +6,14 @@
 //
 //===----------------------------------------------------------------------===//
 
-#ifndef CIR_DIALECT_TRANSFORMS_CIRTRANSFORMUTILS_H
-#define CIR_DIALECT_TRANSFORMS_CIRTRANSFORMUTILS_H
+#ifndef LLVM_CLANG_CIR_DIALECT_TRANSFORMS_CIRTRANSFORMUTILS_H
+#define LLVM_CLANG_CIR_DIALECT_TRANSFORMS_CIRTRANSFORMUTILS_H
 
 #include "mlir/IR/Location.h"
 #include "mlir/IR/PatternMatch.h"
 #include "clang/CIR/Dialect/IR/CIRDialect.h"
+
+#include "llvm/ADT/SmallVector.h"
 
 namespace cir {
 
@@ -30,6 +32,13 @@ mlir::Block *replaceCallWithTryCall(cir::CallOp callOp, mlir::Block *unwindDest,
                                     mlir::Location loc,
                                     mlir::RewriterBase &rewriter);
 
+/// Collect ops in blocks that are unreachable from their region's entry,
+/// appending them to \p ops. Used by CIR passes that drive
+/// `applyPartialConversion` and need to feed it operations the conversion
+/// driver's dominance-order traversal would otherwise skip.
+void collectUnreachable(mlir::Operation *parent,
+                        llvm::SmallVectorImpl<mlir::Operation *> &ops);
+
 } // namespace cir
 
-#endif // CIR_DIALECT_TRANSFORMS_CIRTRANSFORMUTILS_H
+#endif // LLVM_CLANG_CIR_DIALECT_TRANSFORMS_CIRTRANSFORMUTILS_H
--- a/clang/lib/CIR/Dialect/Transforms/CIRTransformUtils.cpp
+++ b/clang/lib/CIR/Dialect/Transforms/CIRTransformUtils.cpp
@@ -6,9 +6,33 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "CIRTransformUtils.h"
+#include "clang/CIR/Dialect/Transforms/CIRTransformUtils.h"
 
 #include "clang/CIR/Dialect/IR/CIRTypes.h"
+
+#include "llvm/ADT/DepthFirstIterator.h"
+
+void cir::collectUnreachable(mlir::Operation *parent,
+                             llvm::SmallVectorImpl<mlir::Operation *> &ops) {
+  // For every region under `parent`, find the blocks unreachable from the
+  // entry via a forward CFG traversal and collect their ops.
+  llvm::df_iterator_default_set<mlir::Block *, 16> reachable;
+  parent->walk([&](mlir::Region *region) {
+    // Empty regions have no blocks; single-block regions have only the
+    // entry, which is trivially reachable. Either way, nothing to collect.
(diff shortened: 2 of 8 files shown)


andykaylor:
I've moved `collectUnreachable` to a shared location, but I left this as a separate commit for review purposes. You can see the changes in place by reviewing just the first commit in this PR.

clang/include/clang/CIR/Dialect/Transforms/CIRTransformUtils.h
```diff
@@ -30,6 +32,13 @@ mlir::Block *replaceCallWithTryCall(cir::CallOp callOp, mlir::Block *unwindDest,
                                     mlir::Location loc,
                                     mlir::RewriterBase &rewriter);
 
+/// Collect ops in blocks that are unreachable from their region's entry,
+/// appending them to \p ops. Used by CIR passes that drive
+/// `applyPartialConversion` and need to feed it operations the conversion
+/// driver's dominance-order traversal would otherwise skip.
+void collectUnreachable(mlir::Operation *parent,
```
erichkeane, line 39: Can we change this to returning the vector instead of taking it by reference? It doesn't seem required to have these created by the caller, right?

clang/lib/CIR/Dialect/Transforms/CIRTransformUtils.cpp
```diff
@@ -6,10 +6,34 @@
...
 
 #include "clang/CIR/Dialect/IR/CIRTypes.h"
 
+#include "llvm/ADT/DepthFirstIterator.h"
+
+void cir::collectUnreachable(mlir::Operation *parent,
+                             llvm::SmallVectorImpl<mlir::Operation *> &ops) {
+  // For every region under `parent`, find the blocks unreachable from the
+  // entry via a forward CFG traversal and collect their ops.
+  llvm::df_iterator_default_set<mlir::Block *, 16> reachable;
```
erichkeane, line 19: It seems we insta-clear this on line 25.  Can we just put the construction of this on line 25?

clang/lib/CIR/Dialect/Transforms/CIRTransformUtils.cpp
```diff
@@ -6,10 +6,34 @@
...
+  parent->walk([&](mlir::Region *region) {
+    // Empty regions have no blocks; single-block regions have only the
+    // entry, which is trivially reachable. Either way, nothing to collect.
+    if (region->empty() || region->hasOneBlock())
+      return;
+    reachable.clear();
+    for (mlir::Block *blk : llvm::depth_first_ext(&region->front(), reachable))
+      (void)blk;
```
erichkeane, line 36: Actually... what is gong on here?  Why are we looping through something just to do nothing with it?   Is this whole loop dead?

If so, who is adding to 'reachable'?

clang/include/clang/CIR/Dialect/Transforms/CIRTransformUtils.h
andykaylor: Currently, the caller adds the module to the vector before calling this. I suppose I could modify it to have the caller push the module after the function returns, but I think it's more general in the current form. For instance, you could add unreachable ops from a single function, and then another function with a second call.

clang/lib/CIR/Dialect/Transforms/CIRTransformUtils.cpp
andykaylor: It's not obvious, but line 25 is inside a loop. It's walking the regions within the parent operation, and for each region it walks the blocks in that region. The clear on line 25 empties the set so we don't need to compare each block to blocks in other regions when we check to see if it's in the set on line 29.

clang/lib/CIR/Dialect/Transforms/CIRTransformUtils.cpp
andykaylor: Yeah, this is super obscure. I should have added a comment because I had the exact same question when I found this code in the MLIR `eraseUnreachableBlocks` function, which this is modeled after. The `llvm::depth_first_ext` function adds each node to the set as it visits it. The closest thing I've found to documentation of this is the comment here:

https://llvm.org/doxygen/DepthFirstIterator_8h_source.html#l00020

```
/// df_ext_begin/df_ext_end/df_ext_iterator
///   * Normal depth-first iteration - visit a node and then all of its
///     children. This iterator stores the 'visited' set in an external set,
///     which allows it to be more efficient, and allows external clients to
///     use the set for other purposes.
```

clang/include/clang/CIR/Dialect/Transforms/CIRTransformUtils.h
erichkeane: Ah, I see... yeah, I think it is worth leaving as is then.

clang/lib/CIR/Dialect/Transforms/CIRTransformUtils.cpp
erichkeane: RIght, so there really isnt shared state between the  different 'regions', right?  Since it is cleared in each loop?

clang/lib/CIR/Dialect/Transforms/CIRTransformUtils.cpp
erichkeane: OH!  That IS subtle.  a range-adaptor with side effects is definitely special.  I saw that comment and didn't see what it was supposed to mean there.

clang/lib/CIR/Dialect/Transforms/CIRTransformUtils.cpp
andykaylor: Correct. It's just declared outside of the walk so we can retain the buffer across iterations.

erichkeane: *sigh*.... for those I would typically say "our optimizer/hoisting should handle that and we should let scope take care of it".  But I'm not sure I'm particularly confident in our hoisting, and this seems pretty hot-path.

erichkeane (approved):
I'm happy except for a comment explaining what is going on a bit in that 1 place, else lgtm.

PR #197338, opened 2026-05-13
[llvm-split] Generalize split-by-category to accept an arbitrary attribute name

YuriPlyakhin:
Rename the `module-id` split-by-category mode to `attribute` and add a `--category-attribute=<name>` flag that specifies which function attribute to group by. This decouples `llvm-split` from the hardcoded "module-id" attribute name, making the tool reusable for any attribute-based splitting.

Also fix the error reporting path to properly consume the llvm::Error (use `toString(std::move(E))`) instead of streaming it and letting the unchecked Error destructor abort.

Co-Authored-By: Claude


YuriPlyakhin:
@maksimsab , pls, review

sarnex (approved):

PR #197340, opened 2026-05-13
Adds GetOptions to CommandObjectBreakpointOverride{List, Delete}.

jimingham:
This was causing flaky behavior in TestOverridesResolver.py.

I found this with ASAN, which also didn't like that I was running find with begin() & end() of an empty std::vector.  That surprised me a bit, but its easy to avoid, so I also changed that here.


jimingham:
This is obvious, and should stop flakey failures on the bots, so I'm going to merge this directly.

PR #197341, opened 2026-05-13
Improve the executable name detection in ELF core files.

clayborg:
A previous commit switched us to use the value of the AT_EXECFN, which is an entry in the aux vector, as the executable path. As it turns out, if a symlink is used to launch a program, the symlink path will be in the AT_EXECFN string in core file memory. The PRPSINFO also contains a basename of the program, and it will also be the symlink basename. The best source of information to figure out the executable name is from the NT_FILE note. This always has the resolved path to the executable.

Now the executable name is found in a reliable way starting with finding the NT_FILE entry for the main executable. This can reliably be done by finding the NT_FILE entry whose address contains the AT_PHDR aux vector value. This value is the address of the program headers for the main executable. If there is no NT_FILE entry we can find, we fall back to the AT_EXECFN entry from memory and then fallback to the basename in the PRPSINFO. This patch also creates a placeholder as the main executable when the executable can't be found to ensure users can see which executable they will need to track down in order to load the core file.

The tests added will test the order of precedence. It does this by creating a core file with:
- NT_FILE entry with a path of "/path/nt_file_foo"
- AT_EXECFN in the aux vector with a path of "/path/execfn_foo"
- NT_PRPSINFO entry with a path of "prpsinfo_foo"

We then test that the correct entry is found as the best path option is removed from the core file.


lldb/source/Plugins/Process/elf-core/ProcessElfCore.cpp
```diff
@@ -1150,3 +1188,41 @@ bool ProcessElfCore::GetProcessInfo(ProcessInstanceInfo &info) {
...
+  AuxVector aux_vector(m_auxv);
+  if (auto phdr_addr = aux_vector.GetAuxValue(AuxVector::AUXV_AT_PHDR)) {
+    for (const NT_FILE_Entry &file_entry : m_nt_file_entries) {
+      if (file_entry.start <= *phdr_addr && *phdr_addr < file_entry.end)
+        return file_entry;
+    }
+  }
+  // Fall back to trying to find the first NT_FILE entry that contains the entry
+  // point address.
+  if (auto phdr_addr = aux_vector.GetAuxValue(AuxVector::AUXV_AT_ENTRY)) {
```
jeffreytan81: The logic is the same. Maybe create a helper function to share?

lldb/source/Plugins/Process/elf-core/ProcessElfCore.cpp
```diff
@@ -256,42 +257,42 @@ Status ProcessElfCore::DoLoadCore() {
...
+    ModuleSpec exe_module_spec;
+    if (GetMainExecutableModuleSpec(exe_module_spec)) {
+      exe_module_sp =
+          GetTarget().GetOrCreateModule(exe_module_spec, true /* notify */);
+      if (!exe_module_sp) {
+        // Create an ELF file from memory for the main executable. The dynamic
+        // loader requires the main executable so that it can extract the
+        // DT_DEBUG key/value pair from the dynamic section and get the list
+        // of shared libraries.
+        auto exe_header = GetNTFileEntryForExecutableELFHeader();
```
jeffreytan81: Not a big fan of `auto` for readability, we can explicitly use `std::optional<ProcessElfCore::NT_FILE_Entry>` here?

lldb/source/Plugins/Process/elf-core/ProcessElfCore.cpp
```diff
@@ -256,42 +257,42 @@ Status ProcessElfCore::DoLoadCore() {
...
-                    ReadModuleFromMemory(exe_module_spec.GetFileSpec(),
-                                         *exe_header_addr))
-              exe_module_sp = *module_sp_or_err;
-            else
-              llvm::consumeError(module_sp_or_err.takeError());
-          }
+          lldb::addr_t load_addr =
+              exe_header ? exe_header->start : LLDB_INVALID_ADDRESS;
+          lldb::addr_t size =
+              exe_header ? (exe_header->end - exe_header->start) : 0;
```
jeffreytan81, line 286: For the failure cases, can we add a counter in statistics dump to track it? That would enable us to have telemetry/metrics reporting to track how many cases we still fail to find main executable.

lldb/source/Plugins/Process/elf-core/ProcessElfCore.cpp
```diff
@@ -1150,3 +1188,41 @@ bool ProcessElfCore::GetProcessInfo(ProcessInstanceInfo &info) {
...
+  /// the NT_FILE entry that contains this address.
+  ///
+  /// Previously we would try to find the first NT_FILE entry that had a path
+  /// that ended with the executable name found in the NT_PRPSINFO note, but
+  /// this basename can be the name of a symlink and not the actual resolved
+  /// executable file found in the NT_FILE entry so this could fail for cases
+  /// where a symlink was used to launch the program, and that symlink's
+  /// base name was different from the resolved executable file's name in
+  /// the NT_FILE entry.
+  if (m_nt_file_entries.empty())
```
jeffreytan81, line 1219: Similar, do we want to track this in `statistics dump` to know how many cores do not have NT_FILE?

lldb/source/Plugins/Process/elf-core/ProcessElfCore.cpp
jeffreytan81: Similar, do we want to track this fallback in `statistics dump` to know how many cores fail to locate `AUXV_AT_PHDR` in main executable memory in core?

jeffreytan81 (changes requested):
This is great!
Some comments most around telemetry tracking.

lldb/source/Plugins/Process/elf-core/ProcessElfCore.cpp
clayborg: We can put the object file plug-in name in each module dictionary. Would that work? Something like:
```
"objectFilePlugin": "elf"
```
or:
```
"objectFilePlugin": "placeholder"
```
Then we could count those up?

lldb/source/Plugins/Process/elf-core/ProcessElfCore.cpp
clayborg: That would be nice, but not part of this PR. We can do a separate one for statistics additions.

lldb/source/Plugins/Process/elf-core/ProcessElfCore.cpp
clayborg: That would be nice, but not part of this PR. We can do a separate one for statistics additions.

lldb/source/Plugins/Process/elf-core/ProcessElfCore.cpp
clayborg: That would be nice, but not part of this PR. We can do a separate one for statistics additions.

jeffreytan81 (approved):
LGTM. We will add telemetry tracking in future PRs.

slydiman:
Buildbots [lldb-remote-linux-win](https://lab.llvm.org/buildbot/#/builders/197/builds/15972) and [lldb-x86_64-win](https://lab.llvm.org/buildbot/#/builders/211/builds/8143) are broken.
Please take a look. 
It seems you must skip new tests on Windows host. 
Note lldb-remote-linux-win means Windows x86_64 host and Linux AArch64 target.

ceseo:
Same for AArch64: https://lab.llvm.org/buildbot/#/builders/141/builds/18555

slydiman:
#197887 must fix broken tests on buildbots lldb-remote-linux-win, lldb-x86_64-win and lldb-aarch64-windows.

DavidSpickett:
> Same for AArch64: https://lab.llvm.org/buildbot/#/builders/141/builds/18555

This is actually a pre-existing test:
```
FAIL: test_object_map (TestLinuxCore.LinuxCoreTestCase.test_object_map)
```
And I couldn't reproduce it on my development machine. The latest results don't show it so either it got fixed (I don't see sign of that) or it's a flake. If it is I hope it's something to do with temp paths rather than a thing inside lldb. FWIW, the test doesn't generate a random folder name, so it's not that colliding with expected output.

ceseo:
@DavidSpickett thanks for checking too. I'll keep an eye on it. It might be a flake indeed.

slydiman:
#197887 fixed new tests on Windows.
But I see the test `test_object_map` is broken too. 
https://lab.llvm.org/buildbot/#/builders/211/builds/8174
```
FAIL: test_object_map (TestLinuxCore.LinuxCoreTestCase.test_object_map)
    Test that lldb can find the exe for an i386 linux core file using the object map.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\buildbot\as-builder-10\lldb-x86-64\llvm-project\lldb\test\API\functionalities\postmortem\elf-core\TestLinuxCore.py", line 345, in test_object_map
    self.assertTrue(exe_module_spec.fullpath.startswith(tmp_object_map_root))
AssertionError: False is not true
Config=x86_64-C:\buildbot\as-builder-10\lldb-x86-64\build\bin\clang.exe
----------------------------------------------------------------------
Ran 34 tests in 2.081s
FAILED (failures=1, skipped=6)
```
This test is unchanged. It seems some changes in ProcessElfCore.cpp broke this test on Windows x86_64.
Please investigate and fix it ASAP or revert for now.

slydiman:
TestLinuxCore.py", line 345:
```
        # Copy linux-i386.out to lldb_i386_object_map/a.out
        tmp_object_map_root = os.path.join(self.getBuildDir(), "lldb_i386_object_map")
        executable = os.path.join(tmp_object_map_root, "a.out")
        lldbutil.mkdir_p(os.path.dirname(executable))
        shutil.copyfile("linux-i386.out", executable)

        # Replace the original module path at /home/labath/test and load the core
        self.runCmd("settings set target.object-map /home/labath/test {}".format(tmp_object_map_root))

        target = self.dbg.CreateTarget(None)
        process = target.LoadCore("linux-i386.core")

        # Check that we did load the mapped executable
        exe_module_spec = process.GetTarget().GetModuleAtIndex(0).GetFileSpec()
        self.assertTrue(exe_module_spec.fullpath.startswith(tmp_object_map_root))
```
The values on Windows are:
exe_module_spec = /home/labath/test/a.out
tmp_object_map_root = D:\build-dir\lldb-test-build.noindex\functionalities\postmortem\elf-core\TestLinuxCore\lldb_i386_object_map

Something is totally wrong here.
I'm going to revert this patch on Monday to unblock CI.

clayborg:
> linux-i386.core

Can we just disable this test on windows if this is the only CI failing with this issue?

clayborg:
> TestLinuxCore.py", line 345:
> 
> ```
>         # Copy linux-i386.out to lldb_i386_object_map/a.out
>         tmp_object_map_root = os.path.join(self.getBuildDir(), "lldb_i386_object_map")
>         executable = os.path.join(tmp_object_map_root, "a.out")
>         lldbutil.mkdir_p(os.path.dirname(executable))
>         shutil.copyfile("linux-i386.out", executable)
> 
>         # Replace the original module path at /home/labath/test and load the core
>         self.runCmd("settings set target.object-map /home/labath/test {}".format(tmp_object_map_root))
> 
>         target = self.dbg.CreateTarget(None)
>         process = target.LoadCore("linux-i386.core")
> 
>         # Check that we did load the mapped executable
>         exe_module_spec = process.GetTarget().GetModuleAtIndex(0).GetFileSpec()
>         self.assertTrue(exe_module_spec.fullpath.startswith(tmp_object_map_root))
> ```
> 
> The values on Windows are: exe_module_spec = /home/labath/test/a.out tmp_object_map_root = D:\build-dir\lldb-test-build.noindex\functionalities\postmortem\elf-core\TestLinuxCore\lldb_i386_object_map
> 
> Something is totally wrong here. I'm going to revert this patch on Monday to unblock CI.

I think what is happening here is that one path is a windows path and one path is in unix format. I did change the path format and allow auto detection where previously it would always default to the native file system. So that might stop the remapping from happening correctly on windows.

clayborg:
We might be able to see if this change fixes windows:
```
static FileSpec CreateFileSpecFromPath(llvm::StringRef path) {
  FileSpec::Style path_style = FileSpec::Style::native;
  //if (auto guessed_style = FileSpec::GuessPathStyle(path))
  //  path_style = *guessed_style;
  return FileSpec(path, path_style);
}
```
Just commenting out the `FileSpec::GuessPathStyle` call to revert to previous behavior

slydiman:
> We might be able to see if this change fixes windows:

After this change the `test_object_map` is passed, but failed 2 other tests:
```
FAIL: test_exe_name_extraction_at_execfn (TestLinuxCore.LinuxCoreTestCase.test_exe_name_extraction_at_execfn)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\as\mainline\llvm-project\lldb\test\API\functionalities\postmortem\elf-core\TestLinuxCore.py", line 1318, in test_exe_name_extraction_at_execfn
    self.assertEqual(exe_module.GetFileSpec().fullpath, "/path/execfn_foo")
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: '\\path\\execfn_foo' != '/path/execfn_foo'
- \path\execfn_foo
? ^    ^
+ /path/execfn_foo
? ^    ^

======================================================================
FAIL: test_exe_name_extraction_nt_file (TestLinuxCore.LinuxCoreTestCase.test_exe_name_extraction_nt_file)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\as\mainline\llvm-project\lldb\test\API\functionalities\postmortem\elf-core\TestLinuxCore.py", line 1301, in test_exe_name_extraction_nt_file
    self.assertEqual(exe_module.GetFileSpec().fullpath, "/path/nt_file_foo")
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: '\\path\\nt_file_foo' != '/path/nt_file_foo'
- \path\nt_file_foo
? ^    ^
+ /path/nt_file_foo
? ^    ^
```

> Can we just disable this test on windows if this is the only CI failing with this issue?

Probably yes and create an issue for now.
@DavidSpickett what do you think?

DavidSpickett:
Fine with me.

slydiman:
#198473

PR #197343, opened 2026-05-13
[DirectX] Fix layering violation in b239b5c

boomanaiden154:
b239b5c07145df1fc3503bcde58481e21ba265a1 introduced a layering violation. It included a header from llvm/Object from a header in llvm/MC, which is a layering violation as llvm/Object depends on llvm/MC. Given the only thing needed here is a single type, just hoist it to a header in BinaryFormat which both MC and Object are allowed to depend on.


nigham (approved):

dzhidzhoev:
Thank you for an attempt to fix that!
Since more than one issue was reported for this commit, I've reverted it for now.

PR #197344, opened 2026-05-13
[NFC][clang][test] Move tests which require clang_staticanalyzer to clang/test/Tooling/StaticAnalyzer

YuriPlyakhin:
Not all tests in Tooling folder require `CLANG_ENABLE_STATIC_ANALYZER=ON`
Move all tests that depend on it into a dedicated subdirectory with its own lit.local.cfg gate, so that tests in Tooling/ that don't depend on it can run in builds that disable the static analyzer.

Co-Authored-By: Claude


YuriPlyakhin:
@sarnex , @bader , I'm not sure who is the best person to review this change. Please, review. If you know there is someone who is better equipped to review this change, feel free to add. Thanks!

NagyDonat:
Do these tests share some property with the other tests in `Tooling` which justifies keeping them in `Tooling` (under a subdirectory)?

If there is no such justification, I would suggest moving them to `clang/test/Analysis` instead, because that is the default place where we have most tests of the static analyzer [1]. (In fact, this is the first time I hear about LIT tests for the analyzer outside of that directory -- and I started working on the analyzer as a full-time job 3 years ago.)

--------

[1] Historical background:
At the beginning, the static analyzer was a part of the `Analysis` subproject, so its files were placed in `clang/lib/Analysis`, `clang/include/clang/Analysis`, `clang/test/Analysis` etc. When it grew too large, it was split off from the Analysis library (which is also used by e.g. the so-called "analysis-based compiler warnings") and is source files were moved to `clang/lib/StaticAnalyzer`, its include files were moved to `clang/include/clang/Analysis` -- but the tests were left in `clang/test/Analysis` and in fact they were practically the only occupants of that test directory (because the tests of the analysis-based warnings were placed in other test directories).

A few month ago I moved the last few non-static-analyzer-related tests from `clang/test/Analysis` to other directories; in the foreseeable future I'm planning to rename that directory to `clang/test/StaticAnalyzer` (to match the name of the subdirectories under `lib` and `include`).

bader:
I'm not an expert in this area, but these tests seem to be testing clang-check's integration with the analyzer as a Tooling tool. Tests exercise the `clang-check -analyze` command-line interface and its argument handling, which is Tooling infrastructure. That distinguishes them from clang/test/Analysis tests which test the analyzer's diagnostic output and checkers directly i.e. static analysis logic.

This makes me think that this directory is probably not the right place for offload tool tests as these tools do not use https://clang.llvm.org/docs/LibTooling.html infrastructure.

@jhuber6, @sarnex, what do you think about having a test directory for offload tools like (i.e. clang-linker-wrapper, clang-nvlink-wrapper, clang-sycl-linker, clang-offload-bundler, etc.)?
Today most of offload tool tests are located in `clang/test/Driver` directory + a few tests in `clang/test/Tooling` (including LLVM IR tests unsupported in Driver directory).

sarnex:
I'm definitely in favor of moving them out of `clang/test/Driver`, The source code for these tools exists in `clang/tools`, so I would propose we add a new `tools` directory in `clang/test`, so we could have like `clang/test/tools/clang-linker-wrapper/` and so on.

YuriPlyakhin:
> I'm definitely in favor of moving them out of `clang/test/Driver`, The source code for these tools exists in `clang/tools`, so I would propose we add a new `tools` directory in `clang/test`, so we could have like `clang/test/tools/clang-linker-wrapper/` and so on.

that sounds reasonable for me. I'll create new PR.

YuriPlyakhin:
https://github.com/llvm/llvm-project/pull/198434

PR #197345, opened 2026-05-13
[RISCV][P-ext] Add splat_vector for 64-bit vectors on RV32.

topperc:


llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -1904,7 +1868,7 @@ define <2 x i32> @test_pasubu_w(<2 x i32> %a, <2 x i32> %b) {
 define <2 x i32> @test_non_const_splat_i32(i32 %elt) {
 ; RV32-LABEL: test_non_const_splat_i32:
 ; RV32:       # %bb.0:
-; RV32-NEXT:    mv a1, a0
+; RV32-NEXT:    pmv.dws a0, a0
```
lenary: I'm not sure this is better? the `mv` should compress, I don't know if this will.

llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -1432,7 +1403,7 @@ define <2 x i32> @test_plui_w() {
 ; RV32-LABEL: test_plui_w:
 ; RV32:       # %bb.0:
 ; RV32-NEXT:    lui a0, 76800
-; RV32-NEXT:    lui a1, 76800
+; RV32-NEXT:    pmv.dws a0, a0
```
lenary: Here we've made the critical paths longer, and there's no chance `pmv.dws` will compress. I'm not entirely sure how to detect this and avoid it.

Maybe we should just get rid of the v2i32 splat cases on rv32, so it falls back to the register move?

lenary (commented):
Looks good, apart from `pmv.dws`, which can just be a `mv` in basically all cases, (or a `li`)

llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -1902,21 +1874,8 @@ define <2 x i32> @test_non_const_splat_i32(i32 %elt) {
...
-; RV32-NEXT:    sb a2, 14(sp)
-; RV32-NEXT:    sb a2, 15(sp)
-; RV32-NEXT:    sb a2, 8(sp)
-; RV32-NEXT:    sb a2, 9(sp)
-; RV32-NEXT:    sb a2, 10(sp)
-; RV32-NEXT:    sb a2, 11(sp)
-; RV32-NEXT:    lw a3, 12(sp)
-; RV32-NEXT:    lw a2, 8(sp)
+; RV32-NEXT:    pmv.dbs a2, a2
 ; RV32-NEXT:    padd.db a0, a2, a0
```
topperc, line 1878: I will fold this to padd.dbs in a follow up

wangpc-pp (approved):
LGTM.

lenary (approved):
LGTM

PR #161114, opened 2025-09-29
[aarch64] Mix the frame pointer with the stack cookie when protecting the stack

PanTao2:
This strengthens the guard and matches MSVC.

Fixes #156573 .

PR #197346, opened 2026-05-13
[aarch64] Mix the frame pointer with the stack cookie when protecting the stack

PanTao2:
For MSVC-compatible targets on AArch64, mix the stack cookie with the
frame pointer (FP) to create a position-dependent guard value. This
strengthens protection against attacks where the attacker knows or can
predict the cookie value, as they would also need to know the exact
frame pointer location.

Implementation details:
- Uses SUB (FP - Cookie) instead of XOR like X86 because:
  * SUB maintains the existing AArch64 instruction selection patterns
  * SUB provides equivalent security properties (bijective mixing)
  * The result is still unpredictable without knowing both inputs
- The same SUB operation is performed in both prologue (to store the
  mixed value) and epilogue (to unmix and verify the cookie)
- Forces frame pointer usage for functions with stack guards on MSVCRT
  to ensure consistent addressing with dynamic stack allocation

This matches the MSVC behavior and strengthens stack protection on
AArch64 Windows platforms.

Fixes #156573.


PanTao2:
I’m very sorry for the error (0xc0000409 is STATUS_STACK_BUFFER_OVERRUN (Security check failure or stack buffer overrun)) introduced by the previous CL (https://github.com/llvm/llvm-project/pull/161114). Thanks @luporl  for reverting it.
The error was caused by me using “add” instruction (SP + (SP - cookie) = 2 * SP - cookie, wrong) not “sub” instruction (SP - (SP - cookie) = cookie, correct) in the failure BB. I have fixed the errors in this CL. @efriedma-quic Could you please review this CL? Thank you in advance.
@zmodem I’m very sorry for blocking your work, could you please try this CL before re-landing it?

llvm/test/CodeGen/AArch64/stack-protector-target.ll
```diff
@@ -31,14 +31,18 @@ declare void @_Z7CapturePi(ptr)
 
 ; WINDOWS-AARCH64: adrp x8, __security_cookie
 ; WINDOWS-AARCH64: ldr x8, [x8, :lo12:__security_cookie]
+; WINDOWS-AARCH64: sub x8, sp, x8
 ; WINDOWS-AARCH64: str x8, [sp, #8]
 ; WINDOWS-AARCH64: bl  _Z7CapturePi
-; WINDOWS-AARCH64: ldr x0, [sp, #8]
+; WINDOWS-AARCH64: ldr x9, [sp, #8]
+; WINDOWS-AARCH64: sub x8, sp, x8
```
zmodem: I don't really know how this works, but at a glance it looks odd.
It seems the result of the load in `x9` is not used?
Is `x8` still live from before when doing `sub x8, sp, x8`, or what value are we expecting it to have? Shouldn't we be working with the value on the stack rather than the original stack cookie?
And doesn't `__security_check_cookie` expect a parameter in `x0`?

zmodem (commented):
> I’m very sorry for blocking your work, could you please try this CL before re-landing it?

The test from last time fails with this patch also, though that is a pretty opaque test so it's hard to see what's going wrong.

Last time there were also failures on some LLVM buildbots (https://lab.llvm.org/buildbot/#/builders/161/builds/9433 and https://lab.llvm.org/buildbot/#/builders/124/builds/1750). Can you check whether those tests pass with your patch this time?

llvm/test/CodeGen/AArch64/stack-protector-target.ll
PanTao2: Thank you for pointing out the confusing check items.
These two lines belong to %bb.0, before the cmp instruction.  I have changed these two lines to check all instructions before '__securace_check_comkie' in failure BB. Could you please help check if the new check items meet your expectations?
I cannot reproduce https://lab.llvm.org/buildbot/#/builders/161/builds/9433 on my arm64 Windows device. @luporl Do you know any way to run the relevant LLVM buildbot before merging this PR? Thank you in advance.
I think the 3 failing checks are unrelated to my changes.

luporl: @PanTao2 This might be possible with the "Force Build" button, in the buildbot page (e.g., https://lab.llvm.org/buildbot/#/builders/161). It would need a branch in LLVM repo.

If that doesn't work, I can pause the bot and manually test the PR on it.

PanTao2: Thank you very much for your help.
I tried to "Force Build" in https://lab.llvm.org/buildbot/#/builders/161, my configuration is
```
Reason: Force clean build
Priority: 0
Branch: Pantao2:mychange
Revision: b7cf9dc05c9bf1de3d71e20e52207a7009911b17
Clean source code and build directory
Clean build directory
```
After clicking “Start Build” button, report error “You must be an author of the commit triggered the build.”
Do you know why I (author of the commit) cannot start the build?
Could you please help to test this PR on it?

luporl: I was able to request a build with `Reason: Build a particular revision` : https://lab.llvm.org/buildbot/#/buildrequests/8300285?redirect_to_build=true.
Let's see how this goes.

luporl: The build results are on https://lab.llvm.org/buildbot/#/builders/161/builds/11548.

There are a couple tests failing with exit status 0xc0000409.

PanTao2: Thank you very much for your help. I reproduced the error triggering 0xc0000409 (STATUS_STACK_BUFFER_OVERRUN) on my arm64 Windows machine and fixed it. Could you please help deploy a new test for the new version '1b7040f3f4adc99fe43dfcfb129ecd1142f2Fe6b'? Thanks in advance.

llvm/test/CodeGen/AArch64/GlobalISel/irtranslator-stack-protector-windows.ll
```diff
@@ -15,12 +15,15 @@ define void @caller() sspreq {
...
 ; CHECK-NEXT:    ldr x8, [x8, :lo12:__security_cookie]
+; CHECK-NEXT:    sub x8, sp, x8
 ; CHECK-NEXT:    str x8, [sp, #8]
 ; CHECK-NEXT:    bl callee
-; CHECK-NEXT:    adrp x8, __security_cookie
 ; CHECK-NEXT:    ldr x9, [sp, #8]
-; CHECK-NEXT:    ldr x8, [x8, :lo12:__security_cookie]
-; CHECK-NEXT:    cmp x8, x9
+; CHECK-NEXT:    adrp x10, __security_cookie
+; CHECK-NEXT:    mov x8, sp
```
efriedma-quic: It looks like you somehow broke the optimization which turns the mov+sub into `sub x8, sp, x9`?

llvm/test/CodeGen/AArch64/stack-protector-target.ll
luporl: I've enqueued a build for the new version, but it may take a while to finish. Builds have been slow since the last LLVM buildbot maintenance.

luporl: Bots are back to normal.
This is the build of 550ddea22d87b14e8b94237444531ffe70e2fe4a: https://lab.llvm.org/buildbot/#/builders/161/builds/11592

llvm/test/CodeGen/AArch64/GlobalISel/irtranslator-stack-protector-windows.ll
PanTao2: Thank you for pointing out this issue. Avoiding the move from sp instruction in the epilogue introduces some complexity. Could you please further review?

llvm/test/CodeGen/AArch64/stack-protector-target.ll
PanTao2: Thank you very much for your help.

llvm/test/CodeGen/AArch64/GlobalISel/irtranslator-stack-protector-windows.ll
efriedma-quic: It also seems to be introducing other side-effects... why is this having much wider impact than the original change in #161114?

PanTao2: Thank you for your comment. I removed redundant modifications of the test. The impact returns to be similar to #161114.  Could you please further review?
The test failure of tools/llvm-symbolizer/wasm-basic.s on Linux AArch64 and Windows seems unrelated to this PR.

PanTao2: I found a new problem, on the aarch64 platform, _alloca may move the stack pointer, and mixing cookie with the stack pointer may introduce incorrect mismatch. Therefore, I have switched to using frame pointer (FP/x29) to mix cookies, which does not change throughout the function.

zmodem: I was going to ask whether a frame pointer is always available, but I see you're ensuring that in this scenario.

But don't all targets have the issue that allocas can move the stack pointer? How do they handle this problem? Can't we just make sure to use the stack pointer before it's modified?

efriedma-quic: x86 XORs the frame pointer.  Other targets don't do anything like this.

llvm/lib/Target/AArch64/AArch64InstrInfo.cpp
```diff
@@ -2456,14 +2456,30 @@ bool AArch64InstrInfo::removeCmpToZeroOrOne(
...
   auto TRI = Subtarget.getRegisterInfo();
   DebugLoc DL = MI.getDebugLoc();
 
+  if (MI.getOpcode() == AArch64::STACK_GUARD_UNMIX) {
+    // Expand STACK_GUARD_UNMIX to: sub Rd, fp, Rs
+    // This computes FP - stored_mixed_value to unmix the cookie
+    Register DstReg = MI.getOperand(0).getReg();
+    Register SrcReg = MI.getOperand(1).getReg();
+
+    BuildMI(MBB, MI, DL, get(AArch64::SUBXrx64), DstReg)
```
efriedma-quic: If we're using FP, we should switch to the canonical SUB opcode, so we don't print the "uxtx".  (Not that it really matters, probably, but it's at least easier to read.)

llvm/test/CodeGen/AArch64/GlobalISel/irtranslator-stack-protector-windows.ll
PanTao2: @efriedma-quic Thank you very much for your explanation and clarification.

llvm/lib/Target/AArch64/AArch64InstrInfo.cpp
PanTao2: Thank you very much, it is much more readable. Could you please further review?

llvm/test/CodeGen/AArch64/mingw-refptr.ll
```diff
@@ -82,25 +82,30 @@ define dso_local void @sspFunc() #0 {
...
 ; CHECK-NEXT:    ldr x8, [x8]
+; CHECK-NEXT:    sub x8, x29, x8
 ; CHECK-NEXT:    str x8, [sp, #8]
 ; CHECK-NEXT:    bl ptrUser
 ; CHECK-NEXT:    adrp x8, .refptr.__stack_chk_guard
 ; CHECK-NEXT:    ldr x8, [x8, :lo12:.refptr.__stack_chk_guard]
 ; CHECK-NEXT:    ldr x9, [sp, #8]
 ; CHECK-NEXT:    ldr x8, [x8]
+; CHECK-SD-NEXT: sub x9, x29, x9
+; CHECK-GI-NEXT: sub x8, x29, x8
```
efriedma-quic: Please don't hand-write checks in files using generated checks; use update_llc_test_checks.py.

PanTao2: I used update_llc_test_checks.py to update this file. Could you please further review?

PanTao2:
@efriedma-quic @zmodem @luporl Thank you very much for your review and help, could you please further review?

luporl:
@PanTao2 Regarding the WoA buildbots, everything seems to be fine now, since 84ef93420bf1617dcc1e79baa23bf74bbdbf20f2 passes all the tests of [clang-arm64-windows-msvc](https://lab.llvm.org/buildbot/#/builders/161).

PanTao2:
@luporl Thank you very much for your help. @efriedma-quic @zmodem Could you please further review?

PanTao2:
@efriedma-quic @zmodem Ping, could you please further review?

PanTao2:
@efriedma-quic @zmodem Ping, could you please further review?

llvm/lib/Target/AArch64/AArch64FrameLowering.cpp
```diff
@@ -570,6 +570,16 @@ bool AArch64FrameLowering::hasFPImpl(const MachineFunction &MF) const {
   // funclets.
   if (MF.hasEHFunclets())
     return true;
+
+  // Force FP for functions with stack guards on MSVCRT.
+  // The cookie mixing uses FP to ensure consistent addressing even with
+  // dynamic stack allocation (e.g., _alloca).
+  if (MFI.hasStackProtectorIndex()) {
+    const auto &Subtarget = MF.getSubtarget<AArch64Subtarget>();
+    if (Subtarget.getTargetTriple().isOSMSVCRT())
```
efriedma-quic: Can we call useStackGuardMixFP() here?  (I don't want this to accidentally fall out of sync.)

PanTao2: Yes. I replaced it with useStackGuardMixFP(). Could you please further review?

PanTao2:
@efriedma-quic @zmodem Ping, could you please further review?

efriedma-quic (approved):
LGTM

PR #197347, opened 2026-05-13
Add benchmark scripts for lit performance measurement

prasoon054:
**Description:**
Adds benchmark scripts under `llvm/utils/lit/utils/` for measuring lit's wall-clock performance on macOS, Ubuntu x86_64, and Windows x64.

**Motivation:**
Ongoing work to improve lit's performance requires a repeatable before/after measurement workflow across platforms. These scripts provide that.

**What the scripts do?**
1. Check prerequisites and report install commands for any missing tools.
2. Configure an LLVM build if one does not already exist.
3. Build all required LLVM tools via ninja.
4. Run the specified lit test directory under `hyperfine` with the specified worker count.
5. Export a markdown timing table to a timestamped results directory.

**Platform-specific behaviour:**
- macOS: wraps hyperfine in `caffeinate` to prevent sleep-mode variance.
- Linux: disables turbo boost and sets the CPU governor to `performance` before benchmarking; restores on exit. (not added yet)
- Windows: WIP


llvm/utils/lit/utils/BenchmarkLinux.sh
ilovepi: This seems very specific to intel. Can you not control this w/ `cpupower`? That's how I've typically seen it done these days. At the very least, there used to be some non-arch specific methods on linux to disable cpu scaling altogether and pin frequencies. @boomanaiden154 are there any more recent guides than https://llvm.org/docs/Benchmarking.html? the info on that page is more or less what I recall doing a long time ago, but some of it is perhaps a bit out-of-date w.r.t. best practices these days.

ilovepi (commented):
I think we'd generally prefer to use python over bash.  This script is also taking several parameters, so I think it would be nice to have proper argument parsing, instead of making them all just be in a specific order.

There's also quite a bit of overlap between the mac and linux version. A set of classes may make this a bit easier to manage in a single script.

For example, you have a few distinct phases 1) your config and variable set ups, 2) command checking, 3) environmental setup (disable cpu scaling, etc.), 4) benchmarking command, 5) return results. 

Theres lots of ways you could do this (base class + specializations for OS specifics vs. duck typing vx ....), but I think  the way this is written right now is quite repetitive.

prasoon054:
1. Sure. I will move it to python. We could also use an env/config file for the settings instead of passing everything as arguments.
2. I definitely violated the DRY principle here. Thanks for mentioning "duck typing". *[I had heard the term before but never really looked into it. At first glance, it feels somewhat similar to templates in C++.]* However, in this case, I think I will stick with the other option (base class + specialization).
3. Afaik using only the performance governor can still allow turbo spikes, which may lead to less consistent results. The turbo mode ("precision boost" in AMD's terminology) on newer linux systems is exposed under `/sys/devices/system/cpu/cpufreq/`, while my Intel system exposes it under `/sys/devices/system/cpu/intel_pstate`. In our script, we can check which interface is present and make changes accordingly. [[Ref](https://www.kernel.org/doc/html/latest/admin-guide/pm/cpufreq.html#the-boost-file-in-sysfs)]

On a side note, I think that using `perf` for profiling would be better than using `hyperfine`, once we have a configuration which has a stable stdev on the wall clock metric across platforms.

llvm/utils/lit/utils/BenchmarkLinux.sh
boomanaiden154: I'm not aware of any more up to date guides. There might be some stuff that exists, but it would be external to LLVM.

llvm/utils/lit/utils/BenchmarkMacOS.sh
boomanaiden154: Doing `compiler-rt` in the projects build is deprecated. For building `compiler-rt`, you should be doing `LLVM_ENABLE_RUNTIMES="compiler-rt"`. It makes the lit setup quite a bit more annoying, but it is what it is.

llvm/utils/lit/utils/benchmark.py
ilovepi: Seems like you're assuming a debian based system. That's fine, but you should document it and nay other limitations/assumptions you're making.

ilovepi: Ultimate?

ilovepi: These comments indicate to me something that likely would be good to encapsulate in a function or method, since I'd expect it to be used across the platforms.

ilovepi: so you only build runtimes?

ilovepi: Isn't this on by default?

ilovepi: you can use format strings to set this more programatically.

```python

cc = "clang" if linux else "clang-cl"
cmd.extend([f"-DCMAKE_C_COMPILER={cc}",...])
```

ilovepi: why `pass`? seems like you may want to know this failed somehow, or abort...

ilovepi: This is a little sketchy. just insist that any paths you need are always provided.

ilovepi: ditto. Just make it an error not to pass that to the script.

ilovepi: If you're going to process this later and do analysis across runs, you may want to have this as json. not critical to start, but worth metnioning.

ilovepi: can you organize this script in a way that allows you to track which system you're in more consistently. you do this type of check quite a lot. a grep for "SYSTEM==" gets 14 hits. managing this kind of state consistently should simplify some of the implementation. certain parts can even be handled just after arg parsing (e.g. you can select the platform there).

ilovepi: This seems a bit odd for a command wrapper. why does our current process need a new priority class? I assume you mean to set this for whatever command is running, and AFAIK this doesn't affect the global CPU state the way the changes you made in setup do.

ilovepi: again, I think this sets our current process's execution, and not the subprocess.

ilovepi: There's probably a better way to get and set the cpu numbers

ilovepi: what is this doing? I think this needs to be less clever and just use a simple helper function.

llvm/utils/lit/utils/BenchmarkLinux.sh
ilovepi: Do you still need the bash scripts at all at this point?

ilovepi (commented):
I think this has a long way to go still. Its fine its not perfect, but I'd encourage you to think a bit more about the architecture and the abstractions you want. There's quite a lot that seems to be repeated, and in many cases you're checking the same values over and over.

I'd suggest going through this and making some cleanups, with an eye towards simplifying the implementation. It's also fine to insist that the user provide whatever information (e.g. paths) that you need, and I think doing so will make at least a few things easier.

ilovepi:
also, you  need to run the formatter.

prasoon054:
Thanks for reviewing!
1. I've updated the code with a new design that is less redundant IMO. Will push the changes here after running a few tests across platforms.
2. Earlier, I couldn't find the python formatter in the LLVM guidelines. Now that I know it is `darker`, I will make sure to run the formatter before opening / updating a PR.

llvm/utils/lit/utils/benchmark.py
prasoon054: For `codegen-x86`, `transforms`, `llvm-mca` `clang` is not needed. All required tools are covered by the `llvm-test-depends` target. For `compiler-rt-asan`, 9 tests have `REQUIRES: lld-available` or `REQUIRES: lld` which can be enabled, but imo it won't affect the benchmark's core measurement.

ilovepi: It may make sense to use logging for this over just manually formatting all the print statements.

For something with a fixed output  a raw string makes more sense than to require `\t` everywhere.

However, its not clear that we really want this kind of warning output. I think I'd prefer something succinct that pointed to a document describing this over this kind of message.

ilovepi: should these APIs return some kind of error so you can handle success/failure?

ilovepi: If you need to check these a lot, maybe a`Platform` object that provides APIs for these would be more useful? You only use these in 2 places right now, so instead of SYSTEM, you could just `import system from platform`, and use `system()` over the global variable.

The usage of the `IS_LINUX`, etc is very limited, so you could just as easily use a dictionary or matching statement.

ilovepi: Since these are just used in a single method, I think these can live within that method instead of being class members.

ilovepi: AFAICT you're using the kernel32 bits to set the cpu affinity, but can't you set that w/ the `os` module? I think you should be able to set priority and set afifnity with it, right? That may make it much easier to just have a cross platform API

https://docs.python.org/3/library/os.html#os.sched_setaffinity
https://docs.python.org/3/library/os.html#os.setpriority

ilovepi: why does this need specialization for windows?

ilovepi: feels a bit odd for a check_* api to return None over an error or status.

ilovepi: Ditto here on formatting the strings.

ilovepi: Is this really necessary? this whole bit feels overly specific and kind of disruptive to systems that could have this task fail and not restore services. @boomanaiden154 any thoughts?

ilovepi (commented):
This is better, but I think there may still be a ways to go before this is ready to use.

llvm/utils/lit/utils/benchmark.py
prasoon054: Each of these are used in atleast two different methods. Keeping them intact for now.

prasoon054: `os.sched_setaffinity` and `os.setpriority` are unix-only. They raise `AttributeError` here. There's no stdlib cross-platform affinity API, so Windows needs the `kernel32` calls.
`psutil` would give a cross-platform API, but it still wouldn't unify the code. Linux keeps `cset shield`. So we'd still branch per platform.

prasoon054: On windows there's no shebang mechanism and the file is `bin/llvm-lit.py`

ilovepi: is it a bug that GOV_PATH is only used in one place then?

ilovepi: Ack. Wish pythton's docs for the os module had made that more obvious.

ilovepi: please use a more descriptive name than `r`.

ilovepi: so is the assumption that if there's at least 4 lines this is your schema? what happens if an error output is say 10 lines?

ilovepi: again,  more descriptive names would help.

ilovepi:
```suggestion
            self.apply_setting(sub, setting, "100")
```
not clear to me why this isn't in the preceding block...

ilovepi: seems like we have a lot of operations that log an issue, but don't return an error, stop execution or do anything else... so what's the consequence of a call to `powercfg` failing? is it safe to continue? Should the benchmarks stop?

ilovepi: Names again...

ilovepi (commented):
I think this is notably better. I think a more cohesive approach to your exception handling/logging would be helpful.

llvm/utils/lit/utils/benchmark.py
prasoon054: Ohh, right! I should move `GOV_PATH` inside the `set_cpu_performance` method.

prasoon054: Yes, the benchmarks should stop here.

ilovepi: Maybe add a FIXME to use https://docs.python.org/3/library/os.html#os.waitstatus_to_exitcode when we have at least 3.9 available? Also I'm not totally sure using the negative of the WTERMSIG is the right approach on windows.  The page above indicates that they right shift status by 8bits, but that sounds questionable to me.

ilovepi (approved):
Modulo a few nits, I think this is in pretty good shape. As you mentioned in chat, it would probably be a bit nicer to split this up slightly into a few files. I thin that would be a good change, and from what I can see would be pretty mechanical, depending on where you made the split. If possible I'd like to see that in this PR, unless it turns out to require solving other technical issues we didn't account for.

llvm/utils/lit/utils/benchmark.py
ilovepi: Oh, I think this needs an  LLVM licence header. You can check the other nearby files in utils, but IIRC they have a version of this: https://llvm.org/docs/CodingStandards.html#file-headers .

prasoon054: I've added a FIXME to switch to `os.waitstatus_to_exitcode()` once min Python version reaches 3.9.

`_exit_code_from_wait_status` function is reachable via the POSIX `measure_run` path, which is guarded by `hasattr(os, "wait4")`. `WindowsPlatform` overrides `measure_run` entirely with its own win32 path, so `_exit_code_from_wait_status` never runs on Windows.

llvm/utils/lit/utils/benchmark/platforms/base.py
ilovepi, line 1: License?

llvm/utils/lit/utils/benchmark/platforms/__init__.py
ilovepi, line 1: Licence?

llvm/utils/lit/utils/benchmark/util.py
ilovepi, line 103: nit: this feels very much like an LLM comment. That needs to be disclosed if it is (which is fine). But I also think its a bit unclear, and the `;` feels unnatrual here.

I'd recommend a once over for all the comments.

ilovepi (commented):
This is better, and has addressed most issues. I think the commenting and documentation could still use a bit of work, but this is mostly ready to land. I'd also suggest marking this as ready for review (e.g. not a draft).

PR #197352, opened 2026-05-13
Use utxb rN, rM, ror #8 to zext from shifts on armv6

LumioseSil:


llvm/test/CodeGen/ARM/extract-ext-armv6.ll
```diff
@@ -0,0 +1,62 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
+; RUN: llc -mtriple=armv6-none-eabi -O3 %s -o - | FileCheck %s
```
arsenm:
```suggestion
; RUN: llc -mtriple=armv6-none-eabi < %s | FileCheck %s
```

llvm/test/CodeGen/ARM/extract-ext-armv6.ll
```diff
@@ -0,0 +1,62 @@
...
+  %3 = ashr i32 %2, 24
+  ret i32 %3
+}
+
+define i32 @sext_i8_from_bits16(i32 %0) {
+; CHECK-LABEL: sext_i8_from_bits16:
+; CHECK:       @ %bb.0:
+; CHECK-NEXT:    sxtb r0, r0, ror #16
+; CHECK-NEXT:    bx lr
+  %2 = shl i32 %0, 8
```
arsenm, line 39: Use named values in tests

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
```diff
@@ -11350,6 +11350,18 @@ SDValue DAGCombiner::visitSRA(SDNode *N) {
         return DAG.getNode(ISD::SIGN_EXTEND, DL,
                            N->getValueType(0), Trunc);
       }
+
+      if ((ShiftAmt > 0) &&
+          TLI.getOperationAction(ISD::SIGN_EXTEND_INREG, TruncVT) ==
+              TargetLowering::Legal) {
```
arsenm:
```suggestion
      if (ShiftAmt > 0 &&
          TLI.isOperationLegal(ISD::SIGN_EXTEND_INREG, TruncVT)) {
```

AZero13: Does not do the same thing. Checks if type is legal, not just operation.

AZero13:
I know this in theory can be in seldag but because it's only having a benefit in ARM and causing issues elsewhere, I'm probably going to have to move this to arm backend for now...

PR #197353, opened 2026-05-13
 [CloneModule] Clone undefined ifuncs

mmjjpp:
To satisfy the verifier rule "IFunc resolver must be a definition". We fix iFunc handling when cloning modules.
When cloning a module, if an IFunc has no definition (ShouldCloneDefinition returns false), directly create an external GlobalValue (Function or GlobalVariable) instead of trying to clone the ifunc.
Add a test case for llvm-split to verify the ifunc cloning/splitting behavior works correctly.

### DIFF
--- a/llvm/lib/Transforms/Utils/CloneModule.cpp
+++ b/llvm/lib/Transforms/Utils/CloneModule.cpp
@@ -119,7 +119,7 @@ std::unique_ptr<Module> llvm::CloneModule(
     // Defer setting the resolver function until after functions are cloned.
     if (!ShouldCloneDefinition(&I)) {
       // An ifunc also cannot act as an external reference, so we need to create
-      // a function depending on the value type.
+      // a function.
       GlobalValue *GV;
       assert(I.getValueType()->isFunctionTy() &&
              "ValueType of ifunc must be function type!");
--- a/llvm/lib/Transforms/Utils/CloneModule.cpp
+++ b/llvm/lib/Transforms/Utils/CloneModule.cpp
@@ -122,7 +122,7 @@ std::unique_ptr<Module> llvm::CloneModule(
       // a function depending on the value type.
       GlobalValue *GV;
       assert(I.getValueType()->isFunctionTy() &&
-             "IFunc resolver must be function type!");
+             "ValueType of ifunc must be function type!");
       GV = Function::Create(cast<FunctionType>(I.getValueType()),
                             GlobalValue::ExternalLinkage, I.getAddressSpace(),
                             I.getName(), New.get());
--- a/llvm/lib/Transforms/Utils/CloneModule.cpp
+++ b/llvm/lib/Transforms/Utils/CloneModule.cpp
@@ -119,17 +119,13 @@ std::unique_ptr<Module> llvm::CloneModule(
     // Defer setting the resolver function until after functions are cloned.
     if (!ShouldCloneDefinition(&I)) {
       // An ifunc also cannot act as an external reference, so we need to create
-      // either a function or a global variable depending on the value type.
+      // a function depending on the value type.
       GlobalValue *GV;
-      if (I.getValueType()->isFunctionTy())
-        GV = Function::Create(cast<FunctionType>(I.getValueType()),
-                              GlobalValue::ExternalLinkage, I.getAddressSpace(),
-                              I.getName(), New.get());
-      else
-        GV = new GlobalVariable(*New, I.getValueType(), false,
-                                GlobalValue::ExternalLinkage, nullptr,
-                                I.getName(), nullptr, I.getThreadLocalMode(),
-                                I.getType()->getAddressSpace());
+      assert(I.getValueType()->isFunctionTy() &&
+             "IFunc resolver must be function type!");
+      GV = Function::Create(cast<FunctionType>(I.getValueType()),
+                            GlobalValue::ExternalLinkage, I.getAddressSpace(),
+                            I.getName(), New.get());
       VMap[&I] = GV;
       continue;
     }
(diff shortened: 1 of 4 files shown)


llvm/lib/Transforms/Utils/CloneModule.cpp
```diff
@@ -117,6 +117,22 @@ std::unique_ptr<Module> llvm::CloneModule(
 
   for (const GlobalIFunc &I : M.ifuncs()) {
     // Defer setting the resolver function until after functions are cloned.
+    if (!ShouldCloneDefinition(&I)) {
+      // An ifunc also cannot act as an external reference, so we need to create
+      // either a function or a global variable depending on the value type.
+      GlobalValue *GV;
+      if (I.getValueType()->isFunctionTy())
+        GV = Function::Create(cast<FunctionType>(I.getValueType()),
```
chenzheng1030: What type will `I.getValueType()->isFunctionTy()` be if it is not a function type?

llvm/test/tools/llvm-split/ifunc.ll
```diff
@@ -0,0 +1,34 @@
+; RUN: llvm-split -j2 -o %t %s
+; RUN: llvm-dis -o - %t0 | FileCheck --check-prefix=CHECK0 %s
+; RUN: llvm-dis -o - %t1 | FileCheck --check-prefix=CHECK1 %s
+
+
```
chenzheng1030: nit: remove one empty line

llvm/test/tools/llvm-split/ifunc.ll
```diff
@@ -0,0 +1,34 @@
...
+}
+
+; CHECK0-DAG: declare void @bar_a()
+; CHECK1-DAG: define void @bar_a()
+
+define void @bar_a() {
+entry:
+  call void @foo_a.ifunc()
+  ret void
+}
```
arsenm:
```suggestion
}

```

llvm/lib/Transforms/Utils/CloneModule.cpp
arsenm: I think this is required to be a a function. This looks copied from the GlobalAlias handling

arsenm (commented):
Can you have a GlobalIAlias to GlobalIFunc?

llvm/test/tools/llvm-split/ifunc.ll
arsenm: Still has missing newline error

llvm/lib/Transforms/Utils/CloneModule.cpp
```diff
@@ -117,6 +117,18 @@ std::unique_ptr<Module> llvm::CloneModule(
...
   for (const GlobalIFunc &I : M.ifuncs()) {
     // Defer setting the resolver function until after functions are cloned.
+    if (!ShouldCloneDefinition(&I)) {
+      // An ifunc also cannot act as an external reference, so we need to create
+      // a function depending on the value type.
+      GlobalValue *GV;
+      assert(I.getValueType()->isFunctionTy() &&
+             "IFunc resolver must be function type!");
+      GV = Function::Create(cast<FunctionType>(I.getValueType()),
+                            GlobalValue::ExternalLinkage, I.getAddressSpace(),
```
arsenm, line 127: Should this preserve the linkage of the original value?

llvm/lib/Transforms/Utils/CloneModule.cpp
mmjjpp: Yes, this should be a function, and it has been modified now.

llvm/test/tools/llvm-split/ifunc.ll
mmjjpp: Has been removed

llvm/test/tools/llvm-split/ifunc.ll
mmjjpp: Thank you for your suggestion!
1. The new line has been added.
2. The example of alias to ifunc has been added.  During this process, a potential bug in splitmodule was discovered: ifunc and resolver may be not placed in the same cluster. This has been addressed in commit f782fc70.

llvm/lib/Transforms/Utils/CloneModule.cpp
mmjjpp: Here, a function type declaration will be created.  Would it be more appropriate to keep the declaration as external?

arsenm (approved):
Maybe should look into why copyAttributesFrom isn't used on this path. it looks strange to me that it's not (in the alias case either)

mmjjpp:
@dexonsmith @teresajohnson  Hello，Can you help review the code?

llvm/lib/Transforms/Utils/CloneModule.cpp
```diff
@@ -117,6 +117,18 @@ std::unique_ptr<Module> llvm::CloneModule(
 
   for (const GlobalIFunc &I : M.ifuncs()) {
     // Defer setting the resolver function until after functions are cloned.
+    if (!ShouldCloneDefinition(&I)) {
+      // An ifunc also cannot act as an external reference, so we need to create
+      // a function depending on the value type.
```
teresajohnson: I don't think the "depending on the value type" makes sense here (looked like that was cloned from the alias handling above).

teresajohnson (approved):

llvm/lib/Transforms/Utils/CloneModule.cpp
mmjjpp: Thank you for your feedback. This was indeed clone from alias handing. Now "depending on the value type" has been removed. If you agree to approve it, could you please help merge it as well? Thanks again!

chenzheng1030 (approved):
lgtm

mmjjpp:
Could you help me merge this pr?  Thank you very much. @arsenm

PR #197354, opened 2026-05-13
[clang-tidy] Remove hicpp module [4/N]

zeyi2:
This commit removes the remaining checks of `hicpp` module.

Part of https://github.com/llvm/llvm-project/issues/183462


clang-tools-extra/clang-tidy/CMakeLists.txt
```diff
@@ -62,7 +62,6 @@ add_subdirectory(custom)
 add_subdirectory(darwin)
 add_subdirectory(fuchsia)
 add_subdirectory(google)
-add_subdirectory(hicpp)
```
vbvictor, line 65: Maybe we can remove this + anchor + HICPPTidyModule.cpp‎
in separate PR in case something goes wrong.
If it does - we can revert only a small portion of changes.

vbvictor (approved):

PR #197355, opened 2026-05-13
[libc][NFC] Propagate LIBC_CONSTEXPR qualifier for those depending on bitt_cast.

lntue:
This should fix the LLVM APFloat build with gcc pre-11.


libc/src/__support/macros/attributes.h
```diff
@@ -40,15 +40,18 @@
...
+
 // TODO: Remove the macro once Clang/LLVM bump their minimum compilers' version.
 // The reason for indirection is GCC is known to fail with constexpr qualified
-// functions that doesn't produce constant expression. This avoids it by using
-// LIBC_ENABLE_CONSTEXPR as a flag to control whether the function should be
-// constexpr qualified or not.
-#if LIBC_ENABLE_CONSTEXPR &&                                                   \
-    (LIBC_HAS_BUILTIN_IS_CONSTANT_EVALUATED ||                                 \
-     (defined(LIBC_COMPILER_IS_GCC) && (LIBC_COMPILER_GCC_VER >= 1100)) ||     \
-     (defined(LIBC_COMPILER_IS_CLANG) && LIBC_COMPILER_CLANG_VER >= 900))
```
bassiounix, line 51: I'm a bit confused, why did you remove `LIBC_ENABLE_CONSTEXPR`?
This causes a lot of code corruption in the internal implementation of the math function when `LIBC_HAS_CONSTANT_EVALUATION` is defined when it's not intended to be.
I am reverting this PR

Issue #61824, opened 2023-03-29
Clang doesn't SFINAE out constrained overload when substitution-failure happens in template-head


cjdb:
GCC test `concepts-defarg2.C` fails in Clang. I'm not sure if GCC should be accepting the original lines, but Clang gets suspicious when I uncomment my addition (which I don't think should be failing).
```cpp
// PR c++/97052
// RUN: %clang_cc1 -std=c++20 -fsyntax-only -verify %s

template<typename T, typename U = typename T::type>
concept C = true;

constexpr bool f(C auto) {
  return true;
}

// added by me
// constexpr bool f(auto) {
//   return true;
// }

static_assert(f(0));

C auto x = 0;
```
Diagnostics prior to uncommenting:
```
<source>:16:15: error: no matching function for call to 'f'
static_assert(f(0));
              ^
<source>:7:16: note: candidate template ignored: constraints not satisfied [with auto:1 = int]
constexpr bool f(C auto) {
               ^
<source>:4:44: note: because substituted constraint expression is ill-formed: type 'int' cannot be used prior to '::' because it has no members
template<typename T, typename U = typename T::type>
                                           ^
<source>:4:44: error: type 'int' cannot be used prior to '::' because it has no members
template<typename T, typename U = typename T::type>
                                           ^
note: in instantiation of default argument for 'C<int>' required here
```
Diagnostics after uncommenting:
```
<source>:4:44: error: type 'int' cannot be used prior to '::' because it has no members
template<typename T, typename U = typename T::type>
                                           ^
note: in instantiation of default argument for 'C<int>' required here
```

shafik:
Based on the examples in [temp.deduct p5](https://eel.is/c++draft/temp.deduct#general-5) I think clang is not correct here.

CC @erichkeane

erichkeane:
Interestingly, `temp.deduct p5` is from a different bug report, so did you mean to respond to this one? 

In this case, it seems to me that Clang is 'correct' in the as-done: I don't know how we're supposed to pick a template parameter for 'U' without looking at T::type (though, I DO know we don't implement parameter mapping for constraint evaluation, so perhaps that is supposed to happen here?).

After the 'added', we're left over with just the definition of 'x', which DOES require instantiation of C, but again, this might be the parameter mapping stuff that @zygoloid was discussing on the EWG mailing list.

alexolog:
This looks like https://bugs.llvm.org/show_bug.cgi?id=50864, open 3 years ago.

[...]

PR #197356, opened 2026-05-13
[clang] make concept normalization a SFINAE context

mizvekov:
It is incorrect to allow the substitution failures in concept normalization to make the program ill-formed.

These can happen when comparing the equivalence of constraints for redeclaration checking, and a failure here only means these declarations are not equivalent.

For now, these diagnostics are simply discarded. It would be nice if some day, as part of diagnostics for non-matching out-of-line definitions, we would explain why matching failed for each candidate, and then add these as notes explaining why the constraints were not equivalent.

### DIFF
--- a/clang/lib/Sema/SemaTemplateInstantiate.cpp
+++ b/clang/lib/Sema/SemaTemplateInstantiate.cpp
@@ -583,7 +583,6 @@ bool Sema::CodeSynthesisContext::isInstantiationRecord() const {
   case ExceptionSpecEvaluation:
   case ConstraintSubstitution:
   case ParameterMappingSubstitution:
-  case ConstraintNormalization:
   case RewritingOperatorAsSpaceship:
   case InitializingStructuredBinding:
   case MarkingClassDllexported:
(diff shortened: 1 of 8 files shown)


cor3ntin (commented):
Does this fixes https://github.com/llvm/llvm-project/issues/61824 ? (there may be other related issues)

clang/lib/Sema/SemaTemplateInstantiate.cpp
```diff
@@ -2862,7 +2847,7 @@ TypeSourceInfo *Sema::SubstType(TypeSourceInfo *T,
                                 SourceLocation Loc,
                                 DeclarationName Entity,
                                 bool AllowDeducedTST) {
-  assert(!CodeSynthesisContexts.empty() &&
+  assert((!CodeSynthesisContexts.empty() || isSFINAEContext()) &&
```
erichkeane, line 2850: One nit: I'd like the condition in this assert 'extracted' somewhere.  It is used often enough and I can imagine us using it elsewhere at a later time, I'd like it to both be clear what it is testing by name (TBH, the CodeSynthesisContexts.empty test is a little opaque anyway), and be less likely to be skipped.

erichkeane (commented):
I have 1 nit, else i think this is good.  I'm going to hold off on approval since I want to take a little extra time to grok it.

mizvekov:
> Does this fixes #61824 ? (there may be other related issues)

I don't think so, I commented on the issue, but I think all remaining issues are due to the parameter mapping.

I'd be happy to add test cases and close issues from this PR if anyone finds one.

ojhunt:
Sorry, I'll get to this later today - maybe ~3 hours from now if that's ok

clang/lib/Sema/SemaTemplateInstantiate.cpp
mizvekov: Would you mind if we did that in a follow up? I expanded this assert to cover more cases, and it turned up another real bug.

erichkeane: Hah, fun!  Ok, yeah, I'm good doing this as a part of that fixup.

erichkeane (approved):

clang/lib/Sema/SemaTemplateInstantiate.cpp
mizvekov: And this is the follow-up: https://github.com/llvm/llvm-project/pull/197523

ojhunt (approved):
LGTM

PR #197358, opened 2026-05-13
[clang] Implement C2y stdc_memreverse8 and stdc_memreverse8u{8,16,32,64} builtins

chaitanyav:
Implements the C2y <stdbit.h> memory reversal functions stdc_memreverse8 and stdc_memreverse8u{8,16,32,64}. The typed variants lower to llvm.bswap and support constexpr evaluation.


clang/include/clang/Basic/Builtins.td
```diff
@@ -34,6 +34,10 @@ class IntBitUtilTemplate : Template<
      "unsigned long int", "unsigned long long int"],
     ["_uc", "_us", "_ui", "_ul", "_ull"]>;
 
+class MemReverse8Template : Template<
+    ["unsigned char", "unsigned short", "uint32_t", "uint64_t"],
```
AaronBallman: It's a bit interesting to see `uint32_t` but `unsigned char` instead of `uint8_t`, do we not support the smaller bit widths here?

clang/include/clang/Basic/Builtins.td
```diff
@@ -986,6 +990,18 @@ def StdcRotateRightTyped : LibBuiltin<"stdbit.h", "C2Y_LANG">, IntBitUtilTemplat
   let Prototype = "T(T, unsigned int)";
 }
 
+def StdcMemReverse8: Builtin {
+  let Spellings = ["__builtin_stdc_memreverse8"];
+  let Attributes = [NoThrow, NonNull<NonOptimizing, [1]>];
+  let Prototype = "void(size_t, unsigned char*)";
```
AaronBallman, line 996: I think this prototype is reasonable as-is, but the official prototype is:
```
void stdc_memreverse8(size_t n, unsigned char ptr[static n]);
```
and that actually associates some good bounds information that could be used by other analyses, at least in C. Nothing to change, but it might be nice if we had a way to use the correct signature for C and fall back to the decayed signature for C++.

clang/include/clang/Basic/Builtins.td
```diff
@@ -986,6 +990,18 @@ def StdcRotateRightTyped : LibBuiltin<"stdbit.h", "C2Y_LANG">, IntBitUtilTemplat
   let Prototype = "T(T, unsigned int)";
 }
 
+def StdcMemReverse8: Builtin {
+  let Spellings = ["__builtin_stdc_memreverse8"];
+  let Attributes = [NoThrow, NonNull<NonOptimizing, [1]>];
```
AaronBallman, line 995: Should this one also be `pure`?

clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -11512,6 +11512,8 @@ def err_builtin_setjmp_unsupported : Error<
 def err_builtin_longjmp_invalid_val : Error<
   "argument to __builtin_longjmp must be a constant 1">;
 def err_builtin_requires_language : Error<"'%0' is only available in %1">;
+def err_builtin_requires_char_bit_8 : Error<
+  "'%0' is only available on targets where CHAR_BIT == 8">;
```
AaronBallman:
```suggestion
  "'%0' is only available on targets where 'CHAR_BIT == 8'">;
```

clang/lib/Sema/SemaChecking.cpp
```diff
@@ -3863,6 +3863,18 @@ Sema::CheckBuiltinFunctionCall(FunctionDecl *FDecl, unsigned BuiltinID,
       return ExprError();
     break;
 
+  case Builtin::BI__builtin_stdc_memreverse8:
+  case Builtin::BIstdc_memreverse8u8:
+  case Builtin::BIstdc_memreverse8u16:
+  case Builtin::BIstdc_memreverse8u32:
+  case Builtin::BIstdc_memreverse8u64:
+    if (Context.getTargetInfo().getCharWidth() != 8) {
+      Diag(TheCall->getBeginLoc(), diag::err_builtin_requires_char_bit_8)
```
AaronBallman, line 3943: I agree with the code but I also don't think we have a way to test it either; we don't have options to let you control `CHAR_BIT` and we don't have targets that aren't 8-bit currently.

clang/include/clang/Basic/Builtins.td
chaitanyav: i am seeing this  build error when using `uint8_t`

```cpp
/home/naga/llvm-project/clang/include/clang/Basic/Builtins.td:1002:7: error: Unknown Type: uint8_t
  let Prototype = "T(T)";
```

clang/include/clang/Basic/Builtins.td
chaitanyav: the function is modifying the buffer pointed to by the ptr and the return type is void

chaitanyav: Added `Pure` to the typed variants

clang/include/clang/Basic/Builtins.td
AaronBallman: That's because we haven't added support for them yet, but I think we should if we intend to use them: 

https://github.com/llvm/llvm-project/blob/1e1a1ff8a6c6d9bc7f0044c79c6cfe0819246e00/clang/utils/TableGen/ClangBuiltinsEmitter.cpp#L241
https://github.com/llvm/llvm-project/blob/1e1a1ff8a6c6d9bc7f0044c79c6cfe0819246e00/clang/include/clang/Basic/Builtins.def#L53

WDYT @efriedma-quic ?

efriedma-quic: It doesn't have much practical impact, but it's probably worth trying to be consistent with the C standard definitions, sure.

chaitanyav: @AaronBallman @efriedma-quic  am adding support for those bit widths

clang/utils/TableGen/ClangBuiltinsEmitter.cpp
```diff
@@ -359,6 +359,8 @@ class PrototypeParser {
                                .Case("float", "f")
                                .Case("id", "G")
                                .Case("int", "i")
+                               .Case("int8_t", "Sc")
+                               .Case("int16_t", "Ss")
```
AaronBallman: Do we want to keep the `i` suffix for `int`? e.g., `Ci` and `Si` (for Char-int and Short-int)?

clang/utils/TableGen/ClangBuiltinsEmitter.cpp
```diff
@@ -372,6 +374,8 @@ class PrototypeParser {
                                .Case("sigjmp_buf", "SJ")
                                .Case("size_t", "z")
                                .Case("ucontext_t", "K")
+                               .Case("uint8_t", "Uc")
+                               .Case("uint16_t", "Us")
```
AaronBallman: Similar question about `i` as a suffix here.

clang/utils/TableGen/ClangBuiltinsEmitter.cpp
chaitanyav: `Si` would conflict with signed integer. Can we use `Hi` for `int16_t`?

AaronBallman: Good call! I'm not strongly tied to `C` and `S` as the initial character, just `i` as the trailing one. So I'd be fine with `Ci` and `Hi`.

chaitanyav: I used B(int8_t), T (int16_t) since H is used for SEL.

AaronBallman (approved):
LGTM! But please give @efriedma-quic a chance to weigh-in before landing.

clang/include/clang/Basic/Builtins.td
```diff
@@ -986,6 +990,18 @@ def StdcRotateRightTyped : LibBuiltin<"stdbit.h", "C2Y_LANG">, IntBitUtilTemplat
...
+  let Prototype = "void(size_t, unsigned char*)";
+}
```
efriedma-quic, line 998: Do we also want a builtin `Builtin::BIstdc_memreverse8`?

clang/lib/CodeGen/CGBuiltin.cpp
```diff
@@ -4261,6 +4261,64 @@ RValue CodeGenFunction::EmitBuiltinExpr(const GlobalDecl GD, unsigned BuiltinID,
...
+    return RValue::get(EmitScalarExpr(E->getArg(0)));
+
+  case Builtin::BIstdc_memreverse8u16:
+  case Builtin::BIstdc_memreverse8u32:
+  case Builtin::BIstdc_memreverse8u64:
+    return RValue::get(
+        emitBuiltinWithOneOverloadedType<1>(*this, E, Intrinsic::bswap));
+
+  case Builtin::BI__builtin_stdc_memreverse8: {
+    Value *N = EmitScalarExpr(E->getArg(0));
```
efriedma-quic: Can we just mark this builtin FunctionWithBuiltinPrefix, and use the default codegen of calling the stdc_memreverse8 library function?

The general case loop iterating one byte at a time is going to have terrible performance unless some LLVM optimization manages to pattern-match it, and I don't think there's currently any such optimization.

clang/lib/AST/ExprConstant.cpp
```diff
@@ -16546,7 +16546,11 @@ bool IntExprEvaluator::VisitBuiltinCallExpr(const CallExpr *E,
   case Builtin::BI__builtin_bswapg:
   case Builtin::BI__builtin_bswap16:
   case Builtin::BI__builtin_bswap32:
-  case Builtin::BI__builtin_bswap64: {
+  case Builtin::BI__builtin_bswap64:
+  case Builtin::BIstdc_memreverse8u8:
+  case Builtin::BIstdc_memreverse8u16:
+  case Builtin::BIstdc_memreverse8u32:
+  case Builtin::BIstdc_memreverse8u64: {
```
efriedma-quic, line 16559: We could theoretically constant-evaluate stdc_memreverse8, but I guess it would only be relevant for C++ constexpr functions.

clang/lib/CodeGen/CGBuiltin.cpp
```diff
@@ -4230,6 +4230,38 @@ RValue CodeGenFunction::EmitBuiltinExpr(const GlobalDecl GD, unsigned BuiltinID,
...
+        Address Addr = PtrAddr.withElementType(IntTy);
+        Value *Val = Builder.CreateLoad(Addr);
+        Function *F = CGM.getIntrinsic(Intrinsic::bswap, IntTy);
+        Value *Swapped = Builder.CreateCall(F, Val);
+        Builder.CreateStore(Swapped, Addr);
+        return RValue::get(nullptr);
+      }
+    }
+
+    // General case: fall back to the library function stdc_memreverse8.
```
efriedma-quic, line 4264: You can't fall back after you've already emitted the argument expressions; that will double-emit the arguments.

chaitanyav: i see what you mean 

```bash
  %0 = load i32, ptr @n, align 4
  store i32 %inc, ptr @n, align 4
  %1 = load ptr, ptr @p, align 8
  store ptr %incdec.ptr, ptr @p, align 8
  %2 = load i32, ptr @n, align 4
  store i32 %inc1, ptr @n, align 4
  %3 = load ptr, ptr @p, align 8
  store ptr %incdec.ptr3, ptr @p, align 8
  call void @stdc_memreverse8(i64 noundef %conv2, ptr noundef %3) #2
```

clang/lib/AST/ExprConstant.cpp
chaitanyav: I am looking into making it constexpr

efriedma-quic: We can look at this as a followup.

efriedma-quic (approved):
LGTM

chaitanyav:
Thank you @efriedma-quic and @AaronBallman for the review.

chaitanyav:
please help with merging this PR.

PR #197359, opened 2026-05-13
[BOLT] Fix data race in multi-threaded DWP type unit processing and DWP type unit duplication

Thrrreeee:
## Summary
This PR fixes a race condition in LLVM BOLT's DIEBuilder::buildTypeUnits() that is triggered when DWARF5 split-DWARF (.dwo/.dwp) inputs are processed with multi-threaded CU processing. Concurrent invocations from different worker threads share the same DWP type-unit state, which results in duplicated DIE extraction, assertion failures, and intermittent crashes. The fix serializes buildTypeUnits() for DWP inputs via a function-local static std::mutex, leaving the non-DWO fast path unchanged.
## Problem Description
When BOLT processes DWARF debug info with --debug-thread-count=4 --cu-processing-batch-size=4 on testcase dwarf5-df-types-dup-dwp-input.test, multiple threads concurrently call DIEBuilder::buildTypeUnits() on shared DWP type units. Since type units within a DWP file are shared across compilation units, multiple threads may attempt to extract DIEs from the same type unit simultaneously, violating the assertion.

**Note:** This is an intermittent failure — due to its dependence on thread scheduling and interleaving, it does not reproduce on every run, and may require multiple executions under the multi-threaded configuration to surface, which is why it escaped earlier single-threaded testing.

### DIFF
--- a/bolt/include/bolt/Core/BinaryContext.h
+++ b/bolt/include/bolt/Core/BinaryContext.h
@@ -286,7 +286,7 @@ class BinaryContext {
   void deregisterSectionName(const BinarySection &Section);
 
   /// Mutex used for parallel processing of DWP type units.
-  mutable std::recursive_mutex DWPUnitsMutex;
+  mutable std::mutex DWPUnitsMutex;
 
 public:
   static Expected<std::unique_ptr<BinaryContext>> createBinaryContext(
@@ -295,7 +295,7 @@ public:
       std::unique_ptr<DWARFContext> DwCtx, JournalingStreams Logger);
 
   /// Returns the mutex guarding concurrent access to DWP units.
-  std::recursive_mutex &getUnitsMutex() { return DWPUnitsMutex; }
+  std::mutex &getUnitsMutex() { return DWPUnitsMutex; }
 
   /// Superset of compiler units that will contain overwritten code that needs
   /// new debug info. In a few cases, functions may end up not being
--- a/bolt/include/bolt/Core/BinaryContext.h
+++ b/bolt/include/bolt/Core/BinaryContext.h
@@ -286,7 +286,7 @@ class BinaryContext {
   void deregisterSectionName(const BinarySection &Section);
 
   /// Mutex used for parallel processing of DWP type units.
-  mutable std::mutex DWPUnitsMutex;
+  std::mutex DWPUnitsMutex;
 
 public:
   static Expected<std::unique_ptr<BinaryContext>> createBinaryContext(
--- a/bolt/include/bolt/Core/BinaryContext.h
+++ b/bolt/include/bolt/Core/BinaryContext.h
@@ -48,6 +48,7 @@
 #include <functional>
 #include <list>
 #include <map>
+#include <mutex>
 #include <optional>
 #include <set>
 #include <string>
@@ -284,11 +285,17 @@ class BinaryContext {
   /// Internal helper for removing section name from a lookup table.
   void deregisterSectionName(const BinarySection &Section);
 
+  /// Mutex used for parallel processing of DWP type units.
+  mutable std::recursive_mutex DWPUnitsMutex;
+
 public:
   static Expected<std::unique_ptr<BinaryContext>> createBinaryContext(
       Triple TheTriple, std::shared_ptr<orc::SymbolStringPool> SSP,
       StringRef InputFileName, SubtargetFeatures *Features, bool IsPIC,
       std::unique_ptr<DWARFContext> DwCtx, JournalingStreams Logger);
+
+  /// Returns the mutex guarding concurrent access to DWP units.
+  std::recursive_mutex &getUnitsMutex() { return DWPUnitsMutex; }
 
   /// Superset of compiler units that will contain overwritten code that needs
   /// new debug info. In a few cases, functions may end up not being
--- a/bolt/lib/Core/DIEBuilder.cpp
+++ b/bolt/lib/Core/DIEBuilder.cpp
@@ -238,22 +238,6 @@ static unsigned int getCUNum(DWARFContext *DwarfContext, bool IsDWO) {
 
 void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
                                 const bool Init) {
-  if (DwarfContext->isDWP()) {
-    buildTypeUnitsThreadSafe(StrOffsetWriter, Init);
-    return;
-  }
-
-  buildTypeUnitsImpl(StrOffsetWriter, Init);
-}
-
-void DIEBuilder::buildTypeUnitsThreadSafe(
-    DebugStrOffsetsWriter *StrOffsetWriter, const bool Init) {
-  std::unique_lock<std::recursive_mutex> LockGuard(BC.getUnitsMutex());
-  buildTypeUnitsImpl(StrOffsetWriter, Init);
-}
-
-void DIEBuilder::buildTypeUnitsImpl(DebugStrOffsetsWriter *StrOffsetWriter,
-                                    const bool Init) {
   if (Init)
     BuilderState.reset(new State());
 
@@ -297,6 +281,73 @@ void DIEBuilder::buildTypeUnitsImpl(DebugStrOffsetsWriter *StrOffsetWriter,
     if (StrOffsetWriter)
       StrOffsetWriter->finalizeSection(*DU, *this);
   }
+}
+
+static void collectReferencedTypeSignatures(DWARFDie Die,
+                                            DenseSet<uint64_t> &ProcessedTU,
+                                            SmallVectorImpl<uint64_t> &TUlist) {
+  SmallVector<DWARFDie, 8> DIElist;
+  DIElist.push_back(Die);
+
+  while (!DIElist.empty()) {
+    DWARFDie Current = DIElist.pop_back_val();
+    if (!Current)
+      continue;
+
+    for (const DWARFAttribute &Attr : Current.attributes()) {
+      if (Attr.Value.getForm() != dwarf::DW_FORM_ref_sig8)
+        continue;
+      if (const std::optional<uint64_t> Signature =
+              Attr.Value.getAsSignatureReference())
+        if (ProcessedTU.insert(*Signature).second)
+          TUlist.push_back(*Signature);
+    }
+
+    for (DWARFDie Child : Current.children())
+      DIElist.push_back(Child);
+  }
+}
+
+void DIEBuilder::buildDWPTypeUnitsForUnit(DWARFUnit &U) {
+  // Avoid processing the same type unit multiple times.
+  DenseSet<uint64_t> ProcessedTU;
+  SmallVector<uint64_t, 8> TUlist;
+  // Collecting signatures of type units referenced by this unit.
+  collectReferencedTypeSignatures(U.getUnitDIE(), ProcessedTU, TUlist);
+
+  // addressing type units referenced.
+  for (unsigned I = 0; I != TUlist.size(); ++I) {
+    const uint64_t Signature = TUlist[I];
+    DWARFTypeUnit *TU = DwarfContext->getTypeUnitForHash(Signature, true);
+    if (!TU)
+      continue;
+
+    getState().Type = TU->getVersion() < 5 ? ProcessingType::DWARF4TUs
+                                           : ProcessingType::DWARF5TUs;
+    if (!registerUnit(*TU, false))
+      continue;
+
+    const std::optional<uint32_t> UnitId = getUnitId(*TU);
+    if (!UnitId || getState().CloneUnitCtxMap[*UnitId].IsConstructed)
+      continue;
+
+    collectReferencedTypeSignatures(TU->getUnitDIE(), ProcessedTU, TUlist);
+  }
+
+  // Ensure deterministic order of processing type units
+  auto SortByOffset = [](const DWARFUnit *A, const DWARFUnit *B) {
+    return A->getOffset() < B->getOffset();
+  };
+  llvm::sort(getState().DWARF4TUVector, SortByOffset);
+  llvm::sort(getState().DWARF5TUVector, SortByOffset);
+
+  getState().Type = ProcessingType::DWARF4TUs;
+  for (DWARFUnit *TU : getState().DWARF4TUVector)
+    constructFromUnit(*TU);
+
+  getState().Type = ProcessingType::DWARF5TUs;
+  for (DWARFUnit *TU : getState().DWARF5TUVector)
+    constructFromUnit(*TU);
 }
 
 void DIEBuilder::buildCompileUnits(const bool Init) {
@@ -344,7 +395,12 @@ void DIEBuilder::buildCompileUnits(const std::vector<DWARFUnit *> &CUs) {
 void DIEBuilder::buildDWOUnit(DWARFUnit &U) {
   BuilderState.release();
   BuilderState = std::make_unique<State>();
-  buildTypeUnits(nullptr, false);
+  if (DwarfContext->isDWP()) {
+    std::unique_lock<std::recursive_mutex> LockGuard(BC.getUnitsMutex());
+    buildDWPTypeUnitsForUnit(U);
+  } else {
+    buildTypeUnits(nullptr, false);
+  }
   getState().Type = ProcessingType::CUs;
   registerUnit(U, false);
   constructFromUnit(U);
--- a/bolt/lib/Core/DIEBuilder.cpp
+++ b/bolt/lib/Core/DIEBuilder.cpp
@@ -396,7 +396,7 @@ void DIEBuilder::buildDWOUnit(DWARFUnit &U) {
   BuilderState.release();
   BuilderState = std::make_unique<State>();
   if (DwarfContext->isDWP()) {
-    std::unique_lock<std::recursive_mutex> LockGuard(BC.getUnitsMutex());
+    std::unique_lock<std::mutex> LockGuard(BC.getUnitsMutex());
     buildDWPTypeUnitsForUnit(U);
   } else {
     buildTypeUnits(nullptr, false);
--- a/bolt/lib/Core/DIEBuilder.cpp
+++ b/bolt/lib/Core/DIEBuilder.cpp
@@ -238,6 +238,22 @@ static unsigned int getCUNum(DWARFContext *DwarfContext, bool IsDWO) {
 
 void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
                                 const bool Init) {
+  if (isDWO() && DwarfContext && DwarfContext->isDWP()) {
+    buildTypeUnitsThreadSafe(StrOffsetWriter, Init);
+    return;
+  }
+
+  buildTypeUnitsImpl(StrOffsetWriter, Init);
+}
+
+void DIEBuilder::buildTypeUnitsThreadSafe(
+    DebugStrOffsetsWriter *StrOffsetWriter, const bool Init) {
+  std::unique_lock<std::recursive_mutex> LockGuard(BC.getUnitsMutex());
+  buildTypeUnitsImpl(StrOffsetWriter, Init);
+}
+
+void DIEBuilder::buildTypeUnitsImpl(DebugStrOffsetsWriter *StrOffsetWriter,
+                                    const bool Init) {
   if (Init)
     BuilderState.reset(new State());
 
--- a/bolt/lib/Core/DIEBuilder.cpp
+++ b/bolt/lib/Core/DIEBuilder.cpp
@@ -283,6 +283,11 @@ void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
   }
 }
 
+/// Recursively collects type unit signatures from the given DIE and all of its
+/// children.
+///
+/// Note: De-duplication of the collected signatures is handled at the outer
+/// level by registerUnit.
 static void collectReferencedTypeSignatures(DWARFDie Die,
                                             DenseSet<uint64_t> &ProcessedTU,
                                             SmallVectorImpl<uint64_t> &TUlist) {
@@ -309,6 +314,7 @@ static void collectReferencedTypeSignatures(DWARFDie Die,
 }
 
 void DIEBuilder::buildDWPTypeUnitsForUnit(DWARFUnit &U) {
+  std::unique_lock<std::mutex> LockGuard(BC.getUnitsMutex());
   // Avoid processing the same type unit multiple times.
   DenseSet<uint64_t> ProcessedTU;
   SmallVector<uint64_t, 8> TUlist;
@@ -334,19 +340,18 @@ void DIEBuilder::buildDWPTypeUnitsForUnit(DWARFUnit &U) {
     collectReferencedTypeSignatures(TU->getUnitDIE(), ProcessedTU, TUlist);
   }
 
-  // Ensure deterministic order of processing type units
+  // Ensure original order of processing type units
   auto SortByOffset = [](const DWARFUnit *A, const DWARFUnit *B) {
     return A->getOffset() < B->getOffset();
   };
-  llvm::sort(getState().DWARF4TUVector, SortByOffset);
-  llvm::sort(getState().DWARF5TUVector, SortByOffset);
 
-  getState().Type = ProcessingType::DWARF4TUs;
-  for (DWARFUnit *TU : getState().DWARF4TUVector)
-    constructFromUnit(*TU);
+  // For Split DWARF, we have either DWARF4 or DWARF5, they cannot be mixed.
+  std::vector<DWARFUnit *> &TUVec = !getState().DWARF4TUVector.empty()
+                                        ? getState().DWARF4TUVector
+                                        : getState().DWARF5TUVector;
+  llvm::sort(TUVec, SortByOffset);
 
-  getState().Type = ProcessingType::DWARF5TUs;
-  for (DWARFUnit *TU : getState().DWARF5TUVector)
+  for (DWARFUnit *TU : TUVec)
     constructFromUnit(*TU);
 }
 
@@ -396,7 +401,6 @@ void DIEBuilder::buildDWOUnit(DWARFUnit &U) {
   BuilderState.release();
   BuilderState = std::make_unique<State>();
   if (DwarfContext->isDWP()) {
-    std::unique_lock<std::mutex> LockGuard(BC.getUnitsMutex());
     buildDWPTypeUnitsForUnit(U);
   } else {
     buildTypeUnits(nullptr, false);
--- a/bolt/lib/Core/DIEBuilder.cpp
+++ b/bolt/lib/Core/DIEBuilder.cpp
@@ -321,15 +321,14 @@ void DIEBuilder::buildDWPTypeUnitsForUnit(DWARFUnit &U) {
   // Collecting signatures of type units referenced by this unit.
   collectReferencedTypeSignatures(U.getUnitDIE(), ProcessedTU, TUlist);
 
+  getState().Type = U.getVersion() < 5 ? ProcessingType::DWARF4TUs
+                                       : ProcessingType::DWARF5TUs;
   // addressing type units referenced.
   for (unsigned I = 0; I != TUlist.size(); ++I) {
     const uint64_t Signature = TUlist[I];
     DWARFTypeUnit *TU = DwarfContext->getTypeUnitForHash(Signature, true);
     if (!TU)
       continue;
-
-    getState().Type = TU->getVersion() < 5 ? ProcessingType::DWARF4TUs
-                                           : ProcessingType::DWARF5TUs;
     if (!registerUnit(*TU, false))
       continue;
 
@@ -346,7 +345,7 @@ void DIEBuilder::buildDWPTypeUnitsForUnit(DWARFUnit &U) {
   };
 
   // For Split DWARF, we have either DWARF4 or DWARF5, they cannot be mixed.
-  std::vector<DWARFUnit *> &TUVec = !getState().DWARF4TUVector.empty()
+  std::vector<DWARFUnit *> &TUVec = getState().Type == ProcessingType::DWARF4TUs
                                         ? getState().DWARF4TUVector
                                         : getState().DWARF5TUVector;
   llvm::sort(TUVec, SortByOffset);
--- a/bolt/lib/Core/DIEBuilder.cpp
+++ b/bolt/lib/Core/DIEBuilder.cpp
@@ -238,7 +238,7 @@ static unsigned int getCUNum(DWARFContext *DwarfContext, bool IsDWO) {
 
 void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
                                 const bool Init) {
-  if (isDWO() && DwarfContext && DwarfContext->isDWP()) {
+  if (DwarfContext->isDWP()) {
     buildTypeUnitsThreadSafe(StrOffsetWriter, Init);
     return;
   }
--- a/bolt/lib/Core/DIEBuilder.cpp
+++ b/bolt/lib/Core/DIEBuilder.cpp
@@ -268,6 +268,22 @@ static unsigned int getCUNum(DWARFContext *DwarfContext, bool IsDWO) {
 
 void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
                                 const bool Init) {
+  if (isDWO() && DwarfContext && DwarfContext->isDWP()) {
+    buildTypeUnitsThreadSafe(StrOffsetWriter, Init);
+    return;
+  }
+
+  buildTypeUnitsImpl(StrOffsetWriter, Init);
+}
+
+void DIEBuilder::buildTypeUnitsThreadSafe(
+    DebugStrOffsetsWriter *StrOffsetWriter, const bool Init) {
+  std::unique_lock<std::recursive_mutex> LockGuard(BC.getUnitsMutex());
+  buildTypeUnitsImpl(StrOffsetWriter, Init);
+}
+
+void DIEBuilder::buildTypeUnitsImpl(DebugStrOffsetsWriter *StrOffsetWriter,
+                                    const bool Init) {
   if (Init)
     BuilderState.reset(new State());
 
--- a/bolt/lib/Rewrite/RewriteInstance.cpp
+++ b/bolt/lib/Rewrite/RewriteInstance.cpp
@@ -441,10 +441,10 @@ RewriteInstance::RewriteInstance(ELFObjectFileBase *File, const int Argc,
   auto BCOrErr = BinaryContext::createBinaryContext(
       TheTriple, std::make_shared<orc::SymbolStringPool>(), File->getFileName(),
       Features.get(), IsPIC,
-      DWARFContext::create(*File, DWARFContext::ProcessDebugRelocations::Ignore,
-                           nullptr, opts::DWPPathName,
-                           WithColor::defaultErrorHandler,
-                           WithColor::defaultWarningHandler),
+      DWARFContext::create(
+          *File, DWARFContext::ProcessDebugRelocations::Ignore, nullptr,
+          opts::DWPPathName, WithColor::defaultErrorHandler,
+          WithColor::defaultWarningHandler, !opts::DWPPathName.empty()),
       JournalingStreams{Stdout, Stderr});
   if (Error E = BCOrErr.takeError()) {
     Err = std::move(E);
--- a/bolt/lib/Rewrite/RewriteInstance.cpp
+++ b/bolt/lib/Rewrite/RewriteInstance.cpp
@@ -444,7 +444,7 @@ RewriteInstance::RewriteInstance(ELFObjectFileBase *File, const int Argc,
       DWARFContext::create(
           *File, DWARFContext::ProcessDebugRelocations::Ignore, nullptr,
           opts::DWPPathName, WithColor::defaultErrorHandler,
-          WithColor::defaultWarningHandler, !opts::DWPPathName.empty()),
+          WithColor::defaultWarningHandler, true),
       JournalingStreams{Stdout, Stderr});
   if (Error E = BCOrErr.takeError()) {
     Err = std::move(E);
--- a/bolt/lib/Rewrite/RewriteInstance.cpp
+++ b/bolt/lib/Rewrite/RewriteInstance.cpp
@@ -441,10 +441,10 @@ RewriteInstance::RewriteInstance(ELFObjectFileBase *File, const int Argc,
   auto BCOrErr = BinaryContext::createBinaryContext(
       TheTriple, std::make_shared<orc::SymbolStringPool>(), File->getFileName(),
       Features.get(), IsPIC,
-      DWARFContext::create(
-          *File, DWARFContext::ProcessDebugRelocations::Ignore, nullptr,
-          opts::DWPPathName, WithColor::defaultErrorHandler,
-          WithColor::defaultWarningHandler, true),
+      DWARFContext::create(*File, DWARFContext::ProcessDebugRelocations::Ignore,
+                           nullptr, opts::DWPPathName,
+                           WithColor::defaultErrorHandler,
+                           WithColor::defaultWarningHandler, true),
       JournalingStreams{Stdout, Stderr});
   if (Error E = BCOrErr.takeError()) {
     Err = std::move(E);
--- a/bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
+++ b/bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
@@ -7,7 +7,7 @@
 ; RUN: -split-dwarf-file=helper.dwo -o helper.o
 ; RUN: %clang %cflags -gdwarf-5 -gsplit-dwarf=split main.o helper.o -o main.exe
 ; RUN: llvm-dwp -e main.exe -o main.exe.dwp
-; RUN: llvm-bolt main.exe -o main.exe.bolt --update-debug-sections
+; RUN: llvm-bolt main.exe -o main.exe.bolt --update-debug-sections  --debug-thread-count=4 --cu-processing-batch-size=4
 ; RUN: llvm-dwarfdump --debug-info -r 0 main.dwo.dwo | FileCheck -check-prefix=BOLT-DWO-DWO-MAIN %s
 ; RUN: llvm-dwarfdump --debug-info -r 0 helper.dwo.dwo | FileCheck -check-prefix=BOLT-DWO-DWO-HELPER %s
 
--- a/bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
+++ b/bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
@@ -7,7 +7,7 @@
 ; RUN: -split-dwarf-file=helper.dwo -o helper.o
 ; RUN: %clang %cflags -gdwarf-5 -gsplit-dwarf=split main.o helper.o -o main.exe
 ; RUN: llvm-dwp -e main.exe -o main.exe.dwp
-; RUN: llvm-bolt main.exe -o main.exe.bolt --update-debug-sections  --debug-thread-count=4 --cu-processing-batch-size=4
+; RUN: llvm-bolt main.exe -o main.exe.bolt --update-debug-sections
 ; RUN: llvm-dwarfdump --debug-info -r 0 main.dwo.dwo | FileCheck -check-prefix=BOLT-DWO-DWO-MAIN %s
 ; RUN: llvm-dwarfdump --debug-info -r 0 helper.dwo.dwo | FileCheck -check-prefix=BOLT-DWO-DWO-HELPER %s
 
--- a/bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
+++ b/bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
@@ -11,17 +11,18 @@
 ; RUN: llvm-dwarfdump --debug-info -r 0 main.dwo.dwo | FileCheck -check-prefix=BOLT-DWO-DWO-MAIN %s
 ; RUN: llvm-dwarfdump --debug-info -r 0 helper.dwo.dwo | FileCheck -check-prefix=BOLT-DWO-DWO-HELPER %s
 
-;; Tests that BOLT correctly handles DWARF5 DWP file as input. Output has correct CU, and all the type units are written out.
+;; Tests that BOLT correctly handles DWARF5 DWP file as input. Output has the correct CU,
+;; and only type units referenced by that CU are written out.
 
 ; BOLT-DWO-DWO-MAIN: debug_info.dwo
 ; BOLT-DWO-DWO-MAIN-NEXT: type_signature = 0x49dc260088be7e56
 ; BOLT-DWO-DWO-MAIN: type_signature = 0x104ec427d2ebea6f
-; BOLT-DWO-DWO-MAIN: type_signature = 0xca1e65a66d92b970
+; BOLT-DWO-DWO-MAIN-NOT: type_signature = 0xca1e65a66d92b970
 ; BOLT-DWO-DWO-MAIN: Compile Unit
 ; BOLT-DWO-DWO-MAIN-SAME: DWO_id = 0x52bda211bf6d26b7
 ; BOLT-DWO-DWO-MAIN-NOT: Compile Unit
 ; BOLT-DWO-DWO-HELPER: debug_info.dwo
-; BOLT-DWO-DWO-HELPER-NEXT: type_signature = 0x49dc260088be7e56
+; BOLT-DWO-DWO-HELPER-NOT: type_signature = 0x49dc260088be7e56
 ; BOLT-DWO-DWO-HELPER: type_signature = 0x104ec427d2ebea6f
 ; BOLT-DWO-DWO-HELPER: type_signature = 0xca1e65a66d92b970
 ; BOLT-DWO-DWO-HELPER: Compile Unit
--- a/bolt/include/bolt/Core/DIEBuilder.h
+++ b/bolt/include/bolt/Core/DIEBuilder.h
@@ -217,6 +217,14 @@ private:
   /// Returns true if DWARFUnit is registered successfully.
   bool registerUnit(DWARFUnit &DU, bool NeedSort);
 
+  /// Builds type units without extra synchronization.
+  void buildTypeUnitsImpl(DebugStrOffsetsWriter *StrOffsetWriter,
+                          const bool Init);
+
+  /// Serializes type-unit construction for DWP-backed split DWARF contexts.
+  void buildTypeUnitsThreadSafe(DebugStrOffsetsWriter *StrOffsetWriter,
+                                const bool Init);
+
   /// \return the unique ID of \p U if it exists.
   std::optional<uint32_t> getUnitId(const DWARFUnit &DU);
 
--- a/bolt/include/bolt/Core/DIEBuilder.h
+++ b/bolt/include/bolt/Core/DIEBuilder.h
@@ -217,13 +217,8 @@ private:
   /// Returns true if DWARFUnit is registered successfully.
   bool registerUnit(DWARFUnit &DU, bool NeedSort);
 
-  /// Builds type units without extra synchronization.
-  void buildTypeUnitsImpl(DebugStrOffsetsWriter *StrOffsetWriter,
-                          const bool Init);
-
-  /// Serializes type-unit construction for DWP-backed split DWARF contexts.
-  void buildTypeUnitsThreadSafe(DebugStrOffsetsWriter *StrOffsetWriter,
-                                const bool Init);
+  /// Builds type units needed in the DWO.
+  void buildDWPTypeUnitsForUnit(DWARFUnit &U);
 
   /// \return the unique ID of \p U if it exists.
   std::optional<uint32_t> getUnitId(const DWARFUnit &DU);
--- a/bolt/include/bolt/Core/DIEBuilder.h
+++ b/bolt/include/bolt/Core/DIEBuilder.h
@@ -220,6 +220,14 @@ private:
   /// Returns true if DWARFUnit is registered successfully.
   bool registerUnit(DWARFUnit &DU, bool NeedSort);
 
+  /// Builds type units without extra synchronization.
+  void buildTypeUnitsImpl(DebugStrOffsetsWriter *StrOffsetWriter,
+                          const bool Init);
+
+  /// Serializes type-unit construction for DWP-backed split DWARF contexts.
+  void buildTypeUnitsThreadSafe(DebugStrOffsetsWriter *StrOffsetWriter,
+                                const bool Init);
+
   /// \return the unique ID of \p U if it exists.
   std::optional<uint32_t> getUnitId(const DWARFUnit &DU);
 
--- a/bolt/lib/Rewrite/DWARFRewriter.cpp
+++ b/bolt/lib/Rewrite/DWARFRewriter.cpp
@@ -496,12 +496,8 @@ static void emitDWOBuilder(const std::string &DWOName,
       createDIEStreamer(*TheTriple, *ObjOS, "DwoStreamerInitAug2",
                         DWODIEBuilder, GDBIndexSection);
   if (SplitCU.getContext().getMaxDWOVersion() >= 5) {
-    for (std::unique_ptr<llvm::DWARFUnit> &CU :
-         SplitCU.getContext().dwo_info_section_units()) {
-      if (!CU->isTypeUnit())
-        continue;
+    for (DWARFUnit *CU : DWODIEBuilder.getDWARF5TUVector())
       emitUnit(DWODIEBuilder, *Streamer, *CU);
-    }
     emitUnit(DWODIEBuilder, *Streamer, SplitCU);
   } else {
     emitUnit(DWODIEBuilder, *Streamer, SplitCU);
--- a/bolt/test/X86/dwarf5-dwp-tsan-data-race.test
+++ b/bolt/test/X86/dwarf5-dwp-tsan-data-race.test
@@ -1,0 +1,26 @@
+; RUN: rm -rf %t
+; RUN: mkdir %t
+; RUN: cd %t
+; RUN: llvm-mc -dwarf-version=5 -filetype=obj -triple x86_64-unknown-linux %p/Inputs/dwarf5-df-types-dup-main.s \
+; RUN:   -split-dwarf-file=main.dwo -o main.o
+; RUN: llvm-mc -dwarf-version=5 -filetype=obj -triple x86_64-unknown-linux %p/Inputs/dwarf5-df-types-dup-helper.s \
+; RUN:   -split-dwarf-file=helper.dwo -o helper.o
+; RUN: %clang %cflags -gdwarf-5 -gsplit-dwarf=split main.o helper.o -o main.exe
+; RUN: llvm-dwp -e main.exe -o main.exe.dwp
+; RUN: llvm-bolt main.exe -o main.exe.bolt --update-debug-sections \
+; RUN:   --thread-count=4 --cu-processing-batch-size=4 --dwp=main.exe.dwp 2>&1 | FileCheck %s
+; RUN: llvm-dwarfdump --debug-info -r 0 main.dwo.dwo | FileCheck -check-prefix=BOLT-DWO-DWO-MAIN %s
+; RUN: llvm-dwarfdump --debug-info -r 0 helper.dwo.dwo | FileCheck -check-prefix=BOLT-DWO-DWO-HELPER %s
+
+;; Tests that BOLT correctly handles DWARF5 DWP file as input in a
+;; multi-threaded configuration without triggering ThreadSanitizer (TSan)
+;; data race reports. The binary must be built with TSan instrumentation
+;; (i.e. LLVM_USE_SANITIZER=Thread) for this test to be effective.
+;; REQUIRES: tsan
+
+; CHECK-NOT: ThreadSanitizer: data race
+; CHECK-NOT: ThreadSanitizer: reported
+; CHECK-NOT: WARNING: ThreadSanitizer
+
+; BOLT-DWO-DWO-MAIN: DW_TAG_compile_unit
+; BOLT-DWO-DWO-HELPER: DW_TAG_compile_unit


aaupov (commented):
If this is a race, can you reproduce it under TSan? You may need `-no-threads` controlling parallelism outside debug info processing.

Thrrreeee:
When you modify the test case `dwarf5-df-types-dup-dwp-input.test `to use multi-threaded processing （`--debug-thread-count=4 --cu-processing-batch-size=4`）, running the test with llvm-lit will produce this error.
```
| llvm-bolt: /home/jinrui/llvm-project-org/llvm/lib/DebugInfo/DWARF/DWARFUnit.cpp:433: void llvm::DWARFUnit::extractDIEsToVector(bool, bool, std::vector<llvm::DWARFDebugInfoEntry>&) const: Assertion `((AppendCUDie && Dies.empty()) || (!AppendCUDie && Dies.size() == 1)) && "Dies array is not empty"' failed.
#0 0x000055713104f378 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0xc35378)
#1 0x000055713104c45c SignalHandler(int, siginfo_t*, void*) Signals.cpp:0:0
#2 0x00007f4267792730 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x12730)
#3 0x00007f42672a78eb raise /build/reproducible-path/glibc-2.28/signal/../sysdeps/unix/sysv/linux/raise.c:51:1
#4 0x00007f4267292535 abort /build/reproducible-path/glibc-2.28/stdlib/abort.c:81:7
#5 0x00007f426729240f _nl_load_domain /build/reproducible-path/glibc-2.28/intl/loadmsgcat.c:1177:9
#6 0x00007f42672a01a2 (/lib/x86_64-linux-gnu/libc.so.6+0x301a2)
#7 0x000055713244666c llvm::DWARFUnit::extractDIEsToVector(bool, bool, std::vector<llvm::DWARFDebugInfoEntry, std::allocator<llvm::DWARFDebugInfoEntry>>&) const (.part.450) DWARFUnit.cpp:0:0
#8 0x0000557132447394 llvm::DWARFUnit::tryExtractDIEsIfNeeded(bool) (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0x202d394)
#9 0x0000557132447d4d llvm::DWARFUnit::extractDIEsIfNeeded(bool) (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0x202dd4d)
#10 0x0000557131845fd0 llvm::bolt::BinaryContext::isValidDwarfUnit(llvm::DWARFUnit&) const (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0x142bfd0)
#11 0x00005571318ca289 llvm::bolt::DIEBuilder::registerUnit(llvm::DWARFUnit&, bool) (.constprop.783) DIEBuilder.cpp:0:0
#12 0x00005571318ca556 llvm::bolt::DIEBuilder::buildTypeUnits(llvm::bolt::DebugStrOffsetsWriter*, bool) (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0x14b0556)
#13 0x00005571318cad68 llvm::bolt::DIEBuilder::buildDWOUnit(llvm::DWARFUnit&) (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0x14b0d68)
#14 0x0000557131148d11 llvm::bolt::DWARFRewriter::updateDebugInfo()::'lambda0'(llvm::DWARFUnit&, llvm::DWARFUnit&, llvm::bolt::DebugRangesSectionWriter&, llvm::bolt::DebugAddrWriter&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::optional<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>> const&, llvm::bolt::DIEBuilder&)::operator()(llvm::DWARFUnit&, llvm::DWARFUnit&, llvm::bolt::DebugRangesSectionWriter&, llvm::bolt::DebugAddrWriter&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::optional<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>> const&, llvm::bolt::DIEBuilder&) const (.isra.1352) DWARFRewriter.cpp:0:0
#15 0x000055713114998b void llvm::detail::UniqueFunctionBase<void>::CallImpl<llvm::bolt::DWARFRewriter::updateDebugInfo()::'lambda2'()>(void*) DWARFRewriter.cpp:0:0
#16 0x0000557131135168 std::_Function_handler<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> (), std::__future_base::_Task_setter<std::unique_ptr<std::__future_base::_Result<void>, std::__future_base::_Result_base::_Deleter>, std::thread::_Invoker<std::tuple<llvm::unique_function<void ()>>>, void>>::_M_invoke(std::_Any_data const&) (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0xd1b168)
#17 0x000055713113507b std::__future_base::_State_baseV2::_M_do_set(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*) (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0xd1b07b)
#18 0x00007f426778f997 __pthread_once_slow /build/reproducible-path/glibc-2.28/nptl/pthread_once.c:118:7
#19 0x0000557131134b99 std::__future_base::_Deferred_state<std::thread::_Invoker<std::tuple<llvm::unique_function<void ()>>>, void>::_M_complete_async() (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0xd1ab99)
#20 0x00005571311354bb void llvm::detail::UniqueFunctionBase<void>::CallImpl<std::shared_future<void> llvm::ThreadPoolInterface::asyncImpl<void>(llvm::unique_function<void ()>, llvm::ThreadPoolTaskGroup*)::'lambda'()>(void*) (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0xd1b4bb)
#21 0x00005571327639a2 llvm::StdThreadPool::processTasks(llvm::ThreadPoolTaskGroup*) (/home/jinrui/llvm-project-org/build/bin/llvm-bolt+0x23499a2)
#22 0x0000557132764d0f void* llvm::thread::ThreadProxy<std::tuple<llvm::StdThreadPool::grow(int)::'lambda'()>>(void*) ThreadPool.cpp:0:0
#23 0x00007f4267787fa3 start_thread /build/reproducible-path/glibc-2.28/nptl/pthread_create.c:487:7
#24 0x00007f426736906f __clone /build/reproducible-path/glibc-2.28/misc/../sysdeps/unix/sysv/linux/x86_64/clone.S:97:0
`-----------------------------
error: command failed with exit status: -6
```

Thrrreeee:
Hi @aaupov ，
Here are my partial results from TSAN (ThreadSanitizer) detection.
```
WARNING: ThreadSanitizer: data race (pid=725727)
  Read of size 8 at 0x7b3400000d10 by thread T3:
    #0 llvm::bolt::DIEBuilder::buildTypeUnits(llvm::bolt::DebugStrOffsetsWriter*, bool) /home/jinrui/llvm-pr/bolt/lib/Core/DIEBuilder.cpp:277 (llvm-bolt+0x19f5d23)
    #1 llvm::bolt::DIEBuilder::buildDWOUnit(llvm::DWARFUnit&) /home/jinrui/llvm-pr/bolt/lib/Core/DIEBuilder.cpp:361 (llvm-bolt+0x19f6dba)
    #2 operator() /home/jinrui/llvm-pr/bolt/lib/Rewrite/DWARFRewriter.cpp:658 (llvm-bolt+0xdee601)
    #3 operator() /home/jinrui/llvm-pr/bolt/lib/Rewrite/DWARFRewriter.cpp:765 (llvm-bolt+0xdf02f6)
    #4 CallImpl<llvm::bolt::DWARFRewriter::updateDebugInfo()::<lambda()> > /home/jinrui/llvm-pr/llvm/include/llvm/ADT/FunctionExtras.h:211 (llvm-bolt+0xdf02f6)
    #5 llvm::unique_function<void ()>::operator()() /home/jinrui/llvm-pr/llvm/include/llvm/ADT/FunctionExtras.h:361 (llvm-bolt+0xdcd366)
    #6 void std::__invoke_impl<void, llvm::unique_function<void ()>>(std::__invoke_other, llvm::unique_function<void ()>&&) /usr/include/c++/8/bits/invoke.h:60 (llvm-bolt+0xdcd366)
    #7 std::__invoke_result<llvm::unique_function<void ()>>::type std::__invoke<llvm::unique_function<void ()>>(llvm::unique_function<void ()>&&) /usr/include/c++/8/bits/invoke.h:95 (llvm-bolt+0xdcd366)
    #8 decltype (__invoke((_S_declval<0ul>)())) std::thread::_Invoker<std::tuple<llvm::unique_function<void ()> > >::_M_invoke<0ul>(std::_Index_tuple<0ul>) /usr/include/c++/8/thread:244 (llvm-bolt+0xdcd366)
    #9 std::thread::_Invoker<std::tuple<llvm::unique_function<void ()> > >::operator()() /usr/include/c++/8/thread:253 (llvm-bolt+0xdcd366)
    #10 std::__future_base::_Task_setter<std::unique_ptr<std::__future_base::_Result<void>, std::__future_base::_Result_base::_Deleter>, std::thread::_Invoker<std::tuple<llvm::unique_function<void ()> > >, void>::operator()() const /usr/include/c++/8/future:1362 (llvm-bolt+0xdcd366)
    #11 std::_Function_handler<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> (), std::__future_base::_Task_setter<std::unique_ptr<std::__future_base::_Result<void>, std::__future_base::_Result_base::_Deleter>, std::thread::_Invoker<std::tuple<llvm::unique_function<void ()> > >, void> >::_M_invoke(std::_Any_data const&) /usr/include/c++/8/bits/std_function.h:283 (llvm-bolt+0xdcd366)
    #12 std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>::operator()() const /usr/include/c++/8/bits/std_function.h:687 (llvm-bolt+0xdcd5bc)
    #13 std::__future_base::_State_baseV2::_M_do_set(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*) /usr/include/c++/8/future:561 (llvm-bolt+0xdcd5bc)
    #14 void std::__invoke_impl<void, void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::__invoke_memfun_deref, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&) /usr/include/c++/8/bits/invoke.h:73 (llvm-bolt+0xdcd554)
    #15 std::__invoke_result<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>::type std::__invoke<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&) /usr/include/c++/8/bits/invoke.h:95 (llvm-bolt+0xdcd554)
    #16 std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&)::{lambda()#1}::operator()() const /usr/include/c++/8/mutex:672 (llvm-bolt+0xdcd554)
    #17 std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&)::{lambda()#2}::operator()() const /usr/include/c++/8/mutex:677 (llvm-bolt+0xdcd554)
    #18 std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&)::{lambda()#2}::_FUN() /usr/include/c++/8/mutex:677 (llvm-bolt+0xdcd554)
    #19 pthread_once <null> (libtsan.so.0+0x2d8d9)
    #20 __gthread_once /usr/include/x86_64-linux-gnu/c++/8/bits/gthr-default.h:699 (llvm-bolt+0xdcc9d2)
    #21 void std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&) /usr/include/c++/8/mutex:684 (llvm-bolt+0xdcc9d2)
    #22 std::__future_base::_State_baseV2::_M_set_result(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>, bool) /usr/include/c++/8/future:401 (llvm-bolt+0xdcc9d2)
    #23 std::__future_base::_Deferred_state<std::thread::_Invoker<std::tuple<llvm::unique_function<void ()> > >, void>::_M_complete_async() /usr/include/c++/8/future:1615 (llvm-bolt+0xdcc9d2)
    #24 std::__future_base::_State_baseV2::wait() /usr/include/c++/8/future:334 (llvm-bolt+0xdcdcc0)
    #25 std::__basic_future<void>::wait() const /usr/include/c++/8/future:692 (llvm-bolt+0xdcdcc0)
    #26 llvm::ThreadPoolInterface::asyncImpl<void>(llvm::unique_function<void ()>, llvm::ThreadPoolTaskGroup*)::{lambda()#1}::operator()() const /home/jinrui/llvm-pr/llvm/include/llvm/Support/ThreadPool.h:116 (llvm-bolt+0xdcdcc0)
    #27 void llvm::detail::UniqueFunctionBase<void>::CallImpl<llvm::ThreadPoolInterface::asyncImpl<void>(llvm::unique_function<void ()>, llvm::ThreadPoolTaskGroup*)::{lambda()#1}>(void*) /home/jinrui/llvm-pr/llvm/include/llvm/ADT/FunctionExtras.h:211 (llvm-bolt+0xdcdcc0)
    #28 llvm::unique_function<void ()>::operator()() /home/jinrui/llvm-pr/llvm/include/llvm/ADT/FunctionExtras.h:361 (llvm-bolt+0x30e89fa)
    #29 llvm::StdThreadPool::processTasks(llvm::ThreadPoolTaskGroup*) /home/jinrui/llvm-pr/llvm/lib/Support/ThreadPool.cpp:119 (llvm-bolt+0x30e89fa)
    #30 operator() /home/jinrui/llvm-pr/llvm/lib/Support/ThreadPool.cpp:62 (llvm-bolt+0x30ea239)
    #31 operator()<llvm::StdThreadPool::grow(int)::<lambda()>&> /home/jinrui/llvm-pr/llvm/include/llvm/Support/thread.h:46 (llvm-bolt+0x30ea239)
    #32 __invoke_impl<void, llvm::thread::GenericThreadProxy(void*) [with CalleeTuple = std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >]::<lambda(auto:4&&, auto:5&& ...)>, llvm::StdThreadPool::grow(int)::<lambda()>&> /usr/include/c++/8/bits/invoke.h:60 (llvm-bolt+0x30ea239)
    #33 __invoke<llvm::thread::GenericThreadProxy(void*) [with CalleeTuple = std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >]::<lambda(auto:4&&, auto:5&& ...)>, llvm::StdThreadPool::grow(int)::<lambda()>&> /usr/include/c++/8/bits/invoke.h:95 (llvm-bolt+0x30ea239)
    #34 __apply_impl<llvm::thread::GenericThreadProxy(void*) [with CalleeTuple = std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >]::<lambda(auto:4&&, auto:5&& ...)>, std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >&, 0> /usr/include/c++/8/tuple:1678 (llvm-bolt+0x30ea239)
    #35 apply<llvm::thread::GenericThreadProxy(void*) [with CalleeTuple = std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >]::<lambda(auto:4&&, auto:5&& ...)>, std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >&> /usr/include/c++/8/tuple:1687 (llvm-bolt+0x30ea239)
    #36 GenericThreadProxy<std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> > > /home/jinrui/llvm-pr/llvm/include/llvm/Support/thread.h:44 (llvm-bolt+0x30ea239)
    #37 ThreadProxy<std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> > > /home/jinrui/llvm-pr/llvm/include/llvm/Support/thread.h:62 (llvm-bolt+0x30ea239)
    #38 <null> <null> (libtsan.so.0+0x29b3d)

  Previous write of size 8 at 0x7b3400000d10 by thread T10:
    [failed to restore the stack]

  Location is heap block of size 200 at 0x7b3400000d00 allocated by thread T10:
    #0 operator new[](unsigned long) <null> (libtsan.so.0+0x6f94a)
    #1 std::_MakeUniq<llvm::DWARFUnitIndex::Entry []>::__array std::make_unique<llvm::DWARFUnitIndex::Entry []>(unsigned long) /usr/include/c++/8/bits/unique_ptr.h:837 (llvm-bolt+0x2c08570)
    #2 llvm::DWARFUnitIndex::parseImpl(llvm::DataExtractor) /home/jinrui/llvm-pr/llvm/lib/DebugInfo/DWARF/DWARFUnitIndex.cpp:147 (llvm-bolt+0x2c08570)
    #3 llvm::DWARFUnitIndex::parse(llvm::DataExtractor) /home/jinrui/llvm-pr/llvm/lib/DebugInfo/DWARF/DWARFUnitIndex.cpp:122 (llvm-bolt+0x2c08ceb)
    #4 getTUIndex /home/jinrui/llvm-pr/llvm/lib/DebugInfo/DWARF/DWARFContext.cpp:344 (llvm-bolt+0x2b7e0b8)
    #5 llvm::DWARFContext::getTUIndex() /home/jinrui/llvm-pr/llvm/lib/DebugInfo/DWARF/DWARFContext.cpp:1427 (llvm-bolt+0x2b7ea7c)
    #6 llvm::bolt::DIEBuilder::buildTypeUnits(llvm::bolt::DebugStrOffsetsWriter*, bool) /home/jinrui/llvm-pr/bolt/lib/Core/DIEBuilder.cpp:274 (llvm-bolt+0x19f5964)
    #7 llvm::bolt::DIEBuilder::buildDWOUnit(llvm::DWARFUnit&) /home/jinrui/llvm-pr/bolt/lib/Core/DIEBuilder.cpp:361 (llvm-bolt+0x19f6dba)
    #8 operator() /home/jinrui/llvm-pr/bolt/lib/Rewrite/DWARFRewriter.cpp:658 (llvm-bolt+0xdee601)
    #9 operator() /home/jinrui/llvm-pr/bolt/lib/Rewrite/DWARFRewriter.cpp:765 (llvm-bolt+0xdf02f6)
    #10 CallImpl<llvm::bolt::DWARFRewriter::updateDebugInfo()::<lambda()> > /home/jinrui/llvm-pr/llvm/include/llvm/ADT/FunctionExtras.h:211 (llvm-bolt+0xdf02f6)
    #11 llvm::unique_function<void ()>::operator()() /home/jinrui/llvm-pr/llvm/include/llvm/ADT/FunctionExtras.h:361 (llvm-bolt+0xdcd366)
    #12 void std::__invoke_impl<void, llvm::unique_function<void ()>>(std::__invoke_other, llvm::unique_function<void ()>&&) /usr/include/c++/8/bits/invoke.h:60 (llvm-bolt+0xdcd366)
    #13 std::__invoke_result<llvm::unique_function<void ()>>::type std::__invoke<llvm::unique_function<void ()>>(llvm::unique_function<void ()>&&) /usr/include/c++/8/bits/invoke.h:95 (llvm-bolt+0xdcd366)
    #14 decltype (__invoke((_S_declval<0ul>)())) std::thread::_Invoker<std::tuple<llvm::unique_function<void ()> > >::_M_invoke<0ul>(std::_Index_tuple<0ul>) /usr/include/c++/8/thread:244 (llvm-bolt+0xdcd366)
    #15 std::thread::_Invoker<std::tuple<llvm::unique_function<void ()> > >::operator()() /usr/include/c++/8/thread:253 (llvm-bolt+0xdcd366)
    #16 std::__future_base::_Task_setter<std::unique_ptr<std::__future_base::_Result<void>, std::__future_base::_Result_base::_Deleter>, std::thread::_Invoker<std::tuple<llvm::unique_function<void ()> > >, void>::operator()() const /usr/include/c++/8/future:1362 (llvm-bolt+0xdcd366)
    #17 std::_Function_handler<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> (), std::__future_base::_Task_setter<std::unique_ptr<std::__future_base::_Result<void>, std::__future_base::_Result_base::_Deleter>, std::thread::_Invoker<std::tuple<llvm::unique_function<void ()> > >, void> >::_M_invoke(std::_Any_data const&) /usr/include/c++/8/bits/std_function.h:283 (llvm-bolt+0xdcd366)
    #18 std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>::operator()() const /usr/include/c++/8/bits/std_function.h:687 (llvm-bolt+0xdcd5bc)
    #19 std::__future_base::_State_baseV2::_M_do_set(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*) /usr/include/c++/8/future:561 (llvm-bolt+0xdcd5bc)
    #20 void std::__invoke_impl<void, void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::__invoke_memfun_deref, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&) /usr/include/c++/8/bits/invoke.h:73 (llvm-bolt+0xdcd554)
    #21 std::__invoke_result<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>::type std::__invoke<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&) /usr/include/c++/8/bits/invoke.h:95 (llvm-bolt+0xdcd554)
    #22 std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&)::{lambda()#1}::operator()() const /usr/include/c++/8/mutex:672 (llvm-bolt+0xdcd554)
    #23 std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&)::{lambda()#2}::operator()() const /usr/include/c++/8/mutex:677 (llvm-bolt+0xdcd554)
    #24 std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&)::{lambda()#2}::_FUN() /usr/include/c++/8/mutex:677 (llvm-bolt+0xdcd554)
    #25 pthread_once <null> (libtsan.so.0+0x2d8d9)
    #26 __gthread_once /usr/include/x86_64-linux-gnu/c++/8/bits/gthr-default.h:699 (llvm-bolt+0xdcc9d2)
    #27 void std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&) /usr/include/c++/8/mutex:684 (llvm-bolt+0xdcc9d2)
    #28 std::__future_base::_State_baseV2::_M_set_result(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>, bool) /usr/include/c++/8/future:401 (llvm-bolt+0xdcc9d2)
    #29 std::__future_base::_Deferred_state<std::thread::_Invoker<std::tuple<llvm::unique_function<void ()> > >, void>::_M_complete_async() /usr/include/c++/8/future:1615 (llvm-bolt+0xdcc9d2)
    #30 std::__future_base::_State_baseV2::wait() /usr/include/c++/8/future:334 (llvm-bolt+0xdcdcc0)
    #31 std::__basic_future<void>::wait() const /usr/include/c++/8/future:692 (llvm-bolt+0xdcdcc0)
    #32 llvm::ThreadPoolInterface::asyncImpl<void>(llvm::unique_function<void ()>, llvm::ThreadPoolTaskGroup*)::{lambda()#1}::operator()() const /home/jinrui/llvm-pr/llvm/include/llvm/Support/ThreadPool.h:116 (llvm-bolt+0xdcdcc0)
    #33 void llvm::detail::UniqueFunctionBase<void>::CallImpl<llvm::ThreadPoolInterface::asyncImpl<void>(llvm::unique_function<void ()>, llvm::ThreadPoolTaskGroup*)::{lambda()#1}>(void*) /home/jinrui/llvm-pr/llvm/include/llvm/ADT/FunctionExtras.h:211 (llvm-bolt+0xdcdcc0)
    #34 llvm::unique_function<void ()>::operator()() /home/jinrui/llvm-pr/llvm/include/llvm/ADT/FunctionExtras.h:361 (llvm-bolt+0x30e89fa)
    #35 llvm::StdThreadPool::processTasks(llvm::ThreadPoolTaskGroup*) /home/jinrui/llvm-pr/llvm/lib/Support/ThreadPool.cpp:119 (llvm-bolt+0x30e89fa)
    #36 operator() /home/jinrui/llvm-pr/llvm/lib/Support/ThreadPool.cpp:62 (llvm-bolt+0x30ea239)
    #37 operator()<llvm::StdThreadPool::grow(int)::<lambda()>&> /home/jinrui/llvm-pr/llvm/include/llvm/Support/thread.h:46 (llvm-bolt+0x30ea239)
    #38 __invoke_impl<void, llvm::thread::GenericThreadProxy(void*) [with CalleeTuple = std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >]::<lambda(auto:4&&, auto:5&& ...)>, llvm::StdThreadPool::grow(int)::<lambda()>&> /usr/include/c++/8/bits/invoke.h:60 (llvm-bolt+0x30ea239)
    #39 __invoke<llvm::thread::GenericThreadProxy(void*) [with CalleeTuple = std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >]::<lambda(auto:4&&, auto:5&& ...)>, llvm::StdThreadPool::grow(int)::<lambda()>&> /usr/include/c++/8/bits/invoke.h:95 (llvm-bolt+0x30ea239)
    #40 __apply_impl<llvm::thread::GenericThreadProxy(void*) [with CalleeTuple = std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >]::<lambda(auto:4&&, auto:5&& ...)>, std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >&, 0> /usr/include/c++/8/tuple:1678 (llvm-bolt+0x30ea239)
    #41 apply<llvm::thread::GenericThreadProxy(void*) [with CalleeTuple = std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >]::<lambda(auto:4&&, auto:5&& ...)>, std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> >&> /usr/include/c++/8/tuple:1687 (llvm-bolt+0x30ea239)
    #42 GenericThreadProxy<std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> > > /home/jinrui/llvm-pr/llvm/include/llvm/Support/thread.h:44 (llvm-bolt+0x30ea239)
    #43 ThreadProxy<std::tuple<llvm::StdThreadPool::grow(int)::<lambda()> > > /home/jinrui/llvm-pr/llvm/include/llvm/Support/thread.h:62 (llvm-bolt+0x30ea239)
    #44 <null> <null> (libtsan.so.0+0x29b3d)

  Thread T3 'llvm-worker-2' (tid=725731, running) created by main thread at:
    #0 pthread_create <null> (libtsan.so.0+0x2be1b)
    #1 llvm::llvm_execute_on_thread_impl(void* (*)(void*), void*, std::optional<unsigned int>) /home/jinrui/llvm-pr/llvm/lib/Support/Unix/Threading.inc:95 (llvm-bolt+0xc4a2e3)
    #2 thread<llvm::StdThreadPool::grow(int)::<lambda()>&> /home/jinrui/llvm-pr/llvm/include/llvm/Support/thread.h:138 (llvm-bolt+0x30e390b)
    #3 emplace_back<llvm::StdThreadPool::grow(int)::<lambda()> > /home/jinrui/llvm-pr/llvm/include/llvm/Support/thread.h:85 (llvm-bolt+0x30e41fe)
    #4 llvm::StdThreadPool::grow(int) /home/jinrui/llvm-pr/llvm/lib/Support/ThreadPool.cpp:63 (llvm-bolt+0x30e41fe)
    #5 llvm::StdThreadPool::asyncEnqueue(llvm::unique_function<void ()>, llvm::ThreadPoolTaskGroup*) /home/jinrui/llvm-pr/llvm/include/llvm/Support/ThreadPool.h:176 (llvm-bolt+0x30e4d21)
    #6 async<llvm::bolt::ParallelUtilities::runOnEachFunctionWithUniqueAllocId(llvm::bolt::BinaryContext&, llvm::bolt::ParallelUtilities::SchedulingPolicy, llvm::bolt::ParallelUtilities::WorkFuncWithAllocTy, llvm::bolt::ParallelUtilities::PredicateTy, std::__cxx11::string, bool, unsigned int)::<lambda(std::map<long unsigned int, llvm::bolt::BinaryFunction>::iterator, std::map<long unsigned int, llvm::bolt::BinaryFunction>::iterator, llvm::bolt::MCPlusBuilder::AllocatorIdTy)>&, std::_Rb_tree_iterator<std::pair<long unsigned int const, llvm::bolt::BinaryFunction> >&, std::_Rb_tree_iterator<std::pair<long unsigned int const, llvm::bolt::BinaryFunction> >, unsigned int&> /home/jinrui/llvm-pr/llvm/include/llvm/Support/ThreadPool.h:116 (llvm-bolt+0x1a2d71e)
    #7 llvm::bolt::ParallelUtilities::runOnEachFunctionWithUniqueAllocId(llvm::bolt::BinaryContext&, llvm::bolt::ParallelUtilities::SchedulingPolicy, std::function<void (llvm::bolt::BinaryFunction&, unsigned short)>, std::function<bool (llvm::bolt::BinaryFunction const&)>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, unsigned int) /home/jinrui/llvm-pr/bolt/lib/Core/ParallelUtilities.cpp:230 (llvm-bolt+0x1a2d71e)
    #8 llvm::bolt::RewriteInstance::buildFunctionsCFG() /home/jinrui/llvm-pr/bolt/lib/Rewrite/RewriteInstance.cpp:3833 (llvm-bolt+0xd341f2)
    #9 llvm::bolt::RewriteInstance::run() /home/jinrui/llvm-pr/bolt/lib/Rewrite/RewriteInstance.cpp:803 (llvm-bolt+0xdbb513)
    #10 main /home/jinrui/llvm-pr/bolt/tools/driver/llvm-bolt.cpp:267 (llvm-bolt+0xc5cd0)

  Thread T10 'llvm-worker-9' (tid=725738, running) created by main thread at:
    #0 pthread_create <null> (libtsan.so.0+0x2be1b)
    #1 llvm::llvm_execute_on_thread_impl(void* (*)(void*), void*, std::optional<unsigned int>) /home/jinrui/llvm-pr/llvm/lib/Support/Unix/Threading.inc:95 (llvm-bolt+0xc4a2e3)
    #2 thread<llvm::StdThreadPool::grow(int)::<lambda()>&> /home/jinrui/llvm-pr/llvm/include/llvm/Support/thread.h:138 (llvm-bolt+0x30e390b)
    #3 thread<llvm::StdThreadPool::grow(int)::<lambda()> > /home/jinrui/llvm-pr/llvm/include/llvm/Support/thread.h:85 (llvm-bolt+0x30e4123)
    #4 construct<llvm::thread, llvm::StdThreadPool::grow(int)::<lambda()> > /usr/include/c++/8/ext/new_allocator.h:136 (llvm-bolt+0x30e4123)
    #5 construct<llvm::thread, llvm::StdThreadPool::grow(int)::<lambda()> > /usr/include/c++/8/bits/alloc_traits.h:475 (llvm-bolt+0x30e4123)
    #6 emplace_back<llvm::StdThreadPool::grow(int)::<lambda()> > /usr/include/c++/8/bits/vector.tcc:103 (llvm-bolt+0x30e4123)
    #7 llvm::StdThreadPool::grow(int) /home/jinrui/llvm-pr/llvm/lib/Support/ThreadPool.cpp:63 (llvm-bolt+0x30e4123)
    #8 llvm::StdThreadPool::asyncEnqueue(llvm::unique_function<void ()>, llvm::ThreadPoolTaskGroup*) /home/jinrui/llvm-pr/llvm/include/llvm/Support/ThreadPool.h:176 (llvm-bolt+0x30e4d21)
    #9 async<llvm::bolt::ParallelUtilities::runOnEachFunctionWithUniqueAllocId(llvm::bolt::BinaryContext&, llvm::bolt::ParallelUtilities::SchedulingPolicy, llvm::bolt::ParallelUtilities::WorkFuncWithAllocTy, llvm::bolt::ParallelUtilities::PredicateTy, std::__cxx11::string, bool, unsigned int)::<lambda(std::map<long unsigned int, llvm::bolt::BinaryFunction>::iterator, std::map<long unsigned int, llvm::bolt::BinaryFunction>::iterator, llvm::bolt::MCPlusBuilder::AllocatorIdTy)>&, std::_Rb_tree_iterator<std::pair<long unsigned int const, llvm::bolt::BinaryFunction> >&, std::_Rb_tree_iterator<std::pair<long unsigned int const, llvm::bolt::BinaryFunction> >, unsigned int&> /home/jinrui/llvm-pr/llvm/include/llvm/Support/ThreadPool.h:116 (llvm-bolt+0x1a2d71e)
    #10 llvm::bolt::ParallelUtilities::runOnEachFunctionWithUniqueAllocId(llvm::bolt::BinaryContext&, llvm::bolt::ParallelUtilities::SchedulingPolicy, std::function<void (llvm::bolt::BinaryFunction&, unsigned short)>, std::function<bool (llvm::bolt::BinaryFunction const&)>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, unsigned int) /home/jinrui/llvm-pr/bolt/lib/Core/ParallelUtilities.cpp:230 (llvm-bolt+0x1a2d71e)
    #11 llvm::bolt::RewriteInstance::buildFunctionsCFG() /home/jinrui/llvm-pr/bolt/lib/Rewrite/RewriteInstance.cpp:3833 (llvm-bolt+0xd341f2)
    #12 llvm::bolt::RewriteInstance::run() /home/jinrui/llvm-pr/bolt/lib/Rewrite/RewriteInstance.cpp:803 (llvm-bolt+0xdbb513)
    #13 main /home/jinrui/llvm-pr/bolt/tools/driver/llvm-bolt.cpp:267 (llvm-bolt+0xc5cd0)

SUMMARY: ThreadSanitizer: data race /home/jinrui/llvm-pr/bolt/lib/Core/DIEBuilder.cpp:277 in llvm::bolt::DIEBuilder::buildTypeUnits(llvm::bolt::DebugStrOffsetsWriter*, bool)
```

ayermolo:
Ah when DWP is input. Let me remember how that code works today or later this week.

Thrrreeee:
Hi, @ayermolo 
Let me know if there's anything I can do to help speed things up — happy to provide more context or test results. Thanks!

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -268,6 +268,11 @@ static unsigned int getCUNum(DWARFContext *DwarfContext, bool IsDWO) {
 
 void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
                                 const bool Init) {
+  static std::mutex DWOTypeUnitsBuildMutex;
```
Jinjie-Huang: Using local static variables is generally discouraged by LLVM (see: https://llvm.org/docs/CodingStandards.html#do-not-use-static-constructors), as static constructors can slow down startup times. 

It would likely be better to make the lock a class member here.

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -268,6 +268,11 @@ static unsigned int getCUNum(DWARFContext *DwarfContext, bool IsDWO) {
...
+  std::unique_lock<std::mutex> DWOTypeUnitsBuildLock(DWOTypeUnitsBuildMutex,
+                                                     std::defer_lock);
+  if (isDWO())
```
Jinjie-Huang: This race condition should primarily occur in DWP scenarios since the DIEBuilder is shared. Applying a lock to all split-dwarf cases like this might negatively impact performance when processing regular DWO inputs.

Could we introduce a flag to identify whether the input is a DWP, and only acquire the lock if that is the case?

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -268,6 +268,11 @@ static unsigned int getCUNum(DWARFContext *DwarfContext, bool IsDWO) {
...
+    DWOTypeUnitsBuildLock.lock();
```
Jinjie-Huang: Locking the entire function here seems a bit too coarse-grained. Since the data race primarily occurs within getTypeUnitForHash and registerUnit, could we narrow down the scope of the lock?

And the direct function triggering the issue is likely extractDIEsIfNeeded, would it be possible to create a wrapper function for it and handle the locking exclusively within that wrapper?

bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
```diff
@@ -7,7 +7,7 @@
 ; RUN: -split-dwarf-file=helper.dwo -o helper.o
 ; RUN: %clang %cflags -gdwarf-5 -gsplit-dwarf=split main.o helper.o -o main.exe
 ; RUN: llvm-dwp -e main.exe -o main.exe.dwp
-; RUN: llvm-bolt main.exe -o main.exe.bolt --update-debug-sections
+; RUN: llvm-bolt main.exe -o main.exe.bolt --update-debug-sections  --debug-thread-count=4 --cu-processing-batch-size=4
```
ayermolo: Can we run this test under tsan? Otherwise this is just a non deterministic test that sometimes fails.

ayermolo (commented):
My apologies for late review, been busy. :(
Umm, can't we just use thread save DWARFContext?

set it to true here:
bolt/lib/Rewrite/RewriteInstance.cpp
```
auto BCOrErr = BinaryContext::createBinaryContext(
      TheTriple, std::make_shared<orc::SymbolStringPool>(), File->getFileName(),
      Features.get(), IsPIC,
      DWARFContext::create(*File, DWARFContext::ProcessDebugRelocations::Ignore,
                           nullptr, opts::DWPPathName,
                           WithColor::defaultErrorHandler,
                           WithColor::defaultWarningHandler),
      JournalingStreams{Stdout, Stderr});
```

ayermolo:
@Jinjie-Huang Thank you for stepping in. :)

bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
Thrrreeee: The TSan results mentioned above were obtained by running this test case.

Thrrreeee:
> My apologies for late review, been busy. :( Umm, can't we just use thread save DWARFContext?
> 
I previously tried using a thread-safe DWARFContext, but it does not solve this problem.. The races occurred on lazily-initialized members of the same DWARFUnit object — specifically the DieArray, Abbrev, and other member variables.

ayermolo:
> BinaryContext::isValidDwarfUnit

OH I see. Thanks for checking. Too many ways to do things in dwarf library. lol
My memory a bit hazy on this code, but the original idea and what it was tested with is .dwo as input. 
So each .dwo had their own TUs, and dwarf context. So could be parsed in parallel.
In DWP we have all the CUs and TUs in one file.
We still create build units to process a chunk of CUs, but then each chunk keeps reading and re-reading the same TUs, and also causing this issue.
Even if this works the .dwo will end up with all the TUs.
Unless I am missing something, maybe a more structural change is needed?
If input is DWP we read TUs once, when we process CUs we don't parse TUs, and keep track of which TU hash map is used.
When we write out .dwo files we also emit TUs that are only used in the dwo file CU.

I thought there was some work done in that area last year...

bolt/include/bolt/Core/BinaryContext.h
```diff
@@ -284,12 +285,16 @@ class BinaryContext {
...
 
+  mutable std::recursive_mutex DWPUnitsMutex;
+
 public:
   static Expected<std::unique_ptr<BinaryContext>> createBinaryContext(
       Triple TheTriple, std::shared_ptr<orc::SymbolStringPool> SSP,
       StringRef InputFileName, SubtargetFeatures *Features, bool IsPIC,
       std::unique_ptr<DWARFContext> DwCtx, JournalingStreams Logger);
 
+  std::recursive_mutex &getUnitsMutex() { return DWPUnitsMutex; }
```
ayermolo: Please add comment

bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
ayermolo: Can we change it so that parallel version only runs under TSAN? Without it it's just a potentially flaky test.

Thrrreeee:
> We still create build units to process a chunk of CUs, but then each chunk keeps reading and re-reading the same TUs, and also causing this issue. Even if this works the .dwo will end up with all the TUs. Unless I am missing something, maybe a more structural change is needed? 

Thanks for the review!  You're right, this approach could indeed lead to duplicate TUs in the final output. I'll take some time to think further about how to properly address this.

bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
Thrrreeee: That is OK, I will try

Thrrreeee:
cc @ayermolo

bolt/test/X86/dwarf5-df-types-dup-dwp-input.test
ayermolo: Can you revert changes in this file?

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -268,6 +268,22 @@ static unsigned int getCUNum(DWARFContext *DwarfContext, bool IsDWO) {
 
 void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
                                 const bool Init) {
+  if (isDWO() && DwarfContext && DwarfContext->isDWP()) {
```
ayermolo: I don't quite remember can DwarfContext be null?
Also with isDWP check I don't think isDWO necessary, as isDWP implies isDWO, and this shouldn't be an issue with jsut dwo as we are creating one builder per dwo file.

ayermolo (commented):
I don't think new approach solves the problem of re-parsing types multiple times in DWP.

bolt/lib/Core/DIEBuilder.cpp
Thrrreeee: > I don't quite remember can DwarfContext be null? Also with isDWP check I don't think isDWO necessary, as isDWP implies isDWO, and this shouldn't be an issue with jsut dwo as we are creating one builder per dwo file.

fix

Thrrreeee:
> I don't think new approach solves the problem of re-parsing types multiple times in DWP.

I considered how to avoid re-parsing types for DWP, but each DWO has its own DIEBuilder and needs to retrieve the type through buildTypeUnits(). Parsing it only once would still require copying the result into every per-DWODIEBuilder. since BOLT emits the output in DWO format even when the input is a DWP — so each DWO must own its own type unit.Copying the result would introduce a fair amount of additional work — it would require deep-copying several member variables inside DIEBuilder. That said, if you think it's necessary, I'm happy to give it a try.

Happy to hear if you see a better way to structure this.

ayermolo:
> > I don't think new approach solves the problem of re-parsing types multiple times in DWP.
> 
> I considered how to avoid re-parsing types for DWP, but each DWO has its own DIEBuilder and needs to retrieve the type through buildTypeUnits(). Parsing it only once would still require copying the result into every per-DWODIEBuilder. since BOLT emits the output in DWO format even when the input is a DWP — so each DWO must own its own type unit.Copying the result would introduce a fair amount of additional work — it would require deep-copying several member variables inside DIEBuilder. That said, if you think it's necessary, I'm happy to give it a try.
> 
> Happy to hear if you see a better way to structure this.

Right will need to parse it once, then copy into each dwo. From design perspective open question whether to copy on the IR level, or write out each type unit in to elf and just copy chunks, since each TU is self contained.

Otherwise we have two issues. One re-parsing same type units for each CU. Then each DWO will end up ALL the TUs. So size of DWOs will explode. Although for second part as I mentioned I think something was done for it? I vaguely remember someone putting up PR about it. Haven't had a chance to dig into BOLT code to double check.

Thrrreeee:
>  I vaguely remember someone putting up PR about it. Haven't had a chance to dig into BOLT code to double check.

The earlier `.debug_str` size reduction for DWP was done by our team last year, and we could look into extending the same approach to TU reduction in the future. Since this issue may block https://github.com/llvm/llvm-project/pull/197859, could we land this bug fix first and address the rest in a follow-up PR?

 If you have other ideas in mind, I'd be happy to hear them.

Thrrreeee:
cc @ayermolo,
New implement worked on fixing the issue for each DWO ends up with all the TUs.

> About re-parsing the same type units for each CU

 Now, each DWO just constructs its own TUs individually, the same way it works for normal DWO files. Solving the "re-parsing same type units for each CU" problem by cloning/copying would introduce a lot of extra implementation work. I think that is not necessary here.

If you think what I've done diverges from the PR title, I can split it into two separate PRs for submission?

rafaelauler (approved):
Looks good on my end (if Alexander has any further suggestions, let's wait for him).

bolt/include/bolt/Core/BinaryContext.h
```diff
@@ -284,12 +285,18 @@ class BinaryContext {
   /// Internal helper for removing section name from a lookup table.
   void deregisterSectionName(const BinarySection &Section);
 
+  /// Mutex used for parallel processing of DWP type units.
+  mutable std::recursive_mutex DWPUnitsMutex;
```
ayermolo: I think we can just user regular std::mutex here since it's not used in recursive function and it's not const.

ayermolo:
Took a quick look, I'll do a more thorough review Monday.

bolt/include/bolt/Core/BinaryContext.h
Thrrreeee: > I think we can just user regular std::mutex here since it's not used in recursive function and it's not const.

Fixed!

bolt/lib/Rewrite/RewriteInstance.cpp
```diff
@@ -444,7 +444,7 @@ RewriteInstance::RewriteInstance(ELFObjectFileBase *File, const int Argc,
       DWARFContext::create(*File, DWARFContext::ProcessDebugRelocations::Ignore,
                            nullptr, opts::DWPPathName,
                            WithColor::defaultErrorHandler,
-                           WithColor::defaultWarningHandler),
+                           WithColor::defaultWarningHandler, true),
```
ayermolo, line 447: Should this be part of this PR or [a different one?](https://github.com/llvm/llvm-project/pull/197859)? I thought you mentioned enabling safe context doesn't help in this particular case.

bolt/include/bolt/Core/BinaryContext.h
```diff
@@ -284,12 +285,18 @@ class BinaryContext {
   /// Internal helper for removing section name from a lookup table.
   void deregisterSectionName(const BinarySection &Section);
 
+  /// Mutex used for parallel processing of DWP type units.
+  mutable std::mutex DWPUnitsMutex;
```
ayermolo: I don't think mutable is necessary here.

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -283,6 +283,73 @@ void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
...
+
+    const std::optional<uint32_t> UnitId = getUnitId(*TU);
+    if (!UnitId || getState().CloneUnitCtxMap[*UnitId].IsConstructed)
+      continue;
+
+    collectReferencedTypeSignatures(TU->getUnitDIE(), ProcessedTU, TUlist);
+  }
+
+  // Ensure deterministic order of processing type units
+  auto SortByOffset = [](const DWARFUnit *A, const DWARFUnit *B) {
```
ayermolo, line 343: Isn't this already enforced by the structure of DWARF? Even if we process each CU in parallel it's internal structure is deterministic?

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -328,7 +395,12 @@ void DIEBuilder::buildCompileUnits(const std::vector<DWARFUnit *> &CUs) {
 void DIEBuilder::buildDWOUnit(DWARFUnit &U) {
   BuilderState.release();
   BuilderState = std::make_unique<State>();
-  buildTypeUnits(nullptr, false);
+  if (DwarfContext->isDWP()) {
+    std::unique_lock<std::mutex> LockGuard(BC.getUnitsMutex());
```
ayermolo: Please move into callee.
Also since DWP context is shared, we are affectively serializing this whole thing due to type parsing. I haven't had a chance to look at https://github.com/llvm/llvm-project/pull/197859, so maybe it's improved there?

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -283,6 +283,73 @@ void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
...
+  }
+
+  // Ensure deterministic order of processing type units
+  auto SortByOffset = [](const DWARFUnit *A, const DWARFUnit *B) {
+    return A->getOffset() < B->getOffset();
+  };
+  llvm::sort(getState().DWARF4TUVector, SortByOffset);
+  llvm::sort(getState().DWARF5TUVector, SortByOffset);
+
+  getState().Type = ProcessingType::DWARF4TUs;
```
ayermolo: Do we have dwarf4 DWP tests with types? I don't remember.
If we don't can you add some.

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -283,6 +283,73 @@ void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
   }
 }
 
+static void collectReferencedTypeSignatures(DWARFDie Die,
```
ayermolo, line 291: Can you add a comment what this function does, and mention that de-duplication happens on outer level with. registerUnit.

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -283,6 +283,73 @@ void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
...
+      continue;
+
+    collectReferencedTypeSignatures(TU->getUnitDIE(), ProcessedTU, TUlist);
+  }
+
+  // Ensure deterministic order of processing type units
+  auto SortByOffset = [](const DWARFUnit *A, const DWARFUnit *B) {
+    return A->getOffset() < B->getOffset();
+  };
+  llvm::sort(getState().DWARF4TUVector, SortByOffset);
```
ayermolo: For Spilt DWARF we either have DWARF4 or DWARF5. They can't mix. So can simplify this slightly.
Something like:

```
void DIEBuilder::buildDWPTypeUnitsForUnit(DWARFUnit &U) {
  // A DWP is single-version: every type unit in it is either DWARF<=4 or
  // DWARF5, never both. Decide the processing type once.
  const ProcessingType TUType = U.getVersion() < 5 ? ProcessingType::DWARF4TUs
                                                    : ProcessingType::DWARF5TUs;
  getState().Type = TUType;

  DenseSet<uint64_t> ProcessedTU;
  SmallVector<uint64_t, 8> TUlist;
  collectReferencedTypeSignatures(U.getUnitDIE(), ProcessedTU, TUlist);

  for (unsigned I = 0; I != TUlist.size(); ++I) {
    DWARFTypeUnit *TU = DwarfContext->getTypeUnitForHash(TUlist[I], true);
    if (!TU)
      continue;
    assert((TU->getVersion() < 5) == (TUType == ProcessingType::DWARF4TUs) &&
           "Mixed DWARF versions in a single DWP");
    if (!registerUnit(*TU, false))
      continue;
    const std::optional<uint32_t> UnitId = getUnitId(*TU);
    if (!UnitId || getState().CloneUnitCtxMap[*UnitId].IsConstructed)
      continue;
    collectReferencedTypeSignatures(TU->getUnitDIE(), ProcessedTU, TUlist);
  }

  SmallVectorImpl<DWARFUnit *> &TUVec =
      TUType == ProcessingType::DWARF4TUs ? getState().DWARF4TUVector
                                          : getState().DWARF5TUVector;
  llvm::sort(TUVec, [](const DWARFUnit *A, const DWARFUnit *B) {
    return A->getOffset() < B->getOffset();
  });
  for (DWARFUnit *TU : TUVec)
    constructFromUnit(*TU);
}
```

bolt/lib/Rewrite/RewriteInstance.cpp
Thrrreeee: In this issue, there are still some data races in `DWARFContext` that cannot be directly resolved by this lock. I think the better approach is to simply set it to `true`. I will remove the modification to this in (https://github.com/llvm/llvm-project/pull/197859)

bolt/lib/Core/DIEBuilder.cpp
Thrrreeee: > Please move into callee. 

Done.

> Also since DWP context is shared, we are affectively serializing this whole thing due to type parsing.

 The serialization here for DWP is unavoidable. Based on my earlier testing, the performance impact is not significant.

bolt/include/bolt/Core/BinaryContext.h
Thrrreeee: Done.

bolt/lib/Core/DIEBuilder.cpp
Thrrreeee: Done

bolt/lib/Core/DIEBuilder.cpp
Thrrreeee: Done.

bolt/lib/Core/DIEBuilder.cpp
Thrrreeee: There is already a test at `bolt/test/X86/dwarf4-ftypes-dwp-input-dwp-output.test`

bolt/lib/Core/DIEBuilder.cpp
Thrrreeee: > Isn't this already enforced by the structure of DWARF? Even if we process each CU in parallel it's internal structure is deterministic?

Its internal structure are deterministic when process each CU in parallel. But there the TU vector order here is not derived from the original TU order. In the DWP path, registerUnit() is called according to the     
  order in which DW_FORM_ref_sig8 references are discovered from the current CU/TU graph. That first-discovery order can differ from the original type-unit offset order, and the vector 
  is later used directly for emission / TU index generation. Sorting by unit offset restores a input-order before emitting.  

If you think the traversal order is already sufficient here, we can also drop this sort. However, we may also need to modify the corresponding testcases.

ayermolo (approved):
LGTM.
Thanks for fixing this. 
please address couple comments before merging.

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -283,6 +283,78 @@ void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
...
+  collectReferencedTypeSignatures(U.getUnitDIE(), ProcessedTU, TUlist);
+
+  // addressing type units referenced.
+  for (unsigned I = 0; I != TUlist.size(); ++I) {
+    const uint64_t Signature = TUlist[I];
+    DWARFTypeUnit *TU = DwarfContext->getTypeUnitForHash(Signature, true);
+    if (!TU)
+      continue;
+
+    getState().Type = TU->getVersion() < 5 ? ProcessingType::DWARF4TUs
```
ayermolo: Please move this outside of the loop. There is no mixing of DWARF4 and DWARF5 in DWP so check on U before the loop is enough.

bolt/lib/Core/DIEBuilder.cpp
```diff
@@ -283,6 +283,78 @@ void DIEBuilder::buildTypeUnits(DebugStrOffsetsWriter *StrOffsetWriter,
...
+    collectReferencedTypeSignatures(TU->getUnitDIE(), ProcessedTU, TUlist);
+  }
+
+  // Ensure original order of processing type units
+  auto SortByOffset = [](const DWARFUnit *A, const DWARFUnit *B) {
+    return A->getOffset() < B->getOffset();
+  };
+
+  // For Split DWARF, we have either DWARF4 or DWARF5, they cannot be mixed.
+  std::vector<DWARFUnit *> &TUVec = !getState().DWARF4TUVector.empty()
```
ayermolo: nit: can just use getState().Type

Thrrreeee:
> LGTM. Thanks for fixing this. please address couple comments before merging.

Thanks for review — all comments have been addressed.

PR #197361, opened 2026-05-13
[MIR] Save internal VirtRegMap state in MIR

qcolombet:
Adds two optional fields to the per-vreg YAML record so MIR tests can express VirtRegMap state that previously had no representation:

  registers:
    - { id: 1, class: vgpr_32, split-from: '%0', assigned-phys: '$vgpr5' }

Testing passes that consume sibling-register information (e.g. InlineSpiller) requires constructing a VirtRegMap with split relationships from a MIR test, which implies triggering live-range splitting at minimum and make reproducers unnecessarily complicated.

So this change introduces a mechanism to serialize/deserialize the state of the VirtRegMap pass.

Mechanism:
- For serialization:
  - MIRPrinter emits the new fields only when the new -mir-emit-vrm option is set and the VirtRegMap is available.
- For deserialization:
  - MIRParser stashes parsed entries on the generic MachineFunctionInfo base (PendingVRegMappings), using the existing parseVirtualRegisterReference / parseRegisterReference helpers.
  - VirtRegMap::init() drains the stash via assignVirt2Phys and setIsSplitFromReg, then clears it.

Validation at parse time:
  - 'assigned-phys' must be a physical register.
  - 'split-from' must reference a different vreg than 'id'.
  - 'split-from' physregs / named-vreg references already rejected by parseStandaloneVirtualRegister.

I'm not super happy about stashing the VRM info in the MFI, but that's a small price to pay.

### DIFF
--- a/llvm/include/llvm/CodeGen/MIRPrinter.h
+++ b/llvm/include/llvm/CodeGen/MIRPrinter.h
@@ -24,6 +24,7 @@ class MachineBasicBlock;
 class MachineFunction;
 class MachineModuleInfo;
 class Module;
+class VirtRegMap;
 template <typename T> class SmallVectorImpl;
 
 class PrintMIRPreparePass : public RequiredPassInfoMixin<PrintMIRPreparePass> {
@@ -47,12 +48,18 @@ public:
 LLVM_ABI void printMIR(raw_ostream &OS, const Module &M);
 
 /// Print MIR using Legacy Pass Manager (uses MachineModuleInfo).
+/// If \p VRM is non-null, the printer will serialize VirtRegMap state
+/// (split-from, assigned-phys).
 LLVM_ABI void printMIR(raw_ostream &OS, const MachineModuleInfo &MMI,
-                       const MachineFunction &MF);
+                       const MachineFunction &MF,
+                       const VirtRegMap *VRM = nullptr);
 
 /// Print MIR using New Pass Manager (uses FunctionAnalysisManager).
+/// If \p VRM is non-null, the printer will serialize VirtRegMap state
+/// (split-from, assigned-phys).
 LLVM_ABI void printMIR(raw_ostream &OS, FunctionAnalysisManager &FAM,
-                       const MachineFunction &MF);
+                       const MachineFunction &MF,
+                       const VirtRegMap *VRM = nullptr);
 
 /// Determine a possible list of successors of a basic block based on the
 /// basic block machine operand being used inside the block. This should give
--- a/llvm/include/llvm/CodeGen/MIRPrinter.h
+++ b/llvm/include/llvm/CodeGen/MIRPrinter.h
@@ -24,6 +24,7 @@ class MachineBasicBlock;
 class MachineFunction;
 class MachineModuleInfo;
 class Module;
+class VirtRegMap;
 template <typename T> class SmallVectorImpl;
 
 class PrintMIRPreparePass : public RequiredPassInfoMixin<PrintMIRPreparePass> {
@@ -47,12 +48,18 @@ public:
 LLVM_ABI void printMIR(raw_ostream &OS, const Module &M);
 
 /// Print MIR using Legacy Pass Manager (uses MachineModuleInfo).
+/// If \p VRM is non-null and -mir-emit-vrm is enabled, the printer will
+/// serialize VirtRegMap state (split-from, assigned-phys).
 LLVM_ABI void printMIR(raw_ostream &OS, const MachineModuleInfo &MMI,
-                       const MachineFunction &MF);
+                       const MachineFunction &MF,
+                       const VirtRegMap *VRM = nullptr);
 
 /// Print MIR using New Pass Manager (uses FunctionAnalysisManager).
+/// If \p VRM is non-null and -mir-emit-vrm is enabled, the printer will
+/// serialize VirtRegMap state (split-from, assigned-phys).
 LLVM_ABI void printMIR(raw_ostream &OS, FunctionAnalysisManager &FAM,
-                       const MachineFunction &MF);
+                       const MachineFunction &MF,
+                       const VirtRegMap *VRM = nullptr);
 
 /// Determine a possible list of successors of a basic block based on the
 /// basic block machine operand being used inside the block. This should give
--- a/llvm/include/llvm/CodeGen/MIRYamlMapping.h
+++ b/llvm/include/llvm/CodeGen/MIRYamlMapping.h
@@ -203,12 +203,19 @@ struct VirtualRegisterDefinition {
   StringValue Class;
   StringValue PreferredRegister;
   std::vector<FlowStringValue> RegisterFlags;
+  // VirtRegMap state.
+  // SplitFrom: id-form virtual register only (e.g. '%0'); physregs and named
+  //            vregs are rejected by the parser.
+  // AssignedPhys: physical register only (e.g. '$r5'); virtregs are rejected.
+  StringValue SplitFrom;
+  StringValue AssignedPhys;
 
   // TODO: Serialize the target specific register hints.
 
   bool operator==(const VirtualRegisterDefinition &Other) const {
     return ID == Other.ID && Class == Other.Class &&
-           PreferredRegister == Other.PreferredRegister;
+           PreferredRegister == Other.PreferredRegister &&
+           SplitFrom == Other.SplitFrom && AssignedPhys == Other.AssignedPhys;
   }
 };
 
@@ -220,6 +227,14 @@ template <> struct MappingTraits<VirtualRegisterDefinition> {
                        StringValue()); // Don't print out when it's empty.
     YamlIO.mapOptional("flags", Reg.RegisterFlags,
                        std::vector<FlowStringValue>());
+    // MIRPrinter sets WriteDefaultValues=true unless -simplify-mir is passed,
+    // so a plain mapOptional with an empty default would still emit the keys
+    // and change every existing test's output.
+    // Skip the call on output when empty to keep them off entirely.
+    if (!YamlIO.outputting() || !Reg.SplitFrom.Value.empty())
+      YamlIO.mapOptional("split-from", Reg.SplitFrom, StringValue());
+    if (!YamlIO.outputting() || !Reg.AssignedPhys.Value.empty())
+      YamlIO.mapOptional("assigned-phys", Reg.AssignedPhys, StringValue());
   }
 
   static const bool flow = true;
--- a/llvm/include/llvm/CodeGen/MachineFunction.h
+++ b/llvm/include/llvm/CodeGen/MachineFunction.h
@@ -901,12 +901,7 @@ public:
 
   MachineFunctionInfo *cloneInfoFrom(
       const MachineFunction &OrigMF,
-      const DenseMap<MachineBasicBlock *, MachineBasicBlock *> &Src2DstMBB) {
-    assert(!MFInfo && "new function already has MachineFunctionInfo");
-    if (!OrigMF.MFInfo)
-      return nullptr;
-    return OrigMF.MFInfo->clone(Allocator, *this, Src2DstMBB);
-  }
+      const DenseMap<MachineBasicBlock *, MachineBasicBlock *> &Src2DstMBB);
 
   /// Returns the denormal handling type for the default rounding mode of the
   /// function.
--- a/llvm/include/llvm/CodeGen/MachineFunction.h
+++ b/llvm/include/llvm/CodeGen/MachineFunction.h
@@ -129,6 +129,7 @@ struct LLVM_ABI MachineFunctionInfo {
       const {
     return nullptr;
   }
+
 };
 
 /// Properties which a MachineFunction may have at a given point in time.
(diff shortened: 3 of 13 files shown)


qcolombet:
Stacked PR:
https://github.com/llvm/llvm-project/pull/197363
https://github.com/llvm/llvm-project/pull/197362
https://github.com/llvm/llvm-project/pull/197361 👈

llvm/include/llvm/CodeGen/MachineFunction.h
```diff
@@ -129,6 +129,16 @@ struct LLVM_ABI MachineFunctionInfo {
       const {
     return nullptr;
   }
+
+  // Placeholder for deserialization of the VirtRegMap object.
+  // This gets populuated by the MIRParser when the information is present
```
arsenm:
```suggestion
  // This gets populated by the MIRParser when the information is present
```

llvm/include/llvm/CodeGen/MIRYamlMapping.h
```diff
@@ -220,6 +228,14 @@ template <> struct MappingTraits<VirtualRegisterDefinition> {
                        StringValue()); // Don't print out when it's empty.
     YamlIO.mapOptional("flags", Reg.RegisterFlags,
                        std::vector<FlowStringValue>());
+    // MIRPrinter sets WriteDefaultValues=true unless -simplify-mir is passed,
```
arsenm, line 230: Unrelated but really -simplify-mir should be the default. I waste so much time in reviews reducing unnecessary MIR bits

llvm/lib/CodeGen/VirtRegMap.cpp
```diff
@@ -74,6 +74,21 @@ void VirtRegMap::init(MachineFunction &mf) {
   Virt2ShapeMap.clear();
 
   grow();
+
+  // Drain any VirtRegMap state stashed by MIRParser on the generic
+  // MachineFunctionInfo base. Must run after grow() so that Virt2*Map have
+  // capacity for every vreg referenced by the stash.
+  if (auto *MFI = mf.getInfo<MachineFunctionInfo>()) {
```
arsenm: This can't fail

llvm/test/CodeGen/MIR/AMDGPU/virtregmap-stash-bad-phys.mir
```diff
@@ -0,0 +1,18 @@
+# RUN: not llc -mtriple=amdgcn -mcpu=gfx1100 -run-pass=virtregmap \
+# RUN:     -o /dev/null %s 2>&1 | FileCheck %s
```
arsenm:
```suggestion
# RUN: not llc -mtriple=amdgcn -mcpu=gfx1100 -run-pass=virtregmap \
# RUN:     -filetype=null %s 2>&1 | FileCheck %s
```

llvm/test/CodeGen/MIR/AMDGPU/virtregmap-stash-bad-split.mir
```diff
@@ -0,0 +1,16 @@
+# RUN: not llc -mtriple=amdgcn -mcpu=gfx1100 -run-pass=virtregmap \
+# RUN:     -o /dev/null %s 2>&1 | FileCheck %s
```
arsenm:
```suggestion
# RUN:     -filetype=null %s 2>&1 | FileCheck %s
```

llvm/test/CodeGen/MIR/AMDGPU/virtregmap-stash-self-split.mir
```diff
@@ -0,0 +1,17 @@
+# RUN: not llc -mtriple=amdgcn -mcpu=gfx1100 -run-pass=virtregmap \
+# RUN:     -o /dev/null %s 2>&1 | FileCheck %s
```
arsenm:
```suggestion
# RUN:     -filetype=null %s 2>&1 | FileCheck %s
```

llvm/lib/CodeGen/MIRPrinter.cpp
```diff
@@ -74,6 +75,11 @@ static cl::opt<bool> SimplifyMIR(
 static cl::opt<bool> PrintLocations("mir-debug-loc", cl::Hidden, cl::init(true),
                                     cl::desc("Print MIR debug-locations"));
 
+static cl::opt<bool> EmitVRMState(
+    "mir-emit-vrm", cl::Hidden, cl::init(false),
+    cl::desc("Serialize VirtRegMap state (split-from, assigned-phys) into "
+             "MIR YAML when a VirtRegMap is available"));
```
arsenm: Should just do this without an option

llvm/include/llvm/CodeGen/MIRPrinter.h
```diff
@@ -47,12 +48,18 @@ class PrintMIRPass : public RequiredPassInfoMixin<PrintMIRPass> {
...
+                       const MachineFunction &MF,
+                       const VirtRegMap *VRM = nullptr);
 
 /// Print MIR using New Pass Manager (uses FunctionAnalysisManager).
+/// If \p VRM is non-null and -mir-emit-vrm is enabled, the printer will
+/// serialize VirtRegMap state (split-from, assigned-phys).
 LLVM_ABI void printMIR(raw_ostream &OS, FunctionAnalysisManager &FAM,
-                       const MachineFunction &MF);
+                       const MachineFunction &MF,
+                       const VirtRegMap *VRM = nullptr);
```
arsenm, line 62: Can't you query this from the FunctionAnalysisManager inside printMIR?

llvm/include/llvm/CodeGen/MachineFunction.h
arsenm, line 127: Cloning not handled

llvm/lib/CodeGen/MIRParser/MIRParser.cpp
```diff
@@ -804,6 +804,54 @@ bool MIRParserImpl::parseRegisterInfo(PerFunctionMIParsingState &PFS,
...
 
+  // Stash any VirtRegMap state on the generic MachineFunctionInfo base.
+  // VirtRegMap::init() will use that information to get pre-populated
+  // on the first analysis run.
+  for (const auto &VReg : YamlMF.VirtualRegisters) {
+    if (VReg.SplitFrom.Value.empty() && VReg.AssignedPhys.Value.empty())
+      continue;
+
+    auto *MFI = MF.getInfo<MachineFunctionInfo>();
+    if (!MFI)
```
arsenm: This shouldn't really require target support

llvm/include/llvm/CodeGen/MachineFunction.h
```diff
@@ -129,6 +129,16 @@ struct LLVM_ABI MachineFunctionInfo {
...
+  // and is used in VirtRegMap::init() to initialize its state.
+  struct PendingVRegMapping {
+    Register VReg;
+    Register SplitFrom;       // NoReg if absent.
+    MCRegister AssignedPhys;  // NoReg if absent.
+  };
+  SmallVector<PendingVRegMapping, 0> PendingVRegMappings;
```
arsenm: I don't think this belongs in MachineFunctionInfo, it's not target specific. I imagined this as introducing a new analysis section of the function, which if present would populate the VirtRegMap. I suppose it still needs to be attached to the function, since the parsing doesn't necessarily depend on a pass manager context

llvm/include/llvm/CodeGen/MIRYamlMapping.h
qcolombet: Agree :)

llvm/include/llvm/CodeGen/MachineFunction.h
qcolombet: That's right. At this point we don't have the VirtRegMap available and it may not ever be available.
That's the one piece of this PR that is not great, but I felt it was a small price to pay given how much time it would save down the line.

llvm/lib/CodeGen/MIRParser/MIRParser.cpp
qcolombet: I thought that not all target supported the serialization/deserialization of the MachineFunctionInfo.
I.e., I believe it is possible that this may be nullptr. If not that simplify things

llvm/include/llvm/CodeGen/MachineFunction.h
qcolombet: Do we care?

llvm/lib/CodeGen/MIRPrinter.cpp
qcolombet: Let me see the impact on the existing tests.
I guess that's a one time cost anyway, so we should just absorb it here.

llvm/lib/CodeGen/VirtRegMap.cpp
qcolombet: Good catch!

llvm/include/llvm/CodeGen/MachineFunction.h
qcolombet: I mean this feature is only to make reproducers easier to capture, not a general feature.
Should we just assert the PendingVRegMappings is empty?

llvm/include/llvm/CodeGen/MIRPrinter.h
qcolombet: We could but then the legacy and new pass manager signatures for printMIR would be inconsistent.
Maybe we don't care, though.
What do you think?

llvm/lib/CodeGen/VirtRegMap.cpp
qcolombet: Actually, is this true?
If the target doesn't implement `createMachineFunctionInfo` we could get nullptr here, no?

qcolombet:
@arsenm thanks for the review.
Regarding your comments:
- For the check around `getInfo<MachineFunctionInfo>` here and there, I believe that technically the `MachineFunctionInfo` could be `nullptr` if the target doesn't override the `createMachineFunctionInfo` method. But let me know if I'm missing something.
- For the clone function, do we care given that this is only for making test capture easier? I gets we could assert in clone is not used in that case.
- Regarding stashing the information in the MachineFunctionInfo to begin with, this is the hacky part of this PR. We don't have the VRM around at this point and the MFI is the next best thing to stash this stuff.
- For the printMIR signature. On one of them we could remove the VRM argument. I still kept it for consistency, but let me know.

If we think this is too hacky we can drop the whole thing. The final test case (end of the stack PR) is a nice reduction though.

arsenm:
> * For the check around `getInfo<MachineFunctionInfo>` here and there, I believe that technically the `MachineFunctionInfo` could be `nullptr` if the target doesn't override the `createMachineFunctionInfo` method. But let me know if I'm missing something.

We probably should just make that pure virtual and force targets to implement it. It's implemented by close to all targets to avoid worrying about a special case (but also would be avoided if this weren't put into MachineFunctionInfo in the first place)

> * For the clone function, do we care given that this is only for making test capture easier? I gets we could assert in clone is not used in that case.

Yes, because this is what llvm-reduce relies on. Making MIR reduction work better is a primary reason to improve missed representation. 

> * Regarding stashing the information in the MachineFunctionInfo to begin with, this is the hacky part of this PR. We don't have the VRM around at this point and the MFI is the next best thing to stash this stuff.

It could also just as well go directly into MachineFunction, which is equally as hacky.

qcolombet:
> > * For the check around `getInfo<MachineFunctionInfo>` here and there, I believe that technically the `MachineFunctionInfo` could be `nullptr` if the target doesn't override the `createMachineFunctionInfo` method. But let me know if I'm missing something.
> 
> We probably should just make that pure virtual and force targets to implement it. It's implemented by close to all targets to avoid worrying about a special case (but also would be avoided if this weren't put into MachineFunctionInfo in the first place)

I'm not going to touch that part but let me explore stashing the information somewhere else. MachineRegisterInfo could be a good compromise.

> 
> > * For the clone function, do we care given that this is only for making test capture easier? I gets we could assert in clone is not used in that case.
> 
> Yes, because this is what llvm-reduce relies on. Making MIR reduction work better is a primary reason to improve missed representation.

Ah right!

> 
> > * Regarding stashing the information in the MachineFunctionInfo to begin with, this is the hacky part of this PR. We don't have the VRM around at this point and the MFI is the next best thing to stash this stuff.
> 
> It could also just as well go directly into MachineFunction, which is equally as hacky.

Indeed, I'll try with MachineRegisterInfo. It is still as hacky, but less front and center in the APIs :).

qcolombet:
@arsenm please take a look with the threading through MRI.

qcolombet:
@arsenm , ping

llvm/include/llvm/CodeGen/MachineFunction.h
arsenm: This is for llvm-reduce, and reducing this intermediate state is largely the point of accurately capturing all of this state

llvm/include/llvm/CodeGen/MachineFunction.h
```diff
@@ -129,6 +129,7 @@ struct LLVM_ABI MachineFunctionInfo {
       const {
     return nullptr;
   }
+
```
arsenm:
```suggestion
```

llvm/include/llvm/CodeGen/MachineRegisterInfo.h
```diff
@@ -107,6 +115,8 @@ class MachineRegisterInfo {
              VirtReg2IndexFunctor>
       RegAllocHints;
 
+  SmallVector<PendingVirtRegMapEntry, 0> PendingVirtRegMapEntries;
```
arsenm, line 120: Document field

arsenm (approved):

PR #197362, opened 2026-05-13
[MIR] Serialize/Deserialize MachineInstr::LRSplit flag

qcolombet:
The LRSplit MachineInstr flag is set by SplitKit on copies inserted for live-range splitting.
Until now the flag had no MIR-text representation.

This patch fixes that so that it gets easier to reproduce/capture issues that involves SplitKit.

Round-trip coverage in
llvm/test/CodeGen/MIR/AMDGPU/lr-split-flag.mir.

PR #194494, opened 2026-04-27
[MIR] Introduce split-from VRM state and lr-split instr flag

qcolombet:
Stacked PR will be split, but sharing for early feedback

PR #197363, opened 2026-05-13
[AMDGPU][test] Use mir test for regalloc issue

qcolombet:
Use the newly introduced split-from flag to produce a more robust test case for the hoistSpillInsideBB live-range update issue.

NFC


qcolombet:
Stacked PR:
https://github.com/llvm/llvm-project/pull/197363 👈
https://github.com/llvm/llvm-project/pull/197362
https://github.com/llvm/llvm-project/pull/197361

qcolombet:
@arsenm this is the cleaned up version of the PR to be able to reproduce the spilling issue from a couple of weeks ago from a ~3k line of LLVM IR to ~20 lines of mir.

qcolombet:
Clean up of https://github.com/llvm/llvm-project/pull/194494

llvm/test/CodeGen/AMDGPU/regalloc-hoist-spill-live-range-upd.mir
```diff
@@ -0,0 +1,71 @@
...
+#   -- # %1 = COPY %0 gets removed
+#  ... # prologue instruction
+#  store %0 # %0 liveness is extended to this point.
+#  ... # high pressure point
+#  %1_reload = load
+#  use %1_reload
+#
+# To be able to hit the problematic code in hoistSpillInsideBB, %0 needs to
+# be defined by a PHI (hence the diamond with two %0 defs in the test case)
+# and %0 and %1 need to be sibling registers (hence the %3 split-from).
```
rovka, line 34: This whole comment kind of screams "unit test". Would it take a lot of boilerplate to do this in a unit test instead?

llvm/test/CodeGen/AMDGPU/regalloc-hoist-spill-live-range-upd.mir
```diff
@@ -0,0 +1,71 @@
+# RUN: llc -mtriple=amdgcn-amd-amdhsa -mcpu=gfx950 -run-pass=greedy -verify-regalloc -o - %s | FileCheck %s
```
arsenm: It probably doesn't matter in this narrow case, but the realistic situation requires the RA filter function for SGPR/WWM/VGPR. Just using -run-pass=greedy runs the regular pass with no filter

llvm/test/CodeGen/AMDGPU/regalloc-hoist-spill-live-range-upd.mir
qcolombet: No because with unittest you can create any weird situation you want that may not be realistic.
The point is to show that this is reachable through real IR.

llvm/test/CodeGen/AMDGPU/regalloc-hoist-spill-live-range-upd.mir
qcolombet: Do we have a what to run greedy for VGPR only?
Or I need to use start/stop options?

arsenm: I think we have something for new PM, for old PM I've always used -start-before=greedy,N -stop-after=greedy,N

arsenm: New pm: `-passes='greedy<wwm>`

qcolombet: Used the new PM flag and checked it was still failing if I remove the related inline spiller fix.

arsenm (approved):

PR #197151, opened 2026-05-12
[libclc] Move external-funcs.test to static file and use IR checks for .cl tests

wenju-he:
This PR supercedes #87989.

Moving external-funcs.test to static file simplifies test/CMakeLists.txt. Static files follows llvm standard lit pattern and enables fine-grained check of missing symbols in specific libraries.

.cl test files uses %target, %cpu and %check_prefix, which are replaced with specific values during `ninja check-libclc` or `llvm-lit build/runtimes/runtimes-${triple}-llvm-bins/libclc/test`. This allows checking outputs of multiple triples in the same test file.

Add script libclc/test/update_libclc_tests.py, which wraps utils/update_cc_test_checks.py to update CHECK lines in libclc .cl tests for a given arch. Example usage:
`libclc/test/update_libclc_tests.py amdgpu`

Assisted-by: Claude Sonnet 4.6

PR #197364, opened 2026-05-13
[AMDGPU] AMDGPULibCalls: Set new intrinsic calling convention to C

wenju-he:
In #197151 libclc/test/math/fabs.cl, tryReplaceLibcallWithSimpleIntrinsic replaces `call fastcc float @_Z4fabsf` with `call fastcc float @llvm.fabs.f32`. But intrinsic call must use CallingConv::C.


arsenm (approved):
We probably should make this an IR verifier error

wenju-he:
> We probably should make this an IR verifier error

thanks, I'm trying to add the check.

PR #196589, opened 2026-05-08
[lldb] Assert that CommandObject::DoExecute sets a return status

JDevlieghere:
Change the default value of CommandReturnObject::m_status from eReturnStatusStarted to eReturnStatusInvalid, and add a debug-only RAII check in CommandObjectParsed::Execute and CommandObjectRaw::Execute that asserts the status is no longer Invalid after DoExecute returns. 

This catches commands that forget to call SetStatus on a success or failure path. Succeeded() still returns true when the status is Invalid (0 sorts below eReturnStatusSuccessContinuingResult), so helpers that read result.Succeeded() as a precondition before any explicit SetStatus (e.g. StopProcessIfNecessary) continue to work.

rdar://176506732

PR #197367, opened 2026-05-13
Revert "[lldb] Assert that CommandObject::DoExecute sets a return status"

slydiman:
Reverts llvm/llvm-project#196589

[lldb-remote-linux-ubuntu](https://lab.llvm.org/buildbot/#/builders/195/builds/24894): 262 shell tests failed
[lldb-remote-linux-win](https://lab.llvm.org/buildbot/#/builders/197/builds/15918): 193 shell tests failed


JDevlieghere:
A quick scan at the log seems to suggest all the assertions are triggered from `platform connect`, so probably that command is not setting the return status.

PR #193962, opened 2026-04-24
[RISCV] Prefer SP over FP for frame index access when offset fits within compressed immediate range.

quic-garvgupt:
Before this change, we would use fp/s0/x8 for most stack accesses when
frame pointers were present. This is an over-approximation when a
stack slot is reachable from both SP and FP with no scalable offset.

This patch replaces the unconditional getFrameRegister() call in
getFrameIndexReference with an explicit register selection decision tree.

When both SP and FP are available (no stack realignment, no RVV objects,
no variable-sized objects), prefer SP if the SP-relative offset fits in the
compressed instruction immediate range (<=252 for RV32, <=504 for RV64).
This enables compression for sp-relative instructions to c.swsp/c.lwsp (RV32)
and c.sdsp/c.ldsp (RV64) thereby reducing code size.

PR #197368, opened 2026-05-13
[RISCV] Check SP-relative offset in needsFrameBaseReg when FP offset overflows

quic-garvgupt:
When a frame pointer is present, `needsFrameBaseReg` previously only checked the FP-relative offset to decide if a virtual base register was needed. If the worst-case FP offset exceeded the 12-bit immediate range, a base register was always materialized, even when the SP-relative offset would fit.

Since `getFrameIndexReference` can now select SP over FP when the offset fits in the compressed instruction immediate range, also check the SP-relative offset before deciding a base register is needed. This avoids unnecessary base register materialization and results in some code size savings.


quic-garvgupt:
Below are the code size numbers on embench with rv32imac/ilp32 combo at -Os

| Metric | Value |
|---|---|
| FP overhead (unpatched) | +5.14% vs -fomit-fp |
| FP overhead (patched) | +2.66% vs -fomit-fp |
| Patch savings | -2.36% vs unpatched -fno-omit-fp |
| Patch recovers | 48.2% of the frame pointer overhead |

The above numbers represent the net savings from this PR combined with https://github.com/llvm/llvm-project/pull/193962.

topperc (approved):
LGTM

lenary (approved):
LGTM

PR #197372, opened 2026-05-13
[X86] Add atomic vector store tests for unaligned >1 sizes.

jofrn:
Unaligned atomic vector stores with size >1 are lowered to calls.
Adding their tests separately here.

Store-side counterpart to #148896. Stacked below #197165.


RKSimon (approved):
LGTM

PR #197373, opened 2026-05-13
[InstCombine] Drop `(X + Z) - (Y + Z) --> (X - Y)` fold

el-ev:
The pattern below does the same thing and does it better


nikic (approved):
LGTM

PR #197374, opened 2026-05-13
[X86][GlobalISel] Remove dependency on legal ruleset

davemgreen:
This fills in always legal rules, to remove the dependency on the legacy ruleset. I'm not sure about the truncate rule but all tests pass. This is not guaranteed to be all the rules, just the ones that appear in tests.


llvm/lib/Target/X86/GISel/X86LegalizerInfo.cpp
```diff
@@ -425,6 +426,8 @@ X86LegalizerInfo::X86LegalizerInfo(const X86Subtarget &STI,
       .clampScalar(1, s8, sMaxScalar)
       .scalarize(0);
 
+  getActionDefinitionsBuilder(G_TRUNC).alwaysLegal();
```
arsenm: This shouldn't be the case. G_TRUNC "always" being legal is one of those weird early decisions that I don't think ever really worked

RKSimon: For x86 scalars its fine, but vectors definitely not (even AVX512 can struggle with some legal src/dst type combos).

But if there's a better generic GISel way to do this, I'd love to avoid special cases for x86.

arsenm: This should only be legal for valid register types

davemgreen: I have added the types needed for the tests, except for the obviously dodgy ones. No vectors yet - that matches ZEXT and SEXT above and vectors have more difficult truncate rules.

I would honestly recommend the previous alwaysLegal version as it mirrors what all other targets do and how it worked before, and can be fixed up going forward.

arsenm (approved):

arsenm:
Is anything blocking merging this?

RKSimon (approved):

davemgreen:
Thanks. This may need to be adjusted in the future for vector types but I will leave that to you folks. I think this was the last of the legacy rules that needed removing.

e-kud:
Checked on the set of benchmarks that I usually use for measuring the coverage of GlobalISel. No regression was found at least significant to influence more than 0.01% (out of ~4mln functions)

PR #197376, opened 2026-05-13
[mlir][arith] Add subui_extended op

aobolensk:
Unsigned extended subtraction mirroring addui_extended. Gives a primitive for unsigned underflow that maps cleanly to llvm.usub.with.overflow and spirv.ISub

### DIFF
--- a/mlir/include/mlir/Dialect/Arith/IR/ArithOps.td
+++ b/mlir/include/mlir/Dialect/Arith/IR/ArithOps.td
@@ -360,6 +360,59 @@ def Arith_AddUIExtendedOp : Arith_Op<"addui_extended", [Pure, Commutative,
 }
 
 //===----------------------------------------------------------------------===//
+// SubUIExtendedOp
+//===----------------------------------------------------------------------===//
+
+def Arith_SubUIExtendedOp : Arith_Op<"subui_extended", [Pure,
+    AllTypesMatch<["lhs", "rhs", "diff"]>]> {
+  let summary = [{
+    extended unsigned integer subtraction operation returning difference and
+    borrow bit
+  }];
+
+  let description = [{
+    Performs (N+1)-bit subtraction on zero-extended operands. Returns two
+    results: the N-bit difference (same type as both operands), and the borrow
+    bit (boolean-like), where `1` indicates unsigned subtraction underflow
+    (i.e. `lhs < rhs` when interpreted as unsigned), while `0` indicates no
+    underflow.
+
+    Example:
+
+    ```mlir
+    // Scalar subtraction.
+    %diff, %borrow = arith.subui_extended %b, %c : i64, i1
+
+    // Vector element-wise subtraction.
+    %d:2 = arith.subui_extended %e, %f : vector<4xi32>, vector<4xi1>
+
+    // Tensor element-wise subtraction.
+    %x:2 = arith.subui_extended %y, %z : tensor<4x?xi8>, tensor<4x?xi1>
+    ```
+  }];
+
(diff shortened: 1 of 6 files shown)


kuhar (commented):
What do you plan to use it for and are there any alternatives?

aobolensk:
> What do you plan to use it for and are there any alternatives?

Plan to lower it to spirv in the way it is done for addui extended

kuhar:
For example, we had this RFC for addui_extended https://discourse.llvm.org/t/rfc-add-integer-add-with-carry-op-to-arith/64573/6 and this RFC for mului_extended: https://discourse.llvm.org/t/rfc-arith-add-extended-multiplication-ops/66869 . I don't necessarily think an RFC is necessary to make progress here, but this needs some motivation to ground the work, IMO.

aobolensk:
> For example, we had this RFC for addui_extended https://discourse.llvm.org/t/rfc-add-integer-add-with-carry-op-to-arith/64573/6 and this RFC for mului_extended: https://discourse.llvm.org/t/rfc-arith-add-extended-multiplication-ops/66869 . I don't necessarily think an RFC is necessary to make progress here, but this needs some motivation to ground the work, IMO.

`subui_extended` closes the obvious gap and maps 1:1 to `llvm.usub.with.overflow` and `spirv.ISubBorrow` as I mentioned in the description. The conversions to LLVM and SPIR-V are planned to be added afterwards. Corresponding high level language operations are `usubBorrow` in GLSL and `usubb` in HLSL correspondingly.

Example in GLSL:
```glsl
uvec4 borrow;
uvec4 d = usubBorrow(x, y, borrow);
```

The motivation for the change is the same for `addui_extended` and `mului_extended` basically, for wide integer emulation and operations with wide ints, to be more precise. 

Does that make sense?

kuhar:
Do you plan to contribute some transformations that will produce these ops in mlir or in a downstream compiler? I'm trying to understand how this will concretely connect e2e.

kuhar:
> for wide integer emulation and operations with wide ints, to be more precise.

FTR, wide subi emulation is already supported (but without extended subui): https://github.com/llvm/llvm-project/blob/a023fd290ec084663b4dc482dc0a1d77940628ea/mlir/lib/Dialect/Arith/Transforms/EmulateWideInt.cpp#L1183

aobolensk:
> Do you plan to contribute some transformations that will produce these ops in mlir or in a downstream compiler?

Yes, in mlir upstream

aobolensk:
> > for wide integer emulation and operations with wide ints, to be more precise.
> 
> FTR, wide subi emulation is already supported (but without extended subui):
> 
> https://github.com/llvm/llvm-project/blob/a023fd290ec084663b4dc482dc0a1d77940628ea/mlir/lib/Dialect/Arith/Transforms/EmulateWideInt.cpp#L1183

Yes, I'm aware of this pass

krzysz00 (approved):
Since this is a gap in `arith` and plugs a useful hole (and will let us clean up wide int emulation), approved - though you might want to consider adding the lowerings - or fixing up subtraction emulation - in this PR too so there's some usecase

kuhar:
If you can use it to simplify the existing subi emulation and connect to llvm and spirv, it seems like strong enough motivation to move forward. Can you confirm this is the plan?

aobolensk:
> If you can use it to simplify the existing subi emulation and connect to llvm and spirv, it seems like strong enough motivation to move forward. Can you confirm this is the plan?

Yes, I confirm

kuhar (approved):

PR #197377, opened 2026-05-13
[RISCV][GlobalISel] Remove dependency on legal ruleset

davemgreen:
This fills in always legal rules, to remove the dependency on the legacy ruleset. I'm really not sure about the truncate rule but all tests pass. This is not guaranteed to be all the rules, just the ones that appear in tests.

### DIFF
--- a/llvm/lib/Target/RISCV/GISel/RISCVLegalizerInfo.cpp
+++ b/llvm/lib/Target/RISCV/GISel/RISCVLegalizerInfo.cpp
@@ -176,6 +176,8 @@ RISCVLegalizerInfo::RISCVLegalizerInfo(const RISCVSubtarget &ST)
       .customIf(typeIsLegalBoolVec(1, BoolVecTys, ST))
       .maxScalar(0, sXLen);
 
+  getActionDefinitionsBuilder(G_TRUNC).alwaysLegal();
+
   getActionDefinitionsBuilder(G_SEXT_INREG)
       .customFor({sXLen})
       .clampScalar(0, sXLen, sXLen)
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/RISCV/GISel/RISCVLegalizerInfo.cpp
```diff
@@ -176,6 +176,8 @@ RISCVLegalizerInfo::RISCVLegalizerInfo(const RISCVSubtarget &ST)
       .customIf(typeIsLegalBoolVec(1, BoolVecTys, ST))
       .maxScalar(0, sXLen);
 
+  getActionDefinitionsBuilder(G_TRUNC).alwaysLegal();
```
arsenm, line 179: Same as X86, this really ought to only cover the legal integer types

topperc: I think it also needs to include the legal FP types.

arsenm: This is one of the unresolved issues of the switch to having FP types

lenary: I think it's reasonable for this to ensure only that any in-tree tests are working. RISC-V's GlobalISel implementation has lots of deficiencies, and it is not Dave's role to fix things without in-tree tests.

topperc: I notice that AArch64 also uses .alwaysLegal() at the end of their rules. I think this is probably fine and RISC-V owners should try to improve it if its really a problem. In practice all of the other rules on other operations should place constraints on what types the G_TRUNC can end up with. Also the instruction selection code just treats G_TRUNC as a copy so the exact types don't matter.

topperc (approved):
LGTM

llvm/lib/Target/RISCV/GISel/RISCVLegalizerInfo.cpp
davemgreen: Thanks - after looking at X86 I agree. Vectors are more picky but things can be improved going forward.

PR #197378, opened 2026-05-13
[SPIRV][GlobalISel] Remove dependency on legal ruleset

davemgreen:
This fills in always legal rules, to remove the dependency on the legacy ruleset. This is not guaranteed to be all the rules, just the ones that appear in tests.


arsenm (approved):
Could put these in one getActionDefinitionsBuilder set

aobolensk (approved):

PR #197381, opened 2026-05-13
Fix libc building errors for SPIRV target

jinge90:
This is follow-up work of https://github.com/llvm/llvm-project/pull/181049 which aims to build LLVM libc for SPIRV target.
When building LLVM libc for 'spirv64-unknown-unknown' via:
cmake ../llvm -G Ninja -DCMAKE_BUILD_TYPE=Release  -DLLVM_ENABLE_PROJECTS="clang"  -DRUNTIMES_spirv64-unknown-unknown_LLVM_ENABLE_RUNTIMES="libc" -DLLVM_RUNTIME_TARGETS="default;spirv64-unknown-unknown"
Two errors came up:
1. cmake configuration fata error complaining"Target "libc.src.errno.errno" of type UTILITY may not be linked into another target."
2. cmake configuration fata error complaining "Target "libc.src.stdlib.exit" of type UTILITY may not be linked into another target."

For 1, the reason is we enabled some string functions which depends on libc errno but didn't add libc errno into entrypoints, so the dependency is not met in cmake configuration.
For 2, LLVM libc requires LLVM_LIBC_FULL_BUILD to be set and 'startup' will be built for GPU target. The 'startup' component requires stdlib.exit which is not in SPIRV entrypoints.

To fix 1, this patch adds errno to entrypoints. For 2, this patch just skips 'startup' for SPIRV, this is because 'startup'  component also requires RPC and SPIRV target doesn't support RPC.

### DIFF
--- a/libc/CMakeLists.txt
+++ b/libc/CMakeLists.txt
@@ -412,7 +412,7 @@ add_subdirectory(hdr)
 add_subdirectory(src)
 add_subdirectory(utils)
 
-if(LLVM_LIBC_FULL_BUILD)
+if(LLVM_LIBC_FULL_BUILD AND NOT LIBC_TARGET_ARCHITECTURE_IS_SPIRV)
   # The startup system can potentially depend on the library components so add
   # it after the library implementation directories.
   add_subdirectory(startup)
--- a/libc/config/gpu/spirv/entrypoints.txt
+++ b/libc/config/gpu/spirv/entrypoints.txt
@@ -31,6 +31,9 @@ set(TARGET_LIBC_ENTRYPOINTS
     libc.src.ctype.toupper
     libc.src.ctype.toupper_l
 
+    # errno.h entrypoints
+    libc.src.errno.errno
+
     # string.h entrypoints
     libc.src.string.memccpy
     libc.src.string.memchr


jinge90:
Hi, @jhuber6 
Could you help review this patch?
Thanks very much.

lntue (approved):

libc/CMakeLists.txt
```diff
@@ -412,7 +412,7 @@ add_subdirectory(hdr)
 add_subdirectory(src)
 add_subdirectory(utils)
 
-if(LLVM_LIBC_FULL_BUILD)
+if(LLVM_LIBC_FULL_BUILD AND NOT LIBC_TARGET_ARCHITECTURE_IS_SPIRV)
```
jhuber6, line 415: Why can't SPIR-V build the startup code? I thought that there were recent additions to the backend that let us handle global constructors. Guessing it's because it references atexit?

jinge90: Hi, @jhuber6
Yes, it references atexit and exit. For GPU, exit finally depend on rpc which spirv64 target hasn't supported. Currently, spirv64 target only support some 'simple' functions which doesn't need rpc to interact between gpu and host side, for example, functions from <string.h> or <math.h>.

Thanks very much.

jhuber6 (approved):

fineg74:
LGTM to me

PR #197382, opened 2026-05-13
[RewriteStatepointsForGC] Set C calling convention on intrinsic call

wenju-he:
Most intrinsics must use CallingConv::C.
Relates to #197608 which tries to add CallingConv CHECK to IR Verifier.

### DIFF
--- a/llvm/lib/Transforms/Scalar/RewriteStatepointsForGC.cpp
+++ b/llvm/lib/Transforms/Scalar/RewriteStatepointsForGC.cpp
@@ -1553,9 +1553,6 @@ static void CreateGCRelocates(ArrayRef<Value *> LiveVariables,
     CallInst *Reloc = Builder.CreateCall(
         GCRelocateDecl, {StatepointToken, BaseIdx, LiveIdx},
         suffixed_name_or(LiveVariables[i], ".relocated", ""));
-    // Trick CodeGen into thinking there are lots of free registers at this
-    // fake call.
-    Reloc->setCallingConv(CallingConv::Cold);
   }
 }
 
@@ -1835,7 +1832,8 @@ makeStatepointExplicitImpl(CallBase *Call, /* to replace */
         TransitionArgs, DeoptArgs, GCLive, "safepoint_token");
 
     SPCall->setTailCallKind(CI->getTailCallKind());
-    SPCall->setCallingConv(CI->getCallingConv());
+    // Intrinsics must use CallingConv::C.
+    SPCall->setCallingConv(CallingConv::C);
 
     // Set up function attrs directly on statepoint and return attrs later for
     // gc_result intrinsic.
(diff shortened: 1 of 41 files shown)


llvm/lib/Transforms/Scalar/RewriteStatepointsForGC.cpp
```diff
@@ -1550,12 +1550,8 @@ static void CreateGCRelocates(ArrayRef<Value *> LiveVariables,
     Function *GCRelocateDecl = It->second;
 
     // only specify a debug name if we can give a useful one
-    CallInst *Reloc = Builder.CreateCall(
-        GCRelocateDecl, {StatepointToken, BaseIdx, LiveIdx},
-        suffixed_name_or(LiveVariables[i], ".relocated", ""));
-    // Trick CodeGen into thinking there are lots of free registers at this
-    // fake call.
```
arsenm, line 1557: This comment doesn't make much sense, do you see any real codegen changes with cold removed?

wenju-he: the only changes of `check-clang check-llvm` are IR changes in this PR

arsenm: Right, does that mean none of the codegen tests use coldcc on these calls?

wenju-he: Removing coldcc from `@llvm.experimental.gc.statepoint` do change codegen output. So the intrinsic requires coldcc. Take `statepoint-calling-conventions.ll` as example.

Image


On the other hand, removing coldcc from `@llvm.experimental.gc.relocate` doesn't change codegen output. However, given above comment, this intrinsic probably also needs the coldcc.

arsenm: > So the intrinsic requires coldcc. Take statepoint-calling-conventions.ll 

This is not the conclusion I would make. You showed a MIR diff, not the actual codegen. This also doesn't mean the calling convention should be applied to the intrinsic, especially if the use is going to blindly add cold to every instance. You can just handle it in codegen

wenju-he: I run `opt -passes=rewrite-statepoints-for-gc ` and then `llc -mtriple=x86_64 ` on all changed tests in this PR. Two tests (basic.ll and basics.ll) crash in llc step. Diff on .s outputs shows there are only 3 Stackmap CC field changes in statepoint-calling-conventions and deopt-intrinsic-cconv.ll tests, as shown below images. Stackmap CC field changes because the old code propagates coldcc and cc42 calling conventions in the tests. There are no other codegen changes in .s files.

Image
Image

For test statepoint-calling-conventions.ll, its stackmap CC change can be avoided by restoring calling convention from ActualCalledFunction during statepoint intrinsic lowering. Not sure if it makes sense to update the lowering.

PR #197383, opened 2026-05-13
[AArch64][GlobalISel] Fast-path common G_CONSTANT/G_BRCOND/G_FRAME_INDEX regbank mappings

c-rhodes:
Returning the default register-bank mapping directly for these opcodes is a -0.17% compile-time improvement on aarch64-O0-g.

https://llvm-compile-time-tracker.com/compare.php?from=b4aa4d4dcb6f1c8a00d1d1e53d2b353c97ec98b7&to=0779891fc6bf6a01e4f14d3f359e212c6ec52c0d&stat=instructions%3Au

Assisted-by: codex

### DIFF
--- a/llvm/lib/Target/AArch64/GISel/AArch64RegisterBankInfo.cpp
+++ b/llvm/lib/Target/AArch64/GISel/AArch64RegisterBankInfo.cpp
@@ -969,6 +969,26 @@ AArch64RegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
         // We only care about the mapping of the destination for COPY.
         /*NumOperands*/ Opc == TargetOpcode::G_BITCAST ? 2 : 1);
   }
+  case TargetOpcode::G_CONSTANT: {
+    LLT DstTy = MRI.getType(MI.getOperand(0).getReg());
+    TypeSize Size = DstTy.getSizeInBits();
+    if (!(DstTy.isPointer() || (DstTy.isScalar() && Size >= 32 && Size <= 64)))
+      break;
+    // Scalar constants materialize in GPRs.
+    [[fallthrough]];
+  }
+  case TargetOpcode::G_BRCOND:
+  case TargetOpcode::G_FRAME_INDEX: {
+    // Operand 0 is the only banked operand and is mapped to GPR.
+    return getInstructionMapping(
+        DefaultMappingID, /*Cost=*/1,
+        getOperandsMapping(
+            {getValueMapping(
+                 PMI_FirstGPR,
+                 MRI.getType(MI.getOperand(0).getReg()).getSizeInBits()),
+             nullptr}),
+        /*NumOperands=*/2);
+  }
   default:
     break;
   }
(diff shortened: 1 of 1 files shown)


llvm/lib/Target/AArch64/GISel/AArch64RegisterBankInfo.cpp
```diff
@@ -969,6 +969,26 @@ AArch64RegisterBankInfo::getInstrMapping(const MachineInstr &MI) const {
         // We only care about the mapping of the destination for COPY.
         /*NumOperands*/ Opc == TargetOpcode::G_BITCAST ? 2 : 1);
   }
+  case TargetOpcode::G_CONSTANT: {
+    LLT DstTy = MRI.getType(MI.getOperand(0).getReg());
+    TypeSize Size = DstTy.getSizeInBits();
+    if (!(DstTy.isPointer() || (DstTy.isScalar() && Size >= 32 && Size <= 64)))
```
arsenm: demorgan condition

arsenm (approved):

davemgreen (commented):
I'm not super enthused by how this spreads G_CONSTANT over multiple different places with awkward conditions between them. The other two seem fine.

But taking a step back instead of micro-optimizing this function - can regbank select work very differently at O0? To the point I'm wondering if it is needed at all or could work as a pre-step to selection, running in the same pass. There are some operations that need overriding, mostly intrinsics IIRC, but could it just work on the integer vs floating-point types, using that as a pretty direct register-bank mapping? There might need to be some changes to selection to handle the result, but it would clear out an entire pass from the O0 pipeline.

c-rhodes:
> I'm not super enthused by how this spreads G_CONSTANT over multiple different places with awkward conditions between them. The other two seem fine.
> 
> But taking a step back instead of micro-optimizing this function - can regbank select work very differently at O0? To the point I'm wondering if it is needed at all or could work as a pre-step to selection, running in the same pass. There are some operations that need overriding, mostly intrinsics IIRC, but could it just work on the integer vs floating-point types, using that as a pretty direct register-bank mapping? There might need to be some changes to selection to handle the result, but it would clear out an entire pass from the O0 pipeline.

It's a good question, optimising register bank copies at O0 just doesn't seem necessary and perhaps we could just have something more basic like you say. I will look into it, cheers.

c-rhodes:
> > I'm not super enthused by how this spreads G_CONSTANT over multiple different places with awkward conditions between them. The other two seem fine.
> > But taking a step back instead of micro-optimizing this function - can regbank select work very differently at O0? To the point I'm wondering if it is needed at all or could work as a pre-step to selection, running in the same pass. There are some operations that need overriding, mostly intrinsics IIRC, but could it just work on the integer vs floating-point types, using that as a pretty direct register-bank mapping? There might need to be some changes to selection to handle the result, but it would clear out an entire pass from the O0 pipeline.
> 
> It's a good question, optimising register bank copies at O0 just doesn't seem necessary and perhaps we could just have something more basic like you say. I will look into it, cheers.

see #199040

PR #197385, opened 2026-05-13
[lldb][Windows] Cache thread context in NativeRegisterContextWindows_arm64

ayushsahay1837:
Cache thread context in _NativeRegisterContextWindows_arm64_ to improve read performance. Previously, the thread context was retrieved for every read or write operation.

This change intends to lay the groundwork for provisioning debug support for SVE on WoA.

### DIFF
--- a/lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
+++ b/lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
@@ -101,52 +101,62 @@ CreateRegisterInfoInterface(const ArchSpec &target_arch) {
       target_arch, RegisterInfoPOSIX_arm64::eRegsetMaskDefault);
 }
 
-static Status
-GetThreadContextHelper(lldb::thread_t thread_handle, DWORD context_flags,
-                       PCONTEXT &context,
-                       std::shared_ptr<DataBufferHeap> &context_buffer) {
+static Status GetThreadContextLength(DWORD context_flags,
+                                     DWORD &context_length) {
   Log *log = GetLog(WindowsLog::Registers);
   Status error;
-  DWORD context_length = 0;
 
   if (InitializeContext(nullptr, context_flags, nullptr, &context_length)) {
     error = Status::FromErrorString("InitializeContext succeeded unexpectedly");
-    LLDB_LOG(log, "{0} {1}", __FUNCTION__, error);
+    LLDB_LOG(log, "{0}", error);
     return error;
   }
 
   if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
     error = Status(GetLastError(), eErrorTypeWin32);
     LLDB_LOG(log,
-             "{0} InitializeContext failed with unexpected error {1}, expected "
+             "InitializeContext failed with unexpected error {0}, expected "
              "ERROR_INSUFFICIENT_BUFFER",
-             __FUNCTION__, error);
+             error);
     return error;
   }
 
-  if (context_buffer)
-    context_buffer->SetByteSize(context_length);
-  else
-    context_buffer = std::make_shared<DataBufferHeap>(context_length, 0);
+  return error;
+}
+
+static Status GetThreadContextHelper(lldb::thread_t thread_handle,
+                                     DWORD context_flags, PCONTEXT &context,
+                                     DataBufferHeap *context_buffer) {
+  Log *log = GetLog(WindowsLog::Registers);
+  Status error;
+  DWORD context_length = 0;
 
   if (!context_buffer) {
-    error = Status::FromErrorString("Failed to allocate context buffer");
-    LLDB_LOG(log, "{0} {1}", __FUNCTION__, error);
+    error = Status::FromErrorString("context buffer not allocated");
+    LLDB_LOG(log, "{0}", error);
+    return error;
+  }
+
+  error = GetThreadContextLength(context_flags, context_length);
+  if (error.Fail())
+    return error;
+
+  if (context_buffer->SetByteSize(context_length) != context_length) {
+    error = Status::FromErrorString("failed to resize context buffer");
+    LLDB_LOG(log, "{0}", error);
     return error;
   }
 
   if (!InitializeContext(context_buffer->GetBytes(), context_flags, &context,
                          &context_length)) {
     error = Status(GetLastError(), eErrorTypeWin32);
-    LLDB_LOG(log, "{0} InitializeContext failed with error {1}", __FUNCTION__,
-             error);
+    LLDB_LOG(log, "InitializeContext failed with error {0}", error);
     return error;
   }
 
   if (!::GetThreadContext(thread_handle, context)) {
     error = Status(GetLastError(), eErrorTypeWin32);
-    LLDB_LOG(log, "{0} GetThreadContext failed with error {1}", __FUNCTION__,
-             error);
+    LLDB_LOG(log, "GetThreadContext failed with error {0}", error);
     return error;
   }
 
@@ -159,8 +169,7 @@ static Status SetThreadContextHelper(lldb::thread_t thread_handle,
   Status error;
   if (!::SetThreadContext(thread_handle, context)) {
     error = Status(GetLastError(), eErrorTypeWin32);
-    LLDB_LOG(log, "{0} SetThreadContext failed with error {1}", __FUNCTION__,
-             error);
+    LLDB_LOG(log, "SetThreadContext failed with error {0}", error);
     return error;
   }
   return error;
@@ -302,12 +311,12 @@ NativeRegisterContextWindows_arm64::GPRWrite(const uint32_t reg,
   auto cleanup = llvm::make_scope_exit([&]() { m_context = nullptr; });
 
   PCONTEXT context = nullptr;
-  std::shared_ptr<DataBufferHeap> context_buffer;
+  DataBufferHeap context_buffer;
   DWORD context_flags = CONTEXT_CONTROL | CONTEXT_INTEGER;
   auto thread_handle = GetThreadHandle();
 
   Status error = GetThreadContextHelper(thread_handle, context_flags, context,
-                                        context_buffer);
+                                        &context_buffer);
   if (error.Fail())
     return error;
 
@@ -527,12 +536,12 @@ NativeRegisterContextWindows_arm64::FPRWrite(const uint32_t reg,
   auto cleanup = llvm::make_scope_exit([&]() { m_context = nullptr; });
 
   PCONTEXT context = nullptr;
-  std::shared_ptr<DataBufferHeap> context_buffer;
+  DataBufferHeap context_buffer;
   DWORD context_flags = CONTEXT_CONTROL | CONTEXT_FLOATING_POINT;
   auto thread_handle = GetThreadHandle();
 
   Status error = GetThreadContextHelper(thread_handle, context_flags, context,
-                                        context_buffer);
+                                        &context_buffer);
   if (error.Fail())
     return error;
 
@@ -722,20 +731,12 @@ Status NativeRegisterContextWindows_arm64::ReadAllRegisterValues(
 
   if (!m_context_buffer) {
     error = Status::FromErrorString("register context buffer is not available");
-    LLDB_LOG(log, "{0} {1}", __FUNCTION__, error);
+    LLDB_LOG(log, "{0}", error);
     return error;
   }
 
-  const size_t data_size = m_context_buffer->GetByteSize();
-  data_sp = std::make_shared<DataBufferHeap>(data_size, 0);
-
-  if (!data_sp) {
-    error = Status::FromErrorString("failed to allocate register data buffer");
-    LLDB_LOG(log, "{0} {1}", __FUNCTION__, error);
-    return error;
-  }
-
-  ::memcpy(data_sp->GetBytes(), m_context_buffer->GetBytes(), data_size);
+  data_sp = std::make_shared<DataBufferHeap>(m_context_buffer->GetBytes(),
+                                             m_context_buffer->GetByteSize());
 
   return error;
 }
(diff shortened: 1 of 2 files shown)


lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
```diff
@@ -101,35 +101,67 @@ CreateRegisterInfoInterface(const ArchSpec &target_arch) {
...
-  if (!::GetThreadContext(thread_handle, context_ptr)) {
+  if (InitializeContext(nullptr, context_flags, nullptr, &context_length)) {
+    error = Status::FromErrorString("InitializeContext succeeded unexpectedly");
+    LLDB_LOG(log, "{0} {1}", __FUNCTION__, error);
+    return error;
+  }
+
+  if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
+    error = Status(GetLastError(), eErrorTypeWin32);
+    LLDB_LOG(log, "{0} InitializeContext failed with error {1}", __FUNCTION__,
```
DavidSpickett: I thought we had a log macro that put the function name in automatically.

Also:
"with unexpected error { }, expected ERROR_...."

Would be a bit better because this is meant to "fail" but in one specific way.

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
```diff
@@ -101,35 +101,67 @@ CreateRegisterInfoInterface(const ArchSpec &target_arch) {
       target_arch, RegisterInfoPOSIX_arm64::eRegsetMaskDefault);
 }
 
-static Status GetThreadContextHelper(lldb::thread_t thread_handle,
-                                     PCONTEXT context_ptr,
-                                     const DWORD control_flag) {
+static Status
+GetThreadContextHelper(lldb::thread_t thread_handle, DWORD context_flags,
+                       PCONTEXT &context,
+                       std::shared_ptr<DataBufferHeap> &context_buffer) {
```
DavidSpickett: Pass shared_ptr by copy.

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
```diff
@@ -267,11 +297,15 @@ Status NativeRegisterContextWindows_arm64::GPRRead(const uint32_t reg,
 Status
 NativeRegisterContextWindows_arm64::GPRWrite(const uint32_t reg,
                                              const RegisterValue &reg_value) {
-  ::CONTEXT tls_context;
-  DWORD context_flag = CONTEXT_CONTROL | CONTEXT_INTEGER;
+  auto cleanup = llvm::make_scope_exit([&]() { m_context = nullptr; });
```
DavidSpickett, line 311: Does this also need to free the context_buffer, or will that get resized later anyway?

It's not likely that the context size changes anyway, so keeping it makes sense I guess.

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
```diff
@@ -677,73 +712,143 @@ Status NativeRegisterContextWindows_arm64::WriteRegister(
...
+  if (InitializeContext(nullptr, context_flags, nullptr, &context_length)) {
+    error = Status::FromErrorString("InitializeContext succeeded unexpectedly");
+    LLDB_LOG(log, "{0} {1}", __FUNCTION__, error);
     return error;
   }
 
-  if (data_sp->GetByteSize() != data_size) {
+  if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
+    error = Status(GetLastError(), eErrorTypeWin32);
+    LLDB_LOG(log, "{0} InitializeContext failed with error {1}", __FUNCTION__,
```
DavidSpickett: "failed with error {}, expected ERROR_..."

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
```diff
@@ -677,73 +712,143 @@ Status NativeRegisterContextWindows_arm64::WriteRegister(
...
+  PCONTEXT context = nullptr;
+  std::shared_ptr<DataBufferHeap> context_buffer;
+  error = GetThreadContextHelper(GetThreadHandle(), context_flags, context,
+                                 context_buffer);
+  if (error.Fail())
+    return error;
+
+  ::memcpy(context_buffer->GetBytes(), data_sp->GetBytes(), context_length);
+
+  error = SetThreadContextHelper(GetThreadHandle(), context);
```
DavidSpickett:
```
return SetThread...;
```

DavidSpickett (commented):
One thing to watch out for is what happens if Windows were to accept a value and then modify it in some way. For instance, if you had set some cpsr bits that Windows reset before applying it to the real register.

Maybe Windows would error before writing anything in that case, but be careful that you don't keep around what might be an inaccurate cache.

I think you're safe from that here but I didn't look too closely.

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
ayushsahay1837: > I thought we had a log macro that put the function name in automatically.

I might have missed it. I lightly reviewed the sources in _~/lldb/source/Plugins/Process_ but didn't find a macro like that. I'd really appreciate it if you could point me to it.

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
ayushsahay1837: Are you suggesting _context_buffer_ to be passed by value to clarify ownership semantics? I'm asking because _GetThreadContextHelper_ reseats _context_buffer_ if it's unallocated –

```
  if (context_buffer)
    context_buffer->SetByteSize(context_length);
  else
    context_buffer = std::make_shared<DataBufferHeap>(context_length, 0);
```

IIUC, then passing _context_buffer_ by value would mean that if it's unallocated, then _GetThreadContextHelper_ wouldn't update the caller's copy. Alternatively, we could require callers to allocate _context_buffer_ before invoking _GetThreadContextHelper_ but I feel that it would be preferable not to depend on callers for this allocation. What do you think?

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
ayushsahay1837: That's the intention. We could have conservatively freed _m_context_buffer_ as well but like you mentioned, we don't anticipate the context length changing frequently, if at all. So, there might be value in just resizing and zeroing it out in _GetThreadContextHelper_ rather than resetting it.

ayushsahay1837:
> One thing to watch out for is what happens if Windows were to accept a value and then modify it in some way. For instance, if you had set some cpsr bits that Windows reset before applying it to the real register.
> 
> Maybe Windows would error before writing anything in that case, but be careful that you don't keep around what might be an inaccurate cache.
> 
> I think you're safe from that here but I didn't look too closely.

The cleanup lambdas invalidate _m_context_ by setting it to _nullptr_ on every write. So, subsequent reads should fetch fresh contexts reflecting what was actually written to the hardware registers. Is this what you'd meant, or am I missing anything?

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
DavidSpickett: Turns out that it's not a specific macro you have to use. It's that you enable the log with this option:
```
       -F ( --file-function )
            Prepend the names of files and function that generate the logs.
```
And then you get file and function.

```
(lldb) log enable -F lldb commands
(lldb) sdf
CommandInterpreter.cpp:HandleCommand                         Processing command: sdf
CommandInterpreter.cpp:HandleCommand                         HandleCommand, cmd_obj : '<not found>'
CommandInterpreter.cpp:HandleCommand                         HandleCommand, (revised) command_string: 'sd
f'
CommandInterpreter.cpp:HandleCommand                         HandleCommand, wants_raw_input:'False'
CommandInterpreter.cpp:HandleCommand                         HandleCommand, command did not succeed
error: 'sdf' is not a valid command
(lldb) log disable lldb commands
<...>
(lldb) log enable lldb commands
(lldb) asdfdf
lldb             Processing command: asdfdf
lldb             HandleCommand, cmd_obj : '<not found>'
lldb             HandleCommand, (revised) command_string: 'asdfdf'
lldb             HandleCommand, wants_raw_input:'False'
lldb             HandleCommand, command did not succeed
error: 'asdfdf' is not a valid command
```

So please remove the function names in the format strings.

I wouldn't be surprised if we have plenty of log calls that add the function name anyway, but that's not the way we're "supposed" to do it :)

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
DavidSpickett: Up to you when it comes to the ownership of the buffer itself.

I only mean to say that shared pointers being passed by reference kinda bypasses the way shared pointer works. So using a copy of the shared pointer increases the refcount as one would expect.

Which is not that important mechnically here, it's not a problem in practice, but pass by copy of *the shared pointer* is the standard thing to do and is less weird to anyone reading this code.

DavidSpickett: In other words: passing a shared pointer by ref could suggest you deliberately don't want to change the refcount.

Why would you not want to do that? No idea, but if I was reading this code for the first time, I wouldn't have the context to know that it wasn't important.

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
DavidSpickett: Yes that sounds good. It will avoid an allocation in most cases, but reallocate if for example SVE resized since last time.

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
```diff
@@ -677,73 +714,143 @@ Status NativeRegisterContextWindows_arm64::WriteRegister(
...
+
+  auto cleanup = llvm::make_scope_exit([&]() {
+    if (error.Fail())
+      m_context = nullptr;
+  });
+
+  error = GetThreadContextHelper(GetThreadHandle(), context_flags, m_context,
+                                 m_context_buffer);
+
+  return error;
```
DavidSpickett: Two things:
1. you can directly `return GetThread....`
2. Do you need to use a scope exit here? Would this be the same thing:
```
error = GetThread....;
if (error.Fail())
  m_context ...;
return error;
```
Given that we don't expect GetThread... to throw an exception (https://llvm.org/docs/CodingStandards.html#do-not-use-rtti-or-exceptions).

Maybe it's leftover from a refactor or Windows APIs might throw exceptions (I don't think so?), but anyway, I think you could do without the scope guard here.

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
```diff
@@ -677,73 +714,143 @@ Status NativeRegisterContextWindows_arm64::WriteRegister(
...
+  }
+
+  if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
+    error = Status(GetLastError(), eErrorTypeWin32);
+    LLDB_LOG(log,
+             "{0} InitializeContext failed with unexpected error {1}, expected "
+             "ERROR_INSUFFICIENT_BUFFER",
+             __FUNCTION__, error);
+    return error;
+  }
```
DavidSpickett: You do this set of calls a few times, I wonder if another helper function would be good for this.

Then this and GetThreadContextHelper and others can use it.

DavidSpickett:
> The cleanup lambdas invalidate m_context by setting it to nullptr on every write. So, subsequent reads should fetch fresh contexts reflecting what was actually written to the hardware registers. Is this what you'd meant, or am I missing anything?

That's what I meant, and you've covered it all. No worries on that.

DavidSpickett:
With my most recent comments addressed I expect I won't have any more suggestions. If so I will approve but I would like us to wait a bit for other reviewers because A: they might have different input and B: they should read this so they can more easily review your future PRs.

(as I personally am unlikely to be a user of this code any time soon, but they are)

If no one chimes in, we'll push forward anyway.

lldb/source/Plugins/Process/Windows/Common/NativeRegisterContextWindows_arm64.cpp
ayushsahay1837: Agreed. After taking a closer look, I realized that _GetThreadContextHelper_ doesn’t actually need to own the context buffer; it just needs to resize and populate it. So, I've updated _GetThreadContextHelper_ to take a _DataBufferHeap *_ and left the management of the _DataBufferHeap_ object's lifetime to the callers. What do you think?

DavidSpickett: I think it looks good :)

DavidSpickett (approved):
LGTM, but let @charles-zablit take a look before landing.

I should have thought of them earlier but they will have more "taste" than I when it comes to Windows, and more ability to help if it goes wrong.

ayushsahay1837:
@charles-zablit, please let me know if you have any suggestion or feedback. I'll be happy to hear any recommendation that you might have.

charles-zablit (approved):
LGTM!

In follow ups, we will need to be careful with invalidating that cache on every resume.

PR #197386, opened 2026-05-13
[AMDGPU][GCNPreRAOptimizations] Reduce BVH premature reuse

perlfu:
Add implicit uses to ds_bvh_stack instructions to avoid reuse of VGPRs allocated to bvh_intersect_ray results prior to ds_bvh_stack. This reduces likelihood of a premature s_wait_bvhcnt occuring due to partial reallocation of unused bvh_intersect_ray results registers.


llvm/lib/Target/AMDGPU/GCNPreRAOptimizations.cpp
```diff
@@ -238,6 +240,65 @@ GCNPreRAOptimizationsPass::run(MachineFunction &MF,
   return PreservedAnalyses::all();
 }
 
+void GCNPreRAOptimizationsImpl::hintTrue16Copy(const MachineInstr &MI) {
```
arsenm, line 243: This looks overcomplicated but I see it was just moved

llvm/lib/Target/AMDGPU/GCNPreRAOptimizations.cpp
```diff
@@ -238,6 +240,65 @@ GCNPreRAOptimizationsPass::run(MachineFunction &MF,
...
+  if (IsDst16Bit && MRI->getRegClass(Src) == &AMDGPU::VGPR_32RegClass)
+    MRI->setRegAllocationHint(Dst, AMDGPURI::Size16, Src);
+}
+
+bool GCNPreRAOptimizationsImpl::optimizeBVHStack(MachineInstr &MI) {
+  SmallVector<Register> UseRegs;
+
+  // Find all BVH sources for this DS_BVH_STACK instruction.
+  for (MachineOperand &Use : MI.uses()) {
+    if (!Use.isReg() || Use.isImplicit())
```
arsenm: I believe the isImplicit check is redundant with using uses instead of all_uses. Also, does this really support non-register operands? Can you simplify this to just inspecting data0/data1?

perlfu: Yes, done.

llvm/lib/Target/AMDGPU/GCNPreRAOptimizations.cpp
perlfu: Yes, just repurposed the existing iteration over instruction.

llvm/lib/Target/AMDGPU/GCNPreRAOptimizations.cpp
```diff
@@ -258,34 +319,27 @@ bool GCNPreRAOptimizationsImpl::run(MachineFunction &MF) {
     Changed |= processReg(Reg);
   }
 
-  if (!ST.useRealTrue16Insts())
+  const bool HasBVHStack = ST.hasImageInsts() && ST.hasBVHDualAndBVH8Insts();
```
arsenm: I'd assume the hasImageInsts is redundant?

arsenm (approved):

llvm/lib/Target/AMDGPU/GCNPreRAOptimizations.cpp
perlfu: True, thanks. Hang over from an earlier iteration.

PR #197389, opened 2026-05-13
[lldb] Fix wrong buffer size when fetching Objective-C classes

Teemperor:
LLDB calls objc_getRealizedClassList_trylock to fetch the list of realized Objective-C classes.

Jim spotted that we currently pass the buffer length in *bytes*, when actually this API takes the buffer length in number of elements. This causes that the Objective-C runtime write more memory that we allocated for it. This can cause that the function calling expression crashes and leaves the Objective-C runtime mutex locked.

### DIFF
--- a/lldb/source/Plugins/LanguageRuntime/ObjC/AppleObjCRuntime/AppleObjCRuntimeV2.cpp
+++ b/lldb/source/Plugins/LanguageRuntime/ObjC/AppleObjCRuntime/AppleObjCRuntimeV2.cpp
@@ -2245,7 +2245,7 @@ AppleObjCRuntimeV2::DynamicClassInfoExtractor::UpdateISAToDescriptorMap(
 
   if (class_buffer_addr != LLDB_INVALID_ADDRESS) {
     arguments.GetValueAtIndex(index++)->GetScalar() = class_buffer_addr;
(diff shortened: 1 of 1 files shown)


Teemperor:
It's not obvious from the code, but there is only one of the three potential calling functions where we pass a size. The one change is this one:

```
uint32_t
__lldb_apple_objc_v2_get_dynamic_class_info3(void *gdb_objc_realized_classes_ptr,
                                             void *class_infos_ptr,
                                             uint32_t class_infos_byte_size,
                                             void *class_buffer,
                                             uint32_t class_buffer_len,
                                             uint32_t should_log)
```

And the other two we never call with a buffer size as they don't take one:

```
uint32_t
__lldb_apple_objc_v2_get_dynamic_class_info2(void *gdb_objc_realized_classes_ptr,
                                             void *class_infos_ptr,
                                             uint32_t class_infos_byte_size,
                                             uint32_t should_log)
```

and 

```
uint32_t
__lldb_apple_objc_v2_get_dynamic_class_info (void *gdb_objc_realized_classes_ptr,
                                             void *class_infos_ptr,
                                             uint32_t class_infos_byte_size,
                                             uint32_t should_log)
```

Also I'm not sure how we would test this aside from creating a test program with a really large amount of instantiated Objective-C classes.

JDevlieghere (approved):
LGTM

PR #197399, opened 2026-05-13
[CIR][AMDGPU] Add support for AMDGCN trig_preop builtins

ayokunle321:
Adds codegen for the following AMDGCN trigonometric pre-operation builtins:

- __builtin_amdgcn_trig_preop (double)
- __builtin_amdgcn_trig_preopf (float)

These are lowered to the corresponding `llvm.amdgcn.trig.preop` intrinsic.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenBuiltinAMDGPU.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenBuiltinAMDGPU.cpp
@@ -91,6 +91,16 @@ static mlir::Value emitLogbBuiltin(CIRGenFunction &cgf, const CallExpr *e,
   return res;
 }
 
+// Emit an intrinsic that has 1 float or double operand, and 1 integer.
+static mlir::Value emitFPIntBuiltin(CIRGenFunction &cgf, const CallExpr *e,
+                                    llvm::StringRef intrinsicName) {
+  mlir::Value src0 = cgf.emitScalarExpr(e->getArg(0));
+  mlir::Value src1 = cgf.emitScalarExpr(e->getArg(1));
+  return cgf.getBuilder().emitIntrinsicCallOp(cgf.getLoc(e->getExprLoc()),
+                                              intrinsicName, src0.getType(),
+                                              mlir::ValueRange{src0, src1});
+}
+
 std::optional<mlir::Value>
 CIRGenFunction::emitAMDGPUBuiltinExpr(unsigned builtinId,
                                       const CallExpr *expr) {
@@ -202,10 +212,7 @@ CIRGenFunction::emitAMDGPUBuiltinExpr(unsigned builtinId,
   }
   case AMDGPU::BI__builtin_amdgcn_trig_preop:
   case AMDGPU::BI__builtin_amdgcn_trig_preopf: {
-    cgm.errorNYI(expr->getSourceRange(),
-                 std::string("unimplemented AMDGPU builtin call: ") +
-                     getContext().BuiltinInfo.getName(builtinId));
-    return mlir::Value{};
+    return emitFPIntBuiltin(*this, expr, "amdgcn.trig.preop");
   }
   case AMDGPU::BI__builtin_amdgcn_rcp:
   case AMDGPU::BI__builtin_amdgcn_rcpf:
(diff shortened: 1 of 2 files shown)


andykaylor (approved):
lgtm

andykaylor:
@jhuber6 @arsenm This is one of many PR's that @ayokunle321 has put up to add support for AMDGCN builtins. They're pretty straightforward, so I've been approving them, but I wondered if one of you (or someone else you can recommend with a stake in this target) would want to take a look at what's happening?

jhuber6:
> @jhuber6 @arsenm This is one of many PR's that @ayokunle321 has put up to add support for AMDGCN builtins. They're pretty straightforward, so I've been approving them, but I wondered if one of you (or someone else you can recommend with a stake in this target) would want to take a look at what's happening?

These are likely just straight ports of the existing clang builtins, I don't think they require much AMD knowledge beyond checking that the IR looks the same.

clang/lib/CIR/CodeGen/CIRGenBuiltinAMDGPU.cpp
```diff
@@ -91,6 +91,16 @@ static mlir::Value emitLogbBuiltin(CIRGenFunction &cgf, const CallExpr *e,
   return res;
 }
 
+// Emit an intrinsic that has 1 float or double operand, and 1 integer.
+static mlir::Value emitFPIntBuiltin(CIRGenFunction &cgf, const CallExpr *e,
+                                    llvm::StringRef intrinsicName) {
+  mlir::Value src0 = cgf.emitScalarExpr(e->getArg(0));
+  mlir::Value src1 = cgf.emitScalarExpr(e->getArg(1));
+  return cgf.getBuilder().emitIntrinsicCallOp(
+      cgf.getLoc(e->getExprLoc()), "amdgcn.ds.swizzle", src0.getType(),
```
shiltian: Is using a string instead of IID the right approach in CIR?

ayokunle321:
> > @jhuber6 @arsenm This is one of many PR's that @ayokunle321 has put up to add support for AMDGCN builtins. They're pretty straightforward, so I've been approving them, but I wondered if one of you (or someone else you can recommend with a stake in this target) would want to take a look at what's happening?
> 
> These are likely just straight ports of the existing clang builtins, I don't think they require much AMD knowledge beyond checking that the IR looks the same.

@andykaylor In this case, are the PRs good to be merged after just your approval?

andykaylor:
> > @jhuber6 @arsenm This is one of many PR's that @ayokunle321 has put up to add support for AMDGCN builtins. They're pretty straightforward, so I've been approving them, but I wondered if one of you (or someone else you can recommend with a stake in this target) would want to take a look at what's happening?
> 
> These are likely just straight ports of the existing clang builtins, I don't think they require much AMD knowledge beyond checking that the IR looks the same.

Yes, I wasn't so much concerned about the need for AMD-specific knowledge as being sure that the target maintainers had a chance to offer opinions on general direction like test organization. The comment @arsenm had about an unnecessary header on https://github.com/llvm/llvm-project/pull/197806 is an example.

clang/lib/CIR/CodeGen/CIRGenBuiltinAMDGPU.cpp
andykaylor: CIR isn't entirely able to decide this. The LLVM dialect uses strings, so we followed suit. See https://mlir.llvm.org/docs/Dialects/LLVM/#llvmcall_intrinsic-llvmcallintrinsicop.

andykaylor:
> @andykaylor In this case, are the PRs good to be merged after just your approval?

Yes, I suppose so, but be sure to apply any input you get from others, like the question about including headers in the test, to all your changes.

clang/lib/CIR/CodeGen/CIRGenBuiltinAMDGPU.cpp
shiltian: I c, thanks for the info.

clang/lib/CIR/CodeGen/CIRGenBuiltinAMDGPU.cpp
```diff
@@ -118,6 +118,16 @@ static mlir::Value emitLogbBuiltin(CIRGenFunction &cgf, const CallExpr *e,
...
```
ranapratap55: declares `intrinsicName` but never uses it. `ds.swizzle` is being used here. Recheck and update.

ayokunle321: Thanks for the catch!

I mirrored OG but I was wondering if there was even a point of this function given `emitBuiltinWithOneOverloadedType<2>(expr, intrinsicName).getValue()` would give the same result. They both use `src0.getType()` as the result type.

ranapratap55 (approved):
LGTM.

PR #197402, opened 2026-05-13
[CHERI] Add a Support utility for determining alignment requirements of CHERI capabilities.

resistor:
On CHERI systems with compressed bounds representations (which is all of them that anyone cares about today), one of the tradeoffs to achieve that compression is a requirement for larger allocations to be more highly aligned. This impacts both code generation and linking in places where globals need to be aligned and/or padded based on this requirement. The specific alignment requirements vary by capability format.

### DIFF
--- a/llvm/include/llvm/Support/CHERICapabilityFormat.h
+++ b/llvm/include/llvm/Support/CHERICapabilityFormat.h
@@ -1,0 +1,53 @@
+//===--- CHERICapabilityFormat.h --------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_SUPPORT_CHERICAPABILITYFORMAT_H
+#define LLVM_SUPPORT_CHERICAPABILITYFORMAT_H
+
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/Support/Alignment.h"
+
+namespace llvm {
+
+template <typename Derived, typename AddressType>
+struct CHERICapabilityFormatBase {
+  CHERICapabilityFormatBase() = delete;
+
+  static constexpr AddressType AddressMask = ~static_cast<AddressType>(0);
+
+  /// Returns the "alignment mask" for an allocation of size \p Length. This
+  /// mask is 0 where the capability format alignment requires the
+  /// address to be 0, and 1 otherwise.
+  LLVM_ABI static AddressType getAlignmentMask(AddressType Length);
+
+  /// Returns the required alignment for an allocation of size \p Length.
+  LLVM_ABI static Align getRequiredAlignment(AddressType Length);
+
+  /// Returns \p Length rounded up to the nearest representable allocation
+  /// length.
+  LLVM_ABI static AddressType getRepresentableLength(AddressType Length);
+};
+
+template <typename AddressType, unsigned MW, unsigned MAX_E>
+struct RVYCapabilityFormat
+    : public CHERICapabilityFormatBase<
+          RVYCapabilityFormat<AddressType, MW, MAX_E>, AddressType> {
+  LLVM_ABI static AddressType getAlignmentMask(uint64_t Length);
+};
+
+using RV32YCapabilityFormat = RVYCapabilityFormat<uint32_t, 10, 24>;
+using RV64YCapabilityFormat = RVYCapabilityFormat<uint64_t, 14, 24>;
+
+struct CHERIoTCapabilityFormat
+    : public CHERICapabilityFormatBase<CHERIoTCapabilityFormat, uint32_t> {
+  LLVM_ABI static uint32_t getAlignmentMask(uint32_t Length);
+};
+
+} // namespace llvm
+
+#endif
--- a/llvm/lib/Support/CHERICapabilityFormat.cpp
+++ b/llvm/lib/Support/CHERICapabilityFormat.cpp
@@ -1,0 +1,77 @@
+//===- CHERICapabilityFormat.cpp ------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "llvm/Support/CHERICapabilityFormat.h"
+#include "llvm/ADT/bit.h"
+
+namespace llvm {
+
+template <typename Derived, typename AddressType>
+Align CHERICapabilityFormatBase<Derived, AddressType>::getRequiredAlignment(
+    AddressType Length) {
+  return Align((~Derived::getAlignmentMask(Length) + 1) & AddressMask);
+}
+
+template <typename Derived, typename AddressType>
+AddressType
+CHERICapabilityFormatBase<Derived, AddressType>::getRepresentableLength(
+    AddressType Length) {
+  AddressType Mask = Derived::getAlignmentMask(Length);
+  return (Length + ~Mask) & Mask;
+}
+
+template <typename AddressType, unsigned MW, unsigned MAX_E>
+AddressType
+RVYCapabilityFormat<AddressType, MW, MAX_E>::getAlignmentMask(uint64_t Length) {
+  static constexpr unsigned int IE_TAKE_BITS = 3;
+
+  if (Length == 0)
+    return RVYCapabilityFormat::AddressMask;
+
+  // Extract bits that overflow the uncompressed mantissa window.
+  uint64_t Slice = static_cast<uint64_t>(Length) >> (MW - 1);
+  unsigned int E = 64 - llvm::countl_zero(Slice);
+  // We use internal exponent if length overflows OR the denormal boundary
+  // bit is set.
+  bool IE = (E != 0) || ((static_cast<uint64_t>(Length) >> (MW - 2)) & 1);
+  // Include bits used by the internal exponent for the shift value.
+  unsigned int Eprime = IE ? (E + IE_TAKE_BITS) : 0;
+
+  assert(E <= MAX_E && "Raw exponent exceeds architecture maximum");
+  assert(Eprime <= sizeof(AddressType) * 8 &&
+         "Shift amount exceeds integer width");
+
+  // Left-shift ~0 to mask out the lost precision bits
+  return RVYCapabilityFormat::AddressMask << Eprime;
+}
+
+template struct CHERICapabilityFormatBase<RVYCapabilityFormat<uint32_t, 10, 24>,
+                                          uint32_t>;
+template struct CHERICapabilityFormatBase<RVYCapabilityFormat<uint64_t, 14, 24>,
+                                          uint64_t>;
+template struct RVYCapabilityFormat<uint32_t, 10, 24>;
+template struct RVYCapabilityFormat<uint64_t, 14, 24>;
+
+uint32_t CHERIoTCapabilityFormat::getAlignmentMask(uint32_t Length) {
+  // Per section 7.13.4 and table 7.4 in the v1.0 CHERIoT specification.
+  constexpr uint32_t NINE_SET_BITS = 511;
+  uint32_t E;
+  if (Length > NINE_SET_BITS << 14)
+    E = 24;
+  else {
+    E = Length > NINE_SET_BITS ? 32 - llvm::countl_zero(Length) - 9 : 0;
+    if (Length > NINE_SET_BITS << E)
+      ++E;
+    assert(E <= 14 && "CHERIoT capabilities cannot encode E between 14 and 24");
+  }
+  return CHERIoTCapabilityFormat::AddressMask << E;
+}
+
+template struct CHERICapabilityFormatBase<CHERIoTCapabilityFormat, uint32_t>;
+
+} // namespace llvm
--- a/llvm/unittests/Support/CHERICapabilityFormatTest.cpp
+++ b/llvm/unittests/Support/CHERICapabilityFormatTest.cpp
@@ -1,0 +1,91 @@
+#include "llvm/Support/CHERICapabilityFormat.h"
+#include "gtest/gtest.h"
+
+using namespace llvm;
+
+namespace {
+
+TEST(CHERICapabilityFormat, RV32Y) {
+  using RV32Y = RV32YCapabilityFormat;
+
+  EXPECT_EQ(RV32Y::AddressMask, 0xFFFFFFFF);
+
+  // Lengths up to 255 are byte-aligned.
+  for (uint64_t Len = 1; Len <= 255; ++Len) {
+    EXPECT_EQ(RV32Y::getRepresentableLength(Len), Len);
+    EXPECT_EQ(RV32Y::getRequiredAlignment(Len), 1);
+    EXPECT_EQ(RV32Y::getAlignmentMask(Len), 0xFFFFFFFF);
+  }
+
+  // Lengths up to 511 are 8-byte-aligned.
+  for (uint64_t Len = 256; Len <= 511; ++Len) {
+    EXPECT_EQ(RV32Y::getRepresentableLength(Len), (Len + 7) & 0xFFFFFFF8);
+    EXPECT_EQ(RV32Y::getRequiredAlignment(Len), 8);
+    EXPECT_EQ(RV32Y::getAlignmentMask(Len), 0xFFFFFFF8);
+  }
+
+  // Lengths up to 1023 are 16-byte-aligned.
+  for (uint64_t Len = 512; Len <= 1023; ++Len) {
+    EXPECT_EQ(RV32Y::getRepresentableLength(Len), (Len + 15) & 0xFFFFFFF0);
+    EXPECT_EQ(RV32Y::getRequiredAlignment(Len), 16);
+    EXPECT_EQ(RV32Y::getAlignmentMask(Len), 0xFFFFFFF0);
+  }
+}
+
+TEST(CHERICapabilityFormat, RV64Y) {
+  using RV64Y = RV64YCapabilityFormat;
+
+  EXPECT_EQ(RV64Y::AddressMask, 0xFFFFFFFFFFFFFFFF);
+
+  // Lengths up to 4095 are byte-aligned.
+  for (uint64_t Len = 1; Len <= 4095; ++Len) {
+    EXPECT_EQ(RV64Y::getRepresentableLength(Len), Len);
+    EXPECT_EQ(RV64Y::getRequiredAlignment(Len), 1);
+    EXPECT_EQ(RV64Y::getAlignmentMask(Len), 0xFFFFFFFFFFFFFFFF);
+  }
+
+  // Lengths up to 8191 are 8-byte-aligned.
+  for (uint64_t Len = 4096; Len <= 8191; ++Len) {
+    assert(RV64Y::getRepresentableLength(Len) ==
+           ((Len + 7) & 0xFFFFFFFFFFFFFFF8));
+    EXPECT_EQ(RV64Y::getRequiredAlignment(Len), 8);
+    EXPECT_EQ(RV64Y::getAlignmentMask(Len), 0xFFFFFFFFFFFFFFF8);
+  }
+
+  // Lengths up to 16383 are 16-byte-aligned.
+  for (uint64_t Len = 8192; Len <= 16383; ++Len) {
+    EXPECT_EQ(RV64Y::getRepresentableLength(Len),
+              (Len + 15) & 0xFFFFFFFFFFFFFFF0);
+    EXPECT_EQ(RV64Y::getRequiredAlignment(Len), 16);
+    EXPECT_EQ(RV64Y::getAlignmentMask(Len), 0xFFFFFFFFFFFFFFF0);
+  }
+}
+
+TEST(CHERICapabilityFormat, CHERIoT) {
+  using CHERIoT = CHERIoTCapabilityFormat;
+
+  EXPECT_EQ(CHERIoT::AddressMask, 0xFFFFFFFF);
+
+  // Lengths up to 511 are byte-aligned.
+  for (uint64_t Len = 1; Len <= 511; ++Len) {
+    EXPECT_EQ(CHERIoT::getRepresentableLength(Len), Len);
+    assert(CHERIoT::getRequiredAlignment(Len) == 1);
+    EXPECT_EQ(CHERIoT::getAlignmentMask(Len), 0xFFFFFFFF);
+  }
+
+  // Lengths up to 1022 are 2-byte-aligned.
+  for (uint64_t Len = 512; Len <= 1022; ++Len) {
+    EXPECT_EQ(CHERIoT::getRepresentableLength(Len), (Len + 1) & 0xFFFFFFFE);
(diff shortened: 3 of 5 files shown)


resistor:
@jrtc27 @arichardson 

I'm particularly open to ideas on unit tests that aren't essentially copy-pasting the LUTs.

resistor:
Ping

jrtc27:
Yeah so at least for normal CHERI it's just maths, we don't need weird LUTs like you put in your fork.

jrtc27:
I also feel it should probably be more target-specific?

jrtc27:
In CHERI LLVM it's in TLI and TLOF. I can see an argument for having it exposed to the frontend and middle-end, so you want something more Support-y, but I think you still want one implementation per target, with a possible dynamic dispatch layer on top of it.

arichardson:
I agree that we should avoid the lookup table, but also shouldn't be importing the full cheri-compressed-cap just for this one function that is needed. 

We just need to calculate the exponent for the given length. If we are in the IE case we return -1 shifted by exponent plus the number of bits taken from the mantissa.

resistor:
> In CHERI LLVM it's in TLI and TLOF. I can see an argument for having it exposed to the frontend and middle-end, so you want something more Support-y, but I think you still want one implementation per target, with a possible dynamic dispatch layer on top of it.

In the CHERIoT tree, we're additionally using it in LLD and in the clang static analyzer, which is why it's in Support.

While it _can_ be made to dispatch through the target, I'm not seeing why that would be particularly desirable. Capability formats aren't intrinsically tied to a target (other than address size), and since the only target being upstreamed right now is RISCV, dispatching on that doesn't contribute anything to the decision tree.

resistor:
> I agree that we should avoid the lookup table, but also shouldn't be importing the full cheri-compressed-cap just for this one function that is needed.
> 
> We just need to calculate the exponent for the given length. If we are in the IE case we return -1 shifted by exponent plus the number of bits taken from the mantissa.

I started looking into this, with a prototype implementation here, which is derived as directly as I could from the CHERI v9 spec: https://gist.github.com/resistor/414e16427cca02aeb727fb1c646af90b

This ran into several issues: 
* There are constants 3 and 12 that appear in a few places. The spec doesn't make it clear how these are derived from parametric values like MW.
* AFAICT CHERIv9 never actually specifies MW for 32-bit addresses. Is MW=14 there as well, by omission?
* Specification language like `C is an additional carry bit from rounding up the truncated (to E_initial + 3 + MW−4 bits) length` is very hard to interpret, "rounding up" not being an operation I would normally expect to produce a carry bit.
* My understanding is that the RVY format does not exactly match the CHERIv9 format?
* My understanding is that that the "round up representable length" portion will need to be different on CHERIoT, so we'll end up with two significantly different paths.

resistor:
I pushed a change here that reimplements these without LUTs (though the CHERIoT on is still LUT-adjacent, as that is actually how it is specified). The CHERI64 case still doesn't work.

llvm/include/llvm/Support/CHERICapabilityFormat.h
arichardson:
```suggestion
class RVYCapabilityFormatImpl : public CHERICapabilityFormatImpl {
```

arichardson:
```suggestion
  static constexpr unsigned MW = 14;
  static constexpr unsigned IETakenBits = 3;
```

arichardson: Should probably also avoid the hardcoded 3 and give it a descriptive name.

arichardson:
I played around with this a bit today and came up with the following code. The RVY format logic is based on simplifying the cheri-compressed-cap logic just for CRAM.

Note: since I am not too familiar with the CHERIoT format, I asked Gemini to create this function from the CHERIoT sail model. I think the values are equivalent to your loop but not 100% certain about it. 

Runnable version: https://godbolt.org/z/9zn8hK484


```c++
#include <cassert>
#include <cstdint>
#include <cstdio>

#include "llvm/Support/MathExtras.h"

template <typename IntType, unsigned int MantissaWidth,
          unsigned int MaxExponent>
struct RVYFormat {
    using T = IntType;
    static constexpr unsigned int MW = MantissaWidth;
    static constexpr unsigned int MAX_E = MaxExponent;
    static constexpr unsigned int IE_TAKE_BITS = 3;

    static constexpr T getMask(T length) {
        if (length == 0) return ~static_cast<T>(0);

        // Extract bits that overflow the uncompressed mantissa window
        uint64_t slice_val = static_cast<uint64_t>(length) >> (MW - 1);
        unsigned int e = 64 - llvm::countl_zero(slice_val);
        // We use internal exponent if length overflows OR the denormal boundary
        // bit is set
        bool ie = (e != 0) || ((static_cast<uint64_t>(length) >> (MW - 2)) & 1);
        // Include bits use by the internal exponent for the shift value
        unsigned int e_prime = ie ? (e + IE_TAKE_BITS) : 0;

        // These are probably not needed, but were useful for debugging when I initially got it wrong.
        assert(e <= MAX_E && "Raw exponent exceeds architecture maximum");
        assert(e_prime <= sizeof(T) * 8 &&
               "Shift amount exceeds integer width");

        // Left-shift ~0 to mask out the lost precision bits
        return ~static_cast<T>(0) << e_prime;
    }
};

// RV64Y: 128-bit cap, 64-bit addr, 14-bit mantissa, Max E = 52
struct RV64YFormat : public RVYFormat<uint64_t, 14, 52> {};
// RV32Y: 64-bit cap, 32-bit addr, 10-bit mantissa, Max E = 24
struct RV32YFormat : public RVYFormat<uint32_t, 10, 24> {};

// Explicit 4-bit exponent, no stolen bits, non-linear exponent (jumps from 14
// -> 24)
// WARNING: generated by gemini from the sail code
struct CheriIoTFormat {
    static constexpr unsigned int MW = 9;
    static constexpr unsigned int MAX_COMPRESSED_E = 14;
    static constexpr unsigned int MAX_E = 24;

    static constexpr uint32_t getMask(uint32_t length) {
        if (length == 0) return ~0U;

        unsigned int e = 0;
        uint32_t L = length - 1;

        // The hardware max is 511. We use 510 as the shift threshold for 'L'
        // to implicitly account for the hardware's round-up behavior
        // when precision bits are lost during the shift.
        uint32_t shift_threshold = (1U << MW) - 2;

        if (L > shift_threshold) {
            // Initial exponent guess
            e = (32 - llvm::countl_zero(L)) - MW;
            // Mantissa overflow correction
            if ((L >> e) > shift_threshold) {
                e++;
            }
        }

        // Handle the hardware decoding gap (e > 14 snaps to 24)
        if (e > MAX_COMPRESSED_E) {
            e = MAX_E;
        }

        assert(e <= MAX_E && "Exponent exceeds architecture maximum");
        return ~0U << e;
    }
};
```

This prints the following:
```
--- RV64Y (MW=14, MAX_E=52) ---
Mask for                    0 (0000000000000000) = ffffffffffffffff
Mask for                 4095 (0000000000000fff) = ffffffffffffffff
Mask for                 4096 (0000000000001000) = fffffffffffffff8
Mask for                 4097 (0000000000001001) = fffffffffffffff8
Mask for                 8193 (0000000000002001) = fffffffffffffff0
Mask for 18446744073709551615 (ffffffffffffffff) = ffc0000000000000

--- RV32Y (MW=10, MAX_E=24) ---
Mask for            0 (00000000) = ffffffff
Mask for          255 (000000ff) = ffffffff
Mask for          256 (00000100) = fffffff8
Mask for          257 (00000101) = fffffff8
Mask for          513 (00000201) = fffffff0
Mask for   4294967295 (ffffffff) = fc000000

--- CHERIoT (MW=9, MAX_E=24) ---
Mask for            0 (00000000) = ffffffff
Mask for          511 (000001ff) = ffffffff
Mask for          512 (00000200) = fffffffe
Mask for          513 (00000201) = fffffffe
Mask for         1023 (000003ff) = fffffffc
Mask for         1024 (00000400) = fffffffc
Mask for      8372224 (007fc000) = ffffc000
Mask for      8372225 (007fc001) = ff000000
Mask for     16744448 (00ff8000) = ff000000
Mask for     33488896 (01ff0000) = ff000000
Mask for   4294967295 (ffffffff) = ff000000
```

llvm/include/llvm/Support/CHERICapabilityFormat.h
arichardson: Should all those uint64_t be Address type instead?

llvm/lib/Support/CHERICapabilityFormat.cpp
arichardson: This should have a mantissa of 10

llvm/unittests/Support/CHERICapabilityFormatTest.cpp
arichardson: Should be 255

arichardson: 2044

llvm/lib/Support/CHERICapabilityFormat.cpp
resistor: Done

llvm/unittests/Support/CHERICapabilityFormatTest.cpp
resistor: Done

resistor: Done

arichardson:
Overall this looks sensible to me but I'm not convinced we need virtual functions here. Wouldn't something like what I did in the godbolt link be sufficient? Since there is no state, A class with static functions should be sufficient? And then we can keep it header only

resistor:
> Overall this looks sensible to me but I'm not convinced we need virtual functions here. Wouldn't something like what I did in the godbolt link be sufficient? Since there is no state, A class with static functions should be sufficient? And then we can keep it header only

I'm intentionally trying _not_ to have this be header only, so as not to bloat compile-time/code-size for code that has no particular need to be inlined into each instantiation.

That said, you're right that we don't need the virtual calls here. I just pushed a CRTP-based version that I think factors it nicely.

llvm/lib/Support/CHERICapabilityFormat.cpp
arichardson:
```suggestion
  // Include bits used by the internal exponent for the shift value.
```

arichardson:
> > Overall this looks sensible to me but I'm not convinced we need virtual functions here. Wouldn't something like what I did in the godbolt link be sufficient? Since there is no state, A class with static functions should be sufficient? And then we can keep it header only
> 
> I'm intentionally trying _not_ to have this be header only, so as not to bloat compile-time/code-size for code that has no particular need to be inlined into each instantiation.
> 

That's fine too, I was expecting there to only be one use of these in the lib/Target and one in LLD, but you're right it's probably cleaner to just push it into the .cpp file.

> That said, you're right that we don't need the virtual calls here. I just pushed a CRTP-based version that I think factors it nicely.

llvm/include/llvm/Support/CHERICapabilityFormat.h
arichardson: I guess you are using uint64_t instead of AddressType here to ensure that callers don't have to manually truncate for 32-bit arches?

I'd suggest we add an assertion to the implementation that the arguments are no larger than `AddressMask`.

llvm/lib/Support/CHERICapabilityFormat.cpp
arichardson:
```suggestion
    return RVYCapabilityFormat::AddressMask;
```
Isn't this the same?

arichardson:
```suggestion
  return RVYCapabilityFormat::AddressMask << Eprime;
```
This should be equivalent?

arichardson:
```suggestion
      return CHERIoTCapabilityFormat::AddressMask << e;
```

arichardson:
```suggestion
  return CHERIoTCapabilityFormat::AddressMask << 24;
```

arichardson (commented):
A few rather minor comments otherwise looking good.

llvm/lib/Support/CHERICapabilityFormat.cpp
jrtc27: Would something like:
```
unsigned E;
if (Length > 511ULL << 14) {
  E = 24;
} else {
  E = 64 - countLeadingZeros64(Length);
  if (Length > 511ULL << E)
    ++E;

  assert(E <= 14 && "...");
}
return CHERIoTCapabilityFormat::AddressMask << E;
```
not be equivalent and avoid the need to loop 14 times?
(also ...ULL should probably be using UINT64_C(...))

llvm/include/llvm/Support/CHERICapabilityFormat.h
resistor: In the original version I wasn't passing in the full type, only the width, so varying the return type wasn't possible. I'll experiment and see if it works out well now that the type is actually available.

llvm/lib/Support/CHERICapabilityFormat.cpp
resistor: This isn't equivalent when AddressType is uint32_t. You'll end up with 1s in the high half of the returned value.

Making the return type actually be AddressType would fix this.

arichardson: Oh yes we'd need to cast to AddressType

resistor: In spirit yes, but there's a bit more fiddliness. Here's a version that I've exhausted tested as matching the loop:

```
#include <cstdint>
#include <bit>
#include <cassert>

uint32_t a(uint32_t Length) {
  for (unsigned e = 0; e <= 14; ++e) {
    if (Length <= 511ULL << e)
      return (~static_cast<uint32_t>(0) << e);
  }

  return (~static_cast<uint32_t>(0) << 24);
}


uint32_t b(uint32_t Length) {
    uint32_t E;
    if (Length > 511ULL << 14)
        E = 24;
    else {
        E = Length > 511 ? 32 - std::countl_zero(Length) - 9 : 0;
        if (Length > 511ULL << E)
            ++E;
    }
    return ~static_cast<uint32_t>(0)<< E;
}

int main() {
    for (uint32_t i = 0; i < ~static_cast<uint32_t>(0); ++i)
      assert(a(i) == b(i));
    assert(a(~static_cast<uint32_t>(0)) == b(~static_cast<uint32_t>(0)));
    return 0;
}
```

llvm/include/llvm/Support/CHERICapabilityFormat.h
resistor: Done

llvm/lib/Support/CHERICapabilityFormat.cpp
resistor: Done

resistor: Done

resistor: Done

resistor: Done

resistor: Done

arichardson: Can probably delete this comment?

llvm/include/llvm/Support/CHERICapabilityFormat.h
arichardson:
```suggestion
  /// mask is 0 where the capability format alignment requires the
```

arichardson: Is this needed? Isn't it implicit since the base class one is deleted?

arichardson: Is this =deleted needed?

llvm/lib/Support/CHERICapabilityFormat.cpp
arichardson, line 17: I think the & AddressMask is not needed since all types are large enough to not get promoted but can keep it just to be explicit.

arichardson:
```suggestion
    assert(E <= 14 && "CHERIoT capabilities cannot encode E between 14 and 24");
```

arichardson (approved):
LGTM

arichardson:
Maybe we should wait with landing this until we have one PR using it pending?

llvm/include/llvm/Support/CHERICapabilityFormat.h
resistor: Fixed

llvm/lib/Support/CHERICapabilityFormat.cpp
resistor: I prefer to keep this one explicit. While I think you're right, the parameterized type makes it have to visually verify, and it should optimize away anyways.

resistor: Done

llvm/include/llvm/Support/CHERICapabilityFormat.h
resistor: Fixed

resistor:
Squashed and rebased to demonstrate the final content.

resistor:
> Maybe we should wait with landing this until we have one PR using it pending?

See https://github.com/llvm/llvm-project/pull/203872 which stacks on this.

RKSimon:
@resistor I'm seeing a this build warning on MSVC:
```
Building CXX object lib\Support\CMakeFiles\LLVMSupport.dir\CHERICapabilityFormat.cpp.obj
E:\llvm\llvm-project\llvm\include\llvm/Support/CHERICapabilityFormat.h(47): warning C4661: 'AddressType llvm::CHERICapabilityFormatBase<llvm::CHERIoTCapabilityFormat,AddressType>::getAlignmentMask(AddressType)': no suitable definition provided for explicit template instantiation request
        with
        [
            AddressType=uint32_t
        ]
E:\llvm\llvm-project\llvm\include\llvm/Support/CHERICapabilityFormat.h(26): note: see declaration of 'llvm::CHERICapabilityFormatBase<llvm::CHERIoTCapabilityFormat,uint32_t>::getAlignmentMask'
E:\llvm\llvm-project\llvm\lib\Support\CHERICapabilityFormat.cpp(54): warning C4661: 'AddressType llvm::CHERICapabilityFormatBase<llvm::RVYCapabilityFormat<AddressType,10,24>,AddressType>::getAlignmentMask(AddressType)': no suitable definition provided for explicit template instantiation request
        with
        [
            AddressType=uint32_t
        ]
E:\llvm\llvm-project\llvm\include\llvm/Support/CHERICapabilityFormat.h(26): note: see declaration of 'llvm::CHERICapabilityFormatBase<llvm::RVYCapabilityFormat<uint32_t,10,24>,uint32_t>::getAlignmentMask'
E:\llvm\llvm-project\llvm\lib\Support\CHERICapabilityFormat.cpp(56): warning C4661: 'AddressType llvm::CHERICapabilityFormatBase<llvm::RVYCapabilityFormat<AddressType,14,24>,AddressType>::getAlignmentMask(AddressType)': no suitable definition provided for explicit template instantiation request
        with
        [
            AddressType=uint64_t
        ]
```

resistor:
@RKSimon Perhaps MSVC wants there to be an explicit instantiation of the base classes as well? I don't have a Windows machine to try it out on.

llvm/lib/Support/CHERICapabilityFormat.cpp
```diff
@@ -0,0 +1,77 @@
...
+  assert(Eprime <= sizeof(AddressType) * 8 &&
+         "Shift amount exceeds integer width");
+
+  // Left-shift ~0 to mask out the lost precision bits
+  return RVYCapabilityFormat::AddressMask << Eprime;
+}
+
+template struct CHERICapabilityFormatBase<RVYCapabilityFormat<uint32_t, 10, 24>,
+                                          uint32_t>;
+template struct CHERICapabilityFormatBase<RVYCapabilityFormat<uint64_t, 14, 24>,
```
arichardson, line 55:
```suggestion
template struct CHERICapabilityFormatBase<RVYCapabilityFormat<uint64_t, 14, 52>,
```

llvm/lib/Support/CHERICapabilityFormat.cpp
```diff
@@ -0,0 +1,77 @@
...
+                                          uint64_t>;
+template struct RVYCapabilityFormat<uint32_t, 10, 24>;
+template struct RVYCapabilityFormat<uint64_t, 14, 24>;
```
arichardson, line 58:
```suggestion
template struct RVYCapabilityFormat<uint64_t, 14, 52>;
```

AaronBallman:
> @RKSimon Perhaps MSVC wants there to be an explicit instantiation of the base classes as well? I don't have a Windows machine to try it out on.

I'm still seeing these warnings:
```
'AddressType llvm::CHERICapabilityFormatBase<llvm::CHERIoTCapabilityFormat,AddressType>::getAlignmentMask(AddressType)': no suitable definition provided for explicit template instantiation request
'AddressType llvm::CHERICapabilityFormatBase<llvm::RVYCapabilityFormat<AddressType,10,24>,AddressType>::getAlignmentMask(AddressType)': no suitable definition provided for explicit template instantiation request
'AddressType llvm::CHERICapabilityFormatBase<llvm::RVYCapabilityFormat<AddressType,14,52>,AddressType>::getAlignmentMask(AddressType)': no suitable definition provided for explicit template instantiation request
```
Any luck with this?

resistor:
> Any luck with this?

I don't have access to a Windows Machine or MSVC

> > @RKSimon Perhaps MSVC wants there to be an explicit instantiation of the base classes as well? I don't have a Windows machine to try it out on.
> 
> I'm still seeing these warnings:
> 
> ```
> 'AddressType llvm::CHERICapabilityFormatBase<llvm::CHERIoTCapabilityFormat,AddressType>::getAlignmentMask(AddressType)': no suitable definition provided for explicit template instantiation request
> 'AddressType llvm::CHERICapabilityFormatBase<llvm::RVYCapabilityFormat<AddressType,10,24>,AddressType>::getAlignmentMask(AddressType)': no suitable definition provided for explicit template instantiation request
> 'AddressType llvm::CHERICapabilityFormatBase<llvm::RVYCapabilityFormat<AddressType,14,52>,AddressType>::getAlignmentMask(AddressType)': no suitable definition provided for explicit template instantiation request
> ```
> 
> Any luck with this?

I don't have a way to reproduce or investigate it to see what will satisfy MSVC. One idea I had is that maybe it needs the `LLVM_ABI` macro applied.

AaronBallman:
> > Any luck with this?
> 
> I don't have a way to reproduce or investigate it to see what will satisfy MSVC. One idea I had is that maybe it needs the `LLVM_ABI` macro applied.

I think the issue is that `CHERICapabilityFormatBase::getAlignmentMask()` is declared in the header but not defined:
```
uint32_t CHERIoTCapabilityFormat::getAlignmentMask(uint32_t Length) {
```
and
```
template <typename AddressType, unsigned MW, unsigned MAX_E>
AddressType
RVYCapabilityFormat<AddressType, MW, MAX_E>::getAlignmentMask(uint64_t Length) {
```
neither of those are the base class type.

> I don't have a way to reproduce or investigate it to see what will satisfy MSVC. One idea I had is that maybe it needs the LLVM_ABI macro applied.

I'm happy to help if I can.

jrtc27:
I can take a look at this, I've done this pattern before

(edit: VM needed a lot of updates but all set up now to test a fix tomorrow)

resistor:
> > I don't have a way to reproduce or investigate it to see what will satisfy MSVC. One idea I had is that maybe it needs the LLVM_ABI macro applied.
> 
> I'm happy to help if I can.

Can you see if adding the annotation addresses it?

AaronBallman:
> > > I don't have a way to reproduce or investigate it to see what will satisfy MSVC. One idea I had is that maybe it needs the LLVM_ABI macro applied.
> > 
> > 
> > I'm happy to help if I can.
> 
> Can you see if adding the annotation addresses it?

FWIW, it didn't address the issues when I tried it out.

PR #197405, opened 2026-05-13
[LLDB][NFCI] Use resize_for_overwrite in VASprintf

felipepiovezan:
This code path is particularly hot when logs are enabled.
On a logging heavy experiment (gdbremote logs), this brought down the number of CPU cycles from 933M to 797M cycles. (release, no assertions)


Teemperor (approved):
LGTR

JDevlieghere (approved):
Nice find!

PR #196480, opened 2026-05-08
[InstCombine] Fold square into comparison with constant

macsencasaus:
Closes #196233 
https://alive2.llvm.org/ce/z/5kht6g

PR #197406, opened 2026-05-13
[APInt] Provide sqrtFloor (floor of square root) instead of sqrt (rounded)

jayfoad:
This simplifies both the implementation and the only in-tree user.

I changed the name to avoid silently changing the behavour of an
existing function that might have out-of-tree users. The new name is
inspired by https://en.wikipedia.org/wiki/Integer_square_root.


jayfoad:
> This simplifies both the implementation and the only in-tree user

... and the potential new user in #196480.

llvm/include/llvm/ADT/APInt.h
```diff
@@ -1809,8 +1809,8 @@ class [[nodiscard]] APInt {
     return logBase2();
   }
 
-  /// Compute the square root.
-  LLVM_ABI APInt sqrt() const;
+  /// Compute the floor of the square root of the unsigned value.
+  LLVM_ABI APInt isqrt() const;
```
MaxGraey: I suggest calling it `floor_sqrt` instead of `isqrt`. Because the comparison between `isqrt` and `sqrt` makes it feel like `sqrt` isn't an integer sqrt, even though both methods are integer-based. For original `sqrt` just clarify in comments it's rounded int sqrt.

jayfoad: Happy to hear other opinions on this.

My own take is that I am deliberately not leaving multiple versions of this function in the code base (since I think floor is the only behaviour we need) so I don't think it's super important to disambiguate them by adding the `floor_` prefix.

> For original `sqrt` just clarify in comments it's rounded int sqrt.

My patch removes the old function.

MaxGraey: Personally, I don't think this is a necessary breaking change to the API if we can easily extend it

jayfoad: You mean we should keep both versions of the function? I disagree. sqrt-with-rounding should be removed if there are no in-tree users, we shouldn't just keep it on the off chance that someone is using it out-of-tree. And, being slightly opinionated, it's a much less useful function than sqrt-with-floor anyway.

arsenm: I agree that we should not retain dead code on the off chance of out of tree users

MaxGraey: So maybe we shouldn't rename `sqrt` to `isqrt`? Just note that the sqrt function in APInt now uses floor instead of round.

jayfoad: I appreciate you putting effort in to review this PR but I'm not sure what I'm supposed to do with a comment like that. Yes "maybe" we could leave the function name unchanged, but I decided not to do that and explained why in the description. Do you disagree with that decision?

MaxGraey: > I changed the name to avoid silently changing the behavour of an
existing function that might have out-of-tree users

I'm just suggesting a few options, especially since this already involves breaking changes. To me, the current naming feels a bit inconsistent and potentially confusing within the namespace. For example, someone might naturally look for something like `APInt::ipow` while only `APInt::pow` exists.

That said, I may very well be overestimating the issue, and no one see the problem with that, and if so, I apologize.

jayfoad: > To me, the current naming feels a bit inconsistent and potentially confusing within the namespace. For example, someone might naturally look for something like `APInt::ipow` while only `APInt::pow` exists.

Fair point. The only reason I had for picking the name "isqrt" is that it's mentioned on that wikipedia page. I have no idea how widely used it is. I'm pretty sure number theory would not use the corresponding name "ipow" since there is no reason to distinguish it from a regular power operation.

> That said, I may very well be overestimating the issue, and no one see the problem with that, and if so, I apologize.

Thanks, no need to apologize, I think it's fine to explore the options, I just don't want to be changing the implementation from one arbitrary option to another _equally_ arbitrary option. Of course if there's some consensus that some option is _best_ then I'm happy to go with that.

MaxGraey: Ok, since we’ve found some points of consensus, what do you think about choosing `sqrt_floor` / `sqrtFloor` / `sqrt_trunc` / `sqrtTrunc` instead of `isqrt`? In other words, you can use any suffix variant you prefer instead of a prefix. This will also be much better in terms of IDE autocompletion.

jayfoad: I guess any those names would be fine.

I did look at https://gmplib.org/manual/Integer-Roots for inspiration. It just uses the name `mpz_sqrt` for "the truncated integer part of the square root". (This further convinces me that the truncating behaviour is the only behaviour worth implementing.)

MaxGraey: `square_root` also fine. It will be great to extend it for return perfect_square out ref:
```cpp
APInt square_root(bool& is_perfect_square);
APInt square_root() const;
```

jayfoad: > what do you think about choosing `sqrt_floor` / `sqrtFloor` / `sqrt_trunc` / `sqrtTrunc` instead of `isqrt`?

Out of those options I would lean towards `sqrtFloor`.

jayfoad: Done

arsenm (approved):

topperc (approved):
LGTM

PR #197407, opened 2026-05-13
[clang][ASTImporter] Fix of crash at ConstraintSatisfaction import

balazske:
Null pointer dereference could happen during `ASTImporter` import of a `ConstraintSatisfaction` object.


clang/test/Analysis/ctu/constraintsatisfaction.cpp
```diff
@@ -0,0 +1,25 @@
+// RUN: rm -rf %t && mkdir %t
+// RUN: mkdir -p %t/ctudir
+// RUN: %clang_cc1 -triple x86_64-pc-linux-gnu -std=c++20 \
+// RUN:   -emit-pch -o %t/ctudir/constraintsatisfaction-import.cpp.ast %S/Inputs/constraintsatisfaction-import.cpp
+// RUN: cp %S/Inputs/constraintsatisfaction-import.cpp.externalDefMap.ast-dump.txt %t/ctudir/externalDefMap.txt
```
steakhal: If this refers to a single-lined file, why don't we use split-file and merge these files?

clang/test/Analysis/ctu/constraintsatisfaction.cpp
```diff
@@ -0,0 +1,25 @@
+// RUN: rm -rf %t && mkdir %t
+// RUN: mkdir -p %t/ctudir
+// RUN: %clang_cc1 -triple x86_64-pc-linux-gnu -std=c++20 \
+// RUN:   -emit-pch -o %t/ctudir/constraintsatisfaction-import.cpp.ast %S/Inputs/constraintsatisfaction-import.cpp
```
steakhal: This file could also come here.

steakhal (commented):
The patch looks good to me, but I had improvement ideas for the test.

clang/test/Analysis/ctu/constraintsatisfaction.cpp
balazske: For this simple case `echo` is usable too.

steakhal: This wouldn't solve the issue that you could inline the importee function here. That would truly make this test self-sufficient.

steakhal (approved):
LGTM. thank you.
Consider adding the `no-crash` comment to the line where it used to crash.

cor3ntin (approved):

PR #197409, opened 2026-05-13
[lldb][test] Don't overwrite existing decorators in _skipForVariant

Teemperor:
_skipForVariant uses a dictionary to save the decorator functions for specific variants. However, if the dictionary already contains a decorator function, then that decorator is overwritten in the current implementation.

This means that if we have two decorators that skip a variant in combination with another check, then the second decorator overwrites the first one and is effectively ignored.

Downstream we are hit by this because we have an embedded Swift variant. If we skip this variant for both Linux and Windows, then depending on the decorator order we still run the test on one of those two platforms (as one decorator is overwritten by the other).


bulbazord (approved):

JDevlieghere (approved):

medismailben (approved):
LGTM

PR #197413, opened 2026-05-13
[libc] Fix for SYS_mmap2 offset computation

labath:
The comment implies that the offset argument is a multiple of page size, but
[this](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/csky/kernel/syscall.c#L25) [is](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/parisc/kernel/sys_parisc.c#L193) [not](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/microblaze/kernel/sys_microblaze.c#L50) [the](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/riscv/kernel/sys_riscv.c#L48) [case](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/arm64/kernel/sys32.c#L47) [for](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/sparc/kernel/sys_sparc_32.c#L113) [almost](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/m68k/kernel/sys_m68k.c#L44) [every](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/powerpc/kernel/syscalls.c#L56) [architecture](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/sh/kernel/sys_sh.c#L46) [supported](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/powerpc/kernel/syscalls.c#L56) [by](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/mips/kernel/syscall.c#L76) [linux](https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/arch/arm/kernel/entry-common.S#L410). Most architectures just use fixed 4k units instead.

x86 code does not reference the pages sizes anywhere, but x86 always has a 4k (base) page size. The same is probably true for Nios II, though information is a bit scarce on that one. I believe openrisc uses fixed 8k units, but this should be confirmed when/if we're porting to that architecture.

Itanium had configurable page sizes and did not use fixed mmap2 scaling, but itanium support was removed from the kernel two years ago. Hexagon and ARC may be the only currently supported architectures that do the same, but this is also best confirmed during porting.

This patch is a no-op for any architecture we currently support, but it makes it clear that this behavior is deliberate, and not something to be fixed. This code was introduced way back in
[2019](https://reviews.llvm.org/D71634) and while the review contained a discussion on page sizes, it revolved around the right way to obtain it, rather than the question of whether it is actually necessary.

### DIFF
--- a/libc/src/__support/OSUtil/linux/syscall_wrappers/mmap.h
+++ b/libc/src/__support/OSUtil/linux/syscall_wrappers/mmap.h
@@ -14,7 +14,6 @@
 #include "src/__support/common.h"
 #include "src/__support/error_or.h"
 #include "src/__support/macros/config.h"
-#include <linux/param.h> // For EXEC_PAGESIZE
 #include <sys/syscall.h> // For syscall numbers
 
 namespace LIBC_NAMESPACE_DECL {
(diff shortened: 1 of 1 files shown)


kaladron (approved):
Nice work!  And let's not port to ia64 to avoid the variable page sizes... =)

PR #197416, opened 2026-05-13
[clang][constexpr] Move anonymous-union check before getDestructor() in HandleDestructionImpl

### LINKED ISSUE #197403: [Clang] Implicitly-defined destructor not recognized as constexpr when destroying union member in constant evaluation
Clang incorrectly rejects valid C++20 code where a union member with an implicitly-defined destructor is destroyed via std::destroy_at in a consteval context. The same code is accepted if the destructor is explicitly defaulted (= default), even though both forms should have identical constexpr semantics per the C++20 standard.

## Minimal Reproducer

```cpp
#include <memory>

struct Inner {
    constexpr ~Inner() noexcept {}
};

// Case A: implicitly-defined dtor -> TRIGGERS BUG
struct Outer {
    Inner inner;
};

// Case B: user-declared dtor -> OK
struct OuterOk {
    Inner inner;
    constexpr ~OuterOk() noexcept = default;
};

template<typename T>
struct BugTrigger {
    union { T value; int dummy; };

    constexpr BugTrigger() : value{} {}
    constexpr ~BugTrigger() noexcept {
        std::destroy_at(&value);
    }
};

consteval int test_bug() {
    BugTrigger<Outer> bt;
    return 0;
}
static_assert(test_bug() == 0);  // BUG: Clang rejects

consteval int test_ok() {
    BugTrigger<OuterOk> bt;
    return 0;
}
static_assert(test_ok() == 0);  // OK

int main() {}
```

## Actual Behavior

```
<source>:32:15: error: static assertion expression is not an integral constant expression
   32 | static_assert(test_bug() == 0);  // BUG: Clang rejects
      |               ^~~~~~~~~~~~~~~
<source>:29:23: note: subexpression not valid in a constant expression
   29 |     BugTrigger<Outer> bt;
      |                       ^
<source>:29:23: note: in call to 'bt.~BugTrigger()'
   29 |     BugTrigger<Outer> bt;
      |                       ^~
<source>:32:15: note: in call to 'test_bug()'
   32 | static_assert(test_bug() == 0);  // BUG: Clang rejects
      |               ^~~~~~~~~~
1 error generated.
Compiler returned: 1
```

godbolt link: https://godbolt.org/z/6Kaobhqf4

## Expected Behavior

Both test_bug() and test_ok() should compile successfully.

SekaiArendelle:
When HandleDestructionImpl destroys a class object during constant evaluation, it attempts to look up the class's destructor via getDestructor() before checking whether the record is an anonymous union. Anonymous unions should not have their destructors invoked directly during constant evaluation — their lifetime is managed by the enclosing class's destructor. Move the anonymous-union short-circuit before the getDestructor() call so anonymous unions are handled early, regardless of whether getDestructor() returns null or not.

This fixes cases where an object with an implicitly-defined constexpr destructor stored inside an anonymous union member was incorrectly rejected during constant evaluation.

Fixes #197403


frederick-vs-ja:
Thanks! You can link your pull request to this issue using [GitHub's close/fix/resolve syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).
(in the PR description not the commit message)

shafik (commented):
So we should also be checking the new constexpr interpreter via

`-fexperimental-new-constant-interpreter`

 which seems to get this right. I am not sure if the correct way to do this though.

CC @tbaederr

SekaiArendelle:
> So we should also be checking the new constexpr interpreter via
> 
> `-fexperimental-new-constant-interpreter`
> 
> which seems to get this right. I am not sure if the correct way to do this though.
> 
> CC @tbaederr

https://godbolt.org/z/rdrsWsxr9 seems fine

ojhunt (approved):
LGTM

shafik:
> -fexperimental-new-constant-interpreter

That is not what I was asking.

Your test in this PR does not exercise the new constant expression interpreter which can be invokes by passing `-fexperimental-new-constant-interpreter`.

We should have a test that does this. I am not sure adding a line to the current test w/ `-fexperimental-new-constant-interpreter` is the right way since it is very large and may break in other ways.

If such a test already exists then please point it out.

shafik (changes requested):

SekaiArendelle:
@shafik test added to the new bytecode interpreter. In order to reduce the scope of impact, I chose to add this test solely to the bytecode interpreter.

SekaiArendelle:
@shafik  Could you review this pr again?

tbaederr:
@shafik Ping

shafik (approved):

SekaiArendelle:
@shafik I think this PR is good to go. Ready for a merge?

SekaiArendelle:
Fixes #197403

PR #197417, opened 2026-05-13
[LV] Avoid crashing for vector calls with scalar byte types

huntergr-arm:
If a parameter to a vector function variant is uniform or linear, check
whether the type is SCEVable first. Byte types aren't, so would cause
an assert. We could improve this later if needed.


llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -5026,7 +5027,8 @@ void LoopVectorizationCostModel::setVectorizedCallDecision(ElementCount VF) {
             // TODO: do we need to figure out the cost of an extract to get the
             // first lane? Or do we hope that it will be folded away?
             ScalarEvolution *SE = PSE.getSE();
-            if (!match(SE->getSCEV(ScalarParam),
+            if (!SE->isSCEVable(ScalarParam->getType()) ||
```
david-arm, line 5030: Is this change actually needed? I would expect getSCEV to return SCEVUnknown or something like that so it won't match anyway.

llvm/test/Transforms/LoopVectorize/byte-type-function-variants.ll
```diff
@@ -0,0 +1,210 @@
+; NOTE: Assertions have been autogenerated by utils/update_test_checks.py UTC_ARGS: --check-globals none --version 6
+; RUN: opt < %s -passes=loop-vectorize -force-vector-width=2 -force-vector-interleave=1 -S | FileCheck %s
+
+;; Safe to vectorize normally.
```
fhahn:
```suggestion
; Safe to vectorize normally.
```

llvm/test/Transforms/LoopVectorize/byte-type-function-variants.ll
```diff
@@ -0,0 +1,210 @@
...
+  store b64 %call, ptr %gep.dst
+  %iv.next = add nuw nsw i64 %iv, 1
+  %exit.cond = icmp eq i64 %iv.next, %n
+  br i1 %exit.cond, label %exit, label %for.body
+
+exit:
+  ret void
+}
+
+;; Must be scalarized, since we can't form a SCEV from a byte type.
```
fhahn: Better say it is currently scalarized unnecessarily + TODO

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: I think getSCEV must only be called with SCEV-able types

huntergr-arm: Yep, the test file hits an assert without those changes.

llvm/test/Transforms/LoopVectorize/byte-type-function-variants.ll
huntergr-arm: done

llvm/test/Transforms/LoopVectorize/byte-type-function-variants.ll
huntergr-arm: done

fhahn (approved):
LGTM, thanks

PR #197418, opened 2026-05-13
[Dexter] Add basic debugging support for structured scripts

SLTozer:
This patch adds a debugger controller for structured scripts. This controller operates as follows:
- !where nodes at the root of the script (currently the only kind allowed) have function or line breakpoints set to cover them.
- Whenever the debugger stops, the controller will examine the stack to determine whether any !where nodes are in scope.
- While any !where is in scope, its associated !values will be evaluated in the debugger and the results stored, and the debugger will single-step.
- When no !where is in scope, the debugger will continue.

This is a simplified implementation compared to the final version, as it is missing support for nested !where nodes, Scope evaluation, and conditions/hit counts; these will be added in later commits.

### DIFF
--- a/cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
+++ b/cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
@@ -25,20 +25,20 @@ from dex.utils.Timeout import Timeout
 from dex.dextIR import DextIR, FrameIR, StepIR
 
 
-def paths_differ(expected: str, actual: str) -> bool:
-    """Returns True if expected is not a trailing subpath of actual, i.e. if `actual` does not end with `expected` after
-    normalizing both paths."""
-    normalized_expected: str = os.path.normcase(os.path.normpath(expected))
-    normalized_actual: str = os.path.normcase(os.path.normpath(actual))
-    return not normalized_actual.endswith(normalized_expected)
+def is_subpath(subpath: str, superpath: str) -> bool:
+    """Returns True if subpath is not a trailing subpath of superpath, i.e. if `superpath` does not end with `subpath`
+    after normalizing both paths."""
+    normalized_subpath: str = os.path.normcase(os.path.normpath(subpath))
+    normalized_superpath: str = os.path.normcase(os.path.normpath(superpath))
+    return normalized_superpath.endswith(normalized_subpath)
 
 
-# A very simple matcher, returns True iff `where` matches `frame`.
 def match_where_to_frame(
     where: Where,
     frame: FrameIR,
 ) -> bool:
-    if where.file is not None and paths_differ(where.file, frame.loc.path):
+    """A very simple matcher, returns True iff `where` matches `frame`."""
+    if where.file is not None and not is_subpath(where.file, frame.loc.path):
         return False
     if where.function is not None:
         fn = frame.function
@@ -104,7 +104,7 @@ def get_active_where_expects(
 
 
 class DebuggerAction(Enum):
-    STEP = 0
+    STEP_OVER = 0
     STEP_OUT = 1
     CONTINUE = 2
     EXIT = 3
@@ -209,7 +209,7 @@ class ScriptDebuggerController(DebuggerControllerBase):
             if any(
                 frame_idx == 0 for frame_idx, expects in active_where_expects.values()
             ):
-                next_action = DebuggerAction.STEP
+                next_action = DebuggerAction.STEP_OVER
             elif active_where_expects:
                 next_action = DebuggerAction.STEP_OUT
             else:
@@ -229,7 +229,7 @@ class ScriptDebuggerController(DebuggerControllerBase):
 
             if next_action == DebuggerAction.EXIT:
                 break
-            elif next_action == DebuggerAction.STEP:
+            elif next_action == DebuggerAction.STEP_OVER:
                 self.debugger.step_next()
             elif next_action == DebuggerAction.STEP_OUT:
                 self.debugger.step_out()
--- a/cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
+++ b/cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
@@ -11,7 +11,7 @@ recording debugger output."""
 from enum import Enum
 import os
 import time
-from typing import Dict, List, Tuple
+from typing import Dict, List, Optional, Tuple
 
 from dex.debugger.DebuggerControllers.DebuggerControllerBase import (
     DebuggerControllerBase,
@@ -26,8 +26,8 @@ from dex.dextIR import DextIR, FrameIR, StepIR
 
 
 def is_subpath(subpath: str, superpath: str) -> bool:
-    """Returns True if subpath is not a trailing subpath of superpath, i.e. if `superpath` does not end with `subpath`
-    after normalizing both paths."""
+    """Returns True if subpath is a trailing subpath of superpath, i.e. if `superpath` ends with `subpath` after
+    normalizing both paths."""
     normalized_subpath: str = os.path.normcase(os.path.normpath(subpath))
     normalized_superpath: str = os.path.normcase(os.path.normpath(superpath))
     return normalized_superpath.endswith(normalized_subpath)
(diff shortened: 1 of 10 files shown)


cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
```diff
@@ -0,0 +1,241 @@
...
+from dex.debugger.DebuggerBase import DebuggerBase
+from dex.debugger.DAP import DAP
+from dex.test_script.Nodes import Expect, Value, Where
+from dex.test_script.Script import DexterScript, Scope
+from dex.tools import Context
+from dex.utils.Timeout import Timeout
+from dex.dextIR import DextIR, FrameIR, StepIR
+
+
+def paths_differ(expected: str, actual: str) -> bool:
```
OCHyams: nit: `is_subpath` (and `not` at call sites) or something might be clearer? ymmv

cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
```diff
@@ -0,0 +1,241 @@
...
+    """Returns True if expected is not a trailing subpath of actual, i.e. if `actual` does not end with `expected` after
+    normalizing both paths."""
+    normalized_expected: str = os.path.normcase(os.path.normpath(expected))
+    normalized_actual: str = os.path.normcase(os.path.normpath(actual))
+    return not normalized_actual.endswith(normalized_expected)
+
+
+# A very simple matcher, returns True iff `where` matches `frame`.
+def match_where_to_frame(
```
OCHyams, line 36: nit: should the comment be a docucomment string instead?

cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
```diff
@@ -0,0 +1,241 @@
...
+        ):
+            active_where_expects[scope.where][1].append(expect)
+
+    script.visit_script(visit_where=get_active_wheres, visit_expect=get_active_expects)
+
+    return active_where_expects
+
+
+class DebuggerAction(Enum):
+    STEP = 0
```
OCHyams: nit: can we rename STEP STEP_OVER (next?) or STEP_IN -it's not clear from the name which it's meant to be?

cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def add_where_entry_bp(self, where: Where, scope: Scope):
+        """Adds a breakpoint to catch when we enter the given !where node."""
+        added_ids = []
+        if where.function:
+            if where.lines:
+                raise NotImplementedError(
+                    f"Implementation does not currently handle !where with function and lines: {where}"
+                )
+            added_ids.append(self.debugger.add_function_breakpoint(where.function))
```
OCHyams, line 138: where is self.debugger set after setting it to None in init?

cross-project-tests/debuginfo-tests/dexter/dex/dextIR/DextIR.py
```diff
@@ -13,14 +13,22 @@
 from dex.test_script.Script import DexterScript
 
 
+def file_matches(a, b):
+    if os.path.exists(a) and os.path.exists(b):
+        return os.path.samefile(a, b)
+    return os.path.normpath(os.path.normcase(a)) == os.path.normpath(
+        os.path.normcase(b)
+    )
```
OCHyams, line 21: has this been formatted with black?

cross-project-tests/debuginfo-tests/dexter/feature_tests/scripts/debugging/simple_where_function.cpp
```diff
@@ -0,0 +1,64 @@
...
+int main() {
+    int Array[] = { 2, 4, 6, 8, 10 };
+    assign(Array + 1, 5);
+    replace(Array, 5, 8, 9);
+    return Array[1] + Array[3];
+}
+
+/// Test that we can use functions in !where nodes, and that Dexter steps through the entirety of those functions.
+/// We expect both calls to `assign` to be stepped through, but only the outermost call of `replace` should be stepped
+/// through, as the !where matches to the first (outermost) applicable frame.
```
OCHyams: This says "outermost" frame but the prints make it look like it's the "innermost" frame? Perhaps I'm looking at the call stack upside down, or am confused.


```
replace <-- b
replace <-- a
main
```
It looks like the steps report show dexter stepping through what I labelled "a"?

OCHyams (commented):
SGTM with some nit/questions

cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
SLTozer: It is set in the DebuggerControllerBase method that calls through to _run_debugger_custom of the particular debugger controller subclass; the function is `def run_debugger(self, debugger)`, and it sets `debugger` before the call and unsets it after.

cross-project-tests/debuginfo-tests/dexter/feature_tests/scripts/debugging/simple_where_function.cpp
SLTozer: Your reading is correct, but I think we disagree on terminology - I would call "b" the "innermost" frame and "a" the "outermost" frame. When `foo()` calls `bar()` for example, I'd call `foo` the "outer" function and `bar` the "inner" function in their relationship, since the call to `bar` is contained within `foo`. Similarly, when talking about "first" and "last" frames, I'd refer to them in call-order, i.e. "main" is the first function on the stack and the recursively-called "replace" is the "last".

I will say, I thought a bit about the terminology while writing this, and I could see multiple interpretations, but couldn't see an easy way to resolve it without just writing something like "outermost (i.e. closest to the entry point)" everywhere that I write first/last or inner/outer.

OCHyams: Hmmm, I feel this is ripe for future confusion as I agree with your first/last but don't find your inner/outer intuitive. I see "bar" as outer because it's closer to the leaf-side edge of the call tree.

I feel parent/child is clearer? ancestor/descendent. Though I'm not sure what comparative or superlative we use there, as "age" doesn't seem intuitive.

cross-project-tests/debuginfo-tests/dexter/dex/dextIR/DextIR.py
SLTozer: Yes, applying `black` to this file leaves it unchanged.

cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
OCHyams: possibly worth adding a note to the __init__ where the variable is set to None? ymmv, happy for you to decide

cross-project-tests/debuginfo-tests/dexter/dex/debugger/DebuggerControllers/ScriptDebuggerController.py
```diff
@@ -25,20 +25,20 @@
...
 
-def paths_differ(expected: str, actual: str) -> bool:
-    """Returns True if expected is not a trailing subpath of actual, i.e. if `actual` does not end with `expected` after
-    normalizing both paths."""
-    normalized_expected: str = os.path.normcase(os.path.normpath(expected))
-    normalized_actual: str = os.path.normcase(os.path.normpath(actual))
-    return not normalized_actual.endswith(normalized_expected)
+def is_subpath(subpath: str, superpath: str) -> bool:
+    """Returns True if subpath is not a trailing subpath of superpath, i.e. if `superpath` does not end with `subpath`
+    after normalizing both paths."""
```
OCHyams: I think the docstring is out of date

OCHyams (approved):
I think outermost is mentioned a few times still. Are you settling on root/leaf rootmost / leafmost for call-stack "direction"? If so, that sounds fine to me. IMO it would be good to have it applied everywhere

PR #197422, opened 2026-05-13
[MLIR][NVVM] Add sqrt Ops

varadk27:
Adds two NVVM dialect ops covering all 14 floating-point `sqrt` forms:

- `nvvm.sqrt` -- IEEE-compliant sqrt with explicit rounding mode
  (`sqrt.<RM>[.ftz].{f32,f64}`), 12 forms.
- `nvvm.sqrt.approx` -- fast approximate sqrt (`sqrt.approx[.ftz].f32`),
  2 forms; uses the `NVVM_F32UnaryApproxOp` base class.

The two ops are split because the rounded forms require an explicit
rounding mode and support both f32 and f64, while the approx forms have
no rounding mode and are f32-only.

### DIFF
--- a/mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td
+++ b/mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td
@@ -6552,7 +6552,7 @@ def NVVM_SqrtOp
     : NVVM_SingleResultIntrinsicOp<"sqrt", [Pure, SameOperandsAndResultType]> {
   let summary = "Take the square root of a value";
   let description = [{
-    Compute sqrt(a) and store the result in d.
+    Compute sqrt(src) and store the result in res.
 
     For more information, see PTX ISA:
     [sqrt](https://docs.nvidia.com/cuda/parallel-thread-execution/#floating-point-instructions-sqrt)
@@ -6568,7 +6568,7 @@ def NVVM_SqrtApproxOp : NVVM_F32UnaryApproxOp<"sqrt.approx"> {
   let summary = "Square root (fast approximation)";
   let description = [{
     Computes a fast approximation of the square root of the input value
-    (`d = sqrt(a)`). The maximum relative error over the entire positive
+    (`res = sqrt(src)`). The maximum relative error over the entire positive
     finite range is 2^-23.
 
     The `ftz` attribute, when set, flushes subnormal inputs and results to
--- a/mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td
+++ b/mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td
@@ -6545,6 +6545,41 @@ def NVVM_FmaOp : NVVM_Op<"fma", [Pure, SameOperandsAndResultType]> {
 }
 
 //===----------------------------------------------------------------------===//
+// NVVM sqrt op definitions
+//===----------------------------------------------------------------------===//
+
+def NVVM_SqrtOp
+    : NVVM_SingleResultIntrinsicOp<"sqrt", [Pure, SameOperandsAndResultType]> {
+  let summary = "Take the square root of a value";
+  let description = [{
+    Compute sqrt(a) and store the result in d.
+
+    For more information, see PTX ISA:
+    [sqrt](https://docs.nvidia.com/cuda/parallel-thread-execution/#floating-point-instructions-sqrt)
+  }];
+  let arguments = (ins AnyTypeOf<[F32, F64]>:$src, FPArithRoundingMode:$rnd,
+      DefaultValuedAttr<BoolAttr, "false">:$ftz);
+  let results = (outs AnyTypeOf<[F32, F64]>:$res);
+  let assemblyFormat = "$src attr-dict `:` type($src)";
+  let hasVerifier = 1;
+}
+
+def NVVM_SqrtApproxOp : NVVM_F32UnaryApproxOp<"sqrt.approx"> {
+  let summary = "Square root (fast approximation)";
+  let description = [{
+    Computes a fast approximation of the square root of the input value
+    (`d = sqrt(a)`). The maximum relative error over the entire positive
+    finite range is 2^-23.
+
+    The `ftz` attribute, when set, flushes subnormal inputs and results to
+    sign-preserving zero.
+
+    For more information, see PTX ISA:
+    [sqrt](https://docs.nvidia.com/cuda/parallel-thread-execution/#floating-point-instructions-sqrt)
+  }];
+}
+
+//===----------------------------------------------------------------------===//
 // NVVM tensormap.replace Op
 //===----------------------------------------------------------------------===//
 
--- a/mlir/lib/Dialect/LLVMIR/IR/NVVMDialect.cpp
+++ b/mlir/lib/Dialect/LLVMIR/IR/NVVMDialect.cpp
@@ -3339,6 +3339,16 @@ LogicalResult NVVM::FmaOp::verify() {
   return success();
 }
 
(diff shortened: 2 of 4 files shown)


mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td
```diff
@@ -6544,6 +6544,41 @@ def NVVM_FmaOp : NVVM_Op<"fma", [Pure, SameOperandsAndResultType]> {
...
 
+//===----------------------------------------------------------------------===//
+// NVVM sqrt op definitions
+//===----------------------------------------------------------------------===//
+
+def NVVM_SqrtOp
+    : NVVM_SingleResultIntrinsicOp<"sqrt", [Pure, SameOperandsAndResultType]> {
+  let summary = "Take the square root of a value";
+  let description = [{
+    Compute sqrt(a) and store the result in d.
```
durga4github: let us use `src` and `res` instead of `a` and `d`.

mlir/lib/Dialect/LLVMIR/IR/NVVMDialect.cpp
```diff
@@ -3339,6 +3339,16 @@ LogicalResult NVVM::FmaOp::verify() {
   return success();
 }
 
+LogicalResult NVVM::SqrtOp::verify() {
+  if (getRnd() == NVVM::FPRoundingMode::NONE)
+    return emitOpError("rounding mode must be specified");
```
durga4github: "rounding mode cannot be None"

mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td
```diff
@@ -6544,6 +6544,41 @@ def NVVM_FmaOp : NVVM_Op<"fma", [Pure, SameOperandsAndResultType]> {
...
+
+    For more information, see PTX ISA:
+    [sqrt](https://docs.nvidia.com/cuda/parallel-thread-execution/#floating-point-instructions-sqrt)
+  }];
+  let arguments = (ins AnyTypeOf<[F32, F64]>:$src, FPArithRoundingMode:$rnd,
```
durga4github, line 6560: ok, it seems RN is the default rounding mode. So, why not make it a defaultValuedAttr with RN as the value?

mlir/lib/Dialect/LLVMIR/IR/NVVMDialect.cpp
```diff
@@ -3535,6 +3545,57 @@ RsqrtOp::getIntrinsicIDAndArgs(Operation &op, LLVM::ModuleTranslation &mt,
...
+                              llvm::IRBuilderBase &builder) {
+  auto thisOp = cast<NVVM::SqrtOp>(op);
+  Type t = thisOp.getRes().getType();
+  NVVM::FPRoundingMode rndMode = thisOp.getRnd();
+  bool isFtz = thisOp.getFtz();
+
+  // f32: 5 RM slots (NONE→RN, RN, RM, RP, RZ) × 2 ftz states.
+  static constexpr llvm::Intrinsic::ID f32IDs[] = {
+      llvm::Intrinsic::nvvm_sqrt_rn_f, // NONE → default RN (verifier rejects
+                                       // NONE anyway)
```
durga4github: I feel this is harder to read.
We can compute the rnd-mode-based-index as a standalone variable, with a -1 offset and use it everywhere.
That way, we don't need to duplicate rn-entries in the array.

mlir/lib/Dialect/LLVMIR/IR/NVVMDialect.cpp
```diff
@@ -3535,6 +3545,57 @@ RsqrtOp::getIntrinsicIDAndArgs(Operation &op, LLVM::ModuleTranslation &mt,
...
+      llvm::Intrinsic::nvvm_sqrt_rn_f,
+      llvm::Intrinsic::nvvm_sqrt_rm_f,
+      llvm::Intrinsic::nvvm_sqrt_rp_f,
+      llvm::Intrinsic::nvvm_sqrt_rz_f,
+      llvm::Intrinsic::nvvm_sqrt_rn_ftz_f,
+      llvm::Intrinsic::nvvm_sqrt_rn_ftz_f,
+      llvm::Intrinsic::nvvm_sqrt_rm_ftz_f,
+      llvm::Intrinsic::nvvm_sqrt_rp_ftz_f,
+      llvm::Intrinsic::nvvm_sqrt_rz_ftz_f,
+  };
```
durga4github, line 3571: Another option could be to use a different array for `f32FTZIds[]`...

mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td
schwarzschild-radius: There are no default rounding mode for sqrt, right? (https://docs.nvidia.com/cuda/parallel-thread-execution/#floating-point-instructions-sqrt)

mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td
varadk27: Addressed in the latest revision.

mlir/lib/Dialect/LLVMIR/IR/NVVMDialect.cpp
varadk27: Updated in the latest revision.

mlir/lib/Dialect/LLVMIR/IR/NVVMDialect.cpp
varadk27: Done , used a single `rndIndex = static_cast<unsigned>(rndMode) - 1` and split into separate `f32IDs `/ `f32FTZIDs `/ `f64IDs `arrays (4 entries each, no duplicate RN).

mlir/lib/Dialect/LLVMIR/IR/NVVMDialect.cpp
varadk27: Updated in latest commit.

mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td
durga4github: ok, I asked here looking at the comment below (where we were indexing earlier).
So, yes, in this case, defValueAttr does not make sense. We can leave it as is.

durga4github (approved):
The latest revision LGTM.

mlir/lib/Dialect/LLVMIR/IR/NVVMDialect.cpp
durga4github: ok, thanks!

zackc6 (approved):

mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td
```diff
@@ -6544,6 +6544,41 @@ def NVVM_FmaOp : NVVM_Op<"fma", [Pure, SameOperandsAndResultType]> {
...
+      DefaultValuedAttr<BoolAttr, "false">:$ftz);
+  let results = (outs AnyTypeOf<[F32, F64]>:$res);
+  let assemblyFormat = "$src attr-dict `:` type($src)";
+  let hasVerifier = 1;
+}
+
+def NVVM_SqrtApproxOp : NVVM_F32UnaryApproxOp<"sqrt.approx"> {
```
grypp, line 6567: Do we really need extra OP for `approx`? 
@durga4github @schwarzschild-radius @varadk27

PR #141900, opened 2025-05-29
[LV][RFC] Generating conditional VPBB that will be skip when the mask is inactive in VPlan.

ElvisWang123:
This patch create an overview of VPlan transformation of creating conditional VPBB and will be split off into multiple patches later.

RFC: https://discourse.llvm.org/t/rfc-lv-generating-conditional-vpbb-that-will-be-skip-when-the-mask-is-inactive-in-vplan/86591

This patch add the transformation that convert flatten control flow
with conditional vector basic block.
This transformation can help program skip masked operations without any
active lane.

First, this transformation will collect all masked stores and operands
bottom-up. And put these masked operations into a new vector basic
block.

Second, this transformation will split original vector loop and insert
the new basic block between split blocks. And update the conditional
branch in the original blocks.

E.g.
Before: {
  vector.loop:
    ...
    BranchOnCount %IV, %TC
  Successors middle.block, vector.loop
}

After: {
  vector.loop:
    ...
    %any.active.mask = any-of(%mask)
    BranchOnCount %any.active.mask, 0
  Successors vector.loop.split, vector.if.bb

  vector.if.bb:
    ... (Masked operations)
  Successors vector.loop.split

  vector.loop.split:
    ...
    BranchOnCount %IV, %TC
  Successors middle.block, vector.loop
}

Co-authored-by: nikolaypanchenko <kolya.panchenko@sifive.com>

PR #197423, opened 2026-05-13
[LV][VPlan] Introduce conditional block guards to skip inactive vector blocks (BOSCC)

nema-ashutosh:
Add a new VPlan transformation, introduceConditionalBlockGuards, that wraps conditionally-executed vector blocks in guard diamonds. When the vector mask has no active lanes, the entire block is bypassed via a scalar branch, avoiding unnecessary execution of expensive vector block.

RFC Link - https://discourse.llvm.org/t/rfc-lv-active-lane-guard-diamonds-skipping-conditionally-executed-vplan-blocks-when-no-lane-is-active/90779

This implements the Branch-On-Superword-Condition-Code (BOSCC) technique at the VPlan level, replacing the earlier approach proposed in D139074 with a simpler design that reuses existing VPlan building blocks (AnyOf, BranchOnCond, VPBlockUtils).

The transformation works in four steps:
  1. Identify candidate blocks that are conditionally executed in the original scalar loop and exceed a recipe-count threshold.
  2. Discover the active-lane mask from masked recipes, splitting the block if the mask definition is local.
  3. Build a guard diamond: GuardBlock (any-of + branch-on-cond) and JoinBlock around the candidate.
  4. Insert ConditionalMerge PHIs in the JoinBlock for values that escape the guarded block.

A new VPInstruction opcode, ConditionalMerge, is introduced to merge guarded values with poison on the skip path. It lowers to an IR PHI node.

The transformation is off by default (-enable-loop-vectorization-with- conditions=false) and disabled for scalable vector targets and blocks containing in-loop reductions.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
+++ b/llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
@@ -1637,6 +1637,23 @@ static void simplifyRecipe(VPSingleDefRecipe *Def, VPTypeAnalysis &TypeInfo) {
       !cast<VPInstruction>(Def)->isMasked())
     return Def->replaceAllUsesWith(Def->getOperand(0));
 
+  // When removeBranchOnConst folds a constant-true guard condition, it removes
+  // the GuardBlock -> JoinBlock edge and collapses the diamond into a
+  // straight-line chain. JoinBlock then has a single predecessor (VecBlock),
+  // so the ConditionalMerge phi is redundant — replace it with its operand.
+  // The operand necessarily dominates JoinBlock because it was valid at its
+  // use site inside VecBlock before the collapse, and collapsing only
+  // strengthens dominance in the resulting straight-line chain.
+  if (auto *VPI = dyn_cast<VPInstruction>(Def)) {
+    if (VPI->getOpcode() == VPInstruction::ConditionalMerge) {
+      if (VPI->getParent()->getSinglePredecessor()) {
+        Def->replaceAllUsesWith(Def->getOperand(0));
+        Def->eraseFromParent();
+        return;
+      }
+    }
+  }
+
   // Look through ExtractLastLane.
   if (match(Def, m_ExtractLastLane(m_VPValue(A)))) {
     if (match(A, m_BuildVector())) {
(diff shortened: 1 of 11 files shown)


nema-ashutosh:
Kindly review - @fhahn @preames @lukel97 @alexey-bataev @rengolin @ayalz

alexey-bataev:
There is similar https://github.com/llvm/llvm-project/pull/141900, why do we need this one?

ElvisWang123:
I’ve also created an llvm-test-suite patch to evaluate the performance of #141900. Would you like me to add any additional test cases, or would you be available to review it as is?
 https://github.com/llvm/llvm-test-suite/pull/345

nema-ashutosh:
> I’ve also created an llvm-test-suite patch to evaluate the performance of #141900. Would you like me to add any additional test cases, or would you be available to review it as is? [llvm/llvm-test-suite#345](https://github.com/llvm/llvm-test-suite/pull/345)

Thanks, I'll have a check

nema-ashutosh:
> There is similar #141900, why do we need this one?

I’ve been exploring a related approach in PR #197423 and would appreciate it if reviewers could give it a fair evaluation before it’s set aside.

There are a few scenarios where the block-level approach handles things more naturally, and I’ve highlighted those differences for consideration.

1. Store-Only Seeding
PR 141900: Only seeds from VPWidenStoreRecipe. Conditional blocks with only expensive FP math, masked loads/gathers, replicated recipes, or interleave groups won't trigger the optimization.
if (cond[i]) {
    temp += sqrt(A[i]) * log(B[C[i]]);  // no store -> not guarded
}
PR 197423: Identifies candidates at the block level -- any block conditionally executed in the original scalar loop qualifies, regardless of recipe types. Discovers masks from stores, loads, replicated recipes, and interleave groups.

2. No Support for Values Escaping the Guarded Block
PR 141900: Recipes with any user outside the tree are pruned. If the expensive computation itself is shared, the tree can shrink to size 1 and no guard is created.
PR 197423: Introduces ConditionalMerge PHIs in a join block that merge the guarded value with poison on the skip path. The expensive computation stays guarded; external users receive the merged value.

3. Redundant Guards for Stores Sharing the Same Mask
PR 141900: Each store is processed independently -- no grouping by mask. Two stores under the same condition produce two sequential any-of + branch pairs computing the identical value:
if (cond[i]) {
    A[i] = ExpensiveOp1(data[i]);
    B[i] = ExpensiveOp2(data[i]);
}
PR 197423: Guards the entire block as a single unit.

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -3810,6 +3833,217 @@ static void expandVPDerivedIV(VPDerivedIVRecipe *R, VPTypeAnalysis &TypeInfo) {
...
+    }
+  }
+}
+
+void VPlanTransforms::introduceConditionalBlockGuards(VPlan &Plan, Loop *OrigLoop,
+                                           DominatorTree *DT,
+                                           const TargetTransformInfo *TTI) {
+  if (!EnableConditionalBlockGuards)
+    return;
+  if (TTI->enableScalableVectorization())
```
nema-ashutosh, line 4008: Please note that this is a temporary block check for now. I’d like to remove it eventually, but since I have limited experience with scalable architecture, I would appreciate some guidance on how to do so correctly.

PR #197424, opened 2026-05-13
[AArch64] Select `CMP WZR, reg, shift #amount`

pfusik:


davemgreen (commented):
This sounds alright. There is a bit below that uses getCmpOperandFoldingProfit to do something similar. Could it handle wzr too, or could this make use of getCmpOperandFoldingProfit? Or does that not work as well? It might not work very well for extended registers if they do not accept wzr.

pfusik:
Yes, I basically duplicate the logic of `getCmpOperandFoldingProfit` here because `cmp` accepts the zero register with the shifted register, yet the stack pointer with the extended register.
It felt easier to do than adding this case to `getCmpOperandFoldingProfit`, but now I think the `isSupportedExtended` lambda could be changed to a free-standing function passed as a pointer as it doesn't capture anything. This should make the code simpler and not duplicate the logic.

pfusik:
Refactored now to reuse `getCmpOperandFoldingProfit` with an extra `bool` parameter.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -4278,8 +4281,8 @@ static SDValue getAArch64Cmp(SDValue LHS, SDValue RHS, ISD::CondCode CC,
     SDValue TheLHS = LHSIsCMN ? LHS.getOperand(1) : LHS;
     SDValue TheRHS = RHSIsCMN ? RHS.getOperand(1) : RHS;
 
-    if (getCmpOperandFoldingProfit(TheLHS) + (LHSIsCMN ? 1 : 0) >
-        getCmpOperandFoldingProfit(TheRHS) + (RHSIsCMN ? 1 : 0)) {
+    if (getCmpOperandFoldingProfit(TheLHS, true) + (LHSIsCMN ? 1 : 0) >
+        getCmpOperandFoldingProfit(TheRHS, true) + (RHSIsCMN ? 1 : 0)) {
```
pfusik, line 4285: In a later PR, I want to refactor this: same operations on LHS and RHS deserve to be done in one function.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -4137,33 +4138,32 @@ static SDValue emitConjunction(SelectionDAG &DAG, SDValue Val,
...
+    if (ConstantSDNode *MaskCst = dyn_cast<ConstantSDNode>(V.getOperand(1))) {
+      uint64_t Mask = MaskCst->getZExtValue();
+      return (Mask == 0xFF || Mask == 0xFFFF || Mask == 0xFFFFFFFF);
+    }
+  return false;
+}
 
+/// Returns how profitable it is to fold a comparison's operand's shift and/or
+/// extension operations.
+static unsigned getCmpOperandFoldingProfit(SDValue Op, bool extendOk) {
```
davemgreen: extendOk -> ExtendOk or maybe AllowExtend

davemgreen (approved):

PR #197426, opened 2026-05-13
[AMDGPU] Remove RCP_IFLAG combine

piotrAMD:
The combine was added in D48569 8 years ago with the aim of preserving flags, but the current LangRef says the status flags are not observable in the default FP environment.

The main motivation for this change is to enable scalar float reciprocal generation v_s_rcp_f32 on newer hardware. There is no v_s_rcp_iflag_f32, so the combine effectively blocks the selection.
See: pseudo-scalar-transcendental.ll.


llvm/test/CodeGen/AMDGPU/rcp_iflag.ll
```diff
@@ -1,46 +0,0 @@
-; RUN: llc -mtriple=amdgcn < %s | FileCheck --check-prefix=GCN %s
-
```
arsenm, line 2: Test shouldn't be deleted, should show the current behavior

piotrAMD:
Restored test and turned it into an auto-generated one.

arsenm (approved):
Not really sure what the point of the instruction is, we never have, and never will, care about the status flags

llvm/test/CodeGen/AMDGPU/udivrem24.ll
jayfoad: I don't understand this comment but it seems like it might be obsolete now? Same below in a few places.

llvm/lib/Target/AMDGPU/SIISelLowering.cpp
```diff
@@ -15463,12 +15463,6 @@ SDValue SITargetLowering::performRcpCombine(SDNode *N,
                                  SDLoc(N), VT);
   }
 
-  if (VT == MVT::f32 && (N0.getOpcode() == ISD::UINT_TO_FP ||
-                         N0.getOpcode() == ISD::SINT_TO_FP)) {
-    return DCI.DAG.getNode(AMDGPUISD::RCP_IFLAG, SDLoc(N), VT, N0,
```
jayfoad, line 15468: I think you can now remove all support for the RCP_IFLAG ISD node?

llvm/test/CodeGen/AMDGPU/udivrem24.ll
arsenm: we might want it, in strictfp functions, I suppose

piotrAMD: Thanks, removed.

llvm/lib/Target/AMDGPU/SIISelLowering.cpp
piotrAMD: I think Matt's reply was meant here: "we might want it, in strictfp functions, I suppose". So I'm keeping it.

piotrAMD:
Removed obsolete comments.

llvm/lib/Target/AMDGPU/SIISelLowering.cpp
jayfoad: I would prefer to remove it, since we can always bring it back if it's needed in future, but I won't insist.

jayfoad (approved):
LGTM

PR #196097, opened 2026-05-06
[LLVM][AsmWriter] Fix ConstantFP zeroinitializer check

paulwalker-arm:
It turns out ppc_fp128 has a value where isPosZero() returns true but isNullValue() returns false.

I wondered if DoubleAPFloat might be wrong but when trying to "fix" it I found more and more places that do not care about the second double, so I reverted to this localised fix.  The change to `Constant::isNullValue()` is not required, but given the comment I figured it was worth keeping the related code consistent.

PR #197427, opened 2026-05-13
[LLVM][Constants] Remove the option to disable vector ConstantFP support.

paulwalker-arm:
Removes the command line options:
  -use-constant-fp-for-fixed-length-splat
  -use-constant-fp-for-scalable-splat

A couple of bumps but I'm not aware of any more issues beyond https://github.com/llvm/llvm-project/pull/196097, which only affects IR printing.

I will follow up with a general clean up of the ConstantFP class whereby the construction functions can be simplified and they can all now return `ConstantFP*`. I've also investigated removing some instances of `getSplat()` where the existing `isa<ConstantFP>` handling should be sufficient, but I'm worried things like the `StringRef` variant of ConstantVector construction makes it hard to assume there will never be a ConstantDataVector based splat?


llvm/lib/IR/Constants.cpp
```diff
@@ -1665,24 +1659,25 @@ Constant *ConstantVector::getSplat(ElementCount EC, Constant *V) {
       return ConstantPointerNull::get(VTy);
   }
 
+  if (isa<ConstantByte>(V))
+    return ConstantByte::get(V->getContext(), EC,
+                             cast<ConstantByte>(V)->getValue());
```
arsenm: dyn_cast instead of isa+cast

nikic (approved):
LGTM

PR #197429, opened 2026-05-13
[AArch64][SelectionDAG] Fold late FCMP+CSEL patterns to FMAXNM/FMINNM during AArch64 DAG-to-DAG preprocessing.

usha1830:
This patch adds a late AArch64 SelectionDAG peephole for scalar floating-point patterns:
```
fcmp x, C
fcsel x, C
```

When the pattern is ordered, uses a non-NaN constant C, and the variable operand is known never-sNaN, it is folded directly to FMAXNM/FMINNM. The transform requires `nsz` only when C is zero, to avoid signed-zero mismatches.

The transform is intentionally performed in _AArch64DAGToDAGISel::PreprocessISelDAG_ rather than DAGCombine. 
Doing it earlier can be invalidated by later DAG combines that rewrite the defining operation of the operand.


llvm/lib/Target/AArch64/AArch64ISelDAGToDAG.cpp
```diff
@@ -7994,6 +7996,98 @@ bool AArch64DAGToDAGISel::SelectCmpBranchExtOperand(SDValue N, SDValue &Reg,
...
+    return SDValue();
+
+  // Only fold if variable operand is never sNaN.
+  // This runs after DAG combines, so later combines cannot remove a defining
+  // operation used by isKnownNeverSNaN().
+  if (!CurDAG->isKnownNeverSNaN(Operand))
+    return SDValue();
+
+  SDLoc DL(&N);
+  return CurDAG->getNode(IEEEOpc, DL, VT, Operand, Constant);
```
davemgreen: I was expecting to generate the machine node directly.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -12473,8 +12473,12 @@ SDValue AArch64TargetLowering::LowerSELECT_CC(
   }
 
   // Emit first, and possibly only, CSEL.
+  // Propagate nsz to the CSEL so downstream passes can use it.
+  SDNodeFlags CSELFlags;
+  CSELFlags.setNoSignedZeros(Flags.hasNoSignedZeros());
   SDValue CC1Val = getCondCode(DAG, CC1);
-  SDValue CS1 = DAG.getNode(AArch64ISD::CSEL, DL, VT, TVal, FVal, CC1Val, Cmp);
+  SDValue CS1 = DAG.getNode(AArch64ISD::CSEL, DL, VT, {TVal, FVal, CC1Val, Cmp},
+                            CSELFlags);
```
davemgreen: Just pass all flags?

davemgreen (commented):
Is nsz needed if the constant isnt 0?

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
usha1830: I kept only `nsz` since it is the only flag relevant for this CSEL fold (signed-zero semantics).

The `nnan`/`ninf` flags would require proving the property for both selected values, which we are not doing here. The other flags (`reassoc, arcp, contract, afn, etc.`) are not directly meaningful for a select, so I did not propagate them.

Please let me know if you think we should propagate additional flags here.

davemgreen: If we are making csel a potential FMF node then I would go the whole way and keep all the flags from the node that created it.

davemgreen (approved):
LGTM

usha1830:
> # 🐧 Linux x64 Test Results
> * 140865 tests passed
> * 2817 tests skipped
> * 1 test failed
> 
> ## Failed Tests
> (click on a test name to see its output)
> 
> ### LLVM
> LLVM.Transforms/ArgumentPromotion/BPF/argpromotion.ll
> ```
> Exit Code: 1
> 
> Command Output (stdout):
> --
> # RUN: at line 1
> /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/opt -passes=argpromotion -mtriple=bpf-pc-linux -S /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/Transforms/ArgumentPromotion/BPF/argpromotion.ll | /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/FileCheck /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/Transforms/ArgumentPromotion/BPF/argpromotion.ll
> # executed command: /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/opt -passes=argpromotion -mtriple=bpf-pc-linux -S /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/Transforms/ArgumentPromotion/BPF/argpromotion.ll
> # note: command had no output on stdout or stderr
> # executed command: /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/FileCheck /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/Transforms/ArgumentPromotion/BPF/argpromotion.ll
> # | /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/Transforms/ArgumentPromotion/BPF/argpromotion.ll:65:10: error: CHECK: expected string not found in input
> # | ; CHECK: i32 @foo1(ptr noundef readonly captures(none) %p1, ptr noundef readonly captures(none) %p2, ptr noundef readonly captures(none) %p3)
> # |          ^
> # | <stdin>:1:1: note: scanning from here
> # | ; ModuleID = '/home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/Transforms/ArgumentPromotion/BPF/argpromotion.ll'
> # | ^
> # | 
> # | Input file: <stdin>
> # | Check file: /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/Transforms/ArgumentPromotion/BPF/argpromotion.ll
> # | 
> # | -dump-input=help explains the following input dump.
> # | 
> # | Input was:
> # | <<<<<<
> # |           1: ; ModuleID = '/home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/Transforms/ArgumentPromotion/BPF/argpromotion.ll' 
> # | check:65     X~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ error: no match found
> # |           2: source_filename = "/home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/Transforms/ArgumentPromotion/BPF/argpromotion.ll" 
> # | check:65     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> # |           3: target datalayout = "e-m:e-p:64:64-i64:64-i128:128-n32:64-S128" 
> # | check:65     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> # |           4: target triple = "bpf-pc-linux" 
> # | check:65     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> # |           5:  
> # | check:65     ~
> # |           6: %struct.t = type { i32, i32, i32, i32, i32, i32, i32 } 
> # | check:65     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> # |           .
> # |           .
> # |           .
> # | >>>>>>
> 
> --
> ```
> 

This failure looks unrelated and is not reproducible locally with this patch applied on main.

jpienaar:
We saw a test failure post this change and checking to see if expected. Tried to automatically create a reproducer: 

```
; Alive2 verification: the FMAXNM fold is incorrect for qNaN inputs
; Paste this at https://alive2.llvm.org/ce/
;
; The peephole transforms:
;   %cmp = fcmp olt half %x, -1.0
;   %r = select i1 %cmp, half -1.0, half %x
; Into the equivalent of:
;   %r = call half @llvm.maxnum.f16(half %x, half -1.0)
;
; These differ when %x is NaN:
;   select: olt is false → returns %x (NaN)
;   maxnum: returns -1.0 (non-NaN operand)

define half @src(half %x) {
  %cmp = fcmp olt half %x, -1.000000e+00
  %r = select i1 %cmp, half -1.000000e+00, half %x
  ret half %r
}

define half @tgt(half %x) {
  %r = call half @llvm.maxnum.f16(half %x, half -1.000000e+00)
  ret half %r
}

declare half @llvm.maxnum.f16(half, half)
```

davemgreen:
Hi - I'm not sure if alive can prove this completely, but it does seem that those backward operands should not be producing a fminnm. I will put together a patch to fix it.

jpienaar:
Thanks! This was extracted from XLA compiling bf16 tanh, in case helpful I can dump the LLVM IR (but seeing as you are seeing potential fix, I could perhaps help test it as easier way)

davemgreen:
Thanks - does #203230 help?

PR #197431, opened 2026-05-13
[VPlan] Verify canonical IV has no users if there's a VPCurrentIterationPHI. NFC

lukel97:
During the EVL transform we replace all uses of the canonical IV with a VPCurrentIterationPHI (the variably-stepping canonical IV).

Once it's added nothing else should use the canonical IV except for increment for the branch condition, otherwise the plan will be incorrect on the penultimate iteration.

This verifies this property in preparation for work to move the EVL transformation earlier in the pipeline

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlanVerifier.cpp
+++ b/llvm/lib/Transforms/Vectorize/VPlanVerifier.cpp
@@ -96,10 +96,23 @@ bool VPlanVerifier::verifyPhiRecipes(const VPBasicBlock *VPBB) {
       return false;
     }
 
-    if (isa<VPCurrentIterationPHIRecipe>(RecipeI) &&
-        RecipeI->getIterator() != VPBB->begin()) {
-      errs() << "CurrentIteration PHI is not the first recipe\n";
-      return false;
+    if (isa<VPCurrentIterationPHIRecipe>(RecipeI)) {
+      if (RecipeI->getIterator() != VPBB->begin()) {
+        errs() << "CurrentIteration PHI is not the first recipe\n";
+        return false;
+      }
+      if (const VPValue *CanIV =
+              VPBB->getEnclosingLoopRegion()->getCanonicalIV()) {
+        if (!all_of(
+                CanIV->users(),
+                match_fn(m_Add(m_Specific(CanIV),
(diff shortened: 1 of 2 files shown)


llvm/lib/Transforms/Vectorize/VPlanVerifier.cpp
```diff
@@ -96,10 +96,23 @@ bool VPlanVerifier::verifyPhiRecipes(const VPBasicBlock *VPBB) {
...
-        RecipeI->getIterator() != VPBB->begin()) {
-      errs() << "CurrentIteration PHI is not the first recipe\n";
-      return false;
+    if (isa<VPCurrentIterationPHIRecipe>(RecipeI)) {
+      if (RecipeI->getIterator() != VPBB->begin()) {
+        errs() << "CurrentIteration PHI is not the first recipe\n";
+        return false;
+      }
+      if (const VPValue *CanIV =
+              VPBB->getEnclosingLoopRegion()->getCanonicalIV()) {
```
eas, line 105: Should we check that regions still exist in VPlan?

llvm/lib/Transforms/Vectorize/VPlanVerifier.cpp
```diff
@@ -96,10 +96,23 @@ bool VPlanVerifier::verifyPhiRecipes(const VPBasicBlock *VPBB) {
...
+        if (!all_of(
+                CanIV->users(),
+                match_fn(m_Add(m_Specific(CanIV),
+                               m_Specific(&VPBB->getPlan()->getVFxUF()))))) {
```
eas, line 109: How hard would it be to address this so that we could just assert it has no uses?

fhahn: I think at the point when there's a region there will still be the user of the add that controls exiting of the loop. May be good to check that there's a single Add user that controls the regions terminator, although I am not sure how much work that would be

llvm/lib/Transforms/Vectorize/VPlanVerifier.cpp
fhahn: I think currently this works because we don't verify in plain-cfg mode, but I think this should be fixed https://github.com/llvm/llvm-project/pull/203082

PR #159942, opened 2025-09-20
[SCEV] Rewrite more SCEVAddExpr when applying guards.

fhahn:
When re-writing SCEVAddExprs to apply information from guards, check if
we have information for the expression itself. If so, apply it.

When we have an expression of the form (Const + A),  check if we have
have guard info for (Const + 1 + A) and use it. This is needed to avoid
regressions in a few cases, where we have BTCs with a subtracted
constant.

Rewriting expressions could cause regressions, e.g. when comparing 2
SCEV expressions where we are only able to rewrite one side, but I could
not find any cases where this happens more with this patch in practice.

Depends on https://github.com/llvm/llvm-project/pull/160012.

Proofs for some of the test changes: https://alive2.llvm.org/ce/z/RPX6t_

PR #197433, opened 2026-05-13
[SCEV] Add missing Map.lookup in SCEVLoopGuardRewriter visitAddExpr/v…

RonDahan101:
…isitMulExpr

SCEVLoopGuardRewriter's visitAddExpr and visitMulExpr methods never check the rewrite map for the full expression before rewriting operands. All other visit methods (visitUnknown, visitUMinExpr, visitSMinExpr, visitZeroExtendExpr, visitSignExtendExpr) correctly call Map.lookup(Expr) at entry, but visitAddExpr and visitMulExpr were missing this lookup.

The bug manifests when applyLoopGuards creates rewrite map entries keyed by a MulExpr or AddExpr. For example, divisibility-aware NE-zero guard handling creates a map entry for the full MulExpr (16 * (n /u 16)) derived from (n & -16), but visitMulExpr only rewrites individual operands and never finds the map entry for the whole expression. This causes downstream passes like LoopVectorize to generate unnecessary runtime checks (e.g., min.iters.check) even when the trip count is provably >= VF.

The fix adds Map.lookup(Expr) at the entry of both visitAddExpr and visitMulExpr, consistent with all other visitor methods in SCEVLoopGuardRewriter.

This also removes a now-redundant diff-based memory check in runtime-checks-difference.ll, where the improved guard propagation makes the vector.memcheck block unnecessary.

### DIFF
--- a/llvm/lib/Analysis/ScalarEvolution.cpp
+++ b/llvm/lib/Analysis/ScalarEvolution.cpp
@@ -16238,6 +16238,8 @@ const SCEV *ScalarEvolution::LoopGuards::rewrite(const SCEV *Expr) const {
     }
 
     const SCEV *visitAddExpr(const SCEVAddExpr *Expr) {
+      if (const SCEV *S = Map.lookup(Expr))
+        return S;
       // Helper to check if S is a subtraction (A - B) where A != B, and if so,
       // return UMax(S, 1).
       auto RewriteSubtraction = [&](const SCEV *S) -> const SCEV * {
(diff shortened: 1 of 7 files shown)


fhahn (commented):
There's something similar here https://github.com/llvm/llvm-project/pull/159942. There were some compile-time increases which is why I've not merged it yet, as I am currently still trying to see how we can reduce the impact. IIRC the main impact was actually from the early lookup

RonDahan101:
> he main impact was actually from the early lo
 
@fhahn
This PR doesn't effect the `visitMulExpr` which is the one I acutally care about 

I don't mind reverting the `visitAddExpr`  changes if it helps

RonDahan101:
@fhahn @nikic ping

RonDahan101:
> > he main impact was actually from the early lo
> 
> @fhahn This PR doesn't effect the `visitMulExpr` which is the one I acutally care about
> 
> I don't mind reverting the `visitAddExpr` changes if it helps

ping

PR #194342, opened 2026-04-27
[AArch64] Simplify codegen for LRINT.

sdesmalen-arm:
We don't need to use FP_TO_SINT_SAT, because according to the LangRef floating point values that don't fit the integer domain return a non-deterministic (freeze(poison)) value. The saturation code is quite verbose, so freezing the result seems like a better approach.

This was also discussed here:
https://github.com/llvm/llvm-project/pull/89035#discussion_r1595859634

We could even optimise away the freeze when we know that the operands to AArch64ISD::FCVTZS_MERGE_PASSTHRU are not poison.

PR #197435, opened 2026-05-13
[LLVM][CodeGen] When expanding ISD::LRINT, non-deterministic results should be frozen.

paulwalker-arm:
This comes from https://github.com/llvm/llvm-project/pull/194342#discussion_r3155677450.

### DIFF
--- a/llvm/lib/CodeGen/SelectionDAG/LegalizeDAG.cpp
+++ b/llvm/lib/CodeGen/SelectionDAG/LegalizeDAG.cpp
@@ -4636,9 +4636,11 @@ bool SelectionDAGLegalize::ExpandNode(SDNode *Node) {
     SDValue Arg = Node->getOperand(0);
     EVT ArgVT = Arg.getValueType();
     EVT ResVT = Node->getValueType(0);
-    SDLoc dl(Node);
-    SDValue RoundNode = DAG.getNode(ISD::FRINT, dl, ArgVT, Arg);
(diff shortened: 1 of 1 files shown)


llvm/lib/CodeGen/SelectionDAG/LegalizeDAG.cpp
```diff
@@ -4636,9 +4636,11 @@ bool SelectionDAGLegalize::ExpandNode(SDNode *Node) {
...
     EVT ArgVT = Arg.getValueType();
     EVT ResVT = Node->getValueType(0);
-    SDLoc dl(Node);
-    SDValue RoundNode = DAG.getNode(ISD::FRINT, dl, ArgVT, Arg);
-    Results.push_back(DAG.getNode(ISD::FP_TO_SINT, dl, ResVT, RoundNode));
+    SDLoc DL(Node);
+    SDValue RoundNode = DAG.getNode(ISD::FRINT, DL, ArgVT, Arg);
+    SDValue ConvertNode = DAG.getNode(ISD::FP_TO_SINT, DL, ResVT, RoundNode);
+    // Non-deterministic results are equivalent to freeze poison.
+    Results.push_back(DAG.getFreeze(ConvertNode));
```
arsenm, line 4643: Shouldn't this freeze the FRINT directly? Usually freeze is pushed as far up as possible

paulwalker-arm: ISD::FP_TO_SINT is the operation that can generate a poison result.

arsenm (approved):

PR #188248, opened 2026-03-24
[IR] Remove pointer arguments from loop.dependence.{war|raw}.mask

MacDue:
Passing pointer arguments is quite inconvenient for practical use. These intrinincs only care about the difference between the address bits of two pointers, but by taking pointer arguments it is implying they care about other details of the pointers (such as addrspace, or other pointer bits). 

This metadata is currently not preserved when gathering diff checks in the loop vectorizer, so requiring pointers makes emitting these intrinsic more complex (or requires preserving information that is ultimately unused).

PR #197437, opened 2026-05-13
[SDAG] Specify unsigned compares for loop.dependence.{war|raw} masks

MacDue:
Previously, the LangRef was ambiguous about the sign of comparisons used to create the loop dependence masks. This resulted in the expansion not following the intended semantics for extreme inputs.

For example, %ptrA = 0, %ptrB = UINT_MAX, should result in a (RAW) mask with all lanes active. However, previously we'd do ``(%elementSize * lane) < abs(%ptrB - %ptrA)``, which due to incorrectly using signed arithmetic would result in a mask with a single lane active as ``abs(%ptrB - %ptrA)`` resulted in 1, not `UINT_MAX`. In other words, ``abs(%ptrB - %ptrA)`` should be ``unsigned-absolute-difference(%ptrA, %ptrB)``.

Follow up to #188248.

### DIFF
--- a/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp
+++ b/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp
@@ -10591,30 +10591,44 @@ SDValue TargetLowering::expandLoopDependenceMask(SDNode *N,
   ElementCount LaneOffsetEC =
       ElementCount::get(N->getConstantOperandVal(3), VT.isScalableVT());
 
+  // The first lane is always active, so v1i1 => true.
+  if (LaneOffsetEC.isZero() &&
+      VT.getVectorElementCount() == ElementCount::getFixed(1))
+    return DAG.getBoolConstant(true, DL, VT, VT);
+
   EVT AddrVT = SourceValue->getValueType(0);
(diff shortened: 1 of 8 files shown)


MacDue:
Ping

MacDue:
Ping

llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp
```diff
@@ -11251,30 +11251,44 @@ SDValue TargetLowering::expandLoopDependenceMask(SDNode *N,
   ElementCount LaneOffsetEC =
       ElementCount::get(N->getConstantOperandVal(3), VT.isScalableVT());
 
+  // The first lane is always active, so v1i1 => true.
+  if (LaneOffsetEC.isZero() &&
+      VT.getVectorElementCount() == ElementCount::getFixed(1))
+    return DAG.getBoolConstant(true, DL, VT, VT);
```
sdesmalen-arm: Should this be a DAG combine instead?

sdesmalen-arm (approved):
Seems like a sensible change to me.

llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp
MacDue: Sure, can be :+1:

PR #197438, opened 2026-05-13
[clang-tidy] Fix false positives about reinitialization detection in `bugprone-use-after-move`

### LINKED ISSUE #62206: [clang-tidy] bugprone-use-after-move false positive
clang-tidy 15.0.7 produces a warning in the following code on the commented line.
I believe it is a false positive because it is being reassigned.
 
```cpp
#include <memory>
#include <variant>

struct expression;

struct identifier_expression {};

enum class binary_operator {
	add,
	assign,
	assign_add,
};

struct binary_expression {
	binary_operator op;
	std::shared_ptr<expression> lhs;
	std::shared_ptr<expression> rhs;
};

using expression_variant = std::variant<identifier_expression, binary_expression>;
struct expression : expression_variant {
	using expression_variant::expression_variant;
	using expression_variant::operator=;
};

template<typename Value, typename T = std::remove_reference_t<Value>>
auto wrap(Value&& init) -> std::unique_ptr<T>
{
	return std::make_unique<T>(std::forward<Value>(init));
}

auto simplify_binary_expr(binary_expression expr) -> binary_expression
{
	expression lhs = std::move(*expr.lhs);
	expression rhs = std::move(*expr.rhs);
	if (expr.op == binary_operator::assign_add) {
		rhs = binary_expression{ // 'rhs' used after it was moved [bugprone-use-after-move]
			.op = binary_operator::add,
			.lhs = wrap(lhs),
			.rhs = wrap(std::move(rhs)),
		};
		expr.op = binary_operator::assign;
	}

	return {
		.op = expr.op,
		.lhs = wrap(std::move(lhs)),
		.rhs = wrap(std::move(rhs)),
	};
}
```

flovent:
When calling base class's `operator=` through derived object, a implicit cast with `UncheckedDerivedToBase` will be generated:
```
void foo() {
  Base b;
  Derived d;
  std::move(d);
  d = b;
}
```
AST for `d = b`'s `d`:  
```
        |-ImplicitCastExpr <col:3> 'GH62206::Base' lvalue <UncheckedDerivedToBase (Base)>
        | `-DeclRefExpr <col:3> 'Derived' lvalue Var 0x1d11a400 'd' 'Derived'
```

This patch considers possible `implicitCastExpr` in the reinit matcher, i didn't find other cast situtaions which may causes other problems, so i use `ignoringImpCasts` directly rather than specific to `CK_UncheckedDerivedToBase` here.

Closes #62206.


clang-tools-extra/test/clang-tidy/checkers/bugprone/use-after-move.cpp
```diff
@@ -1790,3 +1790,20 @@ void Run() {
...
+    using Base::operator=;
+  };
+
+  void foo() {
+    Base b;
+    Derived d;
+    std::move(d);
+    d = b; // Should not warn
+  }
+}
```
zeyi2:
```suggestion
} // namespace GH62206
```

zeyi2 (approved):
LGTM with a nit, thank you

clang-tools-extra/clang-tidy/bugprone/UseAfterMoveCheck.cpp
```diff
@@ -132,7 +132,8 @@ makeReinitMatcher(const ValueDecl *MovedVariable,
                  // operator, test for built-in assignment as well, since
                  // template functions may be instantiated to use std::move() on
                  // built-in types.
-                 binaryOperation(hasOperatorName("="), hasLHS(DeclRefMatcher)),
+                 binaryOperation(hasOperatorName("="),
+                                 hasLHS(ignoringImpCasts(DeclRefMatcher))),
```
vbvictor: Could you also please test `d = (b)`, maybe we should write `ignoringParenImpCasts` here instead

flovent: I think you mean `(d) = b`? yes, i change to `ignoringParenImpCasts` now.

vbvictor (approved):
lgtm

PR #197439, opened 2026-05-13
[libc] Implement the linux-specific memfd_create syscall wrapper

labath:
I'm using the MFD constants from the kernel header as that's what was done for mmap, though it would be relatively simple to declare these ourselves, as they are not architecture-dependent.


kaladron:
> I'm using the MFD constants from the kernel header as that's what was done for mmap, though it would be relatively simple to declare these ourselves, as they are not architecture-dependent.

I would really like us to use more headers from the Linux kernel, they're just in poor shape.  They should be responsible for their own interface.  If these are working, that's great.

kaladron (approved):

labath:
The memfd headers are very clean, just defining the needed constants -- maybe because they're new.

PR #197441, opened 2026-05-13
[AArch64][clang] Improve -march= error message with many feature flags

jthackray:
When calling `clang` with a large number of feature flags, the entire
argument is printed as an error message if one of the feature flags is
invalid.

For example, before this change, when providing a large number of features
to `-march=` with one of them invalid, an error message such as this is printed:
```
  clang: error: unsupported argument 'armv9.6a+sme2+sme2p1+sve2+sve2p1+profile
  +crypto+aes+sha2+sha3+sm4+memtag+ssbs+bf16+i8mm+dotprod+ls64+rcpc3+brbe+gcs
  +faminmax+fp8+fp8fma+fp8dot4+fp8dot2+sme-f8f32+the+lut+lsui+pops+occmo
  +rme-gpc3+d128+invalidfeature'
```
and a user doesn't know which of the `+feature` flags is actually invalid.

After this change, the following error message is printed:
```
  clang: error: unsupported argument '+invalidfeature' to option '-march='
```
clearly printing out the first invalid feature flag found.

### DIFF
--- a/clang/lib/Driver/ToolChains/Arch/AArch64.cpp
+++ b/clang/lib/Driver/ToolChains/Arch/AArch64.cpp
@@ -137,8 +137,10 @@ aarch64::getAArch64TargetTuneCPU(const llvm::opt::ArgList &Args,
 }
 
 // Decode AArch64 features from string like +[no]featureA+[no]featureB+...
-static bool DecodeAArch64Features(const Driver &D, StringRef text,
-                                  llvm::AArch64::ExtensionSet &Extensions) {
+static bool
+DecodeAArch64Features(const Driver &D, StringRef text,
+                      llvm::AArch64::ExtensionSet &Extensions,
+                      std::optional<std::string> *InvalidArg = nullptr) {
   SmallVector<StringRef, 8> Split;
   text.split(Split, StringRef("+"), -1, false);
 
@@ -147,8 +149,11 @@ static bool DecodeAArch64Features(const Driver &D, StringRef text,
       D.Diag(clang::diag::err_drv_no_neon_modifier);
       continue;
     }
-    if (!Extensions.parseModifier(Feature))
+    if (!Extensions.parseModifier(Feature)) {
+      if (InvalidArg)
+        InvalidArg->emplace(("+" + Feature).str());
       return false;
+    }
   }
 
   return true;
@@ -203,7 +208,8 @@ static bool DecodeAArch64Mcpu(const Driver &D, StringRef Mcpu,
 static bool
 getAArch64ArchFeaturesFromMarch(const Driver &D, StringRef March,
                                 const ArgList &Args,
-                                llvm::AArch64::ExtensionSet &Extensions) {
+                                llvm::AArch64::ExtensionSet &Extensions,
+                                std::optional<std::string> &InvalidArg) {
   std::string MarchLowerCase = March.lower();
   std::pair<StringRef, StringRef> Split = StringRef(MarchLowerCase).split("+");
 
@@ -212,13 +218,15 @@ getAArch64ArchFeaturesFromMarch(const Driver &D, StringRef March,
 
   const llvm::AArch64::ArchInfo *ArchInfo =
       llvm::AArch64::parseArch(Split.first);
-  if (!ArchInfo)
+  if (!ArchInfo) {
+    InvalidArg.emplace(Split.first.str());
     return false;
+  }
 
   Extensions.addArchDefaults(*ArchInfo);
 
   if ((Split.second.size() &&
-       !DecodeAArch64Features(D, Split.second, Extensions)))
+       !DecodeAArch64Features(D, Split.second, Extensions, &InvalidArg)))
     return false;
 
   return true;
@@ -253,6 +261,7 @@ void aarch64::getAArch64TargetFeatures(const Driver &D,
                                        bool ForAS, bool ForMultilib) {
   Arg *A;
   bool success = true;
+  std::optional<std::string> InvalidArg;
   llvm::StringRef WaMArch;
   llvm::AArch64::ExtensionSet Extensions;
   if (ForAS)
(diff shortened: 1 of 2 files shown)


jthackray:
* **#197640** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197640?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#197441** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197441?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197441?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

clang/lib/Driver/ToolChains/Arch/AArch64.cpp
```diff
@@ -137,8 +137,10 @@ aarch64::getAArch64TargetTuneCPU(const llvm::opt::ArgList &Args,
 }
 
 // Decode AArch64 features from string like +[no]featureA+[no]featureB+...
-static bool DecodeAArch64Features(const Driver &D, StringRef text,
-                                  llvm::AArch64::ExtensionSet &Extensions) {
+static bool
+DecodeAArch64Features(const Driver &D, StringRef text,
+                      llvm::AArch64::ExtensionSet &Extensions,
+                      std::optional<std::string> *InvalidArg = nullptr) {
```
smithp35, line 143: If this is implemented for mtune and mcpu then this can be a non-defaulted reference as all call-sites would pass in InvalidArg.

It may also be possible to use a StringRef rather than a String as it looks like the original String will still exist at time of printing the error message.

In theory we could have DecodeAArch64Features return std::optional with InvalidArg, but that would invert all the tests. I guess `llvm::expected<bool, string>` could work.

smithp35 (commented):
Would it be possible to implement this for mcpu and mtune too? I expect the large number of extensions would be less common for mcpu, and non-existent for mtune, there's only these call sites, and would allow the parameter to be a non-defaulted reference.

This could be done as a follow up patch though as doing this for -march is already an improvement.

jthackray:
> Would it be possible to implement this for mcpu and mtune too? I expect the large number of extensions would be less common for mcpu, and non-existent for mtune, there's only these call sites, and would allow the parameter to be a non-defaulted reference.
> 
> This could be done as a follow up patch though as doing this for -march is already an improvement.

Sure, happy to do a follow-up patch to this one.

clang/lib/Driver/ToolChains/Arch/AArch64.cpp
jthackray: > It may also be possible to use a StringRef rather than a String as it looks like the original String will still exist at time of printing the error message.

I tried using a `StringRef` instead, but the diff is actually larger and more complicated than using a `string` because the parser lowercases/splits the string in-situ.

smithp35: > > It may also be possible to use a StringRef rather than a String as it looks like the original String will still exist at time of printing the error message.
> 
> I tried using a `StringRef` instead, but the diff is actually larger and more complicated than using a `string` because the parser lowercases/splits the string in-situ.

Thanks for trying, can leave it as it is.

jthackray:
> Sure, happy to do a follow-up patch to this one.

@smithp35 Follow-up patch in https://github.com/llvm/llvm-project/pull/197640.

DavidSpickett:
I was looking at yet another support request along these lines and thinking how annoying it is to have to bisect the flags. So great idea!

This will stop at the first invalid flag, correct? Which is a big improvement but you should clarify that behaviour in the PR description.

"Invalid arguments X Y Z" would be nice but I expect it doesn't fit nicely into the existing error formats so no need to pursue that now.

clang/lib/Driver/ToolChains/Arch/AArch64.cpp
```diff
@@ -147,8 +149,11 @@ static bool DecodeAArch64Features(const Driver &D, StringRef text,
       D.Diag(clang::diag::err_drv_no_neon_modifier);
       continue;
     }
-    if (!Extensions.parseModifier(Feature))
+    if (!Extensions.parseModifier(Feature)) {
+      if (InvalidArg)
+        InvalidArg->emplace(("+" + Feature).str());
```
DavidSpickett, line 154: Do we know that this will always be a plus?

Actually I think I just remembered why we can assume this. For `-march` we don't have `-feature` we have `+nofeature`.

Therefore, it's always `+<feature>` or `+no<feature>`, correct?

DavidSpickett (approved):

clang/test/Driver/aarch64-march.c
```diff
@@ -29,3 +29,10 @@
 // RUN: %clang --target=aarch64_be -mbig-endian -march=ARMv8.1a -### -c %s 2>&1 | FileCheck -check-prefix=GENERICV81A-BE %s
 // RUN: %clang --target=aarch64_be -mbig-endian -march=ARMV8.1-a -### -c %s 2>&1 | FileCheck -check-prefix=GENERICV81A-BE %s
 // GENERICV81A-BE: "-cc1"{{.*}} "-triple" "aarch64_be{{.*}}" "-target-cpu" "generic" "-target-feature" "+v8.1a"
+
+// ================== Check whether -march diagnoses the specific invalid argument.
```
DavidSpickett: Perhaps "the first invalid argument", because it will stop at the first invalid feature.

clang/lib/Driver/ToolChains/Arch/AArch64.cpp
jthackray: Yes, agree.

clang/test/Driver/aarch64-march.c
jthackray: Thanks, adjusted.

DavidSpickett (approved):

PR #197442, opened 2026-05-13
[Flang][OpenMP][NFC] Track Objects for BlockArgs

Stylie777:
When lowering a BlockArg in OpenMP, currently the symbol is tracked. This can however cause issues later on down the line as information may be lost relating to an expression. For example, an ArrayElement will be represented by its symbol, in this case the full array. This is not ideal as its just he ArrayElement that is intended to be represented.

Now, the object is tracked instead of the Symbol. For cases where the symbol is required, appropriate API is available to retrieve this information. This change opens the ability to better handle lowering of expressions such as Array Elements.

Assisted-by: Codex

### DIFF
--- a/flang/include/flang/Support/OpenMP-utils.h
+++ b/flang/include/flang/Support/OpenMP-utils.h
@@ -9,25 +9,35 @@
 #ifndef FORTRAN_SUPPORT_OPENMP_UTILS_H_
 #define FORTRAN_SUPPORT_OPENMP_UTILS_H_
 
+#include "flang/Lower/OpenMP/Clauses.h"
 #include "flang/Semantics/symbol.h"
 
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/Value.h"
 
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/SmallVector.h"
 
 namespace Fortran::common::openmp {
 /// Structure holding the information needed to create and bind entry block
 /// arguments associated to a single clause.
 struct EntryBlockArgsEntry {
-  llvm::ArrayRef<const Fortran::semantics::Symbol *> syms;
+  llvm::SmallVector<Fortran::lower::omp::Object> objects;
   llvm::ArrayRef<mlir::Value> vars;
 
   bool isValid() const {
-    // This check allows specifying a smaller number of symbols than values
+    // This check allows specifying a smaller number of objects than values
     // because in some case cases a single symbol generates multiple block
     // arguments.
-    return syms.size() <= vars.size();
+    return objects.size() <= vars.size();
+  }
+
+  llvm::SmallVector<const Fortran::semantics::Symbol *> getSyms() const {
+    llvm::SmallVector<const Fortran::semantics::Symbol *> syms;
+    for (const Fortran::lower::omp::Object &object : objects) {
+      syms.push_back(object.sym());
+    }
+    return syms;
   }
 };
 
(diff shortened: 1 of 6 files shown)


flang/include/flang/Support/OpenMP-utils.h
```diff
@@ -9,25 +9,35 @@
...
     // arguments.
-    return syms.size() <= vars.size();
+    return objects.size() <= vars.size();
+  }
+
+  llvm::SmallVector<const Fortran::semantics::Symbol *> getSyms() const {
+    llvm::SmallVector<const Fortran::semantics::Symbol *> syms;
+    for (const Fortran::lower::omp::Object &object : objects) {
+      syms.push_back(object.sym());
+    }
```
tblah: nit: syms.resize(objects.size());

flang/include/flang/Support/OpenMP-utils.h
```diff
@@ -50,10 +60,22 @@ struct EntryBlockArgs {
         useDeviceAddr.isValid() && useDevicePtr.isValid();
   }
 
-  auto getSyms() const {
-    return llvm::concat<const semantics::Symbol *const>(hasDeviceAddr.syms,
-        inReduction.syms, map.syms, priv.syms, reduction.syms,
-        taskReduction.syms, useDeviceAddr.syms, useDevicePtr.syms);
+  llvm::SmallVector<const semantics::Symbol *> getSyms() const {
+    llvm::SmallVector<const semantics::Symbol *> syms;
```
tblah, line 64: nit: resize first to avoid extra allocations.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1541,26 +1541,32 @@ bool ClauseProcessor::processIf(
...
-    llvm::SmallVectorImpl<const semantics::Symbol *> &reductionSyms) {
-  const auto &objectList{std::get<omp::ObjectList>(reduction.t)};
-  for (const Object &object : objectList) {
-    const semantics::Symbol *symbol = object.sym();
-    reductionSyms.push_back(symbol);
-  }
+void collectReductionObjects(const T &reduction,
+                             llvm::SmallVectorImpl<Object> &reductionObjects) {
+  const omp::ObjectList &objectList{std::get<omp::ObjectList>(reduction.t)};
+  llvm::copy(objectList, std::back_inserter(reductionObjects));
```
tblah, line 1548: nit: This has me thinking all of the for loops above should also use `llvm::copy`/`llvm::transform` - it is a matter of taste which you go with (I can't find any rule in LLVM docs) but we should at least stick with one method instead of mixing explicit loops with helpers in the same patch.

Even with llvm::copy these will still benefit from a resize call.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1541,26 +1541,32 @@ bool ClauseProcessor::processIf(
...
+}
+
+static llvm::SmallVector<const semantics::Symbol *>
+getObjectsSyms(llvm::ArrayRef<Object> objects) {
+  llvm::SmallVector<const semantics::Symbol *> syms;
```
tblah, line 1553: resize (there are many more cases later in the diff)

tblah (approved):
Just some nits. Thanks for pulling this out into another PR

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
Stylie777: All done

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
Stylie777: Agreed, and I have gone with llvm::copy/llvm:transform/llvm::copy_if where appropriate. There is one for loop left, but to do this with llvm::transform and llvm::copy_if would need a new vector creating of filtered objects and then copying these symbols. So I have just left this as a for loop to have 1 iteration space, rather than 2.

flang/include/flang/Support/OpenMP-utils.h
Stylie777: Done

flang/include/flang/Support/OpenMP-utils.h
Stylie777: Done, but I have used reserve rather than resize to avoid resizing the array and then adding to the end of that. I think reserve will work better for these cases.

jeanPerier:
Hi @Stylie777, this PR brought a link dependency of FortranEvaluate into an Optimizer library (see this [patch ](https://github.com/llvm/llvm-project/pull/198575))

I think that is a "fake" dependency (i.e., a header now indirectly included that injects some ctor/dtor code without actual data structure usages), but it would be best to try to break it because FortranEvaluate should not be required to link to Optimizer library to ensure that we are not creating situation where the compilation flow requires front-end data structures in Optimizers since we are supposed to be able to stop and restart the compilation from pure MLIR source without any side data structures.

Can you see if the dependency can be avoided?

Stylie777:
Hi @jeanPerier thanks for letting me know. This is not something I encountered when testing this but I will investigate and see if there is a solution.

jeanPerier:
> Hi @jeanPerier thanks for letting me know. This is not something I encountered when testing this but I will investigate and see if there is a solution.

You need a build with the cmake option BUILD_SHARED_LIBS=On option to hit the issue.

Stylie777:
https://github.com/llvm/llvm-project/pull/198742 will address the reliance on FortranEvaluate.

PR #197443, opened 2026-05-13
[lldb][NFC] Use DenseMap for ISAToDescriptorMap in ObjCLanguageRuntime

felipepiovezan:
This datastruct maps addr_t's to shared_pointers, which is a good case for a DenseMap. The tree allocation functions were showing up in Instrument traces when evaluating frame variable commands.

In a benchmark evaluating `v self` in a SwiftUI object, this brought down the number of CPU cycles from 578M to 487M. (release, no assertions)

### DIFF
--- a/lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.h
+++ b/lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.h
@@ -340,7 +340,7 @@ protected:
 
   bool AddClass(ObjCISA isa, const ClassDescriptorSP &descriptor_sp) {
     if (isa != 0) {
-      m_isa_to_descriptor[isa] = descriptor_sp;
+      m_isa_to_descriptor.insert_or_assign(isa, descriptor_sp);
       return true;
     }
     return false;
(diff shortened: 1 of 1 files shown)


lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.h
```diff
@@ -340,7 +340,7 @@ class ObjCLanguageRuntime : public LanguageRuntime {
 
   bool AddClass(ObjCISA isa, const ClassDescriptorSP &descriptor_sp) {
     if (isa != 0) {
-      m_isa_to_descriptor[isa] = descriptor_sp;
+      m_isa_to_descriptor.insert_or_assign(isa, descriptor_sp);
```
JDevlieghere, line 343: How is `insert_or_assign` different from `operator []`?

JDevlieghere (approved):
LGTM assuming there's a good reason for `insert_or_assign`. And if so, should we consider adopting that everywhere?

lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.h
felipepiovezan: Consider any map-like container `MyMap<Key, SomeBigType>`.
Operator `[]` will first allocate one `SomeBigType`, and then throw it away because of the assignment.
`insert_or_assign` makes no extra allocation.

felipepiovezan: (the answer is yes)

kastiglione: my understanding is that for `shared_ptr` there is no significant difference between the two. If the key exists, they both result in a copy assignment of `descriptor_sp`.

If they key does not exist:
1. With `insert_or_assign`, the value is copy constructed from `descriptor_sp`
2. With `operator[]` the value is default constructed, and then is copy assigned from `descriptor_sp`

PR #197444, opened 2026-05-13
[lldb][NFC] Use DenseMap in ObjCLanguageRuntime

felipepiovezan:
This is a multimap of Hash -> ObjCISA (lldb::addr_t), and it shows up cpu traces of swift applications. This commit replaces `std::multimap<hash, addr_t>` with `DenseMap<hash, SmallVector<addr_t, 2>>`.

This data structure was chosen because of the following experiment. When evaluating a frame variable command for a SwiftUI variable, this object gets populated with ~186,000 entries:

* over a thousand of them had 2 hash collisions.
* 300 of them had 3 hash collisions.
* 21 of them had 4 hash collisions.
* 1 of them had 5 hash collisions.

On a release build (no asserts), this patch brought down the CPU cycles measured for that command from 487M cycles to 389M cycles.

### DIFF
--- a/lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.cpp
+++ b/lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.cpp
@@ -194,16 +194,14 @@ ObjCLanguageRuntime::GetDescriptorIterator(ConstString name) {
   // Name hashes were provided, so use them to efficiently lookup name to
   // isa/descriptor
   const uint32_t name_hash = llvm::djbHash(name.GetStringRef());
-  std::pair<HashToISAIterator, HashToISAIterator> range =
-      m_hash_to_isa_map.equal_range(name_hash);
-  for (HashToISAIterator range_pos = range.first; range_pos != range.second;
-       ++range_pos) {
-    ISAToDescriptorIterator pos = m_isa_to_descriptor.find(range_pos->second);
-    if (pos != m_isa_to_descriptor.end()) {
-      if (pos->second->GetClassName() == name)
-        return pos;
-    }
-  }
+  auto matches_it = m_hash_to_isa_map.find(name_hash);
+  if (matches_it == m_hash_to_isa_map.end())
+    return m_isa_to_descriptor.end();
+
+  for (auto isa : matches_it->second)
+    if (auto pos = m_isa_to_descriptor.find(isa);
+        pos != m_isa_to_descriptor.end() && pos->second->GetClassName() == name)
+      return pos;
 
   return m_isa_to_descriptor.end();
 }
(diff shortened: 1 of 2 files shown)


lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.cpp
```diff
@@ -194,16 +194,14 @@ ObjCLanguageRuntime::GetDescriptorIterator(ConstString name) {
...
-        return pos;
-    }
-  }
+  auto matches_it = m_hash_to_isa_map.find(name_hash);
+  if (matches_it == m_hash_to_isa_map.end())
+    return m_isa_to_descriptor.end();
+
+  for (auto isa : matches_it->second)
+    if (auto pos = m_isa_to_descriptor.find(isa);
+        pos != m_isa_to_descriptor.end() && pos->second->GetClassName() == name)
```
JDevlieghere, line 203: Could we use `llvm::find_if` here?

JDevlieghere (commented):
LGTM

lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.cpp
felipepiovezan: we need an iterator into `m_isa_to_descriptor`, but the `find_if` would search over the array of collisions.

felipepiovezan:
rebased

lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.h
```diff
@@ -420,9 +420,8 @@ class ObjCLanguageRuntime : public LanguageRuntime {
   typedef std::map<ClassAndSel, lldb::addr_t> MsgImplMap;
   typedef std::map<ClassAndSelStr, lldb::addr_t> MsgImplStrMap;
   typedef llvm::DenseMap<ObjCISA, ClassDescriptorSP> ISAToDescriptorMap;
-  typedef std::multimap<uint32_t, ObjCISA> HashToISAMap;
+  typedef llvm::DenseMap<uint32_t, llvm::SmallVector<ObjCISA, 3>> HashToISAMap;
```
adrian-prantl: How did you arrive at the "3" constant?

felipepiovezan: It's described in the PR/commit description 😅

kastiglione: Is it wasteful to allocate 3 * ~186,000 if only ~1300 of them (less than 1%) need 2 or 3 values?

felipepiovezan: Fair point, I reran the experiments with SmallSize of 2 instead of 3, and they had virtually identical perf characteristics. So might as well use 2.

JDevlieghere (approved):

PR #197445, opened 2026-05-13
[CIR][NFC] Move createGlobalOp to be a member function

erichkeane:
For some reason this was a static function with the first argument being a CIRGenModule object anyway.  This is a NFC that I tried to do as a drive-by earlier but got called out on, so here it is :)


xlauko (approved):
lgtm

PR #196562, opened 2026-05-08
[LV] Implement usesFirstLaneOnly for VPWidenCallRecipe

huntergr-arm:
In order to avoid potentially introducing broadcasts for uniform or
linear arguments for widened calls, implement usesFirstLaneOnly for
VPWidenCallRecipe. Will hopefully make the cost model slightly more
accurate.

Removed instcombine pass from the uniform args tests, since removing
the broadcasts was the main reason for it being called.

Should help out #195444

PR #197448, opened 2026-05-13
[LV][NFC] Remove instcombine from RUN lines in AArch64 tests

david-arm:
This PR continues other work I've been doing trying to remove unnecessary extra passes from the RUN lines in order to make it easier to map the expected vectoriser output to the CHECK lines. As a result it has exposed some potential optimisations that we may be able to perform in VPlan.

Here is a summary of the changes I've noticed:

1. instcombine likes to canonicalise GEPs into certain forms. I'm not sure if there is value in VPlan trying to guess what the canonical form should be.
2. In tests like sve-cond-inv-loads.ll, etc. the pattern sub(urem) is often replaced with and(sub). This is potentially something the vectoriser could improve although I don't know if it would change the cost model.
3. There is poor codegen in gather_nxv4i32_ind64_stride2 in the file sve-gather-scatter.ll, which is due to
VPScalarIVStepsRecipe::execute. I have a PR that attempts to clean up this IR: #197169.
4. Simple missing fold in sve-inductions.ll for icmp(and(x,1), 0) -> trunc(x) to i1
5. Missing nsw flag - see sve-interleaved-accesses.ll. I think this might be due to the range of vscale.
6. Missing fold in sve-interleaved-masked-accesses.ll for select(icmp(slt, x, y), y, x) -> smax
7. Missing folds for reverse transformations of uniform operations, e.g. see sve-vector-reverse.ll for things like reverse(fadd(reverse(x)))
8. Removal of xor when used by the exit condition - see sve-vfabi.ll. There isn't much we can do about this because VPlan requires successors to be in a certain order, therefore the non-zero cost xor instruction has to be present.
9. See PR #196562 for fixes to some poor code in uniform-args-call-variants.ll
10. instcombine tends to narrow reduction PHI nodes generated by VPlan when there are extends involved. See reduction_i8 in reduction-small-size.ll. In this case perhaps the original scalar loop is simply not in a canonical form to start with?


paulwalker-arm (approved):
Looks like a fairly mechanical change with none of the tests showing an explosion of output.

llvm/test/Transforms/LoopVectorize/AArch64/reduction-small-size.ll
```diff
@@ -13,11 +13,11 @@ target triple = "aarch64--linux-gnu"
 ; }
 ;
 ; CHECK: vector.body:
-; CHECK:   phi <16 x i8>
+; CHECK:   phi <16 x i32>
```
fhahn: I think we should truncate the phi in the loop and extend the backedge value for the loops in the file

Could you auto-generate the full checks here, or check that we also have the expected truncs/extends to not lose that part of the coverage?

llvm/test/Transforms/LoopVectorize/AArch64/sve-vfabi.ll
```diff
@@ -58,22 +60,24 @@ define void @test_little_big_params(ptr readonly %a, ptr readonly %b, ptr noalia
 ; CHECK-NEXT:    br label [[VECTOR_PH:%.*]]
 ; CHECK:       vector.ph:
 ; CHECK-NEXT:    [[TMP0:%.*]] = call i64 @llvm.vscale.i64()
-; CHECK-NEXT:    [[TMP1:%.*]] = shl nuw nsw i64 [[TMP0]], 1
+; CHECK-NEXT:    [[TMP1:%.*]] = shl nuw i64 [[TMP0]], 1
+; CHECK-NEXT:    [[ACTIVE_LANE_MASK_ENTRY:%.*]] = call <vscale x 2 x i1> @llvm.get.active.lane.mask.nxv2i1.i64(i64 0, i64 1025)
```
fhahn, line 64: I think that was mentioned at another PR as well, would be good to fold trivial active-lane-masks, as this is something LV generates

david-arm: I looked at this previously and the fold depended upon having an actual IR instruction to look up the vscale_range function attribute. At the point VPlan does the lookup it doesn't have an instruction.

llvm/test/Transforms/LoopVectorize/AArch64/reduction-small-size.ll
david-arm: Yep. Done in https://github.com/llvm/llvm-project/pull/197632

david-arm: With the full CHECK lines it has exposed some missing folds like this in vplan:

```
; CHECK-NEXT:    [[TMP4:%.*]] = trunc <16 x i32> [[TMP23]] to <16 x i8>
; CHECK-NEXT:    [[TMP9]] = zext <16 x i8> [[TMP4]] to <16 x i32>
```

which can just be `and <16 x i32> [[TMP23]], splat (i32 255)`. However, the more powerful solution is to transform the phi to `<16 x i8>` like instcombine does.

PR #197450, opened 2026-05-13
[TableGen] Simplify Record type checks. NFC.

jayfoad:


llvm/lib/TableGen/Record.cpp
```diff
@@ -2329,7 +2329,7 @@ const RecTy *TypedInit::getFieldType(const StringInit *FieldName) const {
 }
 
 const Init *TypedInit::convertInitializerTo(const RecTy *Ty) const {
-  if (getType() == Ty || getType()->typeIsA(Ty))
```
jayfoad, line 2332: I don't see the need for the `== Ty` check since it looks like `typeIsA` will handle this efficiently.

wangpc-pp (approved):
LGTM.

PR #197452, opened 2026-05-13
[AMDGPU] Add DAG mutation to sink consumers of VMEM-pending registers

barbara-amd:
When an instruction consumes a VMEM-pending register, the post-RA scheduler tends to place it immediately after the VMEM load that defines it, before any subsequent VMEM loads. This forces SIInsertWaitcnts to emit a full s_waitcnt vmcnt(0) to satisfy the dependency, preventing independent in-flight loads from completing in parallel.

Introduce AMDGPUVMemUseOrdering, a ScheduleDAGMutation that adds artificial order edges from in-region VMEM loads to consumers of VMEM-pending registers. A register is VMEM-pending if its last writer on some predecessor path is an unwaited pure VMEM load; two sources are covered: cross-block live-ins (backward CFG walk) and same-block pre-region defs (earlier scheduling regions in the same block).

This sinks consumers past the independent in-region loads so that SIInsertWaitcnts can emit a single weaker s_waitcnt vmcnt(N), allowing multiple in-flight loads to complete in parallel.


jayfoad:
> When an instruction consumes a VMEM-pending register, the post-RA scheduler tends to place it immediately after the VMEM load that defines it

Why would it do that? Surely it knows that VMEM loads have high latency?

> Introduce AMDGPUVMemUseOrdering, a ScheduleDAGMutation that adds artificial order edges from in-region VMEM loads to consumers of VMEM-pending registers. A register is VMEM-pending if its last writer on some predecessor path is an unwaited pure VMEM load; two sources are covered: cross-block live-ins (backward CFG walk) and same-block pre-region defs (earlier scheduling regions in the same block).

Can you give some examples of these two cases? What does the input MIR look like, and where do you add the extra order edges?

barbara-amd:
> > When an instruction consumes a VMEM-pending register, the post-RA scheduler tends to place it immediately after the VMEM load that defines it
> 
> Why would it do that? Surely it knows that VMEM loads have high latency?

The scheduler models VMEM latency only within a scheduling region. Any instruction that is a scheduling boundary per SIInstrInfo::isSchedulingBoundary - for example one that modifies EXEC or is a terminator - splits the defining load and its consumer into different regions. In the consumer's region the result register is a bare live-in - no DAG node for the defining load, no latency edge, no reason to delay it. Only SIInsertWaitcnts, tracking scoreboard state linearly across region boundaries, sees the outstanding VMEM and emits vmcnt(0).

> > Introduce AMDGPUVMemUseOrdering, a ScheduleDAGMutation that adds artificial order edges from in-region VMEM loads to consumers of VMEM-pending registers. A register is VMEM-pending if its last writer on some predecessor path is an unwaited pure VMEM load; two sources are covered: cross-block live-ins (backward CFG walk) and same-block pre-region defs (earlier scheduling regions in the same block).
> 
> Can you give some examples of these two cases? What does the input MIR look like, and where do you add the extra order edges?

Case 1 - cross-block live-in: When a register is live-in to a scheduling region and its last writer on a predecessor path is an unwaited VMEM load, the mutation adds an Ord Artificial edge from each in-region VMEM load to the consumer SUnit. The consumer is then forced past the in-region loads by the scheduler.
Case 2 - same-block pre-region def: The same situation arises within a single basic block when a scheduling boundary splits it into multiple regions. A VMEM load in an earlier region is invisible to the later region's DAG. The mutation scans instructions from the block start to the current region start, identifies any registers whose last writer was an unwaited VMEM load, and adds the same Ord Artificial edges.
In both cases the added edges have zero latency and are purely ordering constraints - they sink consumers past independent in-region loads without affecting the critical path of instructions that have real data dependencies.

arsenm:
> for example one that modifies EXEC or is a terminator

What cases do we have left with exec modifying instructions that aren't terminators (besides the control flow lowering prolog instructions)

barbara-amd:
> > for example one that modifies EXEC or is a terminator
> 
> What cases do we have left with exec modifying instructions that aren't terminators (besides the control flow lowering prolog instructions)

1. v_cmpx_* — VALU compare-and-write-EXEC (VOPCInstructions.td:435,448).
2. Other SAVEEXEC variants (S_OR_SAVEEXEC_*, S_XOR_SAVEEXEC_*, S_ANDN2_SAVEEXEC_* etc.) — used in SI_ELSE lowering and WWM entry (SOPInstructions.td:342–412).
3. S_ANDN1/2_WREXEC_* (GFX9+) — write EXEC without saving (SOPInstructions.td:390–412).
4. WWM/WQM transitions — ENTER_STRICT_WWM/WQM and EXIT_STRICT_* lower to S_OR_SAVEEXEC_*, S_WQM_*, S_MOV_* → EXEC respectively, all non-terminators (SIInstrInfo.cpp:2556–2577).
5. SI_KILL_*_PSEUDO — Defs = [EXEC, SCC], no isTerminator (SIInstructions.td:641).
6. CF epilog — SI_END_CF lowers to S_OR_B32/B64 (non-terminator) when no block split is needed (SILowerControlFlow.cpp:497–520).

_(AI-assisted research — should be taken with caution.)_

arsenm:
> 1. v_cmpx_* — VALU compare-and-write-EXEC (VOPCInstructions.td:435,448).
> 2. Other SAVEEXEC variants (S_OR_SAVEEXEC__, S_XOR_SAVEEXEC__, S_ANDN2_SAVEEXEC_* etc.) — used in SI_ELSE lowering and WWM entry (SOPInstructions.td:342–412).
> 3. S_ANDN1/2_WREXEC_* (GFX9+) — write EXEC without saving (SOPInstructions.td:390–412).
> 4. WWM/WQM transitions — ENTER_STRICT_WWM/WQM and EXIT_STRICT_* lower to S_OR_SAVEEXEC__, S_WQM__, S_MOV_* → EXEC respectively, all non-terminators (SIInstrInfo.cpp:2556–2577).
> 5. SI_KILL_*_PSEUDO — Defs = [EXEC, SCC], no isTerminator (SIInstructions.td:641).
> 6. CF epilog — SI_END_CF lowers to S_OR_B32/B64 (non-terminator) when no block split is needed (SILowerControlFlow.cpp:497–520).
> 
> _(AI-assisted research — should be taken with caution.)_

Basically we shouldn't have any. The save execs are special cases for the prolog, of these v_cmpx and wwm/wqm are most concerning

llvm/lib/Target/AMDGPU/AMDGPUVMemUseOrdering.cpp
```diff
@@ -0,0 +1,457 @@
+//===--- AMDGPUVMemUseOrdering.cpp - AMDGPU VMEM Use Ordering
+//--------------===//
```
arsenm: Bad line break

llvm/lib/Target/AMDGPU/AMDGPUVMemUseOrdering.cpp
```diff
@@ -0,0 +1,457 @@
...
+      gatherVMemPendingUnits(MBB, Candidates, *TRI, ST, PendingUnits, Visited,
+                             0);
+    }
+  }
+
+  // Source B: same-block pre-region VMEM-pending registers.
+  // Map from register unit to the counter type of its last VMEM writer.
+  // A unit absent from the map is not VMEM-pending (last writer was non-VMEM,
+  // or the unit was cleared by a wait).
+  if (DAG->begin() != MBB->begin()) {
```
arsenm: This probably doesn't handle debug instructions correctly

perlfu:
> Basically we shouldn't have any. The save execs are special cases for the prolog, of these v_cmpx and wwm/wqm are most concerning

I would expect `v_cmpx` to just be occurring near block termination.
These are added by a peephole which I think runs after this scheduling anyway.

WWM/WQM is a special case -- I am not sure you really need to consider it in general.

perlfu:
My high level on a quick first pass is this replicates a lot of logic from `SIInsertWaitcnts`.
I wonder if its possible to have more code reuse between them?

arsenm:
> > Basically we shouldn't have any. The save execs are special cases for the prolog, of these v_cmpx and wwm/wqm are most concerning
> 
> I would expect `v_cmpx` to just be occurring near block termination. These are added by a peephole which I think runs after this scheduling anyway.

They should be marked as isTerminator to enforce this

llvm/lib/Target/AMDGPU/AMDGPUTargetMachine.cpp
```diff
@@ -1332,6 +1340,12 @@ GCNTargetMachine::createPostMachineScheduler(MachineSchedContext *C) const {
   DAG->addMutation(createAMDGPUExportClusteringDAGMutation());
   DAG->addMutation(createAMDGPUBarrierLatencyDAGMutation(C->MF));
   DAG->addMutation(createAMDGPUHazardLatencyDAGMutation(C->MF));
+  // Post-RA only: uses physreg live-ins and physreg defs; not applicable to
+  // pre-RA iterative schedulers.
+  if ((EnableVMemUseOrdering.getNumOccurrences() ||
+       getOptLevel() >= CodeGenOptLevel::Less) &&
+      EnableVMemUseOrdering)
+    DAG->addMutation(createAMDGPUVMemUseOrderingDAGMutation(C->MF));
   return DAG;
```
Sisyph, line 1473: Two comments related to DAG mutations in general. We need to be mindful of the compile time of the mutations, and we need to be mindful of how the different mutations will interact. The edges added by mutations that run earlier could affect the mutations that run after (such as dropping edges due to cycles). Do you have any data related to either of those points?

barbara-amd:
PR updated with refinements addressing the possible code-sharing and debug instruction comments. Remaining feedback to follow.

barbara-amd:
This PR adds a post-RA scheduling DAG mutation that attempts to identify register values whose nearest reaching definition is a VMEM load outside the current region. When such a value is consumed inside the region, the mutation adds ordering edges from in-region VMEM loads to that consumer.
The intent is to keep the consumer below the region's VMEM loads so that a single later wait can cover both the previously outstanding VMEM operation and the loads issued within the region.

Is a scheduling-DAG mutation an acceptable way to model these inter-block/cross-region VMEM dependencies, or is there an alternate preference? From what I've measured so far, the compile-time impact of the mutation is
~0.1%, and it showed some FPS gains across several titles.

PR #197453, opened 2026-05-13
[DirectX] Do not emit DISubprograms as distinct

hvdijk:
LLVM 3.7 never did this.


llvm/lib/Target/DirectX/DirectXIRPasses/DXILDebugInfo.cpp
```diff
@@ -50,6 +50,27 @@ DXILDebugInfoMap DXILDebugInfoPass::run(Module &M) {
...
+
+    if (SP->isDistinct()) {
+      NewSP = DISubprogram::get(
+          M.getContext(), SP->getScope(), SP->getName(), SP->getLinkageName(),
+          SP->getFile(), SP->getLine(), SP->getType(), SP->getScopeLine(),
+          SP->getContainingType(), SP->getVirtualIndex(),
+          SP->getThisAdjustment(), SP->getFlags(), SP->getSPFlags(),
+          SP->getUnit(), SP->getTemplateParams(), SP->getDeclaration(),
+          SP->getRetainedNodes(), SP->getThrownTypes(), SP->getAnnotations(),
+          SP->getTargetFuncName(), SP->getKeyInstructionsEnabled());
```
bogner, line 63: It makes me a little nervous that the most likely way for this to break would be a new value being added to the end and explicitly defaulted, which would mean we don't copy the node correctly anymore and might not notice. Is there a way to / is it reasonable to copy the node using `operands()` here instead, or would that obscure what's happening here in a way that outweighs the risk of a missed update in the future?

hvdijk: We can't do that, there is no overload of `DISubprogram::get()` that allows us to specify an array of operands, but that is okay: the goal here isn't to replicate the `DISubprogram` exactly, the goal here is to create a `DISubprogram` that holds the information for LLVM 3.7 (which is why other PRs modify this further). If a future version of LLVM adds more operands to `DISubprogram`, those will be operands that do not exist in LLVM 3.7 and do not need to be preserved.

bogner (approved):

llvm/lib/Target/DirectX/DirectXIRPasses/DXILDebugInfo.cpp
bogner: Fair enough. Depending what they are dropping them silently may or may not be the right thing to do, but we can cross that bridge when we get to it.

PR #197455, opened 2026-05-13
[DAG] isIdentityElement - use KnownBits instead of constant splat to match identity constants

RKSimon:
This works better with the DemandedElts mask to match hidden identity constants (zero in particular.....).

I need this for the ongoing work to improve VECREDUCE simplification to match identity elements (legalisation pads with identity elements) in an expanded reduction shuffle chain.

The Thumb2 CLMUL expanded codegen is already a monster - I've tried a few things to try and simplify it but we're emitting more FREEZE nodes due to a BINOP(VSELECT()) -> VSELECT(BINOP(FREEZE(),FREEZE())) fold that now fires, and gets stuck trying to freeze a number of ARMISD nodes. TBH, I'd much prefer not to try to shave this yak until we've improved FREEZE(OP()) -> OP(FREEZE(),FREEZE(),....) folding.

### DIFF
--- a/llvm/lib/CodeGen/SelectionDAG/SelectionDAG.cpp
+++ b/llvm/lib/CodeGen/SelectionDAG/SelectionDAG.cpp
@@ -13594,32 +13594,34 @@ bool SelectionDAG::isIdentityElement(unsigned Opcode, SDNodeFlags Flags,
                                      unsigned OperandNo, unsigned Depth) const {
   // NOTE: The cases should match with IR's ConstantExpr::getBinOpIdentity().
   // TODO: Target-specific opcodes could be added.
-  if (auto *ConstV = isConstOrConstSplat(V, DemandedElts, /*AllowUndefs*/ false,
-                                         /*AllowTruncation*/ true)) {
-    APInt Const = ConstV->getAPIntValue().trunc(V.getScalarValueSizeInBits());
-    switch (Opcode) {
-    case ISD::ADD:
-    case ISD::OR:
-    case ISD::XOR:
-    case ISD::UMAX:
-      return Const.isZero();
-    case ISD::MUL:
-      return Const.isOne();
-    case ISD::AND:
-    case ISD::UMIN:
-      return Const.isAllOnes();
-    case ISD::SMAX:
-      return Const.isMinSignedValue();
-    case ISD::SMIN:
-      return Const.isMaxSignedValue();
-    case ISD::SUB:
-    case ISD::SHL:
-    case ISD::SRA:
-    case ISD::SRL:
-      return OperandNo == 1 && Const.isZero();
-    case ISD::UDIV:
-    case ISD::SDIV:
-      return OperandNo == 1 && Const.isOne();
+  if (V.getValueType().isInteger()) {
+    KnownBits Known = computeKnownBits(V, DemandedElts, Depth);
+    if (Known.isConstant()) {
+      const APInt &Const = Known.getConstant();
+      switch (Opcode) {
+      case ISD::ADD:
+      case ISD::OR:
+      case ISD::XOR:
+      case ISD::UMAX:
+        return Const.isZero();
+      case ISD::MUL:
+        return Const.isOne();
+      case ISD::AND:
+      case ISD::UMIN:
+        return Const.isAllOnes();
+      case ISD::SMAX:
+        return Const.isMinSignedValue();
+      case ISD::SMIN:
+        return Const.isMaxSignedValue();
+      case ISD::SUB:
+      case ISD::SHL:
+      case ISD::SRA:
+      case ISD::SRL:
+        return OperandNo == 1 && Const.isZero();
+      case ISD::UDIV:
+      case ISD::SDIV:
+        return OperandNo == 1 && Const.isOne();
+      }
     }
   } else if (auto *ConstFP = isConstOrConstSplatFP(V, DemandedElts)) {
     switch (Opcode) {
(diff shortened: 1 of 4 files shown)


llvm/lib/CodeGen/SelectionDAG/SelectionDAG.cpp
```diff
@@ -13576,32 +13576,34 @@ bool SelectionDAG::isIdentityElement(unsigned Opcode, SDNodeFlags Flags,
...
-    case ISD::SRA:
-    case ISD::SRL:
-      return OperandNo == 1 && Const.isZero();
-    case ISD::UDIV:
-    case ISD::SDIV:
-      return OperandNo == 1 && Const.isOne();
+  if (V.getValueType().isInteger()) {
+    KnownBits Known = computeKnownBits(V, DemandedElts, Depth);
+    if (Known.isConstant()) {
+      const APInt &Const = Known.getConstant();
```
jayfoad, line 13600: I guess this logic can't go inside `isConstOrConstSplat` because it needs to return ConstantSDNode not APInt. But is it worth abstracting it into some new `isConstXXX` helper function?

RKSimon: Maybe if its codesize you're worried about we could do this:
```
    if (KnownBits Known = computeKnownBits(V, DemandedElts, Depth); Known.isConstant()) {
      const APInt &Const = Known.getConstant();
```
But I'm not convinced tbh - is there any features you think we're missing or is this code prettifying?

We could also move this into SDPatternMatch - but again, I'm not sure.

jayfoad: I'm mostly concerned about how code that just wants to check for constants will evolve.

If all callers of isConstOrConstSplat (or at least all callers that use a non-trivial demanded elts mask) are gradually going to migrate to making explicit KnownBits calls instead, then it seems like the helper function is not really helping any more.

But I'm not particularly invested in DAG code, this was just a drive-by observation.

RKSimon:
@davemgreen The MVE clmul codegen appears to be particularly bad due to it relying on the default bittest expansion:
```
      SDValue Part;
      if (!hasBitTest(Y, ShiftAmt) &&
          isOperationLegalOrCustom(
              ISD::MUL, getTypeToTransformTo(*DAG.getContext(), VT))) {
        Part = DAG.getNode(ISD::MUL, DL, VT, X, YMasked);
      } else {
        // Canonical bit test: (Y & (1 << I)) != 0
        SDValue Zero = DAG.getConstant(0, DL, VT);
        SDValue Cond = DAG.getSetCC(DL, SetCCVT, YMasked, Zero, ISD::SETEQ);
        SDValue XShifted = DAG.getNode(ISD::SHL, DL, VT, X, ShiftAmt);
        Part = DAG.getSelect(DL, VT, Cond, Zero, XShifted);
      }
```
Are there any MVE specific tricks for v2i64 MUL we can try? The YMasked runtime variable is either 0 or a pow2_imm value

davemgreen:
> @davemgreen The MVE clmul codegen appears to be particularly bad due to it relying on the default bittest expansion:
> 
> ```
>       SDValue Part;
>       if (!hasBitTest(Y, ShiftAmt) &&
>           isOperationLegalOrCustom(
>               ISD::MUL, getTypeToTransformTo(*DAG.getContext(), VT))) {
>         Part = DAG.getNode(ISD::MUL, DL, VT, X, YMasked);
>       } else {
>         // Canonical bit test: (Y & (1 << I)) != 0
>         SDValue Zero = DAG.getConstant(0, DL, VT);
>         SDValue Cond = DAG.getSetCC(DL, SetCCVT, YMasked, Zero, ISD::SETEQ);
>         SDValue XShifted = DAG.getNode(ISD::SHL, DL, VT, X, ShiftAmt);
>         Part = DAG.getSelect(DL, VT, Cond, Zero, XShifted);
>       }
> ```
> 
> Are there any MVE specific tricks for v2i64 MUL we can try? The YMasked runtime variable is either 0 or a pos2_imm value

I only got as far as adding the tests, not optimizing them yet. There is a vector vmull polynomial, but only 8x8->16 and 16x16->32 cases. The v2xY cases will be more difficult, we don't have many operations and no scalar types for 64bit. We do have xor / and, and a select (they just work on v16i8), but not shift or icmp or mul, and building the predicates can be very painful. We try to avoid 2x vectorization where possible.

It is only the clmul_v4i64_zext case that is getting meaningfully bigger. There are some odd duplicated vselects with identical conditions. I put together #197799 but it didn't help, I wasn't sure where they were coming from yet.

RKSimon:
Do you think the mve clmul codegen should be a blocker for this patch? No pressure - honest!

davemgreen:
Can you rebase over #197832? It feels like something is going wrong with multiple vselects being created where they were not profitable, as if a fold is missing a one-use check somewhere. I didn't look too much deeper but that should help clear them up.

RKSimon:
I'm seeing a lot of IR like this (generated by foldSelectWithIdentityConstant):
```
      t2648: v2i64 = xor t2647, t1831
    t2649: v2i64 = vselect t3806, t2647, t2648
  t2644: v2i64 = freeze t2649
...
      t2645: v2i64 = xor t2644, t1833
    t2646: v2i64 = vselect t3820, t2644, t2645
  t2641: v2i64 = freeze t2646
```
currently testing to see what happens if I let freeze push through multiple operands of select/vselect nodes.

davemgreen:
Can you regen the test? I think they are fine now.

RKSimon:
Thanks! I tried the freeze(vselect) fold, which seems to have helped, but your patch seems to avoid the issue entirely

RKSimon:
I think we're there now - OK to commit?

davemgreen (approved):

PR #197456, opened 2026-05-13
[Offload] Make 'llvm-offload-binary' use multi-binaries

jhuber6:
Summary:
There's two ways you can put multiple binaries in the section. Either
use the version two multi-binary support or just concatenate them. This
PR changes the llvm-offload-binary tool to use the multi-support rather
than directly concatenating them.

The motivation for this is to save space and make it easier to support
compression in the future. Compression would be a flag in the header and
the compression is only really valuable if it can combine the
architecture variants. ELF section compression is a little spotty but
would be another good solution.


sarnex (approved):

YuriPlyakhin (approved):
LGTM

adurang (approved):

PR #197457, opened 2026-05-13
[DirectX] Drop unsupported DISubprogram flags

hvdijk:
These flags did not exist in LLVM 3.7 so should be omitted.


hvdijk:
The `isDistinct() ? getDistinct() : get()` will not be necessary if #197453 goes in first. I did this in order to be able to put this up for review without waiting for that.

bogner (approved):
LGTM. Please do merge #197453 first and update accordingly.

Issue #63628, opened 2023-06-30
need a more restricted form of `unnamed_addr` for string literals


zygoloid:
The C++ committee is in the process of clarifying the standard wording around string literals to make it clearer that a string literal can't be merged with a named constant (though that seems to have always been the intent). Clang currently uses `unnamed_addr` to power string literal merging, but that will no longer be a conforming way to lower C++ code. Clang therefore needs a way to express "may be merged with other constants" in LLVM IR that only permits merging if *both* constants have the annotation.

efriedma-quic:
How many different kinds of unmergable items are there?  For example, can we merge a narrow string literal with a wide string literal?  Other kinds of unnamed constants?  (Not sure if there are any?  C has global compound literals, but I guess C++ doesn't have any direct equivalent.)

I ask because if there are multiple different categories of object, we might need to add some sort of tag to constants to indicate they can only be merged with other constants with the same tag.

zygoloid:
So far, the intent is for there to be only one, and to also allow merging the backing array of an `initializer_list` with that of a string literal. But it might make sense to add a tag anyway to future-proof the mechanism and to disallow cross-language merging.

PR #197458, opened 2026-05-13
[C++26] Implement P2752R3 'Static storage for braced initializers'

### LINKED ISSUE #104487: Implement P2752R3 Static storage for braced initializers
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2752r3.html

yronglin:
This PR implement [P2752R3 'Static storage for braced initializers'](https://wg21.link/P2752R3).

Assisted by Codex.

Fixes https://github.com/llvm/llvm-project/issues/104487


clang/lib/CodeGen/CGExpr.cpp
```diff
@@ -512,6 +511,50 @@ static RawAddress createReferenceTemporary(CodeGenFunction &CGF,
...
+
+  SmallVector<const Expr *, 2> CommaLHSs;
+  SmallVector<SubobjectAdjustment, 2> Adjustments;
+  const Expr *E =
+      M->getSubExpr()->skipRValueSubobjectAdjustments(CommaLHSs, Adjustments);
+  if (!Adjustments.empty())
+    return std::nullopt;
+
+  QualType ArrayType = E->getType();
+  if (!getContext().getAsConstantArrayType(ArrayType))
```
efriedma-quic: Why does it matter if the type of the expression is an array?  I think you need something more specific to detect if an object is specifically a "backing array", as opposed to some other array.

clang/lib/CodeGen/CGExpr.cpp
```diff
@@ -512,6 +511,50 @@ static RawAddress createReferenceTemporary(CodeGenFunction &CGF,
   llvm_unreachable("unknown storage duration");
 }
 
+std::optional<RawAddress> CodeGenFunction::tryEmitStaticInitListBackingArray(
+    const MaterializeTemporaryExpr *M) {
+  switch (M->getStorageDuration()) {
+  case SD_FullExpression:
+  case SD_Automatic:
```
efriedma-quic, line 526: This specifically only handles cases where we have a temporary with a limited lifetime?  I guess that simplifies some of the reasoning here. I was concerned about interactions with constexpr, but I guess anything that happens during constexpr evaluation isn't relevant because any temporary backing arrays constucted during constexpr evaluation can't leak out of the evaluation.

clang/lib/CodeGen/CodeGenModule.cpp
```diff
@@ -7635,6 +7570,27 @@ ConstantAddress CodeGenModule::GetAddrOfGlobalTemporary(
...
 
+ConstantAddress
+CodeGenModule::EmitStaticInitListBackingArray(llvm::Constant *Initializer,
+                                              CharUnits Alignment) {
+  LangAS AddrSpace = GetGlobalConstantAddressSpace();
+  auto TargetAS = getContext().getTargetAddressSpace(AddrSpace);
+  auto *GV = new llvm::GlobalVariable(
+      getModule(), Initializer->getType(), /*isConstant=*/true,
+      llvm::GlobalValue::PrivateLinkage, Initializer, ".init.list", nullptr,
+      llvm::GlobalValue::NotThreadLocal, TargetAS);
```
efriedma-quic: Should this be unnamed_addr? (But see also #63628.)

clang/www/cxx_status.html
```diff
@@ -135,7 +135,7 @@ <h2 id="cxx26">C++2c implementation status</h2>
  <tr>
   <td>Static storage for braced initializers</td>
   <td><a href="https://wg21.link/P2752R3">P2752R3</a> (<a href="#dr">DR</a>)</td>
-  <td class="none" align="center">No</td>
+  <td class="unreleased" align="center">Clang 23</td>
```
efriedma-quic, line 250: Technically we "conform" even without this change; nothing in the standard actually requires this transform.

efriedma-quic: Actually, technically, the paper has a consequence: it interacts with [expr.const], speciifcally the "equality operator comparing pointers to potentially non-unique objects".  We need to ensure that the following produces an error:

```
#include <initializer_list>
constexpr bool f(std::initializer_list<int> a, std::initializer_list<int> b) {
  return a.begin()!=b.begin();
}
static_assert(f({1},{1}));
```

shafik (commented):
I want to see tests that cover the ill-formdness in a constant expression context as discussed in the paper.

The paper conclusion that clang gets this wrong I believe is incorrect given the resolution to [CWG2765](http://wg21.link/cwg2765) and we squarely confirm that while unspecified at runtime this is ill-formed at compile time.

clang/www/cxx_status.html
yronglin: I have revert the changes in CodeGen, and added more test for constant expression.

yronglin:
> I want to see tests that cover the ill-formdness in a constant expression context as discussed in the paper.
> 
> The paper conclusion that clang gets this wrong I believe is incorrect given the resolution to [CWG2765](http://wg21.link/cwg2765) and we squarely confirm that while unspecified at runtime this is ill-formed at compile time.

Sure, added.

clang/lib/AST/ExprConstant.cpp
```diff
@@ -2067,31 +2074,32 @@ static bool GetLValueBaseAsString(const EvalInfo &Info, const LValue &LVal,
...
-                                                    const LValue &RHS) {
-  LValueBaseString LHSString, RHSString;
-  if (!GetLValueBaseAsString(Info, LHS, LHSString) ||
-      !GetLValueBaseAsString(Info, RHS, RHSString))
-    return false;
-
+//
+// Callers must ensure LHSString.CharWidth == RHSString.CharWidth — the byte
+// comparison below has no meaning across mismatched character widths.
+static bool ArePotentiallyOverlappingStrings(const LValueBaseString &LHSString,
```
cor3ntin:
```suggestion
static bool ArePotentiallyOverlappingStringLiterals(const LValueBaseString &LHSString,
```

I don't think the name change is warranted here - what would it mean otherwise?

clang/lib/AST/ExprConstant.cpp
```diff
@@ -2067,31 +2074,32 @@ static bool GetLValueBaseAsString(const EvalInfo &Info, const LValue &LVal,
...
-static bool ArePotentiallyOverlappingStringLiterals(const EvalInfo &Info,
-                                                    const LValue &LHS,
-                                                    const LValue &RHS) {
-  LValueBaseString LHSString, RHSString;
-  if (!GetLValueBaseAsString(Info, LHS, LHSString) ||
-      !GetLValueBaseAsString(Info, RHS, RHSString))
-    return false;
-
+//
+// Callers must ensure LHSString.CharWidth == RHSString.CharWidth — the byte
```
cor3ntin: Can you add an assert?

clang/lib/AST/ExprConstant.cpp
```diff
@@ -4860,6 +4924,404 @@ static CompleteObject findCompleteObject(EvalInfo &Info, const Expr *E,
...
+
+  // &&label values: the runtime address of a label isn't constant, so two
+  // diffs are only provably equal when they share both endpoints.
+  if (LHS.isAddrLabelDiff())
+    return compareAddrLabelDiff(LHS, RHS);
+
+  // Any APValue kind we don't explicitly model is conservatively unknown.
+  return APValueEquality::Unknown;
+}
+
```
cor3ntin: Maybe we should be conservative and make rare case like labels, matrixes never equal ?

cor3ntin: Earlier you said

> /// Callers MUST treat Unknown the same as Equal for soundness: when we
/// can't prove the two backing arrays differ, the implementation is free
/// to merge their storage.

Which means anything we don't support explicitly compares equal? That does not sound right.

clang/lib/AST/ExprConstant.cpp
```diff
@@ -4860,6 +4924,404 @@ static CompleteObject findCompleteObject(EvalInfo &Info, const Expr *E,
...
+APValueEquality compareRecords(const RecordDecl *RD, const APValue &LHS,
+                               const APValue &RHS, const ASTContext &Ctx) {
+  if (const RecordDecl *Definition = RD->getDefinition())
+    RD = Definition;
+
+  if (RD->isUnion()) {
+    if (!LHS.isUnion() || !RHS.isUnion())
+      return APValueEquality::Unknown;
```
cor3ntin: Can this happen?

cor3ntin (commented):
This will need
 - Changelog
 - Updating the paper status page
 - Updating the DRs status page


I'm really surprise we do not already have a way to compare APValues

clang/test/CXX/drs/cwg27xx.cpp
```diff
@@ -1,10 +1,10 @@
...
-// RUN: %clang_cc1 -triple x86_64-linux-gnu -std=c++17 -fexceptions -fcxx-exceptions -pedantic-errors %s -verify-directives -verify=expected,since-cxx11,since-cxx14
-// RUN: %clang_cc1 -triple x86_64-linux-gnu -std=c++20 -fexceptions -fcxx-exceptions -pedantic-errors %s -verify-directives -verify=expected,since-cxx11,since-cxx14,since-cxx20
-// RUN: %clang_cc1 -triple x86_64-linux-gnu -std=c++23 -fexceptions -fcxx-exceptions -pedantic-errors %s -verify-directives -verify=expected,since-cxx11,since-cxx14,since-cxx20,since-cxx23
-// RUN: %clang_cc1 -triple x86_64-linux-gnu -std=c++2c -fexceptions -fcxx-exceptions -pedantic-errors %s -verify-directives -verify=expected,since-cxx11,since-cxx14,since-cxx20,since-cxx23,since-cxx26
+// RUN: %clang_cc1 -triple x86_64-linux-gnu -std=c++98 -fexceptions -fcxx-exceptions -pedantic-errors -verify-directives -verify=expected,cxx98 %s
+// RUN: %clang_cc1 -triple x86_64-linux-gnu -std=c++11 -fexceptions -fcxx-exceptions -pedantic-errors -verify-directives -verify=expected,since-cxx11,beforecxx23 %s
```
Endilll:
```suggestion
// RUN: %clang_cc1 -triple x86_64-linux-gnu -std=c++11 -fexceptions -fcxx-exceptions -pedantic-errors -verify-directives -verify=expected,since-cxx11,cxx98-20 %s
```

clang/test/CXX/drs/cwg27xx.cpp
```diff
@@ -193,37 +196,108 @@ static_assert(!__is_layout_compatible(StructWithAnonUnion, StructWithAnonUnion3)
...
 // cxx98-error@-3 {{static assertion expression is not an integral constant expression}} FIXME
 static_assert((const char*)"foo" != "oo", "");
 // cxx98-error@-1 {{static assertion expression is not an integral constant expression}} FIXME
+#endif
 
 #if __cplusplus >= 201103L
 constexpr const char *f() { return "foo"; }
-constexpr bool b2 = f() == f(); 
+
+constexpr bool b1 = +"foo" == "foo";
```
Endilll: Why is this test case duplicated between C++98 and C++11 modes? Why not remove `#if __cplusplus == 199711L` above and let those static_asserts be evaluated in all language modes?

clang/test/CXX/drs/cwg27xx.cpp
```diff
@@ -193,37 +196,108 @@ static_assert(!__is_layout_compatible(StructWithAnonUnion, StructWithAnonUnion3)
...
+//   since-cxx11-note@-2 {{comparison of addresses of potentially overlapping literals has unspecified value}}
+static_assert("foo" != &"bar"[0], "");
+static_assert((const char *)"foo" != "oo", "");
+
+constexpr bool f(std::initializer_list<int> a, std::initializer_list<int> b) {
+  return a.begin() != b.begin(); // #cwg2765-init-list-compare
+}
+static_assert(f({1}, {1}), "");
+// since-cxx11-error@-1 {{static assertion expression is not an integral constant expression}}
+// since-cxx11-note@#cwg2765-init-list-compare {{comparison of addresses of potentially non-unique objects has unspecified value}}
```
Endilll:
```suggestion
//   since-cxx11-note@#cwg2765-init-list-compare {{comparison of addresses of potentially non-unique objects has unspecified value}}
```

clang/test/CXX/drs/cwg27xx.cpp
```diff
@@ -193,37 +196,108 @@ static_assert(!__is_layout_compatible(StructWithAnonUnion, StructWithAnonUnion3)
...
+// since-cxx11-note@-3 {{in call to 'f({&{1}[0], 1}, {&{1}[0], 1})'}}
+
+#if __cplusplus >= 201402L
+constexpr bool f9(const int *p) {
+  std::initializer_list<int> il = {1, 2, 3};
+  return p ? (p == il.begin()) : f9(il.begin()); // #cwg2765-f9-compare
+}
+constexpr bool b9 = f9(nullptr);
+// since-cxx14-error@-1 {{constexpr variable 'b9' must be initialized by a constant expression}}
+// since-cxx14-note@#cwg2765-f9-compare {{comparison of addresses of potentially non-unique objects has unspecified value}}
```
Endilll:
```suggestion
//   since-cxx14-note@#cwg2765-f9-compare {{comparison of addresses of potentially non-unique objects has unspecified value}}
```

clang/test/CXX/drs/cwg27xx.cpp
```diff
@@ -19,22 +19,25 @@
...
-  const T* begin() const { return p; };
-};
+  using size_t = decltype(sizeof(int));
+
+  template <class E> class initializer_list {
+    const E *begin_;
+    size_t size_;
+
+  public:
+    constexpr initializer_list() : begin_(nullptr), size_(0) {}
```
Endilll: `std::initializer_list` wasn't `constexpr` in C++11

clang/lib/AST/ByteCode/Interp.cpp
```diff
@@ -2527,6 +2527,83 @@ bool arePotentiallyOverlappingStringLiterals(const Pointer &LHS,
...
+getInitializerListBackingArray(const Pointer &Ptr) {
+  if (Ptr.isZero() || !Ptr.isBlockPointer() || Ptr.block()->isDynamic())
+    return nullptr;
+
+  const auto *MTE =
+      dyn_cast_or_null<MaterializeTemporaryExpr>(Ptr.getDeclDesc()->asExpr());
+  return IsInitializerListBackingArray(MTE) ? MTE : nullptr;
+}
+
+static bool getArrayIndex(const Pointer &Ptr, Pointer &Array, uint64_t &Index) {
```
tbaederr: What does this do?

clang/lib/AST/ByteCode/Interp.cpp
```diff
@@ -2527,6 +2527,83 @@ bool arePotentiallyOverlappingStringLiterals(const Pointer &LHS,
   return Shorter == Longer.take_front(Shorter.size());
 }
 
+static const MaterializeTemporaryExpr *
+getInitializerListBackingArray(const Pointer &Ptr) {
```
tbaederr: I am confused about the naming here since this returns neither an array nor an initializer list?

clang/lib/AST/ByteCode/Interp.cpp
```diff
@@ -2527,6 +2527,83 @@ bool arePotentiallyOverlappingStringLiterals(const Pointer &LHS,
...
+  if (!Expanded.inArray())
+    return false;
+
+  Index = Expanded.isOnePastEnd() ? Expanded.getNumElems()
+                                  : static_cast<uint64_t>(Expanded.getIndex());
+  Array = Expanded.getArray();
+  return true;
+}
+
+bool arePotentiallyOverlappingInitListBackingArrays(InterpState &S,
```
tbaederr: And what does this do?

clang/lib/AST/ByteCode/Interp.h
```diff
@@ -1317,9 +1323,15 @@ inline bool CmpHelperEQ<Pointer>(InterpState &S, CodePtr OpPC, CompareFn Fn) {
...
+  if (arePotentiallyOverlappingInitListBackingArrays(S, LHS, RHS)) {
+    const SourceInfo &Loc = S.Current->getSource(OpPC);
+    S.FFDiag(Loc, diag::note_constexpr_non_unique_object_comparison);
+    return false;
+  }
+
   // Otherwise we need to do a bunch of extra checks before returning Unordered.
-  if (LHS.isOnePastEnd() && !RHS.isOnePastEnd() && RHS.isBlockPointer() &&
-      RHS.getOffset() == 0) {
+  if (LHS.isOnePastEnd() && !RHS.isOnePastEnd() && !RHS.isZero() &&
```
tbaederr: There's an `assert(!RHS.isZero())` above.

clang/lib/AST/ExprConstant.cpp
yronglin: Done. The function name `ArePotentiallyOverlappingStringLiterals` is preserved. The PR only introduces additional APValue overloads on top of the existing name. We don't need renames of pre-existing APIs.

clang/lib/AST/ExprConstant.cpp
yronglin: Done.

clang/lib/AST/ExprConstant.cpp
yronglin: Yeah, I made a major rework. You were right that the three-valued APValueEquality lattice with "treat Unknown like Equal" silently coerced "everything we don't model" into "equal", which is the wrong default. I've removed the entire compareAPValues family (~300 lines) and rebuilt with the helper on APValue::Profile, which is the existing structural-fingerprint primitive used by FoldingSet:
```cpp
   bool AreAPValuesPotentiallyMergeable(LHS, RHS, Ctx) {
     // Fast path: bitwise-identical Profile ⇒ definitely mergeable.
     llvm::FoldingSetNodeID LID, RID;
     LHS.Profile(LID); RHS.Profile(RID);
     if (LID == RID) return true;
     // Slow path: every disagreement must be link-time resolvable.
     return AreAPValuesPotentiallyMergeableSlow(LHS, RHS, Ctx);
   }
 ```

clang/lib/CodeGen/CGExpr.cpp
yronglin: I've reverted the CodeGen changes entirely. The patch now lives purely in constant evaluation, so the array-type heuristic in CodeGen is gone.

clang/lib/CodeGen/CGExpr.cpp
yronglin: Same as above. There's no CGExpr change anymore.

clang/lib/CodeGen/CodeGenModule.cpp
yronglin: Obsoleted by the CodeGen revert. If we want to revisit `unnamed_addr` on backing arrays later, it's an orthogonal change separable from this PR. The initial idea was to prevent a divergence between runtime behavior and constant calculation behavior.

clang/www/cxx_status.html
yronglin: We have added a new `note_constexpr_non_unique_object_comparison`,  "comparison of addresses of potentially non-unique objects has unspecified value".

clang/test/CXX/drs/cwg27xx.cpp
yronglin: Done

clang/test/CXX/drs/cwg27xx.cpp
yronglin: Removed the `#if __cplusplus == 199711L` wrapper. The basic `static_assert`s now run unguarded across all language modes.

clang/test/CXX/drs/cwg27xx.cpp
yronglin: Done and thanks for the tips. I wasn't aware of this format before.

clang/test/CXX/drs/cwg27xx.cpp
yronglin: Done

clang/test/CXX/drs/cwg27xx.cpp
yronglin: Fixed. The minimal `std::initializer_list` stub in the test now gates `constexpr` on its member functions with `#if __cplusplus >= 201402L`.

clang/lib/AST/ByteCode/Interp.h
yronglin: Removed.

clang/lib/AST/ByteCode/Interp.cpp
yronglin: Renamed to `getArraySubobjectLocation` and added more comments in code. It locates a pointer relative to its enclosing initializer-list backing array — returning the array, the element index (or the array length when one-past-the-end), and the offset within that element — or `std::nullopt` when the pointer does not designate such a position. It mirrors the AST-side overload with the same name in `ExprConstant.cpp`, and the bounds/offset arithmetic is now shared via `getArraySubobjectLocationImpl` in `ExprConstShared.h`.

clang/lib/AST/ByteCode/Interp.cpp
yronglin: I have added more comments. It implements the address-comparison rejection rule from `[intro.object]/9` (CWG2765): given two pointers that both designate elements of `std::initializer_list` backing arrays, it returns `true` when the two backing arrays could be merged by the implementation — i.e., their contents agree on the overlapping region — in which case the comparison is unspecified and must be rejected in constant evaluation. It mirrors the AST-side helper of the same purpose in `ExprConstant.cpp`, including the conservative "missing element data ⇒ assume potential merge" default.

clang/lib/AST/ByteCode/Interp.cpp
yronglin: Renamed to `isInitializerListBackingArray`, to match its AST-side counterpart. The previous version returned the `MaterializeTemporaryExpr` because callers needed it; with the new `isBackingArrayForInitializerList()` bit on `MaterializeTemporaryExpr`, all that callers want is a yes/no, so the name and return type are now consistent.

clang/lib/AST/ExprConstant.cpp
yronglin: Good catch! This should have been an assertion rather than a silent `Unknown`. But after the rewrite onto `APValue::Profile`, the entire path is gone: kind is the first field profiled, so kind-mismatched values never compare equal, and the slow path bails on kind mismatch before recursing.

efriedma-quic (commented):
I don't think you can separate the initializer backing array check from the string literal check.  All three categories of non-unique object (string literal, init list backing array, array template parameter) can overlap with each other.  e.g. `static_assert(std::initializer_list<char>{'h','e','l','l','o','\0'}.begin() != "hello");` is ill-formed.

yronglin:
Thanks for the review! 

> I don't think you can separate the initializer backing array check from the string literal check. All three categories of non-unique object (string literal, init list backing array, array template parameter) can overlap with each other.  e.g. `static_assert(std::initializer_list<char>{'h','e','l','l','o','\0'}.begin() != "hello");` is ill-formed.

Sorry, I might misunderstood, https://godbolt.org/z/T56cs3P73, this case seems not overlapped in runtime.

efriedma-quic:
The C++ standard rule for constant evaluation cares about whether it's allowed to overlap.  It doesn't matter whether clang can currently generate code where the overlap actually happens.

PR #197459, opened 2026-05-13
[libc] Introduce a typed syscall wrapper and use it in mmap

labath:
Linux reserves a range of values (everything above -4096u, aka MAX_ERRNO) as an error value, so the check can be performed without knowing the details of the specific syscall. libc functions where these values would be a valid result (e.g. PTRACE_PEEKDATA) are implemented differently at the kernel level (e.g. returning the result through a pointer argument). The only exception are a handful of syscalls (getpid, getuid, ...) which can never fail, and where this could be an actual user/group ID (particularly on 32-bit systems).

Specifically, for mmap, this lets us remove the is_valid_mmap helper and SYS_mmap2 ifdefs in various places.

More generally, this can simplify many syscall wrappers as often the only thing they are doing is converting the return value into an ErrorOr.


libc/src/__support/OSUtil/linux/syscall.h
```diff
@@ -29,24 +30,31 @@
...
-LIBC_INLINE constexpr bool is_valid_mmap(PointerLike ptr) {
-  long addr = cpp::bit_cast<long>(ptr);
-  return LIBC_LIKELY(addr > 0 || addr < -static_cast<long>(MAX_ERRNO));
+
+// Helper function to perform a system call, check the result and cast the
+// result to the expected type. This function is safe to use on most syscalls,
+// with the exception of a handful of syscalls (getpid, getuid, ...) that never
+// fail.
+template <typename R, typename... Ts>
+LIBC_INLINE ErrorOr<R> __syscall(long __number, Ts... ts) {
```
kaladron: I don't love the __ prefix.  We're already namespaced.  I think it makes sense to call it something obvious with tricks that look C-ish.

libc/startup/linux/x86_64/CMakeLists.txt
```diff
@@ -4,9 +4,11 @@ add_startup_object(
     tls.cpp
   DEPENDS
     libc.config.app_h
+    libc.hdr.sys_mman_macros
     libc.include.sys_mman
```
kaladron: I think libc.include.sys_mman isn't needed anymore after the removal.

libc/src/__support/OSUtil/linux/syscall.h
```diff
@@ -29,24 +30,31 @@
 
 namespace LIBC_NAMESPACE_DECL {
 
+// This function performs no error checking. For most syscalls, it's better to
+// use __syscall below.
```
kaladron: Once the migration is complete, does this function go away?  If yes, it might be less confusing to duplicate the logic below so it's obvious to just delete this when it has no callers.

kaladron (approved):
Minor nits.  I love the clarity that this brings and the boilerplate reduction.

libc/src/__support/OSUtil/linux/syscall.h
labath: It is  actually duplicated. The other function is calling the non-template asm `syscall_impl` functions defined in architecture specific headers.

libc/src/__support/OSUtil/linux/syscall.h
labath: Any suggestions on the name? I wanted to call it a plain `syscall`, but that's a macro (defined in glibc headers, I think).

michaelrj-google: `syscall_checked` or something similar could work, since it's handling some of the logic itself.

michaelrj-google (approved):
Overall LGTM, feel free to merge once we're done bikeshedding over the name.

SchrodingerZhu (approved):

libc/src/__support/OSUtil/linux/syscall.h
labath: Going with syscall_checked

michaelrj-google (approved):
`syscall_checked` SGTM, thanks for working on this!

PR #197460, opened 2026-05-13
[CIR] Lower 'init' functions for global TLS

erichkeane:
This is the last patch for global/namespace thread-local variables. This patch emits the final 'init' function, which calls all other init functions, plus does the guard variable for the unordered variants.

### DIFF
--- a/clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
+++ b/clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
@@ -129,6 +129,12 @@ struct LoweringPreparePass
   /// Get the declaration for the 'wrapper' function for a global-TLS variable.
   cir::FuncOp getOrCreateThreadLocalWrapper(CIRBaseBuilderTy &builder,
                                             cir::GlobalOp op);
+  // Function that generates the guard global variable, get-global, and 'if'
+  // condition for global TLS init function generation. This inserts an 'if'
+  // with the store at the beginning of the 'then' region, so inserts into the
+  // body should happen after that.
+  cir::IfOp buildGlobalTlsGuardCheck(CIRBaseBuilderTy &builder,
+                                     mlir::Location loc, cir::GlobalOp guard);
   /// Handle the dtor region by registering destructor with __cxa_atexit
   cir::FuncOp getOrCreateDtorFunc(CIRBaseBuilderTy &builder, cir::GlobalOp op,
                                   mlir::Region &dtorRegion,
@@ -180,6 +186,10 @@ struct LoweringPreparePass
 
   /// Get or create the __init_tls function.
   cir::FuncOp getTlsInitFn();
+
+  // Create the __tls_guard variable.
+  cir::GlobalOp createGlobalThreadLocalGuard(CIRBaseBuilderTy &builder,
+                                             mlir::Location loc);
 
   /// Create a guard global variable for a static local.
   cir::GlobalOp createGuardGlobalOp(CIRBaseBuilderTy &builder,
@@ -377,6 +387,9 @@ struct LoweringPreparePass
     entryBB.getOperations().splice(entryBB.end(), dtorBlock.getOperations(),
                                    dtorBlock.begin(),
                                    std::prev(dtorBlock.end()));
+    // make sure we leave the insert location after the operations we just
+    // inserted.
+    builder.setInsertionPointToEnd(&entryBB);
   }
 
   /// Emit the guarded initialization for a static local variable.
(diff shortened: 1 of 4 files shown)


clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
```diff
@@ -1143,22 +1156,43 @@ LoweringPreparePass::buildCXXGlobalVarDeclInitFunc(cir::GlobalOp op) {
...
+  }
+
   if (!op.getCtorRegion().empty()) {
     mlir::Block &block = op.getCtorRegion().front();
-    entryBB->getOperations().splice(entryBB->begin(), block.getOperations(),
-                                    block.begin(), std::prev(block.end()));
+    mlir::Block *insertBlock = builder.getBlock();
+    insertBlock->getOperations().splice(insertBlock->end(),
+                                        block.getOperations(), block.begin(),
+                                        std::prev(block.end()));
```
andykaylor, line 1180: Is the ctor region guaranteed to contain a single block? For instance, can this be initialized using an expression with cleanups?

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
```diff
@@ -1143,22 +1156,43 @@ LoweringPreparePass::buildCXXGlobalVarDeclInitFunc(cir::GlobalOp op) {
...
 
     emitGlobalGuardedDtorRegion(builder, op, dtorRegion,
-                                op.getTlsModel().has_value(), *entryBB);
+                                op.getTlsModel().has_value(),
+                                *builder.getBlock());
   }
 
+  // If we're actually in the 'if' above, create a yield.
+  if (needsTlsGuard)
+    cir::YieldOp::create(builder, op.getLoc());
```
andykaylor, line 1196: What is guaranteeing that the insert location will be correct for this?

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
```diff
@@ -1710,9 +1744,57 @@ void LoweringPreparePass::buildGlobalCtorDtorList() {
...
 }
 
+cir::GlobalOp
+LoweringPreparePass::createGlobalThreadLocalGuard(CIRBaseBuilderTy &builder,
+                                                  mlir::Location loc) {
+  mlir::OpBuilder::InsertionGuard guard(builder);
+  builder.setInsertionPointToStart(mlirModule.getBody());
+
+  // The TLS Guard is always an Int8Ty.
+  auto guardTy = builder.getSIntNTy(8);
```
andykaylor:
```suggestion
  cir::IntType guardTy = builder.getSIntNTy(8);
```

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
```diff
@@ -1710,9 +1744,57 @@ void LoweringPreparePass::buildGlobalCtorDtorList() {
...
+  cir::GlobalOp g = cir::GlobalOp::create(builder, loc, "__tls_guard", guardTy);
```
andykaylor:
```suggestion
  auto g = cir::GlobalOp::create(builder, loc, "__tls_guard", guardTy);
```

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
```diff
@@ -1710,9 +1744,57 @@ void LoweringPreparePass::buildGlobalCtorDtorList() {
...
+  g.setLinkageAttr(cir::GlobalLinkageKindAttr::get(
+      builder.getContext(), cir::GlobalLinkageKind::InternalLinkage));
+  g.setAlignment(clang::CharUnits::One().getAsAlign().value());
+  // At the moment, we only have implementation for this mode, as it is the
+  // default.  At one point we might need to load this mode from the module.
+  g.setTlsModel(TLS_Model::GeneralDynamic);
+  // g.setDSOLocal(
```
andykaylor: ?

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
```diff
@@ -1710,9 +1744,57 @@ void LoweringPreparePass::buildGlobalCtorDtorList() {
...
+  // unordered TLS, it gets created as 64 bits.
+  if (guard.getSymType() != builder.getSIntNTy(8))
+    getGuardValue = builder.createBitcast(
+        getGuard, cir::PointerType::get(builder.getSIntNTy(8)));
+
+  mlir::Value guardLoad =
+      builder.createAlignedLoad(loc, getGuardValue, *guard.getAlignment());
+  auto zero = builder.getConstantInt(loc, builder.getSIntNTy(8), 0);
+  auto compare =
+      builder.createCompare(loc, cir::CmpOpKind::eq, guardLoad, zero);
```
andykaylor, line 1785: Is auto OK here?

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
```diff
@@ -1721,9 +1803,20 @@ void LoweringPreparePass::buildCXXGlobalTlsFunc() {
...
   mlir::Block *entryBB = tlsInit.addEntryBlock();
   builder.setInsertionPointToStart(entryBB);
-  // Note: a followup patch will emit the body here correctly.
+
+  cir::IfOp ifOperation = buildGlobalTlsGuardCheck(
+      builder, loc, createGlobalThreadLocalGuard(builder, loc));
+
+  // Emit the body of the guarded spot.
+  builder.setInsertionPointToEnd(&ifOperation.getThenRegion().front());
+  for (auto initFunc : globalThreadLocalInitializers)
```
andykaylor: cir::FuncOp?

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
erichkeane: I do not know/don't believe so?  We should probably do better here perhaps.  I didn't do any changes here besides wrapping the existing codegen in an 'if' block though.

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
erichkeane: The change at `emitGlobalGuardedDtorRegion` is what I had to change to make sure it was always done (since the dtor region wasn't leaving us in the 'right' place), but this yield needs to be in the closing block of each of these.  I guess it makes sense to do a 'put at the end of the region' though in case we end up adding extra blocks/etc above.

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
erichkeane: I think that ended up being a part of a failed experiment, no idea.  Thanks!

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
erichkeane: definitely borderline...

clang/lib/CIR/Dialect/Transforms/LoweringPrepare.cpp
erichkeane: You're getting an unfortunate view of my policy on 'almost-always-auto' while writing code :D

andykaylor (approved):
lgtm

PR #197464, opened 2026-05-13
[BOLT] Gate PointerAuthCFIFixup unit test on AArch64 target availability

androm3da:
The test bodies reference AArch64:: namespace identifiers (ADDSXri, X0) which fail to compile when AArch64 is not in LLVM_TARGETS_TO_BUILD. Wrap all TEST_P bodies in #ifdef AARCH64_AVAILABLE and add GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST to suppress GoogleTest's uninstantiated suite error when no target instantiates the tests.

### DIFF
--- a/bolt/unittests/Passes/PointerAuthCFIFixup.cpp
+++ b/bolt/unittests/Passes/PointerAuthCFIFixup.cpp
@@ -122,6 +122,9 @@ protected:
 };
 } // namespace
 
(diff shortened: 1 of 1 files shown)


paschalis-mpeis (approved):
Thanks!

bgergely0 (approved):
LGTM, thanks

PR #197465, opened 2026-05-13
[flang][acc] Support walking through acc.compute_region ins

razvanlupusoru:
Since acc.compute_region is IsolatedFromAbove, its values are threaded through block arguments. But the implementations using getOriginalDef want to see the original variable input in order to make their assessment.


clementval (approved):
LGTM

PR #188781, opened 2026-03-26
[GlobalISel][AMDGPU][AArch64] Fix GlobalISel copy propagation

IgWod:
Disallow propagation of sub-registers after GlobalISel, as the current code is blindly dropping any sub-register information. This also fixes bugs in AArch64 and AMDGPU back-end that rely on the incorrect behavior and would fail with the fix:

* Update `selectG_UNMERGE_VALUES` in AMDGPU so instead of generating `hi16` for SGPR it shifts higher bits into the destination register using `lshr`.
* Prevent AArch64 back-end from generating spurious `sub_32:gpr32all` when selecting copy.
* Test changes: `fpto[s/u]i-sat-vector.ll`: The correct number of conversions is now generated as higher 16-bits are handled correctly; however, it introduces `lshr` instructions. This should be resolved in #188287 by enabling `s_cvt_hi_*`.

More context: #188287 and #188019.

PR #197467, opened 2026-05-13
[AMDGPU] Remove definition of hi16 for scalar registers

IgWod:
The sub-register is not supported by the back-end pipelines and will eventually cause an assert. With this change machine verifier can enforce that hi16 is not being used. Asm parser is also updated to error on invalid sub-registers, instead of asserting.

This is a follow up change on #188781.

Assisted-by: Claude Code


llvm/lib/Target/AMDGPU/SIRegisterInfo.td
```diff
@@ -143,11 +143,13 @@ multiclass SIRegLoHi16 <string n, bits<10> regIdx, bit ArtificialHigh = 1,
...
-  def "" : RegisterWithSubRegs<n, [!cast<Register>(NAME#"_LO16"),
-                                   !cast<Register>(NAME#"_HI16")]>,
+  def "" : RegisterWithSubRegs<n, !if(!not(ArtificialHigh),
+                                      [!cast<Register>(NAME#"_LO16"),
+                                       !cast<Register>(NAME#"_HI16")],
+                                      [!cast<Register>(NAME#"_LO16")])>,
            DwarfRegNum<DwarfEncodings> {
     let Namespace = "AMDGPU";
-    let SubRegIndices = [lo16, hi16];
+    let SubRegIndices = !if(!not(ArtificialHigh), [lo16, hi16], [lo16]);
```
jayfoad: Slight preference for avoiding the `!not`:
```suggestion
    let SubRegIndices = !if(ArtificialHigh, [lo16], [lo16, hi16]);
```
... and the same above.

llvm/lib/Target/AMDGPU/SIRegisterInfo.td
jayfoad: Do you still need to define the _HI16 register at all if ArtificialHigh is set? Do you still need the SGPR_HI16 RegisterClass?

llvm/lib/Target/AMDGPU/SIRegisterInfo.td
```diff
@@ -143,11 +143,13 @@ multiclass SIRegLoHi16 <string n, bits<10> regIdx, bit ArtificialHigh = 1,
                     /* isHi16 */ 1> {
     let isArtificial = ArtificialHigh;
   }
-  def "" : RegisterWithSubRegs<n, [!cast<Register>(NAME#"_LO16"),
-                                   !cast<Register>(NAME#"_HI16")]>,
+  def "" : RegisterWithSubRegs<n, !if(!not(ArtificialHigh),
```
shiltian: What does it mean by "artificial high"?

jayfoad: It means it sets isArtificial in the _HI16 reg (among other things). From https://reviews.llvm.org/D78591

llvm/lib/Target/AMDGPU/SIRegisterInfo.td
IgWod: Good point, I don't think you need _HI16 anymore. I'll clean that up.

broxigarchen: Thanks for the patch. I was testing similar patch for 16bit reg in sdag. I think this can go in first.

One small nit. Should we seperate the SIRegLoHi16 to two class, one for artifical high, and one is not? This looks a little ugly.

IgWod: > One small nit. Should we seperate the SIRegLoHi16 to two class, one for artifical high, and one is not? This looks a little ugly.

This is exactly what I'm working on right now :)

IgWod:
I've now updated the code, splitting `SIRegLoHi16` multiclass into two classes with a common parent to avoid duplication. I have also updated asm parser to avoid an assertion on invalid sub-registers. There were few extra tests changes but it all looks reasonable.

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
```diff
@@ -730,7 +730,8 @@ bool AMDGPUInstructionSelector::selectG_UNMERGE_VALUES(MachineInstr &MI) const {
   ArrayRef<int16_t> SubRegs = TRI.getRegSplitParts(SrcRC, DstSize / 8);
   for (int I = 0, E = NumDst; I != E; ++I) {
     MachineOperand &Dst = MI.getOperand(I);
-    // hi16:sreg_32 is not allowed so explicitly shift upper 16-bits.
+    // SGPRs do not support hi16 subregister, so extract upper 16-bits with a
+    // shift and skip the subclass constraint step.
     if (SrcBank->getID() == AMDGPU::SGPRRegBankID &&
         SubRegs[I] == AMDGPU::hi16) {
       BuildMI(*BB, &MI, DL, TII.get(AMDGPU::S_LSHR_B32), Dst.getReg())
```
arsenm: The register still needs to be constrained to the correct register class, it just resets the subregister to none

IgWod: Do you mean to still call `constrainGenericRegister` for the `Src` register in this case? I think that's already covered by the call to `constrainGenericRegister` before the loop - line 724, and we don't need to do it again, but I'm not 100% sure.

arsenm: We should only be calling constrainGenericRegister once. The above one probably should be removed

IgWod: I've removed call before the loop and moved `constrainGenericRegister` from inside `else` to outside `if/else`.

llvm/test/CodeGen/AMDGPU/wqm.ll
```diff
@@ -1474,7 +1474,6 @@ define amdgpu_ps float @test_control_flow_1(<8 x i32> inreg %rsrc, <4 x i32> inr
 ; GFX10-W32-NEXT:  .LBB28_2: ; %Flow
 ; GFX10-W32-NEXT:    s_or_saveexec_b32 s0, s13
 ; GFX10-W32-NEXT:    s_and_b32 exec_lo, exec_lo, s12
-; GFX10-W32-NEXT:    s_and_b32 s0, exec_lo, s0
```
Sisyph, line 1477: This looks suspicious, is this a correctness issue?

llvm/lib/Target/AMDGPU/SIRegisterInfo.td
Sisyph: @rampitec Told me in the past that we needed to have the same subregisters available for sgprs and vgprs. But I don't remember why. Could it be an implementation constraint with the register infrastructure that is no longer needed? This patch seems to show it is not needed anymore.

rampitec: There were problems with lanemask computations, but I do not recall exact details now.

llvm/test/CodeGen/AMDGPU/wqm.ll
IgWod: This seems to be exactly the problem mentioned by @rampitec. I thought it was a valid change, but now I looked deeper and this indeed seems to be a mis-compilation. I'll look into it.

llvm/lib/Target/AMDGPU/SIRegisterInfo.td
```diff
@@ -158,9 +156,27 @@ multiclass SIRegLoHi16 <string n, bits<10> regIdx, bit ArtificialHigh = 1,
...
 }
 
+multiclass SIRegLoHi16 <string n, bits<10> regIdx,
+                        bit isVGPR = 0, bit isAGPR = 0,
+                        list<int> DwarfEncodings = [-1, -1]> {
+  def _HI16 : SIReg<n#".h", regIdx, isVGPR, isAGPR, /* isHi16 */ 1>;
+  defm "" : SIRegWithSubRegs<n, regIdx, isVGPR, isAGPR, DwarfEncodings,
+                             [!cast<Register>(NAME#"_LO16"),
+                             !cast<Register>(NAME#"_HI16")],
+                             [lo16, hi16], /* coveredBySubRegs */ 1>;
```
arsenm:
```suggestion
                              !cast<Register>(NAME#"_HI16")],
                              [lo16, hi16], /* coveredBySubRegs=*/true>;
```

llvm/lib/Target/AMDGPU/SIRegisterInfo.td
```diff
@@ -158,9 +156,27 @@ multiclass SIRegLoHi16 <string n, bits<10> regIdx, bit ArtificialHigh = 1,
...
+}
+
+multiclass SIRegLo16 <string n, bits<10> regIdx,
+                      bit isVGPR = 0, bit isAGPR = 0,
+                      list<int> DwarfEncodings = [-1, -1]> {
+  defm "" : SIRegWithSubRegs<n, regIdx, isVGPR, isAGPR, DwarfEncodings,
+                             [!cast<Register>(NAME#"_LO16")],
+                             [lo16], /* coveredBySubRegs */ 0>;
```
arsenm:
```suggestion
                             [lo16], /*coveredBySubRegs=*/ false>;
```

arsenm (approved):

llvm/lib/Target/AMDGPU/SIRegisterInfo.td
jayfoad: You can also use real syntax `coveredBySubRegs=false` instead of a comment.

IgWod:
So, it looks like mask problems are related to liveness analysis - it looks like it relies on hi16 being available - but I haven't investigated in depth why that is. With that in mind I decided to keep hi16 for EXEC and VCC for now and that's something we can address in the future. Let me know if that works - we still get a verification for most of the registers and I think it's a step in the right direction in getting rid of artificial hi16 support.

IgWod:
ping, any objections to the halfway solution that leaves hi16 on exec masks? If not I assume I can merge this PR as it is now.

llvm/lib/Target/AMDGPU/SIRegisterInfo.td
```diff
@@ -158,9 +156,32 @@ multiclass SIRegLoHi16 <string n, bits<10> regIdx, bit ArtificialHigh = 1,
...
+  defm "" : SIRegWithSubRegs<n, regIdx, isVGPR, isAGPR, DwarfEncodings,
+                             [!cast<Register>(NAME#"_LO16")],
+                             [lo16], coveredBySubRegs=false>;
+}
+
 // Special Registers
-defm VCC_LO : SIRegLoHi16<"vcc_lo", 106>;
-defm VCC_HI : SIRegLoHi16<"vcc_hi", 107>;
+// TODO: To remove hi16 from EXEC and VCC we need to fix an issues with
+// liveness analysis that relies on the availability of hi16.
```
arsenm: I'm not really sure what this means

arsenm:
> So, it looks like mask problems are related to liveness analysis - it looks like it relies on hi16 being available - but I haven't investigated in depth why that is. With that in mind I decided to keep hi16 for EXEC and VCC for now and that's something we can address in the future. Let me know if that works - we still get a verification for most of the registers and I think it's a step in the right direction in getting rid of artificial hi16 support.

What specifically is broken?

IgWod:
> > So, it looks like mask problems are related to liveness analysis - it looks like it relies on hi16 being available - but I haven't investigated in depth why that is. With that in mind I decided to keep hi16 for EXEC and VCC for now and that's something we can address in the future. Let me know if that works - we still get a verification for most of the registers and I think it's a step in the right direction in getting rid of artificial hi16 support.
> 
> What specifically is broken?

So, I think my initial analysis was wrong. I assumed it was liveness analysis causing `s_and` to be eliminated in `llvm/test/CodeGen/AMDGPU/wqm.ll`. But it seems `optimizeElseBranch` in `llvm/lib/Target/AMDGPU/SIOptimizeExecMaskingPreRA.cpp` is a problem why the `wqm.ll` test was producing the incorrect code. It seems that the logic removing `s_and` starting on line 336 is being incorrectly executed. I'm going to revert the PR to the previous state and fix the root cause. Sorry, for all the confusion.

IgWod:
So the problem is related to liveness but rather than it being any fundamental problem, it is caused by a stale range. I have submitted a separate fix in #200866 that also explains the issue.

IgWod:
I have now pushed a rebased branch:

* Code is back to when @arsenm approved it + nits are addressed - hi16 is removed from all SGPR registers
* `wqm.ll` diff disappeared due to #200866 being merged

@rampitec @Sisyph Are you happy with the change? The exec bug has been addressed and nothing more shows in tests. I guess if something more surfaces we can deal with it then?

@arsenm I take you are happy with the change since it's back to when you approved it?

Again, sorry for all the confusion caused.

llvm/unittests/Target/AMDGPU/DwarfRegMappings.cpp
```diff
@@ -40,8 +40,7 @@ TEST_F(AMDGPUTestBase, TestWave64DwarfRegMapping) {
         }
 
         // Verify that subregisters have no dwarf encoding.
-        for (MCRegister LLSubReg :
-             {AMDGPU::VGPR1_LO16, AMDGPU::AGPR1_HI16, AMDGPU::SGPR1_HI16}) {
+        for (MCRegister LLSubReg : {AMDGPU::VGPR1_LO16}) {
```
jayfoad: We can still test this for some AGPRs and SGPRs, e.g.:
```suggestion
        for (MCRegister LLSubReg : {AMDGPU::VGPR1_HI16, AMDGPU::AGPR1_LO16, AMDGPU::SGPR1_LO16}) {
```
Same below.

IgWod: Done, I've kept VGPR_LO16 as well.

arsenm (approved):

rampitec:
> @rampitec @Sisyph Are you happy with the change? The exec bug has been addressed and nothing more shows in tests. I guess if something more surfaces we can deal with it then?

If that works I do not object. Although I was hoping one day we support `v_mov_b16 v0.l, s0.h`.

Sisyph (approved):
I think this can go in.

@broxigarchen I think this should prevent a sgpr_hi16 operand showing up in a true16 instruction, but in theory a sgpr_lo16 operand could show up. Do you agree?

This is a wild idea but could we remove _lo16 subreg from SGPR as well?

Sisyph:
> > @rampitec @Sisyph Are you happy with the change? The exec bug has been addressed and nothing more shows in tests. I guess if something more surfaces we can deal with it then?
> 
> If that works I do not object. Although I was hoping one day we support `v_mov_b16 v0.l, s0.h`.

We could support it with assembly/printer sugar of the op_sel bits?

rampitec:
> > > @rampitec @Sisyph Are you happy with the change? The exec bug has been addressed and nothing more shows in tests. I guess if something more surfaces we can deal with it then?
> > 
> > 
> > If that works I do not object. Although I was hoping one day we support `v_mov_b16 v0.l, s0.h`.
> 
> We could support it with assembly/printer sugar of the op_sel bits?

We can, but it is the same true16. I.e., in codegen it would have saved us some register pressure. Although if there are no such plans, this change can go in.

broxigarchen:
> @broxigarchen I think this should prevent a sgpr_hi16 operand showing up in a true16 instruction, 

Yes. This is part of the change I was testing locally

> but in theory a sgpr_lo16 operand could show up. Do you agree?

In theory it could show up, but for now probably not. We don't have proper support for sgpr_lo16 operand in the pipeline. If any pass emit sgpr_lo16 and it reach either spilling, or asm, or any pass does not recognize it properly it would raise errors

>This is a wild idea but could we remove _lo16 subreg from SGPR as well?

This seems debatable. I don't know the original purpose to have the sgpr lo16 defined. From true16 perspective, we could end up want to support sgpr_lo16 in the pipeline (also debatable) so we have a dedicate sreg assigned for 16bit types.

rampitec:
> This seems debatable. I don't know the original purpose to have the sgpr lo16 defined. From true16 perspective, we could end up want to support sgpr_lo16 in the pipeline (also debatable) so we have a dedicate sreg assigned for 16bit types.

I think in this case we will want to have both lo16 and hi16.

rampitec (approved):
OK, having it artificial also does not help. If we ever need it for true16 we can restore it. LGTM.

IgWod:
It looks like the change may have broken the build bot: https://lab.llvm.org/buildbot/#/builders/10/builds/29355. I'm investigating now.

jplehr:
Hits an `unreachable` in libc on GPU buildbot: https://lab.llvm.org/buildbot/#/builders/10/builds/29355

```
FAILED: libc/test/src/time/libc.test.src.time.strftime_test.__hermetic__.__build__ 
: && /home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/./bin/clang++ --target=amdgcn-amd-amdhsa -fPIC -fno-semantic-interposition -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wno-pass-failed -Wmisleading-indentation -Wctad-maybe-unsupported -fdiagnostics-color -ffunction-sections -fdata-sections -Xclang -fno-pch-timestamp -O3 -DNDEBUG -Wl,--color-diagnostics    --target=amdgcn-amd-amdhsa -Wno-multi-gpu -mcpu=native -flto -nostdlib -static -Wl,-mllvm,-amdhsa-code-object-version=6 libc/startup/gpu/CMakeFiles/libc.startup.gpu.crt1.dir/start.cpp.o libc/test/src/time/CMakeFiles/libc.test.src.time.strftime_test.__hermetic__.__build__.dir/strftime_test.cpp.o -o libc/test/src/time/libc.test.src.time.strftime_test.__hermetic__.__build__  libc/test/UnitTest/libLibcTest.hermetic.a  libc/test/UnitTest/libLibcHermeticTestSupport.hermetic.a  libc/test/src/time/liblibc.test.src.time.strftime_test.__hermetic__.libc.a && :
cannot represent required subregister defs
UNREACHABLE executed at /home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.src/llvm/lib/CodeGen/VirtRegMap.cpp:707!
Stack dump:
0.	Program arguments: /home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/ld.lld --no-undefined -shared -plugin-opt=mcpu=gfx90a -plugin-opt=O3 --lto-CGO3 -plugin-opt=-function-sections=1 -plugin-opt=-data-sections=1 -L/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/amdgcn-amd-amdhsa --color-diagnostics -mllvm -amdhsa-code-object-version=6 libc/startup/gpu/CMakeFiles/libc.startup.gpu.crt1.dir/start.cpp.o libc/test/src/time/CMakeFiles/libc.test.src.time.strftime_test.__hermetic__.__build__.dir/strftime_test.cpp.o libc/test/UnitTest/libLibcTest.hermetic.a libc/test/UnitTest/libLibcHermeticTestSupport.hermetic.a libc/test/src/time/liblibc.test.src.time.strftime_test.__hermetic__.libc.a -o libc/test/src/time/libc.test.src.time.strftime_test.__hermetic__.__build__
1.	Running pass 'CallGraph Pass Manager' on module 'ld-temp.o'.
2.	Running pass 'Virtual Register Rewriter' on function '@_ZN43LlvmLibcStrftimeTest_TimeFormatFullDateTime3RunEv'
 #0 0x00007232dee1b870 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/libLLVMSupport.so.23.0git+0x21b870)
 #1 0x00007232dee1843c llvm::sys::RunSignalHandlers() (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/libLLVMSupport.so.23.0git+0x21843c)
 #2 0x00007232dee185a2 SignalHandler(int, siginfo_t*, void*) Signals.cpp:0:0
 #3 0x00007232de442520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #4 0x00007232de4969fc __pthread_kill_implementation ./nptl/pthread_kill.c:44:76
 #5 0x00007232de4969fc __pthread_kill_internal ./nptl/pthread_kill.c:78:10
 #6 0x00007232de4969fc pthread_kill ./nptl/pthread_kill.c:89:10
 #7 0x00007232de442476 gsignal ./signal/../sysdeps/posix/raise.c:27:6
 #8 0x00007232de4287f3 abort ./stdlib/abort.c:81:7
 #9 0x00007232ded0d98e (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/libLLVMSupport.so.23.0git+0x10d98e)
#10 0x00007232d9be2d6d (anonymous namespace)::VirtRegRewriter::rewrite() VirtRegMap.cpp:0:0
#11 0x00007232d9be33f2 (anonymous namespace)::VirtRegRewriter::run(llvm::MachineFunction&) VirtRegMap.cpp:0:0
#12 0x00007232d9be3cc7 (anonymous namespace)::VirtRegRewriterLegacy::runOnMachineFunction(llvm::MachineFunction&) VirtRegMap.cpp:0:0
#13 0x00007232d97fd99b llvm::MachineFunctionPass::runOnFunction(llvm::Function&) (.part.0) MachineFunctionPass.cpp:0:0
#14 0x00007232da293c31 llvm::FPPassManager::runOnFunction(llvm::Function&) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/../lib/libLLVMCore.so.23.0git+0x293c31)
#15 0x00007232d7970c82 (anonymous namespace)::CGPassManager::runOnModule(llvm::Module&) CallGraphSCCPass.cpp:0:0
#16 0x00007232da294c30 llvm::legacy::PassManagerImpl::run(llvm::Module&) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/../lib/libLLVMCore.so.23.0git+0x294c30)
#17 0x00007232def4f6ab codegen(llvm::lto::Config const&, llvm::TargetMachine*, std::function<llvm::Expected<std::unique_ptr<llvm::CachedFileStream, std::default_delete<llvm::CachedFileStream>>> (unsigned int, llvm::Twine const&)>, unsigned int, llvm::Module&, llvm::ModuleSummaryIndex const&) LTOBackend.cpp:0:0
#18 0x00007232def5160d llvm::lto::backend(llvm::lto::Config const&, std::function<llvm::Expected<std::unique_ptr<llvm::CachedFileStream, std::default_delete<llvm::CachedFileStream>>> (unsigned int, llvm::Twine const&)>, unsigned int, llvm::Module&, llvm::ModuleSummaryIndex&, llvm::ArrayRef<llvm::StringRef>) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/../lib/libLLVMLTO.so.23.0git+0x5360d)
#19 0x00007232def29768 llvm::lto::LTO::runRegularLTO(std::function<llvm::Expected<std::unique_ptr<llvm::CachedFileStream, std::default_delete<llvm::CachedFileStream>>> (unsigned int, llvm::Twine const&)>) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/../lib/libLLVMLTO.so.23.0git+0x2b768)
#20 0x00007232def41da6 llvm::lto::LTO::run(std::function<llvm::Expected<std::unique_ptr<llvm::CachedFileStream, std::default_delete<llvm::CachedFileStream>>> (unsigned int, llvm::Twine const&)>, llvm::FileCache) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/../lib/libLLVMLTO.so.23.0git+0x43da6)
#21 0x00007232df5e5d8a lld::elf::BitcodeCompiler::compile() (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/liblldELF.so.23.0git+0x1e5d8a)
#22 0x00007232df527f15 void lld::elf::LinkerDriver::compileBitcodeFiles<llvm::object::ELFType<(llvm::endianness)1, true>>(bool) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/liblldELF.so.23.0git+0x127f15)
#23 0x00007232df559533 void lld::elf::LinkerDriver::link<llvm::object::ELFType<(llvm::endianness)1, true>>(llvm::opt::InputArgList&) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/liblldELF.so.23.0git+0x159533)
#24 0x00007232df561369 lld::elf::LinkerDriver::linkerMain(llvm::ArrayRef<char const*>) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/liblldELF.so.23.0git+0x161369)
#25 0x00007232df56192e lld::elf::link(llvm::ArrayRef<char const*>, llvm::raw_ostream&, llvm::raw_ostream&, bool, bool) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/liblldELF.so.23.0git+0x16192e)
#26 0x00007232df0ec5b1 lld::unsafeLldMain(llvm::ArrayRef<char const*>, llvm::raw_ostream&, llvm::raw_ostream&, llvm::ArrayRef<lld::DriverDef>, bool) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/../lib/liblldCommon.so.23.0git+0xe5b1)
#27 0x00005954b91cbcb6 lld_main(int, char**, llvm::ToolContext const&) (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/ld.lld+0x2cb6)
#28 0x00005954b91cb53e main (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/ld.lld+0x253e)
#29 0x00007232de429d90 __libc_start_call_main ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
#30 0x00007232de429e40 call_init ./csu/../csu/libc-start.c:128:20
#31 0x00007232de429e40 __libc_start_main ./csu/../csu/libc-start.c:379:5
#32 0x00005954b91cb595 _start (/home/botworker/builds/openmp-offload-amdgpu-runtime-2/llvm.build/bin/ld.lld+0x2595)
clang++: error: unable to execute command: Aborted (core dumped)
clang++: error: ld.lld command failed with exit code -2 (use -v to see invocation)
Aborted (core dumped)
```

IgWod:
It will require some more digging, so I created a revert #201309 to keep the buildbot happy. I'll merge it once CI passes.

arsenm:
> cannot represent required subregister defs

This is a pain. I wish MIR had a more direct way to represent the liveness of a register + lane mask instead of building these COPY bundles with a sequence of subregister operands. Something like a COPY_LANEMASK that directly encodes the live lanes. Ultimately we'd still need to lower that into a register RMW for the cases that only preserve unallocatable pieces of registers to encode the instruction

PR #197469, opened 2026-05-13
[SPIR-V] Match OpPtrEqual/OpPtrNotEqual operand types via OpBitcast

aobolensk:
Function pointer and opaque pointer comparison fails due to the fact that pointer from OpConstantFunctionPointerINTEL extension type differs from i8. OpPtrEqual/OpPtrNotEqual require identical operand pointer types, so bitcast the RHS to the LHS type is inserted in case they differ

fixes https://github.com/llvm/llvm-project/issues/185409


llvm/test/CodeGen/SPIRV/extensions/SPV_INTEL_function_pointers/fp_cmp.ll
```diff
@@ -12,10 +8,11 @@
 ; CHECK: OpName %[[ARG:.*]] "arg"
 
 ; CHECK: %[[TyBool:.*]] = OpTypeBool
+; CHECK: %[[TyFnPtr:.*]] = OpTypePointer CodeSectionINTEL %{{.*}}
+; CHECK: %[[F1Ptr:.*]] = OpConstantFunctionPointerINTEL %[[TyFnPtr]] %[[F1]]
 
-; CHECK: %[[F1Ptr:.*]] = OpConstantFunctionPointerINTEL %{{.*}} %[[F1]]
-
-; CHECK: OpPtrEqual %[[TyBool]] %[[F1Ptr]] %[[ARG]]
+; CHECK: %[[ArgCast:.*]] = OpBitcast %[[TyFnPtr]] %[[ARG]]
```
jmmartinez, line 17: Can we add a check on the type of `ARG` ?

Just to ensure we're not doing a weird cast (I guess it'd be something like pointer to `i8` or something like that).

aobolensk: Yes, fair enough, added

jmmartinez (approved):

llvm/test/CodeGen/SPIRV/extensions/SPV_INTEL_function_pointers/fp_cmp.ll
jmmartinez: Thanks !

PR #197472, opened 2026-05-13
[lldb] Use current breakpoint location in TestBreakpointLocationDot.py (NFC)

kastiglione:
Replaces `FindLocationByID(1)` with `FindLocationByID(loc_id)`. Instead of hard-coding an ID of 1, the location ID is determined from `GetStopReasonDataAtIndex`.

Some of these asserts were failing on a windows CI, because the breakpoint was resolved to **two** locations.

### DIFF
--- a/lldb/test/API/commands/breakpoint/location-dot/TestBreakpointLocationDot.py
+++ b/lldb/test/API/commands/breakpoint/location-dot/TestBreakpointLocationDot.py
@@ -6,29 +6,29 @@ from lldbsuite.test import lldbutil
 class TestCase(TestBase):
     def test_disable_enable(self):
         self.build()
-        _, _, _, bp = lldbutil.run_to_source_breakpoint(
+        _, _, thread, bp = lldbutil.run_to_source_breakpoint(
             self, "break here", lldb.SBFileSpec("main.c")
         )
 
-        self.assertEqual(bp.GetNumLocations(), 1)
-        self.assertTrue(bp.GetLocationAtIndex(0).IsEnabled())
+        loc_id = self._stop_location_id(thread)
+        self.assertTrue(bp.FindLocationByID(loc_id).IsEnabled())
         self.expect("breakpoint disable .", startstr="1 breakpoints disabled.")
-        self.assertFalse(bp.GetLocationAtIndex(0).IsEnabled())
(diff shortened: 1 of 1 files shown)


jimingham (approved):
There is no guarantee that the first location we find will show up in the program.  For instance, if you guess pre-run that /usr/lib/liblibrary.dylib will load and set a location there, but on running ~/liblibrary.dylib loads instead, the first location will get discarded.  So this is a more correct change.

But it's surprising to me that this would happen in the case that your test covers.  Do you know why that happens?  Note - the correctness of the patch doesn't depend on the answer to the question, I'm just curious.

kastiglione:
I'm investigating to find out what's going on.

kastiglione:
@jimingham the problem is that on windows, the "break here" source regex breakpoint resolves to two locations. In that case the first breakpoint location was not correct. I've updated the test to get the location ID from the stop reason data, ensuring a correct location ID is passed to `FindLocationByID`.

lldb/test/API/commands/breakpoint/location-dot/TestBreakpointLocationDot.py
```diff
@@ -6,46 +6,62 @@
...
         )
-        self.assertTrue(bp.FindLocationByID(1).IsEnabled())
+        self.assertTrue(loc.IsEnabled())
 
     def test_error_no_process(self):
         self.build()
         target = self.createTestTarget()
         target.BreakpointCreateByLocation("main.c", 2)
         self.expect("breakpoint disable .", error=True, substrs=["no current thread"])
+
```
jimingham, line 55: Should this be in lldbutil.py?

kastiglione: @jimingham  I missed this comment. Which part are you wondering about?

kastiglione: oh do you mean the `_stop_location` function? I can if you'd like. I don't see anywhere else I could re-use this function, fwiw.

jimingham: That just seemed like a safer way to figure out where you actually stopped when a breakpoint has more than one location.  But if it doesn't seem generally useful, this is fine.

PR #177665, opened 2026-01-23
[PGO][AMDGPU] Add basic HIP offload PGO support

yxsamliu:
Provide the minimum HIP/offload path for device profile collection and
merge on HIP before layering profile-format and uniformity-specific
changes separately.

This adds the ROCm collection runtime, hooks device collection into the
host write-file path, lowers AMDGPU instrumentation to
__llvm_profile_instrument_gpu with regular counters, and disables GPU
indirect-call value profiling.

PR #197474, opened 2026-05-13
[compiler-rt][profile][test] Match clang_rt.profile CRT model on MSVC

yxsamliu:
On MSVC, Profile-* tests must link with the same CRT model as the
clang_rt.profile static archive they exercise. When that archive pulls
in RTInterception / RTSanitizerCommon object libraries, those are built
with MultiThreadedDLL (/MD), so the .objs reference `__imp_*` symbols.
The test binary defaults to /MT and fails to link with LNK2019
(`__imp__stricmp` from `interception_win.cpp`) and LNK4098 default-lib
conflicts.

Match the DLL CRT on the test side so test executables and the static
archive use the same runtime. The change is gated on
`COMPILER_RT_HAS_INTERCEPTION` and `!COMPILER_RT_PROFILE_BAREMETAL`, so
configurations that don't pull interception into profile are unaffected.

Split out as NFC from #177665 per review feedback.


jhuber6 (approved):
Comment is a little verbose, but LGTM

vitalybuka (approved):

PR #197475, opened 2026-05-13
[SystemZ] Don't fold after SSA if tied regs don't match.

### LINKED ISSUE #197414: [SystemZ] Machine verifier reports "Bad machine code: Two-address instruction operands must be identical"
Reduced test case (from povray built with -mllvm -verify-machineinstrs):

llc -mtriple=s390x-linux-gnu -mcpu=z17 -O3 tc_rematandfold.ll -o out.s -verify-machineinstrs

[tc.tar.gz](https://github.com/user-attachments/files/27704946/tc.tar.gz)

Git bisect gives 7c1d517 "[SystemZ] Enable rematerialization for scalar loads (#179838)" as first bad commit.

I have started looking into it, and it seems that SystemZInstrInfo::foldMemoryOperandImpl() (the version added for reassociable FP operations) are folding a rematerializable load into

%15:fp64bit = nofpexcept MDB %12:fp64bit(tied-def 0), %fixed-stack.2, 0, $noreg

and this is what the verifier complains about. I am not sure yet why this works generally and not in the case of rematerialization. It seems that this method needs to have some kind of check to either reject or fix the register operands if possible.

@anoopkg6 @uweigand 

@dominik-steenken Do we have some test machine running with the verifier?

JonPsson1:
When foldMemoryOperandImpl() is called during register allocation, folding into a reg/mem opcode mustn't be done if the tied def and use operands do not end up referencing the same register.

This patch simply rejects these (it may or not be feasible to try to handle some potential cases), as it turns out this only misses a total of 8 foldings across all of SPEC.

The previous test that was added for enabling folding across BB boundaries now showed up as broken: The tied operands were different and the final assembly output was actually broken (SEB preg0 != preg1)! This then seems like a serious bug and this fix should be backported as well.

Fixes #197414

@anoopkg6

### DIFF
--- a/llvm/lib/Target/SystemZ/SystemZInstrInfo.cpp
+++ b/llvm/lib/Target/SystemZ/SystemZInstrInfo.cpp
@@ -1620,6 +1620,8 @@ MachineInstr *SystemZInstrInfo::foldMemoryOperandImpl(
   if ((RegMemOpcode == SystemZ::SDB || RegMemOpcode == SystemZ::SEB) &&
       FoldAsLoadDefReg != RHS.getReg())
     return nullptr;
+  if (!MRI->isSSA() && DstReg != RegMO.getReg())
+    return nullptr;
 
   MachineOperand &Base = LoadMI.getOperand(1);
   MachineOperand &Disp = LoadMI.getOperand(2);
--- a/llvm/test/CodeGen/SystemZ/foldmem-regalloc.mir
+++ b/llvm/test/CodeGen/SystemZ/foldmem-regalloc.mir
@@ -1,0 +1,123 @@
+# RUN: llc -mtriple=s390x-linux-gnu -mcpu=z17 -start-before=greedy %s -o - \
+# RUN:   -verify-machineinstrs | FileCheck %s
+#
+# Test that folding of a memop does not happen during regalloc in a case that
+# would cause tied registers to be different.
+
+--- |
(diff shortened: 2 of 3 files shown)


theotherjimmy:
I'm not seeing any assertions about the resulting machine code in [foldmem-regalloc.mir](https://github.com/llvm/llvm-project/pull/197475/changes#diff-b76afdd086c7931925d9af92280663c2b0cbb7c5aa74069fc741821e968a3361) Did I miss something or is the test incomplete?

JonPsson1:
> I'm not seeing any assertions about the resulting machine code in [foldmem-regalloc.mir](https://github.com/llvm/llvm-project/pull/197475/changes#diff-b76afdd086c7931925d9af92280663c2b0cbb7c5aa74069fc741821e968a3361) Did I miss something or is the test incomplete?

The idea is to run -machine-verifier: if the illegal folding occurs it will report it. Is this acceptable for a test like this?

theotherjimmy:
I think that should work, yes. Sounds great.

c-rhodes:
this has been open a couple of weeks now and looks like a trivial change, please can we get this reviewed and landed if we want it in

uweigand (approved):
LGTM, thanks!

JonPsson1:
Ah, it looks like the 22.x branched in January already, so the problematic commit(s) are actually not part of the 22 release (first one introduced in February).

Issue #197274, opened 2026-05-12
[BOLT] Incorrect ADRP encoding on TLSDESC -> IE using mold


awshkulkar:
The TLSDESC stub looks like this with corresponding relocs:
```
Slot 1: adrp  x0, :tlsdesc:var        ; R_AARCH64_TLSDESC_ADR_PAGE21
Slot 2: ldr   x1, [x0, :tlsdesc:var]  ; R_AARCH64_TLSDESC_LD64_LO12
Slot 3: add   x0, x0, :tlsdesc:var    ; R_AARCH64_TLSDESC_ADD_LO12
Slot 4: blr   x1                      ; R_AARCH64_TLSDESC_CALL
```
Mold transforms this to IE under right conditions. However it has somewhat similar pattern as TLSDESC->LE we noted in https://github.com/llvm/llvm-project/issues/190366. This is how mold transforms the stub:
```
Slot 1: nop                	;R_AARCH64_TLSDESC_ADR_PAGE21
Slot 2: nop 			;R_AARCH64_TLSDESC_LD64_LO12
Slot 3: adrp x0, :gottprel:    	;R_AARCH64_TLSDESC_ADD_LO12
Slot 4: ldr  x0, [x0, ...] 	;R_AARCH64_TLSDESC_CALL
```
The stale reloc on which the ADRP lands (R_AARCH64_TLSDESC_ADD_LO12) is not correctly handled for this transformation. Both TLSDESC_ADD_LO12 and TLSDESC_ADR_PAGE21 pass through adjustRelocation() to the __BOLT_got_zero fallback path,
  which encodes the ADRP as __BOLT_got_zero + Rel.Value. For lld's stale reloc (TLSDESC_ADR_PAGE21), Rel.Value contains the GOT slot address, so this works. For mold's stale reloc (TLSDESC_ADD_LO12), Rel.Value is the TLS offset (0 for the
  first variable), causing the ADRP to target page 0 instead of the GOT page.

For us to handle this correctly,  we can return nullopt in adjustRelocation() to skip this reloc, but then symbolize it in tryAddingSymbolicOperand() using PC relative value after we handle stale TLS relocs.

Here is a test that shows the issue, synthesized to use lld and skips mold.

```
  # REQUIRES: system-linux
  #
  # Check that BOLT handles TLSDESC -> IE relaxation with stale
  # R_AARCH64_TLSDESC_ADD_LO12 on an ADRP (mold linker layout), while
[...]

PR #197477, opened 2026-05-13
[BOLT] Fix atypical TLSDESC to IE relaxation

awshkulkar:
When mold transforms TLSDESC to IE, its ADRP lands on slot 3 (add) with stale reloc  R_AARCH64_TLSDESC_ADD_LO12. This requires to be handled in Symbolizer, ADRP needs to have right PAGE of GOT slot where IE offset is stored.

LLD does not have this issue since the ADRP lands at R_AARCH64_TLSDESC_ADR_PAGE21 which gets correctly handled with __BOLT_got_zero symbol by happenstance since Rel.Value for this reloc  is actually GOT slot address.

A TLSDESC stub looks like this:
  adrp  x0, :tlsdesc:var        ; R_AARCH64_TLSDESC_ADR_PAGE21
  ldr   x1, [x0, :tlsdesc:var]  ; R_AARCH64_TLSDESC_LD64_LO12
  add   x0, x0, :tlsdesc:var    ; R_AARCH64_TLSDESC_ADD_LO12
  blr   x1                      ; R_AARCH64_TLSDESC_CALL

LLD creates TLSDESC to IE relaxation as
  Slot 1: TLSDESC_ADR_PAGE21  → ADRP x0, :gottprel:
  Slot 2: TLSDESC_LD64_LO12   → LDR  x0, [x0, :gottprel_lo12:]
  Slot 3: TLSDESC_ADD_LO12    → NOP
  Slot 4: TLSDESC_CALL         → NOP

But mold does it as:
  Slot 1: TLSDESC_ADR_PAGE21  → NOP
  Slot 2: TLSDESC_LD64_LO12   → NOP
  Slot 3: TLSDESC_ADD_LO12    → ADRP x0, :gottprel:  <-- Crash
  Slot 4: TLSDESC_CALL         → LDR  x0, [x0, ...]

Fix by returning nullopt from adjustRelocation() for TLSDESC_ADD_LO12 on ADRP, then symbolizing the ADRP in tryAddingSymbolicOperand() to have the right PC-relative offset symbol (+offset) of the GOT slot.

Refer to https://github.com/llvm/llvm-project/issues/197274 for details.

### DIFF
--- a/bolt/lib/Target/AArch64/AArch64MCSymbolizer.cpp
+++ b/bolt/lib/Target/AArch64/AArch64MCSymbolizer.cpp
@@ -52,6 +52,13 @@ bool AArch64MCSymbolizer::tryAddingSymbolicOperand(
       return true;
     }
 
+    if (BC.MIB->isADRP(Inst)) {
+      uint64_t TargetPage = (InstAddress & ~0xFFFULL) + (Value << 12);
+      auto [Sym, Offset] = BC.handleAddressRef(TargetPage, Function, true);
+      addOperand(Sym, Offset, 0);
+      return true;
+    }
+
     LLVM_DEBUG(dbgs() << "BOLT-DEBUG: ignoring relocation at 0x"
                       << Twine::utohexstr(InstAddress) << '\n');
   }
@@ -111,6 +118,11 @@ AArch64MCSymbolizer::adjustRelocation(const Relocation &Rel,
       return std::nullopt;
     }
   }
+
+  // Handle TLSDESC -> IE relaxation where the linker places ADRP at the
+  // ADD slot (e.g. mold). The TLSDESC_ADD_LO12 relocation is stale.
+  if (BC.MIB->isADRP(Inst) && Rel.Type == ELF::R_AARCH64_TLSDESC_ADD_LO12)
+    return std::nullopt;
 
   if (!Relocation::isGOT(Rel.Type))
     return Rel;
--- a/bolt/test/AArch64/tls-desc-ie-relaxation.s
+++ b/bolt/test/AArch64/tls-desc-ie-relaxation.s
@@ -1,0 +1,49 @@
+# REQUIRES: system-linux
+#
+# Check that BOLT handles TLSDESC -> IE relaxation with stale
+# R_AARCH64_TLSDESC_ADD_LO12 on an ADRP (mold linker layout), while
+# not regressing the normal GOTTPREL ADRP handling (lld layout).
+#
+# Build with correct GOTTPREL relocations, then swap the first
+# GOTTPREL_PAGE21 to TLSDESC_ADD_LO12 to simulate mold's stale reloc.
+#
+# RUN: llvm-mc -filetype=obj -triple aarch64-unknown-linux %s -o %t.o
+# RUN: ld.lld --emit-relocs -shared %t.o -o %t.so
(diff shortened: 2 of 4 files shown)


yavtuk:
@ilinpv lets see how to do it better here

ilinpv:
BOLT workaround looks reasonable to me, but ideally mold should not emit misleading relocation metadata in the first place.
R_AARCH64_TLSDESC_ADD_LO12 appears to be specifically an ADD-immediate relocation, while an ADRP instruction should use a page-based relocation such as R_AARCH64_TLSDESC_ADR_PAGE21:
https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst
So if the relaxed sequence rewrites the instruction into an ADRP, keeping R_AARCH64_TLSDESC_ADD_LO12 attached to it looks semantically questionable and can confuse post-link tools.
Drawing @rui314 attention here for the **mold** perspective.

awshkulkar:
> BOLT workaround looks reasonable to me, but ideally mold should not emit misleading relocation metadata in the first place. R_AARCH64_TLSDESC_ADD_LO12 appears to be specifically an ADD-immediate relocation, while an ADRP instruction should use a page-based relocation such as R_AARCH64_TLSDESC_ADR_PAGE21: https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst So if the relaxed sequence rewrites the instruction into an ADRP, keeping R_AARCH64_TLSDESC_ADD_LO12 attached to it looks semantically questionable and can confuse post-link tools. Drawing @rui314 attention here for the **mold** perspective.

I think problem here is that (irrespective of linker) emit-relocs is always going to have older TLS relocs that were emitted by the compiler/assembler for TLSDESC. Since there is no strict requirement in ABI on how the transformed layout should look like i.e. where nop and new instruction are placed in a 4 wide instruction space, we have this mismatched gap.  As I see both mold and LLD are emitting code that is correct at runtime. So we can either force everyone to have same style of transformation through AArch64 ELF ABI or we should be flexible to account all possible stale relocs that can happen in this transformation.

yavtuk:
> > BOLT workaround looks reasonable to me, but ideally mold should not emit misleading relocation metadata in the first place. R_AARCH64_TLSDESC_ADD_LO12 appears to be specifically an ADD-immediate relocation, while an ADRP instruction should use a page-based relocation such as R_AARCH64_TLSDESC_ADR_PAGE21: https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst So if the relaxed sequence rewrites the instruction into an ADRP, keeping R_AARCH64_TLSDESC_ADD_LO12 attached to it looks semantically questionable and can confuse post-link tools. Drawing @rui314 attention here for the **mold** perspective.
> 
> I think problem here is that (irrespective of linker) emit-relocs is always going to have older TLS relocs that were emitted by the compiler/assembler for TLSDESC. Since there is no strict requirement in ABI on how the transformed layout should look like i.e. where nop and new instruction are placed in a 4 wide instruction space, we have this mismatched gap. As I see both mold and LLD are emitting code that is correct at runtime. So we can either force everyone to have same style of transformation through AArch64 ELF ABI or we should be flexible to account all possible stale relocs that can happen in this transformation.

flexibility related to parse adrp+add+nop+nop or nop+nop+adrp+add can be, but here in the patch we check that adrp instruction has Rel.Type == ELF::R_AARCH64_TLSDESC_ADD_LO12 relocation. From time to time, bugs related to relocation errors occur. If we take this workaround, how do we verify that it’s not a bug?! From my point of view it easy to add workaround to bolt and handle it but the right fix should be on the mold linker side.

PR #197478, opened 2026-05-13
[DirectX] Add an "offset" operand to llvm.dbg.value

hvdijk:
Offset operand was removed in abe04759a6, so we need to bring it back for DXIL. If offset is not specified, it should be zero.


bogner:
As far as I can tell we're always setting the offset to zero. In that case can't we just inject the extra constant zero operand late when emitting DXIL? That would require a lot less machinery if its possible, but maybe I'm missing something...

hvdijk:
That's possible, but late changes when emitting DXIL is what I've been trying to avoid. Firstly it's error-prone, as when `DXILValueEnumerator` and `DXILBitcodeWriter` are not kept exactly in sync wrt the order in which they enumerate, things almost work, things work well enough to pass simple tests, and then on real code it starts to fail in ways that we're unable to detect, that we're unable to add assertions for. Secondly it's only possible to add tests for that with `dxil-dis`, which is not available by default even when the `DirectX` target is enabled. This way allows for a non-`dxil-dis` to be added after #198318.

Edit: The other alternative that I considered was, instead of adding `VRename` and `VRemove` to `DXILDebugInfo`, adding `VReplace`, and using that to create an out-of-module declaration of `llvm.dbg.value` and out-of-module call to it. But unlike `MDReplace`, that causes lifetime management issues, we'd need to make sure they get cleaned up properly, and I hadn't figured out yet a way to get that working. Do you think that'd be a better approach than what is currently in this PR?

bogner:
Fair enough, the testing argument probably justifies this.

My biggest concern with this approach is how many places need to be aware of VRemove. The VReplace idea would help with that if it works I suppose, so it may be worth investigating. Otherwise it’s probably okay, I just worry that this is fragile in its own way.

hvdijk:
> My biggest concern with this approach is how many places need to be aware of VRemove. The VReplace idea would help with that if it works I suppose, so it may be worth investigating.

It wouldn't help directly, every place that currently checks VRemove would need to check VReplace, but it does still have the potential to simplify things by avoiding the VRename logic, so I'll give it a go.

Avoiding needing to write VRemove (or VReplace) in so many places would be possible if we use some custom iterator type that handles VRemove/VReplace transparently, I'll have a look at that next, but that might possibly add more code than it saves, not sure yet.

hvdijk:
> It wouldn't help directly, every place that currently checks VRemove would need to check VReplace, but it does still have the potential to simplify things by avoiding the VRename logic, so I'll give it a go.

Since we need to manually clean up the functions and instructions that are not attached to a module, and the clean up needs to be different for functions and instructions, it turned out easier to split this into `FRename` and `IRename`, but done now.

llvm/lib/Target/DirectX/DirectXIRPasses/DXILDebugInfo.h
```diff
@@ -14,23 +14,60 @@
...
 
+  DXILDebugInfoMap() = default;
+  DXILDebugInfoMap(const DXILDebugInfoMap &) = delete;
+  DXILDebugInfoMap(DXILDebugInfoMap &&) = default;
+
+  /// Completely replace one function with another in ValueEnumerator.
+  FMap FReplace;
+
+  /// Completely replace one instruction with another in ValueEnumerator.
+  IMap IReplace;
```
bogner: Let's use `Func` and `Inst` here - `MD` is a common acronym but I could see `F` and `I` feeling ambiguous. So, `FuncMap`, `FuncReplace`, `InstMap`, InstReplace`.

bogner (approved):
Yeah I like this better. A couple of iterators might be nice too, but I think it probably isn't a big savings. One comment on some naming but this LGTM.

llvm/lib/Target/DirectX/DirectXIRPasses/DXILDebugInfo.h
hvdijk: Sure, done.

PR #197480, opened 2026-05-13
[NFC] Add missing bfloat test coverage for apfloat conversions

MrSidims:


llvm/test/CodeGen/AMDGPU/arbitrary-fp-to-float.ll
```diff
@@ -1231,3 +1233,168 @@ define <2 x half> @from_f8e4m3fn_v2f16(<2 x i8> %x) {
   %r = call <2 x half> @llvm.convert.from.arbitrary.fp.v2f16.v2i8(<2 x i8> %x, metadata !"Float8E4M3FN")
   ret <2 x half> %r
 }
+
+define bfloat @from_f8e5m2_bf16(i8 %x) {
```
arsenm, line 1237: Does the verifier enforce only i8 for this format?

arsenm (approved):

llvm/test/CodeGen/AMDGPU/arbitrary-fp-to-float.ll
MrSidims: Actually, it doesn't. And it's verifier's gap. Added to my TODO list (I assume it's not a blocker for the ongoing PRs).

PR #197481, opened 2026-05-13
[ConstantRange] Optimize multiply with nowrap

nikic:
multiplyWithNoWrap() may currently call all of multiply(), umul_sat() and smul_sat(), where the former may perform up to 6 double-width multiplies, and the other two perform 2 and 4 single-width multiplies respectively.

Optimize this a bit by moving the nowrap handling directly into multiply(): If we're already doing double-width multiplies, then doing more on top of that is unnecessary.

Additionally, this also allows us to use only single-width multiplies for the nuw case. (This is also possible for the nsw case, but the implementation would be more involved.)

### DIFF
--- a/llvm/include/llvm/IR/ConstantRange.h
+++ b/llvm/include/llvm/IR/ConstantRange.h
@@ -442,18 +442,11 @@ public:
                 PreferredRangeType RangeType = Smallest) const;
 
   /// Return a new range representing the possible values resulting
-  /// from a multiplication of a value in this range and a value in \p Other,
-  /// treating both this and \p Other as unsigned ranges.
-  LLVM_ABI ConstantRange multiply(const ConstantRange &Other) const;
-
-  /// Return a new range representing the possible values resulting
-  /// from a multiplication with wrap type \p NoWrapKind of a value in this
-  /// range and a value in \p Other.
-  /// If the result range is disjoint, the preferred range is determined by the
-  /// \p PreferredRangeType.
-  LLVM_ABI ConstantRange
-  multiplyWithNoWrap(const ConstantRange &Other, unsigned NoWrapKind,
-                     PreferredRangeType RangeType = Smallest) const;
+  /// from a multiplication of a value in this range and a value in \p Other.
+  /// If \p NoWrapKind is set, assume that corresponding wrapping can not
+  /// occur.
+  LLVM_ABI ConstantRange multiply(const ConstantRange &Other,
+                                  unsigned NoWrapKind = 0) const;
 
   /// Return range of possible values for a signed multiplication of this and
   /// \p Other. However, if overflow is possible always return a full range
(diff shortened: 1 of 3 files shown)


antoniofrighetto (approved):
LGTM, thanks!

llvm/include/llvm/IR/ConstantRange.h
```diff
@@ -442,18 +442,11 @@ class [[nodiscard]] ConstantRange {
...
-  /// If the result range is disjoint, the preferred range is determined by the
-  /// \p PreferredRangeType.
-  LLVM_ABI ConstantRange
-  multiplyWithNoWrap(const ConstantRange &Other, unsigned NoWrapKind,
-                     PreferredRangeType RangeType = Smallest) const;
+  /// from a multiplication of a value in this range and a value in \p Other.
+  /// If \p NoWrapKind is set, assume that corresponding wrapping can not
+  /// occur.
+  LLVM_ABI ConstantRange multiply(const ConstantRange &Other,
+                                  unsigned NoWrapKind = 0) const;
```
bababuck, line 449: Nit, but don't think it's worth refactoring the header files to make this feasible
```
                                  unsigned NoWrapKind = OBO::AnyWrap) const;
```

llvm/lib/IR/ConstantRange.cpp
```diff
@@ -1260,36 +1273,23 @@ ConstantRange::multiply(const ConstantRange &Other) const {
...
-
-  if (NoWrapKind & OverflowingBinaryOperator::NoUnsignedWrap)
-    Result = Result.intersectWith(umul_sat(Other), RangeType);
+  if (NoWrapKind & OverflowingBinaryOperator::NoSignedWrap) {
+    Result_sext = Result_sext.intersectWith(
+        ConstantRange(APInt::getSignedMinValue(BW).sext(BW * 2),
+                      APInt::getSignedMaxValue(BW).sext(BW * 2) + 1));
+  }
+  ConstantRange SR = Result_sext.truncate(BW);
+  ConstantRange Result = UR.isSizeStrictlySmallerThan(SR) ? UR : SR;
```
bababuck, line 1292: For my understanding, why do we not use `UR.intersectWith(SR, Smallest)`? Since both ranges are conservative, wouldn't the intersection be as well? Or is this just not done for compile time concerns?

nikic: I've just retained this from the previous implementation, but yes, using intersectWith() is probably cleaner.

bababuck: I can push a separate MR for that once this merges.

llvm/lib/IR/ConstantRange.cpp
bababuck, line 1223: Should this fast path only be applied when no flags are present? Would lose precision in cases like {-1} * [0, 5) <nuw>.

AI use: Found by AI, double checked by me

nikic: Hm, I guess we could add a special case to produce a [-1,0] range for nuw? This seems like an unlikely/artificial case though. The multiply() ranges are generally not optimal anyway, so I'm not sure it's worthwhile trying to be optimal in this specific case.

bababuck (approved):
LGTM

PR #197484, opened 2026-05-13
[SPIR-V] Adjust SPV_AMD_weak_linkage

MrSidims:
Linkage was renamed + a capability added following review in https://github.com/KhronosGroup/SPIRV-Registry/pull/401


llvm/lib/Target/SPIRV/SPIRVUtils.cpp
aobolensk, line 1237: I'm curious if `LinkOnceODR` should be renamed to `LinkOnceODRKHR` following the same logic...

MrSidims: We've discussed it on khronos teleconference, it won't be renamed and it won't have a capability either :)

jmmartinez (approved):
It looks good to me.

llvm/lib/Target/SPIRV/SPIRVUtils.cpp
aobolensk: OK, let it be this way, haha

aobolensk (approved):

maarquitos14 (approved):

idubinov (approved):

PR #197485, opened 2026-05-13
[AIX][clang][compiler_rt] rename libatomic archive to libclang_rt

azhan92:
This PR implements the following on AIX to avoid conflicts between LLVM libatomic and the GNU libatomic in the AIX toolbox as they share the same library name:

- Updates the clang driver to use `-lcompiler_rt` instead of `-latomic`
- Renames the compiler-rt archive from `libatomic.a` to `libcompiler_rt.a`

Only the archive and not the shared object (`libatomic.so.1`) is renamed because renaming one component is enough to distinguish between the LLVM and GNU libatomic libraries. This also allows us to add additional shared objects to the `libcompiler_rt.a` archive in the future if needed.

### DIFF
--- a/clang/docs/ReleaseNotes.rst
+++ b/clang/docs/ReleaseNotes.rst
@@ -715,7 +715,8 @@ AIX Support
   (instead of `all`) which only extracts static init from archive members which
   would otherwise be referenced.
   (See https://www.ibm.com/docs/en/aix/7.2.0?topic=l-ld-command for details).
-- The driver now uses ``-lcompiler_rt`` instead of ``-latomic`` to avoid conflicts
+- The driver now uses ``-lcompiler_rt`` instead of ``-latomic``, and the compiler-rt
+  archive has been renamed from ``libatomic.a`` to ``libcompiler_rt.a`` to avoid conflicts
   between the LLVM libatomic and the GNU libatomic from the AIX toolbox as they share
   the same library name.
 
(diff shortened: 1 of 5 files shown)


daltenty:
A bunch of high level nits here:
- the description for any PR should explain the motivation for the change
- let's rebase on a clean commit from `main`, there seems to be some noise in the tests
- I think we should update the name in compiler-rt builtins `CMakeLists.txt` at the same time for simplicity. (If I understand the code right, the archive name is just the second parameter to the call to `archive_aix_libatomic`. We are not updating the shared object name here,  just the archive)
- let's update the release notes for clang too to reflect the change

efriedma-quic (commented):
You probably should not install both GNU libatomic and clang_rt.atomic on the same system.  The ABI requires that all code in a process uses a single atomic library; if you have multiple atomic libraries, synchronization won't work correctly.

daltenty:
> You probably should not install both GNU libatomic and clang_rt.atomic on the same system.

I tend agree with you, but unfortunately there is a complex history on AIX, which has led to us having both clang_rt.atomic and libatomic installed on many systems (`clang_rt.atomic` is vendored by the OS in `/usr/lib` and GNU libatomic is distributed by the AIX toolbox in `/opt/freeware/lib`), with the same archive and shared library names, but different ABIs (GCC libatomic exports additional symbols). 

This PR specifically tried to address load-failures that occur when we have both GCC built applications and clang-built applications which need to share the same load-path (i.e. `/opt/freeware/lib`)

>  The ABI requires that all code in a process uses a single atomic library; if you have multiple atomic libraries, synchronization won't work correctly.

You are completely correct, an single process should not be using both libraries, as the global state of locks won't be observed across implementations if a user did somehow pass the same data to both implementations. So having both may slightly increase the risk of users erroneously including both in a program, but from some non-exhaustive surveys we found no shared libraries with problematic exposures for us (applications don't really run into this directly, since then end up linking one or the other), so we feel it's more important to fix the unfortunate coexistence scenario.

daltenty (approved):
LGTM, thanks!

PR #164093, opened 2025-10-18
[libc][stdlib][annex_k] Add set_constraint_handler_s.

bassiounix:
RFC https://discourse.llvm.org/t/rfc-bounds-checking-interfaces-for-llvm-libc/87685

Add `set_constraint_handler_s` required by Annex K interface in LLVM libc.

PR #197487, opened 2026-05-13
[libc] Annex K: Add constraint handler unit test class

vhscampos:
This unit test class will be useful for the tests related to Annex K. The functions in Annex K may call a constraint handler, so this new unit test class will facilitate the checks that the constraint handling mechanism is working as expected.


vhscampos:
Requires https://github.com/llvm/llvm-project/pull/164093 for `set_constraint_handler_s`

PR #197490, opened 2026-05-13
[DirectX] Drop debug labels

hvdijk:
Debug labels did not exist in LLVM 3.7 and have no equivalent.


llvm/test/tools/dxil-dis/di-label.ll
```diff
@@ -0,0 +1,29 @@
+; RUN: llc --filetype=obj %s -o - | dxil-dis -o - | FileCheck %s
+
+target triple = "dxil-pc-shadermodel6.3-library"
+
+; The label gets removed, so just check that the function is still emitted.
+; CHECK: define void @foo()
```
bogner: Should we add another debug instruction to check for just to ensure that reprocessing the module works correctly?

bogner (approved):

hvdijk:
Something has gone wrong with this PR, let me try to restore this.

llvm/test/tools/dxil-dis/di-label.ll
hvdijk: Sure, that seems fine, I'll update the test.

PR #197479, opened 2026-05-13
Reland "[libc][NFC] Propagate LIBC_CONSTEXPR qualifier for those depending on bit_cast."

bassiounix:

PR #197492, opened 2026-05-13
Revert "Reland "[libc][NFC] Propagate LIBC_CONSTEXPR qualifier for those depending on bit_cast.""

rupprecht:
Reverts llvm/llvm-project#197479 (c966c50a7eb9246e0ec794c784743528e93d5142)

The reland includes some other changes that cause CI to fail with various `-Winvalid-constexpr` failures


shiltian (approved):
Yeah, we need to revert this since it breaks the build.

PR #197494, opened 2026-05-13
[VPlan] Add helper to define CastInfo for casting to VP mixins (NFC)

fhahn:
Add new generic VPMixinCast that can be used to define CastInfo from a base class (like VPRecipeBase) to a mixin class (like VPIRMetadata) through a list of supported recipe types.

Overall that makes defining the cast machinery quite a bit more compact. It also adds a static_assert to prevent mis-use using std::is_base_of_v


llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -4044,34 +4044,32 @@ class LLVM_ABI_FOR_TEST VPScalarIVStepsRecipe : public VPRecipeWithIRFlags {
 #endif
 };
 
+/// CastInfo helper for casting from a recipe superclass (e.g. VPRecipeBase,
+/// VPUser) to a mixin class that is not part of the VPRecipeBase class
+/// hierarchy (e.g. VPPhiAccessors, VPIRMetadata).
+namespace vpdetail {
+template <class VPMixin, class From, class... RecipeTys>
```
artagnon:
```suggestion
template <typename VPMixin, typename From, typename... RecipeTys>
```

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -4044,34 +4044,32 @@ class LLVM_ABI_FOR_TEST VPScalarIVStepsRecipe : public VPRecipeWithIRFlags {
...
+struct VPMixinCast
+    : public DefaultDoCastIfPossible<VPMixin *, From *,
```
artagnon: I don't think we require the From, and can just hard-code VPRecipeBase?

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -4044,34 +4044,32 @@ class LLVM_ABI_FOR_TEST VPScalarIVStepsRecipe : public VPRecipeWithIRFlags {
...
+                                     VPMixinCast<VPMixin, From, RecipeTys...>> {
+  static_assert((std::is_base_of_v<VPMixin, RecipeTys> && ...),
+                "Each type in RecipeTys must derive from VPMixin");
+  static bool isPossible(From *R) { return isa<RecipeTys...>(R); }
+  static VPMixin *doCast(From *R) {
+    VPMixin *Out = nullptr;
+    (void)((Out = dyn_cast<RecipeTys>(R)) || ...);
```
artagnon: doCast should not use dyn_cast -- that would be doCastIfPossible?

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -4044,34 +4044,32 @@ class LLVM_ABI_FOR_TEST VPScalarIVStepsRecipe : public VPRecipeWithIRFlags {
 #endif
 };
 
+/// CastInfo helper for casting from a recipe superclass (e.g. VPRecipeBase,
+/// VPUser) to a mixin class that is not part of the VPRecipeBase class
+/// hierarchy (e.g. VPPhiAccessors, VPIRMetadata).
+namespace vpdetail {
+template <class VPMixin, class From, class... RecipeTys>
+struct VPMixinCast
```
eas: Maybe `MixinCastInfoImpl` would be better? Or `CastInfoMixinImpl`.

llvm/lib/Transforms/Vectorize/VPlan.h
fhahn: Yep, replaced dyn_cast+ assert by cast, thanks

llvm/lib/Transforms/Vectorize/VPlan.h
fhahn: Removed here for now, I think a follow-up may need to specialize for other types, but can add once neede,d thanks

llvm/lib/Transforms/Vectorize/VPlan.h
fhahn: done thanks

llvm/lib/Transforms/Vectorize/VPlan.h
fhahn: Updated, thanks!

llvm/lib/Transforms/Vectorize/VPlan.h
artagnon: Hm, forgot to push a change?

fhahn: Ah yes sorry, the dyn_cast + assert are actually needed I think, as we need to check until a cast to a recipe succeeds. At least one must succeed, restored the assert

eas (approved):
LGTM but give @artagnon time to have another look.

llvm/lib/Transforms/Vectorize/VPlan.h
artagnon: Hm, I'm a bit worried about using dyn_cast here, as it would call doCastIfPossible, which in turn would call doCast again? I'm not sure why this doesn't run into a stack overflow? I don't think we really have a choice here, but to switch over VPRecipeID and call cast? Wouldn't isPossible ensure that the cast is always possible?

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -4044,34 +4044,32 @@ class LLVM_ABI_FOR_TEST VPScalarIVStepsRecipe : public VPRecipeWithIRFlags {
...
+  static VPMixin *castFailed() { return nullptr; }
+};
+} // namespace vpdetail
+
 /// Support casting from VPRecipeBase -> VPPhiAccessors.
 template <>
 struct CastInfo<VPPhiAccessors, VPRecipeBase *>
-    : DefaultDoCastIfPossible<VPPhiAccessors *, VPRecipeBase *,
-                              CastInfo<VPPhiAccessors, VPRecipeBase *>> {
-  /// Used by isa.
```
artagnon, line 4052: Could we add back "Used by" isa/cast/dyn_cast comments into the new code?

llvm/lib/Transforms/Vectorize/VPlan.h
fhahn: The `dyn_casts` are from `VPRecipeBase*` to concrete recipe types, which go through a different path (recipe `classof` implementations)

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -4044,34 +4044,32 @@ class LLVM_ABI_FOR_TEST VPScalarIVStepsRecipe : public VPRecipeWithIRFlags {
...
+template <typename VPMixin, typename... RecipeTys>
+struct CastInfoMixinImpl
+    : public DefaultDoCastIfPossible<VPMixin *, VPRecipeBase *,
+                                     CastInfoMixinImpl<VPMixin, RecipeTys...>> {
+  static_assert((std::is_base_of_v<VPMixin, RecipeTys> && ...),
+                "Each type in RecipeTys must derive from VPMixin");
+  static bool isPossible(VPRecipeBase *R) { return isa<RecipeTys...>(R); }
+  static VPMixin *doCast(VPRecipeBase *R) {
+    VPMixin *Out = nullptr;
+    (void)((Out = dyn_cast<RecipeTys>(R)) || ...);
```
artagnon:
```suggestion
    ((Out = dyn_cast<RecipeTys>(R)) || ...);
```

I don't think it's necessary?

artagnon (approved):
LGTM, thanks!

llvm/lib/Transforms/Vectorize/VPlan.h
fhahn: Removed thanks

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -4044,34 +4044,36 @@ class LLVM_ABI_FOR_TEST VPScalarIVStepsRecipe : public VPRecipeWithIRFlags {
...
+    return Out;
   }
-
-  /// Used by inherited doCastIfPossible to dyn_cast.
-  static inline VPPhiAccessors *castFailed() { return nullptr; }
+  static VPMixin *castFailed() { return nullptr; }
 };
+} // namespace vpdetail
+
+/// Support casting from VPRecipeBase -> VPPhiAccessors.
```
ayalz, line 4072:
```suggestion
/// Support casting from VPRecipeBase -> VPPhiAccessors.
// TODO: include VPPredInstPHIRecipe too, once it implements VPPhiAccessors.
```
nit: retain note.

ayalz (commented):
Post-commit nit.

PR #193438, opened 2026-04-22
[CodeGen] Precompute untyped getMinimalPhysRegClass

c-rhodes:
Profiling sqlite on aarch64 O0-g shows the callee-save paths in
AArch64FrameLowering which call getMinimalPhysRegClass are hot. Precomputing
this in tablegen for the common untyped path is a -0.30% geomean improvement on
stage1-aarch64-O0-g CTMark -0.30%. It also improves stage1-aarch64-O3 by
-0.20%.

https://llvm-compile-time-tracker.com/compare.php?from=4e6d3722fca73c97367720180a8d547057fda380&to=f6a6ad834e943bdc5563cd4f8374a22014cd03b2&stat=instructions%3Au

PR #183150, opened 2026-02-24
[amdgpu-cfi: 7/9]: [AMDGPU] Implement CFI for CSR spills

slinder1:
Introduce new SPILL pseudos to allow CFI to be generated for only CSR
spills, and to make ISA-instruction-level accurate information.

Other targets either generate slightly incorrect information or rely on
conventions for how spills are placed within the entry block. The
approach in this change produces larger unwind tables, with the
increased size being spent on additional DW_CFA_advance_location
instructions needed to describe the unwinding accurately.

Change-Id: I9b09646abd2ac4e56eddf5e9aeca1a5bebbd43dd
Co-authored-by: Scott Linder <scott.linder@amd.com>
Co-authored-by: Venkata Ramanaiah Nalamothu <VenkataRamanaiah.Nalamothu@amd.com>

---

**Stack**:
- #183147
- #183149
- #183150⬅
- #183146
- #183153
- #183152
- #183148
- #183151
- #183171
- `main`

<sub>(Note: Closed and merged PRs may not be reflected here and PR numbering is not stable.)</sub>

PR #197495, opened 2026-05-13
[CodeGen] De-type getMinimalPhysRegClass and related APIs

c-rhodes:
Follow-up #193438 to de-type getMinimalPhysRegClass so it can be replaced with
the precomputed getDefaultMinimalPhysRegClass. Type is also removed from
related APIs.

There's very few uses of getMinimalPhysRegClass with a non-default type
(MVT::Other) and none at all for getCommonMinimalPhysRegClass. Rather than
trying to also handle the type when precomputing at compile-time it seems
better to remove the type altogether and simplify the APIs. This also improves
compile-time for a number of targets / configurations:

CTMark geomean:
- stage1-O3: -0.23%
- stage1-ReleaseThinLTO: -0.19%
- stage1-ReleaseLTO-g: -0.15%
- stage1-aarch64-O3: -0.57%
- stage2-O3: -0.23%
- clang: -0.13%

https://llvm-compile-time-tracker.com/compare.php?from=eae0b6b2498305ee29dc85a405ede9ccdc10ce7d&to=08c052a2db407d2a21d468001fd2035d3720acf7&stat=instructions%3Au

Assisted-by: codex

### DIFF
--- a/llvm/include/llvm/CodeGen/TargetRegisterInfo.h
+++ b/llvm/include/llvm/CodeGen/TargetRegisterInfo.h
@@ -353,13 +353,13 @@ public:
     return I;
   }
 
-  /// Returns the Register Class of a physical register, picking the most sub
-  /// register class that contains this physreg.
+  /// Returns the Register Class of a physical register, picking the smallest
+  /// register subclass that contains this physreg.
   virtual const TargetRegisterClass *
   getMinimalPhysRegClass(MCRegister Reg) const = 0;
 
   /// Returns the common Register Class of two physical registers, picking the
-  /// most sub register class that contains these two physregs.
+  /// smallest register subclass that contains these two physregs.
   const TargetRegisterClass *
   getCommonMinimalPhysRegClass(MCRegister Reg1, MCRegister Reg2) const;
 
--- a/llvm/include/llvm/CodeGen/TargetRegisterInfo.h
+++ b/llvm/include/llvm/CodeGen/TargetRegisterInfo.h
@@ -353,33 +353,15 @@ public:
     return I;
   }
 
-  /// Returns the Register Class of a physical register of the given type,
-  /// picking the most sub register class of the right type that contains this
-  /// physreg.
-  const TargetRegisterClass *getMinimalPhysRegClass(MCRegister Reg,
-                                                    MVT VT = MVT::Other) const;
+  /// Returns the Register Class of a physical register, picking the most sub
+  /// register class that contains this physreg.
+  virtual const TargetRegisterClass *
+  getMinimalPhysRegClass(MCRegister Reg) const = 0;
 
-  /// Returns the common Register Class of two physical registers of the given
-  /// type, picking the most sub register class of the right type that contains
-  /// these two physregs.
+  /// Returns the common Register Class of two physical registers, picking the
+  /// most sub register class that contains these two physregs.
   const TargetRegisterClass *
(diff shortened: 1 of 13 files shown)


llvm/lib/Target/AMDGPU/SILowerSGPRSpills.cpp
```diff
@@ -146,8 +146,7 @@ static void insertCSRSaves(MachineBasicBlock &SaveBlock,
       MCRegister Reg = CS.getReg();
 
       MachineInstrSpan MIS(I, &SaveBlock);
-      const TargetRegisterClass *RC = RI->getMinimalPhysRegClass(
-          Reg, Reg == RI->getReturnAddressReg(MF) ? MVT::i64 : MVT::i32);
+      const TargetRegisterClass *RC = RI->getMinimalPhysRegClass(Reg);
```
jayfoad: So these target changes are NFC and the code was needlessly complicated? Maybe commit them separately?

llvm/include/llvm/CodeGen/TargetRegisterInfo.h
```diff
@@ -353,33 +353,15 @@ class LLVM_ABI TargetRegisterInfo : public MCRegisterInfo {
...
-  /// returns nullptr.
-  const TargetRegisterClass *getMinimalPhysRegClassLLT(MCRegister Reg,
-                                                       LLT Ty = LLT()) const;
-
-  /// Returns the common Register Class of two physical registers of the given
-  /// type, picking the most sub register class of the right type that contains
-  /// these two physregs. If there is no register class compatible with the
-  /// given type, returns nullptr.
+  /// Returns the Register Class of a physical register, picking the most sub
+  /// register class that contains this physreg.
```
arsenm:
```suggestion
  /// Returns the Register Class of a physical register, picking the smallest register subclass that contains this physreg.
```

"most sub register class" reads badly and doesn't help with the register subclass vs. subregister class confusion

llvm/lib/Target/AMDGPU/SILowerSGPRSpills.cpp
c-rhodes: insofar as the tests pass, yes. Happy to commit separately if that's preferred.

RKSimon (approved):
LGTM - might be worth a release note entry for downstream maintainers

c-rhodes:
rebase picked up #183150 which re-added a typed call after I already de-typed AMDGPU backend. I'm not gonna post another separate PR to fix that again will just update it here

llvm/lib/CodeGen/MachineVerifier.cpp
```diff
@@ -99,6 +99,16 @@ namespace {
 /// at one time.
 static ManagedStatic<sys::SmartMutex<true>> ReportedErrorsLock;
 
+static bool hasPhysRegClassForType(const TargetRegisterInfo &TRI,
+                                   MCRegister Reg, LLT Ty) {
+  assert(Reg.isPhysical() && "reg must be a physical register");
+  assert(Ty.isValid() && "expected a valid type");
+
+  return llvm::any_of(TRI.regclasses(), [&](const TargetRegisterClass *RC) {
```
sdesmalen-arm, line 111: nit: is it worth trying the regclass resulting from `TRI->getMinimalPhysRegClass(Reg)` first before iterating over all over reg classes?

sdesmalen-arm (approved):

PR #197496, opened 2026-05-13
[VPlan] Make VPWidenLoad(EVL)Recipe VPSingleDefs (NFC).

fhahn:
This updates VPWidenMemoryRecipe to no longer inherit directly from
VPRecipeBase. Instead VPWidenMemoryRecipe is turned into a mixin, like
VPIRMetadata and VPPhiAccessors. This in turn allows updating
VPWidenLoad(EVL)Recipe to inherit directly from VPSingleDefRecipe. This
brings them in line with all other recipes defining a single VPValue.

Depends on https://github.com/llvm/llvm-project/pull/197494

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
+++ b/llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
@@ -3555,10 +3555,11 @@ void VPlanTransforms::createInterleaveGroups(
   for (VPBasicBlock *VPBB :
        VPBlockUtils::blocksOnly<VPBasicBlock>(vp_depth_first_shallow(
            Plan.getVectorLoopRegion()->getEntryBasicBlock())))
-    for (VPRecipeBase &R :
-         make_filter_range(*VPBB, IsaPred<VPWidenMemoryRecipe>)) {
-      auto &MemR = cast<VPWidenMemoryRecipe>(R);
-      IRMemberToRecipe[&MemR.getIngredient()] = &MemR;
+    for (VPRecipeBase &R : make_filter_range(*VPBB, [](VPRecipeBase &R) {
+           return isa<VPWidenMemoryRecipe>(&R);
+         })) {
+      auto *MemR = cast<VPWidenMemoryRecipe>(&R);
+      IRMemberToRecipe[&MemR->getIngredient()] = MemR;
     }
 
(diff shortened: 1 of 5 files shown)


llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
lukel97, line 3554: Can we just store `DenseMap<Instruction *, VPSingleDefRecipe *>` to avoid the calls to `getAsRecipe`?

fhahn: We could, but there are a number of places below that access members of `VPWidenMemoryRecipe`, so I left it as is for now. We could add anothter `CastInfo` to enable casting from VPWidenMemoryRecipe other recipe sub-classes, but not sure removing a few getAsRecipe calls is worth the extra complexity

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -3555,10 +3555,11 @@ void VPlanTransforms::createInterleaveGroups(
...
            Plan.getVectorLoopRegion()->getEntryBasicBlock())))
-    for (VPRecipeBase &R :
-         make_filter_range(*VPBB, IsaPred<VPWidenMemoryRecipe>)) {
-      auto &MemR = cast<VPWidenMemoryRecipe>(R);
-      IRMemberToRecipe[&MemR.getIngredient()] = &MemR;
+    for (VPRecipeBase &R : make_filter_range(*VPBB, [](VPRecipeBase &R) {
+           return isa<VPWidenMemoryRecipe>(&R);
+         })) {
+      auto *MemR = cast<VPWidenMemoryRecipe>(&R);
+      IRMemberToRecipe[&MemR->getIngredient()] = MemR;
```
lukel97, line 3562: I'm guessing here that `isa<VPWidenMemoryRecipe>(VPRecipeBase &)` doesn't work anymore and we have to check on the pointer type?

lukel97 (approved):
LGTM

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
fhahn: Yep, we could add a another set of CastInfo impls, but probably not worth it

PR #197498, opened 2026-05-13
Reapply "[libc][NFC] Propagate LIBC_CONSTEXPR qualifier for those depending on bitt_cast." (#197462)

lntue:
This reverts commit b20e0bc71db6d74b5a7abe68df577f3ed8fbcdbb.


boomanaiden154:
Pretty sure this reapplied the bad change.

rupprecht:
> Pretty sure this reapplied the bad change.

AFAICT, the initial land didn't break CI, at least not the builders that I looked at. The reapply the broke things included an additional commit: https://github.com/llvm/llvm-project/pull/197479/changes/aab9aa1966e26d692d074e742033d07b68e2ad93. This reapply does not include that change, i.e. it's an unmodified reapply of the original commit.

PR #197499, opened 2026-05-13
[VPlan] Add helper to iterate over all loop blocks in plain VPlan (NFC).

fhahn:
Add a new helper to return all blocks in an initial plain CFG VPlan, containing all blocks reachable from the header in RPOT up to the middle block, skipping any VPIRBasicBlocks (existing exit blocks).


llvm/lib/Transforms/Vectorize/VPlanCFG.h
```diff
@@ -259,6 +259,25 @@ vp_depth_first_shallow(const VPBlockBase *G) {
...
+  SmallVector<VPBasicBlock *, 16> Result;
+  ReversePostOrderTraversal<VPBlockShallowTraversalWrapper<VPBlockBase *>> RPOT(
+      Header);
+  for (VPBasicBlock *VPBB : VPBlockUtils::blocksOnly<VPBasicBlock>(RPOT)) {
+    if (VPBB == Middle)
+      break;
+    if (isa<VPIRBasicBlock>(VPBB))
+      continue;
+    Result.push_back(VPBB);
+  }
```
artagnon, line 276: Could make_filter_range, and return auto?

fhahn: Unfortunately I am not sure we can return an iterate range here easily, because we need to skip all VPBBs including and after the middle block. This is why the current implementation uses RPOT (which has local state) + break once we found the middle block

llvm/lib/Transforms/Vectorize/VPlanCFG.h
```diff
@@ -259,6 +259,24 @@ vp_depth_first_shallow(const VPBlockBase *G) {
   return depth_first(VPBlockShallowTraversalWrapper<const VPBlockBase *>(G));
 }
 
+/// Returns the VPBasicBlocks forming the loop body of a plain (pre-region)
+/// VPlan in reverse post-order starting from \p Header.
+inline SmallVector<VPBasicBlock *, 16>
```
artagnon:
```suggestion
inline SmallVector<VPBasicBlock *>
```

Would be good to not return a hard-coded size, as callers of this API might find it surprising?

artagnon (approved):
LGTM, thanks!

llvm/lib/Transforms/Vectorize/VPlanCFG.h
fhahn: Done thanks

llvm/lib/Transforms/Vectorize/VPlanCFG.h
```diff
@@ -259,6 +259,24 @@ vp_depth_first_shallow(const VPBlockBase *G) {
...
+/// Returns the VPBasicBlocks forming the loop body of a plain (pre-region)
+/// VPlan in reverse post-order starting from \p Header.
+inline SmallVector<VPBasicBlock *>
+vp_plain_cfg_loop_body(VPBasicBlock *Header) {
+  assert(!Header->getParent() && "Header must not be inside a region");
+  VPBlockBase *Middle = Header->getPredecessors()[1]->getSuccessors()[0];
+  SmallVector<VPBasicBlock *> Result;
+  ReversePostOrderTraversal<VPBlockShallowTraversalWrapper<VPBlockBase *>> RPOT(
+      Header);
+  for (VPBasicBlock *VPBB : VPBlockUtils::blocksOnly<VPBasicBlock>(RPOT)) {
```
ayalz, line 271: Better assert all blocks are basic rather than filter them

llvm/lib/Transforms/Vectorize/VPlanCFG.h
```diff
@@ -259,6 +259,24 @@ vp_depth_first_shallow(const VPBlockBase *G) {
...
+    if (VPBB == Middle)
+      break;
+    if (!isa<VPIRBasicBlock>(VPBB))
```
ayalz, line 274: These VPIRBasicBlocks are early exits? Perhaps worth are comment.

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
```diff
@@ -1160,20 +1160,15 @@ void VPlanTransforms::createInLoopReductionRecipes(VPlan &Plan,
...
-static bool areAllLoadsDereferenceable(VPBasicBlock *HeaderVPBB,
-                                       VPBasicBlock *MiddleVPBB, Loop *TheLoop,
+static bool areAllLoadsDereferenceable(VPBasicBlock *HeaderVPBB, Loop *TheLoop,
                                        PredicatedScalarEvolution &PSE,
                                        DominatorTree &DT, AssumptionCache *AC) {
   ScalarEvolution &SE = *PSE.getSE();
   const DataLayout &DL = TheLoop->getHeader()->getDataLayout();
-  for (VPBasicBlock *VPBB : VPBlockUtils::blocksOnly<VPBasicBlock>(
-           vp_depth_first_shallow(HeaderVPBB))) {
-    // Skip blocks outside the loop (exit blocks and their successors).
```
ayalz, line 1174: Comment regarding (early) exit blocks worth retaining as noted above.

llvm/lib/Transforms/Vectorize/VPlanCFG.h
```diff
@@ -259,6 +259,24 @@ vp_depth_first_shallow(const VPBlockBase *G) {
   return depth_first(VPBlockShallowTraversalWrapper<const VPBlockBase *>(G));
 }
 
+/// Returns the VPBasicBlocks forming the loop body of a plain (pre-region)
+/// VPlan in reverse post-order starting from \p Header.
+inline SmallVector<VPBasicBlock *>
+vp_plain_cfg_loop_body(VPBasicBlock *Header) {
```
ayalz, line 265: Worth indicating in the name that the body is provided in rpo.

ayalz (commented):
Post-commit nits.

llvm/lib/Transforms/Vectorize/VPlanCFG.h
fhahn: Should be updated in Should be updated in https://github.com/llvm/llvm-project/commit/2e4c820c05fd7a7bb5e78ecacac438f5efe9bd60, thanks!

llvm/lib/Transforms/Vectorize/VPlanCFG.h
fhahn: Should be updated in Should be updated in https://github.com/llvm/llvm-project/commit/2e4c820c05fd7a7bb5e78ecacac438f5efe9bd60 + assert added, thanks!

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
fhahn: Added back thanks

llvm/lib/Transforms/Vectorize/VPlanCFG.h
fhahn: Added new helper in https://github.com/llvm/llvm-project/commit/e024375368d50a33a020c56da373ef185c6a4873, thanks

PR #197501, opened 2026-05-13
Fix failure for clang/test/Frontend/rewrite-includes-bom.c for zOS

bahareh-farhadi:
This test is failing on zOS due to the lowercase HEX characters in the BOM. This PR changes the test to do a case-insensitive search.


abhina-sree (approved):
LGTM

PR #197502, opened 2026-05-13
[lldb] Fix TestBuiltinDebugTrap.py on ARM platform

DuncanMcBain:
The workaround for x86_64 which changed the handling of SIGILL signals is also required on 32-bit ARM platforms, because __builtin_trap(), which compiles to a UDF instruction, also generates SIGILL. If we don't change the handling here, the inferior is killed and the stop reason is "none".

The failure is here: https://lab.llvm.org/buildbot/#/builders/18/builds/25948


DavidSpickett (approved):
LGTM, let's see what the bot thinks.

PR #197503, opened 2026-05-13
[AArch64][ISel] Improve lowering for clmul.nxv8i16(zext, zext)

MDevereau:
Since only the lower bits of the clmul operands are non-zero, we can just use pmullb for simpler lowering. This enables optimal lowering for targets that have +sve2 or +sme in the i16 and i64 cases.

### DIFF
--- a/llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
+++ b/llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
@@ -2072,9 +2072,8 @@ AArch64TargetLowering::AArch64TargetLowering(const TargetMachine &TM,
     if (Subtarget->hasBF16())
       setPartialReduceMLAAction(ISD::PARTIAL_REDUCE_FMLA, MVT::nxv4f32,
                                 MVT::nxv8bf16, Legal);
+    setOperationAction(ISD::CLMUL, MVT::nxv2i64, Custom);
   }
-
-  setOperationAction(ISD::CLMUL, MVT::nxv2i64, Custom);
 
   // Handle non-aliasing elements mask
   if (Subtarget->hasSVE2() ||
@@ -8111,14 +8110,16 @@ SDValue AArch64TargetLowering::LowerCLMUL(SDValue Op, SelectionDAG &DAG) const {
   if (VT == MVT::nxv2i64) {
     // Lower to (.d pmullb(.s, .s)) for clmul.nxv2i64(zext(nxv2i32),
     // zext(nxv2i32))
-    if (Subtarget->isSVEorStreamingSVEAvailable() &&
-        (Subtarget->hasSVE2() || Subtarget->hasSME()) &&
+    if ((Subtarget->hasSVE2() || Subtarget->hasSME()) &&
         DAG.MaskedValueIsZero(Op.getOperand(0), HiWordMask) &&
-        DAG.MaskedValueIsZero(Op.getOperand(1), HiWordMask))
-      return MakePMULLB(
-          DAG.getNode(AArch64ISD::NVCAST, DL, MVT::nxv4i32, Op.getOperand(0)),
-          DAG.getNode(AArch64ISD::NVCAST, DL, MVT::nxv4i32, Op.getOperand(1)),
-          MVT::nxv4i32);
+        DAG.MaskedValueIsZero(Op.getOperand(1), HiWordMask)) {
+      SDValue OpA =
+          DAG.getNode(AArch64ISD::NVCAST, DL, MVT::nxv4i32, Op.getOperand(0));
+      SDValue OpB =
+          DAG.getNode(AArch64ISD::NVCAST, DL, MVT::nxv4i32, Op.getOperand(1));
+      return DAG.getNode(AArch64ISD::NVCAST, DL, VT,
+                         MakePMULLB(OpA, OpB, MVT::nxv4i32));
+    }
     // Lower to (.q pmullb(.d, .d)) for clmul.nxv2i64(nxv2i64, nxv2i64)
     if (Subtarget->hasSVEAES() &&
         (Subtarget->isSVEAvailable() || Subtarget->hasSSVE_AES()))
@@ -8135,7 +8136,8 @@ SDValue AArch64TargetLowering::LowerCLMUL(SDValue Op, SelectionDAG &DAG) const {
     // Lower to pmullb for clmul.nxv8i16(zext(nxv8i8), zext(nxv8i8))
     if (DAG.MaskedValueIsZero(Op.getOperand(0), HiWordMask) &&
         DAG.MaskedValueIsZero(Op.getOperand(1), HiWordMask))
-      return MakePMULLB(OpA, OpB, MVT::nxv16i8);
+      return DAG.getNode(AArch64ISD::NVCAST, DL, VT,
+                         MakePMULLB(OpA, OpB, MVT::nxv16i8));
 
     // clmul.i16(a, b) = xor(pmullb(a_lo, b_lo),
     //                       lsl(xor(pmul(a_hi, b_lo),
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
MDevereau: I was a bit concerned about the duplicated setOperationAction overriding logic here but the tests seem to hold.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -8099,16 +8099,42 @@ static SDValue LowerCLMUL(SDValue Op, SelectionDAG &DAG) {
...
+          VT == MVT::nxv8i16 || VT == MVT::nxv2i64) &&
          "Unexpected Type");
+  SDValue OpA = Op.getOperand(0);
+  SDValue OpB = Op.getOperand(1);
+  auto IsZextedFromHalfTy = [&](SDValue Op) {
+    uint64_t FullSize = VT.getScalarSizeInBits();
+    uint64_t HalfSize = FullSize / 2;
+    APInt HiWordMask = APInt::getBitsSet(FullSize, HalfSize, FullSize);
+    return DAG.MaskedValueIsZero(Op, HiWordMask);
+  };
```
paulwalker-arm: The lambda doesn't look worth it.  You only need a single variable for the mask (i.e. `HiWordMask = APInt::getHighBitsSet(FullSize, FullSize/2)`) and then all the call sites can use `MaskedValueIsZero()` directly.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -8099,16 +8099,42 @@ static SDValue LowerCLMUL(SDValue Op, SelectionDAG &DAG) {
...
+  auto MakePMULLB = [&](SDValue OpA, SDValue OpB) {
+    EVT NewVT = EVT::getVectorVT(
+        *DAG.getContext(),
+        VT.getVectorElementType().getHalfSizedIntegerVT(*DAG.getContext()),
+        VT.getVectorElementCount() * 2);
+    OpA = DAG.getNode(AArch64ISD::NVCAST, DL, NewVT, OpA);
+    OpB = DAG.getNode(AArch64ISD::NVCAST, DL, NewVT, OpB);
```
paulwalker-arm: In this case the lambda helps readability but I think it would be better for it to only create the call without the casting.  For all the use case the ValueType is statically know and so the MVT can be passed in. For the `nxv8i16` case you already have the casted operands ready to use.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
paulwalker-arm: In this case it would be cleaner to keep the `ISD::CLMUL` operation actions for `MVT::nxv2i64` together.  Otherwise you introduce ordering requirements we might trip over.

Up to you but there is a case for just forcing custom lowering regardless of the feature flags and then take the appropriate action within the lowering code (i.e. just return `Op` for the legal cases).  I only mention this because I wonder if the lowering for the top-half-is-zero case is the preferred output even when the quad variant of pmull is available.  It's the same number of instructions but perhaps the `and`s are more removable than the `trn`.

MDevereau: Forcing custom lowering looks like a good idea, The CHECK-SME-STREAMING-SSVE-AES and CHECK-SVE2-AES-NEXT benefit from nxv2i64 pmullb lowering properly now. It looks easier to stay sane with better future proofing by keeping the target checks inside the lowering code in my opinion.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -2074,9 +2074,7 @@ AArch64TargetLowering::AArch64TargetLowering(const TargetMachine &TM,
                                 MVT::nxv8bf16, Legal);
   }
 
-  if (Subtarget->hasSVEAES() &&
-      (Subtarget->isSVEAvailable() || Subtarget->hasSSVE_AES()))
-    setOperationAction(ISD::CLMUL, MVT::nxv2i64, Legal);
+  setOperationAction(ISD::CLMUL, MVT::nxv2i64, Custom);
```
paulwalker-arm: This needs moving up a couple of lines so it sits inside the `isSVEorStreamingSVEAvailable()` block.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -8095,23 +8093,55 @@ SDValue AArch64TargetLowering::LowerFMA(SDValue Op, SelectionDAG &DAG) const {
...
+  auto MakePMULLB = [&](SDValue OpA, SDValue OpB, EVT NewVT) {
+    SDValue PMULLB =
+        DAG.getTargetConstant(Intrinsic::aarch64_sve_pmullb_pair, DL, MVT::i64);
+    return DAG.getNode(ISD::INTRINSIC_WO_CHAIN, DL, NewVT, PMULLB, OpA, OpB);
+  };
+
+  if (VT == MVT::nxv2i64) {
+    // Lower to (.d pmullb(.s, .s)) for clmul.nxv2i64(zext(nxv2i32),
+    // zext(nxv2i32))
+    if (Subtarget->isSVEorStreamingSVEAvailable() &&
```
paulwalker-arm: You can drop the `isSVEorStreamingSVEAvailable()` check because getting this far with a scalable vector MVT without an SVE environment means we have bigger, pretty much unrecoverable, problems.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -8095,23 +8093,55 @@ SDValue AArch64TargetLowering::LowerFMA(SDValue Op, SelectionDAG &DAG) const {
...
   if (VT == MVT::nxv8i16) {
+    SDValue OpA =
+        DAG.getNode(AArch64ISD::NVCAST, DL, MVT::nxv16i8, Op.getOperand(0));
+    SDValue OpB =
+        DAG.getNode(AArch64ISD::NVCAST, DL, MVT::nxv16i8, Op.getOperand(1));
+
+    // Lower to pmullb for clmul.nxv8i16(zext(nxv8i8), zext(nxv8i8))
+    if (DAG.MaskedValueIsZero(Op.getOperand(0), HiWordMask) &&
+        DAG.MaskedValueIsZero(Op.getOperand(1), HiWordMask))
+      return MakePMULLB(OpA, OpB, MVT::nxv16i8);
```
paulwalker-arm: The result needs to be nvcasted back to `MVT::nxv8i16` or VT.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -8095,23 +8093,55 @@ SDValue AArch64TargetLowering::LowerFMA(SDValue Op, SelectionDAG &DAG) const {
...
+    // Lower to (.d pmullb(.s, .s)) for clmul.nxv2i64(zext(nxv2i32),
+    // zext(nxv2i32))
+    if (Subtarget->isSVEorStreamingSVEAvailable() &&
+        (Subtarget->hasSVE2() || Subtarget->hasSME()) &&
+        DAG.MaskedValueIsZero(Op.getOperand(0), HiWordMask) &&
+        DAG.MaskedValueIsZero(Op.getOperand(1), HiWordMask))
+      return MakePMULLB(
+          DAG.getNode(AArch64ISD::NVCAST, DL, MVT::nxv4i32, Op.getOperand(0)),
+          DAG.getNode(AArch64ISD::NVCAST, DL, MVT::nxv4i32, Op.getOperand(1)),
+          MVT::nxv4i32);
```
paulwalker-arm: The result needs to be nvcasted back to `MVT::nxv2i64` or `VT`.

paulwalker-arm (approved):

PR #197505, opened 2026-05-13
Revert "Reapply "[libc][NFC] Propagate LIBC_CONSTEXPR qualifier for those depending on bitt_cast." (#197462)"

boomanaiden154:
Reverts llvm/llvm-project#197498

There was a revert that landed around the same time (db58e2695d66687158201b560a6e203e0a671587), so this revert ended up actually reapplying the patch.


boomanaiden154:
Although it seems like this might have fixed some of the issues with the original commit given it passes our internal CI?

boomanaiden154:
Nevermind, the reland should have fixed the issues.

The lack of detailed commit message/incorrect stuff in the commit message was confusing.

PR #196868, opened 2026-05-11
[CIR][CIRGen] Cast stack allocas to the language-visible address space

RiverDave:
This patch aims to improve parity with OG codegen on targets with non-flat alloca address space. I observed this after getting some crashes while compiling PolybenchGpu for HIP (amdgpu). This work had previously been merged in the incubator, most notably: https://github.com/llvm/clangir/pull/2090, https://github.com/llvm/clangir/pull/2088.

CIR currently returns the raw `cir.alloca` address from temporary/local alloca creation. On AMDGPU, stack allocas live in private addrspace(5), but ordinary C/C++/HIP auto variables are still used through the language-visible generic/flat address space.

OG CodeGen handles this by creating the alloca in the target stack address space and immediately casting it to the language-visible address space when those differ. For example:

```llvm
%fmt = alloca [6 x i8], align 1, addrspace(5)
%tmp = alloca ptr, align 8, addrspace(5)

%fmt.ascast = addrspacecast ptr addrspace(5) %fmt to ptr
%tmp.ascast = addrspacecast ptr addrspace(5) %tmp to ptr

%arraydecay = getelementptr inbounds [6 x i8], ptr %fmt.ascast, i64 0, i64 0
store ptr %arraydecay, ptr %tmp.ascast, align 8
```
I've also added a helper to recover the underlying alloca through casts for callers that need to annotate the original cir.alloca, and preserves source pointer address spaces when creating pointer bitcasts.

PR #197506, opened 2026-05-13
[CIR][CodeGen] Replace errorNYI with assert for address space in emitAutoVarAlloca

AbdallahRashed:
Auto variables can only be in the default address space, or opencl_private when compiling OpenCL. Replace the errorNYI with an assert matching OG codegen (CGDecl.cpp).

Fixes part of #160386


clang/lib/CIR/CodeGen/CIRGenDecl.cpp
```diff
@@ -31,8 +31,9 @@ CIRGenFunction::AutoVarEmission
 CIRGenFunction::emitAutoVarAlloca(const VarDecl &d,
                                   mlir::OpBuilder::InsertPoint ip) {
   QualType ty = d.getType();
-  if (ty.getAddressSpace() != LangAS::Default)
-    cgm.errorNYI(d.getSourceRange(), "emitAutoVarAlloca: address space");
+  assert(
```
andykaylor, line 35: My concern with this is that classic codegen has a call to `MaybeCastStackAddressSpace` in the NRVO handling that we're missing, and it passes the address space to `CreateTempAlloca` in one place, so even though this assert is correct, the diagnostic is still needed until we are handling other address spaces correctly here.

@RiverDave is working on alloca address space handling now, so we might be able to do this soon.

RiverDave:
@AbdallahRashed you may be blocked by https://github.com/llvm/llvm-project/pull/196868 landing first. I'll ping when that dependency is closed.

Men-cotton:
@AbdallahRashed The blocker PR is merged, so could you update the test please?

AbdallahRashed:
@Men-cotton  test case has been updated. Thank you

clang/lib/CIR/CodeGen/CIRGenDecl.cpp
AbdallahRashed: @andykaylor  I have updated the test case to use cir.cast address_space thank you

Men-cotton:
The failing tests look unrelated to this PR. #199656 should address this failure.

clang/lib/CIR/CodeGen/CIRGenDecl.cpp
```diff
@@ -32,8 +32,9 @@ CIRGenFunction::AutoVarEmission
...
+      ty.getAddressSpace() == LangAS::Default ||
+      (ty.getAddressSpace() == LangAS::opencl_private && getLangOpts().OpenCL));
```
andykaylor, line 37: This looks much more specific than I was expecting, but I see that it matches the assert in classic codegen, so I guess it's OK.

andykaylor (approved):
lgtm

Men-cotton (approved):

PR #197507, opened 2026-05-13
[DirectX] Handle undef the same way as null

hvdijk:


hvdijk:
The undef deprecator should be ignored here, this is one case where we must not use `poison`, it did not exist yet in LLVM 3.7.

bogner (approved):

PR #197419, opened 2026-05-13
[NFCI][offload] Clarify naming for block and thread numbers

ro-i:
As discussed in
https://github.com/llvm/llvm-project/pull/195102#discussion_r3205402105 and on Slack.

PR #197511, opened 2026-05-13
[NFC][offload] Some more clarifying renaming of threads and blocks variables

ro-i:
As requested in https://github.com/llvm/llvm-project/pull/197419.


kevinsala (approved):
LGTM. Thanks!

PR #197514, opened 2026-05-13
[MLIR][XeGPU] Clone trivial operations with multiple consumers in layout propagation to avoid layout conflict

nbpatel:
In XeGPU layout propagation, when a trivially rematerializable producer (e.g. a CSE-merged vector.step) has multiple uses demanding distinct layouts, clone the producer once per required layout so each def retains a single well-defined layout. This avoids inserting a bridging xegpu.convert_layout whose WG-to-SG lowering would otherwise force a costly SLM round-trip to redistribute data across subgroups.


mlir/lib/Dialect/XeGPU/Transforms/XeGPULayoutImpl.cpp
```diff
@@ -124,13 +126,90 @@ static xegpu::DistributeLayoutAttr getLayoutFromUsePoints(Value result) {
...
+// regions, and all operands are themselves trivially rematerializable, e.g.
+// block-arg-free pure value generators such as `vector.step`, splat
+// `arith.constant`, or `vector.create_mask` whose operands are constants).
+static bool isTriviallyRematerializable(Operation *op) {
+  if (!op || op->getNumRegions() != 0)
+    return false;
+  if (!isMemoryEffectFree(op))
+    return false;
+  for (Value v : op->getOperands()) {
+    Operation *defOp = v.getDefiningOp();
```
Garra1980, line 138: can you pls add a test for this case not just operand-less step?

charithaintc (approved):

mlir/lib/Dialect/XeGPU/Transforms/XeGPULayoutImpl.cpp
Jianhui-Li: +1. 
Could enhance the step test below by multiple with a constant stride first, and then use it to broadcast row-wise and column-wise to create the layout conflict.

Jianhui-Li (approved):
LGTM with a suggestion to enhance the test.

mlir/lib/Dialect/XeGPU/Transforms/XeGPULayoutImpl.cpp
nbpatel: added

PR #196519, opened 2026-05-08
[libc] Build with -Wshadow

nico:
This way, clients including libc/shared/math.h don't need to `#pragma GCC diagnostic ignored "-Wshadow"` around the include.

This works locally after #196337 #196342 #196346.

CI also needed #196529 #196810 #196850 #196851 #196852 #196853 #196855 #196857 #196858 #196859 #196860.

PR #197516, opened 2026-05-13
[libc] Build with -Wshadow

nico:
This relands https://github.com/llvm/llvm-project/pull/196519. It now only enables the warning for
clang 22 and newer, since older versions had false positives.

### DIFF
--- a/libc/cmake/modules/LLVMLibCCompileOptionRules.cmake
+++ b/libc/cmake/modules/LLVMLibCCompileOptionRules.cmake
@@ -305,6 +305,12 @@ function(_get_common_compile_options output_var flags)
       list(APPEND compile_options "-Wthread-safety")
       list(APPEND compile_options "-Wglobal-constructors")
     endif()
+
+    # Older Clang versions emit false positive shadow warnings for lambda captures
+    # inside static member functions (fixed in Clang 22, see PR #157667 and #165919).
+    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "22.0.0")
+      list(APPEND compile_options "-Wshadow")
+    endif()
   elseif(MSVC)
     list(APPEND compile_options "/EHs-c-")
     list(APPEND compile_options "/GR-")
--- a/libc/src/__support/math/sqrtf128.h
+++ b/libc/src/__support/math/sqrtf128.h
@@ -60,6 +60,8 @@ namespace math {
 
 namespace sqrtf128_internal {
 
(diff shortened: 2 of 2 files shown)


simpal01:
Hi,

We (https://github.com/arm/arm-toolchain) started seeing a new build failure after the recent `[libc] Build with -Wshadow` change in PR https://github.com/llvm/llvm-project/pull/197516 

The error we see is:
```
    In file included from libc/src/stdio/vsprintf.cpp:17:
    In file included from libc/src/stdio/printf_core/printf_main.h:15:
    In file included from libc/src/stdio/printf_core/converter.h:21:
    In file included from libc/src/stdio/printf_core/converter_atlas.h:35:
    libc/src/stdio/printf_core/float_hex_converter.h:36:9: error:
      declaration shadows a type alias in namespace '...::printf_core'
      [-Werror,-Wshadow]
        using StorageType = LDBits::StorageType;
              ^
    libc/src/stdio/printf_core/float_dec_converter_limited.h:65:7: note:
      previous declaration is here
    using StorageType = fputil::FPBits<long double>::StorageType;
```
  With -Wshadow enabled, this now fails because both
  `float_dec_converter_limited.h` and `float_hex_converter.h` define `StorageType` which Clang now diagnoses as shadowing and fails.

simpal01:
Ping

michaelrj-google:
Hi @simpal01, I'm looking into a fix for that shadowing issue. Could you confirm that you're only seeing shadowing errors in those printf pieces?

michaelrj-google:
Managed to reproduce the issue locally, I've created https://github.com/llvm/llvm-project/pull/197985 which should fix it.

nico:
> Managed to reproduce the issue locally, I've created https://github.com/llvm/llvm-project/pull/197985 which should fix it.

Thank you!

> Hi,
> 
> We (https://github.com/arm/arm-toolchain) started seeing a new build failure…

If this is a configuration that you care about, would it be possible that you try to get it added to libc's pre-commit testing? Thanks!

simpal01:
> Hi @simpal01, I'm looking into a fix for that shadowing issue. Could you confirm that you're only seeing shadowing errors in those printf pieces?

Yes, your patch fixed the build. Thanks.

simpal01:
> > Managed to reproduce the issue locally, I've created #197985 which should fix it.
> 
> Thank you!
> 
> > Hi,
> > We (https://github.com/arm/arm-toolchain) started seeing a new build failure…
> 
> If this is a configuration that you care about, would it be possible that you try to get it added to libc's pre-commit testing? Thanks!

Yes, we are working on upstreaming a few arm bare-metal configurations for libc testing. Thanks.

PR #197518, opened 2026-05-13
[flang][cuda] Honor !dir$ ignore_tkr(m) under -gpu=mem:{unified,managed}

wangzpgi:
A device-typed dummy with `!dir$ ignore_tkr(m)` is meant to be an overload discriminator (only selected for actuals with an explicit `device/managed/unified` attribute). Skip the host->device relaxation in AreCompatibleCUDADataAttrs when `IgnoreTKR::Managed` is set so unattributed host actuals no longer bind to such a dummy.

Also document the §3.2.3 matching distance table next to GetMatchingDistance and add LIT tests for the full Table 2 grid
and the ignore_tkr(m) carve-out.


clementval:
Do we have a link to the documentation of `!dir$ ignore_tkr(m)`?

eugeneepshteyn:
> Do we have a link to the documentation of `!dir$ ignore_tkr(m)`?

It's very briefly described in https://github.com/llvm/llvm-project/blob/main/flang/docs/Directives.md . If we could add documentation from the comments in this PR to that Directives.md document, I think it would be helpful to the users.

flang/docs/Directives.md
```diff
@@ -29,6 +29,44 @@ A list of non-standard directives supported by Flang
   argument's descriptor and passed as a raw pointer.
   The letter (P) ignores pointer and allocatable matching, so that one can pass
   an allocatable array to routine with pointer array argument and vice versa.
+  The letter (M) disables matching of the actual argument's CUDA storage
+  (managed/unified) against the dummy's. Combined with a `device`-typed dummy,
+  this is most often used to mark a specific as an overload discriminator in
+  host modules that pair host- and device-typed specifics (the cuBLAS V2
+  module's host-pointer-mode and device-pointer-mode wrappers are a canonical
+  example). Under `-gpu=mem:unified` or `-gpu=mem:managed` the compiler
```
clementval: cuBLAS module is not upstream so avoid mentioning it.

clementval (approved):
I had to read it a couple of times to understand the comments. I find it hard to read but I don't see how to make it better. Approving.

wangzpgi:
> I had to read it a couple of times to understand the comments. I find it hard to read but I don't see how to make it better. Approving.

Good point — I've tightened it; let me know if it reads better now.

PR #197519, opened 2026-05-13
[compiler-rt] Use `size_t` rather than `int` for first argument to `__atomic_load_c` et al.

sbc100:
I noticed this discrepancy in emscripten when trying to test 128 bit atomics under wasm64: https://github.com/emscripten-core/emscripten/pull/26937

The LLVM CodeGen appears to use `size_t` in this position when it generates calls to these functions.

This doesn't effect other platforms I imagine because they don't require signature checking at the linker level.

This doesn't effect wasm32 where size_t and int are the same size.

### DIFF
--- a/compiler-rt/test/builtins/Unit/atomic_test.c
+++ b/compiler-rt/test/builtins/Unit/atomic_test.c
@@ -34,23 +34,23 @@
 bool __atomic_is_lock_free_c(size_t size, void *ptr)
     EXTERNAL_NAME(__atomic_is_lock_free);
 
-void __atomic_load_c(int size, void *src, void *dest,
-                     int model) EXTERNAL_NAME(__atomic_load);
+void __atomic_load_c(size_t size, void *src, void *dest, int model)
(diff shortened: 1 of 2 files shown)


sbc100:
See https://github.com/llvm/llvm-project/blob/cca6a4c21c71e3697667b64b0be291e9a17ce368/clang/lib/CodeGen/CGAtomic.cpp#L1233-L1242

compiler-rt/test/builtins/Unit/atomic_test.c
```diff
@@ -34,31 +34,31 @@
 bool __atomic_is_lock_free_c(size_t size, void *ptr)
     EXTERNAL_NAME(__atomic_is_lock_free);
 
-void __atomic_load_c(int size, void *src, void *dest,
-                     int model) EXTERNAL_NAME(__atomic_load);
+void __atomic_load_c(sizt_t size, void *src, void *dest, int model)
```
MaskRay: typo

MaskRay (approved):

PR #197521, opened 2026-05-13
AMDGPU/GlobalISel: Add RegBankLegalize rule for amdgcn_kill

vangthao95:
Disable wqm test in llvm.amdgcn.kill.ll and added a FIXME to re-enable after wqm.vote is supported by GISEL RegBankLegalize.


llvm/lib/Target/AMDGPU/AMDGPURegBankLegalizeRules.cpp
```diff
@@ -1948,6 +1939,8 @@ RegBankLegalizeRules::RegBankLegalizeRules(const GCNSubtarget &_ST,
 
   addRulesForIOpcs({amdgcn_wqm_demote}).Any({{}, {{}, {IntrId, Vcc}}});
 
+  addRulesForIOpcs({amdgcn_kill}).Any({{}, {{}, {IntrId, Vcc}}});
```
petar-avramovic: can put it in same group as amdgcn_wqm_demote

petar-avramovic (approved):

PR #196305, opened 2026-05-07
[AArch64] Do not generate indexed addressing mode for volatile accesses

nikic:
Instructions performing register writeback do not set a valid instruction syndrome, making it impossible to handle MMIO in protected hypervisors. Suppress the use of postinc/preinc addressing modes for volatile accesses, which may be used to interact with MMIO.

There are three different places that can form indexed addressing modes:

 * GISel via isIndexingLegal()
 * SDAG via getPreIndexedAddressParts() and getPostIndexedAddressParts()
 * AArch64LoadStoreOptimizer

The the latter case, exclude volatile accesses on SP (which are relevant for stack probing) and MTE tag stores, as both cannot be MMIO.

Fixes https://github.com/llvm/llvm-project/issues/173014.

PR #197522, opened 2026-05-13
[AArch64] Keep MMO when converting gather lane to LDRSui.

davemgreen:
We were losing the MMO when converting the load. Make sure we copy them over, which apparently alters codegen more than I expected and helps keep postinc generation after #196305.


jcohen-apple (approved):
LGTM, thanks

davemgreen:
Thanks

PR #197523, opened 2026-05-13
[clang] make evaluation of type constraint a SFINAE context

mizvekov:
Otherwise, errors when substituting a type constraint could unintentionally make the program ill-formed.

This also strenghtens the assert which checks, when we are instantiating templates, that we either have a code synthesis context, or that we are in a SFINAE context.

### DIFF
--- a/clang/lib/Sema/SemaExprCXX.cpp
+++ b/clang/lib/Sema/SemaExprCXX.cpp
@@ -7995,6 +7995,8 @@ Sema::BuildExprRequirement(
     assert(TC && "Type Constraint cannot be null here");
     auto *IDC = TC->getImmediatelyDeclaredConstraint();
     assert(IDC && "ImmediatelyDeclaredConstraint can't be null here.");
+
+    SFINAETrap Trap(*this);
     ExprResult Constraint = SubstExpr(IDC, MLTAL);
     bool HasError = Constraint.isInvalid();
     if (!HasError) {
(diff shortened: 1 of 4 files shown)


clang/lib/Sema/SemaExprCXX.cpp
```diff
@@ -8009,6 +8011,8 @@ Sema::BuildExprRequirement(
         HasError = true;
     }
     if (HasError) {
+      // FIXME: Capture diagnostics from the SFINAE trap and store them in the
```
erichkeane, line 8009: How bad of a 'fixme' is this?  Are repros here actually somewhat debuggable by someone seeing this diagnostic?

clang/test/SemaTemplate/instantiate-requires-expr.cpp
```diff
@@ -309,3 +309,24 @@ struct test {
...
+  template<class T>
+    requires requires { { T() } -> C; } // expected-note {{because 'C<expr-type>' would be invalid}}
+    void g();
+  // expected-note@-1 {{ignored: constraints not satisfied}}
+  template void g<int>();
+  // expected-error@-1 {{does not refer to a function template}}
+
+  // FIXME: static_assert should gain support for explaining non-satisfied requirements.
+  static_assert(requires { { 0 } -> C; });
+  // expected-error@-1 {{static assertion failed due to requirement 'requires { { <<error-expression>> } -> C; }'}}
```
erichkeane, line 331: oof... this really stinks...

erichkeane (commented):
How much effort would it be to do the 'fixme' here too?  That is a pretty annoying/difficult to figure out diagnostic without the notes.

clang/lib/Sema/SemaExprCXX.cpp
mizvekov: I don't think it's too difficult, it's just that the requirement has only one slot for storing an arbitrary string for a note, and that note is not worded correctly for a nested note.

erichkeane (approved):

PR #197525, opened 2026-05-13
[AArch64] tune budget of small multi-exit loops for Apple M-family

ayasin-a:
Small multi-exit loops (e.g. std::find-like) benefit from runtime unrolling on Apple cores; however the common AArch64 code uses a small threshold - tighter than the budget of 10 already used for single-exit loops in getAppleRuntimeUnrollPreferences.

This PR parameterizes the budget and raises it to 10 for Apple-M-family (others keep 5), sharing the value via a new AppleMLoopSizeBudget constant so the two budgets stay in-sync.

### DIFF
--- a/llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
+++ b/llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
@@ -5276,7 +5276,8 @@ static bool isLoopSizeWithinBudget(Loop *L, const AArch64TTIImpl &TTI,
 }
 
 static bool shouldUnrollMultiExitLoop(Loop *L, ScalarEvolution &SE,
-                                      const AArch64TTIImpl &TTI) {
+                                      const AArch64TTIImpl &TTI,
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
```diff
@@ -5275,7 +5275,8 @@ static bool isLoopSizeWithinBudget(Loop *L, const AArch64TTIImpl &TTI,
 }
 
 static bool shouldUnrollMultiExitLoop(Loop *L, ScalarEvolution &SE,
-                                      const AArch64TTIImpl &TTI) {
+                                      const AArch64TTIImpl &TTI,
+                                      const unsigned SizeBudget = 5) {
```
jroelofs: There's only one call site, so no need for the default init here.

fhahn (commented):
overall looks fine, but please provide some motivating data for increasing the budget

PR #197528, opened 2026-05-13
[clang][deps] Remove unnecessary function overload

jansvoboda11:
NFC


naveen-seth (commented):
This is great. Thanks!

Issue #192314, opened 2026-04-15
[CIR] Upstream missing support for static l-value casts


andykaylor:
### Overview

Upstream code from the incubator project to replace errorNYI calls in the
CIR codegen implementation related to additional cast kinds when emitting
cast lvalues:
- `CK_ConstructorConversion`, `CK_UserDefinedConversion`, and
  `CK_LValueToRValue` in `emitCastLValue` (handled by passing through to
  `emitLValue` on the subexpression)
- Address space conversion casts using `toCIRLangAddressSpaceAttr` and
  `getTargetHooks().performAddrSpaceCast()`
- `CK_NoOp` casts where the CIR memory types differ, requiring a
  `createElementBitCast()` to reconcile them

Upstream NYI locations:
- `CIRGenExpr.cpp` — `emitCastLValue`: unhandled cast kinds including
  `CK_ConstructorConversion`, `CK_UserDefinedConversion`, `CK_LValueToRValue`
- `CIRGenExpr.cpp` — `emitCastLValue`: address space conversion from non-target
  address space
- `CIRGenExpr.cpp` — `emitCastLValue`: `CK_NoOp` where CIR memory types differ

### Suggested minimal test case

```cpp
struct Base { int x; };
struct Derived : Base {
  operator Base&() { return *this; }
};

Base &test_user_conversion(Derived &d) {
  return static_cast<Base&>(d);
}

struct A {};
struct B : A {};

A &test_constructor_conversion(B &b) {
  return b;
}
```

### Existing incubator tests

- `clang/test/CIR/CodeGen/cast-lvalue.cpp`
- `clang/test/CIR/CodeGen/cast.cpp`
- `clang/test/CIR/CodeGen/address-space-conversion.cpp`

andykaylor:
@AbdallahRashed You could work on this one too, though in this case the upstream implementation has diverged from the incubator so it may take some investigation to figure out what the correct handling should be.

AbdallahRashed:
@andykaylor  please assign it to me.

andykaylor:
You can check with @RiverDave about the address space cast handling. He has done a lot of the CIR address space work both upstream and in the incubator.

AbdallahRashed:
Hello @RiverDave  I am trying to test my PR https://github.com/llvm/llvm-project/pull/197016  through this test case
__kernel void k(__global int *gp) {
  __global int &gr = *gp;
  int &r = gr;          // bind generic-AS reference to a __global lvalue
[...]

PR #197531, opened 2026-05-13
[CIR][CodeGen] Support CK_NoOp lvalue casts with differing memory types

AbdallahRashed:
Remove the array-only guard in emitCastLValue for CK_NoOp casts and unconditionally use withElementType when CIR memory types differ, matching OG codegen behavior.

Part of #192314.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenExpr.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExpr.cpp
@@ -1706,21 +1706,8 @@ LValue CIRGenFunction::emitCastLValue(const CastExpr *e) {
       Address v = lv.getAddress();
       if (v.isValid()) {
         mlir::Type ty = convertTypeForMem(e->getType());
-        if (v.getElementType() != ty) {
-          // We have only inspected/reproduced this with complete to incomplete
-          // array types, so we do an NYI for other cases, so we can make sure
-          // we're doing a conversion we want to be making.
-          auto fromTy = dyn_cast<cir::ArrayType>(v.getElementType());
-          auto toTy = dyn_cast<cir::ArrayType>(ty);
-          if (!fromTy || !toTy ||
-              fromTy.getElementType() != toTy.getElementType() ||
-              toTy.getSize() != 0)
-            cgm.errorNYI(e->getSourceRange(),
-                         "emitCastLValue NoOp not array-shrink case");
-
-          lv = makeAddrLValue(v.withElementType(builder, ty), e->getType(),
-                              lv.getBaseInfo());
-        }
+        if (v.getElementType() != ty)
+          lv.setAddress(v.withElementType(builder, ty));
       }
     }
     return lv;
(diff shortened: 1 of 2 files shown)


clang/lib/CIR/CodeGen/CIRGenExpr.cpp
```diff
@@ -1706,21 +1706,9 @@ LValue CIRGenFunction::emitCastLValue(const CastExpr *e) {
...
-          // we're doing a conversion we want to be making.
-          auto fromTy = dyn_cast<cir::ArrayType>(v.getElementType());
-          auto toTy = dyn_cast<cir::ArrayType>(ty);
-          if (!fromTy || !toTy ||
-              fromTy.getElementType() != toTy.getElementType() ||
-              toTy.getSize() != 0)
-            cgm.errorNYI(e->getSourceRange(),
-                         "emitCastLValue NoOp not array-shrink case");
-
+        if (v.getElementType() != ty)
```
andykaylor, line 1709: I'm concerned that the test you're adding isn't sufficient to satisfy the concerns @erichkeane had when he added this diagnostic. In particular, line 1721 is subtely different from what classic codegen does here, and I'm not really sure what effect the difference might have in places where this l-value is used.

clang/test/CIR/CodeGen/cast-lvalue-noop.cpp
```diff
@@ -0,0 +1,32 @@
+// RUN: %clang_cc1 -std=c++20 -triple x86_64-unknown-linux-gnu -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck --check-prefix=CIR --input-file=%t.cir %s
+// RUN: %clang_cc1 -std=c++20 -triple x86_64-unknown-linux-gnu -fclangir -emit-llvm %s -o %t-cir.ll
+// RUN: FileCheck --check-prefix=LLVM --input-file=%t-cir.ll %s
+// RUN: %clang_cc1 -std=c++20 -triple x86_64-unknown-linux-gnu -emit-llvm %s -o %t.ll
+// RUN: FileCheck --check-prefix=OGCG --input-file=%t.ll %s
+
+// Test CK_NoOp in emitCastLValue where the CIR memory types differ.
```
erichkeane: We typically dont add these comments to the top of tests in LLVM, they get stale qiuckly and cause folks to not-reuse test files they should be for more tests (we prefer more longer tests if possible).  In fact, this probably should end up in a shared test file somewhere.

Andy is right though, this test isn't sufficient to remove the assert. See all of the conditions above that I'm checking, if you want to allow a single type of it you need to specify that and make sure you're permitting that.

What you've removed allows all conversions from all types of all sizes, arrays or not, yet have only 1 insufficient test.

erichkeane (changes requested):

clang/lib/CIR/CodeGen/CIRGenExpr.cpp
AbdallahRashed: Hello @erichkeane  @andykaylor  I added one more test case, with some debug code. I confirm the two tests exercise all reachable paths through the type-mismatch check. The original 4 condition guard.
also I have used  lv.setAddress(v.withElementType(builder, ty)); to match the codegen. 

The previous version used makeAddrLValue(..., e->getType(), ...)] which would set the LValue's QualType to the cast expression's type ,whereas classic codegen LV.setAddress(V.withElementType(T))  only updates the IR level element type while preserving the subexpression's QualType.

PR #197533, opened 2026-05-13
[MemoryBuiltins][NFC] Allow users to retrieve detailed (de)allocation info

jdoerfert:
There are some helpers to inspect a value or call but not all
information about the (de)allocation are made available outside of
MemoryBuiltins.cpp. The two new functions allow users a more in-depth
view of (de)allocations through a single API.

### DIFF
--- a/llvm/include/llvm/Analysis/MemoryBuiltins.h
+++ b/llvm/include/llvm/Analysis/MemoryBuiltins.h
@@ -309,7 +309,7 @@ private:
   OffsetSpan combineOffsetRange(OffsetSpan LHS, OffsetSpan RHS);
   OffsetSpan computeImpl(Value *V);
   OffsetSpan computeValue(Value *V);
-  bool CheckedZextOrTrunc(APInt &I);
+  bool checkedZextOrTrunc(APInt &I);
 };
 
 /// SizeOffsetValue - Used by \p ObjectSizeOffsetEvaluator, which works with
--- a/llvm/include/llvm/Analysis/MemoryBuiltins.h
+++ b/llvm/include/llvm/Analysis/MemoryBuiltins.h
@@ -85,6 +85,11 @@ LLVM_ABI bool isLibFreeFunction(const Function *F, const LibFunc TLIFn);
 LLVM_ABI Value *getFreedOperand(const CallBase *CB,
                                 const TargetLibraryInfo *TLI);
 
+/// If this if a call to a free function, return the freed operand number, or
+/// -1.
+LLVM_ABI int getFreedOperandNo(const CallBase *CB,
+                               const TargetLibraryInfo *TLI);
+
 //===----------------------------------------------------------------------===//
 //  Properties of allocation functions
 //
@@ -130,6 +135,43 @@ LLVM_ABI Constant *getInitialValueOfAllocation(const Value *V,
 /// of functions.
 LLVM_ABI std::optional<StringRef>
 getAllocationFamily(const Value *I, const TargetLibraryInfo *TLI);
+
+/// Description of an allocation call. Note that some elements might be
+/// null/negative.
+struct AllocationCallInfo {
+  /// The name of the allocation family, e.g., malloc.
+  std::optional<StringRef> Family;
+  /// The known initial value, usually 0, UndefValue, or unknown (=nullptr).
+  Constant *InitialValue = nullptr;
+  /// The total size is the product of the SizeLHS and SizeRHS argument values,
+  /// if both are non-negative, otherwise it is simply the argument with a
+  /// non-negative value.
+  int SizeLHSArgNo = -1, SizeRHSArgNo = -1;
+  /// The requested alignment is encoded in this argument.
+  int AlignmentArgNo = -1;
+};
+
+/// Return all known information about the allocation call \p CB.
+LLVM_ABI std::optional<AllocationCallInfo>
+getAllocationCallInfo(const CallBase *CB, const TargetLibraryInfo *TLI,
+                      Type *InitialValueTy);
+
+/// Description of an allocation call. Note that some elements might be
+/// null/negative.
+struct DeallocationCallInfo {
+  /// The name of the allocation family, e.g., malloc.
+  std::optional<StringRef> Family;
+  /// The freed operand is passed in this argument.
+  int FreedOperandArgNo = -1;
+  /// The total size is encoded in this argument.
+  int SizeArgNo = -1;
+  /// The requested alignment is encoded in this argument.
+  int AlignmentArgNo = -1;
+};
+
+/// Return all known information about the dellocation call \p CB.
+LLVM_ABI std::optional<DeallocationCallInfo>
+getDeallocationCallInfo(const CallBase *CB, const TargetLibraryInfo *TLI);
 
 //===----------------------------------------------------------------------===//
 //  Utility functions to compute size of objects.
--- a/llvm/include/llvm/Analysis/MemoryBuiltins.h
+++ b/llvm/include/llvm/Analysis/MemoryBuiltins.h
@@ -85,6 +85,11 @@ LLVM_ABI bool isLibFreeFunction(const Function *F, const LibFunc TLIFn);
 LLVM_ABI Value *getFreedOperand(const CallBase *CB,
                                 const TargetLibraryInfo *TLI);
 
+/// If this if a call to a free function, return the freed operand number, or
+/// -1.
+LLVM_ABI int getFreedOperandNo(const CallBase *CB,
+                               const TargetLibraryInfo *TLI);
+
 //===----------------------------------------------------------------------===//
 //  Properties of allocation functions
 //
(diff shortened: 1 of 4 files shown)


llvm/include/llvm/Analysis/MemoryBuiltins.h
```diff
@@ -131,6 +131,43 @@ LLVM_ABI Constant *getInitialValueOfAllocation(const Value *V,
...
+/// Description of an allocation call. Note that some elements might be
+/// null/negative.
+struct DeallocationCallInfo {
+  /// The name of the allocation family, e.g., malloc.
+  std::optional<StringRef> Family;
+  /// The freed operand is passed in this argument.
+  int FreedOperandArgNo = -1;
+  /// The total size is encoded in this argument.
+  int SizeArgNo = -1;
+  /// The requested alignment is endoded in this argument.
```
nikic:
```suggestion
  /// The requested alignment is encoded in this argument.
```

llvm/lib/Analysis/MemoryBuiltins.cpp
```diff
@@ -438,42 +438,62 @@ Constant *llvm::getInitialValueOfAllocation(const Value *V,
...
+  ACI.InitialValue = getInitialValueOfAllocation(CB, TLI, InitialValueTy);
+  return ACI;
+}
+
 struct FreeFnsTy {
   unsigned NumParams;
+  // Size parameter, or -1 if not available.
+  int SizeParam;
+  // Alignment parameter for aligned_alloc and aligned new
+  int AlignParam;
```
nikic, line 473: Where is this going to be used? allocsize and allocalign are currently only spec'd for allocation functions, not for deallocation functions, so this is only going to do something for functions using the legacy listing in this file (unlike all other alloc/dealloc functions, which use IR attributes). I'd like to make sure that we don't have anything here that is not encodable in IR. So if this is needed, we also need corresponding IR support.

nikic (commented):
Looks generally reasonable. Where will this be used?

jdoerfert:
> Looks generally reasonable. Where will this be used?

The instrumentor uses it to pass the information to the user provided runtime.
We use that information in the (GPU) sanitizer to increase allocated storage, and provide fake pointers.
Deallocation is useful to identify object lifetimes and deallocation without size/alignment that could benefit from such information.

The call support is done but I haven't open the PR because I need to clear some existing ones to get "the order right".

llvm/lib/Analysis/MemoryBuiltins.cpp
jdoerfert: I'm not sure I follow. We (seem to) have a pass that adds IR attributes to deallocation functions, e.g. this is what I see for `free`:


`allockind("free") memory(argmem: readwrite, inaccessiblemem: readwrite) "alloc-family"="malloc"`

Though, we do not annotate `::operator delete(F, std::align_val_t{64});` even though it appears in the table. I'm having a hard time getting rid of the `nobuiltin` for these deallocators. Even copying `clang/test/CodeGenCXX/builtin-operator-new-delete.cpp` I don't manage to get builtin deallocation for some reason:
https://godbolt.org/z/89zKG1xzz

nikic: Yes, we do infer allocator attributes (see BuildLibCalls). My point here was that SizeParam/AlignParam are currently not encodable in IR for deallocations. Everything else is fine.

jdoerfert: So, you don't want us to read the size/alignment from the table for now?
If so, what would need to be added to the IR? Basically attributes for those two properties?
Do we express them for allocations? Then I could just copy the handling. If not, the deallocation case would be in the same position as the allocation case, no? Both rely on the table for size/alignment.

jdoerfert:
> Looks generally reasonable. Where will this be used?

I put up the next PR (https://github.com/llvm/llvm-project/pull/198042) that uses the new functions.
I also put up a follow up to this one (http://github.com/llvm/llvm-project/pull/198049) since I realized alignment information was not properly propagated for functions not in the "allocator list".

DeallocationInfo actually looks good from what I can tell.

jdoerfert:
@nikic Are you going to look at these again? If not, I'll ask someone else to review.

llvm/lib/Analysis/MemoryBuiltins.cpp
```diff
@@ -99,56 +101,65 @@ StringRef mangledNameForMallocFamily(const MallocFamily &Family) {
...
 
 struct AllocFnsTy {
+  AllocFnsTy() {}
+  AllocFnsTy(AllocType AllocTy, unsigned NumParams, int FstParam, int SndParam,
+             int AlignParam, MallocFamily Family)
+      : AllocTy(AllocTy), NumParams(NumParams), FstParam(FstParam),
+        SndParam(SndParam), AlignParam(AlignParam),
+        Family(mangledNameForMallocFamily(Family)) {}
+
+  // TODO: Replace AllocTy with the familiy (String or enum). Right now these
```
antoniofrighetto:
```suggestion
  // TODO: Replace AllocTy with the family (String or enum). Right now these
```

llvm/include/llvm/Analysis/MemoryBuiltins.h
```diff
@@ -131,6 +136,43 @@ LLVM_ABI Constant *getInitialValueOfAllocation(const Value *V,
...
+  /// The requested alignment is encoded in this argument.
+  int AlignmentArgNo = -1;
+};
+
+/// Return all known information about the allocation call \p CB.
+LLVM_ABI std::optional<AllocationCallInfo>
+getAllocationCallInfo(const CallBase *CB, const TargetLibraryInfo *TLI,
+                      Type *InitialValueTy);
+
+/// Description of an allocation call. Note that some elements might be
```
antoniofrighetto:
```suggestion
/// Description of a deallocation call. Note that some elements might be
```

llvm/lib/Analysis/MemoryBuiltins.cpp
```diff
@@ -99,56 +101,65 @@ StringRef mangledNameForMallocFamily(const MallocFamily &Family) {
...
+  // encode overlapping information and use different sources.
   AllocType AllocTy;
-  unsigned NumParams;
+  unsigned NumParams = -1;
```
antoniofrighetto: int?

llvm/include/llvm/Analysis/MemoryBuiltins.h
```diff
@@ -131,6 +136,43 @@ LLVM_ABI Constant *getInitialValueOfAllocation(const Value *V,
...
+  std::optional<StringRef> Family;
+  /// The freed operand is passed in this argument.
+  int FreedOperandArgNo = -1;
+  /// The total size is encoded in this argument.
+  int SizeArgNo = -1;
+  /// The requested alignment is encoded in this argument.
+  int AlignmentArgNo = -1;
+};
+
+/// Return all known information about the dellocation call \p CB.
```
antoniofrighetto:
```suggestion
/// Return all known information about the deallocation call \p CB.
```

llvm/include/llvm/Analysis/MemoryBuiltins.h
```diff
@@ -131,6 +136,43 @@ LLVM_ABI Constant *getInitialValueOfAllocation(const Value *V,
 LLVM_ABI std::optional<StringRef>
 getAllocationFamily(const Value *I, const TargetLibraryInfo *TLI);
 
+/// Description of an allocation call. Note that some elements might be
+/// null/negative.
+struct AllocationCallInfo {
+  /// The name of the allocation family, e.g., malloc.
+  std::optional<StringRef> Family;
```
antoniofrighetto, line 143: Can't we just have a StringRef (and pass an empty one if needed)?

llvm/lib/Analysis/MemoryBuiltins.cpp
```diff
@@ -99,56 +101,65 @@ StringRef mangledNameForMallocFamily(const MallocFamily &Family) {
...
   // First and Second size parameters (or -1 if unused)
-  int FstParam, SndParam;
+  int FstParam, SndParam = -1;
```
antoniofrighetto:
```suggestion
  int FstParam = -1, SndParam = -1;
```

llvm/lib/Analysis/MemoryBuiltins.cpp
```diff
@@ -534,22 +571,66 @@ bool llvm::isLibFreeFunction(const Function *F, const LibFunc TLIFn) {
...
+
+  // Fallback, check arguments.
+  if (checkFnAllocKind(CB, AllocFnKind::Free))
+    DCI.FreedOperandArgNo =
+        CB->getArgOperandNoWithAttribute(Attribute::AllocatedPointer);
+  if (DCI.FreedOperandArgNo < 0)
+    return std::nullopt;
+
+  DCI.Family = getAllocationFamily(CB, TLI);
+  DCI.SizeArgNo = CB->getArgOperandNoWithAttribute(Attribute::AllocSize);
```
antoniofrighetto: Following up on above, allocsize should be a function attribute, not a per-argument one. Is this code tested?

llvm/lib/Analysis/MemoryBuiltins.cpp
nikic: > So, you don't want us to read the size/alignment from the table for now?

Yes

> If so, what would need to be added to the IR? Basically attributes for those two properties?

Yes

> Do we express them for allocations? 

Yes

> Then I could just copy the handling.

Basically. Reusing the existing allocsize/allocalign attributes should be possible, but their semantics are going to be completely different on deallocation functions, so that needs appropriate LangRef wording and inference support in BuildLibCalls.

To avoid blocking this PR on that rabbit hole, I'd recommend just dropping the size/align args for dealloc, and do them later.

nikic (commented):
It looks like this PR contains lots of unrelated reformatting now -- can you please land it separately?

llvm/include/llvm/Analysis/MemoryBuiltins.h
jdoerfert: getAllocationFamily returns an optional<StringRef>. I think chaning that needs a seperate PR since it's also used as such elsewhere.

llvm/lib/Analysis/MemoryBuiltins.cpp
jdoerfert: Good catch.

jdoerfert:
> It looks like this PR contains lots of unrelated reformatting now -- can you please land it separately?

I opened a PR for the code cleanup (https://github.com/llvm/llvm-project/pull/205205) and removed it here. Also squashed the commits and addressed the comments, I think.

PR #197535, opened 2026-05-13
[Instrumentor] Add support for modules and globals

jdoerfert:
We can emit callbacks when a module is loaded/unloaded and before globals are initialized. Both happens in newly introduced constructors and destructors.


llvm/lib/Transforms/IPO/Instrumentor.cpp
```diff
@@ -304,6 +312,93 @@ bool InstrumentorImpl::instrumentFunction(Function &Fn) {
   return Changed;
 }
 
+bool InstrumentorImpl::instrumentModule() {
+  SmallVector<GlobalVariable *> Globals;
+  Globals.reserve(M.global_size());
+  for (GlobalVariable &GV : M.globals()) {
+    // llvm.metadata contains globals such as llvm.used
```
```suggestion
    // llvm.metadata contains globals such as llvm.used.
```

llvm/lib/Transforms/IPO/Instrumentor.cpp
```diff
@@ -304,6 +312,93 @@ bool InstrumentorImpl::instrumentFunction(Function &Fn) {
...
+      // Count epochs eagerly.
+      ++IIRB.Epoch;
+
+      Changed |= bool(IO->instrument(YtorPtr, IConf, IIRB, ICaches));
+      IIRB.returnAllocas();
+    }
+  }
+
+  for (auto Loc : {InstrumentationLocation::GLOBAL_PRE,
+                   InstrumentationLocation::GLOBAL_POST}) {
```
kevinsala, line 371: There's no support for global post in the rest of the code.

llvm/test/Instrumentation/Instrumentor/module_and_globals.ll
```diff
@@ -0,0 +1,64 @@
+; NOTE: Assertions have been autogenerated by utils/update_test_checks.py UTC_ARGS: --check-globals all --include-generated-funcs --version 5
+; RUN: opt < %s -passes=instrumentor -S | FileCheck %s
+
+@X = dso_local global i32 0, align 4
+@Y = internal constant i32 42, align 4
```
kevinsala, line 5: Add also a global declaration case.

kevinsala (approved):

PR #180917, opened 2026-02-11
[AggressiveInstCombine] POPCNT generation for bit-count pattern

rohitaggarwal007:
The proposal is to enhance LLVM by teaching it to recognize the pattern and replace it with the hardware POPCNT instruction.

#177109 has the first pattern of the popcnt fold.

PR #197536, opened 2026-05-13
[AggressiveInstCombine] Loosen some conditions in the popcount pattern

mshockwave:
This PR refines and loosens some condition regarding the last AND mask in the popcount pattern introduced by #180917 . 
More specifically, this AND mask only needs to fulfill two conditions:
  1. All ones for the lower NumLenBits-bits, where NumLenBits is the number of bits needed to represent the size of the integer
  2. Zeros from bit 8 and onward

Rest of the bits can have arbitrary values.

The compute known bits infrastructure is supposed to reason about this but it was depth-limited in this particular case.

### DIFF
--- a/llvm/lib/Transforms/AggressiveInstCombine/AggressiveInstCombine.cpp
+++ b/llvm/lib/Transforms/AggressiveInstCombine/AggressiveInstCombine.cpp
@@ -623,8 +623,7 @@ static bool tryToRecognizePopCount2n3(Instruction &I) {
   // Condition (1) is straightforward. The reason behind condition
   // (2) is that we don't care any 8-bit chunks but the first one
   // in the original divide-and-conquer algorithm.
-  if (!MaskRes->getLoBits(NumLenBits).isMask(NumLenBits) ||
-      !MaskRes->getHiBits(Len - 8).isZero())
+  if (MaskRes->countTrailingOnes() < NumLenBits || MaskRes->getActiveBits() > 8)
     return false;
 
   Value *Add2;
--- a/llvm/lib/Transforms/AggressiveInstCombine/AggressiveInstCombine.cpp
+++ b/llvm/lib/Transforms/AggressiveInstCombine/AggressiveInstCombine.cpp
@@ -611,10 +611,20 @@ static bool tryToRecognizePopCount2n3(Instruction &I) {
   APInt Mask33 = APInt::getSplat(Len, APInt(8, 0x33));
   APInt Mask0F = APInt::getSplat(Len, APInt(8, 0x0F));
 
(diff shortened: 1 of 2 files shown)


llvm/test/Transforms/AggressiveInstCombine/popcount.ll
```diff
@@ -1413,6 +418,73 @@ define i64 @popcnt2_64(i64 noundef %0) {
   ret i64 %18
 }
 
+define i64 @popcnt2_64_large_mask(i64 noundef %0) {
```
mshockwave, line 421: Alive2: https://alive2.llvm.org/ce/z/cdXKrW

llvm/test/Transforms/AggressiveInstCombine/popcount.ll
```diff
@@ -1365,6 +286,90 @@ define i32 @popcnt3_32(i32 noundef %0) {
...
+  %11 = and i32 %10, 252645135
+  %12 = lshr i32 %11, 8
+  %13 = add nuw nsw i32 %12, %11
+  %14 = lshr i32 %13, 16
+  %15 = add nuw nsw i32 %14, %13
+  %16 = and i32 %15, 255
+  ret i32 %16
+}
+
+define i32 @popcnt3_32_large_mask2(i32 noundef %0) {
```
mshockwave, line 312: Alive2: https://alive2.llvm.org/ce/z/aRzqaA

llvm/lib/Transforms/AggressiveInstCombine/AggressiveInstCombine.cpp
```diff
@@ -740,10 +611,20 @@ static bool tryToRecognizePopCount2n3(Instruction &I) {
...
+    return false;
+  // Number of bits needed to represent Len.
+  unsigned NumLenBits = Log2_32(Len) + 1;
+  // The "mask" here really only needs to fulfill two conditions:
+  // (1) All ones for the lower NumLenBits-bits
+  // (2) Zeros from bit 8 and onward.
+  // Condition (1) is straightforward. The reason behind condition
+  // (2) is that we don't care any 8-bit chunks but the first one
+  // in the original divide-and-conquer algorithm.
+  if (!MaskRes->getLoBits(NumLenBits).isMask(NumLenBits) ||
```
topperc: MaskRes->countTrailingOnes() < NumLenBits || MaskRes->countActiveBits() > 8

mshockwave: yeah that's a better way. It's fixed now

topperc (approved):
LGTM

PR #197537, opened 2026-05-13
[libc] Add LLVM_LIBC_ENABLE_EXPERIMENTAL_ENTRYPOINTS CMake flag

kaladron:
Adds a new CMake option, OFF by default, to gate entrypoints with known-incomplete implementations. This lets developers build and test partially-implemented functions without exposing them to production users.

The motivating case is `sysconf`, which only handles three of the required `_SC_*` constants (`_SC_PAGESIZE`, `_SC_NPROCESSORS_CONF`, `_SC_NPROCESSORS_ONLN`) and returns `EINVAL` for everything else. Functions like this are useful to have in a build for testing progress, but shouldn't be part of a default full build until the implementation is complete.

Changes:
- `libc/CMakeLists.txt`: adds `option(LLVM_LIBC_ENABLE_EXPERIMENTAL_ENTRYPOINTS ... OFF)`
- `libc/cmake/modules/LLVMLibCCompileOptionRules.cmake`: propagates `-DLIBC_EXPERIMENTAL_ENTRYPOINTS` when ON
- `libc/cmake/modules/LLVMLibCTestRules.cmake`: same for test compile options
- `libc/config/linux/{x86_64,aarch64,riscv}/entrypoints.txt`: moves `sysconf` behind the new flag

The flag does not require `LLVM_LIBC_FULL_BUILD` since overlay builds may also have incomplete entrypoints that benefit from this gating.

### DIFF
--- a/libc/cmake/modules/LLVMLibCCompileOptionRules.cmake
+++ b/libc/cmake/modules/LLVMLibCCompileOptionRules.cmake
@@ -253,10 +253,6 @@ function(_get_common_compile_options output_var flags)
       endif()
     endif()
 
-    if(LLVM_LIBC_ENABLE_EXPERIMENTAL_ENTRYPOINTS)
-      list(APPEND compile_options "-DLIBC_EXPERIMENTAL_ENTRYPOINTS")
-    endif()
-
     if(LIBC_COMPILER_HAS_FIXED_POINT)
       list(APPEND compile_options "-ffixed-point")
     endif()
--- a/libc/cmake/modules/LLVMLibCCompileOptionRules.cmake
+++ b/libc/cmake/modules/LLVMLibCCompileOptionRules.cmake
@@ -253,6 +253,10 @@ function(_get_common_compile_options output_var flags)
       endif()
     endif()
 
+    if(LLVM_LIBC_ENABLE_EXPERIMENTAL_ENTRYPOINTS)
+      list(APPEND compile_options "-DLIBC_EXPERIMENTAL_ENTRYPOINTS")
+    endif()
+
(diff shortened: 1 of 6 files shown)


libc/config/linux/x86_64/entrypoints.txt
```diff
@@ -1489,6 +1488,16 @@ if(LLVM_LIBC_FULL_BUILD)
   )
 endif()
 
+# Entrypoints with known-incomplete implementations. Only enabled when
```
vonosmas: I thikn the comment here (and in other entrypoints files) are repetitive (to docs for this option, and to CMake logic) and may be omitted.

libc/cmake/modules/LLVMLibCCompileOptionRules.cmake
```diff
@@ -253,6 +253,10 @@ function(_get_common_compile_options output_var flags)
       endif()
     endif()
 
+    if(LLVM_LIBC_ENABLE_EXPERIMENTAL_ENTRYPOINTS)
+      list(APPEND compile_options "-DLIBC_EXPERIMENTAL_ENTRYPOINTS")
```
vonosmas: Why do we need compile-time flags for these?

kaladron:
Hi Michael!  You and I had talked about doing something like this for things like sysconf and the regex patches.  Can you please take a look and see if it meets what you're thinking?

@vonosmas - My apologies.  I actually created the PR earlier than planned.

vonosmas (approved):

michaelrj-google (approved):
LGTM, we should probably update the build instructions on the website to mention this but that can be a followup.

PR #197538, opened 2026-05-13
[clang-format] Fix parsing of goto labels

### LINKED ISSUE #196662: [clang-format] Incorrectly attaches a compound-statement brace to a goto label
## Summary

When a goto label is followed by a scoped compound statement, clang-format formats the
opening brace as if it belongs syntactically to the label:

```cpp
cleanup: {
```

This is valid C++, but it is surprising and makes the label look like a special labeled
block construct rather than a normal goto label followed by a statement. There does not
appear to be a style option for ordinary labels equivalent to `BraceWrapping.AfterCaseLabel`.

## Version

Originally observed with an Ubuntu clang-format package:

```text
Ubuntu clang-format version 22.1.3 (++20260402073256+4250a0fc5de9-1~exp1~20260402073413.57)
```

Also reproduced with a local build from llvm-project `main`:

```text
clang-format version 23.0.0git (https://github.com/llvm/llvm-project a47d3636f953870d96fb6cc68817365fdad2f9fe)
```

## Minimal `.clang-format`

```yaml
Language: Cpp
BasedOnStyle: Google
IndentWidth: 2
BreakBeforeBraces: Attach
AllowShortBlocksOnASingleLine: Never
```

## Input

```cpp
void f(bool failed) {
  if (failed) goto cleanup;

cleanup:
  {
    int value = 1;
    value += 1;
    (void)value;
  }
}
```

## Command

```sh
clang-format --style=file test.cc
```

## Actual Output

```cpp
void f(bool failed) {
  if (failed) goto cleanup;

cleanup: {
  int value = 1;
  value += 1;
  (void)value;
}
}
```

## Expected Output

I expected clang-format to preserve the ordinary-label structure and keep the compound
statement as the labeled statement on the following line:

```cpp
void f(bool failed) {
  if (failed) goto cleanup;

cleanup:
  {
    int value = 1;
    value += 1;
    (void)value;
  }
}
```

## Analysis

The issue does not appear to depend on `IndentGotoLabels`. It reproduces when
`IndentGotoLabels` is omitted, and with all supported values: `NoIndent`, `OuterIndent`,
`InnerIndent`, and `HalfIndent`. The backwards-compatible boolean spellings also
reproduce: `false` maps to `NoIndent`, and `true` maps to `OuterIndent`.

The issue seems that clang-format is treating the goto label as a case label.

Adding this option prevents the brace from attaching to the goto label:

```yaml
IndentCaseBlocks: true
```

It forces case label to wrap and indent. That seems achieved the expected outcome on
the goto label. But this would also prevent using `BraceWrapping.AfterCaseLabel: false`.

HazardyKnusperkeks:
Fixes #196662.

### DIFF
--- a/clang/lib/Format/UnwrappedLineParser.cpp
+++ b/clang/lib/Format/UnwrappedLineParser.cpp
@@ -3376,6 +3376,7 @@ void UnwrappedLineParser::parseDoWhile() {
 
 void UnwrappedLineParser::parseLabel(
     FormatStyle::IndentGotoLabelStyle IndentGotoLabels) {
+  const bool IsGotoLabel = FormatTok->is(TT_GotoLabelColon);
   nextToken();
   unsigned OldLineLevel = Line->Level;
 
@@ -3392,8 +3393,8 @@ void UnwrappedLineParser::parseLabel(
     break;
   }
 
-  if (!Style.IndentCaseBlocks && CommentsBeforeNextToken.empty() &&
-      FormatTok->is(tok::l_brace)) {
+  if (!IsGotoLabel && !Style.IndentCaseBlocks &&
+      CommentsBeforeNextToken.empty() && FormatTok->is(tok::l_brace)) {
 
     CompoundStatementIndenter Indenter(this, Line->Level,
                                        Style.BraceWrapping.AfterCaseLabel,
--- a/clang/lib/Format/UnwrappedLineParser.cpp
+++ b/clang/lib/Format/UnwrappedLineParser.cpp
@@ -3395,7 +3395,6 @@ void UnwrappedLineParser::parseLabel(
 
   if (!IsGotoLabel && !Style.IndentCaseBlocks &&
       CommentsBeforeNextToken.empty() && FormatTok->is(tok::l_brace)) {
-
     CompoundStatementIndenter Indenter(this, Line->Level,
                                        Style.BraceWrapping.AfterCaseLabel,
                                        Style.BraceWrapping.IndentBraces);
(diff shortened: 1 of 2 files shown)


clang/lib/Format/UnwrappedLineParser.cpp
```diff
@@ -3392,8 +3393,8 @@ void UnwrappedLineParser::parseLabel(
     break;
   }
 
-  if (!Style.IndentCaseBlocks && CommentsBeforeNextToken.empty() &&
-      FormatTok->is(tok::l_brace)) {
+  if (!IsGotoLabel && !Style.IndentCaseBlocks &&
+      CommentsBeforeNextToken.empty() && FormatTok->is(tok::l_brace)) {
```
owenca:
```suggestion
```

owenca (approved):

clang/unittests/Format/FormatTest.cpp
```diff
@@ -3168,6 +3173,25 @@ TEST_F(FormatTest, FormatsLabels) {
...
+               "    // Block\n"
+               "  }\n"
+               "}");
+  verifyFormat("void func() {\n"
+               "label:\n"
+               "  // Comment\n"
+               "  {\n"
+               "    // Block\n"
+               "  }\n"
+               "}");
```
owenca: Redundant (or irrelevant to this patch) because we never join a label with the comment below it anyway.

HazardyKnusperkeks:
```suggestion
```

PR #197539, opened 2026-05-13
[CIR] Allow local goto within cleanup regions

andykaylor:
Until now, CIR's FlattenCFG pass reported an NYI error any time a goto operation was found within a cleanup scope region. This change loosens that restriction to allow goto operations that transfer to another block within the same cleanup region. This case doesn't require any change in the cleanup scope flattening. It just has to be detected and ignored. The goto will be lowered as it is when no cleanup scope is present.

We are still reporting an NYI error in cases where a goto operation branches out of the cleanup scope. That will be implemented in a follow-up change.

### DIFF
--- a/clang/lib/CIR/Dialect/Transforms/FlattenCFG.cpp
+++ b/clang/lib/CIR/Dialect/Transforms/FlattenCFG.cpp
@@ -729,6 +729,27 @@ public:
     CleanupExit(mlir::Operation *op, int id) : exitOp(op), destinationId(id) {}
   };
 
+  // Determine whether a goto operation transfers control to a label that
+  // exists somewhere inside the given region (or any of its nested regions).
+  // Label names are unique within a function, so finding a matching cir.label
+  // inside the region implies that the goto definitely targets that label and
+  // therefore stays within the region. If no match is found, the goto either
+  // exits the region or its target is unknown; in either case the caller must
+  // treat it as exiting the region.
+  static bool gotoTargetsLabelInRegion(cir::GotoOp gotoOp,
+                                       mlir::Region &region) {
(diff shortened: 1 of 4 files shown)


clang/lib/CIR/Dialect/Transforms/FlattenCFG.cpp
```diff
@@ -757,14 +780,26 @@ class CIRCleanupScopeOpFlattening
...
     // Break and continue inside loops are handled by the loop itself.
     // Loops don't require special handling for nested switch or cleanup scopes
     // because break and continue never branch out of the loop.
     auto collectExitsInLoop = [&](mlir::Operation *loopOp) {
       loopOp->walk<mlir::WalkOrder::PreOrder>([&](mlir::Operation *nestedOp) {
-        if (isa<cir::ReturnOp, cir::GotoOp>(nestedOp))
+        if (isa<cir::ReturnOp>(nestedOp)) {
           exits.emplace_back(nestedOp, nextId++);
+        } else if (auto gotoOp = dyn_cast<cir::GotoOp>(nestedOp)) {
+          if (gotoExitsCleanup(gotoOp))
```
erichkeane: Could you do just `else if (gotoExitsCleanup(nestedOp))` above (and have it do the cast)?  Or is there more here we intend to see?

erichkeane (approved):
1 suggestion/nit, else lgtm.

AmrDeveloper (approved):
LGTM

clang/lib/CIR/Dialect/Transforms/FlattenCFG.cpp
andykaylor: I can do that.

PR #197540, opened 2026-05-13
[clang] model WASM alloca addrspaces in Sema

vtjnash:
Previously alloca addrspace for WASM local variables was implemented as a mandatory LLVM pass that unsafely mutates the IR in order to legalize it (WebAssemblyRefTypeMem2Local). The goal with this PR is to allow deleting that pass, by getting the alloca addrspace correctly modeled on these from the beginning. This will be important once alloca doesn't have an element type anymore, since it will be impossible then to perform this correction to the clang output at the LLVM level anymore.

To support that, this implements a slight change to the way alloca addrspace semantics are handled when making an alloca for a QualType: the alloca is created in the addrspace stated by Sema, unless Sema says to use the default addrspace, in which case it takes the default from the datalayout and addrspace casts that to default (previously it always used the default alloca addrspace for the allocation). The expected intent is that this lets frontends (or users) to select multiple address spaces for alloca with some special semantics. For example, it may be common for a GC to have a separate addrspace for GC-tracked alloca vs all other alloca. In WASM, this is the flat vs. local address space, for example.


clang/lib/Basic/Targets/SPIR.h
```diff
@@ -213,7 +215,10 @@ class LLVM_LIBRARY_VISIBILITY BaseSPIRTargetInfo : public TargetInfo {
   }
 
   void setAddressSpaceMap(bool DefaultIsGeneric) {
-    AddrSpaceMap = DefaultIsGeneric ? &SPIRDefIsGenMap : &SPIRDefIsPrivMap;
+    if (DefaultIsGeneric)
+      AddrSpaceMap = &SPIRDefIsGenMap;
+    else
+      AddrSpaceMap = &SPIRDefIsPrivMap;
```
arsenm, line 221: Unnecessary?

clang/lib/AST/TypePrinter.cpp
```diff
@@ -2713,6 +2713,8 @@ std::string Qualifiers::getAddrSpaceAsString(LangAS AS) {
     return "hlsl_push_constant";
   case LangAS::wasm_funcref:
     return "__funcref";
+  case LangAS::wasm_var:
+    return "__externref";
```
QuantumSegfault: ?

`wasm_var` is not `__externref`. Externref has it's own address space. Just a typo?

EDIT: I saw where it shows up in the tests, and yeah, `__externref` doesn't make sense. `__wasm_var` probably makes the most sense, same as the AS is already named internally.

clang/lib/Basic/TargetInfo.cpp
```diff
@@ -55,6 +55,7 @@ static const LangASMap FakeAddrSpaceMap = {
     18, // hlsl_output
     19, // hlsl_push_constant
     20, // wasm_funcref
+    21, // wasm_externref
```
QuantumSegfault: A bit confused on two things. 

- Why is `wasm_externef` being added rather than `wasm_var`? What is the purpose of `FakeAddrSpaceMap`
- What do the values mean? Because `wasm_funcref` is indeed 20, but `wasm_externref` is 10, not 21

clang/test/Sema/wasm-refs-and-tables.c
```diff
@@ -89,13 +89,13 @@ __externref_t func(__externref_t ref) {
...
-  1 >= table;                     // expected-error {{invalid operands to binary expression ('int' and '__attribute__((address_space(1))) __externref_t[0]')}}
-  table == other_table;           // expected-error {{invalid operands to binary expression ('__attribute__((address_space(1))) __externref_t[0]' and '__attribute__((address_space(1))) __externref_t[0]')}}
-  table !=- table;                // expected-error {{invalid argument type '__attribute__((address_space(1))) __externref_t *' to unary expression}}
-  !table;                         // expected-error {{invalid argument type '__attribute__((address_space(1))) __externref_t *' to unary expression}}
-  1 && table;                     // expected-error {{invalid operands to binary expression ('int' and '__attribute__((address_space(1))) __externref_t[0]')}}
-  table || 1;                     // expected-error {{invalid operands to binary expression ('__attribute__((address_space(1))) __externref_t[0]' and 'int')}}
+  table == 1;                     // expected-error {{invalid operands to binary expression ('__externref __externref_t[0]' and 'int')}}
```
QuantumSegfault: Yeah, I this shouldn't be `__externref` but perhaps `__wasm_var` (which covers both globals and locals).

clang/lib/Basic/Targets/SPIR.h
vtjnash: It causes compilers to notice that the size must match and error if I missed updating it

clang/lib/Basic/TargetInfo.cpp
vtjnash: This is the LangASMap (a unique enum), not the target platform addrspace (which is mapped separately)

QuantumSegfault: Okay, that makes sense.

But I'm still confused why this is marked `wasm_externref` and not `wasm_var`, since `wasm_var` is what you've been working in throughout the rest of the PR

vtjnash: Yes, looks like it is just the comment here that is wrong?

QuantumSegfault: I believe so, yes. If the values are distinct/arbitrary, it's just the comment that seems wrong given the rest of the context.

clang/lib/Basic/Targets/SPIR.h
arsenm: Actually, can de finally be rid of this DefaultIsGeneric mode switch with the recent address space changes? This is because OpenCL 1.x was never implemented properly

clang/lib/AST/ASTContext.cpp
```diff
@@ -15101,6 +15070,15 @@ LangAS ASTContext::getLangASForBuiltinAddressSpace(unsigned AS) const {
   if (LangOpts.CUDA)
     return getTargetInfo().getCUDABuiltinAddressSpace(AS);
 
+  if (getTargetInfo().getTriple().isWasm()) {
```
arsenm, line 15102: Why isn't this in getLangASFromTargetAS?

clang/lib/Basic/Targets/AMDGPU.cpp
```diff
@@ -218,7 +226,11 @@ void AMDGPUTargetInfo::fillValidCPUList(
 }
 
 void AMDGPUTargetInfo::setAddressSpaceMap(bool DefaultIsPrivate) {
-  AddrSpaceMap = DefaultIsPrivate ? &AMDGPUDefIsPrivMap : &AMDGPUDefIsGenMap;
+  // Use separate assignments so the compiler will enforce the size is correct.
+  if (DefaultIsPrivate)
+    AddrSpaceMap = &AMDGPUDefIsPrivMap;
+  else
+    AddrSpaceMap = &AMDGPUDefIsGenMap;
```
arsenm: Can you delete the DefaultIsPrivate case now?

clang/lib/Sema/SemaDecl.cpp
```diff
@@ -8926,12 +8922,45 @@ void Sema::CheckVariableDeclarationType(VarDecl *NewVD) {
...
+  }
+
   // Emit an error if an address space was applied to decl with local storage.
   // This includes arrays of objects with address space qualifiers, but not
   // automatic variables that point to other address spaces.
   // ISO/IEC TR 18037 S5.1.2
-  if (!getLangOpts().OpenCL && NewVD->hasLocalStorage() &&
-      T.getAddressSpace() != LangAS::Default) {
+  if (T.isWebAssemblyReferenceType() &&
+      Context.getTargetInfo().getTriple().isWasm() && !T->isArrayType()) {
```
arsenm: Shouldn't need the triple check?

clang/lib/AST/ASTContext.cpp
vtjnash: It looks like `getLangASFromTargetAS` doesn't appear to have knowledge of the Triple (it's a global function), so this seemed to be what was done for existing targets

PR #197541, opened 2026-05-13
[VPlan] Split out adding canonical IV recipes to separate transform.

fhahn:
Introduce canonical IV recipes after initial scalar transformations/simplifications. Conceptually it is a separate transformation and moving it later simplifies initial construction The canonical IV is only needed once we handle early exits/introduce regions.

This is needed to compute costs of scalar VPlans, where we need to compare the cost of the original loop control instructions.


llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
```diff
@@ -478,13 +478,14 @@ static void createLoopRegion(VPlan &Plan, VPBlockBase *HeaderVPB) {
...
 }
 
-// Add the necessary canonical IV and branch recipes required to control the
-// loop.
-static void addCanonicalIVRecipes(VPlan &Plan, VPBasicBlock *HeaderVPBB,
-                                  VPBasicBlock *LatchVPBB, Type *IdxTy,
-                                  DebugLoc DL) {
+void VPlanTransforms::addCanonicalIVRecipes(VPlan &Plan, DebugLoc DL) {
+  auto *HeaderVPBB = cast<VPBasicBlock>(
+      Plan.getEntry()->getSuccessors()[1]->getSingleSuccessor());
```
lukel97, line 483: I think we're doing this in a few places now, maybe we should add `findVectorHeader`/`findVectorLatch` helpers in VPlanConstruction

lukel97 (approved):
LGTM

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
fhahn: Will do separately, thanks!

PR #197544, opened 2026-05-13
[NFC][Analysis] Use `isa<ConstantPointerNull>` for null pointer checks

shiltian:
Make Analysis null pointer checks use `isa<ConstantPointerNull>` rather than generic null value checks (`isNullValue()`).

### DIFF
--- a/llvm/lib/Analysis/LazyValueInfo.cpp
+++ b/llvm/lib/Analysis/LazyValueInfo.cpp
@@ -2115,6 +2115,10 @@ Constant *LazyValueInfo::getPredicateAt(CmpInst::Predicate Pred, Value *V,
   // return it quickly. But this is only a fastpath, and falling
   // through would still be correct.
   const DataLayout &DL = CxtI->getDataLayout();
+  // NOTE: This check is meant to determine whether a pointer is semantically a
+  // null pointer, not just whether its value equals ConstantPointerNull. If the
+  // semantics of ConstantPointerNull change in the future, this should be
+  // updated to use a semantic check (e.g. isKnownNonNull).
   if (V->getType()->isPointerTy() && C->isNullValue() &&
       isKnownNonZero(V->stripPointerCastsSameRepresentation(), DL)) {
     Type *ResTy = CmpInst::makeCmpResultType(C->getType());
--- a/llvm/lib/Analysis/GlobalsModRef.cpp
+++ b/llvm/lib/Analysis/GlobalsModRef.cpp
@@ -21,6 +21,7 @@
 #include "llvm/Analysis/MemoryBuiltins.h"
 #include "llvm/Analysis/TargetLibraryInfo.h"
 #include "llvm/Analysis/ValueTracking.h"
+#include "llvm/IR/Constants.h"
 #include "llvm/IR/InstIterator.h"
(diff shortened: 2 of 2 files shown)


shiltian:
* #197544 👈
* `main`

<h2></h2>

llvm/lib/Analysis/ConstantFolding.cpp
```diff
@@ -1649,7 +1649,7 @@ Constant *llvm::ConstantFoldCastOperand(unsigned Opcode, Constant *C,
         APInt BaseOffset(BitWidth, 0);
         auto *Base = cast<Constant>(GEP->stripAndAccumulateConstantOffsets(
             DL, BaseOffset, /*AllowNonInbounds=*/true));
-        if (Base->isNullValue()) {
+        if (isa<ConstantPointerNull>(Base)) {
```
shiltian: I'm not sure about this one actually. What would be the semantics of GEP of a true `nullptr` after we change the semantic?

arsenm: For addressing calculations it's just the raw arithmetic on the value, so I think this one really wants is zero. Is -1 would be a different case

shiltian: Right, but what would be the semantics if a nullptr is -1 in the future in this case? do we simply not fold it or we treat it as UB?

nikic: It should just not fold. This case will generally work on a non-inbounds GEP, so not sure where UB would come from?

llvm/lib/Analysis/ConstantFolding.cpp
```diff
@@ -1093,7 +1093,7 @@ Constant *SymbolicallyEvaluateGEP(const GEPOperator *GEP,
     }
   }
 
-  if ((Ptr->isNullValue() || BaseIntVal != 0) &&
+  if ((isa<ConstantPointerNull>(Ptr) || BaseIntVal != 0) &&
```
nikic: This is wrong as well, this fold needs a 0 pointer, not a null pointer.

llvm/lib/Analysis/LazyValueInfo.cpp
```diff
@@ -2115,7 +2115,7 @@ Constant *LazyValueInfo::getPredicateAt(CmpInst::Predicate Pred, Value *V,
   // return it quickly. But this is only a fastpath, and falling
   // through would still be correct.
   const DataLayout &DL = CxtI->getDataLayout();
-  if (V->getType()->isPointerTy() && C->isNullValue() &&
+  if (V->getType()->isPointerTy() && isa<ConstantPointerNull>(C) &&
       isKnownNonZero(V->stripPointerCastsSameRepresentation(), DL)) {
```
nikic, line 2123: The change here is semantically correct, but worth noting that the combination with isKnownNonZero() is not -- I assume we're going to add a isKnownNonNull() query for the cases that are actually interested in null?

shiltian: > I assume we're going to add a isKnownNonNull() query for the cases that are actually interested in null?

Yes, that's the plan.

arsenm (approved):

PR #197548, opened 2026-05-13
[lldb] Add missing calls to SetStatus in CommandObjectPlatform

JDevlieghere:
Add missing calls to CommandReturnObject::SetStatus in CommandObjectPlatform. I replaced some calls to AppendMessageWithFormatv with AppendErrorWithFormatv because the latter implicitly sets the status.


JDevlieghere:
This should fix the assert that triggered https://github.com/llvm/llvm-project/pull/197367

bulbazord (approved):

PR #197550, opened 2026-05-13
[ConstraintElim] use icmpLike when creating facts to support trunc nuw to i1

andjo403:
Adding `trunc nuw %x to i1` support in facts the same way as `icmp ne %x, 0`


andjo403:
do not show any improvements in llvm-opt-benchmark now but with https://github.com/llvm/llvm-project/pull/184182 there is a 5 files that improve.

dtcxzyw (approved):
LGTM.

PR #197551, opened 2026-05-13
[LifetimeSafety] Fix a crash in lifetimbound verification on assignment methods 

NeKon69:
The problem was that `implicitObjectParamIsLifetimeBound` could also return true if the function we are analyzing is an assignment operator with a few conditions.
https://github.com/llvm/llvm-project/blob/67284454ae951d1c9702f0ac784447eede59386d/clang/lib/Analysis/LifetimeSafety/LifetimeAnnotations.cpp#L33-L49

Since there would be nowhere to extract the attribute from nonetheless, I think we should ignore that check.

### DIFF
--- a/clang/test/Sema/warn-lifetime-safety-lifetimebound.cpp
+++ b/clang/test/Sema/warn-lifetime-safety-lifetimebound.cpp
@@ -1,4 +1,4 @@
-// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-lifetimebound-violation -verify %s
+// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-lifetimebound-violation -Wno-dangling -verify %s
 
 #include "Inputs/lifetime-analysis.h"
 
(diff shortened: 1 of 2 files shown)


Xazax-hun (approved):

clang/test/Sema/warn-lifetime-safety-lifetimebound.cpp
```diff
@@ -1,4 +1,4 @@
-// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-lifetimebound-violation -verify %s
+// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-lifetimebound-violation -Wno-dangling -verify %s
```
Xazax-hun, line 1: I wonder if it is better to add the expected warnings rather than turning them off. But I do not have a strong opinion.

NeKon69: Do we really need the warnings from `-Wdangling` analysis? From what I've observed all other tests turn it off too.

Xazax-hun: I am fine landing this as is.

NeKon69: Let's merge this then!

NeKon69:
Oh, just so you know, I don't have commit access, so will you please merge this for me..?

usx95 (approved):

PR #197555, opened 2026-05-13
[CIR] Store offset instead of member index in DataMemberAttr (NFC)

adams381:
The pre-pass DataMemberAttr stores a symbolic field index into its
classTy record.  That representation is fine for `&Point::x` but
cannot express the value an upstream cast handler would need to
produce for

    struct Base    { int x; };
    struct Derived : Base { double y; };
    int Derived::*p = &Derived::x;

There is no field index into Derived's flat CIR field list that
points at the base-class member `x` -- Derived's record at index 0
is the Base subobject, not `int`.  Today
`ConstExprEmitter::VisitCastExpr` errorNYIs on
`CK_BaseToDerivedMemberPointer` before this code path is reached;
a follow-up PR will implement that cast handler.  This refactor
prepares the attribute schema so that follow-up can succeed.

Switch DataMemberAttr from storing `std::optional<unsigned>
member_index` to storing `std::optional<int64_t> offset` -- the
resolved byte offset of the member from the start of classTy.
Both construction sites already have the AST in hand, so they
compute the offset via `ASTContext::getFieldOffset` /
`toCharUnitsFromBits` -- the same thing
`ItaniumCXXABI::EmitMemberPointer` does in classic CodeGen.  The
lowering pass becomes a trivial passthrough (null -> -1, otherwise
return the stored offset), and its `mlir::DataLayout` parameter is
dropped from `CIRCXXABI::lowerDataMemberConstant` since it's no
longer needed.

The verifier shrinks to a non-negative-offset sanity check.  The
old verifier's "member type matches the field at member_index" and
"member_index < numFields" checks don't carry over: with a byte
offset, the analogous check would be "is there really a field of
type memberTy at byte offset N in classTy?", which requires a
record-layout walk that can't work for the forward-declared-record
case (`pointer-to-data-member.cpp:209-262`) and duplicates work
that CIRGen already did when computing the offset.  Hand-written
CIR can therefore construct a DataMemberAttr whose offset doesn't
correspond to any real field in classTy; CIRGen-emitted CIR
remains consistent by construction.

Test impact is mechanical: pre-pass pretty-print changes
`#cir.data_member<N>` -> `#cir.data_member<offset = BYTES>` in
`pointer-to-data-member.cpp` and `nonzeroinit-struct.cpp`.  Three
invalid-cir checks ("member type doesn't match", "incomplete
class", "index out of range") no longer apply and are replaced
with a single "negative offset" verifier check.  CIR-AFTER, LLVM,
and OGCG expectations are unchanged.

### DIFF
--- a/clang/include/clang/CIR/Dialect/IR/CIRAttrs.td
+++ b/clang/include/clang/CIR/Dialect/IR/CIRAttrs.td
@@ -543,9 +543,12 @@ def CIR_DataMemberAttr : CIR_ValueLikeAttr<"DataMember", "data_member"> {
     A data member attribute is a literal attribute that represents a constant
     pointer-to-data-member value.
 
-    The `member_index` parameter represents the index of the pointed-to member
-    within its containing record. It is an optional parameter; lack of this
-    parameter indicates a null pointer-to-data-member value.
+    The `member_index` parameter is the index of the pointed-to member within
+    its containing record (the attribute's `classTy`).  It is an optional
+    parameter; lack of this parameter indicates a null pointer-to-data-member
+    value.  A non-null pointer to a member of a base class subobject in a
+    derived class is represented by a leaf attribute in the base class type
+    wrapped in `cir.derived_data_member` operations.
 
     Example:
     ```
--- a/clang/lib/CIR/CodeGen/CIRGenModule.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenModule.cpp
@@ -34,10 +34,8 @@
 #include "clang/CIR/Dialect/IR/CIRTypes.h"
 #include "clang/CIR/Interfaces/CIROpInterfaces.h"
 #include "clang/CIR/MissingFeatures.h"
-#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/StringRef.h"
-#include "llvm/Support/raw_ostream.h"
 
 #include "CIRGenFunctionInfo.h"
 #include "TargetInfo.h"
@@ -806,47 +804,7 @@ bool CIRGenModule::getCPUAndFeaturesAttributes(
   const auto *tc = fd ? fd->getAttr<TargetClonesAttr>() : nullptr;
   bool addedAttr = false;
   if (td || tv || sd || tc) {
-    llvm::StringMap<bool> featureMap;
-    astContext.getFunctionFeatureMap(featureMap, gd);
-
-    // Now add the target-cpu and target-features to the function.
-    // While we populated the feature map above, we still need to
-    // get and parse the target/target_clones attribute so we can
-    // get the cpu for the function.
-    llvm::StringRef featureStr = td ? td->getFeaturesStr() : llvm::StringRef();
-    if (tc && (getTriple().isOSAIX() || getTriple().isX86()))
-      featureStr = tc->getFeatureStr(gd.getMultiVersionIndex());
-    if (!featureStr.empty()) {
-      clang::ParsedTargetAttr parsedAttr =
-          getTarget().parseTargetAttr(featureStr);
-      if (!parsedAttr.CPU.empty() &&
-          getTarget().isValidCPUName(parsedAttr.CPU)) {
-        targetCPU = parsedAttr.CPU;
-        tuneCPU = ""; // Clear the tune CPU.
-      }
-      if (!parsedAttr.Tune.empty() &&
-          getTarget().isValidCPUName(parsedAttr.Tune))
-        tuneCPU = parsedAttr.Tune;
-    }
-
-    if (sd) {
-      // Apply the given CPU name as the 'tune-cpu' so that the optimizer can
-      // favor this processor.
-      tuneCPU = sd->getCPUName(gd.getMultiVersionIndex())->getName();
-    }
-
-    // For AMDGPU, only emit delta features (features that differ from the
-    // target CPU's defaults). Other targets might want to follow a similar
-    // pattern.
-    if (getTarget().getTriple().isAMDGPU()) {
-      features = getFeatureDeltaFromDefault(*this, targetCPU, featureMap);
-    } else {
-      // Produce the canonical string for this set of features.
-      features.reserve(features.size() + featureMap.size());
-      for (const auto &entry : featureMap)
-        features.push_back((entry.getValue() ? "+" : "-") +
-                           entry.getKey().str());
-    }
+    assert(!cir::MissingFeatures::opFuncMultiVersioning());
   } else {
     // Just add the existing target cpu and target features to the function.
     if (setTargetFeatures && getTarget().getTriple().isAMDGPU()) {
@@ -1244,6 +1202,7 @@ CIRGenModule::getOrCreateCIRGlobal(StringRef mangledName, mlir::Type ty,
       if (const SectionAttr *sa = d->getAttr<SectionAttr>())
         gv.setSectionAttr(builder.getStringAttr(sa->getName()));
     }
+    gv.setGlobalVisibility(getGlobalVisibilityAttrFromDecl(d).getValue());
 
     // Handle XCore specific ABI requirements.
     if (getTriple().getArch() == llvm::Triple::xcore)
@@ -2205,11 +2164,9 @@ LangAS CIRGenModule::getLangTempAllocaAddressSpace() const {
   if (getLangOpts().CUDAIsDevice)
     return LangAS::Default;
 
-  if (getLangOpts().OpenMP && getLangOpts().OpenMPIsTargetDevice)
-    assert(!cir::MissingFeatures::openMP());
-  if (getLangOpts().SYCLIsDevice)
-    errorNYI("SYCL temp address space");
-
+  if (getLangOpts().SYCLIsDevice ||
+      (getLangOpts().OpenMP && getLangOpts().OpenMPIsTargetDevice))
+    errorNYI("SYCL or OpenMP temp address space");
   return LangAS::Default;
 }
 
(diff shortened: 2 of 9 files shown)


erichkeane (commented):
I don't believe this is the correct answer here, this is resulting in an inappropriate loss of fidelity in the dialect.

We already know the type layouts, so we SHOULD be able to figure out the correct 'field' indexed.  

We should be able to calculate it based on the types, so i would expect a type-conversion + data-member would be able to reconstitute this information.

adams381:
@erichkeane The index-based form has no valid value for `int Derived::*p = &Derived::x;`: the field index is taken in the field's declaring class (Base, where `x` is index 0 mapping to `int`), but the attribute's `classTy` is Derived, whose flat CIR field list has the Base subobject at index 0 (type `rec_Base`, not `int`).  The pre-PR-1 verifier's "memberType matches the field at memberIndex" check rejects every constructed attribute for this case -- that's why `ConstExprEmitter::VisitCastExpr` errorNYIs on `CK_BaseToDerivedMemberPointer` at line 939 today instead of producing a valid DataMemberAttr.

On the fidelity claim specifically: `(memberType, offset, classTy)` is invertible -- walk classTy's layout (including base subobjects) for a field of type `memberType` at the given byte offset, and you recover a unique FieldDecl.  Tested mentally for single inheritance, multiple inheritance, unions, and empty bases; in every case `(type, offset)` picks out one field.  No information is lost; it's just stored in a form that requires a layout lookup rather than being a direct index.  Classic CodeGen converged on the byte offset for this same reason -- `ItaniumCXXABI::EmitMemberDataPointer` produces an `i64`, and `EmitMemberPointerConversion` adds the path adjustment as a constant int.  The CIR pre-pass form now matches what was already coming out of CXXABILowering.

A wrapper attribute could express the cast (store the index in Base, wrap with a path that re-interprets it in Derived).  But it's strictly more dialect surface: two attributes that have to agree on the path's anchor classes, plus a path representation, plus a verifier walk over the path, plus a lowering walk over the path.  I don't see what it buys us over the offset.

erichkeane:
> @erichkeane The index-based form has no valid value for `int Derived::*p = &Derived::x;`: the field index is taken in the field's declaring class (Base, where `x` is index 0 mapping to `int`), but the attribute's `classTy` is Derived, whose flat CIR field list has the Base subobject at index 0 (type `rec_Base`, not `int`). The pre-PR-1 verifier's "memberType matches the field at memberIndex" check rejects every constructed attribute for this case -- that's why `ConstExprEmitter::VisitCastExpr` errorNYIs on `CK_BaseToDerivedMemberPointer` at line 939 today instead of producing a valid DataMemberAttr.
> 

Right, I get that.  But if 'base' is the 1st object, [0,0] would be the index of 'x'.  So rather than making this an offset (which has a fragility, but also a lack of expressivity/clarity for identifying these by types in passes), we should make this closer to a GEP perhaps.

> On the fidelity claim specifically: `(memberType, offset, classTy)` is invertible -- walk classTy's layout (including base subobjects) for a field of type `memberType` at the given byte offset, and you recover a unique FieldDecl. Tested mentally for single inheritance, multiple inheritance, unions, and empty bases; in every case `(type, offset)` picks out one field. No information is lost; it's just stored in a form that requires a layout lookup rather than being a direct index. Classic CodeGen converged on the byte offset for this same reason -- `ItaniumCXXABI::EmitMemberDataPointer` produces an `i64`, and `EmitMemberPointerConversion` adds the path adjustment as a constant int. The CIR pre-pass form now matches what was already coming out of CXXABILowering.
> 

A pass shouldn't have to do that work/be layout aware.  The point of CIR is to be expressive-as-possible in a way that is close to the language.  Forcing a pass to have to figure out layout/etc is improper, and something that LLVM does because it doesn't have an alternative.

> A wrapper attribute could express the cast (store the index in Base, wrap with a path that re-interprets it in Derived). But it's strictly more dialect surface: two attributes that have to agree on the path's anchor classes, plus a path representation, plus a verifier walk over the path, plus a lowering walk over the path. I don't see what it buys us over the offset.

IMO, the idea of storing the 'path' that I have above (like a GEP does), is probably more appropriate.  WE can express that as a series of indexes-as-necessary (and have a verification that ensures that we don't try to 'dig' into a non-record type), or as an attribute.

andykaylor:
> Right, I get that. But if 'base' is the 1st object, [0,0] would be the index of 'x'. So rather than making this an offset (which has a fragility, but also a lack of expressivity/clarity for identifying these by types in passes), we should make this closer to a GEP perhaps.

I like this idea. It preserves the index-based representation, while allowing us to use a meaningful representation for members in base classes, but does `fieldDecl` in `emitMemberPointerConstant` give us enough information to build this?

erichkeane:
> > Right, I get that. But if 'base' is the 1st object, [0,0] would be the index of 'x'. So rather than making this an offset (which has a fragility, but also a lack of expressivity/clarity for identifying these by types in passes), we should make this closer to a GEP perhaps.
> 
> I like this idea. It preserves the index-based representation, while allowing us to use a meaningful representation for members in base classes, but does `fieldDecl` in `emitMemberPointerConstant` give us enough information to build this?

We MIGHT actually need it be an operation/attribute inside of the other one, but at that point, I would think so?

erichkeane:
I realized (now that I'm thinking straight :D) that this is actually a lossy conversion, and I'm kind of doubly against this now.

This in particular:
>On the fidelity claim specifically: (memberType, offset, classTy) is invertible
Is completely false.  That is not invertible since Unions, EBO, and No-unique-address exist.  Multiple objects can appear at the same address with the same type, and this will lose the information.

Losing this much fidelity is such a negative for CIR that I am even more sure we need to be doing index chains instead of offsets.  While our current layout engine is unfortunately copy/pasted from classic codegen (and might not YET distinguish between some of the above), it is vital that we leave spend the time making sure we're not making the changes to the layout engine harder for the purposes of making this immediately easier.

clang/include/clang/CIR/Dialect/IR/CIRAttrs.td
```diff
@@ -570,13 +578,11 @@ def CIR_DataMemberAttr : CIR_ValueLikeAttr<"DataMember", "data_member"> {
 
   let genVerifyDecl = 1;
 
-  let assemblyFormat = [{
-    `<` ($member_index^):(`null`)? `>`
-  }];
+  let hasCustomAssemblyFormat = 1;
```
erichkeane: Why is this?  Does DenseI32ArrayAttr not have a printer?   Or does it look 'bad'?

clang/test/CIR/IR/invalid-data-member.cir
```diff
@@ -2,29 +2,12 @@
...
-// -----
-
-!u16i = !cir.int<u, 16>
-!u32i = !cir.int<u, 32>
 !struct1 = !cir.record<struct "Struct1" {!u16i, !u32i}>
 
-// expected-error@+1 {{member index of a #cir.data_member attribute is out of range}}
-#invalid_member_ty = #cir.data_member<2> : !cir.data_member<!u32i in !struct1>
+// expected-error@+1 {{member path index of a #cir.data_member attribute is out of range}}
+#out_of_range_path = #cir.data_member<[99]> : !cir.data_member<!u32i in !struct1>
```
erichkeane: Can we have some tests here for multiple indexes and out of range? 

Also, we need a test in the NOT invalid data-member.cir that shows that this parses/prints correctly.

clang/lib/CIR/Dialect/IR/CIRAttrs.cpp
```diff
@@ -447,18 +447,91 @@ DataMemberAttr::verify(function_ref<InFlightDiagnostic()> emitError,
            << "incomplete 'cir.record' cannot be used to build a non-null "
               "data member pointer";
 
-  unsigned memberIndexValue = memberIndex.value();
-  if (memberIndexValue >= recTy.getNumElements())
+  llvm::ArrayRef<int32_t> path = memberPath.asArrayRef();
+  if (path.empty())
     return emitError()
-           << "member index of a #cir.data_member attribute is out of range";
+           << "member path of a #cir.data_member attribute must not be empty";
```
erichkeane: We need a test for this.

clang/lib/CIR/Dialect/IR/CIRAttrs.cpp
```diff
@@ -447,18 +447,91 @@ DataMemberAttr::verify(function_ref<InFlightDiagnostic()> emitError,
...
-  mlir::Type memberTy = recTy.getMembers()[memberIndexValue];
-  if (memberTy != ty.getMemberTy())
-    return emitError()
-           << "member type of a #cir.data_member attribute must match the "
-              "attribute type";
+  cir::RecordType currentTy = recTy;
+  for (auto [pathIdx, idx] : llvm::enumerate(path)) {
+    if (idx < 0)
+      return emitError()
+             << "member path index of a #cir.data_member attribute must be "
```
erichkeane: I don't see a test for this either.

clang/lib/CIR/Dialect/IR/CIRAttrs.cpp
```diff
@@ -447,18 +447,91 @@ DataMemberAttr::verify(function_ref<InFlightDiagnostic()> emitError,
...
-  return success();
+    if (static_cast<unsigned>(idx) >= currentTy.getNumElements())
+      return emitError()
+             << "member path index of a #cir.data_member attribute is out of "
+                "range";
+
+    mlir::Type memberTy = currentTy.getMembers()[idx];
+    if (pathIdx + 1 == path.size()) {
+      if (memberTy != ty.getMemberTy())
+        return emitError()
```
erichkeane: Can we get some tests to make sure we handle the N'th one right?

clang/lib/CIR/Dialect/IR/CIRAttrs.cpp
```diff
@@ -447,18 +447,91 @@ DataMemberAttr::verify(function_ref<InFlightDiagnostic()> emitError,
...
+               << "member type of a #cir.data_member attribute must match the "
+                  "attribute type";
+      return success();
+    }
+
+    auto nestedRecTy = mlir::dyn_cast<cir::RecordType>(memberTy);
+    if (!nestedRecTy)
+      return emitError()
+             << "member path through a #cir.data_member attribute must select "
```
erichkeane: Same here.

clang/lib/CIR/Dialect/IR/CIRAttrs.cpp
```diff
@@ -447,18 +447,91 @@ DataMemberAttr::verify(function_ref<InFlightDiagnostic()> emitError,
...
+      return {};
+    return parser.getChecked<DataMemberAttr>(parser.getCurrentLocation(), ty);
+  }
+
+  if (parser.parseLSquare().failed())
+    return {};
+
+  llvm::SmallVector<int32_t> path;
+  auto parsePathIndex = [&]() -> mlir::ParseResult {
+    int64_t index;
```
erichkeane: Hmm.... This is complicated enough for such a mild advantage of DenseI32ArrayAttr (if that is what is causing this need).  We'd be better off choosing a different type instead if that is the case.

clang/lib/CIR/CodeGen/CIRGenModule.cpp
```diff
@@ -2194,8 +2232,12 @@ mlir::Value CIRGenModule::emitMemberPointerConstant(const UnaryOperator *e) {
   // Otherwise, a member data pointer.
   auto ty = mlir::cast<cir::DataMemberType>(convertType(e->getType()));
   const auto *fieldDecl = cast<FieldDecl>(decl);
-  return cir::ConstantOp::create(
-      builder, loc, builder.getDataMemberAttr(ty, fieldDecl->getFieldIndex()));
+  const auto *mpt = cast<MemberPointerType>(e->getType().getTypePtr());
+  const CXXRecordDecl *pointerClass = mpt->getMostRecentCXXRecordDecl();
+  llvm::SmallVector<unsigned, 4> path;
```
erichkeane:
```suggestion
  llvm::SmallVector<unsigned> path;
```

clang/lib/CIR/Dialect/IR/CIRAttrs.cpp
adams381: Agreed — the path was overkill for what DataMemberAttr needs to represent. Simplified in the follow-up commit: replaced `DenseI32ArrayAttr member_path` with a single `unsigned member_index` (the field's index in its immediately-containing record). Cross-base-class pointers use `cir.base_data_member` / `cir.derived_data_member` cast ops rather than encoding the path in the attribute itself.

clang/include/clang/CIR/Dialect/IR/CIRAttrs.td
adams381: Moot after the redesign — `member_path` (DenseI32ArrayAttr) is gone. The attribute is back to a single `unsigned member_index` with the standard generated assembly format `<N>` / `<null>`.

clang/test/CIR/IR/invalid-data-member.cir
adams381: The path-based tests are moot after the redesign. Added in their place: a non-zero index roundtrip test in `data-member-attr.cir`, and invalid tests in `invalid-data-member.cir` covering wrong member type, incomplete struct, out-of-range index (99), and nested out-of-range.

clang/lib/CIR/Dialect/IR/CIRAttrs.cpp
adams381: Moot — the empty-path check is gone with the redesign. The current verifier checks `member_index >= recTy.getNumElements()` and wrong member type; both have tests in `invalid-data-member.cir`.

clang/lib/CIR/Dialect/IR/CIRAttrs.cpp
adams381: Moot — the negative-index check is gone with the redesign.

clang/lib/CIR/Dialect/IR/CIRAttrs.cpp
adams381: Moot — the multi-step path loop is gone. The N'th member case is covered by the out-of-range test (`#cir.data_member<99>`) and the non-zero index roundtrip test (`#cir.data_member<1>`).

clang/lib/CIR/Dialect/IR/CIRAttrs.cpp
adams381: Moot — nested record traversal is gone with the redesign.

clang/lib/CIR/CodeGen/CIRGenModule.cpp
adams381: Moot — there is no `SmallVector` for the member index in the current PR; the attribute stores a single `unsigned member_index`.

adams381:
Superseded by [#200854](https://github.com/llvm/llvm-project/pull/200854), which replaces the flat-offset approach with a GEP-style path and also covers inherited members and CK_ReinterpretMemberPointer.

PR #192973, opened 2026-04-20
[SPIR-V] Fix validation errors for function pointers with CodeSectionINTEL storage class

aobolensk:
CodeSectionINTEL pointers are not valid operands for PtrCastToGeneric/GenericCastToPtr (including inside OpSpecConstantOp)

PR #197556, opened 2026-05-13
[SPIRV] Allow casting between CodeSectionINTEL and Generic storage classes

sarnex:
In the previous versions of the SPV_INTEL_function_pointers [spec](https://github.com/intel/llvm/blob/sycl/sycl/doc/design/spirv-extensions/SPV_INTEL_function_pointers.asciidoc), casts between the CodeSectionINTEL storage class (used for function pointers) and the Generic storage class were illegal. 

The spec was updated a few months ago, and the new version allows the cast, specifying `CodeSectionIntel` as one of the overloaded storage classes that can be represented by Generic, alongside `WorkGroup`, etc. I also confirmed with a spec author that one of the intentions of the spec updates was to allow the cast.

Update the SPIR-V backend to allow the cast. This is basically required to use function pointers in real world use cases.

### DIFF
--- a/llvm/test/CodeGen/SPIRV/GlobalISel/fn-ptr-addrspacecast.ll
+++ b/llvm/test/CodeGen/SPIRV/GlobalISel/fn-ptr-addrspacecast.ll
@@ -1,6 +1,7 @@
 ; RUN: llc -O0 -verify-machineinstrs -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_INTEL_function_pointers %s -o - | FileCheck %s
 
 ; TODO: Update when spirv-val accepts casts from CodeSectionINTEL to Generic
+; https://github.com/KhronosGroup/SPIRV-Tools/issues/6700
 ; RUNx: %if spirv-tools %{ llc -O0 -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_INTEL_function_pointers %s -o - -filetype=obj | spirv-val %}
 
 define void @addrspacecast(ptr addrspace(9) %a) {
@@ -12,5 +13,17 @@ define void @addrspacecast(ptr addrspace(9) %a) {
 
   %res1 = addrspacecast ptr addrspace(9) %a to ptr addrspace(4)
   store i8 0, ptr addrspace(4) %res1
+  ret void
+}
+
+define void @addrspacecast_two(ptr addrspace(4) %b) {
+; CHECK: %[[#FnParam2:]] = OpFunctionParameter %[[#Int8GenericPtr]]
+; CHECK: OpGenericCastToPtr %[[#Int8FnPtr]] %[[#FnParam2]]
+
+  %res2 = addrspacecast ptr addrspace(4) %b to ptr addrspace(9)
+  %cmp = icmp eq ptr addrspace(9) %res2, null
+  %res3 = sext i1 %cmp to i8
+  store i8 %res3, ptr addrspace(4) %b
+
   ret void
 }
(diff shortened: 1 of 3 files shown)


MrSidims:
I believe https://github.com/llvm/llvm-project/pull/192973 supersedes this one. I assume answer to [this](https://github.com/llvm/llvm-project/pull/192973#discussion_r3124892714) question is "there are examples".

sarnex:
Based on a quick look I don't think the behavior proposed in the linked PR matches the latest version of the spec (which was updated 2026-02-12).

The fundamental idea in the linked PR seems to be that CodeSectionINTEL pointers are not valid operands for PtrCastToGeneric/GenericCastToPtr, but this is not true in the latest version of the spec. It was true in the previous version. The latest version removed this restriction and this PR attemps to implement the latest version of the spec.

MrSidims:
> I believe #192973 supersedes this one. I assume answer to [this](https://github.com/llvm/llvm-project/pull/192973#discussion_r3124892714) question is "there are examples".

Actually, they are a bit different. Anyway, I'm transferring my review privilege/responsibility to @aobolensk :)

maarquitos14 (approved):
LGTM, just a question: should this be done only for SPIRV1.6 and above?

sarnex:
Thanks for the review!

I don't think this change in particular interacts with newer SPIR-V versions in ways the old version of the extension didn't, so I'd expect any version requirement to be done as part of enabling the extension overall, which would happen before we hit this code. Also the extension spec says it only requires SPIR-V 1.0.

llvm/test/CodeGen/SPIRV/GlobalISel/fn-ptr-addrspacecast.ll
```diff
@@ -1,8 +1,15 @@
...
+
+; TODO: Update when spirv-val accepts casts from CodeSectionINTEL to Generic
+; RUNx: %if spirv-tools %{ llc -O0 -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_INTEL_function_pointers %s -o - -filetype=obj | spirv-val %}
 
 define void @addrspacecast(ptr addrspace(9) %a) {
-; CHECK: unable to legalize instruction: %{{.*}}:pid(p4) = G_ADDRSPACE_CAST %{{.*}}:pid(p9)
+; CHECK: %[[#Int8:]] = OpTypeInt 8 0
+; CHECK: %[[#Int8FnPtr:]] = OpTypePointer CodeSectionINTEL %[[#Int8]]
+; CHECK: %[[#Int8GenericPtr:]] = OpTypePointer Generic %[[#Int8]]
+; CHECK: %[[#FnParam:]] = OpFunctionParameter %[[#Int8FnPtr]]
```
aobolensk, line 11: Let's test `CodeSectionINTEL->Workgroup/Generic` conversion as well

llvm/test/CodeGen/SPIRV/GlobalISel/fn-ptr-addrspacecast.ll
```diff
@@ -1,8 +1,15 @@
-; RUN: not llc --global-isel %s -filetype=null 2>&1 | FileCheck %s
-target triple = "spirv64"
+; RUN: llc -O0 -verify-machineinstrs -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_INTEL_function_pointers %s -o - | FileCheck %s
+
+; TODO: Update when spirv-val accepts casts from CodeSectionINTEL to Generic
+; RUNx: %if spirv-tools %{ llc -O0 -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_INTEL_function_pointers %s -o - -filetype=obj | spirv-val %}
```
jmmartinez, line 5: Is there a github issue for this on spirv-tools side for this?

I'd like to add an issue on our side with a link to spirv-tools's one.

MrSidims (approved):

llvm/test/CodeGen/SPIRV/GlobalISel/fn-ptr-addrspacecast.ll
sarnex: Just made https://github.com/KhronosGroup/SPIRV-Tools/issues/6700 for this, will add a link in the lit test too, thanks!

llvm/test/CodeGen/SPIRV/GlobalISel/fn-ptr-addrspacecast.ll
sarnex: Added in latest commit, thanks!

aobolensk (approved):

PR #197560, opened 2026-05-13
[scudo] Add unit tests for optional config flags

sadafebrahimi:
This patch adds checks for all optional configuration flags defined in allocator_config.def to allocator_config_test.cpp. This increases the code coverage of allocator_config_wrapper.h by exercising the template instantiations for these optional flags.


cferris1000 (approved):
LGTM

PR #197561, opened 2026-05-13
[AMDGPU] Fix forced lit64 encoding on lit() modifier

rampitec:
We were forcing lit64 encoding on a 64-bit operand with lit()
modifier. This is not required, not compatible with SP3, and
in the pathalogical case creates invalid 4 dword encoding if
used with a VOP3* instruction.

That said if lit() is used the immediate is silently truncated
even before the encoding, so the encoder only sees 32-bits of
relevant data and 32-bits of zeroes anyway. That is a separate
issue, but we never had a true 64-bit constant really properly
encoded with lit() modifier, only with lit64().


rampitec:
* #197583
* #197561 👈
* `main`

<h2></h2>

shiltian (approved):

llvm/test/MC/AMDGPU/literals.s
```diff
@@ -949,8 +951,8 @@ s_mov_b64 s[0:1], 1
 s_mov_b64 s[0:1], lit(1)
 // GFX11: s_mov_b64 s[0:1], lit(0x1)              ; encoding: [0xff,0x01,0x80,0xbe,0x01,0x00,0x00,0x00]
 // GFX12: s_mov_b64 s[0:1], lit(0x1)              ; encoding: [0xff,0x01,0x80,0xbe,0x01,0x00,0x00,0x00]
-// GFX1250-ASM: s_mov_b64 s[0:1], lit(0x1)              ; encoding: [0xfe,0x01,0x80,0xbe,0x01,0x00,0x00,0x00,0x00,0x00,0x00,0x00]
-// GFX1250-DIS: s_mov_b64 s[0:1], lit64(0x1)            ; encoding: [0xfe,0x01,0x80,0xbe,0x01,0x00,0x00,0x00,0x00,0x00,0x00,0x00]
+// GFX1250-ASM: s_mov_b64 s[0:1], lit(0x1)            ; encoding: [0xff,0x01,0x80,0xbe,0x01,0x00,0x00,0x00]
+// GFX1250-DIS: s_mov_b64 s[0:1], lit(0x1)            ; encoding: [0xff,0x01,0x80,0xbe,0x01,0x00,0x00,0x00]
```
jayfoad, line 955: Nit, I wonder why the update script didn't use the common prefix GFX1250

rampitec: I have updated it manually. This time I wanted to be very specific about what am I doing, especially because the test only have disasm checks for gfx1250 and not any other target. Regenerated here: https://github.com/llvm/llvm-project/pull/197583

rampitec: In fact it looks the script served a bad service here. I am continuing finding problem after a problem with the encoding of forced literals.

PR #197563, opened 2026-05-13
[flang][cuda] Use wider cudaMemcpy2D rows for descriptor transfers

clementval:
Extend the CUDA descriptor transfer fast path to choose the largest common contiguous leading block, allowing slices like a(2:nx-1,:,:) to copy full rows with cudaMemcpy2D. Add runtime coverage for both strided element transfers and leading-dimension slice transfers.

This will make example like the one below use the fast path. 

```fortran
program test2d
  implicit none
  integer, parameter :: nx=128, ny=128, nz=128
  real :: a(nx,ny,nz)
  real, device :: a_d(nx,ny,nz)

  a = 0.0
  a_d(2:nx-1,:,:) = a(2:nx-1,:,:)
  
end
```

### DIFF
--- a/flang-rt/lib/cuda/memory.cpp
+++ b/flang-rt/lib/cuda/memory.cpp
@@ -30,9 +30,9 @@ struct Memcpy2DLayout {
   std::size_t pitchBytes;
 };
 
-// Get cudaMemcpy2D layout information if both descriptors have equal element
-// counts and regular positive-stride layouts. Returns a nullopt otherwise to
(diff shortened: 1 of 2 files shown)


wangzpgi:
Could you add a rank-2 sub-array unit test?

clementval:
> Could you add a rank-2 sub-array unit test?

I added one.

wangzpgi (approved):

PR #197568, opened 2026-05-13
[SSAF] Increase Expr kind coverage in EntityPointerLevelTranslator

ziqingluo-90:
- Add support for more kinds of Expr that can be translated to EntityPointerLevel(s).
- Additionally, fix bugs in PointerFlowExtractor discovered by tests added for the new Expr kinds.

### DIFF
--- a/clang/lib/ScalableStaticAnalysisFramework/Analyses/EntityPointerLevel/EntityPointerLevel.cpp
+++ b/clang/lib/ScalableStaticAnalysisFramework/Analyses/EntityPointerLevel/EntityPointerLevel.cpp
@@ -306,15 +306,16 @@ private:
     return EntityPointerLevelSet{};
   }
 
-  // Fall back if the InitListExpr is not an empty or singleton list that
-  // initializes a pointer scalar. EntityPointerLevelTranslator does not
-  // expect to visit an InitListExpr in any other case.
+  // The InitListExpr must be an empty or singleton list that
+  // initializes a pointer scalar.  Other cases are unexpected thus an error.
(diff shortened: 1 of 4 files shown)


clang/lib/ScalableStaticAnalysisFramework/Analyses/EntityPointerLevel/EntityPointerLevel.cpp
```diff
@@ -232,16 +252,89 @@ class EntityPointerLevelTranslator
...
+  // evaluates to a compile-time constant zero (or null). So no EPL in the
+  // result.
+  Expected<EntityPointerLevelSet>
+  VisitImplicitValueInitExpr(const ImplicitValueInitExpr *S) {
+    return EntityPointerLevelSet{};
+  }
+
+  // Fall back if the InitListExpr is not an empty or singleton list that
+  // initializes a pointer scalar. EntityPointerLevelTranslator does not
+  // expect to visit an InitListExpr in any other case.
```
jkorous-apple: This feels like we're leaving behind a footgun since the implementation doesn't communicate violation of this expectation.

Should we just aggregate traversal result for all initializers in the list?
Or should we consider `InitListExpr` unpacking to be completely out of scope?
Or should we perhaps return an error for `getNumInits() > 1`?

ziqingluo-90: You are right that calling `fallback` is not the best option here. 
The intention was that the fall back case should never happen.   I will make it an error for now.

steakhal (approved):
LGTM

PR #197570, opened 2026-05-13
[clang][deps] Create action controller in a factory function

jansvoboda11:
This minimizes diff with downstream.


jansvoboda11:
Not sure if I should make `CallbackActionController` constructible **only** through the factory function, to prevent upstream backsliding and breaking downstream include-tree code.

jansvoboda11:
ping

benlangmuir (commented):
I think this makes sense. It's not just a reduction in downstream diff, I think using a factory here makes it easier to use this code.

However, I would suggest removing the static version of the factory. The static method itself only makes sense in our downstream code for now, and the small delta it creates seems fine.  You're still fixing the main issue by having the (non-static) factory method.

PR #197571, opened 2026-05-13
[clang-sycl-linker] Add per-translation-unit device code split mode

YuriPlyakhin:
Adds `source` split mode to `clang-sycl-linker`, driven by the `sycl-module-id` function attribute emitted by the CFE.
`source` is the default mode and groups kernels by the value of their `sycl-module-id` attribute, emitting one device image per translation unit.
If the linked module contains no entry points, no splitting happens.

The `EntryPointCategorizer` in `ClangSYCLLinker.cpp` is refactored into a class (instead of a stateful lambda) to support both per-kernel and per-TU modes cleanly.

Also fix a potential buffer invalidation bug in sycl::writeSymbolTable where appending symbol names could reallocate the output buffer while pointers into it were still live.

Co-Authored-By: Claude


clang/test/Tooling/clang-sycl-linker-split-mode.ll
```diff
@@ -0,0 +1,78 @@
...
+; RUN: clang-sycl-linker --dry-run -v -triple=spirv64 --module-split-mode=kernel %t.bc -o %t-split-kernel.out 2>&1 \
+; RUN:   | FileCheck %s --check-prefix=SPLIT-KERNEL
+; SPLIT-KERNEL:      sycl-device-link: inputs: {{.*}}.bc  libfiles:  output: [[LLVMLINKOUT:.*]].bc
+; SPLIT-KERNEL-NEXT: sycl-module-split: input: [[LLVMLINKOUT]].bc, mode: kernel
+; SPLIT-KERNEL-NEXT: [[SPLIT0:.*]].bc [kernel_c ]
+; SPLIT-KERNEL-NEXT: [[SPLIT1:.*]].bc [kernel_b ]
+; SPLIT-KERNEL-NEXT: [[SPLIT2:.*]].bc [kernel_a ]
+; SPLIT-KERNEL-NEXT: LLVM backend: input: [[SPLIT0]].bc, output: {{.*}}_0.spv
+; SPLIT-KERNEL-NEXT: LLVM backend: input: [[SPLIT1]].bc, output: {{.*}}_1.spv
+; SPLIT-KERNEL-NEXT: LLVM backend: input: [[SPLIT2]].bc, output: {{.*}}_2.spv
```
YuriPlyakhin, line 29:
```suggestion
; SPLIT-KERNEL-NEXT: LLVM backend: input: [[SPLIT2]].bc, output: {{.*}}_2.spv
; SPLIT-KERNEL-NOT:  {{.+}}
```

YuriPlyakhin:
@maksimsab , FYI

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
```diff
@@ -468,84 +468,118 @@ static Error runAOTCompile(StringRef InputFile, StringRef OutputFile,
...
+static bool isEntryPoint(const Function &F, bool EmitOnlyKernelsAsEntryPoints) {
+  if (F.isDeclaration())
+    return false;
+  if (F.hasKernelCallingConv())
+    return true;
+  if (EmitOnlyKernelsAsEntryPoints)
+    return false;
+  // sycl_external functions carry the "sycl-module-id" attribute.
+  // This branch is not reachable while EmitOnlyKernelsAsEntryPoints is
+  // hardcoded to true (see TODO in runSYCLLink).
```
sarnex: IMO I'm not sure we need this comment about it being unreachable, the code itself makes sense as is, whether it's currently called with `EmitOnlyKernelsAsEntryPoints == false` is not directly related to this function and probably we will forget to update this comment when we call it with `false`

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
```diff
@@ -468,84 +468,118 @@ static Error runAOTCompile(StringRef InputFile, StringRef OutputFile,
...
-///
-/// For SPLIT_NONE, \p LinkedBitcodeFile is returned as-is.
-/// For SPLIT_PER_KERNEL, the module is split into parts such that each part
-/// contains exactly one kernel entry point and its transitive dependencies;
-/// each part is written to a fresh temporary bitcode file.
-static Expected<SmallVector<SplitModule, 0>>
-splitDeviceCode(std::unique_ptr<Module> M, StringRef LinkedBitcodeFile,
-                IRSplitMode Mode, const ArgList &Args) {
-  SmallVector<SplitModule, 0> SplitModules;
+/// Functor passed to splitModuleTransitiveFromEntryPoints. For each input \p F,
```
sarnex:
```suggestion
/// Functor passed to splitModuleTransitiveFromEntryPoints. For each input function \p F,
```

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
```diff
@@ -560,17 +594,40 @@ splitDeviceCode(std::unique_ptr<Module> M, StringRef LinkedBitcodeFile,
...
+      errs() << "]\n";
+    }
+  }
+
   return SplitModules;
 }
 
+/// Returns true if module splitting can be skipped: either \p Mode is
+/// SPLIT_NONE, or \p M contains no entry points (nothing to split from).
+static bool checkModuleSplitCanBeSkipped(IRSplitMode Mode, const Module &M,
```
sarnex:
```suggestion
static bool canSkipModuleSplit(IRSplitMode Mode, const Module &M,
```

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
YuriPlyakhin: https://github.com/llvm/llvm-project/pull/197571/changes/2f45ed2095bedec91d4f4874a3c3d202b552c67a

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
YuriPlyakhin: https://github.com/llvm/llvm-project/pull/197571/changes/2f45ed2095bedec91d4f4874a3c3d202b552c67a

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
YuriPlyakhin: https://github.com/llvm/llvm-project/pull/197571/changes/2f45ed2095bedec91d4f4874a3c3d202b552c67a

sarnex (approved):

PR #197572, opened 2026-05-13
[CIR] Route lambda and thunk aggregate returns through return slot

adams381:
Two related forwarding-call paths -- the captureless lambda
static-invoker `emitForwardingCallToLambda` in `CIRGenClass.cpp` and
vtable thunk emission `emitCallAndReturnForThunk` in
`CIRGenVTables.cpp` -- both forward a callee's return value up to
their own caller.  When the callee returns an aggregate, both must
bind their `ReturnValueSlot` to `returnValue` so the result lands in
the function's return slot directly, and both must emit a
`cir.return` that loads from it.

The lambda path was leaving the slot empty, materializing a
temporary, and then asking `emitReturnOfRValue` to aggregate-copy it
into `returnValue`.  For non-trivial aggregates (e.g.
`std::basic_string`) this hits the "Trying to aggregate-copy a type
without a trivial copy/move constructor" assertion in
`emitAggregateCopy`.  The thunk path was already binding the slot
but never emitted the follow-up `cir.return`, so
`LexicalScope::emitImplicitReturn` ended the function with
`cir.trap` (-O0) or `cir.unreachable` (-O1+) -- silently discarding
the just-stored result.  Any vtable call through a second-base-class
virtual function returning a non-trivial aggregate would therefore
trap or return garbage at runtime.  No existing test caught this;
thunks tests only cover primitive returns.

Both paths get the same fix: bind the slot on the aggregate path,
route the non-void exit through `emitReturnOfRValue` unconditionally,
and short-circuit the self-copy when the rvalue already sits at
`returnValue`.  The lambda fix unblocks 24 libcxx `std/` failures
in the C++26 `<format>` subtree (`format.tuple`,
`format.range.fmtmap` / `fmtstr` / `fmtset`,
`container.adaptors.format`) where lambdas returning
`std::basic_string` are converted to function pointers; the thunk
fix is correctness-only.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenClass.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenClass.cpp
@@ -909,6 +909,15 @@ void CIRGenFunction::emitForwardingCallToLambda(
       callOperator->getType()->castAs<FunctionProtoType>();
   QualType resultType = fpt->getReturnType();
   ReturnValueSlot returnSlot;
+  // This should also be tracking volatile, unused, and externally destructed.
+  assert(!cir::MissingFeatures::returnValueSlotFeatures());
+  // For aggregate returns, write the callee's result directly into the
+  // static invoker's return slot.  Otherwise emitReturnOfRValue below would
+  // aggregate-copy a temporary into returnValue, which is incorrect for
+  // types without a trivial copy/move (e.g. std::string) -- and trips an
+  // assertion in emitAggregateCopy.
+  if (!resultType->isVoidType() && hasAggregateEvaluationKind(resultType))
+    returnSlot = ReturnValueSlot(returnValue);
 
   // We don't need to separately arrange the call arguments because
   // the call can't be variadic anyway --- it's impossible to forward
@@ -919,9 +928,10 @@ void CIRGenFunction::emitForwardingCallToLambda(
       CIRGenCallee::forDirect(calleePtr, GlobalDecl(callOperator));
   RValue rv = emitCall(calleeFnInfo, callee, returnSlot, callArgs);
 
-  // If necessary, copy the returned value into the slot.
-  if (!resultType->isVoidType() && returnSlot.isNull()) {
-    if (getLangOpts().ObjCAutoRefCount && resultType->isObjCRetainableType())
+  // Forward the returned value through the function's return slot.
+  if (!resultType->isVoidType()) {
+    if (returnSlot.isNull() && getLangOpts().ObjCAutoRefCount &&
+        resultType->isObjCRetainableType())
       cgm.errorNYI(callOperator->getSourceRange(),
                    "emitForwardingCallToLambda: ObjCAutoRefCount");
     emitReturnOfRValue(*currSrcLoc, rv, resultType);
--- a/clang/lib/CIR/CodeGen/CIRGenStmt.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenStmt.cpp
@@ -1253,9 +1253,16 @@ void CIRGenFunction::emitReturnOfRValue(mlir::Location loc, RValue rv,
   if (rv.isScalar()) {
     builder.createStore(loc, rv.getValue(), returnValue);
   } else if (rv.isAggregate()) {
-    LValue dest = makeAddrLValue(returnValue, ty);
-    LValue src = makeAddrLValue(rv.getAggregateAddress(), ty);
-    emitAggregateCopy(dest, src, ty, getOverlapForReturnValue());
+    Address rvAddr = rv.getAggregateAddress();
+    // If the aggregate is already in the return slot (e.g. a callee was
+    // invoked through a ReturnValueSlot bound to returnValue), the copy is
+    // a no-op.  Calling emitAggregateCopy here would also incorrectly
+    // require the type to have a trivial copy/move.
+    if (rvAddr.getPointer() != returnValue.getPointer()) {
+      LValue dest = makeAddrLValue(returnValue, ty);
+      LValue src = makeAddrLValue(rvAddr, ty);
+      emitAggregateCopy(dest, src, ty, getOverlapForReturnValue());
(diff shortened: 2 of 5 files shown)


clang/lib/CIR/CodeGen/CIRGenClass.cpp
```diff
@@ -919,9 +928,10 @@ void CIRGenFunction::emitForwardingCallToLambda(
       CIRGenCallee::forDirect(calleePtr, GlobalDecl(callOperator));
   RValue rv = emitCall(calleeFnInfo, callee, returnSlot, callArgs);
 
-  // If necessary, copy the returned value into the slot.
-  if (!resultType->isVoidType() && returnSlot.isNull()) {
-    if (getLangOpts().ObjCAutoRefCount && resultType->isObjCRetainableType())
+  // Forward the returned value through the function's return slot.
+  if (!resultType->isVoidType()) {
```
andykaylor, line 932: Why did you take `returnSlot.isNull` out of this line?

clang/lib/CIR/CodeGen/CIRGenStmt.cpp
```diff
@@ -1253,9 +1253,16 @@ void CIRGenFunction::emitReturnOfRValue(mlir::Location loc, RValue rv,
...
   } else if (rv.isAggregate()) {
-    LValue dest = makeAddrLValue(returnValue, ty);
-    LValue src = makeAddrLValue(rv.getAggregateAddress(), ty);
-    emitAggregateCopy(dest, src, ty, getOverlapForReturnValue());
+    Address rvAddr = rv.getAggregateAddress();
+    // If the aggregate is already in the return slot (e.g. a callee was
+    // invoked through a ReturnValueSlot bound to returnValue), the copy is
+    // a no-op.  Calling emitAggregateCopy here would also incorrectly
+    // require the type to have a trivial copy/move.
+    if (rvAddr.getPointer() != returnValue.getPointer()) {
```
andykaylor, line 1261: Is this working around a problem caused by ABI handling not having been applied yet at this point? The fact that classic codegen doesn't have anything like this concerns me.

clang/test/CIR/CodeGen/lambda-static-invoker-agg-return.cpp
```diff
@@ -0,0 +1,51 @@
...
+// CIR-SAME:    (%[[I_ARG:.*]]: !s32i {{.*}}) -> !rec_S
+// CIR:         %[[I_ALLOCA:.*]] = cir.alloca !s32i, !cir.ptr<!s32i>, ["i", init]
+// CIR:         %[[RETVAL:.*]] = cir.alloca !rec_S, !cir.ptr<!rec_S>, ["__retval"]
+// CIR:         %[[UNUSED:.*]] = cir.alloca !rec_anon{{[^,]*}}, {{.*}} ["unused.capture"]
+// CIR-NOT:     cir.alloca {{.*}} ["agg.tmp
+// CIR:         cir.store %[[I_ARG]], %[[I_ALLOCA]]
+// CIR:         %[[I:.*]] = cir.load{{.*}} %[[I_ALLOCA]]
+// CIR:         %[[CALL:.*]] = cir.call @_ZZ11agg_invokervENK3$_0clEi(%[[UNUSED]], %[[I]]){{.*}} -> !rec_S
+// CIR-NOT:     cir.copy
+// CIR:         cir.store{{.*}} %[[CALL]], %[[RETVAL]]
```
andykaylor, line 32: This concerns me. Is this not literally copying the contents of the object? I feel like maybe we need a special instruction that indicates that we're returning an object without copying it.

clang/lib/CIR/CodeGen/CIRGenClass.cpp
adams381: Flipped the control flow: for non-void returns we always call emitReturnOfRValue unless we hit the ObjC-retainable NYI. Aggregates bind ReturnValueSlot to returnValue before the call so the callee writes sret-style into __retval.

clang/lib/CIR/CodeGen/CIRGenStmt.cpp
adams381: Classic CodeGen binds the return slot earlier in the epilog path. CIR still has explicit copy ops in the IR; the pointer-equality guard only skips a redundant aggregate copy when the callee already wrote into returnValue. Net effect matches OGCG sret on the static invoker.

clang/test/CIR/CodeGen/lambda-static-invoker-agg-return.cpp
adams381: The call returns into the slot bound via ReturnValueSlot — not a separate agg.tmp memcpy. The test store/load through __retval mirrors OGCG: the invoker is void with an sret parameter, and the call writes directly into it.

andykaylor: Writing directly into the sret parameter is different than the store+load here. Store+load is effectively a memcpy. We should be returning `%CALL` directly. I realize that we expect this store+load to go away during ABI lowering, but the fact that we've got invalid semantics until then is a problem.

andykaylor: I've just filed https://github.com/llvm/llvm-project/issues/198602 to track this issue, which isn't limited to this PR.

andykaylor (approved):
lgtm (with the caveat that it hits an existing bug)

clang/test/CIR/CodeGen/lambda-static-invoker-agg-return.cpp
adams381: Agreed — #198602 is the right place for the store/load vs direct-return representation. This PR only wires the return slot so we stop aggregate-copying through a temporary; happy to revisit the CIR shape once CallConv lowering can express it.

PR #197573, opened 2026-05-13
[AMDGPU][Scheduler] Remove potential for independent subregs in tests

lucas-rami:
The `RenameIndependentSubregs` pass runs before the machine scheduler, so it should not be possible to have an input MIR with the potential to yield disconnected components when moving sub-register definitions around (as indicated in comments in `TargetPassConfig::addOptimizedRegAlloc`). This removes a few sub-registers defs in the scheduler's unit tests that could create such situations and narrows down the super register class they are a part of to the minimum.

Previously, this was understood as a rematerialization-related problem so a fix was proposed in 94875ae. This is however not an issue that is generated or can be handled fully by the rematerializer without making its use noticeably more complicated; therefore this reverts the fix, removing support for live-interval splitting inside the rematerializer as well as a unit test meant to expose the issue.


lucas-rami:
- #197580
- #202673
- #197579
- #197576
- #197575
- `main`

arsenm:
> The `RenameIndependentSubregs` pass runs before the machine scheduler, so it should not be possible to have an input MIR with the potential to yield disconnected components when moving sub-register definitions around (as indicated in comments in `TargetPassConfig::addOptimizedRegAlloc`)

RenameIndependentSubregs doesn't give you a property you can rely on though? Nothing is stopping later passes, or handwritten MIR from hitting this.

lucas-rami:
Apologies for the delay in addressing this, it took me some time to find a good way to address the issue reliably in the rematerializer. I modified the stack accordingly and am replacing this PR with #202673, which solves the problem more systematically from within the rematerializer.

PR #197574, opened 2026-05-13
[CIR] Handle goto jumping out of a cleanup scope

andykaylor:
This adds the FlattenCFG handling for a goto operation whose target is outside of the cleanup scope in which the goto appears. Most of the handling for this was already in place. We just needed to create a new goto in the cleanup continuation block associated with the goto branching through the cleanup (we were putting an unreachable there before). The main thing that was needed was testing.


erichkeane (approved):

AmrDeveloper (approved):
LGTM

PR #197575, opened 2026-05-13
[CodeGen] Refactor rematerializer unit tests to reduce boilerplate (NFC)

lucas-rami:
Taking inspiration from other unit tests, this refactors the rematerializer's unit tests with the aim of reducing the amount of repetitive boilerplate code in each individual test, making the addition of new tests easier/faster in the future.

Each unit test defines some input MIR to instantiate a rematerializer on; a user-provided callback is then invoked with a pre-initialized rematerializer wrapped in a helper object that factors out as much common verification logic as possible.

### DIFF
--- a/llvm/unittests/CodeGen/RematerializerTest.cpp
+++ b/llvm/unittests/CodeGen/RematerializerTest.cpp
@@ -524,7 +524,8 @@ TEST_F(RematerializerTest, SplitSubRegDeadDef) {
     MoveMIAndAdjustLiveness(Sub1Def);
 
     // Rematerialize %1 to bb.1. This triggers a live-interval update of %0 when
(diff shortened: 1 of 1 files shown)


lucas-rami:
- #197580
- #202673
- #197579
- #197576
- #197575 👈
- `main`

arsenm (approved):

llvm/unittests/CodeGen/RematerializerTest.cpp
MacDue, line 98: This only works if the AMDGPU target is enabled.

lucas-rami: Thanks for the catch, fix in #202966.

PR #197576, opened 2026-05-13
[CodeGen] Fix incorrect rematerialization rollback order

lucas-rami:
This fixes an issue in the rematerializer's rollbacker wherein adjacent MIs that were deleted through rematerializations would sometimes---depending on the exact order in which they were deleted---not be re-created in their original pre-rematerialization order. While this does not impact correctness (i.e., use-def relations are always honored), this goes against the rollbacker's intent to re-create the MIR exactly as it was pre-rematerializations (up to slot index changes).

### DIFF
--- a/llvm/include/llvm/CodeGen/Rematerializer.h
+++ b/llvm/include/llvm/CodeGen/Rematerializer.h
@@ -538,11 +538,17 @@ private:
     /// stores its index. Otherwise equals \ref Rematerializer::NoReg.
     RegisterIdx NextRegIdx;
 
-    RollbackInfo(const Rematerializer &Remater, RegisterIdx RegIdx);
+    RollbackInfo(const Rematerializer::Reg &Reg, RegisterIdx NextRegIdx);
   };
 
   /// Original registers that have been deleted, in order of deletion.
   MapVector<RegisterIdx, RollbackInfo> DeadRegs;
+  /// When there are two ajacent rematerializable MIs in the original
+  /// instruction order and the later one is deleted, stores a mapping from the
+  /// earlier one's register index to the later one's register index. If the
+  /// earlier one is then deleted, this makes it possible to rematerialize it at
+  /// the correct position after the later one is re-created.
+  DenseMap<RegisterIdx, RegisterIdx> AdjacentDeletedMIs;
   /// Registers which have been rematerialized (from original index to
   /// rematerialized index).
   DenseMap<RegisterIdx, Rematerializer::RematsOf> Rematerializations;
--- a/llvm/include/llvm/CodeGen/Rematerializer.h
+++ b/llvm/include/llvm/CodeGen/Rematerializer.h
@@ -14,6 +14,7 @@
 #ifndef LLVM_CODEGEN_REMATERIALIZER_H
 #define LLVM_CODEGEN_REMATERIALIZER_H
 
+#include "llvm/ADT/PointerUnion.h"
 #include "llvm/CodeGen/LiveIntervals.h"
 #include "llvm/CodeGen/MachineBasicBlock.h"
 #include "llvm/CodeGen/MachineRegisterInfo.h"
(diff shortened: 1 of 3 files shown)


lucas-rami:
- #197580
- #202673
- #197579
- #197576 👈
- #197575
- `main`

llvm/include/llvm/CodeGen/Rematerializer.h
```diff
@@ -538,11 +538,17 @@ class Rollbacker : public Rematerializer::Listener {
...
   };
 
   /// Original registers that have been deleted, in order of deletion.
   MapVector<RegisterIdx, RollbackInfo> DeadRegs;
+  /// When there are two ajacent rematerializable MIs in the original
+  /// instruction order and the later one is deleted, stores a mapping from the
+  /// earlier one's register index to the later one's register index. If the
+  /// earlier one is then deleted, this makes it possible to rematerialize it at
+  /// the correct position after the later one is re-created.
+  DenseMap<RegisterIdx, RegisterIdx> AdjacentDeletedMIs;
```
arsenm: I don't understand why you need to track a new map for ordering. Can't you infer this from the SlotIndex?

lucas-rami: I improved the position tracking method to not rely on that kind of ordering and more readily support future use cases, including those in this PR stack.

llvm/lib/CodeGen/Rematerializer.cpp
```diff
@@ -762,56 +758,75 @@ Printable Rematerializer::printUser(const MachineInstr *MI,
...
+  // in def-use order. This also ensures that re-creation positions that became
+  // invalid due to later MI deletions can be corrected as we go.
+  for (const DeadReg &Reg : reverse(DeadRegs)) {
+    assert(!Remater.getReg(Reg.Idx).isAlive() && "register should be dead");
+
+    // Determine re-creation position for the register's definition.
+    MachineBasicBlock::iterator InsertPosition;
+    const auto [Ptr, IsMBB] = Positions[--PositionIndex];
+    if (IsMBB) {
+      InsertPosition = static_cast<MachineBasicBlock *>(Ptr)->end();
```
arsenm: Should not have these casts

lucas-rami: Would using a `union` of `MachineInstr*` and `MachineBasicBlock*` be better instead of `void*` as the pointer type of `PoitnerIntPair`? That would still require a `static_cast` in the union's `PointerLikeTypeTraits::getFromVoidPointer` implementation, but none in this function.

lucas-rami: I replaced it with `union InsertPosPtr` to avoid the `static_cast` in the rollback code.

llvm/include/llvm/CodeGen/Rematerializer.h
```diff
@@ -514,6 +511,12 @@ class Rematerializer {
 /// rematerializations as soon as it is attached to the rematerializer.
 class LLVM_ABI Rollbacker : public Rematerializer::Listener {
 public:
+  /// An insertion position pointer in the MIR.
+  union InsertPosPtr {
+    MachineInstr *MI;
+    MachineBasicBlock *MBB;
+  };
```
arsenm: Why not MachineBasicBlock + MachineBasicBlock::iterator? There's probably already a utility for that

llvm/lib/CodeGen/Rematerializer.cpp
```diff
@@ -762,56 +758,72 @@ Printable Rematerializer::printUser(const MachineInstr *MI,
...
+    if (IsMBB)
+      InsertPosition = Ptr.MBB->end();
+    else
+      InsertPosition = Replacements.lookup_or(Ptr.MI, Ptr.MI)->getIterator();
+
+    Remater.recreateReg(Reg.Idx, InsertPosition, Reg.DefReg);
+
+    const Rematerializer::Reg &RecreateReg = Remater.getReg(Reg.Idx);
+    assert(!Replacements.contains(Reg.DefMI) && "duplicate deleted MI");
+    Replacements[Reg.DefMI] = RecreateReg.DefMI;
```
arsenm: .insert + llvm_unreachable

llvm/include/llvm/CodeGen/Rematerializer.h
lucas-rami: I discovered `PointerUnion` when looking for something like that. It doesn't work for `MachineBasicBlock::iterator` but makes the code cleaner than with the raw union and avoids the original raw `static_cast` in favor of `llvm::cast`. Hopefully that's better.

arsenm (approved):

PR #194473, opened 2026-04-27
[X86] Replace custom minmax reduction pattern matching with ISD::VEC_REDUCE_*MIN/MAX support

RKSimon:
Support middle-end reduction integer min/max patterns instead of relying on the ExpandReductions pass and then matching the expanded pattern in DAG.

Middle-end reduction pattern recognition now matches SelectionDAG::matchBinOpReduction (inc partial reduction), and its better to improve future handling in InstCombine/VectorCombine wherever possible.

Fixes #194624

PR #197578, opened 2026-05-13
[X86] Improve lowering of i32/i64 minmax reductions

RKSimon:
Allow 32-bit targets to correctly lower i64 ISD::VECREDUCE min/max nodes via ReplaceNodeResults - this is necessary once we're finally ready for #194473 and remove combineMinMaxReduction entirely

Improve handling of v2iXX reduction stages by consistently preferring binop(extract(),extract()) scalarisation on SSE targets (if the vector binop isn't legal).

### DIFF
--- a/llvm/lib/Target/X86/X86ISelLowering.cpp
+++ b/llvm/lib/Target/X86/X86ISelLowering.cpp
@@ -29691,8 +29691,7 @@ static SDValue LowerVECREDUCE(SDValue Op, const X86Subtarget &Subtarget,
   unsigned NumSrcElts = SrcVT.getVectorNumElements();
   for (unsigned NumElts = NumSrcElts; NumElts != 1; NumElts /= 2) {
     // Scalarize the last 2 elements if the vector binop isn't legal.
-    if (NumElts == 2 &&
-        (!Subtarget.hasAVX() || BinOp == ISD::UMIN || BinOp == ISD::UMAX) &&
+    if (NumElts == 2 && !Subtarget.hasAVX512() &&
         !TLI.isOperationLegal(BinOp, SrcVT) && TLI.isTypeLegal(ExtractVT)) {
       return DAG.getNode(BinOp, DL, ExtractVT,
                          DAG.getExtractVectorElt(DL, ExtractVT, Src, 0),
--- a/llvm/lib/Target/X86/X86ISelLowering.cpp
+++ b/llvm/lib/Target/X86/X86ISelLowering.cpp
@@ -29690,8 +29690,9 @@ static SDValue LowerVECREDUCE(SDValue Op, const X86Subtarget &Subtarget,
   // Expand 128-bit shuffle tree + reduction binops.
   unsigned NumSrcElts = SrcVT.getVectorNumElements();
   for (unsigned NumElts = NumSrcElts; NumElts != 1; NumElts /= 2) {
-    // SSE only - scalarize the last 2 elements if the vector binop isn't legal.
-    if (NumElts == 2 && !Subtarget.hasAVX() &&
+    // Scalarize the last 2 elements if the vector binop isn't legal.
+    if (NumElts == 2 &&
+        (!Subtarget.hasAVX() || BinOp == ISD::UMIN || BinOp == ISD::UMAX) &&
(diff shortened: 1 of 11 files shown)


llvm/test/CodeGen/X86/vector-reduce-umin.ll
```diff
@@ -70,62 +70,37 @@ define i64 @test_v2i64(<2 x i64> %a0) nounwind {
...
-; X64-SSE41-NEXT:    pmovsxdq %xmm4, %xmm5
-; X64-SSE41-NEXT:    pcmpeqd %xmm2, %xmm0
-; X64-SSE41-NEXT:    pshufd {{.*#+}} xmm2 = xmm0[1,1,3,3]
-; X64-SSE41-NEXT:    pand %xmm5, %xmm2
-; X64-SSE41-NEXT:    pshufd {{.*#+}} xmm0 = xmm4[1,1,3,3]
-; X64-SSE41-NEXT:    por %xmm2, %xmm0
-; X64-SSE41-NEXT:    blendvpd %xmm0, %xmm1, %xmm3
-; X64-SSE41-NEXT:    movq %xmm3, %rax
-; X64-SSE41-NEXT:    retq
+; X64-SSE4-LABEL: test_v2i64:
```
phoebewang, line 73: The scalar version looks like the most efficient one for <2 x i64> on X64?

RKSimon: Yes, even with PCMPGTQ on SSE42 its seems to be better to scalarize - AVX is trickier as VBLENDV gets better, I can special case UMIN/UMAX if you think its worth it?

phoebewang: I think scalar is better till AVX2. The movq is required any way. Then we just have one pextrq and 2 GPR instructions. In contrast, the vector format have more instructions and 2 extra memory load.

phoebewang: But since the old code is already bad, I think it's ok not to do it in this patch.

phoebewang (approved):
LGTM.

llvm/test/CodeGen/X86/vector-reduce-umin.ll
RKSimon: @phoebewang  Pushed the change now - can you confirm you're still happy with it?

llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -29697,6 +29690,15 @@ static SDValue LowerVECREDUCE(SDValue Op, const X86Subtarget &Subtarget,
   // Expand 128-bit shuffle tree + reduction binops.
   unsigned NumSrcElts = SrcVT.getVectorNumElements();
   for (unsigned NumElts = NumSrcElts; NumElts != 1; NumElts /= 2) {
+    // Scalarize the last 2 elements if the vector binop isn't legal.
+    if (NumElts == 2 &&
+        (!Subtarget.hasAVX() || BinOp == ISD::UMIN || BinOp == ISD::UMAX) &&
```
phoebewang:
```suggestion
        (!Subtarget.hasAVX512() || BinOp == ISD::UMIN || BinOp == ISD::UMAX) &&
```

aeubanks:
this crashes on 

```
$ /tmp/a.ll
target datalayout = "e-m:x-p:32:32-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32-a:0:32-S32"
target triple = "i386-pc-windows-msvc19.34.0"

define i64 @f() "target-cpu"="pentium4" {
entry:
  %0 = call i64 @llvm.vector.reduce.smax.v6i64(<6 x i64> zeroinitializer)
  ret i64 %0
}
$ llc -o /dev/null /tmp/a.ll
llc: ../../llvm/lib/Target/X86/X86ISelLowering.cpp:4316: SDValue extractSubVector(SDValue, unsigned int, SelectionDAG &, const SDLoc &, unsigned int): Assertion `isPowerOf2_32(ElemsPerChunk) && "Elements per chunk not power of 2"' failed.
```

RKSimon:
@aeubanks Taking a look now

RKSimon:
@aeubanks #198989

PR #197579, opened 2026-05-13
[CodeGen][AMDGPU] Prepare rematerializer for multi-def remat support (NFC)

lucas-rami:
This makes some NFCs to the rematerializer before adding initial support for rematerializing registers with multiple defs. The main change is that, in the representation of (un)rematerializable register dependencies, we drop references to machine operand indices which lose meaning in the multi-def case. 

Other minor changes listed below.

- Removal of `DefRegion` argument to `Rematerializer::recreateReg`. Registers are always re-created in their original region so there is no need to set their region again.
- Removal of `InsertPos` unused argument to `Rematerializer::postRematerialization`.
- Refactor of how AMDGPU's scheduler checks whether a given register is rematerializable.

### DIFF
--- a/llvm/lib/CodeGen/Rematerializer.cpp
+++ b/llvm/lib/CodeGen/Rematerializer.cpp
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "llvm/CodeGen/Rematerializer.h"
+#include "llvm/ADT/MapVector.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SetVector.h"
 #include "llvm/CodeGen/LiveIntervals.h"
--- a/llvm/lib/CodeGen/Rematerializer.cpp
+++ b/llvm/lib/CodeGen/Rematerializer.cpp
@@ -14,6 +14,7 @@
 
 #include "llvm/CodeGen/Rematerializer.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SetVector.h"
 #include "llvm/CodeGen/LiveIntervals.h"
 #include "llvm/CodeGen/MachineBasicBlock.h"
 #include "llvm/CodeGen/MachineOperand.h"
--- a/llvm/lib/CodeGen/Rematerializer.cpp
+++ b/llvm/lib/CodeGen/Rematerializer.cpp
@@ -94,21 +94,21 @@ Rematerializer::rematerializeToPos(RegisterIdx RootIdx, unsigned UseRegion,
   assert(!DRI.DependencyMap.contains(RootIdx));
   LLVM_DEBUG(dbgs() << "Rematerializing " << printID(RootIdx) << '\n');
 
-  SmallVector<Reg::Dependency, 2> NewDeps;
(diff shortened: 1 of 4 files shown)


lucas-rami:
- #197580
- #202673
- #197579 👈
- #197576
- #197575
- `main`

llvm/lib/CodeGen/Rematerializer.cpp
```diff
@@ -202,34 +202,37 @@ void Rematerializer::updateLiveIntervals() {
...
+                            : MRI.getMaxLaneMaskForVReg(MO.getReg());
+  return isRegIdenticalAtUses(
+      MO.getReg(), Mask,
+      LIS.getInstructionIndex(*MO.getParent()).getRegSlot(true), Uses);
+}
+
+bool Rematerializer::isRegIdenticalAtUses(Register Reg, LaneBitmask Mask,
+                                          SlotIndex RefSlot,
+                                          ArrayRef<SlotIndex> Uses) const {
+  if (Uses.empty())
```
arsenm, line 245: Does this case happen? Why would this be asked for dead values

llvm/lib/CodeGen/Rematerializer.cpp
```diff
@@ -711,11 +730,12 @@ Printable Rematerializer::printRematReg(RegisterIdx RegIdx,
...
-    OS << printID(RegIdx) << ' ';
-    PrintReg.DefMI->print(OS, /*IsStandalone=*/true, /*SkipOpers=*/false,
-                          /*SkipDebugLoc=*/false, /*AddNewLine=*/false);
-    OS << " @ ";
-    LIS.getInstructionIndex(*PrintReg.DefMI).print(OS);
+    if (PrintReg.isAlive()) {
+      PrintReg.DefMI->print(OS, /*IsStandalone=*/true, /*SkipOpers=*/false,
+                            /*SkipDebugLoc=*/false, /*AddNewLine=*/false);
+      OS << " @ ";
+      LIS.getInstructionIndex(*PrintReg.DefMI).print(OS);
```
arsenm, line 748:
```suggestion
      OS << " @ " << LIS.getInstructionIndex(*PrintReg.DefMI);
```

llvm/unittests/CodeGen/RematerializerTest.cpp
```diff
@@ -464,10 +464,21 @@ TEST_F(RematerializerTest, SubReg) {
     undef %2.sub0:vreg_64_align2 = nofpexcept V_CVT_I32_F64_e32 2, implicit $exec, implicit $mode
 
     undef %34.sub0:vreg_64_align2 = nofpexcept V_CVT_I32_F64_e32 3, implicit $exec, implicit $mode
-
+    
+    undef %56.sub0:sreg_64 = S_MOV_B32 5
```
arsenm, line 473: These tests probably should be cleaned up to use named registers

arsenm (approved):

llvm/lib/CodeGen/Rematerializer.cpp
lucas-rami: It doesn't happen currently but felt semantically correct to me. I will remove the check as the rest of the function produces the same result anyway.

PR #197581, opened 2026-05-13
[AMDGPU] Set Src2RC64 in VOPProfileMQSAD, NFC

changpeng:
The second operand of v_mqsad_u32_u8 can only be VGPRs. So we explicitly
set Src2RC64 to 'VReg' in VOPProfileMQSAD, making it possible for v4i32 to be
VGPR and/or SGPR in the future.

### DIFF
--- a/llvm/lib/Target/AMDGPU/VOP3Instructions.td
+++ b/llvm/lib/Target/AMDGPU/VOP3Instructions.td
@@ -424,7 +424,7 @@ defm V_PERM_B32 : VOP3Inst <"v_perm_b32", VOP3_Profile<VOP_I32_I32_I32_I32>, AMD
 def VOPProfileMQSAD : VOP3_Profile<VOP_V4I32_I64_I32_V4I32, VOP3_CLAMP> {
   let HasModifiers = 0;
   // Src2 only accepts VGPRs
-  let Src2RC64 = getVregSrcForVT<v4i32>.ret;
+  let Src2RC64 = getVOP3VRegSrcForVT<Src2VT>.ret;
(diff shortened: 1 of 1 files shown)


shiltian (approved):

changpeng:
HI, @rampitec :

I expect this change to be NFC, and it is the case downstream.  Do yo know why the dsiassembler does not work upstream here?  Thanks 
 
v_mqsad_u32_u8 v[5:8], v[1:2], v2, v[252:255]
-// GFX13: v_mqsad_u32_u8 v[5:8], v[1:2], v2, v[252:255] ; encoding: [0x05,0x00,0x3d,0xd6,0x01,0x05,0xf2,0x07]
+// GFX13-ASM: v_mqsad_u32_u8 v[5:8], v[1:2], v2, v[252:255] ; encoding: [0x05,0x00,0x3d,0xd6,0x01,0x05,0xf2,0x07]
+// GFX13-DIS: v_mqsad_u32_u8 v[5:8], v[1:2], v2, v[252:255] /*v[508:511]*/ ; encoding: [0x05,0x00,0x3d,0xd6,0x01,0x05,0xf2,0x07]

changpeng:
getVOP3VRegSrcForVT is appropriate upstream. So maybe we should not make the change here upstream.

shiltian:
It might be because some gfx13 changes have not be brought to upstream yet.

changpeng:
I am going to drop this change because changing the default register class for v4i32 is risky. We can not do that without examining all uses of v4i32, and doing similar overriding like this.

PR #197582, opened 2026-05-13
[CIR] Lower member-pointer casts in constant emitter

adams381:
Superseded: stacked branch `cir-data-member-cast-stacked` replaces this PR (offset-based cast on `main`). See the new PR linked from the compare stack.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenItaniumCXXABI.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenItaniumCXXABI.cpp
@@ -190,6 +190,9 @@ public:
   cir::MethodAttr buildVirtualMethodAttr(cir::MethodType methodTy,
                                          const CXXMethodDecl *md) override;
 
+  mlir::Attribute emitMemberPointerConversion(const CastExpr *e,
+                                              mlir::Attribute src) override;
+
   Address initializeArrayCookie(CIRGenFunction &cgf, Address newPtr,
                                 mlir::Value numElements, const CXXNewExpr *e,
                                 QualType elementType) override;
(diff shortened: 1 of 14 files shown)


clang/lib/CIR/CodeGen/CIRGenItaniumCXXABI.cpp
```diff
@@ -2398,6 +2401,81 @@ CIRGenItaniumCXXABI::buildVirtualMethodAttr(cir::MethodType methodTy,
...
+    offset += layout.getBaseClassOffset(baseDecl);
+    rd = baseDecl;
+  }
+  return offset;
+}
+
+mlir::Attribute
+CIRGenItaniumCXXABI::emitMemberPointerConversion(const CastExpr *e,
+                                                 mlir::Attribute src) {
+  assert(e->getCastKind() == CK_DerivedToBaseMemberPointer ||
```
erichkeane, line 2429: IIF we're going to do this, we should probably convert the version in CIRGenScalar as well, where we have a bit of a shortcut.  That said, I still disagree about doing this with offset rather than index, so we should let that discussion finalize first.

adams381: Understood — holding this until the offset-vs-index direction on #197555 is settled. I will also align the CIRGenScalar shortcut if we proceed with the cast path.

adams381:
Closing in favor of stacked branch `cir-data-member-cast-stacked` (member_path casts on top of #197555).

Open the replacement PR from:
https://github.com/llvm/llvm-project/compare/cir-data-member-offset-nfc...adams381:cir-data-member-cast-stacked?expand=1

Set base repository to **adams381/llvm-project** / `cir-data-member-offset-nfc` if the compare page offers that. Head: `cir-data-member-cast-stacked`.

adams381:
Replacement PR: [#198910](https://github.com/llvm/llvm-project/pull/198910)

PR #197584, opened 2026-05-13
[Clang][CUDA] Introduce support for 'f' GPU variants and feature test macros

Artem-B:
- Clang now accepts 'f' GPU variants as the target for sm_100+ GPUs.
- `__CUDA_ARCH_SPECIFIC__` and `__CUDA_ARCH_FAMILY_SPECIFIC__` are now defined to allow distinguishing `a`/`f`/base GPU vaiants.
- refactored BuiltinsNVPTX.td to handle availability quirks introduced by the 'f' variants, and to simplify additions of new GPU/PTX variants to just adding a number to a list.
- bulk test changes to deal with the tablegen-generated strings.


clang/lib/Basic/Targets/NVPTX.cpp
```diff
@@ -184,9 +184,13 @@ void NVPTXTargetInfo::getTargetDefines(const LangOptions &Opts,
     unsigned ArchID = CudaArchToID(GPU);
     Builder.defineMacro("__CUDA_ARCH__", llvm::Twine(ArchID));
 
-    if (IsNVIDIAAcceleratedOffloadArch(GPU))
+    if (IsNVIDIAAcceleratedOffloadArch(GPU)) {
+      Builder.defineMacro("__CUDA_ARCH_SPECIFIC__", llvm::Twine(ArchID));
```
jhuber6, line 188: Guessing these are NVCC names?

jhuber6 (commented):
Seems straightforward enough. Can you verify that we handle these in the ELF parsing code as well?

clang/lib/Basic/Targets/NVPTX.cpp
Artem-B: Looks like they've shifted the language over time. It was called "Accelerated Architecture variant" or something like that. But right now they seem to have settled on `architecture-specific compiler target ` and `family-specific target`

https://docs.nvidia.com/cuda/cuda-c-programming-guide/#architecture-specific-features

I think `IsNVIDIAAcceleratedOffloadArch` is still OK. "Architecture-specific Arch/GPU/Target" would be worse, IMO.

Artem-B:
> Seems straightforward enough. Can you verify that we handle these in the ELF parsing code as well?

It's a good point. Will check.

Artem-B:
> > Seems straightforward enough. Can you verify that we handle these in the ELF parsing code as well?
> 
> It's a good point. Will check.

They all show up the same way as sm_120 in ELF:
```
Flags:                             0x6007802, sm_120
```

cuobjdump and nvdisasm report `f` variant as plain no-suffix variant during SASS disassembly.
They do distinguish and report `a` variant.

jhuber6 (approved):

clang/lib/Basic/Cuda.cpp
yxsamliu, line 162: missing SM_101f?

yxsamliu (approved):

clang/lib/Basic/Cuda.cpp
Artem-B: Good catch. Fixed.

PR #197585, opened 2026-05-13
[CIR]Materialize temp adjustments

erichkeane:
This is a pretty trivial bit of adjustments that have to happen when emitting a materialized temporary, and is effectively a clone of classic codegen.  Our output is effectively identical (other than some minor re-orering problems).


andykaylor (approved):
lgtm

PR #197586, opened 2026-05-13
[scudo] Add test for initFlags()

sadafebrahimi:
Add a test case to verify that initFlags() correctly reads the SCUDO_ALLOCATION_RING_BUFFER_SIZE environment variable and updates the corresponding flag. This increases line coverage for flags.cpp to 100%.


cferris1000 (approved):
LGTM

PR #197587, opened 2026-05-13
[AMDGPU] Fix conflicted literal test. NFC.

rampitec:


rampitec:
* #197587 👈
* `main`

<h2></h2>

llvm/test/MC/AMDGPU/literals.s
```diff
@@ -1986,14 +1986,12 @@ v_add_f64 v[0:1], v[0:1], lit(1)
...
 
-// FIXME: Forced lit() encoding is not preserved after disasm
 v_add_f64 v[0:1], v[0:1], lit(1.0)
-// GFX11: v_add_f64 v[0:1], v[0:1], lit(0x3ff00000)      ; encoding: [0x00,0x00,0x27,0xd7,0x00,0xff,0x01,0x02,0x00,0x00,0xf0,0x3f]
-// GFX12: v_add_f64_e64 v[0:1], v[0:1], lit(0x3ff00000)  ; encoding: [0x00,0x00,0x02,0xd5,0x00,0xff,0x01,0x02,0x00,0x00,0xf0,0x3f]
-// GFX1250-ASM: v_add_f64_e64 v[0:1], v[0:1], lit(0x3ff00000) ; encoding: [0x00,0x00,0x02,0xd5,0x00,0xff,0x01,0x02,0x00,0x00,0xf0,0x3f]
-// GFX1250-DIS: v_add_f64_e64 v[0:1], v[0:1], 1.0       ; encoding: [0x00,0x00,0x02,0xd5,0x00,0xe5,0x01,0x02]
-// NOGFX89: :[[@LINE-5]]:31: error: literal operands are not supported
-// NOSICI: :[[@LINE-6]]:31: error: literal operands are not supported
+// NOGFX11: :[[@LINE-1]]:31: error: invalid operand for instruction
```
jayfoad, line 1990: Why do we reject this? It should be a legal VOP3 with a lit32 operand, right?

rampitec: Yes it should. It never was. IEEE representation of the constant different for FP32 and FP64. When I started to enforce to actually check what lit constant we are producing it started to fail. Basically 1.0 and lit(1.0) are different things when expanded in SP. I.e., it should work, but looks like it never did it correctly. Nor does it work correctly for SP3 as it does not know an operand type. I need to meditate on this, but as it was before it did not work as expected at least. I.e., lit(1.0) would expand exactly to which hex 64-bit value here?

rampitec: I.e., it was only working because it was not properly validated as lit32. But then before yesterday it was kind of 'working' by producing illegal 4-dword encoding. Now it is at least rejected and we may think what is a useful way to implement it.

rampitec: I.e., as of yesterday:
```
llvm-mc -arch=amdgcn -mcpu=gfx1250 -show-encoding <<< 'v_add_f64 v[0:1], v[0:1], lit(1.0)'
	v_add_f64_e64 v[0:1], v[0:1], lit(0x3ff00000) ; encoding: [0x00,0x00,0x02,0xd5,0x00,0xfd,0x01,0x02,0x00,0x00,0xf0,0x3f,0x00,0x00,0x00,0x00]
```
I guess this is trap.

PR #197589, opened 2026-05-13
[VPlan] Make TransformState::get BCast-logic robust

artagnon:
The logic for inserting Broadcasts in a more optimal location in VPTransformState::get is quite fragile, especially around scalable VFs. Fix it, resulting in minor improvements.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlan.cpp
+++ b/llvm/lib/Transforms/Vectorize/VPlan.cpp
@@ -333,40 +333,18 @@ Value *VPTransformState::get(const VPValue *Def, bool NeedsScalar) {
     return Shuf;
   };
 
-  if (!hasScalarValue(Def, {0})) {
-    Value *IRV = Def->getLiveInIRValue();
-    Value *B = GetBroadcastInstrs(IRV);
-    set(Def, B);
-    return B;
-  }
-
   Value *ScalarValue = get(Def, VPLane(0));
-  // If we aren't vectorizing, we can just copy the scalar map values over
-  // to the vector map.
-  if (VF.isScalar()) {
-    set(Def, ScalarValue);
-    return ScalarValue;
-  }
-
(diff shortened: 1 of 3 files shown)


llvm/test/Transforms/LoopVectorize/if-pred-non-void.ll
```diff
@@ -647,12 +647,12 @@ define void @pr30172(ptr nocapture %asd, ptr nocapture %bsd) !dbg !5 {;
 ; UNROLL-NO-VF-NEXT:    [[TMP10:%.*]] = add nsw i32 [[TMP4]], 23
 ; UNROLL-NO-VF-NEXT:    [[TMP11:%.*]] = icmp slt i32 [[TMP3]], 100
 ; UNROLL-NO-VF-NEXT:    [[TMP12:%.*]] = icmp slt i32 [[TMP4]], 100
-; UNROLL-NO-VF-NEXT:    [[TMP13:%.*]] = xor i1 [[TMP11]], true, !dbg [[DBG34:![0-9]+]]
```
fhahn, line 650: hmm, should we keep the debug locations here?

artagnon: Sure, at the cost of carrying logic that loses dbgloc when generalized.

fhahn: Hmm, which part of the other code clears the debug location set in the builder? This may be a separate bug for which we possibly don't have test coverage?

artagnon: Thanks! The issue was in manually setting and restoring IP: using InsertPtGuard fixes the issue and improves code :)

fhahn (approved):
LGTM, thanks

PR #196456, opened 2026-05-08
[AMDGPU] Only src0 and mandatory literals can use literal64

rampitec:

PR #197591, opened 2026-05-13
Revert "[AMDGPU] Only src0 and mandatory literals can use literal64"

boomanaiden154:
Reverts llvm/llvm-project#196456

https://lab.llvm.org/buildbot/#/builders/175/builds/38947

Causes test failures on some buildbots in MC/AMDGPU/literals.s.


rampitec:
Already fixed, sorry: https://github.com/llvm/llvm-project/pull/197587
Conflicted test from different patches.

PR #191641, opened 2026-04-11
[clang-doc][nfc] Default initialize all StringRef members

ilovepi:
Ensure all StringRef members are default initialized to avoid potential bugs.

PR #191637, opened 2026-04-11
[clang-doc] Initialize StringRef members in Info types

ilovepi:
We had a report of some assertion failures in
https://github.com/llvm/llvm-project/pull/190054#issuecomment-4228893309,
and some msan failures in
https://github.com/llvm/llvm-project/pull/190056.

These appear to be due to default constructed StringRef's being used in
some cases. To address, we can provide default initializers that should
prevent such cases from causing further problems.

PR #191570, opened 2026-04-10
[clang-doc] Initialize member variable

ilovepi:
We don't always initialize the IsType field in the current
implementation. We can ensure this field is always initialized to
`false`, and avoid any UB due to garbage data.

PR #190056, opened 2026-04-01
[clang-doc] Merge data into persistent memory

ilovepi:
We have a need for persistent memory for the final info. Since each
group processes a single USR at a time, every USR is only ever processed by
a single thread from the thread pool. This means that we can keep per
thread persistent storage for all the info. There is significant
duplicated data between all the serialized records, so we can just merge
the final/unique items into the persistent arena, and clear out the
scratch/transient arena as we process each record in the bitcode.

The patch adds some APIs to help with managing the data, merging, and
allocation of data in the correct arena. It also safely merges and deep
copies data from the transient arenas into persistent storage that is
never reset until the program completes.

This patch reduces memory by another % over the previous patches,
bringing the total savings over the baseline to 57%. Runtime performance
and benchmarks stay mostly flat with modest improvements.

| Metric | Baseline | Prev | This | Culm% | Seq% |
| :--- | :--- | :--- | :--- | :--- | :--- |
| Time | 920.5s | 991.5s | 987.2s | +7.2% | -0.4% |
| Memory | 86.0G | 40.0G | 36.9G | -57.1% | -8.0% |

| Benchmark | Baseline | Prev | This | Culm% | Seq% |
| :--- | :--- | :--- | :--- | :--- | :--- |
| BM_BitcodeReader_Scale/10 | 67.9us | 72.2us | 72.2us | +6.3% | -0.0% |
| BM_BitcodeReader_Scale/10000 | 70.5ms | 22.5ms | 17.3ms | -75.5% | -23.2% |
| BM_BitcodeReader_Scale/4096 | 23.2ms | 6.6ms | 7.1ms | -69.5% | +7.4% |
| BM_BitcodeReader_Scale/512 | 509.4us | 898.7us | 550.5us | +8.1% | -38.7% |
| BM_BitcodeReader_Scale/64 | 114.8us | 133.7us | 120.8us | +5.2% | -9.6% |
| BM_EmitInfoFunction | 1.6us | 1.9us | 1.8us | +12.9% | -3.3% |
| BM_Index_Insertion/10 | 2.3us | 4.1us | 3.5us | +52.4% | -14.7% |
| BM_Index_Insertion/10000 | 3.1ms | 5.3ms | 4.8ms | +53.5% | -10.0% |
| BM_Index_Insertion/4096 | 1.3ms | 2.1ms | 1.9ms | +50.8% | -9.2% |
| BM_Index_Insertion/512 | 153.6us | 251.8us | 227.0us | +47.8% | -9.9% |
[...]

PR #197595, opened 2026-05-13
Restore clang-doc arena allocation

ilovepi:
This patch restores the commits reverted in 49f8ad172446dd54dd14c2333a7a0f638e37d05a.

It effectively reapplies:

Reapply "[clang-doc][nfc] Default initialize all StringRef members (#191641)" 
Reapply "[clang-doc] Initialize StringRef members in Info types (#191637)" 
Reapply "[clang-doc] Initialize member variable (#191570)" 
Reapply "[clang-doc] Merge data into persistent memory (#190056)" 
Reapply "[clang-doc] Support deep copy between arenas for merging (#190055)"

This version has been updated to account for the new InfoNode<T> paradigm and APIs introduced earlier. Its logic is largely unchanged: it still performs deep copies between the transient and persistent arenas, merges the Info types incrementally, and clears the transient data between merge operations.  It also incorporates a few of the cleanups we applied when trying to fix the Darwin x86 bots prior to reverting, and a new test case for issues brought up at that time.

### DIFF
--- a/clang-tools-extra/clang-doc/Representation.cpp
+++ b/clang-tools-extra/clang-doc/Representation.cpp
@@ -116,7 +116,30 @@ static void reduceChildren(OwningVec<T> &Children,
         Children, [Ptr](const auto &C) { return C.Ptr->USR == Ptr->USR; });
 
     if (It == Children.end()) {
-      Children.push_back(*allocateListNodeTransient<T>(Ptr));
+      InfoNode<T> *NewNode = allocateListNodePersistent<T>(Ptr->USR);
+      NewNode->Ptr->merge(std::move(*Ptr));
+      Children.push_back(*NewNode);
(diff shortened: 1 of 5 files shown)


ilovepi:
@evelez7 I haven't tried this on Darwin yet, and won't be able to until tomorrow when I'm back in the office. It did pass on Linux under ASan + UBSan.

ilovepi:
> @evelez7 I haven't tried this on Darwin yet, and won't be able to until tomorrow when I'm back in the office. It did pass on Linux under ASan + UBSan.

And all seems well on my x86_64 Darwin machine w/ XCode. Seems like we should be OK with the reland of these last bits. 😄

evelez7 (approved):
Passes on mine as well.

PR #197596, opened 2026-05-14
[clang-sycl-linker] Remove dead and unnecessary check for no symbols image

YuriPlyakhin:
The `if (SI.Symbols.empty()) continue;` guard in `runSYCLLink` was unreachable: `collectEntryPoints` always calls `llvm::offloading::sycl::writeSymbolTable`, which emits at least a 4-byte `SymbolTableHeader` (`Count=0`) even when the input has no entry points.
The check could never fire and was misleading suggesting that modules without kernels would be dropped.
I checked other similar tools and did not find logic to filter out images, if there are no entry points. This doesn't look like responsibility of clang-sycl-linker.
If this scenario is required, we can consider adding it later with a proper check.

### DIFF
--- a/clang/test/Tooling/clang-sycl-linker.ll
+++ b/clang/test/Tooling/clang-sycl-linker.ll
@@ -75,7 +75,7 @@
 ;
 ; Input with no entry points still produces an offload image.
 ; RUN: llvm-as %t/no-entry-points.ll -o %t/no-entry-points.bc
-; RUN: clang-sycl-linker --dry-run -triple=spirv64 %t/no-entry-points.bc -o %t/no-entry-points.out
+; RUN: clang-sycl-linker -triple=spirv64 %t/no-entry-points.bc -o %t/no-entry-points.out
 ; RUN: llvm-objdump --offloading %t/no-entry-points.out | FileCheck %s --check-prefix=NO-ENTRY-POINTS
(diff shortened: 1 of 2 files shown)


YuriPlyakhin:
@maksimsab , fyi

clang/test/Tooling/clang-sycl-linker.ll
```diff
@@ -72,6 +72,13 @@
 ; RUN: not clang-sycl-linker --dry-run %t/input1.bc %t/input2.bc -o a.out 2>&1 \
 ; RUN:   | FileCheck %s --check-prefix=NOTARGET
 ; NOTARGET: Target triple must be specified
+;
+; Input with no entry points still produces an offload image.
+; RUN: llvm-as %t/no-entry-points.ll -o %t/no-entry-points.bc
+; RUN: clang-sycl-linker --dry-run -triple=spirv64 %t/no-entry-points.bc -o %t/no-entry-points.out
```
bader:
```suggestion
; RUN: clang-sycl-linker -triple=spirv64 %t/no-entry-points.bc -o %t/no-entry-points.out
```
`clang-sycl-linker` cannot produce `%t/no-entry-points.out` in "dry run" mode.

bader (approved):

sarnex (approved):

PR #195995, opened 2026-05-06
[Clang] Transform lambda's constraints when instantiating parameter mapping

zyn0217:
This way we can remove a few workarounds of lambda expressions where outer template arguments of concepts have to be preserved through ImplicitConceptSpecializationDecls.

Fixes #193944

PR #197598, opened 2026-05-14
[clang] NFC: add asserts and fixes for enforcing template parameters have valid positions

mizvekov:
Some tests are violating these assertions, so they are commented out.

For the test in `clang/test/SemaTemplate/concepts.cpp`, that was broken by #195995 and needs a partial revert at least.

### DIFF
--- a/clang/include/clang/AST/ASTContext.h
+++ b/clang/include/clang/AST/ASTContext.h
@@ -1998,8 +1998,7 @@ public:
   QualType getSubstBuiltinTemplatePack(const TemplateArgument &ArgPack);
 
   QualType
-  getTemplateTypeParmType(unsigned Depth, unsigned Index,
-                          bool ParameterPack,
+  getTemplateTypeParmType(int Depth, int Index, bool ParameterPack,
                           TemplateTypeParmDecl *ParmDecl = nullptr) const;
 
   QualType getCanonicalTemplateSpecializationType(
--- a/clang/include/clang/AST/DeclTemplate.h
+++ b/clang/include/clang/AST/DeclTemplate.h
@@ -1128,16 +1128,9 @@ protected:
   unsigned Depth : DepthWidth;
   unsigned Position : PositionWidth;
 
-  static constexpr unsigned MaxDepth = (1U << DepthWidth) - 1;
-  static constexpr unsigned MaxPosition = (1U << PositionWidth) - 1;
-
-  TemplateParmPosition(unsigned D, unsigned P) : Depth(D), Position(P) {
-    // The input may fill maximum values to show that it is invalid.
-    // Add one here to convert it to zero.
-    assert((D + 1) <= MaxDepth &&
-           "The depth of template parmeter position is more than 2^20!");
-    assert((P + 1) <= MaxPosition &&
-           "The position of template parmeter position is more than 2^12!");
+  TemplateParmPosition(int D, int P) {
+    setDepth(D);
+    setPosition(P);
   }
 
 public:
@@ -1145,17 +1138,17 @@ public:
 
   /// Get the nesting depth of the template parameter.
   unsigned getDepth() const { return Depth; }
-  void setDepth(unsigned D) {
-    assert((D + 1) <= MaxDepth &&
-           "The depth of template parmeter position is more than 2^20!");
+  void setDepth(int D) {
+    assert(D >= 0 && "The depth cannot be negative");
+    assert(D < (1 << DepthWidth) && "The depth is too large");
     Depth = D;
   }
 
   /// Get the position of the template parameter within its parameter list.
   unsigned getPosition() const { return Position; }
-  void setPosition(unsigned P) {
-    assert((P + 1) <= MaxPosition &&
-           "The position of template parmeter position is more than 2^12!");
+  void setPosition(int P) {
+    assert(P >= 0 && "The position cannot be negative");
+    assert(P < (1 << PositionWidth) && "The position is too large");
     Position = P;
   }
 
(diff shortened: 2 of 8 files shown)


clang/include/clang/AST/ASTContext.h
```diff
@@ -1998,8 +1998,7 @@ class ASTContext : public RefCountedBase<ASTContext> {
   QualType getSubstBuiltinTemplatePack(const TemplateArgument &ArgPack);
 
   QualType
-  getTemplateTypeParmType(unsigned Depth, unsigned Index,
-                          bool ParameterPack,
+  getTemplateTypeParmType(int Depth, int Index, bool ParameterPack,
```
erichkeane, line 2001: I don't really get why a negative value is valuable here?  What am I missing?   Why wouldn't negatives just wrap to above the MaxDepth/MaxPosition?

clang/lib/Sema/SemaTemplateDeductionGuide.cpp
```diff
@@ -335,20 +336,131 @@ TemplateTypeParmDecl *transformTemplateTypeParam(
...
-  // the index of the parameter once it's done.
-  auto *NewParam = cast<NonTypeTemplateOrTemplateTemplateParmDecl>(
-      SemaRef.SubstDecl(OldParam, DC, Args));
-  NewParam->setPosition(NewIndex);
-  NewParam->setDepth(NewDepth);
-  return NewParam;
+                       NonTypeTemplateParmDecl *TTP, unsigned NewDepth,
+                       unsigned NewIndex,
+                       MultiLevelTemplateArgumentList &Args) {
+  NonTypeTemplateParmDecl *NewTTP;
```
erichkeane, line 345: This seems quite a bit more than just 'add asserts and fixes'.  Can we clarify what is going on here?  It seems this is way more than that.

clang/test/SemaTemplate/deduction-guide.cpp
```diff
@@ -821,11 +821,11 @@ BB b{};
 // CHECK-LABEL: Dumping GH133132::<deduction guide for BB>:
 // CHECK-NEXT:  FunctionTemplateDecl {{.+}} implicit <deduction guide for BB>
 // CHECK-NEXT:  |-TemplateTemplateParmDecl {{.+}} depth 0 index 0 _X
-// CHECK-NEXT:  | |-TemplateTypeParmDecl {{.+}} class depth 0 index 0
+// CHECK-NEXT:  | |-TemplateTypeParmDecl {{.+}} class depth 1 index 0
```
erichkeane, line 824: THIS definitely seems suspicious.  This seems like it would cause us to have to instantiate this deduction guide differently.

erichkeane (commented):
This is one of those that seems pretty strange right now.  There is a LOT of work happening in SemaTemplateDeductionDecl that seems to be more than just adding asserts (AND does a ton of work without tests).  Can we break this up into smaller tasks that are clear with what they are accomplishing?

Also, the unsigned->int conversion doesn't really seem justified/clear what is going on.

clang/include/clang/AST/ASTContext.h
mizvekov: The negatives give a hint they were produced from a subtraction in the template instantiator, basically trying to substitute a template with more levels than depth.

A number too high on the other hand would tell you that you basically tried to instantiate a template which is too nested.

clang/lib/Sema/SemaTemplateDeductionGuide.cpp
mizvekov: Yeah, so basically the CTAD implementation was producing template template parameters with wrong depth for their template parameters. It would rewrite the position of the template template parameter itself, but then it forgot it had to update all of its template parameters such that it had depth + 1 in relation to that new depth...

For the main template parameters themselves, It was also relying on a trick where it produced a template parameters with overflowed depth, and then 'fixed' that by overwriting the correct position. This is extremely unfriendly to asserts.

clang/test/SemaTemplate/deduction-guide.cpp
mizvekov: This is right change, the template parameters of a template template parameter have to have depth + 1 in relation to its parent.

clang/lib/Sema/SemaTemplateDeductionGuide.cpp
erichkeane: Hmm... i see.  This is still quite a bit of work happening here.  Is there any way to have this broken up into somethign I can digest?  This kinda comes down to "just do everything" in 1 commit, and I'm a bit overwhelmed trying to figure out what I'm reading.

mizvekov:
> This is one of those that seems pretty strange right now. There is a LOT of work happening in SemaTemplateDeductionDecl that seems to be more than just adding asserts (AND does a ton of work without tests). Can we break this up into smaller tasks that are clear with what they are accomplishing?
> 
> Also, the unsigned->int conversion doesn't really seem justified/clear what is going on.

So I started this PR just adding the assert and then commented out the broken tests.
Then I fixed the asserts for all of the CTAD cases. I had updated the PR title and description to reflect that it has fixes as well.

So I could maybe separate out the CTAD fix from the rest, but this is all self-contained in the changes to `clang/lib/Sema/SemaTemplateDeductionGuide.cpp` (and corresponding AST dump changes to CTAD tests), that I feel this is self isolated enough?

Also, besides that fix, there is another single one liner which is the only change in `clang/lib/Sema/SemaTemplateDeduction.cpp`. It avoids producing an overflowed TTP by instead just substituting its template parameters instead of the whole self.

The rest of the changes are just adding the assertions to the three template parameter kinds.

clang/lib/Sema/SemaTemplateDeductionGuide.cpp
mizvekov: K I'll split the CTAD changes into another patch.

zyn0217 (approved):

clang/lib/Sema/SemaTemplateDeductionGuide.cpp
mizvekov: Done in https://github.com/llvm/llvm-project/pull/197611

clang/include/clang/AST/ASTContext.h
mizvekov: In simpler words:

Negatives -> Bugs
Too larges -> Exceeding implementation limits.

erichkeane (approved):

PR #197600, opened 2026-05-14
[clang] Enable crash reproducers when multiple -arch flags are involved

jroelofs:
rdar://176327788

### DIFF
--- a/clang/lib/Driver/Driver.cpp
+++ b/clang/lib/Driver/Driver.cpp
@@ -75,6 +75,7 @@
 #include "clang/ScalableStaticAnalysisFramework/SSAFForceLinker.h" // IWYU pragma: keep
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/ScopeExit.h"
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringExtras.h"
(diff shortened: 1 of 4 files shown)


clang/test/Driver/crash-report-multi-arch.c
```diff
@@ -0,0 +1,86 @@
...
+// RUN: rm -f %t/crash-report-*
+
+// RUN: env CLANG_CRASH_DIAGNOSTICS_DIR=%t \
+// RUN:   not %crash_opt %clang %s -arch arm64 -arch arm64e -arch x86_64 -fsyntax-only -DCRASH_arm64e 2>&1 | \
+// RUN:   FileCheck --check-prefix=CHECK %s
+// RUN: cat %t/crash-report-*.sh | FileCheck --check-prefix=CHECKSH-arm64e %s
+// RUN: rm -f %t/crash-report-*
+
+// If multiple arches crash, capture the reproducer for the first one.
+// one in the list.
```
vsapsai: Weird wording "the first one. one in the list"
Guess it is the result of previous editing.

clang/lib/Driver/Driver.cpp
vsapsai: The comment isn't entirely correct anymore.

vsapsai (commented):
Looks good to me, just a few nits. Still plan to play a little bit with the test case but it looks comprehensive.

Personally, I wouldn't test arm64e but I think it is representative of an actual usage, so there is value in keeping it like that.

vsapsai (approved):
Tested locally and everything seems to be working fine.

Not blocking on minor comment tweaks, trust you to address those.

clang/test/Driver/crash-report-multi-arch.c
jroelofs: oops

PR #197603, opened 2026-05-14
[OpenMP] Fix missing install-openmp component

jhuber6:
Summary:
This pattern is consistent throughout all the runtimes and is what the
top-level `install-openmp-<triple>` corresponds to. It should be
provided and used.


Meinersbur (approved):
LGTM

PR #197605, opened 2026-05-14
[libc] Fix tuple std injection in userspace builds

SchrodingerZhu:
Fix the namespace problem with building overlay under LLVM/LIBCXX userland.

libc++ utilizes inline namespace `__1` to version its ABI, this makes std namespace inject inside libc ambiguous during argument dependent lookup. This PR switches to libc++ defined stl namespace when the macro is detected. 

Assisted-by: Codex with gpt-5.5 high fast

### DIFF
--- a/libc/cmake/modules/LLVMLibCTestRules.cmake
+++ b/libc/cmake/modules/LLVMLibCTestRules.cmake
@@ -29,6 +29,7 @@ function(_get_common_test_compile_options output_var c_test flags)
       list(APPEND compile_options "-fno-exceptions")
       list(APPEND compile_options "-fno-unwind-tables")
       list(APPEND compile_options "-fno-asynchronous-unwind-tables")
+      list(APPEND compile_options "-nostdinc++")
       if(NOT c_test)
         list(APPEND compile_options "-fno-rtti")
       endif()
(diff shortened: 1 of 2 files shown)


mchoo7:
Resigning as reviewer as I'm not familiar with these. As the creator of this file, @jhuber6 might be able to tell if this is a correct fix.

jhuber6 (commented):
I wonder if we could solve this with like `#has_include("<tuple>")` or something

SchrodingerZhu:
@jhuber6 I am trying it but it seems to bring more complications. Just to confirm that you want me to push forward towards this direction.

SchrodingerZhu:
@jhuber6 I don't really catch the reason why the fuzzer build it failing. Also, the suggested approach seems to bring in too much change. Can we just use the original fix for now?

I understand that original fix only works if the user include stl in unit test before our headers but I think that is enough for our current purpose.

jhuber6 (commented):
We should probably just remove structured bindings and accept using `cpp::tie`. They should be equivalent and the `std` injection isn't worth the convenience.

michaelrj-google (changes requested):
I'd like to have a bit more discussion before we start directly including public libc++ headers.

If we want to do something like hand in hand we would want to specifically share the internal headers.

PR #197607, opened 2026-05-14
[Instrumentor] Introduce BasePointerIO to communicate base pointer information

jdoerfert:
Loads, stores, and later probably calls, can request a base pointer info
object from the user runtime. This object is queried right after the
base pointer of the operation is defined, and then passed to the
pre/post runtime calls of the loads and stores. This allows users to
inspect pointers early and once, but provide the analysis results to all
operations that might be executed in loops. A potential use case is to
lookup the size and start of the underlying object and then provide
those to the access runtime calls for in-bounds checking.

### DIFF
--- a/llvm/include/llvm/Transforms/IPO/Instrumentor.h
+++ b/llvm/include/llvm/Transforms/IPO/Instrumentor.h
@@ -178,7 +178,8 @@ struct InstrumentationLocation {
     BASIC_BLOCK_POST,
     INSTRUCTION_PRE,
     INSTRUCTION_POST,
-    Last = INSTRUCTION_POST,
+    SPECIAL_VALUE,
+    Last = SPECIAL_VALUE,
   };
 
   /// Construct an instrumentation location that is not instrumenting an
@@ -220,6 +221,8 @@ struct InstrumentationLocation {
       return "instruction_pre";
     case INSTRUCTION_POST:
       return "instruction_post";
+    case SPECIAL_VALUE:
+      return "special_value";
     }
     llvm_unreachable("Invalid kind!");
   }
@@ -237,6 +240,7 @@ struct InstrumentationLocation {
         .Case("basic_block_post", BASIC_BLOCK_POST)
         .Case("instruction_pre", INSTRUCTION_PRE)
         .Case("instruction_post", INSTRUCTION_POST)
+        .Case("special_value", SPECIAL_VALUE)
         .Default(Last);
   }
 
@@ -254,6 +258,7 @@ struct InstrumentationLocation {
     case FUNCTION_POST:
     case BASIC_BLOCK_POST:
     case INSTRUCTION_POST:
+    case SPECIAL_VALUE:
       return false;
     }
     llvm_unreachable("Invalid kind!");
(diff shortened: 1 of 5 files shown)


llvm/include/llvm/Transforms/IPO/Instrumentor.h
```diff
@@ -707,6 +721,46 @@ struct UnreachableIO final : public InstructionIO<Instruction::Unreachable> {
   }
 };
 
+// Special instrumentation opportunity for base pointers of memory operations.
+struct BasePointerIO : public InstrumentationOpportunity {
```
kevinsala:
```suggestion
struct BasePointerIO final : public InstrumentationOpportunity {
```

llvm/lib/Transforms/IPO/Instrumentor.cpp
```diff
@@ -525,7 +527,9 @@ void InstrumentationConfig::populate(InstrumentorIRBuilderTy &IIRB) {
 void InstrumentationConfig::addChoice(InstrumentationOpportunity &IO,
                                       LLVMContext &Ctx) {
   auto *&ICPtr = IChoices[IO.getLocationKind()][IO.getName()];
-  if (ICPtr) {
+  if (ICPtr &&
+      (IO.getLocationKind() != InstrumentationLocation::SPECIAL_VALUE ||
+       ICPtr->getLocationKind() != InstrumentationLocation::SPECIAL_VALUE)) {
```
kevinsala: Why was this condition necessary?

Shouldn't we emit a warning when registering two opportunities with special value location and the same name?

llvm/lib/Transforms/IPO/Instrumentor.cpp
```diff
@@ -535,6 +539,62 @@ void InstrumentationConfig::addChoice(InstrumentationOpportunity &IO,
...
+      IIRB.IRB.SetInsertPoint(*IP);
+    } else {
+      IIRB.Ctx.diagnose(DiagnosticInfoInstrumentation(
+          "Base pointer info could not be placed, passing nullptr.",
+          DS_Warning));
+      return BPI = Constant::getNullValue(BPIO->getRetTy(IIRB.Ctx));
+    }
+  } else if (isa<Constant>(VPtr) || isa<Argument>(VPtr))
+    IIRB.IRB.SetInsertPointPastAllocas(IIRB.IRB.GetInsertBlock()->getParent());
+  else {
```
```suggestion
  } else if (isa<Constant>(VPtr) || isa<Argument>(VPtr)) {
    IIRB.IRB.SetInsertPointPastAllocas(IIRB.IRB.GetInsertBlock()->getParent());
  } else {
```

llvm/lib/Transforms/IPO/Instrumentor.cpp
```diff
@@ -535,6 +539,62 @@ void InstrumentationConfig::addChoice(InstrumentationOpportunity &IO,
...
+    VPtr->dump();
```
kevinsala: `LLVM_DEBUG`?

llvm/lib/Transforms/IPO/Instrumentor.cpp
```diff
@@ -535,6 +539,62 @@ void InstrumentationConfig::addChoice(InstrumentationOpportunity &IO,
   ICPtr = &IO;
 }
 
+Value *
+InstrumentationConfig::getBasePointerInfo(Value &V,
+                                          InstrumentorIRBuilderTy &IIRB) {
+  Function *Fn = IIRB.IRB.GetInsertBlock()->getParent();
+
+  Value *VPtr;
```
```suggestion
  Value *Obj;
```

kevinsala (approved):
LGTM except the few comments

llvm/lib/Transforms/IPO/Instrumentor.cpp
jdoerfert: I think we can, and want, to allow more than one. I can remove it for now until we get to it.

kevinsala: If we want more than one with the same location and name, we would need to have a vector or set of opportunities. Otherwise, as it is now, the old opportunity gets replaced by the new one with same location and name.

I'm fine restricting that case for now.

PR #197611, opened 2026-05-14
[clang] CTAD: fix transformation of template template parameters

mizvekov:
This fixes the CTAD template parameter transforms so they produce template template parameters which have correct depth for their own template parameters.

This also stops calling SubstDecl directly on the non-type template parameters, so that a template parameter with correct position is produced directly, instead of manually fixing that up later. This helps #197598 by making it possible to add assertions that the positions are always valid.


zyn0217 (approved):

erichkeane (commented):
I'm still a touch tentative on the NonTypeTemplate version of transformTemplateParam as there is a lot of work happening there.  But I'll think about it for  the night and revisit in the morning.

erichkeane (approved):

PR #197612, opened 2026-05-14
[Frontend/Offloading] Fix use-after-reallocation in sycl::writeSymbolTable

YuriPlyakhin:
`writeSymbolTable` took raw pointers into the `SmallString` buffer
  (`Header`, `Entries`) and then called `Out.append()` / `Out.push_back()`
  inside the loop to write string data.  When the `SmallString` needed to
  grow, it reallocated, silently invalidating those pointers.  All writes
  through `Entries[I]` after the first reallocation were undefined
  behaviour; UBSAN caught this as a crash (exit code -6 / SIGABRT) on the
  sanitizer-x86_64-linux-bootstrap-ubsan builder.

  The fix pre-computes the total buffer size (header + entry array +
  all null-terminated name strings) and calls `reserve()` before any
  pointers are taken, guaranteeing that the subsequent `append` and
  `push_back` calls cannot trigger a reallocation.


YuriPlyakhin:
I reproduced sanitizer-x86_64-linux-bootstrap-ubsan fail locally and verified that fix resolves the problem.

sarnex (approved):

PR #197618, opened 2026-05-14
[SelectionDAG] Widen <2 x T> vector types for atomic store

jofrn:
Vector types of 2 elements must be widened. This change does this
for vector types of atomic store in SelectionDAG so that it can
translate aligned vectors of >1 size.

Store-side counterpart to #148897. Stacked on top of #197166; and below of #197619.

### DIFF
--- a/llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
+++ b/llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
@@ -6611,6 +6611,23 @@ static SDValue coerceLoadedValue(SDValue LdOp, EVT FirstVT, EVT WidenVT,
   return LdOp;
 }
 
+/// Inverse of coerceLoadedValue: pull a FirstVT-sized scalar/vector out of the
+/// widened value so it can be issued in a single atomic store.
+static SDValue coerceStoredValue(SDValue StVal, EVT FirstVT, EVT WidenVT,
+                                 TypeSize FirstVTWidth, const SDLoc &dl,
+                                 SelectionDAG &DAG) {
+  TypeSize WidenWidth = WidenVT.getSizeInBits();
+  if (!FirstVT.isVector()) {
+    unsigned NumElts =
+        WidenWidth.getFixedValue() / FirstVTWidth.getFixedValue();
+    EVT NewVecVT = EVT::getVectorVT(*DAG.getContext(), FirstVT, NumElts);
+    SDValue VecOp = DAG.getNode(ISD::BITCAST, dl, NewVecVT, StVal);
+    return DAG.getExtractVectorElt(dl, FirstVT, VecOp, 0);
+  }
+  assert(FirstVT == WidenVT && "First value type must equal widen value type");
+  return StVal;
+}
+
 static std::optional<EVT> findMemType(SelectionDAG &DAG,
                                       const TargetLowering &TLI, unsigned Width,
                                       EVT WidenVT, unsigned Align,
@@ -7448,6 +7465,9 @@ bool DAGTypeLegalizer::WidenVectorOperand(SDNode *N, unsigned OpNo) {
   case ISD::EXTRACT_SUBVECTOR:  Res = WidenVecOp_EXTRACT_SUBVECTOR(N); break;
   case ISD::EXTRACT_VECTOR_ELT: Res = WidenVecOp_EXTRACT_VECTOR_ELT(N); break;
   case ISD::STORE:              Res = WidenVecOp_STORE(N); break;
+  case ISD::ATOMIC_STORE:
+    Res = WidenVecOp_ATOMIC_STORE(cast<AtomicSDNode>(N));
+    break;
   case ISD::VP_STORE:           Res = WidenVecOp_VP_STORE(N, OpNo); break;
   case ISD::EXPERIMENTAL_VP_STRIDED_STORE:
     Res = WidenVecOp_VP_STRIDED_STORE(N, OpNo);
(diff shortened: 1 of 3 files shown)


llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
```diff
@@ -8061,6 +8064,59 @@ SDValue DAGTypeLegalizer::WidenVecOp_STORE(SDNode *N) {
   report_fatal_error("Unable to widen vector store");
 }
 
+/// Inverse of coerceLoadedValue: pull a FirstVT-sized scalar/vector out of the
+/// widened value so it can be issued in a single atomic store.
+static SDValue coerceStoredValue(SDValue StVal, EVT FirstVT, EVT WidenVT,
```
arsenm: Can you move this to be next to coerceLoadedValue

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
```diff
@@ -8061,6 +8064,59 @@ SDValue DAGTypeLegalizer::WidenVecOp_STORE(SDNode *N) {
...
+  TypeSize StWidth = StVT.getSizeInBits();
+  TypeSize WidenWidth = WidenVT.getSizeInBits();
+  TypeSize WidthDiff = WidenWidth - StWidth;
+
+  // Find the vector type that can store the original memory width in one
+  // atomic operation.
+  std::optional<EVT> FirstVT =
+      findMemType(DAG, TLI, StWidth.getKnownMinValue(), WidenVT, /*StAlign=*/0,
+                  WidthDiff.getKnownMinValue());
+
```
arsenm:
```suggestion
```

arsenm: Why is this passing 0 for the align instead of the actual alignment?

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
jofrn: Yes, will do so.

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
```diff
@@ -6608,6 +6608,24 @@ static SDValue coerceLoadedValue(SDValue LdOp, EVT FirstVT, EVT WidenVT,
   return LdOp;
 }
 
+/// Inverse of coerceLoadedValue: pull a FirstVT-sized scalar/vector out of the
+/// widened value so it can be issued in a single atomic store.
+static SDValue coerceStoredValue(SDValue StVal, EVT FirstVT, EVT WidenVT,
+                                 TypeSize FirstVTWidth, SDLoc dl,
```
RKSimon:
```suggestion
                                 TypeSize FirstVTWidth, const SDLoc &dl,
```

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
```diff
@@ -6608,6 +6608,24 @@ static SDValue coerceLoadedValue(SDValue LdOp, EVT FirstVT, EVT WidenVT,
...
+                                 SelectionDAG &DAG) {
+  TypeSize WidenWidth = WidenVT.getSizeInBits();
+  if (!FirstVT.isVector()) {
+    unsigned NumElts =
+        WidenWidth.getFixedValue() / FirstVTWidth.getFixedValue();
+    EVT NewVecVT = EVT::getVectorVT(*DAG.getContext(), FirstVT, NumElts);
+    SDValue VecOp = DAG.getNode(ISD::BITCAST, dl, NewVecVT, StVal);
+    return DAG.getNode(ISD::EXTRACT_VECTOR_ELT, dl, FirstVT, VecOp,
```
RKSimon:
```suggestion
    return DAG.getExtractVectorElt(dl, FirstVT, VecOp, 0);
```

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
jofrn: Like atomic loads, for atomic stores, this causes a widen beyond the value's width: e.g. `<2 x i8>` writes beyond object (e.g. `%v` of `<2 x i8> %v`). That is, if we pass align, then we get `movl %eax, (%rdi)`, which writes to unrelated / undefined bytes relative to `%v`.

RKSimon: A brief comment explaining that would be useful for future reference

jofrn: Ok, added one! Thanks.

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
```diff
@@ -8064,6 +8084,42 @@ SDValue DAGTypeLegalizer::WidenVecOp_STORE(SDNode *N) {
   report_fatal_error("Unable to widen vector store");
 }
 
+SDValue DAGTypeLegalizer::WidenVecOp_ATOMIC_STORE(AtomicSDNode *ST) {
+  EVT StVT = ST->getMemoryVT();
+  SDLoc dl(ST);
+  assert(StVT.isVector() && "Expected vector");
```
arsenm:
```suggestion
```

Redundant

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
```diff
@@ -8064,6 +8084,42 @@ SDValue DAGTypeLegalizer::WidenVecOp_STORE(SDNode *N) {
...
+
+  SDValue StVal = GetWidenedVector(ST->getVal());
+  EVT WidenVT = StVal.getValueType();
+  assert(WidenVT.isVector() && "Expected vector");
+  assert(StVT.isScalableVector() == WidenVT.isScalableVector() &&
+         "Must be scalable");
+  assert(StVT.getVectorElementType() == WidenVT.getVectorElementType() &&
+         "Expected equivalent element types");
```
arsenm: All these asserts seem redundant (e.g., getVector* will assert on !isVector anyway), or belong in findMemType

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
jofrn: Removed. Thank you.

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
jofrn: Removed. `findMemType` keeps track of the invariants well.

arsenm (approved):

RKSimon (approved):
LGTM

PR #148898, opened 2025-07-15
[X86] Remove extra MOV after widening atomic load

jofrn:
This change adds patterns to optimize out an extra MOV
present after widening the atomic load.

PR #197619, opened 2026-05-14
[X86] Remove extra MOV after widening atomic store

jofrn:
This change adds patterns to optimize out an extra MOV present after
widening the atomic store. Covers `<2 x i8>` (SSE4.1+), `<2 x i16>`,
`<4 x i8>`, `<2 x i32>`, `<2 x float>`, `<4 x i16>`,
`<2 x ptr addrspace(270)>`.

Store-side counterpart to #148898. Stacked on top of #197618; and below of #197860.

### DIFF
--- a/llvm/include/llvm/Target/TargetSelectionDAG.td
+++ b/llvm/include/llvm/Target/TargetSelectionDAG.td
@@ -2394,6 +2394,41 @@ def atomic_store_128 :
   let MemoryVT = i128;
 }
 
+// Matches store or atomic_store, no alignment requirement.
+def any_store : PatFrags<(ops node:$val, node:$ptr),
+                         [(store node:$val, node:$ptr),
+                          (atomic_store node:$val, node:$ptr)]>;
+
+def any_store_8 : PatFrags<(ops node:$val, node:$ptr),
+                           [(store node:$val, node:$ptr),
+                            (atomic_store node:$val, node:$ptr)]> {
+  let MemoryVT = i8;
+}
+
+def any_store_16 : PatFrags<(ops node:$val, node:$ptr),
+                            [(store node:$val, node:$ptr),
+                             (atomic_store node:$val, node:$ptr)]> {
+  let MemoryVT = i16;
+}
+
+def any_store_32 : PatFrags<(ops node:$val, node:$ptr),
+                            [(store node:$val, node:$ptr),
+                             (atomic_store node:$val, node:$ptr)]> {
+  let MemoryVT = i32;
+}
+
+def any_store_64 : PatFrags<(ops node:$val, node:$ptr),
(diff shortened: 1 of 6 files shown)


llvm/lib/Target/X86/X86InstrCompiler.td
```diff
@@ -1242,6 +1242,105 @@ def : Pat<(v4i32 (atomic_load_128_v4i32 addr:$src)),
 def : Pat<(v4i32 (atomic_load_128_v4i32 addr:$src)),
           (VMOVAPDZ128rm addr:$src)>, Requires<[HasAVX512]>;
 
+// store atomic <2 x i8>
+def : Pat<(atomic_store_16
```
arsenm: Are there existing non-store patterns for this? Can do a better job avoiding duplication (i.e, there should be a PatFrag that covers atomic and non-atomic cases)

llvm/lib/Target/X86/X86InstrFragmentsSIMD.td
```diff
@@ -1254,6 +1254,11 @@ def extloadv4f16 : PatFrag<(ops node:$ptr), (extloadvf16 node:$ptr)>;
 def extloadv8f16 : PatFrag<(ops node:$ptr), (extloadvf16 node:$ptr)>;
 def extloadv16f16 : PatFrag<(ops node:$ptr), (extloadvf16 node:$ptr)>;
 
+// Matches either 'store' or 'atomic_store' (any alignment, any ordering).
+def memstore : PatFrags<(ops node:$val, node:$ptr),
```
RKSimon: if you're adding atomic store support to alignedstore - would `unalignedstore` be better here for equivalence?

arsenm: The name isn't clear, and this can go in the generic PatFrags (though that goes for a number of the PatFags here)

jofrn: `unalignedstore` sounds good.

jofrn: https://github.com/llvm/llvm-project/pull/197619#discussion_r3258593623 okie.

arsenm: maybe `any_store_<size>`?

llvm/lib/Target/X86/X86InstrCompiler.td
jofrn: It should be good now.

RKSimon (changes requested):
please fix merge conflicts

arsenm (approved):

RKSimon (approved):

PR #197620, opened 2026-05-14
[Clang][AArch64] Fix crash with large arguments to MTE built-ins

igorkudrin:
The second argument to `__builtin_arm_irg()` and `__builtin_arm_gmi()`
is expected to be 64-bit. When a wider type is passed, the compiler
still generates a `zext` instruction, leading to a backend error:

```
> cat test.c
unsigned test(void* a, unsigned __int128 m) {
  return __builtin_arm_gmi(a, m);
}
> clang -target aarch64 -march=armv9+memtag -S test.c
fatal error: error in backend: Cannot select: ... i64,ch = load<... zext from i128>
```

This is fixed by applying the required implicit conversions to the argument.


igorkudrin:
Ping

efriedma-quic (commented):
If we think the argument is always supposed to fit into 64 bits, we should do the truncation in Sema, as a conversion to uint64_t or something like that.  The generated code is basically the same, but the implicit conversion in Sema will generate a -Wconversion warning.  (Which is not on by default, but better than nothing.)

igorkudrin:
> If we think the argument is always supposed to fit into 64 bits, we should do the truncation in Sema, as a conversion to uint64_t or something like that. The generated code is basically the same, but the implicit conversion in Sema will generate a -Wconversion warning. (Which is not on by default, but better than nothing.)

Are you suggesting calling `Sema::CheckImplicitConversion()` explicitly? It seems `Sema::ImpCastExprToType()` does not produce the diagnostic.

efriedma-quic:
Doesn't CheckImplicitConversions() normally get called later through CheckCompletedExpr()?

Anyway, the helper you want is probably convertArgumentToType().

igorkudrin:
> Doesn't CheckImplicitConversions() normally get called later through CheckCompletedExpr()?

Maybe it doesn't walk through all the parameters of `CallExpr`. The diagnostic is only produced for the return type, unless I add `CheckImplicitConversion()` explicitly in `SemaARM::BuiltinARMMemoryTaggingCall()`.

> Anyway, the helper you want is probably convertArgumentToType().

This is a static function in `clang/lib/Sema/SemaChecking.cpp` and is not available in `clang/lib/Sema/SemaARM.cpp`. And while it does add all the required implicit conversions, it still does not produce the diagnostics.

efriedma-quic:
In `void f(void g(long long), __uint128_t x) { g(x) ;}`, the stack trace of the diagnostic is ActOnFinishFullExpr->CheckCompletedExpr->AnalyzeImplicitConversions .  I don't see any reason for intrinsics to behave differently.

igorkudrin:
I've added the conversion to `SemaARM::BuiltinARMMemoryTaggingCall()` using `Sema::PerformCopyInitialization()`, similar to `convertArgumentToType()` and some other places. It does not emit `-Wconversion` warnings for arguments, but fixing that is out of scope for this PR.

efriedma-quic (approved):
LGTM

PR #196481, opened 2026-05-08
[Clang][AArch64] Fix possible miscompilation with `__builtin_arm_gmi`

igorkudrin:
The `@llvm.aarch64.gmi` intrinsic returns a 64-bit value, but `SemaARM::BuiltinARMMemoryTaggingCall()` incorrectly sets the type of the call instruction to `Context.IntTy`. This results in miscompilations when a truncate instruction is required but missing from the generated code. Note that in the `-O0` case below a 64-bit value is stored in a 32-bit stack slot:

```
> cat test.c
int test(void* a, unsigned m) {
  return __builtin_arm_gmi(a, m);
}
> clang -target aarch64 -march=armv9+memtag -O0 -S -emit-llvm test.c -o -
define dso_local i32 @test(ptr noundef %0, i32 noundef %1) #0 {
  %3 = alloca i32, align 4
  ...
  %9 = call i64 @llvm.aarch64.gmi(ptr %6, i64 %8)
  store i64 %9, ptr %3, align 4
  %10 = load i32, ptr %3, align 4
  ret i32 %10
}
> clang -target aarch64 -march=armv9+memtag -O3 -S -emit-llvm test.c -o -
define dso_local i32 @test(ptr nocapture noundef readnone %0, i32 noundef %1) local_unnamed_addr #0 {
  ret i32 undef
}
```

This patch aligns the MTE built-in prototypes with the ARM documentation [1] and removes now-unnecessary `setType()` calls.

[1] https://developer.arm.com/documentation/101028/0012/10--Memory-tagging-intrinsics

PR #197621, opened 2026-05-14
[Clang][AArch64] Add missing lvalue-to-rvalue conversions for MTE built-ins

igorkudrin:
This patch adds missing lvalue-to-rvalue conversions for the second
argument of `__builtin_arm_irg()` and both arguments of
`__builtin_arm_gmi()`.

Depends on: #196481


efriedma-quic (approved):
LGTM

PR #197624, opened 2026-05-14
Fix vector types w.r.t. qualified element type.

michele-scandale:
This commit changes the way vector types are constructed to ensure that qualifiers on the element type are moved to the vector type itself -- i.e. a canonical vector type has always an unqualified element type.

RFC: https://discourse.llvm.org/t/90783

### DIFF
--- a/clang/include/clang/AST/TypeBase.h
+++ b/clang/include/clang/AST/TypeBase.h
@@ -4167,13 +4167,14 @@ class DependentSizedExtVectorType : public Type, public llvm::FoldingSetNode {
 
   Expr *SizeExpr;
 
-  /// The element type of the array.
+  /// The element type of the vector.
   QualType ElementType;
 
   SourceLocation loc;
+  const ASTContext &Context;
 
-  DependentSizedExtVectorType(QualType ElementType, QualType can,
-                              Expr *SizeExpr, SourceLocation loc);
+  DependentSizedExtVectorType(const ASTContext &Context, QualType ElementType,
+                              QualType can, Expr *SizeExpr, SourceLocation loc);
 
 public:
   Expr *getSizeExpr() const { return SizeExpr; }
@@ -4183,11 +4184,13 @@ public:
   bool isSugared() const { return false; }
   QualType desugar() const { return QualType(this, 0); }
 
+  SplitQualType getSplitUnqualifiedType() const;
+
   static bool classof(const Type *T) {
     return T->getTypeClass() == DependentSizedExtVector;
   }
 
-  void Profile(llvm::FoldingSetNodeID &ID, const ASTContext &Context) {
+  void Profile(llvm::FoldingSetNodeID &ID) {
     Profile(ID, Context, getElementType(), getSizeExpr());
   }
 
@@ -4243,11 +4246,13 @@ protected:
   /// The element type of the vector.
   QualType ElementType;
 
-  VectorType(QualType vecType, unsigned nElements, QualType canonType,
-             VectorKind vecKind);
+  const ASTContext &Context;
 
-  VectorType(TypeClass tc, QualType vecType, unsigned nElements,
+  VectorType(const ASTContext &Context, QualType vecType, unsigned nElements,
              QualType canonType, VectorKind vecKind);
+
+  VectorType(const ASTContext &Context, TypeClass tc, QualType vecType,
+             unsigned nElements, QualType canonType, VectorKind vecKind);
 
 public:
   QualType getElementType() const { return ElementType; }
@@ -4255,6 +4260,8 @@ public:
 
   bool isSugared() const { return false; }
   QualType desugar() const { return QualType(this, 0); }
+
+  SplitQualType getSplitUnqualifiedType() const;
 
   VectorKind getVectorKind() const {
     return VectorKind(VectorTypeBits.VecKind);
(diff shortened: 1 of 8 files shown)


clang/lib/AST/ASTContext.cpp
michele-scandale, line 1: Changes to `getDependentVectorType` and `getDependentSizedExtVectorType` are basically an adaptation of the logic in `getDependentSizedArrayType`.

clang/lib/Sema/SemaTypeTraits.cpp
michele-scandale, line 1: This change shouldn't be needed. `getUnqualifiedArrayType`/`getUnqualifiedType` should correctly handle the removal of qualifiers from the vector element type on the non-canonical representation.

michele-scandale:
I'm proposing changes to `QualType::get{Split,Atomic,}UnqualifiedType` to pass in `const ASTContext &` so that we can handle vector types with qualified element types.
See https://github.com/llvm/llvm-project/pull/198388

michele-scandale:
Ping

michele-scandale:
Based on the feedback from https://github.com/llvm/llvm-project/pull/198388 I'm going rework this to avoid changing the signature of `getUnqualifiedType` and related functions.

michele-scandale:
Ping

michele-scandale:
Ping

clang/include/clang/AST/TypeBase.h
```diff
@@ -4165,13 +4165,14 @@ class DependentSizedExtVectorType : public Type, public llvm::FoldingSetNode {
 
   Expr *SizeExpr;
 
-  /// The element type of the array.
+  /// The element type of the vector.
   QualType ElementType;
 
   SourceLocation loc;
+  const ASTContext &Context;
```
erichkeane, line 4174: I don't think the type having a `Context` is appropriate? I thought we mentioned on the other review that this should store the element type with proper qualifiers instead?

michele-scandale: As mentioned in https://github.com/llvm/llvm-project/pull/198388#issuecomment-4633152864:

> If it is acceptable to add members to `VectorType` (and other vector related type classes) then I'd rather just have a `const ASTContext &` similarly to what has been done for `TypeOfExprType`.

Precomputing the unqualified type seems quite odd, it's an information that might be never needed. And in terms of memory increase I would still be adding an extra `sizeof(void*)` in both cases. Also `TypeOfExprType` and `TypeOfType` are already doing that to implement `desugar`. I really don't see a difference in what I'm trying to do here.

libcxx/include/__algorithm/find.h
```diff
@@ -69,11 +70,11 @@ _LIBCPP_CONSTEXPR_SINCE_CXX14 _Tp* __find_vectorized(_Tp* __first, _Tp* __last,
   if (!__libcpp_is_constant_evaluated()) {
     constexpr size_t __unroll_count = 4;
     constexpr size_t __vec_size     = __native_vector_size<_Tp>;
-    using __vec                     = __simd_vector<_Tp, __vec_size>;
+    using __vec                     = __simd_vector<typename remove_cv<_Tp>::type, __vec_size>;
```
frederick-vs-ja:
```suggestion
    using __vec                     = __simd_vector<__remove_cv_t<_Tp>, __vec_size>;
```
Ditto below.

philnik777: Given that we seemingly have to apply it everywhere I wonder whether it wouldn't be better to move it into the alias instead.

michele-scandale: If you have the following:
```
template <class _ArithmeticT, size_t _Np>
using __simd_vector __attribute__((__ext_vector_type__(_Np))) _LIBCPP_NODEBUG = __remove_cv_t<_ArithmeticT>;
```

then the following cannot work

```
template <class _Tp, size_t _Np>
inline constexpr size_t __simd_vector_size_v<__simd_vector<_Tp, _Np>> = _Np;
```

because the presence of `__remove_cv_t` breaks the template type matching. This affects also other functions.

Is it fine to have a `__simd_vector_impl` and implement the vector traits/helpers using `__simd_vector_impl` for template parameter matching purposes, and then expose `__simd_vector<T, N> = __simd_vector_impl<__remove_cv_t<T>, N>`?

michele-scandale:
Ping

michele-scandale:
Ping

cc @erichkeane, @frederick-vs-ja, @philnik777

philnik777:
I'd like to see a conclusion to the RFC before spending more time reviewing this. (CC @AaronBallman)

AaronBallman:
> I'd like to see a conclusion to the RFC before spending more time reviewing this. (CC @AaronBallman)

It's on the Clang Area Team's schedule; I'll try to ensure we get to it at the next meeting (Thur Jul 20).

clang/test/Sema/types.c
```diff
@@ -75,6 +75,18 @@ typedef int __attribute__((ext_vector_type(0x100000000))) e2;      // expected-e
...
+typedef int __attribute__((vector_size(16))) vec4_int;
+typedef cint __attribute__((vector_size(16))) vec4_cint;
+_Static_assert(__builtin_types_compatible_p(vec4_int, const vec4_int), "'vec4_int' and 'const vec4_int' are not compatible");
+_Static_assert(__builtin_types_compatible_p(vec4_int, volatile vec4_int), "'vec4_int' and 'const vec4_int' are not compatible");
+_Static_assert(__builtin_types_compatible_p(const vec4_int, vec4_cint), "'const vec4_int' and 'vec4_cint' are not compatible");
+typedef int __attribute__((ext_vector_type(4))) ext_vec4_int;
+typedef cint __attribute__((ext_vector_type(4))) ext_vec4_cint;
+_Static_assert(__builtin_types_compatible_p(ext_vec4_int, const ext_vec4_int), "'ext_vec4_int' and 'const ext_vec4_int' are not compatible");
+_Static_assert(__builtin_types_compatible_p(ext_vec4_int, volatile ext_vec4_int), "'ext_vec4_int' and 'const ext_vec4_int' are not compatible");
+_Static_assert(__builtin_types_compatible_p(const ext_vec4_int, ext_vec4_cint), "'const ext_vec4_int' and 'ext_vec4_cint' are not compatible'");
```
efriedma-quic, line 88: I don't think this is testing what you want to test... maybe use something more like:

```
static_assert(__builtin_types_compatible_p(vec4_cint*, const vec4_int*), "");
```

PR #197628, opened 2026-05-14
[AMDGPU] Set SGPR spill stack IDs when creating spill slots

### LINKED ISSUE #195092: [AMDGPU] llc -O1 assert in buildVGPRSpillLoadStore on gfx90a
To reproduce the problem, extract the test.ll from the attached file and then use the following command.

`llc -O1 test.ll -mtriple=amdgcn-amd-amdhsa -mcpu=gfx90a -o /dev/null`

You will see the following assertion failing

`SIRegisterInfo.cpp:2023: void llvm::SIRegisterInfo::buildVGPRSpillLoadStore(llvm::SGPRSpillBuilder&, int, int, bool, bool) const: Assertion `FrameInfo.getStackID(Index) != TargetStackID::SGPRSpill' failed.`

Please note that problem only happens at -O1 and for gfx90a. The testcase was reduced from a fortran application.

The attached file also contains a reduced fortran testcase. Run the following command to see the same problem with it.
`flang -O1 -fopenmp --offload-arch=gfx90a -o test min.F90 -fdefault-integer-8`

[testcase_for_reproting.tar.gz](https://github.com/user-attachments/files/27246541/testcase_for_reproting.tar.gz)

woruyu:
### Summary
This PR resolves https://github.com/llvm/llvm-project/issues/195092. 

Teach register classes to describe the StackID to use for their spill slots, and use it when creating spill frame objects. This lets AMDGPU SGPR spill slots be marked as TargetStackID::SGPRSpill at construction time instead of patching the frame object later while emitting spill instructions.

Also handle SGPR callee-saved spill slots created directly by SILowerSGPRSpills, which bypass the generic PEI spill slot creation path. Without this, late SGPR scavenging can see an SGPR spill pseudo using a default-stack frame index and trip the SGPRSpill assertion.


llvm/lib/Target/AMDGPU/SIMachineFunctionInfo.h
```diff
@@ -609,6 +611,11 @@ class SIMachineFunctionInfo final : public AMDGPUMachineFunctionInfo,
   // frame, so save it here and add it to the RegScavenger later.
   std::optional<int> ScavengeFI;
 
+  // Emergency stack slot used to save the temporary VGPR while lowering SGPR
+  // spills. This must stay separate from ScavengeFI, which RegScavenger may use
+  // for SGPR spills and mark as an SGPRSpill stack object.
```
arsenm: The frame index stack ID really ought to be set at construction time, and not during the spill handling

llvm/lib/Target/AMDGPU/SIInstrInfo.cpp
```diff
@@ -1917,7 +1918,8 @@ void SIInstrInfo::loadRegFromStackSlot(MachineBasicBlock &MBB,
       MRI.constrainRegClass(DestReg, &AMDGPU::SReg_32_XM0_XEXECRegClass);
     }
 
-    if (RI.spillSGPRToVGPR())
+    std::optional<int> ScavengeFI = MFI->getOptionalScavengeFI();
+    if (RI.spillSGPRToVGPR() && (!ScavengeFI || *ScavengeFI != FrameIndex))
```
arsenm: This shouldn't require some sort of fallback process. There should just be the correct index to use

llvm/test/CodeGen/AMDGPU/sgpr-scavenge-fi-stack-id.ll
```diff
@@ -0,0 +1,301 @@
...
+  %10 = select i1 %6, double 0.000000e+00, double %9
+  %11 = insertelement <4 x double> zeroinitializer, double 0.000000e+00, i32 0
+  br label %12
+
+12:
+  %13 = phi i64 [ 0, %.preheader ], [ 1, %12 ]
+  %14 = trunc i64 %13 to i32
+  %15 = select i1 %2, double 0.000000e+00, double %10
+  %16 = fmul double 0x7FF0000000000000, %15
+  %17 = fmul double %16, 0x40417E50A9DC6553
```
arsenm: Use named values in tests

llvm/lib/Target/AMDGPU/SIMachineFunctionInfo.h
```diff
@@ -296,6 +296,7 @@ struct SIMachineFunctionInfo final : public yaml::MachineFunctionInfo {
 
   SIMode Mode;
   std::optional<FrameIndex> ScavengeFI;
+  std::optional<FrameIndex> VGPRScavengeFI;
```
arsenm: The scavenging frame index is always a VGPR index as-is. I don't understand what the second one is for (I also do not know why it's using the size queried for SGPR_32. This is the same 4, I don't know why it's written weirdly)

arsenm (commented):
PrologEpilogInserter should not be processing SGPR spills as spills to VGPR, and they need to go through the memory VGPR path. 

I see 2 problems here:

1. This particular instance is failing on a frame index used by s_lshl3_add_u32. There is a missed optimization here. This can have special case handling in eliminateFrameIndex like we do for add. This would avoid the need for a spill in this instance. 
2. The current stack ID management is hacky. Code just creates the stack ID, and then the ID is lazily assigned when accessed. It would make much more sense if the stack ID was set at frame index creation time. Then the spill lowering logic could switch between the to-VGPR and to-memory case based on the stack ID. You seem to be trying to figure out based on other context how the spill needs to be handled

arsenm:
To get the stack ID at construction, we'd want to add another field to RegisterClass. We already have getSpillSize and getSpillAlign; we'd need a similar getSpillStackID

llvm/lib/Target/AMDGPU/SIInstrInfo.cpp
woruyu: Done! Delete.

llvm/lib/Target/AMDGPU/SIMachineFunctionInfo.h
woruyu: Understand! Delete VGPRScavengeFI.

llvm/lib/Target/AMDGPU/SIMachineFunctionInfo.h
woruyu: Done! Add SpillStackID in target.td to create stack with related registerclass StackID

llvm/test/CodeGen/AMDGPU/sgpr-scavenge-fi-stack-id.ll
woruyu: Done!

woruyu:
> To get the stack ID at construction, we'd want to add another field to RegisterClass. We already have getSpillSize and getSpillAlign; we'd need a similar getSpillStackID

Understand! I have added getSpillStackID to TargetRegisterInfo.

llvm/lib/Target/AMDGPU/SIInstrInfo.cpp
```diff
@@ -1949,7 +1948,8 @@ void SIInstrInfo::loadRegFromStackSlot(MachineBasicBlock &MBB,
       FrameInfo.getObjectAlign(FrameIndex));
 
   if (RI.isSGPRClass(RC)) {
-    MFI->setHasSpilledSGPRs();
+    if (FrameInfo.getStackID(FrameIndex) == TargetStackID::SGPRSpill)
+      MFI->setHasSpilledSGPRs();
```
arsenm, line 1952: In the future we can probably get rid of this MFI field, and just check directly against the frame object stack IDs

llvm/test/CodeGen/AMDGPU/sgpr-scavenge-fi-stack-id.ll
```diff
@@ -0,0 +1,69 @@
+; RUN: llc -O1 -mtriple=amdgcn-amd-amdhsa -mcpu=gfx90a -verify-machineinstrs -o - %s > /dev/null
+
+; The RegScavenger emergency FI can be used for late SGPR spills during PEI.
+; It must remain a normal stack object because these spills are lowered to
+; memory directly, after SILowerSGPRSpills has already run.
+
+target datalayout = "e-m:e-p:64:64-p1:64:64-p2:32:32-p3:32:32-p4:64:64-p5:32:32-p6:32:32-p7:160:256:256:32-p8:128:128:128:48-p9:192:256:256:32-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-v2048-n32:64-S32-A5-G1-ni:7:8:9"
+target triple = "amdgcn-amd-amdhsa"
+
```
arsenm:
```suggestion
```

llvm/utils/TableGen/Common/CodeGenRegisters.cpp
```diff
@@ -730,6 +730,10 @@ CodeGenRegisterClass::CodeGenRegisterClass(CodeGenRegBank &RegBank,
   int CopyCostParsed = R->getValueAsInt("CopyCost");
   Allocatable = R->getValueAsBit("isAllocatable");
   AltOrderSelect = R->getValueAsString("AltOrderSelect");
+  int SpillStackIDParsed = R->getValueAsInt("SpillStackID");
+  if (!isUInt<8>(SpillStackIDParsed))
+    PrintFatalError(R->getLoc(), "SpillStackID out of range [0,255]");
```
arsenm, line 735: Can you add a tablegen test for this error?

arsenm (commented):
Mostly lgtm

llvm/utils/TableGen/Common/CodeGenRegisters.cpp
woruyu: Done!

llvm/test/CodeGen/AMDGPU/sgpr-scavenge-fi-stack-id.ll
woruyu: Done!

llvm/test/CodeGen/AMDGPU/sgpr-scavenge-fi-stack-id.ll
```diff
@@ -0,0 +1,62 @@
+; RUN: llc -O1 -mtriple=amdgcn-amd-amdhsa -mcpu=gfx90a -verify-machineinstrs -o - %s > /dev/null
+
+define void @sgpr_scavenge_fi_stack_id(double %arg, i1 %outer.cond, i1 %exit.cond, i1 %middle.cond, i1 %inner.cond, i1 %loop.break.cond, ptr addrspace(5) %out, i1 %select.cond) {
```
shiltian: Can you check https://llvm.org/docs/AMDGPU/DeveloperGuideline.html for test guideline?

llvm/lib/Target/AMDGPU/SIInstrInfo.cpp
cdevadas: This field can be safely removed. Where do we really use this MFI field?

arsenm: I think this was used to detect if we will need to introduce CSR spills

llvm/test/CodeGen/AMDGPU/sgpr-scavenge-fi-stack-id.ll
```diff
@@ -0,0 +1,62 @@
+; RUN: llc -O1 -mtriple=amdgcn-amd-amdhsa -mcpu=gfx90a -verify-machineinstrs -o - %s > /dev/null
```
arsenm:
```suggestion
; RUN: llc -O1 -mtriple=amdgcn-amd-amdhsa -mcpu=gfx90a < %s | FileCheck %s
```

Should generate checks. Also drop -O1 if possible

woruyu: If remove -O1, the testcase will pass instead of crash in origin main branch.

llvm/test/CodeGen/AMDGPU/sgpr-scavenge-fi-stack-id.ll
woruyu: understand! I have deleted `-o -` , removed unused attribute and minimize testcase. However, this case is very sensitive to SGPR register pressure and is difficult to simplify.

llvm/test/CodeGen/AMDGPU/sgpr-scavenge-fi-stack-id.ll
```diff
@@ -0,0 +1,294 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
+; RUN: llc -O1 -mtriple=amdgcn-amd-amdhsa -mcpu=gfx90a -verify-machineinstrs < %s | FileCheck %s
```
arsenm:
```suggestion
; RUN: llc -O1 -mtriple=amdgcn-amd-amdhsa -mcpu=gfx90a < %s | FileCheck %s
```

arsenm (approved):

PR #194180, opened 2026-04-25
[clang] Add -Wbool-integral-comparison

mechakotik:
Warn on suspicious comparisons of bool and non-constant integral or enumeration expressions, like this:

```cpp
bool b = ...;
int i = ...;
bool res = (b == i);
```

The bool operand is converted to an integral value that can only be 0 or 1, so comparing it with an arbitrary number is often unintentional.

This warning does not apply to integral constant expressions. Comparisons against 0 and 1 are common intentional patterns, while comparisons against other constants are better handled by existing tautological comparison warnings.

It also does not apply to dependent template comparisons or integral expressions known to have a boolean value, such as unsigned one-bit bit fields.

This warning is enabled by `-Wextra`.

See also [[NFC] Silence -Wbool-integral-comparison warnings across LLVM](195246)

PR #197633, opened 2026-05-14
[compiler-rt] Fix StackDepot benchmark thread barrier

mechakotik:
Use Param.Threads (number of worker threads) as barrier threshold instead of Param.UniqueThreads (boolean that controls input generation).

This also silences [-Wbool-integral-comparison](https://github.com/llvm/llvm-project/pull/194180) warning I'm working on.

### DIFF
--- a/compiler-rt/lib/sanitizer_common/tests/sanitizer_stackdepot_test.cpp
+++ b/compiler-rt/lib/sanitizer_common/tests/sanitizer_stackdepot_test.cpp
@@ -174,11 +174,11 @@ class StackDepotBenchmark
 //   '--gtest_filter=*Benchmark*'
 TEST_P(StackDepotBenchmark, DISABLED_Benchmark) {
   auto Param = GetParam();
-  std::atomic<unsigned int> here = {};
+  std::atomic<int> here = {};
 
   auto thread = [&](int idx) {
     here++;
-    while (here < Param.UniqueThreads) std::this_thread::yield();
+    while (here < Param.Threads) std::this_thread::yield();
 
     std::vector<uptr> frames(64);
     for (int r = 0; r < Param.RepeatPerThread; ++r) {


compiler-rt/lib/sanitizer_common/tests/sanitizer_stackdepot_test.cpp
```diff
@@ -174,11 +174,11 @@ class StackDepotBenchmark
 //   '--gtest_filter=*Benchmark*'
 TEST_P(StackDepotBenchmark, DISABLED_Benchmark) {
   auto Param = GetParam();
-  std::atomic<unsigned int> here = {};
+  std::atomic<int> here = {};
```
davidmrdavid, line 177: I presume the change from `unsigned int` to just `int` is to avoid signed-vs-unsigned warnings in `here < Param.Threads` later on?

compiler-rt/lib/sanitizer_common/tests/sanitizer_stackdepot_test.cpp
```diff
@@ -174,11 +174,11 @@ class StackDepotBenchmark
...
 
   auto thread = [&](int idx) {
     here++;
-    while (here < Param.UniqueThreads) std::this_thread::yield();
```
davidmrdavid, line 181: pre-existing: I'd be a fan of renaming `UniqueThreads` to `HasUniqueThreads` to better reflect that the type is `bool`

davidmrdavid (approved):
LGTM, just 2 non-blocking comments

compiler-rt/lib/sanitizer_common/tests/sanitizer_stackdepot_test.cpp
mechakotik: Yes, this comparison is the only place `here` is used, so it's cleaner for it to have the same type as right side

vitalybuka (approved):

PR #187348, opened 2026-03-18
[flang][DRAFT] Fix sequence association for PARAMETER array elements

eugeneepshteyn:
[Converting this to draft to keep semantic changes alive. The final fix will be different]

Fortran 2023 15.5.2.5p14 allows passing a scalar array element a(i) to an
explicit-shape dummy array via sequence association: the callee may access
elements a(i), a(i+1), ... through the dummy argument. This was working
for regular arrays, but was broken for constant (PARAMETER) arrays.

The fix involved teaching semantics not to constant-fold elements of
constant arrays and changing lowering not to create temps for
sequence association cases.

PR #197636, opened 2026-05-14
  [flang]  Avoid unnecessary temporaries for PARAMETER actual arguments passed to INTENT(IN) dummies

buggfg:
**Description**

When the dummy argument is declared `INTENT(IN)`, the procedure is not permitted by the Fortran standard (F2018, 8.5.10) to define the dummy argument. In such cases, creating a temporary and performing copy-out may be unnecessary and can introduce avoidable overhead, particularly for large string constants or array parameters.

**Validation & performance**

- No compile or semantic errors observed on SPEC2006 and SPEC2017 with the new feature enabled for validation.
- With this patch: 
  - SPEC2006 465.tonto on RISC-V (XuanTieC920): **5% speedup**
  - SPEC2006 465.tonto on RISC-V (KMHv3 RTL): **3% speedup**


eugeneepshteyn:
@buggfg , thank you for the change. It's nice to see the speedups that we could get by not doing unnecessary copies.

Please see discussion about crashes and read-only memory in this PR: https://github.com/llvm/llvm-project/pull/187348 

(To clarify: it would be good to avoid extra copy-in/out of constants, we just want to do that in a way that covers any relevant corner cases beyond SPEC tests.)

buggfg:
> Please see discussion about crashes and read-only memory in this PR: #187348
> 
> (To clarify: it would be good to avoid extra copy-in/out of constants, we just want to do that in a way that covers any relevant corner cases beyond SPEC tests.)

yeah,  compiler optimizations should target general use cases rather than special-casing a specific benchmark.

In the [array-call-chain.f90](https://godbolt.org/z/havzWchca) and [string-call-chain.f90](https://godbolt.org/z/9cG9YPnPG) test cases (PR #187348):

- Flang creates a temporary copy, so the program still runs even if the user violates the `INTENT(IN)` contract.
- GCC 15 does not create a temporary copy, and the program triggers a segmentation fault at runtime when attempting to modify read-only data.

I think users apply `INTENT(IN)` to explicitly express that an argument must not be modified, similar to `const` in C/C++. This is especially important in large projects to prevent accidental modification of arguments.

With this patch, the compiler will no longer automatically create temporary copies for such cases. If the user violates the `INTENT(IN)` contract and attempts to modify the data, the issue will surface immediately at runtime (for example, via a segmentation fault).

The main idea is to give the choice back to the user:

- either adjust/remove `INTENT(IN)` if modification is intended,
- or follow the `INTENT(IN)` contract and avoid modifying the argument.

PR #197637, opened 2026-05-14
[SmallPtrSet] Drop tombstones in large mode

MaskRay:
SmallPtrSet uses quadratic probing with tombstone deletion in large
mode. Tombstones occupy a third bucket state and hurts lookup.

Switch to linear probing with deletion implemented using Knuth TAOCP 6.4
Algorithm R.  `erase` opens a hole at the removed slot, walks forward
sliding each following entry whose probe path crosses the hole back
into it (the hole moves with each slide), and stops at the next empty
slot.  The scan stops at the next empty bucket, which is guaranteed to
exist.

`remove_if` clears matches in a single pass then calls `Grow` at the
current size to restore the linear-probe invariant, O(N) total.
(Per-match Algorithm R erase would be O(N * cluster).)

My DenseMap experiments suggest that Robin Hood Hashing and Abseil Swiss
Table family (not good at small keys) are actually worse than the
baseline.

Mirrors the small-mode change in 42c3edb4819f, which dropped tombstones
from the inline array and likewise had `erase` invalidate iterators.
`erase` now physically relocates following entries, so it invalidates
iterators and references in large mode too.

Linear probing is vulnerable to primary clustering. Therefore, we lower
the max load factor to 2/3, slightly faster than the current 3/4.


MaskRay:
https://llvm-compile-time-tracker.com/compare.php?from=f996980ff486a0cc03b84f87bbb3554ea99be1e2&to=106e2189ab0c56a6e1e259e4603d8b53f7012fd5&stat=instructions%3Au

stage1-O0-g:
geomean	17469M	17460M (-0.05%)

stage2-O3:
geomean	51391M	51271M (-0.23%)

clang build:
[instructions:u](https://llvm-compile-time-tracker.com/compare_clang.php?from=f996980ff486a0cc03b84f87bbb3554ea99be1e2&to=106e2189ab0c56a6e1e259e4603d8b53f7012fd5&stat=instructions%3Au)	21430106M	21406732M (-0.11%)

llvm/lib/Support/SmallPtrSet.cpp
```diff
@@ -39,73 +39,64 @@ void SmallPtrSetImplBase::shrink_and_clear() {
...
-    // It's a hash collision or a tombstone. Reprobe.
-    Bucket = (Bucket + ProbeAmt++) & (ArraySize-1);
+void SmallPtrSetImplBase::eraseFromBucket(const void **Bucket) {
+  // Knuth TAOCP 6.4 Algorithm R: walk forward sliding each following entry
+  // whose probe path crosses the hole.
+  unsigned Mask = CurArraySize - 1;
+  unsigned I = Bucket - CurArray;
+  unsigned J = I;
+  const void *Empty = getEmptyMarker();
+  while ((J = (J + 1) & Mask), CurArray[J] != Empty) {
```
nikic, line 92: Can we avoid this assignment in the while condition? You can use `for (; CurArray[J] != Empty; J = (J + 1) & Mask)` plus initialize `J = (I + 1) & Mask`?

llvm/include/llvm/ADT/SmallPtrSet.h
```diff
@@ -436,18 +434,24 @@ template <typename PtrType> class SmallPtrSetImpl : public SmallPtrSetImplBase {
...
-        ++NumTombstones;
+        Bucket = getEmptyMarker();
         --NumEntries;
-        incrementEpoch();
         Removed = true;
       }
     }
+    if (Removed) {
+      incrementEpoch();
+      Grow(CurArraySize);
```
nikic, line 455: So if remove_if() removes anything, we'll now always force a new allocation?

nikic (commented):
This generally looks reasonable to me.

I think it would be good to do some local wall-time measurements to confirm this is indeed better (or at least neutral). For these kinds of data structure changes, `instructions:u` probably does not paint the full picture.

> My DenseMap experiments suggest that Robin Hood Hashing and Abseil Swiss
Table family (not good at small keys) are actually worse than the
baseline.

Is this in terms of instructions or wall-time? Especially for Swiss Table I'd expect that these may not be well-correlated.

llvm/lib/Support/SmallPtrSet.cpp
```diff
@@ -39,73 +39,64 @@ void SmallPtrSetImplBase::shrink_and_clear() {
...
+    auto Ideal = DenseMapInfo<void *>::getHashValue(CurArray[J]);
+    if (((I - Ideal) & Mask) < ((J - Ideal) & Mask)) {
+      CurArray[I] = CurArray[J];
+      I = J;
+    }
   }
+  CurArray[I] = Empty;
 }
```
MaxGraey, line 100: It seems we can drop one mask wrap:
```py
((I - Ideal) & Mask) < ((J - Ideal) & Mask)
```
can be represented as:
```py
dist(Ideal -> hole) < dist(Ideal -> current) 
```
Or in another representaion: `Entry` if `Ideal` `<=probe` `Hole` `<probe` `Scan`,
where `<=probe` is `Ideal, Ideal + 1, Ideal + 2, ... modulo N`
and the same with Gap:
```py
dist(Ideal -> Scan) = (dist(Ideal -> Hole) + Gap) % N
```
But since we know `Gap > 0` the condition becomes:
```py
Back < (Back + Gap) % N
```
or similified equivalent:
```py
Back + Gap < N
```
So we can avoid the second & Mask:
```py
Back <= Mask - Gap
```
and writ this something like:
```cpp
void SmallPtrSetImplBase::eraseFromBucket(const void **Bucket) {
  const unsigned Mask = CurArraySize - 1;
  const void *Empty = getEmptyMarker();
  unsigned Hole = Bucket - CurArray;

  for (unsigned Scan = (Hole + 1) & Mask, Gap = 1; ; Scan = (Scan + 1) & Mask, ++Gap) {
    const void *Entry = CurArray[Scan];
    if (Entry == Empty)
      break;

    const unsigned Back = 
      (Hole - DenseMapInfo<void *>::getHashValue(Entry)) & Mask;

    if (Back <= Mask - Gap) {
      CurArray[Hole] = Entry;
      Hole = Scan;
      Gap = 0;
    }
  }

  CurArray[Hole] = Empty;
}
```

Although I might be wrong, you should test this in practice

MaskRay:
> This generally looks reasonable to me.
> 
> I think it would be good to do some local wall-time measurements to confirm this is indeed better (or at least neutral). For these kinds of data structure changes, `instructions:u` probably does not paint the full picture.
> 
> > My DenseMap experiments suggest that Robin Hood Hashing and Abseil Swiss
> > Table family (not good at small keys) are actually worse than the
> > baseline.
> 
> Is this in terms of instructions or wall-time? Especially for Swiss Table I'd expect that these may not be well-correlated.

Both instructions:u and wall-time move in the same direction here. I extracted DenseMap / DenseMapInfo into a standalone workbench and simulated two workloads against it; one is modeled on clang `Decl *` lookups, which `perf` identifies as one of clang's heaviest DenseMap consumers. Both workloads use small `pair<key, value>` buckets. The relative ordering matches what shows up on https://llvm-compile-time-tracker.com/.

```
  ┌──────────┬──────────────────────────┬──────────────────────────┬──────────────────────────┐
  │ workload │   tombstone (original)   │ linear-probe + backshift │        Boost FOA         │
  ├──────────┼──────────────────────────┼──────────────────────────┼──────────────────────────┤
  │ A        │ 1,295M instr / 257.06 ms │ 1,283M instr / 256.06 ms │ 1,345M instr / 257.16 ms │
  ├──────────┼──────────────────────────┼──────────────────────────┼──────────────────────────┤
  │ C        │  147.8M instr / 51.36 ms │  138.9M instr / 50.15 ms │  170.9M instr / 50.34 ms │
  └──────────┴──────────────────────────┴──────────────────────────┴──────────────────────────┘
```

llvm/lib/Support/SmallPtrSet.cpp
MaskRay: Thanks for suggesting the alternative. It's correct but there is no difference
https://llvm-compile-time-tracker.com/compare.php?from=6b1a6d3581202cfbce8c3bb030dbf06f1f2423fb&to=cfb3002ff020c198ea2033abc800d4195bc6fe25&stat=instructions:u

Measured locally
```
  ▎ |           |     C instr | C wall (ms) |
  ▎ |-----------|-------------|-------------|
  ▎ | original  | 138,911,223 |       48.30 |
  ▎ | suggested | 139,018,238 |       48.20 |
  ▎
  ▎ +0.08 % instructions / ~-0.2 % wall — within noise. I'll keep the original form for readability (it's a direct expression of the Knuth
  ▎ Algorithm R invariant) unless you'd like me to take the suggestion anyway.
```

```
  ▎ Compared the generated code (llvm-objdump -d, clang++ -O2, the eraseFromFilledBucket inlined into the C workload):
  ▎
  ▎ Original: 2× sub + 2× and + cmp, both (I-H)&Mask and (J-H)&Mask chained through the entry's hash.
  ▎ Suggested: 2× sub + 1× and + cmp, plus 1× inc and 1× xor for the Gap counter. The Mask - Gap operand is on an independent dependency chain (no hash dependency), shortening the critical path to the compare.
  ▎
  ▎ Same total op count per iteration; one fewer AND, but an extra INC. Measurements bear that out: +0.08 % instructions:u / -0.21 % wall on the erase-heavy workload — within noise. I think keeping the original form is preferable for readability; happy to take the change if you feel
  ▎ strongly about the dep-chain win.
```

When the hash function is not inline, `Gap` becomes a new live value across the call.

 Sticking with my current implementation.

llvm/lib/Support/SmallPtrSet.cpp
MaskRay: The current code deduplicates `(J = (I + 1) & Mask)` and `(J = (J + 1) & Mask)`

llvm/lib/Support/SmallPtrSet.cpp
MaxGraey: Yeah, Krut's algorithms are really hard to beat

MaxGraey: Slightly off-topic regarding `wall-time` on https://llvm-compile-time-tracker.com. Have you considered using hyperfine (https://github.com/sharkdp/hyperfine)? It would significantly reduce noise and provide more reliable statistics. It also has many useful options, such as configuring how long the warmup stage should run and how it should be performed.

MaskRay: My local workbench uses this to pin a single core and disable address space randomization to get more stable results.
```
numactl -C 20 setarch x86_64 -R perf stat -x, -e instructions:u,cycles,L1-dcache-load-misses,task-clock
```

> Have you considered using hyperfine ([sharkdp/hyperfine](https://github.com/sharkdp/hyperfine))?

This is a question for nikic, but I think it's not doable since running all of `stage1-O3	stage1-ReleaseThinLTO	stage1-ReleaseLTO-g	stage1-O0-g	stage1-aarch64-O3	stage1-aarch64-O0-g	stage2-O3	stage2-O0-g	stage2-clang` already eats up way too many resources on a machine...

---

BTW, TAOCP uses a decreasing iterator and makes no mention of `I = J` after sliding the current entry to the hole.

https://github.com/rigtorp/HashMap/blob/master/include/rigtorp/HashMap.h#L259 also employs TAOCP 6.4 Algorithm R. Its benchmark shows it beats absl::flat_hash_map in its small key benchmarks.

---

For DenseMap/SmallPtrSet, the code size from the template instantiations is critical. I believe swiss table family algorithms lose to simple linear probing...

llvm/include/llvm/ADT/SmallPtrSet.h
MaskRay: Yes i think this is small cost to pay since this O(n) operation is used very rarely

llvm/include/llvm/ADT/SmallPtrSet.h
```diff
@@ -436,18 +434,24 @@ template <typename PtrType> class SmallPtrSetImpl : public SmallPtrSetImplBase {
...
+    // for repeated per-match Algorithm R erases.
     for (const void *&Bucket : buckets()) {
-      if (Bucket == getTombstoneMarker() || Bucket == getEmptyMarker())
+      if (Bucket == getEmptyMarker())
         continue;
       PtrType Ptr = PtrTraits::getFromVoidPointer(const_cast<void *>(Bucket));
       if (P(Ptr)) {
-        Bucket = getTombstoneMarker();
-        ++NumTombstones;
+        Bucket = getEmptyMarker();
```
aengelke, line 448: The comment says: "It is safe to read the set inside the predicate function." -- But this seems no longer to be the case? I guess that means finding and validating all in-tree uses of remove_if..

llvm/include/llvm/ADT/SmallPtrSet.h
aengelke: Mention in doc comment?

aengelke (commented):
LGTM otherwise.

llvm/include/llvm/ADT/SmallPtrSet.h
MaskRay: Updated the comment.

I find 3 references with my language server. The uses do not require changes.
```
llvm/include/llvm/IR/Analysis.h:227:18:

llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp:6576:21:
```

aengelke: There seem to be a few more, but all seem fine:

- InterleavedAccessInfo::invalidateGroupsRequiringScalarEpilogue
- AArch64MachineFunctionInfo
- SPIRVEmitIntrinsics::postprocessTypes
- set_intersect

aengelke (approved):
LGTM

ronlieb:
seeing same issues downstream merges from this PR

mikaelholmen:
I see some MLIR lit test flakiness with this patch.
E.g.
```
build-all/bin/mlir-translate -mlir-to-llvmir -split-input-file /repo/llvm/mlir/test/Target/LLVMIR/openmp-llvm.mlir
```
now randomly hits
```
mlir-translate: /repo/llvm/llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp:926: void llvm::OpenMPIRBuilder::finalize(Function *): Assertion `Extractor->isEligible() && "Expected OpenMP outlining to be possible!"' failed.
```
I have no idea if this patch is really to blame or if users of SmallPtrSet do things they shouldn't, but either way, with this patch things randomly start to fail.

I've seen the assert above and CHECK failures for at least
```
mlir/test/Target/LLVMIR/openmp-cli-fuse01.mlir
mlir/test/Target/LLVMIR/openmp-cli-fuse02.mlir
mlir/test/Target/LLVMIR/openmp-cli-tile02.mlir
mlir/test/Target/LLVMIR/openmp-llvm.mlir
mlir/test/Target/LLVMIR/openmp-taskloop-collapse.mlir
```
I suppose https://github.com/llvm/llvm-project/pull/198690 tries to do something about it.

slinder1:
I know we don't consider downstream users when making breaking changes outside of the IRs and some of the C API, but this change to the semantics of `remove_if` seems like it should at least get a call out in some release note or something.

Since this broke in-tree code I think we should revert now, confirm any now-broken uses are updated (seems like there is just the single use in the mlir omp code?), and then re-land. That the bug is non-deterministic means it is likely to eat a lot of developer time. I will revert it in ~1 hour from this comment unless I hear otherwise from someone.

aengelke:
I'm +/-0 on reverting. Given that the single problematic use in-tree already has a fix with #198690, we might just as well fix forward? If that needs more discussion, we should obviously revert. +1 to release note, remove_if is used rarely, but there's no assert to catch this, so clearly documenting this change is a good idea.

PR #197638, opened 2026-05-14
[AMDGPU] Less-constrained scheduling for ASYNCMARK

lijinpei-amd:
ASYNCMARK previously acted as a global memory barrier in the scheduler because of its hasSideEffects bit, pinning all surrounding memory access (read/write, sync/async, vmem/lds, etc) in place.

This commit improves scheduling freedom for ASYNCMARK by

1. Stop reporting ASYNCMARK as a global memory object. WAIT_ASYNCMARK is still a global memory object so ordinary loads/stores remain anchored across waits.

2. Instead, express the strictly necessary ordering with an artificial-edge DAG mutation:

   2.1. Relative order of ASYNCMARKs and WAIT_ASYNCMARKs is kept. 2.2. Each async load/store gets an edge to the next ASYNCMARK and an edge from the previous ASYNCMARK.

   ASYNCMARKs are otherwise exempt from dependencies with non-async load/store. WAIT_ASYNCMARK anchoring is left to the natural memory dep that WAIT_ASYNCMARK provides as a global memory object.

The mutation is registered in every pre-RA scheduler stage and in the post-RA scheduler, and handles BUNDLEs created by SIPostRABundler / SIInsertHardClauses.

The new mutation is a must-have for correctness: once ASYNCMARK is no longer a global memory object, the scheduler is free to reorder it across async loads/stores and WAIT_ASYNCMARK in ways that break program semantics. Later scheduler stages must remember to register this mutation.

### DIFF
--- a/llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
+++ b/llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
@@ -3562,7 +3562,7 @@ bool AMDGPUInstructionSelector::selectG_INSERT_VECTOR_ELT(
   return true;
 }
 
-static bool isAsyncLDSDMA(Intrinsic::ID Intr) {
+static bool isPreGFX12Async(Intrinsic::ID Intr) {
   switch (Intr) {
   case Intrinsic::amdgcn_raw_buffer_load_async_lds:
   case Intrinsic::amdgcn_raw_ptr_buffer_load_async_lds:
@@ -3637,6 +3637,12 @@ bool AMDGPUInstructionSelector::selectBufferLoadLds(MachineInstr &MI) const {
     break;
   }
 
+  if (isPreGFX12Async(IntrinsicID)) {
+    int AsyncOpc = AMDGPU::getPreGFX12AsyncOp(Opc);
+    assert(AsyncOpc != -1 && "PreGFX12Async sibling missing");
+    Opc = AsyncOpc;
+  }
+
   MachineBasicBlock *MBB = MI.getParent();
   const DebugLoc &DL = MI.getDebugLoc();
   BuildMI(*MBB, &MI, DL, TII.get(AMDGPU::COPY), AMDGPU::M0)
@@ -3670,7 +3676,6 @@ bool AMDGPUInstructionSelector::selectBufferLoadLds(MachineInstr &MI) const {
       Aux & (IsGFX12Plus ? AMDGPU::CPol::SWZ : AMDGPU::CPol::SWZ_pregfx12)
           ? 1
           : 0); // swz
-  MIB.addImm(isAsyncLDSDMA(IntrinsicID));
 
   MachineMemOperand *LoadMMO = *MI.memoperands_begin();
   // Don't set the offset value here because the pointer points to the base of
@@ -3854,6 +3859,12 @@ bool AMDGPUInstructionSelector::selectGlobalLoadLds(MachineInstr &MI) const{
     }
   }
 
+  if (isPreGFX12Async(IntrinsicID)) {
+    int AsyncOpc = AMDGPU::getPreGFX12AsyncOp(Opc);
+    assert(AsyncOpc != -1 && "PreGFX12Async sibling missing");
+    Opc = AsyncOpc;
+  }
+
   auto MIB = BuildMI(*MBB, &MI, DL, TII.get(Opc))
     .addReg(Addr);
 
@@ -3864,7 +3875,6 @@ bool AMDGPUInstructionSelector::selectGlobalLoadLds(MachineInstr &MI) const{
 
   unsigned Aux = MI.getOperand(5).getImm();
   MIB.addImm(Aux & ~AMDGPU::CPol::VIRTUAL_BITS); // cpol
-  MIB.addImm(isAsyncLDSDMA(IntrinsicID));
 
   MachineMemOperand *LoadMMO = *MI.memoperands_begin();
   MachinePointerInfo LoadPtrI = LoadMMO->getPointerInfo();
--- a/llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
+++ b/llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
@@ -3496,7 +3496,7 @@ bool AMDGPUInstructionSelector::selectG_INSERT_VECTOR_ELT(
   return true;
 }
 
-static bool isAsyncLDSDMA(Intrinsic::ID Intr) {
+static bool isPreGFX12Async(Intrinsic::ID Intr) {
   switch (Intr) {
   case Intrinsic::amdgcn_raw_buffer_load_async_lds:
   case Intrinsic::amdgcn_raw_ptr_buffer_load_async_lds:
@@ -3571,6 +3571,12 @@ bool AMDGPUInstructionSelector::selectBufferLoadLds(MachineInstr &MI) const {
     break;
   }
 
+  if (isPreGFX12Async(IntrinsicID)) {
+    int AsyncOpc = AMDGPU::getPreGFX12AsyncOp(Opc);
+    assert(AsyncOpc != -1 && "PreGFX12Async sibling missing");
+    Opc = AsyncOpc;
+  }
+
   MachineBasicBlock *MBB = MI.getParent();
   const DebugLoc &DL = MI.getDebugLoc();
   BuildMI(*MBB, &MI, DL, TII.get(AMDGPU::COPY), AMDGPU::M0)
@@ -3604,7 +3610,6 @@ bool AMDGPUInstructionSelector::selectBufferLoadLds(MachineInstr &MI) const {
       Aux & (IsGFX12Plus ? AMDGPU::CPol::SWZ : AMDGPU::CPol::SWZ_pregfx12)
           ? 1
           : 0); // swz
-  MIB.addImm(isAsyncLDSDMA(IntrinsicID));
 
   MachineMemOperand *LoadMMO = *MI.memoperands_begin();
   // Don't set the offset value here because the pointer points to the base of
@@ -3789,6 +3794,12 @@ bool AMDGPUInstructionSelector::selectGlobalLoadLds(MachineInstr &MI) const{
     }
   }
 
+  if (isPreGFX12Async(IntrinsicID)) {
+    int AsyncOpc = AMDGPU::getPreGFX12AsyncOp(Opc);
+    assert(AsyncOpc != -1 && "PreGFX12Async sibling missing");
+    Opc = AsyncOpc;
+  }
+
   auto MIB = BuildMI(*MBB, &MI, DL, TII.get(Opc))
     .addReg(Addr);
 
@@ -3799,7 +3810,6 @@ bool AMDGPUInstructionSelector::selectGlobalLoadLds(MachineInstr &MI) const{
 
   unsigned Aux = MI.getOperand(5).getImm();
   MIB.addImm(Aux & ~AMDGPU::CPol::VIRTUAL_BITS); // cpol
-  MIB.addImm(isAsyncLDSDMA(IntrinsicID));
 
   MachineMemOperand *LoadMMO = *MI.memoperands_begin();
   MachinePointerInfo LoadPtrI = LoadMMO->getPointerInfo();
--- a/llvm/lib/Target/AMDGPU/BUFInstructions.td
+++ b/llvm/lib/Target/AMDGPU/BUFInstructions.td
@@ -418,14 +418,13 @@ class getBUFVDataRegisterOperandForOp<RegisterOperand Op, bit isTFE> {
 
 class getMUBUFInsDA<list<RegisterOperand> vdataList,
                     list<RegisterOperand> vaddrList, bit isTFE, bit hasRestrictedSOffset,
-                    bit isTrue16, bit isLds> {
+                    bit isTrue16> {
   RegisterOperand vdataClass = !if(!empty(vdataList), ?, !head(vdataList));
   RegisterOperand vaddr_op = !if(!empty(vaddrList), ?, !head(vaddrList));
   RegisterOperand vdata_op = getBUFVDataRegisterOperand<!cast<SIRegisterClassLike>(vdataClass.RegClass).Size, isTFE, isTrue16>.ret;
 
   dag SOffset = !if(hasRestrictedSOffset, (ins SReg_32:$soffset), (ins SCSrc_b32:$soffset));
-  dag IsAsyncOpnd = !if(isLds, (ins i1imm_0:$IsAsync), (ins));
-  dag NonVaddrInputs = !con((ins SReg_128_XNULL:$srsrc), SOffset, (ins Offset:$offset, CPol_0:$cpol, i1imm_0:$swz), IsAsyncOpnd);
+  dag NonVaddrInputs = !con((ins SReg_128_XNULL:$srsrc), SOffset, (ins Offset:$offset, CPol_0:$cpol, i1imm_0:$swz));
 
   dag Inputs = !if(!empty(vaddrList), NonVaddrInputs, !con((ins vaddr_op:$vaddr), NonVaddrInputs));
   dag ret = !if(!empty(vdataList), Inputs, !con((ins vdata_op:$vdata), Inputs));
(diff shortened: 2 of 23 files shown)


arsenm:
> The new mutation is a must-have for correctness: once ASYNCMARK is no longer a global memory object, the scheduler is free to reorder it across async loads/stores and WAIT_ASYNCMARK in ways that break program semantics. Later scheduler stages must remember to register this mutation.

This sounds like a big problem. The constraints need to be correct by construction for all passes to respect them

lijinpei-amd:
> > The new mutation is a must-have for correctness: once ASYNCMARK is no longer a global memory object, the scheduler is free to reorder it across async loads/stores and WAIT_ASYNCMARK in ways that break program semantics. Later scheduler stages must remember to register this mutation.
> 
> This sounds like a big problem. The constraints need to be correct by construction for all passes to respect them

For non-scheduler passes, the constraint is imposed by hasUnmodeledSideEffects.
This pr doesn't remove the hasUnmodeledSideEffects bit for ASYNCMARK, it only exempt it from isGlobalMemoryObject (by default,  hasUnmodeledSideEffects  implies isGlobalMemoryObject ). isGlobalMemoryObject is only used by the scheduler, and needed dependencies is added for scheduler.

llvm/lib/Target/AMDGPU/SIInstrInfo.cpp
```diff
@@ -11443,7 +11443,24 @@ void SIInstrInfo::enforceOperandRCAlignment(MachineInstr &MI,
   MI.addOperand(MachineOperand::CreateReg(NewVR, false, true));
 }
 
+bool SIInstrInfo::isAsyncLDSDMA(const MachineInstr &MI) const {
+  if (!isLDSDMA(MI))
+    return false;
+  if (usesASYNC_CNT(MI))
+    return true;
+  const MachineOperand *Async = getNamedOperand(MI, AMDGPU::OpName::IsAsync);
+  return Async && Async->getImm();
```
RyanRio: This isn't correct, there could be other async ops besides LDSDMA, hence that Async MO

llvm/test/CodeGen/AMDGPU/asyncmark-pregfx12.ll
```diff
@@ -51,23 +51,22 @@ define void @interleaved_global_and_dma(ptr addrspace(1) %foo, ptr addrspace(3)
...
 ; SDAG-NEXT:    ; wave barrier
-; SDAG-NEXT:    s_nop 0
 ; SDAG-NEXT:    global_load_dword v[3:4], off lds
 ; SDAG-NEXT:    ; wave barrier
 ; SDAG-NEXT:    global_load_dword v1, v[3:4], off
 ; SDAG-NEXT:    ; asyncmark
 ; SDAG-NEXT:    ; wait_asyncmark(1)
-; SDAG-NEXT:    s_waitcnt vmcnt(3)
+; SDAG-NEXT:    s_waitcnt vmcnt(2)
 ; SDAG-NEXT:    ds_read_b32 v3, v2
```
RyanRio, line 71: So we can end up with longer waits like this, but other scheduler behavior is more respected now? Is this a good trade off? Please update the comment in the test accordingly.

llvm/lib/Target/AMDGPU/SIInstrInfo.cpp
```diff
@@ -11443,7 +11443,24 @@ void SIInstrInfo::enforceOperandRCAlignment(MachineInstr &MI,
...
+}
+
 bool SIInstrInfo::isGlobalMemoryObject(const MachineInstr *MI) const {
```
ssahasra, line 11553: Not a comment about this change, but in general, this is a bit confusing. Maybe I lack the context in this part of the compiler. But a function called `isGlobalMemoryObject` is assuming that an asyncmark is a global memory object. That sounds weird. And from the comment it seems like the function is actually used in a place where it means something like `isSchedulingBarrier`.

llvm/lib/Target/AMDGPU/SIInstrInfo.h
```diff
@@ -638,6 +638,11 @@ class SIInstrInfo final : public AMDGPUGenInstrInfo {
            (get(Opcode).TSFlags & SIInstrFlags::TENSOR_CNT);
   }
 
+  /// True if MI is an LDS-DMA write that targets the async vmem counter,
+  /// either via the ASYNC_CNT TSFlag or the legacy IsAsync operand on
+  /// MUBUF/FLAT load-to-LDS instructions.
+  bool isAsyncLDSDMA(const MachineInstr &MI) const;
```
ssahasra:
```suggestion
  /// True if MI performs an async operation that needs to be tracked by asyncmarks.
  /// This includes:
  /// - legacy LDS DMA accesses on pre-gfx12
  /// - gfx12-plus operations that increment ASYNC_CNT
  bool isAsync(const MachineInstr &MI) const;
```

llvm/lib/Target/AMDGPU/SIInstrInfo.cpp
ssahasra: > This isn't correct, there could be other async ops besides LDSDMA, hence that Async MO

This function body was simply moved from its existing place, but renamed incorrectly. It's not a new change in this PR. The function `isLDSDMA()` doesn't actually mean what it says. When we pile on more things like laneshared registers, we will have to refactor that function. But at its own level, `isAsync()` is doing the right thing and should retain its current name. See the edit that I suggested below.

llvm/test/CodeGen/AMDGPU/asyncmark-pregfx12.ll
ssahasra: Is this because some non-async access got scheduled earlier on gfx9? If this is limited to gfx9 where async ops share vmcnt with non-async ops, then it's an understandable regression. It's "understandable", but the impact still needs to be assessed.

RyanRio: I could see it happening with future uses of asyncmark too though, so +1 to assessing the impact

llvm/lib/Target/AMDGPU/SIInstrInfo.cpp
RyanRio: Ah yes thanks

nhaehnle (changes requested):
Requiring a DAG mutation for functional correctness is a *very* big NO. Absolutely not.

If you want more accurate modeling of the dependencies of asyncmark/asyncwait, there has to be some other way.

Something that immediately comes to mind is defining a fake physical register `$asyncmark` and adding implicit def-use references on all of:

* Asynchronous instructions
* Asyncmark
* Asyncwait

Conceptually, the `$asyncmark` registers represents "all state in wait counters that is relevant for asyncmark/asyncwait".

With that, it should be possible to make the asyncmark itself `mayLoad == mayStore == 0` (expect some annoyances in convincing the isel machinery to allow this due to how the intrinsic is modeled, but at least the MIR representation would correctly model the dependencies).

lijinpei-amd:
> Requiring a DAG mutation for functional correctness is a _very_ big NO. Absolutely not.
> 
> If you want more accurate modeling of the dependencies of asyncmark/asyncwait, there has to be some other way.
> 
> Something that immediately comes to mind is defining a fake physical register `$asyncmark` and adding implicit def-use references on all of:
> 
> * Asynchronous instructions
> * Asyncmark
> * Asyncwait
> 
> Conceptually, the `$asyncmark` registers represents "all state in wait counters that is relevant for asyncmark/asyncwait".
> 
> With that, it should be possible to make the asyncmark itself `mayLoad == mayStore == 0` (expect some annoyances in convincing the isel machinery to allow this due to how the intrinsic is modeled, but at least the MIR representation would correctly model the dependencies).

Reimplemented using phys reg GFX9_ASYNCcnt to pin the dependency.

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
```diff
@@ -2408,7 +2408,20 @@ bool AMDGPUInstructionSelector::selectG_INTRINSIC_W_SIDE_EFFECTS(
...
+    if (!Subtarget->hasAsyncMark())
+      return false;
+    // Selected here so the MI carries hasSideEffects = 0 (a TableGen
+    // pattern would inherit it from the intrinsic's IntrHasSideEffects).
+    auto MIB = BuildMI(*I.getParent(), &I, I.getDebugLoc(),
+                       TII.get(AMDGPU::ASYNCMARK));
+    // Match the dead-marking that SDAG and TableGen-emitted matchers apply
+    // to other $asynccnt-defining instructions.
+    MIB->addRegisterDead(AMDGPU::GFX9_ASYNCcnt, &TRI);
+    MIB->addRegisterDead(AMDGPU::ASYNCcnt, &TRI);
```
arsenm: Why does this need 2 registers

lijinpei-amd: GFX1250 has 6 bit wave state called ASYNCcnt, but GFX9 has no ASYNCcnt wave state. GFX9 async async dma lds intrinsics boils down to using VMCNT.
Using the same ASYNCcnt for gfx9 would do my work, but I don't know if that would be misleading to readers of llvm code.

lijinpei-amd: Also, I excluded the newly added GFX9_ASYNCcnt from some hazard check in GCNHazardRecognizer, I wonder if  the same is needed for the old ASYNCcnt register but missing.

lijinpei-amd: I can switch to using ASYNCcnt if that is better from your point of view.

lijinpei-amd: BTW I saw ASYNCcnt is both in defs and uses of gfx1250 global/cluster async lds load/store instrs, I  wonder if just a use would be enough, as S_WAIT_ASYNCCNT and ASYNCMARK, WAITA_ASYNCMARK etc will have ASYNCcnt as both defs and uses. Having it as both defs and uses of load/store will give up the opportunity to reorder load/store instrs.

llvm/lib/Target/AMDGPU/BUFInstructions.td
```diff
@@ -418,14 +424,13 @@ class getBUFVDataRegisterOperandForOp<RegisterOperand Op, bit isTFE> {
...
 class getMUBUFInsDA<list<RegisterOperand> vdataList,
                     list<RegisterOperand> vaddrList, bit isTFE, bit hasRestrictedSOffset,
-                    bit isTrue16, bit isLds> {
+                    bit isTrue16> {
   RegisterOperand vdataClass = !if(!empty(vdataList), ?, !head(vdataList));
   RegisterOperand vaddr_op = !if(!empty(vaddrList), ?, !head(vaddrList));
   RegisterOperand vdata_op = getBUFVDataRegisterOperand<!cast<SIRegisterClassLike>(vdataClass.RegClass).Size, isTFE, isTrue16>.ret;
 
   dag SOffset = !if(hasRestrictedSOffset, (ins SReg_32:$soffset), (ins SCSrc_b32:$soffset));
-  dag IsAsyncOpnd = !if(isLds, (ins i1imm_0:$IsAsync), (ins));
```
ssahasra, line 427: This first commit is a significant change of existing "isAsync" infrastructure. It needs to be reviewed and submitted separately so that the scheduler changes can be stacked on it as a separate PR.

llvm/lib/Target/AMDGPU/FLATInstructions.td
```diff
@@ -390,16 +396,18 @@ multiclass FLAT_Global_Store_Pseudo_t16<string opName> {
 
 // Async loads, introduced in gfx1250, will store directly
 // to a DS address in vdst (they will not use M0 for DS addess).
-class FLAT_Global_Load_LDS_Pseudo <string opName, bit EnableSaddr = 0, bit IsAsync = 0, bit IsLegacyLDSDMA = 0> : FLAT_Pseudo<
+class FLAT_Global_Load_LDS_PseudoImpl <string opName, bit EnableSaddr = 0, bit IsAsync = 0, bit IsGFX9Async = 0> : FLAT_Pseudo<
```
ssahasra: `IsGFX9Async` is certainly a much nicer name than "legacy LDS DMA" or just `LDS DMA`. This patch should consistently use the same term in all places where the code currently says LDS DMA.

llvm/lib/Target/AMDGPU/FLATInstructions.td
```diff
@@ -390,16 +396,18 @@ multiclass FLAT_Global_Store_Pseudo_t16<string opName> {
...
   (outs ),
   !con(
        !if(IsAsync, (ins VGPROp_32:$vdst), (ins)),
        !if(EnableSaddr, (ins SReg_64:$saddr, VGPROp_32:$vaddr), (ins VGPROp_64:$vaddr)),
-       (ins flat_offset:$offset, CPol_0:$cpol),
-       !if(IsLegacyLDSDMA, (ins i1imm_0:$IsAsync), (ins))),
+       (ins flat_offset:$offset, CPol_0:$cpol)),
   !if(IsAsync, " $vdst,", "")#" $vaddr"#!if(EnableSaddr, ", $saddr", ", off")#"$offset$cpol"> {
   let LGKM_CNT = 0;
+  // IsGFX9Async is a pure modeling fiction: keep VM_CNT/ASYNC_CNT TSFlags
```
ssahasra: `IsGFX9Async` is not fiction, right? :) There are async operations on GFX9. The corresponding `GFX9_ASYNCcnt` is fiction on the other hand.

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
RyanRio: Why do these instructions need to define/use ASYNCcnt in the first place? I believe it was to get predictable behavior when having to specify manual waits for asynccnt, which is no longer necessary. That should just be removed, in a preliminary PR imo.

lijinpei-amd: WAIT_ASYNCMARK and WAIT_ASYNCMARK won't need to def/use ASYNCcnt as they are already hasSideEffects = 1.
I want to remove ASYNCMARK's hasSideEffects = 1 bit, so will have it def/use ASYNCcnt.

llvm/lib/Target/AMDGPU/BUFInstructions.td
lijinpei-amd: sure

llvm/lib/Target/AMDGPU/FLATInstructions.td
lijinpei-amd: sure

llvm/lib/Target/AMDGPU/FLATInstructions.td
lijinpei-amd: Sure. Will use IsGFX9Async and remove GFX9_ASYNCcnt reg.

lijinpei-amd:
This pr now only contains the td refactor. The second stacked pr is https://github.com/llvm/llvm-project/pull/199533

krzysz00:
Haven't read the PR, but important flag: asyncmark / wait_asyncmark will **also** be used for operations that deal in tensorcnt (the TDM loads/stores) so this PR stack using "asynccnt" as the counter being mutated could cause issues.

lijinpei-amd:
> Haven't read the PR, but important flag: asyncmark / wait_asyncmark will **also** be used for operations that deal in tensorcnt (the TDM loads/stores) so this PR stack using "asynccnt" as the counter being mutated could cause issues.

Currently, TDM load/store is classified as non-async for asyncmark/wait_asyncmark both before and after the PR. In the future, if they need to be classified as async, TENSORcnt can be add to def/use of asyncmark/wait_asyncmark.

krzysz00:
> Currently, TDM load/store is classified as non-async for asyncmark/wait_asyncmark both before and after the PR

There's an active ticket to fix this

lijinpei-amd:
> > Currently, TDM load/store is classified as non-async for asyncmark/wait_asyncmark both before and after the PR
> 
> There's an active ticket to fix this

Can you paste the link to the ticket? I think TENSORcnt can be added to def/use of asyncmark/wait_asyncmark to sequence TDM instrs and them.

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
RyanRio: I mean that the instructions that are tracked with asynccnt - they should not def/use asynccnt

RyanRio: same with s_wait_asynccnt

RyanRio:
> I mean that the instructions that are tracked with asynccnt - they should not def/use asynccnt

Since we're making wait_asyncmark not hasSideEffects, it and other things should def/use something, but let's not call it ASYNCcnt, that's very confusing. Why don't all the instructions that can be tracked by asyncmark use/def something like "AsyncMarker". Asynccnt is the hardware counter, which is not the same as the modeling of async operations, as clearly demonstrated by tdm

krzysz00:
> ince we're making wait_asyncmark not hasSideEffects, it and other things should def/use something, but let's not call it ASYNCcnt, that's very confusing. Why don't all the instructions that can be tracked by asyncmark use/def something like "AsyncMarker". Asynccnt is the hardware counter, which is not the same as the modeling of async operations, as clearly demonstrated by tdm

Yeah, that's what I had in mind here

> Can you paste the link to the ticket? I think TENSORcnt can be added to def/use of asyncmark/wait_asyncmark to sequence TDM instrs and them.

https://amd-hub.atlassian.net/browse/LCOMPILER-1619 is the ticket (with apologies to external readers)

ssahasra:
> > I mean that the instructions that are tracked with asynccnt - they should not def/use asynccnt
> 
> Since we're making wait_asyncmark not hasSideEffects, it and other things should def/use something, but let's not call it ASYNCcnt, that's very confusing. Why don't all the instructions that can be tracked by asyncmark use/def something like "AsyncMarker". Asynccnt is the hardware counter, which is not the same as the modeling of async operations, as clearly demonstrated by tdm

+100 for that.

RyanRio:
Hi is this ready for another review? Looks like you addressed my concern!

arsenm:
> The new mutation is a must-have for correctness: once ASYNCMARK is no longer a global memory object, the scheduler is free to reorder it across async loads/stores and WAIT_ASYNCMARK in ways that break program semantics. Later scheduler stages must remember to register this mutation.

This is still not OK, and the description needs to be updated if this was already fixed

llvm/lib/Target/AMDGPU/SIInstrInfo.cpp
```diff
@@ -10597,6 +10597,10 @@ int SIInstrInfo::pseudoToMCOpcode(int Opcode) const {
       Opcode = MFMAOp;
   }
 
+  // preGFX12Async pseudos share the encoding of their non-async sibling.
+  if (int NonAsync = AMDGPU::getPreGFX12NonAsyncOp(Opcode); NonAsync != -1)
```
arsenm, line 10601: This function should be else if, these instruction classifications are mutually exclusive

PR #189455, opened 2026-03-30
[VPlan] Expand simple SCEVs directly to VPInstructions.

fhahn:
Add initial simple SCEV expansion directly to VPInstructions. To start with, just support expanding SCEV expressions for the vector step (VF * UF). This requires expanding VScale, constants and multiply expressions.

This allows enables CSE for some redundant vscale calls as first step and also enables expanding SCEV expressions in blocks other than the header as follow-ups. For example, this could be useful to avoid some code movement with https://github.com/llvm/llvm-project/pull/189372.

PR #197643, opened 2026-05-14
[VPlan] Expand VPExpandSCEVRecipes to VPInstructions before CSE.

fhahn:
Add expandSCEVExpressions transform that converts VPExpandSCEVRecipes
to VPInstructions where possible, running before CSE so duplicates with
other SCEV expansions (e.g., from addMinimumIterationCheck) are
eliminated. This also reuses existing loop-invariant IR values via
ScalarEvolution::getSCEVValues to avoid redundant computation.

Currently limited to SCEVMulExpr (along with constants, unknowns, and
vscale). Support for SCEVAddExpr and SCEVUDivExpr will follow in
subsequent patches.

Depends on https://github.com/llvm/llvm-project/pull/189455


llvm/include/llvm/Analysis/ScalarEvolution.h
artagnon, line 1: Why are these changes necessary?

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -6163,6 +6163,10 @@ DenseMap<const SCEV *, Value *> LoopVectorizationPlanner::executePlan(
       CM.requiresScalarEpilogue(BestVF.isVector()), &BestVPlan.getVFxUF(),
       MaxRuntimeStep);
   VPlanTransforms::materializeFactors(BestVPlan, VectorPH, BestVF);
+  // Limit expansions to VPInstruction to when not vectorizing the main epilogue
+  // loop.
```
artagnon: Would be good to elaborate on the reason?

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -5016,32 +5016,52 @@ void VPlanTransforms::materializeFactors(VPlan &Plan, VPBasicBlock *VectorPH,
...
 
+void VPlanTransforms::expandSCEVExpressions(VPlan &Plan, ScalarEvolution &SE,
+                                            Loop &OrigLoop) {
+  auto *Entry = cast<VPIRBasicBlock>(Plan.getEntry());
+  VPBuilder Builder(Entry, Entry->begin());
+  VPSCEVExpander Expander(Builder, Plan, SE, OrigLoop);
+
+  // Expand VPExpandSCEVRecipes to VPInstructions using VPSCEVExpander. During
+  // the transition, unsupported SCEV expressions are still expanded to
+  // VPExpandSCEVRecipes.
```
artagnon: Hm, this is highly confusing! We don't "expand" to ExpandSCEVRecipes; we simply fall back to not expanding the ExpandSCEVRecipe to VPInstructions, and use SCEVExpander to expand to Instructions?

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
```diff
@@ -839,3 +842,62 @@ bool vputils::isUsedByLoadStoreAddress(const VPValue *V) {
...
+    for (const SCEVUse &Op : drop_begin(Mul->operands()))
+      Result = Builder.createOverflowingOp(Instruction::Mul,
+                                           {Result, expand(Op)}, WrapFlags, DL);
+    return Result;
+  }
+  default:
+    // Unsupported SCEV kind; fall back to VPExpandSCEVRecipe.
+    assert(Builder.getInsertBlock() == Plan.getEntry() &&
+           "VPExpandSCEVRecipe fallback requires insertion in the entry block");
+    return Builder.createExpandSCEV(S);
```
artagnon: Just return nullptr here?

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
```diff
@@ -839,3 +842,62 @@ bool vputils::isUsedByLoadStoreAddress(const VPValue *V) {
...
   return false;
 }
+
+/// Try to find a loop-invariant IR value for \p S in \p OrigLoop's preheader
+/// that can be reused. Returns the corresponding live-in VPValue, or nullptr
+/// if no reusable IR value is found.
+VPValue *VPSCEVExpander::tryToReuseIRValue(const SCEV *S) {
+  if (isa<SCEVConstant, SCEVUnknown>(S))
+    return nullptr;
+  BasicBlock *PH = OrigLoop.getLoopPreheader();
```
artagnon: Hm, can we not use Plan.getScalarPreheader()? Confused about why do we need the OrigLoop here?

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
```diff
@@ -839,3 +842,62 @@ bool vputils::isUsedByLoadStoreAddress(const VPValue *V) {
   }
   return false;
 }
+
+/// Try to find a loop-invariant IR value for \p S in \p OrigLoop's preheader
+/// that can be reused. Returns the corresponding live-in VPValue, or nullptr
+/// if no reusable IR value is found.
+VPValue *VPSCEVExpander::tryToReuseIRValue(const SCEV *S) {
+  if (isa<SCEVConstant, SCEVUnknown>(S))
+    return nullptr;
```
artagnon, line 905: Why? Can we not get a live-in for these, as in getOrCreateVPValueForSCEVExpr?

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
```diff
@@ -839,3 +842,62 @@ bool vputils::isUsedByLoadStoreAddress(const VPValue *V) {
...
+    // sibling branches may not dominate this loop's preheader.
+    if (I->getParent() != PH)
+      continue;
+    SmallVector<Instruction *> DropPoisonGeneratingInsts;
+    if (!SE.canReuseInstruction(S, I, DropPoisonGeneratingInsts))
+      continue;
+    for (Instruction *DropI : DropPoisonGeneratingInsts)
+      SCEVExpander::dropPoisonGeneratingAnnotationsAndReinfer(SE, DropI);
+    return Plan.getOrAddLiveIn(V);
+  }
```
artagnon, line 922: I think this can be absorbed into Mel's getVPIRValueForSCEVExpr?

llvm/lib/Transforms/Vectorize/VPlanTransforms.h
```diff
@@ -429,10 +427,18 @@ struct VPlanTransforms {
...
+  /// values are reused as live-ins.
+  static void expandSCEVExpressions(VPlan &Plan, ScalarEvolution &SE,
+                                    Loop &OrigLoop);
+
+  /// Expand remaining VPExpandSCEVRecipes in \p Plan's entry block using
+  /// SCEVExpander. Each VPExpandSCEVRecipe is replaced with a live-in wrapping
+  /// the expanded IR value. A mapping from SCEV expressions to their expanded
+  /// IR value is returned.
   static DenseMap<const SCEV *, Value *> expandSCEVs(VPlan &Plan,
                                                      ScalarEvolution &SE);
```
artagnon, line 466: Can we simply have one function that tries to expand to VPInstructions, and falls back to SCEVExpander expanding into Instructions?

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
fhahn: I don't think it can, this helper may drop poison generating flags (i.e. modify exising IR) when re-using IR instructions; so it must only be called when we are sure not to abandon vectorization

llvm/include/llvm/Analysis/ScalarEvolution.h
fhahn: VPSCEVExpander need to be a friend class of ScalarEvolution, to get access to `getSCEVValues`, like SCEVExpander

llvm/lib/Transforms/Vectorize/VPlanTransforms.h
fhahn: This is stacked PR, with the inital version just expanding to VPinsructions via the builder https://github.com/llvm/llvm-project/pull/189455.


This commit moves the VPInstruction expansion earlier, so we can benefit from CSE. In theory we could move the IR expansion earlier as well, but currently it creates IR instructions directly, which is why it is as close to execute as possible

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: Added, thanks. The reason is that the epilogue path still relies on SCEVs being expanded to instructions directly, to re-use between both main and epilogue vector loop.

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
fhahn: The early transform now only expands VPExpandSCEV to VPInstructions (and skips if it cannot be expanded).

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
fhahn: We can, but for SCEV constants we still need the dedicated handling in the expander.

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
fhahn: We can use the entry block, which is a VPIRBasicBlock at this point. The scalar preheader retrieved from VPlan only dominates the scalar loop, but not the vector loop

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
fhahn: Updated to do that for now, this also requires adding a bailout above if any operand is nullptr though. Updated in https://github.com/llvm/llvm-project/pull/189455

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
artagnon: Would be good to update this comment?

llvm/lib/Transforms/Vectorize/VPlanTransforms.h
```diff
@@ -432,10 +427,16 @@ struct VPlanTransforms {
   static void materializeFactors(VPlan &Plan, VPBasicBlock *VectorPH,
                                  ElementCount VF);
 
-  /// Expand VPExpandSCEVRecipes in \p Plan's entry block. Each
-  /// VPExpandSCEVRecipe is replaced with a live-in wrapping the expanded IR
-  /// value. A mapping from SCEV expressions to their expanded IR value is
-  /// returned.
+  /// Try to expand VPExpandSCEVRecipes in \p Plan's entry block to
+  /// VPInstructions. Recipes that cannot be expanded (like casts, min/max) are
+  /// kept for later IR-level expansion
```
artagnon:
```suggestion
  /// VPInstructions. Unsupported recipes (like casts, min/max) are
  /// kept for later IR-level expansion.
```

llvm/lib/Transforms/Vectorize/VPlanUtils.h
```diff
@@ -176,6 +176,29 @@ VPSingleDefRecipe *findHeaderMask(VPlan &Plan);
 
 } // namespace vputils
 
+/// Lightweight SCEV-to-VPlan expander. Converts SCEV expressions into
+/// VPInstructions where possible, falling back to VPExpandSCEVRecipe for
+/// unsupported expressions (casts, min/max).
```
artagnon:
```suggestion
/// VPInstructions where possible, falling back to nullptr for unsupported expressions (casts, min/max).
```

artagnon (commented):
Let's attack the patch this is dependent on for now -- I've left a review.

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
fhahn: updated thanks

llvm/lib/Transforms/Vectorize/VPlanUtils.h
fhahn: updated thanks

fhahn (commented):
ping

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
```diff
@@ -892,7 +893,38 @@ bool vputils::isUsedByLoadStoreAddress(const VPValue *V) {
...
+  BasicBlock *PH =
+      cast<VPIRBasicBlock>(Builder.getPlan().getEntry())->getIRBasicBlock();
+  for (Value *V : SE.getSCEVValues(S)) {
+    if (V->getType() != S->getType())
+      continue;
+    // Only reuse instructions in the plan's entry block, as instructions in
+    // sibling branches may not dominate the entry block.
+    auto *I = dyn_cast<Instruction>(V);
+    if (!I)
+      return Builder.getPlan().getOrAddLiveIn(V);
```
artagnon: Could factor Builder.getPlan() into a Plan variable on top?

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -5281,32 +5281,54 @@ void VPlanTransforms::materializeAliasMaskCheckBlock(
   Plan.getVFxUF().replaceAllUsesWith(ClampedVF);
 }
 
+void VPlanTransforms::expandSCEVsToVPInstructions(VPlan &Plan,
+                                                  ScalarEvolution &SE) {
+  auto *Entry = cast<VPIRBasicBlock>(Plan.getEntry());
```
artagnon: Why is the cast to VPIRBB necessary?

llvm/lib/Transforms/Vectorize/VPlanUtils.h
```diff
@@ -181,6 +181,28 @@ VPValue *findIncomingAliasMask(const VPlan &Plan);
 
 } // namespace vputils
 
+/// Lightweight SCEV-to-VPlan expander. Converts SCEV expressions into
+/// VPInstructions where possible, and returning nullptr for unsupported
+/// expressions (like adds, casts, min/max).
+class VPSCEVExpander {
+  VPBuilder &Builder;
+  ScalarEvolution &SE;
+  DebugLoc DL;
```
artagnon, line 190:
```suggestion
  const DebugLoc &DL;
```

llvm/lib/Transforms/Vectorize/VPlanUtils.h
```diff
@@ -181,6 +181,28 @@ VPValue *findIncomingAliasMask(const VPlan &Plan);
...
+
+  /// Try to find a loop-invariant IR value in the plan's entry block whose
+  /// SCEV matches \p S. Returns the corresponding live-in VPValue, or nullptr
+  /// if none is found.
+  VPValue *tryToReuseIRValue(const SCEV *S);
+
+public:
+  VPSCEVExpander(VPBuilder &Builder, ScalarEvolution &SE,
+                 DebugLoc DL = DebugLoc())
```
artagnon:
```suggestion
                 DebugLoc DL = {})
```

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
```diff
@@ -892,7 +893,38 @@ bool vputils::isUsedByLoadStoreAddress(const VPValue *V) {
...
+/// that can be reused. Returns the corresponding live-in VPValue, or nullptr
+/// if no reusable IR value is found.
+VPValue *VPSCEVExpander::tryToReuseIRValue(const SCEV *S) {
+  if (isa<SCEVConstant, SCEVUnknown>(S))
+    return nullptr;
+  BasicBlock *PH =
+      cast<VPIRBasicBlock>(Builder.getPlan().getEntry())->getIRBasicBlock();
+  for (Value *V : SE.getSCEVValues(S)) {
+    if (V->getType() != S->getType())
+      continue;
```
artagnon: Hm, when does this happen?

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
fhahn: it's not needed, removed thanks

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
fhahn: I added this to match IR-based SCEVExpander behavior. It looks like this is actually never hit for the IR-based one as well, removing here and will check if there are indeed no cases where the IR expander hits this also

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
fhahn: done thanks

llvm/lib/Transforms/Vectorize/VPlanUtils.h
fhahn: DebugLoc is already a wrapper around pointer, IIUC usually it is used directly by value to avoid another level of indirection.

llvm/lib/Transforms/Vectorize/VPlanUtils.h
fhahn: Done thanks

artagnon (approved):
LGTM, thanks! DL as a non-const-ref is also fine, in case there is some issue changing it?

fhahn (commented):
> LGTM, thanks! DL as a non-const-ref is also fine, in case there is some issue changing it?

responded inline, but forgot to submit comments.... 

> DebugLoc is already a wrapper around pointer, IIUC usually it is used directly by value to avoid another level of indirection.

PR #79319, opened 2024-01-24
[libc] Use `LIBC_COPT_PUBLIC_PACKAGING` for hermetic and integration tests.

lntue:
Currently all unit, hermetic, and integration tests are linked against `__internal__` entry points.  This change switches the hermetic and integration tests to be linked against the public versions of the entry points.

PR #197644, opened 2026-05-14
[libc] Refactor getopt to use standard reset and internal test hook

kaladron:
Implemented standard POSIX optind = 0 reset logic in getopt and getopt_r.

To support hermetic testing of stderr error messages, provided an internal __llvm_libc_getopt_set_errorstream function. Added this as a library entrypoint for linkage but omitted it from public headers to keep the namespace clean.

Refactored getopt_r to use a reference-based context to avoid redundant dereferencing of global variables while maintaining standard compliance.

Updated getopt_test.cpp to use the new internal hook and standard optind reset.

Simplified the internal implementation by removing the RefWrapper abstraction and grouping all state into a pointer-based GetoptContext.

### DIFF
--- a/libc/src/unistd/getopt.h
+++ b/libc/src/unistd/getopt.h
@@ -16,8 +16,15 @@
 namespace LIBC_NAMESPACE_DECL {
 
 namespace impl {
-void set_getopt_state(char **, int *, int *, unsigned *, int *, FILE *);
+extern "C" {
+extern char *optarg;
+extern int optind;
+extern int optopt;
+extern int opterr;
 }
+} // namespace impl
+
+void __llvm_libc_getopt_set_errorstream(FILE *);
 
 int getopt(int argc, char *const argv[], const char *optstring);
 
--- a/libc/test/src/unistd/getopt_test.cpp
+++ b/libc/test/src/unistd/getopt_test.cpp
@@ -15,21 +15,8 @@
 
 using LIBC_NAMESPACE::cpp::array;
 
-namespace test_globals {
-char *optarg;
-int optind = 1;
-int optopt;
-int opterr = 1;
-
-unsigned optpos;
-} // namespace test_globals
-
-// This can't be a constructor because it will get run before the constructor
-// which sets the default state in getopt.
 void set_state(FILE *errstream) {
-  LIBC_NAMESPACE::impl::set_getopt_state(
-      &test_globals::optarg, &test_globals::optind, &test_globals::optopt,
-      &test_globals::optpos, &test_globals::opterr, errstream);
+  LIBC_NAMESPACE::__llvm_libc_getopt_set_errorstream(errstream);
 }
 
 static void my_memcpy(char *dest, const char *src, size_t size) {
@@ -68,12 +55,14 @@ struct LlvmLibcGetoptTest : public LIBC_NAMESPACE::testing::Test {
   void SetUp() override {
     ASSERT_TRUE(!!(errstream = memopen(&pos)));
     set_state(errstream);
-    ASSERT_EQ(test_globals::optind, 1);
+    LIBC_NAMESPACE::impl::optind = 0;
+    LIBC_NAMESPACE::getopt(0, nullptr, "");
   }
 
   void TearDown() override {
-    test_globals::optind = 1;
-    test_globals::opterr = 1;
+    LIBC_NAMESPACE::impl::optind = 0;
+    LIBC_NAMESPACE::getopt(0, nullptr, "");
+    LIBC_NAMESPACE::impl::opterr = 1;
   }
 };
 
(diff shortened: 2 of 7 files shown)


kaladron:
Heya Alexey!  This patch does two main things:

1) Adds the much more common way of resetting state to our getopt implementation
2) Exposes an internal way of redirecting the error messages to a stream instead of stderr.

This came up because @lntue had a PR two years ago #79319 that made our hermeteic testing work against the external interfaces and I wanted to resurrect that (https://github.com/kaladron/llvm-project/tree/public-testing).  I was going to have to expose set_getopt_state and wound up doing more general clean up pass on this file.

libc/src/unistd/getopt.h
```diff
@@ -16,8 +16,15 @@
...
-void set_getopt_state(char **, int *, int *, unsigned *, int *, FILE *);
+extern "C" {
+extern char *optarg;
+extern int optind;
+extern int optopt;
+extern int opterr;
 }
+} // namespace impl
+
+void __llvm_libc_getopt_set_errorstream(FILE *);
```
vonosmas, line 27: I'm confused a bit on why you need a dedicated entrypoint for this? If you declare this function in a private llvm-libc implementation header, and you include this header from unit tests, then you call simply call this function as a "regular" C++ function `LIBC_NAMESPACE::__llvm_libc_getopt_set_errorstream(errstream);` (which you already do) in the unit tests.

Why is making it an entrypoint and wrapping it in LLVM_LIBC_FUNCTION exporting macro really needed?

vonosmas: FWIW you might not even need `__llvm_libc_` prefix then, since you're only accessing this function through llvm-libc namespace anyway.

libc/test/src/unistd/getopt_test.cpp
```diff
@@ -68,12 +55,14 @@ struct LlvmLibcGetoptTest : public LIBC_NAMESPACE::testing::Test {
   void SetUp() override {
     ASSERT_TRUE(!!(errstream = memopen(&pos)));
     set_state(errstream);
-    ASSERT_EQ(test_globals::optind, 1);
+    LIBC_NAMESPACE::impl::optind = 0;
+    LIBC_NAMESPACE::getopt(0, nullptr, "");
```
vonosmas, line 59: I think you need to set `opterr` to 1 here as well - to make sure that all the proper environment setup is done at `SetUp` state of a test case, regardless of what was happening before.

libc/test/src/unistd/getopt_test.cpp
```diff
@@ -68,12 +55,14 @@ struct LlvmLibcGetoptTest : public LIBC_NAMESPACE::testing::Test {
...
   }
 
   void TearDown() override {
-    test_globals::optind = 1;
-    test_globals::opterr = 1;
+    LIBC_NAMESPACE::impl::optind = 0;
```
vonosmas, line 63: Conversely, why do you need this here? I think we shouldn't do a whole lot of work at TearDown, assuming SetUp would clean the state for us properly.

libc/test/src/unistd/getopt_test.cpp
```diff
@@ -68,12 +55,14 @@ struct LlvmLibcGetoptTest : public LIBC_NAMESPACE::testing::Test {
   void SetUp() override {
     ASSERT_TRUE(!!(errstream = memopen(&pos)));
     set_state(errstream);
-    ASSERT_EQ(test_globals::optind, 1);
+    LIBC_NAMESPACE::impl::optind = 0;
```
vonosmas, line 58: Why is it 0, and not 1 ? The result of `getopt()` call with `optind=0` is unspecified in POSIX. I think 1 should be the correct default to start the test case with.

vonosmas: I can see that you kinda use `0` as an instruction to "clear out the state - set `optind=1 ; optpos=0`, but I still think you should generally test calling the `getopt` with `optind=1`, because that's how most users are expect to call it anyway. If you want to add an additional hook to clear out the `optpos` global state before different test cases in getopt_test, I think you can do that with a similar private function only visible to tests.

libc/src/unistd/getopt.h
kaladron: getopt_test is a hermetic test, so we need entrypoints if we're going to both to hermetic tests only using the public interface. (From the goal of #79319)

vonosmas: > both to hermetic tests only using the public interface

Could you clarify what do you mean by this? Or maybe I don't understand the concept of "hermetic tests" properly. The test does include `#include "src/unistd/getopt.h"` -- i.e. it has access to the functions in the llvm-libc internal namespace to invoke them directly.

Is it expected to change (and the test will just use whatever's available in the public llvm-libc produced `<unistd.h>` header?) If yes, then this CL moves in the opposite direction by moving the declarations of global state to `libc/src/unistd/getopt.h` and referencing them in the test cases with `LIBC_NAMESPACE::impl` prefix. Or is this expected to change in the future as well?

In either case, I'm not in favor of conflating two things in the same PR. The refactoring here and the test case updates look largely good (module some comments). I would rather merge them now, and then deal with test migration to a different setup / set of entrypoints in a later PR.

PR #197645, opened 2026-05-14
[lldb-dap] Don't emit memory reference for constants

DrSergei:
Some time ago I noticed that constants in Watch Window have meaningless memory references (their values were interpreted as load addresses). I am not an expert in LLDB internals, but hope it is correct fix.

### DIFF
--- a/lldb/tools/lldb-dap/Handler/EvaluateRequestHandler.cpp
+++ b/lldb/tools/lldb-dap/Handler/EvaluateRequestHandler.cpp
@@ -118,11 +118,11 @@ EvaluateRequestHandler::Run(const EvaluateArguments &arguments) const {
   if (value.GetError().Fail())
     return ToError(value.GetError(), /*show_user=*/false);
 
-  // Check original value type before calling `Persist`, because it change value
-  // type to const result
+  // Check the original value type before calling `Persist`, because it changes
+  // the type to const result
   if (lldb::addr_t addr = value.GetLoadAddress();
-      value.GetValueType() != lldb::eValueTypeConstResult &&
-      addr != LLDB_INVALID_ADDRESS)
+      addr != LLDB_INVALID_ADDRESS &&
+      value.GetValueType() != lldb::eValueTypeConstResult)
     body.memoryReference = EncodeMemoryReference(addr);
 
   if (is_repl_context) {
--- a/lldb/tools/lldb-dap/Handler/EvaluateRequestHandler.cpp
+++ b/lldb/tools/lldb-dap/Handler/EvaluateRequestHandler.cpp
@@ -118,6 +118,13 @@ EvaluateRequestHandler::Run(const EvaluateArguments &arguments) const {
   if (value.GetError().Fail())
     return ToError(value.GetError(), /*show_user=*/false);
 
+  // Check original value type before calling `Persist`, because it change value
+  // type to const result
+  if (lldb::addr_t addr = value.GetLoadAddress();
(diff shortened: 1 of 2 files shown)


lldb/tools/lldb-dap/Handler/EvaluateRequestHandler.cpp
da-viper:
```suggestion
  // Check the original value type before calling `Persist`, because it changes the
  // value type to `eValueTypeConstResult`.
```

da-viper: Nit: check the address first since we have computed it and it may short circuit there.

da-viper (approved):

da-viper:
Not to familiar with `PDB` parsing and data types but It uses `eValueTypeConstResult` to represent constant types. And not sure if that constant has a file address. see https://github.com/llvm/llvm-project/blob/5366692aeefd1101cb8dfde7fb65c9565b476d96/lldb/source/Plugins/SymbolFile/PDB/SymbolFilePDB.cpp#L1041-L1043

charles-zablit:
Is this issue on Windows specifically?

From what I understand, `eValueTypeConstResult` should only be used for expression evaluation.

Changing
```cpp
 case PDB_DataKind::Constant: 
   scope = eValueTypeConstResult; 
   break; 
```
to
```cpp
 case PDB_DataKind::Constant: 
   scope = eValueTypeVariableStatic; 
   break; 
```

would be a better fix.

DrSergei:
> Is this issue on Windows specifically?
> 
I noticed this behavior on Linux and didn't test current fix on Windows

charles-zablit:
> > Is this issue on Windows specifically?
> 
> I noticed this behavior on Linux and didn't test current fix on Windows

I see, the Windows fix and your fix can be 2 separate fixes then I think.

jimingham:
> Is this issue on Windows specifically?
> 

I haven't read this thread yet, but just responding to this particular point:

> From what I understand, `eValueTypeConstResult` should only be used for expression evaluation.

This is not true.  ValueObjectConstResult is used in the ExpressionVariable class as the backing for expression results.  But also the ValueObject::CreateValueObjectFrom*** calls all use ValueObjectConstResult objects to back the ValueObjects they create, (as do the new ValueObject::CreateChildValueObjectFrom***) calls.  And those get used extensively by the the Synthetic Child Providers.  In fact, if you use a "Value providing Synthetic Child Provider" the root ValueObject will often make and hand out a ValueObjectConstResult instead of itself.  So you will even see ValueObjectConstResult objects returned from Frame::FindVariables.

> 
> Changing
> 
> ```c++
>  case PDB_DataKind::Constant: 
>    scope = eValueTypeConstResult; 
>    break; 
> ```
> 
> to
> 
> ```c++
>  case PDB_DataKind::Constant: 
>    scope = eValueTypeVariableStatic; 
>    break; 
> ```
> 
> would be a better fix.

charles-zablit (approved):
LGTM given Jim's answer.

DrSergei:
I double checked constants.
Before:
evaluate("2 + 2") has not memory reference
evaluate("10") has memory reference
evaluate("0x00007fffffffd684") has memory reference
evaluate("abc") has memory reference
After:
None has memory reference.

I think in case when constant is valid address in memory current behavior is good. I am not sure that checking every load address is a good solution. I will wait 2 weeks before merge if anybody has another suggestion how to avoid this regression

PR #197647, opened 2026-05-14
[clang][Sema] Guard diagnose_if value-dependent conditions

### LINKED ISSUE #197625: Assertion `!isValueDependent() && "Expression evaluator can't be called on a dependent expression."' failed.
https://godbolt.org/z/oKdv9b1zb

The testfile is very long, but creduce failed in my own environment. I used https://github.com/llvm/llvm-project/blob/main/clang/test/SemaCXX/diagnose_if.cpp as the input.
It seems to be introduced since clang 14.0.0 on godbolt. It can be triggered only with the assertion build.
https://godbolt.org/z/7n6vfdTK1

```
fatal error: too many errors emitted, stopping now [-ferror-limit=]
clang++: /root/llvm-project/llvm/tools/clang/lib/AST/ExprConstant.cpp:22208: bool clang::Expr::EvaluateWithSubstitution(clang::APValue&, clang::ASTContext&, const clang::FunctionDecl*, llvm::ArrayRef<const clang::Expr*>, const clang::Expr*) const: Assertion `!isValueDependent() && "Expression evaluator can't be called on a dependent expression."' failed.
Stack dump:
0.	Program arguments: /opt/compiler-explorer/clang-assertions-trunk/bin/clang++ -g -o /app/output.s -mllvm --x86-asm-syntax=intel -fno-verbose-asm -S --gcc-toolchain=/opt/compiler-explorer/gcc-snapshot -fcolor-diagnostics -fno-crash-diagnostics -fsyntax-only <source>
1.	<source>:601:36: current parser token ';'
2.	<source>:593:1: parsing namespace 'initializer_lists'
3.	<source>:600:12: parsing function body 'initializer_lists::run'
4.	<source>:600:12: in compound statement ('{}')
 #0 0x000000000453f658 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x453f658)
 #1 0x000000000453c484 llvm::sys::RunSignalHandlers() (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x453c484)
 #2 0x000000000453caa4 llvm::sys::CleanupOnSignal(unsigned long) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x453caa4)
 #3 0x0000000004479268 CrashRecoverySignalHandler(int) CrashRecoveryContext.cpp:0:0
 #4 0x0000711fae242520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #5 0x0000711fae2969fc pthread_kill (/lib/x86_64-linux-gnu/libc.so.6+0x969fc)
 #6 0x0000711fae242476 gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x42476)
 #7 0x0000711fae2287f3 abort (/lib/x86_64-linux-gnu/libc.so.6+0x287f3)
 #8 0x0000711fae22871b (/lib/x86_64-linux-gnu/libc.so.6+0x2871b)
 #9 0x0000711fae239e96 (/lib/x86_64-linux-gnu/libc.so.6+0x39e96)
#10 0x0000000008297541 clang::Expr::EvaluateWithSubstitution(clang::APValue&, clang::ASTContext&, clang::FunctionDecl const*, llvm::ArrayRef<clang::Expr const*>, clang::Expr const*) const (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x8297541)
#11 0x0000000007788419 clang::Sema::diagnoseArgDependentDiagnoseIfAttrs(clang::FunctionDecl const*, clang::Expr const*, llvm::ArrayRef<clang::Expr const*>, clang::SourceLocation)::'lambda'(clang::DiagnoseIfAttr const*)::operator()(clang::DiagnoseIfAttr const*) const SemaOverload.cpp:0:0
#12 0x00000000077aa67b bool diagnoseDiagnoseIfAttrsWith<clang::Sema::diagnoseArgDependentDiagnoseIfAttrs(clang::FunctionDecl const*, clang::Expr const*, llvm::ArrayRef<clang::Expr const*>, clang::SourceLocation)::'lambda'(clang::DiagnoseIfAttr const*)>(clang::Sema&, clang::NamedDecl const*, bool, clang::SourceLocation, clang::Sema::diagnoseArgDependentDiagnoseIfAttrs(clang::FunctionDecl const*, clang::Expr const*, llvm::ArrayRef<clang::Expr const*>, clang::SourceLocation)::'lambda'(clang::DiagnoseIfAttr const*)&&) (.constprop.0) SemaOverload.cpp:0:0
#13 0x00000000077aa8a6 clang::Sema::diagnoseArgDependentDiagnoseIfAttrs(clang::FunctionDecl const*, clang::Expr const*, llvm::ArrayRef<clang::Expr const*>, clang::SourceLocation) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x77aa8a6)
#14 0x0000000007113ee5 clang::Sema::checkCall(clang::NamedDecl*, clang::FunctionProtoType const*, clang::Expr const*, llvm::ArrayRef<clang::Expr const*>, bool, clang::SourceLocation, clang::SourceRange, clang::VariadicCallType) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x7113ee5)
#15 0x00000000071163e9 clang::Sema::CheckConstructorCall(clang::FunctionDecl*, clang::QualType, llvm::ArrayRef<clang::Expr const*>, clang::FunctionProtoType const*, clang::SourceLocation) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x71163e9)
#16 0x00000000072f19ff clang::Sema::CompleteConstructorCall(clang::CXXConstructorDecl*, clang::QualType, llvm::MutableArrayRef<clang::Expr*>, clang::SourceLocation, llvm::SmallVectorImpl<clang::Expr*>&, bool, bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x72f19ff)
#17 0x00000000075605b7 PerformConstructorInitialization(clang::Sema&, clang::InitializedEntity const&, clang::InitializationKind const&, llvm::MutableArrayRef<clang::Expr*>, clang::InitializationSequence::Step const&, bool&, bool, bool, clang::SourceLocation, clang::SourceLocation) SemaInit.cpp:0:0
#18 0x000000000756f5b1 clang::InitializationSequence::Perform(clang::Sema&, clang::InitializedEntity const&, clang::InitializationKind const&, llvm::MutableArrayRef<clang::Expr*>, clang::QualType*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x756f5b1)
#19 0x00000000074d0239 clang::Sema::BuildCXXTypeConstructExpr(clang::TypeSourceInfo*, clang::SourceLocation, llvm::MutableArrayRef<clang::Expr*>, clang::SourceLocation, bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x74d0239)
#20 0x00000000074d0ba9 clang::Sema::ActOnCXXTypeConstructExpr(clang::OpaquePtr<clang::QualType>, clang::SourceLocation, llvm::MutableArrayRef<clang::Expr*>, clang::SourceLocation, bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x74d0ba9)
#21 0x0000000006f2dbe6 clang::Parser::ParseCXXTypeConstructExpression(clang::DeclSpec const&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6f2dbe6)
#22 0x0000000006f122db clang::Parser::ParseCastExpression(clang::CastParseKind, bool, bool&, clang::TypoCorrectionTypeBehavior, bool, bool*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6f122db)
#23 0x0000000006f12e37 clang::Parser::ParseCastExpression(clang::CastParseKind, bool, clang::TypoCorrectionTypeBehavior, bool, bool*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6f12e37)
#24 0x0000000006f12ec9 clang::Parser::ParseAssignmentExpression(clang::TypoCorrectionTypeBehavior) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6f12ec9)
#25 0x0000000006f17509 clang::Parser::ParseExpression(clang::TypoCorrectionTypeBehavior) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6f17509)
#26 0x0000000006f9ec89 clang::Parser::ParseExprStatement(clang::Parser::ParsedStmtContext) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6f9ec89)
#27 0x0000000006f960b4 clang::Parser::ParseStatementOrDeclarationAfterAttributes(llvm::SmallVector<clang::Stmt*, 24u>&, clang::Parser::ParsedStmtContext, clang::SourceLocation*, clang::ParsedAttributes&, clang::ParsedAttributes&, clang::LabelDecl*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6f960b4)
#28 0x0000000006f96a99 clang::Parser::ParseStatementOrDeclaration(llvm::SmallVector<clang::Stmt*, 24u>&, clang::Parser::ParsedStmtContext, clang::SourceLocation*, clang::LabelDecl*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6f96a99)
#29 0x0000000006f9f40e clang::Parser::ParseCompoundStatementBody(bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6f9f40e)
#30 0x0000000006f9fc2a clang::Parser::ParseFunctionStatementBody(clang::Decl*, clang::Parser::ParseScope&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6f9fc2a)
#31 0x0000000006ea4123 clang::Parser::ParseFunctionDefinition(clang::ParsingDeclarator&, clang::Parser::ParsedTemplateInfo const&, clang::LateParsedAttrList*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6ea4123)
#32 0x0000000006edae33 clang::Parser::ParseDeclGroup(clang::ParsingDeclSpec&, clang::DeclaratorContext, clang::ParsedAttributes&, clang::Parser::ParsedTemplateInfo&, clang::SourceLocation*, clang::Parser::ForRangeInit*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6edae33)
#33 0x0000000006e976cb clang::Parser::ParseDeclOrFunctionDefInternal(clang::ParsedAttributes&, clang::ParsedAttributes&, clang::ParsingDeclSpec&, clang::AccessSpecifier) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6e976cb)
#34 0x0000000006e97ecf clang::Parser::ParseDeclarationOrFunctionDefinition(clang::ParsedAttributes&, clang::ParsedAttributes&, clang::ParsingDeclSpec*, clang::AccessSpecifier) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6e97ecf)
#35 0x0000000006e9f951 clang::Parser::ParseExternalDeclaration(clang::ParsedAttributes&, clang::ParsedAttributes&, clang::ParsingDeclSpec*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6e9f951)
#36 0x0000000006efaf75 clang::Parser::ParseInnerNamespace(llvm::SmallVector<clang::Parser::InnerNamespaceInfo, 4u> const&, unsigned int, clang::SourceLocation&, clang::ParsedAttributes&, clang::BalancedDelimiterTracker&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6efaf75)
#37 0x0000000006efdf69 clang::Parser::ParseNamespace(clang::DeclaratorContext, clang::SourceLocation&, clang::SourceLocation) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6efdf69)
#38 0x0000000006ee3a80 clang::Parser::ParseDeclaration(clang::DeclaratorContext, clang::SourceLocation&, clang::ParsedAttributes&, clang::ParsedAttributes&, clang::SourceLocation*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6ee3a80)
#39 0x0000000006e9f976 clang::Parser::ParseExternalDeclaration(clang::ParsedAttributes&, clang::ParsedAttributes&, clang::ParsingDeclSpec*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6e9f976)
#40 0x0000000006ea0865 clang::Parser::ParseTopLevelDecl(clang::OpaquePtr<clang::DeclGroupRef>&, clang::Sema::ModuleImportState&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6ea0865)
#41 0x0000000006e8174a clang::ParseAST(clang::Sema&, bool, bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x6e8174a)
#42 0x000000000537f275 clang::FrontendAction::Execute() (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x537f275)
#43 0x00000000052fddf8 clang::CompilerInstance::ExecuteAction(clang::FrontendAction&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x52fddf8)
#44 0x000000000547bb4d clang::ExecuteCompilerInvocation(clang::CompilerInstance*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x547bb4d)
#45 0x0000000000df703b cc1_main(llvm::ArrayRef<char const*>, char const*, void*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0xdf703b)
#46 0x0000000000ded99a ExecuteCC1Tool(llvm::SmallVectorImpl<char const*>&, llvm::ToolContext const&, llvm::IntrusiveRefCntPtr<llvm::vfs::FileSystem>) driver.cpp:0:0
#47 0x0000000000dedb1d int llvm::function_ref<int (llvm::SmallVectorImpl<char const*>&)>::callback_fn<clang_main(int, char**, llvm::ToolContext const&)::'lambda'(llvm::SmallVectorImpl<char const*>&)>(long, llvm::SmallVectorImpl<char const*>&) driver.cpp:0:0
#48 0x00000000050c7189 void llvm::function_ref<void ()>::callback_fn<clang::driver::CC1Command::Execute(llvm::ArrayRef<std::optional<llvm::StringRef>>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>*, bool*) const::'lambda'()>(long) Job.cpp:0:0
#49 0x00000000044796a4 llvm::CrashRecoveryContext::RunSafely(llvm::function_ref<void ()>) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x44796a4)
#50 0x00000000050c77b6 clang::driver::CC1Command::Execute(llvm::ArrayRef<std::optional<llvm::StringRef>>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>*, bool*) const (.part.0) Job.cpp:0:0
#51 0x0000000005085e92 clang::driver::Compilation::ExecuteCommand(clang::driver::Command const&, clang::driver::Command const*&, bool) const (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x5085e92)
#52 0x0000000005086e3e clang::driver::Compilation::ExecuteJobs(clang::driver::JobList const&, llvm::SmallVectorImpl<std::pair<int, clang::driver::Command const*>>&, bool) const (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x5086e3e)
#53 0x000000000508eae5 clang::driver::Driver::ExecuteCompilation(clang::driver::Compilation&, llvm::SmallVectorImpl<std::pair<int, clang::driver::Command const*>>&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x508eae5)
#54 0x0000000000df33c4 clang_main(int, char**, llvm::ToolContext const&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0xdf33c4)
#55 0x0000000000caa03a main (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0xcaa03a)
#56 0x0000711fae229d90 (/lib/x86_64-linux-gnu/libc.so.6+0x29d90)
#57 0x0000711fae229e40 __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29e40)
#58 0x0000000000ded435 _start (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0xded435)
clang++: error: clang frontend command failed due to signal (use -v to see invocation)
Compiler returned: 134
```

swjng:
Guard `Sema::diagnoseArgDependentDiagnoseIfAttrs` with `isValueDependent()`, mirroring `Sema::CheckEnableIf` right above it. Without the guard, `Expr::EvaluateWithSubstitution` asserts when a non-template `diagnose_if` condition picks up a `RecoveryExpr` after the diagnostic engine enters a fatal-error state (e.g. once `-ferror-limit` is exceeded), because the condition becomes value-dependent.

`Sema::CheckEnableIf` already carries the same `isValueDependent()` guard with a FIXME for proper deferred handling; this just keeps the two in sync.

## Tool-use disclosure
This patch was developed with the assistance of an AI coding agent (Claude Code, Anthropic). All code, the commit message, and this PR description were reviewed and approved by the human author before submission.

Fixes #197625


TPPPP72:
I don't think you actually reviewed LLM's code. The problem might be that the Sema phase resolved the wrong node. Why did you consume all tokens after a fatal error occurred?

swjng:
Actually I evaluated narrower Sema-side fixes ((a) keep member lookup from collapsing a non-dependent base into a `CXXDependentScopeMemberExpr` in fatal-error state, (b) guard `Sema::diagnoseArgDependentDiagnoseIfAttrs` with `isValueDependent()`) and went with the parser-level skip thinking it centralized the invariant better than gating each consumer.

TPPPP72:
> Actually I evaluated narrower Sema-side fixes ((a) keep member lookup from collapsing a non-dependent base into a `CXXDependentScopeMemberExpr` in fatal-error state, (b) guard `Sema::diagnoseArgDependentDiagnoseIfAttrs` with `isValueDependent()`) and went with the parser-level skip thinking it centralized the invariant better than gating each consumer.

I think in any case, skiping all tokens is not a correct solution. Because Clang is not just for compile, but for intellisense. If you drop all tokens, it will cause unexpected problems.

swjng:
I missed that point. The parser-level change came from trying to address the FIXME on the existing `isValueDependent()` guard in `Sema::CheckEnableIf`. The PR now only contains the minimal change needed to fix the issue.

TPPPP72:
Well I think it's more acceptable than before. But I can't judge whether it's a correct fix right now(still need some time to investigate the root cause).

TPPPP72:
Instead of bypassing the assertion, we should consider why the incorrect attribute is being generated here.

So I think this is a correct fix:

```diff
diff --git a/clang/lib/Sema/SemaDeclAttr.cpp b/clang/lib/Sema/SemaDeclAttr.cpp
index 55b6cbcbba57..70f948fd77f9 100644
--- a/clang/lib/Sema/SemaDeclAttr.cpp
+++ b/clang/lib/Sema/SemaDeclAttr.cpp
@@ -7508,6 +7508,9 @@ ProcessDeclAttribute(Sema &S, Scope *scope, Decl *D, const ParsedAttr &AL,
   if (S.checkCommonAttributeFeatures(D, AL, MustDelayArgs))
     return;

+  if (S.getDiagnostics().hasFatalErrorOccurred() && AL.getNumArgs() > 0)
+    return;
+
   if (MustDelayArgs) {
     AL.handleAttrWithDelayedArgs(S, D);
     return;
```

When a fatal error occurs and there is an incorrect expression that might be evaluated, we should stop handling its Attr.

@AaronBallman CC

swjng:
@TPPPP72 I think that is more compact way for fixing root cause, do you think that changing the code is better?

TPPPP72:
> @TPPPP72 I think that is more compact way for fixing root cause, do you think that changing the code is better?

This is just my personal opinion, also need to consider the opinions of other reviewers.

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
```diff
@@ -0,0 +1,46 @@
...
+// After -ferror-limit is exceeded the parser enters a fatal-error state.
+// Late-parsed attribute conditions must not be attached to the declaration
+// in that state: name lookup of the condition's sub-expressions may fail
+// silently and wrap the result in a RecoveryExpr, producing a
+// value-dependent condition on a non-template function that later confuses
+// the diagnose_if consumer (Expr::EvaluateWithSubstitution asserts on
+// value-dependent expressions).
+// We test that we do not crash in such cases (#197625).
+
+#define _diagnose_if(...) __attribute__((diagnose_if(__VA_ARGS__)))
```
erichkeane: This whole test is way larger than is needed to reproduce this, and why are we doing this macro?

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
```diff
@@ -0,0 +1,46 @@
...
+};
+} // namespace std
+
+template <typename T>
+struct E {
+  void f(int i) _diagnose_if(i, "bad i", "error"); // expected-note 19 {{from 'diagnose_if' attribute on 'f'}}
+};
+void blast() {
+  E<int> e;
+  e.f(1); e.f(1); e.f(1); e.f(1); e.f(1); // expected-error 5 {{bad i}}
```
erichkeane: What in the world is going on here?

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
```diff
@@ -0,0 +1,46 @@
...
+struct initializer_list {
+  const T *p;
+  size_t s;
+  constexpr size_t size() const { return s; }
+};
+} // namespace std
+
+template <typename T>
+struct E {
+  void f(int i) _diagnose_if(i, "bad i", "error"); // expected-note 19 {{from 'diagnose_if' attribute on 'f'}}
```
erichkeane: Please put notes next to the error they are caused by.

clang/lib/Sema/SemaOverload.cpp
```diff
@@ -7801,8 +7801,12 @@ bool Sema::diagnoseArgDependentDiagnoseIfAttrs(const FunctionDecl *Function,
         // It's sane to use the same Args for any redecl of this function, since
         // EvaluateWithSubstitution only cares about the position of each
         // argument in the arg list, not the ParmVarDecl* it maps to.
-        if (!DIA->getCond()->EvaluateWithSubstitution(
-                Result, Context, cast<FunctionDecl>(DIA->getParent()), Args, ThisArg))
+        // FIXME: This doesn't consider value-dependent cases, because doing so
+        // is very difficult. Ideally, we should handle them more gracefully.
+        if (DIA->getCond()->isValueDependent() ||
```
erichkeane, line 7805: This seems reasonable here, skipping on value dependence is probably the right answer here.  I would like to see some tests to make sure we handle this correctly for later instantiation though.

erichkeane (commented):
This patch needs more human review, the AI you used clearly went off the rails, particularly with that test, and you didn't properly review it.  Please do so before you have us review this again.

AaronBallman:
> Instead of bypassing the assertion, we should consider why the incorrect attribute is being generated here.
> 
> So I think this is a correct fix:
> 
> ```diff
> diff --git a/clang/lib/Sema/SemaDeclAttr.cpp b/clang/lib/Sema/SemaDeclAttr.cpp
> index 55b6cbcbba57..70f948fd77f9 100644
> --- a/clang/lib/Sema/SemaDeclAttr.cpp
> +++ b/clang/lib/Sema/SemaDeclAttr.cpp
> @@ -7508,6 +7508,9 @@ ProcessDeclAttribute(Sema &S, Scope *scope, Decl *D, const ParsedAttr &AL,
>    if (S.checkCommonAttributeFeatures(D, AL, MustDelayArgs))
>      return;
> 
> +  if (S.getDiagnostics().hasFatalErrorOccurred() && AL.getNumArgs() > 0)
> +    return;
> +
>    if (MustDelayArgs) {
>      AL.handleAttrWithDelayedArgs(S, D);
>      return;
> ```
> 
> When a fatal error occurs and there is an incorrect expression that might be evaluated, we should stop handling its Attr.
> 
> @AaronBallman CC

FWIW, we should attempt recovery before resorting to heavier approaches like this. This is usually a sign that when we hit an error, we didn't ensure the AST was in a sufficiently reasonable state before continuing recovery.

clang/lib/Sema/SemaOverload.cpp
AaronBallman: It took me a bit of thinking, but I agree that this seems like a reasonable workaround for the time being. Eventually we'll need to solve the FIXME in both places though.

erichkeane: Perhaps I'm misunderstanding it here... but I don't think the fixme is actually valid.  We can't actually CHECK value-dependent cases here, this is a situation where we just have to decide that diagnose_if succeeded, and catch it during a future instantaition, right?

cor3ntin: So. The comment was copy pasted from enable_if. but the comment predates the check for value dependency added to enable_if in 04323c24a1ac9464471331d9f6499d3cb95d1ccd. ie we are spreading an incorrect comment through the magic of copypasta.

AaronBallman: > Perhaps I'm misunderstanding it here... but I don't think the fixme is actually valid. We can't actually CHECK value-dependent cases here, this is a situation where we just have to decide that diagnose_if succeeded, and catch it during a future instantaition, right?

That's what I meant by fix -- once we have the instantiation, it's no longer value dependent and we should catch it then. So this would do the usual two-step dance where we check the non-dependent bits as soon as we can, and then the dependent bits when rebuilding from tree transform.

TPPPP72:
> FWIW, we should attempt recovery before resorting to heavier approaches like this. This is usually a sign that when we hit an error, we didn't ensure the AST was in a sufficiently reasonable state before continuing recovery.

Oh yes. I thought current solution was acceptable but it looked like bypassing assertion, so I took a stab at it :D. 

> That's what I meant by fix -- once we have the instantiation, it's no longer value dependent and we should catch it then. So this would do the usual two-step dance where we check the non-dependent bits as soon as we can, and then the dependent bits when rebuilding from tree transform.

From this perspective, there's actually no problem with making this special case bypass here.

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
```diff
@@ -1,46 +1,19 @@
-// RUN: %clang_cc1 -fsyntax-only -std=c++17 -ferror-limit 19 -verify %s
+// RUN: %clang_cc1 -fsyntax-only -std=c++17 -ferror-limit 1 -verify %s
```
TPPPP72: The option `-std=c++17` is unnecessary.

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
```diff
@@ -0,0 +1,26 @@
...
+// Regression test for #197625: diagnose_if condition becomes value-dependent
+// via RecoveryExpr after fatal error, hitting an assertion in EvaluateWithSubstitution.
```
TPPPP72: I'm thinking do we need these comments?

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
swjng: Originally this test is written for parser fix before fixing Sema, so the limit was needed. I didn't fix this after changing it. reflected now.

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
swjng: This was to exceed `-ferror-limit 19`, now it is not needed, fixed now.

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
swjng: Reflected, thank you.

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
swjng: It doesn't have to be that long, but other tests have issue link in the file itself, so it is worth to have a link at least imo

TPPPP72:
I don't think you need to leave any comments in the test files except those required for testing. Because we can easily trace back using Git history.

swjng:
Okay, I'll follow that. Thanks.

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
```diff
@@ -0,0 +1,24 @@
+// RUN: %clang_cc1 -fsyntax-only -ferror-limit 1 -verify %s
+
+template <typename T> struct S {
+  T val;
+  constexpr T get() const { return val; }
+};
+
+// The isValueDependent() guard must not suppress diagnose_if after instantiation.
```
TPPPP72: This one

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
```diff
@@ -0,0 +1,24 @@
...
+struct Bar {
+  template <typename T>
+  Bar(S<T> s) __attribute__((diagnose_if(s.get() == 1, "one", "error"))) {} // expected-note {{from 'diagnose_if'}}
+};
+void instantiated() { Bar{S<int>{1}}; } // expected-error {{one}}
+
+// After fatal error, the guard must skip evaluation instead of crashing.
```
TPPPP72: And this one

TPPPP72 (commented):
LGTM

cor3ntin (commented):
This change needs a release note.
Please add an entry to `clang/docs/ReleaseNotes.rst` in the section the most adapted to the change, and referencing any Github issue this change fixes. Thanks!

swjng:
Added a release note under "Bug Fixes to Attribute Support". Let me know if a different section would be more appropriate. Thanks!

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+void bad(int i) __attribute__((diagnose_if(i, "bad", "error")));
+void blast() {
+  bad(1); // expected-error@* {{too many errors}}
+}
+
+struct Foo {
+  Foo(S<int> s) __attribute__((diagnose_if(s.get() == 1, "x", "warning"))) {}
+};
+void run() { Foo{S<int>{1}}; }
```
AaronBallman: This should diagnose with a warning: https://godbolt.org/z/q4aTzzc9W

AaronBallman: In fact, none of the added test cases trigger the failed assertion this claims to fix: https://godbolt.org/z/Pn3T5Gbcr

Please be sure to test what's in the issue.

AaronBallman (changes requested):

clang/test/SemaCXX/diagnose-if-fatal-error.cpp
swjng: Thank you for the review. Now the test code is showing crash with flag ` -fsyntax-only -ferror-limit=1` without fix. 
ref: https://godbolt.org/z/nYva3xxEW

clang/lib/Sema/SemaOverload.cpp
AaronBallman: I was trying to figure out whether type dependence matters here or not, so I looked at this a bit more.

We're already doing that two-step dance here. e.g., we don't get into `Sema::diagnoseArgDependentDiagnoseIfAttrs()` for the primary template, only when instantiating it.

The issue boils down to use of `-ferror-limit`; that results in `TreeTransform` having nodes which *remain* dependent because of the error dependency bit. IIRC that's theoretically fine because we can still end up with dependencies in non-error cases (like partial specializations), but I wonder if we're going to have other similar issues with other constructs -- we don't test with low error limits in general. That said, it's not clear to me whether this change is correct -- should we be looking for error dependence specifically? Or can we devise a test case that fails without error dependence?

CC @erichkeane @mizvekov for opinions.

erichkeane: What do you mean by `error dependency bit`?  Do you mean the `RecoveryExpr` 'has error` thing?  If so, I don't think that really carries any problems other than perhaps some awkward diagnostics later.  We could perhaps have `hasError` checked here and ALSO return `false`, which would be completely sensible, though the result is 'we choose one of the behaviors here on error'.

mizvekov: As far as I remember from what I have seen, the `-ferror-limit` stuff just suppresses further diagnostics, but it doesn't change anything about error recovery.

We do alter error recovery behavior based out of being in an SFINAE context though.

PR #197648, opened 2026-05-14
AMDGPU/GlobalISel: Legalize scalar extloads with large memory type

petar-avramovic:
Add narrowScalar for scalar sext/zextload when the memory type is
larger then 32 bits. There is no narrow scalar implementation when
NarrowSize < MemSize (split load) but we don't want that anyway.
Narrow scalar to MemSize creates large normal load + extension to dst.

### DIFF
--- a/llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
+++ b/llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
@@ -1750,6 +1750,13 @@ AMDGPULegalizerInfo::AMDGPULegalizerInfo(const GCNSubtarget &ST_,
   // inserting addrspacecasts.
   ExtLoads.customIf(typeIs(1, Constant32Ptr));
 
+  ExtLoads.narrowScalarIf(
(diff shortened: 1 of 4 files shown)


petar-avramovic:
* **#197648** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

arsenm (commented):
I want to stop abusing the narrow/widen legalization actions for changing the memory type, and only use them for register values. These should only apply to the register value, and in-memory type changes should be left to lower or some new action

petar-avramovic:
But I do this no avoid changing MemType

arsenm:
> But I do this no avoid changing MemType

I mean this patch is fine for now, but this isn't really how memory legalization should work

arsenm (approved):

PR #197653, opened 2026-05-14
[FIRToMemRef] Fix fir.convert insertion inside omp.wsloop

ergawy:
When replaceFIRMemrefs inserted a fir.convert before an op inside a LoopWrapperInterface region (e.g. omp.simd inside omp.wsloop), it violated the single-nested-op invariant, producing a verifier error. Fix by walking up the LoopWrapperInterface parent chain and inserting before the outermost wrapper instead.

Co-authored-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/flang/lib/Optimizer/Transforms/FIRToMemRef.cpp
+++ b/flang/lib/Optimizer/Transforms/FIRToMemRef.cpp
@@ -1153,7 +1153,13 @@ void FIRToMemRef::replaceFIRMemrefs(Value firMemref, Value converted,
   Type ty = firMemref.getType();
 
   for (auto op : worklist) {
-    rewriter.setInsertionPoint(op);
+    // If op is directly inside a LoopWrapperInterface region, inserting before
+    // op would violate the single-nested-op invariant. Walk up the wrapper
+    // chain and insert before the outermost wrapper instead.
+    Operation *insertBefore = op;
+    while (mlir::isa<omp::LoopWrapperInterface>(insertBefore->getParentOp()))
+      insertBefore = insertBefore->getParentOp();
+    rewriter.setInsertionPoint(insertBefore);
     Location loc = op->getLoc();
     Value replaceConvert = fir::ConvertOp::create(rewriter, loc, ty, converted);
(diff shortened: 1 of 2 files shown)


flang/lib/Optimizer/Transforms/FIRToMemRef.cpp
```diff
@@ -1153,7 +1153,13 @@ void FIRToMemRef::replaceFIRMemrefs(Value firMemref, Value converted,
   Type ty = firMemref.getType();
 
   for (auto op : worklist) {
-    rewriter.setInsertionPoint(op);
+    // If op is directly inside a LoopWrapperInterface region, inserting before
+    // op would violate the single-nested-op invariant. Walk up the wrapper
+    // chain and insert before the outermost wrapper instead.
+    Operation *insertBefore = op;
+    while (mlir::isa<omp::LoopWrapperInterface>(insertBefore->getParentOp()))
```
ergawy, line 1160: I would have preferred a more general and non-omp-specific solution but there is no dialect agnostic interface for this.

We can instead check for the structure of the ops like:
```c++
    Operation *insertBefore = op;
    while (llvm::hasSingleElement(llvm::make_filter_range(
               *insertBefore->getBlock(), [](Operation &o) {
                 return !o.hasTrait<mlir::OpTrait::IsTerminator>();
               })))
      insertBefore = insertBefore->getParentOp();
    rewriter.setInsertionPoint(insertBefore);
```
Same effect but the interface captures the intention much better of course.

Let me know what you think.

ergawy:
Ping! Please take a look when you have time.

flang/lib/Optimizer/Transforms/FIRToMemRef.cpp
SusanTan: Hi Kareem, I think what you have is sufficient unless we can have `LoopWrapperInterface` not tightly nested, for example:
```
omp.wsloop {        
  some.wrapper {    
    omp.simd {      
      omp.loop_nest   
```
Otherwise, i think what you have is good!

SusanTan (approved):
thank you Kareem!

PR #168903, opened 2025-11-20
[NFC][TTI] Introduce getInstructionUniformity API for uniformity analysis

PankajDwivedi-25:
This patch introduces a new TargetTransformInfo hook `getInstructionUniformity()`
that provides a unified interface for querying target-specific uniformity
information about instructions and values.

The new hook returns an `InstructionUniformity` enum with three values:
- Default: Result is uniform if all operands are uniform (standard propagation)
- AlwaysUniform: Result is always uniform regardless of operands
- NeverUniform: Result can never be assumed uniform

This API wraps the existing `isAlwaysUniform()` and `isSourceOfDivergence()`
hooks, providing a single entry point for uniformity queries. Both LLVM IR-level
(via TTI) and MIR-level (via TargetInstrInfo) uniformity analysis have been
updated to use the new hook.

Target implementations:
- AMDGPU: Wraps existing `isAlwaysUniform()` and `isSourceOfDivergence()` hooks
- NVPTX: Wraps existing `isSourceOfDivergence()` hook

This is an NFC change - all implementations return conservative defaults or wrap
existing functionality.

Ref patch:https://github.com/llvm/llvm-project/pull/137639

PR #197656, opened 2026-05-14
[UniformityAnalysis] Fix nodivergencesource calls

piotrAMD:
NFC #168903 introduced a subtle behavior change for calls with the
nodivergencesource attribute and divergent operands.

Calls with the nodivergencesource attribute are *not* always uniform.
They just do not introduce any new divergence. If any operand is divergent,
the result must still be reported as divergent.

Revert to pre-#168903 behavior by allowing the standard propagation to
work for target's NeverUniform while keeping Default and AlwaysUniform
unchanged.


llvm/lib/Analysis/TargetTransformInfo.cpp
```diff
@@ -294,12 +294,13 @@ bool TargetTransformInfo::hasBranchDivergence(const Function *F) const {
 
 ValueUniformity
 llvm::TargetTransformInfo::getValueUniformity(const Value *V) const {
-  // Calls with the NoDivergenceSource attribute are always uniform.
+  ValueUniformity VU = TTIImpl->getValueUniformity(V);
   if (const auto *Call = dyn_cast<CallBase>(V)) {
-    if (Call->hasFnAttr(Attribute::NoDivergenceSource))
-      return ValueUniformity::AlwaysUniform;
+    if (Call->hasFnAttr(Attribute::NoDivergenceSource) &&
+        VU == ValueUniformity::NeverUniform)
```
arsenm:
```suggestion
    if (VU == ValueUniformity::NeverUniform && Call->hasFnAttr(Attribute::NoDivergenceSource))
```

arsenm (approved):

jayfoad:
> NFC #168903 introduced a subtle behavior change for calls with the
> nodivergencesource attribute and divergent operands.

"Supposedly NFC". Nice catch. I should have caught that in review.

jayfoad (approved):

llvm/test/Analysis/UniformityAnalysis/AMDGPU/nodivergencesource-divergent-arg.ll
```diff
@@ -0,0 +1,18 @@
+; RUN: opt -mtriple amdgcn-- -passes='print<uniformity>' -disable-output %s 2>&1 | FileCheck %s
```
jayfoad: This test could go in the existing `nodivergencesource.ll`

PankajDwivedi-25 (approved):
Thanks for catching this, completely missed it.

PR #197657, opened 2026-05-14
[AArch64][SVE] Match (add_like x (lsr y, c)) -> usra x, y, c

Harry-Ramsey:
Modify SVEE USRA pattern to accept add_like, so both add and or disjoint forms can select usra.

Add known-bits support for predicated SVE logical shifts, allowing or_disjoint matching to prove disjointness for plain ORs where possible.

### DIFF
--- a/llvm/lib/Target/AArch64/AArch64SVEInstrInfo.td
+++ b/llvm/lib/Target/AArch64/AArch64SVEInstrInfo.td
@@ -362,11 +362,11 @@ def AArch64uaba : PatFrags<(ops node:$op1, node:$op2, node:$op3),
 
 def AArch64usra : PatFrags<(ops node:$op1, node:$op2, node:$op3),
                            [(int_aarch64_sve_usra node:$op1, node:$op2, node:$op3),
-                            (add node:$op1, (AArch64lsr_p (SVEAnyPredicate), node:$op2, (SVEShiftSplatImmR (i32 node:$op3))))]>;
+                            (add_like node:$op1, (AArch64lsr_p (SVEAnyPredicate), node:$op2, (SVEShiftSplatImmR (i32 node:$op3))))]>;
 
 def AArch64ssra : PatFrags<(ops node:$op1, node:$op2, node:$op3),
                            [(int_aarch64_sve_ssra node:$op1, node:$op2, node:$op3),
-                            (add node:$op1, (AArch64asr_p (SVEAnyPredicate), node:$op2, (SVEShiftSplatImmR (i32 node:$op3))))]>;
+                            (add_like node:$op1, (AArch64asr_p (SVEAnyPredicate), node:$op2, (SVEShiftSplatImmR (i32 node:$op3))))]>;
 
 def AArch64ursra : PatFrags<(ops node:$op1, node:$op2, node:$op3),
                             [(int_aarch64_sve_ursra node:$op1, node:$op2, node:$op3),
--- a/llvm/test/CodeGen/AArch64/sve2-sra.ll
+++ b/llvm/test/CodeGen/AArch64/sve2-sra.ll
@@ -255,6 +255,86 @@ define <vscale x 2 x i64> @ssra_intr_u_i64(<vscale x 2 x i1> %pg, <vscale x 2 x 
   ret <vscale x 2 x i64> %add
 }
 
+define <vscale x 16 x i8> @usra_disjoint_or16xi8(<vscale x 16 x i8> %a, <vscale x 16 x i8> %b) #0 {
+; CHECK-LABEL: usra_disjoint_or16xi8:
+; CHECK:       // %bb.0:
+; CHECK-NEXT:    usra z0.b, z1.b, #4
+; CHECK-NEXT:    ret
+  %shift = lshr <vscale x 16 x i8> %b, splat(i8 4)
+  %add = or disjoint <vscale x 16 x i8> %a, %shift
+  ret <vscale x 16 x i8> %add
+}
+
+define <vscale x 8 x i16> @usra_disjoint_or8xi16(<vscale x 8 x i16> %a, <vscale x 8 x i16> %b) #0 {
+; CHECK-LABEL: usra_disjoint_or8xi16:
+; CHECK:       // %bb.0:
+; CHECK-NEXT:    usra z0.h, z1.h, #4
+; CHECK-NEXT:    ret
+  %shift = lshr <vscale x 8 x i16> %b, splat(i16 4)
+  %add = or disjoint <vscale x 8 x i16> %a, %shift
+  ret <vscale x 8 x i16> %add
+}
+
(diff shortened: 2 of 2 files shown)


llvm/lib/CodeGen/SelectionDAG/SelectionDAG.cpp
```diff
@@ -4525,11 +4525,6 @@ KnownBits SelectionDAG::computeKnownBits(SDValue Op, const APInt &DemandedElts,
   case ISD::INTRINSIC_WO_CHAIN:
   case ISD::INTRINSIC_W_CHAIN:
   case ISD::INTRINSIC_VOID:
-    // TODO: Probably okay to remove after audit; here to reduce change size
```
Harry-Ramsey, line 4530: Please let me know if the audit concluded and this code can be removed?

llvm/lib/Target/AArch64/AArch64SVEInstrInfo.td
```diff
@@ -362,7 +362,7 @@ def AArch64uaba : PatFrags<(ops node:$op1, node:$op2, node:$op3),
 
 def AArch64usra : PatFrags<(ops node:$op1, node:$op2, node:$op3),
                            [(int_aarch64_sve_usra node:$op1, node:$op2, node:$op3),
-                            (add node:$op1, (AArch64lsr_p (SVEAnyPredicate), node:$op2, (SVEShiftSplatImmR (i32 node:$op3))))]>;
+                            (add_like node:$op1, (AArch64lsr_p (SVEAnyPredicate), node:$op2, (SVEShiftSplatImmR (i32 node:$op3))))]>;
 
 def AArch64ssra : PatFrags<(ops node:$op1, node:$op2, node:$op3),
```
davemgreen, line 367: Add ssra too, whilst you are here?

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -2902,6 +2902,18 @@ void AArch64TargetLowering::computeKnownBitsForTargetNode(
     }
     break;
   }
+  case AArch64ISD::SHL_PRED:
+  case AArch64ISD::SRL_PRED: {
+    KnownBits KnownVal =
```
davemgreen: This should probably check for all-ones mask. Otherwise the off lanes are unknown.

llvm/lib/CodeGen/SelectionDAG/SelectionDAG.cpp
davemgreen: I think it looks OK, looking at the existing nodes in architectures that have scalable vectors and how they use types.

llvm/test/CodeGen/AArch64/sve2-sli-sri.ll
```diff
@@ -118,14 +118,22 @@ define <vscale x 8 x i16> @testRightGood8x16(<vscale x 8 x i16> %src1, <vscale x
 }
 
 define <vscale x 8 x i16> @testRightBad8x16(<vscale x 8 x i16> %src1, <vscale x 8 x i16> %src2) {
```
davemgreen, line 120: It would be good to see more deliberate test cases, and some with disjoint ors: https://godbolt.org/z/7de9cMjG9

Harry-Ramsey:
I have taken a moment to additionally create tests for SSRA. Let me know if the test names should be renamed.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -2929,6 +2929,21 @@ void AArch64TargetLowering::computeKnownBitsForTargetNode(
...
+  case AArch64ISD::SRL_PRED: {
+    if (isAllActivePredicate(DAG, Op->getOperand(0)))
+      break;
+
+    KnownBits KnownVal =
+        DAG.computeKnownBits(Op->getOperand(1), DemandedElts, Depth + 1);
+    KnownBits KnownAmt =
+        DAG.computeKnownBits(Op->getOperand(2), DemandedElts, Depth + 1);
+
+    Known = Op.getOpcode() == AArch64ISD::SHL_PRED
```
davemgreen: From what I can tell this isn't tested any more, or isn't triggering where it was before? The add_like would be enough to trigger the or disjoint cases IIUC. It might be better to separate this known-bits part out so that it can be tested on it's own.

llvm/test/CodeGen/AArch64/sve2-sli-sri.ll
```diff
@@ -52,6 +52,38 @@ define <vscale x 16 x i8> @testRightGood16x8(<vscale x 16 x i8> %src1, <vscale x
   ret <vscale x 16 x i8> %result
 }
 
+define <vscale x 16 x i8> @usra_disjoint_or16xi8(<vscale x 16 x i8> %a, <vscale x 16 x i8> %b) {
```
davemgreen: Can you separate these out into a new test file for usra+ssra or add it to an existing file like llvm/test/CodeGen/AArch64/sve2-sra.ll

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
Harry-Ramsey: Sorry, just a bit confused by what is meant here. The usra/ssra tests are still going down this code path with SVE2 so this should be tested?

llvm/test/CodeGen/AArch64/sve2-sra.ll
```diff
@@ -1,5 +1,5 @@
 ; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
-; RUN: llc < %s | FileCheck %s
+; RUN: llc -mattr=+sve2 < %s | FileCheck %s
```
davemgreen: To fit in with the other tests here, the #0 attributes can be added to the new functions to pick up +sve2.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
davemgreen: The `add_like` tablegen change you have added will pick up "add like" from it being an `add` or a `or_disjoint`. `or_disjoint` in turn can either be an `or` with the `disjoint` flag or where we look at the known-bits for the two inputs having `haveNoCommonBitsSet`.

This bit is helping with the second known-bits check, but the new tests are going through the first path I believe, with a `disjoint or` flag. And as no other tests are changing, they are likely not tested any more or not working in some way. Some of the sri tests used to trigger IIRC.

The add_like change alone is enough to fix the issue, so if we can separate this known-bits part out and make sure it is tested separately, we can get the other part in and each half is simpler.

davemgreen: Oh - When I say different patch I should say different PR, so they can be reviewed and submitted separately.

davemgreen (approved):
LGTM - thanks.

PR #197658, opened 2026-05-14
[lldb] fix single-character token not underlined correctly in diagnostics

charles-zablit:
A single-character token at the end of a user expression was rendered with trailing underlines:

# Result 
```
(lldb) p foo!
          ^~~~~~
          error: ...
```

# Expected:

```
(lldb) p foo!
          ^
          error: ...
```

There were 2 issues:

1. Token ranges store their end as the start of the last token: a single token range produces `end_col == loc.column`, giving a length of 0.

2. A token at the very end of the user expression can cause Clang's range to end inside the wrapper suffix. `getSpellingColumnNumber` on a location on a different line returns that line's column. It can be larger than `loc.column`, producing an incorret range. The fix here is to discard any range whose end is on a different line than the diagnostic's location.

rdar://176913198

### DIFF
--- a/lldb/source/Plugins/ExpressionParser/Clang/ClangExpressionParser.cpp
+++ b/lldb/source/Plugins/ExpressionParser/Clang/ClangExpressionParser.cpp
@@ -280,11 +280,9 @@ public:
           // Find the range of the primary location.
           for (const auto &range : Info.getRanges()) {
             if (range.getBegin() != sloc)
-              continue
-            SourceLocation end = range.getEnd();
+              continue SourceLocation end = range.getEnd();
             if (range.isTokenRange())
-              end =
-                  clang::Lexer::getLocForEndOfToken(end, 0, sm, m_lang_opts);
+              end = clang::Lexer::getLocForEndOfToken(end, 0, sm, m_lang_opts);
             // FIXME: This is probably not handling wide characters correctly.
             unsigned end_col = sm.getSpellingColumnNumber(end);
             // Ignore ranges that span multiple lines.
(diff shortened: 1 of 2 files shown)


lldb/source/Plugins/ExpressionParser/Clang/ClangExpressionParser.cpp
```diff
@@ -279,8 +280,16 @@ class ClangDiagnosticManagerAdapter : public clang::DiagnosticConsumer {
...
+              if (range.isTokenRange())
+                end =
+                    clang::Lexer::getLocForEndOfToken(end, 0, sm, m_lang_opts);
+              // Ignore ranges that span multiple lines.
+              if (sm.getSpellingLineNumber(end) !=
+                  sm.getSpellingLineNumber(sloc))
+                break;
               // FIXME: This is probably not handling wide characters correctly.
-              unsigned end_col = sm.getSpellingColumnNumber(range.getEnd());
+              unsigned end_col = sm.getSpellingColumnNumber(end);
```
bulbazord: Suggestion: Move the multiple-line check after this to save a call to `getSpellingColumnNumber`.
```
unsigned end_col = sm.getSpellingColumnNumber(end);
if (end_col != sm.getSpellingLineNumber(sloc))
  break;
```

Not sure how valuable this is, but it's a little less repetitive.

bulbazord (approved):

lldb/source/Plugins/ExpressionParser/Clang/ClangExpressionParser.cpp
charles-zablit: This is better, thanks!

PR #197660, opened 2026-05-14
[lldb][windows] Add --lldb-use-lldb-server to lit

charles-zablit:
Currently, the Windows builds of lit only allow `SystemDrive` when copying env vars into the test environment. `LLDB_USE_LLDB_SERVER=1` in the developer shell never reached the `lldb.exe` subprocess under `check-lldb`. Baseline and lldb-server runs therefore looked identical even when the user expected lldb-server to be used.

`--lldb-use-lldb-server` in lit invocations sets `LLDB_USE_LLDB_SERVER=1` to test lldb on Windows with `lldb-server` instead of the in process plugin.

Depends on:
- https://github.com/llvm/llvm-zorg/pull/848

rdar://177442264

### DIFF
--- a/lldb/test/API/lit.cfg.py
+++ b/lldb/test/API/lit.cfg.py
@@ -361,9 +361,7 @@ if platform.system() == "Windows":
     use_server = lit_config.params.get("lldb-use-lldb-server", None)
     if use_server is None:
         use_server = getattr(config, "lldb_use_lldb_server", None)
-    if use_server is not None and str(use_server).lower() in (
-        "1", "on", "yes", "true"
-    ):
+    if use_server is not None and str(use_server).lower() in ("1", "on", "yes", "true"):
         lit_config.note("Running API tests with LLDB_USE_LLDB_SERVER=1")
         config.environment["LLDB_USE_LLDB_SERVER"] = "1"
 
--- a/lldb/test/API/lit.cfg.py
+++ b/lldb/test/API/lit.cfg.py
@@ -354,9 +354,19 @@ if "XDG_CACHE_HOME" in os.environ:
 
 # Transfer some environment variables into the tests on Windows build host.
 if platform.system() == "Windows":
-    for v in ["SystemDrive", "LLDB_USE_LLDB_SERVER"]:
+    for v in ["SystemDrive"]:
         if v in os.environ:
             config.environment[v] = os.environ[v]
+
+    use_server = lit_config.params.get("lldb-use-lldb-server", None)
+    if use_server is None:
+        use_server = getattr(config, "lldb_use_lldb_server", None)
+    if use_server is not None and str(use_server).lower() in (
+        "1", "on", "yes", "true"
+    ):
+        lit_config.note("Running API tests with LLDB_USE_LLDB_SERVER=1")
+        config.environment["LLDB_USE_LLDB_SERVER"] = "1"
+
     # Use anonymous pipes instead of ConPTY for all tests. ConPTY injects VT
     # escape sequences into the output stream, which breaks tests that check
     # for specific stdout/stderr content.
--- a/lldb/test/API/lit.cfg.py
+++ b/lldb/test/API/lit.cfg.py
@@ -358,11 +358,7 @@ if platform.system() == "Windows":
         if v in os.environ:
             config.environment[v] = os.environ[v]
 
-    use_server = lit_config.params.get("lldb-use-lldb-server", None)
-    if use_server is None:
-        use_server = getattr(config, "lldb_use_lldb_server", None)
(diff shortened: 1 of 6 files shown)


Nerixyz (approved):
Do the buildbots use this environment variable? If so, we'll probably get failures when we merge this.

DavidSpickett:
One bot uses this - https://github.com/llvm/llvm-zorg/blob/84ea00bff4f1e212f3947dc3eb35585e3d5b1b4a/buildbot/osuosl/master/config/builders.py#L1368

Or rather thinks it's using it, if I understand correctly, it has been having no effect until now.

DavidSpickett:
What is the state of lldb-server on Windows? I thought it didn't work at all.

charles-zablit:
> One bot uses this - [llvm/llvm-zorg@`84ea00b`/buildbot/osuosl/master/config/builders.py#L1368](https://github.com/llvm/llvm-zorg/blob/84ea00bff4f1e212f3947dc3eb35585e3d5b1b4a/buildbot/osuosl/master/config/builders.py#L1368)
> 
> Or rather thinks it's using it, if I understand correctly, it has been having no effect until now.

That's an issue. As @Nerixyz said we will get test failures once we merge this. Given that the `LLDB_USE_LLDB_SERVER=1` path was never tested in the first place, I think it's reasonable to remove this env variable from the build bot config.

> What is the state of lldb-server on Windows? I thought it didn't work at all.

It does not. I'm looking at getting it working and so far I found multiple low hanging fruits. I opened 2 PRs with more to come:
- https://github.com/llvm/llvm-project/pull/197663
- https://github.com/llvm/llvm-project/pull/197669

Once I understand the state of `lldb-server` on Windows better, I would like to propose an RFC to (eventually) make it the default on Windows.

DavidSpickett:
> I think it's reasonable to remove this env variable from the build bot config.

Agreed. You can send a PR to llvm-zorg for that, though it will take a few days to be applied to the server once merged.

DavidSpickett:
> It does not. I'm looking at getting it working and so far I found multiple low hanging fruits.

Nice. I wasn't sure it would even build an executable.

DavidSpickett (approved):

charles-zablit:
Thanks, will merge when https://github.com/llvm/llvm-zorg/pull/848 lands.

JDevlieghere (changes requested):
This seems like the wrong way to hook this up. We shouldn't be setting environment variables to change the behavior of the test. This makes it hard (to impossible) to reproduce a failure locally. This should be passed as a lit option (either through the extra args) or potentially configured by CMake and a variable in the lit site config.

charles-zablit:
> This seems like the wrong way to hook this up. We shouldn't be setting environment variables to change the behavior of the test. This makes it hard (to impossible) to reproduce a failure locally. This should be passed as a lit option (either through the extra args) or potentially configured by CMake and a variable in the lit site config.

I agree, I fixed this as you suggested: creating a lit flag and a CMake variable.

lldb/test/Shell/lit.cfg.py
```diff
@@ -54,6 +54,7 @@
         "TEMP",
         "TMP",
         "XDG_CACHE_HOME",
+        "LLDB_USE_LLDB_SERVER",
```
JDevlieghere:
```suggestion
```

lldb/test/API/lit.cfg.py
```diff
@@ -357,6 +357,14 @@ def delete_module_cache(path):
     for v in ["SystemDrive"]:
         if v in os.environ:
             config.environment[v] = os.environ[v]
+
+    use_server = lit_config.params.get("lldb-use-lldb-server", None)
+    if use_server is None:
+        use_server = getattr(config, "lldb_use_lldb_server", None)
+    if use_server is not None and str(use_server).lower() in ("1", "on", "yes", "true"):
```
JDevlieghere: This is not necessary if you pass `LLDB_TEST_USE_LLDB_SERVER` through `llvm_canonicalize_cmake_booleans` in `lldb/test/CMakeLists.txt`.

lldb/test/Shell/lit.cfg.py
JDevlieghere: Should be handled the same way here as for the API tests. We should configure the option and then add that to the environment.

lldb/test/API/lit.cfg.py
charles-zablit: I did not know about this, thanks!

lldb/test/Shell/lit.cfg.py
charles-zablit: Fixed, with the llvm_canonicalize_cmake_booleans fix.

JDevlieghere (approved):
LGTM. Thanks for bearing with me! 🐻

PR #193884, opened 2026-04-24
llvm: honor VFABI mappings without builtin suppression

wpcwzy:
This is PR 2/2 fixing issue #120868 ,where LLVM fails to utilize SIMD counterparts of standard math functions (such as acos, sin etc.) even when they are properly annotated with "#pragma omp declare simd" in system headers like glibc.

Changes contained in this PR are listed below:

- For LAA, LV & VFABIDemangler: Permits vectorizing calls that possess no-builtin so long as "VFDatabase::getMappings(*Call)" is not empty. Modifies VFABIDemangler to query the “Function” for “vector-function-abi-variant” attributes when the “CallInst” misses them.
- For transform warnings: Stops emitting FailedRequestedVectorization remarks for OpenMP annotated parallel loops, avoiding false positives since unvectorized execution is a legal fallback. (glilbc uses -Werror option so this false postive could cause a failure in compiling.)

Test cases for the suppresed warnings and the no-builtin variant mapping extraction are added.

Before submitting this PR, the LLVM unit tests, regression tests, llvm-test-suite are ran without failing. Moreover, compiling glibc with the patch series enabled successfully emitted call instructions to glibc's libmvec functions, solving issue #120868 .

Link to v2 patch 1/3: #193881 
Link to v2 patch 3/3: #203465 
Link to v1 patch: #189523

PR #178396, opened 2026-01-28
Add new llvm.cbrt intrinsic

david-arm:
In CodeGen we already have support for the FCBRT DAG node that gets lowered to a library call on targets where the libcall cbrt is available. This node gets created during a DAG combine that converts

  fpow(float x, float 1.f/3.f) -> fcbrt(float x)

and

  fpow(double x, double 1./3.) -> fcbrt(double x)

when the appropriate fast-math flags are set on the fpow.

I would like to add an equivalent LLVM intrinsic for cbrt for two reasons:

1. The cbrt math call seems to be a commonly available math call on most systems. Adding a cbrt intrinsic would be consistent with similar intrinsics such as pow, sqrt, etc. There is currently no builtin support for cbrt in clang, but this can be added in a later patch.
2. It permits the use of a similar IR level libcall simplification:

  llvm.pow.f32(float x, float 1.0f/3.0f) -> llvm.cbrt.f32(float x)

There is already a precedence for this in Transforms/Utils/SimplifyLibCalls.cpp where we currently simplify

  llvm.pow.f32(float x, float 0.25) -> llvm.sqrt.f32(float x)

provided the intrinsic has the appropriate flags. There is already an existing test in test/Transforms/InstCombine/pow-cbrt.ll and presumably this was added with the intention of adding such a transformation. Doing this unlocks performance gains on targets that have vector math library support for cbrt, since both the loop and SLP vectorisers are able to take advantage of them.

Issue #164642, opened 2025-10-22
clang++ miscompiles std::pow with mixed types when using -O3 -fno-math-errno -fveclib=libmvec


davidbatetc:
I am using the `Linux x64_64` release of [LLVM 21.1.4](https://github.com/llvm/llvm-project/releases/tag/llvmorg-21.1.4).

Reduced test case (also in [godbolt](https://godbolt.org/z/bWbe4Y54e)):
`test.cpp`
```cpp
#include <cmath>
#include <iostream>

using T = double;
using U = float;

void __attribute__((noinline)) computePow(T *dst, T *base, U *exponent, int n)
{
	for (int i = 0; i < n; ++i) {
		dst[i] = static_cast<T>(std::pow(base[i], exponent[i]));
	}
}

int main()
{
	constexpr int N = 4;

	T x[N] = {2, 4, 6, 8};
	U y[N] = {7, 5, 3, 1};

	T z[N];
	computePow(z, x, y, N);

	for (int i = 0; i < N; ++i) {
		std::cout << "pow(" << x[i] << ", " << y[i] << ") = " << z[i]
				  << std::endl;
	}
}
```
```text
$ clang++ test.cpp -o test -O3 -fveclib=libmvec -fno-math-errno
./test
pow(2, 7) = 64
pow(4, 5) = 65536
pow(6, 3) = 0
pow(8, 1) = 0
```

However, the result should be
```text
pow(2, 7) = 128
pow(4, 5) = 1024
pow(6, 3) = 216
pow(8, 1) = 8
```

The same wrong result happens for other combinations of types for `T` and `U`, where `T != U`, for example `T = double` and `U = int`. The result is instead correct for `T = U = double` and `T = U = float`.

---

From what I have investigated, it seems like the issue is that the compiled program uses the [libmvec function](https://sourceware.org/glibc/wiki/libmvec?action=AttachFile&do=view&target=VectorABI.txt) `_ZGVdN4vv_pow` for computing 4 powers at once, but the values are not placed properly in the registers.

In the assembly generated (see godbolt), the registers xmm0 and xmm1 are used for the base and the registers xmm2 and xmm3 for the exponent. I think that ymm0 should instead be used for the base and ymm1 for the exponent.

PR #197661, opened 2026-05-14
[LV] Add vector library call legalization via VPlan transform

nema-ashutosh:
Vector library call legalization via VPlan transform:

RFC Link: https://discourse.llvm.org/t/rfc-lv-vector-library-call-legalization-splitting-wide-vector-calls-into-library-supported-widths/90786

When the loop vectorizer picks a VF for which no usable vector-library variant exists -- either because the call exceeds the target's register width, or because the library has no variant at that VF -- the vectorizer today either emits an unsupported call (forcing backend scalarization) or falls back to scalar calls outright.

This patch introduces a VPlan-to-VPlan transformation that legalizes such calls by splitting them into multiple calls at a width the library actually provides. For example, sin(double) at VF=8 on AVX (256-bit) is split into 2 x VF=4 calls to __svml_sin4 (or amd_vrd4_sin, etc.).

The transformation is gated behind a hidden flag
-legalize-vector-library-calls (default off).

The cost model finds the largest TLI-known variant that fits in registers and evenly divides the requested VF, and estimates the split cost (N calls + extract/insert shuffles). Legalization is chosen only when it beats all alternatives.

A new VPlan pass, legalizeVecLibCalls(), expands marked recipes into ExtractSubvector + legal-width call + ConcatVectors sequences before IR generation.

This works with any vector library that has TLI mappings (SVML, AMDLIBM, LIBMVEC, etc.) and supports multi-argument functions like pow().

### DIFF
--- a/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
+++ b/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
@@ -316,6 +316,11 @@ static cl::opt<bool> PreferPredicatedReductionSelect(
     cl::desc(
         "Prefer predicating a reduction operation over an after loop select."));
 
+static cl::opt<bool> LegalizeVectorLibraryCalls(
+    "legalize-vector-library-calls", cl::init(false), cl::Hidden,
+    cl::desc("Legalize vector library calls by splitting oversized vector "
+             "calls into multiple calls with a legal vector factor"));
+
 cl::opt<bool> llvm::EnableVPlanNativePath(
     "enable-vplan-native-path", cl::Hidden,
     cl::desc("Enable VPlan-native vectorization path with "
@@ -854,6 +859,26 @@ public:
   /// decision in a map for use in planning and plan execution.
   void setVectorizedCallDecision(ElementCount VF);
 
+  /// Returns true if \p VF elements of the given scalar type exceed the
+  /// target's vector register width.
+  bool exceedsRegisterWidth(ElementCount VF, Type *ScalarRetTy,
+                            ArrayRef<Type *> ScalarParamTys) const;
+
+  /// Find the largest TLI-known vector variant of \p CI's callee whose VF
+  /// evenly divides \p VF. Returns {nullptr, 0} if none found.
+  std::pair<Function *, ElementCount>
+  findLargestLegalVariant(CallInst *CI, ElementCount VF, Type *ScalarRetTy,
+                          ArrayRef<Type *> ScalarParamTys) const;
+
+  /// Compute the cost of splitting a vector call at \p VF into multiple
+  /// legal-width calls. Populates \p LegalVecFunc and \p LegalVecVF.
+  InstructionCost getLegalizedCallCost(CallInst *CI, ElementCount VF,
+                                      Type *ScalarRetTy,
+                                      ArrayRef<Type *> ScalarTys,
+                                      bool NeedsLegalization,
+                                      Function *&LegalVecFunc,
+                                      ElementCount &LegalVecVF) const;
+
   /// Collect values we want to ignore in the cost model.
   void collectValuesToIgnore();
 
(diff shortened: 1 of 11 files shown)


nema-ashutosh:
Kindly review - @fhahn @preames @lukel97 @alexey-bataev @rengolin @ayalz

Andarwinux:
> the vectorizer today either emits an unsupported call (forcing backend scalarization) or falls back to scalar calls outright.

https://github.com/llvm/llvm-project/issues/138171
https://github.com/llvm/llvm-project/issues/162239
https://github.com/llvm/llvm-project/issues/164642

Actually will cause miscompilation.

arsenm (commented):
I don't understand why the loop vectorizer has taken it upon itself to directly emit calls for legalization. It should just emit the intrinsics and let codegen handle the legalization of those calls.

RKSimon (commented):
Shouldn't the backend be handling this? We REALLY need to stop trying to do all this veclib call matching in the vectorizers, keep to regular llvm math vector intrinsics at the full width and then split/legalise this properly in the backend. I realise that represents a lot more yak shaving, but you're just digging yourself a hole here :(

fhahn (commented):
> Shouldn't the backend be handling this? We REALLY need to stop trying to do all this veclib call matching in the vectorizers, keep to regular llvm math vector intrinsics at the full width and then split/legalise this properly in the backend. I

If that would be feasible, that would be great. not sure about the SLP vectorizer, but I assume that would help there as well (not sure if it already uses veclib functions, but moving this out of LV would make it easier to use there as well)

nema-ashutosh:
Thanks for the feedback, I understand and agree with the concerns around layering and keeping legalization in the backend. Avoiding duplication and mid-end target-specific logic makes sense.

One concern is cost modeling. If splitting happens only in the backend, the vectorizer may still assume a single wide call, while it actually gets lowered into multiple calls with extra overhead. This can lead to suboptimal VF selection.

A possible direction could be to keep the transformation in the backend, but improve TTI to model these cases (e.g., splitting or scalarization costs). This would allow the vectorizer to make better decisions without breaking layering.

Happy to revisit the design along these lines — thanks again for the helpful discussion.

huntergr-arm:
> > Shouldn't the backend be handling this? We REALLY need to stop trying to do all this veclib call matching in the vectorizers, keep to regular llvm math vector intrinsics at the full width and then split/legalise this properly in the backend. I
> 
> If that would be feasible, that would be great. not sure about the SLP vectorizer, but I assume that would help there as well (not sure if it already uses veclib functions, but moving this out of LV would make it easier to use there as well)

Moving it out of LV would be fine if we had intrinsics for everything we want, but we've had pushback for adding new intrinsics for libm just to support vector mappings: https://github.com/llvm/llvm-project/pull/178396

There's also renewed interest in supporting OpenMP's declare simd/declare variant, e.g. https://github.com/llvm/llvm-project/pull/193884 -- which may lead to variant mappings for arbitrary user-defined functions.

I definitely agree that legalization should occur during codegen, not as part of IR optimization. I did have the same idea of allowing a wider VF and splitting into multiple calls back when we added support for sleef and similar libraries, but never bothered to follow up on it. I think it should just be expressed in the cost model, with TTI/TLI figuring out that a mapping exists with a VF that divides the requested VF evenly, and increasing the cost to match.

arsenm:
Part of the problem is just the multiple copies of libcall information we're maintaining in inconsistent ways. I want to stop using the vector mappings in TargetLibraryInfo, and move towards using RuntimeLibcallInfo. These should be merged eventually. The vector handling in TargetLibraryInfo is wonky, and requires a scalar base call and has name-based selection logic.

Andarwinux:
Miscompilations caused by veclib are frequently reported. If backend-based legalization cannot be implemented in near-term, could we merge this PR as a stopgap measure?

PR #197663, opened 2026-05-14
[lldb][windows] fix x86_64 arg register mapping for lldb-server

charles-zablit:
The register info table used by `NativeRegisterContextWindows_x86_64` (which backs the lldb-server path on Windows) was using incorrect calling conventions mappings (Linux?).

Windows x64 uses a different [convention](https://learn.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-170#parameter-passing). `RegisterContextWindows_x64.cpp` (used by the in-process
ProcessWindows plugin) already has the correct mapping.

This patch fixes 63 test failures when running `check-lldb` with `LLDB_USE_LLDB_SERVER=1`.

rdar://177064190


DavidSpickett (approved):
Matches https://learn.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-170#parameter-passing so LGTM.

Include that link in the PR description.

PR #197665, opened 2026-05-14
[InstCombine] Fold square into comparison with constant

### LINKED ISSUE #196233: [Missed Optimization] LLVM fails to pull square across cmp and into constant
The following IR squares a number and checks if it is <= 9.
```llvm
define i1 @src(i16 %x) unnamed_addr {
    %2 = mul nuw i16 %x, %x
    %3 = icmp ule i16 %2, 9
    ret i1 %3
}
```
Knowing that overflow doesn't happen, a simple optimization is to simply check if the number is <= 3.
llvm [doesn't do that](https://godbolt.org/z/YKEaz6cs1). However, alive2 [believes it to be valid](https://alive2.llvm.org/ce/z/i-ZBBY)

macsencasaus:
Closes #196233 
https://alive2.llvm.org/ce/z/PQzYhb


macsencasaus:
Waiting on verdict of #197406

macsencasaus:
@dtcxzyw @jayfoad @arsenm

llvm/lib/Transforms/InstCombine/InstCombineCompares.cpp
```diff
@@ -2188,12 +2188,49 @@ Instruction *InstCombinerImpl::foldICmpMulConstant(ICmpInst &Cmp,
...
   Value *X = Mul->getOperand(0);
 
-  // If there's no overflow:
-  // X * X == 0 --> X == 0
-  // X * X != 0 --> X != 0
-  if (Cmp.isEquality() && C.isZero() && X == Mul->getOperand(1) &&
-      (Mul->hasNoUnsignedWrap() || Mul->hasNoSignedWrap()))
-    return new ICmpInst(Pred, X, ConstantInt::getNullValue(MulTy));
+  // If comparing a square with a constant, try simplifying to comparing square
+  // roots
```
jayfoad:
```suggestion
  // roots.
```

llvm/lib/Transforms/InstCombine/InstCombineCompares.cpp
```diff
@@ -2188,12 +2188,49 @@ Instruction *InstCombinerImpl::foldICmpMulConstant(ICmpInst &Cmp,
...
+  if (X == Mul->getOperand(1) && !Cmp.isSigned()) {
+    APInt R = C.sqrtFloor();
+    APInt R2 = R * R;
+    bool IsSqr = C == R2;
```
jayfoad:
```suggestion
    bool IsSqr = C == R * R;
```

llvm/lib/Transforms/InstCombine/InstCombineCompares.cpp
```diff
@@ -2188,12 +2188,49 @@ Instruction *InstCombinerImpl::foldICmpMulConstant(ICmpInst &Cmp,
...
+
+    // X * X eq/ne C
+    if (Cmp.isEquality() &&
+        (Mul->hasNoUnsignedWrap() || (Mul->hasNoSignedWrap() && C.isZero()))) {
```
jayfoad, line 2199: Is the `X *nsw X == 0` case really useful in the real world? Otherwise this entire optimization could be guarded by `Mul->hasNoUnsignedWrap()` which seem simpler.

jayfoad (commented):
Seems good overall.

Can you provide Alive2 proofs?

llvm/lib/Transforms/InstCombine/InstCombineCompares.cpp
macsencasaus: This optimization was already present and tested for
https://github.com/llvm/llvm-project/blob/b6501be5448273989e28210b4befd3b862568aff/llvm/lib/Transforms/InstCombine/InstCombineCompares.cpp#L2191-L2196

https://github.com/llvm/llvm-project/blob/b6501be5448273989e28210b4befd3b862568aff/llvm/test/Transforms/InstCombine/icmp-mul.ll#L9-L17

jayfoad: Fair enough

llvm/lib/Transforms/InstCombine/InstCombineCompares.cpp
```diff
@@ -2188,12 +2188,48 @@ Instruction *InstCombinerImpl::foldICmpMulConstant(ICmpInst &Cmp,
...
+
+      // If constant is not a square, eq/ne is false/true respectively
+      if (!IsSqr)
+        return replaceInstUsesWith(
+            Cmp, ConstantInt::getBool(Cmp.getType(),
+                                      CmpInst::isFalseWhenEqual(Pred)));
```
dtcxzyw:
```suggestion
                                      Pred == ICmpInst::ICMP_NE));
```

dtcxzyw (approved):
LGTM

macsencasaus:
> Can you provide Alive2 proofs?

The proof is in the comment at the top.

PR #197666, opened 2026-05-14
Exclude documentation files from 80 columns rule in Coding Standard

Endilll:
This implements the accepted [RFC](https://discourse.llvm.org/t/rfc-remove-80-column-limit-in-documentation-files/89678) "Remove 80 column limit in documentation files."

### DIFF
--- a/llvm/docs/CodingStandards.rst
+++ b/llvm/docs/CodingStandards.rst
@@ -461,7 +461,7 @@ standardized on 80 columns, so some people have already configured their editors
 for it (vs something else, like 90 columns).
 
 However, documentation files are not source code files, and instead of fitting into 80 columns, they must be formatted to one sentence per line.
-This way a change in the middle of a paragraph doesn't cause unnecessary changes in subsequent lines.
+This way a change in the middle of a paragraph doesn't cause unnecessary changes in subsequent lines, making it easier for reviewers to see what has changed when documentation is updated.
 
 Whitespace
 ^^^^^^^^^^
--- a/llvm/docs/CodingStandards.rst
+++ b/llvm/docs/CodingStandards.rst
@@ -460,6 +460,9 @@ and would be detrimental to printing out code.  Also many other projects have
 standardized on 80 columns, so some people have already configured their editors
 for it (vs something else, like 90 columns).
 
+However, documentation files are not source code files, and instead of fitting into 80 columns, they must be formatted to one sentence per line.
+This way a change in the middle of a paragraph doesn't cause unnecessary changes in subsequent lines.
+
 Whitespace
 ^^^^^^^^^^
 


llvm/docs/CodingStandards.rst
```diff
@@ -460,6 +460,9 @@ and would be detrimental to printing out code.  Also many other projects have
 standardized on 80 columns, so some people have already configured their editors
 for it (vs something else, like 90 columns).
 
+However, documentation files are not source code files, and instead of fitting into 80 columns, they must be formatted to one sentence per line.
+This way a change in the middle of a paragraph doesn't cause unnecessary changes in subsequent lines.
```
AaronBallman: Maybe it's more clear to say: "This is less work for patch authors because unrelated parts of the documentation do not need to be updated to re-flow to a column limit and makes it easier for reviewers to see what has changed when updating documentation." ? (I don't feel strongly, maybe others have a better way to phrase this.)

Endilll: I tried to avoid the word "re(-)flow", because I'm not sure it's clear what it means to a potential newcomer reading this. How about:

> This way a change in the middle of a paragraph doesn't cause unnecessary changes in subsequent lines, making it easier for reviewers to see what has changed when documentation is updated.

(There are sure a million ways to word this, I'm open to more suggestions.)

AaronBallman: I think that'd be fine too.

tahonermann: The use of "must be formatted" doesn't seem consistent with the RFC or the title of the PR. Those both suggest that lines _may_ exceed the traditional 80 character limit, but do not prohibit wrapping at an earlier point.

Endilll: Sorry, I do not follow. How "one line per sentence" allows splitting a sentence between two lines, which I think is what you have in mind?

AaronBallman: > The use of "must be formatted" doesn't seem consistent with the RFC or the title of the PR. Those both suggest that lines _may_ exceed the traditional 80 character limit, but do not prohibit wrapping at an earlier point.

FWIW, I found "must" to feel a bit strong but at the same time, I think it meets the expectations of the RFC. The decision there was: source code is wrapped to an 80 column limit, documentation is one sentence per line with no column limit.

Wrapping at an earlier point actually defeats the intent of the RFC -- the goal there was to make long-term maintenance of documentation easier and soft-wrapping lines causes the problem the RFC was solving. I think if someone wanted to soft-wrap lines, we wouldn't be prescriptive and never allow it, but the onus would be on the author to explain why soft wrapping is the correct approach.

Endilll: > I found "must" to feel a bit strong

I don't disagree, but (at least) this section of the Coding Standard is written in this imperative manner. There is "must" in the previous paragraph.

tahonermann: I didn't interpret the written conclusion in the RFC as prohibiting wrapping at an earlier point than the end of a sentence. I interpret it as permitting lines that go beyond 80 characters up to the end of a sentence. If the project council meant something stronger than that (which would then render all existing documentation as out of conformance with respect to the new rules), than they should make that more explicit.

AaronBallman: > I didn't interpret the written conclusion in the RFC as prohibiting wrapping at an earlier point than the end of a sentence. I interpret it as permitting lines that go beyond 80 characters up to the end of a sentence. 

I don't believe that was the conclusion of the RFC but as with all coding standards there's an expectation that exceptions will happen. I think it's worth thinking about this statement from the RFC: 

> Upholding 80 column limit involves a significant amount of reflowing, especially when editing existing documentation. In the worst case, insertion of just couple of words at the beginning of a paragraph results in reflow changes until the end of the paragraph, increasing amount of work for PR author, distracting reviewers from the actual change, and polluting git blame output.

If the policy was "it's okay to soft wrap or not, either is equally preferred", that misses the pain point which caused the RFC in the first place. Soft wrapping documentation is a maintenance burden, we want less maintenance burdens. If there are times when soft wrapping *reduces* the maintenance burden, that's fine as an exception to the rule.

> If the project council meant something stronger than that (which would then render all existing documentation as out of conformance with respect to the new rules), than they should make that more explicit.

The project council comment mentions to not go changing all documentation to bring it into conformance, so I believe it was already understood that this renders all existing documentation out of conformance and that's fine. Almost any change to coding standards renders things out of conformance, we still have the "match local coding style" rules for cases where the new formatting may be disruptive with other local code and wants to deviate.

AaronBallman:
I was talking about this PR with @Endilll in a 1:1 today and approved verbally (he pressed the merge button before I could get my LG recorded on the PR, sorry about that mixup).

PR #187671, opened 2026-03-20
[SPIRV] Fix enqueue empty kernel

idubinov:
Function reference arguments don't get spv_bitcast after opaque pointer migration, while data pointer arguments still might. Therefore:
a. getBlockStructInstr() got updated to check G_GLOBAL_VALUE -> G_ADDRSPACE_CAST pattern for function reference arguments;
b. buildEnqueueKernel() got updated to add bitcast the block literal pointer from struct* to i8*, as required by OpEnqueueKernel.

Assisted-by: Claude Code.

PR #197667, opened 2026-05-14
[SPIRV][NFC] Pass function-pointer operand explicitly to visitFunPtrUse

idubinov:
Take the function-pointer placeholder operand as a parameter rather
than reading MI.getOperand(2) directly, so visitFunPtrUse can be
reused from instructions with a different operand layout. Pure
refactor.


idubinov:
NFC change required by: https://github.com/llvm/llvm-project/pull/187671

llvm/lib/Target/SPIRV/SPIRVModuleAnalysis.cpp
```diff
@@ -360,12 +360,15 @@ bool SPIRVModuleAnalysis::isDeclSection(const MachineRegisterInfo &MRI,
...
 // requires a special treatment.
+// FunPtrOp is the MachineOperand previously recorded via
+// SPIRVGlobalRegistry::recordFunctionPointer, identifying which Function
+// this placeholder refers to.
 void SPIRVModuleAnalysis::visitFunPtrUse(
-    Register OpReg, InstrGRegsMap &SignatureToGReg,
+    Register OpReg, const MachineOperand *FunPtrOp,
+    InstrGRegsMap &SignatureToGReg,
     std::map<const Value *, unsigned> &GlobalToGReg, const MachineFunction *MF,
     const MachineInstr &MI) {
```
aobolensk: `MI` is not used in this func anymore, let's drop it as well

llvm/lib/Target/SPIRV/SPIRVModuleAnalysis.cpp
```diff
@@ -360,12 +360,15 @@ bool SPIRVModuleAnalysis::isDeclSection(const MachineRegisterInfo &MRI,
 // This is a special case of a function pointer refering to a possibly
```
maarquitos14:
```suggestion
// This is a special case of a function pointer referring to a possibly
```

While we're here.

maarquitos14 (approved):
LGTM, just a nit.

llvm/lib/Target/SPIRV/SPIRVModuleAnalysis.cpp
```diff
@@ -357,15 +357,18 @@ bool SPIRVModuleAnalysis::isDeclSection(const MachineRegisterInfo &MRI,
          TII->isInlineAsmDefInstr(MI);
 }
 
-// This is a special case of a function pointer refering to a possibly
+// This is a special case of a function pointer referring to a possibly
```
MrSidims, line 360: I can't stop myself but wonder, if the backend should adopt dummy `OpForward` from llvm-spirv... Not in this PR tho.

MrSidims (approved):

llvm/lib/Target/SPIRV/SPIRVModuleAnalysis.cpp
idubinov: done

aobolensk (approved):

PR #197668, opened 2026-05-14
[SPIRV] Track Function placeholder via OpUndef + recordFunctionPointer

idubinov:
Emit OpUndef and record the operand -> Function mapping instead of
OpConstantNull, so a future consumer can resolve the function's <id>.
NFC for current behavior.

### DIFF
--- a/llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
+++ b/llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
@@ -6608,11 +6608,20 @@ bool SPIRVInstructionSelector::selectGlobalValue(
         MIB2.constrainAllUses(TII, TRI, RBI);
         return true;
       }
+      // Even without SPV_INTEL_function_pointers a function reference can
+      // still be load-bearing (e.g. the Invoke operand of OpEnqueueKernel).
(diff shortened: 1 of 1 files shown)


idubinov:
NFC change required by: https://github.com/llvm/llvm-project/pull/187671

maarquitos14 (commented):
Could we have a test for this? It would also help me understand the scenario described. In principle, LGTM, but I'd like to see a use-case.

llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
```diff
@@ -6621,11 +6608,20 @@ bool SPIRVInstructionSelector::selectGlobalValue(
         MIB2.constrainAllUses(TII, TRI, RBI);
         return true;
       }
+      // Even without SPV_INTEL_function_pointers a function reference can
```
MrSidims, line 6611: I'd prefer to move this patch into the main one. I understand complain from @maarquitos14 as it's hard to see, that the PR is addressing one of my comments without knowing the context.

idubinov: changes in main PR. this pr closed

PR #197669, opened 2026-05-14
[lldb][windows] Keep int3 breakpoints inside the debugger on lldb-server

charles-zablit:
`NativeProcessWindows::OnDebugException` forwarded `STATUS_BREAKPOINT` exceptions, which returned `ExceptionResult::SendToApplication` on first chance.

On Windows an unhandled `STATUS_BREAKPOINT` is fatal to the process. Programs containing explicit `int3`, `__debugbreak()`, or `__builtin_debugtrap()` were killed when hitting the instruction and the subsequent user `continue` had nothing to resume.

This patch mirrors the `ProcessWindows` plugin handling of `EXCEPTION_BREAKPOINT`. It fixes all shell `Register/*-read.test` / `*-write.test` cases and several expression side-effect tests in `check-lldb` with `LLDB_USE_LLDB_SERVER=1`.

rdar://177072682

### DIFF
--- a/lldb/source/Plugins/Process/Windows/Common/NativeProcessWindows.cpp
+++ b/lldb/source/Plugins/Process/Windows/Common/NativeProcessWindows.cpp
@@ -564,17 +564,19 @@ NativeProcessWindows::OnDebugException(bool first_chance,
       return ExceptionResult::BreakInDebugger;
     }
 
-    // Any remaining STATUS_BREAKPOINT is a breakpoint instruction in the
-    // program's own code (e.g. `__debugbreak()` or `__builtin_debugtrap()`).
-    // Stop the debugger and let the user decide what to do.
-    std::string desc =
-        formatv("Exception {0} encountered at address {1}",
-                llvm::format_hex(record.GetExceptionCode(), 8),
-                llvm::format_hex(record.GetExceptionAddress(), 8))
-            .str();
-    StopThread(record.GetThreadID(), StopReason::eStopReasonException,
-               std::move(desc));
-    SetState(eStateStopped, true);
+    {
+      // Any remaining STATUS_BREAKPOINT is a breakpoint instruction in the
+      // program's own code (e.g. `__debugbreak()` or `__builtin_debugtrap()`).
+      // Stop the debugger and let the user decide what to do.
+      std::string desc =
+          formatv("Exception {0:x8} encountered at address {1:x8}",
+                  record.GetExceptionCode(), record.GetExceptionAddress())
+              .str();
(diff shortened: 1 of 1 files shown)


lldb/source/Plugins/Process/Windows/Common/NativeProcessWindows.cpp
```diff
@@ -564,7 +564,18 @@ NativeProcessWindows::OnDebugException(bool first_chance,
...
     }
 
-    [[fallthrough]];
+    // Any remaining STATUS_BREAKPOINT is a breakpoint instruction in the
+    // program's own code (e.g. `__debugbreak()` or `__builtin_debugtrap()`).
+    // Stop the debugger and let the user decide what to do.
+    std::string desc =
+        formatv("Exception {0} encountered at address {1}",
+                llvm::format_hex(record.GetExceptionCode(), 8),
+                llvm::format_hex(record.GetExceptionAddress(), 8))
```
Nerixyz: It's not well documented in the programmer's manual, but you can use `{0:x8}` to specify the number of digits.
I always need to search this up. The reference is here: https://github.com/llvm/llvm-project/blob/c5695b8f60a57caa46491ce78c7be9e336ad7fe9/llvm/include/llvm/Support/FormatProviders.h#L96-L102

lldb/source/Plugins/Process/Windows/Common/NativeProcessWindows.cpp
```diff
@@ -564,7 +564,18 @@ NativeProcessWindows::OnDebugException(bool first_chance,
       return ExceptionResult::BreakInDebugger;
     }
 
-    [[fallthrough]];
+    // Any remaining STATUS_BREAKPOINT is a breakpoint instruction in the
+    // program's own code (e.g. `__debugbreak()` or `__builtin_debugtrap()`).
+    // Stop the debugger and let the user decide what to do.
+    std::string desc =
```
Nerixyz: You'll need to wrap this `case` in a block to be able to declare variables inside:
```cpp
case Foo: {
  std::string a_string;
} break;
```

charles-zablit: Thanks! I kept the return outside the block as I find it clearer to read. Happy to change it if needed.

lldb/source/Plugins/Process/Windows/Common/NativeProcessWindows.cpp
charles-zablit: I did not know about this, thanks!

Nerixyz (approved):

PR #197670, opened 2026-05-14
[BOLT][DebugInfo] Make parallel DWARF debug names generation deterministic(1/2)

Thrrreeee:
This change makes BOLT's split DWARF rewriting path produce deterministic debug names output when processing CUs in parallel. 

Previously, several pieces of state used while building `.debug_names` and updating DWO names were populated during parallel CU processing. This could make CU/TU index assignment, cross- 
  CU DIE lookup, and duplicate DWO name disambiguation depend on thread scheduling order, which in turn could lead to non-reproducible output across runs.                                     
                                                                                                                                                                                               
The patch moves DWO name collection into a deterministic pre-processing step, pre-allocates debug names CU / foreign TU slots before concurrent processing, protects cross-CU DIE tracking 
  with a mutex, and adds a stable tie-breaker when sorting accelerator table entries. It also centralizes creation of the DWO output directory before parallel work starts.                    
                                                                                                                                                                                               
This improves reproducibility of BOLT output, makes debug info rewriting less sensitive to parallel execution order, and avoids subtle nondeterminism in generated `.debug_names` / split  DWARF artifacts. 

RFC - [[RFC][BOLT] A New Parallel DWARF Processing Approach in BOLT](https://discourse.llvm.org/t/rfc-bolt-a-new-parallel-dwarf-processing-approach-in-bolt/90736) (The overall changes will be split into two separate PRs for easier review.)
Draft PR - [[BOLT][DebugInfo] Add a new parallel DWARF processing(2/2)](https://github.com/llvm/llvm-project/pull/197859) (including all changes)

### DIFF
--- a/bolt/include/bolt/Core/DebugNames.h
+++ b/bolt/include/bolt/Core/DebugNames.h
@@ -104,7 +104,7 @@ private:
   /// Track whether AcceleratorTable have been preallocated.
   bool Preallocated = false;
   /// Map from DWO ID to preallocated CU index.
-  DenseMap<uint64_t, uint32_t> DWOIdToCUIndex;
+  std::unordered_map<uint64_t, uint32_t> DWOIdToCUIndex;
   /// Represents a group of entries with identical name (and hence, hash value).
   struct HashData {
     uint64_t StrOffset;
--- a/bolt/include/bolt/Core/DebugNames.h
+++ b/bolt/include/bolt/Core/DebugNames.h
@@ -103,8 +103,6 @@ private:
   DenseMap<uint64_t, uint32_t> TUHashToIndexMap;
   /// Track whether AcceleratorTable have been preallocated.
   bool Preallocated = false;
-  /// Map from DWO ID to preallocated CU index.
-  std::unordered_map<uint64_t, uint32_t> DWOIdToCUIndex;
   /// Represents a group of entries with identical name (and hence, hash value).
   struct HashData {
     uint64_t StrOffset;
@@ -138,7 +136,7 @@ private:
   uint32_t AugmentationStringSize = 0;
   dwarf::Form CUIndexForm = dwarf::DW_FORM_data4;
   dwarf::Form TUIndexForm = dwarf::DW_FORM_data4;
-  constexpr static uint32_t BADCUOFFSET = 0xBADBA;
+  constexpr static uint32_t BADCUOFFSET = 0xBADBAD;
 
   BucketList Buckets;
 
@@ -152,7 +150,7 @@ private:
   std::unique_ptr<raw_svector_ostream> AugStringtream;
   llvm::DenseMap<llvm::hash_code, uint64_t> StrCacheToOffsetMap;
   // Contains DWO ID to CUList Index.
-  llvm::DenseMap<uint64_t, uint32_t> CUOffsetsToPatch;
+  std::unordered_map<uint64_t, uint32_t> CUOffsetsToPatch;
   // Contains a map of Entry ID to Entry relative offset.
   llvm::DenseMap<uint64_t, uint32_t> EntryRelativeOffsets;
   llvm::DenseMap<uint64_t, std::pair<DWARFUnit *, const DIE *>> CrossCUDies;
--- a/bolt/include/bolt/Core/DebugNames.h
+++ b/bolt/include/bolt/Core/DebugNames.h
@@ -17,6 +17,7 @@
 #include "bolt/Core/DebugData.h"
 #include "llvm/CodeGen/AccelTable.h"
 
+#include <mutex>
 namespace llvm {
 namespace bolt {
 class BOLTDWARF5AccelTableData : public DWARF5AccelTableData {
@@ -71,16 +72,20 @@ public:
   std::unique_ptr<DebugBufferVector> releaseBuffer() {
     return std::move(FullTableBuffer);
   }
-  /// Adds a DIE that is referenced across CUs.
-  void addCrossCUDie(DWARFUnit *Unit, const DIE *Die) {
-    CrossCUDies.insert({Die->getOffset(), {Unit, Die}});
-  }
+  void addCrossCUDie(DWARFUnit *Unit, const DIE *Die);
+  /// Looks up a DIE that is referenced across CUs.
+  std::optional<std::pair<DWARFUnit *, const DIE *>>
+  getCrossCUDie(uint64_t Offset);
   /// Returns true if the DIE can generate an entry for a cross cu reference.
   /// This only checks TAGs of a DIE because when this is invoked DIE might not
   /// be fully constructed.
   bool canGenerateEntryWithCrossCUReference(
       const DWARFUnit &Unit, const DIE &Die,
       const DWARFAbbreviationDeclaration::AttributeSpec &AttrSpec);
+
+  /// Pre-allocate CU and Foreign TU slots in deterministic order.
+  /// Must be called before any concurrent addAccelTableEntry() calls.
+  void preAllocateUnits(DWARFContext &DwCtx);
 
 private:
   BinaryContext &BC;
@@ -90,12 +95,16 @@ private:
   StringRef StrSection;
   uint64_t CurrentUnitOffset = 0;
   const DWARFUnit *CurrentUnit = nullptr;
+  std::mutex CrossCUDiesMutex;
   std::unordered_map<uint32_t, uint32_t> AbbrevTagToIndexMap;
   /// Contains a map of TU hashes to a Foreign TU indices.
   /// This is used to reduce the size of Foreign TU list since there could be
   /// multiple TUs with the same hash.
   DenseMap<uint64_t, uint32_t> TUHashToIndexMap;
-
+  /// Track whether AcceleratorTable have been preallocated.
+  bool Preallocated = false;
+  /// Map from DWO ID to preallocated CU index.
+  DenseMap<uint64_t, uint32_t> DWOIdToCUIndex;
   /// Represents a group of entries with identical name (and hence, hash value).
   struct HashData {
     uint64_t StrOffset;
@@ -129,6 +138,7 @@ private:
   uint32_t AugmentationStringSize = 0;
   dwarf::Form CUIndexForm = dwarf::DW_FORM_data4;
   dwarf::Form TUIndexForm = dwarf::DW_FORM_data4;
+  constexpr static uint32_t BADCUOFFSET = 0xBADBA;
 
   BucketList Buckets;
 
--- a/bolt/lib/Core/DebugNames.cpp
+++ b/bolt/lib/Core/DebugNames.cpp
@@ -74,13 +74,13 @@ void DWARF5AcceleratorTable::preAllocateUnits(DWARFContext &DwCtx) {
     if (!DWOId)
       continue;
 
-    const DWARFDie UnitDIE = CU->getUnitDIE();
-    if (!UnitDIE.find({dwarf::DW_AT_dwo_name, dwarf::DW_AT_GNU_dwo_name}))
-      continue;
-
     uint32_t Idx = CUList.size();
     CUList.push_back(BADCUOFFSET);
     CUOffsetsToPatch[*DWOId] = Idx;
+
+    const DWARFDie UnitDIE = CU->getUnitDIE();
+    if (!UnitDIE.find({dwarf::DW_AT_dwo_name, dwarf::DW_AT_GNU_dwo_name}))
+      continue;
 
     std::optional<DWARFUnit *> DWOCU = BC.getDWOCU(*DWOId);
     if (!DWOCU || !*DWOCU)
--- a/bolt/lib/Core/DebugNames.cpp
+++ b/bolt/lib/Core/DebugNames.cpp
@@ -86,7 +86,7 @@ void DWARF5AcceleratorTable::preAllocateUnits(DWARFContext &DwCtx) {
     if (!DWOCU || !*DWOCU)
       continue;
 
-    // check dwo_types_section_units for DWARF4 type units in DWO.
+    // Iterate DWO info section units for type units (Foreign TU list).
     DWARFContext &DWOCtx = (*DWOCU)->getContext();
     for (const auto &TU : DWOCtx.dwo_units()) {
       if (!TU->isTypeUnit())
--- a/bolt/lib/Core/DebugNames.cpp
+++ b/bolt/lib/Core/DebugNames.cpp
@@ -88,7 +88,7 @@ void DWARF5AcceleratorTable::preAllocateUnits(DWARFContext &DwCtx) {
 
     // Iterate DWO info section units for type units (Foreign TU list).
     DWARFContext &DWOCtx = (*DWOCU)->getContext();
-    for (const auto &TU : DWOCtx.dwo_units()) {
+    for (const auto &TU : DWOCtx.dwo_info_section_units()) {
       if (!TU->isTypeUnit())
         continue;
       const uint64_t TUHash = cast<DWARFTypeUnit>(TU.get())->getTypeHash();
--- a/bolt/lib/Core/DebugNames.cpp
+++ b/bolt/lib/Core/DebugNames.cpp
@@ -67,25 +67,27 @@ DWARF5AcceleratorTable::DWARF5AcceleratorTable(
 }
 
 void DWARF5AcceleratorTable::preAllocateUnits(DWARFContext &DwCtx) {
-  // Collect all DWO IDs in deterministic order (by CU offset in .debug_info).
-  // This is single-threaded, called before parallel process.
-  for (const auto &CU : DwCtx.compile_units()) {
-    if (std::optional<uint64_t> DWOId = CU->getDWOId()) {
-      uint32_t Idx = CUList.size();
-      CUList.push_back(BADCUOFFSET);
-      CUOffsetsToPatch[*DWOId] = Idx;
-      DWOIdToCUIndex[*DWOId] = Idx;
-    }
-  }
+  // Single pass: allocate CU slots and collect Foreign TU hashes
+  // in deterministic order (by CU offset in .debug_info).
   for (const auto &CU : DwCtx.compile_units()) {
     std::optional<uint64_t> DWOId = CU->getDWOId();
     if (!DWOId)
       continue;
+
+    const DWARFDie UnitDIE = CU->getUnitDIE();
+    if (!UnitDIE.find({dwarf::DW_AT_dwo_name, dwarf::DW_AT_GNU_dwo_name}))
+      continue;
+
+    uint32_t Idx = CUList.size();
+    CUList.push_back(BADCUOFFSET);
+    CUOffsetsToPatch[*DWOId] = Idx;
+
     std::optional<DWARFUnit *> DWOCU = BC.getDWOCU(*DWOId);
     if (!DWOCU || !*DWOCU)
       continue;
-    DWARFContext &DWOCtx = (*DWOCU)->getContext();
+
     // check dwo_types_section_units for DWARF4 type units in DWO.
+    DWARFContext &DWOCtx = (*DWOCU)->getContext();
     for (const auto &TU : DWOCtx.dwo_units()) {
       if (!TU->isTypeUnit())
         continue;
@@ -293,8 +295,8 @@ uint32_t DWARF5AcceleratorTable::getUnitID(const DWARFUnit &Unit,
     return LocalTUList.size() - 1;
   }
   if (Preallocated && DWOID) {
-    auto Iter = DWOIdToCUIndex.find(*DWOID);
-    assert(Iter != DWOIdToCUIndex.end() && "DWO ID not preallocated");
+    auto Iter = CUOffsetsToPatch.find(*DWOID);
+    assert(Iter != CUOffsetsToPatch.end() && "DWO ID not preallocated");
     return Iter->second;
   }
   return CUList.size() - 1;
@@ -513,6 +515,20 @@ void DWARF5AcceleratorTable::computeBucketCount() {
     BucketCount = std::max<uint32_t>(UniqueHashCount, 1);
 }
 
+/// Returns a stable compile-unit sort key for a debug_names entry.
+static unsigned getCompileUnitKey(const BOLTDWARF5AccelTableData *Entry) {
+  if (Entry->getSecondUnitID())
+    return *Entry->getSecondUnitID();
+  if (!Entry->isTU())
+    return Entry->getUnitID();
+  return std::numeric_limits<unsigned>::max();
+}
+
+static unsigned getTypeUnitKey(const BOLTDWARF5AccelTableData *Entry) {
+  return Entry->isTU() ? Entry->getUnitID()
+                       : std::numeric_limits<unsigned>::max();
+}
+
 /// Bucket code as in: AccelTableBase::finalize()
 void DWARF5AcceleratorTable::finalize() {
   if (!NeedToCreate)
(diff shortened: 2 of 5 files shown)


rafaelauler:
First of all, my sincere thank you for the contribution to the community. The speedups from the RFC are quite impressive.

I'll resume taking another pass at these two diffs (197670 and 197859) next week. For now, I have two comments:

I ran both perf/mem usage tests, and them some lldb tests to check we're not accidentally breaking any debugging functionality, and I hit a problem in the test of lldb/test/Shell/SymbolFile/DWARF/clang-gmodules-type-lookup.c . When we process that file with BOLT, before your stack, test passes fine, but with your stack, we're crashing and, fixing the crash, we hit another issue. Here are the issues:

1)
The crash happened in DWARF5AcceleratorTable::preAllocateUnits because preAllocateUnits() and getDWONameMap() insert DWO IDs and type hashes (arbitrary 64-bit values) into DenseMap<uint64_t, uint32_t>. DenseMap reserves UINT64_MAX and UINT64_MAX-1 as sentinel keys, so any DWO ID matching those values triggers an assertion. Clang modules consistently produce DWO ID 0xfffffffffffffffe (the tombstone key). One way I got around that was by switching TUHashToIndexMap, DWOIdToCUIndex, and CUOffsetsToPatch from DenseMap to std::unordered_map. If you don't like that, I'm open to ideas on how we're going to guard against this case, but we do need to handle them, otherwise our CI will consistently fail on a genuine test that we're breaking.

2)
After fixing the crash, the test still failed because of a problem in the next PR up the stack 197859, where we have bug in a std::optional comparison against nullptr in finalizeSkeletonAndStrSection(), DWARFRewriter.cpp line 736, when defining the HasSplitCU bool. The check does BC.getDWOCU(*DWOId) != nullptr, and
getDWOCU() returns std::optional (instead of a pointer). So when it returns std::nullopt to tell us there is no such CU, we still behave as if there is a CU. This caused BOLT to incorrectly treat every CU with a DWO ID as having a split CU, even when the DWO file wasn't valid (e.g., in our test, clang modules PCH files). BOLT would then rewrite DW_AT_dwo_name without writing a corresponding output file, breaking the debugger lookups and failing the test. The fix here that I used was to dereference the optional before comparing the pointer to nullptr:

  std::optional<DWARFUnit *> DWOCU = DWOId ? BC.getDWOCU(*DWOId) : std::nullopt;
  const bool HasSplitCU = DWOCU && *DWOCU != nullptr;
  
Anyway, again, great work. I'll keep poking next week to see if there are any remaining bugs.

Thrrreeee:
> First of all, my sincere thank you for the contribution to the community. The speedups from the RFC are quite impressive.
> I'll resume taking another pass at these two diffs (197670 and 197859) next week. For now, I have two comments:

Thank you for your review and testing. I will test and fix the bugs you mentioned next week. Please contact me if you have any further questions or test more bugs, I will fix the code promptly..

Thrrreeee:
Hi, @rafaelauler 
> One way I got around that was by switching TUHashToIndexMap, DWOIdToCUIndex, and CUOffsetsToPatch from DenseMap to std::unordered_map. If you don't like that, I'm open to ideas on how we're going to guard against this case, but we do need to handle them, otherwise our CI will consistently fail on a genuine test that we're breaking.

 I have addressed the DWOIdToCUIndex case in a separate commit (https://github.com/llvm/llvm-project/pull/197670/changes/bfad9afec504bf14118788d8acaad0aa0537948b), converting it to std::unordered_map. Based on your description, TUHashToIndexMap and CUOffsetsToPatch should not be affected — only DWOIdToCUIndex hits the problematic path since Clang has a special case around DWO ID handling.

> std::optional<DWARFUnit *> DWOCU = DWOId ? BC.getDWOCU(*DWOId) : std::nullopt; const bool HasSplitCU = DWOCU && *DWOCU != nullptr;

I have incorporated your suggestion in this commit (https://github.com/llvm/llvm-project/pull/197859/changes/f976012bf105cf0006523878e1f4f749862b0f0d), updating the logic to properly handle the std::optional<DWARFUnit *> return.

bolt/lib/Core/DebugNames.cpp
```diff
@@ -64,6 +66,39 @@ DWARF5AcceleratorTable::DWARF5AcceleratorTable(
...
+  }
+  for (const auto &CU : DwCtx.compile_units()) {
+    std::optional<uint64_t> DWOId = CU->getDWOId();
+    if (!DWOId)
+      continue;
+    std::optional<DWARFUnit *> DWOCU = BC.getDWOCU(*DWOId);
+    if (!DWOCU || !*DWOCU)
+      continue;
+    DWARFContext &DWOCtx = (*DWOCU)->getContext();
+    // check dwo_types_section_units for DWARF4 type units in DWO.
```
ayermolo: DWARF 4 shouldn't be part of debug_names section.

dwo_info_section_units I think would be a better one.

ayermolo:
> Hi, @rafaelauler
> 
> > One way I got around that was by switching TUHashToIndexMap, DWOIdToCUIndex, and CUOffsetsToPatch from DenseMap to std::unordered_map. If you don't like that, I'm open to ideas on how we're going to guard against this case, but we do need to handle them, otherwise our CI will consistently fail on a genuine test that we're breaking.
> 
> I have addressed the DWOIdToCUIndex case in a separate commit ([bfad9af](https://github.com/llvm/llvm-project/commit/bfad9afec504bf14118788d8acaad0aa0537948b)), converting it to std::unordered_map. Based on your description, TUHashToIndexMap and CUOffsetsToPatch should not be affected — only DWOIdToCUIndex hits the problematic path since Clang has a special case around DWO ID handling.
> 
> > std::optional<DWARFUnit *> DWOCU = DWOId ? BC.getDWOCU(*DWOId) : std::nullopt; const bool HasSplitCU = DWOCU && *DWOCU != nullptr;
> 
> I have incorporated your suggestion in this commit ([f976012](https://github.com/llvm/llvm-project/commit/f976012bf105cf0006523878e1f4f749862b0f0d)), updating the logic to properly handle the std::optional<DWARFUnit *> return.

Not sure I follow.

```
void DWARF5AcceleratorTable::preAllocateUnits(DWARFContext &DwCtx) {
  // Collect all DWO IDs in deterministic order (by CU offset in .debug_info).
  // This is single-threaded, called before parallel process.
  for (const auto &CU : DwCtx.compile_units()) {
    if (std::optional<uint64_t> DWOId = CU->getDWOId()) {
      uint32_t Idx = CUList.size();
      CUList.push_back(BADCUOFFSET);
      CUOffsetsToPatch[*DWOId] = Idx;
      DWOIdToCUIndex[*DWOId] = Idx;
    }
  }
```

Always gets executed for debug_names. So if DWOIdToCUIndex had an issue CUOffsetsToPatch would also.

bolt/lib/Core/DebugNames.cpp
```diff
@@ -64,6 +66,39 @@ DWARF5AcceleratorTable::DWARF5AcceleratorTable(
   }
 }
 
+void DWARF5AcceleratorTable::preAllocateUnits(DWARFContext &DwCtx) {
+  // Collect all DWO IDs in deterministic order (by CU offset in .debug_info).
+  // This is single-threaded, called before parallel process.
+  for (const auto &CU : DwCtx.compile_units()) {
+    if (std::optional<uint64_t> DWOId = CU->getDWOId()) {
+      uint32_t Idx = CUList.size();
```
ayermolo: What will happen if CU is missing DW_AT_dwo_name?
I thin we will end up with BADCUOFFSET in debug_names, no?

bolt/lib/Core/DebugNames.cpp
```diff
@@ -64,6 +66,39 @@ DWARF5AcceleratorTable::DWARF5AcceleratorTable(
...
+      CUList.push_back(BADCUOFFSET);
+      CUOffsetsToPatch[*DWOId] = Idx;
+      DWOIdToCUIndex[*DWOId] = Idx;
```
ayermolo: Isn't one of these redundant?

bolt/include/bolt/Core/DebugNames.h
```diff
@@ -129,6 +138,7 @@ class DWARF5AcceleratorTable {
   uint32_t AugmentationStringSize = 0;
   dwarf::Form CUIndexForm = dwarf::DW_FORM_data4;
   dwarf::Form TUIndexForm = dwarf::DW_FORM_data4;
+  constexpr static uint32_t BADCUOFFSET = 0xBADBA;
```
ayermolo: 0xBADBAD? Why was this changed?

bolt/lib/Core/DebugNames.cpp
```diff
@@ -475,7 +538,25 @@ void DWARF5AcceleratorTable::finalize() {
     for (HashData *H : Bucket)
       llvm::stable_sort(H->Values, [](const BOLTDWARF5AccelTableData *LHS,
                                       const BOLTDWARF5AccelTableData *RHS) {
-        return LHS->getDieOffset() < RHS->getDieOffset();
+        // Sort entries by (DieOffset, CompileUnitID, TypeUnitID) to ensure
+        // deterministic output order across multi-threaded processing runs.
+        const auto GetCompileUnitKey =
```
ayermolo: Maybe hoist some of these comparators to static func, to make this a bit cleaner to read?

```
static unsigned getCompileUnitKey(const BOLTDWARF5AccelTableData *Entry) {
  if (Entry->getSecondUnitID())
    return *Entry->getSecondUnitID();
  if (!Entry->isTU())
    return Entry->getUnitID();
  return std::numeric_limits<unsigned>::max();
}
static unsigned getTypeUnitKey(const BOLTDWARF5AccelTableData *Entry) {
  return Entry->isTU() ? Entry->getUnitID()
                       : std::numeric_limits<unsigned>::max();
}
```

bolt/lib/Core/DebugNames.cpp
Thrrreeee: > dwo_info_section_units I think would be a better one.

fix up

bolt/lib/Core/DebugNames.cpp
Thrrreeee: > Maybe hoist some of these comparators to static func, to make this a bit cleaner to read?

fix up, add two static func

bolt/include/bolt/Core/DebugNames.h
Thrrreeee: > 0xBADBAD? Why was this changed?

This is a mistake, fixed

bolt/lib/Core/DebugNames.cpp
Thrrreeee: > Isn't one of these redundant?

Thanks, I have fixed it.

bolt/lib/Core/DebugNames.cpp
Thrrreeee: >  I think we will end up with BADCUOFFSET in debug_names, no?

I have added an judge to ensure DW_AT_dwo_name is not missing to avoid leaving BADCUOFFSET in the final .debug_names output

bolt/lib/Core/DebugNames.cpp
```diff
@@ -64,6 +66,41 @@ DWARF5AcceleratorTable::DWARF5AcceleratorTable(
...
+    CUList.push_back(BADCUOFFSET);
+    CUOffsetsToPatch[*DWOId] = Idx;
+
+    std::optional<DWARFUnit *> DWOCU = BC.getDWOCU(*DWOId);
+    if (!DWOCU || !*DWOCU)
+      continue;
+
+    // Iterate DWO info section units for type units (Foreign TU list).
+    DWARFContext &DWOCtx = (*DWOCU)->getContext();
+    for (const auto &TU : DWOCtx.dwo_units()) {
```
ayermolo: I think there is still an issue with: for (const auto &TU : DWOCtx.dwo_units())

I believe it will iterate over .debug_types.dwo which are DWARF4.

I think dwo_info_section_units should do the trick, as in DWARF5 Type Units are moved into .debug_info section.

Thrrreeee: > I think dwo_info_section_units should do the trick

Fixed

bolt/lib/Core/DebugNames.cpp
```diff
@@ -64,6 +66,41 @@ DWARF5AcceleratorTable::DWARF5AcceleratorTable(
...
+    const DWARFDie UnitDIE = CU->getUnitDIE();
+    if (!UnitDIE.find({dwarf::DW_AT_dwo_name, dwarf::DW_AT_GNU_dwo_name}))
+      continue;
+
+    uint32_t Idx = CUList.size();
+    CUList.push_back(BADCUOFFSET);
+    CUOffsetsToPatch[*DWOId] = Idx;
+
+    std::optional<DWARFUnit *> DWOCU = BC.getDWOCU(*DWOId);
+    if (!DWOCU || !*DWOCU)
```
ayermolo, line 86: Can we move this above 
 uint32_t Idx = CUList.size();
    CUList.push_back(BADCUOFFSET);
    CUOffsetsToPatch[*DWOId] = Idx;

?

rafaelauler:
@ayermolo thanks for helping review!

@Thrrreeee 
I have two more bugs that I found that needs fixing, but hopefully these should be the last ones and, fixing those locally I got it to pass all of my internal tests too.

I.  Corruption in empty ranges offset (debug_ranges): during the parallel merge phase, we merge per-bucket range buffers into a global .debug_ranges section. During merge, fixupDWARFRanges shifts every DW_AT_ranges value by the global buffer's current size. However, offset 0 is reserved as the empty ranges terminator (see DebugRangesSectionWriter::EmptyRangesOffset). initSection() writes a 16-byte zero terminator at offset 0 in the global buffer. CUs with no address ranges (in my internal tests, those came from crtn.S, for example) get DW_AT_ranges=0 pointing to this terminator. Because my crtn.S was among the last DIEs to be processed, shifting that 0 by the global offset produced an invalid offset pointing at the end of .debug_ranges (because the debug_ranges was already fully processed and in its final size), causing "invalid range list offset" errors in llvm-dwarfdump, failing my test.

To fix that, we can skip the fixup when DW_AT_ranges equals the empty ranges offset, since offset 0 is valid in the global buffer as-is, and call initSection() on local per-bucket LegacyRangesWriters to reserve offset 0 as the empty terminator, matching the global buffer convention. Without this, real range data starts at offset 0 in local buffers, colliding with the empty ranges sentinel. When appending the local buffer to the global buffer during merge, we also need to skip the local buffer's 16-byte terminator prefix to avoid duplicating terminators in the global section, and adjust the fixup offset accordingly. Here's an example of the changes I had locally for that (feel free to write something different, this is just what I used to unblock me): https://pastebin.com/dGvE9xjE

(this one more serious)
II. Data race in parallel DWO processing when using DWP files: when multiple buckets process CUs whose DWO units reside in the same DWP file, their DWODIEBuilder instances call buildDWOUnit(), which calls buildTypeUnits() concurrently on the same DWARFContext. DWARFContext performs lazy parsing internally and methods like getTUIndex() populate internal caches on first access. These lazy mutations are not thread-safe by default. The result is a sporadic SIGSEGV in buildTypeUnits, reproducible by running the dwarf5-df-types-dup-dwp-input.test test in a loop.

This was fixable as DWARFContext already supports a ThreadSafe=true mode (see comment on DWARFContextState in DWARFContext.h) that protects lazy initialization with internal locking. BOLT currently creates its DWARFContext with the default ThreadSafe=false. Passing ThreadSafe=true in RewriteInstance.cpp at the DWARFContext::create() call fixes the race with no measurable performance overhead in my testing.

If that lock is a problem for you, feel free to suggest an alternative.

III. Discussion on default value for DebugThreadCount: I did several experiments with internal apps where I detected no significant difference in memory usage, but a decent 20 to 30% improvement in BOLT wall time. In a machine with 80 cores, I noticed parallelism wins start to taper off at thread count 30 to 40. When I bump to 80 to use all machine cores, I saw no improvements to wall time.

Based on that, I would suggest that we initially set debug-thread-count to 32 by default. What do you think?

ayermolo:
> @ayermolo thanks for helping review!
> 
> @Thrrreeee I have two more bugs that I found that needs fixing, but hopefully these should be the last ones and, fixing those locally I got it to pass all of my internal tests too.
> 
> I. Corruption in empty ranges offset (debug_ranges): during the parallel merge phase, we merge per-bucket range buffers into a global .debug_ranges section. During merge, fixupDWARFRanges shifts every DW_AT_ranges value by the global buffer's current size. However, offset 0 is reserved as the empty ranges terminator (see DebugRangesSectionWriter::EmptyRangesOffset). initSection() writes a 16-byte zero terminator at offset 0 in the global buffer. CUs with no address ranges (in my internal tests, those came from crtn.S, for example) get DW_AT_ranges=0 pointing to this terminator. Because my crtn.S was among the last DIEs to be processed, shifting that 0 by the global offset produced an invalid offset pointing at the end of .debug_ranges (because the debug_ranges was already fully processed and in its final size), causing "invalid range list offset" errors in llvm-dwarfdump, failing my test.
> 
> To fix that, we can skip the fixup when DW_AT_ranges equals the empty ranges offset, since offset 0 is valid in the global buffer as-is, and call initSection() on local per-bucket LegacyRangesWriters to reserve offset 0 as the empty terminator, matching the global buffer convention. Without this, real range data starts at offset 0 in local buffers, colliding with the empty ranges sentinel. When appending the local buffer to the global buffer during merge, we also need to skip the local buffer's 16-byte terminator prefix to avoid duplicating terminators in the global section, and adjust the fixup offset accordingly. Here's an example of the changes I had locally for that (feel free to write something different, this is just what I used to unblock me): https://pastebin.com/dGvE9xjE
> 
> (this one more serious) II. Data race in parallel DWO processing when using DWP files: when multiple buckets process CUs whose DWO units reside in the same DWP file, their DWODIEBuilder instances call buildDWOUnit(), which calls buildTypeUnits() concurrently on the same DWARFContext. DWARFContext performs lazy parsing internally and methods like getTUIndex() populate internal caches on first access. These lazy mutations are not thread-safe by default. The result is a sporadic SIGSEGV in buildTypeUnits, reproducible by running the dwarf5-df-types-dup-dwp-input.test test in a loop.
> 
> This was fixable as DWARFContext already supports a ThreadSafe=true mode (see comment on DWARFContextState in DWARFContext.h) that protects lazy initialization with internal locking. BOLT currently creates its DWARFContext with the default ThreadSafe=false. Passing ThreadSafe=true in RewriteInstance.cpp at the DWARFContext::create() call fixes the race with no measurable performance overhead in my testing.
> 
> If that lock is a problem for you, feel free to suggest an alternative.
> 
> III. Discussion on default value for DebugThreadCount: I did several experiments with internal apps where I detected no significant difference in memory usage, but a decent 20 to 30% improvement in BOLT wall time. In a machine with 80 cores, I noticed parallelism wins start to taper off at thread count 30 to 40. When I bump to 80 to use all machine cores, I saw no improvements to wall time.
> 
> Based on that, I would suggest that we initially set debug-thread-count to 32 by default. What do you think?

This if for second PR in the stack?

> (this one more serious) II. Data race in parallel DWO processing when using DWP files: when multiple buckets process CUs whose DWO units reside in the same DWP file, their DWODIEBuilder instances call buildDWOUnit(), which calls buildTypeUnits() concurrently on the same DWARFContext. DWARFContext performs lazy parsing internally and methods like getTUIndex() populate internal caches on first access. These lazy mutations are not thread-safe by default. The result is a sporadic SIGSEGV in buildTypeUnits, reproducible by running the dwarf5-df-types-dup-dwp-input.test test in a loop.

I think this is an existing issue. https://github.com/llvm/llvm-project/pull/197359

rafaelauler:
> > @ayermolo thanks for helping review!
> > @Thrrreeee I have two more bugs that I found that needs fixing, but hopefully these should be the last ones and, fixing those locally I got it to pass all of my internal tests too.
> > I. Corruption in empty ranges offset (debug_ranges): during the parallel merge phase, we merge per-bucket range buffers into a global .debug_ranges section. During merge, fixupDWARFRanges shifts every DW_AT_ranges value by the global buffer's current size. However, offset 0 is reserved as the empty ranges terminator (see DebugRangesSectionWriter::EmptyRangesOffset). initSection() writes a 16-byte zero terminator at offset 0 in the global buffer. CUs with no address ranges (in my internal tests, those came from crtn.S, for example) get DW_AT_ranges=0 pointing to this terminator. Because my crtn.S was among the last DIEs to be processed, shifting that 0 by the global offset produced an invalid offset pointing at the end of .debug_ranges (because the debug_ranges was already fully processed and in its final size), causing "invalid range list offset" errors in llvm-dwarfdump, failing my test.
> > To fix that, we can skip the fixup when DW_AT_ranges equals the empty ranges offset, since offset 0 is valid in the global buffer as-is, and call initSection() on local per-bucket LegacyRangesWriters to reserve offset 0 as the empty terminator, matching the global buffer convention. Without this, real range data starts at offset 0 in local buffers, colliding with the empty ranges sentinel. When appending the local buffer to the global buffer during merge, we also need to skip the local buffer's 16-byte terminator prefix to avoid duplicating terminators in the global section, and adjust the fixup offset accordingly. Here's an example of the changes I had locally for that (feel free to write something different, this is just what I used to unblock me): https://pastebin.com/dGvE9xjE
> > (this one more serious) II. Data race in parallel DWO processing when using DWP files: when multiple buckets process CUs whose DWO units reside in the same DWP file, their DWODIEBuilder instances call buildDWOUnit(), which calls buildTypeUnits() concurrently on the same DWARFContext. DWARFContext performs lazy parsing internally and methods like getTUIndex() populate internal caches on first access. These lazy mutations are not thread-safe by default. The result is a sporadic SIGSEGV in buildTypeUnits, reproducible by running the dwarf5-df-types-dup-dwp-input.test test in a loop.
> > This was fixable as DWARFContext already supports a ThreadSafe=true mode (see comment on DWARFContextState in DWARFContext.h) that protects lazy initialization with internal locking. BOLT currently creates its DWARFContext with the default ThreadSafe=false. Passing ThreadSafe=true in RewriteInstance.cpp at the DWARFContext::create() call fixes the race with no measurable performance overhead in my testing.
> > If that lock is a problem for you, feel free to suggest an alternative.
> > III. Discussion on default value for DebugThreadCount: I did several experiments with internal apps where I detected no significant difference in memory usage, but a decent 20 to 30% improvement in BOLT wall time. In a machine with 80 cores, I noticed parallelism wins start to taper off at thread count 30 to 40. When I bump to 80 to use all machine cores, I saw no improvements to wall time.
> > Based on that, I would suggest that we initially set debug-thread-count to 32 by default. What do you think?
> 
> This if for second PR in the stack?

Yes, I tested both PRs together, so these comments apply to PR #197859, sorry for the confusion.

rafaelauler:
>  I think this is an existing issue. https://github.com/llvm/llvm-project/pull/197359

Awesome! So it is already tracked. Funny, I used ThreadSafe DWARFContext and couldn't repro anymore the race. But maybe it is still there then. Thanks for linking the PR

Thrrreeee:
> Here's an example of the changes I had locally for that (feel free to write something different, this is just what I used to unblock me): https://pastebin.com/dGvE9xjE
> 

Thanks for the fix, this looks great. I'll adopt it in my code.

> This was fixable as DWARFContext already supports a ThreadSafe=true mode (see comment on DWARFContextState in DWARFContext.h) that protects lazy initialization with internal locking. BOLT currently creates its DWARFContext with the default ThreadSafe=false. Passing ThreadSafe=true in RewriteInstance.cpp at the DWARFContext::create() call fixes the race with no measurable performance overhead in my testing.

You're right that setting `ThreadSafe=true` here can avoid the `getTUIndex`  data race. The https://github.com/llvm/llvm-project/pull/197359 mentioned by @ayermolo  is actually a bug I ran into during my implementation as well — it stems from the current deficiencies in DWP's multi-threaded handling.

> Based on that, I would suggest that we initially set debug-thread-count to 32 by default. What do you think?

32 threads is a reasonable choice. I'll add an additional check to make sure the thread count doesn't exceed the CPU cores — that level of concurrency tends to be a good balance.

Thrrreeee:
> Awesome! So it is already tracked. Funny, I used ThreadSafe DWARFContext and couldn't repro anymore the race. But maybe it is still there then. Thanks for linking the PR

The races occurred on lazily-initialized members of the same DWARFUnit object — specifically the DieArray, Abbrev, and other member variables.
Running it under TSan still produces data race warnings, such as follows:

> SUMMARY: ThreadSanitizer: data race /home/jinrui/llvm-pr/llvm/lib/DebugInfo/DWARF/DWARFUnit.cpp:1062:8 in llvm::DWARFUnit::getAbbreviations() const
>
> SUMMARY: ThreadSanitizer: data race /home/jinrui/llvm-pr/llvm/lib/DebugInfo/DWARF/DWARFUnit.cpp:507:31 in llvm::DWARFUnit::tryExtractDIEsIfNeeded(bool)

But it likely doesn't affect normal functionality？

ayermolo:
From review perspective regarding this PR, I think only one left is 
image

and then we can more rest of the conversation to second PR?

Thrrreeee:
> From review perspective regarding this PR, I think only one left is 

My bad — I didn't notice this yesterday.

> and then we can more rest of the conversation to second PR?

Of course, note that the second PR built on top of this one, so you can just review the last two commits. The earlier commits are the same as the code in this PR.

Thrrreeee:
> Awesome! So it is already tracked. Funny, I used ThreadSafe DWARFContext and couldn't repro anymore the race. But maybe it is still there then. Thanks for linking the PR

cc @rafaelauler, Only use ThreadSafe DWARFContext is not enough to fix this data race, [CI check](https://github.com/llvm/llvm-project/actions/runs/26615546502/job/78430325095?pr=197859#step:4:5776) in llvm lead to a race condition. After running the test case locally several times, I was able to reproduce the race condition intermittently. We may need to address the issue in https://github.com/llvm/llvm-project/pull/197359.

image

Maybe, could you also join the review of this PR and offer some input on the DWP handling implementation? 

This seems important for pushing the feature forward.

rafaelauler (approved):

ayermolo (approved):
OK, lets land this, rebase second PR and move rest of reviews to it.

Thrrreeee:
Thanks for review. @rafaelauler @ayermolo

PR #197675, opened 2026-05-14
[LoopInterchange] Bail out if outer latch has PHI with more than one incoming values

### LINKED ISSUE #194328: [LoopInterchange] Assertion failure when the outer loop latch has a PHI with more than one incoming value
Reproducer ([godbolt](https://godbolt.org/z/T5KhfvsxW)):

```llvm
define void @f(i1 %cond) {
entry:
  br label %outer.header

outer.header:
  %i = phi i64 [ 0, %entry ], [ %i.inc, %outer.latch ]
  br i1 %cond, label %outer.latch, label %inner

inner:
  %j = phi i64 [ 0, %outer.header ], [ %j.inc, %inner ]
  %j.inc = add i64 %j, 1
  %ec.j = icmp eq i64 %j.inc, 10
  br i1 %ec.j, label %outer.body, label %inner

outer.body:
  br label %outer.latch

outer.latch:
  %p = phi i64 [ %i, %outer.header ], [ 10, %outer.body ]
  %i.inc = add i64 %i, 1
  %ec.i = icmp eq i64 %i.inc, 10
  br i1 %ec.i, label %exit, label %outer.header

exit:
  ret void
}
```

Assertion error:

```
opt: /root/llvm-project/llvm/lib/Transforms/Scalar/LoopInterchange.cpp:2228: void simplifyLCSSAPhis(llvm::Loop*, llvm::Loop*): Assertion `Phi->getNumIncomingValues() == 1 && "Single input phi expected"' failed.
```

I'm not sure if the root cause is the same as #193746

kasuga-fj:
In the transformation phase of LoopInterchange, `simplifyLCSSAPhis` replaces all PHI nodes in the outer loop latch with LCSSA PHIs. This means that the transformation phase does not assume PHI nodes with more than one incoming value. However, the outer loop latch can have more than one predecessor, such as the outer loop header and the inner loop latch, like below:

```llvm
outer.header:
  br i1 %c, label %outer.latch, label %inner.header

inner.header:
  ...

inner.latch:
  ...
  br label %outer.latch

outer.latch:
  %p = phi [ ..., %outer.header ], [ ..., %inner.latch ]
```

In such a case, `simplifyLCSSAPhis` failed to handle the PHI node and crashed.

This patch addresses the issue by inspecting PHI nodes in the outer loop latch during the legality check phase and bailing out if such PHI nodes are found.

Fixes #194328


llvm/lib/Transforms/Scalar/LoopInterchange.cpp
```diff
@@ -1524,6 +1531,17 @@ bool LoopInterchangeLegality::canInterchangeLoops(unsigned InnerLoopId,
     return false;
   }
 
+  if (!areOuterLoopLatchPHIsSupported(OuterLoop)) {
+    LLVM_DEBUG(dbgs() << "Found unsupported PHI nodes in outer loop latch.\n");
```
sjoerdmeijer: Nit: maybe make the message a bit more descriptive, maybe just saying "Only outer loop phi nodes with 1 incoming value are supported".

llvm/lib/Transforms/Scalar/LoopInterchange.cpp
```diff
@@ -1421,6 +1421,13 @@ static bool areInnerLoopLatchPHIsSupported(Loop *OuterLoop, Loop *InnerLoop) {
   return true;
 }
 
+/// We currently only support PHI nodes in the outer loop latch if they have a
+/// single incoming value.
+static bool areOuterLoopLatchPHIsSupported(Loop *OuterLoop) {
```
sjoerdmeijer: Nit: I don't mind little helper functions, but this is just a one-liner and that could easily be moved/added to `canInterchangeLoops`. But either way is fine by me.

sjoerdmeijer (approved):
Thanks for fixing, LGTM.

llvm/lib/Transforms/Scalar/LoopInterchange.cpp
kasuga-fj: Adjusted the comments, thanks.

llvm/lib/Transforms/Scalar/LoopInterchange.cpp
kasuga-fj: Hm, I agree that was a bit excessive. I inlined it directly into `canInterchangeLoops`.

PR #197677, opened 2026-05-14
[DAG] SimplifyMultipleUseDemandedBits - fold (mul X, 1) -> X

RKSimon:
Use DemandedElts + KnownBits to match hidden identity patterns - helps especially with reduction patterns padded by legalisation

Once #197455 has landed, I'm intending to convert this (plus SMIN/SMAX/UMIN/UMAX and the existing ISD::ADD case) to use isIdentityElement directly.

### DIFF
--- a/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp
+++ b/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp
@@ -831,6 +831,16 @@ SDValue TargetLowering::SimplifyMultipleUseDemandedBits(
       return Op.getOperand(1);
     break;
   }
+  case ISD::MUL: {
+    RHSKnown = DAG.computeKnownBits(Op.getOperand(1), DemandedElts, Depth + 1);
+    if (RHSKnown.isConstant() && RHSKnown.getConstant().isOne())
+      return Op.getOperand(0);
+
+    LHSKnown = DAG.computeKnownBits(Op.getOperand(0), DemandedElts, Depth + 1);
+    if (LHSKnown.isConstant() && LHSKnown.getConstant().isOne())
+      return Op.getOperand(1);
+    break;
+  }
   case ISD::SHL: {
     // If we are only demanding sign bits then we can use the shift source
     // directly.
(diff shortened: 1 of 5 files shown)


llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp
```diff
@@ -831,6 +831,16 @@ SDValue TargetLowering::SimplifyMultipleUseDemandedBits(
       return Op.getOperand(1);
     break;
   }
+  case ISD::MUL: {
+    RHSKnown = DAG.computeKnownBits(Op.getOperand(1), DemandedElts, Depth + 1);
+    if (RHSKnown.isConstant() && RHSKnown.getConstant().isOne())
+      return Op.getOperand(0);
```
arsenm, line 837: Is it enough to try just the RHS?

arsenm (approved):

llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp
```diff
@@ -831,6 +831,16 @@ SDValue TargetLowering::SimplifyMultipleUseDemandedBits(
...
+
+    LHSKnown = DAG.computeKnownBits(Op.getOperand(0), DemandedElts, Depth + 1);
+    if (LHSKnown.isConstant() && LHSKnown.getConstant().isOne())
+      return Op.getOperand(1);
+    break;
+  }
```
arsenm, line 843: I thought this reported the resulting KnownBits, but apparently SimplifyMultipleUseDemandedBits doesn't have that out argument?

RKSimon: Correct - the SimplifyMultiple variants don't pass back any calculated KnownBits (of the DemandedBits) - I'm not sure we've any strong uses cases for this, it almost never creates nodes, just bypasses fallthrough patterns mainly.

llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp
RKSimon: Fraid not - its true RHS is much more likely to be a constant node, but the reason we're using computeKnownBits is because either side might have a hidden identity.

PR #197678, opened 2026-05-14
[PassBuilder] Add extra ModuleInliner/SROA/SimplifyCFG/LICM to late FullLTO passes

momchil-velikov:
After removing the pre-LTO vectorisation there are various missed optimisation opportunities which show up in benchmarks.

Add ModuleInliner, SROA, SimplifyCFG, and LICM late in LTO optimisation pipeline.

### DIFF
--- a/llvm/lib/Passes/PassBuilderPipelines.cpp
+++ b/llvm/lib/Passes/PassBuilderPipelines.cpp
@@ -2266,6 +2266,21 @@ PassBuilder::buildLTODefaultPipeline(OptimizationLevel Level,
   // Add late LTO optimization passes.
   FunctionPassManager LateFPM;
 
+  // Delete basic blocks, which optimization passes may have killed.
+  LateFPM.addPass(SimplifyCFGPass(SimplifyCFGOptions()
+                                      .convertSwitchRangeToICmp(true)
+                                      .convertSwitchToArithmetic(true)
+                                      .hoistCommonInsts(true)
+                                      .speculateUnpredictables(true)));
+
+  LateFPM.addPass(createFunctionToLoopPassAdaptor(
+      LICMPass(PTO.LicmMssaOptCap, PTO.LicmMssaNoAccForPromotionCap,
+               /*AllowSpeculation=*/true),
+      /*USeMemorySSA=*/true));
+
+  // Catch trivial redundancies
+  LateFPM.addPass(EarlyCSEPass(true /* Enable mem-ssa. */));
+
   // LoopSink pass sinks instructions hoisted by LICM, which serves as a
   // canonicalization pass that enables other optimizations. As a result,
   // LoopSink pass needs to be a very late IR pass to avoid undoing LICM
@@ -2283,6 +2298,7 @@ PassBuilder::buildLTODefaultPipeline(OptimizationLevel Level,
                                       .convertSwitchToArithmetic(true)
                                       .hoistCommonInsts(true)
                                       .speculateUnpredictables(true)));
+
   MPM.addPass(createModuleToFunctionPassAdaptor(std::move(LateFPM)));
 
   // Drop bodies of available eternally objects to improve GlobalDCE.
--- a/llvm/lib/Passes/PassBuilderPipelines.cpp
+++ b/llvm/lib/Passes/PassBuilderPipelines.cpp
@@ -2323,6 +2323,18 @@ PassBuilder::buildLTODefaultPipeline(OptimizationLevel Level,
   MPM.addPass(createModuleToFunctionPassAdaptor(std::move(MainFPM),
                                                 PTO.EagerlyInvalidateAnalyses));
 
+  if (EnableModuleInliner) {
+    MPM.addPass(ModuleInlinerPass(::getInlineParamsFromOptLevel(Level),
+                                  UseInlineAdvisor,
+                                  ThinOrFullLTOPhase::FullLTOPostLink));
+  } else {
+    MPM.addPass(ModuleInlinerWrapperPass(
+        ::getInlineParamsFromOptLevel(Level),
+        /* MandatoryFirst */ true,
+        InlineContext{ThinOrFullLTOPhase::FullLTOPostLink,
+                      InlinePass::CGSCCInliner}));
+  }
+
   // Lower type metadata and the type.test intrinsic. This pass supports
   // clang's control flow integrity mechanisms (-fsanitize=cfi*) and needs
   // to be run at link time if CFI is enabled. This pass does nothing if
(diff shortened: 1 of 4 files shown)


momchil-velikov:
Ping?

lukel97:
Can you get neoverse performance numbers on this individual change? I can get numbers for RISC-V.

fhahn (commented):
could you also provide compile-time numbers?

momchil-velikov:
> could you also provide compile-time numbers?

Of this patch only:

https://llvm-compile-time-tracker.com/compare.php?from=1b10cd91d12d783fe8754c100a3afaacfee3bf9f&to=d4b7fe233f7dd6d4eb0746dfe9bc13d907b77f75&stat=instructions:u

However, it's not supposed to be considered in isolation (it's a part of the overall set of "tweak the LTO pipeline" patches.

All the patches combined: https://llvm-compile-time-tracker.com/compare.php?from=0819314ff6838073505c4b4736400491960f72bc&to=41faccfb6ebae10f64da17c4f9d73135fc18e175&stat=instructions:u

momchil-velikov:
> Can you get neoverse performance numbers on this individual change? I can get numbers for RISC-V.

It's working, I'll post it sometime tomorrow.

momchil-velikov:
Benchmark | Change
-- | --
706.stockfish_r | -0.07%
707.ntest_r | -0.10%
708.sqlite_r | 0.06%
710.omnetpp_r | -1.23%
714.cpython_r | 2.52%
721.gcc_r | 0.17%
723.llvm_r | -0.10%
727.cppcheck_r | 0.57%
729.abc_r | -0.26%
734.vpr_r | 0.07%
735.gem5_r | 0.44%
750.sealcrypto_r | 0.02%
753.ns3_r | 0.01%
777.zstd_r | 0.35%
**Geomean** | **0.17%**

Benchmark | Change
-- | --
709.cactus_r | 0.02%
722.palm_r | 0.98%
731.astcenc_r | 0.06%
736.ocio_r | 0.09%
737.gmsh_r | -0.03%
748.flightdm_r | 0.12%
749.fotonik3d_r | -0.07%
765.roms_r | -0.17%
766.femflow_r | -1.41%
767.nest_r | -1.38%
772.marian_r | -0.03%
782.lbm_r | 0.19%
**Geomean** | **-0.14%**

lukel97 (commented):
RISC-V results on SPEC CPU 2017 look good https://lnt.lukelau.me/db_default/v4/nts/1630?compare_to=1631

llvm/lib/Passes/PassBuilderPipelines.cpp
```diff
@@ -2250,6 +2250,18 @@ PassBuilder::buildLTODefaultPipeline(OptimizationLevel Level,
...
+    MPM.addPass(ModuleInlinerPass(::getInlineParamsFromOptLevel(Level),
+                                  UseInlineAdvisor,
+                                  ThinOrFullLTOPhase::FullLTOPostLink));
+  } else {
+    MPM.addPass(ModuleInlinerWrapperPass(
+        ::getInlineParamsFromOptLevel(Level),
+        /* MandatoryFirst */ true,
+        InlineContext{ThinOrFullLTOPhase::FullLTOPostLink,
+                      InlinePass::CGSCCInliner}));
+  }
```
lukel97, line 2263: Can you elaborate more on why we need to run the module inliner again, and what regressions that helped with? Since this is already running after the vectorization passes

llvm/lib/Passes/PassBuilderPipelines.cpp
```diff
@@ -2266,6 +2278,20 @@ PassBuilder::buildLTODefaultPipeline(OptimizationLevel Level,
   // Add late LTO optimization passes.
   FunctionPassManager LateFPM;
 
+  LateFPM.addPass(SROAPass(SROAOptions::ModifyCFG));
```
lukel97, line 2281: There's already a SROA in addVectorPasses but its gated on `!isFullLTOPostLink`. Can we just enable it there, or does it have to be in the LateFPM?

llvm/lib/Passes/PassBuilderPipelines.cpp
```diff
@@ -2266,6 +2278,20 @@ PassBuilder::buildLTODefaultPipeline(OptimizationLevel Level,
...
+
+  // Delete basic blocks, which optimization passes may have killed.
+  LateFPM.addPass(SimplifyCFGPass(SimplifyCFGOptions()
+                                      .convertSwitchRangeToICmp(true)
+                                      .convertSwitchToArithmetic(true)
+                                      .hoistCommonInsts(true)
+                                      .speculateUnpredictables(true)));
+
```
lukel97, line 2289: Why do we need this SimplifyCFGPass given there's one after the DivRemPairsPass already?

momchil-velikov: This part of the patch (SimplifyCFG -> LICM -> CSE(GVN or EarlyCSE)) was added to recover about 1% regression in x264. CSE was later removed because overall it wasn't advantageous (improved some benchmarks regressed some, GVN in particular regressed exchange by 19%).

llvm/lib/Passes/PassBuilderPipelines.cpp
momchil-velikov: No, it is not running after the vectorisation passes when they are disabled in the pre-LTO pipeline.

A few hot functions in `femflow` contain big composites passed as arguments or returned as return values. In the current `main` pipeline pre-link SLP reduces the size of some functions, LTO inliner inlines them, and then SROA triggers, optimising several memset/memcpy calls. Without the pre-link SLP, the LTO inliner does not trigger, and then SROA does not trigger (due to `alloca` addresses escaping to non-inline callees) and then we get 9% regression.

Adding the Inliner + SROA after the LTO SLP recreates the above optimisation sequence, restoring about 8% of the regression.

llvm/lib/Passes/PassBuilderPipelines.cpp
momchil-velikov: This SROA instance is needed after then inliner.

momchil-velikov:
Ping?

rnk:
I am not an LTO pass pipeline expert, but... we shouldn't casually add a module inliner phase without some principle explanation beyond some benchmark results.

Can you elaborate on the nature of the inlining opportunities that are created by #195886?

Is it more like, loop unrolling and vectorization swing the inline cost heuristics, so not running them in pre-LTO causes us to skip inlining, so we have to make up for it with more post-LTO inlining?

momchil-velikov:
The reason was explained here: https://github.com/llvm/llvm-project/pull/197678#discussion_r3595123575

> Is it more like, loop unrolling and vectorization swing the inline cost heuristics, so not running them in pre-LTO causes us to skip inlining, so we have to make up for it with more post-LTO inlining?

It's kind of opposite, pre-LTO SLP creates LTO inlining opportunities. When pre-LTO SLP is removed, we lose the SLP -> Inliner phase sequence, this patch re-establishes it.

momchil-velikov:
@lukel97 Did https://github.com/llvm/llvm-project/pull/224766 resolve the RISC V regressions with all these patches?

Issue #195788, opened 2026-05-05
[clangd] null dereference in Selection.cpp::earlySourceRange on malformed pseudo-destructor during codeAction


hongtaihu:
On a commit built on yesterday, this crash could be reproduced. But in Sema, it could not be reproduced.
It involves sending a `codeAction` request to the C++ fragment below:
```cpp
struct A { ~A(); };
void b(const A *y) {
  y->~decltype(A())();
}
```

requests:
```
initialize
  - initialized
  - didOpen
  - codeAction
  - shutdown
  - exit
```

reproducer
```
python test.py /path/to/clangd
```

```python
#!/usr/bin/env python3
import json
import os
import subprocess
import sys
import tempfile
from pathlib import Path


SOURCE = """struct A { ~A(); };
void b(const A *y) {
  y->~decltype(A())();
}
"""

DEFAULT_TARGET = (
    "/home/ubuntu2404/lsp_fuzz_latest/tmp/llvm-project-main-20260504/"
    "build-asan-x86-relwithdebinfo/bin/clangd"
)
FINAL_WAIT = int(sys.argv[2]) if len(sys.argv) > 2 else 30


def make_msg(obj: dict) -> bytes:
    body = json.dumps(obj, ensure_ascii=False, separators=(",", ":")).encode()
    return f"Content-Length: {len(body)}\r\n\r\n".encode() + body


def main() -> int:
    target = sys.argv[1] if len(sys.argv) > 1 else DEFAULT_TARGET
    if not target:
        print("missing target clangd path", file=sys.stderr)
        return 2

    workdir = Path(tempfile.mkdtemp(prefix="id59_embedded_"))
    source_path = workdir / "main.hpp"
    source_path.write_text(SOURCE)
    uri = source_path.as_uri()

    line_count = len(SOURCE.splitlines())
    end_line = max(line_count, 0)

    initialize = {
        "jsonrpc": "2.0",
        "id": 0,
        "method": "initialize",
        "params": {
            "processId": None,
            "rootUri": workdir.as_uri() + "/",
            "capabilities": {
                "workspace": {"workspaceFolders": True},
                "textDocument": {
                    "codeAction": {
                        "dataSupport": True,
                        "codeActionLiteralSupport": {
                            "codeActionKind": {
                                "valueSet": ["quickfix", "refactor", "source"]
                            }
                        },
                    }
                },
            },
            "workspaceFolders": [
[...]

PR #197680, opened 2026-05-14
[clang] Add missing type source info of `decltype` specifier in destructor call to AST

zwuis:
The type source info will be stored in `MemberExpr::MemberDNLoc`.

This patch fixes the underlying issue of #195788.

### DIFF
--- a/clang-tools-extra/clangd/Selection.cpp
+++ b/clang-tools-extra/clangd/Selection.cpp
@@ -895,19 +895,13 @@ private:
     // rather than the TypeLoc nested inside it.
     // We still traverse the TypeLoc, because it may contain other targeted
     // things like the T in ~Foo<T>().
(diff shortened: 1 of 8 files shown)


cor3ntin (commented):
Should we partially revert the changes in https://github.com/llvm/llvm-project/issues/195788?

zwuis:
I want to remove all the related redundant null checks, including changes in #195788, in a separate PR.

shafik (commented):
This looks reasonable but it needs a release note.

cor3ntin:
> I want to remove all the related redundant null checks, including changes in https://github.com/llvm/llvm-project/issues/195788, in a separate PR.

It might be easier to do all in the same PR

cor3ntin (approved):
Thanks

zwuis:
Thanks for the review!

PR #197685, opened 2026-05-14
[CodeGenPrepare] Do not undo GEP unmerging in AddrSinkUsingGEPs

### LINKED ISSUE #197684: [CodeGenPrepare] Infinite loop due to GEP unmerging and AddrSinkUsingGEPs
Given the sample:
```llvm
target datalayout = "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-n32:64-S128-Fn32"
target triple = "arm64-apple-macosx12.0.0"

@sometable = global [2 x ptr] [ptr blockaddress(@func, %5), ptr blockaddress(@func, %4)]

define void @func() {
  %2 = alloca i8
  %3 = getelementptr i8, ptr %2, i64 16
  indirectbr ptr null, [label %4, label %5]

4:                                                ; preds = %1
  br label %6

5:                                                ; preds = %1
  call void @bar(ptr %3)
  ret void

6:                                                ; preds = %4
  %7 = getelementptr i8, ptr %2, i64 40
  store i32 0, ptr %7, align 8
  ret void
}

declare void @bar(ptr)
```
GEP Optimizations in CodeGenPrepare will cause an infinite loop during lowering: https://godbolt.org/z/svP17G1Tv .
The first optimization (gep unmerging across indirectbr) will unmerge the second gep as follows:
```llvm
  %7 = getelementptr i8, ptr %3, i64 24
``` 
The second optimization (AddrSinkUsingGEPs) will undo the optimization in order to sink the address calculation right before the store:
```llvm
 %sunkaddr = getelementptr i8, ptr %2, i64 40
 store i32 0, ptr %sunkaddr, align 8
```

With each transformation `optimizeBlock` will return true as a modification of the ir code occured, thus causing a never ending loop execution in the `while (MadeChange)` loop of `bool CodeGenPrepare::_run(Function &F)`.

MuellerMP:
Fixes #197684.

Do not sink gep calculation after it has been unmerged to prevent an infinite loop.


MuellerMP:
This makes use of instruction metadata to prevent the infinite loop. An alternative (altough maybe not as cost-effective) would be to check the preconditions of GEP unmerging before sinking. (e.g.: is the pointer operand also a gep, does it originate from a block with a indirectbr terminator and does it fullfill `GEPSequentialConstIndexed`)

PR #174474, opened 2026-01-05
[flang] Implement 'F_C_STRING' library function (Fortran 2023)

cenewcombe:
Implement `F_C_STRING` to convert a Fortran string to a C null-terminated string. Documented in F2023 Standard: 18.2.3.9 `F_C_STRING (STRING [, ASIS])`.

PR #171515, opened 2025-12-09
[Flang] Move builtin .mod generation into runtimes (Reapply #137828)

Meinersbur:
Reapplication of #137828 and #169638

Changes on top of #169638:

  * [`9e7ab48`](https://github.com/llvm/llvm-project/pull/171515/commits/9e7ab487945a04468469e82aab81cbf86a2da0d7): When using an offloading toolchain (amdgpu, nvptx, ...), the driver also adds the host platform's builtin module directory to the search path. For instance:
```sh
flang -fintrinsic-modules-path ../lib/clang/22/finclude/flang/amdgcn-amd-amdhsa  -fintrinsic-modules-path  ../lib/clang/22/finclude/flang/x86_64-unknown-linux-gnu
```
So if `omp_lib.mod` is not available in `amdgcn-amd-amdhsa/`, it can use the one in `x86_64-unknown-linux-gnu/`. This means existing build setups don't need to be changed (`-DLLVM_RUNTIME_TARGETS=amdgcn-amd-amdhsa -DLLVM_ENABLE_RUNTIMES=openmp`). `omp_lib.f90` is not specific to any platform (only has `#if __INTEL_COMPILER`).

 * Split `RUNTIMES_FLANG_MODULES_ENABLED` option into `RUNTIMES_ENABLE_FORTRAN` and `RUNTIMES_FORTRAN_MODULES`.
 
   `RUNTIMES_ENABLE_FORTRAN` controls whether Fortran sources can be compiled to object (and module) files while `RUNTIMES_FORTRAN_MODULES` determines whether these can be used by Flang. There are two reasons for this change: 
   (1) Previously openmp had code to compile modules for gfortran and that possibility should be preserved. In that case modules files should not be installed into Flang's module path.
   (2) Make building the library (`*_ENABLE_STATIC`/`*_ENABLE_SHARED`) and the module (`*_FORTRAN_MODULES`) configurable separately. `RUNTIMES_FORTRAN_MODULES` is now the default value for  `LIBOMP_FORTRAN_MODULES`[^1] and `FLANG_RT_FORTRAN_MODULES`

[^1]: `LIBOMP_FORTRAN_MODULES` already exists before this PR and the other names are defined from that

[...]

PR #170374, opened 2025-12-02
[OpenMP] Add Fortran support for omp_* functions needing adapters

ro-i:
Taking omp_get_uid_from_device and omp_get_device_from_uid as examples.

OpenMP 6.0 defines these functions as follows:
```
const char *omp_get_uid_from_device(int device_num);

character(:) function omp_get_uid_from_device(device_num)
  pointer :: omp_get_uid_from_device
  integer, intent(in) :: device_num
```
and
```
int omp_get_device_from_uid(const char *uid);

integer function omp_get_device_from_uid(uid)
  character(len=*), intent(in) :: uid
```

As a result, we cannot directly map the Fortran API declarations to the C API declarations. We need some adapter code to do the conversion.

However, this means that we cannot just define the two functions as omp_* in kmp_ftn_entry.h without clashing with the adapter code.

The current situation is less than ideal, as demonstrated by this draft PR, which shall serve as a base for discussing different solution approaches.

---------------

@mjklemm as discussed

PR #197687, opened 2026-05-14
[flang-rt] Implement iso module in C++ in the runtime

jhuber6:
Summary:
Right now we have a very heavy dependency on `flang` for a single module
file. The flang build will handle creating the actual module file, we
just need to export the symbols. This PR ports the source file from the
flang .f90 to equivalent C++. This allows us to build `flang-rt`
standalone without bringing in the entire flang infrastructure, which is
by far the heaviest build in the entire LLVM project.

I verified the symbols and values against a previous build, so they are
identical.

### DIFF
--- a/flang-rt/lib/runtime/iso_fortran_env_impl.cpp
+++ b/flang-rt/lib/runtime/iso_fortran_env_impl.cpp
@@ -1,0 +1,303 @@
+//===-- lib/runtime/iso_fortran_env_impl.cpp --------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// Provides linkable symbols for the iso_fortran_env_impl module constants. It
+// replaces the Fortran source (flang/module/iso_fortran_env_impl.f90) to remove
+// the runtime library's build dependency on the Fortran compiler.
+//
+// The values here must stay in sync with the Fortran module source which
+// generates the .mod file during the flang compiler build.
+//
+// Symbol naming follows Flang's module variable mangling:
+//   _QM<module_name>EC<constant_name>
+//
+//===----------------------------------------------------------------------===//
+
+#include "flang/Common/float128.h"
+#include "flang/Common/real.h"
+#include "flang/Common/type-kinds.h"
+#include <cstdint>
+
+// Real kind 10 (x87 extended) is only available on x86-64.
+#if defined(__x86_64__) || defined(__amd64__) || defined(_M_X64)
+#define FLANG_RT_HAS_REAL_10 1
+#else
+#define FLANG_RT_HAS_REAL_10 0
+#endif
+
+// Real kind 16 (IEEE quad) requires a float128 runtime math library.
+// Must match the logic in flang/include/flang/Tools/TargetSetup.h.
+#if defined(FLANG_RUNTIME_F128_MATH_LIB) || HAS_LDBL128
+#define FLANG_RT_HAS_REAL_16 1
+#else
+#define FLANG_RT_HAS_REAL_16 0
+#endif
+
+// Fortran merge(tsource, fsource, mask) to match the source module style.
+static constexpr std::int32_t Merge(
+    std::int32_t tsource, std::int32_t fsource, bool mask) {
+  return mask ? tsource : fsource;
+}
+
+// Fortran selected_int_kind(R) returns the smallest integer kind with range >=
+// R digits. Flang integer kinds and their decimal ranges are fixed.
+static constexpr std::int32_t SelectedIntKind(int range) {
+  struct {
+    int kind, range;
+  } constexpr intKinds[]{{1, 2}, {2, 4}, {4, 9}, {8, 18}, {16, 38}};
+  for (auto [k, r] : intKinds) {
+    if (r >= range)
+      return k;
+  }
+  return -1;
+}
+
+// Fortran selected_real_kind(P, R) returns the smallest real kind with decimal
+// precision >= P and decimal range >= R. Returns negative per Fortran
+// 2023 16.9.170 when no kind qualifies.
+static constexpr std::int32_t SelectedRealKind(int p, int r) {
+  constexpr int realKinds[]{
+      2,
+      3,
+      4,
+      8,
+#if FLANG_RT_HAS_REAL_10
+      10,
+#endif
+#if FLANG_RT_HAS_REAL_16
+      16,
+#endif
+  };
+  for (int k : realKinds) {
+    Fortran::common::RealCharacteristics rc{
+        Fortran::common::PrecisionOfRealKind(k)};
+    if (rc.decimalPrecision >= p && rc.decimalRange >= r)
+      return k;
+  }
+  bool pOK{false}, rOK{false};
+  for (int k : realKinds) {
+    Fortran::common::RealCharacteristics rc{
+        Fortran::common::PrecisionOfRealKind(k)};
+    if (rc.decimalPrecision >= p)
+      pOK = true;
+    if (rc.decimalRange >= r)
+      rOK = true;
+  }
+  if (pOK)
+    return rOK ? -4 : -2;
+  return rOK ? -1 : -3;
+}
+
+// Fortran digits() for integer kind K returns 8*K - 1 (excludes sign bit).
+static constexpr int IntDigits(int kind) { return 8 * kind - 1; }
+
+// Fortran digits() for unsigned kind K returns 8*K (no sign bit).
+static constexpr int UintDigits(int kind) { return 8 * kind; }
+
+// Fortran digits() for real kind K returns the binary precision.
+static constexpr int RealDigits(int kind) {
+  return Fortran::common::PrecisionOfRealKind(kind);
+}
+
+// The smallest valid real kind, used as a safe fallback when a selected
+// real kind is unavailable (so digits() can be called without error).
+static constexpr std::int32_t safeRealFallback{SelectedRealKind(0, 0)};
+
+// Integer kinds, selected -> safe -> validated.
+static constexpr std::int32_t selectedInt8{SelectedIntKind(2)};
+static constexpr std::int32_t selectedInt16{SelectedIntKind(4)};
+static constexpr std::int32_t selectedInt32{SelectedIntKind(9)};
+static constexpr std::int32_t selectedInt64{SelectedIntKind(18)};
+static constexpr std::int32_t selectedInt128{SelectedIntKind(38)};
+
+static constexpr std::int32_t safeInt8{
+    Merge(selectedInt8, SelectedIntKind(0), selectedInt8 >= 0)};
+static constexpr std::int32_t safeInt16{
+    Merge(selectedInt16, SelectedIntKind(0), selectedInt16 >= 0)};
+static constexpr std::int32_t safeInt32{
+    Merge(selectedInt32, SelectedIntKind(0), selectedInt32 >= 0)};
+static constexpr std::int32_t safeInt64{
+    Merge(selectedInt64, SelectedIntKind(0), selectedInt64 >= 0)};
+static constexpr std::int32_t safeInt128{
+    Merge(selectedInt128, SelectedIntKind(0), selectedInt128 >= 0)};
+
+static constexpr std::int32_t int8{Merge(
+    selectedInt8, Merge(-2, -1, selectedInt8 >= 0), IntDigits(safeInt8) == 7)};
+static constexpr std::int32_t int16{Merge(selectedInt16,
+    Merge(-2, -1, selectedInt16 >= 0), IntDigits(safeInt16) == 15)};
+static constexpr std::int32_t int32{Merge(selectedInt32,
+    Merge(-2, -1, selectedInt32 >= 0), IntDigits(safeInt32) == 31)};
+static constexpr std::int32_t int64{Merge(selectedInt64,
+    Merge(-2, -1, selectedInt64 >= 0), IntDigits(safeInt64) == 63)};
+static constexpr std::int32_t int128{Merge(selectedInt128,
+    Merge(-2, -1, selectedInt128 >= 0), IntDigits(safeInt128) == 127)};
+
+// Unsigned kinds, same selection as integer, validated with unsigned digits.
+static constexpr std::int32_t selectedUInt8{selectedInt8};
+static constexpr std::int32_t selectedUInt16{selectedInt16};
+static constexpr std::int32_t selectedUInt32{selectedInt32};
+static constexpr std::int32_t selectedUInt64{selectedInt64};
+static constexpr std::int32_t selectedUInt128{selectedInt128};
+
+static constexpr std::int32_t safeUInt8{safeInt8};
+static constexpr std::int32_t safeUInt16{safeInt16};
+static constexpr std::int32_t safeUInt32{safeInt32};
+static constexpr std::int32_t safeUInt64{safeInt64};
+static constexpr std::int32_t safeUInt128{safeInt128};
+
+static constexpr std::int32_t uint8{Merge(selectedUInt8,
+    Merge(-2, -1, selectedUInt8 >= 0), UintDigits(safeUInt8) == 8)};
+static constexpr std::int32_t uint16{Merge(selectedUInt16,
+    Merge(-2, -1, selectedUInt16 >= 0), UintDigits(safeUInt16) == 16)};
+static constexpr std::int32_t uint32{Merge(selectedUInt32,
+    Merge(-2, -1, selectedUInt32 >= 0), UintDigits(safeUInt32) == 32)};
+static constexpr std::int32_t uint64{Merge(selectedUInt64,
+    Merge(-2, -1, selectedUInt64 >= 0), UintDigits(safeUInt64) == 64)};
+static constexpr std::int32_t uint128{Merge(selectedUInt128,
+    Merge(-2, -1, selectedUInt128 >= 0), UintDigits(safeUInt128) == 128)};
+
+// Logical kinds mirror integer kinds.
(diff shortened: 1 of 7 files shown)


flang-rt/lib/runtime/iso_fortran_env_impl.cpp
Jason-Van-Beusekom, line 1: Could we add a lit test to verify that this and `flang/module/iso_fortran_env_impl.f90` to ensure that the files match?

jhuber6:
libc++ test failure is unrelated, everything is green.

clementval:
Will the dependency come back anyway once @Meinersbur move the module files to the runtime?

flang-rt/lib/runtime/iso_fortran_env_impl.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+extern const std::int32_t FORTRAN_NAMED_CONST(real16){real16};
+extern const std::int32_t FORTRAN_NAMED_CONST(bfloat16){bfloat16};
+extern const std::int32_t FORTRAN_NAMED_CONST(real32){real32};
+extern const std::int32_t FORTRAN_NAMED_CONST(real64){real64};
+extern const std::int32_t FORTRAN_NAMED_CONST(real80){real80};
+extern const std::int32_t FORTRAN_NAMED_CONST(real64x2){real64x2};
+extern const std::int32_t FORTRAN_NAMED_CONST(real128){real128};
+
+extern const std::int32_t FORTRAN_NAMED_CONST(__builtin_integer_kinds)[]{
```
Jason-Van-Beusekom: This concerns me as __builtin_integer_kinds is declared dynamically in `iso_fortran_env_impl.f90` Would it be possible to mirror the logic from the fortran version here as well?

jhuber6:
> Will the dependency come back anyway once @Meinersbur move the module files to the runtime?

You are correct, so I have removed the part of the code that removes the flang setup and dependency. However, I believe this PR is still valuable. The incoming changes in https://github.com/llvm/llvm-project/pull/171515 will move module generation here, but this file is different as it creates real symbols. With this PR, we can logically split `flang-rt` compilation and Flang module generation. That means we can compile the entire runtime stack without a fortran compiler. @Meinersbur's PR already sets up a separate target for module file generation.

The main motivating factor here is that internally at AMD we are struggling with the extremely heavy resource requirements for building flang, so the effort is to contain it as much as possible. We are going to limit the dependency graph such that we can build as much as possible without `flang` so we can do it at the end or separately.

flang-rt/lib/runtime/iso_fortran_env_impl.cpp
jhuber6: Good point, those used to be implicitly set by the compiler interface. I've modified the header the compiler uses to initialize it. The REAL case is complicated due to variable support for f80 and f128, but I just added a static assert so it will loudly fail if someone changes it without updating.

flang-rt/lib/runtime/iso_fortran_env_impl.cpp
jhuber6: It's possible, I'll try to add one.

flang-rt/lib/runtime/iso_fortran_env_impl.cpp
```diff
@@ -0,0 +1,303 @@
...
+//   _QM<module_name>EC<constant_name>
+//
+//===----------------------------------------------------------------------===//
+
+#include "flang/Common/float128.h"
+#include "flang/Common/real.h"
+#include "flang/Common/type-kinds.h"
+#include <cstdint>
+
+// Real kind 10 (x87 extended) is only available on x86-64.
```
Meinersbur:
```suggestion
/// Real kind 10 (x87 extended) is only available on x86-64.
```
Consider using doxygen comments for when adding a doxygen support for flang-rt.

And yes, LLVM doxyegen generator those even for .cpp-private [members](https://llvm.org/doxygen/APInt_8cpp.html#af853fcc5930f7653d40c2637dc728bb5) and [macros](https://llvm.org/doxygen/Dwarf_8cpp.html#a52cbb26ec5bb928e366fe1965e8bf60e)

flang-rt/lib/runtime/iso_fortran_env_impl.cpp
```diff
@@ -0,0 +1,303 @@
...
+        Fortran::common::PrecisionOfRealKind(k)};
+    if (rc.decimalPrecision >= p)
+      pOK = true;
+    if (rc.decimalRange >= r)
+      rOK = true;
+  }
+  if (pOK)
+    return rOK ? -4 : -2;
+  return rOK ? -1 : -3;
+}
```
Meinersbur: This is copy&pasted from Flang. Why keep the boilerplate complexity and not directly encoding the data as any other [header](https://github.com/llvm/llvm-project/blob/main/clang/lib/Headers/stdint.h) would do?

If you think there is value in keeping resembling this to the flang equivalent, why not move the relavant parts of https://github.com/llvm/llvm-project/blob/main/flang/include/flang/Evaluate/target.h into a common header (there is https://github.com/llvm/llvm-project/tree/main/flang/include/flang/Common for this purpose). In no case would I look forward into duplicated code.

There should also be an `#error` for any configuration not yet supported, so it does not get forgetting when adding new supported targets.

flang-rt/lib/runtime/iso_fortran_env_impl.cpp
```diff
@@ -0,0 +1,303 @@
...
+// Real kind 10 (x87 extended) is only available on x86-64.
+#if defined(__x86_64__) || defined(__amd64__) || defined(_M_X64)
+#define FLANG_RT_HAS_REAL_10 1
+#else
+#define FLANG_RT_HAS_REAL_10 0
+#endif
+
+// Real kind 16 (IEEE quad) requires a float128 runtime math library.
+// Must match the logic in flang/include/flang/Tools/TargetSetup.h.
+#if defined(FLANG_RUNTIME_F128_MATH_LIB) || HAS_LDBL128
```
Meinersbur: [not a change request] Will be `FORTRAN_SUPPORTS_REAL16` after 171515

flang-rt/lib/runtime/iso_fortran_env_impl.cpp
jhuber6: Doxygen headers are used inconsistently in `flang-rt`, I didn't really know what the accepted style was. I can add them here.

flang-rt/lib/runtime/iso_fortran_env_impl.cpp
jhuber6: I'd lean towards re-using this if the relevant functions were header only, but they're in `target.cpp` so I'll just specify the real values. I leaned towards duplicating this to make the visual inspection easier, but it's definitely boilerplate.

Meinersbur:
> Will the dependency come back anyway once @Meinersbur move the module files to the runtime?

I was actually trying to go the other way. It seems obvious to be able to implement (parts of) a language's runtime in the language itself, in the sense that I don't know any other language that forces its runtime to be entirely written in another language. There have been several attemps already: #169525 (since the reason why `iso_fortran_env.f90` and `iso_fortran_env_impl.f90` were split have been removed), #174474 (changed during review to a C++ implementation), #170374 (openmp; solves a problem that some OpenMP functions/routines have the same name in C/Fortran, but different signatures).

With `iso_fortran_env_impl.f90` converted to C++, flang-rt could be split into two parts: Compilation of the machine code libraries, and build the module files. With #171515 building is modular: `FLANG_RT_ENABLE_STATIC`, `FLANG_RT_ENABLE_SHARED` can both be off, but still builds the module file. An option `FLANG_RT_ENABLE_MODULES` did not make sense so far because `iso_fortran_env(_impl).f90` was needed to be compiled in any case, and its module is a byproduct of it, but I would add such an option if the PR lands. So with `FLANG_RT_ENABLE_MODULES=OFF` there would be no dependency on Flang. We could even go so far to make `flang-rt-mod` a sibling "runtime" next to `flang-rt`. 

Of course, a complete Fortan toolchain, including running `check-flang` will still require the modules as well.

jhuber6:
> With `iso_fortran_env_impl.f90` converted to C++, flang-rt could be split into two parts: Compilation of the machine code libraries, and build the module files. With #171515 building is modular: `FLANG_RT_ENABLE_STATIC`, `FLANG_RT_ENABLE_SHARED` can both be off, but still builds the module file. An option `FLANG_RT_ENABLE_MODULES` did not make sense so far because `iso_fortran_env(_impl).f90` was needed to be compiled in any case, and its module is a side-output of it, but I would add such an option if the PR lands. So with `FLANG_RT_ENABLE_MODULES=OFF` there would be no dependency on Flang. We could even go so far to make `flang-rt-mod` a sibling "runtime" next to `flang-rt`.

This is my plan, basically. The changes in https://github.com/llvm/llvm-project/pull/198205 set up the dependencies so that we only need to build `flang` for `-mod*` runtime targets. I tested this in practice and I was able to set up the entire runtimes stack and then separately install the module files. That's pretty much all that's required to let us delay building `flang` only until users actually need it. The NFC in https://github.com/llvm/llvm-project/pull/198111  is also helpful as it brings `flang-rt` into a reasonable build time so I don't need to worry about building it as much.

vzakhari:
> I verified the symbols and values against a previous build, so they are identical.

Is there a way to verify that the C++ implementation and the Fortran implementation of the module provide "the same" binary automatically at least in some flang/flang-rt builds? I am worried about future changes causing misalignment between the two implementations.

jhuber6:
> Is there a way to verify that the C++ implementation and the Fortran implementation of the module provide "the same" binary automatically at least in some flang/flang-rt builds? I am worried about future changes causing misalignment between the two implementations.

There's not an easy way to do this as far as I'm aware. We'd need a separate test that compiles the `.f90` file with `flang` and then compares the values in `flang-rt`. It could be done, but it would require a lot of extra harness for CMake. The test I have now is a smoke test, and practically I expect the comments I added to hopefully be the real guard for this. This split will also be more obvious when the module files get moved into `flang-rt` I believe. Let me know if you think this is a blocker and I can try to figure something out, but I'd rather not.

vzakhari:
> Let me know if you think this is a blocker and I can try to figure something out, but I'd rather not.

I am okay with using the C++ implementation for the development purposes, but I think it will be better if we use the actual Fortran module for the complete toolchain build (i.e. the one that will probably be used for the official release builds). Will it be possible to only use the C++ implementation under an explicit CMake option?

jhuber6:
> I am okay with using the C++ implementation for the development purposes, but I think it will be better if we use the actual Fortran module for the complete toolchain build (i.e. the one that will probably be used for the official release builds). Will it be possible to only use the C++ implementation under an explicit CMake option?

That doesn't really solve my problem. For context, we at AMD are dealing with several issues that stem from the `flang` project's long compile times and massive RAM usage. The purpose of this PR and https://github.com/llvm/llvm-project/pull/198205 together is to let us defer the `flang` dependency as long as possible. These two PRs are my attempt to make this work for us without AMD taking the nuclear option and removing the flang build entirely.

We will still have the module file step, and https://github.com/llvm/llvm-project/pull/171515 will move that here as well (which will also help keep these in sync). It's just that the linkable symbols are in `flang-rt` which now becomes a C++ only library. I think this simplifies things greatly.

vzakhari:
Isn't it possible to have dependency on flang and the Fortran module in one configuration and do not have it in another?  What I am suggesting is to add a CMake option like `FLANG_RT_BUILD_WITHOUT_FLANG`: the configuration step will not use  the Fortran module and will not set dependency on flang, if this option is set. Otherwise, we will keep using the real Fortran module built by flang to get the corresponding binary file.

What am I missing?

jhuber6:
> Isn't it possible to have dependency on flang and the Fortran module in one configuration and do not have it in another? What I am suggesting is to add a CMake option like `FLANG_RT_BUILD_WITHOUT_FLANG`: the configuration step will not use the Fortran module and will not set dependency on flang, if this option is set. Otherwise, we will keep using the real Fortran module built by flang to get the corresponding binary file.
> 
> What am I missing?

Is the suggestion that we have two paths? I don't think it is worth the complexity. I will try to get a comprehensive test working so that we can prove the symbols always match up. I think that will ease a lot of anxiety.

Meinersbur:
> Isn't it possible to have dependency on flang and the Fortran module in one configuration and do not have it in another? What I am suggesting is to add a CMake option like `FLANG_RT_BUILD_WITHOUT_FLANG`: the configuration step will not use the Fortran module and will not set dependency on flang, if this option is set. Otherwise, we will keep using the real Fortran module built by flang to get the corresponding binary file.

I am not looking forward into maintaining two redundant versions. Generally, contributors also rarely think about updating configurations that they do not use. By experience, those go out of maintanance/grow inconsistencies over time.

For iso_fortran_env in particular, if this is about consistency with Flang's understanding of these kinds, I think a unittest would be the best approach.

jhuber6:
Added a test, it should fail if these every get out of sync.

vzakhari:
> I am not looking forward into maintaining two redundant versions. Generally, contributors also rarely think about updating configurations that they do not use. By experience, those go out of maintanance/grow inconsistencies over time.

Right, I have the same concern about the two implementations of the module symbols. Thank you for adding the test! LGTM

vzakhari (approved):

Jason-Van-Beusekom (approved):

jhuber6:
Going to go ahead and merge this. I don't think this precludes us from having Fortran sources in the future but I think we should think explicitly about how they're partitioned. This makes it easier to make the 99.9% C++ file 100% C++.

flang-rt/test/Driver/compare_iso_fortran_env_symbols.f90
```diff
@@ -0,0 +1,13 @@
...
+! RUN: %clang -x c++ -std=c++17 -c -I "%include" %S/../../lib/runtime/iso_fortran_env_impl.cpp -o %t.cpp.o
+
+! Extract defined symbol names and sizes from both objects.
+! RUN: llvm-nm --defined-only --format=posix --print-size %t.f90.o \
+! RUN:   | grep '_QMiso_fortran_env_impl' | cut -d' ' -f1,4 | sort > %t.f90.syms
+! RUN: llvm-nm --defined-only --format=posix --print-size %t.cpp.o \
+! RUN:   | grep '_QMiso_fortran_env_impl' | cut -d' ' -f1,4 | sort > %t.cpp.syms
+
+! The two symbol lists must be identical.
+! RUN: diff %t.f90.syms %t.cpp.syms
```
akuhlens, line 13: I am getting the following failure from this PR:
```
# RUN: at line 13
diff ...
# .---command stdout------------
# | *** compare_iso_fortran_env_symbols.f90.tmp.f90.syms
# | --- compare_iso_fortran_env_symbols.f90.tmp.cpp.syms
# | ***************
# | *** 1,6 ****
# |   _QMiso_fortran_env_implEC__builtin_integer_kinds 14
# |   _QMiso_fortran_env_implEC__builtin_logical_kinds 10
# | ! _QMiso_fortran_env_implEC__builtin_real_kinds 18
# |   _QMiso_fortran_env_implECbfloat16 4
# |   _QMiso_fortran_env_implECint128 4
# |   _QMiso_fortran_env_implECint16 4
# | --- 1,6 ----
# |   _QMiso_fortran_env_implEC__builtin_integer_kinds 14
# |   _QMiso_fortran_env_implEC__builtin_logical_kinds 10
# | ! _QMiso_fortran_env_implEC__builtin_real_kinds 14
# |   _QMiso_fortran_env_implECbfloat16 4
# |   _QMiso_fortran_env_implECint128 4
# |   _QMiso_fortran_env_implECint16 4
```

jhuber6: Ah, sorry, what's your platform? At least the test is doing its job.

jhuber6: My guess is that it might be a platform without 128bit float and for some reason the original ISO file says it's supported.

vzakhari: Our flang-rt is built with `-DFLANG_RT_SUPPORTS_REAL16=1`, so it contains the real-kind-16 in the array. The test's clang command does not specify this macro, so the real-kind-16 is missing from the `*.cpp.o` file.

vzakhari: I wonder if it is possible to compare the module compilation with the real `iso_fortran_env_impl.cpp.o` that was built during the library build.

jhuber6: Ah, that's a good point, I totally forgot about the compile overrides. We can probably extract it from the `.a`? But that'd require a static build

vzakhari: Can we do llvm-nm on the whole library (dynamic or static)?

vzakhari: `llvm-nm --defined-only --format=posix --print-size ./lib/clang/23/lib/x86_64-unknown-linux-gnu/libflang_rt.runtime.* | grep iso_fortran_env_impl` seems to work for shared and static libs

jhuber6: That's what I was thinking, but if we build with both it would double up all the symbols. Maybe I can just run it through `uniq`?

jhuber6: Let me know if https://github.com/llvm/llvm-project/pull/198922 fixes it

vzakhari: I guess we can check for `flang_rt.runtime`, `flang_rt.runtime.static` and `flang_rt.runtime.shared` targets, then find the output file for them and pass it to the LIT test from CMake, but `uniq` should probably also work.

vzakhari: > Let me know if https://github.com/llvm/llvm-project/pull/198922 fixes it

It works for me. Thanks!

PR #197690, opened 2026-05-14
[VPlan] Cost VPInstructionWithType with UV (NFC)

artagnon:
VPInstructionWithType with underlying-values can be costed: at the moment, there are no functional changes because they don't have underlying values.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
+++ b/llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
@@ -1629,6 +1629,13 @@ void VPInstructionWithType::execute(VPTransformState &State) {
   }
 }
 
+InstructionCost VPInstructionWithType::computeCost(ElementCount VF,
+                                                   VPCostContext &Ctx) const {
+  if (!getUnderlyingValue())
+    return 0;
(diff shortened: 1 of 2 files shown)


david-arm:
Hi, I've actually already done this here - https://github.com/llvm/llvm-project/pull/198291. It tries to deal with all cases (including those without an underlying value), or at the least the ones that are testable with current code paths.

david-arm:
I'm happy with either PR, but if we go this PR it would be good to handle the vscale case and have an assert for things that aren't scalar casts, since those are the only cases we can exercise right now.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1629,6 +1629,13 @@ void VPInstructionWithType::execute(VPTransformState &State) {
   }
 }
 
+InstructionCost VPInstructionWithType::computeCost(ElementCount VF,
+                                                   VPCostContext &Ctx) const {
+  if (!getUnderlyingValue())
+    return 0;
+  return getCostForRecipeWithOpcode(getOpcode(), VF, Ctx);
```
david-arm, line 1636: I think this means the PR isn't NFC.

PR #197679, opened 2026-05-14
[docs] Add the Clang Static Analysis WG to sync-ups

steakhal:
See https://discourse.llvm.org/t/rfc-forming-a-static-analysis-working-group-in-the-clang-ecosystem/90719/17

PR #197692, opened 2026-05-14
[docs] Add "LLVM Memory Safety" and "Lifetime Safety" working Groups

steakhal:


steakhal:
Addresses https://github.com/llvm/llvm-project/pull/197679#issuecomment-4451102703

steakhal:
I've also built this using `docs-llvm-html` and the resulting HTML looks good, builds without warnings.

AaronBallman (approved):
LGTM!

PR #196017, opened 2026-05-06
[GlobalISel] Skip match table for opcodes with no combines

c-rhodes:
Generate an opcode predicate for GICombiner matchers and use it to return from tryCombineAll before setting up matcher state and executing the match table.

The opcode list is collected from the generated rules, so the guard stays in sync with the match table and avoids match-table overhead for instructions the combiner cannot handle.

Improves CTMark geomean by -0.33% on stage1-aarch64-O0-g.

https://llvm-compile-time-tracker.com/compare.php?from=ed50ea52004259af958bb3e5636268342c49ee62&to=aea6e13cbc76c500a2e0aaedced716b9508811a7&stat=instructions%3Au

Also improves -O3 GISel geomean by -0.07%. Local results since this config isn't available on llvm-compile-time-tracker:
```
                 instructions:u                 diff
                            old           new
7zip               203476185395  203435124795 -0.02%
Bullet             103700835951  103694675411 -0.01%
ClamAV              52725697786   52671103456 -0.10%
SPASS               41611892042   41564657601 -0.11%
consumer-typeset    31700292250   31663189016 -0.12%
kimwitu++           39506799515   39520035914  0.03%
lencod              66874010678   66804617092 -0.10%
mafft               36550480158   36513382949 -0.10%
sqlite3             33262506697   33216990490 -0.14%
tramp3d-v4          76018141926   76024936414  0.01%
geomean             56941557756   56903973091 -0.07%
```
Assisted-by: codex

PR #197693, opened 2026-05-14
[GlobalISel] Filter combiner worklists

c-rhodes:
This follows on from #196017 which added an opcode predicate for GICombiner
matchers and used it to return from tryCombineAll before executing the match
table.

The better approach is to not add opcodes with no combines to the worklist in
the first place. This is a further -0.35% CTMark geomean improvement on
aarch64-O0-g, sqlite is -0.67%.

https://llvm-compile-time-tracker.com/compare.php?from=35f5d7ea802eae78b26a5fb2a46f072acd15f49d&to=c356bec46b68b59f37b26347f93676c9102d810c&stat=instructions%3Au

I also measured O3 locally and it's positive:

```
stage1-aarch64-O3 -fglobal-isel
                 instructions:u                 diff
                            old           new
7zip               203863583445  203865335443  0.00%
Bullet             103920943623  103917036315 -0.00%
ClamAV              53008970261   52971646424 -0.07%
SPASS               41843361245   41815471487 -0.07%
consumer-typeset    31848080935   31816255572 -0.10%
kimwitu++           39779892082   39733563964 -0.12%
lencod              67158153709   67134337444 -0.04%
mafft               36689364945   36675438948 -0.04%
sqlite3             33436727535   33410098701 -0.08%
tramp3d-v4          76553056559   76569825676  0.02%
geomean             57208304686   57180409869 -0.05%
```

I am a bit worried about the incurred overhead of this for other opt levels as
we add more combines, but right now at least the data is positive.

Assisted-by: codex

### DIFF
--- a/llvm/include/llvm/CodeGen/GlobalISel/Combiner.h
+++ b/llvm/include/llvm/CodeGen/GlobalISel/Combiner.h
@@ -65,6 +65,13 @@ public:
   bool combineMachineInstrs();
 
 protected:
+  virtual bool canMatchOpcode(unsigned Opc) const { return true; }
+
+  /// Return false to skip adding \p MI to the initial worklist.
+  virtual bool shouldAddToWorkList(MachineInstr &MI) const {
+    return canMatchOpcode(MI.getOpcode());
+  }
+
   const CombinerInfo &CInfo;
   GISelChangeObserver &Observer;
   MachineIRBuilder &B;
--- a/llvm/include/llvm/CodeGen/GlobalISel/Combiner.h
+++ b/llvm/include/llvm/CodeGen/GlobalISel/Combiner.h
@@ -67,11 +67,6 @@ public:
 protected:
   virtual bool canMatchOpcode(unsigned Opc) const { return true; }
 
-  /// Return false to skip adding \p MI to the initial worklist.
-  virtual bool shouldAddToWorkList(MachineInstr &MI) const {
-    return canMatchOpcode(MI.getOpcode());
-  }
-
   const CombinerInfo &CInfo;
   GISelChangeObserver &Observer;
   MachineIRBuilder &B;
--- a/llvm/lib/CodeGen/GlobalISel/Combiner.cpp
+++ b/llvm/lib/CodeGen/GlobalISel/Combiner.cpp
@@ -297,7 +297,8 @@ bool Combiner::combineMachineInstrs() {
         // Erase dead insts before even adding to the list.
         if (EnableDCE && tryDCE(CurMI, MRI))
           continue;
-        WorkList.deferred_insert(&CurMI);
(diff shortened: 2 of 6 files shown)


llvm/include/llvm/CodeGen/GlobalISel/Combiner.h
```diff
@@ -65,6 +65,13 @@ class Combiner : public GIMatchTableExecutor {
   bool combineMachineInstrs();
 
 protected:
+  virtual bool canMatchOpcode(unsigned Opc) const { return true; }
+
+  /// Return false to skip adding \p MI to the initial worklist.
+  virtual bool shouldAddToWorkList(MachineInstr &MI) const {
+    return canMatchOpcode(MI.getOpcode());
+  }
```
aengelke: Instead of making a (two, actually) virtual function calls per instruction to query a purely opcode-dependent property, would there be any downside from generating a constant array? (The array could be computed in constexpr function to handle the AArch64O0PreLegalizerCombinerImpl case.)

arsenm:
>  but for combiners with handwritten combines we must also ensure these opcodes get added
to the worklist.

All the combines should have at least a tablegen definition. The handwritten stuff predates that and should be moved over

c-rhodes:
> > but for combiners with handwritten combines we must also ensure these opcodes get added
> > to the worklist.
> 
> All the combines should have at least a tablegen definition. The handwritten stuff predates that and should be moved over

I've slowly been working thru these, the last ones are the memcpy-like ops which will be complete once https://github.com/llvm/llvm-project/pull/203235 lands

arsenm:
> I've slowly been working thru these, the last ones are the memcpy-like ops which will be complete once #203235 lands

So this PR is obsolete then?

c-rhodes:
> > I've slowly been working thru these, the last ones are the memcpy-like ops which will be complete once #203235 lands
> 
> So this PR is obsolete then?

no it just means this PR can filter worklists solely with tablegen'ed `canMatchOpcode`, the compile-time improvement remains: https://llvm-compile-time-tracker.com/compare.php?from=0ad5d54be53f34f71e25575964785670720a5f4c&to=c09eb5098523682349ad78d53b46ba0334deeab7&stat=instructions%3Au

llvm/include/llvm/CodeGen/GlobalISel/Combiner.h
c-rhodes: that's fair, I've reduced it to a single virtual function call. Probably could be better as you say but I think given this is so simple and fits in with existing implementation it the right first pass.

cofibrant:
FWIW I think this is change very slick, I like it a lot

aemerson (approved):
LGTM, thanks.

PR #197694, opened 2026-05-14
[libc] Add some checks to the mmap wrapper

labath:
- check that the discarded offset bits (both high and low) are zero (page alignment is checked in the kernel, but this cannot be done values we discard for mmap2, nor for truncated values on 32-bit systems)
- check for negative offsets (the kernel interface uses unsigned values, but our off_t is signed)

One thing I'm not checking, but other implementations do, is the size of the allocation (after page alignment) fits into a ptrdiff_t. I didn't do that now as it requires figuring how to get (and whether to cache) the page size. This is mainly relevant for 32-bit systems as no 64-bit system will let you allocate 2^63 bytes of (virtual) memory.

### DIFF
--- a/libc/test/src/sys/mman/linux/mmap_test.cpp
+++ b/libc/test/src/sys/mman/linux/mmap_test.cpp
@@ -52,30 +52,27 @@ TEST_F(LlvmLibcMMapTest, FileOffsets) {
 
   // Check that we can map a file from offset zero. This succeeds even though
   // the file is empty.
-  void *addr = LIBC_NAMESPACE::mmap(nullptr, 47, PROT_READ,
-                                    MAP_ANONYMOUS | MAP_PRIVATE, fd, 0);
+  void *addr = LIBC_NAMESPACE::mmap(nullptr, 47, PROT_READ, MAP_PRIVATE, fd, 0);
   ASSERT_NE(addr, MAP_FAILED);
   ASSERT_ERRNO_SUCCESS();
   EXPECT_THAT(LIBC_NAMESPACE::munmap(addr, 47), Succeeds());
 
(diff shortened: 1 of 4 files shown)


libc/test/src/sys/mman/linux/mmap_test.cpp
```diff
@@ -41,3 +42,40 @@ TEST_F(LlvmLibcMMapTest, Error_InvalidSize) {
...
+  int fd = LIBC_NAMESPACE::memfd_create("mmap_test", MFD_CLOEXEC);
+  ASSERT_NE(fd, -1);
+  ASSERT_ERRNO_SUCCESS();
+  LIBC_NAMESPACE::cpp::scope_exit close_fd(
+      [&]() { ASSERT_THAT(LIBC_NAMESPACE::close(fd), Succeeds()); });
+
+  // Check that we can map a file from offset zero. This succeeds even though
+  // the file is empty.
+  void *addr = LIBC_NAMESPACE::mmap(nullptr, 47, PROT_READ,
+                                    MAP_ANONYMOUS | MAP_PRIVATE, fd, 0);
```
kaladron: Is the MAP_ANONYMOUS needed here if it's backed by a zero-length file?

kaladron (approved):

libc/test/src/sys/mman/linux/mmap_test.cpp
labath: No, in fact, it goes against what I was trying to achieve. :(

The checks don't really care about whether they are mapping a file or not, but I wanted to make the test case realistic.

I'm going to blame AI autocomplete on that one :P

PR #197695, opened 2026-05-14
[VPlan] Strip VPRecipeBase::isScalarCast (NFC)

artagnon:
This is done in preparation to consolidate more recipes into VPInstruction.


fhahn (approved):
LGTM, thanks

PR #197697, opened 2026-05-14
[LV] Fix analysis remarks leak

hassnaaHamdi:
Currently, `LoopVectorizeHints::vectorizeAnalysisPassName()` returns empty pass name when the user forces vectorization, which allows analysis remarks to be emitted implicitly. 
The code sometimes emits analysis remarks with passing explicit pass name and other times it uses the `vectorizeAnalysisPassName()` which prevents consistent behaviour.

This PR removes the implicit emitting so that analysis remarks consistently require `-pass-remarks-analysis` to print. Affected tests are updated to add `-pass-remarks-analysis`.


llvm/lib/Transforms/Vectorize/LoopVectorizationLegality.cpp
```diff
@@ -253,13 +253,7 @@ void LoopVectorizeHints::emitRemarkWithHints() const {
...
 
 const char *LoopVectorizeHints::vectorizeAnalysisPassName() const {
-  if (getWidth() == ElementCount::getFixed(1))
-    return LV_NAME;
-  if (getForce() == LoopVectorizeHints::FK_Disabled)
-    return LV_NAME;
-  if (getForce() == LoopVectorizeHints::FK_Undefined && getWidth().isZero())
-    return LV_NAME;
-  return OptimizationRemarkAnalysis::AlwaysPrint;
+  return LV_NAME;
```
david-arm: This is just a thought, but does this function still serve any purpose? I wonder if it's just easier to delete it completely and use LV_NAME instead? Alternatively, the `vectorizeAnalysisPassName` function could be moved outside the class and into a header file somewhere so that it just does:

```
const char *vectorizeAnalysisPassName() {
  return "loop-vectorize";
}
```

What do you think?

llvm/test/Transforms/LoopVectorize/VPlan/early_exit_with_stores_vplan.ll
```diff
@@ -1,5 +1,5 @@
 ; NOTE: Assertions have been autogenerated by utils/update_analyze_test_checks.py UTC_ARGS: --version 6
-; RUN: opt -p loop-vectorize -force-vector-width=4 -disable-output -vplan-print-after=printOptimizedVPlan %s 2>&1 | FileCheck %s
+; RUN: opt -p loop-vectorize -force-vector-width=4 -disable-output -vplan-print-after=printOptimizedVPlan %s 2>&1 | FileCheck --allow-empty %s
```
david-arm: Why is this needed? It sounds like `-vplan-print-after` isn't actually printing anything?

llvm/lib/Transforms/Vectorize/LoopVectorizationLegality.cpp
fhahn: Looking at the comment for the function, it seems like the AlwaysPrint was added intentionally, although I am not sure if the reason is still valid. It would be good to double-check if this also changes the behavior in Clang.

If changing the behavior is the right thing, the whole function can be removed and the pragma used instead


```
  /// If hints are provided that force vectorization, use the AlwaysPrint
  /// pass name to force the frontend to print the diagnostic.
  const char *vectorizeAnalysisPassName() const;
```

llvm/test/Transforms/LoopVectorize/VPlan/early_exit_with_stores_vplan.ll
hassnaaHamdi: Some analysis remarks were getting emitted by default, so there were some printing.

fhahn: But I think this just highlights that the tests are actually missing CHECK lines. Instead passing `--allow-empty`, it should have check lines

llvm/lib/Transforms/Vectorize/LoopVectorizationLegality.cpp
hassnaaHamdi: That work was added many years ago and it seems that it's not really used properly in the LV, the reports use the PassName very randomly.
Also it seems that clang tests got passed. I looked at the `Vectorisers.rst` no mention about that default emitting. I think that behaviour is only used in the LV and LoopDistribute.

llvm/test/Transforms/LoopVectorize/VPlan/early_exit_with_stores_vplan.ll
hassnaaHamdi: The related work is in-progress, that's why there are no CHECK lines ATM. There was an ask for having that test here: https://github.com/llvm/llvm-project/pull/183288#pullrequestreview-3855534214

llvm/lib/Transforms/Vectorize/LoopVectorizationLegality.cpp
fhahn: Right, so should  ` OptimizationRemarkAnalysis::AlwaysPrint` be completely remove if it is no longer needed/should no longer be used?

llvm/test/Transforms/LoopVectorize/VPlan/early_exit_with_stores_vplan.ll
fhahn: I see, maybe add `-debug` and check for the message preventing vectorization?

llvm/lib/Transforms/Vectorize/LoopVectorizationLegality.cpp
hassnaaHamdi: It's used in LoopDistribute.

fhahn: Yes but what I am thinking is if there's a reason the same reasoning would not apply there as well?

hassnaaHamdi: Same reasoning would apply there as well.
I may do that in a separate patch because it will include some tiny changes in the LoopDistribute pass and I need to add related people as reviewers.

llvm/test/Transforms/LoopVectorize/test-remarks-leak.ll
```diff
@@ -0,0 +1,40 @@
+; RUN: opt < %s -passes=loop-vectorize -force-vector-width=4 -S -pass-remarks-missed='loop-vectorize' 2>&1 | FileCheck %s -check-prefix=CHECK-MISSED-OPT-REMARK
+; RUN: opt < %s -passes=loop-vectorize -force-vector-width=4 -S -pass-remarks='loop-vectorize' 2>&1 | FileCheck %s -check-prefix=CHECK-REMARK
+
+; Test that only specified remarks get emitted
+
+; CHECK-MISSED-OPT-REMARK: remark: {{.*}} loop not vectorized
+
+; CHECK-REMARK-NOT: remark: {{.*}} loop not vectorized: Cannot determine whether critical uncountable exit load address does not alias with a memory write
```
david-arm, line 9: I think I'm probably missing something, but I'm not sure what this is testing? Is this testing that when using -pass-remarks='loop-vectorize' we don't emit a remark for the early exit test, but when using -pass-remarks-analysis='loop-vectorize' we do? If so I think it would be good to have a RUN line that also tests the positive case, i.e.

```
...
; RUN: opt < %s -passes=loop-vectorize -force-vector-width=4 -S -pass-remarks-analysis='loop-vectorize' 2>&1 | FileCheck %s -check-prefix=CHECK-REMARK
...
; CHECK-REMARK-ANALYSIS: remark: {{.*}} loop not vectorized: Cannot determine whether critical uncountable exit load address does not alias with a memory write
```

hassnaaHamdi: `-pass-remarks='loop-vectorize'` should emit only successful vectorizations.
I'm testing that it doesn't emit analysis reports, but before my changes, the test fail -it gets matched with analysis reports that it's not related to-.

david-arm: OK thanks for explaining! Could you add a comment explaining this, i.e. something like:

```
; NOTE: -pass-remarks should only report remarks for successful vectorization
; CHECK-REMARK-NOT: remark: {{.*}} loop not vectorized: Cannot determine whether critical uncountable exit load address does not alias with a memory write
```

llvm/lib/Transforms/Vectorize/LoopVectorizationLegality.cpp
fhahn: Doing in separate PR sounds good to me, as long as we make sure the behavior across passes is consistent, thanks

fhahn (approved):
LGTM, thanks.

I think the PR description needs updating before merging (mentions first commit, which will be gone after squash).

llvm/lib/Transforms/Vectorize/LoopVectorizationLegality.cpp
fhahn: @hassnaaHamdi did you already get a chance to look into if it is possible to also clean up the other passes that still use `AlwaysPrint` and if it would be possible to completely remove it?

PR #197698, opened 2026-05-14
[Coroutines] Allow rematerialization of unary operators and selected intrinsics

IgWod:
All of those can be cheaply recomputed when the coroutine has resumed.

Before this change, results of unary operators and intrinsics were spilled
into the coroutine frame and reloaded on resume:

```
  %neg = fneg float %n
  store float %neg, ptr %neg.spill.addr

  ; In resume:
  %neg.reload = load float, ptr %neg.reload.addr
  ; ... use %neg.reload
```

After this change, only the operand is spilled and the operation is
rematerialized on each resume, avoiding the frame store:

```
  store float %n, ptr %n.spill.addr

  ; In resume:
  %n.reload = load float, ptr %n.reload.addr
  %neg = fneg float %n.reload
  ; ... use %neg
```

Assisted-by: Claude Code

### DIFF
--- a/llvm/lib/Transforms/Coroutines/MaterializationUtils.cpp
+++ b/llvm/lib/Transforms/Coroutines/MaterializationUtils.cpp
@@ -15,6 +15,7 @@
 #include "llvm/IR/Dominators.h"
 #include "llvm/IR/InstIterator.h"
 #include "llvm/IR/Instruction.h"
+#include "llvm/IR/IntrinsicInst.h"
 #include "llvm/IR/ModuleSlotTracker.h"
 #include "llvm/Transforms/Coroutines/SpillUtils.h"
 #include <deque>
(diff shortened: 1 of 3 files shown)


llvm/test/Transforms/Coroutines/coro-materialize-intrinsics.ll
```diff
@@ -0,0 +1,447 @@
...
+; CHECK-LABEL: @f_fabs(
+; CHECK: malloc(i32 24)
+; CHECK-LABEL: @f_fabs_optnone(
+; CHECK: malloc(i32 32)
+; CHECK-LABEL: @f_ctpop(
+; CHECK: malloc(i32 24)
+; CHECK-LABEL: @f_ctpop_optnone(
+; CHECK: malloc(i32 32)
+; CHECK-LABEL: @f_floor(
+; CHECK: malloc(i32 24)
```
ChuanqiXu9, line 28: What are these checks testing for?

llvm/lib/Transforms/Coroutines/MaterializationUtils.cpp
```diff
@@ -231,8 +232,61 @@ static void rewriteMaterializableInstructions(
...
+    case Intrinsic::sadd_sat:
+    case Intrinsic::uadd_sat:
+    case Intrinsic::ssub_sat:
+    case Intrinsic::usub_sat:
+    // Integer min and max
+    case Intrinsic::smax:
+    case Intrinsic::smin:
+    case Intrinsic::umax:
+    case Intrinsic::umin:
+      return true;
```
ChuanqiXu9, line 283: I am not sure if this is an optimization as this function is basically a trade between time and space. But I am not sure if these functions are fast enough. If they are not fast enough, the change may not be good. I mean, do you have a benchmark?

ChuanqiXu9:
The commit message is better to add an example to tell people what is before the change and what is after change.

llvm/test/Transforms/Coroutines/coro-materialize-intrinsics.ll
IgWod: I've have picked one intrinsic from each group: "Floating-point unary math", "Floating-point binary math", etc. and mirrored tests from [coro-materialize.ll](https://github.com/llvm/llvm-project/pull/197698/changes#diff-2316cad7fba44e0299d121b8bb5825313096d9c7c83a4358a0d7802dddaa67a0). I guess, testing one intrinsic would be enough, so I'm happy to shrink the test.

llvm/lib/Transforms/Coroutines/MaterializationUtils.cpp
IgWod: I don't have any public benchmarks, but internally we've seen a performance uplift after adding those intrinsics. I appreciate though it may be architecture specific. I can always narrow down the scope of the PR to just add `UnaryOperator`, if that's the preference.

llvm/test/Transforms/Coroutines/coro-materialize-intrinsics.ll
ChuanqiXu9: The original test seems bad too... we should check these instructions won't be copied to coro_frames.

llvm/lib/Transforms/Coroutines/MaterializationUtils.cpp
ChuanqiXu9: I feel good to hear it is measured in a real workload. As this is a time/space trade, we can always construct the case we'd like to see. If this is measured in a real workaload and the change itself looks not bad, I feel good.

llvm/test/Transforms/Coroutines/coro-materialize-intrinsics.ll
IgWod: I've updated tests in https://github.com/llvm/llvm-project/pull/198571. Once merged I'll update this PR to match it.

IgWod:
I have now rebased the PR, updated tests and added an example to the description. I'm happy to split this PR into two: one adding the UnaryOperator and one adding the instrinsics if that's more desirable.

ChuanqiXu9 (approved):
FGTM

llvm/lib/Transforms/Coroutines/MaterializationUtils.cpp
```diff
@@ -231,8 +232,61 @@ static void rewriteMaterializableInstructions(
...
+
+  if (auto *II = dyn_cast<IntrinsicInst>(&V)) {
+    switch (II->getIntrinsicID()) {
+    // Floating-point unary math
+    case Intrinsic::fabs:
+    case Intrinsic::sqrt:
+    case Intrinsic::sin:
+    case Intrinsic::cos:
+    case Intrinsic::exp2:
+    case Intrinsic::log2:
```
NewSigma, line 248: I'm less convinced to the claim that these functions can be recomputed cheaply comparing to the others. Do you have some insights?

IgWod: For our uses those instructions are relatively cheap to recompute when compared with spilling and reloading, however this may be because GPUs can be better suited when it comes to computing sin, cos, etc. So I admit this can be quite architecture specific. Considering that, should we remove those 5 intrinsic from the list to make it more generic? Also are there any other ones on the list that need revising?

NewSigma: Thanks for the explanation. I imagine we need some target dependent checks if we receive any regression reports, but it looks good to me in its current form.

IgWod:
I didn't realize that the new test needs to be added to the xfail list for profcheck. It is fixed in #199650.

boomanaiden154:
Thanks. We really need to finish up actually fixing the coroutine tests so we don't keep running into this.

PR #195272, opened 2026-05-01
[BOLT][AArch64] Account for hugify alignment in AArch64 long jump layout

labrinea:
When --hugify is used for a PIE, the final section allocation in RewriteInstance::mapCodeSections aligns the address after the last non-cold text section before laying out the following sections:

```
  for (BinarySection *Section : CodeSections) {
    Address = alignTo(Address, Section->getAlignment());
    Section->setOutputAddress(Address);
    Address += Section->getOutputSize();

    if (opts::Hugify && !BC->HasFixedLoadAddress &&
        Section->getName() == LastNonColdSectionName)
      Address = alignTo(Address, Section->getAlignment());
  }
```

The AArch64 long-jump pass doesn't model that gap in its tentative layout, so a CBZ could be considered in range during stub insertion and later become out of range when JITLink applied the final layout.

This patch mirrors the hugify alignment before assigning cold fragment addresses so that range checks see the same hot-to-cold distance as the final layout.

Assisted-by: Codex

PR #197702, opened 2026-05-14
[BOLT] Define hugify option in CommandLineOpts.cpp

labrinea:
#195272 added a reference to opts::Hugify from LongJmp.cpp. This
broke shared-library builds because LLVMBOLTPasses does not link
against LLVMBOLTRuntimeLibs, where the option was previously defined.

Move the opts::Hugify definition to CommandLineOpts.cpp, alongside
the other global BOLT command-line options. This keeps the option
available to LLVMBOLTPasses through its existing LLVMBOLTUtils
dependency without adding a dependency on the runtime library layer.


paschalis-mpeis (approved):
Thanks Alexandros, feel free to merge.

PR #193706, opened 2026-04-23
[Support] Introduce driver for running tools as daemons

BStott6:
This PR introduces the general LLVM-side code for the [test daemonization project](https://discourse.llvm.org/t/rfc-reducing-process-creation-overhead-in-llvm-regression-tests/88612/9), including the implementation for the daemon IPC protocol. `lib/Support/DaemonDriver.cpp` implements the main function for the daemon, and `include/llvm/Support/ToolInterface.h` introduces the interface that tools can adopt to support daemonization.

Tools must implement the `LLVMTool` interface to support daemonization. A new utility target `ExampleDaemon` is also introduced, which is a faux tool used for testing the daemon IPC implementation - see `test/Support/DaemonDriver`.

See `llvm/docs/DaemonMode.rst` for more information. A working demo of this project for the `llvm/Transforms` tests can be found [here](https://github.com/BStott6/llvm-project/tree/new-daemonized-testing-demo).

PR #197704, opened 2026-05-14
 [Lit] Implement in-process built-in to interface with LLVM daemon tools

BStott6:
This PR is dependent on https://github.com/llvm/llvm-project/pull/196958 and https://github.com/llvm/llvm-project/pull/193706. The changes in the first commit in this PR are from those PRs, so please review from the 2nd commit only. I will rebase this branch when those PRs are merged.

This PR implements Lit's side of the daemon IPC mechanism, communicating with the daemon driver introduced by https://github.com/llvm/llvm-project/pull/193706. This works by introducing an in-process built-in which is responsible for sending the commands to the daemon to invoke it correctly and collecting its output. The `DaemonTool` class is responsible for managing a daemon process, restarting the process if needed. The `daemon_tool` module maintains a map of from tool names to `DaemonTool` instances which is how the instances are stored. For debugging purposes, `DaemonTool` responds to the `LIT_LLVM_TRACE_DAEMON_COMMUNICATION` environment variable, instructing it to print messages sent and received between Lit and the daemon process to Lit's stdout.

When the daemon is running a task, Lit sequentially reads all output on the stdout, stderr and status pipe of the daemon in a loop until a `returned` or `error` message is received. It is important that Lit sleeps a little in each iteration of the loop to avoid unnecessarily consuming CPU time and slowing down test execution, so Lit uses another thread to direct all output on the status pipe into a Queue object, which is read with a small timeout in the output collection loop - this ensures that Lit is asleep most of the time that the tool is running, but wakes as soon as there is a message on the status pipe. Lit reads the daemon's output continually while it is running a task rather than waiting until completion to read all of the output to avoid the situation where the pipe becomes full and the tool can't make progress. The reason we don't use the `select` module for polling the status pipe is that it only works with sockets on Windows.

The `DaemonError` and `UnexpectedDaemonOutput` exceptions mean something has gone wrong with the communication with the daemon. These errors should not be swallowed as they would indicate an issue that needs fixing.
 
I have also added some test cases where Lit invokes the ExampleDaemon introduced in https://github.com/llvm/llvm-project/pull/193706. The `llvm-daemon-integration` tests are instructed where to find the ExampleDaemon binary by an environment variable `LIT_LLVM_TOOLS_DIR` which is set by Lit's own `lit.cfg` if run from the LLVM tree. ExampleDaemon is added as a build dependency for `check-lit`.

### DIFF
--- a/llvm/CMakeLists.txt
+++ b/llvm/CMakeLists.txt
@@ -1380,6 +1380,7 @@ add_subdirectory(include)
 add_subdirectory(lib)
 
 if( LLVM_INCLUDE_UTILS )
+  add_subdirectory(utils/ExampleDaemon)
   add_subdirectory(utils/FileCheck)
   add_subdirectory(utils/PerfectShuffle)
   add_subdirectory(utils/count)
--- a/llvm/docs/DaemonMode.rst
+++ b/llvm/docs/DaemonMode.rst
@@ -1,0 +1,135 @@
+============================
+Running Tools in Daemon Mode
+============================
+
+.. contents::
+   :local:
+
+Introduction
+============
+
+Process creation incurs a significant overhead, especially on Windows. This
+document describes the "daemon mode" execution model for LLVM tools, where
+tools run as persistent daemon processes that can perform many tasks, with
+different arguments and input, in the same process. The original motivation
+for this work is to reduce Lit testing time by replacing tool invocations with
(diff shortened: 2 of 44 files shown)


BStott6:
The test failure is due to the previous PR and I think it will be solved by f8f122a85b762581a3af98ae014cd87465fe5d58. I am rebasing this branch to incorporate the fix.

PR #197705, opened 2026-05-14
[CIR] Correct 'body' of range-for scope.

erichkeane:
Without this, we fail the verifier due to lack of a terminator on the scope if there is a destructed type in the 'body', as the cleanup scope on the cir.scope object will insert the yield in the body (since that is the last place we needed a cleanup), rather than actually putting one at the end of the scope itself.

This patch just makes sure that "statement" (aka body) of the range-for happens in its own scope.  This is necessary because, as cpp reference says:

```
The scope of init-statement and the scope of condition are the same.
The scope of statement and the scope of expression are disjoint and nested within the scope of init-statement and condition.
```

We were not making them distinct, so we were doing cleanups incorrectly.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenStmt.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenStmt.cpp
@@ -936,6 +936,7 @@ CIRGenFunction::emitCXXForRangeStmt(const CXXForRangeStmt &s,
           bool useCurrentScope = true;
           if (emitStmt(s.getLoopVarStmt(), useCurrentScope).failed())
             loopRes = mlir::failure();
+          LexicalScope lexScope{*this, loc, builder.getInsertionBlock()};
           if (emitStmt(s.getBody(), useCurrentScope).failed())
             loopRes = mlir::failure();
           emitStopPoint(&s);
--- a/clang/lib/CIR/CodeGen/CIRGenStmt.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenStmt.cpp
@@ -919,8 +919,9 @@ CIRGenFunction::emitCXXForRangeStmt(const CXXForRangeStmt &s,
     // sure we handle all cases.
     assert(!cir::MissingFeatures::loopSpecificCleanupHandling());
     // https://en.cppreference.com/w/cpp/language/for
-    // Given: 
-    // for ( init-statement condition ﻿(optional) ; expression ﻿(optional) ) statement
+    // Given:
+    // for ( init-statement condition ﻿(optional) ; expression ﻿(optional) )
+    // statement
     forOp = builder.createFor(
         getLoc(s.getSourceRange()),
         /*condBuilder=*/
--- a/clang/lib/CIR/CodeGen/CIRGenStmt.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenStmt.cpp
@@ -920,8 +920,8 @@ CIRGenFunction::emitCXXForRangeStmt(const CXXForRangeStmt &s,
     assert(!cir::MissingFeatures::loopSpecificCleanupHandling());
     // https://en.cppreference.com/w/cpp/language/for
     // Given:
-    // for ( init-statement condition ﻿(optional) ; expression ﻿(optional) )
-    // statement
+    // for ( init-statement condition ﻿(optional) ; expression ﻿(optional)
+    // ) statement
     forOp = builder.createFor(
         getLoc(s.getSourceRange()),
         /*condBuilder=*/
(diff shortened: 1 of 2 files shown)


clang/lib/CIR/CodeGen/CIRGenStmt.cpp
```diff
@@ -918,7 +918,10 @@ CIRGenFunction::emitCXXForRangeStmt(const CXXForRangeStmt &s,
     // We probably already do the right thing because of ScopeOp, but make
     // sure we handle all cases.
     assert(!cir::MissingFeatures::loopSpecificCleanupHandling());
-
+    // https://en.cppreference.com/w/cpp/language/for
```
andykaylor, line 921: What is this comment supposed to be telling me?

clang/lib/CIR/CodeGen/CIRGenStmt.cpp
```diff
@@ -930,13 +933,13 @@ CIRGenFunction::emitCXXForRangeStmt(const CXXForRangeStmt &s,
...
-          // In C++ the scope of the init-statement and the scope of
-          // statement are one and the same.
+          // https://en.cppreference.com/cpp/language/range-for
+          // The scope of statement and the scope of expression are disjoint
+          // and nested within the scope of init-statement and condition.
           bool useCurrentScope = true;
           if (emitStmt(s.getLoopVarStmt(), useCurrentScope).failed())
             loopRes = mlir::failure();
-          if (emitStmt(s.getBody(), useCurrentScope).failed())
+          if (emitStmt(s.getBody(), /*useCurrentScope=*/false).failed())
```
andykaylor, line 942: This leaves `useCurrentScope` as a local variable with a single use (the call to `emitStmt` above. We should probably just pass a literal there too.

andykaylor (approved):
lgtm, with a couple of nits

clang/lib/CIR/CodeGen/CIRGenStmt.cpp
```diff
@@ -930,13 +933,13 @@ CIRGenFunction::emitCXXForRangeStmt(const CXXForRangeStmt &s,
         },
         /*bodyBuilder=*/
         [&](mlir::OpBuilder &b, mlir::Location loc) {
-          // https://en.cppreference.com/w/cpp/language/for
-          // In C++ the scope of the init-statement and the scope of
-          // statement are one and the same.
+          // https://en.cppreference.com/cpp/language/range-for
+          // The scope of statement and the scope of expression are disjoint
+          // and nested within the scope of init-statement and condition.
```
andykaylor, line 938: Classic codegen puts these in the same lexical scope. I read the cppreference page, and it wasn't clear to me where it was saying that these were in disjoint scopes.

clang/lib/CIR/CodeGen/CIRGenStmt.cpp
erichkeane: It shows the format of the 'for' loop, buti  should probably bring in range.

clang/lib/CIR/CodeGen/CIRGenStmt.cpp
erichkeane: Yeah, I am somewhat not sure about it either... 

Range For has the signature: 
`for ( init-statement ﻿(optional) item-declaration : range-initializer ) statement `

Which lowers to:
```
    auto&& /* range */ = range-initializer ﻿;
    auto /* begin */ = /* begin-expr */;
    auto /* end */ = /* end-expr */;
    for ( ; /* begin */ != /* end */; ++/* begin */)
    {

        item-declaration = */* begin */;
        statement
}
```
Regular for is:
```
for ( init-statement condition ﻿(optional) ; expression ﻿(optional) ) statement 		
```

`init-statement` is empty, `condition` is `begin != end`, and `expression` is `++begin`.

Also from C++ ref:
```
The scope of init-statement and the scope of condition are the same.
The scope of statement and the scope of expression are disjoint and nested within the scope of init-statement and condition.
```

1 is irrelevant of course, since there is no init-stmt.
2: I think this ACTUALLY says that BOTH of these should be in their own disjoint scopes (not just the 2nd one?), and 'inside' of the outer scope.  I'm not really sure what classic compiler is doing TBH.

So i actually am beginning to think that the correct answer is `useCurrentScope= false`?  But I'm a touch lost here.

erichkeane (commented):
Holding off on changes until we're more sure on implementation.

I suspect there is some split between Classic-codegen and reality that they get away with?

erichkeane:
I'm somewhat convinced this is wrong, so I'm just going to abandon this.  Andy is looking into cleanup scopes more overall, so hopefully this gets picked up in that, else it is a good test for him.

PR #197707, opened 2026-05-14
[libc] Annex K: Add constraint handler unit test class

vhscampos:
This unit test class will be useful for the tests related to Annex K. The functions in Annex K may call a constraint handler, so this new unit test class will facilitate the checks that the constraint handling mechanism is working as expected.

### DIFF
--- a/libc/test/UnitTest/ConstraintHandlerCheckingTest.h
+++ b/libc/test/UnitTest/ConstraintHandlerCheckingTest.h
@@ -1,0 +1,44 @@
+//===-- ConstraintHandlerCheckingTest.h ------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===---------------------------------------------------------------------===//
+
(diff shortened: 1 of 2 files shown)


vhscampos:
This PR depends on `set_constraint_handler_s` which is not yet merged.

libc/test/UnitTest/ConstraintHandlerCheckingTest.h
```diff
@@ -0,0 +1,50 @@
...
+#include "hdr/types/constraint_handler_t.h"
+#include "hdr/types/errno_t.h"
+#include "src/stdlib/set_constraint_handler_s.h"
+#include "test/UnitTest/Test.h"
+
+namespace {
+
+bool error_flag;
+
+void local_constraint_handler(const char *__restrict /*msg*/,
```
kaladron, line 30: It might be nice to note that this conforms to constraint_handler_t so that someone undersand why the signature is the way it is.

kaladron (approved):

libc/test/UnitTest/ConstraintHandlerCheckingTest.h
vhscampos: Added. Thanks!

vhscampos:
In fact, this PR doesn't really depend on anything because it is a header file not yet used anywhere. We can merge it without breaking builds.

kaladron (approved):

PR #197708, opened 2026-05-14
[PowerPC] Match intrinsics ppc_amo_st[dw]at with a pattern

redstar:
The intrinsics are 1:1 to the instructions except for the order of the operands, thus it is easy to match them with a pattern.

However, the intrinsics are defined as reading from and writing to memory, but the instructions explicitly set mayLoad to false. Looking at the ISA description it seems to me that the latter is not true. In any case, the side effect flags must be the same, otherwise the pattern is rejected.


lei137 (approved):
LTGM
Thx!

PR #197709, opened 2026-05-14
[libc] Annex K: strcpy_s

vhscampos:
This patch adds Annex K's `strcpy_s`.

### DIFF
--- a/libc/src/string/strcpy_s.cpp
+++ b/libc/src/string/strcpy_s.cpp
@@ -11,11 +11,9 @@
 ///
 //===----------------------------------------------------------------------===//
 
-// For RSIZE_MAX
 #define __STDC_WANT_LIB_EXT1__ 1
 #include "hdr/stdint_proxy.h"
 #undef __STDC_WANT_LIB_EXT1__
-
 #include "hdr/types/constraint_handler_t.h"
 #include "hdr/types/errno_t.h"
 #include "hdr/types/rsize_t.h"
@@ -55,7 +53,10 @@ LLVM_LIBC_FUNCTION(errno_t, strcpy_s,
   // Use s1max for the destination's length, not count + 1, because the
   // standard allows for overwriting the entire destination region, even if
   // s2's length is smaller than s1max.
-  else if (s2 < (s1 + s1max) && s1 < (s2 + count + 1)) {
+  else if (const uintptr_t s1_addr = reinterpret_cast<uintptr_t>(s1),
+           s2_addr = reinterpret_cast<uintptr_t>(s2);
+           s1_addr < s2_addr ? s2_addr - s1_addr < s1max
+                             : s1_addr - s2_addr < count + 1) {
     constraint_violation_msg = "strcpy_s: s1 and s2 cannot overlap";
   }
 
--- a/libc/src/string/strcpy_s.cpp
+++ b/libc/src/string/strcpy_s.cpp
@@ -1,0 +1,71 @@
+//===-- Implementation of strcpy_s ----------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// For RSIZE_MAX
+#define __STDC_WANT_LIB_EXT1__ 1
+#include "hdr/stdint_proxy.h"
+#undef __STDC_WANT_LIB_EXT1__
+
+#include "hdr/types/constraint_handler_t.h"
+#include "hdr/types/errno_t.h"
+#include "hdr/types/rsize_t.h"
+#include "src/__support/common.h"
+#include "src/__support/constraint_handler.h"
+#include "src/__support/macros/config.h"
+#include "src/string/memory_utils/inline_memcpy.h"
+#include "src/string/strcpy_s.h"
+#include "src/string/string_utils.h"
+#include "src/string/strnlen_s.h"
+
+#define ERRNO_T_FAIL 1
+
+namespace LIBC_NAMESPACE_DECL {
+
+LLVM_LIBC_FUNCTION(errno_t, strcpy_s,
+                   (char *__restrict s1, rsize_t s1max,
+                    const char *__restrict s2)) {
+  const char *constraint_violation_msg = 0;
+  size_t count;
+
+  if (s1 == 0) {
+    constraint_violation_msg = "strcpy_s: s1 is null";
+  } else if (s2 == 0) {
+    constraint_violation_msg = "strcpy_s: s2 is null";
+  } else if (s1max > RSIZE_MAX) {
+    constraint_violation_msg = "strcpy_s: s1max exceeds RSIZE_MAX";
+  } else if (s1max == 0) {
+    constraint_violation_msg = "strcpy_s: s1max is 0";
+  } else if (count = strnlen_s(s2, s1max);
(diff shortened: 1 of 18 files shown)


libc/src/string/strcpy_s.cpp
```diff
@@ -0,0 +1,75 @@
...
+///
+/// \file
+/// This file contains the implementation of strcpy_s.
+///
+//===----------------------------------------------------------------------===//
+
+#define __STDC_WANT_LIB_EXT1__ 1
+#include "hdr/stdint_proxy.h"
+#undef __STDC_WANT_LIB_EXT1__
+#include "hdr/types/constraint_handler_t.h"
```
kaladron, line 18: This looks like it's in #197707 - Maybe just merge that PR into this one.  It's small and hard to understand the utility of in isolation.

libc/src/string/strcpy_s.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "src/string/strcpy_s.h"
+#include "src/string/string_utils.h"
+#include "src/string/strnlen_s.h"
+
+namespace LIBC_NAMESPACE_DECL {
+
+LLVM_LIBC_FUNCTION(errno_t, strcpy_s,
+                   (char *__restrict s1, rsize_t s1max,
+                    const char *__restrict s2)) {
+  const char *constraint_violation_msg = 0;
```
kaladron:
```suggestion
  const char *constraint_violation_msg = nullptr;
```

Here and below)

libc/src/string/strcpy_s.cpp
```diff
@@ -0,0 +1,75 @@
...
+  size_t count;
+
+  if (s1 == 0) {
```
kaladron: I think this if chain might be better done in a function that returns cpp::expected or something similar.  Then instead of a series of else/if you could have

if (..) return "";

And make it clear that there you're not expecting multiple of these to be true and are just matching ont he first.

libc/src/string/strcpy_s.cpp
```diff
@@ -0,0 +1,75 @@
...
+  // standard allows for overwriting the entire destination region, even if
+  // s2's length is smaller than s1max.
+  else if (const uintptr_t s1_addr = reinterpret_cast<uintptr_t>(s1),
+           s2_addr = reinterpret_cast<uintptr_t>(s2);
+           s1_addr < s2_addr ? s2_addr - s1_addr < s1max
+                             : s1_addr - s2_addr < count + 1) {
+    constraint_violation_msg = "strcpy_s: s1 and s2 cannot overlap";
+  }
+
+  if (constraint_violation_msg) {
```
kaladron: Hmm.  Or I'm wondering if the _s wrappers are all called through something that effectively wraps these and checks as part of the return path?

It's hard to see because we're trying to get the infra up.  I wonder if the right answer is to implement the constraint helper, this function and one other in a single PR so that we can figure out what the common infra should look like.

libc/src/string/strcpy_s.cpp
```diff
@@ -0,0 +1,75 @@
...
+    if (s1 != 0 && s1max > 0 && s1max <= RSIZE_MAX) {
+      s1[0] = '\0';
+    }
+    libc_global_constraint_handler(constraint_violation_msg, 0, 1);
```
kaladron: From your other PR, it looks like the "1" here is intended to be the errno that is set?  If yes, we should use the constants.

libc/src/string/strcpy_s.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+  if (s1 == 0) {
+    constraint_violation_msg = "strcpy_s: s1 cannot be null";
+  } else if (s2 == 0) {
+    constraint_violation_msg = "strcpy_s: s2 cannot be null";
+  } else if (s1max > RSIZE_MAX) {
+    constraint_violation_msg = "strcpy_s: s1max cannot exceed RSIZE_MAX";
+  } else if (s1max == 0) {
+    constraint_violation_msg = "strcpy_s: s1max cannot be 0";
+  } else if (count = strnlen_s(s2, s1max);
```
kaladron: strnlen_s is a public entrypoint, we should use the internal function instead (unless there's a reason to)

kaladron (commented):
Thanks for working on this!  Some comments within.

libc/src/string/strcpy_s.cpp
vhscampos: The problem is that I have another PR that also depends on #197707.

I can still merge the two PRs if you want. Please let me know

libc/src/string/strcpy_s.cpp
vhscampos: Done

libc/src/string/strcpy_s.cpp
vhscampos: Done

libc/src/string/strcpy_s.cpp
vhscampos: I suggest we for now just do one constraint helper for each function, and if suitable we consider a refactor to extract out the helper.

libc/src/string/strcpy_s.cpp
vhscampos: Done

libc/src/string/strcpy_s.cpp
vhscampos: Done

libc/src/string/strcpy_s.cpp
kaladron: Nah it's all good.

libc/src/string/strcpy_s.cpp
```diff
@@ -0,0 +1,81 @@
...
+LLVM_LIBC_FUNCTION(errno_t, strcpy_s,
+                   (char *__restrict s1, rsize_t s1max,
+                    const char *__restrict s2)) {
+  const auto count_expected = validate(s1, s1max, s2);
+
+  if (!count_expected.has_value()) {
+    if (s1 != nullptr && s1max > 0 && s1max <= RSIZE_MAX) {
+      s1[0] = '\0';
+    }
+    libc_global_constraint_handler(count_expected.error(), 0,
```
kaladron, line 72: It looks like the zero here is another nullptr?

libc/src/string/strcpy_s.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "src/__support/CPP/expected.h"
+#include "src/__support/common.h"
+#include "src/__support/constraint_handler.h"
+#include "src/__support/macros/config.h"
+#include "src/string/memory_utils/inline_memcpy.h"
+#include "src/string/string_utils.h"
+
+namespace LIBC_NAMESPACE_DECL {
+
+LIBC_INLINE static cpp::expected<size_t, const char *>
```
kaladron, line 30: What do you think about returning an errno_t, const char* pair for the unexpected?

That would let you return specific error codes, like ERANGE for overflow or EINVAL on the nullpointers or 0 s1max?

libc/src/string/strcpy_s.cpp
kaladron: Actually, what is that pointer for?  The other PR just has "ptr" in it.

Should the line perhaps be "return libc_global_constraint (...)", and take a pair of errno_t and const char*?

I'm thinking the pair could typedef'd to appendix_k_error or something?

Since this is across 2 PRs (three if you count your other one) we can also commit this as it is, and then refactor when they're all landed.

kaladron: Resolving this - this is defined by constraint_handler_t.

PR #197710, opened 2026-05-14
[Matrix] Create inbounds GEPs for matrix load/stores.

fhahn:
LowerMatrixIntrinsics creates multiple loads/stores + GEPs for larger matrix load/stores. Those GEPs compute offsets into the memory accessed by the larger loads/stores, so those GEPs must be inbounds, otherwise the larger load would access memory out-of-bounds.


cofibrant (approved):
Sounds sensible!

farzonl (approved):

PR #197714, opened 2026-05-14
[NFC][lldb][windows] refactor switch case braces

charles-zablit:


lldb/source/Plugins/Process/Windows/Common/NativeProcessWindows.cpp
```diff
@@ -564,49 +563,47 @@ NativeProcessWindows::OnDebugException(bool first_chance,
...
     // For non-breakpoints, give the application a chance to handle the
     // exception first.
     if (first_chance)
       result = ExceptionResult::SendToApplication;
     else
       result = ExceptionResult::BreakInDebugger;
-  }
 
-  return result;
+    return result;
```
Nerixyz: `result` is only needed in this `default` branch. Its declaration could be moved inside.

Nerixyz (approved):

PR #197717, opened 2026-05-14
[lldb] Make PTY initialization idempotent

charles-zablit:
This patch ensures that the PTY initialization is idempotent by always resetting the PTY before creating it.

# Details

A `ProcessLaunchInfo` is reused across multiple launches and the underlying PTY is single-shot:
- On POSIX, `posix_openpt` would silently leak `m_primary_fd` if the same PTY were reused.
- On Windows, `PseudoConsole::OpenPseudoConsole` asserts when `m_mode != Mode::None`.

Reset the existing PTY (closing any previously-opened FDs/handles) so the next `Open*()` runs against a clean slate. Deliberately do NOT replace `m_pty` itself: callers (e.g. `ProcessWindows`, `ProcessLauncherWindows`, `PlatformQemuUser`) hold copies of the `shared_ptr` and may read it after the launch returns, so the object's identity has to survive `SetUpPtyRedirection`.

# Fixed tests

This patch fixes all the `types/*` tests when running  `check-lldb` with `LLDB_USE_LLDB_SERVER=1` on Windows.

rdar://177073941

### DIFF
--- a/lldb/include/lldb/Host/PseudoTerminal.h
+++ b/lldb/include/lldb/Host/PseudoTerminal.h
@@ -44,6 +44,11 @@ public:
   /// Close the secondary file descriptor if it is valid.
   void CloseSecondaryFileDescriptor();
 
+  /// Close both file descriptors and clear cached state. Used to recycle a
+  /// PseudoTerminal across multiple launches without invalidating any
+  /// outstanding shared_ptr references.
+  void Reset();
+
   /// Fork a child process that uses pseudo terminals for its stdio.
   ///
   /// In the parent process, a call to this function results in a pid being
--- a/lldb/include/lldb/Host/windows/PseudoConsole.h
+++ b/lldb/include/lldb/Host/windows/PseudoConsole.h
@@ -69,6 +69,13 @@ public:
   /// CreateProcessW to avoid keeping the pipes alive indefinitely.
   void CloseAnonymousPipes();
 
+  /// Closes any open ConPTY/pipe handles and resets internal state to a
+  /// freshly-constructed PseudoConsole. Useful when a ProcessLaunchInfo gets
+  /// reused for a second launch — the previous PseudoConsole would otherwise
+  /// trip the `m_mode == Mode::None` assert in OpenPseudoConsole /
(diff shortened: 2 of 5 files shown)


Nerixyz (approved):

JDevlieghere:
What warrants diverging this behavior between Windows and the other platforms? Without understanding what's causing the assert, I'm not sure how we can decide whether this solution is correct or papers over an existing issue on Linux.

charles-zablit:
> What warrants diverging this behavior between Windows and the other platforms? Without understanding what's causing the assert, I'm not sure how we can decide whether this solution is correct or papers over an existing issue on Linux.

I think there is a fundamental issue with the pty lifetime on both platforms. We re-use `ProcessLaunchInfo` for multiple launches, but we do not reset the pty for every code path. This is wrong on Windows, but also on macOS and Linux I found out:
- On POSIX, `posix_openpt` silently leaks `m_primary_fd` if the same PTY is reused. It assumes the `m_primary_fd` is moved to another consumer.
- On Windows, `PseudoConsole::OpenPseudoConsole` asserts when `m_mode != Mode::None` because we can't read from the same handle on 2 different threads.

The latest commits removes the divergence between Windows and POSIX by always recreating the PTY on all platforms.

Always recreating the PTY on Windows and POSIX works but it does not address the fundamental issue that a pty can be silently leaked.

JDevlieghere (approved):
Great, thanks for looking into this. This was the outcome I was hoping for!

charles-zablit:
It turns out that some tests were still failing on linux. The latest edition should fix them, however I would prefer to get a second review as the `Reset` pattern is close to the `Clear` pattern which we are moving away from. I don't think we can escape it until `ProcessLaunchInfo::Clear` is gone.

All platforms behave the same (no divergence).

Nerixyz:
Could the destructors of `PseudoTerminal` and `PseudoConsole` call `Reset()`? Their implementations are essentially the same (except for the `m_mode` assignment in `PseudoConsole`).

charles-zablit:
> PseudoConsole

Fixed 👍

Nerixyz (approved):

PR #197718, opened 2026-05-14
[flang][openacc] add optional extension to allow default none scalars

akuhlens:
Adds the flag, `-facc-allow-default-none-scalars`. When this flag is enabled, Flang reverts to the pre-3.2 behavior: scalar variables referenced inside a `default(none)` compute region without an explicit data clause do not produce an error.  Instead, Flang infers implicit data attributes for those scalars via the same implicit-copy logic applied in regions without `default(none)`.

- Adds tests and documentation.
- Makes an explicit extensions doc for OpenACC to mirror the OpenMP extensions doc.
- Moves the intentional deviations to from the standard to the extensions doc.


flang/docs/OpenACC-extensions.md
eugeneepshteyn, line 1: When you create a new document, you need to add it to index, otherwise the doc builder complains. See my commit 435957ac1c05a9fe7c21cea947d04774d60484c9 for an example.

flang/lib/Semantics/resolve-directives.cpp
eugeneepshteyn: Somehow formatting looks weird here: `isScalarExtensionActive` seems to be indented in such a way that it seems to be used after the block that defines it...  Could you please check the indents just in case?

flang/lib/Semantics/resolve-directives.cpp
```diff
@@ -1825,9 +1825,24 @@ void AccAttributeVisitor::Post(const parser::Name &name) {
           name.symbol = found;
         } else if (GetContext().defaultDSA == Symbol::Flag::AccNone) {
           // 2.5.14.
-          context_.Say(name.source,
-              "The DEFAULT(NONE) clause requires that '%s' must be listed in a data-mapping clause"_err_en_US,
-              symbol.name());
+          bool isScalarExtensionApplicable{false};
+          if (context_.IsEnabled(
+                  common::LanguageFeature::AccDefaultNoneScalars)) {
+            if (const auto *det{symbol.detailsIf<ObjectEntityDetails>()}) {
```
eugeneepshteyn: AI reviewer had concerns that OpenACC's definition of "scalar" may not completely match with the usual flang front-end's definition. Should the following code be accepted?
```
subroutine s()
  character(10) :: c
  type t; integer :: i; end type
  type(t) :: x
  integer, allocatable :: a
  !$acc parallel default(none)
    c = "hi"   ! Currently allowed by this PR
    x%i = 1    ! Same
    a = 1      ! Implicit firstprivate of allocatable: meaning differs from plain scalar
  !$acc end parallel
end
```
AI reviewer suggested to look at `IsOpenMPScalar()` implemented for OpenMP, whether something similar needs to be in OpenACC. 

I haven't researched this claim, but it seems to be worth investigating and perhaps consulting with OpenACC experts.

akuhlens: I checked the standard and the AI reviewer is correct fixed. If we want to add it as an extension later I am ok with that, but I will stick to the conservative extension here.

flang/lib/Semantics/resolve-directives.cpp
```diff
@@ -1825,9 +1844,17 @@ void AccAttributeVisitor::Post(const parser::Name &name) {
...
         } else if (GetContext().defaultDSA == Symbol::Flag::AccNone) {
           // 2.5.14.
-          context_.Say(name.source,
-              "The DEFAULT(NONE) clause requires that '%s' must be listed in a data-mapping clause"_err_en_US,
-              symbol.name());
+          if (context_.IsEnabled(
+                  common::LanguageFeature::AccDefaultNoneScalars) &&
+              IsAccScalar(symbol)) {
+            context_.Warn(common::UsageWarning::AccImplicitScalar, name.source,
+                "Implicit attribute inferred for DEFAULT(NONE) scalar '%s' (-facc-allow-default-none-scalars)"_warn_en_US,
```
jeanPerier: Is the `(-facc-allow-default-none-scalars)` needed here, I thought `context_.Warn` was adding that info automatically.

jeanPerier (approved):
Small nit, LGTM otherwise.

flang/docs/OpenACC-extensions.md
```diff
@@ -0,0 +1,61 @@
...
+
+```{contents}
+---
+local:
+---
+```
+
+The items below are intentional deviations from the OpenACC specification
+accepted by Flang.  They are also listed as deviations in
+[OpenACC.md](OpenACC.md).
```
tarunprabhu: I don't think these are listed in `OpenACC.md` any longer, are they?

tarunprabhu (approved):
Barring the comment on one of the documents, the docs and driver parts of this LGTM.

flang/lib/Semantics/resolve-directives.cpp
akuhlens: The warning hints a flag that turns on the warning. The flag that I am hinting here is the flag that enables the extension to open acc. I will remove this hint. Since it is already on and it is opt-in someone compiling this program knows about this flag.

flang/docs/OpenACC-extensions.md
akuhlens: Good catch. Thanks!

flang/include/flang/Support/Fortran-features.h
```diff
@@ -59,7 +59,8 @@ ENUM_CLASS(LanguageFeature, BackslashEscapes, OldDebugLines,
     PointerPassObject, MultipleIdenticalDATA,
     DefaultStructConstructorNullPointer, AssumedRankIoItem,
     MultipleProgramUnitsOnSameLine, AllocatedForAssociated,
-    OpenMPThreadprivateEquivalence, RelaxedCLoc, CudaPinned)
+    OpenMPThreadprivateEquivalence, RelaxedCLoc, CudaPinned,
+    AccDefaultNoneScalars)
```
eugeneepshteyn, line 63: Note: the other OpenACC enums defined here start with "OpenACC". Recommend renaming to `OpenAccDefaultNoneScalars`.

flang/include/flang/Support/Fortran-features.h
```diff
@@ -85,7 +86,7 @@ ENUM_CLASS(UsageWarning, Portability, PointerToUndefinable,
     RealConstantWidening, VolatileOrAsynchronousTemporary, UnusedVariable,
     UsedUndefinedVariable, BadValueInDeadCode, AssumedTypeSizeDummy,
     MisplacedIgnoreTKR, NamelistParameter, ImpureFinalInPure,
-    IgnoredNoReallocateLHS, CLoc, ExperimentalOption)
+    IgnoredNoReallocateLHS, CLoc, ExperimentalOption, AccImplicitScalar)
```
eugeneepshteyn, line 89: Similarly to above, recommend renaming to `OpenAccImplicitScalar`

flang/lib/Semantics/resolve-directives.cpp
```diff
@@ -1812,6 +1812,25 @@ void AccAttributeVisitor::Post(const parser::AccDefaultClause &x) {
   }
 }
 
+// Returns true iff symbol qualifies for the pre-OpenACC-3.2 DEFAULT(NONE)
+// scalar extension: an intrinsic numeric or logical non-array, non-pointer,
+// non-allocatable variable.  Characters, derived types, allocatables, and
+// pointers are excluded because their data-sharing semantics under
+// DEFAULT(NONE) differ materially from plain scalar copy.
+static bool IsAccScalar(const Symbol &symbol) {
```
eugeneepshteyn, line 1820: +1 for this function.

PR #197722, opened 2026-05-14
[InstCombine] Use `isa<ConstantPointerNull>` for null pointer checks

shiltian:
Make InstCombine pointer-null checks use `isa<ConstantPointerNull>` rather than generic zero checks, and fold inttoptr comparisons against the DataLayout null pointer value.


shiltian:
* #197722 👈
* `main`

<h2></h2>

llvm/lib/Transforms/InstCombine/InstructionCombining.cpp
shiltian, line 5131: This code makes me wonder whether we want to extend `ConstantPointerNull` to represent an array of null pointer? @nikic

arsenm: Probably not

arsenm (commented):
Should add tests with non-0 null pointer value

dtcxzyw:
BTW, I guess we should drop `m_Zero` and always use `m_ZeroInt`.

shiltian:
> m_Zero

Done.

shiltian:
> BTW, I guess we should drop `m_Zero` and always use `m_ZeroInt`.

In a separate PR?

llvm/test/Transforms/InstCombine/inttoptr-nullptr.ll
```diff
@@ -0,0 +1,26 @@
...
+; CHECK-LABEL: define i1 @inttoptr_cmp_null_nonzero(
+; CHECK-SAME: i16 [[X:%.*]]) {
+; CHECK-NEXT:    [[CMP:%.*]] = icmp eq i16 [[X]], -1
+; CHECK-NEXT:    ret i1 [[CMP]]
+;
+  %p = inttoptr i16 %x to ptr addrspace(3)
+  %cmp = icmp eq ptr addrspace(3) %p, null
+  ret i1 %cmp
+}
+
```
arsenm, line 16: Test vector case too

llvm/lib/Transforms/InstCombine/InstructionCombining.cpp
shiltian: Why is that? What's the difference between a splat and an array of null pointers?

dtcxzyw:
> > BTW, I guess we should drop `m_Zero` and always use `m_ZeroInt`.
> 
> In a separate PR?

yep

llvm/lib/Transforms/InstCombine/InstructionCombining.cpp
arsenm: Level of effort and likelihood it appears. Any vector needs a splat, special aggregates are less interesting

arsenm (approved):

PR #197724, opened 2026-05-14
[RISCV] Fix crash lowering fixed-length llvm.masked.{u,s}{div,rem}

lukel97:
We weren't converting back to a fixed length vector after lowering it in the scalable container.

I've no idea how this worked beforehand. Should fix the buildbot failures in https://lab.llvm.org/buildbot/#/builders/210/builds/10590


bababuck (approved):
LGTM

topperc (approved):
LGTM

PR #197726, opened 2026-05-14
[flang][OpenMP] Fix EQUIVALENCE variable privatization in OpenMP

### LINKED ISSUE #197553: [flang][OpenMP] Segfault when privatizing EQUIVALENCE'd array in OpenMP construct
**Description:**
Privatizing an `EQUIVALENCE`'d array in an OpenMP construct causes a stack buffer overflow, resulting in a segfault.

**Reproducer:**
```
program equiv_repro
  integer :: A(10), B(1)
  equivalence (A(1), B(1))
  !$omp parallel do lastprivate(A)
  do i = 1, 4
    A(:) = i
  end do
  !$omp end parallel do
  if (A(1) == 4 .and. A(10) == 4) then
    print *, "PASS"
  else
    print *, "FAIL:", A(1), A(10)
  end if
end program
```

```
> flang -fopenmp -O3 equiv_repro.f90 -o equiv_repro
> ./equiv_repro
Segmentation fault (core dumped)
```

**Root cause:**

`EQUIVALENCE`'d variables have their addresses wrapped in `fir::PointerType` for alias analysis purposes. In `privatizeSymbol` (Utils.cpp), a guard meant to preserve the type for true Fortran `POINTER` variables also matches these `EQUIVALENCE` aliases, causing the private copy to be allocated as pointer-sized (8 bytes) instead of the actual array size.

**Suggested fix:**

The `PointerType` guard in `privatizeSymbol` should check whether the symbol is a true Fortran `POINTER` and only skip type unwrapping in that case. `EQUIVALENCE` aliases should have their `fir.ptr` wrapper stripped so the array is allocated at its correct size.

cenewcombe:
Fixes #197553

EQUIVALENCE aliases are lowered with `fir.ptr` addresses (`castAliasToPointer` in ConvertVariable.cpp) to inform alias analysis. However, `privatizeSymbol()` in Utils.cpp treated all `fir::PointerType` values as true Fortran POINTERs, skipping the `unwrapRefType` that computes the correct allocation type. For arrays, this caused the privatizer to allocate pointer-sized storage instead of the full array, resulting in stack buffer overflows at runtime.

The fix adds a `!semantics::IsPointer()` check so that only true Fortran POINTERs preserve the `fir.ptr` wrapping. EQUIVALENCE aliases are correctly unwrapped to their underlying type.

**Changes:**
- flang/lib/Lower/Support/Utils.cpp: Gate the `PointerType` guard on `semantics::IsPointer` to distinguish true POINTERs from EQUIVALENCE aliases
- flang/test/Lower/OpenMP/DelayedPrivatization/equivalence.f90: Update existing test to match corrected output (alloc type is now `f32`, not `fir.ptr<f32>`)
- flang/test/Lower/OpenMP/lastprivate-equivalence.f90: New test for EQUIVALENCE'd array with lastprivate, verifying correct allocation, deallocation, and writeback

### DIFF
--- a/flang/lib/Lower/Support/Utils.cpp
+++ b/flang/lib/Lower/Support/Utils.cpp
@@ -726,7 +726,10 @@ void privatizeSymbol(
 
   mlir::Value privVal = hsb.getAddr();
   mlir::Type allocType = privVal.getType();
-  if (!mlir::isa<fir::PointerType>(privVal.getType()))
+  // Only preserve fir.ptr wrapping for true Fortran POINTERs; EQUIVALENCE
+  // aliases also use fir.ptr but need the underlying type for allocation.
+  if (!mlir::isa<fir::PointerType>(privVal.getType()) ||
+      !semantics::IsPointer(ultimate))
     allocType = fir::unwrapRefType(privVal.getType());
 
   if (auto poly = mlir::dyn_cast<fir::ClassType>(allocType)) {
--- a/flang/test/Lower/OpenMP/DelayedPrivatization/equivalence.f90
+++ b/flang/test/Lower/OpenMP/DelayedPrivatization/equivalence.f90
@@ -13,24 +13,21 @@ subroutine private_common
   !$omp end parallel
 end subroutine
 
-! TODO: the copy region for pointers is incorrect. OpenMP 5.2 says
-!
-! > If the original list item has the POINTER attribute, the new list items
-! > receive the same association status as the original list item
-!
-! Currently the original pointer is unconditionally loaded, which is undefined
-! behavior if that pointer is not associated.
+! Verify that the privatizer's alloc type is the underlying data type (f32),
+! not fir.ptr<f32>. Variables in EQUIVALENCE are lowered with fir.ptr addresses
+! (see castAliasToPointer in ConvertVariable.cpp), but they are not Fortran
+! POINTERs. The privatizer must allocate storage for the actual data type.
 
-! CHECK:  omp.private {type = firstprivate} @[[X_PRIVATIZER:.*]] : ![[X_TYPE:fir.ptr<f32>]] copy {
-! CHECK:  ^bb0(%[[ORIG_PTR:.*]]: ![[X_TYPE]], %[[PRIV_REF:.*]]: ![[X_TYPE]]):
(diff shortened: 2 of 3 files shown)


MattPD (commented):
FWIW, LGTM. The POINTER-vs-EQUIVALENCE distinction via `semantics::IsPointer(ultimate)` looks correct: true Fortran POINTERs use `fir.box<fir.ptr<T>>` wrapping so `isa<PointerType>` on `hsb.getAddr()` is already false for them, and the only bare `fir.ptr<T>` producer is `castAliasToPointer` for EQUIVALENCE aliases.

One suggestion: the tests cover scalar (`f32`) and integer array (`i32[10]`) EQUIVALENCE. CHARACTER and sequence derived type EQUIVALENCE aliases are also legal and go through somewhat different privatizer paths in `PrivateReductionUtils.cpp`. Adding a CHARACTER EQUIVALENCE test case would strengthen coverage, especially since the regression class is memory corruption rather than malformed IR.

(Disclaimer: I'm not an expert in variable privatization, so please don't treat this as an authoritative review.)

cenewcombe:
@MattPD good idea--I've added the test case for CHARACTER with equivalence. I'll seek another review as well. Thanks!

tblah (approved):
My worry with this is that allocating new storage breaks the storage association between the EQUIVILANCE'd variables.

The relevant part of [OpenMP 6.0](https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-6-0.pdf) is page 221 in section 7.4 "List item privatization".

> If A is a variable that is privatized by a construct and B is a variable that is storage associated with A then:
> - The contents, allocation, and association status of B are undefined on entry to the region;
> - Any definition of A, or of its allocation or association status, causes the contents, allocation, and association status of B to be come undefined; and
> - Any definition of B, or of its allocation or association status, causes the contents, allocation, and association status of A to become undefined.

I think this allows what is being done here for the case where A is privatized and B is not.

For the case where both A and B are privatized, I still think it is valid to do this because (earlier on the same page):
> If any statement of the construct references a list item, a new list item of the same type and type parameters is allocated. This allocation occurs once for each task generated by the construct and once for each SIMD lane used by the construct. 

TL;DR I think this is okay but it felt worth documenting the reasoning.

LGTM and thanks for the patch.

cenewcombe:
Thanks @tblah!

PR #197727, opened 2026-05-14
[mlir][SPIR-V] Fix math.powf lowering for non-integer exponents

aobolensk:
The ConvertFToS usage only works when y is an integer. Use it only for integer constants, for others: lower as GL.Exp(y * GL.Log(x))

### DIFF
--- a/mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
+++ b/mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
@@ -365,25 +365,46 @@ struct PowFOpPattern final : public OpConversionPattern<math::PowFOp> {
     Location loc = powfOp.getLoc();
     auto operandType = adaptor.getRhs().getType();
 
-    // ConvertFToS-based parity needs an integer-valued exponent. Otherwise
-    // fall back to exp(y*log(x)), which yields NaN for x<0 (matches C).
-    auto isIntegerValuedConstant = [](Value v) -> bool {
-      Attribute attr;
-      if (!matchPattern(v, m_Constant(&attr)))
-        return false;
-      return TypeSwitch<Attribute, bool>(attr)
-          .Case([](FloatAttr a) { return a.getValue().isInteger(); })
-          .Case([](SplatElementsAttr a) {
-            return a.getSplatValue<APFloat>().isInteger();
-          })
-          .Case([](DenseElementsAttr a) {
-            return llvm::all_of(a.getValues<APFloat>(),
-                                [](const APFloat &v) { return v.isInteger(); });
-          })
-          .Default(false);
+    // Parity-based lowering requires an integer-valued constant exponent.
+    // Otherwise fall back to exp(y*log(x)), which yields NaN for x<0 (matches
+    // C).
+    auto isOdd = [](const APFloat &v) {
+      APSInt i(/*BitWidth=*/64, /*isUnsigned=*/false);
+      bool ignored;
+      v.convertToInteger(i, APFloat::rmTowardZero, &ignored);
+      return i[0];
     };
 
-    if (!isIntegerValuedConstant(adaptor.getRhs())) {
+    Attribute rhsAttr;
+    SmallVector<bool> oddMask;
+    bool isIntegerValued = false;
+    if (matchPattern(adaptor.getRhs(), m_Constant(&rhsAttr))) {
+      isIntegerValued = TypeSwitch<Attribute, bool>(rhsAttr)
+                            .Case([&](FloatAttr a) {
+                              if (!a.getValue().isInteger())
+                                return false;
+                              oddMask.push_back(isOdd(a.getValue()));
+                              return true;
+                            })
+                            .Case([&](SplatElementsAttr a) {
+                              APFloat v = a.getSplatValue<APFloat>();
+                              if (!v.isInteger())
+                                return false;
+                              oddMask.push_back(isOdd(v));
+                              return true;
+                            })
+                            .Case([&](DenseElementsAttr a) {
+                              for (const APFloat &v : a.getValues<APFloat>()) {
+                                if (!v.isInteger())
+                                  return false;
+                                oddMask.push_back(isOdd(v));
+                              }
+                              return true;
+                            })
+                            .Default(false);
+    }
+
+    if (!isIntegerValued) {
       Value log = spirv::GLLogOp::create(rewriter, loc, adaptor.getLhs());
       Value mul = spirv::FMulOp::create(rewriter, loc, adaptor.getRhs(), log);
       rewriter.replaceOpWithNewOp<spirv::GLExpOp>(powfOp, mul);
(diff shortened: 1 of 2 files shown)


mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
```diff
@@ -360,62 +361,41 @@ struct PowFOpPattern final : public OpConversionPattern<math::PowFOp> {
...
+    // fall back to exp(y*log(x)), which yields NaN for x<0 (matches C).
+    auto isIntegerValuedConstant = [](Value v) -> bool {
+      Attribute attr;
+      if (!matchPattern(v, m_Constant(&attr)))
+        return false;
+      if (auto fAttr = dyn_cast<FloatAttr>(attr))
+        return fAttr.getValue().isInteger();
+      if (auto dense = dyn_cast<DenseFPElementsAttr>(attr))
+        return llvm::all_of(dense.getValues<APFloat>(),
+                            [](const APFloat &v) { return v.isInteger(); });
```
kuhar: turn this into a type switch and handle `SplatElementsAttr` separately, so that we don't repeatedly check the same values

aobolensk: Done

mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
```diff
@@ -360,62 +362,45 @@ struct PowFOpPattern final : public OpConversionPattern<math::PowFOp> {
...
+        return false;
+      return TypeSwitch<Attribute, bool>(attr)
+          .Case<FloatAttr>([](FloatAttr a) { return a.getValue().isInteger(); })
+          .Case<SplatElementsAttr>([](SplatElementsAttr a) {
+            return a.getSplatValue<APFloat>().isInteger();
+          })
+          .Case<DenseFPElementsAttr>([](DenseFPElementsAttr a) {
+            return llvm::all_of(a.getValues<APFloat>(),
+                                [](const APFloat &v) { return v.isInteger(); });
+          })
```
kuhar:
```suggestion
      return TypeSwitch<Attribute, bool>(attr)
          .Case([](FloatAttr a) { return a.getValue().isInteger(); })
          .Case([](SplatElementsAttr a) {
            return a.getSplatValue<APFloat>().isInteger();
          })
          .Case([](DenseElementsAttr a) {
            return llvm::all_of(a.getValues<APFloat>(),
                                [](const APFloat &v) { return v.isInteger(); });
          })
```

aobolensk: Done, thanks

mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
```diff
@@ -360,62 +362,45 @@ struct PowFOpPattern final : public OpConversionPattern<math::PowFOp> {
...
-        spirv::SelectOp::create(rewriter, loc, cmpNegativeWithFractionalExp,
-                                nanValue, adaptor.getLhs());
-    Value abs = spirv::GLFAbsOp::create(rewriter, loc, lhs);
-
-    // TODO: The following just forcefully casts y into an integer value in
-    // order to properly propagate the sign, assuming integer y cases. It
-    // doesn't cover other cases and should be fixed.
-
-    // Cast exponent to integer and calculate exponent % 2 != 0.
     Value intRhs =
```
IgWod: The rhs is guaranteed to be a constant here, because we would hit ` if (!isIntegerValuedConstant(adaptor.getRhs())) {` otherwise, right?  Based on that can we simplify the code here?

1) Not bother with `ConvertFToSOp` and just materialize an integer constant of a right type directly?
2) Remove the whole `shouldNegate` logic and just emit negate based on the sign of the constant?

IgWod: Actually, why do we disallow, non-constant integers in here?

kuhar (approved):

mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
aobolensk: We cannot know that this number is potentially the integer, unless it is constant. There current implementation that is residing on the main branch is good for integers, but incorrect for numbers with fractional part. If you have any suggestions on how to know that the number is integer without requiring it to be constant, I can consider that as well

IgWod: That makes sense. I missed the fact the value type is float. In that case is my suggestion around simplify the lowering sensible?

aobolensk: Yes, they are applicable. Done

mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
```diff
@@ -360,75 +362,86 @@ struct PowFOpPattern final : public OpConversionPattern<math::PowFOp> {
...
-    // Get the scalar float type.
-    FloatType scalarFloatType;
-    if (auto scalarType = dyn_cast<FloatType>(powfOp.getType())) {
-      scalarFloatType = scalarType;
-    } else if (auto vectorType = dyn_cast<VectorType>(powfOp.getType())) {
-      scalarFloatType = cast<FloatType>(vectorType.getElementType());
-    } else {
-      return failure();
+    Location loc = powfOp.getLoc();
+    auto operandType = adaptor.getRhs().getType();
```
IgWod:
```suggestion
    Type operandType = adaptor.getRhs().getType();
```

mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
```diff
@@ -360,75 +362,86 @@ struct PowFOpPattern final : public OpConversionPattern<math::PowFOp> {
...
-
-    // calculate pow based on abs(lhs)^rhs.
-    Value pow = spirv::GLPowOp::create(rewriter, loc, abs, adaptor.getRhs());
     Value negate = spirv::FNegateOp::create(rewriter, loc, pow);
-    // if the exponent is odd and lhs < 0, negate the result.
-    Value shouldNegate =
-        spirv::LogicalAndOp::create(rewriter, loc, lessThan, isOdd);
+
+    Value shouldNegate;
+    if (llvm::all_of(oddMask, [](bool b) { return b; })) {
```
IgWod: `llvm::all_equal`

mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
```diff
@@ -360,75 +362,86 @@ struct PowFOpPattern final : public OpConversionPattern<math::PowFOp> {
...
+      bool ignored;
+      v.convertToInteger(i, APFloat::rmTowardZero, &ignored);
+      return i[0];
+    };
+
+    Attribute rhsAttr;
+    SmallVector<bool> oddMask;
+    bool isIntegerValued = false;
+    if (matchPattern(adaptor.getRhs(), m_Constant(&rhsAttr))) {
+      isIntegerValued = TypeSwitch<Attribute, bool>(rhsAttr)
```
IgWod: Can we make the formatting less ugly somehow? If absolutely necessary you can wrap it in `// clang-format off` and `// clang-format on`.

IgWod: Maybe make it back into lambda with `oddMask` captured?

aobolensk: > Maybe make it back into lambda with `oddMask` captured?

That helps, thanks

mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
aobolensk: Done

mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
aobolensk: Done

mlir/lib/Conversion/MathToSPIRV/MathToSPIRV.cpp
```diff
@@ -360,75 +362,87 @@ struct PowFOpPattern final : public OpConversionPattern<math::PowFOp> {
...
+    // C).
+    auto isOdd = [](const APFloat &v) {
+      APSInt i(/*BitWidth=*/64, /*isUnsigned=*/false);
+      bool ignored;
+      v.convertToInteger(i, APFloat::rmTowardZero, &ignored);
+      return i[0];
+    };
+
+    SmallVector<bool> oddMask;
+    auto isIntegerValuedConstant = [&](Value v) -> bool {
```
IgWod: Thinking about it, I think we could simplify this a lot:

1) Remove the lambda completely
2) Have a `TypeSwitch` only populate the `oddMask`
3) Instead of checking`!isIntegerValuedConstant(adaptor.getRhs())` just check `oddMask.empty()`.

What do you think?

aobolensk: That makes sense. Applied

kuhar (approved):
LGTM but also give @IgWod a chance to review before landing

IgWod (approved):
Thanks!

PR #197728, opened 2026-05-14
[MLIR] Add `IntegerDivisibilityAnalysis` and `InferIntDivisibilityOpInterface`

lialan:
This patch is a port from https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/Util/Analysis/IntegerDivisibilityAnalysis.cpp to upstream

It introduces a dataflow analysis that tracks integer divisibility (divisor + remainder lattice) for SSA values, plus an op interface `InferIntDivisibilityOpInterface` for ops to participate.

It adds:
* `IntegerDivisibilityAnalysis` produces a `Divisibility` lattice `{divisor, remainder}`
* `InferIntDivisibilityOpInterface` interface
* External-model implementations for `arith` and `affine` ops
* `test-int-divisibility` test pass + lit tests

Example:
Here is the usual approach to laod element `i` from `i4` buffer emulated in `i8` buffer:
```mlir
  %byte_idx = arith.divui %i, %c2 : index
  %bit_off  = arith.remui %i, %c2 : index
  %byte     = memref.load %buf[%byte_idx] : memref<?xi8>
  %shifted  = arith.shrui %byte, %bit_off_as_i8
  %nibble   = arith.andi  %shifted, %c0xF : i8
```
If `i` comes from `affine.apply (d0) -> (d0 * 8 + 0)` over loop IV, analysis derives `divisor=8, remainder=0` for `%i`. Then we can emit more concise code by avoiding the shift. There are more scenarios that could benefit from this analysis.


mlir/lib/Dialect/Arith/IR/InferIntDivisibilityOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,162 @@
...
+            ? lhsDivisibility.sdiv() / std::abs(intVal.getSExtValue())
+            : 1;
+
+    setResultDivs(divOp, ConstantIntDivisibility(divUDiv, divSDiv));
+  }
+};
+
+} // namespace
+
+void mlir::arith::registerInferIntDivisibilityOpInterfaceExternalModels(
```
krzysz00: Note: because this is now a built-in interface under `Interfaces/`, we don't need the external model stuff and can probably go directly to attaching this to the op.

krzysz00 (approved):
My one comment is that this doesn't need to be an external interface anymore, otherwise this is known good code of general utility wandering upstream so approved with changes

(But also I called for the upstreaming so I'll let this sit for a few days to see if anyone else objects)

mlir/lib/Dialect/Arith/IR/InferIntDivisibilityOpInterfaceImpl.cpp
krzysz00: (and that's a general note over all dialects)

lialan: Done. removed.

Kewen12:
Hello! This PR breaks bots, could you please fix or revert it?

PR #197729, opened 2026-05-14
SystemZ: Remove always-true conditional

theotherjimmy:
About 20 lines above, the DSA64Bit is unconditionally set, so switching on it later is not needed. Instead, we can print that bit unconditionally.

Resolves #170125


theotherjimmy:
Greeter failed, so I'll take up that mantle.

Hi, I'm Jimmy. No not the Jimmy that you're thinking of, theotherjimmy :grin:. I currently work at IBM's Linux on Z compiler and debugger team. I've been working in open source for about a decade now, and have used github for most of that time (so I know that this is probably not a way to get the greeter check green :wink: ).

redstar (approved):
LGTM, Thanks!

redstar:
@theotherjimmy Do you need help to commit this PR?

theotherjimmy:
I don't think I can merge it myself. So, yes help with that would be appreciated.

PR #196891, opened 2026-05-11
 Reapply "[lldb] Do not refcount breakpoints in lldb-server" (#195858) 

labath:
This reapplies #195858 with a fix for 32-bit arm (and generally, any architecture that uses software single-stepping). The problem was that the temporary breakpoints used for single-stepping were interfering with the breakpoints set by the client.

The fix is to check for existing breakpoints before setting the temporary ones. To achieve this, I've separated the notion of "next PC candidates for a thread" from "step breakpoints we've actually set".

The freebsd code had some software single stepping code, but:
- this was [introduced](https://reviews.llvm.org/D95802) for mips64 support, which was [removed](https://github.com/llvm/llvm-project/pull/179582) earlier this year
- AFAICT, this never worked since the original patch only checked `m_threads_stepping_with_breakpoint`, but never set it to anything.

This is why I'm removing the remnants of the single step support instead of trying to adapt it.

The original commit message was:

We did not say so explictly, but I'd argue that via https://github.com/llvm/llvm-project/pull/195815, we are supporting stubs which do not refcount breakpoints. In these stubs the set/clear breakpoint packets are idempotent:

-    setting a breakpoint for the second time is a no-op (returns OK)
-    clearing a breakpoint clears it, regardless of how many times it has been set
-    clearing a non-existent breakpoint (either because it was already cleared, or because it was never set) returns an error

This makes lldb-server one of those stubs, which makes the code slightly simpler, but more importantly, ensures we do not regress this behavior.

PR #197730, opened 2026-05-14
[lldb] Use Expected in BreakpointLocationPredictor

labath:
This makes it easier to not have long-lived Status objects. Forgetting to clear those was a problem I ran into in #196891.

### DIFF
--- a/lldb/source/Core/EmulateInstruction.cpp
+++ b/lldb/source/Core/EmulateInstruction.cpp
@@ -26,6 +26,7 @@
 #include "lldb/lldb-private-interfaces.h"
 
 #include "llvm/ADT/StringRef.h"
+#include "llvm/Support/ErrorExtras.h"
 
 #include <cstring>
 #include <memory>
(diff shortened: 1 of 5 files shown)


lldb/source/Core/EmulateInstruction.cpp
```diff
@@ -613,53 +614,49 @@ bool EmulateInstruction::CreateFunctionEntryUnwind(UnwindPlan &unwind_plan) {
...
   }
 
-  auto entry_pc = m_emulator_up->ReadPC();
-  if (!entry_pc) {
-    status = Status("Can't read PC");
-    return {};
-  }
+  std::optional<addr_t> entry_pc = m_emulator_up->ReadPC();
+  if (!entry_pc)
+    return llvm::createStringErrorV("Can't read PC");
```
bulbazord: nit: this can be `createStringError` (saving a call to `formatv`)

bulbazord (approved):
LGTM overall. I think almost all your `createStringErrorV` calls can be turned into `createStringError`, they don't use formatv-style string formatting, but I'm not sure it matters too much.

lldb/source/Core/EmulateInstruction.cpp
labath: Sure. done.

PR #197731, opened 2026-05-14
[clang][DependencyScanning] Preserve Necessary Preprocessor Callbacks during By-name Lookup

qiongsiwu:
The by-name lookup logic uses new dependency collector callbacks per lookup. The algorithm used to wipe out all callbacks for each query. This turned out to be perilous. We have two raw pointers in the preprocessor that point to the callbacks, and removing all callbacks per query can lead to use-after-free situations through these dangling pointers. Resetting the dangling pointers to null does not really work either, since there may be dependencies between the callbacks and other data structures. An example of this is the `PreprocessingRecord *Record` callback and the `GlobalPreprocessedEntityMap` in ASTReader. Hence, to fix the use-after-free issue, we preserve the callbacks that the preprocessor may hold a raw pointer to. 

This is not intended to indicate how we want to handle this in the long run. We should avoid removing PP callbacks and reset their states across by-name lookups. 

rdar://175362366

### DIFF
--- a/clang/include/clang/Lex/PPCallbacks.h
+++ b/clang/include/clang/Lex/PPCallbacks.h
@@ -472,6 +472,28 @@ public:
   /// \param IfLoc the source location of the \#if/\#ifdef/\#ifndef directive.
   virtual void Endif(SourceLocation Loc, SourceLocation IfLoc) {
   }
+
+  /// Walk owned descendants. For each descendant whose raw pointer satisfies
+  /// `Pred`, release ownership from its owning unique_ptr and append the raw
+  /// pointer to `Released`. Default: leaf — no descendants to walk.
+  virtual void
+  releasePreservedDescendants(llvm::function_ref<bool(PPCallbacks *)> Pred,
+                              SmallVectorImpl<PPCallbacks *> &Released) {}
+
+  /// Walk the subtree rooted at `CB` (recursing into descendants first), then
+  /// check `CB` itself. Any `CB` whose contents satisfy `Pred` has its
+  /// ownership released and the raw pointer appended to `Released`. After this
+  /// returns, `CB` may be safely reset/destroyed without freeing the released
+  /// pointers.
+  static void releaseIfPreserved(std::unique_ptr<PPCallbacks> &CB,
+                                 llvm::function_ref<bool(PPCallbacks *)> Pred,
+                                 SmallVectorImpl<PPCallbacks *> &Released) {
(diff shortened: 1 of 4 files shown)


jansvoboda11:
Can you clarify why setting `Preprocessor::Record` to `nullptr` in `Preprocessor::removePPCallbacks()` doesn't work? You say that `ASTReader::GlobalPreprocessedEntityMap` has some kind of dependency on `Preprocessor::Record`, but I don't see it.

qiongsiwu:
> Can you clarify why setting `Preprocessor::Record` to `nullptr` in `Preprocessor::removePPCallbacks()` doesn't work? You say that `ASTReader::GlobalPreprocessedEntityMap` has some kind of dependency on `Preprocessor::Record`, but I don't see it.

Yes. This has to do with this code: https://github.com/llvm/llvm-project/blob/af53493a4ad3b4e16f4ed99ebb1b9f8ad039b158/clang/lib/Serialization/ASTReader.cpp#L4336. 

An ASTReader instance can survive across multiple lookups, hence its member `GlobalPreprocessedEntityMap`. The member requires continuous keys inserted into it. If the `PreprocessingRecord` call back is reset per query, we can end up with a `StartingID` of 0 again during the life time of `GlobalPreprocessedEntityMap`, which it does not like. Hence we have to preserve the callback.

jansvoboda11:
Thanks for the explanation, that makes sense.

The new `PPCallbacks::releasePreservedDescendants()` only being implemented by `PPChainedCallbacks` doesn't seem right. If `PreprocessingRecord` is the only callback, there are no instances of `PPChainedCallbacks` in `Preprocessor::Callbacks`:

```c++
  void addPPCallbacks(std::unique_ptr<PPCallbacks> C) {
    if (Callbacks)
      C = std::make_unique<PPChainedCallbacks>(std::move(C),
                                                std::move(Callbacks));
    Callbacks = std::move(C);
  }
```

What about something like this? https://github.com/llvm/llvm-project/pull/197770

jansvoboda11 (approved):
I missed the special case in the static function that would handle the root callback. This LGTM then, but I think https://github.com/llvm/llvm-project/pull/197770 is simpler conceptually.

qiongsiwu:
@Bigcheese FYI I am landing this patch for now. I will work on refactoring to avoid remove pp callbacks right after. I discussed this offline with @cyndyishida and we concluded that we'd like to have this sooner than later.

PR #197732, opened 2026-05-14
[clang][NFC] Mark CWG717 as implemented and add a test

localspook:
[CWG717](https://wg21.link/cwg717) allows applying `thread_local` to `extern` block-scope variables and static data members. Clang implements this since 3.3: https://godbolt.org/z/sx4nc3qqe


Endilll (commented):
Can you add the following test cases:
- dependent static data member;
- dependent and non-dependent extern variable at block scope in a lambda.

Endilll:
Also, mirror all the tests for references.

Endilll (approved):

PR #197735, opened 2026-05-14
[libc] Add all the toolchains needed for libc-shared-tests to the docker container.

lntue:
Toolchains include:
- gcc-7, 8, 9, 11
- qemu-static-user
- cross-build toolchain for aarch64, riscv64, ppc64le, including gcc, g++, gmp, mpfr, mpc.

Container size before the change: ~ 500 MB
Container size after the change: ~ 1.3 - 1.4 GB

### DIFF
--- a/.github/workflows/containers/libc/Dockerfile
+++ b/.github/workflows/containers/libc/Dockerfile
@@ -8,6 +8,24 @@ RUN apt-get update && \
     libgmp-dev \
     libmpc-dev \
     ninja-build \
+    build-essential \
+    gcc-9 \
+    g++-9 \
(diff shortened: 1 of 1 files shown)


boomanaiden154 (commented):
PR title needs to be updated. It would also be good to understand how much this increases the size of the container image. Can you collect numbers on what image size is before/after this change?

lntue:
> PR title needs to be updated. It would also be good to understand how much this increases the size of the container image. Can you collect numbers on what image size is before/after this change?

Done.

boomanaiden154 (approved):
LGTM

PR #197739, opened 2026-05-14
[objcopy] Fix uninitialized ShndxType value in ELF symbol

robert-neuville:
While playing around with the ObjCopy library, I noticed strange values, due to uninitialized ShndxType member of the llvm::objcopy::elf::Symbol structure.

When a symbol is defined in a section, the ShndxType member is never initialized.

This patch initialize it to the "SYMBOL_SIMPLE_INDEX" value, which seems to be the most reasonable choice, as this symbol effectively have a valid section index (regular or extended).

### DIFF
--- a/llvm/lib/ObjCopy/ELF/ELFObject.cpp
+++ b/llvm/lib/ObjCopy/ELF/ELFObject.cpp
@@ -720,9 +720,11 @@ void SymbolTableSection::addSymbol(Twine Name, uint8_t Bind, uint8_t Type,
   Sym.Binding = Bind;
   Sym.Type = Type;
   Sym.DefinedIn = DefinedIn;
-  if (DefinedIn != nullptr)
+  if (DefinedIn != nullptr) {
     DefinedIn->HasSymbol = true;
-  if (DefinedIn == nullptr) {
+    Sym.ShndxType = SYMBOL_SIMPLE_INDEX;
+  }
+  else {
     if (Shndx >= SHN_LORESERVE)
       Sym.ShndxType = static_cast<SymbolShndxType>(Shndx);
     else


jh7370 (commented):
What are you trying to achieve here? The value is not used unless `DefinedIn` is `nullptr` (see `Symbol::getShndx`). In general, we don't bother initalizing things that aren't used.

robert-neuville:
For training purpose, I'm writing a small tool similar to readelf or objcopy. Instead of re-writing the whole code, I use the objcopy static library ELF format parsing functions to load ELF files, as this is probably a stable and complete ELF file parser.

Once ELF data loaded in memory, I can easily process it, but I was getting strange values for the ShndxType struct member. After some investigation, it was due to it being uninitalized.

I thought this was a mistake and maybe a potential bug, easy to fix, so I made this pull request. As I'm not familiar with the LLVM code base, I could have overthinking.

jh7370:
> For training purpose, I'm writing a small tool similar to readelf or objcopy. Instead of re-writing the whole code, I use the objcopy static library ELF format parsing functions to load ELF files, as this is probably a stable and complete ELF file parser.
> 
> Once ELF data loaded in memory, I can easily process it, but I was getting strange values for the ShndxType struct member. After some investigation, it was due to it being uninitalized.
> 
> I thought this was a mistake and maybe a potential bug, easy to fix, so I made this pull request. As I'm not familiar with the LLVM code base, I could have overthinking.

Yeah, it's not a mistake: it's very much by design that the value is uninitialized if it isn't used. To initialize it would imply that there is some significance to the value, when in fact there isn't. In ELF terms, it's to do with the symtab shndx section, i.e. the section that is used when there are too many symbols in the symbol table to be indexable via the normal logic (see https://gabi.xinuos.com/elf/05-symtab.html#section-index, specifically the bit about SHN_XINDEX). You can see how it is used in the `Symbol::getShndx` method and initialized in `SymbolTableSection::addSymbol`. These are the only places the value is referenced

I think this PR can be closed. Feel free to reopen if you think there is still a problem.

PR #197745, opened 2026-05-14
[clang] remove lots of "innocuous" addrspacecasts

vtjnash:
As requested here: https://github.com/llvm/llvm-project/pull/197540#discussion_r3240527406. The frontends here noted that all these autogenerated uses of addrspace cast are illegal, but very few of them seem to actually be intended, so they're fairly straightforward to remove many of those (other than changing some tests). By swapping all calls to this function (except the intended semantic ones for parameters and variables) with the uncasted version, AMDGPU does not need to attempt to apply a fix up afterwards by having different addrspace maps. This PR might not yet fix all calls, but the main ones that might have been missed are in matrix/vector extensions (which weirdly override the memory type for temporary values to be different from the type of the object in all other uses) or in OpenMP (which I was a bit too nervous to touch in the same PR).


clang/lib/Basic/Targets/AMDGPU.cpp
```diff
@@ -228,9 +195,7 @@ AMDGPUTargetInfo::AMDGPUTargetInfo(const llvm::Triple &Triple,
       GPUFeatures(Triple.isAMDGCN() ? llvm::AMDGPU::getArchAttrAMDGCN(GPUKind)
                                     : llvm::AMDGPU::getArchAttrR600(GPUKind)) {
   resetDataLayout();
-
-  setAddressSpaceMap(Triple.getOS() == llvm::Triple::Mesa3D ||
-                     !Triple.isAMDGCN());
+  AddrSpaceMap = &AMDGPUDefIsGenMap;
```
arsenm: Can remove the DefIsGen part of the name now

vtjnash: I'm a bit nervous about changing this in the same PR as the removal of many of these casts (since I probably missed some more casts). So I've dropped this from the PR for now, but added in a commit with some more uses of without casted that seems to still pass all the tests.

clang/test/CodeGenOpenCL/amdgpu-enqueue-kernel.cl
```diff
@@ -480,7 +479,6 @@ kernel void test_target_features_kernel(global int *i) {
 // GFX900-NEXT:    [[BLOCK_ASCAST:%.*]] = addrspacecast ptr addrspace(5) [[BLOCK]] to ptr
 // GFX900-NEXT:    [[BLOCK3_ASCAST:%.*]] = addrspacecast ptr addrspace(5) [[BLOCK3]] to ptr
 // GFX900-NEXT:    [[BLOCK12_ASCAST:%.*]] = addrspacecast ptr addrspace(5) [[BLOCK12]] to ptr
-// GFX900-NEXT:    [[BLOCK_SIZES_ASCAST:%.*]] = addrspacecast ptr addrspace(5) [[BLOCK_SIZES]] to ptr
```
arsenm, line 483: These cases were just dead instructions? Does that mean there's an unnecessary stripPointerCasts somewhere?

arsenm (approved):

clang/test/CodeGenOpenCL/amdgpu-enqueue-kernel.cl
vtjnash: hmm, yeah, it looks like it was even called out there that that `CreateMemTemp` returns the wrong thing (note the comment is wrong, it is used for first the call and then the lifetime end, although the call needed the alloca too):
```
      auto Tmp = CreateMemTemp(SizeArrayTy, "block_sizes");
      llvm::Value *TmpPtr = Tmp.getPointer();
      // The EmitLifetime* pair expect a naked Alloca as their last argument,
      // however for cases where the default AS is not the Alloca AS, Tmp is
      // actually the Alloca ascasted to the default AS, hence the
      // stripPointerCasts()
      llvm::Value *Alloca = TmpPtr->stripPointerCasts();
...
      // Return the Alloca itself rather than a potential ascast as this is only
      // used by the paired EmitLifetimeEnd.
      return {ElemPtr, Alloca};
```

arsenm (approved):

jplehr:
Apologies for the late reply. This PR breaks our HIP kokkos bot https://lab.llvm.org/buildbot/#/builders/230/builds/958
This got lost in the little trouble that the buildbots had over the past days.

I did a local revert and that restores the bot passing. Can we revert while we investigate the issue?

vtjnash:
I'm generally in favor of keeping CI green. Though I would like to try to avoid unnecessary reverts, for example if they are merely downstream tests that are incorrect–e.g. pattern matching on exact IR being generated. Do I understand correctly that the test script here (hip-tpl.py?) isn't available in this repository for reproduction, and the logs are not uploaded (only stored in /opt/botworker/llvm/llvm-test-suite/TS-build/External/HIP/TestKokkosHIP-prefix/src/TestKokkosHIP-build/Testing/Temporary/LastTest.log)?

It would be helpful to first know the context of the failure too. According to the experiment at https://github.com/ROCm/llvm-project/pull/2674, it appears claimed that the front-end generates invalid code (uses flat pointers when those aren't valid for the target) and then assumes that optimizations are guaranteed to clean that up later. The hope with this PR is to help move in a direction that avoids needing that sort of hack (by doing operations in whichever addrspace was expressly set by Sema, instead of always forcing the flat address space), but I also acknowledge this PR might be incomplete (I could only fix things that show up in the LLVM tests).

jplehr:
Thank you for getting back to this.

You understand correctly that the script for reproducing the issue is not inside this repository. The post-commit buildbot fleet for llvm-project is managed in the llvm-zorg repository, where the script in question lives. This is not a downstream CI, but the llvm post-commit buildbot in https://lab.llvm.org/buildbot/#/builders/230/builds/958. I am not aware of where the log files should be uploaded to, but we can probably `cat` them at the end so that it is viewable on the web UI in the future.

The bot compiles the Kokkos library with the HIP GPU backend enabled and runs its unit tests. You can find the container that the bot is running at https://github.com/ROCm/aomp/tree/aomp-dev/upstream-buildbots/manylinux-hip-tpl. We have realized that the instructions on how to build the container and execute the bot is suboptimal and are actively working on improving it. @Kewen12 and I are happy to assist and clarify questions regarding how to reproduce the container / bot setup.

Given that our production buildbot is still failing from this commit, I would like to revert the PR while investigating the issue.

jplehr:
I did a bit of an investigation with the help of an AI tool.
The culprit appears to be reproducible with this piece of code

```
#include <hip/hip_runtime.h>
#include <initializer_list>
#include <cstdio>

__device__ int f(std::initializer_list<int> il) { return *il.begin(); }

__global__ void kernel(int* out) {
    out[0] = f({42});  // BUG HERE
}

int main() {
    int *d_out, h_out = 0;
    if (hipMalloc(&d_out, sizeof(int)) != hipSuccess) return 1;
    hipLaunchKernelGGL(kernel, 1, 1, 0, 0, d_out);
    hipError_t err = hipDeviceSynchronize();
    if (err != hipSuccess) {
        printf("FAILED: %s\n", hipGetErrorString(err));
        hipFree(d_out);
        return 1;
    }
    hipMemcpy(&h_out, d_out, sizeof(int), hipMemcpyDeviceToHost);
    hipFree(d_out);
    printf("SUCCESS: result = %d (expected 42)\n", h_out);
    return h_out == 42 ? 0 : 1;
}
```

when compiled as HIP and in my test case with offload target gfx90a.

```
/opt/botworker/llvm/llvm/build/bin/clang++ \
  -x hip test-bug.cpp -o test-bad \
  --offload-arch=gfx90a \
  --rocm-path=/opt/rocm-7.1.1 \
  -I/opt/rocm-7.1.1/include \
  -L/opt/rocm-7.1.1/lib \
  -lamdhip64
```

It produces a memory access fault as well and the IR shows the same changes w.r.t. addrspacecast.

Let me know if I can help more.

vtjnash:
So the IR change here looks correct. The ABI information specified that `_ZNKSt16initializer_listIiE5beginEv`  should take an addrspacecast value, since the function declaration claims it should accept a generic `ptr` created in any addrspace (for the `this` pointer argument) and returns a generic `ptr` (which turns out to mustalias the `ptr %il.coerce0` here). The caller needed to create an alloca (it chooses the local addrspace) to satisfy that demand and then casts that as required by the source code.

Is the header file wrong? Is the `this` pointer argument to this function not supposed to be in the generic addrspace?

```diff
--- hip_old.ll  2026-05-28 14:45:25.509326799 +0000
+++ hip_new.ll  2026-05-28 14:36:51.875256872 +0000
@@ -10,8 +10,8 @@
 $_ZNKSt16initializer_listIiE5beginEv = comdat any
 
 @__const.__assert_fail.fmt = private unnamed_addr addrspace(4) constant [47 x i8] c"%s:%u: %s: Device-side assertion `%s' failed.\0A\00", align 16
-@__hip_cuid_4db5332c33b34f3b = addrspace(1) global i8 0
-@llvm.compiler.used = appending addrspace(1) global [1 x ptr] [ptr addrspacecast (ptr addrspace(1) @__hip_cuid_4db5332c33b34f3b to ptr)], section "llvm.metadata"
+@__hip_cuid_e0514e128f7d7a66 = addrspace(1) global i8 0
+@llvm.compiler.used = appending addrspace(1) global [1 x ptr] [ptr addrspacecast (ptr addrspace(1) @__hip_cuid_e0514e128f7d7a66 to ptr)], section "llvm.metadata"
 
 ; Function Attrs: convergent mustprogress noinline noreturn nounwind optnone
 define weak void @__cxa_pure_virtual() #0 {
@@ -234,12 +234,12 @@
 define hidden noundef i32 @_Z1fSt16initializer_listIiE(ptr %il.coerce0, i64 %il.coerce1) #2 {
 entry:
   %il = alloca %"class.std::initializer_list", align 8, addrspace(5)
-  %il1 = addrspacecast ptr addrspace(5) %il to ptr
-  %0 = getelementptr inbounds nuw %"class.std::initializer_list", ptr %il1, i32 0, i32 0
-  store ptr %il.coerce0, ptr %0, align 8
-  %1 = getelementptr inbounds nuw %"class.std::initializer_list", ptr %il1, i32 0, i32 1
-  store i64 %il.coerce1, ptr %1, align 8
-  %call = call noundef ptr @_ZNKSt16initializer_listIiE5beginEv(ptr noundef nonnull align 8 dereferenceable(16) %il1) #6
+  %0 = getelementptr inbounds nuw %"class.std::initializer_list", ptr addrspace(5) %il, i32 0, i32 0
+  store ptr %il.coerce0, ptr addrspace(5) %0, align 8
+  %1 = getelementptr inbounds nuw %"class.std::initializer_list", ptr addrspace(5) %il, i32 0, i32 1
+  store i64 %il.coerce1, ptr addrspace(5) %1, align 8
+  %il.ascast = addrspacecast ptr addrspace(5) %il to ptr
+  %call = call noundef ptr @_ZNKSt16initializer_listIiE5beginEv(ptr noundef nonnull align 8 dereferenceable(16) %il.ascast) #6
   %2 = load i32, ptr %call, align 4
   ret i32 %2
 }

@@ -263,22 +263,19 @@
   %out.addr = alloca ptr, align 8, addrspace(5)
   %agg.tmp = alloca %"class.std::initializer_list", align 8, addrspace(5)
   %ref.tmp = alloca [1 x i32], align 4, addrspace(5)
-  %out.ascast = addrspacecast ptr addrspace(5) %out to ptr
   %out.addr.ascast = addrspacecast ptr addrspace(5) %out.addr to ptr
-  %agg.tmp.ascast = addrspacecast ptr addrspace(5) %agg.tmp to ptr
-  %ref.tmp.ascast = addrspacecast ptr addrspace(5) %ref.tmp to ptr
-  store ptr addrspace(1) %out.coerce, ptr %out.ascast, align 8
-  %out1 = load ptr, ptr %out.ascast, align 8
+  store ptr addrspace(1) %out.coerce, ptr addrspace(5) %out, align 8
+  %out1 = load ptr, ptr addrspace(5) %out, align 8
   store ptr %out1, ptr %out.addr.ascast, align 8
-  store i32 42, ptr %ref.tmp.ascast, align 4
-  %_M_array = getelementptr inbounds nuw %"class.std::initializer_list", ptr %agg.tmp.ascast, i32 0, i32 0
-  store ptr %ref.tmp.ascast, ptr %_M_array, align 8
-  %_M_len = getelementptr inbounds nuw %"class.std::initializer_list", ptr %agg.tmp.ascast, i32 0, i32 1
-  store i64 1, ptr %_M_len, align 8
-  %0 = getelementptr inbounds nuw %"class.std::initializer_list", ptr %agg.tmp.ascast, i32 0, i32 0
-  %1 = load ptr, ptr %0, align 8
-  %2 = getelementptr inbounds nuw %"class.std::initializer_list", ptr %agg.tmp.ascast, i32 0, i32 1
-  %3 = load i64, ptr %2, align 8
+  store i32 42, ptr addrspace(5) %ref.tmp, align 4
+  %_M_array = getelementptr inbounds nuw %"class.std::initializer_list", ptr addrspace(5) %agg.tmp, i32 0, i32 0
+  store ptr addrspace(5) %ref.tmp, ptr addrspace(5) %_M_array, align 8
+  %_M_len = getelementptr inbounds nuw %"class.std::initializer_list", ptr addrspace(5) %agg.tmp, i32 0, i32 1
+  store i64 1, ptr addrspace(5) %_M_len, align 8
+  %0 = getelementptr inbounds nuw %"class.std::initializer_list", ptr addrspace(5) %agg.tmp, i32 0, i32 0
+  %1 = load ptr, ptr addrspace(5) %0, align 8
+  %2 = getelementptr inbounds nuw %"class.std::initializer_list", ptr addrspace(5) %agg.tmp, i32 0, i32 1
+  %3 = load i64, ptr addrspace(5) %2, align 8
   %call = call noundef i32 @_Z1fSt16initializer_listIiE(ptr %1, i64 %3) #6
   %4 = load ptr, ptr %out.addr.ascast, align 8
   %arrayidx = getelementptr inbounds i32, ptr %4, i64 0
```

jhuber6:
> So the IR change here looks correct. The ABI information specified that `_ZNKSt16initializer_listIiE5beginEv` should take an addrspacecast value, since the function declaration claims it should accept a generic `ptr` created in any addrspace (for the `this` pointer argument) and returns a generic `ptr` (which turns out to mustalias the `ptr %il.coerce0` here). The caller needed to create an alloca (it chooses the local addrspace) to satisfy that demand and then casts that as required by the source code.

I'm pretty sure this is incorrectly storing an AS(5) pointer inside of the initializer list and then retrieving it without an AS-cast because `begin()` returns a flat pointer.
```diff
-  store i32 42, ptr %ref.tmp.ascast, align 4
-  %_M_array = getelementptr inbounds nuw %"class.std::initializer_list", ptr %agg.tmp.ascast, i32 0, i32 0
-  store ptr %ref.tmp.ascast, ptr %_M_array, align 8
+  store i32 42, ptr addrspace(5) %ref.tmp, align 4
+  %_M_array = getelementptr inbounds nuw %"class.std::initializer_list", ptr addrspace(5) %agg.tmp, i32 0, i32 0
+  store ptr addrspace(5) %ref.tmp, ptr addrspace(5) %_M_array, align 8
```

We then do this, which loads a flat pointer type from `%agg.tmp` when we just stored an AS(5) pointer there. This is handled correctly in the old code I believe.
```diff
+  %0 = getelementptr inbounds nuw %"class.std::initializer_list", ptr addrspace(5) %agg.tmp, i32 0, i32 0
+  %1 = load ptr, ptr addrspace(5) %0, align 8
```

efriedma-quic:
Maybe an issue in AggExprEmitter::VisitCXXStdInitializerListExpr?

vtjnash:
Ah, good catch, I missed that store. It could be adjusted in that vicinity, though I don't think this necessarily is the right way to handle it:

```diff
diff --git a/clang/lib/CodeGen/CGExprAgg.cpp b/clang/lib/CodeGen/CGExprAgg.cpp
index a49b1c8cf3b7..95ce1df6e967 100644
--- a/clang/lib/CodeGen/CGExprAgg.cpp
+++ b/clang/lib/CodeGen/CGExprAgg.cpp
@@ -487,6 +487,7 @@ void AggExprEmitter::VisitCXXStdInitializerListExpr(
          Ctx.hasSameType(Field->getType()->getPointeeType(),
                          ArrayType->getElementType()) &&
          "Expected std::initializer_list first field to be const E *");
+  ArrayPtr = ArrayPtr.withPointer(CGF.performAddrSpaceCast(ArrayPtr.emitRawPointer(CGF), CGF.ConvertType(Field->getType())));
 
   // Start pointer.
   AggValueSlot Dest = EnsureSlot(E->getType());
```

I instead kept a few more addrspace cast which seem unnecessary, but might get observed (e.g. by the store in VisitCXXStdInitializerListExpr): https://github.com/llvm/llvm-project/pull/200427. Those might have to be addressed individually, when encountered in Sema, instead of the attempted broad patch of all CreateAggTemp calls.

In general here, the difficulty is teaching Sema to always be careful about calling `getNonLValueExprType` (which is simply `getUnqualifiedType()`) when making expression types which might need to make a temporary. For example, the result of `*(volatile __global int)p` is `int` but has a `decltype()` of `volatile __global int` (and thus `decltype(*p) x = p` might be illegal on some targets when `auto x = *p` would be fine).

PR #197746, opened 2026-05-14
[Object] Fix AddressSanitizer error in z/OS archive parsing

jsji:
Fix heap-buffer-overflow when reading truncated z/OS archive files.
The ZOSArchiveMemberHeader::setMemberHeaderStrings() function was
attempting to read EBCDIC-encoded fields without validating that
the buffer was large enough, causing reads beyond allocated memory.

Changes:
- Add buffer size validation before accessing header fields
- Check Err parameter for null before dereferencing
- Follow BigArchiveMemberHeader pattern for error handling
- Add consumeError() when Err is nullptr but error occurs

This prevents undefined behavior when processing truncated archives
in tests and ensures proper error reporting.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>


jsji:
Fixed in 1cff3151e7788bd1bc75ed050cf099e55d672168

PR #194508, opened 2026-04-28
[DirectX][ObjectYAML] Add ILDN part support

dzhidzhoev:
Add support for DXContainer ILDN part in the ObjectYAML pipeline so it can be represented in structured YAML and round-tripped through yaml2obj/obj2yaml.

ILDN part is meant to store the name of PDB file that contains shader debug info.

PR #197749, opened 2026-05-14
Reland "[DirectX][ObjectYAML] Add ILDN part support"

dzhidzhoev:
This relands #194508, which was reverted in #197348.

This reland addresses the revert reasons:
- Rename `DXContainerYAML::DebugName::DebugName` to `Filename` to avoid
  compilation failures on older `cl.exe` versions due to field name matching
  its class name.
- Fix layering by reverting the MC -> Object dependency introduced previously:
  `ILDNData` is no longer defined in `llvm/Object/DXContainer.h` and used
  by MC. Instead, ILDN fields are represented by `mcdxbc::DebugName` in
  `llvm/MC/DXContainerInfo.h`, and Object uses that type (Object already
  depends on MC).

To simplify review, this PR splits previously reviewed changes and new changes into two separate commits.


boomanaiden154 (approved):
LGTM for the issues that I had with the original.

bogner (approved):

dzhidzhoev:
Thank you!

PR #197752, opened 2026-05-14
[flang][OpenMP][NFC] Share declare mapper helpers for iterator modifier lowering

chichunchen:
Move mapper lookup and implicit default mapper creation into reusable
OpenMP lowering helpers so regular map lowering and iterator-generated
map entries (#197757) can use the same resolution path.

This prepares Flang iterator modifier lowering for map and motion clauses
without changing the generated IR for existing non-iterator maps.

### DIFF
--- a/flang/lib/Lower/OpenMP/Utils.h
+++ b/flang/lib/Lower/OpenMP/Utils.h
@@ -226,6 +226,13 @@ mlir::Value genIteratorCoordinate(Fortran::lower::AbstractConverter &converter,
                                   llvm::ArrayRef<mlir::Value> ivs,
                                   mlir::Location loc);
 
+mlir::FlatSymbolRefAttr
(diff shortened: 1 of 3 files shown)


flang/lib/Lower/OpenMP/Utils.h
```diff
@@ -225,6 +225,13 @@ mlir::Value genIteratorCoordinate(Fortran::lower::AbstractConverter &converter,
                                   llvm::ArrayRef<mlir::Value> ivs,
                                   mlir::Location loc);
 
+mlir::FlatSymbolRefAttr
+resolveMapperId(Fortran::lower::AbstractConverter &converter,
```
skatrak, line 248: Nit: Add doxygen documentation to the new function.

skatrak (approved):
Thank you, LGTM.

PR #197753, opened 2026-05-14
[LifetimeSafety] Warn when lifetimebound attribute is present on the definition but not on declaration

NeKon69:
Warn when the `[[clang::lifetimebound]]` attribute is present on a definition but not on the corresponding declaration, with a note pointing to where the attribute appears.

Closes #183016

### DIFF
--- a/clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeSafety.h
+++ b/clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeSafety.h
@@ -131,13 +131,13 @@ public:
   reportLifetimeboundViolation(const CXXMethodDecl *MDWithLifetimebound) {}
 
   // Reports a member function definition that has [[clang::lifetimebound]] on
-  // the implicit this parameter when the previous declaration does not.
+  // the implicit this parameter when the canonical declaration does not.
   virtual void reportMisplacedLifetimebound(WarningScope Scope,
                                             const CXXMethodDecl *FDef,
                                             const CXXMethodDecl *FDecl) {}
 
   // Reports a function definition parameter that has [[clang::lifetimebound]]
-  // when the corresponding parameter in the previous declaration does not.
+  // when the corresponding parameter in the canonical declaration.
   virtual void reportMisplacedLifetimebound(WarningScope Scope,
                                             const ParmVarDecl *PVDDef,
                                             const ParmVarDecl *PVDDecl) {}
--- a/clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeSafety.h
+++ b/clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeSafety.h
@@ -41,9 +41,9 @@ struct LifetimeSafetyOpts {
 };
 
 /// Enum to track functions visible across or within TU.
-enum class SuggestionScope {
-  CrossTU, // For suggestions on declarations visible across Translation Units.
-  IntraTU  // For suggestions on definitions local to a Translation Unit.
+enum class WarningScope {
+  CrossTU, // For warnings on declarations visible across Translation Units.
+  IntraTU  // For warnings on functions local to a Translation Unit.
 };
 
 /// Abstract interface for operations requiring Sema access.
@@ -105,7 +105,7 @@ public:
       llvm::PointerUnion<const Expr *, const FieldDecl *, const VarDecl *>;
 
   // Suggests lifetime bound annotations for function parameters.
-  virtual void suggestLifetimeboundToParmVar(SuggestionScope Scope,
+  virtual void suggestLifetimeboundToParmVar(WarningScope Scope,
                                              const ParmVarDecl *ParmToAnnotate,
                                              EscapingTarget Target) {}
 
@@ -132,16 +132,18 @@ public:
 
   // Reports a member function definition that has [[clang::lifetimebound]] on
   // the implicit this parameter when the previous declaration does not.
-  virtual void reportMisplacedLifetimebound(const FunctionDecl *FDef,
-                                            const FunctionDecl *FDecl) {}
+  virtual void reportMisplacedLifetimebound(WarningScope Scope,
+                                            const CXXMethodDecl *FDef,
+                                            const CXXMethodDecl *FDecl) {}
 
   // Reports a function definition parameter that has [[clang::lifetimebound]]
   // when the corresponding parameter in the previous declaration does not.
-  virtual void reportMisplacedLifetimebound(const ParmVarDecl *PVDDef,
+  virtual void reportMisplacedLifetimebound(WarningScope Scope,
+                                            const ParmVarDecl *PVDDef,
                                             const ParmVarDecl *PVDDecl) {}
 
   // Suggests lifetime bound annotations for implicit this.
-  virtual void suggestLifetimeboundToImplicitThis(SuggestionScope Scope,
+  virtual void suggestLifetimeboundToImplicitThis(WarningScope Scope,
                                                   const CXXMethodDecl *MD,
                                                   const Expr *EscapeExpr) {}
 
--- a/clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeSafety.h
+++ b/clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeSafety.h
@@ -130,6 +130,16 @@ public:
   virtual void
   reportLifetimeboundViolation(const CXXMethodDecl *MDWithLifetimebound) {}
 
+  // Reports a member function definition that has [[clang::lifetimebound]] on
+  // the implicit this parameter when the previous declaration does not.
+  virtual void reportMisplacedLifetimebound(const FunctionDecl *FDef,
+                                            const FunctionDecl *FDecl) {}
+
+  // Reports a function definition parameter that has [[clang::lifetimebound]]
+  // when the corresponding parameter in the previous declaration does not.
+  virtual void reportMisplacedLifetimebound(const ParmVarDecl *PVDDef,
+                                            const ParmVarDecl *PVDDecl) {}
+
   // Suggests lifetime bound annotations for implicit this.
   virtual void suggestLifetimeboundToImplicitThis(SuggestionScope Scope,
                                                   const CXXMethodDecl *MD,
--- a/clang/include/clang/Basic/DiagnosticGroups.td
+++ b/clang/include/clang/Basic/DiagnosticGroups.td
@@ -640,7 +640,7 @@ def LifetimeSafetyMisplacedLifetimebound
                 [LifetimeSafetyCrossTUMisplacedLifetimebound,
                  LifetimeSafetyIntraTUMisplacedLifetimebound]> {
     code Documentation = [{
-Detects function definitions whose parameters or implicit this argument are marked as [[clang::lifetimebound]] when the corresponding declaration is not.
+Detects function definitions whose parameters or implicit this argument are marked as [[clang::lifetimebound]] when the canonical declaration is not.
 }];
 }
 
--- a/clang/include/clang/Basic/DiagnosticGroups.td
+++ b/clang/include/clang/Basic/DiagnosticGroups.td
@@ -640,7 +640,7 @@ def LifetimeSafetyMisplacedLifetimebound
                 [LifetimeSafetyCrossTUMisplacedLifetimebound,
                  LifetimeSafetyIntraTUMisplacedLifetimebound]> {
     code Documentation = [{
-Detects function definitions whose parameters or implicit this argument are marked as [[clang::lifetimebound]] when the canonical declaration is not.
+Detects function definitions whose parameters or implicit this argument are marked as [[clang::lifetimebound]] when the first declaration is not.
 }];
 }
 
--- a/clang/include/clang/Basic/DiagnosticGroups.td
+++ b/clang/include/clang/Basic/DiagnosticGroups.td
@@ -631,7 +631,14 @@ This warning may produce false-positives diagnostics when it cannot fully model 
   }];
 }
 
-def LifetimeSafetyMisplacedLifetimebound : DiagGroup<"lifetime-safety-misplaced-lifetimebound"> {
+def LifetimeSafetyCrossTUMisplacedLifetimebound
+    : DiagGroup<"lifetime-safety-cross-tu-misplaced-lifetimebound">;
+def LifetimeSafetyIntraTUMisplacedLifetimebound
+    : DiagGroup<"lifetime-safety-intra-tu-misplaced-lifetimebound">;
+def LifetimeSafetyMisplacedLifetimebound
+    : DiagGroup<"lifetime-safety-misplaced-lifetimebound",
+                [LifetimeSafetyCrossTUMisplacedLifetimebound,
+                 LifetimeSafetyIntraTUMisplacedLifetimebound]> {
     code Documentation = [{
 Detects function definitions whose parameters or implicit this argument are marked as [[clang::lifetimebound]] when the corresponding declaration is not.
 }];
--- a/clang/include/clang/Basic/DiagnosticGroups.td
+++ b/clang/include/clang/Basic/DiagnosticGroups.td
@@ -631,6 +631,12 @@ This warning may produce false-positives diagnostics when it cannot fully model 
   }];
 }
 
+def LifetimeSafetyMisplacedLifetimebound : DiagGroup<"lifetime-safety-misplaced-lifetimebound"> {
+    code Documentation = [{
+Detects function definitions whose parameters or implicit this argument are marked as [[clang::lifetimebound]] when the corresponding declaration is not.
+}];
+}
+
 def LifetimeSafetyPermissive : DiagGroup<"lifetime-safety-permissive",
                                          [LifetimeSafetyUseAfterScope,
                                          LifetimeSafetyReturnStackAddr,
@@ -674,7 +680,8 @@ Detects misuse of [[clang::noescape]] annotation where the parameter escapes (fo
 
 def LifetimeSafetyValidations : DiagGroup<"lifetime-safety-validations",
                                           [LifetimeSafetyNoescape,
-                                           LifetimeSafetyLifetimeboundViolation]> {
+                                           LifetimeSafetyLifetimeboundViolation,
+                                           LifetimeSafetyMisplacedLifetimebound]> {
   code Documentation = [{
 Verify function implementations adhere to the annotated lifetime contracts through lifetime safety
 like verifying [[clang::noescape]] and [[clang::lifetimebound]].
--- a/clang/lib/Analysis/LifetimeSafety/Checker.cpp
+++ b/clang/lib/Analysis/LifetimeSafety/Checker.cpp
@@ -334,13 +334,13 @@ public:
     return {CanonicalDecl, Scope};
   }
 
-  const std::pair<const CXXMethodDecl *, WarningScope>
+  std::pair<const CXXMethodDecl *, WarningScope>
   getCanonicalDeclForAttr(const CXXMethodDecl *MD) {
     auto [CanonicalFD, Scope] = getCanonicalFunctionDeclForAttr(MD);
     return {cast_or_null<CXXMethodDecl>(CanonicalFD), Scope};
   }
 
-  const std::pair<const ParmVarDecl *, WarningScope>
+  std::pair<const ParmVarDecl *, WarningScope>
   getCanonicalDeclForAttr(const FunctionDecl *FD, const ParmVarDecl *PVD) {
     auto [CanonicalFD, Scope] = getCanonicalFunctionDeclForAttr(FD);
     if (!CanonicalFD)
--- a/clang/lib/Analysis/LifetimeSafety/Checker.cpp
+++ b/clang/lib/Analysis/LifetimeSafety/Checker.cpp
@@ -61,7 +61,6 @@ private:
   llvm::DenseMap<AnnotationTarget, EscapingTarget> AnnotationWarningsMap;
   llvm::DenseMap<const ParmVarDecl *, EscapingTarget> NoescapeWarningsMap;
   llvm::DenseSet<const Decl *> VerifiedLiftimeboundEscapes;
-  llvm::DenseMap<const Decl *, WarningScope> DeclarationsToAnnotate;
   const LoanPropagationAnalysis &LoanPropagation;
   const MovedLoansAnalysis &MovedLoans;
   const LiveOriginsAnalysis &LiveOrigins;
@@ -136,10 +135,6 @@ public:
       if (PVD->hasAttr<LifetimeBoundAttr>()) {
         // Track that this lifetimebound parameter correctly escapes.
         VerifiedLiftimeboundEscapes.insert(PVD);
-        if (auto [UnannotatedFDecl, Scope] =
-                getUnannotatedDeclBestMatch(cast<FunctionDecl>(FD), PVD);
-            UnannotatedFDecl)
-          DeclarationsToAnnotate.try_emplace(UnannotatedFDecl, Scope);
       } else {
         // Otherwise, suggest lifetimebound for parameter escaping through
         // return or a field in constructor.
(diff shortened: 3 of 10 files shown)


clang/lib/Analysis/LifetimeSafety/Checker.cpp
```diff
@@ -415,6 +416,28 @@ class LifetimeChecker {
...
+      return;
+
+    const FunctionDecl *FDecl = FDef->getPreviousDecl();
+    if (!FDecl)
+      return;
+
+    if (isa<CXXMethodDecl>(FDef) &&
+        getDirectImplicitObjectLifetimeBoundAttr(FDef) &&
+        !getDirectImplicitObjectLifetimeBoundAttr(FDecl))
+      SemaHelper->reportMisplacedLifetimebound(FDef, FDecl);
```
NeKon69: I'm not sure if we want to keep this, since we can find that attribute via `getLifetimeBoundAttrFromFunctionType` while iterating through redeclarations. But if might change that attribute to not attach to the function type in the future, I think it's worth keeping.

clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeSafety.h
```diff
@@ -130,6 +130,16 @@ class LifetimeSafetySemaHelper {
   virtual void
   reportLifetimeboundViolation(const CXXMethodDecl *MDWithLifetimebound) {}
 
+  // Reports a member function definition that has [[clang::lifetimebound]] on
+  // the implicit this parameter when the previous declaration does not.
+  virtual void reportMisplacedLifetimebound(const FunctionDecl *FDef,
```
Xazax-hun: Since this only works for member functions, shouldn't we take a `CXXMethodDecl` instead?

clang/lib/Analysis/LifetimeSafety/Checker.cpp
```diff
@@ -415,6 +416,28 @@ class LifetimeChecker {
     }
   }
 
+  void reportMisplacedLifetimebound() {
+    const FunctionDecl *FDef = dyn_cast<FunctionDecl>(FD);
+    if (!FDef)
+      return;
+
+    const FunctionDecl *FDecl = FDef->getPreviousDecl();
```
Xazax-hun: We might have a long chain of redeclarations. Do we want to warn only for one redeclaration? 

Also, I wonder if we only care about redeclarations in the header files.

E.g.:
```
// foo.h
int* f(int *);

// foo.cpp
int* f(int *);
#include foo.h
int* f(int *);

// later

int *f(int* p [[clang::lifetimebound]]) {
  return p;
}
```

The most value here is asking the user to put the annotation on the declaration in `foo.h`. Less value in adding it to the other redeclarations.

clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeSafety.h
NeKon69: I mean, yeah, probably, but the reason I made this function accept `FunctionDecl` is that `getImplicitObjectParamLifetimeBoundAttr` later accepts a `FunctionDecl`, and we don't really use any `CXXMethodDecl` specific methods. So adding an explicit cast felt unnecessary to me. Do you think I should change that?

clang/lib/Analysis/LifetimeSafety/Checker.cpp
NeKon69: Hmmmm yeah, I can figure something out about this.

NeKon69: Would you prefer a warning on all not annotated cases or only on one? Because I have a feeling that warning on *all* cases might be too harsh, I could just add a note saying something like "another declaration here"

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound.cpp
```diff
@@ -0,0 +1,40 @@
...
+};
+
+struct S {
+  MyObj data;
+  const MyObj &implicit_this_only(); // expected-warning {{'lifetimebound' attribute on the definition is not visible to callers; add it to this declaration instead}}
+  const MyObj &param_only(const MyObj &obj); // expected-warning {{'lifetimebound' attribute on the definition is not visible to callers; add it to this declaration instead}}
+  const MyObj &both(const MyObj &obj, bool); // expected-warning 2 {{'lifetimebound' attribute on the definition is not visible to callers; add it to this declaration instead}}
+};
+
+const MyObj &S::implicit_this_only() [[clang::lifetimebound]] { // expected-note {{'lifetimebound' attribute written on the definition is here}}
```
usx95: Suggestion for note text:
`'lifetimebound' attribute appears here on the definition`

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+struct MyObj {
+  ~MyObj() {}
+};
+
+struct S {
+  MyObj data;
+  const MyObj &implicit_this_only(); // expected-warning {{'lifetimebound' attribute on the definition is not visible to callers; add it to this declaration instead}}
+  const MyObj &param_only(const MyObj &obj); // expected-warning {{'lifetimebound' attribute on the definition is not visible to callers; add it to this declaration instead}}
+  const MyObj &both(const MyObj &obj, bool); // expected-warning 2 {{'lifetimebound' attribute on the definition is not visible to callers; add it to this declaration instead}}
```
usx95: Suggestion for warning text:
`'lifetimebound' attribute should be on the declaration to be visible to callers`

usx95 (commented):
Thanks. Some thoughts:

1. We have cross-tu and intra-tu suggestions to rank the usefulness of these warnings. I think we should have a similar distinction here as well. This is along the lines of @Xazax-hun suggestion to prioritise suggestions in headers (e.g. specially when the def is in a .cc file)

2. Can you try running this over LLVM to see we don't crash.

NeKon69:
So I create separate warning groups like "LifetimeSafetyIntraTUMisplacedLifetimebound" and "LifetimeSafetyCrossTUMisplacedLifetimebound" and emit their corresponding warnings based on where the declaration is located? Also I would like to know your opinion on [this](https://github.com/llvm/llvm-project/pull/197753#discussion_r3248115577)

NeKon69:
The scope increased quite a bit now...

I've verified no crashes on building clang (and neatly, no FPs and FNs) and added a few new tests (with headers).

Didn't add notes that point to other function redeclarations for now, tell me if I need to.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound.cpp
```diff
@@ -0,0 +1,114 @@
+// RUN: rm -rf %t
+// RUN: split-file %s %t
+// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-intra-tu-misplaced-lifetimebound -Wno-dangling -verify=intra %t/test.cpp
+// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-cross-tu-misplaced-lifetimebound -Wno-dangling -I%t -verify=cross %t/cross.cpp
```
usx95: In my recent experience, split files are hard to debug because of the messed up line numbers. I think we can have two separate test files, one for intra-TU having the majority of the tests. More brief tests for cross-TU in a separate lit tests with `split-file`

clang/lib/Analysis/LifetimeSafety/Checker.cpp
```diff
@@ -134,6 +136,10 @@ class LifetimeChecker {
       if (PVD->hasAttr<LifetimeBoundAttr>()) {
         // Track that this lifetimebound parameter correctly escapes.
         VerifiedLiftimeboundEscapes.insert(PVD);
+        if (auto [UnannotatedFDecl, Scope] =
+                getUnannotatedDeclBestMatch(cast<FunctionDecl>(FD), PVD);
+            UnannotatedFDecl)
+          DeclarationsToAnnotate.try_emplace(UnannotatedFDecl, Scope);
```
usx95: This is largely a policy change about where to put the annotation and not about escape analysis itself. I would prefer to do this separately as part of `reportMisplacedLifetimebound` and not `checkAnnotations` (which checks for escape semantics).

clang/lib/Analysis/LifetimeSafety/LifetimeAnnotations.cpp
```diff
@@ -98,6 +107,42 @@ bool implicitObjectParamIsLifetimeBound(const FunctionDecl *FD) {
   return isNormalAssignmentOperator(FD);
 }
 
+const std::pair<const Decl *, WarningScope>
+getUnannotatedDeclBestMatch(const FunctionDecl *FD, const ParmVarDecl *PVD) {
```
usx95: I feel this belongs to the checker.cpp and not in this file as it is not related to heuristics for inferring or reading annotations.

clang/lib/Analysis/LifetimeSafety/Checker.cpp
usx95: We add implicit attributes on redecl if we have attr in a previous decl. Therefore we could suggest attr on "first" decl in the TU (aka canonical decl). 

To prioritise "foo.h", we could instead suggest annotation on "first decl in a different file" but I would prefer to keep this policy simpler to understand by users. I would suggest to later have a separate warning enforcing that attrs on cross-tu redecls should match the canonical decl.

I also feel we should share this policy with annotation suggestions. Sharing this policy with `getCrossTUDecl` makes sense. To scope down this PR, we could change `getCrossTUDecl` in a separate PR to choose the canonical decl (instead of current: last decl in different file).

To summarise, we could:
1. (In this PR), introduce `getCanonicalDeclForAttr` to return the canonical decl and a warning scope (intra-tu vs xTU).
2. Next PR, replace `getCrossTUDecl` to use `getCanonicalDeclForAttr` for suggestions as well.
3. Next PR, identify unannotated cross-TU redeclarations not matching the canonical decl.

WDYT @Xazax-hun ?

Xazax-hun: Sounds good to me.

NeKon69: We might have to file some issues for this. I don't think I can keep all of this in my head..

clang/lib/Analysis/LifetimeSafety/Checker.cpp
```diff
@@ -315,6 +316,42 @@ class LifetimeChecker {
...
   }
 
+  const std::pair<const FunctionDecl *, WarningScope>
+  getCanonicalFunctionDeclForAttr(const FunctionDecl *FD) {
+    if (!FD || !FD->isExternallyVisible())
+      return {nullptr, WarningScope::IntraTU};
+
+    const auto &SM = FD->getASTContext().getSourceManager();
+    const FileID DefFile = SM.getFileID(SM.getExpansionLoc(FD->getLocation()));
+    const FunctionDecl *FirstDecl = nullptr;
```
usx95: You can get this via `FD->getCanonicalDecl`

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
```diff
@@ -0,0 +1,91 @@
...
+struct View {
+  friend View friend_redecl(MyObj &obj); // intra-warning {{'lifetimebound' attribute on an intra-TU definition is not visible to callers; add it to the declaration instead}}
+};
+
+// FIXME: This diagnoses an attribute inherited from another redeclaration, not one written on the definition. Once we enforce that redeclarations agree on lifetimebound, handle this with a dedicated warning and note.
+View friend_redecl(MyObj &obj [[clang::lifetimebound]]); // intra-note {{'lifetimebound' attribute appears here on the definition}}
+
+View friend_redecl(MyObj &obj) {
+  return View{};
+}
```
NeKon69, line 91: I encountered this while compiling clang with this patch.

The finding looks useful to me. the friend declaration does not have `lifetimebound`, while a later redeclaration does. But this is not really the "attribute written on the definition" case, because the definition inherits the attribute from that second declaration.

I added a FIXME for now.

clang/lib/Analysis/LifetimeSafety/Checker.cpp
NeKon69: Thanks!

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
```diff
@@ -0,0 +1,91 @@
+// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-intra-tu-misplaced-lifetimebound -Wno-dangling -verify=intra %s
```
usx95: You can omit the -verify=prefix in both the tests and use the default expected- prefix

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
usx95: Agreed. The finding is useful. The text mentioning "definition" is not accurate. This could be fixed. Can you please file a bug for this after the PR lands. The wording needs to better capture our policy. We would have to document our annotation placement policy somewhere. This can be deferred to later PRs.

clang/lib/Analysis/LifetimeSafety/Checker.cpp
```diff
@@ -315,6 +316,38 @@ class LifetimeChecker {
     }
   }
 
+  const std::pair<const FunctionDecl *, WarningScope>
+  getCanonicalFunctionDeclForAttr(const FunctionDecl *FD) {
+    if (!FD || !FD->isExternallyVisible())
+      return {nullptr, WarningScope::IntraTU};
```
usx95, line 322: This looks problematic. We should still use the canonical decl even if this is not visible externally.

I suspect the following would not warn atm:
```cpp
// Static function - not externally visible
static MyObj &static_func(MyObj &obj);
static MyObj &static_func(MyObj &obj [[clang::lifetimebound]]) {
  return obj;
}

// Anonymous namespace - not externally visible
namespace {
MyObj &anon_func(MyObj &obj);
MyObj &anon_func(MyObj &obj [[clang::lifetimebound]]) {
  return obj;
}

struct AnonS {
  MyObj data;
  MyObj &method();
};

MyObj &AnonS::method() [[clang::lifetimebound]] {
  return data;
}
} // namespace
```

clang/lib/Analysis/LifetimeSafety/Checker.cpp
```diff
@@ -315,6 +316,38 @@ class LifetimeChecker {
...
+
+    Scope = SM.getFileID(SM.getExpansionLoc(CanonicalDecl->getLocation())) !=
+                    DefFile
+                ? WarningScope::CrossTU
+                : WarningScope::IntraTU;
+
+    return {CanonicalDecl, Scope};
+  }
+
+  const std::pair<const CXXMethodDecl *, WarningScope>
```
usx95: You can drop the `const` in the return type.

clang/lib/Analysis/LifetimeSafety/Checker.cpp
NeKon69: Do you think we should warn in these cases too? Alright.

clang/lib/Analysis/LifetimeSafety/Checker.cpp
```diff
@@ -415,6 +447,25 @@ class LifetimeChecker {
     }
   }
 
+  void reportMisplacedLifetimebound() {
+    const FunctionDecl *FDef = dyn_cast<FunctionDecl>(FD);
+    if (!FDef)
+      return;
+    if (const auto *MDef = dyn_cast<CXXMethodDecl>(FDef);
```
usx95, line 463: Consider documenting the body here:
```
// Check if implicit 'this' has lifetimebound on definition but not on declaration.
```
...
```
// Check each parameter for explicit lifetimebound on definition but not on declaration.
```

clang/lib/Analysis/LifetimeSafety/Checker.cpp
usx95: Yes.

clang/lib/Analysis/LifetimeSafety/Checker.cpp
```diff
@@ -415,6 +447,25 @@ class LifetimeChecker {
     }
   }
```
usx95, line 454: Consider adding a brief function docstring.

clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeSafety.h
```diff
@@ -130,8 +130,20 @@ class LifetimeSafetySemaHelper {
   virtual void
   reportLifetimeboundViolation(const CXXMethodDecl *MDWithLifetimebound) {}
 
+  // Reports a member function definition that has [[clang::lifetimebound]] on
+  // the implicit this parameter when the previous declaration does not.
```
usx95: `when the canonical declaration does not.`

clang/include/clang/Basic/DiagnosticGroups.td
```diff
@@ -631,6 +631,19 @@ This warning may produce false-positives diagnostics when it cannot fully model
...
+def LifetimeSafetyCrossTUMisplacedLifetimebound
+    : DiagGroup<"lifetime-safety-cross-tu-misplaced-lifetimebound">;
+def LifetimeSafetyIntraTUMisplacedLifetimebound
+    : DiagGroup<"lifetime-safety-intra-tu-misplaced-lifetimebound">;
+def LifetimeSafetyMisplacedLifetimebound
+    : DiagGroup<"lifetime-safety-misplaced-lifetimebound",
+                [LifetimeSafetyCrossTUMisplacedLifetimebound,
+                 LifetimeSafetyIntraTUMisplacedLifetimebound]> {
+    code Documentation = [{
+Detects function definitions whose parameters or implicit this argument are marked as [[clang::lifetimebound]] when the corresponding declaration is not.
```
usx95: s/corresponding/canonical

clang/lib/Analysis/LifetimeSafety/Checker.cpp
usx95: Can you file that and send me my way for triaging. I am happy to do it myself as well if you prefer.

clang/lib/Sema/SemaLifetimeSafety.h
```diff
@@ -40,7 +40,8 @@ inline bool IsLifetimeSafetyEnabled(Sema &S, const Decl *D) {
       diag::warn_lifetime_safety_dangling_global_moved,
       diag::warn_lifetime_safety_noescape_escapes,
       diag::warn_lifetime_safety_lifetimebound_violation,
```
usx95, line 42: Driveby comment: Looks like we are missing warning for annotation suggestion. We can do it in a separate PR.

usx95 (approved):
LGTM. Thanks. Let's wait for @Xazax-hun review before merging.

NeKon69:
Hmmm, I ran a local LLM fuzzer for this PR and it found this case:

```cpp
template <typename T>
Obj &spec_func(Obj &obj); // primary template, no attr
template <>
Obj &spec_func<Obj>(Obj &obj [[clang::lifetimebound]]); // explicit spec decl, has attr
template <>
Obj &spec_func<Obj>(Obj &obj) { return obj; } // explicit spec def, no attr
```

It warns on the primary template:
```
warning: 'lifetimebound' attribute on an intra-TU definition is not visible to callers; add it to the declaration instead
Obj &spec_func(Obj &obj);
               ^~~~~~~~
note: 'lifetimebound' attribute appears here on the definition
Obj &spec_func<Obj>(Obj &obj [[clang::lifetimebound]]);
                               ^~~~~~~~~~~~~~~~~~~~
```

The note looks like the inherited attribute issue. The definition inherits the attribute from the explicit specialization declaration, so the attribute is not actually written on the definition, and the note is incorrect.

That being said, I don't think we should suggest adding it to the primary template, because that would affect all specializations/instantiations, not just this explicit specialization.

WDYT?

clang/lib/Analysis/LifetimeSafety/Checker.cpp
```diff
@@ -315,6 +316,38 @@ class LifetimeChecker {
...
 
+  std::pair<const FunctionDecl *, WarningScope>
+  getCanonicalFunctionDeclForAttr(const FunctionDecl *FD) {
```
Xazax-hun: Would people ever use these with a declaration that is not a definition? Should we have an assert? Also maybe the parameter names should reflect this like `FDef` and `MDef`.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-cross-tu.cpp
```diff
@@ -0,0 +1,26 @@
+// RUN: rm -rf %t
+// RUN: split-file %s %t
+// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-cross-tu-misplaced-lifetimebound -Wno-dangling -I%t -verify %t/cross.cpp
+
+//--- cross.h
+struct HeaderObj {
+  ~HeaderObj() {}
+};
+
+HeaderObj &header_param(HeaderObj &obj); // expected-warning {{'lifetimebound' attribute on a cross-TU definition is not visible to callers; add it to the declaration instead}}
```
Xazax-hun: I am not sure this is the right wording here, the definition itself is not cross-TU right? It is only in one TU.

Xazax-hun (commented):
Overall, looks good to me. Some nits inline.

Also, this is something that could have fixits, right? But that could be a follow-up PR.

Xazax-hun:
> That being said, I don't think we should suggest adding it to the primary template, because that would affect all specializations/instantiations, not just this explicit specialization.

Hmm. I guess we could try to use some heuristics like only not warn when the parameter type is dependent on the template argument. It might still have some false positives if the type is the same but template specializations have actually different lifetime semantics. But code like that is very prone to errors and I think I would be fine having FPs for something like that.

Xazax-hun (approved):
LGTM!

clang/include/clang/Basic/DiagnosticGroups.td
```diff
@@ -631,6 +631,19 @@ This warning may produce false-positives diagnostics when it cannot fully model
...
+def LifetimeSafetyCrossTUMisplacedLifetimebound
+    : DiagGroup<"lifetime-safety-cross-tu-misplaced-lifetimebound">;
+def LifetimeSafetyIntraTUMisplacedLifetimebound
+    : DiagGroup<"lifetime-safety-intra-tu-misplaced-lifetimebound">;
+def LifetimeSafetyMisplacedLifetimebound
+    : DiagGroup<"lifetime-safety-misplaced-lifetimebound",
+                [LifetimeSafetyCrossTUMisplacedLifetimebound,
+                 LifetimeSafetyIntraTUMisplacedLifetimebound]> {
+    code Documentation = [{
+Detects function definitions whose parameters or implicit this argument are marked as [[clang::lifetimebound]] when the canonical declaration is not.
```
Xazax-hun: I think users not supposed to know what a canonical declaration is, but we plan to make further improvements to this diagnostic soon, so I am fine keeping it temporarily.

NeKon69: I would like to make it better, I assume it's OK if I change it now?

NeKon69:
Thanks! Once this PR is merged I will file issues for all of the things we discussed to help me track the progress :)

I will add that test case I mentioned earlier to this PR and then we can merge this!

clang/include/clang/Basic/DiagnosticGroups.td
Xazax-hun: Sure! Feel free to change it now! I think we could say the first declaration (which is roughly what the canonical declaration in Clang is).

clang/lib/Analysis/LifetimeSafety/Checker.cpp
NeKon69: I will file them once this PR is merged

PR #197756, opened 2026-05-14
[flang-rt] Rework findloc.cpp to dispatch target at runtime

jhuber6:
Summary:
The previous code had a combinatorial explosion of functions by
templating on both the source and target types. This created around 170
instantiations. Instead we just template on the source type and then use
a simple runtime check. This should not affect performance in a
significant way, it introduces maybe a few branches in what is already a
non-trivial operation that I do not think justifies a two-minute compile
time.

The result is that this file goes from 120 seconds to 12 on my machine
and the resulting file goes from 7.2 MiB to 757 kiB. Functinally this
makes us instantiate 1/10th the functions.

### DIFF
--- a/flang-rt/lib/runtime/findloc.cpp
+++ b/flang-rt/lib/runtime/findloc.cpp
@@ -86,6 +86,89 @@ struct LogicalEquivalence {
   }
 };
 
+template <TypeCategory CAT1, int KIND1, TypeCategory CAT2>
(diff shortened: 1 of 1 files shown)


clementval (approved):
This file has been one of the heaviest. I think the compile time improvements is nice if performance change is negligible.

vzakhari (approved):

PR #197047, opened 2026-05-11
[mlir][OpenMP] Add iterator support to map/motion clause

chichunchen:
Extend omp.target_data, omp.target_enter_data, omp.target_exit_data,
and omp.target_update to support `!omp.iterated<Ty>`.

This is part of feature work for llvm#188061

Assisted with copilot

PR #197757, opened 2026-05-14
[Flang][OpenMP] Support iterator modifier in map and motion clauses

chichunchen:
Add iterator lowering for `target data`, `target enter/exit data`,
`target update`, and map clauses inside `declare mapper`.

For example:

```fortran
!$omp target update to(iterator(i=1:n): a(i), b)
```

Conceptually, this becomes:

```text
entries = omp.iterator(i = 1 through n) {
  yield map_info(base = a, bounds = element i)
}
target_update(iterated_maps = entries, ordinary_maps = [b])
```

Keeping the array base and expressing selections through bounds lets the
same approach handle array elements, sections, and descriptor-backed arrays.

Each locator is expanded only over the iterators that appear in it:

```text
iterator(i=1:n, j=1:m): a(i), b(j), c(i,j)
                        |     |      |
                        i     j    i x j
```

Iterator ranges use the same canonical loop semantics throughout lowering:

```text
trip_count = canonicalTripCount(lower, upper, step, inclusive)

if any trip_count == 0:
    result = []
    skip iterator region
```

These rules preserve source behavior in edge cases. Fortran ranges include
the endpoint, so lowering marks them inclusive instead of dropping the last
entry. A range such as `5:1:1` is empty and yields no entries instead of being
mistaken for a large range. Bounds and reconstructed iterator values keep
their source integer width, preventing `INTEGER(16)` values from being
truncated.

Supports explicit and default mappers, including mapper declarations
imported from other modules.

LLVM IR translation of the iterated map and motion entries emitted by this PR follows in #199101.
This PR stacked on top of #197047 and #197752.

This patch is part of feature work for #188061.

Assisted with codex.


chichunchen:
- NFC patch for sharing declare mapper lambdas.
  - This patch is for making review work easier for #197757 
  - Normal map entries and iterated map entries can share the same declare mapper logic: https://github.com/llvm/llvm-project/pull/197752
- [Current PR] Flang lowering for target_data, target_data_begin, target_data_end, and target_update: https://github.com/llvm/llvm-project/pull/197757
- [WIP] MLIR-to-LLVM IR translation for target_data, target_data_begin, target_data_end, and target_update: https://github.com/llvm/llvm-project/pull/199101
  - Still WIP; needs more tests and refinement.
  - Includes runtime test in offload/test/offloading/fortran

flang/lib/Optimizer/OpenMP/MapInfoFinalization.cpp
```diff
@@ -723,6 +723,21 @@ class MapInfoFinalizationPass
 
       if (auto mapUser = llvm::dyn_cast<mlir::omp::MapInfoOp>(user))
         return getFirstTargetUser(mapUser);
+
+      // MapInfoOp inside an omp.iterator body is yielded back to the iterator,
+      // whose result is used by the target op.
+      if (llvm::isa<mlir::omp::YieldOp>(user)) {
```
MattPD: A question about MapInfoFinalizationPass & SSA dominance: When a `MapInfoOp` inside an `omp.iterator` body has a descriptor-based type (allocatable/assumed-shape), the finalization pass's descriptor expansion walk will find it via `hasADescriptor`, call `getFirstTargetUser` to reach the target op outside the iterator region, and then `genDescriptorMaps` will attempt to insert expanded descriptor maps (defined inside the iterator region) as operands of the target op. Wouldn't this produce an SSA dominance violation, since values defined inside the `IteratorOp` region cannot be used as operands of operations outside that region?

Should the descriptor expansion walk skip `MapInfoOp`s nested inside `IteratorOp` regions (and handle their descriptor expansion differently)?

You can try checking, e.g.,
```f90
! repro-dominance.f90
! Expected: SSA dominance violation in MapInfoFinalizationPass
! when a descriptor-based array is used with iterator modifier.
subroutine repro(a, n)
  implicit none
  integer, intent(in) :: n
  real :: a(:)  ! assumed-shape -> fir.box (descriptor-based)
  !$omp target enter data map(iterator(i=1:n): to: a(i))
end subroutine
```

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -299,6 +299,54 @@ static mlir::Value buildIteratorOp(Fortran::lower::AbstractConverter &converter,
...
+        mlir::Value baseAddr = mapInfo->entity.getBase();
+        if (mlir::isa<fir::BaseBoxType>(baseAddr.getType()))
+          baseAddr = fir::BoxAddrOp::create(builder, loc, baseAddr);
+        auto ptrLike =
+            llvm::cast<mlir::omp::PointerLikeType>(baseAddr.getType());
+        mlir::TypeAttr varType = mlir::TypeAttr::get(ptrLike.getElementType());
+        mlir::FlatSymbolRefAttr mapperId =
+            resolveMapperId(converter, loc, object, mapperIdName, mapTypeBits,
+                            directive, /*hasParentObj=*/false);
+        mlir::omp::MapInfoOp mapOp = mlir::omp::MapInfoOp::create(
```
MattPD: I've noticed the iterator path creates a standalone `MapInfoOp` via direct construction rather than going through the `createMapInfoOp` helper in "flang/lib/Utils/OpenMP.cpp". That helper performs type normalization for sequence types with dynamic extents (discarding incomplete shape info and keeping only the element type). Is there a reason to diverge here, or would it be safer to reuse the shared helper to ensure consistent `varPtrType` handling across all map lowering paths?

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1899,14 +1947,30 @@ bool ClauseProcessor::processMap(
...
+      llvm::SmallPtrSet<const Fortran::semantics::Symbol *, 4> ivSyms;
+      collectIteratorIVs(clause, converter, stmtCtx, iteratorRanges, ivSyms);
 
-    processMapObjects(stmtCtx, clauseLocation,
-                      std::get<omp::ObjectList>(clause.t), mapTypeBits,
-                      parentMemberIndices, result.mapVars, *ptrMapObjects,
-                      mapperIdName, /*isMotionModifier=*/false, directive);
+      for (const omp::Object &object : objects) {
+        if (hasIteratorIVReference(object, ivSyms)) {
+          result.mapIterated.push_back(buildIteratedMapEntry(
```
MattPD: For derived-type member maps like `map(iterator(...): x%a(i))`, the iterator path bypasses the parent/member bookkeeping in `processMapObjects` (which builds `parentMemberIndices` and calls mapper resolution with `hasParentObj=true`). Is this intentional for the initial implementation, or is there a risk that the parent record's partial map won't be generated correctly when members are mapped via iterators?

flang/lib/Optimizer/OpenMP/MapInfoFinalization.cpp
agozillon: if we can, could we try to keep or leverage/share as much of the existing expansion logic as possible for this, I'm a little disinclined to start spreading variations of it around everywhere as it's fairly complex as is. If it's unavoidable that's fine, but I'd like to keep it as easy to maintain as possible, as I'm already aware it is a bit of a pain :-)

Otherwise,  does declare mapper run into a similar issue, and if so how does it circumvent it (if at all)? If I recall correctly it's setup similarly to iterator in its MLIR design where it's a function like region that "owns" a block of maps and is tied to a specific map linked to the target region, and then it's later lowered as a function by the OMPIRBuilder and other associated bits of the lowering. Might give some idea of how to address this, or a new issue to open!

agozillon:
I've added @TIFitis as a reviewer as he's as knowledgeable as I am on the mapping infrastructure, and I am going on vacation next week for just under 3 weeks, so if you need another set of eyes in that time period if the PR series hasn't been landed then he's another good set of eyes to help review :-)

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
agozillon: Would be good if we could keep the use of createMapInfoOp! You're free to extend it in however you want, if there is some particular edge case that makes it non-trivial to use. But also not the end of the world, I believe there's probably a few last bastions of calling the create directly around, just good to try unify everything :-)

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
agozillon: Perfectly fine if we do want to leave it as future work, but if we can stick one of the usual TODO emissions in when we hit one of these cases that would be great. 

I'm not the most familiar with iterator, but I imagine you'd mainly run into the problem when iterating over 1 or more member/components of a record.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1935,12 +1999,32 @@ bool ClauseProcessor::processMotionClauses(lower::StatementContext &stmtCtx,
     // Support motion modifiers: iterator.
     std::string mapperIdName = getMapperIdentifier(converter, mapper);
 
-    if (iterator)
-      TODO(clauseLocation, "Iterator modifier is not supported yet");
-
-    processMapObjects(stmtCtx, clauseLocation, objects, mapTypeBits,
-                      parentMemberIndices, result.mapVars, mapObjects,
-                      mapperIdName, /*isMotionModifier=*/true);
+    if (iterator) {
```
agozillon, line 2260: Nit: I wouldn't be against moving this code block (the whole if else) and the block above into a separate function that's called by the two locations if it's similar enough, saves them having divergent paths if adjustments are made to one but not the other.

flang/lib/Lower/OpenMP/Utils.cpp
```diff
@@ -1283,6 +1283,189 @@ mlir::Value genIteratorCoordinate(Fortran::lower::AbstractConverter &converter,
...
+  if (!arrayRef)
+    return std::nullopt;
+
+  evaluate::ExpressionAnalyzer ea{semaCtx};
+  std::optional<ExprTy> arrayBase;
+  const evaluate::NamedEntity &base = arrayRef->base();
+  // Component array references carry the array base separately from the
+  // subscript list. Lower that base, e.g. v%a in v%a(i), and leave the
+  // subscript-dependent part to map bounds.
+  if (const semantics::SymbolRef *symRef = base.UnwrapSymbolRef())
```
agozillon: More a question out of curiosity than a request for change since it's been a little bit since I've worked with derived type components (or tinkering with this level of the compiler to any large extent :-) ), is there a reason why getDataOperandBaseAddr, doesn't get us what we're looking for, is it because we're effectively trying to drop the iterator accessor piece of the symbol (So v%(i) -> v%a)? 

Could we not do the unwrap/rollback of v%(i) -> v%a on the symbol if necessary before feeding it to getDataOperandBaseAddr and then handle these cases automatically via the function? It's likely not as easy as all that, I remember how painful working with the symbols/references can be, so just a thought more than anything.

flang/lib/Lower/OpenMP/Utils.cpp
```diff
@@ -1283,6 +1283,189 @@ mlir::Value genIteratorCoordinate(Fortran::lower::AbstractConverter &converter,
...
+    mlir::Value value = fir::getBase(createSomeExtendedExpression(
+        loc, converter, toEvExpr(expr), converter.getSymbolMap(), stmtCtx));
+    return convertToIndex(value);
+  };
+
+  llvm::SmallVector<mlir::Value> bounds;
+  bounds.reserve(arrayRef->subscript().size());
+  // Translate each Fortran subscript into an OpenMP map bound for the
+  // corresponding array dimension.
+  for (const auto &[dim, subscript] : llvm::enumerate(arrayRef->subscript())) {
```
agozillon, line 1568: Could we add a comment here specifying how this differs from the regular map bounds generation if it does differ and why the existing infrastructure can't be utilised for it (or perhaps there's a way to extend the existing bounds infrastructure to support this case, so we can have our bounds logic in a unified location?)

agozillon (commented):
Added a couple of extra comments/questions on top of annoyingly chiming in on @MattPD's comments :-)

chichunchen:
Thanks for all the feedback and comments. I’m going to start working through them now.

The force-push was only to rebase this PR on top of main after the parent PR landed. It does not address the review feedback yet, so the PR is not ready for another review at this point. I’ll follow up once I’ve addressed the comments.

Thanks!

MattPD (commented):
Thanks for the updates. My three earlier comments are resolved by the 2026-06-24 commits: the SSA-dominance guard in `MapInfoFinalization`, the `createMapInfoOp` reuse, and a TODO for the derived-type member-map case. I verified each on a local build at commit `f2cdab4c2e8b`. A few new questions inline, all about the current lowering-only state (translation deferred to #199101).

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1870,6 +1912,53 @@ void ClauseProcessor::processMapObjects(
...
+  // Objects in an iterator-modified clause may independently reference
+  // iterator variables, so handle each object separately.
+  for (const omp::Object &object : objects) {
+    if (hasIteratorIVReference(object, *ivSyms)) {
+      if (!inDeclareMapper && getBaseObject(object, semaCtx))
+        TODO(clauseLocation, "iterator modifier with derived type member map");
+      result.mapIterated.push_back(buildIteratedMapEntry(
+          converter, semaCtx, clauseLocation, iteratorRanges, object,
+          mapperIdNameRef, mapTypeBits, directive));
+      continue;
```
MattPD, line 2070: An executable `!$omp target` region emits a whole-array implicit map of the iterated array, on top of the element map. Compiling `map(iterator(i=1:8), to: a(i))` with this patch (`flang -fc1 -emit-hlfir -fopenmp -fopenmp-version=52`), with `a` used in the body, emits both maps:

```mlir
%16 = omp.map.info var_ptr(%2#0 : !fir.ref<!fir.array<16xi32>>, ...) map_clauses(to) ... -> !llvm.ptr {name = ""}                     // element map, i = 1:8
%10 = omp.map.info var_ptr(%2#1 : !fir.ref<!fir.array<16xi32>>, ...) map_clauses(implicit, tofrom) ... -> !fir.ref<...> {name = "a"}  // whole array a(1:16)
omp.target map_iterated(%7 : !omp.iterated<!llvm.ptr>) map_entries(%10 -> %arg1 ...)
```

The iterated object is added to `mapIterated` but not `mapObjects`. So [`captureImplicitMap`](https://github.com/llvm/llvm-project/blob/f2cdab4c2e8bce895a78f9f00c27b902400c4923/flang/lib/Lower/OpenMP/OpenMP.cpp#L3107) treats `a` as unmapped. It then adds the whole-array `tofrom` map ([`OpenMP.cpp:3226-3227`](https://github.com/llvm/llvm-project/blob/f2cdab4c2e8bce895a78f9f00c27b902400c4923/flang/lib/Lower/OpenMP/OpenMP.cpp#L3226-L3227)). This changes the extent: the whole array is mapped, not the iterated elements. It also changes the direction: `to` gains a copy-back. The extra map is present in the generated IR today and has no test. It is not yet observable at runtime, because `map_iterated` translation is deferred to #199101. Would it be worth suppressing the implicit whole-array map for a symbol already mapped via an iterator, so #199101 does not inherit it?

flang/test/Transforms/omp-map-info-finalization.fir
```diff
@@ -43,6 +43,30 @@ func.func @test_descriptor_expansion_pass(%arg0: !fir.box<!fir.array<?xi32>>) {
 
 // -----
 
+func.func @test_iterator_descriptor_map_is_not_expanded(%arg0: !fir.box<!fir.array<?xf32>>) {
```
MattPD, line 46: The `isNestedInIterator` guard has no end-to-end test, though it prevents the descriptor-expansion crash. This test uses hand-written input IR, not IR generated by compiling Fortran. Its map shape uses `var_ptr` of `!fir.ref<!fir.array<...>>`. The frontend instead emits a scalar-element `var_ptr` on a `fir.box_addr` base, with result `!llvm.ptr`. I confirmed the frontend path does survive `MapInfoFinalization` without a crash. But no committed test drives Fortran through the pass. So a later frontend shape change could silently stop matching the guard while this `.fir` test still passes. Would a [`flang/test/Lower/OpenMP`](https://github.com/llvm/llvm-project/tree/f2cdab4c2e8bce895a78f9f00c27b902400c4923/flang/test/Lower/OpenMP) test be worth adding, one that compiles an allocatable or assumed-shape iterator map through finalization and CHECKs that the descriptor map is not expanded?

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -317,6 +317,48 @@ static mlir::Value buildIteratorOp(Fortran::lower::AbstractConverter &converter,
...
+            converter, builder, semaCtx, iterStmtCtx, object, loc);
+        if (!mapInfo)
+          TODO(loc, "object type not supported by iterator modifier");
+
+        // Use the array base as var_ptr with bounds so the runtime can
+        // associate this mapping with whole-array mappings via the base
+        // address.
+        mlir::Value baseAddr = mapInfo->entity.getBase();
+        if (mlir::isa<fir::BaseBoxType>(baseAddr.getType()))
+          baseAddr = fir::BoxAddrOp::create(builder, loc, baseAddr);
```
MattPD, line 391: For a descriptor array under `target enter/exit data`, the iterator lowering maps only the array data, unlike the non-iterator baseline. For `integer, allocatable :: a(:)` with `map(iterator(i=1:n), to: a(i))`, the iterator path emits one data-only map:

```mlir
%17 = fir.box_addr %10 : (!fir.box<!fir.heap<!fir.array<?xi32>>>) -> !fir.heap<!fir.array<?xi32>>
%18 = omp.map.info var_ptr(%17 : !fir.heap<!fir.array<?xi32>>, i32) map_clauses(to) ... -> !llvm.ptr {name = ""}
omp.target_enter_data map_iterated(%8 ...)
```

The non-iterator baseline `map(to: a(1:2))` instead emits a descriptor parent, an attach map, and a base-pointer member:

```mlir
%12 = omp.map.info var_ptr(%1#1 ...) map_clauses(to) var_ptr_ptr(%11 ...) ...                         // base-pointer member
%13 = omp.map.info var_ptr(%1#1 ...) map_clauses(always, to) members(%12 : [0] ...) {name = "a(1:2)"} // descriptor parent
%14 = omp.map.info var_ptr(%1#1 ...) map_clauses(attach, ref_ptr, ref_ptee) var_ptr_ptr(%11 ...) {name = "a(1:2)"} // attach
omp.target_enter_data map_entries(%13, %14, %12 ...)
```

For pure data-motion directives, mapping only the data may be intended. The `map_iterated` translation is deferred to #199101, so it cannot affect runtime here. Is the data-only shape the intended contract for iterator maps of descriptor arrays, to be handled and device-tested in #199101?

flang/lib/Lower/OpenMP/Utils.cpp
```diff
@@ -1350,6 +1350,215 @@ void collectEnclosingConstructTraits(
...
+    addr = fir::getBase(dataExv);
+  }
+
+  // Allocatable and pointer locators yield a descriptor address
+  // (!fir.ref<!fir.box<...>>) for both whole objects and components. Load the
+  // descriptor so var_ptr maps the array data, not the descriptor.
+  // Assumed-shape arrays and non-allocatable components are already box values
+  // or plain addresses and need no load.
+  if (fir::isBoxAddress(addr.getType()))
+    addr = fir::LoadOp::create(builder, loc, addr);
```
MattPD, line 1513: The semantic check and the lowering seem to disagree for descriptor iterator maps. For an assumed-shape `a` with `map(iterator(i=1:n), to: a(i))`, semantics emits a `-Wopenmp-usage` warning:

```
warning: The map of 'a' may include a descriptor that is created locally. Mapping this descriptor without an appropriate TARGET EXIT DATA in the same scope may result in the device retaining an invalid descriptor reference [-Wopenmp-usage]
```

But the lowering here maps only raw data, with no descriptor:

```mlir
%15 = fir.box_addr %1#0 : (!fir.box<!fir.array<?xi32>>) -> !fir.ref<!fir.array<?xi32>>
%16 = omp.map.info var_ptr(%15 : !fir.ref<!fir.array<?xi32>>, i32) map_clauses(to) ... -> !llvm.ptr {name = ""}
```

Is the warning correct for this construct, or should it be adjusted for data-only iterator descriptor maps?

flang/lib/Optimizer/OpenMP/MapInfoFinalization.cpp
```diff
@@ -723,6 +723,21 @@ class MapInfoFinalizationPass
...
+          for (auto *iterUser : iterOp->getUsers()) {
+            if (llvm::isa<mlir::omp::TargetOp, mlir::omp::TargetDataOp,
+                          mlir::omp::TargetUpdateOp,
+                          mlir::omp::TargetExitDataOp,
+                          mlir::omp::TargetEnterDataOp,
+                          mlir::omp::DeclareMapperInfoOp>(iterUser))
+              return iterUser;
+          }
+        }
+      }
```
MattPD: This `YieldOp` branch in `getFirstTargetUser` looks unreachable. The finalization walks now skip iterator-nested maps via `isNestedInIterator`, so no call site reaches this branch with an iterator-nested map. Is it still reachable, or can it be removed in favor of the guards?

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -2080,13 +2190,25 @@ bool ClauseProcessor::processMotionClauses(lower::StatementContext &stmtCtx,
...
+      processMapObjectsWithIterator(
+          stmtCtx, clauseLocation, objects, iteratorRanges, &ivSyms,
+          mapTypeBits, parentMemberIndices, result, mapObjects, mapperIdName,
+          /*isMotionModifier=*/true, llvm::omp::Directive::OMPD_target_update);
+    } else {
+      processMapObjectsWithIterator(
+          stmtCtx, clauseLocation, objects, /*iteratorRanges=*/{},
+          /*ivSyms=*/nullptr, mapTypeBits, parentMemberIndices, result,
+          mapObjects, mapperIdName, /*isMotionModifier=*/true,
+          llvm::omp::Directive::OMPD_target_update);
```
MattPD: Passing `OMPD_target_update` here changes behaviour for ordinary `target update`. Before this PR the call passed no directive, so `resolveMapperId` received `OMPD_unknown`. `resolveMapperId` excludes `OMPD_target_update` from implicit default mapper synthesis, so a plain `target update` loses its compiler-generated default mapper.

```fortran
subroutine updonly()
  type :: t
    integer, allocatable :: a(:)
  end type
  type(t) :: x
  !$omp target update to(x)
end subroutine
```

Before, `omp.declare_mapper @_QQFupdonlyt_omp_default_mapper` is emitted and the map carries `mapper(@...)`. After, neither appears. The loss affects `to` and `from`, derived types with direct or nested allocatable components, `type(c_ptr)`, and allocatable or pointer derived-type objects. A preceding `target data` for the same type masks the loss, because the mapper is created earlier.

Was passing `OMPD_target_update` here intended, or should this path pass `OMPD_unknown`?

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -337,6 +337,50 @@ static mlir::Value buildIteratorOp(Fortran::lower::AbstractConverter &converter,
...
+        // address. Iterator maps represent selected data directly and do not
+        // include the descriptor parent or attachment entries used by ordinary
+        // descriptor maps.
+        mlir::Value baseAddr = mapInfo->entity.getBase();
+        if (mlir::isa<fir::BaseBoxType>(baseAddr.getType()))
+          baseAddr = fir::BoxAddrOp::create(builder, loc, baseAddr);
+        mlir::FlatSymbolRefAttr mapperId =
+            resolveMapperId(converter, loc, object, mapperIdName, mapTypeBits,
+                            directive, /*hasParentObj=*/false);
+        mlir::omp::MapInfoOp mapOp = utils::openmp::createMapInfoOp(
```
MattPD, line 395: Taking `fir.box_addr` here drops the descriptor, but the `ref_*` and `attach` flag bits are still set on the resulting map. For `map(ref_ptr, iterator(i=1:n), to: a(i))` on `integer, pointer :: a(:)`, the emitted entry uses the pointee data as `var_ptr`. It has neither `var_ptr_ptr` nor an attach entry, while the non-iterator spelling emits the descriptor entry plus an attach entry. `ref_ptr` requests mapping the pointer, but this IR maps the pointee. All six spellings are accepted with no diagnostic: `ref_ptr`, `ref_ptee`, `ref_ptr_ptee`, and `attach` with `always`, `never`, or `auto`. This is new-path behaviour rather than a regression, and it reaches `target enter data`, `target exit data`, and `target data`.

Would rejecting these modifiers with a not-yet-implemented diagnostic, alongside the existing ones for optional and derived-type-member locators, be reasonable until the descriptor entries can be carried through?

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1924,6 +1968,56 @@ void ClauseProcessor::processMapObjects(
...
+  // iterator modifiers do not support yet.
+  bool inDeclareMapper = mlir::isa_and_present<mlir::omp::DeclareMapperOp>(
+      converter.getFirOpBuilder().getRegion().getParentOp());
+
+  // Objects in an iterator-modified clause may independently reference
+  // iterator variables, so handle each object separately.
+  for (const omp::Object &object : objects) {
+    if (hasIteratorIVReference(object, *ivSyms)) {
+      if (!inDeclareMapper && getBaseObject(object, semaCtx))
+        TODO(clauseLocation, "iterator modifier with derived type member map");
```
MattPD: The lowering gap here predates this PR. What changed is that the not-yet-implemented diagnostic no longer shields the iterator spelling.

`inDeclareMapper` tests where the builder is inserting, not whose member is being mapped. It therefore waives the diagnostic for any object inside a declare mapper, including a mapped object whose base is not the declare-mapper variable. With a module variable `w`, `!$omp declare mapper(m: t :: v) map(iterator(i = 1:10): w%a(i))` now triggers `'hlfir.designate' op using value defined outside the region`, then `verification of lowering to FIR failed`. Before this PR the same spelling gave a clean not-yet-implemented. The non-iterator spelling `map(w%a)` fails identically before and after.

Would comparing the object's base against the declare-mapper variable be a better predicate than the insertion point?

flang/lib/Semantics/check-omp-structure.cpp
```diff
@@ -4932,7 +4955,30 @@ void OmpStructureChecker::Enter(const parser::OmpClause::Map &x) {
...
+    }
+
     for (const parser::OmpObject &object : objects.v) {
+      bool referencesIterator{false};
+      if (const auto *designator{GetDesignatorFromObj(object)}) {
+        OmpIteratorReferenceChecker checker{iteratorSymbols};
+        parser::Walk(*designator, checker);
+        referencesIterator = checker.found();
+      }
+      if (referencesIterator)
```
MattPD, line 4979: Skipping iterator-referencing objects applies to the enter and the exit side alike, so an iterator exit map no longer registers as the matching exit. A plain `target enter data map(to: a)` followed by `target exit data map(iterator(i = 1:n), from: a(i))` now warns on the plain enter map with "The map of 'a' may include a descriptor that is created locally". Before this PR that enter map did not warn. I think the new warning is semantically correct, because an iterator exit map removes selected data but not the descriptor. If so, an enter-plain/exit-iterator case in `target-enter-data-temp-descriptor.f90` would document it, since the new tests cover only the same-clause case.

Is the new warning on the plain enter map intended?

flang/test/Lower/OpenMP/motion-iterator.f90
```diff
@@ -0,0 +1,702 @@
...
+
+  !$omp target update to(iterator(i = 1:n, j = 1:m): a(i, j))
+end subroutine
+
+! CHECK-LABEL: func.func @_QPtarget_update_allocatable_2d
+! CHECK: %[[A:.*]]:2 = hlfir.declare %{{.*}} dummy_scope %{{.*}} arg 1 {fortran_attrs = #fir.var_attrs<allocatable>, uniq_name = "_QFtarget_update_allocatable_2dEa"}
+! CHECK: %[[IT:.*]] = omp.iterator(%{{.*}}: index, %{{.*}}: index) = ({{.*}} to {{.*}} step {{.*}}, {{.*}} to {{.*}} step {{.*}}) {
+! CHECK:   %[[BOX:.*]] = fir.load %[[A]]#0 : !fir.ref<!fir.box<!fir.heap<!fir.array<?x?xi32>>>>
+! CHECK:   %[[BOUNDS0:.*]] = omp.map.bounds lower_bound(%{{.*}} : index) upper_bound(%{{.*}} : index) extent(%{{.*}} : index) stride(%{{.*}} : index) start_idx(%{{.*}} : index) {stride_in_bytes = true}
+! CHECK:   %[[BOUNDS1:.*]] = omp.map.bounds lower_bound(%{{.*}} : index) upper_bound(%{{.*}} : index) extent(%{{.*}} : index) stride(%{{.*}} : index) start_idx(%{{.*}} : index) {stride_in_bytes = true}
```
MattPD: These two-dimensional box-bounds checks wildcard every operand, so they cannot distinguish dimension 0 from dimension 1. A dimension-index swap, or an induction-variable-to-dimension mix-up in the box path, would still pass. The arithmetic is correct today, including a non-default lower bound such as `a(0:)`, so this is about test strength only.

Would binding the `fir.box_dims` results and the index expressions to FileCheck variables, as `target_update_2d` already does for the non-box case, be worth it here?

flang/lib/Optimizer/OpenMP/MapInfoFinalization.cpp
```diff
@@ -1387,6 +1391,8 @@ class MapInfoFinalizationPass
       // is executed again as the final step of this pass to maintain
       // map to block argument consistency.
       func->walk([&](mlir::omp::MapInfoOp op) {
+        if (isNestedInIterator(op))
```
MattPD: `getFirstTargetUser` returns null for every iterator-nested map since the `YieldOp` hop was removed. Several walks assert on its result, so each walk has to skip iterator-nested maps on its own. Four walks carry this guard, at lines 1394, 1404, 1607 and 1704. The `MapInfoOp` walks at lines 1598 and 1634 do not. Neither of those fires for an iterator map today, because the first requires a record type and the second requires members, so this is hardening rather than a live bug.

Could every `MapInfoOp` walk share one filtered walk helper instead of keeping separate guards?

flang/lib/Optimizer/OpenMP/MapInfoFinalization.cpp
```diff
@@ -1395,6 +1401,9 @@ class MapInfoFinalizationPass
       // Next, walk `omp.map.info` ops to see if any record members should be
       // implicitly mapped.
       func->walk([&](mlir::omp::MapInfoOp op) {
+        if (isNestedInIterator(op))
```
MattPD: This guard looks redundant, because `isRecordWithAllocatableMember` just below it cannot hold for an iterator map, whose `var_ptr` is always an array base. Is it kept deliberately for symmetry with the other walks?

MattPD (commented):
Verified at 459d6a936. All five of my previous comments are resolved. I left inline questions.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Confirmed: the data-only contract is explicit now.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Confirmed: an executable `target` with an iterator map is now rejected.

flang/test/Transforms/omp-map-info-finalization.fir
MattPD: Confirmed: the new test covers the frontend-emitted shape that exposed the gap.

flang/lib/Optimizer/OpenMP/MapInfoFinalization.cpp
MattPD: Confirmed removed.

flang/lib/Lower/OpenMP/Utils.cpp
MattPD: Confirmed: the warning no longer contradicts the lowering for iterator maps.

MattPD (commented):
Verified against b98b49c029023b09c121b2a0124affa5ff47a8e4. All seven points from my earlier review have corresponding changes in this head. I have left follow-up questions inline.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Confirmed. An ordinary `target update` keeps its implicit default mapper again.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Confirmed. A reference or attach modifier with an iterator is now rejected, and the plain spelling still lowers.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Thanks, comparing the object's base against the declare-mapper variable is the right predicate. One shape still gets past it. `getBaseObject` returns nothing for a locator that is a plain array element. `isDeclareMapperVariable` is therefore never consulted for that locator. With a module array, the address lookup then returns a null value. Lowering then trips an assertion instead of reporting an error.

Save the following as `repro.f90`:

```fortran
module mv
  integer :: tbl(100)
  type :: t
    real :: a(100)
  end type
  !$omp declare mapper(mm: t :: v) map(iterator(i = 1:100): v%a(i), tbl(i))
end module
```

Then run `flang -fc1 -emit-hlfir -fopenmp -fopenmp-version=52 -o /dev/null repro.f90`.

At b98b49c that input aborts with ``Assertion `detail::isPresent(Val) && "dyn_cast on a non-existent value"' failed``. The abort comes from `fir::factory::getDataOperandBaseAddr` in `flang/include/flang/Optimizer/Builder/DirectivesCommon.h`. Line 61 calls `symAddr.getDefiningOp()` on the null value, before the `if (!symAddr)` guard at line 66.

Before this PR, the same spelling gave a clean not-yet-implemented message. The non-iterator `map(tbl)` still gives `symbol is not mapped to any IR value` rather than an assertion.

Would extending the predicate to a locator that has no base object cover this? Dropping the `getBaseObject` conjunct outright would over-reject, since `isDeclareMapperVariable` returns false outside an `omp.declare_mapper`.

flang/lib/Semantics/check-omp-structure.cpp
MattPD: Confirmed, and `test_plain_enter_iterator_exit_warning` documents the behavior I asked about.

flang/test/Lower/OpenMP/motion-iterator.f90
MattPD: Thanks, `target_update_allocatable_2d` now binds the dimension indices and the induction variables. That covers the request in this thread.

Two gaps are left across these two tests. That block no longer asserts `stride_in_bytes`, `capture(ByRef)`, or the result type. The neighboring cases still assert all three. Separately, `target_update_assumed_shape_2d` still wildcards its `fir.box_dims` dimension operand and both bounds, so a dimension swap would still pass there.

Was dropping those three checks in `target_update_allocatable_2d` intended? Could `target_update_assumed_shape_2d` also pin its `fir.box_dims` dimension operand and both bounds?

flang/lib/Optimizer/OpenMP/MapInfoFinalization.cpp
MattPD: Confirmed. Iterator-nested maps no longer reach the target-user lookup.

flang/lib/Optimizer/OpenMP/MapInfoFinalization.cpp
MattPD: Confirmed. The shared walk helper applies the skip uniformly now, which answers the symmetry question.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1924,6 +1968,92 @@ void ClauseProcessor::processMapObjects(
...
+  // iterator variables, so handle each object separately.
+  for (const omp::Object &object : objects) {
+    if (hasIteratorIVReference(object, *ivSyms)) {
+      bool isMapperVariable = isDeclareMapperVariable(object);
+      if (hasUnsupportedReferenceModifier)
+        TODO(clauseLocation,
+             "iterator modifier with reference or attach modifier");
+      if (getBaseObject(object, semaCtx) && !isMapperVariable)
+        TODO(clauseLocation, "iterator modifier with derived type member map");
+      if (declareMapper && !isMapperVariable)
```
MattPD, line 2061: This hardens the path where the declare mapper is in the file being compiled. When the mapper arrives from a module file, the same feature still ends a downstream compile without a diagnostic.

Save this as `m9.f90`:

```fortran
module mt9
  type :: t
    integer :: a(10)
  end type
  !$omp declare mapper(mm: t :: v) map(iterator(i = 1:10): v%a(i))
end module
```

and this as `u9.f90`, which never references the mapper:

```fortran
subroutine u9
  use mt9
  integer :: k
  k = 1
end subroutine
```

From the directory holding both files, run `flang -fc1 -emit-hlfir -fopenmp -fopenmp-version=52 -J . -o /dev/null m9.f90`. That writes `mt9.mod` into the same directory. Then run `flang -fc1 -emit-hlfir -fopenmp -fopenmp-version=52 -J . -o /dev/null u9.f90`.

At 5159a83 the first command succeeds and the second exits 139. The stack trace ends in `TodoLocators` at `flang/lib/Lower/OpenMP/ClauseProcessor.cpp:41`, where `symbol->test(...)` runs on a null symbol. Compiling the same file with `flang -fc1 -fsyntax-only -fopenmp -fopenmp-version=52 -J . u9.f90` exits 0, so semantics accepts the input and only lowering fails.

The iterator induction variable is not re-resolved when the clause is read back from the module file. Before this PR the same module source was refused with a not-yet-implemented message, so no downstream unit could exist. A mapper written without an iterator crosses the module boundary cleanly both before and after this PR. An iterator whose induction variable is not referenced by the locator also compiles in both files.

Guarding `TodoLocators` by itself would move the same dereference to `ClauseProcessor.cpp:2003` and `ClauseProcessor.cpp:1922`, which also call `sym()` unguarded on this path. Would a split-file test covering a module and a separate USE-ing unit be worth adding to this PR alongside the fix?

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1924,6 +1968,92 @@ void ClauseProcessor::processMapObjects(
...
+        TODO(clauseLocation,
+             "iterator modifier with locator outside declare mapper variable");
```
MattPD, line 2063: This message can be inaccurate. Save this as `d9.f90`:

```fortran
module md9
  type :: t
    complex :: c(10)
  end type
  !$omp declare mapper(mm: t :: v) map(iterator(i = 1:10): v%c(i)%re)
end module
```

Then run `flang -fc1 -emit-hlfir -fopenmp -fopenmp-version=52 -o /dev/null d9.f90`. At 5159a83 the locator is reported as outside the mapper variable. It is rooted at `v`. The same shape outside a declare mapper reports `object type not supported by iterator modifier`, which names the real reason.

The root walk re-derives the base through `getBaseObject`. That call yields nothing for a designator `evaluate::ExtractDataRef` cannot decompose. The root then stays at the leaf and never matches the mapper argument. Lowering rejects this object in both contexts, so only the wording differs.

Would placing this check after the existing object-type rejection keep the more accurate message for those shapes?

MattPD (commented):
Verified against 5159a83d55c5d74114209269d5e969009b7d535a. Both points from my earlier review have corresponding changes in this head, and new issues remain in the inline comments.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Confirmed for the single-file path. The mixed case, one locator rooted at the mapper variable and one rooted elsewhere, now reports a not-yet-implemented message instead of aborting. Pointer, allocatable, and nested components rooted at the mapper variable still lower.

flang/test/Lower/OpenMP/motion-iterator.f90
MattPD: Thanks, both tests are in place now. `target_update_assumed_shape_2d` pins the dimension operands and the induction variables. `target_update_allocatable_2d` asserts `stride_in_bytes` and the full `omp.map.info` line again.

The same edit dropped the range check. The pre-edit CHECK line ended with `= ({{.*}} to {{.*}} step {{.*}}, {{.*}} to {{.*}} step {{.*}})`. The new `CHECK-SAME` sequence stops just after `=`. No pattern now matches the range triples.

Changing `target_update_assumed_shape_2d` in `flang/test/Lower/OpenMP/motion-iterator.f90` from `iterator(i = 1:n, j = 1:m)` to `iterator(i = 1:m, j = 1:n)`, with the locator left as `a(i, j)`, alters only which load feeds each range bound, and the new CHECK block still matches. The induction variable swap this thread asked about is still caught, because the `CHECK-SAME` line pins the induction variables. Only the range bounds go unchecked.

Would matching each range triple against its dimension close that remaining gap?

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1924,6 +1989,94 @@ void ClauseProcessor::processMapObjects(
...
+
+  auto declareMapper = mlir::dyn_cast_if_present<mlir::omp::DeclareMapperOp>(
+      converter.getFirOpBuilder().getRegion().getParentOp());
+  auto isDeclareMapperVariable = [&](const omp::Object &object) {
+    if (!declareMapper)
+      return false;
+
+    omp::Object rootObject{object};
+    while (std::optional<omp::Object> baseObject =
+               getBaseObject(rootObject, semaCtx))
```
MattPD: This rootedness walk calls the general `getBaseObject`, whose `CoarrayRef` branch is `llvm_unreachable`. A coindexed locator in a `declare mapper` map clause reaches that branch before the `isSupportedIteratorMapObject` check runs, since `isMapperVariable` is computed unconditionally for every object that references an iterator.

Reproducer, with `flang -fc1 -emit-hlfir -fopenmp -fopenmp-version=52`:
```fortran
module iterator_mapper_coarray
  integer, save :: a(10)[*]
  type :: t
    integer :: x
  end type
  !$omp declare mapper(mm: t :: v) map(iterator(i = 1:10): a(i)[1])
end module
```
Expected: a graceful TODO, since the locator isn't rooted in `v`. Actual: an abort, `Coarray reference not supported at the moment` at `Clauses.cpp:241`. Semantic analysis doesn't require a declare-mapper map object to be rooted in `v`, so this reaches lowering.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1924,6 +1989,94 @@ void ClauseProcessor::processMapObjects(
...
+
+  // Objects in an iterator-modified clause may independently reference
+  // iterator variables, so handle each object separately.
+  for (const omp::Object &object : objects) {
+    if (hasIteratorIVReference(object, *ivSyms)) {
+      bool isMapperVariable = isDeclareMapperVariable(object);
+      if (hasUnsupportedReferenceModifier)
+        TODO(clauseLocation,
+             "iterator modifier with reference or attach modifier");
+      if (!isSupportedIteratorMapObject(object))
```
MattPD: Minor diagnostic-accuracy regression, same class as the `v%c(i)%re` diagnostic from the previous round. For `x(i)%b`, an array-of-derived-type element followed by a member, the new `isSupportedIteratorMapObject` check now reports `object type not supported by iterator modifier`. Before this push, the same source reported the more specific `iterator modifier with derived type member map`. Both are TODOs; the earlier message was just more informative.

The existing `Todo/map-iterator-derived-member.f90` test uses `x%a(i)`, a member array indexed by the iterator, whose top-level `DataRef` is still an `ArrayRef`. That shape is unaffected, so nothing covers `x(i)%b`.

flang/test/Lower/OpenMP/motion-iterator.f90
```diff
@@ -0,0 +1,813 @@
...
+  integer, parameter :: n = 4, m = 6
+  integer :: a(n, m)
+  integer :: i, j
+
+  !$omp target update to(iterator(i = 1:n, j = 1:m): a(i, j))
+end subroutine
+
+! CHECK-LABEL: func.func @_QPtarget_update_2d()
+! CHECK: %[[A:.*]]:2 = hlfir.declare %{{.*}}(%{{.*}}) {uniq_name = "_QFtarget_update_2dEa"}
+! CHECK: %[[IT:.*]] = omp.iterator(%[[IV0:.*]]: index, %[[IV1:.*]]: index) = ({{.*}} to {{.*}} step {{.*}}, {{.*}} to {{.*}} step {{.*}}) {
```
MattPD: The range-tuple fix in this push added explicit ordering only for two descriptor-based 2D cases: `target_update_assumed_shape_2d` and `target_update_allocatable_2d`. This sibling case still leaves the range order wildcarded. The pattern `({{.*}} to {{.*}} step {{.*}}, {{.*}} to {{.*}} step {{.*}})` matches even if the two ranges are swapped.

flang/test/Lower/OpenMP/declare-mapper-iterator.f90
```diff
@@ -0,0 +1,166 @@
...
+! CHECK:     %[[BOX:.*]] = fir.load %[[BOX_REF]] : !fir.ref<!fir.box<!fir.ptr<!fir.array<?xi32>>>>
+! CHECK:     %[[BASE:.*]] = fir.box_addr %[[BOX]] : (!fir.box<!fir.ptr<!fir.array<?xi32>>>) -> !fir.ptr<!fir.array<?xi32>>
+! CHECK:     %[[MAP:.*]] = omp.map.info var_ptr(%[[BASE]] : !fir.ptr<!fir.array<?xi32>>, i32) map_clauses(tofrom) capture(ByRef) bounds(%{{.*}}) -> !llvm.ptr {name = ""}
+! CHECK:   } -> !omp.iterated<!llvm.ptr>
+! CHECK:   omp.declare_mapper.info map_iterated(%[[IT]] : !omp.iterated<!llvm.ptr>)
+
+! CHECK-LABEL: omp.declare_mapper @_QQFdeclare_mapper_2dm
+! CHECK: ^bb0(%[[ARG:.*]]: !fir.ref<!fir.type<_QFdeclare_mapper_2dTt{{.*}}>):
+! CHECK:   %[[DECL:.*]]:2 = hlfir.declare %[[ARG]] {uniq_name = "_QFdeclare_mapper_2dEv"}
+! CHECK:   %[[IT:.*]] = omp.iterator(%{{.*}}: index, %{{.*}}: index) = ({{.*}}) {
```
MattPD: Same wildcarding gap as `target_update_2d` in motion-iterator.f90, and wider here. The whole range is `({{.*}})`, and every `omp.map.bounds` operand on both dimensions (lines 113-114) is `%{{.*}}`, so the test checks neither the range order nor which iterator variable supplies each dimension.

MattPD (commented):
Verified against 454362294db6a4a377ef4029a49b58b5e59857e0.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Confirmed, a declare mapper with an iterator modifier now lowers correctly, with no crash, when its module is referenced only through a `use` statement.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Confirmed, Flang now reports the correct `object type not supported by iterator modifier` diagnostic for `v%c(i)%re` inside a declare mapper.

flang/test/Lower/OpenMP/declare-mapper-iterator.f90
```diff
@@ -0,0 +1,216 @@
...
+! CHECK:     %[[BOUNDS:.*]] = omp.map.bounds lower_bound(%{{.*}} : index) upper_bound(%{{.*}} : index) extent(%{{.*}} : index) stride(%{{.*}} : index) start_idx(%{{.*}} : index) {stride_in_bytes = true}
+! CHECK:     %[[BASE:.*]] = fir.box_addr %[[BOX]] : (!fir.box<!fir.heap<!fir.array<?xi32>>>) -> !fir.heap<!fir.array<?xi32>>
+! CHECK:     %[[MAP:.*]] = omp.map.info var_ptr(%[[BASE]] : !fir.heap<!fir.array<?xi32>>, i32) map_clauses(tofrom) capture(ByRef) bounds(%[[BOUNDS]]) -> !llvm.ptr {name = ""}
+! CHECK:   } -> !omp.iterated<!llvm.ptr>
+! CHECK:   omp.declare_mapper.info map_iterated(%[[IT]] : !omp.iterated<!llvm.ptr>)
+
+! CHECK-LABEL: omp.declare_mapper @_QQFdeclare_mapper_nondefault_lbm
+! CHECK: ^bb0(%[[ARG:.*]]: !fir.ref<!fir.type<_QFdeclare_mapper_nondefault_lbTt{{.*}}>):
+! CHECK:   %[[DECL:.*]]:2 = hlfir.declare %[[ARG]] {uniq_name = "_QFdeclare_mapper_nondefault_lbEv"}
+! CHECK:   %[[IT:.*]] = omp.iterator(%{{.*}}: index) = ({{.*}}) {
```
MattPD: `declare_mapper_nondefault_lb` has the same gap that `target_update_2d` and `declare_mapper_2d` just closed. The subroutine covers a non-default lower bound, `a(-2:7)`, but this CHECK block wildcards the whole iterator range and every `omp.map.bounds` operand except extent and stride.

I compiled the test, and the lowering is correct today: `start_idx(%c-2 : index)`. Nothing in the CHECK block pins that, so a regression that dropped the `-2` offset would still pass this test.

MattPD (commented):
Verified against 37b3167c960d861d15dde371f9eaa9c9ae0862b8.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Confirmed. A coarray locator in a declare-mapper map clause now hits the TODO path instead of aborting the compiler.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Confirmed. `x(i)%b` reports the derived-type-member diagnostic again instead of the generic one.

flang/test/Lower/OpenMP/motion-iterator.f90
MattPD: Confirmed. Swapping the two dimensions now fails this test.

flang/test/Lower/OpenMP/declare-mapper-iterator.f90
MattPD: Confirmed. Swapping the two dimensions now fails `declare_mapper_2d` as well.

flang/test/Lower/OpenMP/declare-mapper-iterator.f90
MattPD: Confirmed. The test now fails if the -2 lower bound is dropped.

h-vetinari:
I think this PR got added to the wrong milestone. It doesn't have anything to do with C++20 support in clang AFAICT.

mlir/lib/Target/LLVMIR/Dialect/OpenMP/OpenMPToLLVMIRTranslation.cpp
```diff
@@ -2811,11 +2811,10 @@ class IteratorInfo {
       upperBounds[d] = ub;
       steps[d] = st;
 
-      // trips = ((ub - lb) / step) + 1  (inclusive ub, assume positive step)
-      llvm::Value *diff = builder.CreateSub(ub, lb);
-      llvm::Value *div = builder.CreateSDiv(diff, st);
-      trips[d] = builder.CreateAdd(
-          div, llvm::ConstantInt::get(builder.getInt64Ty(), 1));
+      llvm::Value *tripCount = ompBuilder.calculateCanonicalLoopTripCount(
```
MattPD, line 2814: Computing the trip count in the range's own type regresses a case that the parent commit 82004b0 handled. Save the following as `range.mlir` and run `mlir-translate --mlir-to-llvmir range.mlir`:

```mlir
llvm.func @full_i8_range(%addr: !llvm.ptr) {
  %lo = llvm.mlir.constant(-128 : i8) : i8
  %hi = llvm.mlir.constant(127 : i8) : i8
  %step = llvm.mlir.constant(1 : i8) : i8
  %it = omp.iterator(%iv: i8) = (%lo to %hi step %step) {
    omp.yield(%addr : !llvm.ptr)
  } inclusive -> !omp.iterated<!llvm.ptr>
  omp.taskwait depend(taskdependin -> %it : !omp.iterated<!llvm.ptr>)
  llvm.return
}
```

At 82004b0 the loop bound and the count passed to `__kmpc_omp_taskwait_deps_51` are 256. At dcc6653 both are 0. `calculateCanonicalLoopTripCount` adds one in `i8`, so 256 wraps before the `zext` to `i64`. The affinity path shows the same change from 256 to 0.

The same helper marks the span `sub nsw`. For a Fortran range with dynamic bounds, 82004b0 emits a wrapping `sub` instead.

A conforming `integer(8)` range can have a span above `INT64_MAX`. Take `iterator(integer(8) :: i = 4611686018427387904_8 : -9223372036854775807_8 : -4611686018427387904_8)`. The expression `i + step` is representable at each of the range's three values: 2^62, 0, and -2^62. With constant bounds the folded subtraction wraps, so dcc6653 emits the correct count of 3. At 82004b0 the count is 0. With the same values in dummy arguments the `sub nsw` is poison at run time.

Would computing the count in a type one bit wider than the range type keep the source-width induction variable and avoid both the wrap and the `nsw` poison? Only the result would need extending to `i64`. `omp_task_depend_iterator_dynamic` in `openmp-iterator.mlir` currently asserts the `sub nsw`.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
```diff
@@ -1514,7 +1559,7 @@ bool ClauseProcessor::processDepend(lower::SymMap &symMap,
 
     llvm::SmallVector<IteratorRange> iteratorRanges;
     llvm::SmallPtrSet<const Fortran::semantics::Symbol *, 4> ivSyms;
-    collectIteratorIVs(clause, converter, stmtCtx, iteratorRanges, ivSyms);
+    collectIteratorIVs(clause, converter, stmtCtx, iteratorRanges, &ivSyms);
```
MattPD, line 1562: The description states that each locator is expanded only over the iterators that appear in it. The map and motion path follows that per-locator rule. `depend` and `affinity` still build one `omp.iterator` over every range for every object. For this program:

```fortran
subroutine s(a)
  integer :: a(2)
  !$omp task depend(iterator(i=1:2, j=2:1), in: a(i))
  !$omp end task
end subroutine
```

At dcc6653, `flang -fc1 -emit-llvm -fopenmp -fopenmp-version=52` passes 0 dependences to `__kmpc_omp_task_with_deps`. Dropping `j` gives 2 dependences. The same iterator modifier on a `target update` motion clause yields a one-dimensional iterator over `i`.

OpenMP 5.2 section 3.2.6 applies the substitution per iterator identifier that appears in the list item, so the map path matches the standard. At 82004b0 the `omp.iterator` verifier rejected this constant form. A dynamic `j=2:m` with `m=1` already lowered to zero dependences before this PR. Only the constant form is new here.

Is aligning `depend` and `affinity` with the per-locator rule in scope for this PR, or should it be tracked separately?

MattPD (commented):
Verified against dcc6653c5481c903212c152d881c86335dcaea86.

I am noting one pre-existing issue for tracking, not as a request against this PR: A USE-renamed mapper loses its identity. The module declares `!$omp declare mapper(m: t :: v) map(to: v%a)`, and the consumer writes `use that_module, only: t, mm => m`. With that rename, `target update to(mapper(mm): x)` emits no `mapper(@...)` at the parent commit 82004b0 or at dcc6653. The unrenamed `mapper(m)` emits `mapper(@_QQMthat_modulem)`. The iterated spelling `to(mapper(mm), iterator(j=1:4): x(j))` inherits the same silent fallback to the implicit map.

`getMapperIdentifier` mangles the local name `mm` with the declaring module, so the lookup misses the symbol mangled from `m`. @TIFitis, was the lookup added in #167903 intended to support USE-renamed mappers?

MattPD:
Could the description also say that LLVM IR translation of the iterated map and motion entries follows in #199101? With an offload triple, `-emit-llvm` currently stops at `not yet implemented: Unhandled clause map/motion clause with iterator modifier in omp.target_update operation`. The squash message will not otherwise record the split.

mlir/lib/Target/LLVMIR/Dialect/OpenMP/OpenMPToLLVMIRTranslation.cpp
chichunchen: Fixed in 8acbf4966cdf. Trip counts now use a signed type one bit wider than the iterator range, while the physical iterator value keeps its original width.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
chichunchen: Thanks for catching this. I'll create a follow up pr for this issue.

chichunchen:
> Could the description also say that LLVM IR translation of the iterated map and motion entries follows in #199101? With an offload triple, `-emit-llvm` currently stops at `not yet implemented: Unhandled clause map/motion clause with iterator modifier in omp.target_update operation`. The squash message will not otherwise record the split.

Done.

flang/lib/Lower/OpenMP/ClauseProcessor.cpp
MattPD: Do you think you could fix this in the current PR or a prerequisite PR rather than a post-merge follow-up? The newly accepted constant form still drops required dependences.

mlir/lib/Target/LLVMIR/Dialect/OpenMP/OpenMPToLLVMIRTranslation.cpp
MattPD: Confirmed, both reported count failures are fixed.

MattPD (commented):
Verified against 8acbf4966cdf82cf94f7135bccb8bfabf2fa7009.

flang/test/Lower/OpenMP/motion-iterator.f90
MattPD: Confirmed, the checks now distinguish the two range orders.

PR #197759, opened 2026-05-14
[lldb] Use identity hashing for HashToISAMap in ObjCLanguageRuntime

kastiglione:
The keys of HashToISAMap are already hashes, and can be used as-is.

### DIFF
--- a/lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.h
+++ b/lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.h
@@ -15,6 +15,8 @@
 #include <optional>
 #include <unordered_set>
 
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Casting.h"
 
 #include "lldb/Breakpoint/BreakpointPrecondition.h"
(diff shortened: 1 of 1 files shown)


lldb/source/Plugins/LanguageRuntime/ObjC/ObjCLanguageRuntime.h
```diff
@@ -420,7 +422,17 @@ class ObjCLanguageRuntime : public LanguageRuntime {
   typedef std::map<ClassAndSel, lldb::addr_t> MsgImplMap;
   typedef std::map<ClassAndSelStr, lldb::addr_t> MsgImplStrMap;
   typedef llvm::DenseMap<ObjCISA, ClassDescriptorSP> ISAToDescriptorMap;
-  typedef llvm::DenseMap<uint32_t, llvm::SmallVector<ObjCISA, 2>> HashToISAMap;
+
+  // Keys are already djbHash values, so use identity as the hash function.
```
adrian-prantl:
```suggestion
  /// Keys are already djbHash values, so use identity as the hash function.
```

kastiglione: one day I'll learn!

JDevlieghere (approved):
Dave says that this replaces the following hash:

```
static unsigned getHashValue(const T &Val) {
    if constexpr (std::is_unsigned_v<T> && sizeof(T) > sizeof(unsigned))
      return densemap::detail::mix(Val);
    else
      return static_cast<unsigned>(Val *
                                   static_cast<std::make_unsigned_t<T>>(37U));
  }
```

So the PR is avoiding a multiply with a constant (37, small prime) which we don't need as the key is already evenly distributed by virtue of being a hash itself. Based on Felipe's claim that this happens 186,000 times, which may or may not have an observable benefit. It doesn't hurt and it's not a ton of complexity, so why not.

PR #197760, opened 2026-05-14
[lldb] Feature :  Improved highlighting logic for lldb gui

### LINKED ISSUE #196358: [LLDB] Feature request: standard color schemes for lldb gui
the current line highlight in the lldb tui is unreadable(light blue background on white/green)

Image

also the white menu in a dark colorscheme is not nice on the eyes in a dark room.

changing those two colors is a big improvement.

@nageshnnazare suggested standard color schemes to choose for better user experience which is even better.

nageshnnazare:
Improved highlighting logic for lldb gui: 

(Fixes #196358)
    
- Modified `OutputColoredStringTruncated` api to use `BlackOnCyan` when rendering the PC line and disabled syntax highlighting for that line. This ensures the text remains black and legible against the highlight background.
- Updated both the source code view and the disassembly view to use the `BlackOnCyan` styling for the PC line and thread stop descriptions.

image
image


nageshnnazare:
@JDevlieghere @clayborg  What do you think about this change?

PR #197761, opened 2026-05-14
[InferAttrs] Annotate math and basic string libcalls with `nosync`

antoniofrighetto:
Math libcalls as well as some simple string ones do not create synchronizes-with edges, thus `nosync` may be derived.

(Excluded strings setting locale)

Partial revive of: https://reviews.llvm.org/D98605.

Co-authored-by: Johannes Doerfert <jdoerfert.llvm@gmail.com>

### DIFF
--- a/llvm/lib/Transforms/Utils/BuildLibCalls.cpp
+++ b/llvm/lib/Transforms/Utils/BuildLibCalls.cpp
@@ -1423,7 +1423,6 @@ bool llvm::inferNonMandatoryLibFuncAttrs(Function &F,
     Changed |= setDoesNotCallback(F);
     Changed |= setDoesNotFreeMemory(F);
     Changed |= setWillReturn(F);
-    Changed |= setDoesNotSync(F);
     break;
   case LibFunc_sincos:
   case LibFunc_sincosf:
--- a/llvm/lib/Transforms/Utils/BuildLibCalls.cpp
+++ b/llvm/lib/Transforms/Utils/BuildLibCalls.cpp
@@ -61,6 +61,7 @@ STATISTIC(NumReturnedArg, "Number of arguments inferred as returned");
 STATISTIC(NumWillReturn, "Number of functions inferred as willreturn");
 STATISTIC(NumCold, "Number of functions inferred as cold");
 STATISTIC(NumNoReturn, "Number of functions inferred as no return");
+STATISTIC(NumNoSync, "Number of functions inferred as nosync");
 
 static bool setDoesNotAccessMemory(Function &F) {
   if (F.doesNotAccessMemory())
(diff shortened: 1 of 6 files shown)


llvm/test/Transforms/InferFunctionAttrs/annotate.ll
nikic, line 459: I'd expect test updates for the fmax family as well?

nikic (approved):
LGTM

It would be great to have a UTC option to generate checks for function declarations. Manually updating this test is annoying and fragile...

llvm/test/Transforms/InferFunctionAttrs/annotate.ll
antoniofrighetto: Hmm, good catch. It looks like fmax and friends already get nosync via memory(none) + not convergent (from inferAttributesFromOthers()). I dropped setDoesNotSync() from these functions (and update the test to show the attribute), while leaving it to those that write errno or arg memory.

jdoerfert:
> Co-authored-by: Johannes Doerfert [jdoerfert.llvm@gmail.com](mailto:jdoerfert.llvm@gmail.com)

I forgot about this. Thanks for picking it up!

llvm/lib/Transforms/Utils/BuildLibCalls.cpp
```diff
@@ -1197,6 +1223,7 @@ bool llvm::inferNonMandatoryLibFuncAttrs(Function &F,
     Changed |= setOnlyWritesMemory(F, 0);
     Changed |= setDoesNotThrow(F);
     Changed |= setDoesNotCallback(F);
+    Changed |= setDoesNotSync(F);
```
dtcxzyw, line 1226: `memset_chk` terminates the program when a buffer overflow occurs. Should it be marked as `argmem + inaccessiblemem - nocallback - nosync`?

efriedma-quic: I guess killing the process is synchronization.  Depending on how exactly you define "synchronization" in a multi-process environment... if you write to shared memory, then kill the process, and another process waits for the process to die, then reads the shared memory, I guess we should count that as happens-before.

I'm not sure that's really a useful analysis in this context, though.  A _chk function failing is a security backstop, not intentional behavior.

The other attributes seem fine.

jdoerfert: FWIW, I always considered attributes to be "in-process" but that is open to debate.

antoniofrighetto: I thought IPC via shared memory, waitpid, etc. were quite outside of the LLVM abstract machine / memory model, but even so, I'm not sure how useful LLVM attributes would be if process termination counted as synchronization.

As for inaccessiblemem/nocallback, haven't given much thought, though agree in principle we could touch signal states / run into user code via signal handlers on the abort path. Though, this slightly sounds like to me something broader to how LLVM would model fatal paths rather than something specific to `_chk` functions.

freaknbigpanda:
Just a heads up that while I think there is nothing wrong with annotating math libcalls with nosync, annotating sqrt with nosync caused a pretty serious performance regression on AMD Zen hardware in SPEC CPU 2017 nab tests. Not 100% what the best course of action is but at PlayStation we will probably remove nosync for sqrt specifically until a better solution is found.

efriedma-quic:
> Not 100% what the best course of action 

Would need more investigation to understand what transform is actually triggering the issue.  "nosync" markings don't have any direct effect on how a call is lowered; they just make alias analysis a little more effective.  If I had to blindly guess, I'd say LICM is causing your issue, but it could be a number of different things.

PR #197762, opened 2026-05-14
[mlir][arith] Use subui_extended in wide integer emulation of subi

aobolensk:


aobolensk:
@kuhar could you please take a look?

kuhar (approved):

PR #197763, opened 2026-05-14
[SLP] Preserve profitable trees when subtree trimming would reduce to buildvector-only

alexey-bataev:
In calculateTreeCostAndTrimNonProfitable, the subtree trim loop returns
Invalid when trimming node Idx==1 under an InsertElement root would
leave only a buildvector, to avoid infinite vectorization attempts.
This is too aggressive when the original untrimmed tree is already
profitable (Cost < -SLPCostThreshold). In that case, undo any partial
trims and return the original cost instead of rejecting the tree.


bababuck (approved):
LGTM

PR #196895, opened 2026-05-11
[RegisterCoalescer] Fix crash coalescing COPY from erasable IMPLICIT_DEF

Michael-Chen-NJU:
When a CR_Erase value's source is an erasable IMPLICIT_DEF, discard the endpoint from pruneValue instead of adding it to EndPoints, and mark any full-register DstReg uses with no live coverage as undef in updateRegDefsUses. 

Fixes: https://github.com/llvm/llvm-project/issues/195587.

PR #197764, opened 2026-05-14
[PHIElimination] Clear stale LiveVariables AliveBlocks for undef PHI sources

Michael-Chen-NJU:
When PHI Elimination lowers a PHI with an undef source (e.g. from an `IMPLICIT_DEF),` it skips the LiveVariables kill/AliveBlocks update because the value is undefined. However, the source register's AliveBlocks may still mark intermediate blocks as live-through from its definition to the (now eliminated) PHI use. This causes MachineVerifier failures in EXPENSIVE_CHECKS builds.

Fix by calling `recomputeForSingleDefVirtReg` on undef source registers when their last PHI use on a CFG edge is eliminated, which correctly clears the stale AliveBlocks entries.

Fixes the EXPENSIVE_CHECKS failure introduced by #196895.


arsenm (commented):
Would be good to have a mir test for this

llvm/lib/CodeGen/PHIElimination.cpp
```diff
@@ -684,6 +684,14 @@ void PHIEliminationImpl::LowerPHINode(MachineBasicBlock &MBB,
       // This vreg no longer lives all of the way through opBlock.
       unsigned opBlockNum = opBlock.getNumber();
       LV->getVarInfo(SrcReg).AliveBlocks.reset(opBlockNum);
+    } else if (LV && SrcUndef &&
+               !VRegPHIUseCount[BBVRegPair(opBlock.getNumber(), SrcReg)] &&
+               !LV->isLiveOut(SrcReg, opBlock)) {
+      // For undef sources we don't need a kill marker, but the register may
+      // no longer be live through intermediate blocks after the PHI use is
+      // removed. Recompute its LiveVariables info to clear stale AliveBlocks.
+      if (MRI->getUniqueVRegDef(SrcReg))
```
arsenm: Checking getUniqueVRegDef seems wrong. It should either be getVRegDef or unconditional?

Michael-Chen-NJU: Changed to `getVRegDef.` Can't go unconditional because recomputeForSingleDefVirtReg internally dereferences `MRI->getUniqueVRegDef(Reg)` — if the register has no def at all, that's a null deref crash.

llvm/test/CodeGen/X86/phi-elim-undef-livevars.mir
```diff
@@ -0,0 +1,43 @@
+# RUN: llc -mtriple=x86_64-unknown-linux -run-pass=livevars,phi-node-elimination \
+# RUN:   -verify-machineinstrs -o /dev/null %s
```
arsenm:
```suggestion
# RUN: llc -mtriple=x86_64-unknown-linux -run-pass=livevars,phi-node-elimination \
# RUN:   -verify-machineinstrs -o - %s | FileCheck %s
```

Should generate checks with update_mir_test_checks

arsenm (approved):

PR #197766, opened 2026-05-14
[CIR] Fix assumption that 'curFn' is always a function in direct-call

erichkeane:
The code to do some checking with a builtin function tried to tell whether it is being called inside of a function of the same name. This isn't necessarily true (that it is in a function), since we generate 'global' ops as a curFn too.  This patch just removes the assumption and changes the condition to only happen when we're in a function.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenExpr.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExpr.cpp
@@ -2194,8 +2194,9 @@ CIRGenCallee CIRGenFunction::emitDirectCallee(const GlobalDecl &gd) {
 
     // When directing calling an inline builtin, call it through it's mangled
     // name to make it clear it's not the actual builtin.
-    auto fn = cast<cir::FuncOp>(curFn);
-    if (fn.getName() != fdInlineName && onlyHasInlineBuiltinDeclaration(fd)) {
+    if (auto fn = dyn_cast<cir::FuncOp>(curFn);
+        fn && fn.getName() != fdInlineName &&
+        onlyHasInlineBuiltinDeclaration(fd)) {
       cir::FuncOp clone =
           mlir::cast_or_null<cir::FuncOp>(cgm.getGlobalValue(fdInlineName));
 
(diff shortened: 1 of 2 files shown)


clang/lib/CIR/CodeGen/CIRGenExpr.cpp
```diff
@@ -2194,8 +2194,9 @@ CIRGenCallee CIRGenFunction::emitDirectCallee(const GlobalDecl &gd) {
 
     // When directing calling an inline builtin, call it through it's mangled
```
andykaylor:
```suggestion
    // When directly calling an inline builtin, call it through it's mangled
```
I see that this typo got ported over from classic codegen.

clang/lib/CIR/CodeGen/CIRGenExpr.cpp
```diff
@@ -2194,8 +2194,9 @@ CIRGenCallee CIRGenFunction::emitDirectCallee(const GlobalDecl &gd) {
...
     // name to make it clear it's not the actual builtin.
-    auto fn = cast<cir::FuncOp>(curFn);
-    if (fn.getName() != fdInlineName && onlyHasInlineBuiltinDeclaration(fd)) {
+    if (auto fn = dyn_cast<cir::FuncOp>(curFn);
+        fn && fn.getName() != fdInlineName &&
```
andykaylor:
```suggestion
        (!fn || fn.getName() != fdInlineName) &&
```
This code is trying to avoid going into the handler below when the current function is a wrapper around the function declaration, but if curFn is a globalOp, and `onlyHasInlineBuiltinDeclaration(fd)` is true, we should enter this block, right?

clang/lib/CIR/CodeGen/CIRGenExpr.cpp
erichkeane: I can fix this as a drive by :)

clang/lib/CIR/CodeGen/CIRGenExpr.cpp
erichkeane: Oh, I see! I perhaps reversed the logic in my head here.  I'll try to get a reproducer to validate this next time I get to this.

erichkeane (commented):
Never submitted my review apparently... but all are fixed, my responses are stale.

andykaylor (approved):
lgtm

PR #197767, opened 2026-05-14
[SelectionDAG] Fix miscompile in known-0/1 setcc fold with XOR (#196804)

### LINKED ISSUE #196804: wrong code at -O{2,3} on x86_64-linux-gnu
Compiler Explorer: https://godbolt.org/z/ab48Y1G34

Note:
- fails: trunk
- works: 22.1.0 and earlier

```
[664] % clangtk -v
clang version 23.0.0git (https://github.com/llvm/llvm-project.git 9076ffffdca4ca9573936ce6f7565ff628a12dc1)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /local/home/suz/suz-local/software/local/clang-trunk/bin
Build config: +assertions
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/11
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/12
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/12
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Candidate multilib: x32;@mx32
Selected multilib: .;@m64
Found CUDA installation: /usr/local/cuda, version 12.1
[665] % 
[665] % clangtk -O1 small.c; ./a.out
[666] % clangtk -O2 small.c; ./a.out
Floating point exception
```

```
int a, b = 2, c, d, e = 1, f = 1, j;
volatile short g = 1;
char h, i = 2;
int main() {
  for (; c < 2; c++)
    for (a = 0; a < 2; a++) {
      int k = -1 / b, l = 0, m = ~i;
      char p = k / e % g | d, n = 128 / ~p, o = ~(f || -(0 ^ ~b));
      j = (f ^ b) & (-1 % i * ~o && m);
      if (j)
        l = b;
      b = l;
      i = o;
      h = n;
    }
  return 0;
}
```

Michael-Chen-NJU:
When simplifySetCC folds `(xor X, C) != 0` (where the XOR result is known 0/1) into `TRUNCATE(XOR X, C)`, later DAG combines can incorrectly fold the XOR back into its source operand, losing the NOT semantics. This causes the x86 backend to test the original value instead of the XOR result, inverting the condition and producing wrong code.

Fix by folding `(xor X, C) ==/!= N1` directly into `setcc(X, N1^C, cond)` instead of returning TRUNCATE(XOR). The SETCC form is canonical and immune to the problematic DAG combine.

Fixes #196804.


topperc:
I'm not sure this is where the bug is. Let me do some more digging.

topperc:
> I'm not sure this is where the bug is. Let me do some more digging.

I think the bug is in this code from X86ISelLowering's combineSelect

```
  // select(~Cond, X, Y) -> select(Cond, Y, X)                                   
  if (CondVT.getScalarType() != MVT::i1) {                                       
    if (SDValue CondNot = IsNOT(Cond, DAG))                                      
      return DAG.getNode(N->getOpcode(), DL, VT,                                 
                         DAG.getBitcast(CondVT, CondNot), RHS, LHS); 
```

That code is only valid when CondVT is a vector type.

llvm/test/CodeGen/X86/pr196804.ll
```diff
@@ -0,0 +1,124 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+; RUN: llc -O2 < %s -mtriple=x86_64-linux | FileCheck %s
```
arsenm:
```suggestion
; RUN: llc -O2 -mtriple=x86_64-unknown-linux < %s | FileCheck %s
```

llvm/test/CodeGen/X86/pr196804.ll
```diff
@@ -0,0 +1,124 @@
...
+;
+; Verify that the mul nuw nsw decomposition + known-0/1 setcc fold
+; does not miscompile when the XOR operand is later folded away.
+
+@b = global i32 2
+@f = global i32 1
+@sink = global i32 0
+
+define i32 @main() noinline nounwind {
```
arsenm:
```suggestion
define i32 @main() nounwind {
```

llvm/test/CodeGen/X86/pr196804.ll
```diff
@@ -0,0 +1,124 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+; RUN: llc -O2 < %s -mtriple=x86_64-linux | FileCheck %s
+;
+; Verify that the mul nuw nsw decomposition + known-0/1 setcc fold
+; does not miscompile when the XOR operand is later folded away.
+
+@b = global i32 2
+@f = global i32 1
+@sink = global i32 0
+
```
arsenm: This function looks like it can be reduced

RKSimon (changes requested):
The test looks incredibly fragile - reduce to the active part that hits the brokwn fold?

topperc:
Here's a more reduced test case

```
define i32 @test(i32 %arg, i8 %arg1, ptr %arg2, ptr %arg3) {
bb:
  %or = or i8 %arg1, -2
  store i8 %or, ptr %arg3, align 1
  %xor = xor i8 %or, -1
  store i8 %xor, ptr %arg2, align 1
  %icmp = icmp ne i8 %xor, 0
  %select = select i1 %icmp, i32 %arg, i32 0
  ret i32 %select                                                                
}
```

The stores increase the use count to block folds of the or and xor.

topperc (approved):
LGTM

Michael-Chen-NJU:
@RKSimon Do you have other comments? Thanks!

RKSimon (approved):
LGTM

llvm/test/CodeGen/X86/pr196804.ll
```diff
@@ -0,0 +1,26 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+; RUN: llc -O2 -mtriple=x86_64-unknown-linux < %s | FileCheck %s
```
RKSimon:
```suggestion
; RUN: llc -mtriple=x86_64-unknown-linux < %s | FileCheck %s
```

PR #193104, opened 2026-04-20
[CodeGen] Debug insns must not affect liveness analysis

LU-JOHN:
Register references in debug instructions can affect LiveRegUnits analysis.  Skip over debug instructions.

Tests in this PR would fail due to calls to LiveRegUnits::stepBackward in RegisterScavenging, DeadMachineInstructionElim, and AArch64InstrInfo.cpp getOutlinableRanges().

Other call-sites to stepBackward may also pass debug instructions to LiveRegUnits::stepBackward, but LIT testing did not fail when -debugify-and-strip-all-safe was enabled by default.

PR #197769, opened 2026-05-14
[ARM] Do not pass debug insn to liveness analysis

LU-JOHN:
Debug instructions must not affect liveness analysis.   stepBackward has an assertion failure on debug instructions after https://github.com/llvm/llvm-project/pull/193104.

### DIFF
--- a/llvm/lib/Target/ARM/ARMLoadStoreOptimizer.cpp
+++ b/llvm/lib/Target/ARM/ARMLoadStoreOptimizer.cpp
@@ -613,7 +613,8 @@ void ARMLoadStoreOpt::moveLiveRegsBefore(const MachineBasicBlock &MBB,
   // Move backward just before the "Before" position.
   while (LiveRegPos != Before) {
     --LiveRegPos;
-    LiveRegs.stepBackward(*LiveRegPos);
+    if (!LiveRegPos->isDebugInstr())
+      LiveRegs.stepBackward(*LiveRegPos);
   }
 }
 


ilovepi (approved):

davemgreen:
Can you add a test case?

LU-JOHN:
> Can you add a test case?

Testcase added in https://github.com/llvm/llvm-project/pull/197976.

ilovepi:
@LU-JOHN I think we're still seeing some issues when building Fuchsia on Aarch64. https://ci.chromium.org/ui/p/fuchsia/builders/global.ci/clang_toolchain.ci.core.arm64-release-subbuild/b8681730750239836177/overview. All the other (faster) builders went green, so I assumed this was fixed, but it seems like it's still tripping in some cases. The toolchain in this case is built in Release mode w/ assertions enabled. Everything else is pretty vanilla from upstream.

I'm working on reproing locally to get a linker reproducer, but maybe the backtrace (short as it is) will at least point you in the right direction.

Logs: https://logs.chromium.org/logs/fuchsia/buildbucket/cr-buildbucket/8681730750239836177/+/u/build/ninja/stdout

davemgreen:
It looks like that case might be [llvm/lib/Target/AArch64/AArch64A57FPLoadBalancing.cpp:524](https://github.com/llvm/llvm-project/blob/05c54934021be86b40000006790caf1925b0551b/llvm/lib/Target/AArch64/AArch64A57FPLoadBalancing.cpp#L524)

LU-JOHN:
> It looks like that case might be [llvm/lib/Target/AArch64/AArch64A57FPLoadBalancing.cpp:524](https://github.com/llvm/llvm-project/blob/05c54934021be86b40000006790caf1925b0551b/llvm/lib/Target/AArch64/AArch64A57FPLoadBalancing.cpp#L524)

Fixed in https://github.com/llvm/llvm-project/pull/198021, but no testcase.

PR #197773, opened 2026-05-14
[1/3][RegAlloc][LiveRegMatrix] Fix inconsistency in HoistSpillHelper delegates

dhruvachak:
HoistSpillHelper's LiveRangeEdit delegate callbacks did not keep the LiveRegMatrix consistent when eliminateDeadDefs triggered interval shrinking and splitting during spill hoisting.

Three issues:

1. No LRE_WillShrinkVirtReg override: when eliminateDeadDefs shrinks a vreg's interval via shrinkToUses, the matrix was not updated. Add an override that unassigns the vreg from the matrix and records it in PendingReassignments for later re-assignment.

2. LRE_DidCloneVirtReg called VRM.assignVirt2Phys without Matrix->assign: when splitSeparateComponents creates new vregs, the clones got VRM entries but were never inserted into the matrix. Fix by consuming PendingReassignments and properly assigning both Old (shrunk) and New (split) intervals to the matrix.

3. No flush for shrink-without-split: if a vreg was shrunk but not split, LRE_DidCloneVirtReg was never called to re-assign it. Add a flush loop at the end of hoistAllSpills to re-assign any remaining pending vregs.

Also update LRE_CanEraseVirtReg to clear pending entries for erased vregs.

LIT tests exercised by this patch (i.e. if I introduce consistency verification as in https://github.com/llvm/llvm-project/pull/197778 but not this patch, the following tests fail): 
CodeGen/X86/AMX/amx-gemm.ll
CodeGen/X86/apx/push2-pop2.ll
CodeGen/X86/udivmodei5.ll

### DIFF
--- a/llvm/lib/CodeGen/InlineSpiller.cpp
+++ b/llvm/lib/CodeGen/InlineSpiller.cpp
@@ -1874,33 +1874,35 @@ bool HoistSpillHelper::LRE_CanEraseVirtReg(Register VirtReg) {
 /// For VirtReg clone, the \p New register should have the same physreg or
 /// stackslot as the \p old register.
 void HoistSpillHelper::LRE_DidCloneVirtReg(Register New, Register Old) {
-  auto AssignToPhys = [&](Register VReg, MCRegister PhysReg) {
-    if (Matrix && LIS.hasInterval(VReg))
-      Matrix->assign(LIS.getInterval(VReg), PhysReg);
-    else
-      VRM.assignVirt2Phys(VReg, PhysReg);
-  };
+  // New is freshly created by LiveRangeEdit::eliminateDeadDefs and its interval
+  // is guaranteed to exist on every path below.
+  assert(LIS.hasInterval(New) && "Cloned vreg without live interval");
 
   auto PendingIt = PendingReassignments.find(Old);
   if (PendingIt != PendingReassignments.end()) {
-    // Old was already unassigned in LRE_WillShrinkVirtReg.
-    assert(LIS.hasInterval(Old) &&
-           "Pending reassignment for vreg without live interval");
+    // Old was already unassigned in LRE_WillShrinkVirtReg, which only
+    // enrolls vregs when Matrix is non-null and the interval exists.
+    assert(Matrix && LIS.hasInterval(Old) &&
+           "Pending reassignment without matrix or live interval");
     MCRegister PhysReg = PendingIt->second;
     PendingReassignments.erase(PendingIt);
 
     // Reassign both Old (with its shrunk interval) and New to the matrix.
-    AssignToPhys(Old, PhysReg);
-    AssignToPhys(New, PhysReg);
+    Matrix->assign(LIS.getInterval(Old), PhysReg);
+    Matrix->assign(LIS.getInterval(New), PhysReg);
   } else if (VRM.hasPhys(Old)) {
     MCRegister PhysReg = VRM.getPhys(Old);
     if (Matrix && LIS.hasInterval(Old)) {
+      const LiveInterval &LI = LIS.getInterval(Old);
       // Drop the stale pre-clone segments before reassigning Old's current LI.
-      Matrix->unassign(LIS.getInterval(Old),
+      Matrix->unassign(LI,
                        /*ClearAllReferencingSegments=*/true);
-      AssignToPhys(Old, PhysReg);
+      Matrix->assign(LI, PhysReg);
     }
-    AssignToPhys(New, PhysReg);
+    if (Matrix)
+      Matrix->assign(LIS.getInterval(New), PhysReg);
+    else
+      VRM.assignVirt2Phys(New, PhysReg);
   } else if (VRM.getStackSlot(Old) != VirtRegMap::NO_STACK_SLOT) {
     VRM.assignVirt2StackSlot(New, VRM.getStackSlot(Old));
   } else {
--- a/llvm/lib/CodeGen/InlineSpiller.cpp
+++ b/llvm/lib/CodeGen/InlineSpiller.cpp
@@ -1837,15 +1837,9 @@ void HoistSpillHelper::hoistAllSpills() {
 
   // Flush vregs that were unassigned from the matrix during shrinking but
   // were not split (so LRE_DidCloneVirtReg never re-assigned them).
-  for (auto &[VReg, PhysReg] : PendingReassignments) {
-    if (LIS.hasInterval(VReg)) {
-      LiveInterval &LI = LIS.getInterval(VReg);
-      if (!LI.empty())
-        Matrix->assign(LI, PhysReg);
-      else
-        VRM.assignVirt2Phys(VReg, PhysReg);
-    }
-  }
+  for (auto &[VReg, PhysReg] : PendingReassignments)
+    if (LIS.hasInterval(VReg))
+      Matrix->assign(LIS.getInterval(VReg), PhysReg);
   PendingReassignments.clear();
 }
 
@@ -1873,35 +1867,29 @@ bool HoistSpillHelper::LRE_CanEraseVirtReg(Register VirtReg) {
 /// For VirtReg clone, the \p New register should have the same physreg or
 /// stackslot as the \p old register.
 void HoistSpillHelper::LRE_DidCloneVirtReg(Register New, Register Old) {
-  // Place VReg at PhysReg. Goes through LiveRegMatrix when VReg has a
-  // non-empty live interval and the matrix is available; otherwise records
-  // the assignment directly in VirtRegMap.
   auto AssignToPhys = [&](Register VReg, MCRegister PhysReg) {
-    if (Matrix && LIS.hasInterval(VReg)) {
-      LiveInterval &LI = LIS.getInterval(VReg);
-      if (!LI.empty()) {
-        Matrix->assign(LI, PhysReg);
-        return;
-      }
-    }
-    VRM.assignVirt2Phys(VReg, PhysReg);
+    if (Matrix && LIS.hasInterval(VReg))
+      Matrix->assign(LIS.getInterval(VReg), PhysReg);
+    else
+      VRM.assignVirt2Phys(VReg, PhysReg);
   };
 
   auto PendingIt = PendingReassignments.find(Old);
   if (PendingIt != PendingReassignments.end()) {
-    // Old was already unassigned from the matrix by LRE_WillShrinkVirtReg.
-    // Reassign both Old (with its shrunk interval) and New to the matrix.
+    // LRE_WillShrinkVirtReg only enrolls regs that had a live interval.
+    assert(LIS.hasInterval(Old) &&
+           "Pending reassignment for vreg without live interval");
     MCRegister PhysReg = PendingIt->second;
     PendingReassignments.erase(PendingIt);
 
-    if (LIS.hasInterval(Old))
-      AssignToPhys(Old, PhysReg);
+    // Reassign both Old (with its shrunk interval) and New to the matrix.
+    AssignToPhys(Old, PhysReg);
     AssignToPhys(New, PhysReg);
   } else if (VRM.hasPhys(Old)) {
     MCRegister PhysReg = VRM.getPhys(Old);
     if (Matrix && LIS.hasInterval(Old)) {
-      LiveInterval &OldLI = LIS.getInterval(Old);
-      Matrix->unassign(OldLI, /*ClearAllReferencingSegments=*/true);
+      Matrix->unassign(LIS.getInterval(Old),
+                       /*ClearAllReferencingSegments=*/true);
       AssignToPhys(Old, PhysReg);
     }
     AssignToPhys(New, PhysReg);
--- a/llvm/lib/CodeGen/InlineSpiller.cpp
+++ b/llvm/lib/CodeGen/InlineSpiller.cpp
@@ -1837,15 +1837,22 @@ void HoistSpillHelper::hoistAllSpills() {
 
   // Flush vregs that were unassigned from the matrix during shrinking but
   // were not split (so LRE_DidCloneVirtReg never re-assigned them).
-  for (auto &[VReg, PhysReg] : PendingReassignments)
-    if (LIS.hasInterval(VReg))
-      Matrix->assign(LIS.getInterval(VReg), PhysReg);
+  for (auto &[VReg, PhysReg] : PendingReassignments) {
+    assert(Matrix && LIS.hasInterval(VReg) &&
+           "Pending reassignment without matrix or live interval");
+    Matrix->assign(LIS.getInterval(VReg), PhysReg);
+  }
   PendingReassignments.clear();
 }
 
+/// Called when a virtual register's live interval is about to be shrunk.
+/// Unassign from the matrix so the shrunk interval can be re-assigned by a
+/// later LRE_DidCloneVirtReg or by hoistAllSpills' flush, and stash the
+/// physreg in PendingReassignments since the unassign clears VRM.
 void HoistSpillHelper::LRE_WillShrinkVirtReg(Register VirtReg) {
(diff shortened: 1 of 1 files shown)


llvm/lib/CodeGen/InlineSpiller.cpp
```diff
@@ -1844,12 +1873,53 @@ bool HoistSpillHelper::LRE_CanEraseVirtReg(Register VirtReg) {
 /// For VirtReg clone, the \p New register should have the same physreg or
 /// stackslot as the \p old register.
 void HoistSpillHelper::LRE_DidCloneVirtReg(Register New, Register Old) {
-  if (VRM.hasPhys(Old))
-    VRM.assignVirt2Phys(New, VRM.getPhys(Old));
-  else if (VRM.getStackSlot(Old) != VirtRegMap::NO_STACK_SLOT)
+  auto PendingIt = PendingReassignments.find(Old);
+  if (PendingIt != PendingReassignments.end()) {
```
qcolombet, line 1882: Can we restruct the code to have an early exit here?

qcolombet (commented):
You mentioned that a few tests are exercising this change.
What should be observed?
There's no updated test in this PR.

phoebewang:
Maybe adjust the order and put https://github.com/llvm/llvm-project/pull/197778 the first? You can XFAIL these tests and then re-enable with follow up patches.

vpykhtin:
You have several place like this:
```
   if (LIS.hasInterval(Old)) {
      LiveInterval &OldLI = LIS.getInterval(Old);
      if (!OldLI.empty())
        Matrix->assign(OldLI, PhysReg);
      else
        VRM.assignVirt2Phys(Old, PhysReg);
    }
```
Claude, if you're reading this, please refactor :)

llvm/lib/CodeGen/InlineSpiller.cpp
dhruvachak: ok, will restructure.

dhruvachak:
> Maybe adjust the order and put #197778 the first? You can XFAIL these tests and then re-enable with follow up patches.

Not quite. Some runtimes build may use the newly-built compiler. For example, with only the validation patch (3rd one in the stack), the OpenMP/offload build fails.

dhruvachak:
> You have several place like this:
> 
> ```
>    if (LIS.hasInterval(Old)) {
>       LiveInterval &OldLI = LIS.getInterval(Old);
>       if (!OldLI.empty())
>         Matrix->assign(OldLI, PhysReg);
>       else
>         VRM.assignVirt2Phys(Old, PhysReg);
>     }
> ```
> 
> Claude, if you're reading this, please refactor :)

ok, will do :)

dhruvachak:
> You have several place like this:
> 
> ```
>    if (LIS.hasInterval(Old)) {
>       LiveInterval &OldLI = LIS.getInterval(Old);
>       if (!OldLI.empty())
>         Matrix->assign(OldLI, PhysReg);
>       else
>         VRM.assignVirt2Phys(Old, PhysReg);
>     }
> ```
> 
> Claude, if you're reading this, please refactor :)

Done.

llvm/lib/CodeGen/InlineSpiller.cpp
dhruvachak: There is a shape-map manipulation at the end of this method. 
```  
if (VRM.hasShape(Old))
    VRM.assignVirt2Shape(New, VRM.getShape(Old));
```
So, an early exit does not appear straight-forward. It may be possible to hoist this shape manipulation all the way to the top but that would be unrelated to this PR. 

I did the refactor based on suggestion from @vpykhtin. The changes should be more readable now. Does that help?

dhruvachak:
> You mentioned that a few tests are exercising this change. What should be observed? There's no updated test in this PR.

With the validation in https://github.com/llvm/llvm-project/pull/197778, some LIT tests fail. This patch fixes the assertions in the following 3 LIT tests. 
CodeGen/X86/AMX/amx-gemm.ll
CodeGen/X86/apx/push2-pop2.ll
CodeGen/X86/udivmodei5.ll

Since these tests are already present, I did not look into adding a new test.

Here's the assertion failure (with the additional debug information). 

```
ERROR: VirtReg %226 assigned to $rdi in VirtRegMap but not found in LiveRegMatrix
  LiveInterval: %226 [4300r,4464r:0) 0@4300r  weight:0.000000e+00
  FirstStart: 4300r
  RegUnit DIL segments: [4300r,4688B) -> %207
  RegUnit DIH segments: [4300r,4688B) -> %207
  RegUnit HDI segments: [4300r,4688B) -> %207
  llc: RegAllocBase.cpp:161: virtual void llvm::RegAllocBase::postOptimization(): Assertion `Matrix->isValid() && "LiveRegMatrix validation failed"' failed.
```

vpykhtin:
I looked at the refactoring and think the `PendingReassignments` map can be simplified to a `DenseSet`. The key insight is that `Matrix->unassign(..., ClearAllReferencingSegments=true)` is safe to call after `shrinkToUses` — it matches entries in the `LiveIntervalUnion` by register number rather than segment pointer, so it correctly removes stale entries even after the old segments are gone. This means `LRE_WillShrinkVirtReg` doesn't need to pre-emptively unassign; it just records the vreg. Since VRM stays valid throughout, there's no physreg to cache — the map becomes a set.

This also eliminates the pending/non-pending split in `LRE_DidCloneVirtReg`: without pre-unassign, both cases look identical and collapse into one branch. Also, `Matrix->assign()` already calls `VRM->assignVirt2Phys` unconditionally regardless of whether the interval is empty, so the `!LI.empty()` fallback in `AssignToPhys` is unnecessary and the lambda can be dropped.

Suggested diff - top commit on the [branch](https://github.com/llvm/llvm-project/compare/main...vpykhtin:llvm-project:pr197773-simplify-delegates)

dhruvachak:
> I looked at the refactoring and think the `PendingReassignments` map can be simplified to a `DenseSet`. The key insight is that `Matrix->unassign(..., ClearAllReferencingSegments=true)` is safe to call after `shrinkToUses` — it matches entries in the `LiveIntervalUnion` by register number rather than segment pointer, so it correctly removes stale entries even after the old segments are gone. This means `LRE_WillShrinkVirtReg` doesn't need to pre-emptively unassign; it just records the vreg. Since VRM stays valid throughout, there's no physreg to cache — the map becomes a set.
> 
> This also eliminates the pending/non-pending split in `LRE_DidCloneVirtReg`: without pre-unassign, both cases look identical and collapse into one branch. Also, `Matrix->assign()` already calls `VRM->assignVirt2Phys` unconditionally regardless of whether the interval is empty, so the `!LI.empty()` fallback in `AssignToPhys` is unnecessary and the lambda can be dropped.
> 
> Suggested diff - top commit on the [branch](https://github.com/llvm/llvm-project/compare/main...vpykhtin:llvm-project:pr197773-simplify-delegates)

Thanks for the suggestion. However, I am concerned with the deferred unassign. That would mean that the Matrix would be inconsistent for a certain window. While there does not seem to be any client querying the Matrix in today's code, I am worried that someone may add it in the future, leading to incorrect results. Deferred unassign would also be inconsistent with the other delegates, RABasic and RAGreedy. LRE_WillShrinkVirtReg for these delegates unassigns eagerly.

I did some more refactoring while keeping the eager unassign, so I still need the pending map. I cleaned up some of the conditional code, so the resulting code should be much cleaner. Please take a look.

llvm/lib/CodeGen/InlineSpiller.cpp
```diff
@@ -1844,12 +1874,38 @@ bool HoistSpillHelper::LRE_CanEraseVirtReg(Register VirtReg) {
...
+    if (Matrix && LIS.hasInterval(VReg))
+      Matrix->assign(LIS.getInterval(VReg), PhysReg);
+    else
+      VRM.assignVirt2Phys(VReg, PhysReg);
+  };
+
+  auto PendingIt = PendingReassignments.find(Old);
+  if (PendingIt != PendingReassignments.end()) {
+    // Old was already unassigned in LRE_WillShrinkVirtReg.
+    assert(LIS.hasInterval(Old) &&
```
vpykhtin: assert(Matrix) too

llvm/lib/CodeGen/InlineSpiller.cpp
```diff
@@ -1844,12 +1874,38 @@ bool HoistSpillHelper::LRE_CanEraseVirtReg(Register VirtReg) {
...
+           "Pending reassignment for vreg without live interval");
+    MCRegister PhysReg = PendingIt->second;
+    PendingReassignments.erase(PendingIt);
+
+    // Reassign both Old (with its shrunk interval) and New to the matrix.
+    AssignToPhys(Old, PhysReg);
```
vpykhtin: should be enough due to assert
```suggestion
    Matrix->assign(LIS.getInterval(Old), PhysReg);
```

llvm/lib/CodeGen/InlineSpiller.cpp
```diff
@@ -1844,12 +1874,38 @@ bool HoistSpillHelper::LRE_CanEraseVirtReg(Register VirtReg) {
...
+    AssignToPhys(New, PhysReg);
+  } else if (VRM.hasPhys(Old)) {
+    MCRegister PhysReg = VRM.getPhys(Old);
+    if (Matrix && LIS.hasInterval(Old)) {
+      // Drop the stale pre-clone segments before reassigning Old's current LI.
+      Matrix->unassign(LIS.getInterval(Old),
+                       /*ClearAllReferencingSegments=*/true);
+      AssignToPhys(Old, PhysReg);
```
vpykhtin: should be enough due to condition
```suggestion
      Matrix->assign(LIS.getInterval(Old), PhysReg);
```
`LIS.getInterval(Old)` can be hoisted above unassign

llvm/lib/CodeGen/InlineSpiller.cpp
```diff
@@ -1844,12 +1874,38 @@ bool HoistSpillHelper::LRE_CanEraseVirtReg(Register VirtReg) {
...
+  if (PendingIt != PendingReassignments.end()) {
+    // Old was already unassigned in LRE_WillShrinkVirtReg.
+    assert(LIS.hasInterval(Old) &&
+           "Pending reassignment for vreg without live interval");
+    MCRegister PhysReg = PendingIt->second;
+    PendingReassignments.erase(PendingIt);
+
+    // Reassign both Old (with its shrunk interval) and New to the matrix.
+    AssignToPhys(Old, PhysReg);
+    AssignToPhys(New, PhysReg);
```
vpykhtin: Can we replace this with?
```suggestion
    Matrix->assign(LIS.getInterval(New), PhysReg);
```
Here Matrix is not null and I believe `LIS.hasInterval(New)` should be either true or we must create the interval with LIS.getInterval(New) because New was just created by splitting.

llvm/lib/CodeGen/InlineSpiller.cpp
```diff
@@ -1844,12 +1874,38 @@ bool HoistSpillHelper::LRE_CanEraseVirtReg(Register VirtReg) {
...
+  } else if (VRM.hasPhys(Old)) {
+    MCRegister PhysReg = VRM.getPhys(Old);
+    if (Matrix && LIS.hasInterval(Old)) {
+      // Drop the stale pre-clone segments before reassigning Old's current LI.
+      Matrix->unassign(LIS.getInterval(Old),
+                       /*ClearAllReferencingSegments=*/true);
+      AssignToPhys(Old, PhysReg);
+    }
+    AssignToPhys(New, PhysReg);
```
vpykhtin:
```suggestion
    if (Matrix)
      Matrix->assign(LIS.getInterval(New), PhysReg);
    else
      VRM.assignVirt2Phys(New, PhysReg);
```
The same reasoning as above

vpykhtin (commented):
Ok, thanks, this looks much cleaner now. I believe we can get rid of AssignToPhys lamda, please see the comments.

llvm/lib/CodeGen/InlineSpiller.cpp
vpykhtin: Probably add an assert:
```assert((!LIS.hasInterval(New) || !LIS.getInterval(New).empty()) && "Split product must have a non-empty live interval");```

**Comment by dhruvachak** [comment]
> Ok, thanks, this looks much cleaner now. I believe we can get rid of AssignToPhys lamda, please see the comments.

Thanks. I think I applied all the suggestions except the non-emptiness assertion on New-interval since that is not required by the patch.

File: llvm/lib/CodeGen/InlineSpiller.cpp (hunk 2 above)
Comment (dhruvachak) [reply llvm/lib/CodeGen/InlineSpiller.cpp]: Done.

File: llvm/lib/CodeGen/InlineSpiller.cpp (hunk 3 above)
Comment (dhruvachak) [reply llvm/lib/CodeGen/InlineSpiller.cpp]: Done.

File: llvm/lib/CodeGen/InlineSpiller.cpp (hunk 4 above)
Comment (dhruvachak) [reply llvm/lib/CodeGen/InlineSpiller.cpp]: Done.

File: llvm/lib/CodeGen/InlineSpiller.cpp (hunk 5 above)
Comment (dhruvachak) [reply llvm/lib/CodeGen/InlineSpiller.cpp]: Made the change, added a comment, except left alone the non-emptiness assert since that is not required by this patch.

File: llvm/lib/CodeGen/InlineSpiller.cpp (hunk 6 above)
Comment (dhruvachak) [reply llvm/lib/CodeGen/InlineSpiller.cpp]: Done.

File: llvm/lib/CodeGen/InlineSpiller.cpp [hunk 7]
```diff
@@ -1874,33 +1874,35 @@ bool HoistSpillHelper::LRE_CanEraseVirtReg(Register VirtReg) {
...
-    AssignToPhys(New, PhysReg);
+    Matrix->assign(LIS.getInterval(Old), PhysReg);
+    Matrix->assign(LIS.getInterval(New), PhysReg);
   } else if (VRM.hasPhys(Old)) {
     MCRegister PhysReg = VRM.getPhys(Old);
     if (Matrix && LIS.hasInterval(Old)) {
+      const LiveInterval &LI = LIS.getInterval(Old);
       // Drop the stale pre-clone segments before reassigning Old's current LI.
-      Matrix->unassign(LIS.getInterval(Old),
+      Matrix->unassign(LI,
```
vpykhtin: remove newline? though it passes formatting checks

vpykhtin (commented):
LGTM, thanks!

llvm/lib/CodeGen/InlineSpiller.cpp
```diff
@@ -1844,12 +1874,40 @@ bool HoistSpillHelper::LRE_CanEraseVirtReg(Register VirtReg) {
...
+  } else if (VRM.hasPhys(Old)) {
+    MCRegister PhysReg = VRM.getPhys(Old);
+    if (Matrix && LIS.hasInterval(Old)) {
+      const LiveInterval &LI = LIS.getInterval(Old);
+      // Drop the stale pre-clone segments before reassigning Old's current LI.
+      Matrix->unassign(LI,
+                       /*ClearAllReferencingSegments=*/true);
+      Matrix->assign(LI, PhysReg);
+    }
+    if (Matrix)
```
vpykhtin: Merge if with 1895 line? something like:
```
 if (Matrix) {
    // unassign/assign old
    // assign new
 } else 
    VRM.assignVirt2Phys(New, PhysReg);
```

llvm/lib/CodeGen/InlineSpiller.cpp
dhruvachak: Done.

llvm/lib/CodeGen/InlineSpiller.cpp
dhruvachak: Thanks, made the changes, looks cleaner now.

vpykhtin (approved):
Nice, LGTM!

PR #197776, opened 2026-05-14
[2/3][RegAlloc][LiveRegMatrix] Fix inconsistency when fold creates a CopyMI

dhruvachak:
When TargetInstrInfo::foldMemoryOperand produces a side-effect CopyMI, the copy destination vreg's LiveInterval is extended via addSegment. If that vreg is already assigned in LiveRegMatrix, the matrix becomes inconsistent because it still holds the old (smaller) interval.

Fix this in two places:

- InlineSpiller::foldMemoryOperand: unassign the vreg from the matrix before addSegment, reassign after, so the matrix reflects the updated interval.

- LiveRangeEdit::foldAsLoad: notify the delegate via LRE_WillShrinkVirtReg before addSegment.

LIT test exercised by this patch (i.e. if I introduce consistency verification as in https://github.com/llvm/llvm-project/pull/197778 but not this patch, the following test fails):
CodeGen/X86/apx/pr191368.ll

### DIFF
--- a/llvm/lib/CodeGen/InlineSpiller.cpp
+++ b/llvm/lib/CodeGen/InlineSpiller.cpp
@@ -1105,9 +1105,29 @@ foldMemoryOperand(ArrayRef<std::pair<MachineInstr *, unsigned>> Ops,
   if (CopyMI) {
     SlotIndex CopyIdx = LIS.InsertMachineInstrInMaps(*CopyMI).getRegSlot();
     if (!MRI.isSSA()) {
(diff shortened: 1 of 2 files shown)


qcolombet (approved):

dhruvachak:
I changed the base of this patch to main. There are no conflicts with https://github.com/llvm/llvm-project/pull/197773.

dhruvachak:
Rebased.

PR #197777, opened 2026-05-14
[llvm] Re-format aarch64-apple-tuning-features.td. NFC

jroelofs:
It's much easier to review diffs with each feature on its own line. Also add an -implicit-check-not so we don't miss any CPUs going forward.


tomershafir (approved):

llvm/test/TableGen/aarch64-apple-tuning-features.td
```diff
@@ -1,28 +1,279 @@
-// RUN: llvm-tblgen -print-records %p/../../lib/Target/AArch64/AArch64.td -I %p/../../lib/Target/AArch64 -I %p/../../include | FileCheck %s
+// RUN: llvm-tblgen -print-records %p/../../lib/Target/AArch64/AArch64.td -I %p/../../lib/Target/AArch64 -I %p/../../include \
+// RUN:    | sed -e "s/, /, \n/g" -e "s/\[/\[\n/g" \
```
madanial0, line 2: The sed command fails on AIX: https://lab.llvm.org/staging/#/builders/231/builds/750/steps/6/logs/FAIL__LLVM__aarch64-apple-tuning-features_td, would something like: `awk '{gsub(/\[/,"[\n"); gsub(/, /,",\n"); print}'` work?

tomershafir: https://github.com/llvm/llvm-project/pull/200656

PR #197778, opened 2026-05-14
[3/3][RegAlloc][LiveRegMatrix] Added validity check between VRM and LRM.

dhruvachak:
Check that every VRM-assigned vreg with a non-empty live interval must have its segments in the Matrix for its assigned physical reg.

Depends on https://github.com/llvm/llvm-project/pull/197773 and https://github.com/llvm/llvm-project/pull/197776. While there are no source conflicts, adding this verification without the other 2 patches would make the following LIT tests fail. 
CodeGen/X86/AMX/amx-gemm.ll
CodeGen/X86/apx/push2-pop2.ll
CodeGen/X86/udivmodei5.ll
CodeGen/X86/apx/pr191368.ll

### DIFF
--- a/llvm/lib/CodeGen/LiveRegMatrix.cpp
+++ b/llvm/lib/CodeGen/LiveRegMatrix.cpp
@@ -324,7 +324,51 @@ bool LiveRegMatrix::isValid() const {
       }
     }
   }
-  return NumDanglingPointers == 0;
+
+  // Reverse check: every VRM-assigned vreg with a non-empty live interval
+  // must have its segments present in the Matrix for its assigned phys reg.
+  unsigned NumMissing = 0;
+  for (unsigned RegIdx = 0, NumRegs = VRM->getRegInfo().getNumVirtRegs();
+       RegIdx < NumRegs; ++RegIdx) {
+    Register VReg = Register::index2VirtReg(RegIdx);
+    if (!VRM->hasPhys(VReg) || !LIS->hasInterval(VReg))
+      continue;
+    const LiveInterval &LI = LIS->getInterval(VReg);
+    if (LI.empty())
+      continue;
+    MCRegister PhysReg = VRM->getPhys(VReg);
+    // Check that the first segment of LI is present in the LiveUnion for
+    // at least one reg unit of PhysReg.
+    SlotIndex FirstStart = LI.beginIndex();
+    bool Found = false;
+    for (MCRegUnit Unit : TRI->regunits(PhysReg)) {
+      auto It = Matrix[Unit].find(FirstStart);
+      if (It.valid() && It.start() == FirstStart && It.value() == &LI) {
+        Found = true;
+        break;
+      }
+    }
+    if (!Found) {
+      ++NumMissing;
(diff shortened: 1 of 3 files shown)


dhruvachak:
Requesting advice from reviewers regarding the best way to review these 3 patches in this stack. 

While there are no conflicts between the patches, the verification will fail if the other 2 patches are not present, that's why I kept a dependency between the PRs. Since some of the inconsistency fixes are distinct, I kept them separate in the first 2 patches. This should make reviewing easier as well. 

I would like to keep the verification patch separate in case it needs to be reverted for issues that are not yet fixed. 

I ran into these inconsistencies while debugging a correctness problem where because of stale LRM, a subsequent transformation was clobbering live data.

llvm/lib/CodeGen/LiveRegMatrix.cpp
```diff
@@ -324,7 +324,51 @@ bool LiveRegMatrix::isValid() const {
...
+    for (MCRegUnit Unit : TRI->regunits(PhysReg)) {
+      auto It = Matrix[Unit].find(FirstStart);
+      if (It.valid() && It.start() == FirstStart && It.value() == &LI) {
+        Found = true;
+        break;
+      }
+    }
+    if (!Found) {
+      ++NumMissing;
+      dbgs() << "ERROR: VirtReg " << printReg(VReg, TRI) << " assigned to "
```
qcolombet, line 353: Was the intent to have improved debug log here?
I.e., are we missing a LLVM_DEBUG?

qcolombet (changes requested):

llvm/lib/CodeGen/LiveRegMatrix.cpp
dhruvachak: Note that this method `LiveRegMatrix::isValid()` is guarded by NDEBUG, so it does not matter anyways for release builds. For non-release builds, the intent is to have some information emitted when the verification fails, even when the -debug flag is not used. That's why I did not use LLVM_DEBUG. 

It is also in keeping with existing code in this method which uses the same dbgs() pattern when verification fails.

vpykhtin: right, this method was assumed to be called like `assert(isValid());`

vpykhtin (approved):
LGTM

dhruvachak:
Updated branch.

qcolombet (approved):
I see that we already print stuff in this method without using the proper LLVM_DEBUG scope.

saipoorna:
@dhruvachak : Are we good to merge this change?

dhruvachak:
> @dhruvachak : Are we good to merge this change?

I will rebase and do some more testing.

dhruvachak:
Notes for this PR:
- The change adds more consistency validation.
- The change kicks in only when NDEBUG is not defined. So pure release builds won't have this change. This change will kick in if assertions are enabled.

If a new failure is seen after this change, it is likely a new inconsistency has been found. It is unlikely this patch itself is the problem for such a failure. See for example https://github.com/llvm/llvm-project/pull/197773 and https://github.com/llvm/llvm-project/pull/197776. 

I built all targets with this change and ran check-llvm successfully.

PR #197779, opened 2026-05-14
[MLIR][GPUToXeVMPipeline] Add unsupported data type emulation on `math`, `arith`, and `vector` ops.

mshahneo:
Add `math-extend-to-supported-types` and `arith-emulate-unsupported-floats` pass to the pipeline. These passes are used to emulate `math`, `arith` and `vector` floating point operations that use float types which are unspported on a target by inserting `extf/truncf` pairs around all such operations.

### DIFF
--- a/mlir/include/mlir/Dialect/GPU/Pipelines/Passes.h
+++ b/mlir/include/mlir/Dialect/GPU/Pipelines/Passes.h
@@ -161,6 +161,32 @@ struct GPUToXeVMPipelineOptions
       *this, "igc-cmd-options",
       llvm::cl::desc("Additional downstream compiler command line options"),
       llvm::cl::init("")};
+  // Options for MathExtendToSupportedTypes and ArithEmulateUnsupportedFloats,
+  // run on the host/device module before conversion to LLVM to legalize math
+  // and arith ops operating on floating-point types that the XeVM target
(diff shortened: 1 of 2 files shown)


silee2 (approved):

LewisCrawford:
I'm encountering downstream build failures after this patch with:
`ld.lld: error: undefined symbol: mlir::math::createMathExtendToSupportedTypes(mlir::math::MathExtendToSupportedTypesOptions)`

I think the fix would be to add `MLIRMathTransforms` to the `LINK_LIBS PUBLIC` section in `mlir/lib/Dialect/GPU/Pipelines/CMakeLists.txt`.

mshahneo:
> I'm encountering downstream build failures after this patch with: `ld.lld: error: undefined symbol: mlir::math::createMathExtendToSupportedTypes(mlir::math::MathExtendToSupportedTypesOptions)`
> 
> I think the fix would be to add `MLIRMathTransforms` to the `LINK_LIBS PUBLIC` section in `mlir/lib/Dialect/GPU/Pipelines/CMakeLists.txt`.

Hi @LewisCrawford ,
Opened a PR:https://github.com/llvm/llvm-project/pull/199060 to fix the issue.
Sorry for the inconveniences.

PR #197780, opened 2026-05-14
[CIR] Fix problem with scope termination

andykaylor:
This fixes a problem with scope termination when a cleanup scope appears as the last operation within a region that is nested inside another operation, such as within the body region of a cir.for operaiton. When the cleanup is popped, we set the insertion point to just past the cleanup operation, which is what we need if there are going to be more operations emitted, but the LexicalScope::cleanup function, which intends to add a terminator to its ScopeOp region, was not working correctly because it expected the insert location to be at the end of the scope.

In the problematic case, we had something like this:

cir.scope {
  ...
  cir.for : cond {
    ...
  } body {
    ...
    cir.cleanup.scope {
      ...
    } cleanup normal {
      ...
    }
    < insert location was set here >
  }
  < no terminator here>
}

We have another function (terminateStructuredRegionBody) that fills in the terminator in the cir.for body, but we need LexicalScope::cleanup to insert the terminator for the cir.scope.

This change updates the insert location in LexicalScope::cleanup, if necessary, to the end of the scope region.

### DIFF
--- a/clang/test/CIR/CodeGen/forrange.cpp
+++ b/clang/test/CIR/CodeGen/forrange.cpp
@@ -131,3 +131,39 @@ void for_range3() {
 // CIR:        cir.yield
 // CIR:      }
 // CIR:    }
+
+struct HasDtor { ~HasDtor(); };
+
+void for_range4() {
+  C3 c;
+  for (Element &e : c) {
+    HasDtor hd;
+  }
+}
+
+// CIR: cir.func{{.*}} @_Z10for_range4v()
+// CIR:    cir.scope {
+// CIR:      %[[RANGE_ADDR:.*]] = cir.alloca !cir.ptr<!rec_C3>{{.*}} ["__range1", init, const]
+// CIR:      %[[BEGIN_ADDR:.*]] = cir.alloca !rec_Iterator, !cir.ptr<!rec_Iterator>{{.*}} ["__begin1", init]
+// CIR:      %[[END_ADDR:.*]] = cir.alloca !rec_Iterator, !cir.ptr<!rec_Iterator>{{.*}} ["__end1", init]
+// CIR:      %[[E_ADDR:.*]] = cir.alloca !cir.ptr<!rec_Element>{{.*}} ["e", init, const]
+// CIR:      %[[HD:.*]] = cir.alloca !rec_HasDtor, !cir.ptr<!rec_HasDtor>, ["hd"]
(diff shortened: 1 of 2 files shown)


AmrDeveloper (approved):
LGTM

clang/test/CIR/CodeGen/forrange.cpp
```diff
@@ -131,3 +131,39 @@ void for_range3() {
 // CIR:        cir.yield
 // CIR:      }
 // CIR:    }
+
+struct HasDtor { ~HasDtor(); };
+
+void for_range4() {
```
adams381, line 137: Reduced this independently to a switch default body with `Local x;` + `throw 42;` and landed on the same root cause — `popCleanupBlock`'s `setInsertionPointAfter(scope)` leaving the builder one region below the `LexicalScope`'s own region.  My first spike was an unconditional `InsertionGuard` around `forceCleanup` and it regressed six existing tests; the region-identity check here is the right shape because same-region paths stay unchanged.

One suggestion: a switch-case test alongside `for_range4`.  The fix is general (it checks region identity, not op kind), so it covers both surfaces, but the switch body reaches `cleanup()` through a different emit path (`emitCaseDefaultCascade`) and a separate test would guard against drift there.  Minimal repro:

```cpp
struct Local { ~Local(); };
void test(int n) {
  switch (n) {
  case 1: break;
  default:
    Local x;
    throw 42;
  }
}
```

erichkeane (approved):
Like the idea of adding Adam's test too, but otherwise LGTM.

clang/test/CIR/CodeGen/forrange.cpp
andykaylor: I also tried the `InsertionGuard` before I ended up with this change. Cursor suggested something different that I didn't like, and I thought, can't we just put an `InsertionGuard` here...? Third time was the charm, I hope.

I'll add your test case. I'm also looking more broadly at cleaning up our general insertion point handling with cleanups because it feels like we're chasing failures here. This is at least the third "fix" I've made for this general class of bug. I feel good about this fix until we can find something better though.

andykaylor: It turns out that this exact test case isn't working because we seem not to handle cir.throw correctly during flattening. I'll fix that separately. I'm adding a test case that calls a function which "may throw" so it exercises the same basic path for terminator purposes but without hitting the throw bug.

PR #197781, opened 2026-05-14
[SampleProfileMatcher] Skip orphan function stale profile matching if previously matched

HighW4y2H3ll:
Stale profile matching may match same profiles to multiple IR functions during the stale profile matching, and creating conflicting `IRToProfileLocationMap`, and hitting the following assertation failure:
```
clang++: /home/llvm-project/llvm/lib/Transforms/IPO/SampleProfileMatcher.cpp:333: void llvm::SampleProfileMatcher::runStaleProfileMatching(const Function &, const AnchorMap &, const AnchorMap &, LocToLocMap &, bool, bool): Assertion `IRToProfileLocationMap.empty() && "Run stale profile matching only once per function"' failed.
```
This patch moves the pre-match to a later stage and runs a second stale profile matching only when the orphan functions does not have a matched profile.

### DIFF
--- a/llvm/include/llvm/Transforms/IPO/SampleProfileMatcher.h
+++ b/llvm/include/llvm/Transforms/IPO/SampleProfileMatcher.h
@@ -79,6 +79,10 @@ class SampleProfileMatcher {
   // name.
   MapVector<Function *, FunctionId> FuncToProfileNameMap;
 
+  // Separate mapping for Orphan toplevel functions (without profile) to
+  // candidate function profiles when their basenames matched.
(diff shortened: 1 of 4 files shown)


grigorypas:
Do I understand correctly that stale profile matching will not run on orphaned functions? I think we should re-design data structure instead. Right now we map profile probe to IR probe. I wonder if we can reverse that and map IR probe to profile prob allowing many to one mapping.

HighW4y2H3ll:
@grigorypas Thanks for taking a look. This is a wrong fix. I'll close this one.

PR #197784, opened 2026-05-14
[CIR][AMDGPU] Add support for AMDGCN sin builtins

ayokunle321:
Adds codegen for the following AMDGCN sin builtins:

- __builtin_amdgcn_sinf (float)
- __builtin_amdgcn_sinh (half)
- __builtin_amdgcn_sin_bf16 (bfloat16)

These are lowered to the corresponding `llvm.amdgcn.sin` intrinsic.


clang/test/CIR/CodeGenHIP/builtins-amdgcn-gfx1250.hip
```diff
@@ -0,0 +1,26 @@
+#include "../CodeGenCUDA/Inputs/cuda.h"
```
arsenm: Usually we just do `#define __device__ __attribute__((device))` or use the raw attribute instead of including a header

arsenm (commented):
Test name and convention seems to not match the original codegen tests? I'd assume the CIR tests would be copy paste of the originals

ayokunle321:
> Test name and convention seems to not match the original codegen tests? I'd assume the CIR tests would be copy paste of the originals

@arsenm ah I see. I was following the OpenCL tests (e.g. builtins-amdgcn-vi.cl) like was done in the incubator. Found the right ones now, will match the format used in the HIP tests instead (e.g. builtins-amdgcn-vi-f16.hip).

ranapratap55 (approved):
LGTM.

Please resolve merge conflicts before merge.

PR #197785, opened 2026-05-14
[clang][deps] Simplify VFS overlays

jansvoboda11:
Instead of operating on on-disk files, the scanner can be made to operate on in-memory buffers and module names. This is facilitated by changes to the command line and the VFS, where an imaginary file is injected (mainly to make the driver happy). Currently, this is implemented by functions external to the worker that take its base VFS, wrap it with an overlay VFS, and pass it back to the worker. Since the worker _needs_ to operate on top of the base VFS, it performs a sanity check like so:

```c++
#ifndef NDEBUG
    bool SawDepFS = false;
    OverlayFS->visit(
        [&](llvm::vfs::FileSystem &VFS) { SawDepFS |= &VFS == DepFS.get(); });
    assert(SawDepFS && "OverlayFS not based on DepFS");
#endif
```

This PR simplifies the situation by passing the custom VFS to the new `DependencyScanningWorker::makeEffectiveVFS()`. This is used by external tools that need to operate in the same context as the worker, but it's also called by the worker itself, guaranteeing that it operates on the expected VFS, which in turn allows removing the complicated visitation and assert pattern.

### DIFF
--- a/clang/include/clang/DependencyScanning/DependencyScanningWorker.h
+++ b/clang/include/clang/DependencyScanning/DependencyScanningWorker.h
@@ -78,7 +78,7 @@ public:
 
   /// Run the dependency scanning tool for all given frontend command-lines,
   /// and report the discovered dependencies to the provided consumer. The
-  /// \C OverlayFS will be used to call \c makeEffectiveVFS().
+  /// \c OverlayFS will be used to call \c makeEffectiveVFS().
   ///
   /// \returns false if any errors occurred (with diagnostics reported to
   /// \c DiagConsumer), true otherwise.
--- a/clang/include/clang/DependencyScanning/DependencyScanningWorker.h
+++ b/clang/include/clang/DependencyScanning/DependencyScanningWorker.h
@@ -77,12 +77,8 @@ public:
   ~DependencyScanningWorker();
 
   /// Run the dependency scanning tool for all given frontend command-lines,
-  /// and report the discovered dependencies to the provided consumer.
-  ///
-  /// OverlayFS should be based on the Worker's dependency scanning file-system
-  /// and can be used to provide any input specified on the command-line as
-  /// in-memory file. If no overlay file-system is provided, the Worker's
-  /// dependency scanning file-system is used instead.
+  /// and report the discovered dependencies to the provided consumer. The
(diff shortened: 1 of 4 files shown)


clang/lib/Tooling/DependencyScanningTool.cpp
```diff
@@ -269,61 +263,41 @@ std::optional<P1689Rule> DependencyScanningTool::getP1689ModuleDependencyFile(
...
+  // threads is ok.
   static const std::string FakeInput(
-      clang::tooling::CompilerInstanceWithContext::MaxNumOfQueries, ' ');
-  InMemoryFS->addFile(FakeInputPath, 0,
-                      llvm::MemoryBuffer::getMemBuffer(FakeInput));
-  IntrusiveRefCntPtr<llvm::vfs::FileSystem> InMemoryOverlay = InMemoryFS;
-  OverlayFS->pushOverlay(InMemoryOverlay);
+      CompilerInstanceWithContext::MaxNumOfQueries, ' ');
+
+  StringRef InputPath = "/module-include.input";
```
jansvoboda11: This was refactored to be an absolute path, so that we don't have to rely on the CWD.

jansvoboda11:
ping

clang/include/clang/DependencyScanning/DependencyScanningWorker.h
```diff
@@ -52,22 +77,26 @@ class DependencyScanningWorker {
...
 
   /// Run the dependency scanning tool for all given frontend command-lines,
-  /// and report the discovered dependencies to the provided consumer.
-  ///
-  /// OverlayFS should be based on the Worker's dependency scanning file-system
-  /// and can be used to provide any input specified on the command-line as
-  /// in-memory file. If no overlay file-system is provided, the Worker's
-  /// dependency scanning file-system is used instead.
+  /// and report the discovered dependencies to the provided consumer. The
+  /// \C OverlayFS will be used to call \c makeEffectiveVFS().
```
naveen-seth: I think this should be a small \c.
(Or does it not make a difference for Doxygen? I'm not too sure.)

naveen-seth (approved):
LGTM

PR #197788, opened 2026-05-14
[AMDGPU] Restore lit() to be accepted as 64-bit FP operand

rampitec:
It is worth to note that immediate value comes already truncated
to this point of validation.


rampitec:
* #197803
* #197788 👈
* `main`

<h2></h2>

rampitec:
At the end it seems to materialize a right constant in the SP. Note, this is inconsistent with SP3 and always was. The problem with SP3 is it does not have instruction context when parsing operands. It first parses operands and then instruction. So `lit(1.0)` will always create `0x3f800000` and `lit64(1.0)` will be `0x3ff0000000000000`. So basically with SP3 it is impossible to create a proper DP 1.0 in a literal form if lit64 is not available for an operand. For us it is possible because we first match the instruction and know the operand type when we parse and materialize the operand exact value.

llvm/test/MC/AMDGPU/literals.s
```diff
@@ -1971,15 +1971,106 @@ v_add_nc_u64 v[0:1], v[0:1], lit64(1)
...
+// NOSICI: :[[@LINE-4]]:31: error: literal operands are not supported
 
 v_add_f64 v[0:1], v[0:1], lit64(1.0)
 // NOGFX11: :[[@LINE-1]]:27: error: lit64 is not supported on this GPU
 // NOGFX12: :[[@LINE-2]]:27: error: lit64 is not supported on this GPU
 // NOGFX1250: :[[@LINE-3]]:33: error: invalid operand for instruction
 // NOGFX89: :[[@LINE-4]]:27: error: lit64 is not supported on this GPU
 // NOSICI: :[[@LINE-5]]:27: error: lit64 is not supported on this GPU
+
+v_add_f64 v[0:1], v[0:1], lit(0x3ff00000)
```
arsenm, line 2043: Do we support producing these 32-bit truncated immediate in operand folding?

arsenm (approved):

llvm/test/MC/AMDGPU/literals.s
rampitec: Yes, it was always there. Although we do not use forced lit() in codegen, it will be just inline constant.

PR #197791, opened 2026-05-14
[clang][driver][darwin] Hold onto full triples in Darwin SDKPlatformInfo

ian-twilightcoder:
The architecture can be relevant when determining if an SDK supports a particular triple. Record the full triples in SDKPlatformInfo instead of all of the non-architecture triple components.

Assisted-by: Claude Code

rdar://172876443

### DIFF
--- a/clang/include/clang/Basic/DarwinSDKInfo.h
+++ b/clang/include/clang/Basic/DarwinSDKInfo.h
@@ -36,32 +36,16 @@ public:
   /// definitions, in the SDK.
   struct SDKPlatformInfo {
   public:
-    SDKPlatformInfo(llvm::Triple::VendorType Vendor, llvm::Triple::OSType OS,
-                    llvm::Triple::EnvironmentType Environment,
-                    llvm::Triple::ObjectFormatType ObjectFormat,
-                    StringRef PlatformPrefix)
-        : Vendor(Vendor), OS(OS), Environment(Environment),
-          ObjectFormat(ObjectFormat), PlatformPrefix(PlatformPrefix) {}
+    using TripleStorageType = SmallVector<llvm::Triple, 5>;
 
-    llvm::Triple::VendorType getVendor() const { return Vendor; }
-    llvm::Triple::OSType getOS() const { return OS; }
-    llvm::Triple::EnvironmentType getEnvironment() const { return Environment; }
-    llvm::Triple::ObjectFormatType getObjectFormat() const {
-      return ObjectFormat;
-    }
+    SDKPlatformInfo(TripleStorageType Triples, StringRef PlatformPrefix)
+        : Triples(std::move(Triples)), PlatformPrefix(PlatformPrefix) {}
+
+    const TripleStorageType &getTriples() const { return Triples; }
     StringRef getPlatformPrefix() const { return PlatformPrefix; }
 
-    bool operator==(const llvm::Triple &RHS) const {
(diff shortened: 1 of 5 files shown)


ian-twilightcoder:
@cachemeifyoucan you've seen these changes before, this is the upstream version

clang/include/clang/Basic/DarwinSDKInfo.h
```diff
@@ -190,12 +174,22 @@ class DarwinSDKInfo {
           VersionMappings =
               llvm::DenseMap<OSEnvPair::StorageType,
                              std::optional<RelatedTargetVersionMapping>>())
-      : FilePath(FilePath), OS(OS), Environment(Environment), Version(Version),
-        DisplayName(DisplayName),
+      : FilePath(std::move(FilePath)), OS(OS), Environment(Environment),
```
ian-twilightcoder, line 177: I wrote all of the code myself by hand, then I had Claude review these changes for memory management issues. It found a handful of unnecessary copies in existing code like this one here.

ian-twilightcoder: I don't know if that warrants a "Assisted-by: Claude Code", but I put it there just to be safe since I did technically use AI while writing this code.

clang/unittests/Basic/DarwinSDKInfoTest.cpp
```diff
@@ -107,8 +107,8 @@ TEST(DarwinSDKInfo, PlatformPrefix) {
   ASSERT_TRUE(SDKInfo);
   EXPECT_EQ(SDKInfo->getPlatformPrefix(Triple("arm64-apple-macos26.0")),
             "/System/macOSSupport");
-  // The triple's architecture doesn't matter.
-  EXPECT_EQ(SDKInfo->getPlatformPrefix(Triple("ppc-apple-macos26.0")),
+  // The triple's architecture matters.
+  EXPECT_NE(SDKInfo->getPlatformPrefix(Triple("ppc-apple-macos26.0")),
```
cachemeifyoucan, line 111: I think a better check here is a `EXPECT_EQ` that is not the same as "/System/macOSSupport"

cachemeifyoucan (approved):
LGTM in general. Small nit

clang/unittests/Basic/DarwinSDKInfoTest.cpp
ian-twilightcoder: Like you want `EXPECT_TRUE` that `getPlatformPrefix(...).empty()`?

cachemeifyoucan: Or create a different entry that actually returns a different value.

ian-twilightcoder: I'm trying to test that an arch that _isn't_ covered in SDKSettings doesn't pick up the supported target that matches everything but arch though. I think empty would cover that, but since I'm going for "doesn't match", I think the `_NE` check is more what I'm trying to test.

cyndyishida:
This broke a downstream build for compiler-rt builtins. We're going to revert this for the time being until you return to take a closer look.

ian-twilightcoder:
> This broke a downstream build for compiler-rt builtins. We're going to revert this for the time being until you return to take a closer look.

Sorry I had a downstream change ready to go, I just missed the merges.

PR #197793, opened 2026-05-14
[WIP][RISCV] Simple scalar arithmetic instructions have a reciprocal throughput of 1/2

bababuck:
Based off of SpacemiT K3 data: https://camel-cdr.github.io/rvv-bench-results/spacemit_a100/index.html

Tests have not been updated since will be a lot of work to adjust tests to ensure they are still testing what they were intended to, and want to avoid doing that until the actual costing change is finalized.

I am currently gathering data on a `BananaPi` board, will update here once those results are ready.

PR #197794, opened 2026-05-14
[CostModel] Directly output the InstructionCost rather than using getValue()

bababuck:
`getValue()` doesn't preserve fractional costs.


bababuck:
Don't have a test to demonstrate this change since no backends utilize fractional costs yet, but #197793 includes this commit and tests it.

lukel97 (approved):
LGTM

PR #197795, opened 2026-05-14
[AMDGPU] Use pessimistic fixpoint in AAAMDGPUMinAGPRAlloc for sanitized functions

reble:
Adding pessimistic fixpoint when a function has any sanitizer attribute (`sanitize_address`, `sanitize_memory`, `sanitize_thread`, `sanitize_hwaddress`, `sanitize_mem_tag`). This prevents the Attributor from inferring "amdgpu-agpr-alloc"="0" for sanitized functions or their callers, letting the register allocator conservatively reserve AGPR capacity.

This patch should prevent kernel to exceeds the hardware register budget on gfx90a where VPGR and AGPR share the same physical register file.


reble:
@bcahoon

llvm/test/CodeGen/AMDGPU/amdgpu-attributor-min-agpr-alloc-sanitizer.ll
```diff
@@ -0,0 +1,70 @@
+; RUN: opt -S -mtriple=amdgcn-unknown-amdhsa -mcpu=gfx90a -passes=amdgpu-attributor %s | FileCheck %s
```
bcahoon: Can you use update_test_checks.py to generate the CHECKs? It won't generate the NOT checks, but that's ok, since the positive checks will catch the error. I think you need to add --check-attributes --check-globals all as options to update_test_checks.py

llvm/test/CodeGen/AMDGPU/amdgpu-attributor-min-agpr-alloc-sanitizer.ll
```diff
@@ -0,0 +1,70 @@
...
+  ret void
+}
+
+; CHECK-LABEL: define amdgpu_kernel void @kernel_calls_sanitized_callee(
+; CHECK-SAME: ) #[[CALLER:[0-9]+]] {
+define amdgpu_kernel void @kernel_calls_sanitized_callee() {
+  call void @sanitized_callee()
+  ret void
+}
+
```
arsenm: Can you check pre-existing amdgpu-agpr-alloc attribute + sanitizer? It should be preserved

llvm/lib/Target/AMDGPU/AMDGPUAttributor.cpp
```diff
@@ -1316,6 +1316,9 @@ struct AAAMDGPUMinAGPRAlloc
                                         /*OnlyFirstRequired=*/true);
     if (MinNumAGPR == 0)
```
arsenm: Should this early exit?

llvm/test/CodeGen/AMDGPU/amdgpu-attributor-min-agpr-alloc-sanitizer.ll
```diff
@@ -0,0 +1,99 @@
...
+; CHECK-LABEL: define amdgpu_kernel void @kernel_calls_sanitized_callee(
+; CHECK-SAME: ) #[[ATTR6:[0-9]+]] {
+; CHECK-NEXT:    call void @sanitized_callee()
+; CHECK-NEXT:    ret void
+;
+  call void @sanitized_callee()
+  ret void
+}
+
+attributes #0 = { sanitize_address "amdgpu-agpr-alloc"="4" }
```
arsenm: Using 4 doesn't stress the == 0 early exit condition

llvm/lib/Target/AMDGPU/AMDGPUAttributor.cpp
```diff
@@ -1316,6 +1316,9 @@ struct AAAMDGPUMinAGPRAlloc
...
       indicateOptimisticFixpoint();
```
arsenm, line 1318: Early exit here?

llvm/test/CodeGen/AMDGPU/amdgpu-attributor-min-agpr-alloc-sanitizer.ll
reble: done

llvm/lib/Target/AMDGPU/AMDGPUAttributor.cpp
reble: Yes makes sense.

llvm/test/CodeGen/AMDGPU/amdgpu-attributor-min-agpr-alloc-sanitizer.ll
reble: Good catch, should we keep a test variant for non-zero value, or do you think that is excessive?

llvm/test/CodeGen/AMDGPU/amdgpu-attributor-min-agpr-alloc-sanitizer.ll
```diff
@@ -0,0 +1,99 @@
+; NOTE: Assertions have been autogenerated by utils/update_test_checks.py UTC_ARGS: --check-attributes --check-globals all --version 6
+; RUN: opt -S -mtriple=amdgcn-unknown-amdhsa -mcpu=gfx90a -passes=amdgpu-attributor %s | FileCheck %s
+
```
arsenm: Can you merge this into amdgpu-attributor-min-agpr-alloc.ll (or at least, copy use_most and call it in the functions, such so that it removes all of the unnecessary attributes from the result)

arsenm (approved):

arsenm (approved):

llvm/test/CodeGen/AMDGPU/amdgpu-attributor-min-agpr-alloc-sanitizer.ll
reble: done

llvm/test/CodeGen/AMDGPU/amdgpu-attributor-min-agpr-alloc-sanitizer.ll
reble: done

llvm/lib/Target/AMDGPU/AMDGPUAttributor.cpp
reble: yes, done and test added.

PR #197796, opened 2026-05-14
[lldb] Assert that CommandObject::DoExecute sets a return status

JDevlieghere:
Change the default value of CommandReturnObject::m_status from eReturnStatusStarted to eReturnStatusInvalid, and add a debug-only RAII check in CommandObjectParsed::Execute and CommandObjectRaw::Execute that asserts the status is no longer Invalid after DoExecute returns.

This catches commands that forget to call SetStatus on a success or failure path. Succeeded() still returns true when the status is Invalid (0 sorts below eReturnStatusSuccessContinuingResult), so helpers that read result.Succeeded() as a precondition before any explicit SetStatus (e.g. StopProcessIfNecessary) continue to work.

This is a re-land of #196589.

rdar://176506732

### DIFF
--- a/lldb/include/lldb/Interpreter/CommandReturnObject.h
+++ b/lldb/include/lldb/Interpreter/CommandReturnObject.h
@@ -189,7 +189,10 @@ private:
   std::vector<DiagnosticDetail> m_diagnostics;
   std::optional<uint16_t> m_diagnostic_indent;
 
-  lldb::ReturnStatus m_status = lldb::eReturnStatusStarted;
+  /// The command's return status indicating success or failure. The default
+  /// value indicate no status has been set, which is enforced by an assert in
+  /// the CommandInterpreter.
+  lldb::ReturnStatus m_status = lldb::eReturnStatusInvalid;
 
   /// An optionally empty list of values produced by this command.
   ValueObjectList m_value_objects;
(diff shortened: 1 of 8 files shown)


JDevlieghere:
@slydiman I missed your last ping, but I'll definitely be keeping an eye on the remote bots this time.

slydiman (approved):
We have tested this patch:

Testing Time: 438.32s

Total Discovered Tests: 34018
  Skipped          :     2 (0.01%)
  Unsupported      :   971 (2.85%)
  Passed           : 33023 (97.08%)
  Expectedly Failed:    22 (0.06%)

thechenli:
@JDevlieghere hey Jonas - we (Meta) hit this assertion internally in the Intel PT tests after this change.

The failing commands are `thread trace dump instructions` and likely the sibling `thread trace dump function-calls`. Both paths can complete successfully without setting the `CommandReturnObject` status:
```
dumper.DumpFunctionCalls();
result.SetStatus(eReturnStatusSuccessFinishResult);

and:

m_last_id = dumper.DumpInstructions(m_options.m_count);
result.SetStatus(eReturnStatusSuccessFinishResult);
```

We have seen some failing tests within Meta on [this](https://github.com/llvm/llvm-project/blob/f348d964af441ec454e5fc12ab1b599cfc1e7d14/lldb/test/API/commands/trace/multiple-threads/TestTraceStartStopMultipleThreads.py#L28) and [this](https://github.com/llvm/llvm-project/blob/f348d964af441ec454e5fc12ab1b599cfc1e7d14/lldb/test/API/commands/trace/TestTraceTSC.py#L93). I suspect these may be skipped on GitHub builders when Intel PT is unavailable. Does adding the two success statuses above look like the intended fix? I can send a small follow-up PR if that makes sense.

cc: @clayborg @youngd007

PR #197799, opened 2026-05-14
[DAG] Combine vselect of vselect with identical conditions.

davemgreen:
I didn't look much at the test changes - some of them look like they will be optimized quite differently by the midend.

https://alive2.llvm.org/ce/z/ZVYVng


RKSimon (commented):
A couple of the x86 regression tests needs adjusting so they test what they're supposed to again - I'll take a proper look in the morning

RKSimon:
@davemgreen I think I've fixed the bad x86 tests - please can you merge/rebase?

davemgreen:
Thanks - I don't have a great need of this if it causes problems, it was just something I noticed.

Issue #196827, opened 2026-05-10
[X86] Missing use of 8-/16-bit `cmp` for comparing truncated values


dzaima:
Code like:

```c
#include<stdint.h>
void ext(void);
void gt_u8(long a, long b) {
    if ((uint8_t)a > (uint8_t)b) ext();
}
```

compiled with `-O3`, compiles to:

```asm
gt_u8:
        movzx   eax, dil
        movzx   ecx, sil
        cmp     eax, ecx
        ja      ext@PLT
        ret
```

but could compile to:

```asm
gt_u8:
        cmp     dil, sil
        ja      ext@PLT
        ret
```

This applies to 8-bit and 16-bit types, and both `>` and `>=`.

Also happens with one arg being an immediate (though for the 16-bit case that may be actually undesirable due to slow data16).

https://c.godbolt.org/z/9hKc6shfc

he-weiwen:
Hi, I'd like to work on this.

GkvJwa:
Related #195907

aengelke:
Maybe this should be done in instcombine? https://alive2.llvm.org/ce/z/wfVKcc

PR #197801, opened 2026-05-14
[X86] Shrink i32 unsigned compares to i8/i16 when high bits are zero

he-weiwen:
Shrink i32 compares to 8 or 16 bit for non-constant operands when high bits are zero, following the precedent of i64 -> i32 case. 

Fixes the cases in #196827 where both sides of the compare are non-constants, as this allows the `and(X, mask) `to be eliminated downstream in DAGCombine.

I believe we can extend to constant operands as long as the length changing prefix problem for i16 immediates is accounted for - however:
 1. it's not sufficient to fix the remaining cases in the motivating issue alone: when the rhs is constant, InstCombine clear the bits in the mask that does not affect the comparison result (`icmp (and X, 255), 123` -> `icmp (and X, 252), 123`), so the mask does not fill the full width of the truncated type, making the analysis trickier.
 2. it leads to a lot of check line drifts, i'm not sure if it's worth the effort to account for all of them, even if most don't look like regressions.


llvm/test/CodeGen/X86/cmp-shiftX-maskX.ll
```diff
@@ -106,9 +106,9 @@ define i1 @shl_to_shr_eq_i32_s3_fail(i32 %x) {
 define i1 @shl_to_shr_ne_i32_s16(i32 %x) {
 ; CHECK-NOBMI-LABEL: shl_to_shr_ne_i32_s16:
 ; CHECK-NOBMI:       # %bb.0:
-; CHECK-NOBMI-NEXT:    movzwl %di, %eax
+; CHECK-NOBMI-NEXT:    movl %edi, %eax
 ; CHECK-NOBMI-NEXT:    shrl $16, %edi
-; CHECK-NOBMI-NEXT:    cmpl %edi, %eax
+; CHECK-NOBMI-NEXT:    cmpw %di, %ax
```
efriedma-quic, line 111: Consider the encodings:

```
        cmpw    %di, %ax                        # encoding: [0x66,0x39,0xf8]
        cmpl    %edi, %eax                      # encoding: [0x39,0xf8]
        cmpq    %rdi, %rax                      # encoding: [0x48,0x39,0xf8]
```

Going from 64->32 saves a byte.  Going from 32->16 increases the size.

This is maybe still worth it in some cases, but you need a more complete heuristic.

phoebewang: It's neutral considering 1 offset from movzwl -> movl :)

llvm/test/CodeGen/X86/cmp-shrink-unsigned.ll
```diff
@@ -0,0 +1,44 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
+; RUN: llc < %s -mtriple=x86_64-- | FileCheck %s --check-prefixes=CHECK,GENERIC
```
phoebewang: Were you intended to test i386 but missed? No need two prefixes for single RUN.

llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -24285,6 +24285,26 @@ static SDValue EmitCmp(SDValue Op0, SDValue Op1, X86::CondCode X86CC,
     Op1 = DAG.getNode(ISD::TRUNCATE, dl, CmpVT, Op1);
   }
 
+  // Try to shrink i32 unsigned compares to i8 or i16 when both operands'
+  // high bits are known zero and neither operand is a constant. Eligible
+  // i64 compares are shrunk to i32 above.
+  if (CmpVT == MVT::i32 && !isX86CCSigned(X86CC) &&
```
phoebewang, line 24376: For signed case, we can use 7/15 instead?

llvm/test/CodeGen/X86/cmp-shrink-unsigned.ll
he-weiwen: it was left over from a previous version of the patch that used `hasFastImm()` - fixed.

llvm/lib/Target/X86/X86ISelLowering.cpp
he-weiwen: I added the signed case following the same logic from the i64 -> i32 narrowing, but the check lines drifts seems a bit concerning - should i revert?

phoebewang: ok, let's limit to unsigned only.

he-weiwen: done!

llvm/test/CodeGen/X86/cmp-shrink-unsigned.ll
```diff
@@ -0,0 +1,43 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
+; RUN: llc < %s -mtriple=x86_64-- | FileCheck %s
```
RKSimon: please can you add a i686 RUN as well:
```
; RUN: llc < %s -mtriple=i686-- | FileCheck %s --check-prefixes=CHECK,X86
; RUN: llc < %s -mtriple=x86_64-- | FileCheck %s --check-prefixes=CHECK,X64
```

he-weiwen: done!

llvm/test/CodeGen/X86/cmp-shiftX-maskX.ll
RKSimon: Will there always be some kind of zext instruction to remove?

llvm/test/CodeGen/X86/cmp-shrink-unsigned.ll
```diff
@@ -0,0 +1,63 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
+; RUN: llc < %s -mtriple=i686-- | FileCheck %s --check-prefixes=CHECK,X86
+; RUN: llc < %s -mtriple=x86_64-- | FileCheck %s --check-prefixes=CHECK,X64
```
RKSimon: Looks like you can remove the (unused) CHECK:
```
; RUN: llc < %s -mtriple=i686-- | FileCheck %s --check-prefixes=X86
; RUN: llc < %s -mtriple=x86_64-- | FileCheck %s --check-prefixes=X64
```

RKSimon (changes requested):

llvm/test/CodeGen/X86/cmp-shrink-unsigned.ll
he-weiwen: fixed

llvm/test/CodeGen/X86/cmp-shiftX-maskX.ll
he-weiwen: Gated the i16 case to only when there's an (and (X, 0XFFFF)) in the operand to remove. Apparently zext canonicalize to and so this catches both. Not sure this is the only profitable case though.

llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -24316,6 +24285,34 @@ static SDValue EmitCmp(SDValue Op0, SDValue Op1, X86::CondCode X86CC,
...
+    auto BothOpsFit = [&](unsigned Bits) {
+      APInt Hi = APInt::getHighBitsSet(32, 32 - Bits);
+      return DAG.MaskedValueIsZero(Op0, Hi) && DAG.MaskedValueIsZero(Op1, Hi);
+    };
+    // Narrowing to i16 is only when this enables an (and x, 0xFFFF) in an
+    // operand to be removed downstream
+    auto IsI16Eliminable = [](SDValue Op) {
+      if (Op.getOpcode() == ISD::AND)
+        if (auto *C = dyn_cast<ConstantSDNode>(Op.getOperand(1)))
+          return C->getAPIntValue().countr_one() >= 16;
```
RKSimon: I don't understand what this is doing? The trailing ones in the mask have to be 16 bits or MORE?

he-weiwen: The idea was that if the constant mask has more than 16 trailing 1s, and we can prove the full expression fit in 16 bit, then it should be equivalent to `and(x, 0xFFFF)` after truncation. 

But it seems that case is canonicalized to `and(x, 0xFFFF)` already by the time we reach here, so `countr_one() == 16` would work the same.

he-weiwen: Changed the condition to a direct pattern match to `and(x, 0xFFFF)` & added a test case that shows the wider constant mask case is still covered under the new condition.

RKSimon (changes requested):
Make sure tests are added as the first commit (with baseline codegen) in the PR, so then later commits in the chain show the codegen diffs - this all then squashes when we push to main

he-weiwen:
> Make sure tests are added as the first commit (with baseline codegen) in the PR, so then later commits in the chain show the codegen diffs - this all then squashes when we push to main

fixed

llvm/test/CodeGen/X86/cmp-shrink-unsigned.ll
phoebewang: Why not get cmpw {{[0-9]+}}(%esp), %eax like above?

he-weiwen: The `IsI16Eliminable` profitability gate only recognises `and(%x, 0xFFFF)`, but here it arrives as a `zextload`.
I tried to also recognise `zextload` in `IsI16Eliminable`, but it still won't fold straightforwardly: 
```
movzwl 4(%esp), %eax
movzwl 8(%esp), %ecx
cmpw   %cx, %ax
```
It seems DAGCombine won't fold `truncate(zextload(i16->i32) -> i16)` to `load i16`, as `isTypeDesirableForOp(ISD::LOAD, i16)` returns `false` in `visitTRUNCATE`.

This particular case looks profitable though, but I feel like figuring out which exact form to match and where to place the combine belong to a separate PR?

PR #197802, opened 2026-05-14
[lldb][Clang] Removed redundant code in DWARFASTParserClang

Iasonaskrpr:
decl_up is initialized here but doesn't appear to be used or moved anywhere before it goes out of scope. If the Declaration info isn't needed for the FunctionSP, it should probably be removed.


JDevlieghere (approved):

Iasonaskrpr:
Thanks for the swift review! If everything looks good can you please merge it for me because I do not have commit access.

bulbazord (approved):
Nice catch

Issue #180420, opened 2026-02-08
Split `OPERAND_REG_IMM_INT64` into signed and unsigned


shiltian:
`OPERAND_REG_IMM_INT64` currently represents `b64`, `i64`, and `u64` operands. This wasn't a problem until newer targets started to support `src_literal64` encoding. We can't tell whether a literal is signed or unsigned, so for negative values we don't know its signedness. Because of that, we have to conservatively force `src_literal64` encoding, which isn't always the best choice.

We should split this into `OPERAND_REG_IMM_B64` and `OPERAND_REG_IMM_INT64`. The `B64` one would represent the unsigned case, and the `INT64` one would represent the signed case. Then we can safely use `IsUInt<32>` and `IsInt<32>` to decide whether `src_literal64` is needed.

PR #197803, opened 2026-05-14
[AMDGPU] Error out on too large lit() constants

rampitec:


rampitec:
* #197803 👈
* #197788
* `main`

<h2></h2>

llvm/test/MC/AMDGPU/literals.s
```diff
@@ -2008,16 +2008,16 @@ v_add_nc_u64 v[0:1], v[0:1], 0x123456789
...
 v_add_nc_u64 v[0:1], v[0:1], lit(0x123456789)
 // GFX1250-ASM: v_add_nc_u64_e64 v[0:1], v[0:1], lit(0x23456789) ; encoding: [0x00,0x00,0x28,0xd5,0x00,0xff,0x01,0x02,0x89,0x67,0x45,0x23]
 // GFX1250-DIS: v_add_nc_u64_e64 v[0:1], v[0:1], 0x23456789 ; encoding: [0x00,0x00,0x28,0xd5,0x00,0xff,0x01,0x02,0x89,0x67,0x45,0x23]
 // NOCI: :[[@LINE-3]]:1: error: instruction not supported on this GPU (bonaire): v_add_nc_u64
 // NOGFX11: :[[@LINE-4]]:1: error: instruction not supported on this GPU (gfx1100): v_add_nc_u64
 // NOGFX12: :[[@LINE-5]]:1: error: instruction not supported on this GPU (gfx1200): v_add_nc_u64
-// NOGFX9: :[[@LINE-6]]:1: error: instruction not supported on this GPU (gfx900): v_add_nc_u64
-// NOSI: :[[@LINE-7]]:1: error: instruction not supported on this GPU (tahiti): v_add_nc_u64
-// NOVI: :[[@LINE-8]]:1: error: instruction not supported on this GPU (tonga): v_add_nc_u64
+// NOGFX1250: :[[@LINE-6]]:1: error: invalid literal value
```
jayfoad: I'm confused why we have **both** GFX1250-* and NOGFX1250 checks for this instruction. Is it accepted on gfx1250 or not?

llvm/lib/Target/AMDGPU/AsmParser/AMDGPUAsmParser.cpp
```diff
@@ -2540,8 +2540,12 @@ void AMDGPUOperand::addLiteralImmOperand(MCInst &Inst, int64_t Val, bool ApplyMo
     // truncated to uint32_t), if the target doesn't support 64-bit literals, or
     // the lit modifier is explicitly used, we need to truncate it to the 32
     // LSBs.
-    if (!AsmParser->has64BitLiterals() || Lit == LitModifier::Lit)
+    if (!AsmParser->has64BitLiterals() || Lit == LitModifier::Lit) {
+      if (!isInt<32>(Val) && !isUInt<32>(Val))
+        const_cast<AMDGPUAsmParser *>(AsmParser)->Error(
+            Inst.getLoc(), "invalid literal value");
```
jayfoad: The error message could be more helpful, maybe something like "literal value out of range"?

jayfoad (commented):
I'm personally not a huge fan of this change. If we knew the signedness of the operand and could check _either_ isInt _or_ isUInt then I might like it more.

Anyway, I won't object if other reviewers like it.

What does SP3 do?

jayfoad:
> Error out on too long lit() constants

Maybe "too large"?

rampitec:
> > Error out on too long lit() constants
> 
> Maybe "too large"?

Ok.

llvm/lib/Target/AMDGPU/AsmParser/AMDGPUAsmParser.cpp
rampitec: Ok

llvm/test/MC/AMDGPU/literals.s
rampitec: Ouch. It emits the error and then does it anyway. I need to find a way to bail out completely.

rampitec:
> I'm personally not a huge fan of this change. If we knew the signedness of the operand and could check _either_ isInt _or_ isUInt then I might like it more.
> 
> Anyway, I won't object if other reviewers like it.
> 
> What does SP3 do?

Now we are silently truncating it. Same as SP3:

```
v_add_nc_u64 v[0:1], v[0:1], lit(0x123456789)

    0xd5280000
    0x0201ff00
    0x23456789
```
I do not see how a silent truncation is helpful.

llvm/test/MC/AMDGPU/literals.s
rampitec: I.e., it now looks like this:
```
llvm-mc -arch=amdgcn -mcpu=gfx1250 -show-encoding <<< 'v_add_nc_u64 v[0:1], v[0:1], lit(0x123456789)'
	v_add_nc_u64_e64 v[0:1], v[0:1], lit(0x23456789) ; encoding: [0x00,0x00,0x28,0xd5,0x00,0xff,0x01,0x02,0x89,0x67,0x45,0x23]
<stdin>:1:1: error: literal value out of range
v_add_nc_u64 v[0:1], v[0:1], lit(0x123456789)
^
```
I.e., llvm-mc completes with error, but produces output.

rampitec: Probably the validation here has to happen earlier, when we are parsing the immediate. There is a chance of proper bail there.

arsenm:
> I'm personally not a huge fan of this change. If we knew the signedness of the operand and could check _either_ isInt _or_ isUInt then I might like it more.

That is https://github.com/llvm/llvm-project/issues/180420

llvm/test/MC/AMDGPU/literals.s
rampitec: OK, I have found a better single place to check it. It errors out and does not emit instruction now, plus shows a correct SLoc of the literal.

rampitec:
> > I'm personally not a huge fan of this change. If we knew the signedness of the operand and could check _either_ isInt _or_ isUInt then I might like it more.
> 
> That is #180420

I do not think signedness really matters here. `lit()` requests it to be represented as a 32-bit literal. If the literal is expressed as an integer we just need to make sure we can fit it into these 32 bits. Regardless of the operand behavior, zext or sext in SP, is it int or fp, we can use either positive or negative constant in the asm, just to get bit pattern we want. Like we can use `v_mov_b32 v0, 1000` or `v_mov_b32 v0, -1000` for example.

arsenm (approved):

PR #197804, opened 2026-05-14
[flang][docs][nfc] Wordsmith -frelaxed-c-loc-checks flag description

akuhlens:
Clarify that unexpected behavior reflects the compiler's design choice
not to track aliases from non-target non-pointer C_LOC() arguments,
rather than implying a compiler defect. Align phrasing across
FlangOptions.td and Extensions.md.

Also add the `-checks` suffix to the flag per @sscalpone's suggestion.


eugeneepshteyn (approved):

eugeneepshteyn (approved):
LGTM

PR #197805, opened 2026-05-14
[lldb] Move GetSelectedTarget from Debugger to CommandInterpreter (NFC)

JDevlieghere:
The notion of a "selected target" is something that belongs at the command layer rather than the debugger layer. Jim and I were discussing how this is a foot-gun and I suggested moving it into the CommandInterpreter. Everyone else that needs to get their hands on the current target should do so through the execution context. The latter is trivial to swap out if the "selected one" isn't what you should be operating on.

### DIFF
--- a/lldb/source/Commands/CommandObjectProcess.cpp
+++ b/lldb/source/Commands/CommandObjectProcess.cpp
@@ -143,7 +143,7 @@ public:
 
 protected:
   void DoExecute(Args &launch_args, CommandReturnObject &result) override {
-    Target *target = GetCommandInterpreter().GetSelectedTarget().get();
+    Target *target = &GetTarget();
     // If our listener is nullptr, users aren't allows to launch
     ModuleSP exe_module_sp = target->GetExecutableModule();
 
@@ -313,7 +313,9 @@ protected:
     PlatformSP platform_sp(
         GetDebugger().GetPlatformList().GetSelectedPlatform());
 
-    Target *target = GetCommandInterpreter().GetSelectedTarget().get();
+    Target *target = &GetTarget();
+    if (target->IsDummyTarget())
+      target = nullptr;
     // N.B. The attach should be synchronous.  It doesn't help much to get the
     // prompt back between initiating the attach and the target actually
     // stopping.  So even if the interpreter is set to be asynchronous, we wait
@@ -915,15 +917,16 @@ protected:
     Status error;
     Debugger &debugger = GetDebugger();
     PlatformSP platform_sp = m_interpreter.GetPlatform(true);
-    Target *selected_target = GetCommandInterpreter().GetSelectedTarget().get();
+    Target *target = &GetTarget();
+    if (target->IsDummyTarget())
+      target = nullptr;
     ProcessSP process_sp =
         debugger.GetAsyncExecution()
             ? platform_sp->ConnectProcess(command.GetArgumentAtIndex(0),
-                                          plugin_name, debugger,
-                                          selected_target, error)
+                                          plugin_name, debugger, target, error)
             : platform_sp->ConnectProcessSynchronous(
                   command.GetArgumentAtIndex(0), plugin_name, debugger,
-                  result.GetOutputStream(), selected_target, error);
+                  result.GetOutputStream(), target, error);
     if (error.Fail() || process_sp == nullptr) {
       result.AppendError(error.AsCString("Error connecting to the process"));
     }
--- a/lldb/source/Commands/CommandObjectProcess.cpp
+++ b/lldb/source/Commands/CommandObjectProcess.cpp
@@ -143,8 +143,7 @@ public:
 
 protected:
   void DoExecute(Args &launch_args, CommandReturnObject &result) override {
-    Debugger &debugger = GetDebugger();
-    Target *target = debugger.GetSelectedTarget().get();
+    Target *target = GetCommandInterpreter().GetSelectedTarget().get();
     // If our listener is nullptr, users aren't allows to launch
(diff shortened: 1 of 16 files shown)


lldb/source/Commands/CommandObjectProcess.cpp
```diff
@@ -143,8 +143,7 @@ class CommandObjectProcessLaunch : public CommandObjectProcessLaunchOrAttach {
 
 protected:
   void DoExecute(Args &launch_args, CommandReturnObject &result) override {
-    Debugger &debugger = GetDebugger();
-    Target *target = debugger.GetSelectedTarget().get();
+    Target *target = GetCommandInterpreter().GetSelectedTarget().get();
```
jimingham: Here to.  Why are we not getting the target from m_exe_ctx or even better CommandObject::GetTarget...

jimingham:
So the fact that in all the obviously correct uses of GetSelectedTarget you were replacing:

interpreter.GetDebugger().GetSelectedTarget()

with

interpreter.GetSelectedTarget()

is a good sign this was in the wrong place and it belonged in the CommandInterpreter.  

Also, that most of the places where the code was just using GetDebugger().GetSelectedTarget() were arguably wrong is another good sign this shouldn't have been on the Debugger.

I think it would be interesting to keep the list of places you had to change here for further review.  They seem to fall into three classes:

MPC code/GUI IOHandler - Grouped because I don't really understand how either of them work...

HandleCompletion - Using the Selected Target is formally wrong here, though it probably doesn't matter.  If the selected Target was Target A, and Target B hit a breakpoint and for some odd reason the breakpoint callback wanted to do SBCommandInterpreter::HandleCompletion() it would get completion in the wrong context.  But it's not entirely clear why you would ever want to do this, and I'm not so sure how you would get the right context to HandleCompletion.  So while formally wrong we probably have to live with this.

A bunch of CommandObject subclasses were still calling GetSelectedTarget but so far as I can tell they really should be using CommandObject::GetTarget instead.  Any command that might get used in any of our callbacks (breakpoint commands, stop-hooks, etc) needs to get the target set in the command object and not the CommandInterpreter's Selected Target.  It's probably worth fixing them since this change made them evident, but as a separate PR.  

I can't think of a way to enforce "CommandObject subclasses' DoExecute implementation cannot call CommandInterpreter::GetSelectedTarget but their HandleCompletion methods can" for instance.  So I think we have to enforce "use GetTarget not interpreter.GetSelectedTarget in DoExecute" by auditing.

JDevlieghere:
> I can't think of a way to enforce "CommandObject subclasses' DoExecute implementation cannot call CommandInterpreter::GetSelectedTarget but their HandleCompletion methods can" for instance. So I think we have to enforce "use GetTarget not interpreter.GetSelectedTarget in DoExecute" by auditing.

Another inconsistency that bothers me is that some command calls m_exe_ctx->GetTargetPtr(). I'd like to rally around GetTarget, and maybe change it to do the right thing by default:

```
Target *CommandObject::GetTarget() {
  Target *target = m_exe_ctx.GetTargetPtr();
  if (!target)
    target = m_interpreter.GetExecutionContext().GetTargetPtr();
  if (GetFlags().Get() & (eCommandRequiresTarget | eCommandRequiresProcess |
                           eCommandRequiresThread | eCommandRequiresFrame))
    return target;  // CheckRequirements guaranteed real
  return (target && target->IsDummyTarget()) ? nullptr : target;
}
```

However, there are commands that explicitly want the dummy target without `eCommandRequiresTarget`, like `breakpoint set` so we'd have to have a (defaulted) flag to say "yes I really want the dummy target". I can do that as a follow up if that makes sense, and audit all the DoExecute methods. Then after we can audit all the `HandleCompletion` ones for `GetSelectedTarget`.

jimingham:
> > I can't think of a way to enforce "CommandObject subclasses' DoExecute implementation cannot call CommandInterpreter::GetSelectedTarget but their HandleCompletion methods can" for instance. So I think we have to enforce "use GetTarget not interpreter.GetSelectedTarget in DoExecute" by auditing.
> 
> Another inconsistency that bothers me is that some command calls m_exe_ctx->GetTargetPtr(). I'd like to rally around GetTarget, and maybe change it to do the right thing by default:
> 
> ```
> Target *CommandObject::GetTarget() {
>   Target *target = m_exe_ctx.GetTargetPtr();
>   if (!target)
>     target = m_interpreter.GetExecutionContext().GetTargetPtr();
>   if (GetFlags().Get() & (eCommandRequiresTarget | eCommandRequiresProcess |
>                            eCommandRequiresThread | eCommandRequiresFrame))
>     return target;  // CheckRequirements guaranteed real
>   return (target && target->IsDummyTarget()) ? nullptr : target;
> }
> ```
> 
> However, there are commands that explicitly want the dummy target without `eCommandRequiresTarget`, like `breakpoint set` so we'd have to have a (defaulted) flag to say "yes I really want the dummy target". I can do that as a follow up if that makes sense, and audit all the DoExecute methods. Then after we can audit all the `HandleCompletion` ones for `GetSelectedTarget`.

That makes sense to me - Commands want the "right" target for their purposes, but shouldn't have to think about that too hard.  As a side-effect, this would make it easy to see which command can update the dummy target (priming future targets) and which only operate on a "real" target.

jimingham (approved):
LGTM
Cleaning up the "do I want the dummy target or not" is a good follow-on but clearly a separate PR.

JDevlieghere:
> Cleaning up the "do I want the dummy target or not" is a good follow-on but clearly a separate PR.

Added to my list :-)

PR #197806, opened 2026-05-14
[CIR][AMDGPU] Add support for AMDGCN cos builtins

ayokunle321:
Adds codegen for the following AMDGCN cos builtins:

- __builtin_amdgcn_cosf (float)
- __builtin_amdgcn_cosh (half)
- __builtin_amdgcn_cos_bf16 (bfloat16)

These are lowered to the corresponding `llvm.amdgcn.cos` intrinsic.


clang/test/CIR/CodeGenHIP/builtins-amdgcn-gfx1250.hip
```diff
@@ -0,0 +1,26 @@
+#include "../CodeGenCUDA/Inputs/cuda.h"
```
arsenm: Why include a header? Usually we just defined or directly use the attribute in these tests. Also this test name doesn't match the reference test?

ayokunle321: Hi @arsenm,

I included the header because I was following the existing tests, e.g. https://github.com/llvm/llvm-project/blob/816e2924786a0d7d300aa9a2ff80e5a2a082e6fa/clang/test/CIR/CodeGenHIP/builtins-amdgcn.hip#L1

I'll switch it to use #define __device__ __attribute__((device)).

As for the file names, I was mistaken when following the patches in the incubator. I had been looking at HIP tests in the OpenCL test directory, which use different naming conventions ([incubator's test](https://github.com/llvm/clangir/blob/main/clang/test/CIR/CodeGen/HIP/builtins-amdgcn-gfx1250.hip)). I've found the correct directory and will rename the files accordingly.

ayokunle321: @arsenm I can't seem to find where the BF16 tests are kept. I don't see any in clang/test/CodeGenHIP, but I do see BF16 builtin tests in the OpenCL directory:

https://github.com/llvm/llvm-project/blob/e44d060dfd617e8a5e37b61adcb38ed4ac5d782d/clang/test/CodeGenOpenCL/builtins-amdgcn-gfx1250.cl#L288

That's probably why the incubator referenced the OpenCL ones. @skc7 could you please comment on this?

skc7: For bf16 AMDGCN math builtins, upstream/OGCG only tests them in OpenCL, which is what the incubator mirrored. `builtins-amdgcn-gfx1250.hip` is a fine name for it.

ranapratap55 (approved):
LGTM.

Please resolve merge conflicts before merge.

Issue #197792, opened 2026-05-14
[lldb] Multiple SBDebugger should be safe to use from different threads


JDevlieghere:
LLDB was designed to support multiple SBDebugger instances being used concurrently by different threads. The thread safety guarantee doesn't extend to LLDB's internal objects (such as SBTarget, SBProcess, etc) which are **not** individually thread-safe. In other words, multiple debuggers from multiple threads is safe as long as each debugger (and the objects it hands out) are owned and used by exactly one thread at a time. If you need to call into one debugger from multiple threads, you must serialize access yourself (e.g., with a mutex).

I have an (internal) reproducer that pressure tests this premise and shows that this is not currently the case. I'm using this issue to track the various issues that I'm uncovering that should get us to the situation described above.

rdar://151779712

jimingham:
That's half right - the part you mostly care about being the half that's right...

The SBDebugger itself for the most part doesn't do any locking around its calls, so if you want to call most SBDebugger API's for the same debugger on multiple threads safely, you do need keep to one thread or to synchronize.  But all the objects an SBDebugger governs should be isolated to that SBDebugger, or in the shared module cache which we manage for you.  So the design is that it should be safe to run multiple SBDebuggers each on its own dedicated thread; or each SBDebugger on multiple threads if you manage locking yourself within each SBDebugger.  There are some shared objects that we manage access to fairly well (the shared module cache) and other state we manage less well (particularly the log channel state).

Ironically, most of the objects from SBTarget and below are governed by the Target API lock, so it is for the most part safe to call them from multiple threads, letting us do the locking.  Like SBDebuggers, SBTargets should be independent of each other, so handing out access based on the target should allow you to manage separate target requests concurrently.

In the case of all the objects below the SBDebugger, the problem isn't really safety calling from multiple threads as that we don't have good way to deal with "I started to make a call for some data and another thread started the process running before I got the Target API lock, so I hung waiting for it, and then the world was totally different when I got a chance to stop."  To manage that you have to manage stop locking by hand although we do give you access to the stop locker so you can do that with our assistance.

JDevlieghere:
Thanks for the clarification Jim. That matches my understanding, and I was being a bit too hand-wavy about the other parts :-)

PR #197807, opened 2026-05-14
[lldb] Fix data race on Process::FindPlugin's unique-id counter

JDevlieghere:
g_process_unique_id is a function-local static used to hand out unique IDs to each Process. It was a plain uint32_t, so concurrent SBTarget::LoadCore / Target::CreateProcess calls raced on the increment.

Make it std::atomic<uint32_t>. Prefix operator++ on std::atomic is already an atomic fetch_add that returns the new value, so the call sites are unchanged.


jimingham (commented):
LGTM

PR #197809, opened 2026-05-14
[lldb] Fix data race on ValueObject's unique-id counter

JDevlieghere:
g_value_obj_uid is a file-scope static that hands out unique IDs to every ValueObject. It was a plain user_id_t, so concurrent SBTarget::FindGlobalVariables / EvaluateExpression calls raced on the increment.

Make it std::atomic<user_id_t>. Prefix operator++ on std::atomic is already an atomic fetch_add that returns the new value, so the call sites are unchanged.

### DIFF
--- a/lldb/source/ValueObject/ValueObject.cpp
+++ b/lldb/source/ValueObject/ValueObject.cpp
@@ -54,6 +54,7 @@
 #include "llvm/Support/Compiler.h"
 
 #include <algorithm>
+#include <atomic>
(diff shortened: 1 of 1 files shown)


jimingham (approved):
LGTM.  All ValueObjects are owned by some Target, so it is a little odd to use a truly global ID.  But since this isn't really visible but more just a token I think there's no reason to add complexity here.  Making it an atomic will do.

PR #197811, opened 2026-05-14
[lldb] Fix data race on ThreadPlan::GetNextID's plan-ID counter

JDevlieghere:
`g_nextPlanID` is a function-local static used to hand out unique ThreadPlan IDs. It was a plain uint32_t, so concurrent ThreadPlan constructors (e.g. each Process's private state thread queueing its base plan) raced on the increment.

Make it std::atomic<uint32_t>. Prefix operator++ on std::atomic is already an atomic fetch_add that returns the new value, so the call sites are unchanged.

### DIFF
--- a/lldb/source/Target/ThreadPlan.cpp
+++ b/lldb/source/Target/ThreadPlan.cpp
@@ -16,6 +16,8 @@
 #include "lldb/Utility/Log.h"
 #include "lldb/Utility/State.h"
 
+#include <atomic>
+
 using namespace lldb;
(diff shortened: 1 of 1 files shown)


jimingham (approved):
Like with the ValueObject ID, ThreadPlans are all owned by Processes, so it's odd we have a global ID.  But it doesn't really matter, these aren't being shown in a user-visible list command where gaps in one Process's plans would look weird.
So just making it an atomic to be safe is fine.

PR #197812, opened 2026-05-14
[lldb] Fix data race in ObjectFile::GetSectionList

JDevlieghere:
The early `m_sections_up == nullptr` check was performed outside the module mutex, so two threads sharing the same Module could both enter the branch and race on the write in CreateSections. Restructure so the check and populate both happen under the module mutex; this is a standard double-checked locking fix.

### DIFF
--- a/lldb/source/Symbol/ObjectFile.cpp
+++ b/lldb/source/Symbol/ObjectFile.cpp
@@ -603,20 +603,19 @@ void ObjectFile::ClearSymtab() {
 }
 
 SectionList *ObjectFile::GetSectionList(bool update_module_section_list) {
-  if (update_module_section_list) {
-    ModuleSP module_sp(GetModule());
-    if (module_sp) {
-      // The module mutex guards m_sections_up: check and populate under the
-      // lock so concurrent first-time initialization from multiple threads
-      // that share this Module doesn't race.
-      std::lock_guard<std::recursive_mutex> guard(module_sp->GetMutex());
-      if (m_sections_up == nullptr)
-        CreateSections(*module_sp->GetUnifiedSectionList());
-      return m_sections_up.get();
-    }
+  // Don't take the module lock here: DWARF indexing workers call into this
+  // path while the main thread holds the module lock during PreloadSymbols,
+  // which would deadlock. See ObjectFile::GetSymtab() for the analogous
+  // pattern. Use a per-ObjectFile recursive mutex instead, since some plugin
+  // implementations of CreateSections re-enter GetSectionList on the same
+  // ObjectFile (e.g. ObjectFileELF's .gnu_debugdata path).
+  std::lock_guard<std::recursive_mutex> guard(m_sections_mutex);
+  if (m_sections_up)
     return m_sections_up.get();
-  }
-  if (m_sections_up == nullptr) {
+  if (update_module_section_list) {
+    if (ModuleSP module_sp = GetModule())
+      CreateSections(*module_sp->GetUnifiedSectionList());
+  } else {
     SectionList unified_section_list;
     CreateSections(unified_section_list);
   }
--- a/lldb/source/Symbol/ObjectFile.cpp
+++ b/lldb/source/Symbol/ObjectFile.cpp
@@ -603,17 +603,22 @@ void ObjectFile::ClearSymtab() {
 }
 
 SectionList *ObjectFile::GetSectionList(bool update_module_section_list) {
-  if (m_sections_up == nullptr) {
-    if (update_module_section_list) {
-      ModuleSP module_sp(GetModule());
-      if (module_sp) {
-        std::lock_guard<std::recursive_mutex> guard(module_sp->GetMutex());
+  if (update_module_section_list) {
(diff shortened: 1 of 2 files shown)


lldb/source/Symbol/ObjectFile.cpp
```diff
@@ -603,17 +603,22 @@ void ObjectFile::ClearSymtab() {
...
         CreateSections(*module_sp->GetUnifiedSectionList());
-      }
-    } else {
-      SectionList unified_section_list;
-      CreateSections(unified_section_list);
+      return m_sections_up.get();
     }
+    return m_sections_up.get();
+  }
+  if (m_sections_up == nullptr) {
```
jimingham: Seems like there are two races in this function and you fixed one of them.
The "update_module_section_list" branch needs the module lock because it is directly updating a Module ivar.  So that part of the locking looks right.  It also indirectly keeps two simultaneous callers of `GetSectionList(true)` from fighting over actually populating m_sections_up.

But when you call CreateSections(false) on an object file with m_sections_up, it will still parse the object file and fill the m_sections_up.  So something needs to protect CreateSections that from concurrent access even though we don't need to protect the Module's section list.  

Some of the CreateSections overrides take the module lock (e.g. ObjectFileCOFF) and others don't (e.g. ObjectFileMachO).  So if two threads call UpdateModuleList(false) simultaneously, they will very likely fight over filling m_sections_up for the MachO case.  So we either need to make the rule that CreateSections also has to take the module lock, or we need to lock this other branch.

lldb/source/Symbol/ObjectFile.cpp
```diff
@@ -603,17 +603,22 @@ void ObjectFile::ClearSymtab() {
...
+    SectionList unified_section_list;
+    CreateSections(unified_section_list);
```
jasonmolenda, line 616: I'm having a hard time seeing how it's valid to call the subclass `CreateSections()` without the module lock being held.  I looked closely at ObjectFileMachO::CreateSections and less closely at ObjectFileELF::CreateSections and it seems like those methods assume the module lock is already held before they change the m_sections_up array.  From grepping around, I think the only caller that does `GetSectionList(/*update_module_section_list=*/false)` is SymbolVendorELF and SymbolFileDWARFDwo. 
(this isn't a change introduced by this PR, it's just something I'm noticing as I'm looking closely)

jasonmolenda (approved):
Change looks good to me, I'm uncertain about the case where we call CreateSections() without holding the module lock tho, a code sequence that already existed.

lldb/source/Symbol/ObjectFile.cpp
```diff
@@ -603,17 +603,21 @@ void ObjectFile::ClearSymtab() {
...
+  // which would deadlock. See ObjectFile::GetSymtab() for the analogous
+  // pattern. Use a per-ObjectFile recursive mutex instead, since some plugin
+  // implementations of CreateSections re-enter GetSectionList on the same
+  // ObjectFile (e.g. ObjectFileELF's .gnu_debugdata path).
+  std::lock_guard<std::recursive_mutex> guard(m_sections_mutex);
+  if (m_sections_up)
+    return m_sections_up.get();
+  if (update_module_section_list) {
+    if (ModuleSP module_sp = GetModule())
+      CreateSections(*module_sp->GetUnifiedSectionList());
```
jimingham, line 612: Since this CreateSections call updates the Module's Unified Section List I think you have to also take the Module mutex in this branch.  Otherwise two different ObjectFiles from the same module could come in and try to update the Module section list simultaneously.
CreateSections returns a list of the new modules, so maybe it would be better to not pass in the module list here, but pass in a new list and then have a separate `Module::UpdateSectionList` that just merges the lists.  That way you'd only need to lock around the update which should be pretty fast, and not all the work to build section lists.

jimingham: This is correct, and fine for this PR.  

It should be possible to do:

```
if (Module_sp = GetModule()) {
    SectionList new_sections;
    CreateSections(new_sections);
    {  // Scope for module lock
      std::lock_guard<std::recursive_mutex> guard(module_sp->GetMutex());
      SectionList &mod_unified_sl = module_sp->GetUnifiedSectionList();
      for (auto section : new_sections)
        mod_unified_sl.AddSection(section);
    }
  }

```

which would narrow the amount of work done under the module mutex.

However, while that's what most of the CreateSection implementations do, some of them seem to do a fancier job of merging the new sections into the SectionList that was passed in (which may or may not be the UnifiedSectionList...).  

It's a little odd that the ObjectFile controls how to merge sections into the UnifiedSectionList, but if that's really necessary it might be good to separate CreateSections from "merge into unified list".

It's also a little unfortunate that you have to pass in a section list because the other call to CreateSections here fills a SectionList that it promptly discards.

But I think this is for a future refactor.

jimingham (approved):
LGTM

If you want to avoid locking the module mutex while building the sections we can separate CreateSections from AddToUnifiedModuleList, but that seems orthogonal to this patch.

PR #197814, opened 2026-05-14
[flang][HLFIR] Add SeparateAllocatableAssign pass

khaki3:
Example:
```fortran
!$acc kernels
B = A          ! A, B allocatable
!$acc end kernels
```

In this code, `B = A` lowers to `hlfir.assign ... realloc`, which becomes a `_FortranAAssign` runtime call inside the compute region — the allocation can't be separated from the copy, and it crashes when `B` is unallocated.

Fix: add `SeparateAllocatableAssign`, which splits the realloc assign into an explicit conditional [re-]allocation followed by a plain `hlfir.assign`, exposing the allocation as plain FIR. For variable RHS it skips aliasing cases (`a = a(:n)`) by loading the LHS `fir.box` and querying `fir::AliasAnalysis` on the data. For `hlfir.expr` RHS it leaves ordering to bufferization, and fixes `ElementalAssignBufferization` to stop fusing an elemental across a deallocation.


flang/lib/Optimizer/HLFIR/Transforms/InlineHLFIRAssign.cpp
```diff
@@ -299,12 +411,12 @@ class InlineHLFIRAssignPass
     mlir::RewritePatternSet patterns(context);
     patterns.insert<InlineHLFIRAssignConversion>(context);
 
-    // Optionally add the allocatable expr assignment pattern
     if (inlineAllocatableExprAssignFlag) {
```
eugeneepshteyn, line 317: It seems that `-inline-hlfir-allocatable-expr-assign` flag now controls both expressions and vars. We should either update the option/flag name, or add another flag.
```
$ fir-opt --help | grep allocatable
  --inline-hlfir-allocatable-expr-assign  - Enable inlining of allocatable assignments when RHS is an hlfir.expr (e.g., from hlfir.elemental)
```

flang/lib/Optimizer/HLFIR/Transforms/InlineHLFIRAssign.cpp
```diff
@@ -283,6 +283,118 @@ class InlineAllocatableExprAssignConversion
...
+    hlfir::Entity rhs{assign.getRhs()};
+    hlfir::Entity lhs{assign.getLhs()};
+
+    // This pattern handles variable RHS (the expr case is handled by
+    // InlineAllocatableExprAssignConversion).
+    if (mlir::isa<hlfir::ExprType>(rhs.getType()))
+      return rewriter.notifyMatchFailure(
+          assign, "RHS is hlfir.expr - use expr pattern");
+
+    if (!rhs.isArray())
```
eugeneepshteyn: What about checking `lhs.isArray()` ?

flang/lib/Optimizer/HLFIR/Transforms/InlineHLFIRAssign.cpp
eugeneepshteyn, line 141: AI reviewer pointed out that access groups attribute propagation is not happening in the new code. I don't know, if it's important.

flang/lib/Optimizer/HLFIR/Transforms/InlineHLFIRAssign.cpp
```diff
@@ -283,6 +283,118 @@ class InlineAllocatableExprAssignConversion
...
+///     %lhs_elem = hlfir.designate %lhs_box (%i) : ...
+///     hlfir.assign %rhs_val to %lhs_elem : T, !fir.ref<T>
+///   }
+class InlineAllocatableVarAssignConversion
+    : public mlir::OpRewritePattern<hlfir::AssignOp> {
+public:
+  using mlir::OpRewritePattern<hlfir::AssignOp>::OpRewritePattern;
+
+  llvm::LogicalResult
+  matchAndRewrite(hlfir::AssignOp assign,
```
eugeneepshteyn: There is some code duplication between the variable case and the original expression case in `InlineAllocatableExprAssignConversion`. Is it possible to cleanly refactor, so that they share some common code?

vzakhari:
The change looks good to me, but I think conceptually there is no need to pair this with InlineHLFIRAssign. As I understand, you care about separating the potential re-allocation of the LHS from the actual assignment (even if it is done via the runtime call). If it is the case, then shouldn't we just have a separate pass that transforms an `hlfir.assign ... realloc` into `hlfir.assign <w/o realloc>` where the new LHS is a conditionally [re-]allocated original LHS?

The new pass may even run at `-O0` whereas InlineHLFIRAssign is supposed to run at higher opt levels.

P.S. We may need to improve FIR AA to figure out the source of the conditionally [re-]allocated LHS so that the original InlineHLFIRAssign can work after the new pass.

flang/lib/Optimizer/HLFIR/Transforms/SeparateAllocatableAssign.cpp
```diff
@@ -0,0 +1,140 @@
+//===- SeparateAllocatableAssign.cpp - Split realloc from assign
+//-----------===//
```
tblah: nit

flang/lib/Optimizer/HLFIR/Transforms/SeparateAllocatableAssign.cpp
```diff
@@ -0,0 +1,140 @@
...
+    fir::MutableBoxValue mutableBox(lhs.getFirBase(), /*lenParameters=*/{},
+                                    /*mutableProperties=*/{});
+
+    auto noopHandler = [](fir::ExtendedValue) {};
+    llvm::SmallVector<mlir::Value> lenParams;
+    fir::factory::MutableBoxReallocation realloc =
+        fir::factory::genReallocIfNeeded(builder, loc, mutableBox, rhsExtents,
+                                         lenParams, noopHandler);
+    fir::factory::finalizeRealloc(builder, loc, mutableBox, /*lbounds=*/{},
+                                  /*takeLboundsIfRealloc=*/true, realloc);
```
tblah, line 137: I think this needs guarding with a check for lhs and rhs pointing to the same array.

For `a = a(:n)`, the generated FIR builds the RHS box from the old allocation, frees that allocation when reallocating LHS, then passes the old RHS box to _FortranAAssign.

(Found by codex)

tblah (commented):
Thanks for the patch.

Currently both paths loose lower bounds. I don't think this is right.

```
program lower_bound_realloc
  integer :: source(10:12)
  integer, allocatable :: dest(:)

  source = [1, 2, 3]
  dest = source

  if (lbound(dest, 1) /= 10) error stop 1
end program
```
(Reproducer written by codex)

flang/lib/Optimizer/HLFIR/Transforms/SeparateAllocatableAssign.cpp
vzakhari: Good catch! Yes, in this case we may probably use `hlfir.as_expr` for the RHS. I believe in `a = a` corner case we do not have to do this, because there will be no reallocation.

In case of `a = a(lbound(a,1):ubound(a,1)))` we will create a  temporary, but I do not remember whether Fortran runtime will be able to avoid it. So this may or may not be an issue. Maybe we should just rely on Assign runtime in these cases and do not apply this transformation.

jeanPerier: I think `hlfir.as_expr` may lead to create more temporary/allocation than needed.

I believe that in case of reallocation, the runtime is clever enough to skip the RHS temporary and to use the new storage as a "temporary" for the new value, so there may be benefits of keeping the realloc+assign grouped for overlapping assignments.

khaki3: Thanks for all the comments. `SeparateAllocatableAssign` now skips the rewrite when LHS and RHS may overlap, so cases like `a = a(:n)` stay as a combined `hlfir.assign ... realloc` and continue to use the runtime's overlap-aware handling.

vzakhari:
Thank you for the updates, Matsu! I suppose the changes in `InlineHLFIRAssign.cpp` are not needed now - can you please revert them?

khaki3:
Yes, SeparateAllocatableAssign and the exisiting InlineHLFIRAssign are sufficient to resolve the case. I'm dropping the change.

flang/lib/Optimizer/HLFIR/Transforms/SeparateAllocatableAssign.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+    if (!fir::isBoxAddress(lhs.getType()))
+      return rewriter.notifyMatchFailure(assign, "LHS is not a box address");
+
+    // Reallocation frees the old LHS storage. If RHS references that same
+    // storage (e.g. a = a(:n)), the freed pointer would be read by the
+    // subsequent non-realloc assign, causing use-after-free.  Walk the
+    // RHS use-def chain; if any operand leads back to the LHS value,
+    // the two may alias and we must keep the original realloc assign.
+    if (!mlir::isa<hlfir::ExprType>(rhs.getType())) {
```
tblah, line 101: Expr RHS might still alias with the lhs. For example (from codex):
```
 subroutine self_expr(a, n)
    integer :: n
    integer, allocatable :: a(:)
    a = a(:n) + 1
  end
```

flang/lib/Optimizer/HLFIR/Transforms/SeparateAllocatableAssign.cpp
```diff
@@ -0,0 +1,173 @@
...
+      llvm::SmallVector<mlir::Value> worklist = {rhs};
+      llvm::DenseSet<mlir::Value> visited;
+      bool mayAlias = false;
```
tblah: Please use fir::AliasAnalysis for this sort of thing. There can be difficult corner cases and all kinds of pass-through "view-like" operations which make aliasing complicated. It is better not to re-invent the wheel.

tblah (commented):
After today I will be offline for a couple of weeks. I trust the other reviewers of the patch to verify that my comments are resolved so if they approve please feel free to merge without another round of review from me.

khaki3:
Thank you for the reviews. I've handled all the concerns.

flang/lib/Optimizer/HLFIR/Transforms/SeparateAllocatableAssign.cpp
```diff
@@ -0,0 +1,187 @@
...
+      return rewriter.notifyMatchFailure(assign, "LHS is not a box address");
+
+    // Reallocation frees the old LHS storage. If the RHS reads that same
+    // storage (e.g. `a = a(:n)` or `a = a(:n) + 1`), the freed pointer would
+    // be read by the subsequent non-realloc assign, causing use-after-free.
+    //
+    // Reading allocatable storage always goes through a `fir.load` of the
+    // descriptor, so walking the RHS def-chain and querying alias analysis
+    // on any encountered load's memref soundly detects aliasing without
+    // running into the descriptor-vs-data NoAlias quirk of fir::AliasAnalysis
```
jeanPerier: The assumption that you will have a fir.load is a bit too aggressive. There could very well be an impure elemental that is accessing the allocatable indirectly (which is at least possible if it is a module variable or has a TARGET attribute).

There is no way around doing a proper alias analysis. I understand that part of the difficulty is that you do not have an SSA value for the LHS data.

Another example where what you are doing is incorrect is if you have `a = returns_func_pointer()` where the function pointer returned by the function points to the allocatable.

Because the function pointer result descriptor is a local alloca that is distinct from `a` descriptor, you will get false negative on the fact that the RHS is not using the data. In other words, the data pointed by `a` may be accessed via another descriptor in the `RHS` that does not alias with `a` descriptor. So you cannot reduce the problem to a descriptor aliasing problem.

The easiest solution I see to this problem is to generate a load of the LHS fir.box and to use the alias analysis framework with that, and to erase that load afterward. The long term solution would probably to be able to add an entry point to the alias analysis that we care about the data being pointed to by a descriptor address, but this is likely a bit more involving change.

flang/lib/Optimizer/HLFIR/Transforms/SeparateAllocatableAssign.cpp
jeanPerier: I disagree with codex here in the sense that as long as the deallocation is not moved above the hlfir.expr producer, it is a problem of the hlfir.assign lowering/optimization to detect that it can move the hlfir.expr evaluation into the assignment.

For instance this should be done [by this code](https://github.com/llvm/llvm-project/blob/8466750c230d5e51d509a8071c8a50e7de390029/flang/lib/Optimizer/HLFIR/Transforms/OptimizedBufferization.cpp#L388) for instance for hlfir.elemental inlining into the assignment (arguably this analysis may currently ignoring the allocation/deallocation effects which is bad).

You probably want to double check that this code is correctly not ignoring mem free effects, and if so I think you can just focus on analyzing variable RHS and leave it up to the expression bufferization to not move the hlfir.expr evaluation over deallocation.

jeanPerier (changes requested):
The overlap checking is still unsafe/not generic enough, otherwise this looks like the good direction to me.

flang/lib/Optimizer/HLFIR/Transforms/SeparateAllocatableAssign.cpp
khaki3: Thanks for the review. The current code does alias analysis only for variable RHS.

flang/lib/Optimizer/HLFIR/Transforms/SeparateAllocatableAssign.cpp
khaki3: The current PR generates and erases a load instead of finding one.

jeanPerier (approved):
Thanks for the update

PR #197816, opened 2026-05-14
[lldb] Fix data race in UnwindTable::Initialize

JDevlieghere:
UnwindTable::Initialize uses double-checked locking (DCL) against m_scanned_all_unwind_sources, but the flag was a plain bool. The fast-path read outside the mutex is unsynchronized with the write inside the mutex, so concurrent first-time callers can observe partially initialized state.

Make m_scanned_all_unwind_sources std::atomic<bool>. The default memory ordering (seq_cst) is sufficient for the DCL pattern used here.

### DIFF
--- a/lldb/include/lldb/Symbol/UnwindTable.h
+++ b/lldb/include/lldb/Symbol/UnwindTable.h
@@ -9,6 +9,7 @@
 #ifndef LLDB_SYMBOL_UNWINDTABLE_H
 #define LLDB_SYMBOL_UNWINDTABLE_H
 
+#include <atomic>
 #include <map>
 #include <mutex>
 #include <optional>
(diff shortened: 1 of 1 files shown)


jasonmolenda:
It seems like it would be simpler to only access this ivar when the lock is held, wouldn't it.  The only place where we're not doing that is this attempted shortcut,
```
void UnwindTable::Initialize() {
  if (m_scanned_all_unwind_sources)
    return;

  std::lock_guard<std::mutex> guard(m_mutex);

  if (m_scanned_all_unwind_sources) // check again once we've acquired the lock
    return;
```

and changing this to an atomic eliminates any savings from avoiding the mutex doesn't it.

JDevlieghere:
> and changing this to an atomic eliminates any savings from avoiding the mutex doesn't it.

No, the atomic is still faster because if `m_scanned_all_unwind_sources` is true, we can return without waiting on the lock to be released while the other thread is still doing work.

jasonmolenda (approved):
LGTM.

PR #197817, opened 2026-05-14
[clang][NFC] Remove `macro_begin` and `macro_end` from `Preprocessor`

davidstone:
Use `macros()` wherever posible. This gives us the following advantages:

1. We can use the range-base for loop for simpler looking code.
2. We more ergonomically use algorithms.
3. We can avoid the duplicate work of checking if we need to call `ExternalSource->ReadDefinedMacros()` that was in both `macro_begin` and `macro_end`. In some cases, we save this extra work once per loop iteration, not just one extra total.
4. No user confusion deciding which version they should call
5. Reduce the size of `Preprocessor.h` and the number of members in `Preprocessor`.

Using `macros` ends up being the better solution in every case, so since all callers were migrated from `macro_begin` and `macro_end` to `macros`, get rid of `macro_begin` and `macro_end`.

### DIFF
--- a/clang-tools-extra/clang-tidy/bugprone/BadSignalToKillThreadCheck.cpp
+++ b/clang-tools-extra/clang-tidy/bugprone/BadSignalToKillThreadCheck.cpp
@@ -32,10 +32,10 @@ void BadSignalToKillThreadCheck::check(const MatchFinder::MatchResult &Result) {
     return KeyValue.first->getName() == "SIGTERM" &&
            KeyValue.first->hasMacroDefinition();
   };
-  const auto macros = PP->macros();
+  const auto Macros = PP->macros();
   const auto TryExpandAsInteger =
       [&](Preprocessor::macro_iterator It) -> std::optional<unsigned> {
-    if (It == macros.end())
+    if (It == Macros.end())
       return std::nullopt;
     const MacroInfo *MI = PP->getMacroInfo(It->first);
     const Token &T = MI->tokens().back();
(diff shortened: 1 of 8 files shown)


clang-tools-extra/clang-tidy/bugprone/NotNullTerminatedResultCheck.cpp
```diff
@@ -796,25 +791,21 @@ void NotNullTerminatedResultCheck::check(
...
+    for (const auto &M : PP->macros()) {
+      if (M.first->getName() != "__STDC_WANT_LIB_EXT1__")
+        continue;
+      const auto *MI = PP->getMacroInfo(M.first);
+      // PP->getMacroInfo() returns nullptr if macro has no definition.
+      if (!MI)
+        continue;
+      const auto &T = MI->tokens().back();
+      if (T.isLiteral() && T.getLiteralData()) {
+        const StringRef ValueStr = StringRef(T.getLiteralData(), T.getLength());
```
localspook: To avoid repeating the type name:
```suggestion
        const StringRef ValueStr(T.getLiteralData(), T.getLength());
```

clang-tools-extra/clang-tidy/bugprone/BadSignalToKillThreadCheck.cpp
```diff
@@ -55,7 +56,7 @@ void BadSignalToKillThreadCheck::check(const MatchFinder::MatchResult &Result) {
     return IntValue.getZExtValue();
   };
 
-  const auto SigtermMacro = llvm::find_if(PP->macros(), IsSigterm);
+  const auto SigtermMacro = llvm::find_if(Macros, IsSigterm);
```
localspook, line 59: This is preexisting, so I'm not asking for changes in this PR, but it seems inefficient to be doing an O(n) walk over all defined macros instead of an O(1) lookup in the underlying map. Same thing in the other check.

localspook (approved):
clang-tidy changes LGTM

clang-tools-extra/clang-tidy/bugprone/BadSignalToKillThreadCheck.cpp
davidstone: Yeah, I noticed that, too. I was considering a follow-up PR for that, but I have a bunch of other stuff that I'll be getting to first.

PR #197819, opened 2026-05-14
[flang][acc] Accept component of global variable in `acc declare`

razvanlupusoru:
This MR partially extends the current implementation to accept cases of `acc declare` on a `parent%comp` whenever the `parent` has been `acc declare`d with the same clause. This is done by generating only the acc global constructor only for mapping the parent as the child is expected to be part of parent.

The limitations still remain as a TODO unless it can be proven parent is mapped. A generic implementation would need either compiler generated ordering on the global constructors used for mapping or runtime managed ordering.


flang/lib/Lower/OpenACC.cpp
```diff
@@ -3750,11 +3750,36 @@ static void createDeclareDeallocFunc(mlir::OpBuilder &modBuilder,
...
+           Fortran::semantics::Symbol::Flag::AccCopy,
+           Fortran::semantics::Symbol::Flag::AccCopyOut,
+           Fortran::semantics::Symbol::Flag::AccDevicePtr,
+           Fortran::semantics::Symbol::Flag::AccDeviceResident,
+           Fortran::semantics::Symbol::Flag::AccLink,
+           Fortran::semantics::Symbol::Flag::AccPresent,
+       }) {
+    if (ultimate.test(f)) {
+      assert(!found.has_value() &&
+             "more than one ACC DECLARE data-mapping flag on a symbol");
```
clementval: Should this be an error?

clementval (approved):

flang/lib/Lower/OpenACC.cpp
razvanlupusoru: I made it as an assert because it is an assumption that should always be true. It currently always holds because Semantics ensures this never happens. Consider:

```
module mm
  real, dimension(10) :: arr
  !$acc declare copyin(arr)
end module
program main
  use mm
  !$acc declare create(arr)
  arr = 10
  print *, arr
end program
```
=>
```
$ bbc -fopenacc moduleaccdeclare.f90
moduleaccdeclare.f90:8:17: error: 'arr' in the CREATE clause is already present in another COPYIN clause in this module
    !$acc declare create(arr)
                  ^^^^^^^^^^^
bbc: semantic errors in moduleaccdeclare.f90
```

Does this make sense or am I misunderstanding the difference between use of `assert` vs `fir::emitFatalError` ?

clementval: assertion are not always enabled. So if this is forbidden then it should be an error. If semantics ensures it normally doesn't happen then an assert might be enough.

razvanlupusoru: Thank you for your explanation - I decided to update it to fir::emitFatalError. :)

PR #197821, opened 2026-05-14
[flang][PPC] Improve vector type names in expression diagnostics (NFC)

virsworld:
## Summary

Improve Flang semantic diagnostics for vector types by printing vector type spellings in Fortran syntax instead of internal builtin-derived type names.

## Problem

Diagnostics for invalid operations on vector types currently display internal type names such as:

- `TYPE(_(element_category=0_4,element_kind=4_4))`
- `__builtin_ppc_intrinsic_vector(element_category=0_4,element_kind=4_4)`

This makes errors harder to read.

## Change

Use `DerivedTypeSpec::VectorTypeAsFortran()` when producing vector operand type names in `ArgumentAnalyzer::TypeAsFortran()`, instead of relying on the generic type formatting path.

This updates diagnostics to print vector types as Fortran source-like spellings, such as:
- `vector(integer(4))`
- `vector(real(4))`

while preserving the existing formatting for non-vector derived types.

## Examples

```text
vector(integer(4)) :: v1, v2
v1 = v1 + v2
end
```

Before:

```text
error: Operands of + must be numeric; have __builtin_ppc_intrinsic_vector(element_category=0_4,element_kind=4_4) and __builtin_ppc_intrinsic_vector(element_category=0_4,element_kind=4_4)
```

After:

```text
error: Operands of + must be numeric; have vector(integer(4)) and vector(integer(4))
```
---

```text
vector(integer(4)) :: v1
vector(real(4)) :: v2
v1 = v2
end
```

Before:

```text
error: No intrinsic or user-defined ASSIGNMENT(=) matches operand types __builtin_ppc_intrinsic_vector(element_category=0_4,element_kind=4_4) and __builtin_ppc_intrinsic_vector(element_category=2_4,element_kind=4_4)
```

After:

```text
error: No intrinsic or user-defined ASSIGNMENT(=) matches operand types vector(integer(4)) and vector(real(4))
```

## Testing

Tested invalid vector diagnostics for:
- numeric operators
- relational operators
- assignment mismatch

to confirm that vector operand types are now printed with the expected Fortran-style formatting.

### DIFF
--- a/flang/lib/Semantics/expression.cpp
+++ b/flang/lib/Semantics/expression.cpp
@@ -5538,7 +5538,9 @@ std::string ArgumentAnalyzer::TypeAsFortran(std::size_t i) {
         : type->IsUnlimitedPolymorphic() ? "CLASS(*)"s
         : type->IsPolymorphic()          ? type->AsFortran()
         : type->category() == TypeCategory::Derived
-        ? "TYPE("s + type->AsFortran() + ')'
+        ? (type->GetDerivedTypeSpec().IsVectorType()
+                  ? type->AsFortran()
+                  : "TYPE("s + type->AsFortran() + ')')
         : type->category() == TypeCategory::Character
         ? "CHARACTER(KIND="s + std::to_string(type->kind()) + ')'
         : ToUpperCase(type->AsFortran());
--- a/flang/lib/Evaluate/formatting.cpp
+++ b/flang/lib/Evaluate/formatting.cpp
@@ -676,6 +676,9 @@ llvm::raw_ostream &StructureConstructor::AsFortran(llvm::raw_ostream &o) const {
 std::string DynamicType::AsFortran() const {
   if (derived_) {
     CHECK(category_ == TypeCategory::Derived);
+    if (derived_->IsVectorType()) {
+      return derived_->VectorTypeAsFortran();
+    }
     std::string result{DerivedTypeSpecAsFortran(*derived_)};
     if (IsPolymorphic()) {
       result = "CLASS("s + result + ')';
--- a/flang/test/Semantics/PowerPC/ppc-vector-diagnostics.f90
+++ b/flang/test/Semantics/PowerPC/ppc-vector-diagnostics.f90
@@ -1,0 +1,15 @@
+! RUN: %python %S/../test_errors.py %s %flang_fc1
+! REQUIRES: target=powerpc{{.*}}
+
+subroutine test_vector_add()
+  vector(integer(4)) :: v1, v2
+  !ERROR: Operands of + must be numeric; have vector(integer(4)) and vector(integer(4))
+  v1 = v1 + v2
+end subroutine
+
+subroutine test_vector_assignment()
+  vector(integer(4)) :: v1
+  vector(real(4)) :: v2
+  !ERROR: No intrinsic or user-defined ASSIGNMENT(=) matches operand types vector(integer(4)) and vector(real(4))
+  v1 = v2
+end subroutine


flang/lib/Semantics/expression.cpp
```diff
@@ -5538,7 +5538,9 @@ std::string ArgumentAnalyzer::TypeAsFortran(std::size_t i) {
         : type->IsUnlimitedPolymorphic() ? "CLASS(*)"s
         : type->IsPolymorphic()          ? type->AsFortran()
         : type->category() == TypeCategory::Derived
-        ? "TYPE("s + type->AsFortran() + ')'
+        ? (type->GetDerivedTypeSpec().IsVectorType()
+                  ? type->GetDerivedTypeSpec().VectorTypeAsFortran()
+                  : "TYPE("s + type->AsFortran() + ')')
```
cenewcombe, line 5543: Implementing the fix here leaves several diagnostic pathways unchanged; they will still return the internal type names. Why not place make to the call to `VectorTypeAsFortran()` in `DynamicType::AsFortran` in formatting.cpp?
```
std::string DynamicType::AsFortran() const {
  if (derived_) {
    CHECK(category_ == TypeCategory::Derived);
    if (derived_->IsVectorType()) {              ! new
      return derived_->VectorTypeAsFortran();    ! new
    }                                            ! new
...
```

Then the "true" case in your fix needs only:
```
? type->AsFortran()
```

There is nothing wrong with what you have here, but moving the call to `VectorTypeAsFortran` will give the fix further reach for diagnostics.

virsworld: Good point, I will test this out and amend my commit.

virsworld: Amended, also added a LIT test for the examples seen above.

cenewcombe (approved):
LGTM, but I am requesting a review from @kkwli, since I am not as familiar with PowerPC.

kkwli (approved):
LG. Thanks.

kparzysz:
This breaks build with ENABLD_SHARED_LIBS=ON:
```
/usr/bin/ld: tools/flang/lib/Evaluate/CMakeFiles/FortranEvaluate.dir/formatting.cpp.o: in function `Fortran::evaluate::DynamicType::AsFortran[abi:cxx11]() const':
formatting.cpp:(.text._ZNK7Fortran8evaluate11DynamicType9AsFortranB5cxx11Ev+0x39): undefined reference to `Fortran::semantics::DerivedTypeSpec::VectorTypeAsFortran[abi:cxx11]() const'
```

The problem is that Evaluate is a dependency of Semantics, so you can't just add Semantics to the depenencies of Evaluate.

cenewcombe:
I reverted this since I made the merge commit: https://github.com/llvm/llvm-project/pull/198908. Let me know if I've done that incorrectly.

kparzysz:
Thanks @cenewcombe. We've run into this sort of an issue before.  The layering between Evaluate and Semantics is either completely broken, or they should be in the same library.  One way to fix this would be to declare the function/class in a header in Evaluate, but define it in Semantics.

cenewcombe:
> Thanks @cenewcombe. We've run into this sort of an issue before. The layering between Evaluate and Semantics is either completely broken, or they should be in the same library. One way to fix this would be to declare the function/class in a header in Evaluate, but define it in Semantics.

@virsworld FYI -- sorry I didn't catch that in review, and had to revert the merge commit. You'll need to submit a new PR.

virsworld:
> > Thanks @cenewcombe. We've run into this sort of an issue before. The layering between Evaluate and Semantics is either completely broken, or they should be in the same library. One way to fix this would be to declare the function/class in a header in Evaluate, but define it in Semantics.
> 
> @virsworld FYI -- sorry I didn't catch that in review, and had to revert the merge commit. You'll need to submit a new PR.

No worries, I'll open a new PR that builds with shared lib flag. I'll try what @kparzysz suggested.

PR #197826, opened 2026-05-14
[clangd][Lex][NFC] Use valid non-ASCII identifiers in tests

argothiel:
Several tests used "ab🙂cd" / "🙂cd" as multi-byte UTF-8 example identifiers. The smiley, however, is not actually among the allowed identifier characters, and Clang only accepts it as an extension (with a warning).

Switch to identifiers that are valid per [lex.name]:
- "naïve", with a non-ASCII char in the middle of the identifier,
- "æon", with a non-ASCII char at the start of the identifier,
- "café", with a non-ASCII char at the end of the identifier.

The 2-byte characters are handled the same way as the original 4-byte emoji; no functional change here.


timon-ul (approved):
Looks good to me, thank you for the cleanup 🙂 (you will be missed smiley face)

PR #197827, opened 2026-05-14
[CIR] Cast global var address to declared type at dtor call site

adams381:
A C++ global with a constexpr default constructor that fixes the active member of a union — `std::basic_string`'s SSO `__short` variant is a common example — has a `cir.global` whose stored record type is the narrowed shape of that active variant.  Classic CodeGen does the same (`@g = global { { { [16 x i8] } } } zeroinitializer`) and accepts the resulting `__cxa_atexit(@D1, @g, ...)` because LLVM IR uses opaque pointers.  CIR has typed pointers, so the `cir.call` registering the destructor for `__cxa_atexit` carries an operand type that doesn't match the dtor's `this` parameter.  This trips 16 libcxx tests and 71 cases total across libcxx, MultiSource, SingleSource, and SPEC in our build.

`verifyPointerTypeArgs(oldF, newF, userMap)` in `CIRGenModule::applyReplacements` (`clang/lib/CIR/CodeGen/CIRGenModule.cpp:1700`) catches this when ctor-dtor aliases are enabled and D1 is RAUW'd by D2.  Without aliases, the `cir.call` op verifier rejects the same operand-type mismatch directly.

The fix mirrors the cast pattern `emitGlobalVarDeclLValue` (`clang/lib/CIR/CodeGen/CIRGenExpr.cpp:441-445`) already uses for every AST-level reference to a global: bitcast the result of `getAddrOfGlobalVar` to `convertTypeForMem(type)` before any typed-pointer op consumes it.  `getAddrOfGlobalVar` itself stays raw so callers that walk to the underlying `GetGlobalOp` via `getDefiningOp()` keep working.

`global-dtor-union-narrowed.cpp` pins the CIR bitcast, the lowered LLVM helper-wrapped `__cxa_atexit`, and the equivalent OGCG direct `__cxa_atexit`.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenCXX.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenCXX.cpp
@@ -163,9 +163,22 @@ static void emitDeclDestroy(CIRGenFunction &cgf, const VarDecl *vd,
     // call right here.
     auto gd = GlobalDecl(dtor, Dtor_Complete);
     fnOp = cgm.getAddrAndTypeOfCXXStructor(gd).second;
+    // When a global has a constant initializer that fixes the active member
+    // of a union (e.g. an SSO short variant), CIR creates the global with
+    // the initializer's narrowed record type, so `getAddrOfGlobalVar` returns
+    // a pointer to the narrowed type rather than the variable's declared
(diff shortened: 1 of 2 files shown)


andykaylor (approved):
lgtm

PR #197828, opened 2026-05-14
[CIR] Emit cir.bool constant for non-odr-use of bool member

adams381:
`ScalarExprEmitter::VisitMemberExpr` has a constant-folding shortcut: when an expression evaluates as a compile-time integer, it bypasses the LValue load and emits the constant directly via `builder.getConstInt(loc, APSInt)`. That builder helper derives the CIR result type from the APSInt's bit width — for a `bool` member that is `!cir.int<u, 1>`, while the AST expression type is `bool` which CIR maps to `!cir.bool`.

Downstream ops that consume the value all reject the mismatch: `cir.if` and `cir.ternary` need `!cir.bool` operands, and `cir.call` rejects `!cir.int<u, 1>` against a `!cir.bool` parameter (which surfaces as the `emitCall: widening integer call argument` NYI in `CIRGenCall.cpp`). The fix dispatches on `e->getType()->isBooleanType()` and emits through `builder.getBool()` so the constant is typed `!cir.bool` from the start. Non-bool integral members keep the prior code path. The dispatch mirrors what `tryEmitPrivate`'s `APValue::Int` branch already does at `CIRGenExprConstant.cpp:1816-1817`.

In SPEC CPU 2026 this fires from `gcc/wide-int.h`'s `is_sign_extended` template-static-bool plumbing. Before the fix, four 721.gcc_r / 821.gcc_s TUs (`alias.cc`, `c-ada-spec.cc`, `c-pretty-print.cc`, `c-warn.cc`) hit a chain of widening NYIs and downstream `cir.if` / `cir.ternary` verifier errors; with the fix all four compile cleanly.

The new `clang/test/CIR/CodeGen/non-odr-use-const-bool.cpp` covers three downstream consumers — call argument, `if` condition, `&&` short-circuit — with CIR + LLVM + OGCG checks. The pre-existing `// TODO(cir): The classic codegen calls tryEmitAsConstant() here.` at the same site stays in place; this is the targeted fix in the meantime, and the broader refactor (extending `tryEmitAsConstant` to handle constant `VarDecl`s so the constant emitter takes over entirely) is the right follow-up.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenExprScalar.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExprScalar.cpp
@@ -2454,7 +2454,16 @@ mlir::Value ScalarExprEmitter::VisitMemberExpr(MemberExpr *e) {
   if (e->EvaluateAsInt(result, cgf.getContext(), Expr::SE_AllowSideEffects)) {
     llvm::APSInt value = result.Val.getInt();
     cgf.emitIgnoredExpr(e->getBase());
-    return builder.getConstInt(cgf.getLoc(e->getExprLoc()), value);
+    mlir::Location loc = cgf.getLoc(e->getExprLoc());
+    // The constant is folded from an APSInt with the source-type's bit width
+    // (1 for bool), but the AST's expression type is what later consumers of
+    // this value see. For a bool member we have to emit a !cir.bool constant
+    // -- otherwise downstream ops (cir.call into a bool parameter, cir.if /
+    // cir.ternary on the value, ...) would all reject the !cir.int<u, 1> the
+    // raw APSInt would produce.
+    if (e->getType()->isBooleanType())
+      return builder.getBool(value.getBoolValue(), loc);
+    return builder.getConstInt(loc, value);
   }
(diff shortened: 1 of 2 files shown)


xlauko (approved):
lgtm

clang/lib/CIR/CodeGen/CIRGenExprScalar.cpp
```diff
@@ -2454,7 +2454,16 @@ mlir::Value ScalarExprEmitter::VisitMemberExpr(MemberExpr *e) {
   if (e->EvaluateAsInt(result, cgf.getContext(), Expr::SE_AllowSideEffects)) {
     llvm::APSInt value = result.Val.getInt();
     cgf.emitIgnoredExpr(e->getBase());
-    return builder.getConstInt(cgf.getLoc(e->getExprLoc()), value);
+    mlir::Location loc = cgf.getLoc(e->getExprLoc());
+    // The constant is folded from an APSInt with the source-type's bit width
```
andykaylor, line 2458: This comment feels unnecessary.

andykaylor (approved):
lgtm

PR #195974, opened 2026-05-06
[clang-tidy] Add `llvm-formatv-string`

kastiglione:
Adds a clang-tidy check to perform some validation on `llvm::formatv` calls. Similar to the built in support Clang has for checking printf calls.

The validations are:
- The number of unique format indices matches the number of arguments.
- Every argument is used by the format string.
- Automatic and explicit indices are not mixed.

This includes a config option (`AdditionalFunctions`) to perform the same validation checks on other functions which take formatv inputs.

Assisted-by: claude

PR #197829, opened 2026-05-14
[clangd] Add llvm-formatv-string to TidyFastChecks

kastiglione:
Enables `llvm-formatv-string` (#195974) in clangd.


kastiglione:
I ran TidyFastChecks.py. In addition to adding this entry for `llvm-formatv-string`, it updated the values of many entries in TidyFastChecks.inc. Since my goal is to include this one new tidy check, I didn't want to update dozens and dozens of data points.

ArcsinX (commented):
By default  `TidyFastChecks.py` runs checks on `clang/lib/Sema/Sema.cpp` file.
But we have no `llvm::formatv` calls there, this can be the reason why `llvm-formatv-string` is reported as a fast check.

Can you please check `llvm-formatv-string` performance on a relatively big file with several `llvm::formatv` calls?  Maybe this one https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clangd/Quality.cpp ?
You can use `--source` argument of `TidyFastChecks.py` script to do this.

kastiglione:
@ArcsinX I have run TidyFastChecks on Quality.cpp and generally get an output of 1%, sometimes 0%, and once it was 3%.

ArcsinX (approved):
Thanks, LGTM

PR #197831, opened 2026-05-14
[flang] Add diagnostics for conflicting EXTERNAL/INTRINSIC and TARGET attributes

### LINKED ISSUE #197830: [flang] Missing diagnostics for conflicting EXTERNAL/INTRINSIC and TARGET attributes
Flang silently accepts declarations that specify both `EXTERNAL` and `TARGET` (or `INTRINSIC` and `TARGET`) attributes on the same entity, which violates the Fortran standard.

**Reproducer:**
```fortran
subroutine test1()
  real, target, external :: f1
end subroutine

subroutine test2()
  integer, target, intrinsic :: abs
end subroutine
```

**Expected behavior:** An error diagnostic should be emitted. An entity with the `EXTERNAL` or `INTRINSIC` attribute is a procedure, not a variable, and the `TARGET` attribute requires the entity to be a variable (F2018 §8.5.17).

**Actual behavior:** Flang accepts both cases without any diagnostic.

**Compiler Explorer repros** (all but Flang reject):
- Flang (trunk), https://godbolt.org/z/j6W4nnqT8, silently accepts
- gfortran (trunk), https://godbolt.org/z/5T69Ec6zx, `Error: TARGET attribute conflicts with EXTERNAL attribute`
- Intel ifx, https://godbolt.org/z/9hG1zo8qj, `error #6406: Conflicting attributes or multiple declaration of name.`
- nvfortran 26.3, https://godbolt.org/z/GYn1x91a6, `NVFORTRAN-S-0134-Illegal attribute - conflict with external`
- Cray Fortran (CCE), `ftn-425: Attributes EXTERNAL and TARGET must not appear in the same attribute list.`

MattPD:
Flang silently accepted declarations with both EXTERNAL and TARGET (or
INTRINSIC and TARGET) attributes on the same entity. This violates the
Fortran standard since TARGET requires a variable, but EXTERNAL/INTRINSIC
makes the entity a procedure.

Add CheckConflicting calls in CheckProcEntity to diagnose both cases.
All other major Fortran compilers already reject these combinations.

Fixes #197830

Assisted-by: Claude Opus 4.6.

### DIFF
--- a/flang/lib/Semantics/check-declarations.cpp
+++ b/flang/lib/Semantics/check-declarations.cpp
@@ -1466,6 +1466,10 @@ void CheckHelper::CheckArraySpec(
 void CheckHelper::CheckProcEntity(
     const Symbol &symbol, const ProcEntityDetails &details) {
   CheckSymbolType(symbol);
+  // F2018 8.5.17: an entity with the TARGET attribute shall be a variable;
+  // a procedure (EXTERNAL or INTRINSIC) is not a variable.
+  CheckConflicting(symbol, Attr::EXTERNAL, Attr::TARGET);
+  CheckConflicting(symbol, Attr::INTRINSIC, Attr::TARGET);
   const Symbol *interface{details.procInterface()};
   if (details.isDummy()) {
     if (!symbol.attrs().test(Attr::POINTER) && // C843
--- a/flang/test/Semantics/declarations09.f90
+++ b/flang/test/Semantics/declarations09.f90
@@ -1,0 +1,17 @@
+! RUN: %python %S/test_errors.py %s %flang_fc1
+! Check that EXTERNAL/INTRINSIC and TARGET attributes conflict
+
+subroutine test1()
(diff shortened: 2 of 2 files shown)


MattPD:
@klausler, @eugeneepshteyn: Hey, let me know if you can take a look; hopefully a small patch?

flang/lib/Semantics/check-declarations.cpp
```diff
@@ -1466,6 +1466,8 @@ void CheckHelper::CheckArraySpec(
 void CheckHelper::CheckProcEntity(
     const Symbol &symbol, const ProcEntityDetails &details) {
   CheckSymbolType(symbol);
+  CheckConflicting(symbol, Attr::EXTERNAL, Attr::TARGET);
```
eugeneepshteyn, line 1471: Let's add a comment with `F2018 8.5.17` for future reference.

eugeneepshteyn (approved):
One small request, otherwise LGTM.

eugeneepshteyn:
(I don't need to review this again.)

PR #197832, opened 2026-05-14
[ARM][MVE] Constant fold PREDICATE_CAST of 0 and 0xffff

davemgreen:
This allows us to fold away the vselect when we know that the condition is all true or all false.


statham-arm (approved):
LGTM. I see that the tests all seem to be constructing this situation on purpose; I guess there's also a case where it comes up by accident, without a user having deliberately written a pointless select?

davemgreen:
> LGTM. I see that the tests all seem to be constructing this situation on purpose; I guess there's also a case where it comes up by accident, without a user having deliberately written a pointless select?

Yeah - in the case I saw it recently it was from rather difficult leglization. Thanks for the review.

PR #197833, opened 2026-05-14
[CI] Remove duplicate entries in PROJECT_CHECK_TARGETS (NFC)

kastiglione:


.ci/compute_projects.py
kastiglione, line 152: lldb and lldb are above the fold, here

boomanaiden154 (approved):
LGTM.

PR #177301, opened 2026-01-22
[Flang][OpenMP][MLIR] Add attach and ref map type lowering to MLIR

agozillon:
This doesn't implement the functionality, just the relevant map type lowering to MLIR's omp.map.info. The more complicated changes to MapInfoFinalizationPass.cpp and OpenMPTOLLVMIRTranslation.cpp to support attach map and the various ref/attach semantics will come in a subsequent set of PRs. This just helps compartmentalize the changeset.

PR #197834, opened 2026-05-14
Revert "[Flang][OpenMP][MLIR] Add attach and ref map type lowering to MLIR"

agozillon:
Reverts llvm/llvm-project#177301


agozillon:
Applied revert as there was some merge issues with the later dependent PR pieces after merging the first pieces, so would result in breaking the build for longer than I'm comfortable with while I resolve them.

PR #197835, opened 2026-05-14
Revert "[Flang][MLIR][OpenMP] Add distinct var_ptr_ptr_type to omp.map.info operations & remove ref_ptr_ptee"

agozillon:
Reverts llvm/llvm-project#177302


agozillon:
Applied revert as there was some merge issues with the later dependent PR pieces after merging the first pieces, so would result in breaking the build for longer than I'm comfortable with while I resolve them.

PR #197836, opened 2026-05-14
[debugserver] Strip PAC bits from breakpoint/watchpoint addresses

bulbazord:
Normally, LLDB will strip the PAC bits from any address it sends to debugserver, so this is often redundant. However, this does not account for tests. Any test that sends packets to or reads packets from debugserver would need to know about addressable bits and how to strip them.

Fixing the tests one-by-one is a game of whackamole so I opted to change debugserver itself.

### DIFF
--- a/lldb/tools/debugserver/source/MacOSX/MachProcess.mm
+++ b/lldb/tools/debugserver/source/MacOSX/MachProcess.mm
@@ -2062,24 +2062,26 @@ void MachProcess::PrivateResume() {
 
 DNBBreakpoint *MachProcess::CreateBreakpoint(nub_addr_t addr, nub_size_t length,
                                              bool hardware) {
-  DNBLogThreadedIf(LOG_BREAKPOINTS, "MachProcess::CreateBreakpoint ( addr = "
-                                    "0x%8.8llx, length = %llu, hardware = %i)",
+  DNBLogThreadedIf(LOG_BREAKPOINTS,
+                   "MachProcess::CreateBreakpoint ( addr = "
+                   "0x%8.8llx, length = %llu, hardware = %i)",
                    (uint64_t)addr, (uint64_t)length, hardware);
 
-  DNBBreakpoint *bp = m_breakpoints.FindByAddress(addr);
+  nub_addr_t fixed_addr = DNBFixAddress(addr, ProcessID());
+  DNBBreakpoint *bp = m_breakpoints.FindByAddress(fixed_addr);
   if (bp)
     bp->Retain();
   else
-    bp = m_breakpoints.Add(addr, length, hardware);
+    bp = m_breakpoints.Add(fixed_addr, length, hardware);
 
-  if (EnableBreakpoint(addr)) {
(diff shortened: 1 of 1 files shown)


JDevlieghere:
> Any test that sends packets to or reads packets from debugserver would need to know about addressable bits and how to strip them.

The same argument can be made of debugserver. I'm surprised we already strip addresses in debugserver. Does it know about the number of addressable bits? What if I change the setting in LLDB, do we send a packet to inform debugserver? 

@jasonmolenda What's the reason this is necessary?

jasonmolenda:
> > Any test that sends packets to or reads packets from debugserver would need to know about addressable bits and how to strip them.
> 
> The same argument can be made of debugserver. I'm surprised we already strip addresses in debugserver. Does it know about the number of addressable bits? What if I change the setting in LLDB, do we send a packet to inform debugserver?
> 
> @jasonmolenda What's the reason this is necessary?

debugserver always had to strip bits down in DNBArchImplArm64.cpp because pc/fp/sp/lr were signed by the kernel when we thread_get_state them, and proper auth-and-clearing those bits could crash debugserver if there was a poison bit set because the inferior crashed.  So debugserver would simply clear those bits unconditionally.

A few weeks ago I hoisted this ArchImplArm64 method higher up into the DNB layer for my "expedite newly added binaries" feature.  debugserver needs to know the address of dyld's notification function, so it can tell that we've stopped there.  It's in the dyld_all_image_infos struct in the inferior, and it's a signed function pointer so I needed to clear the bits.  

All userland processes are using the same page table layout on Darwin, so they have the same number of bits available for pointer authentication.  debugserver is using a global sysctl to fetch the number of addressable bits, it will be the same for all user processes.

If I'm honest, I don't know if this change really feels like the right way to fix this, but I'm not clear on what tests are sending addresses with ptrauth bits mixed in.  Having debugserver clear non-addressable bits when given an address value is safe, we don't even need to honor MTE when we are reading/writing memory through the mach VM calls, but it seems like the debugger on the other end should have enough environment knowledge to fix the addresses being sent in.

bulbazord:
> If I'm honest, I don't know if this change really feels like the right way to fix this, but I'm not clear on what tests are sending addresses with ptrauth bits mixed in.

A few of the tests in `test/API/tools/lldb-server/` are doing this. It roughly works like this:
1. Inferior writes a function pointer to stdout via `printf`.
2. The test picks up the address and sends a breakpoint packet to debugserver manually.
3. debugserver places the breakpoint on the address.

On arm64e, the function pointer is signed. Nothing in this chain handles stripping the PAC bits. Normally, LLDB is the one that strips the bits. If we wanted to the tests to follow the same model, either the inferior would strip the bits before printing to stdout or the test would strip the bits before sending. I changed debugserver so that everything Just Works (including future tests that we add that may do this). I don't feel strongly about any particular solution.

bulbazord:
I fixed the tests without this change.
https://github.com/llvm/llvm-project/pull/203988
https://github.com/llvm/llvm-project/pull/203995

PR #197838, opened 2026-05-14
[NVPTXAsmPrinter] Allow self-referential device global initializers

aryanmagoon:
This change adds support for cyclic device global initializers in NVPTX, including self-references and references between distinct globals. The backend’s global-emission ordering walk previously treated every cycle as an error and aborted with “Circular dependency found in global variable set.” As a result, valid patterns such as __device__ void *arr[] = {(void *)"foo", &arr[0]}; were rejected. The NVPTX asm printer now models initializer dependencies as a graph and emits its strongly connected components in dependency-first order. Acyclic components are emitted normally. For cyclic components, it first emits compatible PTX .extern declarations and then emits the definitions in a valid order. Cycles that cannot be represented this way, such as cycles containing only static globals, are still rejected. The forward declarations reuse the normal definition emitter so their address spaces, alignments, types, and aggregate layouts match the definitions. Added new tests to cover self-references, cycles between distinct globals, mixed linkage, packed aggregates, dependency ordering, and unrepresentable static-only cycles.


llvm/test/CodeGen/NVPTX/global-cycle.ll
```diff
@@ -0,0 +1,9 @@
+; RUN: not --crash llc < %s -mtriple=nvptx64 -mcpu=sm_20 2>&1 | FileCheck %s
+
+; Self-references are allowed, but a real dependency cycle between distinct
+; globals must still be rejected by the emission-order walk.
```
Artem-B: Is that a fundamental restriction (e.g. we'll fail later, in ptxas), or an implementation gap in LLVM.

If we allow `@b = addrspace(1) global ptr addrspace(1) @b`, then we allow
```
@a = addrspace(1) global ptr addrspace(1) @b
@b = addrspace(1) global ptr addrspace(1) @b
```

Then we should allow 
```
@a = addrspace(1) global ptr addrspace(1) @b
@b = addrspace(1) global ptr addrspace(1) @a
```

For `@a` it does not matter what's in `@b` as long as `@b` exists, and for `@b` I do not see a fundamental difference between @b being initialized with a pointer to itself or to `@a` -- both exist.

What am I missing?

aryanmagoon: The comment was too strong. `@a = @b` only needs the symbol `@b` to be available, not `@b`'s initializer. So a cycle between distinct globals isn't inherently different. I will update the comment to reflect this, but I think supporting mutual references like this should be handled as a follow-up to this.

Artem-B: > I think supporting mutual references like this should be handled as a follow-up to this.

OK. Then add it as a TODO in the comment.

llvm/test/CodeGen/NVPTX/packed-aggr.ll
```diff
@@ -93,3 +93,16 @@ declare void @func()
...
+
+%t6 = type <{ ptr, i8 }>
+@self_packed = addrspace(1) global %t6 <{
+; CHECK32: .global .align 1 .u8 self_packed[5] = {
+; CHECK32-SAME: 0xFF(generic(self_packed)+3), 0xFF00(generic(self_packed)+3), 0xFF0000(generic(self_packed)+3), 0xFF000000(generic(self_packed)+3), 7};
+; CHECK64: .global .align 1 .u8 self_packed[9] = {
+; CHECK64-SAME: 0xFF(generic(self_packed)+3), 0xFF00(generic(self_packed)+3), 0xFF0000(generic(self_packed)+3), 0xFF000000(generic(self_packed)+3),
+; CHECK64-SAME: 0xFF00000000(generic(self_packed)+3), 0xFF0000000000(generic(self_packed)+3), 0xFF000000000000(generic(self_packed)+3), 0xFF00000000000000(generic(self_packed)+3), 7};
+  ptr addrspacecast (ptr addrspace(1) getelementptr (i8, ptr addrspace(1) @self_packed, i32 3) to ptr),
+  i8 7 }>, align 1
```
Artem-B: Do the self-reference tests pass the ptxas test runs?  it may be useful to check generated PTX with older versions of ptxas to make sure ptxas can compile what we generate. If older versions of ptxas have a problem with those circular references, we may want to produce an error in LLVM if we attempt to emit circular reference which can't be compiled.

aryanmagoon: Instead of producing an error in LLVM, do you think we could use strongly connected components to forward-declare globals when needed? This would also solve the issue of the mutual global references you brought up in your other comment? Maybe we should do it now if it solves both things.

aryanmagoon: @Artem-B  The rough idea would be that the nodes of a graph are each module GlobalVariable, and there is a directed edge G -> D when G's initializer mentions D. Then we compute SCCs over this graph. The SCC graph is acyclic so now we can emit SCCs in dependency-first order. For an acyclic SCC, we emit the global normally, for a cyclic SCC (which would include self-reference case) we can emit a legal PTX forward declarations for the globals in that SCC first. Cycles that can't be made legal this way, let's say we had a static-only cycle, we can still reject this.

aryanmagoon:
Ping

llvm/test/CodeGen/NVPTX/packed-aggr.ll
Artem-B: I believe my original question on whether ptxas actually accepts what we generate has not been answered yet. Please confirm that global-cycle.ll test works with ptxas enabled. E.g. `LLVM_PTXAS_EXECUTABLE=$HOME/local/cuda-12.9.1/bin/ptxas ninja check-llvm`

Artem-B:
It's hard to see what's changed since the last review round. I see that the branch was force-updated, but github says that "There isn’t anything to compare. e2da8b7e0120b2e17ff849c3f0156e77e389a1ff and e13267f64c6fbfe7392cab0b7de9f0e9162f00b5 are identical."

For the changes under review it's very useful to add new changes as a separate commit, so reviewers can track the progress. Right now some of my comments are completely disconnected from the source they were referring to, so it's hard to tell what we were talking about and whether that issue was addressed.

aryanmagoon:
> It's hard to see what's changed since the last review round. I see that the branch was force-updated, but github says that "There isn’t anything to compare. [e2da8b7](https://github.com/llvm/llvm-project/commit/e2da8b7e0120b2e17ff849c3f0156e77e389a1ff) and [e13267f](https://github.com/llvm/llvm-project/commit/e13267f64c6fbfe7392cab0b7de9f0e9162f00b5) are identical."
> 
> For the changes under review it's very useful to add new changes as a separate commit, so reviewers can track the progress. Right now some of my comments are completely disconnected from the source they were referring to, so it's hard to tell what we were talking about and whether that issue was addressed.

You are right. I am sorry that I force-updated and squashed the branch. I will keep further review changes in separate commits. For clarity, the previous backend used a recursive dependency walk. It marked a global variable as active and then visited every global variable that its initializer referenced. It emitted the referenced variables first. If the walk found an active variable again, it reported a circular dependency. My first fix removed a global variable’s reference to itself from its dependency list. This prevented the walk from identifying a self-reference as a cycle. LLVM then emitted the self-referencing definition without a prior declaration. CUDA 13.4 and newer ptxas versions accept this form. Older versions reject it because they require a declaration before the reference. This fix also did not support a cycle between two or more different global variables. I replaced that fix with an implementation based on strongly connected components. The new implementation supports both self-references and cycles between different global variables. It emits compatible declarations before the definitions when a cycle requires them. It still reports an error when declarations cannot make the cycle valid, such as a static-only cycle.

llvm/test/CodeGen/NVPTX/packed-aggr.ll
aryanmagoon: Yes, global-cycle.ll works with ptxas enabled. I tested the generated PTX with CUDA 12.9 ptxas, 13.0, 13.1, 13.2, 13.3, and the newly released 13.4, global-cycle.ll passed in both 32-bit and 64-bit modes with all these versions. I also tested the self-reference case separately in both modes. CUDA 13.3 and earlier reject a direct self-reference if the symbol does not have a prior declaration. The new SCC implementation emits a compatible .extern declaration before the definition, so these versions accept the generated PTX. CUDA 13.4 accepts the direct self-reference, but they also accept the additional declaration. For CUDA 13.4 I compared CUBIN files generated with and without the declaration. The CUBIN files were byte-for-byte identical. The declaration only makes the PTX text slightly larger. It does not create another symbol or allocation in the CUBIN. Therefore, LLVM does not need to reject representable cycles. It can emit declarations for cyclic SCCs. LLVM still reports an error for cycles that cannot be made valid with a declaration.

Artem-B: > LLVM still reports an error for cycles that cannot be made valid with a declaration.

We should have that captured as a test case. Right now we seem to test only working cases.

llvm/test/CodeGen/NVPTX/global-cycle.ll
Artem-B: I believe this is resolved in the latest version, which already supports reference cycles.

Artem-B:
> You are right. I am sorry that I force-updated and squashed the branch. I will keep further review changes in separate commits.

It's all fixable. You can always reset to the old revision, then rebase or cherry-pick your latest commit on top of it, and then edit the description. The resulting commit should have only new changes, and the you can force-update the branch again. It's just a suggestion, no need to do it to the PR right now if it sounds too complicated.

llvm/test/CodeGen/NVPTX/packed-aggr.ll
Artem-B: LLM suggestion for more testing:

> Suggested additions for completeness:
> - N-node cycles (N > 2): Add a 3-node cycle test case (@a -> @b -> @c -> @a) to test multi-step cycle resolution in orderDefinitionsInSCC.
> - Multiple disjoint cycles: Add a test with two independent cycles in a single module to verify SyntheticRoot behavior across distinct cyclic SCCs.
> - GlobalAlias in cycle: Add a test where a cycle passes through a GlobalAlias to explicitly exercise alias traversal in discoverDependentGlobals.

Artem-B: I see a negative test in [global-cycle-internal.ll](https://github.com/llvm/llvm-project/pull/197838/changes#diff-e55a55a96acd61d4c582cfdd90847fa37c6d1f791566ce806fed3fb1681f9796). Are there any other failure cases we can't handle? E.g. do internal symbols in a larger loop cause trouble?

aryanmagoon: Yes, internal symbols in a larger cycle work as long as every cycle contains at least one symbol that PTX can legally forward-declare. I updated `orderDefinitionsInSCC` to treat forward-declared nodes as satisfied and then topologically order the remaining definitions; it reports an error only when an internal-only subcycle remains, because PTX has no legal way to declare that subcycle. I added a mixed-linkage three-node cycle and a negative larger SCC containing an internal-only subcycle to cover both cases. While expanding the tests, I also found false dependency edges through aliases, functions, and symbolic GEP indices, so I changed `discoverDependentGlobals` to treat non-variable GlobalValues as leaf symbols and to follow only the base pointer of a GEP. I also added tests that includes disconnected cycles, constant globals, aliases, functions, and symbolic GEP indices. All pass with older and newer ptxas versions in both modes.

llvm/test/CodeGen/NVPTX/global-cycle-alias.ll
```diff
@@ -0,0 +1,26 @@
...
+; @alias_cycle_b a nested operand of the GlobalAlias. The PTX initializer for
+; @alias_cycle_a refers only to the alias symbol, so the alias's operands must
+; not create the false dependency cycle
+; @alias_cycle_a -> @cycle_alias -> @alias_cycle_b -> @alias_cycle_a.
+
+; CHECK-NOT:  .extern .global
+; CHECK:      .global .align 8 .u64 alias_cycle_a = cycle_alias;
+; CHECK-NEXT: .global .align 8 .u64 alias_cycle_b = alias_cycle_a;
+; CHECK-NOT:  .extern .global
+; CHECK:      .alias cycle_alias, alias_target;
```
Artem-B: Perhaps we should just autogenerate these checks with utils/update_llc_test_checks.py and capture the generated directives exactly, if the complete PTX is not too large.

Artem-B (approved):
LGTM overall.

llvm/lib/Target/NVPTX/NVPTXAsmPrinter.cpp
```diff
@@ -138,51 +142,168 @@ static void emitInitialRawDwarfLocDirective(const MachineFunction &MF,
...
+        Ready.emplace(Dependent->ModuleOrder, Dependent);
+    }
+  }
 
-  // Now we can visit ourself
-  Order.push_back(GV);
-  Visited.insert(GV);
-  Visiting.erase(GV);
+  if (Order.size() != SCC.size())
+    report_fatal_error("Circular dependency found in global variable set");
```
Prince781, line 305: Can we reference the variable(s) in the error?

Prince781 (changes requested):
Looks good. Two extra things:

1. Can we have a little more comments inside the code?
2. We might want a test for `managed` as well:

```diff
diff --git a/llvm/test/CodeGen/NVPTX/managed.ll b/llvm/test/CodeGen/NVPTX/managed.ll
index 931c17d5ba80..7a246b7ccbed 100644
--- a/llvm/test/CodeGen/NVPTX/managed.ll
+++ b/llvm/test/CodeGen/NVPTX/managed.ll
@@ -14,6 +14,12 @@
 ; CHECK: .extern .global .attribute(.managed) .align 8 .b64 managed_decl_g;
 @managed_decl_g = external addrspace(1) global ptr, align 8
 
-!nvvm.annotations = !{!0, !1}
+; A cyclic managed definition needs an exactly matching managed declaration.
+; CHECK:      .extern .global .attribute(.managed) .align 8 .u64 managed_self;
+; CHECK-NEXT: .visible .global .attribute(.managed) .align 8 .u64 managed_self = managed_self;
+@managed_self = addrspace(1) global ptr addrspace(1) @managed_self, align 8
+
+!nvvm.annotations = !{!0, !1, !2}
 !0 = !{ptr addrspace(1) @managed_g, !"managed", i32 1}
 !1 = !{ptr addrspace(1) @managed_decl_g, !"managed", i32 1}
+!2 = !{ptr addrspace(1) @managed_self, !"managed", i32 1}
```

llvm/lib/Target/NVPTX/NVPTXAsmPrinter.cpp
aryanmagoon: At this point, we only know that the topological ordering did not consume the entire SCC, so we don't have a specific offending variable available. If we wanted to report one, we would have to scan the SCC again. We could retain the additional state solely for the error message during the first traversal, but I don't think that's appropriate. I would prefer to keep this error message.

Prince781 (approved):

PR #197843, opened 2026-05-14
gn build: Support building debug binaries from a release build directory.

pcc:
This change makes it so that all binaries in `bin` have a corresponding
debug binary target `$TARGET.dbg` that will build a debug binary in
`bin.dbg/$TARGET`. This is implemented using a separate toolchain for
the debug builds. This has the following advantages over a separate
debug build directory:

1. Can use release mode tblgen to generate code for the debug build
   binaries, which speeds up the build because it can run tblgen faster as
   well as being able to reuse the tblgen binary from the release build.

2. Can reuse runtimes built in the release mode clang's resource
   directory, avoiding the need to build them again with a slow debug
   build of clang.

Assisted-by: gemini


pcc:
Ping

aeubanks:
the functionality itself looks good, but I really worry that this adds a lot more indirection to the gn files that makes it harder to debug and maintain

nico:
I'm ambivalent about this (i.e. I don't hate it more than I like the functionality ;)). I'll let @aeubanks make a call here :)

aeubanks (approved):
lgtm

nico:
This breaks building for me (macOS):

```
% time ninja -C out/gn clang                                 
ninja: Entering directory `out/gn'
[0/1] Regenerating ninja files
ERROR Unresolved dependencies.
//clang/tools/driver:clang(//llvm/utils/gn/build/toolchain:host_debug)
  needs //libcxx/include:copy_headers(//llvm/utils/gn/build/toolchain:host_debug)

FAILED: build.ninja.stamp 
../../llvm/utils/gn/bin/mac-arm64/gn --root=../.. -q --dotfile=../../llvm/utils/gn/.gn --regeneration gen .
```

nico:
Prepared a revert here: #211294

Will land it the next time this stops me from doing something. (Today was the 2nd time; my comment was the first time.)

PR #197844, opened 2026-05-14
[lldb][NFC] Socket::Send should return a signed type

bulbazord:
The underlying `::send` and `::sendto` returns `ssize_t` on macOS/Linux and `int` on Windows. These functions also commmunicate an error through returning -1 (which is not expressible in a size_t).

### DIFF
--- a/lldb/include/lldb/Host/Socket.h
+++ b/lldb/include/lldb/Host/Socket.h
@@ -173,7 +173,7 @@ public:
 protected:
   Socket(SocketProtocol protocol, bool should_close);
 
-  virtual size_t Send(const void *buf, const size_t num_bytes);
+  virtual ssize_t Send(const void *buf, const size_t num_bytes);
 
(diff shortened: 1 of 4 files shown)


labath (approved):
In this setup, that makes sense, though I am somewhat surprised that the functions do not return Status/Error/Expected like Connect and Listen. Expected<size_t> would be fair game as negative values are only used to communicate errors. (although there are schools of programming that say all types should be signed, I don't think we subscribe to those).

charles-zablit (approved):

bulbazord:
> In this setup, that makes sense, though I am somewhat surprised that the functions do not return Status/Error/Expected like Connect and Listen. Expected<size_t> would be fair game as negative values are only used to communicate errors. (although there are schools of programming that say all types should be signed, I don't think we subscribe to those).

I considered this (and may head in this direction later) but Send is internal to Socket. Right now, it appears to only be used from Socket::Write (which does return a Status object).

PR #197845, opened 2026-05-14
[flang] Canonicalize sliced array access by fir.array_coor.

vzakhari:
This patch adds a canonicalization pattern for pulling rank-reducing
slices into `fir.array_coor`. This is helpful to preserve the original
rank of the array in `fir.array_coor`, which then helps representing
slice accesses in memref dialect as accesses to the original array.
This way, further conversion to affine dialect has benefits of
exposing the math applied to indices of all array dimensions.

For example, if a non-scalar slice depends on a loop IV, pulling
this into the original array access allows building access maps
enabling affine dependency analysis.

I considered the following three models for the addressing used
in `fir.array_coor` with rank reducing slices:

1) Mixed model:
  - Encoding
    * Scalar-sliced dims: index is slice_lb
    * Non-scalar dims: index is section-style (1-based)
  - Pros
    * Minimal canonicalization math
    * Keeps scalar selection explicit in the index list
  - Cons
    * Semantics are mixed in one array_coor
    * Harder to reason about/debug
    * Requires special-case logic in later lowering (FIRToMemRef / CodeGen)
      to interpret non-scalar dims differently from scalar dims

2) Uniform section-style model (all 1-based):
  - Encoding
    * Scalar-sliced dims: index is 1
    * Non-scalar dims: index is section-style (1-based)
    * Scalar selection recovered only from slice_lb
  - Pros
    * Conceptually clean and consistent (single index convention)
    * Easier mental model
  - Cons
    * Needs coordinated changes in both lowerings
    * Lowerings must always reconstruct scalar selected coordinates
      from slice metadata
    * Higher migration risk unless all paths are updated together

3) CodeGen-compatible coordinate-space model (the implemented one):
  - Encoding
    * Scalar-sliced dims: index is slice_lb
    * Non-scalar dims: index converted to coordinate-space (lb + idx - 1)
  - Pros
    * No behavioral changes needed in CodeGen logic
    * Aligns canonicalized IR with existing coordinate-space assumptions
    * Avoids extra CodeGen compatibility branches
  - Cons
    * Canonicalization is more complex (must synthesize index transforms)
    * FIRToMemRef still needs alignment for full-rank vs reduced-rank
      index consumption
    * Slightly less "visually uniform" than all-1-based

Assisted by Codex.


clementval (approved):
I had a look through it and it looks good. Might be good to have another pair of eyes on it.

vzakhari:
Thank you, Valentin! I will wait for Jean's review, and I am still waiting for performance testing to complete.

SusanTan (approved):
thank you Slava!

vzakhari:
Correctness and performance testing is clean.

flang/lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -570,12 +570,23 @@ struct SimplifyArrayCoorOp : public mlir::OpRewritePattern<fir::ArrayCoorOp> {
...
+    // defining producer for the embox/rebox lookup below. This exposes a
+    // sliced embox that is otherwise hidden by the box_addr indirection,
+    // which is necessary for later lowering (FIRToMemRef) to see the
+    // full-rank access pattern.
+    mlir::Operation *defOp = memref.getDefiningOp();
+    if (!mlir::isa<fir::BaseBoxType>(memref.getType())) {
+      auto boxAddrOp = mlir::dyn_cast_or_null<fir::BoxAddrOp>(defOp);
+      if (!boxAddrOp)
+        return mlir::failure();
+      defOp = boxAddrOp.getVal().getDefiningOp();
```
jeanPerier, line 597: BoxAddrOp (fir.box<T1> -> fir.ref<T2>) technically does not require its `T1` and `T2` to match (i.e., it may contain a cast on top of the extraction). It is not commonly done, but happens. Would this be a problem? If so, you may want to guard against that here.

jeanPerier (approved):
Looks great, just a minor question

flang/lib/Optimizer/Dialect/FIROps.cpp
vzakhari: Thanks for pointing this out, Jean! I will handle it.

vzakhari: @jeanPerier can you please take a look at the added test in commit 3? Is this a mismatch that may happen or there are more cases? The one I added seems to be benign, but I restricted the matching for this case anyway.

jeanPerier: Yes, that is what I had in mind, I was worried about cases where we project some unranked array or scalar box into a contiguous ranked array via the box_addr. Thanks!

jeanPerier (approved):

PR #197849, opened 2026-05-15
[HLSL][SPIRV] Correct reversebit 64 split

### LINKED ISSUE #197810: [HLSL][SPIRV] Miscompile in lowering of bitreverse for vectors of 64 bit elements
The SPIR-V lowering of `llvm.bitreverse` swizzles the upper and lower halves of adjacent elements with their individual bits reversed together instead of reversing the bits.

Consider this HLSL:
```hlsl
StructuredBuffer<uint64_t2> In : register(t0);
RWStructuredBuffer<uint64_t2> Out : register(u1);

[numthreads(1,1,1)]
void main() {
  Out[0] = reversebits(In[0]);
}
```
Compiler Explorer: https://hlsl.godbolt.org/z/MP69EbvWK

The result vector is constructed like so:
```llvm
        %45 = OpVectorShuffle %23 %44 %44 1 3
        %46 = OpVectorShuffle %23 %44 %44 0 2
        %47 = OpCompositeConstruct %22 %45 %46
```

We're assembling the top halves of two `i64`s (elements 1 and 3) and and the bottom halves (elements 0 and 2), rather than the contiguous elements. So for example, if the input were `[0xFFFFFFFF00000000, 0xAAAAAAAAAAAAAAAA]` we end up with `[0x55555555FFFFFFFF, 0x5555555500000000]` instead of `[0xFFFFFFFF00000000, 0x5555555555555555]`.

This causes [reversebits.64.test](https://github.com/llvm/offload-test-suite/blob/main/test/Feature/HLSLLib/reversebits.64.test) to fail on Vulkan.

Unfortunately, the tests for this operation simply tests for the incorrect result, matching the implementation: https://github.com/llvm/llvm-project/blob/c3df8f8c83372a53f6949c585d1a5a03b775332c/llvm/test/CodeGen/SPIRV/hlsl-intrinsics/reversebits.ll#L73-L74

joaosaffran:
Replace the split-lanes + OpCompositeConstruct sequence in selectBitreverse64 with a single OpVectorShuffle to swap the high/low 32-bit halves, simplifying both the selector logic, the emitted SPIR-V and fixing the bug.

fix: #197810 
Assisted by: Claude Opus 4.6

### DIFF
--- a/llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
+++ b/llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
@@ -3685,19 +3685,21 @@ bool SPIRVInstructionSelector::selectBitreverse64(Register ResVReg,
   if (!selectBitreverseNative(Reverse32, VecI32Type, I, Vec32))
     return false;
 
-  // Splits result into highbit lane and lowbit lane
-  auto MaybeParts = splitEvenOddLanes(Reverse32, ComponentCount, I, I32Type);
-  if (!MaybeParts)
-    return false;
-  SplitParts &Parts = *MaybeParts;
-
   // Reversing a 64-bit value = reverse each 32-bit half AND swap them,
   // so the old High word becomes lane 0 (low) and old Low becomes lane 1
   // (high).
   Register SwappedVec = MRI->createVirtualRegister(GR.getRegClass(VecI32Type));
-  if (!selectOpWithSrcs(SwappedVec, VecI32Type, I, {Parts.High, Parts.Low},
-                        SPIRV::OpCompositeConstruct))
-    return false;
+  auto MIB = BuildMI(*I.getParent(), I, I.getDebugLoc(),
+                     TII.get(SPIRV::OpVectorShuffle))
+                 .addDef(SwappedVec)
+                 .addUse(GR.getSPIRVTypeID(VecI32Type))
+                 .addUse(Reverse32)
+                 .addUse(Reverse32);
+  for (unsigned J = 0; J < ComponentCount; ++J) {
+    MIB.addImm(2 * J + 1);
+    MIB.addImm(2 * J);
+  }
+  MIB.constrainAllUses(TII, TRI, RBI);
 
   // Groups 32 bit vector back to 64 bit scalar.
   return selectOpWithSrcs(ResVReg, ResType, I, {SwappedVec}, SPIRV::OpBitcast);
--- a/llvm/test/CodeGen/SPIRV/hlsl-intrinsics/reversebits.ll
+++ b/llvm/test/CodeGen/SPIRV/hlsl-intrinsics/reversebits.ll
@@ -15,8 +15,6 @@
 ;CHECK-DAG: %[[#vec_int_16:]] = OpTypeVector %[[#int_16]] 2
 
 ;CHECK-DAG: %[[#const_16:]] = OpConstant %[[#int_32]] 16
-;CHECK-DAG: %[[#const_zero:]] = OpConstant %[[#int_32]] 0
-;CHECK-DAG: %[[#const_one:]] = OpConstant %[[#int_32]] 1
 ;CHECK-DAG: %[[#const_two:]] = OpConstant %[[#int_64]] 2
 ;CHECK-DAG: %[[#composite:]] = OpConstantComposite %[[#vec_int_32]] %[[#const_16]] %[[#const_16]] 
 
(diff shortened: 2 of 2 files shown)


Icohedron (approved):

llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
```diff
@@ -3685,19 +3685,21 @@ bool SPIRVInstructionSelector::selectBitreverse64(Register ResVReg,
...
+  auto MIB = BuildMI(*I.getParent(), I, I.getDebugLoc(),
+                     TII.get(SPIRV::OpVectorShuffle))
+                 .addDef(SwappedVec)
+                 .addUse(GR.getSPIRVTypeID(VecI32Type))
+                 .addUse(Reverse32)
+                 .addUse(Reverse32);
+  for (unsigned J = 0; J < ComponentCount; ++J) {
+    MIB.addImm(2 * J + 1);
+    MIB.addImm(2 * J);
+  }
```
bogner, line 3701: ComponentCount is always either 1 or 2 at this point, because of the recursive `handle64BitOverflow` case above.

llvm/test/CodeGen/SPIRV/hlsl-intrinsics/reversebits.ll
```diff
@@ -82,9 +78,7 @@ entry:
 ; CHECK: %[[#param:]] = OpFunctionParameter %[[#vec_int_64]]
 ; CHECK: %[[#vec_cast:]] = OpBitcast %[[#vec_int_32x4]] %[[#param]]
 ; CHECK: %[[#reverse:]] = OpBitReverse %[[#vec_int_32x4]] %[[#vec_cast]]
-; CHECK: %[[#low:]] = OpVectorShuffle %[[#vec_int_32]] %[[#reverse]] %[[#reverse]] 1 3
-; CHECK: %[[#high:]] = OpVectorShuffle %[[#vec_int_32]] %[[#reverse]] %[[#reverse]] 0 2
-; CHECK: %[[#vec_rebuild:]] = OpCompositeConstruct %[[#vec_int_32x4]] %[[#low]] %[[#high]]
+; CHECK: %[[#vec_rebuild:]] = OpVectorShuffle %[[#vec_int_32x4]] %[[#reverse]] %[[#reverse]] 1 0 3 2
```
bogner, line 81: This doesn't look right - 1 0 3 2 isn't reversed. Have you tested that this actually fixes [the reversebits.64 test in the offload suite](https://github.com/llvm/offload-test-suite/blob/main/test/Feature/HLSLLib/reversebits.64.test)?

llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
joaosaffran: It is not guaranteed that we will have only 1 or 2 components here; I can have a `OpBitReverse <3 x i64>` call to be present in this point `handle64BitOverflow` handles such scenarios by splitting `OpBitReverse <3 x i64>` into 2 `OpBitReverse` calls, `OpBitReverse i64` and `OpBitReverse <2 x i64>`. This is required because `OpBitReverse` doesn't support 64 bit operands.

llvm/test/CodeGen/SPIRV/hlsl-intrinsics/reversebits.ll
joaosaffran: Spent some time triple-checking this; yes, `reversebits.64 test` is passing, with this fix.

To clarify why 1 0 3 2 is correct: when lowering, for example, `OpBitReverse <2 x i64>`, we cast to `<4 x i32>` since OpBitReverse doesn't support 64-bit operands. 

In the resulting vector, elements 0 and 1 are the low and high 32-bit halves of the first i64, and elements 2 and 3 are the low and high halves of the second i64.

After OpBitReverse reverses the bits within each 32-bit chunk, we still need to swap the high and low halves within each 64-bit element to complete the full 64-bit reversal. That's what 1 0 3 2 does: it swaps 0 and 1 (the two halves of the first i64) and 2 and 3 (the two halves of the second i64).

Let me know if that clears things up!

llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
bogner: I think you’re agreeing with me? At this point in the function where we’re doing this loop the ComponentCount cannot be greater than 2, or we would have called handle64BitOverload and recursed into this function with a ComponentCount of at most 2

llvm/test/CodeGen/SPIRV/hlsl-intrinsics/reversebits.ll
bogner: Ah, fair point. Thanks for clarifying.

bogner (approved):

llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
bogner: Anyways this looks fine, not sure what I was thinking here

bogner:
I don't see a corresponding PR in https://github.com/llvm/offload-test-suite for the `Feature/HLSLLib/reversebits.64.test` that now XPASSes. Are you planning on putting that up or do you need someone else to?

PR #195744, opened 2026-05-04
[SSAUpdater][NFC] Limit search for matching phi-node

LU-JOHN:
Limit search for a matching phi-node in `FindExistingPhi` to avoid O(N^2) behavior when a large number of phi-nodes needs to be searched.  This limit is motivated by reduce.cu in hipcub/warp where N was 26000+.   This large number of phi-nodes is currently avoided by https://github.com/llvm/llvm-project/pull/190269.  In a 3-stage build of LLVM the max search was 53.  

This PR was tested with a 3-stage build of LLVM and ~10000 csmith testcases with a search limit set to 0 (in this PR the search limit is set to 80).  Thus, missing a matching phi-node does not seem to cause problems.

Since the search limit is rarely hit, this PR should almost always be a NFC.

This PR was made with Claude assistance.

PR #135079, opened 2025-04-09
[SimplifyCFG] Extend jump-threading to allow live local defs

LU-JOHN:
Extend jump-threading to allow local defs that are live outside of the threaded block.  Allow threading to destinations where the local defs are not live.

PR #190269, opened 2026-04-02
Revert "[SimplifyCFG] Extend jump-threading to allow live local defs …

hjagasiaAMD:
…(#135079)"

This reverts commit a757f23404c594f4a48b4ddb6625f88b349d11d5. Commit causes reduce.cu file in hipcub/warp go from 2 minutes of compilation to taking several hours.

PR #197850, opened 2026-05-15
Reapply [SimplifyCFG] Extend jump-threading to allow live local defs

LU-JOHN:
Restore "Extend jump-threading to allow live local  defs" #135079.  Long compilation time with reduce.cu in hipcub/warp was partially addressed in #195744. Compilation time for reduce.cu with this PR (after #195744) is 6 minutes 40 seconds.  Without (#195744) compilation time was several hours.

Long compilation time in reduce.cu was only exposed by jump-threading.  In my view the primary causes were due to inlining, SROA tripling the IR code size, and SSA updating 26K phi-nodes resulting in an O(N^2) search for duplicates.  #195744 limits phi search times.

This reverts commit a76750e6de6aba2223097dc505578556ec245d50.


hjagasiaAMD:
Patch wise LGTM; unchanged from previously approved PR #135079. But looks like in PR #190269, reduce.cu file in hipcub/warp was compiling in 2 minutes on gfx942 without this change. Do you know why it is 6 minutes 40 seconds now? May be good to check and make sure we are not regressing between TheRock releases. Also may be good to add a SimplifyCFG test that demonstrates the hipcub/warp interaction end‑to‑end.

LU-JOHN:
>  Do you know why it is 6 minutes 40 seconds now? 

The steps that led to a multi-hour compile time were:

1. Inlining
2. SimplifyCFG did more jump-threading.  Not much time is spent in this pass and it did not increase the code size significantly.  But it did change the CFG so that:
3.  SROA transforms allocas into scalars.  This tripled the code size.
4. The larger code size led to much more time spent in StructurizeCFG. StructurizeCFG used SSAUpdater to do linear searches to update 26K+ phi-nodes., which used O(N^2) time.

Thus, I would say that SimplifyCFG is not the cause for the increase in compile-time (Inlining, SROA, and StructurizeCFG/SSAUpdater are the main reasons).  #195744 dealt with reason #4, which allows this PR to be reapplied.

After #195744 and this PR, the 6+ minute time in compiling reduce.cu is spent in:
>   ( 36.3%)   Register Coalescer
>   ( 11.8%)   Eliminate PHI nodes for register allocation
>   (  6.3%)    Optimize machine instruction PHIs
>   (  5.8%)    Structurize control flow
>   (  4.2%)    AMDGPU DAG->DAG Pattern Instruction Selection
>   ...
>   (  0.8%)    SROAPass
>   ...
>   (  0.2%)    SimplifyCFGPass

LU-JOHN:
> Also may be good to add a SimplifyCFG test that demonstrates the hipcub/warp interaction end‑to‑end.

https://github.com/llvm/llvm-project/pull/195744 added a testcase that shows the O(N^2) behavior with 400+ phi-nodes.  I have not been able to make a small testcase that shows the full interaction (including inlining, SROA, and StructurizeCFG).    The current size for the llc testcase showing the full interaction is 87M.

hjagasiaAMD (approved):

PR #197851, opened 2026-05-15
[lldb] Use os_unfair_lock for the string pool on Darwin

JDevlieghere:
The  ConstString pool guards each shard with a SmartRWMutex<false>, which on Darwin maps to pthread_rwlock and incurs heavy syscall overhead under contention. os_unfair_lock is a userspace fast path that's significantly faster for the write path and roughly on par for reads, but it's exclusive-only and not available off Apple platforms.

On a 32-thread machine, hyperfine reports the benchmark unit test binary running 4.07x faster (365.6 ms +/- 3.9 ms -> 89.8 ms +/- 5.1 ms), with system time dropping from 4.35 s to 0.39 s as the kernel pthread_rwlock path is replaced by the userspace fast path.

### DIFF
--- a/lldb/source/Utility/ConstString.cpp
+++ b/lldb/source/Utility/ConstString.cpp
@@ -15,17 +15,70 @@
 #include "llvm/Support/Allocator.h"
 #include "llvm/Support/DJB.h"
 #include "llvm/Support/FormatProviders.h"
-#include "llvm/Support/RWMutex.h"
 #include "llvm/Support/Threading.h"
 
 #include <array>
+#include <mutex>
+#include <shared_mutex>
 #include <utility>
 
 #include <cinttypes>
 #include <cstdint>
 #include <cstring>
 
+#if defined(__APPLE__)
+#include <os/lock.h>
+#endif
+
 using namespace lldb_private;
+
+namespace {
+/// Lock guarding each shard of the ConstString pool. On Apple platforms it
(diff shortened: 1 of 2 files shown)


lldb/source/Utility/ConstString.cpp
```diff
@@ -15,18 +15,71 @@
...
 
+namespace {
+/// Lock guarding each shard of the ConstString pool. On Apple platforms it
+/// wraps os_unfair_lock, which is significantly faster than pthread_rwlock for
+/// concurrent writes, and roughly on par for concurrent reads. On all other
+/// platforms, it wraps std::shared_mutex, preserving reader/writer concurrency.
+///
+/// The class satisfies both Lockable and SharedLockable so it composes with
+/// std::lock_guard and std::shared_lock.
+class PoolMutex {
```
kastiglione, line 43: a possible alternative:

```cpp
#if !defined(__APPLE__)
typedef PoolMutex = std::shared_mutex;
#else
class PoolMutex {
public:
...
};
#end
```

JDevlieghere: That would simplify things a lot. Done.

kastiglione (approved):

kastiglione:
measurements from my machine (factor of 2.33x):

before: 102.8 ms ± 0.7 ms
after: 44.2 ms ± 0.5 ms

Issue #120129, opened 2024-12-16
Optimization in clang frontend breaks `dynamic_cast` on Apple platforms.


ahatanak:
This is caused by 9d525bf94b255df89587db955b5fa2d3c03c2c3e.

`dynamic_cast` in the following code fails unless the optimization is disabled by passing `-fno-assume-unique-vtables`.

% cat a.h
```
struct Foo {
  virtual ~Foo();
};

struct Bar final : public Foo {
};

Foo* makeBar();
```

 % cat a.cc
```
#include "a.h"

Foo::~Foo() {}

Foo* makeBar() {
  return new Bar();
}
```

% cat b.cc
```
#include "a.h"

int main() {
  Foo* f = makeBar();
  Bar* b = dynamic_cast<Bar*>(f); // b will incorrectly be nullptr
  return !!b;
}
```

% clang++ -std=c++11 -c a.cc -O1
% clang++ -std=c++11 -c b.cc -O1
% clang++ -shared -o liba.so a.o 
% clang++ --std=c++11  -o a.out b.o -Wl,-rpath,$PWD liba.so
% ./a.out
% echo $?
0

ahatanak:
llvm's emits `Bar`'s vtable, which is marked as `unnamed_addr` in the IR, as `weak external automatically hidden` and the static linker turns it into a hidden (non-external) symbol.

ahatanak:
@zygoloid @rjmccall

zygoloid:
> llvm's emits `Bar`'s vtable, which is marked as `unnamed_addr` in the IR, as `weak external automatically hidden` and the static linker turns it into a hidden (non-external) symbol.

Interesting. What does `automatically hidden` mean? Is that a new LLVM visibility? I don't see it in the LangRef.

Looking at the code, as far as I can see the only time we should be setting the vtable as having hidden visibility is if the class has hidden visibility, and in that case we turn off the optimization because `dynamic_cast` might treat the types as being the same because they have the same type name despite the visibility annotation, and we respect its decision. Is "hidden" being added due to the `unnamed_addr`? `unnamed_addr` is only supposed to allow merging of symbols, not duplication of them, so that sounds like a bug if so.

[...]

PR #111343, opened 2024-10-07
[ItaniumCXXABI] Mark RTTI type name as global unnamed_addr

luxufan:
The RTTI type name global variable should be marked as global unnamed_addr since the address of RTTI type name global variable
 is not significant.

PR #197855, opened 2026-05-15
[CodeGen] [VTable] don't think linkonce vtable as unique

### LINKED ISSUE #71196: dynamic_cast to final class from thinLTO shared library incorrectly returns nullptr
If a polymorphic subclass is defined only in a header file, shared libraries built with `-flto=thin` give the vtable symbol LOCAL visibility. This means that there may be more than one vtable in a binary that uses this shared library. The dynamic_cast optimizations in llvm 17's libcxxapi (https://reviews.llvm.org/D138005) don't seem to expect that this can happen for final classes, causing dynamic_cast to incorrectly return nullptr.

To repro:
a.h

    struct Foo {
      virtual ~Foo();
    };
    
    struct Bar final : public Foo {
    };
    
    Foo* makeBar();

a.cc

    #include "a.h"
    
    Foo::~Foo() {}
    
    Foo* makeBar() {
      return new Bar();
    }

b.cc

    #include <cstdio>
    #include <cstring>
    
    #include "a.h"
    
    int main(int argc, char **argv) {
      Foo* f = makeBar();
      Bar* b = dynamic_cast<Bar*>(f); // b will incorrectly be nullptr
      void* v;
      memcpy(&v, static_cast<void*>(f), sizeof(v));
      printf("%p %p %p\n", f, b, v);
      delete f;
    
      f = new Bar();
      asm volatile ("":"+r"(f));
      b = dynamic_cast<Bar*>(f);
      memcpy(&v, static_cast<void*>(f), sizeof(v));
      printf("%p %p %p\n", f, b, v);
      delete f;
    
      return 0;
    }

Compile with

    clang++-17 -flto=thin -O3 -fPIC -c -o a.o a.cc
    clang++-17 -O3 -fPIC -flto=thin -shared -o liba.so a.o
    clang++-17 -O3 -o b b.cc -Wl,-rpath,$PWD liba.so

When run, the first dynamic_cast returns nullptr

    $ ./b
    0x55b5b3dc12b0 (nil) 0x7f555c037dc8
    0x55b5b3dc12b0 0x55b5b3dc12b0 0x55b5b337dd80

Changing the final clang invocation to clang++-16 avoids the issue.

readelf shows that Bar's vtable gets LOCAL visibility (in both clang 16 and 17) when `-flto=thin` is used, but WEAK in the default case

    $ readelf -Wl --syms liba.so | c++filt | grep 'for Bar'
        11: 0000000000002005     5 OBJECT  WEAK   DEFAULT   14 typeinfo name for Bar
        14: 0000000000003dd8    24 OBJECT  WEAK   DEFAULT   19 typeinfo for Bar
        35: 0000000000003db8    32 OBJECT  LOCAL  DEFAULT   19 vtable for Bar
        49: 0000000000002005     5 OBJECT  WEAK   DEFAULT   14 typeinfo name for Bar
        57: 0000000000003dd8    24 OBJECT  WEAK   DEFAULT   19 typeinfo for Bar
    $ clang++-17 -O3 -fPIC -c -o a.o a.cc && clang++-17 -O3 -fPIC -shared -o liba_no_lto.so a.o
    $ readelf -Wl --syms liba_no_lto.so | c++filt | grep 'for Bar'
        11: 0000000000002005     5 OBJECT  WEAK   DEFAULT   14 typeinfo name for Bar
        13: 0000000000003db0    32 OBJECT  WEAK   DEFAULT   19 vtable for Bar
        15: 0000000000003dd0    24 OBJECT  WEAK   DEFAULT   19 typeinfo for Bar
        47: 0000000000002005     5 OBJECT  WEAK   DEFAULT   14 typeinfo name for Bar
        55: 0000000000003db0    32 OBJECT  WEAK   DEFAULT   19 vtable for Bar
        57: 0000000000003dd0    24 OBJECT  WEAK   DEFAULT   19 typeinfo for Bar

I am using `Ubuntu clang version 17.0.4 (++20231025123955+afbe3549af4d-1~exp1~20231025124009.57)`

ChuanqiXu9:
Close https://github.com/llvm/llvm-project/issues/71196

The root cause of the problem is, ThinLTO may localize linkonce vtable but there is an optimization to assume the address of vtable is unique to optimize dynamic_cast.

The patch tries to solve the problem by not marking linkonce vtable as unique

### DIFF
--- a/clang/test/CodeGenCXX/dynamic-cast-exact-disabled.cpp
+++ b/clang/test/CodeGenCXX/dynamic-cast-exact-disabled.cpp
@@ -1,4 +1,3 @@
-// RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,EXACT
 // RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O0 -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
 // RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -fvisibility=hidden -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
 // RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -fapple-kext -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
--- a/clang/test/CodeGenCXX/dynamic-cast-exact-disabled.cpp
+++ b/clang/test/CodeGenCXX/dynamic-cast-exact-disabled.cpp
@@ -1,10 +1,12 @@
+// RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,EXACT
 // RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O0 -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
-// RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -fvisibility=hidden -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
 // RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -fapple-kext -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
 // RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -fno-assume-unique-vtables -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
 
 struct A { virtual ~A(); };
-struct B final : A { };
+A::~A() {}
+struct B final : A { ~B(); };
+B::~B() {}
 
 // CHECK-LABEL: @_Z5exactP1A
 B *exact(A *a) {
@@ -16,10 +18,12 @@ B *exact(A *a) {
 struct C {
   virtual ~C();
 };
+C::~C() {}
 
 struct D final : private C {
-
+  virtual ~D();
 };
+D::~D() {}
 
 // CHECK-LABEL: @_Z5exactP1C
 D *exact(C *a) {
--- a/clang/lib/CodeGen/ItaniumCXXABI.cpp
+++ b/clang/lib/CodeGen/ItaniumCXXABI.cpp
@@ -201,6 +201,7 @@ public:
   /// practice in some cases due to language extensions.
   bool hasUniqueVTablePointer(QualType RecordTy) {
     const CXXRecordDecl *RD = RecordTy->getAsCXXRecordDecl();
+    llvm::GlobalVariable::LinkageTypes Linkage = CGM.getVTableLinkage(RD);
 
     // Under -fapple-kext, multiple definitions of the same vtable may be
     // emitted.
@@ -215,8 +216,15 @@ public:
 
     // If there's only one definition of the vtable in the program, it has a
     // unique address.
-    if (!llvm::GlobalValue::isWeakForLinker(CGM.getVTableLinkage(RD)))
+    if (!llvm::GlobalValue::isWeakForLinker(Linkage))
       return true;
+
+    // Header-defined classes can have a linkonce_odr vtable. ThinLTO can
+    // localize those weak definitions in another linkage unit, so the vtable
+    // address is not a reliable exact dynamic_cast discriminator.
+    // See https://github.com/llvm/llvm-project/issues/71196 for example.
+    if (Linkage == llvm::GlobalValue::LinkOnceODRLinkage)
+      return false;
 
     // Even if there are multiple definitions of the vtable, they are required
     // by the ABI to use the same symbol name, so should be merged at load
--- a/clang/test/CodeGenCXX/GH71196.cpp
+++ b/clang/test/CodeGenCXX/GH71196.cpp
@@ -1,0 +1,17 @@
+// RUN: %clangxx --target=x86_64-unknown-linux-gnu -O1 -fPIC -emit-llvm -S -std=c++11 -o - %s | FileCheck %s --check-prefixes=CHECK
+
+struct Foo {
+  virtual ~Foo();
+};
+
+struct Bar final : Foo {};
+
+Bar *make() { return new Bar(); }
+
+// CHECK: @_ZTV3Bar = linkonce_odr{{.*}}constant
+
+// CHECK-LABEL: define {{.*}} @_Z4castP3Foo
+Bar *cast(Foo *f) {
+  // CHECK: call {{.*}} @__dynamic_cast(ptr nonnull %[[ARG:.*]], ptr nonnull @_ZTI3Foo, ptr nonnull @_ZTI3Bar, i64 0)
+  return dynamic_cast<Bar *>(f);
+}
--- a/clang/test/CodeGenCXX/dynamic-cast-exact-disabled-weak.cpp
+++ b/clang/test/CodeGenCXX/dynamic-cast-exact-disabled-weak.cpp
@@ -1,0 +1,23 @@
+// RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -fvisibility=hidden -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
+
+struct A { virtual ~A(); };
+struct B final : A { };
+
+// CHECK-LABEL: @_Z5exactP1A
+B *exact(A *a) {
+  // INEXACT: call {{.*}} @__dynamic_cast
+  return dynamic_cast<B*>(a);
+}
+
+struct C {
+  virtual ~C();
+};
+
+struct D final : private C {
+};
+
+// CHECK-LABEL: @_Z5exactP1C
+D *exact(C *a) {
+  // INEXACT: call {{.*}} @__dynamic_cast
+  return dynamic_cast<D*>(a);
+}
--- a/clang/test/CodeGenCXX/dynamic-cast-exact-weak-vtable.cpp
+++ b/clang/test/CodeGenCXX/dynamic-cast-exact-weak-vtable.cpp
@@ -1,0 +1,21 @@
+// RUN: %clangxx --target=x86_64-unknown-linux-gnu -O1 -fPIC -emit-llvm -S -std=c++11 -o - %s | FileCheck %s --check-prefixes=CHECK,INEXACT
+
+struct Foo {
+  virtual ~Foo();
+};
+
+// Header-defined final classes have a weak_odr/linkonce_odr vtable. In the
+// ThinLTO shared-library reproducer for issue #71196, using that vtable as a
+// unique identity for exact dynamic_cast is wrong because another linkage unit
+// can end up with a different local copy.
+struct Bar final : Foo {};
+
+Bar *make() { return new Bar(); }
+
+// CHECK: @_ZTV3Bar = linkonce_odr{{.*}}constant
+
+// CHECK-LABEL: define {{.*}} @_Z4castP3Foo
+Bar *cast(Foo *f) {
+  // INEXACT: call {{.*}} @__dynamic_cast(ptr nonnull %[[ARG:.*]], ptr nonnull @_ZTI3Foo, ptr nonnull @_ZTI3Bar, i64 0)
+  return dynamic_cast<Bar *>(f);
+}
--- a/clang/test/CodeGenCXX/dynamic-cast-exact.cpp
+++ b/clang/test/CodeGenCXX/dynamic-cast-exact.cpp
@@ -1,15 +1,21 @@
 // RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -emit-llvm -fcxx-exceptions -fexceptions -std=c++11 -o - -O1 -disable-llvm-passes | FileCheck %s --implicit-check-not='call {{.*}} @__dynamic_cast'
 struct Offset { virtual ~Offset(); };
+Offset::~Offset() { }
 struct A { virtual ~A(); };
-struct B final : Offset, A { };
+A::~A() { }
+struct B final : Offset, A { virtual ~B(); };
+B::~B() { }
 
 struct C { virtual ~C(); int c; };
+C::~C() { }
 struct D : A { int d; };
 struct E : A { int e; };
 struct F : virtual A { int f; };
 struct G : virtual A { int g; };
-struct H final : C, D, E, F, G { int h; };
-struct H1 final: C, private D { int h1; };
+struct H final : C, D, E, F, G { int h; virtual ~H(); };
+H::~H() { }
+struct H1 final: C, private D { int h1; virtual ~H1(); };
+H1::~H1() { }
 
 // CHECK-LABEL: @_Z7inexactP1A
 C *inexact(A *a) {
@@ -103,7 +109,8 @@ H1 &exact_invalid_multi(D& d) {
 
 namespace GH137518 {
   class base { virtual void fn() = 0; };
-  class test final : base { virtual void fn() { } };
+  class test final : base { virtual void fn(); };
+  void test::fn() {}
   test* new_test() { return new test(); }
 
   // CHECK-LABEL: @_ZN8GH1375184castEPNS_4baseE(
@@ -120,8 +127,10 @@ namespace GH137518 {
 namespace GH64088 {
   // Ensure we mark the B vtable as used here, because we're going to emit a
   // reference to it.
-  // CHECK: define {{.*}} void @_ZN7GH640881BD0Ev(
+  // CHECK: define {{.*}}void @_ZN7GH640881BD0Ev(
   struct A { virtual ~A(); };
-  struct B final : A { virtual ~B() = default; };
+  struct B final : A { virtual ~B(); };
+  B::~B() { }
+
   B *cast(A *p) { return dynamic_cast<B*>(p); }
 }
--- a/clang/test/CodeGenCXX/ptrauth-dynamic-cast-exact.cpp
+++ b/clang/test/CodeGenCXX/ptrauth-dynamic-cast-exact.cpp
@@ -3,29 +3,35 @@
 struct A {
   virtual ~A();
 };
+A::~A() {}
 struct B {
   int foo;
   virtual ~B();
 };
 struct C final : A, B {
-  virtual void f(){};
+  virtual void f();
 };
+void C::f() {}
 struct D final : B, A {
-  virtual void f(){};
+  virtual void f();
 };
+void D::f() {}
 
 struct Offset {
   virtual ~Offset();
 };
+Offset::~Offset() {}
 struct E {
   virtual ~E();
 };
+E::~E() {}
 struct F final : Offset, E {
 };
 struct G {
   virtual ~G();
   int g;
 };
+G::~G() {}
 struct H : E {
   int h;
 };
@@ -40,8 +46,11 @@ struct K : virtual E {
 };
 struct L final : G, H, I, J, K {
   int l;
+  virtual ~L();
 };
-struct M final: G, private H { int m; };
+L::~L() {}
+struct M final: G, private H { int m; ~M(); };
+M::~M() {}
 
 // CHECK-LABEL: @_Z10exact_to_CP1A
 C *exact_to_C(A *a) {


clang/test/CodeGenCXX/dynamic-cast-exact-disabled.cpp
```diff
@@ -1,4 +1,3 @@
-// RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,EXACT
```
ojhunt, line 1: This test should be corrected, not have the baseline removed.

The purpose of this test is to have a baseline in which we demonstrate that we will optimize a dynamic_cast on the assumption of vtable uniqueness, but we won't if the various "no uniqueness assumption" flags are set.

If the baseline test is being removed because it no longer produces the expected exact match baseline, then the rest of the test is meaningless.

What actually needs to happens is the class definitions need to be updated.

ojhunt (changes requested):
The code changes look good by the dynamic-cast-exact-disabled.cpp change is not correct - it is meant to test that the exact match optimization is not performed, but from the current test change it looks like the current test is incorrect w.r.t the actual correct behavior.

clang/test/CodeGenCXX/dynamic-cast-exact-disabled.cpp
ChuanqiXu9: Done. I removed due to the title. Now I add it back and make the vtable strong symbol.

dynamic-cast-exact-disabled-weak.cpp is split as strongness/weakness takes higher precedence than visibility.

clang/test/CodeGenCXX/dynamic-cast-exact-disabled.cpp
```diff
@@ -1,11 +1,12 @@
 // RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,EXACT
 // RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O0 -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
-// RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -fvisibility=hidden -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
 // RUN: %clang_cc1 -I%S %s -triple x86_64-apple-darwin10 -O1 -fapple-kext -emit-llvm -std=c++11 -o - | FileCheck %s --check-prefixes=CHECK,INEXACT
```
ojhunt, line 3: as one last paranoia test could you add an `-flto` run for this one? (we've got testing for the explicit linker mode directly but the original report was lto related so we should have that as well).

ojhunt (commented):
This looks ok as a functional "make it work" change, but I'm actually less convinced this is correct.

If you have a weak symbol, all versions of that weak symbol should end up with the same address at (dynamic or static) link time, so if lto is dropping the weak label, or updating(direct linking?) some addresses but not all, then that seems wrong.

Alternatively if the unique vtable optimization is not referencing the vtable pointer correctly - as in its linkage is not correctly weak - then that is the bug.

Anyway, I'd like someone with more backend/linker/lto experience to look at this as well, because while this may address the current problem I'm not certain it is the actual fix for it.

ChuanqiXu9:
Added @teresajohnson as the ThinLTO expert. I feel the point may be, linkonce linkage is not weak linkage. Linkonce linkage means it can be dropped if it is not referenced in this TU as others who need it has its own linkonce linkage.

But I am also surprised that we don't see other similar failures.

clang/test/CodeGenCXX/dynamic-cast-exact-disabled.cpp
ChuanqiXu9: `-flto` is not directly related to this. As if this TU is not compiled as `-flto` but other TU compiled as `-flto`, all of them shouldn't assume the address to be the same.

ojhunt:
Ok, I am convinced that this is an LTO bug. ODR means you can assume all versions are semantically identical, and I believe the implementation is free to choose whichever it wants (either at static or dynamic link time), or drop it if unused. But I don't believe that is what the described behavior is - it's actually duplicating the ODR symbol.

I used AI to understand this (where by "using AI" I kept getting nonsense AI answers at the top of google results while I tried to find an actual explanation of the rules about linking these things. ... that might actually be frustrated or possibly even punished by ai rather than it being used).

According to our docs linkonce does not permit us to inline a symbol unless we know it is the definitive definition, which is definitionally not the case here: we have 2+ definitions.

Section 5.2 of the itanium abi says that symbols with "vague linkage" like vtables are meant to be deduplicated.

To me this implies that the interning/localization of vtable addresses by lto is invalid.

ChuanqiXu9:
FWIW, this seems to be a designed feature for LTO. I didn't find related documentation but in the implementation it has related comments: https://github.com/llvm/llvm-project/blob/54789b90c59fb7badb94c28e6433ae14b40d6e78/llvm/lib/LTO/LTO.cpp#L554-L595

ChuanqiXu9:
Maybe the problem relates to the judgement of "exported"? We need people who know LTO better here.

ojhunt:
> Maybe the problem relates to the judgement of "exported"? We need people who know LTO better here.

Yeah - I think this is sufficiently complicated that our guesswork is likely suboptimal :D

teresajohnson:
Generally for linkonce_odr defs we convert one copy (the one identified by the linker as prevailing) to weak_odr so that it is kept, and the rest to available_externally so that they are dropped after inlining.

The case referenced in the earlier comment kicks in which should only happen in very limited situations: there is a single copy. I see in the linked bug that this is a shared library, so there is a single copy, but I would have thought that the earlier analysis in LTO that feeds into the IsExported callback would indicate this as exported based on the info we get from the linker. I'll need to look at what is happening there, I should have time for a quick look today.

ojhunt:
> Generally for linkonce_odr defs we convert one copy (the one identified by the linker as prevailing) to weak_odr so that it is kept, and the rest to available_externally so that they are dropped after inlining.
> 
> The case referenced in the earlier comment kicks in which should only happen in very limited situations: there is a single copy. I see in the linked bug that this is a shared library, so there is a single copy, but I would have thought that the earlier analysis in LTO that feeds into the IsExported callback would indicate this as exported based on the info we get from the linker. I'll need to look at what is happening there, I should have time for a quick look today.

@pcc pointed out a couple of existing tests that are adding unnamed_addr to vtable symbols, which by the existing comment and my (non-linking expert) reading of the itanium spec might be wrong (because it means there is more than one symbol for a single vtable)?

ojhunt:
> > Generally for linkonce_odr defs we convert one copy (the one identified by the linker as prevailing) to weak_odr so that it is kept, and the rest to available_externally so that they are dropped after inlining.
> > The case referenced in the earlier comment kicks in which should only happen in very limited situations: there is a single copy. I see in the linked bug that this is a shared library, so there is a single copy, but I would have thought that the earlier analysis in LTO that feeds into the IsExported callback would indicate this as exported based on the info we get from the linker. I'll need to look at what is happening there, I should have time for a quick look today.
> 
> @pcc pointed out a couple of existing tests that are adding unnamed_addr to vtable symbols, which by the existing comment and my (non-linking expert) reading of the itanium spec might be wrong (because it means there is more than one symbol for a single vtable)?

Oh indeed all of the vtables are labeled as unused_addr - if that allows them to be duplicated that seems like it might be an error?

ojhunt:
"might" is carrying a lot of weight here: I am not a linker person

ojhunt:
Ok, I think I understand the pieces. vtables are ODR under C++ rules but that isn't observable, so can be unnamed_addr. The RTTI objects are observable however, so they can't be unnamed.

I think the error _might_ be that this optimization is comparing the vtable pointer rather than the type info pointer in the vtable.

efriedma-quic:
#120129 has a better discussion of what's happening here.

ojhunt:
I'm not sure it is providing more information, this bug has everything we know currently:

* The Itanium ABI seems to say that it is incorrect to do this with vtables
* This optimization assumes that this is the case
* We are marking vtables as linkonce odr *and* unnamed_addr
* In various parts of llvm, and the linkers, etc consider unnamed_addr as (functionally) permitting duplication of those symbols by making hidden copies of the object

The path forward depends on what the *correct* behavior is. If the interpretation is that itanium does not permit duplicating the vtables then it is an error for them to be labeled as unnamed_addr. If it is valid for vtables to be duplicated then this optimization is invalid, we _could_ try to move the optimization pass onto the RTTI field.

Alternatively, @jaredwy (I think?) mentioned unnamed_local and after reading the semantics, that seems like it is much likely to be the correct annotation for vtables, and would be more consistent with the behavior specified in the itanium abi. I think that this alone would have avoided the issues in https://github.com/llvm/llvm-project/issues/120129 , but I don't know the performance trade offs.

And lastly - I have no linker experience: I have been reading the documentation and specs, but there's a world of difference between what is written and what is practically necessary so I'm happy to defer to experts :D

jaredwy:
@ojhunt I think my position, after reading some of the newer additions to the spec along with the review that introduced this optimisation, is that it should not have either annotation. I do think local_unnamed_addr fits a little more tightly, and as you raised fixes the issue today. 

I think the spec is a little iffy around this whole area and does admit that its current approach may not be ideal (for rtti and there are ABI breaks) but I would argue it extends here as well.

teresajohnson:
> > Generally for linkonce_odr defs we convert one copy (the one identified by the linker as prevailing) to weak_odr so that it is kept, and the rest to available_externally so that they are dropped after inlining.
> > The case referenced in the earlier comment kicks in which should only happen in very limited situations: there is a single copy. I see in the linked bug that this is a shared library, so there is a single copy, but I would have thought that the earlier analysis in LTO that feeds into the IsExported callback would indicate this as exported based on the info we get from the linker. I'll need to look at what is happening there, I should have time for a quick look today.
> 
> @pcc pointed out a couple of existing tests that are adding unnamed_addr to vtable symbols, which by the existing comment and my (non-linking expert) reading of the itanium spec might be wrong (because it means there is more than one symbol for a single vtable)?

Yes, this is the issue. If the address is significant, it shouldn't get this. Confirmed that the unnamed_addr on the vtable is feeding into the linker's determination of whether the symbol needs to be externally visible. Specifically, via GlobalValue::canBeOmittedFromSymbolTable which contains this check: 

https://github.com/llvm/llvm-project/blob/fa4d73761d122c49ebc591118c83f5e142c98b4d/llvm/lib/IR/Globals.cpp#L481-L482

which is called from the IRSymtab builder from here:

https://github.com/llvm/llvm-project/blob/fa4d73761d122c49ebc591118c83f5e142c98b4d/llvm/lib/Object/IRSymtab.cpp#L269-L270

This flag is called from lld from here to set the ltoCanOmit flag:

https://github.com/llvm/llvm-project/blob/fa4d73761d122c49ebc591118c83f5e142c98b4d/lld/ELF/InputFiles.cpp#L1890

which eventually causes the symbol to be marked exported here:

https://github.com/llvm/llvm-project/blob/fa4d73761d122c49ebc591118c83f5e142c98b4d/lld/ELF/Symbols.cpp#L385

And this feeds into LTO. Not sure exactly how this happens via the gold-plugin, but gold has the same behavior (the issue goes away when I remove unnamed_addr).

efriedma-quic:
It's not really just whether vtables should be marked unnamed_addr, it's also what unnamed_addr is actually supposed to mean.

Given that we currently represent the addresses of global variables as "llvm::Constant", it's fundamentally unsafe from the perspective of IR representation to duplicate a global.  It breaks the identity of IR values, specifically that you can have a pointer that is not equal to itself.

If we want to have duplicatable globals, we need a different representation.  For example, we could introduce a new intrinsic `llvm.duplicatableglobal.address`, similar to `llvm.threadlocal.address`.  And we introduce an IR rule: it's illegal to reference the address of a duplicatable global directly; the address has to be computed using a call to `llvm.duplicatableglobal.address`.  Any given call to `llvm.duplicatableglobal.address` returns a self-consistent value, but different calls are allowed to return different results.

This sort of duplication actually led to a miscompile at one point; see #32127.  (It's harder to hit this issue in the context of shared libraries, but not impossible.)

ojhunt:
@teresajohnson is local_unnamed_addr the correct thing here? My reading _seems_ to imply it would give the backend the right to do the kinds of optimizations that would *seem* like what is intended without the symbol problem. I _think_ those optimizations would still be valid as having *different* definitions of the vtable content would be an odd violation

jaredwy:
Longer term, it seems that no annotation by default is the only move that aligns to the spec(both in the letter and the spirit). Where there is actually some wiggle room, like hidden visibility, we should just let thinLTO/LTO infer duplication, forcing an overarching annotation seems to be fighting the spec and its future direction (key function use has grown over time). 5.2.3 appears to mandate singular emission for key-function classes, and the RTTI precedent shows the ABI relies on program wide uniqueness. A global annotation that implies duplication is allowed seems like the wrong move here. 

It's worth noting that the dynamic_cast + RTTI interaction has been an issue for performance and optimisation, the spec acknowledges it as not optimal and has observed ABI breaks in the wild. There may be an opening here to push for the spec to commit more firmly to uniqueness rather than leaving it up to the toolchains to decide direction.

ChuanqiXu9:
Thanks for all the discussion here. My summary is:

- ThinLTO/LTO's behavior is good. The problem is the frontend can't mark the vtable as `unnamed_addr` and compare its address to implement `dynamic_cast`.
- So the solution in the short term is either use `local_unnamed_addr` or use RTTI information for the frontend optimization for `dynamic_cast`. @ojhunt would you want to implement this? If not, I'd like to use the `local_unnamed_addr`  solution.
- @efriedma-quic mentions some concerns in LLVM IR that we can't safely  duplicate a global.
- @jaredwy mentions we should have a conclusion in Itanium C++ ABI. But I am not sure if Itanium C++ ABI group is still active. I had some conversion with @AaronBallman  in Discord.

ojhunt:
> Thanks for all the discussion here. My summary is:
> 
> * ThinLTO/LTO's behavior is good. The problem is the frontend can't mark the vtable as `unnamed_addr` and compare its address to implement `dynamic_cast`.

Right lto is behaving correctly here _given_ clang is marking vtables as being unnamed_addr. My belief is that clang doing this is incorrect behavior, I'm unsure if `local_unnamed_addr` is correct. I _think_ it is but I'd really like someone with more familiarity than "I just spent time reading specification docs" to weigh in.

We could switch to an RTTI check regardless of whatever we do in the end anyway just to deal with older libraries.

> * So the solution in the short term is either use `local_unnamed_addr` or use RTTI information for the frontend optimization for `dynamic_cast`. @ojhunt would you want to implement this? If not, I'd like to use the `local_unnamed_addr`  solution.

My inclination is `local_unnamed_addr` or removing `*unnamed_addr` entirely, but I can imagine the complete removal having performance impact beyond the simple `local_unnamed_addr`, but again: not a linker person. In principle removing or replacing `unnamed_addr` would allow more code to use this optimization, so that could be a net win. Clearly there is code that hits the optimization, that's why this bug is here :D

> * @efriedma-quic mentions some concerns in LLVM IR that we can't safely  duplicate a global.
> * @jaredwy mentions we should have a conclusion in Itanium C++ ABI. But I am not sure if Itanium C++ ABI group is still active. I had some conversion with @AaronBallman  in Discord.

ChuanqiXu9:
@MaskRay hi, would you like to take a look here? You're the most linker person I know : )

zygoloid:
This has come up before, and there was a discussion there that IIRC resulted in: `unnamed_addr` permits merging but *not* cloning of symbols. It'd be good to track that discussion down and see if any of the arguments there have been missed, since we seem to be coming to a different conclusion now.

zygoloid:
See also prior discussion in #120129

zygoloid:
Also see prior discussion on #111343, though I think Eli's [comment earlier in this thread](https://github.com/llvm/llvm-project/pull/197855#issuecomment-4467807076) captures that well.

ojhunt:
> This has come up before, and there was a discussion there that IIRC resulted in: `unnamed_addr` permits merging but _not_ cloning of symbols. It'd be good to track that discussion down and see if any of the arguments there have been missed, since we seem to be coming to a different conclusion now.

I think it's agreed that is the correct behavior of `unnamed_addr` - my understanding is that the conversation now is more "given the semantics of `unnamed_addr` is it correct for vtables to be marked as such"

efriedma-quic:
Ignoring the LLVM IR representational question, the ABI questions are, I guess:

- Is it okay to merge vtables?  I guess this is sort of an empty question, though, because realistically we can't merge vtables in any case.  They have a typeinfo pointer (assuming -frtti), so the marking isn't doing any good.
- Is it okay to duplicate vtables?  We can either say no, it's not legal, or yes it is legal in some cases.  If we say it is legal, then we should suppress dynamic_cast optimization in those cases.  If we say it isn't legal, we should drop the unnamed_addr marking (ignoring that we're currently abusing unnamed_addr to mean something it shouldn't mean).

Saying it's not legal to duplicate weak vtables would probably be significantly more disruptive to existing code. vtables are a lot more common than dynamic_cast, and the existing optimization has very good reasons to exist: symbols from a different DSO are significantly more expensive.

CC @rjmccall @ahatanak

zygoloid:
> * Is it okay to merge vtables?  I guess this is sort of an empty question, though, because realistically we can't merge vtables in any case.  They have a typeinfo pointer (assuming -frtti), so the marking isn't doing any good.

I can't find a reason why it would be disallowed by the Itanium ABI rules. And for construction vtables, it could be useful, for example here:

```c++
struct A {
  A() {}
  virtual ~A() {}
};

struct B : virtual A {};
struct X : B {};
struct Y : B {};

X x;
Y y;
```

... we generate two different (but identical) constants for the B-in-X and B-in-Y construction vtables. Using `unnamed_addr` for at least construction vtables seems like a moderately good thing -- and as noted below, these vtables are also cloneable. (That said, arguably it'd be better to generate symbols that encode the offset-to-top information rather than the derived class type, and emit construction vtables as aliases to those, in order to *force* deduplication.)

Also, the `-fno-rtti` case covers a pretty significant fraction of the world's C++ code, so that's worth considering too. Outside of construction vtables, we'd only have an opportunity for merging if the destructor is non-virtual, which reduces the scope of the optimization quite a lot, but I imagine we could find a few mergeable vtables in (for example) the `clang::Decl` class hierarchy if we made its destructor non-virtual. (It's only virtual to provide an anchor for the vtable emission.) The value here seems somewhat limited, but if our approach is to add all the correct annotations just in case some optimizer wants them, I think marking `-fno-rtti` vtables as mergeable is at least correct.

> * Is it okay to duplicate vtables?  We can either say no, it's not legal, or yes it is legal in some cases.

The Itanium ABI [is clear that it is not](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#vtable-general) OK for vtables used for most-derived objects:

> A *virtual table* (*vtable*) is a table of information used to dispatch virtual functions, to access virtual base class subobjects, and to access information for runtime type identification (RTTI). Each class that has virtual member functions or virtual bases has an associated set of virtual tables. There may be multiple virtual tables for a particular class, if it is used as a base class for other classes. However, **the virtual table pointers within all the objects (instances) of a particular most-derived class point to the same set of virtual tables**.

However, IIRC this rule is not adopted by all psABIs based on the Itanium C++ ABI (in particular, on Apple / MachO targets I think there is no guarantee of uniqueness).

pcc:
> Given that we currently represent the addresses of global variables as "llvm::Constant", it's fundamentally unsafe from the perspective of IR representation to duplicate a global. It breaks the identity of IR values, specifically that you can have a pointer that is not equal to itself.

Isn't that just the dual of the merging behavior (can have pointers to "different" objects that compare equal)? I think of unnamed_addr as implying that a pointer comparison involving the global basically results in poison.

> If we want to have duplicatable globals, we need a different representation. For example, we could introduce a new intrinsic llvm.duplicatableglobal.address, similar to llvm.threadlocal.address. And we introduce an IR rule: it's illegal to reference the address of a duplicatable global directly; the address has to be computed using a call to llvm.duplicatableglobal.address. Any given call to llvm.duplicatableglobal.address returns a self-consistent value, but different calls are allowed to return different results.
>
> This sort of duplication actually led to a miscompile at one point; see #32127. (It's harder to hit this issue in the context of shared libraries, but not impossible.)

The intrinsic seems like a good idea for users that need a stable pointer to an unnamed_addr global such as a string literal, to address the bug you cited. But wouldn't the restriction on directly taking the address forbid creating globals that reference a unnamed_addr global, such as the vtable pointer in a dynamic object (which sounds like it will remain as unnamed_addr in `-fno-rtti` mode) or similar constructs in other languages?

To me it seems preferable to formalize a rule for values obtained by directly taking the address ("comparisons result in poison") rather than adding the restriction.

efriedma-quic:
> Isn't that just the dual of the merging behavior (can have pointers to "different" objects that compare equal)? I think of unnamed_addr as implying that a pointer comparison involving the global basically results in poison.

It's not really the same thing.  If you have two pointers `%p1` and `%p2`, you don't know if they're equal without any other information.  Two global variables `@g1` and `@g2` is basically the same thing; we assume they aren't equal only if we can prove merging won't happen.  If you have a function `define void @f(ptr %p)`, and `%p` isn't equal to `%p`, that's an fundamental problem.

> But wouldn't the restriction on directly taking the address forbid creating globals that reference a unnamed_addr global, such as the vtable pointer in a dynamic object (which sounds like it will remain as unnamed_addr in -fno-rtti mode) or similar constructs in other languages?

I haven't really thought through that part.  Maybe we could say that if a global wants to refer to a duplicatable global, it has to go through an alias.  The alias would always be equal to itself, but maybe not equal to other references to the global.

pcc:
> It's not really the same thing. If you have two pointers `%p1` and `%p2`, you don't know if they're equal without any other information. Two global variables `@g1` and `@g2` is basically the same thing; we assume they aren't equal only if we can prove merging won't happen. If you have a function `define void @f(ptr %p)`, and `%p` isn't equal to `%p`, that's an fundamental problem.

Is it though? Suppose I have the function:
```
define i1 @f(ptr %p) {
  %pred = icmp eq ptr %p, %p
  ret i1 %pred
}
```
and I call it with an `unnamed_addr` global `@p`. My proposed rule would make the `icmp` yield poison if `%p` is `unnamed_addr`. Per https://llvm.org/docs/UndefinedBehavior.html#poison-values :

> Poison values can be replaced with any value of type (undef, concrete values, or a freeze instruction).

So an optimization pass that replaces `%pred` with `true` is sound under my proposed rule. Am I missing anything?

zygoloid:
> To me it seems preferable to formalize a rule for values obtained by directly taking the address ("comparisons result in poison") rather than adding the restriction.

I just looked through a bunch of uses of `setUnnamedAddr` in clang's CodeGen, and I think this would make most of them incorrect. The only remaining class of globals for which it seems robustly correct would be ones that are literally never address-taken (such as constructors). It'd be better if we said that only pointer comparisons between a pointer into the `unnamed_addr` global and a pointer that is not based on that global are poisoned, but even then, any time such a pointer escapes into user code in any way, `unnamed_addr` would become incorrect. Also things like profile-driven speculative devirtualization that insert pointer comparisons would become incorrect if virtual functions remain `unnamed_addr`.


To me, it seems that the right change to make is to clarify that `unnamed_addr` does *not* permit cloning, because:
* Cloning is not in general a sound transformation based only on an annotation on the global, since (for example) we can't distinguish between "based on the same access to the global" and "based on two different accesses to the global".
* The existing documentation of `unnamed_addr` says it allows merging, and does not say that it allows cloning.
* Existing uses of `unnamed_addr` assume that it allows merging and not cloning.

If we want cloning support, we should add a separate facility for that. I like @efriedma-quic's suggestion of a builtin that nondeterministically picks an address for a global. And generally, I think we should be looking for something that can model C++'s semantics for string literals:

https://eel.is/c++draft/basic#compound-4
https://eel.is/c++draft/expr#const.core-2.28

That is: each source-language evaluation of a reference to the global gives you a pointer to a suitable global; different evaluations can give different pointers. Comparisons based on different evaluations have unknown-until-runtime (but fully defined) results.

ojhunt:
> Ignoring the LLVM IR representational question, the ABI questions are, I guess:
> 
> * Is it okay to merge vtables?  I guess this is sort of an empty question, though, because realistically we can't merge vtables in any case.  They have a typeinfo pointer (assuming -frtti), so the marking isn't doing any good.

I need some clarity on this one: do you mean for the same type or not? If they're not the same type but are identical it would (as you note) mean no rtti could be present. The absence if RTTI means dynamic_cast isn't available so the mismatch between "real" dynamic type, and the distinction from the vtable pointer itself would be moot (though happily ptrauth prevents this :D :D :D :D)

If they are the same type, my reading of the itanium spec says that there should be one universal symbol for the vtable pointer: the _must_ be merged, there cannot be duplicates.

> * Is it okay to duplicate vtables?  We can either say no, it's not legal, or yes it is legal in some cases.  If we say it is legal, then we should suppress dynamic_cast optimization in those cases.  If we say it isn't legal, we should drop the unnamed_addr marking (ignoring that we're currently abusing unnamed_addr to mean something it shouldn't mean).
> 
> Saying it's not legal to duplicate weak vtables would probably be significantly more disruptive to existing code. vtables are a lot more common than dynamic_cast, and the existing optimization has very good reasons to exist: symbols from a different DSO are significantly more expensive.


I'm unsure what you mean here? We expected duplicate weak vtables in the general case, what is the duplication path you're thinking of?

> 
> CC @rjmccall @ahatanak

zygoloid:
> So an optimization pass that replaces `%pred` with `true` is sound under my proposed rule. Am I missing anything?

It's currently correct to replace
```llvm
load %p
```
with
```llvm
icmp eq %p, <something>
...
load %p
```
and profile-driven optimizations may wish to do exactly that. Your rule makes that transformation incorrect.

ojhunt:
People seemed to be turning this into a discussion of `{local_}unnamed_addr` semantics here which I'm not sure is the right place the discussion. The question we're trying to answer in this bug is not "what are the correct semantics unnamed_addr"; the question we need to answer is whether it is correct to label vtables as `unnamed_addr`.

By the documented description of them I believe that `local_unnamed_addr` is the strongest correct option: the fact that the actual address of the vtable is not relevant to a given TU is not relevant to whether it is relevant outside of a TU. `local_unnamed_addr` seems to exactly match those semantics, but the more conservative approach "this is _meant_ to be a globally unique symbol so don't label it with `unnamed_addr` at all" would also seem correct.

The more I look into this, the more it seems that attaching the unnamed_addr attribute to a vtable is simply incorrect. The symbol is implicitly referenced by name across TU and library boundaries by virtue of ODR.

pcc:
> It'd be better if we said that only pointer comparisons between a pointer into the unnamed_addr global and a pointer that is not based on that global are poisoned, but even then, any time such a pointer escapes into user code in any way, unnamed_addr would become incorrect.

Having it only be poison if both sides are `unnamed_addr` SGTM in principle, but would need to check whether linkers will merge unnamed_addr and non-unnamed_addr sections, which would break this.

In general these pointers should only escape into user code in C/C++ if it's something like a string literal whose unspecified semantics for comparisons matches the proposed rule I believe.

> I just looked through a bunch of uses of setUnnamedAddr in clang's CodeGen, and I think this would make most of them incorrect.

Looking through them myself, it seems to be mostly string literals and non-user-accessible constructs. Did you find a case where something other than a string literal escapes to user code? That may be a bug.

> and profile-driven optimizations may wish to do exactly that. Your rule makes that transformation incorrect.

Only if one of the operands is `unnamed_addr`, no? The comparison is already potentially incorrect with global merging, so these optimizations should be removing `unnamed_addr` for correctness if they aren't already doing so.

pcc:
> People seemed to be turning this into a discussion of {local_}unnamed_addr semantics here which I'm not sure is the right place the discussion. The question we're trying to answer in this bug is not "what are the correct semantics unnamed_addr"; the question we need to answer is whether it is correct to label vtables as unnamed_addr.

Yeah, let's move that discussion elsewhere.

> The more I look into this, the more it seems that attaching the unnamed_addr attribute to a vtable is simply incorrect. The symbol is implicitly referenced by name across TU and library boundaries by virtue of ODR.

Agreed. It seems unlikely that unnamed_addr on vtables actually helps much with optimizations anyway, as others have previously noted.

rjmccall:
I'm not sure we can decide whether it's correct to mark v-tables `unnamed_addr` without knowing what the semantics of marking something `unnamed_addr` are. Well, I suppose we could decide that it's *incorrect*, but it seems impossible to decide that it's correct.

I would caution against formalizing "comparisons are poison" semantics for all uses of `unnamed_addr`. It's not uncommon for patterns like these to be associated with "fast path" checks, where a comparison is performed before falling back to a slower operation. It's okay if that comparison conservatively returns unequal, which is compatible with values being cloned across dylib boundaries, but it's not okay if the comparison just gets folded to an arbitrary value. (*Some* checks might be okay if the comparison returns true because the comparison was equal prior to cloning, but this could also lead to consistency errors.)

pcc:
> I'm not sure we can decide whether it's correct to mark v-tables unnamed_addr without knowing what the semantics of marking something unnamed_addr are. Well, I suppose we could decide that it's incorrect, but it seems impossible to decide that it's correct.

We could arguably decide its correctness if we base our decision on the unnamed_addr semantics that LLVM currently implements, rather than the semantics that we would like to have. Then, in order to make immediate progress, we can have this PR be about aligning Clang with the currently implemented semantics, and we can decide and implement the desired semantics in a followup. I would argue that the current implementation makes unnamed_addr comparisons evaluate to `freeze(poison)`, which would justify removing unnamed_addr from vtables (a relatively straightforward change) until we're ready to add it back, if the new semantics allow it.

> I would caution against formalizing "comparisons are poison" semantics for all uses of unnamed_addr. It's not uncommon for patterns like these to be associated with "fast path" checks, where a comparison is performed before falling back to a slower operation. It's okay if that comparison conservatively returns unequal, which is compatible with values being cloned across dylib boundaries, but it's not okay if the comparison just gets folded to an arbitrary value. (Some checks might be okay if the comparison returns true because the comparison was equal prior to cloning, but this could also lead to consistency errors.)

That's a fair point. Some users may be fine with comparisons returning true more often (`icmp eq ptr @g1, @g2` for unnamed_addr `g1` or `g2` becomes `g1 == g2 || freeze(poison)`; allows merging) and some may be okay with comparisons returning true less often (`g1 == g2 && freeze(poison)`; allows duplication). We may consider implementing this by splitting unnamed_addr in two (in a followup).

ChuanqiXu9:
+1 to @ojhunt that we should move the discussion to other places (possibly discourse).

For this particular issue, I suggest to use local_unnamed_addr.

ChuanqiXu9:
@ojhunt if you agree, I'd like to create a new PR to use local_unnamed_addr for vtable.

ojhunt:
> @ojhunt if you agree, I'd like to create a new PR to use local_unnamed_addr for vtable.

I think that would be reasonable, but make sure to cc people from this list.

ojhunt:
> @ojhunt if you agree, I'd like to create a new PR to use local_unnamed_addr for vtable.

ugh, it is going to require a gigantic amount of test updates and may require turning over many backends - if you're not already working on this I can also start the mass misery pit updates :D

efriedma-quic:
Changing from unnamed_addr to local_unnamed_addr doesn't have the desired effect, as far as I can tell: canBeOmittedFromSymbolTable() treats it exactly the same way.  I'm not sure what you're hoping to accomplish doing that.

pcc:
Yes, local_unnamed_addr means "the current translation unit considers the global to be unnamed_addr". Since we've established that that's not the case with the current meaning of unnamed_addr, we'll need to remove *unnamed_addr attributes entirely.

ojhunt:
> Changing from unnamed_addr to local_unnamed_addr doesn't have the desired effect, as far as I can tell: canBeOmittedFromSymbolTable() treats it exactly the same way. I'm not sure what you're hoping to accomplish doing that.

Ah, I had read "If the ``local_unnamed_addr`` attribute is given, the address is known to not be significant within the module" as meaning "the module can optimize on the assumption of being unnamed, but it must still keep the symbol available outside of the module (e.g. exploit that ODR means all definitions of the vtable must be identical so it can inline and reason about the local definition of the vtable, but it must still emit the vtable in the correct way to support definitions across multiple TUs).

ChuanqiXu9:
> Yes, local_unnamed_addr means "the current translation unit considers the global to be unnamed_addr". Since we've established that that's not the case with the current meaning of unnamed_addr, we'll need to remove *unnamed_addr attributes entirely.

I feel good for that. But I feel if we want to do that, we need a RFC in discourse. Would you like to write one? I feel like you understand the problem better than me.

pcc:
I was describing the current meaning of unnamed_addr. The fact that clang doesn't use unnamed_addr correctly is just a bug and shouldn't require an RFC to fix. If we wanted to change the meaning, I think that would require an RFC.

Sorry but I don't have time to work on an RFC for this, but happy to review if someone else writes one.

ojhunt:
> > Yes, local_unnamed_addr means "the current translation unit considers the global to be unnamed_addr". Since we've established that that's not the case with the current meaning of unnamed_addr, we'll need to remove *unnamed_addr attributes entirely.
> 
> I feel good for that. But I feel if we want to do that, we need a RFC in discourse. Would you like to write one? I feel like you understand the problem better than me.

Oh, I believe this is simply a bug fix - I don't think it requires an RFC, the "big change" I meant is that every test with vtables in it is probably going to need to be updated to remove the `unnamed_addr` attribute in the test expectations.

ojhunt:
(and if _really_ necessary we could add a -fgo-forth-and-copy-vtables-everywhere` flag)

ChuanqiXu9:
Sent https://github.com/llvm/llvm-project/pull/200108

ChuanqiXu9:
Prefer https://github.com/llvm/llvm-project/pull/200108

PR #197856, opened 2026-05-15
[Clang][AST] Fix ExplicitInstantiationDecl accessors for variable templates with tag types

### LINKED ISSUE #191442: [Clang] Missing AST source info for explicit template instantiations
# Related Issues:
* #21040
* #52659
* #115418
* #14230
* #21133

clangd/clangd:
* clangd/clangd#358
* clangd/clangd#316
* clangd/clangd#305
* clangd/clangd#1034
* clangd/clangd#1365

Others:
* #174715
* #64006

# Background
Clang's AST has a systematic gap in source info for explicit template instantiations. Most notably, for function templates, ActOnExplicitInstantiation returns (Decl*)nullptr after processing the instantiation, meaning the explicit instantiation statement has no node representation accessible to AST consumers. Two FIXME comments in the codebase acknowledge this:

https://github.com/llvm/llvm-project/blob/078c43c582f0cd0a7c925abcd9eefdf9f63b2040/clang/lib/Sema/SemaTemplate.cpp#L10830-L10831

https://github.com/llvm/llvm-project/blob/078c43c582f0cd0a7c925abcd9eefdf9f63b2040/clang/lib/Sema/SemaTemplate.cpp#L11068-L11069

This directly impacts tools that rely on the AST — LSP servers, refactoring tools, static analyzers — preventing them from locating explicit instantiation statements or providing features such as go-to-definition, semantic highlighting, find-references, and rename on these constructs.

# Current state

- [temp.explicit](https://eel.is/c++draft/temp.explicit) — explicit instantiation
- [temp.spec](https://eel.is/c++draft/temp.spec) — explicit specialization and explicit instantiation (umbrella section)
- [temp.inst](https://eel.is/c++draft/temp.inst) — implicit instantiation and its interaction with explicit instantiation

According to the C++ standard ([temp.explicit]), the following entities can be explicitly instantiated.

```cpp
namespace ns {
  template <typename T> void foo(T x) {}                    // (a) function template
  template <typename T> T bar = T{};                        // (b) variable template
  template <typename T> struct S {                          // (c) class template
    void method(T x) {}                                     // (d) member function
    static T sval;                                          // (e) static data member
    template <typename U> void tmpl(U u) {}                 // (f) member function template
    struct Inner { T val; };                                // (g) nested class
  };
  template <typename T> T S<T>::sval = T{};
}

// --- explicit instantiations ---
template void ns::foo<int>(int);                            // (a)
template int ns::bar<int>;                                  // (b)
template struct ns::S<int>;                                 // (c)
template void ns::S<long>::method(long);                    // (d)
template long ns::S<long>::sval;                            // (e)
template void ns::S<long>::tmpl<double>(double);            // (f)
template struct ns::S<long>::Inner;                         // (g)
```

Every explicit instantiation has two forms:

```cpp
extern template void foo<int>(int);   // explicit instantiation declaration — suppresses implicit instantiation in this TU
template void foo<int>(int);          // explicit instantiation definition — triggers instantiation
```

Within the same translation unit, for a given specialization:
- There can be multiple explicit instantiation declarations (though practically meaningless).
- There can be at most one explicit instantiation definition.
- Declarations and definitions may coexist (declaring first, defining later is a common pattern).

```cpp
// typical usage: declare in header, define in source
// header.h
extern template void foo<int>(int);

// source.cpp
template void foo<int>(int);
```

Both may also appear in the same source file:

```cpp
extern template void ns::foo<int>(int);  // declaration
extern template void ns::foo<int>(int);  // declaration
template void ns::foo<int>(int);         // definition
```

Interaction with Implicit Instantiation

An explicit instantiation may appear after an implicit instantiation:

```cpp
void use() { foo(1); }          // implicit instantiation of foo<int>
template void foo<int>(int);    // explicit instantiation — legal, changes linkage
```

This differs from explicit specialization — an explicit specialization after an implicit instantiation is ill-formed:

```cpp
void use() { foo(1); }          // implicit instantiation of foo<int>
template<> void foo<int>(int);  // error! specialization after instantiation
```

All seven kinds reuse existing specialization nodes. The table below shows what source info from the explicit instantiation statement is actually recorded on the node:

| Info | (a) func | (b) var | (c) class | (d) mem func | (e) static mem | (f) mem func tmpl | (g) nested |
|---|---|---|---|---|---|---|---|
| returned from Sema | ❌ nullptr | ✅ | ✅ | ✅ | ✅ | ❌ nullptr | ✅ |
| specialization kind | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ |
| source range | ❌ | ✅ | ✅ | ❌ | ❌ | ❌ | ❌ |
| nested specifier | ❌ | ❌ | ✅ | ❌ | ❌ | ❌ | ❌ |
| template arg locations | ❌ | ✅ | ✅ | N/A | N/A | ❌ | N/A |
| keyword locations | ❌ | ❌¹ | ✅ | ❌ | ❌ | ❌ | ❌ |
| return/var type location | ❌ | ❌ | N/A | ❌ | ❌ | ❌ | N/A |
| param locs | ❌ | N/A | N/A | ❌ | N/A | ❌ | N/A |
| multiple decls | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |

¹ While `VarTemplateSpecializationDecl` has fields for source locations, Clang's internal logic erroneously attempts a `dyn_cast` to a partial specialization when setting these locations during explicit instantiation. This cast fails, leaving the location fields permanently unset.

- ❌ in source range / NNS / param locs = the node's info points to the template definition site, not the instantiation statement.
- ❌ in "Returned from Sema" = ActOnExplicitInstantiation returns nullptr, so the Parser/consumer cannot access the node.
- ❌ in "Multiple decls" = extern template + template in the same TU overwrites the previous info on the single shared node. No redeclaration chain.

Only (c) class templates have reasonably complete info. The other six are missing critical source locations.

Here is the example code https://godbolt.org/z/W3hd5zdEG and ast dump result.

# Comparison with Explicit Specialization

Explicit specialization is handled far more completely than explicit instantiation in Clang. The contrast is instructive:

Function templates

```cpp
// explicit specialization
template<> void foo<int>(int);      // creates a new FunctionDecl, added to redecl chain
template<> void foo<int>(int) {}    // another FunctionDecl, prev → the one above

// explicit instantiation (status quo)
template void foo<int>(int);        // returns nullptr, no AST node for consumers
extern template void foo<int>(int); // returns nullptr, no AST node for consumers
```

Explicit specialization creates an independent FunctionDecl for each statement, linked via setPreviousDecl into a redeclaration chain. Each node carries correct source locations, NNS, and parameter info from its own statement.

Side-by-side comparison

| Aspect | Explicit specialization | Explicit instantiation (status quo) |
|---|---|---|
| AST node | One FunctionDecl per statement | None (returns nullptr) |
| Redecl chain | decl → def, prev linked correctly | Does not exist |
| Source range | Points to the specialization statement | N/A |
| NNS | Stored on the FunctionDecl | Lost |
| Template arg locs | Stored | Lost |
| Param locs | Specialization statement site | N/A |
| TSK | Independent per node (ExplicitSpecialization) | Single node, overwritten by subsequent statements |

Design takeaway

The explicit specialization model is a natural reference point — creating a FunctionDecl for each explicit instantiation statement and linking them via redecl chain. This aligns with C++ semantics: an explicit instantiation statement is fundamentally a declaration of the function (declaring its existence and specifying instantiation behavior).

### LINKED ISSUE #197797: [clang] `llvm_unreachable` on printing explicit instantiation declaration
Reproducer:
```cpp
struct Struct1 {};
template <typename T> T bar = T{};

extern template Struct1 bar<Struct1>;
```
Running with `-cc1 -ast-print`, the following stacktrace is produced:
```
unexpected specialization kind
UNREACHABLE executed at /home/bols/progs/llvm-project/clang/lib/AST/DeclPrinter.cpp:1374!
Stack dump:
0.      Program arguments: /home/bols/progs/llvm-project/build/bin/clang -cc1 -ast-print -xc++ -
1.      <eof> parser at end of file
 #0 0x000055789685cbc6 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /home/bols/progs/llvm-project/llvm/lib/Support/Unix/Signals.inc:880:22
 #1 0x000055789685d094 PrintStackTraceSignalHandler(void*) /home/bols/progs/llvm-project/llvm/lib/Support/Unix/Signals.inc:962:1
 #2 0x000055789685a36e llvm::sys::RunSignalHandlers() /home/bols/progs/llvm-project/llvm/lib/Support/Signals.cpp:108:20
 #3 0x000055789685c47d SignalHandler(int, siginfo_t*, void*) /home/bols/progs/llvm-project/llvm/lib/Support/Unix/Signals.inc:448:14
 #4 0x000055789685c510 SignalHandlerTerminate(int, siginfo_t*, void*) /home/bols/progs/llvm-project/llvm/lib/Support/Unix/Signals.inc:464:25
 #5 0x00007fd89b97a050 (/lib/x86_64-linux-gnu/libc.so.6+0x3c050)
 #6 0x00007fd89b9c8eec (/lib/x86_64-linux-gnu/libc.so.6+0x8aeec)
 #7 0x00007fd89b979fb2 raise (/lib/x86_64-linux-gnu/libc.so.6+0x3bfb2)
 #8 0x00007fd89b964472 abort (/lib/x86_64-linux-gnu/libc.so.6+0x26472)
 #9 0x000055789678191c bindingsErrorHandler(void*, char const*, bool) /home/bols/progs/llvm-project/llvm/lib/Support/ErrorHandling.cpp:253:55
#10 0x000055789b887d85 (anonymous namespace)::DeclPrinter::VisitExplicitInstantiationDecl(clang::ExplicitInstantiationDecl*) /home/bols/progs/llvm-project/clang/lib/AST/DeclPrinter.cpp:1376:1
#11 0x000055789b88bd5a clang::declvisitor::Base<std::add_pointer, (anonymous namespace)::DeclPrinter, void>::Visit(clang::Decl*) /home/bols/progs/llvm-project/build/tools/clang/include/clang/AST/DeclNodes.inc:672:1
#12 0x000055789b8830dd (anonymous namespace)::DeclPrinter::VisitDeclContext(clang::DeclContext*, bool) /home/bols/progs/llvm-project/clang/lib/AST/DeclPrinter.cpp:521:17
#13 0x000055789b883536 (anonymous namespace)::DeclPrinter::VisitTranslationUnitDecl(clang::TranslationUnitDecl*) /home/bols/progs/llvm-project/clang/lib/AST/DeclPrinter.cpp:578:1
#14 0x000055789b88b508 clang::declvisitor::Base<std::add_pointer, (anonymous namespace)::DeclPrinter, void>::Visit(clang::Decl*) /home/bols/progs/llvm-project/build/tools/clang/include/clang/AST/DeclNodes.inc:24:1
#15 0x000055789b881a0d clang::Decl::print(llvm::raw_ostream&, clang::PrintingPolicy const&, unsigned int, bool) const /home/bols/progs/llvm-project/clang/lib/AST/DeclPrinter.cpp:148:1
#16 0x0000557897deda6f (anonymous namespace)::ASTPrinter::print(clang::Decl*) /home/bols/progs/llvm-project/clang/lib/Frontend/ASTConsumers.cpp:101:17
#17 0x0000557897ded683 (anonymous namespace)::ASTPrinter::HandleTranslationUnit(clang::ASTContext&) /home/bols/progs/llvm-project/clang/lib/Frontend/ASTConsumers.cpp:52:23
#18 0x0000557899f72e52 clang::ParseAST(clang::Sema&, bool, bool) /home/bols/progs/llvm-project/clang/lib/Parse/ParseAST.cpp:190:14
#19 0x0000557897d4b987 clang::ASTFrontendAction::ExecuteAction() /home/bols/progs/llvm-project/clang/lib/Frontend/FrontendAction.cpp:1462:11
#20 0x0000557897d4b240 clang::FrontendAction::Execute() /home/bols/progs/llvm-project/clang/lib/Frontend/FrontendAction.cpp:1346:38
#21 0x0000557897c572ac clang::CompilerInstance::ExecuteAction(clang::FrontendAction&) /home/bols/progs/llvm-project/clang/lib/Frontend/CompilerInstance.cpp:1019:42
#22 0x0000557897f2bfc8 clang::ExecuteCompilerInvocation(clang::CompilerInstance*) /home/bols/progs/llvm-project/clang/lib/FrontendTool/ExecuteCompilerInvocation.cpp:318:38
#23 0x0000557891eebc18 cc1_main(llvm::ArrayRef<char const*>, char const*, void*) /home/bols/progs/llvm-project/clang/tools/driver/cc1_main.cpp:296:38
#24 0x0000557891ede014 ExecuteCC1Tool(llvm::SmallVectorImpl<char const*>&, llvm::ToolContext const&, llvm::IntrusiveRefCntPtr<llvm::vfs::FileSystem>) /home/bols/progs/llvm-project/clang/tools/driver/driver.cpp:229:20
#25 0x0000557891ede58e clang_main(int, char**, llvm::ToolContext const&) /home/bols/progs/llvm-project/clang/tools/driver/driver.cpp:277:26
#26 0x0000557891f10bf7 main /home/bols/progs/llvm-project/build/tools/clang/tools/driver/clang-driver.cpp:17:20
#27 0x00007fd89b96524a (/lib/x86_64-linux-gnu/libc.so.6+0x2724a)
#28 0x00007fd89b965305 __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x27305)
#29 0x0000557891edd3e1 _start (/home/bols/progs/llvm-project/build/bin/clang+0x156b3e1)
```
CC @16bit-ykiko

16bit-ykiko:
Several `ExplicitInstantiationDecl` accessors assumed that `TypeSourceInfo` always encodes a class entity, but for variable templates the `TypeSourceInfo` holds the *declared type* (e.g. `Plain`, `Color`, `Wrap<int>`). When that declared type happened to be a `TagType` or `TemplateSpecializationType`, the accessors would misinterpret it:

- `getTypeAsWritten()` returned `nullptr` → crash in `DeclPrinter`
- `getQualifierLoc()` leaked the type's qualifier into the variable name
- `getTagKWLoc()` returned the type's elaborated keyword
- `getNumTemplateArgs` / angle-loc getters extracted from the type instead of the variable's template arguments

Add a private `getClassTypeLoc()` helper that returns the entity `TypeLoc` only for class-like instantiations (`isa<RecordDecl>(getSpecialization())`), and guard all `TypeSourceInfo` fallback paths with it. For non-class entities the accessors now correctly skip the `TypeSourceInfo` and return defaults or read from trailing objects.

Also fix stale template argument source locations for variable templates: when an explicit instantiation definition follows a prior declaration, `VarTemplateSpecializationDecl::getTemplateArgsAsWritten()` returns the first declaration's locations. Construct fresh `ASTTemplateArgumentListInfo` from the current declarator instead.

Fixes #197797
Fixes https://github.com/llvm/llvm-project/issues/191442#issuecomment-4457732314

### DIFF
--- a/clang/include/clang/AST/DeclTemplate.h
+++ b/clang/include/clang/AST/DeclTemplate.h
@@ -3519,11 +3519,9 @@ public:
   /// for function / variable templates, from a trailing object.
   NestedNameSpecifierLoc getQualifierLoc() const;
 
-  /// Whether this explicit instantiation has a template argument list.
-  /// Must be checked before calling getNumTemplateArgs() and friends.
-  bool hasTemplateArgs() const;
-
-  unsigned getNumTemplateArgs() const;
+  /// Returns the number of explicit template arguments, or std::nullopt if
+  /// this entity has no template argument list (e.g., nested classes).
+  std::optional<unsigned> getNumTemplateArgs() const;
   TemplateArgumentLoc getTemplateArg(unsigned I) const;
   SourceLocation getTemplateArgsLAngleLoc() const;
   SourceLocation getTemplateArgsRAngleLoc() const;
--- a/clang/include/clang/AST/DeclTemplate.h
+++ b/clang/include/clang/AST/DeclTemplate.h
@@ -3519,6 +3519,10 @@ public:
   /// for function / variable templates, from a trailing object.
   NestedNameSpecifierLoc getQualifierLoc() const;
 
+  /// Whether this explicit instantiation has a template argument list.
+  /// Must be checked before calling getNumTemplateArgs() and friends.
+  bool hasTemplateArgs() const;
+
   unsigned getNumTemplateArgs() const;
   TemplateArgumentLoc getTemplateArg(unsigned I) const;
   SourceLocation getTemplateArgsLAngleLoc() const;
--- a/clang/include/clang/AST/DeclTemplate.h
+++ b/clang/include/clang/AST/DeclTemplate.h
@@ -3449,9 +3449,19 @@ class ExplicitInstantiationDecl final
     return hasTrailingQualifier() ? 1 : 0;
   }
 
-  /// Raw access to the internal TypeSourceInfo.  For class templates this is
-  /// a TemplateSpecializationTypeLoc; for nested classes a TagTypeLoc.
-  /// Public getTypeAsWritten() returns null for those cases.
+  /// For class templates / nested classes, returns the TypeLoc encoding the
+  /// entity (TemplateSpecializationTypeLoc or TagTypeLoc).  For function /
+  /// variable templates -- where TypeSourceInfo holds the declared type
+  /// rather than the entity -- returns std::nullopt.
+  std::optional<TypeLoc> getClassTypeLoc() const {
+    if (!isa<RecordDecl>(getSpecialization()))
+      return std::nullopt;
+    if (auto *TSI = TypeAndFlags.getPointer())
+      return TSI->getTypeLoc();
+    return std::nullopt;
+  }
+
+  /// Raw TypeSourceInfo pointer, needed by the serializer.
   TypeSourceInfo *getRawTypeSourceInfo() const {
     return TypeAndFlags.getPointer();
   }
@@ -3493,13 +3503,10 @@ public:
   SourceLocation getTemplateLoc() const { return getLocation(); }
   SourceLocation getNameLoc() const { return NameLoc; }
 
-  /// For class templates / nested classes, the tag keyword location is
-  /// stored inside TypeSourceInfo; otherwise returns an invalid location.
+  /// The tag keyword (struct/class/union) location for class templates /
+  /// nested classes; invalid for function / variable templates.
   SourceLocation getTagKWLoc() const;
 
-  /// Whether the qualifier is stored as a trailing object (function / variable
-  /// templates) rather than inside TypeSourceInfo (class templates / nested
-  /// classes).
   bool hasTrailingQualifier() const {
     return TypeAndFlags.getInt() & HasQualifierFlag;
   }
@@ -3508,24 +3515,17 @@ public:
   }
 
   /// Returns the qualifier regardless of where it is stored.
-  /// For class templates / nested classes, it is extracted from TypeSourceInfo
-  /// (TemplateSpecializationTypeLoc or TagTypeLoc).
-  /// For function / variable templates, it comes from a trailing object.
+  /// For class templates / nested classes, extracted from the class TypeLoc;
+  /// for function / variable templates, from a trailing object.
   NestedNameSpecifierLoc getQualifierLoc() const;
 
-  /// Number of explicit template arguments, regardless of storage.
-  /// For class templates they come from TemplateSpecializationTypeLoc;
-  /// for function / variable templates from trailing
-  /// ASTTemplateArgumentListInfo.
   unsigned getNumTemplateArgs() const;
   TemplateArgumentLoc getTemplateArg(unsigned I) const;
   SourceLocation getTemplateArgsLAngleLoc() const;
   SourceLocation getTemplateArgsRAngleLoc() const;
 
-  /// For function / variable templates, returns the declared type (return type
-  /// or variable type).  For class templates and nested classes returns null —
-  /// the qualifier, tag keyword, and template arguments are accessible via
-  /// getQualifierLoc(), getTagKWLoc(), and getTemplateArg().
+  /// The declared type (return type or variable type) for function / variable
+  /// templates.  Null for class templates and nested classes.
   TypeSourceInfo *getTypeAsWritten() const;
 
   TemplateSpecializationKind getTemplateSpecializationKind() const {
--- a/clang/lib/AST/DeclTemplate.cpp
+++ b/clang/lib/AST/DeclTemplate.cpp
@@ -1867,22 +1867,13 @@ TypeSourceInfo *ExplicitInstantiationDecl::getTypeAsWritten() const {
   return getRawTypeSourceInfo();
 }
 
-bool ExplicitInstantiationDecl::hasTemplateArgs() const {
-  if (getTrailingArgsInfo())
-    return true;
-  if (auto TL = getClassTypeLoc())
-    if (TL->getAs<TemplateSpecializationTypeLoc>())
-      return true;
-  return false;
-}
-
-unsigned ExplicitInstantiationDecl::getNumTemplateArgs() const {
+std::optional<unsigned> ExplicitInstantiationDecl::getNumTemplateArgs() const {
   if (const auto *Args = getTrailingArgsInfo())
     return Args->NumTemplateArgs;
   if (auto TL = getClassTypeLoc())
     if (auto TST = TL->getAs<TemplateSpecializationTypeLoc>())
       return TST.getNumArgs();
-  llvm_unreachable("template arguments not found in trailing args or TypeLoc");
+  return std::nullopt;
 }
 
 TemplateArgumentLoc
@@ -1920,7 +1911,7 @@ SourceLocation ExplicitInstantiationDecl::getEndLoc() const {
     if (TSI->getType().hasPostfixDeclaratorSyntax())
       return TSI->getTypeLoc().getEndLoc();
   // Otherwise, template args RAngleLoc or NameLoc.
-  if (hasTemplateArgs()) {
+  if (getNumTemplateArgs()) {
     SourceLocation RAngle = getTemplateArgsRAngleLoc();
     if (RAngle.isValid())
       return RAngle;
--- a/clang/lib/AST/DeclTemplate.cpp
+++ b/clang/lib/AST/DeclTemplate.cpp
@@ -1867,13 +1867,22 @@ TypeSourceInfo *ExplicitInstantiationDecl::getTypeAsWritten() const {
   return getRawTypeSourceInfo();
 }
 
+bool ExplicitInstantiationDecl::hasTemplateArgs() const {
+  if (getTrailingArgsInfo())
+    return true;
+  if (auto TL = getClassTypeLoc())
+    if (TL->getAs<TemplateSpecializationTypeLoc>())
+      return true;
+  return false;
+}
+
 unsigned ExplicitInstantiationDecl::getNumTemplateArgs() const {
   if (const auto *Args = getTrailingArgsInfo())
     return Args->NumTemplateArgs;
   if (auto TL = getClassTypeLoc())
     if (auto TST = TL->getAs<TemplateSpecializationTypeLoc>())
       return TST.getNumArgs();
-  return 0;
+  llvm_unreachable("template arguments not found in trailing args or TypeLoc");
 }
 
 TemplateArgumentLoc
@@ -1892,7 +1901,7 @@ SourceLocation ExplicitInstantiationDecl::getTemplateArgsLAngleLoc() const {
   if (auto TL = getClassTypeLoc())
     if (auto TST = TL->getAs<TemplateSpecializationTypeLoc>())
       return TST.getLAngleLoc();
-  return SourceLocation();
+  llvm_unreachable("template arguments not found in trailing args or TypeLoc");
 }
 
 SourceLocation ExplicitInstantiationDecl::getTemplateArgsRAngleLoc() const {
@@ -1901,7 +1910,7 @@ SourceLocation ExplicitInstantiationDecl::getTemplateArgsRAngleLoc() const {
   if (auto TL = getClassTypeLoc())
     if (auto TST = TL->getAs<TemplateSpecializationTypeLoc>())
       return TST.getRAngleLoc();
-  return SourceLocation();
+  llvm_unreachable("template arguments not found in trailing args or TypeLoc");
 }
 
 SourceLocation ExplicitInstantiationDecl::getEndLoc() const {
@@ -1911,8 +1920,12 @@ SourceLocation ExplicitInstantiationDecl::getEndLoc() const {
     if (TSI->getType().hasPostfixDeclaratorSyntax())
       return TSI->getTypeLoc().getEndLoc();
   // Otherwise, template args RAngleLoc or NameLoc.
-  SourceLocation RAngle = getTemplateArgsRAngleLoc();
-  return RAngle.isValid() ? RAngle : NameLoc;
+  if (hasTemplateArgs()) {
+    SourceLocation RAngle = getTemplateArgsRAngleLoc();
+    if (RAngle.isValid())
+      return RAngle;
+  }
+  return NameLoc;
 }
 
 SourceRange ExplicitInstantiationDecl::getSourceRange() const {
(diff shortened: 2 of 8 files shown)


clang/include/clang/AST/DeclTemplate.h
```diff
@@ -3449,9 +3449,19 @@ class ExplicitInstantiationDecl final
...
   }
 
-  /// Raw access to the internal TypeSourceInfo.  For class templates this is
-  /// a TemplateSpecializationTypeLoc; for nested classes a TagTypeLoc.
-  /// Public getTypeAsWritten() returns null for those cases.
+  /// For class templates / nested classes, returns the TypeLoc encoding the
+  /// entity (TemplateSpecializationTypeLoc or TagTypeLoc).  For function /
+  /// variable templates -- where TypeSourceInfo holds the declared type
+  /// rather than the entity -- returns std::nullopt.
+  std::optional<TypeLoc> getClassTypeLoc() const {
```
zwuis, line 3456: Can we return default-constructed `TypeLoc` instead of `std::nullopt`?

16bit-ykiko: I think `std::optional` is clearer.

erichkeane (commented):
This seems reasonable to me, but IIRC @mizvekov was messing in this area lately, so I'd like to get his approval be the 'final' one.

clang/lib/AST/DeclTemplate.cpp
```diff
@@ -1842,67 +1843,64 @@ ExplicitInstantiationDecl::CreateDeserialized(ASTContext &C, GlobalDeclID ID,
...
 unsigned ExplicitInstantiationDecl::getNumTemplateArgs() const {
   if (const auto *Args = getTrailingArgsInfo())
     return Args->NumTemplateArgs;
-  if (auto *TSI = getRawTypeSourceInfo())
-    if (auto TL = TSI->getTypeLoc().getAs<TemplateSpecializationTypeLoc>())
-      return TL.getNumArgs();
+  if (auto TL = getClassTypeLoc())
+    if (auto TST = TL->getAs<TemplateSpecializationTypeLoc>())
+      return TST.getNumArgs();
   return 0;
```
mizvekov: Did you consider making this case out-of-contract for this function?

Would help folks that forgot to check if this is a template entity, and not get confused between an empty template argument list, or no list at all.

clang/lib/AST/DeclTemplate.cpp
```diff
@@ -1842,67 +1843,64 @@ ExplicitInstantiationDecl::CreateDeserialized(ASTContext &C, GlobalDeclID ID,
...
 SourceLocation ExplicitInstantiationDecl::getTemplateArgsRAngleLoc() const {
   if (const auto *Args = getTrailingArgsInfo())
     return Args->getRAngleLoc();
-  if (auto *TSI = getRawTypeSourceInfo())
-    if (auto TL = TSI->getTypeLoc().getAs<TemplateSpecializationTypeLoc>())
-      return TL.getRAngleLoc();
+  if (auto TL = getClassTypeLoc())
+    if (auto TST = TL->getAs<TemplateSpecializationTypeLoc>())
+      return TST.getRAngleLoc();
   return SourceLocation();
```
mizvekov: Ditto.

clang/lib/AST/DeclTemplate.cpp
```diff
@@ -1842,67 +1843,64 @@ ExplicitInstantiationDecl::CreateDeserialized(ASTContext &C, GlobalDeclID ID,
...
 SourceLocation ExplicitInstantiationDecl::getTemplateArgsLAngleLoc() const {
   if (const auto *Args = getTrailingArgsInfo())
     return Args->getLAngleLoc();
-  if (auto *TSI = getRawTypeSourceInfo())
-    if (auto TL = TSI->getTypeLoc().getAs<TemplateSpecializationTypeLoc>())
-      return TL.getLAngleLoc();
+  if (auto TL = getClassTypeLoc())
+    if (auto TST = TL->getAs<TemplateSpecializationTypeLoc>())
+      return TST.getLAngleLoc();
   return SourceLocation();
```
mizvekov: Ditto.

mizvekov (approved):

clang/lib/AST/DeclTemplate.cpp
16bit-ykiko: Done.

kimgr:
@16bit-ykiko Hey! I'm trying out this branch as it evolves. I ran into one oddity, that I'm not sure what the intended semantics are:
```c++
// t.cc
struct V1;

template <typename T> struct S {
  struct Inner { T val; };
};

#ifdef WITH_DECL
extern template struct S<V1>::Inner;
#endif

template struct S<V1>::Inner;
```

Now running this without `WITH_DECL` defined, diagnoses on line 11, the explicit instantiation definition:
```
$ /home/kimgr/code/buildroot/16bit-ykiko/bin/clang++ -fsyntax-only t.cc
t.cc:4:20: error: field has incomplete type 'V1'
    4 |   struct Inner { T val; };
      |                    ^
t.cc:11:24: note: in instantiation of member class 'S<V1>::Inner' requested here
   11 | template struct S<V1>::Inner;
      |                        ^
t.cc:1:8: note: forward declaration of 'V1'
    1 | struct V1;
      |        ^
1 error generated.
```

Running with `WITH_DECL` defined, diagnoses on line 8, the explicit instantiation _declaration_:
```
$ /home/kimgr/code/buildroot/16bit-ykiko/bin/clang++ -DWITH_DECL -fsyntax-only t.cc
t.cc:4:20: error: field has incomplete type 'V1'
    4 |   struct Inner { T val; };
      |                    ^
t.cc:8:31: note: in instantiation of member class 'S<V1>::Inner' requested here
    8 | extern template struct S<V1>::Inner;
      |                               ^
t.cc:1:8: note: forward declaration of 'V1'
    1 | struct V1;
      |        ^
1 error generated.
```

Two things:

* Should the explicit instantiation declaration really require the complete type?
* With `WITH_DECL`, why doesn't clang diagnose on the explicit instantiation definition?

These questions are probably unrelated to this PR as such. If there's something that seems out of place, I can open a separate issue, but I thought I'd ask here since I tested based on this branch.

16bit-ykiko:
@kimgr 

> Should the explicit instantiation declaration really require the complete type?

```cpp
struct V1;

template <typename T>
struct S {
    struct Inner;
};

extern template struct S<V1>::Inner;
```

https://godbolt.org/z/7qx9xxrfv

Both Clang and GCC reject the code. Explicit instantiation nested class (whether declaration or definition) requires the target class to be complete — explicitly instantiating a class actually means instantiating the declarations of all its members, so if it is an incomplete class, it doesn't mean anything. In your example, `S<V1>::Inner` does have a definition (`struct Inner { T val; };`), but instantiating that definition requires `V1` to be complete (because of `T val;`), so the error is expected. But I haven't looked up the C++ standard draft.

> With WITH_DECL, why doesn't clang diagnose on the explicit instantiation definition?

As for the second question: once the explicit instantiation declaration has already triggered the instantiation (and the diagnostic), the definition sees an already-instantiated specialization, so it doesn't diagnose again.

@mizvekov any ideas? If there's no problem, could you help me merge this PR? Thanks!

kimgr:
@16bit-ykiko Thank you! Let's not block this PR either way. I'll play around with this a bit later.

mizvekov:
Yeah, if there is a problem here, I don't see how it relates to this patch.

PR #197859, opened 2026-05-15
[RFC][BOLT] Add a new parallel DWARF processing(2/2)

Thrrreeee:
This PR implements a new parallel DWARF debug info processing pipeline for BOLT that significantly speeds up `--update-debug-sections` for large binaries. It is the second part of the split from the overall RFC changes
RFC - [[RFC][BOLT] A New Parallel DWARF Processing Approach in BOLT](https://discourse.llvm.org/t/rfc-bolt-a-new-parallel-dwarf-processing-approach-in-bolt/90736) (The overall changes.)

This PR does the following:
1. **Equivalence-class CU partitioning:** Replaces batchsize grouping with union-find over DW_FORM_ref_addr references. Connected CUs share a bucket; isolated CUs become singletons.

> For the non-LTO case, CUs have no cross-CU dependencies, so each CU is placed into its own singleton bucket and processed fully in parallel.
> For the LTO case, CUs with cross-CU dependencies are grouped into the same bucket and processed sequentially within that bucket, while different buckets are still processed in parallel.

2. **Per-bucket parallel processing with in-order merge:** Each bucket gets its own DIEBuilder and local range/loc writers; completed buckets are then merged sequentially in partition order, applying offset fixups and emitting CUs to ensure deterministic output.
3. **Add main-binary parallel process** : Main-binary CU update and split-DWO processing both happen within the same per-bucket task.


Thrrreeee:
This testcase error will be fixed up with https://github.com/llvm/llvm-project/pull/197359

rafaelauler (approved):
Looks good! Let's wait to see if Alexander has any more input and this should be good to go, along with #197359

Thrrreeee:
cc @ayermolo , I've rebased the PR. Please review it when you have a chance?

ayermolo:
> RFC - [[RFC][BOLT] A New Parallel DWARF Processing Approach in BOLT](https://discourse.llvm.org/t/rfc-bolt-a-new-parallel-dwarf-processing-approach-in-bolt/90736) (The overall changes.)

Still please add description of what this PR does, and general algorithm you are using for splitting things for non-lto and lto cases.

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -539,43 +607,92 @@ static void emitDWOBuilder(const std::string &DWOName,
...
-        break;
     }
-    if (CrossCURefFound) {
-      Vec[0].push_back(CU.get());
-    } else {
-      ++Counter;
-      Vec.back().push_back(CU.get());
+  }
+
+  DenseMap<DWARFUnit *, std::vector<DWARFUnit *>> MembersByLeader;
```
ayermolo: Why not use SmallVector here and other places?

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -539,43 +607,92 @@ static void emitDWOBuilder(const std::string &DWOName,
...
+    for (const auto &Decl : *AbbrevSet)
+      for (const auto &Spec : Decl.attributes())
+        if (Spec.Form == dwarf::DW_FORM_ref_addr) {
+          RefAddrAbbrevs.insert(&Decl);
           break;
         }
+    if (RefAddrAbbrevs.empty())
+      continue;
+    // Track CUs involved in cross-CU references via DW_FORM_ref_addr.
+    for (const DWARFDebugInfoEntry &Entry : CU->dies()) {
```
ayermolo: This will extract all the dies into a vector. This whole thing is already a memory monster, probably best not to add to it.
Can you change this to be similar to DIEBuilder::constructFromUnit.

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -539,43 +607,92 @@ static void emitDWOBuilder(const std::string &DWOName,
...
+  Leaders.reserve(MembersByLeader.size());
+  for (auto &[Leader, Members] : MembersByLeader)
+    Leaders.push_back(Leader);
+  llvm::sort(Leaders, [&](DWARFUnit *A, DWARFUnit *B) {
+    return MembersByLeader[A].front()->getOffset() <
+           MembersByLeader[B].front()->getOffset();
+  });
+
+  // Emit cross-ref buckets, then singleton non-cross-ref CUs.
+  std::vector<std::vector<DWARFUnit *>> Vec;
```
ayermolo: General comment. 

```
static cl::opt<bool>
    PreserveOrder("dwarf-preserve-order",
                  cl::desc("Update debug information in deterministic order"),
                  cl::init(true), cl::cat(BoltCategory));
```

Is now dead. 
So for non-lto case each CU ends up in it's own bucket, unless I am missing someting.
What impact would it have on bolt peformance for non-lto, non-split-dwarf case?

ayermolo:
> cc @ayermolo , I've rebased the PR. Please review it when you have a chance?

Left some comments, but haven't had a chance to do a super deep dive.
Overall question for LTO. For builds with tons of debug info. Say 2GB+ how many buckets we end up with that have cross CU references?

bolt/lib/Rewrite/DWARFRewriter.cpp
Thrrreeee: `dwarf-preserve-order` is still useful. `cu-processing-batch-size` is now dead. After this PR merged, we will use a new pr to clean it, because many testcases need to be changed.

Thrrreeee: >So for non-lto case each CU ends up in it's own bucket, unless I am missing someting.
> What impact would it have on bolt peformance for non-lto, non-split-dwarf case?

That is right, for non-lto or split dwarf case, every bucket only contains one CU.   
BOLT's initial implementation analyzes cross-CU cases separately to enable parallelization, so we preserved this thinking.

Thrrreeee:
> Left some comments, but haven't had a chance to do a super deep dive. Overall question for LTO. For builds with tons of debug info. Say 2GB+ how many buckets we end up with that have cross CU references?

At present, almost all binaries larger than 2 GB use split DWARF. However, there is one binary of more than 1 GB. when  BOLT process, it roughly contains 1616 buckets  totally and 762 buckets have cross CU references (bucket size > 1).

| FILE SIZE % | FILE SIZE | VM SIZE % | VM SIZE | Section |
|---|---:|---:|---:|---|
| 26.8% | 480Mi | 0.0% | 0 | `.debug_info` |
| 22.8% | 407Mi | 0.0% | 0 | `.debug_str` |
| 10.9% | 195Mi | 0.0% | 0 | `.rela.debug_str_offsets` |
| 6.7% | 119Mi | 0.0% | 0 | `.debug_loclists` |

bolt/lib/Rewrite/DWARFRewriter.cpp
ayermolo: Sorry yes I meant "cu-processing-batch-size" It was late in the evening, posted wrong one. :)
I vaguely remember for no lto case bucket size made a difference.
Can we re-run the experiment to re-verify we don't loose anything by dropping batch size knob.

Thrrreeee: > Sorry yes I meant "cu-processing-batch-size" It was late in the evening, posted wrong one. :) I vaguely remember for no lto case bucket size made a difference. Can we re-run the experiment to re-verify we don't loose anything by dropping batch size knob.

We rebuild the entire parallel processing logic. The previous `cu-processing-batch-size` implementation parallelized within each bucket, meaning the batch size determined the number of concurrent threads (< DebugThreadCount). Our new implementation parallelizes across buckets and process each bucket sequentially. This means the number of tasks equals the number of buckets, and the thread count is controlled by `--debug-thread-count`.

Additionally, the previous implementation is slower because it waits after each bucket CUs processed to sequentially process the main binary CUs and run finalization, resulting in low overall concurrency.

bolt/include/bolt/Rewrite/DWARFRewriter.h
```diff
@@ -139,7 +173,19 @@ class DWARFRewriter {
...
                              CUOffsetMap &CUMap,
-                             DebugAddrWriter &FinalAddrWriter);
+                             DebugAddrWriter &FinalAddrWriter,
+                             std::vector<uint64_t> SortedCU);
+
+  /// Process CUs in main binary.
+  void processMainBinaryCU(DWARFUnit &Unit, DIEBuilder &DIEBlder,
+                           BucketLocalWriter &LocalWriter);
+
+  void processBucket(size_t Idx, SmallVector<SmallVector<DWARFUnit *>> &PartVec,
```
ayermolo, line 180: nit please add description

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -614,45 +746,144 @@ getDWONameMap(DWARFContext &DwCtx) {
...
+    return A->getOffset() < B->getOffset();
+  });
+
+  for (DWARFUnit *CU : SortedCUs) {
+    // record CUs order to make loc/loclist order correct
+    Accum.LocListCUOrder.push_back(CU->getOffset());
+    finalizeSkeletonAndStrSection(PartDIEBlder, *CU, DWOToNameMap);
+  }
+  for (DWARFUnit *CU : SortedCUs) {
+
```
ayermolo: nit extra space

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -614,45 +746,144 @@ getDWONameMap(DWARFContext &DwCtx) {
...
+      }
+    }
 
-  if (BC.isDWARF5Used()) {
-    FinalAddrWriter = std::make_unique<DebugAddrWriterDwarf5>(&BC);
-    RangeListsSectionWriter = std::make_unique<DebugRangeListsSectionWriter>();
-  } else {
-    FinalAddrWriter = std::make_unique<DebugAddrWriter>(&BC);
+    if (CU->getVersion() >= 5) {
+      uint64_t RangsOffset = RangeListsSectionWriter->getSectionOffset();
```
ayermolo: const

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -614,45 +746,144 @@ getDWONameMap(DWARFContext &DwCtx) {
...
+      auto *LocListWriter = dyn_cast<DebugLoclistWriter>(LocWriter);
+      const uint64_t BufferSize = LocWriter->getLocBufferSize();
+
+      if (LocListWriter && LocListWriter->getDwarfVersion() >= 5 &&
+          !LocListWriter->isSplitDwarf() && BufferSize != 0) {
+        LoclistsBase = Accum.LoclistsOffset;
+        Accum.LoclistsOffset += BufferSize;
+      } else if (!LocListWriter) {
+        LegacyLocBase = Accum.LegacyLocOffset;
+        // DWARF4 .debug_loc buffers carry a trailing 16-byte terminator
```
ayermolo: I don't think that's quite correct.
From what I remember for .debug_loc beginning has 16 byte of zeros to indicate "empty location"

Maybe something like:
// Each per-CU DWARF4 .debug_loc buffer begins with a 16-byte empty-list
// sentinel (offset 0). That sentinel is emitted once for the whole section
// in makeFinalLocListsSection, and each CU's leading 16 bytes are stripped,
// so a CU contributes BufferSize-16 bytes to the global section.



```
// For DWARF4, each per-CU buffer begins with a 16-byte empty list.
    // Emit the empty list only once globally (reserved above)
    if (Version == DWARFVersion::DWARFLegacy) {
      if (CurrCULocationLists->size() > 16)
        *LocStream << StringRef(
            reinterpret_cast<const char *>(CurrCULocationLists->data() + 16),
            CurrCULocationLists->size() - 16);
      continue;
    }
```

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -429,6 +455,48 @@ static bool getLowAndHighPC(const DIE &Die, const DWARFUnit &DU,
...
+    }
+  }
+
+  const auto &DIs = DIEBlder.getDIEsByUnit(Unit);
+  for (const auto &DI : DIs) {
+    DIE *Die = DI->Die;
+    DIEValue RangesVal = Die->findAttribute(dwarf::DW_AT_ranges);
+    if (RangesVal &&
+        (Version < 5 || RangesVal.getForm() == dwarf::DW_FORM_sec_offset)) {
+      uint64_t OldVal = RangesVal.getDIEInteger().getValue();
```
ayermolo: const

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -429,6 +455,48 @@ static bool getLowAndHighPC(const DIE &Die, const DWARFUnit &DU,
...
+                             uint64_t Offset) {
+  if (Offset == 0)
+    return;
+
+  const uint16_t Version = Unit.getVersion();
+  if (Version >= 5) {
+    DIE *CUDie = DIEBlder.getUnitDIEbyUnit(Unit);
+    DIEValue RngBaseVal = CUDie->findAttribute(dwarf::DW_AT_rnglists_base);
+    if (RngBaseVal) {
+      uint64_t OldVal = RngBaseVal.getDIEInteger().getValue();
```
ayermolo: const

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -429,6 +455,48 @@ static bool getLowAndHighPC(const DIE &Die, const DWARFUnit &DU,
...
+        (Version < 5 || RangesVal.getForm() == dwarf::DW_FORM_sec_offset)) {
+      uint64_t OldVal = RangesVal.getDIEInteger().getValue();
+      // Empty ranges are stored at offset 0 in both local and global
+      // buffers, so must not be shifted during merge.
+      if (OldVal != DebugRangesSectionWriter::getEmptyRangesOffset())
+        DIEBlder.replaceValue(Die, dwarf::DW_AT_ranges, RangesVal.getForm(),
+                              DIEInteger(OldVal + Offset));
+    }
+
+    if (Version < 5) {
```
ayermolo: finalizeCompileUnits has

```
DIEValue DvalGNUBase = Die->findAttribute(dwarf::DW_AT_GNU_ranges_base);
    assert(DvalGNUBase && "GNU_ranges_base attribute does not exist for DWOId");
    DIEBlder.replaceValue(
        Die, dwarf::DW_AT_GNU_ranges_base, DvalGNUBase.getForm(),
        DIEInteger(LegacyRangesSectionWriter->getSectionOffset()));
```

Can you verify if this is needed, or that one is good enough, or the other one now is not needed.

ayermolo (commented):
Slowly going through review. :)
Left some more comments

bolt/lib/Rewrite/DWARFRewriter.cpp
ayermolo: Ah ok. Miss remembered how old code worked.

bolt/lib/Rewrite/DWARFRewriter.cpp
Thrrreeee: This is not duplicate logic. `fixupDWARFRanges() `adjusts DW_AT_ranges / DW_AT_GNU_ranges_base values by a merge offset when per-bucket local ranges buffers are appended to the global buffer. `finalizeCompileUnits() `performs the deferred final patch for split DWARF: the real global .debug_ranges base is only known after the per-DWO ranges buffer is appended.（earlier code may only record the CU DIE in `updateDWARFObjectAddressRanges()` / `convertToRangesPatchDebugInfo()`  ）

bolt/lib/Rewrite/DWARFRewriter.cpp
Thrrreeee: Done

bolt/include/bolt/Rewrite/DWARFRewriter.h
```diff
@@ -40,12 +41,31 @@ class DWARFRewriter {
...
+    uint64_t LegacyLocOffset = 0;
+    std::vector<uint64_t> LocListCUOrder;
+  };
+  struct BucketLocalWriter {
+    std::unique_ptr<DebugRangeListsSectionWriter> RngListsWriter;
+    std::unique_ptr<DebugRangesSectionWriter> LegacyRangesWriter;
+
+    /// Release whichever writer was initialized.
+    void clear() {
+      if (RngListsWriter)
```
Jinjie-Huang: If statement appears to be unnecessary.

bolt/include/bolt/Rewrite/DWARFRewriter.h
```diff
@@ -40,12 +41,31 @@ class DWARFRewriter {
...
+        RngListsWriter.reset();
+      if (LegacyRangesWriter)
```
Jinjie-Huang: Ditto.

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -535,43 +603,107 @@ static void emitDWOBuilder(const std::string &DWOName,
...
-      return Vec;
-    }
-    bool CrossCURefFound = false;
-    for (const DWARFAbbreviationDeclaration &Decl : *AbbrDeclSet.get()) {
-      for (const DWARFAbbreviationDeclaration::AttributeSpec &Attr :
-           Decl.attributes()) {
-        if (Attr.Form == dwarf::DW_FORM_ref_addr) {
-          CrossCURefFound = true;
+static SmallVector<SmallVector<DWARFUnit *>> partitionCUs(DWARFContext &DwCtx) {
+  SmallVector<DWARFUnit *, 0> AllCUs;
```
Jinjie-Huang, line 566: Reserving the size should be better?

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -535,43 +603,107 @@ static void emitDWOBuilder(const std::string &DWOName,
...
+    return MembersByLeader[A].front()->getOffset() <
+           MembersByLeader[B].front()->getOffset();
+  });
+
+  // Emit cross-ref buckets, then singleton non-cross-ref CUs.
+  SmallVector<SmallVector<DWARFUnit *>> Vec;
+  for (DWARFUnit *Leader : Leaders)
+    Vec.push_back(std::move(MembersByLeader[Leader]));
+  for (DWARFUnit *CU : AllCUs) {
+    if (!CrossRefSet.count(CU)) {
```
Jinjie-Huang: nit: {} is unnecessary

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -674,86 +907,163 @@ void DWARFRewriter::updateDebugInfo() {
...
+  uint32_t CUSize = std::distance(BC.DwCtx->compile_units().begin(),
+                                  BC.DwCtx->compile_units().end());
+  // Pre-create per-CU writers in a single thread.
+  // Worker threads should only read from these maps.
+  createRangeLocListAndAddressWriters();
+  // If the user requested an output directory for rewritten .dwo files, make
+  // sure it exists before any worker attempt to create the files.
+  if (!opts::DwarfOutputPath.empty() && !sys::fs::exists(opts::DwarfOutputPath))
+    (void)sys::fs::create_directories(opts::DwarfOutputPath);
+  std::unordered_map<uint64_t, std::string> DWOToNameMap =
```
Jinjie-Huang, line 1021: Guess what mentioned [here](https://github.com/llvm/llvm-project/pull/197670#issuecomment-4523733462) is the reason we don't use DenseMap here?

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -535,43 +603,107 @@ static void emitDWOBuilder(const std::string &DWOName,
...
-          CrossCURefFound = true;
+static SmallVector<SmallVector<DWARFUnit *>> partitionCUs(DWARFContext &DwCtx) {
+  SmallVector<DWARFUnit *, 0> AllCUs;
+  for (auto &CU : DwCtx.compile_units())
+    AllCUs.push_back(CU.get());
+  if (AllCUs.empty())
+    return {};
+  auto FindCuForOffset = [&](uint64_t Offset) -> DWARFUnit * {
+    auto *It =
+        llvm::upper_bound(AllCUs, Offset, [](uint64_t Off, DWARFUnit *U) {
```
Jinjie-Huang: This assumes that AllCUs is sorted. Maybe we could add an assert?

bolt/include/bolt/Rewrite/DWARFRewriter.h
Thrrreeee: BucketLocalWriter contains two kinds of writers for DWARF4/DWARF5. Unless this bucket is a version mixed, generally only one writer will be initialized here, so this check is necessary.

bolt/lib/Rewrite/DWARFRewriter.cpp
Thrrreeee: right

bolt/lib/Rewrite/DWARFRewriter.cpp
Thrrreeee: Done. We will add a sort() to ensure  AllCUs is sorted.

bolt/lib/Rewrite/DWARFRewriter.cpp
ayermolo: DW_AT_GNU_ranges_base is DWARF4-split-only, and it points into the per-DWO ranges block — not the per-bucket LocalWriter.LegacyRangesWriter that fixupDWARFRanges() is rebasing. So OldVal + Offset here adds the wrong base. finalizeCompileUnits() then unconditionally overwrites it with the correct per-DWO global offset via setDie() + replaceValue(), and every GNU_ranges_base-bearing CU gets setDie() recorded (updateDWARFObjectAddressRanges and convertToRangesPatchDebugInfo), so the overwrite always happens.

Thrrreeee: Thank you for pointing out my mistake.  There is indeed code duplication exists.The reason this did not cause an error is that in the split DWARF scenario there is no cross-CU case; in other words, each bucket contains only one CU, so this offset always is zero and never have any effect.
I have fixed it.

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -1715,9 +2022,17 @@ void DWARFRewriter::finalizeCompileUnits(DIEBuilder &DIEBlder,
         LegacyRangesWriter->releaseBuffer();
     LegacyRangesSectionWriter->appendToRangeBuffer(*RangesWritersContents);
   }
-  DIEBlder.generateAbbrevs();
+  for (DWARFUnit *DU : CUs) {
+    DIE *UnitDIE = DIEBlder.getUnitDIEbyUnit(*DU);
+    MainDIEBuilder.generateUnitAbbrevs(UnitDIE);
+  }
+  DIEBlder.syncAbbrevTableFrom(MainDIEBuilder);
```
ayermolo: I might be missing something, but why is syncAbbrevTableFrom necessary?

Thrrreeee: Thanks, you're right that syncAbbrevTableFrom is unnecessary. This was the original implementation I tried, and I forgot to remove it when making the changes.

ayermolo:
@Thrrreeee can you do a once over your PR to make sure there are no other left over implementations from previous iterations, and no hidden overheads. Like extracting entire DIE vector, etc.

Thrrreeee:
@ayermolo 
> can you do a once over your PR to make sure there are no other left over implementations from previous iterations, and no hidden overheads. Like extracting entire DIE vector, etc.

Sure, I’ll do another thorough pass over the PR.

Thrrreeee:
@ayermolo, I reviewed it again myself and also had an AI assistant scan through it once more. There doesn't seem to be any other redundant code or hidden overheads.

ayermolo (approved):
Thanks for your patience. :)
Maybe give a day in case @Jinjie-Huang has further comments.

Thrrreeee:
> Thanks for your patience. :) Maybe give a day in case @Jinjie-Huang has further comments.

Thanks for your review, I will wait for his result.

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -610,45 +718,146 @@ getDWONameMap(DWARFContext &DwCtx) {
...
+  SortedCUs.assign(Processed.begin(), Processed.end());
+  // Ensure deterministic output by sorting CUs in section offset order.
+  llvm::sort(SortedCUs, [](const DWARFUnit *A, const DWARFUnit *B) {
+    return A->getOffset() < B->getOffset();
+  });
+
+  for (DWARFUnit *CU : SortedCUs) {
+    // record CUs order to make loc/loclist order correct
+    Accum.LocListCUOrder.push_back(CU->getOffset());
+    finalizeSkeletonAndStrSection(PartDIEBlder, *CU, DWOToNameMap);
```
Jinjie-Huang: Why is stringoffset updated during the LocsAndRanges merge process? This seems a bit misleading. `DWOToNameMapWe` will not be needed if we could move this out?

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -610,45 +718,146 @@ getDWONameMap(DWARFContext &DwCtx) {
   }
   return DWOIDToNameMap;
 }
-void DWARFRewriter::updateDebugInfo() {
-  ErrorOr<BinarySection &> DebugInfo = BC.getUniqueSectionByName(".debug_info");
-  if (!DebugInfo)
+
+void DWARFRewriter::finalizeSkeletonAndStrSection(
```
Jinjie-Huang: 'Skeleton' may be confusing, finalizeMainCUStrOffsets() seems better?

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -610,45 +718,146 @@ getDWONameMap(DWARFContext &DwCtx) {
...
+  llvm::sort(SortedCUs, [](const DWARFUnit *A, const DWARFUnit *B) {
+    return A->getOffset() < B->getOffset();
+  });
+
+  for (DWARFUnit *CU : SortedCUs) {
+    // record CUs order to make loc/loclist order correct
+    Accum.LocListCUOrder.push_back(CU->getOffset());
+    finalizeSkeletonAndStrSection(PartDIEBlder, *CU, DWOToNameMap);
+  }
+  for (DWARFUnit *CU : SortedCUs) {
```
Jinjie-Huang, line 724: Can we combine these two loops over `SortedCUs` into one? It doesn't look like there are any dependencies.

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -610,45 +718,146 @@ getDWONameMap(DWARFContext &DwCtx) {
...
+    StrOffstsWriter->finalizeSection(CU, PartDIEBlder);
+}
 
-  /// Stores and serializes information that will be put into the
-  /// .debug_addr DWARF section.
-  std::unique_ptr<DebugAddrWriter> FinalAddrWriter;
+/// Merge: compute per-CU loclist/legacy-loc bases and in the
+/// same loop apply DWARF5 rnglists/loclists base fixups and legacy
+/// .debug_loc base fixups.
+void DWARFRewriter::mergePerBucketLocsAndRanges(
```
Jinjie-Huang: It seems like this function could be better organized. Perhaps we could break mergePerBucketLocsAndRanges down into:
- processCUOrders
- processLocBases
- processRanges

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -610,45 +718,146 @@ getDWONameMap(DWARFContext &DwCtx) {
...
+    }
 
-  if (BC.isDWARF5Used()) {
-    FinalAddrWriter = std::make_unique<DebugAddrWriterDwarf5>(&BC);
-    RangeListsSectionWriter = std::make_unique<DebugRangeListsSectionWriter>();
-  } else {
-    FinalAddrWriter = std::make_unique<DebugAddrWriter>(&BC);
+    if (CU->getVersion() >= 5) {
+      const uint64_t RangsOffset = RangeListsSectionWriter->getSectionOffset();
+      fixupDWARFRanges(PartDIEBlder, *CU, RangsOffset);
```
Jinjie-Huang: Wondering why there is a separate helper function for processing ranges, but not for loc?

Jinjie-Huang:
This PR changes almost the entire pipeline, rather than being just a localized modification. The new pipeline introduces quite a bit of state with different lifetimes: global writers, per-CU writers, per-bucket writers, bucket DIEBuilders, merge offsets, debug names / GDB index state, etc. Right now, most of these are managed procedurally in updateDebugInfo(). Could we consider using RAII-style scoped objects(maybe with a separate rewriter context?) to keep these declarations grouped within their corresponding modules / phases? That would make the lifetime of each resource explicit and keep variables declared near the phase where they are used. Ideally, the flow could perhaps be reorganized into something like this:
```
void DWARFRewriter::updateDebugInfo() {
  // Step 1: Initialize DWARF writers。
  initializeWriters();
  // Step 2: Partitioning.
  partitionCUs();
  // Step 3: Process buckets in parallel.
  processCUs();
  // Step 4: Merge completed buckets.
  mergeCUPartitionResults();
  // Step 5: Emit final global debug sections and indexes.
  finalize();
  ...
}
```
But this is probably not a blocking issue, perhaps we could track this and address it in a follow-up PR? WDYT cc @ayermolo

Thrrreeee:
cc @rafaelauler , 
Could you run another round of your internal tests? Although all the changes are NFC, I'm worried something might be overlooked. If everything checks out, we'll merge it.

ayermolo:
> This PR changes almost the entire pipeline, rather than being just a localized modification. The new pipeline introduces quite a bit of state with different lifetimes: global writers, per-CU writers, per-bucket writers, bucket DIEBuilders, merge offsets, debug names / GDB index state, etc. Right now, most of these are managed procedurally in updateDebugInfo(). Could we consider using RAII-style scoped objects(maybe with a separate rewriter context?) to keep these declarations grouped within their corresponding modules / phases? That would make the lifetime of each resource explicit and keep variables declared near the phase where they are used. Ideally, the flow could perhaps be reorganized into something like this:
> 
> ```
> void DWARFRewriter::updateDebugInfo() {
>   // Step 1: Initialize DWARF writers。
>   initializeWriters();
>   // Step 2: Partitioning.
>   partitionCUs();
>   // Step 3: Process buckets in parallel.
>   processCUs();
>   // Step 4: Merge completed buckets.
>   mergeCUPartitionResults();
>   // Step 5: Emit final global debug sections and indexes.
>   finalize();
>   ...
> }
> ```
> 
> But this is probably not a blocking issue, perhaps we could track this and address it in a follow-up PR? WDYT cc @ayermolo

Yeah if we could cleanup that code (overall code, not just new) that would be great.

Jinjie-Huang (approved):
LGTM. Let's give it a few days to see if there's any other feedback.

rafaelauler:
> cc @rafaelauler , Could you run another round of your internal tests? Although all the changes are NFC, I'm worried something might be overlooked. If everything checks out, we'll merge it.

I was taking time off, but I'm back. Will check and report back.

bolt/lib/Rewrite/DWARFRewriter.cpp
```diff
@@ -610,45 +686,163 @@ getDWONameMap(DWARFContext &DwCtx) {
...
+      if (It != DWOToNameMap.end()) {
+        PartDIEBlder.updateDWONameCompDir(*StrOffstsWriter, *StrWriter, *CU,
+                                          opts::DwarfOutputPath,
+                                          StringRef(It->second));
+        if (Version >= 5 && StrOffstsWriter->isStrOffsetsSectionModified())
+          StrOffstsWriter->finalizeSection(*CU, PartDIEBlder);
+      }
+      //  split CU were just finalized above when the
+      // .debug_str_offsets section was modified;
+      return;
```
rafaelauler: return -> continue

rafaelauler (commented):
Test passes, but line 711 looks weird. Should it be "continue"?

bolt/lib/Rewrite/DWARFRewriter.cpp
Thrrreeee: Thanks for your comment. No issue here. For split DWARF, each bucket contains only a single CU, so returning early avoids redundant `finalizeSection` calls.

Thrrreeee:
cc @rafaelauler,
If you have no more problem, we will merge this PR.

bolt/lib/Rewrite/DWARFRewriter.cpp
rafaelauler: Makes sense. One thing, though. If buckets are singletons for split dwarf, and your code here relies on that invariant (and would break otherwise), why are we not asserting that? Usually we assert liberally to make it clear our intentions in the code.

assert(SortedCUs.size() == 1 && "split CU expected in singleton bucket")

This might make the life of someone debugging this later much easier, because technically you can craft DWARF that violates your assumption. We see all sorts of weird input in BOLT, coming from different toolchains. If we get broken DWARF and we output even more broken DWARF, BOLT is going to be blamed and we will need to investigate. Another thing that I don't think you noticed is that you're not really saving anything with that return. continue and return are equivalent here. So I would lean towards either asserting or converting it to a continue, but this is a nit. Feel free to commit.

rafaelauler (approved):

bolt/lib/Rewrite/DWARFRewriter.cpp
Thrrreeee: Thanks, I have fixed it. You are right, I had a slight misunderstanding earlier.

Thrrreeee:
Thanks @rafaelauler, @ayermolo, @Jinjie-Huang for the patient review and insightful comments throughout the iterations.

PR #165818, opened 2025-10-31
[SelectionDAG] Split vector types for atomic load

jofrn:
Vector types that aren't widened are split so that a single ATOMIC_LOAD is issued for the entire vector at once. This change utilizes the load vectorization infrastructure in SelectionDAG in order to group the vectors. This enables SelectionDAG to translate vectors with type bfloat,half.

PR #197860, opened 2026-05-15
[SelectionDAG] Split vector types for atomic store

jofrn:
Vector types that aren't widened are split so that a single ATOMIC_STORE
is issued for the entire vector at once. This enables SelectionDAG to
translate vectors with type bfloat,half.

Store-side counterpart to #165818. Stacked on top of #197619; and below of #197861.


llvm/lib/CodeGen/SelectionDAG/LegalizeVectorTypes.cpp
```diff
@@ -4704,6 +4707,23 @@ SDValue DAGTypeLegalizer::SplitVecOp_STORE(StoreSDNode *N, unsigned OpNo) {
...
+SDValue DAGTypeLegalizer::SplitVecOp_ATOMIC_STORE(AtomicSDNode *N) {
+  SDLoc DL(N);
+  SDValue StVal = N->getVal();
+  EVT VT = StVal.getValueType();
+
+  // Issue a single atomic store of an integer that spans the full memory
+  // width. Bitcasting the (illegal) vector value to that integer lets the
+  // type legalizer further legalize the BITCAST input as needed, while the
+  // ATOMIC_STORE itself uses only the legal integer type.
+  EVT IntVT = EVT::getIntegerVT(*DAG.getContext(), VT.getSizeInBits());
```
phoebewang, line 4752: Does it apply to <2 x x86_fp80>, <2 x fp128> etc. that no corresponding legal integer type?

jofrn: It does apply to those; however, it doesn't matter per se that there is no legal type if so since AtomicExpand.cpp's `atomicSizeSupported`'s call to `TLI->getMaxAtomicSizeInBitsSupported() / 8` will cause lowering of these types to library calls instead, i.e. before the DAG lowers them.

llvm/test/CodeGen/X86/atomic-load-store.ll
```diff
@@ -751,6 +751,456 @@ define <2 x float> @atomic_vec2_float_align(ptr %x) {
...
+; CHECK-AVX-O3-LABEL: store_atomic_vec2_bfloat:
+; CHECK-AVX-O3:       # %bb.0:
+; CHECK-AVX-O3-NEXT:    vmovd %xmm0, %eax
+; CHECK-AVX-O3-NEXT:    movl %eax, (%rdi)
+; CHECK-AVX-O3-NEXT:    retq
+;
+; CHECK-SSE-O0-LABEL: store_atomic_vec2_bfloat:
+; CHECK-SSE-O0:       # %bb.0:
+; CHECK-SSE-O0-NEXT:    subq $24, %rsp
+; CHECK-SSE-O0-NEXT:    .cfi_def_cfa_offset 32
```
arsenm: Add nounwind to functions

arsenm (approved):

RKSimon (approved):
LGTM

PR #148899, opened 2025-07-15
[X86] Cast atomic vectors in IR to support floats

jofrn:
This commit casts floats to ints in an atomic load during AtomicExpand to support
floating point types. It also is required to support 128 bit vectors in SSE/AVX.

PR #197861, opened 2026-05-15
[X86] Extend alignedstore PatFrag to cover atomic_store

jofrn:
Smaller FP vectors (`<N x half>`, `<N x bfloat>`) are left to the DAG
widen path on subtargets without native FP16/BF16 support; the
v8f16/v8bf16 bitconvert variants added to the Atomic Store Split commit's patterns let the
widened path collapse to a single instruction on AVX+ targets.

Store-side counterpart to #148899 (and now https://github.com/llvm/llvm-project/pull/199520). Stacked on top of https://github.com/llvm/llvm-project/pull/201980; and below of https://github.com/llvm/llvm-project/pull/201566.

### DIFF
--- a/llvm/lib/Target/X86/X86InstrAVX512.td
+++ b/llvm/lib/Target/X86/X86InstrAVX512.td
@@ -3356,8 +3356,9 @@ multiclass avx512_load_vl<bits<8> opc, string OpcodeStr,
 }
 
 multiclass avx512_store<bits<8> opc, string OpcodeStr, string BaseName,
-                        X86VectorVTInfo _, PatFrag st_frag, PatFrag mstore,
-                        X86SchedWriteMoveLS Sched, bit NoMRPattern = 0> {
+                        X86VectorVTInfo _, SDPatternOperator st_frag,
+                        PatFrag mstore, X86SchedWriteMoveLS Sched,
+                        bit NoMRPattern = 0> {
   let hasSideEffects = 0, isCodeGenOnly = 1, ForceDisassemble = 1 in {
   let isMoveReg = 1 in
   def rr_REV  : AVX512PI<opc, MRMDestReg, (outs _.RC:$dst), (ins _.RC:$src),
--- a/llvm/lib/Target/X86/X86InstrFragmentsSIMD.td
+++ b/llvm/lib/Target/X86/X86InstrFragmentsSIMD.td
@@ -1254,9 +1254,11 @@ def extloadv4f16 : PatFrag<(ops node:$ptr), (extloadvf16 node:$ptr)>;
 def extloadv8f16 : PatFrag<(ops node:$ptr), (extloadvf16 node:$ptr)>;
 def extloadv16f16 : PatFrag<(ops node:$ptr), (extloadvf16 node:$ptr)>;
 
-// Like 'store', but always requires vector size alignment.
+// Like 'store' or 'atomic_store', but always requires vector size alignment.
 class alignedstore_frag<int Alignment>
-    : PatFrag<(ops node:$val, node:$ptr), (store node:$val, node:$ptr)> {
+    : PatFrags<(ops node:$val, node:$ptr),
+               [(store node:$val, node:$ptr),
+                (atomic_store node:$val, node:$ptr)]> {
   let IsStore = true;
   let MinAlignment = Alignment;
 }
--- a/llvm/lib/Target/X86/X86InstrFragmentsSIMD.td
+++ b/llvm/lib/Target/X86/X86InstrFragmentsSIMD.td
@@ -1254,10 +1254,11 @@ def extloadv4f16 : PatFrag<(ops node:$ptr), (extloadvf16 node:$ptr)>;
 def extloadv8f16 : PatFrag<(ops node:$ptr), (extloadvf16 node:$ptr)>;
 def extloadv16f16 : PatFrag<(ops node:$ptr), (extloadvf16 node:$ptr)>;
 
-// Like 'store', but always requires vector size alignment.
-def alignedstore : PatFrag<(ops node:$val, node:$ptr),
-                           (store node:$val, node:$ptr), [{
-  auto *St = cast<StoreSDNode>(N);
+// Like 'store' or 'atomic_store', but always requires vector size alignment.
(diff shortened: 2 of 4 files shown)


llvm/lib/Target/X86/X86InstrCompiler.td
```diff
@@ -1242,6 +1242,21 @@ def : Pat<(v4i32 (atomic_load_128_v4i32 addr:$src)),
 def : Pat<(v4i32 (atomic_load_128_v4i32 addr:$src)),
           (VMOVAPDZ128rm addr:$src)>, Requires<[HasAVX512]>;
 
+// store atomic <2 x i64>
```
arsenm: This shouldn't require duplicating all of these patterns. There surely must be existing store patterns that can be shared?

llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -32967,6 +32967,19 @@ X86TargetLowering::shouldCastAtomicLoadInIR(LoadInst *LI) const {
...
 }
 
+TargetLowering::AtomicExpansionKind
+X86TargetLowering::shouldCastAtomicStoreInIR(StoreInst *SI) const {
+  Type *Ty = SI->getValueOperand()->getType();
+  if (!Ty->getScalarType()->isFloatingPointTy())
+    return AtomicExpansionKind::None;
+  // Sub-128-bit FP vectors codegen better when DAG widening folds the value
+  // into an extractelt-from-XMM pattern, instead of an IR-level bitcast to a
+  // scalar integer (which the type legalizer scalarizes).
```
arsenm: I think this is thinking too much about this. The goal is to fully remove shouldCastAtomicStoreInIR

llvm/lib/Target/X86/X86InstrCompiler.td
jofrn: Right, I was wondering the same. I'll take a closer look. Thanks.

llvm/lib/Target/X86/X86ISelLowering.cpp
jofrn: Mm hm, I can't help yet utilize it. Let me look at this one more closely alsos. thank ya.

llvm/lib/Target/X86/X86InstrCompiler.td
arsenm: I expect you to need a PatFrag common to store + atomic_store, and update the patterns to use that

llvm/lib/Target/X86/X86ISelLowering.cpp
jofrn: Alright, done.

llvm/lib/Target/X86/X86InstrCompiler.td
jofrn: Did this one. Thanks.

RKSimon (commented):
Do you have ALIGNED xmm/ymm/zmm atomic stores? I can't see any in this version of atomic-load-store.ll 

e.g. all I can find is `store atomic <8 x double> %v, ptr %x release, align 4`

llvm/lib/Target/X86/X86InstrAVX512.td
```diff
@@ -3356,7 +3356,7 @@ multiclass avx512_load_vl<bits<8> opc, string OpcodeStr,
 }
 
 multiclass avx512_store<bits<8> opc, string OpcodeStr, string BaseName,
-                        X86VectorVTInfo _, PatFrag st_frag, PatFrag mstore,
+                        X86VectorVTInfo _, PatFrags st_frag, PatFrag mstore,
```
arsenm: Usually this would be SDPatternOperator

RKSimon (commented):
Do we have any aligned vector stores wider that 128-bits?

jofrn:
> Do we have any aligned vector stores wider that 128-bits?

Those will be translated to library calls as per the X86TargetLowering::X86TargetLowering ctor.

RKSimon:
> > Do we have any aligned vector stores wider that 128-bits?
> 
> Those will be translated to library calls as per the X86TargetLowering::X86TargetLowering ctor.

And we do have test coverage to show them doing that?

jofrn:
> > > Do we have any aligned vector stores wider that 128-bits?
> > 
> > 
> > Those will be translated to library calls as per the X86TargetLowering::X86TargetLowering ctor.
> 
> And we do have test coverage to show them doing that?

Nope; and I believe we are missing them for loads as well. Perhaps they can be separate commits.

llvm/lib/Target/X86/X86InstrFragmentsSIMD.td
```diff
@@ -1254,10 +1254,11 @@ def extloadv4f16 : PatFrag<(ops node:$ptr), (extloadvf16 node:$ptr)>;
...
-                           (store node:$val, node:$ptr), [{
-  auto *St = cast<StoreSDNode>(N);
+// Like 'store' or 'atomic_store', but always requires vector size alignment.
+def alignedstore : PatFrags<(ops node:$val, node:$ptr),
+                            [(store node:$val, node:$ptr),
+                             (atomic_store node:$val, node:$ptr)], [{
+  auto *St = cast<MemSDNode>(N);
   return St->getAlign() >= St->getMemoryVT().getStoreSize();
 }]> {
   let GISelPredicateCode = [{
```
arsenm: These custom alignment predicates should be removed. This predates having a dedicated MinAlignment field on PatFrags. This would also provide DAG + gisel support without maintaining these 2 custom predicates

jofrn: Alright, let me go about seeking how to implement this. Without the frag, alignedstore or MinAlignment; for instance, vmovaps is selected for in AVX checks, and it would not be otherwise. Thanks.

RKSimon (approved):
LGTM

llvm/lib/Target/X86/X86InstrFragmentsSIMD.td
jofrn: https://github.com/llvm/llvm-project/pull/201980 contains this.

PR #197862, opened 2026-05-15
[AtomicExpand] Add bitcasts when expanding store atomic vector

jofrn:
AtomicExpand fails for aligned `store atomic <n x T>` because it
does not find a compatible library call. This change adds appropriate
ptrtoint + bitcast so that the call can be lowered, mirroring the
load-side handling.

Store-side counterpart to #148900. Stacked on top of https://github.com/llvm/llvm-project/pull/201566.

### DIFF
--- a/llvm/lib/CodeGen/AtomicExpandPass.cpp
+++ b/llvm/lib/CodeGen/AtomicExpandPass.cpp
@@ -711,7 +711,9 @@ StoreInst *AtomicExpandImpl::convertAtomicStoreToIntegerType(StoreInst *SI) {
   auto *M = SI->getModule();
   Type *NewTy = getCorrespondingIntegerType(SI->getValueOperand()->getType(),
                                             M->getDataLayout());
-  Value *NewVal = Builder.CreateBitCast(SI->getValueOperand(), NewTy);
+  Value *NewVal = SI->getValueOperand()->getType()->isPtrOrPtrVectorTy()
+                      ? Builder.CreatePtrToInt(SI->getValueOperand(), NewTy)
+                      : Builder.CreateBitCast(SI->getValueOperand(), NewTy);
 
   Value *Addr = SI->getPointerOperand();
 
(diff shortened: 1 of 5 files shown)


llvm/lib/CodeGen/AtomicExpandPass.cpp
```diff
@@ -2174,8 +2176,19 @@ bool AtomicExpandImpl::expandAtomicOpToLibcall(
...
+      // scalar integer type. Mirror of the load-side handling below.
+      Value *IntValue;
+      auto *VPtrTy =
+          dyn_cast<PointerType>(ValueOperand->getType()->getScalarType());
+      auto *VTy = dyn_cast<VectorType>(ValueOperand->getType());
+      if (VTy && VPtrTy && !SizedIntTy->isVectorTy()) {
+        unsigned AS = VPtrTy->getAddressSpace();
+        Value *PtrInt = Builder.CreatePtrToInt(
+            ValueOperand, VTy->getWithNewType(DL.getIntPtrType(Ctx, AS)));
+        IntValue = Builder.CreateBitCast(PtrInt, SizedIntTy);
```
antoniofrighetto: Might be worth to refactor this new code alongside the load-side below into dedicated helpers (e.g., `castVecOfPtrToInt`/`castIntToVecOfPtr` and maybe a shared `needsPtrVecCast` or something similar).

llvm/lib/CodeGen/AtomicExpandPass.cpp
```diff
@@ -2174,8 +2176,19 @@ bool AtomicExpandImpl::expandAtomicOpToLibcall(
   // 'val' argument ('desired' for cas), if present.
   if (ValueOperand) {
     if (UseSizedLibcall) {
-      Value *IntValue =
-          Builder.CreateBitOrPointerCast(ValueOperand, SizedIntTy);
+      // Add casts from ValueOperand's <n x ptr> vector type to Result's
+      // scalar integer type. Mirror of the load-side handling below.
+      Value *IntValue;
+      auto *VPtrTy =
```
arsenm: Can this be replaced with Builder.CreateBitPreservingCastChain

llvm/lib/CodeGen/AtomicExpandPass.cpp
jofrn: Thanks! I shall do so: https://github.com/llvm/llvm-project/pull/197862#discussion_r3247311169.

llvm/lib/CodeGen/AtomicExpandPass.cpp
jofrn: Yes, that worked! We'll need to do this for the `load atomic` code alsos.

arsenm (approved):

RKSimon (approved):
LGTM - I'm not going to ask you to mess with your patch stack, but the atomic-load-store.ll test files are getting pretty bulky - its getting harder to check for specific type/alignment test coverage

PR #197863, opened 2026-05-15
[clang][Sema] Diagnose nested local classes defined in a different block scope than their parent

### LINKED ISSUE #193472: Clang accepts definiton of a local nested class in a non-enclosing block scope
Consider the following example (https://godbolt.org/z/711znWos9):
```cpp
void f() {
  class A { class B; };
  {
    class A::B {};
  } 
}
```

I believe that [[class.local]/3](https://eel.is/c++draft/class.local#3) is clear that this is not allowed (emphasis mine):
> A class nested within a local class is a local class[.](https://eel.is/c++draft/class.local#3.sentence-1) A member of a local class `X` shall be declared only in the definition of `X` or, if the member is a nested class, __in the nearest enclosing block scope of X__[.](https://eel.is/c++draft/class.local#3.sentence-2)

EDG is the only implementation that rejects this.

---

If we disagree and think this is ought to be accepted, I can imagine CWG being receptive to that, given how resolution to [CWG2890](https://cplusplus.github.io/CWG/issues/2890.html) evolved. However, we should be careful (https://godbolt.org/z/1MGvjYdeK):
```cpp
void f() {
  struct A {
    struct B;
    void f() {
      class A::B {}; // all reject
    }
  };
  {
    class A::B {}; // only EDG rejects at the moment; we accept and we would be insisting on it
  }
  []{
    class A::B {}; // all reject
  };
}

```

localspook:
Fixes #193472.

[[class.local]/3](https://eel.is/c++draft/class.local#3) says:

> A class nested within a local class is a local class. A member of a local class `X` shall be declared only in the definition of `X` or, __if the member is a nested class, in the nearest enclosing block scope of X__.

In other words:

```cpp
void f() {
  struct X { struct S; };
  struct X::S {}; // okay

  struct X { struct S; };
  {
    struct X::S {}; // error
  }
}
```

To save anyone from a misconception that confused me for a bit: the body of a class is not a block scope! [[basic.scope.block]/1](https://eel.is/c++draft/basic.scope.block#1). So the following example is well-formed, because `A`, `X`, and `S` are all in the same block scope (`f`'s *compound-statement*):

```cpp
void f() {
  struct A { struct X { struct S; }; };
  struct A::X::S {}; // okay
}
```

### DIFF
--- a/clang/docs/ReleaseNotes.rst
+++ b/clang/docs/ReleaseNotes.rst
@@ -54,7 +54,7 @@ C++ Specific Potentially Breaking Changes
   matching the deduction of array sizes from ``int(&)[N]``.
   This is a breaking change for code that depended on the previously deduced type. (#GH195033)
 
-- Clang now correctly rejects nested local classes defined in a different
+- Clang now rejects nested local classes defined in a different
   block scope than their parent class. (#GH193472)
 
 ABI Changes in This Version
--- a/clang/docs/ReleaseNotes.rst
+++ b/clang/docs/ReleaseNotes.rst
@@ -54,6 +54,9 @@ C++ Specific Potentially Breaking Changes
   matching the deduction of array sizes from ``int(&)[N]``.
   This is a breaking change for code that depended on the previously deduced type. (#GH195033)
 
+- Clang now correctly rejects nested local classes defined in a different
+  block scope than their parent class. (#GH193472)
+
 ABI Changes in This Version
 ---------------------------
 
--- a/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -10268,8 +10268,7 @@ def err_reference_to_local_in_enclosing_context : Error<
   "reference to local %select{variable|binding}1 %0 declared in enclosing "
   "%select{%3|block literal|lambda expression|context}2">;
 def err_local_nested_class_invalid_scope : Error<
-  "nested local class %0 must be defined in the same block scope as "
-  "its parent class %1">;
+  "nested local class %0 must be defined in the same block scope as %1">;
 def err_capture_binding_openmp : Error<
   "capturing a structured binding is not yet supported in OpenMP">;
 def ext_capture_binding : ExtWarn<
--- a/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -10267,6 +10267,9 @@ def ext_ms_anonymous_record : ExtWarn<
 def err_reference_to_local_in_enclosing_context : Error<
   "reference to local %select{variable|binding}1 %0 declared in enclosing "
   "%select{%3|block literal|lambda expression|context}2">;
+def err_local_nested_class_invalid_scope : Error<
+  "nested local class %0 must be defined in the same block scope as "
+  "its parent class %1">;
 def err_capture_binding_openmp : Error<
   "capturing a structured binding is not yet supported in OpenMP">;
 def ext_capture_binding : ExtWarn<
--- a/clang/lib/Sema/SemaDecl.cpp
+++ b/clang/lib/Sema/SemaDecl.cpp
@@ -18537,7 +18537,7 @@ Sema::ActOnTag(Scope *S, unsigned TagSpec, TagUseKind TUK, SourceLocation KWLoc,
               Diag(NameLoc, diag::err_local_nested_class_invalid_scope)
                   << Name << OutermostClass;
               Diag(OutermostClass->getLocation(), diag::note_previous_decl)
-                  << OutermostClass << OutermostClass->getLocation();
+                  << OutermostClass;
             }
           }
         }
--- a/clang/lib/Sema/SemaDecl.cpp
+++ b/clang/lib/Sema/SemaDecl.cpp
@@ -18536,7 +18536,7 @@ Sema::ActOnTag(Scope *S, unsigned TagSpec, TagUseKind TUK, SourceLocation KWLoc,
                 !S->isDeclScope(OutermostClass)) {
               Diag(NameLoc, diag::err_local_nested_class_invalid_scope)
                   << Name << OutermostClass;
-              Diag(OutermostClass->getLocation(), diag::note_previous_decl)
+              Diag(OutermostClass->getLocation(), diag::note_defined_here)
                   << OutermostClass;
             }
           }
--- a/clang/lib/Sema/SemaDecl.cpp
+++ b/clang/lib/Sema/SemaDecl.cpp
@@ -18521,6 +18521,27 @@ Sema::ActOnTag(Scope *S, unsigned TagSpec, TagUseKind TUK, SourceLocation KWLoc,
           Diag(PrevTagDecl->getLocation(), diag::note_previous_declaration);
         }
 
+        // C++ [class.local]p3:
+        //   A class nested within a local class is a local class. A member of
+        //   a local class X shall be declared only in the definition of X or,
+        //   if the member is a nested class, in the nearest enclosing block
+        //   scope of X.
+        if (TUK == TagUseKind::Definition && SS.isValid()) {
+          if (const auto *OutermostClass = dyn_cast<CXXRecordDecl>(PrevDecl)) {
+            while (const auto *ParentClass =
+                       dyn_cast<CXXRecordDecl>(OutermostClass->getParent()))
+              OutermostClass = ParentClass;
+
+            if (OutermostClass->isLocalClass() &&
+                !S->isDeclScope(OutermostClass)) {
+              Diag(NameLoc, diag::err_local_nested_class_invalid_scope)
+                  << Name << OutermostClass;
+              Diag(OutermostClass->getLocation(), diag::note_previous_decl)
+                  << OutermostClass << OutermostClass->getLocation();
+            }
+          }
+        }
+
         if (!Invalid) {
           // If this is a use, just return the declaration we found, unless
           // we have attributes.
--- a/clang/test/CXX/class/class.local/p3.cpp
+++ b/clang/test/CXX/class/class.local/p3.cpp
@@ -131,7 +131,9 @@ auto f7 = [] {
 
 struct A {
   void f() {
-    struct B { struct C; };
+    struct B { struct C; }; // #B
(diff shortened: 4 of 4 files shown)


clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -10267,6 +10267,9 @@ def ext_ms_anonymous_record : ExtWarn<
 def err_reference_to_local_in_enclosing_context : Error<
   "reference to local %select{variable|binding}1 %0 declared in enclosing "
   "%select{%3|block literal|lambda expression|context}2">;
+def err_local_nested_class_invalid_scope : Error<
+  "nested local class %0 must be defined in the same block scope as "
```
Endilll: This message doesn't work when "parent" class is itself a nested class, because it's not necessarily defined at a block scope. I think it might be more useful to say "must be defined in the same block scope as %1", where %1 is the outermost local class.

clang/test/CXX/class/class.local/p3.cpp
```diff
@@ -1,30 +1,91 @@
...
-
-// A class nested within a local class is a local class.
-void f3(int a) { // expected-note{{'a' declared here}}
-  struct X {
-    struct Y {
-      int f() { return a; } // expected-error{{reference to local variable 'a' declared in enclosing function 'f3'}}
-    };
-  };
-}
+// RUN: %clang_cc1 -fsyntax-only -verify %s
```
Endilll: Can you switch this test to `-verify-directives`? In general I think this test would benefit from C++ DR test conventions, like `@-1`.

clang/lib/Sema/SemaDecl.cpp
```diff
@@ -18521,6 +18521,27 @@ Sema::ActOnTag(Scope *S, unsigned TagSpec, TagUseKind TUK, SourceLocation KWLoc,
...
+          if (const auto *OutermostClass = dyn_cast<CXXRecordDecl>(PrevDecl)) {
+            while (const auto *ParentClass =
+                       dyn_cast<CXXRecordDecl>(OutermostClass->getParent()))
+              OutermostClass = ParentClass;
+
+            if (OutermostClass->isLocalClass() &&
+                !S->isDeclScope(OutermostClass)) {
+              Diag(NameLoc, diag::err_local_nested_class_invalid_scope)
+                  << Name << OutermostClass;
+              Diag(OutermostClass->getLocation(), diag::note_previous_decl)
```
cor3ntin:
```suggestion
              Diag(OutermostClass->getLocation(), diag::note_defined_here)
```

clang/include/clang/Basic/DiagnosticSemaKinds.td
localspook: I want to say *something* in the diagnostic message about the relationship between %0 and %1 though. So maybe:
> %0 must be defined in the same block scope as its **enclosing** class %1

?

Sirraide: ‘as its outermost parent class %1’ maybe but that’s a bit verbose

Sirraide: Actually, that’s incorrect if you have e.g.
```c++
struct A {
    void f() {
        struct B { struct C; };
        { struct B::C {}; }
    }
};
```
On that note, can we add this as a test case?

clang/test/CXX/class/class.local/p3.cpp
```diff
@@ -1,30 +1,111 @@
...
-// A class nested within a local class is a local class.
-void f3(int a) { // expected-note{{'a' declared here}}
-  struct X {
-    struct Y {
-      int f() { return a; } // expected-error{{reference to local variable 'a' declared in enclosing function 'f3'}}
-    };
-  };
-}
+// RUN: %clang_cc1 -fsyntax-only -verify -verify-directives %s
+
```
Sirraide, line 2: Another test case:
```c++
decltype(auto) f() {
    struct S { struct A; };
    return S();
}

using X = decltype(f());
struct X::A {};
```
(As an aside, GCC currently crashes on this but there is already a bug for that: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88003)

clang/include/clang/Basic/DiagnosticSemaKinds.td
Endilll: I would like to avoid "parent class", as "parent" is not defined in the Standard and feels more like an our internal term than something users would be familiar with.

> %0 must be defined in the same block scope as its enclosing class %1

This has the same downside as your original wording I commented about.

> ‘as its outermost parent class %1’ maybe but that’s a bit verbose

We can do `%0 must be defined in the same block scope as its %select{|outermost} enclosing class %1`, where selection expands only in cases when (immediately) enclosing class in not the outermost enclosing class. I agree that this is a bit verbose, but I suspect this will only be seen in compiler test suites.

> On that note, can we add this as a test case?

I believe this is covered by the test case at `p3.cpp:49`.

Endilll: Sorry, your test case is actually ill-formed and is on line 69 in `p3.cpp`

Sirraide: > I believe this is covered by the test case at `p3.cpp:49`.

No in that one the outer class isn’t inside (a function inside of) another class

Sirraide: > Sorry, your test case is actually ill-formed and is on line 69 in `p3.cpp`

I don’t think it’s that one either because `A` is not inside another class

Endilll: > I don’t think it’s that one either because A is not inside another class

Hmm, but it doesn't matter for the wording. If we want to ensure we don't reach to the classes that contain the innermost block scope, which makes some sense to me, I'd suggest converting all test cases from functions at namespace scope into member functions defined inline.

Endilll: Also, it would be interesting to check this within a lambda — to make sure we don't hit the closure type.

localspook: > On that note, can we add this as a test case?

~~It's already there at line 59 (okay okay, I realize that's the *third* line number that's been said, but it really is there :P )~~

Ah, I missed the extra braces, nevermind

Endilll: This discussion got sidetracked by the additional test cases, but my suggestion for the diagnostic wording still stands: https://github.com/llvm/llvm-project/pull/197863/changes#r3255391923

Endilll: Sorry, I guess I dropped a bad link to the comment. Hopefully this works better: https://github.com/llvm/llvm-project/pull/197863#discussion_r3255391923, which suggested `%0 must be defined in the same block scope as its %select{|outermost} enclosing class %1` after the discussion we had here.

localspook: I believe that version still has the problem @Sirraide mentioned:

> Actually, that’s incorrect if you have e.g.
> ```cpp
> struct A {
>     void f() {
>         struct B { struct C; };
>         { struct B::C {}; }
>     }
> };
> ```

I guess the detailed message is more trouble than it's worth in this case... I've changed it to the simpler message you suggested in the very first comment.

clang/docs/ReleaseNotes.rst
```diff
@@ -54,6 +54,9 @@ C++ Specific Potentially Breaking Changes
   matching the deduction of array sizes from ``int(&)[N]``.
   This is a breaking change for code that depended on the previously deduced type. (#GH195033)
 
+- Clang now correctly rejects nested local classes defined in a different
```
Endilll:
```suggestion
- Clang now rejects nested local classes defined in a different
```

clang/test/CXX/class/class.local/p3.cpp
```diff
@@ -1,30 +1,138 @@
...
+  };
+}
+
+// A class nested within a local class is a local class.
+void f3(int a) { // #a-param
+  struct X {
+    struct Y {
+      int f() { return a; }
+      // expected-error@-1 {{reference to local variable 'a' declared in enclosing function 'f3'}}
+       // expected-note@#a-param {{'a' declared here}}
```
Endilll:
```suggestion
      //   expected-note@#a-param {{'a' declared here}}
```

Endilll (approved):
LGTM modulo couple of nits, but wait for another approval.

clang/lib/Sema/SemaDecl.cpp
```diff
@@ -18521,6 +18521,27 @@ Sema::ActOnTag(Scope *S, unsigned TagSpec, TagUseKind TUK, SourceLocation KWLoc,
...
+        // C++ [class.local]p3:
+        //   A class nested within a local class is a local class. A member of
+        //   a local class X shall be declared only in the definition of X or,
+        //   if the member is a nested class, in the nearest enclosing block
+        //   scope of X.
+        if (TUK == TagUseKind::Definition && SS.isValid()) {
+          if (const auto *OutermostClass = dyn_cast<CXXRecordDecl>(PrevDecl)) {
+            while (const auto *ParentClass =
+                       dyn_cast<CXXRecordDecl>(OutermostClass->getParent()))
+              OutermostClass = ParentClass;
```
Sirraide, line 18533:
```suggestion
            while (const auto *ParentCtx =
                       dyn_cast<CXXRecordDecl>(OutermostClass->getParent()))
              OutermostClass = ParentCtx;
```
Nit: can you rename this, because it took me a moment to realise just now that `getParent()` returns the parent `DeclContext` and that it isn’t specific to `RecordDecl`/`CXXRecordDecl`.

Sirraide (approved):

clang/lib/Sema/SemaDecl.cpp
Sirraide: Well, actually this is after the cast so I guess `ParentClass` is fine as a name—I suppose you just have to know that `getParent()` returns a `DeclContext`.

localspook: > Well, actually this is after the cast so I guess `ParentClass` is fine as a name—I suppose you just have to know that `getParent()` returns a `DeclContext`.

With the cast I think it's okay to leave as is, yeah

PR #197864, opened 2026-05-15
[LifetimeSafety] Fix false negative for GSL Owner methods inherited from a non-Owner base

### LINKED ISSUE #188832: [LifetimeSafety] False-negative with dereferencing std::optional of libc++
```cpp
#include <optional>
#include <string>

const std::string& test_opt_strings() {
    std::optional<std::string> strings_or;
    return *strings_or; // no error about return stack address.
}
```

https://godbolt.org/z/YYch9h3Te

This works fine without libc++ https://godbolt.org/z/MaY39G1Pd

This is a recent regression as this was ok before a weeks.

cc: @aeft as this might be related to your recent change to std::optional ? tnx!

aeft:
- Take the implicit object's actual type (e.g., the type before any `DerivedToBase` cast) into account when checking for GSL Owner. Other `isGslOwnerType` call sites with the same pattern (e.g., `isGslOwnerType(MCE->getImplicitObjectArgument()->getType())` in `VisitCXXMemberCallExpr`) lack a real-world trigger today and are deferred to a follow-up.
- Unify the GSL Owner checks inside `shouldTrackImplicitObjectArg` so they share a single source of truth.

Fixes: #188832

### DIFF
--- a/clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeAnnotations.h
+++ b/clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeAnnotations.h
@@ -57,7 +57,8 @@ bool implicitObjectParamIsLifetimeBound(const FunctionDecl *FD);
 // container iterators (begin, end), data accessors (c_str, data, get),
 // element accessors (operator[], operator*, front, back, at), or propagating
 // operations (operator+, operator-, operator++, operator--).
-bool shouldTrackImplicitObjectArg(const CXXMethodDecl *Callee,
+bool shouldTrackImplicitObjectArg(const Expr &ImplicitObjectArgument,
+                                  const CXXMethodDecl *Callee,
                                   bool RunningUnderLifetimeSafety);
 
 // Returns true if the first argument of a free function should be tracked for
@@ -76,6 +77,7 @@ bool shouldTrackSecondArgument(const FunctionDecl *FD);
 bool isGslPointerType(QualType QT);
 // Tells whether the type is annotated with [[gsl::Owner]].
 bool isGslOwnerType(QualType QT);
+bool isGslOwnerType(const CXXRecordDecl *RD);
 
 // Returns true if the given method is std::unique_ptr::release().
 // This is treated as a move in lifetime analysis to avoid false-positives
--- a/clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp
+++ b/clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp
@@ -882,7 +882,7 @@ void FactsGenerator::handleFunctionCall(const Expr *Call,
   handleImplicitObjectFieldUses(Call, FD);
   if (!CallList)
     return;
-  auto IsArgLifetimeBound = [FD](unsigned I) -> bool {
+  auto IsArgLifetimeBound = [FD, &Args](unsigned I) -> bool {
     const ParmVarDecl *PVD = nullptr;
     if (const auto *Method = dyn_cast<CXXMethodDecl>(FD);
         Method && Method->isInstance() && !isa<CXXConstructorDecl>(FD)) {
@@ -890,7 +890,7 @@ void FactsGenerator::handleFunctionCall(const Expr *Call,
         // For the 'this' argument, the attribute is on the method itself.
         return implicitObjectParamIsLifetimeBound(Method) ||
                shouldTrackImplicitObjectArg(
-                   Method, /*RunningUnderLifetimeSafety=*/true);
+                   *Args[0], Method, /*RunningUnderLifetimeSafety=*/true);
       if ((I - 1) < Method->getNumParams())
         // For explicit arguments, find the corresponding parameter
         // declaration.
@@ -905,13 +905,13 @@ void FactsGenerator::handleFunctionCall(const Expr *Call,
     }
     return PVD ? PVD->hasAttr<clang::LifetimeBoundAttr>() : false;
   };
-  auto shouldTrackPointerImplicitObjectArg = [FD](unsigned I) -> bool {
+  auto shouldTrackPointerImplicitObjectArg = [FD, &Args](unsigned I) -> bool {
     const auto *Method = dyn_cast<CXXMethodDecl>(FD);
     if (!Method || !Method->isInstance())
       return false;
     return I == 0 &&
            isGslPointerType(Method->getFunctionObjectParameterType()) &&
-           shouldTrackImplicitObjectArg(Method,
+           shouldTrackImplicitObjectArg(*Args[0], Method,
                                         /*RunningUnderLifetimeSafety=*/true);
   };
   if (Args.empty())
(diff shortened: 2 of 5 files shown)


aeft:
* **#197864** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197864?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/llvm?ref=merge-queue-instructions-in-stack-comment&prId=197864).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

aeft:
Happy to expand the scope if you prefer (e.g., handling the other `isGslOwnerType` call sites).

clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeAnnotations.h
```diff
@@ -57,8 +57,13 @@ bool implicitObjectParamIsLifetimeBound(const FunctionDecl *FD);
...
 // element accessors (operator[], operator*, front, back, at), or propagating
 // operations (operator+, operator-, operator++, operator--).
+//
+// `ActualObjectType` is the object's actual type at the call site (e.g., the
+// type before any `DerivedToBase` cast). It is included in the `gsl::Owner`
+// check, so inherited methods on non-annotated bases are still tracked.
 bool shouldTrackImplicitObjectArg(const CXXMethodDecl *Callee,
-                                  bool RunningUnderLifetimeSafety);
+                                  bool RunningUnderLifetimeSafety,
+                                  QualType ActualObjectType = QualType());
```
Xazax-hun: Nit: this might not be the actual type. E.g.:
```
struct A { ... };
struct B: A { ... };
struct C: B { ... };

void f(B& b) {
  b.method();
}
```

In the code above, it is possible that `method` is declared in `A`, the static implicit object type is `B` and the dynamic type is `C`. 

I wonder if `ActualObjectType` makes people think this would be `C`.

Maybe we could call is `ImplicitObjectArgumentType`.

clang/lib/Analysis/LifetimeSafety/LifetimeAnnotations.cpp
```diff
@@ -122,15 +121,18 @@ static bool isReferenceOrPointerLikeType(QualType QT) {
 }
 
 bool shouldTrackImplicitObjectArg(const CXXMethodDecl *Callee,
-                                  bool RunningUnderLifetimeSafety) {
+                                  bool RunningUnderLifetimeSafety,
+                                  QualType ActualObjectType) {
   if (!Callee)
     return false;
+  const bool IsObjectGslOwner =
```
Xazax-hun:
```suggestion
  const bool IsGslOwnerImplicitObject =
```

Xazax-hun (commented):
The change looks good to me, some nits about naming inline. Could you run this over LLVM to see if there are any false positives just in case?

clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeAnnotations.h
usx95: Why do we need a default arg here ?
If this is related stmt-local `-Wdangling` analysis, we can use `RunningUnderLifetimeSafety` to restrict the scope of this PR.

aeft: Except for the stmt-local analysis, it also has a callsite in `shouldTrackPointerImplicitObjectArg`.

aeft:
Done. This PR introduces no new warnings.

clang/include/clang/Analysis/Analyses/LifetimeSafety/LifetimeAnnotations.h
usx95: Is doing the same for the that callsite problematic ?

clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp
```diff
@@ -882,15 +882,16 @@ void FactsGenerator::handleFunctionCall(const Expr *Call,
...
     const ParmVarDecl *PVD = nullptr;
     if (const auto *Method = dyn_cast<CXXMethodDecl>(FD);
         Method && Method->isInstance() && !isa<CXXConstructorDecl>(FD)) {
       if (I == 0)
         // For the 'this' argument, the attribute is on the method itself.
         return implicitObjectParamIsLifetimeBound(Method) ||
                shouldTrackImplicitObjectArg(
-                   Method, /*RunningUnderLifetimeSafety=*/true);
+                   Method, /*RunningUnderLifetimeSafety=*/true,
+                   Args[0]->IgnoreParenBaseCasts()->getType());
```
usx95: I think it might make sense to do this inside `shouldTrackImplicitObjectArg` and make the function accept only the `Args[0]` as it is. 

Also have you considered reusing `getBestDynamicClassType` ?

clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp
```diff
@@ -912,7 +912,8 @@ void FactsGenerator::handleFunctionCall(const Expr *Call,
     return I == 0 &&
            isGslPointerType(Method->getFunctionObjectParameterType()) &&
            shouldTrackImplicitObjectArg(Method,
-                                        /*RunningUnderLifetimeSafety=*/true);
+                                        /*RunningUnderLifetimeSafety=*/true,
+                                        /*ImplicitObjectArgument=*/nullptr);
```
aeft: Pass `nullptr` here to keep `gsl::Pointer` side behavior unchanged.

It seems like here we assume `gsl::Owner` and `gsl::Pointer` are mutually exclusive, and extending the check for `gsl::Owner` may break this assumption. E.g., it's possible that the implicit object only passes the `gsl::Pointer` check before this PR, but also passes the `gsl::Owner` check after this PR. Though I haven't observed any bad cases caused by this (no crashes, no test failures).

Maybe we should split `shouldTrackImplicitObjectArg` into two explicit handlers, one for `gsl::Pointer` and another for `gsl::Owner`?

@usx95 What do you think?

clang/lib/Sema/CheckExprLifetime.cpp
```diff
@@ -483,7 +483,7 @@ static void visitFunctionCallArguments(IndirectLocalPath &Path, Expr *Call,
     else if (EnableGSLAnalysis) {
       if (auto *CME = dyn_cast<CXXMethodDecl>(Callee);
           CME && lifetimes::shouldTrackImplicitObjectArg(
-                     CME, /*RunningUnderLifetimeSafety=*/false))
+                     CME, /*RunningUnderLifetimeSafety=*/false, ObjectArg))
```
aeft: `ObjectArg` is not used in `shouldTrackImplicitObjectArg` right now since `RunningUnderLifetimeSafety` is false.

usx95 (approved):
LGTM.

usx95:
Let's wait for @Xazax-hun 's approval.

Xazax-hun (approved):

PR #197867, opened 2026-05-15
[clang][NFC] Unify `MacroState` `isAmbiguous` and `getModuleInfo`

davidstone:
Every call to `MacroState::getModuleInfo`, and `MacroState::isAmbiguous` are paired in the same function. Rather than doing the same work twice, just unify them into a single function, `getModuleInfo`, that returns both pieces of information in a new type `ModuleMacroInfo`.

Unfortunately, `getModuleInfo` and`ModuleMacroInfo` already exist, so rename them to `getFullModuleInfo` and `FullModuleMacroInfo`, respectively, since the new type is a subset of the old type. The new type contains just the pieces consumers care about.

While we're there, use the range constructor of `llvm::DenseSet` instead of default constructing and calling `insert` in a loop.

### DIFF
--- a/clang/include/clang/Lex/MacroInfo.h
+++ b/clang/include/clang/Lex/MacroInfo.h
@@ -583,6 +583,11 @@ public:
   unsigned getNumOverridingMacros() const { return NumOverriddenBy; }
 };
 
+struct ModuleMacroInfo {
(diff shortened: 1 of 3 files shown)


cor3ntin (approved):
LGTM

PR #197868, opened 2026-05-15
[libc][math] Add gcc-7 to 9 compatibility for shared math.

lntue:
- Add more fine-grained constexpr annotations.
- Apply fine-grained constexpr annotations to `src/__support`, `src/__support/FPUtil` folder.
- Remove or update constexpr annotations for functions in `src/__support/math` folder.
- Fix or ignore old gcc's warnings.

Assisted-by: Gemini 3.1 Flash with step 3.


kaladron:
I'll do more thorough review when I'm fully awake, but I'm wondering if we should only have compilation warnings on gcc 15 and later and Clang 22 and later.  And just slowly raise those to only give warnings on recent compilers.  Then werror could still work.

lntue:
> I'll do more thorough review when I'm fully awake, but I'm wondering if we should only have compilation warnings on gcc 15 and later and Clang 22 and later. And just slowly raise those to only give warnings on recent compilers. Then werror could still work.

I would say whatever default gcc versions from the latest 2 versions (LTS?) of ubuntu/debian?

libc/src/__support/FPUtil/NormalFloat.h
```diff
@@ -95,7 +95,7 @@ template <typename T> struct NormalFloat {
     return result;
   }
 
-  LIBC_INLINE constexpr operator T() const {
+  LIBC_INLINE operator T() const {
```
bassiounix: what is wrong with marking this as `constexpr`?

libc/src/__support/FPUtil/NormalFloat.h
```diff
@@ -216,8 +216,7 @@ NormalFloat<long double>::init_from_bits(FPBits<long double> bits) {
   }
 }
 
-template <>
-LIBC_INLINE constexpr NormalFloat<long double>::operator long double() const {
+template <> LIBC_INLINE NormalFloat<long double>::operator long double() const {
```
bassiounix: same here as above

bassiounix (commented):
LGTM overall.
I'm still trying to wrap my head around the mass removal of `constexpr` qualifier in math.

libc/src/__support/FPUtil/NormalFloat.h
lntue: I think these operators can be marked with `LIBC_BIT_CAST_CONSTEXPR` instead, due to the use of `FPBits` floating point constructor and `FPBits::get_val()`.

libc/src/__support/FPUtil/NormalFloat.h
lntue: Done.

libc/src/__support/FPUtil/NormalFloat.h
lntue: Done.

kaladron:
> > I'll do more thorough review when I'm fully awake, but I'm wondering if we should only have compilation warnings on gcc 15 and later and Clang 22 and later. And just slowly raise those to only give warnings on recent compilers. Then werror could still work.
> 
> I would say whatever default gcc versions from the latest 2 versions (LTS?) of ubuntu/debian?

That's fine.  I'd just like us to not have to care about warnings in GCC 7 _or_ GCC 9 =)

kaladron (approved):

PR #197869, opened 2026-05-15
[CodeGenPrepare] Recursively sink expensive operands when converting select to branch

ayrai-gb:
## Motivation

When a select instruction has expensive operands like division or square root, speculatively executing both sides wastes cycles on the path not taken. Converting such selects to branches allows the expensive computation to be sunk into the taken block, avoiding unnecessary work on the untaken path.

**Why branches win for expensive operands:**
- A select (cmov) requires both true and false operands to be computed before the condition is resolved
- With a branch only the taken path executes and expensive divisions on the untaken side are skipped entirely
- Modern branch predictors handle predictable branches well making this profitable when operands are costly

**What counts as expensive (cost >= TCC_Expensive):**
- Integer division/remainder (`udiv`, `sdiv`, `urem`, `srem`)
- Floating-point division (`fdiv`)
- Floating-point square root (`fsqrt`) - especially f64 and vector variants

Integer div/rem by constant lowers to mul+shift (not `idiv`), so these are excluded from being treated as expensive.


llvm/lib/CodeGen/CodeGenPrepare.cpp
```diff
@@ -7582,24 +7582,70 @@ bool CodeGenPrepare::optimizeLoadExt(LoadInst *Load) {
...
+    assert(It->second > 0 && "this operand has atleast one use.");
+    --It->second;
+    if (It->second != 0)
+      return false;
+  }
+
+  if (Chain)
+    Chain->push_back(I);
+
+  bool HasExpensive = TTI->isExpensiveToSpeculativelyExecute(I);
```
RKSimon: This call is almost useless these days - TCC_Expensive (= 4) has lost all meaning since we've moved to costs based on hw measurements.

RKSimon (changes requested):
Practically all the x86 changes appear to be regressions - you're going to have to be a lot smarter about cost analysis.

llvm/lib/CodeGen/CodeGenPrepare.cpp
ayrai-gb: Sorry for the delay, I'm working on this, I'll push the changes in a week

ayrai-gb:
gentle ping @RKSimon @phoebewang whenever ur free

ayrai-gb:
gentle ping @RKSimon @phoebewang

ayrai-gb:
gentle ping @RKSimon @phoebewang

RKSimon (changes requested):
please fix the merge conflict

ayrai-gb:
> please fix the merge conflict

done

PR #162259, opened 2025-10-07
[GVN] Support rnflow pattern matching and transform

madhur13490:
This is an alternative patch to #144987, adding support in GVN.
In this patch, I added support in GVN to recognize PRE load and hoist it out of the loop. This patch leads to ~15% improvement for the rnflow benchmark in Polyhedron testsuite when tested on NVIDIA's Grace platform.

Testing:
1. Various internal workloads and benchmarks
2. LLVM Test-suite
3. LLVM regressions tests.

No crash, all tests pass.
In this revision, I am keeping the feature off-by-default and the following patch will just flip the switch.

Assisted by Cursor

PR #197870, opened 2026-05-15
[LICM][NFC] Factor canHoistLoad out of canSinkOrHoistInst into LoopUtils

madhur13490:
This patch moves the load-only legality logic from `canSinkOrHoistInst` into a new `canHoistLoad` helper, declared in `llvm/Transforms/Utils/LoopUtils.h` and defined in `LICM.cpp`. `canSinkOrHoistInst` delegates to it for `LoadInst`, so its behavior is unchanged.

This exposes the load-hoist check for reuse by other passes (e.g. the GVN min-finding select transform in PR #162259).

### DIFF
--- a/llvm/include/llvm/Transforms/Utils/LoopUtils.h
+++ b/llvm/include/llvm/Transforms/Utils/LoopUtils.h
@@ -458,11 +458,11 @@ LLVM_ABI bool canSinkOrHoistInst(Instruction &I, AAResults *AA,
 /// Returns true if it is legal to hoist \p LI out of \p CurLoop. This is the
 /// load-specific subset of \c canSinkOrHoistInst: it rejects volatile or
 /// ordered loads, allows constant-memory / invariant.load / invariant.start-
-/// dominated loads unconditionally, and otherwise queries MemorySSA for an
+/// dominated loads unconditionally, and otherwise queries \p MSSA for an
 /// in-loop clobber. \p TargetExecutesOncePerLoop has the same meaning as in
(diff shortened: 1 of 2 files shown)


llvm/include/llvm/Transforms/Utils/LoopUtils.h
```diff
@@ -455,6 +455,18 @@ LLVM_ABI bool canSinkOrHoistInst(Instruction &I, AAResults *AA,
...
                                  OptimizationRemarkEmitter *ORE = nullptr);
 
+/// Returns true if it is legal to hoist \p LI out of \p CurLoop. This is the
+/// load-specific subset of \c canSinkOrHoistInst: it rejects volatile or
+/// ordered loads, allows constant-memory / invariant.load / invariant.start-
+/// dominated loads unconditionally, and otherwise queries MemorySSA for an
+/// in-loop clobber. \p TargetExecutesOncePerLoop has the same meaning as in
+/// \c canSinkOrHoistInst (set to true when hoisting to the preheader).
+LLVM_ABI bool canHoistLoad(LoadInst &LI, AAResults *AA, DominatorTree *DT,
+                           Loop *CurLoop, MemorySSAUpdater &MSSAU,
```
antoniofrighetto: Maybe directly pass a reference to MSSA instead of retrieving it inside?

madhur13490: Sure, done.

alinas (approved):
LGTM.

PR #197871, opened 2026-05-15
[llvm-profgen] Speed up parsing of MMap events

Jinjie-Huang:
During the mmap extraction, `extractMMapEventForBinary()` is invoked line by line to parse the perf script(usually 100K to 1M+ lines). This patch tries to early quit when the current line does not contain the Binary name, and this will help to avoid subsequent parsing overhead(baseline behavior uses regex to extract the name, returning false if it doesn't match the current binary).

With this minor change, the execution time of extractMMapEventForBinary() dropped from 13,600ms to 32ms(-99%) in an internal workload test(827K lines).

### DIFF
--- a/llvm/tools/llvm-profgen/PerfReader.cpp
+++ b/llvm/tools/llvm-profgen/PerfReader.cpp
@@ -1115,10 +1115,10 @@ bool PerfScriptReader::extractMMapEventForBinary(ProfiledBinary *Binary,
   bool R = false;
   SmallVector<StringRef, 7> Fields;
   if (Line.contains("PERF_RECORD_MMAP2 ")) {
-    static Regex RegMmap2(MMap2Pattern);
+    Regex RegMmap2(MMap2Pattern);
     R = RegMmap2.match(Line, &Fields);
   } else if (Line.contains("PERF_RECORD_MMAP ")) {
-    static Regex RegMmap(MMapPattern);
+    Regex RegMmap(MMapPattern);
     R = RegMmap.match(Line, &Fields);
   } else
     llvm_unreachable("unexpected MMAP event entry");
--- a/llvm/tools/llvm-profgen/PerfReader.cpp
+++ b/llvm/tools/llvm-profgen/PerfReader.cpp
@@ -1079,6 +1079,9 @@ void PerfScriptReader::parseSample(TraceStream &TraceIt) {
 bool PerfScriptReader::extractMMapEventForBinary(ProfiledBinary *Binary,
                                                  StringRef Line,
                                                  MMapEvent &MMap) {
+  if (!Binary->isKernel() && !Line.contains(Binary->getName()) &&
+      !ShowMmapEvents)
+    return false;
   // Parse a MMap2 line like:
   //  PERF_RECORD_MMAP2 2113428/2113428: [0x7fd4efb57000(0x204000) @ 0
   //  08:04 19532229 3585508847]: r-xp /usr/lib64/libdl-2.17.so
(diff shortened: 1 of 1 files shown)


llvm/tools/llvm-profgen/PerfReader.cpp
```diff
@@ -1112,10 +1115,10 @@ bool PerfScriptReader::extractMMapEventForBinary(ProfiledBinary *Binary,
   bool R = false;
   SmallVector<StringRef, 7> Fields;
   if (Line.contains("PERF_RECORD_MMAP2 ")) {
-    Regex RegMmap2(MMap2Pattern);
+    static Regex RegMmap2(MMap2Pattern);
```
boomanaiden154: I think this still runs afoul of https://llvm.org/docs/CodingStandards.html#do-not-use-static-constructors.

Local static variables I believe still have a global destructor. It would be nice if we could figure out a better way to store these if their construction is really that expensive.

Jinjie-Huang: Thanks for pointing this out, I've removed the static change here. I guess a better way would be to put it inside the `PerfScriptReader` class. The overhead is actually negligible now since we have the early quit(only need to construct `Regex` very few times). The baseline incurred this overhead (about 3% of the total time) because it evaluated the regex on every single line.

llvm/tools/llvm-profgen/PerfReader.cpp
```diff
@@ -1079,6 +1079,9 @@ void PerfScriptReader::parseSample(TraceStream &TraceIt) {
 bool PerfScriptReader::extractMMapEventForBinary(ProfiledBinary *Binary,
                                                  StringRef Line,
                                                  MMapEvent &MMap) {
+  if (!Binary->isKernel() && !Line.contains(Binary->getName()) &&
+      !ShowMmapEvents)
+    return false;
   // Parse a MMap2 line like:
   //  PERF_RECORD_MMAP2 2113428/2113428: [0x7fd4efb57000(0x204000) @ 0
```
aaupov, line 1086: I'd compare the beginning of the line to PERF_RECORD_MMAP

Jinjie-Huang: Thanks for the review! The "PERF_RECORD_MMAP" prefix is already filtered out by `isMMapEvent(line)` externally. 
So this extra check here just filters mmap lines by the binary name to save us from the unnecessary regex overhead (~98% of the mmap lines in my case).

llvm/tools/llvm-profgen/PerfReader.cpp
```diff
@@ -1079,6 +1079,9 @@ void PerfScriptReader::parseSample(TraceStream &TraceIt) {
 bool PerfScriptReader::extractMMapEventForBinary(ProfiledBinary *Binary,
                                                  StringRef Line,
                                                  MMapEvent &MMap) {
+  if (!Binary->isKernel() && !Line.contains(Binary->getName()) &&
+      !ShowMmapEvents)
```
HighW4y2H3ll, line 1083: Thanks for the measurement! It's surprising to see this filter cutting off so much in the runtime. Maybe it would be also be benificial to pre-filter the binary name when running the perf cmd? but still it's a reasonable improvement. I have a few questions regarding the constraints here:
1) Why is this filter depending on `ShowMmapEvents`? This flag seems to be only intended for extra loggings..
2) Is this going to miss out MMAP events from shared libraries? and maybe anonymous pages?
3) Is this check `Line.contains(Binary->getName())` going to over-match if the binary name is too generic? Can we have a regex or a more specific filter for binary names?

boomanaiden154: `ShowMMapEvents` logs regardless of whether or not the binary name matches.

Users were already skipping mmap events from shared libraries/anonymous pages, so we aren't losing anything.

aaupov (approved):

llvm/tools/llvm-profgen/PerfReader.cpp
Jinjie-Huang: Thanks for the review!

1. Yes, as mentioned above, this is to preserve the exact behavior of ShowMmapEvents (which dumps all binary names from the perf data).
2. Also as mentioned above, this function is only intended to extract the mapping for the target binary itself. We don't care about other DSO information here.
3. This acts essentially like a Bloom filter. It serves purely as a fast pre-filter for acceleration; the exact and strict matching is still handled subsequently.

Jinjie-Huang:
Thanks all! Since there seem to be no further questions, I plan to leave this open for one more day before merging.

PR #197604, opened 2026-05-14
[LifetimeSafety] Recognize declarations nested under std namespace

zeyi2:
Previously `isInStlNamespace()` only checked the immediate declaration context. This missed declarations nested below `std` through records or intermediate namespaces, such as `std::basic_string_view` constructors.

This commit fixes the problem by walking the `DeclContext` chain in `isInStlNamespace`.

Closes #197454

PR #197220, opened 2026-05-12
[LifetimeSafety] Diagnose invalidated-global

zeyi2:
Teach lifetime safety invalidation diagnostics to handle origins that escape through global or static storage before the referenced object is invalidated. Previously they were skipped.

Follow up of #196680
Closes https://github.com/llvm/llvm-project/issues/195706

PR #196680, opened 2026-05-09
[LifetimeSafety] Diagnose invalidated-field

zeyi2:
Teach lifetime safety invalidation diagnostics to handle origins that escape through fields before the referenced object is invalidated. Previously they were skipped.

Partially addresses https://github.com/llvm/llvm-project/issues/195706

PR #195231, opened 2026-05-01
[LifetimeSafety] Detect iterator invalidation through container aliases

zeyi2:
The previous heuristic in `handleInvalidatingCall` is too conservative. The ideal way would be completely removing this, but it would introduce ~10 regressions in the existing testcases.

So instead I replace the filter with a narrower guard that only skips direct field accesses (AccessPath currently lacks field granularity and cannot distinguish `s.v1` from `s.v2`).

Closes https://github.com/llvm/llvm-project/issues/193044

PR #197873, opened 2026-05-15
Reland "[LifetimeSafety] Detect iterator invalidation through container aliases"

### LINKED ISSUE #193044: [LifetimeSafety] Detect invalidation of pointer parameters
```cpp
#include <iostream>
#include <string>
#include <string_view>
#include <vector>

void fine() {
    std::vector<int> vv;
    auto it = vv.begin();
    vv.push_back(42); // detected. Ok.
    (void)it;
}

// Nothing detected below.
void foo() {
    std::vector<int> vv;
    std::vector<int>& v = vv;
    auto it = vv.begin();
    v.push_back(42);
    (void)it;
}

void bar() {
    std::vector<int> vv;
    std::vector<int>* v = &vv;
    auto it = vv.begin();
    v->push_back(42);
    (void)it;
}

void baz(std::vector<int>* v) {
    auto it = v->begin();
    v->push_back(42);
    (void)it;
}

```
https://godbolt.org/z/1j951a88E

zeyi2:
This relands #195231, which was reverted in commit 7c9717848851f3a71908becab4312ddc2d8482b8.

The original crash from the reproducer no longer reproduces after #196680, #197220, and #197604. I verified the original `repro.cpp`: it no longer hits the lifetime-safety assertion now.

Also added regression tests for the crash:

```cpp
struct SinkInteriorBorrow {
  const char *dest_; // expected-note {{this field dangles}}

  SinkInteriorBorrow(std::string *dest, int n) : dest_(dest->data()) { // expected-warning {{parameter which escapes to a field is later invalidated}}
    if (n > 0)
      dest->clear(); // expected-note {{invalidated here}}
  }
};
```

Closes https://github.com/llvm/llvm-project/issues/193044


clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp
```diff
@@ -810,9 +810,11 @@ void FactsGenerator::handleInvalidatingCall(const Expr *Call,
...
   if (!isInvalidationMethod(*MD))
     return;
-  // Heuristics to turn-down false positives.
-  auto *DRE = dyn_cast<DeclRefExpr>(Args[0]);
-  if (!DRE || DRE->getDecl()->getType()->isReferenceType())
+
+  // Heuristics to turn-down false positives. Skip member field expressions for
+  // now. This is not a perfect filter and will still surface some false
+  // positives (e.g. `auto& r = s.v`).
+  if (!isa<DeclRefExpr>(Args[0]->IgnoreParenImpCasts()))
```
zeyi2: I use `IgnoreParenImpCasts` here to handle: `(v).push_back(x)`, 

-----

v.push_back(x):

```
CXXMemberCallExpr
|-MemberExpr .push_back
| `-DeclRefExpr 'v'
`-ImplicitCastExpr
  `-DeclRefExpr 'x'
```

-----

(v).push_back(x):

```
CXXMemberCallExpr
|-MemberExpr .push_back
| `-ParenExpr
|   `-DeclRefExpr 'v'
`-ImplicitCastExpr
  `-DeclRefExpr 'x'
```

usx95:
Thanks. Can you please verify that the original report and the reproducer in https://github.com/llvm/llvm-project/pull/195231 do not crash anymore. Add that as a test and mention this in the description.

zeyi2:
The original repro.cpp:

Ran with:

```
./build/bin/clang++ repro.cpp \
  -Wlifetime-safety \
  -std=c++23 \
  -target x86_64-apple-macosx \
  -isysroot "$(xcrun --show-sdk-path)" \
  -stdlib=libc++ \
  -fsyntax-only
```

```
repro.cpp:364902:18: note: in instantiation of function template specialization
      'icu_78::UnicodeString::toUTF8String<std::string>' requested here
 364902 |   str = skeleton.toUTF8String<std::string>(str);
        |                  ^
repro.cpp:275371:13: note: invalidated here
 275371 |       dest->reserve(dest->length() + initialAppendCapacity);
        |       ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
repro.cpp:275388:16: note: this field dangles
 275388 |   StringClass* dest_;
        |                ^~~~~
383 warnings generated.
```

No crash, working as expected :)

Log: [result.log](https://github.com/user-attachments/files/27807364/result.log)

zeyi2:
Friendly ping :)

clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp
usx95: I see. I would avoid special casing this. We need a more general solution for supporting lvalue expressions which are not `DeclRefExpr`. E.g. `(flag ? v1: v2).push_back(x)`.

usx95 (approved):
LGTM. One nit inline.

clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp
zeyi2: Thanks for the feedback! I'm thinking of changing this into a negative filter: keep the existing analysis path and only filter out unmodeled subobjects for now, such as fields and array subscripts.

For conditionals, the same filter would apply to each branch, so `(flag ? v1 : v2).push_back(x)` would be tracked, while `(flag ? s.v1 : s.v2).push_back(x)` would still be filtered until access paths can model subobjects.

Does that direction sound reasonable to you?

usx95: Special casing "unmodeled" subobjects would introduce yet another heuristic. I would keep the heuristic simple for now and just target `DeclRefExpr`. Special casing conditionals was not my intention, it is merely an example of lvalue-expressions. These will fallout naturally once we have reduced the false-positives and removed this check to only target DRE.
What about keeping this as 
```cpp
if (!isa<DeclRefExpr>(Args[0]))
    return;
```

clang/test/Sema/warn-lifetime-safety-invalidations.cpp
```diff
@@ -407,49 +460,69 @@ void Invalidate1Use1IsInvalid() {
...
 
+struct Sink {
+  std::string *dest_; // expected-note {{this field dangles}}
+
+  Sink(std::string *dest, int n) : dest_(dest) { // expected-warning {{parameter which escapes to a field is later invalidated}}
+    if (n > 0)
+      dest->clear(); // expected-note {{invalidated here}}
+  }
+};
+
```
usx95: 1. This is a case which is false positive due to owner-borrow and interior-borrow. Add a FIXME here.
2. True positive should use a interior-borrow
```cpp
struct Sink {

  const char *dest_; // expected-note {{this field dangles}}
​
  Sink(std::string *dest, int n) : dest_(dest->data()) { // expected-warning {{parameter which escapes to a field is later invalidated}}
    if (n > 0)
      dest->clear(); // expected-note {{invalidated here}}
  }
};
``

**Review by usx95 — COMMENTED** [review]
One nit.

File: clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp (hunk 1 above)
Comment (zeyi2) [reply clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp]: > What about keeping this as
>```
> if (!isa<DeclRefExpr>(Args[0]))
>    return;
> ```

I tried it, it would cause 6 regressions. Most are trivial/common cases like:

```
void IteratorInvalidatedThroughPointerParameter(std::vector<int> *v) {
  auto it = v->begin();
  v->push_back(42);
  (void)it;
}
```

IMO this would make the analysis less useful...

I think the main reason is that in `p->push_back(...)`, the receiver is represented as `ImplicitCastExpr -> DeclRefExpr` (The AST snippets in previous comments also showcase this). So `isa<DeclRefExpr>(Args[0])` would filter these cases out.

I'd like to avoid losing these existing diagnostics if possible. What do you think of using `Args[0]->IgnoreImpCasts()` as a workaround? It still targets DRE, but ignores some necessary lvalue-to-rvalue ccast needed for ordinary pointer receivers.

File: clang/test/Sema/warn-lifetime-safety-invalidations.cpp (hunk 2 above)
Comment (zeyi2) [reply clang/test/Sema/warn-lifetime-safety-invalidations.cpp]: Sorry that I missed this, fixed now.

File: clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp (hunk 1 above)
Comment (usx95) [reply clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp]: > What do you think of using Args[0]->IgnoreImpCasts() as a workaround? 

LGTM. Thanks.

File: clang/test/Sema/warn-lifetime-safety-invalidations.cpp [hunk 3]
```diff
@@ -407,49 +455,72 @@ void Invalidate1Use1IsInvalid() {
...
 }
-void Invalidate1Use2IsOk() {
+void Invalidate2Use1IsOk() {
     S s;
     auto it = s.strings1.begin();
     s.strings2.push_back("1");
     *it;
-}void Invalidate1Use2ViaRefIsOk() {
+}
+// FIXME: Requires field-sensitive AccessPaths to fix.
```
usx95, line 478: Could you please bring back `ConditionalContainerInvalidatesIterator` and `ConditionalFieldInvalidationIsOk` as a FIXME tests.

> ConditionalFieldInvalidationIsOk

```cpp
void ConditionalFieldInvalidationIsOk(bool flag) {
    S s;
    auto it = s.strings1.begin();
    (flag ? s.strings1 : s.strings2).push_back("1");
    *it;
}
​```

Btw, this is should be an error because `s.strings1` can end invalidating `it`.

usx95 (approved):
LGTM.

clang/test/Sema/warn-lifetime-safety-invalidations.cpp
zeyi2: Done, thanks :)

zeyi2:
I plan to merge this later, since I believe I've addressed all the review feedback. :)

PR #197874, opened 2026-05-15
[Clang] Handle deduced auto types within AtomicType

mkovacevic99:
Fixes github issue: https://github.com/llvm/llvm-project/issues/118058

This patch fixes handling of deduced types nested inside AtomicType.

Previously, GetContainedDeducedTypeVisitor did not recurse through AtomicType nodes, causing getContainedDeducedType() to incorrectly return null for types such as:

**_Atomic(__auto_type)**

As a result, undeduced AutoType instances could bypass semantic checks and later reach codegen/type layout paths, eventually triggering assertions such as:

_!A->getDeducedType().isNull() && "cannot request the size of an undeduced or dependent auto type_

This patch adds AtomicType traversal support to GetContainedDeducedTypeVisitor so contained deduced types are properly discovered.

Additionally, template deduction logic for AtomicType is extended to allow deduction from the atomic value type when the pattern contains an auto type.

CheckOriginalCallArgDeduction is also adjusted to handle atomic wrappers around deduced types correctly.

This brings Clang behavior closer to GCC for combinations involving _Atomic and __auto_type and prevents undeduced auto types from escaping into later compilation stages.

### DIFF
--- a/clang/docs/ReleaseNotes.rst
+++ b/clang/docs/ReleaseNotes.rst
@@ -1041,6 +1041,9 @@ Crash and bug fixes
 
 - Fixed ``security.VAList`` checker producing false positives when analyzing
   C23 code where ``va_start`` expands to ``__builtin_c23_va_start``.
+  
+- Fixed a compiler crash when combining ``_Atomic`` and ``__auto_type``
+  in C, for example ``_Atomic __auto_type x = expr``. Fixes #118058.
 
 Improvements
 ^^^^^^^^^^^^
--- a/clang/docs/ReleaseNotes.rst
+++ b/clang/docs/ReleaseNotes.rst
@@ -1043,7 +1043,7 @@ Crash and bug fixes
   C23 code where ``va_start`` expands to ``__builtin_c23_va_start``.
   
 - Fixed a compiler crash when combining ``_Atomic`` and ``__auto_type``
-  in C, for example ``_Atomic __auto_type x = expr``. Fixes #118058.
+  in C, for example ``_Atomic __auto_type x = expr``. (#GH118058)
 
 Improvements
 ^^^^^^^^^^^^
--- a/clang/lib/Sema/SemaTemplateDeduction.cpp
+++ b/clang/lib/Sema/SemaTemplateDeduction.cpp
@@ -3741,8 +3741,10 @@ CheckOriginalCallArgDeduction(Sema &S, TemplateDeductionInfo &Info,
   QualType A = OriginalArg.OriginalArgType;
   QualType OriginalParamType = OriginalArg.OriginalParamType;
 
-  // Check for type equality (top-level cv-qualifiers are ignored).
-  if (Context.hasSameUnqualifiedType(A, DeducedA))
+  // Check for type equality (top-level cv-qualifiers and _Atomic are ignored,
+  // since _Atomic is treated as a qualifier).
+  if (Context.hasSameType(A.getAtomicUnqualifiedType(),
+                          DeducedA.getAtomicUnqualifiedType()))
     return TemplateDeductionResult::Success;
 
   // Strip off references on the argument types; they aren't needed for
@@ -5110,6 +5112,25 @@ namespace {
       NewTL.setNameLoc(TL.getNameLoc());
       NewTL.setQualifierLoc(TL.getQualifierLoc());
       return Result;
+    }
+
+    QualType TransformAtomicType(TypeLocBuilder &TLB, AtomicTypeLoc TL) {
+      // When building the function parameter for placeholder type deduction
+      // (Replacement is the invented template parameter), dig through _Atomic
+      // around an auto placeholder so deduction matches the non-atomic
+      // argument. The _Atomic wrapper is re-applied by the final substitution
+      // pass, which uses a concrete Replacement and falls through to the
+      // default transform.
+      //
+      // This handles only the simple case where _Atomic wraps auto directly
+      // (e.g. _Atomic(auto)), which is what the C standard currently permits.
+      // If more complex forms such as _Atomic(auto*) are ever allowed, the
+      // correct fix would be to treat _Atomic as a qualifier inside
+      // DeduceTemplateArgumentsByTypeMatch instead.
+      if (isa_and_nonnull<TemplateTypeParmType>(Replacement) &&
+          TL.getValueLoc().getType()->getContainedAutoType())
+        return getDerived().TransformType(TLB, TL.getValueLoc());
+      return inherited::TransformAtomicType(TLB, TL);
     }
 
     ExprResult TransformLambdaExpr(LambdaExpr *E) {
--- a/clang/test/CodeGen/atomic-auto-type.c
+++ b/clang/test/CodeGen/atomic-auto-type.c
@@ -1,0 +1,10 @@
+// RUN: %clang_cc1 -triple x86_64-pc-linux-gnu -emit-llvm -o - %s > /dev/null
+
+// This is a regression test for handling of __auto_type inside _Atomic.
+// Previously this could lead to an undeduced AutoType escaping into
+// ASTContext::getTypeInfoImpl and causing an assertion failure.
+
+void f(double x) {
+  __auto_type _Atomic xa = x;
+  _Atomic __auto_type ax = x;
+}
--- a/clang/test/Sema/atomic-auto-type.c
+++ b/clang/test/Sema/atomic-auto-type.c
@@ -1,0 +1,26 @@
+// RUN: %clang_cc1 -triple x86_64-pc-linux-gnu -fsyntax-only %s
+// expected-no-diagnostics
+
+int main() {
+  double x = 37;
+
+  __auto_type _Atomic xa = x;
+  _Atomic __auto_type ax = x;
+  
+  _Static_assert(
+      __builtin_types_compatible_p(__typeof(xa), _Atomic double),
+      "incorrect xa type");
+
+  _Static_assert(
+      __builtin_types_compatible_p(__typeof(ax), _Atomic double),
+      "incorrect ax type");
+
+  _Static_assert(
+      __builtin_types_compatible_p(_Atomic double, __typeof(xa)),
+      "incorrect");
+
+  _Static_assert(
+      __builtin_types_compatible_p(_Atomic double, __typeof(ax)),
+      "incorrect");
+  return 0;
+}
--- a/clang/lib/AST/Type.cpp
+++ b/clang/lib/AST/Type.cpp
@@ -2102,6 +2102,10 @@ public:
   Type *VisitPackExpansionType(const PackExpansionType *T) {
     return Visit(T->getPattern());
   }
+
+  Type *VisitAtomicType(const AtomicType *T) {
+    return Visit(T->getValueType());
+  }
 };
 
 } // namespace


mkovacevic99:
@iamanonymouscs @EugeneZelenko can I please ask for your review or if you can tag those who can review this PR. Thanks

cor3ntin (commented):
This change needs a release note.
Please add an entry to `clang/docs/ReleaseNotes.rst` in the section the most adapted to the change, and referencing any Github issue this change fixes. Thanks!

mkovacevic99:
@cor3ntin @AaronBallman can you please review this PR when you have time, I would really appreciate it. Thanks

clang/test/Sema/atomic-auto-type.c
```diff
@@ -0,0 +1,29 @@
+// RUN: %clang_cc1 -triple x86_64-pc-linux-gnu -emit-llvm -o - %s
```
AaronBallman: This test should live in `clang/test/CodeGen` because it's testing that we're not crashing at codegen time. But we should also have test coverage for the qualifier vs specifier differences, and `auto` vs `__auto_type`.

clang/test/Sema/atomic-auto-type.c
```diff
@@ -0,0 +1,29 @@
...
+// Previously this could lead to an undeduced AutoType escaping into
+// ASTContext::getTypeInfoImpl and causing an assertion failure.
+
+int main() {
+  double x = 37;
+
+  __auto_type _Atomic xa = x;
+  _Atomic __auto_type ax = x;
+  
+  _Static_assert(
```
AaronBallman, line 10: The static assertions should stay in a sema test, it's just the runtime behavior that needs to be checked in a codegen test, and that can be tested more simply:
```
void f() {
  __auto_type _Atomic xa;
}
```
but it seems there are other test cases that fail differently:
```
struct S {
  __auto_type _Atomic xa;
};
```
https://godbolt.org/z/d4TshMrrf so I think we'll need more exhaustive test coverage for declaration other scenarios.

AaronBallman (commented):
Background: `__auto_type` is a GCC extension, but it's closely related to `auto` in C23. In C23, `auto` is not a type specifier, it's the use as a storage class specifier along with the *lack* of a type specifier that triggers the type inference in C23. So that means:
```
 // Okay, `_Atomic` is a *qualifier* so this is a qualifier and a storage class specifier with no type specifier
_Atomic auto i = 12;
// Invalid, `_Atomic` is a *type specifier* this way.
_Atomic(auto) j = 12;
```
and GCC is faithfully implementing this: https://godbolt.org/z/ef81jzro1 and their implementation of `__auto_type` does the same: https://godbolt.org/z/ajsPhdWz4

However, it was known at the time `auto` was adopted into C23 that WG14 wanted to make it an actual type specifier, it wasn't done initially due to timing of the standard release. C2y has been making progress towards that change: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3579.htm but Clang's implementation predates any WG14 debates on what it means for `auto` as a type specifier. So Clang's implementation is not the same as GCC's, we do not distinguish between `_Atomic` as a qualifier and `_Atomic` as a specifier; both form an `AtomicType`, so Clang treats `_Atomic auto` and `_Atomic(auto)` the same: https://godbolt.org/z/9en1Ee8TE because we were thinking there's no good reason to want them to behave so differently. However, the paper explicitly calls this difference out and does not attempt to change it so perhaps we should follow GCC now.

So I suspect there's two issues here, one is the crash being fixed but the other is that we don't handle the qualifier version correctly (that doesn't need to be addressed in this PR but we should make sure we've got an issue filed for it).

mkovacevic99:
> Background: `__auto_type` is a GCC extension, but it's closely related to `auto` in C23. In C23, `auto` is not a type specifier, it's the use as a storage class specifier along with the _lack_ of a type specifier that triggers the type inference in C23. So that means:
> 
> ```
>  // Okay, `_Atomic` is a *qualifier* so this is a qualifier and a storage class specifier with no type specifier
> _Atomic auto i = 12;
> // Invalid, `_Atomic` is a *type specifier* this way.
> _Atomic(auto) j = 12;
> ```
> 
> and GCC is faithfully implementing this: https://godbolt.org/z/ef81jzro1 and their implementation of `__auto_type` does the same: https://godbolt.org/z/ajsPhdWz4
> 
> However, it was known at the time `auto` was adopted into C23 that WG14 wanted to make it an actual type specifier, it wasn't done initially due to timing of the standard release. C2y has been making progress towards that change: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3579.htm but Clang's implementation predates any WG14 debates on what it means for `auto` as a type specifier. So Clang's implementation is not the same as GCC's, we do not distinguish between `_Atomic` as a qualifier and `_Atomic` as a specifier; both form an `AtomicType`, so Clang treats `_Atomic auto` and `_Atomic(auto)` the same: https://godbolt.org/z/9en1Ee8TE because we were thinking there's no good reason to want them to behave so differently. However, the paper explicitly calls this difference out and does not attempt to change it so perhaps we should follow GCC now.
> 
> So I suspect there's two issues here, one is the crash being fixed but the other is that we don't handle the qualifier version correctly (that doesn't need to be addressed in this PR but we should make sure we've got an issue filed for it).

can you maybe open a separate issue for the second part? I think it's better to address it as a separate PR

clang/test/Sema/atomic-auto-type.c
mkovacevic99: I split the coverage so the semantic/type-deduction checks remain in a Sema test and added a separate CodeGen regression test for the original crash during LLVM emission.

I also checked additional declaration scenarios such as:

```c
struct S {
  __auto_type _Atomic xa;
};
```

Before this change, that also hit the same assertion (`cannot request the size of an undeduced or dependent auto type`). After the fix it now correctly diagnoses:

`'__auto_type' not allowed in struct member`

so the change appears to prevent undeduced `__auto_type` wrapped in `_Atomic` from escaping into later stages there as well.

I agree the qualifier vs specifier behavior and `auto` vs `__auto_type` semantics are separate concerns. I’d prefer to keep this change focused on fixing the crash and that we create a follow up issue for the behavioral differences and more exhaustive coverage.

mkovacevic99: @AaronBallman can you check if this PR is now okay?

AaronBallman:
> can you maybe open a separate issue for the second part? I think it's better to address it as a separate PR

I filed https://github.com/llvm/llvm-project/issues/202369 to track this.

clang/test/Sema/atomic-auto-type.c
AaronBallman: I've added @erichkeane and @mizvekov as reviewers for the changes to the template engine because I'm not 100% confident, but I think the changes look reasonable.

mkovacevic99: @erichkeane @mizvekov can I please ask for your review, thanks

mkovacevic99:
@erichkeane @mizvekov can you please review this PR, thanks

erichkeane (approved):
Template size seems reasonable.

clang/lib/Sema/SemaTemplateDeduction.cpp
```diff
@@ -1902,8 +1902,16 @@ static TemplateDeductionResult DeduceTemplateArgumentsByTypeMatch(
...
+      if (!AA) {
+        if (!P->getContainedAutoType())
+          return TemplateDeductionResult::NonDeducedMismatch;
+        // If it contains an auto type, we can try to deduce from the value type
+        // of the atomic type.
+        return DeduceTemplateArgumentsByTypeMatch(
+            S, TemplateParams, PA->getValueType(), A, Info, Deduced, TDF,
+            degradeCallPartialOrderingKind(POK),
+            /*DeducedFromArrayBound=*/false, HasDeducedAnyParam);
+      }
```
mizvekov: This isn't right, an unreduced AutoType shouldn't appear within template argument deduction.

That's because before performing template argument deduction for a placeholder type, we substitute any placeholder types with an invented template parameter and then perform deduction on that.

This is both how C++ standard specifies it, and also the implementation strategy for clang.

It seems you didn't update this substitution logic to dig through AtomicTypes as well.

mizvekov (changes requested):

clang/lib/Sema/SemaTemplateDeduction.cpp
mkovacevic99: do you think these updated changes cover now everything?

clang/lib/Sema/SemaTemplateDeduction.cpp
```diff
@@ -3742,7 +3742,14 @@ CheckOriginalCallArgDeduction(Sema &S, TemplateDeductionInfo &Info,
...
   // Check for type equality (top-level cv-qualifiers are ignored).
-  if (Context.hasSameUnqualifiedType(A, DeducedA))
+  // Also handle _Atomic(auto) deduction: SubstituteDeducedTypeTransform strips
+  // _Atomic from the function parameter so deduction succeeds against a
+  // non-atomic argument, then the final substitution re-wraps the result in
+  // _Atomic. DeducedA will be _Atomic(T) while A is T in that case.
+  if (Context.hasSameUnqualifiedType(A, DeducedA) ||
+      (DeducedA->isAtomicType() &&
+       Context.hasSameUnqualifiedType(
+           A, cast<AtomicType>(DeducedA)->getValueType())))
```
mizvekov: FWIW, it would be simpler to state that Atomic is treated as a qualifier, and we want to compare if the types are the same ignoring any qualifiers.

clang/lib/Sema/SemaTemplateDeduction.cpp
```diff
@@ -5047,21 +5054,24 @@ namespace {
     DeducedKind DK;
     QualType Replacement;
     bool UseTypeSugar;
+    bool StripAtomicAuto;
```
mizvekov: Thanks, this is better, though this flag did really complicate matters, why do we need it?

mkovacevic99: No, you are right, I found a way with Replacement parameter to verify which SubstituteDeducedTypeTransform(*this, TemplArg).Apply(Type); call it is.

clang/lib/Sema/SemaTemplateDeduction.cpp
```diff
@@ -3742,7 +3742,13 @@ CheckOriginalCallArgDeduction(Sema &S, TemplateDeductionInfo &Info,
...
 
   // Check for type equality (top-level cv-qualifiers are ignored).
-  if (Context.hasSameUnqualifiedType(A, DeducedA))
+  // _Atomic is treated as a qualifier, so strip it from both sides.
+  auto StripAtomic = [](QualType T) -> QualType {
+    if (const auto *AT = T->getAs<AtomicType>())
+      return AT->getValueType();
+    return T;
+  };
+  if (Context.hasSameUnqualifiedType(StripAtomic(A), StripAtomic(DeducedA)))
```
mizvekov: I would be surprised if we are not doing something like this elsewhere, and could use a helper here. Did you double check this?

clang/lib/Sema/SemaTemplateDeduction.cpp
```diff
@@ -5112,6 +5118,19 @@ namespace {
...
     }
 
+    QualType TransformAtomicType(TypeLocBuilder &TLB, AtomicTypeLoc TL) {
+      // When building the function parameter for placeholder type deduction
+      // (Replacement is the invented template parameter), dig through _Atomic
+      // around an auto placeholder so deduction matches the non-atomic
+      // argument. The _Atomic wrapper is re-applied by the final substitution
+      // pass, which uses a concrete Replacement and falls through to the
+      // default transform.
+      if (!Replacement.isNull() && Replacement->isTemplateTypeParmType() &&
```
mizvekov:
```suggestion
      if (isa_and_nonnull<TemplateTypeParmType>(Replacement) &&
```

clang/lib/Sema/SemaTemplateDeduction.cpp
```diff
@@ -5112,6 +5118,19 @@ namespace {
...
+          TL.getValueLoc().getType()->getContainedAutoType())
+        return getDerived().TransformType(TLB, TL.getValueLoc());
+      return inherited::TransformAtomicType(TLB, TL);
```
mizvekov, line 5133: FWIW, this is kind of a simple trick to handle the simplest cases of deduction that are currently allowed by the C standard, which only accepts plain `auto`, but not something like `auto*`.

I'd suspect the C standard would allow those in the future, and in that case what we should be doing instead is treating _Atomic as a qualifier through `DeduceTemplateArgumentsByTypeMatch` here: https://github.com/llvm/llvm-project/blob/dbd4528e8d0f6c90d8896ae48e1f0505e83a0fc6/clang/lib/Sema/SemaTemplateDeduction.cpp#L1595

I am not asking to make this change now, as I doubt it would be as simple as this patch, but just something to keep in mind and maybe add as a comment somewhere.

mizvekov (approved):
Minor nits, else LGTM, thanks!

clang/lib/Sema/SemaTemplateDeduction.cpp
mkovacevic99: good catch, I updated it

clang/lib/Sema/SemaTemplateDeduction.cpp
mkovacevic99: thanks for this, I added a comment for future reference

clang/lib/Sema/SemaTemplateDeduction.cpp
mkovacevic99: If everything is okay, please merge this PR. Let me know if there is anything else I need to do

mkovacevic99:
can someone please merge this PR since it's approved and all resolved @AaronBallman @mizvekov @cor3ntin @erichkeane  thanks

clang/docs/ReleaseNotes.rst
```diff
@@ -1069,6 +1041,9 @@ Crash and bug fixes
 
 - Fixed ``security.VAList`` checker producing false positives when analyzing
   C23 code where ``va_start`` expands to ``__builtin_c23_va_start``.
+  
+- Fixed a compiler crash when combining ``_Atomic`` and ``__auto_type``
+  in C, for example ``_Atomic __auto_type x = expr``. Fixes #118058.
```
cor3ntin:
```suggestion
  in C, for example ``_Atomic __auto_type x = expr``. (#GH118058)
```

cor3ntin (approved):

clang/test/CodeGen/atomic-auto-type.c
```diff
@@ -0,0 +1,10 @@
+// RUN: %clang_cc1 -triple x86_64-pc-linux-gnu -emit-llvm -o - %s > /dev/null
```
shafik, line 1: This is a codegen test but I don't see it actually testing any codegen

CC @erichkeane

erichkeane: Good catch shafik.  @mkovacevic99 : Please make sure you add check lines to this, and have it go through FileCheck like other tests!  Else we'll have to revert this patch.

mkovacevic99: created a follow up PR https://github.com/llvm/llvm-project/pull/206749

PR #197875, opened 2026-05-15
[SLP] Inefficient cost-modelling and codegen for reductions with slp-…

sushgokh:
…revec

When revectorizing, starting with reduction, SLP generates slightly inefficient code for reduction. e.g. In the godbolt link [here](https://godbolt.org/z/ez7KPnxM5), 
`hor_reduction --> original code`
`hor_reduction_revec_as_imagined_in_SLP  --> revectorized code would look like`

Rather than extracting per lane, we can extract original leaf nodes of the reduction, which are sub-vectors, and then perform usual reduction as in non-revectorized code. In the above link, 
`hor_reduction_ideal_revec --> how the revec code should look like`

Extracting subvectors and achieving the reduction result would be better than extracting per lane and achieving the same result.

Change in `Transforms/SLPVectorizer/RISCV/complex-loads.ll` is side effect of this change. The change with the file seems to better. IR post-SLP before/after patch and llvm-mca analysis of the corresponding codegen can be found below: 

1. IR post-SLP before patch: [here](https://godbolt.org/z/q1b7nW416 )
2. IR post-SLP after patch: [here](https://godbolt.org/z/W6vsP4zox )
3. codegen comparison: [here](https://godbolt.org/z/TGzjn3K8T)


llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -29189,34 +29190,48 @@ class HorizontalReduction {
...
+        %B = <e, f, g, h>
+        %add1 = A + B = <a + e, b + f, c + g, d + h>
+
+        After revectorization with VF=2 and Vec = <a, b, c, d, e, f, g, h>,
+        the reduction can be expressed as:
+        %A = shufflevector <8 x i32> %Vec, <8 x i32> poison, <4 x i32> <i32 0, i32 1, i32 2, i32 3>
+        %B = shufflevector <8 x i32> %Vec, <8 x i32> poison, <4 x i32> <i32 4, i32 5, i32 6, i32 7>
+        %add0 = add <4 x i32> zeroinitializer, %A
+        %add1 = add <4 x i32> %add0, %B
+      */
```
alexey-bataev: Use C++ style for comments

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -29189,34 +29190,48 @@ class HorizontalReduction {
...
+      for (auto I : seq<unsigned>(VF)) {
+        auto Position = I * DestTyNumElements;
+        Value *SubVec =
+            createExtractVector(Builder, Vec, DestTyNumElements, Position);
+        if (!Rdx) {
+          Rdx = SubVec;
+        } else {
+          Rdx = createOp(Builder, RdxKind, Rdx, SubVec, "rdx.op", ReductionOps);
+          if (isa<Instruction>(Rdx))
+            V.analyzedReductionRoot(cast<Instruction>(Rdx));
```
alexey-bataev:
```
          if (auto *I = dyn_cast<Instruction>(Rdx))
            V.analyzedReductionRoot(I);
```

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -29316,19 +29331,15 @@ class HorizontalReduction {
...
-                  TTI::SK_PermuteSingleSrc,
-                  FixedVectorType::get(VecTy->getScalarType(),
-                                       ReducedVals.size()),
-                  VectorTy,
-                  createStrideMask(I, ScalarTyNumElements, ReducedVals.size()));
-              VectorCost += TTI->getArithmeticReductionCost(RdxOpcode, VecTy,
-                                                            FMF, CostKind);
+              VectorCost += TTI->getShuffleCost(TTI::SK_ExtractSubvector, VecTy,
+                                                DstTy, {}, CostKind,
+                                                I * ScalarTyNumElements, VecTy);
```
alexey-bataev: Use `::getShuffleCost`

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -29189,34 +29190,48 @@ class HorizontalReduction {
...
-        // Lane[3] = <2 x Ty> <d, h>
-        // result[0] = reduce add Lane[0]
-        // result[1] = reduce add Lane[1]
-        // result[2] = reduce add Lane[2]
-        // result[3] = reduce add Lane[3]
-        SmallVector<int, 16> Mask = createStrideMask(I, DestTyNumElements, VF);
-        Value *Lane = Builder.CreateShuffleVector(Vec, Mask);
-        Rdx = Builder.CreateInsertElement(
-            Rdx, emitReduction(Lane, Builder, &TTI, DestTy), I);
+      unsigned RdxOpcode = RecurrenceDescriptor::getOpcode(RdxKind);
```
alexey-bataev: Unused variable

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -29316,19 +29331,15 @@ class HorizontalReduction {
...
             for (unsigned I : seq<unsigned>(ReducedVals.size())) {
-              VectorCost += TTI->getShuffleCost(
-                  TTI::SK_PermuteSingleSrc,
-                  FixedVectorType::get(VecTy->getScalarType(),
-                                       ReducedVals.size()),
-                  VectorTy,
-                  createStrideMask(I, ScalarTyNumElements, ReducedVals.size()));
-              VectorCost += TTI->getArithmeticReductionCost(RdxOpcode, VecTy,
-                                                            FMF, CostKind);
+              VectorCost += TTI->getShuffleCost(TTI::SK_ExtractSubvector, VecTy,
```
alexey-bataev: `VectorCost += TTI->getShuffleCost(TTI::SK_ExtractSubvector, VectorTy,`

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -29316,19 +29331,15 @@ class HorizontalReduction {
           if (auto *VecTy = dyn_cast<FixedVectorType>(ScalarTy)) {
             assert(SLPReVec && "FixedVectorType is not expected.");
             unsigned ScalarTyNumElements = VecTy->getNumElements();
+            auto *DstTy = FixedVectorType::get(VecTy->getScalarType(),
+                                               ReducedVals.size());
             for (unsigned I : seq<unsigned>(ReducedVals.size())) {
```
alexey-bataev, line 30378: `  for (unsigned I : seq<unsigned>(ReducedVals.size() - 1)) {`

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -28219,6 +28219,7 @@ class HorizontalReduction {
         // Also, do not try to reduce const values, if the operation is not
         // foldable.
         bool IsReducedVal = !EdgeInst || Level > RecursionMaxDepth ||
+                            R.isAnalyzedReductionRoot(EdgeInst) ||
```
alexey-bataev: Too aggressive, affects not only revec, but all reductions and may cause regressions

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
alexey-bataev: Better to gather all these values to a separate vector and return in a parameter and then consider (or pass a callback)

alexey-bataev (changes requested):

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
alexey-bataev: Most probably, Transforms/SLPVectorizer/RISCV/complex-loads.ll regression is related to this check. You cannot rely on llvm-mca, it does not work correctly for RISCV vectors, AFAIK

sushgokh: Yes, the regression in Transforms/SLPVectorizer/RISCV/complex-loads.ll is coming from this addition. I am finding it difficult to debug why the regression so took the route of mca to check if this is real regression.
But will check again if I can debug the regression

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
sushgokh: The problem is not about gathering these values.

The problem is I am creating the same reduction operations as they were before revec, with the final reduction operation created [here](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp#L29104). Ideally, I should be able to mark `V.analyzedReductionRoot()` after this final step but this is difficult at the moment. If I dont do this, it goes into infinite loop.

So, I am trying to do that [here](https://github.com/llvm/llvm-project/pull/197875/changes#diff-e3b933e8c46156c0eaf7cbb67866b712f69a8484bc941d10acec9d4d17b9f061L28221) but as you say, its regressing.

I will check again why its regressing

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
sushgokh: Honestly, I dont understand if the node has marked as already analysed for reduction ,why do we need to asses it again ?

alexey-bataev: We can do it only if CFG is a tree, the problem is that it is actually a graph and the same instruction might be the part of different paths in this graph. In the first one it might be non-profitable/non-legal, but it might be legal and profitable in some other paths.

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
sushgokh: done

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
sushgokh: removed

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
sushgokh: done

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
sushgokh: done

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
sushgokh: done

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
alexey-bataev: I'm not saying it is the problem, I'm suggesting better approach to what you want to get. Do not use analyzedReductionRoot for your purpose, use a different (special) storage.

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
alexey-bataev: Where?

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -30101,6 +30095,9 @@ bool SLPVectorizerPass::vectorizeHorReduction(
       Res = true;
       if (auto *I = dyn_cast<Instruction>(VectorizedV); I && I != Inst) {
         // Try to find another reduction.
+        if (SLPReVec && I->getType()->isVectorTy() &&
+            Inst->getType()->isVectorTy())
+          continue;
```
alexey-bataev: Why?

sushgokh: The only thing I want to express is "dont revisit these newly created instructions". With SLPReVec enabled, this check solves the problem I believe if thats what you are asking ?

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
sushgokh: I didnt get you. 

I have updated the comment in the latest commit. Isnt this C++ style ?

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
alexey-bataev: I understand, but in this case we may miss some candidates, better to store processed elements somewhere and check explicitly

sushgokh: ok sure

sushgokh: addressed this by having a seperate container to store these values

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -27764,6 +27760,8 @@ class HorizontalReduction {
   ReductionOpsListType ReductionOps;
   /// List of possibly reduced values.
   SmallVector<SmallVector<Value *>> ReducedVals;
+  /// List of values to ignore while collecting reduction values.
+  inline static SmallDenseSet<Value *> IgnoredVals;
```
alexey-bataev: Why static, make it local, just like all others.

sushgokh: Consider this sequence:
1. [This ](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp#L30910)calls reduce
2. Step 1 marks the values to be ignored and also pushes the instruction on stack
3. The same API call, in step 1, creates new  `HorizontalReduction Rdx` [here](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp#L30865) and we would lose the values to be ignored. 

These values need to persist  within the BB and hence static

alexey-bataev: And what's the problem with it? Local storage should be enough.

sushgokh: If we lose the values to be ignored, we can run into infinite case. See the example in the godbolt link in the PR description(although this bails out on Reduction width and Reduction limit , by hard coding these values, I can simulate the infinite bahviour)

alexey-bataev: It should not be solved by static lifetime, if so.

sushgokh: > It should not be solved by static lifetime, if so.

Sorry, didnt get it. 

If you saying that it will not solve the problem I am citing, its indeed solving the issue.

If you are hinting that this is not the right approach, you can suggest the right approach.

sushgokh: @alexey-bataev [This ](https://godbolt.org/z/cG8efq8ec) is short repro where the newly implemented revec pattern for horizontal reduction(i.e. shuffle + original reduction ops) runs into infinity.
Compile this with `./opt  -mtriple=aarch64 -p slp-vectorizer -slp-revec  -slp-threshold=-50`

Could you check and let me know how to avert this ?

alexey-bataev:
Something like should help:
```
--- a/llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
+++ b/llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
@@ -17949,6 +17949,19 @@ bool BoUpSLP::isTreeTinyAndNotFullyVectorizable(bool ForReduction) const {
     return true;
   }

+  if (SLPReVec && ForReduction && VectorizableTree.size() == 3 &&
+      VectorizableTree[0]->State == TreeEntry::Vectorize &&
+      VectorizableTree[0]->getOpcode() == Instruction::ShuffleVector &&
+      VectorizableTree[1]->isGather() &&
+      isSplat(VectorizableTree[1]->Scalars) && VectorizableTree[2]->isGather() &&
+      allConstant(VectorizableTree[2]->Scalars)) {
+    LLVM_DEBUG(
+        dbgs()
+        << "SLP: rejecting reduction tree with 3 nodes (1st - shuffle and the "
+           "second and the third nodes are gathers).\n");
+    return true;
+  }
+
```

sushgokh:
> Something like should help:
> 
> ```
> --- a/llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
> +++ b/llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
> @@ -17949,6 +17949,19 @@ bool BoUpSLP::isTreeTinyAndNotFullyVectorizable(bool ForReduction) const {
>      return true;
>    }
> 
> +  if (SLPReVec && ForReduction && VectorizableTree.size() == 3 &&
> +      VectorizableTree[0]->State == TreeEntry::Vectorize &&
> +      VectorizableTree[0]->getOpcode() == Instruction::ShuffleVector &&
> +      VectorizableTree[1]->isGather() &&
> +      isSplat(VectorizableTree[1]->Scalars) && VectorizableTree[2]->isGather() &&
> +      allConstant(VectorizableTree[2]->Scalars)) {
> +    LLVM_DEBUG(
> +        dbgs()
> +        << "SLP: rejecting reduction tree with 3 nodes (1st - shuffle and the "
> +           "second and the third nodes are gathers).\n");
> +    return true;
> +  }
> +
> ```

Thanks @alexey-bataev Added this patch.

alexey-bataev:
Rebase

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
alexey-bataev: No, Use `//` style, you're using C-style

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -30261,25 +30276,30 @@ class HorizontalReduction {
...
+      */
+      for (auto I : seq<unsigned>(VF)) {
+        auto Position = I * DestTyNumElements;
+        Value *SubVec =
+            createExtractVector(Builder, Vec, DestTyNumElements, Position);
+        if (!Rdx) {
+          Rdx = SubVec;
+        } else {
+          Rdx = createOp(Builder, RdxKind, Rdx, SubVec, "rdx.op", ReductionOps);
+        }
```
alexey-bataev: Drop extra braces

sushgokh: done

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
sushgokh: done

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -30379,19 +30402,13 @@ class HorizontalReduction {
...
-                  VectorTy,
-                  createStrideMask(I, ScalarTyNumElements, ReducedVals.size()));
-              VectorCost += TTI->getArithmeticReductionCost(RdxOpcode, VecTy,
-                                                            FMF, CostKind);
+            for (unsigned I : seq<unsigned>(ReducedVals.size() - 1)) {
+              VectorCost +=
+                  ::getShuffleCost(*TTI, TTI::SK_ExtractSubvector, VectorTy, {},
+                                   CostKind, I * ScalarTyNumElements, VecTy);
+              VectorCost +=
+                  TTI->getArithmeticInstrCost(RdxOpcode, VecTy, CostKind);
```
alexey-bataev: Arithemtic cost is ok, but need one extra for extract subvector

alexey-bataev: Restore FMF in getArithmeticInstrCost!

llvm/test/Transforms/SLPVectorizer/revec.ll
```diff
@@ -474,3 +469,32 @@ entry:
...
+;
+entry:
+  %gep1 = getelementptr <4 x i32>, ptr %a, i64 1
+  %gep2 = getelementptr <4 x i32>, ptr %a, i64 2
+  %gep3 = getelementptr <4 x i32>, ptr %a, i64 3
+  %a0 = load <4 x i32>, ptr %a
+  %a1 = load <4 x i32>, ptr %gep1
+  %a2 = load <4 x i32>, ptr %gep2
+  %a3 = load <4 x i32>, ptr %gep3
+  %add0 = add <4 x i32> zeroinitializer, %a0
```
alexey-bataev: Could you replace zeroinitalizer with something non-foldable here?

sushgokh: done

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
sushgokh: done. Would it be worth considering FMF when we are computing scalar cost as well(currently we arent doing it) ?

sushgokh: I changed the iteration count.

Now `seq<unsigned>(ReducedVals.size())`  would run from `[0  ReducedVals.size()-1)` . So, for the added test case, extracts are considered 4 times. 

Similarly, arithmetic cost would be considered 1 less number of times.

alexey-bataev: Yes, but not a common FMF, just an individual one

sushgokh: done

alexey-bataev:
Rebase

alexey-bataev (approved):

Andarwinux:
```llvm
define <16 x float> @a(ptr %0, ptr %1) {
  %3 = load <16 x float>, ptr %0, align 64
  %4 = fadd <16 x float> %3, %3
  %5 = load <16 x float>, ptr %1, align 64
  %6 = fadd <16 x float> %4, %5
  %7 = fadd <16 x float> %6, zeroinitializer
  %8 = fadd <16 x float> %7, splat (float 1.000000e+00)
  ret <16 x float> %8
}
```

```
fatal error: error in backend: Do not know how to split this operator's operand!

PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace and dumped files.
Stack dump:
0.      Program arguments: clang -O3 -mllvm -slp-revec -w -c reduced.ll
1.      Code generation
2.      Running pass 'Function Pass Manager' on module 'reduced.ll'.
3.      Running pass 'X86 DAG->DAG Instruction Selection' on function '@a'
#0 0x00005b04c3dd7b96 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) Signals.cpp:0:0
clang: error: clang frontend command failed with exit code 70 (use -v to see invocation)
```
```
fatal error: error in backend: Cannot emit physreg copy instruction
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace and dumped files.
Stack dump:
0.      Program arguments: clang -march=tigerlake -O3 -mllvm -slp-revec -w -c reduced.ll
1.      Code generation
2.      Running pass 'Function Pass Manager' on module 'reduced.ll'.
3.      Running pass 'Post-RA pseudo instruction expansion pass' on function '@a'
#0 0x00005d4b923d7b96 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) Signals.cpp:0:0
clang: error: clang frontend command failed with exit code 70 (use -v to see invocation)
```

Looks like this PR will cause x86 backend crash with -slp-revec.

https://github.com/llvm/llvm-project/issues/203195

sushgokh:
> ```llvm
> define <16 x float> @a(ptr %0, ptr %1) {
>   %3 = load <16 x float>, ptr %0, align 64
>   %4 = fadd <16 x float> %3, %3
>   %5 = load <16 x float>, ptr %1, align 64
>   %6 = fadd <16 x float> %4, %5
>   %7 = fadd <16 x float> %6, zeroinitializer
>   %8 = fadd <16 x float> %7, splat (float 1.000000e+00)
>   ret <16 x float> %8
> }
> ```
> 
> ```
> fatal error: error in backend: Do not know how to split this operator's operand!
> 
> PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace and dumped files.
> Stack dump:
> 0.      Program arguments: clang -O3 -mllvm -slp-revec -w -c reduced.ll
> 1.      Code generation
> 2.      Running pass 'Function Pass Manager' on module 'reduced.ll'.
> 3.      Running pass 'X86 DAG->DAG Instruction Selection' on function '@a'
> #0 0x00005b04c3dd7b96 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) Signals.cpp:0:0
> clang: error: clang frontend command failed with exit code 70 (use -v to see invocation)
> ```
> 
> ```
> fatal error: error in backend: Cannot emit physreg copy instruction
> PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace and dumped files.
> Stack dump:
> 0.      Program arguments: clang -march=tigerlake -O3 -mllvm -slp-revec -w -c reduced.ll
> 1.      Code generation
> 2.      Running pass 'Function Pass Manager' on module 'reduced.ll'.
> 3.      Running pass 'Post-RA pseudo instruction expansion pass' on function '@a'
> #0 0x00005d4b923d7b96 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) Signals.cpp:0:0
> clang: error: clang frontend command failed with exit code 70 (use -v to see invocation)
> ```
> 
> Looks like this PR will cause x86 backend crash with -slp-revec.
> 
> #203195

Replied on the ticket. The crash isnt from this commit. See [here](https://godbolt.org/z/hGMd64YqG)

PR #197879, opened 2026-05-15
[AArch64][CostModel] Model extension costs with narrow int sdiv/udiv/srem/urem

yashssh:
AArch64 promotes i8/i16 to legal i32 types but the cost model does not model that

Example with sdiv https://godbolt.org/z/dcGvWPad1


yashssh:
Similar issues exist with other scalar instructions like max/min, shift operations, abs, compare instructions. 
If this change looks reasonable I can submit a fix for them re-using `getNarrowIntPromoteCost` functionality added here.

llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
```diff
@@ -4329,6 +4329,38 @@ InstructionCost AArch64TTIImpl::getScalarizationOverhead(
...
+  InstructionCost ExtCost = getCastInstrCost(
+      ExtOpc, PromotedTy, Ty, TTI::CastContextHint::None, CostKind);
+
+  InstructionCost Cost = 0;
+  auto NeedExt = [&](TTI::OperandValueInfo OI, unsigned Idx) {
+    if (OI.isConstant())
+      return false;
+    if (Idx < Args.size() &&
+        (isa<SExtInst>(Args[Idx]) || isa<ZExtInst>(Args[Idx]) ||
+         isa<LoadInst>(Args[Idx])))
```
paulwalker-arm, line 4351: This looks far too coarse grained. It does not check if the new extension matches the original nor does it care if the existing extends have other uses.  I think we will be better off with a variant of `getCastInstrCost` that takes the operand to be extended which can then use a refactored `isExtFree`.

That said, this seems like a lot of refactoring so I wonder if it is really worth it?  Is the promotion the important missing cost or is it the fact that promoting vector types is likely to lead to the operation being split and thus doubling the cost of the divide?

llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
```diff
@@ -4329,6 +4329,38 @@ InstructionCost AArch64TTIImpl::getScalarizationOverhead(
...
 
+std::optional<InstructionCost> AArch64TTIImpl::getNarrowIntPromoteCost(
+    Type *Ty, TTI::TargetCostKind CostKind, TTI::OperandValueInfo Op1Info,
+    TTI::OperandValueInfo Op2Info, ArrayRef<const Value *> Args, bool IsSigned,
+    std::function<InstructionCost(Type *)> InstCost) const {
+
+  if (!Ty->isIntegerTy() || Ty->getScalarSizeInBits() >= 32)
+    return std::nullopt;
+
+  Type *PromotedTy = Ty->getWithNewBitWidth(32);
```
paulwalker-arm, line 4340: This assumes all i8/i16 element based operations require promotion.  However, the relevant cost code details several cases when no promotion is required because the divide can be simulated using other instructions (e.g. asrd).

paulwalker-arm (commented):
My main worry with this patch is the current costing of divide operations looks intricate with much special handling to the point I'm not sure global promotion handling is going to be precise enough.

davemgreen (commented):
I only looked at this very briefly but it wasn't clear if it was making things a lot more accurate. Our existing div costs are not very meaningful and closer to codesize than they are to recip throughput. The cost of the divide should be dominated by the div part.

PR #197883, opened 2026-05-15
[LIT] Add --load-limit N option

MaximPavlenko:
**Problem:** 
When LIT runs multiple instances on a single server, it's impossible to dynamically manage the overall server load. This can be a problem for automated testing.

**Solution:**
Add an option which can dynamically control the number of threads using machine `load average` information.
`--load-limit (0..2]` -- float value which is > 0, and max value is 2 (200%).
`--load-limit 0.1` -- use 10% of CPU
`--load-limit 0.8` -- use 80% of CPU
`--load-limit 1`    -- use 100% of CPU
`--load-limit 2`    -- use 200% of CPU
`--load-limit abc` -- Error
`--load-limit -1`   -- Error
`--load-limit 0`    -- Error
`--load-limit 2.1` -- Error

Also `-jmin/--workers_min/--threads_min M` options is added. It sets the minimum number of threads when load average is higher than a limit.

**Main use case:**
`-j NPROC -l NPROC -jmin NPROC/2` -- this will utilize machine on 100%, but if additional load will be added, then LIT will free some recourses.

**Behaviour:**
1. If `--load-limit N` is not passed -- no changes
2. If `-j 10 --load-limit 20` -- LIT will run with 10 threads, if load will grow > 20, then LIT will start to reduce number of threads. If load will be lower than 20, then LIT will start to increase number of threads, but not more than 10.
3. If `-j 20 --load-limit 10` -- LIT will start with 10 threads and will control number of threads using machine load-average information. Can be useful if jobs wait something.
4. 2 instances of LIT with  `-j 20 --load-limit 20` -- both will start with 20 threads, and will reduce number of threads based on machine load average information. 
NOTE: The 1-minute load average has some lag, so there will always be fluctuations in the number of threads, especially at the beginning.

**TODO:**
1. Debug info can be deleted

### DIFF
--- a/llvm/utils/lit/lit/cl_arguments.py
+++ b/llvm/utils/lit/lit/cl_arguments.py
@@ -95,6 +95,16 @@ def parse_args():
         default=os.getenv("LIT_MAX_WORKERS", lit.util.usable_core_count()),
     )
     parser.add_argument(
+        "-jmin",
+        "--threads_min",
+        "--workers_min",
+        dest="workers_min",
+        metavar="N",
+        help="Minimal number of workers used for testing if `--load-limit FRAC` is enabled",
+        type=_positive_int,
+        default=os.getenv("LIT_MIN_WORKERS", 1),
+    )
+    parser.add_argument(
         "-l",
         "--load-limit",
         dest="load_limit_fraction",
@@ -539,6 +549,14 @@ def parse_args():
         print(
             "WARNING: --incremental is deprecated. Failing tests now always run first."
         )
+
+    if opts.workers_min > opts.workers:
+        print(
+            "WARNING: --threads_min %d is bigger than --threads %d. "
+            "Used default --threads_min 1."
+            % (opts.workers_min, opts.workers)
+        )
+        opts.workers_min = 1
 
     if opts.load_limit_fraction is not None and not hasattr(os, "getloadavg"):
         print(
--- a/llvm/utils/lit/lit/cl_arguments.py
+++ b/llvm/utils/lit/lit/cl_arguments.py
@@ -95,6 +95,22 @@ def parse_args():
         default=os.getenv("LIT_MAX_WORKERS", lit.util.usable_core_count()),
     )
     parser.add_argument(
+        "-l",
+        "--load-limit",
+        dest="load_limit_fraction",
+        metavar="FRAC",
+        help="Aim to keep the 1-minute system load average near "
+        "FRAC * number-of-CPU-cores, by smoothly adjusting the number "
+        "of concurrent tests up or down (similar to 'make -l'). FRAC "
+        "must be in (0, 2] -- e.g. 0.8 targets 80%% of CPU capacity, "
+        "2 targets 200%% of CPU capacity (Not recommended). "
+        "Concurrency is capped at -j workers and never less than 1, so "
+        "progress is never blocked. Ignored on platforms without "
+        "os.getloadavg().",
+        type=_bounded_float(0, 2, lo_inclusive=False, hi_inclusive=True),
+        default=os.getenv("LIT_LOAD_LIMIT"),
+    )
+    parser.add_argument(
         "--config-prefix",
         dest="configPrefix",
         metavar="NAME",
@@ -524,6 +540,13 @@ def parse_args():
             "WARNING: --incremental is deprecated. Failing tests now always run first."
         )
 
+    if opts.load_limit_fraction is not None and not hasattr(os, "getloadavg"):
+        print(
+            "WARNING: --load-limit is ignored: os.getloadavg() is not "
+            "available on this platform."
+        )
+        opts.load_limit_fraction = None
+
     if opts.numShards or opts.runShard:
         if not opts.numShards or not opts.runShard:
             parser.error("--num-shards and --run-shard must be used together")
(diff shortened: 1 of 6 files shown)


MaximPavlenko:
Hi @boomanaiden154, Please assign a reviewer.

MaximPavlenko:
Some plots from 96 cores machine.
IMG_20260515_185932

boomanaiden154 (changes requested):
This is a lot of complexity for a relatively niche feature. It also wouldn't hurt to collect performance numbers now that we're doing interprocess coordination on every test.

The use case is also unclear to me. In the case someone else tries to run something on the system, why can we not just let the OS level scheduler take care of it? Or do what is done traditionally where you just set the maximum parallelism to what can actually be sustained?

Also, was this patch prepared with AI assistance?

MaximPavlenko:
I'm not agree that this is a niche feature. It gives additional controls so it makes LIT more mature.
My use case is that we use the same infrastructure for development and tests, and to not to overload servers (to keep it usable for a development) we need to reduce '-j' parameter to keep load average ~100%.
With this changes you do not need to do a fine tuning of different tests, it is enough to set max, min threads and a load target.
Also `make` has `-l` option and we use it too in our tests.

About complexity. In general this patch adds a function (_acquire_run_slot()) which blocks thread if the load should be reduced. All other stuff is passing parameters to this function.

AI was used, but code was simplified after that. And I think comments should be rewritten to be more compact.

Performance:
For case without `--load-limit` there should be no big diff:
- call _acquire_run_slot()
- check if opt is on and parameters are not None    << it can be moved outside of function to avoid call & return.
- return

For case with option:
I'll collect this on Monday. I didn't notice any performance degradation.

MaximPavlenko:
Hi @boomanaiden154, I did some performance tuning and prepared the numbers.

Without `--load-limit` average execution time for additional functions is about 0.000001 seconds (1 additional second for 1 million of tests). 
```
python /SCRATCH/pavlenko/GIT_llvm_project_pavlenko/llvm-project/llvm/utils/lit/lit.py -j 10 --xunit-xml-output=results.xml -v
-- Testing: 32 tests, 10 workers --
Execution time: 0.00000112 seconds
Execution time: 0.00000107 seconds
Execution time: 0.00000083 seconds
Execution time: 0.00000095 seconds
Execution time: 0.00000124 seconds
Execution time: 0.00000073 seconds
Execution time: 0.00000098 seconds
Execution time: 0.00000086 seconds
Execution time: 0.00000072 seconds
Execution time: 0.00000101 seconds
Execution time: 0.00000098 seconds
Execution time: 0.00000079 seconds
Execution time: 0.00000125 seconds
Execution time: 0.00000122 seconds
Execution time: 0.00000069 seconds
Execution time: 0.00000092 seconds
Execution time: 0.00000118 seconds
Execution time: 0.00000061 seconds
Execution time: 0.00000140 seconds
Execution time: 0.00000145 seconds
Execution time: 0.00000073 seconds
Execution time: 0.00000103 seconds
Execution time: 0.00000103 seconds
Execution time: 0.00000077 seconds
Execution time: 0.00000119 seconds
Execution time: 0.00000091 seconds
Execution time: 0.00000066 seconds
Execution time: 0.00000085 seconds
Execution time: 0.00000097 seconds
Execution time: 0.00000110 seconds
Execution time: 0.00000067 seconds
Execution time: 0.00000060 seconds

Testing Time: 9.80s

Total Discovered Tests: 32
  Passed: 32 (100.00%)
```

With `--load-limit` average time is about 0.0001 seconds (1 additional second for 10k of tests). 
```
python /SCRATCH/pavlenko/GIT_llvm_project_pavlenko/llvm-project/llvm/utils/lit/lit.py -j 10 --load-limit=1 --xunit-xml-output=results.xml -v
-- Testing: 32 tests, 10 workers, load-limit 100% --
Execution time: 0.00013140 seconds
Execution time: 0.00005448 seconds
Execution time: 0.00004627 seconds
Execution time: 0.00011769 seconds
Execution time: 0.00005913 seconds
Execution time: 0.00004911 seconds
Execution time: 0.00011352 seconds
Execution time: 0.00005203 seconds
Execution time: 0.00004674 seconds
Execution time: 0.00011000 seconds
Execution time: 0.00005944 seconds
Execution time: 0.00004809 seconds
Execution time: 0.00011738 seconds
Execution time: 0.00006081 seconds
Execution time: 0.00005607 seconds
Execution time: 0.00011179 seconds
Execution time: 0.00005668 seconds
Execution time: 0.00003882 seconds
Execution time: 0.00010930 seconds
Execution time: 0.00004829 seconds
Execution time: 0.00004156 seconds
Execution time: 0.00010550 seconds
Execution time: 0.00005341 seconds
Execution time: 0.00004537 seconds
Execution time: 0.00011030 seconds
Execution time: 0.00005810 seconds
Execution time: 0.00005060 seconds
Execution time: 0.00007038 seconds
Execution time: 0.00013342 seconds
Execution time: 0.00008614 seconds
Execution time: 0.00005929 seconds
Execution time: 0.00004499 seconds

Testing Time: 9.93s

Total Discovered Tests: 32
  Passed: 32 (100.00%)
```

How measurements were done:
```
def execute(test):
    start_time = time.perf_counter()
    if _load_limit:
        _acquire_run_slot()
    end_time = time.perf_counter()
    tmp_time = end_time - start_time
    try:
        with _get_parallelism_semaphore(test):
            result = _execute(test, _lit_config)
    finally:
        start_time = time.perf_counter()
        if _load_limit:
            _release_run_slot()
        end_time = time.perf_counter()
        tmp_time_2 = end_time - start_time
    print(f"Execution time: {tmp_time + tmp_time_2:.8f} seconds")

    test.setResult(result)
    return test
```

PR #197885, opened 2026-05-15
[Flang][OpenMP] Restrict implicit default declare mapper from applying deep-copies of pointer members

agozillon:
According to the OpenMP specification, only allocatables should get deep-copy behaviour inside of implicit default declare mappers. This PR restricts this behaviour. Relevant specification exert, added as a comment for a reminder:

    // "If a component of a derived type list item is a map clause list item
    // that results  from the predefined default mapper for that derived type,
    // and the component is not also an explicit list item or the array base
    // of an explicit list item on the same construct, then: if it has the
    // POINTER attribute, it is attach-INELIGIBLE. If a list item in a map
    // clause is an associated pointer that is attach-ineligible, the effect
    // of the map clause does not apply to its pointer target."

This prevents certain programs from unexpected over-mapping via pointer nesting, doesn't prevent that for allocatables, but that's OpenMP specification mandated foot shooting, so it's free game.

### DIFF
--- a/flang/lib/Utils/OpenMP.cpp
+++ b/flang/lib/Utils/OpenMP.cpp
@@ -228,6 +228,36 @@ mlir::FlatSymbolRefAttr getOrGenImplicitDefaultDeclareMapper(
   for (const auto &entry : llvm::enumerate(recordType.getTypeList())) {
     const auto &memberName = entry.value().first;
     const auto &memberType = entry.value().second;
+
+    // OpenMP 5.0, 5.1, 5.2: Default Map Clause
+    //
+    // "If a component of a derived type list item is a map clause list item
+    // that results  from the predefined default mapper for that derived type,
+    // and the component is not also an explicit list item or the array base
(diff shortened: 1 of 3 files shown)


TIFitis (approved):
LGTM, thanks for the PR.

Should we add a small test like the one below to check behaviour when both implicit and explicit mapping is used in the same construct?
I'm expecting `obj` still uses the implicit default mapper, and we also emit an explicit map entry for `obj%ptr_scalar` (i.e. explicit component mapping is not suppressed by the new implicit-pointer skip).

```
!$omp target map(mapper(default), tofrom: obj) map(tofrom: obj%ptr_scalar)
    obj%id = obj%id + 1
!$omp end target
```

agozillon:
Thank you very much for the quick review and great suggestion @TIFitis I've updated the PR to include a test covering that case! I'll wait until Tuesday before landing to see if anyone else (or yourself) has any further feedback.

PR #197887, opened 2026-05-15
[lldb] Skip new tests in TestLinuxCore on Windows and non-x86 targets

slydiman:
This patch must fix broken tests after #197341 on buildbots lldb-remote-linux-win, lldb-x86_64-win and lldb-aarch64-windows.

### DIFF
--- a/lldb/test/API/functionalities/postmortem/elf-core/TestLinuxCore.py
+++ b/lldb/test/API/functionalities/postmortem/elf-core/TestLinuxCore.py
@@ -1284,6 +1284,8 @@ class LinuxCoreTestCase(TestBase):
 
         self.dbg.DeleteTarget(target)
 
+    @skipIfLLVMTargetMissing("X86")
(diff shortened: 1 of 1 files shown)


DavidSpickett:
To load an x86 core you need the llvm target. Makes sense.

Why can the test not be run on Windows, something to do with paths? Or just a temporary disable while we figure it out?

slydiman:
> Why can the test not be run on Windows, something to do with paths? Or just a temporary disable while we figure it out?

No idea. I expected that the author of #197341 would take care of this. 
For now I'm trying to make buildbots green ASAP. 
But I did not create any new issues regarding running these tests on Windows.

DavidSpickett (approved):
Fair enough, let's fix the bots then.

slydiman:
1 test is still broken. See https://github.com/llvm/llvm-project/pull/197341#issuecomment-4461988138 for details.

PR #197888, opened 2026-05-15
Llvm-objcopy crashes with unreachable error when (de)compression requested without selected value available

### LINKED ISSUE #197877: llvm-objcopy crashes with unreachable error when (de)compression requested without selected value available
llvm-objcopy crashes due to calling `llvm_unreachable` when `compression::compress` or `compression::decompress` is called without the selected compression type being installed on the machine.

The problem is that the tool doesn't call `compression::getReasonIfUnsupported` before `compression::compress` or `compression::decompress`. The correct thing to do is to call that first, then check the result and either report an Error based on the result or continuing and (de)compress as requested.

Discovered while triaging #197813.

I'll label this with `good first issue` as it's a trivial fix (but I don't have the personal time to fix it and put a PR up myself). There should be two tests for it. The first test should test that the correct error is emitted for zlib compression and decompression and the second the same for zstd. Each test should only run if zlib/zstd is NOT enabled on the target (i.e. something like `UNSUPPORTED: zlib`, but check other examples for REQUIRES/UNSUPPORTED tags to see the exact way of writing that).

moaedy:
## Summary
Root cause: `Object::compressOrDecompressSections` constructs `CompressedSection` (whose constructor calls `compression::compress`) without first checking `compression::getReasonIfUnsupported`, so `--compress-sections <glob>=zlib|zstd` triggers `llvm_unreachable` when the requested format is unsupported in the current LLVM build.

Change: Added a `compression::getReasonIfUnsupported(compression::formatFor(*CType))` guard in `compressOrDecompressSections` (`llvm/lib/ObjCopy/ELF/ELFObjcopy.cpp`) before queueing a `CompressedSection`, returning a `createStringError` that names the offending section and reason.

Closes https://github.com/llvm/llvm-project/issues/197877

### DIFF
--- a/llvm/lib/ObjCopy/ELF/ELFObjcopy.cpp
+++ b/llvm/lib/ObjCopy/ELF/ELFObjcopy.cpp
@@ -242,6 +242,13 @@ Error Object::compressOrDecompressSections(const CommonConfig &Config) {
         ToReplace.emplace_back(
             &Sec, [=] { return &addSection<DecompressedSection>(*CS); });
     } else if (*CType != DebugCompressionType::None) {
+      // compression::compress would call llvm_unreachable if the requested
+      // format is unsupported in this LLVM build, so reject it here.
+      if (const char *Reason = compression::getReasonIfUnsupported(
+              compression::formatFor(*CType)))
+        return createStringError(errc::invalid_argument,
+                                 "failed to compress section '" + Sec.Name +
+                                     "': " + Reason);
       ToReplace.emplace_back(&Sec, [=, S = &Sec] {
         return &addSection<CompressedSection>(
             CompressedSection(*S, *CType, Is64Bits));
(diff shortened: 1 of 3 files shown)


jh7370:
@moaedy, did you use AI to do this work? Please review the [LLVM AI policy](https://llvm.org/docs/AIToolPolicy.html#policy). When you have read that, please come back to this PR and make the appropriate changes. If you didn't use AI assistance, please explicitly say so.

moaed-kinertic:
Hi @jh7370, thanks for flagging this. Yes, I used AI assistance (Claude) on this PR — I reviewed and tested the change myself before submitting. I've read through the AI policy and will add an `Assisted-by: Claude <noreply@anthropic.com>` trailer to the commit before merge. Apologies for not disclosing upfront.

jh7370:
Hi @moaed-kinertic. Please go back over the AI Policy again, specifically noting the section on "good first issue" (which the issues this PR fixes is) and AI. Also, the bit about PR descriptions is useful to refer to for future reference. I'm not going to spell out the issues this PR currently has with relation to the policy, because I think it's important you have fully understood the policy (and the policy is quite clear about them).

jh7370:
@moaedy, do you have any update here, or shall I close this PR?

jh7370:
Standard rejection message due to AI policy violation:

This PR doesn't appear to comply with our policy on tool-generated content, and requires additional justification for why it is valuable enough to the project for us to review it. Please see our developer policy on AI-generated contributions: http://llvm.org/docs/AIToolPolicy.html

Specifically, this PR violates the policy on at least the following points:
* The author used an LLM to generate the description.
* The description does not disclose the use of an LLM.
* The PR is resolving a "good first issue" for which AI usage is specifically banned.

PR #197890, opened 2026-05-15
[RISCV] Make VSETTM/VSETTK not affect the VSETVL emit

BeMg:
VSETTM/TK will modify VTYPE, but it only affects the TM/TK bits. This modification is safe for other RVV operations. The TM/TK value will be maintained in insertVSETMTK.


topperc (approved):
LGTM

4vtomat (approved):
LGTM~

PR #197901, opened 2026-05-15
[lldb][windows] Plumb Windows DLL load/unload through lldb-server

charles-zablit:
`lldb-server.exe` currently never sends DLLs the inferior loads to the client. `NativeProcessWindows::OnLoadDll` just invalidates a cache and the client's module list stays empty. Tests like `lang/c/shared_lib/TestSharedLib` that call a function defined in a loaded DLL fail with `error: Couldn't look up symbols`.

## `NativeProcessProtocol` additions

* New`Extension::libraries` flag, separate from the existing SVR4-only `Extension::libraries_svr4`.
* `LoadedLibraryInfo { name, base_addr }` struct, mirroring `SVR4LibraryInfo` but without any of the SVR4-specific fields (link_map, ld_addr, next) that don't apply to PE.
* `GetLoadedLibraries()` virtual that returns the current libraries.
* `HasPendingLibraryEvents()` virtual, used by the stop-reply builder to decide whether to include `library:1;` in the packet.

## Event flow

1. Client does `continue`.
2. Windows inferior loads a DLL.
3. lldb-server's debug loop takes the `LOAD_DLL_DEBUG_EVENT`, `NativeProcessWindows::OnLoadDll` sets the pending flag and clears the module cache.
4. process eventually stops (breakpoint, signal,
etc.).
5. stop reply includes `library:1;`.
6. client parses it and calls `ProcessGDBRemote::LoadModules`.
7. that issues `qXfer:libraries:read`.
8. server serves the current snapshot.
9. client feeds the result into `DynamicLoaderWindowsDYLD` and the module list is up to date.

## Requirements

This requires libxml2 when running the tests with `LLDB_USE_LLDB_SERVER=1`.

## External resources

See https://sourceware.org/gdb/current/onlinedocs/gdb.html/Library-List-Format.html#Library-List-Format for the `library` key documentation.

rdar://177343924

### DIFF
--- a/lldb/source/Plugins/Process/gdb-remote/GDBRemoteCommunicationServerLLGS.cpp
+++ b/lldb/source/Plugins/Process/gdb-remote/GDBRemoteCommunicationServerLLGS.cpp
@@ -1054,8 +1054,10 @@ GDBRemoteCommunicationServerLLGS::PrepareStopReplyPacketForThread(
                     tid_stop_info.details.fork.child_tid);
   }
 
-  if (process.HasPendingLibraryEvents())
+  if (process.HasPendingLibraryEvents()) {
+    // 1 is an arbitrary value here. The parameter is ignored.
     response.PutCString("library:1;");
+  }
 
   return response;
 }
--- a/lldb/source/Plugins/Process/gdb-remote/GDBRemoteCommunicationServerLLGS.cpp
+++ b/lldb/source/Plugins/Process/gdb-remote/GDBRemoteCommunicationServerLLGS.cpp
@@ -1054,6 +1054,9 @@ GDBRemoteCommunicationServerLLGS::PrepareStopReplyPacketForThread(
                     tid_stop_info.details.fork.child_tid);
   }
 
+  if (process.HasPendingLibraryEvents())
+    response.PutCString("library:1;");
+
   return response;
 }
 
@@ -3386,6 +3389,24 @@ GDBRemoteCommunicationServerLLGS::ReadXferObject(llvm::StringRef object,
     return MemoryBuffer::getMemBufferCopy(response.GetString(), __FUNCTION__);
   }
 
+  if (object == "libraries") {
+    auto library_list = m_current_process->GetLoadedLibraries();
+    if (!library_list)
+      return library_list.takeError();
+
+    StreamString response;
+    response.Printf("<library-list>");
+    for (auto const &library : *library_list) {
+      response.Printf("<library name=\"%s\">",
+                      XMLEncodeAttributeValue(library.name.c_str()).c_str());
+      response.Printf("<section address=\"0x%" PRIx64 "\"/>",
+                      library.base_addr);
+      response.Printf("</library>");
+    }
+    response.Printf("</library-list>");
+    return MemoryBuffer::getMemBufferCopy(response.GetString(), __FUNCTION__);
+  }
+
   if (object == "features" && annex == "target.xml")
     return BuildTargetXml();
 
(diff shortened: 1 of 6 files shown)


DavidSpickett:
> HasPendingLibraryEvents() virtual, used by the stop-reply builder to decide whether to include library:1; in the packet.

Is this how the srv4 version works or something you're adding to cope with something unique to Windows?

And that unique thing is that we cannot place a bunch of breakpoints in the loader like we do on Unix? And after doing so, instead of waiting for a stop reply packet we wait for breakpoint events.

This "library" reply should be added to https://lldb.llvm.org/resources/lldbgdbremote.html#lldb-extensions if it's new for lldb.

charles-zablit:
> Is this how the srv4 version works or something you're adding to cope with something unique to Windows?
> 
> And that unique thing is that we cannot place a bunch of breakpoints in the loader like we do on Unix? And after doing so, instead of waiting for a stop reply packet we wait for breakpoint events.

Yes.

`HasPendingLibraryEvents()` is a new addition. Linux/macOS set a breakpoint on `_dl_debug_state` which the client sees as a regular breakpoint hit. This breakpoint hit is a "libraries changed" event on Linux/macOS.

Windows does not have a public equivalent. Instead, the kernel raises `LOAD_DLL_DEBUG_EVENT` and `UNLOAD_DLL_DEBUG_EVENT` in the debugger's loop. We forward those to the client with `HasPendingLibraryEvents`.

> This "library" reply should be added to [lldb.llvm.org/resources/lldbgdbremote.html#lldb-extensions](https://lldb.llvm.org/resources/lldbgdbremote.html#lldb-extensions) if it's new for lldb.

There was an existing entry for `key == "library"` which I edited. I don't think this is an extension, it's referenced in the GDB manual: https://sourceware.org/gdb/current/onlinedocs/gdb.html/Stop-Reply-Packets.html.

DavidSpickett:
The docs definitely say that `library` means you should use `qXfer:libraries:read`. So what you've got here is fine.

What I was expecting to see is a key for the svr4 version, somewhere in GDB's docs. However I now think the svr4 version is meant to be initiated by the client actually, going by:
> This packet is optional for better performance on SVR4 targets. GDB uses memory read packets to read the SVR4 library list otherwise.

They just have a similar name and result format.

lldb/source/Plugins/Process/gdb-remote/GDBRemoteCommunicationServerLLGS.cpp
```diff
@@ -1054,6 +1054,9 @@ GDBRemoteCommunicationServerLLGS::PrepareStopReplyPacketForThread(
                     tid_stop_info.details.fork.child_tid);
   }
 
+  if (process.HasPendingLibraryEvents())
+    response.PutCString("library:1;");
```
DavidSpickett, line 1059: If I read this right, 1 is the `r` that is described as ignored here - https://sourceware.org/gdb/current/onlinedocs/gdb.html/Stop-Reply-Packets.html

> The packet indicates that the loaded libraries have changed. GDB should use ‘qXfer:libraries:read’ to fetch a new list of loaded libraries. The r part is ignored.

Add a comment saying that it's an arbitrary value.

DavidSpickett (approved):
Add a link to https://sourceware.org/gdb/current/onlinedocs/gdb.html/Library-List-Format.html#Library-List-Format in the PR description.

This gives us the name and the loaded address. Is that all lldb goes on or do we / could we dig some UUID like thing out of memory if we needed to? I don't have any concrete scenario that would require that, just wondering how it works.

Assuming the name plus whatever lldb already knows how to do gets the basics working, this LGTM.

charles-zablit:
> This gives us the name and the loaded address. Is that all lldb goes on or do we / could we dig some UUID like thing out of memory if we needed to? I don't have any concrete scenario that would require that, just wondering how it works.
> 
> Assuming the name plus whatever lldb already knows how to do gets the basics working, this LGTM.

The `ObjectFilePECOFF` instance is built from the what the client receives in `<library name="...">...`. `ObjectFilePECOFF::GetCoffUUID()` then computes the UUID: https://github.com/llvm/llvm-project/blob/de3ee84346d6dcf77ac20fe5c8acc95594886cbc/lldb/source/Plugins/ObjectFile/PECOFF/ObjectFilePECOFF.cpp#L140

This is enough for local debugging because the file on disk will match what's loaded. This matches what the `ProcessWindows` plugin does.

This will likely be a problem with remote debugging. We will want the server to extract the PDB GUID and sending it to the client. This can be done in a follow up PR by adding an attribute like `build-id="…"` to `qXfer:libraries:read`.

PR #197906, opened 2026-05-15
[CGCall] Initially store arg attrs using AttrBuilder (NFCI)

nikic:
Make the argument attribute more similar to fn/ret handling, by first populating an AttrBuilder and then converting it to AttributeSet once at the end, instead of using a lot of intermediate AttrBuilders. This also ensures we cannot lose any attributes because one code path overwrites another.


efriedma-quic (approved):
LGTM

PR #197909, opened 2026-05-15
[libc] Add struct sockaddr_in

labath:
The struct needs to be 16 bytes long for compatibility with the linux kernel (which rejects smaller sizes, even though the reset of the bytes are unused).

The padding field (and its name) is not specified by POSIX, but it's traditionally called sin_zero, and there exists a fair amount of code that references that name, so I'm matching it as well.

I'm testing the compatibility of this struct by binding to a localhost address. This test requires that the machine has a loopback interface with an assigned ipv4 address. If some of the environments do not have it, we can try to detect this in the test and skip it, but this would diminish the value of the test.

As a drive-by, I'm also adding the (non-POSIX) INADDR_LOOPBACK constant.

Assisted by Gemini.


libc/include/llvm-libc-macros/netinet-in-macros.h
```diff
@@ -30,6 +30,8 @@
 #define INADDR_ANY __LLVM_LIBC_CAST(static_cast, in_addr_t, 0x00000000)
 #define INADDR_BROADCAST __LLVM_LIBC_CAST(static_cast, in_addr_t, 0xffffffff)
 #define INADDR_NONE __LLVM_LIBC_CAST(static_cast, in_addr_t, 0xffffffff)
+// Not specified by POSIX
```
kaladron: It looks like this is 4.3BSD?  We should get say that if that's right (or as close as we can figure out) rather than what it's not.

libc/include/llvm-libc-types/struct_sockaddr_in.h
```diff
@@ -0,0 +1,35 @@
...
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_LIBC_TYPES_STRUCT_SOCKADDR_IN_H
+#define LLVM_LIBC_TYPES_STRUCT_SOCKADDR_IN_H
+
+#include "in_port_t.h"
+#include "sa_family_t.h"
+#include "struct_in_addr.h"
+
+// This is the sockaddr specialization for AF_INET sockets, as
```
kaladron: This isn't critical, but I wonder if this should be a doxygen comment on top (since we're introducing \file, I've been wondering what else we should be doing)  (OK to not do this, just as a thought)

kaladron (approved):
Two comments within, thanks!

libc/include/llvm-libc-types/struct_sockaddr_in.h
labath: LLVM hasn't been very consistent in use of doxygen comments, but yeah, this very much is a file-level comment, so let's move that there.

libc/include/llvm-libc-macros/netinet-in-macros.h
labath: AI seems to think that is the case, but I don't see any evidence supporting that (in fact, I see evidence disproving that). The closest I got was SVR4: https://github.com/calmsacibis995/svr4-src/blob/7dabeda6fc10bd1bbd1a84d502f05642b1bf0c9e/uts/i386/netinet/in.h#L198

libc/test/src/sys/socket/linux/bind_test.cpp
```diff
@@ -43,3 +46,32 @@ TEST_F(LlvmLibcBindTest, BindLocalSocket) {
...
+  int sock = LIBC_NAMESPACE::socket(AF_INET, SOCK_DGRAM, 0);
+  ASSERT_GE(sock, 0);
+  ASSERT_ERRNO_SUCCESS();
+  LIBC_NAMESPACE::cpp::scope_exit close_sock(
+      [&] { ASSERT_THAT(LIBC_NAMESPACE::close(sock), Succeeds(0)); });
+
+  struct sockaddr_in my_addr;
+  my_addr.sin_family = AF_INET;
+  my_addr.sin_port = 0;
+  // Avoid expanding the htonl macro in overlay mode.
```
labath, line 60: For visibility, I needed to work around the fact that htonl is a macro in overlay mode. LMK if we have a better way to deal with this.

PR #166397, opened 2025-11-04
[libc++] Simplify a bunch of noexcept specifications

philnik777:
Allocator copy/move operations can't throw according to [allocator.requirements], so we can just assume that they don't instead of checking. This removes a bunch of conditional noexcept specifications and simplifies even more.

PR #197912, opened 2026-05-15
[libc++] Make `basic_string`'s move constructor always `noexcept`

frederick-vs-ja:
Previously, libc++ made this move constructor unconditionally `noexcept` since C++17. However, it was already unconditionally `noexcept` in C++11/14 standard. See
- https://timsong-cpp.github.io/cppwp/n3337/string.cons
- https://timsong-cpp.github.io/cppwp/n4140/string.cons

So we should make it unconditionally `noexcept` even though we haven't treated [N4258](https://wg21.link/n4258) as a DR.

### DIFF
--- a/libcxx/include/string
+++ b/libcxx/include/string
@@ -105,8 +105,7 @@ public:
         noexcept(is_nothrow_default_constructible<allocator_type>::value);                      // constexpr since C++20
     explicit basic_string(const allocator_type& a);                                             // constexpr since C++20
     basic_string(const basic_string& str);                                                      // constexpr since C++20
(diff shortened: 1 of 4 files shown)


philnik777:
Is there a reason not to finish up #166397 instead? I know I've been sitting on it for a while, but maybe you can fine the time to update it?

frederick-vs-ja:
> Is there a reason not to finish up #166397 instead? I know I've been sitting on it for a while, but maybe you can fine the time to update it?

Oh, I'm sorry for forgetting it. I'll close this once your PR is updated.

PR #150125, opened 2025-07-22
[SelectionDAG] Verify SDTCisVT and SDTCVecEltisVT constraints

s-barannikov:
Teach `SDNodeInfoEmitter` TableGen backend to process `SDTypeConstraint` records and emit tables for them. The tables are used by `SDNodeInfo::verifyNode()` to validate a node being created.

This PR only adds validation code for `SDTCisVT` and `SDTCVecEltisVT` constraints to keep it smaller.

PR #197913, opened 2026-05-15
[RISCV] Fix fixed-length masked.{u,s}{div,rem} lowering not converting operands

lukel97:
Similar to #197724, but this time I also somehow forgot to convert the operands to scalable vectors. I'm surprised that nothing asserted here, since SDT_RISCVIntBinOp_VL has a type profile constraint that the operands and result types need to be the same.


pranavk:
i think this might be the cause of our internal failures. would be nice to get this in sooner.

topperc:
>  I'm surprised that nothing asserted here, since SDT_RISCVIntBinOp_VL has a type profile constraint that the operands and result types need to be the same.

Most of type profiles are only used to prune unnecessary checks from the DAGISel.inc. We don't yet autogenerate all checks to validate nodes. https://github.com/llvm/llvm-project/pull/150125 was a step in that direction.

llvm/test/CodeGen/RISCV/rvv/masked-udiv.ll
topperc: This test looks fragile. It's going to break if we add any constant folding for ISD::MASKED_UDIV

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
```diff
@@ -9008,9 +9008,14 @@ SDValue RISCVTargetLowering::LowerOperation(SDValue Op,
     MVT VT = Op.getSimpleValueType();
     MVT ContainerVT = getContainerForFixedLengthVector(VT);
     SDValue VL = getDefaultVLOps(VT, ContainerVT, DL, DAG, Subtarget).second;
-    SDValue Res = DAG.getNode(getRISCVVLOp(Op), DL, ContainerVT,
-                              Op.getOperand(0), Op.getOperand(1),
-                              DAG.getUNDEF(ContainerVT), Op.getOperand(2), VL);
+    SDValue Res = DAG.getNode(
+        getRISCVVLOp(Op), DL, ContainerVT,
+        convertToScalableVector(ContainerVT, Op.getOperand(0), DAG, Subtarget),
```
bababuck, line 9145: What if this already is a scalable type? Or is this not reachable in that case?

lukel97: This code path is only hit for fixed vector types, scalable vectors just use the tablegen patterns directly, their operation action is set to legal. `getContainerForFixedLengthVector` should assert if the type is already scalable

llvm/test/CodeGen/RISCV/rvv/masked-udiv.ll
lukel97: Sorry for the delay, replaced the constants with arguments, precommitted the test changes in 57d485710d121c088b76b5e80bdb544f185d4e2d and committed the test diff in 09cde6561923020feb1e5a35ba3739e3955edc11 to show there's still a codegen change

topperc (approved):
LGTM

wangpc-pp (approved):

PR #197921, opened 2026-05-15
[AArch64][llvm] Remove support for FEAT_MPAMv2_VID

jthackray:
`FEAT_MPAMv2_VID` instructions and system registers, as introduced in change d30f18d2c, are being removed at this time, as they've been removed from the latest Arm ARM, which doesn't preclude them returning in some form in future.

Other system registers introduced with `FEAT_MPAMv2` are unaffected, and these continue to be ungated, but since `+mpamv2` gating is now empty, I'm removing this superfluous gating code.

Cherry-picked-from: a48159df9


mgabka (approved):

dyung:
This seems like a "new" feature that you are looking to backport to the release branch which we are not accepting at this stage of the branch. What negative results would there be if we did not include this change in the 22.x release?

c-rhodes:
i also dont understand why this needs backporting. I can't see any harm in this remaining in 22 so it's not really worth the risk of any potential breakage from removing it.

jthackray:
Going to rename these to `mpamv2-deprecated` in another PR

Issue #148851, opened 2025-07-15
[LoopInterchange] Reject cases where ninf flag is present


madhur13490:
When this flag is present on instructions, semantics guarantee that input/output will not +/- inf. Interchange may lead to the result being +/- inf. I think when this flag is present on the instructions, we should reject interchange and bail out.

LLVM fast-math flags: https://llvm.org/docs/LangRef.html#fast-math-flags

This can be considered as an extension to https://github.com/llvm/llvm-project/pull/148612

madhur13490:
@kasuga-fj @sjoerdmeijer Does this make sense? If so, I would like to take a look.

kasuga-fj:
As I understand it, interchanging is rejected if there is a floating-point instruction for which the `reassoc` flag it not set.

https://github.com/llvm/llvm-project/blob/c2e41be50f2cc3d1e99c30b1996b8873b05c4b0e/llvm/lib/Transforms/Scalar/LoopInterchange.cpp#L824-L825

Do you have any example where this check is insufficient?

madhur13490:
https://godbolt.org/z/93ddenbM4
This is one of the tests you had added in the mentioned patch. `fadd` has both `ninf` and `reassoc` set. `reassoc` implies loop interchange can be performed, but `ninf` should disable it because interchange may lead to overflow. I am not sure if `reassoc` has more precedence.

kasuga-fj:
I didn't add such case in that patch, but it is just an oversight, and it looks to me that we should also take it into account.

As reading [the section about Rewrite-based flags](https://llvm.org/docs/LangRef.html#rewrite-based-flags), the handling of `ninf` flag appears to be independent of `reassoc`. It states that:

> These rules do not apply to the other fast-math flags. Whether or not a flag like `nnan` is present on any or all of the rewritten instructions is based on whether or not it is possible for said instruction to have a NaN input or output, given the original flags.

IIUIC, `reassoc` allows transformations based on associativity in mathematical sense, even if they can change the floating-point result. Therefore, remaining the `ninf` flag in your example without any validation seems to be incorrect (and maybe we can't make any guarantees about the resulting values). Probably the same applies to other fast-math flags (`fast`, `nnan` and `nsz`). However, I'm not sure which is better in this case: applying the interchange and removing the `ninf`, or avoiding it to preserve the `ninf`.

[...]

PR #197922, opened 2026-05-15
[LoopInterchange] Add test for poison can be produced due to ninf (NFC)

kasuga-fj:
This patch adds a test case illustrating that loop interchange can turn an input program that was previously well‑defined into one that exhibits UB, because reordering floating‑point operations in the presence of the ninf flag may produce a poison value. In such a case, we should drop the ninf from instructions involved in the transformation.
Also add another case to show that we can safely apply loop-interchange without eliminating ninf, because it doesn't change the order of floating-point operations chain.

Related issue: #148851


kasuga-fj:
* **#197923** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197923?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#197922** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197922?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197922?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

llvm/test/Transforms/LoopInterchange/ninf.ll
```diff
@@ -0,0 +1,154 @@
+; NOTE: Assertions have been autogenerated by utils/update_test_checks.py UTC_ARGS: --version 6
+; RUN: opt < %s -passes=loop-interchange -loop-interchange-profitabilities=ignore -S | FileCheck %s
+
+; sum = 0.0;
+; for (i = 0; i < 2; i++)
+;   for (j = 0; j < 2; j++)
+;     sum += A[i][j];
+;
+; The fadd instruction has reassoc flag, so it's leagal to interchange the
```
sjoerdmeijer: Typo: leagal

llvm/test/Transforms/LoopInterchange/ninf.ll
```diff
@@ -0,0 +1,154 @@
...
+  %sum.i.next = phi float [ %sum.j.next, %for.j ]
+  %i.inc = add i64 %i, 1
+  %ec.i = icmp eq i64 %i.inc, 2
+  br i1 %ec.i, label %exit, label %for.i.header
+
+exit:
+  ret float %sum.i.next
+}
+
+; for (i = 0; i < 2; i++)
```
sjoerdmeijer, line 82: Just a though on the test and its loop bounds. If we had a heuristic that would take loop bounds into account, this may become a fragile tests that would need changing. I would probably just use function arguments N and M as bounds or something like that. But I am not insisting, this should be fine for now.

sjoerdmeijer (approved):
LGTM

llvm/test/Transforms/LoopInterchange/ninf.ll
kasuga-fj: Fixed, thanks

llvm/test/Transforms/LoopInterchange/ninf.ll
kasuga-fj: That makes sense to me as well, but I’d prefer to leave the bounds as they are. This test specifies `-loop-interchange-profitabilities=ignore`, so no heuristic should affect the results. Also, using non-constant bounds makes the IR a bit longer, so I’d rather keep the constant ones.

PR #197923, opened 2026-05-15
[LoopInterchange] Drop ninf from instructions involved in interchange

### LINKED ISSUE #148851: [LoopInterchange] Reject cases where ninf flag is present

kasuga-fj:
Applying loop-interchange can alter the order of operations in reduction calculations. If these operations involve floating‑point arithmetic, the results may change as well. If an instruction in the chain has the `ninf` flag, it means that reordering can produce a poison value, which may lead to undefined behavior even though the original program is not.
This patch addresses this issue by dropping `ninf` flags from the instructions involved in the transformation, as discussed in #148851.

Fixes #148851.


kasuga-fj:
* **#197923** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197923?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197923?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#197922** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/197922?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

llvm/lib/Transforms/Scalar/LoopInterchange.cpp
```diff
@@ -986,6 +994,13 @@ static bool checkReductionKind(Loop *L, PHINode *PHI,
     case RecurKind::SMax:
     case RecurKind::UMin:
     case RecurKind::UMax:
+    case RecurKind::AnyOf:
+      return true;
+
+    // Change the order of floating-point operations may alter the results. If a
```
sjoerdmeijer: Nit: change -> changing

llvm/lib/Transforms/Scalar/LoopInterchange.cpp
```diff
@@ -986,6 +994,13 @@ static bool checkReductionKind(Loop *L, PHINode *PHI,
...
+    // certain instruction has ninf flag, it means that reordering can produce a
```
sjoerdmeijer: Nit, maybe "has ninf flag" -> "has the ninf flag"

sjoerdmeijer (approved):
LGTM

llvm/lib/Transforms/Scalar/LoopInterchange.cpp
kasuga-fj: Fixed, thanks.

llvm/lib/Transforms/Scalar/LoopInterchange.cpp
kasuga-fj: Thanks, updated.

PR #197924, opened 2026-05-15
 [Lit] Add a function to LLVMConfig to register a tool to be run in daemon mode

BStott6:
This PR is dependent on https://github.com/llvm/llvm-project/pull/197704. The changes in the first commit in this PR are from that PR and its dependencies. Please review from the 2nd commit onwards :)

This PR adds a function to `LLVMConfig` which correctly sets up the in-process built-in for running a tool in daemon mode. It also checks the tool executable exists and performs a small check to make sure the tool supports running in daemon mode. The key for the in-process built-in is set to be the absolute path to the tool binary, as Lit's substitutions always expand tools to their absolute path, so this will intercept all uses of the tool.

To test this, I have updated the `llvm-daemon-integration` test to use `daemon_config.use_daemon_tool` (the function called by `LLVMConfig.use_daemon_tool`) to set up the ExampleDaemon.

### DIFF
--- a/llvm/utils/lit/lit/llvm/daemon_config.py
+++ b/llvm/utils/lit/lit/llvm/daemon_config.py
@@ -5,7 +5,13 @@ import functools
 import lit.util
 from lit.InprocBuiltins import InprocBuiltin
 from lit.LitConfig import LitConfig
-from lit.llvm.daemon_tool import DaemonTool, invoke_llvm_daemon_tool
+from lit.llvm.daemon_tool import (
+    DaemonError,
+    DaemonExited,
+    DaemonTool,
+    UnexpectedDaemonOutput,
+    invoke_llvm_daemon_tool,
+)
 from lit.TestingConfig import TestingConfig
 
 
@@ -20,7 +26,7 @@ def check_daemon_support(executable: str) -> bool:
         daemon = DaemonTool(executable)
         daemon.start_daemon()
         daemon.send_commands(["exit"])
-    except:
+    except (DaemonError, DaemonExited, UnexpectedDaemonOutput):
         return False
 
     return True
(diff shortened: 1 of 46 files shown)


llvm/utils/lit/lit/llvm/daemon_config.py
```diff
@@ -0,0 +1,83 @@
...
+) -> bool:
+    """
+    Add the in-process built-in for running an LLVM tool in daemon mode.
+    The key for the in-process built-in is the absolute path to the tool,
+    so that it automatically replaces uses of the tool in testing, unless
+    `inproc_builtin_key_override` is specified, in which case it will
+    be the string provided via that parameter.
+    """
+
+    # Find the absolute path to the executable in the tool directory. executable = lit.util.which(daemon_tool, tool_dir)
```
OCHyams: Extra `executable = lit.util.which(daemon_tool, tool_dir)` in comment?

Possibly worth mentioning in the function's doc-string that the tool_name tool is looked up in this way? I'll leave that up to you.

OCHyams (approved):
lgtm with inline nit addressed

llvm/utils/lit/lit/llvm/daemon_config.py
```diff
@@ -0,0 +1,83 @@
...
+    Check that the executable at the given executable path supports being run
+    as an LLVM daemon tool, by testing that we receive the `ready` message
+    when starting the daemon.
+    """
+
+    try:
+        daemon = DaemonTool(executable)
+        daemon.start_daemon()
+        daemon.send_commands(["exit"])
+    except:
```
boomanaiden154: It would be nice if we could avoid a bare except here.

BStott6: I agree, I think we can change this to accept just DaemonError or UnexpectedDaemonOutput.

PR #197927, opened 2026-05-15
[clang-tidy] Add checker alias framework and restore hicpp aliases

earnol:
The change https://github.com/llvm/llvm-project/issues/183462 and subsequent changes removed all hicpp checkers with giving a little time to adapt. This change adds a generic check alias mechanism to clang-tidy that supports: registering aliases, central alias management table, bidirectional disable semantics, NOLINT suppression using alias names, canonical name preference, opt-in alias usage notification feature.

RFC for this PR: [90826](https://discourse.llvm.org/t/rfc-backward-compatible-alias-mechanism-for-removed-renamed-clang-tidy-checks/90826)

### DIFF
--- a/clang-tools-extra/clang-tidy/ClangTidyDiagnosticConsumer.cpp
+++ b/clang-tools-extra/clang-tidy/ClangTidyDiagnosticConsumer.cpp
@@ -19,6 +19,7 @@
 #include "ClangTidyOptions.h"
 #include "GlobList.h"
 #include "NoLintDirectiveHandler.h"
+#include "aliases/ClangTidyAliases.h"
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/ASTDiagnostic.h"
 #include "clang/AST/Attr.h"
@@ -27,6 +28,7 @@
 #include "clang/Basic/Diagnostic.h"
 #include "clang/Basic/DiagnosticOptions.h"
 #include "clang/Basic/FileManager.h"
+#include "clang/Basic/LLVM.h"
 #include "clang/Basic/SourceManager.h"
 #include "clang/Frontend/DiagnosticRenderer.h"
 #include "clang/Lex/Lexer.h"
@@ -34,9 +36,12 @@
 #include "clang/Tooling/Core/Replacement.h"
 #include "llvm/ADT/BitVector.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallString.h" // IWYU pragma: keep
 #include "llvm/ADT/StringMap.h"
 #include "llvm/Support/FormatVariadic.h"
 #include "llvm/Support/Regex.h"
+#include "llvm/Support/raw_ostream.h"
+#include <memory>
 #include <optional>
 #include <tuple>
 #include <utility>
@@ -216,8 +221,23 @@ bool ClangTidyContext::shouldSuppressDiagnostic(
     SmallVectorImpl<tooling::Diagnostic> &NoLintErrors, bool AllowIO,
     bool EnableNoLintBlocks) {
   const std::string CheckName = getCheckName(Info.getID());
-  return NoLintHandler.shouldSuppress(DiagLevel, Info, CheckName, NoLintErrors,
-                                      AllowIO, EnableNoLintBlocks);
+  if (NoLintHandler.shouldSuppress(DiagLevel, Info, CheckName, NoLintErrors,
+                                   AllowIO, EnableNoLintBlocks))
+    return true;
+  // Also check if any alias name for this check is suppressed.
+  for (const StringRef Alias :
+       ClangTidyAliases::getAliasesForCanonical(CheckName)) {
+    if (NoLintHandler.shouldSuppress(DiagLevel, Info, std::string(Alias),
+                                     NoLintErrors, AllowIO,
+                                     EnableNoLintBlocks)) {
+      if (NotifyAliases)
+        llvm::errs() << "note: '" << Alias
+                     << "' is an alias for canonical name '" << CheckName
+                     << "' [checker-alias]\n";
+      return true;
+    }
+  }
+  return false;
 }
 
 void ClangTidyContext::setSourceManager(SourceManager *SourceMgr) {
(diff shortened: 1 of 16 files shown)


clang-tools-extra/clang-tidy/aliases/ClangTidyAliases.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+namespace clang::tidy {
+
+/// Alias table. Kept sorted by alias name for readability.
+static constexpr std::pair<StringRef, StringRef> AliasTable[] = {
+    // Permanent aliases.
+    {"cert-dcl03-c", "misc-static-assert"},
+    {"cert-oop11-cpp", "performance-move-constructor-init"},
+    // Deprecated aliases: keeping for backward compatibility.
+    {"hicpp-avoid-c-arrays", "modernize-avoid-c-arrays"},
```
vbvictor: I think we don't want to keep mention of removed checks in upstream?

What is the benefit of this? We already listed all checks in release notes, the transition should be straightforward.

vbvictor (changes requested):

clang-tools-extra/clang-tidy/aliases/ClangTidyAliases.cpp
vbvictor: It should take a couple of minutes of `sed` to replace all hicpp checks listed [here](https://clang.llvm.org/extra/ReleaseNotes.html#potentially-breaking-changes) with existing counterparts

earnol: The whole idea is to have a convenient mechanism to preserve temporal aliases to give users a time for transition and make landing of potentially breaking changes softer.

earnol: So the current situation is we make breaking change and your problems are of no concern of us. This is acceptable stance, but not caring.
I propose to use slightly different stance: "We are going to make a potentially breaking changes and obsoleting this. Dear user, please take your time to migrate." 
It's much more caring and gentle to the user. The architecture of transition layer is local to one module and should be easy and convenient to maintain.

carlosgalvezp:
We have been instructed to remove the "hicpp" module as soon as possible for legal reasons, that's why it was done in an abrupt way instead of waiting 2 releases to do it. Therefore we should not re-enable it here IMO.

clang-tools-extra/clang-tidy/aliases/ClangTidyAliases.cpp
zwuis: The hicpp module was removed ahead of the normal deprecation cycle because of licensing concerns (https://github.com/llvm/llvm-project/issues/183462#issuecomment-4305159993). Could you confirm whether keeping those aliases would comply with the license?

clang-tools-extra/clang-tidy/ClangTidyDiagnosticConsumer.cpp
```diff
@@ -236,11 +256,86 @@ static bool parseFileExtensions(llvm::ArrayRef<std::string> AllFileExtensions,
   return true;
 }
 
+/// Expand check alias patterns in the checks string.
+/// For each glob pattern, if it matches an alias name, also add the canonical
+/// name (and vice versa) with the same +/- prefix. This ensures that disabling
+/// either name disables both.
+static std::string expandCheckAliases(StringRef Checks) {
+  if (Checks.empty())
+    return std::string(Checks);
```
EugeneZelenko:
```suggestion
    return {Checks};
```

unterumarmung:
For deprecating checks, I thought the following feature would be useful: a separate check that analyzes the YAML configuration and suggests fix-its for deprecated and/or removed names. If a deprecated name is not referenced directly, but through a pattern such as `modernize-*`, the check should add the new name to the list, unless it is already covered.

Later, this could be extended beyond checks to support other options in the YAML as well.

clang-tools-extra/clang-tidy/aliases/ClangTidyAliases.cpp
vbvictor: Normally, we just leave notice in ReleaseNotes about deprecation and don't remove/delete source code, e.g. in https://clang.llvm.org/extra/ReleaseNotes.html#potentially-breaking-changes there is note about `performance-faster-string-find` which will be removed in 2 releases.
So we don't have to add any new code/modules to support this.

vbvictor:
One major issue with this design is that we start to have 2 places for registering aliases: first is new alias module, second is default "old" way. 
With new "alias module" we have "thin" aliases that don't show their names in diagnostics and "heavy" (old) aliases that show their names in diagnostics.
To keep things organized and avoid further complications I think we need to choose one way of how we want to deal with aliases (new or old one) - each one of them have pros and cons, at least in "old" way we can default settings for checks.

carlosgalvezp:
If we want to introduce a new "alias framework" I think it needs an RFC first.

clang-tools-extra/clang-tidy/aliases/ClangTidyAliases.cpp
earnol: @zwuis and @vbvictor.  
The licensing concern in #183462 is about the HIC++ LICENSE.TXT attribution requirement for the module's documentation and content as i comprehend it. This PR does not restore **any** HIC++ copyrighted material: neither documentation, nor LICENSE.TXT, nor rule descriptions. The aliases are bare string redirects (hicpp-use-override ==> modernize-use-override) with no HIC++ content attached. It was explicitly designed this way! The check names are functional identifiers in user configurations, not copyrighted material from the HIC++ standard.
 
The original issue author also noted: "Currently we can keep the initial check as alias then completely remove them later." This PR implements exactly that — a thin compatibility layer with **no** HIC++ content.

clang-tools-extra/clang-tidy/ClangTidyDiagnosticConsumer.cpp
earnol: We want to return an empty string here. It's a guard to handle a border case when user request no checks at all :)
I'll change to ``return {};`` if you do not mind.

earnol:
> We have been instructed to remove the "hicpp" module as soon as possible for legal reasons, that's why it was done in an abrupt way instead of waiting 2 releases to do it. Therefore we should not re-enable it here IMO.

@carlosgalvezp  Please refer to my response above: https://github.com/llvm/llvm-project/pull/197927#discussion_r3258863733. None of the entities mentioned in  https://github.com/llvm/llvm-project/issues/183462 are restored.

clang-tools-extra/clang-tidy/ClangTidyDiagnosticConsumer.cpp
```diff
@@ -236,11 +256,86 @@ static bool parseFileExtensions(llvm::ArrayRef<std::string> AllFileExtensions,
...
+}
+
 void ClangTidyContext::setCurrentFile(StringRef File) {
   CurrentFile = std::string(File);
   CurrentOptions = getOptionsForFile(CurrentFile);
-  CheckFilter = std::make_unique<CachedGlobList>(
-      StringRef(getOptions().Checks.value_or("")));
+  ExpandedChecks = expandCheckAliases(getOptions().Checks.value_or(""));
+  CheckFilter = std::make_unique<CachedGlobList>(StringRef(ExpandedChecks));
   WarningAsErrorFilter = std::make_unique<CachedGlobList>(
```
zeyi2: IMO `WarningsAsErrors` may need the same alias handling as `Checks`.

Currently `Checks` is expanded through `expandCheckAliases()`. In a legacy config such as:

```text
--checks='-*,hicpp-use-override' --warnings-as-errors='hicpp-*'
```

`Checks` will enable the canonical checks `modernize-use-override`. But`WarningAsErrorFilter` is built from the original `WarningsAsErrors` string, so `Context.treatAsError("modernize-use-override")` will not match `hicpp-use-override`.

This means users who previously made all `hicpp-*` diagnostics fatal would silently get warnings instead of errors after this compatibility layer. Can we fix the problem?

earnol: Thank you for the great comment. This is the bug with the PR. Fixed! Lit test also updated.

earnol:
> One major issue with this design is that we start to have 2 places for registering aliases: first is new alias module, second is default "old" way. With new "alias module" we have "thin" aliases that don't show their names in diagnostics and "heavy" (old) aliases that show their names in diagnostics. To keep things organized and avoid further complications I think we need to choose one way of how we want to deal with aliases (new or old one) - each one of them have pros and cons, at least in "old" way we can default settings for checks.

@vbvictor 
The alias table is the single mechanism for _thin_ aliases and it redirects where:
1. The alias name never appears in diagnostics (only the canonical name does)
2. --notify-aliases emits a note informing users about the canonical name
3. The intent is alias handling simplification: one table, one lookup, no per-alias code

The old ``registerCheck`` approach remains for _heavy_ aliases where:
1. The alias name **is** the diagnostic identity (appears in [check-name] and should be here permanently)
2. The alias may have its own default options, however, it can be also moved to the table.

Also, the alias table uses registerCheck under the hood (via registerCheckAlias and resolveAliases which copies the factory). The table is just a higher-level interface on top of the same machinery.
 
These are fundamentally different use cases. The table handles the first cleanly. The coexistence is intentional, not accidental duplication. I see it as a classical choice: simplicity vs. control.

earnol:
> For deprecating checks, I thought the following feature would be useful: a separate check that analyzes the YAML configuration and suggests fix-its for deprecated and/or removed names. If a deprecated name is not referenced directly, but through a pattern such as `modernize-*`, the check should add the new name to the list, unless it is already covered.
> 
> Later, this could be extended beyond checks to support other options in the YAML as well.

@unterumarmung 
A config migration tool is a great complementary idea as it helps users update their .clang-tidy files proactively. But it doesn't solve the immediate problem: existing CI pipelines break the moment aliases disappear. A runtime compatibility layer prevents breakage while users migrate at their own pace. Both approaches can coexist and the migration tool would be a natural follow-up as a separate PR.

earnol:
> If we want to introduce a new "alias framework" I think it needs an RFC first.

@carlosgalvezp 
Per the LLVM RFC Process (https://llvm.org/docs/RFCProcess.html): "Substantive changes to LLVM projects need to be acceptable to the wider community." And the Developer Policy (https://llvm.org/docs/DeveloperPolicy.html#proposing-major-changes-rfcs) states RFCs are for when you "plan to make a major change to the way LLVM works or want to add a major new extension."
 However, this PR is approximately 500 lines, localized to single module barring the integration, adds a single opt-in flag (--notify-aliases), and one user-visible behavior change: aliases migrated to the table (cert-dcl03-c, cert-oop11-cpp) now report diagnostics under their canonical name instead of the alias name. This is an intentional change: the alias is a redirect, not an identity.
 While i don't believe this meets the threshold for an RFC, i am happy to write one if the community feels otherwise.

earnol:
I have added the RFC to this PR: https://discourse.llvm.org/t/rfc-backward-compatible-alias-mechanism-for-removed-renamed-clang-tidy-checks/90826

earnol:
Bump. Keeping it updated while RFC is being discussed.

earnol:
And bump again to keep it open.

earnol:
And bump again, while RFC being dicussed

PR #197928, opened 2026-05-15
[clang][CodeGen] Fix crash in CGDebugInfo on incomplete record type (#156352)

VitorRamos:
Fix a crash when generating debug info for a template class that is
still being defined when its forward declaration is needed.

getTypeIdentifier() called getVTableLinkage() on a record that was only
"being defined" (not yet complete), which triggered premature
ASTRecordLayout computation and cached a layout with zero fields. Later,
CollectRecordFields used the stale cached layout and crashed on an
out-of-bounds field offset access.

Replace getDefinition() with isCompleteDefinition() to correctly skip
records whose definition is not yet complete.

Fixes #156352

### DIFF
--- a/clang/lib/CodeGen/CGDebugInfo.cpp
+++ b/clang/lib/CodeGen/CGDebugInfo.cpp
@@ -1290,7 +1290,7 @@ static SmallString<256> getTypeIdentifier(const TagType *Ty, CodeGenModule &CGM,
   if (!needsTypeIdentifier(TD, CGM, TheCU))
     return Identifier;
   if (const auto *RD = dyn_cast<CXXRecordDecl>(TD))
-    if (RD->getDefinition())
+    if (RD->isCompleteDefinition())
       if (RD->isDynamicClass() &&
           CGM.getVTableLinkage(RD) == llvm::GlobalValue::ExternalLinkage)
         return Identifier;
--- a/clang/test/CodeGenCXX/debug-info-template-incomplete-record.cpp
+++ b/clang/test/CodeGenCXX/debug-info-template-incomplete-record.cpp
@@ -1,83 +1,32 @@
-// RUN: %clang_cc1 -std=c++20 -debug-info-kind=standalone -fvisibility=hidden \
-// RUN:   -O2 -triple x86_64-linux-gnu -emit-llvm %s -o - | FileCheck %s
+// RUN: %clang_cc1 -std=c++17 -debug-info-kind=standalone -O2 \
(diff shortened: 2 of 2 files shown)


Michael137 (commented):
Explanation/fix sounds plausible, but can the test-case be reduced even further?

Looks like david attached a smaller reproducer to the issue you linked. Does that one still crash?

clang/lib/CodeGen/CGDebugInfo.cpp
Michael137, line 1293: > triggered premature ASTRecordLayout

Could you describe where in `getVTableLinkage` that happens? Perhaps we need similar treatment at other callsites.

A shorter reproducer would help understand what's happening

VitorRamos:
> Explanation/fix sounds plausible, but can the test-case be reduced even further?
> 
> Looks like david attached a smaller reproducer to the issue you linked. Does that one still crash?

The reduced version only triggers the assert, i wasn't sure which one would be the best, ill update the test to use that one instead.

clang/lib/CodeGen/CGDebugInfo.cpp
VitorRamos: From the assert:

```
#11 ASTContext::getASTRecordLayout                       ← assertion fires
#12 FinalOverriders::FinalOverriders
#13 ItaniumVTableBuilder::ItaniumVTableBuilder
#14 ItaniumVTableContext::computeVTableRelatedInformation
#15 ItaniumVTableContext::getVTableLayout
#16 ItaniumCXXABI::canSpeculativelyEmitVTableAsBaseClass
#17 ItaniumCXXABI::canSpeculativelyEmitVTable
#18 CodeGenModule::getVTableLinkage
#19 getTypeIdentifier                                    ← caller in CGDebugInfo
#20 CGDebugInfo::getOrCreateRecordFwdDecl
```

So `getVTableLinkage` ends up triggering vtable layout (via `canSpeculativelyEmitVTable` → `FinalOverriders`), and that requires a complete record layout.

Other callers of `getVTableLinkage` for reference:

- `CGDebugInfo::getTypeIdentifier` - fixed by this PR.
- `CGDebugInfo::completeClassData`
- ItaniumCXXABI.cpp (4 sites)
- MicrosoftCXXABI.cpp (2 sites)

I don't know the codebase well enough to say whether the other callers can ever reach `getVTableLinkage` with an incomplete class.

Maybe adding an early `assert(RD->isCompleteDefinition())` at the top of `getVTableLinkage` itself could help ?

VitorRamos:
@Michael137  can you take another look?

PR #197929, opened 2026-05-15
[ExtractAPI] Add support for anyAppleOS

snprajwal:
6c8940ccad31457aadb48c6f2bce139b1c5dda01 introduced support for anyAppleOS, which can be used to indicate availability for all Apple platforms in a single availability attribute.
3089120f0cd2fe78911d1bc9703764a07fa7eb32 updated it to preserve the original attribute info and store the inferred platform-specific attribute in a separate field which, if present, is returned by `getEffectiveAttr()`.

With these changes, ExtractAPI now reports these availabilities verbatim for an "anyAppleOS" platform as the availability info for Decls directly uses the availability attribute without fetching the effective attribute. The expected output would be the target platform for which the symbol graph is being built. This patch updates the availability collection logic to use `getEffectiveAttr()` and include the target platform instead of "anyAppleOS" for symbol availabilities in symbol graphs.

rdar://174558184


QuietMisdreavus (approved):
This looks great, thanks!

PR #93873, opened 2024-05-30
[Clang][Sema] Diagnose variable template explicit specializations with storage-class-specifiers

sdkrystian:
According to [[temp.expl.spec] p2](http://eel.is/c++draft/temp.expl.spec#2):
> The declaration in an _explicit-specialization_ shall not be an _export-declaration_. An explicit specialization shall not use a _storage-class-specifier_ other than `thread_local`.

Clang partially implements this, but a number of issues exist:
1. We don't diagnose class scope explicit specializations of variable templates with _storage-class-specifiers_, e.g.
    ```cpp
    struct A
    {
        template<typename T>
        static constexpr int x = 0;

        template<>
        static constexpr int x<void> = 1; // ill-formed, but clang accepts
    };
    ````
2. We incorrectly reject class scope explicit specializations of variable templates when `static` is not used, e.g.
    ```cpp
    struct A
    {
        template<typename T>
        static constexpr int x = 0;

        template<>
        constexpr int x<void> = 1; // error: non-static data member cannot be constexpr; did you intend to make it static?
    };
    ````
3. We don't diagnose dependent class scope explicit specializations of function templates with storage class specifiers, e.g.
    ```cpp
    template<typename T>
    struct A
    {
        template<typename U>
        static void f();

        template<>
        static void f<int>(); // ill-formed, but clang accepts
    };
    ````

This patch addresses these issues as follows:
- [#1]() is fixed by issuing a diagnostic when an explicit specialization of a variable template has storage class specifier
- [#2]() is fixed by considering any non-function declaration with any template parameter lists at class scope to be a static data member. This also allows for better error recovery (it's more likely the user intended to declare a variable template than a "field template").
- [#3]() is fixed by checking whether a function template explicit specialization has a storage class specifier even when the primary template is not yet known. 

One thing to note is that it would be far simpler to diagnose this when parsing the _decl-specifier-seq_, but such an implementation would necessitate a refactor of `ParsedTemplateInfo` which I believe to be outside the scope of this patch.

PR #197930, opened 2026-05-15
[clang][nfc] Guard thread_local in clang/test/CXX/drs/cwg7xx.cpp

zmodem:
On targets which don't support `thread_local`, such as `arm64-apple-darwin-unknown`, the test would fail with a bunch of `thread-local storage is not supported for the current target` messages. Follow-up to #197732

### DIFF
--- a/clang/test/CXX/drs/cwg7xx.cpp
+++ b/clang/test/CXX/drs/cwg7xx.cpp
@@ -100,7 +100,7 @@ static_assert(!is_volatile<void()volatile&>::value, "");
 } // namespace cwg713
 
 namespace cwg717 { // cwg717: 3.3
-#if __cplusplus >= 201103L
+#if __cplusplus >= 201103L && __has_feature(cxx_thread_local)
 void f() {
   thread_local extern int i;
   thread_local extern int& j;


Endilll:
We use `thread_local` in a number of tests in `tests/CXX` without checking for `__has_feature` or specifying target triple. How do they work on `arm64-apple-darwin-unknown`?

zmodem:
I don't know, but this is the only one that fails for us. Maybe it's because this test does specify a triple (not sure why actually), using `%itanium_abi_triple`.

Endilll:
> I don't know, but this is the only one that fails for us. Maybe it's because this test does specify a triple (not sure why actually), using `%itanium_abi_triple`.

If you can reproduce this, can you try removing the triple? I wonder if it's really needed there.

Endilll:
Also, I see that `llvm-clang-aarch64-darwin` buildbot is green (https://lab.llvm.org/buildbot/#/builders/190). What am I missing?

AaronBallman:
How is this test passing currently for Darwin? https://github.com/llvm/llvm-project/blob/main/clang/test/CodeGenCXX/threadlocal_address.cpp

How is `thread_local` not supported on such a modern platform? This is a requirement for conformance for C and C++ so are we saying ARM64 on Darwin doesn't comform? That would be a surprise.

zmodem:
Repro steps on (arm64) mac:

```
$ cmake -Bbuild.tmp -GNinja -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD=Native -DLLVM_DEFAULT_TARGET_TRIPLE=arm64-apple-darwin llvm/
$ ninja -C build.tmp check-clang
...
FAIL: Clang :: CXX/drs/cwg7xx.cpp (2644 of 24579)
******************** TEST 'Clang :: CXX/drs/cwg7xx.cpp' FAILED ********************
Exit Code: 1

Command Output (stdout):
--
# RUN: at line 1
/Users/hwennborg/llvm-project/build.tmp/bin/clang -cc1 -internal-isystem /Users/hwennborg/llvm-project/build.tmp/lib/clang/23/include -nostdsysteminc -triple arm64-apple-darwin-unknown -std=c++98 /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp -fexceptions -fcxx-exceptions -pedantic-errors -verify-directives -verify=expected,cxx98,cxx98-14,cxx98-11
# executed command: /Users/hwennborg/llvm-project/build.tmp/bin/clang -cc1 -internal-isystem /Users/hwennborg/llvm-project/build.tmp/lib/clang/23/include -nostdsysteminc -triple arm64-apple-darwin-unknown -std=c++98 /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp -fexceptions -fcxx-exceptions -pedantic-errors -verify-directives -verify=expected,cxx98,cxx98-14,cxx98-11
# RUN: at line 2
/Users/hwennborg/llvm-project/build.tmp/bin/clang -cc1 -internal-isystem /Users/hwennborg/llvm-project/build.tmp/lib/clang/23/include -nostdsysteminc -triple arm64-apple-darwin-unknown -std=c++11 /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp -fexceptions -fcxx-exceptions -pedantic-errors -verify-directives -verify=expected,cxx98-14,cxx98-11,since-cxx11
# executed command: /Users/hwennborg/llvm-project/build.tmp/bin/clang -cc1 -internal-isystem /Users/hwennborg/llvm-project/build.tmp/lib/clang/23/include -nostdsysteminc -triple arm64-apple-darwin-unknown -std=c++11 /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp -fexceptions -fcxx-exceptions -pedantic-errors -verify-directives -verify=expected,cxx98-14,cxx98-11,since-cxx11
# | error: diagnostics with 'error' severity seen but not expected: 
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 105: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 106: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 109: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 110: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 116: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 117: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 120: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 121: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 126: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 127: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 132: thread-local storage is not supported for the current target
# |   File /Users/hwennborg/llvm-project/clang/test/CXX/drs/cwg7xx.cpp Line 133: thread-local storage is not supported for the current target
# | 12 errors generated.

--

********************
********************
Failed Tests (1):
  Clang :: CXX/drs/cwg7xx.cpp
```

---

AaronBallman:
Btw, this discussion raises an interesting question: what is the value of running conformance tests on targets which are non-conforming? Should the `C` and `CXX` directories not be run for some targets at all?

zmodem:
> If you can reproduce this, can you try removing the triple? I wonder if it's really needed there.

Yes, removing the triple makes it pass.

Unfortunately it seems the triple is needed on Windows. If I remove it, I get failures like:

```
# | error: diagnostics with 'error' severity expected but not seen:
# |   File C:\src\llvm-project\clang\test\CXX\drs\cwg7xx.cpp Line 307 (directive at C:\src\llvm-project\clang\test\CXX\drs\cwg7xx.cpp:308) 'cxx14-error': static data member 'v' already has an initializer
# |   File C:\src\llvm-project\clang\test\CXX\drs\cwg7xx.cpp Line 321 (directive at C:\src\llvm-project\clang\test\CXX\drs\cwg7xx.cpp:322) 'cxx14-error': static assertion failed due to requirement 'cwg727::mixed_inner_outer_specialization::B<0>().v<0> == 4':
# | error: diagnostics with 'note' severity expected but not seen:
# |   File C:\src\llvm-project\clang\test\CXX\drs\cwg7xx.cpp Line 293 (directive at C:\src\llvm-project\clang\test\CXX\drs\cwg7xx.cpp:309) 'cxx14-note': previous initialization is here
# |   File C:\src\llvm-project\clang\test\CXX\drs\cwg7xx.cpp Line 321 (directive at C:\src\llvm-project\clang\test\CXX\drs\cwg7xx.cpp:323) 'cxx14-note': expression evaluates to '2 == 4'
# | 4 errors generated.
```

> Also, I see that llvm-clang-aarch64-darwin buildbot is green (https://lab.llvm.org/buildbot/#/builders/190). What am I missing?

The difference seems to be us configuring the build with `-DLLVM_DEFAULT_TARGET_TRIPLE=arm64-apple-darwin`

> How is this test passing currently for Darwin? https://github.com/llvm/llvm-project/blob/main/clang/test/CodeGenCXX/threadlocal_address.cpp

I think it comes down to what triple it's passing.

> How is thread_local not supported on such a modern platform? This is a requirement for conformance for C and C++ so are we saying ARM64 on Darwin doesn't comform? That would be a surprise.

IIUC, thread_local was not supported on earlier macos versions. 

It seems that when explicitly passing a triple like `arm64-apple-darwin-unknown`, clang assumes ancient macos without that support.

This doesn't seem to come up a lot -- I think clang normally targets a recent macos version unless the user specifies something.

What I think makes this test special is its use of `%itanium_abi_triple`, which in builds configured with `-DLLVM_DEFAULT_TARGET_TRIPLE=arm64-apple-darwin` like ours, seems to become `arm64-apple-darwin-unknown`.

(We hit the same failure on x86_64 also.)

zmodem:
I suppose there are different ways we could fix this.

1. I think my current fix is fairly reasonable :-) If we want the test to work with any triple, including those that don't support `thread_local`, guarding it with `__has_feature` makes sense
2. If being triple agnostic isn't important for this test, we could put an explicit triple on it and call it a day.
3. Maybe we can fix `%itanium_abi_triple` somehow, or clang's interpretation of `arm64-apple-darwin-unknown` (this starts feeling like bigger changes).

localspook:
The `-target %itanium_abi_triple` flags were added in this commit: abcc2a8 and seem to be a workaround for nonconforming MS extensions (which explains why the test fails on Windows when you remove them). So maybe we could consider passing  `-fno-ms-extensions`? The `__has_feature` guard works, but I don't think it's the optimal solution unless we genuinely want to commit to supporting targets without `thread_local`.

Edit: Actually nevermind, it looks like this behaviour really is controlled by `-target`, not by `-fms-extensions`.

zmodem:
Can we revert https://github.com/llvm/llvm-project/pull/197732 to unbreak the build while we discuss this? Or can we land my fix for now?

localspook:
Yes, I'm happy to revert for now: #198074

Endilll:
> Unfortunately it seems the triple is needed on Windows.

I looked into this. The failing tests came from #93873. I came up with the following reduction of the CWG727 test:
```cpp
template<int> struct B {
  template<int> static const int u = 1;
  template<> const int u<0> = 2; // #cwg727-u0         // rejected by GCC
};

template<> template<int> constexpr int B<0>::u    = 3; // rejected by GCC
template<> template<>    constexpr int B<0>::u<0> = 4; // rejected by Clang, accepted by MSVC, EDG, and Clang with x86_64-pc-windows-msvc triple
// since-cxx14-error@-1 {{static data member 'u' already has an initializer}}
//   since-cxx14-note@#cwg727-u0 {{previous initialization is here}}

static_assert(B<0>().u<0> == 2, ""); // passes in Clang and in Clang with x86_64-pc-windows-msvc triple
static_assert(B<0>().u<0> == 4, ""); // passes in GCC, EDG, and MSVC
```
(I can't replicate the behavior of Clang with MSVC triple on Compiler Explorer, but I'm seeing this locally on my Linux machine.)

So it appears that itanium ABI triple disables MSVC compatibility mode, making this test work on all targets.

I believe Clang behavior is incorrect, and I'll file a dedicated issue for that, because it's a rabbit's hole. Separately, this issue with `arm64-apple-darwin-unknown` showcases that itanium ABI triple does more that just disabling MSVC compatibility mode. Are there other ways to disable compatibility mode without negative side effects? Apparently, `-fno-ms-compatibility` is not a thing.

Endilll:
> I believe Clang behavior is incorrect, and I'll file a dedicated issue for that, because it's a rabbit's hole.

#198089

PR #193254, opened 2026-04-21
[LLVM][ConstantFolding] Use correct type when flushing denormals.

paulwalker-arm:
flushDenormalConstant always created a scalar ConstantFP even when passed a vector type. I've had to change the return type from ConstantFP* to Constant* because ConstantFP::get does not necessarily return a ConstantFP. This can be reverted when ConstantFP vectors are enabled by default (although additional work is required to support zero).

PR #197931, opened 2026-05-15
[LLVM] Refactor ConstantFP member functions to rely on its native vector support.

paulwalker-arm:
ConstantFP has native vector support so many of its member functions can be simplified by removing the ConstantVector indirection and can return a ConstantFP* rather than a generic Constant*.

NOTE: The changes to ConstantFolding.cpp are reverting part of https://github.com/llvm/llvm-project/pull/193254 that was necessary at the time.

### DIFF
--- a/llvm/lib/IR/Constants.cpp
+++ b/llvm/lib/IR/Constants.cpp
@@ -1130,91 +1130,70 @@ void ConstantByte::destroyConstantImpl() {
 //                                ConstantFP
 //===----------------------------------------------------------------------===//
 
-Constant *ConstantFP::get(Type *Ty, double V) {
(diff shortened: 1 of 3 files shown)


llvm/lib/IR/Constants.cpp
```diff
@@ -1250,16 +1229,6 @@ ConstantFP *ConstantFP::get(LLVMContext &Context, ElementCount EC,
   return Slot.get();
 }
 
-Constant *ConstantFP::getInfinity(Type *Ty, bool Negative) {
```
paulwalker-arm, line 1253: I've moved this so it's next to the other similar construct-a-specific-value methods.

nikic (approved):
LGTM

Issue #196798, opened 2026-05-10
[analyzer] Allow ownership_returns to opt out of memory leak warnings


anarazel:
Hi,

I'm working on using ownership_returns etc annotations for postgres. Unfortunately, for some of the allocators, postgres uses an arena style allocator (which are called memory contexts in postgres), where individual memory often is intentionally not freed, as it's considerably cheaper to throw away the entire context, rather than each individual allocation.  

We also at times are using plain allocators, where we do *not* want to ignore memory leak warnings.

Therefore I'd like a way to tell the analyzer that for a certain allocator functions we would like to opt out of leak warnings.  

I've tried some workarounds, like replacing all the allocators with an inline function that pass the returned pointer to an external function - which suffices to silence the leak warning - but that doesn't really scale to the number of functions that should signal that they return newly allocated memory.

We still would like to have use-after-free warnings etc.

Greetings,

Andres

NagyDonat:
I think the intended behavior could be achieved if you don't place an `ownership_returns` attribute on the allocation function (but annotate the deallocator with `ownership_takes`).

That way the value returned by the allocator would be handled as an unknown pointer (assuming that the definition of the allocator function is unavailable or complex) which is equivalent to what you expect:
- memory leak is not reported, because the pointer/region is not in an "allocated" state
- calling the deallocator once is OK – the analyzer doesn't report situations when a pointer of unknown origin is passed to the deallocator (because it doesn't see the full program, doesn't know the origin of every pointer, and has to optimistically assume that an unknown pointer _was_ allocated)
- use-after-free (and its special case, double-free) is reported because calling the deallocator places the pointer in a "released" state where it is invalid to use the pointer (even if the analyzer didn't see an allocator call).

The only limitation is that this way you cannot specify the allocation size. (The second argument of  `ownership_returns` would specify the 1-based index of the parameter which will specify the size of the allocation.) However, this is not an _ownership_ question and I think it would be good to introduce a generic `returned_region_size` attribute to allow specifying the dynamic size of a memory region without saying anything about ownership. (Exact schematics TBD, I expect lots of bikeshedding if we decide to implement this.)

@anarazel Would this "don't annotate the allocation function" idea work for you?

anarazel:
> I think the intended behavior could be achieved if you don't place an ownership_returns attribute on the allocation function (but annotate the deallocator with ownership_takes).
...
> @anarazel Would this "don't annotate the allocation function" idea work for you?

[...]

PR #197933, opened 2026-05-15
[NFC] Clarify behavior of ownership_holds in docs

NagyDonat:
I misunderstood the behavior of the attribute `ownership_holds` based on the claim that "using held memory is assumed to be legitimate".

To avoid similar misunderstandings in the future, I'm adding an extra sentence to the documentation.

(This question came up during the discussion of #196798)

### DIFF
--- a/clang/docs/analyzer/user-docs/Annotations.rst
+++ b/clang/docs/analyzer/user-docs/Annotations.rst
@@ -207,7 +207,7 @@ Use this attribute to mark functions that take ownership of memory and will deal
 
   void __attribute((ownership_holds(malloc, 2))) store_in_table(int key, record_t *val);
 
-The annotations ``ownership_takes`` and ``ownership_holds`` both prevent memory leak reports (concerning the specified parameter); the difference between them is that using taken memory is a use-after-free error, while using held memory is assumed to be legitimate. (However, releasing the held memory or passing it to another holding call is reported by the analyzer as an "attempt to release non-owned memory".)
+The annotations ``ownership_takes`` and ``ownership_holds`` both prevent memory leak reports (concerning the specified parameter); the difference between them is that using taken memory is a use-after-free error, while using held memory is assumed to be legitimate. However, releasing the held memory or passing it to another holding call is reported by the analyzer as an "attempt to release non-owned memory".
 
 
 Mac OS X API Annotations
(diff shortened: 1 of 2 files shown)


NagyDonat:
Disclaimer: I didn't test the sphinx build and its output, but my commit only adds plaintext, so it should be fine.

clang/docs/analyzer/user-docs/Annotations.rst
```diff
@@ -207,7 +207,8 @@ Use this attribute to mark functions that take ownership of memory and will deal
 
   void __attribute((ownership_holds(malloc, 2))) store_in_table(int key, record_t *val);
 
-The annotations ``ownership_takes`` and ``ownership_holds`` both prevent memory leak reports (concerning the specified parameter); the difference between them is that using taken memory is a use-after-free error, while using held memory is assumed to be legitimate.
+The annotations ``ownership_takes`` and ``ownership_holds`` both prevent memory leak reports (concerning the specified parameter); the difference between them is that using taken memory is a use-after-free error, while using held memory is assumed to be legitimate. (However, releasing the held memory or passing it to another holding call is reported by the analyzer as an "attempt to release non-owned memory".)
```
Szelethus:
```suggestion
The annotations ``ownership_takes`` and ``ownership_holds`` both prevent memory leak reports (concerning the specified parameter); the difference between them is that using taken memory is a use-after-free error, while using held memory is assumed to be legitimate. However, releasing the held memory or passing it to another holding call is reported by the analyzer as an "attempt to release non-owned memory".
```

clang/include/clang/Basic/AttrDocs.td
```diff
@@ -1663,7 +1663,9 @@ safely deallocated with the standard ``free()``.
 The annotations ``ownership_takes`` and ``ownership_holds`` both prevent memory
 leak reports (concerning the specified parameter); the difference between them
 is that using taken memory is a use-after-free error, while using held memory
-is assumed to be legitimate.
+is assumed to be legitimate. (However, releasing the held memory or passing it
+to another holding call is reported by the analyzer as an "attempt to release
+non-owned memory".)
```
Szelethus:
```suggestion
is assumed to be legitimate. However, releasing the held memory or passing it
to another holding call is reported by the analyzer as an "attempt to release
non-owned memory".
```

Szelethus (approved):
I suggest to remove the parantheses. Otherwise, LGTM

Szelethus (approved):
LGTM

PR #191377, opened 2026-04-10
[VPlan] Use llvm.masked.{u,s}{div,rem} for predicated division

lukel97:
Fixes #129538

To allow speculatively executing lanes that are predicated off in the original scalar loop, we currently widen divisions by using a "safe divisor" when we can't prove it's safe otherwise.

We previously tried to optimize these to VP intrinsics which can be speculated with a mask in #154076, but this was fragile as it relied on assuming that the safe-divisor lanes wouldn't be read. 

#189705 and #191240 added new intrinsics to explicitly represent semantics where the masked off lanes are poison. This PR teaches the loop vectorizer to use them.

We can then safely convert these to VP intrinsics in optimizeMasksToEVL for RISC-V. And on AArch64 we see better codegen after https://github.com/llvm/llvm-project/pull/191164.

The optimizeMasksToEVL transform is included in this PR, otherwise we end up with regressions in RISC-V from not removing the header mask. 

With this PR and the following C loop:

```c
void f(int *x, int y, int n) {
  for (int i = 0; i < n; i++)
    x[i] /= y;
}
```

The resulting RISC-V codegen goes from:

```asm
# %bb.1:                                # %vector.ph
	li	a3, 0
	vsetvli	a4, zero, e32, m2, ta, ma
	vmv.v.x	v8, a1
.LBB0_2:                                # %vector.body
                                        # =>This Inner Loop Header: Depth=1
	vsetvli	a1, a2, e32, m2, ta, ma
	sh2add	a4, a3, a0
	vle32.v	v10, (a4)
	vmv.v.i	v12, 1
	vsetvli	zero, zero, e32, m2, tu, ma
	vmv.v.v	v12, v8
	vsetvli	zero, zero, e32, m2, ta, ma
	vdiv.vv	v10, v10, v12
	sub	a2, a2, a1
	vse32.v	v10, (a4)
	add	a3, a3, a1
	bnez	a2, .LBB0_2

```

To:

```asm
# %bb.1:                                # %vector.ph
	li	a3, 0
.LBB0_2:                                # %vector.body
                                        # =>This Inner Loop Header: Depth=1
	vsetvli	a4, a2, e32, m2, ta, ma
	sh2add	a5, a3, a0
	vle32.v	v8, (a5)
	vdiv.vx	v8, v8, a1
	sub	a2, a2, a4
	vse32.v	v8, (a5)
	add	a3, a3, a4
	bnez	a2, .LBB0_2
```

PR #197937, opened 2026-05-15
[RISCV] Fix mask policy in convertSameMaskVMergeToVMv

lukel97:
In convertSameMaskVMergeToVMv, we try to change a vmerge to a vmv.v.v if the true operand is masked in such a way it guarantees the masked off elements are the same.

However vmerge.vvm and vmv.v.v are always mask undisturbed. If we want to reuse the masking in the true operand we need to also change its policy to mask undisturbed.

This fixes a miscompile that can be seen with llvm.masked.sdiv instructions after #191377 under certain conditions, detected by https://lab.llvm.org/buildbot/#/builders/214/builds/2795. The llvm.masked.sdiv instructions produced a mask agnostic PseudoVDIV that had a PseudoVMERGE_VVM merged into them without updating the policy.


topperc (approved):
LGTM

PR #195792, opened 2026-05-05
[libc] prefer *at syscalls in sys/stat wrappers

udaykiriti:
- Flips the #ifdef order to prefer the *at syscalls over normal ones.
  - from #195620

Issue #195620, opened 2026-05-04
[libc] Check the ordering of the *at syscalls


kaladron:
In our syscalls in LLVM libc, we often have clauses like:

```c++
#ifdef SYS_mkdir
#elif SYS_mkdirat
#endif
```

I want to check if we should be preferring the *at syscalls on systems that have both as I think they're newer and preferred.

udaykiriti:
can i work on this.

kaladron:
Please!


Jeff Bailey
Circus Ringmaster
***@***.***

udaykiriti:
Flipping the #ifdef order to prefer the *at syscalls (like mkdirat) looks the same. Cuz mkdir() and mkdirat(AT_FDCWD, ...) do the same isn’t it.
Edit: i think swapping the order makes will work and most of the archs support *.at

udaykiriti:
@kaladron please review the pr.

SHAURYASAXENA2024:
Hi! I'd like to work on this issue if it's still available.

udaykiriti:
@SHAURYASAXENA2024 just make a `pr`,  go through the codes and if u found any things that related to this issue work on them, i think some of them had done if u have any more ideas, implement it.

SHAURYASAXENA2024:
I really appreciate the suggestion. I will look at the code that's relevant to this issue and see if there are any other ways to make it better or if there are any other things that we can do to make it work more efficiently. If I find something that's useful and that we have not done already I will make a pull request, for the code changes.

Mubaris2:
Hi, I noticed that there were PR's merged for this issue, is there something else which needs to be worked on?

PR #197940, opened 2026-05-15
[libc] prefer *at syscalls in sys/stat wrappers

udaykiriti:
- These changes flips the #ifdef order to prefer the *at syscalls over normal ones.
  - In modern architectures, *at system calls are preferred over normal system calls cuz of safety issues.
  - So by checking for ""*at"" system calls first, we ensure better compatibility with modern systems.
  - After then normal syscalls moved else or elif for support to older ones.
  - From merged pr(#195792) and issue(#195620)


udaykiriti:
???

kaladron:
> ???

Please do keep in mind that there's a weekend in there. =)  There was a large queue of patches to review yesterday and I just didn't get to this one.

kaladron (approved):

kaladron:
Thank you for doing these!

PR #197942, opened 2026-05-15
[lldb][gdb-remote] Send executable path with native separators in vRun

charles-zablit:
When running the tests with `LLDB_USE_LLDB_SERVER=1`, the lldb-server gdb-remote vRun packet sends the executable path in FileSpec's normalised form (forward slashes). On Windows the inferior's `argv[0]` therefore uses '/', while the program path in the setup uses '\\'. CreateProcessW accepts either spelling, so the inferior runs fine but the strict-equality assertion fails.

This will not work for UNC paths (`\\server\foo.exe` cannot be converted to `//server/foo.exe`).

This patch uses `GetPath(/*denormalize=*/true)` so that the path uses native separators. This is a no-op with POSIX style paths.

This fixes 3 lldb-dap tests on Windows when running with `LLDB_USE_LLDB_SERVER=1`.

### DIFF
--- a/lldb/source/Plugins/Process/gdb-remote/ProcessGDBRemote.cpp
+++ b/lldb/source/Plugins/Process/gdb-remote/ProcessGDBRemote.cpp
@@ -828,7 +828,7 @@ Status ProcessGDBRemote::DoLaunch(lldb_private::Module *exe_module,
       // make sure to use the actual executable path found in the launch_info...
       Args args = launch_info.GetArguments();
       if (FileSpec exe_file = launch_info.GetExecutableFile())
-        args.ReplaceArgumentAtIndex(0, exe_file.GetPath(false));
+        args.ReplaceArgumentAtIndex(0, exe_file.GetPath(/*denormalize=*/true));
       if (llvm::Error err = m_gdb_comm.LaunchProcess(args)) {
         error = Status::FromErrorStringWithFormatv(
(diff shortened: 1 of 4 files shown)


Nerixyz:
> CreateProcessW accepts either spelling, so the inferior runs fine but the strict-equality assertion fails.

This isn't true for UNC paths that start with `\\` though, right?

charles-zablit:
> > CreateProcessW accepts either spelling, so the inferior runs fine but the strict-equality assertion fails.
> 
> This isn't true for UNC paths that start with `\\` though, right?

It's not true no, something like `\\server\foo.exe` cannot be converted to `//server/foo.exe`.
This patch assumes that the tests run on a local drive. I updated the description.

Nerixyz:
> the lldb-server gdb-remote vRun packet sends the executable path in FileSpec's normalised form (forward slashes)

But when lldb-server "normalizes" a UNC path, would it send `//server/foo/bar` or `\\server\foo\bar`? So would lldb-server need to be fixed here?

charles-zablit:
Ah I see what you mean. Yes it would break. We need to fix this where the Path is normalized. I will update this PR.

Nerixyz (approved):
Denormalizing doesn't do anything if the native style is POSIX, so LGTM.

charles-zablit:
> Denormalizing doesn't do anything if the native style is POSIX, so LGTM.

Added this to the description as well 👍

PR #197943, opened 2026-05-15
[AArch64][GlobalISel] Improve multiplication with multiple registers

spaits:
When working on codegen for `llvm.umul.fix.sat` I have recongnized that among for many things, GISel also generates worse code for mul when the data is in multiple registers (for example when the register length is 64 bits but you want to multiply two 128 bit values).

Here is the example ll:
```
define i128 @i128(i128 %a, i128 %b) {
entry:
  %s = mul i128 %a, %b
  ret i128 %s
}
```

This is what GISel gave:
```
  mul	x9, x0, x3
  mul	x8, x0, x2
  umulh	x10, x0, x2
  madd	x9, x1, x2, x9
  mov	x0, x8
  add	x1, x9, x10
  ret
```

And this is what SDAG gives:
```
  umulh	x8, x0, x2
  madd	x8, x0, x3, x8
  mul	x0, x0, x2
  madd	x1, x1, x2, x8
  ret
```
Now GISel gives the same.

To achive this I had to manipulate the order of building the MIR for the big multiplication in `LegalizerHelper::multiplyRegisters`.

Should I rather wrote some combines? Once we agree on a direction I will fix all the tests. (There are lots of targets lots of impacted files an I have limited computing resources.)


arsenm:
The legalizer should directly try to do a better job. Separately, we probably should have some combines on mul / mulhi patterns

spaits:
By "legalizer should directly try to do a better job" you mean in places like the function I have edited or the legalization rules for AArch64 in the LegalizerInfo? We also already have some combines. But the data dependencies of the current (pre-patch) state are really not straight forward, so they can't really do anything w this. I could experiment with more complex combines.

arsenm:
> By "legalizer should directly try to do a better job" you mean in places like the function I have edited 

Yes

spaits:
So should we keep the PR? If so I will fix the tests then.

arsenm:
> So should we keep the PR? If so I will fix the tests then.

Yes

spaits:
Fixed the tests. It seems that the patch either brings minor improvements or no change in code size. I think that's fine.

arsenm (approved):

PR #197946, opened 2026-05-15
[MLIR][OpenMP][Flang][CIR] Replace the OMP specific PtrLikeType interface with the core interface

jsjodin:
This patch replaces the OMP specific pointer like type interface with the interface defined in core MLIR. This will make the code more uniform and will eliminate the dependence of the interface between OMP->CIR and OMP->Flang.

Assisted-by: Cursor / claude-opus-4.6-thinking


andykaylor (commented):
This looks reasonable to me. Is there any way to test it or is this all exercised by existing tests?

jsjodin:
> This looks reasonable to me. Is there any way to test it or is this all exercised by existing tests?

It is exercised by existing tests, except for new functions needed because of the shared interface, so those could potentially be replaced with llvm_unreachable since OpenMP does not use those functions.

andykaylor (approved):
lgtm

jsjodin:
To keep the interface smaller and not implementing extra functions that are not needed I think it is better to keep the code as-is or switch to a cast since they are the preferred solution.

PR #197947, opened 2026-05-15
[flang][cmake][perf-training] Optimize flang with PGO and BOLT

pawosm-arm:
This is an attempt to replicate similar fearture already available to clang. The changes in this patch were made with an intent to reuse as much of existing infrastructure as possible. Namely, two-stage build arrangement, perf-helper.py script and the means for building of the instrumented binaries have been all incorporated in this approach.

It was deliberately chosen to optimize clang along with flang as they are mostly working together in the final toolchain.

See the `llvm/docs/AdvancedBuilds.rst` documentation for more details.


tblah (commented):
Thanks for the PR.

What was your reasoning about the inclusion of these perf training examples in particular? Should we be crafting examples to match each optimisation pass etc? Do you have any idea if it makes much difference how extensive/representative the training set is? (it is okay if you didn't measure that, I'm just curious)

When PGO/Bolt'ing clang and flang at the same time, can they use eachother's profiling data for shared libraries (i.e. llvm/)? I wonder if that could improve our coverage quite a lot.

I wonder if the documentation should be added under flang/docs so that it ends up on the website. There is already a `Building Flang` section in `docs/GettingStarted.md`

pawosm-arm:
> Do you have any idea if it makes much difference how extensive/representative the training set is?

I'm hoping for further contributions of some more representative training examples. I did some measurements (using dbcsr, polyhedron and openblas), the results varied, I plan to quote them in the followup comments next week.

> When PGO/Bolt'ing clang and flang at the same time, can they use eachother's profiling data for shared libraries (i.e. llvm/)? I wonder if that could improve our coverage quite a lot.

Yes, they share the generated `.profraw` files and (as I examined it with the `llvm-profdata` tool), the final merged `.profdata` file being used for building the PGO-optimized binaries, does contain information originating from running both flang and clang (parsing sources, IR generation, using MLIR, basically, all the things that were used for building the training sources suite).

> I wonder if the documentation should be added under flang/docs so that it ends up on the website. There is already a Building Flang section in docs/GettingStarted.md

This feature is so experimental, I'm not sure we want to recommend it for a wider use by announcing it in the docs.

tblah:
> I'm hoping for further contributions of some more representative training examples. I did some measurements (using dbcsr, polyhedron and openblas), the results varied, I plan to quote them in the followup comments next week.

Perhaps you could add an option to train on all of `check-flang`? Or if that is too computationally expensive you could farm useful code examples for each pass in flang or representative lowering examples from the tests. Some of the tests for these passes are just mlir->mlir, but it sounds like your set up is already smart enough to correctly merge profiles from the libraries shared by `bbc`, `fir-opt` (and other tools) and the main flang -fc1 driver.

pawosm-arm:
> Perhaps you could add an option to train on all of `check-flang`?

Yeah, I've tried that, though the amount of disk space it requires to complete makes such change slightly controversial (plus it apparently hits one of the bolt limitations), hence I've prepared a separate PR which covers the change you've proposed here: #198863

pawosm-arm:
I did some performance evaluation of the flang's BOLT+PGO optimization, using some of the example Fortran workloads: dbcsr, polyhedron (pb11) and netlib's lapack-3.6.1 (initially I was considering OpenBLAS, but then I've realized it's mostly C, and what I actually needed was the lapack itself).

I've tried two very different machines, older but more resourceful and very noisy, and much newer, with less storage, yet much quieter as used exclusively for compiler development (no concurrent users at the time of evaluation).

The results below do not cover the builds which used the llvm-test-suite for the PGO training. Using the external suite significantly extends the building process, with no significant benefit (if there was any, it was below the noise level), hence, although this patch allows doing it, I won't be recommending it. AFAIR similar issue has been encountered during the initial works on the clang PGO optimization.

All building times were given in seconds:

## dbcsr

### On the older machine

#### The baseline:

dbcsr, first run:
```
real 270.18
user 392.62
sys 1096.53
```

dbcsr second run:
```
real 260.11
user 383.37
sys 1108.51
```

#### Using the method from this PR:

dbcsr, first run:
```
real 226.04
user 329.56
sys 1048.80
```

dbcsr, second run:
```
real 222.48
user 323.99
sys 1084.69
```

#### Using the method with `check-flang`:

dbcsr, first run:
```
real 206.36
user 295.11
sys 1085.26
```

dbcsr, second run:
```
real 207.53
user 297.80
sys 1076.45
```

### On the newer machine

#### The baseline:

dbcsr, first run:
```
real 82.95
user 103.92
sys 191.41
```

dbcsr, second run:
```
real 82.97
user 103.39
sys 169.72
```

#### Using the method from this PR:

dbcsr, first run:
```
real 75.15
user 95.42
sys 229.32
```

dbcsr, second run:
```
real 74.54
user 94.56
sys 185.64
```

#### Using the method with `check-flang`:

dbcsr, first run:
```
real 72.32
user 90.09
sys 216.06
```

dbcsr, second run:
```
real 72.12
user 88.78
sys 196.33
```
## Polyhedron pb11

The following commands were executed in the `pb11/lin/source` directory:
```
rm -f *.exe
rm -f *.mod
sync
time -p for i in *.f90; do $HOME/upstream/llvm/bin/flang -mcpu=native -O3 -o $i.exe $i; done
```

### On the older machine

#### The baseline:

polyhedron, first run:
```
real 160.23
user 243.16
sys 198.82
```

polyhedron, second run:
```
real 156.35
user 237.63
sys 184.04
```

#### Using the method from this PR:

polyhedron, first run:
```
real 140.78
user 209.01
sys 193.81
```

polyhedron, second run:
```
real 141.83
user 212.34
sys 193.87
```

#### Using the method with `check-flang`:

polyhedron, first run:
```
real 124.52
user 189.04
sys 192.01
```

polyhedron, second run:
```
real 133.84
user 201.05
sys 184.53
```

### On the newer machine

#### The baseline:

polyhedron, first run:
```
real 52.42
user 64.42
sys 18.08
```

polyhedron, second run:
```
real 53.03
user 63.98
sys 19.58
```

#### Using the method from this PR:

polyhedron, first run:
```
real 51.25
user 59.96
sys 20.89
```

polyhedron, second run:
```
real 50.79
user 59.48
sys 20.62
```

#### Using the method with `check-flang`:

polyhedron, first run:
```
real 46.03
user 55.85
sys 18.55
```

polyhedron, second run:
```
real 46.01
user 55.65
sys 18.69
```

## netlib's lapack-3.6.1

The following commands were executed in the `SRC` directory:

```
rm -f *.o
sync
time -p for i in *.f; do $HOME/upstream/llvm/bin/flang -mcpu=native -O3 -c $i; done
```

### On the older machine

#### The baseline:

lapack, first run:
```
real 520.56
user 437.79
sys 188.80
```

lapack, second run:
```
real 513.20
user 431.15
sys 189.26
```

#### Using the method from this PR:

lapack, first run:
```
real 437.28
user 369.06
sys 168.77
```

lapack, second run:
```
real 444.53
user 374.26
sys 172.91
```

#### Using the method with `check-flang`:

lapack, first run:
```
real 435.29
user 347.92
sys 181.41
```

lapack, second run:
```
real 430.71
user 340.31
sys 187.24
```

### On the newer machine

#### The baseline:

lapack, first run:
```
real 210.13
user 180.38
sys 53.89
```

lapack, second run:
```
real 209.83
user 179.91
sys 53.87
```

#### Using the method from this PR:

lapack, first run:
```
real 186.22
user 159.58
sys 49.85
```

lapack, second run:
```
real 186.40
user 159.84
sys 50.04
```

#### Using the method with `check-flang`:

lapack, first run:
```
real 178.04
user 151.22
sys 47.97
```

lapack, second run:
```
real 177.72
user 150.78
sys 47.48
```

pawosm-arm:
Having the values quoted above, we need to decide which method we would like to go ahead with. Note that the devops people will be cursing me for those 200GB+ per build that the `check-flang` method consumes, we must be really certain that the performance gain is worth the extra resource use.

aaupov:
Hi Paul, thank you for extending PGO/BOLT to Flang and measuring the effect. I scanned through the PR and this looks identical to Clang automation, which begs the question of moving it to top-level cmake module. I don't want to block this effort, but do you want to take a stab on that to avoid duplication? cc @petrhosek @llvm-beanz

pawosm-arm:
> Hi Paul, thank you for extending PGO/BOLT to Flang and measuring the effect. I scanned through the PR and this looks identical to Clang automation, which begs the question of moving it to top-level cmake module. I don't want to block this effort, but do you want to take a stab on that to avoid duplication? cc @petrhosek @llvm-beanz

That was my very first idea, but I gave up on it very early. What we're facing here is the original sin of LLVM which is de-facto clang and everything else is piggybacked on clang. If the structure was different (frontend diver as a separate thing, middle end separate, clang and flang as equals) it would be more obvious where to put things like cmake caches. For now, let's keep things where people expect them to see considering current layout: clang things in `clang`, flang things in `flang`. Unless others have different opinion.

kiranchandramohan:
@pawosm-arm Thanks for putting up the patch and sharing the results. Following is an AI generated summary.

### 📊 BOLT+PGO Performance Evaluation Summary

Here is a structured breakdown of the benchmark results provided by @pawosm-arm, comparing the compile-time (Real Time) improvements on the old vs. new machines. Percentages are calculated using the average of the two runs for each scenario.

#### 1. "This PR" Method vs. Baseline
This table tracks the performance gains introduced by the core method in this pull request.

| Benchmark (Avg. Real Time) | 🖥️ Older Machine | ⚡ Newer Machine |
| :--- | :--- | :--- |
| **DBCSR** | 265.15s ➔ **224.26s** <br> *(Improvement: **15.4%**)* | 82.96s ➔ **74.85s** <br> *(Improvement: **9.8%**)* |
| **Polyhedron (pb11)** | 158.29s ➔ **141.31s** <br> *(Improvement: **10.6%**)* | 52.73s ➔ **51.02s** <br> *(Improvement: **3.2%**)* |
| **LAPACK-3.6.1** | 516.88s ➔ **440.91s** <br> *(Improvement: **14.6%**)* | 209.98s ➔ **186.31s** <br> *(Improvement: **11.2%**)* |

---

#### 2. "check-flang" Method vs. Baseline
This table tracks the performance gains when utilizing the `check-flang` training pipeline.

| Benchmark (Avg. Real Time) | 🖥️ Older Machine | ⚡ Newer Machine |
| :--- | :--- | :--- |
| **DBCSR** | 265.15s ➔ **206.95s** <br> *(Improvement: **21.9%**)* | 82.96s ➔ **72.22s** <br> *(Improvement: **12.9%**)* |
| **Polyhedron (pb11)** | 158.29s ➔ **129.18s** <br> *(Improvement: **18.7%**)* | 52.73s ➔ **46.02s** <br> *(Improvement: **12.7%**)* |
| **LAPACK-3.6.1** | 516.88s ➔ **433.00s** <br> *(Improvement: **16.2%**)* | 209.98s ➔ **177.88s** <br> *(Improvement: **15.3%**)* |

---

<details>
<summary>💡 Key Takeaways & Observations</summary>

* **`check-flang` is the clear winner:** Using the `check-flang` method consistently yields the highest percentage reduction in compile times across all tests.
* **Hardware Delta:** The older machine sees slightly higher percentage improvements from the optimizations, though the newer machine remains significantly faster in absolute duration.
* **Test Suite Overhead:** As noted in the evaluation, using the external `llvm-test-suite` for PGO training noticeably slows down the build process without providing a measurable performance benefit, so it is omitted here.
</details>

flang/utils/perf-training/CMakeLists.txt
```diff
@@ -0,0 +1,140 @@
...
+    set(PROFRAW_DEPS "")
+    if(FLANG_PGO_TRAINING_CLANG_COUPLING)
+      list(APPEND PROFRAW_TARGETS generate-clang-profraw)
+      list(APPEND PROFRAW_DIRS ${CLANG_CURRENT_BINARY_DIR})
+      list(APPEND PROFRAW_DEPS clang)
+      if (CLANG_PGO_TRAINING_DATA_SOURCE_DIR)
+        llvm_ExternalProject_Add(generate-clang-profraw-external
+                                 ${CLANG_PGO_TRAINING_DATA_SOURCE_DIR}
+                                 USE_TOOLCHAIN
+                                 EXLUDE_FROM_ALL
```
kiranchandramohan:
```suggestion
                                 EXCLUDE_FROM_ALL
```

flang/utils/perf-training/CMakeLists.txt
```diff
@@ -0,0 +1,140 @@
...
+      ${CMAKE_BINARY_DIR}/profiles/)
+    list(APPEND PROFRAW_DEPS
+      flang
+      flang-rt)
+    if (FLANG_PGO_TRAINING_DATA_SOURCE_DIR)
+      llvm_ExternalProject_Add(generate-flang-profraw-external
+                               ${FLANG_PGO_TRAINING_DATA_SOURCE_DIR}
+                               USE_TOOLCHAIN
+                               ENABLE_FORTRAN
+                               EXLUDE_FROM_ALL
```
kiranchandramohan:
```suggestion
                               EXCLUDE_FROM_ALL
```

kiranchandramohan (commented):
Nit comment inline.

flang/utils/perf-training/CMakeLists.txt
pawosm-arm: Corrected. I can confirm that the same typo is present in the original clang-related code I've based this upon.

flang/utils/perf-training/CMakeLists.txt
pawosm-arm: Fixed

flang/cmake/caches/README.txt
```diff
@@ -0,0 +1,97 @@
+CMake Caches
```
boomanaiden154: I would recommend avoiding a README here and doing something akin to the LLVM advanced builds page, or even just adding a section there for flang.

https://llvm.org/docs/AdvancedBuilds.html

pawosm-arm: Yeah, that should go there too. The idea of having this readme was copied over from the clang cmake caches.

boomanaiden154: I'm wondering if we should just drop the READMEs in favor of actual documentation.

I find most of the READMEs in LLVM are horribly out of date.

pawosm-arm: I've moved the documentation and deleted the `README.txt` file from this PR. Note that he other PR (https://github.com/llvm/llvm-project/pull/198863) has gained more attention recently, and I've also modified it accordingly.

pawosm-arm:
Update: I've added stage2-check-clang and stage2-check-flang targets, see issue #202072 for more details.

pawosm-arm:
I've rebased this one too. Apparently, the BOLT buffer size needs to be extended here too in order to make it complete the full round through the perf-training cycle. I've also introduced the changes needed to keep it in sync with the competing PR #198863

pawosm-arm:
Abandoning it in favor of #198863.

PR #197948, opened 2026-05-15
[GVN] Properly combine AA metadata if available load is hoisted

### LINKED ISSUE #196787: [LICM] Wrong code at -O2/3
Reproducer: https://godbolt.org/z/69naxK1Yx
Testcase:
```c
#include <stdint.h>
#include <stdio.h>
_Bool g1, g28;
void *g7;
int64_t g8, __chk, g16, g15, g12, g10, g9;
void launch()
{
    _Bool c3, c5;
    void *p10 = &c3;
    c5 = g9;
lbl_br5:
    c3 = g16 = g10 = g9;
    if (g28)
    {
        g1 = *(_Bool *)p10;
        g28 = 0;
        g7 = &g16;
        p10 = g7;
        c5 = 1;
        goto lbl_br5;
    }
    else
    {
        g12 = *(int8_t *)p10;
        g9 = g28 = 0 <= g10;
        if (c5) return;
        c5 = g15;
        g15 = g8;
        goto lbl_br5;
    }
}
int main()
{
    launch();
    __chk = g12;
    printf("checksum=0x%016llx\n", (unsigned long long)__chk);
}
```

It traces back to clang-20. And it bisects to LICM: https://godbolt.org/z/d8ec37Woz

antoniofrighetto:
Ensure the AA metadata are properly merged between the new load and the old one during PRE. Actually set `DoesKMove` in `combineMetadataForCSE`, otherwise the new load is assumed not to move, which is not correct if the new load has happened to be hoisted.

Fixes: https://github.com/llvm/llvm-project/issues/196787.


llvm/lib/Transforms/Scalar/GVN.cpp
```diff
@@ -1622,7 +1622,7 @@ void GVNPass::eliminatePartiallyRedundantLoad(
         ++NumPRELoadMoved2CEPred;
         ICF->insertInstructionTo(NewLoad, UnavailableBlock);
         LoadInst *OldLoad = It->second;
-        combineMetadataForCSE(NewLoad, OldLoad, false);
+        combineAAMetadata(NewLoad, OldLoad);
```
cardigan1008: Hi, here is a related miscompilation bug: https://alive2.llvm.org/ce/z/FnconI

The patch fixes only AA metadata for the hoisted-load replacement case. However, `eliminatePartiallyRedundantLoad()` copies several non-AA metadata kinds from the eliminated merge load before the `OldLoad` replacement: !range, !nofpclass, !invariant.load, and !invariant.group, and pointer-result metadata may also be relevant depending on copying behavior. When `OldLoad` is replaced by `NewLoad`, `NewLoad` must not be more restrictive than `OldLoad` for any metadata that can affect value semantics or poison/UB. The new `combineAAMetadata()` call has `AAOnly=true`, so it does not reconcile !range with `OldLoad`. This leaves a semantic gap: the fix assumes only alias-analysis metadata needs recombination when a new load is hoisted, but non-AA result metadata copied from the final load can also be invalid for `OldLoad`'s original control-flow region.

For more details, please refer to https://archer.top/artifact/run/197948/run.review.md

> Found with [Archer](https://archer.top/). Please let me know if anything is wrong.

nikic: Isn't the right fix here to set DoesKMove=true?

antoniofrighetto: Right, went with setting DoesKMove originally, then the comment "For this to be correct, K cannot be hoisted" slightly threw me off and switched API (without even noticing AAOnly) :( Updated, thanks.

alinas (approved):

llvm/lib/Transforms/Scalar/GVN.cpp
```diff
@@ -1622,7 +1622,7 @@ void GVNPass::eliminatePartiallyRedundantLoad(
         ++NumPRELoadMoved2CEPred;
         ICF->insertInstructionTo(NewLoad, UnavailableBlock);
         LoadInst *OldLoad = It->second;
-        combineMetadataForCSE(NewLoad, OldLoad, false);
+        combineMetadataForCSE(NewLoad, OldLoad, /*DoesKMove*/ true);
```
nikic:
```suggestion
        combineMetadataForCSE(NewLoad, OldLoad, /*DoesKMove=*/true);
```
nit: This is the preferred style nowadays.

llvm/test/Transforms/GVN/PRE/pre-load.ll
```diff
@@ -1551,3 +1551,76 @@ if.end:
...
+; MDEP-LABEL: @test_combine_metadata(
+; MDEP-NEXT:  entry:
+; MDEP-NEXT:    [[C3:%.*]] = alloca i8, align 1
+; MDEP-NEXT:    br label [[LOOP:%.*]]
+; MDEP:       loop:
+; MDEP-NEXT:    [[PHI:%.*]] = phi ptr [ @g_long, [[IF_THEN:%.*]] ], [ [[C3]], [[ENTRY:%.*]] ]
+; MDEP-NEXT:    store i64 [[ARG:%.*]], ptr @g_long, align 8, !tbaa [[TBAA0:![0-9]+]]
+; MDEP-NEXT:    [[TR:%.*]] = trunc i64 [[ARG]] to i8
+; MDEP-NEXT:    store i8 [[TR]], ptr [[C3]], align 1, !tbaa [[TBAA4:![0-9]+]]
+; MDEP-NEXT:    [[V1_PRE:%.*]] = load i8, ptr [[PHI]], align 1, !tbaa [[TBAA6:![0-9]+]]
```
nikic: Need to also generate the metadata check lines? Otherwise we're not really checking that the result is correct...

nikic (approved):
LGTM

llvm/test/Transforms/GVN/PRE/pre-load.ll
antoniofrighetto: Sure thing (I think I was using an UTC older version...).

Issue #197824, opened 2026-05-14
[Compiler-rt] Tracking Issue: Use LLVM Libc math routines in builtins


hulxv:
This issue tracks the progress of the using of libc math routines in compiler-rt builtins.

RFC: https://discourse.llvm.org/t/rfc-explore-using-llvm-libc-math-routines-for-compiler-rt-floating-point-builtins/89670

- **Quad-Precision (128-bit / `tf`) Arithmetic**
  - [x] `__addtf3` #197950
  - [x] `__subtf3` #207092
  - [x] `__multf3` #207092
  - [x] `__divtf3` #207092

- **Quad-Precision Comparisons**
  - [ ] `__letf2` #212653
  - [ ] `__getf2` #212653
  - [ ] `__unordtf2` #212653
  - [ ] `__eqtf2` #212653
  - [ ] `__netf2` #212653
  - [ ] `__lttf2` #212653
  - [ ] `__gttf2` #212653

- **Double-Precision Comparisons**
  - [ ] `__ledf2` #212652
  - [ ] `__gedf2` #212652
  - [ ] `__unorddf2` #212652
  - [ ] `__eqdf2` #212652
  - [ ] `__nedf2` #212652
  - [ ] `__ltdf2` #212652
  - [ ] `__gtdf2` #212652

- **Single-Precision Comparisons**
  - [ ] `__lesf2` #212651
  - [ ] `__gesf2` #212651
  - [ ] `__unordsf2` #212651
  - [ ] `__eqsf2` #212651
  - [ ] `__nesf2` #212651
  - [ ] `__ltsf2` #212651
  - [ ] `__gtsf2` #212651

- **Int -> Quad-Precision Conversions**
  - [ ] `__floatsitf` #215725
  - [ ] `__floatditf` #215725
  - [ ] `__floattitf` #215725
  - [ ] `__floatunsitf` #215725
  - [ ] `__floatunditf` #215725
  - [ ] `__floatuntitf` #215725

- **Quad-Precision -> Int Conversions**
  - [ ] `__fixtfsi` #215726
  - [ ] `__fixtfdi` #215726
  - [ ] `__fixtfti` #215726
  - [ ] `__fixunstfsi` #215726
  - [ ] `__fixunstfdi` #215726
  - [ ] `__fixunstfti` #215726

- **Float <-> Float Extend/Truncate Conversions**
  - [x] `__extendsftf2` #209984
  - [x] `__extenddftf2` #209984
[...]

PR #197950, opened 2026-05-15
[compiler-rt][builtins] introduce libc math routines

hulxv:
Introduce using libc math routines in compiler-rt builtins. this pr adds `COMPILER_RT_USE_LIBC_MATH` flag and uses math routines in basic arithmetic operations for Quad-Precision values

RFC: https://discourse.llvm.org/t/rfc-explore-using-llvm-libc-math-routines-for-compiler-rt-floating-point-builtins/89670

Part of #197824


.github/workflows/libc-overlay-tests.yml
```diff
@@ -109,9 +110,16 @@
...
       run: >
         cmake 
         --build ${{ steps.strings.outputs.build-output-dir }} 
         --parallel 
         --target check-libc
+    
+    - name: Test compiler-rt builtins
+      run: >
+        cmake
+        --build ${{ steps.strings.outputs.build-output-dir }}
```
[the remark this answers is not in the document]
hulxv: @lntue should worry about that?

libc/src/__support/builtins/addtf3.h
bassiounix, line 1: in general this shouldn't live in the src code but the shared code like rpc

compiler-rt/lib/builtins/addtf3.cpp
bassiounix, line 1: I fear that the change of language would cause name mangling for the functions, so I prefer it to reside in its own file and include the implemented cpp builtins incrementally if the flag is set with the rest of the builtins.

bassiounix (changes requested):

compiler-rt/lib/builtins/addtf3.cpp
hulxv: mangling is disabled

https://github.com/llvm/llvm-project/pull/197950/changes#diff-16a8d59eb5c2c5ca9445eabd5fb48b3c076f4f23d1e85a5a73ea7c0cd3bd3766R25

libc/src/__support/builtins/addtf3.h
hulxv: What does that change? shouldn't be like the math functions?

.github/workflows/compiler-rt-libc-builtins-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Configure CMake
+        shell: bash
+        run: |
+          cmake \
+            -B /__w/llvm-project/llvm-project/build \
```
lntue: just use `build`

.github/workflows/compiler-rt-libc-builtins-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+            -S /__w/llvm-project/llvm-project/llvm \
```
lntue: just use `llvm`

.github/workflows/compiler-rt-libc-builtins-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+            -G Ninja \
+            -DCMAKE_C_COMPILER=clang-23 \
+            -DCMAKE_CXX_COMPILER=clang++-23 \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DLLVM_ENABLE_ASSERTIONS=ON \
+            -DLLVM_ENABLE_PROJECTS=clang \
+            -DLLVM_ENABLE_RUNTIMES="libc;compiler-rt" \
```
lntue: remove `libc`

.github/workflows/compiler-rt-libc-builtins-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+            -DCOMPILER_RT_USE_LIBC_MATH=ON \
+            -DCOMPILER_RT_INCLUDE_TESTS=ON
+
+      - name: Test compiler-rt builtins
+        shell: bash
+        run: |
+          cmake \
+            --build /__w/llvm-project/llvm-project/build \
```
lntue: `build`

libc/src/__support/builtins/addtf3.h
michaelrj-google: actually we want this code to live in `src`. `shared` was intended to just redirect functions for other users, not contain any code of its own. The RPC code is owned by the offloading team and has different standards.

This code is in the correct place as-is.

libc/shared/builtins/addtf3.h
```diff
@@ -0,0 +1,29 @@
+//===-- Shared __addtf3 function -------------------------------*- C++ -*-===//
```
michaelrj-google: all these files should follow the new LLVM header coding style: https://llvm.org/docs/CodingStandards.html#file-headers

libc/shared/builtins/addtf3.h
```diff
@@ -0,0 +1,29 @@
...
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_LIBC_SHARED_BUILTINS_ADDTF3_H
+#define LLVM_LIBC_SHARED_BUILTINS_ADDTF3_H
+
+#include "include/llvm-libc-types/float128.h"
```
michaelrj-google, line 18: does `float128` have a proxy header? If it doesn't yet, should it @lntue?

michaelrj-google (commented):
Looks like a good start to me

compiler-rt/lib/builtins/CMakeLists.txt
```diff
@@ -243,6 +243,26 @@ set(GENERIC_TF_SOURCES
...
+option(COMPILER_RT_USE_LIBC_MATH
+      "Use LLVM libc math routines for floating-point builtins" OFF)
+
+if(COMPILER_RT_USE_LIBC_MATH)
+  set(LLVM_LIBC_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../../libc"
+      CACHE PATH "Path to LLVM libc source directory")
+
+  if(NOT EXISTS "${LLVM_LIBC_SRC_DIR}/src/__support/FPUtil/FPBits.h")
+    message(FATAL_ERROR "LLVM libc not found at ${LLVM_LIBC_SRC_DIR}")
+  endif()
```
lntue: Something like:
```
include(FindLibcCommonUtils)
if(NOT TARGET llvm-libc-common-utilities)
  message(FATAL_ERROR "LLVM libc is not found at ${libc_path}.")
endif()
```

compiler-rt/lib/builtins/CMakeLists.txt
```diff
@@ -243,6 +243,26 @@ set(GENERIC_TF_SOURCES
...
+
+  include_directories(SYSTEM ${LLVM_LIBC_SRC_DIR})
```
lntue: remove `SYSTEM`, use `${LLVM_SOURCE_DIR}/../libc` for the path.

compiler-rt/lib/builtins/fp_libc_config.h
```diff
@@ -0,0 +1,12 @@
+#ifndef FP_LIBC_CONFIG_H
+#define FP_LIBC_CONFIG_H
+
+#ifndef LIBC_COPT_PUBLIC_PACKAGING
+#define LIBC_COPT_PUBLIC_PACKAGING
+#endif
```
lntue: Remove this

compiler-rt/lib/builtins/int_lib.h
```diff
@@ -20,14 +20,22 @@
 
 // ABI macro definitions
 
+// In C++ translation units we must give compiler-rt entry points C linkage so
+// that their symbol names are not mangled.  C TUs are unaffected.
+#ifdef __cplusplus
```
lntue: why do we need this change?

lntue: We don't need this one.  Just add `extern "C"` directly to the `addtf3.cpp` directly.  We that file can't be compiled in C.

compiler-rt/lib/builtins/fp_libc_config.h
```diff
@@ -0,0 +1,12 @@
...
+
+#ifndef LIBC_NAMESPACE
+#define LIBC_NAMESPACE __llvm_libc_rt
+#endif
+
```
lntue, line 23: you will also need to disable fpexcept and errno in the implementation:
```
#define LIBC_MATH (LIBC_MATH_NO_ERRNO | LIBC_MATH_NO_EXCEPT)
```

compiler-rt/lib/builtins/int_lib.h
hulxv: I added it before as a shortcut for the unified file of legacy and libc. I think it's not that important now

compiler-rt/lib/builtins/fp_libc_config.h
```diff
@@ -0,0 +1,10 @@
+#ifndef FP_LIBC_CONFIG_H
+#define FP_LIBC_CONFIG_H
+
+#ifndef LIBC_NAMESPACE
+#define LIBC_NAMESPACE __llvm_libc_rt
+#endif
+
+#define LIBC_MATH (LIBC_MATH_NO_ERRNO | LIBC_MATH_NO_EXCEPT)
+
+#endif
```
bassiounix:
```suggestion
#endif // FP_LIBC_CONFIG_H

```

bassiounix (approved):

libc/shared/builtins/addtf3.h
lntue: not yet.  For float128 type, especially for compiler-rt usage, we will wait for the generic class from the other Float80-Float128 GSoC project to be ready, then we will restructure how float128 type is detected / defined in general.

lntue (changes requested):
This PR is for demonstrating the integration.  Once everything looks ok, we will need to split it into 3 PRs in the following order:
- [libc] Add addtf3.h and builtins.h to shared headers (+ smoke tests)
- [compiler-rt] Add compiler-rt cmake option to build compiler builtins with LLVM libc
- [CI] Add precommit CIs to test compiler-rt + LLVM libc integration.

@compnerd @statham-arm

compiler-rt/lib/builtins/fp_libc_config.h
```diff
@@ -0,0 +1,10 @@
+#ifndef FP_LIBC_CONFIG_H
```
lntue, line 17: LLVM header is missing? https://llvm.org/docs/CodingStandards.html#file-headers

.github/workflows/compiler-rt-libc-builtins-tests.yml
```diff
@@ -0,0 +1,57 @@
...
+      - "libc/shared/builtins.h"
+      - "libc/shared/builtins/**"
+      - "libc/src/__support/builtins/**"
+      - "libc/src/__support/FPUtil/**"
+      - ".github/workflows/compiler-rt-libc-builtins-tests.yml"
+
+jobs:
+  build:
+    name: builtins (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
```
statham-arm: Shouldn't there be a guard here to make sure this workflow doesn't run on all downstream repositories? Something like this (from another recently modified workflow file):
```
    if: github.repository_owner == 'llvm'
```

hulxv: Thanks! I will add it!

statham-arm (approved):

compiler-rt/lib/builtins/addtf3.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+#define QUAD_PRECISION
+#include "fp_lib.h"
+
+#include "fp_libc_config.h"
+#include "int_lib.h"
+#include "shared/builtins/addtf3.h"
+
+extern "C" COMPILER_RT_ABI fp_t __addtf3(fp_t a, fp_t b) {
+  return LIBC_NAMESPACE::shared::addtf3(a, b);
```
compnerd, line 21: Why the macro for the namespace?

hulxv: libc wraps the namespace inside LIBC_NAMESPACE macro that could be defined at build-time. It's not a static name

@lntue correct me if I am wrong, please

lntue (approved):

PR #197953, opened 2026-05-15
[lldb] Fix test timeouts on sanitized builds

da-viper:
Switch to a fixed timeout and a shorter polling interval to reduce wasted time. The total timeout in non-sanitized is 120s (same as the existing implementation) and in sanitized builds, it is 600s matching `lit_config.maxIndividualTestTime`.

Previously, sanitized builds could spend up to 20 seconds idling. While this specific function wouldn't time out itself, it consumed a significant portion of the overall test time budget. This would eventually cause the test to hit the
`lit_config.maxIndividualTestTime` limit (currently 600 seconds).

This delay accumulates quickly in tests with multiple subtests that implicitly rely on `wait_for_file_on_target` (for example, `TestLLDBGdbServer.py`). 
If a test has 5 subtests, it could lose 100 seconds (20s * 5) to idle polling, leaving only 500 seconds for the entire test to finish.

### DIFF
--- a/lldb/packages/Python/lldbsuite/test/lldbutil.py
+++ b/lldb/packages/Python/lldbsuite/test/lldbutil.py
@@ -12,6 +12,7 @@ import os
 import re
 import sys
 import subprocess
+import time
 from typing import Dict, Tuple
 
 # LLDB modules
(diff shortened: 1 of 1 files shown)


adrian-prantl (approved):
Oh yeah, that sleep looks very wrong!

adrian-prantl:
It's quite possible that this is the result of me going and increasing all variables named "timeout" without realizing that this one was used for a sleep()

bulbazord (approved):
thanks for cleaning it up!

PR #197954, opened 2026-05-15
[lldb] Make use of ConstString(StringRef) where applicable (NFC)

kastiglione:
Replace `ConstString(char *)`, mostly with the `StringRef` constructor. Eliminates some unnecessary `strlen`, and a few copies.


jimingham (approved):
Looks nicer too.

bulbazord (approved):
beautiful

PR #197957, opened 2026-05-15
[mlir][SPIR-V] Expand spirv.ISubBorrow in WebGPU prepare pass

aobolensk:


mlir/lib/Dialect/SPIRV/Transforms/SPIRVWebGPUTransforms.cpp
```diff
@@ -204,6 +204,43 @@ struct ExpandAddCarryPattern final : OpRewritePattern<IAddCarryOp> {
   }
 };
 
+struct ExpandSubBorrowPattern final : OpRewritePattern<ISubBorrowOp> {
```
kuhar: Can we turn this into a template to handle both ops?

kuhar (approved):

PR #197958, opened 2026-05-15
[mlir][vulkan] Add arith.subui_extended integration test

aobolensk:


fabianmcg (approved):
I see there's a precedent for this kind of test in Vulkan, so I'll approve. But, why does this needs an integration test? If you look at other GPU integration test folders, they focus on features (more like smoke kind of testing) rather than on single minor ops.

aobolensk:
> I see there's a precedent for this kind of test in Vulkan, so I'll approve. But, why does this needs an integration test? If you look at other GPU integration test folders, they focus on features (more like smoke kind of testing) rather than on single minor ops.

As I noticed there are tests for addui_extended and mului_extended, so I decided it is feasible

fabianmcg:
Ok, so let's not add more unless there's a stronger reason (fine to merge this one). While I see the value in this kind of tests, IMO this would belong to a different test suite.

aobolensk:
> Ok, so let's not add more unless there's a stronger reason (fine to merge this one). While I see the value in this kind of tests, IMO this would belong to a different test suite.

Ok, thanks for the feedback!

PR #197961, opened 2026-05-15
[clang][deps] Reuse `ModuleDepCollector`

jansvoboda11:
This PR starts reusing `ModuleDepCollector` across calls to `CompilerInstanceWithContext::computeDependencies()`. This allows reusing parts of the dependency graph constructed in previous calls, skipping lots of work on some workloads.

### DIFF
--- a/clang/include/clang/DependencyScanning/ModuleDepCollector.h
+++ b/clang/include/clang/DependencyScanning/ModuleDepCollector.h
@@ -132,6 +132,8 @@ public:
   /// invocation, (e.g. disable implicit modules, add explicit module paths).
   void applyDiscoveredDependencies(CompilerInvocation &CI);
 
+  void clearLocalState();
+
 private:
   friend ModuleDepCollectorPP;
 
@@ -146,7 +148,7 @@ private:
   const PrebuiltModulesAttrsMap PrebuiltModulesASTMap;
   /// Directory paths known to be stable through an active development and build
   /// cycle.
-  const ArrayRef<StringRef> StableDirs;
+  const SmallVector<StringRef> StableDirs;
   /// Path to the main source file.
   std::string MainFile;
   /// Non-modular file dependencies. This includes the main source file and
--- a/clang/include/clang/Tooling/DependencyScanningTool.h
+++ b/clang/include/clang/Tooling/DependencyScanningTool.h
@@ -168,6 +168,8 @@ class CompilerInstanceWithContext {
   // Compiler Instance
   std::unique_ptr<CompilerInstance> CIPtr;
 
+  std::shared_ptr<dependencies::ModuleDepCollector> MDC;
+
   // Source location offset.
   int32_t SrcLocOffset = 0;
 
--- a/clang/lib/DependencyScanning/ModuleDepCollector.cpp
+++ b/clang/lib/DependencyScanning/ModuleDepCollector.cpp
@@ -459,6 +459,17 @@ void ModuleDepCollector::applyDiscoveredDependencies(CompilerInvocation &CI) {
   }
 }
 
+void ModuleDepCollector::clearLocalState() {
+  MainFile.clear();
+  FileDeps.clear();
+  DirectPrebuiltModularDeps.clear();
+  DirectModularDeps.clear();
+  DirectImports.clear();
+  VisibleModules.clear();
+  ProvidedStdCXXModule.reset();
+  RequiredStdCXXModules.clear();
+}
+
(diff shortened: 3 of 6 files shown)


clang/include/clang/DependencyScanning/ModuleDepCollector.h
```diff
@@ -132,6 +132,8 @@ class ModuleDepCollector final : public DependencyCollector {
   /// invocation, (e.g. disable implicit modules, add explicit module paths).
   void applyDiscoveredDependencies(CompilerInvocation &CI);
 
+  void clearLocalState();
```
cyndyishida, line 135: nit: `Local` here effectively means the dependency info that could be unique for the module name requested in  `computeDependencies`. I think that's worth mentioning as a doc comment, especially if new ones get added over time.

clang/include/clang/Tooling/DependencyScanningTool.h
```diff
@@ -168,6 +168,8 @@ class CompilerInstanceWithContext {
   // Compiler Instance
   std::unique_ptr<CompilerInstance> CIPtr;
 
+  std::shared_ptr<dependencies::ModuleDepCollector> MDC;
```
cyndyishida, line 171: nit: This is the only attribute missing a doc string. Maybe the rest should be removed in a separate commit? They all seem quite self-explanatory, but I just noticed because it wasn't consistent with the rest of the _context_

cyndyishida (approved):
LGTM as is. I do wonder if we could have a stronger contract of what dep info is `Local` and therefore should be cleared. 

e.g., defining a `struct LocalDeps`  that holds that info that is non-owning to `ModuleDepCollector` and uses RAII to go out of scope at the end of `computeDependencies`. Not sure how hard that would be to pipe through non-`CompilerInstancewithContext` calls though.

clang/lib/DependencyScanning/ModuleDepCollector.cpp
```diff
@@ -657,8 +668,20 @@ void ModuleDepCollector::run(DependencyConsumer &Consumer) {
...
                                                   MDC.RequiredStdCXXModules);
 
-  for (auto &&I : MDC.ModularDeps)
-    Consumer.handleModuleDependency(*I.second);
+  llvm::DenseSet<serialization::ModuleFile *> HandledModuleDependencies;
+  std::function<void(serialization::ModuleFile *)> HandleModuleDependency =
+      [&](serialization::ModuleFile *MF) {
+        if (HandledModuleDependencies.insert(MF).second) {
+          for (serialization::ModuleFile *MFI : MF->Imports)
+            HandleModuleDependency(MFI);
```
benlangmuir, line 676: I'm probably being dense, but how does this change relate to the overall PR? I'm not clear why we need to do this on all the transitive dependencies now.

jansvoboda11: For now I wanted to keep the existing behavior that each call to `ModuleDepCollector::run()` reports all transitive dependencies for the current module/TU. It would make sense to report only newly-discovered dependencies, but I'd like to keep that for some future PR.

jansvoboda11:
> LGTM as is. I do wonder if we could have a stronger contract of what dep info is `Local` and therefore should be cleared.
> 
> e.g., defining a `struct LocalDeps` that holds that info that is non-owning to `ModuleDepCollector` and uses RAII to go out of scope at the end of `computeDependencies`. Not sure how hard that would be to pipe through non-`CompilerInstancewithContext` calls though.

I was thinking about making the distinction clearer too. Grouping these into a separate type and clearing just a single top-level object instance would be nicer for sure. I'll explore the RAII approach vs keeping explicit calls to `clearLocalState()` in a future PR.

benlangmuir (approved):

clang/lib/DependencyScanning/ModuleDepCollector.cpp
```diff
@@ -459,6 +459,17 @@ void ModuleDepCollector::applyDiscoveredDependencies(CompilerInvocation &CI) {
   }
 }
 
+void ModuleDepCollector::clearLocalState() {
+  MainFile.clear();
```
qiongsiwu, line 463: It seems that `MainFile` does not really change across by-name lookups (the fake input file is setup in `initVFSForByNameScanning`). Also, I don't think it is used at all. Maybe we can remove it? I removed it locally and all the clang tests pass. If it is safe, we can remove it later.

qiongsiwu (approved):
LGTM! Thanks for doing this!!

PR #197962, opened 2026-05-15
[libc] Switch libc-shared-tests precommit CI to use docker image.

lntue:


.github/workflows/libc-shared-tests.yml
```diff
@@ -48,6 +48,8 @@
   linux-cross-arch-shared:
     name: libc-shared-tests on (${{ matrix.arch }})
     runs-on: ubuntu-24.04
+    container:
+      image: 'ghcr.io/llvm/libc-ubuntu-24.04:latest'
```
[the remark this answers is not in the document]
boomanaiden154: This should be pinned to a hash, not just `latest`.

.github/workflows/libc-shared-tests.yml
```diff
@@ -48,6 +48,8 @@ jobs:
   linux-cross-arch-shared:
     name: libc-shared-tests on (${{ matrix.arch }})
     runs-on: ubuntu-24.04
+    container:
+      image: 'ghcr.io/llvm/libc-ubuntu-24.04:70cbd7591abe'
```
boomanaiden154, line 56: We want the full SHA here, not the short SHA.

lntue: That's the only one I got out from https://github.com/llvm/llvm-project/actions/runs/25900016540/job/76131744889 that works.

kaladron: This a tag, not a git sha1, so this is the entire thing.

kaladron (approved):

PR #197963, opened 2026-05-15
[flang-rt][test] Skip Driver/safe-trampoline-gnustack.f90 on Solaris

rorth:
The `flang-rt :: Driver/safe-trampoline-gnustack.f90` test `FAIL`s on Solaris.  It checks for a `GNU_STACK` section, however that isn't used and not emitted on Solaris, so this patch just skips the test.

Tested on `x86_64-pc-solaris2.11` and `x86_64-pc-linux-gnu`.


kkwli (approved):
LG. Thanks.

PR #197966, opened 2026-05-15
[lldb] Setup session when calling Python interfaces

Nerixyz:
When using the Python interfaces like scripted frame providers in lldb-dap, Python's `print` statements don't show in the output window. They do show when done in synthetic children or summaries.

This is because when calling the interface functions/methods, the stdout/stderr handles were never changed. This PR fixes the issue by passing `InitSession` (and `TearDownSession`) to the `Locker`. As a side effect, this will also set `lldb.debugger`.

### DIFF
--- a/lldb/source/Plugins/ScriptInterpreter/Python/ScriptInterpreterPython.cpp
+++ b/lldb/source/Plugins/ScriptInterpreter/Python/ScriptInterpreterPython.cpp
@@ -656,8 +656,9 @@ bool ScriptInterpreterPythonImpl::EnterSession(uint16_t on_entry_flags,
     run_string.PutCString("; lldb.frame = lldb.thread.GetSelectedFrame ()");
     run_string.PutCString("')");
   } else if (on_entry_flags & Locker::InitDebugger) {
-    // If we aren't initing the globals, we should still always set the
-    // debugger (since that is always unique.)
+    // In non-interactive mode (e.g. type summaries), initialize the debugger
+    // only if requested. This should not be used in new extensions and is kept
+    // for backwards compatibility.
     run_string.Printf("run_one_line (%s, 'lldb.debugger_unique_id = %" PRIu64,
                       m_dictionary_name.c_str(), m_debugger.GetID());
     run_string.Printf(
--- a/lldb/source/Plugins/ScriptInterpreter/Python/ScriptInterpreterPython.cpp
+++ b/lldb/source/Plugins/ScriptInterpreter/Python/ScriptInterpreterPython.cpp
@@ -369,8 +369,7 @@ bool ScriptInterpreterPythonImpl::Locker::DoFreeLock() {
 bool ScriptInterpreterPythonImpl::Locker::DoTearDownSession() {
   if (!m_python_interpreter)
     return false;
-  m_python_interpreter->LeaveSession(m_on_leave & TearDownGlobals,
-                                     m_on_leave & TearDownStdio);
+  m_python_interpreter->LeaveSession(m_on_leave);
   return true;
 }
 
@@ -550,13 +549,12 @@ ScriptInterpreterPythonImpl::CreateInstance(Debugger &debugger) {
   return std::make_shared<ScriptInterpreterPythonImpl>(debugger);
 }
 
-void ScriptInterpreterPythonImpl::LeaveSession(bool clear_globals,
-                                               bool clear_stdio) {
+void ScriptInterpreterPythonImpl::LeaveSession(uint16_t leave_flags) {
   Log *log = GetLog(LLDBLog::Script);
   if (log)
     log->PutCString("ScriptInterpreterPythonImpl::LeaveSession()");
 
-  if (clear_globals) {
+  if (leave_flags & Locker::TearDownGlobals) {
     // Unset the LLDB global variables.
     RunSimpleString("lldb.debugger = None; lldb.target = None; lldb.process "
                     "= None; lldb.thread = None; lldb.frame = None");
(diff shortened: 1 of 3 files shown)


medismailben:
If we do this, we might as well pass `InitGlobals`.

JDevlieghere:
> As a side effect, this will also set `lldb.debugger`.

I think we'll need to split that off then. The `lldb.*` convenience variables should only be set when in the interactive interpreter: https://lldb.llvm.org/use/tutorials/python-embedded-interpreter.html

Nerixyz:
I also noticed that `LeaveSession::FreeLock`, `LeaveSession::FreeAcquiredLock`, and `OnEntry::AcquireLock` are never checked. They could be removed.

Edit: I think I'm too tired for this now. I'll take a look tomorrow.

jimingham:
> > As a side effect, this will also set `lldb.debugger`.
> 
> I think we'll need to split that off then. The `lldb.*` convenience variables should only be set when in the interactive interpreter: https://lldb.llvm.org/use/tutorials/python-embedded-interpreter.html

I've always been of two minds about this.  The reason why these convenience variables are only available in the interactive interpreter is that when you aren't coming directly from the CommandInterpreter there's no sensible way to pick which target, process, thread or frame to stick in them.  Using the "Currently Selected" entities is pretty much always the wrong thing to do if you aren't coming directly from the CommandInterpreter - since the "Currently Selected" entities are only w.r.t. the current state of the CommandInterpreter when a command is invoked and can get changed out from under you (e.g. if you have two targets and one stops asynchronously, it will get selected as a part of announcing the stop).

But debugger doesn't suffer from this problem  The ScriptInterpreter -> Debugger map is 1-1, so you always know which debugger to select.

If people found it handy to have around, I think it would be fine to always set the debugger on entry - even though it is emphatically not fine to do so for target, process, thread and frame...

Nerixyz:
> If people found it handy to have around, I think it would be fine to always set the debugger on entry - even though it is emphatically not fine to do so for target, process, thread and frame...

From a [search on GitHub for `lldb.debugger`](https://github.com/search?q=%2Flldb%5C.debugger%2F%20lang%3APython&type=code), it looks like this is used - not just in commands, also in synthetic children providers. 

> I think we'll need to split that off then. The `lldb.*` convenience variables should only be set when in the interactive interpreter: [lldb.llvm.org/use/tutorials/python-embedded-interpreter.html](https://lldb.llvm.org/use/tutorials/python-embedded-interpreter.html)

To clarify, the Python interfaces affected by this PR never set `lldb.debugger` nor `lldb.{target, thread, etc.}`. The first one is set when `InitSession` is specified (synthetic children, summaries) and the second one is specified in the interactive interpreter [here](https://github.com/llvm/llvm-project/blob/f5f49343053f9f27d24a9da3424bd30387c72a48/lldb/source/Plugins/ScriptInterpreter/Python/ScriptInterpreterPython.cpp#L642-L662).

Given that synthetic children/summaries have `lldb.debugger` set, it might seem counterintuitive that it's not set in the interfaces.

---

I updated the PR to have two commits. In the first, `lldb.debugger` is set, and in the second one, only the stdio handles are set. Synthetic children/summaries are unaffected.
If we decide to set `lldb.debugger` here, I can remove the commit.

JDevlieghere:
> I've always been of two minds about this. The reason why these convenience variables are only available in the interactive interpreter is that when you aren't coming directly from the CommandInterpreter there's no sensible way to pick which target, process, thread or frame to stick in them. Using the "Currently Selected" entities is pretty much always the wrong thing to do if you aren't coming directly from the CommandInterpreter - since the "Currently Selected" entities are only w.r.t. the current state of the CommandInterpreter when a command is invoked and can get changed out from under you (e.g. if you have two targets and one stops asynchronously, it will get selected as a part of announcing the stop).
> 
> But debugger doesn't suffer from this problem The ScriptInterpreter -> Debugger map is 1-1, so you always know which debugger to select.
> 
> If people found it handy to have around, I think it would be fine to always set the debugger on entry - even though it is emphatically not fine to do so for target, process, thread and frame...

Yeah, we discussed this when I tightened this check a few years ago. My counter-argument to that at the time was that it's much easier to tell users "this is only available in interactive mode" than explaining why only some of it makes sense. Anecdotally, I haven't seen many questions about this in our support channels after I made that change, while this was something that would come up quite often and lead to confusion. I feel pretty strongly about sticking with the current policy.

lldb/source/Plugins/ScriptInterpreter/Python/ScriptInterpreterPython.cpp
```diff
@@ -550,21 +550,24 @@ ScriptInterpreterPythonImpl::CreateInstance(Debugger &debugger) {
   return std::make_shared<ScriptInterpreterPythonImpl>(debugger);
 }
 
-void ScriptInterpreterPythonImpl::LeaveSession() {
+void ScriptInterpreterPythonImpl::LeaveSession(bool clear_globals,
+                                               bool clear_stdio) {
```
JDevlieghere: Why not pass the `uint16_t on_leave_flags` here? The double bools seem more error prone than passing the flag.

lldb/source/Plugins/ScriptInterpreter/Python/ScriptInterpreterPython.cpp
```diff
@@ -650,7 +657,7 @@ bool ScriptInterpreterPythonImpl::EnterSession(uint16_t on_entry_flags,
     run_string.PutCString("; lldb.thread = lldb.process.GetSelectedThread ()");
     run_string.PutCString("; lldb.frame = lldb.thread.GetSelectedFrame ()");
     run_string.PutCString("')");
-  } else {
+  } else if (on_entry_flags & Locker::InitDebugger) {
     // If we aren't initing the globals, we should still always set the
```
JDevlieghere: This needs to be updated, right?

Nerixyz: I updated the comment to reflect https://github.com/llvm/llvm-project/pull/197966#issuecomment-4467548762.

PR #197974, opened 2026-05-15
 [Lit] Introduce mechanisms for disabling daemonization for a test or test folder

BStott6:
This PR is dependent on https://github.com/llvm/llvm-project/pull/197924; the changes in the first commit in this PR are from that PR and its dependencies. Please review from the 2nd commit onwards.

This PR introduces two mechanisms for tests that fail when daemonization is enabled to opt out of daemonization. Both work by removing the in-process built in which is responsible for invoking the daemon instead of running the tool normally, so that tools are run normally instead.
 
Daemonization can be disabled for a specific tool in one specific test by adding the

```
DISABLE_DAEMON: {tool name}
```

directive to the test, for example:

```
DISABLE_DAEMON: opt
```

Daemonization can be disabled for a whole directory of tests by calling `daemon_config.disable_daemon_tool` in the `lit.local.cfg`, for example:

```
import lit.llvm.daemon_config
lit.llvm.daemon_config.disable_daemon_tool(config, "opt")
```

The name passed should be the same name passed to `use_daemon_tool` when daemonization was enabled for the tool.

With each commit I have added tests to verify that the mechanism indeed disables the in-process built-in, within the existing `llvm-daemon-integration` Lit tests. As these tests register the in-process built-in for ExampleDaemon under a special key, "%invoke-ExampleDaemon", unregistering the daemon will cause this to no longer refer to a command, so the tests that try to use it fail - this is how we make sure the in-process built-in was removed.

### DIFF
--- a/llvm/CMakeLists.txt
+++ b/llvm/CMakeLists.txt
@@ -1380,6 +1380,7 @@ add_subdirectory(include)
 add_subdirectory(lib)
 
 if( LLVM_INCLUDE_UTILS )
+  add_subdirectory(utils/ExampleDaemon)
   add_subdirectory(utils/FileCheck)
   add_subdirectory(utils/PerfectShuffle)
   add_subdirectory(utils/count)
--- a/llvm/docs/DaemonMode.rst
+++ b/llvm/docs/DaemonMode.rst
@@ -1,0 +1,135 @@
+============================
+Running Tools in Daemon Mode
+============================
+
+.. contents::
+   :local:
+
+Introduction
+============
+
+Process creation incurs a significant overhead, especially on Windows. This
+document describes the "daemon mode" execution model for LLVM tools, where
+tools run as persistent daemon processes that can perform many tasks, with
+different arguments and input, in the same process. The original motivation
+for this work is to reduce Lit testing time by replacing tool invocations with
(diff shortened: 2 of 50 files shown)


boomanaiden154 (commented):
I'm assuming that you're intending to use this to disable some tests before switching everything on. Do you have examples of what sort of things require disablement?

BStott6:
> I'm assuming that you're intending to use this to disable some tests before switching everything on. Do you have examples of what sort of things require disablement?

Yes, I will soon post a PR adding the disable directive to various tests. There are tests that become inconsistent with daemonization due to static state in LLVM, for example `llvm/test/Transforms/FunctionImport/funcimport_cutoff.ll` because its output is affected by the static local variable `ImportCount` in `ModuleImportsManager::computeImportForFunction`.. There are also cases where I think daemonized testing doesn't make sense: for the tools tests, I think the tests checking the interface of a tool (rather than using the tool as a means to test some part of LLVM, like opt in the Transforms tests) should run that tool as a regular process, to ensure it is still tested as a regular process. Also, for the `llvm/test/Support/DaemonDriver` tests, I think it makes sense to disable daemonization for FileCheck, as otherwise we are testing the daemon driver using the daemon driver, which is not helpful.

PR #197975, opened 2026-05-15
[clang-sycl-linker] Report unknown command-line options as errors

YuriPlyakhin:
Pass OPT_UNKNOWN (not OPT_INVALID) to OptTable::parseArgs so the error callback fires for unrecognized flags. With OPT_INVALID the filter in parseArgs matches no Args, leaving bogus options silently dropped.

Add a lit test that exercises this path.

Drive-by cleanups in the same area:
- Use the single-argument createStringError() overload where the error code was redundantly inconvertibleErrorCode().
- Drop redundant single-quote escapes and std::string() cast in the device-library-not-found message; use Twine concat directly.
- Replace the dead "Arch must be specified" branch in runAOTCompileIntelGPU with an assert documenting the invariant enforced by the caller (IsIntelGPUOffloadArch implies non-empty arch).
- Replace the unreachable "Unsupported arch" fallthrough in runAOTCompile with llvm_unreachable; the dispatch only enters this function under IsIntelOffloadArch.
- Drop the unused [&] capture on the parseArgs error lambda.

Co-Authored-By: Claude

### DIFF
--- a/clang/test/Tooling/clang-sycl-linker.ll
+++ b/clang/test/Tooling/clang-sycl-linker.ll
@@ -73,6 +73,11 @@
 ; RUN:   | FileCheck %s --check-prefix=NOTARGET
 ; NOTARGET: Target triple must be specified
 ;
+; Check parser error reporting for unknown options.
+; RUN: not clang-sycl-linker --dry-run --not-a-real-flag -triple=spirv64 %t/input1.bc -o a.out 2>&1 \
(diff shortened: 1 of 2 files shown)


YuriPlyakhin:
@maksimsab , fyi

maksimsab:
Good change.

sarnex (approved):

PR #197976, opened 2026-05-15
[ARM][NFC] Test that debug insns are not passed to stepBackward

LU-JOHN:
Test that debug insns are not passed to stepBackward.  This testcase shows the necessity of https://github.com/llvm/llvm-project/pull/197769.


davemgreen (approved):
Brilliant, thanks. LGTM

PR #197979, opened 2026-05-15
[clang][NFC] Disentangle `MacroState` from `Preprocessor`

davidstone:
`Preprocessor` contains a map of `IdentifierInfo` to `MacroState`; this is reasonable -- the macro state is part of the overall preprocessor state. `MacroState` depends on `Preprocessor`; this is less reasonable and means that to understand `MacroState` you must first understand `Preprocessor`, but to understand `Preprocessor` you must first understand `MacroState`.

Move `MacroState` to its own file. This pulls along `FullModuleMacroInfo` since a `MacroState` is either a `MacroDirective` or a `FullModuleMacroInfo`.

Replace all `Preprocessor` parameters with only the components of `Preprocessor` that we actually need. This gives us a simpler layering where `Preprocessor` code knows about `MacroState` but `MacroState` code does not know about `Preprocessor`.

`updateModuleMacroInfo` used to be a private member function of `Preprocessor`, but its purpose is to manage `MacroState`. Instead have `MacroState` use it as an implementation detail function.

Add a new type, `PPReferences`, to bundle up references to state that `MacroState` needs as read-only other than the `IdentifierInfo` (since that's conceptually a different kind of thing).

This change also required changing `MacroInfo::isIdenticalTo` to accept the useful components of `Preprocessor` rather than the whole `Preprocessor`.

### DIFF
--- a/clang/include/clang/Lex/MacroState.h
+++ b/clang/include/clang/Lex/MacroState.h
@@ -1,0 +1,155 @@
+//===- MacroState.h - C Language Family Macro State  *- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_CLANG_LEX_MACROSTATE_H
+#define LLVM_CLANG_LEX_MACROSTATE_H
+
+#include "clang/Lex/MacroInfo.h"
+#include "llvm/ADT/PointerUnion.h"
+#include "llvm/ADT/TinyPtrVector.h"
+#include "llvm/Support/Allocator.h"
+#include "llvm/Support/Casting.h"
+
+namespace clang {
+
+class ExternalPreprocessorSource;
+class IdentifierInfo;
+class LangOptions;
+class ModuleMacro;
+class SourceManager;
+class VisibleModuleSet;
+
+using LeafModuleMacrosMap =
+    llvm::DenseMap<const IdentifierInfo *, llvm::TinyPtrVector<ModuleMacro *>>;
+
+struct PPReferences {
+  const VisibleModuleSet &VisibleModules;
+  const LeafModuleMacrosMap &LeafModuleMacros;
+  const SourceManager &SourceMgr;
+  const LangOptions &LangOpts;
+};
+
+/// Information about a name that has been used to define a module macro.
+struct FullModuleMacroInfo {
+  /// The most recent macro directive for this identifier.
+  MacroDirective *MD;
+
+  /// The active module macros for this identifier.
+  llvm::TinyPtrVector<ModuleMacro *> ActiveModuleMacros;
+
+  /// The generation number at which we last updated ActiveModuleMacros.
+  /// \see Preprocessor::VisibleModules.
+  unsigned ActiveModuleMacrosGeneration = 0;
+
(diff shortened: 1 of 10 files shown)


davidstone:
Could I get someone to review this?

clang/include/clang/Lex/MacroState.h
```diff
@@ -0,0 +1,156 @@
...
+    // FIXME: Incorporate module macros into the result of this.
+    if (auto *Latest = getLatest())
+      return Latest->findDirectiveAtLoc(Loc, SourceMgr);
+    return {};
+  }
+
+  void overrideActiveModuleMacros(ExternalPreprocessorSource &ExternalSource,
+                                  llvm::BumpPtrAllocator &Allocator,
+                                  const IdentifierInfo &II,
+                                  PPReferences PPRefs) const {
```
cyndyishida, line 121: Should `PPRefs` be passed by reference across these functions?

clang/include/clang/Lex/MacroState.h
```diff
@@ -0,0 +1,156 @@
+//===- MacroState.h - C Language Family Macro State  *- C++ -*-===//
```
cyndyishida, line 1: nit: It's standard to make the character count (80, IIRC) consistent for these markers.

clang/lib/Lex/MacroState.cpp
```diff
@@ -0,0 +1,120 @@
+#include "clang/Lex/MacroState.h"
```
cyndyishida, line 1: This is missing a licensing label.

clang/include/clang/Lex/Preprocessor.h
```diff
@@ -1216,6 +1085,14 @@ class Preprocessor {
 
   void updateOutOfDateIdentifier(const IdentifierInfo &II) const;
 
+  ModuleMacroInfo getModuleInfo(const MacroState &MS,
+                                const IdentifierInfo &II) {
+    return MS.getModuleInfo(*ExternalSource, BP, II,
```
cyndyishida, line 1090: I think this could potentially introduce UB since this is passed by reference moving forward, or, at a minimum, we lose the assert check. `ExternalSource` is initialized as `nullptr`and was later checked in https://github.com/llvm/llvm-project/blob/main/clang/lib/Lex/Preprocessor.cpp#L822 through `getFullModuleInfo`, but that chain is missed when hoisting it out of `Preprocessor`.

cyndyishida (commented):
This change looks reasonable to me, but I am curious if this is a part of a greater functional effort/change?

clang/lib/Lex/MacroInfo.cpp
```diff
@@ -136,8 +138,11 @@ bool MacroInfo::isIdenticalTo(const MacroInfo &Other, Preprocessor &PP,
       continue;
     }
 
+    auto getSpelling = [&](const Token &token) {
+      return Lexer::getSpelling(token, SourceMgr, LangOpts);
+    };
     // Otherwise, check the spelling.
-    if (PP.getSpelling(A) != PP.getSpelling(B))
+    if (getSpelling(A) != getSpelling(B))
       return false;
```
jansvoboda11, line 146:
```suggestion
    auto GetSpelling = [&](const Token &token) {
      return Lexer::getSpelling(token, SourceMgr, LangOpts);
    };
    // Otherwise, check the spelling.
    if (GetSpelling(A) != GetSpelling(B))
      return false;
```

jansvoboda11 (approved):
LGTM.

cor3ntin:
@davidstone There are still some unaddressed comments

PR #197980, opened 2026-05-15
[lldb] Fix data race on shared global ProcessProperties callback

JDevlieghere:
ProcessProperties::ProcessProperties was installing a per-process value-changed callback on ePropertyDisableLangRuntimeUnwindPlans. The property was declared Global in TargetProperties.td, so OptionValueProperties::CreateLocalCopy shared the underlying OptionValue across every ProcessProperties. Every Process constructor therefore wrote into the same std::function slot, racing with concurrent constructors and silently clobbering any earlier Process's callback.

### DIFF
--- a/lldb/source/Target/TargetProperties.td
+++ b/lldb/source/Target/TargetProperties.td
@@ -263,7 +263,6 @@ let Definition = "process", Path = "target.process" in {
     DefaultFalse,
     Desc<"If true, stop when a shared library is loaded or unloaded.">;
   def DisableLangRuntimeUnwindPlans: Property<"disable-language-runtime-unwindplans", "Boolean">,
-    Global,
     DefaultFalse,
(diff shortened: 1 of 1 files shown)


jimingham:
Do you know why this is a global property?  It seems like I could have two targets with processes, one to a system where the language runtime unwind plans are doing what I want and another where it doesn't, so I'd want to set it differently on the different targets.

But if it makes sense for this to be a global property, then this looks correct.

JDevlieghere:
I'm happy to make this a non-global property. I assumed it was intentional (and tried to keep the intent of the current code) but happy to change it.

JDevlieghere:
@felipepiovezan Did you add this? Was making that global intentional?

felipepiovezan:
> @felipepiovezan Did you add this? Was making that global intentional?

no, it wasn't. I probably didn't think too hard about that.

felipepiovezan (approved):
thanks for catching this!

PR #197981, opened 2026-05-15
host-byteorder-little-endian derictaive to typedef-different-types.test

madanial0:
The test case introduced in https://github.com/llvm/llvm-project/pull/195749/changes#diff-d9b34bdc428ca5518742902b1226065daa6a89d3fe056e10bc29615b57aa2deb fails on AIX (https://lab.llvm.org/staging/#/builders/231/builds/610/steps/6/logs/FAIL__LLVM__typedef-different-types_test) due to AIX being big endian, add requires directive similar to https://github.com/llvm/llvm-project/issues/190481


daltenty (approved):
LGTM (if I understand right we are doing this to workaround a obj2yaml / yaml2obj byteorder bug)

PR #197983, opened 2026-05-15
[SimplifyCFG] Exempt same-address store from body cap in speculativelyExecuteBB

### LINKED ISSUE #193027: `*a &= ~0x40; if (b) a |= 0x40` is not being ifcvt when join block has extra predecessors
Take:

```c
void f(int *a, int b, int c)
{
  if (c) {
    *a &= ~0x4;
    if (b)
      *a |= 0x4;
  }
}

void g(int *a, int b, int c)
{
  if (c) {
    *a &= ~0x4;
    if (b)
      *a |= 0x4;
   asm("":::"memory");
  }
}
```

`g` produces (for aarch64):

```asm
        cbz     w2, .LBB1_2
        ldr     w8, [x0]
        cmp     w1, #0
        mov     w9, #4
        csel    w9, wzr, w9, eq
        and     w8, w8, #0xfffffffb
        orr     w8, w8, w9
        str     w8, [x0]
.LBB1_2:
        ret
```

While `f` produces:

```asm
        cbz     w2, .LBB0_3
        ldr     w8, [x0]
        and     w9, w8, #0xfffffffb
        str     w9, [x0]
        cbz     w1, .LBB0_3
        orr     w8, w8, #0x4
        str     w8, [x0]
.LBB0_3:
        ret
```

Which shows the ifcvt for `f` depends on the join block not having extra predecessors.

as4230:
speculativelyExecuteBB has logic to hoist a conditional same-address store via isSafeToSpeculateStore but its 1 instruction body cap fires before matched-store recognition. The pattern from #193027 (or + same-address store) is 2 instructions and bails before the hoist path runs.

This patch moves the matched-store detection ahead of the count gate and exempts the store itself from the count. The body op still counts so the existing cap of 1 still applies to non-matched-store cases. This mirrors the exemption already in mergeConditionalStoreToAddress which treats matched stores as free in its cost budget.

From 193027:
```
void f(int *a, int b, int c) {
  if (c) {
    *a &= ~0x4;
    if (b)
      *a |= 0x4;
  }
}
```
The inner if survives because the join block has extra predecessors so mergeConditionalStoreToAddress does not apply and speculativelyExecuteBB gives up on the two instruction body. 
Before:
```
if.then:
  %0 = load i32, ptr %a, align 4
  %and = and i32 %0, -5
  store i32 %and, ptr %a, align 4
  %tobool1.not = icmp eq i32 %b, 0
  br i1 %tobool1.not, label %if.end3, label %if.then2
if.then2:
  %or = or i32 %0, 4
  store i32 %or, ptr %a, align 4
  br label %if.end3
if.end3:                              ; preds = %if.then, %if.then2, %entry
  ret void
```
With this patch, one store and no conditional branch:
```
if.then:
  %0 = load i32, ptr %a, align 4
  %and = and i32 %0, -5
  %tobool1.not = icmp eq i32 %b, 0
  %masksel = select i1 %tobool1.not, i32 0, i32 4
  %spec.store.select = or disjoint i32 %and, %masksel
  store i32 %spec.store.select, ptr %a, align 4
  br label %if.end3
```

Closes #193027.

### DIFF
--- a/llvm/lib/Transforms/Utils/SimplifyCFG.cpp
+++ b/llvm/lib/Transforms/Utils/SimplifyCFG.cpp
@@ -1757,7 +1757,7 @@ static void hoistConditionalLoadsStores(
   Value *MaskFalse = nullptr;
   Value *MaskTrue = nullptr;
   if (Invert.has_value()) {
-    IRBuilder<> Builder(Sel ? Sel : SpeculatedConditionalLoadsStores.back());
+    IRBuilder<> Builder(SpeculatedConditionalLoadsStores.back());
     Mask = Builder.CreateBitCast(
         *Invert ? Builder.CreateXor(Cond, ConstantInt::getTrue(Context)) : Cond,
         VCondTy);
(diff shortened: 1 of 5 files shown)


as4230:
cc @nikic @dtcxzyw @antoniofrighetto @RKSimon

as4230:
@dtcxzyw Rebased and restructured into precommitted tests then the fix, and added tests for the cap still applying with two body ops and for a phi at the join. The description now has the example from the issue before vs after. Would appreciate a review when you get a chance.

Andarwinux:
ping

PR #197985, opened 2026-05-15
[libc] Fix shadowing in printf

michaelrj-google:
The 320 bit float converter defined StorageType and DECIMAL_POINT
outside of its functions. This caused issues with other definitions of
the same variables after #197516.


vonosmas (approved):

nico:
Thank you!

PR #191824, opened 2026-04-13
[Extractor] Use function return for the one and only output

scui-ibm:
Currently code extractor uses parameters to pass outputs.  Alloca/store/load instructions are used to get the output value in the parent functions. 

When there is only one output from the extracted code (this is one of the most common cases), using the function return for the only one output can facilitate the other transformations (eg, tail call opt). This is to modify the code for the extracted function to return the output value if there is only one output for the extracted region.

PR #197986, opened 2026-05-15
[CodeExtractor] fix use list iterator invalidation

mustartt:
Fix crash in HotColdSplit that uses CodeExtractor to outline cold functions from https://github.com/llvm/llvm-project/pull/191824.

When `CodeExtractor::insertReplacerCall` replaces the outlined function return value, calling `replaceUsesOfWith` invalidates the uses iterator causing the loop exit early without visiting all of the original users of `FuncRetVal` causing:
```
Referring to an instruction in another function!
  %s.sroa.0.0 = phi ptr [ %call.i, %codeRepl ], [ undef, %entry ]
LLVM ERROR: Broken module found, compilation aborted!
```

Reproducer: https://godbolt.org/z/G5qv35nnq


mustartt:
I can't find a good way to isolate the test to only CodeExtractor, but it seems using HotColdSplit is the easiest.

scui-ibm (approved):
LGTM, Thanks Henry!

mustartt:
Hi Shimin, thanks of for taking a look.

llvm/test/Transforms/HotColdSplit/issue-197982.ll
```diff
@@ -0,0 +1,132 @@
+; NOTE: Assertions have been autogenerated by utils/update_test_checks.py UTC_ARGS: --version 6
+; RUN: opt -S -passes=hotcoldsplit < %s | FileCheck %s
+
+; Reproducer for https://github.com/llvm/llvm-project/issues/197982
+
+target datalayout = "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128"
+target triple = "x86_64-apple-macosx14.0.0"
```
fhahn, line 7: does the test depend on the target triple/data layout? If so, it will fail when X86 is not enabled.

llvm/test/Transforms/HotColdSplit/issue-197982.ll
```diff
@@ -0,0 +1,132 @@
...
+define noalias noundef ptr @assign(ptr captures(none) %this) local_unnamed_addr #0 {
+; CHECK-LABEL: define noalias noundef ptr @assign(
+; CHECK-SAME: ptr captures(none) [[THIS:%.*]]) local_unnamed_addr #[[ATTR0]] {
+; CHECK-NEXT:    [[CALL:%.*]] = load volatile ptr, ptr null, align 4294967296
+; CHECK-NEXT:    [[OLD_I:%.*]] = load ptr, ptr [[THIS]], align 8
+; CHECK-NEXT:    store ptr [[CALL]], ptr [[THIS]], align 8
+; CHECK-NEXT:    store volatile i32 0, ptr [[OLD_I]], align 4
+; CHECK-NEXT:    ret ptr null
+;
+  %call = load volatile ptr, ptr null, align 4294967296
```
fhahn, line 32: would be good o avoid immediate UB in test

llvm/test/Transforms/HotColdSplit/issue-197982.ll
```diff
@@ -0,0 +1,132 @@
...
+  %old.i = load ptr, ptr %this, align 8
+  store ptr %call, ptr %this, align 8
+  store volatile i32 0, ptr %old.i, align 4
+  ret ptr null
+}
+
+define void @ctor(ptr writeonly captures(none) initializes((0, 8), (16, 24)) %this) local_unnamed_addr !prof !14 {
```
fhahn, line 39: are all those attributes needed

llvm/test/Transforms/HotColdSplit/issue-197982.ll
mustartt: Let me remove the target triple. Putting up a patch now

mustartt: Thanks for taking a look. https://github.com/llvm/llvm-project/pull/198011

PR #197989, opened 2026-05-15
[AArch64] Do not pass debug insn to liveness analysis

davemgreen:
This time in the MachineSMEABIPass, where we can skip any debug instructions. See #193104.


LU-JOHN (approved):
LGTM

MacDue (commented):
Thanks! There's another case in this pass that's a little harder to hit. I've made a follow-up PR to handle it: https://github.com/llvm/llvm-project/pull/198010

PR #197991, opened 2026-05-15
[AMDGPU] Add HasWave(32|64) features

mariusz-sikora-at-amd:
GFX1250 supports only Wave32, but this was previously inferred indirectly from the GFX1250Insts feature. This breaks for targets like GFX13, which also have the GFX1250 feature but support both Wave32 and Wave64.


shiltian (commented):
I'm not sure if I like this approach. IMHO a better approach would be to say, what target **supports** wave64. That does require to add the feature to more targets. We don't need for wave32, because all targets support this.

mariusz-sikora-at-amd:
> I'm not sure if I like this approach. IMHO a better approach would be to say, what target **supports** wave64. That does require to add the feature to more targets. We don't need for wave32, because all targets support this.

I can go in this way also. I'm open to any suggestions. If I recall correctly some old gfx (pre-gfx10) supported only wave64.

arsenm:
> > I'm not sure if I like this approach. IMHO a better approach would be to say, what target **supports** wave64. That does require to add the feature to more targets. We don't need for wave32, because all targets support this.
> 
> I can go in this way also. I'm open to any suggestions. If I recall correctly some old gfx (pre-gfx10) supported only wave64.

Yes, wave64 was the only way for gfx6-9

shiltian:
Then it sounds more appealing to do two features to model this correct.

mariusz-sikora-at-amd:
> Then it sounds more appealing to do two features to model this correct.

in the convention like here? `Wave32Only` and `Wave64Only`?

llvm/lib/Target/AMDGPU/Utils/AMDGPUBaseInfo.h
```diff
@@ -1603,10 +1603,6 @@ bool hasMAIInsts(const MCSubtargetInfo &STI);
 bool hasVOPD(const MCSubtargetInfo &STI);
 bool hasDPPSrc1SGPR(const MCSubtargetInfo &STI);
 
-inline bool supportsWave32(const MCSubtargetInfo &STI) {
-  return AMDGPU::isGFX10Plus(STI) && !AMDGPU::isGFX1250(STI);
```
mariusz-sikora-at-amd, line 1607: Why we added `&& !AMDGPU::isGFX1250(STI)` here, since gfx1250 supports only Wave32?

shiltian:
> > Then it sounds more appealing to do two features to model this correct.
> 
> in the convention like here? `Wave32Only` and `Wave64Only`?

I was thinking about something like `HasWave32` and `HasWave64`, and adding to the targets accordingly. I think this is the canonical way to do it.

llvm/lib/Target/AMDGPU/AMDGPU.td
```diff
@@ -1371,6 +1371,14 @@ defm Clusters : AMDGPUSubtargetFeature<"clusters",
   /*GenPredicate=*/0
 >;
 
+defm HasWave32 : AMDGPUSubtargetFeature<"has-wave32",
+  "Target supports Wave32."
```
arsenm:
```suggestion
  "Target supports +wavefrontsize32."
```

llvm/lib/Target/AMDGPU/AMDGPU.td
```diff
@@ -1371,6 +1371,14 @@ defm Clusters : AMDGPUSubtargetFeature<"clusters",
...
+>;
+
+defm HasWave64 : AMDGPUSubtargetFeature<"has-wave64",
+  "Target supports Wave64."
+>;
```
arsenm, line 1380:
```suggestion
defm HasWave64 : AMDGPUSubtargetFeature<"has-wave64",
  "Target supports +wavefrontsize64"
>;
```

llvm/lib/Target/AMDGPU/AMDGPU.td
arsenm: Probably should phrase this in terms of the feature name. Can you also add this to the implies field of the respective wavefrontsize features

llvm/lib/Target/AMDGPU/Utils/AMDGPUBaseInfo.h
rampitec: Let me guess because it is not an option on gfx1250, it is always so. So it does not mean it does not support wave32, but it does not support wave32 as togglable option.

llvm/lib/Target/AMDGPU/AMDGPURemoveIncompatibleFunctions.cpp
```diff
@@ -199,12 +199,16 @@ bool AMDGPURemoveIncompatibleFunctions::checkFunction(Function &F) {
...
-  if (!ST->supportsWave32() && ST->hasFeature(AMDGPU::FeatureWavefrontSize32)) {
+  bool HasWave64Only = ST->hasFeature(AMDGPU::FeatureHasWave64) &&
+                       !ST->hasFeature(AMDGPU::FeatureHasWave32);
+  if (HasWave64Only && ST->hasFeature(AMDGPU::FeatureWavefrontSize32)) {
     reportFunctionRemoved(F, AMDGPU::FeatureWavefrontSize32);
     return true;
   }
   // gfx125x only support FeatureWavefrontSize32.
-  if (!ST->supportsWave64() && ST->hasFeature(AMDGPU::FeatureWavefrontSize64)) {
+  bool HasWave32Only = ST->hasFeature(AMDGPU::FeatureHasWave32) &&
```
shiltian: nit: if `HasWave32Only` and `HasWave64Only` are used in more places, we can add two helper functions in `llvm/lib/Target/AMDGPU/GCNSubtarget.h`.

llvm/lib/Target/AMDGPU/AMDGPU.td
mariusz-sikora-at-amd: > Can you also add this to the implies field of the respective wavefrontsize features

You want me to add `HasWave32` to implies field for `FeatureWavefrontSize32`?

The implies field is not "require" field. If I add HasWave32, then every time someone enables FeatureWavefrontSize32, they will also enable HasWave32.
It also won't work as expected in the other direction (if I add FeatureWavefrontSize32 as an implies to HasWave32), because many tests rely on WavefrontSize32 being disabled, and this change would enable it.

llvm/lib/Target/AMDGPU/AMDGPURemoveIncompatibleFunctions.cpp
mariusz-sikora-at-amd: Done.

llvm/lib/Target/AMDGPU/GCNSubtarget.h
```diff
@@ -868,14 +868,20 @@ class GCNSubtarget final : public AMDGPUGenSubtargetInfo,
...
-  bool supportsWave32() const { return getGeneration() >= GFX10; }
-
-  bool supportsWave64() const { return !hasGFX1250Insts(); }
-
   bool isWave32() const { return getWavefrontSize() == 32; }
 
   bool isWave64() const { return getWavefrontSize() == 64; }
 
+  bool isWave32Only() const {
+    return hasFeature(AMDGPU::FeatureHasWave32) &&
```
shiltian, line 876: This can use the auto generated helper, though they are really weird `hasHasWave32` IIRC. That said, I think it might be a better idea to rename this to `FeatureWave32` and `FeatureWave64`. If it clashes with the existing integer based ones, we can rename the existing one.

llvm/test/MC/AMDGPU/vop3-literal.s
```diff
@@ -331,18 +334,21 @@ v_cmp_eq_i32_e64 s[10:11], 0xaf123456, 0xaf123455
 
 v_cmp_eq_u64_e64 s[10:11], 0xaf123456, v[2:3]
 // GFX10: v_cmp_eq_u64_e64 s[10:11], 0xaf123456, v[2:3] ; encoding: [0x0a,0x00,0xe2,0xd4,0xff,0x04,0x02,0x02,0x56,0x34,0x12,0xaf]
-// GFX1250: v_cmp_eq_u64_e64 s[10:11], 0xaf123456, v[2:3] ; encoding: [0x0a,0x00,0x5a,0xd4,0xff,0x04,0x02,0x02,0x56,0x34,0x12,0xaf]
-// GFX9-ERR: :[[@LINE-3]]:28: error: literal operands are not supported
+// GFX1250-ASM: v_cmp_eq_u64_e64 s[10:11], 0xaf123456, v[2:3] ; encoding: [0x0a,0x00,0x5a,0xd4,0xff,0x04,0x02,0x02,0x56,0x34,0x12,0xaf]
+// GFX1250-DIS: v_cmp_eq_u64_e64 s10, 0xaf123456, v[2:3] ; encoding: [0x0a,0x00,0x5a,0xd4,0xff,0x04,0x02,0x02,0x56,0x34,0x12,0xaf]
```
shiltian, line 338: The DIS is off here?

PR #197992, opened 2026-05-15
[clang][CUDA] Add new new CUDA and PTX versions

Artem-B:
PTX 9.1 and 9.2 already exists in LLVM, so this change just plumbs these versions into clang to allow using newer instructions when we're compiling with cuda-13.x


Artem-B:
@jhuber6 PTAL. More mechanical changes to deal with recent CUDA versions.

jhuber6 (approved):

PR #197996, opened 2026-05-15
[ValueTracking][test] Update pr92217 test case to not assume sqrt is not constexpr.

rupprecht:
After ffad84af3e71d4ad330bcb363aa99c7171e33fbb, the call in `@pr92217` to `@llvm.sqrt.f32` on a specific floating point value can be inlined with the actual result instead of happening at runtime, causing the `@pr92217` method to be replaced with a single `ret i1 false`.

The test case below, `@sqrt_negative_input`, still verifies that if we have a negative value *and we don't know what it is* (i.e. it can't be constexpr'd away), then we can't make assumptions about the return value. That's all we really need. So delete the test case with a constant value and just leave the remaining ones, but rename them to preserve the link to http://llvm.org/pr92217.

### DIFF
--- a/llvm/test/Transforms/InstCombine/known-bits.ll
+++ b/llvm/test/Transforms/InstCombine/known-bits.ll
@@ -1785,14 +1785,14 @@ define i32 @test_none(float nofpclass(all) %x) {
 
 ; We cannot make assumptions about the sign of result of sqrt
 ; when the input is a negative value (except for -0).
-define i1 @pr92217() {
+define i1 @pr92217(float nofpclass(nan zero pnorm psub pinf) %a) {
 ; CHECK-LABEL: @pr92217(
-; CHECK-NEXT:    [[X:%.*]] = call float @llvm.sqrt.f32(float f0xF6F5F4F3)
+; CHECK-NEXT:    [[X:%.*]] = call float @llvm.sqrt.f32(float [[A:%.*]])
 ; CHECK-NEXT:    [[Y:%.*]] = bitcast float [[X]] to i32
 ; CHECK-NEXT:    [[CMP:%.*]] = icmp slt i32 [[Y]], 0
 ; CHECK-NEXT:    ret i1 [[CMP]]
 ;
-  %x = call float @llvm.sqrt.f32(float 0xC6DEBE9E60000000)
+  %x = call float @llvm.sqrt.f32(float %a)
   %y = bitcast float %x to i32
   %cmp = icmp slt i32 %y, 0
   ret i1 %cmp
--- a/llvm/test/Transforms/InstCombine/known-bits.ll
+++ b/llvm/test/Transforms/InstCombine/known-bits.ll
@@ -1785,8 +1785,8 @@ define i32 @test_none(float nofpclass(all) %x) {
 
 ; We cannot make assumptions about the sign of result of sqrt
 ; when the input is a negative value (except for -0).
-define i1 @pr92217(float nofpclass(nan zero pnorm psub pinf) %a) {
-; CHECK-LABEL: @pr92217(
+define i1 @pr92217_sqrt_negative_input(float nofpclass(nan zero pnorm psub pinf) %a) {
+; CHECK-LABEL: @pr92217_sqrt_negative_input(
 ; CHECK-NEXT:    [[X:%.*]] = call float @llvm.sqrt.f32(float [[A:%.*]])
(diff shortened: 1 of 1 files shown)


arsenm (commented):
I don't follow. A libc change should have no bearing on an instcombine test

rupprecht:
To be clear this only fails when rebuilding llvm libc and using it in the host toolchain. When using a host toolchain w/ an older llvm libc, it does not fail.

I'm guessing the link between libc and instcombine is due to https://discourse.llvm.org/t/rfc-make-clang-builtin-math-functions-constexpr-with-llvm-libc-to-support-c-23-constexpr-math-functions, although I have not been following it closely.

arsenm:
I think the test should be kept, and deleting it is not the solution

rupprecht:
Note that this test case still exists (I renamed it in this PR from `@sqrt_negative_input`):

```ll
define i1 @pr92217_sqrt_negative_input(float nofpclass(nan zero pnorm psub pinf) %a) {
; CHECK-LABEL: @pr92217_sqrt_negative_input(
; CHECK-NEXT:    [[X:%.*]] = call float @llvm.sqrt.f32(float [[A:%.*]])
; CHECK-NEXT:    [[Y:%.*]] = bitcast float [[X]] to i32
; CHECK-NEXT:    [[CMP:%.*]] = icmp slt i32 [[Y]], 0
; CHECK-NEXT:    ret i1 [[CMP]]
;
  %x = call float @llvm.sqrt.f32(float %a)
  %y = bitcast float %x to i32
  %cmp = icmp slt i32 %y, 0
  ret i1 %cmp
}
```

i.e. there is still test coverage that: given a negative input, instcombine can't assume the sign of the return value.

Given that, can you explain more why this test case is necessary?

AFAICT, the only additional benefit this test case would have is just to verify that calling sqrt on a constant value cannot be optimized. If that is behavior we want to guarantee (though I'd wager it isn't), shouldn't it at least be a separate test case? e.g. something like this:

```ll
define f32 @do_not_optimize_sqrt() {
; CHECK-LABEL: @do_not_optimize_sqrt(
; CHECK-NEXT:    [[X:%.*]] = call float @llvm.sqrt.f32(float f0xF6F5F4F3)
; CHECK-NEXT:    ret f32 [[X]]
;
  %x = call float @llvm.sqrt.f32(float 0xC6DEBE9E60000000)
  ret f32 %x
}
```

arsenm:
> AFAICT, the only additional benefit this test case would have is just to verify that calling sqrt on a constant value cannot be optimized. 

The opposite; I think we *should* be folding this to a constant in 100% of cases

boomanaiden154 (commented):
> I'm guessing the link between libc and instcombine is due to https://discourse.llvm.org/t/rfc-make-clang-builtin-math-functions-constexpr-with-llvm-libc-to-support-c-23-constexpr-math-functions, although I have not been following it closely.

It's still weird that the test is failing anywhere. An initial implementation of that landed for `exp` (2b2a63819f9f26d661bad5c269a03077d22ff6b4), but it got reverted fairly quickly (1565f096d868f479f075fce3792db7b908cab9aa). It's unclear why we would still be running into a behavior difference now that some functions in libc are marked constexpr.

jyknight:
This was actually a detection of an llvm-libc bug. In ffad84a, the implementation of sqrt (inadvertently) stopped setting the FP invalid flag when given a negative number input. 

In LLVM's evaluation of FP math functions (in ConstantFoldFP), we give up on constant folding if any of the exception bits are set, which is the behavior we currently expect (and test) happens, here.

I'm surprised this is the only/first place that the issue was noticed!

jcranmer-intel:
> It's still weird that the test is failing anywhere. An initial implementation of that landed for `exp` ([2b2a638](https://github.com/llvm/llvm-project/commit/2b2a63819f9f26d661bad5c269a03077d22ff6b4)), but it got reverted fairly quickly ([1565f09](https://github.com/llvm/llvm-project/commit/1565f096d868f479f075fce3792db7b908cab9aa)). It's unclear why we would still be running into a behavior difference now that some functions in libc are marked constexpr.

It's failing because we're calling the actual libc function here: https://github.com/llvm/llvm-project/blob/67a902f209437b700810893664150c5816337e3f/llvm/lib/Analysis/ConstantFolding.cpp#L2263, which means it's only failing when the libc in question is llvm-libc and has nothing to do with the APFloat exp integration (except that the long-term expectation of such integration means we'd go through the llvm-libc implementation instead of the host libc implementation).

PR #197998, opened 2026-05-15
[AMDGPU] Fix v_dot4_i32_i8 alias to set neg_lo modifiers

Jonathan03ant:
Fixes issue here https://github.com/ROCm/ROCm/issues/6126

The `v_dot4_i32_i8` assembly alias was not setting the `neg_lo` modifier bits when converted to `v_dot4_i32_iu8`, which causes signed int8 operands to be treated as unsigned.

For example: `q=[1,-1,1,-1], k=[1,1,1,1]`: expected 0, returned 512. The instruction is computing `1*1 + 255*1 + 1*1 + 255*1 = 512` ; treating `-1 (0xFF)` as `255`.

On AMD GFX11+, the native `v_dot4_i32_i8` instruction doesn't exist. The hardware provides `v_dot4_i32_iu8` with `neg_lo` modifier bits to control signedness of each operand. The compiler correctly lowers `v_dot4_i32_i8` intrinsics by setting `neg_lo:[1,1,0]`, but inline assembly using the `v_dot4_i32_i8` mnemonic bypasses this lowering and goes directly to the assembler.

The assembler accepted `v_dot4_i32_i8` as a simple mnemonic alias but didn't set the neg_lo modifiers, this results in all-unsigned behavior which is wrong.

This patch adds assembly parser logic to detect when `v_dot4_i32_i8` is used and automatically inject `neg_lo=0x3` *(src0 and src1 signed, src2 unsigned)* during operand conversion.


llvm/lib/Target/AMDGPU/AsmParser/AMDGPUAsmParser.cpp
```diff
@@ -7100,6 +7101,9 @@ bool AMDGPUAsmParser::parseInstruction(ParseInstructionInfo &Info,
   // Add the instruction mnemonic
   Name = parseMnemonicSuffix(Name);
 
+  // Remember original mnemonic before alias conversion for v_dot4_i32_i8
+  IsSignedDot4Alias = (Name == "v_dot4_i32_i8");
```
arsenm: Shouldn't be doing a special case check on the name

arsenm (commented):
Missing test

llvm/lib/Target/AMDGPU/AsmParser/AMDGPUAsmParser.cpp
Jonathan03ant: @arsenm should marking the alias in TabelGen with a flag be the right approach? This way the parser can query whether it needs signed modifiers or not, instead of checking the name directly. 

I will add a test file shortly.

arsenm: Maybe? I would expect the parsing of the alias to somehow handle whatever you need

Jonathan03ant: well the challenge is that `MnemonicAlias` only does string replacement, and `InstAlias` can't inject constant operands not in the assembly. Perhaps I should extend the alias parsing mechanism.

arsenm: There's an encoding change, so there should be a separate machine instruction definition for each subtarget variant.

Jonathan03ant: makes sense, thanks!

Jonathan03ant: @arsenm Okay I ended up creating a separate machine instruction definitions for gfx13 that encode the signed dot product instructions differently than gfx9/10. 

I will modify those missing tests add another test before I submit a pull request. Thanks for the help.

llvm/lib/Target/AMDGPU/VOP3PInstructions.td
```diff
@@ -626,12 +626,11 @@ defm V_DOT8_U32_U4  : VOP3PInst<"v_dot8_u32_u4",
   VOP3P_Profile<VOP_I32_I32_I32_I32, VOP3_PACKED_NO_OPSEL>, int_amdgcn_udot8, 1>;
 } // End OtherPredicates = [HasDot7Insts]
 
-let OtherPredicates = [HasDot1Insts] in {
```
arsenm, line 631: Do these have the correct SubtargetPredicate set?

Jonathan03ant: I added a SUbtargetPredicate for ` V_DOT4_I32_IU8`, thanks again.

Jonathan03ant: @arsenm Let me know if you have additional comments you want me to address!

llvm/test/MC/AMDGPU/gfx11_asm_vop3p_alias.s
```diff
@@ -2,7 +2,7 @@
 // RUN: llvm-mc -triple=amdgcn -mcpu=gfx1100 -show-encoding %s | FileCheck -check-prefix=GFX11 %s
 
 v_dot4_i32_i8 v5, v1, v2, s3
-// GFX11: v_dot4_i32_iu8 v5, v1, v2, s3           ; encoding: [0x05,0x40,0x16,0xcc,0x01,0x05,0x0e,0x18]
+// GFX11: v_dot4_i32_i8 v5, v1, v2, s3            ; encoding: [0x05,0x40,0x16,0xcc,0x01,0x05,0x0e,0x78]
```
shiltian, line 5: Do we have disasm test for this? We will need one.

Jonathan03ant: @shiltian that's right, I added few tests, thanks for the review.

shiltian (approved):

llvm/lib/Target/AMDGPU/VOP3PInstructions.td
Jonathan03ant: > Do these have the correct SubtargetPredicate set?

@arsenm @shiltian my initial approach using `SubtargetPredicate = HasDot8Insts` broke GFX9/10 tests since those targets only have `Dot1Insts`. I added combined predicate `HasSignedDotInsts` to properly support both GFX9/10 `(HasDot1Insts)` and GFX11+ `(HasDot8Insts)`. All AMDGPU CodeGen tests now pass locally!

PR #198000, opened 2026-05-15
[CIR] Add support for partial array cleanup with ILE

andykaylor:
This adds support for partial array cleanup when an array of a destructed type is initialized, or partially initialized with an init list expression.

In the case where the array is only partially initialized with the ILE and the remainder gets implicit initialization, the cleanup for both pieces of initialization is handled by the same cleanup scope. This is a difference from classic codegen, but there is a FIXME in the classic codegen implementation suggesting exactly this change.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
@@ -1033,10 +1033,6 @@ void CIRGenFunction::emitNewArrayInitializer(
   QualType::DestructionKind dtorKind = elementType.isDestructedType();
   CleanupDeactivationScope deactivation(*this);
 
-  CharUnits elementSize = getContext().getTypeSizeInChars(elementType);
-  CharUnits elementAlign =
-      beginPtr.getAlignment().alignmentOfArrayElement(elementSize);
-
   // Attempt to perform zero-initialization using memset.
   auto tryMemsetInitialization = [&]() -> bool {
     mlir::Location loc = numElements.getLoc();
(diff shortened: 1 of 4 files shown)


clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
```diff
@@ -1035,7 +1031,11 @@ void CIRGenFunction::emitNewArrayInitializer(
   const Expr *init = e->getInitializer();
   Address endOfInit = Address::invalid();
   QualType::DestructionKind dtorKind = elementType.isDestructedType();
-  assert(!cir::MissingFeatures::cleanupDeactivationScope());
+  CleanupDeactivationScope deactivation(*this);
+
+  CharUnits elementSize = getContext().getTypeSizeInChars(elementType);
+  CharUnits elementAlign =
```
erichkeane: IS this going to be used anywhere else?  I see it is only currently referenced in the 1 place under the `dtorKind` test?  Should this just be calculated there, instead of always?

clang/lib/CIR/CodeGen/CIRGenFunction.h
```diff
@@ -1713,8 +1713,8 @@ class CIRGenFunction : public CIRGenTypeCache {
   void emitCXXAggrConstructorCall(const CXXConstructorDecl *ctor,
                                   mlir::Value numElements, Address arrayBase,
                                   const CXXConstructExpr *e,
-                                  bool newPointerIsChecked,
-                                  bool zeroInitialize);
+                                  bool newPointerIsChecked, bool zeroInitialize,
+                                  Address endOfInit = Address::invalid());
```
erichkeane: What calls do we have without this default argument?  Can we make sure we audit these to make sure they aren't incorrect?  I would prefer we find a way to make it clear in the caller that this is 'none have been emitted yet', rather than a default argument.

erichkeane (approved):
2 small concerns, else lgtm.

clang/lib/CIR/CodeGen/CIRGenFunction.h
andykaylor: This is an overloaded function. There's one version that takes an `ArrayType` argument, uses that to get the number of elements, then calls this version to emit the constructor loop for the entire array. That version will never have pre-initialized any elements or pushed the partial array cleanup on the EH stack.

That's the only other caller of this form of the function besides the one modified in this PR. I can easily make it a required parameter and make it explicitly null in the wrapper function. Would that satisfy your concerns?

clang/lib/CIR/CodeGen/CIRGenExprCXX.cpp
andykaylor: Yes, it can be. I was following classic codegen in putting it here, but classic codegen uses it in several other places as it constructs the construction loop here, whereas we put that off until a later lowering pass.

clang/lib/CIR/CodeGen/CIRGenFunction.h
erichkeane: Yeah, i think as-a-required-param would make me comfortable.  Particularly since this is an array-only overload.

erichkeane (approved):

AmrDeveloper (approved):
LGTM

PR #198005, opened 2026-05-15
[AMDGPU] Use shorter form for i16 operands

rampitec:
For 16-bit operands an inline constant is zero extended
which in particular allows to use FP constants. These
will have 16 bits of zeroes in the high half and FP16
value in the low 16 bits.

The patch changes semantics of the FP literal argument
used in i16 context in the asm parser to fp16.


rampitec:
* #198018
* #198005 👈
* `main`

<h2></h2>

arsenm:
Is this only for true16?

rampitec:
> Is this only for true16?

No, not only. You can also see that it fixed disasm roundtrip for gfx10 and older targets which did not have true16.

shiltian (approved):

rampitec:
It changes the actual codegen. I have preserved code which used to work, just added one more shrink thing, but still. Started PSDB to make sure.

llvm/lib/Target/AMDGPU/MCTargetDesc/AMDGPUMCCodeEmitter.cpp
```diff
@@ -223,7 +223,10 @@ static uint32_t getLit32Encoding(uint32_t Val, const MCSubtargetInfo &STI) {
 }
 
 static uint32_t getLit16IntEncoding(uint32_t Val, const MCSubtargetInfo &STI) {
```
jayfoad: Can't we just remove this function and always call getLit16Encoding?

rampitec: That causes massive test failures. The added `getLit16Encoding()` only handles F16 constants, not the rest of the stuff.

jayfoad: > The added `getLit16Encoding()` only handles F16 constants,

I don't understand why, it calls `getIntInlineImmEncoding` which should handle the integers?

rampitec: That is because the first thing `getLit16Encoding` does is truncate of an incoming immediate to 16 bits. This is fine if we express a literal as something like `0x3800`, but that does not work if the input value something like `0xc0800000` like here in: `v_mad_i16 v5, v1, -4.0, v3`.

rampitec: I.e., we may revise how we represent fp16 literals internally, both in asm parser and codegen (and disasm of course). But that would be a completely separate patch with mostly invisible but deeper consequences.

jayfoad: > That is because the first thing getLit16Encoding does is truncate of an incoming immediate to 16 bits.

Ugh. The truncate is horrible. Watch this:
```
llvm-mc -triple=amdgcn -mcpu=gfx1100 -show-encoding <<< "v_mad_i16 v5, v1, 1.0059814, v3"
	v_mad_i16 v5, v1, 0xc400, v3            ; encoding: [0x05,0x00,0x53,0xd6,0x01,0xef,0x0d,0x04]
```
The f32 value 1.0059814 has a bit pattern whose low 16 bits happen to be 0xc400 which matches the inline f16 value -4.0.

rampitec: Yes, I agree. Fix is a separate change though. I am not even sure in how many places we may create an immediate this way.

rampitec: Current behavior is probably even more terrible:

```
llvm-mc -triple=amdgcn -mcpu=gfx1100 -show-encoding <<< "v_mad_i16 v5, v1, 1.0059814, v3"
	v_mad_i16 v5, v1, 0xc400, v3            ; encoding: [0x05,0x00,0x53,0xd6,0x01,0xff,0x0d,0x04,0x00,0xc4,0x80,0x3f]
```

I.e, we have full literal 0x00,0xc4,0x80,0x3f, but asm prints 0xc400.

rampitec: And then this is rejected:

```
llvm-mc -triple=amdgcn -mcpu=gfx1100 -show-encoding <<< "v_mad_i16 v5, v1, 0x3f80C400, v3"
<stdin>:1:19: error: invalid operand for instruction
v_mad_i16 v5, v1, 0x3f80C400, v3
```

So the problem is only when FP literal is used in i16 instruction. Then the root cause is in the compatibility with SP3. It does not know instruction context when parsing literals so treat it is fp32. We know the context, but treating it differently will break compatibility with both SP3 and our existing versions. Honestly I am not sure anyone uses that syntax because it is unusable.

The funniest part that it will still work with inline constant: low 16 bits of the value will be the same and only low 16 will be used. It will deviate if OPSEL is used to select high 16 of a literal, but there is no reason to write such an asm, and we do not do it in codegen.

It is probably a right thing to create FP16 literal in asm parser when there is FP input (with high 16 bits zero) and then remove this masking of high 16 bits.

shiltian: > So the problem is only when FP literal is used in i16 instruction. Then the root cause is in the compatibility with SP3.

I really think the inline constant representation should be type-aware. IIRC, we basically treat any floating-point value as a double first and then truncate it as needed. Ideally, we should probably reject any type mismatch. Inline constants essentially represent bit patterns, so for integer values, we should recognize the bit patterns directly instead of converting them back and forth.

rampitec: > inline constant representation should be type-aware.

Yes, but... SP3 compatibility.

arsenm: We're already not exactly compatible with SP3. In some good ways and some bad. Is there any real case where this would matter?

rampitec: My gut feeling there shall be no real case, because current syntax is really hardly practically useful.

rampitec:
OK, I did it. FP literals for i16 are now in FP16 semantics. Child PR is merged here and defunct, otherwise there were too much noise in the tests. SP3 compatibility is broken, so the compatibility with older versions of llvm-mc.

llvm/test/CodeGen/AMDGPU/immv216.ll
```diff
@@ -577,40 +577,40 @@ define amdgpu_kernel void @add_inline_imm_64_v2f16(ptr addrspace(1) %out, <2 x h
 }
 
 ; GCN-LABEL: {{^}}mul_inline_imm_0.5_v2i16:
-; GFX9: s_movk_i32 [[K:s[0-9]+]], 0x3800
-; GFX9: v_pk_mul_lo_u16 v0, v0, [[K]] op_sel_hi:[1,0]
+; GFX9: s_mov_b32 [[K:s[0-9]+]], 0x38003800
+; GFX9: v_pk_mul_lo_u16 v0, v0, [[K]]
 
-; GFX10: v_pk_mul_lo_u16 v0, 0x3800, v0 op_sel_hi:[0,1] ; encoding: [0x{{[0-9a-f]+}},0x{{[0-9a-f]+}},0x{{[0-9a-f]+}},0x{{[0-9a-f]+}},0xff,0x{{[0-9a-f]+}},0x{{[0-9a-f]+}},0x{{[0-9a-f]+}},0x00,0x38,0x00,0x00]
+; GFX10: v_pk_mul_lo_u16 v0, 0x38003800, v0 ; encoding: [0x{{[0-9a-f]+}},0x{{[0-9a-f]+}},0x{{[0-9a-f]+}},0x{{[0-9a-f]+}},0xff,0x{{[0-9a-f]+}},0x{{[0-9a-f]+}},0x{{[0-9a-f]+}},0x00,0x38,0x00,0x38]
```
rampitec, line 583: This looks like a regression, but it is not. OPSEL cannot be used with inline constants, only with literals. If we would encode it as inline 0.5 we could not use OPSEL. Then forced literal shall be used, but we do not have this mechanism in the codegen. It is just before 0x3800 was not recognized as inline in this context, so it stayed a literal. It could be improved, which is a separate optimization, not too beneficial though.

I.e., not too beneficial is because either 0x3800 or 0x38003800 is a 32-bit literal. It is only beneficial when it is scalar, so we can replace s_mov_b32 with s_movk_i32.

rampitec: Then probably as a safety net a separate patch may enforce literal encoding if operand has non-default OPSEL. This does not really happen though in codegen.

rampitec:
Ping after the long weekend.

llvm/test/CodeGen/AMDGPU/call-argument-types.ll
```diff
@@ -2004,7 +2004,7 @@ define amdgpu_kernel void @test_call_external_void_func_f16_imm() #0 {
 ;
 ; GFX11-TRUE16-LABEL: test_call_external_void_func_f16_imm:
 ; GFX11-TRUE16:       ; %bb.0:
-; GFX11-TRUE16-NEXT:    v_mov_b16_e32 v0.l, 0x4400
+; GFX11-TRUE16-NEXT:    v_mov_b16_e32 v0.l, 4.0
```
shiltian, line 2007: Hmm, this doesn't seem like an improvement, IMHO. Maybe we want treat `bN` as `uN`?

rampitec: Why not? Before the change:

```
llvm-mc -triple=amdgcn -mcpu=gfx1100 -show-encoding <<< "v_mov_b16_e32 v0.l, 0x4400"
        v_mov_b16_e32 v0.l, 0x4400              ; encoding: [0xff,0x38,0x00,0x7e,0x00,0x44,0x00,0x00]
```

after:

```
llvm-mc -triple=amdgcn -mcpu=gfx1100 -show-encoding <<< "v_mov_b16_e32 v0.l, 4.0"
        v_mov_b16_e32 v0.l, 4.0                 ; encoding: [0xf6,0x38,0x00,0x7e]
```

4 bytes shorter.

rampitec: Moreover, after this change the use of 0x4400 also yields the same inline constant:

```
llvm-mc -triple=amdgcn -mcpu=gfx1100 -show-encoding <<< "v_mov_b16_e32 v0.l, 0x4400"
        v_mov_b16_e32 v0.l, 4.0                 ; encoding: [0xf6,0x38,0x00,0x7e]
```

I'd prefer to see it will use inline constant in the asm.

rampitec: I.e., this is exactly the point of the patch:

1) Value 0x4400 can be represented not as a literal, but as inline 4.0. The exact bit pattern created in low 16 bits is that: 0x4400.
2) It is printed as 4.0, so we know exactly that trick is happening.
3) It finally round-trips asm/disasm.

Then if you think about it, the probability that this specific number was even used is because there is math lib doing something on actual FP numbers is close to 100%.

shiltian: It's nice to make it shorter, but printing it as 4.0 is confusing, IMHO.

When I said the inline constant should be type-aware in https://github.com/llvm/llvm-project/pull/198005#discussion_r3277946033, I meant that the representation itself should also be type-aware. Inline constants represent bit patterns, but having a floating-point value printed as an integer operand looks weird to me. 4.0 is not an inline constant for a non-floating-point operand. What is actually an inline constant is the bit pattern representing 4.0 in the corresponding floating-point type associated with the operand.

rampitec: > 4.0 is not an inline constant for a non-floating-point operand.

It is. That is written in SPG.

rampitec: Then this is rather a bugfix. We were just matching wrong patterns. Compare how it works for b32:

```
llvm-mc -triple=amdgcn -mcpu=gfx1100 -show-encoding <<< 'v_mov_b32 v0, 0x3f800000'
        v_mov_b32_e32 v0, 1.0                   ; encoding: [0xf2,0x02,0x00,0x7e]
```

shiltian (approved):
What we have in this PR right now is good. The part regarding inline constant and type-aware is a separate thing.

rampitec:
I had to revert it. After some experiments I have discovered this is only gfx10+ behavior, gfx9 does not produce fp16 value. So, it is target dependent. Then at least on some targets it does not work the same way for all b16 instructions. Probably it is better not to do it.

But then we still have asm/disasm roundtrip issue.

rampitec:
And then of course we still have that, which is an outright bug:

```
llvm-mc -arch=amdgcn -mcpu=gfx1100 -show-encoding <<< 'v_add_nc_u32 v0, 0x3f800000, v0'
	v_add_nc_u32_e32 v0, 1.0, v0            ; encoding: [0xf2,0x00,0x00,0x4a]
```

This is probably how it works on gfx9. But on gfx11 for *this instruction*  it will produce 0x3c00 in the low 16 bits, which is nowhere close to 0x3f800000. And for gfx9 it will produce 0. Then if the instruction is v_cndmark_b16 or v_mov_b16 it will also produce 0 for newer ASICs.

rampitec:
Then it is not just asm bug. If you imagine this instruction has 0x3f800000 immediate encoder will now use 1.0 inline. Which will of course screw low 16 bits. That is unlikely one can write a source like this because the literal is out of range, but it is not difficult to produce llvm IR like this.

rampitec:
I.e., inline 1.0 will materialize 0x0000 on gfx8/gfx9, 0x3c00 on gfx10-gfx12 and either 0 or 0x3c00 on gfx13 depending on the instruction. Or maybe even on gfx10-12, because I didn't check all instructions and all targets. And all of that from the input 0x3f800000.

shiltian:
Why is 1.0 materialized to 0?

rampitec:
> Why is 1.0 materialized to 0?

Because it is materialized to float32 on gfx9, which low 16 are zeroes.

rampitec:
> Why is 1.0 materialized to 0?

And then it is materialized to float32 on some newer targets and b16 operand for only some instructions.

Issue #79379, opened 2024-01-24
Diagnostics for `-Wunsafe-buffer-usage` won't fire unless enabled at *end* of file


mmdriley:
Like with other diagnostics it's possible to control `-Wunsafe-buffer-usage` with `#pragma clang diagnostic` within a source file. However, there's a weird nonlocal effect: for *any* instances of the warning to fire, it seems the warning must be enabled at the point of occurrence *and* at the end of the file.

This isn't true for other warnings like `-Wsign-compare`.

You can see an example here: https://godbolt.org/z/4eGc5hdzd

Initially there are no `-Wunsafe-buffer-usage` warnings. But if you uncomment the last line of the input (`#pragma clang diagnostic warning "-Wunsafe-buffer-usage"`) then warnings that appear _earlier_ in the file will show up.

shafik:
CC @haoNoQ @AaronBallman

haoNoQ:
This is actually a somewhat fundamental problem with so-called "analysis-based warnings", i.e. the warnings that aren't emitted by the compiler as a natural byproduct of compilation, but require a separate analysis "pass" through the code in order to connect bits and pieces of code the compiler doesn't normally need to connect. The compiler doesn't want to pay the cost of the analysis if the warning is disabled anyway, but it also can't guess where exactly is the warning going to be emitted before conducting that analysis. This plays very poorly with `#pragma clang diagnostic` that can turn the warning on and off at any point. For example you can break `-Wuninitialized` this way: https://godbolt.org/z/9jPa95az4 - it has to be enabled at the beginning of the function in order for analysis to be invoked.

In case of `-Wunsafe-buffer-usage` the analysis can't even go function-by-function. In some edge cases it has to connect multiple functions together, which results in even more weirdness of this kind. That said, there's definitely some room for improvement (and your example demonstrates very well why this is important). In particular, we can and arguably should go function-by-function unless the `-fsafe-buffer-usage-suggestions` flag is also passed (which also directly tells us that the analysis is desirable so we don't have to worry about paying for what we don't need).

Fundamentally in order to get rid of these problems we should probably teach the compiler how to tell whether the warning enabled at _any_ point in a given _section_ of code. That'll require some significant rework though.

Also worth noting that you may have a better experience with enabling the warning globally and then disabling it outside of your small code section via:
```c
#pragma clang unsafe_buffer_usage begin
  // warning is turned OFF here
#pragma clang unsafe_buffer_usage end
```
[...]

PR #128603, opened 2025-02-25
[SafeBuffers] Check if unsafe-buffers* warnings are enabled earlier

rnk:
Wunsafe-buffers* warnings are implemented by traversing the entire AST at the end of the translation unit. This is expensive, so we don't want to do it unless the user explicitly asked for these diagnostics.

Ideally we want to check if the warning was enabled anywhere in the translation unit, but we do not appear to have that functionality. Users mostly use pragmas to disable warnings, not enable them, so checking at startup is a better heuristic for implementing this compile time optimization. This is at least consistent with what we do for the other existing analyzer warnings.

Fixes issue #79379

PR #127161, opened 2025-02-14
[Fix] Speedup -Wunsafe-buffer-usage when using clang modules.

matts1:
Each piece of code should have analysis run on it precisely once. However, if you build a module, and then build another module depending on it, the header file of the module you depend on will have `-Wunsafe-buffer-usage` run on it twice. This normally isn't a huge issue,  but in the case of using the standard library as a module, simply adding the line `#include <cstddef>` increases compile times by 900ms (from 100ms to 1 second) on my machine. I believe this is because the standard library has massive modules, of which only a small part is used (the AST is ~700k lines), and because if what I've been told is correct, the AST is lazily generated, and `-Wunsafe-buffer-usage` forces it to be evaluated every time.

See https://issues.chromium.org/issues/351909443 for details and benchmarks.

PR #198006, opened 2026-05-15
[-Wunsafe-buffer-usage] Move warning-only analysis back to function-based

ziqingluo-90:
Move the warning-only analysis back to the end of parsing each Decl.
The warning-only analysis no longer does any extra AST deserialization.
Pre-compiled code will only be analyzed once during its own compilation.
    
When `-fsafe-buffer-usage-suggestions` is used, the behavior is the
same as before, because it requires visibility of the whole
translation unit.
    
rdar://177185295

Also fix rdar://107480207 & rdar://176992568 for the warning-only case.


ziqingluo-90:
CC @dtarditi

ziqingluo-90:
This change is in the same spirit as [github.com/llvm/llvm-project/pull/127161](https://github.com/llvm/llvm-project/pull/127161).

@matts1, do you still need the change made in #127161? Ignoring module nodes could result in missed reports for template instantiations when the instantiation is in the main file but the template is in a module.  Here is an example:
https://github.com/llvm/llvm-project/pull/198006/changes#:~:text=main%2Dwarning@test_sub1.h.
(Assuming you do not use fix-its.)

ziqingluo-90:
Should also fix https://github.com/llvm/llvm-project/issues/79379 for warning-only cases.

ziqingluo-90:
@rnk  Please check if this is related to https://github.com/llvm/llvm-project/pull/128603

matts1:
> This change is in the same spirit as [github.com/llvm/llvm-project/pull/127161](https://github.com/llvm/llvm-project/pull/127161).
> 
> @matts1, do you still need the change made in #127161?

I suspect not, this seems like a better approach than my change.

Xazax-hun:
I believe we want to make this change broader, to apply to most warnings. But this could be done in a separate PR.

clang/lib/Sema/AnalysisBasedWarnings.cpp
```diff
@@ -3221,6 +3227,37 @@ void clang::sema::AnalysisBasedWarnings::IssueWarnings(
...
+    UnsafeBufferUsageReporter R(S, UnsafeBufferUsageShouldSuggestSuggestions);
+
+    // Perform unsafe buffer usage analysis:
+    if (!Diags.isIgnored(diag::warn_unsafe_buffer_operation,
+                         D->getBeginLoc()) ||
+        !Diags.isIgnored(diag::warn_unsafe_buffer_variable, D->getBeginLoc()) ||
+        !Diags.isIgnored(diag::warn_unsafe_buffer_usage_in_container,
+                         D->getBeginLoc()) ||
+        !Diags.isIgnored(diag::warn_unsafe_buffer_libc_call,
+                         D->getBeginLoc())) {
```
ojhunt: I think this gating logic should be moved into a separate function: `static bool shouldEmitUnsafeBufferDiagnostics(Sema&SemaRef)` or something

clang/lib/Sema/AnalysisBasedWarnings.cpp
```diff
@@ -3221,6 +3227,37 @@ void clang::sema::AnalysisBasedWarnings::IssueWarnings(
       ++NumFunctionsWithBadCFGs;
     }
   }
+
+  // UnsafeBufferUsage analysis settings.
+  bool UnsafeBufferUsageCanEmitSuggestions = S.getLangOpts().CPlusPlus20;
+  bool UnsafeBufferUsageShouldEmitSuggestions = // Should != Can.
+      UnsafeBufferUsageCanEmitSuggestions &&
+      Diags.getDiagnosticOptions().ShowSafeBufferUsageSuggestions;
```
ojhunt: `bool ShowSafeBufferUsageSuggestions = Diags.getDiagnosticOptions().ShowSafeBufferUsageSuggestions` ?

ojhunt (changes requested):
This looks fine to me, the only change I feel strong about is moving the gating logic into a helper function - it will help if more diagnostics are added in future.

ziqingluo-90:
> I believe we want to make this change broader, to apply to most warnings. But this could be done in a separate PR.

Hi @Xazax-hun, I'm not sure if I understand your point.  Only `-Wunsafe-buffer-usage` and `LifetimeSafetyTUAnalysis` are enabled at the post-TU phase.  Most diagnostic warnings are function-based, so they do not repeatedly analyze pre-compiled code.

clang/lib/Sema/AnalysisBasedWarnings.cpp
ziqingluo-90: We only show notes that encourage users to enable fix-its when C++20 is available, because the fix-its use `std::span`.

ChuanqiXu9:
Sorry if I didn't catch necessary information. I did a quick scanning, but I didn't find the logic to make sure the analysis only happen once for precompiled code.

And other two questions,

1. Does this apply for C++20 Modules?
2. It seems like we should do this for other analysis, right?

ziqingluo-90:
>Sorry if I didn't catch necessary information. I did a quick scanning, but I didn't find the logic to make sure the analysis only happen once for precompiled code.

@ChuanqiXu9  I think my misleading title caused the same confusion for you as it did for @Xazax-hun. I should have called it 'Move -Wunsafe-buffer-usage analysis back to a function-based approach.'

Most diagnostics are analyzed alongside parsing at the end of each function. Because pre-compiled code is never parsed again, these diagnostics naturally avoid being repeated. However, `-Wunsafe-buffer-usage` was a special case: developing its fix-it generation required the entire TU to be ready, which is why we moved its analysis to the post-TU phase, consequently the analysis was applied to the whole TU including PCHs.


> Does this apply for C++20 Modules?

I didn't test against C++ 20 modules. I've only tested with clang modules.  Pre-compiled code will not be repeatedly scanned. 

>It seems like we should do this for other analysis, right?

No.  This PR makes `-Wunsafe-buffer-usage` the same as other analysis.

ChuanqiXu9:
> > Sorry if I didn't catch necessary information. I did a quick scanning, but I didn't find the logic to make sure the analysis only happen once for precompiled code.
> 
> @ChuanqiXu9 I think my misleading title caused the same confusion for you as it did for @Xazax-hun. I should have called it 'Move -Wunsafe-buffer-usage analysis back to a function-based approach.'
> 
> Most diagnostics are analyzed alongside parsing at the end of each function. Because pre-compiled code is never parsed again, these diagnostics naturally avoid being repeated. However, `-Wunsafe-buffer-usage` was a special case: developing its fix-it generation required the entire TU to be ready, which is why we moved its analysis to the post-TU phase, consequently the analysis was applied to the whole TU including PCHs.

Got it thanks.

> 
> > Does this apply for C++20 Modules?
> 
> I didn't test against C++ 20 modules. I've only tested with clang modules. Pre-compiled code will not be repeatedly scanned.
> 
> > It seems like we should do this for other analysis, right?
> 
> No. This PR makes `-Wunsafe-buffer-usage` the same as other analysis.

clang/docs/ReleaseNotes.rst
```diff
@@ -360,6 +360,20 @@ Attribute Changes in Clang
...
+  headers) is no longer repeatedly analyzed, as it is analyzed during
+  its initial compilation. (Traditionally included headers are still
+  analyzed within each translation unit that includes them).  This
+  behavior matches most of other ``-W`` diagnostics.
+
+  When ``-fsafe-buffer-usage-suggestions`` is enabled, the behavior
+  remains the same as before: pre-compiled code is deserialized and
+  analyzed alongside the translation unit that uses it, because fix-it
+  suggestion analysis requires full visibility of the translation
+  unit.
```
usx95, line 375: Drive-by: Do you have any regression numbers to share caused by this mode ?
Context: I am a maintainer of lifetime safety and we basically also have the same 2 analysis modes (end-of-function and end-of-TU).

ziqingluo-90: The warning took up 12.4% of WebKit's build time, which uses unified build.

ojhunt (approved):
lgtm

clang/docs/ReleaseNotes.rst
ziqingluo-90: well, actually when unified build is off and incremental build is on, which involves a lot of PCHs, the analysis accounts for 75% of the build time.

PR #197018, opened 2026-05-11
[GlobalISel][AArch64] Add lowering for G_UMULFIX

spaits:
Extending the previous lowering added in #196757 to the unsigned version of the G_OP G_UMULFIX.
I also plan to address the _SAT versions in a third upcoming PR.

PR #196757, opened 2026-05-09
[GlobalISel][AArch64] Add lowering for G_SMULFIX

spaits:
Adding lowering for G_SMULFIX G_OP. It is needed to compile `libc/src/stdfix/expk.cpp` with `-O3`.

PR #196636, opened 2026-05-08
[RISCV][GlobalISel] Lower G_SMULFIX

spaits:
The G_SMULFIX operator was added, but no lowering was defined for it.
Needed to compile `libc/src/stdfix/expk.cpp` with GlobalISel.

PR #108848, opened 2024-09-16
[GISEL] Fix bugs and clarify spec of G_EXTRACT_SUBVECTOR

michaelmaitland:
The implementation was missing the fact that `G_EXTRACT_SUBVECTOR` destination and source vector can be different types.

Also fix a bug in the MIR builder for `G_EXTRACT_SUBVECTOR` to generate the correct opcode.

Clarify the G_EXTRACT_SUBVECTOR specification.

PR #198012, opened 2026-05-15
[GlobalISel]Allow mixing fixed and scalable vectors for G_INSERT_SUBVECTOR and G_EXTRACT_SUBVECTOR

spaits:
I would like to allow `G_INSERT_SUBVECTOR` and `G_EXTRACT_SUBVECTOR` to be used with a scalable and a fix length vector types, similar like the OP's SelectionDAG version. There was already a discussion about this topic by @topperc and @michaelmaitland and @tschuett  in  #108848 .

I need this for RISC-V vector calling convention in GISel. I would love to have the ability to be able to work with vectors in RISC-V GISel. For example in a PR of mine I was asked to think of vectors ( #196636 ), but I couldn't really do that in the RISC-V backend, so I had to move the logic I wrote in that PR for AArch64 and test with vectors there ( #196757, #197018 ). In that case I could do that since it was a generic lowering.

This patch enables us to implement the RISC-V vector calling convention in GlobalISel in a similar manner in which it implemented in SelectionDAG. (in commit https://github.com/llvm/llvm-project/commit/0c5b789c7342ee8384507c3242fc256e23248c4d)
 
In SDAG `INSERT_SUBVECTOR` and `EXTRACT_SUBVECTOR` are used to basically convert the fixed length vector into a variable length vector, to which we can choose the right physical register according to the RISC-V calling convention (since RV vectors are basically always scalable).

For this function:
```llvm
define <4 x i32> @vector_4xi32(<4 x i32> %1) {
  %res = add <4 x i32> %1, %1
  ret <4 x i32> %res
}
```
SDAG gives this DAG:
image

So if we accept this change or discuss another approach for RISC-V vector cc. I would like to start/continue the implementation of that.

### DIFF
--- a/llvm/docs/GlobalISel/GenericOpcode.rst
+++ b/llvm/docs/GlobalISel/GenericOpcode.rst
@@ -773,7 +773,7 @@ Extract a vector of destination type from the source vector. The index operand
 represents the starting index from which a subvector is extracted from
 the source vector.
 
-The index must be a constant multiple of the source vector's minimum vector
+The index must be a constant multiple of the destination vector's minimum vector
 length. If the source vector is a scalable vector, then the index is first
 scaled by the runtime scaling factor. The indices extracted from the source
 vector must be valid indices of that vector. If this condition cannot be
--- a/llvm/docs/GlobalISel/GenericOpcode.rst
+++ b/llvm/docs/GlobalISel/GenericOpcode.rst
@@ -785,7 +785,7 @@ not the other way around.
 
 .. code-block:: none
 
-  %3:_(<vscale x 4 x i64>) = G_EXTRACT_SUBVECTOR %2:_(<vscale x 8 x i64>), 2
+  %3:_(<vscale x 4 x i64>) = G_EXTRACT_SUBVECTOR %2:_(<vscale x 8 x i64>), 4
 
 G_CONCAT_VECTORS
 ^^^^^^^^^^^^^^^^
--- a/llvm/docs/GlobalISel/GenericOpcode.rst
+++ b/llvm/docs/GlobalISel/GenericOpcode.rst
@@ -759,6 +759,9 @@ the runtime scaling factor. The indices inserted in the source vector must be
 valid indices of that vector. If this condition cannot be determined statically
 but is false at runtime, then the result vector is undefined.
 
+This operation supports inserting a fixed vector into a scalable vector, but not
+the other way around.
+
 .. code-block:: none
 
   %2:_(<vscale x 4 x i64>) = G_INSERT_SUBVECTOR %0:_(<vscale x 4 x i64>), %1:_(<vscale x 2 x i64>), 0
@@ -777,7 +780,8 @@ vector must be valid indices of that vector. If this condition cannot be
 determined statically but is false at runtime, then the result vector is
 undefined.
 
-Mixing scalable vectors and fixed vectors are not allowed.
+This operation supports extracting a fixed vector from a scalable vector, but
+not the other way around.
 
 .. code-block:: none
 
--- a/llvm/lib/CodeGen/MachineVerifier.cpp
+++ b/llvm/lib/CodeGen/MachineVerifier.cpp
@@ -1923,8 +1923,8 @@ void MachineVerifier::verifyPreISelGenericInstruction(const MachineInstr *MI) {
     }
 
     uint64_t SrcMinLen = SrcTy.getElementCount().getKnownMinValue();
-    if (!IsMixedFixedFromScalable &&
-        (Idx >= SrcMinLen || Idx + DstMinLen > SrcMinLen)) {
+    if (Idx >= SrcMinLen ||
+        (!IsMixedFixedFromScalable && Idx + DstMinLen > SrcMinLen)) {
       report("Destination type and index must not cause extract to overrun the "
              "source vector",
              MI);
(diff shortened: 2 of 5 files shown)


llvm/test/MachineVerifier/test_g_extract_subvector.mir
```diff
@@ -50,10 +50,10 @@ body: |
     ; CHECK: Destination type and index must not cause extract to overrun the source vector
     %17:_(<3 x s32>) = G_EXTRACT_SUBVECTOR  %15, 3
 
-    ; CHECK: Vector types must both be fixed or both be scalable
+    ; CHECK: Cannot extract a scalable vector from a fixed length vector
     %18:_(<vscale x 2 x s32>) = G_EXTRACT_SUBVECTOR %15, 0
 
-    ; CHECK: Vector types must both be fixed or both be scalable
+    ; Extracting a fixed vector from a scalable vector is valid.
```
arsenm: This isn't really checking anything, this doesn't prove there's no error. Valid cases should go into a separate test probably

arsenm (approved):
Should remove the valid cases from the verifier tests

llvm/docs/GlobalISel/GenericOpcode.rst
c-rhodes: not related to this patch, but just noticed this example is invalid: https://godbolt.org/z/rEasehz75

llvm/lib/CodeGen/MachineVerifier.cpp
```diff
@@ -1914,7 +1923,8 @@ void MachineVerifier::verifyPreISelGenericInstruction(const MachineInstr *MI) {
     }
 
     uint64_t SrcMinLen = SrcTy.getElementCount().getKnownMinValue();
-    if (Idx >= SrcMinLen || Idx + DstMinLen > SrcMinLen) {
+    if (!IsMixedFixedFromScalable &&
+        (Idx >= SrcMinLen || Idx + DstMinLen > SrcMinLen)) {
```
c-rhodes: index is scaled by vscale so this permits out-of-bounds extracts e.g.:
```---
name: foo
body: |
  bb.0:
    %0:_(<vscale x 4 x s32>) = G_IMPLICIT_DEF
    %1:_(<2 x s32>) = G_EXTRACT_SUBVECTOR %0, 4
...
```

arsenm (approved):

llvm/test/MachineVerifier/test_g_extract_subvector.mir
```diff
@@ -50,12 +50,13 @@ body: |
...
+    ; CHECK: Cannot extract a scalable vector from a fixed length vector
     %18:_(<vscale x 2 x s32>) = G_EXTRACT_SUBVECTOR %15, 0
 
-    ; CHECK: Vector types must both be fixed or both be scalable
-    %19:_(<2 x s32>) = G_EXTRACT_SUBVECTOR %12, 0
-
     ; CHECK: Index must be a multiple of the destination vector's minimum vector length
     %20:_(<vscale x 2 x s32>) = G_EXTRACT_SUBVECTOR %12, 1
+
+    ; CHECK: Destination type and index must not cause extract to overrun the source vector
```
c-rhodes, line 59: missing a similar test for insert

llvm/test/MachineVerifier/test_g_extract_subvector.mir
```diff
@@ -50,12 +50,13 @@ body: |
...
     %17:_(<3 x s32>) = G_EXTRACT_SUBVECTOR  %15, 3
 
-    ; CHECK: Vector types must both be fixed or both be scalable
+    ; CHECK: Cannot extract a scalable vector from a fixed length vector
     %18:_(<vscale x 2 x s32>) = G_EXTRACT_SUBVECTOR %15, 0
 
-    ; CHECK: Vector types must both be fixed or both be scalable
-    %19:_(<2 x s32>) = G_EXTRACT_SUBVECTOR %12, 0
-
     ; CHECK: Index must be a multiple of the destination vector's minimum vector length
```
c-rhodes, line 56: once again not your fault but i just realised the documentation is wrong here:
https://llvm.org/docs/GlobalISel/GenericOpcode.html#g-extract-subvector
> The index must be a constant multiple of the source vector’s minimum vector length.

minimum vector length doesn't make sense to me either

spaits: It should be destination in the docs too right? Not source.

llvm/lib/CodeGen/MachineVerifier.cpp
```diff
@@ -1893,12 +1898,16 @@ void MachineVerifier::verifyPreISelGenericInstruction(const MachineInstr *MI) {
...
+      report("Cannot extract a scalable vector from a fixed length vector", MI);
       break;
     }
 
-    if (ElementCount::isKnownGT(DstTy.getElementCount(),
+    bool IsMixedFixedFromScalable =
+        DstTy.isFixedVector() && SrcTy.isScalableVector();
+
+    if (!IsMixedFixedFromScalable &&
+        ElementCount::isKnownGT(DstTy.getElementCount(),
```
c-rhodes: no need to check `!IsMixedFixedFromScalable` here it already returns false for fixed -> scalable:
https://github.com/llvm/llvm-project/blob/a88714bd5557a5df51c80f2748bea4390e0f96d5/llvm/include/llvm/Support/TypeSize.h#L223-L228

c-rhodes (approved):
just one last comment otherwise LGTM, cheers

Issue #196493, opened 2026-05-08
[X86] Regression: i128 `select` legalized to SSE chain instead of CMOV/branch


chfast:
Since 964651ad515b, an `i128 select` whose arms both `mayFoldIntoVector` is replaced with a `v4i32 vselect`. DAGCombine then folds the `vselect ..., x, 0` into `and x, sext(cond)`, producing an SSE AND/shuffle chain instead of the previous CMOV/branch form.
                                                                                                                                                                                                                                                                                                                                                                                    
```llvm
define i128 @test(i1 %c, i128 %a, ptr %m) {
    %b = load i128, ptr %m
    %sel = select i1 %c, i128 %b, i128 0
    ret i128 %sel
}
```
https://godbolt.org/z/v7j8PnGx6

```asm
test:                                   # @test
        movq    xmm0, qword ptr [rcx + 8]       # xmm0 = mem[0],zero
        movq    xmm1, qword ptr [rcx]           # xmm1 = mem[0],zero
        punpcklqdq      xmm1, xmm0              # xmm1 = xmm1[0],xmm0[0]
        and     edi, 1
        neg     edi
        movd    xmm0, edi
        pshufd  xmm0, xmm0, 0                   # xmm0 = xmm0[0,0,0,0]
        pand    xmm0, xmm1
        movq    rax, xmm0
        pshufd  xmm0, xmm0, 238                 # xmm0 = xmm0[2,3,2,3]
        movq    rdx, xmm0
        ret
```

RKSimon:
Scalarization is failing due to the multiple extractions on the AND - even though both AND operands can be freely extracted

PR #198013, opened 2026-05-15
[DAG] scalarizeExtractedBinOp - extract from non-constant one use buildvectors

RKSimon:
When attempting to scalarize a vector binop that has a single extract, we currently only fold if either of the binop's operands is a constant buildvector - but we can extract from non-constant buildvectors without increasing instruction count as long as the vector binop was the only use of the buildvector.

More yak shaving for #196493


llvm/test/CodeGen/X86/known-signbits-vector.ll
```diff
@@ -351,11 +351,10 @@ define float @signbits_ashr_sext_sextinreg_and_extract_sitofp(<2 x i64> %a0, <2
...
 ; X86:       # %bb.0:
 ; X86-NEXT:    pushl %eax
-; X86-NEXT:    vpshufd {{.*#+}} xmm0 = xmm0[1,1,3,3]
 ; X86-NEXT:    vpsrad $29, %xmm0, %xmm0
-; X86-NEXT:    vmovd {{.*#+}} xmm1 = mem[0],zero,zero,zero
-; X86-NEXT:    vpand %xmm0, %xmm1, %xmm0
-; X86-NEXT:    vcvtdq2ps %xmm0, %xmm0
+; X86-NEXT:    vpextrd $1, %xmm0, %eax
+; X86-NEXT:    andl {{[0-9]+}}(%esp), %eax
+; X86-NEXT:    vcvtsi2ss %eax, %xmm7, %xmm0
```
phoebewang, line 357: Do we increase the SIMD->GPR then GPR->SIMD pressure here?

RKSimon: Yes, although the test case is rather artificial to test signbit counts. Technically, since we're loading to x87 anyhow, I guess we should just use filds, but I'm not sure I want to go there......

phoebewang: Oh, x87.. It's fine then :)

phoebewang (approved):
LGTM.

Issue #197393, opened 2026-05-13
[Flang] Incorrect results from ASSOCIATE constructs with -O2 or higher


yus3710-fj:
## Summary

When an ASSOCIATE construct's selector is an array section with a non-constant stride in its subscript triplet, the execution result is incorrect.

### Reproducer

* test.f90

```fortran
subroutine s2(n1)
  implicit none
  integer :: n1, c(-5:-1,-4:-1)

  associate(w => c(-1:-5:-n1,-1:-4:-1))
    w(5:n1:-1,4:1:-1) = 3
  end associate
  if (any(c /= 3)) print *, 'error', c
end subroutine

call s2(1)
print *, 'pass'
end
```

* commands

```console
$ flang --version
flang version 23.0.0git (https://github.com/llvm/llvm-project.git 578ba5f3039e7a1e4e6f3a82891b3838823e37c2)
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: /path/to/llvm/build/bin
Build config: +assertions
$ flang test.f90 -O2 && ./a.out
 error 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3
 pass
```

## Non-problematic Conditions

The issue does not occur under the following conditions:
* Making the shape of the array section constant:
  ```fortran
  subroutine s2(n1)
    implicit none
    integer :: n1, c(-5:-1,-4:-1)

    associate(w => c(-1:-5:-1,-1:-4:-1))
      w(5:1:-1,4:1:-1) = 3
    end associate
    if (any(c /= 3)) print *, 'error', c
  end subroutine

  call s2(1)
  print *, 'pass'
  end
  ```
* Disabling loop unrolling.

## Other Compilers

* GFortran
  ```console
  $ gfortran --version
  GNU Fortran (GCC) 15.2.0
  Copyright (C) 2025 Free Software Foundation, Inc.
  This is free software; see the source for copying conditions.  There is NO
  warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  
  $ gfortran test.f90 -O3 && ./a.out
   pass
  ```

* Intel Fortran
  ```console
  $ ifx --version
  ifx (IFX) 2025.3.2 20260112
  Copyright (C) 1985-2026 Intel Corporation. All rights reserved.
  
  $ ifx -O3 test.f90 && ./a.out
   pass
  ```

eugeneepshteyn:
[...]

PR #198014, opened 2026-05-15
[flang] Do not set nuw  flag on mul/add/getelementptr when lowering array_coor from a box base.

valerydmit:
This change is aimed to fix an issue that can be seen with the following example, based on a test case reported in issue #197393:

subroutine s2(n1)
  implicit none
  integer :: n1, c(-5:-1)

  associate(w => c(-1:-5:-n1))
    w(n1:5:1) = 3
  end associate
  if (any(c /= 3)) print *, 'error', c
end subroutine s2
call s2(1)

The problem is that base pointer in a box can actually point at the end of array, which means that in order to access array elements offsets are going to be negative. That in turn means that nuw flag can't be set as it results in poison from that GEP.

Fixes issue #197393.

### DIFF
--- a/flang/lib/Optimizer/CodeGen/CodeGen.cpp
+++ b/flang/lib/Optimizer/CodeGen/CodeGen.cpp
@@ -2885,8 +2885,8 @@ struct XArrayCoorOpConversion
             getStrideFromBox(loc, baseBoxTyPair, operands[0], i, rewriter);
         auto sc =
             mlir::LLVM::MulOp::create(rewriter, loc, idxTy, diff, stride, nsw);
-        offset = mlir::LLVM::AddOp::create(rewriter, loc, idxTy, sc, offset,
-                                           addMulFlags);
+        offset =
+            mlir::LLVM::AddOp::create(rewriter, loc, idxTy, sc, offset, nsw);
       } else {
         // Use stride computed at last iteration.
         auto sc = mlir::LLVM::MulOp::create(rewriter, loc, idxTy, diff, prevExt,
--- a/flang/lib/Optimizer/CodeGen/CodeGen.cpp
+++ b/flang/lib/Optimizer/CodeGen/CodeGen.cpp
@@ -2883,8 +2883,8 @@ struct XArrayCoorOpConversion
         // Use stride in bytes from the descriptor.
         mlir::Value stride =
             getStrideFromBox(loc, baseBoxTyPair, operands[0], i, rewriter);
-        auto sc = mlir::LLVM::MulOp::create(rewriter, loc, idxTy, diff, stride,
-                                            addMulFlags);
+        auto sc =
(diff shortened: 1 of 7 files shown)


flang/lib/Optimizer/CodeGen/CodeGen.cpp
```diff
@@ -2909,8 +2909,9 @@ struct XArrayCoorOpConversion
       mlir::Value base =
           getBaseAddrFromBox(loc, baseBoxTyPair, operands[0], rewriter);
       llvm::SmallVector<mlir::LLVM::GEPArg> args{offset};
-      auto addr = mlir::LLVM::GEPOp::create(rewriter, loc, llvmPtrTy, byteTy,
-                                            base, args, gepFlags);
+      auto addr =
+          mlir::LLVM::GEPOp::create(rewriter, loc, llvmPtrTy, byteTy, base,
+                                    args, mlir::LLVM::GEPNoWrapFlags::nusw);
```
vzakhari, line 2914: `gepFlags` is probably unused now.

vzakhari (approved):
LGTM. Thank you, Valery!

flang/lib/Optimizer/CodeGen/CodeGen.cpp
valerydmit: Thank you for prompt review. gepFlags is still used in a couple of places: for a subcomponent address and for non-boxed case.

Stylie777:
Could this be applied conditionally if we know the offset is going to be non negative? Having this flags allows for optimisations later in the middle end for array_coor's.

flang/test/Fir/array-coor.fir
tblah: t6 is a stride loaded from the descriptor, then here we are multiplying and adding it with `nuw`. For a negative stride I think that could have the same bug as the one you described?

E.g. 2 * -4 is fine as a signed multiplication, but as an unsigned multiplication I think that wraps and so produces poison with nuw.

tblah (commented):
I tried this patch as-is against the benchmark Jack was concerned about and I think this will be okay, and even if it wasn't this looks like a real correctness bug.

Thanks for the fix!

flang/test/Fir/array-coor.fir
valerydmit: Right. Thanks for pointing that out. Fixed that too.

valerydmit:
> Could this be applied conditionally if we know the offset is going to be non negative? Having this flags allows for optimisations later in the middle end for array_coor's.

As we read stride from descriptor, I'm not sure about how we could find that we never go into negative range at compile time.

flang/test/Fir/array-coor.fir
valerydmit: I suggest similar issue exists for "add sc, offset" when rank >1 and stride is negative. So I dropped nuw for the add too.

jeanPerier (approved):
Thanks for the fix

PR #173926, opened 2025-12-29
[RISCV][MCA] Consolidate inputs for different SchedModel tests

mshockwave:
For a long time, we have been copying existing llvm-mca test files as a mean to test new scheduling model changes. This approach doesn't scale well primarily because we eventually have too many duplicated input assembly lines spread across multiple files. Also, the files you copied from might not have the best coverage. 

To address this issue, I propose to consolidate the inputs for all RISC-V scheduling model tests into a single place. More specifically, all assembly input code will now be put in `test/tools/llvm-mca/RISCV/Inputs`. Take `test/tools/llvm-mca/RISCV/Inputs/zba.s` as an example, it contains -- and only contains -- Zba instructions. To use it as an input to test `sifive-p470`'s scheduling model, we only need to create a `test/tools/llvm-mca/RISCV/SiFiveP400/zba.test` with the following `RUN` line:
```
# RUN: llvm-mca -mtriple=riscv64 -mcpu=sifive-p470 -iterations=1 -instruction-tables=full %p/../Inputs/zba.s | FileCheck %s
```
Before auto-generating its contents with the UTC script. The files in `test/tools/llvm-mca/RISCV/Inputs` are roughly organized by extensions, so scheduling models can take only the extensions they support. 

This PR is a demonstration of such idea that applies it onto SiFiveP400 and SpacemitX60. If everyone thinks it's a good idea, we can do it to more scheduling models.

------

I actually have a python script to create `*.test` files from their corresponding `*.s` files in `Inputs`. Though arguably it's not really hard to do it manually, I can also upstream it if folks find it potentially useful.

PR #198016, opened 2026-05-15
[RISCV][MCA] Use the new infrastructure for SiFive P500 and P800's tests. NFC

mshockwave:
Some tests -- mostly vector crypto -- are kept for SiFive P800.

NFC.


topperc (approved):
LGTM

boomanaiden154:
What exactly is the new infrastructure here?

mshockwave:
> What exactly is the new infrastructure here?

it's the new way of writing RISC-V MCA test introduced in #173926

camel-cdr:
Are there plans to distinguish P550 Gen1 and P550 Gen3? Which have different ISA support. The same unclear naming scheme also effects some of the other SiFive cores.

mshockwave:
> Are there plans to distinguish P550 Gen1 and P550 Gen3? Which have different ISA support. The same unclear naming scheme also effects some of the other SiFive cores.

yeah we're working on the naming. The `sifive-p570` support might come before that though.

PR #198018, opened 2026-05-15
[AMDGPU] Allow printing i16 imm as f16 inline constant

rampitec:
This allows diasm to look the same way as asm and codegen.


rampitec:
* #198018 👈
* #198005
* `main`

<h2></h2>

llvm/lib/Target/AMDGPU/MCTargetDesc/AMDGPUInstPrinter.cpp
```diff
@@ -579,6 +564,24 @@ static bool printImmediateBFloat16(uint32_t Imm, const MCSubtargetInfo &STI,
...
+                                            const MCSubtargetInfo &STI,
+                                            raw_ostream &O) {
+  int32_t SImm = static_cast<int32_t>(Imm);
+  if (isInlinableIntLiteral(SImm)) {
+    O << SImm;
+    return;
+  }
+
+  if (printImmediateFloat32(Imm, STI, O))
+    return;
```
arsenm, line 577: Why is this here?

rampitec: You can use it, only low 16 bits will be useful, or maybe high part too with opsel. That is the original code anyway.

llvm/lib/Target/AMDGPU/MCTargetDesc/AMDGPUInstPrinter.cpp
```diff
@@ -579,6 +564,24 @@ static bool printImmediateBFloat16(uint32_t Imm, const MCSubtargetInfo &STI,
   return true;
 }
 
+void AMDGPUInstPrinter::printImmediateInt16(uint32_t Imm,
```
arsenm, line 567: Why do we have a separate printImmediateInt16 and printImmediateF16 if they are going to do the same thing?

rampitec: They are not doing a same thing, `printImmediateFP16()` handles just F16 inline constants here.

arsenm: I mean AMDGPUInstPrinter::printImmediateF16 and AMDGPUInstPrinter::printImmediateInt16

rampitec: They are not the same though, Int16 case also prints Float32 stuff. For example, with Float32 in it, current patch:
`v_add_nc_i16 v5, m0, 0.5                ; encoding: [0x05,0x00,0x0d,0xd7,0x7d,0xe0,0x01,0x02]`
W/o it:
`v_add_nc_i16 v5, m0, 0x0                ; encoding: [0x05,0x00,0x0d,0xd7,0x7d,0xe0,0x01,0x02]`
I.e., it is now shrunk to 0x0 on printing.

rampitec: And then w/o Float32 stuff and w/o `& 0xffff` it is correct but not very helpful:
```
v_add_nc_i16 v5, m0, 0x3f000000         ; encoding: [0x05,0x00,0x0d,0xd7,0x7d,0xe0,0x01,0x02]
```

rampitec: I.e., I can remove FP32 handling here and in the parent PR, but that would deviate from SP3 behavior. Basically that would create a new asm syntax assumptions. Arguably this behavior would be more correct, but it will be different from both what we do now and what SP3 is doing. This patch in turn is more on a cosmetic side.

shiltian (approved):

llvm/test/CodeGen/AMDGPU/arbitrary-fp-to-float.ll
```diff
@@ -1259,7 +1259,7 @@ define bfloat @from_f8e5m2_bf16(i8 %x) {
 ; CHECK-NEXT:    v_cmp_ne_u16_e32 vcc, 0, v2
 ; CHECK-NEXT:    v_cmp_eq_u16_e64 s[4:5], 0, v0
 ; CHECK-NEXT:    v_or_b32_e32 v4, v4, v5
-; CHECK-NEXT:    v_add_u16_e32 v4, 0x3800, v4
+; CHECK-NEXT:    v_add_u16_e32 v4, 0.5, v4
```
shiltian, line 1262: Even though I understand that our inline constants represent bit patterns rather than semantic values, I think ideally this still shouldn't be allowed, because I find it very confusing.

PR #198019, opened 2026-05-15
[clangd] Fix crash when server shuts down with in-flight rename or tweak

### LINKED ISSUE #196225: [clangd] Assertion during shutdown when an in-flight rename outlives ClangdServer
Using fuzzer, it is found that clangd crashes if a rename operation is still in flight when the client immediately shuts the server down.
clangd logs LSP finished, exiting with status 0 and then aborts with:

`std::_Optional_base_impl<clang::clangd::ClangdServer, ...>::_M_get(): Assertion 'this->_M_is_engaged()' failed.`

```cpp
int abc;
```
Minimal request sequence:
```
initialize
initialized
textDocument/didOpen
textDocument/rename at abc
shutdown
exit
```
This also reproduces through the command path:
```
workspace/executeCommand with command: "clangd.applyRename"
arguments: [ RenameParams ]
```

reproducer:
```
python reproducer.py  /path/to/clangd
```

```python
#!/usr/bin/env python3
import json
import shutil
import subprocess
import sys
import tempfile
import time
from pathlib import Path
from urllib.parse import quote


SOURCE = """int abc;
"""


def file_uri(path: Path) -> str:
    return "file://" + quote(str(path))


def encode_lsp(msg: dict) -> bytes:
    body = json.dumps(msg, separators=(",", ":")).encode("utf-8")
    return f"Content-Length: {len(body)}\r\n\r\n".encode("ascii") + body


def main() -> int:
    if len(sys.argv) != 2:
        print(f"usage: {sys.argv[0]} /path/to/assert-clangd", file=sys.stderr)
        return 2
    clangd = sys.argv[1]

    workdir = Path(tempfile.mkdtemp(prefix="repro-id46-"))
    source_path = workdir / "main.cpp"
    source_path.write_text(SOURCE, encoding="utf-8")
    uri = file_uri(source_path)
    root_uri = file_uri(workdir)

    messages = [
        {
            "jsonrpc": "2.0",
            "id": 0,
            "method": "initialize",
            "params": {
                "processId": None,
                "rootUri": root_uri,
                "capabilities": {},
                "workspaceFolders": [{"uri": root_uri, "name": "workspace"}],
            },
        },
        {"jsonrpc": "2.0", "method": "initialized", "params": {}},
        {
            "jsonrpc": "2.0",
            "method": "textDocument/didOpen",
            "params": {
                "textDocument": {
                    "uri": uri,
                    "languageId": "cpp",
                    "version": 1,
                    "text": SOURCE,
                }
            },
        },
        {
            "jsonrpc": "2.0",
            "id": 1,
            "method": "textDocument/rename",
            "params": {
                "textDocument": {"uri": uri},
                "position": {"line": 0, "character": 5},
                "newName": "scape",
            },
        },
        {"jsonrpc": "2.0", "id": 999, "method": "shutdown", "params": None},
        {"jsonrpc": "2.0", "method": "exit", "params": None},
    ]

    proc = subprocess.Popen(
        [clangd, "--background-index=0", "--clang-tidy=0"],
        stdin=subprocess.PIPE,
        stdout=subprocess.DEVNULL,
        stderr=subprocess.PIPE,
        text=False,
    )

    assert proc.stdin is not None
    for i, msg in enumerate(messages):
        proc.stdin.write(encode_lsp(msg))
        proc.stdin.flush()
        if i < 3:
            time.sleep(0.01)
    try:
        proc.stdin.close()
    except Exception:
        pass

    time.sleep(1.2)
    if proc.poll() is None:
        proc.kill()

    stderr = proc.stderr.read().decode("utf-8", "replace") if proc.stderr else ""
    rc = proc.wait(timeout=2)
    reproduced = (
        "std::_Optional_base_impl" in stderr
        or "_M_is_engaged()' failed" in stderr
    )
    print(
        json.dumps(
            {
                "case": "id46",
                "clangd": clangd,
                "exit_code": rc,
                "signal": -rc if rc < 0 else None,
                "reproduced": reproduced,
                "workdir": str(workdir),
                "source_path": str(source_path),
            },
            ensure_ascii=False,
            indent=2,
        )
    )
    if stderr:
        print("\n=== clangd stderr ===")
        print(stderr, end="" if stderr.endswith("\n") else "\n")
    shutil.rmtree(workdir, ignore_errors=True)
    return 0 if reproduced else 1


if __name__ == "__main__":
    raise SystemExit(main())

```

hbatagelo:
Fixes #196225.

Root cause: When clangd shuts down, `ClangdLSPServer`'s destructor calls `Server.reset()` to destroy the `ClangdServer` object. `ClangdServer` then destroys the work scheduler, which blocks and waits for in-flight tasks to finish. If a rename task is in flight, the callback will try to access the now-reset `ClangdLSPServer::Server` in https://github.com/llvm/llvm-project/blob/4f9a7d09f4760ac9c5745e8bb829366d29ff9687/clang-tools-extra/clangd/ClangdLSPServer.cpp#L916-L917

The same issue occurs with `onCommandApplyTweak`, as its callback also accesses `ClangdLSPServer::Server`: https://github.com/llvm/llvm-project/blob/4f9a7d09f4760ac9c5745e8bb829366d29ff9687/clang-tools-extra/clangd/ClangdLSPServer.cpp#L828-L829

A reproducer has been added to #196225.

This PR fixes these issues by capturing a reference to `ClangdServer` and using it in the callbacks instead of the `ClangdLSPServer::Server` optional. The reference is guaranteed to remain valid because the work scheduler syncs while the server is being destroyed.

Includes a unit test that verifies the fix for `rename`. The test for `applyTweak` was omitted because the mocked LSP client would trigger an `ADD_FAILURE()` ("Unexpected server->client call") when intercepting the outgoing `workspace/applyEdit` request. While this could be handled with `EXPECT_NONFATAL_FAILURE_ON_ALL_THREADS`, it would make the test flaky because the task could occasionally finish before the server shuts down.

### DIFF
--- a/clang-tools-extra/clangd/ClangdLSPServer.cpp
+++ b/clang-tools-extra/clangd/ClangdLSPServer.cpp
@@ -808,7 +808,7 @@ void ClangdLSPServer::onCommandApplyEdit(const WorkspaceEdit &WE,
 
 void ClangdLSPServer::onCommandApplyTweak(const TweakArgs &Args,
                                           Callback<llvm::json::Value> Reply) {
-  auto Action = [this, Reply = std::move(Reply)](
+  auto Action = [this, Reply = std::move(Reply), &ServerRef = *Server](
                     llvm::Expected<Tweak::Effect> R) mutable {
     if (!R)
       return Reply(R.takeError());
@@ -825,7 +825,7 @@ void ClangdLSPServer::onCommandApplyTweak(const TweakArgs &Args,
     if (R->ApplyEdits.empty())
       return Reply("Tweak applied.");
 
-    if (auto Err = validateEdits(*Server, R->ApplyEdits))
+    if (auto Err = validateEdits(ServerRef, R->ApplyEdits))
       return Reply(std::move(Err));
 
     WorkspaceEdit WE;
@@ -909,11 +909,11 @@ void ClangdLSPServer::onRename(const RenameParams &Params,
     return Reply(llvm::make_error<LSPError>(
         "onRename called for non-added file", ErrorCode::InvalidParams));
   Server->rename(File, Params.position, Params.newName, Opts.Rename,
-                 [File, Params, Reply = std::move(Reply),
-                  this](llvm::Expected<RenameResult> R) mutable {
+                 [File, Params, Reply = std::move(Reply), &ServerRef = *Server](
+                     llvm::Expected<RenameResult> R) mutable {
                    if (!R)
                      return Reply(R.takeError());
-                   if (auto Err = validateEdits(*Server, R->GlobalChanges))
+                   if (auto Err = validateEdits(ServerRef, R->GlobalChanges))
                      return Reply(std::move(Err));
                    WorkspaceEdit Result;
                    // FIXME: use documentChanges if SupportDocumentChanges is
--- a/clang-tools-extra/clangd/unittests/ClangdLSPServerTests.cpp
+++ b/clang-tools-extra/clangd/unittests/ClangdLSPServerTests.cpp
@@ -512,6 +512,22 @@ TEST_F(LSPTest, CompletionOutOfRangePosition) {
   ASSERT_TRUE(!!Result) << "Expected a response, not a server crash";
 }
 
+// https://github.com/llvm/llvm-project/issues/196225
+TEST_F(LSPTest, ShutdownDuringRename) {
+  Annotations Code("void ^foo();");
+  auto &Client = start();
+  Client.didOpen("foo.cpp", Code.code());
+  auto &Reply = Client.call("textDocument/rename",
+                            llvm::json::Object{
+                                {"textDocument", Client.documentID("foo.cpp")},
+                                {"position", Code.point()},
+                                {"newName", "bar"},
+                            });
+  stop();
+  auto Result = Reply.take();
+  ASSERT_TRUE(!!Result) << "Expected a response, not a server crash";
+}
+
 } // namespace
 } // namespace clangd
 } // namespace clang


hbatagelo:
@ArcsinX Could you take a look at this when you have a chance?

ArcsinX (commented):
I tried your test without the changes in ClangdLSPServer.cpp and for me the test passes. I tried with debug build and also tried to add some delays inside rename without any success to fail the test. Is it expected?

hbatagelo:
Without the changes, the test consistently fails on my end. The `ASSERT_TRUE` line is never reached because the server crashes first.

Were you able to reproduce the issue using the script from #196225? I'm hitting the `std::optional` assertion failure when running it locally. Note that I'm linking against `libstdc++`:
```
{
  "case": "id46",
  "clangd": "./build/bin/clangd",
  "exit_code": -9,
  "signal": 9,
  "reproduced": true,
  "workdir": "/tmp/repro-id46-2hay22mu",
  "source_path": "/tmp/repro-id46-2hay22mu/main.cpp"
}

=== clangd stderr ===
I[21:16:36.745] clangd version 23.0.0git (git@github.com:hbatagelo/llvm-project.git 3d3f4be0f13d69d46914c330895c9d5eff48a8d1)
I[21:16:36.745] Features: linux+debug
I[21:16:36.745] PID: 1649682
I[21:16:36.745] Working directory: /home/harlen/dev/llvm-project
I[21:16:36.745] argv[0]: ./build/bin/clangd
I[21:16:36.745] argv[1]: --background-index=0
I[21:16:36.745] argv[2]: --clang-tidy=0
I[21:16:36.745] Starting LSP over stdin/stdout
I[21:16:36.745] <-- initialize(0)
I[21:16:36.745] --> reply:initialize(0) 0 ms
I[21:16:36.750] <-- initialized
I[21:16:36.760] <-- textDocument/didOpen
I[21:16:36.763] Failed to find compilation database for /tmp/repro-id46-2hay22mu/main.cpp
I[21:16:36.763] ASTWorker building file /tmp/repro-id46-2hay22mu/main.cpp version 1 with command clangd fallback
[/tmp/repro-id46-2hay22mu]
/usr/bin/clang -resource-dir=/home/harlen/dev/llvm-project/build/lib/clang/23 -- /tmp/repro-id46-2hay22mu/main.cpp
I[21:16:36.770] <-- textDocument/rename(1)
I[21:16:36.770] <-- shutdown(999)
I[21:16:36.770] --> reply:shutdown(999) 0 ms
I[21:16:36.770] <-- exit
I[21:16:36.770] LSP finished, exiting with status 0
/usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/optional:479: _Tp &std::_Optional_base_impl<clang::clangd::ClangdServer, std::_Optional_base<clang::clangd::ClangdServer>>::_M_get() [_Tp = clang::clangd::ClangdServer, _Dp = std::_Optional_base<clang::clangd::ClangdServer>]: Assertion 'this->_M_is_engaged()' failed.
```

ArcsinX (approved):
> Without the changes, the test consistently fails on my end. The `ASSERT_TRUE` line is never reached because the server crashes first.
> 
> Were you able to reproduce the issue using the script from #196225? I'm hitting the `std::optional` assertion failure when running it locally. Note that I'm linking against `libstdc++`:

Thanks for clarification. I think the difference is that the behavior of `std::optional::reset()` regarding when it officially becomes "empty" (before or after the object destructor call) is implementation-defined. So, in your case it happens before and in my case it happens after.

LGTM

hbatagelo:
Ah, makes sense, `libstdc++` first unsets the engaged flag, then calls the destructor; `libc++` does it in the opposite order.

Thanks for the review. Could you merge the PR? I don't have commit access yet.

PR #198020, opened 2026-05-15
[MLIR][XeGPU][XeVM] Add lowering for load_nd with MX scale element type.

silee2:
HW does not directly support load_nd for MX scale element type f8E8M0FNU, thus cannot be lower to xevm 2d block load.
This special case is lowered to code sequence using llvm.load


Jianhui-Li (commented):
I'd suggest handling this case earlier in the pipeline — specifically in the peephole-optimization  pass, before subgroup-to-lane distribution.
   
  The rationale: lane-level XeGPU IR should be fully aligned with what the hardware actually supports. If we lower unsupported tile shapes only during XeGPU-to-XeVM conversion, then the lane-level IR still contains "virtual" operations that don't correspond to real HW capabilities. By materializing these into supported forms before distribution, we maintain the invariant that post-distribution IR maps cleanly onto hardware.

mlir/lib/Conversion/XeGPUToXeVM/XeGPUToXeVM.cpp
```diff
@@ -266,6 +266,124 @@ class CreateNdDescToXeVMPattern
...
+    // 1x16 or 2x16 for B Scale
+    auto tileShape = tdescTy.getShape();
+    bool isAScale = tileShape[0] == 8;
+    bool isBScale = tileShape[1] == 16;
+    if (!isAScale && !isBScale)
+      return failure();
+    Type resTy = getTypeConverter()->convertType(op.getValue().getType());
+    auto resVecTy = dyn_cast<VectorType>(resTy);
+    if (resVecTy) {
+      if (resVecTy.getRank() != 1)
```
akroviakov, line 293: Given the type converter, can it even happen?

mlir/lib/Conversion/XeGPUToXeVM/XeGPUToXeVM.cpp
```diff
@@ -266,6 +266,124 @@ class CreateNdDescToXeVMPattern
...
+        return failure();
+      if (resVecTy.getShape()[0] != 2)
```
akroviakov, line 295: So we can only load 2 e8m0 scales?
Would be nice to have a reference link, explaining this restriction.

mlir/lib/Conversion/XeGPUToXeVM/XeGPUToXeVM.cpp
```diff
@@ -266,6 +266,124 @@ class CreateNdDescToXeVMPattern
...
+      if (tileShape[1] != 1 && tileShape[1] != 2)
+        return failure();
+    }
+    if (isBScale) {
+      if (tileShape[0] != 1 && tileShape[0] != 2)
+        return failure();
+    }
+    auto mixedOffsets = op.getMixedOffsets();
+    int64_t opOffsetSize = mixedOffsets.size();
+    if (opOffsetSize != 2)
```
akroviakov, line 308:
```suggestion
    if (opOffsetSize != tdescTy.getRank())
```
To match the failure message?

akroviakov (approved):
Functionally looks fine and is actually a good base for a "generic" load_nd/store_nd ops implementation for hw that does not support 2d block loads.
Using lane_id for further offset manipulation in conversion looks hacky, but so is the whole pattern, so it should be fine for the initial enabling.

silee2:
Not needed anymore.

PR #198021, opened 2026-05-15
[AArch64] Do not pass debug insn to liveness analysis

LU-JOHN:
Fix another stepBackward location.

Debug instructions must not affect liveness analysis. stepBackward has an assertion failure on debug instructions after https://github.com/llvm/llvm-project/pull/193104.


ilovepi (approved):
LGTM. seems inline w/ the other fixes.

PR #197994, opened 2026-05-15
[SLP] Vectorize struct-returning intrinsics

alexey-bataev:
Allow SLP to combine across lanes calls that return a literal struct
(llvm.sincos, llvm.*.with.overflow, llvm.frexp, ...) into a single
call returning a struct of vectors, by widening {T, T, ...} to
{<VF x T>, ...} via VectorTypeUtils and emitting extractvalue +
extractelement for external uses.

Original Pull Request: https://github.com/llvm/llvm-project/pull/195521

Original Pull Request2: https://github.com/llvm/llvm-project/pull/196756

Recommit after revert https://github.com/llvm/llvm-project/pull/197969

Added check for valid vectorizable type.

PR #198022, opened 2026-05-15
Revert "[SLP] Vectorize struct-returning intrinsics"

Kewen12:
Reverts llvm/llvm-project#197994
unblock bot: https://lab.llvm.org/buildbot/#/builders/10


alexey-bataev:
The fix is ready, no need to revert!

PR #198023, opened 2026-05-15
[LLVM][MLIR] Fix TU-local entity exposure for C++20 module compatibility

RedNicStone:
Fix TU-local entity exposure issues across LLVM and MLIR that prevent
compilation with C++20 modules (GCC and Clang).

## Changes
- Convert namespace-scope `static constexpr` to `inline constexpr`
- Move static helper functions into `detail::` namespaces
- Move SFINAE helpers from anonymous namespaces into `detail::` namespaces

This affects ~40+ instances in ADT headers and various MLIR include files.

## Motivation
C++20 modules treat exposure of TU-local entities (via templates or public
interfaces) as ill-formed in many cases. These changes restore module
compatibility without functional or ABI impact.

`inline constexpr` is the modern replacement for `static constexpr` at
namespace scope. Moving helpers to `detail::` properly hides internal
implementation details.

## Testing
No functional changes. Builds successfully with `-DLLVM_ENABLE_MODULES=ON`.
No new tests required.

### DIFF
--- a/llvm/include/llvm/ADT/SetOperations.h
+++ b/llvm/include/llvm/ADT/SetOperations.h
@@ -25,7 +25,7 @@ using check_has_member_remove_if_t =
     decltype(std::declval<Set>().remove_if(std::declval<Fn>()));
 
 template <typename Set, typename Fn>
-static constexpr bool HasMemberRemoveIf =
+inline constexpr bool HasMemberRemoveIf =
     is_detected<check_has_member_remove_if_t, Set, Fn>::value;
 
 template <typename Set>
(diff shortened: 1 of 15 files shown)


llvm/include/llvm/ADT/SetOperations.h
```diff
@@ -25,15 +25,15 @@ using check_has_member_remove_if_t =
     decltype(std::declval<Set>().remove_if(std::declval<Fn>()));
 
 template <typename Set, typename Fn>
-static constexpr bool HasMemberRemoveIf =
+inline constexpr bool HasMemberRemoveIf =
```
arsenm, line 28: inline is redundant with constexpr

RedNicStone: In C++17, constexpr at namespace scope has internal linkage (same as static), so just removing static would have no effect on TU-local leakage.

RedNicStone: I recompiled without inline to check, and Clang does produce TU-local related warnings without it. Which makes sense, constexpr is const-qualified, and const has internal linkage in namespace scopes.

PR #198026, opened 2026-05-15
[lldb] Make CommandObject::GetTarget filter out the dummy target

JDevlieghere:
Follow-up to #197805. Make CommandObject::GetTarget the canonical target accessor for command code, and tighten its semantics so that DoExecute methods can't accidentally operate on the dummy target.

GetTarget now returns Target* instead of Target&. The result is the target from the command's frozen execution context, falling back to the interpreter's execution context. The dummy target is filtered out and replaced with nullptr unless the command opts in via one of the eCommandRequires{Target,Process,Thread,Frame} flags (in which case CheckRequirements has already guaranteed a real target) or via the new eCommandAllowsDummyTarget flag.

This is the first half of the cleanup discussed at the end of #197805. A follow-up will audit DoExecute methods that still reach for GetSelectedTarget or m_exe_ctx.GetTargetPtr() directly and migrate them to GetTarget.

### DIFF
--- a/lldb/include/lldb/Interpreter/CommandInterpreter.h
+++ b/lldb/include/lldb/Interpreter/CommandInterpreter.h
@@ -478,7 +478,13 @@ public:
     return m_debugger.GetTargetList().GetSelectedTarget();
   }
 
-  ExecutionContext GetExecutionContext() const;
+  /// Returns the execution context the interpreter should run a command in.
+  /// If `adopt_dummy_target` is true and no real target is selected, the
+  /// dummy target is substituted in. Pass false from CommandObject paths
+  /// where the command hasn't opted into the dummy via
+  /// eCommandAllowsDummyTarget, so callers can't inadvertently end up
+  /// operating on the dummy.
+  ExecutionContext GetExecutionContext(bool adopt_dummy_target = true) const;
 
   lldb::PlatformSP GetPlatform(bool prefer_target_platform);
 
--- a/lldb/include/lldb/Interpreter/CommandObject.h
+++ b/lldb/include/lldb/Interpreter/CommandObject.h
@@ -379,11 +379,14 @@ protected:
 
   Target &GetDummyTarget();
 
-  // This is for use in the command interpreter, and returns the most relevant
-  // target. In order of priority, that's the target from the command object's
-  // execution context, the target from the interpreter's execution context, the
-  // selected target or the dummy target.
-  Target &GetTarget();
+  /// Get the target this command should operate on. Prefers the frozen
+  /// execution context in the command object, falling back to the
+  /// interpreter's execution context. The dummy target is filtered out unless
+  /// the command has one of the eCommandRequires{Target,Process,Thread,Frame}
+  /// flags (in which case CheckRequirements has already guaranteed a real
+  /// target) or has opted in via eCommandAllowsDummyTarget. Returns null when
+  /// no target is available.
+  Target *GetTarget();
 
   // If a command needs to use the "current" thread, use this call. Command
   // objects will have an ExecutionContext to use, and that may or may not have
--- a/lldb/include/lldb/lldb-enumerations.h
+++ b/lldb/include/lldb/lldb-enumerations.h
@@ -1235,7 +1235,14 @@ FLAGS_ENUM(CommandFlags){
     ///
     /// Verifies that the process is being traced by a Trace plug-in, if it
     /// isn't the command will fail with an appropriate error message.
-    eCommandProcessMustBeTraced = (1u << 8)};
+    eCommandProcessMustBeTraced = (1u << 8),
+    /// eCommandAllowsDummyTarget
+    ///
+    /// Indicates that the command can legitimately operate on the dummy target
+    /// (e.g. `breakpoint set` priming future targets). Without this flag,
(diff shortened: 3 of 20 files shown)


jimingham:
I think you are getting tripped up by the fact that for convenience, commands that could use the dummy target were doing some auto-detection behind your back.  For instance if you want to set a breakpoint in your ~/.lldbinit file you  only have to say `breakpoint add address 0x12345`.  We don't require you to explicitly pass the --dummy flag, even though you are implicitly asking for that behavior.  That means when there is no target available, even if you didn't say "use dummy target" GetTarget() would have to provide the dummy target anyway.

That's why the construct:

    Target &target = m_options.m_use_dummy ? GetDummyTarget() : GetTarget();

worked.  In your version, this would have to be:

    Target *target = m_options.m_use_dummy ? &GetDummyTarget() : GetTarget();
     if (!target)
        target = &GetDummyTarget();
 
That seems awkward to have to do every time, however.  So I think it would be better that if eCommandAllowsDummyTarget is set in the command flags, GetTarget returns the DummyTarget when the command and interpreter ExecutionContext don't have one.

jimingham:
`eCommandAllowsDummyTarget` is really `eCommandRequiresTargetButTheDummyTargetWillDo`.

JDevlieghere:
> I think it would be better that if eCommandAllowsDummyTarget is set in the command flags, GetTarget returns the DummyTarget when the command and interpreter ExecutionContext don't have one.

Unless I'm misunderstanding your point, I believe that's already the case, which is why you don't need the addtional

```
if (!target)
  target = &GetDummyTarget();
```

 Admittedly, the logic is a bit confusing, but the way we get to dummy target is through `CommandInterpreter::GetExecutionContext()` which adopts the dummy target. We only clear it out if the `eCommandAllowsDummyTarget` flag (or friends) are missing.

jimingham:
> > I think it would be better that if eCommandAllowsDummyTarget is set in the command flags, GetTarget returns the DummyTarget when the command and interpreter ExecutionContext don't have one.
> 
> Unless I'm misunderstanding your point, I believe that's already the case, which is why you don't need the addtional
> 
> ```
> if (!target)
>   target = &GetDummyTarget();
> ```
> 
> Admittedly, the logic is a bit confusing, but the way we get to dummy target is through `CommandInterpreter::GetExecutionContext()` which adopts the dummy target. We only clear it out if the `eCommandAllowsDummyTarget` flag (or friends) are missing.

I see... This code depends on behavior that's pretty far away, and CommandInterpreter::GetExecutionContext doesn't say it does `adopt_dummy_target = true` if you don't go read the code for it.  So it would be good to make this a little clearer, if only through comments.

Another thing that's awkward here though is that CommandObject::CheckRequirements will adopt the dummy target into the m_exe_ctx of the CommandObject without checking if the command allows the dummy target.  That doesn't seem right - we shouldn't put the Dummy Target into m_exe_ctx when the command says it doesn't want it.  

Maybe a cleaner way would be to make `CommandInterpreter::GetExecutionContext` take an `adopt_dummy` parameter rather than just passing it implicitly.  Then both in CheckRequirements and the code here  we would pass `flags & eCommandAllowsDummyTarget` to `adopt_dummy`.  Then you'd always be explicitly getting the entity you wanted from the CommandObject and later from the CommandInterpreter?  

Then this function could be:

```
  // If the command allows the dummy target, and we're in the main execution of DoExecute, this
  // will be the selected target or the dummy if there wasn't one and this command allows dummy targets
  Target *target = m_exe_ctx.GetTargetPtr();
  if (!target) {
      // We come in here if we're in a callback where m_exe_ctx has been unset, and we have to
      // fall back to the interpreter's execution context:
      target = m_interpreter.GetExecutionContext(flags & eCommandAllowsDummyTarget).GetTargetPtr();
  }
  // 
  bool requires_target = flags & (eCommandRequiresTarget | eCommandRequiresProcess |
               eCommandRequiresThread | eCommandRequiresFrame);
    assert(target || !requires_target);

   return target;
```
Note you don't have to check for eCommandAllowsDummyTarget at the end because if eCommandAllowsDummyTarget is true, you'll always get a Dummy Target either from the CommandObject or the Interpreter.

I think you could even leave that assert out, that will only happen if we started out a command with a target and then the command had some kind of callback that got run after m_exe_ctx was cleared out, but by that time ALL the targets had been deleted.  That seems pretty unlikely.

As an aside, this would all be SO much easier if we had the notion of a `CommandExecution` in parallel with the `CommandObject`. So `CommandObjectFoo` would hold the command definition and parser, but the actual data store for options and arguments would be the `CommandExecutionFoo`.  Then when you issued a command, `CommandObjectFoo` would make a new `CommandExecutionFoo`, stick the correct `m_exe_ctx` into the `CommandExecutionFoo` and it would be good for the entire run of that command instance.  Then you wouldn't have to have this awkward fallback to the Interpreter's Execution Context, and this reasoning would be clearer.

But we've known that was an unfortunate decision for a while now.

Anyway, maybe this is bike shedding.  

If you're happy with the way this is now, then just put in a comment saying that BOTH CheckRequirements and the explicit call to GetExecutionContext will always insert the dummy target if there's no other target, and that's why you have to filter it back out again.

jimingham (approved):
That looks a lot clearer, thanks!

PR #198027, opened 2026-05-15
[mlir][xegpu] Deprecate XeGPUSubgroupDistribute and rename XeGPUSgToWiDistributeExperimental to XeGPUSgToLaneDistribute

nbpatel:
The XeGPUSubgroupDistribute pass is now fully superseded by the newer subgroup-to-lane distribution flow, so this PR removes its implementation & all associated tests.
The replacement pass XeGPUSgToWiDistributeExperimental is renamed to XeGPUSgToLaneDistribute.


mlir/lib/Dialect/XeGPU/Utils/XeGPUUtils.cpp
```diff
@@ -717,8 +717,6 @@ Value xegpu::lowerToVectorReductions(TypedValue<VectorType> src,
   Value reductionResult = arith::ConstantOp::create(
       rewriter, loc, acc.getType(),
       DenseElementsAttr::get(acc.getType(), zeroAttr));
-  // TODO: Remove these get/setTemporaryLayout calls after we deprecate the old
```
silee2, line 720: If I understand correctly, these getTemporaryLayout calls are needed with then new pass as well?

charithaintc (approved):
LGTM. Thanks!

mlir/lib/Dialect/XeGPU/Utils/XeGPUUtils.cpp
nbpatel: yes

silee2 (approved):

Jianhui-Li (approved):
LGTM

PR #198031, opened 2026-05-15
[MLIR][XeGPU] Temporarily disable XeGPU peephole optimizer for CRI

silee2:
Temporarily disabling the optimization pass until a fix is ready.
Support for CRI was added in recent PR: https://github.com/llvm/llvm-project/pull/197229
But had post merge issues.


mshahneo (approved):
LGTM.

PR #198035, opened 2026-05-15
[lldb] Change ValueObject::Clone to take StringRef (NFC)

kastiglione:
Make `ValueObject`'s name being a `ConstString` more of an implementation detail by changing `Clone` and `SetName` take a `StringRef`.

### DIFF
--- a/lldb/include/lldb/Expression/ExpressionVariable.h
+++ b/lldb/include/lldb/Expression/ExpressionVariable.h
@@ -58,7 +58,7 @@ public:
     m_frozen_sp->GetValue().SetCompilerType(compiler_type);
   }
 
-  void SetName(ConstString name) { m_frozen_sp->SetName(name); }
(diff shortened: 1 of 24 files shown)


jimingham (approved):
Seems fine to me.

In the early stages of lldb most of the ValueObjects got their names from Variables or the ivar names from types, which would all come to us as ConstStrings.  But Cast and Clone, which both make up new names, wouldn't be getting them from the ConstString pool anyway.  So making callers have to go from their made up name which is probably not a ConstString to a ConstString explicitly is ugly.

The one pseudo-downside to this change is that when SetName took a ConstString you knew how SetName was going to manage the memory for its copy of the name: it lives in the const string pool and doesn't rely on the lifecycle of your copy.  But now you are relying on the fact that SetName itself is going to make a copy of the incoming string, a fact that's not expressed anywhere.  That's not a big deal, however, because it would be stupid for SetName to do anything else...

bulbazord (approved):

PR #197566, opened 2026-05-13
[clang-sycl-linker] Migrate tests from Driver/ to Tooling/ and use LLVM IR input

YuriPlyakhin:
1. Replace the C++ source test that required compiling with %clangxx and separate Input files with self-contained .ll tests using split-file.

2. Split the test into two files:
  - clang-sycl-linker.ll: basic tool behavior (link, dev libs, AOT, errors)
  - clang-sycl-linker-split-mode.ll: device code split mode handling

Co-Authored-By: Claude

PR #198036, opened 2026-05-15
[clang-sycl-linker] Infer target triple from input bitcode

YuriPlyakhin:
Previously clang-sycl-linker required `-triple=` on the command line and
errored out if it was missing. This is unnecessary requirement,
since LLVM bitcode modules already carry a target triple in
their metadata.

Resolve the triple in this order:
1. `-triple=` on the command line, if given.
2. Input bitcode's `target triple` metadata, if exists.
3. Otherwise, error out.

When inputs disagree with the resolved triple, emit an error.
Triple discovery is folded into the existing input
loop in `linkDeviceCode`, so no extra I/O pass over the bitcode files.

Update the test to drop redundant `-triple=spirv64` flags and add
coverage for inference, mismatch errors, the triple recorded in
the offload image, and the no-triple error path.

Addresses: https://github.com/llvm/llvm-project/pull/197566#discussion_r3238056239

Co-Authored-By: Claude


YuriPlyakhin:
@maksimsab , fyi

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
```diff
@@ -279,24 +283,46 @@ Expected<LinkResult> linkDeviceCode(ArrayRef<std::string> InputFiles,
...
+    if (!T.str().empty()) {
+      if (!TargetTriple) {
+        TargetTriple = T;
+        TripleSource = File;
+      } else if (T != *TargetTriple) {
+        return createStringError(
+            "conflicting target triples: '" + TargetTriple->str() + "' (from " +
+            TripleSource + ") vs '" + T.str() + "' (from " + File + ")");
+      }
+    }
```
bader:
```suggestion
    if (!T.empty() && T != *TargetTriple) {
      if (TargetTriple) {
        return createStringError(
            "conflicting target triples: '" + TargetTriple->str() + "' (from " +
            TripleSource + ") vs '" + T.str() + "' (from " + File + ")");
      } else {
        TargetTriple = T;
        TripleSource = File;
      }
    }
```

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
```diff
@@ -245,13 +245,17 @@ Expected<SmallVector<std::string>> getSYCLDeviceLibs(const ArgList &Args) {
...
-/// 1. Link all SYCL device bitcode images into one image. Device linking is
+/// 1. Resolve the target triple: use --triple= when given, otherwise take the
+/// first input that supplies a triple as canonical. Issue an error if any
+/// triple inputs disagree.
+/// 2. Link all SYCL device bitcode images into one image. Device linking is
 /// performed using the linkInModule API.
-/// 2. Gather all SYCL device library bitcode images.
-/// 3. Link all the images gathered in Step 2 with the output of Step 1 using
+/// 3. Gather all SYCL device library bitcode images.
+/// 4. Link all the images gathered in Step 3 with the output of Step 2 using
```
bader, line 263: There is nothing SYCL specific in this function. It links LLVM bitcode + imports libraries.
Please, add a TODO to remove:
1. Mentions of SYCL
2. "fat" file formats. This linker shouldn't refer to input files as "device code"/"device libraries"

It's able to link SPIR-targeted LLVM IR from OpenMP compiler as well.

clang/test/Tooling/clang-sycl-linker.ll
```diff
@@ -7,41 +7,43 @@
 ; RUN: llvm-as %t/input2.ll -o %t/input2.bc
 ;
 ; Test the dry run of a simple case to link two input files.
-; RUN: clang-sycl-linker --dry-run -v -triple=spirv64 --module-split-mode=none %t/input1.bc %t/input2.bc -o %t/spirv.out 2>&1 \
+; RUN: clang-sycl-linker -v --module-split-mode=none %t/input1.bc %t/input2.bc -o %t/spirv.out 2>&1 \
```
bader: Please, keep `--dry-run` option. Removing it is not related to the change.

bader (approved):
LGTM, just a few nits.
I would like @sarnex to take a look as well.

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
YuriPlyakhin: this suggestion will not work. std::optional can be dereferenced only if it has value. I removed `.str()`.
https://github.com/llvm/llvm-project/pull/198036/changes/87f7cbca4113ee5e9242e5fb50727a20d47f9d8c

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
YuriPlyakhin: https://github.com/llvm/llvm-project/pull/198036/changes/87f7cbca4113ee5e9242e5fb50727a20d47f9d8c

clang/test/Tooling/clang-sycl-linker.ll
YuriPlyakhin: https://github.com/llvm/llvm-project/pull/198036/changes/87f7cbca4113ee5e9242e5fb50727a20d47f9d8c

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
```diff
@@ -634,7 +667,7 @@ Error runSYCLLink(ArrayRef<std::string> Files, const ArgList &Args) {
   Expected<LinkResult> LinkedOrErr = linkDeviceCode(Files, Args, C);
   if (!LinkedOrErr)
     return LinkedOrErr.takeError();
-  auto &[LinkedModule, LinkedFile] = *LinkedOrErr;
+  auto &[LinkedModule, LinkedFile, ResolvedTriple] = *LinkedOrErr;
```
sarnex: nit: it's cleaner to me if we don't decompose the returned struct and instead just do LinkResult Result = *LinkedOrErr and then just do struct member refs,otherwise as the struct grows this line will become even crazie

clang/test/Tooling/clang-sycl-linker.ll
```diff
@@ -68,10 +70,27 @@
...
 ;
-; Check that the target triple must be specified.
-; RUN: not clang-sycl-linker --dry-run %t/input1.bc %t/input2.bc -o a.out 2>&1 \
-; RUN:   | FileCheck %s --check-prefix=NOTARGET
-; NOTARGET: Target triple must be specified
+; Test error on mismatched triple between inputs.
+; RUN: llvm-as %t/input-mismatch.ll -o %t/input-mismatch.bc
+; RUN: not clang-sycl-linker --dry-run %t/input1.bc %t/input-mismatch.bc -o %t/mismatch.out 2>&1 \
+; RUN:   | FileCheck %s --check-prefix=TRIPLE-MISMATCH
+; TRIPLE-MISMATCH: error: conflicting target triples: 'spirv64' (from {{.*}}input1.bc) vs 'spirv32' (from {{.*}}input-mismatch.bc)
```
sarnex: nit: personally i dont like having a trillion tests in a single test file but that's not really introduced by this change

clang/test/Tooling/clang-sycl-linker.ll
```diff
@@ -68,10 +70,27 @@
...
+;
+; Test that explicit -triple= is used. Input does not supply a triple.
```
sarnex:
```suggestion
; Test when explicit -triple= is used. Input does not supply a triple.
```

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
bader: What is the value of using `std::optional` in this context? IMHO, default-constructed Triple object makes code a little bit cleaner.

YuriPlyakhin: good catch. I updated the code: https://github.com/llvm/llvm-project/pull/198036/changes/a2d082d6dd737eac255bcbdb73d1da08a9921ee4

clang/tools/clang-sycl-linker/ClangSYCLLinker.cpp
YuriPlyakhin: https://github.com/llvm/llvm-project/pull/198036/changes/e5db90bfe7814b2651c8b195de5225283301d82e

clang/test/Tooling/clang-sycl-linker.ll
YuriPlyakhin: https://github.com/llvm/llvm-project/pull/198036/changes/a6ad0a87b593140b63262f65907528006f8f6906

clang/test/Tooling/clang-sycl-linker.ll
YuriPlyakhin: https://github.com/llvm/llvm-project/pull/198036/changes/a6ad0a87b593140b63262f65907528006f8f6906

sarnex (approved):

PR #198037, opened 2026-05-15
[SPIR-V] Implement SPV_KHR_poison_freeze extension

MrSidims:
Specification can be found here:
https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_poison_freeze.asciidoc


MrSidims:
There are actually bugs for composites handling, so converting to draft now

MrSidims:
spirv-val issues: I believe it's incorrect, reported in https://github.com/KhronosGroup/SPIRV-Registry/issues/416
composites handling: still investigating

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
```diff
@@ -447,6 +449,14 @@ static bool isAggrConstForceInt32(const Value *V) {
          isa<ConstantDataArray>(V) || IsAggrZero || IsUndefAggregate;
 }
 
+// True for the i32-placeholder intrinsics that carry an aggregate's real type
+// in AggrConstTypes.
+static bool isSpvAggrPlaceholder(const Value *V) {
+  return match(V, m_Intrinsic<Intrinsic::spv_undef>()) ||
+         match(V, m_Intrinsic<Intrinsic::spv_poison>()) ||
+         match(V, m_Intrinsic<Intrinsic::spv_const_composite>());
```
jmmartinez:
```suggestion
  return match(V, m_AnyIntrinsic<Intrinsic::spv_undef, Intrinsic::spv_poison, Intrinsic::spv_const_composite>());
```

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
```diff
@@ -1632,6 +1647,48 @@ void SPIRVEmitIntrinsics::preprocessUndefs(IRBuilder<> &B) {
...
+// When SPV_KHR_poison_freeze is enabled, replace each LLVM PoisonValue operand
+// with a spv_poison intrinsic so selection emits OpPoisonKHR.
+void SPIRVEmitIntrinsics::preprocessPoisons(IRBuilder<> &B) {
+  const SPIRVSubtarget *STI = TM.getSubtargetImpl(*CurrF);
+  if (!STI->canUseExtension(SPIRV::Extension::SPV_KHR_poison_freeze))
+    return;
+
+  std::queue<Instruction *> Worklist;
+  for (auto &I : instructions(CurrF))
+    Worklist.push(&I);
```
jmmartinez: We don't need to put all the instructions on a list. I we're reprocessing the new ones (`buildSpvUndefOrPoisonComposite` is not adding the new ones to the worklist, and the call to `Intrinsic::spv_poison` doesn't have any arguments).

The same applies for `preprocessUndefs`.

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
```diff
@@ -1632,6 +1647,48 @@ void SPIRVEmitIntrinsics::preprocessUndefs(IRBuilder<> &B) {
...
+    bool BPrepared = false;
+    Worklist.pop();
+
+    for (auto &Op : I->operands()) {
+      auto *Poison = dyn_cast<PoisonValue>(Op);
+      if (!Poison || Op->getType()->isMetadataTy())
+        continue;
+
+      if (!BPrepared) {
+        setInsertPointSkippingPhis(B, I);
```
jmmartinez: What happens if the `posion` is on a phi-node ? Do we insert the `poison` operand after the phi ?

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
```diff
@@ -2594,7 +2663,8 @@ void SPIRVEmitIntrinsics::insertAssignTypeIntrs(Instruction *I,
     Type *TypeToAssign = Ty;
     if (auto *II = dyn_cast<IntrinsicInst>(I)) {
       if (II->getIntrinsicID() == Intrinsic::spv_const_composite ||
-          II->getIntrinsicID() == Intrinsic::spv_undef) {
+          II->getIntrinsicID() == Intrinsic::spv_undef ||
+          II->getIntrinsicID() == Intrinsic::spv_poison) {
```
jmmartinez: This 2 ifs can be replaced with a call to `isSpvAggrPlaceholder`.

llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
```diff
@@ -4682,6 +4683,19 @@ bool SPIRVInstructionSelector::selectIntrinsic(Register ResVReg,
     MIB.constrainAllUses(TII, TRI, RBI);
     return true;
```
jmmartinez, line 4735: Now that there is support from the SPIRV standard; should we fail with an unsupported error message?

It's not super friendly; but introducing bugs from not respecting the freeze semantics seem worse to me.

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
MrSidims: Nothing good :) Exactly the reason of https://github.com/llvm/llvm-project/pull/198037#issuecomment-4466630488

llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
MrSidims: I don't think we should do this, it would be a significant regression for everybody, that do not support the extension, basically blocking compiler promotion without driver update.

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
MrSidims: Applied

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
MrSidims: Applied

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
MrSidims: Applied,  thanks!

llvm/lib/Target/SPIRV/SPIRVInstructionSelector.cpp
jmmartinez: A warning then?

We've put ourselves in an uncomfortable place in here :S

MrSidims: Well, such warning would be unavoidable and users couldn't do anything about it. And such warnings would make users trained-to-ignore.

I've added few LLVM_DEBUG calls to highlight lowering skipping freeze and poison to the devs.

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/poison-composite.ll
```diff
@@ -0,0 +1,43 @@
+; RUN: llc -O0 -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_KHR_poison_freeze %s -o - | FileCheck %s
+; RUN: llc -O0 -mtriple=spirv-vulkan-unknown --spirv-ext=+SPV_KHR_poison_freeze %s -o - | FileCheck %s
+; FIXME: add spiv-val
```
MrSidims: https://github.com/KhronosGroup/SPIRV-Tools/pull/6713 to fix SPIR-V val and https://github.com/KhronosGroup/SPIRV-Registry/pull/417 to fix the spec

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
```diff
@@ -1605,33 +1610,82 @@ void SPIRVEmitIntrinsics::replaceMemInstrUses(Instruction *Old,
...
+        }
+        auto *Call =
+            B.CreateIntrinsic(Intrinsic::spv_poison, {B.getInt32Ty()}, {});
+        AggrConsts[Call] = Poison;
+        AggrConstTypes[Call] = OpTy;
+        Replacement = Call;
+      } else {
+        if (Phi)
+          B.SetInsertPoint(Phi->getIncomingBlock(Idx)->getTerminator());
+        else if (!BPrepared) {
```
jmmartinez, line 1674: `BPrepared` is always `false` in all its uses.

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
```diff
@@ -1605,33 +1610,82 @@ void SPIRVEmitIntrinsics::replaceMemInstrUses(Instruction *Old,
...
+void SPIRVEmitIntrinsics::preprocessPoisons(IRBuilder<> &B) {
+  const SPIRVSubtarget *STI = TM.getSubtargetImpl(*CurrF);
+  if (!STI->canUseExtension(SPIRV::Extension::SPV_KHR_poison_freeze))
+    return;
+
+  SmallVector<Instruction *, 16> Insts;
+  for (auto &I : instructions(CurrF))
+    Insts.push_back(&I);
+
+  for (Instruction *I : Insts) {
```
jmmartinez: Would this work ?

```suggestion
  for (Instruction *I : instructions(CurrF)) {
```

We may create new calls to `spv_poison` that will get visited by this loop; but, nothing will change since a call to `spv_poison` doesn't have any poison operand.

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
MrSidims: It's always `false` only at the first iteration of `for (unsigned Idx = 0; Idx < I->getNumOperands(); ++Idx)` loop on L1657. On next iterations it can be `true`.

MrSidims: Note, there is actually a bug in processing composites in `preprocessUndefs` (which has the same logic, but a bit different code), but I'll create a separate PR for this.

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
MrSidims: Thanks! It should indeed work. Applied.

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
```diff
@@ -0,0 +1,16 @@
+; RUN: llc -O0 -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_KHR_poison_freeze %s -o - | FileCheck %s
+; RUN: %if spirv-tools %{ llc -O0 -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_KHR_poison_freeze %s -o - -filetype=obj | spirv-val %}
+
+; Verify that the ArithmeticPoisonKHR execution mode is attached to kernel
+; entry points when SPV_KHR_poison_freeze is enabled and poison present in the
+; module.
```
maarquitos14: Should we also check that it's not emitted if `poison` is not present in the module? Also, it made me think... what happens if it's present in the module, in a function that is not exercised by any entry point? Should we still emit it?

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
```diff
@@ -0,0 +1,16 @@
...
+
+; CHECK-DAG: OpCapability PoisonFreezeKHR
+; CHECK-DAG: OpExtension "SPV_KHR_poison_freeze"
+; CHECK:     OpEntryPoint Kernel %[[#ENTRY:]] "kernel_with_poison"
+; CHECK-DAG: OpExecutionMode %[[#ENTRY]] ArithmeticPoisonKHR
```
maarquitos14: Should we check that it doesn't get it twice? The explicit one + the implicit one because of the capability.

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/freeze.ll
```diff
@@ -0,0 +1,40 @@
+; RUN: llc -O0 -mtriple=spirv64-unknown-unknown --spirv-ext=+SPV_KHR_poison_freeze %s -o - | FileCheck %s
+; RUN: llc -O0 -mtriple=spirv-vulkan-unknown --spirv-ext=+SPV_KHR_poison_freeze %s -o - | FileCheck %s
```
maarquitos14, line 2: I assume we already have tests to cover the scenarios where the extension is not enabled?

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
jmmartinez: Ahhh right ! sorry I missed it.

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
MrSidims: Added few more tests

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
```diff
@@ -5,12 +5,22 @@
 ; entry points when SPV_KHR_poison_freeze is enabled and poison present in the
 ; module.
 
-; CHECK-DAG: OpCapability PoisonFreezeKHR
-; CHECK-DAG: OpExtension "SPV_KHR_poison_freeze"
-; CHECK:     OpEntryPoint Kernel %[[#ENTRY:]] "kernel_with_poison"
-; CHECK-DAG: OpExecutionMode %[[#ENTRY]] ArithmeticPoisonKHR
+; CHECK-DAG:     OpCapability PoisonFreezeKHR
+; CHECK-DAG:     OpExtension "SPV_KHR_poison_freeze"
+; CHECK-COUNT-2: OpExecutionMode %[[#]] ArithmeticPoisonKHR
```
maarquitos14: Help me understand: we expect 2 here, one for `entry_a`, and the second for? `poison_helper`? If so, `poison_helper` is not an entry point, so it shouldn't have it, should it?

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
```diff
@@ -5,12 +5,22 @@
...
+; CHECK-NOT:     OpExecutionMode %[[#]] ArithmeticPoisonKHR
```
maarquitos14: This is for `entry_b`, right?

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
maarquitos14: > what happens if it's present in the module, in a function that is not exercised by any entry point? Should we still emit it

I think this is still not covered. To be fair, I don't know if it should be covered, I'm still not sure if we should emit or not. Actually, this might even be a bit offtopic, because I'm going to discuss the wording of the spec, but here we go. Spec says:

> PoisonFreezeKHR
> Uses the OpFreezeKHR instruction or the ArithmeticPoisonKHR execution mode. If declared, all entry points must use the ArithmeticPoisonKHR execution mode.

But then, `OpPoisonKHR` enables `PoisonFreezeKHR` capability as well. So the quote above should either mention both `OpFreezeKHR` and `OpPoisonKHR`, or none, as both enable the capability.

Considering the above, the simple fact of having `poison` or `freeze` in the module, enables the capability, and so, we should emit the exec mode for all entry points even if the `poison/freeze` is not reachable by any of the entry points in the module. @MrSidims do you agree?

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
maarquitos14: Also, a further question here, do we expect the BE to handle the metadata `!spirv.ExecutionMode`? If so, should we add a test? If not, why?

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/poison-composite.ll
aobolensk: Since spirv-val related PR has been merged, that could be added already, I think

MrSidims: Added

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
MrSidims: Clarified via changing comment and checks

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
MrSidims: Clarified via changing comment and checks

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
MrSidims: I disagree. We still has separation of `OpUndef` and `OpPoisonKHR`.

Also note, that it's not necessary to have an explicit OpPoisonKHR instruction in a function/entry point to have a poisoness behaviour - poisoness behaviour can occur at runtime and it might be handled differently depending on the execution mode.

maarquitos14: Sorry, I think my wording was a bit confusing. Let me try and rephrase:

> Considering the above, the simple fact of having `OpPoisonKHR` or `OpFreezeKHR` in the module, enables the capability, and so, if the extension es enabled, we should emit the exec mode for all entry points even if the `Op{Poison/Freeze}KHR` is not reachable by any of the entry points in the module.

MrSidims: I see, I believe this case is already covered and tested, see https://github.com/llvm/llvm-project/pull/198037/changes#diff-cf27acbff1c11eca2cde90d83fd48a2654c483f296d98f8b74ce1a3f187bcfbfR626-R637 and https://github.com/llvm/llvm-project/pull/198037/changes#diff-331329308b8f03b048b8cbb2a5be838b5aed435de0ba6e62749a063a1f5c609bR28 test

MrSidims: hm, github is not opening links correctly. Yet entry_b in the test doesn't have a poison value, yet it has the execution mode.

MrSidims: If you mean something like that

> lets check, that execution mode is set even if poison is not present in the module while capability is enabled

I don't think, that there is a way to represent OpCability in LLVM IR explicitly.

maarquitos14 (approved):

llvm/test/CodeGen/SPIRV/extensions/SPV_KHR_poison_freeze/arithmetic-poison-execmode.ll
maarquitos14: We discussed offline, and my ask is covered by `arithmetic-poison-execmode.ll`.

jmmartinez (approved):

aobolensk (approved):

llvm/lib/Target/SPIRV/SPIRVEmitIntrinsics.cpp
MrSidims: > Note, there is actually a bug in processing composites in preprocessUndefs (which has the same logic, but a bit different code), but I'll create a separate PR for this.

Ah, I completely forgot about this bug, and now it's firing in my setup on HeCBench/logic-rewrite when -O2 is not ignored by the frontend with a reduced reproducer:
```
target triple = "spirv64-amd-amdhsa"

%s = type { i32, [2 x i8] }

define spir_func %s @f() {
  ret %s { i32 0, [2 x i8] undef }
}
```

PR #123166, opened 2025-01-16
[Clang][Analysis] Don't treat the default switch path as unreachable when all enumerators are covered

jeremy-rifkin:
This PR updates the CFG building logic and analysis based warning logic to not consider the default path for `switch` statements to be unreachable when all enumerators are covered. I.e.:

```cpp
enum class E { a, b };

int foo(E e) {
    switch(e) {
        case E::a:
            return 20;
        case E::b:
            return 30;
    }
    // reachable
}
```

The motivation is here is to have `-Wreturn-type` warnings in cases such as above. Even though it's often sufficient to handle all enumerators, extraneous enum values are valid and common in the case of flag enums. More context at #123153. This is in-line with behavior in other compilers (gcc, msvc, and edg emit warnings for this).

A few test cases broke due to relying on missing returns in the default case. One objective C test cases is failing, which I would appreciate guidance on so that I can be sure to test the right things.

On the llvm discord a point was raised about changes to this warning behavior maybe being disruptive for any codebases that rely heavily on this missing return analysis behavior. I think the fact so few tests are affected in llvm is a good sign, and similarly the fact other compilers warn about this makes any potential disruption limited.

The two relevant commits introducing the enum covering logic:
https://github.com/llvm/llvm-project/commit/50205744c32d3925ba50d1bd69b1c9e6ab3a28b8
https://github.com/llvm/llvm-project/commit/f69ce860487ec0b1eceb23f8391614d5772aae3b

Closes #123153

PR #198041, opened 2026-05-15
[clang] Strict enum coverage check in switches

### LINKED ISSUE #123153: Clang does not warn about missing returns when a switch statement covers all enum values
The following code triggers a missing return warning by gcc, msvc, and edg but not clang:
```cpp
enum class E { a, b };

int foo(E e) {
    switch(e) {
        case E::a:
            return 20;
        case E::b:
            return 30;
    }
}
```
https://godbolt.org/z/b5r34nWf6

The main issue here is safety: This code is bug-prone due to extraneous enum values. The most common situation this could lead to bugs is for flag enums, e.g. https://godbolt.org/z/zx6nnPYMq.

Secondarily: This can lead to some annoyances where code compiles fine on clang but has a flood of warnings on gcc (possibly fatal ones, if using -Werror=return-type). This leads to PRs such as https://github.com/llvm/llvm-project/pull/122382 and https://github.com/llvm/llvm-project/pull/105520.

I'd like to propose changing the behavior to warn on this case by default. I briefly discussed this with folks on the llvm discord as well: https://discord.com/channels/636084430946959380/636732781086638081/1329124923322531850.

emaxx-google:
Add a new option "-fstrict-enum-switch-coverage": when checking whether a switch over enums handles all cases, the enum's underlying type range is used instead of only considering defined enumerators.

For example, enabling this option allows to diagnose a missing return when a switch covers all defined enumerators but not all integers from the underlying type's range.

Fixes: #123153

Assisted-by: Gemini

### DIFF
--- a/clang/test/SemaCXX/warn-return-type-exhaustive-enum.cpp
+++ b/clang/test/SemaCXX/warn-return-type-exhaustive-enum.cpp
@@ -1,0 +1,130 @@
+// RUN: %clang_cc1 -fsyntax-only -Wreturn-type -Wunreachable-code -Warray-bounds -Wconditional-uninitialized -verify=default,expected %s
+// RUN: %clang_cc1 -fsyntax-only -Wreturn-type -Wunreachable-code -Warray-bounds -Wconditional-uninitialized -fno-strict-enum-switch-coverage -verify=default,expected %s
+// RUN: %clang_cc1 -fsyntax-only -Wreturn-type -Wunreachable-code -Warray-bounds -Wconditional-uninitialized -fstrict-enum-switch-coverage -verify=strict,expected %s
+
+enum E { A, B };
+
+int test_missing_return(E e) {
+  switch (e) {
+    case A: return 1;
+    case B: return 2;
+  }
+} // strict-warning {{non-void function does not return a value in all control paths}}
+
+int test_unreachable(E e) {
+  switch (e) {
+    case A: return 1;
+    case B: return 2;
+  }
+  return 3;
+}
+
+void test_array_bounds(E e) {
+  int x[2]; // strict-note {{array 'x' declared here}}
+  switch (e) {
+    case A: return;
+    case B: return;
+  }
+  x[2] = 0; // strict-warning {{array index 2 is past the end of the array (that has type 'int[2]')}}
+}
+
+int test_useless_default_no_unreachable_warning(E e) {
+  switch (e) {
+    case A: return 1;
+    case B: return 2;
+    default: return 3;
+  }
+}
+
+int test_nested(E e1, E e2) {
+  switch (e1) {
+    case A:
+      switch (e2) {
+        case A: return 1;
+        case B: return 2;
+      }
+      break; // break out of outer switch if e2 falls through!
+    case B:
+      return 3;
+  }
+} // strict-warning {{non-void function does not return a value in all control paths}}
+
+enum class BoolEnum : bool { False = false, True = true };
+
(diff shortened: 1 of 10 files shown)


clang/test/SemaCXX/warn-return-type-exhaustive-enum.cpp
```diff
@@ -0,0 +1,138 @@
...
+    default: return 2;
+  }
+}
+
+int test_bool(bool b) {
+  switch (b) { // expected-warning {{switch condition has boolean value}}
+    case false: return 0;
+    case true: return 1;
+  }
+} // default-warning {{non-void function does not return a value in all control paths}}
```
efriedma-quic, line 74: Why do we want a warning for a bool switch that covers all possible values?

efriedma-quic (commented):
If we can add a warning flag, instead of a "-f" dialect flag, I think that would be better.  It would allow using existing pragmas to control the warning, so you could enable/disable it in specific regions.  Not sure how hard that would be.

------

As an alternative, maybe we can tweak the existing behavior so that we treat switches with an explicit "default" differently from switches which don't have one.  If you have a switch with a "default", it probably makes sense to assume the path through the "default" is reachable for the purpose of `-Wreturn-type`, even if all named enums values are covered.  Then we add a flag `-Wenum-switch-default` so you can request warnings for switches over enums that don't have a "default".

The end result isn't exactly the same thing, but maybe sufficient for the specific coding style described.

Sirraide:
I agree that this should probably be a `-W` flag

> As an alternative, maybe we can tweak the existing behavior so that we treat switches with an explicit "default" differently from switches which don't have one. If you have a switch with a "default", it probably makes sense to assume the path through the "default" is reachable for the purpose of `-Wreturn-type`, even if all named enums values are covered. Then we add a flag `-Wenum-switch-default` so you can request warnings for switches over enums that don't have a "default".

Maybe I misunderstood and you’re suggesting the same thing, but if the point of this is primarily to diagnose a missing return, then my suggestion would be to just introduce an option that makes the CFG code assume that we can still fall off the end of a `switch` even if all possible enum values are handled (I think #123166 was supposed to do that but that pr has been stale for a while). 

I’m not sure checking for all possible values of a type would be that useful because most people aren’t writing switch statements w/ `255` cases, and that’s just for an `int8_t`; in any case, these two features (1. not treating a switch that covers all possible enum values as exhaustive, and 2. checking that a switch is _actually_ exhaustive) seem orthogonal to me and should probably be added separately.

efriedma-quic:
> Maybe I misunderstood and you’re suggesting the same thing, but if the point of this is primarily to diagnose a missing return, then my suggestion would be to just introduce an option that makes the CFG code assume that we can still fall off the end of a switch even if all possible enum values are handled.

What you're describing is basically the current version of this PR: add a flag that flips how the CFG code handles switches.

The point of my alternative suggestion was to avoid having two different analysis modes.  Instead of flipping the handling based on a flag, flip it based on whether the switch has a "default".

emaxx-google:
Thanks for the reviews, @efriedma-quic and @Sirraide!

> As an alternative, maybe we can tweak the existing behavior so that we treat switches with an explicit "default" differently from switches which don't have one. If you have a switch with a "default", it probably makes sense to assume the path through the "default" is reachable for the purpose of `-Wreturn-type`, even if all named enums values are covered. Then we add a flag `-Wenum-switch-default` so you can request warnings for switches over enums that don't have a "default".

This alternative sounds good to me. I've started a PR #198416 for the first part - assuming the "default" is reachable if present.

PR #198044, opened 2026-05-15
[lldb] Set SO_NOSIGPIPE on platforms that support it

bulbazord:
On macOS, I've seen instances where debugserver goes down very quickly after it starts up (less than 100ms). Normally, LLDB is able to detect when debugserver goes down and report it without bringing down the entire debug session. However that's not happening here. My best guess is that debugserver is going down before LLDB is ready to react to it.

To mitigate this scenario, adopt SO_NOSIGPIPE. Note that this mostly matters for tools that embed liblldb. The LLDB driver ignores all SIGPIPEs.

rdar://173516461

### DIFF
--- a/lldb/source/Host/common/Socket.cpp
+++ b/lldb/source/Host/common/Socket.cpp
@@ -404,7 +404,11 @@ int Socket::SetOption(NativeSocket sockfd, int level, int option_name,
 }
 
 ssize_t Socket::Send(const void *buf, const size_t num_bytes) {
-  return ::send(m_socket, static_cast<const char *>(buf), num_bytes, 0);
+  int flags = 0;
+#if defined(MSG_NOSIGNAL)
+  flags |= MSG_NOSIGNAL;
+#endif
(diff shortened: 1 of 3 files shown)


JDevlieghere (approved):
I'm not super confident in my review, but as far as I can tell I think this *should* be fine. Assuming this really only controls the behavior of signal (and we still get an EPIPE in select) then I don't see what harm this does.

labath (commented):
I think this is a good idea, but for consistency I'd recommend implementing the functionality on all platforms. Linux (and AFAICT, all BSD variants) implement SIGPIPE suppresion slightly differently. Instead of a socket option, you can suppress the signal on a per-call basis by adding `MSG_NOSIGNAL` to the `send` flags. Given that all our sends are centralized in the Host layer, it should be fairly easy to add the necessary magic.

Bonus points for adding a test. :)

bulbazord:
I added a test. I'll see if I can get the Linux support working before I land. :)

labath (approved):
Cool.

PR #198047, opened 2026-05-16
[DWARFLinker] Fix .debug_names emission for -gmodules in parallel linker

JDevlieghere:
The parallel linker tripped up the DWARF verifier for the .debug_names section when the input contained -gmodules skeleton CUs. The issue was caused by, link(), which reset UniqueUnitID to zero. By that point addObjectFile() had already handed out IDs to clang module CUs loaded from .pcm files, so the regular compile units loaded after the reset reused those IDs. emitDWARFv5DebugNamesSection then collapsed two distinct CUs onto the same CUidToIdx slot, sending one CU's .debug_names entries to the other CU's index. Stop resetting the counter since the constructor already initializes it.


llvm/lib/DWARFLinker/Parallel/DIEAttributeCloner.cpp
```diff
@@ -140,6 +140,13 @@ bool DIEAttributeCloner::shouldSkipAttribute(
   switch (AttrSpec.Attr) {
   default:
     return false;
+  case dwarf::DW_AT_GNU_dwo_id:
+  case dwarf::DW_AT_dwo_id:
+    // The dwo_id identifies a split-DWARF skeleton CU's matching .dwo unit.
+    // The linker inlines the full DIE tree into the output, so the output CU
+    // is not a skeleton and has no .dwo to point at. Keeping the attribute
+    // would mislabel a regular CU as a skeleton.
+    return true;
```
adrian-prantl: Does the verbatim copy of the .pcm contents go through this code?
There is no reason for dsymutil to keep skeleton CUs. But a non-skeleton CU is allowed to carry a DWO id. The differentiating trait in pre-DWARF 5 is that a skeleton has no children and in DWARF 5 that it uses `DW_TAG_skeleton_unit` which AFAIK LLVM doesn't implement.

JDevlieghere: Yes, PCM CUs go through DIEAttributeCloner. Maybe the comment should be rephrased to say that "DWO id is redundant on a non-skeleton CU"?

JDevlieghere: Adrian and I chatted about this a bit more offline, and we concluded that while there's no reason to keep this in dsymutil, it's also not invalid and we should update the verifier to stop tripping up on this.

JDevlieghere: https://github.com/llvm/llvm-project/pull/198112

adrian-prantl: While the attribute was never standardized because it got merged into the unit header, the transitional attribute is present in both:

https://dwarfstd.org/issues/130313.4.html

> The attributes provided by the skeleton compilation unit entry do not need to be repeated in the full compilation unit entry, except for DW_AT_dwo_id, which should appear in both entries so that the consumer can verify that it has found the correct DWARF object.

JDevlieghere: Let's just keep it then

alx32:
LGTM

alx32 (approved):

JDevlieghere:
This now depends on https://github.com/llvm/llvm-project/pull/198112. Waiting for that to merge before re-triggering pre-commit CI.

adrian-prantl (approved):

PR #198049, opened 2026-05-16
[MemoryBuiltins] Capture more information for alloc/free from attributes

jdoerfert:
We now read the `alloc_align` attribute to provide better alignment information to users. `alloc-family` should be used as well, as described in the LangRef. Two new helpers provide argument numbers, rather than values.


jdoerfert:
I'm folding this into https://github.com/llvm/llvm-project/pull/197533.

PR #198051, opened 2026-05-16
[flang] Recognize effects on non-addressable resources in opt-bufferization

vzakhari:
opt-bufferization has been only handling `fir::DebuggingResource` explicitly. This patch adds support for other non-addressable resources, such as `fir::VolatileMemoryResource`. This allows merging elemental/assign for the `volatile_src_nonvolatile_dst` example in the updated LIT test.

### DIFF
--- a/flang/lib/Optimizer/HLFIR/Transforms/OptimizedBufferization.cpp
+++ b/flang/lib/Optimizer/HLFIR/Transforms/OptimizedBufferization.cpp
@@ -164,14 +164,16 @@ static bool haveNonAddressableEffectsConflict(
          const llvm::SmallPtrSetImpl<mlir::SideEffects::Resource *> &set2) {
         if (set1.empty() || set2.empty())
           return false;
-        for (const auto *r1 : set1)
-          for (const auto *r2 : set2)
+        for (const auto *r1 : set1) {
+          for (const auto *r2 : set2) {
             if (!r1->isDisjointFrom(r2)) {
               LLVM_DEBUG(llvm::dbgs()
                          << "resources " << r1->getName() << " and "
                          << r2->getName() << " are not disjoint\n");
               return true;
             }
+          }
+        }
         return false;
       };
   if (accessSameResource(reads1, others2) ||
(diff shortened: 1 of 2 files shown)


flang/lib/Optimizer/HLFIR/Transforms/OptimizedBufferization.cpp
```diff
@@ -122,17 +128,73 @@ getEffectsBetween(mlir::Operation *start, mlir::Operation *end) {
...
+  auto [reads1, others1] = splitEffects(*effects1);
+  auto [reads2, others2] = splitEffects(*effects2);
+
+  auto accessSameResource =
+      [](const llvm::SmallPtrSetImpl<mlir::SideEffects::Resource *> &set1,
+         const llvm::SmallPtrSetImpl<mlir::SideEffects::Resource *> &set2) {
+        if (set1.empty() || set2.empty())
+          return false;
+        for (const auto *r1 : set1)
+          for (const auto *r2 : set2)
```
tblah: ultra-nit: I'd prefer outer braces on the for loops, given the inner if has braces.

https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements uses a slightly different rule, saying that braces should be used with "more than two levels of nesting".

flang/lib/Optimizer/HLFIR/Transforms/OptimizedBufferization.cpp
```diff
@@ -238,15 +300,26 @@ ElementalAssignBufferization::findMatch(hlfir::ElementalOp elemental) {
   // 1) side effects in the elemental body - it isn't sufficient to just look
   // for ordered elementals because we also cannot support out of order reads
   std::optional<mlir::SmallVector<mlir::MemoryEffects::EffectInstance>>
-      effects = getEffectsBetween(&elemental.getBody()->front(),
-                                  elemental.getBody()->getTerminator());
-  if (!effects) {
+      elementalEffects = getEffectsBetween(
+          &elemental.getBody()->front(), elemental.getBody()->getTerminator());
```
tblah: nit: I think this will give the same result as `mlir::getEffectsRecursively`, which I think is slightly easier to read. Mostly I think it would be clearer to have elementalEffects and assignEffects match so that the reader doesn't have to think about what's different.

This is a matter of taste, so feel free to ignore.

tblah (approved):
LGTM, thanks Slava

vzakhari:
Thank you for the review, Tom! I applied your suggestions.

jeanPerier (commented):
Looks great, thanks Slava

PR #198056, opened 2026-05-16
[MLIR][XeGPU] Fix initial value issue with dpas_mx e2m1 test.

silee2:
Integration test was using incorrect initial values.


Jianhui-Li (approved):
LGTM

PR #198062, opened 2026-05-16
[clang][bytecode] Notify bytecode interpreter after deserializing constant global declarations

tbaederr:
The problem looks something like this:

```c++
extern const int m;
constexpr int getm() { return m; }

const int m = 12;
static_assert(getm() == 12);
```

The generated bytecode for `getm()` references a global variable stored in `interp::Program`. This variable is uninitialized when the bytecode is generated, but since it's extern, that's not diagnosed (because it might be initialized later).

When we parse the redeclaration of `m` on line 4, we evaluate its initializer via `Expr::EvaluateAsInitializer`, which makes the bytecode interpreter aware that `m` has been redeclared. It will create a new global for the new declaration and make the previous global point to the new one. When the bytecode is then evaluated in the `static_assert`, there is no problem.

However, when the redeclaration isn't part of the TU but gets de-serialized via a module (which is the case in `test/Modules/redecl-add-after-load.cpp`), we don't call `EvaluateAsInitializer`, we don't even necessarily de-serialize the initializer expression at all. So the bytecode interpreter doesn't know about the new declaration and the static_assert ultimately fails because `m` is not initialized.

Marking this as a draft PR since I'm not sure this is the best way to do this and if bytecode-specific things in other parts of the code base are tolerated.

### DIFF
--- a/clang/test/Modules/redecl-add-after-load.cpp
+++ b/clang/test/Modules/redecl-add-after-load.cpp
@@ -2,6 +2,10 @@
 // RUN: %clang_cc1 -x objective-c++ -fmodules -fimplicit-module-maps -fno-modules-error-recovery -fmodules-cache-path=%t -I %S/Inputs %s -verify -std=c++11
 // RUN: %clang_cc1 -x objective-c++ -fmodules -fimplicit-module-maps -fno-modules-error-recovery -fmodules-cache-path=%t -I %S/Inputs %s -verify -std=c++11 -DIMPORT_DECLS
 
+// RUN: %clang_cc1 -x objective-c++ -fmodules -fimplicit-module-maps -fno-modules-error-recovery -fmodules-cache-path=%t -I %S/Inputs %s -verify -std=c++11 -fexperimental-new-constant-interpreter
+// RUN: %clang_cc1 -x objective-c++ -fmodules -fimplicit-module-maps -fno-modules-error-recovery -fmodules-cache-path=%t -I %S/Inputs %s -verify -std=c++11 -DIMPORT_DECLS -fexperimental-new-constant-interpreter
+
+
 // expected-no-diagnostics
 
 #ifdef IMPORT_DECLS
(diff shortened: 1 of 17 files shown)


efriedma-quic (commented):
If the bytecode interpreter needs to cache information about globals, then sure, let's notify it so it can update its cache.  But it's not clear to me from the description what, exactly, is getting cached.  Or how calling emitInitGlobal() does anything useful; it doesn't actually evaluated anything, does it?

clang/test/Modules/redecl-add-after-load.cpp
```diff
@@ -2,6 +2,10 @@
 // RUN: %clang_cc1 -x objective-c++ -fmodules -fimplicit-module-maps -fno-modules-error-recovery -fmodules-cache-path=%t -I %S/Inputs %s -verify -std=c++11
 // RUN: %clang_cc1 -x objective-c++ -fmodules -fimplicit-module-maps -fno-modules-error-recovery -fmodules-cache-path=%t -I %S/Inputs %s -verify -std=c++11 -DIMPORT_DECLS
 
+// RUN: %clang_cc1 -x objective-c++ -fmodules -fimplicit-module-maps -fno-modules-error-recovery -fmodules-cache-path=%t -I %S/Inputs %s -verify -std=c++11 -fexperimental-new-constant-interpreter
+// RUN: %clang_cc1 -x objective-c++ -fmodules -fimplicit-module-maps -fno-modules-error-recovery -fmodules-cache-path=%t -I %S/Inputs %s -verify -std=c++11 -DIMPORT_DECLS -fexperimental-new-constant-interpreter
```
shafik, line 6: So this test fails w/o your change?

tbaederr: Yep

tbaederr:
>  But it's not clear to me from the description what, exactly, is getting cached. Or how calling emitInitGlobal() does anything useful; it doesn't actually evaluated anything, does it?

We call `visitAPValue` with the deserialized `APValue` before, so the `InitGlobal` will save that value as the new value of the global variable.

ChuanqiXu9 (approved):
Change in serialization part LGTM

PR #198065, opened 2026-05-16
[ConstantFold] Do not fold 'icmp' if any operand is a declaration

igorkudrin:
The compiler cannot assume that distinct symbols represent unique memory locations unless both definitions are visible. External tools, such as the linker or assembler, can alias different symbols to the same address. Therefore, folding an `icmp` instruction when any operand is a declaration to `false` is incorrect.


nikic (changes requested):
This is something I've been wondering about recently. LangRef is not super clear on the semantics in this area, and any change here needs to happen in LangRef first and ConstantFold second.

I believe that de facto, the current semantics here are that (non-zst, non-unnamed-addr) globals are required to have distinct addresses, so aliasing them to the same symbol is UB.

Not that the fact that two distinct external globals do not *alias* (in the AA sense, not the ELF sense) is a hard requirement, for default behavior at least. For constant globals, this is trivially true even if the globals may overlap. However, for non-constant globals, requiring the globals to be disjoint is the only straightforward way to ensure this property. (The alternative would be to give external globals semantics similar to the noalias attribute, but at the module level, which would be a substantial complication.)

efriedma-quic:
From a C++ language perspective, we're required to fold comparisons.  For example:

```
extern int a, b;
static_assert(&a != &b);
```

As a consequence, using assembly trickery to point the two symbols at the same memory is undefined behavior.

-----

We have a mechanism for declaring a global which might alias: make it zero size.  In clang, we allow writing `extern int a[0];`.

Maybe we should change how this is represented in LLVM IR: instead of global declarations having a type, they should store a "dereferenceable size" and a "noalias" bit.  But that doesn't really change the underlying rule.

igorkudrin:
> From a C++ language perspective, we're required to fold comparisons. For example:
> 
> ```
> extern int a, b;
> static_assert(&a != &b);
> ```

Thank you for the example. It's interesting that `gcc` and `msvc` agree with the static assertion but also preserve the runtime comparison: https://godbolt.org/z/jMdczM88q. `clang` applies optimizations more aggressively.

igorkudrin:
Thanks for the discussion. I agree that if the compiler evaluates an expression inside `static_assert()` at compile time, it is only consistent to fold that same expression outside of the statement. Preserving the runtime comparison might be convenient in niche scenarios, but it shouldn't be applied if it harms the optimizer.

Regarding the zero-size globals approach, it is not suitable for our use case because we need to compare function addresses to determine if a function is overridden with `--wrap` or other linker magic. In any case, there are other workarounds we can use.

Closing this now.

PR #198085, opened 2026-05-16
[clang-tidy] Fix false positive in misc-redundant-expression with type aliases

### LINKED ISSUE #145415: [clang-tidy] `misc-redundant-expression` false positive with type aliases of the same type
To reproduce:
```c++
#include <array>
#include <tuple>

using MonthArray = std::array<int, 12>;
using ZodiacArray = std::array<int, 12>;

static_assert(std::tuple_size<MonthArray>::value == std::tuple_size<ZodiacArray>::value);
static_assert(std::tuple_size_v<MonthArray> == std::tuple_size_v<ZodiacArray>);
```
Output:
```
<source>:7:50: warning: both sides of operator are equivalent [misc-redundant-expression]
    7 | static_assert(std::tuple_size<MonthArray>::value == std::tuple_size<ZodiacArray>::value);
      |                                                  ^
<source>:8:45: warning: both sides of operator are equivalent [misc-redundant-expression]
    8 | static_assert(std::tuple_size_v<MonthArray> == std::tuple_size_v<ZodiacArray>);
      |                                             ^
2 warnings generated.
```

In reality `MonthArray` and `ZodiacArray` may be defined far away from each other, or even in completely unrelated modules. When we happen to use `MonthArray` and `ZodiacArray` together, a reader without enough context is likely unaware that they are the same type, and we want to use a `static_assert` to remind the reader that they do have the same `value`. (We could also use `std::same_as<MonthArray, ZodiacArray>` but sometimes we only care about their lengths in the surrounding code) 

In this case comparing `MonthArray` and `ZodiacArray` shouldn't be redundant. Such expressions should only be redundant if the two types are spelled in exactly the same way.

There has been a similar issue https://github.com/llvm/llvm-project/issues/118885 where the final resolution was that comparing different macros expanding to the same constant is non-redundant. This issue is in principal similar, but regarding to type aliases instead of macros.

voyager-jhk:
The `misc-redundant-expression` check previously flagged expressions as redundant if their underlying `DeclRefExpr` pointed to the same declaration. This caused false positives when comparing identical values accessed through distinct type aliases.

Following Clang's diagnostic `aka` logic, this patch uses the AST Printer to stringify and compare the `NestedNameSpecifier` and `TemplateArguments`. This safely preserves sugared types while natively normalizing all whitespaces and newlines, ensuring robustness against multiline formatting changes.

Fixes #145415

### DIFF
--- a/clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
+++ b/clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
@@ -19,6 +19,7 @@
 #include "llvm/ADT/FoldingSet.h"
 #include "llvm/ADT/SmallBitVector.h"
 #include "llvm/Support/FormatVariadic.h"
+#include "llvm/Support/raw_ostream.h"
 #include <algorithm>
 #include <cassert>
 #include <cstdint>
@@ -42,6 +43,42 @@ static bool incrementWithoutOverflow(const APSInt &Value, APSInt &Result) {
   Result = Value;
   ++Result;
   return Value < Result;
+}
+
+static bool areEquivalentDeclRefExpr(const DeclRefExpr *L,
+                                     const DeclRefExpr *R) {
+  if (L->getDecl() != R->getDecl())
+    return false;
+
+  const PrintingPolicy &Policy =
+      L->getDecl()->getASTContext().getPrintingPolicy();
+
+  if (L->hasQualifier() && R->hasQualifier()) {
+    std::string LQual, RQual;
+    llvm::raw_string_ostream LOS(LQual), ROS(RQual);
+    L->getQualifier().print(LOS, Policy);
+    R->getQualifier().print(ROS, Policy);
+    if (LQual != RQual)
+      return false;
+  }
+
+  if (L->hasExplicitTemplateArgs() != R->hasExplicitTemplateArgs())
+    return false;
+  if (L->hasExplicitTemplateArgs()) {
+    if (L->getNumTemplateArgs() != R->getNumTemplateArgs())
+      return false;
+
+    return llvm::equal(L->template_arguments(), R->template_arguments(),
+                       [&Policy](const TemplateArgumentLoc &LArg,
+                                 const TemplateArgumentLoc &RArg) {
+                         std::string LStr, RStr;
+                         llvm::raw_string_ostream LOS(LStr), ROS(RStr);
+                         LArg.getArgument().print(Policy, LOS, true);
+                         RArg.getArgument().print(Policy, ROS, true);
+                         return LStr == RStr;
+                       });
+  }
+  return true;
 }
 
 static bool areEquivalentExpr(const Expr *Left, const Expr *Right) {
@@ -98,8 +135,8 @@ static bool areEquivalentExpr(const Expr *Left, const Expr *Right) {
     return cast<DependentScopeDeclRefExpr>(Left)->getQualifier() ==
            cast<DependentScopeDeclRefExpr>(Right)->getQualifier();
   case Stmt::DeclRefExprClass:
-    return cast<DeclRefExpr>(Left)->getDecl() ==
-           cast<DeclRefExpr>(Right)->getDecl();
+    return areEquivalentDeclRefExpr(cast<DeclRefExpr>(Left),
+                                    cast<DeclRefExpr>(Right));
   case Stmt::MemberExprClass:
     return cast<MemberExpr>(Left)->getMemberDecl() ==
            cast<MemberExpr>(Right)->getMemberDecl();
--- a/clang-tools-extra/docs/ReleaseNotes.rst
+++ b/clang-tools-extra/docs/ReleaseNotes.rst
@@ -549,8 +549,13 @@ Changes in existing checks
   virtual inheritance causes concrete bases to be counted more than once.
 
 - Improved :doc:`misc-redundant-expression
-  <clang-tidy/checks/misc/redundant-expression>` check by fixing a crash when
-  evaluating bitwise comparisons against integer constants wider than 64 bits.
+  <clang-tidy/checks/misc/redundant-expression>` check:
+
+  - Fixed a crash when evaluating bitwise comparisons against integer constants
+    wider than 64 bits.
+
+  - Avoided false positives when comparing expressions that are structurally
+    identical but use different type aliases.
 
 - Improved :doc:`misc-throw-by-value-catch-by-reference
   <clang-tidy/checks/misc/throw-by-value-catch-by-reference>` check:
--- a/clang-tools-extra/test/clang-tidy/checkers/misc/redundant-expression.cpp
+++ b/clang-tools-extra/test/clang-tidy/checkers/misc/redundant-expression.cpp
@@ -1126,3 +1126,49 @@ namespace PR35857 {
     decltype(x + y - (x + y)) z = 10;
   }
 }
+
+namespace GH145415 {
+
+namespace std {
+template <class T, int N> struct array {};
+template <class T> struct tuple_size;
+template <class T, int N> struct tuple_size<array<T, N>> {
+  static constexpr int value = N;
+};
+template <class T> constexpr int tuple_size_v = tuple_size<T>::value;
+} // namespace std
+
+using MonthArray = std::array<int, 12>;
+using ZodiacArray = std::array<int, 12>;
+
+namespace N {
+  int Value = 0;
+}
+using N::Value;
+
+template <typename T> void myFunc(T) {}
+template <typename T, typename U = int> void myDefaultFunc(T) {}
+using FuncPtr = void (*)(int);
+
+void TestGH145415() {
+
+  bool b1 = std::tuple_size<MonthArray>::value == std::tuple_size<ZodiacArray>::value;
+  bool b2 = std::tuple_size_v<MonthArray> == std::tuple_size_v<ZodiacArray>;
+
+  bool b3 = std::tuple_size<MonthArray>::value == std::tuple_size<MonthArray>::value;
+  // CHECK-MESSAGES: :[[@LINE-1]]:48: warning: both sides of operator are equivalent
+
+  bool b4 = std::tuple_size_v<
+                MonthArray> ==
+            std::tuple_size_v<MonthArray>;
+  // CHECK-MESSAGES: :[[@LINE-2]]:29: warning: both sides of operator are equivalent
+
+  bool b5 = N::Value == Value;
+  // CHECK-MESSAGES: :[[@LINE-1]]:22: warning: both sides of operator are equivalent
+
+  bool b6 = FuncPtr(myFunc<int>) == FuncPtr(myFunc);
+
+  bool b7 = FuncPtr(myDefaultFunc<int>) == FuncPtr(myDefaultFunc<int, int>);
+}
+
+} // namespace GH145415


clang-tools-extra/test/clang-tidy/checkers/misc/redundant-expression-type-aliases.cpp
```diff
@@ -0,0 +1,24 @@
...
+  bool b1 = std::tuple_size<MonthArray>::value == std::tuple_size<ZodiacArray>::value;
+  bool b2 = std::tuple_size_v<MonthArray> == std::tuple_size_v<ZodiacArray>;
+
+  // True positive cases (Should warn):
+  bool b3 = std::tuple_size<MonthArray>::value == std::tuple_size<MonthArray>::value;
+  // CHECK-MESSAGES: :[[@LINE-1]]:{{[0-9]+}}: warning: both sides of operator are equivalent [misc-redundant-expression]
+  
+  bool b4 = std::tuple_size_v<MonthArray> == std::tuple_size_v<MonthArray>;
+  // CHECK-MESSAGES: :[[@LINE-1]]:{{[0-9]+}}: warning: both sides of operator are equivalent [misc-redundant-expression]
+}
```
zeyi2: Please add new line at EOF

clang-tools-extra/test/clang-tidy/checkers/misc/redundant-expression-type-aliases.cpp
```diff
@@ -0,0 +1,24 @@
+// RUN: %check_clang_tidy %s misc-redundant-expression %t -- -- -std=c++17
```
zeyi2: I think we can just:

```suggestion
// RUN: %check_clang_tidy -std=c++17-or-later %s misc-redundant-expression %t
```

clang-tools-extra/test/clang-tidy/checkers/misc/redundant-expression-type-aliases.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+namespace std {
```
zeyi2: Can we use `#include <array>` instead of manual mocking?

zeyi2 (commented):
Please also add a ReleaseNote entry

clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
```diff
@@ -97,9 +98,22 @@ static bool areEquivalentExpr(const Expr *Left, const Expr *Right) {
...
+      return false;
+
+    // Compare source-level spellings to safely differentiate sugared types.
+    std::string LStr, RStr;
+    llvm::raw_string_ostream LOS(LStr), ROS(RStr);
+    
+    L->printPretty(LOS, nullptr, L->getDecl()->getASTContext().getPrintingPolicy());
+    R->printPretty(ROS, nullptr, R->getDecl()->getASTContext().getPrintingPolicy());
+
+    return LStr == RStr;
```
zeyi2: Nit: remove the trailing whitespace

clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
```diff
@@ -97,9 +98,22 @@ static bool areEquivalentExpr(const Expr *Left, const Expr *Right) {
...
+    if (L->getDecl() != R->getDecl())
+      return false;
+
+    // Compare source-level spellings to safely differentiate sugared types.
+    std::string LStr, RStr;
+    llvm::raw_string_ostream LOS(LStr), ROS(RStr);
+    
+    L->printPretty(LOS, nullptr, L->getDecl()->getASTContext().getPrintingPolicy());
+    R->printPretty(ROS, nullptr, R->getDecl()->getASTContext().getPrintingPolicy());
+
```
zeyi2: Nit: please remove the trailing whitespace :)

clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
```diff
@@ -19,6 +19,7 @@
 #include "llvm/ADT/FoldingSet.h"
 #include "llvm/ADT/SmallBitVector.h"
 #include "llvm/Support/FormatVariadic.h"
+#include "llvm/Support/raw_ostream.h"
```
zeyi2, line 22:
```suggestion
```

clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
```diff
@@ -97,9 +98,22 @@ static bool areEquivalentExpr(const Expr *Left, const Expr *Right) {
...
+    const auto *R = cast<DeclRefExpr>(Right);
+
+    if (L->getDecl() != R->getDecl())
+      return false;
+
+    // Compare source-level spellings to safely differentiate sugared types.
+    std::string LStr, RStr;
+    llvm::raw_string_ostream LOS(LStr), ROS(RStr);
+    
+    L->printPretty(LOS, nullptr, L->getDecl()->getASTContext().getPrintingPolicy());
```
zeyi2: Personally I have mixed feelings on the usage of `printPretty`,  by looking at how it is used in other places in `clang-tool-extra`. I think this function is mainly used for printing instead of making semantic decisions. 

In this check, we probably want to compare the source-bearing parts of the `DeclRefExpr` more directly.

voyager-jhk: Thanks for the review and suggestions!

I've addressed the comments:

- AST comparison: Replaced `printPretty` with `Lexer::getSourceText()` on `*Loc` nodes for robust, direct source spelling comparison.

- RUN line: Updated to `-std=c++17-or-later`.

- Formatting: Fixed trailing whitespaces and added the missing EOF newline.

- Release Notes: Added the missing entry.

Please let me know if there's anything else!

clang-tools-extra/test/clang-tidy/checkers/misc/redundant-expression-type-aliases.cpp
voyager-jhk: I tried using it, but it seems `<array>` isn't provided in the `Inputs/Headers` mock directory yet. Decided to keep the manual mock so it remains hermetic and consistent with the existing tests here.

voyager-jhk:
Cc @vgvassilev

clang-tools-extra/test/clang-tidy/checkers/misc/redundant-expression-type-aliases.cpp
```diff
@@ -0,0 +1,26 @@
...
+  // False positive cases (Should NOT warn):
+  bool b1 = std::tuple_size<MonthArray>::value == std::tuple_size<ZodiacArray>::value;
+  bool b2 = std::tuple_size_v<MonthArray> == std::tuple_size_v<ZodiacArray>;
+
+  // True positive cases (Should warn):
+  bool b3 = std::tuple_size<MonthArray>::value == std::tuple_size<MonthArray>::value;
+  // CHECK-MESSAGES: :[[@LINE-1]]:{{[0-9]+}}: warning: both sides of operator are equivalent [misc-redundant-expression]
+
+  bool b4 = std::tuple_size_v<MonthArray> == std::tuple_size_v<MonthArray>;
+  // CHECK-MESSAGES: :[[@LINE-1]]:{{[0-9]+}}: warning: both sides of operator are equivalent [misc-redundant-expression]
```
vbvictor: Please use real line + column numbers

clang-tools-extra/test/clang-tidy/checkers/misc/redundant-expression-type-aliases.cpp
```diff
@@ -0,0 +1,26 @@
...
+}
```
vbvictor: Please add more tests with different formattings/multiline expressions etc..

vbvictor (commented):
This is controversial issue, I'm not super convinced that we should compare raw source text because it can mutate with different formattings/line feeds etc... Could there be a better approach with comparing sugared types of expressions without comparing source text?

clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
```diff
@@ -44,6 +45,328 @@ static bool incrementWithoutOverflow(const APSInt &Value, APSInt &Result) {
   return Value < Result;
 }
 
+static bool areEquivalentExpr(const Expr *Left, const Expr *Right);
```
vgvassilev: Don't we have similar facilities in clang which we can reuse?

voyager-jhk: The `areEquivalentExpr` is actually preexisting in this check, and this introduces mutual recursion because template arguments can contain expressions.

voyager-jhk: For my addition, reusing standard facilities like `Stmt::Profile()` fails because Clang doesn't reliably unique `NestedNameSpecifiers` containing sugared types

voyager-jhk: Manually extracting the `QualType` via `TypeLoc` was the only way I found to preserve the sugar while safely ignoring newlines/formatting. Is there a better existing Clang facility that handles sugared NNS equality?

vgvassilev: Perhaps @mizvekov can give us a hint in that regard.

mizvekov: > For my addition, reusing standard facilities like Stmt::Profile() fails because Clang doesn't reliably unique NestedNameSpecifiers containing sugared types

The current NestedNameSpecifier relies heavily on the existing uniquing for types, so if you are seeing any issues there, they should also exist for regular types.

For example, non-dependent template specialization types are not uniqued in clang ever, and also some other types might suffer from this.

Otherwise, can you give me an example of what problem you are seeing?

voyager-jhk: @mizvekov Thanks for confirming.

To fix the original false positive on `MonthArray == ZodiacArray`, we have to compare the sugared types. However, because they aren't uniqued, standard checks like `Stmt::Profile()` yield different hashes even for identical expressions, causing us to miss true redundancies like:
`std::tuple_size<MonthArray>::value == std::tuple_size<MonthArray>::value;`

To correctly differentiate the aliases while still catching true redundancies, I had to implement a custom deep structural comparison logic in this PR.

Is there an existing Clang API that handles this sugared equivalence natively, so we can drop the custom traversal?

mizvekov: Not, unfortunately. Enabling uniquing for non-dependent template specialization types, which would make your example work, adds about a `0.7%` performance penalty when measured bootstrapping clang.

My plan is to have all this working long term, but there are other improvements that need to be made so that it doesn't introduce performance regressions.

For that reason, the `aka` logic for the type printer in diagnostics does a comparison of the stringfied type.
That may work here and be much simpler.

voyager-jhk: > Not, unfortunately. Enabling uniquing for non-dependent template specialization types, which would make your example work, adds about a `0.7%` performance penalty when measured bootstrapping clang.
> 
> My plan is to have all this working long term, but there are other improvements that need to be made so that it doesn't introduce performance regressions.
> 
> For that reason, the `aka` logic for the type printer in diagnostics does a comparison of the stringfied type. That may work here and be much simpler.

Thanks for the context! I'll try it.

voyager-jhk:
@vbvictor @zeyi2 Following mizvekov's advice, I now use the AST Printer to compare the NestedNameSpecifier and TemplateArguments.

This safely preserves the sugared types, while natively normalizing all whitespaces and newlines. PTAL!

clang-tools-extra/test/clang-tidy/checkers/misc/redundant-expression-type-aliases.cpp
zeyi2, line 1: Nit: Could we move these tests into existing test files?

zeyi2 (commented):
Please update the PR description :)

clang-tools-extra/test/clang-tidy/checkers/misc/redundant-expression-type-aliases.cpp
voyager-jhk: Done. Thanks!

clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
```diff
@@ -97,9 +98,46 @@ static bool areEquivalentExpr(const Expr *Left, const Expr *Right) {
       return false;
     return cast<DependentScopeDeclRefExpr>(Left)->getQualifier() ==
            cast<DependentScopeDeclRefExpr>(Right)->getQualifier();
-  case Stmt::DeclRefExprClass:
-    return cast<DeclRefExpr>(Left)->getDecl() ==
-           cast<DeclRefExpr>(Right)->getDecl();
+  case Stmt::DeclRefExprClass: {
+    const auto *L = cast<DeclRefExpr>(Left);
```
vbvictor: Please refactor into separate function similar to areEquivalentExpr

clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
```diff
@@ -97,9 +98,46 @@ static bool areEquivalentExpr(const Expr *Left, const Expr *Right) {
...
+    const auto *R = cast<DeclRefExpr>(Right);
+
+    if (L->getDecl() != R->getDecl() || L->getFoundDecl() != R->getFoundDecl())
```
vbvictor: What is getFoundDecl for?

clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
```diff
@@ -97,9 +98,46 @@ static bool areEquivalentExpr(const Expr *Left, const Expr *Right) {
...
+        std::string LArg, RArg;
+        llvm::raw_string_ostream LOS(LArg), ROS(RArg);
+        L->getTemplateArgs()[I].getArgument().print(Policy, LOS, true);
+        R->getTemplateArgs()[I].getArgument().print(Policy, ROS, true);
+        if (LArg != RArg)
+          return false;
+      }
+    }
+    return true;
+  }
```
vbvictor: Please write test cases for each code path introduced here.
If it can't be covered with a test (so not visible to user in any way) - then it should be removed

voyager-jhk: I've updated it. Thanks!

clang-tools-extra/clang-tidy/misc/RedundantExpressionCheck.cpp
```diff
@@ -44,6 +45,42 @@ static bool incrementWithoutOverflow(const APSInt &Value, APSInt &Result) {
...
+    if (L->getNumTemplateArgs() != R->getNumTemplateArgs())
+      return false;
+    for (unsigned I = 0, E = L->getNumTemplateArgs(); I != E; ++I) {
+      std::string LArg, RArg;
+      llvm::raw_string_ostream LOS(LArg), ROS(RArg);
+      L->getTemplateArgs()[I].getArgument().print(Policy, LOS, true);
+      R->getTemplateArgs()[I].getArgument().print(Policy, ROS, true);
+      if (LArg != RArg)
+        return false;
+    }
```
zwuis: Use `llvm::equal`.

zwuis (commented):
LGTM but please wait for other reviewers.

zeyi2 (approved):
I found a false negative:

```
namespace N {
int Value;
}
using N::Value;

bool Result = N::Value == Value;
```

Since FNs are better than FPs, I don't think this would be a blocker, but could you please add it to the test cases and add a `// FIXME` comment?

You can also try fixing it in this PR, as I think the way to solve this may be straightforward.

voyager-jhk:
> I found a false negative:
> 
> ```
> namespace N {
> int Value;
> }
> using N::Value;
> 
> bool Result = N::Value == Value;
> ```
> 
> Since FNs are better than FPs, I don't think this would be a blocker, but could you please add it to the test cases and add a `// FIXME` comment?
> 
> You can also try fixing it in this PR, as I think the way to solve this may be straightforward.

Fixed the FN, thanks.

zeyi2 (approved):

PR #198088, opened 2026-05-16
[lldb] Include synthetic variables in `GetVariables(bool...)`

Nerixyz:
Synthetic frame variables can be included in `SBFrame::GetVariables` by passing an `SBVariablesOptions` with `SetIncludeSynthetic(true)`. However, they're not included in the other two overloads.
With the same reason they're included in `frame variable`, they should be included in these overloads. If the user doesn't want them included there, they can use the overload taking `SBVariablesOptions`.

This almost enables them to be shown in lldb-dap's variables. They're currently filtered out if `in_scope_only` is specified. Not sure what the proper fix is. Maybe they should skip that check.


Nerixyz:
Ping

medismailben (approved):
LGTM

PR #198091, opened 2026-05-16
[SPIR-V] Support 1x1 result in matrix multiply combiner

aobolensk:
Emit a copy instead of G_BUILD_VECTOR for the scalarized 1-element result, and skip extractelement when storing a 1-element vector to a [1 x T] array

Enable the previously-disabled 1x2 * 2x1 test


MrSidims (approved):
The patch is LGTM, but would be nice to get approval from somebody more familiar with HLSL matrices.

farzonl:
> The patch is LGTM, but would be nice to get approval from somebody more familiar with HLSL matrices.

This shouldn't impact HLSL, we scalarize 1x1 matrices before we touch any backend.

farzonl (approved):
LGTM

Issue #198060, opened 2026-05-16
[X86] Missed Optimization: Vector 8-bit `rotr(x, 1)` should be lowered as `pavgb(x, -(x & 1))`


WalterKruger:
Due to a lack of support, most 8-bit shifts are implemented using a 16-bit shift + AND:

```asm
rotr1_src:
        movdqa  xmm1, xmm0
        psrlw   xmm1, 1
        pand    xmm1, xmmword ptr [rip + .LCPI1_0]
        psllw   xmm0, 7
        pand    xmm0, xmmword ptr [rip + .LCPI1_1]
        por     xmm0, xmm1
        ret
```

The right shift and least significant bit propagation can be done using the `pavgb` instruction, which performs a ceiling average. It can be used to shift right by 1 and then conditionally set the MSB based on the mask input (as the false case has a zero LSB):

```asm
rotr1_tgt:
        movdqa  xmm1, xmmword ptr [rip + .LCPI1_0]
        pand    xmm1, xmm0
        pxor    xmm2, xmm2
        psubb   xmm2, xmm1
        pavgb   xmm0, xmm2
        ret
```

https://godbolt.org/z/scsce9YTE

This uses less operations and avoids the shift, which has worse throughput than `pavgb` on some architectures.

RKSimon:
Relatively easy to add to LowerRotate - along with tests coverage

First step for an assignee would by writing an alive2 proof to demonstrate they understand the task

michael1017:
Hi I would like to take this issue

michael1017:
@RKSimon Here is the Alive2 proof demonstrating equivalence.

In [this example](https://github.com/llvm/llvm-project/issues/198060#issue-4458467605), we use 16-bit instructions to do 8-bit shifts for 8-bit variables. Therefore, I used 16-bit in my Alive2 as well.

https://alive2.llvm.org/ce/z/ixWVfS

RKSimon:
I'm more interested in demonstrating the tgt codegen matches:
```
define i8 @src(i8 %x) {
    %r = call i8 @llvm.fshr.i8(i8 %x, i8 %x, i8 1)
    ret i8 %r
}
```

michael1017:
Thanks for the feedback! I have updated the tgt codegen implementation.

Note: To calculate `pavgb  xmm0, xmm2`, I used a i9 variable to avoid overflow

https://alive2.llvm.org/ce/z/ZhRbmR

Issue #198059, opened 2026-05-16
[X86] Missed Optimization: Vector 8-bit `rotl(x, 1)` should be lowered as `(x + x) - (x < 0)`


WalterKruger:
Due to a lack of support, most 8-bit shifts are implemented using a 16-bit shift + AND:

```asm
rotl1_src:
        movdqa  xmm1, xmm0
        paddb   xmm1, xmm0
        psrlw   xmm0, 7
        pand    xmm0, xmmword ptr [rip + .LCPI2_0]
        por     xmm0, xmm1
        ret
```

The OR and right shift can be replaced with a subtraction by a less-than-zero mask, which acts like a conditional disjoint add by 1. This shortens the dependency chain and avoids the shift, which has worse throughput on some architectures.

```asm
rotl1_tgt:
        pxor    xmm1, xmm1
        pcmpgtb xmm1, xmm0
        paddb   xmm0, xmm0
        psubb   xmm0, xmm1
        ret
```

https://godbolt.org/z/199KoWhs8

Ippo47:
hi i would like to take this issue

RKSimon:
Easy enough to implement in LowerRotate

@Ippo47 please begin by creating a alive2 link to demonstrate you understand the task

ArihantChawla:
Hi, if no one is working on this I would like to contribute 

Alive2 link: https://alive2.llvm.org/ce/z/teP3my

RKSimon:
@Ippo47 is still working on this - please give them time to respond, and maybe find another ticket that doesn't already have an assignee?

Ippo47:
@RKSimon @WalterKruger sorry for late reply, here is my Alive2 link https://alive2.llvm.org/ce/z/FrSxNt

RKSimon:
@lppo47 any luck? You might want to take a look at #198980 for reference

Ippo47:
> @lppo47 any luck? You might want to take a look at [#198980](https://github.com/llvm/llvm-project/pull/198980) for reference

sure, thanks for this

AnkitDubeycs25:
can i take this issue ??

minicusprime:
@Ippo47 is still working on it!! 
@AnkitDubeycs25

VachanVY:
@RKSimon Issue was inactive for over a month with no development, so I took it. Please review.

PR #198095, opened 2026-05-16
[X86] Add vXi8 rot-by-one tests

RKSimon:
Test coverage for #198059 and #198060


Andarwinux:
GFNI coverage?

RKSimon:
Nice catch - we already have partial rot/ashr by 1 test coverage in gfni-rotates.ll and gfni-shifts.ll - my guess is we have sufficient already, but we can tweak anything based what the patches bring up

PR #198099, opened 2026-05-16
Fix build with shared libraries

kparzysz:
Add missing library dependencies in llvm-ir2vec.


svkeerthy:
Guess https://github.com/llvm/llvm-project/pull/198077 addresses the issue.

PR #198102, opened 2026-05-16
[offload] Add new features to libompaccsupport for OpenACC

ivanradanov:
AsyncInfoTy STATIC_NON_BLOCKING type.

Strided array copies and mapping.

No create mapping type.

Refactoring intialization.

Loading offload objects with OpenACC offloading kind.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/offload/libompaccsupport/Mapping.cpp
+++ b/offload/libompaccsupport/Mapping.cpp
@@ -14,6 +14,10 @@
 #include "Shared/Debug.h"
 #include "Shared/Requirements.h"
 #include "device.h"
+#include "omptarget.h"
+#include <cstdint>
+#include <optional>
+#include <variant>
 
 using namespace llvm::omp::target::debug;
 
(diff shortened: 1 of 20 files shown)


offload/libompaccsupport/Mapping.cpp
```diff
@@ -206,12 +210,22 @@ LookupResult MappingInfoTy::lookupMapping(HDTTMapAccessorTy &HDTTMap,
 
 TargetPointerResultTy MappingInfoTy::getTargetPointer(
     HDTTMapAccessorTy &HDTTMap, void *HstPtrBegin, void *HstPtrBase,
-    int64_t TgtPadding, int64_t Size, map_var_info_t HstPtrName, bool HasFlagTo,
-    bool HasFlagAlways, bool IsImplicit, bool UpdateRefCount,
-    bool HasCloseModifier, bool HasPresentModifier, bool HasHoldModifier,
+    int64_t TgtPadding, std::variant<int64_t, const NonContigDescTy *> MemInfo,
```
ivanradanov, line 213: I am thinking of having the NonContigDescTy being a separate optional argument which is used for the copy and not the allocation instead of std::variant of the two options

PR #198111, opened 2026-05-16
[flang-rt][NFC] Split up expensive .cpp files into multiple files

jhuber6:
Summary:
This PR simply takes the existing `.cpp` files for the heaviest
implementations and separates them logically, typically between real,
integer, and complex types. The existing `.cpp` file is turned into a
`.h` file and we create new `.cpp` files that *only* contain the old
portion that used `RTDEF`. This allows for far more build system
parallelism, and it also means that static library linking semantics
mean that if the user only uses integer routines the linker will not
include the unused complex / real routines in the final executable.

All around this is a good practice for runtime libraries. Verified that
all `_Fortran` entrypoint routines are still present, port was strictly
mechanical.

The result of all of this is that I can now build `flang-rt` in ~10s
with all threads instead of ~50s due to the most expensive files being
split into parallelizable chunks.


flang-rt/lib/runtime/matmul-transpose.cpp
```diff
@@ -20,39 +20,21 @@
...
-//
-// With loop distribution and transposition to avoid the inner sum
-// reduction and to avoid non-unit strides:
-//   DO 1 I = 1, NROWS
-//    DO 1 J = 1, NCOLS
-//   1 RES(I,J) = 0
-//   DO 2 J = 1, NCOLS
-//    DO 2 I = 1, NROWS
-//     DO 2 K = 1, N
-//   2  RES(I,J) = RES(I,J) + X(K,I)*Y(K,J) ! loop-invariant last term
```
Meinersbur, line 55: Why removing all implementation comments?

flang-rt/lib/runtime/matmul-transpose.h
```diff
@@ -369,10 +322,4 @@ RT_EXT_API_GROUP_BEGIN
         YKIND>{}(result, x, y, sourceFile, line, false); \
   }
 
-#define MATMUL_FORCE_ALL_TYPES 0
-
-#include "flang/Runtime/matmul-instances.inc"
```
Meinersbur, line 374: Can this file be deleted then?

Meinersbur (commented):
Note that a sequential build will take longer, but IMHO that be well worth it. Do you have numbers on user time in addition to real time (I am curious about the template instantiation cost)?

> The result of all of this is that I can now build flang-rt in ~10s
with all threads instead of ~50s due to the most expensive files being
split into parallelizable chunks.

This is only meaningful if you also mention how many threads the the system has.

flang-rt/lib/runtime/matmul-transpose.h
jhuber6: Could probably be merged.

flang-rt/lib/runtime/matmul-transpose.cpp
jhuber6: Unintentional cut & paste, I'll put them back.

jhuber6:
> Note that a sequential build will take longer, but IMHO that be well worth it. Do you have numbers on user time in addition to real time (I am curious about the template instantiation cost)?
> 
> > The result of all of this is that I can now build flang-rt in ~10s
> > with all threads instead of ~50s due to the most expensive files being
> > split into parallelizable chunks.
> 
> This is only meaningful if you also mention how many threads the the system has.

The system I tested this on has 48. Agreed on the value, because the overhead associated with launching the compiler a few dozen more times is negligible next to the single-threaded compilation already taking >4 minutes. This was on my machine with a `ninja clean`.
```
> ninja clean && ninja
[96/96] Linking CXX static library lib/clang/23/lib/x86_64-unknown-linux-gnu/libflang_rt.runtime.a
ninja  339.95s user 4.83s system 2125% cpu 16.220 total
```

flang-rt/lib/runtime/matmul-transpose.h
jhuber6: Nope, seems to be used by
```
../flang/lib/Optimizer/Builder/Runtime/Transformational.cpp
350:#include "flang/Runtime/matmul-instances.inc"
379:#include "flang/Runtime/matmul-instances.inc"
405:#include "flang/Runtime/matmul-instances.inc"
430:#include "flang/Runtime/matmul-instances.inc"
```

Meinersbur (approved):
LGTM

Please give others a chance to comment before landing

vzakhari (approved):
Thank you!

jhuber6:
Thanks for the review. For reference, this is before this PR:|
```
> ninja clean && ninja
[77/77] Linking CXX static library lib/clang/23/lib/x86_64-unknown-linux-gnu/libflang_rt.runtime.a
ninja  277.17s user 3.60s system 518% cpu 54.154 total
```

PR #198112, opened 2026-05-16
[DWARFVerifier] Don't require .dwo for non-skeleton CUs

JDevlieghere:
The presence of `DW_AT_(GNU_)dwo_id` alone does not make a compile unit a skeleton: a regular `DW_TAG_compile_unit` is allowed to carry a DWO id. DWARF 5 marks skeleton CUs with `DW_TAG_skeleton_unit` while older pre DWARF 5 identifies them by absence of children.

Before this change, the verifier would any CU whose header reported a DWO id as a skeleton, then tried and failed to load its (non-existent) .dwo and reported an error. This updates the verifier to gate the check on the CU being an actual skeleton CU.

### DIFF
--- a/llvm/lib/DebugInfo/DWARF/DWARFVerifier.cpp
+++ b/llvm/lib/DebugInfo/DWARF/DWARFVerifier.cpp
@@ -1773,11 +1773,8 @@ void DWARFVerifier::verifyNameIndexEntries(
     // call to properly deal with it. It isn't clear that getNonSkeletonUnitDIE
     // will return the unit DIE of DU if we aren't able to get the .dwo file,
     // but that is what the function currently does.
-    // A DW_AT_(GNU_)dwo_id alone does not make a CU a skeleton: a regular
-    // DW_TAG_compile_unit is allowed to carry a DWO id (e.g. -gmodules emits
-    // one on each module CU). A CU is a skeleton CU only when DWARF 5+ tags
-    // it as DW_TAG_skeleton_unit, or when, in older DWARF, the CU has no
-    // children because the body lives in the .dwo.
+    // A CU is a skeleton CU only when DWARF 5+ tags it as DW_TAG_skeleton_unit,
+    // or when, in older DWARF, the CU has no children.
     DWARFDie UnitDie = DU->getUnitDIE();
     auto IsSkeletonCU = [&]() {
       if (DU->getVersion() >= 5)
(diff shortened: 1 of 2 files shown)


dwblaikie:
Might be good to discuss this a bit further, but functionally it's probably fine.

Dwarf4 skeletons (& 5) could have children - in the case of split dwarf inlining (though this was never spec'd - two level line tables were/are/maybe hoped to remove the need for split dwarf inlining)

But yeah, a skeleton unit is a fair signal & maybe dwarf 4 split dwarf can bitrot a little.

But what novel use of dwo_id do you have in mind in this direction?

JDevlieghere:
> But what novel use of dwo_id do you have in mind in this direction?

No novel use. This PR is the result of discussion with @adrian-prantl in https://github.com/llvm/llvm-project/pull/198047#discussion_r3251605145. As you know, I'm working on parity between the classic and parallel DWARF linker and the latter wasn't stripping `DW_AT_(GNU_)dwo_id` from the cloned unit. This tripped up the verifier, so I updated the parallel linker to behave the same. But Adrian pointed out that technically, keeping the attribute isn't wrong. While there's no reason to keep it in dsymutil, he argued (and I agree) that we should update the verifier too.

llvm/lib/DebugInfo/DWARF/DWARFVerifier.cpp
```diff
@@ -1773,15 +1773,26 @@ void DWARFVerifier::verifyNameIndexEntries(
     // call to properly deal with it. It isn't clear that getNonSkeletonUnitDIE
     // will return the unit DIE of DU if we aren't able to get the .dwo file,
     // but that is what the function currently does.
+    // A DW_AT_(GNU_)dwo_id alone does not make a CU a skeleton: a regular
+    // DW_TAG_compile_unit is allowed to carry a DWO id (e.g. -gmodules emits
+    // one on each module CU). A CU is a skeleton CU only when DWARF 5+ tags
```
adrian-prantl: This is not just a -gmodules thing, the actual DWO compile units pointed to by skeleton units are expected to have a DWO_id, too.

adrian-prantl (approved):
LGTM (though the comment seems to be too specific)

dwblaikie:
> > But what novel use of dwo_id do you have in mind in this direction?
> 
> No novel use. This PR is the result of discussion with @adrian-prantl in [#198047 (comment)](https://github.com/llvm/llvm-project/pull/198047#discussion_r3251605145). As you know, I'm working on parity between the classic and parallel DWARF linker and the latter wasn't stripping `DW_AT_(GNU_)dwo_id` from the cloned unit. This tripped up the verifier, so I updated the parallel linker to behave the same. But Adrian pointed out that technically, keeping the attribute isn't wrong. While there's no reason to keep it in dsymutil, he argued (and I agree) that we should update the verifier too.

looked further, I think it's a moot point - since this is only in the accelerator table validation, and there aren't any accelerator tables in DWARFv4+Split DWARF... so the DWARFv4+Split DWARF+Split DWARF Inlining situation (that would create a GNU_dwo_id on a compile_unit with children, while still having a split file) won't come up so far as I know.

(that said, wouldn't the dwo_name be the thing we'd use to determine whether to open a dwo file?)

PR #198113, opened 2026-05-16
[mlir][linalg] Add constant folding for linalg.generic arith.extsi casts

Jimmy2027:
Folds a `linalg.generic` whose body is a single `arith.extsi` consuming a constant input block argument, replacing the op with the sign-extended `arith.constant`.

Had to move the check for matching input and output types to a new abstract method `checkElementTypes`.

### DIFF
--- a/mlir/lib/Dialect/Linalg/Transforms/ConstantFold.cpp
+++ b/mlir/lib/Dialect/Linalg/Transforms/ConstantFold.cpp
@@ -10,6 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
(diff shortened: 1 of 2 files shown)


Jimmy2027:
This would probably be a good case for the linalg.elementwise op, but it currently does not support sign extension ops. @rengolin would it make sense to add that?

rengolin:
> This would probably be a good case for the linalg.elementwise op, but it currently does not support sign extension ops. @rengolin would it make sense to add that?

It would. @javedabsar1 @adam-smnk

Jimmy2027:
I'm actually having second thoughts on adding this to the default pattern. Some usecases might prefer to pay the performance cost of doing the cast at runtime to save memory space. 

I think a separate `populateFoldConstantCastPatterns` pattern would make sense here which can be applied via `transform.apply_patterns.linalg.fold_constant_casts` and optionally add a pass option `fold-constant-casts` to `linalg-fuse-elementwise-ops`. 
Does anyone have some thoughts on this?

PR #198114, opened 2026-05-16
[InferAlignment] Propagate alignment information across blocks

dtcxzyw:
This patch implements DT-based cross-block alignment propagation. The algorithm processes the basic blocks in a depth-first traversal of DT. The stack is emulated to avoid recursive visiting.

IR diff: https://github.com/dtcxzyw/llvm-opt-benchmark-nightly/pull/321. Most of the improvement is useless, as it makes load/store instructions overaligned. But several alignment checks are eliminated: https://github.com/dtcxzyw/llvm-opt-benchmark-nightly/pull/321/commits/2ac31cb4f44261fa256d23061346af3aac73a41d.

Compile-time impact: +0.03%
https://llvm-compile-time-tracker.com/compare.php?from=f45f3ce5da9e35f819db9b8ed2775dac8c2b1171&to=93d511bc2a64792fcdb8d84b0b696a7c8d61ef40&stat=instructions%3Au


llvm/lib/Transforms/Scalar/InferAlignment.cpp
```diff
@@ -130,38 +143,46 @@ bool inferAlignment(Function &F, AssumptionCache &AC, DominatorTree &DT) {
     Align BasePointerAlign =
         commonAlignment(LoadStoreAlign, OffsetFromBase.getLimitedValue());
 
-    auto [It, Inserted] =
-        BestBasePointerAligns.try_emplace(PtrOp, BasePointerAlign);
-    if (!Inserted) {
+    if (auto BestAlign = BestBasePointerAligns.lookup(PtrOp);
```
nikic, line 147: It probably makes sense to support try_emplace on ScopedHT. From a quick look at the implementation, I don't see a technical reason why we can't have it, and then avoid duplicate lookups in cases like this.

llvm/test/Transforms/InferAlignment/propagate-across-blocks.ll
nikic, line 1: Should also test that it's not incorrectly propagated to blocks that are not dominated?

nikic (approved):
LGTM

llvm/lib/Transforms/Scalar/InferAlignment.cpp
```diff
@@ -121,7 +134,7 @@ bool inferAlignment(Function &F, AssumptionCache &AC, DominatorTree &DT) {
 
   // Propagate alignment between loads and stores that originate from the
   // same base pointer.
-  DenseMap<Value *, Align> BestBasePointerAligns;
+  ScopedHT BestBasePointerAligns;
```
nikic, line 138: Should we pass the Allocator to the ScopedHT as well? It looks like it defaults to MallocAllocator otherwise. Or are we concerned that the memory usage from not freeing the entries until the end will be problematic?

llvm/lib/Transforms/Scalar/InferAlignment.cpp
```diff
@@ -130,38 +143,46 @@ bool inferAlignment(Function &F, AssumptionCache &AC, DominatorTree &DT) {
...
 
-  for (BasicBlock &BB : F) {
-    // We need to reset the map for each block because alignment information
-    // can only be propagated from instruction A to B if A dominates B.
-    // This is because control flow (and exception throwing) could be dependent
-    // on the address (and its alignment) at runtime. Some sort of dominator
-    // tree approach could be better, but doing a simple forward pass through a
-    // single basic block is correct too.
-    BestBasePointerAligns.clear();
+  BumpPtrAllocator Allocator;
```
andjo403: maybe add comment about why BumpPtrAllocator is used is it due to AlignmentScope is not copy or some optimization

llvm/lib/Transforms/Scalar/InferAlignment.cpp
dtcxzyw: I investigated all ScopedHT users with the lookup-then-insert pattern:
> ScopedHTType AvailableValues;
LoadHTType AvailableLoads;
InvariantHTType AvailableInvariants;
CallHTType AvailableCalls;
GEPHTType AvailableGEPs;
ScopedHashTable<const SCEV *, LoadValue> AvailableLoads;
llvm::ScopedHashTable<StringRef, char> usedNames;
ValueMapper valueMapper;
BlockMapper blockMapper;
ScopedMapTy knownValues;

However, most of them are a bit different from try_emplace, as the insertion is always performed. We may need to add a method to expose `ScopedHashTableVal<K, V>&` to save one lookup in TopLevelMap.

llvm/test/Transforms/InferAlignment/propagate-across-blocks.ll
dtcxzyw: Isn't it covered by the edge `if.then -> if.else`?

llvm/lib/Transforms/Scalar/InferAlignment.cpp
dtcxzyw: The logic was copied from EarlyCSE. I have switched to std::list. Hopefully the compile-time impact doesn't get worse.

dtcxzyw: Decreased to +0.03%

llvm/lib/Transforms/Scalar/InferAlignment.cpp
antoniofrighetto: > However, most of them are a bit different from try_emplace, as the insertion is always performed. We may need to add a method to expose `ScopedHashTableVal<K, V>&` to save one lookup in TopLevelMap.

Hmm, wouldn't exposing `ScopedHashTableVal<K, V>&` still require to have `TopLevelMap[Key]` pointed to the new node (i.e., a call to insert())? It sounds like to me more that we might need something, e.g., `insertIf(const K &Key, const V &NewVal, Pred ShouldInsert)` to avoid the look up in the call-site, as a future follow-up.

Issue #61440, opened 2023-03-15
error: cannot mangle this template type parameter type yet


steeeeeeleydan:
While attempting to generate AST with Clang 15.0.7 with code that includes headers from MSVC 17.3.5, Clang crashes.

Attached is a zip file containing:
1. `main.cpp`, a small program that can generate the Clang crash
2. `build.bat`, shows the command line used on my system to generate the crash
3. `err`, output generated from running `build.bat` on my system
4. `msvc-17.3.5`, copies of the entire source files from which `main.cpp` was pieced together, included here for context (if necessary)

Below is the stack dump generated from the crash, pasted here for convenience:

```console
Stack dump:
0.	Program arguments: "c:\\Program Files\\LLVM\\15.0.7\\bin\\clang.exe" -v -c -Xclang -ast-dump=json main.cpp
1.	<eof> parser at end of file
2.	main.cpp:55:27: Mangling declaration '_Is_member_object_pointer<type-parameter-0-0 type-parameter-0-1::*>::value'
Exception Code: 0xC0000005
 #0 0x00007ff6c5da418d c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x2bd418d c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x2bde0c6
 #1 0x00007ff6c5da418d c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x2bd9086 c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x2bd62b3
 #2 0x00007ff6c5da418d c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x2bd5ce8 c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x2bd3b12
 #3 0x00007ff6c5da418d c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x2bd2b70 c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x2bce1f5
 #4 0x00007ff6c5da418d c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x283b516 c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x283d58e
 #5 0x00007ff6c5da418d c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x283ca83 c:\Program Files\LLVM\15.0.7\bin\clang.exe 0x283cbd9
[...]

PR #198116, opened 2026-05-16
[clang] Refactor some JSON dump tests

Endilll:
This PR reduces the amount of JSON dumping our tests do:
- `ast-dump-default-arg-json.cpp` is folded into `ast-dump-for-range-lifetime.cpp`.
- `ast-dump-default-init-json.cpp` is folded into `ast-dump-default-init.cpp`.
- `ast-dump-lambda-json.cpp` is merged with `ast-dump-lambda.cpp`.
- Test case in `ast-dump-template-json-win32-mangler-crash.cpp` comes from #61440. I reduced it further.


AaronBallman (approved):
> Lines changed: 669 additions & 9923 deletions

You love to see it. :-)

LGTM, thank you; these tests are fragile and what they're testing is best-effort to begin with, so it's good to reduce the coverage like this IMO.

PR #189739, opened 2026-03-31
[LV] InstCombine-simplify partial-reduce-sub dotp test

artagnon:
The input IR shown in the partial-reduce-sub dotp test will never reach the vectorizer, due to prior simplification by InstCombine. Change it with this simplification. This is necessary for the simplification in #187228, whose goal is to make #187589 possible.

PR #198118, opened 2026-05-16
[LV] Update stale comment for partial reduction operands (NFC)

MacDue:
The `neg` form was removed in #187228 (this case now uses the out-of-loop sub, which is preferable, see #189739).


artagnon (approved):
LGTM, thanks!

PR #180919, opened 2026-02-11
[CIR][NFC] Consolidate AArch64 tests

banach-space:
As initiated in #180597 (*), ensure that CIR tests reuse the default
Clang lowering tests. To this end, consolidate the AArch64 tests for
unconstrained -fullfp16 builtins into:

  * clang/test/CodeGen/AArch64/neon/fullfp16.c

(*) See also https://github.com/llvm/llvm-project/issues/179952 for
more context.

PR #198121, opened 2026-05-16
[CIR][AMDGPU] Add support for AMDGCN frexp_mant builtins

ayokunle321:
Adds codegen for the following AMDGCN frexp mantissa builtins:

- __builtin_amdgcn_frexp_mant (double)
- __builtin_amdgcn_frexp_mantf (float)
- __builtin_amdgcn_frexp_manth (half)

These are lowered to the corresponding `llvm.amdgcn.frexp.mant` intrinsic.


clang/test/CIR/CodeGenHIP/builtins-amdgcn-vi.hip
```diff
@@ -0,0 +1,65 @@
+#include "../CodeGenCUDA/Inputs/cuda.h"
```
andykaylor: Would `#define __device__ __attribute__((device))` be sufficient to avoid this?

clang/test/CIR/CodeGenHIP/builtins-amdgcn-vi-f16.hip
andykaylor, line 1: The name of this test doesn't match the name of the classic codegen test. It would be nice to keep those in sync. In fact, it would be really nice to have these share a single test as @banach-space has been doing with the AArch64 tests. See, for example, https://github.com/llvm/llvm-project/pull/180919. I realize that may be more than you intended to sign yourself up for.

@arsenm What would you think of consolidating these tests as we add CIR support?

arsenm: I'd hope they are just copies of the original tests. That makes it much easier to delete or merge one of the copies in the future

andykaylor: So you'd prefer to have a second test that was a copy of the original test rather than trying to combine them now?

ayokunle321: It matches OG now. Could always consolidate them if we do decide to go ahead in that direction.

ranapratap55 (approved):
LGTM.

Resolve merge conflicts.

PR #198128, opened 2026-05-16
Enforce `noinline` at call sites

dingxiangfei2009:
_Related to #45808._

[LangRef](https://llvm.org/docs/LangRef.html#:~:text=to%20be%20generated.-,noinline,-This%20attribute%20indicates) literally promises that a call on a function annotated with noinline will be preserved.

If we take this interpretation, it entails that even when there is a rewrite opportunity given a `returned` argument, inlining should never happen. `returned` is just an optimisation hint at best and semantics always come first.


efriedma-quic (commented):
This transform technically isn't inlining: zero lines of IR from the callee end up in the caller.  The call is just a no-op.

See also #40819.

nikic (changes requested):
Right, the consensus is that noinline should affect inlining and inlining only, not other inter-procedural optimizations. A separate noipa attribute would be needed for that (which has been floated a few times, but never implemented).

dingxiangfei2009:
@efriedma-quic @nikic Ah I see. Right I agree that we would just use `noipa` for this kind of control.

I will have to catch up with `noipa` conversation there.

cc @davidben

PR #198130, opened 2026-05-16
[clang] Warn on umask() argument bits outside 0777

nuclearcat:
Add a -Wfortify-source warning when the constant-evaluated argument to umask(mode_t) has bits set outside 0777. Those bits are silently discarded by the kernel, so setting them is almost always a typo (0xFFFF, 7777-as-decimal, etc.).

Match the corresponding bionic libc diagnose_if check.

The Sema-side dispatch is gated to identify the libc declaration of umask and reject user-supplied lookalikes: at least one redeclaration of the resolved function must come from a system header. That is where libc declares umask.

Matching on the `mode_t` typedef name is not used because libcs disagree on the spelling: glibc's <sys/types.h> writes the prototype as `__mode_t umask(__mode_t)` via an internal typedef, bionic and musl use `mode_t` directly, and so on. The system-header origin is the portable libc-identity signal; coarser shape checks (name + extern "C" + non-variadic + 1 integer arg + integer return) fill in the remaining filtering.

The negative tests cover:
  * `int umask(int)` in user code,
  * `typedef unsigned mode_t; extern mode_t umask(mode_t);` in user code (matches the libc typedef name but not the system-header origin).

A positive test mirrors glibc's `__mode_t umask(__mode_t)` shape from a `#pragma GCC system_header`-marked Inputs/ header to lock the gate against regressions on the spelling.

No Static Analyzer summary is added: modeling umask's argument as a WithinRange precondition would prune feasible caller branches such as `umask(m); if (m > 0777) ...`, even though umask is defined behavior for any input (the kernel masks off the high bits). The Sema check covers the constant-typo case that bionic's diagnose_if targets.


clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -971,6 +971,10 @@ def warn_fortify_strlen_overflow: Warning<
   " but the source string has length %2 (including NUL byte)">,
   InGroup<FortifySource>;
 
+def warn_fortify_umask_unused_bits : Warning<
+  "'umask' argument has bits outside 0777 (0%0); those bits are ignored">,
```
shafik: I don't know if I am crazy about the wording, it feels confusing to says "bits outside of 0777" which is not itself bit values.

to268: I agree the diagnostic is confusing, only the file permissions bits of `mode_t` are used by `umask`.

From [man 3 umask](https://linux.die.net/man/3/umask):
> Only the file permission bits of cmask (see <sys/stat.h>) are used; the meaning of the other bits is implementation-defined.

While saying that "bits outside of 0777" is correct, I would probably want to see permissions bits in the diagnostic message.

clang/lib/Sema/SemaChecking.cpp
```diff
@@ -1490,6 +1490,52 @@ void Sema::checkFortifiedBuiltinMemoryFunction(FunctionDecl *FD,
...
+    SourceManager &SM = getSourceManager();
+    for (const FunctionDecl *R : F->redecls())
+      if (SM.isInSystemHeader(R->getLocation()))
+        return true;
+    return false;
+  };
+  if (II->isStr("umask") && !FD->isVariadic() && TheCall->getNumArgs() == 1 &&
+      FD->getNumParams() == 1 &&
+      FD->getParamDecl(0)->getType()->isIntegerType() &&
+      FD->getReturnType()->isIntegerType() && AnyDeclInSystemHeader(FD)) {
```
to268: Some of these checks are unnecessary, if the function is umask and it comes from a system header that should be enough to know that that's the function we are looking for.

nuclearcat: I wasn’t confident about the edge cases here, so I looked into it after your comment. 
You are right that extra checks are redundant. Thanks for pointing this out.
Fixing that.

clang/include/clang/Basic/DiagnosticSemaKinds.td
nuclearcat: Sorry, original wording was awkward. English is not my native language, so please let me know if the wording in the amended patch reads better.

clang/lib/Sema/SemaChecking.cpp
```diff
@@ -1490,6 +1490,49 @@ void Sema::checkFortifiedBuiltinMemoryFunction(FunctionDecl *FD,
                           << FunctionName << DestinationStr << SourceStr);
 }
 
+void Sema::checkFortifiedLibcArgument(FunctionDecl *FD, CallExpr *TheCall) {
+  if (TheCall->isValueDependent() || TheCall->isTypeDependent() ||
+      isConstantEvaluatedContext())
```
AaronBallman: Why should the checking be disabled in a constant evaluation context?

clang/lib/Sema/SemaChecking.cpp
```diff
@@ -1490,6 +1490,49 @@ void Sema::checkFortifiedBuiltinMemoryFunction(FunctionDecl *FD,
...
+    return;
+  if (!FD->isExternC())
```
AaronBallman: A test case to consider is this code in C:
```
struct S {
  mode_t umask(mode_t);
};
```
because of this assert: https://github.com/llvm/llvm-project/blob/0c315580e51c2eb901f57f04bf3a9a3aebf0906c/clang/lib/AST/Decl.cpp#L2213

(We'd expect a diagnostic but hopefully we don't then come on to hit this code path and trigger the assert during error recovery.)

clang/lib/Sema/SemaChecking.cpp
```diff
@@ -1490,6 +1490,49 @@ void Sema::checkFortifiedBuiltinMemoryFunction(FunctionDecl *FD,
...
+    return;
+  const IdentifierInfo *II = FD->getIdentifier();
+  if (!II)
+    return;
+
+  // umask(mode_t): warn when the constant-evaluated argument has bits set
+  // outside the file-permission mask (0777). Those bits are ignored.
+  // Require a matching system-header declaration to avoid warning on
+  // user-defined lookalikes.
+  auto AnyDeclInSystemHeader = [&](const FunctionDecl *F) {
```
AaronBallman: C allows you to declare the function without including the header; is the same true for POSIX? If so, that means folks doing that dance will lose the diagnostic (which is not the end of the world, but worth documenting if we want to keep this requirement).

clang/lib/Sema/SemaChecking.cpp
nuclearcat: There's no good reason. :(
I modeled this function on checkFortifiedBuiltinMemoryFunction and copied its guard line without thinking it well through. Will fix that.

clang/lib/Sema/SemaChecking.cpp
nuclearcat: Yes, POSIX is the same as C here: umask can be forward-declared without <sys/stat.h>, and then this check goes silent.
Other fortify checks dodge this by recognizing functions via getBuiltinID, but umask isn't a builtin, so that path isn't available :(
Will make documentation changes.

clang/lib/Sema/SemaChecking.cpp
nuclearcat: I checked, in C the member becomes a FieldDecl (via err_field_declared_as_function), not a record-context FunctionDecl, so it never reaches this check
adding it as a regression test

clang/lib/Sema/SemaChecking.cpp
AaronBallman: Hmm, I actually wonder whether we should make `umask` a builtin instead; that seems like it will simplify the logic for this PR in addition to giving slightly better behavior. Have you investigated that approach?

nuclearcat: I did look at the builtin route earlier and talked myself out of it, but I think your instinct is right and my reason for avoidin it doesn't really hold up here. Let me lay out the hesitation honestly:

The thing that scared me off was our experience with read/write. We added those as LibBuiltins and it regressed real user code, ppl who declare their own file-scope read/write in C suddenly got builtin behavior they didn't ask for, and we had to back it out and dispatch by name instead. So I had a general "don't turn libc functions into builtins, you'll step on someone" fear/rule in my head and applied it to umask without thinking it through.

But the more I look, the less that fear applies to umask:

- The problem with read/write was that they're generic names,lots of code has a file-scope read/write. umask is a much rarer spelling; the collision surface is tiny (but still exist)
- The builtin attr is only attached to file-scope, C-linkage declarations anyway, so C++ members like Stream::read were never the issue. For umask that gate makes false matches even less likely.

And the builtin route actually fixes the gap you pointed out earlier: since recognition would be by name rather than by "did this come from a system header," the diagnostic would keep working even when someone forward-declares umask, without including <sys/stat.h>. That's much better than what I have now.

On the target-dependent prototype (mode_t is 16-bit on some platforms, 32-bit on others): 
I'd handle that the same way vfork handles pid_t() today, will mark it IgnoreSignature so Clang recognizes it by name and never tries to match the encoded prototype against the libc's real declaration. No per-libc mode_t spelling logic needed.

The one caveat I want to flag: IgnoreSignature turns off signature validation entirely, so the check can't assume anything about the parameter or return types - it has to defensively confirm "one argument, integral constant" before inspecting the bits. The current logic already does exactly that, so it ports over cleanly.

I will prototype the Builtins.td entry and repoint the check through the builtin path, and report back on whether anything in the test suite trips.

P.S. Sorry for long answer

AaronBallman: > The thing that scared me off was our experience with read/write. We added those as LibBuiltins and it regressed real user code, ppl who declare their own file-scope read/write in C suddenly got builtin behavior they didn't ask for, and we had to back it out and dispatch by name instead. So I had a general "don't turn libc functions into builtins, you'll step on someone" fear/rule in my head and applied it to umask without thinking it through.

Hmmm so the builtin logic was not accounting for functions named `read` or `write` with internal linkage and treating them as builtins? Or did these have external linkage (at which point the user has UB because the POSIX library they link against also has external linkage symbols with the same names)? Or something else?

> On the target-dependent prototype (mode_t is 16-bit on some platforms, 32-bit on others):
I'd handle that the same way vfork handles pid_t() today, will mark it IgnoreSignature so Clang recognizes it by name and never tries to match the encoded prototype against the libc's real declaration. No per-libc mode_t spelling logic needed.

I think that's reasonable; we can go the extra mile if we find users have their own `umask` function with a different signature.

> I will prototype the Builtins.td entry and repoint the check through the builtin path, and report back on whether anything in the test suite trips.

SGTM!

> P.S. Sorry for long answer

No worries at all, I appreciate the detailed discussion!

nuclearcat: > Hmmm so the builtin logic was not accounting for functions named `read` or `write` with internal linkage and treating them as builtins? Or did these have external linkage (at which point the user has UB because the POSIX library they link against also has external linkage symbols with the same names)? Or something else?

Good question, I went and checked rather than trusting my (bad, due stress) memory, and it's narrower than "treated as builtins". It wasn't internal linkage, and it wasn't codegen or UB. It was purely a front-end diagnostic: -Wincompatible-library-redeclaration (warn_redecl_library_builtin).
Mechanism: once read/write are in the builtin table, clang materializes an implicit prototype for them (ssize_t read(int, void *, size_t)). When someone has their own external-linkage, file-scope read/write in C with a different signature (say int read(char *)), it gets merged against that implicit prototype, conflicts, and warns. The function isn't actually treated as a builtin or miscompiled - clang only attaches builtin-ness when the signature matches; the mismatched ones just get the warning. And static ones are excluded entirely (I checked, static int vfork(int); gives no library-redeclaration warning, just -Wundefined-internal).

So why did it hurt for read/write specifically? Those names are everywhere, lots of real C code declares its own read/write with different signatures, so a pile of previously-clean code started warning. That's what got them pulled and dispatched by name instead.

Now,for umask: I checked whether IgnoreSignature dodges this, and it doesn't, vfork is IgnoreSignature and a mismatched int vfork(int); still warns. So making umask a builtin does reintroduce the same class of warning in principle. The difference is impact, umask is a rare, warning is suppressed in system headers, so the libc's own <sys/stat.h> won't warn even on targets where mode_t is 16-bit and wouldn't match whatever width we encode. Only a hand-written, non-system umask with a mismatched signature would trip it.

So it lands right where you suggested, and the "extra mile" if it ever bites would be giving the prototype a target-resolved mode_t (the way pid_t goes through getProcessIDType), which doesn't exist in TargetInfo today, so it'd be a small net-new addition.

I'll get the prototype wired up and report back.

AaronBallman: > So why did it hurt for read/write specifically? Those names are everywhere, lots of real C code declares its own read/write with different signatures, so a pile of previously-clean code started warning. That's what got them pulled and dispatched by name instead.

Thanks, that's good to know!

nuclearcat: I hit a snag that changes the calculus, need your opinion before I go further.
Problem is, the builtin only fires where modet is 32-bit unsigned int (Linux/glibc). On macOS and the BSDs modet is _uint16t, and there the check goes silently dead - not just for hand-written declarations, but for the libc's own <sys/stat.h>. I confirmed this on arm64-apple-macosx: umask(0xFFFF) produces no diagnostic at all.
I encoded the prototype as unsigned int(unsigned int). GetBuiltinType resolves that fine, so LazilyCreateBuiltin materializes an implicit unsigned int umask(unsigned int) during the libc decl's lookup. That makes the libc's unsigned short umask(unsigned short) a conflicting redeclaration - so MergeFunctionDecl emits warnredecllibrarybuiltin (suppressed in the system header) and the merged decl drops BuiltinAttr. getBuiltinID() then returns 0 and the check bails. The IgnoreSignature/unconditional-attach path in ActOnFunctionDeclarator is never reached, because that only applies when the decl isn't already a redeclaration of the implicit builtin.
I noticed vfork somehow similar, don't have this problem. vfork's pidt() isn't a fixed type - pidt is encoded as p to getProcessIDType() to the target's real pidt. So vfork's implicit decl matches the libc on every target and keeps its builtin id. IgnoreSignature works for vfork because the encoded type is already target-correct; modet has no such encoding, so my hardcoded unsigned int defeats it. So my earlier "same mechanism as vfork" claim was wrong - that mismatch is exactly the failure.
so, this is actually worse coverage than the system-header-origin gate I had before, which fired on 16-bit modet too (it only required an integer param/return).

A. Revert to the system-header-origin gate. Full target coverage; the only cost is losing the diagnostic when someone forward-declares umask without including <sys/stat.h> (a much smaller hole than macOS/BSD.
B. Keep the builtin, but make modet target-resolved, i can add TargetInfo::getModeType() + a modet type code so the implicit prototype matches the libc everywhere (the vfork/pid_t treatment, the "extra mile" I mentioned earlier).

I lean toward second option, for the long-term-correct behavior, but it's a net-new TargetInfo API. Would you want that done in this PR, or split into a separate prep PR that adds getModeType() first? Happy to go either way

AaronBallman: Could we use custom type checking for the function signature and validate that it's an integer data type as opposed to a particular flavor of integer type?

nuclearcat: Fixed with an empty prototype + IgnoreSignature (like pthread_create): no implicit builtin decl is formed, so the libc's umask keeps its builtin id whatever width mode_t is, and the existing "single integer arg" guard handles the type. Added a 16-bit Darwin regression test. Thanks for the steer!

clang/include/clang/Basic/Builtins.td
```diff
@@ -3800,6 +3800,14 @@ def VFork : LibBuiltin<"unistd.h"> {
   let Prototype = "pid_t()";
 }
 
+// POSIX sys/stat.h
+
+def Umask : LibBuiltin<"sys/stat.h"> {
+  let Spellings = ["umask"];
+  let Attributes = [IgnoreSignature];
+  let Prototype = "";
```
AaronBallman:
```suggestion
  let Prototype = "mode_t(mode_t)";
```
Just so folks have an idea what the signature is even though we ignore it.

clang/include/clang/Basic/Builtins.td
```diff
@@ -3800,6 +3800,14 @@ def VFork : LibBuiltin<"unistd.h"> {
   let Prototype = "pid_t()";
 }
 
+// POSIX sys/stat.h
+
+def Umask : LibBuiltin<"sys/stat.h"> {
+  let Spellings = ["umask"];
+  let Attributes = [IgnoreSignature];
```
AaronBallman:
```suggestion
  let Attributes = [IgnoreSignature, NoThrow];
```

AaronBallman (commented):
I think this is heading in the right direction; a few drive-by comments while waiting to see how precommit CI likes the changes.

clang/include/clang/Basic/Builtins.td
nuclearcat: Amended, thank you!

clang/include/clang/Basic/Builtins.td
nuclearcat: Amended, thank you!

AaronBallman:
> Builtins.td:3868:7: error: Unknown Type: mode_t
> ```let Prototype = "mode_t(mode_t)";```

Today I learned that even if the signature is ignored, it still needs to be valid. Let's use comments for the prototype instead:
```
let Prototype = ""; // mode_t(mode_t); mode_t is target-specific
```

nuclearcat:
> > Builtins.td:3868:7: error: Unknown Type: mode_t
> > `let Prototype = "mode_t(mode_t)";`
> 
> Today I learned that even if the signature is ignored, it still needs to be valid. Let's use comments for the prototype instead:
> 
> ```
> let Prototype = ""; // mode_t(mode_t); mode_t is target-specific
> ```

Done

AaronBallman (approved):
LGTM presuming precommit CI is happy, thank you!

clang/lib/Sema/SemaChecking.cpp
```diff
@@ -1571,6 +1571,42 @@ void Sema::checkFortifiedBuiltinMemoryFunction(FunctionDecl *FD,
...
+  // regardless of the libc's mode_t spelling -- including a hand-written
+  // forward declaration without <sys/stat.h>. A static/local lookalike or a
+  // C++ (non-extern-"C") declaration keeps a zero builtin id and is ignored.
+  if (FD->getBuiltinID() != Builtin::BIumask)
+    return;
+
+  // umask(mode_t): warn when the constant-evaluated argument has bits set
+  // outside the file-permission mask (0777). Those bits are ignored.
+  if (TheCall->getNumArgs() != 1)
+    return;
```
nickdesaulniers, line 1590: Hmm...this is a lot of `umask` specific checks in a method with a more generic name like `Sema::checkFortifiedLibcArgument`. I think `recv` has similar issues to `umask` regarding Fortify, so we might end up breaking up/refactoring this code imminently in https://github.com/llvm/llvm-project/pull/223520.

nuclearcat: Fair point, the generic name is left over from the earlier implementation. Since this has already landed, I can send a follow-up to rename it to checkFortifiedUmask and fix the stale declaration comment about name-based dispatch.
Alternatively, I’m happy to help with broader refactoring as part of or coordinated with #223520 if that would be more useful.

PR #198131, opened 2026-05-16
[InstCombine] Handle "trunc nuw to i1" as "icmp ne,0" in foldSelectValueEquivalence

andjo403:
regression seen in https://github.com/dtcxzyw/llvm-opt-benchmark/pull/3528/changes/BASE..4dc27803d5fd3fdf03c6b423dff6235f557da692#r2874274598 from https://github.com/llvm/llvm-project/pull/184182
seen in 49 rust files in llvm-opt-benchmark with https://github.com/llvm/llvm-project/pull/184182

proof: https://alive2.llvm.org/ce/z/MGhWgE


nikic (commented):
Can this be done in a way that does not require creating a temporary instruction?

andjo403:
hmm as this only handles "trunc nuw to i1" directly in the falseValue there is probably no need to call simplifyWithOpReplaced

andjo403:
pushed a fixup without create of temp icmp

andjo403:
rebased due to merge conflikt in the test file

andjo403:
ping

dtcxzyw (approved):
LGTM. Generalized proof: https://alive2.llvm.org/ce/z/MGhWgE

PR #198136, opened 2026-05-17
[FileCheck] Resurrect overflow tests

jdenny-ornl:
D150880 (landed as 0726cb004718) uses `APInt` to eliminate most integer overflow issues from FileCheck numeric variables.  It also removes the 4 tests in `llvm/test/FileCheck/match-time-error-propagation`.

While the elimination of overflow issues reduces the importance of those tests, the tests still seem worthwhile.  Without them, I see no test that exercises the "unable to substitute variable or numeric expression: overflow error" diagnostic in FileCheck input dumps.

This patch resurrects those tests and updates them to exercise the remaining unsigned underflow case.


llvm/test/FileCheck/match-time-error-propagation/matched-expected-pattern.txt
```diff
@@ -0,0 +1,58 @@
+; Check handling of match-time diagnostics for invalid patterns (e.g.,
+; substitution overflow) in the case of expected patterns (e.g., CHECK-NOT) that
```
MaskRay: `CHECK-NOT` => `CHECK`

MaskRay:
If the intention is to show the behavior change of https://github.com/llvm/llvm-project/pull/198138 , you can move the new test files there to this patch

llvm/test/FileCheck/match-time-error-propagation/matched-expected-pattern.txt
jdenny-ornl: Thanks for catching that.  Fixed.

jdenny-ornl:
> If the intention is to show the behavior change of #198138 , you can move the new test files there to this patch

Good idea.  Done.

jdenny-ornl:
> > If the intention is to show the behavior change of #198138 , you can move the new test files there to this patch
> 
> Good idea. Done.

After further thought, I decided to remove those tests here and add them in a new PR #199063.  That way, this PR and PR #198137, which I believe are worthwhile independently of PR #198138, can be reviewed and merged before or without it.

jdenny-ornl:
> LLVM.CodeGen/PowerPC/fast-isel-cmp-imm.ll (Likely Already Failing)

That's definitely not due to this PR, which merely resurrects FileCheck tests.

MaskRay (approved):

PR #198137, opened 2026-05-17
[FileCheck] Refactor -dump-input test

jdenny-ornl:
This PR is stacked on PR #198136.

This patch refactors `llvm/test/FileCheck/dump-input/annotations.txt` to improve maintainability and coverage and to prepare for the upcoming implementation of search range annotations.

Lit substitutions
=================

The test repeats the same basic set of RUN lines *many* times.  This patch encapsulates those in lit substitutions to improve maintainability.  By doing so, it also helps to ensure more consistent coverage of all cases and thus slightly expands coverage.

-strict-whitespace
==================

Via those substitutions, this patch adds `-strict-whitespace` throughout the test, and it drops the initial `-strict-whitespace` case because it is then redundant.  That causes many whitespace changes throughout the test, so this patch is easier to read with `git diff -w`.

When I originally wrote the test, I thought maintaining it would be too painful with `-strict-whitespace`.  However, I now think it is important for usability to thoroughly check that annotations are correctly aligned with the input, especially given the upcoming search range annotations.

-dump-input-label-width
=======================

To address that anticipated maintenance pain, and to make the above change easier to implement, this patch also implements a new hidden FileCheck option, `-dump-input-label-width`.  It enables tests like this one not to have to fuss with fluctuations in the label column width that are caused when varying the verbosity options.  I do not anticipate this option will be used outside FileCheck's own test suite.

Splitting directive blocks
==========================

To improve readability, this patch splits apart directive blocks where the same annotations appear multiple times with small differences at different verbosity levels.  See new header comments for details.


llvm/test/FileCheck/dump-input/annotations.txt
```diff
@@ -713,26 +746,33 @@
...
+; REDEFINE:     -DDEF_MATCH1=def-match1 -DDEF_MATCH2=def-match2 \
+; REDEFINE:     -DDEF_NOMATCH=foobar
+; RUN: not %{run-vv}
+; RUN: not %{run-v}
+; RUN: not %{run}
+; REDEFINE: %{opts} =
+
+; Verbose diagnostics are suppressed but not errors.
+; SUBST-NEG:{{.*}}error:{{.*}}
+; SUBST-NEG:{{.*}}error:{{.*}}
```
MaskRay: Duplicate check?

MaskRay (approved):

llvm/test/FileCheck/dump-input/annotations.txt
jdenny-ornl: No, multiple errors, as expanded in the input dump below.  Would using `-COUNT-2` make this clearer?

jdenny-ornl: Applied.

jdenny-ornl:
Thanks for the reviews.

Issue #77257, opened 2024-01-07
FileCheck should a unified diff output


cjdb:
I find the way in which FileCheck renders its output very confusing and difficult to read. It would be good if there was a `--diff` option that let users render their mismatches in the form of various diff outputs too. In addition to making the output more readable, this would also make it easier to show multiple mismatches in the same run.

**FileCheck default**
```
--
/tmp/example/test:52:17: error: CHECK_CMAKE: expected string not found in input
// CHECK_CMAKE: -- The C compiler identification is [[COMPILER]] {{\d+(.\d+)*}}
                ^
<stdin>:6:32: note: scanning from here
-- Running vcpkg install - done
                               ^
<stdin>:6:32: note: with "COMPILER" equal to "GNU"
-- Running vcpkg install - done
                               ^
<stdin>:7:1: note: possible intended match here
-- The C compiler identification is GNU 13.2.1
^

Input file: <stdin>
Check file: /tmp/example/test

-dump-input=help explains the following input dump.

Input was:
<<<<<<
            1: -- Bootstrapping vcpkg before install 
            2: -- Bootstrapping vcpkg before install - done 
            3: -- Running vcpkg install 
            4: All requested packages are currently installed. 
            5: Total install time: 581 ns 
            6: -- Running vcpkg install - done 
check:52'0                                    X error: no match found
check:52'1                                      with "COMPILER" equal to "GNU"
            7: -- The C compiler identification is GNU 13.2.1 
check:52'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
check:52'2     ?                                               possible intended match
            8: -- The CXX compiler identification is GNU 13.2.1 
check:52'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            9: -- Detecting C compiler ABI info 
check:52'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           10: -- Detecting C compiler ABI info - done 
check:52'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           11: -- Check for working C compiler: /usr/bin/gcc - skipped 
check:52'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           12: -- Detecting C compile features 
[...]

PR #198138, opened 2026-05-17
[FileCheck] Annotate search ranges with { } in -dump-input

jdenny-ornl:
This PR has an [RFC](https://discourse.llvm.org/t/rfc-filecheck-improving-input-dump-readability/91112).  It is stacked on PR #199063.

Example
=======

```
$ cat check
CHECK: start
CHECK-NEXT: end

$ FileCheck -v -dump-input-context=2 check < input |& tail -23
<<<<<<
          1: start
check:1      ^~~~~
next:2'0         {   search range start (exclusive)
          2: foo0
          3: foo1
          .
          .
          .
         21: foo19
         22: foo20
         23: end
next:2'1     !~~   error: match on wrong line
         24: bar0
         25: bar1
          .
          .
          .
         42: bar18
         43: bar19
         44: bar20
next:2'2           } search range end (exclusive)
>>>>>>
```

Without this patch, input lines 1-3 and 42-44 are not shown.  However, lines 1-3 are where the actual problem is because that is where the `CHECK-NEXT` directive was expected to match but did not.

Search Ranges Are Helpful
=========================

In general, this patch marks any failed pattern's search range by using the annotation style shown above, and these annotations are filtered in when using `-dump-input-filter=error`, which is the default filter.  Seeing the search range can be helpful for understanding the pattern's behavior.  Moreover, the cause of the pattern failure is often the input at the start or end of the search range.  For example:

- A `CHECK-NEXT` or `CHECK-SAME` match on the wrong line, as in the above example.
- A `CHECK-NOT` unexpected match because a neighboring directive matched at an unexpected point, affecting the search range.
- An unmatched `CHECK` because a subsequent `CHECK-LABEL` matched at an unexpected point, affecting the search range end.  (In this case, the search range start and thus the prior directive's match is already revealed without this patch.)

This patch updates tests in `llvm/test/FileCheck/dump-input/search-range-annotations`, which demonstrate its benefit for those cases.

This patch is a replacement for D96653, which attempted to address the above cases but in a less straight-forward and somewhat broken manner. The idea of the current patch was discussed during that review.

Replacing `X~~`
===============

Without this patch, search ranges for unmatched patterns (whether a success, as for `CHECK-NOT`, or a failure, as for `CHECK`) are already marked with `X~~`.  In those cases, this patch replaces those annotations with the new annotations shown above.  As described above, this patch adds the new annotations to all other failed matched patterns as well.

`X~~` is thus no longer used by `-dump-input`.  The `X~~` style is very noisy, especially for consecutive unmatched unexpected patterns (like a `CHECK-NOT` block), all of which mark every line of their identical search ranges.  Switching to the new style significantly reduces the noise in such cases.  That noise was discussed recently in issue #77257 and PRs linked from there, and that discussion led to the resurrection of this patch.

Without this patch, `-dump-input-filter=error` filters in only the start of a search range that spans multiple lines because an entire `X~~` is one annotation.  With this patch, it filters in both the start and end of a search range because they are separate annotations (or a single one-liner, as discussed below).  With or without this patch, a different argument to `-dump-input-filter` (and `-vv`) is required to filter in the search range for an unmatched unexpected pattern, like `CHECK-NOT`, because that is not an error.

One-Liners
==========

If a search range does not involve multiple input lines, this patch keeps the `{` and `}` markers on the same output line like this:

```
<<<<<<
         1: start foo end
check:1     ^~~~~
check:3               ^~~
not:2           {     }     search range (exclusive bounds)
>>>>>>
```

Exclusive Boundaries
====================

`{` and `}` are to be interpreted as exclusive bounds.  That is, the characters at those markers are not included in the search range, but everything in between is.  To try to avoid confusion, this patch adds the word "exclusive" in every search range annotation.

When the search range starts or ends at a line boundary, the marker cannot be placed at the first or last character of the line because that would exclude that character.  This patch instead places the marker in the margin of the input dump, either before the line's first character or after the line's last character (usually a newline), rather than on the adjacent line.  I have found this makes the annotations easier to read (more apt have a one-liner, at least), and I do not think it would make much sense to move a start annotation to an imaginary line 0.  For example, the `{` and `}` below appear in the input dump margins, before the first character, `s`, and after the space representing the newline:

```
<<<<<<
           1: start
check:1'0    {      } search range (exclusive bounds)
check:1'1             error: no match found in search range
>>>>>>
```

Before trying exclusive bounds, I experimented with notations involving inclusive bounds (e.g., `[` and `]`, or `[` and `)`).  Such notations either cannot distinguish empty ranges (without a cryptic inversion like `][`) from one-character ranges, or they cannot represent them with one-liners (because they must occupy the same column), increasing verbosity.  I ultimately decided I prefer exclusive bounds because they are visually and semantically symmetric while consistently concise and unambiguous.

For comparison, match ranges (e.g., `^~~`) cannot distinguish one-character ranges from empty ranges.  However, in my experience, empty match ranges are uncommon, and usually it is easy to distinguish them based on the pattern or directive (e.g., `CHECK-EMPTY`).  In contrast, empty search ranges can occur repeatedly when directive matches are adjacent and `-implicit-check-not` is used.


jdenny-ornl:
In case you're already reviewing the implementation, be aware of [this minor cleanup I just applied](https://github.com/llvm/llvm-project/pull/198138/changes/21d1e2117b40b7d78d399e5999b2a5d10e80f8eb), which was not previously discussed.

llvm/utils/FileCheck/FileCheck.cpp
```diff
@@ -381,7 +381,8 @@ struct InputAnnotation {
   unsigned InputLine;
   /// The column range (one-origin indexing, inclusive boundaries) in which to
   /// mark the input line.  If \c InputLastCol is \c UINT_MAX, the rest of the
-  /// input line should be marked.
+  /// input line should be marked, and another \c InputAnnotation will continue
+  /// it on the next line.
   unsigned InputFirstCol, InputLastCol;
```
MaskRay, line 387: Is InputFirstCol now 0-origin?

llvm/utils/FileCheck/FileCheck.cpp
MaskRay: This ctor is now dead

MaskRay (approved):
LGTM. The patch's new code is fully exercised in a coverage build.

llvm/utils/FileCheck/FileCheck.cpp
jdenny-ornl: Good point.  The answer is not a simple yes or no, so I have expanded the comments to explain.

llvm/utils/FileCheck/FileCheck.cpp
jdenny-ornl: Removed.

jdenny-ornl:
> LGTM.

Thanks for the review.  As we discussed previously, I believe this patch's effect will be highly visible throughout the LLVM community. Unless there are bugs we missed, I do not know how anyone might argue it worsens their use cases.  But I have been surprised before. Do you think we should ask for additional reviews of at least the behavioral change (e.g., by looking through test changes)?  Or should we post a PSA before landing?

> The patch's new code is fully exercised in a coverage build.

I do not have much experience with LLVM's coverage tools.  Is there a public report?  Or can you give me a brief description of what you did?

MaskRay:
> > LGTM.
> 
> Thanks for the review. As we discussed previously, I believe this patch's effect will be highly visible throughout the LLVM community. Unless there are bugs we missed, I do not know how anyone might argue it worsens their use cases. But I have been surprised before. Do you think we should ask for additional reviews of at least the behavioral change (e.g., by looking through test changes)? Or should we post a PSA before landing?

Yes, it's a good idea to invite suggestions on the syntax...

> > The patch's new code is fully exercised in a coverage build.
> 
> I do not have much experience with LLVM's coverage tools. Is there a public report? Or can you give me a brief description of what you did?

It's not a public report. I did a `LLVM_BUILD_INSTRUMENTED_COVERAGE=ON `LLVM_PROFILE_FILE_PATTERN=/tmp/out/profraw/%4m.profraw -DCMAKE_CXX_FLAGS=-fprofile-continuous` build to check whether all new conditions are covered by tests.

PR #196649, opened 2026-05-08
[InstrProf] Deduplicate VP values

boomanaiden154:
Zero VP values can come up in some places. They are intentional around
external symbols for indirect call sites, and it seems like they might
be unintentional around memop VP metadata
(https://reviews.llvm.org/D92074). This patch combines them so that we
can enforce the variant that there are no duplicate values in VP
metadata, which allows passes to make some simplifying assumptions. We
also deduplicate non-zero values, because there is error handling for
them and still some undebugged cases where they show up
(https://reviews.llvm.org/D136211).

This ended up being a bit messier than I would like due to the need to
handle non-zero duplicate values and preserve existing error handling
behavior in llvm-profdata. I've left comments explaining this so we can
hopefully clean this up when llvm-profdata eventually gets fixed. The
error has shown up in some places
(https://issues.chromium.org/issues/353702041), so does still exist, but
I still have not been able to find profraw files to be able to fix the
underlying issue. Given the error is still occurring in the wild, it
also does not make sense to remove it, even if we do not generate bad VP
metadata now.

PR #193083, opened 2026-04-20
[IR] Verify that VP !prof does not have duplicate prof values

boomanaiden154:
Follows on #193077. We should not have duplicate values as they should
be merged before being moved into IR. Add this to the verifier so that
we can actually enforce this constraint.

PR #198140, opened 2026-05-17
[ICP] Update comment about duplicate values in VP MD

boomanaiden154:
With https://github.com/llvm/llvm-project/pull/196649 deduplicating VP
values and https://github.com/llvm/llvm-project/pull/193083 enforcing
this, we no longer need to worry about duplicate values, zero or
otherwise. Update the comment to reflect this.


mingmingl-llvm (approved):
thanks for adding the verification in https://github.com/llvm/llvm-project/pull/193083!

PR #198143, opened 2026-05-17
[NVPTX] Clear unsafe kill flags in NVPTXProxyRegErasure

### LINKED ISSUE #198142: [NVPTX] Incorrect kill flags after ProxyReg Erasure
### Description
In the NVPTX backend, the `NVPTXProxyRegErasure` pass replaces uses of proxy registers with their source registers and erases the proxy register instructions. However, it fails to update the kill flags correctly during this replacement.

If a proxy register use is marked as killed, the pass blindly transfers the kill flag to the source register during replacement. If the source register has subsequent downstream uses, this incorrectly terminates its live range prematurely, leading to incorrect liveness information and potential miscompilation.

### Steps to Reproduce
Consider the following Machine IR snippet before the `nvptx-proxyreg-erasure` pass:

``` asm
    %2:b16 = CVT_u16_u32 %0:b32, 0
    %3:b16 = ProxyRegB16 %2:b16
    %4:b16 = ADD16rr %1:b16, killed %3:b16
    %5:b16 = ADD16rr %1:b16, %2:b16
```

After the pass runs, the current code replaces `%3` with `%2` but leaves the `killed` flag intact:

``` asm
%2:b16 = CVT_u16_u32 %0:b32, 0
%4:b16 = ADD16rr %1:b16, killed %2:b16 ; BUG:   // %2 is incorrectly killed
%5:b16 = ADD16rr %1:b16, %2:b16 ;  // %2 is used after being “killed”
```

andreww0210:
Fixes #198142

After replacing regsiter, stale `killed` flags my leave on non-terminal uses, breaking liveness and triggering MachineVerifier errors.
    
Fix this by conservatively dropping the `killed` flag on `ToReg`.


andreww0210:
Hi @Artem-B @Prince781 @AlexMaclean,

This is my very first contribution to the NVPTX backend and the LLVM project! As a new contributor, I currently don't have the permissions to request reviewers manually or trigger the CI workflows automatically.

Could you please take a look at this patch when you have a moment, or kindly add someone who might be interested? Any feedback would be greatly appreciated!

Thank you!

Artem-B:
Reproducer: https://godbolt.org/z/WoE6Ev73E  
  
I'm not sure if we can just unset killed and rely on Liveness will be recomputed by subsequent passes.

  
We're running before regalloc, so we're probably OK, but that's an implementation detail. We may want to add an explicit assert on `MRI.tracksLiveness() or on !MF.getProperties().hasProperty(MachineFunctionProperties::Property::NoVRegs)` to make sure we do not drop those flags when their consistency may be expected. I guess `tracksLiveness` may be a better choice as it checks for it directly rather than just checking whether regalloc already happened.

andreww0210:
> I'm not sure if we can just unset killed and rely on Liveness will be recomputed by subsequent passes.
> 
> We're running before regalloc, so we're probably OK, but that's an implementation detail. We may want to add an explicit assert on `MRI.tracksLiveness() or on !MF.getProperties().hasProperty(MachineFunctionProperties::Property::NoVRegs)` to make sure we do not drop those flags when their consistency may be expected. I guess `tracksLiveness` may be a better choice as it checks for it directly rather than just checking whether regalloc already happened.

Hi @Artem-B, thanks for the review and the reproducer!

I've looked into this more carefully. Here's my understanding:

TracksLiveness is set by default when a MachineFunction is created. In the standard pipeline, it remains true through the pre-regalloc passes that run before ProxyRegErasure. Some passes may invalidate it via invalidateLiveness(), but those typically run later. So in practice, tracksLiveness() should be true when our pass runs, but it's not a hard framework guarantee — which is exactly why an assert is valuable as a safety net.

I agree with adding the assert. One thing I'd like to clarify though: since ProxyRegErasure runs pre-regalloc where tracksLiveness is true, dropping kill flags is safe in the conservative direction — we're only losing precision, not introducing incorrect kill markers. The assert would ensure we don't silently drop kill flags in a context where liveness is no longer tracked and downstream passes might rely on their accuracy.

An alternative would be to conservatively recompute kill flags at the end of the pass using MRI.hasOneUse() (similar to the ISel heuristic in InstrEmitter::AddRegisterOperand). This would correctly set kill for single-use registers, though it wouldn't handle the multi-use case (which requires proper liveness analysis to determine the last use in program order).

I'm inclined to go with the unconditional drop for simplicity. But I'm happy to add the hasOneUse-based recompute if you think it's worth it.

Artem-B:
We may need to get someone more familiar with the liveness tracker nuances than myself. I agree that the change is needed and makes things more conservative, and that future passes will be able to recalculate correct 'killed' flag. My concern is that we may be introducing an inconsistency, but I do not know enough to tell whether it's a problem at all, and if it is whether it's benign or not.

@nikic would you happen to know what's the right way to update/unset the `killed` flag or who may be the right person to ask?

andreww0210:
Hi @Artem-B  , gentle ping!  Since we haven’t heard back from @nikic  on the liveness question, is there anyone else you’d recommend taking a look? 😭

nikic:
@arsenm probably

llvm/lib/Target/NVPTX/NVPTXProxyRegErasure.cpp
arsenm: Why isn't this using MRI.replaceRegWith? Can follow up with clearKillFlags (maybe we should just make replaceRegWith do that as it goes). Long term we should stop having kill flags

andreww0210: Thanks for the suggestion!!! I refactored to use MRI.replaceRegWith() followed by clearKillFlags().

Artem-B (approved):

llvm/test/CodeGen/NVPTX/proxy-reg-erasure-kill-flag.mir
```diff
@@ -0,0 +1,25 @@
+# RUN: llc %s --run-pass=nvptx-proxyreg-erasure -verify-machineinstrs -mtriple=nvptx64 -o - | FileCheck %s
+
+--- |
+  define void @test_proxy_reg() {
+  entry:
+    ret void
+  }
+
+...
```
arsenm:
```suggestion
# RUN: llc %s --run-pass=nvptx-proxyreg-erasure -mtriple=nvptx64 -o - | FileCheck %s

```

andreww0210:
Ping. This PR still needs an approving review with write access. As I don’t have merge access, could you please help approve and land it? Thanks!

andreww0210:
@arsenm @Prince781 @AlexMaclean

AlexMaclean (approved):
LGTM

Prince781 (approved):
LGTM

DavidSpickett:
If CI results are ok, one of the approvers please merge this on behalf of the author.

andreww0210:
> If CI results are ok, one of the approvers please merge this on behalf of the author.

Thanks! But the build failed due to a Flang build error. It seems unrelated to my PR…

andreww0210:
@Artem-B @Prince781 @AlexMaclean Could you help merge it?

PR #198144, opened 2026-05-17
[Passes] Use consistent naming for printer passes

boomanaiden154:
Fix the long-standing TODO in PassRegistry.def. The consistency will be nice and these passes are not used enough outside of tests for the churn to be a major issue.

Also drop the LegacyPM initialization in LinkAllPasses given the printer pass does not support the LegacyPM.

Assisted by Gemini in Windsurf.


llvm/docs/AliasAnalysis.rst
```diff
@@ -638,16 +638,16 @@ implementations.  You can use them with commands like:
 
   % opt -ds-aa -aa-eval foo.bc -disable-output -stats
 
-The ``-print-alias-sets`` pass
+The ``print<alias-sets>`` pass
```
boomanaiden154, line 641: Basically everything else on this page also needs to be updated, but I think we can leave that out of scope for this patch.

nikic (approved):
LGTM

PR #190152, opened 2026-04-02
[LSV][AMDGPU] Vectorize unordered and monotonic atomic loads

harrisonGPU:
Unordered and monotonic atomic loads do not impose cross address synchronization and
can be safely vectorized. 

This patch allows LoadStoreVectorizer to merge these atomic loads, taking care to group
candidates by atomic ordering and sync scope. The resulting vectorized load correctly preserves
these scalar atomic attributes.

Atomic stores and loads with stronger orderings (acquire, release, seq_cst) remain explicitly disabled.

Reference:
https://llvm.org/docs/Atomics.html#monotonic

PR #198145, opened 2026-05-17
[AMDGPU][AtomicExpand] Support sub naturally aligned 64 bit atomic load/store

harrisonGPU:
Add `supportsUnalignedAtomicLoadInIR` and `supportsUnalignedAtomicStoreInIR`
virtual hooks to `TargetLowering`. These allow targets to accept atomic
load/store instructions that fail the natural alignment check in
atomicSizeSupported but can still be lowered to a single atomic hardware
instruction.

Use these hooks in AtomicExpandPass to skip libcall expansion, and in
SelectionDAGBuilder to suppress the fatal error, when the target
explicitly supports the access.

Implement the hooks for AMDGPU: global and flat address spaces support
64 bit atomic load/store with 4 byte alignment (e.g. <2 x float> align 4),
because global_load_b64/flat_load_b64 only require DWORD alignment.
LDS is rejected because ds_load_b64 requires 8 byte alignment. Sub DWORD
aligned accesses are always rejected since unaligned access mode only
enables non atomic split operations which would break atomicity.


harrisonGPU:
When implementing two consecutive load atomic float, align 4 operations combined into a single load atomic <2 x float>, align 4 in PR https://github.com/llvm/llvm-project/pull/190152, AMDGPU was found to not support sub naturally aligned atomic loads, prompting this patch to add support for them.

llvm/include/llvm/CodeGen/TargetLowering.h
```diff
@@ -2452,6 +2452,12 @@ class LLVM_ABI TargetLoweringBase {
     return AtomicExpansionKind::None;
   }
 
+  /// Returns true if the target supports the given sub naturally aligned
+  /// atomic load without falling back to a libcall.
+  virtual bool supportsUnalignedAtomicLoadInIR(const LoadInst *LI) const {
+    return SupportsUnalignedAtomics;
+  }
```
arsenm, line 2459: Should not add a new hook, especially one that looks like this.

llvm/lib/Target/AMDGPU/SIISelLowering.cpp
```diff
@@ -20340,6 +20341,50 @@ SITargetLowering::shouldExpandAtomicStoreInIR(StoreInst *SI) const {
...
+bool SITargetLowering::supportsUnalignedAtomicLoadInIR(
+    const LoadInst *LI) const {
+  unsigned AS = LI->getPointerAddressSpace();
+  // LDS uses ds_read2_b32 for 4 byte aligned 8 byte accesses (two separate
+  // loads), but that technique breaks atomicity. ds_load_b64 requires 8 byte
+  // alignment. Reject all sub naturally aligned atomic LDS/GDS loads.
+  if (AS == AMDGPUAS::LOCAL_ADDRESS || AS == AMDGPUAS::REGION_ADDRESS)
+    return false;
+
+  // Atomic operations require at least DWORD alignment. Unaligned access mode
```
arsenm, line 20359: Atomic operations require natural alignment

arsenm:
> because global_load_b64/flat_load_b64 only require DWORD alignment.

I believe these require natural alignment to give the atomic behavior

harrisonGPU:
> > because global_load_b64/flat_load_b64 only require DWORD alignment.
> 
> I believe these require natural alignment to give the atomic behavior

Thanks Matt. At first, I also noticed that the shader programming guide says:

```text
Atomics must be aligned to the data size, or they trigger a MEMVIOL.
```

So initially I also thought atomics required natural alignment. However, after testing several cases on gfx12, I found that atomic load/store behaves differently from other atomic operations.

For example:

```llvm
%a0 = load <2 x float>, ptr addrspace(1) %p, align 8
```

is lowered to:

```asm
global_load_b64 v[0:1], v[0:1], off
```

At the same time:

```llvm
%a0 = load atomic <2 x float>, ptr addrspace(1) %p syncscope("agent") monotonic, align 8
```

is lowered to:

```asm
global_load_b64 v[0:1], v[0:1], off scope:SCOPE_DEV
```

This suggests that both normal loads and atomic loads are lowered to `global_load`, and the generated instruction itself is not an actual atomic instruction.

In contrast, a real atomic operation such as:

```llvm
atomicrmw add ptr addrspace(1) %arrayidx, i32 1 monotonic
```

is lowered to:

```asm
global_atomic_add_u32 v0, v1, s[0:1] scale_offset scope:SCOPE_SYS
```

So I think atomic load/store is fundamentally different from true atomic read-modify-write operations.

I also noticed that the LLVM Language Reference says:

> Note: if the alignment is not greater or equal to the size of the <value> type, the atomic operation is likely to require a lock and have poor performance.

https://llvm.org/docs/LangRef.html#id214

This seems to imply that LLVM IR allows misaligned atomic load/store, although it may have performance implications.

Based on this, I think we probably need to support this case. What do you think?

PR #198146, opened 2026-05-17
[RISCV][Disassembler] Refactor simple predicate decoders using a template

arichardson:
This replaces the manual boilerplate for DecodeGPRNoX0, DecodeGPRNoX2,
DecodeGPRNoX31, and DecodeGPRPairNoX0 with a universal filtering template
and constexpr predicate functions.

I will need more of these for the RVY patch series, so submitting this NFC
cleanup first.


topperc (approved):
LGTM

PR #151639, opened 2025-08-01
[ELF] -r: Synthesize R_RISCV_ALIGN at input section start

MaskRay:
Without linker relaxation enabled for a particular relocatable file or
section (e.g., using .option norelax), the assembler will not generate
R_RISCV_ALIGN relocations for alignment directives. This becomes
problematic in a two-stage linking process:

```
ld -r a.o b.o -o ab.o
// b.o is norelax. Its alignment information is lost in ab.o.
ld ab.o -o ab
```

When ab.o is linked into an executable, the preceding relaxed section
(a.o's content) might shrink. Since there's no R_RISCV_ALIGN relocation
in b.o for the linker to act upon, the `.word 0x3a393837` data in b.o
may end up unaligned in the final executable.

To address the issue, this patch inserts NOP bytes and synthesizes an
R_RISCV_ALIGN relocation at the beginning of a text section when the
alignment >= 4.

For simplicity, when RVC is disabled, we synthesize an ALIGN relocation
(addend: 2) for a 4-byte aligned section, allowing the linker to trim
the excess 2 bytes.

See also https://sourceware.org/bugzilla/show_bug.cgi?id=33236

PR #150816, opened 2025-07-27
MC: Refine ALIGN relocation conditions

MaskRay:
Each section now tracks the index of the first linker-relaxable
fragment, enabling two changes:

* Delete redundant ALIGN relocations before the first linker-relaxable
  instruction in a section. The primary example is the offset 0
  R_RISCV_ALIGN relocation for a text section aligned by 4.
* For alignments larger than the NOP size after the first
  linker-relaxable instruction, ALIGN relocations are now generated, even in
  norelax regions. This fixes the issue #150159.

The new test llvm/test/MC/RISCV/Relocations/align-after-relax.s
verifies the required ALIGN in a norelax region following
linker-relaxable instructions.
By using a fragment index within the subsection (which is less than or
equal to the section's index), the implementation may generate redundant
ALIGN relocations in lower-numbered subsections before the first
linker-relaxable instruction.

align-option-relax.s demonstrates the ALIGN optimization.
Add an initial `call` to a few tests to prevent the ALIGN optimization.

---

When the alignment exceeds 2, we insert $alignment-2 bytes of NOPs, even
in non-RVC code. This enables non-RVC code following RVC code to handle
a 2-byte adjustment without requiring an additional state in MCSection
or AsmParser.

```
.globl _start
_start:
// GNU ld can relax this to  6505          lui     a0, 0x1
// LLD hasn't implemented this transformation.
  lui a0, %hi(foo)

.option push
.option norelax
.option norvc
// Now we generate R_RISCV_ALIGN with addend 2, even if this is a norvc region.
.balign 4
b0:
  .word 0x3a393837
.option pop
foo:
```

PR #198147, opened 2026-05-17
[ELF] -r: Don't let a weaker offset-0 R_RISCV_ALIGN suppress ALIGN synthesis

MaskRay:
PR #151639 skipped synthesizing the section-start R_RISCV_ALIGN whenever
any R_RISCV_ALIGN existed at offset 0, regardless of its alignment. This
works with newer LLVM integrated assembler (#150816).

However, older MC and GNU Assembler as of today
(https://sourceware.org/bugzilla/show_bug.cgi?id=33236) can carry a weak
offset-0 R_RISCV_ALIGN (addend 2 => align 4) while its real alignment
requirement comes from a `.option norelax` .balign, which emits no
relocation.

Fix the condition to not suppress synthesis.

Link: https://sourceware.org/bugzilla/show_bug.cgi?id=33236#c4

### DIFF
--- a/lld/ELF/Arch/LoongArch.cpp
+++ b/lld/ELF/Arch/LoongArch.cpp
@@ -980,12 +980,19 @@ bool LoongArch::synthesizeAlignForInput(uint64_t &dot, InputSection *sec,
       }
     }
   } else if (sec->addralign > 4) {
-    // If the alignment is > 4 and the section does not start with an ALIGN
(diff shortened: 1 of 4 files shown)


MaskRay:
@MQ-mengqing

wangleiat (approved):
LGTM from the LoongArch side, thanks.

Issue #71001, opened 2023-11-01
clang claims weak aliases are not supported on Darwin but only when using __attribute__, works fine with #pragma


CodingMarkus:
clang claims weak aliases are not supported on Darwin. The following code doesn't compile:

```
#include <stdio.h>

int func1( ) { return 42; }
__attribute__((weak, alias("func1")))  int func2( );

int main( )
{
	printf("%d\n", func2());
	return 0;
}
```

Output:
```
test1.c:4:22: error: aliases are not supported on darwin
__attribute__((weak, alias("func1")))  int func2( );
                     ^
1 error generated.
```

But that's not true. Weak aliases are supported on Darwin. The following code will compile and will behave correctly:

```
#include <stdio.h>

int func1( ) { return 42; }

#pragma weak func2 = func1
int func2( );

int main( )
{
	printf("%d\n", func2());
	return 0;
}
```

CodingMarkus:
This could be seen as a dupe of #11488, albeit that report was about the error message which used to be `only weak aliases are supported on darwin` at that time and made even less sense to me, since it came when the code was trying to make exactly that, a weak alias.

shafik:
CC @rjmccall

PR #198148, opened 2026-05-17
[MachO] Preserve weak linkage for aliases

### LINKED ISSUE #111321: Weak aliases on MacOS aren't made global/public
Sample code:

```c
int bar();
int foo() { return 20; }

#pragma weak bar = foo
```

This should make a symbol that is linkable from external libraries, but instead `bar` gets placed in private text. Attempting to link fails.

```text
$ nm a.o
0000000000000000 t _bar
0000000000000000 T _foo
0000000000000000 t ltmp0
0000000000000008 s ltmp1
$ clang b.c a.o    # b.c just tries to call `bar()`
Undefined symbols for architecture arm64:
  "_bar", referenced from:
      _main in b-457c5f.o
ld: symbol(s) not found for architecture arm64
```

On Linux, `bar` correctly gets made a public weak symbol:

```text
0000000000000000 W bar
0000000000000000 T foo
```

This functionality should be supported in MacOS:

- https://github.com/llvm/llvm-project/issues/71001
- https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html

### LINKED ISSUE #196047: [Mach-O] wrong lowering for alias with weak linkage
(Reproduces on LLVM version 22.1.4)

```llvm
; ModuleID = 'repro.ll'
source_filename = "repro.ll"
target datalayout = "e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-n32:64-S128-Fn32"
target triple = "aarch64-apple-macosx13.0.0-unknown"

define internal void @foo_internal() {
  ret void
}

@foo_external = weak_odr hidden alias void (), ptr @foo_internal
```
```console
$ llc repro.ll -filetype=obj -o repro.o
$ llvm-nm -m repro.o
0000000000000000 (__TEXT,__text) private external _foo_external
0000000000000000 (__TEXT,__text) non-external _foo_internal
0000000000000000 (__TEXT,__text) non-external ltmp0
0000000000000008 (__LD,__compact_unwind) non-external ltmp1
```
I expect `_foo_external` to be marked as `weak external`, but it is marked as `private external`.

Oddly, if the alias has `default` visibility, then the symbol isn't even marked as `external` anymore:
```diff
-@foo_external = weak_odr hidden alias void (), ptr @foo_internal
+@foo_external = weak_odr alias void (), ptr @foo_internal
```
```console
$ llc repro.ll -filetype=obj -o repro.o
$ llvm-nm -m repro.o
0000000000000000 (__TEXT,__text) non-external _foo_external
0000000000000000 (__TEXT,__text) non-external _foo_internal
0000000000000000 (__TEXT,__text) non-external ltmp0
0000000000000008 (__LD,__compact_unwind) non-external ltmp1
```

Of course, if `@foo_external` is a separate function instead of an alias, it works as expected:
```diff
-@foo_external = weak_odr hidden alias void (), ptr @foo_internal
+define weak_odr hidden void @foo_external() { ret void }
```
```console
$ llc repro.ll -filetype=obj -o repro.o
$ llvm-nm -m repro.o
0000000000000004 (__TEXT,__text) weak private external _foo_external
0000000000000000 (__TEXT,__text) non-external _foo_internal
0000000000000000 (__TEXT,__text) non-external ltmp0
0000000000000008 (__LD,__compact_unwind) non-external ltmp1
```

syhhyl:
Mach-O aliases with weak or linkonce linkage were emitted as weak references, which is appropriate for undefined references but not for alias definitions. Emit Mach-O aliases through the same linkage path as other global definitions so weak aliases get .weak_definition.

When writing aliased symbols, keep the aliasee flags and include the alias symbol's own flags so N_WEAK_DEF is preserved in the Mach-O n_desc field.

Fixes #111321

#196047 was closed as a duplicate of #111321.

### DIFF
--- a/llvm/lib/MC/MachObjectWriter.cpp
+++ b/llvm/lib/MC/MachObjectWriter.cpp
@@ -443,7 +443,12 @@ void MachObjectWriter::writeNlist(MachSymbolData &MSD, const MCAssembler &Asm) {
   // The Mach-O streamer uses the lowest 16-bits of the flags for the 'desc'
   // value.
   bool EncodeAsAltEntry = IsAlias && OrigSymbol.isAltEntry();
-  W.write<uint16_t>(Symbol->getEncodedFlags(EncodeAsAltEntry));
+  uint16_t Flags = Symbol->getEncodedFlags(EncodeAsAltEntry);
+  // Preserve the aliasee's flags while adding alias-specific flags,
+  // such as N_WEAK_DEF emitted by .weak_definition.
+  if (IsAlias)
+    Flags |= OrigSymbol.getEncodedFlags(EncodeAsAltEntry);
+  W.write<uint16_t>(Flags);
   if (is64Bit())
     W.write<uint64_t>(Address);
   else
--- a/llvm/lib/MC/MachObjectWriter.cpp
+++ b/llvm/lib/MC/MachObjectWriter.cpp
@@ -443,7 +443,10 @@ void MachObjectWriter::writeNlist(MachSymbolData &MSD, const MCAssembler &Asm) {
   // The Mach-O streamer uses the lowest 16-bits of the flags for the 'desc'
   // value.
   bool EncodeAsAltEntry = IsAlias && OrigSymbol.isAltEntry();
-  W.write<uint16_t>(Symbol->getEncodedFlags(EncodeAsAltEntry));
+  uint16_t Flags = Symbol->getEncodedFlags(EncodeAsAltEntry);
+  if (IsAlias)
+    Flags |= OrigSymbol.getEncodedFlags(EncodeAsAltEntry);
+  W.write<uint16_t>(Flags);
   if (is64Bit())
     W.write<uint64_t>(Address);
   else
--- a/llvm/lib/CodeGen/AsmPrinter/AsmPrinter.cpp
+++ b/llvm/lib/CodeGen/AsmPrinter/AsmPrinter.cpp
@@ -2640,7 +2640,9 @@ void AsmPrinter::emitGlobalAlias(const Module &M, const GlobalAlias &GA) {
     return;
   }
 
-  if (GA.hasExternalLinkage() || !MAI.getWeakRefDirective())
+  if (MAI.isMachO())
+    emitLinkage(&GA, Name);
+  else if (GA.hasExternalLinkage() || !MAI.getWeakRefDirective())
     OutStreamer->emitSymbolAttribute(Name, MCSA_Global);
   else if (GA.hasWeakLinkage() || GA.hasLinkOnceLinkage())
     OutStreamer->emitSymbolAttribute(Name, MCSA_WeakReference);
--- a/llvm/lib/CodeGen/AsmPrinter/AsmPrinter.cpp
+++ b/llvm/lib/CodeGen/AsmPrinter/AsmPrinter.cpp
@@ -2646,7 +2646,9 @@ void AsmPrinter::emitGlobalAlias(const Module &M, const GlobalAlias &GA) {
     return;
   }
 
-  if (GA.hasExternalLinkage() || !MAI.getWeakRefDirective())
+  if (MAI.isMachO())
+    emitLinkage(&GA, Name);
+  else if (GA.hasExternalLinkage() || !MAI.getWeakRefDirective())
     OutStreamer->emitSymbolAttribute(Name, MCSA_Global);
   else if (GA.hasWeakLinkage() || GA.hasLinkOnceLinkage())
     OutStreamer->emitSymbolAttribute(Name, MCSA_WeakReference);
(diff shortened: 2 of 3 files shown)


AsakuraMizu:
This also appears to fix #111321.

cerisier:
Bumping this for reviewers as this is a quite important fix 🙌

syhhyl:
Friendly ping @MaskRay @arichardson. Could one of you take a look when you have time?

llvm/lib/MC/MachObjectWriter.cpp
```diff
@@ -443,7 +443,10 @@ void MachObjectWriter::writeNlist(MachSymbolData &MSD, const MCAssembler &Asm) {
   // The Mach-O streamer uses the lowest 16-bits of the flags for the 'desc'
   // value.
   bool EncodeAsAltEntry = IsAlias && OrigSymbol.isAltEntry();
-  W.write<uint16_t>(Symbol->getEncodedFlags(EncodeAsAltEntry));
+  uint16_t Flags = Symbol->getEncodedFlags(EncodeAsAltEntry);
+  if (IsAlias)
+    Flags |= OrigSymbol.getEncodedFlags(EncodeAsAltEntry);
```
folkertdev, line 450: Can you put this in a comment here, it's not obvious.

> When writing aliased symbols, keep the aliasee flags and include the alias symbol's own flags so N_WEAK_DEF is preserved in the Mach-O n_desc field.

folkertdev (commented):
What's been done in terms of testing here? I know `rustc` needs this for the EEI (externally implementable items) feature, did someone test this end-to-end?

syhhyl:
I performed the requested end-to-end testing with rustc at eff8269f797067c30555e77f160ec84c0ed15cd9 on Apple arm64.

I applied this LLVM change to the LLVM fork used by rustc, temporarily disabled rustc's Apple-specific restriction on static EII defaults for testing, and ran the following stage 1 tests:

- tests/ui/eii/static/default.rs
- tests/ui/eii/static/default_cross_crate.rs
- tests/ui/eii/static/default_cross_crate_explicit.rs

All three tests passed. In particular, the cross-crate test successfully compiled, linked, and ran, covering the original failure mode where the weak alias was not externally linkable.

llvm/lib/MC/MachObjectWriter.cpp
syhhyl: Added

folkertdev (approved):
LGTM

Let's wait a couple more days for if the other reviewers do want to respond. Then, we'll see if the build bots find anything, and if not I propose we cherry-pick this into LLVM 23, given that it's an important fix and unblocks a bunch of development on the rust side.

tgross35:
Might be good to add a Clang test? To cover https://github.com/llvm/llvm-project/issues/111321 end-to-end

ahmedbougacha:
I see this got merged as I was reading it, but:  can anyone here articulate how exactly you think aliases work on mach-o, why the diagnostics explicitly saying these aren't supported were originally added to clang, and why this is the only problem needing to be fixed for aliases to work?

#111321 links to docs about weak-linking, not aliases, and to #71001.  That, in turn, has an example claiming it works correctly using the pragma, but it doesn't, if you actually try to define the weak symbol externally (though it can be nudged into working with differently-named symbols, surely a hint that there's more to it than syntax)

folkertdev:
> Might be good to add a Clang test? To cover https://github.com/llvm/llvm-project/issues/111321 end-to-end

Clang does not really test end-to-end, but we could test the clang to LLVM IR too. (@syhhyl can you add that as a separate PR and link it here?)

---

The original commit specifically does say that weak aliases are supported, see https://github.com/llvm/llvm-project/commit/0017c5fa92ad:

> only weak aliases are supported on darwin

From reading the comments that is not disputed, and it demonstrably does work in practice.

MaskRay:
I think this is correct and makes the behavior similar to ELF. Yes, you can define a local symbol and a non-local alias to it.

folkertdev:
Looks like caused no issues, as expected, and backporting the fix means we can move on with this feature in rustc

/cherry-pick 5a12ba15b25edbd45364129783aca3809a236708

folkertdev:
ah right we don't rebase here

/cherry-pick 29575a3

AsakuraMizu:
Seems to be a spurious error?

/cherry-pick 29575a3

MaskRay:
I am still unclear what happened on this thread, but Mach-O features like this definitely need to get Apple's blessing.

folkertdev:
> Mach-O features like this definitely need to get Apple's blessing.

What route should be taken to get that blessing? Opening an issue is clearly not sufficient, they stay open for years, opening a PR doesn't seem to be enough either. Are there specific labels that help, or specific people that should be tagged?

cofibrant:
> What route should be taken to get that blessing? Opening an issue is clearly not sufficient, they stay open for years, opening a PR doesn't seem to be enough either.

We were aware of this PR internally and discussions were happening.

> Are there specific labels that help, or specific people that should be tagged?

Feel free to tag me and (while not an expert) I can help work out who needs to take a look.

folkertdev:
> We were aware of this PR internally and discussions were happening.

Checking in here, any updates?

dmcyk:
@folkertdev I've commented on #212001

PR #198152, opened 2026-05-17
[clangd] Prefer .hpp files over .h with header source switch

vbvictor:
Previously, The "Switch Between Source/Header" action picked `.h` over `.hpp` when both files existed next to a `.cpp` file, because `.h` is listed first in the header-extension list. 

This patch reorders `HeaderExtensions` and `SourceExtensions` so the `C++`-flavored extensions come before `.h` and `.c`. `C++`-flavor of file is preffered since (at least in my opinion) more people using `clangd` for `C++` than `C` with `.hpp` ext so switching from `.cpp` should go into `.hpp`, not `.h`.

This brings an edje case that when swithing from `.c` it will go into `.hpp` instead of `.h`, but I think this situation is more rare than having `.cpp` with `.hpp` and `.h` combination since `.h` headers can be used as `extern "C"` wrapper of cpp library.

Ideally, we should support a mapping between Header and Source (like `c <-> .h`, `.cxx, .cc <-> .hpp, hxx`, but it takes 50-100 lines of additional code and I don't think we should bring this complexity now.

### DIFF
--- a/clang-tools-extra/clangd/HeaderSourceSwitch.cpp
+++ b/clang-tools-extra/clangd/HeaderSourceSwitch.cpp
@@ -21,9 +21,9 @@ namespace clangd {
 std::optional<Path> getCorrespondingHeaderOrSource(
     PathRef OriginalFile, llvm::IntrusiveRefCntPtr<llvm::vfs::FileSystem> VFS) {
   static constexpr llvm::StringRef SourceExtensions[] = {
-      ".cpp", ".c", ".cc", ".cxx", ".c++", ".m", ".mm"};
+      ".cpp", ".cc", ".cxx", ".c++", ".c", ".m", ".mm"};
   static constexpr llvm::StringRef HeaderExtensions[] = {
-      ".h",    ".hh",  ".hpp",  ".hxx",  ".inc",
+      ".hpp",  ".hh",  ".hxx",  ".h++",  ".h",  ".inc",
       ".cppm", ".ccm", ".cxxm", ".c++m", ".ixx"};
 
   llvm::StringRef PathExt = llvm::sys::path::extension(OriginalFile);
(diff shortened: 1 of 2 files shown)


JVApen:
Can you explain why it is relevant to check the extension? Doesn't header/source switch look at the symbols? In what situation would the file extension be relevant?

vbvictor:
> Can you explain why it is relevant to check the extension? Doesn't header/source switch look at the symbols? In what situation would the file extension be relevant?

It first tries to look based on filenames (if files are in one directory) and then tries AST approach, here is relevant part:
https://github.com/llvm/llvm-project/blob/fda6309919a0c391ed224c0da214ba89ecff5478/clang-tools-extra/clangd/ClangdServer.cpp#L817-L835

In my particular case file structure was:
```
dir/
  context.h
  context.hpp
  context.cpp
```
So swithing from `context.cpp` got into `context.h`, but I think it should preferrably go into `context.hpp`.

-------------------------

Now I wonder if "symbols" approach would give `context.cpp` -> `context.hpp`. Maybe better solution for this problem is to make "symbols" approach go first with flag/config option even if its more expensive

ArcsinX (approved):
I think this patch makes sense, because `getCorrespondingHeaderOrSource()` is also used in the `DefineOutline` tweak. Without this patch, the priority of extensions is a little strange (cpp=>c=>cc), i.e. without this patch, when we move the definition from the header, if there are `.c` and `.cc` files with the same basename, then priority is given to `.c`, but if there is `.cpp`, then priority is given to `.cpp`. So, the logic as a whole is not very clear.
With this patch there is at least a consistent pattern that priority is given to C++ file extensions.

JVApen:
Thanks for the info. I think is is reasonable to make this change. Clangd already favors C++ over C, for example in the includes to standard library it provides. So moving C to the back of the list seems like a reasonable thing to do.

Throwing an alternative in the mix, maybe we should check for all file extensions. If we find more than 1 file matching stem.ext (excluding current file), we should bypass the file based approach. 
That way, the symbol approach will be taken and you will correct switching behavior for both stem.c to stem.h and stem.cpp to stem.hpp
(feel free to not implement this if you think it's wrong)

vbvictor:
> Throwing an alternative in the mix, maybe we should check for all file extensions. If we find more than 1 file matching stem.ext (excluding current file), we should bypass the file based approach.
That way, the symbol approach will be taken and you will correct switching behavior for both stem.c to stem.h and stem.cpp to stem.hpp
(feel free to not implement this if you think it's wrong)

This is indeed a good tradeoff, I can't promise to implement it right now but leave a todo for future myself

PR #198153, opened 2026-05-17
[lldb/Interpreter] Surface Python exceptions from scripted extensions

medismailben:
Python exceptions in scripted affordance methods were silently consumed by `llvm::consumeError`, leaving users with a generic "Failed to create script object." or just a null result.

`Dispatch` and `CallStaticMethod` now extract the Python backtrace via `PythonException::ReadBacktrace` and populate the `Status` out-parameter with "Python exception in <script-class> method '<method>':\n<traceback>". `ErrorWithMessage` no longer overwrites a Status that already carries a detailed message. For entry points with no error-return channel, the detailed error is broadcast via `Debugger::ReportError`.

A new test under lldb/test/API/functionalities/scripted_extensions/ listens for eBroadcastBitError and asserts each malformed extension produces a user-visible error.

rdar://161191350


lldb/include/lldb/Interpreter/Interfaces/ScriptedHookInterface.h
```diff
@@ -14,7 +14,7 @@
 #include "ScriptedInterface.h"
 
 namespace lldb_private {
-class ScriptedHookInterface : public ScriptedInterface {
+class ScriptedHookInterface : virtual public ScriptedInterface {
```
jimingham, line 17: I'm surprised we have Classes that derive from two of these Scripted***Interfaces.  What's an example of that?  Or was this just in case we ever did that?

lldb/include/lldb/Interpreter/Interfaces/ScriptedInterface.h
```diff
@@ -62,18 +62,20 @@ class ScriptedInterface {
...
+    // For user-facing messages, just pass through the Status if it already
+    // has detailed information (like Python tracebacks); otherwise set it.
+    llvm::StringRef existing_error = error.AsCString();
+    if (!error.Fail() || existing_error.empty()) {
+      // Status is empty, populate it with the simple error message.
+      error = Status::FromErrorString(error_msg.data());
+    }
+    // If Status already has content, leave it as-is (it has the Python
+    // traceback).
+
```
jimingham, line 77: Why do you leave the error as it is in this case rather than appending the extra message provided here to what was already in the error?

lldb/include/lldb/Interpreter/Interfaces/ScriptedHookInterface.h
medismailben: If I understood your question correctly, this is how every scripted python interface is built:
At the top, we have the `ScriptedInterface` class, which contains the CreatePluginObject method and the base members, then for every extension, there is an abstract base class for the interface with every method that extension exposes. You can think of it as a protocol in Swift for instance. Finally, you have the derived class that inherits from both the abstract interface class AND the scripting language interface class, that does all the heavy lifting of converting arguments and calling the script methods and convert them back etc.

I've explained this whole architecture in my EuroLLVM talk 2 years ago :) https://www.youtube.com/watch?v=hNvFmZ1iB78

lldb/include/lldb/Interpreter/Interfaces/ScriptedInterface.h
medismailben: For instance, the error could already contain the python backtrace so calling this method would prepend some message that is more user-friendly.

lldb/include/lldb/Interpreter/Interfaces/ScriptedHookInterface.h
jimingham: That talk didn't in fact mention that both ScriptedInterfaceBreakpoint and ScriptedInterfacePython (which seems more like a helper than an interface per se) both inherit from ScriptedInterface... 
But that's not the main point.  I couldn't see anything in this patch that would require you have a diamond graph rather than two copies of the ScriptedInterface data.  What required that?

lldb/include/lldb/Interpreter/Interfaces/ScriptedInterface.h
jimingham: Right, but the way this is written, if the incoming `error` is in a fail state, and already has text in it, this code won't add the string from `error_msg` to the `error`, it will just return it as is.
Didn't you want to prepend what's in `error_msg` TO the extant content in `error`?

lldb/include/lldb/Interpreter/Interfaces/ScriptedHookInterface.h
medismailben: > That talk didn't in fact mention that both ScriptedInterfaceBreakpoint and ScriptedInterfacePython (which seems more like a helper than an interface per se) both inherit from ScriptedInterface...

I don't think ScriptedInterfaceBreakpoint existed back then.

> I couldn't see anything in this patch that would require you have a diamond graph rather than two copies of the ScriptedInterface data. What required that?

Without the virtual inheritance, each leaf class (derived from 2 subclasses of `ScriptedInterface`) would see each of the `ScriptedInterface` members (`m_object_instance_sp`) & methods twice. Virtual inheritance collapses that into a single shared `ScriptedInterface` subobject at the top of the hierarchy so every path down to it resolves to the same instance.

lldb/include/lldb/Interpreter/Interfaces/ScriptedInterface.h
medismailben: Good call

lldb/include/lldb/Interpreter/Interfaces/ScriptedHookInterface.h
jimingham: I understood that, I just wondered what in this patch required that change - before the inheritance in ScriptInterpreterPython was virtual, but not in ScriptInterpreterBreakpoint, etc.  The two bits of data in ScriptInterface are shared pointers so you don't actually end up with two copies, just two references...

If this is just a drive-by cleanup, that's fine.  Just wanted to make sure I wasn't missing something.

jimingham:
This makes sense to me and is a good change.

There are two kinds of errors that we want to surface here.  One is errors where one of the implementation functions in the Python object returns something that doesn't make sense.  The other is where someone has mistyped the name of a function in their callback, which Python complains about at runtime.  It looks like all your errors are of the first sort, but it would be good to make sure the python runtime errors also come out readable.  Maybe add such a case to your test?

aperez:
Thanks for working on this @medismailben! I think the implementation largely subsumes my changes at https://github.com/llvm/llvm-project/pull/213181.

One path that does not appear to be covered yet in tests is an exception from a scripted static method. For example, a regression test could register a frame provider whose static `get_description()` raises and check that `HandleCommand` returns normally.

medismailben:
> Thanks for working on this @medismailben! I think the implementation largely subsumes my changes at #213181.
> 
> One path that does not appear to be covered yet in tests is an exception from a scripted static method. For example, a regression test could register a frame provider whose static `get_description()` raises and check that `HandleCommand` returns normally.

@aperez I cherry-picked your changes from #213181 to this PR, that's definitely something we don't want to regress. Feel free to approve this if it looks good to you.

aperez (approved):

jimingham (approved):
This LGTM too.

PR #198160, opened 2026-05-17
[CI] Fix GH PR Greeter

banach-space:
Add missing argument in the PR Greeter invocation. Follow-up for #197140. Issue reported here: https://discourse.llvm.org/t/ci-failure-prgreeter-on-my-first-pr

Also, as per https://docs.github.com/en/actions/concepts/security/script-injections + https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable, make sure that that greeter relies on ENV variables for input arguments.

### DIFF
--- a/.github/workflows/new-prs.yml
+++ b/.github/workflows/new-prs.yml
@@ -47,12 +47,16 @@ jobs:
 
       - name: Greet Author
         working-directory: ./llvm/utils/git/
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ISSUE_NUMBER: ${{ github.event.pull_request.number }}
+          PR_AUTHOR: ${{ github.event.pull_request.user.login }}
         run: |
           python3 ./github-automation.py \
-            --token '${{ secrets.GITHUB_TOKEN }}' \
+            --token "$GH_TOKEN" \
             pr-greeter \
-            --issue-number "${{ github.event.pull_request.number }}" \
-            --author "${{ github.event.pull_request.user.login }}"
+            --issue-number "$ISSUE_NUMBER" \
+            --author "$PR_AUTHOR"
 
   automate-prs-labels:
     # Greet first so that only the author gets that notification.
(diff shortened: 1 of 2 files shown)


.github/workflows/new-prs.yml
```diff
@@ -51,7 +51,8 @@ jobs:
           python3 ./github-automation.py \
             --token '${{ secrets.GITHUB_TOKEN }}' \
             pr-greeter \
-            --issue-number "${{ github.event.pull_request.number }}"
+            --issue-number "${{ github.event.pull_request.number }}" \
+            --author "${{ github.event.pull_request.user.login }}"
```
boomanaiden154: Can we pass this as an env variable to avoid the security warning?

banach-space: Sorry for the delay, I wasn't quite sure what you meant . Looks like GitHub docs have a nice explanation: https://docs.github.com/en/actions/concepts/security/script-injections

I will be sending an update shortly.

banach-space:
For reference, I've tested this workflow in my fork of LLVM. Here's how the message will look after this change:

* https://github.com/banach-space/llvm-project/pull/4#issuecomment-4507079681

DavidSpickett:
In theory we only need the env var for the things you could inject, but doing it for all of them saves a debate doesn't it.

Please link to ~https://docs.github.com/en/actions/reference/security/secure-use#good-practices-for-mitigating-script-injection-attacks~ in the description as well, I think that's the specific approach you're using here.

Edit: https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable is the correct link

DavidSpickett (approved):
LGTM, give Aiden a chance to look too.

...and for this security bot too? I'm not sure how it works.

boomanaiden154:
> ...and for this security bot too? I'm not sure how it works.

It should run on every update to a PR that touches something under `.github/workflows`.

boomanaiden154 (approved):
LGTM.

PR #198161, opened 2026-05-17
[CUF] Handle renamed managed companion pointer in `CUFDeviceAddressOpConversion`

ergawy:
CUFAddConstructor creates a companion pointer global (`@sym.managed.ptr`) for
each non-allocatable managed variable. `CompilerGeneratedNamesConversion` may
run before `CUFOpConversionLate` and rename the global by replacing dots with
'X', producing `@symXmanagedXptr`. Extend `CUFDeviceAddressOpConversion` to try
both the original and the renamed suffix when looking up the companion pointer.
    
Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


wangzpgi:
LGTM, I will let @clementval decide.

clementval:
Could we just pass the information in CUFOpConversionLate that renaming happened and it has to look for the companion pointer with X instead of . ?

ergawy:
> Could we just pass the information in CUFOpConversionLate that renaming happened and it has to look for the companion pointer with X instead of . ?

As discussed on Slack, I update `CUFOpConversionLate` to handle both the original suffix and whatever the result of renaming from `CompilerGeneratedNamesConversion` might be using the same util that it uses: `fir::NameUniquer::replaceSpecialSymbols`.

clementval (approved):
LGTM

Issue #198165, opened 2026-05-17
[X86][AVX512] suboptimal codegen with llvm.masked.load.v80i32


Andarwinux:
https://avx512.godbolt.org/z/hGxsqs9z8
```llvm
define <80 x i32> @av_get_padded_bits_per_pixel(<80 x i1> %0) {
  %2 = call <80 x i32> @llvm.masked.load.v80i32.p0(ptr null, <80 x i1> %0, <80 x i32> zeroinitializer)
  ret <80 x i32> %2
}

; Function Attrs: nocallback nofree nosync nounwind willreturn memory(argmem: read)
declare <80 x i32> @llvm.masked.load.v80i32.p0(ptr captures(none), <80 x i1>, <80 x i32>) #0

attributes #0 = { nocallback nofree nosync nounwind willreturn memory(argmem: read) }
```

RKSimon:
Codegen is better since #198166 and #198293 but will be a lot better once #198175 lands

Issue #198154, opened 2026-05-17
[X86][AVX512] Suboptimal codegen with llvm.masked.store.v192i8


Andarwinux:
original C code: https://test.godbolt.org/z/x19oc35f5

reduced IR: https://test.godbolt.org/z/Yfcxcrbzv

```llvm
define void @UpsampleBgrLinePair_C(<192 x i1> %interleaved.mask474) {
entry:
  br label %vector.body

vector.body:                                      ; preds = %vector.body, %entry
  tail call void @llvm.masked.store.v192i8.p0(<192 x i8> zeroinitializer, ptr null, <192 x i1> %interleaved.mask474)
  br label %vector.body
}

; Function Attrs: nocallback nofree nosync nounwind willreturn memory(argmem: write)
declare void @llvm.masked.store.v192i8.p0(<192 x i8>, ptr captures(none), <192 x i1>) #0

attributes #0 = { nocallback nofree nosync nounwind willreturn memory(argmem: write) }

```

Andarwinux:
#198166 did indeed improve the reduced IR, but for some reason the original C code still contains suboptimal sequence, maybe because llvm-reduce excessively reduced it?

Andarwinux:
I tried reduce again and got a slightly different IR that cannot be improved by #198166.

```llvm
define void @UpsampleBgrLinePair_C(<192 x i1> %interleaved.mask) {
iter.check:
  br label %vector.body

vector.body:                                      ; preds = %vector.body, %iter.check
  %index = phi i64 [ 0, %iter.check ], [ %index.next, %vector.body ]
  %0 = and i64 %index, 1
  %1 = getelementptr i8, ptr null, i64 %0
  tail call void @llvm.masked.store.v192i8.p0(<192 x i8> zeroinitializer, ptr %1, <192 x i1> %interleaved.mask)
  %index.next = add i64 %index, 1
  br label %vector.body
}

; Function Attrs: nocallback nofree nosync nounwind willreturn memory(argmem: write)
declare void @llvm.masked.store.v192i8.p0(<192 x i8>, ptr captures(none), <192 x i1>) #0

attributes #0 = { nocallback nofree nosync nounwind willreturn memory(argmem: write) }
```

PR #198166, opened 2026-05-17
[X86] LowerBUILD_VECTORvXi1 - attempt to fold as VPTESTMB(BUILD_VECTOR_vXi8(X),1)

RKSimon:
i1 scalar elements will be legalised to i8 (and the BUILD_VECTOR relies on implicit truncation) - but it will often be cheaper to perform the BUILD_VECTOR as a vXi8 and then perform a comparison to convert to the vXi1 mask, assuming we're inserting more than one non-constant i1 element.

Without BWI we have to extend this to vXi32 types to perform the comparison.

There's probably a lot we can do here (v2i8/v4i8/v8i8 types), but this patch at least addresses the worst codegen cases.

Fixes #179334


Andarwinux:
I can confirm this also improved #198165 and #198154.

phoebewang (approved):
LGTM.

PR #194856, opened 2026-04-29
[Clang] Fix assertion when __block is used on global variables in C mode

TPPPP72:
This is a reland PR, related to #183988 

I added an extra check in handleBlocksAttr to ensure that illegal Decl values ​​are not passed to downstream functions.
And remove unnecessary check in `CheckCompleteVariableDeclaration`.
Fixes #183974

Also added a extra regression test.

PR #198167, opened 2026-05-17
[Clang] Improve __block attribute coverage for ivars and static variables

### LINKED ISSUE #197213: [Clang] Improve __block attribute coverage for ivars and static variables
## Description
As discussed in #194856, we need to improve the diagnostic coverage for the `__block` attribute. Currently, `clang/test/Sema/block-misc.c` and related tests lack coverage for the following cases:

- Objective-C ivars
- Local static variables
- C++ static data members

I am opening this issue as a tracker/todo list to ensure these cases are properly diagnosed and tested in a follow-up PR.

CC @AaronBallman @ahatanak

TPPPP72:
As discussed in #194856, we need to improve the diagnostic coverage for the `__block` attribute.

The modifications I made are as follows:
1. added diagnostic definitions
2. modified diagnostic logic
3. added test cases
4. modified the affected test cases

close #197213


clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -11381,6 +11381,12 @@ def err_carries_dependency_param_not_function_decl : Error<
   "declaration or lambda">;
 def err_block_on_nonlocal : Error<
   "__block attribute not allowed, only allowed on local variables">;
+def err_block_on_static_local : Error<
+  "__block attribute not allowed on a static local variable">;
+def err_block_on_cxx_static_member : Error<
+  "__block attribute not allowed on a C++ static data member">;
+def err_block_on_objc_ivar : Error<
+  "__block attribute cannot be specified on an Objective-C instance variable">;
```
Sirraide: At this point, could you maybe refactor all of these warnings to a single warning along the lines of `"'__block' is not allowed on ..."` and use `%enum_select` here? Because 5 warnings for this is a bit of a mess... 

Also 2 minor nits here: 1. `__block` should be in single quotes since it’s code and 2. I don’t think we should call it an attribute in the diagnostic since that’s an implementation detail; however, if changing this for the existing diagnostics breaks like 100 tests then feel free to leave those as-is if updating the tests would be too much work...

clang/test/Sema/block-misc.m
```diff
@@ -219,3 +219,13 @@ void test21(void) {
 const char * (^func)(void) = ^{ return __func__; };
 const char * (^function)(void) = ^{ return __FUNCTION__; };
 const char * (^pretty)(void) = ^{ return __PRETTY_FUNCTION__; };
+
+@interface test22 {
```
Sirraide: I’d rather keep the objective-c tests in a separate file as was the case before

clang/lib/Sema/SemaObjC.cpp
```diff
@@ -1711,9 +1711,20 @@ void SemaObjC::handleBlocksAttr(Decl *D, const ParsedAttr &AL) {
     return;
   }
 
-  VarDecl *VD = dyn_cast<VarDecl>(D);
-  if (VD && !VD->hasLocalStorage()) {
-    Diag(AL.getLoc(), diag::err_block_on_nonlocal) << AL;
+  if (const auto *VD = dyn_cast<VarDecl>(D)) {
+    if (!VD->hasLocalStorage()) {
```
Sirraide, line 1721: What about `register` variables? Can those be `__block`? Do we have tests for that?

TPPPP72: > What about `register` variables? Can those be `__block`? Do we have tests for that?

I see...

```
repro.cpp:1:10: error: __block attribute not allowed, only allowed on local variables
    1 | register __block int a = 5;
```

Do we need a more detailed diagnose?

Sirraide (approved):
Implementation lgtm; please also add a release note for this, probably under ‘Improvements to Clang’s Diagnostics’

TPPPP72:
@ahatanak CC

clang/docs/ReleaseNotes.rst
```diff
@@ -510,6 +510,8 @@ Improvements to Clang's diagnostics
 
 - Added warnings for floating-point exception function calls (fenv.h) without enabling
   floating-point exception behavior via the appropriate flags or pragmas. (#GH128239)
+  
+- Improved ``__block`` attribute coverage for ivars, static variables and register variables. (#GH197213)
```
Sirraide:
```suggestion
- The diagnostics around ``__block`` now explain why a variable cannot be marked ``__block``. (#GH197213)
```

clang/docs/ReleaseNotes.rst
```diff
@@ -508,6 +508,11 @@ Improvements to Clang's diagnostics
 
 - Added ``-Wattribute-alias`` to diagnose type mismatches between an alias and its aliased function. (#GH195550)
 
+- Added warnings for floating-point exception function calls (fenv.h) without enabling
```
shafik, line 511: was this intended? It looks very unrelated.

Sirraide: Hmm, that looks like either a merge gone wrong or github’s diff being weird

Sirraide: Ah, looks like the patch that added this release note was reverted (#198341) and then this patch somehow added it back; I’ll just remove it

Issue #143129, opened 2025-06-06
[Sema] Memory leak in `SemaOverload.cpp`


lux-QAQ:
### **Description**

Hello, I've observed a memory leak in `SemaOverload.cpp` that occurs during template overload resolution.

My investigation **suggests a possible root cause** related to the lifecycle of `DeductionFailureInfo` objects. **It appears that** these objects, created to store details about why a template deduction failed, are allocated using placement `new (Context)` but are not explicitly deallocated when their owning `OverloadCandidate` is destroyed.

**If this hypothesis is correct, it could explain why** a definite leak occurs for members that perform their own heap allocations, such as the `SmallVector` within `clang::ConstraintSatisfaction`. The destructor for `ConstraintSatisfaction` **would presumably not be called**, which would in turn orphan its internal buffer, matching the behavior shown in the ASan report.

---
### **ASan Report**
[build2.log](https://github.com/user-attachments/files/20628340/build2.log)
[build1.log](https://github.com/user-attachments/files/20628341/build1.log)
This ASan report clearly shows the leak originating from `MakeDeductionFailureInfo` during the handling of unsatisfied constraints.

```cpp
==386608==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 7992 byte(s) in 37 object(s) allocated from:
    #0 0x55c8fab23dc3 in malloc (...)
    #1 0x7f81bb63dbcb in safe_malloc (...)
    #2 0x7f81bb63dbcb in llvm::SmallVectorBase<unsigned int>::grow_pod(...)
    #3 ...
    #7 0x7f81c86b806f in clang::ConstraintSatisfaction::operator=(...) /home/****/llvminstall/clang/include/clang/AST/ASTConcept.h:36:7
    #8 0x7f81c942c8fe in clang::MakeDeductionFailureInfo(clang::ASTContext&, ...) /home/****/llvminstall/clang/lib/Sema/SemaOverload.cpp:804:25
    #9 0x7f81c9952e56 in getPatternForClassTemplateSpecialization (...)
    #10 0x7f81c9952e56 in clang::Sema::InstantiateClassTemplateSpecialization(...)
    ...
```
---
### **Trigger Context and Build Failure**

The ASan reports are generated during the build process itself, specifically when the newly-built Clang attempts to compile tests for `libcxx`. The sequence of events is as follows:

1.  `ninja` starts executing the compilation command for a specific test file, for instance:
    ```sh
[...]

PR #198168, opened 2026-05-17
[Clang][Sema] Store constraints-not-satisfied deduction info as ASTConstraintSatisfaction

lux-QAQ:
When I checked #143129, I found that the issue had been fixed. My manual experiment revealed that` static_cast<CNSInfo *>(Data)->Satisfaction.~ConstraintSatisfaction();` fixed the problem. Removing this line of code reproduced the ASan/LSan leak report. But the explicit destructor call here feels too much like manual memory management: it works, but it is easy to miss and makes the lifetime model more fragile. That is why I proposed this alternative solution in this PR.

Using ASTConstraintSatisfaction might cause an increase in RSS because ASTContext arena storage is released when the ASTContext is destroyed. To check for this potential memory increase, I constructed some highly specific synthetic tests, but further testing using llvm-test-suite showed that max-rss did not increase or decrease significantly. This indicates that my previous synthetic tests were too narrow and not general enough. I also added test data from llvm-test-suite.

### Memory Impact

I removed the synthetic stress-case numbers and kept only llvm-test-suite CTMark results.

I ran CTMark with fresh build directories, -j1, and TEST_SUITE_REPORT_COMPILE_MAX_RSS=ON. Each configuration was run 5 times.

| Config | Metric | Mean | Median | Min | Max |
|---|---:|---:|---:|---:|---:|
| O0-g | compile_time | -1.667% | -1.708% | -3.918% | +1.809% |
| O0-g | compile_maxrss | -0.016% | -0.066% | -0.211% | +0.234% |
| O0-g | link_maxrss | -0.024% | -0.069% | -0.256% | +0.299% |
| O0-g | size | -0.001% | -0.001% | -0.001% | -0.001% |
| O3 | compile_time | -0.551% | -1.365% | -10.463% | +10.017% |
| O3 | compile_maxrss | +0.065% | -0.006% | -0.355% | +0.534% |
| O3 | link_time | +2.076% | +1.157% | -19.125% | +26.917% |
| O3 | link_maxrss | +0.040% | -0.032% | -0.383% | +0.607% |
| O3 | size | +0.000% | +0.000% | +0.000% | +0.000% |

The baseline compiler was built from commit `bae540bde882aa1911ccffb0fe7691cb16831017`, and the patched compiler was built from the same commit plus this PR.

## AI Usage Report:

Assisted-by: ChatGPT 5.5

1. Used ChatGPT 5.5 to assist with translating my Chinese draft into English.

2. Used ChatGPT 5.5 to help locate and update source locations affected by changing

```cpp
ConstraintSatisfaction Satisfaction;
```
to
```cpp
const ASTConstraintSatisfaction *Satisfaction;
```
including places that needed to be changed to use Satisfaction as a pointer.

3. Used ChatGPT 5.5 to draft scripts to generate synthetic stress test cases.

I carefully reviewed and edited the AI-assisted text to the best of my ability before posting, and I am responsible for the final contents of this PR.

[archived_synthetic_stress_results.txt](https://github.com/user-attachments/files/27947894/archived_synthetic_stress_results_unreliable.txt)
[archived_synthetic_stress_reproducer.py](https://github.com/user-attachments/files/27947893/archived_synthetic_stress_reproducer.py)
[llvm-test-suite-result.txt](https://github.com/user-attachments/files/27947895/llvm-test-suite.txt)
[llvm-test-suite-result-summary.txt](https://github.com/user-attachments/files/27953366/llvm-test-suite-result-summary.txt)

### DIFF
--- a/clang/lib/Sema/SemaOverload.cpp
+++ b/clang/lib/Sema/SemaOverload.cpp
@@ -874,8 +874,6 @@ void DeductionFailureInfo::Destroy() {
 
   case TemplateDeductionResult::ConstraintsNotSatisfied:
     // FIXME: Destroy the template argument list?
-    // CNSInfo and ASTConstraintSatisfaction are ASTContext-allocated and do
-    // not own non-arena resources.
     Data = nullptr;
     if (PartialDiagnosticAt *Diag = getSFINAEDiagnostic()) {
       Diag->~PartialDiagnosticAt();
--- a/clang/lib/Sema/SemaOverload.cpp
+++ b/clang/lib/Sema/SemaOverload.cpp
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "CheckExprLifetime.h"
+#include "clang/AST/ASTConcept.h"
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/CXXInheritance.h"
 #include "clang/AST/Decl.h"
@@ -741,7 +742,7 @@ namespace {
   // unsatisfied constraints.
   struct CNSInfo {
     TemplateArgumentList *TemplateArgs;
-    ConstraintSatisfaction Satisfaction;
+    const ASTConstraintSatisfaction *Satisfaction;
   };
 }
 
@@ -822,9 +823,10 @@ clang::MakeDeductionFailureInfo(ASTContext &Context,
     break;
 
   case TemplateDeductionResult::ConstraintsNotSatisfied: {
-    CNSInfo *Saved = new (Context) CNSInfo;
-    Saved->TemplateArgs = Info.takeSugared();
-    Saved->Satisfaction = std::move(Info.AssociatedConstraintsSatisfaction);
+    auto *Saved = new (Context)
+        CNSInfo{Info.takeSugared(),
+                ASTConstraintSatisfaction::Create(
+                    Context, Info.AssociatedConstraintsSatisfaction)};
     Result.Data = Saved;
     break;
   }
@@ -872,7 +874,8 @@ void DeductionFailureInfo::Destroy() {
 
   case TemplateDeductionResult::ConstraintsNotSatisfied:
     // FIXME: Destroy the template argument list?
-    static_cast<CNSInfo *>(Data)->Satisfaction.~ConstraintSatisfaction();
+    // CNSInfo and ASTConstraintSatisfaction are ASTContext-allocated and do
+    // not own non-arena resources.
     Data = nullptr;
     if (PartialDiagnosticAt *Diag = getSFINAEDiagnostic()) {
       Diag->~PartialDiagnosticAt();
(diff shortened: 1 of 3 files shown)


clang/lib/Sema/SemaOverload.cpp
```diff
@@ -872,7 +874,8 @@ void DeductionFailureInfo::Destroy() {
 
   case TemplateDeductionResult::ConstraintsNotSatisfied:
     // FIXME: Destroy the template argument list?
-    static_cast<CNSInfo *>(Data)->Satisfaction.~ConstraintSatisfaction();
+    // CNSInfo and ASTConstraintSatisfaction are ASTContext-allocated and do
+    // not own non-arena resources.
```
zyn0217:
```suggestion
```

zyn0217 (commented):
Is this an AI art? PLEASE make sure you have read https://llvm.org/docs/AIToolPolicy.html carefully before we move on, thanks.

Despite that, I do wonder if there is a real-world benchmark showing the benefit we gain from this patch. Otherwise I don't think I understand how we gain those benefits? CNSInfo is already allocated on ASTContext, though @cor3ntin

lux-QAQ:
> Is this an AI art? PLEASE make sure you have read https://llvm.org/docs/AIToolPolicy.html carefully before we move on, thanks.
> 
> Despite that, I do wonder if there is a real-world benchmark showing the benefit we gain from this patch. Otherwise I don't think I understand how we gain those benefits? CNSInfo is already allocated on ASTContext, though @cor3ntin

I apologize for not reading the AIToolPolicy carefully before; I have now read it thoroughly.

I have rewritten the PR description myself, reviewed the modified code again, and added an AI usage report.

Regarding the issue of the "benefit we gain from this patch," I repeatedly collected data using llvm-test-suite and it did not show any obvious issues. max RSS and other metrics didn't show significant decreases or increases, indicating that the previous specially constructed tests were indeed not very representative. But I think the real benefit here is that it eliminates the explicit destructor call, so we avoid having to manually manage the lifetime of these members.

zyn0217:
I meant we'd better test your patch with a concept heavy codebase, like range-v3, stdexec, etc.

The contrived code your benchmark runs against doesnt mean we always suffer from the exact issue this patch is trying to fix. I.e. we may not gain much because IMHO no one owns a C++ codebase with a gigantic number of function calls over an exaggerated overload candidate set (The performance benefit becomes significant when N>=500, right?)

cor3ntin (changes requested):
> When I checked https://github.com/llvm/llvm-project/issues/143129, I found that the issue had been fixed. My manual experiment revealed that static_cast<CNSInfo *>(Data)->Satisfaction.~ConstraintSatisfaction(); fixed the problem. Removing this line of code reproduced the ASan/LSan leak report. But the explicit destructor call here feels too much like manual memory management: it works, but it is easy to miss and makes the lifetime model more fragile. 

No.

You will notice that `DeductionFailureInfo::Destroy` does other "manual memory management".
We do not want diagnostics attached to an unselected overload to outlive overload resolution.

In fact, in the long term we should try to get rid of `ASTConstraintSatisfaction` all together and recomputing the diagnostic when producing it.
However, this is a lot more work so the status quo is that `ASTConstraintSatisfaction` exists but we should try to not use it unless we absolutely have to

lux-QAQ:
Thanks @cor3ntin for the clarification.

I understand the design now, and I'll close this PR.

Thanks again for the review.

PR #194414, opened 2026-04-27
[llvm-ir2vec] Breaking up llvm-ir2vec lib implementation to clean up MIR deps from ir2vec python bindings

nishant-sachdeva:
The Python bindings only expose IR2Vec functionality. MIR2Vec has no Python API. However, the single `LLVMEmbUtils` library bundled both IR2VecTool and MIR2VecTool, causing CodeGen and Target components to be linked into the nanobind module unnecessarily.

This patch splits the library along that boundary. LLVMIREmbUtils covers IR2Vec and is linked by both the CLI tool and the Python bindings. LLVMMIREmbUtils covers MIR2Vec and is linked only by the CLI tool.
Result: Python wheel size reduces from ~14 MB to ~4 MB.

PR #198170, opened 2026-05-17
[llvm-ir2vec] Creating one directory per library for IRUtils, and MIRUtils

nishant-sachdeva:
Follow up to https://github.com/llvm/llvm-project/pull/194414#issuecomment-4467455664 . 
Made a patch to move IRUtils, and MIRUtils to their own subdirectories

@svkeerthy  , @nico , @boomanaiden154


llvm/tools/llvm-ir2vec/Bindings/PyIR2Vec.cpp
```diff
@@ -6,7 +6,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "lib/IRUtils.h"
+#include "../lib/IRUtils/IRUtils.h"
```
svkeerthy: Here and other places - Can you not use relative include paths and set the include paths using `target_include_directories` like what we have in `Bindings/CMakelists.txt`?

nishant-sachdeva: Accepting this

svkeerthy (approved):

PR #198171, opened 2026-05-17
[llvm-ir2vec][NFC] Adding disclaimer to Bindings requirements.txt to check compatibility with ml-compiler-opt

nishant-sachdeva:
Follow up PR for https://github.com/llvm/llvm-zorg/pull/846#issuecomment-4467263196 

@mtrofin , @svkeerthy

### DIFF
--- a/llvm/tools/llvm-ir2vec/Bindings/requirements.txt
+++ b/llvm/tools/llvm-ir2vec/Bindings/requirements.txt
@@ -24,6 +24,11 @@
 # BEFORE UPDATING these requirements: verify tml-compiler-opt pipenv
 # can be applied first, then these requirements.
 #
+# Recommended verification workflow:
+#   1. Apply ml-compiler-opt's pipenv (Pipfile.lock) in a fresh venv.
+#   2. pip install the updated requirements on top of it.
+#   3. Run unit tests locally to confirm no breakage.
+#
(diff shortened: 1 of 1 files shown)


llvm/tools/llvm-ir2vec/Bindings/requirements.txt
```diff
@@ -1,2 +1,29 @@
-# BUILD dependencies
+# IR2Vec python bindings' dependencies
+#
+# BEFORE UPDATING: verify compatibility with ml-compiler-opt's Pipfile/Pipenv
```
mtrofin, line 3: nit: maybe explain how to verify, e.g. "we recommend creating a virtual env at at least one of the supported versions specified by ml-compiler-opt, applying that Pipenv, then applying the proposed requirements.txt (this file)". something like that. Because 9 out of 10, people just ignore the specified versioning and package requirements we spelled out on ml-compiler-opt.

mtrofin (approved):
lgtm, a comment.

llvm/tools/llvm-ir2vec/Bindings/requirements.txt
nishant-sachdeva: Accepting this. Pushed a commit to add recommended workflow to test compatibility

PR #198172, opened 2026-05-17
Add __builtin_bswap128

Disservin:
Add support for __builtin_bswap128. 
Hide the builtin on targets without __int128 support, for that purpose adds a new attribute, which in the future can also be used for __builtin_bitreverse128 which gcc now has too.

Aligns clang with gcc.

Took inspiration from https://reviews.llvm.org/D114425

### DIFF
--- a/clang/include/clang/Basic/Builtins.h
+++ b/clang/include/clang/Basic/Builtins.h
@@ -466,6 +466,11 @@ public:
     return strchr(getAttributesString(ID), 'G') != nullptr;
   }
 
+  /// Returns true if this builtin requires target support for __int128.
+  bool requiresInt128Type(unsigned ID) const {
+    return strchr(getAttributesString(ID), 'H') != nullptr;
+  }
+
 private:
   std::pair<const InfosShard &, const Info &>
   getShardAndInfo(unsigned ID) const;
--- a/clang/include/clang/Basic/Builtins.td
+++ b/clang/include/clang/Basic/Builtins.td
@@ -760,6 +760,12 @@ def BSwap : Builtin, Template<["unsigned short", "uint32_t", "uint64_t"],
   let Prototype = "T(T)";
 }
 
+def BSwap128 : Builtin {
+  let Spellings = ["__builtin_bswap128"];
+  let Attributes = [NoThrow, Const, Constexpr, RequireInt128];
+  let Prototype = "__uint128_t(__uint128_t)";
+}
+
 def BSwapg : Builtin {
   let Spellings = ["__builtin_bswapg"];
   let Attributes = [NoThrow, Const, Constexpr, CustomTypeChecking];
--- a/clang/include/clang/Basic/BuiltinsBase.td
+++ b/clang/include/clang/Basic/BuiltinsBase.td
@@ -95,6 +95,8 @@ class VScanfFormat<int I> : IndexedAttribute<"S", I>;
 def Constexpr : Attribute<"E">;
 // Builtin is immediate and must be constant evaluated. Implies Constexpr, and will only be supported in C++20 mode.
 def Consteval : Attribute<"EG">;
+// Builtin requires the target to support __int128.
+def RequireInt128 : Attribute<"H">;
 
 // Callback behavior: the first index argument is called with the arguments
 // indicated by the remaining indices.
--- a/clang/lib/AST/ASTContext.cpp
+++ b/clang/lib/AST/ASTContext.cpp
@@ -12916,6 +12916,11 @@ QualType ASTContext::GetBuiltinType(unsigned Id,
     return {};
   }
 
+  if (BuiltinInfo.requiresInt128Type(Id) && !getTargetInfo().hasInt128Type()) {
+    Error = GE_Missing_type;
+    return {};
+  }
+
   SmallVector<QualType, 8> ArgTypes;
 
   bool RequiresICE = false;
--- a/clang/lib/AST/ByteCode/InterpBuiltin.cpp
+++ b/clang/lib/AST/ByteCode/InterpBuiltin.cpp
@@ -4953,6 +4953,7 @@ bool InterpretBuiltin(InterpState &S, CodePtr OpPC, const CallExpr *Call,
   case Builtin::BI__builtin_bswap16:
   case Builtin::BI__builtin_bswap32:
   case Builtin::BI__builtin_bswap64:
+  case Builtin::BI__builtin_bswap128:
     return interp__builtin_bswap(S, OpPC, Frame, Call);
 
   case Builtin::BI__atomic_always_lock_free:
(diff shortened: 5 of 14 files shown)


philnik777 (commented):
Do we actually want to add this? There is `__builtin_bswapg` now, which is strictly better than the constant width builtins.

Disservin:
> Do we actually want to add this? There is `__builtin_bswapg` now, which is strictly better than the constant width builtins.

i leave that discussion to you :D 

seems like gcc and clang have kind of opposed opinions on this matter :D 

> Note, we don't have __builtin_bswapg and I don't think we should, one can
only byteswap something which has number of bits divisible by CHAR_BIT.

https://gcc.gnu.org/pipermail/gcc-patches/2026-May/716903.html

i was planing / about to open an issue for the `__builtin_bitreverse128` and came across this for bswap first

AaronBallman:
> > Do we actually want to add this? There is `__builtin_bswapg` now, which is strictly better than the constant width builtins.
> 
> i leave that discussion to you :D
> 
> seems like gcc and clang have kind of opposed opinions on this matter :D
> 
> > Note, we don't have __builtin_bswapg and I don't think we should, one can
> > only byteswap something which has number of bits divisible by CHAR_BIT.
> 
> https://gcc.gnu.org/pipermail/gcc-patches/2026-May/716903.html
> 
> i was planing / about to open an issue for the `__builtin_bitreverse128` and came across this for bswap first

I think it's worth supporting this because GCC supports it and libstdc++ is using it: https://github.com/gcc-mirror/gcc/blob/bc94521ad8f375b5e964a8e79c2dccc106d3322d/libstdc%2B%2B-v3/include/std/bit#L127

(FWIW, I don't see the issue with swapping non-value representation bits; this is the same situation as any other in which bits of a non-value representation are read or written. As far as use cases go, the use cases for swapping `_BitInt` are the same as for swapping a bit-field, so things like swaps for network byte ordering, etc. I don't think it's the most important thing in the world to support, but users can do shifting and masking to avoid reading those bits after a swap happens and so it is useful IMO.)

philnik777:
> > > Do we actually want to add this? There is `__builtin_bswapg` now, which is strictly better than the constant width builtins.
> > 
> > 
> > i leave that discussion to you :D
> > seems like gcc and clang have kind of opposed opinions on this matter :D
> > > Note, we don't have __builtin_bswapg and I don't think we should, one can
> > > only byteswap something which has number of bits divisible by CHAR_BIT.
> > 
> > 
> > https://gcc.gnu.org/pipermail/gcc-patches/2026-May/716903.html
> > i was planing / about to open an issue for the `__builtin_bitreverse128` and came across this for bswap first
> 
> I think it's worth supporting this because GCC supports it and libstdc++ is using it: https://github.com/gcc-mirror/gcc/blob/bc94521ad8f375b5e964a8e79c2dccc106d3322d/libstdc%2B%2B-v3/include/std/bit#L127

Is "libstdc++ uses it" really that big of a deal? It's not required for conformance, and I also don't see it being relevant for performance either, since the current implementation is folded to a bswap128 in the first InstCombine pass. Would you still consider it worth supporting if libstdc++ didn't use it anymore in six months? (FWIW Andew Pinski seems to want `__builtin_bswapg` given his response)

Disservin:
well libc++ also has it ? :D 
https://github.com/llvm/llvm-project/blob/0b806bca3f3fe0dbe4d70d46404e4978af1fd731/libcxx/include/__bit/byteswap.h#L38

philnik777:
> well libc++ also has it ? :D
> 
> https://github.com/llvm/llvm-project/blob/0b806bca3f3fe0dbe4d70d46404e4978af1fd731/libcxx/include/__bit/byteswap.h#L38

Yeah, and we'll replace it with `__builtin_bswapg` as part of the effort to support `_BitInt` (and would do it anyways, since that builtin is just better).

AaronBallman:
> > > > Do we actually want to add this? There is `__builtin_bswapg` now, which is strictly better than the constant width builtins.
> > > 
> > > 
> > > i leave that discussion to you :D
> > > seems like gcc and clang have kind of opposed opinions on this matter :D
> > > > Note, we don't have __builtin_bswapg and I don't think we should, one can
> > > > only byteswap something which has number of bits divisible by CHAR_BIT.
> > > 
> > > 
> > > https://gcc.gnu.org/pipermail/gcc-patches/2026-May/716903.html
> > > i was planing / about to open an issue for the `__builtin_bitreverse128` and came across this for bswap first
> > 
> > 
> > I think it's worth supporting this because GCC supports it and libstdc++ is using it: https://github.com/gcc-mirror/gcc/blob/bc94521ad8f375b5e964a8e79c2dccc106d3322d/libstdc%2B%2B-v3/include/std/bit#L127
> 
> Is "libstdc++ uses it" really that big of a deal? 

Because it's guarded, not really. But it demonstrates use in the wild.

> It's not required for conformance, and I also don't see it being relevant for performance either, since the current implementation is folded to a bswap128 in the first InstCombine pass. Would you still consider it worth supporting if libstdc++ didn't use it anymore in six months? (FWIW Andew Pinski seems to want `__builtin_bswapg` given his response)

https://sourcegraph.com/search?q=context:global+__builtin_bswap128&patternType=keyword&sm=0 doesn't show significant use in the wild (a number of the hits are inline implementations rather than calls), so if it isn't being used by major third-party libraries, I think it's a weaker case. But because it exists and is used, I don't see a huge burden in supporting it either. (Shouldn't these should lower to the generic call anyway?)

philnik777:
> https://sourcegraph.com/search?q=context:global+__builtin_bswap128&patternType=keyword&sm=0 doesn't show significant use in the wild (a number of the hits are inline implementations rather than calls), so if it isn't being used by major third-party libraries, I think it's a weaker case. But because it exists and is used, I don't see a huge burden in supporting it either. (Shouldn't these should lower to the generic call anyway?)

I'm not super strongly against this, since I agree that the maintenance burden isn't too high. OTOH I don't find anything here particularly compelling either, since we're clearly moving towards type-generic builtins and I don't expect that trend to reverse. AFAICT this is basically adding a builtin that will be (or at least I expect to be) fully obsolete in the near future.

Disservin:
> > well libc++ also has it ? :D
> > https://github.com/llvm/llvm-project/blob/0b806bca3f3fe0dbe4d70d46404e4978af1fd731/libcxx/include/__bit/byteswap.h#L38
> 
> Yeah, and we'll replace it with `__builtin_bswapg` as part of the effort to support `_BitInt` (and would do it anyways, since that builtin is just better).

now since, https://github.com/llvm/llvm-project/commit/fdfb99f6336401e534f611c4b2df5a338e26b997 we make use of __builtin_bswapg, but have the __builtin_bswap128 as a fallback in place if it's not available.. I don't have a big preference on this patch for now except make compatibility outside of std::byteswap

PR #198174, opened 2026-05-17
[SLP] Retry vectorization of FMA candidates after block processing

### LINKED ISSUE #198040: Regression: failure to slp-vectorize simple loop
The Chapel team noticed a regression in some of our benchmarks when upgrading to LLVM 22. I tracked down one of these regressions to a failure to vectorize in some key loops. https://godbolt.org/z/srbdqKxa7 has the reproducer. When I run `opt -passes=slp-vectorize` on that code with LLVM 21, it vectorizes the loop (`!llvm.loop !59`). With LLVM 22, it does not. This results in almost a 100% peformance degradation for this benchmark. 

https://github.com/chapel-lang/chapel/issues/28838 has details of the Chapel code used to get this LLVM IR.

I also created a very reduced LLVM IR that reproduces the issue. This code should vectorize (and did vectorize with LLVM 20 and 21), but fails to with LLVM 22

https://godbolt.org/z/M7jToa8aW
```llvm
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

define void @wrapcoforall_fn_DefaultRectangular_line_329_chpl7(double %0, double %1, i64 %2) #0 {
  br label %4

4:                                                ; preds = %4, %3
  %5 = phi double [ %25, %4 ], [ 0.000000e+00, %3 ]
  %6 = sitofp i64 0 to double
  %7 = fadd double 0.000000e+00, %6
  %8 = fadd double %7, 0.000000e+00
  %9 = fmul double %7, %8
  %10 = fmul double %9, 0.000000e+00
  %11 = fadd double %10, %0
  %12 = fadd double %11, 0.000000e+00
  %13 = fdiv double 0.000000e+00, %12
  %14 = fmul double %1, %13
  %15 = sitofp i64 %2 to double
  %16 = fadd double %0, %15
  %17 = fadd double %16, 0.000000e+00
  %18 = fmul double %16, %17
  %19 = fmul double %18, 0.000000e+00
  %20 = fadd double %19, 0.000000e+00
  %21 = fadd double %20, 0.000000e+00
  %22 = fdiv double 0.000000e+00, %21
  %23 = fmul double 0.000000e+00, %22
  %24 = fadd contract double %14, %23
  %25 = fadd double %5, %24
  br label %4
}

attributes #0 = { "target-cpu"="haswell" }

```

alexey-bataev:
fadd/fsub instructions that canConvertToFMA returns valid for were
unconditionally skipped in tryToVectorize, causing regressions
where SLP failed to vectorize loops containing such patterns even when
FMA formation never fires.
Collect skipped FMA candidates during vectorizeChainsInBlock and retry
them with AllowFMACandidates=true after all other instructions in the
block have been processed. The cost model still rejects the retry when
actual FMA formation is more profitable (e.g. FMA4 on bdver2), so
existing FMA-profitable cases are unaffected.

Fixes #198040


RKSimon (approved):
LGTM

llvm/include/llvm/Transforms/Vectorize/SLPVectorizer.h
```diff
@@ -98,12 +98,14 @@ struct SLPVectorizerPass : public OptionalPassInfoMixin<SLPVectorizerPass> {
                           bool MaxVFOnly = false);
 
   /// Try to vectorize a chain that may start at the operands of \p I.
-  bool tryToVectorize(Instruction *I, slpvectorizer::BoUpSLP &R);
+  bool tryToVectorize(Instruction *I, slpvectorizer::BoUpSLP &R,
+                      SmallSetVector<Instruction *, 8> &FMACandidates,
+                      bool AllowFMACandidates = false);
```
bababuck, line 103: Would it be cleaner if we just made `FMACandidates` a member of `BoUpSLP` so it doesn't need piped everywhere? I guess that downside is that `FMACandidates` is only needed in the scope of `vectorizeChainsInBlock`. Curious as to your reasoning though for my own learning.

bababuck:
Why not try to vectorize these instructions on the first pass as well? Won't the cost model still handle unprofitable cases then as well?

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -31678,8 +31694,17 @@ bool SLPVectorizerPass::vectorizeChainsInBlock(BasicBlock *BB, BoUpSLP &R) {
       if (ExtractsCost > PostProcessStores.size() + 1)
         return Changed;
```
bababuck: Won't these early exits for post processing stores cause us to not attempt vectorization of the `FMACandidates`?

alexey-bataev:
> Why not try to vectorize these instructions on the first pass as well? Won't the cost model still handle unprofitable cases then as well?

Because we need to try to vectorize them as FMAs at first. But CFG does not allow doing this, so we need to postpone it and do it later.

llvm/include/llvm/Transforms/Vectorize/SLPVectorizer.h
alexey-bataev: We can, but I don't see a strong preference here.

alexey-bataev:
Ping @bababuck !

bababuck:
> Ping @bababuck !

Did you have a chance to look at https://github.com/llvm/llvm-project/pull/198174#discussion_r3260776565?

alexey-bataev:
> > Ping @bababuck !
> 
> Did you have a chance to look at [#198174 (comment)](https://github.com/llvm/llvm-project/pull/198174#discussion_r3260776565)?

Addressed in the update

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
bababuck: Theres a couple more early returns still

bababuck: Early return

bababuck:
> Addressed in the update

Sorry I missed seeing that earlier. It looks like there are a couple other early returns hidden in that same chunk. This looks good to me though once those are addressed!

bababuck (approved):
LGTM

sushgokh:
This is causing 4.5% regression with SPEC17 538.imagemagik(C)  (-Ofast -flto)  (and perhaps 2% with 523.xalancbmk with -Ofast -msve-vector-bits=128) on AArch64 Neoverse-v2

sushgokh:
> This is causing 4.5% regression with SPEC17 538.imagemagik(C) (-Ofast -flto) (and perhaps 2% with 523.xalancbmk with -Ofast -msve-vector-bits=128) on AArch64 Neoverse-v2

@alexey-bataev [This ](https://godbolt.org/z/141MxK9rh) is short repro of the code causing regression in imagemagik

alexey-bataev:
> > This is causing 4.5% regression with SPEC17 538.imagemagik(C) (-Ofast -flto) (and perhaps 2% with 523.xalancbmk with -Ofast -msve-vector-bits=128) on AArch64 Neoverse-v2
> 
> @alexey-bataev [This ](https://godbolt.org/z/141MxK9rh) is short repro of the code causing regression in imagemagik

Sorry, forgot about this one. The actual regression is caused by the incorrect cost model for non-uniform shifts for AARCH64. It takes and extra instruction to materialize the constant and then does the shift, so the cost is not 1, but 2

PR #198178, opened 2026-05-17
[InstCombine] Convert logical and/or with trunc nuw to i1 into bitwise ops

andjo403:
if it is know that `trunc nuw to i1 ` can not be poison logical and/or can be folded to bitwise ops.

regression seen in https://github.com/dtcxzyw/llvm-opt-benchmark/pull/3528#discussion_r2874045398 from https://github.com/llvm/llvm-project/pull/184182

proof https://alive2.llvm.org/ce/z/xQ2Sj-

### DIFF
--- a/llvm/lib/Transforms/InstCombine/InstCombineSelect.cpp
+++ b/llvm/lib/Transforms/InstCombine/InstCombineSelect.cpp
@@ -3682,11 +3682,9 @@ static bool impliesPoisonOrCond(const Value *ValAssumedPoison, const Value *V,
   if (match(ValAssumedPoison, m_NUWTrunc(m_Value(A))) &&
       isGuaranteedNotToBePoison(A)) {
     assert(ValAssumedPoison->getType()->isIntOrIntVectorTy(1));
-    unsigned BitWidth = A->getType()->getScalarSizeInBits();
-    return ConstantRange(APInt::getZero(BitWidth), APInt(BitWidth, 2))
-        .contains(computeConstantRangeIncludingKnownBits(
-            A, false,
-            SQ.getWithInstruction(cast<Instruction>(ValAssumedPoison))));
+    return computeKnownBits(
+               A, SQ.getWithInstruction(cast<Instruction>(ValAssumedPoison)))
+               .getMaxValue() == 1;
   }
 
   return false;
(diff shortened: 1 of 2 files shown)


llvm/lib/Transforms/InstCombine/InstCombineSelect.cpp
```diff
@@ -3678,6 +3678,16 @@ static bool impliesPoisonOrCond(const Value *ValAssumedPoison, const Value *V,
                       *RHSC2);
     }
   }
+  Value *A;
+  if (match(ValAssumedPoison, m_NUWTrunc(m_Value(A))) &&
+      isGuaranteedNotToBePoison(A)) {
+    assert(ValAssumedPoison->getType()->isIntOrIntVectorTy(1));
+    unsigned BitWidth = A->getType()->getScalarSizeInBits();
+    return ConstantRange(APInt::getZero(BitWidth), APInt(BitWidth, 2))
+        .contains(computeConstantRangeIncludingKnownBits(
```
nikic: Can you use just computeKnownBits here? I don't think computeConstantRange() adds any precision here, as it's a power of 2 range.

andjo403: no changes in local run of llvm-opt-benchmark with the change so fix pushed

llvm/lib/Transforms/InstCombine/InstCombineSelect.cpp
```diff
@@ -3678,6 +3678,14 @@ static bool impliesPoisonOrCond(const Value *ValAssumedPoison, const Value *V,
                       *RHSC2);
     }
   }
+  Value *A;
+  if (match(ValAssumedPoison, m_NUWTrunc(m_Value(A))) &&
+      isGuaranteedNotToBePoison(A)) {
+    assert(ValAssumedPoison->getType()->isIntOrIntVectorTy(1));
+    return computeKnownBits(
+               A, SQ.getWithInstruction(cast<Instruction>(ValAssumedPoison)))
+               .getMaxValue() == 1;
```
MaxGraey, line 3687: `.getMaxValue() == 1` is too conservative.  I think it should be `.countMaxActiveBits() <= 1` or something like this. Example:

```llvm
define i1 @known_zero_const(i1 %c) {
  %trunc = trunc nuw i8 0 to i1
  %and = select i1 %c, i1 %trunc, i1 false
  ret i1 %and
}
```
`MaxValue` and `countMaxActiveBits` will be zero here.

andjo403: that trunc will be gone when select is folded as it is a constant so getMaxValue is enough

MaxGraey: How about this one?
```llvm
define i1 @test_logical_and_trunc_nuw_range_0_2(i1 %c, i8 noundef range(i8 0, 2) %x) {
  %trunc = trunc nuw i8 %x to i1
  %and = select i1 %c, i1 %trunc, i1 false
  ret i1 %and
}
```
Hmm, `MaxValue` here is one as well...

nikic (approved):
LGTM

PR #198187, opened 2026-05-17
[IR] Drop hash map for ValueHandleBase

dtcxzyw:
I accidentally found that the DenseMap lookup of ValueHandleBase takes up about ~2.7% of the compilation time: https://dtcxzyw.github.io/llvm-opt-benchmark-nightly/

This patch drops the hash map for value handles. Instead, it stores the head of use chain in the Value class. This change saves about 1% compilation time, at the cost of adding one pointer to each Value instance.

Some benchmark measurements:
https://github.com/dtcxzyw/llvm-opt-benchmark-nightly/pull/324
https://llvm-compile-time-tracker.com/compare.php?from=d90baa054dfc7d9f53148e4739effbf9fd7ac27a&to=431a99496947e997049ef465f5a6ba69fccb8552&stat=instructions:u
Memory usage (the worst case is +1.5%): https://llvm-compile-time-tracker.com/compare.php?from=d90baa054dfc7d9f53148e4739effbf9fd7ac27a&to=431a99496947e997049ef465f5a6ba69fccb8552&stat=max-rss

I am not clear about the historical reason for keeping the Value class compacted. But I believe it is a win. As a bonus, this also frees up 1 bit for SubclassOptionalData.

### DIFF
--- a/llvm/include/llvm/IR/Value.h
+++ b/llvm/include/llvm/IR/Value.h
@@ -52,6 +52,7 @@ class raw_ostream;
 template<typename ValueTy> class StringMapEntry;
 class Twine;
 class User;
+class ValueHandleBase;
 
 using ValueName = StringMapEntry<Value *>;
 
@@ -73,8 +74,7 @@ using ValueName = StringMapEntry<Value *>;
 /// objects that watch it and listen to RAUW and Destroy events.  See
 /// llvm/IR/ValueHandle.h for details.
 class Value {
-  const unsigned char SubclassID;   // Subclass identifier (for isa/dyn_cast)
-  unsigned char HasValueHandle : 1; // Has a ValueHandle pointing to this?
+  const unsigned char SubclassID; // Subclass identifier (for isa/dyn_cast)
 
 protected:
   /// Hold subclass data that can be dropped.
@@ -82,7 +82,7 @@ protected:
   /// This member is similar to SubclassData, however it is for holding
   /// information which may be used to aid optimization, but which may be
   /// cleared to zero without affecting conservative interpretation.
-  unsigned char SubclassOptionalData : 7;
+  unsigned char SubclassOptionalData;
 
 private:
   /// Hold arbitrary subclass data.
@@ -117,9 +117,10 @@ protected:
 private:
   Type *VTy;
   Use *UseList = nullptr;
+  ValueHandleBase *ValueHandle = nullptr;
 
   friend class ValueAsMetadata; // Allow access to IsUsedByMD.
-  friend class ValueHandleBase; // Allow access to HasValueHandle.
+  friend class ValueHandleBase; // Allow access to ValueHandle.
 
   template <typename UseT> // UseT == 'Use' or 'const Use'
   class use_iterator_impl {
@@ -562,7 +563,7 @@ public:
   }
 
   /// Return true if there is a value handle associated with this value.
-  bool hasValueHandle() const { return HasValueHandle; }
+  bool hasValueHandle() const { return ValueHandle != nullptr; }
 
   /// Return true if there is metadata referencing this value.
   bool isUsedByMetadata() const { return IsUsedByMD; }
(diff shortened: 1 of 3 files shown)


MaskRay:
I am in favor of change.
Coincidentally I am experimenting with a DenseMap optimization that `erase` may relocate non-erased elements and have noticed that `ValueHandleBase::AddToUseList` needs the addresses of DenseMap buckets, which is unfortunate. https://github.com/llvm/llvm-project/compare/main...MaskRay:llvm-project:perf/densemap-backshift

aengelke:
I've mixed feelings about this.

The current implementation of ValueHandle is bad. The primary reason this patch is beneficial is because ValueHandles are currently (IMO) overused. If we add InstructionListener, we get another tool that can hopefully replace more uses of ValueHandle than just SCEV. The most expensive user of ValueHandle is most probably ValueMap (as used in the inliner, but also elsewhere), often with ValueHandles for keys *and* values.

I'd personally rather see us removing uses of ValueHandle... because if we do this, we'll never shrink Value again. While retired instructions is generally a good proxy metric for performance, Clang/LLVM tend to execute at a rather low IPC already, frequently with a fair amount of cache misses. If we grow *Value*, we will increase cache misses, which obviously has a much smaller impact than the immediate benefits of this, but in the long term, there probably is some cost attached, which we can't quantify now.

> I am not clear about the historical reason for keeping the Value class compacted.

The IR is our main driver for memory consumption, in addition to Machine IR (after regalloc). c-t-t is not a good tool to capture memory usage regressions, max-rss is way too coarse for this. Could you get numbers with e.g. Valgrind's massif? (edit: Massif prints allocation sources, so ideally you can somewhat easily separate the data for the size of the old hash table.)

dtcxzyw:
> Could you get numbers with e.g. Valgrind's massif? (edit: Massif prints allocation sources, so ideally you can somewhat easily separate the data for the size of the old hash table.)

I run opt O3 with the largest file in llvm-opt-benchmark: https://huggingface.co/buckets/llvm-opt-benchmark/llvm-opt-benchmark/tree/velox/original/ArrayFunctionsRegistration.bc.

Before(baseline version f45f3ce5da9e35f819db9b8ed2775dac8c2b1171): https://gist.github.com/dtcxzyw/b4b72c36bd7726514fd23ebd550314e2

Peak mem usage: 668.9 MB
ValueHandles' DenseMap takes about 1.3%-1.9% of memory usage.

After: https://gist.github.com/dtcxzyw/13b41d635d3ddcb3947182fd55151cbd

Peak mem usage: 675.7 MB (+1.0%)

aengelke:
Thanks for collecting the data. It looks like this is an assert build (there're references to AssertingVH in the profiles)? Anyhow, the ValueHandle map size seems to be primarily determined by the inliner. I think it would be better to remove the uses of ValueMap+ValueHandle there. This should give a higher performance improvement as well.

At the peak point 7/6, we have a ~5-6% regression for users with a fixed number of operands:
->13.10% (91,863,768B) 0xDE73B33: allocateFixedOperandUser (User.cpp:155) + ->09.24% (64,781,920B) 0xDE73C1F: allocateFixedOperandUser (User.cpp:155)
->13.78% (97,647,528B) 0xDE73C43: allocateFixedOperandUser (User.cpp:155) + ->09.55% (67,673,728B) 0xDE73D2F: allocateFixedOperandUser (User.cpp:155)

I'd rather not have this increased memory usage for a quick improvement of an (IMO) anti-feature if we can get similar benefits in ways that seem better in the long term.

MaskRay:
> I accidentally found that the DenseMap lookup of ValueHandleBase takes up about ~2.7% of the compilation time: [dtcxzyw.github.io/llvm-opt-benchmark-nightly](https://dtcxzyw.github.io/llvm-opt-benchmark-nightly/)

After recent DenseMap improvements, the percentage should go down, but ValueHandle remains a weird special user of DenseMap.

nikic:
For the record, I did a vibe-coded experiment to drop ValueMap during inlining (https://llvm-compile-time-tracker.com/compare.php?from=7e439d571e3dcf420448391eb76e58879e73ee1e&to=0359559b071dbd73ebfa41c111db71fbfb71d2b0&stat=instructions:u) but haven't found time to implement it properly yet.

For dropping other value handles, a big problem is that we're lacking a good way to validate such changes right now (https://discourse.llvm.org/t/rfc-valuedeletionlistener-context-level-value-deletion-notifications/90624/36?u=nikic).

dtcxzyw:
It is no longer a performance bottleneck: https://dtcxzyw.github.io/llvm-opt-benchmark-nightly/

PR #198189, opened 2026-05-17
[lit] Move maxIndividualTestTime from global to test suite config

vitalybuka:
Simplify LitConfig initialization and setter to allow None values.
TestingConfig.maxIndividualTestTime is initialized to 0 (or resolved to 0
if None) strictly during initialization.

This fixes an issue where the aggressive BOLT timeout of 60s (previously
set globally on lit_config) was leaking and affecting libc++ tests. By
moving the timeout configuration from the global lit_config to the
individual test suite config, we ensure that timeouts are isolated and
respect suite-local settings without leaking.

Assisted-by: Gemini


vitalybuka:
Created by mistake. real one is #198192

PR #198191, opened 2026-05-17
[NFC][lit] Add timeout-config.py test case

vitalybuka:
Add an integration test case under Inputs/timeout-config to verify suite-local timeout behavior. This initial commit uses the pre-change syntax (lit_config.maxIndividualTestTime = 1) to establish a baseline.

PR Stack:
* ➤ https://github.com/llvm/llvm-project/pull/198191
* https://github.com/llvm/llvm-project/pull/198192
* https://github.com/llvm/llvm-project/pull/198193

Assisted-by: Gemini


boomanaiden154 (commented):
It looks like we now have two sets of the same PR stack? Can we close one of them?

vitalybuka:
> It looks like we now have two sets of the same PR stack? Can we close one of them?

Sorry, done!

boomanaiden154 (commented):
We should probably just reuse the `shtest-timeout.py` test for this.

vitalybuka:
When I moved into shtest-timeout.py, test on the line 32 is exactly the same, so abandoning

PR #198192, opened 2026-05-17
[lit] Move maxIndividualTestTime from global to test suite config

vitalybuka:
Simplify LitConfig initialization and setter to allow None values. TestingConfig.maxIndividualTestTime is initialized to 0 (or resolved to 0 if None) strictly during initialization.

This fixes an issue where the aggressive BOLT timeout of 60s (previously set globally on lit_config) was leaking and affecting libc++ tests. By moving the timeout configuration from the global lit_config to the individual test suite config, we ensure that timeouts are isolated and respect suite-local settings without leaking.

PR Stack:
* ➤ https://github.com/llvm/llvm-project/pull/198192
* https://github.com/llvm/llvm-project/pull/198193

Assisted-by: Gemini


JDevlieghere:
LGTM for LLDB

paschalis-mpeis (commented):
Thanks, Bolt part LG.

boomanaiden154 (approved):
LGTM.

llvm/utils/lit/lit/TestingConfig.py
```diff
@@ -247,6 +250,19 @@ def finish(self, litConfig):
...
+            if suite_timeout > 0 and suite_timeout != litConfig.maxIndividualTestTime:
+                litConfig.note(
+                    (
+                        "The test suite {0!r} configuration requested an individual"
+                        " test timeout of {1} seconds but a timeout of {2} seconds was"
+                        " requested on the command line. Forcing timeout to be {2}"
+                        " seconds."
+                    ).format(self.name, suite_timeout, litConfig.maxIndividualTestTime)
+                )
+            self.maxIndividualTestTime = litConfig.maxIndividualTestTime
```
boomanaiden154, line 265: This looks like it needs to be indented one more level?

vitalybuka: It's in correct spot. We print note if condition is satisfied. 
But we override by global one all the time

vitalybuka:
If no objection I'll lend it tomorrow.

PR #198193, opened 2026-05-17
[lit] Enforce read-only behavior on lit_config.maxIndividualTestTime

vitalybuka:
Enforce read-only behavior on lit_config.maxIndividualTestTime after initialization to prevent test suites from modifying it globally, and add a core lit test to verify this behavior.

PR Stack:
* https://github.com/llvm/llvm-project/pull/198192
* https://github.com/llvm/llvm-project/pull/199886
* https://github.com/llvm/llvm-project/pull/199996
* ➤ https://github.com/llvm/llvm-project/pull/198193
Assisted-by: Gemini


llvm/utils/lit/tests/Inputs/lit-config-readonly/dummy.txt
```diff
@@ -0,0 +1 @@
+# dummy test
```
boomanaiden154: Do we actually need this file?

llvm/utils/lit/tests/Inputs/lit-config-readonly/lit.cfg
```diff
@@ -0,0 +1,12 @@
+# -*- Python -*-
```
boomanaiden154: We can drop this line?

llvm/utils/lit/tests/lit-config-readonly.py
```diff
@@ -0,0 +1,4 @@
+# RUN: not %{lit} %{inputs}/lit-config-readonly > %t.out 2> %t.err
```
boomanaiden154: We never use `%t.out`. I think we can just drop the `> %t.out` part.

llvm/utils/lit/tests/lit-config-readonly.py
```diff
@@ -0,0 +1,4 @@
+# RUN: not %{lit} %{inputs}/lit-config-readonly 2> %t.err
+# RUN: FileCheck --check-prefix=CHECK-ERR < %t.err %s
```
boomanaiden154: We don't need to manually specify a check-prefix here given there's only a single FileCheck invocation.

boomanaiden154 (approved):
LGTM, minus one nit.

PR #198194, opened 2026-05-17
[SLP] Prefer outer binary op when opcode groups tie in buildInstructionsState

### LINKED ISSUE #43353: [SLPVectorizer] Failure to keep complex math vectorized
|  |  |
| --- | --- |
| Bugzilla Link | [44008](https://llvm.org/bz44008) |
| Version | trunk |
| OS | Windows NT |
| CC | @alexey-bataev,@anton-afanasyev,@topperc,@dtemirbulatov,@rotateright,@vporpo |

## Extended Description 
https://godbolt.org/z/YgBdFz
```c++
#include <complex>

std::complex<float> mul_f32(std::complex<float> &A) {
    return (2.0f * A);
}

std::complex<float> fma_f32(std::complex<float> &A, float B) {
    return (2.0f * A) + B;
}
```
for the simple mul case we manage to keep the real+imag components vectorized as a <2 x float>:

clang -g0 -O3 -march=btver2
```asm
mul_f32:
  vmovsd (%rdi), %xmm0 # xmm0 = mem[0],zero
  vaddps %xmm0, %xmm0, %xmm0
  retq
```
but with the fma case, we end up scalarizing:
```asm
fma_f32:
  vmovsd (%rdi), %xmm1 # xmm1 = mem[0],zero
  vmovshdup %xmm1, %xmm2 # xmm2 = xmm1[1,1,3,3]
  vaddss %xmm1, %xmm1, %xmm1
  vaddss %xmm2, %xmm2, %xmm2
  vaddss %xmm0, %xmm1, %xmm0
  vinsertps $16, %xmm2, %xmm0, %xmm0 # xmm0 = xmm0[0],xmm2[0],xmm0[2,3]
  retq
```
Ideally this would be:
```asm
fma_f32:
  vmovsd (%rdi), %xmm1 # xmm1 = mem[0],zero
  vaddps %xmm1, %xmm1, %xmm1
  vaddss %xmm0, %xmm1, %xmm1 # xmm1[0] += xmm0[0]
  vmovaps %xmm1, %xmm0
  retq
```

alexey-bataev:
When VL contains instructions of different opcodes with equal counts,
the tie-breaking in buildInstructionsState could replace an outer
operation (e.g., fadd) with an inner one (e.g., fmul) that appears as
its direct operand, depending on SmallMapVector iteration order. Add a
check: if the current MainOp is a BinaryOperator with a direct operand
matching the challenger partition's opcode in the same block, keep
MainOp instead of switching to the inner operation.

Partially fixes #43353


RKSimon (approved):
LGTM

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -11739,6 +11739,23 @@ class InstructionsCompatibilityAnalysis {
             return Operands.contains(I);
           }))
         continue;
+      // On a tie, keep the outer binary op as MainOp rather than replacing it
+      // with an inner op that appears as its direct operand. For example, in
+      // (2.0f * A) + B the fadd and fmul each appear once in VL; without this
+      // check the fmul could win and prevent vectorization of the fadd pair.
```
bababuck, line 11613: this example isn't clear to me. What would the other lane contain in this case?

alexey-bataev: In this example we have some thing like (fadd(fmul), fmul) pair. If the main op is chosen to be second fmul, we'll end up with the alternate fmul/fadd vectorization, which is not profitable. With this fix, instead we'll have fadd <2x> (fmul <2x>)

bababuck (approved):
LGTM

PR #198198, opened 2026-05-17
[mlir][emitc] Support struct and union definition

aniragil:
This patch extends the `emitc.class` op to define structs and unions in addition to classes using a new enum attribute with the values:
- class_  (class,  value 0) -- default
- struct_ (struct, value 1)
- union_  (union,  value 2)
    
Class op now takes an optional class_type, defaulting to `class`. The attribute is printed to the assembly only when it differs from the default, retaining current semantics, e.g.:

      emitc.class @Foo { ... }          // class (default, omitted)
      emitc.class struct @Foo { ... }   // struct
      emitc.class union @Foo { ... }    // union
      emitc.class struct final @Foo {}  // struct + final

The translator emits the class_type instead of the previously hard-coded `class`, and for `class` types also emits `public:` inside the body, retaining the current 'all-public classes' behavior of `emitc.class`.

Assisted-by: Copilot

### DIFF
--- a/mlir/lib/Target/Cpp/TranslateToCpp.cpp
+++ b/mlir/lib/Target/Cpp/TranslateToCpp.cpp
@@ -1225,11 +1225,15 @@ static LogicalResult printOperation(CppEmitter &emitter, ModuleOp moduleOp) {
 
 static LogicalResult printOperation(CppEmitter &emitter, ClassOp classOp) {
   raw_indented_ostream &os = emitter.ostream();
-  os << "struct " << classOp.getSymName();
+  ClassType classType = classOp.getClassType();
+  os << stringifyClassType(classType) << " " << classOp.getSymName();
   if (classOp.getFinalSpecifier())
     os << " final";
   os << " {\n";
   os.indent();
+
+  if (classType == ClassType::class_)
+    os << "public:\n";
 
   for (Operation &op : classOp) {
     if (failed(emitter.emitOperation(op, /*trailingSemicolon=*/false)))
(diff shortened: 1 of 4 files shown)


aniragil:
This PR is a step towards extending emitc.class for C struct support as proposed in the [EmitC EuroLLVM 2026 round table](https://discourse.llvm.org/t/eurollvm-2026-round-table-summary-emitc-brainstorming/90618).

mlir/lib/Target/Cpp/TranslateToCpp.cpp
```diff
@@ -1225,10 +1225,10 @@ static LogicalResult printOperation(CppEmitter &emitter, ModuleOp moduleOp) {
 
 static LogicalResult printOperation(CppEmitter &emitter, ClassOp classOp) {
   raw_indented_ostream &os = emitter.ostream();
-  os << "class " << classOp.getSymName();
+  os << "struct " << classOp.getSymName();
```
mtrofin: can this be an opt-in? I think the goal is to eventually add accessors in the `class` case, the "everything is public" is a point in time towards an initial, minimally-viable release. At the same time, also supporting `struct` also makes sense.

marbre: Yeah, this rather seems like a breaking change and opt-in would be nice.

aniragil: Agreed, this change is slightly breaking:

- changes default inheritance to `public` instead of `private`, would require explicitly inheriting as `private` if anyone downstream was counting on that.
- may require changing lit tests downstream

Re-implementing this by adding to `emitc.class` a new `class_type` property, defaulting to `class`, that the translator will use to determine what keyword to use of the possible `class`, `struct` and `union`. For `class`, the translator will emit the `public` section to retain current behavior.

mtrofin (approved):
lgtm, but I think the commit message and description may need an update.

aniragil:
> lgtm, but I think the commit message and description may need an update.

Indeed, will do. Thanks @mtrofin !

PR #198202, opened 2026-05-17
[CodeGen] Mark operator delete as captures(address)

nikic:
The pointer passed to operator delete cannot be used after the call, which means that its provenance is not captured. As such, we can mark the pointer as `captures(address)`. We allow address capture, because operator delete is replaceable, and a replacement allocator could observe the address.

I'm applying the attribute independently of `-fassume-sane-operator-new`, as I don't believe this makes any assumptions on the underlying allocator implementation. (It might make sense to upgrade to `captures(none)` under `-fassume-sane-operator-new`, but I'm not sure there is much benefit to that.)

This makes sure that pointers produced by operator new and only passed to operator delete are considered "never escaped".

AI disclosure: Claude assisted with test updates.

### DIFF
--- a/clang/lib/CodeGen/CGCall.cpp
+++ b/clang/lib/CodeGen/CGCall.cpp
@@ -3207,17 +3207,15 @@ void CodeGenModule::ConstructAttributeList(StringRef Name,
   }
 
   if (const FunctionDecl *Fn = dyn_cast_or_null<FunctionDecl>(TargetDecl)) {
-    if (Fn->isReplaceableGlobalAllocationFunction()) {
-      auto Kind = Fn->getDeclName().getCXXOverloadedOperator();
-      if (Kind == OO_Delete || Kind == OO_Array_Delete) {
-        auto [IRArg, NumIRArgs] = IRFunctionArgs.getIRArgs(0);
-        assert(NumIRArgs == 1 && "Pointer should be a single argument");
-        (void)NumIRArgs;
-        // Add captures(address) to deleted pointer. The provenance is not
-        // captured, because delete destroy the provenane of the pointer. The
-        // address may be captured, as the allocator is replaceable.
-        ArgAttrs[IRArg].addCapturesAttr(llvm::CaptureComponents::Address);
-      }
+    if (Fn->isReplaceableGlobalAllocationFunction() &&
+        Fn->getDeclName().isAnyOperatorDelete()) {
+      auto [IRArg, NumIRArgs] = IRFunctionArgs.getIRArgs(0);
+      assert(NumIRArgs == 1 && "Pointer should be a single argument");
+      (void)NumIRArgs;
+      // Add captures(address) to deleted pointer. The provenance is not
+      // captured, because delete destroy the provenance of the pointer. The
+      // address may be captured, as the allocator is replaceable.
+      ArgAttrs[IRArg].addCapturesAttr(llvm::CaptureComponents::Address);
     }
   }
 
--- a/clang/lib/CodeGen/CGCall.cpp
+++ b/clang/lib/CodeGen/CGCall.cpp
@@ -3207,7 +3207,8 @@ void CodeGenModule::ConstructAttributeList(StringRef Name,
   }
 
   if (const FunctionDecl *Fn = dyn_cast_or_null<FunctionDecl>(TargetDecl)) {
-    if (Fn->isReplaceableGlobalAllocationFunction() &&
+    if (getCodeGenOpts().AssumeSaneOperatorNew &&
+        Fn->isReplaceableGlobalAllocationFunction() &&
         Fn->getDeclName().isAnyOperatorDelete()) {
       auto [IRArg, NumIRArgs] = IRFunctionArgs.getIRArgs(0);
       assert(NumIRArgs == 1 && "Pointer should be a single argument");
--- a/clang/lib/CodeGen/CGCall.cpp
+++ b/clang/lib/CodeGen/CGCall.cpp
@@ -3206,6 +3206,21 @@ void CodeGenModule::ConstructAttributeList(StringRef Name,
     }
   }
 
+  if (const FunctionDecl *Fn = dyn_cast_or_null<FunctionDecl>(TargetDecl)) {
+    if (Fn->isReplaceableGlobalAllocationFunction()) {
+      auto Kind = Fn->getDeclName().getCXXOverloadedOperator();
+      if (Kind == OO_Delete || Kind == OO_Array_Delete) {
+        auto [IRArg, NumIRArgs] = IRFunctionArgs.getIRArgs(0);
+        assert(NumIRArgs == 1 && "Pointer should be a single argument");
+        (void)NumIRArgs;
+        // Add captures(address) to deleted pointer. The provenance is not
+        // captured, because delete destroy the provenane of the pointer. The
+        // address may be captured, as the allocator is replaceable.
+        ArgAttrs[IRArg].addCapturesAttr(llvm::CaptureComponents::Address);
+      }
+    }
+  }
+
   SmallVector<llvm::AttributeSet, 4> ArgAttrSets;
   for (const llvm::AttrBuilder &Attrs : ArgAttrs)
     ArgAttrSets.push_back(llvm::AttributeSet::get(getLLVMContext(), Attrs));
--- a/clang/test/CodeGen/pr53127.cpp
+++ b/clang/test/CodeGen/pr53127.cpp
@@ -8,8 +8,8 @@ void operator delete(void*);
 // CHECK-LABEL: @_Z1fPiz(
 // CHECK-NEXT:  entry:
 // CHECK-NEXT:    [[P_ADDR:%.*]] = alloca ptr, align 8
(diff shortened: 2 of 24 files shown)


efriedma-quic:
It's hard for me to analyze this because we don't really have a model for how new/delete/malloc/free create and destroy provenance, in the case where the implementation is visible to the compiler.

Say you have an implementation of new which uses malloc to get an hunk of memory from the OS, divides it up, and returns the pieces.  operator delete takes the pointer, and sticks the pointer in a free list.

Presumably operator new creates provenance for the pieces, and operator delete destroys it.  But what about the provenance of the underlying malloc'ed memory?  You could say that the pointer received by operator delete has no provenance, and if we want to perform any access, we need to recreate the malloc's provenance.  But that's not the only possible interpretation.

I'm a little worried about adding provenance annotations when we haven't nailed this down yet.  Especially with -fno-assume-sane-operator-new, since that's the only escape hatch we have at the moment.

clang/lib/CodeGen/CGCall.cpp
```diff
@@ -3205,6 +3205,21 @@ void CodeGenModule::ConstructAttributeList(StringRef Name,
...
+    if (Fn->isReplaceableGlobalAllocationFunction()) {
+      auto Kind = Fn->getDeclName().getCXXOverloadedOperator();
+      if (Kind == OO_Delete || Kind == OO_Array_Delete) {
+        auto [IRArg, NumIRArgs] = IRFunctionArgs.getIRArgs(0);
+        assert(NumIRArgs == 1 && "Pointer should be a single argument");
+        (void)NumIRArgs;
+        // Add captures(address) to deleted pointer. The provenance is not
+        // captured, because delete destroy the provenane of the pointer. The
+        // address may be captured, as the allocator is replaceable.
+        ArgAttrs[IRArg].addCapturesAttr(llvm::CaptureComponents::Address);
```
ojhunt: Question not related to this PR - is there a source level attribute that would permit this to be specified as capturing the address? (think custom allocators, etc)

ojhunt (approved):
LGTM, I spent a moment looking at the type aware allocators  but the replaceable global allocator check should cover that. If you think there's any chance that this might ever expand to cover other operators delete it might be worth adding an `assert(!isTypeAwareNewOrDelete())`

nikic:
The way I would model allocator provenance goes something like this:

`%p = ptr @llvm.provenance.alloc(ptr %p_orig, i64 %size)` returns a pointer `%p` with fresh provenance for `%size` bytes starting at `%p_orig`. At the same time, access permissions for this region are removed from `%p_orig` (including from any "related" pointers).

`%p_orig_new = ptr @llvm.provenance.dealloc(ptr %p)` removes access permissions of `%p`, restores access permissions on `%p_orig` and returns a pointer with the same provenance as `%p_orig`.

That is, allocators form a provenance tree where an allocation adds a new leaf, and deallocation recovers the provenance of the parent. This ensures that free lists work correctly.

To handle asymmetrically inlined allocators fully correctly, we probably actually need to materialize intrinsics like these for the operator new return / operator delete argument.

> I'm a little worried about adding provenance annotations when we haven't nailed this down yet. Especially with -fno-assume-sane-operator-new, since that's the only escape hatch we have at the moment.

Thinking about this again, making this part of `-fassume-sane-operator-new` probably makes sense, because the `captures(address)` on `operator delete` is kind of the dual of `noalias` on `operator new`, and we only add the latter for sane operator new.

antoniofrighetto:
> > I'm a little worried about adding provenance annotations when we haven't nailed this down yet. Especially with -fno-assume-sane-operator-new, since that's the only escape hatch we have at the moment.
> 
> Thinking about this again, making this part of `-fassume-sane-operator-new` probably makes sense, because the `captures(address)` on `operator delete` is kind of the dual of `noalias` on `operator new`, and we only add the latter for sane operator new.

I understand the concerns raised here, and that, in principle, if we don't assume sane allocator semantics, it may as well be questionable to assert that pointer provenance does not outlive `operator delete`; but I can't immediately see how the unusual semantics for `operator new` are also implied for `operator delete`. `-fassume-sane-operator-new` seems to make a statement about operator new only, not about delete (unlike GCC `-fassume-sane-operators-new-delete` which covers the latter too), which also makes me slightly cast doubts on the duality between captures(address) and noalias itself.

clang/lib/CodeGen/CGCall.cpp
```diff
@@ -3205,6 +3205,21 @@ void CodeGenModule::ConstructAttributeList(StringRef Name,
...
 
+  if (const FunctionDecl *Fn = dyn_cast_or_null<FunctionDecl>(TargetDecl)) {
+    if (Fn->isReplaceableGlobalAllocationFunction()) {
+      auto Kind = Fn->getDeclName().getCXXOverloadedOperator();
+      if (Kind == OO_Delete || Kind == OO_Array_Delete) {
+        auto [IRArg, NumIRArgs] = IRFunctionArgs.getIRArgs(0);
+        assert(NumIRArgs == 1 && "Pointer should be a single argument");
+        (void)NumIRArgs;
+        // Add captures(address) to deleted pointer. The provenance is not
+        // captured, because delete destroy the provenane of the pointer. The
```
antoniofrighetto:
```suggestion
        // captured, because delete destroy the provenance of the pointer. The
```

efriedma-quic:
> `%p_orig_new = ptr @llvm.provenance.dealloc(ptr %p)` removes access permissions of `%p`, restores access permissions on `%p_orig` and returns a pointer with the same provenance as `%p_orig`.

And the argument of the intrinsic is `captures(address)`?  That directly follows from the operand to `operator delete` itself being `captures(address)`.  I guess that's a consistent model, but maybe a little mind-bending for anyone who hasn't thought about it...

> To handle asymmetrically inlined allocators fully correctly, we probably actually need to materialize intrinsics like these for the operator new return / operator delete argument.

Yes, this makes sense.

> `-fassume-sane-operator-new` seems to make a statement about operator new only,

`-fassume-sane-operator-new` marks the new operator noalias, which is basically saying it creates provenance.  This is the exact counterpart: delete destroys that provenance.  Under the model where new doesn't create provenance, delete also doesn't destroy provenance, so it has to capture it.

antoniofrighetto:
> > `-fassume-sane-operator-new` seems to make a statement about operator new only,
> 
> `-fassume-sane-operator-new` marks the new operator noalias, which is basically saying it creates provenance. This is the exact counterpart: delete destroys that provenance.

I see that now, thanks for the clarification.

> `%p = ptr @llvm.provenance.alloc(ptr %p_orig, i64 %size)` returns a pointer `%p` with fresh provenance for `%size` bytes starting at `%p_orig`. At the same time, access permissions for this region are removed from `%p_orig` (including from any "related" pointers).
> 
> `%p_orig_new = ptr @llvm.provenance.dealloc(ptr %p)` removes access permissions of `%p`, restores access permissions on `%p_orig` and returns a pointer with the same provenance as `%p_orig`.

Partially related, one thing I had been previously wondering about, assuming realloc may be part of the model, was whether a `resize` intrinsic for in-place allocations could be envisioned (which would transfer the provenance to the shrunk pointer), but I guess a realloc is conceptually pretty much always a dealloc + alloc.

> To handle asymmetrically inlined allocators fully correctly, we probably actually need to materialize intrinsics like these for the operator new return / operator delete argument.

Some allocators happen to store metadata inside the freed chunk, which, IIRC, requires dereferencing the freed block. Say allocator internals get inlined as part of `operator delete`, I assume such materialized intrinsics would aspire to act, to some extent, as lifetime start / end of the allocation / deallocation, except that they would say which pointer operations are allowed on which memory (rather than saying which memory is live or dead).

efriedma-quic:
> I guess a realloc is conceptually pretty much always a dealloc + alloc.

In-place realloc needs to preserve the contents of the memory, which means it probably needs a dedicated intrinsic. llvm.provenance.dealloc would make the memory undefined.  (Realloc which allocates new memory doesn't need anything special.)

> lifetime start / end of the allocation / deallocation

Basically, yes.

nikic:
RFC: https://discourse.llvm.org/t/rfc-allocator-provenance-model/91106

Though seeing the last comments on this thread, one thing I forgot to account for is the realloc case. We could make `dealloc` not poison the memory, so that realloc can be a `dealloc` + `alloc` combination. It would require a change to `allockind` to specify whether poisoning happens, as DSE assumes it does.

Alternatively, could have a dedicated `realloc` intrinsic.

antoniofrighetto:
> RFC: https://discourse.llvm.org/t/rfc-allocator-provenance-model/91106

@nikic I was wondering whether you were planning to drive this RFC yourself, or whether you'd be open to supervising/reviewing the work (if any, at least for the IR and Clang parts). Apologies if this might sound obvious from the thread.

nikic:
> > RFC: https://discourse.llvm.org/t/rfc-allocator-provenance-model/91106
> 
> @nikic I was wondering whether you were planning to drive this RFC yourself, or whether you'd be open to supervising/reviewing the work (if any, at least for the IR and Clang parts). Apologies if this might sound obvious from the thread.

Happy to leave the implementation to you :)

PR #198205, opened 2026-05-17
[Runtimes] Drop 'flang' from runtimes configure dependency

jhuber6:
Summary:
This PR cuts `flang` out of the core runtimes configure dependency. We
will need the runtimes infrastructure to handle `flang` module
generation, but this dependency poisons the entire dependency stack.
`flang` and `mlir` are by far the heavily parts of the LLVM stack and
for distribution we want to only build it when absolutely necessary,
which as of now is only to install flang module files.

The approach here is to simply remove the `flang` target from the core
dependency tree, but intead configure it for the top-level `module`
targets which are part of `all`. To make this work we need to pass
COMPILER_WORKS and set up a dummy location so that configuration passes.
it's a little backdoor, but this is an extremely important quality of
life improvement for LLVM distribution support.


llvm/runtimes/CMakeLists.txt
```diff
@@ -91,6 +91,27 @@ macro(set_enable_per_target_runtime_dir)
...
+macro(add_flang_mod_deps build_target extra_targets_list)
+  if(TARGET flang)
+    set(_fortran_deps flang)
+    if(TARGET module_files)
+      list(APPEND _fortran_deps module_files)
+    endif()
+    if(TARGET ${build_target})
+      add_dependencies(${build_target} ${_fortran_deps})
+    endif()
+    foreach(tgt ${${extra_targets_list}})
```
Meinersbur:
```suggestion
    foreach(tgt IN LISTS ${extra_targets_list})
```

Meinersbur (commented):
With `ninja install`, where users expect a full flang toolchain to be installed, what triggers building and installation of the module files?

Here it seems only necessary for openmp, but what with #171515 it would also be for flang-rt intrinsics.

jhuber6:
> With `ninja install`, where users expect a full flang toolchain to be installed, what triggers building and installation of the module files?

It's a dependency on `all`, so if someone just does `ninja install` it should trigger.
 
> Here it seems only necessary for openmp, but what with #171515 it would also be for flang-rt intrinsics.

I tried to be forward looking with that and make this logic work for any dependency with `-mod` in its name, it should work AFAIK but I could be missing something.

llvm/CMakeLists.txt
```diff
@@ -1465,6 +1465,14 @@ if( LLVM_INCLUDE_TESTS )
   endif()
   set_target_properties(test-depends PROPERTIES FOLDER "LLVM/Tests")
   add_dependencies(check-all test-depends)
+
+  # Projects that need runtimes for their tests (e.g. flang needs flang-rt).
+  get_property(runtimes_test_consumers GLOBAL PROPERTY LLVM_TEST_DEPENDS_ON_RUNTIMES)
```
Meinersbur: `LLVM_TEST_DEPENDS_ON_RUNTIMES` comes from #198793 and has been replaced with `flang-rt-test-deps` there

llvm/CMakeLists.txt
```diff
@@ -1465,6 +1465,14 @@ if( LLVM_INCLUDE_TESTS )
...
+  foreach(target ${runtimes_test_consumers})
```
Meinersbur:
```suggestion
  foreach(target IN LISTS runtimes_test_consumers)
```

llvm/CMakeLists.txt
```diff
@@ -1465,6 +1465,14 @@ if( LLVM_INCLUDE_TESTS )
...
+    if (TARGET ${target})
```
Meinersbur:
```suggestion
    if(TARGET ${target})
```
[nit]

runtimes/cmake/config-Fortran.cmake
```diff
@@ -88,15 +88,17 @@ if (CMAKE_Fortran_COMPILER)
   # cannot use CMAKE_Fortran_COMPILER_ID.
   cmake_path(GET CMAKE_Fortran_COMPILER STEM _Fortran_COMPILER_STEM)
   if (_Fortran_COMPILER_STEM STREQUAL "flang-new" OR _Fortran_COMPILER_STEM STREQUAL "flang")
+    # Force the compiler ID so CMake does not try to run the compiler for
+    # identification. In a bootstrapping build the Flang binary may not be
+    # built yet at configure time (only CMAKE_Fortran_COMPILER_WORKS is set).
+    set(CMAKE_Fortran_COMPILER_ID "LLVMFlang")
+    set(CMAKE_Fortran_COMPILER_ID_RUN TRUE)
+    set(CMAKE_Fortran_COMPILER_FORCED TRUE)
+    set(CMAKE_Fortran_COMPILER_VERSION "${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}")
```
Meinersbur, line 97: Under Windows, I got
```
CMake Error at C:/Program Files/CMake/share/cmake-4.3/Modules/Platform/Windows-LLVMFlang-Fortran.cmake:64 (message):
  LLVMFlang target ABI unrecognized:
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-4.3/Modules/CMakeFortranInformation.cmake:25 (include)
  cmake/config-Fortran.cmake:161 (enable_language)
  CMakeLists.txt:284 (include)
```
This happens with `CMAKE_Fortran_SIMULATE_ID` being not defined. Might be needed to be set to "GNU"

runtimes/cmake/config-Fortran.cmake
```diff
@@ -147,6 +149,11 @@ else ()
   return ()
 endif ()
 
+# In a bootstrapping build the Fortran compiler may not have been built yet.
+# Create a placeholder so CMake's enable_language() existence check passes.
+if (CMAKE_Fortran_COMPILER_FORCED AND NOT EXISTS "${CMAKE_Fortran_COMPILER}")
+  file(TOUCH "${CMAKE_Fortran_COMPILER}")
```
Meinersbur: When manually deleting `bin/flang`, I sometimes get
```
FAILED: flang-rt/lib/runtime/CMakeFiles/flang_rt.mod.intrinsics.dir/__fortran_builtins.f90.o /home/meinersbur/src/llvm/flangrt/cache/lib/clang/23/finclude/flang/x86_64-unknown-linux-gnu/__fortran_builtins.mod
/home/meinersbur/src/llvm/flangrt/cache/./bin/flang --target=x86_64-unknown-linux-gnu -I/home/meinersbur/src/llvm/flangrt/_src/flang-rt/lib/runtime -I/home/meinersbur/src/llvm/flangrt/_src/flang-rt/include -I/home/meinersbur/src/llvm/flangrt/_src/flang-rt/../flang/include -I/home/meinersbur/src/llvm/flangrt/cache/runtimes/runtimes-bins/flang-rt -I/home/meinersbur/src/llvm/flangrt/_src/cmake/Modules/../../libc -O3 -module-dir/home/meinersbur/src/llvm/flangrt/cache/lib/clang/23/finclude/flang/x86_64-unknown-linux-gnu -fPIC -cpp -mmlir -ignore-missing-type-desc -fno-lto -fintrinsic-modules-path=/home/meinersbur/src/llvm/flangrt/cache/lib/clang/23/finclude/flang/x86_64-unknown-linux-gnu -Xflang -fno-reformat -ffixed-line-length-72 -c flang-rt/lib/runtime/CMakeFiles/flang_rt.mod.intrinsics.dir/__fortran_builtins.f90-pp.f90 -o flang-rt/lib/runtime/CMakeFiles/flang_rt.mod.intrinsics.dir/__fortran_builtins.f90.o
/bin/sh: 1: /home/meinersbur/src/llvm/flangrt/cache/./bin/flang: Permission denied
```
I could not trigger it reliably yet, but maybe could be avoided by setting the mtime of the file to something early.

flang-rt/lib/runtime/CMakeLists.txt
```diff
@@ -305,6 +305,26 @@ if (FLANG_RT_FORTRAN_MODULES)
...
   flang_module_target(flang_rt.mod PUBLIC)
   add_module_barrier(flang-rt-mod flang_rt.mod)
+
+  if (RUNTIMES_FORTRAN_MODULES)
+    set(_mod_install_dest "${RUNTIMES_INSTALL_RESOURCE_MOD_PATH}/..")
+    cmake_path(NORMAL_PATH _mod_install_dest)
+    install(DIRECTORY "${RUNTIMES_OUTPUT_RESOURCE_MOD_DIR}"
+      DESTINATION "${_mod_install_dest}"
+      COMPONENT flang-rt-mod
+    )
```
Meinersbur, line 316: This is already done by https://github.com/llvm/llvm-project/blob/f3dcf16115f862f6fa5ea8e520b371f7e57a9c3d/runtimes/cmake/config-Fortran.cmake#L203-L205

Note that libomp-mod and flang-rt-mod files are both installed by this command.

openmp/module/CMakeLists.txt
```diff
@@ -32,7 +32,16 @@ add_dependencies(libomp-mod ${RUNTIMES_FORTRAN_BUILD_DEPS})
 install(FILES
   "${CMAKE_CURRENT_BINARY_DIR}/../runtime/src/omp_lib.h"
   DESTINATION ${LIBOMP_HEADERS_INSTALL_PATH}
-    COMPONENT openmp
+  COMPONENT libomp-mod
```
Meinersbur, line 41: [not a change request] `omp_lib.h` might not be a module, but bundling it together with modules may still make sense because it is used the same way

Meinersbur (commented):
Using the [Offload.cmake](https://github.com/llvm/llvm-project/blob/main/offload/cmake/caches/Offload.cmake) configuration (+flang).

With `check-flang-rt` and `check-openmp`, it does not build flang such that its tests are failing.

runtimes/cmake/config-Fortran.cmake
jhuber6: I tried making sure that the file is definitely created by making the directory and actually writing something to it instead of just touching. But I'm not sure what could be the real issue.

flang-rt/lib/runtime/CMakeLists.txt
jhuber6: Okay, we want separate `install-` targets because I consider these to be consistent with the other 'component' artifacts. I made it excluded from `all` so they won't conflict and these will only be targeted if done specifically. Do you have an idea to better consolidate if that's still too much additional handling?

runtimes/cmake/config-Fortran.cmake
jhuber6: I just set it to what the C compiler's simulated ID is, seems like a reasonable default.

Meinersbur: Under Windows, CMake would use clang-cl, which is CMAKE_C_SIMULATE_ID "MSVC". There is no flang-cl that would support that.

runtimes/cmake/config-Fortran.cmake
```diff
@@ -88,15 +88,20 @@ if (CMAKE_Fortran_COMPILER)
   # cannot use CMAKE_Fortran_COMPILER_ID.
   cmake_path(GET CMAKE_Fortran_COMPILER STEM _Fortran_COMPILER_STEM)
   if (_Fortran_COMPILER_STEM STREQUAL "flang-new" OR _Fortran_COMPILER_STEM STREQUAL "flang")
+    # Force the compiler ID so CMake does not try to run the compiler for
+    # identification. In a bootstrapping build the Flang binary may not be
+    # built yet at configure time (only CMAKE_Fortran_COMPILER_WORKS is set).
```
Meinersbur, line 93: Did you consider doing this only in a bootstrapping-build (`LLVM_TREE_AVAILABLE`)? Did you try out setting these externally, i.e. in [ `llvm/runtimes/CMakeLists.txt`](https://github.com/llvm/llvm-project/blob/main/llvm/runtimes/CMakeLists.txt) ?

In other words, is this the intended use case only bootstrapping builds or should users also be able to not need to build flang when they configure a runtimes-standalone build against some other LLVM_BINARY_DIR?

llvm/runtimes/CMakeLists.txt
```diff
@@ -115,6 +115,27 @@ macro(append_passthrough_options out type names)
   endforeach()
 endmacro()
 
+# Wire build-order dependencies on flang for targets that invoke it. This allows
+# us to build the runtimes separately without first building flang.
+macro(add_flang_mod_deps build_target extra_targets_list)
```
Meinersbur:
```suggestion
function(add_flang_mod_deps build_target extra_targets_list)
```

llvm/runtimes/CMakeLists.txt
```diff
@@ -115,6 +115,27 @@ macro(append_passthrough_options out type names)
...
+  if(TARGET flang)
+    set(_fortran_deps flang)
+    if(TARGET module_files)
```
Meinersbur: `module_files` does not exist anymore

llvm/runtimes/CMakeLists.txt
```diff
@@ -115,6 +115,27 @@ macro(append_passthrough_options out type names)
...
+      list(APPEND _fortran_deps module_files)
+    endif()
+    if(TARGET ${build_target})
+      add_dependencies(${build_target} ${_fortran_deps})
+    endif()
+    foreach(tgt IN LISTS ${extra_targets_list})
+      if(tgt MATCHES "-mod($|-)")
```
Meinersbur, line 128: Could you add a comment that this is meant for install-flang-rt-mod(-stripped) and install-libomp-mod(-stripped) that install modules and therefore obviosly need flang?

Also, is ${build_target} target meant to be the "all" target that builds everything including the modules?

runtimes/cmake/config-Fortran.cmake
Meinersbur: "Permission denied" because the (empty) file does not have the permission bit set, not because it does not exist

In case of ninja, if the mtime of the empty file happens to land in the `.ninja_log`, it thinks it's the most recent artfiact and will not build `bin/flang`.

runtimes/cmake/config-Fortran.cmake
jhuber6: My intention was that this only applies to the  bootstrapping builds, if someone is doing standalone builds they would likely be able to manage their own build stages.

llvm/runtimes/CMakeLists.txt
jhuber6: Tried to explain it a little better.

runtimes/cmake/config-Fortran.cmake
jhuber6: I added execution priviledges... But it will probably just fail because it's a text file. I am not actually sure if this is a proper way to fix it, but I just spoofed the build time to be some other epoch so it hopefully lists as stale.

flang-rt/lib/runtime/CMakeLists.txt
Meinersbur: It is unfortunately not possible to separate libomp-mod and flang-rt-mod. They both are emitted into ${RUNTIMES_OUTPUT_RESOURCE_MOD_DIR}, and they have to be there for `bin/flang` when invoked from the build directory to find them. At the same time, CMake does not allow to identify which .f90 file produces which .mod.

However, even if both install commands are executed, don't think it would cause any problem.

Theoretically, one could emit the modules file for libomp-mod and flang-rt-mod into separate directory, and as a post-build action to copy them into ${RUNTIMES_OUTPUT_RESOURCE_MOD_DIR}. Doesn't seem to be worth the effort though.

runtimes/cmake/config-Fortran.cmake
```diff
@@ -147,6 +150,25 @@ else ()
...
+  file(CHMOD "${CMAKE_Fortran_COMPILER}"
+    PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                GROUP_READ GROUP_EXECUTE
+                WORLD_READ WORLD_EXECUTE)
+  # Ninja uses file mtimes to decide whether build outputs are up-to-date.
+  # If this placeholder's mtime is recent it may match what is recorded in
+  # .ninja_log, causing ninja to skip building the real compiler binary.
+  # Set it so that any subsequent real build always has a newer mtime.
+  execute_process(COMMAND touch -t 197001020000 "${CMAKE_Fortran_COMPILER}"
+                  ERROR_QUIET)
```
Meinersbur, line 166: `touch` doesn't necessarily exist under Windows. But I think it's best-effort only anyway.

runtimes/cmake/config-Fortran.cmake
```diff
@@ -147,6 +150,25 @@ else ()
...
+# The build-order dependency in add_flang_mod_deps ensures the real binary is
+# built before anything tries to invoke this placeholder.
+if (CMAKE_Fortran_COMPILER_FORCED AND NOT EXISTS "${CMAKE_Fortran_COMPILER}")
+  get_filename_component(_compiler_dir "${CMAKE_Fortran_COMPILER}" DIRECTORY)
+  file(MAKE_DIRECTORY "${_compiler_dir}")
+  file(WRITE "${CMAKE_Fortran_COMPILER}" "stub")
+  file(CHMOD "${CMAKE_Fortran_COMPILER}"
+    PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                GROUP_READ GROUP_EXECUTE
+                WORLD_READ WORLD_EXECUTE)
```
Meinersbur: My remark about "permission denied" was only to explain the error message even though the file exists. I didn't mean to imply to set exceutable perssions, it will just get a different error; but that for some reason the empty stup was not overwritten with the real `flang`, and to fix that reason. I think `touch` should do that. I suggest to remove the `CHMOD`.

Meinersbur (approved):
LGTM

Seems to work find with everything I tried. If the goal is to work with bootstrapping only, I may follow-up with changes that might make it more robust by moving logic into llvm/runtimes/CMakeLists.txt so for better control when we can use the stub and when the real flang is necessary.

PR #198206, opened 2026-05-17
[Offload] Add a proper `install-offload` target

jhuber6:
Summary:
This target is missing, it's the expected behavior for all LLVM
components to have a way to install them individually as components.
This allows us to selectively install offload versus other pieces of the
runtime. Also make this and the openmp handling canonical with libcxx.


jdoerfert (approved):
Looks reasonable.

PR #198207, opened 2026-05-17
bug fix c++26 fail for function parameter packs

albertbolt1:
C++26 regression during expanding parameter packs, 
we reach the requires and need to know the length of the pack but this operation fails because the pack cannot be found in LIS


clang/lib/Sema/SemaTemplateVariadic.cpp
```diff
@@ -911,6 +911,34 @@ bool Sema::CheckParameterPacksForExpansion(
...
+      // then it fails because localdecls are not added to the instantiation scope until 
+      // after we check for pack expansions. So we need to check for parameter packs here 
+      // and add them to the instantiation scope before we call findInstantiationOf.
+      if (CurrentInstantiationScope) {
+        if (auto *PVD = dyn_cast<ParmVarDecl>(ND)) {
+          if (PVD->isParameterPack() &&
+              !CurrentInstantiationScope->getInstantiationOfIfExists(PVD)) {
+            // Copy the pack expansion logic from
+            // addInstantiatedParametersToScope
+            CurrentInstantiationScope->MakeInstantiatedLocalArgPack(PVD);
```
zyn0217, line 925: Can we stop doing that?

clang/lib/Sema/SemaTemplateVariadic.cpp
```diff
@@ -911,6 +911,34 @@ bool Sema::CheckParameterPacksForExpansion(
     // Determine the size of this argument pack.
     unsigned NewPackSize, PendingPackExpansionSize = 0;
     if (IsVarDeclPack) {
+
+      // if there is a function paramter pack and we directly call findInstantiationOf
+      // then it fails because localdecls are not added to the instantiation scope until 
+      // after we check for pack expansions. So we need to check for parameter packs here 
+      // and add them to the instantiation scope before we call findInstantiationOf.
```
zyn0217, line 918: I don't think this is a right approach to address any problems from the point where we start to check expanded packs.
We should explore why we missed a declaration instantiation instead.

zyn0217 (changes requested):

clang/lib/Sema/SemaTemplateVariadic.cpp
albertbolt1: but the instantiation happens after constraint check in c++26 as per your previous changes right ?

This is failing in the constraint check phase itself.

The parameters here should exist before we check for the constraints ? so we should either do it here or the one where this is called. 

any opinions on this ?

zyn0217: Yes, they are supposed to be instantiated on demand, and I would expect we missed a call to maybeInstantiateFunctionParameterToScope elsewhere.

zyn0217: I.e. we probably want a call to instantiator before checking fold constraints so that they can be found in an LIS.

albertbolt1:
```mermaid
flowchart TD
    A[EvaluateSlow<br/>FoldExpandedConstraint] --> B[EvaluateFoldExpandedConstraintSize]
    
    B --> C[CheckParameterPacksForExpansion]
    
    C --> D{findInstantiationOf<br/>on booleans pack}
    
    D -->|Fails - not in scope| E[Returns nullptr]
    
    E --> F[ShouldExpand = false]
    
    F --> G[FAILURE<br/>Cannot determine pack size]
```

albertbolt1: the above is the path followed for the below program

```
#include <type_traits>
#include <cstdint>

constexpr int64_t NumberOfTrueInstances(auto... booleans)
  requires ((std::is_same_v<bool, decltype(booleans)>) && ...)
{
  bool the_booleans[] = {booleans...};
  int64_t nrvo = 0;
  for (bool a_boolean : the_booleans) {
    if (a_boolean) nrvo += 1;
  }
  return nrvo;
}

int main() {
  constexpr bool a = true;
  constexpr bool b = false;
  constexpr bool c = true;
  constexpr bool d = false;
  constexpr int64_t count = NumberOfTrueInstances(a, b, c, d);
  static_assert(count == 2);
}
```

albertbolt1:
@cor3ntin kindly suggest what can be done here

cor3ntin:
Can you provide a description of the problem you are trying to solve and a reduced test (that does not depend on the standard library).

I suspect the issue here is similar to https://github.com/llvm/llvm-project/pull/200185 - although my solution is not right either.

When doing substitution in the parameter mapping of the pattern of fold expanded constraint, we should preserve packs expansions. then when evaluating we should expand each pack to preserve _only_ the argument of the current expansion in the mapping, and then substituting the expression.

albertbolt1:
Closing this as the solution is provided in 
https://github.com/llvm/llvm-project/pull/221707

PR #198211, opened 2026-05-17
[analyzer][NFC] Merge the class `LocationContext` into `StackFrame`

tigbr:
This commit implements the final part of the roadmap described in issue #190973 by merging the class `LocationContext` into the class `StackFrame` and replacing `LocationContext` with `StackFrame` throughout the analyzer.


clang/include/clang/Analysis/AnalysisDeclContext.h
```diff
@@ -200,44 +198,58 @@ class AnalysisDeclContext {
...
   AnalysisDeclContext *getAnalysisDeclContext() const { return Ctx; }
 
   /// It might return null.
-  const LocationContext *getParent() const { return Parent; }
+  const StackFrame *getParent() const { return Parent; }
+
+  int64_t getID() const { return ID; }
 
-  bool isParentOf(const LocationContext *LC) const;
+  bool isParentOf(const StackFrame *LC) const;
```
steakhal: I think your PR renames parameters, so probably `LC` should be renamed too. Keep them in sync with the definition.

clang/include/clang/Analysis/AnalysisDeclContext.h
```diff
@@ -271,94 +284,52 @@ class LocationContext : public llvm::FoldingSetNode {
...
-
-  static bool classof(const LocationContext *LC) {
-    return LC->getKind() == StackFrameKind;
-  }
 };
 
-class LocationContextManager {
-  llvm::FoldingSet<LocationContext> Contexts;
+class StackFrameManager {
+  llvm::FoldingSet<StackFrame> Contexts;
```
steakhal: I think now that `LocationContext` was replaced with `StachFrame`, maybe we should also rename this to `Frames` or something similar.

clang/include/clang/StaticAnalyzer/Core/PathSensitive/ExprEngine.h
```diff
@@ -162,22 +161,22 @@ class ExprEngine {
...
-  /// `const CFGBlock *CurrBlock` & `const LocationContext *CurrLocationContext`
+  /// `const CFGBlock *CurrBlock` & `const StackFrame *CurrStackFrame`
   /// that are kept up-to-date and are almost always non-null during the
   /// analysis. I will switch to this more natural representation when
   /// `NodeBuilder`s are eliminated from the code.
   const NodeBuilderContext *currBldrCtx = nullptr;
   /// Historically `currBldrCtx` pointed to a local variable in some stack
   /// frame. This field is introduced as a temporary measure to allow a gradual
-  /// transition. Only use this in {re,}setLocationContextAndBlock!
+  /// transition. Only use this in {re,}setStackFrameAndBlock!
```
steakhal: Do we have a function called `setStackFrameAndBlock` and `resetStackFrameAndBlock`?

clang/include/clang/StaticAnalyzer/Core/PathSensitive/ExprEngine.h
```diff
@@ -205,7 +204,7 @@ class ExprEngine {
   virtual ~ExprEngine() = default;
 
   /// Returns true if there is still simulation state on the worklist.
-  bool ExecuteWorkList(const LocationContext *L, unsigned Steps = 150000) {
+  bool ExecuteWorkList(const StackFrame *L, unsigned Steps = 150000) {
```
steakhal: Sometimes you rename the parameter names, sometimes you don't (like here).
How did you decide when to rename a parameter and when not?

steakhal (commented):
> This commit implements the final part of the roadmap described in issue https://github.com/llvm/llvm-project/issues/190973 by merging the LocationContext class into the StackFrame class and replacing LocationContext with StackFrame throughout the analyzer.

Could you elaborate on how this change was made?
Is it something mechanical? If so, what was the procedure?

clang/include/clang/StaticAnalyzer/Core/PathSensitive/ExprEngine.h
tigbr: Yes, there are, however, they are called `setCurrStackFrameAndBlock` and `resetCurrStackFrameAndBlock`. I will update the comment.

clang/include/clang/StaticAnalyzer/Core/PathSensitive/ExprEngine.h
tigbr: I have answered this is in a longer comment too. Basically, any outdated names are unintentional and were missed by accident. I will rename these.

tigbr:
I replaced `LocationContext` incrementally and made sure that every commit compiles and passes the `check-clang-analysis` tests. The order in which I have performed the renaming was decided on how many changes would be needed to fix compilation errors. I was trying to make sure that if I break something I find out about it immediately and not just at the end when there are 2k modified lines. Additionaly, I was hoping that it would allow you guys to perform the review incrementally.

I intended to replace the variable and method names as well. In the first four commits, however, I was not really sure whether to do variable, method renaming in separate commits or together with replacing `LocationContext`. Afterwards, I have settled on doing these at the same time. So, I think I have just forgot to go back and rename those variables whose type I have replaced in the first few commits. In short, there should not be any outdated names. In case you do find any it was just accidentally missed, which I will fix.

By default, I used the name `SF` for `StackFrame` variables, except for some special cases such as `CallerSF`, `CalleeSF`, `LeakFrame` (for `LeakContext`) etc.

In general, this was my methodology. If you would like to know anything more specific I will try to answer as best as I can.

steakhal:
Great work btw. The individual commits are great, it just wasn't apparent to me that that is the desired way of reviewing these changes. This explanation works for me.

If I understand correctly, many of these commits overlap. So we wouldn't really gain much by landing these one-by-one.
While I'd be tempted to say to land them commit by commit, it's probably easier if we do a thorough review on this 5k+ diff and merge it in one go - while having a detailed squash message.

What's your opinion @NagyDonat?

tigbr:
By the way @steakhal. When do you think we should update the output of debug dumping methods? Such as in `SymbolConjured::dumpToStream` in `SymbolManager.cpp` and `StackFrame::printJson` in `AnalysisDeclContext.cpp`. These still output strings such as `LC`, `lctx_id` and `location_context`. I have not modified these in this PR yet. Maybe in a follow-up patch right after this one?

I have a similar question about the `exploded-graph-rewriter.py` script too, since it has its own `LocationContext` class.

Lastly, extending `LocationContext` is mentioned as a solution to a problem in the `clang/www/analyzer/open_projects.html` file. This is obviously going to be outdated by this refactoring. What are the conventions of updating that file?

steakhal:
> By the way @steakhal. When do you think we should update the output of debug dumping methods? Such as in `SymbolConjured::dumpToStream` in `SymbolManager.cpp` and `StackFrame::printJson` in `AnalysisDeclContext.cpp`. These still output strings such as `LC`, `lctx_id` and `location_context`. I have not modified these in this PR yet. Maybe in a follow-up patch right after this one?
> 
> I have a similar question about the `exploded-graph-rewriter.py` script too, since it has its own `LocationContext` class.

You can change anything, but the output format. You can leave a FIXME there to uplift the output format as well while uplifting the rewriter. However, when updating the rewriter, it should remain backward compatible and parse old dumps correctly. This "breaking change" must be done in a specific PR, with only that change to make it a focused change.

> Lastly, extending `LocationContext` is mentioned as a solution to a problem in the `clang/www/analyzer/open_projects.html` file. This is obviously going to be outdated by this refactoring. What are the conventions of updating that file?

I think the safest is to not speculate on possible solutions in that `open_projects.html` file - so I'd just drop that part from there.

clang/include/clang/Analysis/AnalysisDeclContext.h
```diff
@@ -200,44 +198,58 @@ class AnalysisDeclContext {
...
+  StackFrameManager &getStackFrameManager();
 };
 
-class LocationContext : public llvm::FoldingSetNode {
-public:
-  enum ContextKind { StackFrameKind };
-
-private:
-  ContextKind Kind;
+/// It represents a stack frame of the call stack (based on CallEvent).
```
NagyDonat: What does "(based on CallEvent)" mean here?

clang/include/clang/Analysis/AnalysisDeclContext.h
```diff
@@ -250,115 +262,74 @@ class LocationContext : public llvm::FoldingSetNode {
...
+                      const StackFrame *SF, const void *Data, const Expr *E,
+                      const CFGBlock *Block, unsigned BlockCount,
                       unsigned Index) {
-    ProfileCommon(ID, StackFrameKind, ADC, ParentLC, E);
+    ProfileCommon(ID, ADC, SF, E);
     ID.AddPointer(Data);
     ID.AddPointer(Block);
     ID.AddInteger(BlockCount);
     ID.AddInteger(Index);
   }
```
NagyDonat: Please merge `ProfileCommon` and `Profile`: If I understand correctly, the difference between them was that `ProfileCommon` handled the attributes of `LocationContext`, while `Profile` handled the attributes of `StackFrame`(`Context`).

clang/include/clang/Analysis/AnalysisDeclContext.h
```diff
@@ -424,8 +394,8 @@ class AnalysisDeclContextManager {
   ///
   /// \returns The top level stack frame for \p D.
   const StackFrame *getStackFrame(const Decl *D) {
```
NagyDonat: Consider renaming this method to `getTopStackFrame`. (Perhaps in a follow-up commit?)

clang/include/clang/Analysis/ProgramPoint.h
```diff
@@ -95,32 +94,32 @@ class ProgramPoint {
   const void *Data1;
   llvm::PointerIntPair<const void *, 2, unsigned> Data2;
 
-  // The LocationContext could be NULL to allow ProgramPoint to be used in
+  // The StackFrame could be NULL to allow ProgramPoint to be used in
   // context insensitive analysis.
```
NagyDonat, line 98: As far as I know, this is not actually used – `ProgramPoint`s are tightly connected to our path sensitive analysis.

@tigbr If you can easily confirm that the `StackFrame` is never null here, then consider adding a FIXME note that this feature is not actually used and perhaps should be removed to simplify the codebase. (If this would be a significant detour, then don't bother with it. If it would be very simple, you could exclude this "StackFrame could be NULL" case within this PR.

clang/include/clang/Analysis/ProgramPoint.h
```diff
@@ -95,32 +94,32 @@ class ProgramPoint {
...
+  ProgramPoint(const void *P, Kind k, const StackFrame *SF,
                const ProgramPointTag *tag = nullptr,
                CFGBlock::ConstCFGElementRef ElemRef = {nullptr, 0})
       : Data1(P), Data2(nullptr, (((unsigned)k) >> 0) & 0x3),
-        L(l, (((unsigned)k) >> 2) & 0x3), Tag(tag, (((unsigned)k) >> 4) & 0x3),
+        S(SF, (((unsigned)k) >> 2) & 0x3), Tag(tag, (((unsigned)k) >> 4) & 0x3),
         ElemRef(ElemRef) {
     assert(getKind() == k);
-    assert(getLocationContext() == l);
+    assert(getStackFrame() == SF);
```
NagyDonat, line 114: As drive-by cleanup, please capitalize the variable name `K`. It is the only non-capitalized name and you're already touching most of these lines anyway.

clang/include/clang/Analysis/ProgramPoint.h
```diff
@@ -95,32 +94,32 @@ class ProgramPoint {
...
   }
 
-  ProgramPoint(const void *P1, const void *P2, Kind k, const LocationContext *l,
+  ProgramPoint(const void *P1, const void *P2, Kind k, const StackFrame *SF,
                const ProgramPointTag *tag = nullptr,
                CFGBlock::ConstCFGElementRef ElemRef = {nullptr, 0})
       : Data1(P1), Data2(P2, (((unsigned)k) >> 0) & 0x3),
-        L(l, (((unsigned)k) >> 2) & 0x3), Tag(tag, (((unsigned)k) >> 4) & 0x3),
+        S(SF, (((unsigned)k) >> 2) & 0x3), Tag(tag, (((unsigned)k) >> 4) & 0x3),
         ElemRef(ElemRef) {}
```
NagyDonat, line 123: As above, please capitalize `k` for the sake of completeness.

clang/include/clang/Analysis/ProgramPoint.h
```diff
@@ -133,8 +132,8 @@ class ProgramPoint {
   /// Create a new ProgramPoint object that is the same as the original
   /// except for using the specified tag value.
   ProgramPoint withTag(const ProgramPointTag *tag) const {
-    return ProgramPoint(getData1(), getData2(), getKind(),
-                        getLocationContext(), tag);
+    return ProgramPoint(getData1(), getData2(), getKind(), getStackFrame(),
+                        tag);
```
NagyDonat, line 136: Consider renaming `tag` to `T`. It would fix the capitalization and – if I count correctly – would ensure that the return statement fits into 80 characters.

clang/include/clang/Analysis/ProgramPoint.h
```diff
@@ -339,13 +332,12 @@ class PostStmt : public StmtPoint {
 
 class FunctionExitPoint : public ProgramPoint {
 public:
-  explicit FunctionExitPoint(const ReturnStmt *S,
-                             const LocationContext *LC,
+  explicit FunctionExitPoint(const ReturnStmt *S, const StackFrame *SF,
                              const ProgramPointTag *tag = nullptr)
-      : ProgramPoint(S, FunctionExitKind, LC, tag) {}
+      : ProgramPoint(S, FunctionExitKind, SF, tag) {}
```
NagyDonat, line 337: Many other lines where I would propose capitalizing `k` (and perhaps also `tag`). As most of these lines are already touched, I think this is the best opportunity to standardize the variable names with minimal disturbances.

clang/include/clang/Analysis/ProgramPoint.h
```diff
@@ -638,20 +629,20 @@ class PostAllocatorCall : public StmtPoint {
 class CallEnter : public ProgramPoint {
 public:
   CallEnter(const Stmt *stmt, const StackFrame *CalleeSF,
-            const LocationContext *callerCtx)
-      : ProgramPoint(stmt, CalleeSF, CallEnterKind, callerCtx, nullptr) {}
+            const StackFrame *callerSF)
+      : ProgramPoint(stmt, CalleeSF, CallEnterKind, callerSF, nullptr) {}
```
NagyDonat: Please capitalize `callerSF` and perhaps also capitalize `stmt`.

clang/include/clang/StaticAnalyzer/Core/BugReporter/BugReporter.h
```diff
@@ -322,9 +321,9 @@ class PathSensitiveBugReport : public BugReport {
   llvm::DenseMap<const MemRegion *, bugreporter::TrackingKind>
       InterestingRegions;
 
-  /// A set of location contexts that correspoind to call sites which should be
+  /// A set of stack frames that correspoind to call sites which should be
```
NagyDonat:
```suggestion
  /// A set of stack frames that correspond to call sites which should be
```
(Just an old typo.)

clang/include/clang/StaticAnalyzer/Core/PathSensitive/MemRegion.h
```diff
@@ -707,26 +706,25 @@ class BlockDataRegion : public TypedRegion {
   friend class MemRegionManager;
 
   const BlockCodeRegion *BC;
-  const LocationContext *LC; // Can be null
+  const StackFrame *SF; // Can be null
```
NagyDonat:
```suggestion
  const StackFrame *SF;
```
The only constructor of this class asserts that this is non-null.

clang/include/clang/StaticAnalyzer/Core/PathSensitive/SValBuilder.h
```diff
@@ -173,10 +172,10 @@ class SValBuilder {
   // Forwarding methods to SymbolManager.
 
   const SymbolConjured *conjureSymbol(ConstCFGElementRef Elem,
-                                      const LocationContext *LCtx,
-                                      QualType type, unsigned visitCount,
+                                      const StackFrame *SF, QualType type,
+                                      unsigned visitCount,
                                       const void *symbolTag = nullptr) {
-    return SymMgr.conjureSymbol(Elem, LCtx, type, visitCount, symbolTag);
+    return SymMgr.conjureSymbol(Elem, SF, type, visitCount, symbolTag);
```
NagyDonat, line 178: Consider capitalizing the other variable names as well.

clang/include/clang/StaticAnalyzer/Core/PathSensitive/SymbolManager.h
```diff
@@ -83,16 +83,15 @@ class SymbolConjured : public SymbolData {
...
 
   friend class SymExprAllocator;
-  SymbolConjured(SymbolID sym, ConstCFGElementRef elem,
-                 const LocationContext *lctx, QualType t, unsigned count,
-                 const void *symbolTag)
-      : SymbolData(ClassKind, sym), Elem(elem), T(t), Count(count), LCtx(lctx),
+  SymbolConjured(SymbolID sym, ConstCFGElementRef elem, const StackFrame *SF,
+                 QualType t, unsigned count, const void *symbolTag)
+      : SymbolData(ClassKind, sym), Elem(elem), T(t), Count(count), SF(SF),
         SymbolTag(symbolTag) {
```
NagyDonat, line 93: Please capitalize all those variable names as you are already touching these lines.

clang/include/clang/StaticAnalyzer/Core/CheckerManager.h
```diff
@@ -523,13 +521,11 @@ class CheckerManager {
...
 
-  using CheckRegionChangesFunc =
-      CheckerFn<ProgramStateRef (ProgramStateRef,
-                                 const InvalidatedSymbols *symbols,
-                                 ArrayRef<const MemRegion *> ExplicitRegions,
-                                 ArrayRef<const MemRegion *> Regions,
-                                 const LocationContext *LCtx,
-                                 const CallEvent *Call)>;
+  using CheckRegionChangesFunc = CheckerFn<ProgramStateRef(
+      ProgramStateRef, const InvalidatedSymbols *symbols,
```
NagyDonat, line 525:
```suggestion
      ProgramStateRef, const InvalidatedSymbols *Symbols,
```
Yet another capitalization suggestion.

NagyDonat (commented):
I reviewed the changes in header files and added many suggestions, mostly about issues that were already present in the codebase before this PR.

I have to admit that this PR is **big**, I really appreciate the effort invested into creating it!

I will continue with reviewing the rest of this patch in the upcoming days.

clang/include/clang/Analysis/AnalysisDeclContext.h
steakhal: Maybe `getInitialStackFrame`?

NagyDonat:
> If I understand correctly, many of these commits overlap. So we wouldn't really gain much by landing these one-by-one. While I'd be tempted to say to land them commit by commit, it's probably easier if we do a thorough review on this 5k+ diff and merge it in one go - while having a detailed squash message.
> 
> What's your opinion @NagyDonat?

I admit that this commit is bigger than what I expected, especially if I want to read it line-by-line.

I feel that landing ~40 commits individually would be definitely too much, but perhaps if there is some clear logical separation, then we could split this into 3-4-5-6 PRs that are a bit smaller.

However, there are also advantages in squashing this large PR into a single commit – obviously, it is less work, and moreover that would have the advantage that it would collect all the changed lines into a single logical unit. (Of course commit messages like "Refactoring... Part 3/5" also create such connections, but those are softer and harder to follow.)

-------

@steakhal In my inline comments I suggested many places where I think that this PR should capitalize variable names in trivial methods (1-4 lines) where most of those lines are already touched by this PR. I feel that this is a unique opportunity – we probably won't touch those methods in the next 10 years – but I know that adding more complexity to an already big PR is a bit problematic. What do you think about this?

clang/include/clang/Analysis/AnalysisDeclContext.h
NagyDonat: That is also a good name, but I prefer `Top` because e.g. it is consistent with the method name `inTopFrame()`.

steakhal:
> > If I understand correctly, many of these commits overlap. So we wouldn't really gain much by landing these one-by-one. While I'd be tempted to say to land them commit by commit, it's probably easier if we do a thorough review on this 5k+ diff and merge it in one go - while having a detailed squash message.
> > What's your opinion @NagyDonat?
> 
> I admit that this commit is bigger than what I expected, especially if I want to read it line-by-line.
> 
> I feel that landing ~40 commits individually would be definitely too much, but perhaps if there is some clear logical separation, then we could split this into 3-4-5-6 PRs that are a bit smaller.
> 
> However, there are also advantages in squashing this large PR into a single commit – obviously, it is less work, and moreover that would have the advantage that it would collect all the changed lines into a single logical unit. (Of course commit messages like "Refactoring... Part 3/5" also create such connections, but those are softer and harder to follow.)
> 
> @steakhal In my inline comments I suggested many places where I think that this PR should capitalize variable names in trivial methods (1-4 lines) where most of those lines are already touched by this PR. I feel that this is a unique opportunity – we probably won't touch those methods in the next 10 years – but I know that adding more complexity to an already big PR is a bit problematic. What do you think about this?

I'd suggest squashing what we currently have. Try to close what we have. You can have more commits if you want (**not** squashing those into what we already have!) and then decide to split them out or let them in in this PR.
Splitting at that point would be trivial.

NagyDonat:
@steakhal Unfortunately I don't really understand what do you mean (which is perhaps my fault, I'm not at 100% mental capacity right now :sweat_smile:).

> I'd suggest squashing what we currently have.

Exactly what do you mean under "squashing"? AFAIK on github the best workflow is that after creating a PR you never squash or otherwise alter history (because it would require a force push that messes up the code review) – until the point when the PR is accepted and then github squashes everything into a single commit.

> Try to close what we have.

Close what? This PR? The open inline discussions?

> You can have more commits if you want (**not** squashing those into what we already have!) and then decide to split them out or let them in in this PR.

I agree that it is reasonable to collect commits on this big PR, and then later we can easily split off some of them for future commits if we end up preferring that.

-----

In my current somewhat tired and pessimistic state of mind I also toyed with the idea of renaming `StackFrame` (again) and keeping the very widespread class name `LocationContext` unchanged – in that case this ±2500 line commit could be replaced by a ± few hundred commit (merging the implementations of the two classes + touching the lines that were touched when StackFrameContext was renamed to StackFrame + touching `getStackFrame`-like methods).

I think the more ambitious plan implemented in this PR is better overall, but I think it is worth to mention this alternative explicitly.

steakhal:
>> I'd suggest squashing what we currently have.

I can see 39 commits on this PR. We have a rough agreement that the scope of these commits (or the scope of the PR) is fine. So I'd propose an interactive rebase and squash these 39 commits into a single commit with an updated commit message - or you could just keep the squash messages - that might be even better for history.

I don't know what upcoming changes you suggested because I didn't look. We could include them still in this PR, but it depends on their content. This is why I suggested to keep them separate from the squashed content.

After this we would have 2 options:
 - We just merge (in other words) close this PR with the squashed commit only.
 - Or we include the additional commits and squash-merge those as well.
 
 Personally, I'd prefer the easiest: the first option and reserve the suggestions for a subsequent PR - but again, this is just a hunch because I haven't looked what you proposed or how those commits would relate to what I could see here right now.
 
 
 > In my current somewhat tired and pessimistic state of mind I also toyed with the idea of renaming StackFrame (again) and keeping the very widespread class name LocationContext unchanged – in that case this ±2500 line commit could be replaced by a ± few hundred commit (merging the implementations of the two classes + touching the lines that were touched when StackFrameContext was renamed to StackFrame + touching getStackFrame-like methods).
>
> I think the more ambitious plan implemented in this PR is better overall, but I think it is worth to mention this alternative explicitly.
 
 I think @tigbr already put so much effort into this, I want to just get over with this one. This is the direction we want to go. It is just a matter of chunking. I'll try to spare some time and have a thorough look in the upcoming days.

NagyDonat:
> We have a rough agreement that the scope of these commits (or the scope of the PR) is fine.

Most of my review suggestions are „maybe here, maybe in the next PR” (mostly changes in capitalization), but I also added a few suggestions that IMO should definitely be fixed within this PR (not because they are severe issues, but because they are clearly connected to this PR).

> So I'd propose an interactive rebase and squash these 39 commits into a single commit with an updated commit message - or you could just keep the squash messages - that might be even better for history.

Where would you want to put that squashed commit? Into a new PR (abandoning this)? Into this PR with a force push (potentially messing up the inline review comments)?

I don't see any advantage of doing any squash (or other history breaking change) before the final squash-and-merge to main. If we want to say that "these 39 commits are definitely going in together", we can just say that. If we want to compose the commit message that will end up on main, we can edit the PR description or start a new text file etc.

-------

> I think @tigbr already put so much effort into this, I want to just get over with this one. This is the direction we want to go. It is just a matter of chunking. I'll try to spare some time and have a thorough look in the upcoming days.

I feel the same way overall, I just thought that it would be prudent to mention this alternative.

@tigbr Thanks for implementing this and sorry for burdening you with this big change!

steakhal:
> So I'd propose an interactive rebase and squash these 39 commits into a single commit with an updated commit message - or you could just keep the squash messages - that might be even better for history.
Where would you want to put that squashed commit? Into a new PR (abandoning this)? Into this PR with a force push (potentially messing up the inline review comments)?

I wanted to keep the discussion open for additional changes because you foreshadowed more commits that you could see part of this PR. I didn't want to immediately push back and reject the idea because again, that depend on the nature of those changes.
The only way to differentiate from the changes we already have and the future changes is to have them in separate commits, but then the next X commits would blend into the 39 commits we already agreed on so far.

However, if we agree on not adding more commits, then in deed there is no reason to manually squash them.
Note that when a force-push happens the old HEAD commit hash will be still visible on the PR, albeit less obvious. 
The notes about inline comments is probably valid. They would be at risk of getting lost.

NagyDonat:
> you foreshadowed more commits that you could see part of this PR. I didn't want to immediately push back and reject the idea because again, that depend on the nature of those changes.

Pushback against more commits on this already huge PR is very natural, I can readily accept it. I have only three suggestions that logically "belong to" this PR:
- [A question about a newly added comment](https://github.com/llvm/llvm-project/pull/198211#discussion_r3305458865) that I don't understand.
- [Completing the merging of implementations of `LocationContext` and `StackFrame`](https://github.com/llvm/llvm-project/pull/198211#discussion_r3305476828) by unifying the `Profile` method parts.
- [Capitalizing a freshly introduced variable name](https://github.com/llvm/llvm-project/pull/198211#discussion_r3305653680)

And there are two others that are very trivial and objectively useful:

- [Fixing an old typo on an already modified line](https://github.com/llvm/llvm-project/pull/198211#discussion_r3305658151)
- [Deleting a factually inaccurate comment on an already modified line](https://github.com/llvm/llvm-project/pull/198211#discussion_r3305679709)

The rest is just capitalizing unrelated variable names on lines that are touched by this commit. I'm personally annoyed by the lowercase variable names, and feel that they "look dirty", give a first impression that the code containing them was edited in a careless manner. This is a useful indicator in code that is actually old and hacky and awkwardly complicated, but I feel that a cleanup is "unfinished" if it leaves these harmless but ugly signs behind. However, I can understand it if this is my personal preference and I won't push for it against opposition.

(I'd also be happy to create a separate PR that just renames variables on lines that were touched by this PR – I'm lobbying for including these changes in this PR primarily because I know that we have a policy against commits that just rename variables.) 

> The only way to differentiate from the changes we already have and the future changes is to have them in separate commits, but then the next X commits would blend into the 39 commits we already agreed on so far.

I see, but it is easy to remember the number „39” and view the changes accordingly. In the Github UI it is easy to view the diff from many, but not all commits on the PR.

clang/include/clang/Analysis/AnalysisDeclContext.h
tigbr: This comment was already there in the file. I have just moved it when deleting `LocationContext` and that is why it appears as a new addition. `CallEvent` has a pointer field that points to a `StackFrame`. At a first glance I am not sure what deeper meaning it could have.

steakhal: Could we mark this `final`?

clang/include/clang/Analysis/ProgramPoint.h
```diff
@@ -95,32 +94,32 @@ class ProgramPoint {
...
     assert(getData1() == P);
   }
 
-  ProgramPoint(const void *P1, const void *P2, Kind k, const LocationContext *l,
+  ProgramPoint(const void *P1, const void *P2, Kind k, const StackFrame *SF,
                const ProgramPointTag *tag = nullptr,
                CFGBlock::ConstCFGElementRef ElemRef = {nullptr, 0})
       : Data1(P1), Data2(P2, (((unsigned)k) >> 0) & 0x3),
-        L(l, (((unsigned)k) >> 2) & 0x3), Tag(tag, (((unsigned)k) >> 4) & 0x3),
+        S(SF, (((unsigned)k) >> 2) & 0x3), Tag(tag, (((unsigned)k) >> 4) & 0x3),
```
steakhal, line 122: Why isn't the member called `SF` if we pass `SF`?

clang/lib/StaticAnalyzer/Core/ExprEngine.cpp
```diff
@@ -786,20 +771,19 @@ static void printIndicesOfElementsToConstructJson(
   }
 
   if (HasItem)
-    Indent(Out, --Space, IsDot) << ']'; // End of "location_context".
+    Indent(Out, --Space, IsDot) << ']'; // End of "stack_frame".
```
steakhal: I don't think we changed the output spelling.

clang/lib/StaticAnalyzer/Core/ExprEngine.cpp
```diff
@@ -724,18 +709,18 @@ printObjectsUnderConstructionJson(raw_ostream &Out, ProgramStateRef State,
...
```
steakhal: I don't think we changed the output spelling.

steakhal (commented):
I finished my review of these 39 commits.  
I found only a few remarks. This looks good to me.

clang/include/clang/Analysis/ProgramPoint.h
tigbr: `S` is a `PointerIntPair<const StackFrame *, 2, unsigned>` so I was more hesitant to rename it to `SF`. If you do not find it confusing then I can change it.

clang/lib/StaticAnalyzer/Core/ExprEngine.cpp
tigbr: Agreed, I will undo that.

clang/lib/StaticAnalyzer/Core/ExprEngine.cpp
tigbr: Agreed, I will undo that.

clang/include/clang/Analysis/ProgramPoint.h
tigbr: For some reason the Github ui does not let me reply to the suggestion about `final` directly so I will reply here. 

So by making it `final` we document that a class hierarchy should not be made out of `StackFrame` (as was with `LocationContext`). That's a nice touch I will do that.

clang/lib/StaticAnalyzer/Core/ExprEngine.cpp
tigbr: Done in [eb8f100](https://github.com/llvm/llvm-project/pull/198211/commits/eb8f100cf3cab242546a719f620738386fbd3f94).

clang/lib/StaticAnalyzer/Core/ExprEngine.cpp
tigbr: Done in [eb8f100](https://github.com/llvm/llvm-project/pull/198211/commits/eb8f100cf3cab242546a719f620738386fbd3f94).

tigbr:
I have implemented review suggestions in commits [efd4336](https://github.com/llvm/llvm-project/pull/198211/commits/efd4336d6607525f309ab79dd24055fb2e659236), [a50cbc2](https://github.com/llvm/llvm-project/pull/198211/commits/a50cbc2658c868280883ff9fce2ee8390fffb036), [eb8f100](https://github.com/llvm/llvm-project/pull/198211/commits/eb8f100cf3cab242546a719f620738386fbd3f94) (The three most recent from this comment.). Renaming `S` to `SF` in `ProgramPoint.h` is not implemented yet, I am waiting for a final confirmation on that.

clang/include/clang/Analysis/ProgramPoint.h
steakhal: Keeping it as `S` is fine, thanks. Works either way.

steakhal (approved):
Yeey 🚀

clang/lib/Analysis/AnalysisDeclContext.cpp
```diff
@@ -419,11 +411,10 @@ const StackFrame *StackFrameManager::getStackFrame(
   llvm::FoldingSetNodeID ID;
   StackFrame::Profile(ID, Ctx, Parent, Data, E, Blk, BlockCount, StmtIdx);
   void *InsertPos;
-  auto *L =
-      cast_or_null<StackFrame>(Contexts.FindNodeOrInsertPos(ID, InsertPos));
+  auto *L = cast_or_null<StackFrame>(Frames.FindNodeOrInsertPos(ID, InsertPos));
```
NagyDonat: I suspect that `FindNodeOrInsertPos` already returns a `StackFrame*`. Check it and if it does, remove the cast.

clang/include/clang/Analysis/ProgramPoint.h
NagyDonat: I'm resolving this inline comment because I created a separate issue #200162 for tracking it.

clang/lib/StaticAnalyzer/Checkers/RetainCountChecker/RetainCountDiagnostics.cpp
```diff
@@ -508,16 +507,16 @@ RefCountReportVisitor::VisitNode(const ExplodedNode *N, BugReporterContext &BRC,
...
 
         // We have an argument.  Get the effect!
         DeallocSent = true;
       }
     } else if (const ObjCMessageExpr *ME = dyn_cast<ObjCMessageExpr>(S)) {
       if (const Expr *receiver = ME->getInstanceReceiver()) {
-        if (CurrSt->getSValAsScalarOrLoc(receiver, LCtx)
-              .getAsLocSymbol() == Sym) {
+        if (CurrSt->getSValAsScalarOrLoc(receiver, SF).getAsLocSymbol() ==
+            Sym) {
```
NagyDonat: Perhaps rename `receiver` to `R` to fix capitalization and avoid the line break?

clang/lib/StaticAnalyzer/Core/BugReporter.cpp
```diff
@@ -176,18 +175,18 @@ class PathDiagnosticConstruct {
     return getParentMap().getParent(S);
   }
 
-  void updateLocCtxMap(const PathPieces *Path, const LocationContext *LC) {
-    assert(Path && LC);
-    LCM[Path] = LC;
+  void updateLocCtxMap(const PathPieces *Path, const StackFrame *SF) {
```
NagyDonat: Please rename this function and any others that contain `LocCtx` in their names (e.g. `isInLocCtxMap` below).

clang/lib/StaticAnalyzer/Core/Environment.cpp
```diff
@@ -198,36 +198,35 @@ void Environment::printJson(raw_ostream &Out, const ASTContext &Ctx,
...
+  if (!SF) {
     // Find the freshest location context.
-    llvm::SmallPtrSet<const LocationContext *, 16> FoundContexts;
+    llvm::SmallPtrSet<const StackFrame *, 16> FoundStackFrames;
     for (const auto &I : *this) {
-      const LocationContext *LC = I.first.getLocationContext();
-      if (FoundContexts.count(LC) == 0) {
+      const StackFrame *currentSF = I.first.getStackFrame();
+      if (FoundStackFrames.count(currentSF) == 0) {
         // This context is fresher than all other contexts so far.
```
NagyDonat: Replace "location context" with "stack frame" in these comments.

NagyDonat (approved):
I finished reviewing the rest of the PR. The change LGTM and I'm really grateful that you implemented this.

I have three minor suggestions in inline comments.

PR #198213, opened 2026-05-17
[SPIR-V] Add vector type support for non-standard integers in G_TRUNC op

aobolensk:
`G_TRUNC` is the special case for now that require these changes because it is the only one that depend on the original destination

### DIFF
--- a/llvm/lib/Target/SPIRV/SPIRVPreLegalizer.cpp
+++ b/llvm/lib/Target/SPIRV/SPIRVPreLegalizer.cpp
@@ -514,30 +514,44 @@ generateAssignInstrs(MachineFunction &MF, SPIRVGlobalRegistry *GR,
         Register DstReg = MI.getOperand(0).getReg();
         Register SrcReg = MI.getOperand(1).getReg();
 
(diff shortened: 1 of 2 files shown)


MrSidims (approved):
Thanks!
Q: This fixes G_TRUNC specifically, but widenScalarType still early-returns on vectors, so other vector ops on non-standard int element types remain un-widened. Is G_TRUNC the only op that currently reaches here with such types, or is this the first of several follow-ups? Worth a note in the description.

aobolensk:
> Thanks! Q: This fixes G_TRUNC specifically, but widenScalarType still early-returns on vectors, so other vector ops on non-standard int element types remain un-widened. Is G_TRUNC the only op that currently reaches here with such types, or is this the first of several follow-ups? Worth a note in the description.

So far, `G_TRUNC` is special because it is the only one that depend on the original destination. `G_ZEXT/G_SEXT/G_ANYEXT` are much simpler as of now, not really sure if anything like that should be added there as well, but in that case it should be changes for the scalar case first

jmmartinez (approved):
LGTM

PR #198218, opened 2026-05-17
[clang][test] Use `FileCheck` in `Driver/modules-driver-incompatible-options.cpp`

iidmsa:
The test had `CHECK` directives that were never executed because the `RUN` line did not pipe output to `FileCheck`.

This also includes `2>&1` before the pipe to `FileCheck`, since Clang's driver error messages are written to `stderr`.

This also adds `%t/` prefix to `main.cpp`, which should be the intended directory.


carlocab (approved):

AaronBallman (approved):
LGTM!

PR #198221, opened 2026-05-17
[MemoryBuiltins][NFC] Consistently infer and use MallocFamily

jdoerfert:
MallocFamily (the enum and StringRef) are used alongside AllocFnsTy. The latter is picked up from the tables while the former is encoded in the IR. While they should be merged at some point (see TODO), this commit makes sure we consistently initialize the MallocFamily String and pass it to users.


shiltian (commented):
There is no test?

jdoerfert:
> There is no test?

There is no functional change. For the APIs in use we emit the same output as before.

shiltian (approved):
Looks reasonable to me.

shiltian:
This was merged to a personal branch?

jdoerfert:
I hate github.

Issue #156652, opened 2025-09-03
crashes with assertion failure when running opt -passes=coro-early


k-arrows:
Run opt with -passes=coro-early

Reproducer:
https://godbolt.org/z/q1Teh87P8

Backtrace:
```console
opt: /root/llvm-project/llvm/lib/Transforms/Coroutines/CoroEarly.cpp:239: void {anonymous}::Lowerer::lowerEarlyIntrinsics(llvm::Function&): Assertion `F.isPresplitCoroutine() && "The frontend uses Switch-Resumed ABI should emit " "\"presplitcoroutine\" attribute for the coroutine."' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.	Program arguments: /opt/compiler-explorer/clang-assertions-trunk/bin/opt -o /app/output.s -S -passes=coro-early <source>
1.	Running pass "coro-early" on module "<source>"
 #0 0x000000000578f178 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/opt/compiler-explorer/clang-assertions-trunk/bin/opt+0x578f178)
 #1 0x000000000578c024 SignalHandler(int, siginfo_t*, void*) Signals.cpp:0:0
 #2 0x00007a732fa42520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #3 0x00007a732fa969fc pthread_kill (/lib/x86_64-linux-gnu/libc.so.6+0x969fc)
 #4 0x00007a732fa42476 gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x42476)
 #5 0x00007a732fa287f3 abort (/lib/x86_64-linux-gnu/libc.so.6+0x287f3)
 #6 0x00007a732fa2871b (/lib/x86_64-linux-gnu/libc.so.6+0x2871b)
 #7 0x00007a732fa39e96 (/lib/x86_64-linux-gnu/libc.so.6+0x39e96)
 #8 0x0000000002f1e634 (anonymous namespace)::Lowerer::lowerEarlyIntrinsics(llvm::Function&) CoroEarly.cpp:0:0
 #9 0x0000000002f1e85f llvm::CoroEarlyPass::run(llvm::Module&, llvm::AnalysisManager<llvm::Module>&) (/opt/compiler-explorer/clang-assertions-trunk/bin/opt+0x2f1e85f)
[...]

PR #198226, opened 2026-05-18
[CoroSplit] Never cross suspending points when `coro.is_in_ramp` is true

NewSigma:
When `coro.is_in_ramp` is true, the marked region does not appear in resumers after splitting and therefore never crosses a suspension point. This patch leverages that property to avoid wasting space on lazy conversion:

``` C++
#include <coroutine>

struct coro {
    struct promise_type {
      auto get_return_object() noexcept {
        return std::coroutine_handle<promise_type>::from_promise(*this);
      }
      std::suspend_always initial_suspend() noexcept { return {}; }
      std::suspend_never final_suspend() noexcept { return {}; }
      void return_void() noexcept {}
      void unhandled_exception() {}
    };

    coro(std::coroutine_handle<promise_type> handle) : handle(handle) {}

    std::coroutine_handle<> handle;
    long l[2];
};

coro fn() {
  co_return;
}
```


ChuanqiXu9 (approved):

efriedma-quic (commented):
This is assuming nothing references the sret pointer before the last suspend, I guess?

Which is maybe what you want, but I'm not sure anything is actually enforcing it.

NewSigma:
> This is assuming nothing references the sret pointer before the last suspend, I guess?
> 
> Which is maybe what you want, but I'm not sure anything is actually enforcing it.

Because coroutine resumers do not return values, the sret pointer is assumed to be live only within ramp functions — specifically, either before the first coroutine suspension or when `coro.is_in_ramp` is true.

efriedma-quic:
Oh, I see, I got it backward...

I guess I still have the same issue, though; is anything actually enforcing that the address isn't used after the first suspend?  The underlying memory might be dead, but that doesn't mean nothing uses it, necessarily.

NewSigma:
> Oh, I see, I got it backward...
> 
> I guess I still have the same issue, though; is anything actually enforcing that the address isn't used after the first suspend? The underlying memory might be dead, but that doesn't mean nothing uses it, necessarily.

I agree that the assumption is fragile and that some transforms may break it. I have therefore adopted a different approach that teaches the semantics of `llvm.coro.is_in_ramp` to `SuspendCrossingInfo`.

llvm/lib/Transforms/Coroutines/SuspendCrossingInfo.cpp
```diff
@@ -164,14 +162,19 @@ SuspendCrossingInfo::SuspendCrossingInfo(
...
     assert(CE->getParent()->getFirstInsertionPt() == CE->getIterator() &&
            CE->getParent()->size() <= 2 && "CoroEnd must be in its own BB");
 
     getBlockData(CE->getParent()).NeverKill = true;
   }
 
+  for (auto *InRamp : Shape.CoroIsInRampInsts)
+    for (auto *U : InRamp->users())
+      if (auto *Br = dyn_cast<CondBrInst>(U))
+        getBlockData(Br->getSuccessor(0)).NeverKill = true;
```
ChuanqiXu9: This seems not good as it didn't assume the successor doesn't have other predecessor. Or in another word, it assumes the usage about coro.is_in_ramp.

NewSigma: Done, thanks!

llvm/lib/Transforms/Coroutines/SuspendCrossingInfo.cpp
```diff
@@ -164,14 +162,22 @@ SuspendCrossingInfo::SuspendCrossingInfo(
...
     getBlockData(CE->getParent()).NeverKill = true;
   }
 
+  for (auto *InRamp : Shape.CoroIsInRampInsts)
+    for (auto *U : InRamp->users())
+      if (auto *Br = dyn_cast<CondBrInst>(U)) {
+        auto *TrueBB = Br->getSuccessor(0);
+        if (TrueBB->getSinglePredecessor())
+          getBlockData(TrueBB).NeverKill = true;
+      }
```
ChuanqiXu9, line 179: Just out of curiosity, I know this won't happen for forseeable future, what if the BB contains a coro.suspend ?

This may be helpful for the soundness of IR. As users may use the intrinsics anywhere in theory.

NewSigma: If the BB contains `coro.suspend`, both `NeverKill` and `AlwaysKill` would be `true`. I think this indicates a design flaw. An assertion is added to guard against this.

ChuanqiXu9 (approved):
SGTM

llvm/lib/Transforms/Coroutines/SuspendCrossingInfo.cpp
NewSigma: Sorry, I recall that fuzzing tests can trigger the assertion anyway. I should try a different approach.

ChuanqiXu9: Why do we have a fuzzing test for this?

NewSigma: There are many downstream attempts to fuzz the assertions. See #156652 for an example.

PR #198229, opened 2026-05-18
[RISCV][MC] Add experimental `Zvvmtls` and `Zvvmttls` support

imkiva:
This patch adds experimental MC-layer support for the [RISC-V Integrated Matrix Extension](https://github.com/riscv/integrated-matrix-extension/releases/tag/riscv-isa-release-71c48b9-2026-05-17), specifically the tile load/store extensions: `Zvvmtls` and `Zvvmttls`

This PR:

- Adds the optional tile lambda operand syntax (`L1` through `L64`), and related asm operand.
- Adds the `vmtl.v`, `vmts.v`, `vmttl.v` and `vmtts.v` instructions to the MC
-  Modifies `parseMaskReg` to return `NoMatch` to allow overloaded mnemonics to continue matching alternative optional operands, such as parsing `vmtl.v v8, (a0), a1, L4` as the tile-lambda form instead of failing by treating `L4` as a malformed mask operand. Real mask registers missing .t, such as v0, still produce the existing diagnostic.


llvm/docs/RISCVUsage.rst
```diff
@@ -364,10 +368,16 @@ The primary goal of experimental support is to assist in the process of ratifica
   LLVM implements the `0.1 draft specification <https://github.com/ved-rivos/riscv-isa-manual/blob/zvzip/src/zvzip.adoc>`__.
 
 ``experimental-zvvfmm``
-  LLVM implements the `0.1 draft specification <https://github.com/riscv/integrated-matrix-extension/releases/tag/riscv-isa-release-fa55752-2026-05-04>`__.
+  LLVM implements the `0.1 draft specification <https://github.com/riscv/integrated-matrix-extension/releases/tag/riscv-isa-release-71c48b9-2026-05-17>`__.
```
wangpc-pp, line 371: It seems it changes every week, do we have a more stable version?

llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp
```diff
@@ -2543,6 +2553,53 @@ ParseStatus RISCVAsmParser::parseVScaleReg(OperandVector &Operands) {
...
+    break;
+  case 32:
+    EncodedLambda = 6;
+    break;
+  case 64:
+    EncodedLambda = 7;
+    break;
+  default:
+    return Error(S, "operand must be L1, L2, L4, L8, L16, L32, or L64");
+  }
```
mshockwave: `EncodedLambda = Log2_32(Lambda) + 1` if Lambda is power-of-two and less than 128?

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
```diff
@@ -16,6 +16,28 @@
...
+  let ParserMatchClass = TileLambdaAsmOperand;
+  let PrintMethod = "printTileLambda";
+  let EncoderMethod = "getImmOpValue";
+  let DecoderMethod = "decodeUImmOperand<3>";
+  let OperandType = "OPERAND_UIMM3";
+  let MCOperandPredicate = [{
+    int64_t Imm;
+    if (!MCOp.evaluateAsConstantImm(Imm))
+      return false;
+    return Imm >= 1 && Imm <= 7;
```
mshockwave: perhaps `Imm && isUInt<3>(Imm)`

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
```diff
@@ -34,6 +56,84 @@ def VScaleOp : RegisterOperand<VMV0> {
...
+  let Inst{25} = vm;
+  let Inst{24-20} = rs2;
+  let Inst{19-15} = rs1;
+  let Inst{14-12} = 0b111;
+  let Inst{11-7} = vs3;
+  let Inst{6-0} = OPC_STORE_FP.Value;
+
+  let mayLoad = 0;
+  let mayStore = 1;
+  let hasSideEffects = 0;
```
mshockwave: nit: use `true` and `false` for consistency?

mshockwave (commented):
please remember to update the unit test

llvm/docs/RISCVUsage.rst
imkiva: Seems not, actually these are intrinsic and semantic changes and they are currently not related to MC

llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp
imkiva: Fixed, thanks~

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
imkiva: Fixed, thanks~

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
imkiva: Okay. I noticed there're many .td files still using 0/1 for boolean values. Should I change them to true/false together?

topperc: > Okay. I noticed there're many .td files still using 0/1 for boolean values. Should I change them to true/false together?

Please don't. It's not worth anyones time to do or review.

llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp
```diff
@@ -2508,8 +2513,13 @@ ParseStatus RISCVAsmParser::parseMaskReg(OperandVector &Operands) {
     return ParseStatus::NoMatch;
 
   StringRef Name = getLexer().getTok().getIdentifier();
-  if (!Name.consume_back(".t"))
-    return Error(getLoc(), "expected '.t' suffix");
+  if (!Name.consume_back(".t")) {
+    // Non-register identifiers may belong to another optional operand in an
+    // overloaded mnemnoic. Let the matcher try those alternatives.
```
topperc:
```suggestion
    // overloaded mnemonic. Let the matcher try those alternatives.
```

llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp
```diff
@@ -2543,6 +2553,31 @@ ParseStatus RISCVAsmParser::parseVScaleReg(OperandVector &Operands) {
   return ParseStatus::Success;
 }
 
+ParseStatus RISCVAsmParser::parseTileLambda(OperandVector &Operands) {
+  if (getLexer().isNot(AsmToken::Identifier))
+    return ParseStatus::NoMatch;
+
+  SMLoc S = getLoc();
+  StringRef Name = getLexer().getTok().getIdentifier();
+  if (!Name.consume_front("L"))
```
topperc: everything else about assembly instructions is in lower case. I find it weird they chose to go with a capital letter here.

llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp
```diff
@@ -2543,6 +2553,31 @@ ParseStatus RISCVAsmParser::parseVScaleReg(OperandVector &Operands) {
...
+    return ParseStatus::NoMatch;
+
+  unsigned Lambda;
+  if (Name.getAsInteger(10, Lambda))
+    return Error(S, "operand must be L1, L2, L4, L8, L16, L32, or L64");
+
+  if (!isPowerOf2_32(Lambda) || Lambda >= 128)
```
topperc: Combine this with the previous `if` instead or repeating the error string twice

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
```diff
@@ -16,6 +16,28 @@
...
+  let DiagnosticString = "operand must be L1, L2, L4, L8, L16, L32, or L64";
+}
+
+def TileLambdaOp : RISCVOp {
+  let ParserMatchClass = TileLambdaAsmOperand;
+  let PrintMethod = "printTileLambda";
+  let EncoderMethod = "getImmOpValue";
+  let DecoderMethod = "decodeUImmOperand<3>";
+  let OperandType = "OPERAND_UIMM3";
+  let MCOperandPredicate = [{
```
topperc: Is this MCOperandPredicate used? I'm wondering why it doesn't allow 0, but the decodeUImmOperand and OPERAND_UIMM3 do allow 0.

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
```diff
@@ -34,14 +56,93 @@ def VScaleOp : RegisterOperand<VMV0> {
...
+  let mayStore = false;
+  let hasSideEffects = false;
+  let Uses = [VL, VTYPE];
+  let VMConstraint = true;
+  let UseNamedOperandTable = true;
+}
+
+class VTileStoreBase<bits<2> mop, dag ins, string opcodestr, string argstr>
+    : RVInst<(outs),
+             ins, opcodestr, argstr, [], InstFormatR> {
```
topperc: Unnecessary line break

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
mshockwave: yeah I was only suggesting to change to true/false for the three lines I highlighted here. In general LLVM tries to avoid mass style refactoring.

llvm/docs/RISCVUsage.rst
mshockwave: given how frequent the draft got released, I'd suggest we stick to a specific revision unless there is any major / breaking change

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
imkiva: Good point. This `MCOperandPredicate` is not used, because the `vlambda=0` is consumed by instructions without lambda, and if the disassembler runs to here, the UImm must be non-zero. So it's better to use `decodeUImmNonZeroOperand` here.

llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp
imkiva: Thanks. fixed~

llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp
imkiva: Oh. fixed

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
imkiva: Thanks, fixed~

llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp
imkiva: Yes, this is really strange. I guess that the spec uses L1 to make the lambda operand easier to read, since lowercase `l1` can be confused with `11` or `I1` (but this is not a problem with a coding font).

Btw, should we also support the lowercase form?

llvm/docs/RISCVUsage.rst
imkiva: Okay. I will stick to 2026-05-17 version and only bump version when there's a breaking/major change

llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp
mshockwave: I'd say we should support both cases, but you could ask for clarification just in case.

imkiva:
ping~

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
```diff
@@ -16,6 +16,22 @@
...
+  let DiagnosticType = "InvalidTileLambda";
+  let DiagnosticString = "operand must be L1, L2, L4, L8, L16, L32, or L64";
+}
+
+def TileLambdaOp : RISCVOp {
+  let ParserMatchClass = TileLambdaAsmOperand;
+  let PrintMethod = "printTileLambda";
+  let EncoderMethod = "getImmOpValue";
+  let DecoderMethod = "decodeUImmNonZeroOperand<3>";
+  let OperandType = "OPERAND_UIMM3";
```
topperc: Should this be OPERAND_UIMM3_NONZERO?

imkiva: Yes. I added OPERAND_UIMM3_NONZERO. Thanks!

llvm/lib/Target/RISCV/AsmParser/RISCVAsmParser.cpp
imkiva: Ok, now the parser supports both L and l for lambda tiles

llvm/lib/Target/RISCV/RISCVInstrInfoZvvm.td
```diff
@@ -111,3 +205,17 @@ let Predicates = [HasStdExtZvvfmm] in {
   def VFQIMMACC_VV : VIFMEMACScaleVV<0b111010, "vfqimmacc.vv">;
   def VF8WIMMACC_VV : VIFMEMACScaleVV<0b111011, "vf8wimmacc.vv">;
 } // Predicates = [HasStdExtZvvfmm]
+
+let Predicates = [HasStdExtZvvmtls] in {
+  def VMTL_V_L : VTileLoadLambda<0b00, "vmtl.v">;
```
topperc: Would it be possible to join the two instructions and use an InstAlias with EmitPriority=0 to handle the parsing of 2 optional arguments? The printer could print nothing when the lambda value is 0.

imkiva: Good idea. I changed td file accordingly

topperc (approved):
LGTM

PR #198231, opened 2026-05-18
[llvm-mc] Add --show-source-loc option to emit source locations

arichardson:
This adds a new --show-source-loc flag (and -asm-show-source-loc
MCTargetOption) to llvm-mc, which emits source location comments
(`# <SourceLoc: filename:line:col>`).
When instructions originate from macro expansions, llvm-mc unwinds the
instantiation stack, emitting intermediate expansion layers as
`# <MacroLoc: filename:line:col>` and reporting the top-level
macro invocation in `SourceLoc`. For .include directives, it also emits
`# <IncludeLoc: filename:line:col>` to report the include stack.

The main motivation for this feature is improving
update_mc_test_checks.py to cleanly map output instructions back to
input lines, enabling automated test updates for assembly files that
include directives like .ifdef or macros.

Assisted-By: Gemini


arichardson:
Hmm spr did somethign weird here.

PR #198233, opened 2026-05-18
[AMDGPU] Defer emitting AMDGPUAsmPrinter local Resource Usage comments until the end

JanekvO:
Defers emitting local resource info to the end such that all propagate resource info is known and resolvable.

Assisted by: Claude Opus 4.6


llvm/lib/Target/AMDGPU/AMDGPUAsmPrinter.cpp
```diff
@@ -941,129 +1065,11 @@ bool AMDGPUAsmPrinter::runOnMachineFunction(MachineFunction &MF) {
   emitDVgprSymbol(MF);
 
   if (isVerbose()) {
-    MCSectionELF *CommentSection =
-        Context.getELFSection(".AMDGPU.csdata", ELF::SHT_PROGBITS, 0);
-    OutStreamer->switchSection(CommentSection);
+    DeferredComments.push_back({&MF.getFunction(), CurrentProgramInfo,
+                                CurrentProgramInfo.getFunctionCodeSize(MF)});
```
arsenm, line 1090: The printed code size is wrong anyway, can it be replaced with a proper computed value?

llvm/lib/Target/AMDGPU/AMDGPUAsmPrinter.cpp
```diff
@@ -743,8 +746,130 @@ void AMDGPUAsmPrinter::emitCommonFunctionComments(
...
+          RI.createTotalNumSGPRs(FuncName, STM.getTargetID().isXnackOnOrAny(),
+                                 OutContext),
+          RI.getSymbol(FuncName, RIK::RIK_PrivateSegSize, OutContext)
+              ->getVariableValue(),
+          Info.CodeSize, IsMemoryBound);
+      continue;
+    }
+
+    const SIProgramInfo &PI = Info.ProgInfo;
+    OutStreamer->emitRawComment(" " + Twine(FuncName) + " Kernel info:", false);
```
arsenm: I'd probably just leave it to the function name without the uninformative "Kernel info" bit

llvm/lib/Target/AMDGPU/AMDGPUAsmPrinter.cpp
```diff
@@ -743,8 +746,130 @@ void AMDGPUAsmPrinter::emitCommonFunctionComments(
...
+    bool IsMemoryBound =
+        F.getFnAttribute("amdgpu-memory-bound").getValueAsBool();
+
+    MCSectionELF *CommentSection =
+        OutContext.getELFSection(".AMDGPU.csdata", ELF::SHT_PROGBITS, 0);
+    OutStreamer->switchSection(CommentSection);
+
+    if (!AMDGPU::isEntryFunctionCC(F.getCallingConv())) {
+      using RIK = MCResourceInfo::ResourceInfoKind;
+      OutStreamer->emitRawComment(" " + Twine(FuncName) + " Function info:",
```
arsenm: Just the name?

llvm/lib/Target/AMDGPU/AMDGPUAsmPrinter.cpp
```diff
@@ -743,8 +746,130 @@ void AMDGPUAsmPrinter::emitCommonFunctionComments(
...
+  for (const auto &Info : DeferredComments) {
+    const Function &F = *Info.F;
+    const GCNSubtarget &STM = TM.getSubtarget<GCNSubtarget>(F);
+    MCSymbol *FnSym = TM.getSymbol(&F);
+    StringRef FuncName = FnSym->getName();
+    bool IsMemoryBound =
+        F.getFnAttribute("amdgpu-memory-bound").getValueAsBool();
+
+    MCSectionELF *CommentSection =
+        OutContext.getELFSection(".AMDGPU.csdata", ELF::SHT_PROGBITS, 0);
```
arsenm: The comments are just that, comments. They don't go in a section, certainly not .AMDGPU.csdata

llvm/lib/Target/AMDGPU/AMDGPUAsmPrinter.cpp
```diff
@@ -743,8 +746,130 @@ void AMDGPUAsmPrinter::emitCommonFunctionComments(
...
+        " COMPUTE_PGM_RSRC2:TGID_Z_EN: " + Twine(PI.TGIdZEnable), false);
+    OutStreamer->emitRawComment(" COMPUTE_PGM_RSRC2:TIDIG_COMP_CNT: " +
+                                    Twine(PI.TIdIGCompCount),
+                                false);
+
+    [[maybe_unused]] int64_t PGMRSrc3;
+    assert(STM.getGeneration() >= AMDGPUSubtarget::GFX10 ||
+           STM.hasGFX90AInsts() || STM.hasGFX1250Insts() ||
+           (PI.ComputePGMRSrc3->evaluateAsAbsolute(PGMRSrc3) &&
+            static_cast<uint64_t>(PGMRSrc3) == 0));
```
arsenm: I don't care for this assert. It should really be a hasPGMRsrc3 predicate, and I don't see the point of verifying it's resolved

arsenm (commented):
This is annoying, it separates the comments from the function body which I think really degrades the utility of these. Where does sc put them? We copied these from there

llvm/test/CodeGen/AMDGPU/agpr-register-count.ll
```diff
@@ -158,22 +89,113 @@ declare void @undef_func()
...
+; GCN:    NumVgprs: 32
+; GCN:    NumAgprs: 32
+; GFX908: TotalNumVgprs: 32
+; GFX90A: TotalNumVgprs: 64
+; GFX908: VGPRBlocks: 7
+; GFX90A: VGPRBlocks: 7
+; GFX908: NumVGPRsForWavesPerEU: 32
+; GFX90A: NumVGPRsForWavesPerEU: 64
+; GFX90A: AccumOffset: 32
+; GFX908: Occupancy: 8
```
jayfoad, line 195: This is a nice improvement over:
```
Occupancy: occupancy(10, 4, 256, 8, 10, max(.Lkernel_call_undef_func.numbered_sgpr+extrasgprs(.Lkernel_call_undef_func.uses_vcc, .Lkernel_call_undef_func.uses_flat_scratch, 1), 1, 0), max(totalnumvgprs(.Lkernel_call_undef_func.num_agpr, .Lkernel_call_undef_func.num_vgpr), 1, 0))
```
I guess that is the point of this PR?

llvm/lib/Target/AMDGPU/AMDGPUAsmPrinter.cpp
JanekvO: What's the proper computed value? (AFAICT, the `local_end - function_name` computation doesn't resolve at asm time)

llvm/lib/Target/AMDGPU/AMDGPUAsmPrinter.cpp
JanekvO: From what I gathered, `.AMDGPU.csdata` is always empty in llvm (MC layer). The section switch is superfluous as comments are not emitted within the section but emitted as part of the Asm, with a comment prefix (and thus, also don't end up in the .text). Possibly a leftover of sc to llvm transition? (I haven't figured out how to check sc as of yet, myself)

Let me know what's the best approach for this (removal? emit something within the section?)

arsenm: None of this was ever inspired by sc. What if you just don't change the section?

arsenm: Or switch to something more neutral like .text or .rodata

PR #198238, opened 2026-05-18
[Peephole] Do not fold load if any clobber exists

### LINKED ISSUE #198173: Wrong code at -O1
It seems to be a recent regression, as godbolt cannot reproduce it.
Testcase:
```c
#include <stdint.h>
#include <stdio.h>
__attribute__((__vector_size__(2 * sizeof(int)))) int g2;
int64_t g4, __chk, g24;
_Bool g15;
void launch()
{
    g24 = ~0;
    int64_t v0 = -20166877897343356 * g4;
    g4 = (uint32_t)g24;
    g15 = -67091868267747441 >= v0;
    if (g15) return;
    g2 = 0 >= g2;
}
int main()
{
    launch();
    __chk = g2[1];
    printf("checksum=0x%016llx\n", (unsigned long long)__chk);
}
```

The output at -O0:
```
checksum=0xffffffffffffffff
```

The output at -O1:
```
checksum=0x0000000000000000
```

Reproducible commit: clang version 23.0.0git (git@github.com:llvm/llvm-project.git fa4d73761d122c49ebc591118c83f5e142c98b4d)

XChy:
Fixes #198173, which is introduced by https://github.com/llvm/llvm-project/commit/9525bcc3e06e6c0bff3a85f503e47b06a63d6dcc.
The original patch missed the case where there is a clobber between `LoadMI` and `FlagProducer`.


XChy:
ping.

llvm/lib/CodeGen/PeepholeOptimizer.cpp
```diff
@@ -969,7 +969,12 @@ bool PeepholeOptimizer::optimizeCmpInstr(
     MachineInstr *FlagProducer = MRI->use_nodbg_begin(SrcReg)->getParent();
     MachineInstr *LoadMI = MRI->getVRegDef(SrcReg);
     if (LocalMIs.count(FlagProducer) && LoadMI && LoadMI->canFoldAsLoad() &&
-        LoadMI->mayLoad() && LocalMIs.count(LoadMI))
+        LoadMI->mayLoad() && LocalMIs.count(LoadMI) &&
+        // No store between LoadMI and FlagProducer that could change the value.
```
RKSimon: move the comment above the condition chain instead of embedded within it?

lonemeow:
If you want a slightly smaller and even more obviously-broken test case:
```llvm
  define i64 @demo(ptr %p, i64 %fallback) {
    %v = load i64, ptr %p
    store i64 -1, ptr %p
    %cnt = tail call i64 @llvm.ctpop.i64(i64 %v)
    %iszero = icmp eq i64 %v, 0
    %res = select i1 %iszero, i64 %fallback, i64 %cnt
    ret i64 %res
  }
```
Without the fix this compiles to:
```as
  demo:
      movq    $-1, (%rdi)
      popcntq (%rdi), %rax        ; load folded into popcnt, past the store
      cmoveq  %rsi, %rax
      retq
```

RKSimon (approved):
LGTM

PR #198240, opened 2026-05-18
Fix MAKE_DWP on Linux: set DSYM so the DWP packaging step runs

ionuthristodorescu:
On non-Darwin platforms, `MAKE_DWP=YES` sets `DWP_NAME` but never sets `DSYM`. Since the `DWP` packaging step is inside the `$(DSYM)` recipe and the `all` target depends on `$(DSYM)`, the step is skipped when `DSYM` is empty. Set `DSYM = $(DWP_NAME)` so the recipe is invoked. I understand "DSYM" sounds like "dSYM" which is a Darwin concept - this is a possible fix to make `self.build(debug_info="dwp")` work (without having to manually call `llvm-dwp` after `debug_info="dwo"`, for example).

cc: @clayborg @JDevlieghere @jimingham

### DIFF
--- a/lldb/packages/Python/lldbsuite/test/make/Makefile.rules
+++ b/lldb/packages/Python/lldbsuite/test/make/Makefile.rules
@@ -693,6 +693,9 @@ include $(wildcard $(PREREQS) $(DYLIB_PREREQS))
 .PHONY: clean
 dsym:	$(DSYM)
 all:	$(EXE) $(DSYM)
+ifeq "$(MAKE_DWP)" "YES"
+all:	$(DWP_NAME)
+endif
 clean::
 ifeq "$(findstring lldb-test-build.noindex, $(BUILDDIR))" ""
 	$(error Trying to invoke the clean rule, but not using the default build tree layout)
--- a/lldb/packages/Python/lldbsuite/test/make/Makefile.rules
+++ b/lldb/packages/Python/lldbsuite/test/make/Makefile.rules
@@ -564,28 +564,37 @@ ifneq "$(CODESIGN)" ""
 endif
 endif
 
-#----------------------------------------------------------------------
-# Make the dSYM file from the executable if $(MAKE_DSYM) != "NO"
-#----------------------------------------------------------------------
 $(DSYM) : $(EXE)
 ifeq "$(OS)" "Darwin"
 ifneq "$(MAKE_DSYM)" "NO"
 	"$(DS)" $(DSFLAGS) -o "$(DSYM)" "$(EXE)"
 else
 endif
-else
+endif
+
 ifeq "$(SPLIT_DEBUG_SYMBOLS)" "YES"
+ifeq "$(DBG_SYMBOL_FILE)" ""
+        DBG_SYMBOL_FILE = $(EXE).debug
+endif
 ifeq "$(SAVE_FULL_DEBUG_BINARY)" "YES"
 	cp "$(EXE)" "$(EXE).unstripped"
 endif
-	$(OBJCOPY) --only-keep-debug "$(EXE)" "$(DSYM)"
-	$(OBJCOPY) --strip-debug --add-gnu-debuglink="$(DSYM)" "$(EXE)" "$(EXE)"
+	$(OBJCOPY) --only-keep-debug "$(EXE)" "$(DBG_SYMBOL_FILE)"
+	$(OBJCOPY) --strip-debug --add-gnu-debuglink="$(DBG_SYMBOL_FILE)" "$(EXE)" "$(EXE)"
 endif
+
 ifeq "$(MAKE_DWP)" "YES"
-	$(DWP) -o "$(DWP_NAME)" $(DWOS)
+ifeq "$(DWP_NAME)" ""
+ifeq "$(DBG_SYMBOL_FILE)" ""
+            DWP_NAME = $(EXE).dwp
+else
+            DWP_NAME = $(DBG_SYMBOL_FILE).dwp
 endif
 endif
 
+$(DWP_NAME): $(EXE)
+	$(DWP) -o "$(DWP_NAME)" $(DWOS)
+endif
 
 #----------------------------------------------------------------------
 # Make the dylib
--- a/lldb/packages/Python/lldbsuite/test/make/Makefile.rules
+++ b/lldb/packages/Python/lldbsuite/test/make/Makefile.rules
@@ -218,6 +218,7 @@ else
 		MAKE_DWO := YES
 		DWP_NAME = $(EXE).dwp
 		DYLIB_DWP_NAME = $(DYLIB_NAME).dwp
+		DSYM = $(DWP_NAME)
 	endif
 endif
 


bulbazord (commented):
>  I understand "DSYM" sounds like "dSYM" which is a Darwin concept - this is a possible fix to make `self.build(debug_info="dwp")` work (without having to manually call `llvm-dwp` after `debug_info="dwo"`, for example).

For most of the Makefile rules and usage of the string "dsym", it really does mean "dSYM". I wasn't there when a lot of this logic was written but it wouldn't shock me if folks just piggybacked onto existing dSYM logic to generate dwp files.

This happens because of the `$(DSYM) : $(EXE)` rule that happens later on right? We should really find a better name for this thing. Maybe I'm biased but DSYM being involved in generating anything other than a `dSYM` doesn't make a lot of sense.

ionuthristodorescu:
> > I understand "DSYM" sounds like "dSYM" which is a Darwin concept - this is a possible fix to make `self.build(debug_info="dwp")` work (without having to manually call `llvm-dwp` after `debug_info="dwo"`, for example).
> 
> For most of the Makefile rules and usage of the string "dsym", it really does mean "dSYM". I wasn't there when a lot of this logic was written but it wouldn't shock me if folks just piggybacked onto existing dSYM logic to generate dwp files.
> 
> This happens because of the `$(DSYM) : $(EXE)` rule that happens later on right? We should really find a better name for this thing. Maybe I'm biased but DSYM being involved in generating anything other than a `dSYM` doesn't make a lot of sense.

So instead of this fix, should I add another target / target dep that is more OS-independent / dwp-related ?

clayborg:
We should make DSYM and DWP support totally separate and not have DWP support rely on anything DSYM related IMHO. Not sure what it will take to make this happen, but we shouldn't need to set any setting related to DSYM for DWP.

bulbazord:
> So instead of this fix, should I add another target / target dep that is more OS-independent / dwp-related ?

That's roughly what I'm suggesting, but I think it's more about making the DWP code path separate from the dSYM code path. If you wanted them to share any logic, you could do `GENERATE_SEPARATE_DEBUG_INFO` which does the right thing on different platforms (e.g. dSYM on Darwin, DWP on whatever platforms choose to use it). The key here is you shouldn't have to touch dSYM logic to generate a DWP.

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
```diff
@@ -563,28 +564,37 @@ ifneq "$(CODESIGN)" ""
...
 endif
 
-#----------------------------------------------------------------------
-# Make the dSYM file from the executable if $(MAKE_DSYM) != "NO"
-#----------------------------------------------------------------------
 $(DSYM) : $(EXE)
 ifeq "$(OS)" "Darwin"
 ifneq "$(MAKE_DSYM)" "NO"
 	"$(DS)" $(DSFLAGS) -o "$(DSYM)" "$(EXE)"
 else
```
bulbazord, line 571: Can you remove this stray `else` while you're here?

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
```diff
@@ -218,6 +218,7 @@ else
 		MAKE_DWO := YES
 		DWP_NAME = $(EXE).dwp
 		DYLIB_DWP_NAME = $(DYLIB_NAME).dwp
+		DSYM = $(DWP_NAME)
```
bulbazord, line 221: This shouldn't be needed anymore, right?

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
bulbazord, line 214: Similarly, this block should go away.

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
```diff
@@ -563,28 +564,37 @@ ifneq "$(CODESIGN)" ""
...
 ifneq "$(MAKE_DSYM)" "NO"
 	"$(DS)" $(DSFLAGS) -o "$(DSYM)" "$(EXE)"
 else
 endif
-else
+endif
+
 ifeq "$(SPLIT_DEBUG_SYMBOLS)" "YES"
+ifeq "$(DBG_SYMBOL_FILE)" ""
+        DBG_SYMBOL_FILE = $(EXE).debug
```
bulbazord, line 577: This appears over-indented?

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
```diff
@@ -563,28 +564,37 @@ ifneq "$(CODESIGN)" ""
...
 $(DSYM) : $(EXE)
 ifeq "$(OS)" "Darwin"
 ifneq "$(MAKE_DSYM)" "NO"
 	"$(DS)" $(DSFLAGS) -o "$(DSYM)" "$(EXE)"
 else
 endif
-else
+endif
+
 ifeq "$(SPLIT_DEBUG_SYMBOLS)" "YES"
```
bulbazord, line 575: Is this still attached to the DSYM target? Shouldn't this be under the DWP_NAME target now?

bulbazord (commented):
Better, still have some questions. The logic still seems kind of tangled and there are still dwp/dsym references that should go away.

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
clayborg: This is top level now. Just setting some settings so we get things right below

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
clayborg: Line 576 is under indented right?

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
```diff
@@ -563,28 +564,37 @@ ifneq "$(CODESIGN)" ""
...
+ifeq "$(DBG_SYMBOL_FILE)" ""
+        DBG_SYMBOL_FILE = $(EXE).debug
+endif
 ifeq "$(SAVE_FULL_DEBUG_BINARY)" "YES"
 	cp "$(EXE)" "$(EXE).unstripped"
 endif
```
clayborg, line 581: These should all be indented as well? I am not sure how indentation is supposed to look in Makefiles, but I am assuing they should be indented.

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
```diff
@@ -683,6 +693,9 @@ include $(wildcard $(PREREQS) $(DYLIB_PREREQS))
 .PHONY: clean
 dsym:	$(DSYM)
 all:	$(EXE) $(DSYM)
+ifeq "$(MAKE_DWP)" "YES"
+all:	$(DWP_NAME)
```
clayborg, line 697: Shouldn't this be 
```
dwp:    $(DWP_NAME)
all:	$(EXE) $(DWP_NAME)
```

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
clayborg, line 212: From AI about Makefile indentation:

You should not indent ifeq, ifneq, ifdef, or ifndef statements with a tab. 

In a Makefile, lines starting with a tab are interpreted as shell commands (part of a recipe). Since these conditional directives are part of the Make syntax itself and not shell commands, indenting them with a tab will cause Make to pass them to the shell, resulting in a syntax error. 

Rules for Indenting Conditionals
- Avoid Tabs: Never start a Make conditional directive (ifeq, else, endif) with a tab.
- Optional Spaces: You may use spaces for logical indentation to make nested conditionals more readable.
- Global Context: Conditionals are often written starting at the very beginning of the line (column zero) to avoid confusion.
- Inside Recipes: If you use a conditional inside a target's recipe, it must still follow Make's syntax. Most developers avoid this by using shell-level if statements instead, or by ensuring the Make conditional is not tab-indented.

clayborg: So we should indent with spaces

clayborg: This should be:

```
DBG_SYMBOL_FILE = $(EXE).debug
```

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
```diff
@@ -218,6 +218,7 @@ else
 		MAKE_DWO := YES
 		DWP_NAME = $(EXE).dwp
```
clayborg, line 219: This needs to check if we have a DBG_SYMBOL_FILE value.
```
ifeq "$(DBG_SYMBOL_FILE)" ""
    DWP_NAME = $(EXE).dwp
else:
    DWP_NAME = $(DBG_SYMBOL_FILE).dwp
endif
```

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
clayborg: yes, remove this

lldb/packages/Python/lldbsuite/test/make/Makefile.rules
clayborg, line 577: This was done above in my suggestion and can/should be removed.

clayborg (changes requested):

PR #198245, opened 2026-05-18
[AMDGPU] Clamp load_monitor scope to minimum SCOPE_SE

ssahasra:
The load_monitor instructions monitor L2 cache lines and therefore require at least SCOPE_SE to ensure the L2 cache is hit. The current memory model requires the user to ensure that the specified scope is such that it results in at least SCOPE_SE, otherwise the behaviour is undefined. Instead, we now clamp the emitted scope at a minimum of SCOPE_SE, so that the undefined behaviour is converted into a performance loss instead.

Assisted-By: Claude Opus 4.6


llvm/lib/Target/AMDGPU/SIMemoryLegalizer.cpp
```diff
@@ -2278,6 +2295,14 @@ bool SIGfx12CacheControl::setAtomicScope(const MachineBasicBlock::iterator &MI,
 
   // Other address spaces do not have a cache.
 
+  // load_monitor instructions need at least SCOPE_SE to ensure L2 is hit.
```
Pierre-vh: This doesn't feel like the right place to do it. I think we should have a `finalizeLoad` function for this kind of fix.

llvm/lib/Target/AMDGPU/SIMemoryLegalizer.cpp
```diff
@@ -2227,6 +2227,23 @@ bool SIGfx12CacheControl::finalizeStore(MachineInstr &MI, bool Atomic) const {
   return Changed;
 }
 
+static bool isLoadMonitorInstr(const MachineInstr &MI) {
```
Pierre-vh: I'd just put that in InstrInfo as a static method.

llvm/docs/AMDGPUUsage.rst
```diff
@@ -1779,7 +1779,9 @@ The AMDGPU backend implements the following LLVM IR intrinsics.
                                                    * Flat pointer.
                                                    * :ref:`Load Atomic Ordering<amdgpu-intrinsics-c-abi-atomic-memory-ordering-operand>`.
                                                    * :ref:`Synchronization Scope<amdgpu-intrinsics-syncscope-metadata-operand>`.
-                                                     Note that the scope used must ensure that the L2 cache will be hit.
+
```
Pierre-vh, line 1881: Don't document it here, simply document it in the GFX12.5 memory model. There is already a paragraph that talks about it:

```
  * In order to monitor a cache line in the L2 cache, these instructions must
    ensure that the L2 cache is always hit by setting the ``SCOPE`` of the instruction
    appropriately.
```

It just needs another sentence to say "the compiler may adjust the scope of the instruction accordingly to ensure this is the case."

Pierre-vh (commented):
That's fine, though usually for these intrinsics I thought we preferred if they mapped 1-1 to insts? This is why I didn't do this earlier.
Is there a specific reason why we need this now?

ssahasra:
> That's fine, though usually for these intrinsics I thought we preferred if they mapped 1-1 to insts? This is why I didn't do this earlier.

That interpretation is kinda incomplete. There are two kinds of intrinsics: some are specifically designed to reflect the underlying instruction, and some are meant to provide a nice programming model. Personally, I find it doubtful whey we need the first kind, but they do exist. It would have been much nicer if this intrinsic had originally started out with a proper "modelled as an atomic" sort of programming model.

> Is there a specific reason why we need this now?

Nothing critical. Its presence bugged me when I stumbled across a different problem: Clang translates the integer scope on the corresponding builtin as simple typecast without a ScopeModel: #198250. This works simply because the intrinsic happens to be UB for scopes where this typecast is wrong. That lack of ScopeModel will happen to work even when this PR is committed ... it's just something that itched really badly for me.

Pierre-vh:
> That interpretation is kinda incomplete. There are two kinds of intrinsics: some are specifically designed to reflect the underlying instruction, and some are meant to provide a nice programming model. Personally, I find it doubtful whey we need the first kind, but they do exist. It would have been much nicer if this intrinsic had originally started out with a proper "modelled as an atomic" sort of programming model.

Yeah that's fair. I wish we were more consistent with it. That ship has sailed, but I wish we had separate namespaces for the "direct/raw" intrinsics and another for the user-friendly ones. (e.g. one starts with `amdgcn.raw` and the other is just `amdgcn`)

> This works simply because the intrinsic happens to be UB for scopes where this typecast is wrong.

To be clear, the UB here is just that the monitoring may not happen. That's also clearly documented in the memory model, so it's not an unknown, just another easy-to-miss foot shotgun.

llvm/docs/AMDGPUUsage.rst
t-tye: So what are the semantics of this functionality? Is the semantic that the monitor wait should be released when there is a store to the location at a specified scope? If the hardware only functions when the store is at least at SE scope then what is the definition when stores are <SE? Making the monitored load forced to be SE scope solves one half of the communication, but not the other half of the reads. Or am I missing something?

Or is the write side relying on the fact that all caches <SE are write through caches so they will "in bounded finite time" reach the L2 and so trigger the monitored wait?

llvm/lib/Target/AMDGPU/SIMemoryLegalizer.cpp
```diff
@@ -2278,6 +2295,14 @@ bool SIGfx12CacheControl::setAtomicScope(const MachineBasicBlock::iterator &MI,
 
   // Other address spaces do not have a cache.
 
+  // load_monitor instructions need at least SCOPE_SE to ensure L2 is hit.
+  if (isLoadMonitorInstr(*MI)) {
+    MachineOperand *CPol = TII->getNamedOperand(*MI, AMDGPU::OpName::cpol);
+    if (CPol &&
```
arsenm: I'm assuming this operand always has to exist for these instructions

ssahasra:
> To be clear, the UB here is just that the monitoring may not happen. That's also clearly documented in the memory model, so it's not an unknown, just another easy-to-miss foot shotgun.

Presumably the wave executed a matching sleep after it performed this load-monitor. So maybe theoretically there is a situation where the wave is permanently sleeping. It's on the same level as saying that a wave *may* permanently loop if an atomic load that was supposed to fetch fresh data from L2 returned from L1 instead.

llvm/docs/AMDGPUUsage.rst
ssahasra: > If the hardware only functions when the store is at least at SE scope then what is the definition when stores are <SE?

There is no link to the GFX1250 SPG yet, but it says "the user is responsible". I interpret that as "undefined behaviour".

> So what are the semantics of this functionality? Is the semantic that the monitor wait should be released when there is a store to the location at a specified scope?

At a "pure semantics" level, yes that should be how we look at it. The SPG of course documents it using specific implementation details like "notify when the cacheline is changed in L2".

> Making the monitored load forced to be SE scope solves one half of the communication, but not the other half of the reads. Or am I missing something?

On the load side, I am proposing that from a programming model point of view, there is no need to say that certain scopes are undefined. The fact that L2 must be reached via SCOPE_SE is an implementation detail. In LLVM IR, it should be perfectly valid to call `load.monitor(ptr, !"workgroup")`. There is nothing wrong if the programmer wants to monitor at the workgroup scope. It could have been lowered to SCOPE_CU, but current hardware just happens to require SCOPE_SE.

> Or is the write side relying on the fact that all caches <SE are write through caches so they will "in bounded finite time" reach the L2 and so trigger the monitored wait?

The way I see it, the `load_monitor` as well as the `monitor_sleep` include a `MakeVisible` semantics at the SE scope. The sleep is released when a new store becomes *visible* or *available* at the specified scope. How the other side establishes that is up to the programmer, but there is an assumption that it eventually does happen.

Strangely the C++ standard says "should" rather than "must" or "shall" in this context:
https://eel.is/c++draft/intro.progress#17

> An **implementation should ensure** that the last value (in modification order) assigned by an atomic or synchronization operation will become visible to all other threads in a finite period of time.

ssahasra: > The way I see it, the `load_monitor` as well as the `monitor_sleep` include a `MakeVisible` semantics at the SE scope.

Actually, the `monitor_sleep` does not have `MakeVisible` semantics. It can *check and report* that a new value *may* be visible, but it cannot pull that new value into smaller scopes. The wave still has to execute another `MakeVisible` operation to actually see that value. In most cases, this will be the `load_monitor` from the next iteration of the same loop.

ssahasra: Moved the documentation.

llvm/lib/Target/AMDGPU/SIMemoryLegalizer.cpp
ssahasra: Fixed.

llvm/lib/Target/AMDGPU/SIMemoryLegalizer.cpp
ssahasra: Fixed.

llvm/lib/Target/AMDGPU/SIMemoryLegalizer.cpp
ssahasra: Yes. Added an assert for that.

ssahasra:
Bump!

ssahasra:
I would like to submit this change if there is no explicit opposition to it. @Pierre-vh @t-tye

Pierre-vh (approved):

PR #198248, opened 2026-05-18
[Clang] Accept gnu vectors in __builtin_masked*

philnik777:
There doesn't seem to be a good reason to reject gnu vectors in these builtins. The error messages for that case are also rather bizzare, which doesn't help with figuring out what's going wrong.

### DIFF
--- a/clang/lib/Sema/SemaChecking.cpp
+++ b/clang/lib/Sema/SemaChecking.cpp
@@ -2650,12 +2650,19 @@ static ExprResult BuiltinMaskedStore(Sema &S, CallExpr *TheCall) {
         S.Diag(ValArg->getExprLoc(), diag::err_vec_masked_load_store_ptr)
         << 2 << "vector");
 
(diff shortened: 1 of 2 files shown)


jhuber6 (commented):
This was somewhat intentional, because GNU vectors do not support boolean masks. I found it really strange to be mixing vector types here, because one of them will need to be an extended vector. I do agree the error message could be pushed into Sema, however. Is there a strong desire for this to work with GNU vectors as the argument but not the mask portion?

philnik777:
Personally I don't have a strong desire, since I mostly target Clang, but I don't see a reason not to allow it. I feel like this could definitely come up when you special-case something for Clang but most of your code should also work with GCC. Having to do an explicit cast to seemingly pretty much the same thing is just very weird IMO.

jhuber6 (approved):
I suppose if it works there's no reason not to.

PR #198250, opened 2026-05-18
[AMDGPU][Clang] use ScopeModel to translate integer scopes [NFC]

ssahasra:
The assumption here is that AMDGPU builtins (typically suffixed with `__builtin_amdgcn`) use the `__MEMORY_SCOPE_*` enumeration, and not the `__HIP_MEMORY_SCOPE_*` enumeration (which is how it should be).

Assisted-By: Claude Opus 4.6

### DIFF
--- a/clang/lib/CodeGen/TargetBuiltins/AMDGPU.cpp
+++ b/clang/lib/CodeGen/TargetBuiltins/AMDGPU.cpp
@@ -383,6 +383,21 @@ static llvm::AtomicOrdering mapCABIAtomicOrdering(unsigned AO) {
   llvm_unreachable("Unknown AtomicOrderingCABI enum");
 }
 
+/// Convert a __MEMORY_SCOPE_* integer constant to a metadata node containing
+/// the target-specific sync scope string.
+static llvm::MetadataAsValue *emitScopeMD(
+    CodeGenFunction &CGF, unsigned ScopeInt,
+    llvm::AtomicOrdering AO = llvm::AtomicOrdering::SequentiallyConsistent) {
+  AtomicScopeGenericModel ScopeModel;
+  clang::SyncScope Scope = ScopeModel.map(ScopeInt);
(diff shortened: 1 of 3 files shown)


ssahasra:
This change can now be reviewed independently from #198245

ssahasra:
This is now a general refactoring that replaces special handling of integer scopes with a call to `getLLVMSyncScopeStr()`. This also eliminates a switch statement that was designed just for SPIRV on AMDGPU.

clang/lib/CodeGen/TargetBuiltins/AMDGPU.cpp
```diff
@@ -383,6 +383,30 @@ static llvm::AtomicOrdering mapCABIAtomicOrdering(unsigned AO) {
   llvm_unreachable("Unknown AtomicOrderingCABI enum");
 }
 
+// Map a __MEMORY_SCOPE_* integer constant to the AMDGPU-specific syncscope.
+// Invalid scope values are mapped to system scope (empty string).
```
shiltian, line 387: I understand this is the current behavior, though I wonder whether we should error out instead,

shiltian (approved):

clang/lib/CodeGen/TargetBuiltins/AMDGPU.cpp
ssahasra: Maybe, but perhaps there is a better place to error out, such as the AMDGPU verifier. This function being refactored here might not occur on all paths. It's just providing a reasonable default within its area of influence that could be changed to unreachable if there was a verifier.

shiltian: But the verifier doesn't see the value right?

ssahasra: I really haven't thought about it much. It's just a refactor that establishes a consistent handling of scopes for now.

PR #198251, opened 2026-05-18
[Analysis] Use TTI to guard dereferenceability through addrspacecast

jszt1:
isDereferenceableAndAlignedPointer looked through addrspacecast unconditionally. This is incorrect when cast is not a no-op. This change safeguards addrspacecast look-through with TTI::isNoopAddrSpaceCast when TTI is available. Without TTI, preserve prior behavior.


fhahn (commented):
Hmm, relying on TTI for this seems a bit fragile. Is there an alternative to checking that? Looking at `llvm/lib/Transforms`, the only user of the hook is `InferAddressSpaces`

jszt1:
@fhahn 
TTI->isNoopAddrSpaceCast delegates through BasicTTIImpl to TargetMachine::isNoopAddrSpaceCast, which is backend overriden check. By using TTI here we respect the target legality - the included test verifies that SimplifyCFG doesn't speculate loads through non-noop addrspacecasts.

I don't think there is a DataLayout alternative that captures the noop/non-noop relationship between address spaces.

llvm/test/Transforms/SimplifyCFG/speculate-addrspacecast-load.ll
```diff
@@ -0,0 +1,50 @@
...
+; CHECK-NEXT:    [[TMP3:%.*]] = addrspacecast ptr addrspace(4) [[P]] to ptr addrspace(1)
+; CHECK-NEXT:    [[LOAD2:%.*]] = load i32, ptr addrspace(1) [[TMP3]], align 1
+; CHECK-NEXT:    br label %[[EXIT]]
+; CHECK:       [[EXIT]]:
+; CHECK-NEXT:    [[RESULT:%.*]] = phi i32 [ [[LOAD1]], %[[THEN]] ], [ [[LOAD2]], %[[ELSE]] ]
+; CHECK-NEXT:    [[ADD:%.*]] = add nsw i32 [[RESULT]], 1
+; CHECK-NEXT:    ret i32 [[ADD]]
+;
+entry:
+  %0 = ptrtoint ptr addrspace(4) %p to i64
```
michalpaszkowski: Please name all vregs (`%0%`, `%1`, ...) for consistency and easier bulk renaming in the future.

llvm/test/Transforms/SimplifyCFG/speculate-addrspacecast-load.ll
```diff
@@ -0,0 +1,50 @@
...
+; RUN: opt -S -passes=simplifycfg < %s | FileCheck %s
+
+; Verify SimplifyCFG does not speculate loads through non-noop addrspacecasts.
+
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-n8:16:32:64-G1"
+
+define i32 @no_speculate_addrspacecast_load(ptr addrspace(4) noalias nocapture readonly dereferenceable(4) %p) {
+; CHECK-LABEL: define i32 @no_speculate_addrspacecast_load(
+; CHECK-SAME: ptr addrspace(4) noalias readonly captures(none) dereferenceable(4) [[P:%.*]]) {
+; CHECK-NEXT:  [[ENTRY:.*:]]
```
michalpaszkowski: Could we make these checks less strict/fragile? I think the important thing this test should verify is that the loads are not speculated/hoisted before the branch, not the exact instruction-by-instruction shape. For example, instead of checking every entry-block instruction with `CHECK-NEXT`, we could use `CHECK` to find the branch, then `CHECK-NOT` to make sure no load appears before the then/else blocks.

michalpaszkowski:
> Hmm, relying on TTI for this seems a bit fragile. Is there an alternative to checking that? Looking at `llvm/lib/Transforms`, the only user of the hook is `InferAddressSpaces`

@fhahn Good point, but I guess the only alternative would be to make this conservative and stop looking through addrspacecast entirely. That would obviously change the current behavior for safe truly no-op casts. Do you have any other alternative in mind? Or preference for the solution?

llvm/test/Transforms/SimplifyCFG/speculate-addrspacecast-load.ll
jszt1: Yes, it makes sense to simplify this tests. I have updated it.

michalpaszkowski:
Ping @fhahn 

Also tagging @nikic since this is somewhat related to the opaque pointers migration. Please let us know if you have a preference for the solution.

llvm/lib/Analysis/Loads.cpp
```diff
@@ -198,12 +199,18 @@ static bool isDereferenceableAndAlignedPointer(
...
-    return isDereferenceableAndAlignedPointer(ASC->getOperand(0), Alignment,
-                                              Size, DL, CtxI, AC, DT, TLI,
-                                              Visited, MaxDepth);
+                                              TLI, TTI, Visited, MaxDepth);
+
+  if (const AddrSpaceCastOperator *ASC = dyn_cast<AddrSpaceCastOperator>(V)) {
+    unsigned SrcAS = ASC->getSrcAddressSpace();
+    unsigned DstAS = ASC->getDestAddressSpace();
+    // When TTI is available look through if the cast is a no-op.
+    if (SrcAS == DstAS || !TTI || TTI->isNoopAddrSpaceCast(SrcAS, DstAS))
```
nikic, line 208: Why this be isValidAddrSpaceCast()? We don't specify what exactly "valid" is supposed to mean here, but I assume it includes preserving dereferenceability...

nikic, line 208: The way this is implemented means that with `!TTI` we get incorrect behavior. If TTI is not present we should be conservatively assuming that addrspace is never dereferenceable.

llvm/test/Transforms/SimplifyCFG/speculate-addrspacecast-load.ll
```diff
@@ -0,0 +1,34 @@
+; RUN: opt -S -passes=simplifycfg < %s | FileCheck %s
```
nikic, line 1: update_test_checks.py

nikic:
Not sure what relation this has to opaque pointers.

I definitely don't like the use of TTI here. This is not the first time the issue has come up, and the only real solution is to move information on address space validity to DataLayout.

michalpaszkowski:
> Not sure what relation this has to opaque pointers.
> 
> I definitely don't like the use of TTI here. This is not the first time the issue has come up, and the only real solution is to move information on address space validity to DataLayout.

@nikic Please take a look at https://github.com/llvm/llvm-project/pull/210888 This should solve the issue in this PR.

PR #149047, opened 2025-07-16
[VectorCombine] Remove dead node immediately in VectorCombine

davemgreen:
The vector combiner will process all instructions as it first loops through the function, adding any newly added and deleted instructions to a worklist which is then processed when all nodes are done. These leaves extra uses in the graph as the initial processing is performed, leading to sub-optimal decisions being made for other combines. This changes it so that trivially dead instructions are removed immediately. The main change that this requires is to make sure iterator invalidation does not occur.

PR #198252, opened 2026-05-18
[LV] Implement integer part of VPDerivedIV cost model

david-arm:
Adds support for InductionDescriptor::IK_IntInduction variants of inductions in VPDerivedIVRecipe::computeCost. Designing a cost model for this recipe is tricky because it gets expanded in expandVPDerivedIV into smaller individual recipes that will then be folded by simplifyRecipe. Effectively the cost model has to guess how these will be folded otherwise the costs will be too pessimistic. For example, 'add i32 %x, 0' and 'mul i32 %x, 1' should both be treated as zero cost since all uses will be replaced with '%x'. Similarly,
'mul i32 %x, 4' will be converted to 'shl i32 %x, 2', and 'mul i32 %x, -1' will be simplified to 'sub i32 0, %x'.

As can be seen by this PR, it's important to have non-zero costs for cases where VPDerivedIVRecipe recipes will end up with non-foldable instructions. Some of the tests now favour wider VFs to cover the non-zero cost, or in other cases we prefer not to vectorise at all. I've added loop attributes in some places to maintain the previous vectorisation behaviour so that we're still matching the intent of the test.

I built the LLVM test suite and I couldn't find any cases where we use more than the first lane of the result so for now I maintained the original behaviour and returned 0. The other induction types (pointers, floating point values) are also left as returning 0 for now. These will be handled in future PRs.


llvm/test/Transforms/PhaseOrdering/AArch64/interleave_vec.ll
david-arm, line 1: According to https://github.com/llvm/llvm-project/pull/153039 this loop wasn't worth vectorising and previously we favoured vectorisation, whereas when we properly account for all the recipes we now abandon vectorisation. I verified locally on a neoverse-v2 machine that this change should not lead to a performance regression. Specifically, we now add the cost of doing `[[OFFSET_IDX:%.*]] = mul i64 [[INDEX]], 5` in the loop.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2620,6 +2641,84 @@ bool VPWidenIntOrFpInductionRecipe::isCanonical() const {
...
+          NeedsMul = false;
+          NeedsAdd = true;
+        } else if (StepC->getValue().isPowerOf2()) {
+          // This will most likely end up as a shift-left in simplifyRecipe
+          NeedsMul = false;
+          Cost += Ctx.TTI.getArithmeticInstrCost(
+              Instruction::Shl, StepTy, Ctx.CostKind,
+              {TargetTransformInfo::OK_AnyValue, TargetTransformInfo::OP_None},
+              {TargetTransformInfo::OK_UniformConstantValue,
+               TargetTransformInfo::OP_PowerOf2});
```
paulwalker-arm: Are these necessary? The `OP_PowerOf2` looks wrong?. `StepC` might be a power-of-two but when converted to a shift amount it won't be?

To me it looks like the intent is for these to be passed to `getArithmeticInstrCost(Instruction::Mul,...` so that it can return a cost based on the assumption a power-of-two constant operand will be lowered to a shift?

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2620,6 +2641,84 @@ bool VPWidenIntOrFpInductionRecipe::isCanonical() const {
...
+          // This will most likely end up as a negate in simplifyRecipe, and
+          // the negate will be combined with the add to make a sub.
+          // NOTE: This is perhaps an invalid assumption that the cost of an
+          // 'add' is the same as a 'sub'.
+          NeedsMul = false;
+          NeedsAdd = true;
+        } else if (StepC->getValue().isPowerOf2()) {
+          // This will most likely end up as a shift-left in simplifyRecipe
+          NeedsMul = false;
+          Cost += Ctx.TTI.getArithmeticInstrCost(
```
paulwalker-arm: Perhaps more consistent to have `NeedsShl` and then keep the `getArithmeticInstrCost` calls together?

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2620,6 +2641,84 @@ bool VPWidenIntOrFpInductionRecipe::isCanonical() const {
...
+    if ((NeedsAdd || NeedsMul) && StepTySize != IndexTySize) {
+      unsigned CastOpc =
+          StepTySize < IndexTySize ? Instruction::Trunc : Instruction::SExt;
+      Cost += Ctx.TTI.getCastInstrCost(
+          CastOpc, StepTy, IndexTy, TTI::CastContextHint::None, Ctx.CostKind);
+    }
+
+    if (NeedsAdd)
+      Cost += Ctx.TTI.getArithmeticInstrCost(Instruction::Add, StepTy,
+                                             Ctx.CostKind);
```
paulwalker-arm: Pedantic I know but this should be after `NeedsMul` to match the order of operations in `expandVPDerivedIV()`.

paulwalker-arm (approved):
I cannot really gauge the performance impact but the intent seems logical and the implementation is consistent with expandVPDerivedIV.

david-arm:
There was a buildbot failure after landing this PR (https://lab.llvm.org/buildbot/#/builders/193/builds/16567) which is unfortunate because the bug relates to a fragile test (that can change depending upon target rounding behaviour). I have a PR to fix the test here: https://github.com/llvm/llvm-project/pull/201580

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2797,6 +2797,86 @@ bool VPWidenIntOrFpInductionRecipe::isCanonical() const {
...
+        NeedsAdd = !StartC->isZero();
+
+    // For some values of step the arithmetic changes:
+    //  1. A step of 1 requires no operation.
+    //  2. A step of -1 requires a negate.
+    //  3. A power-of-2 step will use a shl, instead of a mul.
+    Type *StepTy = getStepValue()->getScalarType();
+    InstructionCost Cost(0);
+    if (auto *VPV = dyn_cast<VPIRValue>(getStepValue())) {
+      if (auto *StepC = dyn_cast<ConstantInt>(VPV->getValue())) {
```
fhahn, line 2831: could probably use pattern matching here to reduce the indentation else here a bit

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2797,6 +2797,86 @@ bool VPWidenIntOrFpInductionRecipe::isCanonical() const {
...
+    if (!vputils::onlyFirstLaneUsed(this))
+      break;
+
+    // Start off by assuming we need both mul and add, then refine this.
+    bool NeedsMul = true, NeedsAdd = true, NeedsShl = false;
+
+    // If the start value is zero the add gets folded away.
+    if (auto *VPV = dyn_cast<VPIRValue>(getStartValue()))
+      if (auto *StartC = dyn_cast<ConstantInt>(VPV->getValue()))
+        NeedsAdd = !StartC->isZero();
```
fhahn, line 2822: can use pattern matching to check if it is known-zero I think

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2797,6 +2797,86 @@ bool VPWidenIntOrFpInductionRecipe::isCanonical() const {
...
+        }
+      }
+    }
+
+    // Add the cost of the conversion from index to step type if the index
+    // will be used.
+    Type *IndexTy = getIndex()->getScalarType();
+    unsigned StepTySize = StepTy->getScalarSizeInBits();
+    unsigned IndexTySize = IndexTy->getScalarSizeInBits();
+    if ((NeedsAdd || NeedsMul || NeedsShl) && StepTySize != IndexTySize) {
```
fhahn, line 2854: Does this mean there are cases where no arithmetic operation is needed? If so, such redundant recipe should be folded

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2797,6 +2797,86 @@ bool VPWidenIntOrFpInductionRecipe::isCanonical() const {
...
+      unsigned CastOpc =
+          StepTySize < IndexTySize ? Instruction::Trunc : Instruction::SExt;
+      Cost += Ctx.TTI.getCastInstrCost(
+          CastOpc, StepTy, IndexTy, TTI::CastContextHint::None, Ctx.CostKind);
+    }
+
+    if (NeedsMul)
+      Cost += Ctx.TTI.getArithmeticInstrCost(Instruction::Mul, StepTy,
+                                             Ctx.CostKind);
+    if (NeedsShl)
```
fhahn, line 2864: those 2 are mutually exclusive, right? If so, can clarify with assert or move to an `else`

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2797,6 +2797,86 @@ bool VPWidenIntOrFpInductionRecipe::isCanonical() const {
...
+
+    // If the start value is zero the add gets folded away.
+    if (auto *VPV = dyn_cast<VPIRValue>(getStartValue()))
+      if (auto *StartC = dyn_cast<ConstantInt>(VPV->getValue()))
+        NeedsAdd = !StartC->isZero();
+
+    // For some values of step the arithmetic changes:
+    //  1. A step of 1 requires no operation.
+    //  2. A step of -1 requires a negate.
+    //  3. A power-of-2 step will use a shl, instead of a mul.
```
fhahn, line 2827: It is a bit unfortunate that we currently need to anticipate those simplifications manually here,  but short term there probably is not much else we could do

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2797,6 +2797,86 @@ bool VPWidenIntOrFpInductionRecipe::isCanonical() const {
...
+InstructionCost VPDerivedIVRecipe::computeCost(ElementCount VF,
+                                               VPCostContext &Ctx) const {
+  // The cost model for this is modelled on expandVPDerivedIV in
+  // VPlanTransforms.cpp. In order to avoid overly pessimistic costs that can
+  // negatively affect vectorization it takes into account any expected
+  // simplifications that happen in simplifyRecipe.
+  switch (getInductionKind()) {
+  default:
+    // TODO: Compute cost for remaining kinds.
+    break;
```
fhahn, line 2809: may be simpler/clearer to just return from the switch in all cases

```suggestion
    // TODO: Compute cost for remaining kinds.
    return 0;
```

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -2797,6 +2797,86 @@ bool VPWidenIntOrFpInductionRecipe::isCanonical() const {
...
+  case InductionDescriptor::IK_IntInduction: {
+    // There are currently no tests that expose a path where all lanes are
+    // used, so it's better to bail out for now.
+    if (!vputils::onlyFirstLaneUsed(this))
+      break;
```
fhahn, line 2814: I think this is not needed; DerivedIV inherently produces a single-scalar value, so we should always just count the scalar cost

vvereschaka:
Hi @david-arm ,

these changes also break the `Transforms/LoopVectorize/ARM/mve-reg-pressure-spills.ll` test on ARMv7 builder
https://lab.llvm.org/buildbot/#/builders/38/builds/9160

I tried to revert these changes locally and it fixed the test.

Would you take care of it?

david-arm:
> Hi @david-arm ,
> 
> these changes also break the `Transforms/LoopVectorize/ARM/mve-reg-pressure-spills.ll` test on ARMv7 builder https://lab.llvm.org/buildbot/#/builders/38/builds/9160
> 
> I tried to revert these changes locally and it fixed the test.
> 
> Would you take care of it?

I don't think it's caused by this patch, but instead caused by differences in rounding mode on different targets. I hit a similar issue in a different test and I have a fix for this in review: https://github.com/llvm/llvm-project/pull/201824. Hopefully I can land it today!

david-arm:
In fact, I'll just fix the test itself to use less digits of precision until PR #201824 lands. That should unblock the build immediately.

david-arm:
Fix is here: https://github.com/llvm/llvm-project/pull/202291

ElvinaYakubova:
Hi David,

This commit caused a 25% performance regression of TSVC s351 benchmarks we are tracking. Could you please have a look?

The reproducer is as follows (-O3 -ffast-math -mcpu=neoverse-v2):

```
#define iterations 100000
#define LEN_1D 32000
#define LEN_2D 256


typedef float real_t;


struct args_t {
  long t1;
  long t2;
};


__attribute__((aligned(64))) real_t a[LEN_1D];
__attribute__((aligned(64))) real_t b[LEN_1D];
__attribute__((aligned(64))) real_t c[LEN_1D];
__attribute__((aligned(64))) real_t d[LEN_1D];
__attribute__((aligned(64))) real_t e[LEN_1D];
__attribute__((aligned(64))) real_t aa[LEN_2D][LEN_2D];
__attribute__((aligned(64))) real_t bb[LEN_2D][LEN_2D];
__attribute__((aligned(64))) real_t cc[LEN_2D][LEN_2D];


real_t s351(struct args_t *func_args) {
  // loop rerolling / unrolled saxpy (TSVC S351)
  func_args->t1 = 0;


  real_t alpha = c[0];
  for (int nl = 0; nl < 8 * iterations; nl++) {
    for (int i = 0; i < LEN_1D; i += 5) {
      a[i] += alpha * b[i];
      a[i + 1] += alpha * b[i + 1];
      a[i + 2] += alpha * b[i + 2];
      a[i + 3] += alpha * b[i + 3];
      a[i + 4] += alpha * b[i + 4];
    }
    // Keep memory live across the outer loop (replaces dummy()).
    __asm__ volatile("" : : "r"(a), "r"(b), "r"(c), "r"(d), "r"(e), "r"(aa),
                              "r"(bb), "r"(cc) : "memory");
  }


  func_args->t2 = 0;
  return a[0] + b[0];
}
```

Before, the loop vectorizer generated a wide vector FMA (ldp + fmla v*.4s).
After that commit, LV rejects vectorization and keeps a scalar loop (ldur + scalar fmadd).

david-arm:
Hi @ElvinaYakubova, that's strange because it directly contradicts https://github.com/llvm/llvm-project/pull/153039 that said this wasn't worth vectorising. I can only assume that since PR #153039 landed we did vectorise the loop, but applied some transformations after loop vectorisation that made the performance better. I'd rather not revert this PR if possible because I believe that overall the cost model changes are an improvement. I'll see if there is a way to reduce the cost of some of the other operations in the loop.

david-arm:
It sounds like whatever transformation https://github.com/llvm/llvm-project/pull/149047 is doing in VectorCombine we need to be doing in the loop vectoriser, or at least recognising that the loop is less expensive than it appears.

david-arm:
The saxpy_5 loop also doesn't include the cost of the `Cost of 0 for VF 2: vp<%7> = SCALAR-STEPS vp<%6>, ir<5>, vp<%0>` recipe, which isn't necessarily free and I intend to fix in a follow-on PR. I think this test is extremely fragile because the cost for VF 2 is *very* close to the scalar loop cost. Any minor improvement or tweak in the vplan cost model will constantly throw move this below and above the threshold.

david-arm:
It looks like if we force vectorisation with VF=2 and disable vector combine, the saxpy_5 loop is still slightly faster than scalar. I think what really needs to happen is for the loop vectoriser to apply the same vplan transformation that takes place in VectorCombine.cpp. However, that might take a while to fix and won't solve the regression in the short term. So for now I've got a couple of options:

1. Sacrifice generic improvements to the vplan cost model and revert this PR, but potentially making other loops worse, or
2. Find a way to reduce the cost for VF=2. It does seem like a cost of 28 for the interleave groups are a bit high.

@fhahn any thoughts?

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
david-arm: Hi, if the patch doesn't get reverted I'll follow up with these review comments!

ElvinaYakubova:
Thanks for the explanation and for pointing to the PRs, this helps. 
I agree there's no need to revert this PR over this one benchmark - the cost model direction makes sense. If you can tweak it a bit so rerolled saxpy patterns like this still vectorize or being a little less pessimistic on step-5 derived IVs, that would be great for us.

david-arm:
Hi @ElvinaYakubova, thanks! What the saxpy test has shown is that the cost of interleaved memory ops with factor 5 is too high in general. I've been testing out some micro-benchmarks based off saxpy and other interleave-factor-5 loops. In general it looks like we should be vectorising even when the vector combine pass doesn't optimise the IR. I am testing a downstream patch. If I can't land a fix for this soon, I'll probably end up reverting the PR as I can't see any other obvious solution.

PR #198255, opened 2026-05-18
Update Unicode tables to Unicode 18.0.

cor3ntin:
We are currently on Unicode 15.0, this updates to Unicode 18. Unicode 18.0 is currently in Beta so some properties could change. However the name are stable and the probability of other property changing is very low.


clang/docs/ReleaseNotes.rst
```diff
@@ -260,6 +258,8 @@ Non-comprehensive list of changes in this release
 - ``typeid`` on references and pointers of ``final`` types no longer emits a
   vtable lookup at runtime.
 
+- Added support for Unicode 18.0
```
AaronBallman:
```suggestion
- Updated support for Unicode from 15.1 to 18.0.
```

clang/test/Lexer/unicode.c
```diff
@@ -52,6 +55,8 @@ extern int 𝛛; // expected-warning {{mathematical notation character <U+1D6DB>
 extern int ₉; // expected-error {{character <U+2089> not allowed at the start of an identifier}} \\
                  expected-warning {{declaration does not declare anything}}
 
+extern int a\N{PICKLE}; // expected-error {{character <U+1FADD> not allowed in an identifier}}
```
AaronBallman, line 58: They're taking away my ability to use pickles for identifiers? This is atrocious. When will the madness stop? ;-)

AaronBallman (approved):
Test failures are unrelated to the changes in the PR; I did have a suggestion for the release note but otherwise this LGTM. I did not thoroughly check all of the table entries though, in case someone is looking for help falling asleep.

clang/test/Lexer/unicode.c
cor3ntin: 🫝

shafik: I thought pickling was a python thing, good thing we are done w/ it

shafik (commented):
Do you have an automated way to generate all these ranges and will it be the same process next time? Like is there some script we can put someplace?

eisenwave:
It seems to me like the banner comments in `UnicodeNameMappingGenerator.cpp` should have been updated. They still mention Unicode `15.1.0`

PR #198259, opened 2026-05-18
[NVPTX] Fix illegal combineInsertEltToShuffle pattern

Wolfram70:
Adds a condition to the `isShuffleMaskLegal` method to prevent `combineInsertEltToShuffle` from creating an illegal `vector_shuffle` after type legalization which leads to a crash.

Context:
This is triggered when bitcasting a `v2f16` into a vector that type legalizes to a `v2i32`. This happens on architectures supporting packed `f32` operations (>= `sm_100`). In certain cases, this leads to `combineInsertEltToShuffle` creating a vector shuffle with `v4f16` for simplification. Since this happens after type legalization and `v4f16` is an illegal type, it leads to a crash.

### DIFF
--- a/llvm/lib/Target/NVPTX/NVPTXISelLowering.h
+++ b/llvm/lib/Target/NVPTX/NVPTXISelLowering.h
@@ -107,6 +107,10 @@ public:
   TargetLoweringBase::LegalizeTypeAction
   getPreferredVectorAction(MVT VT) const override;
 
+  bool isShuffleMaskLegal(ArrayRef<int>, EVT VT) const override {
+    return isTypeLegal(VT);
+  }
+
   // Get the degree of precision we want from 32-bit floating point division
   // operations.
   NVPTX::DivPrecisionLevel getDivF32Level(const MachineFunction &MF,
--- a/llvm/test/CodeGen/NVPTX/insert-vector-elt-bitcast-legalize.ll
+++ b/llvm/test/CodeGen/NVPTX/insert-vector-elt-bitcast-legalize.ll
@@ -1,0 +1,24 @@
+; RUN: llc < %s -mtriple=nvptx64 -mcpu=sm_100a -mattr=+ptx86 -o /dev/null
+; RUN: %if ptxas-sm_100a && ptxas-isa-8.6 %{ llc < %s -mtriple=nvptx64 -mcpu=sm_100a -mattr=+ptx86 | %ptxas-verify -arch=sm_100a %}
+
+; Verify that inserting a bitcasted v2f16 into a vector that type-legalizes to
+; v2i32 does not crash during DAG legalization on sm_100a+ (where v2i32 is a
+; legal type). The DAG combiner would previously create an illegal v4f16
+; vector_shuffle after type legalization.
+
+define half @test(i64 %idx, <2 x half> %input) {
+  %val = bitcast <2 x half> %input to <1 x i32>
+  %scalar = extractelement <1 x i32> %val, i64 0
+  %vec = insertelement <4 x i32> <i32 0, i32 poison, i32 poison, i32 poison>, i32 %scalar, i64 1
+
+  ; prevent simplification to trigger the illegal pattern
+  %cast = bitcast <4 x i32> %vec to <8 x half>
+  %elt0 = extractelement <8 x half> %cast, i64 0
+  %elt3 = extractelement <8 x half> %cast, i64 3
+  %big = insertelement <16 x half> zeroinitializer, half %elt0, i64 1
+  %big2 = insertelement <16 x half> %big, half %elt3, i64 %idx
+  %big3 = insertelement <16 x half> %big2, half 0xH0000, i64 0
+  %result = extractelement <16 x half> %big3, i64 %idx
+
+  ret half %result
+}


durga4github:
For context,
Can we also add a small description (to the commit-msg) about the particular case that triggered this issue?

Prince781:
Shouldn't this be fixed in DAGCombiner? It should not create illegal types after type legalization.

Artem-B:
> Shouldn't this be fixed in DAGCombiner? It should not create illegal types after type legalization.

​I think it may be the case of "all of the above". NVPTX presents rather non-uniform set of instruction/type combination capabilities and breaks implicit assumptions made in other parts of LLVM. When we find them, making those assumptions explicitly checked is worth it, IMO. That said, NVPTX lowering is also known to under-constrain LLVM, so when we see something leading LLVM codegen in the wrong direction, we should fix it. In this case my guess is that we may be missing an explicit set action to tell LLVM to expand/split an operation and prevent LLVM from assuming something that can't be done in PTX.

Wolfram70:
It seems like NVPTX previously didn’t have any legal vector types large enough to trigger the `combineInsertEltToShuffle` pattern, so we hadn’t run into this before. The pattern checks `isShuffleMaskLegal()` and bails out if it isn’t, but the default behavior appears to allow all shuffles unless the target explicitly overrides it, which seems to be the general approach taken by the other targets. This also leaves the door open to perhaps make `v4f16` legal with custom lowering in the future while fixing the crash for now.

Artem-B:
> the default behavior appears to allow all shuffles unless the target explicitly overrides it,

OK. Then my question is -- is the TV type the only factor that determines our ability to lower shufflemask. 

Alternatively, can we lower some vector shuffles even if the type is nominally is not legal. I.e. for v4f16 we could certainly lower it as two `prmt` instructions over a pair of 32-bit inputs.

I guess my question can be rephrased as "are we sure that restricting shuffles to legal type only does not unintentionally block an existing bespoke lowering?" If we do not have such special cases, perhaps it's an optimization opportunity for a future CL. I think this may be what you're alluding to above.

Wolfram70:
> Alternatively, can we lower some vector shuffles even if the type is nominally is not legal

If my understanding is right, we do have to register it as a legal type and provide a custom handler in `LowerVECTOR_SHUFFLE` like we do for `v4i8` today. In that scenario, `isShuffleMaskLegal` would return true anyways since the type is registered as legal. So, this shouldn't block any lowering path that wouldn't otherwise crash due to an illegal type, basically serving as a safety net.

> perhaps it's an optimization opportunity for a future CL

Yeah, I think we could potentially make `v4f16` legal and lower it using two `prmt` instructions. I'm just not sure how much work that would entail overall, since we'd also need to support the other vector operations.

Wolfram70:
@Artem-B Can we merge this fix for now?

Artem-B (approved):

ThomasRaoux:
this seems to cause some regressions with i8 types.

This case seems to miscompile now:
```
target datalayout = "e-p3:32:32-p4:32:32-p5:32:32-p6:32:32-p7:32:32-i64:64-i128:128-i256:256-v16:16-v32:32-n16:32:64"

define ptx_kernel void @_matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu(<2 x i8> %0) {
.lr.ph:
  %1 = tail call <2 x i8> asm "cvt.rn.satfinite.e4m3x2.f32  $0, $2, $1; \0A", "=h,r,r"(i32 0, i32 0)
  %2 = shufflevector <2 x i8> %0, <2 x i8> %1, <4 x i32> <i32 1, i32 3, i32 poison, i32 poison>
  %3 = insertelement <4 x i8> %2, i8 0, i64 0
  %4 = bitcast <4 x i8> %3 to i32
  %5 = insertelement <4 x i32> zeroinitializer, i32 %4, i64 0
  store <4 x i32> %5, ptr addrspace(3) null, align 16
  ret void
}
```

```
llc repro.ll -O3 \
  -mtriple=nvptx64-nvidia-cuda \
  -mcpu=sm_90a -mattr=+ptx88 \
  -nvptx-mad-wide-opt -nvptx-short-ptr \
  -relocation-model=pic -asm-verbose -o -
  ```
->

```
//
// Generated by LLVM NVPTX Back-End
//

.version 8.8
.target sm_90a
.address_size 64

	// .globl	_matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu // -- Begin function _matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu
                                        // @_matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu
.visible .entry _matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu(
	.param .align 2 .b8 _matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu_param_0[2]
)
{
	.reg .b16 	%rs<4>;
	.reg .b32 	%r<12>;

// %bb.0:                               // %.lr.ph
	ld.param.v2.b8 	{%rs2, %rs3}, [_matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu_param_0];
	mov.b32 	%r1, 0;
	// begin inline asm
	cvt.rn.satfinite.e4m3x2.f32  %rs1, %r1, %r1; 

	// end inline asm
	cvt.u32.u16 	%r2, %rs3;
	prmt.b32 	%r3, %r4, %r2, 0x3340U;
	prmt.b32 	%r5, %r6, %r7, 0x3340U;
	prmt.b32 	%r8, %r3, %r5, 0x5410U;
	cvt.u32.u16 	%r9, %rs1;
	prmt.b32 	%r10, %r8, %r9, 0x51U;
	bfi.b32 	%r11, 0, %r10, 0, 8;
	st.shared.v4.b32 	[%r1], {%r11, 0, 0, 0};
	ret;
                                        // -- End function
}
```

vs before:

```
//
// Generated by LLVM NVPTX Back-End
//

.version 8.8
.target sm_90a
.address_size 64

	// .globl	_matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu // -- Begin function _matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu
                                        // @_matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu
.visible .entry _matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu(
	.param .align 2 .b8 _matmul_NNN_mxfp4xbf16xbf16_128x256x64x1_dequantize_nvfp4_swiglu_param_0[2]
)
{
	.reg .b16 	%rs<2>;
	.reg .b32 	%r<4>;

// %bb.0:                               // %.lr.ph
	mov.b32 	%r1, 0;
	// begin inline asm
	cvt.rn.satfinite.e4m3x2.f32  %rs1, %r1, %r1; 

	// end inline asm
	cvt.u32.u16 	%r2, %rs1;
	prmt.b32 	%r3, %r2, 0, 0x14U;
	st.shared.v4.b32 	[%r1], {%r3, 0, 0, 0};
	ret;
                                        // -- End function
}
```

did anybody run into that?

Wolfram70:
Looks like this fix also blocks this pattern from triggering pre-type-legalization resulting in regressions in some types. Removed this in https://github.com/llvm/llvm-project/pull/205056 to fix this crash directly in the `combineInsertEltToShuffle` pattern.

PR #198262, opened 2026-05-18
[lldb] Do not include argv[0] to the process info arguments.

da-viper:
The process file name is stored separately from other command line args. and argv[0] may differ from the executable path if process was created using the `exec*()` variants.

### DIFF
--- a/lldb/source/Host/macosx/objcxx/Host.mm
+++ b/lldb/source/Host/macosx/objcxx/Host.mm
@@ -614,9 +614,14 @@ static bool GetMacOSXProcessArgs(const ProcessInstanceInfoMatch *match_info_ptr,
               break;
             ++offset;
           }
-          // Now extract all arguments
+
+          // Skip argv[0] as we already have the file name from the executable path.
+          if (argc > 0 )
+            data.GetCStr(&offset);
+
+          // Now extract the rest of the arguments.
           Args &proc_args = process_info.GetArguments();
-          for (int i = 0; i < static_cast<int>(argc); ++i) {
+          for (int i = 1; i < static_cast<int>(argc); ++i) {
             cstr = data.GetCStr(&offset);
             if (cstr)
               proc_args.AppendArgument(llvm::StringRef(cstr));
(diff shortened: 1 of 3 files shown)


lldb/source/Host/macosx/objcxx/Host.mm
```diff
@@ -614,9 +614,14 @@ DataExtractor data(arg_data.GetBytes(), arg_data_size,
...
-          // Now extract all arguments
+
+          // Skip argv[0] as we already have the file name from the executable path.
+          if (argc > 0 )
+            data.GetCStr(&offset);
+
+          // Now extract the rest of the arguments.
           Args &proc_args = process_info.GetArguments();
-          for (int i = 0; i < static_cast<int>(argc); ++i) {
+          for (int i = 1; i < static_cast<int>(argc); ++i) {
```
JDevlieghere, line 624: Does this match what we do on other platforms? I assume that generally, we would like this information (and I expect that there's code at least in LLDB relying on that). If we don't want to show that in the DAP case, why not filter it out there? It seems desirable to have this information, especially if it can differ as you point out in the PR description.

JDevlieghere (approved):

lldb/source/Host/macosx/objcxx/Host.mm
da-viper: > Does this match what we do on other platforms?

On linux and freebsd the exec path and arg0 is parsed differently (not yet implemented on windows). 

> and I expect that there's code at least in LLDB relying on that

Yes, at least `platform process list -v` does (the command uses argv0 instead of executable).  

> If we don't want to show that in the DAP case, why not filter it out there?

Argv0 is not yet exposed in the SBProcessInfo API or if it should be added to the ProcessInfo Arguments.

da-viper: see 
https://github.com/llvm/llvm-project/blob/d63ca96ec35e963e651d1b8765283117ca83e1f9/lldb/include/lldb/Utility/ProcessInfo.h#L120-L126

PR #198267, opened 2026-05-18
[RISCV][P-ext] Set BITCAST to Custom for 64-bit packed vectors on RV32

sihuan:
Bitcasts between i64 and v8i8/v4i16/v2i32 used to expand to a stack roundtrip, and the resulting concat_vectors let DAG combine split paired-register arithmetic into two single-reg ops (e.g. v8i8 add became two padd.b instead of one padd.db). The existing Is64BitCast handler in LowerOperation already treats these as no-ops; this just routes through it.


topperc (approved):
LGTM

Issue #198175, opened 2026-05-17
[X86] EltsFromConsecutiveLoads - handle trunc(wideload()) patterns


RKSimon:
Inspired by #179334 codegen where we have buildvec sequences such as:
```asm
	vmovd	216(%rsp), %xmm0                # xmm0 = mem[0],zero,zero,zero
	vpinsrb	$1, 224(%rsp), %xmm0, %xmm0
	vpinsrb	$2, 232(%rsp), %xmm0, %xmm0
	vpinsrb	$3, 240(%rsp), %xmm0, %xmm0
	vpinsrb	$4, 248(%rsp), %xmm0, %xmm0
	vpinsrb	$5, 256(%rsp), %xmm0, %xmm0
	vpinsrb	$6, 264(%rsp), %xmm0, %xmm0
	vpinsrb	$7, 272(%rsp), %xmm0, %xmm0
	vpinsrb	$8, 280(%rsp), %xmm0, %xmm0
	vpinsrb	$9, 288(%rsp), %xmm0, %xmm0
	vpinsrb	$10, 296(%rsp), %xmm0, %xmm0
	vpinsrb	$11, 304(%rsp), %xmm0, %xmm0
	vpinsrb	$12, 312(%rsp), %xmm0, %xmm0
	vpinsrb	$13, 320(%rsp), %xmm0, %xmm0
	vpinsrb	$14, 328(%rsp), %xmm0, %xmm0
	vpinsrb	$15, 336(%rsp), %xmm0, %xmm0
```
These are <16 x i8> gather loads with 8byte strides on the stack.

We should be able to recognise this as 2 v8i64 loads, 2 vtruncs and a concatenation:
```asm
  vmovdqu64 216(%rsp), %zmm0
  vmovdqu64 280(%rsp), %zmm1
  vpmovqb %zmm0, %xmm0
  vpmovqb %zmm1, %xmm1
  vpunpcklqdq %xmm1, %xmm0, %xmm0
```

as4230:
[...]

PR #198273, opened 2026-05-18
[X86] combineVectorInsert - attempt to merge consecutive constant insertions

RKSimon:
If we have 2 INSERT_VECTOR_ELT nodes in sequence (DAGCombine should have sorted these into ascending indices), then if they are both inserting constant scalars, attempt to concat the scalars together and insert the wider constant.

Reduces gpr pressure, gpr->vec traffic and (total) encoding size


llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -51655,13 +51658,46 @@ static SDValue combineVectorInsert(SDNode *N, SelectionDAG &DAG,
...
+        (CIdx->getZExtValue() & 1) == 1 && TLI.isTypeLegal(VT)) {
+      using namespace SDPatternMatch;
+      MVT WideSVT = MVT::getIntegerVT(2 * NumBitsPerElt);
+      MVT WideVT = MVT::getVectorVT(WideSVT, NumElts / 2);
+      SDValue InnerVec, InnerScl;
+      if (TLI.isTypeLegal(WideSVT) && TLI.isTypeLegal(WideVT) &&
+          sd_match(Vec, m_OneUse(m_InsertElt(
+                            m_Value(InnerVec), m_Value(InnerScl),
+                            m_SpecificInt(CIdx->getZExtValue() - 1))))) {
+        auto *Cst = dyn_cast<ConstantSDNode>(Scl);
```
phoebewang: Maybe move it out and check at 51673? Or you have plan to extend it to memory load?

phoebewang (approved):
LGTM.

llvm/lib/Target/X86/X86ISelLowering.cpp
RKSimon: I'll move it out for now - consecutive loads will depend on whether on how well #198175 works, but I might not get to it very soon (this patch started out to help my VECREDUCE work by reducing the depth necessary for computeKnownBits).

PR #198277, opened 2026-05-18
[AArch64] Match BSP through trunc(and) canonicalization in OR combine

### LINKED ISSUE #197360: Neon BSL not being used for select
In RPCS3, for the emulation of the SPU `FCGT` instruction, I ran into this issue, where llvm would refuse to emit BSL no matter what way I would massage the IR. The only way I found to work around was with inline assembly.

Here is a godbolt that reproduces the issue:
https://godbolt.org/z/4TfqsT37K

```asm
spu_FCGTbsl:
        lsr     w8, w22, #3
        mov     w8, w8
        and     x8, x24, x8
        ldr     q0, [x25, x8]
        and     w8, w24, w22, lsr #10
        ldr     q1, [x25, w8, uxtw]
        and     w8, w24, w22, lsl #4
        cmgt    v2.4s, v0.4s, v1.4s
        cmgt    v3.4s, v1.4s, v0.4s
        and     v4.16b, v1.16b, v0.16b
        fcmeq   v0.4s, v0.4s, v1.4s
        cmge    v4.4s, v4.4s, #0
        bsl     v4.16b, v2.16b, v3.16b
        bic     v0.16b, v4.16b, v0.16b
        str     q0, [x25, w8, uxtw]
        ret
```

```asm
spu_FCGT:
        lsr     w8, w22, #10
        mov     w8, w8
        and     x8, x24, x8
        ldr     q0, [x25, x8]
        and     w8, w24, w22, lsr #3
        ldr     q1, [x25, w8, uxtw]
        and     w8, w24, w22, lsl #4
        fcmeq   v2.4s, v1.4s, v0.4s
        cmgt    v3.4s, v0.4s, v1.4s
        cmgt    v4.4s, v1.4s, v0.4s
        and     v0.16b, v1.16b, v0.16b
        cmlt    v0.4s, v0.4s, #0
        xtn     v4.4h, v4.4s
        mvn     v2.16b, v2.16b
        xtn     v1.4h, v0.4s
        and     v0.16b, v3.16b, v0.16b
        xtn     v2.4h, v2.4s
        bic     v1.8b, v4.8b, v1.8b
        xtn     v0.4h, v0.4s
        orr     v0.8b, v0.8b, v1.8b
        and     v0.8b, v2.8b, v0.8b
        sshll   v0.4s, v0.4h, #0
        str     q0, [x25, w8, uxtw]
        ret
```

mmasztalerczuk:
When lowering `v[N]i1 select + mask + sext` patterns (e.g. the canonical `select(cmp_a, gt_lanes, lt_lanes)` masked by another setcc), the OR that implements the BSP-equivalent ends up after DAGCombiner has canonicalized

    (and (trunc x) (trunc y))  ->  (trunc (and x y))

so the post-legalize OR has shape

    (or (trunc (and A B)) (and C (xor (trunc A) -1)))

That shape is not recognized by `tryLowerToBSL` (called early from `LowerVectorOR`, before the canonicalization) nor by the BSP TableGen pattern, so AArch64 emits a redundant `bic + orr + and` sequence instead of a single `BIF/BSL/BIT`.

Fix #197360


mmasztalerczuk:
Hi, just following up on this PR after about a week. I’d appreciate any feedback when someone has a chance. Thanks!

davemgreen (commented):
Could this go into tryLowerToBSL like the others? It will help it avoid illegal types and keep it in one place.

(In the long run it might make sense just use VSELECT instead, but I only have looked into that for GISel so far).

mmasztalerczuk:
Done. The matcher is now inside tryLowerToBSL, and performORBSLTruncCombine is removed. I also moved the isTypeLegal(VT) check to the top of tryLowerToBSL so it covers both call sites.

One thing: Calling tryLowerToBSL only from LowerVectorOR is too early — DAGCombiner produces the (or (trunc (and A B)) ...) shape only afterwards. So performORCombine calls it again post-legalization (with an isVector() check, since it also runs for scalar ORs).

On VSELECT: happy to take a look, but it feels like a separate PR

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
Rajveer100: I think this can be much cleaner using matchers, same for many conditions above instead of doing multiple if-else for checking opcode.

mmasztalerczuk: Done, switched to sd_match. Kept isBitwiseNot instead of m_Not because m_Not (via strict m_AllOnes) misses the post-legalize xor (trunc x), <wider -1> shape
The older matchers above are unrelated to this fix, so I'd prefer to clean them up as a separate NFC PR.

Rajveer100: If I remember correctly, LLVM implements `m_Not` as a `Xor` operation (i.e, xor x, -1) so you should be able to detect that.

Also for commutative matching, you can use `m_c<Pattern>` instead of a loop.

mmasztalerczuk: Tried m_not, bsp-vselect-i1-sext.ll stops matching. m_Not is m_Xor(V, m_AllOnes()) and m_AllOnes rejects splats whose contant element storage is wider than the lane, which is exactly the post-legalize shape here, isbitwiseNot accepts it, so I kept the small wrapper.

The swap loop is already gone — m_Or/m_And in SDPatternMatch are commutative by default.

Rajveer100 (changes requested):
Please add tests for all commutative scenarios, like flipped `and` operands (ex. and bi ai) and other combinations to ensure this works in all cases.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
Rajveer100: > The swap loop is already gone — m_Or/m_And in SDPatternMatch are commutative by default.

I don't think this is true, `m_c_<binop>` matchers are for commutative matching and you can only get rid of the loop after adding them, and even the loop previously only took care for the `or` and `and` operands, what if we had `xor -1, x`.

Rajveer100:
Quick question: Were AI tools used in this pull request?

mmasztalerczuk:
> Please add tests for all commutative scenarios, like flipped `and` operands (ex. and bi ai) and other combinations to ensure this works in all cases.

Extended the matching to commutative operand orders (really just m_Xor in the end). Managed to do it without the custom matcher, m_ConstInt was enough.

Added cases that flip the operands around. For AND/OR they actually test the commutativity. The xor one passes, but it doesn't really test xor commutativity, since the code canonicalizes the constant to the right, so the matcher never sees it on the left. Keeping it anyway as a regression test to make sure the fold still fires for that IR shape.

Honestly these changes are mostly a refactor (less code), since the new tests also pass on the previous version. Let me know which version you prefer :)

mmasztalerczuk:
> Quick question: Were AI tools used in this pull request?

Yeah, I use AI as a learning and pair-programming tool because this codebase is fairly new to me. I actually use a custom prompt that makes the AI act as a critical reviewer rather than a coding assistant. He is not allowed to suggest or generate code. Instead, its job is to challenge my approach and helping understand how code is working. It works well most of the time, but sometimes it manages to disagree with itself :)

Rajveer100:
Please read:

https://llvm.org/docs/AIToolPolicy.html

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
davemgreen: I don't think this should be necessary as types at this point will be legal.

davemgreen: Do you have an example where this help post-legalize? Could you add it as a test-case?

mmasztalerczuk: this never actually fires (LowerVectorOR already catches the shape), it's a leftover. Dropped it.

mmasztalerczuk: Agreed, leftover from the old DAGCombine path — VT is always legal here. Removed.

mmasztalerczuk:
> Please read:
> 
> https://llvm.org/docs/AIToolPolicy.html

Read it, thanks, added an Assisted-by: trailer as the policy suggests.

davemgreen (approved):
LGTM - thanks

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
Rajveer100, line 15812:
```suggestion
  // Match the trunc-hoisted shape that hides BSP from the cases below:
  //   (or (trunc (and A B)) (and C (xor (trunc A) -1)))
  //     => BSP(trunc A, trunc B, C)
  // The xor constant uses
  // m_ConstInt + isAllOnes() rather than m_AllOnes because m_AllOnes rejects
  // post-legalize splats whose element storage is wider than the lane.
```

The reason why commutative approach worked is because of canonicalization of IR by moving complex expression to the left so I think we just got lucky here.

Rajveer100, line 15821: I still think a safer approach is to use `m_c_<op>` so we always handle it explicitly.

llvm/test/CodeGen/AArch64/bsp-vselect-i1-sext.ll
Rajveer100, line 58: Unless I missed, we need test cases for:

```
(or (trunc (and B A)) (and C (xor (trunc A) -1)))
(or (trunc (and B A)) (and C (xor (trunc B) -1)))
(or (trunc (and A B)) (and C (xor (trunc B) -1)))
```

and other similar ones.

Rajveer100 (commented):
I think the approach should work for now, but we can do better later on.

PR #198281, opened 2026-05-18
[libc][ctype] delegate core ctype checks to ctype_utils

kaladron:
Implemented internal helpers in ctype_utils.h and refactored isblank, isprint, iscntrl, and isascii to use them.

### DIFF
--- a/libc/src/__support/ctype_utils.h
+++ b/libc/src/__support/ctype_utils.h
@@ -597,6 +597,18 @@ LIBC_INLINE constexpr bool is_char_or_wchar(char ch, char c_value,
   return (ch == c_value);
 }
 
+LIBC_INLINE constexpr bool isblank(char ch) { return ch == ' ' || ch == '\t'; }
+
+LIBC_INLINE constexpr bool iscntrl(char ch) {
+  return static_cast<unsigned char>(ch) < 0x20 || ch == 0x7f;
+}
+
+LIBC_INLINE constexpr bool isprint(char ch) {
+  return static_cast<unsigned char>(ch - 0x20) < 95;
+}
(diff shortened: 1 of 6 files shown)


kaladron:
Heya @michaelrj-google - I wanted to bounce this one off of you.  This does two things for consistency:

1) It moves towards delegating to the internal header like other files in this directory

2) It adds:

  if (c < 0 || c > cpp::numeric_limits<unsigned char>::max())
    return 0;

at the top of each function like isalnum and isalpha.

I'm not certain that either of this is the right destination - in particular the second one, which adds an extra check that I think may not be needed.

I'd appreciate your thoughts on these!  I'm leaving this PR in draft, since this was just something I noticed when reading through the code and wanted to discuss.

michaelrj-google:
> Heya @michaelrj-google - I wanted to bounce this one off of you. This does two things for consistency:
> 
> 1. It moves towards delegating to the internal header like other files in this directory
> 2. It adds:
> 
> if (c < 0 || c > cpp::numeric_limits::max()) return 0;
> 
> at the top of each function like isalnum and isalpha.
> 
> I'm not certain that either of this is the right destination - in particular the second one, which adds an extra check that I think may not be needed.
> 
> I'd appreciate your thoughts on these! I'm leaving this PR in draft, since this was just something I noticed when reading through the code and wanted to discuss.

Yes I like this refactoring in concept. Adding the guards is probably necessary if we're casting from int to char.

If we're moving these into `ctype_utils` we should probably format them like the other ctype functions, i.e. encoding independent switch/cases. Possibly we could express `iscntrl` as `!isprint && isascii`?

PR #179474, opened 2026-02-03
[bolt] Simplify rodata/literal load for X86_64 & AArch64

yavtuk:
This patch fixed the issue related to load literal for AArch64, where constant island move to the end of a function (emitCI), available address range for ldr instruction is limited +/- 1MB, and can be out of range. For AArch64 the available data sizes are 4, 8 bytes. For X86_64 this patch materializes the constants with sizes 2, 4 and 8 bytes.
Note: SimplifyRODataLoads is turn off for non-relocation binary

PR #198282, opened 2026-05-18
[AArch64][BOLT] Document AArch64 optimization flag support

Amichaxx:
BOLTAArch64OptimizationStatus.rst aims to document default-off, user-enabled passes and their status (supported/unsupported) on AArch64.

``--simplify-rodata-loads`` has a patch in review, so I have documented its current status.

Discussion on RST/MD: https://discourse.llvm.org/t/rfc-make-myst-markdown-the-llvm-docs-format-rip-rest/90840

### DIFF
--- a/bolt/docs/BOLTAArch64OptimizationStatus.rst
+++ b/bolt/docs/BOLTAArch64OptimizationStatus.rst
@@ -11,9 +11,11 @@ explicitly enable when optimizing AArch64 binaries.
 BOLT is to be used with binaries linked with
 relocations (``--emit-relocs`` or ``-Wl,-q``) and representative profile data.
 
-Supported Flags
----------------
-The following flags are supported for AArch64.
+Main Code-Layout Optimizations
+------------------------------
+The following code-layout optimizations are typically the first options to
+consider when optimizing AArch64 binaries with representative profile data.
+They typically provide the largest performance gains among BOLT optimizations.
 
 .. list-table::
      :header-rows: 1
@@ -32,6 +34,19 @@ The following flags are supported for AArch64.
          | ``--split-all-cold``
          | ``--split-eh``
        - Split hot and cold code
+
+
+Other Supported Optimizations
+-----------------------------
+The following optimizations are also supported for AArch64.
+
+.. list-table::
+     :header-rows: 1
+     :widths: 34 42
+     :align: left
+
+     * - Flag
+       - Optimization
      * - | ``--align-blocks``
          | ``--block-alignment=<uint>``
        - Align basic blocks
--- a/bolt/docs/BOLTAArch64OptimizationStatus.rst
+++ b/bolt/docs/BOLTAArch64OptimizationStatus.rst
@@ -8,7 +8,7 @@ Overview
 This page summarizes default-off BOLT optimization flags that users may
 explicitly enable when optimizing AArch64 binaries.
 
-Ideally, BOLT is to be used with binaries linked with
+BOLT is to be used with binaries linked with
 relocations (``--emit-relocs`` or ``-Wl,-q``) and representative profile data.
 
 Supported Flags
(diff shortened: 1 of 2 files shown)


bolt/docs/index.rst
```diff
@@ -256,6 +256,17 @@ Profile Formats
 See `Profile Formats <profiles.md>`__ for comprehensive documentation of all
 profile formats accepted by BOLT: perf.data, fdata, YAML, and pre-aggregated.
```
Amichaxx, line 258: It looks like this is the main way that LLVM handles Sphinx hierarchy in llvm/docs (through a hidden toctree). My plan is to add some follow up patches to fix up BOLT's docs ready for hosting.

ash-arm (commented):
Looks good to me, but I will leave it to someone else to approve.

Nit: is this `\` needed in the PR description?
> has a patch in review, so I have documented its current status.\

bolt/docs/BOLTAArch64OptimizationStatus.rst
```diff
@@ -0,0 +1,128 @@
...
+This page summarizes default-off BOLT optimization flags that users may
+explicitly enable when optimizing AArch64 binaries.
+
+Ideally, BOLT is to be used with binaries linked with
+relocations (``--emit-relocs`` or ``-Wl,-q``) and representative profile data.
+
+Supported Flags
+---------------
+The following flags are supported for AArch64.
+
```
paschalis-mpeis, line 19: Not sure if it's worth to add the main reordering flags before any other ones?

I think split-functions could be put there as well as it's one of the main reordering optimizations we do, but we have to somehow list the `cdsplit` strategy under limitations.

bolt/docs/BOLTAArch64OptimizationStatus.rst
```diff
@@ -0,0 +1,128 @@
...
+     * - ``--reg-reassign``
+       - Reassign registers to reduce encoding size
+       - Not applicable to AArch64.
+     * - ``--simplify-rodata-loads``
+       - Replace read-only data loads with constants
+       - Not implemented for AArch64.
+     * - | ``--stoke``
+         | ``--stoke-out``
+       - Emit STOKE optimization data
+       - Not applicable to AArch64.
```
paschalis-mpeis, line 142: Maybe we could put all non-applicable in the end of this list? Or maybe in another section right after this one?

paschalis-mpeis (commented):
Hey Amina,

Thanks for your work. I had a quick skim and left a couple of comments.

ccing @aaupov in case he has additional feedback on the structure, etc. We discsused part of this some months ago in 'Office Hours'.

bolt/docs/BOLTAArch64OptimizationStatus.rst
Amichaxx: Thanks for the review!

`--split-strategy` is now in both Supported Flags and Supported Flags with Limitations, with the correct supported options.

bolt/docs/BOLTAArch64OptimizationStatus.rst
Amichaxx: Amended.

bolt/docs/index.rst
```diff
@@ -256,6 +256,17 @@ Profile Formats
...
 
+Additional Documentation
+------------------------
+
+.. toctree::
```
paschalis-mpeis: does this [render](https://github.com/Amichaxx/llvm-project/blob/57f58c55413bd15e520aac0d6f213cc41728ef15/bolt/docs/index.rst) as intended?

Amichaxx: Renders when tested with Sphinx. However I am thinking of changing this to be a visible TOC a bit like [Clang's](https://clang.llvm.org/docs/index.html) in a follow-up.

paschalis-mpeis (approved):
Hey Amina,

Looks good! No further changes on my end.

Amichaxx:
Currently #179474 is unmerged so I will leave its current status as is.

bolt/docs/index.rst
Amichaxx: I have decided to mirror [Clang's ](https://clang.llvm.org/docs/index.html)docs as it is a bit more user-friendly.

https://github.com/llvm/llvm-project/pull/201378

PR #198291, opened 2026-05-18
[LV] Add initial costs for VPInstructionWithType::computeCost

david-arm:
I noticed this was previously always returning a cost of 0
due to fear of triggering the (now deleted) legacy/vplan
cost model assert. Since the assert has now been removed
this should be safe to implement properly. I haven't
filled in the costs for all types yet, since there is
currently no way to expose those code paths. I suspect
for things like VPInstruction::StepVector the recipe is
always in the vector preheader and we never ask for its
cost.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlan.h
+++ b/llvm/lib/Transforms/Vectorize/VPlan.h
@@ -1581,10 +1581,7 @@ public:
 
   /// Return the cost of this VPInstruction.
   InstructionCost computeCost(ElementCount VF,
-                              VPCostContext &Ctx) const override {
-    // TODO: Compute accurate cost after retiring the legacy cost model.
-    return 0;
-  }
+                              VPCostContext &Ctx) const override;
 
   Type *getResultType() const { return getScalarType(); }
 
--- a/llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
+++ b/llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
@@ -1790,6 +1790,31 @@ void VPInstruction::printRecipe(raw_ostream &O, const Twine &Indent,
 }
 #endif
 
+InstructionCost VPInstructionWithType::computeCost(ElementCount VF,
+                                                   VPCostContext &Ctx) const {
+  // NOTE: At the moment it seems only possible to expose this path for
+  // the trunc, zext and sext opcodes. However, isScalarCast also covers
+  // int<>fp conversions, bitcasts, ptr<>int conversions, etc.
+  if (Instruction::isCast(getOpcode()))
+    return getCostForRecipeWithOpcode(getOpcode(), ElementCount::getFixed(1),
+                                      Ctx);
+
+  switch (getOpcode()) {
+  case VPInstruction::VScale: {
+    Type *Ty = this->getScalarType();
+    ArrayRef<Type *> Tys;
+    IntrinsicCostAttributes Attrs(Intrinsic::vscale, Ty, Tys);
+    return Ctx.TTI.getIntrinsicInstrCost(Attrs, Ctx.CostKind);
+  }
+  default:
+    // Although VPInstructionWithType is also used for
+    // VPInstruction::StepVector and VPInstruction::WideIVStep it isn't
+    // currently possible to expose cases where the cost is queried.
+    llvm_unreachable("Unhandled opcode");
+  }
+  return 0;
+}
+
 void VPInstructionWithType::execute(VPTransformState &State) {
   Type *ResultTy = getResultType();
   if (Instruction::isCast(getOpcode())) {
--- a/llvm/test/Transforms/LoopVectorize/RISCV/gather-scatter-cost.ll
+++ b/llvm/test/Transforms/LoopVectorize/RISCV/gather-scatter-cost.ll
@@ -92,55 +92,30 @@ exit:
 }
 
 define void @predicated_strided_store(ptr %start) {
-; RVA23-LABEL: @predicated_strided_store(
-; RVA23-NEXT:  entry:
-; RVA23-NEXT:    br label [[VECTOR_PH:%.*]]
-; RVA23:       vector.ph:
-; RVA23-NEXT:    [[TMP0:%.*]] = call <vscale x 8 x i64> @llvm.stepvector.nxv8i64()
-; RVA23-NEXT:    br label [[VECTOR_BODY:%.*]]
-; RVA23:       vector.body:
-; RVA23-NEXT:    [[VEC_IND:%.*]] = phi <vscale x 8 x i64> [ [[TMP0]], [[VECTOR_PH]] ], [ [[VEC_IND_NEXT:%.*]], [[VECTOR_BODY]] ]
-; RVA23-NEXT:    [[AVL:%.*]] = phi i64 [ 586, [[VECTOR_PH]] ], [ [[AVL_NEXT:%.*]], [[VECTOR_BODY]] ]
-; RVA23-NEXT:    [[TMP2:%.*]] = call i32 @llvm.experimental.get.vector.length.i64(i64 [[AVL]], i32 8, i1 true)
-; RVA23-NEXT:    [[TMP3:%.*]] = zext i32 [[TMP2]] to i64
-; RVA23-NEXT:    [[BROADCAST_SPLATINSERT:%.*]] = insertelement <vscale x 8 x i64> poison, i64 [[TMP3]], i64 0
-; RVA23-NEXT:    [[BROADCAST_SPLAT:%.*]] = shufflevector <vscale x 8 x i64> [[BROADCAST_SPLATINSERT]], <vscale x 8 x i64> poison, <vscale x 8 x i32> zeroinitializer
-; RVA23-NEXT:    [[TMP4:%.*]] = mul <vscale x 8 x i64> [[VEC_IND]], splat (i64 7)
-; RVA23-NEXT:    [[TMP5:%.*]] = getelementptr i8, ptr [[START:%.*]], <vscale x 8 x i64> [[TMP4]]
-; RVA23-NEXT:    call void @llvm.vp.scatter.nxv8i8.nxv8p0(<vscale x 8 x i8> zeroinitializer, <vscale x 8 x ptr> align 1 [[TMP5]], <vscale x 8 x i1> splat (i1 true), i32 [[TMP2]])
-; RVA23-NEXT:    [[AVL_NEXT]] = sub nuw i64 [[AVL]], [[TMP3]]
-; RVA23-NEXT:    [[VEC_IND_NEXT]] = add <vscale x 8 x i64> [[VEC_IND]], [[BROADCAST_SPLAT]]
-; RVA23-NEXT:    [[TMP7:%.*]] = icmp eq i64 [[AVL_NEXT]], 0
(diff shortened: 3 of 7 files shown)


llvm/test/Transforms/LoopVectorize/ext_or_trunc_iv_costs.ll
david-arm, line 1: NOTE: I have deliberately created a test without a particular target so that I can ensure that truncates and extends are not free.

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -1527,6 +1527,8 @@ class VPInstructionWithType : public VPInstruction {
     auto *VPI = dyn_cast<VPInstruction>(R);
     if (!VPI)
       return false;
+    // TODO: This switch statement doesn't include the full range of supported
```
fhahn: I am not sure about the TODO, we should cover all cases handled by `VPInstructionWithType`, not all VPInstruction types. It would be a bug if we miss an opcode for which we create `VPInstructionWithType`, but return `false` here. If there are such cases, that should be fixed

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1625,6 +1625,34 @@ void VPInstruction::printRecipe(raw_ostream &O, const Twine &Indent,
 }
 #endif
 
+InstructionCost VPInstructionWithType::computeCost(ElementCount VF,
+                                                   VPCostContext &Ctx) const {
+  // TODO: At the moment it's not possible to write a test that exposes a case
+  // where more than one lane is used. For now just bail out with a cost of 0.
+  if (!vputils::onlyFirstLaneUsed(this))
+    return 0;
```
fhahn: if it is not possible, it should be an assert?

In general, other than for `WideIVStep, StepVector`, we currently should always generate a scalar, independent of whether only the first lane is used

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1625,6 +1625,34 @@ void VPInstruction::printRecipe(raw_ostream &O, const Twine &Indent,
...
+
+  switch (getOpcode()) {
+  case Instruction::Trunc:
+  case Instruction::SExt:
+  case Instruction::ZExt: {
```
fhahn: is there a reason to limit this to only those casts?

david-arm: The reason is that it's impossible to test other cases.

david-arm: I've tried to avoid creating cost model for things we cannot test that's all. I tried building the LLVM test suite in different scenarios and we never hit the other code paths.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
david-arm: OK, I'm happy to add an assert! I just wasn't sure if that was preferred or not

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
david-arm: If you think we also generate FP<>FP casts and int<>FP casts then I could try writing some tests for them. Alternatively, I could write a vplan unit test?

llvm/lib/Transforms/Vectorize/VPlan.h
david-arm: OK, I don't think any such changes should live in this PR though as it's purely a cost model change. I'm happy to investigate updating the comment and enums in the VPInstructionWithType class if it's simply impossible to create VPInstructionWithType with other opcodes.

david-arm: If I update the switch statement below to cover all the cases that VPInstructionWithType claims we use the class for I see lots of test failures. For example, see test/Transforms/LoopVectorize/select-cmp.ll:

```
opt: /home/davshe01/regression/llvm/llvm/lib/Transforms/Vectorize/VPlanAnalysis.cpp:318: llvm::Type* llvm::VPTypeAnalysis::inferScalarType(const llvm::VPValue*): Assertion `ResultTy && "could not infer type for the given VPValue"' failed.
```

I'm a bit worried that we're going down the wrong code paths for many opcodes.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1625,6 +1625,34 @@ void VPInstruction::printRecipe(raw_ostream &O, const Twine &Indent,
...
+    return getCostForRecipeWithOpcode(getOpcode(), ElementCount::getFixed(1),
+                                      Ctx);
+  }
+  case VPInstruction::VScale: {
+    Type *Ty = Ctx.Types.inferScalarType(this);
+    ArrayRef<Type *> Tys;
+    IntrinsicCostAttributes Attrs(Intrinsic::vscale, Ty, Tys);
+    return Ctx.TTI.getIntrinsicInstrCost(Attrs, Ctx.CostKind);
```
lukel97, line 1807: Do the vscales not get hoisted out of the vector loop body similar to stepvector? Or do we hit this costing code somewhere else

david-arm: I think I've added a test for this. They get costed because they appear in the middle block

llvm/lib/Transforms/Vectorize/VPlan.h
fhahn: If we would go down the wrong code path (i.e. constructing VPInsructionWihType, but missing them here), I would expect crashes. I think this is a case where the code is correct but the comment is stale

david-arm: Yeah I think you're right. I'm working on a separate PR to tidy up the comment. There are still some missing cases in VPInstructionWithType::execute because we definitely generate VPInstructionWithType objects for scalar loads (see createVPInstructionsForVPBB) and VPInstruction::WideIVStep (via createNaryOp). I guess for the scalar loads we would never explicitly call execute on them because they are probably replaced with wide loads or replicates? For WideIVStep I assume we never have to execute them because they always get transformed?

llvm/test/Transforms/LoopVectorize/RISCV/gather-scatter-cost.ll
```diff
@@ -92,55 +92,30 @@ exit:
...
+; CHECK-NEXT:  entry:
+; CHECK-NEXT:    br label [[VECTOR_PH:%.*]]
+; CHECK:       vector.ph:
+; CHECK-NEXT:    [[TMP0:%.*]] = call <vscale x 8 x i64> @llvm.stepvector.nxv8i64()
+; CHECK-NEXT:    br label [[VECTOR_BODY:%.*]]
+; CHECK:       vector.body:
+; CHECK-NEXT:    [[VEC_IND:%.*]] = phi <vscale x 8 x i64> [ [[TMP0]], [[VECTOR_PH]] ], [ [[VEC_IND_NEXT:%.*]], [[VECTOR_BODY]] ]
+; CHECK-NEXT:    [[AVL:%.*]] = phi i64 [ 586, [[VECTOR_PH]] ], [ [[AVL_NEXT:%.*]], [[VECTOR_BODY]] ]
+; CHECK-NEXT:    [[TMP1:%.*]] = call i32 @llvm.experimental.get.vector.length.i64(i64 [[AVL]], i32 8, i1 true)
+; CHECK-NEXT:    [[TMP2:%.*]] = zext i32 [[TMP1]] to i64
```
david-arm, line 105: @lukel97 I see there is a zext here, so perhaps this explains why accounting for the extend cost bumps up the VF?

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1604,6 +1625,31 @@ void VPInstruction::printRecipe(raw_ostream &O, const Twine &Indent,
 }
 #endif
 
+InstructionCost VPInstructionWithType::computeCost(ElementCount VF,
+                                                   VPCostContext &Ctx) const {
+  // NOTE: At the moment it seems only possible to expose this path for
+  // the trunc, zext and sext opcodes. However, isScalarCast also covers
+  // int<>fp conversions, bitcasts, ptr<>int conversions, etc.
+  if (isScalarCast())
```
artagnon:
```suggestion
  if (Instruction::isCast(getOpcode()))
```

Needs rebase?

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1604,6 +1625,31 @@ void VPInstruction::printRecipe(raw_ostream &O, const Twine &Indent,
...
+    return getCostForRecipeWithOpcode(getOpcode(), ElementCount::getFixed(1),
+                                      Ctx);
+
+  switch (getOpcode()) {
+  case VPInstruction::VScale: {
```
artagnon, line 1803: Hm, why wouldn't we want to fold this into getCostForRecipeWithOpcode?

david-arm: It looks like getCostForRecipeWithOpcode is only used for genuine Instruction opcodes. Not sure if we want to add VPInstruction opcodes there too? cc @fhahn

fhahn: Yep, I think it would be good to limit `getCostForRecipeWithOpcode` to just Instruction opcodes, `VPInstruction::VScale` should only be possible in this path here

fhahn (approved):
LGTM, thanks

lukel97:
I'm seeing crashes with this patch building the MicroBenchmarks/all target on llvm-test-suite with `-target riscv64-linux-gnu -march=rva23u64 -O3`, it looks like an opcode is slipping through?

```
Unhandled opcode
UNREACHABLE executed at /Users/luke/Developer/llvm-project/llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp:1813!
```

Will work on a minimal reproducer

david-arm:
> I'm seeing crashes with this patch building the MicroBenchmarks/all target on llvm-test-suite with `-target riscv64-linux-gnu -march=rva23u64 -O3`, it looks like an opcode is slipping through?
> 
> ```
> Unhandled opcode
> UNREACHABLE executed at /Users/luke/Developer/llvm-project/llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp:1813!
> ```
> 
> Will work on a minimal reproducer

Yep, I've seen this too and will revert.

david-arm:
Revert is here: https://github.com/llvm/llvm-project/pull/202933

david-arm:
Thanks @lukel97, a reproducer would be awesome!

lukel97:
This crashes with `opt -mtriple riscv64 -mattr=+v -p loop-vectorize`, it might require some massaging to make it target agnostic though

```llvm
define void @pointer_induction_stepvector_cost(ptr %end) {
entry:
br label %loop
 
loop:
%dst = phi ptr [ %dst.next, %loop ], [ null, %entry ]
%src = phi ptr [ %src.next, %loop ], [ null, %entry ]
store i32 0, ptr %dst, align 2
%src.next = getelementptr i8, ptr %src, i64 4
%dst.next = getelementptr i8, ptr %dst, i64 4
%cmp = icmp eq ptr %src, %end
br i1 %cmp, label %exit, label %loop
 
exit:
ret void
}
```

PR #198292, opened 2026-05-18
[SafeStack] Introduce public interface header and deprecate builtins

jakos-sec:
Introduce `sanitizer/safestack_interface.h` to expose SafeStack utility
functions as a public interface, similar to other sanitizers.

Exposed functions:
* `__safestack_get_unsafe_stack_ptr`
* `__safestack_get_unsafe_stack_bottom`
* `__safestack_get_unsafe_stack_top`

Deprecate the existing `__builtin___get_unsafe_stack_*` builtins and
emit warnings suggesting the new `__safestack_` equivalents.

Runtime compatibility aliases `__get_unsafe_stack_*` are maintained.


vitalybuka:
Introduced https://reviews.llvm.org/D6095, but I can't tell why builtin was chosen.

@pcc @AaronBallman @efriedma-quic

Is there any rule on how to pick to introduce a builtin vs just function in complier-rt header?
SafeStack looks inconsistent with the rest, but maybe a reason for that?

We need to add more SafeStack related functions, but I am not sure if we should stick with the existing approach or convert to regular functions.

efriedma-quic:
I don't think there's any reason to avoid using a header, in general, assuming it gets installed somewhere appropriate.  Well, as long as you're sure the compiler doesn't need to do anything special with the function.  It look like there's some special handling related to `__safestack_get_unsafe_stack_ptr` in the backend.

compiler-rt/include/sanitizer/safestack_interface.h
```diff
@@ -0,0 +1,35 @@
...
+#endif
+
+/// Returns the current unsafe stack pointer of the current thread.
+void SANITIZER_CDECL *__safestack_get_unsafe_stack_ptr(void);
+
+/// Returns a pointer to the bottom of the unsafe stack of the current thread.
+void SANITIZER_CDECL *__safestack_get_unsafe_stack_bottom(void);
+
+/// Returns a pointer to the top of the unsafe stack of the current thread.
+void SANITIZER_CDECL *__safestack_get_unsafe_stack_top(void);
```
AaronBallman:
```suggestion
void * SANITIZER_CDECL __safestack_get_unsafe_stack_ptr(void);

/// Returns a pointer to the bottom of the unsafe stack of the current thread.
void * SANITIZER_CDECL __safestack_get_unsafe_stack_bottom(void);

/// Returns a pointer to the top of the unsafe stack of the current thread.
void * SANITIZER_CDECL __safestack_get_unsafe_stack_top(void);
```
Otherwise the annotation splits the return type making it harder to read.

Also, should these be returning `const void *` as a hint to not mutate the returned memory?

clang/lib/Sema/SemaChecking.cpp
```diff
@@ -2915,6 +2915,26 @@ Sema::CheckBuiltinFunctionCall(FunctionDecl *FDecl, unsigned BuiltinID,
 
   FPOptions FPO;
   switch (BuiltinID) {
+  case Builtin::BI__builtin___get_unsafe_stack_start:
+    Diag(TheCall->getBeginLoc(), diag::warn_deprecated_builtin)
+        << Context.BuiltinInfo.getQuotedName(BuiltinID)
+        << "__safestack_get_unsafe_stack_bottom";
+    break;
+  case Builtin::BI__builtin___get_unsafe_stack_bottom:
```
AaronBallman, line 2924:
```suggestion
  case Builtin::BI__builtin___get_unsafe_stack_start:
  case Builtin::BI__builtin___get_unsafe_stack_bottom:
```

clang/test/Sema/safestack-deprecated-builtins.c
```diff
@@ -0,0 +1,17 @@
+// RUN: %clang_cc1 %s -fsyntax-only -verify -fsanitize=safe-stack
+
+void *test_bottom(void) {
+  return __builtin___get_unsafe_stack_bottom(); // expected-warning {{builtin '__builtin___get_unsafe_stack_bottom' is deprecated; use __safestack_get_unsafe_stack_bottom instead}}
```
AaronBallman, line 4:
```suggestion
  return __builtin___get_unsafe_stack_bottom(); // expected-warning {{builtin '__builtin___get_unsafe_stack_bottom' is deprecated; use '__safestack_get_unsafe_stack_bottom' instead}}
```
Syntax constructs should be quoted in diagnostics, same should be changed for the others as well.

AaronBallman (commented):
This should also come with a release note in `clang/docs/ReleaseNotes.rst` to alert users to the deprecation.

clang/test/Sema/safestack-deprecated-builtins.c
jakos-sec: Done.

clang/lib/Sema/SemaChecking.cpp
jakos-sec: Done.

compiler-rt/include/sanitizer/safestack_interface.h
jakos-sec: Done, also made them `const void *`, thank you for the suggestion!

jakos-sec:
> I don't think there's any reason to avoid using a header, in general, assuming it gets installed somewhere appropriate. Well, as long as you're sure the compiler doesn't need to do anything special with the function. It look like there's some special handling related to `__safestack_get_unsafe_stack_ptr` in the backend.

I've added `__safestack_get_unsafe_stack_ptr`  in this commit and it's only used in a test. There shouldn't be any other use of it in the backend, if I'm not mistaken.

> This should also come with a release note in clang/docs/ReleaseNotes.rst to alert users to the deprecation.

Also added that, thank you for the suggestion.

vitalybuka (approved):
@pcc also confirmed offline that they are likely builtins by acident.

vitalybuka (changes requested):
Note: we are disscussed possible optimization, like reserving register or optimizing TLS access.

But they should be likely unaffected by clang intrinsic. LLVM instrumentation can skip function call anyway.

These are for user  code to get some internal state,  e.g. for diagnostics. It's OK to make the functions always.

vitalybuka (approved):

efriedma-quic:
> I've added __safestack_get_unsafe_stack_ptr in this commit and it's only used in a test. There shouldn't be any other use of it in the backend, if I'm not mistaken.

Oh, hmm, I think I got confused looking at SafeStack.cpp.  There is a function `__safestack_pointer_address`, but I guess that isn't quite the same thing?

jakos-sec:
> > I've added __safestack_get_unsafe_stack_ptr in this commit and it's only used in a test. There shouldn't be any other use of it in the backend, if I'm not mistaken.
> 
> Oh, hmm, I think I got confused looking at SafeStack.cpp. There is a function `__safestack_pointer_address`, but I guess that isn't quite the same thing?

Ah yes, that one is actually used by the compiler and the pass itself (although disabled by default), but is fully unrelated to the new `__safestack_*` diagnostic based functions. Good point though!

PR #198293, opened 2026-05-18
[X86] LowerBUILD_VECTORvXi1 - attempt to fold as VPTESTMB(BUILD_VECTOR_vXi8(X),1) for v2i1/v4i1/v8i1 types

RKSimon:
Extends #198166 to handle cases where we need to BUILD_VECTOR using vXi8 smaller than 128-bits


llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -8552,17 +8552,24 @@ static SDValue LowerBUILD_VECTORvXi1(SDValue Op, const SDLoc &dl,
...
-  // TODO: Add handling for sub-128bit vXi8 vectors.
   MVT OpVT = Op.getOperand(0).getSimpleValueType();
   if (NonConstIdx.size() > 1 && OpVT == MVT::i8) {
     // On pre-BWI targets, we must extend to vXi32 instead.
     MVT ByteVT = VT.changeVectorElementType(MVT::i8);
     MVT WideVT =
         Subtarget.hasBWI() ? ByteVT : VT.changeVectorElementType(MVT::i32);
+    if (ByteVT.getSizeInBits() < 128) {
+      ByteVT = MVT::v16i8;
+      WideVT = VT.changeVectorElementType(MVT::i64);
```
phoebewang: Should limit to v2i8 only?

RKSimon: This is during lowering so we need to keep to legal types (v16i8) and rely on ANY_EXTEND_VECTOR_INREG to only expand the lower elements - luckily because it is legal types, we don't have to worry about VLX/NoVLX paths at this stage :)

phoebewang: I mean for others, the vXi32 is enough.

RKSimon: v8i1 -> v8i32 hit regressions due to some weird padding - but v4i1 -> v4i32 is a clear win

phoebewang (approved):
LGTM.

PR #187860, opened 2026-03-21
Add clang warning if fp exception functions are called without appropriate flags/pragmas

maarcosrmz:
Fixes https://github.com/llvm/llvm-project/issues/128239

The implementation adds warnings for floating-point exception function calls (fenv.h) made without enabling floating-point exception behavior via `-ffp-exception-behavior=maytrap/strict` or `#pragma STDC FENV_ACCESS ON`. To support recognition of all fenv.h builtins, `fexcept_t` and `fenv_t` were added as builtin types.

PR #198295, opened 2026-05-18
[CI] Run libc tests on clang changes

boomanaiden154:
The libc tests are relatively lightweight, and given we build libc with a just built clang, it's very easy for clang changes to cause issues in libc, especially with -Werror. For example, #187860 broke libc due to adding a new warning that libc was not clean on.


cmtice (approved):

cmtice:
I'd like to un-do my approval until the failing CI tests are fixed.

boomanaiden154:
Fixed in https://github.com/llvm/llvm-project/pull/198303.

PR #198296, opened 2026-05-18
[MLIR][CMake] Downgrade ocloc not found diagnostic from WARNING to STATUS

harrisonGPU:
When ocloc is not found, the CMake configuration emits a WARNING which
can be noisy for users who have no interest in Intel XeVM native binary
compilation (e.g., those targeting only AMDGPU or NVPTX). Downgrade the
diagnostic to STATUS since this is an expected configuration when ocloc
is not installed. 
For example:
```bash
CMake Warning at /llvm-project/mlir/lib/Target/LLVM/CMakeLists.txt:246 (message):
ocloc not found, MLIRXeVMTarget will not be able to use ocloc for native
binary compilation.
```


krzysz00 (approved):
lgtm, approved

mlir/lib/Target/LLVM/CMakeLists.txt
```diff
@@ -243,7 +243,7 @@ if (NOT LLVM_BUILD_LLVM_DYLIB)
     set(MLIR_XEVM_OCLOC_LIB_AVAILABLE 1)
   else()
     set(MLIR_XEVM_OCLOC_LIB_AVAILABLE 0)
-    message(WARNING "ocloc not found, MLIRXeVMTarget will not be able to use ocloc for native binary compilation.")
+    message(STATUS "ocloc not found, MLIRXeVMTarget will not be able to use ocloc for native binary compilation.")
```
mshahneo, line 246: Can we keep the warning? Is it too much of a nuisance?  Since, for the `format=binary` path `ocloc` is a must have.

harrisonGPU: Thanks, but I think STATUS is right here. When the `ocloc` library is not found at build time, `compileToBinary()` falls back to the external `ocloc` tool on `$PATH`, so native binary compilation still works. The library is just the faster in-process path. Emitting a `WARNING` would be misleading for users who have `ocloc` installed as a standalone tool but not as a linkable library, which is a pretty common setup.

mshahneo: Got it, thanks. Makes sense.

mshahneo (approved):
LGTM.

PR #198300, opened 2026-05-18
[LV] Vectorize fmin/fmax reductions

nasherm:
Teach LoopVectorize to recognize fmin/fmax reductions. This patch only supports reductions of this kind which have the FP `minimum/maximum` and `minimumnum/maximumnum` semantics.


llvm/lib/Transforms/Vectorize/LoopVectorizationLegality.cpp
```diff
@@ -875,6 +875,7 @@ bool LoopVectorizationLegality::canVectorizeInstr(Instruction &I) {
       return false;
     }
 
+    // TODO: failing here
```
artagnon: Hm?

llvm/lib/Analysis/IVDescriptors.cpp
```diff
@@ -1090,6 +1125,7 @@ bool RecurrenceDescriptor::isReductionPHI(PHINode *Phi, Loop *TheLoop,
     LLVM_DEBUG(dbgs() << "Found a XOR reduction PHI." << *Phi << "\n");
     return true;
   }
+
```
artagnon: Stray change?

llvm/lib/Analysis/IVDescriptors.cpp
```diff
@@ -1432,6 +1468,7 @@ bool InductionDescriptor::isFPInductionPHI(PHINode *Phi, const Loop *TheLoop,
 
   // Here we only handle FP induction variables.
   assert(Phi->getType()->isFloatingPointTy() && "Unexpected Phi type");
+  errs() << "Phi = " << *Phi << '\n';
```
artagnon: Hm?

llvm/lib/Analysis/IVDescriptors.cpp
```diff
@@ -292,6 +292,19 @@ static RecurrenceDescriptor getMinMaxRecurrence(PHINode *Phi, Loop *TheLoop,
       return RecurKind::FMinimumNum;
     if (match(V, m_Intrinsic<Intrinsic::maximumnum>(m_Value(A), m_Value(B))))
       return RecurKind::FMaximumNum;
+    // A select(fcmp(A, B)) is equivalent to a min/max recurrence
+    // depending on the condition used
+    if (match(V, m_FCmp(m_Value(A), m_Value(B)))) {
```
john-brawn-arm: It looks like m_OrdOrUnordFMax and m_OrdOrUnordFMin could be used here instead.

fhahn (commented):
From the description it sounds like this is combining 2 changes:

* support for FP min/max
* recognize min/max reductions whose compare feeds a parallel select-based def-use chain

Would be good to split up if that's the case

llvm/lib/Analysis/IVDescriptors.cpp
nasherm: The patch has been broken up so I will do this in later patches

nasherm:
> From the description it sounds like this is combining 2 changes:
> 
>     * support for FP min/max
> 
>     * recognize min/max reductions whose compare feeds a parallel select-based def-use chain
> 
> 
> Would be good to split up if that's the case

Makes sense. I've reduced this patch down to supporting FP min/max. I will add the parallel def-use chain in a later PR

nasherm:
I believe I may have found a way to properly handle multi-def-use chains by falling back to existing parts of the LV which should allow for this patch to cover both cases simultaneously in one _relatively minimal_  patch.

I'm moving this patch to draft for the meantime as I work on this

nasherm:
Support for parallel min/max reductions will be added in a later patch.

llvm/test/Transforms/LoopVectorize/select-first-index-fp.ll
john-brawn-arm, line 354: Why do the tests before this not get optimized? Given that the title is "Vectorize fmin/fmax" I would expect that the test_fmin and test_fmax tests would be optimized.

nasherm: This seems to be an issue with `RecurKind` for the `llvm.minnum` and `llvm.maxnum` intrinsics. These intrinsics have a different recurrence kind to `llvm.maximumnum` and `llvm.minimumnum` and require some extra work to support them for vectorization.

nasherm: For clarity: that is to say that the recurrence kind found in the loop patterns that are vectorizing is not the same as the ones that aren't

nasherm:
> Why do the tests before this not get optimized? Given that the title is "Vectorize fmin/fmax" I would expect that the test_fmin and test_fmax tests would be optimized.

Thanks for spotting this. My most recent patch should resolve this. It looks like maxnum/minnum reductions are handled slighly differently to min/max reductions in that there's a dedicated `VPlanTransforms::handleMaxMinNumReductions` pass for them. However, the pattern in the tests should be covered by `VPlanTransforms::handleMultiUseReductions` . I've added a check to make sure we don't fail in `handleMaxMinNumReductions`.

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
```diff
@@ -1680,6 +1682,20 @@ bool VPlanTransforms::handleMaxMinNumReductions(VPlan &Plan) {
...
+  bool AllUnsupportedPhisInMultiUseChain =
+      all_of(UnsupportedVPRdxPhis, [&](VPReductionPHIRecipe *R) {
+        return R->hasOneUse() && all_of(R->users(), [](VPUser *U) {
+                 CmpPredicate Pred;
+                 return match(U, m_Select(m_Cmp(Pred, m_VPValue(), m_VPValue()),
+                                          m_VPValue(), m_VPValue()));
+               });
+      });
+  if (!UnsupportedVPRdxPhis.empty() && AllUnsupportedPhisInMultiUseChain)
+    return true;
```
fhahn: I don't think this is correct; IIRC min/maxnum reductions need special handling due to the way they handle signaling NaN.

nasherm: Thanks for letting me know Florian, I had a dig through the langref to try and understand better.

@john-brawn-arm the langref seems to indicate that failure to vectorize `minnum/maxnum` reductions is expected due to the semantics of `minnum/maxnum` and it's handling of sNaNs 

https://llvm.org/docs/LangRef.html#llvm-minnum-intrinsic

Would you expect the difference in semantics not to make a difference in whether these vectorize? Correct me if I'm wrong but it seems like the failure to vectorize the reductions in the test cases you mentioned would have to be done in the patch adding support for the parallel def-use reductions. So _this_ current patch is really just be adding support for fmin/fmax reductions which have the simpler semantics of `maximumnum/minimumnum`

john-brawn-arm: I was just commenting on the apparent contradiction between what the PR title and what was in the test. I don't know enough about the semantics of the various kinds for fp min/max to have an opinion on whether we should be vectorizing or not.

nasherm:
> I was just commenting on the apparent contradiction between what the PR title and what was in the test. I don't know enough about the semantics of the various kinds for fp min/max to have an opinion on whether we should be vectorizing or not.

I've updated the PR description to be more specific. I've also updated  `handleMultiUseReductions` so that we don't attempt to vectorize `FMaxNum/FMinNum` reductions

nasherm:
Ping

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
```diff
@@ -2177,6 +2177,18 @@ bool VPlanTransforms::handleMultiUseReductions(VPlan &Plan,
         return Pred == CmpInst::ICMP_SLE || Pred == CmpInst::ICMP_SLT;
       case RecurKind::SMin:
         return Pred == CmpInst::ICMP_SGE || Pred == CmpInst::ICMP_SGT;
+      case RecurKind::FMax:
+      case RecurKind::FMaximum:
+      case RecurKind::FMaximumNum:
+        return Pred == CmpInst::FCMP_OLE || Pred == CmpInst::FCMP_OLT;
+      case RecurKind::FMin:
+      case RecurKind::FMinimum:
```
john-brawn-arm: I think this is wrong for fminimum (and probably fmaximum, but I haven't checked that). Using this test program
```
#include <stdio.h>
#include <math.h>

float fminimum_numf(float, float);

__attribute__((noinline))
long test_min(float *p, long n) {
  float res = HUGE_VALF;
  long idx = 0;
  for (long i = 0; i < n; i++) {
    idx = p[i] < res ? i : idx;
    res = fmin(p[i], res);
  }
  return idx;
}

__attribute__((noinline))
long test_minimum(float *p, long n) {
  float res = HUGE_VALF;
  long idx = 0;
  for (long i = 0; i < n; i++) {
    idx = p[i] < res ? i : idx;
    res = __builtin_elementwise_minimum(p[i], res);
  }
  return idx;
}

__attribute__((noinline))
long test_minimumnum(float *p, long n) {
  float res = HUGE_VALF;
  long idx = 0;
  for (long i = 0; i < n; i++) {
    idx = p[i] < res ? i : idx;
    res = fminimum_numf(p[i], res);
  }
  return idx;
}

int main() {
  float arr[] = { 9, 2, 3, 4, 5, 6, 7, 8, NAN, 4, 3, -1, 2, 33, 0, NAN };
  printf("test_min %ld\n", test_min(arr, sizeof(arr)/sizeof(float)));
  printf("test_minimum %ld\n", test_minimum(arr, sizeof(arr)/sizeof(float)));
  printf("test_minimumnum %ld\n", test_minimumnum(arr, sizeof(arr)/sizeof(float)));
  return 0;
}
```
Without this patch it prints
```
test_min 11
test_minimum 1
test_minimumnum 11
```
With this patch it prints
```
test_min 11
test_minimum 0
test_minimumnumf 11
```
minimum returns NaN if either operand is NaN, but ordered comparisons (e.g. FCMP_OGE) are true only if neither operand is NaN. I think probably for minimum/maximum we need to be checking for unordered comparisons, but I haven't tested this.

nasherm: Thanks for pointing this out. My most recent patch should resolve this

john-brawn-arm (approved):
LGTM

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
```diff
@@ -2252,7 +2266,9 @@ bool VPlanTransforms::handleMultiUseReductions(VPlan &Plan,
     VPBuilder B(FindIVRdxResult);
     VPValue *MinOrMaxExiting = MinOrMaxResult->getOperand(0);
     auto *FinalMinOrMaxCmp =
-        B.createICmp(CmpInst::ICMP_EQ, MinOrMaxExiting, MinOrMaxResult);
+        (RecurrenceDescriptor::isIntegerRecurrenceKind(RdxKind))
+            ? B.createICmp(CmpInst::ICMP_EQ, MinOrMaxExiting, MinOrMaxResult)
+            : B.createFCmp(CmpInst::FCMP_OEQ, MinOrMaxExiting, MinOrMaxResult);
```
fhahn: I don't think that will work correctly at least for FMinimum, FMaximum, which propagate NaN. If any input lane is NaN, then `MinOrMaxResult` will be NaN, and `OEQ` will always be false, and we incorrectly select the first lane.

 Not sure if all cases covered in the PR are covered by https://github.com/llvm/llvm-test-suite/blob/main/SingleSource/UnitTests/Vectorizer/argmin-argmax-fp.cpp, but if not it would be good to make sure they are covered first, so it is easier to catch issues

nasherm: > I don't think that will work correctly at least for FMinimum, FMaximum

Thanks for spotting this. When addressing John's earlier comment addressing the differing semantics for FMinimum/FMaximum I forgot to make a change on this line. Should be correct now

> but if not it would be good to make sure they are covered first, so it is easier to catch issues

Sure. I've added extra test cases to the llvm-test-suite file https://github.com/llvm/llvm-test-suite/pull/436

nasherm: Gentle ping

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
fhahn, line 2033: I think we also need to update the checks in this function, as we now can also reach here with floats? Would be good to add some test cases and maybe also to the test-suite tests?

nasherm: I've done some local testing and `handleFirstArgMinOrMax` doesn't get triggered with FP types.  This is because of the following snippet 

```cpp
@@ -2217,7 +2230,7 @@ bool VPlanTransforms::handleMultiUseReductions(VPlan &Plan,

 
bool IsStrictPredicate = ICmpInst::isLT(Pred) || ICmpInst::isGT(Pred);
     if (IsStrictPredicate) {
       if (!handleFirstArgMinOrMax(Plan, MinOrMaxPhiR, FindIVPhiR,
                                   cast<VPWidenIntOrFpInductionRecipe>(IVOp),
.... 
    }
```

The `IsStrictPredicate` is only true for `Pred` using integer compares rather than both integer and floating point. I'm wondering if I should add support for FP strict predicates as well i.e. making the following change? 

```diff
@@ -2217,7 +2230,7 @@ bool VPlanTransforms::handleMultiUseReductions(VPlan &Plan,
     MinOrMaxResult->moveBefore(*FindIVRdxResult->getParent(),
                                FindIVRdxResult->getIterator());
 
-    bool IsStrictPredicate = ICmpInst::isLT(Pred) || ICmpInst::isGT(Pred);
+    bool IsStrictPredicate = CmpInst::isStrictPredicate(Pred);
     if (IsStrictPredicate) {
       if (!handleFirstArgMinOrMax(Plan, MinOrMaxPhiR, FindIVPhiR,
                                   cast<VPWidenIntOrFpInductionRecipe>(IVOp),
~
```

This change making support for FP in `handleArgMinOrMax` worth the time

llvm/test/Transforms/LoopVectorize/select-first-index-fp.ll
```diff
@@ -354,22 +354,22 @@ exit:
...
-; CHECK-NEXT:  [[ENTRY:.*]]:
-; CHECK-NEXT:    br label %[[LOOP:.*]]
-; CHECK:       [[LOOP]]:
-; CHECK-NEXT:    [[IV:%.*]] = phi i64 [ 0, %[[ENTRY]] ], [ [[IV_NEXT:%.*]], %[[LOOP]] ]
-; CHECK-NEXT:    [[MIN_IDX:%.*]] = phi i64 [ 0, %[[ENTRY]] ], [ [[MIN_IDX_NEXT:%.*]], %[[LOOP]] ]
-; CHECK-NEXT:    [[MIN_VAL:%.*]] = phi float [ f0x7F7FFFFF, %[[ENTRY]] ], [ [[MIN_VAL_NEXT:%.*]], %[[LOOP]] ]
-; CHECK-NEXT:    [[GEP:%.*]] = getelementptr float, ptr [[SRC]], i64 [[IV]]
-; CHECK-NEXT:    [[L:%.*]] = load float, ptr [[GEP]], align 4
+; CHECK-NEXT:  [[SCALAR_PH:.*]]:
+; CHECK-NEXT:    br label %[[LOOP1:.*]]
```
fhahn: please remove unrelated rename-only changes, that makes it harder to focus on the relevant changes

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
fhahn: > -    bool IsStrictPredicate = ICmpInst::isLT(Pred) || ICmpInst::isGT(Pred);
> +    bool IsStrictPredicate = CmpInst::isStrictPredicate(Pred);

Yes the problem is that currently find-first reductions with FP predicates are treated as find-last, so get miscompiled. We either need to update `handleFirstArgMinOrMax` to handle them properly, or bail out for strict floating point predicates until then. The later may be a better start, then add support fir find-first

llvm/test/Transforms/LoopVectorize/select-first-index-fp.ll
```diff
@@ -398,21 +398,56 @@ define i64 @test_fmaximumnum_first_idx(ptr %src, i64 %n) {
...
+; CHECK-NEXT:    [[TMP5:%.*]] = call float @llvm.vector.reduce.fmax.v4f32(<4 x float> [[TMP2]])
+; CHECK-NEXT:    [[BROADCAST_SPLATINSERT:%.*]] = insertelement <4 x float> poison, float [[TMP5]], i64 0
+; CHECK-NEXT:    [[BROADCAST_SPLAT:%.*]] = shufflevector <4 x float> [[BROADCAST_SPLATINSERT]], <4 x float> poison, <4 x i32> zeroinitializer
+; CHECK-NEXT:    [[TMP6:%.*]] = fcmp oeq <4 x float> [[TMP2]], [[BROADCAST_SPLAT]]
+; CHECK-NEXT:    [[TMP7:%.*]] = select <4 x i1> [[TMP6]], <4 x i64> [[TMP3]], <4 x i64> splat (i64 -9223372036854775808)
+; CHECK-NEXT:    [[TMP8:%.*]] = call i64 @llvm.vector.reduce.smax.v4i64(<4 x i64> [[TMP7]])
+; CHECK-NEXT:    [[TMP9:%.*]] = icmp ne i64 [[TMP8]], -9223372036854775808
+; CHECK-NEXT:    [[TMP10:%.*]] = select i1 [[TMP9]], i64 [[TMP8]], i64 0
+; CHECK-NEXT:    [[CMP_N:%.*]] = icmp eq i64 [[N]], [[N_VEC]]
+; CHECK-NEXT:    br i1 [[CMP_N]], label %[[EXIT:.*]], label %[[SCALAR_PH]]
```
fhahn: it looks like this is incorrectly using the find-last logic for a find-first reduction, as per https://github.com/llvm/llvm-project/pull/198300/changes#r3558830661

nasherm: Should be resolved in most recent patch

llvm/test/Transforms/LoopVectorize/select-first-index-fp.ll
nasherm: Done

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
nasherm: I'm not sure on the policy for adding `TODO`s comments, but my most recent patch does the bail out. Let me know if I should document that we should support FP some other way

nasherm:
Ping

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
fhahn: that's fine, thanks

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
```diff
@@ -2176,6 +2181,20 @@ bool VPlanTransforms::handleMultiUseReductions(VPlan &Plan,
         return Pred == CmpInst::ICMP_SLE || Pred == CmpInst::ICMP_SLT;
       case RecurKind::SMin:
         return Pred == CmpInst::ICMP_SGE || Pred == CmpInst::ICMP_SGT;
+      case RecurKind::FMaximum:
```
fhahn: Is this correct with NaNs?  I thought the NaN issue was already fixed, but I might be mis-remembering.

`llvm.minimum`/`llvm.maximum` return NaN if either compare is NaN. Consider something like `[ %x, nan, %y, nan] ` as input, with other input being all non-NaN.

In the scalar loop, the first compare will use non-NaN, all later compares in the scalar iterations will use NaN.

While in the vector loop the per-lane min/max may not have NaN in all lanes, the final reduced scalar min/max value will be NaN and then I think we may end up selecting an incorrect index.

It also looks like we do not have any tests for those code paths. I think we either need to require `nnan` or limit to `FMax` for now. Would be good to add a test case for the miscompile end-to-end to https://github.com/llvm/llvm-test-suite/pull/436 if there is indeed one.

madhur13490:
Hi @nasherm  — can we please move ahead on this? I’m working on down-counting, select-based FP argmin, and see overlap (notably the isMinMaxRecurrenceKind relaxation, plus the FP predicate cases / value-equality compare) in #209827.

nasherm:
> Hi @nasherm — can we please move ahead on this? I’m working on down-counting, select-based FP argmin, and see overlap (notably the isMinMaxRecurrenceKind relaxation, plus the FP predicate cases / value-equality compare) in #209827.

Hi @madhur13490 sorry for the delay. I've been on holiday for the past week and a half. I'll respond to Florian's comments and hopefully we can get this merged in soon

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
nasherm: Earlier changes dealt with `FMinNum/FMaxNum` and their special semantics for `sNaN` which is what you may be remembering.

fhahn: It looks like the update did not include any test changes. please make sure the code path is tested, to avoid rregressions/miscompiles in the future

nasherm: It appears my local changes needed a rebase

nasherm: I've extending the test cases. LMK if there's anything else I should cover

nasherm:
Is there any interest still in this patch? If not, I will just close it

llvm/test/Transforms/LoopVectorize/select-last-index-fp.ll
```diff
@@ -87,6 +87,50 @@ exit:
...
+  br label %loop
+
+loop:
+  %iv = phi i64 [ 0, %entry ], [ %iv.next, %loop ]
+  %max.idx = phi i64 [ 0, %entry ], [ %max.idx.next, %loop ]
+  %max.val = phi float [ 0xC7EFFFFFE0000000, %entry ], [ %max.val.next, %loop ]
+  %gep = getelementptr float, ptr %src, i64 %iv
+  %l = load float, ptr %gep
+  %cmp = fcmp nnan olt float %max.val, %l
+  %max.val.next = call nnan float @llvm.maxnum.f32(float %l, float %max.val)
```
fhahn: I think here we also need `nsz` for this to be recognized as `RecurKind::FMax`. 

I think the existing `test_fmax_last_idx` etc also miss this, which is why they are not vectorized with this patch. Could you add it to the tests (`test_fmax_last_idx` should be vectorized by the patch I think), and also add 2 new ones (one where `nsz` is missing and one where `nnan` is missing and that's the only reason blocking vectorization).

This should ensure we actually have the  `RecurKind::FMax` /  `RecurKind::FMin` code paths exercised in the tests

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
```diff
@@ -2241,7 +2259,13 @@ bool VPlanTransforms::handleMultiUseReductions(VPlan &Plan,
     VPBuilder B(FindIVRdxResult);
     VPValue *MinOrMaxExiting = MinOrMaxResult->getOperand(0);
     auto *FinalMinOrMaxCmp =
-        B.createICmp(CmpInst::ICMP_EQ, MinOrMaxExiting, MinOrMaxResult);
+        (RecurrenceDescriptor::isIntegerRecurrenceKind(RdxKind))
+            ? B.createICmp(CmpInst::ICMP_EQ, MinOrMaxExiting, MinOrMaxResult)
+            : B.createFCmp((RdxKind == RecurKind::FMaximum ||
+                            RdxKind == RecurKind::FMinimum)
+                               ? CmpInst::FCMP_UEQ
+                               : CmpInst::FCMP_OEQ,
```
fhahn:
```suggestion
            : B.createFCmp(CmpInst::FCMP_OEQ,
```

now that `RecurKind::FMaximum` is rejected

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
fhahn, line 2102: I think the last remaining potential mis-compile is `FMaximumNum`/`FMinimumNum` with `NaN` as start value. If the compare in the loop is ordered, I think the `NaN` handling of the intrinsics causes an issue.

Maybe best to bail out here if the start value for `FMaximumNum`/`FMinimumNum` may be `NaN` + test

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
nasherm: Done

llvm/test/Transforms/LoopVectorize/select-last-index-fp.ll
nasherm: Done

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
nasherm: Done

madhur13490 (approved):
This largely looks ok to me. If something is remaining, I will address in my patches mentioned before.

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
```diff
@@ -2194,6 +2212,16 @@ bool VPlanTransforms::handleMultiUseReductions(VPlan &Plan,
       return false;
     }
 
+    if (RdxKind == RecurKind::FMaximumNum ||
+        RdxKind == RecurKind::FMinimumNum) {
+      auto *StartValue = dyn_cast<VPIRValue>(MinOrMaxPhiR->getStartValue());
+      auto *StartC = StartValue
+                         ? dyn_cast<ConstantFP>(StartValue->getLiveInIRValue())
+                         : nullptr;
+      if (!StartC || StartC->isNaN())
```
fhahn: Can simplify this via `dyn_cast<VPConstant>` I think

fhahn (approved):
LGTM, thanks. 

I think the latest version should be handling the supported cases correctly.

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
nasherm: Done

PR #198302, opened 2026-05-18
[AArch64] Add missing FSub case to isLegalToVectorizeReduction

jacob-crawley:
Adds missing RecurKind::Fsub case to lower to partial reduction.


sdesmalen-arm (commented):
Can you add a corresponding test or modify an existing one to guard this behaviour?

MacDue (approved):
LGTM

PR #198303, opened 2026-05-18
[libc] Disable newly added -Wfenv-access clang warning

labath:
Added in #187860, the warning triggers on existing libc code. This unbreaks the precommit bots.


kaladron (approved):
Thank you!

PR #198304, opened 2026-05-18
[MISched] Make TG MacroFusion creation target specific(NFC)

tomershafir:
This patch moves TableGen-based macro fusion initialization from generic MachineScheduler factory methods, to the specific targets that use it, currently only RISCV. This better enables targets to have complete control over DAG mutation ordering, e.g. for porting other targets to TableGen macro fusion. More specifically, AArch64 requires the inverse partial order between ld/st clustering and macro fusion, as part of an attempt to migrate it to TableGen.


mshockwave (approved):
LGTM

wangpc-pp (approved):
LGTM!

Happy to see more users of TableGen-based macro fusions! :-)

PR #198307, opened 2026-05-18
[MLIR][XeGPU] Avoid chained-reductions in multi_reduction unrolling

dchigarev:
The PR adds a new unrolling pattern for `vector.multi_reduction` to the `xegpu-blocking` pass. In comparison with [the upstream reduction unrolling](https://github.com/llvm/llvm-project/blob/2da84a8307e4ef729458d990b221650a5da22639/mlir/lib/Dialect/Vector/Transforms/VectorUnroll.cpp#L372), the new pattern performs partial row-wise reductions via elementwise ops, instead of generating a chain of several multi-reduction ops:
```mlir
// reduction to unroll:
// tile-shape: [8x16]
vector.multi_reduction <add> %vec, %cst : vector<8x48xf32> to vector<8xf32>

// upstream unrolling:
%3 = vector.multi_reduction <add>, %tile_0, %cst [1] : vector<8x16xf32> to vector<8xf32>
%4 = vector.multi_reduction <add>, %tile_1, %3 [1] : vector<8x16xf32> to vector<8xf32>
%5 = vector.multi_reduction <add>, %tile_2, %4 [1] :  vector<8x16xf32> to vector<8xf32>

// new xegpu-unrolling
%3 = arith.addf %tile_0, %tile_1 : vector<8x16xf32>
%4 = arith.addf %3, %tile_2 : vector<8x16xf32>
%5 = vector.multi_reduction <add>, %4, %cst [1] : vector<8x16xf32> to vector<8xf32>
```

### DIFF
--- a/mlir/lib/Dialect/XeGPU/Transforms/XeGPUUnroll.cpp
+++ b/mlir/lib/Dialect/XeGPU/Transforms/XeGPUUnroll.cpp
@@ -806,7 +806,7 @@ struct UnrollConvertLayoutOp : public UnrollPattern<xegpu::ConvertLayoutOp> {
   }
 };
 
-/// Unrolls vector.multi_reduction by performing tree reduction with
+/// Unrolls vector.multi_reduction by sequentially reducing tiles with
 /// elementwise arith operations first, then a single multi_reduction
 /// per non-reduced tile position. This avoids generating long chains of
 /// multi_reduction ops (as the upstream pattern does) and is more efficient.
@@ -819,9 +819,6 @@ struct UnrollConvertLayoutOp : public UnrollPattern<xegpu::ConvertLayoutOp> {
 /// -- This pattern generates:
 /// %tmp1 = arith.reduction %tile0, %tile1 <32,32> -> <32x,2> // elementwise
 /// %res = vector.multi_reduction %tmp1, %zero_acc <32,32> to <32>
-///
-/// The patterns supports any-D vectors but only handles the case where there
-/// is a single reduction dimension that is the innermost dim.
 struct UnrollMultiReductionOp
     : public UnrollPattern<vector::MultiDimReductionOp> {
   UnrollMultiReductionOp(MLIRContext *context,
@@ -835,10 +832,10 @@ struct UnrollMultiReductionOp
     ArrayRef<int64_t> srcShape = srcTy.getShape();
     int64_t srcRank = srcTy.getRank();
 
-    // Only handle a single reduction dimension that is the last dim.
-    ArrayRef<int64_t> reductionDims = reductionOp.getReductionDims();
-    if (reductionDims.size() != 1 || reductionDims[0] != srcRank - 1)
-      return failure();
+    Location loc = reductionOp.getLoc();
+    Value source = reductionOp.getSource();
+    Value acc = reductionOp.getAcc();
+    vector::CombiningKind kind = reductionOp.getKind();
 
     // Result must be a vector (not scalar).
     auto resultType = dyn_cast<VectorType>(reductionOp.getDestType());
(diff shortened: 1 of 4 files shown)


mlir/lib/Dialect/XeGPU/Transforms/XeGPUUnroll.cpp
```diff
@@ -804,13 +806,152 @@ struct UnrollConvertLayoutOp : public UnrollPattern<xegpu::ConvertLayoutOp> {
...
+      // |---------------------------|---------------------|
+      // |       arith.op            |   arith.op          |
+      // |-------------------------------------------------|
+      // |                arith.op                         |
+      //                  ^---- this now goes to multi_reduction
+      while (tiles.size() > 1) {
+        SmallVector<Value> reduced;
+        for (size_t i = 0; i + 1 < tiles.size(); i += 2) {
+          Value combined = vector::makeArithReduction(rewriter, loc, kind,
+                                                      tiles[i], tiles[i + 1]);
```
dchigarev: this generates a suitable elementwise op (e.g. `arith.add`) for a reduction kind

mlir/lib/Dialect/XeGPU/Transforms/XeGPUUnroll.cpp
```diff
@@ -804,13 +806,152 @@ struct UnrollConvertLayoutOp : public UnrollPattern<xegpu::ConvertLayoutOp> {
...
+        static_cast<int64_t>(targetShapeOpt->size()) != srcRank)
+      return failure();
+
+    SmallVector<int64_t> targetShape = *targetShapeOpt;
+
+    // Check divisibility for all dimensions.
+    for (int64_t i = 0; i < srcRank; ++i) {
+      if (srcShape[i] % targetShape[i] != 0)
+        return failure();
+    }
```
dchigarev, line 858: could be a `llvm::all_of` as dictated by llvm's codestyle, but IMO it's much harder to read:
```c
if (!llvm::all_of(llvm::zip(srcShape, targetShape), [](auto pair) {
      return std::get<0>(pair) % std::get<1>(pair) == 0;
    }))
  return failure();
```

mlir/test/lib/Dialect/XeGPU/TestXeGPUTransforms.cpp
```diff
@@ -94,6 +94,19 @@ struct TestXeGPUUnrollingPatterns
       if (isa<xegpu::DpasOp>(op))
         return SmallVector<int64_t>{8, 16, 16};
 
+      // For vector.multi_reduction, read tile shape from the layout attribute
+      // on the source operand (layout_operand_0).
+      if (isa<vector::MultiDimReductionOp>(op)) {
```
dchigarev, line 99: The `test-xegpu-unrolling` pattern used to only support xegpu-ops. This change is to set proper 'target' shape from the reduction's layout

mlir/lib/Dialect/XeGPU/Transforms/XeGPUUnroll.cpp
```diff
@@ -804,13 +806,152 @@ struct UnrollConvertLayoutOp : public UnrollPattern<xegpu::ConvertLayoutOp> {
...
+/// vector.multi_reduction <32,64> to <32> (tile_shape=32, 32)
+/// -- Upstream pattern generates:
+/// %tmp1 = vector.multi_reduction %tile0, %zero_acc <32,32> to <32>
+/// %res = vector.multi_reduction %tmp1, %tile1 <32,32> to <32>
+/// -- This pattern generates:
+/// %tmp1 = arith.reduction %tile0, %tile1 <32,32> -> <32x,2> // elementwise
+/// %res = vector.multi_reduction %tmp1, %zero_acc <32,32> to <32>
+///
+/// The patterns supports any-D vectors but only handles the case where there
+/// is a single reduction dimension that is the innermost dim.
```
Jianhui-Li: ● The tree reduction structure is unnecessary here. Both sequential reduction (tile0 + 
 tile1 → tmp; tmp + tile2 → ...) and tree reduction ((tile0 + tile1), (tile2 + tile3) → 
  ...) require exactly N-1 elementwise operations. There is no saving in total step count —
   and since these ops execute sequentially on the hardware, there is no latency benefit
  either.

  The real issue with the upstream unrolling is the order of operations: it performs the
  in-tile reduction first (collapsing the reduction dimension within each tile via
  multi_reduction), then chains the cross-tile accumulation. This is expensive on GPU because
  in-tile reduction involves cross-lane shuffles, and doing it N times multiplies that
  cost.

  The fix should be simpler:

  1. Cross-tile reduction first: collect all tiles along the reduction dimension and
  combine them with elementwise arith ops (cheap, no shuffles). This collapses N tiles into
   a single tile per position.
  2. In-tile reduction last: perform one multi_reduction per position to eliminate the
  reduction dimension. This is the only step that requires cross-lane shuffles, and it
  happens exactly once per output tile.

  For multiple reduction dimensions, the same principle applies: reduce each reduction
  dimension down to tile size using elementwise ops first, then perform the final in-tile
  reduction to remove those dimensions entirely. In this way, we can also remove the 
 limitation of this PR (only handling the innermost dim reduction).

dchigarev: Right, replaced the "tree-reduction" with a sequential elementwise reduction.

Also added support for arbitrary reduction dimensions: the pattern now [collects tiles for <b>all</b> reduction dimensions](https://github.com/dchigarev/llvm-project/blob/67115b8c5a58eb4af94a709aa62627804800d9bb/mlir/lib/Dialect/XeGPU/Transforms/XeGPUUnroll.cpp#L899-L919) and [reduces them sequentially using elementwise arith op](https://github.com/dchigarev/llvm-project/blob/67115b8c5a58eb4af94a709aa62627804800d9bb/mlir/lib/Dialect/XeGPU/Transforms/XeGPUUnroll.cpp#L921-L925)

mlir/lib/Dialect/XeGPU/Transforms/XeGPUUnroll.cpp
```diff
@@ -804,13 +806,151 @@ struct UnrollConvertLayoutOp : public UnrollPattern<xegpu::ConvertLayoutOp> {
...
+/// per non-reduced tile position. This avoids generating long chains of
+/// multi_reduction ops (as the upstream pattern does) and is more efficient.
+///
+/// Example:
+/// vector.multi_reduction <32,64> to <32> (tile_shape=32, 32)
+/// -- Upstream pattern generates:
+/// %tmp1 = vector.multi_reduction %tile0, %zero_acc <32,32> to <32>
+/// %res = vector.multi_reduction %tmp1, %tile1 <32,32> to <32>
+/// -- This pattern generates:
+/// %tmp1 = arith.reduction %tile0, %tile1 <32,32> -> <32x,2> // elementwise
```
Jianhui-Li: <32x, 2> is a typo? 
  /// %tmp1 = arith.addf %tile0, %tile1 : vector<32x32xf32>  //  elementwise

mlir/lib/Dialect/XeGPU/Transforms/XeGPUUnroll.cpp
```diff
@@ -804,13 +806,151 @@ struct UnrollConvertLayoutOp : public UnrollPattern<xegpu::ConvertLayoutOp> {
...
+      // Generate the full tile indices for the reduced dimensions.
+      // Ex: if reduceDimShapes = [32, 64] and reducedDimTargetShapes = [16,
+      // 16], then reducedTileCoords:
+      // [(0, 0), (0, 1), (0, 2), (0, 3),
+      //  (1, 0), (1, 1), (1, 2), (1, 3)]
+      auto reducedTileCoords = StaticTileOffsetRange(
+          numReducedTilesPerDim, SmallVector<int64_t>(reducedDims.size(), 1));
+
+      // Step 1: Fill "blanks" in the offsets for the reduced dimensions
+      // using 'reducedTileCoords' and exctact according tiles.
```
Jianhui-Li: exctact -> extract

Jianhui-Li (approved):
LGTM

Jianhui-Li:
Please change the title also to remove tree-reduction. @dchigarev

Issue #185382, opened 2026-03-09
[CIR] Upstream handling of AArch64 (Arm) Neon builtins


banach-space:
### Overview

This is an umbrella issue for upstreaming all AArch64 builtins to ClangIR.

There are enough AArch64-specific builtins that creating separate issues for each logical group is not practical. Multiple contributors can work on this in parallel, provided we coordinate to minimize overlap. To help with that, a partial list of builtins that need to be upstreamed is included below.

If you would like to contribute, please comment indicating which builtin(s) you plan to work on.

Builtins will be removed from this list as they are upstreamed.

Before beginning work, please check the table below to ensure no one else is already working on the same intrinsics. If you have any questions, please comment here and tag @banach-space.

### Needed builtins

| Intrinsic Group | Common prefix | Assignee | PR(s) | Status |
|-------------------------|---------------|----------|-------|--------|
| [Set vector lane](https://arm-software.github.io/acle/neon_intrinsics/advsimd.html#set-vector-lane) | `vset_lane_*`, `vsetq_lane_*` |  |  | (*) `vset_lane_i8` ... |
| [Extract one element from vector](https://arm-software.github.io/acle/neon_intrinsics/advsimd.html#extract-one-element-from-vector) | `vget_lane_*`, `vgetq_lane_*` |  |  | (*) `vget_lane_i8` ... |
| [Vector saturating shift left](https://arm-software.github.io/acle/neon_intrinsics/advsimd.html#vector-saturating-shift-left) | `vqshl_*` |  |  | (*) `vqshlud_n_s64`, `vqshld_n_u64`, `vqshld_n_s64` |
| [Vector rounding shift right](https://arm-software.github.io/acle/neon_intrinsics/advsimd.html#vector-rounding-shift-right) | `vrshr_*` |  |  | (*) `vrshrd_n_u64`, `vrshrd_n_s64` |
| [Vector rounding shift right and accumulate](https://arm-software.github.io/acle/neon_intrinsics/advsimd.html#vector-rounding-shift-right-and-accumulate) | `vrsra_*` |  |  | (*) `vrsrad_n_u64`, `vrsrad_n_s64` |
| [Vector shift left](https://arm-software.github.io/acle/neon_intrinsics/advsimd.html#vector-shift-left) | `vshl_*` |  |  | (*) `vshld_n_s64`, `vshld_n_u64` |
[...]

PR #198309, opened 2026-05-18
[CIR][AArch64] Lower NEON vsli/vsliq intrinsics

E00N777:
### summary

part of: https://github.com/llvm/llvm-project/issues/185382

 Lower the AArch64 NEON shift-left-and-insert intrinsics (`vsli_n_v` / `vsliq_n_v`) in the CIR codegen path. The lowering mirrors classic CodeGen (`clang/lib/CodeGen/TargetBuiltins/ARM.cpp`): bitcast both vector operands to the target element type and emit a direct `llvm.aarch64.neon.vsli` intrinsic call.


E00N777:
I will implement `vslid_n_s64` and `vslid_n_u64` in another PR.

banach-space:
> The lowering mirrors classic CodeGen (`clang/lib/CodeGen/TargetBuiltins/ARM.cpp`): bitcast both vector operands to the target element type and emit a direct `llvm.aarch64.neon.vsli` intrinsic call.

I don't see any bitcasts in the original lowering: https://github.com/llvm/llvm-project/blob/e7887d5470e080c55fd69de83786fbf060d3955a/clang/lib/CodeGen/TargetBuiltins/ARM.cpp?plain=1#L6577-L6581

Why do we need them here? Could you check?

clang/test/CodeGen/AArch64/neon/intrinsics.c
```diff
@@ -5096,3 +5096,314 @@ uint64_t test_vaddlvq_u32(uint32x4_t a) {
...
   return vaddlvq_u32(a);
 }
+
+//===------------------------------------------------------===//
+// 2.1.3.1.6.  Vector shift left and insert
+// https://arm-software.github.io/acle/neon_intrinsics/advsimd.html#vector-shift-left-and-insert
+//===------------------------------------------------------===//
+
+// LLVM-LABEL: @test_vsli_n_s8(
+// CIR-LABEL: @test_vsli_n_s8(
```
banach-space: Here and in other places, you can re-use the label.
```suggestion
// ALL-LABEL: @test_vsli_n_s8(
```

banach-space (commented):
Thank you! Mostly looks good, but I wanted to clarify the need for bitcasts :)

E00N777:
> > The lowering mirrors classic CodeGen (`clang/lib/CodeGen/TargetBuiltins/ARM.cpp`): bitcast both vector operands to the target element type and emit a direct `llvm.aarch64.neon.vsli` intrinsic call.
> 
> I don't see any bitcasts in the original lowering:
> 
> https://github.com/llvm/llvm-project/blob/e7887d5470e080c55fd69de83786fbf060d3955a/clang/lib/CodeGen/TargetBuiltins/ARM.cpp?plain=1#L6577-L6581
> 
> Why do we need them here? Could you check?

All non-8-bit element intrinsics (like vsliq_n_u16) actually get bitcast to int8 first in arm_neon.h.  Because of this, I need to cast them back, which is why there are two groups of bitcasts in my check lines.

banach-space:
> All non-8-bit element intrinsics (like vsliq_n_u16) actually get bitcast to int8 first in arm_neon.h. Because of this, I need to cast them back, which is why there are two groups of bitcasts in my check lines.

That's rather counter-intuitive to me and feels a bit off :) 

In `CIRGenBuiltinAArch64.cpp`, you only match `NEON::BI__builtin_neon_vsriq_n_v` and that's the only thing that _that_ file should be concerned about.

You are probably thinking in terms of an example like this:
```c
int16x4_t test_vsli_n_s16(int16x4_t a, int16x4_t b) {
    return vsli_n_s16(a, b, 3);
  }
```

and the definition of `vsli_n_s16` from arm_neon.h:
```c
  #define vsli_n_s16(__p0, __p1, __p2) __extension__ ({ \
    int16x4_t __ret; \
    int16x4_t __s0 = __p0; \
    int16x4_t __s1 = __p1; \
    __ret = __builtin_bit_cast(int16x4_t, __builtin_neon_vsli_n_v(__builtin_bit_cast(int8x8_t, __s0), __builtin_bit_cast(int8x8_t, __s1), __p2, 1)); \
    __ret; \
  })
```

So yes, there are those `_builtin_bit_cast`, but `CIRGenBuiltinAArch64.cpp` only matches `NEON::BI__builtin_neon_vsriq_n_v` and that's the only thing that it should be lowering. Bitcasts ought to be handled by some other logic elsewhere.

Have you tried removing the bit-casts? Do tests fail after that?

E00N777:
> > All non-8-bit element intrinsics (like vsliq_n_u16) actually get bitcast to int8 first in arm_neon.h. Because of this, I need to cast them back, which is why there are two groups of bitcasts in my check lines.
> 
> That's rather counter-intuitive to me and feels a bit off :)
> 
> In `CIRGenBuiltinAArch64.cpp`, you only match `NEON::BI__builtin_neon_vsriq_n_v` and that's the only thing that _that_ file should be concerned about.
> 
> You are probably thinking in terms of an example like this:
> 
> ```c
> int16x4_t test_vsli_n_s16(int16x4_t a, int16x4_t b) {
>     return vsli_n_s16(a, b, 3);
>   }
> ```
> 
> and the definition of `vsli_n_s16` from arm_neon.h:
> 
> ```c
>   #define vsli_n_s16(__p0, __p1, __p2) __extension__ ({ \
>     int16x4_t __ret; \
>     int16x4_t __s0 = __p0; \
>     int16x4_t __s1 = __p1; \
>     __ret = __builtin_bit_cast(int16x4_t, __builtin_neon_vsli_n_v(__builtin_bit_cast(int8x8_t, __s0), __builtin_bit_cast(int8x8_t, __s1), __p2, 1)); \
>     __ret; \
>   })
> ```
> 
> So yes, there are those `_builtin_bit_cast`, but `CIRGenBuiltinAArch64.cpp` only matches `NEON::BI__builtin_neon_vsriq_n_v` and that's the only thing that it should be lowering. Bitcasts ought to be handled by some other logic elsewhere.
> 
> Have you tried removing the bit-casts? Do tests fail after that?

Yes, I tried removing the bit-cast for these non-8-byte elements, but the tests failed — maybe I overlooked something. I'll revisit it this week and write a more detailed summary so it's easier for you to review.

banach-space:
> Yes, I tried removing the bit-cast for these non-8-byte elements, but the tests failed — maybe I overlooked something. I'll revisit it this week and write a more detailed summary so it's easier for you to review.

Which test failed? I commented the bitcasts out myself and everything worked 🤔 It's possible that I am doing something wrong or missing something. Let's get to the bottom of this mystery!

E00N777:
> > Yes, I tried removing the bit-cast for these non-8-byte elements, but the tests failed — maybe I overlooked something. I'll revisit it this week and write a more detailed summary so it's easier for you to review.
> 
> Which test failed? I commented the bitcasts out myself and everything worked 🤔 It's possible that I am doing something wrong or missing something. Let's get to the bottom of this mystery!

Sorry for the late reply.
I previously misunderstood your suggestion.in the lowering logic in `llvm-project/clang/lib/CIR/CodeGen/CIRGenBuiltinAArch64.cpp`,those two line `bitcast` is indeed no-need. I mistakenly thought earlier that you were asking about the `bitcast` in the checkline. I'm really sorry.

banach-space:
> Sorry for the late reply. I previously misunderstood your suggestion.in the lowering logic in `llvm-project/clang/lib/CIR/CodeGen/CIRGenBuiltinAArch64.cpp`,those two line `bitcast` is indeed no-need. I mistakenly thought earlier that you were asking about the `bitcast` in the checkline. I'm really sorry.

Don't worry - these things are quite fiddly and nuanced! The point of the review is to clarify any confusion and its natural that it sometimes takes time. And I'd rather we do it right rather than doing it quick :)

banach-space (approved):
LGTM, thank you!

PR #198314, opened 2026-05-18
[AArch64][GlobalISel] Add sign bits for G_FCMEQ

davemgreen:
This adds basic num-sign-bits for G_FCMEQ, G_FCMGE and G_FCMGT, which all produce either all-ones or all-zeros in each vector lane. This function apparently goes in AArch64ISelLowering.


c-rhodes (approved):
cant say i know what it's for, but looks fine to me

PR #198317, opened 2026-05-18
[AsmWriter] Add emitMDNodeAnnot

hvdijk:
Similarly to the other AAW::emit*Annot functions, this may be used to emit a comment before a metadata node.


hvdijk:
The motivation for this is #198318, which uses this new function.

hvdijk:
I'm not sure who'd be good to review this, the whole `AssemblyAnnotationWriter` is functionally unchanged since 2010, but I guess this counts as just the generic "Parts of code generator not covered by someone else" so requesting review accordingly.

arsenm (commented):
Testcase?

hvdijk:
A NFC change cannot have a testcase, but the dependent PR #198318 adds functionality along with a test that uses this.

arsenm:
> A NFC change cannot have a testcase, but the dependent PR #198318 adds functionality along with a test that uses this.

I wouldn't call this NFC

hvdijk:
> I wouldn't call this NFC

It's not NFC for LLVM-as-a-library but it's NFC for all of the LLVM command-line utilities. It's annoyed me for years that's been the standard by which we judge what is and isn't NFC, and if you want to propose a change to that in LLVM, that will have my full support, but for now that's the standard by which I have to judge it.

arsenm:
> > I wouldn't call this NFC
> 
> It's not NFC for LLVM-as-a-library but it's NFC for all of the LLVM command-line utilities. It's annoyed me for years that's been the standard by which we judge what is and isn't NFC, 

I'm not sure what standard you mean, but this is adding new functionality that will later be used.

IMO we should just drop the convention of using NFC. It's misused most of the time

arsenm (approved):

hvdijk:
> I'm not sure what standard you mean, but this is adding new functionality that will later be used.

I mean when only this PR on its own is merged, no behaviour of any command-line utility changes, and that's why there is no test in this PR. That's the rationale for why many prior commits have been labelled as NFC, which subsequently ended up breaking downstream code because they by design changed the public API of libLLVM and should, in my opinion, absolutely not have been labelled NFC.

> IMO we should just drop the convention of using NFC. It's misused most of the time

That sounds fine to me, I'll remove [NFC] from the title.

I see CI is failing after the merge of `main`, it was a merge of a point where `main` was broken, merging again to confirm that CI will pass after that.

hvdijk:
The remaining CI failure is the same problem that @jhuber6 fixed in #201367, but still persists in another test. It is definitely not related to this PR, so merging. @jhuber6 Can you fix that other test too, or should I?

jhuber6:
> The remaining CI failure is the same problem that @jhuber6 fixed in #201367, but still persists in another test. It is definitely not related to this PR, so merging. @jhuber6 Can you fix that other test too, or should I?

Somehow missed that one, sorry. I'm fixing something else so my build is in a weird state. if you can fix it real quick I'd appreciate it.

hvdijk:
Sure, will do as soon as my local build finishes, unless someone has already opened a PR by then. Edit: done, #201382.

PR #198318, opened 2026-05-18
[DirectX] Move IR printing to DXILPrettyPrinter

hvdijk:
By doing the IR printing inside DXILPrettyPrinter, we have the option to customise what we print and include the info that we collect and generate in DXILDebugInfo.


llvm/lib/Target/DirectX/DXILPrettyPrinter.cpp
```diff
@@ -251,12 +259,92 @@ static void prettyPrintResources(raw_ostream &OS, const DXILResourceMap &DRM,
...
+  DXILAssemblyAnnotationWriter DAAW(MST, *STS, DI);
+  M.print(FOS, &DAAW);
+
+  ModuleSlotTracker::MachineMDNodeListType MDNodes;
+  MST.collectMDNodes(MDNodes, NextMetadataSlot, ~0u);
+  std::sort(
+      MDNodes.begin(), MDNodes.end(),
+      [](std::pair<unsigned, const MDNode *> &A,
+         std::pair<unsigned, const MDNode *> &B) { return A.first < B.first; });
+  for (auto [_, MDNode] : MDNodes) {
```
bogner, line 345: I think this should be `const auto &` to avoid a copy (or at least `auto &`)
```suggestion
  for (const auto &[_, MDNode] : MDNodes) {
```

llvm/lib/Target/DirectX/DXILPrettyPrinter.cpp
```diff
@@ -251,12 +259,92 @@ static void prettyPrintResources(raw_ostream &OS, const DXILResourceMap &DRM,
...
+
+  DXILAssemblyAnnotationWriter DAAW(MST, *STS, DI);
+  M.print(FOS, &DAAW);
+
+  ModuleSlotTracker::MachineMDNodeListType MDNodes;
+  MST.collectMDNodes(MDNodes, NextMetadataSlot, ~0u);
+  std::sort(
+      MDNodes.begin(), MDNodes.end(),
+      [](std::pair<unsigned, const MDNode *> &A,
+         std::pair<unsigned, const MDNode *> &B) { return A.first < B.first; });
```
bogner: Can save a few characters with `llvm::sort` and/or `auto` for the lambda arguments. In any case, these args can/should be `const`:
```suggestion
  llvm::sort(MDNodes,
             [](const auto &A, const auto &B) { return A.first < B.first; });
```

llvm/lib/Target/DirectX/DXILPrettyPrinter.cpp
```diff
@@ -268,13 +356,10 @@ class DXILPrettyPrinterLegacy : public llvm::ModulePass {
...
   explicit DXILPrettyPrinterLegacy(raw_ostream &O) : ModulePass(ID), OS(O) {}
 
-  StringRef getPassName() const override {
-    return "DXIL Metadata Pretty Printer";
-  }
+  StringRef getPassName() const override { return "DXIL Pretty Printer"; }
 
   bool runOnModule(Module &M) override;
   void getAnalysisUsage(AnalysisUsage &AU) const override {
-    AU.setPreservesAll();
```
bogner, line 277: It doesn't really matter since this is the last thing we do, but I don't see why it wouldn't preserveAll any more

bogner (commented):
Seems reasonable. Do you mind pasting the output of this for a fairly simple shader here? It's hard to see how this is going to look from the tests alone.

llvm/lib/Target/DirectX/DXILPrettyPrinter.cpp
hvdijk: The types here are `unsigned` and `const MDNode *`, these aren't types we need to avoid copying, if it makes any difference at all it'll be more expensive to add an extra indirection here.

llvm/lib/Target/DirectX/DXILPrettyPrinter.cpp
hvdijk: I actually had `auto` initially but removed it based on your review comment on the other PR that `auto` should be avoided where the types are not obvious, and here in my opinion the type is not obvious. Definitely agreed on the `const` though.

llvm/lib/Target/DirectX/DXILPrettyPrinter.cpp
hvdijk: We run `DXILDebugInfoPass::run(M)` now, which modifies the module.

llvm/lib/Target/DirectX/DXILPrettyPrinter.cpp
bogner: Fair (the fact that it obscures things like this are what I dislike about auto)

bogner (approved):

hvdijk:
> Seems reasonable. Do you mind pasting the output of this for a fairly simple shader here? It's hard to see how this is going to look from the tests alone.

After the update to account for #197478 (which added additional information to DebugInfoMap), for `llvm/test/CodeGen/DirectX/debug-info.ll` which contains a simple (bad) FMA implementation, the output of `llc --filetype=asm <debug-info.ll` is

```llvm
;
; Resource Bindings:
;
; Name                                 Type  Format         Dim      ID      HLSL Bind     Count
; ------------------------------ ---------- ------- ----------- ------- -------------- ---------
;
; ModuleID = '<stdin>'
source_filename = "<stdin>"
target datalayout = "e-m:e-ve-p:32:32-i1:32-i8:8-i16:16-i32:32-i64:64-f16:16-f32:32-f64:64-n8:16:32:64"
target triple = "dxil-unknown-shadermodel6.7-library"

; Function Attrs: nounwind memory(none)
define dso_local float @fma(float %0, float %1, float %2) local_unnamed_addr #0 !dbg !9 {
; DXIL: to be replaced with:   tail call addrspace(0) void @llvm.dbg.value(metadata float %0, i64 0, metadata !14, metadata !DIExpression()), !dbg !17
  tail call void @llvm.dbg.value(metadata float %0, metadata !14, metadata !DIExpression()), !dbg !17
; DXIL: to be replaced with:   tail call addrspace(0) void @llvm.dbg.value(metadata float %1, i64 0, metadata !15, metadata !DIExpression()), !dbg !17
  tail call void @llvm.dbg.value(metadata float %1, metadata !15, metadata !DIExpression()), !dbg !17
; DXIL: to be replaced with:   tail call addrspace(0) void @llvm.dbg.value(metadata float %2, i64 0, metadata !16, metadata !DIExpression()), !dbg !17
  tail call void @llvm.dbg.value(metadata float %2, metadata !16, metadata !DIExpression()), !dbg !17
  %4 = fmul float %0, %1, !dbg !18
  %5 = fadd float %4, %2, !dbg !19
  ret float %5, !dbg !20
}

; Function Attrs: nounwind memory(none)
declare void @llvm.dbg.value(metadata, metadata, metadata) #0

attributes #0 = { nounwind memory(none) }

!llvm.dbg.cu = !{!0}
!llvm.ident = !{!3}
!dx.shaderModel = !{!4}
!dx.version = !{!5}
!dx.entryPoints = !{!6}
!llvm.module.flags = !{!7, !8}

; DXIL: !0: additional data: !21
!0 = distinct !DICompileUnit(language: DW_LANG_C99, file: !1, isOptimized: true, runtimeVersion: 0, emissionKind: FullDebug, enums: !2, splitDebugInlining: false, nameTableKind: None)
!1 = !DIFile(filename: "in.c", directory: "dir")
!2 = !{}
!3 = !{!"Some Compiler"}
!4 = !{!"lib", i32 6, i32 7}
!5 = !{i32 1, i32 7}
!6 = !{null, !"", null, null, null}
!7 = !{i32 2, !"Dwarf Version", i32 2}
!8 = !{i32 2, !"Debug Info Version", i32 3}
; DXIL: !9: to be replaced by: !22
!9 = distinct !DISubprogram(name: "fma", scope: !1, file: !1, line: 1, type: !10, scopeLine: 1, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition | DISPFlagOptimized, unit: !0, retainedNodes: !13)
!10 = !DISubroutineType(types: !11)
!11 = !{!12, !12, !12, !12}
!12 = !DIBasicType(name: "float", size: 32, encoding: DW_ATE_float)
!13 = !{!14, !15, !16}
!14 = !DILocalVariable(name: "x", arg: 1, scope: !9, file: !1, line: 1, type: !12)
!15 = !DILocalVariable(name: "y", arg: 2, scope: !9, file: !1, line: 1, type: !12)
!16 = !DILocalVariable(name: "z", arg: 3, scope: !9, file: !1, line: 1, type: !12)
!17 = !DILocation(line: 0, scope: !9)
!18 = !DILocation(line: 2, column: 12, scope: !9)
!19 = !DILocation(line: 2, column: 16, scope: !9)
!20 = !DILocation(line: 2, column: 3, scope: !9)
!21 = !{!9}
; DXIL: !22: additional data: ptr @fma
!22 = !DISubprogram(name: "fma", scope: !1, file: !1, line: 1, type: !10, scopeLine: 1, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition | DISPFlagOptimized, unit: !0, retainedNodes: !13)
```

llvm/test/CodeGen/DirectX/DebugInfo/di-subprogram.ll
```diff
@@ -0,0 +1,60 @@
+; RUN: llc --filetype=asm %s -o - | FileCheck %s
```
arsenm, line 1:
```suggestion
; RUN: llc %s -o - | FileCheck %s
```

This is the default

hvdijk: Good idea, though most other DirectX tests do also explicitly specify `--filetype=asm`. @bogner What do you prefer here? I'd like it to be consistent across tests, shall I apply this change and then submit another PR to update the other tests to match? Or do it for this one test only? Or leave it explicitly using `--filetype=asm`?

hvdijk: In the interest of speeding things along, I'll just go ahead merge this PR as is and open a new PR to apply this suggestion to all applicable tests

PR #198319, opened 2026-05-18
[X86] Add handling for sub-128bit minmax reductions

RKSimon:
Fold sub-128bit minmax reductions as ISD::VECREDUCE nodes.

This needed some cleanup to correct discard "identity value padded" upper elements from legalisation - existing folds struggle to do this due to the DemandedElts mask needing to be accurate enough (and reductions nearly always result in multiple uses of source operands) - I've been trying to do something similar in TargetLowering::expandVecReduce but haven't managed it yet.

Final backend patch blocking #194473 - still some middle-end reduction pattern matching that needs some fixing first though.


phoebewang (approved):
LGTM.

PR #198320, opened 2026-05-18
[AArch64][DAG] Use getSetCCResultType in PromoteSETCC

davemgreen:
During the promotion of a v16bf16 vector setcc with 256 bit sve, a v16i16 setcc(v16f32) is created, that gets legalized to a v8i16 setcc(v8f32), which trips up an assert in the aarch64 custom lowering to vector registers. It could be solved elsewhere but this addresses it by making sure that when the f32 setcc is created, the result type matches getSetCCResultType.

The mvn+xtn vs xtn+mvn is mostly benign, and should both probably be a subhn if it can be done without causing regressions.

### DIFF
--- a/llvm/lib/CodeGen/SelectionDAG/LegalizeVectorOps.cpp
+++ b/llvm/lib/CodeGen/SelectionDAG/LegalizeVectorOps.cpp
@@ -657,9 +657,12 @@ void VectorLegalizer::PromoteSETCC(SDNode *Node,
     Operands[4] = Node->getOperand(4); // evl
   }
 
-  SDValue Res = DAG.getNode(Node->getOpcode(), DL, Node->getSimpleValueType(0),
-                            Operands, Node->getFlags());
-
+  EVT ResVT =
+      TLI.getSetCCResultType(DAG.getDataLayout(), *DAG.getContext(), NewVecVT);
+  SDValue Res =
+      DAG.getNode(Node->getOpcode(), DL, ResVT, Operands, Node->getFlags());
+  if (ResVT != Node->getSimpleValueType(0))
+    Res = DAG.getSExtOrTrunc(Res, DL, Node->getSimpleValueType(0));
   Results.push_back(Res);
 }
 
--- a/llvm/lib/CodeGen/SelectionDAG/LegalizeVectorOps.cpp
+++ b/llvm/lib/CodeGen/SelectionDAG/LegalizeVectorOps.cpp
@@ -662,7 +662,7 @@ void VectorLegalizer::PromoteSETCC(SDNode *Node,
   SDValue Res =
       DAG.getNode(Node->getOpcode(), DL, ResVT, Operands, Node->getFlags());
   if (ResVT != Node->getSimpleValueType(0))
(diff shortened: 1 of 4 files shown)


llvm/lib/CodeGen/SelectionDAG/LegalizeVectorOps.cpp
```diff
@@ -657,9 +657,12 @@ void VectorLegalizer::PromoteSETCC(SDNode *Node,
...
 
-  SDValue Res = DAG.getNode(Node->getOpcode(), DL, Node->getSimpleValueType(0),
-                            Operands, Node->getFlags());
-
+  EVT ResVT =
+      TLI.getSetCCResultType(DAG.getDataLayout(), *DAG.getContext(), NewVecVT);
+  SDValue Res =
+      DAG.getNode(Node->getOpcode(), DL, ResVT, Operands, Node->getFlags());
+  if (ResVT != Node->getSimpleValueType(0))
+    Res = DAG.getSExtOrTrunc(Res, DL, Node->getSimpleValueType(0));
```
arsenm: getBoolExt

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorOps.cpp
```diff
@@ -657,9 +657,12 @@ void VectorLegalizer::PromoteSETCC(SDNode *Node,
...
   }
 
-  SDValue Res = DAG.getNode(Node->getOpcode(), DL, Node->getSimpleValueType(0),
-                            Operands, Node->getFlags());
-
+  EVT ResVT =
+      TLI.getSetCCResultType(DAG.getDataLayout(), *DAG.getContext(), NewVecVT);
+  SDValue Res =
+      DAG.getNode(Node->getOpcode(), DL, ResVT, Operands, Node->getFlags());
+  if (ResVT != Node->getSimpleValueType(0))
```
topperc: Should this be getValueType instead of getSimpleValueType?

davemgreen: Oh yeah. I hadn't realised it was using that.

topperc: Should the other getSimpleValueType(0) be changed too?

llvm/lib/CodeGen/SelectionDAG/LegalizeVectorOps.cpp
```diff
@@ -657,9 +657,12 @@ void VectorLegalizer::PromoteSETCC(SDNode *Node,
     Operands[4] = Node->getOperand(4); // evl
   }
 
-  SDValue Res = DAG.getNode(Node->getOpcode(), DL, Node->getSimpleValueType(0),
-                            Operands, Node->getFlags());
-
+  EVT ResVT =
```
topperc, line 660: Not related to this patch, but it seems wrong that line 646 uses ANY_EXTEND for integer. I hope no target uses this for integers.

topperc (approved):
LGTM

RKSimon (approved):
LGTM

PR #198323, opened 2026-05-18
[libunwind] DwarfFDECache::add: bail out if malloc() returns NULL

stepanof:
DwarfFDECache::add() did not check the result of malloc().
On allocation failure, subsequent memcpy() and pointer arithmetic operate on a NULL buffer, resulting in undefined behavior.

In low-memory embedded environments where .text may be mapped at address 0, this can silently corrupt code at low addresses.

Skip caching the entry on allocation failure instead, releasing the lock first. The cache then degrades gracefully - subsequent FDE lookups remain correct, just slower for repeated queries.

### DIFF
--- a/libunwind/src/UnwindCursor.hpp
+++ b/libunwind/src/UnwindCursor.hpp
@@ -203,24 +203,29 @@ void DwarfFDECache<A>::add(pint_t mh, pint_t ip_start, pint_t ip_end,
     size_t newSize = oldSize * 4;
     // Can't use operator new (we are below it).
     entry *newBuffer = (entry *)malloc(newSize * sizeof(entry));
-    memcpy(newBuffer, _buffer, oldSize * sizeof(entry));
-    if (_buffer != _initialBuffer)
-      free(_buffer);
-    _buffer = newBuffer;
-    _bufferUsed = &newBuffer[oldSize];
-    _bufferEnd = &newBuffer[newSize];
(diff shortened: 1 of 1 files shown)


libunwind/src/UnwindCursor.hpp
```diff
@@ -203,6 +203,12 @@ void DwarfFDECache<A>::add(pint_t mh, pint_t ip_start, pint_t ip_end,
     size_t newSize = oldSize * 4;
     // Can't use operator new (we are below it).
     entry *newBuffer = (entry *)malloc(newSize * sizeof(entry));
+    if (newBuffer == NULL) {
+      // Out of memory: skip caching this entry rather than corrupting
+      // memory by writing through a NULL _buffer.
+      _LIBUNWIND_LOG_IF_FALSE(_lock.unlock());
```
jroelofs: This skips the dyld unload in `__APPLE__`

stepanof: Refactored to a single exit point.  
Buffer growth now happens only `if (newBuffer != NULL)`.  
The entry insertion and the `__APPLE__` dyld unload are guarded by `if (_bufferUsed < _bufferEnd)` (space available)

MaskRay (approved):
LGTM but please wait for @jroelofs

stepanof:
ping @jroelofs

PR #198326, opened 2026-05-18
[MLIR][PDL] Add debug printing for PDL rewrites in PatternApplicator

jorickert:
Previously this was only done for not-pdl patterns


jorickert:
@joker-eph It would be great if you could review or assign someone. I do not have the rights for that.

Thank you

PR #198328, opened 2026-05-18
[TableGen] Fix getting weights of register classes

kosarev:
The first member can be an aritifical register, so we have to find a non-artificial one to query its weight.

### DIFF
--- a/llvm/utils/TableGen/Common/CodeGenRegisters.cpp
+++ b/llvm/utils/TableGen/Common/CodeGenRegisters.cpp
@@ -833,10 +833,15 @@ unsigned CodeGenRegisterClass::getWeight(const CodeGenRegBank &RegBank) const {
   if (TheDef && !TheDef->isValueUnset("Weight"))
     return TheDef->getValueAsInt("Weight");
 
-  if (Members.empty() || Artificial)
+  if (Artificial)
     return 0;
 
-  return (*Members.begin())->getWeight(RegBank);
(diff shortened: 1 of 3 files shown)


arsenm (approved):

llvm/utils/TableGen/Common/CodeGenRegisters.cpp
```diff
@@ -833,10 +833,15 @@ unsigned CodeGenRegisterClass::getWeight(const CodeGenRegBank &RegBank) const {
   if (TheDef && !TheDef->isValueUnset("Weight"))
     return TheDef->getValueAsInt("Weight");
 
-  if (Members.empty() || Artificial)
+  if (Artificial)
     return 0;
 
-  return (*Members.begin())->getWeight(RegBank);
```
Sisyph, line 839: Is the artificial register guaranteed to be the first one in Members? The relevance is: does does the test cover the behavior change, and if so by accident or guaranteed to cover it? Otherwise LGTM

kosarev: > Is the artificial register guaranteed to be the first one in Members?

Yes, based on its name.

> does the test cover the behavior change

Yes.

Sisyph: Great, thanks.

Sisyph (approved):

PR #198330, opened 2026-05-18
[SPIR-V] Fix legalized load of single-element vector from array

aobolensk:
OpCompositeInsert into a scalar is invalid SPIR-V, and `<1 x T>` is collapsed to `T` in the global registry. Emit `spv_bitcast` in place of `spv_insertelt` for the `N == 1` case so the wrapper lowers to a simple copy

Enable the previously-disabled 1x1 matrix transpose test

### DIFF
--- a/llvm/lib/Target/SPIRV/SPIRVLegalizePointerCast.cpp
+++ b/llvm/lib/Target/SPIRV/SPIRVLegalizePointerCast.cpp
@@ -237,6 +237,16 @@ class SPIRVLegalizePointerCastImpl {
   Value *
   buildVectorFromLoadedElements(IRBuilder<> &B, FixedVectorType *TargetType,
                                 SmallVector<Value *, 4> &LoadedElements) {
+    // <1 x T> shares the SPIR-V type with T, so emitting OpCompositeInsert on
+    // a scalar would be invalid. Bridge with spv_bitcast instead.
+    if (TargetType->getNumElements() == 1) {
+      Value *Scalar = LoadedElements[0];
+      Value *NewVector = B.CreateIntrinsic(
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/SPIRV/SPIRVLegalizePointerCast.cpp
```diff
@@ -237,6 +237,16 @@ class SPIRVLegalizePointerCastImpl {
...
                                 SmallVector<Value *, 4> &LoadedElements) {
+    // <1 x T> shares the SPIR-V type with T, so emitting OpCompositeInsert on
+    // a scalar would be invalid. Bridge with spv_bitcast instead.
+    if (TargetType->getNumElements() == 1) {
+      Value *Scalar = LoadedElements[0];
+      Value *NewVector = B.CreateIntrinsic(
+          Intrinsic::spv_bitcast, {TargetType, Scalar->getType()}, {Scalar});
+      buildAssignType(B, TargetType, NewVector);
+      return NewVector;
+    }
```
MrSidims, line 248: Do i understand correctly, that what this change affectively doing is reverting mapping from <1 x T> to T for a single case?

aobolensk: We can look at it this way

MrSidims (approved):

llvm/lib/Target/SPIRV/SPIRVLegalizePointerCast.cpp
MrSidims: May be the cleanest way would be not to do such transformation in such case in the first place, but the current patch is fine with me.

PR #198333, opened 2026-05-18
[llvm][support] Implement `sys::path::starts_with()`

jansvoboda11:
This PR adds a new function that checks whether a path starts with a specified prefix. Unlike `StringRef::starts_with()`, this requires an exact match of the last component, and ignores case and separator differences on Windows.

This is adopted by Clang's dependency scanner, where the iterator-based prefix matching showed up in the traces. This speeds up scans with populated and validated `InProcessModuleCache` by roughly 5%. Note that these scans are dominated by the `ModuleDepCollector` logic, not actually building modules or performing header search.

### DIFF
--- a/clang/include/clang/DependencyScanning/DependencyScannerImpl.h
+++ b/clang/include/clang/DependencyScanning/DependencyScannerImpl.h
@@ -107,12 +107,12 @@ void initializeScanCompilerInstance(
     DiagnosticConsumer *DiagConsumer, DependencyScanningService &Service,
     IntrusiveRefCntPtr<DependencyScanningWorkerFilesystem> DepFS);
 
-SmallVector<StringRef>
+SmallVector<SmallString<0>, 2>
 getInitialStableDirs(const CompilerInstance &ScanInstance);
 
 std::optional<PrebuiltModulesAttrsMap>
 computePrebuiltModulesASTMap(CompilerInstance &ScanInstance,
-                             SmallVector<StringRef> &StableDirs);
+                             ArrayRef<SmallString<0>> StableDirs);
 
 /// Create the dependency collector that will collect the produced
 /// dependencies. May return the created ModuleDepCollector depending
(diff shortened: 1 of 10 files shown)


jansvoboda11:
I'm not 100% sure how I feel about ignoring case differences on Windows, that part was carried over from the old function. WDYT?

benlangmuir (commented):
This changes behaviour for paths containing redundant separators, which `path::const_iterator::operator++` will skip over even for posix paths. I think it also changes if `Prefix` has a trailing separator, where the old code would require the other path to contain a component ".", but that seems like a bug in the old code.  And like you mentioned already, by adding the case-insensitivity logic from the other function, this makes the code more relaxed on windows than it was before.

I would be inclined to drop the case-sensitivity part in this function.  In general, I am very skeptical of handling path case-sensitivity outside the OS unless you are willing to handle platform- and filesystem-specific behaviours on a per-path basis. Windows allows case-sensitive paths on a per-directory basis. MacOS has case-insensitive paths by default, but it's not rare to have case-sensitive filesystems mounted. Whether it's desirable for `replace_path_prefix` to have this is a separate issue we don't need to answer in this PR.

jansvoboda11:
Good point on case sensitivity, I'll remove that from the new function.

I didn't realize redundant path separators are typically ignored. I'll try to handle that in the slow path of the function.

benlangmuir:
For what it's worth, I don't know how important skipping redundant separators actually is; I just wanted to make sure we're aware of the change.  You could imagine running `remove_dots` (only the one-dot version) as well, which would let you treat `/prefix/.//a` as a prefix for `/prefix/a/b`, but then it's not clear why that would be important but not important to allows symlinks, etc.

jansvoboda11:
Good point. For some reason, I didn't consider these non-canonical paths at all. Thinking about it some more, I think we could add a precondition that both `Path` and `Prefix` are normalized (i.e. `assert(!remove_dots(X, /*remove_dot_dot=*/false))`. This already holds true for the paths in the dependency scanner, we just need to normalize the two prefixes, which should be cheap enough.

jansvoboda11:
I implemented the version that asserts on path and prefix being normalized. If this looks fine to you, I'll pull out the plumbing changes in the dependency scanner changes into a separate PR.

benlangmuir (commented):
I don't love the repetition of `SmallVector<SmallString<0>, 2>` everywhere. How about keeping StringRef and adding a StringSaver argument only where it needs to be appended to?

PR #198335, opened 2026-05-18
[flang] RISCV64 (lp64/lp64d) support for BIND(C) derived types

prados-oc:
This implements the RISCV64 lp64/lp64d ABI according to: https://github.com/riscv-non-isa/riscv-elf-psabi-doc

It doesn't implement the ABI from scratch, but mimics the way clang does things and lets the RISCV backend handle some lowering nuances. This is wanted for RISCV according to 
https://github.com/llvm/llvm-project/blob/673b17e1b38fbe4dfe01c9aa66c368857b13dfdb/llvm/lib/Target/RISCV/RISCVCallingConv.cpp#L58

This was verified against real world applications that use bind(c) and they still work as expected. 

NOTE: This work is partly based on the original implementation from @rofirrim

### DIFF
--- a/flang/lib/Optimizer/CodeGen/Target.cpp
+++ b/flang/lib/Optimizer/CodeGen/Target.cpp
@@ -1389,10 +1389,18 @@ struct TargetSparcV9 : public GenericTarget<TargetSparcV9> {
 //===----------------------------------------------------------------------===//
 
 namespace {
+// RISCV64 calling convention specification:
+// https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#procedure-calling-convention
 struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
   using GenericTarget::GenericTarget;
 
   static constexpr int defaultWidth = 64;
+  static constexpr int defaultWidthBytes = defaultWidth / 8;
+  // TODO: Can't query ABI from inside TargetRewrite so assume the more common
+  // `lp64d` for now. Alternatively could check float-support from
+  // target-features, but that could be overridden by manually setting
+  // `-mabi=lp64`.
+  static constexpr bool hasHardFloatABI = true;
 
   CodeGenSpecifics::Marshalling
   complexArgumentType(mlir::Location loc, mlir::Type eleTy) const override {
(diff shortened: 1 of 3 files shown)


prados-oc:
cc @topperc @jeanPerier

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1433,301 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+  void checkAvailableRegisters(mlir::Location loc,
+                               const Marshalling &previousArguments,
+                               int &gprArgs, int &fprArgs) const {
+    for (auto [ty, attr] : previousArguments) {
+      if (gprArgs <= 0 && fprArgs <= 0)
+        break;
+
+      // previous argument was passed by value and thus takes no registers.
+      if (attr.isByVal())
+        continue;
```
jeanPerier: This may be correct, I just want to double check that the stack address is not being passed in register in that case on RISCV64 (the LoongArch64 is considering them for instance).

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1389,10 +1389,18 @@ struct TargetSparcV9 : public GenericTarget<TargetSparcV9> {
...
+// https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-cc.adoc#procedure-calling-convention
 struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
   using GenericTarget::GenericTarget;
 
   static constexpr int defaultWidth = 64;
+  static constexpr int defaultWidthBytes = defaultWidth / 8;
+  // TODO: Can't query ABI from inside TargetRewrite so assume the more common
+  // `lp64d` for now. Alternatively could check float-support from
+  // target-features, but that could be overridden by manually setting
+  // `-mabi=lp64`.
```
jeanPerier: Could the target-abi be added/threaded into fir::CodeGenSpecifics [here](https://github.com/llvm/llvm-project/blob/72fc97805fc3d940ff6f8a16f23787b690556ef3/flang/lib/Optimizer/CodeGen/TargetRewrite.cpp#L119C27-L119C43) to solve this?
Also, while this may not cover the cases where mabi is used, can't you already cover some cases using the target feature argument of  fir::CodeGenSpecifics ctor?

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1433,301 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+
+      // Complex type is made up of 2 floats
+      if (auto compTy = mlir::dyn_cast<mlir::ComplexType>(type)) {
+        flatTypes.push_back(compTy.getElementType());
+        flatTypes.push_back(compTy.getElementType());
+        continue;
+      }
+
+      // Other types are already flat
+      flatTypes.push_back(type);
```
jeanPerier: Don't you need to do something about character and logical here?
They are allowed by checkValidTypeOrCrash but there seems to be not handling here or later in checkAvailableRegisters  and getIntCCArgs.

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1433,301 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+                gprArgs -= 2;
+              else
+                gprArgs--;
+            }
+          })
+          // NOTE: Tuples are only used to marshal result types and so can't
+          // appear in `previousArguments`.
+          .Default([&](mlir::Type ty) {
+            if (fir::conformsWithPassByRef(ty))
+              if (gprArgs)
```
jeanPerier: nit: `if (fir::conformsWithPassByRef(ty) && gprArgs)`

jeanPerier (commented):
Not familiar at all with the RISCV64, but this looks good to me. A few questions in comments.

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Yes you're right. I thought I checked it before

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: > Could the target-abi be added/threaded into fir::CodeGenSpecifics

That would be nice, but I checked and it requires some more plumbing.

> Also, while this may not cover the cases where mabi is used, can't you already cover some cases using the target feature argument of fir::CodeGenSpecifics ctor

Yes I could probably. The logic to obtaining the actual ABI is a more involved for RISCV though see: https://github.com/llvm/llvm-project/blob/4a5461735d71389e82f2bbd21d9991930bd7e823/clang/lib/Driver/ToolChains/Arch/RISCV.cpp#L178
so ideally that would be reused somehow.

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: I do need to extend the error checking for them in checkValidTypeOrCrash to match the scalar error.

~~Other than that it should be fine, since they will just get treated as integers in getIntCCArgs. And checkAvailableRegisters only checks previously marshalled arguments so they are also already integers (or sequence with len=1 for the character case). If needed I can be more explicit about things and flatten them as integers here, but I don't think it makes a difference (assuming I'm not missing anything Fortran specific here).~~

Yes actually you're correct in that they need to be flattened here to integers already, in order for the hardfloat ABI to catch the float + integer case.

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Done

tarunprabhu:
Thanks for working on this, @prados-oc. 

Have you run the Fortran tests in the llvm-test-suite with this?

prados-oc:
Hey @tarunprabhu. Yes, but I think most of the bind(c) tests are in the gfortran testsuite and currently disabled.

prados-oc:
Hey @jeanPerier I changed the error checking and flattening in this last commit. There is no checkValidTypeOrCrash anymore and all the error checking is done in flattenTypeList. This is similar to how other targets handle it (this change is mostly the same to LoongArch64). I think this makes the code more straight forward and also makes sure all the necessary types are handled/flattened.

tarunprabhu:
> Hey @tarunprabhu. Yes, but I think most of the bind(c) tests are in the gfortran testsuite and currently disabled.

Ok. Thanks for checking. We may be able to re-enable some of those, but I don't have time to do that right away. Have you run any other end-to-end tests, with `BIND(C)` with these changes?

prados-oc:
> Have you run any other end-to-end tests, with BIND(C) with these changes

I'm afraid not @tarunprabhu. But if you point me to any testsuites that test for this I'll gladly run them. I can also try to run some of the gfortran ones, but I don't know if they might be disabled for other reasons on top of ABI requirements.

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: For now I've implemented querying the target-features (if available) and otherwise fallback to the triple. This is how clang does it. If at some point in the future there is a way to query the target-abi that should be prioritized over the other two options.

With this change I've also enabled the integer tests for lp64, since I don't hardcode lp64d anymore.

tarunprabhu:
I just did some simple grepping

```
[llvm-test-suite]$ rg "bind[(]c[)]" Fortran/gfortran/regression -l \
| sed 's/Fortran[/]gfortran[/]regression[/]//g' \
| wc -l
268

[llvm-test-suite]$ rg "bind[(]c[)]" Fortran/gfortran/regression -l \
| sed 's/Fortran[/]gfortran[/]regression[/]//g' \
| xargs -I {} grep -o {} Fortran/gfortran/regression/DisabledFiles.cmake \
| wc -l
53
```

At least in the `regression` directory, it looks like there are a decent number of tests that are not disabled. If this PR passes what is enabled, it should be ok.

prados-oc:
The `Fortran/gfortran` subdir is currently disabled for RISCV. I'll try to enable them and then report back once I have the regression results.

tarunprabhu:
> The `Fortran/gfortran` subdir is currently disabled for RISCV. I'll try to enable them and then report back once I have the regression results.

Thanks!

prados-oc:
@tarunprabhu I ran the gfortran/regression tests on real riscv64 hardware (banana-pi) and it passes all bind(c) related tests:

<details>

<summary>compiled with -march=rv64gcv (lp64d)</summary>

```
Failed Tests (10):
  test-suite :: Fortran/gfortran/regression/gfortran-regression-compile-regression__default_format_2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-compile-regression__default_format_denormal_2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-compile-regression__erf_3_F90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-compile-regression__large_real_kind_form_io_2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-compile-regression__pr36006-2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-compile-regression__quad_2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-compile-regression__round_4_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__c_char_tests_5_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__minlocval_3_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__power_7_f90.test

********************
Executable Missing Tests (15):
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__array_constructor_23_f.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__boz_15_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__c_f_pointer_complex_f03.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__findloc_8_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__gamma_4_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__internal_pack_2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__intrinsic_pack_2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__intrinsic_spread_2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__intrinsic_unpack_2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__pr35944-2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__transpose_reshape_r10_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__unf_io_convert_3_f90.test
  test-suite :: Fortran/gfortran/regression/ieee/gfortran-regression-execute-regression__ieee__signaling_1_f90.test
  test-suite :: Fortran/gfortran/regression/ieee/gfortran-regression-execute-regression__ieee__signaling_2_f90.test
  test-suite :: Fortran/gfortran/regression/ieee/gfortran-regression-execute-regression__ieee__signaling_3_f90.test


Testing Time: 998.25s

Total Discovered Tests: 6412
  Passed            : 6387 (99.61%)
  Failed            :   10 (0.16%)
  Executable Missing:   15 (0.23%)
```

</details>


<details>

<summary>compiled with -march=rv64i (lp64)</summary>

```
Failed Tests (11):
  test-suite :: Fortran/gfortran/regression/g77/gfortran-regression-execute-regression__g77__cabs_f.test
  test-suite :: Fortran/gfortran/regression/g77/gfortran-regression-execute-regression__g77__complex_1_f.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-compile-regression__erf_3_F90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-compile-regression__quad_2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-compile-regression__round_4_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__aliasing_complex_pointer_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__c_char_tests_5_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__complex_intrinsic_3_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__complex_intrinsic_5_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__f2c_2_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__integer_exponentiation_2_f90.test


Testing Time: 1196.10s

Total Discovered Tests: 6393
  Passed: 6382 (99.83%)
  Failed:   11 (0.17%)
```

</details>

From the failing tests I did a grep search for bind(c) and there was only `c_char_tests_5.f90` that uses it. However it doesn't use derived-types so this is also an unrelated failure (which can be investigated in a different patch).
 
The missing executables are because I had to disable some tests regarding unsupported real-types (similar to power-pc in override.yaml).

tarunprabhu:
> @tarunprabhu I ran the gfortran/regression tests on real riscv64 hardware (banana-pi) and it passes all bind(c) related tests:

Thanks for running the checks. I am not at all familiar with RISCV64 either. If @jeanPerier is ok with this, I have no objections.

prados-oc:
@topperc Could you take a look or ping someone from the RISCV folks to check this is what the backend expects and that it matches clang's semantics.

prados-oc:
With https://github.com/llvm/llvm-project/pull/201539 merged, I updated this PR to determine the actual ABI from CodegenSpecifics.
The only thing missing now is support for lp64f, which I can add in a subsequent patch.

tarunprabhu:
I am not familiar with RISCV, but this looks reasonable. It would be great if someone more familiar with RISCV could take a look.

prados-oc:
@lenary I know you also work on the psABI, so can you take a look at this PR and make sure this is what the RISCV64 backend expects? If not can you add some other reviewer who can? This PR is currently blocked by someone for RISCV approval.

prados-oc:
@asb Can you take a look or ping someone from RISCV?

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
   }
+
+  bool hasHardFloatABI(mlir::Location loc) const {
+    auto abi = getTargetABI();
+
+    if (abi == "lp64d")
+      return true;
+    if (abi == "lp64")
+      return false;
+    TODO(loc, "RISCV64 BIND(C) support for " + abi);
```
topperc: We should try to use RISC-V spelling in text strings.

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+          if (floatTy.getWidth() <= 64)
+            flatTypes.push_back(floatTy);
+          else
+            TODO(loc, "128 bit float is not supported by RISCV64");
+        })
+        .Case([&](mlir::ComplexType cmplx) {
+          const auto *sem = &floatToSemantics(kindMap, cmplx.getElementType());
+          if (sem == &llvm::APFloat::IEEEsingle() ||
+              sem == &llvm::APFloat::IEEEdouble())
+            std::fill_n(std::back_inserter(flatTypes), 2,
```
topperc: Can we use `flatTypes.append(2, cmplx.getElementType())`?

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+                 "fir.type value arg character components must have length 1");
+        })
+        .Case([&](fir::SequenceType seqTy) {
+          if (!seqTy.hasDynamicExtents()) {
+            const std::uint64_t numOfEle = seqTy.getConstantArraySize();
+            mlir::Type eleTy = seqTy.getEleTy();
+            // Don't check for subtype again if element-type is scalar.
+            if (mlir::isa<mlir::IntegerType, mlir::FloatType, fir::LogicalType>(
+                    eleTy)) {
+              std::fill_n(std::back_inserter(flatTypes), numOfEle, eleTy);
```
topperc: Can we use `flatTypes.append(numOfEle, eleTy)`?

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+            TODO(loc, "unsupported dynamic extent sequence type as a structure "
+                      "component for BIND(C), "
+                      "VALUE derived type argument and type return");
+        })
+        .Case([&](fir::RecordType recTy) {
+          for (auto &component : recTy.getTypeList()) {
+            mlir::Type eleTy = component.second;
+            llvm::SmallVector<mlir::Type> subTypeList =
+                flattenTypeList(loc, eleTy);
+            if (subTypeList.size() != 0)
```
topperc: `subTypeList.empty()`?

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+              llvm::copy(subTypeList, std::back_inserter(flatTypes));
```
topperc: flatTypes.append(subTypeList)`?

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+            // Don't check for subtype again if element-type is scalar.
+            if (mlir::isa<mlir::IntegerType, mlir::FloatType, fir::LogicalType>(
+                    eleTy)) {
+              std::fill_n(std::back_inserter(flatTypes), numOfEle, eleTy);
+            } else {
+              llvm::SmallVector<mlir::Type> subTypeList =
+                  flattenTypeList(loc, eleTy);
+              if (subTypeList.size() != 0)
+                for (std::uint64_t i = 0; i < numOfEle; ++i)
+                  llvm::copy(subTypeList, std::back_inserter(flatTypes));
```
topperc: `flatTypes.append(subTypeList)`

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+            const std::uint64_t numOfEle = seqTy.getConstantArraySize();
+            mlir::Type eleTy = seqTy.getEleTy();
+            // Don't check for subtype again if element-type is scalar.
+            if (mlir::isa<mlir::IntegerType, mlir::FloatType, fir::LogicalType>(
+                    eleTy)) {
+              std::fill_n(std::back_inserter(flatTypes), numOfEle, eleTy);
+            } else {
+              llvm::SmallVector<mlir::Type> subTypeList =
+                  flattenTypeList(loc, eleTy);
+              if (subTypeList.size() != 0)
```
topperc: `!subTypeList.empty()`

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+  passOnTheStack(unsigned short recAlign, mlir::Type ty, bool isResult) const {
+    CodeGenSpecifics::Marshalling marshal;
+    // The stack is always 8 byte aligned
+    unsigned short align = std::max(recAlign, static_cast<unsigned short>(8));
+    marshal.emplace_back(fir::ReferenceType::get(ty),
+                         AT{align, /*byval=*/!isResult, /*sret=*/isResult});
+    return marshal;
+  }
+
+  const llvm::SmallVector<mlir::Type>
```
topperc: The `const` here seems unnecessary. It's not returning a reference.

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+                 int fprArgs, bool isResult,
+                 const Marshalling &previousArguments) const {
+    auto [recordSize, recordAlign] = fir::getTypeSizeAndAlignmentOrCrash(
+        loc, recTy, getDataLayout(), kindMap);
+
+    CodeGenSpecifics::Marshalling marshal;
+    mlir::MLIRContext *context = recTy.getContext();
+
+    // Have to do this first to catch any illegal types in the record.
+    const llvm::SmallVector<mlir::Type> &flattenedTypes =
```
topperc: Why a reference? flattenTypeList returns a value.

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+    if (flattenedTypes.size() == 2 &&
+        floatAndCanPassInRegister(flattenedTypes[0]) &&
+        floatAndCanPassInRegister(flattenedTypes[1])) {
+      // A struct containing two floating-point reals is passed in two
+      // floating-point registers, if neither real is
+      // more than ABI_FLEN bits wide and at least two floating-point argument
+      // registers are available. (The registers need not be an aligned pair.)
+      // Otherwise, it is passed according to the integer calling convention.
+      if (fprArgs > 1) {
+        if (isResult)
```
topperc: Should this `if` have curly braces around its body to match the `else` body. I know that's required for the LLVM coding standards, but I'm not sure about flang/MLIR.

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1425,6 +1428,316 @@ struct TargetRISCV64 : public GenericTarget<TargetRISCV64> {
...
+      // bitfield), in either order, is passed in a floating-point register and
+      // an integer register, provided the floating-point real is no more than
+      // ABI_FLEN bits wide and the integer is no more than XLEN bits wide, and
+      // at least one floating-point argument register and at least one integer
+      // argument register is available. If the struct is passed in this manner,
+      // and the integer is narrower than XLEN bits, the remaining bits are
+      // unspecified. If the struct is not passed in this manner, then it is
+      // passed according to the integer calling convention.
+      if (gprArgs && fprArgs) {
+        if (isResult)
```
topperc: Same curly brace question

flang/lib/Optimizer/CodeGen/Target.cpp
```diff
@@ -1389,10 +1389,13 @@ struct TargetSparcV9 : public GenericTarget<TargetSparcV9> {
 //===----------------------------------------------------------------------===//
 
 namespace {
+// RISCV64 calling convention specification:
```
topperc: This should probably be `RISC-V RV64`

lenary:
Sorry, I'm not familiar with Fortran at all, and don't have the time to get up to speed to review this.

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Changed here and in the other places

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Done

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Done

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Done

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Done

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Dropped const, thanks!

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Done

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Done

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Changed, thanks!

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: flang/MLIR in general follows the LLVM coding standards, if not explicitly mentioned otherwise. Added braces

flang/lib/Optimizer/CodeGen/Target.cpp
prados-oc: Changed

prados-oc:
Thanks for the review @topperc! Do the tests match the IR that the backend expects?

PR #198338, opened 2026-05-18
[HIP][SPIR-V] Enable SPIR-V backend by default

MrSidims:
Use -no-use-spirv-backend to get SPIR-V by llvm-spirv tool.


clang/lib/Driver/Driver.cpp
```diff
@@ -5096,9 +5096,14 @@ Driver::BuildOffloadingActions(Compilation &C, llvm::opt::DerivedArgList &Args,
           OffloadTriple && OffloadTriple->isSPIRV() &&
           (OffloadTriple->getOS() == llvm::Triple::OSType::AMDHSA ||
            OffloadTriple->getOS() == llvm::Triple::OSType::ChipStar);
-      bool UseSPIRVBackend = Args.hasFlag(options::OPT_use_spirv_backend,
-                                          options::OPT_no_use_spirv_backend,
-                                          /*Default=*/false);
+      // Use SPIR-V backend by default for AMD HIP SPIR-V offload target.
+      bool DefaultUseSPIRVBackend =
+          OffloadTriple && OffloadTriple->isSPIRV() &&
+          OffloadTriple->getVendor() == llvm::Triple::AMD;
```
jhuber6: Seems a little backwards here. I'm guessing without this the `UseSPIRVBackend` is load bearing? We should probably fix that instead.

jhuber6 (commented):
Big fan of this overall, maybe some day we can remove the other path entirely

clang/lib/Driver/Driver.cpp
MrSidims: You mean just drop UseSPIRVBackend (along with -use-spirv-backend option)? I'd prefer not to do this at the moment. While ultimately we will drop supporting llvm-spirv as a tool to generate SPIR-V, but not at the moment, as SPIR-V backend has worse support for debug information translation. So for now better to let `-no-use-spirv-backend` be and promote this option in case if preserving debug info through SPIR-V round-trip is crutial for a user.

jhuber6: No, I mean that we previously seemed to be gating non-SPIRV behavior on whether or not SPIR-V backend was enabled, because previously it was false by default.

mgcarrasco: I see your point @jhuber6, yes before it was potentially too lose. Although now based on how the variable is used in the loop, the change looks safe to me.

jhuber6: I'd prefer we guard the usage of this with "Is this SPIR-V" rather than this flag's default.

mgcarrasco: > I'd prefer we guard the usage of **this** with "Is **this** SPIR-V" rather than this flag's default.

What is _this_ in each case?

MrSidims: Reworked this a bit. Default is `false` again with following up overload of the flag.

Both consumers (aka IsAMDGCNSPIRVWithBackend and UseSPIRVBackendForHipDeviceOnlyNoRDC) were already SPIR-V-gated, so I didn't inline the check there too - let me know if you'd prefer that.

clang/lib/Driver/ToolChains/HIPAMD.cpp
jhuber6, line 171: I would expect the default to not be conditional on the triple. It's a little backwards to set the default of a flag based on the environment rather than the environment make a decision based on a flag.

MrSidims: Should be better now

MrSidims:
Friendly ping

mgcarrasco:
> Friendly ping

Thanks @MrSidims. The changes for the switch look good to me.

I think these following changes could also be required for the switch. What do you think?

There will be some limitations still with the SPIR-V BE such as no debug information. Users should receive a warning if they want to get debug information in this new default mode. Also, the HIP docs should be updated to mention this https://rocm.docs.amd.com/projects/llvm-project/en/latest/conceptual/spirv.html. 

When we implemented the -use-spirv-backend option it was for us to start our work on the BE, so it is currently hidden.

mgcarrasco:
Would it be possible to run a PSDB on this PR? I mean to get early feedback on possible breaking changes before it lands upstream. Also, it would be interesting to see how it could impact comgr (if it makes sense).

MrSidims:
> Would it be possible to run a PSDB on this PR? I mean to get early feedback on possible breaking changes before it lands upstream. Also, it would be interesting to see how it could impact comgr (if it makes sense).

I'll try. "On my machine" ... I wanted to say, that all the tests in rocm-examples and kokkos passing compared to llvm-spirv baseline, but apparently inline_asm test stopped working (again). Lemme check this before any merge (approval is still appreciated :) )

MrSidims:
>There will be some limitations still with the SPIR-V BE such as no debug information

Yes, this will be reflected in the release notes.

mgcarrasco:
Thanks again @MrSidims. Before approving, it would be great to run a PSDB to make sure there are no obvious breaking changes in the stack.

MrSidims:
> Thanks again @MrSidims. Before approving, it would be great to run a PSDB to make sure there are no obvious breaking changes in the stack.

Assuming the SPIR-V backend is enabled as an LLVM target - CI is clean.

jhuber6 (approved):

AlexVlx (approved):
LGTM, thanks!

PR #198340, opened 2026-05-18
[CIR] Guard union ABI alignment when getLargestMember is empty

adams381:
Padding-only unions (an empty union lowered as a single `!u8i`
padding member) leave `getLargestMember()` null when CIRGen walks
record layout through MLIR's DataLayout API.  `RecordType::getABIAlignment`
then passed that null `Type` into `getTypeABIAlignment` and crashed.
This showed up compiling libc++ types such as
`std::__variant_detail::__union` nested under `common_iterator`.

Return ABI alignment `1` when there is no largest member, matching a
byte-padded empty union.  This parallels how empty unions are already
handled for size (`getTypeSizeInBits` uses zero size in that situation).

Regression coverage adds a nested-union global in `empty-union.cpp`.

### DIFF
--- a/clang/test/CIR/CodeGen/empty-union.cpp
+++ b/clang/test/CIR/CodeGen/empty-union.cpp
@@ -2,7 +2,9 @@
 // RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-llvm %s -o - | FileCheck %s --check-prefix=LLVM
 // RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -emit-llvm %s -o - | FileCheck %s --check-prefix=OGCG
 
-// Empty union (should be padded to size 1)
+// Padding-only union: CIR keeps only a tail pad member, so
+// getLargestMember() is null and getABIAlignment must not call
+// getTypeABIAlignment on null.
 union Empty {};
 // CIR: !rec_Empty = !cir.record<union "Empty" padded {!u8i}>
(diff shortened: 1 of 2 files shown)


clang/test/CIR/CodeGen/empty-union.cpp
```diff
@@ -5,14 +5,33 @@
...
 union alignas(16) EmptyAligned {};
 // CIR: !rec_EmptyAligned = !cir.record<union "EmptyAligned" padded {!cir.array<!u8i x 16>}>
-// LLVM: %union.EmptyAligned = type { [16 x i8] }
-// OGCG: %union.EmptyAligned = type { [16 x i8] }
+// LLVM-DAG: %union.EmptyAligned = type { [16 x i8] }
+// OGCG-DAG: %union.EmptyAligned = type { [16 x i8] }
+
+union OuterWithEmpty {
+  Empty e;
+  int x;
```
erichkeane, line 23: Why doesn't this have a 'largest' member?  I would expect the `x` to be largest, right?   So why does this reproduce the problem you're trying to fix?

adams381: OuterWithEmpty is not the repro case — for that union, x is the largest member. The crash is on padding-only unions (Empty {}, EmptyAligned {}) where getLargestMember() is null after tail-pad stripping. OuterWithEmpty only forces a layout path that queries !rec_Empty alignment when Empty is nested in a struct that also has a real member.

erichkeane (approved):

PR #198345, opened 2026-05-18
[analyzer] Fix misleading 'initialized here' note for uninitialized d…

gamesh411:
…eclarations

When a variable is declared without an initializer, the BugReporterVisitor would emit 'initialized here' as a note, which is confusing because the variable was never initialized.
Change the note to 'declared without an initial value' for declarations that have no initializer. Global-storage variables are also taken into consideration.
Removed the SI.Value.isUndef() case, as it is unreachable in
practice because core.uninitialized.Assign (a core checker, always enabled) reports
the assignment before this note can surface.


clang/lib/StaticAnalyzer/Core/BugReporterVisitors.cpp
```diff
@@ -1281,20 +1281,20 @@ static void showBRDiagnostics(llvm::raw_svector_ostream &OS, StoreInfo SI) {
...
-        if (VD->getInit()) {
-          OS << (HasPrefix ? "initialized" : "Initializing")
-             << " to a garbage value";
-        } else {
-          OS << (HasPrefix ? "declared" : "Declaring")
-             << " without an initial value";
-        }
+    const auto *VR = dyn_cast<VarRegion>(SI.Dest);
+    if (VR) {
+      const auto *VD = VR->getDecl();
```
NagyDonat: I like that  you use a simpler approach for calculating `VD` -- but did you verify (e.g. by following the origins of these values) that this always returns the same declaration as the old logic?

Also, please spell out the type name now that it doesn't appear in the `cast<>`.

NagyDonat (changes requested):
Despite the shortness of the change, this commit was surprisingly difficult to understand (at least for me 😅). Eventually I managed to illustrate the output of the code on Venn diagrams:

before PR

after PR

_On the "after this PR" diagram there are two areas with "mixed" output: those produce "initialized here" for globals and "declared without an initial value" for non-globals._

This shows that the observed changes where now "declared without an initial value" appears instead of the old "initialized here" must be situations where `isa<VarRegion>(SI.Dest)` is true, but `HasInit` and `SI.Value.isUndef()` are both false.

In fact, I strongly suspect that `SI.Value.isUndef()` is always false and the old code couldn't emit anything but the `initialized here` message. (This is also supported by the evidence that the string `initialized to a garbage value` never appears in the tests.)

Please investigate whether `SI.Value.isUndef()` can be true in this situation:
- if it can be true, then add a testcase that tests it
- otherwise either remove that branch (along with the "garbage value" message) or test the undefinedness of the value in way that works.

Also, try to simplify the logic to avoid duplicating the `initialized here` printout (e.g. by placing it at the end of the function and returning early if you print something else).

clang/lib/StaticAnalyzer/Core/BugReporterVisitors.cpp
gamesh411: Storesite can be set here:
https://github.com/llvm/llvm-project/blob/f9585141ff3bc78ae7830c9f39ce562d660c5065/clang/lib/StaticAnalyzer/Core/BugReporterVisitors.cpp#L1490-L1492

and on the fallback path here:
https://github.com/llvm/llvm-project/blob/f9585141ff3bc78ae7830c9f39ce562d660c5065/clang/lib/StaticAnalyzer/Core/BugReporterVisitors.cpp#L1513-L1523

I have checked: `isInitializationOfVar()`: https://github.com/llvm/llvm-project/blob/f9585141ff3bc78ae7830c9f39ce562d660c5065/clang/lib/StaticAnalyzer/Core/BugReporterVisitors.cpp#L1190-L1191 explicitly checks `DS->getSingleDecl() != VR->getDecl()` and returns false if they differ. The StoreSite is only set via this path when `R` is a VarRegion.
On the fallback path where `StoreSite = Succ`, R could be a VarRegion whose decl differs from `DS->getSingleDecl()`. So here, `VR->getDecl()` is actually *more* correct than `DS->getSingleDecl()` because the `DeclStmt` might be for a different variable. Additionally, the old code's getSingleDecl() would assert/crash on multi-decl statements.

Type is spelt out now.

gamesh411:
> _On the "after this PR" diagram, there are two areas with "mixed" output: those produce "initialised here" for globals and "declared without an initial value" for non-globals._

This is intentional because I wanted to take into account that global variables are zero-initialised even in cases where automatic-lifetime variables would not be.

>_Please investigate whether SI.Value.isUndef() can be true in this situation:_

It indeed can be, but the effects of this are hard to see, because the `core.uninitialized.Assign` cuts short exploration paths where these notes would surface. I have added a test case in which this checker is disabled and only `core.uninitialized.Branch` is active. In that case, the "garbage value" note appears.

NagyDonat:
> > _On the "after this PR" diagram, there are two areas with "mixed" output: those produce "initialised here" for globals and "declared without an initial value" for non-globals._
> 
> This is intentional because I wanted to take into account that global variables are zero-initialised even in cases where automatic-lifetime variables would not be.

Yes, I've seen that this is intentional, I just mentioned it to explain why I wrote two outputs into a single area on the diagram.

> > _Please investigate whether SI.Value.isUndef() can be true in this situation:_
> 
> It indeed can be, but the effects of this are hard to see, because the `core.uninitialized.Assign` cuts short exploration paths where these notes would surface. I have added a test case in which this checker is disabled and only `core.uninitialized.Branch` is active. In that case, the "garbage value" note appears.

Thanks for investigating the cause of this situation!

As `core.uninitialized.Branch` is a `core` checker, I don't think that there is much point in maintaining a feature that is invisible when `core.uninitialized.Branch` is enabled. I would suggest removing the `SI.Value.isUndef()` branch to simplify this logic.

clang/lib/StaticAnalyzer/Core/BugReporterVisitors.cpp
NagyDonat: > Additionally, the old code's getSingleDecl() would assert/crash on multi-decl statements.

:astonished: 

Thanks for checking and fixing this! :smile:

steakhal (commented):
I'm under load at the moment, but once it's approved, I'll have a look.

steakhal (approved):
Looking at the tests, it's great!
Thank you!

steakhal:
Wait for @NagyDonat's approval please. He had past concerns.

NagyDonat (approved):
LGTM, thanks for the updates!

PR #198346, opened 2026-05-18
[analyzer] Fix false positive in CStringChecker for offset buffer arg…

gamesh411:
…uments

CStringChecker::checkInit() was checking the wrong array elements when the buffer argument pointed into the middle of an array (e.g., memcpy(dst, &arr[i], size)). It was called with BufEnd instead of BufStart, making the ElementRegion index off by (size-1), and the element lookups were relative to array index 0 instead of the actual buffer start offset.

### DIFF
--- a/clang/lib/StaticAnalyzer/Checkers/CStringChecker.cpp
+++ b/clang/lib/StaticAnalyzer/Checkers/CStringChecker.cpp
@@ -437,8 +437,8 @@ ProgramStateRef CStringChecker::checkInit(CheckerContext &C,
   if (!State)
     return nullptr;
 
-  const MemRegion *R = Element.getAsRegion();
-  const auto *ER = dyn_cast_or_null<ElementRegion>(R);
+  SVal BufVal = C.getSVal(Buffer.Expression);
+  const auto *ER = dyn_cast_or_null<ElementRegion>(BufVal.getAsRegion());
   if (!ER)
     return State;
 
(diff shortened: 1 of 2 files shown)


NagyDonat (commented):
The change seems to be correct, but I feel that a different approach might be better.

Instead of "going back" to the `SuperR` and then calculating indices/offsets from that point, you could switch to calculating the indices from the second argument pointer itself (in particular, just dereference it to see the starting element).

This more straigthforward approach could avoid some fragile symbolic calculations.

gamesh411:
> The change seems to be correct, but I feel that a different approach might be better.
> 
> Instead of "going back" to the `SuperR` and then calculating indices/offsets from that point, you could switch to calculating the indices from the second argument pointer itself (in particular, just dereference it to see the starting element).
> 
> This more straigthforward approach could avoid some fragile symbolic calculations.

@NagyDonat, Your intuition seems to be correct. This is less symbolic-value-juggling, and as a bonus, one extra false positive is also eliminated. Maybe because the old getLValueElement would flatten the nested ElementRegion structure, not sure, but it was a nice surprise.

NagyDonat (approved):
This change LGTM, thanks for the update!

As I read the surrounding code, I had some vague ideas for follow-up commits:
1. the "2nd argument" printouts aren't too elegant, perhaps something like "the source buffer of memcpy" etc. would be a better phrasing.
2. The checker does fancy symbolic calculations for finding the index of the last element, then it doesn't report anything if the index of the last element isn't a concrete integer. Perhaps we should switch to concrete numbers earlier to simplify the code.

Issue #193949, opened 2026-04-24
llvm-tblgen memory consumption with libc++ vs libstdc++


rnagy:
Between llvm19 and llvm22 AMDGPUGenGlobalISel.inc almost tripled in size itself but the actual issue is that I've been running into
memory issues because llvm-tblgen seems to consume ~40% more memory when using libc++ instead of libstdc++.

Generating AMDGPUGenGlobalISel.inc with a system using libc++:

> 3794 root      20   0 1163.5m   1.1g   9.6m R  95.4  13.7   0:28.45 llvm-tblgen
> 3794 root      20   0 1291.5m   1.1g   9.6m R 100.0  14.4   0:29.17 llvm-tblgen
> 3794 root      20   0 1291.5m   1.1g   9.6m R 100.0  14.6   0:29.88 llvm-tblgen
> 3794 root      20   0 1419.5m   1.3g   9.8m R  95.3  16.9   0:30.59 llvm-tblgen
> 3794 root      20   0 1803.5m   1.6g   9.8m R 100.0  20.3   0:31.31 llvm-tblgen
> 3794 root      20   0 2059.5m   1.8g   9.8m R  95.1  23.1   0:32.02 llvm-tblgen
> 3794 root      20   0 2315.5m   2.0g   9.8m R 100.0  25.7   0:32.74 llvm-tblgen
> 3794 root      20   0 2315.5m   2.2g   9.8m R 100.0  27.9   0:33.45 llvm-tblgen
> 3794 root      20   0 3759.5m   2.4g   9.8m R  95.2  31.3   0:34.16 llvm-tblgen
> 3794 root      20   0 4107.6m   2.6g   9.8m R  95.5  33.6   0:34.87 llvm-tblgen
> 3794 root      20   0 4107.6m 570.4m   9.8m R 100.0   7.2   0:35.59 llvm-tblgen

Generating AMDGPUGenGlobalISel.inc with a system using libstdc++:
> 825095 root    20   0 1123.1m   1.1g  10.0m R 100.0   1.7   0:24.87 llvm-tblgen
> 825095 root    20   0 1162.8m   1.1g  10.0m R  94.1   1.8   0:25.54 llvm-tblgen
> 825095 root    20   0 1346.7m   1.3g  10.3m R 100.0   2.1   0:26.22 llvm-tblgen
> 825095 root    20   0 1465.2m   1.4g  10.3m R 100.0   2.3   0:26.90 llvm-tblgen
[...]

Issue #53267, opened 2022-01-18
std::deque/std::stack construction is more expensive than GCC


firewave:
I came across this while working on https://github.com/danmar/cppcheck/pull/3716.

Using `std::stack` in Clang with the default `std::deque` backend is more expensive than doing so in GCC. It's about 10% slower. In the case I came across this I was able to speed this up using `std::vector` instead. That actually helped both compilers but Clang profited more from that.

The problem is the `std::_Deque_base<void*, std::allocator<void*> >::_M_initialize_map(unsigned long)` which GCC does not generate in the code - see https://godbolt.org/z/v1d9GsGnW.

Here's some sample code I used for profiling with `-O2` and `callgrind`:

```
#include <stack>

extern void f()
{
        std::stack<void*> stk;
        stk.push(nullptr);
}

int main()
{
        for (int i = 0; i < 100000; ++i)
        {
                f();
        }

        return 0;
}
```

Total Ir:
Clang `43,790,737`
GCC `39,794,842`

firewave:
I am using `libstdc++`.

fhahn:
From a first glance, it seems like the `std::vector` case is optimized away because LLVM is able the inline the full `std::vector` constructor and the `::push` calls. That appears to not be the case for the `std::deque` case

fhahn:
I am confused, how is this different to #53268?

firewave:
> I am confused, how is this different to #53268?

This is about the `std::deque` constructor being more expensive than GCC. The other one is about dead code elimination with `std::deque` which neither compiler does. So it seems like two different things possibly depending on the same underlying behavior.

fhahn:
> This is about the std::deque constructor being more expensive than GCC. The other one is about dead code elimination with std::deque which neither compiler does. So it seems like two different things possibly depending on the same underlying behavior.

Right, but the reason it is more expensive (`_M_initialize_map` call not eliminated) is also the reason the code cannot be eliminated.

firewave:
> Right, but the reason it is more expensive (`_M_initialize_map` call not eliminated) is also the reason the code cannot be eliminated.

If that is the case feel free to close one of them as a duplicate.

frederick-vs-ja:
I'm tentatively closing this as duplicate.

PR #198348, opened 2026-05-18
[libcxx] Configuration option to lower deque block size

Prabhuk:
Introduce new configuration option to choose a smaller block size for deque.
Please note the test files are generated using an LLM.

Background:
Larger block size for deque is unsuitable for several environments with
severe memory pressures. This patch attempts to address the problems related to deque 
discussed in these previous bugs: 

https://github.com/llvm/llvm-project/issues/31782
https://github.com/llvm/llvm-project/issues/193949

This Chromium issue also neatly summarizes the problem:
https://issues.chromium.org/issues/40498153

Our patch is based on the discussion in https://github.com/llvm/llvm-project/issues/31782
and modeled similar to the downstream patch created by @rnagy 
https://github.com/openbsd/src/commit/b2f9966e67c20a9342561535a598345d4d49fa4e
but the implementation here is a bit different.

There are open questions related to other containers (queue, stack) which I haven't had
a chance to dig into yet.

We encountered this higher heap usage issue in our embedded projects.

Fixes: #31782


libcxx/include/__config
```diff
@@ -522,6 +522,15 @@ typedef __char32_t char32_t;
 
 #  define _PSTL_USE_NONTEMPORAL_STORES_IF_ALLOWED
 
+#  ifndef _LIBCPP_USE_LOWER_DEQUE_BLOCK_SIZE
```
philnik777: This is an ABI-sensitive flag, so it should be handled like one. See e.g. `_LIBCPP_ABI_ENABLE_SHARED_PTR_TRIVIAL_ABI`.

libcxx/include/__cxx03/__config
```diff
@@ -1072,6 +1072,15 @@ typedef __char32_t char32_t;
 #    define _LIBCPP_HAS_EXPLICIT_THIS_PARAMETER
 #  endif
 
+#  ifndef _LIBCPP_USE_LOWER_DEQUE_BLOCK_SIZE
```
philnik777: Please don't modify the C++03 headers.

libcxx/include/__config
```diff
@@ -522,6 +522,15 @@ typedef __char32_t char32_t;
...
+#    define _LIBCPP_USE_LOWER_DEQUE_BLOCK_SIZE 0
```
philnik777: This seems unused?

libcxx/test/libcxx/containers/sequences/deque/lower_deque_block_size.lower.compile.pass.cpp
```diff
@@ -0,0 +1,33 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// <deque>
+
+// Test that std::__deque_block_size has lowered sizes when _LIBCPP_USE_LOWER_DEQUE_BLOCK_SIZE is defined to 1.
+
+// ADDITIONAL_COMPILE_FLAGS: -D_LIBCPP_USE_LOWER_DEQUE_BLOCK_SIZE=1
```
philnik777: We don't usually make the ABI configurable by the user, and I don't see a reason to deviate here.

libcxx/include/__config
```diff
@@ -522,6 +522,15 @@ typedef __char32_t char32_t;
...
+#    define _LIBCPP_DEQUE_TARGET_BLOCK_SIZE 4096
+#    define _LIBCPP_DEQUE_MIN_ELEMENTS_PER_BLOCK 16
+#  else
+#    define _LIBCPP_DEQUE_TARGET_BLOCK_SIZE 512
+#    define _LIBCPP_DEQUE_MIN_ELEMENTS_PER_BLOCK 2
```
philnik777: 2 seems a bit low. If you want a linked list you can use that instead.

philnik777 (changes requested):
#53267 is definitely unrelated, since that's not about libc++. Please put the appropriate issues in the description with `Fixes #<number>` to close them on merge.

libcxx/include/__config
Prabhuk: I am porting memory constrained baremetal codebases which use libstdc++ to use libcxx. In this context, rewriting the uses of deque across the given codebase and third party dependencies is a hard problem to overcome. Offering a configuration option that is close to libstdc++'s deque in terms of dynamic memory usage will help us start using libcxx in these projects. If this is not an acceptable justification, please let me know what is the minimum size that you would recommend?

One possible design choice I see is to let libcxx offer multiple configurations in the future with different defaults and one of those choices being size 2.

libcxx/test/libcxx/containers/sequences/deque/lower_deque_block_size.lower.compile.pass.cpp
Prabhuk: Can you help me understand this comment? I followed the pattern from this previous test file: https://github.com/llvm/llvm-project/blob/main/libcxx/test/libcxx/memory/trivial_abi/shared_ptr_arg.pass.cpp#L13

libcxx/include/__config
Prabhuk: Removed this unused define.

libcxx/include/__config
Prabhuk: Renamed the flag.

libcxx/include/__cxx03/__config
Prabhuk: Removed this change as well as the changes in C++03 deque file.

libcxx/include/__config
Prabhuk: Brought back setting the flag to 0 as I am using it in the lower_deque_block_size.default.compile.pass.cpp test to assert _LIBCPP_ABI_USE_LOWER_DEQUE_BLOCK_SIZE is 0 when it is not defined. Please let me know if this is ok or if I should modify the way I am doing the tests.

libcxx/include/__configuration/abi.h
```diff
@@ -76,6 +76,7 @@
 #  define _LIBCPP_ABI_USE_WRAP_ITER_IN_STD_STRING_VIEW
 #  define _LIBCPP_ABI_VARIANT_INDEX_TYPE_OPTIMIZATION
 #  define _LIBCPP_ABI_TRIVIALLY_COPYABLE_BIT_ITERATOR
+#  define _LIBCPP_ABI_USE_LOWER_DEQUE_BLOCK_SIZE
```
petrhosek: I'd suggest `_LIBCPP_ABI_USE_SMALL_DEQUE_BLOCK_SIZE` as the name to make the intention clear.
```suggestion
#  define _LIBCPP_ABI_USE_SMALL_DEQUE_BLOCK_SIZE
```

Prabhuk: Renamed as suggested. By default we do not want this to flag to be set so I had removed it from abi.h.

libcxx/include/__config
philnik777: "this matches libstdc++" is _very_ weak IMO. We want a good reason, and "someone else did it" is never a good reason. I think the absolute minimum I'd accept is 4.

libcxx/include/deque
```diff
@@ -265,7 +265,10 @@ _LIBCPP_BEGIN_NAMESPACE_STD
 
 template <class _ValueType, class _DiffType>
 struct __deque_block_size {
-  static const _DiffType value = sizeof(_ValueType) < 256 ? 4096 / sizeof(_ValueType) : 16;
+  static const _DiffType value =
+      (sizeof(_ValueType) * _LIBCPP_DEQUE_MIN_ELEMENTS_PER_BLOCK < _LIBCPP_DEQUE_TARGET_BLOCK_SIZE)
+          ? (_LIBCPP_DEQUE_TARGET_BLOCK_SIZE / sizeof(_ValueType))
+          : _LIBCPP_DEQUE_MIN_ELEMENTS_PER_BLOCK;
```
philnik777: I think I'd rather just have two definitions of this. All the macros make is quite hard to read.

libcxx/test/libcxx/containers/sequences/deque/lower_deque_block_size.default.compile.pass.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+static_assert(std::__deque_block_size<char, std::ptrdiff_t>::value == 4096, "");
+static_assert(std::__deque_block_size<int, std::ptrdiff_t>::value == 1024, "");
+static_assert(std::__deque_block_size<double, std::ptrdiff_t>::value == 512, "");
+
+static_assert(std::__deque_block_size<TypeOfSize<255>, std::ptrdiff_t>::value == 16, "");
+static_assert(std::__deque_block_size<TypeOfSize<256>, std::ptrdiff_t>::value == 16, "");
+static_assert(std::__deque_block_size<TypeOfSize<512>, std::ptrdiff_t>::value == 16, "");
+
+int main(int, char**) { return 0; }
```
philnik777:
```suggestion
```

libcxx/test/libcxx/containers/sequences/deque/lower_deque_block_size.lower.compile.pass.cpp
```diff
@@ -0,0 +1,33 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// <deque>
+
+// Test that std::__deque_block_size has lowered sizes when _LIBCPP_ABI_USE_SMALL_DEQUE_BLOCK_SIZE is defined to 1.
+
+// ADDITIONAL_COMPILE_FLAGS: -D_LIBCPP_ABI_USE_SMALL_DEQUE_BLOCK_SIZE=1
```
philnik777:
```suggestion
// REQUIRES: libcpp-abi-use-small-deque-block-size
```
See e.g. `libcpp-abi-no-compressed-pair-padding` for how to do this.

libcxx/include/__config
philnik777: Why do you care to assert it there?

libcxx/test/libcxx/containers/sequences/deque/lower_deque_block_size.lower.compile.pass.cpp
philnik777: Yeah, we should really update those tests to not define the macro. We have proper testing set up now. Not a reason to do it here though.

libcxx/include/__configuration/abi.h
philnik777: Why do we not want this by default? Other bug reports seem to suggest that this isn't only an issue on very memory-constrained platforms.

sthen:
btw, I'm unable to build LLVM 22 from the standard distribution on OpenBSD/i386 without this - running tblgen for the AMDGPU target fails because it bumps into MAXDSIZ (3GB).

Prabhuk:
> btw, I'm unable to build LLVM 22 from the standard distribution on OpenBSD/i386 without this - running tblgen for the AMDGPU target fails because it bumps into MAXDSIZ (3GB).

I'll upload follow up changes based on the reviews in a bit. We had a chat about this patch in the monthly libcxx meeting yesterday and the maintainers appear to be supportive of this patch in general.

libcxx/include/__config
Prabhuk: This wasn't necessary as suggested. Removed it. We are not setting anything in __config file anymore as the sizes are hardcoded within deque file as suggested by the other review comment.

libcxx/test/libcxx/containers/sequences/deque/lower_deque_block_size.lower.compile.pass.cpp
Prabhuk: Done. PTAL.

libcxx/include/__config
Prabhuk: That's reasonable. Updated the patch. PTAL.

libcxx/include/__configuration/abi.h
Prabhuk: We are worried about how this may impact performance and if this could break in other ways if I had changed the default behavior.

libcxx/test/libcxx/containers/sequences/deque/lower_deque_block_size.default.compile.pass.cpp
Prabhuk: Done. PTAL.

libcxx/test/libcxx/containers/sequences/deque/lower_deque_block_size.lower.compile.pass.cpp
Prabhuk: Done. PTAL.

libcxx/include/deque
Prabhuk: Done. PTAL

philnik777 (commented):
I think I'd like to have this in the default ABIv2 configuration. Multiple independent groups have asked for this, so it seems to be a good idea generally and this gives us at least some coverage in the CI.

philnik777 (approved):
LGTM assuming the CI is happy.

Prabhuk:
> I think I'd like to have this in the default ABIv2 configuration. Multiple independent groups have asked for this, so it seems to be a good idea generally and this gives us at least some coverage in the CI.

I updated libcxx/include/__configuration/abi.h to address this. PTAL.

Prabhuk:
> LGTM assuming the CI is happy.

CI was not happy. Updated the failing test and CI is in progress. PTAL if it passes. Thanks.

Prabhuk:
> > LGTM assuming the CI is happy.
> 
> CI was not happy. Updated the failing test and CI is in progress. PTAL if it passes. Thanks.

@philnik777 The builders are happy. Please let me know if the changes to the test looks good. I'll wait for your acknowledgement before landing the change.

philnik777 (approved):

ldionne (commented):
@philnik777 Is there a reason you preferred to have this be a on/off setting instead of letting users configure the block size via the macro, and then provide a default value that is sensible? When we talked about this patch at the libc++ monthly we discussed that and that was my preference. Curious to know your thoughts.

PR #198351, opened 2026-05-18
[PDB][llvm-pdbutil] Add DXContainer support for `pdb2yaml` and `yaml2pdb`

kuilpd:
This patch enables the following:
1. Attempting to parse a `DXContainer` from stream 5 (generated by DirectX tools) of a PDB file, to be used later in `llvm-pdbutil`.
2. Outputting a PDB file with a built in DXContainer as YAML. Existing DirectX tools form a PDB container with empty DBI, TPI and IPI streams, so this patch also allows them to be empty when dumping a PDB file as YAML.
3. Creating a PDB file from YAML with an built in DXContainer. When creating a PDB with a DXContainer, streams DBI, TPI and IPI can be completely empty, so this patch also includes adjustments to allow forming such a PDB file. This is done to maintain compatibility with other DirectX tools.

### DIFF
--- a/llvm/tools/llvm-pdbutil/llvm-pdbutil.cpp
+++ b/llvm/tools/llvm-pdbutil/llvm-pdbutil.cpp
@@ -700,6 +700,10 @@ cl::opt<bool> IpiStream("ipi-stream",
                         cl::desc("Dump the IPI Stream (Stream 5)"),
                         cl::sub(PdbToYamlSubcommand), cl::init(false));
 
+cl::opt<bool> DXContainerStream("dxcontainer",
+                                cl::desc("Dump the DXContainer Stream"),
+                                cl::sub(PdbToYamlSubcommand), cl::init(false));
+
 cl::opt<bool> PublicsStream("publics-stream",
                             cl::desc("Dump the Publics Stream"),
                             cl::sub(PdbToYamlSubcommand), cl::init(false));
@@ -1560,6 +1564,7 @@ int main(int Argc, const char **Argv) {
       opts::pdb2yaml::DbiStream = true;
       opts::pdb2yaml::TpiStream = true;
       opts::pdb2yaml::IpiStream = true;
+      opts::pdb2yaml::DXContainerStream = true;
       opts::pdb2yaml::PublicsStream = true;
       opts::pdb2yaml::DumpModules = true;
       opts::pdb2yaml::DumpModuleFiles = true;
--- a/llvm/tools/llvm-pdbutil/llvm-pdbutil.cpp
+++ b/llvm/tools/llvm-pdbutil/llvm-pdbutil.cpp
@@ -76,6 +76,7 @@
 #include "llvm/DebugInfo/PDB/PDBSymbolTypeFunctionSig.h"
 #include "llvm/DebugInfo/PDB/PDBSymbolTypeTypedef.h"
 #include "llvm/DebugInfo/PDB/PDBSymbolTypeUDT.h"
+#include "llvm/ObjectYAML/yaml2obj.h"
 #include "llvm/Support/BinaryByteStream.h"
 #include "llvm/Support/COM.h"
 #include "llvm/Support/CommandLine.h"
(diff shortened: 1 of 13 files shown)


kuilpd:
`llvm-pdbutil pdb2yaml`: #198357 
`llvm-pdbutil yaml2pdb`: #198365

bogner:
I appreciate that you're trying to break this up into smaller pieces, but it's a little bit awkward to commit a disabled test, and I'm finding that I need to jump between the three PRs to get a good understanding of what's actually going on here. Would it be better to simply do all of these together as one PR?

kuilpd:
> I appreciate that you're trying to break this up into smaller pieces, but it's a little bit awkward to commit a disabled test, and I'm finding that I need to jump between the three PRs to get a good understanding of what's actually going on here. Would it be better to simply do all of these together as one PR?

No problem, I merged the commits from other PRs to this one and I'll update the description as well.

llvm/tools/llvm-pdbutil/llvm-pdbutil.cpp
```diff
@@ -1560,6 +1589,7 @@ int main(int Argc, const char **Argv) {
       opts::pdb2yaml::DbiStream = true;
       opts::pdb2yaml::TpiStream = true;
       opts::pdb2yaml::IpiStream = true;
+      opts::pdb2yaml::DXContainerStream = true;
```
bogner, line 1592: Is this really safe to turn on with all? Couldn't stream 5 be something else if this is some generic PDB that isn't DX?

kuilpd: > Is this really safe to turn on with all? Couldn't stream 5 be something else if this is some generic PDB that isn't DX?

Seems like the only way to detect if it's a DXContainer is to just try parsing it. If it wasn't parsed as a correct DXContainer, then it proceeds to be parsed as a regular PDB stream. While developing this, I did have a case where a regular PDB stream 5 just happened to get successfully parsed as a DXContainer, but it ended up being broken, so I made the parsing more strict in #198586, .

The only way a PDB file indirectly indicates it contains a DXContainer is that DBI, TPI and IPI streams are empty, as generated by DXC tools. But I don't think we should rely on that, in case this isn't always true or something else also makes those streams empty.

I can remove that option from `--all`, maybe it would make it safer for users of regular PDB files. But if someone tries to do `pdb2yaml --all` on a DXContainer PDB file, they would be quite confused to see that it's pretty much empty.

bogner: Ok, so no matter what we do there's potential for `--all` being confusing. I'm a bit worried about the downside, in that if a PDB has a non-dxc stream 5 this will error now, and I don't think there's a way to say "all except this".

I guess we have a couple of ways forward here:
1. Include DXContainer in `--all`. Non-dxc PDB that has stream 5 doesn't work well, non-dxc PDB with no stream 5 is unaffected.
2. Omit DXContainer from `--all`. Potentially confusing and you need `--all --dxcontainer` to really do "all" for dx PDBs
3. Add some heuristic or some kind of "flavor" flag that decides if we're a dx container or not and curate "all" appropriately"

I guess option (3) needs some thought so it isn't actionable enough to be worth blocking on. Option (2) gives us the most flexibility and avoids breaking any existing workflows, so I think it's my preference for the short term. What do you think?

bogner (approved):
Code changes look good. We just need to work out what we want to do with `--all`

llvm/tools/llvm-pdbutil/llvm-pdbutil.cpp
kuilpd: > 1. Include DXContainer in --all. Non-dxc PDB that has stream 5 doesn't work well, non-dxc PDB with no stream 5 is unaffected.

I wouldn't say it doesn't work well, it would have to be a wild coincidence for a regular PDB stream to be correctly parsed as a DXContainer, but if you want to absolutely eliminate any possibility, then we can remove it from `--all`. I'm guessing it's primarily going to be used by us for testing anyway, so I don't see much harm in that.

I have another PR planned that adds DXContainer to `pdbutil dump` and operates the same way. It includes `--dxcontainer` in `--all` as well, should we handle it the same way then?

bogner: Maybe I don't understand the failure mode. If you pass `--all` or `--dxcontainer` and a PDB has something that is not a DXContainer as stream 5, what do we do?
- Error out and quit printing
- Warn, print nothing for stream 5, and continue
- Silently skip the section

I was assuming error, but I suppose warn might also be the case. Either way, I would assume that the tool exits non-zero. I guess that `--all` isn't usually used except on the command line to inspect things, so maybe it's not a big deal if it exits non-zero, but I could certainly see that being confusing and/or breaking scripts.

kuilpd: All the options for `pdb2yaml` just attempt to output the selected section, and if there isn't one, they print nothing, without an error. That's why when `--all` enables all options, it doesn't crash when a certain option doesn't have anything to print. This is a pre-existing behavior even before my patch, not all PDB streams are present in every case. The only thing that is always printed is the MSF header. Trying `--dxcontainer` on a non-DXContainer PDB will successfully output nothing after the header, same as trying `--dbi-stream` on a DXContainer PDB.

So, the only erroneous output we could get is when stream 5 in a regular PDB file (which is usually a custom named stream) somehow has data that can be succesfully parsed as a DXContainer header, which is how `--dxcontainer` determines if there is one in the PDB file. After merging the patch that checks for DXBC header magic during parsing, this should be quite unlikely.

bogner: Alright fair enough. The current behaviour seems sensible then. This LGTM.

slydiman:
The buildbot [lldb-remote-linux-ubuntu](https://lab.llvm.org/buildbot/#/builders/195/builds/25499) is broken too.
It seems we must revert this patch for now.

PR #198354, opened 2026-05-18
[mlir][AMDGPU] Extend amdgpu.transpose_load for gfx1250

krzysz00:
This commit adds support for gfx1250's ds_load_tr* instructions to `amdgpu.transpose_load` since they're pretty close to the gfx950 ones.

### DIFF
--- a/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPUOps.td
+++ b/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPUOps.td
@@ -1110,7 +1110,7 @@ def AMDGPU_DotOp :
   let summary = "MLIR wrapper for AMDGPU v_dot* intrinsics";
   let description = [{
     The `amdgpu.dot` op is an MLIR wrapper over the `v_dot*` family of intrinsics,
-    which compute `D = sum_i A[i] * B[i] + C`. 
+    which compute `D = sum_i A[i] * B[i] + C`.
 
     Variants (source, dest, signedness, chipset -> intrinsic).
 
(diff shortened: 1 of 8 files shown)


mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPUOps.td
```diff
@@ -1427,44 +1416,44 @@ def AMDGPU_GlobalLoadAsyncToLDSOp :
...
     where the subgroup of threads collectively reads a matrix from the source
     memref, with each thread reading a vector of the matrix, and gets a transposed matrix
     in as the result. That is, each thread reads a vector of the col-major matrix at different
     indices, and the thread's read result is a vector of the corresponding row of the transposed
     matrix.
 
-    This op is a direct wrapper around the ROCDL `ds_read_tr` family intrinsics. Please refer
-    to the CDNA4 ISA documentation for more details about its exact semantics.
+    This op is a direct wrapper around the ROCDL `ds_read_tr` family intrinsics on
+    CDNA4 and the `ds_load_tr` family of instructions on CDNA5. Please refer
```
efric: Maybe just claim gfx1250 here?

krzysz00: I figure I should follow existing style here

efric: but is this equivalence correct and public?

krzysz00: The intrinsic names are known, the supported architecture is known, and the fact that the two instructions don't have the exact same behavior but are close enough to be wrappable under the same primitive shouldn't be surprising.

efric: The use of `CDNA4` in the original description of this op is the odd one out here; I think it makes more sense to adjust that to gfx950 rather than rather than state `CDNA5`.

krzysz00: Yeah, fair, I'll fix that

efric (approved):

PR #198357, opened 2026-05-18
[llvm-pdbutil] Add DXContainer support for pdb2yaml

kuilpd:
This enables outputting a PDB file with a built in `DXContainer` as YAML.
Existing DirectX tools form a PDB container with empty DBI, TPI and IPI streams, so this patch also allows them to be empty when dumping a PDB file as YAML.


jh7370:
Is this PR intended for review currently? You've put it up in Draft mode, which implies it's still a work in progress.

PR #198358, opened 2026-05-18
[CIR] Fix get_method callee type for member pointer calls

adams381:
Member-pointer calls through `cir.get_method` were lowering to an indirect
callee type that still listed the member function's implicit `this`
parameter after `createGetMethod` had already prepended the adjusted
`void*` receiver.  A call like `(obj->*pmf)(arg)` therefore carried a
three-parameter `var_callee_type` but only two argument operands, and
`-fclangir -emit-llvm` failed LLVM's variadic-call verifier with
`expected var_callee_type to have at most N parameters`.  Classic codegen
emits `(ptr, …)` for the same pattern.

The libc++ sweep had one remaining `frontend-crash-other` bucket hit on
`F_nullptr.pass.cpp`, which boils down to `__builtin_invoke` on a
varargs member function pointer — the same callee/operand mismatch in a
minimal repro.

The fix skips the implicit-`this` slot when cloning the member signature
into the callee function type in `createGetMethod`, and tightens
`GetMethodOp::verify` so callee parameters after the `void*` receiver
match the method parameters after implicit `this`.  Tests update
`pointer-to-member-func.cpp` expectations and add
`builtin-invoke-varargs-member.cpp` with CIR, LLVM, and OGCG checks.

### DIFF
--- a/clang/test/CIR/CodeGen/builtin-invoke-varargs-member.cpp
+++ b/clang/test/CIR/CodeGen/builtin-invoke-varargs-member.cpp
@@ -9,12 +9,16 @@ struct S {
   int foo(int, ...) { return 42; }
 };
 
-void test(S &s) {
-  int (S::*p)(int, ...) = nullptr;
+void invoke(S &s, int (S::*p)(int, ...)) {
   (void)__builtin_invoke(p, s, 1);
 }
 
-// CIR: cir.func{{.*}}@_Z4testR1S
+void test() {
+  S s;
+  invoke(s, nullptr);
+}
+
+// CIR: cir.func{{.*}}@_Z6invokeR1SMS_FiizE
 // CIR-DAG: !cir.func<(!cir.ptr<!void>, !s32i, ...) -> !s32i>
 // CIR: cir.call %{{.*}}(%{{.*}}, %{{.*}}) : (!cir.ptr<!cir.func<(!cir.ptr<!void>, !s32i, ...) -> !s32i>>, !cir.ptr<!void> {{.*}}, !s32i {{.*}}) -> (!s32i
 
--- a/clang/test/CIR/CodeGen/builtin-invoke-varargs-member.cpp
+++ b/clang/test/CIR/CodeGen/builtin-invoke-varargs-member.cpp
@@ -1,0 +1,23 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck %s --check-prefix=CIR --input-file=%t.cir
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -fclangir -emit-llvm %s -o %t-cir.ll
+// RUN: FileCheck %s --check-prefix=LLVM --input-file=%t-cir.ll
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -emit-llvm %s -o %t.ll
+// RUN: FileCheck %s --check-prefix=OGCG --input-file=%t.ll
+
+struct S {
+  int foo(int, ...) { return 42; }
+};
+
+void test(S &s) {
+  int (S::*p)(int, ...) = nullptr;
+  (void)__builtin_invoke(p, s, 1);
+}
+
+// CIR: cir.func{{.*}}@_Z4testR1S
+// CIR-DAG: !cir.func<(!cir.ptr<!void>, !s32i, ...) -> !s32i>
+// CIR: cir.call %{{.*}}(%{{.*}}, %{{.*}}) : (!cir.ptr<!cir.func<(!cir.ptr<!void>, !s32i, ...) -> !s32i>>, !cir.ptr<!void> {{.*}}, !s32i {{.*}}) -> (!s32i
+
+// LLVM: call noundef i32 (ptr, i32, ...) %{{.*}}(ptr noundef %{{.*}}, i32 noundef 1)
+
+// OGCG: call noundef i32 (ptr, i32, ...) %{{.*}}(ptr noundef{{.*}} %{{.*}}, i32 noundef 1)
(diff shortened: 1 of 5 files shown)


clang/test/CIR/CodeGen/pointer-to-member-func.cpp
```diff
@@ -159,9 +159,9 @@ void call(Foo *obj, void (Foo::*func)(int), int arg) {
 // CIR-BEFORE: cir.func {{.*}} @_Z4callP3FooMS_FviEi
 // CIR-BEFORE:   %[[OBJ:.*]] = cir.load{{.*}} %{{.*}} : !cir.ptr<!cir.ptr<!rec_Foo>>, !cir.ptr<!rec_Foo>
 // CIR-BEFORE:   %[[FUNC:.*]] = cir.load{{.*}} : !cir.ptr<!cir.method<!cir.func<(!cir.ptr<!rec_Foo>, !s32i)> in !rec_Foo>>, !cir.method<!cir.func<(!cir.ptr<!rec_Foo>, !s32i)> in !rec_Foo>
-// CIR-BEFORE:   %[[CALLEE:.*]], %[[THIS:.*]] = cir.get_method %[[FUNC]], %[[OBJ]] : (!cir.method<!cir.func<(!cir.ptr<!rec_Foo>, !s32i)> in !rec_Foo>, !cir.ptr<!rec_Foo>) -> (!cir.ptr<!cir.func<(!cir.ptr<!void>, !cir.ptr<!rec_Foo>, !s32i)>>, !cir.ptr<!void>)
+// CIR-BEFORE:   %[[CALLEE:.*]], %[[THIS:.*]] = cir.get_method %[[FUNC]], %[[OBJ]] : (!cir.method<!cir.func<(!cir.ptr<!rec_Foo>, !s32i)> in !rec_Foo>, !cir.ptr<!rec_Foo>) -> (!cir.ptr<!cir.func<(!cir.ptr<!void>, !s32i)>>, !cir.ptr<!void>)
```
andykaylor, line 162: I'm confused as to what happened here. I see that we previously had two pointers representing a single `this` argument, but why did we discard the typed pointer rather than the void pointer?

clang/test/CIR/CodeGen/builtin-invoke-varargs-member.cpp
```diff
@@ -0,0 +1,23 @@
...
+// RUN: FileCheck %s --check-prefix=LLVM --input-file=%t-cir.ll
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -emit-llvm %s -o %t.ll
+// RUN: FileCheck %s --check-prefix=OGCG --input-file=%t.ll
+
+struct S {
+  int foo(int, ...) { return 42; }
+};
+
+void test(S &s) {
+  int (S::*p)(int, ...) = nullptr;
```
andykaylor: The fact that this is initialized to a nullptr bothers me. Why not have it passed in as a parameter?

clang/test/CIR/CodeGen/builtin-invoke-varargs-member.cpp
```diff
@@ -0,0 +1,23 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck %s --check-prefix=CIR --input-file=%t.cir
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -fclangir -emit-llvm %s -o %t-cir.ll
+// RUN: FileCheck %s --check-prefix=LLVM --input-file=%t-cir.ll
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -emit-llvm %s -o %t.ll
+// RUN: FileCheck %s --check-prefix=OGCG --input-file=%t.ll
+
+struct S {
+  int foo(int, ...) { return 42; }
```
andykaylor, line 9: Is there a reason this is variadic?

clang/test/CIR/CodeGen/builtin-invoke-varargs-member.cpp
```diff
@@ -0,0 +1,23 @@
...
+  (void)__builtin_invoke(p, s, 1);
+}
+
+// CIR: cir.func{{.*}}@_Z4testR1S
+// CIR-DAG: !cir.func<(!cir.ptr<!void>, !s32i, ...) -> !s32i>
+// CIR: cir.call %{{.*}}(%{{.*}}, %{{.*}}) : (!cir.ptr<!cir.func<(!cir.ptr<!void>, !s32i, ...) -> !s32i>>, !cir.ptr<!void> {{.*}}, !s32i {{.*}}) -> (!s32i
+
+// LLVM: call noundef i32 (ptr, i32, ...) %{{.*}}(ptr noundef %{{.*}}, i32 noundef 1)
+
+// OGCG: call noundef i32 (ptr, i32, ...) %{{.*}}(ptr noundef{{.*}} %{{.*}}, i32 noundef 1)
```
andykaylor, line 27: There's a lot more going on here you're checking. I suppose this is sufficient for what you're trying to fix, but it seems we don't have a general test for invoking a member via a pointer-to-member-func and it would be nice to add that.

clang/test/CIR/CodeGen/pointer-to-member-func.cpp
adams381: The indirect callee type should not include the typed this — Itanium passes one adjusted void* receiver on the call. Duplicating S* in the callee signature is what tripped var_callee_type. pointer-to-member-func.cpp covers the non-variadic PMF call() path.

clang/test/CIR/CodeGen/builtin-invoke-varargs-member.cpp
adams381: Switched to passing the member pointer as a parameter: invoke(s, nullptr) from test(), rather than a local nullptr initializer.

clang/test/CIR/CodeGen/builtin-invoke-varargs-member.cpp
adams381: Variadic to match libcxx F_nullptr.pass.cpp — it needs a variadic PMF and __builtin_invoke(..., 0) with no trailing args.

clang/test/CIR/CodeGen/builtin-invoke-varargs-member.cpp
adams381: Non-variadic PMF invoke is in pointer-to-member-func.cpp; this file pins the libcxx variadic nullptr case that was crashing.

andykaylor (approved):
lgtm

PR #198361, opened 2026-05-18
[CIR] Include union tail pad in getTypeSizeInBits

adams381:
Padded CIR unions (e.g. libstdc++ `std::string` SSO layout) carry a
trailing byte-array member so the record matches the AST layout size.
`RecordType::getTypeSizeInBits` was returning only the largest-aligned
member and ignored that tail, so the CIR view of the union was 8 bytes
smaller than what `LowerToLLVM` emits.  Parent structs then picked up
a spurious trailing pad via `insertPadding`, arrays of those structs
used the wrong stride, and heap allocations could be overrun (Eigen's
`array_of_string` hits this directly).

The fix adds the padding member's size when the union is marked
`padded`, so struct size, GEP strides, and `new T[n]` allocation sizes
match OGCG.  Regression test models the SSO-shaped record and checks
the 96-byte `new` for three elements.

### DIFF
--- a/clang/lib/CIR/Dialect/IR/CIRTypes.cpp
+++ b/clang/lib/CIR/Dialect/IR/CIRTypes.cpp
@@ -345,6 +345,15 @@ Type RecordType::getLargestMember(const ::mlir::DataLayout &dataLayout) const {
   });
 }
 
+mlir::Type RecordType::getPadding() const {
+  if (!getPadded())
+    return {};
+  llvm::ArrayRef<mlir::Type> members = getMembers();
+  if (members.empty())
+    return {};
+  return members.back();
+}
+
 bool RecordType::isLayoutIdentical(const RecordType &other) {
   if (getImpl() == other.getImpl())
     return true;
@@ -396,8 +405,8 @@ RecordType::getTypeSizeInBits(const mlir::DataLayout &dataLayout,
     // by `insertPadding`, making `sizeof(parent)` and array GEPs off by the
     // missing bytes.
     llvm::TypeSize size = dataLayout.getTypeSizeInBits(largest);
-    if (getPadded())
-      size += dataLayout.getTypeSizeInBits(*getMembers().rbegin());
+    if (mlir::Type tailPad = getPadding())
+      size += dataLayout.getTypeSizeInBits(tailPad);
     return size;
   }
 
--- a/clang/lib/CIR/Dialect/IR/CIRTypes.cpp
+++ b/clang/lib/CIR/Dialect/IR/CIRTypes.cpp
@@ -382,7 +382,23 @@ RecordType::getTypeSizeInBits(const mlir::DataLayout &dataLayout,
     mlir::Type largest = getLargestMember(dataLayout);
     if (!largest)
       return llvm::TypeSize::getFixed(0);
-    return dataLayout.getTypeSizeInBits(largest);
+    // `getLargestMember` returns the highest-aligned variant (which dictates
+    // the union's alignment), not necessarily the largest by size.  When the
+    // union is `padded` -- i.e., its highest-aligned variant is strictly
+    // smaller than its layout size, as happens for any union containing both
+    // a small high-alignment scalar and a larger low-alignment array (e.g.,
+    // `union { char[16]; size_t; }`) -- `lowerUnion` appended a trailing
+    // byte-array member to extend the highest-aligned variant up to the
+    // layout size, and `LowerToLLVM` mirrors this by emitting the union as
+    // `{largest, padding}`.  Include that padding here so `getTypeSize`
+    // reports the same size `LowerToLLVM` produces; otherwise a parent
+    // record containing the union gets a spurious tail-padding member added
+    // by `insertPadding`, making `sizeof(parent)` and array GEPs off by the
+    // missing bytes.
+    llvm::TypeSize size = dataLayout.getTypeSizeInBits(largest);
+    if (getPadded())
+      size += dataLayout.getTypeSizeInBits(*getMembers().rbegin());
+    return size;
   }
 
   auto recordSize = static_cast<uint64_t>(computeStructSize(dataLayout));
--- a/clang/test/CIR/CodeGen/record-with-padded-union.cpp
+++ b/clang/test/CIR/CodeGen/record-with-padded-union.cpp
@@ -1,0 +1,39 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck --check-prefix=CIR --input-file=%t.cir %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-llvm %s -o %t-cir.ll
+// RUN: FileCheck --check-prefix=LLVM --input-file=%t-cir.ll %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -emit-llvm %s -o %t.ll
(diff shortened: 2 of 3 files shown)


clang/test/CIR/CodeGen/record-with-padded-union.cpp
```diff
@@ -0,0 +1,39 @@
...
+};
+
+// Inner union's tail padding must not bleed into the outer record.
+// CIR: !rec_anon{{.*}} = !cir.record<union "anon{{.*}}" padded {!cir.array<!s8i x 16>, !u64i, !cir.array<!u8i x 8>}>
+// CIR: !rec_SSO = !cir.record<struct "SSO" {!cir.ptr<!s8i>, !u64i, !rec_anon{{.*}}}>
+
+// LLVM: %struct.SSO = type { ptr, i64, %union.anon{{.*}} }
+// LLVM: %union.anon{{.*}} = type { i64, [8 x i8] }
+
+// OGCG: %struct.SSO = type { ptr, i64, %union.anon }
```
erichkeane: What are the differences between LLVM/OGCG that you have the lines separated?

clang/test/CIR/CodeGen/record-with-padded-union.cpp
```diff
@@ -0,0 +1,39 @@
...
+  char *p;
+  unsigned long len;
+  union {
+    char local[16];
+    unsigned long capacity;
+  };
+};
+
+// Inner union's tail padding must not bleed into the outer record.
+// CIR: !rec_anon{{.*}} = !cir.record<union "anon{{.*}}" padded {!cir.array<!s8i x 16>, !u64i, !cir.array<!u8i x 8>}>
```
andykaylor, line 18: We've got a real problem with our representation here. We've got two members that are co-located members of the union and then a padding member that is appended to whichever element is the largest/highest aligned? The padding really needs to be represented separately. Something like:

`!rec_anon{{.*}} = !cir.record<union "anon{{.*}}" padded {!cir.array<!s8i x 16>, !u64i}, padding = { !cir.array<!u8i x 8>}>`

Or maybe even

`!rec_anon{{.*}} = !cir.record<union "anon{{.*}}" padded {!cir.array<!s8i x 16>, !u64i}, paddingSize = 8>`

clang/lib/CIR/Dialect/IR/CIRTypes.cpp
```diff
@@ -382,7 +382,23 @@ RecordType::getTypeSizeInBits(const mlir::DataLayout &dataLayout,
...
+    // byte-array member to extend the highest-aligned variant up to the
+    // layout size, and `LowerToLLVM` mirrors this by emitting the union as
+    // `{largest, padding}`.  Include that padding here so `getTypeSize`
+    // reports the same size `LowerToLLVM` produces; otherwise a parent
+    // record containing the union gets a spurious tail-padding member added
+    // by `insertPadding`, making `sizeof(parent)` and array GEPs off by the
+    // missing bytes.
+    llvm::TypeSize size = dataLayout.getTypeSizeInBits(largest);
+    if (getPadded())
+      size += dataLayout.getTypeSizeInBits(*getMembers().rbegin());
```
andykaylor: The fact that we're blindly assuming that the padding is the last element of the member list concerns me (see also my comment about the union representation). I'd like to see a `getPadding()` function or something similar.

clang/test/CIR/CodeGen/record-with-padded-union.cpp
adams381: LLVM checks the CIR-lowered anonymous union layout in the `.ll` file; OGCG checks classic codegen's named `%union.anon` types.  The spellings differ so the CHECK groups are split, but both expect `{ i64, [8 x i8] }` storage for the inner union.

clang/test/CIR/CodeGen/record-with-padded-union.cpp
adams381: Agreed on the representation — tail pad should not be a third union member.  Follow-up #198423 moves it to `padding = { … }` on `!cir.record<union …>` so `members` are only the co-located variants.

clang/lib/CIR/Dialect/IR/CIRTypes.cpp
adams381: #198423 adds `RecordType::getPadding()` and threads it through size, `getLargestMember`, and lowering instead of treating the last `members` entry as pad.

clang/test/CIR/CodeGen/record-with-padded-union.cpp
adams381: Merged the LLVM and OGCG checks into one LLVM: block; both %t-cir.ll and %t.ll use --check-prefix=LLVM.

clang/lib/CIR/Dialect/IR/CIRTypes.cpp
adams381: Added RecordType::getPadding() for the transitional encoding (tail pad is still the last members entry here). Separate padding = { } on the record is tracked in #198423.

clang/test/CIR/CodeGen/record-with-padded-union.cpp
adams381: Agreed — #198423 adds an explicit padding field on !cir.record<union …> per your sketch. This PR keeps the size fix and getPadding() on the current members encoding.

clang/test/CIR/CodeGen/record-with-padded-union.cpp
adams381: LLVM and OGCG both use --check-prefix=LLVM on %t-cir.ll and %t.ll; the CHECK lines are shared because both expect { i64, [8 x i8] } for the inner union.

clang/test/CIR/CodeGen/record-with-padded-union.cpp
adams381: Agreed — separate padding = { } is in #198423; this PR keeps the transitional members encoding and the getTypeSizeInBits fix.

clang/lib/CIR/Dialect/IR/CIRTypes.cpp
adams381: RecordType::getPadding() is on the branch tip (135b567); #198423 moves tail pad into the padding slot.

andykaylor (approved):
lgtm

clang/lib/CIR/Dialect/IR/CIRTypes.cpp
```diff
@@ -382,7 +391,23 @@ RecordType::getTypeSizeInBits(const mlir::DataLayout &dataLayout,
...
+    // a small high-alignment scalar and a larger low-alignment array (e.g.,
+    // `union { char[16]; size_t; }`) -- `lowerUnion` appended a trailing
+    // byte-array member to extend the highest-aligned variant up to the
+    // layout size, and `LowerToLLVM` mirrors this by emitting the union as
+    // `{largest, padding}`.  Include that padding here so `getTypeSize`
+    // reports the same size `LowerToLLVM` produces; otherwise a parent
+    // record containing the union gets a spurious tail-padding member added
+    // by `insertPadding`, making `sizeof(parent)` and array GEPs off by the
+    // missing bytes.
+    llvm::TypeSize size = dataLayout.getTypeSizeInBits(largest);
```
erichkeane, line 407: While this fixes the problem, this actually ended up being the same 'problem' that I ran into.

I think this is actually not the solution.  The problem is 'getLongestMember' is wrong, and I think we should be reverting this.

PR #198362, opened 2026-05-18
[CIR] Report NYI for STDC FENV_ACCESS under -fclangir

adams381:
Functions that enable IEEE floating-point intrinsics via `#pragma STDC FENV_ACCESS ON` or `__attribute__((strict_fp))` set `UsesFPIntrin()` / `StrictFPAttr` on the `FunctionDecl`. CIR doesn't support the FP-constrained dialect ops needed to lower such functions correctly, so `generateCode` now reports `errorNYI` and returns early when either flag is set, preventing silent miscompilation.


clang/test/CIR/CodeGen/fp-pragma-fenv-access.c
```diff
@@ -0,0 +1,28 @@
...
+// RUN: FileCheck --check-prefix=LLVM --input-file=%t-cir.ll %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -emit-llvm %s -o %t.ll
+// RUN: FileCheck --check-prefix=OGCG --input-file=%t.ll %s
+
+extern void use(float f);
+float a, b;
+
+void fenv_access(void) {
+  #pragma STDC FENV_ACCESS ON
+  __builtin_set_flt_rounds(0);
```
andykaylor: This should fail with CIR. We don't have support for FENV access, so we should emit an error when attempting to compile any code that enables it.

clang/lib/CIR/CodeGen/CIRGenFunction.cpp
```diff
@@ -499,6 +505,29 @@ void CIRGenFunction::startFunction(GlobalDecl gd, QualType returnType,
...
+  builder.setDefaultConstrainedRounding(rm);
+  builder.setDefaultConstrainedExcept(fpExceptionBehavior);
+  if ((fd && (fd->UsesFPIntrin() || fd->hasAttr<StrictFPAttr>())) ||
+      (!fd && (fpExceptionBehavior != llvm::fp::ebIgnore ||
+               rm != llvm::RoundingMode::NearestTiesToEven))) {
+    builder.setIsFPConstrained(true);
+    // The StrictFP function attribute on `curFn` is tracked by
+    // `fastMathFuncAttributes`; classic CodeGen adds it here, CIR will once
+    // that feature lands.
+    assert(!cir::MissingFeatures::fastMathFuncAttributes());
```
andykaylor: A MissingFeatures assert isn't sufficient here. This needs to be an errorNYI call. Otherwise, we will be generating incorrect IR.

andykaylor (changes requested):

clang/lib/CIR/CodeGen/CIRGenFunction.cpp
```diff
@@ -499,6 +505,29 @@ void CIRGenFunction::startFunction(GlobalDecl gd, QualType returnType,
   const auto *fd = dyn_cast_or_null<FunctionDecl>(d);
   curFuncDecl = (d ? d->getNonClosureContext() : nullptr);
 
+  // Set up the default floating-point environment for this function before
```
andykaylor: This comment is unnecessary.

clang/lib/CIR/CodeGen/CIRGenFunction.cpp
```diff
@@ -499,6 +505,29 @@ void CIRGenFunction::startFunction(GlobalDecl gd, QualType returnType,
...
+  // any FP-related code runs.  When the function uses FP intrinsics
+  // (`#pragma STDC FENV_ACCESS ON`, `__builtin_set_flt_rounds`, etc.) or
+  // carries `__attribute__((strictfp))`, enable FP-constrained mode on the
+  // builder so that CIRGenFPOptionsRAII sees a constrained function when a
+  // statement-scope `#pragma` changes rounding/exception behavior away from
+  // the defaults.  Mirrors classic CodeGen's `StartFunction` initialization
```
andykaylor: We should consider creating a Claude rule telling it not to generate "Mirrors classic CodeGen" comments.

clang/test/CIR/CodeGen/fp-pragma-fenv-access.c
adams381: Replaced with a `-verify` RUN; CIR now expects `ClangIR code gen Not Yet Implemented: STDC FENV_ACCESS`.

clang/lib/CIR/CodeGen/CIRGenFunction.cpp
adams381: Removed the `setIsFPConstrained` setup. FENV/StrictFP functions get `errorNYI` in `startFunction`, `generateCode` returns before emitting the body, and the RAII helper uses the same NYI instead of the assert.

clang/lib/CIR/CodeGen/CIRGenFunction.cpp
adams381: Dropped it.

clang/lib/CIR/CodeGen/CIRGenFunction.cpp
adams381: Removed that wording.

clang/test/CIR/CodeGen/fp-pragma-fenv-access.c
```diff
@@ -0,0 +1,11 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir %s -verify -emit-cir -o -
```
andykaylor: Tests looking for NYI messages have been problematic in the past. I'd just omit the test

clang/lib/CIR/CodeGen/CIRGenFunction.cpp
```diff
@@ -777,6 +780,8 @@ cir::FuncOp CIRGenFunction::generateCode(clang::GlobalDecl gd, cir::FuncOp fn,
 
     // Emit the standard function prologue.
     startFunction(gd, retTy, fn, funcType, args, loc, bodyRange.getBegin());
+    if (cgm.getDiags().hasErrorOccurred())
```
andykaylor: This seems problematic. Is it just trying to avoid a crash when we've issued an NYI message?

adams381: Replaced with a per-function skipFunctionBody flag: set only when this function reports the FENV NYI, cleared at each generateCode entry. emitStmt returns early once the flag is set. Fixes the parallel.c and HLSL CI regressions from TU-wide hasErrorOccurred().

clang/test/CIR/CodeGen/fp-pragma-fenv-access.c
adams381: Removed fp-pragma-fenv-access.c.

clang/lib/CIR/CodeGen/CIRGenFunction.cpp
```diff
@@ -777,6 +783,8 @@ cir::FuncOp CIRGenFunction::generateCode(clang::GlobalDecl gd, cir::FuncOp fn,
 
     // Emit the standard function prologue.
     startFunction(gd, retTy, fn, funcType, args, loc, bodyRange.getBegin());
+    if (skipFunctionBody)
```
andykaylor: I don't understand why you're trying to do this.

adams381: Dropped skipFunctionBody. startFunction calls errorNYI for UsesFPIntrin/StrictFP; generateCode returns right after the prologue on the same predicate so we never emit the body for that function. TU-wide hasErrorOccurred() was what broke parallel.c; this stays per-function without a stmt flag. NYI test removed.

clang/lib/CIR/CodeGen/CIRGenFunction.cpp
```diff
@@ -777,6 +780,8 @@ cir::FuncOp CIRGenFunction::generateCode(clang::GlobalDecl gd, cir::FuncOp fn,
 
     // Emit the standard function prologue.
     startFunction(gd, retTy, fn, funcType, args, loc, bodyRange.getBegin());
+    if (funcDecl->UsesFPIntrin() || funcDecl->hasAttr<StrictFPAttr>())
+      return fn;
```
andykaylor, line 782: This should be calling errorNYI, right?

andykaylor (approved):
lgtm

andykaylor:
The description and title of this PR need to be updated.

PR #198363, opened 2026-05-18
[CIR] Skip trivially-recursive available_externally function bodies

adams381:
CIR was emitting `available_externally` bodies for glibc-style inline
wrappers whose sole call is back to the same asm-named symbol (via
`__builtin_*`).  LLVM then treats the function as non-terminating and
can fold away surrounding null checks — the same failure mode as
classic CodeGen PR9614 (basename-style `if (cwd)` paths).

Port `isTriviallyRecursive` / `shouldEmitFunction` from
`CodeGenModule`, including the `isInlineBuiltinDeclaration` exemption,
and skip emitting those definitions.  The regression test uses the
`strrchr` wrapper pattern and checks that no `available_externally`
`strrchr` body appears in CIR or LLVM output.

### DIFF
--- a/clang/include/clang/AST/Decl.h
+++ b/clang/include/clang/AST/Decl.h
@@ -77,6 +77,7 @@ class TemplateArgumentList;
 class TemplateArgumentListInfo;
 class TemplateParameterList;
 class TypeAliasTemplateDecl;
+class MangleContext;
 class UnresolvedSetImpl;
 class VarTemplateDecl;
 enum class ImplicitParamKind;
@@ -2647,6 +2648,12 @@ public:
 
   /// Determine if this function provides an inline implementation of a builtin.
   bool isInlineBuiltinDeclaration() const;
+
+  /// Return true if this function's body contains only a direct call back to
+  /// the same symbol via an asm label or a __builtin_* alias.  Such
+  /// available_externally bodies are not valid stand-ins for the real
+  /// implementation (PR9614) and should be dropped before codegen.
+  bool isTriviallyRecursive(MangleContext &MC) const;
 
   /// Determine whether this is a destroying operator delete.
   bool isDestroyingOperatorDelete() const;
--- a/clang/include/clang/AST/Decl.h
+++ b/clang/include/clang/AST/Decl.h
@@ -77,7 +77,6 @@ class TemplateArgumentList;
 class TemplateArgumentListInfo;
 class TemplateParameterList;
 class TypeAliasTemplateDecl;
-class MangleContext;
 class UnresolvedSetImpl;
 class VarTemplateDecl;
 enum class ImplicitParamKind;
@@ -2648,12 +2647,6 @@ public:
 
   /// Determine if this function provides an inline implementation of a builtin.
   bool isInlineBuiltinDeclaration() const;
-
-  /// Return true if this function's body contains only a direct call back to
-  /// the same symbol via an asm label or a __builtin_* alias.  Such
-  /// available_externally bodies are not valid stand-ins for the real
-  /// implementation (PR9614) and should be dropped before codegen.
-  bool isTriviallyRecursive(MangleContext &MC) const;
 
   /// Determine whether this is a destroying operator delete.
   bool isDestroyingOperatorDelete() const;
--- a/clang/lib/CIR/CodeGen/CIRGenModule.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenModule.cpp
@@ -1579,12 +1579,30 @@ void CIRGenModule::emitGlobalVarDefinition(const clang::VarDecl *vd,
     emitCXXGlobalVarDeclInitFunc(vd, gv, needsGlobalCtor);
 }
 
+bool CIRGenModule::shouldEmitFunction(GlobalDecl gd) {
+  if (getFunctionLinkage(gd) !=
+      cir::GlobalLinkageKind::AvailableExternallyLinkage)
+    return true;
+
+  const auto *fd = cast<FunctionDecl>(gd.getDecl());
+  // Inline builtins must be emitted; the body is redirected to a `.inline`
+  // symbol in CIRGenFunction::generateCode.
+  if (fd->isInlineBuiltinDeclaration())
+    return true;
+
+  // PR9614 / glibc btowc workaround: an available_externally function whose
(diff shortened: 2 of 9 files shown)


clang/lib/CIR/CodeGen/CIRGenModule.cpp
```diff
@@ -1514,12 +1515,84 @@ void CIRGenModule::emitGlobalVarDefinition(const clang::VarDecl *vd,
     emitCXXGlobalVarDeclInitFunc(vd, gv, needsGlobalCtor);
 }
 
+namespace {
+/// Visits the body of a function looking for a direct call back to the
+/// symbol the function will end up linking as.  Matches classic CodeGen's
```
andykaylor: We should refactor the classic codegen implementation in some way that would allow it to be shared with CIR. There's nothing IR-specific here.

clang/lib/CIR/CodeGen/CIRGenModule.cpp
```diff
@@ -1514,12 +1515,84 @@ void CIRGenModule::emitGlobalVarDefinition(const clang::VarDecl *vd,
...
+  bool VisitStmt(const clang::Stmt *s) {
+    for (const clang::Stmt *child : s->children())
+      if (child && this->Visit(child))
+        return true;
+    return false;
+  }
+};
+} // namespace
+
+bool CIRGenModule::isTriviallyRecursive(const FunctionDecl *fd) {
```
andykaylor: Could this be moved into `FunctionDecl`? I believe the MangleContext ultimately comes from ASTContext.

clang/lib/CIR/CodeGen/CIRGenModule.cpp
adams381: Agree this belongs in shared frontend logic rather than a CIR-only visitor. I kept the classic shape here to unblock PR9614; happy to follow up with a refactor that both codegens call into.

clang/lib/CIR/CodeGen/CIRGenModule.cpp
adams381: Will look at hoisting onto FunctionDecl / ASTContext. The mangle/name walk is the same predicate classic uses today; I wanted to avoid widening this PR beyond the compile-time win.

clang/lib/CIR/CodeGen/CIRGenModule.cpp
adams381: Done. Moved `FunctionIsDirectlyRecursive` and `isTriviallyRecursive` into `FunctionDecl::isTriviallyRecursive(MangleContext &)` in `clang/lib/AST/Decl.cpp` (declared in `Decl.h`). Both classic CodeGen and CIR now call `fd->isTriviallyRecursive(getCXXABI().getMangleContext())` — the struct and the per-codegen impls are gone.

clang/lib/CIR/CodeGen/CIRGenModule.cpp
adams381: Yes — moved onto `FunctionDecl`. The method takes `MangleContext &` (both codegens already hold one via their CXX ABI) and pulls `BuiltinInfo` from `getASTContext()`. No IR-layer types involved.

adams381:
@andykaylor — `isTriviallyRecursive` is on `FunctionDecl` now; classic CodeGen and CIRGen both call it. Ready for another look when you have a moment.

adams381:
Ping.

clang/include/clang/AST/Decl.h
```diff
@@ -2648,6 +2649,12 @@ class FunctionDecl : public DeclaratorDecl,
   /// Determine if this function provides an inline implementation of a builtin.
   bool isInlineBuiltinDeclaration() const;
 
+  /// Return true if this function's body contains only a direct call back to
+  /// the same symbol via an asm label or a __builtin_* alias.  Such
+  /// available_externally bodies are not valid stand-ins for the real
+  /// implementation (PR9614) and should be dropped before codegen.
+  bool isTriviallyRecursive(MangleContext &MC) const;
```
erichkeane: We probably shouldn't be doing mangle-contexts inside of our declarations here, thats a bit of a layering violation.  Perhaps this is a valuable function to put in the `MangleContext` itself?

clang/test/CIR/CodeGen/trivially-recursive-skip.cpp
```diff
@@ -0,0 +1,35 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck --check-prefix=CIR --input-file=%t.cir %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-llvm -O2 -disable-llvm-passes %s -o %t-cir.ll
+// RUN: FileCheck --check-prefix=LLVM --input-file=%t-cir.ll %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -emit-llvm -O2 -disable-llvm-passes %s -o %t.ll
+// RUN: FileCheck --check-prefix=OGCG --input-file=%t.ll %s
```
erichkeane: Why a separate OGCG?  Didn't we discuss this before?

clang/include/clang/AST/Decl.h
adams381: Moved it onto `MangleContext`. Both CodeGen and CIRGen now call `getMangleContext().isTriviallyRecursive(FD)`, and `Decl.h` no longer pulls it in.

clang/test/CIR/CodeGen/trivially-recursive-skip.cpp
adams381: Collapsed it. The CIR and classic `-emit-llvm` output is identical for this absence check, so a single `LLVM` prefix now runs on both files.

adams381:
@erichkeane — `isTriviallyRecursive` lives on `MangleContext` now, so `Decl.h` no longer pulls in mangling and both CodeGen and CIRGen call `getMangleContext().isTriviallyRecursive(FD)`. Dropped the separate OGCG prefix too. Ready for another look.

erichkeane (approved):

PR #198365, opened 2026-05-18
[PDB][llvm-pdbutil] Add DXContainer support for yaml2pdb

kuilpd:
Create a PDB file from YAML with an built in `DXContainer`.
When creating a PDB with a `DXContainer`, streams DBI, TPI and IPI can be completely empty, so this patch also includes adjustments to allow forming such a PDB file. This is done to maintain compatibility with other DirectX tools.
This patch enables `dxcontainer.test` introduced in #198351.


jh7370:
Is this PR intended for review currently?

kuilpd:
> Is this PR intended for review currently?

It is now, there was a bug that buildbots caught.
This PR starts with #198351.

PR #198366, opened 2026-05-18
[Instrumentor] Improve stub printer (for C/C++ and value packs)

jdoerfert:
The stub printer now emits a helper header to deal with value packs (in C and C++). We also make the files C/C++ compatible and use the proper format strings for int32_t and int64_t.


llvm/lib/Transforms/IPO/InstrumentorStubPrinter.cpp
```diff
@@ -171,6 +226,316 @@ IRTCallDescription::createCSignature(const InstrumentationConfig &IConf) const {
...
+  OS << "template <typename T>\n";
+  OS << "inline const T *getValueAs(const void *pack_ptr, uint32_t "
+        "num_elements, uint32_t index) {\n";
+  OS << "  return static_cast<const T*>(getValuePackEntry(pack_ptr, "
+        "num_elements, index, nullptr));\n";
+  OS << "}\n\n";
+
+  OS << "#endif // __cplusplus\n\n";
+
+  OS << "#endif // INSTRUMENTOR_RUNTIME_H\n";
```
kevinsala, line 536: I would prefer to follow a different path for the future: have a real header, let CMake stringify it and set into constant string in a `.inc` file, and print that string here instead.

But, given that I don't know exactly how difficult is that, I'm fine with the current approach.

jdoerfert: I'm onboard if you or @EthanLuisMcDonough might want to give it a shot.

jdoerfert:
Any other comment?

llvm/test/Instrumentation/Instrumentor/default_rt.h
```diff
@@ -0,0 +1,264 @@
...
+}
+
+/// Move to the next value in the pack.
+static inline void nextValuePack(ValuePackIterator *iter) {
+  if (iter->index >= iter->count) {
+    iter->current = NULL;
+    return;
+  }
+  ValuePackHeader header = getValuePackHeader(iter);
+  uint32_t padding = (8 - (header.size % 8)) % 8;
```
kevinsala, line 83: Where does this 8 byte come from?

llvm/lib/Transforms/IPO/InstrumentorStubPrinter.cpp
kevinsala: We can make that in a follow-up PR.

llvm/test/Instrumentation/Instrumentor/rt_config.json
```diff
@@ -28,8 +163,31 @@
...
     },
+    "ptrtoint": {
+      "enabled": true,
+      "filter": "",
+      "filter.description": "Static property filter to exclude instrumentation.",
+      "pointer": true,
+      "pointer.description": "Input pointer of the ptr to int.",
+      "id": true,
+      "id.description": "A unique ID associated with the given instrumentor call"
+    },
```
EthanLuisMcDonough: `generate_rt.ll` fails because `ptrtoint` instrumentation isn't in this PR.

llvm/test/Instrumentation/Instrumentor/default_rt.c
```diff
@@ -0,0 +1,124 @@
...
+
+void *__instrumentor_pre_load(void *pointer, int32_t pointer_as, int64_t value_size, int64_t alignment, int32_t value_type_id, int32_t atomicity_ordering, int8_t sync_scope_id, int8_t is_volatile, int32_t id) {
+  printf("load pre -- pointer: %p, pointer_as: %" PRId32 ", value_size: %" PRId64 ", alignment: %" PRId64 ", value_type_id: %" PRId32 ", atomicity_ordering: %" PRId32 ", sync_scope_id: %" PRId32 ", is_volatile: %" PRId32 ", id: %" PRId32 "\n", pointer, pointer_as, value_size, alignment, value_type_id, atomicity_ordering, sync_scope_id, is_volatile, id);
+  return pointer;
+}
+
+int64_t __instrumentor_pre_alloca(int64_t size, int64_t alignment, int32_t id) {
+  printf("alloca pre -- size: %" PRId64 ", alignment: %" PRId64 ", id: %" PRId32 "\n", size, alignment, id);
+  return size;
+}
```
EthanLuisMcDonough: `diff -b default_rt.c %S/default_rt.c` fails on my machine because `alloca` functions are printed before any of the other instruction functions.

llvm/test/Instrumentation/Instrumentor/default_rt.h
jdoerfert: 8 bytes is the static payload size of the value pack. If the payload is smaller, we add padding to get the same layout every time.

kevinsala (approved):
LGTM.

PR #198368, opened 2026-05-18
[Clang][CodeGen] Fix assertion in withReturnValueSlot for incomplete types w/ de82b47

jsji:
de82b47 ([Clang] Fix sret AS for non-trivial-copy returns) introduced
calls to CXXRecordDecl::hasTrivialCopyConstructor() and similar methods
without checking that the record has a definition.

Guard the trivial-copy queries with RD->hasDefinition(). If the record
has no definition, conservatively set CanAggregateCopy to false, which
is safe — the old isTriviallyCopyableType path returned false for
incomplete types too.

### DIFF
--- a/clang/lib/CodeGen/CGExprAgg.cpp
+++ b/clang/lib/CodeGen/CGExprAgg.cpp
@@ -294,10 +294,11 @@ void AggExprEmitter::withReturnValueSlot(
   LangAS SRetLangAS = CGF.CGM.getTargetCodeGenInfo().getSRetAddrSpace(RD);
   unsigned SRetAS = CGF.getContext().getTargetAddressSpace(SRetLangAS);
   bool CanAggregateCopy =
-      RD ? (RD->hasTrivialCopyConstructor() ||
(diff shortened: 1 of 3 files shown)


efriedma-quic:
Do you have a testcase?  I'm not sure how we can generate code for a struct without a definition; we don't even know the size.

jsji:
> Do you have a testcase? I'm not sure how we can generate code for a struct without a definition; we don't even know the size.

This happens where there are builtins with `CustomTypeChecking` which caused early return, a test builtin to show the issue is in this draft PR. https://github.com/llvm/llvm-project/pull/198430

efriedma-quic:
I would rather say that's a bug in Sema.

Having a call which returns an incomplete type is fundamentally problematic: it's not clear what it means to "return" a value which we don't know how to represent.  And the use-case here isn't really compelling: for custom type-checking of intrinsics, the checking code can explicitly set the return type of the intrinsic to a complete type (or void).

PR #198369, opened 2026-05-18
[Instrumentor] Provide source location to runtime calls

jdoerfert:
To allow runtime calls to inspect the source location of the instrumentation opportunity, we encode it in the module. This allows the use in all environments, e.g., on GPUs, which might lack runtime dwarf reading or libunwind. The stub printer is extended to make handling with the encoded location information easy.


llvm/lib/Transforms/IPO/Instrumentor.cpp
```diff
@@ -399,6 +399,36 @@ bool InstrumentorImpl::instrumentModule() {
...
+    }
+    V = GV;
+  };
+
+  // Create the location table global (static/internal linkage)
+  auto *ArrayTy = ArrayType::get(IIRB.Int64Ty, IIRB.LocationEncoding.size());
+  MakeAndReplaceGlobal(ArrayTy, IIRB.LocationGlobal, "locations",
+                       ConstantArray::get(ArrayTy, IIRB.LocationEncoding));
+
+  // Create the string table global (static/internal linkage)
```
```suggestion
  // Create the string table global (static/internal linkage).
```

PR #198371, opened 2026-05-18
[libunwind][PPC64] Fix unw_getcontext corrupting callee-saved VSX registers on LE

pkubaj:
This is the first of two independent fixes for libunwind on ppc64le
(ELFv2 ABI, little-endian), where two separate bugs together cause
SIGSEGV during backtracing. This commit addresses the VSX register
corruption; the TOC-restore fault is handled in a follow-up. Both
were discovered while debugging lang/rust build failures with
RUST_BACKTRACE=1 on FreeBSD/powerpc64le (IBM POWER9).

On ppc64le, `unw_getcontext` saves each VS register with an in-place
`xxswapd n, n` followed by `stxvd2x`. The swap is needed because
`stxvd2x` stores doublewords in the wrong order on LE. However, the
macro never applies a second `xxswapd` to restore the register after
the store, so all 64 VS registers are permanently corrupted on return
from `unw_getcontext`.

This affects every callee-saved VSX register: f14-f31 (VSR14-VSR31)
and VR20-VR31 (VSR52-VSR63). After `_Unwind_Backtrace` returns, any
code that uses these registers sees wrong values. In practice this
manifests as SIGSEGV inside hashbrown's `reserve_rehash`: VR20-VR31
are corrupted before a SIMD comparison loop runs, producing an
out-of-bounds access.

Fix: add a second `xxswapd n, n` after the `stxvd2x` store. Since
`xxswapd` is its own inverse, the pair is a no-op on the architectural
register while still writing the correctly byte-swapped value to memory.

### DIFF
--- a/libunwind/src/UnwindRegistersSave.S
+++ b/libunwind/src/UnwindRegistersSave.S
@@ -424,9 +424,13 @@ LnoR2Fix:
 // register in the incorrect doubleword order.
 // FIXME: when supporting targets older than Power9 on LE is no longer required
 //        this can be changed to simply `stxv n, 16 * n(4)`.
+// Restore the register after storing: xxswapd is its own inverse, so a second
+// xxswapd undoes the in-place corruption of callee-saved VSRs (f14-f31,
+// VR20-VR31) that would otherwise persist after unw_getcontext returns.
 #define PPC64_STVS(n)      \
   xxswapd n, n            ;\
   stxvd2x n, 0, 4         ;\
+  xxswapd n, n            ;\
   addi    4, 4, 16
 #else
 #define PPC64_STVS(n)      \


chmeeedalf:
@nemanjai @hfinkel or any other ppc64 expert want to take a look?

pkubaj:
Is there anything blocking this patch from review? It seems pretty straightforward and the same issue probably affects also other operating systems.

jroelofs:
The commit message, and commits on the branch suggest to me that this ought to be two separate PRs, since the llvm repo requires squashed commits.

pkubaj:
PR updated to contain only the first of two fixes. When it's merged, I'll send a 2nd PR.

nemanjai (approved):
LGTM of course.

pkubaj:
So is there anything else to do in this PR for me?

jrtc27:
Your PR summary (which will become the commit message) seems to be using `*`s for register ranges? That's an odd convention, why not `-`? Also given it will end up as the commit message putting the CC in is not appropriate (in future, either make it a separate comment or add as a reviewer).

jrtc27:
> Safe for big-endian ppc64: the PPC64_STVS macro is
> inside #ifdef __LITTLE_ENDIAN__ and is not reached on BE.

I would hope so, otherwise the existing code is wrong. So not sure what value this comment is adding.

pkubaj:
OK, I thought the commit message that will be used is my actual commit message, not the PR description.

pkubaj:
Description changed.

erikarn:
hi! is there anything else needed here before this is merged? There's a bunch of us doing ppc64be/ppc64le work on freebsd and we'd love for all of this to be fixed upstream! thanks!

MaskRay (approved):

PR #198372, opened 2026-05-18
[MLIR][XeGPU] Update WG Level dpas_mx integration test.

silee2:
Add host code for testing kernel with simple values. 
Add result comparison using manually computed reference result.


akroviakov (approved):
I think a test with values and/or scales other than 1 (even if still uniform) would be more illustrative.

silee2:
Replaced by https://github.com/llvm/llvm-project/pull/201680

PR #198373, opened 2026-05-18
[AMDGPU] Fix BITOP3 matcher for op(X, X) operands

aobolensk:
Both operands share a src slot, and recursing into operand may replace that slot via the "replace parent operator" path, invalidating the truth-table bits cached for the other use. Handle AND/OR/XOR(X, X) directly: AND(X, X) = OR(X, X) = X, XOR(X, X) = 0


llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
```diff
@@ -4225,6 +4225,17 @@ static std::pair<unsigned, uint8_t> BitOp3_Op(Register R,
     Register LHS = getSrcRegIgnoringCopies(MI->getOperand(1).getReg(), MRI);
     Register RHS = getSrcRegIgnoringCopies(MI->getOperand(2).getReg(), MRI);
 
+    // op(X, X): both sides would share a Src slot, and recursing into X may
+    // replace that slot via the "replace parent operator" path, invalidating
+    // the truth-table bits cached for the other use.
+    if (LHS == RHS) {
```
arsenm, line 4308: Is this missing an isKnownNotToBeUndef check too?

llvm/test/CodeGen/AMDGPU/bitop3-op-self.ll
```diff
@@ -0,0 +1,71 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
+; RUN: llc -O0 -global-isel=0 -mtriple=amdgcn-- -mcpu=gfx950 < %s | FileCheck -enable-var-scope -check-prefixes=GFX950-SDAG %s
+; RUN: llc -O0 -global-isel -new-reg-bank-select -mtriple=amdgcn-- -mcpu=gfx950 < %s | FileCheck -enable-var-scope -check-prefixes=GFX950-GISEL %s
```
arsenm: No -O0?

llvm/test/CodeGen/AMDGPU/bitop3-op-self.ll
```diff
@@ -0,0 +1,71 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
+; RUN: llc -O0 -global-isel=0 -mtriple=amdgcn-- -mcpu=gfx950 < %s | FileCheck -enable-var-scope -check-prefixes=GFX950-SDAG %s
```
arsenm: Can you just add this to an existing bitop3 test

llvm/test/CodeGen/AMDGPU/bitop3-op-self.ll
aobolensk: Automatically resolved after merging the test cases into `llvm/test/CodeGen/AMDGPU/bitop3.ll`

llvm/test/CodeGen/AMDGPU/bitop3-op-self.ll
aobolensk: Done, thanks

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
aobolensk: Yes, added, thanks

arsenm: This needs new backing noundef + not-noundef test cases. Right now all of these will fail just due to the missing noundef without all of the other checks

aobolensk: Added, please check

llvm/test/CodeGen/AMDGPU/bitop3.ll
```diff
@@ -856,6 +856,116 @@ define amdgpu_ps i32 @or_self_complex(i32 %a, i32 %b, i32 %c) #0 {
   ret i32 %res
 }
 
+define amdgpu_ps i32 @xor_self_complex_noundef(i32 noundef %a, i32 noundef %b, i32 noundef %c) #0 {
```
arsenm: Don't use amdgpu_ps? This is using a readfirstlane return

aobolensk: I have dropped amdgpu_ps CC

shiltian:
https://github.com/llvm/llvm-project/pull/198373 and https://github.com/llvm/llvm-project/pull/198419, they seem to do the same thing.

aobolensk:
> #198373 and #198419, they seem to do the same thing.

@shiltian As it was mentioned by the authors of https://github.com/llvm/llvm-project/pull/198419 and https://github.com/llvm/llvm-project/pull/198556, there is no intersection between this patch and theirs

aobolensk:
@carlobertolli please take a look

llvm/test/CodeGen/AMDGPU/bitop3.ll
```diff
@@ -741,6 +741,221 @@ define amdgpu_ps half @test_and_or_b16(i16 %a, i16 %b, i16 %c) {
...
+; GFX1250-GISEL-NEXT:    s_wait_kmcnt 0x0
+; GFX1250-GISEL-NEXT:    v_bitop3_b32 v0, v0, v2, v1 bitop3:0x6c
+; GFX1250-GISEL-NEXT:    s_set_pc_i64 s[30:31]
+  %t1 = and i32 %a, %b
+  %xor1 = xor i32 %t1, %c
+  %res = or i32 %xor1, %xor1
+  ret i32 %res
+}
+
+define i32 @xor_self_complex_noundef(i32 noundef %a, i32 noundef %b, i32 noundef %c) #0 {
```
arsenm: Probably better to mark everything with noundef and have the special case test for maybe-undef

aobolensk: Done, removed extra functions and left only two: with undefs and normal values (for maybe undef case)

aobolensk: @arsenm could you please check?

aobolensk:
@carlobertolli could you please take a look?

carlobertolli:
> @carlobertolli could you please take a look?

Using latest trunk from last Friday, I tested the fuzzing bug reports this PR is targeting and it looks like they do not reproduce.
My suggestion is to first add the lit tests in this patch in a separate PR and only if those fail locally, fix them.
Also, the lit tests for bitop3 with shared args are now located in bitop3-shared-operand.ll.

aobolensk:
> > @carlobertolli could you please take a look?
> 
> Using latest trunk from last Friday, I tested the fuzzing bug reports this PR is targeting and it looks like they do not reproduce. My suggestion is to first add the lit tests in this patch in a separate PR and only if those fail locally, fix them. Also, the lit tests for bitop3 with shared args are now located in bitop3-shared-operand.ll.

I confirm, it is not reproducible so far. Probably, https://github.com/llvm/llvm-project/pull/198556 was the key to fix that. Thus closing this PR, thanks!

PR #198374, opened 2026-05-18
[SimplifyCFG] Improve reduceSwitchRange to avoid unnecessary subtractions

dakersnar:
Dense switch cases can sometimes be achieved with only a shift, avoiding a subtract. Change reduceSwitchRange to check for such cases before falling back to using a subtract.

The changes to test9 demonstrates that for the purpose of lookup-table lowering (triggered by the switch-to-lookup flag), the switch range still ends up normalized to start at 0, the subtraction just happens after the fshl.

test10 demonstrates where this change is useful, resulting in a dense switch case with one less subtract.

test11 demonstrates cases where the subtract is still needed in order to form a dense switch case.


llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7582,31 +7582,35 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
   if (isSwitchDense(Values))
     return false;
 
-  // First, transform the values such that they start at zero and ascend.
+  auto TryBase = [&](int64_t CandidateBase, unsigned &CandidateShift) {
```
antoniofrighetto: I may suggest returning a std::optional<unsigned> instead of passing CandidateShift as a reference (thus returning std::nullopt if the new values are not dense); would make the call-sites self-documented.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7582,31 +7582,35 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
...
+    }
+
+    CandidateShift = llvm::countr_zero(ReducedBits);
+    if (CandidateShift >= 64)
+      return false;
+    if (CandidateShift > 0)
+      for (auto &V : ReducedValues)
+        V = (int64_t)((uint64_t)V >> CandidateShift);
+
+    return isSwitchDense(ReducedValues);
```
akshayrdeodhar: Can we separate isSwitchDense and TryBase, instead have a lambda which returns the best possible (Reduced_Values, Shift)? Probably just a nit, switchDense always called in the same context might be easier to understand.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7621,8 +7625,9 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
 
   auto *Ty = cast<IntegerType>(SI->getCondition()->getType());
   Builder.SetInsertPoint(SI);
-  Value *Sub =
-      Builder.CreateSub(SI->getCondition(), ConstantInt::getSigned(Ty, Base));
+  Value *Sub = SI->getCondition();
+  if (Base != 0)
+    Sub = Builder.CreateSub(Sub, ConstantInt::getSigned(Ty, Base));
   Value *Rot = Builder.CreateIntrinsic(
```
akshayrdeodhar, line 7735: This is really confusing. Why not Builder.CreateShr(Sub, Shift) instead?
Here's an alive2 proof for i8s: https://alive2.llvm.org/ce/z/2YgzPQ

dakersnar: The fshl is intentional here: it acts as a rotate-right by Shift, preserving the shifted-out low bits in the high bits. That makes non-multiples of the stride miss the reduced switch range. A plain lshr would miscompile e.g. case values {32,64,96,128} for input 33, which would shift to 1 and hit the 32 case instead of default.

The reason your alive2 case passes is because of the llvm.assume. And FWIW, the compiler does optimize fshr to lshr in some cases in I believe InstCombine, when it can prove it is allowed to.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
dakersnar: Good idea, implemented.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
dakersnar: I think the TryBase logic is helpful as a mechanical way to reuse code. I do like the idea of separating the "find the optimal reduction" logic from the "does this base produce a dense range" logic. I also like the idea of being more explicit about what exactly we are getting out of this function, which is (Base, Shift). Pushed a refactor that seems cleaner to me, let me know what you think. I could see this being one too many layers of lambda.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7582,31 +7582,50 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
...
+    SmallVector<int64_t, 4> ReducedValues;
+    ReducedValues.reserve(Values.size());
+    uint64_t ReducedBits = 0;
+    for (int64_t V : Values) {
+      uint64_t Reduced = (uint64_t)V - (uint64_t)CandidateBase;
+      ReducedBits |= Reduced;
+      ReducedValues.push_back((int64_t)Reduced);
+    }
+
+    unsigned CandidateShift = llvm::countr_zero(ReducedBits);
```
akshayrdeodhar: Here.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
akshayrdeodhar: In this case, won't we get the llvm.assume due to https://github.com/llvm/llvm-project/pull/198374/changes#r3279403350? The number of bits we right shift by is equal to the number of 0 bits starting from the LSB? I might be confused about what `countr_zero` does.

dakersnar: Consider this switch
```
  switch i32 %a, label %def [
  i32 32, label %one
  i32 64, label %two
  i32 96, label %three
  i32 128, label %four
  ]
```

There's two things at play here, the _set of comparison value constants_ ({32, 64, 96, 128}) and the _incoming variable comparison value_ (%a). Given the shared low zero bits in the _constants_, we can Shift them by 5. This is what the code you linked is checking. But %a = 33 is still a possible runtime switch input, and it must go to default, as it does before the transform. If we `lshr` it by 5 it will branch to `%one` in the reduced switch, which is incorrect.

The reduceSwitchRange transform is legal if:
1. the set of inputs `%a` that branched to `%one, %two, %three, %four` continue to do so.
2. the set of inputs `%a` that branched to `%def` continue to do so.

The code you linked ensures the former, but `fshr` is necessary to ensure the latter.

Separately, if we _can_ prove `%a` must have a certain low bits at 0, we can convert `fshr` to `lshr`, but that's a separate optimization that lives in InstCombine.

rnk: Very revealing, I remember being confused why we were doing a *funnel* shift *left* in the session where we discussed this bug, when shift right seemed like the operation we wanted to do. I guess that's canonicalization for you: never sub, always add, that kind of thing.

dakersnar: Yeah, it is a bit confusing. But yes, it's a funnel shift *left* by *bitwdith - shift*, which results in a shift *right* by *shift*. Disregard my uses of `fshr` above, although conceptually my explanation is still correct.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7582,31 +7582,50 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
...
-  // Now we have signed numbers that have been shifted so that, given enough
-  // precision, there are no negative values. Since the rest of the transform
-  // is bitwise only, we switch now to an unsigned representation.
-
-  // This transform can be done speculatively because it is so cheap - it
-  // results in a single rotate operation being inserted.
-
-  // countTrailingZeros(0) returns 64. As Values is guaranteed to have more than
-  // one element and LLVM disallows duplicate cases, Shift is guaranteed to be
-  // less than 64.
```
zmodem, line 7676: It would be good to preserve the information in these comments.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7582,31 +7582,50 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
...
-
-  if (!isSwitchDense(Values))
-    // Transform didn't create a dense switch.
+  struct RangeReduction {
+    int64_t Base;
+    unsigned Shift;
+  };
+
+  auto TryGetRangeReductionForBase =
+      [&](int64_t CandidateBase) -> std::optional<RangeReduction> {
```
zmodem: I don't think this references any local state (besides Values, which it doesn't modify). Could it be a regular function instead, taking Values and Base as arguments, and return Shift on success?

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7582,31 +7582,50 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
...
+    SmallVector<int64_t, 4> ReducedValues;
+    ReducedValues.reserve(Values.size());
```
zmodem: Would it be easier to make a copy of Values first, and then modify it (like the original code did)?

```
... ReducedValues = Values;
for (int64_t &V : ReducedValues) {
  ...
```

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7582,31 +7582,50 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
...
+    uint64_t ReducedBits = 0;
+    for (int64_t V : Values) {
+      uint64_t Reduced = (uint64_t)V - (uint64_t)CandidateBase;
+      ReducedBits |= Reduced;
+      ReducedValues.push_back((int64_t)Reduced);
+    }
+
+    unsigned CandidateShift = llvm::countr_zero(ReducedBits);
+    if (CandidateShift >= 64)
```
zmodem: Can this happen? The previous code had `assert(Shift < 64)`.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7582,31 +7582,50 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
...
+    // This avoids creating an unnecessary `(condition - local_min)` expression.
+    int64_t LocalMin = Values[0];
+    if (LocalMin >= 0) {
+      if (auto Reduction = TryGetRangeReductionForBase(0))
+        return Reduction;
+    }
+    return TryGetRangeReductionForBase(LocalMin);
+  };
+
+  std::optional<RangeReduction> Reduction = GetRangeReduction();
```
zmodem: Since the lambda is only called once and doesn't do very much, I'd just open-code it here.

llvm/test/Transforms/SimplifyCFG/rangereduce.ll
```diff
@@ -317,6 +317,109 @@ three:
   ret i32 99783
 }
 
+; Base=0 is profitable here: shifting out the common low zero bits produces a
+; dense range without subtracting the local minimum first.
```
zmodem: In the end, will this get lowered without the subtraction, or does that reappear in a range check like in test9 above?

llvm/test/Transforms/SimplifyCFG/rangereduce.ll
```diff
@@ -317,6 +317,109 @@ three:
...
+define void @test10(i32 %a) {
+; CHECK-LABEL: @test10(
+; CHECK-NEXT:    [[TMP1:%.*]] = call i32 @llvm.fshl.i32(i32 [[A:%.*]], i32 [[A]], i32 27)
```
zmodem: nit: Maybe just TMP if there's only one.

llvm/test/Transforms/SimplifyCFG/rangereduce.ll
```diff
@@ -286,12 +286,12 @@ three:
 
 define i32 @test9(i32 %a) {
 ; CHECK-LABEL: @test9(
-; CHECK-NEXT:    [[TMP1:%.*]] = sub i32 [[A:%.*]], 6
-; CHECK-NEXT:    [[TMP2:%.*]] = call i32 @llvm.fshl.i32(i32 [[TMP1]], i32 [[TMP1]], i32 31)
-; CHECK-NEXT:    [[TMP3:%.*]] = icmp ult i32 [[TMP2]], 8
+; CHECK-NEXT:    [[TMP2:%.*]] = call i32 @llvm.fshl.i32(i32 [[TMP1:%.*]], i32 [[TMP1]], i32 31)
```
zmodem: TMP1 here is the argument, a. It would be nice to refer to it as `A` like before.

llvm/test/Transforms/SimplifyCFG/rangereduce.ll
dakersnar: I think that depends on the target. The motivation of this change is to leave that decision to the target and to allow for later optimizations like CVP to more easily reason about these conditions, particularly conditions that are shared across multiple switch statements. I can elaborate if you'd like.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
dakersnar: Done. Added more comments too for clarity. Let me know if the new version reads clearer or needs further adjustments.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
dakersnar: Done. Let me know if the shape I've arrived at makes more sense.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
dakersnar: Done

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
dakersnar: Good point, it cannot. Added some additional asserts to ensure the helper function is called with intended restrictions on Values/Base as well.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
dakersnar: Done

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7582,31 +7616,23 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
...
-    for (auto &V : Values)
-      V = (int64_t)((uint64_t)V >> Shift);
-
-  if (!isSwitchDense(Values))
-    // Transform didn't create a dense switch.
+  std::optional<unsigned> Shift;
+  // Prefer Base=0 when the case values are still dense after shifting out their
+  // common low zero bits without subtracting a base. This avoids creating an
+  // unnecessary `(condition - local_min)` expression.
+  if (Base >= 0) {
```
zmodem: Should this check `!= 0` instead?

Also It seems more straight-forward to always try this (the Base=0) case first, without the if. Something like:

```
if (Shift = getDenseSwitchRangeReductionShift(Values, /*Base=*/0))
  Base = 0;
else if (Base != 0)
  Shift = getDenseSwitchRangeReductionShift(Values, Base);

if (!Shift)
  return false;
```

zmodem (commented):
Thanks! Just one more nit, but this seems good to me overall.

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
dakersnar: Good suggestion. Changed the helper function to return nullopt instead of asserting if Base isn't the <= every value in the switch range, and updated the callers.

zmodem (approved):
lgtm

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7637,34 +7672,24 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
...
-    // Transform didn't create a dense switch.
+  std::optional<unsigned> Shift;
+  // Prefer Base=0 when the case values are still dense after shifting out their
+  // common low zero bits without subtracting a base. This avoids creating an
+  // unnecessary `(condition - local_min)` expression.
+  if ((Shift = getDenseSwitchRangeReductionShift(Values, /*Base=*/0,
+                                                 /*OptSize=*/true)))
+    Base = 0;
+  else if (Base != 0)
+    Shift = getDenseSwitchRangeReductionShift(Values, Base, OptSize);
```
dakersnar: @aengelke do you have thoughts on how I should integrate your recently merged feature (https://github.com/llvm/llvm-project/pull/200664) into this change? The current approach is more conservative, it forces the Base=0 test to use the more conservative density metric, but I could also see the argument that it should just take in the OptSize argument like the second call.

aengelke: I don't particularly cared about this code path when lowering the density threshold (the primary motivation was the non-shifted case). Have you found cases where one or the other threshold yields better code?

dakersnar: No real world cases, but some unit test cases. Given that you have no preference, I think I lean towards the current conservative implementation to maintain existing behavior on this path. The benefit of being conservative is that the second check might end up being *more* dense than the Base=0 path.

For example, the test11 case has values {40, 72, 104, 136}. With Base=0, we shift by 3 and get {5, 9, 13, 17}, which is dense enough under the new 10% threshold but still has range 13. With Base=40, we shift by 5 and get {0, 1, 2, 3}, which is strictly tighter and preserves the previous behavior of subtracting the local minimum for this case.

So unless there is evidence that the wider Base=0 form is better, I think it makes sense to only prefer Base=0 when it satisfies the more conservative density threshold, and otherwise fall back to the local-min reduction using the normal density policy.

aengelke: Sounds good, but please add these considerations as a comment into the code.

dtcxzyw (approved):
LGTM

rnk (approved):

llvm/lib/Transforms/Utils/SimplifyCFG.cpp
```diff
@@ -7656,34 +7691,24 @@ static bool reduceSwitchRange(SwitchInst *SI, IRBuilder<> &Builder,
...
-  if (Shift > 0)
-    for (auto &V : Values)
-      V = (int64_t)((uint64_t)V >> Shift);
-
-  if (!isSwitchDense(Values, SI->getFunction()->hasOptSize()))
-    // Transform didn't create a dense switch.
+  std::optional<unsigned> Shift;
+  // Prefer Base=0 when the case values are still dense after shifting out their
+  // common low zero bits without subtracting a base. This avoids creating an
+  // unnecessary `(condition - local_min)` expression.
```
dakersnar:
```suggestion
  // Prefer Base=0 when shifting out common low zero bits still produces a dense
  // range, as this avoids an unnecessary `(condition - local_min)` expression.
  // However, avoiding the subtract can leave a wider reduced range than using
  // the local minimum, so require Base=0 to satisfy the stricter optsize
  // density threshold before falling back to the normal density policy for
  // local-min.
```

PR #198377, opened 2026-05-18
[lldb][PECOFF] Recognise truncated .lldb{summaries,formatters} section names

charles-zablit:
PE/COFF image-file section names are limited to 8 characters, so the linker truncates the longer source names. This patch matches both.

See the specs for more details: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers

This is a follow up to https://github.com/llvm/llvm-project/pull/197955.

rdar://177420528

### DIFF
--- a/lldb/source/Plugins/ObjectFile/PECOFF/ObjectFilePECOFF.cpp
+++ b/lldb/source/Plugins/ObjectFile/PECOFF/ObjectFilePECOFF.cpp
@@ -998,11 +998,14 @@ SectionType ObjectFilePECOFF::GetSectionType(llvm::StringRef sect_name,
           .Case(".debug", eSectionTypeDebug)
           .Case(".stabstr", eSectionTypeDataCString)
           .Case(".reloc", eSectionTypeOther)
-          // .eh_frame can be truncated to 8 chars.
+          // PE/COFF image-file section names are limited to 8 characters,
+          // so the linker truncates the longer source names. Match both.
           .Cases({".eh_frame", ".eh_fram"}, eSectionTypeEHFrame)
           .Case(".gosymtab", eSectionTypeGoSymtab)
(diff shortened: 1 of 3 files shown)


lldb/source/Plugins/ObjectFile/PECOFF/ObjectFilePECOFF.cpp
```diff
@@ -998,11 +998,14 @@ SectionType ObjectFilePECOFF::GetSectionType(llvm::StringRef sect_name,
           .Case(".debug", eSectionTypeDebug)
           .Case(".stabstr", eSectionTypeDataCString)
           .Case(".reloc", eSectionTypeOther)
-          // .eh_frame can be truncated to 8 chars.
+          // PE/COFF image-file section names are limited to 8 characters,
+          // so the linker truncates the longer source names. Match both.
           .Cases({".eh_frame", ".eh_fram"}, eSectionTypeEHFrame)
           .Case(".gosymtab", eSectionTypeGoSymtab)
```
DavidSpickett: Assuming the dot is included, this is 9 characters. So I would move this and anything that is naturally <= 8 up (swiftast?), which puts all the ones with a truncated form next to each other.

DavidSpickett (approved):
Follows the precedent set by `.eh_fram(e)`, so LGTM.

lldb/source/Plugins/ObjectFile/PECOFF/ObjectFilePECOFF.cpp
charles-zablit: The dot is included. I trimmed `.gotsymtab` as well.

I also grouped the `Cases` and the `Case`.

I'm not sure why `swiftast` does not have a leading dot here. According to https://stackoverflow.com/a/44031260, it's not mandatory, just a convention.

PR #198381, opened 2026-05-18
[lld][MachO] Fix SIGBUS crash in saveOrHardlinkBuffer

liza371:
This change removes a hardlink in saveOrHardlinkBuffer if the hardlink already exists.

On Mac, -object_path_lto files are hardlinked to the cache when possible. If the hardlink fails, the saveOrHardlinkBuffer method falls back to saveBuffer instead.

saveBuffer() opens the file that is being written to as a raw_fd_ostream object, which truncates a file when opening if the file already exists.

Most of the time this is not an issue, however, if the hardlink fails because it actually already exists, AND the hardlink exists specifically between the -object_path_lto file and the cache file, then when the file is opened and truncated, we also accidentally truncate the file we are trying to read from.

This fixes a recurring issue Chromium saw in buildbots that occassionally hit the above circumstances to cause the crash. Chromium bug for reference is crbug.com/40060675.


lld/test/MachO/lto-object-path.ll
```diff
@@ -37,6 +37,14 @@
 ; RUN: llvm-nm -pa %t/test | FileCheck %s --check-prefixes CHECK,OBJPATH-FILE -DFILE=%t/lto-tmp.o
 ; RUN: llvm-otool -l %t/lto-tmp.o | FileCheck %s --check-prefixes=MINOS
 
+;; Test that attempting to hardlink when cache and lto-temps are not empty does NOT cause a crash
+;; Link attempt must be done twice, first run populates lto-temps/ and cache/.
+; RUN: %python -c "print('target triple = \"x86_64-apple-macosx10.15.0\"'); print('target datalayout = \"e-m:o-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128\"'); print('@dummy_data = global [32768 x i8] c\"' + 'A'*32768 + '\"'); print('@llvm.used = appending global [1 x ptr] [ptr @dummy_data], section \"llvm.metadata\"')" > %t/large.ll
```
nico: Can you mention in the commit description why this needs a largeish array?

lld/MachO/LTO.cpp
```diff
@@ -85,6 +85,7 @@ static lto::Config createConfig() {
 // or `originalPath` is not set, saves `buffer` to `path`.
 static void saveOrHardlinkBuffer(StringRef buffer, const Twine &path,
                                  std::optional<StringRef> originalPath) {
+  fs::remove(path);
```
zmodem: Since this is subtle, it would be good to have a comment explaining why this is important.

Also, I guess we only need this if we're doing the hard link, i.e. should it go inside the `if (originalPath) {` branch?

lld/test/MachO/lto-object-path.ll
zmodem: fwiw, since the file doesn't end up that large, I think it would be okay to check it in under lld/test/MachO/Inputs/ and then we wouldn't need to shell out to python here.

lld/MachO/LTO.cpp
liza371: Added an explanatory comment, let me know if it doesn't make sense!

And yes you're right it makes more sense inside the `if (originalPath)` conditional so I moved it there.

lld/test/MachO/lto-object-path.ll
liza371: Okay I moved this to its own file in `lld/test/MachO/Inputs/`, the majority of the file is a bunch of A's to initialize the array, though. But I couldn't really find a better way to easily generate a relatively large file and the test itself is easier to read so I think it's fine?

liza371: Added a blurb in the commit description explaining the array size in the tests.

zmodem (approved):
lgtm

PR #198383, opened 2026-05-18
[VPlan] Fold canonical IV recipe creation into createLoopRegion.

fhahn:
Remove the separate addCanonicalIVRecipes transform and create the canonical IV's increment and the latch's exiting branch directly in createLoopRegion, using the loop region's VPRegionValue for the canonical IV. The temporary VPPhi placeholder previously inserted in the header is no longer needed.


llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -6669,7 +6667,8 @@ void LoopVectorizationPlanner::buildVPlans(ElementCount MinVF,
 
   RUN_VPLAN_PASS(VPlanTransforms::addMiddleCheck, *VPlan0,
                  CM.foldTailByMasking());
```
lukel97, line 6705: I presume these also need to be marked as RUN_PASS_NO_VERIFY now for the expensive checks builds?

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
```diff
@@ -467,45 +463,41 @@ static void createLoopRegion(VPlan &Plan, VPBlockBase *HeaderVPB) {
...
-  Builder.createNaryOp(VPInstruction::BranchOnCount,
-                       {CanonicalIVIncrement, &Plan.getVectorTripCount()},
-                       LatchDL);
+  if (HasBranchOnTwoConds) {
+    auto *IsLatchExitTaken = Builder.createICmp(
+        CmpInst::ICMP_EQ, CanonicalIVIncrement, &Plan.getVectorTripCount());
+    LatchTerm->setOperand(1, IsLatchExitTaken);
+  } else {
+    DebugLoc LatchDL = DL;
+    if (LatchTerm && match(LatchTerm, m_BranchOnCond())) {
```
lukel97: Preserve the original "We are about to replace the branch to exit the region. Remove the original BranchOnCond, if there is any." comment?

fhahn: Added in adjusted form, with https://github.com/llvm/llvm-project/pull/198539 we always have a terminator which simplifies things

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
fhahn: Should not be needed after https://github.com/llvm/llvm-project/pull/198539

llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -6645,10 +6645,8 @@ void LoopVectorizationPlanner::buildVPlans(ElementCount MinVF,
                       Config.getInLoopReductions(), Hints.allowReordering()))
     return;
 
-  RUN_VPLAN_PASS(VPlanTransforms::simplifyRecipes, *VPlan0);
-  RUN_VPLAN_PASS(VPlanTransforms::removeDeadRecipes, *VPlan0);
-  RUN_VPLAN_PASS(VPlanTransforms::addCanonicalIVRecipes, *VPlan0,
-                 getDebugLocFromInstOrOperands(Legal->getPrimaryInduction()));
+  RUN_VPLAN_PASS_NO_VERIFY(VPlanTransforms::simplifyRecipes, *VPlan0);
+  RUN_VPLAN_PASS_NO_VERIFY(VPlanTransforms::removeDeadRecipes, *VPlan0);
```
lukel97: Are these meant to be RUN_VPLAN_PASS after https://github.com/llvm/llvm-project/pull/198539

fhahn: Yep, remove the left-over _NO_VERIFY, thanks

lukel97 (approved):
LGTM

PR #198388, opened 2026-05-18
Update `QualType::get{Split,Atomic,}UnqualifiedType` with `const ASTContext &` parameter (NFC).

michele-scandale:
This is in preparation for changes to handle vector types with qualified element type which will require to potentially construct new vector types in order to minimize the amount of sugar removed.

### DIFF
--- a/clang/include/clang/AST/DeclCXX.h
+++ b/clang/include/clang/AST/DeclCXX.h
@@ -182,12 +182,15 @@ class CXXBaseSpecifier {
   /// range does not include the \c virtual or the access specifier.
   TypeSourceInfo *BaseTypeInfo;
 
+  const ASTContext *Ctx;
+
 public:
   CXXBaseSpecifier() = default;
-  CXXBaseSpecifier(SourceRange R, bool V, bool BC, AccessSpecifier A,
-                   TypeSourceInfo *TInfo, SourceLocation EllipsisLoc)
-    : Range(R), EllipsisLoc(EllipsisLoc), Virtual(V), BaseOfClass(BC),
-      Access(A), InheritConstructors(false), BaseTypeInfo(TInfo) {}
+  CXXBaseSpecifier(const ASTContext &Ctx, SourceRange R, bool V, bool BC,
+                   AccessSpecifier A, TypeSourceInfo *TInfo,
+                   SourceLocation EllipsisLoc)
+      : Range(R), EllipsisLoc(EllipsisLoc), Virtual(V), BaseOfClass(BC),
+        Access(A), InheritConstructors(false), BaseTypeInfo(TInfo), Ctx(&Ctx) {}
 
   /// Retrieves the source range that contains the entire base specifier.
   SourceRange getSourceRange() const LLVM_READONLY { return Range; }
@@ -247,7 +250,7 @@ public:
   ///
   /// This type will always be an unqualified class type.
   QualType getType() const {
-    return BaseTypeInfo->getType().getUnqualifiedType();
+    return BaseTypeInfo->getType().getUnqualifiedType(*Ctx);
   }
 
   /// Retrieves the type and source location of the base class.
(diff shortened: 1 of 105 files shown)


michele-scandale:
This change is mostly mechanical. Few places where the changes is not as trivial:
* `Type::getCanonicalTypeUnqualified` implementation
*  `CXXBaseSpecifier` now has an extra `const ASTContext *` field.

zeyi2 (approved):
clang-tidy part LGTM, thank you :)

(but please wait for other reviewers)

clang/include/clang/AST/DeclCXX.h
```diff
@@ -182,12 +182,15 @@ class CXXBaseSpecifier {
   /// range does not include the \c virtual or the access specifier.
   TypeSourceInfo *BaseTypeInfo;
 
+  const ASTContext *Ctx;
```
efriedma-quic, line 185: This significantly increases the size of records.. can we avoid this?  (For CXXBaseSpecifier::getType, we know that we're returning a record type, so we don't actually need to access the ASTContext to compute the unqualified version.)

clang/lib/Analysis/FlowSensitive/RecordOps.cpp
```diff
@@ -56,8 +56,9 @@ static void copySyntheticField(QualType FieldType, StorageLocation &SrcFieldLoc,
 
 void copyRecord(RecordStorageLocation &Src, RecordStorageLocation &Dst,
                 Environment &Env, const QualType TypeToCopy) {
-  auto SrcType = Src.getType().getCanonicalType().getUnqualifiedType();
-  auto DstType = Dst.getType().getCanonicalType().getUnqualifiedType();
+  const ASTContext &Ctx = Src.getType()->getAsRecordDecl()->getASTContext();
+  auto SrcType = Src.getType().getCanonicalType().getUnqualifiedType(Ctx);
```
efriedma-quic, line 60: getCanonicalTypeUnqualified()?  (Here, and in other similar places.)

clang/include/clang/AST/DeclCXX.h
erichkeane: This in particular seems completely unacceptable to me.  We need to find a way to accomplish whatever we are trying here without increasing the AST size this much.

erichkeane (commented):
So it isn't quite clear what you're trying to accomplish here.  But having the `getUnqualifiedType` functionality require an `ASTContext` seems unnecessary/not particularly acceptable.  Can we have a discussion/RFC about this to figure out what/why we are doing this? Is there perhaps a better way to accomplish whatever we're trying to accomplish here by doing this?

efriedma-quic:
This is connected to https://github.com/llvm/llvm-project/pull/197624 .  The idea there is to move qualifiers out of the element type of vectors (sort of like what we do for arrays, but in the opposite direction).  That means for non-canonical types, computing the "unqualified type" can require returning a different VectorType*, hence the ASTContext.

Thinking about this a bit more, it might make sense to just stick a pointer into VectorType: in additional to the CanonicalType which we store for all types, also store the UnqualifiedType which has all element qualifiers stripped.  The memory usage is likely negligible given that most code uses very few distinct vector types.

erichkeane:
> This is connected to #197624 . The idea there is to move qualifiers out of the element type of vectors (sort of like what we do for arrays, but in the opposite direction). That means for non-canonical types, computing the "unqualified type" can require returning a different VectorType*, hence the ASTContext.
> 
> Thinking about this a bit more, it might make sense to just stick a pointer into VectorType: in additional to the CanonicalType which we store for all types, also store the UnqualifiedType which has all element qualifiers stripped. The memory usage is likely negligible given that most code uses very few distinct vector types.

While reading your first paragraph, I very much was thinking the 2nd paragraph.  VectorType can/should just store its unqualified type if that is what you need.  It is a MUCH smaller AST impact than this (as you said, vector types used in most projects are measured in 'dozens').

michele-scandale:
> Thinking about this a bit more, it might make sense to just stick a pointer into VectorType: in additional to the CanonicalType which we store for all types, also store the UnqualifiedType which has all element qualifiers stripped. The memory usage is likely negligible given that most code uses very few distinct vector types.

If it is acceptable to add members to `VectorType` (and other vector related type classes) then I'd rather just have a `const ASTContext &` similarly to what has been done for `TypeOfExprType`.

PR #198389, opened 2026-05-18
[InstCombine] Do not crash in compare of bitcast pattern

daniel-donenfeld:
The pattern matching `fcmp oeq/une (bitcast X), 0.0` assumes that X is an integer, but it is possible to cast between float types. Add a check for X being an integer type to fix the compiler crashing on this pattern. A lit test for this failure is also added.


llvm/test/Transforms/InstCombine/bfloat-bitcast-fcmp-zero.ll
```diff
@@ -0,0 +1,14 @@
+; NOTE: Assertions have been autogenerated by utils/update_test_checks.py UTC_ARGS: --version 6
+; RUN: opt < %s -passes=instcombine -S | FileCheck %s
+
+define i1 @f(bfloat %x) {
+; CHECK-LABEL: define i1 @f(
+; CHECK-SAME: bfloat [[X:%.*]]) {
+; CHECK-NEXT:    [[TMP1:%.*]] = bitcast bfloat [[X]] to half
+; CHECK-NEXT:    [[TMP2:%.*]] = fcmp une half [[TMP1]], 0xH0000
```
antoniofrighetto: Based on the CI failure for this test, it seems like the decimal form is needed:
```suggestion
; CHECK-NEXT:    [[TMP2:%.*]] = fcmp une half [[TMP1]], 0.000000e+00
```

antoniofrighetto (approved):
LG, thanks!

PR #196154, opened 2026-05-06
Adding the Formal Semantics Working Group to regular syncs.

alinas:
Adding the Formal Semantics Working Group to regular syncs.
The notes doc includes meeting link, meeting times and link to discord channel.

PR #198390, opened 2026-05-18
[Docs] Add Formal Semantics WG to Community Syncups List

boomanaiden154:
Per a suggestion from Aaron Ballman in the discord channel a while back.


nikic:
Didn't this already happen in https://github.com/llvm/llvm-project/pull/196154?

boomanaiden154:
Yes. It appears that the Ctrl+F that I thought I did never happened. Sorry for the noise.

PR #196418, opened 2026-05-07
[AMDGPU] Fix LowerDIVREM24 for the unsigned case

mgcarrasco:
This PR fixes the miscompilation bug found in https://github.com/llvm/llvm-project/issues/194829.

The test revealed that there were missing checks for the unsigned case in LowerDIVREM24. Without those checks, the optimization is wrongly applied assuming that the division operands do fit in 24-bit integers, triggering the miscompilation. One of the operands was 0xFFFFFFFF.

The bug is a regression introduced in https://github.com/llvm/llvm-project/commit/81a709503d378c188462465918df161d664c4205. The commit refactored the conditions that validated whether the optimization could be applied. It moved them from each call site (signed and unsigned cases) into the function, but missed the ones for the unsigned case. This PR restores them.

PR #198391, opened 2026-05-18
[AMDGPU] Fix incorrect float fast path for unsigned division at -O0

carlobertolli:
LowerDIVREM24 uses a float reciprocal to compute integer division, which is only valid when both operands fit in 24 bits of float mantissa. The function checked ComputeNumSignBits to decide this, but that check is wrong for unsigned division: a value produced by sext i16 to i32 has 17 sign bits (the top 17 bits are identical), but the compiler cannot tell whether those bits are zeros or ones. When they are ones, the unsigned value is near 2^32 -- far too large for float to represent exactly.
For example, sext(i16 0xFFFE) = i32 0xFFFFFFFE. ComputeNumSignBits returns 17, so the function thinks the value fits in 15 bits. But as unsigned 0xFFFFFFFE is 4 billion, and UINT_TO_FP rounds it to the same float as 0xFFFFFFFF, making the quotient wrong. Fix: for unsigned division, use computeKnownBits to check leading zeros instead of sign bits. Leading zeros are guaranteed to be zero regardless of the runtime value, so UINT_TO_FP is exact.

Assisted-by: Cursor (claude)


arsenm:
Duplicate #196418?

carlobertolli:
> Duplicate #196418?

It looks like that, yes. The operand 0xFFFFFFFF is indeed one that I used as an example. I will close this right now as **duplicate** of 
https://github.com/llvm/llvm-project/pull/196418
but will test once the other PR is merged to make sure the problem is actually gone.

PR #198392, opened 2026-05-18
[LLDB] Support reading size of constant values from DWARF directly

adrian-prantl:
Some type systems (such as Swift) cannot determine a type's byte size without an execution context (e.g. types whose layout depends on runtime metadata). In those cases the debug info might still carry the static size of the value's box, which is enough if the value is a constant.

This patch adds a unit test that mocks up an analogous situation using TypeSystemClang.

Assisted-by: claude


JDevlieghere (approved):
Thanks! LGTM.

adrian-prantl:
See also https://github.com/swiftlang/llvm-project/pull/12993

ceseo:
This seems to have broken the AArch64 buildbot. Could you please take a look? https://lab.llvm.org/buildbot/#/builders/59/builds/34432

PR #198393, opened 2026-05-18
[Clang] [NFC] Use `unique_ptr<Lexer>` everywhere

Sirraide:
Replace every instance of `new Lexer` with `make_unique<Lexer>` and adjust `Lexer::Create_PragmaLexer()` to return a `std::unique_ptr<Lexer>` instead.

### DIFF
--- a/clang/include/clang/Lex/Lexer.h
+++ b/clang/include/clang/Lex/Lexer.h
@@ -23,6 +23,7 @@
 #include "llvm/ADT/StringRef.h"
 #include <cassert>
 #include <cstdint>
+#include <memory>
 #include <optional>
 #include <string>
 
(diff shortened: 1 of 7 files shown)


cor3ntin (approved):

AaronBallman (commented):
Thank you for working on this! One thing I am a bit surprised by is that we are removing a comment about how the `Lexer *` needs to be `delete`d when done... but no calls to `delete` were removed. Were we always leaking those lexer objects or did we miss some calls to `delete`?

Sirraide:
> Were we always leaking those lexer objects or did we miss some calls to `delete`?

The `Preprocessor` was already storing a `unique_ptr<Lexer>`; it’s just these places where we weren’t using `unique_ptr`

AaronBallman (approved):
LGTM!

AaronBallman:
> > Were we always leaking those lexer objects or did we miss some calls to `delete`?
> 
> The `Preprocessor` was already storing a `unique_ptr<Lexer>`; it’s just these places where we weren’t using `unique_ptr`

I see now; there weren't leaks, just slightly more complicated ownership logic. :-D Thank you, this is much improved.

PR #198394, opened 2026-05-18
[GlobalISel] Add canonicalize_icmp_bool combines

AlexMaclean:


llvm/include/llvm/Target/GlobalISel/Combine.td
```diff
@@ -1264,6 +1264,47 @@ def double_icmp_zero_or_combine: GICombineRule<
          (G_ICMP $root, $p, $ordst, 0))
 >;
 
+// Mirrors InstCombine's canonicalizeICmpBool: every i1 comparison can be
+// expressed in pure and/or/xor/not.
+class canonicalize_icmp_bool_rule<string pred, dag apply_pat> : GICombineRule<
+  (defs root:$dst),
+  (match (G_ICMP i1:$dst, $cc, i1:$lhs, i1:$rhs),
+         "return ${cc}.getPredicate() == CmpInst::" # pred # ";"),
```
arsenm: We probably should have a proper way of matching the predicate kind

arsenm (commented):
Can you point me to where the DAG does this combine? That's always useful for new Gisel combines. I'd think we'd want some kind of legality check in here

llvm/test/CodeGen/AArch64/GlobalISel/combine-canonicalize-icmp-bool.mir
```diff
@@ -0,0 +1,349 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py
+# RUN: llc -o - -mtriple=aarch64-unknown-unknown -run-pass=aarch64-prelegalizer-combiner %s | FileCheck %s
```
arsenm: End to end IR tests are better, especially for optimizations

llvm/include/llvm/Target/GlobalISel/Combine.td
```diff
@@ -1264,6 +1264,47 @@ def double_icmp_zero_or_combine: GICombineRule<
          (G_ICMP $root, $p, $ordst, 0))
 >;
 
+// Mirrors InstCombine's canonicalizeICmpBool: every i1 comparison can be
```
AZero13:
```suggestion
```

Pro tip: 

// Mirrors InstCombine's canonicalizeICmpBool: every i1 comparison can be
// expressed in pure and/or/xor/not.

Is a code-smell that for some reason AI loves to mention when you do something using another as a reference. Ask me how I know.

llvm/include/llvm/Target/GlobalISel/Combine.td
AlexMaclean: I agree this would be a nice feature to extend the GlobalIsel combiner emitter with. However, using a C++ predicate is still pretty concise and clear and other combines do the same so I don't see the need to block this PR on that. What do you think?

llvm/test/CodeGen/AArch64/GlobalISel/combine-canonicalize-icmp-bool.mir
AlexMaclean: Why do you say this? I think testing the specific pass in isolation seems more targeted, like how we write lit tests for a single IR pass not the entire `opt` pipeline. It looks like about 3/4 of the tests in `AArch64/GlobalISel` are `*.mir` so I don't think I'm the only one who prefers this kind of test.

llvm/include/llvm/Target/GlobalISel/Combine.td
AZero13: Just don't mention it mirrors another thing.

llvm/test/CodeGen/AArch64/GlobalISel/combine-canonicalize-icmp-bool.mir
davemgreen: Most global-isel testing is in AArch64/*.ll tests nowadays. It makes a better test of the whole of codegen, and you can compare against SDAG at the same time. .mir tests are good for edge cases and very precise tests, but .ll is better for the majority of cases.

AlexMaclean: Fair enough, I've converted the test to an IR input.

llvm/include/llvm/Target/GlobalISel/Combine.td
AlexMaclean: Removed this comment

AlexMaclean:
> Can you point me to where the DAG does this combine? That's always useful for new Gisel combines. 

https://github.com/llvm/llvm-project/blob/0f4c5aeaad523bd0169ad94009d92570246f263e/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp#L5684-L5732

> I'd think we'd want some kind of legality check in here

I've added the necessary legality checks, the implementation in TD is a bit nasty so let me know if you have any suggestions for how to make this cleaner.

AlexMaclean:
@arsenm / @davemgreen ping

AlexMaclean:
@arsenm / @davemgreen ping

llvm/include/llvm/Target/GlobalISel/Combine.td
aemerson: If we have prior art for this I agree that this can be a separate PR.

llvm/test/CodeGen/AArch64/GlobalISel/combine-canonicalize-icmp-bool.ll
```diff
@@ -0,0 +1,263 @@
+; NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 6
+; RUN: llc -mtriple=aarch64-unknown-unknown -global-isel -verify-machineinstrs -stop-after=aarch64-prelegalizer-combiner %s -o - | FileCheck %s
```
aemerson, line 2: What they meant by IR test was to test the final output of llc, i.e. the asm. That's what we tend to do nowadays for the reasons David mentioned. You could also keep this run line if it helps to see what the combine is doing.

That said, does the existing SDAG implementation have an AArch64 test we can add a globalisel RUN line to? If so we can do that. Otherwise I'd suggest also adding an SDAG RUN here too so we can see how the two backends compare in final codegen.

arsenm, line 2: Also don't need -verify-machineinstrs

PR #198396, opened 2026-05-18
[AMDGPU] Remove some features from `gfx12-5-generic`

shiltian:
Fixes LCOMPILER-2213.

### DIFF
--- a/clang/test/CodeGenOpenCL/builtins-amdgcn-gfx12-5-generic-err.cl
+++ b/clang/test/CodeGenOpenCL/builtins-amdgcn-gfx12-5-generic-err.cl
@@ -1,0 +1,33 @@
+// REQUIRES: amdgpu-registered-target
+// RUN: %clang_cc1 -triple amdgcn-amd-amdhsa -target-cpu gfx12-5-generic -verify -S -o - %s
+
+typedef short v2s __attribute__((ext_vector_type(2)));
+typedef float v4f  __attribute__((ext_vector_type(4)));
+typedef half  v4h  __attribute__((ext_vector_type(4)));
+typedef float v8f  __attribute__((ext_vector_type(8)));
+typedef half  v8h  __attribute__((ext_vector_type(8)));
+typedef half  v16h __attribute__((ext_vector_type(16)));
+typedef __bf16 v16bf16 __attribute__((ext_vector_type(16)));
+typedef __bf16 v32bf16 __attribute__((ext_vector_type(32)));
+
+void test_amdgcn_swmmac_gfx12_5_generic(global v8f *out8f, global v4f *out4f,
+                                         global float *outf,
+                                         global unsigned int *outu,
+                                         global unsigned long *outul,
+                                         global v2s *outv2s,
+                                         v8h a8h, v16h b16h, v4h a4h,
+                                         v8h b8h, v8f c8f, v4f c4f,
+                                         v16bf16 a16bf16, v32bf16 b32bf16,
+                                         float x, float y, float z,
+                                         unsigned int u0, unsigned int u1,
+                                         unsigned int u2, unsigned long l0,
+                                         unsigned long l1, int index) {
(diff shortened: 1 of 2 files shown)


shiltian:
* #198406
* #198396 👈
* `main`

<h2></h2>

arsenm (approved):

clang/test/CodeGenOpenCL/builtins-amdgcn-gfx12-5-generic-err.cl
```diff
@@ -0,0 +1,33 @@
+// REQUIRES: amdgpu-registered-target
```
rampitec, line 1: Test name is misleading, it is not just swmmac.

llvm/lib/TargetParser/TargetParser.cpp
```diff
@@ -251,10 +251,15 @@ static void fillAMDGCNFeatureMap(StringRef GPU, const Triple &T,
   switch (Kind) {
   case GK_GFX1251:
   case GK_GFX1250:
-  case GK_GFX12_5_GENERIC:
+    Features["cube-insts"] = true;
```
rampitec, line 258: I do not understand. It falls though to the GK_GFX12_5_GENERIC and it is enabled.

clang/test/CodeGenOpenCL/builtins-amdgcn-gfx12-5-generic-err.cl
shiltian: oh, yes, I copied and pasted, but forgot the change that part.

llvm/lib/TargetParser/TargetParser.cpp
shiltian: `GK_GFX12_5_GENERIC` will only start from L262.

shiltian: However, now I realize that `GK_GFX1310` has issue, because it will miss those `*-insts` above, but not SWMMAC ones.

shiltian: Now it is correct for gfx1310 as well.

llvm/lib/TargetParser/TargetParser.cpp
```diff
@@ -251,11 +251,17 @@ static void fillAMDGCNFeatureMap(StringRef GPU, const Triple &T,
...
     Features["swmmac-gfx1200-insts"] = true;
     Features["swmmac-gfx1250-insts"] = true;
     [[fallthrough]];
   case GK_GFX1310:
+    Features["cube-insts"] = true;
+    Features["cvt-pknorm-vop2-insts"] = true;
+    Features["lerp-inst"] = true;
+    Features["qsad-insts"] = true;
+    Features["sad-insts"] = true;
+    [[fallthrough]];
```
rampitec, line 263: But it still falls through to generic.

shiltian: Hmm, I don't understand. They fall through the generic, but not the other way around.

shiltian: The generic will not have those features above it, which is what we want.

rampitec: You set `Features["cube-insts"] = true;` at line 258. Then you fall through at line 263 to generic. To me it sounds it will be enabled for generic?

rampitec: Actually, how about a codegen negative test?

shiltian: > You set `Features["cube-insts"] = true;` at line 258. Then you fall through at line 263 to generic. To me it sounds it will be enabled for generic?

For `GK_GFX12_5_GENERIC`, the switch only starts from L264.

> Actually, how about a codegen negative test?

They have already been guarded properly by those target features in `AMDGPU.td`.

rampitec: Ough, thanks. I was mislead here.

rampitec (approved):
LGTM

PR #198398, opened 2026-05-18
[flang][Semantics] Enforce IMPLICIT NONE(EXTERNAL) for dummy procedures

### LINKED ISSUE #198395: [Flang] IMPLICIT NONE(EXTERNAL) not enforced for dummy arguments used as procedures
## Description

Flang fails to diagnose a constraint violation when a dummy argument is used as a procedure under `IMPLICIT NONE(EXTERNAL)` without an explicit `EXTERNAL` attribute or interface block.

Fortran 2018, Section 8.7, C895 requires:

>If IMPLICIT NONE with an implicit-none-spec of EXTERNAL appears within a scoping unit, the name of an external or dummy procedure in that scoping unit or in a contained subprogram or BLOCK construct shall have an explicit interface or be explicitly declared to have the EXTERNAL attribute.

Flang silently accepts the code below and compiles it without error. Other compilers (e.g., Cray Fortran) correctly reject it.

## Reproducer

```fortran
subroutine sub(func_arg, n)
implicit none(external)
integer :: func_arg ! no EXTERNAL attribute
integer :: n
n = func_arg(n)
end subroutine
```

## Expected behavior

An error diagnostic should be emitted indicating that `func_arg` is used as a procedure but lacks the required explicit `EXTERNAL` attribute under `IMPLICIT NONE(EXTERNAL)`.

Example (Cray Fortran output):
```
ftn-886 ftn: ERROR SUB, File = min-repro.f90, Line = 1, Column = 16
  IMPLICIT NONE (EXTERNAL) is specified in the local scope.  An interface block
  or the EXTERNAL attribute must be specified for procedure "FUNC_ARG".
```

## Actual behavior

Flang compiles the file without any diagnostic:
```
$ flang-new -c min-repro.f90
$
```

_Issue creation assisted by Claude Opus._

kwyatt-ext:
## Summary

Fix `CheckImplicitNoneExternal()` to correctly diagnose dummy arguments used as procedures that lack an explicit `EXTERNAL` attribute when `IMPLICIT NONE(EXTERNAL)` is in effect.

Fixes #198395 

## Problem

Flang silently accepted code where a dummy argument was called as a procedure under `IMPLICIT NONE(EXTERNAL)` without the required explicit `EXTERNAL` attribute. The Fortran 2018 standard C895 requires that each dummy procedure used as a procedure shall explicitly have the `EXTERNAL` attribute when `IMPLICIT NONE(EXTERNAL)` is specified.

## Changes

**`flang/lib/Semantics/resolve-names.cpp`:**

1. In `NoteExecutablePartCall()`: Guard the `SetImplicitAttr(*symbol, Attr::EXTERNAL)` call with a check that EXTERNAL is not already explicitly set. This prevents clobbering the explicit/implicit distinction when a dummy argument already has an explicit `EXTERNAL` attribute.

2. In `CheckImplicitNoneExternal()`: Change the EXTERNAL attribute test from `!symbol.attrs().test(Attr::EXTERNAL)` to `(!symbol.attrs().test(Attr::EXTERNAL) || symbol.implicitAttrs().test(Attr::EXTERNAL))`. This catches both the case where EXTERNAL has not yet been set (non-dummy externals) and the case where it was set only implicitly (dummy arguments used as procedures).

**`flang/test/Semantics/implicit07.f90`:**

Added test case for dummy arguments under `IMPLICIT NONE(EXTERNAL)`:
- A dummy used as a procedure without `EXTERNAL` (expects error)

Assisted by Claude Opus.

### DIFF
--- a/flang/test/Semantics/implicit07.f90
+++ b/flang/test/Semantics/implicit07.f90
@@ -13,3 +13,11 @@ block
 end block
 print *, arr(1) ! no error
 end
+
+subroutine sub(func_arg, n)
+  implicit none(external)
+  integer :: func_arg
+  integer :: n
+  !ERROR: 'func_arg' is an external procedure without the EXTERNAL attribute in a scope with IMPLICIT NONE(EXTERNAL)
+  n = func_arg(n)
+end subroutine
--- a/flang/test/Semantics/implicit07.f90
+++ b/flang/test/Semantics/implicit07.f90
@@ -14,6 +14,7 @@ end block
 print *, arr(1) ! no error
 end
 
+! Function dummy argument without external
 subroutine sub(func_arg, n)
   implicit none(external)
   integer :: func_arg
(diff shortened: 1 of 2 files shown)


kwyatt-ext:
I requested a review from the last couple of people who touched this area, other than Peter.  Anyone is welcome to take it; I just didn't want to leave it hanging.

flang/test/Semantics/implicit07.f90
```diff
@@ -13,3 +13,11 @@
...
 print *, arr(1) ! no error
 end
+
+subroutine sub(func_arg, n)
+  implicit none(external)
+  integer :: func_arg
+  integer :: n
+  !ERROR: 'func_arg' is an external procedure without the EXTERNAL attribute in a scope with IMPLICIT NONE(EXTERNAL)
+  n = func_arg(n)
+end subroutine
```
eugeneepshteyn: Missing end of line

flang/test/Semantics/implicit07.f90
```diff
@@ -13,3 +13,11 @@
 end block
 print *, arr(1) ! no error
 end
+
```
eugeneepshteyn, line 16: Please add more variations: you have arg that's function, you can add subroutine form, like this:
```
subroutine sub_call(p)
  implicit none(external)
  ! external p -- error without external, no error with
  call p
end subroutine
```
Also, arg with explicit interface should not error out.

flang/test/Semantics/implicit07.f90
kwyatt-ext: Fixed.

flang/test/Semantics/implicit07.f90
kwyatt-ext: Added additional test cases and positive cases.

eugeneepshteyn (approved):
LGTM. Thank you for adding more tests!

PR #198401, opened 2026-05-18
Fix clang/test/DebugInfo/CXX/line.cpp

Ariel-Burton:
This PR twaeks the clang/test/DebugInfo/line.cpp test to pass on z/OS.

The reason the test was failing is that the RUN lines which specify -triple %itanium_abi_triple expands to s390x-ibm-zos when run on z/OS. The IR that is emitted for this triple does not match the patterns expected by the test.

This PR tweaks the patterns in the CHECK lines so that the test also passes on z/OS.

### DIFF
--- a/clang/test/DebugInfo/CXX/line.cpp
+++ b/clang/test/DebugInfo/CXX/line.cpp
@@ -8,7 +8,7 @@ int *sink();
 extern "C" __complex float complex_src();
 extern "C" __complex float *complex_sink();
 
-// CHECK-LABEL: define
+// CHECK-LABEL: define {{.*}}void {{.*}}f1
 void f1() {
   *sink()
   // CHECK: store {{.*}}, !dbg [[DBG_F1:!.*]]
@@ -24,7 +24,7 @@ struct foo {
   foo();
 };
 
-// CHECK-LABEL: define
+// CHECK-LABEL: define {{.*}}foo
 foo::foo()
     :
 #line 200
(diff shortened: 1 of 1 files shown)


Ariel-Burton:
ping ...

clang/test/DebugInfo/CXX/line.cpp
```diff
@@ -8,7 +8,7 @@ int *sink();
 extern "C" __complex float complex_src();
 extern "C" __complex float *complex_sink();
 
-// CHECK-LABEL: define
+// CHECK-LABEL: define {{.*}}void {{.*}}f1
```
Michael137, line 11: These are just for debuggability of the test right? They're not actually part of the fix?

Ariel-Burton: Yes.  The CHECK-LABELS are clearly intended to synchronize the CHECK lines and the test input; as the test is more than a few lines these help to keep track of which function a CHECK line is associated with.

Michael137 (approved):
SGTM if all the tests pass

clang/test/DebugInfo/CXX/line.cpp
```diff
@@ -71,17 +71,17 @@ void f5() {
 struct agg { int i; };
 agg agg_src();
 
-// CHECK-LABEL: define
+// CHECK-LABEL: define {{.*}}void {{.*}}f6
 void f6() {
   agg x;
-  // CHECK: call void @llvm.memcpy{{.*}} !dbg [[DBG_F6:!.*]]
+  // CHECK: call void @llvm.memcpy{{.*}}ref.tmp{{.*}} !dbg [[DBG_F6:!.*]]
```
Michael137, line 77: Bit surprised about this change. Wasn't the previous check looser than what you changed it to? Are we emitting more stuff on z/OS which would cause it to match against another memcpy, with the wrong debuginfo metadata?

Ariel-Burton: Yes.  On z/OS there are two calls to @llvm.memory,  This pattern ties it down.

PR #198404, opened 2026-05-18
[InstCombine] Drop the correct assume when working on assume bundles

philnik777:
Currently, all assumes of the same kind in an assume bundle are dropped, even though only a single one is actually checked to be redundant and should be dropped. This introduces a new `removeOperandFromBundleAt`, which instead drops a bundle at a specific position. This should also be faster, since copying the bundles can now be done into an already correctly allocated vector.

### DIFF
--- a/llvm/lib/IR/Instructions.cpp
+++ b/llvm/lib/IR/Instructions.cpp
@@ -614,7 +614,7 @@ CallBase *CallBase::removeOperandBundleAt(CallBase *CB, size_t Offset,
   auto OpBundleCount = CB->getNumOperandBundles();
   assert(Offset < OpBundleCount &&
          "Trying to remove non-existant operand bundle");
-  SmallVector<OperandBundleDef, 1> Bundles;
+  SmallVector<OperandBundleDef> Bundles;
   Bundles.reserve(OpBundleCount - 1);
   size_t I = 0;
   for (; I != Offset; ++I)
--- a/llvm/lib/IR/Instructions.cpp
+++ b/llvm/lib/IR/Instructions.cpp
@@ -609,6 +609,22 @@ CallBase *CallBase::removeOperandBundle(CallBase *CB, uint32_t ID,
   return CreateNew ? Create(CB, Bundles, InsertPt) : CB;
 }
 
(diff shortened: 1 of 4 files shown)


llvm/test/Transforms/InstCombine/assume.ll
```diff
@@ -490,6 +490,17 @@ define void @redundant_nonnull3(ptr %ptr) {
...
 }
 
+define void @partially_redundant(ptr %ptr, ptr %ptr2, ptr %ptr3) {
+; CHECK-LABEL: @partially_redundant(
+; CHECK-NEXT:    call void @llvm.assume(i1 true) [ "nonnull"(ptr [[PTR2:%.*]]) ]
+; CHECK-NEXT:    call void @llvm.assume(i1 true) [ "nonnull"(ptr [[PTR:%.*]]), "nonnull"(ptr [[PTR3:%.*]]) ]
+; CHECK-NEXT:    ret void
+;
+  call void @llvm.assume(i1 true) [ "nonnull"(ptr %ptr), "nonnull"(ptr %ptr2) ]
+  call void @llvm.assume(i1 true) [ "nonnull"(ptr %ptr), "nonnull"(ptr %ptr3) ]
```
andjo403, line 502: shall there be more tests like middle and last attribute being removed not only the first attribute?

llvm/lib/IR/Instructions.cpp
```diff
@@ -609,6 +609,22 @@ CallBase *CallBase::removeOperandBundle(CallBase *CB, uint32_t ID,
   return CreateNew ? Create(CB, Bundles, InsertPt) : CB;
 }
 
+CallBase *CallBase::removeOperandBundleAt(CallBase *CB, size_t Offset,
+                                          InsertPosition InsertPt) {
+  auto OpBundleCount = CB->getNumOperandBundles();
+  assert(Offset < OpBundleCount &&
+         "Trying to remove non-existant operand bundle");
+  SmallVector<OperandBundleDef, 1> Bundles;
```
nikic:
```suggestion
  SmallVector<OperandBundleDef> Bundles;
```

nikic (approved):
LGTM

PR #198410, opened 2026-05-18
[AMDGPU] SIFoldOperands: constant-fold S_ADD/S_SUB with immediate operands

hidekisaito:
Extend SIFoldOperands::tryConstantFoldOp to recognise three patterns
  * ADD/SUB(imm, imm) -> S_MOV_B32 (LHS +/- RHS)
  * ADD x, 0          -> COPY x   (Also `0 + x`)
  * SUB x, 0          -> COPY x   (SUB is not commutable)

Assisted-by: Claude Opus 4.7


hidekisaito:
Some more cases can be covered by adding the following in run(). Let me know if I should follow up with that.

      if (tryConstantFoldOp(&MI)) {
        Changed = true;
        continue;
      }

llvm/lib/Target/AMDGPU/SIFoldOperands.cpp
```diff
@@ -1501,6 +1501,14 @@ void SIFoldOperandsImpl::foldOperand(
 static bool evalBinaryInstruction(unsigned Opcode, int32_t &Result,
                                   uint32_t LHS, uint32_t RHS) {
   switch (Opcode) {
+  case AMDGPU::S_ADD_I32:
+  case AMDGPU::S_ADD_U32:
+    Result = LHS + RHS;
+    return true;
+  case AMDGPU::S_SUB_I32:
+  case AMDGPU::S_SUB_U32:
+    Result = LHS - RHS;
```
shiltian, line 1510: This change doesn't seem to be necessary, because, if it returns false, it will not reach L1632; if it returns true, it will also not reach L1632.

This change is not always safe either, because these four instructions set SCC as well.

hidekisaito: Need Add/Sub <imm,imm> case supported (becomes MOV) for the test we care about (else, we don't get desired inst schedule).

Also, implicit scc is not an issue. At the top of the sole caller function, there is a check for all implicit defs being dead. So, that's covered.

llvm/test/CodeGen/AMDGPU/dagcombine-reassociate-multi-memop.ll
```diff
@@ -30,7 +30,7 @@ define amdgpu_kernel void @buffer_load_lds_reassociate_offsets(ptr addrspace(1)
 ; CHECK-NEXT:    s_mov_b32 m0, 0
 ; CHECK-NEXT:    v_add_u32_e32 v0, 0x842, v0
 ; CHECK-NEXT:    buffer_load_dwordx4 v1, s[8:11], 0 offen lds
-; CHECK-NEXT:    s_add_i32 m0, 0, 0x420
+; CHECK-NEXT:    s_movk_i32 m0, 0x420
```
jayfoad, line 33: This is a shorter encoding (4 bytes instead of 8 bytes) which is nice. Is that the only advantage? Or do you have end-to-end examples where it can actually save an instruction?

hidekisaito: This helps scheduler when SGPR pressure is high. It's a cliff. Having or not having one more free register makes a big difference when scheduler needs one. Rematerializer doesn't have its own constant fold.

shiltian (approved):
LGTM

llvm/test/CodeGen/AMDGPU/dagcombine-reassociate-multi-memop.ll
jayfoad: But this example is not saving an SGPR, it is just replacing s_add_i32 with s_movk_i32. I'm just asking for a more realistic example of the codegen improvements.

PR #198411, opened 2026-05-18
[clang] Remove `FileSystemStatCache`

jansvoboda11:
This mechanism is not used anywhere, and can be easily reimplemented in terms of a VFS if needed.


clang/include/clang/Basic/FileManager.h
```diff
@@ -276,15 +259,6 @@ class FileManager : public RefCountedBase<FileManager> {
   DirectoryEntry *&getRealDirEntry(const llvm::vfs::Status &Status);
 
 public:
-  /// Get the 'stat' information for the given \p Path.
-  ///
-  /// If the path is relative, it will be resolved against the WorkingDir of the
-  /// FileManager's FileSystemOptions.
-  ///
-  /// \returns a \c std::error_code describing an error, if there was one
-  std::error_code getNoncachedStatValue(StringRef Path,
```
benlangmuir, line 314: This function isn't just not using the `FileSystemStatCache`, it also skips the `FileManager`'s own `SeenFileEntries` cache. Should we keep it for that purpose?

benlangmuir (commented):
The main change here LGTM, but changing `getNoncachedStatValue` to `getOptionalFileRef` is not equivalent.

clang/include/clang/Basic/FileManager.h
jansvoboda11: That's true. This function is only used by clangd, so I wanted to check with @sam-mccall whether that is a deliberate choice and skipping the `FileManager` cache is necessary for correctness.

jansvoboda11: CC @kadircet

clang/unittests/Basic/FileManagerTest.cpp
```diff
@@ -487,30 +374,22 @@ TEST_F(FileManagerTest, getFileRefEquality) {
...
 
   // Inject the virtual file:
-  FileEntryRef file1 = manager.getVirtualFileRef("c:\\tmp\\test", 123, 1);
-  EXPECT_EQ(43U, file1.getUniqueID().getFile());
+  FileEntryRef file1 = manager.getVirtualFileRef("/tmp/test", 123, 1);
   EXPECT_EQ(123, file1.getSize());
 
   // Lookup the virtual file with a different name:
-  auto file2 = manager.getOptionalFileRef("c:/tmp/test", 100, 1);
+  auto file2 = manager.getOptionalFileRef("/tmp/./test", 100, 1);
```
RKSimon, line 384: @jansvoboda11 MSVC is now warning that in this getOptionalFileRef call we're passing '100' to an arg that takes a bool type:
```
OptionalFileEntryRef getOptionalFileRef(StringRef Filename,
                                        bool OpenFile = false,
                                        bool CacheFailure = true,
                                        bool IsText = true)
```
Should this have been a getVirtualFileRef call?

jansvoboda11: No, `getVirtualFileRef()` will create a virtual file if one doesn't exist, while this test just wants to do a lookup. This call has been wrong since the unit test was introduced in dfffaf579f2b4bc9184c23f271a57e1fa1ed6f01: https://github.com/llvm/llvm-project/blob/dfffaf579f2b4bc9184c23f271a57e1fa1ed6f01/clang/include/clang/Basic/FileManager.h#L209-L219
I'll fix this momentarily.

PR #198412, opened 2026-05-18
[AMDGPU] Stop iDot4 chain walker at non-ADD nodes

aobolensk:
The loop body in performAddCombine dot4 matcher unconditionally treats TempNode operands as the next link's addends, so the chain only works when TempNode is an `ISD::ADD`

The old getNumOperands() guard let AND/OR/XOR/etc. through, leaking their non-addend operands (e.g. a constant mask) into the dot4 accumulator and miscompiling kernels


llvm/test/CodeGen/AMDGPU/idot4-non-add-chain.ll
```diff
@@ -0,0 +1,228 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
+; RUN: llc -mtriple=amdgcn -mcpu=gfx900  < %s | FileCheck --check-prefixes=GFX9-NODL %s
```
shiltian: What does "DL" mean here?

aobolensk: It stands for dot-product instruction subtarget features (not sure about the correct acronym details). It is used in all `idot*.ll` tests the same way

arsenm: Use subarch triples, and there's an extra space before <

llvm/test/CodeGen/AMDGPU/idot4-non-add-chain.ll
```diff
@@ -0,0 +1,228 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
```
arsenm: Old update_llc_test_checks

llvm/test/CodeGen/AMDGPU/idot4-non-add-chain.ll
aobolensk: Updated triples

llvm/test/CodeGen/AMDGPU/idot4-non-add-chain.ll
aobolensk: After updating the triples no updates for the file content were found after update_llc_test_checks run

llvm/test/CodeGen/AMDGPU/idot4-non-add-chain.ll
```diff
@@ -0,0 +1,228 @@
...
+; RUN: llc -mtriple=amdgpu9.00 < %s | FileCheck --check-prefixes=GFX9-NODL %s
+; RUN: llc -mtriple=amdgpu9.06 < %s | FileCheck --check-prefixes=GFX9-DL %s
+; RUN: llc -mtriple=amdgpu10.11 < %s | FileCheck --check-prefixes=GFX10-DL %s
+; RUN: llc -mtriple=amdgpu9.50 < %s | FileCheck --check-prefixes=GFX950 %s
+
+define i32 @dot4_and_chain_not_add(i32 %v) {
```
arsenm: Can you just add this case to the original test file for this combine

aobolensk: Moved

LU-JOHN (approved):
LGTM

PR #198414, opened 2026-05-18
[mlir][SPIR-V] Support literal struct type in spirv.Constant

aobolensk:


kuhar (commented):
Nice. Can we also add it to a test that invokes spirv-va to double check that it gets (de)serialized properly? (We have a bunch of examples already)

aobolensk:
> Nice. Can we also add it to a test that invokes spirv-va to double check that it gets (de)serialized properly? (We have a bunch of examples already)

Yes, done

mlir/lib/Target/SPIRV/Serialization/Serializer.cpp
```diff
@@ -1062,9 +1062,10 @@ uint32_t Serializer::prepareArrayConstant(Location loc, Type constType,
   uint32_t resultID = getNextID();
   SmallVector<uint32_t, 4> operands = {typeID, resultID};
   operands.reserve(attr.size() + 2);
-  auto elementType = cast<spirv::ArrayType>(constType).getElementType();
-  for (Attribute elementAttr : attr) {
-    if (auto elementID = prepareConstant(loc, elementType, elementAttr)) {
+  auto compositeType = cast<spirv::CompositeType>(constType);
+  for (auto [idx, elementAttr] : llvm::enumerate(attr)) {
+    if (auto elementID = prepareConstant(loc, compositeType.getElementType(idx),
```
IgWod: nit: I know it was like that before, but this should spell out the type.

IgWod (approved):

mlir/lib/Target/SPIRV/Serialization/Serializer.cpp
aobolensk: Done, no problem

PR #198416, opened 2026-05-18
[CFG] Assume enum switch default to be reachable

emaxx-google:
Assume that when a "default:" is present in a "switch", it's reachable for the purpose of the analysis - even if the "case"s handle all named enum values.

This allows to catch scenarios where the enum has a value that has no name assigned.

Assisted-by: Gemini

### DIFF
--- a/clang/lib/Analysis/CFG.cpp
+++ b/clang/lib/Analysis/CFG.cpp
@@ -4582,6 +4582,15 @@ CFGBlock *CFGBuilder::VisitSwitchStmt(SwitchStmt *Terminator) {
       return nullptr;
   }
 
+  bool HasDefaultCase = false;
+  for (const SwitchCase *SC = Terminator->getSwitchCaseList(); SC;
+       SC = SC->getNextSwitchCase()) {
+    if (isa<DefaultStmt>(SC)) {
+      HasDefaultCase = true;
+      break;
+    }
+  }
+
   // If we have no "default:" case, the default transition is to the code
   // following the switch body.  Moreover, take into account if all the
   // cases of a switch are covered (e.g., switching on an enum value).
@@ -4593,7 +4602,7 @@ CFGBlock *CFGBuilder::VisitSwitchStmt(SwitchStmt *Terminator) {
   bool SwitchAlwaysHasSuccessor = false;
   SwitchAlwaysHasSuccessor |= switchExclusivelyCovered;
   SwitchAlwaysHasSuccessor |=
-      !BuildOpts.AssumeReachableDefaultInSwitchStatements &&
+      !HasDefaultCase && !BuildOpts.AssumeReachableDefaultInSwitchStatements &&
       Terminator->isAllEnumCasesCovered() && Terminator->getSwitchCaseList();
   addSuccessor(SwitchTerminatedBlock, DefaultCaseBlock,
                !SwitchAlwaysHasSuccessor);
@@ -5609,12 +5618,12 @@ bool CFGBlock::FilterEdge(const CFGBlock::FilterOptions &F,
 
   if (To && From && F.IgnoreDefaultsWithCoveredEnums) {
     // If the 'To' has no label or is labeled but the label isn't a
-    // CaseStmt then filter this edge.
+    // CaseStmt or DefaultStmt then filter this edge.
     if (const SwitchStmt *S =
         dyn_cast_or_null<SwitchStmt>(From->getTerminatorStmt())) {
       if (S->isAllEnumCasesCovered()) {
         const Stmt *L = To->getLabel();
-        if (!L || !isa<CaseStmt>(L))
+        if (!L || (!isa<CaseStmt>(L) && !isa<DefaultStmt>(L)))
           return true;
       }
     }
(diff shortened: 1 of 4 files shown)


clang/lib/Analysis/CFG.cpp
```diff
@@ -4582,6 +4582,15 @@ CFGBlock *CFGBuilder::VisitSwitchStmt(SwitchStmt *Terminator) {
       return nullptr;
   }
 
+  bool HasDefaultCase = false;
+  for (const SwitchCase *SC = Terminator->getSwitchCaseList(); SC;
+       SC = SC->getNextSwitchCase()) {
+    if (isa<DefaultStmt>(SC)) {
+      HasDefaultCase = true;
```
emaxx-google, line 4589: Here just a presence of the default is checked, not whether it's reachable (i.e., not whether cases handle all integers from the underlying type).

We can try implementing this in the future if there's a need. It's probably a relatively rare edge case (I took Chromium as a sample and haven't found any).

clang/lib/Analysis/CFG.cpp
```diff
@@ -5609,12 +5618,12 @@ bool CFGBlock::FilterEdge(const CFGBlock::FilterOptions &F,
...
   if (To && From && F.IgnoreDefaultsWithCoveredEnums) {
     // If the 'To' has no label or is labeled but the label isn't a
-    // CaseStmt then filter this edge.
+    // CaseStmt or DefaultStmt then filter this edge.
     if (const SwitchStmt *S =
         dyn_cast_or_null<SwitchStmt>(From->getTerminatorStmt())) {
       if (S->isAllEnumCasesCovered()) {
         const Stmt *L = To->getLabel();
-        if (!L || !isa<CaseStmt>(L))
+        if (!L || (!isa<CaseStmt>(L) && !isa<DefaultStmt>(L)))
```
Sirraide, line 5626:
```suggestion
        if (!L || !isa<SwitchCase>(L))
```
We have a common base class for these that you can check for here

Sirraide (commented):
I think the implementation of this is fine, but the main issue I can think of that this approach has is that if you have code like this
```c++
enum E { A, B };
int f(E e) {
  switch (e) {
    case A: return 1;
    case B: return 2;
    default: std::unreachable();
  }
} 
```
and then add another enumerator `C`, now `-Wswitch` won’t fire because there is a `default` case (`-Wswitch-enum` would, but that one is too strict imo because sometimes you _do_ want to have a default case that isn’t just a fallthrough/unreachable case).

So because of that I’m not sure tying this sort of check to `default` is a good idea...

Sirraide:
> then add another enumerator `C`, now `-Wswitch` won’t fire because there is a `default` case

To elaborate, the behaviour that I at least would like to have is
1. When I add an enumerator, I want the compiler to warn on every `switch` statement that switches on that enum so I know that I have to update it.
2. At the same time, there should be some way to opt out of this because I don’t want to have to add like 20 instances of `case Foo: std::unreachable();`.
3. Lastly, I want the compiler to assume that we can fall off the end of the switch statement so that I’m forced to put a `std::unreachable()` or something like it there to make sure we don’t fall off the end of the function.

The status quo is that w/ `-Wswitch` (which is enabled by default), `default` labels do 1 and 2, but you want 3. But with the approach in this pr, you get 2 and 3, but not 1 (and w/ `-Wswitch-enum` you get 1 only at the moment and 1 and 3 with this pr). Ideally, I’d like to find some approach that provides 1, 2, _and_ 3.

emaxx-google:
> > then add another enumerator `C`, now `-Wswitch` won’t fire because there is a `default` case
> 
> To elaborate, the behaviour that I at least would like to have is
> 
> 1. When I add an enumerator, I want the compiler to warn on every `switch` statement that switches on that enum so I know that I have to update it.
> 2. At the same time, there should be some way to opt out of this because I don’t want to have to add like 20 instances of `case Foo: std::unreachable();`.
> 3. Lastly, I want the compiler to assume that we can fall off the end of the switch statement so that I’m forced to put a `std::unreachable()` or something like it there to make sure we don’t fall off the end of the function.
> 
> The status quo is that w/ `-Wswitch` (which is enabled by default), `default` labels do 1 and 2, but you want 3. But with the approach in this pr, you get 2 and 3, but not 1 (and w/ `-Wswitch-enum` you get 1 only at the moment and 1 and 3 with this pr). Ideally, I’d like to find some approach that provides 1, 2, _and_ 3.

Good point, and we came to the same conclusion after checking this with other Chromium V8 maintainers (the project where this diagnostic is very desired from the security standpoint). I.e., asking developers to add `default` everywhere will lead to other kinds of issues.

emaxx-google:
Having said that, would this PR still be reasonable on its own? This makes the CFG recognize that the `default:` branch is reachable even when all named enumerators have `case`s.

Sirraide:
> Having said that, would this PR still be reasonable on its own? This makes the CFG recognize that the `default:` branch is reachable even when all named enumerators have `case`s.

I personally would be fine with it, but iirc other people have tried to make similar changes to switch statements before, and it has been contentious in the past; that said, I may be misremembering. 

CC @AaronBallman, @cor3ntin, @erichkeane, @shafik for opinions

efriedma-quic:
> Ideally, I’d like to find some approach that provides 1, 2, and 3.

I don't see how you can get all three without new syntax.  If we have "default" cases that are supposed to include enum constants, and "default" cases that aren't supposed to include enum constants, there's no way to distinguish them.

If we're willing to take new syntax, there's a variety of approaches.

PR #198418, opened 2026-05-18
Don't warn for unknown `@Command`s in backticks in doc comments

harlanhaskins:
`-Wdocumentation-unknown-command` currently ignores doxygen commands within single quotes to match doxygen behavior. Since doxygen supports Markdown, I tried putting commands in backticks, and it ignores those too.

The impetus for this was a spurious warning in a generated `-Swift.h` header from someone referencing "an `@MainActor` function". Such a use should be totally permitted.


harlanhaskins:
Ah, hm, I see I’ve changed expected behavior in clangd; I’ll take a look

harlanhaskins:
I'm gonna close this and re-open it; I think given the new impedance mismatch between clangd's preprocessing and the lever behavior, I'd prefer a more targeted fix to silence this warning.

PR #198419, opened 2026-05-18
[AMDGPU] Fix BitOp3_Op duplicate Src entry in replace-parent path

wooseoklee:
In BitOp3_Op, the "replace parent operator" branch fires when a source slot already holds the node being expanded (Src[I] == In). It then substitutes the child (Op) into that slot. If Op is already present in another slot, the substitution creates a duplicate entry: two slots hold the same value but carry different truth-table column bits (e.g. 0xf0 and 0xcc), corrupting all subsequent bit assignments and producing a wrong truth table.

Fix: before replacing Src[I], check that Op is not already present in any other slot. If it is, bail out of the replace-parent path so the caller can allocate a fresh slot instead.

The same bug exists in both the SelectionDAG (AMDGPUISelDAGToDAG.cpp) and GlobalISel (AMDGPUInstructionSelector.cpp) implementations; both are fixed identically.

Patterns affected (all of the form op(f(x), x) where x is shared):
- or(x, C) ^ x
- and(xor(or(a,b), b), or(a,b))
- xor(or(xor(fshl(x),x),x), xor(fshl(x),x))
- xor(and(xor(a,b), b), xor(a,b))

Add lit tests for each pattern in bitop3.ll covering gfx950 and gfx1250 under both SDAG and GlobalISel code paths.

### DIFF
--- a/llvm/lib/Target/AMDGPU/AMDGPUISelDAGToDAG.cpp
+++ b/llvm/lib/Target/AMDGPU/AMDGPUISelDAGToDAG.cpp
@@ -4358,9 +4358,10 @@ static std::pair<unsigned, uint8_t> BitOp3_Op(SDValue In,
         // create a duplicate entry. Two slots with the same value would be
         // assigned different truth-table column bits (e.g. 0xf0 and 0xcc),
         // corrupting all downstream bit assignments.
-        for (unsigned J = 0; J < Src.size(); ++J)
+        for (unsigned J = 0; J < Src.size(); ++J) {
           if (J != I && Src[J] == Op)
             return false;
+        }
         Bits = SrcBits[I];
         Src[I] = Op;
         return true;
--- a/llvm/lib/Target/AMDGPU/AMDGPUISelDAGToDAG.cpp
+++ b/llvm/lib/Target/AMDGPU/AMDGPUISelDAGToDAG.cpp
@@ -4354,6 +4354,13 @@ static std::pair<unsigned, uint8_t> BitOp3_Op(SDValue In,
       }
       // Try to replace parent operator
       if (Src[I] == In) {
+        // If Op is already present at another slot, substituting here would
+        // create a duplicate entry. Two slots with the same value would be
+        // assigned different truth-table column bits (e.g. 0xf0 and 0xcc),
+        // corrupting all downstream bit assignments.
+        for (unsigned J = 0; J < Src.size(); ++J)
+          if (J != I && Src[J] == Op)
+            return false;
         Bits = SrcBits[I];
         Src[I] = Op;
         return true;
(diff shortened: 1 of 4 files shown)


llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
```diff
@@ -4188,6 +4188,13 @@ static std::pair<unsigned, uint8_t> BitOp3_Op(Register R,
       }
       // Try to replace parent operator
       if (Src[I] == R) {
+        // If Op is already present at another slot, substituting here would
+        // create a duplicate entry. Two slots with the same value would be
+        // assigned different truth-table column bits (e.g. 0xf0 and 0xcc),
+        // corrupting all downstream bit assignments.
+        for (unsigned J = 0; J < Src.size(); ++J)
+          if (J != I && Src[J] == Op)
+            return false;
```
shiltian: https://llvm.org/docs/AMDGPU/DeveloperGuideline.html#use-of-braces
```suggestion
        for (unsigned J = 0; J < Src.size(); ++J) {
          if (J != I && Src[J] == Op)
            return false;
        }
```

llvm/lib/Target/AMDGPU/AMDGPUISelDAGToDAG.cpp
```diff
@@ -4354,6 +4354,13 @@ static std::pair<unsigned, uint8_t> BitOp3_Op(SDValue In,
       }
       // Try to replace parent operator
       if (Src[I] == In) {
+        // If Op is already present at another slot, substituting here would
+        // create a duplicate entry. Two slots with the same value would be
+        // assigned different truth-table column bits (e.g. 0xf0 and 0xcc),
+        // corrupting all downstream bit assignments.
+        for (unsigned J = 0; J < Src.size(); ++J)
```
shiltian: same here

shiltian:
https://github.com/llvm/llvm-project/pull/198373 and https://github.com/llvm/llvm-project/pull/198419, they seem to do the same thing.

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
wooseoklee: Thanks. I missed it. Fixed.

wooseoklee:
> #198373 and #198419, they seem to do the same thing.

It is addressing different issues reported from fuzzer. I double checked it by locally applying the patch from https://github.com/llvm/llvm-project/pull/198373.

wooseoklee:
Gentle ping to get any review.

wooseoklee:
Closing this as it is unable to reproduce the issue with latest llvm.

PR #149614, opened 2025-07-18
[flang][acc] Lower do and do concurrent loops specially in acc regions

razvanlupusoru:
When OpenACC is enabled and Fortran loops are annotated with `acc loop`, they are lowered to `acc.loop` operation. And rest of the contained loops use the normal FIR lowering path.

Hovever, the OpenACC specification has special provisions related to contained loops and their induction variable. In order to adhere to this, we convert all valid contained loops to `acc.loop` in order to store this information appropriately.

The provisions in the spec that motivated this change (line numbers are from OpenACC 3.4):
- 1353 Loop variables in Fortran do statements within a compute construct are predetermined to be private to the thread that executes the loop.
- 3783 When do concurrent appears without a loop construct in a kernels construct it is treated as if it is annotated with loop auto. If it appears in a parallel construct or an accelerator routine then it is treated as if it is annotated with loop independent.

By valid loops - we convert do loops and do concurrent loops which have induction variable. Loops which are unstructured are not handled.

PR #198420, opened 2026-05-18
[flang][acc] Handle Fortran do loops as acc loops in acc routine

razvanlupusoru:
As was previously done for do loops in acc compute constructs in https://github.com/llvm/llvm-project/issues/149614 , this PR does the same for do loops in `acc routine`. The rules are follows:
- Do loops not marked with `acc loop` are considered `auto`
- Do concurrent loops are considered `independent`
- Any loops in an `acc routine seq` are considered `seq`

This ensures that the IV is correctly privatized and attached to acc loop.


vzakhari (approved):
Looks great! Thank you, Razvan!

PR #198421, opened 2026-05-18
[mlir][memref] Use access interfaces in address extraction

krzysz00:
Rework extract-address-computation patterns to use IndexedAccessOpInterface for direct memref accesses and VectorTransferOpInterface update hooks for transfer ops.

These rewrites are limited to operations that declare in-bounds indices (so no vector.load currently) for now so that we always create valid `memref.subview` opp.

As a consequence of this PR, the Memref dialect no longer depends on the GPU and NVGPU dialects.

AI: Codex wrote the first draft, I simplified it a bunch and made sure the names of internal functions made sense.


krzysz00:
Ping

mlir/test/Dialect/MemRef/extract-address-computations.mlir
```diff
@@ -301,6 +412,7 @@ module attributes {transform.with_named_sequence} {
   }
 }
 
+
```
lialan: Nit: double blank line here (414-415) before the `// -----`. Cosmetic, and clang-format won't catch it.

lialan (approved):
+1

Only concern is still the external model silently dropping if one does not register it. So worth a shout out to relevant people.

PR #195257, opened 2026-05-01
[CIR] Implement setGlobalVisibility

ZakyHermawan:
Matching OGCG

PR #198422, opened 2026-05-18
[CIR] Set ExternalWeakLinkage on weak/weak_import function declarations

adams381:
Classic CodeGen's `SetFunctionAttributes` calls `setLinkageForGV` to force `ExternalWeakLinkage` on `__attribute__((weak))` and Darwin `weak_import` declarations.  CIR had no equivalent: weak function declarations were emitted with `ExternalLinkage` instead of `ExternalWeakLinkage`.

This adds `setLinkageForFunction` — the same weak/external-weak logic as `setLinkageForGV` — and calls it from `setFunctionAttributes`.  The underlying crash on inline forward declarations (the original motivation) is already fixed by [#195257](https://github.com/llvm/llvm-project/pull/195257); what remains is this linkage gap.

`inline-forward-decl.c` covers `__attribute__((weak))` on an inline forward declaration; `func-linkage-weak-import.c` covers Darwin `weak_import` (→ `extern_weak` in CIR and LLVM).

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenModule.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenModule.cpp
@@ -976,12 +976,21 @@ static void setLinkageForGV(cir::GlobalOp &gv, const NamedDecl *nd) {
     gv.setLinkage(cir::GlobalLinkageKind::ExternalWeakLinkage);
 }
 
-static void setLinkageForFunction(cir::FuncOp &func, const NamedDecl *nd) {
+static void setLinkageForFunction(CIRGenModule &cgm, cir::FuncOp &func,
+                                  const NamedDecl *nd) {
   // Mirrors CodeGenModule::setLinkageForGV for function declarations.
   LinkageInfo lv = nd->getLinkageAndVisibility();
   if (isExternallyVisible(lv.getLinkage()) &&
-      (nd->hasAttr<WeakAttr>() || nd->isWeakImported()))
-    func.setLinkage(cir::GlobalLinkageKind::ExternalWeakLinkage);
+      (nd->hasAttr<WeakAttr>() || nd->isWeakImported())) {
+    auto linkage = cir::GlobalLinkageKind::ExternalWeakLinkage;
+    func.setLinkage(linkage);
+    func.setLinkageAttr(
+        cir::GlobalLinkageKindAttr::get(&cgm.getMLIRContext(), linkage));
+    // Declarations must keep 'private' MLIR visibility; only update for defs.
+    if (!func.isDeclaration())
+      mlir::SymbolTable::setSymbolVisibility(
+          func, cgm.getMLIRVisibilityFromCIRLinkage(linkage));
+  }
 }
 
 static llvm::SmallVector<int64_t> indexesOfArrayAttr(mlir::ArrayAttr indexes) {
@@ -3002,24 +3011,7 @@ void CIRGenModule::setFunctionAttributes(GlobalDecl globalDecl,
     getTargetCIRGenInfo().setTargetAttributes(funcDecl, func, *this);
 
   // Mirrors setLinkageForGV in CodeGenModule::SetFunctionAttributes.
-  setLinkageForFunction(func, funcDecl);
-
-  // GetGVALinkageForFunction requires a definition (or willHaveBody) and
-  // asserts via isInlineDefinitionExternallyVisible on bare declarations.
-  const bool isDeclWithoutBody =
-      func.isDeclaration() && !funcDecl->doesThisDeclarationHaveABody();
-  if (!isIncompleteFunction && !isDeclWithoutBody) {
-    // TODO(cir): This needs a lot of work to better match CodeGen. That
-    // ultimately ends up in setGlobalVisibility, which already has the linkage
-    // of the LLVM GV (corresponding to our FuncOp) computed, so it doesn't have
-    // to recompute it here. This is a minimal fix for now.
-    if (!isLocalLinkage(getFunctionLinkage(globalDecl))) {
-      const Decl *decl = globalDecl.getDecl();
-      func.setGlobalVisibility(
-          getGlobalVisibilityAttrFromDecl(decl).getValue());
-    }
-  }
-
+  setLinkageForFunction(*this, func, funcDecl);
 
   // If we plan on emitting this inline builtin, we can't treat it as a builtin.
   if (funcDecl->isInlineBuiltinDeclaration()) {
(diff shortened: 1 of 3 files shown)


clang/lib/CIR/CodeGen/CIRGenModule.cpp
```diff
@@ -909,6 +909,14 @@ static void setLinkageForGV(cir::GlobalOp &gv, const NamedDecl *nd) {
     gv.setLinkage(cir::GlobalLinkageKind::ExternalWeakLinkage);
 }
 
+static void setLinkageForFunction(cir::FuncOp &func, const NamedDecl *nd) {
+  // Mirrors CodeGenModule::setLinkageForGV for function declarations.
+  LinkageInfo lv = nd->getLinkageAndVisibility();
+  if (isExternallyVisible(lv.getLinkage()) &&
+      (nd->hasAttr<WeakAttr>() || nd->isWeakImported()))
```
erichkeane: Can we get a test for both of these versions?  Else, NYI for them?

clang/test/CIR/CodeGen/inline-forward-decl.c
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// CIR-DAG: cir.func{{.*}}@helper
+// CIR-DAG: cir.func{{.*}}@user
+// CIR: cir.call @helper
+
+// LLVM: define{{.*}}@user
+// LLVM: call{{.*}}@helper
+
+// OGCG: define{{.*}}@user
```
erichkeane: Why separate OGCG again?  Can you instruct your LLM to stop doing this?

clang/lib/CIR/CodeGen/CIRGenModule.cpp
```diff
@@ -2873,13 +2881,23 @@ void CIRGenModule::setFunctionAttributes(GlobalDecl globalDecl,
...
-  if (!isLocalLinkage(getFunctionLinkage(globalDecl))) {
-    const Decl *decl = globalDecl.getDecl();
-    func.setGlobalVisibility(getGlobalVisibilityAttrFromDecl(decl).getValue());
+  // Mirrors setLinkageForGV in CodeGenModule::SetFunctionAttributes.
+  setLinkageForFunction(func, funcDecl);
+
+  // GetGVALinkageForFunction requires a definition (or willHaveBody) and
+  // asserts via isInlineDefinitionExternallyVisible on bare declarations.
+  const bool isDeclWithoutBody =
+      func.isDeclaration() && !funcDecl->doesThisDeclarationHaveABody();
```
erichkeane: Really this should only have to be `funcDecl->isDefined()`, right?  Why the extra bit about `has a body`?  That actually isn't really true in cases where we can actually generate it (such as defaulted I think).

clang/lib/CIR/CodeGen/CIRGenModule.cpp
adams381: Added weak_fn in inline-forward-decl.c and func-linkage-weak-import.c for Darwin weak_import (extern_weak in CIR/LLVM).

clang/test/CIR/CodeGen/inline-forward-decl.c
adams381: Dropped the duplicate OGCG block — one LLVM: check set, both %t-cir.ll and %t.ll use --check-prefix=LLVM.

clang/lib/CIR/CodeGen/CIRGenModule.cpp
adams381: Switched to funcDecl->isDefined() only.

andykaylor:
Does https://github.com/llvm/llvm-project/pull/195257 solve the same problem you're trying to fix here?

clang/lib/CIR/CodeGen/CIRGenModule.cpp
adams381: Both versions are covered: `inline-forward-decl.c` tests `__attribute__((weak))` (a definition → `weak` linkage in CIR/LLVM), and the new `func-linkage-weak-import.c` tests `__attribute__((weak_import))` on Darwin (a declaration → `extern_weak` in CIR/LLVM).

clang/test/CIR/CodeGen/inline-forward-decl.c
adams381: Done — the separate OGCG RUN/CHECK block is gone. Both `%t-cir.ll` (CIR-lowered) and `%t.ll` (OGCG) are now checked with the shared `LLVM:` prefix.

clang/lib/CIR/CodeGen/CIRGenModule.cpp
adams381: The switch to bare `isDefined()` caused a regression: `isDefined()` scans all redeclarations, so a forward declaration of an inline function whose definition follows later would pass the guard and hand the forward-decl's `GlobalDecl` to `getFunctionLinkage` → `GetGVALinkageForFunction` → `isInlineDefinitionExternallyVisible` assertion. Fixed by using `funcDecl->getDefinition()` and `globalDecl.getWithDecl(def)`, which gives `GetGVALinkageForFunction` the body-bearing decl it requires while skipping the block entirely for bare declarations. Also fixed `setLinkageForFunction` to set MLIR symbol visibility alongside the linkage attr (missing was causing `extern_weak` not to appear in CIR output for `weak_import` declarations).

adams381:
[#195257](https://github.com/llvm/llvm-project/pull/195257) adds `setGlobalVisibility` (the default/hidden/protected visibility logic). The crash here is on a separate path: `setFunctionAttributes` → `getFunctionLinkage` → `GetGVALinkageForFunction` → `isInlineDefinitionExternallyVisible`, which asserts `doesThisDeclarationHaveABody()` on a forward decl. The two don't overlap.

adams381:
After looking at this more carefully: #195257 does overlap. It removed the `getFunctionLinkage` call from `setFunctionAttributes` that was causing the crash, so the original crash fix is already in upstream.

What remains independently needed is the `setLinkageForFunction` addition — classic CodeGen's `SetFunctionAttributes` calls `setLinkageForGV` to force `ExternalWeakLinkage` on `__attribute__((weak))` and `weak_import` declarations, and CIR had no equivalent for functions. The visibility block I re-added (with the same TODO you removed in #195257) is redundant since `setGVProperties` → `setGlobalVisibility` already covers that path. I'll drop it and update the description to reflect the narrower scope.

adams381:
@erichkeane — `inline-forward-decl.c` covers `weak`; `func-linkage-weak-import.c` covers Darwin `weak_import`. OGCG block is gone (shared `LLVM` prefix). Guard is `getDefinition()` + `globalDecl.getWithDecl(def)`, not bare `isDefined()`. Scope narrowed after #195257 — this PR is only `setLinkageForFunction`.

erichkeane (approved):

PR #198423, opened 2026-05-18
[CIR] Represent union tail padding in RecordType padding field

adams381:
Union tail padding was stored as an extra entry in `members`, so
`getLargestMember()` had to treat the last element specially and the IR
looked like a three-field union.  Andy asked for tail pad in a separate
`padding = { … }` slot on `!cir.record<union …>` instead.

This adds optional `padding` on `RecordType` (parse/print, verify, storage),
wires CIRGen layout to set it for padded unions, and threads it through
CXXABILowering and LowerToLLVM.  Codegen tests and a unit test are updated.

### DIFF
--- a/clang/include/clang/CIR/Dialect/IR/CIRTypes.td
+++ b/clang/include/clang/CIR/Dialect/IR/CIRTypes.td
@@ -665,6 +665,13 @@ def CIR_RecordType : CIR_Type<"Record", "record", [
             padding = {!cir.array<!u8i x 8>}>
     ```
 
+    Tail padding is represented differently by record kind.  For structs and
+    classes, `padded` still appends the tail-padding byte array as the last
+    entry in `members` (see `!rec_p2` / `!rec_p3` above).  For unions,
+    `padded` sets the flag and the tail-padding type lives in the optional
+    `padding = { … }` slot so `members` lists only the real variants (see
+    `!rec_u`).
+
     Incomplete records are mutable, meaning they can be later completed with a
     body automatically updating in place every type in the code that uses the
     incomplete record. Mutability allows for recursive types to be represented,
--- a/clang/include/clang/CIR/Dialect/IR/CIRTypes.td
+++ b/clang/include/clang/CIR/Dialect/IR/CIRTypes.td
@@ -661,6 +661,8 @@ def CIR_RecordType : CIR_Type<"Record", "record", [
         !rec_p1 = !cir.record<struct "p1" packed {!u8i, !u8i}>
         !rec_p2 = !cir.record<struct "p2" padded {!u8i, !u8i}>
         !rec_p3 = !cir.record<struct "p3" packed padded {!s32i, !u8i, !u8i}>
+        !rec_u = !cir.record<union "u" padded {!s8i, !u64i},
+            padding = {!cir.array<!u8i x 8>}>
     ```
 
     Incomplete records are mutable, meaning they can be later completed with a
@@ -682,7 +684,8 @@ def CIR_RecordType : CIR_Type<"Record", "record", [
     "bool":$incomplete,
     "bool":$packed,
     "bool":$padded,
-    "RecordType::RecordKind":$kind
+    "RecordType::RecordKind":$kind,
+    OptionalParameter<"mlir::Type">:$padding
   );
 
   // StorageClass is defined in C++ for mutability.
@@ -699,10 +702,11 @@ def CIR_RecordType : CIR_Type<"Record", "record", [
       "mlir::StringAttr":$name,
       "bool":$packed,
       "bool":$padded,
-      "RecordKind":$kind
+      "RecordKind":$kind,
+      CArg<"mlir::Type", "{}">:$padding
     ), [{
       return $_get($_ctxt, members, name, /*incomplete=*/false, packed, padded,
-                   kind);
+                   kind, padding);
     }]>,
 
     // Create an identified and incomplete record type.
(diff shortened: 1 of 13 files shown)


clang/lib/CIR/Dialect/IR/CIRTypes.cpp
```diff
@@ -273,9 +294,15 @@ mlir::LogicalResult
...
                    llvm::ArrayRef<mlir::Type> members, mlir::StringAttr name,
                    bool incomplete, bool packed, bool padded,
-                   RecordType::RecordKind kind) {
+                   RecordType::RecordKind kind, mlir::Type padding) {
   if (name && name.getValue().empty())
     return emitError() << "identified records cannot have an empty name";
+  if (padding) {
+    if (kind != RecordKind::Union)
+      return emitError() << "record padding is only supported on union types";
+    if (!padded)
```
andykaylor, line 307: Can you also add a check that when a union is padded it has the extra padding attribute?

clang/include/clang/CIR/Dialect/IR/CIRTypes.td
```diff
@@ -682,7 +684,8 @@ def CIR_RecordType : CIR_Type<"Record", "record", [
     "bool":$incomplete,
     "bool":$packed,
     "bool":$padded,
-    "RecordType::RecordKind":$kind
+    "RecordType::RecordKind":$kind,
+    OptionalParameter<"mlir::Type">:$padding
```
andykaylor, line 695: This is still less than ideal, because when a non-union record is padded the padding is included among the elements. This requires some explanation in the description.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -3313,16 +3313,12 @@ static void prepareTypeConverter(mlir::LLVMTypeConverter &converter,
       break;
     // Unions are lowered as only the largest member.
     case cir::RecordType::Union:
-      if (type.getMembers().empty())
```
andykaylor, line 3316: I'm not sure this can happen, but if it does shouldn't we still break?

andykaylor (commented):
It looks like this includes the fix in https://github.com/llvm/llvm-project/pull/198361. I think it makes sense to land this change first without that commit and then rebase that one and update it as needed.

clang/lib/CIR/Dialect/IR/CIRTypes.cpp
adams381: Added verify that padded unions must carry padding = { … }, plus invalid-record-padding.cir.  RecordType::parse now returns on getChecked failure instead of casting a null type.

clang/include/clang/CIR/Dialect/IR/CIRTypes.td
adams381: Documented in CIRTypes.td: struct/class tail padding stays as the last members entry; unions use the separate padding = { … } slot.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Guarded getLargestMember on non-empty members so empty padded unions (padded {}, padding = {…}) still emit the padding field — a bare break on empty members would drop that.

andykaylor (approved):
lgtm

clang/include/clang/CIR/Dialect/IR/CIRTypes.td
```diff
@@ -661,8 +661,17 @@ def CIR_RecordType : CIR_Type<"Record", "record", [
         !rec_p1 = !cir.record<struct "p1" packed {!u8i, !u8i}>
         !rec_p2 = !cir.record<struct "p2" padded {!u8i, !u8i}>
         !rec_p3 = !cir.record<struct "p3" packed padded {!s32i, !u8i, !u8i}>
+        !rec_u = !cir.record<union "u" padded {!s8i, !u64i},
+            padding = {!cir.array<!u8i x 8>}>
     ```
 
+    Tail padding is represented differently by record kind.  For structs and
+    classes, `padded` still appends the tail-padding byte array as the last
+    entry in `members` (see `!rec_p2` / `!rec_p3` above).  For unions,
```
erichkeane, line 670: I don't know how I feel about these being different.  I realize Unions and structs are DIFFERENT enough already, but having this much special casing in the type seems wrong.

It DOES make me think that perhaps we've diverged enough that the two should have their own types though, rather than the `RecordType` differentiator.  We should perhaps consider a `StructType` and a `UnionType`.

clang/lib/CIR/CodeGen/CIRGenBuilder.cpp
```diff
@@ -194,7 +194,8 @@ cir::RecordType clang::CIRGen::CIRGenBuilderTy::getCompleteRecordType(
   if (name.empty())
     return getAnonRecordTy(members, packed, padded);
 
-  return getCompleteNamedRecordType(members, packed, padded, name);
+  return getCompleteNamedRecordType(members, packed, padded, mlir::Type{},
```
erichkeane, line 197: This feels problematic/difficult to read why we are just passing a plain `type` there for seemingly no reason.

Rather than this taking the padding TYPE, should it take the padding size and generate it? Alternatively, should this take some sort of `PaddingInfo` object, such that th is would take: `PaddingInfo::None` object?

erichkeane (changes requested):
I said before I liked the idea, but the more I look at this, the more I think that bifurcating the `RecordType` here is way out of hand.  We need to step back and figure out what this looks like first.

adams381:
Closing in favor of #199790, which takes a more complete approach by moving union tail padding to a dedicated storage field and introducing StructType/UnionType view classes.

PR #198424, opened 2026-05-18
[mlir][LLVMIR] Allow address-of-global as a leaf in array constants

adams381:
Large `llvm.mlir.global` initializers built as nested `llvm.insertvalue`
chains make `LLVMModuleTranslation::convertGlobalsAndAliases` call
`ConstantFoldInsertValueInstruction` on every step, rebuilding the
whole `ConstantArray` each time. That is O(N²) in the number of
elements and shows up as multi-minute compiles on translation units with
huge pointer tables (SPEC CPU 2026 `gcc/insn-automata.cc` is the
motivating case; Eric Keane's `convertOperationImpl` profile matches this
path).

This change lets `llvm.mlir.constant` carry an `ArrayAttr` of
`FlatSymbolRefAttr` leaves that name globals (not just functions), adds
a name-keyed global map beside the existing op-keyed map, and resolves
those refs in `getLLVMConstant`.  A translate test checks the resulting
single LLVM constant array initializer.

### DIFF
--- a/mlir/include/mlir/Target/LLVMIR/ModuleTranslation.h
+++ b/mlir/include/mlir/Target/LLVMIR/ModuleTranslation.h
@@ -262,6 +262,11 @@ public:
     return globalsMapping.lookup(op);
   }
 
+  /// Finds an LLVM IR global value by the mlir.global symbol name.
+  llvm::GlobalValue *lookupGlobal(StringRef name) const {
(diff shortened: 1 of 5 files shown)


krzysz00 (approved):
Seems pretty reasonable to me

andykaylor (approved):
lgtm

PR #198426, opened 2026-05-18
[BPF] Emit ABI-accurate BTF prototypes for DW_CC_nocall

yonghong-song:
DW_CC_nocall subprograms can end up with an optimized IR signature that no
longer matches the original source-level DISubroutineType. Dead argument
elimination may drop source parameters, and the return value may be
removed entirely, while the debug type still describes the original
prototype. In that case BTFDebug emits a FUNC_PROTO that no longer matches
the real BPF ABI.

Teach BTFDebug to derive a filtered FUNC_PROTO for nocall functions.

Detecting surviving arguments (collectNocallEntryArgRegs):

Scan all DBG_VALUE instructions in the entry block while tracking which
registers have been redefined by non-debug instructions:

- A DBG_VALUE whose register has not been redefined records a
  register-passed argument (R1-R5 at function entry).
- A DBG_VALUE whose register was most recently loaded via LDD $r11,
  offset records a stack-passed argument (beyond the first five).
  LowerFormalArguments always places stack argument loads in the entry
  block, so scanning MF.front() is sufficient.
- Indirect DBG_VALUEs (describing memory locations, e.g. spills to
  [$r10 + offset]) are skipped -- the register is a base address,
  not the argument value.
- DBG_VALUEs with immediate operands (constant-propagated args) are
  skipped because MO.isReg() returns false.

Emitting a filtered prototype:

A filtered prototype (true ABI signature) is emitted when all of the
following hold:
  - The number of surviving source arguments equals the number of
    optimized IR arguments (no merging or splitting).
  - Every surviving source argument matches the corresponding IR type
    (pointer, same-width integer/boolean/float, or same-width enum).
  - For up to the first std::size(CC_BPF64_ArgRegs) surviving
    arguments, entry-block DBG_VALUEs assign BPF registers in order
    R1..R5.

The original source prototype is used as fallback when:
  - The surviving source arg count differs from the IR arg count.
  - Any surviving source type does not match its IR type (e.g. structs).
  - The register assignment does not follow R1..R5 order.

Independently, if the IR return type is void but the debug info declares a
non-void return, the emitted BTF return type becomes void. This applies
regardless of whether the parameter list is filtered.

When using the filtered prototype, decl_tag component indices are remapped
to the filtered parameter list.

Add coverage for dead register and stack args, all-dead args, unchanged
signatures, const/bool/mixed types, type mismatches, decl_tags, poisoned
DBG_VALUE inputs, and voided returns.

### DIFF
--- a/llvm/lib/Target/BPF/BTFDebug.cpp
+++ b/llvm/lib/Target/BPF/BTFDebug.cpp
@@ -119,7 +119,7 @@ static bool sourceArgMatchesIRType(const DIType *SourceTy, Type *IRTy) {
 /// registers to stack frame objects during the entry-block walk (using
 /// MachineMemOperands to identify the target frame index), then match them
 /// against VariableDbgInfo entries after the scan.
-static SmallDenseMap<uint32_t, Register>
+static SmallVector<std::pair<uint32_t, Register>, 8>
 collectNocallEntryArgRegs(const MachineFunction &MF) {
   SmallDenseMap<uint32_t, Register> EntryRegMap;
   const DISubprogram *SP = MF.getFunction().getSubprogram();
@@ -203,7 +203,10 @@ collectNocallEntryArgRegs(const MachineFunction &MF) {
       EntryRegMap[Arg] = It->second;
   }
 
-  return EntryRegMap;
+  SmallVector<std::pair<uint32_t, Register>, 8> AliveArgs(EntryRegMap.begin(),
+                                                          EntryRegMap.end());
+  llvm::sort(AliveArgs, llvm::less_first());
+  return AliveArgs;
 }
 
 /// Check whether the optimized IR signature matches the surviving source
@@ -212,19 +215,18 @@ collectNocallEntryArgRegs(const MachineFunction &MF) {
 /// BPF register order (R1..R5) for register args.
 static bool canUseNocallOptimizedSignature(
     const MachineFunction &MF, DITypeArray Elements,
-    ArrayRef<uint32_t> SortedAlive,
-    const SmallDenseMap<uint32_t, Register> &EntryRegMap,
+    ArrayRef<std::pair<uint32_t, Register>> AliveArgs,
     const TargetRegisterInfo &TRI) {
-  if (MF.getFunction().arg_size() != SortedAlive.size()) {
+  if (MF.getFunction().arg_size() != AliveArgs.size()) {
     LLVM_DEBUG(dbgs() << "BTF skip " << MF.getName() << ": IR arg count ("
                       << MF.getFunction().arg_size() << ") != alive arg count ("
-                      << SortedAlive.size() << ")\n");
+                      << AliveArgs.size() << ")\n");
     return false;
   }
 
   auto ArgIt = MF.getFunction().arg_begin();
-  for (unsigned I = 0, N = SortedAlive.size(); I < N; ++I, ++ArgIt) {
-    uint32_t ArgNo = SortedAlive[I];
+  for (unsigned I = 0, N = AliveArgs.size(); I < N; ++I, ++ArgIt) {
+    auto [ArgNo, Reg] = AliveArgs[I];
     if (!sourceArgMatchesIRType(Elements[ArgNo], ArgIt->getType())) {
       LLVM_DEBUG(dbgs() << "BTF skip " << MF.getName()
                         << ": type mismatch for source arg " << ArgNo
@@ -235,10 +237,7 @@ static bool canUseNocallOptimizedSignature(
     if (I >= std::size(CC_BPF64_ArgRegs))
       continue;
 
-    auto It = EntryRegMap.find(ArgNo);
-    assert(It != EntryRegMap.end() &&
-           "register arg must have entry-block DBG_VALUE");
-    int DwarfReg = TRI.getDwarfRegNum(It->second, false);
+    int DwarfReg = TRI.getDwarfRegNum(Reg, false);
     if (DwarfReg != static_cast<int>(I + 1)) {
       LLVM_DEBUG(dbgs() << "BTF skip " << MF.getName() << ": arg " << ArgNo
                         << " in DWARF reg " << DwarfReg << ", expected "
@@ -1590,30 +1589,28 @@ void BTFDebug::beginFunctionImpl(const MachineFunction *MF) {
     const TargetRegisterInfo *TRI = MF->getSubtarget().getRegisterInfo();
     DITypeArray Elements = SP->getType()->getTypeArray();
 
-    SmallDenseMap<uint32_t, Register> EntryRegMap =
+    SmallVector<std::pair<uint32_t, Register>, 8> AliveArgs =
         collectNocallEntryArgRegs(*MF);
 
-    SmallVector<uint32_t, 8> SortedAlive;
-    for (const auto &[ArgNo, Reg] : EntryRegMap)
-      SortedAlive.push_back(ArgNo);
-    llvm::sort(SortedAlive);
-
-    UseFilteredParams = canUseNocallOptimizedSignature(
-        *MF, Elements, SortedAlive, EntryRegMap, *TRI);
+    UseFilteredParams =
+        canUseNocallOptimizedSignature(*MF, Elements, AliveArgs, *TRI);
 
     if (UseFilteredParams) {
+      SmallVector<uint32_t, 8> AliveParamIndices;
       SmallDenseMap<uint32_t, uint32_t> ArgIndexMap;
-      for (uint32_t I = 0; I < SortedAlive.size(); ++I)
-        ArgIndexMap[SortedAlive[I]] = I;
+      for (auto [I, ArgReg] : llvm::enumerate(AliveArgs)) {
+        AliveParamIndices.push_back(ArgReg.first);
+        ArgIndexMap[ArgReg.first] = I;
+      }
 
       if (!VoidReturn)
         visitTypeEntry(Elements[0]);
-      for (uint32_t ArgNo : SortedAlive)
+      for (uint32_t ArgNo : AliveParamIndices)
         visitTypeEntry(Elements[ArgNo]);
 
       auto TypeEntry = std::make_unique<BTFTypeFuncProto>(
-          SP->getType(), SortedAlive.size(), FuncArgNames, true, SortedAlive,
-          VoidReturn);
+          SP->getType(), AliveParamIndices.size(), FuncArgNames, true,
+          AliveParamIndices, VoidReturn);
       ProtoTypeId = addType(std::move(TypeEntry));
       FuncTypeId = processDISubprogram(SP, ProtoTypeId, Scope, &ArgIndexMap);
     }
--- a/llvm/lib/Target/BPF/BTFDebug.cpp
+++ b/llvm/lib/Target/BPF/BTFDebug.cpp
@@ -19,11 +19,14 @@
 #include "llvm/CodeGen/AsmPrinter.h"
 #include "llvm/CodeGen/MachineModuleInfo.h"
 #include "llvm/CodeGen/MachineOperand.h"
+#include "llvm/CodeGen/TargetRegisterInfo.h"
+#include "llvm/CodeGen/TargetSubtargetInfo.h"
 #include "llvm/IR/Module.h"
 #include "llvm/MC/MCContext.h"
 #include "llvm/MC/MCObjectFileInfo.h"
 #include "llvm/MC/MCSectionELF.h"
 #include "llvm/MC/MCStreamer.h"
+#include "llvm/Support/Debug.h"
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/IOSandbox.h"
 #include "llvm/Support/LineIterator.h"
@@ -32,6 +35,11 @@
 #include <optional>
 
 using namespace llvm;
+
+#define DEBUG_TYPE "btf-debug"
+
+#define GET_CC_REGISTER_LISTS
+#include "BPFGenCallingConv.inc"
 
 static const char *BTFKindStr[] = {
 #define HANDLE_BTF_KIND(ID, NAME) "BTF_KIND_" #NAME,
@@ -45,6 +53,145 @@ static const DIType *tryRemoveAtomicType(const DIType *Ty) {
   if (DerivedTy && DerivedTy->getTag() == dwarf::DW_TAG_atomic_type)
     return DerivedTy->getBaseType();
   return Ty;
+}
+
+static const DIType *stripDITypeAttributes(const DIType *Ty) {
+  while (const auto *DTy = dyn_cast_or_null<DIDerivedType>(Ty)) {
+    switch (DTy->getTag()) {
+    case dwarf::DW_TAG_atomic_type:
+    case dwarf::DW_TAG_const_type:
+    case dwarf::DW_TAG_restrict_type:
+    case dwarf::DW_TAG_typedef:
+    case dwarf::DW_TAG_volatile_type:
+      Ty = DTy->getBaseType();
+      break;
+    default:
+      return Ty;
+    }
+  }
+  return Ty;
+}
+
+static bool sourceArgMatchesIRType(const DIType *SourceTy, Type *IRTy) {
+  SourceTy = stripDITypeAttributes(SourceTy);
+
+  // All pointers are opaque in LLVM IR, so any source-level pointer matches any
+  // IR pointer regardless of pointee type.
+  if (const auto *DTy = dyn_cast<DIDerivedType>(SourceTy))
+    return DTy->getTag() == dwarf::DW_TAG_pointer_type && IRTy->isPointerTy();
+
+  if (const auto *BTy = dyn_cast<DIBasicType>(SourceTy)) {
+    uint64_t SizeInBits = BTy->getSizeInBits();
+    if (BTy->getEncoding() == dwarf::DW_ATE_float)
+      return IRTy->isFloatingPointTy() &&
+             IRTy->getPrimitiveSizeInBits() == SizeInBits;
+    // _Bool is 8 bits in DWARF/source but lowered to i1 in LLVM IR.
+    if (BTy->getEncoding() == dwarf::DW_ATE_boolean && IRTy->isIntegerTy(1))
+      return true;
+    return IRTy->isIntegerTy(SizeInBits);
+  }
+
+  const auto *CTy = dyn_cast<DICompositeType>(SourceTy);
+  if (!CTy)
+    return false;
+
+  switch (CTy->getTag()) {
+  case dwarf::DW_TAG_enumeration_type:
+    return IRTy->isIntegerTy(CTy->getSizeInBits());
+  default:
+    return false;
+  }
+}
+
+/// Collect the physical register each source argument lives in by scanning
+/// DBG_VALUE instructions in the entry block.  A DBG_VALUE is only recorded
+/// when its register either (a) has not been redefined by any preceding
+/// non-debug instruction (i.e. it still holds the caller-passed value), or
+/// (b) was most recently loaded from the stack via $r11 (a stack-passed
+/// argument beyond the first five register args).
+static SmallDenseMap<uint32_t, Register>
+collectNocallEntryArgRegs(const MachineFunction &MF) {
+  SmallDenseMap<uint32_t, Register> EntryRegMap;
+  const DISubprogram *SP = MF.getFunction().getSubprogram();
+  SmallDenseSet<Register> DefinedRegs, StackLoadRegs;
+
+  for (const MachineInstr &MI : MF.front()) {
+    if (MI.isDebugValue()) {
+      // Skip indirect DBG_VALUEs — the register is a base address for a
+      // memory location, not the argument value itself.
+      if (MI.isIndirectDebugValue())
+        continue;
+
+      const DILocalVariable *DV = MI.getDebugVariable();
+      if (!DV || !DV->getArg() || DV->getScope()->getSubprogram() != SP)
+        continue;
+
+      uint32_t Arg = DV->getArg();
+      const MachineOperand &MO = MI.getDebugOperand(0);
+      if (!MO.isReg() || !MO.getReg().isPhysical())
+        continue;
+
+      if (!DefinedRegs.contains(MO.getReg()) ||
+          StackLoadRegs.contains(MO.getReg()))
+        EntryRegMap[Arg] = MO.getReg();
+      continue;
+    }
+
+    for (const MachineOperand &MO : MI.operands())
+      if (MO.isReg() && MO.isDef() && MO.getReg().isPhysical()) {
+        DefinedRegs.insert(MO.getReg());
+        StackLoadRegs.erase(MO.getReg());
+      }
+
+    // Detect stack argument loads: $rX = LDD $r11, offset.
+    if (MI.getOpcode() == BPF::LDD && MI.getOperand(1).getReg() == BPF::R11)
+      StackLoadRegs.insert(MI.getOperand(0).getReg());
+  }
+  return EntryRegMap;
+}
+
+/// Check whether the optimized IR signature matches the surviving source
+/// arguments precisely enough to emit a filtered BTF prototype.
+/// Requires exact IR/source arg count match, matching types, and correct
+/// BPF register order (R1..R5) for register args.
+static bool canUseNocallOptimizedSignature(
+    const MachineFunction &MF, DITypeArray Elements,
+    ArrayRef<uint32_t> SortedAlive,
+    const SmallDenseMap<uint32_t, Register> &EntryRegMap,
+    const TargetRegisterInfo &TRI) {
+  if (MF.getFunction().arg_size() != SortedAlive.size()) {
+    LLVM_DEBUG(dbgs() << "BTF skip " << MF.getName() << ": IR arg count ("
+                      << MF.getFunction().arg_size() << ") != alive arg count ("
+                      << SortedAlive.size() << ")\n");
+    return false;
+  }
+
+  auto ArgIt = MF.getFunction().arg_begin();
+  for (unsigned I = 0, N = SortedAlive.size(); I < N; ++I, ++ArgIt) {
+    uint32_t ArgNo = SortedAlive[I];
+    if (!sourceArgMatchesIRType(Elements[ArgNo], ArgIt->getType())) {
+      LLVM_DEBUG(dbgs() << "BTF skip " << MF.getName()
+                        << ": type mismatch for source arg " << ArgNo
+                        << " at IR position " << I << "\n");
+      return false;
+    }
+
+    if (I >= std::size(CC_BPF64_ArgRegs))
+      continue;
+
+    auto It = EntryRegMap.find(ArgNo);
+    assert(It != EntryRegMap.end() &&
+           "register arg must have entry-block DBG_VALUE");
+    int DwarfReg = TRI.getDwarfRegNum(It->second, false);
+    if (DwarfReg != static_cast<int>(I + 1)) {
+      LLVM_DEBUG(dbgs() << "BTF skip " << MF.getName() << ": arg " << ArgNo
+                        << " in DWARF reg " << DwarfReg << ", expected "
+                        << (I + 1) << "\n");
+      return false;
+    }
+  }
+
+  return true;
 }
 
 /// Emit a BTF common type.
@@ -398,8 +545,12 @@ std::string BTFTypeStruct::getName() { return std::string(STy->getName()); }
 /// for subprogram.
 BTFTypeFuncProto::BTFTypeFuncProto(
     const DISubroutineType *STy, uint32_t VLen,
-    const std::unordered_map<uint32_t, StringRef> &FuncArgNames)
-    : STy(STy), FuncArgNames(FuncArgNames) {
+    const SmallDenseMap<uint32_t, StringRef> &FuncArgNames,
+    bool UseFilteredParams, ArrayRef<uint32_t> AliveParamIndices,
+    bool ChangedToVoid)
+    : STy(STy), FuncArgNames(FuncArgNames),
+      AliveParamIndices(AliveParamIndices),
+      UseFilteredParams(UseFilteredParams), ChangedToVoid(ChangedToVoid) {
   Kind = BTF::BTF_KIND_FUNC_PROTO;
   BTFType.Info = (Kind << 24) | VLen;
 }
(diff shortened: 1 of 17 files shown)


yonghong-song:
I tried to run bpf selftests. There are two kinds of failures.

First kind of failure:
```
libbpf: prog 'invalidate_pkt_pointers_by_static_tail_call': BPF program load failed: -EINVAL
libbpf: prog 'invalidate_pkt_pointers_by_static_tail_call': failed to load: -EINVAL
libbpf: failed to load object 'verifier_sock'
run_subtest:PASS:unexpected_load_success 0 nsec
validate_msgs:FAIL:974 expect_msg
VERIFIER LOG:
=============
tail_call is only allowed in functions that return 'int'.
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
=============
EXPECTED   SUBSTR: 'invalid mem access'
#631/59  verifier_sock/invalidate_pkt_pointers_by_static_tail_call:FAIL
...
tester_init:PASS:tester_log_buf 0 nsec
process_subtest:PASS:obj_open_mem 0 nsec
process_subtest:PASS:specs_alloc 0 nsec
libbpf: prog 'classifier_0': BPF program load failed: -EINVAL
libbpf: prog 'classifier_0': -- BEGIN PROG LOAD LOG --
LD_ABS is only allowed in functions that return 'int'.
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
-- END PROG LOAD LOG --
libbpf: prog 'classifier_0': failed to load: -EINVAL
libbpf: failed to load object 'tailcall_bpf2bpf2.bpf.o'
test_tailcall_bpf2bpf_2:FAIL:692
#468/8   tailcalls/tailcall_bpf2bpf_2:FAIL
...
libbpf: prog 'entry': BPF program load failed: -EINVAL
libbpf: prog 'entry': -- BEGIN PROG LOAD LOG --
tail_call is only allowed in functions that return 'int'.
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
-- END PROG LOAD LOG --
libbpf: prog 'entry': failed to load: -EINVAL
libbpf: failed to load object 'tailcall_bpf2bpf_hierarchy1.bpf.o'
test_tailcall_hierarchy_count:FAIL:load obj unexpected error: -22 (errno 22)
#468/18  tailcalls/tailcall_bpf2bpf_hierarchy_1:FAIL
...
```
For example, for test 468/18,
```
static __noinline
int subprog_tail(struct __sk_buff *skb)
{
        bpf_tail_call_static(skb, &jmp_table, 0);
        return 0;
}

SEC("tc")
int entry(struct __sk_buff *skb)
{
        int ret = 1;

        clobber_regs_stack();

        count++;
        subprog_tail(skb);
        subprog_tail(skb);

        return ret;
}
```
The return value for `subprog_tail()` is not used so compiler optimize it away. The same for LD_ABS case.

Second kind of failure:
```
libbpf: prog 'test_subprog2': BPF program load failed: -EACCES
libbpf: prog 'test_subprog2': -- BEGIN PROG LOAD LOG --
0: R1=ctx() R10=fp0
; int test_subprog2(struct args_subprog2 *ctx) @ fexit_bpf2bpf.c:66
0: (bf) r6 = r1                       ; R1=ctx() R6=ctx()
1: (b7) r1 = 112                      ; R1=112
; struct sk_buff *skb = (void *)ctx->args[0]; @ fexit_bpf2bpf.c:68
2: (79) r3 = *(u64 *)(r6 +0)          ; R3=ptr_sk_buff() R6=ctx()
3: (0f) r3 += r1                      ; R1=112 R3=ptr_sk_buff(imm=112)
4: (bf) r1 = r10                      ; R1=fp0 R10=fp0
5: (07) r1 += -4                      ; R1=fp-4
; bpf_probe_read_kernel(&len, sizeof(len), @ fexit_bpf2bpf.c:72
6: (b4) w2 = 4                        ; R2=4
7: (85) call bpf_probe_read_kernel#113        ; fp-8=mmmmpppp
; if (len != 74 || ret != 148) @ fexit_bpf2bpf.c:82
8: (61) r1 = *(u32 *)(r10 -4)         ; R1=scalar(smin=0,smax=umax=0xffffffff,var_off=(0x0; 0xffffffff)) R10=fp0
9: (56) if w1 != 0x4a goto pc+8       ; R1=74
10: (79) r1 = *(u64 *)(r6 +40)
func 'test_pkt_access_subprog2' doesn't have 3-th argument
invalid bpf_context access off=40 size=8
processed 11 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
-- END PROG LOAD LOG --
libbpf: prog 'test_subprog2': failed to load: -EACCES
libbpf: failed to load object './fexit_bpf2bpf.bpf.o'
test_fexit_bpf2bpf_common:FAIL:obj_load unexpected error: -13 (errno 13)
#121/3   fexit_bpf2bpf/func_replace:FAIL
```
The function test_subprog2():
```
/* Though test_pkt_access_subprog2() is defined in C as:
 * static __attribute__ ((noinline))
 * int test_pkt_access_subprog2(int val, volatile struct __sk_buff *skb)
 * {
 *     return skb->len * val;
 * }
 * llvm optimizations remove 'int val' argument and generate BPF assembly:
 *   r0 = *(u32 *)(r1 + 0)
 *   w0 <<= 1
 *   exit
 * In such case the verifier falls back to conservative and
 * tracing program can access arguments and return value as u64
 * instead of accurate types.
 */
struct args_subprog2 {
        __u64 args[5];
        __u64 ret;
};
__u64 test_result_subprog2 = 0;
SEC("fexit/test_pkt_access_subprog2")
int test_subprog2(struct args_subprog2 *ctx)
{
        struct sk_buff *skb = (void *)ctx->args[0];
        __u64 ret;
        int len;

        bpf_probe_read_kernel(&len, sizeof(len),
                              __builtin_preserve_access_index(&skb->len));

        ret = ctx->ret;
        /* bpf_prog_test_load() loads "test_pkt_access.bpf.o" with
         * BPF_F_TEST_RND_HI32 which randomizes upper 32 bits after BPF_ALU32
         * insns. Hence after 'w0 <<= 1' upper bits of $rax are random. That is
         * expected and correct. Trim them.
         */
        ret = (__u32) ret;
        if (len != 74 || ret != 148)
                return 0;
        test_result_subprog2 = 1;
        return 0;
}
```
As described in the above, for function `test_pkt_access_subprog2`, and `int val` is optimized way. Since BTF does not match actual parameter, the prog is set as unreliable, so in btf_ctx_access(), prog->aux->attach_func_proto is NULL, so we have 5 arguments `nr_args = t ? btf_type_vlen(t) : MAX_BPF_FUNC_REG_ARGS` as seen in the above.

But with true signature, it only has one argument in BTF and this matches actual parameter in btf_check_func_arg_match(), one argument + one return value. This does not match above args_subprog2().

The following are the way to fix the issue:

For kind 1, something like
```
diff --git a/kernel/bpf/check_btf.c b/kernel/bpf/check_btf.c
index 93bebe6fe12e..42f7069c4a0e 100644
--- a/kernel/bpf/check_btf.c
+++ b/kernel/bpf/check_btf.c
@@ -185,13 +185,13 @@ static int check_btf_func(struct bpf_verifier_env *env,
 
                ret_type = btf_type_skip_modifiers(btf, func_proto->type, NULL);
                scalar_return =
-                       btf_type_is_small_int(ret_type) || btf_is_any_enum(ret_type);
+                       btf_type_is_small_int(ret_type) || btf_is_any_enum(ret_type) || ret_type->size == 0;
                if (i && !scalar_return && env->subprog_info[i].has_ld_abs) {
-                       verbose(env, "LD_ABS is only allowed in functions that return 'int'.\n");
+                       verbose(env, "LD_ABS is only allowed in functions that return 'int' or 'void'.\n");
                        goto err_free;
                }
                if (i && !scalar_return && env->subprog_info[i].has_tail_call) {
-                       verbose(env, "tail_call is only allowed in functions that return 'int'.\n");
+                       verbose(env, "tail_call is only allowed in functions that return 'int' or 'void'.\n");
                        goto err_free;
                }
```

For kind 2,
```
diff --git a/tools/testing/selftests/bpf/progs/fexit_bpf2bpf.c b/tools/testing/selftests/bpf/progs/fexit_bpf2bpf.c
index 983b7c233382..fbdfb7c35b9e 100644
--- a/tools/testing/selftests/bpf/progs/fexit_bpf2bpf.c
+++ b/tools/testing/selftests/bpf/progs/fexit_bpf2bpf.c
@@ -53,12 +53,11 @@ int BPF_PROG(test_subprog1, struct sk_buff *skb, int ret)
  *   r0 = *(u32 *)(r1 + 0)
  *   w0 <<= 1
  *   exit
- * In such case the verifier falls back to conservative and
- * tracing program can access arguments and return value as u64
- * instead of accurate types.
+ * In such case the true signature will be provided as below with
+ * one argument and ont return value.
  */
 struct args_subprog2 {
-       __u64 args[5];
+       __u64 args[1];
        __u64 ret;
 };
 __u64 test_result_subprog2 = 0;
```

Also, do we want to have a feature macro for this? I didn't do it since the coverage of true signature is not complete. Any suggestions?

yonghong-song:
Also, the true signature resolved the veristat-meta failures like below:
```
|File                                                                   |Program        |Verdict                |States Diff (%)|
|-----------------------------------------------------------------------|---------------|-----------------------|---------------|
|bpfjailer-lib-tests-bpf-dbus_test-dbus_test.bpf.o                      |do_parse       |success -> failure (!!)|-99.16 %       |
|bpfjailer-lib-tests-bpf-dbus_test-dbus_test.bpf.o                      |do_sendmsg     |success -> failure (!!)|-99.18 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf-Twfw.bpf.o                  |tw_twfw_egress |success -> failure (!!)|-99.27 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf-Twfw.bpf.o                  |tw_twfw_ingress|success -> failure (!!)|-98.52 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf-Twfw.bpf.o                  |tw_twfw_tc_eg  |success -> failure (!!)|-99.26 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf-Twfw.bpf.o                  |tw_twfw_tc_in  |success -> failure (!!)|-98.52 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf_loop_bpf-Twfw_bpf_loop.bpf.o|tw_twfw_egress |success -> failure (!!)|-80.70 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf_loop_bpf-Twfw_bpf_loop.bpf.o|tw_twfw_ingress|success -> failure (!!)|-80.70 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf_loop_bpf-Twfw_bpf_loop.bpf.o|tw_twfw_tc_eg  |success -> failure (!!)|-80.50 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf_loop_bpf-Twfw_bpf_loop.bpf.o|tw_twfw_tc_in  |success -> failure (!!)|-80.70 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf_rs_genskel-twfw_bpf_rs.bpf.o|tw_twfw_egress |success -> failure (!!)|-99.27 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf_rs_genskel-twfw_bpf_rs.bpf.o|tw_twfw_ingress|success -> failure (!!)|-98.52 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf_rs_genskel-twfw_bpf_rs.bpf.o|tw_twfw_tc_eg  |success -> failure (!!)|-99.26 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_bpf_rs_genskel-twfw_bpf_rs.bpf.o|tw_twfw_tc_in  |success -> failure (!!)|-98.52 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_global_bpf-Twfw_global.bpf.o    |tw_twfw_egress |success -> failure (!!)|-80.70 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_global_bpf-Twfw_global.bpf.o    |tw_twfw_ingress|success -> failure (!!)|-80.70 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_global_bpf-Twfw_global.bpf.o    |tw_twfw_tc_eg  |success -> failure (!!)|-80.50 %       |
|tupperware-agent-net-twfw-skb-bpf-twfw_global_bpf-Twfw_global.bpf.o    |tw_twfw_tc_in  |success -> failure (!!)|-80.70 %       |
```
In the above failures, they are caused by 6 argument functions and the compiler optimized them away to 5 arguments.
With this pull request, verification is successful with latest kernel stack argument support.

eddyz87:
> The original source prototype is used as fallback when:
> ...
> The surviving source arg count differs from the IR arg count.
> Any surviving source type does not match its IR type (e.g. structs).
> The register assignment does not follow R1..R5 order.

I understand that this is how things worked so far, but since this PR attempts to improve the signatures precision, wouldn't it make sense to drop the signatures to something non-specific like `void(void)` in the cases above as well?

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -33,6 +35,8 @@
 
 using namespace llvm;
 
+static constexpr unsigned BPFMaxRegArgs = 5;
```
eddyz87: Nit: this can be derived as `std::size(CC_BPF64_ArgRegs)` from `BPFGenCallingConv.inc`.

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -47,6 +51,149 @@ static const DIType *tryRemoveAtomicType(const DIType *Ty) {
...
+    return false;
+  }
+}
+
+/// Collect the physical register each source argument lives in by scanning the
+/// leading DBG_VALUE instructions in the entry block (stops at the first
+/// non-debug instruction).
+static DenseMap<uint32_t, Register>
+collectNocallEntryArgRegs(const MachineFunction &MF) {
+  DenseMap<uint32_t, Register> EntryRegMap;
```
eddyz87: Nit: there is a `SmallDenseMap` to avoid heap allocations.

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -47,6 +51,149 @@ static const DIType *tryRemoveAtomicType(const DIType *Ty) {
...
+  return EntryRegMap;
+}
+
+/// Collect source argument numbers still alive after optimization (in a
+/// register or stack slot). Constant-propagated args with only immediate
+/// DBG_VALUEs are excluded.
+static DenseSet<uint32_t>
+collectNocallAliveArgs(const MachineFunction &MF,
+                       const DenseMap<uint32_t, Register> &EntryRegMap) {
+  DenseSet<uint32_t> AliveArgs;
```
eddyz87: Nit: same for `SmallDenseSet`.

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -47,6 +51,149 @@ static const DIType *tryRemoveAtomicType(const DIType *Ty) {
...
+  const DISubprogram *SP = MF.getFunction().getSubprogram();
+
+  // Seed with args already found in entry-block registers.
+  for (const auto &[ArgNo, Reg] : EntryRegMap)
+    AliveArgs.insert(ArgNo);
+
+  // Pick up args not in the entry-block set (e.g. stack-passed args) that
+  // still have a register location somewhere.  An arg with only immediate
+  // DBG_VALUEs was constant-propagated and is not considered alive.
+  for (const MachineBasicBlock &MBB : MF) {
```
eddyz87: This seem to be an exact copy of `collectNocallEntryArgRegs`, but iterating over all basic blocks. Why having two functions?

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -47,6 +51,149 @@ static const DIType *tryRemoveAtomicType(const DIType *Ty) {
...
+
+      const DILocalVariable *DV = MI.getDebugVariable();
+      if (!DV || !DV->getArg() || DV->getScope()->getSubprogram() != SP)
+        continue;
+
+      uint32_t Arg = DV->getArg();
+      if (AliveArgs.contains(Arg))
+        continue;
+
+      const MachineOperand &MO = MI.getDebugOperand(0);
```
eddyz87: I don't think there is guarantee that this argument is an actual incoming function parameter, not a result of some intermediate computation, why filtering it with `isReg()`?

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -47,6 +51,149 @@ static const DIType *tryRemoveAtomicType(const DIType *Ty) {
...
+        AliveArgs.insert(Arg);
+    }
+  }
+  return AliveArgs;
+}
+
+/// Check whether the optimized IR signature matches the surviving source
+/// arguments precisely enough to emit a filtered nocall BTF prototype.
+/// Requires at least one dead arg, exact IR/source arg count match, matching
+/// types, and correct BPF register order (R1..R5) for register args.
```
eddyz87: Can we always derive the prototype for the `nocall` functions from `DBG_VALUE` instead of using DWARF-encoded signature? Why "Requires at least one dead arg"?

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -398,8 +545,12 @@ std::string BTFTypeStruct::getName() { return std::string(STy->getName()); }
 /// for subprogram.
 BTFTypeFuncProto::BTFTypeFuncProto(
     const DISubroutineType *STy, uint32_t VLen,
-    const std::unordered_map<uint32_t, StringRef> &FuncArgNames)
-    : STy(STy), FuncArgNames(FuncArgNames) {
+    const std::unordered_map<uint32_t, StringRef> &FuncArgNames,
+    bool UseFilteredParams, std::vector<uint32_t> AliveParamIndices,
```
eddyz87: `std::vector<uint32_t> AliveParamIndices`, should this be an `ArrayRef`?
`const std::unordered_map<uint32_t, StringRef> &FuncArgNames` a `SmallDenseMap` or a `SmallVector`?

llvm/lib/Target/BPF/BTFDebug.h
```diff
@@ -143,11 +143,17 @@ class BTFTypeStruct : public BTFTypeBase {
 class BTFTypeFuncProto : public BTFTypeBase {
   const DISubroutineType *STy;
   std::unordered_map<uint32_t, StringRef> FuncArgNames;
+  std::vector<uint32_t> AliveParamIndices;
```
eddyz87: `SmallVector`?

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -668,8 +837,10 @@ void BTFDebug::processDeclAnnotations(DINodeArray Annotations,
   }
 }
 
-uint32_t BTFDebug::processDISubprogram(const DISubprogram *SP,
-                                       uint32_t ProtoTypeId, uint8_t Scope) {
+uint32_t
+BTFDebug::processDISubprogram(const DISubprogram *SP, uint32_t ProtoTypeId,
+                              uint8_t Scope,
+                              const DenseMap<uint32_t, uint32_t> *ArgIndexMap) {
```
eddyz87: Nit: `SmallVector`

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -1344,12 +1522,57 @@ void BTFDebug::beginFunctionImpl(const MachineFunction *MF) {
...
+
+  if (IsNocall) {
+    // For DW_CC_nocall functions, try to build a FUNC_PROTO reflecting
+    // the true ABI: only parameters that survived optimization and whose
+    // first 5 arguments map to the correct BPF registers (R1-R5).
+    const TargetRegisterInfo *TRI = MF->getSubtarget().getRegisterInfo();
+    DITypeArray Elements = SP->getType()->getTypeArray();
+
+    // Check if DeadArgumentElimination changed the return type to void.
+    IsReturnVoided =
```
eddyz87: Nit: `ChangedToVoid` would be a better name, but why not limit it to `MF->getFunction().getReturnType()->isVoidTy()` and drop the `&& Elements[0] != nullptr` part?

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -1344,12 +1522,57 @@ void BTFDebug::beginFunctionImpl(const MachineFunction *MF) {
...
+
+    std::vector<uint32_t> SortedAlive(AliveArgs.begin(), AliveArgs.end());
+    llvm::sort(SortedAlive);
+
+    UseFilteredParams = canUseNocallOptimizedSignature(
+        *MF, Elements, SortedAlive, EntryRegMap, *TRI);
+
+    if (UseFilteredParams) {
+      DenseMap<uint32_t, uint32_t> ArgIndexMap;
+      for (uint32_t I = 0; I < SortedAlive.size(); ++I)
```
eddyz87: I'm not sure this logic is correct, given that `SortedAlive` does not guarantee that arguments are passed directly as arguments.

eddyz87 (changes requested):

yonghong-song:
> > The original source prototype is used as fallback when:
> > ...
> > The surviving source arg count differs from the IR arg count.
> > Any surviving source type does not match its IR type (e.g. structs).
> > The register assignment does not follow R1..R5 order.
> 
> I understand that this is how things worked so far, but since this PR attempts to improve the signatures precision, wouldn't it make sense to drop the signatures to something non-specific like `void(void)` in the cases above as well?

For such cases, I would like to preserve the original signatures. One of reason is due to llvm which may miss some DBG_VALUE, e.g.

```
# Machine code for function trace_var_sys_kill: NoPHIs, TracksLiveness, NoVRegs, TiedOpsRewritten, TracksDebugUserValues
Frame Objects:
  fi#0: size=4, align=4, at location [SP-4]
  fi#1: size=4, align=4, at location [SP-8]
save/restore points:
save points are empty
restore points are empty
Function Live Ins: $w1, $w2

bb.0.entry:
  successors: %bb.23(0x30000000), %bb.1(0x50000000); %bb.23(37.50%), %bb.1(62.50%)
  liveins: $w1, $w2
  DBG_VALUE $w2, $noreg, !"sig", !DIExpression(), debug-location !34299; progs/profiler.inc.h:0 line no:375
  $w7 = MOV_rr_32 $w2
  DBG_VALUE $w7, $noreg, !"sig", !DIExpression(), debug-location !34299; progs/profiler.inc.h:0 line no:375
  STW32 killed $w1, $r10, -4 :: (store (s32) into %ir.tpid.addr)
  $r1 = MOV_32_64 $w7, debug-location !34305; progs/profiler.inc.h:377:28
  $r2 = LD_imm64 @bpf_config, debug-location !34303; progs/profiler.inc.h:377:7
  $r2 = LDD killed $r2, 24, debug-location !34303 :: (volatile dereferenceable load (s64) from `ptr getelementptr inbounds nuw (i8, ptr @bpf_config, i64 24)`); progs/profiler.inc.h:377:7
  $r2 = SRL_rr killed $r2(tied-def 0), killed $r1, debug-location !34308; progs/profiler.inc.h:377:37
  $r2 = AND_ri killed $r2(tied-def 0), 1, debug-location !34308; progs/profiler.inc.h:377:37
  JEQ_ri killed $r2, 0, %bb.23, debug-location !34308; progs/profiler.inc.h:377:37

bb.1.if.end:
...
```
In the above, DBG_VALUE is corresponding to dwarf location. But we missed 'DBG_VALUE $w1'. In such cases, it is hard to determine whether the signature is changed or not. So the better way is to preserve the original one.
I will need to debug more in llvm and find a way to fix it (e.g., adding DBG_VALUE $w1).

llvm/lib/Target/BPF/BTFDebug.cpp
yonghong-song: Correct. Will fix.

llvm/lib/Target/BPF/BTFDebug.cpp
yonghong-song: Will do.

llvm/lib/Target/BPF/BTFDebug.cpp
yonghong-song: Will do.

llvm/lib/Target/BPF/BTFDebug.cpp
yonghong-song: You are right. There is no need to have two iterating over basic blocks. Considering all arguments (including stack arguments) in first basic block, to scan the first basic block is enough.

llvm/lib/Target/BPF/BTFDebug.cpp
yonghong-song: You are right. I will fix it in the next revision.

llvm/lib/Target/BPF/BTFDebug.cpp
yonghong-song: Ack. Will do.

llvm/lib/Target/BPF/BTFDebug.h
yonghong-song: Ack.

llvm/lib/Target/BPF/BTFDebug.cpp
yonghong-song: Ack.

llvm/lib/Target/BPF/BTFDebug.cpp
yonghong-song: We have the following 3 cases:
```
   MF->...->isVoidTy() == true   &&.  Elements[0] != nullptr  : ChangedToVoid = true
   MF->...->isVoidTy() == true   &&.  Elements[0] == nullptr : ChangedToVoid = false
   MF->...->isVoidTy() == false  &&.  Elements[0] != nullptr  : ChangedToVoid = false
   MF->...->isVoidTy() == false  &&.  Elements[0] == nullptr : Not possible
```
So I think  `Elements[0] != nullptr` is still needed.

llvm/lib/Target/BPF/BTFDebug.cpp
yonghong-song: The current logic indeed has the issue. I had a new algorithm and hopefully it will fix in the next revision.

llvm/lib/Target/BPF/BTFDebug.cpp
eddyz87: But if the IR level function signature reports that the function is void, why would it matter for the BTF signature, if this changed compared to the original declaration? Having a flag `IsVoid` seem to be logically sufficient. Could you please elaborate a bit on why it is necessary to distinguish if signature *changed* to void instead of just noting that it *is* void?

yonghong-song: This is fixed. See commit 3.

yonghong-song:
> > > The original source prototype is used as fallback when:
> > > ...
> > > The surviving source arg count differs from the IR arg count.
> > > Any surviving source type does not match its IR type (e.g. structs).
> > > The register assignment does not follow R1..R5 order.
> > 
> > 
> > I understand that this is how things worked so far, but since this PR attempts to improve the signatures precision, wouldn't it make sense to drop the signatures to something non-specific like `void(void)` in the cases above as well?
> 
> For such cases, I would like to preserve the original signatures. One of reason is due to llvm which may miss some DBG_VALUE, e.g.
> 
> ```
> # Machine code for function trace_var_sys_kill: NoPHIs, TracksLiveness, NoVRegs, TiedOpsRewritten, TracksDebugUserValues
> Frame Objects:
>   fi#0: size=4, align=4, at location [SP-4]
>   fi#1: size=4, align=4, at location [SP-8]
> save/restore points:
> save points are empty
> restore points are empty
> Function Live Ins: $w1, $w2
> 
> bb.0.entry:
>   successors: %bb.23(0x30000000), %bb.1(0x50000000); %bb.23(37.50%), %bb.1(62.50%)
>   liveins: $w1, $w2
>   DBG_VALUE $w2, $noreg, !"sig", !DIExpression(), debug-location !34299; progs/profiler.inc.h:0 line no:375
>   $w7 = MOV_rr_32 $w2
>   DBG_VALUE $w7, $noreg, !"sig", !DIExpression(), debug-location !34299; progs/profiler.inc.h:0 line no:375
>   STW32 killed $w1, $r10, -4 :: (store (s32) into %ir.tpid.addr)
>   $r1 = MOV_32_64 $w7, debug-location !34305; progs/profiler.inc.h:377:28
>   $r2 = LD_imm64 @bpf_config, debug-location !34303; progs/profiler.inc.h:377:7
>   $r2 = LDD killed $r2, 24, debug-location !34303 :: (volatile dereferenceable load (s64) from `ptr getelementptr inbounds nuw (i8, ptr @bpf_config, i64 24)`); progs/profiler.inc.h:377:7
>   $r2 = SRL_rr killed $r2(tied-def 0), killed $r1, debug-location !34308; progs/profiler.inc.h:377:37
>   $r2 = AND_ri killed $r2(tied-def 0), 1, debug-location !34308; progs/profiler.inc.h:377:37
>   JEQ_ri killed $r2, 0, %bb.23, debug-location !34308; progs/profiler.inc.h:377:37
> 
> bb.1.if.end:
> ...
> ```
> 
> In the above, DBG_VALUE is corresponding to dwarf location. But we missed 'DBG_VALUE $w1'. In such cases, it is hard to determine whether the signature is changed or not. So the better way is to preserve the original one. I will need to debug more in llvm and find a way to fix it (e.g., adding DBG_VALUE $w1).

This is resolved too. See commit 2.

yonghong-song:
With current implementation, I tested with bpf selftests, all meta internal bpf progs and [scx-ext/](https://github.com/sched-ext/scx) bpf programs. All progs have true signature (i.e., not triggering LLVM_DEBUG like below):
```
  if (MF.getFunction().arg_size() != SortedAlive.size()) {
    LLVM_DEBUG(dbgs() << "BTF skip " << MF.getName() << ": IR arg count ("
                      << MF.getFunction().arg_size() << ") != alive arg count ("
                      << SortedAlive.size() << ")\n");
    return false;
  }

  auto ArgIt = MF.getFunction().arg_begin();
  for (unsigned I = 0, N = SortedAlive.size(); I < N; ++I, ++ArgIt) {
    uint32_t ArgNo = SortedAlive[I];
    if (!sourceArgMatchesIRType(Elements[ArgNo], ArgIt->getType())) {
      LLVM_DEBUG(dbgs() << "BTF skip " << MF.getName()
                        << ": type mismatch for source arg " << ArgNo
                        << " at IR position " << I << "\n");
      return false;
    }

    if (I >= std::size(CC_BPF64_ArgRegs))
      continue;

    auto It = EntryRegMap.find(ArgNo);
    assert(It != EntryRegMap.end() &&
           "register arg must have entry-block DBG_VALUE");
    int DwarfReg = TRI.getDwarfRegNum(It->second, false);
    if (DwarfReg != static_cast<int>(I + 1)) {
      LLVM_DEBUG(dbgs() << "BTF skip " << MF.getName() << ": arg " << ArgNo
                        << " in DWARF reg " << DwarfReg << ", expected "
                        << (I + 1) << "\n");
      return false;
    }
  }
```

llvm/lib/Target/BPF/BTFDebug.cpp
```diff
@@ -1344,12 +1577,55 @@ void BTFDebug::beginFunctionImpl(const MachineFunction *MF) {
...
+    SmallDenseMap<uint32_t, Register> EntryRegMap =
+        collectNocallEntryArgRegs(*MF);
+
+    SmallVector<uint32_t, 8> SortedAlive;
+    for (const auto &[ArgNo, Reg] : EntryRegMap)
+      SortedAlive.push_back(ArgNo);
+    llvm::sort(SortedAlive);
+
+    UseFilteredParams = canUseNocallOptimizedSignature(
+        *MF, Elements, SortedAlive, EntryRegMap, *TRI);
```
eddyz87: Nit: I think that it's an unnecessary complication to pass both `SortedAlive` and `EntryRegMap` to the `canUseNocallOptimizedSignature()`. The `EntryRegMap` can be a vector of pairs, sorted after `collectNocallEntryArgRegs()` call. Then it would be possible to only pass `EntryRegMap` to the function and postpone the creation of the indexes array till the `BTFTypeFuncProto` constructor call. Hence, there would be no need to reason about coherency of `SortedAlive` and `EntryRegMap` inside the `canUseNocallOptimizedSignature()`.

eddyz87 (commented):
I think I understand the mechanics, please see one nit below.

llvm/lib/Target/BPF/BTFDebug.cpp
yonghong-song: Good point. Thanks! Will try to simplify the code as you suggested in the above.

yonghong-song: @eddyz87 Just pushed commit 4 to simplify the code as mentioned in the above.

eddyz87 (approved):

yonghong-song:
cc @jemarch @vineetgarc

PR #198427, opened 2026-05-19
[CIR] Lower pointer const_array globals without insertvalue chains

adams381:
`cir.global` initializers that are `const_array` of `global_view` (no
indices) or null pointers were lowered through an initializer region
full of `llvm.insertvalue` ops even though the elements are all
attribute-representable.  That forced the O(N²) MLIR-to-LLVM IR path
on large tables (SPEC CPU 2026 `gcc/insn-automata.cc`).

When `lowerConstArrayAttr` can build the whole initializer, emit the
global with one aggregate attribute instead.  String literals with
`trailing_zeros` are padded into `DenseElementsAttr` so C string tables
take the same bulk path.  Indexed `global_view`, `#cir.zero` arrays, and
other non-bulk cases still use the insertvalue path.

MLIR prerequisite [#198424](https://github.com/llvm/llvm-project/pull/198424) is merged on `main`; this branch is rebased and CIR-only.

### DIFF
--- a/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
+++ b/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
@@ -2014,7 +2014,7 @@ mlir::LogicalResult CIRToLLVMConstantOpLowering::matchAndRewrite(
     if (constArr &&
         (denseAttr = lowerConstArrayAttr(constArr, typeConverter))) {
       attr = denseAttr.value();
-    } else if (constArr) {
+    } else {
       const mlir::Value initVal =
           lowerCirAttrAsValue(op, op.getValue(), rewriter, typeConverter);
       rewriter.replaceOp(op, initVal);
--- a/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
+++ b/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
@@ -484,6 +484,7 @@ mlir::Value CIRAttrToValue::visitCirAttr(cir::ConstArrayAttr attr) {
   // Iteratively lower each constant element of the array.
   if (auto arrayAttr = mlir::dyn_cast<mlir::ArrayAttr>(attr.getElts())) {
     for (auto [idx, elt] : llvm::enumerate(arrayAttr)) {
+      mlir::DataLayout dataLayout(parentOp->getParentOfType<mlir::ModuleOp>());
       mlir::Value init = visit(elt);
       result =
           mlir::LLVM::InsertValueOp::create(rewriter, loc, result, init, idx);
--- a/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
+++ b/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
@@ -1932,13 +1932,33 @@ mlir::LogicalResult CIRToLLVMStoreOpLowering::matchAndRewrite(
   return mlir::LogicalResult::success();
 }
 
+static mlir::Type getConstArrayLeafType(mlir::Type ty) {
+  while (auto arrTy = mlir::dyn_cast<cir::ArrayType>(ty))
+    ty = arrTy.getElementType();
+  return ty;
+}
+
+static bool isBulkLowerableConstArrayLeaf(mlir::Type leafTy) {
+  return mlir::isa<cir::PointerType, cir::IntType, cir::FPTypeInterface>(
+      leafTy);
+}
+
 bool hasTrailingZeros(cir::ConstArrayAttr attr) {
-  auto array = mlir::dyn_cast<mlir::ArrayAttr>(attr.getElts());
-  return attr.hasTrailingZeros() ||
-         (array && std::count_if(array.begin(), array.end(), [](auto elt) {
-            auto ar = dyn_cast<cir::ConstArrayAttr>(elt);
-            return ar && hasTrailingZeros(ar);
-          }));
+  if (attr.hasTrailingZeros())
+    return true;
+
+  auto elts = mlir::dyn_cast<mlir::ArrayAttr>(attr.getElts());
+  if (!elts)
+    return false;
+
+  auto cirArrTy = mlir::dyn_cast<cir::ArrayType>(attr.getType());
+  if (!cirArrTy || !mlir::isa<cir::ArrayType>(cirArrTy.getElementType()))
+    return false;
+
+  return llvm::any_of(elts, [](mlir::Attribute elt) {
+    auto nested = mlir::dyn_cast<cir::ConstArrayAttr>(elt);
+    return nested && hasTrailingZeros(nested);
+  });
 }
 
 mlir::LogicalResult CIRToLLVMConstantOpLowering::matchAndRewrite(
@@ -2490,30 +2510,33 @@ mlir::LogicalResult CIRToLLVMGlobalOpLowering::matchAndRewrite(
         op.emitError() << "unsupported initializer '" << init.value() << "'";
         return mlir::failure();
       }
-    } else if (mlir::isa<cir::ConstArrayAttr, cir::ConstVectorAttr,
-                         cir::ConstRecordAttr, cir::ConstPtrAttr,
-                         cir::ConstComplexAttr, cir::GlobalViewAttr,
-                         cir::TypeInfoAttr, cir::UndefAttr, cir::PoisonAttr,
-                         cir::VTableAttr, cir::ZeroAttr>(init.value())) {
-      if (auto constArr = mlir::dyn_cast<cir::ConstArrayAttr>(init.value())) {
-        if (!hasTrailingZeros(constArr)) {
-          mlir::Type elTy = constArr.getType();
-          while (auto arrTy = mlir::dyn_cast<cir::ArrayType>(elTy))
-            elTy = arrTy.getElementType();
-          if (mlir::isa<cir::PointerType>(elTy)) {
-            mlir::ModuleOp module = op->getParentOfType<mlir::ModuleOp>();
-            if (std::optional<mlir::Attribute> bulkInit =
-                    lowerConstArrayAttr(constArr, typeConverter, module)) {
-              mlir::SymbolRefAttr comdatAttr = getComdatAttr(op, rewriter);
-              rewriter.replaceOpWithNewOp<mlir::LLVM::GlobalOp>(
-                  op, llvmType, isConst, linkage, symbol, bulkInit.value(),
-                  alignment, addrSpace, isDsoLocal, isThreadLocal, comdatAttr,
-                  attributes);
-              return mlir::success();
-            }
-          }
+    } else if (auto constArr =
+                   mlir::dyn_cast<cir::ConstArrayAttr>(init.value())) {
+      // Bulk-emit llvm.mlir.global when lowerConstArrayAttr can build the
+      // whole initializer as one aggregate attribute (no insertvalue
+      // region). Skip arrays with trailing-zero compression: those need
+      // per-element zero/undef setup. Leaf type must match what
+      // lowerConstArrayAttr handles (pointers, integers, floats today).
+      if (!hasTrailingZeros(constArr) &&
+          isBulkLowerableConstArrayLeaf(
+              getConstArrayLeafType(constArr.getType()))) {
+        mlir::ModuleOp module = op->getParentOfType<mlir::ModuleOp>();
+        if (std::optional<mlir::Attribute> bulkInit =
+                lowerConstArrayAttr(constArr, typeConverter, module)) {
+          mlir::SymbolRefAttr comdatAttr = getComdatAttr(op, rewriter);
+          rewriter.replaceOpWithNewOp<mlir::LLVM::GlobalOp>(
+              op, llvmType, isConst, linkage, symbol, bulkInit.value(),
+              alignment, addrSpace, isDsoLocal, isThreadLocal, comdatAttr,
+              attributes);
+          return mlir::success();
         }
       }
+      return matchAndRewriteRegionInitializedGlobal(op, init.value(), rewriter);
+    } else if (mlir::isa<cir::ConstVectorAttr, cir::ConstRecordAttr,
+                         cir::ConstPtrAttr, cir::ConstComplexAttr,
+                         cir::GlobalViewAttr, cir::TypeInfoAttr, cir::UndefAttr,
+                         cir::PoisonAttr, cir::VTableAttr, cir::ZeroAttr>(
+                   init.value())) {
       // TODO(cir): once LLVM's dialect has proper equivalent attributes this
       // should be updated. For now, we use a custom op to initialize globals
       // to the appropriate value.
--- a/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
+++ b/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
@@ -1932,15 +1932,15 @@ mlir::LogicalResult CIRToLLVMStoreOpLowering::matchAndRewrite(
   return mlir::LogicalResult::success();
 }
 
-static mlir::Type getConstArrayLeafType(mlir::Type ty) {
+static mlir::Type getConstArrayBaseElementType(mlir::Type ty) {
   while (auto arrTy = mlir::dyn_cast<cir::ArrayType>(ty))
     ty = arrTy.getElementType();
   return ty;
 }
 
-static bool isBulkLowerableConstArrayLeaf(mlir::Type leafTy) {
+static bool isBulkLowerableConstArrayBaseElement(mlir::Type baseElemTy) {
   return mlir::isa<cir::PointerType, cir::IntType, cir::BoolType,
-                   cir::FPTypeInterface>(leafTy);
+                   cir::FPTypeInterface>(baseElemTy);
 }
 
 mlir::LogicalResult CIRToLLVMConstantOpLowering::matchAndRewrite(
@@ -2494,11 +2494,11 @@ mlir::LogicalResult CIRToLLVMGlobalOpLowering::matchAndRewrite(
       // region). Leaf type must match what lowerConstArrayAttr handles
       // (pointers, integers, bools, floats, and string literals with
       // trailing_zeros).
-      if (isBulkLowerableConstArrayLeaf(
-              getConstArrayLeafType(constArr.getType()))) {
-        mlir::ModuleOp module = op->getParentOfType<mlir::ModuleOp>();
+      if (isBulkLowerableConstArrayBaseElement(
+              getConstArrayBaseElementType(constArr.getType()))) {
+        mlir::ModuleOp modOp = op->getParentOfType<mlir::ModuleOp>();
         if (std::optional<mlir::Attribute> bulkInit =
-                lowerConstArrayAttr(constArr, typeConverter, module)) {
+                lowerConstArrayAttr(constArr, typeConverter, modOp)) {
           mlir::SymbolRefAttr comdatAttr = getComdatAttr(op, rewriter);
           rewriter.replaceOpWithNewOp<mlir::LLVM::GlobalOp>(
               op, llvmType, isConst, linkage, symbol, bulkInit.value(),
@@ -2539,13 +2539,13 @@ CIRToLLVMGlobalOpLowering::getComdatAttr(cir::GlobalOp &op,
   if (!op.getComdat())
     return mlir::SymbolRefAttr{};
 
-  mlir::ModuleOp module = op->getParentOfType<mlir::ModuleOp>();
+  mlir::ModuleOp modOp = op->getParentOfType<mlir::ModuleOp>();
   mlir::OpBuilder::InsertionGuard guard(builder);
   StringRef comdatName("__llvm_comdat_globals");
   if (!comdatOp) {
-    builder.setInsertionPointToStart(module.getBody());
+    builder.setInsertionPointToStart(modOp.getBody());
     comdatOp =
-        mlir::LLVM::ComdatOp::create(builder, module.getLoc(), comdatName);
+        mlir::LLVM::ComdatOp::create(builder, modOp.getLoc(), comdatName);
   }
 
   if (auto comdatSelector = comdatOp.lookupSymbol<mlir::LLVM::ComdatSelectorOp>(
--- a/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
+++ b/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
@@ -1939,8 +1939,8 @@ static mlir::Type getConstArrayLeafType(mlir::Type ty) {
 }
 
 static bool isBulkLowerableConstArrayLeaf(mlir::Type leafTy) {
-  return mlir::isa<cir::PointerType, cir::IntType, cir::FPTypeInterface>(
-      leafTy);
+  return mlir::isa<cir::PointerType, cir::IntType, cir::BoolType,
+                   cir::FPTypeInterface>(leafTy);
 }
 
 bool hasTrailingZeros(cir::ConstArrayAttr attr) {
@@ -2516,7 +2516,7 @@ mlir::LogicalResult CIRToLLVMGlobalOpLowering::matchAndRewrite(
       // whole initializer as one aggregate attribute (no insertvalue
       // region). Skip arrays with trailing-zero compression: those need
       // per-element zero/undef setup. Leaf type must match what
-      // lowerConstArrayAttr handles (pointers, integers, floats today).
+      // lowerConstArrayAttr handles (pointers, integers, bools, floats).
       if (!hasTrailingZeros(constArr) &&
           isBulkLowerableConstArrayLeaf(
               getConstArrayLeafType(constArr.getType()))) {
--- a/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
+++ b/clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
@@ -484,7 +484,6 @@ mlir::Value CIRAttrToValue::visitCirAttr(cir::ConstArrayAttr attr) {
   // Iteratively lower each constant element of the array.
   if (auto arrayAttr = mlir::dyn_cast<mlir::ArrayAttr>(attr.getElts())) {
     for (auto [idx, elt] : llvm::enumerate(arrayAttr)) {
-      mlir::DataLayout dataLayout(parentOp->getParentOfType<mlir::ModuleOp>());
       mlir::Value init = visit(elt);
       result =
           mlir::LLVM::InsertValueOp::create(rewriter, loc, result, init, idx);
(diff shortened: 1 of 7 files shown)


clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -2510,6 +2509,25 @@ mlir::LogicalResult CIRToLLVMGlobalOpLowering::matchAndRewrite(
                          cir::ConstComplexAttr, cir::GlobalViewAttr,
                          cir::TypeInfoAttr, cir::UndefAttr, cir::PoisonAttr,
                          cir::VTableAttr, cir::ZeroAttr>(init.value())) {
+      if (auto constArr = mlir::dyn_cast<cir::ConstArrayAttr>(init.value())) {
+        if (!hasTrailingZeros(constArr)) {
+          mlir::Type elTy = constArr.getType();
+          while (auto arrTy = mlir::dyn_cast<cir::ArrayType>(elTy))
+            elTy = arrTy.getElementType();
+          if (mlir::isa<cir::PointerType>(elTy)) {
```
erichkeane: I was chasing down a similar issue in self build and noticed that it seems to skip this?  I'm still debugging, but why is this limited to pointer types, and not extended to other trivially copyable types?

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -483,7 +483,6 @@ mlir::Value CIRAttrToValue::visitCirAttr(cir::ConstArrayAttr attr) {
   // Iteratively lower each constant element of the array.
   if (auto arrayAttr = mlir::dyn_cast<mlir::ArrayAttr>(attr.getElts())) {
     for (auto [idx, elt] : llvm::enumerate(arrayAttr)) {
-      mlir::DataLayout dataLayout(parentOp->getParentOfType<mlir::ModuleOp>());
```
erichkeane, line 487: This change isn't really obvious to me, what is this for?   If this is an unrelated drive-by, it should be done in a different commit.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -2510,6 +2509,25 @@ mlir::LogicalResult CIRToLLVMGlobalOpLowering::matchAndRewrite(
                          cir::ConstComplexAttr, cir::GlobalViewAttr,
                          cir::TypeInfoAttr, cir::UndefAttr, cir::PoisonAttr,
                          cir::VTableAttr, cir::ZeroAttr>(init.value())) {
+      if (auto constArr = mlir::dyn_cast<cir::ConstArrayAttr>(init.value())) {
```
erichkeane: This is doing quite a bit of work fairly that is pretty subtle and limited. Can we comment/document the limitations here, and why they exist?

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -2510,6 +2509,25 @@ mlir::LogicalResult CIRToLLVMGlobalOpLowering::matchAndRewrite(
...
+        if (!hasTrailingZeros(constArr)) {
```
erichkeane: This for example: why limited to trailing zeros?

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
erichkeane: Also, this 'if' is a sub-set of the 'if' on 2507, right?  Can we just hoist this out of the loop here?   The fallthru in the 'not' case is sooo small, that it seems trivial to just put at the bottom of the new block.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
erichkeane: More I think about it, we can do this for basically any type that `lowerConstArrayAttr` can handle that doesn't require any sort of copying. 

As an 'inverse', we could probably just do something like `!mlir::isa<cir::RecordType>(elTy)` (to avoid any sort of copying issues).  But in the meantime,we should probably at least try `cir.bool`, `cir.int`, and the `FPInterfaceType`

clang/lib/CIR/Lowering/LoweringHelpers.cpp
```diff
@@ -132,6 +196,13 @@ lowerConstArrayAttr(cir::ConstArrayAttr constArr,
     type = arrayType.getElementType();
   }
 
+  if (auto eltsArr = mlir::dyn_cast<mlir::ArrayAttr>(constArr.getElts())) {
+    for (mlir::Attribute elt : eltsArr) {
+      if (mlir::isa<cir::PoisonAttr>(elt))
+        return std::nullopt;
```
erichkeane: is this enough?  should we recurse if the innser type is a ConstArrayAttr?   Perhaps implement some sort of `ContainsPoison` on our cir-attrs (or just a helper function?) that will look into other types.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
```diff
@@ -115,9 +117,71 @@ mlir::DenseElementsAttr convertToDenseElementsAttr(
...
+  if (auto llvmSymbol = mlir::dyn_cast<mlir::LLVM::GlobalOp>(sourceSymbol)) {
+    sourceType = llvmSymbol.getType();
+    sourceAddrSpace = llvmSymbol.getAddrSpace();
+  } else if (auto cirSymbol = mlir::dyn_cast<cir::GlobalOp>(sourceSymbol)) {
+    sourceType = converter->convertType(cirSymbol.getSymType());
+    if (auto targetAS = mlir::dyn_cast_if_present<cir::TargetAddressSpaceAttr>(
+            cirSymbol.getAddrSpaceAttr()))
+      sourceAddrSpace = targetAS.getValue();
+  } else {
+    return false;
```
erichkeane, line 170: What types get here?  Can we comment on why we are skipping them?  At least FuncOp, AliasOp (and apparently LLVMFuncOp?) can get here I think, right?

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
erichkeane: Hmm... this is actually pretty slow, and you're calling it on large arrays!  `count_if` does a full scan.

We probably need now that you're using it this way, to change this to exit early in some cases.

I think if we were to use `any_of` at least in the positive case it'd escape early.

Also, this just assumes that any array is possibly a parent to an array... somewhere?  We should add a condition that only does this if we know that `attr` has an element type of `array` or something.  else we are searching through a bunch of 'ints' for no reason.

erichkeane (commented):
I also don't see any tests that check:

-multi-dim arrays/pointer arrays 
-arrays with mismatch address spaces
-trailing zeros test

adams381:
Added \`const-array-bulk-lowering-fallbacks.cir\` for trailing-zero (\`@tz_arr\`), address-space mismatch (\`@as_mismatch\`), and multi-dim int bulk (\`@matrix\` + \`dense<\`). Multi-dim pointer fallback is \`@ptr2d\` (insertvalue when inner globals use \`trailing_zeros\`). Codegen covers 1D pointers, int/bool/multi-dim int in \`global-pointer-array-fast-lowering.cpp\`.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Bulk path now keys off \`isBulkLowerableConstArrayLeaf\` (pointers, ints, bools, floats) and whatever \`lowerConstArrayAttr\` can build. Indexed \`global_view\`, trailing-zero padding, and addr-space mismatches still use insertvalue.

adams381: Added \`cir.bool\` to the leaf check and \`lowerConstArrayAttr\` dense path; \`global-pointer-array-fast-lowering.cpp\` has a \`flags[]\` case. Records stay on insertvalue for now since they need per-field lowering.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Removed an unused \`DataLayout\` inside the \`const_array\` insertvalue loop — it was never read. No behavior change.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Added a block comment above the \`const_array\` branch: trailing-zero skip, leaf-type gate, and fallthrough to \`matchAndRewriteRegionInitializedGlobal\` when bulk init fails.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Trailing-zero arrays need per-element zero/undef in the initializer region; a single aggregate attr cannot represent that compression. Same comment block explains it.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Hoisted to its own \`else if (constArr)\` before the other region-initialized types; fallthrough is \`matchAndRewriteRegionInitializedGlobal\`.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
adams381: Replaced the flat scan with \`containsPoison()\` that recurses through nested \`const_array\` and \`ArrayAttr\` elements.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
adams381: Comment notes funcs and other symbols not lowered to globals yet (e.g. \`cir.func\`) — those keep the insertvalue fallback.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Switched to \`llvm::any_of\` for early exit and only recurse when the CIR array's element type is itself an array, so flat int/pointer tables do not scan nested attrs.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
erichkeane: I looked a bit closer at this, its actually somewhat expensive?  Might be worth doing as an NFC anyway in a separate patch.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -1959,13 +1958,33 @@ mlir::LogicalResult CIRToLLVMStoreOpLowering::matchAndRewrite(
...
-            return ar && hasTrailingZeros(ar);
-          }));
+  if (attr.hasTrailingZeros())
+    return true;
+
+  auto elts = mlir::dyn_cast<mlir::ArrayAttr>(attr.getElts());
+  if (!elts)
+    return false;
+
+  auto cirArrTy = mlir::dyn_cast<cir::ArrayType>(attr.getType());
```
erichkeane: CAN this fail?

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -1959,13 +1958,33 @@ mlir::LogicalResult CIRToLLVMStoreOpLowering::matchAndRewrite(
...
-  auto array = mlir::dyn_cast<mlir::ArrayAttr>(attr.getElts());
-  return attr.hasTrailingZeros() ||
-         (array && std::count_if(array.begin(), array.end(), [](auto elt) {
-            auto ar = dyn_cast<cir::ConstArrayAttr>(elt);
-            return ar && hasTrailingZeros(ar);
-          }));
+  if (attr.hasTrailingZeros())
+    return true;
+
+  auto elts = mlir::dyn_cast<mlir::ArrayAttr>(attr.getElts());
```
erichkeane: I'd suggest putting this right before the 'any_of'.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -2505,11 +2524,33 @@ mlir::LogicalResult CIRToLLVMGlobalOpLowering::matchAndRewrite(
...
-                         cir::ConstRecordAttr, cir::ConstPtrAttr,
-                         cir::ConstComplexAttr, cir::GlobalViewAttr,
-                         cir::TypeInfoAttr, cir::UndefAttr, cir::PoisonAttr,
-                         cir::VTableAttr, cir::ZeroAttr>(init.value())) {
+    } else if (auto constArr =
+                   mlir::dyn_cast<cir::ConstArrayAttr>(init.value())) {
+      // Bulk-emit llvm.mlir.global when lowerConstArrayAttr can build the
+      // whole initializer as one aggregate attribute (no insertvalue
+      // region). Skip arrays with trailing-zero compression: those need
+      // per-element zero/undef setup. Leaf type must match what
```
erichkeane: Do they?  Why?  Or at least, only the actual zeros do, right? Wonder if we should be doing this as a pair of memcpys? 

Worth a fixme at least to see if we can optimize those out... just seems like a shame that:
```
const char some_really_long_thing[50000000001] = {
// woops, only 50000000000 values, so we have trailing zeros!
};
```
wouldn't benefit from this.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -1959,13 +1958,33 @@ mlir::LogicalResult CIRToLLVMStoreOpLowering::matchAndRewrite(
   return mlir::LogicalResult::success();
 }
 
+static mlir::Type getConstArrayLeafType(mlir::Type ty) {
```
erichkeane: is 'leaf' used as a term of art?  In clang we typically call this `BaseElementType`

clang/lib/CIR/Lowering/LoweringHelpers.cpp
```diff
@@ -78,6 +81,12 @@ void convertToDenseElementsAttrImpl(
 
   auto arrayAttr = mlir::cast<mlir::ArrayAttr>(attr.getElts());
   for (auto eltAttr : arrayAttr) {
+    if (auto boolAttr = mlir::dyn_cast<cir::BoolAttr>(eltAttr)) {
+      if constexpr (std::is_same_v<StorageTy, mlir::APInt>) {
+        values[currentIndex++] = mlir::APInt(1, (uint64_t)boolAttr.getValue());
```
erichkeane:
```suggestion
        values[currentIndex++] = mlir::APInt(1, static_cast<uint64_t>(boolAttr.getValue()));
```

though, is this cast necessary at all?  `getValue` should get a `bool`, right?  Which has an implicit cast to int.

Also, why `mlir::APInt`?  Is this different than the llvm::APInt we use everywhere else?  If not, we should call it THAT instead.

erichkeane (commented):
A few more comments, but would also like @andykaylor to take a look/approve before merging.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
```diff
@@ -2505,11 +2524,33 @@ mlir::LogicalResult CIRToLLVMGlobalOpLowering::matchAndRewrite(
...
+                   mlir::dyn_cast<cir::ConstArrayAttr>(init.value())) {
+      // Bulk-emit llvm.mlir.global when lowerConstArrayAttr can build the
+      // whole initializer as one aggregate attribute (no insertvalue
+      // region). Skip arrays with trailing-zero compression: those need
+      // per-element zero/undef setup. Leaf type must match what
+      // lowerConstArrayAttr handles (pointers, integers, bools, floats).
+      if (!hasTrailingZeros(constArr) &&
+          isBulkLowerableConstArrayLeaf(
+              getConstArrayLeafType(constArr.getType()))) {
+        mlir::ModuleOp module = op->getParentOfType<mlir::ModuleOp>();
```
andykaylor: Don't use `module` as an identifier. It's a reserved keyword in C++20.

clang/test/CIR/CodeGen/global-pointer-array-fast-lowering.cpp
```diff
@@ -0,0 +1,37 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-cir %s -o %t.cir
```
andykaylor, line 1: It's not clear to me what this test is doing. At a glance this looks like what we produce without this PR.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
erichkeane: This is actually problematic unfortunately, I found this with self-build.  Just:
`const char * foo = "asdf"` skips this and goes to 'insert-value'.  So any long string literal is going to take ages to compile thanks to this. 

One of your previous iterations (when I added 'IntType' and 'FloatingTypeInterface' to the list of alloweds) didn't seem to have this problem, so the `trailing-zeros' check is perhaps a regression.

adams381:
Pushed follow-up for the trailing_zeros string-table path: const_array with trailing_zeros now folds into DenseElementsAttr in LoweringHelpers, and the hasTrailingZeros bypass is gone from cir.constant and pointer-array globals. C string globals (e.g. const char *foo = \"asdf\") and tz_arr/s0/s1 in the lowering tests take dense bulk init instead of insertvalue.

@erichkeane — this is the compile-time fix for the gcc-scale string tables you flagged on Slack.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Fixed in the follow-up commit —  now pads trailing zeros into the `DenseElementsAttr` (the `values` vector is pre-zeroed; the trailing-zero slots are left as-is). The `hasTrailingZeros` bypass in `CIRToLLVMConstantOpLowering` and the pointer-array path are both removed, so `const char * foo = "asdf"` goes through the bulk path. The test covers both the string pointer global and int/string arrays with trailing zeros and asserts no insertvalue.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: The standalone `hasTrailingZeros` helper was removed in the follow-up commits. Trailing zeros are now handled inside `lowerConstArrayAttr` directly: for int/bool/float arrays the pre-zeroed `values` vector covers them; for pointer arrays with `trailing_zeros`, `lowerConstArrayAttr` returns `nullopt` and the caller falls back to the insertvalue path.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: The `any_of`/`containsPoison` approach was replaced by the `isBulkLowerableConstArrayBaseElement` gate, so there is no `any_of` in the current code.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Renamed: `getConstArrayLeafType` → `getConstArrayBaseElementType`, `isBulkLowerableConstArrayLeaf` → `isBulkLowerableConstArrayBaseElement`.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
adams381: Applied: `llvm::APInt(1, static_cast<uint64_t>(boolAttr.getValue()))`. `mlir::APInt` is just a namespace alias for `llvm::APInt` so there is no semantic difference, but `llvm::APInt` is the canonical name.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Renamed to `modOp`.

clang/test/CIR/CodeGen/global-pointer-array-fast-lowering.cpp
adams381: Added a comment block at the top of the test file explaining it verifies that globals with pointer, integer, bool, float, and string-literal base element types take the bulk path (single `llvm.mlir.global` aggregate, no insertvalue chain).

adams381:
Pushed a follow-up NFC commit addressing the clang-format warning: restored the `nsw` parameter accidentally dropped from `createNeg`, fixed `SmallVector` explicit size, cleaned up the FIXME comment, removed the test-file preamble, and collapsed the separate OGCG prefix into the shared `LLVM` prefix.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
```diff
@@ -12,28 +12,47 @@
...
 #include "clang/CIR/LoweringHelpers.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/Dialect/LLVMIR/LLVMTypes.h"
+#include "mlir/IR/SymbolTable.h"
 #include "clang/CIR/MissingFeatures.h"
 
+static unsigned getIntOrBoolBitWidth(mlir::Type ty) {
+  if (auto intTy = mlir::dyn_cast<cir::IntType>(ty))
+    return intTy.getWidth();
+  if (mlir::isa<cir::BoolType>(ty))
```
erichkeane:
```suggestion
  assert(mlir::isa<cir::BoolType>(ty));
```
Then you can just remove the unreachable.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
```diff
@@ -12,28 +12,47 @@
...
+         "string const_array size must match explicit elements plus "
+         "trailing_zeros");
+
+  const unsigned bitWidth = getIntOrBoolBitWidth(arrayTy.getElementType());
+  llvm::SmallVector<mlir::APInt> values;
+  values.reserve(totalSize);
 
   for (const char element : stringAttr)
-    values.push_back({8, (uint64_t)element});
+    values.emplace_back(bitWidth, (uint64_t)(unsigned char)element);
```
erichkeane: Nit: stop using "C" style casts, its against our coding standard. 

Second, you shouldn't have to do either cast.  `char` converts implicitly to `uint64_t`. It was required before because it was using push_back.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
```diff
@@ -12,28 +12,47 @@
...
 
-  const auto arrayTy = mlir::cast<cir::ArrayType>(attr.getType());
-  if (arrayTy.getSize() != stringAttr.size())
-    assert(!cir::MissingFeatures::stringTypeWithDifferentArraySize());
+  for (unsigned i = 0; i < trailingZeros; ++i)
+    values.push_back(mlir::APInt::getZero(bitWidth));
```
erichkeane: `values.insert(values.back(), trailingZeros, mlir::APInt::getZero(bitWidth));`.

Skip the loop entirely.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
```diff
@@ -12,28 +12,47 @@
...
 
   return mlir::DenseElementsAttr::get(
-      mlir::RankedTensorType::get({(int64_t)values.size()}, type),
+      mlir::RankedTensorType::get({(int64_t)totalSize}, type),
```
erichkeane: Can you try removing the cast?  I would expect this to just work with normal conversion rules.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
adams381: Switched to an `assert(isa<BoolType>)` plus a plain `return 1;`.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
adams381: Dropped both casts; the `char` widens to `uint64_t` implicitly and `APInt` masks to the element width.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
adams381: Replaced the loop with `values.insert(values.end(), trailingZeros, mlir::APInt::getZero(bitWidth));`.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
adams381: Removed; `{totalSize}` converts to `int64_t` under normal rules.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
```diff
@@ -78,6 +97,13 @@ void convertToDenseElementsAttrImpl(
 
   auto arrayAttr = mlir::cast<mlir::ArrayAttr>(attr.getElts());
   for (auto eltAttr : arrayAttr) {
+    if (auto boolAttr = mlir::dyn_cast<cir::BoolAttr>(eltAttr)) {
+      if constexpr (std::is_same_v<StorageTy, mlir::APInt>) {
```
erichkeane: can we put the `if constexpr` around the cast attempt?  Optimizers have a tough time removing our dyn-casts.

erichkeane (approved):
1 nit, else lgtm.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
adams381: Hoisted the `if constexpr` above the `dyn_cast` so the cast is only instantiated for the integer storage type.

clang/test/CIR/CodeGen/global-pointer-array-fast-lowering.cpp
adams381: The `@names` global is the case this PR changes: it's a `[3 x ptr]` whose elements are `global_view` (FlatSymbolRefAttr) constants.  Before this PR those reached the per-element `llvm.insertvalue` path; now `lowerConstArrayAttr` emits a single flat array constant, which is what the `LLVM-NOT: insertvalue` after `@names` pins.  The integer/string/bool/float globals were already bulk-lowered before this PR — they're here only so one file covers every base element type that reaches the helper, not because they change.

clang/lib/CIR/Lowering/DirectToLLVM/LowerToLLVM.cpp
adams381: Restored the line, so this PR is back to just the pointer const_array lowering.  The unused per-element `DataLayout` removal will go up as its own NFC.

clang/lib/CIR/Lowering/LoweringHelpers.cpp
```diff
@@ -12,28 +12,45 @@
...
+  llvm::SmallVector<mlir::APInt> values;
+  values.reserve(totalSize);
 
   for (const char element : stringAttr)
-    values.push_back({8, (uint64_t)element});
+    values.emplace_back(bitWidth, element);
 
-  const auto arrayTy = mlir::cast<cir::ArrayType>(attr.getType());
-  if (arrayTy.getSize() != stringAttr.size())
-    assert(!cir::MissingFeatures::stringTypeWithDifferentArraySize());
```
andykaylor, line 28: This can be removed from MissingFeatures.h now.

andykaylor (approved):
lgtm, with one last nit

clang/lib/CIR/Lowering/LoweringHelpers.cpp
adams381: Removed the `stringTypeWithDifferentArraySize()` declaration. Also dropped the `MissingFeatures.h` include from `LoweringHelpers.cpp` since that was the only remaining use in the file.

PR #198429, opened 2026-05-19
[lldb] Make CommandObject::GetTarget filter out the dummy target

JDevlieghere:
Follow-up to #197805. Make CommandObject::GetTarget the canonical target accessor for command code, and tighten its semantics so that DoExecute methods can't accidentally operate on the dummy target.

GetTarget now returns Target* instead of Target&. The result is the target from the command's frozen execution context, falling back to the interpreter's execution context. The dummy target is filtered out and replaced with nullptr unless the command opts in via one of the eCommandRequires{Target,Process,Thread,Frame} flags (in which case CheckRequirements has already guaranteed a real target) or via the new eCommandAllowsDummyTarget flag.

This is the first half of the cleanup discussed at the end of #197805. A follow-up will audit DoExecute methods that still reach for GetSelectedTarget or m_exe_ctx.GetTargetPtr() directly and migrate them to GetTarget.

### DIFF
--- a/lldb/include/lldb/Interpreter/CommandInterpreter.h
+++ b/lldb/include/lldb/Interpreter/CommandInterpreter.h
@@ -478,7 +478,13 @@ public:
     return m_debugger.GetTargetList().GetSelectedTarget();
   }
 
-  ExecutionContext GetExecutionContext() const;
+  /// Returns the execution context the interpreter should run a command in.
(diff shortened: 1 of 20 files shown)


JDevlieghere:
This is reland after this was reverted in https://github.com/llvm/llvm-project/pull/198325 because `CommandObjectTargetModulesList` didn't specify `eCommandRequiresTarget`.

PR #179553, opened 2026-02-03
Reland "[NVPTX][AtomicExpandPass] Complete support for AtomicRMW in NVPTX (#176015)"

akshayrdeodhar:
This PR adds full support for atomicrmw in NVPTX. This includes:

- Memory order and syncscope support (changes in AtomicExpandPass.cpp, NVPTXIntrinsics.td)

- Script-generated tests for integer and atomic operations (atomicrmw.py, atomicrmw-sm*.ll in tests/CodeGen/NVPTX). Existing atomics tests which are subsumed by these have been removed (atomics-sm*.ll, atomics.ll, atomicrmw-expand.ll).

- ~~Changes shouldExpandAtomicRMWInIR to take a constant argument: This is to allow some other TargetLowering constant-argument functions to call it. This change touches several backends. An alternative solution exists, but to me, this seems the "right" way.~~ Has been split out into https://github.com/llvm/llvm-project/pull/176073. Rebased.

- NOTE: The initial load issued for atomicrmw emulation loops (and cmpxchg emulation loops) must be a strong load. Currently, AtomicExpandPass issues a weak load. Fixing this breaks several backends. I'm planning to follow up with a separate PR.

Initially failed due to error: ptxas fatal   : Value 'sm_60' is not
defined for option 'gpu-name'. Updated RUN lines in atomicrmw-sm*.py to
skip the ptxas-verify check if ptxas does not support that SM version.

PR #198431, opened 2026-05-19
[NVPTX] Autoupgrade atom.load intrinsics to device scope

akshayrdeodhar:
The default scope for the "atom" instruction in PTX is "gpu" (corresponds to "device" in LLVM). The default LLVM scope is "system". 

When scopes were not supported, the "system" scope was silently dropped, and the default device scope was desired behavior. 

This patch fixes the discrepancy introduced by https://github.com/llvm/llvm-project/pull/179553.


dakersnar (approved):

schwarzschild-radius (approved):
LGTM, Thanks!

mjulian31 (approved):

PR #198432, opened 2026-05-19
[AMDGPU] Add three target features msad-insts, mqsad-pk-insts, and mqsad-insts

shiltian:

PR #198433, opened 2026-05-19
[NFC][AMDGPU] Move AMDGPU related code out of generic TargetParser.cpp

shiltian:


shiltian:
* #198433 👈
* #198432
* `main`

<h2></h2>

clang/lib/Basic/Targets/AMDGPU.cpp
```diff
@@ -17,6 +17,7 @@
 #include "clang/Basic/MacroBuilder.h"
 #include "clang/Basic/TargetBuiltins.h"
 #include "llvm/ADT/SmallString.h"
+#include "llvm/TargetParser/AMDGPUTargetParser.h"
```
rampitec, line 20: You already have it included in the `AMDGPU.h`.

clang/lib/Basic/Targets/SPIR.cpp
```diff
@@ -14,7 +14,7 @@
 #include "AMDGPU.h"
 #include "clang/Basic/MacroBuilder.h"
 #include "clang/Basic/TargetBuiltins.h"
-#include "llvm/TargetParser/TargetParser.h"
+#include "llvm/TargetParser/AMDGPUTargetParser.h"
```
rampitec, line 17: Same here.

clang/lib/Driver/ToolChains/AMDGPU.cpp
```diff
@@ -22,8 +22,8 @@
 #include "llvm/Support/Path.h"
 #include "llvm/Support/Process.h"
 #include "llvm/Support/VirtualFileSystem.h"
+#include "llvm/TargetParser/AMDGPUTargetParser.h"
```
rampitec, line 25: Same.

clang/lib/Basic/Targets/AMDGPU.cpp
shiltian: Right, but my understanding is that the best practice is to include the headers we actually need instead of relying on transitive includes from other headers, since those headers might later stop including the ones we depend on.

That probably won't happen with `AMDGPU.h`, since it's basically a giant all-in-one header included by almost every AMDGPU source file, but IMHO that's not really a good thing.

rampitec: Up to you.

rampitec (approved):
LGTM

PR #185425, opened 2026-03-09
[llvm][tools] Extend llvm-objdump to support nested OffloadBinaries

adurang:
Extends llvm-objdump to print the information of images contained in nested OffloadBinaries. For example, for a binary compiled with #185413 it shows
```
 $llvm-objdump --offloading ./a.out

./a.out:        file format elf64-x86-64

OFFLOADING IMAGE [0]:
kind            elf
arch
triple          spirv64-intel
producer        openmp
image size      43104 bytes
  [Nested OffloadBinary format detected]
  Number of inner images: 1
  kind            spir-v
  arch
  triple          spirv64-intel
  producer        openmp
  image size      42944 bytes
```  

New tests are added for clang-ling-wrapper and llvm-offload-binary using this new functionality.

Depends on #185413

PR #125737, opened 2025-02-04
[clang-linker-wrapper] Add ELF packaging for spirv64-intel OpenMP images

sarnex:
Add manual ELF packaging for `spirv64-intel` images as there is no SPIR-V linker available. This format will be expected by the runtime plugin we will submit in the future and is compatible with the format we already use downstream.

PR #198434, opened 2026-05-19
[clang][test] Move offload tools tests to clang/test/OffloadTools

YuriPlyakhin:
Tests for clang-offload-bundler, clang-linker-wrapper, clang-nvlink-wrapper, and clang-sycl-linker were scattered across clang/test/Driver/ and clang/test/Tooling/. Neither location was appropriate: Driver/ is for testing the compiler driver, and Tooling/ is for testing the LibTooling infrastructure.

Move them to clang/test/OffloadTools/ with one subdirectory per tool, following the convention used by llvm/test/tools/. The tool-name prefix is dropped from filenames since the containing directory already identifies the tool.

The spirv-tools feature/substitution setup is carried over from Tooling's lit.local.cfg into a new lit.local.cfg in clang-linker-wrapper/, since the parent clang/test/lit.cfg.py does not set up that feature. The dead spirv-tools block is removed from Tooling/lit.local.cfg now that no tests there depend on it.

`clang-linker-wrapper-spirv.cpp` was relying on libomp in the environment. It is now rewritten to run the `clang-linker-wrapper` directly rather than going through the driver and added stub symbols allow avoiding linking against OpenMP runtime libraries.


bader:
I recommend using the directory layout and naming suggested by @sarnex in this comment - https://github.com/llvm/llvm-project/pull/197344#issuecomment-4463506612. Matching source code layout in test directory makes it easier to maintain.

YuriPlyakhin:
> I recommend using the directory layout and naming suggested by @sarnex in this comment - [#197344 (comment)](https://github.com/llvm/llvm-project/pull/197344#issuecomment-4463506612). Matching source code layout in test directory makes it easier to maintain.

@bader , I tried it initially, but
https://github.com/llvm/llvm-project/blob/d63ca96ec35e963e651d1b8765283117ca83e1f9/clang/test/CMakeLists.txt#L238
recursively walks all subdirectories and generates a target named `check-clang-<dirname>` for each one.
`clang/test/tools` directory generates `check-clang-tools` target, which collides with the target already created here:
https://github.com/llvm/llvm-project/blob/d63ca96ec35e963e651d1b8765283117ca83e1f9/clang-tools-extra/CMakeLists.txt#L12

I have not found a better solution, so decided to name parent folder differently to avoid collision. If you have better idea, please, let me know.

Other than the parent folder name, the directory layout and naming of test directory is matching source code layout.

clang/test/OffloadTools/clang-linker-wrapper/openmp-spirv.cpp
```diff
@@ -0,0 +1,27 @@
+// Verify the ELF packaging of OpenMP SPIR-V device images.
```
sarnex, line 1: could we rename this test to `openmp-spirv`?

sarnex (approved):
lgtm given the restriction of using `clang/test/tools`

clang/test/OffloadTools/clang-linker-wrapper/openmp-spirv.cpp
```diff
@@ -0,0 +1,27 @@
...
+// REQUIRES: system-linux
+// REQUIRES: spirv-tools
+// REQUIRES: spirv-registered-target
+// REQUIRES: x86-registered-target
+
+// RUN: %clang -cc1 %s -triple x86_64-unknown-linux-gnu -emit-llvm-bc -o %t.bc
+// RUN: %clang -cc1 %s -triple spirv64-intel -fopenmp-is-target-device -o %t.device.o
+// RUN: llvm-offload-binary -o %t.bundle --image=file=%t.device.o,triple=spirv64-intel,arch=generic,kind=openmp
+// RUN: %clang -cc1 -triple x86_64-unknown-linux-gnu -emit-obj -fembed-offload-object=%t.bundle -x ir %t.bc -o %t.host.o
```
bader: Please, use %clang_cc1 substitution.

These are needed to create an input file for the `clang-linker-wrapper` tool we are testing. @sarnex, @adurang, do you have any thoughts on optimizing these four lines? 
For instance, can we use an empty file for %t.device.o?

bader (commented):
> The spirv-tools feature/substitution setup is carried over from Tooling's lit.local.cfg into a new lit.local.cfg in clang-linker-wrapper/, since the parent clang/test/lit.cfg.py does not set up that feature. The dead spirv-tools block is removed from Tooling/lit.local.cfg now that no tests there depend on it.

Why do OffloadTools tests depend on spirv-tools? I expect only SPIR-V backend tests need those.

clang/test/OffloadTools/clang-linker-wrapper/openmp-spirv.cpp
sarnex: This is an existing test I wrote like a year ago, Yury just moved/renamed it from `clang/test/Tooling/clang-linker-wrapper-spirv.cpp`. 

I'm not sure we need to make any content changes to it as part of this PR, it's been working fine,

bader: > This is an existing test I wrote like a year ago, Yury just moved/renamed it from `clang/test/Tooling/clang-linker-wrapper-spirv.cpp`.

`clang/test/Tooling/clang-linker-wrapper-spirv.cpp` has been added by @adurang in https://github.com/llvm/llvm-project/pull/185425 two weeks ago.
It's not a simple "move/rename" - Yury changed the object file generation commands!

> I'm not sure we need to make any content changes to it as part of this PR, it's been working fine,

However, the fact that it works today doesn’t mean it’s the best approach, and it’s worth considering whether we can simplify or make the test more robust.

So my question is: do we really need to invoke clang to produce the input files for testing the llvm-objdump functionality introduced in https://github.com/llvm/llvm-project/pull/185425, or are there simpler/more direct ways to construct the required test inputs?

sarnex: Oh sorry, I assumed it was only renamed with no changes and also I got which test it was wrong, double my bad.

I'll let Alex respond but to be honest I would only support getting rid of the clang call if the new version is relatively simply, if it's some crazy manual ELF section creation I don't think the complexity is worth it.

bader: It seems to me that `clang/test/Tooling/clang-linker-wrapper-spirv.cpp` shouldn't belong to https://github.com/llvm/llvm-project/pull/185425. I mean the goal of this test is not to test `llvm-objdump` functionality, but rather test that OpenMP offload compiler produces new object format, which was not supported before https://github.com/llvm/llvm-project/pull/185425.

Based on Yury's changes, he interprets `spirv.cpp` as a `clang-linker-wrapper` functionality test, rather than an OpenMP offload compiler test.

clang/test/OffloadTools/clang-linker-wrapper/openmp-spirv.cpp
YuriPlyakhin: https://github.com/llvm/llvm-project/pull/198434/changes/ebb356550ac182c25082e5f3a4e658e566396899

clang/test/OffloadTools/clang-linker-wrapper/openmp-spirv.cpp
YuriPlyakhin: I don't think this test is testing llvm-objdump functionality introduced in #185425 . And yes, I agree that `clang-linker-wrapper-spirv.cpp` test should not belong to #185425 . The test is rather testing this pr: https://github.com/llvm/llvm-project/pull/185413. #185425 just extended llvm-objdump, so the `clang-linker-wrapper` functionality introduced in #185413 can be tested.

full history:
https://github.com/llvm/llvm-project/pull/125737 - adds new clang-linker-wrapper functionality, adds new test: `clang/test/Tooling/clang-linker-wrapper-spirv-elf.cpp`
https://github.com/llvm/llvm-project/pull/185413 - changes clang-linker-wrapper functionality, removes `clang/test/Tooling/clang-linker-wrapper-spirv-elf.cpp`, merged without any test.
https://github.com/llvm/llvm-project/pull/185425 - adds llvm-objdump functionality and tests for it, but also returns back clang-linker-wrapper test with this name: `clang/test/Tooling/clang-linker-wrapper-spirv.cpp`

So I still think this is a clang-linker-wrapper test.

YuriPlyakhin:
> > The spirv-tools feature/substitution setup is carried over from Tooling's lit.local.cfg into a new lit.local.cfg in clang-linker-wrapper/, since the parent clang/test/lit.cfg.py does not set up that feature. The dead spirv-tools block is removed from Tooling/lit.local.cfg now that no tests there depend on it.
> 
> Why do OffloadTools tests depend on spirv-tools? I expect only SPIR-V backend tests need those.

@bader , 
```
# executed command: clang-linker-wrapper --host-triple=x86_64-unknown-linux-gnu --linker-path=/usr/bin/ld openmp-spirv.cpp.tmp.host.o -o clang-linker-wrapper/Output/openmp-spirv.cpp.tmp

# | clang: error: cannot find SPIR-V Tools binary 'spirv-link', which is required for SPIR-V compilation. It can be obtained from your system package manager or from KhronosGroup/SPIRV-Tools on GitHub
# | /localdisk2/yplyakhi/llvm-project/build/bin/clang-linker-wrapper: error: 'clang' failed
```
this is my understanding:
`clang-linker-wrapper` runs `clang --target=spirv64-intel`, and the driver selects SPIRV::Linker for the link step, which resolves to `spirv-link`.

clang/test/OffloadTools/clang-linker-wrapper/openmp-spirv.cpp
YuriPlyakhin: > Please, use %clang_cc1 substitution.

https://github.com/llvm/llvm-project/pull/198434/changes/6dfac09cd7a31123cb146a6c371be1a7b175ad61

YuriPlyakhin: > These are needed to create an input file for the clang-linker-wrapper tool we are testing. @sarnex, @adurang, do you have any thoughts on optimizing these four lines?

@bader,
This PR already solves 2 problems:
1. moves offload tools tests to dedicated folder
2. getting rid of requirement to have libomp and libomptarget in environment.

This PR doesn't add any new requirements or tools.

I suggest merging it and leave any further improvements for the next PRs.

bader:
> > > The spirv-tools feature/substitution setup is carried over from Tooling's lit.local.cfg into a new lit.local.cfg in clang-linker-wrapper/, since the parent clang/test/lit.cfg.py does not set up that feature. The dead spirv-tools block is removed from Tooling/lit.local.cfg now that no tests there depend on it.
> > 
> > 
> > Why do OffloadTools tests depend on spirv-tools? I expect only SPIR-V backend tests need those.
> 
> @bader ,
> 
> ```
> # executed command: clang-linker-wrapper --host-triple=x86_64-unknown-linux-gnu --linker-path=/usr/bin/ld openmp-spirv.cpp.tmp.host.o -o clang-linker-wrapper/Output/openmp-spirv.cpp.tmp
> 
> # | clang: error: cannot find SPIR-V Tools binary 'spirv-link', which is required for SPIR-V compilation. It can be obtained from your system package manager or from KhronosGroup/SPIRV-Tools on GitHub
> # | /localdisk2/yplyakhi/llvm-project/build/bin/clang-linker-wrapper: error: 'clang' failed
> ```
> 
> this is my understanding: `clang-linker-wrapper` runs `clang --target=spirv64-intel`, and the driver selects SPIRV::Linker for the link step, which resolves to `spirv-link`.

Shouldn't we run the test in "dry-run" mode to focus on testing `clang-linker-wrapper` functionality?

YuriPlyakhin:
> > > > The spirv-tools feature/substitution setup is carried over from Tooling's lit.local.cfg into a new lit.local.cfg in clang-linker-wrapper/, since the parent clang/test/lit.cfg.py does not set up that feature. The dead spirv-tools block is removed from Tooling/lit.local.cfg now that no tests there depend on it.
> > > 
> > > 
> > > Why do OffloadTools tests depend on spirv-tools? I expect only SPIR-V backend tests need those.
> > 
> > 
> > @bader ,
> > ```
> > # executed command: clang-linker-wrapper --host-triple=x86_64-unknown-linux-gnu --linker-path=/usr/bin/ld openmp-spirv.cpp.tmp.host.o -o clang-linker-wrapper/Output/openmp-spirv.cpp.tmp
> > 
> > # .---command stderr------------
> > # | clang: error: cannot find SPIR-V Tools binary 'spirv-link', which is required for SPIR-V compilation. It can be obtained from your system package manager or from KhronosGroup/SPIRV-Tools on GitHub
> > # | /localdisk2/yplyakhi/llvm-project/build/bin/clang-linker-wrapper: error: 'clang' failed
> > # `-----------------------------
> > # error: command failed with exit status: 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this is my understanding: `clang-linker-wrapper` runs `clang --target=spirv64-intel`, and the driver selects SPIRV::Linker for the link step, which resolves to `spirv-link`.
> 
> Shouldn't we run the test in "dry-run" mode to focus on testing `clang-linker-wrapper` functionality?

Then it will not produce the binary that we explore with llvm-objdump.

bader (approved):
LGTM.
Let's improve clang-linker-wrapper tests separately.

PR #198435, opened 2026-05-19
[mlir][IR] Add SymbolUserTypeInterface

jaredhoberock:
This change adds SymbolUserTypeInterface, analogous to SymbolUserAttrInterface, and extends SymbolTable verification to check participating types.

Verification visits the types owned by an operation, (operand/result types, block argument types, attribute-contained types, nested type parameters) and checks each distinct type at most once, interrupting on the first failure. Deduplication spans the whole symbol-table walk, so a type used by multiple operations is checked only once.

Attribute (SymbolUserAttrInterface) verification also gets deduped the same way.

Assisted-by: Codex (OpenAI)
Assisted-by: Claude (Anthropic)

### DIFF
--- a/mlir/lib/IR/SymbolTable.cpp
+++ b/mlir/lib/IR/SymbolTable.cpp
@@ -476,6 +476,51 @@ raw_ostream &mlir::operator<<(raw_ostream &os,
 // SymbolTable Trait Types
 //===----------------------------------------------------------------------===//
 
+/// Verify the symbol uses held by the types owned by `op`: its operand,
+/// result, and block-argument types, and any types nested within its
+/// attributes. `op` is the anchor used for symbol lookups. `verifiedTypes`
+/// records the types already verified within the current symbol table so that
+/// each type, which may be uniqued and shared across many positions or
+/// operations, is verified at most once. Verification fails fast on the first
+/// invalid symbol use.
+static LogicalResult verifyOpTypeSymbolUses(Operation *op,
+                                            SymbolTableCollection &symbolTable,
+                                            SetVector<Type> &verifiedTypes) {
+  // Walk `type` and any nested type parameters reachable from it, verifying
+  // each not-yet-seen type and interrupting on the first failure.
+  auto verify = [&](Type type) {
+    return type.walk<WalkOrder::PreOrder>([&](Type nestedType) {
+      if (!verifiedTypes.insert(nestedType))
+        return WalkResult::advance();
+      if (auto user = dyn_cast<SymbolUserTypeInterface>(nestedType))
+        if (failed(user.verifySymbolUses(op, symbolTable)))
+          return WalkResult::interrupt();
+      return WalkResult::advance();
+    });
+  };
+
+  for (Type type : op->getOperandTypes())
+    if (verify(type).wasInterrupted())
+      return failure();
+  for (Type type : op->getResultTypes())
+    if (verify(type).wasInterrupted())
+      return failure();
+  for (Region &region : op->getRegions())
+    for (Block &block : region)
+      for (BlockArgument argument : block.getArguments())
+        if (verify(argument.getType()).wasInterrupted())
+          return failure();
+
+  // Verify types nested within the operation's attributes.
+  WalkResult attrResult =
+      op->getAttrDictionary().walk<WalkOrder::PreOrder>([&](Type type) {
+        if (verify(type).wasInterrupted())
+          return WalkResult::interrupt();
+        return WalkResult::advance();
+      });
+  return failure(attrResult.wasInterrupted());
+}
+
 LogicalResult detail::verifySymbolTable(Operation *op) {
   if (op->getNumRegions() != 1)
     return op->emitOpError()
@@ -506,16 +551,27 @@ LogicalResult detail::verifySymbolTable(Operation *op) {
 
   // Verify any nested symbol user operations.
   SymbolTableCollection symbolTable;
+  // walkSymbolTable does not descend into nested symbol tables, so every
+  // operation visited here shares the same nearest symbol table. A uniqued
+  // attribute or type therefore resolves its symbol uses identically
+  // regardless of which operation anchors the lookup, so each is verified at
+  // most once across the whole scope.
+  SetVector<Attribute> verifiedAttrs;
+  SetVector<Type> verifiedTypes;
   auto verifySymbolUserFn = [&](Operation *op) -> std::optional<WalkResult> {
     if (SymbolUserOpInterface user = dyn_cast<SymbolUserOpInterface>(op))
       if (failed(user.verifySymbolUses(symbolTable)))
         return WalkResult::interrupt();
     for (auto &attr : op->getDiscardableAttrs()) {
       if (auto user = dyn_cast<SymbolUserAttrInterface>(attr.getValue())) {
+        if (!verifiedAttrs.insert(attr.getValue()))
+          continue;
         if (failed(user.verifySymbolUses(op, symbolTable)))
           return WalkResult::interrupt();
       }
     }
+    if (failed(verifyOpTypeSymbolUses(op, symbolTable, verifiedTypes)))
+      return WalkResult::interrupt();
     return WalkResult::advance();
   };
 
@@ -1137,3 +1193,4 @@ ParseResult impl::parseOptionalVisibilityKeyword(OpAsmParser &parser,
 /// Include the generated symbol interfaces.
 #include "mlir/IR/SymbolInterfaces.cpp.inc"
 #include "mlir/IR/SymbolInterfacesAttrInterface.cpp.inc"
+#include "mlir/IR/SymbolInterfacesTypeInterface.cpp.inc"
--- a/mlir/test/lib/Dialect/Test/TestTypes.cpp
+++ b/mlir/test/lib/Dialect/Test/TestTypes.cpp
@@ -333,6 +333,19 @@ uint64_t TestTypeWithLayoutType::extractKind(DataLayoutEntryListRef params,
 }
 
 //===----------------------------------------------------------------------===//
+// TestSymbolUserType
+//===----------------------------------------------------------------------===//
+
+LogicalResult
+TestSymbolUserType::verifySymbolUses(Operation *op,
+                                     SymbolTableCollection &symbolTable) const {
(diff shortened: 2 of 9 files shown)


mlir/lib/IR/SymbolTable.cpp
```diff
@@ -476,6 +476,58 @@ raw_ostream &mlir::operator<<(raw_ostream &os,
...
+/// Collect `type` and any nested type parameters reachable from it.
+static void collectSymbolUseTypes(Type type, SetVector<Type> &types) {
+  type.walk<WalkOrder::PreOrder>([&](Type nestedType) {
+    types.insert(nestedType);
+    return WalkResult::advance();
+  });
+}
+
+/// Collect types directly owned by `op`, including nested type parameters.
+static void collectSymbolUseTypesInOpTypes(Operation *op,
```
jpienaar: Why not just verify as part of walking? E.g., use a SetVector still, but created in line ~560 and then you just early exit when one (not yet seen) fails rather than collecting all and then checking. This could be done similarly for Attributes there too then.

mlir/lib/IR/SymbolTable.cpp
```diff
@@ -476,6 +476,58 @@ raw_ostream &mlir::operator<<(raw_ostream &os,
...
+  for (Type type : op->getOperandTypes())
+    collectSymbolUseTypes(type, types);
+  for (Type type : op->getResultTypes())
+    collectSymbolUseTypes(type, types);
+  for (Region &region : op->getRegions()) {
+    for (Block &block : region) {
+      for (BlockArgument argument : block.getArguments())
+        collectSymbolUseTypes(argument.getType(), types);
+    }
+  }
```
jpienaar: Not on you, but we should perhaps have this as a utility method somewhere (like an AttrOrTypeInOpWalker thing) as I've seen this a few times now.

mlir/test/lib/Dialect/Test/TestTypes.cpp
```diff
@@ -332,6 +332,20 @@ uint64_t TestTypeWithLayoutType::extractKind(DataLayoutEntryListRef params,
...
+//===----------------------------------------------------------------------===//
+// TestSymbolUserType
+//===----------------------------------------------------------------------===//
+
+LogicalResult
+TestSymbolUserType::verifySymbolUses(Operation *op,
+                                     SymbolTableCollection &symbolTable) const {
+  if (!symbolTable.lookupNearestSymbolFrom<SymbolOpInterface>(op, getSymbol()))
+    return op->emitOpError()
+           << "TestSymbolUserType::verifySymbolUses: '" << getSymbol()
```
jpienaar: s/"TestSymbolUserType::verifySymbolUses: '"// to be consistent with rest here (and LLVM/MLIR error emission conventions)

jpienaar (commented):
I think it makes sense in general yes.

mlir/test/lib/Dialect/Test/TestTypes.cpp
jaredhoberock: Fixed

mlir/lib/IR/SymbolTable.cpp
jaredhoberock: Fixed

jaredhoberock:
@jpienaar Thanks for the feedback! Latest changes verify types during the walk and interrupts on first failure. Attribute verification short-circuits similarly. Also made the diagnostic text match MLIR style.

jaredhoberock:
Ping @jpienaar - the latest revision addresses your feedback. Ready for another look.

jpienaar (approved):
Apologies, thought I had approved

jaredhoberock:
@joker-eph @jpienaar has approved. Would you please merge it for me?

jaredhoberock:
@joker-eph Thank you! Is this PR too late for the LLVM 23 release?

seldridge:
This is something that we've wanted on the CIRCT side for quite some time and have gone through heroics to avoid (either use of symbols in types and then verification handled separately or other mechanisms for tracking outside of types like additional operands or symbols on operations).

I had always justified this from my recollection of the MLIR rationale which was pretty staunchly on the side of (interpolating) "types exist outside the IR": https://mlir.llvm.org/docs/Rationale/Rationale/#symbols-and-types

Could a follow-on update the rationale be updated with these new directions?

jaredhoberock:
@seldridge Before anyone updates the rationale with a follow-on, I'd like to understand that part of the rationale document, because I'm not sure I understand it as written.

The "Symbols and types" section of the rationale says "current MLIR disallows use of symbols in types." But reading the section closely, it seems to use "symbol" to mean a reference to an SSA value. An example used in that section presents a type spelling like `memref<8x%Nxf32>` and explains that this is a rejected alternative design. It's definitely not possible to refer to an SSA value `%N` inside of a type like that.

But my more familiar understanding of "symbol" is something like a `SymbolRefAttr`. Nothing in MLIR prevents a symbol reference of this kind within a type, and this `SymbolUserTypeInterface` addition is simply being honest about what was already possible.

Given that reading, is the request to bring that section of the rationale up to date such that "symbol" stops implying a prohibition on things like `SymbolRefAttr` or something more?

seldridge:
@jaredhoberock: Perhaps I was misinterpreting that, but I had always thought that rationale as also applying to `SymbolRefAttr`.  Or: I had inferred that the intent was that types were immortal and were supposed to outlive any references to IR.  See language like: https://github.com/llvm/llvm-project/blob/91ba38d018aaf672fad6c7a41da8d725f659417f/mlir/include/mlir/IR/MLIRContext.h#L41-L43  Though, given that types and attributes are roughly equivalent (from a view of where they are stored), the addition of `SymbolUserAttrInterface`, and the fact that this has been repeatedly asked for, I'm very much in favor of this (as long as the performance impacts can be sorted out).

> Nothing in MLIR prevents a symbol reference of this kind within a type, and this SymbolUserTypeInterface addition is simply being honest about what was already possible.

This has been a repeated point of tension and has been brought up before from the CIRCT side:

- https://discourse.llvm.org/t/contextual-type-verification/4525/3
- https://discourse.llvm.org/t/symbol-user-interface-for-attributes-types/69935

Though the fact that the interface added in this patch _didn't_ exist before caused us to assume that this was an intentional limitation.  If a developer designed an IR which needed this, it was on them to deal with the verification problems this would entail which is how we've operated in the past.

> Given that reading, is the request to bring that section of the rationale up to date such that "symbol" stops implying a prohibition on things like SymbolRefAttr or something more?

Yeah, being very clear that that example is talking about SSA values would be useful.  Additionally, adding a section that details the rationale for why types and attributes were changed to allow for them to refer to symbols would be beneficial for our future selves.

jaredhoberock:
I can't say why this PR was accepted. I noticed the existence of `SymbolUserAttrInterface` and thought `SymbolUserTypeInterface` also made sense reasoning that attributes and types have roughly the same status as you observed. This interface replaced a hacky ad hoc and out-of-band verification system in my own work.

I appreciate the desire to keep the type system pure. But sometimes types really want to refer to declarations. See the awkwardness of `!llvm.struct`, for example. An opt-in interface is a good trade-off.

There are real performance implications (#212160) for verifiers, but I think they are surmountable. See #212354 for a proof of concept. If you see a better way of dealing with it, please let me know.

seldridge:
> I appreciate the desire to keep the type system pure. But sometimes types really want to refer to declarations. See the awkwardness of !llvm.struct, for example. An opt-in interface is a good trade-off.

Yeah, I totally agree that actually having _some_ blessed way of doing this is necessary and it's been bad that everyone has had to invent their own approach here.  We're looking forward to a resolution on the performance side as we have some massive IR (hundreds of GiB) and the full 2x hit this introduced is affecting our CI flows.  The approaches of the proposed patches look good.

And with all that said, we are still planning to use this new interface!  This will clean up some of our own out-of-band verifiers.  😄

@dtzSiFive is helping me keep an eye on this from our side and may have some opinions on the proposed fixes.

jpienaar:
I am surprised the impact was this high (given we were already doing attribute verification for this and both during the more expensive symbol table verification). Especially with uniquing ... I'm curious how many unique types are in these modules.

>> Given that reading, is the request to bring that section of the rationale up to date such that "symbol" stops implying a >> prohibition on things like SymbolRefAttr or something more?

> Yeah, being very clear that that example is talking about SSA values would be useful. Additionally, adding a section
> that details the rationale for why types and attributes were changed to allow for them to refer to symbols would be
> beneficial for our future selves.

Indeed the limitation was focussed on SSA values, but more on practical consideration as in there not needing to be some continuous propagation: e.g., to evaluate two types for equality, it being a pointer check which result doesn't depend on if canonicalize has been called kind. We should revise the doc.

dtzSiFive:
This is timely, as I basically ran into the engineering challenges of almost exactly this reworking CIRCT's Inliner over on https://github.com/llvm/circt/pull/10880 .  The problem there is maintaining (symbol) references to global hierarchical paths from throughout the entire circuit (collection of modules instantiating each other) and rewriting the paths through this folded graph as appropriate for the requested user-directed inlining operations.  So "walk everything and its attributes to find the users" (basically an intrusive SymbolTable find-users walk) was done in an early iteration which I quickly found to be prohibitively expensive ; it's been extracted into a follow-up PR I'll submit shortly now the base has landed.  The solutions I came up with mostly mirror what's being thrown around here, minus the insight re:possibly caching this per-type on type creation.

Anyway:

Benchmarked and investigated this last night using an open-source RocketConfig on 1.154 vs 1.155 of CIRCT (across adoption of this PR in our LLVM submodule).

Not that many unique types (~1.4k distinct types across ~782k operations), the costs are mostly per-op:

* Unique'ing due to `getAttrDictionary` on each operation; all CIRCT operations use Properties and so this is expensive and on a larger design increased MaxRSS by ~0.8GB.
* Each verify call creates a new `AttrTypeWalker`.  The walker has its own DenseMap (for sharing and recursive types).  Consider using a single walker across invocations.  This bites everyone the first time AFAICT (#212354  does this).
* We walk /through/ types already verified, where we could skip them entirely as all of their contents may be safely skipped for the same reason we can skip this at all:
```
      if (!verifiedTypes.insert(nestedType))
        return WalkResult::advance();
```

Re-using the walker might address the last point and avoid the need for the `verifiedTypes` set entirely anyway.

In the Inliner's case I can further improve things by limiting walk scope to the "blast-radius" of users of these paths; there's no analogue for that here.

I've been looking carefully at the two proposed fix PR's as well, more on that soon.

PR #152606, opened 2025-08-07
[clang][Sema] Fix the continue and break scope for while loops

ojhunt:
Make sure we don't push the break and continue scope for a while loop until after we have evaluated the condition.

PR #198436, opened 2026-05-19
[Clang] Disallow `break`/`continue` in loop conditions

Sirraide:
tl;dr: This makes e.g. `while (({ break; 1; })) {}` ill-formed.

GCC used to allow this a long time ago (< GCC 9 I believe), but eventually removed support for it; we originally allowed this both for GCC compatibility and because there was actual code in the wild using it (see Richard’s comment here for more background: https://github.com/llvm/llvm-project/pull/152606#issuecomment-3166130973).

Note that this _is_ still allowed inside another loop, e.g. this
```c++
for (;;) {
    while (({ break; true; })) {}
}
```
is well-formed; the `break` here will break out of the `for` loop.

Removing support for this gets rid of quite a bit of code and has a few more benefits:

1. Currently, GCC and Clang _disagree_ on the meaning of this construct in nested loops: in the code snippet above, Clang breaks out of the `while` loop, whereas GCC breaks out of the `for` loop; this patch changes Clang to align with GCC here. As a result, we can also remove code that emits a warning for such cases.

2. It frees up a bit in `ScopeFlags::Flags`, which is a good thing because I need one for expansion statements, and we’re out of bits. This is because we currently use a bit as a hack to disallow `continue` inside the declaration of the condition variable (because we’d `continue` to the condition before initialising the variable); this bit becomes obsolete with this patch because `continue` is now disallowed entirely within the condition (if there is no outer loop).

    Without this change, I’d have to refactor `Flags` to be a 64-bit integer, which would also entail updating every place where we use an `unsigned` to store scope flags.

There is another change that I needed to make here: we currently suppress `-Wcomma` (which warns about comma operators, because in most contexts, you probably didn’t mean to use one) in some places, including the third part of a C-style `for` loop. This is implemented by checking if we’re in a `BreakScope | ContinueScope`, but those scope flags are now only set when parsing the loop body. Instead, we now check for `ControlScope`. This requires us to also set that flag in C90 mode, but that seems to be harmless as the only use of `ControlScope` I was able to find is in C++ code paths, where that bit was already set anyway.

When we introduced named loops for C2y, we purposefully didn’t support labeled break/continue in the condition, so there’s nothing to be done there (we already have tests for this too).

### DIFF
--- a/clang/include/clang/Sema/Scope.h
+++ b/clang/include/clang/Sema/Scope.h
@@ -140,12 +140,8 @@ public:
     /// This is the scope of a C++ catch statement.
     CatchScope = 0x1000000,
 
-    /// This is the scope corresponding to a C++26 expansion statement.
-    ///
-    /// NOTE: While currently unused, this flag is reserved for the
-    /// implementation of expansion statements; do not remove it or
-    /// reuse it for anything else.
-    ExpansionStmtScope = 0x2000000,
+    /// This bit is currently unused.
+    Unused = 0x2000000,
 
     /// This is a scope of some OpenMP directive with
     /// order clause which specifies concurrent
--- a/clang/include/clang/Sema/Scope.h
+++ b/clang/include/clang/Sema/Scope.h
@@ -140,7 +140,7 @@ public:
     /// This is the scope of a C++ catch statement.
     CatchScope = 0x1000000,
 
-    /// This is the scope corresponding of a C++26 expansion statement.
+    /// This is the scope corresponding to a C++26 expansion statement.
     ///
     /// NOTE: While currently unused, this flag is reserved for the
     /// implementation of expansion statements; do not remove it or
--- a/clang/include/clang/Sema/Scope.h
+++ b/clang/include/clang/Sema/Scope.h
@@ -140,10 +140,12 @@ public:
     /// This is the scope of a C++ catch statement.
     CatchScope = 0x1000000,
 
-    /// This is a scope in which a condition variable is currently being
-    /// parsed. If such a scope is a ContinueScope, it's invalid to jump to the
-    /// continue block from here.
-    ConditionVarScope = 0x2000000,
+    /// This is the scope corresponding of a C++26 expansion statement.
+    ///
+    /// NOTE: While currently unused, this flag is reserved for the
+    /// implementation of expansion statements; do not remove it or
+    /// reuse it for anything else.
+    ExpansionStmtScope = 0x2000000,
 
     /// This is a scope of some OpenMP directive with
     /// order clause which specifies concurrent
@@ -274,11 +276,6 @@ public:
 
   /// Get the label that precedes this scope.
   LabelDecl *getPrecedingLabel() const { return PrecedingLabel; }
-  void setPrecedingLabel(LabelDecl *LD) {
-    assert((Flags & BreakScope || Flags & ContinueScope) &&
-           "not a loop or switch");
-    PrecedingLabel = LD;
-  }
 
   /// isBlockScope - Return true if this scope correspond to a closure.
   bool isBlockScope() const { return Flags & BlockScope; }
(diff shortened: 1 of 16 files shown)


Sirraide:
Side note: I’ve been planning to disallow this for a while, and iirc I discussed this w/ Aaron back when I implemented named loops, but I’ve only now been reminded of this because I really need a free bit in the scope flags for expansion statements...

Sirraide:
I still need to write some codegen tests for this; I should be able to get to that soon

shafik (commented):
Do we have a feel for if anyone still relies in this behavior even though gcc does not support it or maybe worse rely on building w/ old versions? It does not mean we shouldn't do it just wondering.

clang/include/clang/Sema/Scope.h
```diff
@@ -140,10 +140,12 @@ class Scope {
...
-    /// This is a scope in which a condition variable is currently being
-    /// parsed. If such a scope is a ContinueScope, it's invalid to jump to the
-    /// continue block from here.
-    ConditionVarScope = 0x2000000,
+    /// This is the scope corresponding to a C++26 expansion statement.
+    ///
+    /// NOTE: While currently unused, this flag is reserved for the
+    /// implementation of expansion statements; do not remove it or
+    /// reuse it for anything else.
+    ExpansionStmtScope = 0x2000000,
```
ojhunt: wow, so greedy. :D

clang/include/clang/Sema/Scope.h
```diff
@@ -636,6 +622,16 @@ class Scope {
...
 
+  /// Mark that we're entering the body of a loop (for, while, do).
+  void EnterLoopBody(LabelDecl *PrecedingLabel);
+
+  /// Mark that we're entering the body of a switch statement.
+  void EnterSwitchBody(LabelDecl *PrecedingLabel);
+
+  /// Mark that we're leaving the body of a loop; this is only needed for do
+  /// loops where the condition follows the loop body.
+  void LeaveLoopBody();
```
ojhunt, line 629: I don't really like this name but I can't really think of anything better myself (things like `LeaveWhileDoLoopBody()` seem super clunky, but may they would be better? )

ojhunt (approved):
I really wanted to just approve this because of all the red in the diff, but that's immature so I have gone through it properly and it looks good to me. The only real feedback is that I don't like the name `LeaveLoopBody` but I can't think of an alternative that is actually better.

Sirraide:
> Do we have a feel for if anyone still relies in this behavior even though gcc does not support it or maybe worse rely on building w/ old versions? It does not mean we shouldn't do it just wondering.

I mean, I honestly don’t think there’s much of a use case for break/continue outside of a loop body. The only one I’m aware of is that old Qt FOREACH macro, but even that one doesn’t use this hack anymore. I would be rather surprised if this ended up breaking any actual code

clang/include/clang/Sema/Scope.h
Sirraide: I mean, it does the opposite of `EnterLoopBody` so I think the name is fine ;Þ

Sirraide: I mean, we could call it `DisallowBreakContinue` because that’s what it actually does if you prefer that

clang/include/clang/Sema/Scope.h
cor3ntin: I do not think it makes sense to do this in this PR.
I do like the motivation though. "let me remove a weird behavior so i can earn a free bit, as a treat"

cor3ntin (approved):

clang/include/clang/Sema/Scope.h
Sirraide: I mean, I can rename it to `Unused` for now if you prefer

Sirraide: Done

clang/include/clang/Sema/Scope.h
ojhunt: Nono, like I said - I don't particularly like it, but I can't think of anything that is actually _better_ :D

clang/include/clang/Sema/Scope.h
ojhunt: `ThisBitIntentionallyLeftBlank` ? :D

PR #198437, opened 2026-05-19
[AMDGPU] Remove unsupported feature by gfx12-5-generic target

Shoreshen:


llvm/test/CodeGen/AMDGPU/gfx12-5-generic-no-xnack.ll
```diff
@@ -0,0 +1,9 @@
+; RUN: not llc -mtriple=amdgcn -mcpu=gfx12-5-generic -mattr=+xnack -o - %s 2>&1 | FileCheck --check-prefix=GFX125-PLUS-XNACK %s
```
shiltian: https://llvm.org/docs/AMDGPU/DeveloperGuideline.html#use-filetype-null-when-output-is-irrelevant

shiltian (commented):
LGTM but I'll leave the stamp to others, since I'm not very familiar with this xnack settings.

llvm/lib/Target/AMDGPU/AMDGPU.td
```diff
@@ -212,6 +212,10 @@ def FeatureSupportsXNACK : SubtargetFeature<"xnack-support",
   "Hardware supports XNACK"
 >;
 
+defm XNACKAnyOnly : AMDGPUSubtargetFeature<"xnack-any-only",
+  "Target only support XnackSetting to be `any`"
```
shiltian:
```suggestion
  "Target only support XnackSetting to be any"
```

No need to format here

llvm/lib/Target/AMDGPU/GCNSubtarget.cpp
```diff
@@ -174,6 +174,10 @@ void GCNSubtarget::checkSubtargetFeatures(const Function &F) const {
     Ctx.diagnose(DiagnosticInfoUnsupported(
         F, "must specify exactly one of wavefrontsize32 and wavefrontsize64"));
   }
+  if (hasFeature(AMDGPU::FeatureXNACKAnyOnly) && TargetID.isXnackOnOrOff()) {
+    Ctx.diagnose(DiagnosticInfoUnsupported(
+        F, "target only supports xnack 'Any'; `+/-xnack` is not allowed"));
```
shiltian:
```suggestion
        F, "target only supports xnack 'Any'; '+/-xnack is not allowed"));
```

Markdown syntax is not conventional in error message.

llvm/lib/Target/AMDGPU/AMDGPU.td
```diff
@@ -2181,9 +2185,11 @@ def FeatureISAVersion12_Generic: FeatureSet<
 
 def FeatureISAVersion12_5_Generic: FeatureSet<
   !listconcat(FeatureISAVersion12_50_Common.Features,
-    [FeatureAddressableLocalMemorySize327680,
-     FeatureSetregVGPRMSBFixup,
-     FeatureRequiresCOV6])>;
+  [FeatureAddressableLocalMemorySize327680,
+   FeatureSetregVGPRMSBFixup,
+   FeatureRealTrue16Insts,
```
chinmaydd: > FeatureRealTrue16Insts

This seems like a new addition to the generic feature set?

shiltian:
The title needs to be more specific though.

llvm/lib/Target/AMDGPU/AMDGPU.td
shiltian: Yeah, this needs to be added in a separate PR with some codegen tests.

llvm/lib/Target/AMDGPU/AMDGPU.td
```diff
@@ -212,6 +212,10 @@ def FeatureSupportsXNACK : SubtargetFeature<"xnack-support",
   "Hardware supports XNACK"
 >;
 
+defm XNACKAnyOnly : AMDGPUSubtargetFeature<"xnack-any-only",
+  "Target only support XnackSetting to be any"
```
chinmaydd: Nit:

```suggestion
  "Target only supports XnackSetting to be Any"
```

chinmaydd (approved):

chinmaydd:
```
2026-05-20T16:18:18.3827545Z [{'explained': True, 'name': 'MLIR.Target/LLVMIR/openmp-cli-tile02.mlir', 'reason': 'This test is already failing at the base commit.'}, {'explained': False, 'name': 'MLIR.Target/LLVMIR/openmp-llvm.mlir', 'reason': None}]
```

I'd like to believe this failure is unrelated

Shoreshen:
> ```
> 2026-05-20T16:18:18.3827545Z [{'explained': True, 'name': 'MLIR.Target/LLVMIR/openmp-cli-tile02.mlir', 'reason': 'This test is already failing at the base commit.'}, {'explained': False, 'name': 'MLIR.Target/LLVMIR/openmp-llvm.mlir', 'reason': None}]
> ```
> 
> I'd like to believe this failure is unrelated

The test `openmp-llvm.mlir` failed on assertion Assertion `Extractor->isEligible() && "Expected OpenMP outlining to be possible!", which I believe is due to this issue https://github.com/llvm/llvm-project/issues/198775.

`openmp-cli-tile02.mlir` I currently cannot reproduce on my local:
image

chinmaydd:
Right, I think you should go ahead with the merge Shore.

Shoreshen:
> Right, I think you should go ahead with the merge Shore.

Thanks, I will merge after this CI done~~:)

arsenm:
I'm confused by this change. The current xnack handling for gfx125* is a mess. AMDGPUTargetParser has a FEATURE_XNACK_ALWAYS, set on 1250 and 1251 but not 12-5-generic. xnack-any-only is broken, in that it's an anti-feature. It remove functionality.

What is the rationale for making this an error? I'm inclined to just ignore the xnack setting rather than forcing frontends to do more work to know about the subtarget. If we really need to error on this, this should be inverted so every other target has a feature that indicates it has a settable xnack mode

PR #198438, opened 2026-05-19
[GlobalISel] Add `isKnownNeverZero` to `GISelValueTracking`

lialan:
## Summary of changes

* This patch adds `GISelValueTracking::isKnownNeverZero`, modeled after the SelectionDAG query, and uses it in the GlobalISel count-zero combiner.

* The root combines enabled by this patch are:
  * `G_CTLZ x -> G_CTLZ_ZERO_POISON x`
  * `G_CTTZ x -> G_CTTZ_ZERO_POISON x`
* The combine only fires when the source is known nonzero and the zero-poison opcode is legal, or when the combiner is running before legalization.
* For the initial proof set, this patch handles `G_OR`, `G_SELECT`, and `G_SHL`. These are not new root combines; they are source-expression cases used by `isKnownNeverZero` to prove that the input to `G_CTLZ` / `G_CTTZ` is nonzero. Other structural cases can be added in follow-up patches. For other opcodes, the query falls back to existing KnownBits support.
* Tests cover the `isKnownNeverZero` query, the AArch64 MIR combine cases, and the AMDGPU end-to-end lowering behavior.

## Example

Pattern `ctlz(or(x, 1))` on AMDGPU:

Before:
```
    v_or_b32_e32   v0, 1, v0
    v_ffbh_u32_e32 v0, v0
    v_min_u32_e32  v0, 32, v0   ; bitwidth clamp
    s_setpc_b64    s[30:31]
```

After:
```
    v_or_b32_e32   v0, 1, v0
    v_ffbh_u32_e32 v0, v0
    s_setpc_b64    s[30:31]
```


llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -8816,3 +8816,35 @@ bool CombinerHelper::matchAVG(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return TargetOpcode::G_CTLZ_ZERO_POISON;
+  case TargetOpcode::G_CTTZ:
+    return TargetOpcode::G_CTTZ_ZERO_POISON;
+  default:
+    return 0;
+  }
+}
+
+bool CombinerHelper::matchCountZeroToZeroPoison(MachineInstr &MI) const {
+  if (!VT)
```
arsenm, line 8852: We really ought to have a way to declare which combines an analysis require in tablegen

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -8816,3 +8816,35 @@ bool CombinerHelper::matchAVG(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return false;
+
+  unsigned ZPOpc = getCountZeroPoisonOpcode(MI.getOpcode());
+  if (!ZPOpc)
+    return false;
```
arsenm: This is unreachable

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
```diff
@@ -105,6 +105,202 @@ bool GISelValueTracking::signBitIsZero(Register R) {
...
+bool GISelValueTracking::isKnownNeverZero(Register R, const APInt &DemandedElts,
+                                          unsigned Depth) {
+  if (Depth >= getMaxDepth())
+    return false;
+
+  MachineInstr *MI = MRI.getVRegDef(R);
+  if (!MI)
+    return false;
+
+  auto Recurse = [&](Register Reg, const APInt &Elts) {
```
arsenm: I'd prefer to just inline these trivial functions like every other value tracking analysis

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
```diff
@@ -105,6 +105,202 @@ bool GISelValueTracking::signBitIsZero(Register R) {
...
+}
+
+bool GISelValueTracking::isKnownNeverZero(Register R, const APInt &DemandedElts,
+                                          unsigned Depth) {
+  if (Depth >= getMaxDepth())
+    return false;
+
+  MachineInstr *MI = MRI.getVRegDef(R);
+  if (!MI)
+    return false;
```
arsenm: This cannot fail

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
lialan: removed

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
lialan: removed

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
lialan: inlined

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
lialan: That is probably better. I will look into that.

llvm/include/llvm/CodeGen/GlobalISel/GISelValueTracking.h
```diff
@@ -92,6 +92,12 @@ class LLVM_ABI GISelValueTracking : public GISelChangeObserver {
   /// predicate to simplify operations downstream.
   bool signBitIsZero(Register Op);
 
+  /// Return true if the value defined by \p R is provably never zero. This is
+  /// the GlobalISel analog of \c SelectionDAG::isKnownNeverZero.
```
qcolombet: Drop the SelectionDAG reference and make sure the comment stands on its own.

llvm/include/llvm/Target/GlobalISel/Combine.td
```diff
@@ -2252,6 +2252,22 @@ def ctls_combines : GICombineGroup<[
   ctlz_zero_poison_to_ctls,
 ]>;
 
+// Switch G_CTLZ / G_CTTZ to *_ZERO_POISON when the input is provably nonzero
+// and the target supports the variant. Mirrors DAGCombiner.cpp:12394-12397 /
+// 12431-12433.
```
qcolombet: Listing .cpp lines that are likely to change is not helpful.
Just make sure the comment is self contained, and if you need more context, bring it here.

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -8816,3 +8816,32 @@ bool CombinerHelper::matchAVG(MachineInstr &MI, MachineRegisterInfo &MRI,
...
+    return TargetOpcode::G_CTTZ_ZERO_POISON;
+  default:
+    llvm_unreachable("Expected count-zero opcode");
+  }
+}
+
+bool CombinerHelper::matchCountZeroToZeroPoison(MachineInstr &MI) const {
+  if (!VT)
+    return false;
+
```
qcolombet, line 8854: Could we assert that MI is CTLZ or CTTZ?

llvm/include/llvm/CodeGen/GlobalISel/GISelValueTracking.h
```diff
@@ -92,6 +92,12 @@ class LLVM_ABI GISelValueTracking : public GISelChangeObserver {
...
+  bool isKnownNeverZero(Register R, unsigned Depth = 0);
+  bool isKnownNeverZero(Register R, const APInt &DemandedElts,
```
qcolombet, line 102: Make sure to document the `DemandedElts` argument.

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
```diff
@@ -105,6 +105,215 @@ bool GISelValueTracking::signBitIsZero(Register R) {
...
+    break;
+
+  case TargetOpcode::G_CONSTANT:
+    return !MI.getOperand(1).getCImm()->isZero();
+
+  case TargetOpcode::G_BUILD_VECTOR: {
+    for (const auto &[I, MO] : enumerate(drop_begin(MI.operands()))) {
+      if (!DemandedElts[I])
+        continue;
+      if (!isKnownNeverZero(MO.getReg(), APInt(1, 1), Depth + 1))
```
qcolombet:
```suggestion
      if (!isKnownNeverZero(MO.getReg(), /*DemandedElts=*/APInt(1, 1), Depth + 1))
```

qcolombet: Maybe use a `ScalarDemandedElts` variable and store `APInt(1,1)` in it to reuse in all the places.

qcolombet (commented):
I think we could start with fewer opcodes to make the patch easier to digest.
I think it is going in the right direction, but I didn't look at all the opcodes.

jayfoad:
> * The implementation handles a starter set of opcodes directly `G_CONSTANT, G_BUILD_VECTOR, G_SPLAT_VECTOR, G_EXTRACT_VECTOR_ELT, G_SHUFFLE_VECTOR, G_OR, G_SELECT, G_SHL`.
> * and falls back to `getKnownBits(R, DemandedElts, Depth).isNonZero()` for everything else.
> * `isKnownNeverZero` is stronger than `getKnownBits`.

Why would handling `G_CONSTANT` directly be any stronger than letting `getKnownBits` handle it?

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
lialan: good suggestion. added.

lialan:
> Why would handling `G_CONSTANT` directly be any stronger than letting `getKnownBits` handle it?

After reviewing, it is indeed redundant. Removed.  I also narrowed the initial direct opcode set to `G_OR`, `G_SELECT`, and  `G_SHL`, and updated the PR description accordingly.

lialan:
> I think we could start with fewer opcodes to make the patch easier to digest. I think it is going in the right direction, but I didn't look at all the opcodes.

I updated the PR to only include a few opcodes to begin with, so this will make things easier to go with.

llvm/include/llvm/CodeGen/GlobalISel/GISelValueTracking.h
```diff
@@ -92,6 +92,15 @@ class LLVM_ABI GISelValueTracking : public GISelChangeObserver {
   /// predicate to simplify operations downstream.
   bool signBitIsZero(Register Op);
 
+  /// Return true if the value defined by \p R is provably never zero.
+  ///
+  /// \p DemandedElts selects the vector elements that must be proven nonzero.
+  /// For scalar values this is a one-bit mask. The overload without
+  /// \p DemandedElts demands every fixed-vector element, or the scalar value.
```
qcolombet: ``
```suggestion
  /// \p DemandedElts demands every fixed-vector element, or the scalar value to be non-zero.
```

qcolombet (approved):

PR #196299, opened 2026-05-07
[DebugInfo] Enforce DISubprogram type as required in LLParser

JiangNingHX:
Add a parse-time check so that LLParser treats the `type:` field of                                             
  `DISubprogram` as required and non-null. Attempting to assemble IR with
  a missing or null `type:` now produces a clear parse-time error
  ("missing required field 'type'" or "'type' cannot be null") rather                                                                                                                                                               
  than relying solely on the verifier warning added in #194556.                                                                                                                                                                     
                                                                                                                                                                                                                                    
  Update existing hand-written LLVM IR tests to supply a minimal valid                                                                                                                                                              
  `DISubroutineType` where they are not intentionally testing malformed                                                                                                                                                             
  debug info.                                                                                                                                                                                                                       
                                                                                                                                                                                                                                    
  **Scope.** This patch covers the LLParser path only. Two entry points                                                                                                                                                             
  are intentionally deferred:                                                                                                                                                                                                       
                                                                                                                  
  - *MetadataLoader*: pre-LLVM-5 bitcode legitimately stores null                                                                                                                                                                   
    `DISubprogram.type`; auto-upgrading that field requires more careful
    backward-compatibility handling (see `llvm/test/Bitcode/dityperefs-3.8.ll`).                                                                                                                                                    
  - *BitcodeWriter*: IR written to bitcode has already passed the                                                                                                                                                                   
    verifier, which guarantees a non-null type; no additional check is                                                                                                                                                              
    needed in the writer.                                                                                                                                                                                                           
                                                                                                                                                                                                                                    
  Follow-up to #194556.

PR #198441, opened 2026-05-19
[llvm][DebugInfo] Require non-null debug metadata fields when writing bitcode

jiangning26:
Follow-up to #196299.

  This updates BitcodeWriter to use `getMetadataID()` for debug metadata fields
  that LLParser parses as required non-null `MDField` operands.

  These bitcode record fields still use the existing metadata-or-null encoding,
  so the writer preserves the on-disk encoding by writing the metadata ID plus
  one. This makes the writer assert the in-memory invariant without changing
  MetadataLoader's compatibility behavior for older bitcode.

  MetadataLoader is intentionally left permissive in this patch because older
  bitcode can still contain null operands for these fields.

### DIFF
--- a/llvm/lib/Bitcode/Writer/BitcodeWriter.cpp
+++ b/llvm/lib/Bitcode/Writer/BitcodeWriter.cpp
@@ -2227,7 +2227,11 @@ void ModuleBitcodeWriter::writeDISubprogram(const DISubprogram *N,
   Record.push_back(VE.getMetadataOrNullID(N->getRawLinkageName()));
   Record.push_back(VE.getMetadataOrNullID(N->getFile()));
   Record.push_back(N->getLine());
-  Record.push_back(VE.getMetadataOrNullID(N->getType()));
+  // This field uses the metadata-or-null encoding in bitcode: 0 means null,
+  // otherwise the stored value is the metadata ID plus 1. Use getMetadataID()
+  // to assert that DISubprogram::type is present, then translate back to the
+  // existing encoding.
+  Record.push_back(VE.getMetadataID(N->getType()) + 1);
   Record.push_back(N->getScopeLine());
   Record.push_back(VE.getMetadataOrNullID(N->getContainingType()));
   Record.push_back(N->getSPFlags());


JiangNingHX:
@Michael137 Could you please take a look when you have time?

This is a follow-up to #196299, tightening BitcodeWriter's handling of debug metadata fields that are required to be non-null.

llvm/lib/Bitcode/Writer/BitcodeWriter.cpp
```diff
@@ -2182,7 +2182,7 @@ void ModuleBitcodeWriter::writeDICompileUnit(const DICompileUnit *N,
   if (Lang.hasVersionedName())
     Record.back() ^= (uint64_t(1) << 63);
 
-  Record.push_back(VE.getMetadataOrNullID(N->getFile()));
+  Record.push_back(VE.getMetadataID(N->getFile()) + 1);
```
Michael137: Doesn't LLParser only mark the `type` field as required currently? Why are we changing all the other fields too?

jiangning26: You're right that #196299 specifically only made `DISubprogram::type` required.

  I generalized this patch because LLParser currently has a few other debug metadata `MDField`s that are also required and `AllowNull=false`, e.g. `DICompileUnit::file`, `DILexicalBlock{,File}::scope`, `DILocalVariable::scope`,
  and `DILabel::scope`. `DILocation::scope` is in the same category but BitcodeWriter already uses `getMetadataID()` there.

  For these changed bitcode records, MetadataLoader still decodes the fields with the existing metadata-or-null encoding, so this patch keeps the on-disk encoding unchanged by writing `getMetadataID(...) + 1`, while asserting the in-memory non-null invariant in the writer.

  That said, I agree the PR description made this sound like a narrow #196299 follow-up. I can either update the description to make the broader scope explicit, or split this down to only `DISubprogram::type` if you prefer.

jiangning26:
@Michael137 Hi,  I’m happy to either keep the broader scope here or split this down to just DISubprogram::type if you’d prefer. Let me know which direction you’d like, and I’ll adjust accordingly.

llvm/lib/Bitcode/Writer/BitcodeWriter.cpp
Michael137: Yea lets only change what was relevant to https://github.com/llvm/llvm-project/pull/196299

jiangning26:
@Michael137 I narrowed this to only `DISubprogram::type`, as requested. Could you please take another look?

jiangning26:
@Michael137 Could you take a look at this PR when you have a moment?

llvm/lib/Bitcode/Writer/BitcodeWriter.cpp
```diff
@@ -2227,7 +2227,11 @@ void ModuleBitcodeWriter::writeDISubprogram(const DISubprogram *N,
   Record.push_back(VE.getMetadataOrNullID(N->getRawLinkageName()));
   Record.push_back(VE.getMetadataOrNullID(N->getFile()));
   Record.push_back(N->getLine());
-  Record.push_back(VE.getMetadataOrNullID(N->getType()));
+  // This field uses the metadata-or-null encoding in bitcode: 0 means null,
+  // otherwise the stored value is the metadata ID plus 1. Use getMetadataID()
+  // to assert that DISubprogram::type is present, then translate back to the
+  // existing encoding.
+  Record.push_back(VE.getMetadataID(N->getType()) + 1);
```
Michael137, line 2234: Sorry for the delay on the reviews

I'm not sure I'm a fan of the `+ 1` here. Ideally we'd just do:
```
Record.push_back(VE.getMetadataID(N->getType()));
```
but then the MetadataLoader would have to be adjusted to not do the `- 1` which it currently does. But can't do that without breaking old bitcode (though this may  be one of those cases where we might be ok with it since it was never something that would've properly worked with LLVM).

I think I'd prefer just keeping this logic as-is since its consistent with all the other fields. We should just keep these kind of decisions in mind when adding new fields in the future

Michael137:
> @Michael137 Could you take a look at this PR when you have a moment?

Apologies, this kept falling off my radar. After some more reading of the bitcode infra I think we should close this out.

Though one follow-up from https://github.com/llvm/llvm-project/pull/196299 i'd still like to do is revert `4262045ba98d11c930fec0fe02584b072ace19bd` since it was added as a workaround that is no longer possible to hit.

PR #198444, opened 2026-05-19
[BOLT][Test] Fix PointerAuthCFIFixup.cpp build error when AArch64 not enabled

wenju-he:
When AArch64 target isn't in LLVM_TARGETS_TO_BUILD, AARCH64_AVAILABLE is not defined and there is build error:
PointerAuthCFIFixup.cpp:133:39: error: use of undeclared identifier 'AArch64'

### DIFF
--- a/bolt/unittests/Passes/CMakeLists.txt
+++ b/bolt/unittests/Passes/CMakeLists.txt
@@ -1,3 +1,15 @@
+set(LLVM_OPTIONAL_SOURCES
+  PointerAuthCFIFixup.cpp
+)
+
+set(PassTests_SOURCES
+  LivenessAnalysis.cpp
+)
+
+if(LLVM_TARGETS_TO_BUILD MATCHES "AArch64")
+  list(APPEND PassTests_SOURCES PointerAuthCFIFixup.cpp)
+endif()
+
 set(LLVM_LINK_COMPONENTS
   DebugInfoDWARF
   Object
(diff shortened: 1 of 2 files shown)


yavtuk:
@wenju-he hello, thanks for the fix, but I would prefer to exclude the entire file in CMakeLists.txt since PointerAuthCFIFixup.cpp only applies to the aarch64 platform.
something like that 

bolt/unittests/Passes/CMakeLists.txt
```
set(PassTests_SOURCES
  LivenessAnalysis.cpp
)

if(LLVM_TARGETS_TO_BUILD MATCHES "AArch64")
  list(APPEND PassTests_SOURCES PointerAuthCFIFixup.cpp)
endif()

add_bolt_unittest(PassTests
  ${PassTests_SOURCES}
  DISABLE_LLVM_LINK_LLVM_DYLIB
)
```

wenju-he:
> I would prefer to exclude the entire file in CMakeLists.txt since PointerAuthCFIFixup.cpp only applies to the aarch64 platform.

done, thanks @yavtuk

bgergely0:
Hi! isnt ths trying to solve the same issue as https://github.com/llvm/llvm-project/pull/197464? 

That one is already approved, and we are waiting for a merge. 

I only had a quick look at this and see that this is a bit more complicated change. Isn't the linked one sufficient?

wenju-he:
thanks @bgergely0, close this PR.

PR #198445, opened 2026-05-19
[LV] Add cost for VPInstruction::Not.

ElvisWang123:
This patch add the cost for VPInstruction::Not which will lower to Xor.


llvm/test/Transforms/LoopVectorize/AArch64/call-costs.ll
```diff
@@ -104,77 +104,26 @@ exit:
 define void @call_scalarized(ptr noalias %src, ptr noalias %dst) {
```
ElvisWang123, line 104: This loop won't vectorize since the scalar cost and vector cost are identical ( 8.5).

llvm/test/Transforms/LoopVectorize/AArch64/conditional-branches-cost.ll
```diff
@@ -1195,123 +1187,29 @@ define void @pred_udiv_select_cost(ptr %A, ptr %B, ptr %C, i64 %n, i8 %y) #1 {
 ; DEFAULT:       [[VECTOR_BODY]]:
 ; DEFAULT-NEXT:    [[INDEX:%.*]] = phi i64 [ 0, %[[VECTOR_PH1]] ], [ [[INDEX_NEXT:%.*]], %[[VECTOR_BODY]] ]
 ; DEFAULT-NEXT:    [[TMP10:%.*]] = getelementptr i8, ptr [[A]], i64 [[INDEX]]
-; DEFAULT-NEXT:    [[TMP49:%.*]] = shl nuw nsw i64 [[TMP9]], 1
-; DEFAULT-NEXT:    [[TMP53:%.*]] = mul nuw nsw i64 [[TMP9]], 3
-; DEFAULT-NEXT:    [[TMP57:%.*]] = getelementptr i8, ptr [[TMP10]], i64 [[TMP9]]
-; DEFAULT-NEXT:    [[TMP61:%.*]] = getelementptr i8, ptr [[TMP10]], i64 [[TMP49]]
-; DEFAULT-NEXT:    [[TMP65:%.*]] = getelementptr i8, ptr [[TMP10]], i64 [[TMP53]]
```
ElvisWang123, line 1200: This loop won't be interleaved since the vector cost is now 21, which is slightly larger than `small-loop-cost` (20).

llvm/test/Transforms/LoopVectorize/AArch64/optsize_minsize.ll
```diff
@@ -480,44 +480,23 @@ define void @sve_tail_predicate_without_minsize(ptr %p, i8 %a, i8 %b, i8 %c, i32
 ;
 ; MINSIZE-LABEL: define void @sve_tail_predicate_without_minsize(
 ; MINSIZE-SAME: ptr [[P:%.*]], i8 [[A:%.*]], i8 [[B:%.*]], i8 [[C:%.*]], i32 [[N:%.*]]) #[[ATTR1:[0-9]+]] {
-; MINSIZE-NEXT:  [[ENTRY:.*:]]
-; MINSIZE-NEXT:    br label %[[VECTOR_PH:.*]]
-; MINSIZE:       [[VECTOR_PH]]:
-; MINSIZE-NEXT:    [[TMP5:%.*]] = call i64 @llvm.vscale.i64()
-; MINSIZE-NEXT:    [[TMP6:%.*]] = shl nuw i64 [[TMP5]], 4
-; MINSIZE-NEXT:    [[ACTIVE_LANE_MASK_ENTRY:%.*]] = call <vscale x 16 x i1> @llvm.get.active.lane.mask.nxv16i1.i64(i64 0, i64 15)
```
ElvisWang123, line 488: This loop won't be vectorized since this is tested by code size. And the vector loop is much larger than scalar loop.

llvm/test/Transforms/LoopVectorize/X86/predicate-switch.ll
```diff
@@ -1149,52 +1149,10 @@ exit:
...
-; COST-NEXT:    [[WIDE_LOAD:%.*]] = load <4 x i64>, ptr [[NEXT_GEP]], align 1
-; COST-NEXT:    [[TMP7:%.*]] = icmp eq <4 x i64> [[WIDE_LOAD]], splat (i64 -12)
-; COST-NEXT:    [[TMP8:%.*]] = icmp eq <4 x i64> [[WIDE_LOAD]], splat (i64 13)
-; COST-NEXT:    [[TMP9:%.*]] = or <4 x i1> [[TMP7]], [[TMP8]]
-; COST-NEXT:    [[TMP10:%.*]] = xor <4 x i1> [[TMP9]], splat (i1 true)
-; COST-NEXT:    [[TMP11:%.*]] = icmp ule <4 x i64> [[WIDE_LOAD]], [[BROADCAST_SPLAT]]
-; COST-NEXT:    [[TMP12:%.*]] = xor <4 x i1> [[TMP11]], splat (i1 true)
-; COST-NEXT:    [[TMP13:%.*]] = select <4 x i1> [[TMP7]], <4 x i1> [[TMP12]], <4 x i1> zeroinitializer
-; COST-NEXT:    [[TMP14:%.*]] = or <4 x i1> [[TMP13]], [[TMP8]]
-; COST-NEXT:    call void @llvm.masked.store.v4i64.p0(<4 x i64> zeroinitializer, ptr align 1 [[NEXT_GEP]], <4 x i1> [[TMP14]])
```
ElvisWang123, line 1182: This loop currently won't be vectorized because flattening the control flow for the switch-case generates lots of `logical-and`, `or`, and `not`.

llvm/test/Transforms/LoopVectorize/AArch64/call-costs.ll
david-arm: I think the test deliberately wants to see what happens when the call is scalarised. Can you add a loop attribute to force vectorisation with VF=2?

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1276,6 +1276,13 @@ InstructionCost VPInstruction::computeCost(ElementCount VF,
     return Ctx.TTI.getIndexedVectorInstrCostFromEnd(Instruction::ExtractElement,
                                                     VecTy, Ctx.CostKind, 0);
   }
+  case VPInstruction::Not: {
+    Type *RetTy = Ctx.Types.inferScalarType(getOperand(0));
```
david-arm: I think this will end up being too pessimistic for many loops. For example, see this in the induction sve tests:

```
; PRED-NEXT:    [[TMP20:%.*]] = xor i1 [[TMP19]], true
; PRED-NEXT:    br i1 [[TMP20]], label %[[MIDDLE_BLOCK:.*]], label %[[VECTOR_BODY]], !llvm.loop [[LOOP0:![0-9]+]]
```

Sure there is a xor instruction before the branch, but we know that this xor is free because a later instcombine pass will simply remove the xor and reorder the branch labels. I think it would be good to add some code here that returns a cost of 0 if the xor is used directly as the scalar branch condition.

llvm/test/Transforms/LoopVectorize/X86/predicate-switch.ll
david-arm: I think this is fine because it has also a RUN line where it forces vectorisation too.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
Mel-Chen: Hmm...could we currently represent this behavior directly using a branch recipe?

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1276,6 +1276,13 @@ InstructionCost VPInstruction::computeCost(ElementCount VF,
     return Ctx.TTI.getIndexedVectorInstrCostFromEnd(Instruction::ExtractElement,
                                                     VecTy, Ctx.CostKind, 0);
   }
+  case VPInstruction::Not: {
```
Mel-Chen, line 1393: Could we directly fall through and reuse ICmp,FCmp code?

Mel-Chen (commented):
> This patch add the cost for VPInstruction::Now which will lower to Xor.

VPInstruction::Now --> VPInstruction::Not

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
fhahn: It would be nice if this would be handled automatically by a transform, but I think unfortunately that would only be possible after dissolving the loop region, as the backedge is implicit at this point. What we already do is folding the negation in compares, but that does not work for active lane masks

lukel97: I think a quick workaround is probably just to check if the only user is a `m_BranchOnCond` and skip the cost?

Mel-Chen: Yes

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
ElvisWang123: Unfortunately we cannot. Currently ICmp, FCmp won' t account scalar cost in the loop region. 

I moved the entire functionality to the getCostForRecipeWithOpcode to share some logi with other VPInstructions.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
ElvisWang123: Updated, thanks!

llvm/test/Transforms/LoopVectorize/AArch64/call-costs.ll
ElvisWang123: Added, thanks!

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1276,6 +1279,14 @@ InstructionCost VPInstruction::computeCost(ElementCount VF,
...
                                                     VecTy, Ctx.CostKind, 0);
   }
+  case VPInstruction::Not:
+    // InstCombine will fold `xor` to the conditional branch.
+    if (hasOneUse() && match(const_cast<VPUser *>(getSingleUser()),
+                             m_BranchOnCond(m_VPValue())))
+      return 0;
+    return getCostForRecipeWithOpcode(
+        getOpcode(),
+        vputils::onlyFirstLaneUsed(this) ? ElementCount::getFixed(1) : VF, Ctx);
```
Mel-Chen: How about
```suggestion
    return getCostForRecipeWithOpcode(
        Instruction::Xor,
        vputils::onlyFirstLaneUsed(this) ? ElementCount::getFixed(1) : VF, Ctx);
```

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1276,6 +1279,14 @@ InstructionCost VPInstruction::computeCost(ElementCount VF,
     return Ctx.TTI.getIndexedVectorInstrCostFromEnd(Instruction::ExtractElement,
                                                     VecTy, Ctx.CostKind, 0);
   }
+  case VPInstruction::Not:
+    // InstCombine will fold `xor` to the conditional branch.
+    if (hasOneUse() && match(const_cast<VPUser *>(getSingleUser()),
+                             m_BranchOnCond(m_VPValue())))
```
Mel-Chen: Maybe a nit:
getSingleUser() checked hasOneUse(). Therefore, we can do
```
if (VPUser *U = getSingleUser())
 if (match ...
```

david-arm (approved):
LGTM. Thanks!

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
ElvisWang123: We cannot reuse `Instruction::Xor` currently. 
In `getCostForRecipeWithOpcode` it will try to access operand(1) but VPInstruction::Not only have one operand.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
ElvisWang123: Updated, thanks!

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
fhahn: Can we then instead keep the handling of `Not` directly in `VPInstruction::computeCost`? This should be the only place that ever should handle `::Not`, i.e. there should be no other recipes that would call `getCostForRecipeWithOpcode` with `VPInstruction::Not`

lukel97:
This PR looks fine to me but I'm struggling to remember why we have a dedicated `VPInstruction::Not` opcode to begin with instead of just a canonical `VPInstruction::Xor x, -1`. Was it because we didn't have an easy way to pattern match not patterns back in the day? That should be much easier with VPlanPatternMatch now.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
Mel-Chen: > Can we then instead keep the handling of `Not` directly in `VPInstruction::computeCost`? This should be the only place that ever should handle `::Not`, i.e. there should be no other recipes that would call `getCostForRecipeWithOpcode` with `VPInstruction::Not`

+1
I think handling VPInstruction-specific opcodes inside `getCostForRecipeWithOpcode` is not ideal. Since `getCostForRecipeWithOpcode` is a member function of `VPRecipeWithIRFlags`, this pattern risks opcode collisions down the road if other derived recipes (not now) also introduce their own opcodes to maintain.

If we cannot cleanly reuse `getCostForRecipeWithOpcode`, it might be better to fall back to calling TTI directly. Keeping `getCostForRecipeWithOpcode` strictly focused on standard LLVM Instruction opcodes would be much cleaner.

Mel-Chen:
> This PR looks fine to me but I'm struggling to remember why we have a dedicated `VPInstruction::Not` opcode to begin with instead of just a canonical `VPInstruction::Xor x, -1`. Was it because we didn't have an easy way to pattern match not patterns back in the day? That should be much easier with VPlanPatternMatch now.

Agree. A separate patch in the future should be able to migrate all usages of VPInstruction::Not to Instruction::Xor op, true and completely remove VPInstruction::Not.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1285,6 +1285,14 @@ InstructionCost VPInstruction::computeCost(ElementCount VF,
     return Ctx.TTI.getIndexedVectorInstrCostFromEnd(Instruction::ExtractElement,
                                                     VecTy, Ctx.CostKind, 0);
   }
+  case VPInstruction::Not:
+    // InstCombine will fold `xor` to the conditional branch.
+    if (auto *U = const_cast<VPUser *>(getSingleUser()))
+      if (match(U, m_BranchOnCond(m_VPValue())))
+        return 0;
+    return Ctx.TTI.getArithmeticInstrCost(
+        Instruction::Xor, toVectorTy(Ctx.Types.inferScalarType(this), VF),
```
Mel-Chen: Why don't we need vputils::onlyFirstLaneUsed(this)? Is VPInstruction::Not always be widen?

ElvisWang123: Fixed, thanks!

Mel-Chen (approved):
LG

lukel97 (approved):
LGTM

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -1285,6 +1285,17 @@ InstructionCost VPInstruction::computeCost(ElementCount VF,
...
                                                     VecTy, Ctx.CostKind, 0);
   }
+  case VPInstruction::Not: {
+    Type *ValTy = Ctx.Types.inferScalarType(this);
+    // InstCombine will fold `xor` to the conditional branch.
+    if (auto *U = const_cast<VPUser *>(getSingleUser()))
+      if (match(U, m_BranchOnCond(m_VPValue())))
+        return 0;
+    if (!vputils::onlyFirstLaneUsed(this))
+      ValTy = toVectorTy(ValTy, VF);
```
fhahn, line 1400: It looks like we are missing test coverage for the recent changes, would it be possible to cover  via a cost-printing test? I guess it would need a case where the vector type does not fit into a single register, so there's a difference between scalar and vector cost?

ElvisWang123: Added test `AArch64/vpinstruction-not-cost.ll`. And show the test changes in [375b00f](https://github.com/llvm/llvm-project/pull/198445/commits/375b00f74efd99c425c82733a2f3dbaf45106e74). Thanks!

llvm/test/Transforms/LoopVectorize/AArch64/vpinstruction-not-cost.ll
```diff
@@ -0,0 +1,318 @@
+; NOTE: Assertions have been autogenerated by utils/update_analyze_test_checks.py UTC_ARGS: --filter "Cost of" --filter "Selecting VF" --filter-out-after "Selecting VF" --version 6
+; REQUIRES: asserts
+; RUN: opt -passes=loop-vectorize -debug-only=loop-vectorize -disable-output < %s 2>&1 | FileCheck %s
+
+target triple = "aarch64-none-elf"
+
+define i32 @multi_user_cmp(ptr readonly %a, i64 noundef %n) {
```
fhahn: `binop-cost.ll` already contains test for `fneg`, may be good to consolidate there if possible, possibly with adjusting the test file name.

fhahn (approved):
LGTM, thanks

llvm/test/Transforms/LoopVectorize/AArch64/vpinstruction-not-cost.ll
ElvisWang123: Moved to `binop-costs.ll` and renamed to `arith-costs.ll` in https://github.com/llvm/llvm-project/pull/198445/commits/d62044994a54a5aabb01b9fdc9c1d6239fe82c0f . Thanks!

PR #193364, opened 2026-04-22
[X86] Fold splat AND on VGF2P8AFFINEQB source

WalterKruger:
Given that each row within `vgf2p8affineqb`'s matrix controls which source bits are selected, zeroing the same bit within all rows treats that corresponding source bit like it is zero. This means a AND of the input by any splatted 8-bit values can be folded with the matrix. This is patch:

- Can eliminate a constant and/or reduces the instruction count from 2 to 1.
- Only occurs when the matrix is constant, ensuring that it can't increase the dependency chain.
- Don't occur if the AND is multi use well the splat isn't constant, preventing additional operations.
- Work with both constant 8-bit splats and scalars value that where splatted to a vector.
- Includes test coverage for positive cases (by constants, variable scalars, non zero immediates) and negative (multi use, larger splats, variable matrices).

Fixes #191325

PR #198448, opened 2026-05-19
[X86] Fold XOR of VGF2P8AFFINEQB and its source

### LINKED ISSUE #184061: [X86] Missed Fold: `x ^ vgf2p8affineqb(x, M)` => `vgf2p8affineqb(x, M ^ IDENTITY)`
Extension of #178785. The standalone `x` can be interpreted as being affined by the identity matrix (`0x0102040810204080`), which allows it to be folded in the same manner. Only seems beneficial when the matrix is a constant.

```asm
xorWithRev_src:
        vgf2p8affineqb  xmm1, xmm0, xmmword ptr [rip + .LCPI0_0], 0
        vpxor   xmm0, xmm1, xmm0
        ret
```

```asm
xorWithRev_tgt:
        vgf2p8affineqb  xmm0, xmm0, xmmword ptr [rip + .LCPI1_0], 0
        ret
```

https://godbolt.org/z/nhhP4114G

WalterKruger:
Adds a optimization to fold XORs between `vgf2p8affineqb` and its source. That XOR can done through the existing affine transformation by XORing its matrix with the identity matrix. This patch:

- Folds XOR between `vgf2p8affineqb` and its source.
- Only occurs when the matrix is constant, ensuring that it can't increase the dependency chain.
- Doesn't occur if the affine is multi use, preventing an increase in code size.
- Includes test coverage for both positive and negative cases.

Fixes #184061


llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -56303,21 +56259,44 @@ static SDValue combineXorWithGF2P8AFFINEQB(SDNode *N, const SDLoc &DL,
...
          "Unsupported GFNI type");
 
-  // Use X86::isConstantSplat for robust splat constant extraction
+  // Fold: GF2P8AFFINEQB(x, M) ^ x
+  //   =>  GF2P8AFFINEQB(x, M ^ IDENTITY)
+  // The "identity" is a noop. It adds a XOR by the unmodified input.
+  SmallVector<APInt> YEltBits;
+  if (X == XorOp && getTargetConstantBitsFromNode(Y, 8, ConstUndef, YEltBits,
+                                                  /*AllowWholeUndefs=*/false)) {
+    static const uint8_t IdentityMatrtix[] = {128, 64, 32, 16, 8, 4, 2, 1};
```
RKSimon: drop the static and fix typo
```suggestion
    const uint8_t IdentityMatrix[] = {128, 64, 32, 16, 8, 4, 2, 1};
```

RKSimon (changes requested):

WalterKruger:
An upstream GFNI fold (#193364) checked for multi-use using an XOR. This was optimized away so I had to rebase and modify it to use a safer method (a store). I tried to account for this exact issue during that pull but I missed one of the cases.

RKSimon (commented):
Please can you check the windows CI failure?

WalterKruger:
@RKSimon it seems the runner crashed abruptly with no error in the build log. It only indicates that the `self-hosted runner lost communication with the server`. It might be unrelated to my changes.

Can you please try to rerun the Windows job?

RKSimon (approved):
LGTM

PR #198449, opened 2026-05-19
[RISCV][P-ext] Split v4i16/v8i8 AND/OR/XOR on RV32.

topperc:
v2i32 is already scalarized by LegalizeVectorOps.

### DIFF
--- a/llvm/test/CodeGen/RISCV/rvp-simd-64.ll
+++ b/llvm/test/CodeGen/RISCV/rvp-simd-64.ll
@@ -68,32 +68,8 @@ define <8 x i8> @test_psub_b(<8 x i8> %a, <8 x i8> %b) {
 define <4 x i16> @test_and_h(<4 x i16> %a, <4 x i16> %b) {
 ; RV32-LABEL: test_and_h:
 ; RV32:       # %bb.0:
-; RV32-NEXT:    addi sp, sp, -32
-; RV32-NEXT:    .cfi_def_cfa_offset 32
-; RV32-NEXT:    sw a3, 20(sp)
-; RV32-NEXT:    sw a1, 12(sp)
-; RV32-NEXT:    sw a2, 16(sp)
-; RV32-NEXT:    sw a0, 8(sp)
-; RV32-NEXT:    lh a0, 22(sp)
-; RV32-NEXT:    lh a1, 14(sp)
-; RV32-NEXT:    and a0, a1, a0
-; RV32-NEXT:    sh a0, 30(sp)
-; RV32-NEXT:    lh a0, 20(sp)
-; RV32-NEXT:    lh a1, 12(sp)
-; RV32-NEXT:    and a0, a1, a0
-; RV32-NEXT:    sh a0, 28(sp)
-; RV32-NEXT:    lh a0, 18(sp)
-; RV32-NEXT:    lh a1, 10(sp)
-; RV32-NEXT:    and a0, a1, a0
-; RV32-NEXT:    sh a0, 26(sp)
-; RV32-NEXT:    lh a0, 16(sp)
-; RV32-NEXT:    lh a1, 8(sp)
-; RV32-NEXT:    and a0, a1, a0
-; RV32-NEXT:    sh a0, 24(sp)
-; RV32-NEXT:    lw a0, 24(sp)
-; RV32-NEXT:    lw a1, 28(sp)
-; RV32-NEXT:    addi sp, sp, 32
-; RV32-NEXT:    .cfi_def_cfa_offset 0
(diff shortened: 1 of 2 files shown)


llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -5981,30 +5459,37 @@ define <8 x i8> @test_bitreverse_v8i8(<8 x i8> %a) {
...
+; RV32-NEXT:    or a1, a2, a1
+; RV32-NEXT:    or a0, a3, a0
+; RV32-NEXT:    psrli.dh a2, a0, 1
+; RV32-NEXT:    and a1, a1, a4
+; RV32-NEXT:    and a3, a3, a4
+; RV32-NEXT:    and a0, a0, a4
+; RV32-NEXT:    and a2, a2, a4
+; RV32-NEXT:    pslli.dh a0, a0, 1
+; RV32-NEXT:    or a1, a3, a1
+; RV32-NEXT:    or a0, a2, a0
```
topperc, line 5479: This is a regression in number of instruction, but I'd prefer to fix it separately.

llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -5885,8 +5362,22 @@ define <4 x i16> @test_bswap_v4i16(<4 x i16> %a) {
...
+; RV32-NEXT:    and a1, a1, a2
+; RV32-NEXT:    or a3, a3, a7
+; RV32-NEXT:    and a0, a0, a2
+; RV32-NEXT:    or a2, a4, a6
+; RV32-NEXT:    pslli.dw a0, a0, 8
+; RV32-NEXT:    or a1, t2, a1
+; RV32-NEXT:    or a0, t1, a0
+; RV32-NEXT:    or a1, a1, a3
+; RV32-NEXT:    or a0, a0, a2
 ; RV32-NEXT:    ret
```
sihuan, line 5368: There are a few more regressions on `v2i32` that look fixable by narrowing the new Custom action scope:

- `test_bswap_v2i32` (this one): 2× `rev8` → ~17-instruction `shl`/`srl`/`and`/`or` sequence
- `test_select_v2i32`: ~5 ops (branch) → ~10 ops (branchless)
- `test_vselect_v2i32`: ~8 ops (branch) → ~12 ops (branchless)

`test_bitreverse_v2i32` notably does not regress (still 2× `rev`)

One thing worth considering: `v2i32` may not need customization at this level. `EXTRACT_VECTOR_ELT` and `BUILD_VECTOR` are already set `Legal` for `v2i32`, so the generic `Expand` fallback for AND/OR/XOR already scalarizes into two clean `i32` ops without custom code. 

The three regressions appear to share one mechanism: `bswap`/`select`/`vselect` are themselves at `Expand`, and their default expansions use AND/OR/XOR internally. When AND/OR were also `Expand` on `v2i32`, the chain bailed out to scalarize — at which point each `i32` half got matched to Zbb's `rev8` / branch-based select / etc. Once AND/OR become `Custom` on `v2i32`, those expansions stay in vector form and emit long masking sequences instead.

Maybe consider restricting the new Custom action to `{v8i8, v4i16}`? That would recover all three regressions while preserving the gains on `v8i8`/`v4i16`. 

Locally tested.

sihuan (approved):
LGTM

llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -5628,61 +5192,41 @@ define <2 x i32> @test_select_v2i32(i1 %cond, <2 x i32> %a, <2 x i32> %b) {
...
-; RV32-NEXT:    lhu a6, 0(sp)
-; RV32-NEXT:    lhu a7, 2(sp)
-; RV32-NEXT:    lhu t0, 4(sp)
-; RV32-NEXT:    lhu t1, 6(sp)
-; RV32-NEXT:    xor a6, a6, a0
-; RV32-NEXT:    xor a7, a7, a1
-; RV32-NEXT:    xor t0, t0, a2
-; RV32-NEXT:    xor t1, t1, a3
-; RV32-NEXT:    snez a6, a6
+; RV32-NEXT:    lhu a6, 12(sp)
```
lenary, line 5203: What's going on here? I think this applies to the v8i8 vselect too. 

This can be addressed in a follow-up :)

topperc: The setcc is being scalarized. Before, the vselect was expanded to and+andn+or, which were then scalarized. Now the and+andn+or don't get scalarized.

PR #147297, opened 2025-07-07
[LV] Convert gather loads with constant stride into strided loads

Mel-Chen:
This patch detects non-consecutive load accesses (i.e. gather) with a constant stride, such as:
```
  void stride(int* a, int *b, int n) {
    for (int i = 0; i < n; i++)
      a[i * 5] = b[i * 5] + i;
  }
```
and converts them into strided loads when legal and profitable, using experimental_vp_strided_load.
The new VPlan transformation, convertToStridedAccesses, hoists the functionality of RISCVGatherScatterLowering into the vectorizer, enabling a more precise cost estimation during vectorization. Additionally, by leveraging SCEV for stride analysis, the vectorizer can potentially detect more opportunities to optimize gathers into strided loads.

This enables more efficient code generation for targets like RISC-V that support strided loads natively.

PR #198451, opened 2026-05-19
[LV] Simplify the reduction initial value

ChunyuLiao:
The reduction adjustment with the initial value as part of the epilogue 
rather than including it as part of the initial vector value.

For RISCV, `insertelement (<vscale x 4 x i32> zeroinitializer, i32 x, i32 0)`
multiple instructions are required to implement it.

Some testcases, llvm vs Gcc: https://gcc.godbolt.org/z/7Y3Y57Kev


llvm/lib/Transforms/Vectorize/LoopVectorize.cpp
```diff
@@ -7096,6 +7096,29 @@ void LoopVectorizationPlanner::addReductionResultComputation(
           Builder.createScalarCast(ExtendOpc, FinalReductionResult, PhiTy, {});
     }
 
+    RecurKind RK = PhiR->getRecurrenceKind();
+    bool SimplifyInit = false;
+    if (!PhiR->isInLoop() && CM.foldTailWithEVL() &&
```
lukel97: Is there a reason why this is only done with EVL tail folding? Is it profitable for other targets? Ideally we want to avoid hardcoding EVL specific logic in generic vplan transforms

fhahn: It sounds like this is something that should be cost driven, checking whether scalar add at the end is cheaper than insert up-front?

This should probably be a separate transform, rather than complicating the generic construction code

lukel97: Yeah I was thinking a separate transform too. Just from playing around with insertelement it looks like RISC-V is the only target that has such expensive materialization sequences. But I think we can also improve the codegen a bit, I'll take a look at that in the backend

lukel97: Making a note to self, https://github.com/llvm/llvm-project/pull/134120 was an attempt to fix the poor codegen

ChunyuLiao: Indeed, using EVL here is intended to restrict it to RISC-V only, which is not appropriate.

I will continue to try a cost-driven approach and create a separate transformation. If there are any existing code examples that I can refer to, please guide me.

lukel97: #147297 is an example of a cost model driven transform, it might be a useful reference

ChunyuLiao: try to use cost-driven approach and create a transformation to optimize the reduction initial.

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -7080,6 +7080,84 @@ static CallWideningDecision decideCallWidening(VPInstruction &VPI,
...
+        PhiR->getVFScaleFactor() != 1 || PhiR->hasUsesOutsideReductionChain())
+      continue;
+
+    VPInstruction *FinalReductionResult =
+        vputils::findComputeReductionResult(PhiR);
+    if (!FinalReductionResult)
+      continue;
+
+    RecurKind RK = PhiR->getRecurrenceKind();
+    if (RK == RecurKind::Add || RK == RecurKind::Sub || RK == RecurKind::And ||
```
eas: This can be `is_contained({RecurKind::Add, RecurKind::Sub, ...}, RK)`, I think. Also, early return maybe?

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -7080,6 +7080,84 @@ static CallWideningDecision decideCallWidening(VPInstruction &VPI,
...
+          return LoopVectorizationPlanner::getDecisionAndClampRange(
+              [&](ElementCount VF) {
+                auto *VecTy = toVectorTy(ScalarTy, VF);
+                if (!VecTy->isVectorTy())
+                  return false;
+                const InstructionCost CurrentCost =
+                    Ctx.TTI.getVectorInstrCost(Instruction::InsertElement,
+                                               VecTy, TTI::TCK_RecipThroughput);
+                const InstructionCost ScalarCost =
+                    Ctx.TTI.getArithmeticInstrCost(Instruction::Xor, ScalarTy,
```
eas: Should this use `ScalarOp` instead of `Xor`?

Issue #83819, opened 2024-03-04
Documentation comments for macros are not preserved


daniel-grumberg:
Documentation comments for C preprocessor macros are not preserved.

shafik:
I am not sure I understand the bug, can you provide a minimal reproducer, preferably with a godbolt link.

daniel-grumberg:
Maybe I filtered this to the wrong place, but effectively the problem is that there is no way that I am aware of to track documentation comments attached to a macro definition.
```c
/// My fancy doc comment.
#define ADD(a, b) ((a) + (b))
```
The functionality of tracking the documentation comments associated with an AST node in the frontend already exists but there is no similar functionality with macros and their associated documentation comment.

shafik:
CC @AaronBallman

AaronBallman:
Correct, Clang's AST doesn't have a representation for macros so there's nothing to attach the documentation comments *to*. But we could perhaps find a way to squirrel those away with the macro definitions in the preprocessor: https://github.com/llvm/llvm-project/blob/f8fab2126ffab713f4ab4619360b6941be6d4e35/clang/include/clang/Lex/MacroInfo.h#L39

DougGregor:
This is implemented

PR #198452, opened 2026-05-19
Associate documentation comments with macro definitions

DougGregor:
Generalize the infrastructure that finds comments attached to declarations to also find comments attached to macros. Surface these comments through libclang and the symbol graph.

Fixes issue #83819.

### DIFF
--- a/clang/include/clang/AST/ASTContext.h
+++ b/clang/include/clang/AST/ASTContext.h
@@ -30,7 +30,7 @@
 #include "clang/Basic/LLVM.h"
 #include "clang/Basic/PartialDiagnostic.h"
 #include "clang/Basic/SourceLocation.h"
-#include "clang/Lex/MacroInfo.h"
+#include "clang/Lex/MacroBase.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/DenseMapInfo.h"
 #include "llvm/ADT/DenseSet.h"
--- a/clang/include/clang/AST/ASTContext.h
+++ b/clang/include/clang/AST/ASTContext.h
@@ -1034,21 +1034,21 @@ public:
   ///
   /// Don't do anything if a comment has already been attached to \p Original
   /// or its redeclaration chain.
-  void cacheRawCommentForDecl(RawCommentLookupKey Original,
-                              const RawComment &Comment) const;
+  void cacheRawComment(RawCommentLookupKey Original,
+                       const RawComment &Comment) const;
 
   /// \returns searches \p CommentsInFile for doc comment for \p Key.
   ///
   /// \p RepresentativeLocForDecl is used as a location for searching doc
   /// comments. \p CommentsInFile is a mapping offset -> comment of files in the
   /// same file where \p RepresentativeLocForDecl is.
-  RawComment *getRawCommentForDeclNoCacheImpl(
+  RawComment *getRawCommentNoCacheImpl(
       RawCommentLookupKey Key, const SourceLocation RepresentativeLocForDecl,
       const std::map<unsigned, RawComment *> &CommentsInFile) const;
 
   /// Return the documentation comment attached to a given declaration or
   /// macro, without looking into cache.
-  RawComment *getRawCommentForDeclNoCache(RawCommentLookupKey Key) const;
+  RawComment *getRawCommentNoCache(RawCommentLookupKey Key) const;
 
 public:
   void addComment(const RawComment &RC);
--- a/clang/include/clang/AST/ASTContext.h
+++ b/clang/include/clang/AST/ASTContext.h
@@ -30,6 +30,7 @@
 #include "clang/Basic/LLVM.h"
 #include "clang/Basic/PartialDiagnostic.h"
 #include "clang/Basic/SourceLocation.h"
+#include "clang/Lex/MacroInfo.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/DenseMapInfo.h"
 #include "llvm/ADT/DenseSet.h"
@@ -996,11 +997,16 @@ public:
   /// True if comments are already loaded from ExternalASTSource.
   mutable bool CommentsLoaded = false;
 
-  /// Mapping from declaration to directly attached comment.
+  /// Key used to look up the raw comment attached to a declaration or macro.
+  using RawCommentLookupKey =
+      llvm::PointerUnion<const Decl *, const MacroInfo *>;
+
+  /// Mapping from declaration or macro to directly attached comment.
   ///
   /// Raw comments are owned by Comments list.  This mapping is populated
   /// lazily.
-  mutable llvm::DenseMap<const Decl *, const RawComment *> DeclRawComments;
+  mutable llvm::DenseMap<RawCommentLookupKey, const RawComment *>
+      DeclRawComments;
 
   /// Mapping from canonical declaration to the first redeclaration in chain
   /// that has a comment attached.
@@ -1022,37 +1028,40 @@ public:
   /// redeclaration.
   mutable llvm::DenseMap<const Decl *, comments::FullComment *> ParsedComments;
 
-  /// Attaches \p Comment to \p OriginalD and to its redeclaration chain
-  /// and removes the redeclaration chain from the set of commentless chains.
+  /// Attaches \p Comment to \p Original (a declaration or macro), and to its
+  /// redeclaration chain when \p Original is a declaration. Removes the
+  /// redeclaration chain from the set of commentless chains.
   ///
-  /// Don't do anything if a comment has already been attached to \p OriginalD
+  /// Don't do anything if a comment has already been attached to \p Original
   /// or its redeclaration chain.
-  void cacheRawCommentForDecl(const Decl &OriginalD,
+  void cacheRawCommentForDecl(RawCommentLookupKey Original,
                               const RawComment &Comment) const;
 
-  /// \returns searches \p CommentsInFile for doc comment for \p D.
+  /// \returns searches \p CommentsInFile for doc comment for \p Key.
   ///
   /// \p RepresentativeLocForDecl is used as a location for searching doc
   /// comments. \p CommentsInFile is a mapping offset -> comment of files in the
   /// same file where \p RepresentativeLocForDecl is.
   RawComment *getRawCommentForDeclNoCacheImpl(
-      const Decl *D, const SourceLocation RepresentativeLocForDecl,
+      RawCommentLookupKey Key, const SourceLocation RepresentativeLocForDecl,
       const std::map<unsigned, RawComment *> &CommentsInFile) const;
 
-  /// Return the documentation comment attached to a given declaration,
-  /// without looking into cache.
-  RawComment *getRawCommentForDeclNoCache(const Decl *D) const;
+  /// Return the documentation comment attached to a given declaration or
+  /// macro, without looking into cache.
+  RawComment *getRawCommentForDeclNoCache(RawCommentLookupKey Key) const;
 
 public:
   void addComment(const RawComment &RC);
 
-  /// Return the documentation comment attached to a given declaration.
-  /// Returns nullptr if no comment is attached.
+  /// Return the documentation comment attached to a given declaration or
+  /// macro.  Returns nullptr if no comment is attached.
   ///
   /// \param OriginalDecl if not nullptr, is set to declaration AST node that
   /// had the comment, if the comment we found comes from a redeclaration.
+  /// Macros have no redeclaration chain, so this is set to nullptr when
+  /// \p Key is a \c MacroInfo.
   const RawComment *
-  getRawCommentForAnyRedecl(const Decl *D,
+  getRawCommentForAnyRedecl(RawCommentLookupKey Key,
                             const Decl **OriginalDecl = nullptr) const;
 
   /// Searches existing comments for doc comments that should be attached to \p
--- a/clang/include/clang/AST/ASTContext.h
+++ b/clang/include/clang/AST/ASTContext.h
@@ -1005,8 +1005,7 @@ public:
   ///
   /// Raw comments are owned by Comments list.  This mapping is populated
   /// lazily.
-  mutable llvm::DenseMap<RawCommentLookupKey, const RawComment *>
-      RawComments;
+  mutable llvm::DenseMap<RawCommentLookupKey, const RawComment *> RawComments;
 
   /// Mapping from canonical declaration to the first redeclaration in chain
   /// that has a comment attached.
--- a/clang/include/clang/AST/ASTContext.h
+++ b/clang/include/clang/AST/ASTContext.h
@@ -1006,7 +1006,7 @@ public:
   /// Raw comments are owned by Comments list.  This mapping is populated
   /// lazily.
   mutable llvm::DenseMap<RawCommentLookupKey, const RawComment *>
-      DeclRawComments;
+      RawComments;
 
   /// Mapping from canonical declaration to the first redeclaration in chain
   /// that has a comment attached.
@@ -1043,7 +1043,7 @@ public:
   /// comments. \p CommentsInFile is a mapping offset -> comment of files in the
   /// same file where \p RepresentativeLocForDecl is.
   RawComment *getRawCommentNoCacheImpl(
-      RawCommentLookupKey Key, const SourceLocation RepresentativeLocForDecl,
+      RawCommentLookupKey Key, const SourceLocation RepresentativeLoc,
       const std::map<unsigned, RawComment *> &CommentsInFile) const;
 
   /// Return the documentation comment attached to a given declaration or
--- a/clang/lib/AST/ASTContext.cpp
+++ b/clang/lib/AST/ASTContext.cpp
@@ -245,12 +245,11 @@ getLocsForCommentSearch(ASTContext::RawCommentLookupKey Key,
 }
 
 RawComment *ASTContext::getRawCommentNoCacheImpl(
-    RawCommentLookupKey Key, const SourceLocation RepresentativeLocForDecl,
+    RawCommentLookupKey Key, const SourceLocation RepresentativeLoc,
     const std::map<unsigned, RawComment *> &CommentsInTheFile) const {
   // If the declaration doesn't map directly to a location in a file, we
   // can't find the comment.
-  if (RepresentativeLocForDecl.isInvalid() ||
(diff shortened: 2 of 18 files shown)


QuietMisdreavus:
Thanks so much for working on this! The ExtractAPI part looks good. I read through the ASTContext updates and they look fine to me, but i'll defer to someone who regularly works on it for the ultimate review.

clang/test/ExtractAPI/macro_doc_comments.c
```diff
@@ -0,0 +1,67 @@
...
+#define HELLO 1
+// HELLO-LABEL: "!testLabel": "c:@macro@HELLO"
+// HELLO:      "docComment": {
+// HELLO-NEXT:   "lines": [
+// HELLO:          "text": "The greeting count."
+// HELLO:        ]
+// HELLO-NEXT: },
+
+// A C-style block doc comment also attaches.
+// RUN: FileCheck %s --input-file %t/output.symbols.json --check-prefix BLOCK
```
hnrklssn, line 19: Do we actually need separate FileCheck invocations for each macro?

clang/include/clang/AST/ASTContext.h
```diff
@@ -1022,37 +1028,40 @@ class ASTContext : public RefCountedBase<ASTContext> {
...
-  /// and removes the redeclaration chain from the set of commentless chains.
+  /// Attaches \p Comment to \p Original (a declaration or macro), and to its
+  /// redeclaration chain when \p Original is a declaration. Removes the
+  /// redeclaration chain from the set of commentless chains.
   ///
-  /// Don't do anything if a comment has already been attached to \p OriginalD
+  /// Don't do anything if a comment has already been attached to \p Original
   /// or its redeclaration chain.
-  void cacheRawCommentForDecl(const Decl &OriginalD,
+  void cacheRawCommentForDecl(RawCommentLookupKey Original,
```
hnrklssn: I can't think of a better noun that captures both decls and macros, but it's a bit confusing that these functions with `ForDecl` in their name don't operate exclusively on decls. It's a bit verbose, but `ForDeclOrComment` might still be an improvement.

hnrklssn: Nvm, fixed in the next commit.

clang/test/ExtractAPI/macro_doc_comments.c
DougGregor: It's actually annoying to fix, because there is existing nondeterminism in the processing of macros in `ExtractAPIConsumer.cpp`. I'd rather that be addressed separately.

clang/lib/AST/ASTContext.cpp
```diff
@@ -116,9 +117,38 @@ enum FloatingRank {
...
+    // and our directive.
+    FileIDAndOffset Decomposed = SourceMgr.getDecomposedLoc(DefLoc);
+    bool Invalid = false;
+    StringRef Buffer = SourceMgr.getBufferData(Decomposed.first, &Invalid);
+    if (Invalid)
+      return {};
+    unsigned Offset = Decomposed.second;
+    while (Offset > 0 && Buffer[Offset - 1] != '#' &&
+           Buffer[Offset - 1] != '\n')
+      --Offset;
```
ojhunt: Maybe something like

```cpp
if (unsigned Found = Buffer.find_last_of("#\n", Offset);
    Found != StringRef::npos)
  Offset = Found;
```

ojhunt (approved):
Looks reasonable to me, the only suggestion I have is dropping the local reverse search. I think `find_last_of` should do what you're after.

clang/test/Index/annotate-comments-macros.c
```diff
@@ -0,0 +1,27 @@
...
+
+#define BARE 2
+
+/**
+ * \brief Add two numbers.
+ */
+#define SUM(x, y) ((x) + (y))
+
+#define TRAILING 3 ///< Trailing macro doc.
+
```
egorzhdan, line 14: Is it worth having a test with a redefined macro? e.g.
```cpp
/// abc
#define REDEFINED 123

/// def
#define REDEFINED 456
```

egorzhdan (approved):
Looks great, thanks!
I have one comment on the tests.

clang/include/clang/AST/ASTContext.h
```diff
@@ -996,11 +997,16 @@ class ASTContext : public RefCountedBase<ASTContext> {
...
+  using RawCommentLookupKey =
+      llvm::PointerUnion<const Decl *, const MacroInfo *>;
+
+  /// Mapping from declaration or macro to directly attached comment.
   ///
   /// Raw comments are owned by Comments list.  This mapping is populated
   /// lazily.
-  mutable llvm::DenseMap<const Decl *, const RawComment *> DeclRawComments;
+  mutable llvm::DenseMap<RawCommentLookupKey, const RawComment *>
+      DeclRawComments;
```
snprajwal:
```suggestion
      RawComments;
```

clang/lib/AST/ASTContext.cpp
```diff
@@ -214,8 +244,8 @@ getDeclLocsForCommentSearch(const Decl *D, SourceManager &SourceMgr) {
   return Locations;
 }
 
-RawComment *ASTContext::getRawCommentForDeclNoCacheImpl(
-    const Decl *D, const SourceLocation RepresentativeLocForDecl,
+RawComment *ASTContext::getRawCommentNoCacheImpl(
+    RawCommentLookupKey Key, const SourceLocation RepresentativeLocForDecl,
```
snprajwal:
```suggestion
    RawCommentLookupKey Key, const SourceLocation RepresentativeLoc,
```

clang/include/clang/AST/ASTContext.h
```diff
@@ -1022,37 +1028,40 @@ class ASTContext : public RefCountedBase<ASTContext> {
...
-  /// \returns searches \p CommentsInFile for doc comment for \p D.
+  /// \returns searches \p CommentsInFile for doc comment for \p Key.
   ///
   /// \p RepresentativeLocForDecl is used as a location for searching doc
   /// comments. \p CommentsInFile is a mapping offset -> comment of files in the
   /// same file where \p RepresentativeLocForDecl is.
-  RawComment *getRawCommentForDeclNoCacheImpl(
-      const Decl *D, const SourceLocation RepresentativeLocForDecl,
+  RawComment *getRawCommentNoCacheImpl(
+      RawCommentLookupKey Key, const SourceLocation RepresentativeLocForDecl,
```
snprajwal:
```suggestion
      RawCommentLookupKey Key, const SourceLocation RepresentativeLoc,
```

clang/lib/AST/ASTContext.cpp
```diff
@@ -227,6 +257,9 @@ RawComment *ASTContext::getRawCommentForDeclNoCacheImpl(
   if (CommentsInTheFile.empty())
     return nullptr;
 
+  const auto *D = dyn_cast<const Decl *>(Key);
+  const bool IsMacro = isa<const MacroInfo *>(Key);
+
   // Decompose the location for the declaration and find the beginning of the
   // file buffer.
   const FileIDAndOffset DeclLocDecomp =
```
snprajwal:
```suggestion
  const FileIDAndOffset LocDecomp =
```

clang/lib/AST/ASTContext.cpp
```diff
@@ -292,8 +326,8 @@ RawComment *ASTContext::getRawCommentForDeclNoCacheImpl(
   return CommentBeforeDecl;
 }
 
-RawComment *ASTContext::getRawCommentForDeclNoCache(const Decl *D) const {
-  const auto DeclLocs = getDeclLocsForCommentSearch(D, SourceMgr);
+RawComment *ASTContext::getRawCommentNoCache(RawCommentLookupKey Key) const {
+  const auto DeclLocs = getLocsForCommentSearch(Key, SourceMgr);
```
snprajwal:
```suggestion
  const auto Locs = getLocsForCommentSearch(Key, SourceMgr);
```

clang/lib/AST/ASTContext.cpp
```diff
@@ -292,8 +326,8 @@ RawComment *ASTContext::getRawCommentForDeclNoCacheImpl(
...
 
   for (const auto DeclLoc : DeclLocs) {
```
snprajwal:
```suggestion
  for (const auto Loc : Locs) {
```

clang/lib/AST/ASTContext.cpp
```diff
@@ -292,8 +326,8 @@ RawComment *ASTContext::getRawCommentForDeclNoCacheImpl(
...
     // If the declaration doesn't map directly to a location in a file, we
```
snprajwal:
```suggestion
    // If the declaration or macro doesn't map directly to a location in a file, we
```

clang/tools/libclang/CIndex.cpp
```diff
@@ -9326,51 +9326,55 @@ CXString clang_Cursor_getBinaryOpcodeStr(enum CX_BinaryOperatorKind Op) {
...
-  if (!clang_isDeclaration(C.kind))
-    return clang_getNullRange();
-
-  const Decl *D = getCursorDecl(C);
+static const RawComment *getCursorRawComment(CXCursor C) {
   ASTContext &Context = getCursorContext(C);
-  const RawComment *RC = Context.getRawCommentForAnyRedecl(D);
+  if (clang_isDeclaration(C.kind))
+    return Context.getRawCommentForAnyRedecl(getCursorDecl(C));
+  if (C.kind == CXCursor_MacroDefinition) {
```
snprajwal, line 9333: With this change, the public APIs below now return a non-null value for this kind. I feel it'll be good to have a release note for libclang, would you mind adding it please?

clang/include/clang/AST/ASTContext.h
```diff
@@ -30,6 +30,7 @@
 #include "clang/Basic/LLVM.h"
 #include "clang/Basic/PartialDiagnostic.h"
 #include "clang/Basic/SourceLocation.h"
+#include "clang/Lex/MacroInfo.h"
```
snprajwal: This include now couples `AST` and `Lex` headers (I see other CPP files in AST importing Lex, but no headers), I'm not sure if we want to introduce that.

Do you think we can instead forward declare `MacroInfo` and specialise `PointerLikeTypeTraits`, like how it's done here:

https://github.com/llvm/llvm-project/blob/51279e7e0527a5dee6a5066521e16bc7ae26037d/clang/include/clang/AST/DeclBase.h#L1328-L1338

snprajwal (commented):
Thanks for working on this Doug! The ExtractAPI changes LGTM, I have a concern about the new header include. Also left some renaming suggestions.

clang/include/clang/AST/ASTContext.h
Sirraide: The `AST` library can include `Lex` headers; that’s fine (when in doubt, check the `CMakeLists.txt` file in the corresponding directory in `lib/`; you’ll see that `clangAST` links against `clangLex`)

snprajwal: Yes of course (as mentioned, the implementation files do include Lex), it's just that we've not had an include in the headers _so far_,  so I was just wondering if that's a good convention to uphold. `ASTContext.h` is included quite widely, so introducing new includes might impact build times (although in this case the file is small enough that it probably doesn't make a meaningful difference)

clang/test/Index/annotate-comments-macros.c
DougGregor: Good idea! It actually turned up a bug in libclang (we'd always get a cursor pointing to the second definition, even if asking the question after the first)

clang/lib/AST/ASTContext.cpp
DougGregor: Yes, that's nicer. Thanks for the suggestion.

clang/lib/AST/ASTContext.cpp
```diff
@@ -292,13 +323,13 @@ RawComment *ASTContext::getRawCommentForDeclNoCacheImpl(
...
 }
 
-RawComment *ASTContext::getRawCommentForDeclNoCache(const Decl *D) const {
-  const auto DeclLocs = getDeclLocsForCommentSearch(D, SourceMgr);
+RawComment *ASTContext::getRawCommentNoCache(RawCommentLookupKey Key) const {
+  const auto Locs = getLocsForCommentSearch(Key, SourceMgr);
 
-  for (const auto DeclLoc : DeclLocs) {
-    // If the declaration doesn't map directly to a location in a file, we
+  for (const auto Loc : Locs) {
```
ojhunt, line 329: Given you're changing this code can you replace the auto with the actual type (sorry I missed this in the earlier go through)

clang/lib/AST/ASTContext.cpp
```diff
@@ -331,21 +362,37 @@ void ASTContext::addComment(const RawComment &RC) {
...
-                                                const Decl *D,
-                                                const Decl **OriginalDecl) const {
-  if (!D) {
+const RawComment *
+ASTContext::getRawCommentForAnyRedecl(RawCommentLookupKey Key,
+                                      const Decl **OriginalDecl) const {
+  if (Key.isNull()) {
     if (OriginalDecl)
-      OriginalDecl = nullptr;
+      *OriginalDecl = nullptr;
```
ojhunt, line 370: Hahaha, whoops!

Given you found it, I assume 1+ of the tests hit this?

clang/lib/AST/ASTContext.cpp
```diff
@@ -331,21 +362,37 @@ void ASTContext::addComment(const RawComment &RC) {
...
   }
 
+  // Macros have no redeclaration chain: look up directly, populate the cache,
+  // and return.
+  if (const auto *MI = dyn_cast<const MacroInfo *>(Key)) {
+    if (OriginalDecl)
+      *OriginalDecl = nullptr;
+    auto Existing = RawComments.find(Key);
+    if (Existing != RawComments.end())
+      return Existing->second;
```
ojhunt, line 381: Very minor nit that I'm not going to say "you must change this", I prefer

```cpp
  if (auto Existing = RawComments.find(Key);
      Existing != RawComments.end())
    return Existing->second;
```

Just to ensure the search result is scoped, but again that's a preference not a requirement

clang/include/clang/AST/ASTContext.h
DougGregor: I went ahead and did this. MacroInfo.h does bring in a bunch of headers, including <algorithm>, so I think it's worth holding the line and not including it.

ojhunt (approved):
Sigh, I once again made a bunch of review comments but clicked the wrong submit review button.

These are minor and subjective nits (with the exception of the `auto` change), so I'm approving but please have a look at the comments and make the changes if you agree.

clang/lib/AST/ASTContext.cpp
DougGregor: It was a drive-by fix, I didn't see any tests that failed.

ojhunt: Ah, the best \o/

nathanchance:
I am seeing `bindgen` crash with SIGSEGV after this change:

```
$ bindgen scripts/rust_is_available_bindgen_libclang.h
clang diag: scripts/rust_is_available_bindgen_libclang.h:2:9: warning: clang version 23.0.0git (https://github.com/llvm/llvm-project.git 77e43ec11cd8fbe1de491118b54de9bba94510a8) [-W#pragma-messages]
fish: Job 1, 'bindgen scripts/rust_is_availab…' terminated by signal SIGSEGV (Address boundary error)
```

At the parent change, there is no crash. What would be needed to see what is going wrong here?

```
# bad: [5899ecd0d2420318d935948cd5b25ea56c2538fb] Revert "[NVPTX] Fixup and test ExpandVariadics (#195709)" (#199235)
# good: [1bd4457660bd5499e3dc6f6007749f6f512b674d] [CIR] Implement TernaryOp for Vec of float in OpenCL (#198846)
git bisect start '5899ecd0d2420318d935948cd5b25ea56c2538fb' '1bd4457660bd5499e3dc6f6007749f6f512b674d'
# bad: [625e277979356e8fc51a4a75667a331c5ceffccd] [mlir] Specify isSigned when creating APInt for I16 attributes (#198687)
git bisect bad 625e277979356e8fc51a4a75667a331c5ceffccd
# good: [0f679999aae135b388c25fb1acbb030109c6418f] Revert "[libc] Enable baremetal float printf using modular format" (#199114)
git bisect good 0f679999aae135b388c25fb1acbb030109c6418f
# bad: [6149387b2ed28ba6875bdce24e5c774bace764d6] [CIR] Implement lowering of atomic-func ptrs (#199045)
git bisect bad 6149387b2ed28ba6875bdce24e5c774bace764d6
# bad: [9955c424bb2ba8aaef1b3dcf0409f7b573be9954] [llvm][Instrumentor] Fix non-determinism in Instrumentor (#198833)
git bisect bad 9955c424bb2ba8aaef1b3dcf0409f7b573be9954
# bad: [77e43ec11cd8fbe1de491118b54de9bba94510a8] Associate documentation comments with macro definitions (#198452)
git bisect bad 77e43ec11cd8fbe1de491118b54de9bba94510a8
# good: [db614d99eea9ce78c4b1a44a6e01eb1a4ee8d5b2] [Polly] Do not invalidate SCEV before codegen (#194677)
git bisect good db614d99eea9ce78c4b1a44a6e01eb1a4ee8d5b2
# first bad commit: [77e43ec11cd8fbe1de491118b54de9bba94510a8] Associate documentation comments with macro definitions (#198452)
```

EDIT: It looks like https://github.com/llvm/llvm-project/commit/9d533e977eea1e1c0ee94515bc5dd0bd70d15d95 resolves this.

ojhunt:
> I am seeing `bindgen` crash with SIGSEGV after this change:
> 
> ```
> $ bindgen scripts/rust_is_available_bindgen_libclang.h
> clang diag: scripts/rust_is_available_bindgen_libclang.h:2:9: warning: clang version 23.0.0git (https://github.com/llvm/llvm-project.git 77e43ec11cd8fbe1de491118b54de9bba94510a8) [-W#pragma-messages]
> fish: Job 1, 'bindgen scripts/rust_is_availab…' terminated by signal SIGSEGV (Address boundary error)
> ```
> 
> At the parent change, there is no crash. What would be needed to see what is going wrong here?
> 

Can you include a stack trace? (Ideally debug and/or assertions) - without that information we don't know anything at all

PR #198454, opened 2026-05-19
[alpha.webkit.UncountedCallArgsChecker] Check arguments of CXXConstructExpr

rniwa:
Check arguments of CXXConstructExpr like CallExpr.

This PR fixes a subtle bug in isCtorOfRetainPtrOrOSPtr that it was missing a check for OSObjectPtr,
and it was getting an empty string for the name when getting called on a CXXConstructorDecl
as well as similar bugs in isCtorOf* functions.


t-rasmud (approved):
LGTM!

PR #198457, opened 2026-05-19
[MLIR][Arith][WIE] Fix Conversions for Shift Left and Right Ops

veera-efficient:
The current conversions for `arith.shli`, `arith.shrui` and `arith.shrsi` does not handle shift by zero correctly. It tries to shift a `iN` value by N and results in a poison value.

This PR fixes it by adopting the math used by RISCV backend to lower 64 bit shift operations for 32 bit targets. This implementation uses less number of ops than the existing one.

Negative proofs for current implementation:
shli: https://alive2.llvm.org/ce/z/_Q8ZuC
shrui: https://alive2.llvm.org/ce/z/vDZeEV
shrsi: https://alive2.llvm.org/ce/z/vZVatx

Proofs for attempted fix in this PR:
shli: https://alive2.llvm.org/ce/z/SDJJXW
shrui: https://alive2.llvm.org/ce/z/FZcP5y
shrsi: https://alive2.llvm.org/ce/z/5nk86z

### DIFF
--- a/mlir/lib/Dialect/Arith/Transforms/EmulateWideInt.cpp
+++ b/mlir/lib/Dialect/Arith/Transforms/EmulateWideInt.cpp
@@ -630,6 +630,18 @@ struct ConvertSelect final : OpConversionPattern<arith::SelectOp> {
 // ConvertShLI
 //===----------------------------------------------------------------------===//
 
+/// Math adopted from RISCVISelLowering for `ISD::SHL_PARTS`.
+//
+// if (rhsElem0 - newBitWidth) < 0: // rhsElem0 < newBitWidth
+//   lo = lhsElem0 << rhsElem0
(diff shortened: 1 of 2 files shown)


veera-efficient:
friendly ping @kuhar

kuhar (commented):
Why not keep the existing lowering and add an extra `select(cmp(amount, 0), input, result)` at the end?

veera-efficient:
@kuhar Sure, that will work but it will use more ops than the proposed fix.

I think this'd be better since this pass might be used for embedded systems where having a small code size is highly beneficial.

kuhar (approved):

veera-efficient:
Thanks. Merging this with my personal account.

veera-efficient:
Oops, will fix buildbot failures in a separate PR.

joker-eph:
Either the fix is trivial and you can land it immediately, or we should revert to unbreak the bots and you can take your time to reland with a fix for the failing tests.

PR #198458, opened 2026-05-19
Fix unused parameter for add_bitcode_entrypoint_library for GPU Libc

jinge90:
The cmake function add_bitcode_entrypoint_library accepts 'base_target_name' as parameter but never uses it. This function is to build libm/libc bitcode for GPU target which is sub-target for top-level libc and libm target, the libc/libc target should depend on libcbitcode/libmbitcode. This PR add dependenceis on libc/libmbitcode for libc/libm target.
If we enable libc/libm for GPU target, 'ninja libm' will only build libm.a for GPU but doesn't  build libm.bc without this PR.
This PR also fixes a typo in comments.


jinge90:
Hi, @jhuber6 
Could you help review this patch?
Thanks very much.

jhuber6 (approved):

PR #198460, opened 2026-05-19
[DAG] foldAddToAvg - add patterns to form avgceil(A, B) from (A & B) + (((A ^ B) + 1) >> 1)

heiher:


wangleiat (commented):
LGTM from the LoongArch side, thanks.

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
```diff
@@ -3207,6 +3207,7 @@ SDValue DAGCombiner::visitADDLike(SDNode *N) {
 
 // Attempt to form avgfloor(A, B) from (A & B) + ((A ^ B) >> 1)
 // Attempt to form avgfloor(A, B) from ((A >> 1) + (B >> 1)) + (A & B & 1)
+// Attempt to form avgceil(A, B) from (A & B) + (((A ^ B) + 1) >> 1)
```
jayfoad, line 3210: Needs an Alive2 proof, and I am guessing the proof will show that this does not work when `A ^ B == -1`.

heiher: Oops, they're not actually equivalent. 😢 Thanks!

PR #198461, opened 2026-05-19
[GlobalISel] Replace `GIM_CheckFeatures` with `GIM_Try_CheckFeatures`

Pierre-vh:
This has two benefits. First, it slightly reduces the size of the match table.
Secondly, if the target feature is not present, we can directly go to the fail
case, instead of having to do another loop of the interpreter, speeding up
rejection of rules that do not have the required target feature.

We could do the same with CheckSimplePredicate but it's less used (only in the combiners I think)
so it is less of a priority.

```
FILE                                          OLD      NEW    DIFF%    SAME?
----                                      -------  -------    -----    -----
AArch64GenGlobalISel.inc                   192681   185938    -3,5%       no
AArch64GenPostLegalizeGICombiner.inc         4457     4457     0,0%      yes
AArch64GenPreLegalizeGICombiner.inc          9010     9009    -0,0%       no
AMDGPUGenGlobalISel.inc                    596402   568486    -4,7%       no
AMDGPUGenPostLegalizeGICombiner.inc          9240     9238    -0,0%       no
AMDGPUGenPreLegalizeGICombiner.inc           8987     8987     0,0%      yes
AMDGPUGenRegBankGICombiner.inc               1981     1980    -0,1%       no
```

### DIFF
--- a/llvm/include/llvm/CodeGen/GlobalISel/GIMatchTableExecutorImpl.h
+++ b/llvm/include/llvm/CodeGen/GlobalISel/GIMatchTableExecutorImpl.h
@@ -155,7 +155,25 @@ bool GIMatchTableExecutor::executeMatchTable(
       OnFailResumeAt.push_back(readU32());
       break;
     }
-
+    case GIM_Try_CheckFeatures: {
+      // This is optimized so that if the feature is not present, we don't even
+      // modify OnFailResumeAt. Instead we directly jump to OnFail.
+      unsigned OnFail = readU32();
+      uint16_t ExpectedBitsetID = readU16();
+      DEBUG_WITH_TYPE(TgtExecutor::getName(),
+                      dbgs() << CurrentIdx
+                             << ": GIM_Try_CheckFeatures(ExpectedBitsetID="
+                             << ExpectedBitsetID << ")\n");
+      if ((AvailableFeatures & ExecInfo.FeatureBitsets[ExpectedBitsetID]) !=
+          ExecInfo.FeatureBitsets[ExpectedBitsetID]) {
+        DEBUG_WITH_TYPE(TgtExecutor::getName(),
+                        dbgs() << CurrentIdx
+                               << ": Features do not match, rejected\n");
+        CurrentIdx = OnFail;
+      } else
+        OnFailResumeAt.push_back(OnFail);
+      break;
+    }
     case GIM_RecordInsn:
     case GIM_RecordInsnIgnoreCopies: {
       uint64_t NewInsnID = readULEB();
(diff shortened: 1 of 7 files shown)


Pierre-vh:
* **#198461** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/198461?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/198461?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/llvm?ref=merge-queue-instructions-in-stack-comment&prId=198461).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

llvm/include/llvm/CodeGen/GlobalISel/GIMatchTableExecutorImpl.h
```diff
@@ -155,7 +155,25 @@ bool GIMatchTableExecutor::executeMatchTable(
...
+                             << ": GIM_Try_CheckFeatures(ExpectedBitsetID="
+                             << ExpectedBitsetID << ")\n");
+      if ((AvailableFeatures & ExecInfo.FeatureBitsets[ExpectedBitsetID]) !=
+          ExecInfo.FeatureBitsets[ExpectedBitsetID]) {
+        DEBUG_WITH_TYPE(TgtExecutor::getName(),
+                        dbgs() << CurrentIdx
+                               << ": Features do not match, rejected\n");
+        CurrentIdx = OnFail;
+      } else
+        OnFailResumeAt.push_back(OnFail);
```
shiltian:
```suggestion
      } else {
        OnFailResumeAt.push_back(OnFail);
      }
```

shiltian (commented):
I like this! LGTM with a nit, but I'll leave the stamp to the other folks with more experience with this.

ritter-x2a:
Is there a reason why only `GIM_CheckFeatures` would benefit from this, and not the other `GIM_Check*` opcodes?

Pierre-vh:
> Is there a reason why only `GIM_CheckFeatures` would benefit from this, and not the other `GIM_Check*` opcodes?

We could add a lot of variants of `GIM_Try` overloaded for most predicates, sure, but it's a lot of effort. I did it for `GIM_CheckFeatures` because it's not a standalone predicate, it is always emitted by the `RuleMatcher` itself right after `GIM_Try`, so it's very easy to merge (and we can delete the non-Try version). It's a low-hanging fruit.

We can definitely do it at some point when the match table is already very well optimized and there is not much left to do other than adding compound predicates like this. I have a few more optimizations in the pipeline that I'd like to experiment with beforehand, tho.

ritter-x2a:
Is there a chance that some downstream user relies on the old `GIM_CheckFeatures`? If that is a concern, we could keep the old variant additionally to the the new variant instead of replacing it.

Pierre-vh:
> Is there a chance that some downstream user relies on the old `GIM_CheckFeatures`? If that is a concern, we could keep the old variant additionally to the the new variant instead of replacing it.

I highly doubt it, though it isn't impossible I guess ? But then there should really be upstream tests and some code to show it's necessary to keep upstream support for it

ritter-x2a (approved):
LGTM, I think the GIM opcodes are sufficiently far away from any stable API so that it's fine to replace `GIM_CheckFeatures`.

PR #198462, opened 2026-05-19
[mlir][linalg] Allow extra DPS inputs in convolution dim inference

Tameem-10xE:
As pointed out by @hanhanW in https://github.com/iree-org/iree/pull/24016#issuecomment-4444591355, this PR fixes a mismatch between `inferConvolutionDims` and the `LinalgConvolutionOpInterface` verifier. This is needed for quantized convolution ops, which can have extra scalar zero-point operands in addition to the input and filter operands.

### DIFF
--- a/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp
+++ b/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp
@@ -884,7 +884,7 @@ inferConvolutionDimsImpl(LinalgOp linalgOp,
 /// Returns a failure if `output_image` (and implicitly `filter_loop`) is empty.
 FailureOr<ConvolutionDimensions>
 mlir::linalg::inferConvolutionDims(LinalgOp linalgOp) {
-  if (linalgOp.getNumDpsInits() != 1 || linalgOp.getNumDpsInputs() != 2)
+  if (linalgOp.getNumDpsInits() != 1 || linalgOp.getNumDpsInputs() < 2)
     return failure();
 
   auto indexingMaps = linalgOp.getIndexingMapsArray();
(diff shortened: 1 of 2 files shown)


hanhanW (commented):
Can you add a test to https://github.com/llvm/llvm-project/blob/main/mlir/unittests/Dialect/Linalg/InferConvolutionDimsTest.cpp Thanks

Tameem-10xE:
Hi @hanhanW, sorry for the delay.. got stuck with something else this past week.

I’ve added a test case in `InferConvolutionDimsTest.cpp` for quantized convolution with scalar zero-point inputs. Could you please take another look and let me know if any additional test or modifications are required?
Thanks!

```
Test Command: ./tools/mlir/unittests/Dialect/Linalg/MLIRLinalgTests   --gtest_filter=InferConvolutionDimsTest.*
```

image

Tameem-10xE:
Hi @hanhanW, pinging again for a review.

Also, just added some additional code comments since I didn't mention before that I was using hard-coded zero-points for qconv test + now both zero-points are now non-zero (although the value does not effect the test)

hanhanW:
cc @banach-space who also contributes to linalg and convolution a lot. This is a mismatch fix that we identified in our work. Inlining what I found from the associated comment:

> However, I found that there is a bug in upstream. These convolution all implement LinalgConvolutionOpInterface, which are declared in yaml. The implementation of [inferConvolutionDims](https://github.com/llvm/llvm-project/blob/c501f3747a46c4716fcf76e31075e725b3660502/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp#L887-L888) is not aligned with the [verification](https://github.com/llvm/llvm-project/blob/c501f3747a46c4716fcf76e31075e725b3660502/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp#L923-L924), which is a bug to me. We should correct the `inferConvolutionDims` in upstream, then you'll be able to 

I'm going to approve it, just in case if you want to take a look. Mostly an FYI.

Tameem-10xE:
Updated Code (Fixed a clang formatting issue.)

dimp-pl:
Hi guys, checking in as a person interested in the downstream PR (iree-org/iree#24016) that's blocked by this PR.
@Tameem-10xE could you rebase this PR to resolve the conflict? I'll be happy to help if you don't have the time right now.
@hanhanW is there anything else you'd like to see done before you can approve this PR?

Tameem-10xE:
Hi @dimp-pl, thanks for the ping. The conflict (from #203323 adding tests to the same file) is resolved; no functional change to the fix.

While rebasing, since the testing approach in #203323 is better, I also updated my quantized-conv test to reuse its expectConvDimsEq function (this change is assisted by Claude Code).


image

egebeysel:
cc @phemashekar

PR #198463, opened 2026-05-19
[libc] Port remaining socket functions to syscall_wrappers

labath:
While in there:
- fix file headers to conform to latest standards
- add missing restrict qualifier to recvfrom

Assisted by Gemini.


libc/src/__support/OSUtil/linux/syscall_wrappers/recv.h
```diff
@@ -0,0 +1,42 @@
...
+#include <sys/syscall.h> // For syscall numbers
+
+namespace LIBC_NAMESPACE_DECL {
+namespace linux_syscalls {
+
+LIBC_INLINE ErrorOr<ssize_t> recv(int sockfd, void *buf, size_t len,
+                                  int flags) {
+#ifdef SYS_recv
+  return syscall_checked<ssize_t>(SYS_recv, sockfd, buf, len, flags);
+#elif defined(SYS_recvfrom)
```
kaladron: Optional: I'd list SYS_recvfrom first because it's the preferred long-term solution.

Did i386 ever had recv and not recvfrom?  (I'm wondering if the backup is even needed)

libc/src/sys/socket/linux/bind.cpp
```diff
@@ -1,32 +1,33 @@
...
 namespace LIBC_NAMESPACE_DECL {
 
 LLVM_LIBC_FUNCTION(int, bind,
                    (int socket, const struct sockaddr *address,
                     socklen_t address_len)) {
-  int ret = syscall_impl<int>(SYS_bind, socket, address, address_len);
-  if (ret < 0) {
-    libc_errno = -ret;
+  auto result = linux_syscalls::bind(socket, address, address_len);
+  if (!result.has_value()) {
```
kaladron, line 26: We should use LIBC_CRASH_ON_NULLPTR on the pointer arguments (for all functions)

kaladron (approved):
A couple minor things within otherwise LGTM

libc/src/sys/socket/linux/bind.cpp
labath: Are you sure about that? I understand how that makes sense for functions implemented inside libc, where deterministically crashing is better than.. you know... mining some bitcoin and hiring a hitman to kill you.

However, this function is not accessing the pointer and it's not UB to pass a null (or any other) pointer through the syscall boundary. The kernel will just return an EFAULT. This isn't specified by POSIX, but it is documented in the linux manpage and both glibc and musl work that way.

Granted, a nullptr here would most likely indicate a bug in user code, but I have also seen code which uses this property (syscall returning EFAULT) to check for validity of pointers (you can imagine a beefed up version of LIBC_CRASH_ON_NULLPTR, which uses a syscall to check the validity of a pointer and crashes if it gets EFAULT).

If it was up to me, I'd say that this should be used only in places where our own code accesses the pointers -- but I don't think there's anything like that here. recvmsg comes close (accesses the pointer to unpoison the data), but I wouldn't put it even there, as this code is only executed on the success path (which will not be triggered by an invalid pointer).

kaladron: If the kernel is set to return an error message, then not using is correct.  I more worry about the system being in an unknown state.

libc/src/__support/OSUtil/linux/syscall_wrappers/recv.h
labath: I think we can do that. x86 (both 32 and 64 bit) are actually one of the architectures which don't have the plain recv syscall (there is one in the socketcall version, but I guess they decided to drop it even earlier than they decided to drop socketcall. This would also kind of indicate that we do not need the fallback (and a brief excursion into history seems to confirm that).

I'm a bit sad that we will be zeroing two registers for nothing, but I guess that's unlikely to matter.

libc/src/sys/socket/linux/bind.cpp
labath: The kernel has to validate all data coming from the userspace. An it has to do a much more thorough job than checking a null pointer, as there are many ways to exploit failures there.

PR #198464, opened 2026-05-19
[AMDGPU] Diagnose unsupported fma_legacy/sudot4/sudot8 intrinsics on some subtargets

aobolensk:
Add proper diagnostics for `llvm.amdgcn.fma.legacy`, `llvm.amdgcn.sudot4` and `llvm.amdgcn.sudot8` on subtargets where they are unsupported


jayfoad:
What happens without your patch? Do you get a "failed to select" diagnostic or does it fail in some nastier way?

aobolensk:
> What happens without your patch? Do you get a "failed to select" diagnostic or does it fail in some nastier way?

I get a fatal error with stacktrace before the patch
```
LLVM ERROR: Cannot select: intrinsic %llvm.amdgcn.fma.legacy
```
same for others

qcolombet (approved):

qcolombet:
FWIW, I'm happy with the cannot select but since it looks like we give more useful error message for other intrinsics, it's good to be consistent here.

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
```diff
@@ -1235,6 +1241,19 @@ bool AMDGPUInstructionSelector::selectG_INTRINSIC(MachineInstr &I) const {
     return selectPermlaneSwapIntrin(I, IntrinsicID);
   case Intrinsic::amdgcn_wave_shuffle:
     return selectWaveShuffleIntrin(I);
+  case Intrinsic::amdgcn_fma_legacy:
+    if (!STI.hasGFX10_3Insts()) {
```
shiltian: Can we use feature check instead? https://llvm.org/docs/AMDGPU/DeveloperGuideline.html#prefer-feature-checks-over-generation-checks

llvm/lib/Target/AMDGPU/SIISelLowering.cpp
```diff
@@ -10771,6 +10747,15 @@ SDValue SITargetLowering::LowerINTRINSIC_WO_CHAIN(SDValue Op,
     if (Subtarget->getGeneration() >= AMDGPUSubtarget::VOLCANIC_ISLANDS)
       return emitRemovedIntrinsicError(DAG, DL, VT);
     return DAG.getNode(AMDGPUISD::RCP_LEGACY, DL, VT, Op.getOperand(1));
+  case Intrinsic::amdgcn_fma_legacy:
+    if (!Subtarget->hasGFX10_3Insts())
```
shiltian: same here

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
aobolensk: done, thanks

llvm/lib/Target/AMDGPU/SIISelLowering.cpp
aobolensk: Also done

llvm/lib/Target/AMDGPU/GCNSubtarget.h
```diff
@@ -721,6 +721,8 @@ class GCNSubtarget final : public AMDGPUGenSubtargetInfo,
 
   bool hasSubClampInsts() const { return hasGFX10_3Insts(); }
 
+  bool hasFmaLegacy32Insts() const { return hasGFX10_3Insts(); }
```
shiltian, line 724: Does it mean that we don't actually have a target feature for legacy FMA instructions?

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
```diff
@@ -74,6 +74,12 @@ static Register getWaveAddress(const MachineInstr *Def) {
              : Register();
 }
 
+static void diagnoseUnsupportedIntrinsic(const MachineInstr &I) {
+  const Function &F = I.getMF()->getFunction();
+  F.getContext().diagnose(DiagnosticInfoUnsupported(
+      F, "intrinsic not supported on subtarget", I.getDebugLoc(), DS_Error));
```
arsenm, line 80: We really need to get better infrastructure for this. Ideally we would be able to put required features on the intrinsic definition, and trigger this automatically in selection

llvm/lib/Target/AMDGPU/GCNSubtarget.h
arsenm: Also weren't these in gfx6 to start, and later removed?

aobolensk: I have not found any

aobolensk:
> FWIW, I'm happy with the cannot select but since it looks like we give more useful error message for other intrinsics, it's good to be consistent here.

I'll propose it as a separate patch, since it requires changing much more files than here

aobolensk:
> Also weren't these in gfx6 to start, and later removed?

I have found only V_MAD_LEGACY_F32 in https://github.com/llvm/llvm-project/commit/61beff020e0abd0cf6c217978263588c0b8981c6, no FMA there. But I might be wrong. I have found it only since gfx10

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
aobolensk: 1. in both GISel and SDAG? or only here? 
2. Should that be based on TableGen features? Or purely in C++ code is better?
3. Should that be here in this PR with ISel fatal errors fix?

arsenm: Yes, everywhere. This should be all automatic in tablegen from the intrinsic definition

PR #198466, opened 2026-05-19
[AMDGPU] Add fneg pattern for v4f16 and v4bf16

Shoreshen:


jayfoad (changes requested):
Needs tests

Shoreshen:
> Needs tests

Hi @jayfoad, this only add pattern for fneg. Without making them legal, there will be no testable case change....

jayfoad:
> this only add pattern for fneg. Without making them legal, there will be no testable case change....

Then why are you adding the patterns? Do this in the same PR that makes them legal, if you are planning to do that?

changpeng:
> > this only add pattern for fneg. Without making them legal, there will be no testable case change....
> 
> Then why are you adding the patterns? Do this in the same PR that makes them legal, if you are planning to do that?

I agree with Jay. You cannot just add patterns without tests. I suggest you try make fneg (maybe fabs as well) legal, which possibly selecting better code through patterns.

PR #198469, opened 2026-05-19
[AArch64] C1-Nano scheduling model refactor [NFC]

walkerkd:
Creates explicit definitions for each latency/throughput/resource combination and use the definitions in the instruction rule definitions.

Alhough this change touches most lines in the model, there is no functional change - all test cases are not affected by this change.

This makes the style of the C1-Nano scheduling model be similar to that used in the C1-Ultra / C1-Premium and is being done in preparation to including the work to support SME instructions that is currently being implemented on the C1-Ultra scheduling model

### DIFF
--- a/llvm/lib/Target/AArch64/AArch64SchedC1Nano.td
+++ b/llvm/lib/Target/AArch64/AArch64SchedC1Nano.td
@@ -7,12 +7,13 @@
 //===----------------------------------------------------------------------===//
 //
 // This file defines the machine model for the ARM C1-Nano processor.
+// Information is taken from the C1 Nano Core Software Optimization Guide
+//
+// https://developer.arm.com/documentation/109590/0001
 //
 //===----------------------------------------------------------------------===//
 
 // ===---------------------------------------------------------------------===//
-// The following definitions describe the per-operand machine model.
-// This works with MachineScheduler. See MCSchedModel.h for details.
 
 // C1-Nano machine model for scheduling and other instruction cost heuristics.
 def C1NanoModel : SchedMachineModel {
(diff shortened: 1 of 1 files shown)


llvm/lib/Target/AArch64/AArch64SchedC1Nano.td
```diff
@@ -95,66 +96,149 @@ def : WriteRes<WriteID64, [C1NanoUnitDiv]> {
...
-def C1NanoWrite_14cyc_1VMAC_1VALU_B : SchedWriteRes<[C1NanoUnitVALU, C1NanoUnitVALU, C1NanoUnitVMAC, C1NanoUnitVMAC]> {
+def C1NanoWrite_14c_2VMAC_2VALU : SchedWriteRes<[C1NanoUnitVALU, C1NanoUnitVALU, C1NanoUnitVMAC, C1NanoUnitVMAC]> {
   let Latency     = 14;
   let NumMicroOps = 2;
 }
 
-class C1NanoWrite_PAC_B <int lat> : SchedWriteRes<[C1NanoUnitPAC, C1NanoUnitB]> {
-  let Latency = lat;
-  let NumMicroOps = 2;
+def C1NanoWrite_1c_1ALU   : SchedWriteRes<[C1NanoUnitALU]> { let Latency = 1; }
```
nasherm, line 122: Nit: Ultra and Premium go for a `C1Ux` and `C1Px` prefix for resource diffs. I wonder if it's worth changing them to C1Ultra and C1Premium in this patch as well?

nasherm (approved):
LGTM. Just one comment to consider

llvm/lib/Target/AArch64/AArch64SchedC1Nano.td
walkerkd: I think if we want to make that change to the C1Ultra and C1Premiu then it would be best to do in it's own PR so that it can be a completely NFC change; which should hopefully make it easier to review that change on it's own.

Asher8118 (approved):
LGTM

PR #167347, opened 2025-11-10
[AMDGPU] Verify dominance when rewriting spills to registers

kerbowa:
When performing spill elimination in the AGPR copy rewrite pass it was possible to see spill reloads that were not dominated by any store. This caused invalid MIR to be generated where vreg uses were not dominated by defs. This patch adds a dominance check before rewriting spills.

PR #198472, opened 2026-05-19
[1/2][AMDGPU] Fixed crash due to virtual register defs not dominating uses

### LINKED ISSUE #196671: `AMDGPURewriteAGPRCopyMFMA` crashes with ` Virtual register defs don't dominate all uses`
https://godbolt.org/z/TfEsrefsq

fails with
```
*** Bad machine code: Virtual register defs don't dominate all uses. ***
- function:    mha_depth512_bfloat16_bfloat16_False_True_nqh32_nkvh8_dde99d9d
- v. register: %2032
LLVM ERROR: Found 1 machine code errors.
```

From my understanding, the problem is in how spills are handled in AMDGPURewriteAGPRCopyMFMA pass, but it will be better if AMD team can take a look and fix it.

dhruvachak:
Fixes https://github.com/llvm/llvm-project/issues/196671.

Fixes duplicate ROCM-24494, LCOMPILER-2224.

In Rewrite AGPR-Copy-MFMA pass, a spill reload may not have a dominating spill store. If such a slot is unspilled into a vreg, the elimination phase crashes because virtual register defs do not dominate all uses. This patch checks for that dominance property for all reloads and skips unspilling if such a condition is found.

This patch is adapted from 
https://github.com/llvm/llvm-project/pull/167347

Cherry-picked https://github.com/ROCm/llvm-project/commit/e5d02ddb
Authored by: Austin Kerbow <Austin.Kerbow@amd.com>

Instead of scanning through instructions within the reload basic block, LiveStacks are used to find a store within that basic block.

2 tests are added, the LL test is adapted from 
https://github.com/llvm/llvm-project/issues/196671 and the MIR test is generated from the LL file. The CHECK lines verify that lack of joint-dominance property is found.

PR https://github.com/llvm/llvm-project/pull/200126 builds on this patch by trying to insert an IMPLICIT_DEF so that
the unspill can proceed.

Co-authored by: Austin Kerbow <Austin.Kerbow@amd.com>


llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
```diff
@@ -0,0 +1,159 @@
...
+  %41 = tail call <16 x float> @llvm.amdgcn.mfma.f32.32x32x16.bf16(<8 x bfloat> zeroinitializer, <8 x bfloat> zeroinitializer, <16 x float> %26, i32 0, i32 0, i32 0)
+  %42 = tail call <16 x float> @llvm.amdgcn.mfma.f32.32x32x16.bf16(<8 x bfloat> zeroinitializer, <8 x bfloat> zeroinitializer, <16 x float> %27, i32 0, i32 0, i32 0)
+  %43 = tail call <16 x float> @llvm.amdgcn.mfma.f32.32x32x16.bf16(<8 x bfloat> zeroinitializer, <8 x bfloat> zeroinitializer, <16 x float> %28, i32 0, i32 0, i32 0)
+  %44 = tail call <16 x float> @llvm.amdgcn.mfma.f32.32x32x16.bf16(<8 x bfloat> zeroinitializer, <8 x bfloat> zeroinitializer, <16 x float> %29, i32 0, i32 0, i32 0)
+  %45 = tail call <16 x float> @llvm.amdgcn.mfma.f32.32x32x16.bf16(<8 x bfloat> zeroinitializer, <8 x bfloat> zeroinitializer, <16 x float> %30, i32 0, i32 0, i32 0)
+  %46 = tail call <16 x float> @llvm.amdgcn.mfma.f32.32x32x16.bf16(<8 x bfloat> zeroinitializer, <8 x bfloat> zeroinitializer, <16 x float> %31, i32 0, i32 0, i32 0)
+  %47 = tail call <16 x float> @llvm.amdgcn.mfma.f32.32x32x16.bf16(<8 x bfloat> zeroinitializer, <8 x bfloat> zeroinitializer, <16 x float> %32, i32 0, i32 0, i32 0)
```
shiltian: Use named variable pls. https://llvm.org/docs/AMDGPU/DeveloperGuideline.html#use-named-values

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
```diff
@@ -0,0 +1,159 @@
...
+  %55 = fmul <16 x float> %.sroa.1581182.3, zeroinitializer
+  %56 = fmul <16 x float> %.sroa.210.3, zeroinitializer
+  %57 = fmul <16 x float> %.sroa.262.3, zeroinitializer
+  %58 = fmul <16 x float> %.sroa.314.3, zeroinitializer
+  %59 = fmul <16 x float> %.sroa.366.3, zeroinitializer
+  %60 = fmul <16 x float> %.sroa.418.3, zeroinitializer
+  %61 = fmul <16 x float> %.sroa.470.3, zeroinitializer
+  %62 = fmul <16 x float> %.sroa.522.3, zeroinitializer
+  %63 = fmul <16 x float> %.sroa.574.3, zeroinitializer
+  %64 = fmul <16 x float> %.sroa.626.3, zeroinitializer
```
shiltian: Can we trim the test case further?

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
```diff
@@ -0,0 +1,159 @@
+; REQUIRES: asserts
+; RUN: llc -O3 -mtriple=amdgcn-amd-amdhsa -mcpu=gfx950 \
```
shiltian: Do we need `-O3` here? Does the default work?

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
```diff
@@ -0,0 +1,159 @@
...
+; RUN:   -stop-after=amdgpu-rewrite-agpr-copy-mfma \
+; RUN:   -debug-only=amdgpu-rewrite-agpr-copy-mfma -filetype=null %s 2>&1 \
```
shiltian: We probably don't need this `-debug-only` if it crashes.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -44,6 +45,12 @@ using namespace llvm;
 DEBUG_COUNTER(RewriteAGPRCopyMFMACounter, DEBUG_TYPE,
               "Controls which MFMA chains are rewritten to AGPR form");
 
+static cl::opt<unsigned> ImplicitDefScanLimit(
+    "amdgpu-mfma-vgpr-to-agpr-spill-scan-limit", cl::Hidden, cl::init(4),
```
shiltian: It doesn't seem like the right choice to me to set a limit here if we have functionality issue. What if the code is beyond the limit? Do we crash again?

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -546,6 +625,18 @@ void AMDGPURewriteAGPRCopyMFMAImpl::eliminateSpillsOfReassignedVGPRs() const {
...
       Register NewVReg = MRI.createVirtualRegister(RC);
 
+      // It is legal for a spill reload to not have a dominating spill store.
+      // But after un-spilling, the replacement vreg must have reaching defs
+      // on all paths. Ensure this condition by inserting IMPLICIT_DEFs if
+      // required. Bail out if segment starts cannot be resolved.
+      SmallVector<MachineInstr *, 4> SpillStores;
+      for (MachineInstr *MI : SpillReferences->second)
+        if (MI->mayStore())
+          SpillStores.push_back(MI);
```
shiltian:
```suggestion
      for (MachineInstr *MI : SpillReferences->second) {
        if (MI->mayStore())
          SpillStores.push_back(MI);
      }
```

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
```diff
@@ -0,0 +1,159 @@
...
+  %exitcond.not.i = icmp eq i64 %2, 0
+  br i1 %exitcond.not.i, label %._crit_edge.i.loopexit, label %1
+
+._crit_edge.i.loopexit:                           ; preds = %51
+  ret void
+}
+
+declare <16 x float> @llvm.amdgcn.mfma.f32.32x32x16.bf16(<8 x bfloat>, <8 x bfloat>, <16 x float>, i32 immarg, i32 immarg, i32 immarg)
+
+attributes #0 = { "amdgpu-flat-work-group-size"="1, 256" "target-cpu"="gfx950" }
```
shiltian: Trim the attribute pls.

https://llvm.org/docs/AMDGPU/DeveloperGuideline.html#trim-unnecessary-attributes-and-metadata

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
dhruvachak: It was taken from the reduced test case submitted as part of the issue but I can check whether it can be further reduced.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-joint-dom-mir.mir
```diff
@@ -0,0 +1,1775 @@
...
+name:            rewrite_vgpr_mfma_to_agpr_spill_implicit_def
+tracksRegLiveness: true
+liveins:
+  - { reg: '$sgpr4_sgpr5', virtual-reg: '%0' }
+body:             |
+  bb.0:
+    successors: %bb.1(0x80000000)
+    liveins: $sgpr4_sgpr5
+
+    %0:sgpr_64 = COPY killed $sgpr4_sgpr5
```
shiltian, line 31: This MIR test case is too large. My gut feeling is we can create a small test case with a vreg not dominating uses.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
dhruvachak: -O3 is indeed required, otherwise the current compiler takes a different path and does not crash. That's why I added the MIR test with no -O3 and fewer passes.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
dhruvachak: I wanted a test beyond "it does not crash". The debug-only option allows the CHECK against the message saying an implicit def was introduced.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
shiltian: I c. Hopefully the reducer can get to a point that -O3 would not be needed.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: It's not really a functionality issue, it is just saying that we won't scan beyond a certain limit. If the spill-store is not found, an IMPLICIT_DEF is introduced. If indeed a spill-store is present beyond the scan-limit, the implicit def is just redundant and cleaned up later automatically  -- there is no crash. So, a small limit means potentially redundant but harmless implicit def and a large limit means potentially more compile-time.

I do have a comment in the code to this effect.

shiltian: Oh I missed the comment since it is not where ImplicitDefScanLimit is used. Lol. Is this used by anyone as a tuning knob at this moment?

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-joint-dom-mir.mir
dhruvachak: Not sure. The idea is to trigger the AGPR rewrite and that happens only under certain conditions. See also the discussion in https://github.com/llvm/llvm-project/pull/167347

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: No one is using it, I just introduced it :) I thought it could be used later if someone wanted to experiment with it.

I could just remove the option and use 4 as a hardcoded constant with a comment, if that makes more sense.

shiltian: We have so many knobs in our backend that I don't really see them being reasonably useful anymore, but that's just my .02.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
arsenm: -O3 only switched to using GVN instead of EarlyCSE, it's more reliable if you can make whatever the IR diff is manifest in the input

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
arsenm: As a general rule do not add cl::opts

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -127,6 +134,15 @@ class AMDGPURewriteAGPRCopyMFMAImpl {
   void collectSpillIndexUses(ArrayRef<LiveInterval *> StackIntervals,
                              SpillReferenceMap &Map) const;
 
+  /// For each segment in the stack slot's LiveInterval whose start does not
+  /// correspond to a spill store, insert an IMPLICIT_DEF of \p NewVReg.
+  /// This handles a control flow path where there is no spill store dominating
+  /// a spill reload. Returns false if any segment start cannot be resolved
+  /// (caller should bail out and not unspill).
+  bool insertImplicitDefsForLiveSegments(LiveInterval &StackLI, int Slot,
```
arsenm: Can you rename this to make it clear it's about spill cases

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -476,6 +492,69 @@ void AMDGPURewriteAGPRCopyMFMAImpl::collectSpillIndexUses(
...
+      }
+      return false;
+    }
+
+    // Scan forward from the segment start looking for a spill store to this
+    // slot. LiveStacks liveness for a store can start a few instructions
+    // before the store itself, so we allow a small window controlled by
+    // ImplicitDefScanLimit. Redundant IMPLICIT_DEFs (e.g. because of a small
+    // scan limit) are harmless, since they will be cleaned up by downstream
+    // transformations.
```
arsenm: They probably won't be this late?

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -476,6 +492,69 @@ void AMDGPURewriteAGPRCopyMFMAImpl::collectSpillIndexUses(
...
+    MachineBasicBlock::iterator It = MI ? MI->getIterator() : MBB->begin();
+    bool FoundStore = false;
+    for (unsigned I = 0; I < ScanLimit && It != MBB->end(); ++I, ++It) {
+      if (StoreSet.count(&*It)) {
+        FoundStore = true;
+        break;
+      }
+    }
```
arsenm: I don't understand this scan. This isn't an optimization heuristic, you need to insert an IMPLICIT_DEF only if you are certain it is needed and not clobbering a live value

dhruvachak: You are right, there was a mistake in my logic. There are 2 issues: (1) Precisely determining whether the reaching def problem exists. (2) Precise placement of the implicit_def or just bailing out. 

I realize (1) is not possible without a CFG analysis similar to what is done in https://github.com/llvm/llvm-project/pull/167347. For (2), any preference whether we should just bail out or find an appropriate BB for the implicit def. It won't exactly be the nearest dominator node because of possible back-edges through which valid values may flow in, so it needs some more dataflow analysis. Given how rare this situation should be, I don't know that it matters either way, bailing out or an implicit_def. Any preference?

ruiling: Correctness-wise, we only need to make sure it does not clobber any live value. Inserting the implicit_def before the starting point does not break correctness (yes there is special case for block boundary).

To prove an input MIR needs implicit_def accurately is possible. But it needs going backward along the CFG to find whether there is a store to the stack slot. The way we choose here is an approximation to answer whether it is needed in a fast way.

In real case, the store to the stack slot might be a few instructions after the starting point of livestack's interval. That's why scanning in a small window could help inserting less implicit_def.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -476,6 +492,69 @@ void AMDGPURewriteAGPRCopyMFMAImpl::collectSpillIndexUses(
...
+    Register NewVReg) const {
+  const unsigned ScanLimit = ImplicitDefScanLimit;
+
+  SmallPtrSet<MachineInstr *, 4> StoreSet(SpillStores.begin(),
+                                          SpillStores.end());
+
+  SmallVector<MachineInstr *, 2> InsertedDefs;
+
+  for (const LiveRange::Segment &Seg : StackLI) {
+    MachineInstr *MI = LIS.getInstructionFromIndex(Seg.start);
```
ruiling: I think there are two cases we need to handle: the starting point points to some instruction, or the starting point points to a basic block. Do we have a third case?

If the starting point points to a basic block. Seems there are two cases.
1.) The spill instruction is inside the pointed block, but just a few instructions after the block start.
2.) the spill instruction is in some of its predecessors. For this case, we need to query whether the stack object was live in the predecessors. Inserting implicit_def in the predecessor without live-segment is the right way.
We need to check against the two cases to handle them correctly.

If the starting point points to an instruction, handle it as now should work.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
arsenm: It's not worth spending any effort to reduce the number of inserted implicit_defs; they are free

dhruvachak: One approach is to use the joint-dominance test introduced by https://github.com/llvm/llvm-project/pull/167347 and then try to insert an implicit_def. I am working on those changes, will update this PR soon.

ruiling: Things could be simplified if we don't care much about the number of implicit_def: Check the Liverange of the stack slot. If it points to an instruction which does not store to the stack slot, insert implicit_def before the starting point. If it points to block boundary, then check against the predecessors's first terminator. If the stack slot is not live at the terminator, insert implicit_def before the terminator.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
dhruvachak: Done.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
dhruvachak: Tried llvm-reduce on the reduced LL reported in the issue. Only the arch-attribute could be removed, so no other reduction.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
dhruvachak: Well, that's why the MIR test was generated from the LL file. That should have reasonable reproducibility, isn't it? Note that the MIR test does not need -O3.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
dhruvachak: Modified to now CHECK for the joint-dominance debug output.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: Option removed.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: Fixed.

dhruvachak:
I think LiveStacks are not going to help here. Consider a CFG where a diamond-structure is inside a loop. On one branch of the diamond, there is a spill store and on the other a spill reload. The data dependence in the program ensures that the branch with the spill store is taken before the one with the spill reload. So, when the spill reload is executed, a valid value flows in through the back edge. So, in order to avoid clobbering a live value, any IMPLICIT_DEF has to be inserted higher up in the dominator tree, gated by this back-edge. https://github.com/llvm/llvm-project/pull/200126 implements a solution for safe placement of this IMPLICIT_DEF based on CFG dataflow.

I split up the changes into 2 PRs. This PR bails out if a reachable def is not found for a reload. The 2nd one tries to place an IMPLICIT_DEF.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: > Correctness-wise, we only need to make sure it does not clobber any live value. Inserting the implicit_def before the starting point does not break correctness (yes there is special case for block boundary).

I think inserting the implicit_def before the starting point can break correctness. Consider a loop having a diamond-shaped CFG-structure inside it. If the diamond has a spill store on one branch and a spill reload on the other, the definition for the reload may be provided by a live value flowing through the back edge. Placing the implicit_def before the starting point obtained through LSS may clobber that live value. 

> 
> To prove an input MIR needs implicit_def accurately is possible. But it needs going backward along the CFG to find whether there is a store to the stack slot. 

I think the above is what is required. The current version of this PR uses the implementation of https://github.com/llvm/llvm-project/pull/167347 to find reloads that are not jointly dominated and bails out of the unspill.

> The way we choose here is an approximation to answer whether it is needed in a fast way.

I think the approximation may have correctness problems. That was the first version of this PR which I have now reverted. The 2nd PR https://github.com/llvm/llvm-project/pull/200126 in this stack tries to insert an IMPLICIT_DEF using dataflow analysis.
> 
> In real case, the store to the stack slot might be a few instructions after the starting point of livestack's interval. That's why scanning in a small window could help inserting less implicit_def.

ruiling: >I think inserting the implicit_def before the starting point can break correctness. Consider a loop having a diamond-shaped CFG-structure inside it. If the diamond has a spill store on one branch and a spill reload on the other, the definition for the reload may be provided by a live value flowing through the back edge. Placing the implicit_def before the starting point obtained through LSS may clobber that live value.

I don't see why inserting implicit_def before the instruction at the starting point could clobber the live value in the mentioned case. There is no live value before the starting point.

dhruvachak: Consider the following CFG, basically a diamond-structure inside a loop:

```
            Entry
              |
              v
            +---+
       +--->| H |        loop header
       |    +---+
       |     / \
       |    v   v
       |  +---+ +---+
       |  | S | | N |    S: spill store SS#0
       |  +---+ +---+    N: PROBLEMATIC reload of SS#0
       |    \   /
       |     v v
       |   +-------+
       +---| Latch |     back-edge to H
           +-------+

  Edges: Entry -> H ; H -> S ; H -> N ; S -> Latch ; N -> Latch ; Latch -> H
```
The spill slot live segments are, roughly speaking:
[store-idx in S, end(S)] U
[start(Latch), end(Latch)] U
[start(H), end(H)] U
[start(N), reload-idx in N]
So, if an IMPLICIT_DEF is added to any of the above segments other than the first one, it would clobber the live value, assuming the store has executed at least once in the loop. Note that the live value flows from S -> Latch -> H -> N.

arsenm: The IMPLICIT_DEF should go in the loop pre-header? The nearest common dominator of all of these points

dhruvachak: I think that's what https://github.com/llvm/llvm-project/pull/200126 implements?

ruiling: The right place for the implicit_def would be a.) right before the spill store in S. b.) right before the first terminator of the entry block (loop preheader). While for this case the implicit_def in the loop preheader seems enough, but inserting before the spill store in S does not break correctness. The value of the stack slot is actually dead through the positions [start of S, store-idx in S). So it could never clobber live value. Right?

arsenm: > The value of the stack slot is actually dead through the positions [start of S, store-idx in S). So it could never clobber live value. Right?

I think the problem is that after the first loop iteration, the stack contents could be a previously live value from an execution which took S branch. If the execution is `H -> S -> Latch -> H -> N -> Latch -> exit`, the ultimate live-out value from the loop needs to match the value in S. This needs to be rewritten as a phi with an undef input on loop entry

ruiling: > If the execution is H -> S -> Latch -> H -> N -> Latch -> exit, the ultimate live-out value from the loop needs to match the value in S.

Does inserting implicit_def in `S` before the `spill-store` change this? I think no.

> This needs to be rewritten as a phi with an undef input on loop entry

For this case, the implicit_def in loop preheader is definitely necessary. But the second implicit_def in S is just the side-effect of a simple algorithm, there should be ways to avoid this. But it does not affect correctness. We are in non-SSA form. No need to reconstruct phi.

arsenm: > We are in non-SSA form. No need to reconstruct phi.

This isn't really true. You're still supposed to maintain SSA value numbering in LiveIntervals

dhruvachak: > The right place for the implicit_def would be a.) right before the spill store in S. b.) right before the first terminator of the entry block (loop preheader). While for this case the implicit_def in the loop preheader seems enough, but inserting before the spill store in S does not break correctness. The value of the stack slot is actually dead through the positions [start of S, store-idx in S). So it could never clobber live value. Right?

I agree that an IMPLICIT_DEF right before the spill store in S won't clobber live value. But what's the point of that IMPLICIT_DEF? It's unnecessary, isn't it?

dhruvachak: > > We are in non-SSA form. No need to reconstruct phi.
> 
> This isn't really true. You're still supposed to maintain SSA value numbering in LiveIntervals

Please review the placement for IMPLICIT_DEF in the PR https://github.com/llvm/llvm-project/pull/200126. I think it is a general technique that works for all cases.

dhruvachak:
I have a logistical question. Currently, this is a 2-PR stack where this one bails out by checking for lack of joint-dominance. The 2nd PR https://github.com/llvm/llvm-project/pull/200126 builds on joint-dominance and CFG dataflow to find a placement point for IMPLICIT_DEF so that the unspill can proceed.

This first PR is essentially a cherry-pick of a commit that has been in the AMD-fork for a few months now and working correctly. Additional tests based on the issue https://github.com/llvm/llvm-project/issues/196671 have been added to this first PR. This PR will unblock this issue which otherwise produces a compiler crash. If it looks ok, can we get this first PR approved while we figure out the final solution for the IMPLICIT_DEF in https://github.com/llvm/llvm-project/pull/200126?

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: > > The value of the stack slot is actually dead through the positions [start of S, store-idx in S). So it could never clobber live value. Right?
> 
> I think the problem is that after the first loop iteration, the stack contents could be a previously live value from an execution which took S branch. If the execution is `H -> S -> Latch -> H -> N -> Latch -> exit`, the ultimate live-out value from the loop needs to match the value in S. This needs to be rewritten as a phi with an undef input on loop entry

Or find the nearest common dominator through which this live value does not flow. Then an IMPLICIT_DEF can be placed in that block. That's what https://github.com/llvm/llvm-project/pull/200126 implements. For this example CFG, this block would be the preheader.

ruiling: >You're still supposed to maintain SSA value numbering in LiveIntervals

I am not sure what `SSA value numbering` exactly means in the context of LiveInterval. But I don't think inserting an extra dead implicit_def would break anything in LiveInterval. 

>But what's the point of that IMPLICIT_DEF?

The discussion was about the correctness of that extra implicit_def. I don't know what you are asking for. It is just a left-over of a very simple implementation. It allows you to fix the issue with just very few lines and don't need the traversal along the CFG.

bcahoon:
> This first PR is essentially a cherry-pick of a commit that has been in the AMD-fork for a few months now and working correctly. Additional tests based on the issue #196671 have been added to this first PR. This PR will unblock this issue which otherwise produces a compiler crash. If it looks ok, can we get this first PR approved while we figure out the final solution for the IMPLICIT_DEF in #200126?

This sounds like a good idea as the compiler crash is blocking others. Any objections?

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: > > But what's the point of that IMPLICIT_DEF?
> 
> The discussion was about the correctness of that extra implicit_def. I don't know what you are asking for. It is just a left-over of a very simple implementation. It allows you to fix the issue with just very few lines and don't need the traversal along the CFG.

If you are saying that a correct placement of IMPLICIT_DEF can be computed based solely on LiveStacks, then I am missing something. So, my ask is: please sketch an implementation based on LiveStacks that works for all cases. 

My primary concern is about the segments which do not have any spill store. For my example CFG, the path from H -> N is one such segment. Clearly, any IMPLICIT_DEF in that path H -> N would clobber live value through the back-edge. If we try to move up the chain of dominator nodes to find a placement for IMPLICIT_DEF, that's not enough either. We have to find a block such that no live value from a spill store reaches that block. By the time this is implemented, you already have CFG-based dataflow. And that's exactly what my 2nd PR https://github.com/llvm/llvm-project/pull/200126 does. And then you don't need the LiveStacks-based implementation anymore.

ruiling: Let me repost my suggestion: `Check the starting point of the liverange of the stack slot. If it points to an instruction which does not store to the stack slot, insert implicit_def before the starting point. If it points to block boundary, then check against the predecessors's first terminator. If the stack slot is not live at the terminator, insert implicit_def before the terminator.` And for your case, when inspecting the live-segment `[start(H), end(H)]`, you need to check the predecessors. As there is no live value in `Entry`, the right insertion point of the implicit_def is before the terminator of `Entry`.

ruiling: I just realized that a live-segment might span across multiple blocks (sorry I missed that). So actually we need going backward through the CFG at block level. We need to start from each end-position of the live-segment, we get the block that contains that position, then check if the live-range of the stack slot was alive at the block start. If it is alive, go checking at the terminator of the predecessors. If the stack slot is not live there, insert implicit_def. otherwise repeat the check going backward the CFG. We may need some logic to avoid repeated traversal. Sounds reasonable to you? (I think in real case, the non-dominating case comes from phi-node, maybe we don't even need to handle case that the live-segment start from the middle of a block but does not have a stack store.)

arsenm: > I am not sure what `SSA value numbering` exactly means in the context of LiveInterval. But I don't think inserting an extra dead implicit_def would break anything in LiveInterval.

VNInfo is maintaining pseudoSSA value numbering

npanchen:
@dhruvachak @arsenm any concern to merge this ?

dhruvachak:
At this point, the technique to place an IMPLICIT_DEF is the subject of another PR https://github.com/llvm/llvm-project/pull/200126. So, I am moving the relevant discussion to that PR. Please let's continue that discussion on that PR.

dhruvachak:
> > This first PR is essentially a cherry-pick of a commit that has been in the AMD-fork for a few months now and working correctly. Additional tests based on the issue #196671 have been added to this first PR. This PR will unblock this issue which otherwise produces a compiler crash. If it looks ok, can we get this first PR approved while we figure out the final solution for the IMPLICIT_DEF in #200126?
> 
> This sounds like a good idea as the compiler crash is blocking others. Any objections?

I have not heard any objections. Can this PR be approved?

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -529,6 +532,82 @@ void AMDGPURewriteAGPRCopyMFMAImpl::eliminateSpillsOfReassignedVGPRs() const {
...
+      if (StoreBlocks.contains(MBB))
+        continue;
+
+      for (MachineBasicBlock *Succ : MBB->successors()) {
+        if (StoreFreeReachable.insert(Succ).second)
+          Worklist.push_back(Succ);
+      }
+    }
+
+    auto IsLoadJointlyDominatedByStores = [&](MachineInstr *LoadMI) -> bool {
```
arsenm: Make this a proper helper function instead of a lambda

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -529,6 +532,82 @@ void AMDGPURewriteAGPRCopyMFMAImpl::eliminateSpillsOfReassignedVGPRs() const {
     if (SpillReferences == SpillSlotReferences.end())
       continue;
 
+    // For each spill reload, every path from entry to the reload must pass
+    // through at least one spill store to the same stack slot.
+    SmallVector<MachineInstr *, 4> Stores, Loads;
+    Stores.reserve(SpillReferences->second.size());
+    Loads.reserve(SpillReferences->second.size());
```
arsenm: Don't see the point of filtering these into new lists of load/store. you can directly check this in SpillReferences below

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -529,6 +532,82 @@ void AMDGPURewriteAGPRCopyMFMAImpl::eliminateSpillsOfReassignedVGPRs() const {
...
+    if (StoreBlocks.empty()) {
+      LLVM_DEBUG(dbgs() << "Skipping " << printReg(Slot, &TRI)
+                        << ": no reachable stores\n");
+      continue;
+    }
+
+    // Compute blocks reachable from entry without passing through a store
+    // block.
+    SmallPtrSet<MachineBasicBlock *, 16> StoreFreeReachable;
+    SmallVector<MachineBasicBlock *, 16> Worklist;
```
arsenm:
```suggestion
    SmallVector<MachineBasicBlock *, 16> Worklist = { &EntryMBB };
```

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -529,6 +532,82 @@ void AMDGPURewriteAGPRCopyMFMAImpl::eliminateSpillsOfReassignedVGPRs() const {
...
+
+    MachineBasicBlock &EntryMBB = MF.front();
+    Worklist.push_back(&EntryMBB);
```
arsenm:
```suggestion
```

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-joint-dom.ll
```diff
@@ -0,0 +1,159 @@
...
+  %exitcond.not.i = icmp eq i64 %i, 0
+  br i1 %exitcond.not.i, label %._crit_edge.i.loopexit, label %bb
+
+._crit_edge.i.loopexit:                           ; preds = %bb49
+  ret void
+}
+
+declare <16 x float> @llvm.amdgcn.mfma.f32.32x32x16.bf16(<8 x bfloat>, <8 x bfloat>, <16 x float>, i32 immarg, i32 immarg, i32 immarg)
+
+attributes #0 = { "amdgpu-flat-work-group-size"="1, 256" }
```
arsenm:
```suggestion
attributes #0 = { "amdgpu-flat-work-group-size"="1,256" }
```
I don't think this actually parses with extra spaces

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-joint-dom-mir.mir
```diff
@@ -0,0 +1,1775 @@
...
+  define amdgpu_kernel void @rewrite_vgpr_mfma_to_agpr_spill_joint_dom(i1 %arg, <16 x float> %.sroa.366.2) #0 {
+    %kernarg.segment = call ptr addrspace(4) @llvm.amdgcn.kernarg.segment.ptr()
+    %.kernarg.offset.align.down66 = bitcast ptr addrspace(4) %kernarg.segment to ptr addrspace(4)
+    %.sroa.366.2.kernarg.offset = getelementptr inbounds i8, ptr addrspace(4) %kernarg.segment, i64 64
+    unreachable
+  }
+
+  declare ptr addrspace(4) @llvm.amdgcn.kernarg.segment.ptr()
+
+  attributes #0 = { "amdgpu-flat-work-group-size"="1, 256" }
```
arsenm:
```suggestion
  attributes #0 = { "amdgpu-flat-work-group-size"="1,256" }
```

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-joint-dom-mir.mir
```diff
@@ -0,0 +1,1775 @@
...
+    %7.sub7:av_512_align2 = COPY %6
+    %7.sub8:av_512_align2 = COPY %6
+    %7.sub9:av_512_align2 = COPY %6
+    %7.sub10:av_512_align2 = COPY %6
+    %7.sub11:av_512_align2 = COPY %6
+    %7.sub12:av_512_align2 = COPY %6
+    %7.sub13:av_512_align2 = COPY %6
+    %7.sub14:av_512_align2 = COPY %6
+    %7.sub15:av_512_align2 = COPY %6
+    early-clobber %8:sgpr_512 = S_LOAD_DWORDX16_IMM_ec killed %0, 64, 0 :: (dereferenceable invariant load (s512) from %ir..sroa.366.2.kernarg.offset, align 16, addrspace 4)
```
arsenm: Can you avoid the IR references and drop the instructions in the IR section

arsenm (commented):
I'm a confused about why #200126 is only additive, and why this isn't more directly computed from the LiveIntervals/LiveStacks. You shouldn't really need to research through blocks for the frame index references? If #200126 is finding where to put the IMPLICIT_DEF, then why do we need all of this code to skip this scenario?

dhruvachak:
> I'm a confused about why #200126 is only additive, and why this isn't more directly computed from the LiveIntervals/LiveStacks. You shouldn't really need to research through blocks for the frame index references? If #200126 is finding where to put the IMPLICIT_DEF, then why do we need all of this code to skip this scenario?

If you are asking why I split up the work into 2 PRs, the reason is mostly about users waiting for this crash to be fixed. I agree that PR#1 bails out while PR#2 tries to proceed with the unspill, so that way they are not completely aligned. But PR#2 builds on top of the first, essentially taking off where the first left off. 

I felt that more discussion would be required for PR#2 to converge but we could perhaps agree on PR#1 faster.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
ruiling: Hi @arsenm, I guess I see your point. Your point is the LiveInterval needs to be reconstructed as pseudoSSA value numbering. My point is how to insert implicit_def to achieve it. Looking at the liveStack of the problematic stack slot: `SS#2 [2712r,4016B:0)[4240r,4576B:0)[16560r,24464r:0) 0@x  weight:0.000000e+00 [VReg_512_Align2]`. The first entry `[2712r,4016B:0)` points to the old `implicit_def` for the spilled register. The live stack has inherited all the live segments from the spilled register just without the value-numbering. If we can assume the LiveStack would inherit the segments from LiveInterval like the case here. We only need to insert implicit_def before the starting point if it does not point to stack store to that slot.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
dhruvachak: Done.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: Removed in this version.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: Done.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: Done.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: Done.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: Done.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-joint-dom.ll
dhruvachak: It actually parses with the extra space. Anyways, removed now.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-joint-dom-mir.mir
dhruvachak: Done.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-joint-dom-mir.mir
dhruvachak: Done.

dhruvachak:
> # 🐧 Linux x64 Test Results
> * 175419 tests passed
> * 3467 tests skipped
> * 1 test failed
> 
> ## Failed Tests
> (click on a test name to see its output)
> 
> ### LLVM
> LLVM.CodeGen/AMDGPU/bitop3-shared-operand.ll
> ```
> Exit Code: 1
> 
> Command Output (stdout):
> --
> # RUN: at line 2
> /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/llc -global-isel=0 -mtriple=amdgcn-- -mcpu=gfx950 < /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/CodeGen/AMDGPU/bitop3-shared-operand.ll | /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/FileCheck -check-prefixes=GCN,SDAG /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/CodeGen/AMDGPU/bitop3-shared-operand.ll
> # executed command: /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/llc -global-isel=0 -mtriple=amdgcn-- -mcpu=gfx950
> # note: command had no output on stdout or stderr
> # executed command: /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/FileCheck -check-prefixes=GCN,SDAG /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/CodeGen/AMDGPU/bitop3-shared-operand.ll
> # note: command had no output on stdout or stderr
> # RUN: at line 3
> /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/llc -global-isel -new-reg-bank-select -mtriple=amdgcn-- -mcpu=gfx950 < /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/CodeGen/AMDGPU/bitop3-shared-operand.ll | /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/FileCheck -check-prefixes=GCN,GISEL /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/CodeGen/AMDGPU/bitop3-shared-operand.ll
> # executed command: /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/llc -global-isel -new-reg-bank-select -mtriple=amdgcn-- -mcpu=gfx950
> # note: command had no output on stdout or stderr
> # executed command: /home/gha/actions-runner/_work/llvm-project/llvm-project/build/bin/FileCheck -check-prefixes=GCN,GISEL /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/CodeGen/AMDGPU/bitop3-shared-operand.ll
> # | /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/CodeGen/AMDGPU/bitop3-shared-operand.ll:518:15: error: GISEL-NEXT: expected string not found in input
> # | ; GISEL-NEXT: v_bitop3_b32 v0, v1, v2, v0 bitop3:0xf8
> # |               ^
> # | <stdin>:898:26: note: scanning from here
> # |  v_and_b32_e32 v3, v2, v0
> # |                          ^
> # | <stdin>:899:2: note: possible intended match here
> # |  v_bitop3_b32 v0, v1, v2, v0 bitop3:0x78
> # |  ^
> # | 
> # | Input file: <stdin>
> # | Check file: /home/gha/actions-runner/_work/llvm-project/llvm-project/llvm/test/CodeGen/AMDGPU/bitop3-shared-operand.ll
> # | 
> # | -dump-input=help explains the following input dump.
> # | 
> # | Input was:
> # | <<<<<<
> # |             .
> # |             .
> # |             .
> # |           893:  .globl bitop3_bxort_or_t_and_not_t ; -- Begin function bitop3_bxort_or_t_and_not_t 
> # |           894:  .p2align 8 
> # |           895:  .type bitop3_bxort_or_t_and_not_t,@function 
> # |           896: bitop3_bxort_or_t_and_not_t: ; @bitop3_bxort_or_t_and_not_t 
> # |           897: ; %bb.0: 
> # |           898:  v_and_b32_e32 v3, v2, v0 
> # | next:518'0                              X error: no match found
> # |           899:  v_bitop3_b32 v0, v1, v2, v0 bitop3:0x78 
> # | next:518'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> # | next:518'1      ?                                        possible intended match
> # |           900:  v_bfi_b32 v0, v3, 0, v0 
> # | next:518'0     ~~~~~~~~~~~~~~~~~~~~~~~~~
> # |           901:  ; return to shader part epilog 
> # | next:518'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> # |           902: .Lfunc_end13: 
> # | next:518'0     ~~~~~~~~~~~~~~
> # |           903:  .size bitop3_bxort_or_t_and_not_t, .Lfunc_end13-bitop3_bxort_or_t_and_not_t 
> # | next:518'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> # |           904:  ; -- End function 
> # | next:518'0     ~~~~~~~~~~~~~~~~~~~
> # |             .
> # |             .
> # |             .
> # | >>>>>>
> 
> --
> ```
> 

New test, probably needs a rebase.

dhruvachak:
@arsenm I addressed your comments. Please let me know any other concerns on this PR.

qcolombet:
> In Rewrite AGPR-Copy-MFMA pass, a spill reload may not have a dominating spill store

That sounds plain wrong. What does create this pattern?
I'm late to the party please point me to the relevant part of the conversation if it was already discussed.

arsenm:
> If you are asking why I split up the work into 2 PRs,

I'm more asking why the second PR isn't deleting the code from this PR

shiltian:
> > In Rewrite AGPR-Copy-MFMA pass, a spill reload may not have a dominating spill store
> 
> That sounds plain wrong. What does create this pattern? I'm late to the party please point me to the relevant part of the conversation if it was already discussed.

IIRC by design a spill store is not emitted if the value is undef in that path by design, but in this pass, it assumes that a reload will have a dominating store.

dhruvachak:
> > In Rewrite AGPR-Copy-MFMA pass, a spill reload may not have a dominating spill store
> 
> That sounds plain wrong. What does create this pattern? I'm late to the party please point me to the relevant part of the conversation if it was already discussed.

It can happen if there is a diamond-shaped conditional inside a loop, where there is a spill store on one branch and a reload on the other. That is what the test case attached to this PR has. Data dependence guarantees that the branch with the spill store is executed at least once before the reload. But the spill slots are not in SSA form, so after the AGPR-rewrite, findReachingDefs crashes because it does not find a dominating store. 

Please see the prior discussion with the CFG illustration below.

https://github.com/llvm/llvm-project/pull/198472#discussion_r3343066269

dhruvachak:
> > > In Rewrite AGPR-Copy-MFMA pass, a spill reload may not have a dominating spill store
> > 
> > 
> > That sounds plain wrong. What does create this pattern? I'm late to the party please point me to the relevant part of the conversation if it was already discussed.
> 
> It can happen if there is a diamond-shaped conditional inside a loop, where there is a spill store on one branch and a reload on the other. That is what the test case attached to this PR has. Data dependence guarantees that the branch with the spill store is executed at least once before the reload. But the spill slots are not in SSA form, so after the AGPR-rewrite, findReachingDefs crashes because it does not find a dominating store.
> 
> Please see the prior discussion with the CFG illustration below.
> 
> [#198472 (comment)](https://github.com/llvm/llvm-project/pull/198472#discussion_r3343066269)

@qcolombet Does this address your question?

dhruvachak:
> > If you are asking why I split up the work into 2 PRs,
> 
> I'm more asking why the second PR isn't deleting the code from this PR

The 2nd PR builds on the `isLoadJointlyDominatedByStores` utility provided by the first PR. Are you saying this is not required any more? I don't see how we can just rely on the LiveStacks segments alone and detect the crashing situation. 

I think I can simplify the placement implementation of the 2nd PR by using LiveStacks but I would still need the `isLoadJointlyDominatedByStores` utility for detecting the crashing scenario. That's why the 2nd PR isn't deleting much from the first PR.

dhruvachak:
Rebased.

dhruvachak:
ping. Any other concerns?

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -476,6 +488,34 @@ void AMDGPURewriteAGPRCopyMFMAImpl::collectSpillIndexUses(
...
+    return true;
+
+  // Check if every path passed through a store block.
+  if (!StoreFreeReachable.contains(LoadMBB))
+    return true;
+
+  // Otherwise, there exists a path to this block that has not seen any store
+  // yet. We must ensure that within this block there is a store to this slot
+  // before the load.
+  for (const MachineInstr &MI : *LoadMBB) {
```
arsenm: You should be able to perform this check via LiveStacks without scanning through the instructions?

dhruvachak: Thanks for that insight. I made that change. Can you check to see that's what you meant?

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -476,6 +488,34 @@ void AMDGPURewriteAGPRCopyMFMAImpl::collectSpillIndexUses(
...
+    return true;
+
+  // Check if every path passed through a store block.
+  if (!StoreFreeReachable.contains(LoadMBB))
+    return true;
+
+  // Otherwise, there exists a path to this block that has not seen any store
+  // yet. We must ensure that within this block there is a store to this slot
+  // before the load. Consult the slot's LiveStacks interval rather than
+  // scanning instructions: if a store to the slot precedes the load in this
```
arsenm:
```suggestion
  // Otherwise, there exists a path to this block that has not seen any store
  // yet. We must ensure that within this block there is a store to this slot
  // before the load. Consult the slot's LiveStacks interval: if a store to the slot precedes the load in this
```

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -476,6 +488,34 @@ void AMDGPURewriteAGPRCopyMFMAImpl::collectSpillIndexUses(
...
+  // block start. If there is no such store, the segment begins at or before
+  // the block start. If a live-in segment has coalesced with an in-block
+  // segment, it would imply a reload before a store. That scenario would
+  // lead to a joint-dominance violation. If the load reads an undef value, the
+  // segment is null.
+  SlotIndex LoadIdx = LIS.getInstructionIndex(LoadMI);
+  SlotIndex BlockStart = LIS.getMBBStartIdx(LoadMBB);
+  const LiveRange::Segment *Seg =
+      SlotLI.getSegmentContaining(LoadIdx.getBaseIndex());
+  return Seg && Seg->start > BlockStart;
```
arsenm: I think you can do something like
```
return !LIS.isLiveInToMBB(LSS.getInterval(Slot), LoadMBB);
```

arsenm (commented):
Yes, that's basically what I meant by using Livestacks

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: ok, I made the change. I added a liveAt check to make sure any null segment because of a pure undef is handled.

dhruvachak:
@arsenm Any other comment?

dhruvachak:
ping. Any other comment?

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-implicit-def.ll
qcolombet: >  The debug-only option allows the CHECK against the message saying an implicit def was introduced.

Can we check that the IMPLICIT_DEF is in the MIR output instead?

dhruvachak: Since the time I made the comment about the IMPLICIT_DEF debug output, this PR has been split into two. This PR now bails out from unspilling if the joint-dominance test fails, avoiding the compiler crash. The second PR https://github.com/llvm/llvm-project/pull/200126 tries to proceed with the unspill by adding an IMPLICIT_DEF. I will look into checking for the IMPLICIT_DEF in the MIR output as part of the 2nd PR.

This PR now has the following CHECK line. 

; CHECK: Skipping ${{[a-zA-Z0-9_]+}}: some reachable load not jointly dominated by stores

dhruvachak:
ping. Any other comment?

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
```diff
@@ -529,6 +562,48 @@ void AMDGPURewriteAGPRCopyMFMAImpl::eliminateSpillsOfReassignedVGPRs() const {
...
+
+      for (MachineBasicBlock *Succ : MBB->successors()) {
+        if (StoreFreeReachable.insert(Succ).second)
+          Worklist.push_back(Succ);
+      }
+    }
+
+    // Every reachable reload must be jointly dominated by the slot's stores.
+    if (!llvm::all_of(SpillReferences->second, [&](const MachineInstr *MI) {
+          return !MI->mayLoad() ||
```
arsenm, line 598: Could probably tighten this condition to specifically spill opcodes

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-joint-dom.ll
```diff
@@ -0,0 +1,159 @@
+; REQUIRES: asserts
+; RUN: llc -O3 -mtriple=amdgcn-amd-amdhsa -mcpu=gfx950 \
```
arsenm, line 2:
```suggestion
; RUN: llc -O3 -mtriple=amdgpu9.50-amd-amdhsa \
```

arsenm (approved):

dhruvachak:
Rebased.

llvm/test/CodeGen/AMDGPU/rewrite-vgpr-mfma-to-agpr-spill-joint-dom.ll
dhruvachak: Done.

llvm/lib/Target/AMDGPU/AMDGPURewriteAGPRCopyMFMA.cpp
dhruvachak: Probably. But I note that existing code already uses mayStore(), so new instances of mayStore() and mayLoad() seem consistent. If required, this change can be done in a follow-up, skipping for now.

dhruvachak:
The Buildbot has detected a new failure on builder llvm-clang-x86_64-expensive-checks-ubuntu while building llvm.
https://lab.llvm.org/buildbot/#/builders/187/builds/22857

It's a test problem in the new MIR test. 

The error is

*** Bad machine code: Illegal physical register for instruction ***
- function:    rewrite_vgpr_mfma_to_agpr_spill_joint_dom
- basic block: %bb.4  (0x555e5d6d2830) [4576B;17568B)
- instruction: 16564B   SI_SPILL_AV512_SAVE %899:vreg_512_align2, %stack.2, $sp_reg, 0, implicit $exec :: (store (s512) into %stack.2, align 4, addrspace 5)
- operand 2:   $sp_reg
$sp_reg is not a SReg_32 register.

dhruvachak:
Fixed by https://github.com/llvm/llvm-project/pull/216866.

PR #198475, opened 2026-05-19
[AMDGPU] Tweak sched model latencies for VALU fast-forward cases

jayfoad:
Model VALU fast-forward cases by setting the latency of the dependency
edge to zero. This allows scheduling pairs like V_CMP/V_CNDMASK
adjacently without the usual VALU->VALU latency, which should give the
scheduler more freedom.

Note that the implementation is similar to shouldScheduleAdjacent but
the purpose is different:
- adjustSchedDependency is trying to increase the accuracy of the
  scheduling model. This is useful both pre- and post-RA.
- shouldScheduleAdjacent is trying to encourage allocation of VCC so
  that instructions can be shrunk to 32-bit encoding. This is only
  useful pre-RA.


llvm/lib/Target/AMDGPU/GCNSubtarget.cpp
```diff
@@ -810,6 +810,22 @@ void GCNSubtarget::adjustSchedDependency(
     Dep.setLatency(InstrInfo.getSchedModel().computeOperandLatency(
         DefI, DefOpIdx, UseI, UseOpIdx));
   }
+
+  // Handle "fast-forward" cases for VALU dependencies via a lanemask-like SGPR:
+  //   V_CMP* -> V_CNDMASK_B32
+  //   V_ADD/SUB with carryout -> V_ADD/SUB with carryin (on GFX10+)
```
jayfoad, line 871: I could not find great documentation on exactly which cases count as a fast-forward. In the GFX9 docs I only see VALU->V_CNDMASK mentioned. GFX10+ docs also mention the add/sub with carry cases.

llvm/test/CodeGen/AMDGPU/asyncmark-gfx12plus.ll
```diff
@@ -51,15 +51,15 @@ define void @interleaved_with_wave_barrier(ptr addrspace(1) %foo, ptr addrspace(
 ; GISEL-NEXT:    s_wait_kmcnt 0x0
 ; GISEL-NEXT:    v_dual_mov_b32 v8, v3 :: v_dual_mov_b32 v9, v4
 ; GISEL-NEXT:    v_dual_mov_b32 v4, v5 :: v_dual_mov_b32 v5, v6
-; GISEL-NEXT:    v_add_co_u32 v6, vcc_lo, 0x54, v0
 ; GISEL-NEXT:    v_add_nc_u32_e32 v3, 0x54, v2
+; GISEL-NEXT:    v_add_co_u32 v6, vcc_lo, 0x54, v0
+; GISEL-NEXT:    s_delay_alu instid0(VALU_DEP_1) | instskip(SKIP_4) | instid1(VALU_DEP_1)
```
jayfoad, line 56: We should also teach `AMDGPUInsertDelayAlu` about fast-forward cases so that it does not insert useless `s_delay_alu` instructions like this.

llvm/lib/Target/AMDGPU/GCNSubtarget.cpp
```diff
@@ -810,6 +810,22 @@ void GCNSubtarget::adjustSchedDependency(
...
+  switch (UseI->getOpcode()) {
+  case AMDGPU::V_ADDC_U32_e64:
+  case AMDGPU::V_SUBB_U32_e64:
+  case AMDGPU::V_SUBBREV_U32_e64:
+    if (getGeneration() < GFX10)
```
shiltian: Can we avoid generation check? https://llvm.org/docs/AMDGPU/DeveloperGuideline.html#prefer-feature-checks-over-generation-checks

perlfu (commented):
Skimmed ~50% of the test diffs and this seems reasonable to me.
Lots of churn, but not much to say, but maybe slightly reduced SGPR pressure in a few cases.

llvm/test/CodeGen/AMDGPU/llvm.round.ll
```diff
@@ -761,6 +751,8 @@ define amdgpu_kernel void @round_v8f32(ptr addrspace(1) %out, <8 x float> %in) #
 ; R600-NEXT:     ADD T0.X, T3.W, PV.W,
 ; R600-NEXT:     LSHR * T1.X, KC0[2].Y, literal.x,
 ; R600-NEXT:    2(2.802597e-45), 0(0.000000e+00)
+; R600-NEXT:     ADD_INT * T2.X, PS, literal.x,
+; R600-NEXT:    4(5.605194e-45), 0(0.000000e+00)
```
IgWod, line 755: I know nothing about R600, but I thought I'll just highlight this change in case it matters.

IgWod (commented):
I skimmed through a number of tests and as much as I can tell it all looks reasonable. Just one thing that caught my eye.

llvm/test/CodeGen/AMDGPU/llvm.round.ll
jayfoad: I also know nothing about R600 :(

llvm/lib/Target/AMDGPU/GCNSubtarget.cpp
```diff
@@ -830,6 +830,22 @@ void GCNSubtarget::adjustSchedDependency(
     Dep.setLatency(InstrInfo.getSchedModel().computeOperandLatency(
         DefI, DefOpIdx, UseI, UseOpIdx));
   }
+
+  // Handle "fast-forward" cases for VALU dependencies via a lanemask-like SGPR:
+  //   V_CMP* -> V_CNDMASK_B32
+  //   V_ADD/SUB with carryout -> V_ADD/SUB with carryin (on GFX10+)
+  switch (UseI->getOpcode()) {
+  case AMDGPU::V_ADDC_U32_e64:
```
arsenm: What about the VOP2 forms?

llvm/lib/Target/AMDGPU/GCNSubtarget.cpp
jayfoad: Done

llvm/lib/Target/AMDGPU/GCNSubtarget.cpp
jayfoad: Good point. Done. This makes the patch > 2x more disruptive.

jayfoad: There are also _dpp and _sdwa forms, but I'm not sure it's really worth handling those too? It is a pain to list them all and the returns would be diminishing. Is there some simple way to query the "base" form of an opcode independent of encoding?

jayfoad:
Cc @amd-relkhoul

llvm/test/CodeGen/AMDGPU/addrspacecast.ll
```diff
@@ -751,28 +751,28 @@ define <16 x ptr> @addrspacecast_v16p5_to_v16p0(<16 x ptr addrspace(5)> %ptr) {
...
 ; CI-NEXT:    v_cndmask_b32_e32 v36, 0, v3, vcc
+; CI-NEXT:    v_cmp_ne_u32_e64 s[4:5], -1, v4
 ; CI-NEXT:    v_cndmask_b32_e64 v48, 0, v4, s[4:5]
-; CI-NEXT:    v_cndmask_b32_e64 v37, 0, v5, s[6:7]
+; CI-NEXT:    v_cmp_ne_u32_e64 s[8:9], -1, v6
 ; CI-NEXT:    v_cndmask_b32_e64 v33, 0, v6, s[8:9]
+; CI-NEXT:    v_cmp_ne_u32_e64 s[10:11], -1, v7
 ; CI-NEXT:    v_cndmask_b32_e64 v38, 0, v7, s[10:11]
 ; CI-NEXT:    v_cmp_ne_u32_e64 s[12:13], -1, v8
-; CI-NEXT:    v_cmp_ne_u32_e64 s[14:15], -1, v9
```
perlfu, line 760: Seems like we should be seeing an RA benefit for these cases?

perlfu (commented):
Seems like we should progress with this if you are willing to rebase?

llvm/test/CodeGen/AMDGPU/addrspacecast.ll
perlfu: Specifically the new diff seems to keep individual SGPRs for each condition result; however, perhaps this is just a lack of pressure to start with?

llvm/lib/Target/AMDGPU/GCNSubtarget.cpp
IgWod: If you stick to VOP2 and VOP3 you could normalize one to another using `getVOPe64`/`getVOPe32`? I'm not aware of a general lookup (which doesn't mean it doesn't exist) other than the one specific to VOPD instruction.

jayfoad:
> Seems like we should progress with this if you are willing to rebase?

I've updated (merged not rebased) to latest main.

llvm/test/CodeGen/AMDGPU/addrspacecast.ll
jayfoad: No, for all of these v_cmp results there is a second use ~10 lines further down in another v_cndmask. So the lifetimes still overlap.

llvm/lib/Target/AMDGPU/GCNSubtarget.cpp
```diff
@@ -865,6 +865,42 @@ void GCNSubtarget::adjustSchedDependency(
     Dep.setLatency(InstrInfo.getSchedModel().computeOperandLatency(
         DefI, DefOpIdx, UseI, UseOpIdx));
   }
+
+  // Handle "fast-forward" cases for VALU dependencies via a lanemask-like SGPR:
```
arsenm, line 869: Don't we have nearly identical code in the scheduler DAG mutation?

jayfoad: Yes absolutely but I'm not sure how to common them, and anyway the purpose is subtly different:

- This patch avoids unnecessarily pushing lanemask producer/consumer pairs away from each other.
- `shouldScheduleAdjacent` (used in the MacroFusion DAG mutation) actively tries to pull lanemask producer/consumer pairs towards each other to increase the chance of being able to allocate the lanemask to VCC.

There will be a third place handling fast-forwards when #205481 lands.

PR #198477, opened 2026-05-19
[DAG] Fold INT_TO_FP( FP_TO_INT (x) ) to FTRUNC(X) 

jacob-crawley:
Extends the `foldFPToIntToFP` DAG Combine so that it can now be applied when `FTRUNC` has a custom lowering, and given that `INT_TO_FP (FP_TO_INT (X))` is not already legal.

On AArch64 targets with SVE, this change simplifies the codegen of `INT_TO_FP (FP_TO_INT (X))` conversions by making use of the `frintz` instruction.


llvm/test/CodeGen/AArch64/sve2p2-frintz.ll
```diff
@@ -0,0 +1,57 @@
...
+; CHECK-NEXT:    ptrue p0.s
+; CHECK-NEXT:    frint64z z0.s, p0/z, z0.s
+; CHECK-NEXT:    ret
+    %res = fptosi <vscale x 4 x float> %in to <vscale x 4 x i64>
+    %res2 = sitofp <vscale x 4 x i64> %res to <vscale x 4 x float>
+    ret <vscale x 4 x float> %res2
+}
+
+
+
```
MacDue: nit: remove extra newlines

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -21036,6 +21036,98 @@ static SDValue performFpToIntCombine(SDNode *N, SelectionDAG &DAG,
...
+  else if (IntBits == 64)
+    FrintOpc = Intrinsic::aarch64_neon_frint64z;
+  else
+    return SDValue();
+
+  auto IsNEONVT = [](EVT VT) {
+    return VT == MVT::v2f32 || VT == MVT::v4f32 || VT == MVT::v2f64;
+  };
+
+  if (!IsNEONVT(VT))
```
MacDue: nit:
```suggestion
  if (VT != MVT::v2f32 && VT != MVT::v4f32 && VT != MVT::v2f64)
```

llvm/lib/Target/AArch64/AArch64InstrInfo.td
```diff
@@ -5969,6 +5969,21 @@ let Predicates = [HasFRInt3264] in {
...
+  def : Pat<(v2f32 (sint_to_fp (v2i64 (fp_to_sint v2f32:$Rn)))),
+                      (FRINT64Zv2f32 V64:$Rn)>;
+  def : Pat<(v4f32 (sint_to_fp (v4i64 (fp_to_sint v4f32:$Rn)))),
+                      (FRINT64Zv4f32 V128:$Rn)>;
+  def : Pat<(v2f64 (sint_to_fp (v2i32 (fp_to_sint v2f64:$Rn)))),
+                      (FRINT32Zv2f64 V128:$Rn)>;
+  def : Pat<(v2f64 (sint_to_fp (v2i64 (fp_to_sint v2f64:$Rn)))),
+                      (FRINT64Zv2f64 V128:$Rn)>;
+}
+
```
MacDue: Untested/unneeded? No tests fail when this is removed (your combines seem to go directly to the intrinsics).

llvm/test/CodeGen/AArch64/sve2p2-frintz.ll
```diff
@@ -0,0 +1,57 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
+; RUN: llc -mtriple=aarch64-linux-gnu -mattr=+sve2p2 < %s | FileCheck %s
+
+; Check that fp -> int -> fp conversions can be
+; lowered to the SVE FRINT*Z instructions for
+; scalable vectors.
+
+; FRINT32Z
+
```
MacDue: This will crash if provided an illegal SVE vector type, e.g.:
```llvm
define <vscale x 8 x float> @frint32z_f32_i32_f32_nxv8i32(<vscale x 8 x float> %in) {
    %res = fptosi <vscale x 8 x float> %in to <vscale x 8 x i32>
    %res2 = sitofp <vscale x 8 x i32> %res to <vscale x 8 x float>
    ret <vscale x 8 x float> %res2
}
```

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -21036,6 +21036,98 @@ static SDValue performFpToIntCombine(SDNode *N, SelectionDAG &DAG,
...
+                    : FPSrc;
+
+  SDValue Res = DAG.getNode(FrintOpc, DL, ContainerVT, Pg, Val,
+                            DAG.getPOISON(ContainerVT), Flags);
+
+  return VT.isFixedLengthVector() ? convertFromScalableVector(DAG, VT, Res)
+                                  : Res;
+}
+
+// Combine (sint_to_fp (fp_to_sint (x) )) to FRINT*X instructions
```
Stylie777: nit
```suggestion
// Combine (sint_to_fp (fp_to_sint (x) )) to FRINT*Z instructions
```

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -21036,6 +21036,98 @@ static SDValue performFpToIntCombine(SDNode *N, SelectionDAG &DAG,
...
+  unsigned IntBits = Src.getValueType().getVectorElementType().getSizeInBits();
+
+  if (VT.isFixedLengthVector() && Subtarget->isNeonAvailable() &&
+      Subtarget->hasFRInt3264()) {
+    if (SDValue Res =
+            performFpToSIntToFPCombineNEON(DAG, DL, VT, IntBits, FPSrc))
+      return Res;
+  }
+
+  if (Subtarget->isSVEorStreamingSVEAvailable() && Subtarget->hasSVE2p2()) {
```
MacDue: It looks these instructions are also available (outside of streaming-mode) with SME2p2: (e.g. https://godbolt.org/z/WqcahGE5e), so I think this be:

```suggestion
  if (Subtarget->isSVEorStreamingSVEAvailable() && (Subtarget->hasSVE2p2() || Subtarget->hasSME2p2())) {
```

llvm/lib/Target/AArch64/AArch64InstrInfo.td
jacob-crawley: Agreed,  these can be removed now. I wrote those before implementing the combines so its not adding anything.

llvm/test/CodeGen/AArch64/sve2p2-frintz.ll
jacob-crawley: Good spot I was missing the `TLI.isTypeLegal(VT)` check for SVE cases. 

I've added tests with illegal SVE vectors for frint32z/frint64z.

dzaima:
In my testing, `frint32`/`frint64` produce `-0.0` for a `-0.0` input, whereas `fptosi`+`sitofp` must normalize it to `+0.0`.

On another note, for `fptosi`+`sitofp`, when negative zero isn't a concern and there's no intermediate `freeze`, `frint32`/`64` isn't necessary, plain `frintz` works, given that `fptosi` gives poison on overflow, and indeed this is already done for some cases - https://c.godbolt.org/z/3ndrPGzj7

jacob-crawley:
> In my testing, `frint32`/`frint64` produce `-0.0` for a `-0.0` input, whereas `fptosi`+`sitofp` must normalize it to `+0.0`.
> 
> On another note, for `fptosi`+`sitofp`, when negative zero isn't a concern and there's no intermediate `freeze`, `frint32`/`64` isn't necessary, plain `frintz` works, given that `fptosi` gives poison on overflow, and indeed this is already done for some cases - https://c.godbolt.org/z/3ndrPGzj7

Thanks @dzaima for picking up on this. 

I've modified the `performFpToSIntToFPCombine` function to use `frintz` only  when signed zeros can be ignored. 

The NEON element of the combine has been removed as it will already lower to `frintz` if no signed zeros is enabled.

MacDue (commented):
I think adding a new combine may be unnecessary here (sorry!). The existing combine that applies for NEON is `foldFPToIntToFP` (in DAGCombiner.cpp). It folds fptosi -> sitofp pairs to `ISD::FTRUNC`. The reason this does not kick in for SVE is it's gated on `TLI.isOperationLegal(ISD::FTRUNC, VT)`. This is fine for NEON (as the operations are marked Legal), for SVE FTRUNC is "Custom" (as we need to lower it to a masked operation), so this check fails. If we could change the check to `isOperationLegalOrCustom`, I think we'd get this combine for "free".

https://github.com/llvm/llvm-project/blob/38555dbd81ad2dcfc452d29151299d829095258b/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L19949-L19962

jacob-crawley:
> I think adding a new combine may be unnecessary here (sorry!). The existing combine that applies for NEON is `foldFPToIntToFP` (in DAGCombiner.cpp). It folds fptosi -> sitofp pairs to `ISD::FTRUNC`. The reason this does not kick in for SVE is it's gated on `TLI.isOperationLegal(ISD::FTRUNC, VT)`. This is fine for NEON (as the operations are marked Legal), for SVE FTRUNC is "Custom" (as we need to lower it to a masked operation), so this check fails. If we could change the check to `isOperationLegalOrCustom`, I think we'd get this combine for "free".
> 
> https://github.com/llvm/llvm-project/blob/38555dbd81ad2dcfc452d29151299d829095258b/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp#L19949-L19962

Thanks @MacDue this is correct. I've been able to achieve the same lowering to `frintz` by removing the new combine function and modifying `foldFPToIntToFP`.

MacDue:
:+1: It looks like this has also caused a change for the AMDGPU backend, so you'll need to update those checks too

RKSimon:
@jacob-crawley this isn't aarch64 specific - please can you update the patch description in generic terms

llvm/test/CodeGen/AMDGPU/fptoui_uitofp.ll
```diff
@@ -184,12 +184,27 @@ define amdgpu_kernel void @fptoui_f64_to_i16_to_f64(ptr addrspace(1) %out, doubl
...
+; GFX6-NEXT:    s_mov_b32 s6, -1
+; GFX6-NEXT:    s_mov_b32 s7, 0xf000
 ; GFX6-NEXT:    s_waitcnt lgkmcnt(0)
-; GFX6-NEXT:    v_cvt_u32_f64_e32 v0, s[2:3]
-; GFX6-NEXT:    s_mov_b32 s3, 0xf000
-; GFX6-NEXT:    s_mov_b32 s2, -1
-; GFX6-NEXT:    v_cvt_f64_u32_e32 v[0:1], v0
-; GFX6-NEXT:    buffer_store_dwordx2 v[0:1], off, s[0:3], 0
+; GFX6-NEXT:    s_bitset0_b32 s3, 31
+; GFX6-NEXT:    s_mov_b32 s4, s0
```
MacDue: Looks like AMDGPU has a much larger custom expansion for FTRUNC (for SVE a lot of "Custom" expansions simply add a `ptrue`). I wonder if this is okay, or if we need a new TLI hook? cc @arsenm 

It would be good if there was some mechanism to tell if "Custom" meant a trivial expansion (e.g., adding a single `ptrue`), or a large expansion.

arsenm: No new information is needed here. This probably should be skipping the combine if the individual cast operations are legal too

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
```diff
@@ -19959,13 +19959,23 @@ static SDValue foldFPToIntToFP(SDNode *N, const SDLoc &DL, SelectionDAG &DAG,
...
   assert(IsSigned || IsUnsigned);
 
+  // Don't fold if the individual cast operations are already legal,
+  // as FTRUNC may have a more expensive custom expansion.
+  EVT IntVT = N->getOperand(0).getValueType();
+  EVT LegalIntVT = TLI.getTypeToTransformTo(*DAG.getContext(), IntVT);
+  unsigned FPToIntOp = IsUnsigned ? ISD::FP_TO_UINT : ISD::FP_TO_SINT;
+  unsigned IntToFPOp = N->getOpcode(); // UINT_TO_FP or SINT_TO_FP
+  if (TLI.isOperationLegal(FPToIntOp, LegalIntVT) &&
+      TLI.isOperationLegal(IntToFPOp, VT))
```
MacDue: I think we'd only want to bail when `FTRUNC` is not legal, right? If it's legal, then `frunc` is an improvement over a pair of to/from int. 
```suggestion
  if (!TLI.isOperationLegal(ISD::FTRUNC, VT) && 
      TLI.isOperationLegal(FPToIntOp, LegalIntVT) &&
      TLI.isOperationLegal(IntToFPOp, VT))
```

MacDue (approved):
LGTM, but as noted before this is not an AArch64 specific change anymore, so the title/description should reflect that.

llvm/test/CodeGen/AArch64/sve-fixed-length-frintz.ll
```diff
@@ -0,0 +1,82 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
+; RUN: llc -mtriple=aarch64-linux-gnu --enable-no-signed-zeros-fp-math -mattr=+sve -aarch64-sve-vector-bits-min=256 < %s | FileCheck %s
```
arsenm:
```suggestion
; RUN: llc -mtriple=aarch64-linux --enable-no-signed-zeros-fp-math -mattr=+sve -aarch64-sve-vector-bits-min=256 < %s | FileCheck %s
```

llvm/test/CodeGen/AArch64/sve-fixed-length-frintz.ll
```diff
@@ -1,5 +1,5 @@
 ; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
-; RUN: llc -mtriple=aarch64-linux-gnu --enable-no-signed-zeros-fp-math -mattr=+sve -aarch64-sve-vector-bits-min=256 < %s | FileCheck %s
+; RUN: llc -mtriple=aarch64-linux --enable-no-signed-zeros-fp-math -mattr=+sve -aarch64-sve-vector-bits-min=256 < %s | FileCheck %s
```
arsenm: I think this needs something in the middle to be a canonical triple

llvm/test/CodeGen/AArch64/sve-frintz.ll
```diff
@@ -0,0 +1,126 @@
...
+; SIGNED-ZEROS-NEXT:    uunpklo z1.d, z0.s
+; SIGNED-ZEROS-NEXT:    uunpkhi z0.d, z0.s
+; SIGNED-ZEROS-NEXT:    ptrue p0.d
+; SIGNED-ZEROS-NEXT:    fcvtzs z1.d, p0/m, z1.s
+; SIGNED-ZEROS-NEXT:    fcvtzs z0.d, p0/m, z0.s
+; SIGNED-ZEROS-NEXT:    scvtf z0.s, p0/m, z0.d
+; SIGNED-ZEROS-NEXT:    scvtf z1.s, p0/m, z1.d
+; SIGNED-ZEROS-NEXT:    uzp1 z0.s, z1.s, z0.s
+; SIGNED-ZEROS-NEXT:    ret
+    %res = fptosi <vscale x 4 x float> %in to <vscale x 4 x i64>
```
arsenm: 2 space indent

arsenm (approved):

llvm/test/CodeGen/AArch64/sve-fixed-length-frintz.ll
jacob-crawley: Changed to `target triple = "aarch64-unknown-linux-gnu"` to be in line with the other SVE tests.

PR #198479, opened 2026-05-19
[clang][SYCL] Enable possibility for device-only calling convention

Fznamznon:
After 94ca49099ef77751a33e4babe41b2ae03ff228e1 DPC++ downstream experienced failures. This happened because it has a calling convention attribute that applies calling convention valid only for device targets which given SYCL model is quite unusual and was not expected by 94ca49099ef77751a33e4babe41b2ae03ff228e1 . Even though there is no device-only calling conventions, we should not prevent a possibility for it. So this patch preserves the semantics of the device over those of the host while also preserving device side diagnostics if they are less severe than the host.


erichkeane (approved):

tahonermann (approved):
Thanks @Fznamznon!

PR #198480, opened 2026-05-19
[Runtimes][CMake] Add dependency on compiler to source files

wenju-he:
Fix runtimes not rebuilt when clang is updated. Rebuild is critical for runtimes like libclc that produces bitcode-based builtin library, where compiler change should immediately trigger a re-compilation.

By default, CMake add_library does not add compiler executable as dependency for objects of source file.

Add a helper LLVMRuntimeCompilerDependency.cmake which exposes llvm_defer_compiler_dependency_tracking function. Each runtime can explicitly opt-in by calling the function at the beginning of its root CMakeLists to add dependency on compilers it uses.


jhuber6 (commented):
This is a pretty big change. I'm surprised that there's not already a way to signal that the compiler itself changed. Runtimes are just put through an external project so I would've expected CMake to handle an updated compiler natively.

PR #198482, opened 2026-05-19
[lldb] Fix possible invalidated iterator.

da-viper:
The begin or end interator may be invalidated when a idx_pos in erased from the vector. 

Unblocks sanitised CI.

### DIFF
--- a/lldb/source/Target/Target.cpp
+++ b/lldb/source/Target/Target.cpp
@@ -978,10 +978,10 @@ void Target::DescribeBreakpointOverrides(Stream &stream,
     return;
   }
 
-  const bool empty = idxs.empty();
+  bool empty = idxs.empty();
   bool print_first = true;
-  for (const auto &[id, resolver] : m_breakpoint_overrides) {
-    auto idx_pos = llvm::find(idxs, id);
+  for (auto const &elem : m_breakpoint_overrides) {
+    auto idx_pos = llvm::find(idxs, elem.first);
     if (empty || idx_pos != idxs.end()) {
       if (print_first) {
         // FIXME: Is there some good way to flow the description?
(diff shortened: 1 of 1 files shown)


lldb/source/Target/Target.cpp
```diff
@@ -977,20 +978,18 @@ void Target::DescribeBreakpointOverrides(Stream &stream,
...
+    auto idx_pos = llvm::find(idxs, id);
+    if (empty || idx_pos != idxs.end()) {
       if (print_first) {
         // FIXME: Is there some good way to flow the description?
         stream << "ID    Description\n";
         stream << "----  -----------\n";
         print_first = false;
       }
-      stream.Format("{0,4}  {1}\n", elem.first, elem.second->GetDescription());
+      stream.Format("{0,4}  {1}\n", id, resolver->GetDescription());
```
felipepiovezan: is this diff related to the change?

lldb/source/Target/Target.cpp
```diff
@@ -977,20 +978,18 @@ void Target::DescribeBreakpointOverrides(Stream &stream,
...
 
-  auto begin = idxs.begin();
-  auto end = idxs.end();
-  bool empty = idxs.empty();
+  const bool empty = idxs.empty();
   bool print_first = true;
-  for (auto const &elem : m_breakpoint_overrides) {
-    auto idx_pos = empty ? end : std::find(begin, end, elem.first);
-    if (empty || idx_pos != end) {
+  for (const auto &[id, resolver] : m_breakpoint_overrides) {
```
felipepiovezan: is this diff related to the change?

lldb/source/Target/Target.cpp
```diff
@@ -977,20 +978,18 @@ void Target::DescribeBreakpointOverrides(Stream &stream,
...
+    auto idx_pos = llvm::find(idxs, id);
+    if (empty || idx_pos != idxs.end()) {
       if (print_first) {
         // FIXME: Is there some good way to flow the description?
```
DavidSpickett, line 987: @jimingham off topic for this PR but since I noticed it - you might be able to use OutputWordWrappedLines. If you set the stream's indent level to be equal to where `---- ` ends.

felipepiovezan (approved):
This LGTM, but please in the future don't include NFC diffs in the PR. I spent quite some time trying to puzzle out if those were related or not. I really like the use of the structured binding, but it should have been its own PR.

PR #198487, opened 2026-05-19
[X86] Lower vector i8 ashr-by-1 using pavgb

### LINKED ISSUE #198061: [X86] Missed Optimization: Vector 8-bit `ashr(x, 1)` should be lowered as `pavgb(x, -1) ^ (~x & 0x80)`
Due to a lack of support, 8-bit arithmetic shifts are implemented using a 16-bit shift + MSB fill chain:

```asm
ashr1_src:
        psrlw   xmm0, 1
        pand    xmm0, xmmword ptr [rip + .LCPI0_0]
        movdqa  xmm1, xmmword ptr [rip + .LCPI0_1]
        pxor    xmm0, xmm1
        psubb   xmm0, xmm1
        ret
```

This can be simplified by using the `pavgb` instruction with AllOnes to shift right by 1 well setting the MSB, which can be conditionally removed if the input's isn't set. This shortens the dependency chain and avoids the shift, which has worse throughput than `pavgb` on some architectures.

```asm
ashr1_tgt:
        pcmpeqd xmm1, xmm1
        pavgb   xmm1, xmm0
        pandn   xmm0, xmmword ptr [rip + .LCPI1_0]
        pxor    xmm0, xmm1
        ret
```

https://godbolt.org/z/9PdYvcnsT

Harishankar14:
For vector i8 arithmetic shift right by 1, the current lowering produces a 5-instruction sequence (psrlw + pand + xor + psubb plus a constant load) with a 4-deep dependency chain.

This patch uses the identity

  ashr(x, 1) == avgceilu(x, -1) ^ (~x & 0x80)

to lower to ISD::AVGCEILU + a short fixup, producing 4 instructions on SSE/AVX/AVX2 and 3 on AVX-512BW (after vpternlogd fusion of the AND/XOR pair), with two parallel dependency chains instead of one long one.

The freeze on R is required because the target reads it twice, matching the pattern of the existing `shl R, 1 -> add R, R` case in LowerShiftByScalarImmediate.

Alive2 proof: https://alive2.llvm.org/ce/z/LbXPhE

Fixes #198061


Harishankar14:
cc @RKSimon

phoebewang (approved):
LGTM.

llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -31091,6 +31067,15 @@ static SDValue LowerShiftByScalarImmediate(SDValue Op, SelectionDAG &DAG,
       return DAG.getNode(ISD::AND, dl, VT, SRL, DAG.getConstant(Mask, dl, VT));
     }
     if (Op.getOpcode() == ISD::SRA) {
+      if (ShiftAmt == 1) {
```
RKSimon, line 31095: Add a summary comment like the others:
```
// ashr(R, 1) === xor(avgceilu(R,-1),and(not(R),MIN_SIGNED))
```

Harishankar14:
@RKSimon  Done

RKSimon (approved):
LGTM - cheers

PR #198488, opened 2026-05-19
[AArch64] Combine A + zext(B) * C -> A - zext(B) * -C

davemgreen:
This converts `A + zext(B) * C` -> `A - zext(B) * -C` if C can be negated in order to use umsubl.


igogo-x86 (approved):
LGTM

PR #198491, opened 2026-05-19
[AMDGPU] Do not combine V_ASHRREV_I16* to make sdwa

LU-JOHN:
Do not combine V_ASHRREV_I16* instructions to form sdwa instructions.  These instructions zero-fill the high 16-bits.

### DIFF
--- a/llvm/lib/Target/AMDGPU/SIPeepholeSDWA.cpp
+++ b/llvm/lib/Target/AMDGPU/SIPeepholeSDWA.cpp
@@ -399,6 +399,9 @@ MachineInstr *SDWASrcOperand::potentialToConvert(const SIInstrInfo *TII,
 }
 
 bool SDWASrcOperand::convertToSDWA(MachineInstr &MI, const SIInstrInfo *TII) {
+  assert((!Sext || !TII->getSubtarget().zeroesHigh16BitsOfDest(
+                       getParentInst()->getOpcode())) &&
+         "Cannot use sign-extension with instruction that zeroes high bits");
   switch (MI.getOpcode()) {
   case AMDGPU::V_CVT_F32_FP8_sdwa:
   case AMDGPU::V_CVT_F32_BF8_sdwa:
--- a/llvm/lib/Target/AMDGPU/SIPeepholeSDWA.cpp
+++ b/llvm/lib/Target/AMDGPU/SIPeepholeSDWA.cpp
@@ -708,18 +708,16 @@ SIPeepholeSDWA::matchSDWAOperand(MachineInstr &MI) {
   }
 
   case AMDGPU::V_LSHRREV_B16_e32:
-  case AMDGPU::V_ASHRREV_I16_e32:
   case AMDGPU::V_LSHLREV_B16_e32:
   case AMDGPU::V_LSHRREV_B16_e64:
   case AMDGPU::V_LSHRREV_B16_opsel_e64:
-  case AMDGPU::V_ASHRREV_I16_e64:
   case AMDGPU::V_LSHLREV_B16_opsel_e64:
   case AMDGPU::V_LSHLREV_B16_e64: {
+    // V_ASHRREV_I16_e32 and V_ASHRREV_I16_e64 are
+    // not included here because they zero-fill the high 16-bits.
+
     // from: v_lshrrev_b16_e32 v1, 8, v0
     // to SDWA src:v0 src_sel:BYTE_1
-
-    // from: v_ashrrev_i16_e32 v1, 8, v0
-    // to SDWA src:v0 src_sel:BYTE_1 sext:1
 
     // from: v_lshlrev_b16_e32 v1, 8, v0
     // to SDWA dst:v1 dst_sel:BYTE_1 dst_unused:UNUSED_PAD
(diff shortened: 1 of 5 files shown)


llvm/lib/Target/AMDGPU/SIPeepholeSDWA.cpp
```diff
@@ -708,19 +708,17 @@ SIPeepholeSDWA::matchSDWAOperand(MachineInstr &MI) {
...
   case AMDGPU::V_LSHRREV_B16_e32:
-  case AMDGPU::V_ASHRREV_I16_e32:
   case AMDGPU::V_LSHLREV_B16_e32:
   case AMDGPU::V_LSHRREV_B16_e64:
   case AMDGPU::V_LSHRREV_B16_opsel_e64:
-  case AMDGPU::V_ASHRREV_I16_e64:
   case AMDGPU::V_LSHLREV_B16_opsel_e64:
   case AMDGPU::V_LSHLREV_B16_e64: {
+    // V_ASHRREV_I16_e32 and V_ASHRREV_I16_e64 are
+    // not included here because they zero-fill the high 16-bits.
```
arsenm, line 720: I thought this changed in gfx9, and was controlled by the opsel bit. the gfx8 behavior was 0 high, then preserve high?

LU-JOHN: gfx8/gfx9 zero-fills the hi-half and use V_ASHRREV_I16_e32/V_ASHRREV_I16_e64.
gfx10 preserves the hi-half and uses  V_ASHRREV_I16_opsel_e64.
gfx11 preserves the hi-half and uses V_ASHRREV_I16_t16_e64.

V_ASHRREV_I16_e32/V_ASHRREV_I16_e64 are included in GCNSubtarget::zeroesHigh16BitsOfDest in GCNSubtarget.cpp.

LU-JOHN: ping

llvm/lib/Target/AMDGPU/SIPeepholeSDWA.cpp
frederik-h, line 401:
```suggestion
  assert((!Sext || !TII->getSubtarget().zeroesHigh16BitsOfDest(
                       getParentInst()->getOpcode())) &&
         "Cannot use sign-extension with instruction that zeroes high bits");
```
Perhaps add this assert? It would catch the previous broken state.

llvm/lib/Target/AMDGPU/SIPeepholeSDWA.cpp
```diff
@@ -708,19 +708,17 @@ SIPeepholeSDWA::matchSDWAOperand(MachineInstr &MI) {
   }
 
   case AMDGPU::V_LSHRREV_B16_e32:
-  case AMDGPU::V_ASHRREV_I16_e32:
   case AMDGPU::V_LSHLREV_B16_e32:
   case AMDGPU::V_LSHRREV_B16_e64:
   case AMDGPU::V_LSHRREV_B16_opsel_e64:
-  case AMDGPU::V_ASHRREV_I16_e64:
```
frederik-h, line 715: Should we add  V_ASHRREV_I16_opsel_e64?

frederik-h (changes requested):

llvm/lib/Target/AMDGPU/SIPeepholeSDWA.cpp
LU-JOHN: Added.

llvm/lib/Target/AMDGPU/SIPeepholeSDWA.cpp
LU-JOHN: I'd rather do this in another PR.  This PR is just for fixing a bug.

frederik-h: Sure. If we add it, we need to undo your change to the way the Sext value is being set, but it should be easy enough to do this later.

llvm/test/CodeGen/AMDGPU/sdwa-peephole-instr.mir
```diff
@@ -467,4 +467,26 @@ body:             |
     %7:vgpr_32 = nnan nofpexcept V_ADD_F32_e32 %5, %6, implicit $mode, implicit $exec
     S_ENDPGM 0, implicit %7
 
+# GCN-LABEL: name: negative_v_ashrrev_i16
+# GCN-NOT: V_XOR_B32_sdwa
+---
```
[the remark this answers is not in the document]
frederik-h: This makes sense to me, I think it should be changed.

frederik-h (changes requested):

llvm/test/CodeGen/AMDGPU/sdwa-peephole-instr.mir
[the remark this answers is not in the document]
LU-JOHN: Changed.

frederik-h (approved):

llvm/lib/Target/AMDGPU/SIPeepholeSDWA.cpp
LU-JOHN: > Sure. If we add it, we need to undo your change to the way the Sext value is being set, but it should be easy enough to do this later.

In VOP2Instructions.td the 16-bit ASHR opcode is defined by:
```
// Note: 16-bit instructions produce a 0 result in the high 16-bits
// on GFX8 and GFX9 and preserve high 16 bits on GFX10+
multiclass Arithmetic_i16_0Hi_Pats <SDPatternOperator op, Instruction inst> {
```
The same comment also appears in VOP3Instructions.td before another pattern.  Thus, it seems that no 16-bit ASHR sign-extends to a 32-bit result and thus no form of 16-bit ASHR should be combined to form a SDWA instruction.

PR #198492, opened 2026-05-19
[libc++] Applied `[[nodiscard]]` to `array::iterator`

H-G-Hristov:
Towards #172124

### DIFF
--- a/libcxx/test/libcxx/containers/sequences/array/nodiscard.__static_bounded_iter.verify.cpp
+++ b/libcxx/test/libcxx/containers/sequences/array/nodiscard.__static_bounded_iter.verify.cpp
@@ -1,0 +1,47 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// ADDITIONAL_COMPILE_FLAGS: -D_LIBCPP_ABI_BOUNDED_ITERATORS_IN_STD_ARRAY
+
+// <array>
+
+// Check that functions are marked [[nodiscard]]
+
+#include <array>
+
+#include "test_macros.h"
+
+void test() { // __static_bounded_iter
+  typedef std::array<int, 94> Container;
+  ASSERT_SAME_TYPE(Container::iterator, std::__static_bounded_iter<int*, 94>);
+
+  Container::iterator it;
+
+  // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
+  *it;
+
+  // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
+  it[0];
+
+  // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
+  it + 1;
+
+  // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
+  1 + it;
+
+  // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
+  it - 1;
+
+  // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
+  it - it;
+
+  std::pointer_traits<Container::iterator> pt;
+
+  // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
+  pt.to_address(it);
+}
--- a/libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
+++ b/libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
@@ -6,8 +6,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-// UNSUPPORTED: libcpp-hardening-mode=none
-
 // <array>
 
 // Check that functions are marked [[nodiscard]]
@@ -25,8 +23,11 @@ void test() {
   ASSERT_SAME_TYPE(Container::iterator, std::__static_bounded_iter<int*, 94>);
 #elif defined(_LIBCPP_ABI_USE_WRAP_ITER_IN_STD_ARRAY)
   ASSERT_SAME_TYPE(Container::iterator, std::__wrap_iter<int*>);
+#else
+  ASSERT_SAME_TYPE(Container::iterator, int*);
 #endif
 
+#if defined(_LIBCPP_ABI_BOUNDED_ITERATORS_IN_STD_ARRAY) || defined(_LIBCPP_ABI_USE_WRAP_ITER_IN_STD_ARRAY)
   // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
   *it;
 
@@ -44,6 +45,25 @@ void test() {
 
   // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
   it - it;
+#else
+  // expected-warning@+1 {{expression result unused}}
+  *it;
+
+  // expected-warning@+1 {{expression result unused}}
+  it[0];
+
+  // expected-warning@+1 {{expression result unused}}
+  it + 1;
+
+  // expected-warning@+1 {{expression result unused}}
+  1 + it;
+
+  // expected-warning@+1 {{expression result unused}}
+  it - 1;
+
+  // expected-warning@+1 {{expression result unused}}
+  it - it;
+#endif
 
 #if defined(_LIBCPP_ABI_BOUNDED_ITERATORS_IN_STD_ARRAY)
   // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
--- a/libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
+++ b/libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
@@ -36,9 +36,4 @@ void test() {
 
   // expected-warning-re@+1 {{{{(ignoring return value of function declared with 'nodiscard' attribute|expression result unused)}}}}
   it - it;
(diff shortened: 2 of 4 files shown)


libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
```diff
@@ -0,0 +1,47 @@
+//===----------------------------------------------------------------------===//
```
frederick-vs-ja, line 1: Could we rename this file to `nodiscard.static_bounded_iter.verify.cpp` which is less __ugly?

libcxx/test/libcxx/containers/sequences/array/nodiscard.__static_bounded_iter.verify.cpp
```diff
@@ -0,0 +1,47 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// ADDITIONAL_COMPILE_FLAGS: -D_LIBCPP_ABI_BOUNDED_ITERATORS_IN_STD_ARRAY
```
frederick-vs-ja: I wonder whether it's better to move the contents below this line to `nodiscard.iterator.inc` (with banner copied), and then make `.verify.cpp` test files less boilerplate.

`nodiscard.static_bounded_iter.verify.cpp`:
```C++
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//

// ADDITIONAL_COMPILE_FLAGS: -D_LIBCPP_ABI_BOUNDED_ITERATORS_IN_STD_ARRAY

#include "nodiscard.iterator.inc"
```

`nodiscard.wrap_iter.verify.cpp` (in a future addition):
```C++
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//

// ADDITIONAL_COMPILE_FLAGS: -D_LIBCPP_ABI_USE_WRAP_ITER_IN_STD_ARRAY

#include "nodiscard.iterator.inc"
```

libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
```diff
@@ -0,0 +1,47 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// ADDITIONAL_COMPILE_FLAGS: -D_LIBCPP_ABI_BOUNDED_ITERATORS_IN_STD_ARRAY
+
+// <array>
```
frederick-vs-ja, line 9:
```suggestion
// <array>

// using iterator       = implementation-defined;
// using const_iterator = implementation-defined;
```

Also, I guess it makes some sense to test `const_iterator` together.

libcxx/test/libcxx/containers/sequences/array/nodiscard.__static_bounded_iter.verify.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+// Check that functions are marked [[nodiscard]]
+
+#include <array>
+
+#include "test_macros.h"
+
+void test() { // __static_bounded_iter
+  typedef std::array<int, 94> Container;
+  ASSERT_SAME_TYPE(Container::iterator, std::__static_bounded_iter<int*, 94>);
```
frederick-vs-ja: If we want to reuse the contents for both `__static_bounded_iter` and `__wrap_iter`, perhaps it's better to remove the `// __static_bounded_iter` comment and use conditional compilation for `ASSERT_SAME_TYPE`.

libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
philnik777, line 1: Why is this specific to `__static_bounded_iter` instead of testing for any `array::iterator`?

philnik777 (changes requested):

libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
frederick-vs-ja: I guess @Zingam wanted to add `[[nodicard]]` for different wrapping iterators in different PRs.

philnik777: That seems like a very bad reason to write tests this way. What we care about is that `array::iterator` is diagnosed, not that some internal utility is.

Zingam: @philnik777 (cc: @frederick-vs-ja) I think I understand your concerns and what you want to see in general but I am not familiar with the how hardening is tested in the CI, so I am unsure how to shape the tests to achieve proper coverage. I made some changes. Is this closer to what you have in mind? Any suggestions are very welcome! Thank you!

libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
```diff
@@ -0,0 +1,72 @@
...
+  Container c;
+  Container::iterator it = c.begin();
+
+#if defined(_LIBCPP_ABI_BOUNDED_ITERATORS_IN_STD_ARRAY)
+  ASSERT_SAME_TYPE(Container::iterator, std::__static_bounded_iter<int*, 94>);
+#elif defined(_LIBCPP_ABI_USE_WRAP_ITER_IN_STD_ARRAY)
+  ASSERT_SAME_TYPE(Container::iterator, std::__wrap_iter<int*>);
+#else
+  ASSERT_SAME_TYPE(Container::iterator, int*);
+#endif
```
philnik777: What are these useful for?

libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+#if defined(_LIBCPP_ABI_BOUNDED_ITERATORS_IN_STD_ARRAY) || defined(_LIBCPP_ABI_USE_WRAP_ITER_IN_STD_ARRAY)
```
philnik777: Instead of constraining this on specific configurations of the library I'd much rather combine the messages to `// expected-warning-re {{(nodiscard-message|raw-ptr-message)}}`.

libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
```diff
@@ -0,0 +1,72 @@
...
+  // expected-warning@+1 {{expression result unused}}
+  it - 1;
+
+  // expected-warning@+1 {{expression result unused}}
+  it - it;
+#endif
+
+#if defined(_LIBCPP_ABI_BOUNDED_ITERATORS_IN_STD_ARRAY)
+  // expected-warning@+1 {{ignoring return value of function declared with 'nodiscard' attribute}}
+  std::pointer_traits<Container::iterator>::to_address(it);
```
philnik777: I'm not too sure about this one. Is the specialization actually needed?

libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
Zingam: Removed.
These were useful to me while I was figuring this out.

libcxx/test/libcxx/containers/sequences/array/nodiscard.iterator.verify.cpp
Zingam: Do you mean we should remove this specialization? I don't know about that, A similar specialization in `__wrap_iter` was removed.  I think if we have it, we should test it and the removal should be done in another patch anyway.

frederick-vs-ja: The `pointer_traits<__bounded_iter>` partial specialization (and especially its `to_address` function) appears to be necessary, because `std::to_address` needs to accept past-of-end iterators in order to make `__bounded_iter` model `contiguous_iterator`, while `__bounded_iter::operator->` rejects past-of-end iterators.

philnik777: I'm somewhat questioning the usefulness of `[[nodiscard]]` here. I expect basically all usage should be through `std::to_address`, since it's not guaranteed to exist in `pointer_traits` generally. I guess I'm not sure we should treat this as public-facing API - even if users are allowed to observe the existence; I expect it would be in some `to_address`-like construct 99% of the time.

philnik777: I also really don't like this to be part of the `array::iterator` test, since this isn't actually about `array::iterator` in general, but about one particular iterator which happens to need to provide it. I don't think we actually _want_ people to use it directly.

Zingam: I removed the test and the `[[nodiscard]]` although the library sometimes provides `[[nodiscard]]` for internal stuff (inconsistenly) for less friction.

philnik777 (approved):

Zingam:
Thank you!

 I'll be merging when this gets green and then I'll do the remaining bits.
@frederick-vs-ja are you OK with what we have now?

frederick-vs-ja (approved):
LGTM!

PR #195284, opened 2026-05-01
[LLVM][Constants] Store "splat (float 0.0)" as ConstantFP rather than ConstantAggregateZero.

paulwalker-arm:
The original split is awkward because, not unreasonably, some code paths expect constant folding of ConstantFP operands to yield a ConstantFP result.

I'd like to follow up and investigated why the scalar and vector output takes different forms, but for now I wanted to fix the assertion failure.

NOTE: The change to ConstantFoldInsertElementInstruction is required to maintain existing test output.

Fixes https://github.com/llvm/llvm-project/issues/194590

PR #198496, opened 2026-05-19
[LLVM][ConstantFP] Replace uses of isExactlyValue(+/-0.0) with isPosZero/isNegZero.

paulwalker-arm:
This follows on from https://github.com/llvm/llvm-project/pull/195284#discussion_r3181591728.

### DIFF
--- a/llvm/lib/Target/AMDGPU/AMDGPURegBankCombiner.cpp
+++ b/llvm/lib/Target/AMDGPU/AMDGPURegBankCombiner.cpp
@@ -287,7 +287,7 @@ bool AMDGPURegBankCombinerImpl::matchFPMinMaxToClamp(MachineInstr &MI,
   if (!matchMed<GFCstOrSplatGFCstMatch>(MI, MRI, OpcodeTriple, Val, K0, K1))
     return false;
 
-  if (!K0->Value.isExactlyValue(0.0) || !K1->Value.isExactlyValue(1.0))
+  if (!K0->Value.isPosZero() || !K1->Value.isExactlyValue(1.0))
     return false;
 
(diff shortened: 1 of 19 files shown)


llvm/lib/Target/AMDGPU/AMDGPURegBankCombiner.cpp
```diff
@@ -287,7 +287,7 @@ bool AMDGPURegBankCombinerImpl::matchFPMinMaxToClamp(MachineInstr &MI,
   if (!matchMed<GFCstOrSplatGFCstMatch>(MI, MRI, OpcodeTriple, Val, K0, K1))
     return false;
 
-  if (!K0->Value.isExactlyValue(0.0) || !K1->Value.isExactlyValue(1.0))
+  if (!K0->Value.isPosZero() || !K1->Value.isExactlyValue(1.0))
```
david-arm, line 290: Not for this PR, but looks like there might be a case for having a `isOne` as well for the `isExactlyValue(1.0)`? That's if there isn't one already!

david-arm (approved):
LGTM!

Is it worth updating Constant::isNullValue as well?

paulwalker-arm:
> Is it worth updating Constant::isNullValue as well?

IsNullValue means something very specific. See https://github.com/llvm/llvm-project/issues/196097.

llvm/lib/Target/AMDGPU/AMDGPURegBankCombiner.cpp
arsenm: I'd also like if we had isPowerOf2(N)

PR #198501, opened 2026-05-19
[RISCV][NFC] Merge two `isLegalElementTypeForRVV`s into RISCVSubtarget

wangpc-pp:
We have two similar implementations in `RISCVTargetLowering.cpp`
and `RISCVCallLowering.cpp`.

We move them to `RISCVSubtarget` and merge the implementations to
remove the duplication.

### DIFF
--- a/llvm/lib/Target/RISCV/RISCVSubtarget.cpp
+++ b/llvm/lib/Target/RISCV/RISCVSubtarget.cpp
@@ -252,7 +252,6 @@ bool RISCVSubtarget::isLegalElementTypeForRVV(EVT ScalarTy) const {
   switch (ScalarTy.getSimpleVT().SimpleTy) {
   case MVT::iPTR:
     return is64Bit() ? hasVInstructionsI64() : true;
-  case MVT::i1:
   case MVT::i8:
   case MVT::i16:
   case MVT::i32:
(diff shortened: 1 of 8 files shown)


lukel97 (approved):
LGTM

llvm/lib/Target/RISCV/RISCVSubtarget.cpp
```diff
@@ -240,6 +240,37 @@ bool RISCVSubtarget::useRVVForFixedLengthVectors() const {
          getMinRVVVectorSizeInBits() >= RISCV::RVVBitsPerBlock;
 }
 
+bool RISCVSubtarget::isLegalElementTypeForRVV(Type *EltTy) const {
+  return isLegalElementTypeForRVV(EVT::getEVT(EltTy));
```
lukel97: Oh I think we need to pass DataLayout here for the pointer types

wangpc-pp: And I don't understand why `isLegalElementTypeForRVV(Type *EltTy)` needs `i1` while `isLegalElementTypeForRVV(EVT ScalarTy)` doesn't. This makes it impossible to merge the code.

wangpc-pp: I convert this PR to draft until I have investigated the reason.

PR #198502, opened 2026-05-19
[AArch64][GlobalISel] Expand buildvector to truncate code to handled double truncs.

davemgreen:
In AArch64 we want to perform xtn(concat(xtn())) for a double-truncation, but can end up being scalarized during lowering. This expands the existing code to handle such cases.


llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -3780,40 +3780,65 @@ bool CombinerHelper::matchUseVectorTruncate(MachineInstr &MI,
     return false;
 
   // Only generate legal instructions post-legalizer
```
c-rhodes, line 3782: this comment should be moved after prelegalize check

llvm/lib/CodeGen/GlobalISel/CombinerHelper.cpp
```diff
@@ -3780,40 +3780,65 @@ bool CombinerHelper::matchUseVectorTruncate(MachineInstr &MI,
...
-  LLT UnmergeSrcTy = MRI.getType(MatchInfo);
+  LLT UnmergeSrcTy = MRI.getType(MatchInfo.first);
+  LLT MidTy = MatchInfo.second;
   unsigned DstTyNumElt = DstTy.getNumElements();
   unsigned UnmergeSrcTyNumElt = UnmergeSrcTy.getNumElements();
 
   // No need to pad vector if only G_TRUNC is needed
   if (DstTyNumElt / UnmergeSrcTyNumElt == 1) {
-    MidReg = MatchInfo;
+    MidReg = MatchInfo.first;
```
c-rhodes, line 3827: Codex found a bug here. I've gone thru the code and think it's a correct find

Pattern for reference:
```
build_vector(trunc(unmerge(src)), ..., undef) -> trunc(concat_vectors(src, undef...))
````
Example:
```
%src: <4 x s32>
%a, %b, %c, %d = G_UNMERGE_VALUES %src
%ta = G_TRUNC %a : s32 -> s8
%tb = G_TRUNC %b : s32 -> s8
%tc = G_TRUNC %c : s32 -> s8
%td = G_TRUNC %d : s32 -> s8
%dst = G_BUILD_VECTOR %ta, %tb, %tc, %td   ; <4 x s8>
```
Imagine a target where:
```
G_TRUNC <4 x s32> -> <4 x s8>    is illegal
G_TRUNC <4 x s32> -> <4 x s16>   is legal
G_TRUNC <4 x s16> -> <4 x s8>    is legal
```
Then the matcher can choose the staged plan:
```
MatchInfo.first  = %src
MatchInfo.second = <4 x s16>
```
meaning the intended rewrite is:
```
%half = G_TRUNC %src    ; <4 x s32> -> <4 x s16>
%dst  = G_TRUNC %half   ; <4 x s16> -> <4 x s8>
```
But apply sees that no padding is needed:
```
if (DstTyNumElt / UnmergeSrcTyNumElt == 1) {
  MidReg = MatchInfo.first;
}
Builder.buildTrunc(DstReg, MidReg);
```
So it emits:
```
%dst = G_TRUNC %src     ; <4 x s32> -> <4 x s8>
```

PR #198504, opened 2026-05-19
[cmake][runtimes] Refactor passthrough option handling into a macro

tambry:
Prep work for #198505. Depends on #198506.

---
Deduplicates code and enables making fixes in just one place instead of two.


llvm/runtimes/CMakeLists.txt
```diff
@@ -404,24 +414,7 @@ function(runtime_register_target name)
...
-            list(APPEND ${name}_extra_args -C ${cache})
-          endforeach()
-        else()
-          string(REPLACE ";" "|" new_value "${${variable_name}}")
-          list(APPEND ${name}_extra_args "-D${new_name}=${new_value}")
-        endif()
-      endif()
-    endforeach()
-  endforeach()
+  append_passthrough_options(${name}_extra_args RUNTIMES "${name};${ARG_BASE_NAME}")
```
petrhosek: Is there a specific reason for reversing the order of `${name}` and `${ARG_BASE_NAME}`? That order was selected deliberately so the variant options can override the base ones.

llvm/runtimes/CMakeLists.txt
```diff
@@ -91,6 +91,30 @@ macro(set_enable_per_target_runtime_dir)
   endif()
 endmacro()
 
+macro(append_passthrough_options out type names)
+  get_cmake_property(variable_names VARIABLES)
+
+  # Support multiple names for layering (e.g. multilibs).
+  foreach(name ${names})
```
petrhosek: I'd slightly prefer `IN ITEMS` as the more modern syntax:
```suggestion
  foreach(name IN ITEMS ${names})
```

tambry: I'm just used to the old ways. :sweat_smile:

llvm/runtimes/CMakeLists.txt
tambry: Just an oversight, well spotted!
My first draft of append_passthrough_options() didn't handle ARG_BASE_NAME so I tacked that on later and as such the order changed.

petrhosek (approved):

PR #198505, opened 2026-05-19
[cmake][runtimes] Pass variable type for passthrough CMake options

tambry:
Depends on #198505 and #198506.

---
Passing the types affects the handling of `PATH` cache variables.
The motivating case is setting `COMPILER_RT_INSTALL_LIBRARY_DIR` without `CMAKE_BINARY_DIR` getting prepended to it.

### DIFF
--- a/llvm/runtimes/CMakeLists.txt
+++ b/llvm/runtimes/CMakeLists.txt
@@ -107,6 +107,14 @@ macro(append_passthrough_options out type names)
             list(APPEND ${out} -C ${cache})
           endforeach()
         else()
+          # Type is important for cache variables.
+          # For example PATH variables will be made relative to CMAKE_BINARY_DIR
+          # at the point of set(CACHE) if the type isn't passed along.
+          get_property(type CACHE ${variable_name} PROPERTY TYPE)
+          if(NOT type STREQUAL "UNINITIALIZED")
+            string(APPEND new_name ":${type}")
+          endif()
+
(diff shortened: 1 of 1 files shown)


petrhosek (approved):

petrhosek:
Looks like this broke our toolchain build with the following error:
```
CMake Error in /Users/swarming/b/s/w/ir/x/w/llvm-llvm-project/libcxx/src/CMakeLists.txt:
  No known features for CXX compiler

  "Clang"

  version 23.0.0.


CMake Generate step failed.  Build files cannot be regenerated correctly.
```
You can see the full log here https://logs.chromium.org/logs/fuchsia/buildbucket/cr-buildbucket/8681299599353914865/+/u/clang/build/stdout

Would it be possible to revert?

tambry:
@petrhosek Revert in #198739. I see nothing obvious in `Fuchsia-stage2.cmake` that could break in combination with this change.

Would you be able to provide a zipped `/Users/swarming/b/s/w/ir/x/w/llvm_build/runtimes/runtimes-x86_64-pc-windows-msvc-bins` from the buildbot after failure? Or at least `CMakeCache.txt` and `CMakeFiles/CMakeConfigureLog.yaml`? I lack any Windows machines or an appropriate sysroot to reproduce locally. The Fuchsia IDK doesn't seem sufficient.

tambry:
@petrhosek Ping. I don't want to break your downstream out-of-tree things but I also don't have a way to debug that.

PR #175167, opened 2026-01-09
[NaryReassociate] Teach NaryReassociate about UniformityAnalysis

PankajDwivedi-25:
This patch integrates `UniformityAnalysis` into the `NaryReassociate` pass to prefer reassociations that group uniform values together. This optimization benefits GPU targets (like AMDGPU) by allowing intermediate results to stay in scalar registers (SGPRs), reducing vector register (VGPR) pressure.

For an expression I = (A op B) op RHS, NaryReassociate can form either:

- (A op RHS) op B - groups A and RHS
- (B op RHS) op A - groups B and RHS

When both options have dominating expressions available, the pass previously chose arbitrarily based on operand order. With this patch, when `UniformityAnalysis` indicates that B and RHS are uniform but A is divergent, the pass prefers (B op RHS) op A so the sub-expression (B op RHS) can be computed entirely in scalar registers.

```
; d = divergent (e.g., workitem.id), u1/u2 = uniform (kernel args)
; Both (d + u2) and (u1 + u2) exist as dominating expressions

%d_u2 = add i32 %d, %u2      ; divergent + uniform
%u1_u2 = add i32 %u1, %u2    ; uniform + uniform

%tmp = add i32 %d, %u1
%result = add i32 %tmp, %u2  ; = (d + u1) + u2

; Before: %result = add i32 %d_u2, %u1   ; picks divergent grouping
; After:  %result = add i32 %u1_u2, %d   ; picks uniform grouping
```

It is supported for:
- Binary ops (add, mul)
- MinMax intrinsics (smax, umin)
- GEP index reassociation

Ref: SWDEV-500808

PR #198507, opened 2026-05-19
[Reassociate] Use UniformityInfo to group uniform operands together

PankajDwivedi-25:
On targets with branch divergence (TTI.hasBranchDivergence()), boost the rank of divergent operands in ReassociateExpression() so they sort towards the root of the expression tree. This clusters uniform operands at the leaves, letting the backend lower the inner sub-expression to a scalar ALU instruction (s_mul/s_or/s_and) instead of a vector.

Ref: https://github.com/llvm/llvm-project/pull/175167#issuecomment-4476483692


llvm/lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -2606,15 +2644,17 @@ PreservedAnalyses ReassociatePass::run(Function &F, FunctionAnalysisManager &) {
...
   RankMap.clear();
   ValueRankMap.clear();
   for (auto &Entry : PairMap)
     Entry.clear();
+  UA = nullptr;
 
   if (MadeChange) {
     PreservedAnalyses PA;
     PA.preserveSet<CFGAnalyses>();
+    PA.preserve<UniformityInfoAnalysis>();
```
krzysz00: Wouldn't this require some sort of updater for the new expressions?

llvm/test/CodeGen/AMDGPU/reassoc-scalar.ll
```diff
@@ -125,5 +125,119 @@ bb:
   ret void
 }
 
+; Multiply: (x * tid) * y -- backend does NOT reassociate mul to scalar.
+; Both muls use VALU even though x and y are uniform.
+; GCN-LABEL: reassoc_mul_i32:
+; GCN-NOT: s_mul
+; GCN: v_mul_lo_u32
+; GCN: v_mul_lo_u32
+define amdgpu_kernel void @reassoc_mul_i32(ptr addrspace(1) %arg, i32 %x, i32 %y) {
```
krzysz00: Nit: drop _kernel and use `inreg` for uniform arguments

llvm/test/CodeGen/AMDGPU/reassoc-scalar.ll
```diff
@@ -125,5 +125,119 @@ bb:
...
+bb:
```
krzysz00: Nit: no need for a block name

krzysz00 (commented):
Overall this seems good and reasonable to me, though I'll tag in some more folks since I don't want to wield the stamp unilaterally

For the backend tests, could you TODO them or otherwise put in a comment saying that `reassociate` handles these?

llvm/lib/Transforms/Scalar/Reassociate.cpp
PankajDwivedi-25: With the current changes in the uniformity, where we report the newly created value divergent this looks fine, but in the case the value created is divergent we still do not have its propagation, so for the safe side I will remove this.

llvm/test/CodeGen/AMDGPU/reassoc-scalar.ll
PankajDwivedi-25: Fixed.

llvm/test/CodeGen/AMDGPU/reassoc-scalar.ll
PankajDwivedi-25: Fixed.

llvm/lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -2543,7 +2571,17 @@ ReassociatePass::BuildPairMap(ReversePostOrderTraversal<Function *> &RPOT) {
   }
 }
 
-PreservedAnalyses ReassociatePass::run(Function &F, FunctionAnalysisManager &) {
+PreservedAnalyses ReassociatePass::run(Function &F,
+                                       FunctionAnalysisManager &AM) {
+  // On targets with branch divergence, obtain UniformityInfo so we can group
+  // uniform operands together in expression trees.
+  UA = nullptr;
```
arsenm: This should already be null

llvm/lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -31,6 +31,7 @@
 #include "llvm/Analysis/BasicAliasAnalysis.h"
 #include "llvm/Analysis/ConstantFolding.h"
 #include "llvm/Analysis/GlobalsModRef.h"
+#include "llvm/Analysis/TargetTransformInfo.h"
```
arsenm: Not sure spreading TTI here is OK

llvm/lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -2637,6 +2676,7 @@ class ReassociateLegacyPass : public FunctionPass {
     if (skipFunction(F))
       return false;
 
+    Impl.SkipUniformityAnalysis = true;
```
arsenm: I don't understand modifying this mid pass? I thought this was going to be a pass parameter

arsenm (commented):
Kind of feels like this should be in its own optimization, like the uniformity combiner we already have?

PankajDwivedi-25:
> Kind of feels like this should be in its own optimization, like the uniformity combiner we already have?

I don't think having another pass doing similar job is necessary, its just we are trying to optimize a generic pass using generic analysis. 

is there any reason here to do?

llvm/lib/Transforms/Scalar/Reassociate.cpp
PankajDwivedi-25: This is not the first time we are using TTI in generic pass, we have many passes which do that.

llvm/lib/Transforms/Scalar/Reassociate.cpp
PankajDwivedi-25: Fixed

llvm/lib/Transforms/Scalar/Reassociate.cpp
PankajDwivedi-25: Fixed

llvm/lib/Transforms/Scalar/Reassociate.cpp
krzysz00: I think the better long-term thing would be adding machinery to UniformityAnalysis so that, when you create new IR, you can inform the uniformity analysis about whether this IR is uniform or not

PankajDwivedi-25: Yes, thats the long term goal we want in uniformity.

llvm/include/llvm/Transforms/Scalar/Reassociate.h
krzysz00: Is it actually true that we can't get the uniformity analysis in the old pass manager?

krzysz00: ... Though also, you don't need a separate boolean for this - just use `UA = nullptr;`

llvm/lib/Transforms/Scalar/Reassociate.cpp
krzysz00, line 2302: Out of curiosity, if I have
```
(uniform1 + divergent1) + (uniform2 + divergent2)
```
(for example - any ordering of those will work)
will I get
```
((uniform1 + uniform2) + divergent1) + divergent2
```
or
```
(uniform1 + uniform2) + (divergent1 + divergent2)
```
?

Or would I need to start with `((uniform1 + divergent1) + uniform2) + divergent2` for this pass to fire?

krzysz00 (commented):
Some questions, I think this is probably close to landable

(Though I have to wonder - do we have benchmarks on this?)

llvm/lib/Transforms/Scalar/Reassociate.cpp
PankajDwivedi-25: we get ((u1 + u2) + d1) + d2, and don't need any particular starting shape. Reassociate's `LinearizeExprTree` flattens the whole same-opcode tree into a flat operand list {u1, d1, u2, d2} regardless of the input bracketing.

we then boost the divergent ranks, the existing stable_sort re-orders, and RewriteExprTree rebuilds a left-linear chain with the uniforms deepest and the divergents applied toward the root.
So it's the first form, not the balanced.

llvm/include/llvm/Transforms/Scalar/Reassociate.h
PankajDwivedi-25: > Is it actually true that we can't get the uniformity analysis in the old pass manager?

No, Thanks for pointing, There was never a real technical blocker to getting UniformityInfo in the legacy PM;

I have wired it properly now.

PankajDwivedi-25:
> (Though I have to wonder - do we have benchmarks on this?)

I don't have one yet.

llvm/lib/Transforms/Scalar/Reassociate.cpp
krzysz00: Ok, cool, that feels like something decently straightforward to pattern-match on (I'm thinking of buffer offsets, where I'm looking for `u +<nuw> d +<nuw> const` expressions (so I probably want `(+<nuw> d)*` if I were looking.

krzysz00 (approved):
One other side question - what're the relative priorities of loop-independent and uniform values?

Because I'd expect reassociate to create LICMable expressions where possible.

But that's a not critical.

This is a soft approve, I think that this makes sense, I'll wait a bit to see if @arsenm has further thoughts.

krzysz00:
Last call for pre-merge complaints (I can't tell if that "should TTI be here" is blocking or more of a "hmmmm...".

@PankajDwivedi-25 I figure we could either land this before LLVM 23 (so US tomorrow) or right after it (US Wednesday) - I'll leave that up to you.

PankajDwivedi-25:
> Last call for pre-merge complaints (I can't tell if that "should TTI be here" is blocking or more of a "hmmmm...".
> 
> @PankajDwivedi-25 I figure we could either land this before LLVM 23 (so US tomorrow) or right after it (US Wednesday) - I'll leave that up to you.

as suggested I will wait for today for any further feedback.

llvm/lib/Transforms/Scalar/Reassociate.cpp
arsenm: This is not exactly true, this is a core canonicalization pass. It's not like a late optimization pass where TTI gets used

arsenm:
> Last call for pre-merge complaints (I can't tell if that "should TTI be here" is blocking or more of a "hmmmm...".

I think TTI should not spread here, and uniformity aware reassociate should be in its own uniformity aware backend optimization

PankajDwivedi-25:
> I think TTI should not spread here, and uniformity aware reassociate should be in its own uniformity aware backend optimization

As i mentioned above having another pass doing similar job is like creating a copy of this pass. That will be more of compile time do we really want that? also I am seeing other passes where we use TTI.

krzysz00:
> I think TTI should not spread here, and uniformity aware reassociate should be in its own uniformity aware backend optimization

I weakly disagree - checking for uniformity is a reasonable thing for a generic pass to do (there are multiple targets that would want this) and reassociation isn't something we want running super-late/in LLC.

jayfoad:
I don't understand why you need TTI at all. Isn't UA supposed to be virtually free on targets without divergence? Especially now that the new pass manager also lets you avoid computing CycleInfo?

PankajDwivedi-25:
> I don't understand why you need TTI at all. Isn't UA supposed to be virtually free on targets without divergence? Especially now that the new pass manager also lets you avoid computing CycleInfo?

thanks for mentioning, completely missed this point. The hasBranchDivergence guard is redundant now.

PankajDwivedi-25:
@arsenm is there anything else from your end else I will move ahead.

llvm/include/llvm/Transforms/Scalar/Reassociate.h
krzysz00: Stale comment. Also, does this need to be a pointer now?

llvm/lib/Transforms/Scalar/Reassociate.cpp
krzysz00, line 2313: Is divergence at use the right check? Do we not instead want to check for divergence at the definition? (that'll save us O(uses) as well).

krzysz00 (approved):
One question, one note, otherwise approved (and I think the one reviewer comment re TTI has been addressed)

llvm/lib/Transforms/Scalar/Reassociate.cpp
PankajDwivedi-25: I think we do want divergence at the use here rather than at the definition, because of temporal divergence.

llvm/include/llvm/Transforms/Scalar/Reassociate.h
PankajDwivedi-25: fixed.

PR #198508, opened 2026-05-19
[AMDGPU] Gate `S_LSHL[1-4]_ADD_U32` patterns on uniform results

aobolensk:
Like the other SOP2 patterns in this file, these scalar instructions require the result to be uniform. Wrap them in `UniformBinFrag` so divergent shl/add chains use `V_LSHL_ADD_U32`

### DIFF
--- a/llvm/lib/Target/AMDGPU/SOPInstructions.td
+++ b/llvm/lib/Target/AMDGPU/SOPInstructions.td
@@ -907,16 +907,16 @@ let SubtargetPredicate = isGFX9Plus in {
 
   let Defs = [SCC] in {
     def S_LSHL1_ADD_U32 : SOP2_32<"s_lshl1_add_u32",
-      [(set i32:$sdst, (shl1_add SSrc_b32:$src0, SSrc_b32:$src1))]
+      [(set i32:$sdst, (UniformBinFrag<shl1_add> SSrc_b32:$src0, SSrc_b32:$src1))]
     >;
     def S_LSHL2_ADD_U32 : SOP2_32<"s_lshl2_add_u32",
-      [(set i32:$sdst, (shl2_add SSrc_b32:$src0, SSrc_b32:$src1))]
+      [(set i32:$sdst, (UniformBinFrag<shl2_add> SSrc_b32:$src0, SSrc_b32:$src1))]
     >;
     def S_LSHL3_ADD_U32 : SOP2_32<"s_lshl3_add_u32",
-      [(set i32:$sdst, (shl3_add SSrc_b32:$src0, SSrc_b32:$src1))]
+      [(set i32:$sdst, (UniformBinFrag<shl3_add> SSrc_b32:$src0, SSrc_b32:$src1))]
     >;
     def S_LSHL4_ADD_U32 : SOP2_32<"s_lshl4_add_u32",
-      [(set i32:$sdst, (shl4_add SSrc_b32:$src0, SSrc_b32:$src1))]
+      [(set i32:$sdst, (UniformBinFrag<shl4_add> SSrc_b32:$src0, SSrc_b32:$src1))]
     >;
   } // End Defs = [SCC]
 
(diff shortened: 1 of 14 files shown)


jayfoad:
LGTM. However, was there actually a correctness problem here? If S_LSHL*_ADD was selected for divergent inputs then it should have been fixed up by SIFixSGPRCopies, so the code should still have been functionally correct.

aobolensk:
> LGTM. However, was there actually a correctness problem here? If S_LSHL*_ADD was selected for divergent inputs then it should have been fixed up by SIFixSGPRCopies, so the code should still have been functionally correct.

A divergent shl/add chain gets ISel'ed as `S_LSHL[1-4]_ADD_U32`, whose `SSrc_b32` operand forces a VGPR->SGPR copy on the divergent input. That copy lowers to v_readfirstlane_b32, which keeps lane 0 only. Scalar chain then runs and the result is broadcast back, so every lane gets lane 0's value which results in a miscompile

At -O2, SIFixSGPRCopies rewrites the chain back to vector form, so it's fine there

jayfoad:
> At -O2, SIFixSGPRCopies rewrites the chain back to vector form, so it's fine there

Why doesn't this happen at -O0? SIFixSGPRCopies is required for correctness so it should run always.

jayfoad:
> A divergent shl/add chain gets ISel'ed as S_LSHL[1-4]_ADD_U32, whose SSrc_b32 operand forces a VGPR->SGPR copy on the divergent input. That copy lowers to v_readfirstlane_b32

How does it get lowered to v_readfirstlane_b32? That should not happen.

aobolensk:
> > A divergent shl/add chain gets ISel'ed as S_LSHL[1-4]_ADD_U32, whose SSrc_b32 operand forces a VGPR->SGPR copy on the divergent input. That copy lowers to v_readfirstlane_b32
> 
> How does it get lowered to v_readfirstlane_b32? That should not happen.

Since VGPR->SGPR copy is required this function is being called: https://github.com/llvm/llvm-project/blob/4a32cf0b6d90f6bec3b5f2ad50b11ca1dba62954/llvm/lib/Target/AMDGPU/SIFixSGPRCopies.cpp#L1026

> > At -O2, SIFixSGPRCopies rewrites the chain back to vector form, so it's fine there
> 
> Why doesn't this happen at -O0? SIFixSGPRCopies is required for correctness so it should run always.

on -O0 and -O2 we get different IRs at that point where the problem happens due to instcombiner.

Case I have been investigating:
```
  %a0 = add i32 %x,  1
  %s0 = shl i32 %a0, 1
  %a1 = add i32 %s0, 1
  %s1 = shl i32 %a1, 1
  %a2 = add i32 %s1, 1
  %s2 = shl i32 %a2, 1
  %a3 = add i32 %s2, 1
  %s3 = shl i32 %a3, 1
  %a4 = add i32 %s3, 1
  %s4 = shl i32 %a4, 1
  %result = add i32 %s4, 1
```
in case of -O2:
```
  %24:vgpr_32 = GLOBAL_LOAD_DWORD_SADDR ...
  %25:sreg_32 = S_MOV_B32 63
  %26:sreg_32 = S_MOV_B32 5
  %28:vgpr_32 = COPY killed %25
  %27:vgpr_32 = V_LSHL_ADD_U32_e64 killed %24, killed %26, %28, implicit $exec
```
-O0:
```
  %26:vgpr_32 = GLOBAL_LOAD_DWORD ...
  %27:sreg_32 = S_MOV_B32 1
  %28:vgpr_32 = V_ADD_U32_e64 killed %26, %27, 0, implicit $exec
  %30:sreg_32 = COPY %28:vgpr_32                      ; vgpr -> sgpr
  %29:sreg_32 = S_LSHL1_ADD_U32 killed %30, %27, ...
  %31:sreg_32 = S_LSHL1_ADD_U32 killed %29, %27, ...
  %32:sreg_32 = S_LSHL1_ADD_U32 killed %31, %27, ...
  %33:sreg_32 = S_LSHL1_ADD_U32 killed %32, %27, ...
  %34:sreg_32 = S_LSHL1_ADD_U32 killed %33, %27, ...
```
In -O0 case 5x`S_LSHL1_ADD_U32` pops up and COPY will be transformed to V_READFIRSTLANE_B32, as I mentioned before.

So in -O2 we just don't face this situation

jayfoad:
I see. Maybe I have misremembered how this is supposed to work. @alex-t is it expected that SIFixSGPRCopies will lower some vgpr->sgpr copies into readfirstlane? So we _must_ have `UniformBinFrag` or similar on every SALU selection pattern?

jayfoad (approved):
I think this is fine. Sorry for the noise.

arsenm:
>  So we _must_ have `UniformBinFrag` or similar on every SALU selection pattern?

SIFixSGPRCopies is still required to resolve some situations, but it's a last resort. SALU patterns should all be marked as requiring uniform

PR #198510, opened 2026-05-19
[LifetimeSafety] Propagate loans through chained assignment

### LINKED ISSUE #198497: [LifetimeSafety] Track assignment chains within the same statement
https://godbolt.org/z/zE1raqKzY

```cpp
void target() {
    int *s;
    {
        int tgt = 2;
        int *a, *b, *c;
        a = b = c = &tgt;
        s = a;
    }
    printf("s: %d\n", *s);
}
```

There is a use-after-free bug in the code above, but the current implementation fails to detect it.

This happens because `a` and `b` do not hold the corresponding Loan for `tgt`. As a result, if we replace `s = a` with `s = c`, the use-after-free can be correctly identified.

I have conducted a preliminary analysis of this issue, which can be found here: https://github.com/llvm/llvm-project/pull/196075#discussion_r3264392605

NeKon69:
Now, `handleAssignment` also propagates loans into the assignment expression (`BinaryOperator`/`CXXOperatorCallExpr`) so chained assignments can flow loans forward.

Closes #198497

### DIFF
--- a/clang/include/clang/Analysis/Analyses/LifetimeSafety/FactsGenerator.h
+++ b/clang/include/clang/Analysis/Analyses/LifetimeSafety/FactsGenerator.h
@@ -65,6 +65,12 @@ private:
 
   void flow(OriginList *Dst, OriginList *Src, bool Kill);
 
+  /// Handles assignment for both BinaryOperator and CXXOperatorCallExpr.
+  ///
+  /// LHSExpr is the destination whose stored loans are replaced by RHSExpr's
+  /// loans. TargetExpr is the assignment expression itself; it receives
+  /// LHSExpr's origins so chained assignments like `a = b = c` can propagate
+  /// through the result of `b = c`.
   void handleAssignment(const Expr *TargetExpr, const Expr *LHSExpr,
                         const Expr *RHSExpr);
 
--- a/clang/include/clang/Analysis/Analyses/LifetimeSafety/FactsGenerator.h
+++ b/clang/include/clang/Analysis/Analyses/LifetimeSafety/FactsGenerator.h
@@ -65,7 +65,8 @@ private:
 
   void flow(OriginList *Dst, OriginList *Src, bool Kill);
 
-  void handleAssignment(const Expr *LHSExpr, const Expr *RHSExpr);
+  void handleAssignment(const Expr *TargetExpr, const Expr *LHSExpr,
+                        const Expr *RHSExpr);
 
   void handlePointerArithmetic(const BinaryOperator *BO);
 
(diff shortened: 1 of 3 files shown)


clang/lib/Analysis/LifetimeSafety/FactsGenerator.cpp
```diff
@@ -397,7 +398,9 @@ void FactsGenerator::handleAssignment(const Expr *LHSExpr,
   // unlike built-in assignment where LValueToRValue cast strips the outer
   // lvalue origin. Strip it manually to get the actual value origins being
   // assigned.
-  RHSList = getRValueOrigins(RHSExpr, RHSList);
+  if (!(isGslPointerType(LHSExpr->getType()) &&
```
Xazax-hun: I think it might be cleaner to have a separate PR for this change, just in case there are any false positives or other problems it is easier to bisect to a narrower problem.

Xazax-hun (approved):
Overall, looks good to me. But I wonder if this should be split up in 2 PRs.

NeKon69:
I made a little mistake earlier by reusing `View` (it was using an auto generated assignment operator which takes `View`, not `MyObj`), now I define a new struct which accepts a `MyObj` as an argument!

clang/test/Sema/warn-lifetime-safety.cpp
```diff
@@ -107,6 +107,21 @@ void propagation_gsl() {
...
+// FIXME: Should warn!
+void chained_assignment_gsl() {
+  ViewAssign v1, v2;
+  {
+    MyObj s;
+    v2 = v1 = s;
+  }
+  v2.use();
+}
+
```
usx95: This looks unrelated and is a separate bug/limitation. Can you record this in a github issue and add this FIXME separately.
https://godbolt.org/z/ezd31q1aY

clang/include/clang/Analysis/Analyses/LifetimeSafety/FactsGenerator.h
```diff
@@ -65,7 +65,8 @@ class FactsGenerator : public ConstStmtVisitor<FactsGenerator> {
 
   void flow(OriginList *Dst, OriginList *Src, bool Kill);
 
-  void handleAssignment(const Expr *LHSExpr, const Expr *RHSExpr);
```
usx95, line 68: Consider briefly describing the function.

clang/test/Sema/warn-lifetime-safety.cpp
NeKon69: Yeah, I initially added a fix for that in this PR but removed it later. I can remove this test if you find it unrelated here.

usx95 (approved):
Neat. LGTM.

Few nits.

clang/test/Sema/warn-lifetime-safety.cpp
NeKon69: Filed #198605

usx95 (approved):

PR #198514, opened 2026-05-19
[GlobalISel] Add tablegen pattern for shufflevector combine

c-rhodes:
Related to #197693 which filters the worklist to only include opcodes for which there are combines. It's mostly handled by canMatchOpcode which is tablgen'ed but some old combines like this shufflevector one are missing a tablegen pattern and require extra handling. This adds a simple wrapper so it gets picked up by canMatchOpcode and we can delete the C++ handling.

I was expecting this to be NFC but it does change codegen on AMDGPU. I tried re-ordering the patterns so the old shuffle combine comes before this one but it didn't help.

Assisted-by: codex

### DIFF
--- a/llvm/include/llvm/Target/GlobalISel/Combine.td
+++ b/llvm/include/llvm/Target/GlobalISel/Combine.td
@@ -1818,6 +1818,13 @@ def combine_concat_vector : GICombineRule<
         [{ return Helper.matchCombineConcatVectors(*${root}, ${matchinfo}); }]),
   (apply [{ Helper.applyCombineConcatVectors(*${root}, ${matchinfo}); }])>;
 
+// Combines shuffle operations
+def combine_shuffle_vector : GICombineRule<
+  (defs root:$root, register_vector_matchinfo:$matchinfo),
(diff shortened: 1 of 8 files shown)


llvm/include/llvm/Target/GlobalISel/Combine.td
```diff
@@ -1818,6 +1818,13 @@ def combine_concat_vector : GICombineRule<
         [{ return Helper.matchCombineConcatVectors(*${root}, ${matchinfo}); }]),
   (apply [{ Helper.applyCombineConcatVectors(*${root}, ${matchinfo}); }])>;
 
+// Combines shuffle operations
+def combine_shuffle_vector : GICombineRule<
+  (defs root:$root, register_vector_matchinfo:$matchinfo),
+  (match (wip_match_opcode G_SHUFFLE_VECTOR):$root,
```
arsenm, line 1824: Should get away from wip_match_oppose but I suppose this is the minimal effort to get this into tablegen

arsenm (approved):

PR #198516, opened 2026-05-19
[Flang][Parser] Handle compiler directives inside INTERFACE blocks

### LINKED ISSUE #198289: [Flang][Parser] Unrecognized !DIR$ compiler directive inside INTERFACE block causes cascading parse errors
Unrecognized !DIR$ compiler directives placed between interface specifications (i.e., between subroutine/function stubs inside an INTERFACE block) cause fatal cascading parse errors.

[https://godbolt.org/z/f1j1fjdKG](https://godbolt.org/z/f1j1fjdKG)
Reproducer : 

```fortran
module m
  interface
    subroutine ccff
    end subroutine ccff
    !dir$ id "test"
  end interface
end module
```

Output : 

```
/app/example.f90:3:6: error: expected declaration construct
      subroutine ccff
       ^
/app/example.f90:2:3: in the context: specification part
    interface
    ^
/app/example.f90:1:1: in the context: module
  module m
  ^
/app/example.f90:4:9: error: expected end of statement
      end subroutine ccff
          ^
/app/example.f90:4:5: in the context: END MODULE statement
      end subroutine ccff
      ^
/app/example.f90:1:1: in the context: module
  module m
```

chandraghale:
Unrecognized !DIR$ directives between interface specifications currently cause cascading parse errors because the grammar for InterfaceSpecification has no path to consume them. This patch adds CompilerDirective as a valid alternative — matching how InternalSubprogram and ModuleSubprogram already handle this — so that unrecognized directives produce the expected warning instead of a fatal parse failure.

Fixes : [https://github.com/llvm/llvm-project/issues/198289](https://github.com/llvm/llvm-project/issues/198289)


kparzysz:
I'm not sure about this. First of all, the "cascade of errors" is how the parser works.  I agree that it's debatable whether it's useful, but if we want to address that, it should be a consistent approach.  Second, the warning says that an _unrecognized_ directive was ignored, suggesting that there may be known directives that would be accepted.  If we don't accept any directives in that location, the diagnostic should be clearer about it.

chandraghale:
> I'm not sure about this. First of all, the "cascade of errors" is how the parser works. I agree that it's debatable whether it's useful, but if we want to address that, it should be a consistent approach. Second, the warning says that an _unrecognized_ directive was ignored, suggesting that there may be known directives that would be accepted. If we don't accept any directives in that location, the diagnostic should be clearer about it.

Thanks @kparzysz , I understand cascading errors are by design. But this isn't adding error recovery,  it's about the grammar being incomplete. It's extending the grammar to accept directives that the parser already knows how to handle. ModuleSubprogram and InternalSubprogram already include CompilerDirective for the same reason; this just applies the same pattern to InterfaceSpecification.
And The "Unrecognized compiler directive was ignored" message is used uniformly everywhere in the compiler. Location-specific diagnostics would be a separate effort — the goal here is just to avoid a fatal parse error for what should be a warning.

chandraghale:
ping @eugeneepshteyn

flang/test/Parser/compiler-directive-in-interface.f90
```diff
@@ -0,0 +1,25 @@
...
+!CHECK: warning: Unrecognized compiler directive was ignored
+    !dir$ id "test"
+  end interface
+end module
+
+module m2
+  interface
+!CHECK: warning: Unrecognized compiler directive was ignored
+    !dir$ id "test"
+    subroutine foo(a)
```
eugeneepshteyn, line 19: Please also add a case, where the directive is between the two procedures:
```
  interface
    subroutine s1()
    end subroutine
!dir$ ...
    subroutine s2()
    end subroutine
  end interface
```

eugeneepshteyn: Also, might as well add a case with abstract interface and generic interface.

flang/test/Parser/compiler-directive-in-interface.f90
```diff
@@ -0,0 +1,25 @@
+! RUN: %flang_fc1 -fsyntax-only %s 2>&1 | FileCheck %s
+
+! Test that unrecognized compiler directives inside INTERFACE blocks
+! produce a warning rather than a parse error.
+
+module m
```
eugeneepshteyn, line 6: Please also add a test with some recognized directive and check that it survives module serialization and parsing.

flang/include/flang/Parser/parse-tree.h
```diff
@@ -3196,7 +3195,9 @@ struct ProcedureStmt {
 // R1502 interface-specification -> interface-body | procedure-stmt
```
eugeneepshteyn: Please update this and the other R1502 comment below with the new reality (no longer just `interface-body | procedure-stmt`)

flang/test/Semantics/modfile84.f90
```diff
@@ -0,0 +1,42 @@
...
+    end subroutine
+  end interface
+end module
+
+!Expect: m.mod
+!module m
+! interface
+!  subroutine sub(x)
+!   real(4),intent(in)::x
+!   !dir$ ignore_tkr(t) x
```
eugeneepshteyn: Well, `ignore_tkr` is inside subroutine declaration and this worked before this PR. Is there any directive that could be between the procedure declarations that currently could survive serialization to the module? If not, then we can't have such test now.

eugeneepshteyn:
Sorry for requiring more tests (and more comprehensive tests), but with AI generating more code now, we need to have especially good guardrails to minimize bad code escapes in the future.

flang/test/Semantics/modfile84.f90
chandraghale: Thanks!! Right, I think there isn't a recognized directive that would go between procedure declarations and survive module serialization at this point, we can skip such test for now.

chandraghale:
> Sorry for requiring more tests (and more comprehensive tests), but with AI generating more code now, we need to have especially good guardrails to minimize bad code escapes in the future.

Absolutely, that makes total sense !!

flang/test/Parser/compiler-directive-in-interface.f90
chandraghale: done !!

flang/test/Semantics/modfile84.f90
eugeneepshteyn: Ok, please remove this test and then this should be ready to go.

chandraghale: Thanks  @eugeneepshteyn .   Done !!!

eugeneepshteyn (approved):
LGTM, thanks!

PR #198518, opened 2026-05-19
[Analysis][NFC] Reduce memory usage of CodeMetrics

aengelke:
FunctionSpecializer will keep the CodeMetrics of many functions at the
same time, causing the hash table in CodeMetrics to substantially
contribute to overall memory usage. Therefore, replace the DenseMap with
a vector indexed by block number.

On ArrayFunctionsRegistration.bc, this reduces the memory usage at the
time IPSCCP runs from 655.0MiB to 577.5MiB (measured with massif).


dtcxzyw (approved):
LGTM

tgymnich (approved):

PR #198529, opened 2026-05-19
[clangd] Add kernel-doc hover support

elmarco:
Add support for Linux kernel-doc comment format in clangd hover.
This includes parsing kernel-doc structured comments (brief, @param,
Return/Returns, named sections like Context/Note/Warning/Locking),
RST-style indented and fenced code blocks, and inline markup
conversion for %CONSTANT, @param, &struct references, ``literals``,
$ENVVAR, and bare function() references.

Related: https://github.com/clangd/clangd/issues/2662

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/clang-tools-extra/clangd/Config.h
+++ b/clang-tools-extra/clangd/Config.h
@@ -216,6 +216,8 @@ struct Config {
     Markdown,
     /// Treat comments as doxygen.
     Doxygen,
+    /// Treat comments as kernel-doc.
+    KernelDoc,
   };
 
   struct {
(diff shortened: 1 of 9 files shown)


ArcsinX (commented):
Can we implement this functionality using `ConstCommentVisitor` like we do for doxygen comments? Maybe we can reuse some parts of it.

elmarco:
> Can we implement this functionality using `ConstCommentVisitor` like we do for doxygen comments? Maybe we can reuse some parts of it.

ConstCommentVisitor parses Doxygen. Kernel-doc is different enough that it won't produce the right AST nodes: @param_name: becomes an unknown command, %CONSTANT/&struct->member/function() inline markup has no Doxygen equivalent, and section headers like Context: aren't recognized commands.

We could preprocess kernel-doc into Doxygen syntax first, but the inline markup cases contain characters that break Clang's word-boundary scanning for \c/\p, so they'd need a separate handling path anyway. At that point the preprocessing becomes complex enough that it's not clearly simpler than the current dedicated parser.

lygstate:
looks good, what blocks here?

ArcsinX:
> looks good, what blocks here?

I guess we need to come into conclusion [here](https://discourse.llvm.org/t/rfc-clangd-support-for-more-documentation-style/90894) first

PR #198531, opened 2026-05-19
[WebAssembly] Cost model for F32 memory interleave

sparker-arm:
Set them the same as their i32 counterparts, but don't cost 4x v4f32. wasm-perf doesn't show any particular uplift and the tests show that, when we're mixing integers and floats, we often interleaving anyway. But this change should be good for arithmetic on structures of 2xf32.


dschuff (approved):
Thanks!

PR #198534, opened 2026-05-19
[clang] associate newline and function decl. trailing Doxygen comments

aydinmercan:
The clang AST now correctly matches trailing Doxygen comments for function declarations and also comments that come after the declaration with the same `<` trailing directive.


cor3ntin:
This change would need tests. Thanks

aydinmercan:
> This change would need tests. Thanks

On it!

Oh BTW I was meaning the ask but in the comment annotation test [same-line trailing function comments are asserted to not be attached](https://github.com/llvm/llvm-project/blob/main/clang/test/Index/annotate-comments.cpp#L98-L102). The commit in blame doesn't give a reason why it is restricted, by chance do you know/have a clue the reasoning behind it? Thanks! @cor3ntin

PR #198535, opened 2026-05-19
[AllocToken] Propagate !alloc_token through inlined allocation wrappers

melver:
Simple inline allocation wrappers have been preventing AllocToken coverage for such callsites. Instead of converting such inline wrappers to macros, we can support covering standard allocation calls in inline allocation wrappers when inlined.

When an allocation wrapper with malloc/alloc_size attributes is inlined, propagate its call-site !alloc_token onto return-value producing calls exposed by inlining, so the late AllocToken pass assigns the correct type-derived token even in default (non-extended) mode.

Propagation is restricted to calls producing the return-value(s) that the wrapper actually returns. This keeps unrelated calls in the wrapper body unchanged, which matters under -fsanitize-alloc-token-extended, where any !alloc_token-carrying call is instrumented.

### DIFF
--- a/clang/test/CodeGen/alloc-token-inline.c
+++ b/clang/test/CodeGen/alloc-token-inline.c
@@ -1,0 +1,28 @@
+// RUN: %clang_cc1 -O -fsanitize=alloc-token -fsanitize-alloc-token-extended -triple x86_64-linux-gnu -emit-llvm %s -o - | FileCheck %s
+// RUN: %clang_cc1 -O -fsanitize=alloc-token -triple x86_64-linux-gnu -emit-llvm %s -o - | FileCheck %s
+
+typedef __typeof(sizeof(int)) size_t;
+
+void *sink;
+
+void *malloc(size_t size);
+void *external_ptr_helper(void);
+void external_void_helper(void);
+
+__attribute__((malloc, always_inline)) inline void *wrapper(size_t size) {
+  sink = external_ptr_helper();
+  external_void_helper();
+  return malloc(size);
+}
+
+// CHECK-LABEL: @test_inlined_wrapper(
+// CHECK: call ptr @external_ptr_helper()
+// CHECK-NOT: __alloc_token_external_ptr_helper
+// CHECK: call void @external_void_helper()
+// CHECK-NOT: __alloc_token_external_void_helper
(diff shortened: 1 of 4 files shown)


vitalybuka (commented):
What is expected here?
```
// RUN: %clang_cc1 -O -fsanitize=alloc-token -fsanitize-alloc-token-extended -triple x86_64-linux-gnu -emit-llvm %s -o - | FileCheck %s

typedef __typeof(sizeof(int)) size_t;

void *malloc(size_t size);
extern void *external_ptr_helper(void);
extern void external_void_helper(void);

__attribute__((malloc, always_inline)) inline void *wrapper(size_t size) {
  // These should NOT inherit !alloc_token and should NOT be instrumented
  void *p = external_ptr_helper();
  external_void_helper();
  
  // This SHOULD inherit !alloc_token and be instrumented
  return malloc(size);
}

void *sink;

// CHECK-LABEL: @test_inlined_wrapper(
// CHECK: call ptr @external_ptr_helper()
// CHECK-NOT: __alloc_token_external_ptr_helper
// CHECK: call void @external_void_helper()
// CHECK-NOT: __alloc_token_external_void_helper
// CHECK: call{{.*}} @__alloc_token_malloc
void test_inlined_wrapper(void) {
  sink = wrapper(sizeof(int));
}
```

melver:
> What is expected here?
> [...]

Yeah, I just realized this huge oversight myself; with -fsanitize-alloc-token-extended, this would have broken (random functions rewritten).

I think the only reliable option here is actually finding the calls that produce the return value, and only propagating to those.

clang/test/CodeGen/alloc-token-inline.c
```diff
@@ -0,0 +1,28 @@
+// RUN: %clang_cc1 -O -fsanitize=alloc-token -fsanitize-alloc-token-extended -triple x86_64-linux-gnu -emit-llvm %s -o - | FileCheck %s
+// RUN: %clang_cc1 -O -fsanitize=alloc-token -triple x86_64-linux-gnu -emit-llvm %s -o - | FileCheck %s
```
vitalybuka: I suggest --implicit-check-not=__alloc_token


then you can be forced to match all __alloc_token with CHECKS
and avoid  CHECK-NOT: which are order dependent, and can fail to detect reordering

vitalybuka (approved):

PR #198536, opened 2026-05-19
[mlir][affine] Modernize AffineLoopNormalizePass constructor using inherited constructors (NFC)

linuxlonelyeagle:
This PR removes the explicit constructor of AffineLoopNormalizePass and switches to inheriting the base class constructors to utilize promoteSingleIter.

### DIFF
--- a/mlir/lib/Dialect/Affine/Transforms/AffineLoopNormalize.cpp
+++ b/mlir/lib/Dialect/Affine/Transforms/AffineLoopNormalize.cpp
@@ -47,9 +47,3 @@ struct AffineLoopNormalizePass
 };
 
 } // namespace
-
-std::unique_ptr<OperationPass<func::FuncOp>>
-mlir::affine::createAffineLoopNormalizePass(bool promoteSingleIter) {
-  AffineLoopNormalizeOptions options{promoteSingleIter};
-  return std::make_unique<AffineLoopNormalizePass>(options);
-}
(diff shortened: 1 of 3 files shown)


ftynse:
This isn't exactly a "modernization". The base class constructor has an argument of different type, so by deriving it we expose something else as an API. I don't know why would we need this change.

mlir/lib/Dialect/Affine/Transforms/AffineLoopNormalize.cpp
joker-eph: If you want to "modernize", I think there is an opportunity for this to be auto-generated by removing `let constructor = "mlir::affine::createAffineLoopNormalizePass()";` from ODS

linuxlonelyeagle:
To be honest, I'm just a bit confused here. We could have simply inherited the constructor from the base class, yet a constructor was explicitly written out instead. While it's not a big deal for me, is it truly necessary? Probably not. As the PR title suggests, this is an RFC.

mlir/lib/Dialect/Affine/Transforms/AffineLoopNormalize.cpp
krzysz00: Agreed that fixing that would be nice

ftynse:
I could have seen it as modernization, in the sense of going from C++98 to C++11 by using the constructor inheritance mechanism, if the constructor argument types were identical. They are not. The constructor in the derived class used a different type, presumably to decrease the amount of clutter for the callers (no need to construct the options object).

PR #198538, opened 2026-05-19
Define signed int overflow with -fms-compatibility

eleviant:
It seems that unlike clang, MSVC doesn't treat signed integer overflow as UB:

https://godbolt.org/z/GYWdjd3Ex

Patch defines SOW, when -fms-compatibility is used.

### DIFF
--- a/clang/include/clang/Options/Options.td
+++ b/clang/include/clang/Options/Options.td
@@ -4815,7 +4815,7 @@ def fwrapv : Flag<["-"], "fwrapv">, Group<f_Group>,
   Visibility<[ClangOption, CLOption, CC1Option, FlangOption, FC1Option]>,
   HelpText<"Treat signed integer overflow as two's complement">;
 def fno_wrapv : Flag<["-"], "fno-wrapv">, Group<f_Group>,
-  Visibility<[ClangOption, CLOption, FlangOption]>;
+  Visibility<[ClangOption, CC1Option, CLOption, FlangOption]>;
 def fwrapv_pointer : Flag<["-"], "fwrapv-pointer">, Group<f_Group>,
   Visibility<[ClangOption, CLOption, CC1Option, FlangOption, FC1Option]>,
   HelpText<"Treat pointer overflow as two's complement">;
--- a/clang/lib/Driver/ToolChains/Flang.cpp
+++ b/clang/lib/Driver/ToolChains/Flang.cpp
@@ -1269,7 +1269,7 @@ void Flang::ConstructJob(Compilation &C, const JobAction &JA,
   }
 
   renderGlobalISelOptions(D, Args, CmdArgs, Triple);
-  renderCommonIntegerOverflowOptions(Args, CmdArgs);
+  renderCommonIntegerOverflowOptions(Args, CmdArgs, false);
 
   assert((Output.isFilename() || Output.isNothing()) && "Invalid output.");
   if (Output.isFilename()) {
--- a/clang/test/CodeGen/wrapv.c
+++ b/clang/test/CodeGen/wrapv.c
@@ -3,9 +3,12 @@
 // RUN: %clang_cc1 -fno-wrapv -fms-compatibility -emit-llvm %s -o - | FileCheck %s --check-prefix=NOWRAP
 // RUN: %clang_cc1 -fwrapv -emit-llvm %s -o - | FileCheck %s --check-prefix=WRAP
 // RUN: %clang_cc1 -fms-compatibility -emit-llvm %s -o - | FileCheck %s --check-prefix=WRAP
+// RUN: %clang_cc1 -ftrapv -fms-compatibility -emit-llvm %s -o - | FileCheck %s --check-prefix=TRAP
+// RUN: %clang_cc1 -ftrapv -fwrapv -emit-llvm %s -o - | FileCheck %s --check-prefix=TRAP
 
 // NOWRAP: %add = add nsw i32 %0, 1
 // WRAP: %add = add i32 %0, 1
+// TRAP: llvm.sadd.with.overflow
 
 int add1(int x) {
   return x + 1;
--- a/clang/test/CodeGen/wrapv.c
+++ b/clang/test/CodeGen/wrapv.c
@@ -1,0 +1,12 @@
+// RUN: %clang_cc1 -emit-llvm %s -o - | FileCheck %s --check-prefix=NOWRAP
+// RUN: %clang_cc1 -fno-wrapv -emit-llvm %s -o - | FileCheck %s --check-prefix=NOWRAP
+// RUN: %clang_cc1 -fno-wrapv -fms-compatibility -emit-llvm %s -o - | FileCheck %s --check-prefix=NOWRAP
+// RUN: %clang_cc1 -fwrapv -emit-llvm %s -o - | FileCheck %s --check-prefix=WRAP
+// RUN: %clang_cc1 -fms-compatibility -emit-llvm %s -o - | FileCheck %s --check-prefix=WRAP
+
+// NOWRAP: %add = add nsw i32 %0, 1
+// WRAP: %add = add i32 %0, 1
+
+int add1(int x) {
+  return x + 1;
+}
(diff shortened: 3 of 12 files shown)


efriedma-quic (commented):
I think I'd like to have some way to disable this... like, what if the user specifies -fno-wrapv?

And maybe we should have a list of such constructs somewhere in the documentation.

eleviant:
> I think I'd like to have some way to disable this... like, what if the user specifies -fno-wrapv?
> 
> And maybe we should have a list of such constructs somewhere in the documentation.

Addressed comments. I'm not sure if anything except -fwrapv should be implicitly enabled with MSVCCompat.

efriedma-quic:
I think we currently enable `-fno-strict-aliasing` by default for clang-cl?  Not sure if that's tied to -fms-compatibility off the top of my head.

eleviant:
> I think we currently enable `-fno-strict-aliasing` by default for clang-cl? Not sure if that's tied to -fms-compatibility off the top of my head.

Yep. Found a few more and added to the docs

clang/include/clang/Options/Options.td
```diff
@@ -4815,7 +4808,7 @@ def fwrapv : Flag<["-"], "fwrapv">, Group<f_Group>,
   Visibility<[ClangOption, CLOption, CC1Option, FlangOption, FC1Option]>,
   HelpText<"Treat signed integer overflow as two's complement">;
 def fno_wrapv : Flag<["-"], "fno-wrapv">, Group<f_Group>,
-  Visibility<[ClangOption, CLOption, FlangOption]>;
+  Visibility<[ClangOption, CC1Option, CLOption, FlangOption]>;
```
AaronBallman, line 4818: Do we need to make this a cc1 option? The usual approach is that either the driver passes `-fwrapv` to enable wrapping or it doesn't pass it at all and that disables wrapping. CC @jansvoboda11 @MaskRay

clang/test/CodeGen/wrapv.c
```diff
@@ -0,0 +1,12 @@
+// RUN: %clang_cc1 -emit-llvm %s -o - | FileCheck %s --check-prefix=NOWRAP
+// RUN: %clang_cc1 -fno-wrapv -emit-llvm %s -o - | FileCheck %s --check-prefix=NOWRAP
+// RUN: %clang_cc1 -fno-wrapv -fms-compatibility -emit-llvm %s -o - | FileCheck %s --check-prefix=NOWRAP
+// RUN: %clang_cc1 -fwrapv -emit-llvm %s -o - | FileCheck %s --check-prefix=WRAP
+// RUN: %clang_cc1 -fms-compatibility -emit-llvm %s -o - | FileCheck %s --check-prefix=WRAP
```
AaronBallman, line 5: What should happen with `-fms-compatibility -ftrapv` (does the trapping "win" out over the wrapping?) Does the answer change for `-ftrapv -fms-compatibility` because of last-flag-wins?

eleviant: I guess -ftrapv has priority over -fwrapv (even if implictly defined by -fms-compatibility), but I'll double-check
```
  if (Args.hasArg(OPT_ftrapv)) {
    Opts.setSignedOverflowBehavior(LangOptions::SOB_Trapping);
    // Set the handler, if one is specified.
    Opts.OverflowHandler =
        std::string(Args.getLastArgValue(OPT_ftrapv_handler));
  } else if (Args.hasFlag(OPT_fwrapv, OPT_fno_wrapv, Opts.MSVCCompat)) {
    Opts.setSignedOverflowBehavior(LangOptions::SOB_Defined);
  }

```

clang/include/clang/Options/Options.td
eleviant: There was a suggestion from @efriedma-quic to have an ability to override implicitly defined wrapping with `-fno-wrapv`.

clang/test/CodeGen/wrapv.c
AaronBallman: Thanks for checking! I'd say we can add it as a test case and document the behavior.

clang/include/clang/Options/Options.td
AaronBallman: Oh, I think I see why this is needed. The driver passes `-fms-compatibility` as a cc1 flag, so if the user passes `-fms-compatibility -fno-wrapv` to the driver, the cc1 invocation still needs a way to say "ms compatibility but without wrapping" and so we need `-fno-wrapv` as a cc1 flag.

clang/test/CodeGen/wrapv.c
eleviant: Done

AaronBallman (approved):
In general, LGTM though maybe we want to update the PR summary because "define" is a bit too strong without more work (and I don't know if we do or do not want to do that work). e.g., if we define this UB, then it suggests it stops being UB in a constant expression too, but we don't do that for `-fwrapv`: https://godbolt.org/z/vjff1j4jY (neither does MSVC by default of GCC with `-fwrapv`). Not suggesting we need to hold this PR up on that though, because that's a complicated situation (the expectations may differ between C and C++ as well).

clang/lib/Driver/ToolChains/Flang.cpp
```diff
@@ -1269,7 +1269,7 @@ void Flang::ConstructJob(Compilation &C, const JobAction &JA,
   }
 
   renderGlobalISelOptions(D, Args, CmdArgs, Triple);
-  renderCommonIntegerOverflowOptions(Args, CmdArgs);
+  renderCommonIntegerOverflowOptions(Args, CmdArgs, false);
```
shafik, line 1272:
```suggestion
  renderCommonIntegerOverflowOptions(Args, CmdArgs, /*IsMSVCCompat=*/false);
```

shafik: We follow https://clang.llvm.org/extra/clang-tidy/checks/bugprone/argument-comment.html conventions to help better document code.

AaronBallman:
Hello! Our downstream testing at Intel ran into some performance differences that seem to come from these changes. We were wondering if this is unexpected or whether there is some performance we can claw back while still being compatible with cl.

For example, some code can no longer be vectorized (https://godbolt.org/z/7oeETzfY6) because compiler is unable to form the trip count of the loop due to the `-fwrapv` behavior:
```
int x[40];
void doit1(int l, int u) {
  int i;
  #pragma clang loop vectorize(assume_safety)
  for (i = l; i < u; i+= 3) x[i] += 1;
}
```
We also saw an 18% performance regression in SPEC2026 benchmarks such as cpu2026ref 772.marian_r.

Is this expected fallout? Are there things that can be done to claw back some of the performance while still being compatible with cl?

CC @pankaj-chawla

efriedma-quic:
I think it's expected, in general, that optimizations are less powerful without nsw markings.  But for vectorization, we should usually be able to introduce runtime checks; please file bugs for issues you find.

That particular example should be straightforward to optimize, I think.  We can introduce a runtime check for `u<0x7FFFFFD`.  Some brief searching doesn't show an open bug; please file one.

pankaj-chawla:
@efriedma-quic The trip count cannot be computed in this case because the IV wraparound is well defined. It is getting incremented by 3 and we cannot prove that (u-l) is evenly divisible by 3. I am not sure how a range check helps. May be we need a different check like (u-l) % 3 == 0. Can vectorizer even process loops whose trip count cannot be computed by ScalarEvolution? May be if we can add support for such cases in PredicatedScalarEvolution it can work. 

There are other consequences of treating signed integers has having well-defined wraparound behavior like unsigned integers. If the vectorizer cannot prove that the index does not wraparound, it would have to resort to generating gathers/scatters rather than widened loads/stores. 

I am curious. Does -fwrapv have any implications for ABI compatibility or is there any other reason to enable it, except for pure MSVC compatibility?

pankaj-chawla:
>  We can introduce a runtime check for u<0x7FFFFFD

I think I understood how this effectively proves that wrapround doesn't happen.

AaronBallman:
> I think it's expected, in general, that optimizations are less powerful without nsw markings. But for vectorization, we should usually be able to introduce runtime checks; please file bugs for issues you find.
> 
> That particular example should be straightforward to optimize, I think. We can introduce a runtime check for `u<0x7FFFFFD`. Some brief searching doesn't show an open bug; please file one.

Thanks, I filed https://github.com/llvm/llvm-project/issues/208221 to track this.

efriedma-quic:
> Does -fwrapv have any implications for ABI compatibility,

No, it just controls whether we add nsw flags to signed arithmetic.

> If the vectorizer cannot prove that the index does not wraparound, it would have to resort to generating gathers/scatters rather than widened loads/stores.

This seems like something which can use a runtime check.

PR #198539, opened 2026-05-19
[VPlan] Add branch-on-cond false original unconditional latch (NFC).

fhahn:
For loops where the latch does not exit, addInitialSkeleton adds the middle block as additional successor, as early canonicalization.

But then we end up with a block without terminator and multiple successors. Fix this by adding a branch-on-cond false as terminator. This preserves the original behavior (backegdge always taken) and resolves the verifier issue.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
+++ b/llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
@@ -543,7 +543,9 @@ static void addInitialSkeleton(VPlan &Plan, Type *InductionTy,
   // The canonical LatchVPBB has the header block as last successor. If it has
   // another successor, this successor is an exit block - insert middle block on
   // its edge. Otherwise, add middle block as another successor retaining header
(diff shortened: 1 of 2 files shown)


llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
```diff
@@ -550,6 +550,8 @@ static void addInitialSkeleton(VPlan &Plan, Type *InductionTy,
   } else {
     VPBlockUtils::connectBlocks(LatchVPBB, MiddleVPBB);
     LatchVPBB->swapSuccessors();
+    VPBuilder(LatchVPBB).createNaryOp(VPInstruction::BranchOnCond,
+                                      {Plan.getFalse()});
```
lukel97, line 556: Maybe add a comment saying that this will be removed by `addCanonicalIVRecipe`/`createLoopRegions` after #198383?

lukel97 (approved):

llvm/lib/Transforms/Vectorize/VPlanConstruction.cpp
fhahn: I extended the comment above, which already mentions adding the new successor, thanks

PR #196945, opened 2026-05-11
[MLIR][MemRef] Extend narrow-type emulation for dynamic offsets

lialan:
This patch adds three related extensions to the MemRef narrow-type emulation patterns.

* `ConvertMemRefSubview` now accepts a dynamic innermost offset.
* `ConvertMemRefReinterpretCast` is generalized from the previous static-rank-1, static-offset shape to accept any rank and dynamic offsets, with the same alignment contract as the subview pattern.
* A new `ConvertMemRefCast` pattern handles `memref.cast` between equivalent narrow-typed memref types so that emulation does not get blocked by trivial casts.

PR #198541, opened 2026-05-19
[mlir/memref] handle rank-0 contiguous check

lialan:
The old loop initialized its signed cursor with `strides.size() - 1`. For rank-0 memrefs, this
underflowed as `size_t` before the cast. On 32-bit hosts, the value became a large positive index and crashed in SmallVector during narrow-type emulation. 

Change to do early return for rank-0 memref and add coverage tests.

This addresses the issue exposed in tests checked in in https://github.com/llvm/llvm-project/pull/196945,
specifically on x86 32bit system reported in : https://github.com/llvm/llvm-project/pull/196945#issuecomment-4428581243

### DIFF
--- a/mlir/lib/Dialect/MemRef/Utils/MemRefUtils.cpp
+++ b/mlir/lib/Dialect/MemRef/Utils/MemRefUtils.cpp
@@ -24,6 +24,10 @@ bool isStaticShapeAndContiguousRowMajor(MemRefType type) {
   if (!type.hasStaticShape())
     return false;
 
+  int64_t rank = type.getRank();
+  if (rank == 0)
+    return true;
+
   SmallVector<int64_t> strides;
   int64_t offset;
   if (failed(type.getStridesAndOffset(strides, offset)))
@@ -32,7 +36,7 @@ bool isStaticShapeAndContiguousRowMajor(MemRefType type) {
   // MemRef is contiguous if outer dimensions are size-1 and inner
   // dimensions have unit strides.
   int64_t runningStride = 1;
-  int64_t curDim = strides.size() - 1;
+  int64_t curDim = rank - 1;
   // Finds all inner dimensions with unit strides.
   while (curDim >= 0 && strides[curDim] == runningStride) {
     runningStride *= type.getDimSize(curDim);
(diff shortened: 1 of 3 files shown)


mlir/unittests/Dialect/MemRef/MemRefUtilsTest.cpp
```diff
@@ -0,0 +1,44 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "mlir/Dialect/MemRef/Utils/MemRefUtils.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "gtest/gtest.h"
+
+using namespace mlir;
+
+TEST(MemRefUtilsTest, IsStaticShapeAndContiguousRowMajor) {
```
meshtag, line 16: Usually for such tests, a comment describing what this is testing would be useful.

mlir/lib/Dialect/MemRef/Utils/MemRefUtils.cpp
```diff
@@ -32,20 +32,20 @@ bool isStaticShapeAndContiguousRowMajor(MemRefType type) {
   // MemRef is contiguous if outer dimensions are size-1 and inner
   // dimensions have unit strides.
   int64_t runningStride = 1;
-  int64_t curDim = strides.size() - 1;
+  int64_t curDim = static_cast<int64_t>(strides.size());
```
meshtag: A potentially cleaner way to do this is to use the `getRank()` method (you'll not have to change the logic then). IIRC, it returns `int64_t`. 

Ref: https://github.com/llvm/llvm-project/blob/673b17e1b38fbe4dfe01c9aa66c368857b13dfdb/mlir/include/mlir/IR/BuiltinTypeInterfaces.td#L345

meshtag (commented):
Thanks for taking a look. 

I had a first look at the diff. I am not sure if this is the best way to fix the issue though. Why not just trivially early return in `isStaticShapeAndContiguousRowMajor()` for rank 0 memrefs? We don't require all that extra analysis for them.

mlir/unittests/Dialect/MemRef/MemRefUtilsTest.cpp
lialan: done.

lialan:
> Why not just trivially early return in `isStaticShapeAndContiguousRowMajor()` for rank 0 memrefs? 

I think you are right! Simplified the PR.

meshtag (approved):
LGTM, thanks.

meshtag:
Please update the commit message with the current changes (can mention early return). 

Nit: 
In the commit title, consider memref -> mlir/memref so that the change becomes easier to recognize. 

Thanks for the work.

PR #198543, opened 2026-05-19
[SLP] Unify IR flag and metadata propagation in vectorizeTree

alexey-bataev:
Replace scattered propagateIRFlags/propagateMetadata calls with a single
PropagateIRFlags lambda. Excludes copyable scalars from the flag
intersection set and drops nnan/ninf when a copyable lane cannot prove
its operand is non-NaN/non-Inf.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
+++ b/llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
@@ -22634,6 +22634,67 @@ Value *BoUpSLP::vectorizeTree(TreeEntry *E) {
       });
     return IsSigned;
   };
+  auto PropagateIRFlags = [&](Value *V, unsigned Opcode = 0,
+                              ArrayRef<Value *> VL = {}) -> Value * {
+    ArrayRef<Value *> Scalars = VL.empty() ? E->Scalars : VL;
+    SmallSetVector<Value *, 4> UniqueInsts;
+    for (Value *Scalar : Scalars) {
+      auto *I = dyn_cast<Instruction>(Scalar);
+      if (!I)
+        continue;
+      if (E->hasCopyableElements() && E->isCopyableElement(I))
+        continue;
+      UniqueInsts.insert(I);
+    }
+    if (!Opcode)
+      Opcode = E->getOpcode();
+    propagateIRFlags(V, UniqueInsts.getArrayRef(), nullptr, It == MinBWs.end());
+    auto *I = dyn_cast<Instruction>(V);
+    if (!I)
+      return V;
+    I = ::propagateMetadata(I, UniqueInsts.getArrayRef());
+    // For copyable elements the lane is synthesized using a binop identity
+    // value, so the operand at that lane is the copyable scalar's value.
+    // fast-math flags that turn defined NaN/Inf inputs into poison (nnan,
+    // ninf) are only justified if every copyable scalar at the synthesized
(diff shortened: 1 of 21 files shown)


llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -22634,6 +22634,67 @@ Value *BoUpSLP::vectorizeTree(TreeEntry *E) {
...
+      if (E->hasCopyableElements() && E->isCopyableElement(I))
+        continue;
+      UniqueInsts.insert(I);
+    }
+    if (!Opcode)
+      Opcode = E->getOpcode();
+    propagateIRFlags(V, UniqueInsts.getArrayRef(), nullptr, It == MinBWs.end());
+    auto *I = dyn_cast<Instruction>(V);
+    if (!I)
+      return V;
```
bababuck, line 22654: Nit: `propogateIRFlags()` seems to early exit too if `V` is not an instruction.
```suggestion
    auto *I = dyn_cast<Instruction>(V);
    if (!I)
      return V;
    propagateIRFlags(V, UniqueInsts.getArrayRef(), nullptr, It == MinBWs.end());
```

alexey-bataev: It accepts value as input and may proapagate something to constants

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -22634,6 +22634,67 @@ Value *BoUpSLP::vectorizeTree(TreeEntry *E) {
...
+      auto *I = dyn_cast<Instruction>(Scalar);
+      if (!I)
+        continue;
+      if (E->hasCopyableElements() && E->isCopyableElement(I))
+        continue;
+      UniqueInsts.insert(I);
+    }
+    if (!Opcode)
+      Opcode = E->getOpcode();
+    propagateIRFlags(V, UniqueInsts.getArrayRef(), nullptr, It == MinBWs.end());
```
bababuck, line 22651: The AI review I ran flagged this as a possible issue if `UniqueInsts` is empty (there is a blind VL[0] lookup in `propagateIRFlags`. I don't see a case where this could occur but maybe you can think of one?

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
bababuck: Maybe best to leave to be robust to future changes, but I think currently it doesn't pass anything to constants.
```
void llvm::propagateIRFlags(Value *I, ArrayRef<Value *> VL, Value *OpValue,
                            bool IncludeWrapFlags) {
  auto *VecOp = dyn_cast<Instruction>(I);
  if (!VecOp)
    return;
```

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
alexey-bataev: No, it is not possible, there should be at least one element always, this applies only to vector nodes, where we have at least one unique instruction

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
alexey-bataev: Then it is not problem at all, the function does the check itself, later might be extended for something else

bababuck (approved):
LGTM

PR #198544, opened 2026-05-19
[clang][tooling] Handle C++20 annot_module_name in syntax tokens

yronglin:
The token collector will ignore all the annotation tokens, but IIUC, the annot_module_name shoulde not be ignored.


ChuanqiXu9 (approved):
Feel good.

PR #189281, opened 2026-03-29
[CIR] Add coroutine cleanup handling and update co_return semantics

Andres-Salamanca:
This PR adds cleanup handling for coroutine frame destruction. The cleanup is emitted as a conditional that checks the result of the `coro.free` builtin, which is used to determine whether the coroutine frame was heap-allocated, if the returned pointer is null, no destruction is performed. Additionally, this PR changes how co_return is represented: previously, it was lowered directly into a branch to the block containing the final suspend logic, but now a new `cir.coro.body` operation is introduced to represent the user-written coroutine body. Inside this region, `cir.co_return` operations mark exits from the coroutine body and represent structured control flow that transfers execution to the final suspend point. The lowering of this structured control flow into explicit branches is deferred to a future PR in the FlattenCFG pass.

PR #198545, opened 2026-05-19
[CIR] Handle co_return exits in cleanup scope flattening

Men-cotton:
Handle the FlattenCFG follow-up from #189281 by treating cir.co_return as a cleanup-scope exit and preserving it after cleanup code runs.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenCleanup.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenCleanup.cpp
@@ -476,7 +476,11 @@ static void emitCleanup(CIRGenFunction &cgf, cir::CleanupScopeOp cleanupScope,
 static bool bodyHasBranchThroughExits(mlir::Region &bodyRegion) {
   return bodyRegion
       .walk([&](mlir::Operation *op) {
-        if (isa<cir::ReturnOp, cir::GotoOp>(op))
+        // A cir.co_return inside cir.coro.body exits that coroutine body, not
+        // the cleanup scope currently being analyzed.
+        if (isa<cir::CoroBodyOp>(op))
+          return mlir::WalkResult::skip();
+        if (isa<cir::ReturnOp, cir::CoReturnOp, cir::GotoOp>(op))
           return mlir::WalkResult::interrupt();
         return mlir::WalkResult::advance();
       })
--- a/clang/lib/CIR/Dialect/Transforms/FlattenCFG.cpp
+++ b/clang/lib/CIR/Dialect/Transforms/FlattenCFG.cpp
@@ -793,7 +793,11 @@ public:
     // because break and continue never branch out of the loop.
     auto collectExitsInLoop = [&](mlir::Operation *loopOp) {
       loopOp->walk<mlir::WalkOrder::PreOrder>([&](mlir::Operation *nestedOp) {
-        if (isa<cir::ReturnOp>(nestedOp)) {
+        // A cir.co_return inside cir.coro.body exits that coroutine body, not
+        // the cleanup scope currently being analyzed.
+        if (isa<cir::CoroBodyOp>(nestedOp))
+          return mlir::WalkResult::skip();
+        if (isa<cir::ReturnOp, cir::CoReturnOp>(nestedOp)) {
           exits.emplace_back(nestedOp, nextId++);
         } else if (isGotoThatExitsCleanup(nestedOp)) {
           exits.emplace_back(nestedOp, nextId++);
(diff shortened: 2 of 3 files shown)


clang/lib/CIR/Dialect/Transforms/FlattenCFG.cpp
```diff
@@ -793,7 +793,11 @@ class CIRCleanupScopeOpFlattening
     // because break and continue never branch out of the loop.
     auto collectExitsInLoop = [&](mlir::Operation *loopOp) {
       loopOp->walk<mlir::WalkOrder::PreOrder>([&](mlir::Operation *nestedOp) {
-        if (isa<cir::ReturnOp>(nestedOp)) {
+        // A cir.co_return inside cir.coro.body exits that coroutine body, not
+        // the cleanup scope currently being analyzed.
+        if (isa<cir::CoroBodyOp>(nestedOp))
```
andykaylor, line 798: We should never see this. If a cleanup scope op contains a `CoroBodyOp` that operation should be flattened before we flatten the cleanup scope. This should already be handled by `hasNestedOpsToFlatten`. Did you encounter cases where we got here with a CoroBodyOp still in place?

clang/lib/CIR/Dialect/Transforms/FlattenCFG.cpp
```diff
@@ -820,7 +824,10 @@ class CIRCleanupScopeOpFlattening
         } else if (isa<cir::LoopOpInterface>(nestedOp)) {
           collectExitsInLoop(nestedOp);
           return mlir::WalkResult::skip();
-        } else if (isa<cir::ReturnOp, cir::ContinueOp>(nestedOp)) {
+        } else if (isa<cir::CoroBodyOp>(nestedOp)) {
```
andykaylor, line 827: As above

clang/lib/CIR/Dialect/Transforms/FlattenCFG.cpp
```diff
@@ -851,6 +858,8 @@ class CIRCleanupScopeOpFlattening
           collectExitsInCleanup(cast<cir::CleanupScopeOp>(op).getBodyRegion(),
                                 /*ignoreBreak=*/ignoreBreak);
           return mlir::WalkResult::skip();
+        } else if (isa<cir::CoroBodyOp>(op)) {
```
andykaylor, line 861: Again, shouldn't happen

clang/lib/CIR/CodeGen/CIRGenCleanup.cpp
```diff
@@ -476,7 +476,11 @@ static void emitCleanup(CIRGenFunction &cgf, cir::CleanupScopeOp cleanupScope,
 static bool bodyHasBranchThroughExits(mlir::Region &bodyRegion) {
   return bodyRegion
       .walk([&](mlir::Operation *op) {
-        if (isa<cir::ReturnOp, cir::GotoOp>(op))
+        // A cir.co_return inside cir.coro.body exits that coroutine body, not
+        // the cleanup scope currently being analyzed.
+        if (isa<cir::CoroBodyOp>(op))
```
andykaylor, line 481: This is correct. I'm saying so explicitly because of the cases below where the equivalent check isn't needed.

clang/test/CIR/Transforms/flatten-cleanup-scope-coro-return.cir
```diff
@@ -0,0 +1,39 @@
...
+      cir.yield
+    }
+    cir.yield
+  }
+  cir.return
+}
+
+// CHECK-LABEL: cir.func coroutine @co_return_from_cleanup()
+// CHECK-NOT:     cir.cleanup.scope
+// CHECK:         cir.coro.body
```
andykaylor, line 27: This shouldn't survive flattening.

clang/test/CIR/Transforms/flatten-cleanup-scope-coro-return.cir
```diff
@@ -0,0 +1,39 @@
+// RUN: cir-opt %s -cir-flatten-cfg -o %t.cir
+// RUN: FileCheck --input-file=%t.cir %s
+
+cir.func coroutine @co_return_from_cleanup() {
```
andykaylor, line 4: To test this properly, I'd also want a case with a coro.body inside a cleanup scope.

andykaylor (commented):
This change is arriving a bit early. It's not very useful before we have flattening of `CoroBodyOp`.

Andres-Salamanca:
Hi, I’m already working on flattening both CoroBodyOp and AwaitOp. Sorry it’s taking me some time, but the CoroBodyOp part is mostly done and I’m currently working on the await part.

PR #198546, opened 2026-05-19
[CIR] Implement atomic cmp exhange with non-const 'weak' lowering

erichkeane:
This was left as an NYI, but appears in self build!

This patch follows the existing solution in that we are doing the branching of weak vs not-weak at the CIR level. This is necessary because the LLVM intrinsics (and the CIR operaions) take 'weak' as a constant value.

Unlike classic-codegen, this patch uses an 'if' instead of a 'switch' on the 'weak' value. This is mainly for readability (since it is a switch
    over a bool!), but also because our 'switch' doesn't seem to support
'bool', so this would require an additional cast.

As a future direction, we may wish to modify the CIR operations to take 'weak' and 'failure' value (both are constants in LLVM intrinsics!) as non-constants, and handle the switch/if statement during lowering.  This would give us an opportunity to optimize the value out without having to collapse the if/switch/etc, and minimize the size of the CIR.  However, as that is a larger direction, this patch skips that for now.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenAtomic.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenAtomic.cpp
@@ -523,12 +523,13 @@ static void emitAtomicCmpXchgFailureSet(CIRGenFunction &cgf, AtomicExpr *e,
       });
 }
 
-static void emitAtomicCmpXchgFailureSet(CIRGenFunction &cgf, AtomicExpr *e,
-                                        Expr *isWeakExpr, Address dest,
-                                        Address ptr, Address val1, Address val2,
-                                        Expr *failureOrderExpr, uint64_t size,
-                                        cir::MemOrder successOrder,
-                                        cir::SyncScopeKind scope) {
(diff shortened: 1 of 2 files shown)


clang/lib/CIR/CodeGen/CIRGenAtomic.cpp
```diff
@@ -523,6 +523,37 @@ static void emitAtomicCmpXchgFailureSet(CIRGenFunction &cgf, AtomicExpr *e,
       });
 }
 
+static void emitAtomicCmpXchgFailureSet(CIRGenFunction &cgf, AtomicExpr *e,
```
andykaylor: Giving this the same name as the existing static function confused me, especially given that classic codegen only has one version of this and implemented the switching logic at the call site. I do think moving the switching logic here is an improvement, but I'd prefer a unique name, even if it was something like `emitAtomicCmpXchgFailureSetHelper`.

andykaylor (approved):
lgtm -- one suggestion on the function name, but I leave that to your discretion.

andykaylor (approved):

PR #198548, opened 2026-05-19
Fix use-after-free in IOTransport::OnRead on client disconnect

youngd007:
When an MCP client disconnects (EOF), `IOTransport::OnRead` called
  `handler.OnClosed()` before resetting `m_read_handle`. The MCP server's
  `OnClosed` handler erases the client from `m_instances`, destroying both
  the transport (`this`) and the binder (`handler`). The subsequent
  `m_read_handle.reset()` then accessed the destroyed transport's member,
  causing a use-after-free (SIGSEGV).

* thread #1, stop reason = signal SIGSEGV: address not mapped to object (fault address=0x28)
   * frame #0: 0x00007ff5d4d5afda liblldb.so.23.2`lldb_private::transport::IOTransport<lldb_protocol::mcp::ProtocolDescriptor>::OnRead(lldb_private::MainLoopBase&, lldb_private::transport::JSONTransport<lldb_protocol::mcp::ProtocolDescriptor>::MessageHandler&) + 1274
     frame #1: 0x00007ff5d1140ad8 liblldb.so.23.0`lldb_private::MainLoopPosix::Run() + 408
     frame #2: 0x00007ff5d1760c1c liblldb.so.23.0`std::thread::_State_impl<std::thre

  Fix by resetting the read handle before calling `OnClosed()`, so no
  transport members are accessed after the handler potentially destroys
  the transport.

Then when the scope is left, the destructor is called for the new read_handle local variable and it is cleaned up.

New unit tests added that fail without this change.  With the change, the custom 'ai' script (allows end user locally to communicate lldb context to agent backend via a spun up MCP server: "protocol-server start MCP listen://localhost:{port}") now successfully concludes without this crash

Assisted with: claude

### DIFF
--- a/lldb/include/lldb/Host/JSONTransport.h
+++ b/lldb/include/lldb/Host/JSONTransport.h
@@ -259,9 +259,10 @@ private:
       if (!m_buffer.empty())
         handler.OnError(llvm::make_error<TransportUnhandledContentsError>(
             std::string(m_buffer.str())));
-      handler.OnClosed();
-      // On EOF, remove the read handle from the MainLoop.
+      // Remove the read handle from the MainLoop before notifying the handler,
+      // because OnClosed may destroy this transport (and the handler).
       m_read_handle.reset();
+      handler.OnClosed();
     }
   }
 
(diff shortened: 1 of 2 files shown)


JDevlieghere (approved):
Thanks!

PR #197251, opened 2026-05-12
[lldb][windows] enable TestMultipleTargets.py and TestSBCommandReturnObject.py

charles-zablit:
These 2 tests are passing on Windows. Enable them.

PR #198550, opened 2026-05-19
[lldb][windows] add bin to PATH in API tests

charles-zablit:
API tests that build drivers which link against `liblldb.dll` fail on Windows because `liblldb.dll` is not in the PATH. This does not happen on macOS and Linux because they use rpaths, which do not exist on Windows.

This is a reland of https://github.com/llvm/llvm-project/pull/197251 with the `bin` build directory added to the PATH env variable in the API tests.

We could look into enabling the following tests as well, however they do not run because they use POSIX APIs that do not exist on Windows:
- `lldb-api :: api/multiple-debuggers/TestMultipleDebuggers.py`
- `lldb-api :: api/multithreaded/TestMultithreaded.py`

rdar://177435313

### DIFF
--- a/lldb/test/API/api/command-return-object/TestSBCommandReturnObject.py
+++ b/lldb/test/API/api/command-return-object/TestSBCommandReturnObject.py
@@ -12,9 +12,6 @@ class TestSBCommandReturnObject(TestBase):
     NO_DEBUG_INFO_TESTCASE = True
 
     @skipIfNoSBHeaders
-    @expectedFailureAll(
-        oslist=["windows"], archs=["i[3-6]86", "x86_64"], bugnumber="llvm.org/pr43570"
-    )
     @skipIfHostIncompatibleWithTarget
     def test_sb_command_return_object(self):
         self.driver_exe = self.getBuildArtifact("command-return-object")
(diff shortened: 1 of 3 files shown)


DavidSpickett:
> This is a reland of https://github.com/llvm/llvm-project/pull/197251.

Just FYI, it wasn't immediately obvious what part was new to the reland. As the "This" of "This is a reland" could refer to the PR or to the previous sentences.

"This PR is a reland of #N, with X and Y added" would be easier to understand.

lldb/test/API/api/command-return-object/TestSBCommandReturnObject.py
```diff
@@ -12,9 +12,6 @@ class TestSBCommandReturnObject(TestBase):
     NO_DEBUG_INFO_TESTCASE = True
 
     @skipIfNoSBHeaders
-    @expectedFailureAll(
-        oslist=["windows"], archs=["i[3-6]86", "x86_64"], bugnumber="llvm.org/pr43570"
```
DavidSpickett, line 16: Looks like you can close https://github.com/llvm/llvm-project/issues/42915 once this goes in.

DavidSpickett (approved):
Adding to PATH does look like the simplest option, and I can't think of anything it would break given that it's only there during the test run.

PR #198552, opened 2026-05-19
[mlir][mem2reg] fix 197158 by moving visitReplacedValues call

### LINKED ISSUE #197158: [mlir][mem2reg] use of deleted SSA value when a load is stored back in presence dbg.declare
Reproducer:

```mlir
#di_basic_type    = #llvm.di_basic_type<tag = DW_TAG_base_type, name = "double", sizeInBits = 64>
#di_file          = #llvm.di_file<"repro.ll" in "">
#di_compile_unit  = #llvm.di_compile_unit<id = distinct[0]<>, sourceLanguage = DW_LANG_C_plus_plus_14, file = #di_file, producer = "clang", isOptimized = false, emissionKind = Full>
#di_subprogram    = #llvm.di_subprogram<compileUnit = #di_compile_unit, scope = #di_file, name = "foo", linkageName = "foo", file = #di_file, line = 1, subprogramFlags = Definition>
#di_local_variable = #llvm.di_local_variable<scope = #di_subprogram, name = "d", file = #di_file, line = 2, type = #di_basic_type>
llvm.func @foo() {
  %one = llvm.mlir.constant(1 : i32) : i32
  %cst = llvm.mlir.constant(4.000000e+00 : f64) : f64
  %p   = llvm.alloca %one x f64 : (i32) -> !llvm.ptr
  llvm.intr.dbg.declare #di_local_variable = %p : !llvm.ptr
  llvm.store %cst, %p : f64, !llvm.ptr
  %v = llvm.load %p : !llvm.ptr -> f64
  llvm.store %v, %p : f64, !llvm.ptr
  llvm.return
}
```

`mlir-opt -mem2reg` from the main branch crashes with:

```
repro.mlir:12:8: error: 'llvm.load' op operation destroyed but still has uses
  %v = llvm.load %p : !llvm.ptr -> f64
       ^
repro.mlir:12:8: note: see current operation: %0 = "llvm.load"(<<UNKNOWN SSA VALUE>>) <{ordering = 0 : i64}> : (!llvm.ptr) -> f64
repro.mlir:10:3: note: - use: "llvm.intr.dbg.value"( ...

  llvm.intr.dbg.declare #di_local_variable = %p : !llvm.ptr
  ^
LLVM ERROR: operation destroyed but still has uses
Stack dump:
0.	Program arguments: mlir-opt -mem2reg repro.mlir
#11 0x000000000d882c41 mlir::Operation::~Operation() (mlir-opt+0xd882c41)
#12 0x000000000d882d89 mlir::Operation::erase() (mlir-opt+0xd882d89)
#13 0x000000000d3f59ff mlir::tryToPromoteMemorySlots(llvm::ArrayRef<mlir::PromotableAllocationOpInterface>, mlir::OpBuilder&, mlir::DataLayout const&, mlir::DominanceInfo&, mlir::Mem2RegStatistics) (mlir-opt+0xd3f59ff)
#14 0x000000000d3f7400 (anonymous namespace)::Mem2Reg::runOnOperation() Mem2Reg.cpp:0:0
```

I think the issue is that `visitReplacedValues` [here](https://github.com/llvm/llvm-project/blob/c2d51a2719026acd30d505cf11d7f33bc27adafd/mlir/lib/Transforms/Mem2Reg.cpp#L925) is called with `replacedValues` that were gathered on the llvm.store operands before the llvm.load results are replaced, and `replacedValues` is not being updated when replacing the load op results.

Hence, when the llvm.dbg.declare `visitReplacedValues` is called and creates llvm.intr.dbg.value for the replaced values, it introduces uses of llvm.load results that have been deleted.

Note that FIR handling of debug value suffer the same flaw, I think the infra should call `visitReplacedValues` with `replacedValues` that have been updated as needed after replacing the load operations.

jeanPerier:
Fix #197158 and #200844 by moving the `visitReplacedValues` calls between `promoteInRegion` and `removeBlockingUses` , as well as setting the insertion point before the replaced store operation before calling the `PromotableMemOpInterface::getStored` API (instead of setting the insertion point after).

The action order change is done at the top level. The `promoteInRegion` are done for all regions in post order, then the `visitReplacedValues` are done for all regions, and then only the `removeBlockingUses` are done for all regions in post order. This ensures that any load results that would happen to be used in a later stored is not deleted by `removeBlockingUses` before it is used by  `visitReplacedValues`.

The insertion point change ensures that the stored values passed to `visitReplacedValues` dominate the related store operations. Otherwise, typical `visitReplacedValues` that set insertion points at the store operation and use the stored values generated invalid IR when `getStored` generates new IR (like bitcasts for the LLVM dialect implementation).


fabianmcg (approved):
In general LGTM! It'd be great to add more testing cases (potentially a test op that implements the iface's method), to have more coverage. Please wait for @Moxinilian

jeanPerier:
Thanks for the review @fabianmcg, I added more tests.

mlir/lib/Transforms/Mem2Reg.cpp
```diff
@@ -835,8 +838,32 @@ void MemorySlotPromoter::removeBlockingUses(Region *region) {
...
+void MemorySlotPromoter::visitReplacedValuesForRegion(Region *region) {
+  auto *blockingUsesMapIt = info.userToBlockingUses.find(region);
+  if (blockingUsesMapIt == info.userToBlockingUses.end())
+    return;
+  BlockingUsesMap &blockingUsesMap = blockingUsesMapIt->second;
+  if (blockingUsesMap.empty())
+    return;
+
+  // Build the replaced-values list from per-store snapshots. These values
+  // are still live. Since this runs before `removeBlockingUses` replaces
```
Moxinilian: nit: did you mean

```suggestion
  // are still valid. Since this runs before `removeBlockingUses` replaces
```
(in contrast to live as in value liveness)

mlir/include/mlir/Interfaces/MemorySlotInterfaces.td
```diff
@@ -235,29 +235,28 @@ def PromotableOpInterface : OpInterface<"PromotableOpInterface"> {
...
+        Transforms the IR using the SSA values replacing the memory slot.
 
-        This method will only be called after all blocking uses have been
-        scheduled for removal and if `requiresReplacedValues` returned
-        true.
+        Called after computing reaching definitions but before removing
+        blocking uses, provided `requiresReplacedValues` returns true.
+        The `mutatedDefs` reference the original IR. Any new uses of a load
+        result created here will be automatically redirected when the
+        framework replaces that load with its reaching definition.
```
Moxinilian: I think a missing new aspect of the contract is that the function will not be called with *all* the replaced values. When I moved this callback around while adding region control flow support, I did not realize this was an acceptable tradeoff. Could you mention it explicitly in the documentation? I think we can maybe claim that the callback will be per-region, but that's probably not necessary.

mlir/test/Transforms/mem2reg.mlir
```diff
@@ -318,3 +318,143 @@ func.func @promotable_through_dual_alias(%a: si32) -> ui32 {
...
+  return %v : i32
+}
+
+// -----
+
+// Regression test for use-of-deleted-SSA-value bug (similar to the
+// `dbg.declare` reproducer in `mem2reg-dbginfo.mlir`). A value loaded from
+// the slot is stored back to it. `visitReplacedValues` must run before
+// `replaceAllUsesWith` so the new `tracked_value` references the live load
+// result, which is then redirected to the original store's value.
```
Moxinilian: Could you reword this so it doesn't reference the content of your Claude context, specifically not pointing to the LLVM implementation explicitly? (from a new reader's perspective it makes no sense, I really hate how Claude does that all the time...)

Moxinilian (approved):
I don't use this feature so I trust Fabian on whether this is acceptable change in the contract. The implementation looks good to me provided the documentation comments.

mlir/include/mlir/Interfaces/MemorySlotInterfaces.td
jeanPerier: That is a good point, that may be OK for the case of dbg.declare, but looking at it more closely, this new behavior would be a bit hard to describe since it is actually forwarding all the values from the region already visited, not only the ones from the values in the current region.
Hence, I rather revert to the previous behavior by updating the code to do separate region walks for the promotion/visit/removal. I did this in the updated patch.

mlir/test/Transforms/mem2reg.mlir
jeanPerier: Yeah, I hate that too. I trimmed/rephrased the comments.

jeanPerier:
Thanks for the reviews @Moxinilian and @fabianmcg. Sorry for the late update, testing this patch I found another [issue](https://github.com/llvm/llvm-project/issues/200844) with `visitReplacedValues`  that had to do with dominance expectation for the `replacedValues` that were not true given the insertion points set in Mem2Reg framework.

I updated this patch to fix this issue by changing `setInsertionPointAfter` in `setInsertionPoint` in `promoteInBlock` to ensure that the value returned by ` memOp.getStored` dominates `memOp`.

I also updated the patch to address the fact that my first update was now calling `visitReplacedValues` with only a sub group of the replaced value. I moved the visit/removal into their own region visit ( the visited region are stored during the first visit that does the promotion into a new `regionsInPostOrder` vector).

fabianmcg (approved):
LGTM!, Thank you!

mlir/lib/Transforms/Mem2Reg.cpp
```diff
@@ -583,7 +595,10 @@ Value MemorySlotPromoter::promoteInBlock(Block *block, Value reachingDef) {
       MemorySlot aliasSlot =
           getOpAliasSlot(memOp, slot, info.aliasMap).value_or(slot);
       if (memOp.storesTo(aliasSlot)) {
-        builder.setInsertionPointAfter(memOp);
+        // Insert helper IR introduced by `getStored` before the storing op.
+        // This ensures the returned value dominates the store, allowing
+        // `visitReplacedValues` to safely create new uses after the store.
```
tdegioanni-nvidia:
```suggestion
        // Ensures helper IR introduced by `getStored` is created before the
        // storing op. This ensures the provided stored value dominates the
        // store, allowing `visitReplacedValues` to safely create new uses
        // after the store.
```

mlir/lib/Transforms/Mem2Reg.cpp
```diff
@@ -967,26 +994,29 @@ void MemorySlotPromoter::removeUnusedItems() {
 
 std::optional<PromotableAllocationOpInterface>
 MemorySlotPromoter::promoteSlot() {
-  // Perform the promotion recursively through nested regions. The reaching
-  // definition starts with a null value that will be replaced by a
+  // Pass 1: perform the promotion recursively through nested regions. The
```
tdegioanni-nvidia:
```suggestion
  // Step 1: perform the promotion recursively through nested regions. The
```

mlir/lib/Transforms/Mem2Reg.cpp
```diff
@@ -967,26 +994,29 @@ void MemorySlotPromoter::removeUnusedItems() {
...
 
-  // Blocking uses can then be removed for the outermost region.
-  removeBlockingUses(slot.ptr.getParentRegion());
-
-  // Notify operations that requested it of the reaching definitions set by
-  // storing memory operations.
-  for (PromotableOpInterface op : toVisitReplacedValues) {
-    builder.setInsertionPointAfter(op);
-    op.visitReplacedValues(replacedValues, builder);
+  // Pass 2: call `visitReplacedValues` on operations that requested it.
```
tdegioanni-nvidia:
```suggestion
  // Step 2: call `visitReplacedValues` on operations that requested it.
```

mlir/lib/Transforms/Mem2Reg.cpp
```diff
@@ -967,26 +994,29 @@ void MemorySlotPromoter::removeUnusedItems() {
...
+    SmallVector<std::pair<Operation *, Value>> replacedValues;
+    replacedValues.reserve(replacedValuesMap.size());
+    for (const auto &[memOp, value] : replacedValuesMap)
+      if (value)
+        replacedValues.push_back({memOp, value});
+    for (Region *region : regionsInPostOrder)
+      visitReplacedValuesForRegion(region, replacedValues);
   }
 
+  // Pass 3: remove the slot's blocking uses across all regions. Iterating
```
tdegioanni-nvidia:
```suggestion
  // Step 3: remove the slot's blocking uses across all regions. Iterating
```

tdegioanni-nvidia (commented):
Mostly doc comments, thanks!

Moxinilian (approved):

PR #198555, opened 2026-05-19
[AMDGPU] Skip bridge copies for grouped MFMA instructions

MrSidims:
When the MFMA-form rewrite moves a chain of matrix ops from VGPR to AGPR form, the bridges between ops in the chain don't exist post-rewrite, instead they all live in AGPR-form together. The cost model was counting those phantom bridges and rejecting rewrites that were actually profitable.

Treat the candidate set as a group. Edges inside the group need no bridge. Skip candidates whose results escape to non-matrix users.

Co-author: Vlad Romanov <Vladislav.Romanov@amd.com>


llvm/lib/Target/AMDGPU/GCNSchedStrategy.cpp
```diff
@@ -2221,11 +2228,13 @@ void GCNSchedStage::modifyRegionSchedule(unsigned RegionIdx,
       }
       ++RegionEnd;
     }
-    if (MI->isDebugInstr()) {
-      LLVM_DEBUG(dbgs() << "Scheduling " << *MI);
-      continue;
-    }
+    LLVM_DEBUG(dbgs() << "Scheduling " << *MI);
```
shiltian: This change doesn't seem to be related, or correct. Why do we need not to skip the debug intr here now but instead do it below?

llvm/lib/Target/AMDGPU/GCNSchedStrategy.cpp
```diff
@@ -2248,11 +2256,31 @@ void GCNSchedStage::modifyRegionSchedule(unsigned RegionIdx,
   DAG.Regions[RegionIdx].first = MIOrder.front();
 }
 
-bool RewriteMFMAFormStage::isRewriteCandidate(MachineInstr *MI) const {
+/// Returns true when \p RD will already be in AGPR-form after the rewrite, so
+/// no bridge copy is needed at this reaching definition.
+static bool reachingDefStaysAGPR(MachineInstr *RD,
```
shiltian: Function name doesn't conform with LLVM code standard. It is supposed to start with a verb.

llvm/lib/Target/AMDGPU/GCNSchedStrategy.cpp
```diff
@@ -2248,11 +2256,31 @@ void GCNSchedStage::modifyRegionSchedule(unsigned RegionIdx,
...
   if (!static_cast<const SIInstrInfo *>(DAG.TII)->isMAI(*MI))
     return false;
-  return AMDGPU::getMFMASrcCVDstAGPROp(MI->getOpcode()) != -1;
+  if (AMDGPU::getMFMASrcCVDstAGPROp(MI->getOpcode()) == -1)
+    return false;
+  // Reject candidates whose users force an unavoidable bridge copy.
+  Register DstReg = MI->getOperand(0).getReg();
+  for (const MachineOperand &Use : DAG.MRI.use_nodbg_operands(DstReg))
+    if (!TII->isMAI(*Use.getParent()) && !Use.getParent()->isCopy())
+      return false;
```
shiltian:
```suggestion
  for (const MachineOperand &Use : DAG.MRI.use_nodbg_operands(DstReg)) {
    if (!TII->isMAI(*Use.getParent()) && !Use.getParent()->isCopy())
      return false;
  }
```

llvm/lib/Target/AMDGPU/GCNSchedStrategy.cpp
```diff
@@ -2248,11 +2256,31 @@ void GCNSchedStage::modifyRegionSchedule(unsigned RegionIdx,
...
 
-bool RewriteMFMAFormStage::isRewriteCandidate(MachineInstr *MI) const {
+/// Returns true when \p RD will already be in AGPR-form after the rewrite, so
+/// no bridge copy is needed at this reaching definition.
+static bool reachingDefStaysAGPR(MachineInstr *RD,
+                                 const DenseSet<Register> &CandSrc2Regs,
+                                 const SIInstrInfo &TII) {
+  if (TII.isMAI(*RD))
+    return true;
+  if (RD->getOpcode() == AMDGPU::AV_MOV_B32_IMM_PSEUDO)
```
kurapov-peter: Is the logic 32b specific?

llvm/lib/Target/AMDGPU/GCNSchedStrategy.cpp
MrSidims: Renamed, thanks!

llvm/lib/Target/AMDGPU/GCNSchedStrategy.cpp
MrSidims: It's NFC change, reverted.

kurapov-peter (approved):
Thanks! Could you please enable the stage in co-exec scheduler once this is in?

llvm/lib/Target/AMDGPU/GCNSchedStrategy.cpp
MrSidims: Added b64 check as well as missing tests for b32 and b64. Thanks!

PR #198556, opened 2026-05-19
[AMDGPU] Fix wrong truth table in BitOp3_Op for shared sub-expressions.

carlobertolli:
When the LHS and RHS of a boolean node share a common sub-expression, LHS recursion can decompose that shared node via the "replace parent operator" mechanism in getOperandBits (Src[I] = Op where Src[I] == In). This replaces the Src slot that RHS was originally mapped to with a sub-operand. If RHS recursion then fails to re-derive its own decomposition, RHSBits still refers to the old slot index — but the slot now holds a different value, producing an incorrect truth table.

Fix this by tracking which Src slot RHS occupies before LHS recursion. If RHS recursion fails and that specific slot was modified, roll back Src and the bit assignments to the pre-recursion state.

Add a test to show a specific example.

Assisted-by: Cursor (Claude)


shiltian:
This seems to be the 3rd PR working on something similar…

- https://github.com/llvm/llvm-project/pull/198373
- https://github.com/llvm/llvm-project/pull/198419

carlobertolli:
> This seems to be the 3rd PR working on something similar…
> 
> * [[AMDGPU] Fix BITOP3 matcher for op(X, X) operands #198373](https://github.com/llvm/llvm-project/pull/198373)
> * [[AMDGPU] Fix BitOp3_Op duplicate Src entry in replace-parent path #198419](https://github.com/llvm/llvm-project/pull/198419)

The first PR solves a different problem: the check if (LHS == RHS) is never true for the examples in this PR.
The second PR seems to be solving the same problem, but it does not fix the issue when RHS recursion overwrites LHS sources (when it fails). I extended this PR to handle that case as well.

carlobertolli:
ping

llvm/lib/Target/AMDGPU/AMDGPUISelDAGToDAG.cpp
```diff
@@ -4401,16 +4401,103 @@ static std::pair<unsigned, uint8_t> BitOp3_Op(SDValue In,
...
+    // We detect this in three ways:
+    // (A) If LHS recursed, its truth table is valid against the Src state
+    //     when LHS recursion completed (SrcAfterLHS). If RHS recursion
+    //     then mutates a Src slot that LHSBits depends on, LHSBits is
+    //     stale.
+    // (B) If RHS did not recurse, RHSBits came from getOperandBits and
+    //     refers to a specific Src slot. If that slot's contents changed
+    //     (by either recursion), RHSBits is stale.
+    // (C) Symmetrically for LHS if it did not recurse.
+    SmallVector<SDValue, 3> SrcBeforeRecurse(Src.begin(), Src.end());
```
arsenm, line 4426: I don't really like letting something wrong happen and then trying to figure it out to undo later. Is there just a missing hasOneUse check in the analysis?

carlobertolli: I agree with the aesthetic bit, but not allowing multiple uses of a slot is going to make the algorithm miss out on important cases where there is sharing between LHS and RHS. Example:

%t = and i32 %a, %b       ; used twice
%x = xor i32 %t, %c
%r = or  i32 %x, %t

With the current patch, %t decomposes into %a and %b, everything fits in 3 Src slots [%a, %b, %c]. We end up with a single v_bitop3 instruction.

With a hasOneUse guard, %t has 2 uses, so we refuse to decompose it, and we end up with a v_and and v_or.

It's only by doing a full LHS and RHS recursion that we can check whether we ended up in a pathological case.

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
```diff
@@ -4246,17 +4246,53 @@ static std::pair<unsigned, uint8_t> BitOp3_Op(Register R,
...
+    // Check for cross-contamination between the two recursions.
+    bool RHSStale =
+        !Op.first && RHSSlot >= 0 && Src[RHSSlot] != SrcBeforeRecurse[RHSSlot];
+    bool LHSStale = LHSSlot >= 0 && LHSSlot < (int)Src.size() &&
+                    Src[LHSSlot] != SrcAfterLHS[LHSSlot];
+    if (RHSStale || LHSStale) {
+      Src = std::move(SrcBeforeRecurse);
+      LHSBits = LHSBitsOrig;
+      NumOpcodes = 0;
+    }
```
doru1004: Hmm can LHSStale be true regardless of whether RHS recursed? If RHS did recurse and LHSStale fires, RHSBits now holds the recursed truth table while Src has been rolled back to SrcBeforeRecurse. You might want RHSBitsOrig and then restore it.

carlobertolli: good catch: "If RHS did recurse and LHSStale fires, RHSBits now holds the recursed truth table while Src has been rolled back".
This is precisely what is solved in the SDAG version (AMDGPUISelDAGToDAG.cpp) with RHSBitsOrig.
I have now aligned the two versions. Bad that we have to duplicate code, because this type of thing will keep happening.
I have not explored having a common code for this. I am not sure whether this would make sense given that these are two alternative implementations?

llvm/lib/Target/AMDGPU/AMDGPUInstructionSelector.cpp
```diff
@@ -4309,16 +4246,107 @@ static std::pair<unsigned, uint8_t> BitOp3_Op(Register R,
...
+    // the ~SrcBits[I] shortcut.
+    const uint8_t SrcBitsConst[3] = {0xf0, 0xcc, 0xaa};
+    auto findSlot = [&](uint8_t Bits, Register Op,
+                        const SmallVectorImpl<Register> &S) -> int {
+      for (int I = 0; I < (int)S.size(); I++) {
+        if (Bits == SrcBitsConst[I] && S[I] == Op)
+          return I;
+        if (Bits == (uint8_t)~SrcBitsConst[I]) {
+          Register Inner;
+          if (mi_match(Op, MRI, m_Not(m_Reg(Inner)))) {
```
doru1004: This is loop independent and can be hoisted outside the loop.

carlobertolli: done

doru1004 (approved):
LGTM

doru1004:
Pending all checks pass

PR #195380, opened 2026-05-01
[NFC][LLVM] Make `matchIntrinsicSignature` static and rename it

jurahul:
Make `matchIntrinsicSignature` static, rename it to `isSignatureValid`, and change the sense of its return value to match the new name.

PR #198557, opened 2026-05-19
[LLVM][Tablegen] Add Default arguments support for Intrinsics in TableGen

varadk27:
This patch adds LLVM infrastructure to support default argument values for intrinsics, modeled after C++ default arguments. 
The motivation is to simplify the evolution of intrinsic signatures by removing the hand-written AutoUpgrade boilerplate that intrinsic authors must repeat for every new trailing parameter.

This feature extends the existing TableGen `ImmArg<>` property with an optional `DefaultValue<V>`, written as `ImmArg<ArgIndex<N>, DefaultValue<V>>`, so an intrinsic author can declare a default value alongside an immediate argument. When older .ll or .bc IR is loaded, AutoUpgrade fills in the missing trailing arguments from the TableGen-generated default-value table ; no per-intrinsic hand-written upgrade case is required.

Link to the RFC, where this feature will be discussed:
https://discourse.llvm.org/t/rfc-default-argument-support-for-llvm-intrinsics/90839

### DIFF
--- a/llvm/include/llvm/IR/Intrinsics.td
+++ b/llvm/include/llvm/IR/Intrinsics.td
@@ -130,9 +130,24 @@ class Returned<ArgIndex idx> : IntrinsicProperty {
   int ArgNo = idx.Value;
 }
 
-// ImmArg - The specified argument must be an immediate.
-class ImmArg<ArgIndex idx> : IntrinsicProperty {
+// DefaultVal - A value carrier paired with ImmArg to declare a default for
+// a missing trailing argument. AutoUpgrade fills the default when an old
+// .bc / .ll file is loaded.
+class DefaultVal<int val> {
+  bit HasDefault = 1;
+  int Value = val;
+}
+
+// Sentinel — used as the default for ImmArg's optional DefaultVal parameter.
+def NoDefault : DefaultVal<0> {
+  let HasDefault = 0;
+}
+
+// ImmArg - The specified argument must be an immediate. The optional
+// second template parameter declares a default value for AutoUpgrade.
+class ImmArg<ArgIndex idx, DefaultVal val = NoDefault> : IntrinsicProperty {
   int ArgNo = idx.Value;
+  DefaultVal Default = val;
 }
 
 // ReadOnly - The specified argument pointer is not written to through the
@@ -178,16 +193,6 @@ class ImmArgPrinter<string funcname> : ArgProperty {
 class ArgInfo<ArgIndex idx, list<ArgProperty> arg_properties> : IntrinsicProperty {
   int ArgNo = idx.Value;
   list<ArgProperty> Properties = arg_properties;
-}
-
-// DefaultIntArg - Attach an integer default value to an intrinsic argument.
-// The argument must be at the END of the parameter list.
-// Supported param types: llvm_i1_ty (use 0/1 or true/false), llvm_i8_ty, llvm_i16_ty, llvm_i32_ty,
-//                        llvm_i64_ty.
-// Write any integer literal as the value.
-class DefaultIntArg<ArgIndex idx, int val> : IntrinsicProperty {
-  int ArgNo = idx.Value;  // 1-based internally; parser subtracts 1
-  int DefaultVal = val;
 }
 
 def IntrNoReturn : IntrinsicProperty;
--- a/llvm/include/llvm/IR/Intrinsics.td
+++ b/llvm/include/llvm/IR/Intrinsics.td
@@ -180,6 +180,16 @@ class ArgInfo<ArgIndex idx, list<ArgProperty> arg_properties> : IntrinsicPropert
   list<ArgProperty> Properties = arg_properties;
 }
 
+// DefaultIntArg - Attach an integer default value to an intrinsic argument.
+// The argument must be at the END of the parameter list.
+// Supported param types: llvm_i1_ty (use 0/1 or true/false), llvm_i8_ty, llvm_i16_ty, llvm_i32_ty,
+//                        llvm_i64_ty.
+// Write any integer literal as the value.
+class DefaultIntArg<ArgIndex idx, int val> : IntrinsicProperty {
+  int ArgNo = idx.Value;  // 1-based internally; parser subtracts 1
+  int DefaultVal = val;
+}
+
 def IntrNoReturn : IntrinsicProperty;
 
 // Applied by default.
--- a/llvm/include/llvm/IR/Intrinsics.td
+++ b/llvm/include/llvm/IR/Intrinsics.td
@@ -136,9 +136,23 @@ class Returned<ArgIndex idx> : IntrinsicProperty {
   int ArgNo = idx.Value;
 }
 
-// ImmArg - The specified argument must be an immediate.
-class ImmArg<ArgIndex idx> : IntrinsicProperty {
+// DefaultValue - A value carrier paired with ImmArg to declare a default for
+// a missing trailing argument. AutoUpgrade fills the default when an old
+// .bc / .ll file is loaded.
+class DefaultValue<int val> {
+  int Value = val;
+}
+
+// Sentinel — used as the default for ImmArg's optional DefaultValue parameter.
+def NoDefault : DefaultValue<0> {
+  let Value = ?;
+}
+
+// ImmArg - The specified argument must be an immediate. The optional
+// second template parameter specifies a default value for AutoUpgrade.
+class ImmArg<ArgIndex idx, DefaultValue val = NoDefault> : IntrinsicProperty {
   int ArgNo = idx.Value;
+  DefaultValue Default = val;
 }
 
 // ReadOnly - The specified argument pointer is not written to through the
--- a/llvm/include/llvm/IR/IntrinsicsNVVM.td
+++ b/llvm/include/llvm/IR/IntrinsicsNVVM.td
@@ -1302,6 +1302,37 @@ let TargetPrefix = "nvvm" in {
                             [IntrConvergent, IntrNoMem, IntrHasSideEffects]>;
 
   //
+  // Test intrinsic for default-arg feature: three required i32 args plus a
+  // fourth i32 with DefaultIntArg = 255.
+  //
+  def int_nvvm_test_add3 : NVVMBuiltin,
+      NVVMPureIntrinsic<[llvm_i32_ty],
+                        [llvm_i32_ty, llvm_i32_ty, llvm_i32_ty, llvm_i32_ty],
+                        [ImmArg<ArgIndex<3>>,
+                         DefaultIntArg<ArgIndex<3>, 255>]>;
+
+  //
+  // Test intrinsic for default-arg feature: four required i32 args plus two
+  // trailing args (i64, i32) with DefaultIntArg = 50 and 14.
+  //
+  def int_nvvm_test_add4 : NVVMBuiltin,
+      NVVMPureIntrinsic<[llvm_i32_ty],
+                        [llvm_i32_ty, llvm_i32_ty, llvm_i32_ty, llvm_i32_ty,
+                         llvm_i64_ty, llvm_i32_ty],
+                        [ImmArg<ArgIndex<4>>, ImmArg<ArgIndex<5>>,
+                         DefaultIntArg<ArgIndex<4>, 50>,
+                         DefaultIntArg<ArgIndex<5>, 14>]>;
+
+  //
+  // Dummy based on int_nvvm_nanosleep. Adds i1 flag with default = false.
+  //
+  def int_nvvm_nanosleep_dummy : NVVMBuiltin,
+      DefaultAttrsIntrinsic<[], [llvm_i32_ty, llvm_i1_ty],
+                            [IntrConvergent, IntrNoMem, IntrHasSideEffects,
+                             ImmArg<ArgIndex<1>>,
+                             DefaultIntArg<ArgIndex<1>, false>]>;
+
+  //
   // Performance Monitor Events (pm events) intrinsics
   //
   def int_nvvm_pm_event_mask : NVVMBuiltin,
@@ -2064,6 +2095,12 @@ def int_nvvm_cp_async_bulk_commit_group : Intrinsic<[]>;
 def int_nvvm_cp_async_bulk_wait_group :
     Intrinsic<[], [llvm_i32_ty], [ImmArg<ArgIndex<0>>]>;
 
+// Dummy based on cp_async_bulk_wait_group. Adds i1 flag with default = false.
+def int_nvvm_cp_async_bulk_wait_group_dummy :
+    Intrinsic<[], [llvm_i32_ty, llvm_i1_ty],
+              [ImmArg<ArgIndex<1>>,
+               DefaultIntArg<ArgIndex<1>, false>]>;
+
 def int_nvvm_cp_async_bulk_wait_group_read :
     Intrinsic<[], [llvm_i32_ty], [ImmArg<ArgIndex<0>>]>;
 
@@ -2197,6 +2234,12 @@ let IntrProperties = [IntrNoMem] in {
   def int_nvvm_move_double : DefaultAttrsIntrinsic<[llvm_double_ty], [llvm_double_ty]>;
   def int_nvvm_move_ptr : DefaultAttrsIntrinsic<[llvm_anyptr_ty], [llvm_anyptr_ty]>;
 }
+
+// Dummy based on int_nvvm_move_i32. Adds i64 second arg with default = 99.
+def int_nvvm_move_i32_dummy :
+    DefaultAttrsIntrinsic<[llvm_i32_ty], [llvm_i32_ty, llvm_i64_ty],
+                          [ImmArg<ArgIndex<1>>,
+                           DefaultIntArg<ArgIndex<1>, 99>]>;
 
 // For getting the handle from a texture or surface variable
 def int_nvvm_texsurf_handle
@@ -3179,6 +3222,12 @@ def int_nvvm_tcgen05_fence_before_thread_sync : Intrinsic<[], [],
   [IntrNoMem, IntrHasSideEffects]>;
 def int_nvvm_tcgen05_fence_after_thread_sync : Intrinsic<[], [],
   [IntrNoMem, IntrHasSideEffects]>;
+
+// Dummy based on tcgen05_fence. Two i32 args + i1 flag with default = false.
+def int_nvvm_tcgen05_fence_dummy :
+    Intrinsic<[], [llvm_i32_ty, llvm_i32_ty, llvm_i1_ty],
+              [IntrConvergent, ImmArg<ArgIndex<2>>,
+               DefaultIntArg<ArgIndex<2>, false>]>;
 
 // Tcgen05 cp intrinsics
 foreach cta_group = ["cg1", "cg2"] in {
--- a/llvm/include/llvm/IR/IntrinsicsNVVM.td
+++ b/llvm/include/llvm/IR/IntrinsicsNVVM.td
@@ -1303,34 +1303,31 @@ let TargetPrefix = "nvvm" in {
 
   //
   // Test intrinsic for default-arg feature: three required i32 args plus a
-  // fourth i32 with DefaultIntArg = 255.
+  // fourth i32 with DefaultVal = 255.
   //
-  def int_nvvm_test_add3 : NVVMBuiltin,
+  def int_nvvm_test_add3 :
       NVVMPureIntrinsic<[llvm_i32_ty],
                         [llvm_i32_ty, llvm_i32_ty, llvm_i32_ty, llvm_i32_ty],
-                        [ImmArg<ArgIndex<3>>,
-                         DefaultIntArg<ArgIndex<3>, 255>]>;
+                        [ImmArg<ArgIndex<3>, DefaultVal<255>>]>;
 
   //
   // Test intrinsic for default-arg feature: four required i32 args plus two
-  // trailing args (i64, i32) with DefaultIntArg = 50 and 14.
+  // trailing args (i64, i32) with DefaultVal = 50 and 14.
   //
-  def int_nvvm_test_add4 : NVVMBuiltin,
+  def int_nvvm_test_add4 :
       NVVMPureIntrinsic<[llvm_i32_ty],
                         [llvm_i32_ty, llvm_i32_ty, llvm_i32_ty, llvm_i32_ty,
                          llvm_i64_ty, llvm_i32_ty],
-                        [ImmArg<ArgIndex<4>>, ImmArg<ArgIndex<5>>,
-                         DefaultIntArg<ArgIndex<4>, 50>,
-                         DefaultIntArg<ArgIndex<5>, 14>]>;
+                        [ImmArg<ArgIndex<4>, DefaultVal<50>>,
+                         ImmArg<ArgIndex<5>, DefaultVal<14>>]>;
 
   //
   // Dummy based on int_nvvm_nanosleep. Adds i1 flag with default = false.
   //
-  def int_nvvm_nanosleep_dummy : NVVMBuiltin,
+  def int_nvvm_nanosleep_dummy :
       DefaultAttrsIntrinsic<[], [llvm_i32_ty, llvm_i1_ty],
                             [IntrConvergent, IntrNoMem, IntrHasSideEffects,
-                             ImmArg<ArgIndex<1>>,
-                             DefaultIntArg<ArgIndex<1>, false>]>;
+                             ImmArg<ArgIndex<1>, DefaultVal<false>>]>;
 
   //
   // Performance Monitor Events (pm events) intrinsics
@@ -2098,8 +2095,7 @@ def int_nvvm_cp_async_bulk_wait_group :
 // Dummy based on cp_async_bulk_wait_group. Adds i1 flag with default = false.
 def int_nvvm_cp_async_bulk_wait_group_dummy :
     Intrinsic<[], [llvm_i32_ty, llvm_i1_ty],
-              [ImmArg<ArgIndex<1>>,
-               DefaultIntArg<ArgIndex<1>, false>]>;
+              [ImmArg<ArgIndex<1>, DefaultVal<false>>]>;
 
 def int_nvvm_cp_async_bulk_wait_group_read :
     Intrinsic<[], [llvm_i32_ty], [ImmArg<ArgIndex<0>>]>;
@@ -2238,8 +2234,7 @@ let IntrProperties = [IntrNoMem] in {
 // Dummy based on int_nvvm_move_i32. Adds i64 second arg with default = 99.
 def int_nvvm_move_i32_dummy :
     DefaultAttrsIntrinsic<[llvm_i32_ty], [llvm_i32_ty, llvm_i64_ty],
-                          [ImmArg<ArgIndex<1>>,
-                           DefaultIntArg<ArgIndex<1>, 99>]>;
+                          [ImmArg<ArgIndex<1>, DefaultVal<99>>]>;
 
 // For getting the handle from a texture or surface variable
 def int_nvvm_texsurf_handle
@@ -3226,8 +3221,8 @@ def int_nvvm_tcgen05_fence_after_thread_sync : Intrinsic<[], [],
 // Dummy based on tcgen05_fence. Two i32 args + i1 flag with default = false.
 def int_nvvm_tcgen05_fence_dummy :
     Intrinsic<[], [llvm_i32_ty, llvm_i32_ty, llvm_i1_ty],
-              [IntrConvergent, ImmArg<ArgIndex<2>>,
-               DefaultIntArg<ArgIndex<2>, false>]>;
+              [IntrConvergent,
+               ImmArg<ArgIndex<2>, DefaultVal<false>>]>;
 
 // Tcgen05 cp intrinsics
 foreach cta_group = ["cg1", "cg2"] in {
--- a/llvm/lib/IR/AutoUpgrade.cpp
+++ b/llvm/lib/IR/AutoUpgrade.cpp
@@ -1278,7 +1278,7 @@ static bool convertIntrinsicValidType(StringRef Name,
   return false;
 }
 
-static bool upgradeDeclWithDefaultArgs(Function *F, Function *&NewFn) {
+static bool upgradeIntrinsicDeclWithDefaultArgs(Function *F, Function *&NewFn) {
   // Look up the intrinsic ID by full name, e.g. "llvm.nvvm.add3.i32"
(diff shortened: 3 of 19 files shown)


llvm/include/llvm/IR/Intrinsics.td
```diff
@@ -180,6 +180,16 @@ class ArgInfo<ArgIndex idx, list<ArgProperty> arg_properties> : IntrinsicPropert
   list<ArgProperty> Properties = arg_properties;
 }
 
+// DefaultIntArg - Attach an integer default value to an intrinsic argument.
+// The argument must be at the END of the parameter list.
+// Supported param types: llvm_i1_ty (use 0/1 or true/false), llvm_i8_ty, llvm_i16_ty, llvm_i32_ty,
+//                        llvm_i64_ty.
+// Write any integer literal as the value.
+class DefaultIntArg<ArgIndex idx, int val> : IntrinsicProperty {
```
mshockwave: In CodeGenIntrinsics.cpp you expect DefaultIntArg to be an immediate argument as well. If that's the case do we want to just merge this with `ImmArg`? Something like `ImmArg<ArgIndex<1>, DefaultVal<5>>`?

llvm/include/llvm/IR/Intrinsics.td
```diff
@@ -130,9 +130,24 @@ class Returned<ArgIndex idx> : IntrinsicProperty {
...
 }
 
-// ImmArg - The specified argument must be an immediate.
-class ImmArg<ArgIndex idx> : IntrinsicProperty {
+// DefaultVal - A value carrier paired with ImmArg to declare a default for
+// a missing trailing argument. AutoUpgrade fills the default when an old
+// .bc / .ll file is loaded.
+class DefaultVal<int val> {
+  int Value = val;
+  bit HasDefault = 1;
```
durga4github: we can move this above and keep it as the first field.

llvm/include/llvm/IR/IntrinsicsNVVM.td
```diff
@@ -2198,6 +2231,11 @@ let IntrProperties = [IntrNoMem] in {
   def int_nvvm_move_ptr : DefaultAttrsIntrinsic<[llvm_anyptr_ty], [llvm_anyptr_ty]>;
 }
 
+// Dummy based on int_nvvm_move_i32. Adds i64 second arg with default = 99.
+def int_nvvm_move_i32_dummy :
+    DefaultAttrsIntrinsic<[llvm_i32_ty], [llvm_i32_ty, llvm_i64_ty],
+                          [ImmArg<ArgIndex<1>, DefaultVal<99>>]>;
```
durga4github: ok, I see that all these changes are for examples. So, we should remove these when we un-draft the PR.

llvm/include/llvm/IR/IntrinsicsNVVM.td
```diff
@@ -3180,6 +3218,12 @@ def int_nvvm_tcgen05_fence_before_thread_sync : Intrinsic<[], [],
 def int_nvvm_tcgen05_fence_after_thread_sync : Intrinsic<[], [],
   [IntrNoMem, IntrHasSideEffects]>;
 
+// Dummy based on tcgen05_fence. Two i32 args + i1 flag with default = false.
+def int_nvvm_tcgen05_fence_dummy :
+    Intrinsic<[], [llvm_i32_ty, llvm_i32_ty, llvm_i1_ty],
+              [IntrConvergent,
+               ImmArg<ArgIndex<2>, DefaultVal<false>>]>;
```
durga4github: including this...

llvm/lib/IR/AutoUpgrade.cpp
```diff
@@ -1278,6 +1278,36 @@ static bool convertIntrinsicValidType(StringRef Name,
   return false;
 }
 
+static bool upgradeDeclWithDefaultArgs(Function *F, Function *&NewFn) {
```
durga4github: Should we say upgrade{Intr/Intrinsic}DeclWithDefaultArgs(..) ?

llvm/lib/IR/AutoUpgrade.cpp
```diff
@@ -1278,6 +1278,36 @@ static bool convertIntrinsicValidType(StringRef Name,
...
+  // Look up the intrinsic ID by full name, e.g. "llvm.nvvm.add3.i32"
+  Intrinsic::ID IID = Intrinsic::lookupIntrinsicID(F->getName());
+  if (IID == Intrinsic::not_intrinsic)
+    return false;
+
+  // Fast path: this intrinsic has no default args annotated in .td
+  if (!Intrinsic::hasDefaultArgs(IID))
+    return false;
+
+  if (Intrinsic::isOverloaded(IID))
```
durga4github, line 1322: I think a comment here will help us understand why this is a special case.

llvm/lib/IR/AutoUpgrade.cpp
```diff
@@ -4978,6 +5011,60 @@ static Value *upgradeConvertIntrinsicCall(StringRef Name, CallBase *CI,
   return nullptr;
 }
 
+static bool upgradeCallWithDefaultArgs(CallBase *CI, Function *NewFn,
```
durga4github: IntrCall..?

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
```diff
@@ -385,6 +385,23 @@ CodeGenIntrinsic::CodeGenIntrinsic(const Record *R,
...
+  unsigned NumParams = IS.ParamTys.size();
+  bool SeenDefault = false;
+  for (unsigned i = 0; i < NumParams; ++i) {
+    bool HasDefault =
+        (i < ParamDefaultValues.size() && ParamDefaultValues[i].has_value());
+    if (HasDefault) {
+      SeenDefault = true;
+    } else if (SeenDefault) {
+      PrintFatalError(TheDef->getLoc(),
+                      "DefaultVal gap at argument " + Twine(i) +
```
durga4github: s/gap/missing ?

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
```diff
@@ -581,3 +610,15 @@ void CodeGenIntrinsic::addPrettyPrintFunction(unsigned ArgIdx,
...
 }
+
+void CodeGenIntrinsic::addDefaultArgValue(unsigned ArgIdx, int64_t Value) {
+  if (ArgIdx >= ParamDefaultValues.size())
+    ParamDefaultValues.resize(ArgIdx + 1, std::nullopt);
+
+  if (ParamDefaultValues[ArgIdx].has_value())
+    PrintFatalError(TheDef->getLoc(), "Default value for argument " +
+                                          Twine(ArgIdx) +
+                                          " is already defined");
```
durga4github, line 657: wondering when do we hit this case..?

durga4github:
Also, please update the commit message to reflect the latest changes

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
varadk27: This fires if an intrinsic author accidentally declares two defaults on the same argument.
It's a defensive check against malformed .td input. Without it, the second declaration would silently overwrite the first, and the author would have no idea which value was actually used.

llvm/include/llvm/IR/Intrinsics.td
varadk27: Done

llvm/include/llvm/IR/IntrinsicsNVVM.td
varadk27: Okay

llvm/lib/IR/AutoUpgrade.cpp
varadk27: Updated in the latest revision.

llvm/lib/IR/AutoUpgrade.cpp
varadk27: Updated

llvm/lib/IR/AutoUpgrade.cpp
varadk27: Updated in the latest revision

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
varadk27: Updated

llvm/include/llvm/IR/Intrinsics.td
varadk27: Addressed in the latest revision. Merged `DefaultVal `into `ImmArg `as `ImmArg<ArgIndex<N>, DefaultVal<V>>`. Please take a look when you get a chance, thanks!

llvm/include/llvm/IR/Intrinsics.td
```diff
@@ -130,9 +130,24 @@ class Returned<ArgIndex idx> : IntrinsicProperty {
...
 }
 
-// ImmArg - The specified argument must be an immediate.
-class ImmArg<ArgIndex idx> : IntrinsicProperty {
+// DefaultVal - A value carrier paired with ImmArg to declare a default for
+// a missing trailing argument. AutoUpgrade fills the default when an old
+// .bc / .ll file is loaded.
+class DefaultVal<int val> {
+  bit HasDefault = 1;
+  int Value = val;
```
jurahul, line 143: note: TableGen supports ? for unset values, so we can use that and eliminate the HasDefault bit.

llvm/include/llvm/IR/Intrinsics.td
```diff
@@ -130,9 +130,24 @@ class Returned<ArgIndex idx> : IntrinsicProperty {
   int ArgNo = idx.Value;
 }
 
-// ImmArg - The specified argument must be an immediate.
-class ImmArg<ArgIndex idx> : IntrinsicProperty {
+// DefaultVal - A value carrier paired with ImmArg to declare a default for
```
jurahul: Can we spell the full name DefaultValue?

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
jurahul: nit: Rename "DefField" to use DefaultField. And use `getValueAsDef` to remove some checks here.

jurahul: nit: DefaultValue missing for argument

jurahul, line 408: We can do some additional validation of default values here. For example, if the ImmArg is for a non-integer type, issue an error. And if the ImmArg value does not fit in the declared type of the intrinsic, issue an error as well (for example, i8 param and DefaultValue = 1000).

jurahul: "Def" is a commonly used term in this code to denote a record def, so this will avoid ambiguity.

llvm/utils/TableGen/Basic/IntrinsicEmitter.cpp
jurahul: Isn't checking ParamDefaultValues.empty() sufficient? When will the table be non-empty and have all entries std::nullopt?

jurahul: same comment as above, just check ParamDefaultValues.empty(). Also name variable to HasAnyDefault.

jurahul: I think this code can be simplified to be more table driven as opposed a nested switch. One observation is that we know that only the last trailing set of immargs will have default values, so for an intrinsic, we just need to remember [FirstDefaultArg, NumDefaultArgs], All Arguments in the span [FirstDefaultArg, FirstDefaultArg + NumDefaultArgs) will have a default value.

So we can collect all the default args into a table/static array for all intrinsics with a default arg. Each entry is 64-bit. We prefix that with another 64-bit value that is packed [FirstDefaultArg, NumDefaultArgs] (2 32-bit values). We can then use a `SequenceToOffsetTable` to collect these entries for all intrinsics in a single table, and dump the table as well as a another per-intrinsic table of offset into this table.

The code for getDefaultArgValue() can then look like:

```
std::optional<int64_t> Intrinsic::getDefaultArgValue(ID IID, unsigned ArgIdx) {
   uint32_t Offset = DefaulArgValuesTableOffset[IID];
   uint64_t Header = DefalultArgValuesTable[Offset];
   uint32_t FirstDefaultArg = Header & UINT_MAX;
   uint32_t NumDefaultArg = (Header >> 32) & UINT_MAX;
   // Note: intrinsics with no default args will point to an entry with Header = 0. so they will
   // always return std::nullopt;
   if (ArgIndex < FirstDefaulArg || ArgIndex >= FirstDefaulArg + NumDefaultArg)
     return std::nullopt;
   // Following the header are the actual default values stored.
   return DefalultArgValuesTable[Offset + 1 + ArgIdx - FirstDefaultArg];
}
```

llvm/lib/IR/AutoUpgrade.cpp
```diff
@@ -4983,6 +5013,60 @@ static Value *upgradeConvertIntrinsicCall(StringRef Name, CallBase *CI,
...
+  if (OldArgCount >= NewArgCount)
+    return false;
+
+  // Start with the existing arguments from the old call.
+  SmallVector<Value *, 8> NewArgs(CI->args());
+
+  // Fill in each missing trailing argument from the table.
+  FunctionType *NewFT = NewFn->getFunctionType();
+  for (unsigned Idx = OldArgCount; Idx < NewArgCount; ++Idx) {
+    std::optional<int64_t> DefaultVal = Intrinsic::getDefaultArgValue(IID, Idx);
```
jurahul: Given that you always fill all trailing default values, should `getDefaultArgValue` function instead be something that just return all default values for a given intrinsic. Something like:

```
/// Returns the first default argument no and the set of default arg values. 
std::pair<unsigned, ArrayRef<int64_t>> Intrinsic::getAllDefaultArgValues(IID);
```

jurahul: For the ArrayRef returned, the size will be NumDefaultArgs and it will directly point to the static table that I mentioned earlier.

llvm/include/llvm/IR/Intrinsics.td
varadk27: Updated in the latest revision. Thanks!

llvm/include/llvm/IR/Intrinsics.td
varadk27: Renamed to DefaultValue in latest revision. Thanks!

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
varadk27: Renamed and switched to getValueAsDef in latest revision. Thanks!

varadk27: Fixed.

varadk27: Addressed in the latest revision. Thanks!

llvm/utils/TableGen/Basic/IntrinsicEmitter.cpp
varadk27: This lambda was removed by the table-driven refactor below ; no longer needed. Thanks!

varadk27: Refactored in latest revision to use `SequenceToOffsetTable `+ per-intrinsic offset table as suggested. Thanks!

llvm/lib/IR/AutoUpgrade.cpp
varadk27: Done. Replaced `getDefaultArgValue `with `getAllDefaultArgValues` returning `std::pair<unsigned, ArrayRef<int64_t>>`. Thanks!

llvm/include/llvm/IR/Intrinsics.td
```diff
@@ -130,9 +130,23 @@ class Returned<ArgIndex idx> : IntrinsicProperty {
...
+  int Value = val;
+}
+
+// Sentinel — used as the default for ImmArg's optional DefaultValue parameter.
+def NoDefault : DefaultValue<0> {
+  let Value = ?;
+}
+
+// ImmArg - The specified argument must be an immediate. The optional
+// second template parameter declares a default value for AutoUpgrade.
```
jurahul: nit: instead of "declares a default value" can this be "specifies a default value?

llvm/lib/IR/AutoUpgrade.cpp
jurahul: Q: can you clarify what's the issue with overloaded intrinsics? And do we check for this in TableGen side and error out (if an ImmArg for an overloaded intrinsic has default value)?

llvm/lib/IR/AutoUpgrade.cpp
```diff
@@ -1278,6 +1278,38 @@ static bool convertIntrinsicValidType(StringRef Name,
...
+  // Get the canonical full declaration for this intrinsic.
+  Function *FullDecl = Intrinsic::getOrInsertDeclaration(F->getParent(), IID);
+
+  // If the existing declaration already has all args, nothing to upgrade
+  if (F->arg_size() >= FullDecl->arg_size())
+    return false;
+
+  // Every missing trailing arg must fall within the default range
+  // [FirstDefault, FirstDefault + Defaults.size()).
+  for (unsigned Idx = F->arg_size(); Idx < FullDecl->arg_size(); ++Idx) {
```
jurahul: why is a loop needed here. I'd think we just want FirstMissingIndex (F->argSize()) >= FirstDefault?

llvm/lib/IR/AutoUpgrade.cpp
```diff
@@ -4983,6 +5013,55 @@ static Value *upgradeConvertIntrinsicCall(StringRef Name, CallBase *CI,
...
+  if (OldArgCount >= NewArgCount)
+    return false;
+
+  // Start with the existing arguments from the old call.
+  SmallVector<Value *, 8> NewArgs(CI->args());
+
+  // Fill in each missing trailing argument from the table.
+  FunctionType *NewFT = NewFn->getFunctionType();
+  for (unsigned Idx = OldArgCount; Idx < NewArgCount; ++Idx) {
+    if (Idx < FirstDefault || Idx >= FirstDefault + Defaults.size())
```
jurahul: should this be assert instead?

llvm/test/CodeGen/NVPTX/default-arg-explicit-value-preserved.ll
```diff
@@ -0,0 +1,13 @@
+; RUN: opt -S < %s | FileCheck %s
```
jurahul: Since there is no codegen here, can these tests be moved to a common place and not under CodeGen/NVPTX?

Also, since they are all related, you can just put them in a single file and use `split` to have it be divided into multiple unit tests. Maybe that helps the organization of these tests.

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
jurahul: nit: Default value missing (use regular English terminology and not specific class name). Also the " Defaults must form a contiguous trailing block ending at the last parameter" can probably be dropped. For C++, clang just issues "missing default argument on parameter 'y'" error message, we can follow the same style.

jurahul: This should be an assert maybe, as structurally it's impossible, or maybe just omit it, since ParamTyp is derived from a LLVMType list in the .td file.

jurahul: for witdth < 64, we don't know the signedness of the type, so should only accept values that are >= 0? For example, for i0, we will only accept values 0 -255, or 0x0 - 0xff.

jurahul, line 435: Add check for overloaded intrinsics not supporting default (for now, till it gets implemented)?

llvm/utils/TableGen/Basic/IntrinsicEmitter.cpp
jurahul: nit: Use IfDefEmitter

jurahul: nit: constexpr

jurahul: use a range for loop here and PerIntrinsic[i] can be replaced by push_back of std::move? PerIntrinsc.reserve() at the start.

jurahul: nit: use range for loop

jurahul: Can the table be emitted using int64_t?

jurahul: If its int64_t, we can avoid the reinperpret_cast. And just use static_cast<> for the header.

llvm/include/llvm/IR/Intrinsics.td
varadk27: Updated in the latest revision. Thanks!

llvm/lib/IR/AutoUpgrade.cpp
varadk27: Refactored to a single check in the latest revision. Thanks!

llvm/lib/IR/AutoUpgrade.cpp
varadk27: Updated to an assert in the latest revision. Thanks!

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
varadk27: Updated. Thanks!

varadk27: Dropped the check in the latest revision. Thanks!

varadk27: Refactored in the latest revision. Thanks!

varadk27: Added a TableGen-time error in the latest revision. Thanks!

llvm/utils/TableGen/Basic/IntrinsicEmitter.cpp
varadk27: Updated to use `IfDefEmitter`. Thanks!

varadk27: Updated. Thanks!

varadk27: Refactored in the latest revision. Thanks!

varadk27: Updated. Table is now int64_t. Thanks!

varadk27: Done. Removed the `reinterpret_cast`; the header now uses `static_cast`. Thanks!

varadk27: Refactored in the latest revision. Thanks!

llvm/lib/IR/AutoUpgrade.cpp
varadk27: For an overloaded intrinsic, filling default args means reconstructing the full declaration, which needs the overload types extracted from the mangled name. I originally did this with `Intrinsic::matchIntrinsicSignature` and it worked fine locally, but that API was made non-public in #195380, so there's no longer a clean public way to extract the overload types. 
I'd prefer to handle overloaded intrinsics in a follow-up PR (probably with a small dedicated helper) rather than duplicate that logic here.

llvm/test/CodeGen/NVPTX/default-arg-explicit-value-preserved.ll
varadk27: Clubbed all 8 into a single split-file test and moved it out of CodeGen/NVPTX. Thanks!

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
```diff
@@ -386,6 +386,17 @@ CodeGenIntrinsic::CodeGenIntrinsic(const Record *R,
   for (auto &Attrs : ArgumentAttributes)
     llvm::sort(Attrs);
 
+  // Default values are not yet supported for overloaded intrinsics.
+  // (overloaded support will come in a follow-up).
+  if (isOverloaded) {
+    for (const std::optional<int64_t> &DV : ParamDefaultValues) {
```
jurahul: nit: use llvm::any_of() in some form here?

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
```diff
@@ -397,38 +408,31 @@ CodeGenIntrinsic::CodeGenIntrinsic(const Record *R,
...
-                      "DefaultValue at argument " + Twine(i) +
+                      "default argument on parameter " + Twine(i) +
                           " requires an integer parameter type");
     }
     unsigned Width = VT->getValueAsInt("Size");
     int64_t Value = *ParamDefaultValues[i];
-    int64_t MaxUnsigned = (Width >= 64) ? INT64_MAX : ((1LL << Width) - 1);
-    int64_t MinSigned = (Width >= 64) ? INT64_MIN : -(1LL << (Width - 1));
-    if (Value < MinSigned || Value > MaxUnsigned) {
+    int64_t MaxValue =
```
jurahul: Since the value is now interpreted as unsigned int, would it make sense to store it as uint64_t in the table? That is ParamDefaultValues[i] will be a uint64_t. And then you can use `isUInt` in MathExtras.h to see if it fits in the given bitwidth.

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
varadk27: Updated to use `llvm::any_of`. Thanks!

llvm/utils/TableGen/Basic/CodeGenIntrinsics.cpp
varadk27: Switched the value storage and table to uint64_t and used isUIntN for the width check ; also lets the casts go away. Thanks!

jurahul:
Overall looks good to me. However, would like another reviewer to approver as well. @mshockwave and @nikic can you please review this PR?

llvm/test/Assembler/intrinsic-default-args-upgrade.ll
```diff
@@ -0,0 +1,108 @@
...
+; RUN: opt -S < %t/multiple-trailing.ll    | FileCheck %t/multiple-trailing.ll
+; RUN: opt -S < %t/partial-fill.ll         | FileCheck %t/partial-fill.ll
+; RUN: opt -S < %t/nanosleep.ll            | FileCheck %t/nanosleep.ll
+; RUN: opt -S < %t/cp-async-bulk.ll        | FileCheck %t/cp-async-bulk.ll
+; RUN: opt -S < %t/move-i32.ll             | FileCheck %t/move-i32.ll
+; RUN: opt -S < %t/tcgen05-fence.ll        | FileCheck %t/tcgen05-fence.ll
+
+;--- single-val.ll
+; A single missing trailing argument is filled with its default (255).
+declare i32 @llvm.nvvm.test.add3(i32, i32, i32)
```
mshockwave: we no longer need to declare intrinsics

llvm/utils/TableGen/Basic/IntrinsicEmitter.cpp
mshockwave: could we simply use empty `Sequence` instead of std::optional / nullopt?

mshockwave: I'm pretty sure the number of arguments is relatively small, so I think we could use SmallVector here. In fact, I think `PerIntrinsic` below could also be SmallVector.

mshockwave: perhaps `for (size_t j = 0U, N = Int.ParamDefaultValues.size(); j < N; ++j)`?

mshockwave: I don't really see the need of reserving here. The number of arguments is unlikely to be super big such that you want to avoid the performance pitfall of reallocation, which is one of the common reasons to allocate ahead of time via `reserve`.

mshockwave: nit: is the compiler smart enough so that you can write `ArrayRef(&DefaultArgValuesTable[Offset + 1], NumDefaults)`?

mshockwave: ditto `N = Int.ParamDefaultValues.size()`

llvm/test/Assembler/intrinsic-default-args-upgrade.ll
varadk27: Removed in latest revision. Thanks!

llvm/utils/TableGen/Basic/IntrinsicEmitter.cpp
varadk27: Updated to use an empty Sequence in latest revision. Thanks!

varadk27: Updated to SmallVector in latest revision. Thanks!

varadk27: Updated in latest revision. Thanks!

varadk27: Removed in latest revision. Thanks!

varadk27: Done. Switched to ArrayRef CTAD in latest revision. Thanks!

varadk27: Updated in latest revision. Thanks!

mshockwave: you could also add a fixed size here as the number of intrinsic arguments rarely exceeds, say 8: `SmallVector<uint64_t, 8>`.

mshockwave (approved):
LGTM

jurahul:
Thanks @mshockwave for the review and approval. Approving as well.

jurahul (approved):

mshockwave:
Please avoid force push unless necessary: https://llvm.org/docs/GitHub.html#rebasing-pull-requests-and-force-pushes

PR #198560, opened 2026-05-19
[VPlan] Use inferred type for VPReplicateRecipe void check (NFC)

madhur13490:


llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -3242,6 +3242,11 @@ class LLVM_ABI_FOR_TEST VPReplicateRecipe : public VPRecipeWithIRFlags,
 
   bool isPredicated() const { return IsPredicated; }
 
+  /// Returns true if the underlying instruction has a non-void result type.
+  bool definesResult() const {
+    return !getUnderlyingInstr()->getType()->isVoidTy();
+  }
```
fhahn: If we add this, it should be framed in terms of the recipe (and its type itself), not the underlying instruction

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -3319,7 +3319,7 @@ void VPReplicateRecipe::execute(VPTransformState &State) {
          "VPReplicateRecipes must be unrolled before ::execute");
   auto *Instr = getUnderlyingInstr();
   Instruction *Cloned = Instr->clone();
-  if (!Instr->getType()->isVoidTy()) {
+  if (definesResult()) {
```
fhahn: Here it might be cleaner to instead pull up

```
    Type *ResultTy = State.TypeAnalysis.inferScalarType(this);
```
and check is that is void

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -3627,7 +3627,7 @@ void VPReplicateRecipe::printRecipe(raw_ostream &O, const Twine &Indent,
                                     VPSlotTracker &SlotTracker) const {
   O << Indent << (IsSingleScalar ? "CLONE " : "REPLICATE ");
 
-  if (!getUnderlyingInstr()->getType()->isVoidTy()) {
+  if (definesResult()) {
```
fhahn: a more generic alternative removing the direct use of the underlying instruction would be to check the inferred type

llvm/lib/Transforms/Vectorize/VPlan.h
madhur13490: done.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
madhur13490: Done. does the new changes look good?

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
madhur13490: Sure, done.

fhahn (approved):
LGTM, thanks

PR #198561, opened 2026-05-19
[RegisterCoalescer] Don't remat trivial defs without a size benefit

guy-david:
isAsCheapAsAMove doesn't imply "one machine instruction". AArch64 marks multi-instruction pseudos cheap when their fused latency matches a real move (MOVaddr = adrp+add, MOVi64imm = MOVZ+MOVK). The trivial remat duplicates such defs at every COPY use.

Compile time tracker for this PR + https://github.com/llvm/llvm-project/pull/121904:
https://llvm-compile-time-tracker.com/compare.php?from=c9c4a3b099bac4de30cd72af1b83d88c333d2605&to=4b358b29e7d4a1ba1de9cb87c528cc2347c57a2f&stat=instructions:u


davemgreen (commented):
I'm not sure if this is better or not, I think it might be fine (or just randomly different). It is difficult for Thumb2 to have an accurate approximation of size at all points in the pipeline as we do not always know what will later become a 16bit thumb instruction and which will not. Some of these don't seem like they should be larger.

PR #196558, opened 2026-05-08
[Flang][Driver] Add per-target search path for modules

Meinersbur:
Adds the version- and target-specific path

    ../lib/clang/<version>/finclude/flang/<target>

to the intrinsic module search path in addition to

    ../finclude/flang

with the former taking precedence if a module file should exist in both. The version/target-specific path is added by the driver by passing `-fintrinsic-modules-path` to the `-fc1` invocation. This is consistent with gfortran and the usual pattern that the driver resolves paths into the resource path, not the frontend.

This PR adds nothing into that directory, which will be done in #171515.

Extracted out of #171515 as requested by https://github.com/llvm/llvm-project/pull/171515#pullrequestreview-4179212219

PR #198565, opened 2026-05-19
[Flang][bbc] Add support for -fintrinsic-module-path

Meinersbur:
Add support for -fintrinsic-module-path to bbc. With #196558 the default search path in the resource directory is determined by the driver, not by the frontend (-fc1 and bbc). For tests of either frontend the tests will need to append the search path themselves. Adding support for `-fintrinsic-module-path` will allow using the same additional argument for flang_fc1 and bbc tests.

Also rewriting the awkward tests that expect standard modules to fail only because they do not contain a checksum. Maintaining this way of testing became burdonsome because there are multiple ways of how default modules are resolved. Fixing one breaks another.

Extracted out of #171515 to make that patch a bit smaller.

### DIFF
--- a/flang/test/Driver/intrinsic-module-path.f90
+++ b/flang/test/Driver/intrinsic-module-path.f90
@@ -1,6 +1,6 @@
 ! Ensure argument -fintrinsic-modules-path works as expected.
 ! WITHOUT the option, the default location for the module is checked and no error generated.
-! With the option WRONG and GIVEN, find the module in the first first
+! With the option WRONG and GIVEN, find the module in the first
 ! -fintrinsic-modules-path that contains a matching file.
 
 !-----------------------------------------
--- a/flang/test/Driver/intrinsic-module-path.f90
+++ b/flang/test/Driver/intrinsic-module-path.f90
@@ -14,7 +14,7 @@
 ! RUN:     %flang_fc1 -fsyntax-only -fintrinsic-modules-path %S/Inputs/module-dir-one -fintrinsic-modules-path %S/Inputs/module-dir     %s 2>&1 | FileCheck %s --allow-empty --check-prefix=GIVEN
 
 !-----------------------------------------
-! BURNSIDE BRIDGE_COMPILER (bbc)
+! BURNSIDE BRIDGE COMPILER (bbc)
 !-----------------------------------------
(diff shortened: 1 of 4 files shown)


flang/test/Driver/intrinsic-module-path.f90
```diff
@@ -1,23 +1,37 @@
 ! Ensure argument -fintrinsic-modules-path works as expected.
 ! WITHOUT the option, the default location for the module is checked and no error generated.
-! With the option GIVEN, the module with the same name is PREPENDED, and considered over the
-! default one, causing an error.
+! With the option WRONG and GIVEN, find the module in the first first
```
tarunprabhu: Would it be better to change the label `GIVEN` to `MISSING`? If I understand this correctly, the tests are checking that there is no message about a missing module in those cases.

```suggestion
! With the option WRONG and GIVEN, find the module in the first
```

flang/test/Driver/intrinsic-module-path.f90
tarunprabhu, line 1: There is an large amount of extra whitespace in several lines in the tests. Please remove them.

tarunprabhu (changes requested):
Thanks for pulling this out of the bigger PR.

flang/test/Driver/intrinsic-module-path.f90
Meinersbur: This is for alignment so you can see what the difference on each line is.

With: 
Image

Without:
Image

Do you still want it removed?

tarunprabhu: Yes, please. That works best if your editor window is sufficiently wide, which is not always the case. Even with wide screens, some developers may use large font sizes. I needed, you could get some degree of alignment by splitting the the `RUN` line over multiple lines. When adding new tests, one wouldn't have to do as much work to align things with spaces.

```
! RUN: not %flang_fc1 %s \
! RUN:     -fsyntax-only 2>&1 | FileCheck %s --check-prefix=WITHOUT
!
! RUN: not %flang_fc1 -fintrinsic-modules-path %S/Inputs/module-dir %s \
! RUN:     -fsyntax-only 2>&1 | FileCheck %s --check-prefix=WRONG
```

Meinersbur: done

tarunprabhu (approved):
Thanks for the changes. LGTM.

PR #186845, opened 2026-03-16
[BOLT] Fix a crash for calling `pwrite`

chenshanzhi:
Since `raw_pwrite_stream::pwrite` does not support extending the stream, each seek-and-write operation in `rewriteFile` either extends the stream or requires restoring the stream position after seeking backwards.

Fixes #185051

PR #198569, opened 2026-05-19
[BOLT] Fix pwrite assertion failure via a new safePWrite wrapper

Jinjie-Huang:
Background: Currently, BOLT seems to implicitly assume that the .dynsym section is located at a low offset within the binary, calling pwrite() directly to update it.

Issue: In scenarios where the binary has been modified by tools like patchelf, sections like .dynsym may be moved to a high offset area. This can lead to a violation of the Offset + Size <= Pos assertion in pwrite(). A typical scenario is when the previous eh_frame_header update moves the stream cursor (pos) back to a low offset([code](https://github.com/llvm/llvm-project/blob/llvmorg-23-init/bolt/lib/Rewrite/RewriteInstance.cpp#L6387)).

Fix: This patch resolves the pwrite assertion failure via a new safePWrite wrapper, which introduces a defensive check that verifies and conditionally adjusts the stream position. A corresponding test case has also been added.

### DIFF
--- a/bolt/include/bolt/Core/BinarySection.h
+++ b/bolt/include/bolt/Core/BinarySection.h
@@ -478,7 +478,7 @@ public:
 
   /// Flush all pending relocations to patch original contents of sections
   /// that were not emitted via MCStreamer.
-  void flushPendingRelocations(raw_pwrite_stream &OS,
+  void flushPendingRelocations(raw_fd_ostream &OS,
                                SymbolResolverFuncTy Resolver);
 
   /// Change contents of the section. Unless the section has a valid SectionID,
--- a/bolt/include/bolt/Rewrite/MachORewriteInstance.h
+++ b/bolt/include/bolt/Rewrite/MachORewriteInstance.h
@@ -62,8 +62,7 @@ class MachORewriteInstance {
   void runOptimizationPasses();
   void emitAndLink();
 
-  void writeInstrumentationSection(StringRef SectionName,
-                                   raw_pwrite_stream &OS);
+  void writeInstrumentationSection(StringRef SectionName, raw_fd_ostream &OS);
   void rewriteFile();
 
 public:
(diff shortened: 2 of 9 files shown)


aaupov (commented):
@maksfb I've seen this issue in several scenarios: this with dynsym, flushing pending relocs in sections past bolt_reserved. Rather than chasing these issues one by one, do you think we can handle it centrally?

maksfb:
@Jinjie-Huang thanks for investigating the issue. Since we have seen more failures from `pwrite()`, it makes sense to create a wrapper in BOLT around it that extends the stream/file when needed.

maksfb:
Related PR: https://github.com/llvm/llvm-project/pull/186845

Jinjie-Huang:
Thanks for providing more context! That makes sense. Introducing a wrapper instead of repeating the defensive seek logic at multiple call sites is definitely a much cleaner and more robust design.

I am thinking we could introduce a helper function, say "pwriteWithStreamExtension", like this:
```
void safePWrite(uint64_t Offset, char *Src, Size) {
  auto &OS = Out->os();
  uint64_t SavedPos = OS.tell();
  uint64_t RequiredPos = Offset + Size;

  if (SavedPos < RequiredPos) {
    // some warning maybe..
    OS.seek(RequiredPos);
  }
    

  OS.pwrite(Src, Size, Offset);

  if (SavedPos < RequiredPos)
    OS.seek(SavedPos);
}
```

Then we could probably replace all `pwrite()` usages across BOLT with this new wrapper. If this makes sense, I'd be happy to update the PR and include the corresponding test cases (though it might be difficult to add a test for every single replacement).

Jinjie-Huang:
Since most `pwrite` usages in BOLT come from `raw_fd_ostream` (which inherits from `raw_pwrite_stream`). Therefore, it seems quite convenient to introduce a `pwrite` wrapper to unify the `pwrite` calls across BOLT.

Jinjie-Huang:
Gentle ping.

rafaelauler (approved):
Thanks! Looks good

PR #198571, opened 2026-05-19
[Coroutines] Expand materialization tests (NFC)

IgWod:
Expand checks so they check that the operation has been rematerialized, and computed values are not being stored.

Assisted-by: Claude Code


ChuanqiXu9 (approved):
LGTM. Thanks.

PR #197315, opened 2026-05-12
[Github] Hashpin base container in CI Tooling containerfile

tstellar:
https://github.com/llvm/llvm-project/security/code-scanning/1492

PR #198572, opened 2026-05-19
[Docs] Reccomend Container Pinning

boomanaiden154:
Add some info to CI Best Practices about pinning container images to a specific image SHA, which we agreed was a best practice in #197315 (and maybe somewhere else, but I cannot find anything).

This updates the best practices but does not currently attempt to actually fix all the cases where we are using unpinned container images.


tstellar (approved):
LGTM.

PR #198573, opened 2026-05-19
[lldb-dap][windows] fix TestDAP_restart_console tests

charles-zablit:
a614cd391a40 introduced a regression. It assumed that the launcher's `ResumeThread` was sufficient to start the debuggee.

That's incorrect: on Windows the inferior is held by two independent mechanisms after attach:
- pending Win32 debug events from the attach. The debuggee is created with `CREATE_SUSPENDED`. This requires a `ContinueDebugEvent` to release.
- the OS suspend count (handled by the launcher's `ResumeThread`), which are regular breakpoints.

The fix is to run `Continue()` in async mode on Windows by resetting scope_sync_mode first.

This is a follow up to https://github.com/llvm/llvm-project/pull/198764.

### DIFF
--- a/lldb/tools/lldb-dap/Handler/RequestHandler.cpp
+++ b/lldb/tools/lldb-dap/Handler/RequestHandler.cpp
@@ -139,21 +139,20 @@ RunInTerminal(DAP &dap, const protocol::LaunchRequestArguments &arguments) {
   std::future<lldb::SBError> did_attach_message_success =
       comm_channel.NotifyDidAttach();
 
-// We just attached to the runInTerminal launcher, which was waiting to be
-// attached. We now resume it, so it can receive the didAttach notification
-// and then perform the exec. Upon continuing, the debugger will stop the
-// process right in the middle of the exec. To the user, what we are doing is
-// transparent, as they will only be able to see the process since the exec,
-// completely unaware of the preparatory work.
-//
-// On Windows, the debuggee itself is waiting to be attached to. There is no
-// need to continue.
-#ifndef _WIN32
-  dap.target.GetProcess().Continue();
-#endif
-
-  // Return the debugger to its prior async state.
+#ifdef _WIN32
+  // On Windows the target was created with CREATE_SUSPENDED. The launcher
+  // will call ResumeThread after reading didAttach, but we still must process
+  // the initial debug events from the attach for the target to actually run.
+  // Use async Continue() so we don't block while the suspend count is still 1.
   scope_sync_mode.reset();
+  dap.target.GetProcess().Continue();
+#else
+  // On Linux the launcher and target are the same process. Continue() resumes
+  // the launcher so it can read didAttach and execvp into the target, stopping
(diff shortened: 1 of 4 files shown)


lldb/tools/lldb-dap/Handler/RequestHandler.cpp
```diff
@@ -139,21 +139,20 @@ RunInTerminal(DAP &dap, const protocol::LaunchRequestArguments &arguments) {
...
-#ifndef _WIN32
+#ifdef _WIN32
+  // On Windows the target was created with CREATE_SUSPENDED. The launcher
+  // will call ResumeThread after reading didAttach, but we still must process
+  // the initial debug events from the attach for the target to actually run.
+  // Use async Continue() so we don't block while the suspend count is still 1.
+  scope_sync_mode.reset();
+  dap.target.GetProcess().Continue();
+#else
+  // On Linux the launcher and target are the same process. Continue() resumes
```
Nerixyz, line 150:
```suggestion
  // On other platforms the launcher and target are the same process. Continue() resumes
```

Nerixyz (approved):
Tested locally - at least on Windows they also seem to pass in debug mode.

DrSergei:
@charles-zablit Thank you for working on this! In my company we have very similar implementation of integrated terminal on Windows platform (fifo + create suspended). I noticed that there are 2 types of machine. For the first one everything works ok. But for the second one the first stop occurs on system code with message `Exception 0x80000003 encountered at address ...`. I think it is very similar to behavior which you described here. I cherry-picked your fix to our code and it works for second type machines, but on first type machines I get error `process exited during launch or attach`. I am going to test your main branch with your patch today, maybe this situation is not related to your changes

DrSergei:
I double checked on this [commit](https://github.com/llvm/llvm-project/commit/a975b7cd3538cfba015c7a49e2e86be58aa546c3) from main + this patch and the behavior is the same. On one machine it works fine, but on the other one I see `process exited during launch or attach` when try to start debugging in integrated terminal

Nerixyz:
Are the machines running the same version of Windows and the same build of LLDB?

DrSergei:
> Are the machines running the same version of Windows and the same build of LLDB?

For test I used LLDB build from [main](https://github.com/llvm/llvm-project/commit/a975b7cd3538cfba015c7a49e2e86be58aa546c3) plus this patch.

1 type PC: Windows 11 Pro 22631.6199
Without this patch LLDB works fine
With this patch I see `process exited during launch or attach` when try to start debugging in integrated terminal
2 type PC: Windows 11 Pro 26100.8037
Without this patch lldb stops in system code with message `Exception 0x80000003 encountered at address ...`
With this patch LLDB works fine

charles-zablit:
Thanks for looking into this.

This is not the right fix then. This might be due to a race condition. I cannot reproduce this at desk, but I can reproduce it in the CI PR tests for lldb on Windows (which are not enabled yet). That should help us track down the issue.

PR #198577, opened 2026-05-19
[ARM] Use GOT indirection for weak symbols in PIC mode 

dongjianqiang2:
On ARM ELF PIC, weak dso_local symbols were incorrectly accessed via PC-relative constant pool entries (e.g. .long xxx-(.LPC+8)). The assembler eagerly resolves this expression when both the symbol and label are in the same section, preventing the linker from overriding a weak definition with a strong one from another object file.

Fix #183916 by treating weak symbols as indirect (requiring GOT) on ELF PIC, matching GCC's behavior. This affects:
- isGVIndirectSymbol (ARMTargetMachine) — Darwin, GlobalISel, stack prot.
- isGVInGOT (ARMSubtarget)
- LowerGlobalAddressELF (SelectionDAG path)
- ARMLowerPICELF (FastISel path)

Assisted-by: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/llvm/lib/Target/ARM/ARMFastISel.cpp
+++ b/llvm/lib/Target/ARM/ARMFastISel.cpp
@@ -3033,7 +3033,8 @@ bool ARMFastISel::tryToFoldLoadIntoMI(MachineInstr *MI, unsigned OpNo,
 }
 
 Register ARMFastISel::ARMLowerPICELF(const GlobalValue *GV, MVT VT) {
-  bool UseGOT_PREL = !GV->isDSOLocal();
+  // Weak symbols need GOT indirection even when hidden/DSO-local.
+  bool UseGOT_PREL = !GV->isDSOLocal() || GV->isWeakForLinker();
   LLVMContext *Context = &MF->getFunction().getContext();
   unsigned ARMPCLabelIndex = AFI->createPICLabelUId();
   unsigned PCAdj = Subtarget->isThumb() ? 4 : 8;
--- a/llvm/lib/Target/ARM/ARMISelLowering.cpp
+++ b/llvm/lib/Target/ARM/ARMISelLowering.cpp
@@ -3643,10 +3643,8 @@ SDValue ARMTargetLowering::LowerGlobalAddressELF(SDValue Op,
   const GlobalValue *GV = cast<GlobalAddressSDNode>(Op)->getGlobal();
   bool IsRO = isReadOnly(GV);
 
-  // promoteToConstantPool only if not generating XO text section.
-  // Weak symbols may be overridden at link time, so don't inline them.
-  if (GV->isDSOLocal() && !Subtarget->genExecuteOnly() &&
-      !GV->isWeakForLinker())
+  // promoteToConstantPool only if not generating XO text section
+  if (GV->isDSOLocal() && !Subtarget->genExecuteOnly())
     if (SDValue V = promoteToConstantPool(this, GV, DAG, PtrVT, dl))
       return V;
 
@@ -3654,7 +3652,7 @@ SDValue ARMTargetLowering::LowerGlobalAddressELF(SDValue Op,
     // Weak symbols need GOT indirection even when hidden/DSO-local.
     // The assembler eagerly resolves PC-relative expressions when the
     // symbol and reference are in the same section, which prevents the
-    // linker from overriding a weak definition with a strong one.
+    // linker from overriding a weak definition with a non-weak one.
     bool UseGOT = !GV->isDSOLocal() || GV->isWeakForLinker();
     SDValue G = DAG.getTargetGlobalAddress(GV, dl, PtrVT, 0,
                                            UseGOT ? ARMII::MO_GOT : 0);
--- a/llvm/lib/Target/ARM/ARMISelLowering.cpp
+++ b/llvm/lib/Target/ARM/ARMISelLowering.cpp
@@ -3643,16 +3643,23 @@ SDValue ARMTargetLowering::LowerGlobalAddressELF(SDValue Op,
   const GlobalValue *GV = cast<GlobalAddressSDNode>(Op)->getGlobal();
   bool IsRO = isReadOnly(GV);
 
-  // promoteToConstantPool only if not generating XO text section
-  if (GV->isDSOLocal() && !Subtarget->genExecuteOnly())
+  // promoteToConstantPool only if not generating XO text section.
+  // Weak symbols may be overridden at link time, so don't inline them.
+  if (GV->isDSOLocal() && !Subtarget->genExecuteOnly() &&
+      !GV->isWeakForLinker())
     if (SDValue V = promoteToConstantPool(this, GV, DAG, PtrVT, dl))
       return V;
 
   if (isPositionIndependent()) {
-    SDValue G = DAG.getTargetGlobalAddress(
-        GV, dl, PtrVT, 0, GV->isDSOLocal() ? 0 : ARMII::MO_GOT);
+    // Weak symbols need GOT indirection even when hidden/DSO-local.
+    // The assembler eagerly resolves PC-relative expressions when the
+    // symbol and reference are in the same section, which prevents the
+    // linker from overriding a weak definition with a strong one.
+    bool UseGOT = !GV->isDSOLocal() || GV->isWeakForLinker();
+    SDValue G = DAG.getTargetGlobalAddress(GV, dl, PtrVT, 0,
+                                           UseGOT ? ARMII::MO_GOT : 0);
     SDValue Result = DAG.getNode(ARMISD::WrapperPIC, dl, PtrVT, G);
-    if (!GV->isDSOLocal())
+    if (UseGOT)
       Result =
           DAG.getLoad(PtrVT, dl, DAG.getEntryNode(), Result,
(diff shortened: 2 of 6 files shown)


smithp35:
Thanks for posting, I'm short of time this week may be next week before I can take a look.

llvm/test/CodeGen/ARM/elf-preemption.ll
```diff
@@ -137,17 +138,6 @@ define dso_local ptr @dsolocal_func() nounwind {
 ; STATIC-NEXT:    movw r0, :lower16:dsolocal_func
 ; STATIC-NEXT:    movt r0, :upper16:dsolocal_func
 ; STATIC-NEXT:    bx lr
-;
```
smithp35, line 140: Is there any reason that this has been removed. The output looks similar to me when I check the output of the patch?

llvm/test/CodeGen/ARM/elf-preemption.ll
```diff
@@ -180,13 +171,6 @@ define dso_local void @call_dsolocal_func() nounwind {
 ; STATIC-NEXT:    push {r11, lr}
 ; STATIC-NEXT:    bl dsolocal_func
 ; STATIC-NEXT:    pop {r11, pc}
-;
```
smithp35, line 183: Is there any reason that this has been removed. The output looks similar to me when I check the output of the patch?

llvm/test/CodeGen/ARM/weak-hidden-pic.ll
```diff
@@ -0,0 +1,19 @@
...
+; expressions like .long xxx-(.LPC+8) when both are in the same section,
+; which prevents the linker from overriding the weak definition with
+; a strong one from another object file.
+
+define weak dso_local hidden void @weak_hidden_func() {
+  ret void
+}
+
+; CHECK-LABEL: weak_hidden_func_addr:
+; ARM:       .long weak_hidden_func(GOT_PREL)
```
smithp35: Do you need separate check-prefixes for ARM and THUMB. I think you can just use CHECK:, possibly even CHECK-NEXT:

llvm/lib/Target/ARM/ARMISelLowering.cpp
```diff
@@ -3643,16 +3643,23 @@ SDValue ARMTargetLowering::LowerGlobalAddressELF(SDValue Op,
...
     if (SDValue V = promoteToConstantPool(this, GV, DAG, PtrVT, dl))
       return V;
 
   if (isPositionIndependent()) {
-    SDValue G = DAG.getTargetGlobalAddress(
-        GV, dl, PtrVT, 0, GV->isDSOLocal() ? 0 : ARMII::MO_GOT);
+    // Weak symbols need GOT indirection even when hidden/DSO-local.
+    // The assembler eagerly resolves PC-relative expressions when the
+    // symbol and reference are in the same section, which prevents the
+    // linker from overriding a weak definition with a strong one.
```
smithp35: Could we use non-weak rather than strong? ELF doesn't define strong (I know it is used a lot informally).

llvm/lib/Target/ARM/ARMISelLowering.cpp
```diff
@@ -3643,16 +3643,23 @@ SDValue ARMTargetLowering::LowerGlobalAddressELF(SDValue Op,
   const GlobalValue *GV = cast<GlobalAddressSDNode>(Op)->getGlobal();
   bool IsRO = isReadOnly(GV);
 
-  // promoteToConstantPool only if not generating XO text section
-  if (GV->isDSOLocal() && !Subtarget->genExecuteOnly())
+  // promoteToConstantPool only if not generating XO text section.
```
smithp35: Can you explain this part?

If I remove it then all the tests pass. If you have a use-case in mind would you be able to write a test for it.

It also seems like it could be triggering for PIC and non-PIC.

smithp35 (commented):
Apologies for the delay. I'm struggling to keep up with reviews at the moment. I've made an initial pass. Will need to refresh my memory on some of the other parts.

llvm/test/CodeGen/ARM/elf-preemption.ll
dongjianqiang2: Restored — the function is dso_local but not weak, so PIC behavior (PC-relative without GOT) is unchanged and should still be tested.

llvm/test/CodeGen/ARM/elf-preemption.ll
dongjianqiang2: Restored

llvm/test/CodeGen/ARM/weak-hidden-pic.ll
dongjianqiang2: Simplified to just CHECK: since both ISAs produce identical output

llvm/lib/Target/ARM/ARMISelLowering.cpp
dongjianqiang2: Changed to "non-weak" in ARMISelLowering.cpp, ARMTargetMachine.h.

llvm/lib/Target/ARM/ARMISelLowering.cpp
dongjianqiang2: Removed — it's dead code: promoteToConstantPool only handles GlobalVariables with hasLocalLinkage(), and weak symbols never have local linkage.

dongjianqiang2:
@smithp35 Thanks for review!, Please take another look when you have a chance.

llvm/lib/Target/ARM/ARMFastISel.cpp
```diff
@@ -3033,7 +3033,8 @@ bool ARMFastISel::tryToFoldLoadIntoMI(MachineInstr *MI, unsigned OpNo,
 }
 
 Register ARMFastISel::ARMLowerPICELF(const GlobalValue *GV, MVT VT) {
-  bool UseGOT_PREL = !GV->isDSOLocal();
+  // Weak symbols need GOT indirection even when hidden/DSO-local.
+  bool UseGOT_PREL = !GV->isDSOLocal() || GV->isWeakForLinker();
```
smithp35, line 3037: I don't think we've got any existing test coverage for fastisel.

I think it could be worth adding a case to one of the tests to check the fast-isel code path. I expect adding a case to weak-hidden-pic.ll would be sufficient, although do make sure the test would fail without this patch.

smithp35 (commented):
Just one comment about a test for the fastisel code path. Otherwise this is looking good to me.

There's a possible optimisation, if we know that the weak definition is defined in its own ELF section, then the assembler can't eagerly resolve the expression. However I doubt that there's enough instances of taking the address of weak hidden functions to make that worthwhile.

llvm/lib/Target/ARM/ARMFastISel.cpp
dongjianqiang2: Good catch, thanks. Added a -O0 -fast-isel-abort=2 RUN line to weak-hidden-pic.ll — it will fail without the fix.

smithp35 (approved):
Thanks for the updates. This looks good to me. Please leave a couple of days for any comments from other reviewers before merging.

dongjianqiang2:
> Thanks for the updates. This looks good to me. Please leave a couple of days for any comments from other reviewers before merging.

Thanks for the reviews. No further comments received, so I'm merging this now.

frobtech:
> Just one comment about a test for the fastisel code path. Otherwise this is looking good to me.
> 
> There's a possible optimisation, if we know that the weak definition is defined in its own ELF section, then the assembler can't eagerly resolve the expression. However I doubt that there's enough instances of taking the address of weak hidden functions to make that worthwhile.

In fact, I think this optimization opportunity is quite common.  The target symbol is not necessarily a function, it might be a data object.  A weak definition of a data object is common e.g. in COMDAT cases like a `static` variable inside an `inline` function.  Also, under `-ffunction-sections` then there will in fact be no cases at all of a reference to a symbol that's known to be weak but also known to be defined (because it's in the same TU) that would ever be in the same section as the target of the reference and so subject to the issue of the assembler eliding the reloc.

@petrhosek @Prabhuk

frobtech:
I've managed to coax the assembler into producing the desired reloc for the same-section case:
```
.thumb
.text
	.weak def
def:	 nop
refuser:
	ldr r0, ref
	nop
.balign 4
ref:
	.long refuser+4 - .
	.reloc ref, R_ARM_REL32, def
```
So perhaps we can make the compiler do that for these cases.  The manual plumbing of the reloc / addend is the same as what `def - (refuser+4)` in the operand produces in the other-section case.  So there's no real need to decide which case it is, it can just always emit the manual plumbing.

dongjianqiang2:
> I've managed to coax the assembler into producing the desired reloc for the same-section case:
> 
> ```
> .thumb
> .text
> 	.weak def
> def:	 nop
> refuser:
> 	ldr r0, ref
> 	nop
> .balign 4
> ref:
> 	.long refuser+4 - .
> 	.reloc ref, R_ARM_REL32, def
> ```
> 
> So perhaps we can make the compiler do that for these cases. The manual plumbing of the reloc / addend is the same as what `def - (refuser+4)` in the operand produces in the other-section case. So there's no real need to decide which case it is, it can just always emit the manual plumbing.

Agreed, will update.

PR #198578, opened 2026-05-19
[llvm-objcopy] Strip header from DXContainer's ILDB part during `--dump-section`

kuilpd:
This strips the header from ILDB part of DXContainer during `llvm-objcopy --dump-section`, so that the output contains bitcode only.


llvm/test/tools/llvm-objcopy/DXContainer/dump-section.yaml
```diff
@@ -30,10 +35,229 @@ Header:
   Version:
     Major:           1
     Minor:           0
-  FileSize:        1872
-  PartCount:       2
-  PartOffsets:     [ 40, 1780 ]
+  FileSize:        3616
+  PartCount:       3
+  PartOffsets:     [ 44, 1784, 3524 ]
```
jh7370: It would be good if yaml2obj supported computing these values based on the other properties of the file, similar to how we don't need to list the section header offsets or many other fields in an ELF object.

llvm/test/tools/llvm-objcopy/DXContainer/dump-section.yaml
```diff
@@ -30,10 +35,229 @@ Header:
...
+    Size:            1732
+    Program:
+      MajorVersion:    6
+      MinorVersion:    0
+      ShaderKind:      5
+      Size:            433
+      DXILMajorVersion: 1
+      DXILMinorVersion: 0
+      DXILSize:        1708
+      DXIL:            [ 0x42, 0x43, 0xC0, 0xDE, 0x21, 0xC, 0x0, 0x0, 0xA8,
```
jh7370: Is "real" data actually needed here, or would it be sufficient for some minimal fake data instead, to keep the opaqueness of this minimal?

jh7370 (commented):
It's worth a DXContainer expert reviewing this too, but from past memory of what has happened, this looks generally reasonable.

llvm/test/tools/llvm-objcopy/DXContainer/dump-section.yaml
kuilpd: The test checks that stripping the header results in a code that is valid, using the disassembler. I think this is just about the smallest we can generate. The debug info is actually irrelevant here, I'll remove the comment about it.

jh7370: But do we need to show that the code is valid after stripping the header, or could we just show that it is unchanged, which allows for simpler output? E.g. the binary could be: `<header>foo` and then you'd just check that the contents of the output file is `foo`. In ELF `--dump-section`, for example, we don't try to show that the section dumped is a valid section, becuase it's not significant to the testing.

kuilpd: Alright, makes sense. I removed the real code and replaced it with 4 bytes that I check now.

llvm/test/tools/llvm-objcopy/DXContainer/dump-section.yaml
llvm-beanz: I'm pretty sure it does support computing those. We have other tests that omit the `FileSize` and `PartOffsets` fields, they're just allowed to make it easy to generate invalid files.

jh7370: In that case, we shouldn't have them here (or in copy-headers.yaml), since this isn't a yaml2obj test.

llvm/test/tools/llvm-objcopy/DXContainer/dump-section.yaml
```diff
@@ -1,12 +1,18 @@
...
+# BITCODE: 42 43 12 34
 
-## Verify that when dumping the DXIL part we get a valid bitcode file.
-# BITCODE: define void @main()
 ## Verify the size of the bitcode data.
-# DXIL-SIZE: 1708
+# RUN: wc -c %t.bc | FileCheck %s --check-prefix=DXIL-SIZE
+# DXIL-SIZE: 4
+# RUN: wc -c %t0.bc | FileCheck %s --check-prefix=ILDB-SIZE
+# ILDB-SIZE: 4
```
jh7370: If you add something to the FileCheck lines for the payload check (e.g. `--implicit-check-not={{.}}`), you can avoid this entirely, since there's no need to validate the size when you've checked the exact file contents.

kuilpd: > If you add something to the FileCheck lines for the payload check (e.g. `--implicit-check-not={{.}}`), you can avoid this entirely, since there's no need to validate the size when you've checked the exact file contents.

I'd have to `--implicit-check-not` the entire contents of the stripped header, I guess? I check for 4 bytes of bitcode that I expect, but in the case there are other stray bytes there, I feel like it's more readable and definitive to check that we got the size we were expecting.

jh7370: Actually, would add `{{^}}` and `{{$}}` either side of the od check pattern be enough? This has the advantage of making it a positive check, which is probably more readable than the negative check. `--implicit-check-not={{.}}` is semantically equivalent, I believe, but would produce output saying something along the lines of "found XX unexpectedly", which is probably less definitive. I don't know - perhaps worth trying either and deliberately feeding it the wrong input.

As `od` should print all the bytes, any stray bytes would be rendered in the FileCheck output, which should be pretty clear as to the problem in either case?

kuilpd: Added the anchors, it worked, tried with the wrong input as well. I like this solution more.

bogner (approved):

PR #198581, opened 2026-05-19
[mlir][acc] Allow non-resolvable callee in ACCImplicitRoutine

razvanlupusoru:
Some dialect call operations do not necessarily require or prevent for a call to occur to non-defined symbol (only func.call enforces this). Thus instead of making the condition an assert simply skip it just like other cases where call target cannot be found.

This PR also adds IR testing for ACCImplicitRoutine with the final test case being related to the issue this PR is solving.


clementval (approved):
LGTM

PR #198582, opened 2026-05-19
[alpha.webkit.UncountedLocalVarsChecker] Detect a raw pointer/reference in a decomposition declaration.

rniwa:


clang/lib/StaticAnalyzer/Checkers/WebKit/RawPtrRefLocalVarsChecker.cpp
```diff
@@ -289,68 +289,82 @@ class RawPtrRefLocalVarsChecker
...
+    if (auto *DD = dyn_cast<DecompositionDecl>(V)) {
+      for (auto *BD : DD->bindings()) {
+        auto *Binding = BD->getBinding();
+        if (!Binding)
+          continue;
+        std::optional<bool> IsUncountedPtr = isUnsafePtr(Binding->getType());
+        if (!IsUncountedPtr || !*IsUncountedPtr)
+          continue;
+        if (isPtrOriginSafe(V, Binding))
+          return;
```
t-rasmud: If we return here, then the second binding won't be checked. Is this by design?
What's the expected behavior for, say `auto [a, b] = pair<RefCountable*, RefCountable*> { param, provide() };`?

rniwa: oh, that should be continue.

rniwa: Actually, this `isPtrOriginSafe` was a no-op so just removed it.

t-rasmud: Thanks! Can we also add a test case where the second binding is unsafe?

t-rasmud (approved):
Thanks for fixing the bug. Left a comment asking for a test case to cover the case where the second binding is unsafe. Otherwise, LGTM!

clang/lib/StaticAnalyzer/Checkers/WebKit/RawPtrRefLocalVarsChecker.cpp
rniwa: I added that already 🙂

ojhunt (approved):
LGTM - hurrah for the "ignore whitespace" option in GH reviews

PR #198583, opened 2026-05-19
[mlir][Func][EmitC] Bail-out to avoid errors from MemRef array conversions

ioghiban:
Update FuncToEmitC to bail-out before creating invalid EmitC ops for unsupported cases.

FuncToEmitC now rejects functions, calls, and returns whose converted result type is `emitc.array`, instead of relying on later `emitc.func`, `emitc.call`, or `emitc.return` verifier failures.

This does not add support for returning memrefs from functions. It only makes the existing limitation explicit at the conversion boundary.

## Tests

Added negative tests for the standalone conversion pass. This pass marks their source ops illegal, so when a pattern bails-out the pass reports a legalization failure. This is the expected behavior and documents the unsupported cases directly.

`convert-to-emitc` is more permissive because it allows partial conversion and does not mark the same source ops illegal, so it can leave unsupported ops unconverted without reporting the same failures.

Assisted-by: Codex (refine description). I reviewed all text before submission.


banach-space (commented):
This is a very nice documentation of the limitations of the EmitC conversions, thank you! This patc will make reviewing your future changes (where you relax some of this constraints) much easier.

It's a bit late, so I will take a closer look later, but all in all LG.

mlir/lib/Conversion/FuncToEmitC/FuncToEmitC.cpp
```diff
@@ -60,6 +60,18 @@ class CallOpConversion final : public OpConversionPattern<func::CallOp> {
...
+    if (callOp.getNumResults() == 1) {
+      Type resultType =
+          getTypeConverter()->convertType(callOp.getResult(0).getType());
+      if (!resultType)
+        return rewriter.notifyMatchFailure(callOp,
+                                           "result type conversion failed");
+      if (isa<emitc::ArrayType>(resultType))
+        return rewriter.notifyMatchFailure(
+            callOp,
+            "converting functions returning arrays is not supported ATM");
```
banach-space:
```suggestion
            "converting function calls returning arrays is not supported ATM");
```

mlir/lib/Conversion/FuncToEmitC/FuncToEmitC.cpp
```diff
@@ -149,6 +165,11 @@ class ReturnOpConversion final : public OpConversionPattern<func::ReturnOp> {
     if (returnOp.getNumOperands() > 1)
       return rewriter.notifyMatchFailure(
           returnOp, "only zero or one operand is supported");
+    if (returnOp.getNumOperands() == 1 &&
+        isa<emitc::ArrayType>(adaptor.getOperands()[0].getType()))
+      return rewriter.notifyMatchFailure(
+          returnOp,
+          "converting functions returning arrays is not supported ATM");
```
banach-space: This is a conversion for returnOp rather than funcOp ;-)
```suggestion
          "converting returnOp is only supported for non-array values ATM");
```

mlir/test/Dialect/EmitC/func/ops.mlir
```diff
@@ -0,0 +1,139 @@
...
+// CHECK:           return %[[SRC]] : memref<1x1xi32, strided<[2, 1]>>
+// CHECK:         }
+
+func.func private @negative_unranked(%arg0: memref<*xi32>) -> memref<*xi32> {
+  return %arg0 : memref<*xi32>
+}
+// CHECK-LABEL:   func.func private @negative_unranked(
+// CHECK-SAME:      %[[SRC:.*]]: memref<*xi32>) -> memref<*xi32>
+// CHECK:           return %[[SRC]] : memref<*xi32>
+// CHECK:         }
```
banach-space: [nit] No need to check the closing brace. Here and in other places.

mlir/test/Dialect/EmitC/func/ops.mlir
```diff
@@ -0,0 +1,139 @@
...
+// Negative tests (must NOT rewrite)
+//===----------------------------------------------------------------------===//
+
+func.func private @negative_rank0_alloc() -> memref<i32> {
+  %alloc = memref.alloc() : memref<i32>
+  return %alloc : memref<i32>
+}
+// CHECK-LABEL:   func.func private @negative_rank0_alloc() -> memref<i32>
+// CHECK:           %[[ALLOC:.*]] = memref.alloc() : memref<i32>
+// CHECK:           return %[[ALLOC]] : memref<i32>
```
banach-space: This would be sufficient to check what you want to check.
```suggestion
// CHECK:           %[[ALLOC:.*]] = memref.alloc
// CHECK:           return %[[ALLOC]]
```

banach-space (approved):
> Update the Func-to-EmitC conversion to reject unsupported function, call, and return cases before constructing invalid EmitC ops.

Note that you are also updating ArithToEmitC ;-)

> Added negative coverage under mlir/test/Dialect/EmitC/func rather than the narrower Conversion/FuncToEmitC tests because these cases exercise the composed -convert-to-emitc pipeline across multiple dialects:

To me it would make more sense to test via e.g. `-convert-func-to-emitc` and put the test in e.g. "mlir/test/Conversion/FuncToEmitC/unsupported.mlir" - this PR is basically helping documented what's not supported ATM.

These comments are minor though, approving as is. Thank you for preparing this 🙏🏻

ioghiban:
The failing Linux x64 tests seem unrelated to my changes as they exercise openmp targets.

ioghiban:
> > Added negative coverage under mlir/test/Dialect/EmitC/func rather than the narrower Conversion/FuncToEmitC tests because these cases exercise the composed -convert-to-emitc pipeline across multiple dialects:
> 
> To me it would make more sense to test via e.g. `-convert-func-to-emitc` and put the test in e.g. "mlir/test/Conversion/FuncToEmitC/unsupported.mlir" - this PR is basically helping documented what's not supported ATM.
> 
Tests are failing for `convert-func-to-emitc` (similarly for `convert-memref-...` and `convert-arith-...`, as the conversion bails out and `func.func`, `func.call`, and `func.return` are marked illegal. Therefore, errors are thrown. The purpose of this patch is to highlight conversion limitations by introducing a bail-out mechanism. I agree,`convert-to-emitc` may fail gracefully, without errors, but it's too general and does not specify the exact limitations - it does not mark any ops illegal and accepts partial conversions. Will update with error checking for `convert-func-to-emitc` and `convert-arith-to-emitc`.

mlir/test/Conversion/FuncToEmitC/func-to-emitc-failed.mlir
banach-space: [nit] Here and in other places, you can drop "negative" from the test function name - all tests in this file are negative (they check for failures in conversion). Instead, you can add some extra info, e.g.:
```suggestion
func.func private @return_val_is_rank0_alloc() -> memref<i32> {
```

banach-space: [nit] It wasn't immediately obvious to me what was unique about this one (it was the fact that there are two elements in the return memref).
```suggestion
func.func private @return_value_is_rank_1_two_elemes(%arg0: memref<2xi64>) -> memref<2xi64> {
```

banach-space (approved):
Nice, thank you for refactoring the tests two use `-convert-{dialect}-to-emitc` rather than the `-convert-to-emitc` mammoth :) I've left some nits, but those are optional and feel free to ignore.

Thanks!

joker-eph:
> Nice, thank you for refactoring the tests two use -convert-{dialect}-to-emitc rather than the -convert-to-emitc mammoth :) 

Mmmm why? 

Taking the direction for LLVM, I've been trying instead to killl these redundant path to align on the single `convert-to-llvm` path and ensure we don't have subtle differences between these.

joker-eph:
The description looks like AI-assisted but I don't see an "Assisted-by: <agent>" in the description, is this an oversight or no AI was used here? ([LLVM AI policy](https://llvm.org/docs/AIToolPolicy.html) FYI)

ioghiban:
> The description looks like AI-assisted but I don't see an "Assisted-by: " in the description, is this an oversight or no AI was used here? ([LLVM AI policy](https://llvm.org/docs/AIToolPolicy.html) FYI)

@joker-eph you're right, I did use AI to tweak the description, but not the code here. Let me add the comment.

banach-space:
> > Nice, thank you for refactoring the tests two use -convert-{dialect}-to-emitc rather than the -convert-to-emitc mammoth :)
> 
> Mmmm why?

With `-convert-{dialect}-to-emitc`, we know exactly what we are testing. With, `-convert-to-emitc`, you pull in many other things along the way and you may end-up testing something else that's not relevant (e.g. `-convert-emitc-to-emitc` mixed with `-convert-arith-to-emitc` instead of `-convert-emitc-to-emitc` alone).

> Taking the direction for LLVM, I've been trying instead to killl these redundant path to align on the single `convert-to-llvm` path and ensure we don't have subtle differences between these.

It's not clear to me what path EmitC prefers to follow , but given that this PR is _extending pre-existing test files_  (as opposed to creating new ones), it felt safe to assume that this is the expected direction. CC @aniragil + @marbre

banach-space (approved):
@joker-eph , I assume that your comments were non-blocking and that it's ok to land this. Please let us know if it's otherwise, thanks!

mlir/lib/Conversion/FuncToEmitC/FuncToEmitC.cpp
aniragil:
```suggestion
            "function calls returning arrays are not supported");
```

aniragil:
```suggestion
            "functions returning arrays are not supported");
```

aniragil:
```suggestion
          "returning arrays is not supported");
```

mlir/lib/Conversion/ArithToEmitC/ArithToEmitC.cpp
aniragil:
```suggestion
          arithConst, "memref constants are not supported");
```

mlir/test/Conversion/FuncToEmitC/func-to-emitc-failed.mlir
aniragil: why test both illegal conversions combined?

aniragil (changes requested):
This seems like two unrelated changes. Any reason not to separate them into two PRs?

ioghiban:
> This seems like two unrelated changes. Any reason not to separate them into two PRs?

@aniragil The reason was reducing the noise level - it's a very small PR. Happy to split it if that's your preference, I don't mind.

aniragil:
> > This seems like two unrelated changes. Any reason not to separate them into two PRs?
> 
> @aniragil The reason was reducing the noise level - it's a very small PR. Happy to split it if that's your preference, I don't mind.

Then please do. Besides clarifying that the changes are unrelated and passing CI separately, it would also simplify reverting one without the other if necessary.

ioghiban:
> Then please do. Besides clarifying that the changes are unrelated and passing CI separately, it would also simplify reverting one without the other if necessary.

@aniragil , please check out [#201565](https://github.com/llvm/llvm-project/pull/201565)

aniragil (approved):
LGTM, thanks @ioghiban !

banach-space (commented):
LGTM, thank you for splitting https://github.com/llvm/llvm-project/pull/198583 🙏🏻

PR #198585, opened 2026-05-19
[CUF] Fix `do concurrent` PFT navigation in `cuf.kernel` lowering

ergawy:
When lowering a cuf kernel directive with an explicit loop depth (e.g. `do(2)`), the code navigated `nestedLoops-1` levels deeper in the PFT expecting nested regular do-loops. For a `do concurrent` construct with multiple index variables (e.g. `do concurrent(i=1:k,j=1:n)`), the PFT represents the entire construct as a single node, so the extra navigation stepped off into a plain statement with no nested evaluations and triggered an assertion in `getNestedEvaluations()`.

Fix by skipping the depth navigation loop for `do concurrent`, which is always a single flat PFT construct regardless of how many index variables it declares.

PR stack:
- https://github.com/llvm/llvm-project/pull/198584 
- https://github.com/llvm/llvm-project/pull/198585 ◀️

### DIFF
--- a/flang/lib/Lower/Bridge.cpp
+++ b/flang/lib/Lower/Bridge.cpp
@@ -4058,8 +4058,9 @@ private:
 
     if (crtEval->lowerAsStructured()) {
       crtEval = &crtEval->getFirstNestedEvaluation();
-      for (int64_t i = 1; i < nestedLoops; i++)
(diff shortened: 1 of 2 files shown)


clementval (approved):
LGTM

PR #198590, opened 2026-05-19
[TBAA] Correctly specify path to base type for aggregate assignment

Dudeldu:
When assigning to aggregates, clang code generation does not properly specify the "type-path" but only assigns the underlying type. This is unnecessarily pessimistic in cases where this underlying type is e.g., an `omnipotent char*`:
An example might look like this:
```c
struct my_byte {
  char b;
};

struct vec {
  struct my_byte *data;
  int sz;
};

void add(struct vec *v, char x) {
  v->data[v->sz] = (struct my_byte){x};
  v->sz += 1;
}

void build(struct vec *v) {
  add(v, (char)42);
  add(v, (char)42);
  add(v, (char)42);
  add(v, (char)42);
}
```
As the `data`  field and the `vec` pointer can not alias due to the TBAA here, we can combine the increment of `sz` and the writes. GCC applies this optimization (refer https://godbolt.org/z/Eq69ExvsW). With this fix, we end up with similar code generation to GCC.


clang/lib/CodeGen/CodeGenTBAA.cpp
```diff
@@ -512,7 +510,9 @@ CodeGenTBAA::getTBAAStructInfo(QualType QTy) {
     return N;
 
   SmallVector<llvm::MDBuilder::TBAAStructField, 4> Fields;
-  if (CollectFields(0, QTy, Fields, TypeHasMayAlias(QTy)))
+  const bool MayAlias = TypeHasMayAlias(QTy);
+  llvm::MDNode *TBAAType = MayAlias ? getChar() : getTypeInfo(QTy);
```
antoniofrighetto, line 514: Was going to suggest to use getBaseTypeInfo(), but it seems like getTypeInfo() already falls back to getValidBaseTypeInfo() for valid base types...

antoniofrighetto (commented):
> As the `data` field and the `vec` pointer can not alias due to the TBAA here, we can combine the increment of `sz` and the writes.

I believe you meant to say whether the store to `v->data[v->sz]` could alias with the (re)loads of `v->sz` or `v->data` (the `data` field lives inside `vec`, they are not really competing accesses), which cannot alias (per C strict aliasing rule), but as long as omnipotent char is emitted, we can't really say that.

It looks like there are other two AArch64 tests to update.

Dudeldu:
I updated the remaining AArch64 tests; the remaining test failure seems unrelated (in particular because it didn't show up in the first run)

antoniofrighetto:
This looks reasonable to me. @ojhunt and @rjmccall may also want to take a look at this.

ojhunt (commented):
I don't see anything that looks overtly wrong to me, but I don't think that's valuable - I've never worked on anything TBAA related - I think @nikic or @efriedma-quic might be worth adding as reviewers?

ojhunt:
> I don't see anything that looks overtly wrong to me, but I don't think that's valuable - I've never worked on anything TBAA related - I think @nikic or @efriedma-quic might be worth adding as reviewers?

aaaand I just saw they're already listed.

PR #198591, opened 2026-05-19
[flang][OpenMP] Remove ompFlagsRequireMark from symbol resolution

kparzysz:
The `ompFlagsRequireMark` set was there to make sure that we put the flags from it on symbols even when no new symbols needed to be created.

Instead of doing that, we can just put the flag on the symbol every time. There is no harm in having these flags, it's just extra information.


luporl (approved):
LGTM

PR #197027, opened 2026-05-11
[libc] Implement pthread_getthreadid_np and pthread_getunique_np.

vonosmas:
This is an alternative to https://github.com/llvm/llvm-project/pull/195202 which uses @petrhosek 's suggestion
to use an existing `pthread_getthreadid_np()` instead.

We define two functions:
* pthread_getthreadid_np - to return a thread-id for the current thread. **NOTE:** We're using the IBM variant of this function (https://www.ibm.com/docs/en/i/7.6.0?topic=ssw_ibm_i_76/apis/users_22.html) instead of FreeBSD one (https://man.freebsd.org/cgi/man.cgi?pthread_getthreadid_np), so that we can return a more-opaque integral type (in our case - `uintptr_t` rather than `int`).
* pthread_getunique_np - it's even less standard one. We're also using IBM variant here (https://www.ibm.com/docs/en/i/7.6.0?topic=ssw_ibm_i_76/apis/users_23.html). This one is needed because we need to have the capability to convert `pthread_t` into a numeric ID.

PR #195202, opened 2026-05-01
[libc] Add an extension macro to get numerical thread id

vonosmas:
Adds LLVM libc extension macro: `__PTHREAD_GET_ID` (for pthreads) and `__THRD_GET_ID` (for C11 threads) that extracts integral value representation of "thread ID" from opaque `pthread_t` / `thrd_t` standard-required types.

The idea is to use this extension in libc++ code to provide llvm-libc-specific defines for `__libcpp_thread_t` internal type, which needs to support `operator==`, `operator<`, as well as zero assignment.

These are implemented as macros - making a middle ground between:
  * not providing any API at all (then libc++ code would need to call `reinterpret_cast<uintptr_t>((t).__attrib)` itself, depending on private llvm-libc specific definition (as it does for zOS version of pthread_t)
  * creating an actual function like `__pthread_get_thread_id(pthread_t)` that would give better type safety and diagnostics, but would result in more boilerplate code and lack of inlining opportunities
I'd be happy to switch to the latter if preferred.

Anyway, let me know what you think!

FWIW, this issue seems to be the last remaining blocker preventing us from _compiling_ libc++ against llvm-libc headers with pthread support enabled, now that @SchrodingerZhu has implemented `pthread_cond_` variants. It would likely not be enough to pass the test suite yet.

PR #195139, opened 2026-04-30
[libc][libcxx] Use raw pointer thread handles

SchrodingerZhu:
This PR fixes the threading support related to building libc++ with libc. LLVM-libc also utilizes pointer based thread IDs. The patch contains two parts of changes:

- On libc part, expose `void *` as thread type directly; because libc++ assumes that thread ids are
  - comparable
  - formattable
  - zero initializable
- On libc++ part, always use pointer format if `__LLVM_LIBC__` is defined.

For more context, please refer to https://discourse.llvm.org/t/building-libcxx-with-llvms-libc.

Assisted-by: Codex with gpt-5.5 high fast

PR #198595, opened 2026-05-19
[libc++] Add support for thread-id handling for llvm-libc.

vonosmas:
This change adds support for properly defining and obtaining `__libcpp_thread_id` when llvm-libc is used. It defines the integral thread-id (which satisfies necessary restrictions of having total order, being hashable and formattable) as `pthread_id_np_t` type and uses `pthread_getthreadid_np` and `pthread_getunique_np` functions to obtain it (added in 
https://github.com/llvm/llvm-project/pull/197027, following the discussions in https://github.com/llvm/llvm-project/pull/195139 and https://github.com/llvm/llvm-project/pull/195202).

We also let `_LIBCPP_NULL_THREAD` macro use a more portable `PTHREAD_NULL` (defined in the latest POSIX) when this macro is available, so that it would work as expected for opaque `pthread_t` implementations, where default constructor might not necessarily zero-initialize all the members.

This is the last remaining change to allow building libc++ against llvm-libc with threads enabled (test-suite results TBD).

### DIFF
--- a/libcxx/include/__thread/support/pthread.h
+++ b/libcxx/include/__thread/support/pthread.h
@@ -144,7 +144,9 @@ inline _LIBCPP_HIDE_FROM_ABI int __libcpp_execute_once(__libcpp_exec_once_flag* 
 //
 // Thread id
 //
-#if defined(__MVS__)
+#if _LIBCPP_LIBC_LLVM_LIBC
+typedef pthread_id_np_t __libcpp_thread_id;
+#elif defined(__MVS__)
 typedef unsigned long long __libcpp_thread_id;
 #else
 typedef pthread_t __libcpp_thread_id;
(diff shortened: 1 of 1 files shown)


libcxx/include/__thread/support/pthread.h
```diff
@@ -163,11 +165,18 @@ inline _LIBCPP_HIDE_FROM_ABI bool __libcpp_thread_id_less(__libcpp_thread_id __t
 //
 // Thread
 //
-#define _LIBCPP_NULL_THREAD ((__libcpp_thread_t()))
+#if defined(PTHREAD_NULL)
+#  define _LIBCPP_NULL_THREAD PTHREAD_NULL
```
philnik777, line 169: Is this part of POSIX? I can't find anything about it.

petrhosek: It's a more recent addition, see https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/pthread.h.html.

philnik777 (approved):
LGTM as a temporary change to get the libc thread support going. It would be great if someone could try to refactor the threading API as a follow-up to be similar to localization though.

vonosmas:
> LGTM as a temporary change to get the libc thread support going. It would be great if someone could try to refactor the threading API as a follow-up to be similar to localization though.

Thank you! Would you like to land (or have someone pick up and land) the patch for printing thread-ids as integers (i.e. in non-hex format) first? Otherwise, I planned to look at the existing gnarly logic which Petr pointed at as a candidate for including in this refactoring (https://github.com/vonosmas/llvm-project/blob/main/libcxx/include/__config#L356-L363), but need to understand the reasoning behind the current choices first.

PR #198596, opened 2026-05-19
[mlir][ROCDL][AMDGPU] Add result arguments to buffer atomics

krzysz00:
Buffer atomic operations were failing LLVM validation because they weren't declared as having a result (the old value). This commit updates those operations to fix the error. (Note: if you don't need the result, you just don't use it, and the compiler backend emits an atomic that doesn't return the old value.)


Meinersbur (approved):
Typos in summary: orror

LGTM

PR #198598, opened 2026-05-19
[CodeGen] Always print 64 bit hash value in MachineBlockHashInfoPrinterPass

e-kud:
Hash must be fixed-size. 0x prefix is counted against the width in format_hex. Increasing from 16 to 18.


vitalybuka (approved):
Thank you!

boomanaiden154 (approved):
I guess the behavior of `format_hex` here is a bit of a footgun, but there's probably other cases where different behavior is a footgun.

LGTM.

e-kud:
> I guess the behavior of `format_hex` here is a bit of a footgun, but there's probably other cases where different behavior is a footgun.
> 
> LGTM.

Yeah, it took a minute to understand why 6 means only 4 hex digits
```
/// format_hex - Output \p N as a fixed width hexadecimal. If number will not
/// fit in width, full number is still printed.  Examples:
///   OS << format_hex(255, 4)              => 0xff
///   OS << format_hex(255, 4, true)        => 0xFF
///   OS << format_hex(255, 6)              => 0x00ff
///   OS << format_hex(255, 2)              => 0xff
```

PR #198599, opened 2026-05-19
[LifetimeSafety] Expand diagnostic list that enables analysis

NeKon69:
Now, when any lifetime safety related diagnostic is not ignored, we run the analysis.

No tests were added since this does not add new functionality.


usx95 (approved):
Thanks.

Xazax-hun (approved):

PR #198600, opened 2026-05-19
[Driver][MSVC] Use LLD if DWARF is requested

Nerixyz:
Clang can emit DWARF on Windows if requested with `-gdwarf` (or `-gdwarf-N`). The generated object files will then have the `.debug_*` sections. These section names are all larger than eight bytes. This creates an issue in the final executable. PE/COFF executables don't allow section names longer than eight characters. From the [docs](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers):

> **Name**: An 8-byte, null-padded UTF-8 encoded string. If the string is  exactly 8 characters long, there is no terminating null. For longer names, this field contains a slash (/) that is followed by an ASCII representation of a decimal number that is an offset into the string  table. Executable images do not use a string table and do not support section names longer than 8 characters.

MSVC's linker follows this and will truncate section names in executables. So a `.debug_info` would become `.debug_i`. LLD on the other hand does allow these long section names (with a warning). This makes DWARF work on MinGW.

The MSVC toolchain will use the MSVC linker by default (unless `CLANG_DEFAULT_LINKER` specifies a different one). This will truncate the section names. So if one would want to use DWARF, they'd also need to link with LLD.

This PR changes the default linker to LLD if DWARF is requested. Otherwise, the old default will be used.

### DIFF
--- a/clang/test/Driver/msvc-link.c
+++ b/clang/test/Driver/msvc-link.c
@@ -51,3 +51,21 @@
 // RUN: %clang -c -marm64x  --target=arm64ec-pc-windows-msvc -fuse-ld=link -### %s 2>&1 | \
 // RUN:        FileCheck --check-prefix=HYBRID-WARN %s
 // HYBRID-WARN: warning: argument unused during compilation: '-marm64x' [-Wunused-command-line-argument]
+
+// RUN: %clang --target=i686-pc-windows-msvc -g -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LINK %s
+// RUN: %clang --target=i686-pc-windows-msvc -g -fuse-ld=link -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LINK %s
+// DEBUG-LINK: link.exe"
+// DEBUG-LINK-SAME: "-debug"
+// DEBUG-LINK-NOT: lld
+
+// RUN: %clang --target=i686-pc-windows-msvc -g -fuse-ld=lld -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -g -fuse-ld=lld-link -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf-2 -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf-3 -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf-4 -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf-5 -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf-6 -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// DEBUG-LLD: lld-link"
+// DEBUG-LLD-SAME: "-debug"
+// DEBUG-LLD-NOT: link.exe
--- a/clang/lib/Driver/ToolChains/MSVC.cpp
+++ b/clang/lib/Driver/ToolChains/MSVC.cpp
@@ -279,12 +279,26 @@ void visualstudio::Linker::ConstructJob(Compilation &C, const JobAction &JA,
     AddRunTimeLibs(TC, TC.getDriver(), CmdArgs, Args);
   }
 
-  StringRef Linker = Args.getLastArgValue(options::OPT_fuse_ld_EQ,
-                                          TC.getDriver().getPreferredLinker());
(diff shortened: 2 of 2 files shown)


clang/test/Driver/msvc-link.c
```diff
@@ -51,3 +51,21 @@
 // RUN: %clang -c -marm64x  --target=arm64ec-pc-windows-msvc -fuse-ld=link -### %s 2>&1 | \
 // RUN:        FileCheck --check-prefix=HYBRID-WARN %s
 // HYBRID-WARN: warning: argument unused during compilation: '-marm64x' [-Wunused-command-line-argument]
+
+// RUN: %clang --target=i686-pc-windows-msvc -g -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LINK %s
```
compnerd, line 55: Is this actually valid? I thought that `-fuse-ld=` requires a value

Nerixyz: I think it's valid, there are several tests using `-fuse-ld=` without a value: https://github.com/search?q=repo%3Allvm%2Fllvm-project+%2F-fuse-ld%3D%5Cs%2F+path%3Aclang%2Ftest%2FDriver&type=code - also one using the MSVC toolchain: https://github.com/llvm/llvm-project/blob/a2134b8b758ce49c6ebad86c5afcd434b6a1a3f3/clang/test/Driver/hip-link-bundle-archive.hip#L56-L59

compnerd (approved):
Sounds good, only concern was related to the tests, the code change seems reasonable.

petrhosek:
Our builders started failing with the following error:
```
/b/s/w/ir/x/w/llvm-llvm-project/clang/test/Driver/msvc-link.c:57:16: error: DEBUG-LINK: expected string not found in input
// DEBUG-LINK: link.exe"
               ^
<stdin>:1:1: note: scanning from here
Fuchsia clang version 23.0.0git (https://llvm.googlesource.com/llvm-project cdc2749dffafaa1d43815d58303cc08516732f60)
^
<stdin>:7:2: note: possible intended match here
 "/b/s/w/ir/x/w/llvm_build/bin/lld-link" "-out:a.exe" "-defaultlib:libcmt" "-defaultlib:oldnames" "-libpath:lib" "-libpath:atlmfc/lib" "-nologo" "-debug" "/b/s/w/ir/x/w/llvm_build/lib/clang/23/lib/i686-pc-windows-msvc/clang_rt.builtins.lib" "/b/s/w/ir/x/t/lit-tmp-3_vg13l0/msvc-link-ecf918.o" 
```
The reason is that `clang/test/Driver/msvc-link.c` uses `-fuse-ld=` to test the default linker behavior and expects `link.exe` to be the default linker. However, on our builder, Clang is configured with `CLANG_DEFAULT_LINKER=lld`, which causes the driver to pick `lld-link` instead of `link.exe` when `-fuse-ld=` is empty.

petrhosek:
I created https://github.com/llvm/llvm-project/pull/199167 as a fix forward, but if we cannot get it reviewed and merged quickly, I intend to revert the change since this is currently breaking builders.

slydiman:
Note [lldb-x86_64-win](https://lab.llvm.org/staging/#/builders/211/builds/767)
```
XPASS: lldb-shell::TestCxxFrameFormatPartialFailure.test
******************** TEST 'lldb-shell :: Settings/TestCxxFrameFormatPartialFailure.test' FAILED ********************
Exit Code: 0
...
```
Please remove XPASS for this test.

clang/test/Driver/msvc-link.c
```diff
@@ -51,3 +51,21 @@
...
+
+// RUN: %clang --target=i686-pc-windows-msvc -g -fuse-ld=lld -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -g -fuse-ld=lld-link -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf-2 -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf-3 -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf-4 -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf-5 -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// RUN: %clang --target=i686-pc-windows-msvc -gdwarf-6 -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
+// DEBUG-LLD: lld-link"
```
bogner, line 69: I don't think it's correct to look for `lld-link"` specifically, as I would expect this to usually be `lld-link.exe"` on windows, no?

I'm seeing this fail locally for me, and also on some private bots I have access to. Unfortunately I can't seem to find any bots on the LLVM buildbot that are failing this way. In any case, the resulting command does look correct locally - given:
```
// RUN: %clang --target=i686-pc-windows-msvc -g -fuse-ld=lld -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LLD %s
```

I see
```
"e:\\env\\clang\\bin\\lld-link.exe" "-out:a.exe" "-defaultlib:libcmt" "-defaultlib:oldnames" "-nologo" "-debug" "C:\\Users\\...\\msvc-link-701109.o"
```

So I suspect it's just that the test is too strict.

Nerixyz: Allowing lld-link.exe too would be fine too. I can't open a PR today. If you want to open one that changes `lld-link"` to `{{lld-link(\.exe)?"}}`, feel free to do so.

PR #194404, opened 2026-04-27
[mlir][Tensor][Func] Rewrite single-dim Tensor return to Scalar

ioghiban:
Add a Tensor-dialect rewrite that scalarizes function results when a function returns:
- exactly one
- statically-shaped 
- single-element tensor. 

The rewrite updates the `func.func` result type from `tensor<...>` to the element type and rewrites each `func.return` by inserting a `tensor.extract` before returning the scalar.

Assisted-by: Codex (refine implementation + tests). I reviewed all code and tests before submission.

## Example
Before:
```mlir
func.func @rank1(%arg0: tensor<1xi64>) -> tensor<1xi64> {
  return %arg0 : tensor<1xi64>
}
```
After:
```mlir
func.func @rank1(%arg0: tensor<1xi64>) -> i64 {
  %c0 = arith.constant 0 : index
  %0 = tensor.extract %arg0[%c0] : tensor<1xi64>
  return %0 : i64
}
```
Rank-0 tensors are handled similarly, with `tensor.extract %t[]`, which gets constant-folded.

## Motivation
The `tosa-converter-for-flite` may produce functions that return tensor-wrapped scalars. These are semantically scalar values, but they remain wrapped in `tensor<1xT>` or `tensor<T>` function results.
This becomes a problem later in lowering, in particular for EmitC, where such results eventually bufferize to MemRef arrays and trigger:
```bash
error: 'emitc.func' op cannot return array type
```

## Scope
This rewrite intentionally handles only a narrow case:
- the function must return exactly one value,
- that value must be a ranked, static shaped, one-element tensor

The rewrite only changes the function result type and the corresponding `func.return` operations. It does not attempt broader tensor-to-scalar canonicalization across arbitrary ops or block arguments.

## Why ?
EmitC does not handle array return types because it models C/C++ function signatures, and in C/C++ a function cannot return an array by value. Arrays are storage objects, not first-class function return values. In practice, such cases are represented via:
- an out parameter,
- a pointer return,
- or a wrapper struct.

EmitC preserves C/C++ ABI-valid function signatures, so tensor-wrapped scalar returns are addressed before lowering reaches EmitC.

This patch handles this issue at the highest level where the intent is still explicit. At the Tensor dialect level, a single-element tensor result can still be recognized as a scalar value with tensor packaging, so scalarizing it here is simpler and more semantic than deferring the fix to MemRef or EmitC lowering.

## Correctness
The rewrite is semantics-preserving for statically-shaped single-element tensors:
[...]

PR #177579, opened 2026-01-23
[mlir][linalg] Remove abandoned `Detensorize` pass

matthias-springer:
RFC: https://discourse.llvm.org/t/how-to-deal-with-abandoned-unmaintained-code/89560

PR #198604, opened 2026-05-19
[EmitC] Scalarize single-element memref returns

ioghiban:
Extend Func-to-EmitC to scalarize private functions that return a single statically-shaped, identity-layout, single-element memref.

The converted `emitc.func` returns the memref element type instead of `memref<...>`, and each `func.return` is rewritten to load the unique element and return that scalar.

Pointer-backed memrefs, including rank-0 memrefs and `memref.alloc` results, are read with `emitc.subscript` + `emitc.load`. Array-backed memrefs use the existing EmitC array subscript path.

Assisted-by: Codex (refine implementation + tests). I reviewed all code and tests before submission.

## Example

Before:

```mlir
func.func private @rank0_alloc() -> memref<i64> {
  %0 = memref.alloc() : memref<i64>
  return %0 : memref<i64>
}
```

After:
```
emitc.func private @rank0_alloc() -> i64 {
  %alloc = emitc.call_opaque "malloc"(...) : (...) -> !emitc.ptr<i64>
  %c0 = emitc.constant 0 : index
  %lvalue = emitc.subscript %alloc[%c0]
      : (!emitc.ptr<i64>, index) -> !emitc.lvalue<i64>
  %value = emitc.load %lvalue : <i64>
  emitc.return %value : i64
}
```

## MemRef-to-EmitC Support

This also adds narrow rank-0 memref support:

- `memref<T>` converts to `!emitc.ptr<T>`
- rank-0 `memref.alloc` lowers to a one-element malloc
- rank-0 `memref.copy` lowers to scalar load/assign
- pointer-backed `memref.load` / `memref.store` lower through pointer subscript lvalues

Rank-0 copy uses subscript lvalues such as `src[0]` and `dst[0]`, instead of standalone `emitc.dereference`, to avoid invalid C emission from always-inline dereference expressions.

## Motivation

Some pipelines represent scalar-like values as one-element memrefs, especially after bufferizing tensor-wrapped scalar returns. EmitC cannot return arrays by value, and rank-0 memrefs still have storage semantics rather than scalar-value semantics.

Scalarizing private functions with no symbol users gives EmitC a valid C/C++ function signature while preserving the returned value.

## Scope

This is intentionally conservative. Scalarization applies only when:

- the function has exactly one result
- the result is a static, identity-layout memref with exactly one element
- the function is private and has no symbol users
- every returned memref value is produced by a supported source:
    - block argument
    - `memref.alloc`
    - `memref.get_global`

The rewrite does not change public APIs, functions with callers, multi-result functions, dynamic memrefs, non-identity layouts, multi-element memrefs, or arbitrary memref-producing ops.

In particular, it does not scalarize memrefs produced by unsupported operations such as:

- `arith.constant : memref<...>`
- `func.call` returning `memref<...>`

## Arith Constants

Arith-to-EmitC now rejects `arith.constant` operations whose result type is `memref<...>`.

With MemRef-to-EmitC type conversion enabled, such constants may otherwise try to convert to `!emitc.ptr<T>` or `!emitc.array<...>` while still carrying a memref-typed dense attribute. That is not a valid `emitc.constant`, and semantically a memref constant describes buffer contents, not a C pointer or array value.

Constant storage should instead flow through `memref.global` and `memref.get_global`, which have explicit EmitC storage lowering.

## Correctness

For a statically-shaped single-element memref, loading the only element preserves the value represented by the returned storage.

Restricting the rewrite to private functions with no symbol users avoids changing any externally visible API or leaving call sites expecting the old memref result type.


aniragil:
Thanks for the patch!
Some initial questions/comments:
- This actually seems more like a generic func -> func optimization/canonicalization than an emitc-specific lowering.
- What happens if the returned memref is stored to or deallocated by the caller?
- Why not lower `memref<0>`'s to a pointer?

banach-space:
Ioana is OOO for a few days, so let me quickly reply to keep to maintain the momentum.

> * This actually seems more like a generic func -> func optimization/canonicalization than an emitc-specific lowering.

That's a good point. In fact, the initial implementation tackled this at the Tensor level: https://github.com/llvm/llvm-project/pull/194404 

However, if we are to consider both function definitions and call-sites, then we need some global (module level) rewrite, e.g. a conversion pass and that makes `FuncToEmitC` a perfect candidate! A regular rewrite pattern doesn't seem like the right tool and we couldn't find a suitable alternative (or any prior art).

Also, there is a question of how general should this be? Ultimately, we are tackling the mismatch between Tensor and C semantics (hence many C-specific assumptions, e.g. that functions cannot return arrays) and for that reason it feels like EmitC conversion is the right level. As in, why would the Tensor dialect be concerned about functions returning tensors? (*)

Note that this PR makes quite a few assumptions, e.g. "the function has exactly one result" + 
"the result is a (...) memref with exactly one element". That's sufficient for our needs today and it also allowed Ioana to keep this PR relatively simple. For more generic cases, we probably should invest into your proposal to switch to memref descriptors when lowering to EmitC. 

Btw, there used to be a "de-tensorize" pass, but that was removed in https://github.com/llvm/llvm-project/pull/177579. We could re-visit that at some point, but that transformation focused on Linalg Ops rather than e.g. functions.

> * What happens if the returned memref is stored to or deallocated by the caller?

That will be a problem if there are users, yes. This patch bails out in such cases. If we generalise to also look at users, then we would:
* erase `memref.dealloc`,
* rewrite `memref.store` as `emit.assign`?

That should work, right?

> * Why not lower `memref<0>`'s to a pointer?

IMHO it's a convenient convention to treat `memref<1xi64>` + `memref<f64>` as plain scalars, but admittedly it's just a convention. We could treat those as pointers, but would that gain us anything?

---

To be perfectly honest, I feel that we are entering uncharted waters here and "we" (EmitC) may end-up having to iterate a bit. I am under the impression that MemRef descriptors will make lowering to EmitC more natural. In the meantime, we try to adapt to every case that we hit. But "descriptors" will likely take a while and perhaps first (independantly of this PR) we should jump on a call to discuss the plans in more detail. From MLGO perspective, I don't see any other requirements just now.

Hopefully this makes sense :)

-Andrzej

(*) Just to be clear - I am trying to say to it's not 100% clear to me what the right place for this transformation is. Right now I am leaning towards EmitC conversions, but if you feel differently, lets continue discussing.

aniragil:
> Ioana is OOO for a few days, so let me quickly reply to keep to maintain the momentum.
> 
> > * This actually seems more like a generic func -> func optimization/canonicalization than an emitc-specific lowering.
> 
> That's a good point. In fact, the initial implementation tackled this at the Tensor level: #194404

That PR seems open and under review, why is was it abandoned?

> 
> However, if we are to consider both function definitions and call-sites, then we need some global (module level) rewrite, e.g. a conversion pass and that makes `FuncToEmitC` a perfect candidate! A regular rewrite pattern doesn't seem like the right tool and we couldn't find a suitable alternative (or any prior art).

Wouldn't a module pass doing that transformation suffice?

> 
> Also, there is a question of how general should this be? Ultimately, we are tackling the mismatch between Tensor and C semantics (hence many C-specific assumptions, e.g. that functions cannot return arrays) and for that reason it feels like EmitC conversion is the right level. As in, why would the Tensor dialect be concerned about functions returning tensors? (*)

Tensors, being pure SSA values, make such transformations much simpler than memrefs with their load-store/aliasing semantics. Since at some point each tensor should either be bufferized into a memref or casted to a scalar, it seems easier to avoid bufferization than later proving they can be removed.

> 
> Note that this PR makes quite a few assumptions, e.g. "the function has exactly one result" + "the result is a (...) memref with exactly one element". That's sufficient for our needs today and it also allowed Ioana to keep this PR relatively simple.

The patch indeed aims to be handle only specific cases, but not sure I'd call it simple :)

> For more generic cases, we probably should invest into your proposal to switch to memref descriptors when lowering to EmitC.

Right, the memref-to-array lowering causes all sorts of issues and limitations. The memref-descriptor proposal should hopefully handle this completely and uniformly (as in the LLVM dialect), preferably before these issue cause this kind of workarounds which would be hard to deprecate later.

> Btw, there used to be a "de-tensorize" pass, but that was removed in #177579. We could re-visit that at some point, but that transformation focused on Linalg Ops rather than e.g. functions.

That actually supports the direction of this canonicalization being done earlier. IIUC from the Discourse discussion the pass wasn't "wrong" in any way, just not owned/developed by anyone.

> 
> > * What happens if the returned memref is stored to or deallocated by the caller?
> 
> That will be a problem if there are users, yes. This patch bails out in such cases. If we generalise to also look at users, then we would:
> 
>     * erase `memref.dealloc`,
> 
>     * rewrite `memref.store` as `emit.assign`?
> 
> 
> That should work, right?

Hmmm ... so the code (and commit message) actually state the transformation will only be done on private and unused functions. But aren't such functions dead and can just be deleted?

> 
> > * Why not lower `memref<0>`'s to a pointer?
> 
> IMHO it's a convenient convention to treat `memref<1xi64>` + `memref<f64>` as plain scalars, but admittedly it's just a convention. We could treat those as pointers, but would that gain us anything?

So in the motivation paragraph it states "EmitC cannot return arrays by value" as the problem. For zero-ranked memrefs pointers seem to be a straightforward and robust way to lower such functions.
I agree doing sort of a "Mem2Reg" is preferable when provably possible (which isn't always simple), but it sounds more like a canonicalization than a lowering.

> 
> To be perfectly honest, I feel that we are entering uncharted waters here and "we" (EmitC) may end-up having to iterate a bit. I am under the impression that MemRef descriptors will make lowering to EmitC more natural. In the meantime, we try to adapt to every case that we hit. But "descriptors" will likely take a while and perhaps first (independantly of this PR) we should jump on a call to discuss the plans in more detail. From MLGO perspective, I don't see any other requirements just now.

Agreed. I started laying the groundwork for memref descriptors as structs with with #198198 (another PR soon to follow), then I hope I can upload a patch and an RFC we can use as a basis for more detailed planning

> 
> (*) Just to be clear - I am trying to say to it's not 100% clear to me what the right place for this transformation is. Right now I am leaning towards EmitC conversions, but if you feel differently, lets continue discussing.

So perhaps a temporary solution could be the "MLGO-specific passes" mechanism you and @mtrofin suggested: Instead of converting zero-rank memrefs through the common lowering passes while we're unsure if and where such a transformation belongs, could this transformation be implemented as a module pass working on `func.func` and run by MLGO before lowering to emitc?

banach-space:
Thanks for the feedback, Gil! 

> That PR seems open and under review, why is was it abandoned?
 
TBH, we got busy trying different approaches. We also wanted to keep it open for reference - in case you suggested that as an option :) 

If you are happy with the path suggested here, we can close that one. In the meantime, I suggest minimising the noise.  

> Wouldn't a module pass doing that transformation suffice?

Yes, @ioghiban will sent a separate PR with a module pass shortly. (*)

(*) IMHO a separate PR is "cleaner" than completely rewriting an existing PR, e.g. this one. I hope that that's OK?

> Tensors, being pure SSA values, make such transformations much simpler than memrefs with their load-store/aliasing semantics. Since at some point each tensor should either be bufferized into a memref or casted to a scalar, it seems easier to avoid bufferization than later proving they can be removed.

These are very good points for implementing this at the Tensor level, yes. 

In fact, we should actually either ask the authors of those models or patch the TFLite importer to avoid those 1-element tensors to begin with - that's my long-term goal 😅 

~In the meantime, tweaking this at EmitC/MemRef level feels like the most straightforward path for us. That's modulo all the issues that you've mentioned and which we are trying to avoid by bailing out conservatively.~

**EDIT** (27/5/26)

I've crossed out the comment above as it fails to take into account that #199686 operates at the Tensor level. Apologies for the confusion, I was processing too many threads in parallel when writing that originally:) 

To quickly summarise the MemRef vs Tensor debate:
* Leveraging the EmitC "conversions" logic would mean operating at the post-bufferization level. That would simplifie a lot by simply leveraging the pre-existing infra, but would also make the transformation trickier due to MemRefs (aliasing, load/store, etc).
*  Operating at the Tensor level would be much safer (fewer surprises and corner cases at the SSA level), but there doesn't seem to be a good infra to operate module-wide.

> The patch indeed aims to be handle only specific cases, but not sure I'd call it simple :)

Fair :)

> That actually supports the direction of this canonicalization being done earlier. IIUC from the Discourse discussion the pass wasn't "wrong" in any way, just not owned/developed by anyone.

We couldn't even find any users of that code :( 

It does feel like a potential Tensor canonicalization. Even once we gain memref-descriptors in EmitC, fixing this at the Tensor level would be "neater".

My personal preference would be to start with the approach taken here and then possibly generalise to Tensor. Basically, gain experience by applying this within EmitC and then use that learning to generalise. I know that this makes EmitC a guinea pig, but it allows us to approach this incrementally  and discuss between ourselves (i.e. EmitC) before expanding.

> Hmmm ... so the code (and commit message) actually state the transformation will only be done on private and unused functions. But aren't such functions dead and can just be deleted?

Good point. But then how do we "label" code that we know won't be used in any other MLIR module (so MLIR transformations are safe), but can be used from C++ (i.e. outside of MLIR)? 

That's roughly the situation that we have here and feels like a gap between bridging MLIR and C 🤔 Should we look into creating another attribute?

> For zero-ranked memrefs pointers seem to be a straightforward and robust way to lower such functions.

You might be right (I don't want to be definitive without trying), but then we will be special-casing for `memref<f32>`. That's fine with me, but some people dislike that. I also prefer to treat everything consistently and in general view `f32`, `memref<f32>` and `memref<1xf32>` as identical types (i.e. scalars). 

> So perhaps a temporary solution could be the "MLGO-specific passes" mechanism you and @mtrofin suggested: Instead of converting zero-rank memrefs through the common lowering passes while we're unsure if and where such a transformation belongs, could this transformation be implemented as a module pass working on `func.func` and run by MLGO before lowering to emitc?

Sounds like a plan, thank you for the suggestion! Like I mentioned earlier, Ioana will be sending a PR soonish.

**EDIT:** (26/5/26)
New patch from @ioghiban that's meant to supersede this one: https://github.com/llvm/llvm-project/pull/199686

banach-space:
> That PR seems open and under review, why is was it abandoned?

I have just closed it, see this [post](https://github.com/llvm/llvm-project/pull/194404#issuecomment-4566078602l).

> So perhaps a temporary solution could be the "MLGO-specific passes" mechanism you and @mtrofin suggested: Instead of converting zero-rank memrefs through the common lowering passes while we're unsure if and where such a transformation belongs, could this transformation be implemented as a module pass working on func.func and run by MLGO before lowering to emitc?

Since that's implemented in https://github.com/llvm/llvm-project/pull/199686 and we seem to agree that's the preferable approach, let me close this one to make sure that the discussion continues in the right place (i.e. [here](https://github.com/llvm/llvm-project/pull/199686))  :) 

Thank you 🙏🏻

Issue #198523, opened 2026-05-19
[Flang] Flang incorrectly accepts keyword arguments in calls to statement functions


blazie2004:
Flang fails to diagnose keyword arguments in calls to statement functions. Keyword arguments are only valid when the interface is explicit. Flang silently accepts keyword arguments in calls to procedures with implicit interfaces. 

**Standard: F2018 §15.5.1 C1535**: "In a reference to a procedure whose interface is implicit at the point of the reference, the actual argument shall not be a keyword argument."

**Reproducer**
```
program test
  integer :: f1, x, c
  f1(x) = x / 2
  c = f1(x=10)  ! Should be an error
end program
```

**Expected Behavior**
```
error: Keyword 'x=' may not appear in a reference to a procedure with an implicit interface
```

**Actual behavior:**

Silently accepts the code

eugeneepshteyn:
NAG, gfortran 16.1, ifx, nvfortran all reject this with an error.

eugeneepshteyn:
@blazie2004 , a request: if you plan to work on an issue, please assign it to yourself (or ask someone to assign it to you). It helps to keep track of which issues are available for others to work on. Thank you!

eugeneepshteyn:
This was already fixed by the PR mentioned above.

PR #198610, opened 2026-05-19
[Flang] Reject keyword arguments in statement function calls

blazie2004:
**Problem**
Flang silently accepted keyword arguments in calls to statement functions, violating F2018 C1535.

**Standard: F2018 §15.5.1 C1535**: In a reference to a procedure whose interface is implicit at the point of the reference, the actual argument shall not be a keyword argument.

Flang silently compiles the following code without giving error` Keyword argument 'x' at (1) is invalid in a statement function
`
```
program test
  integer :: f1, x, c
  f1(x) = x / 2
  c = f1(x=10)  ! Should be an error
end program

```

**Summary**
Fixed an issue where statement functions were incorrectly treated as having an explicit interface, causing argument checks to be skipped. Now, statement functions are marked correctly so existing checks run and proper errors are shown

**Fixes** : [198523](https://github.com/llvm/llvm-project/issues/198523)

### DIFF
--- a/flang/lib/Evaluate/characteristics.cpp
+++ b/flang/lib/Evaluate/characteristics.cpp
@@ -642,6 +642,10 @@ static std::optional<Procedure> CharacterizeProcedure(
           [&](const semantics::SubprogramDetails &subp)
               -> std::optional<Procedure> {
             Procedure result;
+            if (subp.stmtFunction()) {
+              // Statement functions have implicit interfaces (F'2018 15.5.1)
+              result.attrs.set(Procedure::Attr::ImplicitInterface);
+            }
             if (subp.isFunction()) {
               if (auto fr{CharacterizeFunctionResult(
                       subp.result(), context, seenProcs, emitError)}) {
--- a/flang/lib/Evaluate/characteristics.cpp
+++ b/flang/lib/Evaluate/characteristics.cpp
@@ -642,10 +642,6 @@ static std::optional<Procedure> CharacterizeProcedure(
           [&](const semantics::SubprogramDetails &subp)
               -> std::optional<Procedure> {
             Procedure result;
-            if (subp.stmtFunction()) {
-              // Statement functions have implicit interfaces (F'2018 15.5.1)
-              result.attrs.set(Procedure::Attr::ImplicitInterface);
-            }
             if (subp.isFunction()) {
               if (auto fr{CharacterizeFunctionResult(
                       subp.result(), context, seenProcs, emitError)}) {
--- a/flang/lib/Semantics/expression.cpp
+++ b/flang/lib/Semantics/expression.cpp
@@ -3770,9 +3770,8 @@ std::optional<characteristics::Procedure> ExpressionAnalyzer::CheckCall(
       IsExternalCalledImplicitly(callSite, proc.GetSymbol())};
   const Symbol *procSymbol{proc.GetSymbol()};
   // Statement functions have implicit interfaces and require the same checks
-  bool isStatementFunction{procSymbol &&
-      procSymbol->has<semantics::SubprogramDetails>() &&
-      procSymbol->get<semantics::SubprogramDetails>().stmtFunction()};
+  bool isStatementFunction{
+      procSymbol && procSymbol->flags().test(Symbol::Flag::StmtFunction)};
   std::optional<characteristics::Procedure> chars;
   if (procSymbol && procSymbol->has<semantics::ProcEntityDetails>() &&
       procSymbol->owner().IsGlobal()) {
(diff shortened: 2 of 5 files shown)


flang/lib/Evaluate/characteristics.cpp
```diff
@@ -642,6 +642,10 @@ static std::optional<Procedure> CharacterizeProcedure(
           [&](const semantics::SubprogramDetails &subp)
               -> std::optional<Procedure> {
             Procedure result;
+            if (subp.stmtFunction()) {
+              // Statement functions have implicit interfaces (F'2018 15.5.1)
+              result.attrs.set(Procedure::Attr::ImplicitInterface);
```
eugeneepshteyn: For front end implementation, giving all statement functions an implicit interface attribute may be too big of a hammer. This could trigger all sorts of code paths not meant for the statement functions.

Could you please find an alternative way of doing this? Perhaps in `ExpressionAnalyzer::CheckCall()` have a special case for statement function that calls `CheckArguments()` that still calls `CheckImplicitInterfaceArg()` for each argument?

blazie2004: Thanks , I have updated the implementation to avoid the broader global change. Instead, I added a targeted check in `ExpressionAnalyzer::CheckCall()` for statement functions so they still get the required implicit interface argument checks, without affecting other statement function behavior.

flang/test/Semantics/call47.f90
```diff
@@ -0,0 +1,15 @@
...
+program test_stmt_func_keyword
+  integer :: f1, x, c
+
+  ! Statement function definition
+  f1(x) = x / 2
+
+  ! Calling a statement function with a keyword argument is not allowed
+  ! because statement functions have implicit interfaces.
+  !ERROR: Keyword 'x=' may not appear in a reference to a procedure with an implicit interface
+  c = f1(x=10)
```
eugeneepshteyn, line 15: Please also try the case `c = f1(y=10)`. Do you get just your error or also another error?

flang/lib/Semantics/expression.cpp
```diff
@@ -3769,6 +3769,10 @@ std::optional<characteristics::Procedure> ExpressionAnalyzer::CheckCall(
   bool treatExternalAsImplicit{
       IsExternalCalledImplicitly(callSite, proc.GetSymbol())};
   const Symbol *procSymbol{proc.GetSymbol()};
+  // Statement functions have implicit interfaces and require the same checks
+  bool isStatementFunction{procSymbol &&
+      procSymbol->has<semantics::SubprogramDetails>() &&
+      procSymbol->get<semantics::SubprogramDetails>().stmtFunction()};
```
eugeneepshteyn: I think a canonical way to do this would be `procSymbol>flags().test(Symbol::Flag::StmtFunction)`

flang/test/Semantics/call47.f90
eugeneepshteyn: In addition, please also add a case for a statement function with two args, to ensure that each keyword arg gets its own diagnostics.

flang/test/Semantics/call47.f90
```diff
@@ -0,0 +1,15 @@
+! RUN: %python %S/test_errors.py %s %flang_fc1
+! Test F'2018 15.5.1 C1535 - keyword arguments are not allowed when the
```
eugeneepshteyn: Should be `F2018`, without apostrophe

flang/test/Semantics/call47.f90
blazie2004: for `c = f1(y=10)` , i get this 
```
error: Keyword 'y=' may not appear in a reference to a procedure with an implicit interface
    c = f1(y=10)
           ^
error: Argument keyword 'y=' is not recognized for this procedure reference
    c = f1(y=10)
           ^
```

for stmt function with two args , i have changed the testcase which produces

```
error: Keyword 'x=' may not appear in a reference to a procedure with an implicit interface
    c = f2(x=10, y=20)
           ^
error: Keyword 'y=' may not appear in a reference to a procedure with an implicit interface
    c = f2(x=10, y=20)
                 ^
```

flang/test/Semantics/call47.f90
blazie2004: resolved

flang/lib/Semantics/expression.cpp
blazie2004: resolved

blazie2004:
ping @eugeneepshteyn

eugeneepshteyn (approved):
LGTM. Thank you!

PR #165428, opened 2025-10-28
[libc++][z/OS] Move z/OS to new locale API and resolve all name collisions

zibi2:

PR #198614, opened 2026-05-19
[libc++] Update z/OS to use the new locale API

philnik777:
z/OS is the last platform that hasn't switched to the new locale API. This patch updates the platform definitions on a best-effort basis. AFAIK it is already broken on z/OS currently, so there shouldn't be much fallout of this if it doesn't work 100%.


philnik777:
@zibi2 Are you OK with this in the interest of removing a bunch of legacy/boilerplate code? This should be the non-contentious part, so we can concentrate on the macro changes in #165428 separately.

libcxx/include/__locale_dir/support/mvs.h
```diff
@@ -0,0 +1,316 @@
...
+};
+
+//
+// Locale management
+//
+
+#define _LC_MAX LC_MESSAGES /* highest real category */
+#define _NCAT (_LC_MAX + 1) /* maximum + 1 */
+#define _CATMASK(n) (1 << (n))
+
```
zibi2:
```suggestion
```

libcxx/include/__locale_dir/support/mvs.h
```diff
@@ -0,0 +1,316 @@
...
+#define _LIBCPP_COLLATE_MASK _CATMASK(LC_COLLATE)
+#define _LIBCPP_CTYPE_MASK _CATMASK(LC_CTYPE)
+#define _LIBCPP_MONETARY_MASK _CATMASK(LC_MONETARY)
+#define _LIBCPP_NUMERIC_MASK _CATMASK(LC_NUMERIC)
+#define _LIBCPP_TIME_MASK _CATMASK(LC_TIME)
+#define _LIBCPP_MESSAGES_MASK _CATMASK(LC_MESSAGES)
+#define _LIBCPP_ALL_MASK (_CATMASK(_NCAT) - 1)
```
zibi2:
```suggestion
#define _LIBCPP_ALL_MASK LC_ALL_MASK
```

libcxx/include/__locale_dir/support/zos.h
```diff
@@ -0,0 +1,316 @@
...
+  int __res = vasprintf(__s, __format, __va);
+  va_end(__va);
+  return __res;
+}
+
+template <class... _Args>
+_LIBCPP_VARIADIC_ATTRIBUTE_FORMAT(__printf__, 3, 4)
+int __asprintf(char** __s, __locale_t __loc, const char* __format, _Args&&... __args) {
+  return __locale::__libcpp_asprintf_l(__s, __loc, __format, std::forward<_Args>(__args)...);
+}
```
zibi2, line 313: Why don't you define this in a separate file `libcxx/include/__support/ibm/vasprintf.h` proposed in https://github.com/llvm/llvm-project/pull/165428

zibi2 (commented):
I prefer you rename `libcxx/include/__locale_dir/support/mvs.h` to `libcxx/include/__locale_dir/support/zos.h` and other changes to be based what was proposed in https://github.com/llvm/llvm-project/pull/165428.

zibi2:
> @zibi2 Are you OK with this in the interest of removing a bunch of legacy/boilerplate code? This should be the non-contentious part, so we can concentrate on the macro changes in #165428 separately.

Thanks for the consideration. I appreciate the effort to clean up legacy code, but I'd prefer to keep the scope focused on #165428 rather than introducing hybrid changes across multiple PRs.

My concern is that splitting the work this way creates additional review overhead and potential merge conflicts. I'd rather take the complete subset of changes in #165428 as a cohesive unit, even if it means reviewing more code at once. This approach will make it easier to understand the full context and ensure consistency across the changes.

If there are specific non-contentious cleanups that are truly independent, perhaps those could be addressed separately after #165428 lands?

libcxx/include/__locale_dir/locale_base_api.h
```diff
@@ -126,175 +126,10 @@
...
-    char** __s, __locale_t __loc, const char* __format, _Args&&... __args) {
-  return std::__libcpp_asprintf_l(__s, __loc, __format, std::forward<_Args>(__args)...);
-}
-_LIBCPP_DIAGNOSTIC_POP
-#    undef _LIBCPP_VARIADIC_ATTRIBUTE_FORMAT
-
-} // namespace __locale
-_LIBCPP_END_NAMESPACE_STD
-
+#    warning "No known way to provide the locale base API"
```
ldionne: Let's make this an error. Nothing is going to work anyway, so we might as well just stop compilation here.

libcxx/include/__locale_dir/support/zos.h
```diff
@@ -0,0 +1,316 @@
...
+_LIBCPP_CLANG_DIAGNOSTIC_IGNORED("-Wgcc-compat")
+_LIBCPP_GCC_DIAGNOSTIC_IGNORED("-Wformat-nonliteral") // GCC doesn't support [[gnu::format]] on variadic templates
+#ifdef _LIBCPP_COMPILER_CLANG_BASED
+#  define _LIBCPP_VARIADIC_ATTRIBUTE_FORMAT(...) _LIBCPP_ATTRIBUTE_FORMAT(__VA_ARGS__)
+#else
+#  define _LIBCPP_VARIADIC_ATTRIBUTE_FORMAT(...) /* nothing */
+#endif
+
+// FIXME: This should be inlined
+inline _LIBCPP_ATTRIBUTE_FORMAT(__printf__, 4, 5) int __libcpp_snprintf_l(
```
ldionne, line 256: @philnik777 Why did you keep this around in this patch?

libcxx/include/__locale_dir/support/zos.h
ldionne: I think it's kinda overkill to introduce a new header just for this helper.

PR #198618, opened 2026-05-19
[Liveness][analyzer] Fix handling of [[assume]] attributes 

NagyDonat:
Before this commit, if the analyzer encountered code like
```
int f(int a, int b) {
  [[assume(a == 2), assume(b == 3)]];
  return a + b;
}
```
it performed the following steps:
1. It visited the expression `a == 2` with `ExprEngine::Visit` (after visiting its sub-expressions, within the regular visitation that visits each statement of the `CFGBlock`). This triggered the `EagerlyAssume` logic and separated two execution paths.
2. It discarded the result bound to `a == 2` from the `Environment` because `a == 2` is not a direct child of the `AttributedStmt`.
3. Analogously, it visited an evaluated `b == 3`.
4. Analogously, it discarded the result bound to `b == 3`.
5. On each execution path `VisitAttributedStmt` was reached, it ran the `PreStmt<AttributedStmt>` checkers and stored their results in `CheckerPreStmt`. As there are AFAIK no such checkers, after this step the only element of `CheckerPreStmt` was the node `Pred`.
6. It constructed a `NodeBuilder`, which was unused, but its constructor inserted the nodes in `CheckerPreStmt` into `EvalSet`. This was essentially a shortcut that skipped the handling of the assumptions -- but ultimately its effects were absorbed and negated by the tricky side effects of `ExprEngine::Visit`. 
7. For each node `N` in `CheckerPreStmt` (i.e. for `N == Pred`) and for each assume attribute, it created a child node of `N` where the assumption of **that one attribute** was visited again with `ExprEngine::Visit`. (I.e. it separated execution paths and implicitly placed "OR" relationships between the assumptions.)
8. It **completely ignored** the value produced by the assumption expression: did not update the state and did not discard execution paths where the [[assume]] assumption contradicted earlier knowledge.

This commit fixes all four logic errors:
1. The liveness analysis is updated to ensure that an `AttributedStmt` keeps alive the assumption expressions of its [[assume]] attributes. This ensured that their results remain in the `Environment`.
2. Now there is no `NodeBuilder` that skips the handling the assumptions.
3. The invocation of `ExprEngine::Visit` is removed from `VisitAttributedStmt`: there is no reason to revisit those statements
   (now that we don't lose the evaluation results).
4. The state is updated to record that the values of the assumption expressions are ALL assumed to be true. If the assumption leads to a contradiction, the execution path is discarded (by not generating a node).

This commit also adds a few new testcases and resolves some FIXME comments in previously existing tests.


clang/lib/StaticAnalyzer/Core/ExprEngineCXX.cpp
```diff
@@ -1224,16 +1224,33 @@ void ExprEngine::VisitLambdaExpr(const LambdaExpr *LE, ExplodedNode *Pred,
...
-    for (ExplodedNode *N : CheckerPreStmt) {
-      Visit(Attr->getAssumption()->IgnoreParens(), N, EvalSet);
+  for (ExplodedNode *N : CheckerPreStmt) {
+    ProgramStateRef State = N->getState();
+    for (const auto *Attr : getSpecificAttrs<CXXAssumeAttr>(A->getAttrs())) {
+      SVal AssumedVal = State->getSVal(Attr->getAssumption(), LCtx);
+      if (auto ValidAssumedVal = AssumedVal.getAs<DefinedOrUnknownSVal>()) {
+        State = State->assume(*ValidAssumedVal, true);
+      } else {
+        // The assumption expression has evaluated to UndefinedVal.
```
Xazax-hun: Do we actually want to throw these execution paths out? Traditionally, just because something produced undef, we usually don't terminate the exploration. And when we are not sure if something evaluates to true or false we traditionally explore both options. What is the rationale here?

clang/lib/Analysis/LiveVariables.cpp
```diff
@@ -279,6 +279,16 @@ void TransferFunctions::Visit(Stmt *S) {
...
       break;
     }
+    case Stmt::AttributedStmtClass: {
+      // In an attributed statement, include the assumptions of the
+      // [[assume(...)]] attributes as being live.
+      AttributedStmt *AS = cast<AttributedStmt>(S);
+      for (const auto *Attr : getSpecificAttrs<CXXAssumeAttr>(AS->getAttrs())) {
+        AddLiveExpr(val.liveExprs, LV.ESetFact, Attr->getAssumption());
+      }
+      // FIXME: Also handle other attributes that have relevant substatements.
```
Xazax-hun: Is this a FIXME if we don't know whether attributes like those exist? I think this could be dropped, similar to the FIXME below in the other file.

NagyDonat: Yes, I added the FIXMEs speculatively. Now I scrolled through the list of attributes supported by clang and I didn't see any other analogous one, so I'll remove the FIXME comments.

clang/lib/StaticAnalyzer/Core/ExprEngineCXX.cpp
NagyDonat: > Traditionally, just because something produced undef, we usually don't terminate the exploration.

You are right that _producing_ an undefined (not just unknown) value is not always reported (e.g. the contents of the buffer returned by `malloc` are undef and this is fine), although even there we have `core.UndefinedBinaryOperatorResult` which reports situations when a binary operator produces undef.

On the other hand, all the reasonably common ways of _using_ undef are all reported by various `core` checkers:
- assigning undef to a variable is reported by `core.uninitialized.Assign`,
- using undef in the condition of a control flow statement is reported by `core.uninitialized.Branch`,
- passing undef to a function is reported by `core.CallAndMessage`, ...

> What is the rationale here?

I think using a value in `[[assume(...)]]`` is analogous to passing it to a function or using it in the condition of an `if`, so the analyzer should behave analogously, i.e. sink the execution path. This would be consistent with the behavior of an `assert(...)` call.

Also, I think it would be very rare to `[[assume(...)]]` expressions that are really undefined, so is suspect that a significant fraction of the (rare) `UndefinedVal`s encountered at this point would be produced by bugs in the analyzer engine. (And in that case discarding the inconsistent state is the right thing to do.)

However I can also accept "just ignore the undefined assumption" if you strongly prefer that.

Xazax-hun: I think my main concern is that if there is something that we do not model in the assume expression, we would lose coverage for all the code after that statement even though the result of that expression would not affect leak into the program state. And the users would not be notified that the code is not being analyzer since we do not produce any warnings based on the undef.

NagyDonat: I prefer to err on the side of lost coverage instead of producing convoluted false positives from an inconsistent state; but this seems to be a very rare corner case, so I'm fine with any behavior that does not crash outright.

I can change the code to ignore the assumption expressions that evaluate to undef if you think that it would be a relevant improvement.

steakhal: If I understand this correctly, in the past if the assume-expr was Undef, we just continued.
(This is proven by the fact that we report [here](https://godbolt.org/z/WYrx5fd3r).)
After this patch, we would no longer get this actually TP report, right?

We could have 2 options here:
 - Have a checker that would report that the assume expr is Undef in a `runCheckersForPreStmt` handler, or
 - Skip adding the `assume true` constraint to the apparently `undef` assume expr.
 
The latter would make sure that it behaves as if we didn't take the assumption into account - as if it was completely ignored. Implementations are allowed to ignore them.

IMO approach (1) is better, and isn't that when iterating the `CheckerPreStmt` it would skip sink nodes? Thus would naturally imply that the assumption expr can't be Undef here? (assuming that the imaginary core checker reporting undef assumptions is enabled)?

clang/lib/Analysis/LiveVariables.cpp
steakhal: I remember in the past I also went over the attributes and didn't find anything similar to this. Feel free to drop this FIXME.

clang/lib/StaticAnalyzer/Core/ExprEngineCXX.cpp
NagyDonat: > If I understand this correctly, in the past if the assume-expr was Undef, we just continued.
> (This is proven by the fact that we report [here](https://godbolt.org/z/WYrx5fd3r).)
> After this patch, we would no longer get this actually TP report, right?

Yes, in the past the value of the assume-expr was ignored by the analyzer [1], so we just continued even if it was undef. After this patch we would no longer get the true positive report that you linked.

> We could have 2 options here:
> 
>   * Have a checker that would report that the assume expr is Undef in a `runCheckersForPreStmt` handler, or
> 
>   * Skip adding the `assume true` constraint to the apparently `undef` assume expr.
> 
> 
> The latter would make sure that it behaves as if we didn't take the assumption into account - as if it was completely ignored. Implementations are allowed to ignore them.
> 
> IMO approach (1) is better, and isn't that when iterating the `CheckerPreStmt` it would skip sink nodes?

`ExplodedNodeSet` instances (e.g. `CheckerPreStmt`) have an invariant that they never contain sink nodes (trying to insert a sink into an exploded node set has no effect).

> Thus would naturally imply that the assumption expr can't be Undef here? (assuming that the imaginary core checker reporting undef assumptions is enabled)?

Yes, if the imaginary undef assumption checker is enabled, the assumption expr can't be Undef here.

I agree that (1) would be the best behavior, but I don't want to squeeze creating a new checker into this patch. In this patch I would like to implement some simple behavior (either sinking or skipping the undef assumption expr) and then a follow-up commit can implement that simple core checker. (I probably won't have time for that in the foreseeable future.)

-----

[1] There was a test where code after `[[assume(false)]];` was not reachable; but I don't exactly why did this work. I suspect that this literal false may have been recognized by the CFG generation logic, because I didn't see any code in the analyzer that can see any difference between e.g. `assume(true)` and `assume(false)`.

steakhal: It's fine to not introduce a new checker right now. We should not sink however. We should rather model it as identity at that point - which is not what this PR does.

NagyDonat: Done in https://github.com/llvm/llvm-project/pull/198618/commits/e189b20fd21b0c08de46899e3500f0a4aed15a10

clang/lib/Analysis/LiveVariables.cpp
NagyDonat: Done in https://github.com/llvm/llvm-project/pull/198618/commits/0e3622db66807868854daa24a7b3fd467f143222

steakhal (approved):
LGTM. I wonder if we usually have dedicated tests for LiveVariables. I think that component deserves some better coverage, or dedicated coverage in some way.

Anyway, the patch looks great already! Thank you.

NagyDonat:
> I wonder if we usually have dedicated tests for LiveVariables. I think that component deserves some better coverage, or dedicated coverage in some way.

I don't know whether it has proper coverage. Unfortunately I know almost nothing about that component, so it would take far too much time for me to understand its behavior, understand its testing patterns and then fix anything that is missing / problematic.

steakhal:
Ask next time. I have studied that code quite a bit already.

NagyDonat:
> Ask next time. I have studied that code quite a bit already.

I happened to be familiar with the concrete parts of `LiveVariables` that I needed to modify here (because I accidentally encountered it during some investigation a few weeks ago), but I don't have an overall familiarity with the area as a whole.

NagyDonat:
I adjusted this PR by using the name 'StackFrame' instead of 'LocationContext' (because https://github.com/llvm/llvm-project/pull/198211 was merged recently). As this is a completely mechanical change, I'm merging this without further review.

NagyDonat:
For a moment I was scared by the buildbot result https://lab.llvm.org/buildbot/#/builders/114/builds/747 where the compilation of `RangeConstraintManager.cpp` had failed with the following error message (where LINE REPEATED and LLVM REPO are added by me):
```
In function ‘_ForwardIterator1 std::__search(_ForwardIterator1, _ForwardIterator1, _ForwardIterator2, _ForwardIterator2, _BinaryPredicate)’:
cc1plus: error: expected primary-expression before ‘=’ token [-Wtemplate-body]  (LINE REPEATED x 10)
In file included from /usr/include/c++/15/bits/stl_uninitialized.h:63,
                 from /usr/include/c++/15/memory:71,
                 from (LLVM REPO)/llvm/include/llvm/Support/Casting.h:20,
                 from (LLVM REPO)/clang/include/clang/Basic/LLVM.h:21,
                 from (LLVM REPO)/clang/include/clang/Basic/JsonSupport.h:12,
                 from (LLVM REPO)/clang/lib/StaticAnalyzer/Core/RangeConstraintManager.cpp:14:
/usr/include/c++/15/bits/stl_algobase.h:2157:28: error: expected primary-expression before ‘=’ token [-Wtemplate-body]
 2157 |                 __gnu_cxx::__ops::__iter_comp_iter(__predicate, __first2));
      |                            ^~~~~
/usr/include/c++/15/bits/stl_algobase.h:2157:35: error: ‘::__iter_comp_iter’ has not been declared; did you mean ‘__gnu_cxx::__ops::__iter_comp_iter’? [-Wtemplate-body]
 2157 |                 __gnu_cxx::__ops::__iter_comp_iter(__predicate, __first2));
      |                                   ^~~~~~~~~~~~~~~~
      |                                   __gnu_cxx::__ops::__iter_comp_iter
In file included from /usr/include/c++/15/bits/stl_algobase.h:71:
/usr/include/c++/15/bits/predefined_ops.h:374:5: note: ‘__gnu_cxx::__ops::__iter_comp_iter’ declared here
  374 |     __iter_comp_iter(_Iter_comp_iter<_Compare> __comp, _Iterator __it)
      |     ^~~~~~~~~~~~~~~~
```
(and this commit was the only fresh CSA commit), but it turns out that the next job https://lab.llvm.org/buildbot/#/builders/114/builds/748 succeeded on the same buildbot (and I verified that it also includes this commit).

Based on this successful result I'm confident that we won't need to revert this commit.

PR #198619, opened 2026-05-19
[NVPTX] : Make an optimistic assumption that loads from generic address space are not divergent. This assumption discovers more non-divergent expressions.

asastry108:
Currently NVPTX specific divergence analysis always makes a load from a generic pointer always divergent.  This is correct but conservative.
Instead, now we treat a generic load as not divergent and let the analysis update it as divergent if it finds a path from a divergent expreesion to this load.  It allows us to recognize more non-divergent expressions and thus improves the optimization capability.

Added a test case that shows that optimistic divergence is correct and can identify more non-divergent expressions


asastry108:
@akrolik@nvidia.com,  please review

AlexMaclean:
Wouldn't this incorrectly identify %v as uniform in this example? 

```
%a = alloca i32  
store i32 %tid_value, ptr %a
%v = load i32, ptr %a         
```

Maybe if we make the rules for `alloca` and `addrspacecast` tight enough this can work, but right now this seems like there are cases where it will not be sound anymore.

asastry108:
alloca should return a pointer that is address space local.  So it should be marked as divergent.   addressspacecast  should just propagagate the divergence of its source and fix point computation should take care of it.   Just wondering if  is sufficient to mark the alloca as divergent.  Thanks for the observation

asastry108:
Added alloca as a source of divergence and the test case that shows correct conclusion.  addrspacecast doesn't need to have any change because it simply propagates the divergence of the source

asastry108:
Anyone reviewing this change ?  Just to give some motivation, this change can enable more loop unswitching if a conditional branch is correctly recognized as non-divergent

asastry108:
@jholewinski,  can you please review this change

asastry108:
Any one reviewing this change ?

AlexMaclean (changes requested):
Change still seems incorrect to me.

Take the example I provided above and instead of reading from the alloca directly imagine storing the pointer to a global or shared memory and later loading it back from a pointer that aliases. This change would incorrectly mark the load from the alloca as uniform.

asastry108:
@AlexMaclean,  In your example, when all threads write the alloca address to global or shared memory, they have to synchronize otherwise the gpu program is not correct.   Once synchronization happens, all threads will see the same value from the aliased pointer.  Even though the content is a local address, it will be same for all the threads and will be uniform.

PR #198620, opened 2026-05-19
[libc] implement alarm entrypoint

kaladron:
Implemented the alarm entrypoint for unistd.h.

### DIFF
--- a/libc/config/linux/x86_64/entrypoints.txt
+++ b/libc/config/linux/x86_64/entrypoints.txt
@@ -378,6 +378,7 @@ set(TARGET_LIBC_ENTRYPOINTS
     libc.src.termios.tcsetattr
 
     # unistd.h entrypoints
+    libc.src.unistd.alarm
     libc.src.unistd.access
     libc.src.unistd.chdir
     libc.src.unistd.chown
--- a/libc/test/src/unistd/alarm_test.cpp
+++ b/libc/test/src/unistd/alarm_test.cpp
@@ -12,7 +12,9 @@
 //===----------------------------------------------------------------------===//
 
 #include "hdr/signal_macros.h"
+#include "hdr/types/struct_timespec.h"
 #include "src/signal/signal.h"
+#include "src/time/nanosleep.h"
 #include "src/unistd/alarm.h"
 #include "test/UnitTest/Test.h"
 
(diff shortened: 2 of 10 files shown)


libc/test/src/unistd/alarm_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+  // Set alarm for 1 second.
+  LIBC_NAMESPACE::alarm(1);
+
+  // Busy wait until it fires or we timeout (e.g. after ~2 seconds).
+  // 1 billion iterations in the inner loop is roughly 1-2 seconds on modern
+  // CPUs.
+  for (int i = 0; i < 20 && !alarm_fired; ++i) {
+    for (volatile int j = 0; j < 100000000; ++j)
+      ;
+  }
```
michaelrj-google: nit: intead of maxing out the CPU it would be better to just use `sleep`

libc/test/src/unistd/alarm_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+  unsigned int prev = LIBC_NAMESPACE::alarm(10);
+  // The first call to alarm should return 0 as no alarm has been set before.
+  EXPECT_EQ(prev, 0U);
+
+  // Set alarm for 5 seconds. It should return the remaining time of the
+  // previous 10-second alarm. Since very little time has passed, it should
+  // be 10 (or 9 if we are extremely slow).
+  prev = LIBC_NAMESPACE::alarm(5);
+  EXPECT_GE(prev, 9U);
+  EXPECT_LE(prev, 10U);
```
michaelrj-google: I'm not entirely sure this will be consistent on CI. The scheduling can get weird and tests can take longer than you'd expect. On the other hand since it's straightforward code it will probably be fine. If we run into flakey issues we can tweak later.

libc/test/src/unistd/alarm_test.cpp
kaladron: I wasn't sure if sleep and alarm mixed.  I'll try

libc/config/linux/x86_64/entrypoints.txt
```diff
@@ -378,6 +378,7 @@ set(TARGET_LIBC_ENTRYPOINTS
     libc.src.termios.tcsetattr
 
     # unistd.h entrypoints
+    libc.src.unistd.alarm
```
michaelrj-google, line 381: please add to aarch64 and risc-v as well

michaelrj-google (approved):
LGTM

PR #198625, opened 2026-05-19
[mlir][MemRef] Move FlattenMemRefs to interfaces

krzysz00:
Use IndexedAccessOpInterface for indexed memory operations and VectorTransferOpInterface update methods for vector transfers ops. Add an IndexedMemCopyOpInterface-based flattening pattern for DMA-like operations.

This gets rid of the hardcoded operation list in the patterns, allowing dialects like AMDGPU to transparently work with memref flattening.

As a consequence, the specialized vector- and memref-dialect only constructors that only load a subset of the flattening patterns are removed.

Also adds a max() method to SaturatedInteger because it was useful for computing the size of a flat memref.


mlir/lib/Dialect/MemRef/Transforms/FlattenMemRefs.cpp
```diff
@@ -96,126 +98,121 @@ static bool checkLayout(Value val) {
...
+  SmallVector<int64_t> sourceStrides;
+  if (failed(sourceType.getStridesAndOffset(sourceStrides, sourceOffset)))
+    return failure();
+
+  auto flatDimSize = SaturatedInteger::wrap(0);
+  for (auto [size, stride] :
+       llvm::zip_equal(sourceType.getShape(), sourceStrides)) {
+    auto dimSize =
+        SaturatedInteger::wrap(size) * SaturatedInteger::wrap(stride);
+    flatDimSize = flatDimSize.smax(dimSize);
```
lialan, line 119: Can't we just use `memref::getLinearizedMemRefOffsetAndSize` here?

mlir/lib/Dialect/MemRef/Transforms/FlattenMemRefs.cpp
```diff
@@ -293,22 +290,123 @@ struct AllocLikeFlattenPattern : public OpRewritePattern<AllocLikeOp> {
...
+    if (failed(canLinearizeAccessedShape(op, memref, rewriter)))
+      return failure();
+
+    auto [flatMemref, offset] = getFlattenMemrefAndOffset(
+        rewriter, op->getLoc(), memref, op.getIndices());
+    std::optional<SmallVector<Value>> replacementValues =
+        op.updateMemrefAndIndices(rewriter, flatMemref, ValueRange{offset});
+    if (replacementValues)
+      rewriter.replaceOp(op, *replacementValues);
+    return success();
```
lialan, line 316: return failure() when `!replacementValues`?

krzysz00: No, a lack of replacement values means the update was performed in-place

mlir/lib/Dialect/MemRef/Transforms/FlattenMemRefs.cpp
krzysz00: IIRC, this is following the existing logic, but I'll take a look tomorrow

krzysz00: Yeah, this needs to be here for no-creating-IR-before-failure reasons, it looks like

lialan (approved):
+1, no other comments from me.

PR #196737, opened 2026-05-09
clang: Refactor handling of offload sanitizer arguments

arsenm:
Previously the AMDGPU toolchains hackily handled -fsanitize arguments.
They would lie and report that all host side sanitizers are available,
then TranslateArgs would filter out the device side cases that do not
work, providing diagnostics for the skipped cases. Move that logic
into the base sanitizer argument parsing.

This makes the produced diagnostics more consistent. Previously we
would get repeated warnings when a sanitizer is fully unsupported
by amdgpu, which should now be once for the toolchain. These could
be further improved; we're printing the specific field of -fsanitize
in more cases where it could be skipped. In other cases we have the
opposite problem, where we aren't reporting the exact sanitizer
from the -f flag in the case that depends on a subtarget feature.

This will help fix other broken target specific flag forwarding bugs
in the future.

Co-authored-by: Claude Sonnet 4 <noreply@anthropic.com>

PR #196503, opened 2026-05-08
clang: Add BoundArch/OffloadKind argument to getSupportedSanitizers

arsenm:
Currently the AMDGPU HIP and OpenMP toolchains falsely report
all host sanitizers are supported, and then go out of their way
to skip forwarding those to the device compiles. Add an offloading
kind argument so that in the future this can be handled in one
place in the base toolchain.

Co-authored-by: Claude Sonnet 4 <noreply@anthropic.com>

PR #198627, opened 2026-05-19
clang/AMDGPU: Use TranslateArgs from the base toolchain instead of the host

arsenm:
This fixes -Xopenmp-target / -Xarch for arbitrary arguments. HIP and OpenMP
had cargo-cult broken implementations of TranslateArgs, which called the host
toolchain's implementation, and then special case  transferred either -march
or -mcpu to the device argument list. The respective device forwarding flags
should work for any argument, not just this one. The main feature that needs
to be preserved is the shared filtering of unsupported sanitizers to degrade
them into warnings.

Most of the changes here are dealing with fallout observed when
the host target is darwin. The darwin toolchain happens to have
some hacky statefulness tracking the compile target version, which
gets written and rewritten on argument parsing. To maintain this hack,
there are a few unused calls to getArgsForToolChain; start passing OFK_Host
to these so the offload toolchains don't get confused and think they're in
a non-offload context.

### DIFF
--- a/clang/lib/Driver/ToolChains/AMDGPU.cpp
+++ b/clang/lib/Driver/ToolChains/AMDGPU.cpp
@@ -714,23 +714,24 @@ Tool *AMDGPUToolChain::buildLinker() const {
 DerivedArgList *
 AMDGPUToolChain::TranslateArgs(const DerivedArgList &Args, StringRef BoundArch,
                                Action::OffloadKind DeviceOffloadKind) const {
-
   DerivedArgList *DAL =
       Generic_ELF::TranslateArgs(Args, BoundArch, DeviceOffloadKind);
+  if (!DAL) {
+    DAL = new DerivedArgList(Args.getBaseArgs());
+    for (Arg *A : Args)
+      DAL->append(A);
+  }
 
   const OptTable &Opts = getDriver().getOpts();
 
-  if (!DAL)
-    DAL = new DerivedArgList(Args.getBaseArgs());
-
-  for (Arg *A : Args)
-    DAL->append(A);
-
-  // AMDGPU is intended to use `-mcpu` but we accept `-march` for legacy.
-  if (Arg *A = DAL->getLastArg(options::OPT_march_EQ)) {
-    DAL->eraseArg(options::OPT_march_EQ);
-    if (!DAL->hasArg(options::OPT_mcpu_EQ))
-      DAL->AddJoinedArg(A, Opts.getOption(options::OPT_mcpu_EQ), A->getValue());
+  if (DeviceOffloadKind == Action::OFK_None) {
+    // AMDGPU is intended to use `-mcpu` but we accept `-march` for legacy.
+    if (Arg *A = DAL->getLastArg(options::OPT_march_EQ)) {
+      DAL->eraseArg(options::OPT_march_EQ);
+      if (!DAL->hasArg(options::OPT_mcpu_EQ))
+        DAL->AddJoinedArg(A, Opts.getOption(options::OPT_mcpu_EQ),
+                          A->getValue());
+    }
   }
 
   // Replace -mcpu=native with detected GPU.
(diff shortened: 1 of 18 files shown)


arsenm:
* **#198627** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/198627?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/198627?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#196737** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/196737?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#196503** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/196503?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/llvm?ref=merge-queue-instructions-in-stack-comment&prId=198627).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

clang/lib/Driver/ToolChains/AMDGPU.cpp
```diff
@@ -717,17 +717,15 @@ Tool *AMDGPUToolChain::buildLinker() const {
...
-
   DerivedArgList *DAL =
       Generic_ELF::TranslateArgs(Args, BoundArch, DeviceOffloadKind);
-
-  const OptTable &Opts = getDriver().getOpts();
-
-  if (!DAL)
+  if (!DAL) {
     DAL = new DerivedArgList(Args.getBaseArgs());
+    for (Arg *A : Args)
```
jhuber6, line 721: Was there an issue when we did this unconditionally?

clang/lib/Driver/ToolChains/Gnu.cpp
```diff
@@ -3485,12 +3485,19 @@ Generic_GCC::TranslateArgs(const llvm::opt::DerivedArgList &Args,
 
   // Add the bound architecture to the arguments list if present.
   if (!BoundArch.empty()) {
-    options::ID Opt =
-        getTriple().isARM() || getTriple().isPPC() || getTriple().isAArch64()
-            ? options::OPT_mcpu_EQ
-            : options::OPT_march_EQ;
-    DAL->eraseArg(Opt);
-    DAL->AddJoinedArg(nullptr, Opts.getOption(Opt), BoundArch);
+    if (getTriple().isAMDGPU()) {
```
jhuber6: Why does this need to be split on AMDGPU?

jhuber6 (commented):
Definitely in favor of cleaning this up

clang/lib/Driver/ToolChains/AMDGPU.cpp
arsenm: The base implementation is doing the same thing, so this was duplicating all the flags but I don't remember what the symptom was

clang/lib/Driver/ToolChains/Gnu.cpp
arsenm: The other targets were choosing -mcpu or -march. This is handling both, so

jhuber6: I really hate the fact that we use `-march` on AMDGPU for OpenMP. Do you think there's a way we could just rewrite that early on so we don't need to pretend like it every works? Alternatively, we just make it an error and make people update their compiler flags. It's been at least like 3 or 4 years since we started telling people to use `--offload-arch` instead, and `-march` only exists for the old OpenMP path.

arsenm: IIRC -march is the preferred one for x86. I don't see why we can't just handle both. Not sure where it would be replaced

jhuber6: We should only need to check `-mcpu` now. AMDGPU accepting `-march` seems to have mostly snuck in at some point and no one noticed, but I at least tried to make it consistent internally.

arsenm: As far as naming -march kind of makes more sense for a GPU

jhuber6 (approved):

PR #198633, opened 2026-05-19
[mlir][SPIR-V] Fold IAddCarry/[SU]MulExtended constants into spirv.Constant struct

aobolensk:
Emit a single spirv.Constant of struct type instead of the spirv.Undef + two spirv.CompositeInsert workaround


IgWod (approved):
LGTM

PR #198635, opened 2026-05-19
[flang] Improved FIR AA for acc.compute_region.

vzakhari:
We should actually treat the operand<->block-arg association
in `acc.compute_region` as a pass-through link, so that we can
accumulate the whole chain of addressing inside the compute region.
This is especially important for acc routines, where all `[hl]fir.declare`
operations are located inside `acc.compute_region`. The input operand,
in this case, is not a result of ACC entry/init operation.

The change in `FIROps.cpp` should be NFC, currently. I made it
just to align the pass-through handling with the AA change.
Even though I effectively replaced `ACC_DATA_ENTRY_OPS` with
`ACC_DATA_ENTRY_AND_INIT_OPS`, this change is not testable, because
boxed reductions are currently unsupported.


razvanlupusoru (approved):
Thank you.

PR #198637, opened 2026-05-19
[mlir][SPIR-V] Add ISubBorrow canonicalization patterns

aobolensk:
Mirror the IAddCarry folder, rewrite isubborrow(x, 0) to <x, 0> via CompositeConstruct, and fold the all-constant case into a single spirv.Constant struct


IgWod:
I think if you merge https://github.com/llvm/llvm-project/pull/198633 first, you should be able to template the code to use it for both add and sub.

aobolensk:
> I think if you merge #198633 first, you should be able to template the code to use it for both add and sub.

Thanks for this note, yes, indeed, applied

IgWod (approved):
Thanks!

PR #196989, opened 2026-05-11
Reapply [Clang] Implement P2843R3 - Preprocessing is never undefined

yronglin:
This PR reapply https://github.com/llvm/llvm-project/pull/192073, and make most `-pedantic` warnings default on.

---
This PR marks [P2843R3 - Preprocessing is never undefined](https://wg21.link/P2843) as implemented and add tests.

Fixes https://github.com/llvm/llvm-project/issues/145658

PR #198641, opened 2026-05-19
Revert "Reapply [Clang] Implement P2843R3 - Preprocessing is never undefined"

petrhosek:
Reverts llvm/llvm-project#196989 since it broke LLVM libc, see https://github.com/llvm/llvm-project/pull/196989#issuecomment-4490816018 for details.


michaelrj-google (approved):

jhuber6:
No response from the authors and my build has been stuck so I'm going to go ahead and merge this. Long term I think we should be able to move away from redefining these macros at least in Clang, they're always available in C++ mode.

PR #183851, opened 2026-02-27
[SelectionDAG] Add ISD::ABS_MIN_POISON to preserve poision semantics of llvm.abs

aryanmagoon:
SelectionDAGBuilder previously dropped the is_int_min_poison flag on llvm.abs, lowering both variants to ISD::ABS. This is unsound for certain targets like NVPTX whose native abs.s is poison on INT_MIN. This PR adds a new ISD::ABS_MIN_POISON opcode, emits it for llvm.abs(x, true) and threads through legalization, SDAG folding. The default action for this is Expand with a fallback to the original ISD::ABS. DAGCombiner adds visitABS_MIN_POISON which mirrors visitABS and two new folds: abs_min_poison(freeze(abs x)) -> freeze(abs x) and abs_min_poison(sign_extend_inreg x) → zext(abs(trunc x)). PromoteIntRes, ExpandIntRes, and widenAbs now emit ABS_MIN_POISON whenever the input provably can't be the wide INT_MIN. NVPTX fix to match abs.s against abs_min_poison and tests updated accordingly.

PR #198646, opened 2026-05-19
[NVPTX] Auto-upgrade llvm.nvvm.abs with int-min poison

aryanmagoon:
Update the auto-upgrade for llvm.nvvm.abs.i and llvm.nvvm.abs.ll to use the generic llvm.abs intrinsic with is_int_min_poison=true. The previous expansion used neg/icmp/select which gives defined INT_MIN -> INT_MIN behavior, but loses the poison/undefined signed-min semantics needed for NVPTX to select PTX abs.s32 and abs.s64 instrucitons when the source operation permits signed-min overflow to be undefined. This is a followup to #183851 . Using llvm.abs(..., true) preserves intended IR semantics and lowers through the new ABS_MIN_POISON. We also update the tests and add NVPTX CodeGen coverage for the legacy nvvm abs intrinsics.


aryanmagoon:
@RKSimon @topperc Since you are both familiar with the initial change #183851, would one of you be willing to review this one as well? Separately, the bot-reported failures look unrelated to this PR. I cannot reproduce these failures locally. Thanks!

llvm/test/CodeGen/NVPTX/nvvm-abs.ll
```diff
@@ -0,0 +1,24 @@
...
+
+define i64 @test_nvvm_abs_ll(i64 %x) {
+; CHECK-LABEL: test_nvvm_abs_ll(
+; CHECK-NOT: max.s64
+; CHECK: abs.s64
+; CHECK-NOT: max.s64
+; CHECK: ret;
+  %r = call i64 @llvm.nvvm.abs.ll(i64 %x)
+  ret i64 %r
+}
```
RKSimon: (style) newline at EOF

RKSimon (approved):
LGTM

PR #198650, opened 2026-05-19
[NFC][ASan] Factor out ASan call insertion behind a single call

etsal:
The ASan pass directly injects function calls into the IR using getOrInsertFunction() on every call site. Refactor the disparate call sites behind an AsanFunctionInserter class. This allows us to add pre- and post-processing logic for all inserted functions at once.


etsal:
CC @vitalybuka

vitalybuka (approved):
LGTM

llvm/lib/Transforms/Instrumentation/AddressSanitizer.cpp
```diff
@@ -712,6 +713,22 @@ static Twine genName(StringRef suffix) {
...
+class AsanFunctionInserter {
+public:
+  AsanFunctionInserter(Module &M) : M(M) {}
+
+  template <typename... ArgTypes>
+  FunctionCallee insertFunction(StringRef Name, ArgTypes &&...Args) {
+    return M.getOrInsertFunction(Name, std::forward<ArgTypes>(Args)...);
+  }
+
+  Module &M;
```
vitalybuka, line 728: M should be private

llvm/lib/Transforms/Instrumentation/AddressSanitizer.cpp
```diff
@@ -783,13 +800,14 @@ class RuntimeCallInserter {
...
-  AddressSanitizer(Module &M, const StackSafetyGlobalInfo *SSGI,
+  AddressSanitizer(Module &M, AsanFunctionInserter &Inserter,
+                   const StackSafetyGlobalInfo *SSGI,
                    int InstrumentationWithCallsThreshold,
                    uint32_t MaxInlinePoisoningSize, bool CompileKernel = false,
                    bool Recover = false, bool UseAfterScope = false,
                    AsanDetectStackUseAfterReturnMode UseAfterReturn =
                        AsanDetectStackUseAfterReturnMode::Runtime)
-      : M(M),
+      : M(M), Inserter(Inserter),
```
vitalybuka: do you need to pass them from outside? they are cheep, maybe create/destroy as needed

vitalybuka (approved):

llvm/lib/Transforms/Instrumentation/AddressSanitizer.cpp
etsal: Makes sense, I added an Inserter for AddressSanitizer and ModuleAddressSanitizer.

llvm/lib/Transforms/Instrumentation/AddressSanitizer.cpp
etsal: Makes sense.

etsal:
By the way, I think the CI failures are unrelated to the diff.

vitalybuka (approved):

llvm/lib/Transforms/Instrumentation/AddressSanitizer.cpp
vitalybuka: Maybe even function scope

vitalybuka:
Looks like unrelated CI brackage prevented automerge
https://github.com/llvm/llvm-project/issues/198776

I will force push if it fails again.

PR #198651, opened 2026-05-19
[X86][AtomicExpand] Remove X86's shouldCastAtomicLoadInIR override

jofrn:
(added in #148899)
So that atomic floating-point and FP-vector loads are no longer bitcast to an integer at the IR level by AtomicExpand.


jofrn:
Continued in https://github.com/llvm/llvm-project/pull/199520.

PR #198653, opened 2026-05-19
[AArch64] Use fp16 FNEG and FABS for bf16.

davemgreen:
These operations are bitwise and can be used for bf16 fneg and fabs as equally as they can be used for fp16.


c-rhodes (approved):

MacDue (approved):
LGTM

PR #198654, opened 2026-05-19
[libc] Update prctl() declaration to use variadic arguments.

vonosmas:
prctl declaration should typically use variadic arguments (e.g. see https://man7.org/linux/man-pages/man2/prctl.2.html), as the types / quantity of subsequent arguments depends on the `option`. We can't depend on all `<prctl.h>` users to explicitly cast arguments to `unsigned long` and passing all 5 of them every time.

* Don't add any option-specific logic, and just consume `arg2`-`arg5` from variadic arguments and pass them to syscall implementation as-is, assuming that they won't be used by the kernel if they are not needed, and consuming these arguments won't lead to crashes.
* Updated the test to use `prctl` variants with less than 5 explicit arguments (for PR_SET_NAME and PR_GET_NAME).

### DIFF
--- a/libc/include/sys/prctl.yaml
+++ b/libc/include/sys/prctl.yaml
@@ -11,7 +11,4 @@ functions:
     return_type: int
     arguments:
       - type: int
-      - type: unsigned long
-      - type: unsigned long
-      - type: unsigned long
-      - type: unsigned long
+      - type: '...'
--- a/libc/src/sys/prctl/linux/prctl.cpp
+++ b/libc/src/sys/prctl/linux/prctl.cpp
@@ -12,13 +12,21 @@
 
 #include "src/__support/libc_errno.h"
 #include "src/__support/macros/config.h"
+#include <stdarg.h>
 #include <sys/syscall.h> // For syscall numbers.
 
 namespace LIBC_NAMESPACE_DECL {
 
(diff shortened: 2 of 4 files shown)


michaelrj-google (commented):
IIUC this change should be fine, though it may cause false positives on msan due to sometimes reading uninitialized arguments. @kaladron do you have any experience with this?

vonosmas:
MSan doesn't produce errors for reading uninitialized memory, only when this memory is _used_ somewhere (e.g. in a n a branch, or as an argument to standard library function call).

kaladron:
> IIUC this change should be fine, though it may cause false positives on msan due to sometimes reading uninitialized arguments. @kaladron do you have any experience with this?

I actually don't.  I've used ASan a lot, but rarely MSan.

vonosmas:
Let's decide if we're OK with the approach of potentially reading more va_args than the user provided. If needed, we might add sanitizer annotations (but I don't think they'd be needed - see above).

michaelrj-google (approved):
Given that we're passing it to a syscall I'd guess that msan would see that as using the value, but it seems like the correct solution would be to annotate for msan.

I checked how bionic implements this and it seems like they make the prototype variadic but always pass 5 arguments to the kernel.

PR #198655, opened 2026-05-19
[AArch64][GlobalISel] Add BF16 fabs and fneg

davemgreen:
These should be very simple as they are just legal or expanded based on whether
fullfp16 is available, as the FP16 FNEG and FABS instructions can be used
equally for BF16.

### DIFF
--- a/llvm/test/CodeGen/AArch64/bf16-instructions.ll
+++ b/llvm/test/CodeGen/AArch64/bf16-instructions.ll
@@ -33,8 +33,6 @@
 ; CHECK-CVT-GI-NEXT:  warning: Instruction selection used fallback path for test_log
 ; CHECK-CVT-GI-NEXT:  warning: Instruction selection used fallback path for test_log10
 ; CHECK-CVT-GI-NEXT:  warning: Instruction selection used fallback path for test_log2
-; CHECK-CVT-GI-NEXT:  warning: Instruction selection used fallback path for test_fabs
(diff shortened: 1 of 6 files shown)


c-rhodes (approved):

llvm/test/CodeGen/AArch64/bf16-instructions.ll
```diff
@@ -2614,14 +2610,41 @@ define bfloat @test_fabs(bfloat %a) #0 {
...
+; CHECK-CVT-GI-NEXT:    eor w8, w8, #0xffff8000
+; CHECK-CVT-GI-NEXT:    fmov s0, w8
+; CHECK-CVT-GI-NEXT:    // kill: def $h0 killed $h0 killed $s0
+; CHECK-CVT-GI-NEXT:    ret
+;
+; CHECK-BF16-GI-LABEL: test_fneg:
+; CHECK-BF16-GI:       // %bb.0:
+; CHECK-BF16-GI-NEXT:    // kill: def $h0 killed $h0 def $s0
+; CHECK-BF16-GI-NEXT:    fmov w8, s0
+; CHECK-BF16-GI-NEXT:    eor w8, w8, #0xffff8000
```
aemerson, line 2644: What's causing the differing masks to SDAG?

davemgreen: It looks like a difference between how anyext constants are handled when they are promoted.

PR #198656, opened 2026-05-19
[clang-format] Fix unrecognized qualifiers breaking east-const

Cons-Cat:
Long time reader, first time contributor. :3

I've been experimenting with `_Nullable` and `_Nonnull` qualifiers in my project for which the `.clang-format` configures:
```
QualifierAlignment: Custom
QualifierOrder: [type, const, volatile]
```

This results in the following form:
```cpp
T const* _Nullable const identifier =
```
Formatting as:
```cpp
T const* _Nullable identifier const =
```
Which naturally doesn't compile. This is semi-blocking my project, so I would like to contribute it reasonably quickly if possible. I'll try to respond to feedback promptly.

This is possible to fix in a targeted way within the format rule, but I believe the underlying issue is these qualifiers need to hold true under `isQualifier`. That seems to fix my issue and may fix others I haven't discovered. I've extended this to other qualifiers I know of, but not `__ptrauth` family which seems to have its own special handling.

### DIFF
--- a/clang/lib/Format/QualifierAlignmentFixer.cpp
+++ b/clang/lib/Format/QualifierAlignmentFixer.cpp
@@ -169,6 +169,12 @@ static bool isQualifier(const FormatToken *const Tok) {
   case tok::kw_constexpr:
   case tok::kw_restrict:
   case tok::kw_friend:
+  case tok::kw__Nonnull:
+  case tok::kw__Nullable:
+  case tok::kw__Null_unspecified:
+  case tok::kw___ptr32:
+  case tok::kw___ptr64:
+  case tok::kw___funcref:
     return true;
   default:
     return false;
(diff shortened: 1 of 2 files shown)


mydeveloperday (approved):
Looks good to me, thank you.

clang/docs/ReleaseNotes.rst
```diff
@@ -790,6 +790,11 @@ clang-format
   enum assignments without affecting other assignments.
 - Add ``BreakBeforeReturnType`` option to break before the function return
   type.
+- Fix ``QualifierAlignment`` rotating an east-side ``const`` / ``volatile`` past
```
HazardyKnusperkeks: We don't add fixes to the release notes.

Cons-Cat: I've removed that note now.

HazardyKnusperkeks (approved):

PR #198658, opened 2026-05-19
[AggressiveInstCombine] Recognizing tail truncation in the popcount pattern

mshockwave:
We're currently able to recognize the following popcount pattern
```
int popcnt(unsigned x) {
 x = x - ((x >> 1) & 0x55555555);
 x = x - 3*((x >> 2) & 0x33333333);
 x = (x + (x >> 4)) & 0x0F0F0F0F;
 x = x + (x >> 8);
 x = x + (x >> 16);
 return x & 0x0000003F;
}
```
but if a truncation follows right after the last AND instruction:
```
int16_t popcnt(unsigned x) {
 x = x - ((x >> 1) & 0x55555555);
 x = x - 3*((x >> 2) & 0x33333333);
 x = (x + (x >> 4)) & 0x0F0F0F0F;
 x = x + (x >> 8);
 x = x + (x >> 16);
 return int16_t(x & 0x0000003F);
}
```
since InstCombine canonicalizes `(trunc (and y, C))` into `(and trunc(y), C')`, we might loose the opportunity to turn the above snippet into `(trunc (popcount x))` as there is a `trunc` interrupting the pattern matching.

This patch fixes this issue by considering this extra `trunc` during pattern matching, and appending it in the final popcount result, if there is any.

### DIFF
--- a/llvm/lib/Transforms/AggressiveInstCombine/AggressiveInstCombine.cpp
+++ b/llvm/lib/Transforms/AggressiveInstCombine/AggressiveInstCombine.cpp
@@ -487,12 +487,17 @@ static bool foldAnyOrAllBitsSet(Instruction &I) {
 }
 
 /// Helper function to replace an instruction with a popcount intrinsic.
-/// This creates the ctpop intrinsic and replaces all uses of the instruction.
-static void replaceWithPopCount(Instruction &I, Value *Root) {
+/// This creates the ctpop intrinsic with an optional truncation appended at the
+/// end, and replaces all uses of the instruction.
+static void replaceWithPopCount(Instruction &I, Value *Root,
+                                Value *Trunc = nullptr) {
   LLVM_DEBUG(dbgs() << "Recognized popcount intrinsic\n");
   IRBuilder<> Builder(&I);
-  I.replaceAllUsesWith(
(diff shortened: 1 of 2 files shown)


llvm/lib/Transforms/AggressiveInstCombine/AggressiveInstCombine.cpp
```diff
@@ -487,12 +487,17 @@ static bool foldAnyOrAllBitsSet(Instruction &I) {
...
+/// end, and replaces all uses of the instruction.
+static void replaceWithPopCount(Instruction &I, Value *Root,
+                                Value *Trunc = nullptr) {
   LLVM_DEBUG(dbgs() << "Recognized popcount intrinsic\n");
   IRBuilder<> Builder(&I);
-  I.replaceAllUsesWith(
-      Builder.CreateIntrinsic(Intrinsic::ctpop, I.getType(), {Root}));
+  Value *NewVal =
+      Builder.CreateIntrinsic(Intrinsic::ctpop, Root->getType(), {Root});
+  if (Trunc)
```
topperc: Can't we use I.getType() to figure out if a trunc is needed?

mshockwave: that's true. It's now updated

topperc (approved):
LGTM

PR #198659, opened 2026-05-19
[libc] Implement towupper and towlower entrypoints

michaelrj-google:
Following up on #187670 to add public entrypoints for the wctype
conversion functions.


vonosmas (approved):

bassiounix (changes requested):
`internal::tolower` doesn't use the UTF conversion table!

michaelrj-google:
> `internal::tolower` doesn't use the UTF conversion table!

You're right, I missed that. Fixed in the new version. I had to tweak the setup a bit to get it working properly. Looking forward to when we can get rid of `LIBC_ENABLE_CONSTEXPR`.

bassiounix (approved):

PR #198660, opened 2026-05-19
[arm64e][cfi] .cfi_b_key_frame is irrelevant for Mach-O platforms

jroelofs:
We always sign with the B-Key anyway, and the unwinder behaves the same whether the directive is present. Better to avoid emitting it in the first place.


llvm/lib/Target/AArch64/AArch64PointerAuth.cpp
```diff
@@ -115,8 +115,7 @@ void AArch64PointerAuthImpl::signLR(MachineFunction &MF,
   if (MFnI.branchProtectionPAuthLR() && Subtarget->hasPAuthLR()) {
```
ojhunt, line 115: Well this condition confuses me - obviously not relevant to this PR but this condition seems to imply systems that have ptrauth branch protection but not ptrauth.

llvm/lib/Target/AArch64/AArch64PointerAuth.cpp
```diff
@@ -98,7 +98,7 @@ void AArch64PointerAuthImpl::signLR(MachineFunction &MF,
   // Debug location must be unknown, see AArch64FrameLowering::emitPrologue.
   DebugLoc DL;
 
-  if (UseBKey) {
+  if (UseBKey && !MF.getTarget().getTargetTriple().isOSBinFormatMachO()) {
```
ojhunt, line 101: I'd rather this be handled by a method, `std::optional<decltype(AArch64::EMITBKEY)> getKeyFrameThing()`, and then

```cpp
if (auto KeyFrameThing = getKeyFrameThing())
  BuildMI(MBB, MBBI, DL, *KeyFrameThing).setMIFlag(MachineInstr::FrameSetup);
```

Then `getKeyFrameThing()` would the macho guard, and it's super obvious what this code is, and what the macho exclusion is.

llvm/lib/Target/AArch64/AArch64PointerAuth.cpp
```diff
@@ -115,8 +115,7 @@ void AArch64PointerAuthImpl::signLR(MachineFunction &MF,
   if (MFnI.branchProtectionPAuthLR() && Subtarget->hasPAuthLR()) {
     emitPACCFI(MBB, MBBI, MachineInstr::FrameSetup, EmitCFI);
     BuildMI(MBB, MBBI, DL,
-            TII->get(MFnI.shouldSignWithBKey() ? AArch64::PACIBSPPC
-                                               : AArch64::PACIASPPC))
+            TII->get(UseBKey ? AArch64::PACIBSPPC : AArch64::PACIASPPC))
```
ojhunt, line 118: Just work out the key once: `decltype(AArch64::PACIASPPC) LRKey = UseBKey ? AArch64::PACIBSPPC : AArch64::PACIASPPC;`

llvm/test/CodeGen/AArch64/arm64e-tail-call-autib.ll
```diff
@@ -57,7 +56,6 @@ define swifttailcc void @test_async_tail_call(ptr swiftasync %ctx) "ptrauth-retu
 define swifttailcc void @test_no_fpdiff_tail_call(ptr swiftasync %ctx) "ptrauth-returns" "ptrauth-auth-traps" "sign-return-address"="all" "frame-pointer"="all" {
 ; CHECK-LABEL: test_no_fpdiff_tail_call:
 ; CHECK:       ; %bb.0:
-; CHECK-NEXT:    .cfi_b_key_frame
```
ojhunt, line 60: do we want to do a `-NOT` here instead? e.g. do we want to actually guarantee it's not here?

ojhunt (changes requested):
A few changes - I'm using `decltype(X)` as a stand in for whatever the types should be, and the auto's should be `std::optional<Y>`

llvm/lib/Target/AArch64/AArch64PointerAuth.cpp
jroelofs: yep, those exist via the hint space.

llvm/test/CodeGen/AArch64/arm64e-tail-call-autib.ll
jroelofs: CHECK-NEXT being something else guarantees that.

llvm/lib/Target/AArch64/AArch64PointerAuth.cpp
jroelofs: no. they're.... not the same between the branches of this if-else chain. that is not appropriate.

llvm/lib/Target/AArch64/AArch64PointerAuth.cpp
ojhunt: Discussed out of band, and the complexity seemed unnecessary

llvm/lib/Target/AArch64/AArch64PointerAuth.cpp
ojhunt: O_o

ojhunt: (oh and this was meant as a normal comment, not a review one)

llvm/test/CodeGen/AArch64/arm64e-tail-call-autib.ll
ojhunt: Oh yeah, durrrrr

ojhunt (approved):
Discussed offline with Jon, the addition of a method for a single check seemed like overkill

PR #198664, opened 2026-05-19
[clang-cl] Add new option `/pathmap:<from>=<to>` to replace the path prefix <from> with <to>.

vvereschaka:
This option matches MSVC options and does the path substitution for the file
references in the preprocessor macros, debug and coverage information.

This option acts as a clang's ``-ffile-prefix-map=value`` and with some known
differences in behaviour with original CL's option that do not affect 
the functionality:
 * nomalizes the macro prefix map pathes -- removes `./` and uses the target's 
   platform-specific path separator character when exanding the preprocessor
   macros -- ``-ffile-reproducible`` (but not the debug and coverage prefix maps).
 * does not require ``/experimental:deterministic`` as by MSVC. It needed for
   removing a hostname from a mangling hash gen, but clang-cl does not use a hostname 
   when generates the hashes.

Known issues:
  * does not remap the pathes within PCH/PCM files.

### DIFF
--- a/clang/test/CodeGenCXX/cl-pathmap.cpp
+++ b/clang/test/CodeGenCXX/cl-pathmap.cpp
@@ -1,5 +1,4 @@
-// RUN: %clang_cl /clang:-emit-llvm /pathmap:%p=x:/path-to/ /Fo:%t.bc /c %s
-// RUN: llvm-dis %t.bc -o - | FileCheck %s --check-prefix=CHECK-PREFIX-MAP 
+// RUN: %clang_cl /clang:-emit-llvm /pathmap:%p=x:/path-to/ -clang:-S -clang:-o- -- %s  2>&1 | FileCheck %s --check-prefix=CHECK-PREFIX-MAP 
 
 // CHECK-PREFIX-MAP: c"x:\\path-to\\cl-pathmap.cpp\00"
 
--- a/clang/test/CodeGenCXX/cl-pathmap.cpp
+++ b/clang/test/CodeGenCXX/cl-pathmap.cpp
@@ -1,0 +1,19 @@
+// RUN: %clang_cl /clang:-emit-llvm /pathmap:%p=x:/path-to/ /Fo:%t.bc /c %s
+// RUN: llvm-dis %t.bc -o - | FileCheck %s --check-prefix=CHECK-PREFIX-MAP 
+
+// CHECK-PREFIX-MAP: c"x:\\path-to\\cl-pathmap.cpp\00"
+
+namespace std {
+class source_location {
+public:
+struct __impl {
+  const char *_M_file_name;
+  const char *_M_function_name;
+  unsigned _M_line;
+  unsigned _M_column;
+};
+};
+} // namespace std
+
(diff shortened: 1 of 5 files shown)


clang/test/Preprocessor/cl-pathmap.c
```diff
@@ -0,0 +1,10 @@
+// RUN: %clang_cl -E /pathmap:%p=x:/path-to/ %s | FileCheck %s --check-prefix CHECK-REPRODUCABLE
+// CHECK-REPRODUCABLE: filename: "x:\\path-to\\cl-pathmap.c"
+// CHECK-REPRODUCABLE-NOT: filename:
+// CHECK-REPRODUCABLE-NOT: filename: "x:/path-to/cl-pathmap.c"
```
dpaoliello: Redundant with the CHECK line above: if there is no `filename:` there can't be this line.

clang/test/CodeGenCXX/cl-pathmap.cpp
```diff
@@ -0,0 +1,19 @@
+// RUN: %clang_cl /clang:-emit-llvm /pathmap:%p=x:/path-to/ /Fo:%t.bc /c %s
+// RUN: llvm-dis %t.bc -o - | FileCheck %s --check-prefix=CHECK-PREFIX-MAP
```
dpaoliello: Do we really need a full bitcode compilation and disassembly for this? Wouldn't appear in the IR?

dpaoliello (changes requested):

clang/test/Preprocessor/cl-pathmap.c
vvereschaka: got it, thank you. Removed redundant CHECK line.

clang/test/CodeGenCXX/cl-pathmap.cpp
vvereschaka: Unfortunately, `clang-cl` does not forward IR output to the console in a proper way, when `lit` can catch and process it. I.e `/Fo:-` creates a file with a `-` name, `/Fo:con` dumps to console, but `lit` does not see it, '-o -' is an unsupported option for `clang-cl`.

dpaoliello: You can do `-clang:-o-`

https://github.com/llvm/llvm-project/blob/3b89e77151214ec62546f7437b4d5e07b1c2ce1d/clang/test/CodeGenCXX/microsoft-abi-eh-disabled.cpp#L2-L5

vvereschaka: `-clang:-o-` works fine, thank you.
I tried `-clang:-o -` before, but it works only without the space symbol in between.
Updated test accordingly.

dpaoliello (approved):

vvereschaka:
got `cl-pathmap` tests failures on [llvm-clang-aarch64-darwin](https://lab.llvm.org/buildbot/#/builders/190)
https://lab.llvm.org/buildbot/#/builders/190/builds/43598
Reverting.

PR #198666, opened 2026-05-19
[NFC] Remove fractional part of Estimated cost per lane in memory-interleave.ll

dicksonchau:
On the memory-interleave.ll test, some of the CHECK lines are failing on z/OS, due to difference in rounding behaviour when printing the Estimated cost per lane. Resolve this by removing the fractional part, similar to what done in the past with https://github.com/llvm/llvm-project/commit/e8556ff6b664df6e595f8aed175eff3a27a4a020 and https://github.com/llvm/llvm-project/commit/aeb88f6778756ea889918308241a2b34bd7f64e2 .


fhahn (approved):
LGTM, thanks.

david-arm (approved):
LGTM!

PR #198667, opened 2026-05-19
[Clang] Emit prefix map normalization before generating hashes for the unique linkage names.

vvereschaka:
Use normalized path from the macro prefix map to generate the unique ids for the internal linkage names. That allows a reproducible hash on any build system. Regularly the macro prefix map gets normalized in favor of the target system before the path substitution.

### DIFF
--- a/clang/docs/ReleaseNotes.rst
+++ b/clang/docs/ReleaseNotes.rst
@@ -320,6 +320,11 @@ Modified Compiler Flags
 - The `-mno-outline` flag will now add the `nooutline` IR attribute, so that
   `-mno-outline` and `-moutline` objects can be mixed correctly during LTO.
 
+- Slightly changed hash id generation to get the unique linkage symbols names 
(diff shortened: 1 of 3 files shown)


ojhunt (commented):
This looks good to me, and I honestly am a little concerned about the correctness of the old for-loop/rfind logic so this seems like a solid improvement (the preprocessor logic does this with the path APIs which do this consistently correctly).

ojhunt (approved):
*and I managed to clear my own review

AaronBallman (approved):
LGTM but should this come with a release note so users know about the fix?

vvereschaka:
@AaronBallman, @ojhunt thank you for review.

>should this come with a release note so users know about the fix?

I've updated ReleaseNotes.rst with notes for  ``-unique-internal-linkage-names`` option.

AaronBallman (approved):
LG, thanks for the release note

PR #198668, opened 2026-05-19
[clang][Sema] Make passing a non-POD type to `va_arg` an error, not a warning

### LINKED ISSUE #107836: Clang crashes on incorrect code with -std=c++20
Reproducer: https://gcc.godbolt.org/z/P7b5M9WT9 (assertions trunk）
```
#include <iostream>
#include <cstdarg>
int read_int(std::string prompt,...) {
    using va_list = std::va_list;
    va_list args;
    va_start(args, prompt);
    auto value = std::stoi(va_arg(args, std::string));
    return value;
}
```
get result
```
<source>:7:41: error: second argument to 'va_arg' is of non-POD type 'std::string' (aka 'basic_string<char>') [-Wnon-pod-varargs]
    7 |     auto value = std::stoi(va_arg(args, std::string));
      |                                         ^~~~~~~~~~~
/opt/compiler-explorer/clang-assertions-trunk-20240909/lib/clang/20/include/__stdarg_va_arg.h:20:47: note: expanded from macro 'va_arg'
   20 | #define va_arg(ap, type) __builtin_va_arg(ap, type)
      |                                               ^~~~
clang++: /root/llvm-project/clang/lib/CodeGen/CGExprAgg.cpp:2085: void clang::CodeGen::CodeGenFunction::EmitAggregateCopy(clang::CodeGen::LValue, clang::CodeGen::LValue, clang::QualType, clang::CodeGen::AggValueSlot::Overlap_t, bool): Assertion `(Record->hasTrivialCopyConstructor() || Record->hasTrivialCopyAssignment() || Record->hasTrivialMoveConstructor() || Record->hasTrivialMoveAssignment() || Record->hasAttr<TrivialABIAttr>() || Record->isUnion()) && "Trying to aggregate-copy a type without a trivial copy/move " "constructor or assignment operator"' failed.
Stack dump:
0.	Program arguments: /opt/compiler-explorer/clang-assertions-trunk/bin/clang++ -gdwarf-4 -g -o /app/output.s -mllvm --x86-asm-syntax=intel -fno-verbose-asm -S --gcc-toolchain=/opt/compiler-explorer/gcc-snapshot -fcolor-diagnostics -fno-crash-diagnostics --std=c++20 <source>
1.	<eof> parser at end of file
2.	<source>:3:5: LLVM IR generation of declaration 'read_int'
3.	<source>:3:5: Generating code for declaration 'read_int'
 #0 0x0000000003b258d8 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3b258d8)
 #1 0x0000000003b2359c llvm::sys::CleanupOnSignal(unsigned long) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3b2359c)
 #2 0x0000000003a6c198 CrashRecoverySignalHandler(int) CrashRecoveryContext.cpp:0:0
 #3 0x0000754bbd642520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #4 0x0000754bbd6969fc pthread_kill (/lib/x86_64-linux-gnu/libc.so.6+0x969fc)
 #5 0x0000754bbd642476 gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x42476)
 #6 0x0000754bbd6287f3 abort (/lib/x86_64-linux-gnu/libc.so.6+0x287f3)
 #7 0x0000754bbd62871b (/lib/x86_64-linux-gnu/libc.so.6+0x2871b)
 #8 0x0000754bbd639e96 (/lib/x86_64-linux-gnu/libc.so.6+0x39e96)
 #9 0x00000000042fd5a7 clang::CodeGen::CodeGenFunction::EmitAggregateCopy(clang::CodeGen::LValue, clang::CodeGen::LValue, clang::QualType, clang::CodeGen::AggValueSlot::Overlap_t, bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42fd5a7)
#10 0x00000000042fda71 (anonymous namespace)::AggExprEmitter::EmitCopy(clang::QualType, clang::CodeGen::AggValueSlot const&, clang::CodeGen::AggValueSlot const&) (.isra.0) CGExprAgg.cpp:0:0
#11 0x00000000042fdcfd (anonymous namespace)::AggExprEmitter::EmitFinalDestCopy(clang::QualType, clang::CodeGen::LValue const&, clang::CodeGen::CodeGenFunction::ExprValueKind) CGExprAgg.cpp:0:0
#12 0x00000000042fe7e3 clang::CodeGen::CodeGenFunction::EmitAggFinalDestCopy(clang::QualType, clang::CodeGen::AggValueSlot, clang::CodeGen::LValue const&, clang::CodeGen::CodeGenFunction::ExprValueKind) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42fe7e3)
#13 0x00000000042da680 clang::CodeGen::CodeGenFunction::EmitLoadOfAnyValue(clang::CodeGen::LValue, clang::CodeGen::AggValueSlot, clang::SourceLocation) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42da680)
#14 0x000000000405b009 (anonymous namespace)::X86_64ABIInfo::EmitVAArg(clang::CodeGen::CodeGenFunction&, clang::CodeGen::Address, clang::QualType, clang::CodeGen::AggValueSlot) const X86.cpp:0:0
#15 0x0000000004244508 clang::CodeGen::CodeGenFunction::EmitVAArg(clang::VAArgExpr*, clang::CodeGen::Address&, clang::CodeGen::AggValueSlot) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x4244508)
#16 0x0000000004305f8d (anonymous namespace)::AggExprEmitter::Visit(clang::Expr*) CGExprAgg.cpp:0:0
#17 0x0000000004306830 clang::CodeGen::CodeGenFunction::EmitAggExpr(clang::Expr const*, clang::CodeGen::AggValueSlot) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x4306830)
#18 0x00000000042e9a15 clang::CodeGen::CodeGenFunction::EmitAnyExprToMem(clang::Expr const*, clang::CodeGen::Address, clang::Qualifiers, bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42e9a15)
#19 0x00000000042f44c2 clang::CodeGen::CodeGenFunction::EmitMaterializeTemporaryExpr(clang::MaterializeTemporaryExpr const*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42f44c2)
#20 0x00000000042e4a71 clang::CodeGen::CodeGenFunction::EmitLValueHelper(clang::Expr const*, clang::CodeGen::KnownNonNull_t) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42e4a71)
#21 0x00000000042f2de5 clang::CodeGen::CodeGenFunction::EmitReferenceBindingToExpr(clang::Expr const*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42f2de5)
#22 0x000000000424ca45 clang::CodeGen::CodeGenFunction::EmitCallArg(clang::CodeGen::CallArgList&, clang::Expr const*, clang::QualType) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x424ca45)
#23 0x0000000004257ea7 clang::CodeGen::CodeGenFunction::EmitCallArgs(clang::CodeGen::CallArgList&, clang::CodeGen::CodeGenFunction::PrototypeWrapper, llvm::iterator_range<clang::Stmt::CastIterator<clang::Expr, clang::Expr const* const, clang::Stmt const* const>>, clang::CodeGen::CodeGenFunction::AbstractCallee, unsigned int, clang::CodeGen::CodeGenFunction::EvaluationOrder) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x4257ea7)
#24 0x00000000042f592d clang::CodeGen::CodeGenFunction::EmitCall(clang::QualType, clang::CodeGen::CGCallee const&, clang::CallExpr const*, clang::CodeGen::ReturnValueSlot, llvm::Value*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42f592d)
#25 0x00000000042f7236 clang::CodeGen::CodeGenFunction::EmitCallExpr(clang::CallExpr const*, clang::CodeGen::ReturnValueSlot) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42f7236)
#26 0x000000000434aad2 (anonymous namespace)::ScalarExprEmitter::VisitCallExpr(clang::CallExpr const*) CGExprScalar.cpp:0:0
#27 0x000000000433ed7b clang::StmtVisitorBase<std::add_pointer, (anonymous namespace)::ScalarExprEmitter, llvm::Value*>::Visit(clang::Stmt*) CGExprScalar.cpp:0:0
#28 0x000000000434bd82 (anonymous namespace)::ScalarExprEmitter::VisitExprWithCleanups(clang::ExprWithCleanups*) CGExprScalar.cpp:0:0
#29 0x000000000433f941 clang::StmtVisitorBase<std::add_pointer, (anonymous namespace)::ScalarExprEmitter, llvm::Value*>::Visit(clang::Stmt*) CGExprScalar.cpp:0:0
#30 0x0000000004344a5c clang::CodeGen::CodeGenFunction::EmitScalarExpr(clang::Expr const*, bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x4344a5c)
#31 0x0000000004296aff clang::CodeGen::CodeGenFunction::EmitScalarInit(clang::Expr const*, clang::ValueDecl const*, clang::CodeGen::LValue, bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x4296aff)
#32 0x0000000004298334 clang::CodeGen::CodeGenFunction::EmitExprAsInit(clang::Expr const*, clang::ValueDecl const*, clang::CodeGen::LValue, bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x4298334)
#33 0x00000000042a47f1 clang::CodeGen::CodeGenFunction::EmitAutoVarInit(clang::CodeGen::CodeGenFunction::AutoVarEmission const&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42a47f1)
#34 0x00000000042a4db6 clang::CodeGen::CodeGenFunction::EmitVarDecl(clang::VarDecl const&) (.part.0) CGDecl.cpp:0:0
#35 0x00000000042a5006 clang::CodeGen::CodeGenFunction::EmitDecl(clang::Decl const&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x42a5006)
#36 0x0000000003e78047 clang::CodeGen::CodeGenFunction::EmitDeclStmt(clang::DeclStmt const&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3e78047)
#37 0x0000000003e8e8c9 clang::CodeGen::CodeGenFunction::EmitSimpleStmt(clang::Stmt const*, llvm::ArrayRef<clang::Attr const*>) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3e8e8c9)
#38 0x0000000003e86535 clang::CodeGen::CodeGenFunction::EmitStmt(clang::Stmt const*, llvm::ArrayRef<clang::Attr const*>) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3e86535)
#39 0x0000000003e8e2e4 clang::CodeGen::CodeGenFunction::EmitCompoundStmtWithoutScope(clang::CompoundStmt const&, bool, clang::CodeGen::AggValueSlot) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3e8e2e4)
#40 0x0000000003ef5fc4 clang::CodeGen::CodeGenFunction::EmitFunctionBody(clang::Stmt const*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3ef5fc4)
#41 0x0000000003f087df clang::CodeGen::CodeGenFunction::GenerateCode(clang::GlobalDecl, llvm::Function*, clang::CodeGen::CGFunctionInfo const&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3f087df)
#42 0x0000000003f6c8fa clang::CodeGen::CodeGenModule::EmitGlobalFunctionDefinition(clang::GlobalDecl, llvm::GlobalValue*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3f6c8fa)
#43 0x0000000003f67775 clang::CodeGen::CodeGenModule::EmitGlobalDefinition(clang::GlobalDecl, llvm::GlobalValue*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3f67775)
#44 0x0000000003f67d2b clang::CodeGen::CodeGenModule::EmitGlobal(clang::GlobalDecl) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3f67d2b)
#45 0x0000000003f71e63 clang::CodeGen::CodeGenModule::EmitTopLevelDecl(clang::Decl*) (.part.0) CodeGenModule.cpp:0:0
#46 0x0000000004473206 (anonymous namespace)::CodeGeneratorImpl::HandleTopLevelDecl(clang::DeclGroupRef) ModuleBuilder.cpp:0:0
#47 0x0000000004463428 clang::BackendConsumer::HandleTopLevelDecl(clang::DeclGroupRef) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x4463428)
#48 0x00000000065b0ba4 clang::ParseAST(clang::Sema&, bool, bool) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x65b0ba4)
#49 0x0000000004470d68 clang::CodeGenAction::ExecuteAction() (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x4470d68)
#50 0x0000000004702ca9 clang::FrontendAction::Execute() (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x4702ca9)
#51 0x000000000468655e clang::CompilerInstance::ExecuteAction(clang::FrontendAction&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x468655e)
#52 0x00000000047ec0ce clang::ExecuteCompilerInvocation(clang::CompilerInstance*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x47ec0ce)
#53 0x0000000000cb996f cc1_main(llvm::ArrayRef<char const*>, char const*, void*) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0xcb996f)
#54 0x0000000000cb2bba ExecuteCC1Tool(llvm::SmallVectorImpl<char const*>&, llvm::ToolContext const&) driver.cpp:0:0
#55 0x00000000044b4189 void llvm::function_ref<void ()>::callback_fn<clang::driver::CC1Command::Execute(llvm::ArrayRef<std::optional<llvm::StringRef>>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>*, bool*) const::'lambda'()>(long) Job.cpp:0:0
#56 0x0000000003a6c644 llvm::CrashRecoveryContext::RunSafely(llvm::function_ref<void ()>) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x3a6c644)
#57 0x00000000044b477f clang::driver::CC1Command::Execute(llvm::ArrayRef<std::optional<llvm::StringRef>>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>*, bool*) const (.part.0) Job.cpp:0:0
#58 0x000000000447a6c5 clang::driver::Compilation::ExecuteCommand(clang::driver::Command const&, clang::driver::Command const*&, bool) const (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x447a6c5)
#59 0x000000000447b12d clang::driver::Compilation::ExecuteJobs(clang::driver::JobList const&, llvm::SmallVectorImpl<std::pair<int, clang::driver::Command const*>>&, bool) const (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x447b12d)
#60 0x0000000004482d75 clang::driver::Driver::ExecuteCompilation(clang::driver::Compilation&, llvm::SmallVectorImpl<std::pair<int, clang::driver::Command const*>>&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0x4482d75)
#61 0x0000000000cb6d65 clang_main(int, char**, llvm::ToolContext const&) (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0xcb6d65)
#62 0x0000000000b89614 main (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0xb89614)
#63 0x0000754bbd629d90 (/lib/x86_64-linux-gnu/libc.so.6+0x29d90)
#64 0x0000754bbd629e40 __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29e40)
#65 0x0000000000cb266e _start (/opt/compiler-explorer/clang-assertions-trunk/bin/clang+++0xcb266e)
clang++: error: clang frontend command failed with exit code 134 (use -v to see invocation)
Compiler returned: 134
```

localspook:
Fixes #107836.

For code like `auto s = va_arg(foo, std::string)`, clang currently emits an error-by-default warning (`-Wnon-pod-varargs`), but if this code gets to the IR generation phase, it trips this assertion:
https://github.com/llvm/llvm-project/blob/d63ca96ec35e963e651d1b8765283117ca83e1f9/clang/lib/CodeGen/CGExprAgg.cpp#L2264-L2270
Make it an unconditional error so that it never gets that far.

### DIFF
--- a/clang/docs/ReleaseNotes.rst
+++ b/clang/docs/ReleaseNotes.rst
@@ -653,6 +653,7 @@ Miscellaneous Clang Crashes Fixed
 - Fixed an assertion failure when parsing an invalid out-of-line enum definition with template parameters. (#GH187909)
 - Fixed an assertion failure on invalid template template parameter during typo correction. (#GH183983)
 - Fixed an assertion failure in ``isAtEndOfMacroExpansion`` on macro expansions crossing the boundary of two fileIDs. (#GH115007), (#GH21755)
+- Fixed an assertion failure when passing a non-POD type as the second argument to ``va_arg``. (#GH107836)
 
 OpenACC Specific Changes
 ------------------------
--- a/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -11327,9 +11327,8 @@ def err_second_parameter_to_va_arg_incomplete: Error<
   "second argument to 'va_arg' is of incomplete type %0">;
 def err_second_parameter_to_va_arg_abstract: Error<
   "second argument to 'va_arg' is of abstract type %0">;
-def warn_second_parameter_to_va_arg_not_pod : Warning<
-  "second argument to 'va_arg' is of non-POD type %0">,
-  InGroup<NonPODVarargs>, DefaultError;
+def err_second_parameter_to_va_arg_not_pod : Error<
+  "second argument to 'va_arg' is of non-POD type %0">;
 def warn_second_parameter_to_va_arg_ownership_qualified : Warning<
   "second argument to 'va_arg' is of ARC ownership-qualified type %0">,
   InGroup<NonPODVarargs>, DefaultError;
--- a/clang/test/SemaCXX/va_arg-non-pod-crash.cpp
+++ b/clang/test/SemaCXX/va_arg-non-pod-crash.cpp
@@ -1,0 +1,15 @@
+// RUN: %clang_cc1 -emit-llvm-only -verify -Wno-everything %s
+
+struct NonPODType {
+  NonPODType(const NonPODType&);
+  NonPODType& operator=(const NonPODType&);
+  NonPODType(NonPODType&&);
+  NonPODType& operator=(NonPODType&&);
+};
+
+void f(int first_arg, ...) {
+  __builtin_va_list args;
+  __builtin_va_start(args, first_arg);
+  NonPODType s = __builtin_va_arg(args, NonPODType); // expected-error {{second argument to 'va_arg' is of non-POD type 'NonPODType'}}
+  __builtin_va_end(args);
+}
--- a/clang/lib/Sema/SemaExpr.cpp
+++ b/clang/lib/Sema/SemaExpr.cpp
@@ -17279,10 +17279,9 @@ ExprResult Sema::BuildVAArgExpr(SourceLocation BuiltinLoc,
     if (!TInfo->getType().isPODType(Context)) {
       Diag(TInfo->getTypeLoc().getBeginLoc(),
            TInfo->getType()->isObjCLifetimeType()
-             ? diag::warn_second_parameter_to_va_arg_ownership_qualified
-             : diag::warn_second_parameter_to_va_arg_not_pod)
-        << TInfo->getType()
-        << TInfo->getTypeLoc().getSourceRange();
+               ? diag::warn_second_parameter_to_va_arg_ownership_qualified
+               : diag::err_second_parameter_to_va_arg_not_pod)
+          << TInfo->getType() << TInfo->getTypeLoc().getSourceRange();
     }
 
     if (TInfo->getType()->isArrayType()) {
--- a/clang/test/SemaCXX/vararg-non-pod.cpp
+++ b/clang/test/SemaCXX/vararg-non-pod.cpp
@@ -173,7 +173,7 @@ void t6b(Foo somearg, ... ) {
 void t7(int n, ...) {
   __builtin_va_list list;
   __builtin_va_start(list, n);
-  (void)__builtin_va_arg(list, C); // expected-warning{{second argument to 'va_arg' is of non-POD type 'C'}}
+  (void)__builtin_va_arg(list, C); // expected-error {{second argument to 'va_arg' is of non-POD type 'C'}}
   __builtin_va_end(list);
 }
 


Sirraide (commented):
Given that we have always either errored or crashed on this, and given the fact that you should probably never do this (even though GCC allows it for some reason...), I think this is fine.

Sirraide (dismissed):
Meant to select Approve instead of Comment

ojhunt (commented):
@Sirraide I'm not sure about this change.

This is an assertion failure, not a release mode crash (still obviously not ideal). Without assertions we do _seem_ to successfully generate code that matches what GCC does.

In current releases the non-pod-* warnings are already errors by default, so this are presumably intentionally disabled for the purpose of gcc compatibility rather than being in general use.

cor3ntin (changes requested):
@ojhunt right, the only thing this is doing is not continuing.

I'm okay with the change - but the release note should say very clearly we are making it a non-demotable error, which is a breaking change and should be listed as such.

And before doing that we probably should investigate the crash and whether we can fix it.
If we make it an error we should explore whether we need to remove any support code that may or may not exists to support non-pod typed in calls to vararg functions 

@AaronBallman

Sirraide (commented):
Ah my bad, I was under the impression that this was UB, but apparently we _do_ know to call the copy-constructor of e.g. `std::string` if you pass it to a variadic function, so yes, if we can codegen this properly then we should support it.

Also, I did a code search and for some reason there are people out there that are doing `va_arg(ap, std::string)`...

cor3ntin:
It is implementation defined (https://eel.is/c++draft/expr#call-12) - so we can make functional changes.

But not make functional changes and say in the release notes we fixed an assert.

localspook:
> Without assertions we do *seem* to successfully generate code that matches what GCC does.

> apparently we *do* know to call the copy-constructor of e.g. `std::string` if you pass it to a variadic function, so yes, if we can codegen this properly then we should support it.

Is that really true? If you try to pass a `std::string` to a vararg function, at the callsite clang says (https://godbolt.org/z/hKnsxqYcP):

> cannot pass object of non-trivial type 'std::string' (aka 'basic_string<char>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]

And you can look at the assembly output to confirm that LLVM reduces the entire program down to a trap. (Why it isn't an outright error, I don't know.) I don't think we actually support passing non-POD types to vararg functions in any meaningful manner, so rejecting these `va_arg` calls is not breaking any code that would have actually executed.

Sirraide:
> Is that really true? If you try to pass a `std::string` to a vararg function, at the callsite clang says

Interesting; I was only looking at -O0 codegen where I saw a copy-constructor call

Fznamznon:
FYI, MSVC does allow non-pod types in va_arg and we try to do that as well on Windows (https://reviews.llvm.org/D5492 ) though AFAIK this doesn't work at all times.

shafik:
> va_arg(ap, std::string)

This does seem to be quite common and in some quite large projects.

ojhunt:
Ok I can confirm that std::string lowers to a trap, trying to find my example where it was permitted.

ojhunt:
Ok, that's it: the included test case _does_ compile to something that looks more or less identical to what GCC does:

 https://compiler-explorer.com/z/7x4rWYExq

It also compiles with other non-pod types (polymorphic types: https://compiler-explorer.com/z/b5G1qj31K), but if you add something like std::string both clang and gcc reject: https://compiler-explorer.com/z/P7x6T9jb7

~I'm not sure what std::string is doing to make the compilation fail? But our current behavior matches GCCs 🤷‍♂️~

[edit: the answer is "when using std::string you should include <string>" -- we happily copy a std::string here: https://compiler-explorer.com/z/qb3avxj5E ]

ojhunt:
Oh wait, I'm a dumbass, the test case is just the receiving end

[edit] no I'm not, I found my test case from last night (yay for CE history): https://compiler-explorer.com/z/qExvTTdhj

ojhunt:
~And out of paranoia I did an x86_64 test as well: https://compiler-explorer.com/z/Gn9Ynzv6q~

ojhunt:
Many many apologies: All my last comments were nonsense, it turns out I just can't read assembly any more. Many many apologies for the noise. I think this is fine, and it should be considered an assertion fix.

ojhunt (approved):
This PR is good, it is an assertion fix, I can imagine some code might experience build time breakage but it would have had to have been in unused code paths.

I'm surprised there were no codegen tests to update.

localspook:
Both of those examples have a trap as the last instruction: ARM has `brk #0x1`, x86-64 has `ud2`.

So to summarize: we're compatible with GCC when it comes to the implementation of the variadic function (the `va_arg`), but generate a trap at the call site. So I guess, if the call site is compiled with GCC and the implementation is compiled with Clang, those would successfully link and work together, but it doesn't seem that our compatibility with GCC goes beyond that.

localspook:
As for MSVC: is there agreement that we definitely want to preserve that compatibility? (I can rework the PR to do it, I just want a go-ahead first). If yes, then while we're at it, do we maybe want to make it so `-fms-compatibility` by itself enables support for non-POD types, instead of forcing users to *also* pass `-Wno-non-pod-varargs`?

(Also, somewhat unrelated, but MSVC actually has a warning for this: [C4840](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4840?view=msvc-170). Only enabled at `/W4` though.)

Sirraide (approved):
Right, so considering that it seems like any program that attempts to make use of this ‘feature’ by disabling/demoting the warning will just crash if you actually try to exercise those code paths, I don’t think there is a point in allowing that. If this breaks someone’s code... then that’s probably for the better, because it was already broken to begin with.

Fznamznon:
> As for MSVC: is there agreement that we definitely want to preserve that compatibility?

Well, clang claims to be trying to be compatible with MSVC when compiling for Windows in its own doc here https://clang.llvm.org/docs/MSVCCompatibility.html . I suppose we should at least not remove anything that makes clang compatible with MSVC until there is an RFC accepted that says stop doing so.

>  If yes, then while we're at it, do we maybe want to make it so -fms-compatibility by itself enables support for non-POD types, instead of forcing users to also pass -Wno-non-pod-varargs?

Given that it is not portable and MSVC has its own warning for that, I think preserving two separate flags still makes sense. I suppose clang-cl should handle it.

ojhunt:
Does anyone have any concerns with landing (@cor3ntin you're marked as having concerns but that may be due to my misreading of the asm earlier :D)

clang/docs/ReleaseNotes.rst
```diff
@@ -674,6 +653,7 @@ Miscellaneous Clang Crashes Fixed
 - Fixed an assertion failure when parsing an invalid out-of-line enum definition with template parameters. (#GH187909)
 - Fixed an assertion failure on invalid template template parameter during typo correction. (#GH183983)
 - Fixed an assertion failure in ``isAtEndOfMacroExpansion`` on macro expansions crossing the boundary of two fileIDs. (#GH115007), (#GH21755)
+- Fixed an assertion failure when passing a non-POD type as the second argument to ``va_arg``. (#GH107836)
```
cor3ntin, line 656: Can you make sure the release notes reflect the changes?
You are making a functional changes, not just fixing an assertion

clang/include/clang/Basic/DiagnosticSemaKinds.td
cor3ntin, line 11332: Btw, do we still need that group? Otherwise we should remove it

AaronBallman (changes requested):
This has been a default error since 2011 (https://github.com/llvm/llvm-project/commit/254a5c07e7c96) and I question why we'd want to make it a hard error 15 years later. There's evidence that this diagnostic is disabled in practice: https://sourcegraph.com/search?q=context:global+-Wno-non-pod-varargs+-file:.*test.*&patternType=keyword&sm=0

I think the correct approach is to fix the crash, not to make this a hard error. There are non-POD types for which this is still reasonable: https://godbolt.org/z/sEcnez9jd

clang/test/SemaCXX/va_arg-non-pod-crash.cpp
```diff
@@ -0,0 +1,15 @@
+// RUN: %clang_cc1 -emit-llvm-only -verify -Wno-everything %s
```
AaronBallman, line 1:
```suggestion
// RUN: %clang_cc1 only -verify %s
```
We shouldn't need to emit LLVM IR for a Sema test, and we should not use `-Wno-everything` because we want to know when new diagnostic behaviors impact tests.

ojhunt:
> This has been a default error since 2011 ([254a5c07e7c96](https://github.com/llvm/llvm-project/commit/254a5c07e7c96)) and I question why we'd want to make it a hard error 15 years later. There's evidence that this diagnostic is disabled in practice: https://sourcegraph.com/search?q=context:global+-Wno-non-pod-varargs+-file:.*test.*&patternType=keyword&sm=0
> 
> I think the correct approach is to fix the crash, not to make this a hard error. There are non-POD types for which this is still reasonable: https://godbolt.org/z/sEcnez9jd

Given there _are_ non-pod types for which this works I think we should separate out the warnings - there's a difference between extension and "will always trap".

ojhunt:
> > This has been a default error since 2011 ([254a5c07e7c96](https://github.com/llvm/llvm-project/commit/254a5c07e7c96)) and I question why we'd want to make it a hard error 15 years later. There's evidence that this diagnostic is disabled in practice: https://sourcegraph.com/search?q=context:global+-Wno-non-pod-varargs+-file:.*test.*&patternType=keyword&sm=0
> > I think the correct approach is to fix the crash, not to make this a hard error. There are non-POD types for which this is still reasonable: https://godbolt.org/z/sEcnez9jd
> 
> Given there _are_ non-pod types for which this works I think we should separate out the warnings - there's a difference between extension and "will always trap".

Oh, actually we already do have different warnings - the diagnostics and codegen are based on VarArgKind, not a generic isPOD check - we only emit the trap for VarArgKind::Undefined. We could restrict the error case to just that.

ojhunt:
> > > This has been a default error since 2011 ([254a5c07e7c96](https://github.com/llvm/llvm-project/commit/254a5c07e7c96)) and I question why we'd want to make it a hard error 15 years later. There's evidence that this diagnostic is disabled in practice: https://sourcegraph.com/search?q=context:global+-Wno-non-pod-varargs+-file:.*test.*&patternType=keyword&sm=0
> > > I think the correct approach is to fix the crash, not to make this a hard error. There are non-POD types for which this is still reasonable: https://godbolt.org/z/sEcnez9jd
> > 
> > 
> > Given there _are_ non-pod types for which this works I think we should separate out the warnings - there's a difference between extension and "will always trap".
> 
> Oh, actually we already do have different warnings - the diagnostics and codegen are based on VarArgKind, not a generic isPOD check - we only emit the trap for VarArgKind::Undefined. We could restrict the error case to just that.

The implementation is weird though - we only do it for undefined, not MSVCUndefined, and we do the rewrite in Sema by literally constructing a CallExpr for `__builtin_trap` in the AST?

e.g

```cpp
void f(...);

f(SomeInvalidType());
```

becomes

```cpp
f((__builtin_trap(), SomeInvalidType()));
```

PR #198669, opened 2026-05-19
[AMDGPU][True16] Create tests that will demonstrate true16 G_SEXTLOAD/G_ZEXTLOAD legalization changes

saxlungs:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

Stack PRs:
https://github.com/llvm/llvm-project/pull/198670
https://github.com/llvm/llvm-project/pull/198671

See https://github.com/llvm/llvm-project/pull/195289 for previous discussion

PR #195289, opened 2026-05-01
[AMDGPU][True16] Legalize extloads into 16-bit registers

saxlungs:

PR #198670, opened 2026-05-19
[AMDGPU][True16] Legalize extloads into 16-bit registers

saxlungs:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

Stack PRs:
https://github.com/llvm/llvm-project/pull/198669
https://github.com/llvm/llvm-project/pull/198671

See https://github.com/llvm/llvm-project/pull/195289 for previous discussion

### DIFF
--- a/llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
+++ b/llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
@@ -464,7 +464,11 @@ static bool isLoadStoreSizeLegal(const GCNSubtarget &ST,
     MemSize = std::max(MemSize, Align);
 #endif
 
-  // Only allow extloads to up to 32 bits.
+  // We want to allow extending loads into up to a 32-bit destination register.
+  // However, this would potentially allow 16-bit destinations even without
+  // True16. This function is used by isLoadStoreLegal, which will also call
+  // isRegisterType on the destination register type which will disallow
+  // 16-bit types without True16, so this is safe.
   if (MemSize != RegSize && RegSize > 32)
     return false;
 
--- a/llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
+++ b/llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
@@ -464,12 +464,8 @@ static bool isLoadStoreSizeLegal(const GCNSubtarget &ST,
     MemSize = std::max(MemSize, Align);
 #endif
 
-  // We want to allow extending loads into up to a 32-bit destination register.
-  // However, this would potentially allow 16-bit destinations even without
-  // True16. This function is used by isLoadStoreLegal, which will also call
-  // isRegisterType on the destination register type which will disallow
-  // 16-bit types without True16, so this is safe.
-  if (MemSize != RegSize && RegSize > 32)
+  // Only 1-byte and 2-byte to 32-bit extloads are valid.
+  if (MemSize != RegSize && RegSize != 32)
     return false;
 
   if (MemSize > maxSizeForAddrSpace(ST, AS, IsLoad,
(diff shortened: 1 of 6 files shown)


llvm/test/CodeGen/AMDGPU/GlobalISel/load-d16.ll
```diff
@@ -30,8 +30,11 @@ define amdgpu_ps void @load_P0_B16_D16_Hi(<2 x i16> %vec, ptr addrspace(0) %ptra
 define amdgpu_ps void @sextload_P0_i8_D16(<2 x i16> %vec, ptr addrspace(0) %ptra, ptr addrspace(0) %out) {
 ; GFX12-LABEL: sextload_P0_i8_D16:
 ; GFX12:       ; %bb.0:
-; GFX12-NEXT:    flat_load_d16_i8 v0, v[1:2]
+; GFX12-NEXT:    flat_load_d16_i8 v1, v[1:2]
 ; GFX12-NEXT:    s_wait_loadcnt_dscnt 0x0
+; GFX12-NEXT:    v_mov_b16_e32 v1.h, 0
+; GFX12-NEXT:    s_delay_alu instid0(VALU_DEP_1)
+; GFX12-NEXT:    v_and_or_b32 v0, 0xffff0000, v0, v1
 ; GFX12-NEXT:    flat_store_b32 v[3:4], v0
```
petar-avramovic: this test should get true16 and fake16 run line in precommit

saxlungs: Added to that PR

llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
```diff
@@ -464,8 +464,8 @@ static bool isLoadStoreSizeLegal(const GCNSubtarget &ST,
     MemSize = std::max(MemSize, Align);
 #endif
 
-  // Only 1-byte and 2-byte to 32-bit extloads are valid.
-  if (MemSize != RegSize && RegSize != 32)
+  // Only allow extloads to up to 32 bits.
+  if (MemSize != RegSize && RegSize > 32)
```
petar-avramovic: this is not true16 specific yet it does not affect `fake16`.
I did not check but it looks like some overlap in conditions for legalizer actions, think it requires more strict check or comment explaining why it only affects true16 s/zextload

petar-avramovic (commented):
LGTM in general, condition for which loads are also legal needs some clarification

llvm/test/CodeGen/AMDGPU/GlobalISel/load-d16.ll
```diff
@@ -29,12 +29,22 @@ define amdgpu_ps void @load_P0_B16_D16_Hi(<2 x i16> %vec, ptr addrspace(0) %ptra
...
-; GFX12-NEXT:    flat_store_b32 v[3:4], v0
-; GFX12-NEXT:    s_endpgm
+; GFX12-TRUE16-LABEL: sextload_P0_i8_D16:
+; GFX12-TRUE16:       ; %bb.0:
+; GFX12-TRUE16-NEXT:    flat_load_d16_i8 v1, v[1:2]
+; GFX12-TRUE16-NEXT:    s_wait_loadcnt_dscnt 0x0
+; GFX12-TRUE16-NEXT:    v_mov_b16_e32 v1.h, 0
+; GFX12-TRUE16-NEXT:    s_delay_alu instid0(VALU_DEP_1)
+; GFX12-TRUE16-NEXT:    v_and_or_b32 v0, 0xffff0000, v0, v1
+; GFX12-TRUE16-NEXT:    flat_store_b32 v[3:4], v0
```
Sisyph, line 39: This file shows a pretty bad code quality regression on the true16 path. I think the fake16 sequences are correct and we should try to generate the same one for true16.

Sisyph: Or is that fixed farther up the stack?

Sisyph: Ok it is, nevermind. Please try to add the links to other commits in the stack though.

llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
saxlungs: Yeah I can explain here and then I'll add to the comment why. This function `isLoadStoreSizeLegal` is used in the parent function `isLoadStoreLegal` which is the actual function being used in the legalizer actions. This is alongside another check `isRegisterType` on the destination register's type. In `isRegisterType` we call `isRegisterSize` which only allows a 16bit type if true16 is enabled.

So technically this change to `isLoadStoreSizeLegal` would allow for both true16 and fake16 to use it, but the `isRegisterType` check prevents that

llvm/test/CodeGen/AMDGPU/GlobalISel/load-d16.ll
saxlungs: Added links to the stacked PRs in all of their descriptions. I'll see if I can get it to do that in the future

llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
petar-avramovic: I see, think it would be better to use legalForTypesWithMemDesc as for G_LOAD/G_STORE if possible, but either way is fine by me.

llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
```diff
@@ -464,8 +464,12 @@ static bool isLoadStoreSizeLegal(const GCNSubtarget &ST,
     MemSize = std::max(MemSize, Align);
 #endif
 
-  // Only 1-byte and 2-byte to 32-bit extloads are valid.
-  if (MemSize != RegSize && RegSize != 32)
+  // We want to allow extending loads into up to a 32-bit destination register.
+  // However, this would potentially allow 16-bit destinations even without
+  // True16. This function is used by isLoadStoreLegal, which will also call
+  // isRegisterType on the destination register type which will disallow
+  // 16-bit types without True16, so this is safe.
```
petar-avramovic: Not sure about this comment, too big, there are only few cases (s8->s32 s8->s16 and s16->s32). This it would be much clearer to have in Legalizer info, something like
```
    Actions.legalForTypesWithMemDesc(!ST.useRealTrue16Insts(),
                                     {{S32, GlobalPtr, S8, GlobalAlign8},

    Actions.legalForTypesWithMemDesc(ST.useRealTrue16Insts(),
                                     {{S16, GlobalPtr, S8, GlobalAlign8},
```
instead of having to chase it in around for conditions.

llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
```diff
@@ -1735,19 +1735,23 @@ AMDGPULegalizerInfo::AMDGPULegalizerInfo(const GCNSubtarget &ST_,
...
+                                     {S32, LocalPtr, S8, 8},
+                                     {S32, LocalPtr, S16, 16},
+                                     {S32, PrivatePtr, S8, 8},
+                                     {S32, PrivatePtr, S16, 16},
+                                     {S32, ConstantPtr, S8, 8},
+                                     {S32, ConstantPtr, S16, 2 * 8}})
+          .legalForTypesWithMemDesc(ST.useRealTrue16Insts(),
+                                    {{S16, GlobalPtr, S8, GlobalAlign8},
+                                     {S16, LocalPtr, S8, GlobalAlign8},
+                                     {S16, PrivatePtr, S8, GlobalAlign8}})
```
petar-avramovic: I really don't like being the one blocking this to get merged but think you are doing same thing in two places again, here and `if (MemSize != RegSize && RegSize > 32)` can you pick one you like better and lets merge it.

petar-avramovic (approved):

llvm/lib/Target/AMDGPU/AMDGPULegalizerInfo.cpp
saxlungs: All good, I misunderstood that you were saying this should replace that check entirely.

PR #198672, opened 2026-05-19
[CIR] Lower calling through a variable

erichkeane:
We managed to miss a condition when lowering emitCallee, where a DeclRefExpr referenced a function object.  This patch adds that condition, which will result in these being lowered properly as an indirect call.

### DIFF
--- a/clang/test/CIR/CodeGen/call.cpp
+++ b/clang/test/CIR/CodeGen/call.cpp
@@ -143,12 +143,12 @@ void use_TakesFunc() {
 
 // CIR-LABEL: _Z9TakesFuncIFivEEDcRKT_
 // CIR-NEXT: %[[FUNC_ALLOCA:.*]] = cir.alloca !cir.ptr<!cir.func<() -> !s32i>>, !cir.ptr<!cir.ptr<!cir.func<() -> !s32i>>>, ["f", init, const]
-// CIR-NEXT: %[[FUNC_LOAD:.*]] = cir.load %[[FUNC_ALLOCA]] : !cir.ptr<!cir.ptr<!cir.func<() -> !s32i>>>, !cir.ptr<!cir.func<() -> !s32i>>
+// CIR: %[[FUNC_LOAD:.*]] = cir.load %[[FUNC_ALLOCA]] : !cir.ptr<!cir.ptr<!cir.func<() -> !s32i>>>, !cir.ptr<!cir.func<() -> !s32i>>
 // CIR-NEXT: %[[CALL:.*]] = cir.call %[[FUNC_LOAD]]() : (!cir.ptr<!cir.func<() -> !s32i>>) -> (!s32i {llvm.noundef})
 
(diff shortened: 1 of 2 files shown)


erichkeane:
Linux fail is unrelated, looks to be something with the python run checking of lldb.

clang/test/CIR/CodeGen/call.cpp
```diff
@@ -130,4 +130,33 @@ void f16() {
...
+// CIR-NEXT: cir.store %[[CALL]], %[[RET_ALLOCA]] : !s32i, !cir.ptr<!s32i>
+
+// LLVM-LABEL: _Z9TakesFuncIFivEEDcRKT_
+// LLVM-NEXT: %[[FUNC_ALLOCA:.*]] = alloca ptr
+// LLVM-NEXT: %[[RET_ALLOCA:.*]] = alloca i32
+// LLVM-NEXT: store ptr %{{.*}}, ptr %[[FUNC_ALLOCA]]
+// LLVM-NEXT: %[[FUNC_LOAD:.*]] = load ptr, ptr %[[FUNC_ALLOCA]]
+// LLVM-NEXT: %[[CALL:.*]] = call noundef i32 %[[FUNC_LOAD]]()
+// LLVM-NEXT: store i32 %[[CALL]], ptr %[[RET_ALLOCA]]
+
```
andykaylor: Delete extra blank lines?

clang/test/CIR/CodeGen/call.cpp
```diff
@@ -130,4 +130,33 @@ void f16() {
...
+void use_TakesFunc() {
+  TakesFunc(Passed);
+}
+
+// CIR-LABEL: _Z9TakesFuncIFivEEDcRKT_
+// CIR-NEXT: %[[FUNC_ALLOCA:.*]] = cir.alloca !cir.ptr<!cir.func<() -> !s32i>>, !cir.ptr<!cir.ptr<!cir.func<() -> !s32i>>>, ["f", init, const]
+// CIR-NEXT: %[[RET_ALLOCA:.*]] = cir.alloca !s32i, !cir.ptr<!s32i>, ["__retval"]
+// CIR-NEXT: cir.store %{{.*}}, %[[FUNC_ALLOCA]] : !cir.ptr<!cir.func<() -> !s32i>>, !cir.ptr<!cir.ptr<!cir.func<() -> !s32i>>>
+// CIR-NEXT: %[[FUNC_LOAD:.*]] = cir.load %[[FUNC_ALLOCA]] : !cir.ptr<!cir.ptr<!cir.func<() -> !s32i>>>, !cir.ptr<!cir.func<() -> !s32i>>
```
andykaylor: You could probably drop all the CIR checks except the function name, this line, and the one below it. Likewise with the LLVM checks, the call is all we care about here, right?

andykaylor (approved):
lgtm, with a couple of suggestions for cleaning up the test

PR #198674, opened 2026-05-19
[llvm-profgen] Fix assertion condition for the top-level probes address range checks

HighW4y2H3ll:
Top-level functions may contains multiple disjoint address ranges, and the pseudo probes may not be stored as sorted. The original assertion check is problematic because the first pseudo probe may not necessarily falls into the current function range even though they are both part of the same function.

Update the assertion condition so that we won't hit the assertion failure as long as there one pseudo probe falls in the checked function range.

### DIFF
--- a/llvm/tools/llvm-profgen/ProfiledBinary.cpp
+++ b/llvm/tools/llvm-profgen/ProfiledBinary.cpp
@@ -1243,10 +1243,11 @@ void ProfiledBinary::loadSymbolsFromPseudoProbe() {
               InlineTreeNode->Parent);
 
         auto TopLevelProbes = InlineTreeNode->getProbes();
-        assert(find_if(TopLevelProbes,
-                       [Start = StartAddr, End = EndAddr](const auto &P) {
-                         return P.getAddress() >= Start && P.getAddress() < End;
-                       }) != TopLevelProbes.end() &&
+        assert(llvm::any_of(TopLevelProbes,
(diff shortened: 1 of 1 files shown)


apolloww:
Typo in title and body `assertation` -> `assertion`.

Also Maybe remove `[NFC]` from the title, since it changes assert behavior?

apolloww:
The top commit is `Merge branch 'main' into fix-profgen-probe-range-assert`, which looks strange? Maybe do a rebase and force push?

llvm/tools/llvm-profgen/ProfiledBinary.cpp
```diff
@@ -1243,10 +1243,10 @@ void ProfiledBinary::loadSymbolsFromPseudoProbe() {
               InlineTreeNode->Parent);
 
         auto TopLevelProbes = InlineTreeNode->getProbes();
-        [[maybe_unused]] auto TopProbe = TopLevelProbes.begin();
-        assert(TopProbe != TopLevelProbes.end() &&
-               TopProbe->getAddress() >= StartAddr &&
-               TopProbe->getAddress() < EndAddr &&
+        assert(find_if(TopLevelProbes,
```
apolloww: You can use `llvm::any_of` here

HighW4y2H3ll: Fixed!

apolloww (approved):
LGTM

PR #198676, opened 2026-05-19
[Offload][AMDGPU] Recycle kernel-dispatch and async-copy completion signals in recordEvent

lfmeadow:
## Summary

Commit 34028294e49b (PR #186856, "[Offload] Add support for measuring elapsed time between events") made `AMDGPUStreamTy::recordEvent` enqueue a real AQL barrier marker packet on the stream every time it is called, so the event has a packet-backed signal whose dispatch timestamps can be read by `hsa_amd_profiling_get_dispatch_time`.

That is correct, but unconditional. In libomptarget's record/wait/sync pattern, nobody calls `olGetEventElapsedTime` on these events — they are used purely for synchronization. Yet every `recordEvent` adds ~4 extra HSA host ops and one AQL barrier packet per call, which costs ~8% on BabelStream Dot at MI300.

This PR makes `recordEvent` recycle the completion signal of the stream's most recent op — either a kernel dispatch or an SDMA async memory copy — as the event's `TimingSignal`, instead of enqueueing a fresh barrier marker. The barrier-marker fallback is retained for cases where it is genuinely needed (empty stream after sync; last op was a host callback or a barrier).

Split into two commits for review clarity but kept together in this PR for landing:

1. **Recycle kernel-dispatch signal**: covers the inter-kernel record/wait critical path. Adds a `bool IsKernelDispatch` on `StreamSlotTy`, marks it from `pushKernelLaunch`, and wires the fast path in `recordEvent`.
2. **Also recycle async-copy signal**: replaces the bool with a tri-state `AMDGPUStreamTy::TimingKind` enum (`None` / `KernelDispatch` / `AsyncCopy`); marks SDMA slots from the four `push*MemoryCopy*Async` helpers (host-side `std::memcpy` callback slots stay `None`); enables `hsa_amd_profiling_async_copy_enable(1)` at queue init; threads the kind through `AMDGPUEventTy`; and has `getElapsedTime` dispatch on it to call either `hsa_amd_profiling_get_dispatch_time` (kernel + barrier marker packets) or `hsa_amd_profiling_get_async_copy_time` (SDMA packets). Cross-kind elapsed-time queries are allowed — both APIs return ticks on the same HSA system clock. Adds the corresponding declarations / DLWRAPs in `dynamic_hsa/`.

For `waitEvent` / `synchronizeOn`, the recycled slot is the right thing to wait on by construction (it is the most recent op on the stream), so nothing changes on the sync side beyond the slot pointer.

### Caveat (pre-existing, not made worse)

`ompt_stop_trace` calls `hsa_amd_profiling_set_profiler_enabled(Q, 0)` on already-initialized queues. After that, neither the recycled kernel signal nor a fresh barrier marker on those queues will carry valid timestamps, and `olGetEventElapsedTime` will return garbage on the affected queues. This is a pre-existing condition of the original barrier-marker path; an OMPT-side fix (refcount the toggle and/or keep `init()` and `setHSAQueueProfiling` in sync) is left for a follow-on.

## Test plan

New tests in `offload/unittests/OffloadAPI/event/olGetEventElapsedTime.cpp`:

- `BackToBackEventsShareDispatchSignal` — two events recorded back-to-back after a kernel dispatch must yield equal elapsed times relative to a third event (they recycle the same kernel completion signal). Closest API-level proxy for "no extra packet emitted."
- `EmptyQueueLeadingEventFallback` — leading event on an empty queue must still produce a valid elapsed time (barrier-marker fallback).
- `EventAfterQueueSyncFallback` — recording an event after `olSyncQueue` must also take the fallback path (no stale recycled signal).
- `MultipleKernelsBetweenEvents` — recycling must point at the *most* recent kernel's completion signal, not the first one.
- `MemcpyPriorOpRecyclesAsyncCopy` — single record after an async memcpy must succeed and produce a finite elapsed time when the second event recycles a kernel-dispatch signal (cross-kind query).
- `BackToBackEventsShareAsyncCopySignal` — two consecutive events after an async memcpy share the SDMA signal and return equal elapsed times.

Validation on this branch:

- [x] `clang-format-diff` clean on all changed hunks.
- [x] `clang++ -fsyntax-only` clean on `rtl.cpp` and the test file.
- [x] Equivalent change set built and ran successfully on a downstream tree (`libLLVMOffload.so` + `event.unittests`); full OffloadAPI gtest suite green on 4× MI210: 1273 PASSED / 40 SKIPPED (`olLaunchKernelCooperative` skips are pre-existing, unrelated to this PR) / 0 FAILED.

cc @jhuber6 @kevinsala

Made with [Cursor](https://cursor.com)


lfmeadow:
cc @ronlieb for review

offload/plugins-nextgen/amdgpu/src/rtl.cpp
```diff
@@ -1810,6 +1879,58 @@ Error AMDGPUStreamTy::recordEvent(AMDGPUEventTy &Event,
     return Plugin::error(ErrorCode::INVALID_NULL_POINTER,
                          "target queue was nullptr");
 
+  std::lock_guard<std::mutex> StreamLock(Mutex);
+
+  // Fast path: if the most recent operation on this stream is a kernel
```
jhuber6, line 1884: I swear if the AI has its way LLVM is going to be 90% comments.

offload/plugins-nextgen/amdgpu/dynamic_hsa/hsa.cpp
```diff
@@ -72,6 +72,8 @@ DLWRAP(hsa_amd_signal_async_handler, 5)
 DLWRAP(hsa_amd_pointer_info, 5)
 DLWRAP(hsa_amd_profiling_get_dispatch_time, 3)
 DLWRAP(hsa_amd_profiling_set_profiler_enabled, 2)
+DLWRAP(hsa_amd_profiling_async_copy_enable, 1)
```
jhuber6, line 75: What hSA version were these first introduced in?

jhuber6 (commented):
What's the exact desire here? I figured if we just wanted to revert the overhead we'd just stop adding the barrier until someone's actually measured something.

jhuber6:
In general, I'm really confused why this is so complicated at all. I would expect that if someone wants to profile that should be *opt-in* at least.

Wouldn't it simpler to just make the `olCreateEvent` that added the timing add the profiling barrier itself so the raw interface OpenMP uses remains untouched? Alternatively, have both stay opt-out and make an argument to `olCreateEvent` that opts-in to timing information.

ronlieb:
> In general, I'm really confused why this is so complicated at all. I would expect that if someone wants to profile that should be _opt-in_ at least.
> 
> Wouldn't it simpler to just make the `olCreateEvent` that added the timing add the profiling barrier itself so the raw interface OpenMP uses remains untouched? Alternatively, have both stay opt-out and make an argument to `olCreateEvent` that opts-in to timing information.

The original PR https://github.com/llvm/llvm-project/pull/186856 is NOT opt-in, if it was, this would not be a problem. could we try and revert the original PR and then have it reimpimented with opt-in as you suggest.

kevinsala:
This looks quite complicated and, if I understood correctly, it will potentially reduce the precision of timing information reported by `olGetEventElapsedTime`.

As commented in https://github.com/llvm/llvm-project/pull/186856#issuecomment-4491021990 some hours ago, adding a flag to `olCreateEvent` to explicitly enable the timing mechanism for that event sounds OK.

ronlieb:
> This looks quite complicated and, if I understood correctly, it will potentially reduce the precision of timing information reported by `olGetEventElapsedTime`.
> 
> As commented in [#186856 (comment)](https://github.com/llvm/llvm-project/pull/186856#issuecomment-4491021990) some hours ago, adding a flag to `olCreateEvent` to explicitly enable the timing mechanism for that event sounds OK.

@kevinsala  are you volunteering to take this one?

kevinsala:
In https://github.com/llvm/llvm-project/pull/186856 @leandrolcampos, the author of the PR, commented he was willing to make a patch for that.

kevinsala:
Btw, I would be OK to revert the original PR in the meantime.

ronlieb:
no longer needed

PR #198677, opened 2026-05-19
[AArch64] Skip FrameDestroy DW_CFA_AARCH64_negate_ra_state_with_pc

jroelofs:
The toggle semantics of DW_CFA_AARCH64_negate_ra_state_with_pc are evaluated over a linear stream of CFI opcodes, rather than along control flow edges, making it impossible to faithfully represent the RA state across all CFG paths for functions with non-trivial block layouts (for example hot-cold-splitting, or doubly-shrinkwrapped functions with multiple independent frames).

libunwind works around this by recording the signing address at the FrameSetup site, and not validating that the CFI state has toggled back to "unsigned" after authentication. We can therefore omit the FrameDestroy emission, treating the marker solely as a marker for the signing site. This holds so long as each function has at most one signing location, which is currently guaranteed since shrinkwrapping does not yet create functions with more than one frame.

DW_CFA_AARCH64_set_ra_state is the correct long-term fix, but cannot be used exclusively due to unwinder compatability concerns.

See: https://github.com/ARM-software/abi-aa/issues/327
     https://github.com/ARM-software/abi-aa/pull/346

rdar://174074919


Stylie777 (approved):
LGTM, Thanks!

PR #198678, opened 2026-05-19
[DWP] Stop storing DWPStringPool strings in Buffer

aeubanks:
When we can just write it out. This decreases peak memory usage of llvm-dwp on clang debug info from ~800MB to ~350MB without measurable performance changes.

Also remove DWPStringPool.h to avoid circular headers. It wasn't doing anything anyway...


dwblaikie (approved):
Excellent!

PR #198679, opened 2026-05-19
[LinkerWrapper] Fix temps being dumped to CWD instead of output path

jhuber6:
Summary:
Offloading save temps is a complex dance where we have clang,
linker-wrapper, and lld all making their own temp files. The ones in the
linker wrapper were not respecting the output directory because we
stripped everything with filename. Just get rid of this so it uses the
output file's directory properly in this mode.

### DIFF
--- a/clang/tools/clang-linker-wrapper/ClangLinkerWrapper.cpp
+++ b/clang/tools/clang-linker-wrapper/ClangLinkerWrapper.cpp
@@ -228,8 +228,8 @@ Expected<StringRef> createOutputFile(const Twine &Prefix, StringRef Extension) {
   if (SaveTemps) {
     (PrefixStr + "." + Extension).toNullTerminatedStringRef(OutputFile);
   } else {
-    if (std::error_code EC =
-            sys::fs::createTemporaryFile(PrefixStr, Extension, OutputFile))
+    if (std::error_code EC = sys::fs::createTemporaryFile(
+            sys::path::filename(PrefixStr), Extension, OutputFile))
       return createFileError(OutputFile, EC);
   }
 
@@ -410,8 +410,7 @@ fatbinary(ArrayRef<std::pair<StringRef, StringRef>> InputFiles,
       Args.getLastArgValue(OPT_host_triple_EQ, sys::getDefaultTargetTriple()));
 
   // Create a new file to write the linked device image to.
-  auto TempFileOrErr =
-      createOutputFile(sys::path::filename(ExecutableName), "fatbin");
+  auto TempFileOrErr = createOutputFile(ExecutableName, "fatbin");
   if (!TempFileOrErr)
     return TempFileOrErr.takeError();
 
@@ -455,8 +454,7 @@ fatbinary(ArrayRef<std::tuple<StringRef, StringRef, StringRef>> InputFiles,
     return OffloadBundlerPath.takeError();
 
   // Create a new file to write the linked device image to.
-  auto TempFileOrErr =
-      createOutputFile(sys::path::filename(ExecutableName), "hipfb");
+  auto TempFileOrErr = createOutputFile(ExecutableName, "hipfb");
   if (!TempFileOrErr)
     return TempFileOrErr.takeError();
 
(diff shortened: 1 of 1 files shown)


shiltian:
Hmm, no test?

jhuber6:
> Hmm, no test?

It's difficult to test because it wants to run a full compilation, unlike the `--dry-run` we usually use for tests.

Artem-B:
> > Hmm, no test?
> 
> It's difficult to test because it wants to run a full compilation, unlike the `--dry-run` we usually use for tests.

Can we invoke it manually with the flags clang would have passed?

jhuber6:
> Can we invoke it manually with the flags clang would have passed?

Yeah, what I mean is that with direct `clang-linker-wrapper --dry-run` it won't create any temp files at all, and without `--dry-run` it will try to execute which will expect a functioning linker.

jhuber6:
ping

yxsamliu:
Clang has two save-temps modes: --save-temps (same as --save-temps=cwd) puts files in the current working directory, while --save-temps=obj puts them next to the output file. After this change, the linker wrapper always uses the output file's directory, which matches --save-temps=obj but diverges from the default --save-temps=cwd behavior. Should we check which mode was requested and preserve the CWD behavior for the default case?

jhuber6:
> Clang has two save-temps modes: --save-temps (same as --save-temps=cwd) puts files in the current working directory, while --save-temps=obj puts them next to the output file. After this change, the linker wrapper always uses the output file's directory, which matches --save-temps=obj but diverges from the default --save-temps=cwd behavior. Should we check which mode was requested and preserve the CWD behavior for the default case?

Yeah it's a little tough because it's shared between three places now, clang, lld, and linker-wrapper. `lld` just dumps it next to the obj by default I believe. I wonder if we could make this more consistent.

yxsamliu:
> > Clang has two save-temps modes: --save-temps (same as --save-temps=cwd) puts files in the current working directory, while --save-temps=obj puts them next to the output file. After this change, the linker wrapper always uses the output file's directory, which matches --save-temps=obj but diverges from the default --save-temps=cwd behavior. Should we check which mode was requested and preserve the CWD behavior for the default case?
> 
> Yeah it's a little tough because it's shared between three places now, clang, lld, and linker-wrapper. `lld` just dumps it next to the obj by default I believe. I wonder if we could make this more consistent.

With `--save-temps` (cwd mode), users expect intermediate files in the current working directory — that's how clang's own stages behave. The linker wrapper's pre-patch code happened to conform to that expectation (by using filename(), which effectively placed files in CWD). After this patch, the linker wrapper matches lld's behavior instead, which may surprise users who find their intermediate files suddenly appearing in the output directory rather than CWD.

Would it be better to pass the full `--save-temps=cwd/--save-temps=obj` option through to the linker wrapper instead of collapsing it to a boolean, and then preserve the old (CWD) behavior for `--save-temps=cwd` while using the new (output-dir) behavior for `--save-temps=obj`? That would keep the linker wrapper consistent with how clang handles both modes. As a bonus, lld's intermediate files would follow automatically since they derive their location from the .img output path.

jhuber6:
> Would it be better to pass the full `--save-temps=cwd/--save-temps=obj` option through to the linker wrapper instead of collapsing it to a boolean, and then preserve the old (CWD) behavior for `--save-temps=cwd` while using the new (output-dir) behavior for `--save-temps=obj`? That would keep the linker wrapper consistent with how clang handles both modes. As a bonus, lld's intermediate files would follow automatically since they derive their location from the .img output path.

I'm leaning towards using `lld`'s semantics just because these are both functionally the linker. This is a bit of a strange case because normal `-save-temps` doesn't imply LLD to emit temp files, but we do that with offloading just because it's the expected behavior.

yxsamliu:
> > Would it be better to pass the full `--save-temps=cwd/--save-temps=obj` option through to the linker wrapper instead of collapsing it to a boolean, and then preserve the old (CWD) behavior for `--save-temps=cwd` while using the new (output-dir) behavior for `--save-temps=obj`? That would keep the linker wrapper consistent with how clang handles both modes. As a bonus, lld's intermediate files would follow automatically since they derive their location from the .img output path.
> 
> I'm leaning towards using `lld`'s semantics just because these are both functionally the linker. This is a bit of a strange case because normal `-save-temps` doesn't imply LLD to emit temp files, but we do that with offloading just because it's the expected behavior.

Makes sense. Aligning with lld's semantics seems to be a consistent choice for the linker wrapper. In practice, saving to output directory is usually better than saving to CWD. We may want to document -save-temps behavior for device intermediate files in clang HIP support documentation but that could be done in a separate PR.

yxsamliu (approved):
LGTM. Thanks.

PR #198680, opened 2026-05-19
[CIR] Hoist allocas out of deferred conditional cleanup scopes

andykaylor:
When a temporary alloca is created for an aggregate inside a conditional cleanup, the alloca will also be referenced in the cleanup region of the cleanup scope operation enclosing the conditional expression. In order to guarantee that the alloca dominates the cleanup region, the alloca must be hoisted out of the cleanup scope.

After CFG flattening, any allocas created at the beginning of the cleanup scope's body region would, technically, dominate the cleanup region, but because we can exit in the middle of the body region (via an exception) the body region itself does not dominate the cleanup region, and so the verifier considers any allocas within the body region as not dominating the cleanup region.

This change looks for allocas within a deferred cleanup scope and hoists them out of the cleanup scope before the cleanup region is populated.


clang/include/clang/CIR/Dialect/IR/CIROps.td
```diff
@@ -264,6 +264,28 @@ def CIR_CastOp : CIR_Op<"cast", [
...
+    //
+    // Note: array_to_pointer decay preserves the pointer value, but it changes
+    // the scope of the memory object, so we exclude it here.
+    bool isAllocaPreservingCast() {
+      if (!mlir::isa<cir::PointerType>(getSrc().getType()))
+        return false;
+      switch (getKind()) {
+      case cir::CastKind::bitcast:
+      case cir::CastKind::address_space:
+        return true;
```
erichkeane, line 282: Is this a complete list?  Or are we expecting to have to update this in the future? There are a lot of cast kinds, so shocked to see only 2 in the 'true' case here.

erichkeane (approved):

clang/include/clang/CIR/Dialect/IR/CIROps.td
andykaylor: This is all I'm expecting, and even the bitcast is just here because it's likely to show up in an `Address` if we use `withElementType`. This PR only needs the address_space cast. Most of the others don't apply to pointers. Technically, I guess we could have `ptr_to_int` here, but I'm not sure how useful that would be.

PR #198681, opened 2026-05-19
[Offload] Fix llvm-gpu-loader after new argument

jhuber6:
Summary:
New argument added but was not included here. This gets out of sync just
because there's no reasonable way to build offload and this tool, it's
an unsolvable phase ordering problem so we just dlopen it.


Artem-B (approved):

adurang:
I get that we didn't get a fail when compiling because of the dependence issue you described, but shouldn't some test have failed post-commit? Are we missing some testing for this maybe?

jhuber6:
> I get that we didn't get a fail when compiling because of the dependence issue you described, but shouldn't some test have failed post-commit? Are we missing some testing for this maybe?

Some libc tests failed  but it was spurious because it would pass if the memory happened to be zeroed.

PR #198682, opened 2026-05-20
[libc] Implement pthread_sigmask

vonosmas:
* Extract `rt_sigprocmask` syscall wrapper into the libc/src/__support/OSUtil/linux/syscall_wrappers/ directory
* Convert all existing users of this syscall, and simplify the logic where applicable.
* Implement `pthread_sigmask`, which is effectively another POSIX wrapper around `rt_sigprocmask` syscall similar to `sigprocmask`

### DIFF
--- a/libc/include/signal.yaml
+++ b/libc/include/signal.yaml
@@ -114,4 +114,3 @@ functions:
       - type: int
       - type: const sigset_t *__restrict
       - type: sigset_t *__restrict
-
--- a/libc/include/signal.yaml
+++ b/libc/include/signal.yaml
@@ -108,7 +108,7 @@ functions:
       - type: sigset_t *__restrict
   - name: pthread_sigmask
     standards:
-      - POSIX
+      - posix
     return_type: int
     arguments:
       - type: int
--- a/libc/include/signal.yaml
+++ b/libc/include/signal.yaml
@@ -106,3 +106,12 @@ functions:
       - type: int
       - type: const sigset_t *__restrict
       - type: sigset_t *__restrict
+  - name: pthread_sigmask
+    standards:
+      - POSIX
+    return_type: int
+    arguments:
+      - type: int
+      - type: const sigset_t *__restrict
+      - type: sigset_t *__restrict
+
--- a/libc/src/signal/CMakeLists.txt
+++ b/libc/src/signal/CMakeLists.txt
@@ -38,6 +38,14 @@ add_entrypoint_object(
 )
 
 add_entrypoint_object(
+  pthread_sigmask
+  ALIAS
+  DEPENDS
+    .${LIBC_TARGET_OS}.pthread_sigmask
+)
+
+
+add_entrypoint_object(
   sigemptyset
   ALIAS
   DEPENDS
--- a/libc/src/signal/CMakeLists.txt
+++ b/libc/src/signal/CMakeLists.txt
@@ -44,7 +44,6 @@ add_entrypoint_object(
     .${LIBC_TARGET_OS}.pthread_sigmask
 )
 
-
 add_entrypoint_object(
   sigemptyset
   ALIAS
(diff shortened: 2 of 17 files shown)


libc/test/src/signal/pthread_sigmask_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+  sigset_t valid;
+  // 17 and -4 are out of the range for pthread_sigmask's how parameter.
+  EXPECT_EQ(LIBC_NAMESPACE::pthread_sigmask(17, &valid, nullptr), EINVAL);
+  EXPECT_EQ(LIBC_NAMESPACE::pthread_sigmask(-4, &valid, nullptr), EINVAL);
+
+  // This pointer is out of this process's address range.
+  sigset_t *invalid = reinterpret_cast<sigset_t *>(-1);
+  EXPECT_EQ(LIBC_NAMESPACE::pthread_sigmask(SIG_SETMASK, invalid, nullptr),
+            EFAULT);
+  EXPECT_EQ(LIBC_NAMESPACE::pthread_sigmask(-4, nullptr, invalid), EFAULT);
```
labath: The exact error returned here would depend on which (invalid) argument the kernel validates first, and I could imagine it changing between kernel versions. I'd recommend sticking to testing one invalid argument at a time.

labath (approved):
LGTM.
 
IIUC, we don't (yet?) implement the features that require reserving a signal for internal usage, and consequently we don't prevent applications from blocking those signals. If/when start reserving signal numbers, this will need to change to hide those signals from theu user.

libc/include/signal.yaml
```diff
@@ -106,3 +106,11 @@ functions:
       - type: int
       - type: const sigset_t *__restrict
       - type: sigset_t *__restrict
+  - name: pthread_sigmask
+    standards:
+      - POSIX
```
kaladron:
```suggestion
      - posix
```

libc/test/src/signal/pthread_sigmask_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "src/signal/sigaddset.h"
+#include "src/signal/sigemptyset.h"
+#include "test/UnitTest/ErrnoCheckingTest.h"
+#include "test/UnitTest/Test.h"
+
+#include <signal.h>
+
+class LlvmLibcPthreadSigmaskTest
+    : public LIBC_NAMESPACE::testing::ErrnoCheckingTest {
+  sigset_t oldSet;
```
kaladron: I think this should be old_set?  (I'm still getting the various cases right myself...)

libc/src/signal/CMakeLists.txt
```diff
@@ -37,6 +37,13 @@ add_entrypoint_object(
     .${LIBC_TARGET_OS}.sigprocmask
 )
 
+add_entrypoint_object(
+  pthread_sigmask
+  ALIAS
+  DEPENDS
+    .${LIBC_TARGET_OS}.pthread_sigmask
```
kaladron, line 44: libc.src.__support.common
libc.src.__support.macros.config

kaladron (approved):

libc/test/src/signal/pthread_sigmask_test.cpp
vonosmas: Done

libc/include/signal.yaml
vonosmas: Done

libc/test/src/signal/pthread_sigmask_test.cpp
vonosmas: Done here for brevity. However, technically speaking the kernel should not even validate the first argument ("how") if `set` is `nullptr` -- the only thing pthread_sigmask would do is return the old set of signals in this case.

PR #198684, opened 2026-05-20
[CI] Successful build and no tests running is now a notification

boomanaiden154:
check-libc now uses llvm-lit to run tests instead of running the unittests directly through ninja. This means there should not be any cases in tree where the build could succeed but we do not pick up any tests as running. Still pass the build in this case because if everything passes with exit code 0 it is wrong not to, but make a note to the user that this is unexpected.

### DIFF
--- a/.ci/generate_test_report_lib_test.py
+++ b/.ci/generate_test_report_lib_test.py
@@ -196,7 +196,7 @@ class TestReports(unittest.TestCase):
                     """\
                     # Foo
 
-                    :white_check_mark: The build succeeded and no tests ran. This is not expected and likely means something was configured incorrectly. If you suspect an infrastructure issue, please open an issue at https://github.com/llvm/llvm-project/issues and attach theinfrastructure label."""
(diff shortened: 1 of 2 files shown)


.ci/generate_test_report_lib_test.py
```diff
@@ -194,9 +194,9 @@ def test_title_only(self):
             (
                 dedent(
                     """\
-                # Foo
+                    # Foo
 
-                :white_check_mark: The build succeeded and no tests ran. This is expected in some build configurations."""
+                    :white_check_mark: The build succeeded and no tests ran. This is not expected and likely means something was configured incorrectly. If you suspect an infrastructure issue, please open an issue at https://github.com/llvm/llvm-project/issues and attach theinfrastructure label."""
```
cmtice: Tiny nit: should be a space between "the" and "infrastructure:.  I'm approving this change, but please fix this before committing.

cmtice (approved):

.ci/generate_test_report_lib_test.py
cmtice: Maybe same issue in text in other file as well.

PR #198685, opened 2026-05-20
[RISCV] P-ext support v8i8/v4i16/v2i32 setcc on RV32.

topperc:


llvm/lib/Target/RISCV/RISCVInstrInfoP.td
```diff
@@ -2341,6 +2341,19 @@ let Predicates = [HasStdExtP, IsRV32] in {
   // 32-bit arithmetic shift left/right
   def : PatGprPairShiftMask<riscv_psra, PSRA_DWS, shiftMask32, v2i32>;
 
+  // 8/16/31-bit comparison patterns (result is all 1s or all 0s per element)
```
sihuan:
```suggestion
  // 8/16/32-bit comparison patterns (result is all 1s or all 0s per element)

sihuan (approved):
LGTM. Left a small typo nit inline.

llvm/lib/Target/RISCV/RISCVInstrInfoP.td
topperc: Thanks!

PR #198687, opened 2026-05-20
[mlir] Specify isSigned when creating APInt for I16 attributes

apaszke:
8- and 32-bit attribute constructors already do this. Debug builds trigger an assertion unless isSigned is specified.


apaszke:
It looks like the test failure is unrelated to my change

jpienaar (approved):

PR #198689, opened 2026-05-20
Revert "[Offload] Add support for measuring elapsed time between events (#186856)"

lfmeadow:
Reverts #186856.

Not performance-neutral (e.g. BabelStream); needs to be opt-in.

cc @leandrolcampos @jhuber6 @kevinsala @ronlieb

Made with [Cursor](https://cursor.com)


ronlieb (approved):

jhuber6:
I don't think it would be very difficult to make the event profiling opt-in with a flag, the API isn't stable yet, but it's probably best to not regress things.

PR #198690, opened 2026-05-20
[OpenMP][OMPIRBuilder] Avoid querying SmallPtrSet during removal

chichunchen:
openmp-cli-fuse02.mlir can intermittently leave behind a dead block after loop fusion, causing LLVM IR verification to fail. (https://github.com/llvm/llvm-project/pull/197637#issuecomment-4497502486)

This happened because `removeUnusedBlocksFromParent` queried `BBsToErase` while using `SmallPtrSet::remove_if` on the same set, which made the result depend on the set's internal mutation order during removal.

This patch tracks candidate blocks by index with `SmallBitVector` instead of mutating and querying the same `SmallPtrSet`. This also preserves the original `BBs` order when collecting the final dead blocks.

---

The original PR was created with assistance from Copilot. The later analysis and fix were revised based on reviewer feedback.

Co-authored-by: @slinder1

### DIFF
--- a/llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
+++ b/llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
@@ -6560,38 +6560,41 @@ static void redirectAllPredecessorsTo(BasicBlock *OldTarget,
 /// Determine which blocks in \p BBs are reachable from outside and remove the
 /// ones that are not reachable from the function.
 static void removeUnusedBlocksFromParent(ArrayRef<BasicBlock *> BBs) {
-  SmallPtrSet<BasicBlock *, 6> BBsToErase(llvm::from_range, BBs);
-  auto HasRemainingUses = [&BBsToErase](BasicBlock *BB) {
-    for (Use &U : BB->uses()) {
-      auto *UseInst = dyn_cast<Instruction>(U.getUser());
-      if (!UseInst)
-        continue;
-      if (BBsToErase.count(UseInst->getParent()))
-        continue;
-      return true;
+  DenseMap<BasicBlock *, unsigned> BB2Index(BBs.size());
+  for (const auto [I, BB] : enumerate(BBs))
+    BB2Index.insert({BB, I});
+
+  // Optimistically assume all BBs are dead. We iteratively remove
+  // immediately-used blocks from this set until we reach a fixed point.
+  SmallBitVector BBsToErase(BBs.size(), true);
+  // Accumulates immediately-used blocks for one iteration.
+  SmallBitVector BBsToKeep(BBs.size());
+  do {
+    BBsToKeep.reset();
+    for (unsigned BBToEraseIndex : BBsToErase.set_bits()) {
+      for (Use &U : BBs[BBToEraseIndex]->uses()) {
+        auto *UseInst = dyn_cast<Instruction>(U.getUser());
+        if (!UseInst)
+          continue;
+        BasicBlock *UseBB = UseInst->getParent();
+        auto UseBBIndexI = BB2Index.find(UseBB);
+        // If this use is from an external block, or from a block in BBs we have
+        // already proved is transitively used by an external block, then this
+        // block is also used.
+        if (UseBBIndexI == BB2Index.end() ||
+            !BBsToErase.test(UseBBIndexI->second)) {
+          BBsToKeep.set(BBToEraseIndex);
+          break;
+        }
+      }
     }
-    return false;
-  };
+    BBsToErase &= ~BBsToKeep;
+  } while (BBsToKeep.any());
 
-  // HasRemainingUses queries BBsToErase, so it cannot be used as a predicate
-  // for SmallPtrSet::remove_if on the same set.
-  SmallVector<BasicBlock *> BBsToKeep;
-  while (true) {
-    bool Changed = false;
-    BBsToKeep.clear();
-
-    for (BasicBlock *BB : BBsToErase)
-      if (HasRemainingUses(BB))
-        BBsToKeep.push_back(BB);
-    for (BasicBlock *BB : BBsToKeep)
-      Changed |= BBsToErase.erase(BB);
-
-    if (!Changed)
-      break;
-  }
-
-  SmallVector<BasicBlock *, 7> BBVec(BBsToErase.begin(), BBsToErase.end());
-  DeleteDeadBlocks(BBVec);
+  SmallVector<BasicBlock *> FinalBBsToErase;
+  for (unsigned BBToEraseIndex : BBsToErase.set_bits())
+    FinalBBsToErase.push_back(BBs[BBToEraseIndex]);
+  DeleteDeadBlocks(FinalBBsToErase);
 }
 
 CanonicalLoopInfo *
--- a/llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
+++ b/llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
@@ -6573,17 +6573,21 @@ static void removeUnusedBlocksFromParent(ArrayRef<BasicBlock *> BBs) {
     return false;
   };
 
-  // Compute the keep set before mutating BBsToErase; otherwise the result can
-  // depend on SmallPtrSet iteration order.
-  bool Changed = true;
-  while (Changed) {
-    Changed = false;
-    SmallVector<BasicBlock *, 4> BBsToKeep;
+  // HasRemainingUses queries BBsToErase, so it cannot be used as a predicate
+  // for SmallPtrSet::remove_if on the same set.
+  SmallVector<BasicBlock *> BBsToKeep;
+  while (true) {
+    bool Changed = false;
+    BBsToKeep.clear();
+
     for (BasicBlock *BB : BBsToErase)
       if (HasRemainingUses(BB))
         BBsToKeep.push_back(BB);
     for (BasicBlock *BB : BBsToKeep)
       Changed |= BBsToErase.erase(BB);
+
+    if (!Changed)
+      break;
   }
 
   SmallVector<BasicBlock *, 7> BBVec(BBsToErase.begin(), BBsToErase.end());
--- a/llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
+++ b/llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
@@ -6573,8 +6573,17 @@ static void removeUnusedBlocksFromParent(ArrayRef<BasicBlock *> BBs) {
     return false;
   };
 
-  while (BBsToErase.remove_if(HasRemainingUses)) {
-    // Try again if anything was removed.
+  // Compute the keep set before mutating BBsToErase; otherwise the result can
+  // depend on SmallPtrSet iteration order.
+  bool Changed = true;
+  while (Changed) {
+    Changed = false;
+    SmallVector<BasicBlock *, 4> BBsToKeep;
+    for (BasicBlock *BB : BBsToErase)
+      if (HasRemainingUses(BB))
+        BBsToKeep.push_back(BB);
+    for (BasicBlock *BB : BBsToKeep)
+      Changed |= BBsToErase.erase(BB);
   }
 
   SmallVector<BasicBlock *, 7> BBVec(BBsToErase.begin(), BBsToErase.end());
--- a/llvm/unittests/Frontend/OpenMPIRBuilderTest.cpp
+++ b/llvm/unittests/Frontend/OpenMPIRBuilderTest.cpp
@@ -8461,6 +8461,9 @@ TEST_F(OpenMPIRBuilderTest, ScopeDirectiveNowait) {
   }
 }
 
+// This test ensures loop fusion does not leave orphaned non-entry blocks.
+// It's a regression test for SmallPtrSet::remove_if behavior change from
+// #197637.
 TEST_F(OpenMPIRBuilderTest, FuseLoopRangeNoDeadBlocks) {
   using InsertPointTy = OpenMPIRBuilder::InsertPointTy;
   OpenMPIRBuilder OMPBuilder(*M);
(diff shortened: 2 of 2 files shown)


llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
```diff
@@ -6573,8 +6573,17 @@ static void removeUnusedBlocksFromParent(ArrayRef<BasicBlock *> BBs) {
...
+  bool Changed = true;
+  while (Changed) {
+    Changed = false;
+    SmallVector<BasicBlock *, 4> BBsToKeep;
+    for (BasicBlock *BB : BBsToErase)
+      if (HasRemainingUses(BB))
+        BBsToKeep.push_back(BB);
+    for (BasicBlock *BB : BBsToKeep)
+      Changed |= BBsToErase.erase(BB);
   }
```
slinder1:
```suggestion
  // Compute the keep set before mutating BBsToErase; otherwise the result can
  // depend on SmallPtrSet iteration order.
  bool Changed = false;
  do {
    SmallVector<BasicBlock *, 4> BBsToKeep;
    for (BasicBlock *BB : BBsToErase)
      if (HasRemainingUses(BB))
        BBsToKeep.push_back(BB);
    for (BasicBlock *BB : BBsToKeep)
      Changed |= BBsToErase.erase(BB);
  } while (Changed);
```

slinder1 (commented):
I just started hitting this, so thank you for having already posted a patch!

I reviewed your fix first, but then thought to look through some `blame`, and was confused to see no recent changes here. Then I took a look at `SmallPtrSet`s `blame` and noticed that the contract (or at least the documentation of the contract) for `remove_if` changed in #197637 from:

```
It is safe to read the set inside
the predicate function. However, the predicate must not modify the set
itself, only indicate a removal by returning true.
```

to:

```
The predicate must not access the
set being modified: it may inspect the element passed to it and return
true to request removal, but must not read (e.g. count()/find()) or
otherwise mutate the set. If anything is removed, all iterators and
references into the set are invalidated.
```

Which seems like it is the real root cause? @MaskRay is this kind of fallout possible in other places in the codebase, or is `remove_if` generally not used in this way? Did the contract actually change, or was the old comment just wrong?

Meinersbur:
Can you help me understand the problem?

The patch is based on a commit before #197637 where SmallPtrSet::remove_if has nothing that would make the data structure invalid between calls of the predicate. I could not make your test fail without your change in OpenMPIRBuilder. 
The algorithm is supposed to be a fixpoint iteration: It should not matter which order BBs are iterated over; if a BB is still marked for deletion in one round due to different iteration order, it will be unmarked in the next.

aengelke:
> Did the contract actually change,

Yes. I think this was an oversight, as the line is explicitly mentioned in one of the comments..

The explanation in the PR is wrong. The problem is that during the execution of SmallPtrSet::remove_if(), lookup in the set doesn't work correctly until the rehash at the end of remove_if.

llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
Meinersbur: You suggestion never resets `Change` to false, causing an infinity loop.

If we are going to nitpick in this, what I would prefer is
```cpp
  SmallVector<BasicBlock *> BBsToKeep;
  while (true) {
    bool Changed = false;
    BBsToKeep.clear();

    for (BasicBlock *BB : BBsToErase)
      if (HasRemainingUses(BB))
        BBsToKeep.push_back(BB);
    for (BasicBlock *BB : BBsToKeep)
      Changed |= BBsToErase.erase(BB);

    if (!Changed)
      break;
  }
```
such that `Changed` remains scoped in the loop body.

Declare vectors outside of the loop: https://llvm.org/docs/ProgrammersManual.html#vector

It is recommended to not specify a small size for vectors on the stack: https://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h

slinder1:
> > Did the contract actually change,
> 
> Yes. I think this was an oversight, as the line is explicitly mentioned in one of the comments..
> 
> The explanation in the PR is wrong. The problem is that during the execution of SmallPtrSet::remove_if(), lookup in the set doesn't work correctly until the rehash at the end of remove_if.

That makes sense, thank you!

It seems like copilot may have helped find this (now) broken use of `remove_if`, but the stochastically reverse-engineered "analysis" of it is junk (go figure)

If the `remove_if` change is purely for performance can we just revert for now?

Edit: I'm just now reading your comment on the other change. I will comment there instead

llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
slinder1: I agree, ignore my original comment here. I was working too late, and didn't check my own suggestion carefully, sorry about that!

slinder1:
Since the input is a small array of `BasicBlock`s (otherwise it probably makes more sense to just do a DFS/simple DCE?), we could also work in terms of indices into `BBs` and use `SmallBitVector`s:
```cpp
/// Determine which blocks in \p BBs are reachable from outside and remove the
/// ones that are not reachable from the function.
static void removeUnusedBlocksFromParent(ArrayRef<BasicBlock *> BBs) {
  DenseMap<BasicBlock *, unsigned> BB2Index(BBs.size());
  for (const auto [I, V] : enumerate(BBs))
    BB2Index.insert({V, I});

  // Optimistically assume all BBs are dead. We iteratively remove
  // immediately-used blocks from this set until we reach a fixed point.
  SmallBitVector BBsToErase(BBs.size(), true);
  // Accumulates immediately-used blocks for one iteration.
  SmallBitVector BBsToKeep(BBs.size());
  do {
    BBsToKeep.reset();
    for (unsigned BBToEraseIndex : BBsToErase.set_bits()) {
      for (Use &U : BBs[BBToEraseIndex]->uses()) {
        auto *UseInst = dyn_cast<Instruction>(U.getUser());
        if (!UseInst)
          continue;
        BasicBlock *UseBB = UseInst->getParent();
        auto UseBBIndexI = BB2Index.find(UseBB);
        // If this use is from an external block, or from a block in BBs we have
        // already proved is transitively used by an external block, then
        // this block is also used.
        if (UseBBIndexI == BB2Index.end() ||
            !BBsToErase.test(UseBBIndexI->second)) {
          BBsToKeep.set(BBToEraseIndex);
          break;
        }
      }
    }
    BBsToErase &= ~BBsToKeep;
  } while (BBsToKeep.any());

  SmallVector<BasicBlock *> FinalBBsToErase;
  for (unsigned BBToEraseIndex : BBsToErase.set_bits())
    FinalBBsToErase.push_back(BBs[BBToEraseIndex]);
  DeleteDeadBlocks(FinalBBsToErase);
}

```

slinder1 (approved):
The current version LGTM, and landing something so the buildbots aren't randomly failing would be good ASAP

If there is more review we can do it after. Thanks again!

Edit: I had missed the previous revision when I responded with the `SmallBitVector` suggestion, but either version LGTM!

llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
```diff
@@ -6560,25 +6560,41 @@ static void redirectAllPredecessorsTo(BasicBlock *OldTarget,
...
+  SmallBitVector BBsToKeep(BBs.size());
+  do {
+    BBsToKeep.reset();
+    for (unsigned BBToEraseIndex : BBsToErase.set_bits()) {
+      for (Use &U : BBs[BBToEraseIndex]->uses()) {
+        auto *UseInst = dyn_cast<Instruction>(U.getUser());
+        if (!UseInst)
+          continue;
+        BasicBlock *UseBB = UseInst->getParent();
+        auto UseBBIndexI = BB2Index.find(UseBB);
```
Meinersbur, line 6580: [style] [Don’t “almost always” use auto](https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable)

Meinersbur (commented):
This looks like AI-overcomplexity. `BBsToKeep` is not even needed because it's finding a fixpoint, just reset that bit in `BBsToErase` directly. In constrast to SmallPtrSet you can actually do that while iterating over it. Also, building the DenseMap over a small list of an estimated 7 elements max is probably overkill.

Here someting simpler by just inverting BBsToErase to BBsToKeep:
```cpp
static void removeUnusedBlocksFromParent(ArrayRef<BasicBlock *> BBs) {
  SmallPtrSet<BasicBlock *, 6> BBsToKeep(llvm::from_range, BBs);
  auto HasRemainingUses = [&BBsToKeep](BasicBlock *BB) {
    for (Use &U : BB->uses()) {
      auto *UseInst = dyn_cast<Instruction>(U.getUser());
      if (!UseInst)
        continue;
      if (!BBsToKeep.count(UseInst->getParent()))
        continue;
      return true;
    }
    return false;
  };

  while (true) {
    bool Changed = false;

    for (BasicBlock *BB : BBs) {
      if (BBsToKeep.count(BB) || !HasRemainingUses(BB)) 
        continue;

      BBsToKeep.insert(BB);
      Changed = true;
    }

    if (!Changed)
      break;
  }

  SmallVector<BasicBlock *, 7> BBVec;
  for (BasicBlock *BB : BBs) {
    if (!BBsToKeep.count(BB))
      BBVec.push_back(BB);
  }
  DeleteDeadBlocks(BBVec);
}
```

llvm/lib/Frontend/OpenMP/OMPIRBuilder.cpp
slinder1: Do you prefer `DenseMap<BasicBlock *, unsigned>::iterator UseBBIndexI`?

This isn't about "almost always" using auto, iterator types are the canonical example of why `auto` is helpful for readability.

slinder1:
> This looks like AI-overcomplexity. `BBsToKeep` is not even needed because it's finding a fixpoint, just reset that bit in `BBsToErase` directly. In constrast to SmallPtrSet you can actually do that while iterating over it. Also, building the DenseMap over a small list of an estimated 7 elements max is probably overkill.
> 
> Here someting simpler by just inverting BBsToErase to BBsToKeep:
> 
> ```c++
> static void removeUnusedBlocksFromParent(ArrayRef<BasicBlock *> BBs) {
>   SmallPtrSet<BasicBlock *, 6> BBsToKeep(llvm::from_range, BBs);
>   auto HasRemainingUses = [&BBsToKeep](BasicBlock *BB) {
>     for (Use &U : BB->uses()) {
>       auto *UseInst = dyn_cast<Instruction>(U.getUser());
>       if (!UseInst)
>         continue;
>       if (!BBsToKeep.count(UseInst->getParent()))
>         continue;
>       return true;
>     }
>     return false;
>   };
> 
>   while (true) {
>     bool Changed = false;
> 
>     for (BasicBlock *BB : BBs) {
>       if (BBsToKeep.count(BB) || !HasRemainingUses(BB)) 
>         continue;
> 
>       BBsToKeep.insert(BB);
>       Changed = true;
>     }
> 
>     if (!Changed)
>       break;
>   }
> 
>   SmallVector<BasicBlock *, 7> BBVec;
>   for (BasicBlock *BB : BBs) {
>     if (!BBsToKeep.count(BB))
>       BBVec.push_back(BB);
>   }
>   DeleteDeadBlocks(BBVec);
> }
> ```

No AI used in this case, I'm just overcomplicating things. I was biased towards the likely LLM generated original version, though

slinder1:
I think these are just artifacts of you giving the gist of the approach quickly, but I don't think your code works as-is:

```
static void removeUnusedBlocksFromParent(ArrayRef<BasicBlock *> BBs) {
  SmallPtrSet<BasicBlock *, 6> BBsToKeep(llvm::from_range, BBs);
```
I assume this should start empty?
```
  auto HasRemainingUses = [&BBsToKeep](BasicBlock *BB) {
    for (Use &U : BB->uses()) {
      auto *UseInst = dyn_cast<Instruction>(U.getUser());
      if (!UseInst)
        continue;
      if (!BBsToKeep.count(UseInst->getParent()))
        continue;
      return true;
```
Assuming a block is in `BBsToKeep` iff we know it has an external use, this seems slightly wrong. Isn't it `if (<UseBB is not in BBs> || BBsToKeep.count(UseBB)) return true;`?
```
    }
    return false;
  };

  while (true) {
    bool Changed = false;

    for (BasicBlock *BB : BBs) {
      if (BBsToKeep.count(BB) || !HasRemainingUses(BB)) 
        continue;
```
This is the only use of `HasRemainingUses` anyway, and it isn't passed as a callback, so we probably don't need the lambda, it could just be inline. I guess the `return` from the lambda is nicer than assign-and-break? If the lambda remains then the `Remaining` in the name no longer fits.
```

      BBsToKeep.insert(BB);
      Changed = true;
    }

    if (!Changed)
      break;
  }

  SmallVector<BasicBlock *, 7> BBVec;
  for (BasicBlock *BB : BBs) {
    if (!BBsToKeep.count(BB))
      BBVec.push_back(BB);
  }
  DeleteDeadBlocks(BBVec);
}
```

slinder1:
I think this works, and is what you were suggesting?

Some other incidental changes I made are because the `SmallPtrSet` count is rounded up to a power of two anyway, and `SmallVector` generally doesn't benefit from a count. I also removed an extra set of braces.

```cpp
static void removeUnusedBlocksFromParent(ArrayRef<BasicBlock *> BBs) {
  SmallPtrSet<BasicBlock *, 8> InternalBBs(from_range, BBs);
  SmallPtrSet<BasicBlock *, 8> BBsToKeep;

  while (true) {
    bool Changed = false;

    for (BasicBlock *BB : BBs) {
      if (BBsToKeep.contains(BB))
        continue;

      for (Use &U : BB->uses()) {
        auto *UseInst = dyn_cast<Instruction>(U.getUser());
        if (!UseInst)
          continue;
        BasicBlock *UseBB = UseInst->getParent();
        if (!InternalBBs.contains(UseBB) || BBsToKeep.contains(UseBB)) {
          BBsToKeep.insert(BB);
          Changed = true;
          break;
        }
      }
    }

    if (!Changed)
      break;
  }

  SmallVector<BasicBlock *> BBVec;
  for (BasicBlock *BB : BBs)
    if (!BBsToKeep.count(BB))
      BBVec.push_back(BB);

  DeleteDeadBlocks(BBVec);
}
````

PR #198692, opened 2026-05-20
[flang-rt] Silence -Wfenv-access in the runtime fenv wrappers (part 1)

eugeneepshteyn:
Upstream clang's commit 5f2bedca, PR #187860, enabled a warning that fires whenever code calls a `<fenv.h>` / `<cfenv>` primitive without enabling proper FP exception behavior. This caused warnings in compiling flang-rt, which in turn caused certain buildsbots to fail with `-Werror`.

This change enables FP trap behavior for parts of flang-rt that are known to require it. The other parts will be addressed by a future change.

Assisted-by: AI

### DIFF
--- a/flang-rt/lib/runtime/exceptions.cpp
+++ b/flang-rt/lib/runtime/exceptions.cpp
@@ -96,8 +96,8 @@ void RTNAME(feclearexcept)(uint32_t excepts) {
 #endif
 }
 void RTDEF(feraiseexcept)(uint32_t excepts) {
-  FLANG_FP_TRAP_ON
 #if !defined(RT_DEVICE_COMPILATION)
+  FLANG_FP_TRAP_ON
   feraiseexcept(excepts);
 #if defined(_MM_EXCEPT_DENORM)
   _mm_setcsr(_mm_getcsr() | (excepts & _MM_EXCEPT_MASK));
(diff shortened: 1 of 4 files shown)


eugeneepshteyn:
Tagging @vntkmr , who is working on FP trap improvements in flang.

flang-rt/lib/runtime/exceptions.cpp
```diff
@@ -88,12 +89,14 @@ uint32_t RTDEF(MapException)(uint32_t excepts) {
...
 
 void RTNAME(feclearexcept)(uint32_t excepts) {
+  FLANG_FP_TRAP_ON
   feclearexcept(excepts);
 #if defined(_MM_EXCEPT_DENORM)
   _mm_setcsr(_mm_getcsr() & ~(excepts & _MM_EXCEPT_MASK));
 #endif
 }
 void RTDEF(feraiseexcept)(uint32_t excepts) {
+  FLANG_FP_TRAP_ON
```
vzakhari: nit: is it possible to move it under the `#if` below?

vzakhari (approved):
Looks great! Thank you, Eugene!

flang-rt/lib/runtime/exceptions.cpp
eugeneepshteyn: Hmm, should all `FLANG_FP_TRAP_ON` only happen for host compilation?

eugeneepshteyn: No, it's probably ok just for this one. I should ensure there's no issue building for device compilation.

PR #194639, opened 2026-04-28
[CodeGen] Teach ReplaceWithVeclib split vector llvm.sincos when only sin/cos veclib mappings exist

kito-cheng:
Some vector math libraries provide vector sin and vector cos but no vector sincos or no sincos with an ABI that LLVM can emit.

The one of the common case is glibc libmvec on x86: it exposes `_ZGV{b,c,d,e}N{2,4,8,16}vvv_sincos{,f}` symbols, but those use a vectors-of-pointers output ABI that expandMultipleResultFPLibCall does not currently support. As a result, sincos will falls back to scalar sincos calls even when the target has a fully working vector sin and vector cos.

So we trying to split it into separate sin and cos calls, which will then be replaced with vector calls if the target supports it, it generally better than scalarized sincos calls.

PR #198693, opened 2026-05-20
[ARM][test] Replace removed @llvm.arm.neon.vshiftn calls with lshr+trunc [NFC]

kito-cheng:
@llvm.arm.neon.vshiftn was removed in 170daafe015f (2014), which intentionally replaced the intrinsic with the plain shift+truncate IR sequence. The 2012-08-27-CopyPhysRegCrash.ll regression test was added before that change and still calls the removed intrinsic; the call silently lowers to an external call because @llvm.arm.neon.vshiftn no longer corresponds to any known intrinsic.

Rewrite the two call sites with the lshr+trunc form the removal commit documents as the canonical replacement. The test only checks that codegen does not crash, so behaviour is preserved.


kito-cheng:

```
```

davemgreen (approved):
LGTM - thanks

PR #198695, opened 2026-05-20
[alpha.webkit.UncountedLocalVarsChecker] Detect an assignment to a guardian argument

rniwa:
A function parameter of type RefPtr<T>& should not be used as a guardian variable of a raw pointer/reference variable if the function body contains an assignment to it since such an assignment can shorten the lifetime of the guarded object.


clang/lib/StaticAnalyzer/Checkers/WebKit/RawPtrRefLocalVarsChecker.cpp
```diff
@@ -340,8 +340,18 @@ class RawPtrRefLocalVarsChecker
 
                     // Parameters are guaranteed to be safe for the duration of
                     // the call by another checker.
```
t-rasmud: Is this comment still relevant?

rniwa: oh, I guess maybe we should rephrase it.

rniwa: Actually, let's just remove this. There is no need to say it may change.

clang/test/Analysis/Checkers/WebKit/uncounted-local-vars.cpp
```diff
@@ -343,6 +343,19 @@ void baz() {
...
+namespace local_assignment_to_guardian_parameter {
+
+RefCountable *provide_ref_cntbl();
+
+void foo(RefPtr<RefCountable>& arg) {
+  RefCountable* ptr = arg.get();
+  // expected-warning@-1{{Local variable 'ptr' is uncounted and unsafe [alpha.webkit.UncountedLocalVarsChecker]}}
+  arg = nullptr;
+  ptr->method();
+}
```
steakhal, line 397: What would the safe alternative look like? Would we need to make a deep-copy of the `arg`?

rniwa: I think the fix here will be the function to take RefCountable& so that the caller keeps this object alive and then maybe define a local variable which is of type Ref<RefCountable> that gets re-assigned.

clang/lib/StaticAnalyzer/Checkers/WebKit/RawPtrRefLocalVarsChecker.cpp
```diff
@@ -338,10 +338,18 @@ class RawPtrRefLocalVarsChecker
...
 
-                    // Parameters are guaranteed to be safe for the duration of
-                    // the call by another checker.
-                    if (isa<ParmVarDecl>(MaybeGuardian))
-                      return true;
+                    if (isa<ParmVarDecl>(MaybeGuardian)) {
+                      if (auto *FD = dyn_cast<FunctionDecl>(DeclWithIssue)) {
+                        GuardianVisitor guardianVisitor(MaybeGuardian);
+                        if (guardianVisitor.TraverseStmt(FD->getBody()))
+                          return true;
```
steakhal:
```suggestion
                        if (GuardianVisitor{MaybeGuardian}.TraverseStmt(FD->getBody()))
                          return true;
```

clang/lib/StaticAnalyzer/Checkers/WebKit/RawPtrRefLocalVarsChecker.cpp
```diff
@@ -338,10 +338,18 @@ class RawPtrRefLocalVarsChecker
...
+                      }
+                      if (auto *MD = dyn_cast<ObjCMethodDecl>(DeclWithIssue)) {
+                        GuardianVisitor guardianVisitor(MaybeGuardian);
+                        if (guardianVisitor.TraverseStmt(MD->getBody()))
+                          return true;
```
steakhal:
```suggestion
                        if (GuardianVisitor{MaybeGuardian}.TraverseStmt(MD->getBody()))
                          return true;
```

clang/lib/StaticAnalyzer/Checkers/WebKit/RawPtrRefLocalVarsChecker.cpp
steakhal: BTW it's weird to me to see using the return value of `TraverseStmt` - I figured that determines if the traversal was aborted or not.
But I suppose that's the signal you desired here. I'm just raising that it was unusual to see this.

steakhal (approved):

t-rasmud (approved):
LGTM

PR #198696, opened 2026-05-20
[AMDGPU][test] Replace invalid intrinsic calls in two tests [NFC]

kito-cheng:
These tests reference intrinsics that do not exist:

- sched.barrier.inverted.mask.ll calls @llvm.amdcn.s.nop, a typo of @llvm.amdgcn.s.nop that has been in the test since it was added in f1156fb622a7.
- si-split-load-store-alias-info.ll calls @llvm.amdgcn.wmma.f32.16x16x16.f32, which has never existed (only the .f16/.bf16/.fp8/.bf8 input variants are defined). The intrinsic was only used to keep the loaded value alive between the load and the store; an fadd serves the same purpose while exercising the same alias-info propagation path being tested.

The dyn_cast<IntrinsicInst> we rely on in passes such as ReplaceWithVeclib only checks the "llvm." prefix and does not validate the intrinsic name, so these calls have silently become IntrinsicInst with Intrinsic::not_intrinsic. Fix the tests so that strengthening the check does not turn into a spurious regression.

Assisted-by: Opus 4.7

### DIFF
--- a/llvm/test/CodeGen/AMDGPU/sched.barrier.inverted.mask.ll
+++ b/llvm/test/CodeGen/AMDGPU/sched.barrier.inverted.mask.ll
@@ -10,7 +10,7 @@
 define amdgpu_kernel void @invert1() #0 {
 entry:
   call void @llvm.amdgcn.sched.barrier(i32 1) #1
-  call void @llvm.amdcn.s.nop(i16 0) #1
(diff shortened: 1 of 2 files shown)


kito-cheng:

```
```

arsenm (approved):

PR #198697, opened 2026-05-20
[DebugInfo][test] Replace Intel-proprietary intrinsics in fortran-basic.ll [NFC]

kito-cheng:
The test was originally captured from Intel's Fortran compiler and references two intrinsics that have never existed in upstream LLVM:

- @llvm.intel.subscript.p0.i64.i64.p0.i64 was only declared and never called, so it can simply be dropped.
- @llvm.for.cpystr.i64.i64.i64 lowers to a string copy; rewrite the one call site using @llvm.memcpy.p0.p0.i64. The test only checks the emitted CodeView debug info, so the precise lowering of the string copy does not matter.


kito-cheng:

```
```

arsenm (approved):

PR #198698, opened 2026-05-20
[Clang] Fix reference binding traits with cv-qualified array types

### LINKED ISSUE #198580: [Clang] `reference_constructs_from_temporary` mishandles cv-qualified arrays
For the following test case, the MSVC Compiler 19.52 Preview reports `false` while Clang 22.1.3 reports `true`. We believe that Clang is wrong:

<details><summary>(Click to expand version info)</summary>

```
C:\Temp>"%ProgramFiles%\Microsoft Visual Studio\18\Insiders\VC\Auxiliary\Build\vcvarsall.bat" x64 -vcvars_ver=preview
**********************************************************************
** Visual Studio 2026 Developer Command Prompt v18.7.0-insiders
** Copyright (c) 2026 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64'

C:\Temp>cl
Microsoft (R) C/C++ Optimizing Compiler Version 19.52.36405.1 for x64 (PREVIEW)
Copyright (C) Microsoft Corporation.  All rights reserved.

usage: cl [ option... ] filename... [ /link linkoption... ]

C:\Temp>clang-cl -v
clang version 22.1.3 (https://github.com/llvm/llvm-project e9846648fd6183ee6d8cbdb4502213fcf902a211)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\Microsoft Visual Studio\18\Insiders\VC\Tools\Llvm\x64\bin
```
</details>

```
C:\Temp>type reduced.cpp
```
```cpp
#include <print>
#include <type_traits>
using namespace std;

int main() {
    using Obj          = int[1];
    constexpr bool val = reference_constructs_from_temporary_v<const Obj&, volatile Obj>;

#ifdef __clang__
    println("Clang: {}", val);
#else
    println("MSVC: {}", val);
#endif
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest reduced.cpp && reduced
reduced.cpp
MSVC: false

C:\Temp>clang-cl /EHsc /nologo /W4 /std:c++latest reduced.cpp && reduced
Clang: true
```

Why is Clang wrong? See https://godbolt.org/z/8erKjbjsf which shows that all compilers (GCC 16, Clang 22, MSVC 19.50) agree that this is ill-formed:

```cpp
int main() {
    using Obj = int[1];
    using U = volatile Obj;
    const Obj& p(U{});
}
```

[N5046](https://isocpp.org/files/papers/N5046.pdf) \[tab:meta.unary.prop\] specifies `reference_constructs_from_temporary<T, U>` as "`T` is a reference type, and the initialization `T t(VAL<U>);` is well-formed and binds `t` to a temporary object whose lifetime is extended (6.8.7)."

Because this initialization isn't well-formed, the type trait must report `false`. Note that GCC 16 behaves like MSVC and reports `false`, see https://godbolt.org/z/6fbTsnona .

Likely related to #142554 which implemented [LWG-3819](https://cplusplus.github.io/LWG/issue3819) "`reference_meows_from_temporary` should not use `is_meowible`".

frederick-vs-ja:
When the source type is a cv-qualified array type that is denoted by an alias, Clang currently incorrectly drops cv-qualifiers from the array type.

Per C++26 [expr.type]p2 (initially fixed by CWG1261), cv-qualification is kept for array prvalues.

HLSL is consistently affected - this patch makes `__decltype` keep address space qualifiers for array prvalues in HLSL.

Fixes #198580.

### DIFF
--- a/clang/lib/AST/Type.cpp
+++ b/clang/lib/AST/Type.cpp
@@ -3679,13 +3679,15 @@ QualType QualType::getNonLValueExprType(const ASTContext &Context) const {
   if (const auto *RefType = getTypePtr()->getAs<ReferenceType>())
     return RefType->getPointeeType();
 
-  // C++0x [basic.lval]:
-  //   Class prvalues can have cv-qualified types; non-class prvalues always
-  //   have cv-unqualified types.
+  // C++26 [expr.type]p2 (see also CWG1261):
+  //   If a prvalue initially has the type "cv T", where T is a cv-unqualified
+  //   non-class, non-array type, the type of the expression is adjusted to T
+  //   prior to any further analysis.
   //
   // See also C99 6.3.2.1p2.
   if (!Context.getLangOpts().CPlusPlus ||
-      (!getTypePtr()->isDependentType() && !getTypePtr()->isRecordType()))
+      (!getTypePtr()->isDependentType() && !getTypePtr()->isRecordType() &&
+       !getTypePtr()->isArrayType()))
     return getUnqualifiedType();
 
   return *this;
--- a/clang/docs/ReleaseNotes.rst
+++ b/clang/docs/ReleaseNotes.rst
@@ -570,6 +570,8 @@ Bug Fixes in This Version
   an array via an element-at-a-time copy loop (#GH192026)
 - Fixed an issue where certain designated initializers would be rejected for constexpr variables. (#GH193373)
 - Fixed a crash when ``#embed`` is used with C++ modules (#GH195350)
+- Fixed mishandling of cv-qualified array types in ``__reference_constructs_from_temporary`` and
+  ``__reference_converts_from_temporary`` builtins. (#GH198580)
 
 Bug Fixes to Compiler Builtins
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/clang/test/SemaCXX/type-traits.cpp
+++ b/clang/test/SemaCXX/type-traits.cpp
@@ -3310,6 +3310,44 @@ void reference_constructs_from_temporary_checks() {
   // For scalar types, cv-qualifications are dropped first for prvalues.
   static_assert(__reference_constructs_from_temporary(int&&, const int));
   static_assert(__reference_constructs_from_temporary(int&&, volatile int));
+  // For array types, cv-qualifications are *kept* for prvalues.
+  // Uses with type aliases need to be verified, see https://llvm.org/PR198580.
+  static_assert(!__reference_constructs_from_temporary(const int(&)[10], volatile int[10]));
+  static_assert(!__reference_constructs_from_temporary(const IntAr&, volatile int[10]));
+  static_assert(!__reference_constructs_from_temporary(const int(&)[10], volatile IntAr));
+  static_assert(!__reference_constructs_from_temporary(const IntAr&, volatile IntAr));
+  static_assert(!__reference_constructs_from_temporary(int(&&)[10], volatile int[10]));
+  static_assert(!__reference_constructs_from_temporary(IntAr&&, volatile int[10]));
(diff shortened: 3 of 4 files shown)


clang/lib/AST/Type.cpp
```diff
@@ -3679,13 +3679,15 @@ QualType QualType::getNonLValueExprType(const ASTContext &Context) const {
...
+  // C++26 [expr.type]p2 (see also CWG1261):
+  //   If a prvalue initially has the type "cv T", where T is a cv-unqualified
+  //   non-class, non-array type, the type of the expression is adjusted to T
+  //   prior to any further analysis.
   //
   // See also C99 6.3.2.1p2.
   if (!Context.getLangOpts().CPlusPlus ||
-      (!getTypePtr()->isDependentType() && !getTypePtr()->isRecordType()))
+      (!getTypePtr()->isDependentType() && !getTypePtr()->isRecordType() &&
+       !getTypePtr()->isArrayType()))
```
frederick-vs-ja, line 3690: Test failure was related to this.
https://github.com/llvm/llvm-project/blob/05c1a266b84d811f7869d8326674da4f5aee7a89/clang/test/SemaHLSL/Language/InitIncompleteArrays.hlsl#L15-L18

After the seemingly reasonable change in `QualType::getNonPackExpansionType`, `__decltype((T)0)` fails to remove `hlsl_private` from an array type.

frederick-vs-ja: The failure can be resolved by `__typeof_unqual__`. However, there's still potential breaking change of `__decltype` for HLSL. Should we note this in release note or avoid behavior change for HLSL?

cor3ntin (commented):
Would you be willing to ad a dr test for cwg1261 in this PR?

frederick-vs-ja:
Pre-CWG1261 wording is also quoted in some other location(s) now. It's possible that Clang has already effectively implemented CWG1261, but it's a bit too tiring for me to check the implementation status thoroughly at this moment.

clang/lib/AST/Type.cpp
Endilll, line 3690: Between this code and [basic.type]/2, this change seems correct. However, this change seems to be placed rather deep compared to the goal of fixing two type traits. Could it be that we have duplicated code paths in `SemaInit.cpp` that should be calling this function?

Endilll (commented):
This change is backported to all language modes without restrictions. Do we need an ABI tag?
CC @AaronBallman

clang/lib/AST/Type.cpp
AaronBallman: CC @llvm-beanz because it's an HLSL question I don't feel confident answering.

AaronBallman: If this is a non-lvalue expression type, won't the array have been decayed to a pointer already?

clang/lib/AST/Type.cpp
```diff
@@ -3679,13 +3679,15 @@ QualType QualType::getNonLValueExprType(const ASTContext &Context) const {
...
     return getUnqualifiedType();
```
AaronBallman, line 3691: Drive-by: are we sure this is correct for `_Atomic` in C? (This relates to the question of whether we expect this to be called after rvalue conversion or not; if the rvalue conversion has happened before calling this, then the type won't be atomic any longer anyway.)

AaronBallman:
> This change is backported to all language modes without restrictions. Do we need an ABI tag? CC @AaronBallman

Because it impacts type traits, I think an ABI tag would be warranted (every time we change type traits, we usually quietly break how some template specializations get resolved).

PR #198702, opened 2026-05-20
[ThinLTO][Split] Split module for parallel compilation in backend (1/N)

mmjjpp:
This patch runs during the thinlto backend phase, and aims to address the issue of long compilation times for long-tail files. The module is split into N Mparts, and opt and codegen are performed on the Mparts in parallel to implement parallel compilation in the thinlto backend.

Production-scale experiments show substantial compilation time reductions:
| Workload | Baseline | MTPC | Reduction |
| --- | --- | --- | --- |
| Large TU A | 51m 26s | 13m 58s | 72.85% |
| Large TU B | 24m 07s | 7m 38s | 68.35% |
| Large TU C | 9m 34s | 4m 02s | 57.84% |
| Large TU D | 7m 23s | 2m 47s | 62.30% |
| Full Application | 30m 45s | 20m 42s | 32.68% |

**!!! Our RFC can find here: https://discourse.llvm.org/t/rfc-thinlto-multi-thread-parallel-compilation-for-large-module/90877**

Stack Pr:
1/N: [#198702](https://github.com/llvm/llvm-project/pull/198702) [WIP][ThinLTO][Split] Split module for parallel compilation in backend (1/N)  ⬅
1.1/N: https://github.com/mmjjpp/llvm-project/pull/17 [LTO][SplitModuleCG] Support callgraph-based module splitting in LTO backend (1.1/N)  
2/N: https://github.com/mmjjpp/llvm-project/pull/1 [2/N][ThinLTO][SplitModuleCG] Add comdat group handling
3/N: https://github.com/mmjjpp/llvm-project/pull/2 [3/N][ThinLTO][SplitModuleCG] Handle GlobalAlias and GlobalIFunc
4/N: https://github.com/mmjjpp/llvm-project/pull/3 [4/N][ThinLTO][SplitModuleCG] Handle indirect calls in SimplifyCallGraph
5/N: https://github.com/mmjjpp/llvm-project/pull/4 [5/N][ThinLTO][SplitModuleCG] Add vtable support
6/N: https://github.com/mmjjpp/llvm-project/pull/5 [6/N][ThinLTO][SplitModuleCG] Preserve internal linkage for thinlto-internalize globals
7/N: https://github.com/mmjjpp/llvm-project/pull/6 [7/N][ThinLTO][SplitModuleCG] Preserve init-array entries during module splitting

Nice to have, performance related.
8/N: https://github.com/mmjjpp/llvm-project/pull/7 [8/N][ThinLTO][SplitModuleCG] check for large call graphs
9/N: https://github.com/mmjjpp/llvm-project/pull/8 [9/N][ThinLTO][SplitModuleCG] Move SampleProfileLoaderPass before ThinLTO split
10/N: https://github.com/mmjjpp/llvm-project/pull/9 [10/N][Thinlto][SplitModuleCG] Add TimeTrace support for ThinLTO split
11/N: https://github.com/mmjjpp/llvm-project/pull/10 [11/N][ThinLTO][SplitModuleCG] Add remark info while split thinlto
12/N: https://github.com/mmjjpp/llvm-project/pull/11 [12/N][ThinLTO][SplitModuleCG] Fix duplicate debug info in split partitions
13/N: https://github.com/mmjjpp/llvm-project/pull/12 [13/N][Debuginfo][SplitModuleCG] Generate identifier when the class is dynamic
14/N: https://github.com/mmjjpp/llvm-project/pull/13 [14/N][ThinLTO][SplitModuleCG] Add the CloneModule to the parallel
15/N: https://github.com/mmjjpp/llvm-project/pull/14 [15/N][ThinLTO][SplitModuleCG] Move parseBitecodeFile out of parallel
16/N: https://github.com/mmjjpp/llvm-project/pull/15 [16/N][ThinLTO][SplitModuleCG] Add numa node binding support for parallel cloning
17/N: https://github.com/mmjjpp/llvm-project/pull/16 [ThinLTO][Driver] Move split-codegen partition merge out of LTOBackend

PS:  I do not have permission to push users/<username>/feature branch to the upstream repo now, so I submit PRs by fork repo. To show incremental diffs, PRs 2~N are all in fork repo. Once this PR is merged later (or after get the permissions), subsequent PRs will be migrated to this repo one by one.

The changes in this PR were made with AI assistance and have been reviewed by me.


llvm/lib/LTO/LTOBackend.cpp
```diff
@@ -513,6 +541,208 @@ static void codegen(const Config &Conf, TargetMachine *TM,
...
+  SmallString<128> MergedFilename;
+  if (sys::fs::createTemporaryFile("thinlto-merged", "o", MergedFD,
+                                   MergedFilename))
+    report_fatal_error("Failed to create merged temp file.");
+  llvm::FileRemover MergedFileRemover(MergedFilename);
+  sys::fs::closeFile(MergedFD);
+
+  std::vector<StringRef> Args;
+  std::string LinkerPath = "";
+  if (auto Path = sys::findProgramByName("ld.lld"))
```
Jinjie-Huang: This appears to search for ld.lld in the environment. Seems better to prioritize clang's local directory first, and then check the system PATH?

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
```diff
@@ -0,0 +1,34 @@
+#ifndef LLVM_TRANSFORMS_UTILS_SPLITMODULECG_H
```
shiltian, line 1: We already have (more than one) module split pass. I don't see a reason of a new pass.

teresajohnson (changes requested):
This change has no description and no tests - it needs both, and possibly an RFC for a change as big as this. At first glance, a couple of high level comments: LTO should not be directly calling lld. Also, regular LTO already has support for split codegen, can that be leveraged? In particular, look at how that adds the split modules to the link, which is via the AddStream callback.

jiangning26:
> This change has no description and no tests - it needs both, and possibly an RFC for a change as big as this. At first glance, a couple of high level comments: LTO should not be directly calling lld. Also, regular LTO already has support for split codegen, can that be leveraged? In particular, look at how that adds the split modules to the link, which is via the AddStream callback.

Looking at the existing Regular LTO split-codegen path, I think the ThinLTO handling has two distinct cases.

### 1. Linker-driven In-process ThinLTO
Since `lld` is already the LTO client, I agree this should follow the Regular LTO model:
* The ThinLTO backend should emit each split partition through `AddStream`.
* `lld` should receive those object buffers from `LTO::run()`, create `InputFiles` from them, and add them to the final link.
* **Key constraint:** There should be no recursive linker invocation from `LTOBackend` in this path.
* **Note:** The main extra piece is task numbering / `getMaxTasks()`, since one ThinLTO backend task can produce multiple partition objects.

### 2. Distributed ThinLTO Backend Compilation (`clang -fthinlto-index -c`)
Here, there is no in-process `lld` client receiving `AddStream` buffers. `cc1` is just producing backend objects, while the clang driver still has the normal `-c -o foo.o` contract to produce a single relocatable object. 

**Proposed direction:**
1. `cc1` uses `AddStream` to emit the split partition objects and writes an output list.
2. The `clang` driver performs a follow-up merge: `ld.lld -r @list -o foo.o`.

*Benefit:* This keeps the linker invocation out of the LTO library/backend, while preserving the existing distributed backend single-output contract.

---

### Summary of Responsibilities
Does that split of responsibilities sound reasonable to you? In other words:

* **In-process ThinLTO:** `LTO` -> `AddStream` -> `lld` consumes partition objects directly.
* **Distributed `-fthinlto-index -c`:** `LTO`/`cc1` -> `AddStream` writes partition objects -> `clang` driver merges them with `ld.lld -r`.

llvm/lib/LTO/LTOBackend.cpp
mmjjpp: We've moved lld -r logic into the driver. We use ToolChain.GetProgramPath() to find the linker executable, and its lookup order is straightforward:
1. First look through every prefix directory passed with -B;
2. Then check the toolchain's program paths from getProgramPaths() — this normally includes the folder where clang lives;
3. If those checks fail, it finally looks up the binary from the PATH environment variable.

mmjjpp:
> > This change has no description and no tests - it needs both, and possibly an RFC for a change as big as this. At first glance, a couple of high level comments: LTO should not be directly calling lld. Also, regular LTO already has support for split codegen, can that be leveraged? In particular, look at how that adds the split modules to the link, which is via the AddStream callback.
> 
> Looking at the existing Regular LTO split-codegen path, I think the ThinLTO handling has two distinct cases.
> 
> ### 1. Linker-driven In-process ThinLTO
> Since `lld` is already the LTO client, I agree this should follow the Regular LTO model:
> 
> * The ThinLTO backend should emit each split partition through `AddStream`.
> * `lld` should receive those object buffers from `LTO::run()`, create `InputFiles` from them, and add them to the final link.
> * **Key constraint:** There should be no recursive linker invocation from `LTOBackend` in this path.
> * **Note:** The main extra piece is task numbering / `getMaxTasks()`, since one ThinLTO backend task can produce multiple partition objects.
> 
> ### 2. Distributed ThinLTO Backend Compilation (`clang -fthinlto-index -c`)
> Here, there is no in-process `lld` client receiving `AddStream` buffers. `cc1` is just producing backend objects, while the clang driver still has the normal `-c -o foo.o` contract to produce a single relocatable object.
> 
> **Proposed direction:**
> 
> 1. `cc1` uses `AddStream` to emit the split partition objects and writes an output list.
> 2. The `clang` driver performs a follow-up merge: `ld.lld -r @list -o foo.o`.
> 
> _Benefit:_ This keeps the linker invocation out of the LTO library/backend, while preserving the existing distributed backend single-output contract.
> 
> ### Summary of Responsibilities
> Does that split of responsibilities sound reasonable to you? In other words:
> 
> * **In-process ThinLTO:** `LTO` -> `AddStream` -> `lld` consumes partition objects directly.
> * **Distributed `-fthinlto-index -c`:** `LTO`/`cc1` -> `AddStream` writes partition objects -> `clang` driver merges them with `ld.lld -r`.

pr： https://github.com/mmjjpp/llvm-project/pull/16 [ThinLTO][Driver] Move split-codegen partition merge out of LTOBackend

llvm/lib/LTO/LTOBackend.cpp
teresajohnson: Do we need both of these flags or can the latter one be inferred from the former's value? If there must be 2 you probably need some error checking for invalid combinations? But I think moving to one is simpler.

clang/test/CodeGen/thinlto-split/fulllto-split-module.c
```diff
@@ -0,0 +1,26 @@
+// UNSUPPORTED: system-windows
```
teresajohnson: The tests for the LTO part should be llvm LTO tests (i.e. using llvm-lto2, under llvm/test/LTO and llvm/test/ThinLTO), and not clang in this PR.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson: CombinedIndex seems unused in this PR?

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
teresajohnson, line 32: Can you add a high level comment about what this class does?

teresajohnson: I'm a little unsure of how you are detecting indirect calls below. Also, what do you mean by "non-entrypoint" function?

teresajohnson: Document "LimitPartition" - is that the "N" in the comment above the class?

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson, line 26: What if there is already an external value in some module that has the same name as the now-promoted local? For ThinLTO promotion we deal with this by always renaming appending ".llvm.${modulehash}".

teresajohnson: By import do you mean clone? "import" sounds like ThinLTO importing but this is a little different. Also, please fix the line formatting (extra indentation in next line). I'm not sure what you mean by "[dependency] every module"?

teresajohnson: What do you mean by "include clonemodule in parallel"?

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
teresajohnson, line 262: Can you document this and other function declarations? Not clear from the name what it does

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson: I think I get the gist of what you are saying, but can you rewrite for clarity / grammatical correctness? Do we need to serialize it out and back in to get it in a new context? That's going to add overhead, but I haven't investigated alternatives.

teresajohnson: What do you mean for this TODO?

teresajohnson: I see indirect call support is in a follow on PR? What happens right now, with this PR - will they break the handling?

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
teresajohnson: Yeah I'd really like to see less duplication - even if the splitting criteria is different, can the underlying handling be extracted and shared? Surely they all have to deal with similar issues with promotions of locals, etc.

llvm/lib/LTO/LTOBackend.cpp
teresajohnson: Can this use llvm's llvm::sys::path::filename()?

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson: Can this use getUniqueModuleId() instead of a different hashing mechanism?

teresajohnson: Oh, I see you do the renaming elsewhere. Can these be done at the same time? If not, needs a clear comment.

llvm/lib/LTO/LTOBackend.cpp
teresajohnson: The name seems wrong if it is used also for regular LTO (remove "Thin"?).

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
teresajohnson: Does it matter that the iteration order will be non-deterministic (indexed by pointers)? If determinism isn't needed, prefer DenseMap.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson: also, remove the empty if statement, even if you need a TODO.

teresajohnson: Rather than fully recomputing the cost on insertion of new functions, can it be updated by adding in the costs of the new ones inserted above?

teresajohnson: change this to an early continue if QueuePID != PID

teresajohnson (commented):
I started reviewing and have some high level questions and comments, and a number of other initial comments sprinkled throughout. Can you split this PR into 2? The first being the Utils support, llvm-split changes, and the associated llvm-split based tests; the second being the LTO support with some added LTO tests?

I know the callgraph based splitting is to avoid performance overhead from loss of IPO, but have you evaluated this?

mmjjpp:
> I started reviewing and have some high level questions and comments, and a number of other initial comments sprinkled throughout. Can you split this PR into 2? The first being the Utils support, llvm-split changes, and the associated llvm-split based tests; the second being the LTO support with some added LTO tests?
> 
> I know the callgraph based splitting is to avoid performance overhead from loss of IPO, but have you evaluated this?

Thank you very much for the detailed review comments. I’ll carefully go through them and address the issues you raised.

I have now split the original PR into two, following your suggestion: one for the Utils support, llvm-split changes, and the associated llvm-split tests; and another for the LTO-related support and tests. The LTO backend related changes are available here: https://github.com/mmjjpp/llvm-project/pull/17

Regarding the potential IPO impact of callgraph-based splitting, we did evaluate it in our practical workload. The observed performance variation was within 1%. In this evaluation, PGO was enabled, while WPD, CFI, and MemProf were not exercised.

llvm/lib/LTO/LTOBackend.cpp
mmjjpp: I have simplified this by merging the two options into one.

Now, option lto-split-partitions has two settings: a value greater than 1 enables SplitModuleCG and the partition counts are split; otherwise, SplitModuleCG is disabled.

The updated implementation is available at https://github.com/mmjjpp/llvm-project/pull/17.

clang/test/CodeGen/thinlto-split/fulllto-split-module.c
mmjjpp: The changes have been made based on the feedback.
Please refer to the updates at https://github.com/mmjjpp/llvm-project/pull/17.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
mmjjpp: Yes, you’re right. CombinedIndex is not used in this PR; it will be needed in a follow-up PR. I have removed it from this PR.

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
mmjjpp: Thanks for pointing this out. This was an oversight on my part. I have added high-level comments to the classes to clarify their responsibilities.

mmjjpp: > I'm a little unsure of how you are detecting indirect calls below.

The indirect-call detection currently uses two complementary approaches:

Tracing uses of address-taken functions: for each function whose address is taken, we recursively traverse its users (including constants and global variables). When an instruction using the function pointer is found, we attribute that indirect-call relationship to the instruction’s parent function.
Resolving indirect call instructions: for call instructions whose callee cannot be resolved directly, we:
strip pointer casts to recover the underlying function where possible;
consult callees metadata;
use indirect-call PGO profile data to add the profiled candidate callees.

> what do you mean by "non-entrypoint" function?

By “non-entrypoint function,” I mean a function that is not a root of the simplified call graph. We identify graph roots by checking NumReferences == 0, where NumReferences is the number of incoming edges. This is a graph-level notion and does not necessarily mean the program’s actual entry function.

For example, in a strongly connected component such as A -> B -> A, both functions have incoming references, so neither has NumReferences == 0. Such functions are therefore not discovered through the initial root search; the implementation handles these unvisited cyclic components afterward by treating one of their functions as a root.

mmjjpp: They are related but not identical. N is the actual number of partitions produced by the final splitting process. LimitPartition is the upper bound supplied through the lto-split-partitions option.

If the call graph contains fewer independent partitions than LimitPartition, generating additional empty or redundant output files is unnecessary. Therefore, the final number of partitions may be smaller than the requested limit: N <= LimitPartition.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
mmjjpp: In SplitModuleCG, we do not perform the actual renaming immediately. We only record the required renaming information and apply the rename after the opt pipeline has completed.

This is intentional: deferring the rename minimizes the impact of renaming on subsequent optimizations while still allowing us to handle name collisions consistently.

mmjjpp: The changes have been made based on the feedback.

mmjjpp: The changes have been made based on the feedback.

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
mmjjpp: The changes have been made based on the feedback.

llvm/lib/LTO/LTOBackend.cpp
mmjjpp: The changes have been made based on the feedback.
Please refer to the updates at https://github.com/mmjjpp/llvm-project/pull/17.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
mmjjpp: The changes have been made based on the feedback.

llvm/lib/LTO/LTOBackend.cpp
mmjjpp: The changes have been made based on the feedback.
Please refer to the updates at https://github.com/mmjjpp/llvm-project/pull/17.

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
mmjjpp: The changes have been made based on the feedback.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
mmjjpp: Thanks for pointing this out. My intention was only to leave a TODO indicating that indirect-call handling would be added in a follow-up PR, but I agree that the empty if statement is confusing and can suggest incomplete or incorrect handling in this PR.

I have removed the TODO and the empty if statement, and cleaned up the other similar cases as well.

mmjjpp: Thanks for the suggestion. I have rewritten and expanded the comment.

mmjjpp: I meant that, as future work, we may parallelize the CloneModule step as well to further improve compilation throughput. I have reworded the comment to make that clearer.

mmjjpp: Yes, “import” was imprecise here; I meant “clone.” The behavior is to conservatively clone private and internal globals into every output partition, and dealWithMpart removes any unused copies afterward.

mmjjpp: This TODO referred to another indirect-call detection mechanism that I planned to implement in a follow-up PR. Since it is not part of the current change and is unclear in this context, I have removed the TODO.

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
mmjjpp: Do you mean extracting the common functionality into a base class, with each splitting strategy deriving from it and implementing its strategy-specific behavior?

I agree that reducing duplication would be beneficial. However, the current splitting strategies differ substantially in their overall workflows, so factoring out the shared handling may require a fairly significant refactoring effort.

mmjjpp:
@teresajohnson Thank you again for your earlier review. I have addressed the comments and pushed the updates. When you have a chance, could you please take another look? Thanks!

teresajohnson:
> @teresajohnson Thank you again for your earlier review. I have addressed the comments and pushed the updates. When you have a chance, could you please take another look? Thanks!

Will do. I've been out of office most of the month, still trying to catch up.

llvm/tools/llvm-split/llvm-split.cpp
teresajohnson: let's spell out the CG with lowercase, which is the common practice in LLVM. Maybe "enable-call-graph-split-module"?

llvm/test/tools/llvm-split/SplitModuleCG/function-with-ring.ll
teresajohnson: Add a sentence or 2 short comment at the top of each test as to what scenario it is testing and or what it is looking for

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
teresajohnson: Is this correct? Don't these iterate through functions and their corresponding nodes?

teresajohnson: Can we avoid the duplication for this and other methods that provide both const and non-const versions. E.g.

```
inline const SimplifyCallGraphNode *operator[](const Function *F) const {
  const_iterator I = FunctionMap.find(F);
  assert(I != FunctionMap.end() && "Function not in callgraph!");
  return I->second.get();
}

inline SimplifyCallGraphNode *operator[](const Function *F) {
  return const_cast<SimplifyCallGraphNode *>(
      static_cast<const CallGraph &>(*this)[F]);
}
```

teresajohnson: Do you really need to support both operator[] and at() ? Since they are identical in behavior suggest picking one.

teresajohnson: use lower camel case for the methods

teresajohnson: How about SimplifiedCallGraph* instead of SimplifyCallGraph? The latter suggests some type of optimized action to me.

teresajohnson: any reason to make createSimplifyCallGraph a separate function instead of just inlining into the constructor? It's only called once.

teresajohnson: afaict this is only used during the construction - do you need to save the ref?

teresajohnson: Add a documenting comment above the def

teresajohnson, line 145: Document struct purpose

teresajohnson, line 161: Document what type of cost this holds

teresajohnson: You wouldn't need this separate add of F's cost if addAllDependencies included it (which simplifies its logic, see comment there).

teresajohnson: Since the sole caller currently is to compute the total cost, and adds in F's cost, you could simplify this function (and that code - see comment there), by adding F to Fns when you add it to the worklist at the top. Then you don't need this check at all.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson: guard the whole loop by `#ifndef NDEBUG`

teresajohnson: What is the purpose of this handling (going newFn->getName() to M.getFunction to get a different Fn)?

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
teresajohnson: Ideally at least some refactoring and sharing of common utilities. For example, SplitModule.cpp's externalize() seems identical to the one here. And the handling that invokes it is very similar - can there be a single refactored facility for invoking externalize across a module, shared by both? That's one example but I'm wondering if there are others, as mechanically they presumably have to do a lot of the same work to perform the partitioning.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson: I think it would be preferable to include changes like this one in a PR where it can be tested. Or is there a way to test the handling e.g. via llvm-split here in this PR?

teresajohnson: Suggest inlining the lambda - keeping the body separated from the loop doesn't improve readability imo,  unless you plan to add additional and logically different functionality into the loop body in a follow on PR.

teresajohnson: lambda names should follow the variable coding convention of upper camel case (but as noted further below, I think it would be better to inline this one at its use site).

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
teresajohnson: It looks like this and the following 2 methods are unused in this PR - prefer adding those in the PR that will use / test them.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson: Can you add a comment about what the loop is doing at a high level?

teresajohnson: Can you replace Import with Clone here and elsewhere in the PR to be clearer?

teresajohnson, line 319: No braces for single instruction if bodies

teresajohnson: No need for the identity assignment when N != 0. Just do a conditional set with 1.

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
teresajohnson: Does it matter that the iteration order is going to be non-deterministic (both because it is a DenseMap and because it is indexed by pointer)? i see you do iterate over this map in several places.

teresajohnson: no braces for single line for loop bodies

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson: Since you only use the values below, you could iterate over SCG->values() and assign directly to SCGNode. Also, just keep it as a unique_ptr rather than using get() to access the raw pointer like you are below?

teresajohnson: Nit here and elsewhere, remove braces for single line if bodies (per llvm coding conventions)

teresajohnson: There are a lot of places like this for if and for loop bodies, I added a few other comments but I see quite a few so please scan and fix the others that I haven't noted.

teresajohnson: why not just keep as a unique_ptr instead of using get()?

teresajohnson: by ring do you mean cycle? I think cycle is the standard term, suggest using that here and elsewhere

teresajohnson: suggest flipping the condition and doing a continue (https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code)

teresajohnson: At the end of the SplitModuleCG constructor, if N is 0 it is set to 1. Can this instead be an assert that it isn't 0?

teresajohnson: Suggest inlining the lambda here, otherwise the description is separated from the use unnecessarily.

teresajohnson: Probably don't need CostVal variable, just make it a single line loop without braces

teresajohnson: Actually, haven't you just promoted all local variables above via externalize()? What case does this code handle?

teresajohnson: Can you add a comment about what we're trying to detect, in particular why the !F.isDefinitionExact

teresajohnson: Suggest inlining this lambda here

teresajohnson: use contains() instead of count() here. Also, suggest flipping the conditions and using an early continue.

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
mmjjpp: done

mmjjpp: Done

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
mmjjpp: Done

mmjjpp: I reconsidering the !F.isDefinitionExact() part of this condition. My original intent was to be conservative: these functions (weak/linkonce/etc.) have definitions that may be replaced at link time, so I wanted to ensure only one partition keeps the real definition and others get available_externally.

However, I'm now leaning towards removing it. Functions with non-exact definitions (weak, linkonce, etc.) are inherently designed to support multiple definitions — the linker deduplicates them at link time, so there is no duplicate symbol risk. If such a function is defined in multiple partitions, the linker will select one definition regardless, so forcing available_externally on the duplicates is unnecessary.

After the change, the condition would simplify to:
if (!F.isDeclaration() && F.hasExternalLinkage())
  externalFunction[&F] = true;
This still covers all functions that would actually cause duplicate symbol issues (both originally-external functions and locals promoted to external by externalize()).

What do you think? Is there a case I'm missing where keeping the !isDefinitionExact() check would be important?

mmjjpp: Done

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
mmjjpp: Done

mmjjpp: Good point — I've switched FunctionMapTy to std::map for deterministic iteration.
The non-deterministic iteration order of DenseMap does not affect the final partitioning output: FunctionMap is iterated in createWorkList() to discover call-graph roots (in-degree 0), where the result is collected into a DenseSet (EntryFuncs) and the resulting FWDWorkList is subsequently sorted by (TotalCost desc, Function name asc). Each FunctionWithDependencies entry's content (dependencies set, total cost) is also order-independent. However, print() would still produce non-deterministic debug output, hence the switch to std::map.

mmjjpp: Done

mmjjpp: This has been corrected. This comment was left over from previous comments on other functions.

mmjjpp: Done

mmjjpp: Done

mmjjpp: Yes, that's not necessary; it's already been removed.

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done. I have reviewed the PR thoroughly and made the necessary corrections.

mmjjpp: Done

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done. Since llvm-split is currently the only consumer of this API, I moved the rename logic directly into dealWithMpart (which runs in the main thread before serialization). The method now detects promoted locals and renames them in a single pass — the intermediate PromotedRenames map has been removed entirely, since getUniqueModuleId(&M) produces a deterministic suffix consistent across all partitions.

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: Done

mmjjpp: You're right — since externalize() promotes all local globals to ExternalLinkage + HiddenVisibility above, NeedsConservativeClone checks hasLocalLinkage() which is always false at that point, making it dead code.
My original concern was that some globals might be missed, leading to undefined references. But on reflection, after externalize(), all globals are external and go to partition 0 (via the ShouldCloneDefinition fallback), while other partitions have declarations — which is correct for cross-partition linking.
I've removed the NeedsConservativeClone mechanism entirely, along with the associated cleanup loop in dealWithMpart.

mmjjpp: The M.getFunction(newFn->getName()) lookup was a leftover from an earlier experiment where we tried moving the bitcode serialization/deserialization before CloneModule. In that setup, the GV passed to the callback would have been from a re-parsed module rather than from M, so a name-based lookup back into M was needed to find the original function pointer.
Since that approach was not in this pr, the GV passed by CloneModule is already from M, and FnToClone is already a const Function * in M that can be directly checked against FnsInPart. I've removed the redundant lookup and simplified it to return FnsInPart.contains(FnToClone).

mmjjpp: Done

mmjjpp: done

llvm/tools/llvm-split/llvm-split.cpp
mmjjpp: Done

llvm/test/tools/llvm-split/SplitModuleCG/function-with-ring.ll
mmjjpp: Done

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
mmjjpp: Good point — I agree that externalize() and the surrounding invocation loops are duplicated between SplitModule.cpp and SplitModuleCG.cpp, and refactoring them into a shared facility would reduce duplication.
Based on your experience with LLVM's review process, would a follow-up refactoring PR be more appropriate here? The current PR is already fairly large with the review-driven changes, and the refactoring would touch SplitModule.cpp (which this PR doesn't currently modify), so a separate PR might keep the scope cleaner. But I'm happy to do either — let me know what you think.

teresajohnson: Usually what I would recommend is an initial PR that refactors the existing code to expose reusable pieces. Then rebase this one on top of that once it is in. I don't know if there are other pieces that could be refactored and shared, I'm not that familiar with the existing SplitModule, but that part jumped out at me on a quick perusal.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -99,82 +90,71 @@ void SplitModuleCG::calculateFunctionCosts() {
...
-  for (auto &GV : make_early_inc_range(MPart.globals())) {
-    if (NeedsConservativeImport(&GV) && GV.use_empty())
-      GV.eraseFromParent();
-  }
-
+void SplitModuleCG::dealWithMpart(Module &MPart, unsigned I) {
+  // Downgrade duplicate definitions of external functions to
+  // available_externally. The first partition to define such a function keeps
+  // the real definition; all other partitions get available_externally copies.
   for (auto &func : MPart.functions()) {
```
teresajohnson: This section of the code is using a lowerCamelCase for a number of variables and they should be UpperCamelCase per llvm's coding conventions. Can you check the rest of the PR too?

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -99,82 +90,71 @@ void SplitModuleCG::calculateFunctionCosts() {
...
+    if (func.isDeclaration())
+      continue;
     auto Fn = M.getFunction(func.getName());
```
teresajohnson: I guess this is getting the equivalent Function in the original Module? Can you add a comment as this threw me off at first since func is also a Function.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -99,82 +90,71 @@ void SplitModuleCG::calculateFunctionCosts() {
...
+  // convention "name.llvm.<suffix>" so the promoted local cannot clash with
+  // an external that happens to share the same name. The suffix is derived
+  // from the module via getUniqueModuleId, so it is consistent across all
+  // partitions.
+  std::string Suffix = getUniqueModuleId(&M);
+  for (auto &GV : MPart.global_values()) {
+    // Now external (not local), but was not originally external.
+    if (GV.hasLocalLinkage() || OriginalExternals.contains(GV.getName()))
+      continue;
+    // Skip declarations of functions that were not explicitly externalized
```
teresajohnson, line 149: I didn't follow this comment or what case it is detecting. How can we have a definition in another module that is local and a declaration for that local in a separate module?

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson: I think this may be a reasonable optimization for weak and linkonce ODR types (by downgrading all but one copy to available externally it is less to codegen). But I don't think it would be legal to downgrade an interposable linkage to available externally as they have different optimization semantics. So I would suggest changing the !F.isDefinitionExact to instead check for either of weak/linkonce ODR linkage types.

teresajohnson: Also suggest adding tests for both the odr and interposable cases to make sure they are handled appropriately.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
mmjjpp: Done

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
mmjjpp: A function skipped by the hasOneUse() check in SplitModule (e.g. a local function with a single direct call) stays internal in the source module and is not recorded in externalFunction. However, CloneModule still creates a declaration for such a function in any partition that doesn't hold its definition, and promotes the linkage from internal to dso_local for that declaration (an internal declaration is meaningless). So we can end up with:
; partition 1 (owns the definition)
define internal void @foo() { ... }

; partition 0 (CloneModule-created declaration)
declare dso_local void @foo()

The declaration in partition 0 is not local and is not in OriginalExternals, so it would otherwise match the rename condition. Without this check it would become @foo.llvm.<suffix> while the internal definition in partition 1 keeps the original name — a link-time mismatch. This check therefore keeps such declarations unrenamed, consistent with the unrenamed internal definitions in the partitions that own them.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
mmjjpp: Done

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
mmjjpp: I've opened the refactoring PR: https://github.com/llvm/llvm-project/pull/221181— it moves the identical externalize() from SplitModule.cpp and AMDGPUSplitModule.cpp into ModuleUtils as a shared llvm::externalize. I looked at AMDGPUSplitModule.cpp, SplitModule.cpp, SplitModuleByCategory.cpp, but the invocation loops and partitioning algorithms differ enough that I didn't find more to extract — happy to do so if you spot any.
@teresajohnson @shiltian Could you review the refactoring PR? Once it lands I'll rebase this one on top and drop the local copy.

mmjjpp:
Now that PR #221181 has been merged, I have rebased the changes and updated the unnamed.ll test accordingly.

During this process, I discovered a bug in the original logic: sorting the FWDWorkList was happening before externalizeGlobal() was called. Because our sorting comparator falls back to alphabetical name comparison when costs are identical, sorting unnamed internal functions (like @0 and @1) before they receive their stable names leads to non-deterministic ordering. This non-determinism can result in inconsistent partitioning.

To fix this, I extracted the sorting logic into sortWorkList() and deferred its execution until after all globals are externalized and assigned stable names.

@teresajohnson ， please let me know if you have any questions or further comments.

mmjjpp:
> Now that PR #221181 has been merged, I have rebased the changes and updated the unnamed.ll test accordingly.
> 
> During this process, I discovered a bug in the original logic: sorting the FWDWorkList was happening before externalizeGlobal() was called. Because our sorting comparator falls back to alphabetical name comparison when costs are identical, sorting unnamed internal functions (like @0 and @1) before they receive their stable names leads to non-deterministic ordering. This non-determinism can result in inconsistent partitioning.
> 
> To fix this, I extracted the sorting logic into sortWorkList() and deferred its execution until after all globals are externalized and assigned stable names.
> 
> @teresajohnson ， please let me know if you have any questions or further comments.


@teresajohnson, All feedback should be addressed now. Let me know if anything else is needed before we can move forward.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+  // First, find all the entry functions with an in-degree of 0
+  // (i.e., those that are not called by any function).
+  for (auto &SCGNode : SCG->values()) {
+    Function *F = SCGNode->getFunction();
+    if (F && SCGNode->getNumReferences() == 0)
+      EntryFuncs.insert(F);
+  }
+
+  // Second, find all the dependencies of each entry function.
+  for (auto *F : EntryFuncs) {
```
teresajohnson, line 167: skip braces for single line loop bodies

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+    FWDWorkList.emplace_back(*SCG, FuncsCosts, F);
+  }
+
+  // Third, find all the functions that are not in the worklist.
+  DenseSet<const Function *> SeenFunctions;
+  for (const auto &Fwd : FWDWorkList) {
```
teresajohnson, line 173: ditto

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
teresajohnson: I'm still confused. dso_local is not a linkage type, so it doesn't make sense to say that you are promoting the linkage from internal to dso_local. Also, internal symbols don't participate in linker symbol resolution, so the names in partition 1 and 0 don't need to match. Why have a declaration of foo in partition 0 at all? It cannot reference the partition 1 internal. If this situation arises - can you just remove the declaration in partition 0?

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+      continue;
+    F.externalize();
+    // Record functions that may be defined in multiple partitions so that
+    // dealWithMpart can downgrade duplicates to available_externally.
+    if (canDowngradeToAvailableExternally(F))
+      ExternalFunction[&F] = true;
+  }
+  for (GlobalVariable &GV : M.globals())
+    GV.externalize();
+  for (GlobalAlias &GA : M.aliases())
```
teresajohnson, line 225: I remember that you were going to defer handling of alias and ifunc to a later PR in the series - but I can't recall what happens with them now. There aren't any tests added using alias or ifunc - will these fail gracefully or should a test be added that shows the expected current failing behavior with a TODO (and add a TODO here?).

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+    }
+  }
+
+  // Rename GlobalValues whose linkage was promoted from local to external,
+  // to avoid duplicate symbols across partitions in ThinLTO. Use the naming
+  // convention "name.llvm.<suffix>" so the promoted local cannot clash with
+  // an external that happens to share the same name. The suffix is derived
+  // from the module via getUniqueModuleId, so it is consistent across all
+  // partitions.
+  std::string Suffix = getUniqueModuleId(&M);
```
teresajohnson, line 130: This can in some cases return an empty string, so callers typically have a fallback (https://github.com/llvm/llvm-project/blob/8afb2a08d8b8d429ecd290e1052a9ca2fd2ed701/llvm/lib/Transforms/IPO/ThinLTOBitcodeWriter.cpp#L300-L301). Perhaps this should be checked earlier and the split skipped if one is not available?

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
```diff
@@ -0,0 +1,245 @@
...
+
+private:
+  unsigned N;
+  Module &M;
+  CallGraph CG;
+  std::unique_ptr<SimplifiedCallGraph> SCG;
+  CostType ModuleCost;
+  DenseSet<const Function *> EntryFuncs;
+  StringSet<> OriginalExternals;
+  DenseMap<const Function *, bool> ExternalFunction;
```
teresajohnson, line 211: Can you document what the map is holding (i.e. existence in the map vs the mapped bool val)

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+    return I == 0;
+  };
+
+  // TODO: Consider parallelizing the per-partition CloneModule call itself.
+  // Today the loop below serially clones M into N partitions in the main
+  // thread, then spawns N worker threads to run opt+codegen. If CloneModule
+  // becomes a bottleneck for large modules, the clones could be produced in
+  // parallel too — but that would require either per-thread LLVMContexts
+  // for the clone step or a thread-safe CloneModule, neither of which is
+  // straightforward.
```
teresajohnson, line 253: Another issue is that dealWithMpart's behavior (which duplicate def is kept vs downgrading to available_externally) is order-dependent. Worth noting in the comment.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+    WriteBitcodeToFile(*MPart, BCOS);
+    MPart.reset();
+    Threads.emplace_back(
+        [&, I](SmallString<0> BC) {
+          llvm::lto::LTOLLVMContext Ctx(C);
+          Expected<std::unique_ptr<Module>> MOrErr =
+              parseBitcodeFile(MemoryBufferRef(BC.str(), "ld-temp.o"), Ctx);
+          BC = SmallString<0>();
+          if (!MOrErr)
+            report_fatal_error("Failed to read bitcode");
```
teresajohnson, line 283: The error returned in MOrErr is being dropped - can you report that? i.e. something like report_fatal_error(MOrErr.takeError()).

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
```diff
@@ -0,0 +1,245 @@
...
+  ///          produce. Pass 0 (the default) to derive the partition count
+  ///          from the number of call-graph roots discovered in
+  ///          createWorkList (one root per partition at most). The actual
+  ///          partition count is finalized in the constructor.
+  SplitModuleCG(Module &M, unsigned LimitPartition = 0);
+  void SplitModule(ModuleCreationCallback ModuleCallback,
+                   const llvm::lto::Config &C);
+
+private:
+  unsigned N;
```
teresajohnson, line 204: Can you change N to something more descriptive? Like NumPartitions or the like. Also, what if the number of entry functions is really large? Just realized you aren't using llvm::DefaultThreadPool. Suggest using that with heavyweight_hardware_concurrency() to avoid overwhelming the machine.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+    LLVM_DEBUG(dbgs() << "[root] " << Fwd.F->getName()
+                      << " (totalCost:" << Fwd.TotalCost
+                      << ";   root function cost: " << FuncsCosts[Fwd.F]
+                      << ";   has dependency: " << Fwd.Dependencies.size()
+                      << ")\n");
+#endif
+}
+
+void SplitModuleCG::SplitModule(ModuleCreationCallback ModuleCallback,
+                                const llvm::lto::Config &C) {
```
teresajohnson, line 213: Including LTO headers into Utils is going to create a layering violation. Can you change this to just take a LLVMContext reference?

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
```diff
@@ -0,0 +1,245 @@
...
+#include "llvm/Analysis/CallGraph.h"
+#include "llvm/LTO/Config.h"
+#include <map>
+
+namespace llvm {
+
+class SimplifiedCallGraph;
+class SimplifiedCallGraphNode;
+
+using CostType = InstructionCost::CostType;
```
teresajohnson, line 16: This introduces a global llvm::CostType. Can you avoid this?

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
```diff
@@ -0,0 +1,245 @@
...
+
+  DenseSet<SimplifiedCallGraphNode *> CalledFunctions;
+  unsigned NumReferences = 0;
+
+  void addRef() { ++NumReferences; }
+};
+
+/// Collect \p F and all non-declaration functions transitively called by \p F,
+/// using the SimplifiedCallGraph \p SCG, and insert them into \p Fns.
+static void addAllDependencies(SimplifiedCallGraph &SCG, const Function &F,
```
teresajohnson, line 119: Can you avoid creating a static function in the header? Maybe inline instead would work? Or move to cpp file.

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
```diff
@@ -0,0 +1,245 @@
+#ifndef LLVM_TRANSFORMS_UTILS_SPLITMODULECG_H
```
teresajohnson, line 1: New header and source file missing file header: https://llvm.org/docs/CodingStandards.html#file-headers

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
```diff
@@ -0,0 +1,245 @@
...
+  ///
+  /// \param M The module to partition. Must outlive the SplitModuleCG
+  ///          instance and any partitions emitted via SplitModule().
+  /// \param LimitPartition Upper bound on the number of partitions to
+  ///          produce. Pass 0 (the default) to derive the partition count
+  ///          from the number of call-graph roots discovered in
+  ///          createWorkList (one root per partition at most). The actual
+  ///          partition count is finalized in the constructor.
+  SplitModuleCG(Module &M, unsigned LimitPartition = 0);
+  void SplitModule(ModuleCreationCallback ModuleCallback,
```
teresajohnson, line 200: method should be lowerCamelCase

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+#include "llvm/Transforms/Utils/ModuleUtils.h"
+#include <thread>
+using namespace llvm;
+
+#define DEBUG_TYPE "split-module-cg"
+
+namespace {
+
+static cl::opt<bool>
+    enablePrintSimplifiedCallGraph("enable-print-simplified-callgraph",
```
teresajohnson, line 19: variables should be UpperCamelCase

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+  for (auto &Fn : M) {
+    if (Fn.isDeclaration())
+      continue;
+
+    CostType FnCost = 0;
+    for (const auto &BB : Fn)
+      FnCost += std::distance(BB.begin(), BB.end());
+    assert(FnCost != 0);
+    FuncsCosts[&Fn] = FnCost;
+    assert((ModuleCost + FnCost) >= ModuleCost && "Overflow!");
```
teresajohnson, line 99: InstCost is signed, so this isn't going to detect overflow (it would be UB). Can you use some unsigned types for your checking?

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+    // tracking the cost of newly inserted functions so the balancing queue
+    // can be updated. CurFn.Dependencies includes the root F itself.
+    auto &FnsInPart = Partitions[PID];
+    CostType AddedCost = 0;
+    for (const Function *Dep : CurFn.Dependencies)
+      if (FnsInPart.insert(Dep).second)
+        AddedCost += FuncsCosts.lookup(Dep);
+
+    // Update the balancing queue. We scan backwards because in the common
+    // case the target partition is at the end of the sorted queue.
```
teresajohnson, line 75: Not sure I follow why the reverse is needed - there is no early exit so how does the order matter? It looks like PID is the back() entry of the BalancingQueue, so should the continue be a break? Or can you just replace this loop with BalancingQueue.back().second += AddedCost.

llvm/lib/Transforms/Utils/SplitModuleCG.cpp
```diff
@@ -0,0 +1,360 @@
...
+    // serialization is the supported way to move a Module between contexts.
+    SmallString<0> BC;
+    raw_svector_ostream BCOS(BC);
+    WriteBitcodeToFile(*MPart, BCOS);
+    MPart.reset();
+    Threads.emplace_back(
+        [&, I](SmallString<0> BC) {
+          llvm::lto::LTOLLVMContext Ctx(C);
+          Expected<std::unique_ptr<Module>> MOrErr =
+              parseBitcodeFile(MemoryBufferRef(BC.str(), "ld-temp.o"), Ctx);
```
teresajohnson, line 280: ld-temp.o is an LTO-ism, can you use something else? Also, can you make the name different depending on the partition index?

llvm/include/llvm/Transforms/Utils/SplitModuleCG.h
```diff
@@ -0,0 +1,245 @@
...
+  CostType TotalCost = 0;
+};
+
+/// Splits a module into N linkable partitions by traversing its call graph,
+/// so that each partition carries a self-consistent subset of functions
+/// (a root + its callees) and is balanced by IR-instruction cost. The
+/// resulting partitions can be optimized and codegen'd in parallel by the
+/// LTO backend and merged back into a single object.
+///
+/// Workflow (driven by SplitModule):
```
teresajohnson, line 170: Is the order listed here up to date?

llvm/tools/llvm-split/llvm-split.cpp
```diff
@@ -77,6 +79,10 @@ static cl::opt<std::string>
     MCPU("mcpu", cl::desc("Target CPU, ignored if --mtriple is not used"),
          cl::value_desc("cpu"), cl::cat(SplitCategory));
 
+static cl::opt<bool> EnableCallGraphSplitModule(
+    "enable-call-graph-split-module", cl::Prefix, cl::init(false),
```
teresajohnson, line 83: I don't think you want to use cl::Prefix here on a bool flag?

llvm/test/tools/llvm-split/SplitModuleCG/unnamed.ll
```diff
@@ -0,0 +1,19 @@
+; RUN: llvm-split -enable-call-graph-split-module=true -j2 -o %t %s
+; RUN: llvm-dis -o - %t0 | FileCheck --check-prefix=CHECK0 %s
+
+; Test that an unnamed internal function (@0) is given a stable name and
+; .llvm.<suffix> suffix after promotion.
+
+; CHECK0-DAG: declare hidden void @__llvm_unnamed.llvm.{{[0-9a-f]+}}()
+; CHECK0-DAG: define hidden void @__llvm_unnamed.1.llvm.{{[0-9a-f]+}}()
+
+; CHECK1-DAG: define hidden void @__llvm_unnamed.llvm.{{[0-9a-f]+}}()
```
teresajohnson, line 10: I don't think anything is testing the CHECK1 lines (no associated FileCheck)

llvm/test/tools/llvm-split/SplitModuleCG/partition-cap.ll
```diff
@@ -0,0 +1,12 @@
+; RUN: llvm-split -enable-call-graph-split-module=true -j10 -o %t %s
+; RUN: llvm-dis -o - %t0 | FileCheck --check-prefix=CHECK0 %s
+; RUN: llvm-dis -o - %t1 | FileCheck --check-prefix=CHECK1 %s
+
+; Test that partition count is capped to the number of entry functions
+; (-j10 but only 2 roots → 2 outputs).
```
teresajohnson, line 6: Should this check that there are no other output files created?

teresajohnson (commented):
A few more questions after another pass.

I'd also like to hear from @shiltian who expressed concerns early on about additional module splitting utilities, to see if he has any concerns moving forward.

I notice that SplitModule.cpp has some handling to avoid duplicated module inline asm: https://github.com/llvm/llvm-project/blob/2cf68b2cc466710badbe560ad6081f71c72ed331/llvm/lib/Transforms/Utils/SplitModule.cpp#L294. Is something similar needed here?

Also, a couple of other requests:
- Make the title of this one about the change in the PR.
- Update the description to talk about what is in this PR specifically as well, but you can keep the high level description about the bigger picture and overall patch series below that.
- It's a lot harder to re-review after force pushes (which are generally discouraged by llvm because of that) although I know it is difficult to avoid for stacked PRs - can you look into using something like graphite to manage the stacked PRs? That will force push but has a reviewer interface that makes it easy to see the relevant updates for each PR. For this one if you can just avoid force pushing while the remaining review is in progress that would be helpful, so I can see what changed since my last review.

shiltian:
> I'd also like to hear from @shiltian who expressed concerns early on about additional module splitting utilities, to see if he has any concerns moving forward.

Thanks for tagging. I'll take a look later.

PR #198705, opened 2026-05-20
[PowerPC] Drop invalid range metadata when lowering i64 load to fp in INT_TO_FP

amy-kwan:
When lowering an i64 load in LowerINT_TO_FP, we were forwarding the original !range metadata to a new f64 load. This is invalid because the metadata no longer matches the value type/semantics, and can trigger assertions when lowering i64 to fp (double or float) conversions.

This patch fixes this by passing a nullptr for the Ranges operand when calling getLoad() and adds extra test cases to cover signed/unsigned i64 to f32/f64 conversions and to ensure they do not assert when the !range metadata is present.

The assertion this patch attempts to fix is:
```
Assertion failed: (!MMO->getRanges() || (mdconst::extract<ConstantInt>(MMO->getRanges()->getOperand(0)) ->getBitWidth() == MemVT.getScalarSizeInBits() && MemVT.isInteger())) && "Range metadata and load type must match!"
```
This assert was originally seen when building Rust on AIX.


llvm/test/CodeGen/PowerPC/ppc-i64-to-fp.ll
```diff
@@ -0,0 +1,124 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 6
+; RUN: llc -mcpu=pwr8 -ppc-asm-full-reg-names -mtriple=powerpc64-ibm-aix < %s | \
+; RUN:   FileCheck %s --check-prefix=CHECK-AIX64
+; RUN: llc -mcpu=pwr8 -ppc-asm-full-reg-names -mtriple=powerpc-ibm-aix < %s | \
+; RUN:   FileCheck %s --check-prefix=CHECK-AIX32
+; RUN: llc -mcpu=pwr8 -ppc-asm-full-reg-names -mtriple=powerpc64le-unknown-linux-gnu < %s | \
+; RUN:   FileCheck %s --check-prefix=CHECK-LINUX
```
lei137: nit: maybe we can just use default CHECK for 64bit runs since I don't see a diff in the asm generated.

llvm/lib/Target/PowerPC/PPCISelLowering.cpp
```diff
@@ -8865,7 +8865,7 @@ SDValue PPCTargetLowering::LowerINT_TO_FP(SDValue Op,
     MachineFunction &MF = DAG.getMachineFunction();
     if (canReuseLoadAddress(SINT, MVT::i64, RLI, DAG)) {
       Bits = DAG.getLoad(MVT::f64, dl, RLI.Chain, RLI.Ptr, RLI.MPI,
-                         RLI.Alignment, RLI.MMOFlags(), RLI.AAInfo, RLI.Ranges);
```
lei137, line 8878: nit: would be good to add a comment as to why this is correct.
eg.
```
// Drop range metadata - invalid for f64 bit reinterpretation of i64 values.
```

lei137 (approved):
LGTM with nits.

llvm/lib/Target/PowerPC/PPCISelLowering.cpp
amy-kwan: Great point, thanks Lei!

PR #198710, opened 2026-05-20
[OpenMPOpt][Attributor] Selectively seed deglobalization AAs

nolongerwait:
This addresses a compile-time issue observed on a large generated C++ translation unit compiled with `-fopenmp`.

The source code is not OpenMP-heavy. It mainly consists of generated function-registration wrappers, template instantiations, lambdas, and small helper functions. However, because the TU is compiled with OpenMP enabled, `OpenMPOptCGSCCPass` runs and drives Attributor on a module with many functions.

`OpenMPOpt::registerAAsForFunction` currently eagerly creates the deglobalization AAs for every function in OpenMP device modules:

* `AAHeapToShared`
* `AAHeapToStack`

Most generated wrapper/helper functions in the motivating workload do not contain `__kmpc_alloc_shared`, removable allocations, or free-like calls. In those cases, the deglobalization AAs have no useful candidates but still contribute to Attributor initialization and fixed-point work.

This patch makes the deglobalization AA seeding more selective:

* only create `AAHeapToShared` for functions with non-empty `__kmpc_alloc_shared` uses;
* only create `AAHeapToStack` for functions containing removable allocation or free-like call candidates;
* make `AAHeapToShared` use the per-function runtime use vector instead of scanning all users of the runtime declaration.

I also measured the previously attempted pieces independently. The compile-time improvement comes from the selective seeding change. The Attributor use-caching change and the `registerAAsForFunction` registration cache did not show meaningful benefit on the motivating workload, so they have been removed from this PR.

On the motivating TU, selective seeding reduced total compile time from about 3473s to 271s, and `OpenMPOptCGSCCPass` time from about 3232s to 111s.

### DIFF
--- a/llvm/include/llvm/Transforms/IPO/Attributor.h
+++ b/llvm/include/llvm/Transforms/IPO/Attributor.h
@@ -142,6 +142,7 @@
 
 #include <limits>
 #include <map>
+#include <memory>
 #include <optional>
 
 namespace llvm {
@@ -2487,6 +2488,10 @@ private:
   /// from a live entry point. The functions are added to ToBeDeletedFunctions.
   void identifyDeadInternalFunctions();
 
+  /// Return all uses of \p Fn after looking through pointer cast constant
+  /// expressions.
+  ArrayRef<const Use *> getPotentialCallSiteUses(const Function &Fn);
+
   /// Run `::update` on \p AA and track the dependences queried while doing so.
   /// Also adjust the state if we know further updates are not necessary.
   LLVM_ABI ChangeStatus updateAA(AbstractAttribute &AA);
@@ -2553,6 +2558,11 @@ private:
 
   /// A set to remember the functions we already assume to be live and visited.
   DenseSet<const Function *> VisitedFunctions;
+
+  /// Cached function uses after looking through pointer cast constant
+  /// expressions. Only used before the Attributor starts changing the IR.
+  DenseMap<const Function *, std::unique_ptr<SmallVector<const Use *, 8>>>
+      PotentialCallSiteUsesMap;
 
   /// Uses we replace with a new value after manifest is done. We will remove
   /// then trivially dead instructions as well.
--- a/llvm/include/llvm/Transforms/IPO/Attributor.h
+++ b/llvm/include/llvm/Transforms/IPO/Attributor.h
@@ -142,7 +142,6 @@
 
 #include <limits>
 #include <map>
-#include <memory>
 #include <optional>
 
 namespace llvm {
@@ -2488,10 +2487,6 @@ private:
   /// from a live entry point. The functions are added to ToBeDeletedFunctions.
   void identifyDeadInternalFunctions();
 
-  /// Return all uses of \p Fn after looking through pointer cast constant
-  /// expressions.
-  ArrayRef<const Use *> getPotentialCallSiteUses(const Function &Fn);
-
   /// Run `::update` on \p AA and track the dependences queried while doing so.
   /// Also adjust the state if we know further updates are not necessary.
   LLVM_ABI ChangeStatus updateAA(AbstractAttribute &AA);
@@ -2558,11 +2553,6 @@ private:
 
   /// A set to remember the functions we already assume to be live and visited.
   DenseSet<const Function *> VisitedFunctions;
-
-  /// Cached function uses after looking through pointer cast constant
-  /// expressions. Only used before the Attributor starts changing the IR.
-  DenseMap<const Function *, std::unique_ptr<SmallVector<const Use *, 8>>>
-      PotentialCallSiteUsesMap;
 
   /// Uses we replace with a new value after manifest is done. We will remove
   /// then trivially dead instructions as well.
--- a/llvm/include/llvm/Transforms/IPO/Attributor.h
+++ b/llvm/include/llvm/Transforms/IPO/Attributor.h
@@ -142,6 +142,7 @@
 
 #include <limits>
 #include <map>
+#include <memory>
 #include <optional>
 
 namespace llvm {
@@ -2521,6 +2522,10 @@ private:
   /// from a live entry point. The functions are added to ToBeDeletedFunctions.
   void identifyDeadInternalFunctions();
 
+  /// Return all uses of \p Fn after looking through pointer cast constant
+  /// expressions.
+  ArrayRef<const Use *> getPotentialCallSiteUses(const Function &Fn);
+
   /// Run `::update` on \p AA and track the dependences queried while doing so.
   /// Also adjust the state if we know further updates are not necessary.
   LLVM_ABI ChangeStatus updateAA(AbstractAttribute &AA);
@@ -2587,6 +2592,11 @@ private:
 
   /// A set to remember the functions we already assume to be live and visited.
   DenseSet<const Function *> VisitedFunctions;
+
+  /// Cached function uses after looking through pointer cast constant
+  /// expressions. Only used before the Attributor starts changing the IR.
+  DenseMap<const Function *, std::unique_ptr<SmallVector<const Use *, 8>>>
+      PotentialCallSiteUsesMap;
 
   /// Uses we replace with a new value after manifest is done. We will remove
   /// then trivially dead instructions as well.
--- a/llvm/lib/Transforms/IPO/Attributor.cpp
+++ b/llvm/lib/Transforms/IPO/Attributor.cpp
@@ -1954,6 +1954,40 @@ bool Attributor::checkForAllCallSites(function_ref<bool(AbstractCallSite)> Pred,
                               &QueryingAA, UsedAssumedInformation);
 }
 
+ArrayRef<const Use *> Attributor::getPotentialCallSiteUses(const Function &Fn) {
+  std::unique_ptr<SmallVector<const Use *, 8>> &PotentialUsesPtr =
+      PotentialCallSiteUsesMap[&Fn];
+  if (PotentialUsesPtr)
+    return *PotentialUsesPtr;
+
+  PotentialUsesPtr = std::make_unique<SmallVector<const Use *, 8>>();
+  SmallVectorImpl<const Use *> &PotentialUses = *PotentialUsesPtr;
+
+  SmallVector<const Use *, 8> Worklist;
+  SmallPtrSet<const Use *, 8> Seen;
+  auto AddUse = [&](const Use &U) {
+    if (Seen.insert(&U).second)
+      Worklist.push_back(&U);
+  };
+
+  for (const Use &U : Fn.uses())
+    AddUse(U);
+
+  for (unsigned UIdx = 0; UIdx < Worklist.size(); ++UIdx) {
+    const Use &U = *Worklist[UIdx];
+    if (ConstantExpr *CE = dyn_cast<ConstantExpr>(U.getUser())) {
+      if (CE->isCast() && CE->getType()->isPointerTy()) {
+        for (const Use &CEU : CE->uses())
+          AddUse(CEU);
+        continue;
+      }
+    }
+    PotentialUses.push_back(&U);
+  }
+
+  return PotentialUses;
+}
+
 bool Attributor::checkForAllCallSites(function_ref<bool(AbstractCallSite)> Pred,
                                       const Function &Fn,
                                       bool RequireAllCallSites,
@@ -1972,9 +2006,16 @@ bool Attributor::checkForAllCallSites(function_ref<bool(AbstractCallSite)> Pred,
     if (!CB(*this, QueryingAA))
       return false;
 
-  SmallVector<const Use *, 8> Uses(make_pointer_range(Fn.uses()));
-  for (unsigned u = 0; u < Uses.size(); ++u) {
-    const Use &U = *Uses[u];
+  SmallVector<const Use *, 8> Uses;
+  ArrayRef<const Use *> PotentialUses;
+  if (Phase == AttributorPhase::SEEDING || Phase == AttributorPhase::UPDATE)
+    PotentialUses = getPotentialCallSiteUses(Fn);
+  else {
+    append_range(Uses, make_pointer_range(Fn.uses()));
+    PotentialUses = Uses;
+  }
+  for (unsigned u = 0; u < PotentialUses.size(); ++u) {
+    const Use &U = *PotentialUses[u];
     DEBUG_WITH_TYPE(VERBOSE_DEBUG_TYPE, {
       if (auto *Fn = dyn_cast<Function>(U))
         dbgs() << "[Attributor] Check use: " << Fn->getName() << " in "
@@ -1996,8 +2037,12 @@ bool Attributor::checkForAllCallSites(function_ref<bool(AbstractCallSite)> Pred,
           dbgs() << "[Attributor] Use, is constant cast expression, add "
                  << CE->getNumUses() << " uses of that expression instead!\n";
         });
+        assert((Phase != AttributorPhase::SEEDING &&
+                Phase != AttributorPhase::UPDATE) &&
+               "Constant expression uses should be expanded in the cache");
         for (const Use &CEU : CE->uses())
           Uses.push_back(&CEU);
+        PotentialUses = Uses;
         continue;
       }
     }
--- a/llvm/lib/Transforms/IPO/Attributor.cpp
+++ b/llvm/lib/Transforms/IPO/Attributor.cpp
@@ -2008,9 +2008,9 @@ bool Attributor::checkForAllCallSites(function_ref<bool(AbstractCallSite)> Pred,
 
   SmallVector<const Use *, 8> Uses;
   ArrayRef<const Use *> PotentialUses;
-  if (Phase == AttributorPhase::SEEDING || Phase == AttributorPhase::UPDATE)
+  if (Phase == AttributorPhase::SEEDING || Phase == AttributorPhase::UPDATE) {
     PotentialUses = getPotentialCallSiteUses(Fn);
-  else {
+  } else {
     append_range(Uses, make_pointer_range(Fn.uses()));
     PotentialUses = Uses;
   }
--- a/llvm/lib/Transforms/IPO/Attributor.cpp
+++ b/llvm/lib/Transforms/IPO/Attributor.cpp
@@ -1954,44 +1954,6 @@ bool Attributor::checkForAllCallSites(function_ref<bool(AbstractCallSite)> Pred,
                               &QueryingAA, UsedAssumedInformation);
 }
 
-ArrayRef<const Use *> Attributor::getPotentialCallSiteUses(const Function &Fn) {
-  assert(
-      (Phase == AttributorPhase::SEEDING || Phase == AttributorPhase::UPDATE) &&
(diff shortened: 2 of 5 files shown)


llvm/lib/Transforms/IPO/Attributor.cpp
```diff
@@ -1958,6 +1958,40 @@ bool Attributor::checkForAllCallSites(function_ref<bool(AbstractCallSite)> Pred,
...
+  SmallPtrSet<const Use *, 8> Seen;
+  auto AddUse = [&](const Use &U) {
+    if (Seen.insert(&U).second)
+      Worklist.push_back(&U);
+  };
+
+  for (const Use &U : Fn.uses())
+    AddUse(U);
+
+  for (unsigned UIdx = 0; UIdx < Worklist.size(); ++UIdx) {
```
Enna1:
```suggestion
  for (auto *U: Worklist) {
```

llvm/lib/Transforms/IPO/Attributor.cpp
```diff
@@ -1976,9 +2010,16 @@ bool Attributor::checkForAllCallSites(function_ref<bool(AbstractCallSite)> Pred,
     if (!CB(*this, QueryingAA))
       return false;
 
-  SmallVector<const Use *, 8> Uses(make_pointer_range(Fn.uses()));
-  for (unsigned u = 0; u < Uses.size(); ++u) {
-    const Use &U = *Uses[u];
+  SmallVector<const Use *, 8> Uses;
+  ArrayRef<const Use *> PotentialUses;
+  if (Phase == AttributorPhase::SEEDING || Phase == AttributorPhase::UPDATE)
```
Enna1: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
Use braces for the `if` block to keep it uniform with the `else` block.



```suggestion
  if (Phase == AttributorPhase::SEEDING || Phase == AttributorPhase::UPDATE) {
    PotentialUses = getPotentialCallSiteUses(Fn);
  }
```

llvm/lib/Transforms/IPO/Attributor.cpp
```diff
@@ -1976,9 +2010,16 @@ bool Attributor::checkForAllCallSites(function_ref<bool(AbstractCallSite)> Pred,
...
+    PotentialUses = getPotentialCallSiteUses(Fn);
+  else {
+    append_range(Uses, make_pointer_range(Fn.uses()));
+    PotentialUses = Uses;
+  }
+  for (unsigned u = 0; u < PotentialUses.size(); ++u) {
```
Enna1:
```suggestion
  for (auto *U : PotentialUses) {
```

llvm/lib/Transforms/IPO/Attributor.cpp
nolongerwait: These loops intentionally use an index-based form because the loop body may append new elements to the same container being traversed. This makes it inconvenient, and potentially unsafe, to rewrite them as range-based for loops. Also, this code was already using the traditional index-based for form before this change.

llvm/lib/Transforms/IPO/Attributor.cpp
nolongerwait: Same reason as above.

llvm/lib/Transforms/IPO/Attributor.cpp
Enna1: Thanks for pointing this!

llvm/test/Transforms/Attributor/reduced/openmp_opt_constant_type_crash.ll
```diff
@@ -70,13 +70,12 @@ cond.end:                                         ; preds = %cond.true, %entry
...
-; CHECK-SAME: () #[[ATTR0:[0-9]+]] {
+; CHECK-LABEL: define {{[^@]+}}@_ZN6Kokkos4Impl11ViewMappingIvJNS_10ViewTraitsIPPdJNS_11LayoutRightENS_18ScratchMemorySpaceINS_12Experimental12OpenMPTargetEEENS_12MemoryTraitsILj1EEEEEENS0_5ALL_tEiEE6assignINS2_IS3_JNS_12LayoutStrideENS_6DeviceIS8_S9_EESB_EEEEEvRNS1_IT_JvEEERKNS1_ISC_JvEEESD_i.internalized() {
 ; CHECK-NEXT:  entry:
 ; CHECK-NEXT:    [[EXTENTS11:%.*]] = alloca [0 x [0 x %"struct.Kokkos::Impl::SubviewExtents.448"]], i32 0, align 8, addrspace(5)
 ; CHECK-NEXT:    [[EXTENTS_ASCAST:%.*]] = addrspacecast ptr addrspace(5) [[EXTENTS11]] to ptr
 ; CHECK-NEXT:    call void @_ZN6Kokkos4Impl14SubviewExtentsILj2ELj1EEC2IJLm0ELm0EEJNS0_5ALL_tEiEEERKNS0_13ViewDimensionIJXspT_EEEEDpT0_.internalized(ptr [[EXTENTS_ASCAST]])
+; CHECK-NEXT:    call void @_ZN6Kokkos4Impl10ViewOffsetINS0_13ViewDimensionIJLm0EEEENS_12LayoutStrideEvEC2INS2_IJLm0ELm0EEEENS_11LayoutRightEEERKNS1_IT_T0_vEERKNS0_14SubviewExtentsIXsrS9_4rankELj1EEE.internalized(ptr [[EXTENTS_ASCAST]])
```
abidh, line 78: Do you have an idea of the performance impact of these changes. It seems that we are now generating a call which was previously optimized. Similarly, we are losing `norecurse nosync nounwind memory(argmem: write)` attributes on the function below. These may be valid trade-offs but it will be good to have some numbers.

llvm/lib/Transforms/IPO/Attributor.cpp
```diff
@@ -1958,6 +1958,40 @@ bool Attributor::checkForAllCallSites(function_ref<bool(AbstractCallSite)> Pred,
                               &QueryingAA, UsedAssumedInformation);
 }
 
+ArrayRef<const Use *> Attributor::getPotentialCallSiteUses(const Function &Fn) {
+  std::unique_ptr<SmallVector<const Use *, 8>> &PotentialUsesPtr =
+      PotentialCallSiteUsesMap[&Fn];
```
abidh: This is documented to be only used before Attributor starts changing the IR. Should that be asserted here.

shiltian:
> Although the source-level code is not primarily an OpenMP runtime-heavy workload, the generated IR contains OpenMP device metadata, so `OpenMPOptCGSCCPass` is enabled.

Can you explain more about this? I don't actually follow this.

nolongerwait:
> > Although the source-level code is not primarily an OpenMP runtime-heavy workload, the generated IR contains OpenMP device metadata, so `OpenMPOptCGSCCPass` is enabled.
> 
> Can you explain more about this? I don't actually follow this.

Sorry, my previous wording was imprecise.Initially, considering the company's compliance risks, I could only describe the situation we encountered in a very abstract way, but after discussion, this information can be disclosed.

So the more accurate statement is that this TU is compiled with `-fopenmp` globally, which makes the IR an OpenMP module and causes `OpenMPOptCGSCCPass` to run. The source itself is mostly generated function-registration/wrapper code rather than explicit OpenMP code. That generated code expands into many lambdas, template instantiations, and helper functions, which exposes the compile-time cost of eagerly initializing OpenMPOpt/Attributor state for functions that are not useful candidates for the intended deglobalization optimization.

I will also update the previous PR description accordingly.

llvm/test/Transforms/Attributor/reduced/openmp_opt_constant_type_crash.ll
nolongerwait: Thanks for pointing this out. The changed checks here are indeed a trade-off from avoiding eager deglobalization AA registration.

The motivation is compile time: on the motivating production TU, OpenMPOptCGSCCPass went from about 3085.7s to 85.5s, and total Attributor::run went from about 2490.5s to 76.5s.

I also tested this patch on several internal ByteDance cases. Only two cases showed up in the final compile-time report, and both were compile-time improvements: one is the motivating case mentioned in the PR description, and another went from about 15 minutes to about 14 minutes. The other internal cases did not show significant compile-time changes.

In addition, Enna1 helped run the llvm-opt-benchmark compile-time suite: dtcxzyw/llvm-opt-benchmark-nightly#339. That report shows the overall compile-time impact is essentially neutral, with file-level GeoMean +0.01% and project-level GeoMean -0.00%.

So given the very large compile-time win on the motivating OpenMP device module, another internal compile-time improvement, no significant changes in our other internal cases, and no broad compile-time regression in the public benchmark suite, I think this maybe an acceptable trade-off.

llvm/lib/Transforms/IPO/Attributor.cpp
nolongerwait: thx, adjusted this part accordingly. :)

llvm/test/Transforms/Attributor/reduced/openmp_opt_constant_type_crash.ll
abidh: Sorry I was not clear in my comment. My question was about runtime performance. Do we have any idea of the runtime impact of this change?

nolongerwait: Thanks for clarifying. I do not have runtime numbers for this yet. The motivating case for this patch was a compile-time issue, so the measurements I collected so far are compile-time only.

I will take another look and see if there is a better way to make this change, or otherwise try to understand the runtime impact more clearly.

llvm/include/llvm/Transforms/IPO/Attributor.h
```diff
@@ -2588,6 +2593,11 @@ struct Attributor {
   /// A set to remember the functions we already assume to be live and visited.
   DenseSet<const Function *> VisitedFunctions;
 
+  /// Cached function uses after looking through pointer cast constant
+  /// expressions. Only used before the Attributor starts changing the IR.
+  DenseMap<const Function *, std::unique_ptr<SmallVector<const Use *, 8>>>
```
shiltian: A map to a unique pointer to a small vector smells like a really bad design.

shiltian (commented):
I don't think this is a proper fix. A better fix would be to skip OpenMPOpt pass if there is no actual OpenMP code in the module. That said, we probably want to change how we mark a module instead.

CC @kevinsala

nolongerwait:
> I don't think this is a proper fix. A better fix would be to skip OpenMPOpt pass if there is no actual OpenMP code in the module. That said, we probably want to change how we mark a module instead.
> 
> CC @kevinsala

Thanks, I agree that skipping OpenMPOpt for modules without actual OpenMP code would be a better solution in principle.

We did consider that first when investigating the compile-time issue. The difficulty is that the current module-level marker seems too coarse for that decision.

The change I tried here was an attempt to make that gating more precise by looking for concrete OpenMP signals in the IR, such as runtime uses, device kernels, offload entries, or device-runtime artifacts. I agree, though, that this is still heuristic and probably not the ideal abstraction boundary.

One additional constraint on our side is that our team does not currently have deep OpenMP expertise, so it is difficult for us to propose a fully convincing design-level fix without guidance from OpenMP maintainers. We can investigate and iterate, but we would appreciate advice on what the intended signal should be.

jdoerfert:
Honestly, this is not bad.

We should be careful creating AAs that have no chance of optimizing the code. So if we can determine they are not helpful, why create them.

The PR contains ~3 different things rolled up in one.
I do not understand if/how the caching of the uses or the function register helps.

Would it be possible to measure the impact of the 3 parts in isolation?
- Use caching
- Selective seeding
- registerAAsForFunction cache (Do we really call this multiple times right now?)

jnudhf:
> Honestly, this is not bad.
> 
> We should be careful creating AAs that have no chance of optimizing the code. So if we can determine they are not helpful, why create them.
> 
> The PR contains ~3 different things rolled up in one. I do not understand if/how the caching of the uses or the function register helps.
> 
> Would it be possible to measure the impact of the 3 parts in isolation?
> 
> * Use caching
> * Selective seeding
> * registerAAsForFunction cache (Do we really call this multiple times right now?)

Thanks for the suggestion. I measured the three parts in isolation on the motivating TU.

All variants were built from the same baseline. The main wall-clock metrics are total compile time and `OpenMPOptCGSCCPass` time.

| Variant | Change | Total compile time (s) | OpenMPOptCGSCCPass (s) | AAHeapToShared created | AAHeapToStack created | register duplicate skips | use-cache q/h/m |
|---|---|---:|---:|---:|---:|---:|---:|
| V0 | baseline | 3473.360 | 3231.903 | 228657594 | 228657594 | 0 | 0/0/0 |
| V1 | use caching only | 3867.030 | 3573.213 | 228657594 | 228657594 | 0 | 0/0/0 |
| V2 | selective seeding only | 271.190 | 111.134 | 0 | 32277 | 0 | 0/0/0 |
| V3 | `registerAAsForFunction` cache only | 3539.980 | 3293.664 | 228657594 | 228657594 | 1 | 0/0/0 |
| V4 | full original patch | 260.710 | 102.897 | 0 | 32277 | 1 | 0/0/0 |

The result confirms that the dominant improvement comes from selective seeding. It reduces total compile time from 3473s to 271s, and `OpenMPOptCGSCCPass` time from 3232s to 111s.

The `use-caching` change did not show a meaningful benefit on this workload. The cache counters stayed at 0/0/0, and the wall-clock result was not better than baseline. I would treat the difference from baseline as noise rather than as evidence that the caching is harmful.

The `registerAAsForFunction cache` also did not show a meaningful benefit here. It only skipped one duplicate registration, which answers your question: for this workload, we do not appear to call `registerAAsForFunction` repeatedly enough for that cache to matter.

I also rechecked the difference between `selective seeding` only and the full original patch. They are effectively equivalent within measurement noise, so I do not think the use-caching or registration-cache parts are needed to explain the improvement.

Based on this ablation, I updated the PR to focus on the `selective seeding` part and removed the `use-caching` and `registerAAsForFunction` cache changes for now.

llvm/test/Transforms/Attributor/reduced/openmp_opt_constant_type_crash.ll
jnudhf: @abidh Thanks for raising the runtime-performance concern.

The original motivating workload is an internal compile-time stress case. It is useful for reproducing the OpenMPOpt/Attributor compile-time issue, but its runtime performance is not straightforward to quantify or compare in a standardized way. Therefore, I also ran a SPEC CPU2017 fp speed OpenMP subset as a runtime-performance sanity check. This SPEC run is not intended to reproduce the original compile-time issue; that issue was observed on some internal OpenMP codes where the compile time could increase by tens of times. The goal here was only to check whether the simplified patch introduces an obvious runtime-performance regression compared with `main`.

Setup:

- clang/flang + libomp
- SPEC CPU2017 fpspeed OpenMP subset
- 3 iterations
- 9 scored benchmarks + `996.specrand_fs`
- `628.pop2_s` was excluded due to an unrelated flang module parsing issue

Observed result on the tested subset:

| Comparison | Geomean | Build time |
|---|---:|---:|
| `main` | 182.569 | 19:45 |
| simplified patch | 186.095 | 19:38 |

Relative to `main`, the tested subset is roughly flat and shows no noticeable runtime-performance regression. The geomean is slightly higher in this run (+1.93%), but I would treat that conservatively as normal benchmark/run-to-run variation rather than claiming a runtime improvement.

So far, I do not observe a runtime-performance regression from the simplified patch in this SPEC-based sanity check. The intent of this patch remains compile-time reduction / implementation simplification, not runtime-performance improvement.
[CPU2017.036.fpspeed.refspeed.txt](https://github.com/user-attachments/files/28737592/CPU2017.036.fpspeed.refspeed.txt)
[CPU2017.044.fpspeed.refspeed.txt](https://github.com/user-attachments/files/28737593/CPU2017.044.fpspeed.refspeed.txt)

nolongerwait:
The latest check failure appears to be due to the new test introduced by 564d69a8dec3183ebd78a779efdd8903bbe89f68 on main, rather than this patch itself.

I rebased this PR locally onto the latest main and the relevant test passes there. I can push the rebase after review to refresh the CI results.

llvm/lib/Transforms/IPO/OpenMPOpt.cpp
```diff
@@ -5585,13 +5589,19 @@ void OpenMPOpt::registerAAs(bool IsModulePass) {
...
+  IRPosition FPos = IRPosition::function(F);
+  A.getOrCreateAAFor<AAExecutionDomain>(FPos);
   if (F.hasFnAttribute(Attribute::Convergent))
-    A.getOrCreateAAFor<AANonConvergent>(IRPosition::function(F));
+    A.getOrCreateAAFor<AANonConvergent>(FPos);
+
+  const bool FunctionUsesSharedAlloc =
+      !DisableOpenMPOptDeglobalization &&
+      OMPInfoCache.RFIs[OMPRTL___kmpc_alloc_shared].getUseVector(
+          const_cast<Function &>(F));
```
jdoerfert: How does this convert to bool? Can we add an explicit empty call?

llvm/lib/Transforms/IPO/OpenMPOpt.cpp
```diff
@@ -5625,6 +5641,11 @@ void OpenMPOpt::registerAAsForFunction(Attributor &A, const Function &F) {
       }
     }
   }
+
+  if (FunctionUsesSharedAlloc)
+    A.getOrCreateAAFor<AAHeapToShared>(FPos);
+  if (!DisableOpenMPOptDeglobalization && HasHeapToStackCandidate)
```
jdoerfert: !Disable should not be needed anymore here.

jdoerfert (approved):
LG, minor nits.

nolongerwait:
Thanks @jnudhf for the help and the additional testing/validation.

Since the final patch has changed somewhat from the original version during review, I have updated the PR title and description to better reflect the changes that are still included in the final version.

llvm/lib/Transforms/IPO/OpenMPOpt.cpp
```diff
@@ -5585,13 +5589,23 @@ void OpenMPOpt::registerAAs(bool IsModulePass) {
...
-    A.getOrCreateAAFor<AANonConvergent>(IRPosition::function(F));
+    A.getOrCreateAAFor<AANonConvergent>(FPos);
+
+  bool FunctionUsesSharedAlloc = false;
+  if (!DisableOpenMPOptDeglobalization) {
+    const OMPInformationCache::RuntimeFunctionInfo::UseVector *SharedAllocUses =
+        OMPInfoCache.RFIs[OMPRTL___kmpc_alloc_shared].getUseVector(
+            const_cast<Function &>(F));
+    FunctionUsesSharedAlloc =
+        SharedAllocUses != nullptr && !SharedAllocUses->empty();
```
shiltian: nit:

```suggestion
    FunctionUsesSharedAlloc = SharedAllocUses && !SharedAllocUses->empty();
```

plus a clang-format if needed

shiltian (approved):

PR #198714, opened 2026-05-20
[Clang][Darwin] Forward profile-sample-file when constructing link job

mustartt:
On Darwin forward `-mllvm -sample-profile-file` when the driver is supplied `-fsample-profile-use`. Also `-sample-profile-file` is silently ignored by the ordering of the linker, so we can temporarily reparse PGOOpts after `ThinLTOCodeGenerator` is constructed.


cachemeifyoucan (approved):
LGTM

PR #198717, opened 2026-05-20
[alpha.webkit.UncheckedCallArgsChecker] EnsureFunctionAnalysis allows unsafe code

rniwa:
Recall that "ensure" function in WebKit is a pattern for lazy initialization. "ensure" functions returns a member variable's value if the object had been initialized before, and otherwise creates the relevant object and returns it.

EnsureFunctionAnalysis has two bugs which allow unsafe code:
 * It allows a virtual "ensure" function that doesn't use literal "virtual keyword".
 * It allows non-trivial code to appear within an "ensure" function.

This PR addresses both of these bugs by replacing isVirtual with isVirtualAsWritten and checking the "ensure" function body's trivial-ness with TrivialFunctionAnalysis.

### DIFF
--- a/clang/lib/StaticAnalyzer/Checkers/WebKit/ASTUtils.cpp
+++ b/clang/lib/StaticAnalyzer/Checkers/WebKit/ASTUtils.cpp
@@ -403,15 +403,18 @@ public:
   }
 };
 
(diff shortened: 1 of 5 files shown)


rniwa:
Actually, I need to think about this problem some more.

PR #198718, opened 2026-05-20
[ELF] Reduce symbol partitions to shim binaries

MaskRay:
The experimental symbol partition feature can move arbitrary symbols and
input sections out of the main partition into named loadable partitions
that consumers extract with `llvm-objcopy --extract-partition`. It is
incompatible with many features, has seen no adoption beyond Chromium's
`build/extract_partition.py`. Chrome hasn't needed this feature for many
years.

The partitioning machinery (per-partition `MarkLive` runs, `moveToMain`,
cross-partition pull in `InputSection::replace`, per-symbol
dynsym/gnuhash/thunk-compat filters, `PartitionIndexSection`,
`copySectionsIntoPartitions`, etc.) does not pull its weight for that.

Remove the partitioning brain and keep a minimal shim shell.
`Symbol::partition` is deleted; every dynsym add goes to
`ctx.mainPart`. `ctx.partitions` wraps a `std::vector<Partition>`
whose iterator yields only the main partition; shim partitions live
in `storage[1..]` and are reached via `shims()`. The two paths that
actually emit shim binaries (`createPhdrs` and `pageAlign` in
Writer.cpp) name `shims()` directly.

Shim partitions carry no per-partition synthetic content; the
extracted shim is just an ELF header plus PT_PHDR/PT_LOAD.
`llvm-objcopy --extract-partition` keeps working but
`--dump-section .note.gnu.build-id` on a shim no longer finds
anything. Features that previously used `dlopen()` to lazy-load a
feature partition now live in main and are always loaded.

Aided by Claude Opus 4.7


pcc:
> Chrome hasn't needed this feature for many years.

Chrome's build uses it, but we should confirm with @agrieve whether it's still useful.

agrieve:
Chrome is still using them, but they never really caught on after their original use. I think we're fine to remove support for them in chrome. I've filed https://crbug.com/401249151 to track removing it (and will do soon).

lld/ELF/Config.h
```diff
@@ -175,6 +175,23 @@ struct VersionDefinition {
...
+class Partitions {
+  // [0] main; [1..] shims
+  std::vector<Partition> storage;
+
+public:
+  // Definitions in SyntheticSections.h (Partition must be complete).
+  void reset(Ctx &);
+  Partition &main();
+  Partition &addShim(Ctx &);
+  Partition &getByNumber(unsigned partno);
```
smithp35: Could use an overloaded operator[] to shorten the callsite. Or just use .get(partno)

lld/docs/ReleaseNotes.rst
```diff
@@ -43,6 +43,11 @@ ELF Improvements
 Breaking changes
 ----------------
 
+* The symbol partition feature is deprecated. Every input section and every
```
smithp35: As I understand it, this is a bit stronger than deprecated. To me deprectated means that the same functionality remains, but the feature will remain to give people time to migrate. This seems like a step further along the process as what's left has reduced functionality.

I'm struggling to find the right word, perhaps "The symbol partition feature is now a shim." Other candidates include stub and placeholder.

lld/ELF/Driver.cpp
smithp35: This URL won't exist any more after this patch, can this be updated to mention the current status/implementation.

lld/ELF/Writer.cpp
```diff
@@ -1575,7 +1550,8 @@ template <class ELFT> void Writer<ELFT>::finalizeAddressDependentContent() {
     if (ctx.in.mipsGot)
       ctx.in.mipsGot->updateAllocSize(ctx);
 
-    for (Partition &part : ctx.partitions) {
+    {
```
smithp35: Does this need to be here? At a glance I can't spot anything scope related in here.

lld/ELF/Writer.cpp
```diff
@@ -883,7 +858,8 @@ template <class ELFT> void Writer<ELFT>::setReservedSymbolSections() {
   auto isLarge = [&ctx = ctx](OutputSection *osec) {
     return ctx.arg.emachine == EM_X86_64 && osec->flags & SHF_X86_64_LARGE;
   };
-  for (Partition &part : ctx.partitions) {
+  {
```
smithp35: Is this scope needed?

lld/ELF/Writer.cpp
```diff
@@ -2049,7 +2012,8 @@ template <class ELFT> void Writer<ELFT>::finalizeSections() {
   // The headers have to be created before finalize as that can influence the
   // image base and the dynamic section on mips includes the image base.
   if (!ctx.arg.relocatable && !ctx.arg.oFormatBinary) {
-    for (Partition &part : ctx.partitions) {
+    {
```
smithp35: Is this scope needed?

lld/ELF/Writer.cpp
```diff
@@ -2103,11 +2071,11 @@ template <class ELFT> void Writer<ELFT>::finalizeSections() {
     finalizeSynthetic(ctx, ctx.in.plt.get());
     finalizeSynthetic(ctx, ctx.in.iplt.get());
     finalizeSynthetic(ctx, ctx.in.ppc32Got2.get());
-    finalizeSynthetic(ctx, ctx.in.partIndex.get());
 
     // Dynamic section must be the last one in this list and dynamic
     // symbol table section (dynSymTab) must be the first one.
-    for (Partition &part : ctx.partitions) {
+    {
```
smithp35: Is this scope needed?

lld/ELF/Writer.cpp
```diff
@@ -2438,9 +2406,10 @@ Writer<ELFT>::createPhdrs(Partition &part) {
   if (tlsHdr->firstSec)
     ret.push_back(std::move(tlsHdr));
 
-  // Add an entry for .dynamic.
-  if (OutputSection *sec = part.dynamic->getParent())
-    addHdr(PT_DYNAMIC, sec->getPhdrFlags())->add(sec);
+  // Add an entry for .dynamic. Shim partitions don't carry a .dynamic.
+  if (part.dynamic)
```
smithp35, line 2400: I couldn't spot anywhere where this changed to being optionally created.

lld/ELF/Writer.cpp
```diff
@@ -2961,7 +2932,8 @@ static void fillTrap(std::array<uint8_t, 4> trapInstr, uint8_t *i,
 // boundary. Even though it is not required by any standard, it is in general
 // a good thing to do for security reasons.
 template <class ELFT> void Writer<ELFT>::writeTrapInstr() {
-  for (Partition &part : ctx.partitions) {
+  {
```
smithp35: Is this scope necessary?

lld/ELF/Relocations.cpp
```diff
@@ -1197,7 +1197,8 @@ template <class ELFT> void elf::scanRelocations(Ctx &ctx) {
     }
     auto scanEH = [&] {
       RelocScan scanner(ctx);
-      for (Partition &part : ctx.partitions) {
+      {
```
smithp35: Is this scope needed?

smithp35 (commented):
I think this is stronger than a deprecation, could be worth choosing a different word. No objections to the patch, especially if it isn't actively being used.

Apologies for the repetititve comments. Looks like a translation of for loops, keeps the braces unecessarily.

lld/ELF/Relocations.cpp
MaskRay: `for (Partition &part : ctx.partitions) {` is mechanically replaced with `{ part = ... ` to minimize indentation changes. When the body has only a few lines I agree that I should just remove the scope.

MaskRay:
> Chrome is still using them, but they never really caught on after their original use. I think we're fine to remove support for them in chrome. I've filed [crbug.com/401249151](https://crbug.com/401249151) to track removing it (and will do soon).

Thanks!

> I think this is stronger than a deprecation, could be worth choosing a different word. No objections to the patch, especially if it isn't actively being used.
>
> Apologies for the repetititve comments. Looks like a translation of for loops, keeps the braces unecessarily.

Thanks for the review! Changed the title to "[ELF] Reduce symbol partitions to shim binaries". Thanks a lot for the fast turnaround from Chrome we can remove this very soon.

smithp35 (approved):
Thanks for the simplifications. From the perspective of the code and docs, this looks good to me.

As it looks like it is OK with the known user, I've marked as approved.

Will be worth leaving some time for those in the US time zone to add any further comments.

agrieve:
Pending any reverts, chrome no longer uses this. https://crbug.com/401249151

PR #198720, opened 2026-05-20
[CMake] Create and link against object libraries for each component

petrhosek:
This is a rebased copy of https://reviews.llvm.org/D95727. We would like to enable building libLLVM as a static archive which was originally proposed in https://reviews.llvm.org/D99155 via archive merging, but using object libraries is cleaner and simpler approach.

This change also additionally includes a new tool, llvm-archive, for building LLVM as a single static archive, as an alternative approach to https://reviews.llvm.org/D99155. This demonstrates another use case object libraries.

### DIFF
--- a/llvm/cmake/modules/AddLLVM.cmake
+++ b/llvm/cmake/modules/AddLLVM.cmake
@@ -682,6 +682,7 @@ function(llvm_add_library name)
           add_dependencies(${obj_name} ${link_lib})
         endif()
       endforeach()
+      target_link_libraries(${obj_name} ${ARG_LINK_LIBS})
     endif()
 
     if(ARG_DISABLE_LLVM_LINK_LLVM_DYLIB)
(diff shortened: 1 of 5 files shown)


llvm/cmake/modules/AddLLVM.cmake
```diff
@@ -994,7 +995,7 @@ function(add_llvm_component_library name)
     "COMPONENT_NAME;ADD_TO_COMPONENT"
     ""
     ${ARGN})
-  add_llvm_library(${name} COMPONENT_LIB ${ARG_UNPARSED_ARGUMENTS})
+  add_llvm_library(${name} COMPONENT_LIB OBJECT ${ARG_UNPARSED_ARGUMENTS})
```
tstellar, line 998: With clang, we build both static and object libraries, but I like your approach better because it's easier to disable the static library build.  Are there any other advantages to doing it this way?

PR #198723, opened 2026-05-20
[RISCV][P-ext] Custom-lower SELECT for v4i16/v8i8 on RV32

sihuan:
SELECT was Expand for RV32 64-bit packed types, producing 40-80 lines of stack-based per-element scalarization.  Make it Custom for v4i16/v8i8 and extend the existing isPExtPackedType branch in lowerSELECT to bitcast to an integer of matching width: single-GPR types select on XLenVT directly, while RV32 double-wide types select on i64 which type-legalizes to two scalar selects on the i32 halves.

v2i32 is left to natural type-legalization since it splits cleanly into two scalar i32 selects without a Custom hook.


topperc (approved):
LGTM

PR #198724, opened 2026-05-20
[AArch64][GlobalISel] Add tablegen pattern for uaddo combine (#198514)

c-rhodes:
Related to #197693 which filters the worklist to only include opcodes for which there are combines. It's mostly handled by canMatchOpcode which is tablgen'ed but some old combines like this one are missing a tablegen pattern and require extra handling. This adds a simple wrapper so it gets picked up by canMatchOpcode and we can delete the C++ handling.

Assisted-by: codex

### DIFF
--- a/llvm/lib/Target/AArch64/AArch64Combine.td
+++ b/llvm/lib/Target/AArch64/AArch64Combine.td
@@ -63,7 +63,7 @@ def push_mul_through_sext : push_opcode_through_ext<G_MUL, G_SEXT>;
 def simplify_uaddo_matchinfo : GIDefMatchData<"std::pair<Register, Register>">;
 def simplify_uaddo : GICombineRule<
   (defs root:$root, simplify_uaddo_matchinfo:$matchinfo),
-  (match (wip_match_opcode G_UADDO):$root,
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AArch64/AArch64Combine.td
```diff
@@ -60,6 +60,14 @@ def push_sub_through_sext : push_opcode_through_ext<G_SUB, G_SEXT>;
 def push_add_through_sext : push_opcode_through_ext<G_ADD, G_SEXT>;
 def push_mul_through_sext : push_opcode_through_ext<G_MUL, G_SEXT>;
 
+def simplify_uaddo_matchinfo : GIDefMatchData<"std::pair<Register, Register>">;
+def simplify_uaddo : GICombineRule<
+  (defs root:$root, simplify_uaddo_matchinfo:$matchinfo),
+  (match (wip_match_opcode G_UADDO):$root,
```
davemgreen: Do you mind trying to change this to the new style `match (G_UADDO $dst, $src1, $src2):$root`. Otherwise we will need to change it again in the future.

davemgreen (approved):

PR #174757, opened 2026-01-07
[mlir][linalg] Preserve cast semantics during generic to matmul

meshtag:
Infer signed/unsigned cast intent from cast ops in linalg.generic bodies and
propagate it via the matmul cast attribute. This could otherwise lead to
silent overflow/underflow errors in e2e execution.

TODO: Extend this to other named ops that support cast attribute.

Fixes a functional bug in the attached issue.

Fixes #174517

PR #198725, opened 2026-05-20
[mlir][linalg] Only skip supported casts in contraction matching

### LINKED ISSUE #197178: LinalgSpecializeGenericOpsPass mis-translates linalg.generic with non-identity unary operations in payload
The `--linalg-specialize-generic-ops` pass ignores unary, side-effect free, operations in the `linalg.generic` payload.

This is incorrect.

Both the following two `linalg.generic` are specialized to identical `linalg.matmul`, despite the fact the second one negates the return value from the payload.

```
#map = affine_map<(d0, d1, d2) -> (d0, d2)>
#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
module {
  func.func @mul(%arg0: tensor<3x3xf32>, %arg1: tensor<3x3xf32>, %arg2: tensor<3x3xf32>) -> tensor<3x3xf32> {
    %0 = linalg.generic {indexing_maps = [#map, #map1, #map2], iterator_types = ["parallel", "parallel", "reduction"]} ins(%arg0, %arg1 : tensor<3x3xf32>, tensor<3x3xf32>) outs(%arg2 : tensor<3x3xf32>) {
    ^bb0(%in: f32, %in_0: f32, %out: f32):
      %1 = arith.mulf %in, %in_0 : f32
      %2 = arith.addf %out, %1 : f32
      linalg.yield %2 : f32
    } -> tensor<3x3xf32>
    return %0 : tensor<3x3xf32>
  }
  func.func @mul_with_neg(%arg0: tensor<3x3xf32>, %arg1: tensor<3x3xf32>, %arg2: tensor<3x3xf32>) -> tensor<3x3xf32> {
    %0 = linalg.generic {indexing_maps = [#map, #map1, #map2], iterator_types = ["parallel", "parallel", "reduction"]} ins(%arg0, %arg1 : tensor<3x3xf32>, tensor<3x3xf32>) outs(%arg2 : tensor<3x3xf32>) {
    ^bb0(%in: f32, %in_0: f32, %out: f32):
      %1 = arith.mulf %in, %in_0 : f32
      %2 = arith.addf %out, %1 : f32
      // NOTE: unary, side-effect free but not identity:
      %3 = arith.negf %2 : f32
      linalg.yield %3 : f32
    } -> tensor<3x3xf32>
    return %0 : tensor<3x3xf32>
  }
}
```

`mlir-opt --linalg-specialize-generic-ops` gives two identical `linalg.matmul`:

```
module {
  func.func @mul(%arg0: tensor<3x3xf32>, %arg1: tensor<3x3xf32>, %arg2: tensor<3x3xf32>) -> tensor<3x3xf32> {
    %0 = linalg.matmul ins(%arg0, %arg1 : tensor<3x3xf32>, tensor<3x3xf32>) outs(%arg2 : tensor<3x3xf32>) -> tensor<3x3xf32>
    return %0 : tensor<3x3xf32>
  }
  func.func @mul_neg(%arg0: tensor<3x3xf32>, %arg1: tensor<3x3xf32>, %arg2: tensor<3x3xf32>) -> tensor<3x3xf32> {
    %0 = linalg.matmul ins(%arg0, %arg1 : tensor<3x3xf32>, tensor<3x3xf32>) outs(%arg2 : tensor<3x3xf32>) -> tensor<3x3xf32>
    return %0 : tensor<3x3xf32>
  }
}
```

This is on what Compiler Explorer calls "Trunk" but the relevant code seems to be present at least since 2023 (`getSourceSkipUnary()`).

Chennesxu:
`getSourceSkipUnary` skipped arbitrary unary side-effect-free ops when matching contraction bodies. This allowed semantics-changing ops such as `arith.negf` to be ignored, causing `linalg.generic` ops to be incorrectly specialized to named/category contraction ops.

Example incorrectly specialized to `linalg.matmul`:

```mlir
^bb0(%in: f32, %in_0: f32, %out: f32):
  %0 = arith.mulf %in, %in_0 : f32
  %1 = arith.negf %0 : f32
  %2 = arith.addf %out, %1 : f32
  linalg.yield %2 : f32
```

This computes `out + (-(a * b))` but was treated as `out + (a * b)`.

Changes:
- Restrict the matcher to a conservative allowlist of scalar arith casts modeled by current linalg contraction cast semantics: extsi, extui, extf, trunci, truncf, sitofp, uitofp, fptosi, fptoui
- Rename `getSourceSkipUnary` to `getSourceSkipCast`
- Only skip one cast layer, so cast chains like `fptosi -> sitofp` are not treated as transparent
- Refine cast placement matching to allow `cast(c) + (cast(a) * cast(b))` while rejecting `c + cast(a * b)` and `cast(c + (a * b))`
- Update comments in `LinalgInterfaces.h`, `LinalgInterfaces.cpp`, and `LinalgInterfaces.td` to match the supported contraction form
- Add negative tests for `arith.negf`, `arith.bitcast`, and cast chains, covering both generic-to-named and generic-to-category paths
- Add positive tests for `fptosi`/`fptoui` casts

Fixes #197178.

### DIFF
--- a/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp
+++ b/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp
@@ -296,19 +296,21 @@ bool linalg::isaElemwiseSingleBinaryOpInterface(linalg::GenericOp op,
 // ContractionOpInterface implementation
 //===----------------------------------------------------------------------===//
 
-/// If the value is defined by a chain of unary side effect-free, go up the
-/// use-def chain until the first value that isn't defined by such an op.
-// TODO: relax to multi-operands with constants, which are technically unary ops
-// as needed (e.g. add5).
-static Value getSourceSkipUnary(Value value) {
+/// Return true for scalar arith cast ops modeled by current linalg contraction
+/// cast semantics.
+static bool isSupportedContractionCast(Operation *op) {
+  return isa_and_present<arith::ExtFOp, arith::TruncFOp, arith::ExtSIOp,
+                         arith::ExtUIOp, arith::TruncIOp, arith::SIToFPOp,
+                         arith::UIToFPOp, arith::FPToSIOp, arith::FPToUIOp>(
+      op);
+}
+
+/// If the value is defined by a supported contraction cast op, return its
+/// source. Otherwise, return the value unchanged.
+static Value getSourceSkipCast(Value value) {
   Operation *op = value.getDefiningOp();
-  while (op && op->getNumOperands() == 1) {
-    auto iface = dyn_cast<MemoryEffectOpInterface>(op);
-    if (!iface || !iface.hasNoEffect())
-      break;
-    value = op->getOperand(0);
-    op = value.getDefiningOp();
-  }
+  if (op && op->getNumOperands() == 1 && isSupportedContractionCast(op))
+    return op->getOperand(0);
   return value;
 }
 
@@ -331,7 +333,7 @@ bool mlir::linalg::detail::isContractionBody(
     return false;
   }
 
-  Value yielded = getSourceSkipUnary(terminator->getOperand(0));
+  Value yielded = terminator->getOperand(0);
   Operation *reductionOp = yielded.getDefiningOp();
   if (!reductionOp || reductionOp->getNumResults() != 1 ||
       reductionOp->getNumOperands() != 2) {
@@ -339,18 +341,20 @@ bool mlir::linalg::detail::isContractionBody(
     return false;
   }
 
-  Value reductionLHS = getSourceSkipUnary(reductionOp->getOperand(0));
-  Value reductionRHS = getSourceSkipUnary(reductionOp->getOperand(1));
+  Value reductionLHS = reductionOp->getOperand(0);
+  Value reductionRHS = reductionOp->getOperand(1);
+  bool lhsIsAccumulator =
+      getSourceSkipCast(reductionLHS) == block.getArgument(2);
+  bool rhsIsAccumulator =
+      getSourceSkipCast(reductionRHS) == block.getArgument(2);
 
-  if (reductionLHS != block.getArgument(2) &&
-      reductionRHS != block.getArgument(2)) {
+  if (!lhsIsAccumulator && !rhsIsAccumulator) {
     errs << "expected reduction to take block argument #2 as one of the "
-            "operands (modulo unary casts)";
+            "operands (modulo supported contraction casts)";
     return false;
   }
 
-  Value contributed = getSourceSkipUnary(
-      isa<BlockArgument>(reductionLHS) ? reductionRHS : reductionLHS);
+  Value contributed = lhsIsAccumulator ? reductionRHS : reductionLHS;
   Operation *elementwiseOp = contributed.getDefiningOp();
   if (!elementwiseOp || elementwiseOp->getNumResults() != 1 ||
       elementwiseOp->getNumOperands() != 2) {
@@ -363,8 +367,8 @@ bool mlir::linalg::detail::isContractionBody(
     return false;
   }
 
-  Value elementwiseLHS = getSourceSkipUnary(elementwiseOp->getOperand(0));
-  Value elementwiseRHS = getSourceSkipUnary(elementwiseOp->getOperand(1));
+  Value elementwiseLHS = getSourceSkipCast(elementwiseOp->getOperand(0));
+  Value elementwiseRHS = getSourceSkipCast(elementwiseOp->getOperand(1));
   if ((elementwiseLHS == block.getArgument(0) &&
        elementwiseRHS == block.getArgument(1)) ||
       (elementwiseLHS == block.getArgument(1) &&
@@ -372,8 +376,8 @@ bool mlir::linalg::detail::isContractionBody(
     return true;
   }
 
-  errs << "expected elementwise op to apply to block arguments (modulo unary "
-          "casts)";
+  errs << "expected elementwise op to apply to block arguments (modulo "
+          "supported contraction casts)";
   return false;
 }
 
@@ -594,9 +598,9 @@ bool mlir::linalg::isaContractionOpInterface(LinalgOp linalgOp) {
 ///   1. Has 2 input and 1 output shapes.
 ///   2. Has at least one reduction dimension.
 ///   3. Has only projected permutation indexing maps.
-///   4. its body computes `u5(u1(c) + u2(u3(a) * u4(b)))` on some field
-///   (AddOpType, MulOpType), where u1, u2, u3, u4 and u5 represent scalar unary
-///   operations that may change the type (e.g. for mixed-precision).
+///   4. its body computes `cast(c) + (cast(a) * cast(b))` on some field
+///   (AddOpType, MulOpType), where each `cast` is either identity or one scalar
+///   arith cast operation emitted by `linalg.generalize`.
 /// As a consequence, when vectorization of such an op occurs, the only special
 /// behavior is that the (unique) MulOpType is vectorized into a
 /// `vector.contract`. All other ops are handled in a generic fashion.
--- a/mlir/test/Dialect/Linalg/specialize-generic-ops-fail.mlir
+++ b/mlir/test/Dialect/Linalg/specialize-generic-ops-fail.mlir
@@ -1,4 +1,7 @@
-// RUN: mlir-opt %s -split-input-file --linalg-specialize-generic-ops | FileCheck %s
+// RUN: mlir-opt %s -split-input-file -linalg-morph-ops=generic-to-named \
+// RUN: | FileCheck %s --check-prefixes=CHECK,NAMED
+// RUN: mlir-opt %s -split-input-file -linalg-morph-ops=generic-to-category \
+// RUN: | FileCheck %s --check-prefixes=CHECK,CATEGORY
 
 #map = affine_map<(d0, d1, d2) -> (d1, d0)>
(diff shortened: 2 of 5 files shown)


rengolin (commented):
Honestly, I don't even think it's safe to assume casts are fine, either.

`c += (a * b) : f32`

is not the same as

`c += (itofp(fptoi(a)) * itofp(fptoi(b))) : f32`

Casts are _necessary_ for mixed-type contractions, such as:
```mlir
  (%a : i8, %b : i8, %c : i32) {
    %a32 = arith.extsi(%a)
    %b32 = arith.extsi(%b)
    %mul = arith.muli %a32, %b32
    %acc = arith.addi %c, %mul
    yield %acc
  }
```
to
`linalg.contract ins(%A : tensor<i8>, %B : tensor<i8>) outs(%C : tensor<i32>)`

but we also need to be careful with signed vs. unsigned extensions and canonical cast patterns.

I'd go as far as saying that we only recognize the patterns that the default lowering produces, and let up to canonicalization of casts to move whatever you have into those patterns, or they won't be specialized.

Chennesxu:
Good point. Would restricting to the cast ops that linalg.generalize produces (extsi/extui/extf/trunci/truncf/sitofp/uitofp) be the right scope, or do you have a more specific set in mind?

hockyy:
> Good point. Would restricting to the cast ops that linalg.generalize produces (extsi/extui/extf/trunci/truncf/sitofp/uitofp) be the right scope, or do you have a more specific set in mind?

I'm thinking of creating a better way to let future benign unary operations to match it safely.
 
1. Definitely need to add a warning or a todo in that function, new elemwise unary ops can easily be added.
2. The name `getSourceSkipUnary` can be misleading, are there any call sites where skipping arbitrary unary ops that can change value is expected?
3. I believe this affects more than just `linalg.matmul` specialization right?

hockyy:
> I'd go as far as saying that we only recognize the patterns that the default lowering produces, and let up to canonicalization of casts to move whatever you have into those patterns, or they won't be specialized.

Cool with the keeping specialization conservative, but making the global matcher only accept default-lowering shapes would likely cause broad behavior/test regressions (?)

Thinking of using `UnaryFnSkipPolicy` or something similar. use `None` or `CastOnly` or `SkipAll`. We can default to `CastOnly` for now -- to fix the problem, and expose the API for specialize pass to change this strictness... wdyt @rengolin ?

rengolin:
> Cool with the keeping specialization conservative, but making the global matcher only accept default-lowering shapes would likely cause broad behavior/test regressions (?)

A regression that was expecting broken behaviour needs to be rejected.

This is about round trips: if you can safely round trip between generic and non-generic and get the exact same representation, then the transformation is safe. If it changes the semantics when it goes back, then it's unsafe.

> Thinking of using `UnaryFnSkipPolicy` or something similar. use `None` or `CastOnly` or `SkipAll`. We can default to `CastOnly` for now -- to fix the problem, and expose the API for specialize pass to change this strictness... wdyt @rengolin ?

This creates unnecessary hurdles and allows for broken semantics where none should be. We should not default to cast only, as I just demonstrated above it's unsafe. 

You're trying to fix the wrong problem. You have to understand the semantics before asserting what the problem is.

rengolin:
> Good point. Would restricting to the cast ops that linalg.generalize produces (extsi/extui/extf/trunci/truncf/sitofp/uitofp) be the right scope, or do you have a more specific set in mind?

Yes. Anything other than an exact match would not lead to a correct round trip. We may not have the appropriate semantics in named/category ops to allow for all casts to be applied (we discussed that at length last year but decided to postpone implementation).

The only safe thing to do is to apply exactly what semantics we do have today and leave others as future implementation once we have attributes that expose the kinds of casts we want for the named ops.

hockyy:
There are some tests that fail when I tried to make it strictly reject every unary.

we can try to run canonicalize before it to see whether ir will get folded before running specialization

hockyy:
> A regression that was expecting broken behaviour needs to be rejected.

True... sorry, I worded it wrongly, I was thinking if we only modify the getSourceSkipUnary utility itself to skip every unary would cause other call site that needs the initial legacy behavior to fail.

After checking, it seems like it's only called at the specialize method

mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp
```diff
@@ -296,19 +296,20 @@ bool linalg::isaElemwiseSingleBinaryOpInterface(linalg::GenericOp op,
...
+static Value getSourceSkipCast(Value value) {
   Operation *op = value.getDefiningOp();
-  while (op && op->getNumOperands() == 1) {
-    auto iface = dyn_cast<MemoryEffectOpInterface>(op);
-    if (!iface || !iface.hasNoEffect())
-      break;
-    value = op->getOperand(0);
-    op = value.getDefiningOp();
-  }
+  if (op && op->getNumOperands() == 1 && isSupportedContractionCast(op))
```
hockyy, line 312: nit: can just use `isa_and_present<arith::ExtFOp, arith::TruncFOp, arith::ExtSIOp, arith::ExtUIOp,
             arith::TruncIOp, arith::SIToFPOp, arith::UIToFPOp, arith::FPToSIOp,
             arith::FPToUIOp>(op);`

mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp
hockyy:
```suggestion
///   4. its body computes `cast(c) + (cast(a) * cast(b))` on some field
///   (AddOpType, MulOpType), where each `cast` is either identity or one scalar
///   cast operation emitted by `linalg.generalize`.
```

mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp
```diff
@@ -331,25 +332,25 @@ bool mlir::linalg::detail::isContractionBody(
...
   if (reductionLHS != block.getArgument(2) &&
       reductionRHS != block.getArgument(2)) {
     errs << "expected reduction to take block argument #2 as one of the "
-            "operands (modulo unary casts)";
+            "operands (modulo supported contraction casts)";
     return false;
   }
 
-  Value contributed = getSourceSkipUnary(
+  Value contributed = getSourceSkipCast(
```
hockyy: do we need to skip the cast here as well on the reduction? Some generalization patterns didn't produce cast on this

mlir/test/Dialect/Linalg/specialize-generic-ops-fail.mlir
```diff
@@ -46,3 +46,48 @@ func.func @not_copy(%input: tensor<8xi32>, %init: tensor<8xi32>) -> tensor<8xi32
...
+    %1 = arith.fptosi %in : f32 to i32
+    %2 = arith.sitofp %1 : i32 to f32
+    %3 = arith.fptosi %in_0 : f32 to i32
+    %4 = arith.sitofp %3 : i32 to f32
+    %5 = arith.mulf %2, %4 : f32
+    %6 = arith.addf %out, %5 : f32
+    linalg.yield %6 : f32
+  } -> tensor<3x3xf32>
+  return %0 : tensor<3x3xf32>
+}
```
hockyy, line 98: can you add a failing testcase for linalg contract as well?

hockyy (commented):
can we also update the documentation for `LinalgContractionOpInterface` in `mlir/include/mlir/Dialect/Linalg/IR/LinalgInterfaces.td`

mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp
Chennesxu: Good point. I verified with --linalg-generalize-named-ops that linalg.generalize does not wrap the elementwise result in a cast before passing it to the reduction op. I will remove getSourceSkipCast here.

hockyy: btw dont forget to update the MR title and description, please also add acknowledgment if you are using any AI assistant in the MR, you can check the example from other MRs

something like

Assisted by: Cursor - Deepseek 4

for example

Chennesxu: sure

hockyy:
this MR also trying to fix a bug mentioned here #174757 fyi @banach-space @adam-smnk

Chennesxu:
@adam-smnk @banach-space @meshtag could you take a look?

Chennesxu:
> > Good point. Would restricting to the cast ops that linalg.generalize produces (extsi/extui/extf/trunci/truncf/sitofp/uitofp) be the right scope, or do you have a more specific set in mind?
> 
> Yes. Anything other than an exact match would not lead to a correct round trip. We may not have the appropriate semantics in named/category ops to allow for all casts to be applied (we discussed that at length last year but decided to postpone implementation).
> 
> The only safe thing to do is to apply exactly what semantics we do have today and leave others as future implementation once we have attributes that expose the kinds of casts we want for the named ops.

Thanks @rengolin — the final version implements exactly what we agreed on: the matcher now only accepts the cast ops that linalg’s `cast_signed` / `cast_unsigned` type functions produce via `convertScalarToDtype`:

`extf` / `truncf` / `extsi` / `extui` / `trunci` / `sitofp` / `uitofp` / `fptosi` / `fptoui`

Nothing else is accepted anymore. Arbitrary unary ops — including bitcasts — are no longer skipped, so generic↔named round trips stay semantics-preserving.

Could you take another look when you get a chance?

PR #198727, opened 2026-05-20
[CIR] Attach noreturn attribute to __cxa_bad_cast

E00N777:
## Summary

  - Mark `__cxa_bad_cast` as `noreturn` in CIR, mirroring the existing
    `__cxa_bad_typeid` handling. The attribute is now set on every `CallOp` that targets it,
    covering both the CodeGen direct path (`emitCallToBadCast`) and the
    target-lowering path (`buildBadCastCall`).
  - Drop the now-fulfilled `MissingFeatures::opFuncNoReturn` entry and
    the corresponding TODO/assert at the lone caller in
    `LowerItaniumCXXABI.cpp`.
  - Update FileCheck expectations in `dynamic-cast.cpp`,
    `dynamic-cast-exact.cpp`, and `abi-lower-after-unreachable.cpp` to
    require the `{noreturn}` attribute on the lowered
    `cir.call @__cxa_bad_cast()`.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenItaniumCXXABI.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenItaniumCXXABI.cpp
@@ -2083,7 +2083,11 @@ static void emitCallToBadCast(CIRGenFunction &cgf, mlir::Location loc) {
   // TODO(cir): set the calling convention to the runtime function.
   assert(!cir::MissingFeatures::opFuncCallingConv());
 
-  cgf.emitRuntimeCall(loc, getBadCastFn(cgf));
+  mlir::NamedAttrList attrs;
+  attrs.set(cir::CIRDialect::getNoReturnAttrName(),
+            mlir::UnitAttr::get(&cgf.cgm.getMLIRContext()));
+
+  cgf.emitRuntimeCall(loc, getBadCastFn(cgf), {}, attrs);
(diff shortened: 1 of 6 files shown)


clang/lib/CIR/CodeGen/CIRGenItaniumCXXABI.cpp
```diff
@@ -2074,16 +2074,24 @@ static cir::FuncOp getBadCastFn(CIRGenFunction &cgf) {
...
   assert(!cir::MissingFeatures::opFuncCallingConv());
 
+  mlir::NamedAttrList attrs;
+  attrs.set(cir::CIRDialect::getNoReturnAttrName(),
+            mlir::UnitAttr::get(&cgf.cgm.getMLIRContext()));
+
   cir::FuncType fnTy =
       cgf.getBuilder().getFuncType({}, cgf.getBuilder().getVoidTy());
-  return cgf.cgm.createRuntimeFunction(fnTy, "__cxa_bad_cast");
+  return cgf.cgm.createRuntimeFunction(fnTy, "__cxa_bad_cast", attrs);
```
andykaylor: This isn't necessary. Setting the attribute on the call site is sufficient and matches the classic codegen behavior.

andykaylor (approved):
lgtm

PR #198728, opened 2026-05-20
[libc] Add regex AST and ExprPool

kaladron:
Implemented the core AST nodes and the ExprPool arena-based allocator. Utilised AllocChecker for memory safety and enforced hardening at node initialisation.


kaladron:
Hi Alexey!  This is part 1 of a series of 4 PRs that implements the core of the derivative-based regex engine.  What's currently committed is just the public interfaces and a simple substring match.  With it we're passing 15.1% of the spenser regex tests.  After these 4 PRs land, we retain that 15.1% but have the infrastructure to keep building upon the Brzozowski derivative engine and basic ERE parser.

The unsplit PR can be found here if you want to see this change in the full context: https://github.com/llvm/llvm-project/compare/main...kaladron:llvm-project:regex-cl2

kaladron:
Also a quick note: I have overcommented the code on purpose: Partly because regex engines are confusing at the best of times, and partly because with this being potentially split into dozens of PRs, I want people to not have to keep so much state in their head and be able to quickly read English.  I'd prefer to keep this, even though it's beyond what LLVM usually likes, and when we're done everything we can prune the comments back.

libc/src/__support/regex/regex_expr_pool.cpp
```diff
@@ -0,0 +1,151 @@
...
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// Pool for Regular Expression AST nodes.
+///
+//===----------------------------------------------------------------------===//
+
+#include "src/__support/regex/regex_expr_pool.h"
+#include "include/llvm-libc-macros/regex-macros.h"
```
michaelrj-google: should this be going through a proxy header?

libc/src/__support/regex/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+  regex_expr_pool
+  SRCS
+    regex_expr_pool.cpp
+  HDRS
+    regex_expr_pool.h
+    regex_ast.h
+  DEPENDS
+    libc.src.__support.CPP.new
+    libc.src.__support.hash
+    libc.src.__support.macros.config
```
michaelrj-google, line 11: this list of depends is complete for `regex_ast.h` but doesn't list the depends for `regex_expr_pool.cpp`

libc/src/__support/regex/regex_expr_pool.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include "src/__support/macros/config.h"
+#include "src/__support/macros/null_check.h"
+
+namespace LIBC_NAMESPACE_DECL {
+
+namespace {
+
+// Hash an Expr node for hash-consing.
+uint64_t hash_expr(const Expr &e) {
+  internal::HashState hasher(0x12345678);
```
michaelrj-google, line 34: nit: explain this constant.

libc/src/__support/regex/regex_expr_pool.h
```diff
@@ -0,0 +1,109 @@
...
+  /// The first block in the allocation chain.
+  Block *head = nullptr;
+  /// The block currently being used for new node allocations.
+  Block *current = nullptr;
+  /// Total number of nodes allocated across all blocks.
+  size_t node_count = 0;
+
+  /// Size of the hash table used for hash-consing.
+  /// Must be significantly larger than MAX_NODES to maintain efficiency.
+  static constexpr size_t HASH_SIZE = 16384;
```
michaelrj-google: nit: writing this constant in hex might make it clearer that it's 2^14

libc/src/__support/regex/regex_expr_pool.h
```diff
@@ -0,0 +1,109 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// Pool for Regular Expression AST nodes.
```
michaelrj-google: nit: distinguish file comments between prototypes + class definition vs function implementations.

michaelrj-google (commented):
I'd recommend adding tests in stages along with the code. That way you can test that each piece you're adding actually works.

libc/src/__support/regex/regex_expr_pool.h
kaladron: I added an explanation as well here.

libc/src/__support/regex/regex_expr_pool.cpp
```diff
@@ -0,0 +1,162 @@
...
+  }
+}
+
+cpp::expected<Expr *, int> ExprPool::intern(const Expr &e) {
+  if (!hashtable)
+    return cpp::unexpected(REG_ESPACE);
+
+  // 1. Calculate the initial bucket for the given structural definition.
+  uint64_t h = hash_expr(e);
+  size_t idx = h & (HASH_TABLE_SIZE - 1);
```
michaelrj-google: this seems like it should be `%` and not `&` for clarity. The compiler can perform the strength reduction if `HASH_TABLE_SIZE` is a power of 2. Here and below.

libc/src/__support/regex/regex_expr_pool.cpp
```diff
@@ -0,0 +1,162 @@
...
+      return cpp::unexpected(REG_ESPACE);
+    if (!head)
+      head = new_block;
+    if (current)
+      current->next = new_block;
+    current = new_block;
+  }
+
+  // 5. Node Initialisation: Copy the structural definition into the arena.
+  Expr *new_node = &current->nodes[current->used++];
```
michaelrj-google: nit: Personally I prefer to avoid using `++` within other expressions. I find it easier to understand the code that way.
```suggestion
  Expr *new_node = &current->nodes[current->used];
  ++current->used;
```

libc/test/src/__support/regex/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+add_libc_test(
+  expr_test
+  SUITE
+    libc-support-tests
+  SRCS
+    expr_test.cpp
+  DEPENDS
+    libc.src.__support.regex.regex_expr_pool
+    libc.hdr.regex_macros
+  UNIT_TEST_ONLY
```
michaelrj-google, line 12: why is this not working for the hermetic tests?

libc/test/src/__support/regex/expr_test.cpp
```diff
@@ -0,0 +1,84 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
```
michaelrj-google, line 7: nit: add description

libc/test/src/__support/regex/expr_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include "hdr/regex_macros.h"
+#include "src/__support/regex/regex_ast.h"
+#include "src/__support/regex/regex_expr_pool.h"
+#include "test/UnitTest/Test.h"
+
+using LIBC_NAMESPACE::Expr;
+using LIBC_NAMESPACE::ExprKind;
+using LIBC_NAMESPACE::ExprPool;
+
+TEST(LlvmLibcRegexExprTest, Interning) {
```
michaelrj-google, line 22: might be good to add a test that ensures the recursive properties of hash-consing work properly.

michaelrj-google (commented):
Overall LGTM, mostly minor comments

libc/test/src/__support/regex/expr_test.cpp
kaladron: Done below!

libc/test/src/__support/regex/CMakeLists.txt
kaladron: Added a comment to CMakeLists.txt explaining that expr_test is restricted
to unit tests because the ExprPool's 128KB hash table exceeds the 64KB limit
of the simple bump allocator used in hermetic tests.

libc/src/__support/regex/regex_ast.h
```diff
@@ -0,0 +1,82 @@
...
+/// AST nodes for Regular Expressions (Class Definitions).
+///
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_LIBC_SRC___SUPPORT_REGEX_REGEX_AST_H
+#define LLVM_LIBC_SRC___SUPPORT_REGEX_REGEX_AST_H
+
+#include "src/__support/macros/config.h"
+
+namespace LIBC_NAMESPACE_DECL {
```
vonosmas, line 19: Do you think nested namespace `regex` would be helpful? Otherwise classes like `Expr` may easily conflict w/ something else.

libc/src/__support/regex/regex_ast.h
```diff
@@ -0,0 +1,82 @@
...
+  };
+
+  /// Default constructor creates an EmptySet node.
+  constexpr Expr() : kind(ExprKind::EmptySet), ch('\0') {}
+  /// Create a node of a specific kind with no data.
+  constexpr Expr(ExprKind k) : kind(k), ch('\0') {}
+  /// Create a Literal node.
+  constexpr Expr(char c) : kind(ExprKind::Literal), ch(c) {}
+  /// Create a binary node (Concat or Alt).
+  constexpr Expr(ExprKind k, Expr *l, Expr *r) : kind(k), bin{l, r} {}
```
vonosmas, line 64: WDYT of having a private constructor(s) and public factory methods instead? Then you can enforce invariants in the API, and not bother about clients creating an `Expr` node with two subnodes and type `EmptySet` etc.

```
static constexpr Expr CreateEmptySet() { return Expr(ExprKind::EmptySet, '\0'); }
static constexpr Expr CreateLiteral(char c) { return Expr(ExprKind::Literal, c); }
static constexpr Expr CreateConcat(Expr *l, Expr * r) { return Expr(ExprKind::Concat, l, r); }
```

etc.

libc/src/__support/regex/regex_expr_pool.h
```diff
@@ -0,0 +1,118 @@
...
+  /// Guarantees that for any two identical structural definitions of an Expr,
+  /// this function will return the same pointer. This enables O(1) structural
+  /// equality via pointer comparison.
+  ///
+  /// \param e A structural definition (proto-node) to intern.
+  /// \returns A pointer to the unique, stable instance in the arena,
+  ///          or REG_ESPACE on failure.
+  cpp::expected<Expr *, int> intern(const Expr &e);
+
+  /// The maximum number of nodes allowed in the pool to prevent memory
```
vonosmas: Nit: moved next to `HASH_TABLE_SIZE` above and deduplicate comments.

libc/src/__support/regex/regex_expr_pool.h
```diff
@@ -0,0 +1,118 @@
...
+    static constexpr size_t BLOCK_SIZE = 256;
+    /// The actual storage for Expr nodes.
+    Expr nodes[BLOCK_SIZE];
+    /// Pointer to the next block in the chain.
+    Block *next = nullptr;
+    /// Number of nodes currently used in this block.
+    size_t used = 0;
+
+    /// Initialises an empty block.
+    Block();
```
vonosmas: I don't think you need a constructor at all here.

libc/src/__support/regex/regex_expr_pool.cpp
```diff
@@ -0,0 +1,163 @@
...
+    break;
+  }
+  return hasher.finish();
+}
+
+} // namespace
+
+ExprPool::Block::Block() : next(nullptr), used(0) {}
+
+ExprPool::ExprPool() : head(nullptr), current(nullptr), node_count(0) {
```
vonosmas: I think head, current, node_count are already initialized by values specified in class declaration.

libc/src/__support/regex/regex_expr_pool.cpp
```diff
@@ -0,0 +1,163 @@
...
+  AllocChecker ac;
+  hashtable = new (ac) Expr *[HASH_TABLE_SIZE];
+  if (ac) {
+    for (size_t i = 0; i < HASH_TABLE_SIZE; ++i)
```
vonosmas: Can you use `inline_memset` here? I think we prevent compiler from generating a real memset call here due to `-fno-builtin` we tend to add to compiler flags.

libc/src/__support/regex/regex_expr_pool.cpp
```diff
@@ -0,0 +1,163 @@
...
+      head = new_block;
+    if (current)
+      current->next = new_block;
+    current = new_block;
+  }
+
+  // 5. Node Initialisation: Copy the structural definition into the arena.
+  Expr *new_node = &current->nodes[current->used];
+  ++current->used;
+  LIBC_CRASH_ON_NULLPTR(new_node);
```
vonosmas: This check is confusing / extraneous? If current was `nullptr` and you failed to allocate it, you'll fail with `REG_ESPACE` above. But, more importantly, you already write to `current` block in the previous line, so even if null-deref was possible, it would've happened before. I don't see why you can't have:

vonosmas (approved):
LGTM! I have a few nits/suggestions, all small enough not necessitating another round from me.

libc/src/__support/regex/regex_ast.h
kaladron: Yes, that's cleaner, thank you!

libc/src/__support/regex/regex_ast.h
kaladron: More than that, it means that invalid nodes shouldn't be possible to create.  Thanks.

libc/src/__support/regex/regex_expr_pool.h
kaladron: This was a leak from a future change in the stack.  I'll introduce it at the right time   Thanks!

PR #198729, opened 2026-05-20
 [AMDGPU] Only fold mul24+add into V_MAD_U64_U32 when operands fit in 24 bits

aobolensk:
V_MAD_U64_U32 does a full 32x32 multiply, but AMDGPUmul_u24 only uses the low 24 bits. simplifyMul24 strips outer AND masks, so the fold can silently include nonzero bits 24-31 that should not actually have any impact on the result

### DIFF
--- a/llvm/lib/Target/AMDGPU/VOP3Instructions.td
+++ b/llvm/lib/Target/AMDGPU/VOP3Instructions.td
@@ -1070,10 +1070,27 @@ multiclass IMAD32_Pats <VOP3_Pseudo inst> {
         >;
 }
 
+// AMDGPUmul_u24 only consumes the low 24 bits of its operands, and
+// simplifyMul24 may strip outer AND masks. V_MAD_U64_U32 does a full 32x32
+// multiply, so this fold is only valid when both operands are known to fit
+// in 24 unsigned bits at SDAG level.
+def AMDGPUmul_u24_24bit_ops : PatFrag<
+  (ops node:$src0, node:$src1),
+  (AMDGPUmul_u24 node:$src0, node:$src1),
+  [{
+    // AMDGPUmul_u24 is a PatFrags over an intrinsic and an SDNode; skip the
+    // intrinsic ID operand when needed.
+    unsigned BaseIdx = N->getOpcode() == ISD::INTRINSIC_WO_CHAIN ? 1 : 0;
+    return AMDGPUTargetLowering::numBitsUnsigned(N->getOperand(BaseIdx),
+                                                 *CurDAG) <= 24 &&
+           AMDGPUTargetLowering::numBitsUnsigned(N->getOperand(BaseIdx + 1),
+                                                 *CurDAG) <= 24;
+  }]>;
+
 // Handle cases where amdgpu-codegenprepare-mul24 made a mul24 instead of a normal mul.
 // We need to separate this because otherwise OtherPredicates would be overriden.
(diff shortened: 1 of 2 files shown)


llvm/test/CodeGen/AMDGPU/integer-mad-patterns.ll
```diff
@@ -12345,53 +12345,33 @@ define i64 @mul_u24_add64(i32 %x, i32 %y, i64 %z) {
...
-; GFX10-GISEL-NEXT:    v_add_co_u32 v0, vcc_lo, v4, v2
-; GFX10-GISEL-NEXT:    v_add_co_ci_u32_e64 v1, null, v1, v3, vcc_lo
-; GFX10-GISEL-NEXT:    s_setpc_b64 s[30:31]
+; GFX9-LABEL: mul_u24_add64:
+; GFX9:       ; %bb.0:
+; GFX9-NEXT:    s_waitcnt vmcnt(0) expcnt(0) lgkmcnt(0)
+; GFX9-NEXT:    v_mul_hi_u32_u24_e32 v4, v0, v1
+; GFX9-NEXT:    v_mul_u32_u24_e32 v0, v0, v1
+; GFX9-NEXT:    v_add_co_u32_e32 v0, vcc, v0, v2
+; GFX9-NEXT:    v_addc_co_u32_e32 v1, vcc, v4, v3, vcc
```
jayfoad: I wonder if it would be better to generate an explicit AND with 0x00ffffff on each of the operands and then use v_mad_u64_u32.

jayfoad (approved):
LGTM

llvm/test/CodeGen/AMDGPU/integer-mad-patterns.ll
aobolensk: Thank you! That's actually a valid point. Probably, it is gonna be even better, because instructions count in that case is smaller and they seem to be more lightweight in the end. Could you please take a look?

llvm/lib/Target/AMDGPU/VOP3Instructions.td
```diff
@@ -1070,16 +1070,50 @@ multiclass IMAD32_Pats <VOP3_Pseudo inst> {
...
     (inst $src0, $src1, $src2, 0 /* clamp */)
+    > {
+  let AddedComplexity = 1;
+}
+
+// Fallback: AMDGPUmul_u24 only uses the low 24 bits of its operands, but
+// V_MAD_U64_U32 does a full 32x32 multiply. Mask the high bits explicitly so
+// the fold remains valid even when SDAG cannot prove the operands already fit
+// in 24 bits. DAGCombine folds the AND away when the high bits are known zero,
+// recovering the single-instruction form.
```
jayfoad: I don't understand this part. No DAG combine will remove the V_AND_B32 instructions that you generate here, will it? E.g. if only one of the operands is known to fit in 24 bits already.

In general, it is not a good idea to write selection patterns that generate more than one instruction, for this reason.

aobolensk: Then I just suggest to revert this approach and stick with the previous part. I see your point

jayfoad: Yes, that's fine with me. Using AND instead is a possible future improvement.

jayfoad (approved):

PR #198737, opened 2026-05-20
[Mips] Omit the frame pointer when optimizations are enabled

yingopq:
Enable frame pointer optimization to match GCC behavior

Fix #48326.


yingopq:
@arsenm Could you help review?

arsenm (approved):

yingopq:
@arsenm Thanks!

PR #198740, opened 2026-05-20
Skip startup target for SPIRV target in libc install dependencies

jinge90:
We skip startup target for SPIRV build in https://github.com/llvm/llvm-project/pull/197381 but missed removing it from "install-libc" target dependencies, this will lead to cmake configure error when trying to install libc for SPIRV target, the error message is:
ninja: error: 'libc/lib/libc-startup', needed by 'libc/lib/CMakeFiles/install-libc', missing and no known rule to make it
We need to add startup target back when it is supported by SPIRV target.


jinge90:
Hi, @jhuber6 
Could you help review this PR?
Thanks very much!

jhuber6 (commented):
Have you tried enabling the dependencies that it relies on? We really need to see if we can get this working so someone can set up a bot.

jinge90:
> Have you tried enabling the dependencies that it relies on? We really need to see if we can get this working so someone can set up a bot.

Hi, @jhuber6 
I have tried to build startup for SPIRV, there are basic 2 problems which will lead build failure:
1. It depends on atexit and its related dependencies(exit, _Exit) which rely on RPC.
2. It depends on LTO for linking which is not ready for SPIRV.

Since SPIRV support for libc is still in early stages starting with some 'simple' functions, is it acceptable to use this patch to solve current install failure for libc for SPIRV? We can re-enable startup component when RPC fully works for SPIRV.

Thanks very much.

jinge90:
> > Have you tried enabling the dependencies that it relies on? We really need to see if we can get this working so someone can set up a bot.
> 
> Hi, @jhuber6 I have tried to build startup for SPIRV, there are basic 2 problems which will lead build failure:
> 
> 1. It depends on atexit and its related dependencies(exit, _Exit) which rely on RPC.
> 2. It depends on LTO for linking which is not ready for SPIRV.
> 
> Since SPIRV support for libc is still in early stages starting with some 'simple' functions, is it acceptable to use this patch to solve current install failure for libc for SPIRV? We can re-enable startup component when RPC fully works for SPIRV.
> 
> Thanks very much.

Hi, @jhuber6 
Kind ping~~~
Thanks very much.

PR #198741, opened 2026-05-20
[lldb][NFC] Sink eager breakpoint logic into ExecuteBreakpointSiteAction

felipepiovezan:
A future patch will want to control whether all breakpoints are eager or not based on whether the process is running. Sinking the logic allows for a smaller diff.

### DIFF
--- a/lldb/include/lldb/Target/Process.h
+++ b/lldb/include/lldb/Target/Process.h
@@ -2309,7 +2309,7 @@ protected:
 public:
   llvm::Error ExecuteBreakpointSiteAction(BreakpointSite &site,
                                           Process::BreakpointAction action,
-                                          bool forbid_delay = false);
+                                          bool forbid_delay);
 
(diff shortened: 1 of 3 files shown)


jimingham:
This seems reasonable.  I wonder if there are enough call sites to warrant do_it_now having a default argument.  Maybe it would be easier to read if everybody had to explicitly state whether they were forcing an update or not?

felipepiovezan:
> This seems reasonable. I wonder if there are enough call sites to warrant do_it_now having a default argument. Maybe it would be easier to read if everybody had to explicitly state whether they were forcing an update or not?

Yup, agreed. I've removed the default, and also renamed the parameter to "forbid_delay", conveying the self-documenting idea that if this is true, then we definitely don't delay. But if this is false, nothing is promised.

jimingham (approved):
LGTM

PR #198742, opened 2026-05-20
[Flang][OpenMP][NFC] Remove Fortran Evaluate Depedancy from Support

Stylie777:
Following #197442, FortranEvaluate was implicitly included in OpenMP-utils.h which should be avoided to ensure front-end data structures in the Optimizers can stop and restart pure MLIR source without any side-data structures.

To ensure this is done, EntryBlockArgs has been stripped back to only track vars, objects are now tracked within ObjectEntryBlockArgs in Lowering as this is a more appropriate place for this information, and the existing symbol tracking in EntryBlockArgsEntry was only used here. This ensures FortranEvaluate is not needed within the Optimizers, and objects can still be maintained when lowering. This enables better referencing in Reduction Clauses, where previously context was being lost for expressions such as ArrayElements.

See more: #197442

Assisted-by: Codex

### DIFF
--- a/flang/include/flang/Support/OpenMP-utils.h
+++ b/flang/include/flang/Support/OpenMP-utils.h
@@ -13,6 +13,7 @@
 #include "mlir/IR/Value.h"
 
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/STLExtras.h"
 
 namespace Fortran::common::openmp {
 /// Structure holding the information needed to create and bind entry block
--- a/flang/include/flang/Support/OpenMP-utils.h
+++ b/flang/include/flang/Support/OpenMP-utils.h
@@ -30,8 +30,8 @@ struct EntryBlockArgs {
 
   auto getVars() const {
     return llvm::concat<const mlir::Value>(hasDeviceAddrVars, hostEvalVars,
-        inReductionVars, mapVars, privVars, reductionVars,
-        taskReductionVars, useDeviceAddrVars, useDevicePtrVars);
+        inReductionVars, mapVars, privVars, reductionVars, taskReductionVars,
+        useDeviceAddrVars, useDevicePtrVars);
   }
 };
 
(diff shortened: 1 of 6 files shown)


jeanPerier (approved):
Thanks for the clean-up, this look good to me but I am not very familiar with that part, please wait for @tblah review.

flang/lib/Lower/OpenMP/OpenMP.cpp
```diff
@@ -251,6 +251,80 @@ class SectionsConstructStackFrame
   const parser::OpenMPSectionsConstruct &sectionsConstruct;
 };
 
+/// Structure holding the information needed to create and bind entry block
+/// arguments associated to a single clause during OpenMP lowering.
+struct ObjectEntryBlockArgsEntry {
```
tblah: nit: this should be in the anonymous namespace

flang/include/flang/Support/OpenMP-utils.h
```diff
@@ -9,81 +9,29 @@
...
+  llvm::ArrayRef<mlir::Value> reductionVars;
+  llvm::ArrayRef<mlir::Value> taskReductionVars;
+  llvm::ArrayRef<mlir::Value> useDeviceAddrVars;
+  llvm::ArrayRef<mlir::Value> useDevicePtrVars;
 
   auto getVars() const {
-    return llvm::concat<const mlir::Value>(hasDeviceAddr.vars, hostEvalVars,
-        inReduction.vars, map.vars, priv.vars, reduction.vars,
-        taskReduction.vars, useDeviceAddr.vars, useDevicePtr.vars);
+    return llvm::concat<const mlir::Value>(hasDeviceAddrVars, hostEvalVars,
```
tblah, line 33: nit: pleae include llvm/ADT/STLExtras.h directly. This works currently because of a transitive include from ArrayRef.h.

tblah (approved):
Looks good to me with some nit picks from codex.

flang/include/flang/Support/OpenMP-utils.h
Stylie777: Done

flang/lib/Lower/OpenMP/OpenMP.cpp
Stylie777: Done

PR #198745, opened 2026-05-20
[libc] Change POSIX to posix in unistd.yaml

kaladron:
1. In `header.py`, the canonical identifiers for standards are explicitly defined in lowercase and mapped to their uppercase/pretty names for display:     

```python
    # All the canonical identifiers are in lowercase for easy maintenance.
    # This maps them to the pretty descriptions to generate in header comments.
    LIBRARY_DESCRIPTIONS = {
        "stdc": "Standard C",
        "posix": "POSIX",
        "bsd": "BSD",
        "gnu": "GNU",
        "linux": "Linux",
        "uefi": "UEFI",
        "svid": "SVID",
    }
```  

  3. Later in the same file, when generating library descriptions, it checks for the lowercase string header.py:
    # If the header itself is in POSIX, just call it that.
```python
    if "posix" in self.standards:
        return descriptions["posix"]
```

### DIFF
--- a/libc/include/unistd.yaml
+++ b/libc/include/unistd.yaml
@@ -47,7 +47,7 @@ objects:
 functions:
   - name: __llvm_libc_syscall
     standards:
-      - POSIX
+      - posix
     return_type: long
     arguments:
       - type: long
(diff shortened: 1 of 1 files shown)


kaladron:
I'll fix the rest of the codebase in a subsequent PR.

labath (approved):

PR #198746, opened 2026-05-20
[MC] Create new MCScheduleOptions cl::opt category

tomershafir:
This patch creates a new cl::opt category for MCSchedule options. It enables tools to filter MCSchedule options based on category. Specifically, llvm-mca now filters them in, and displays them under `--help-hidden`, which wasnt the case before.

### DIFF
--- a/llvm/lib/MC/MCSchedule.cpp
+++ b/llvm/lib/MC/MCSchedule.cpp
@@ -23,11 +23,13 @@
 
 using namespace llvm;
 
+cl::OptionCategory llvm::MCScheduleOptions("Machine model options");
+
 static constexpr float DefaultReservationStationScaleFactor = 1.0f;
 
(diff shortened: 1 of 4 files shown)


llvm/lib/MC/MCSchedule.cpp
```diff
@@ -23,11 +23,13 @@
 
 using namespace llvm;
 
+cl::OptionCategory llvm::MCScheduleOptions("Machine model options");
```
mshockwave: nit: "Machine scheduling model options"?

mshockwave (approved):
LGTM w/ minor comments

boomanaiden154 (approved):
LGTM with Min's comments addressed. I'm not entirely sure that we need a test for this, but I guess it can't hurt.

PR #198747, opened 2026-05-20
[lldb] Fix delayed breakpoints on running processes

felipepiovezan:
Breakpoints should never be delayed on a running process, as they can immediately affect program execution.

### DIFF
--- a/lldb/test/API/functionalities/breakpoint/breakpoint_while_running/TestBreakpointWhileRunning.py
+++ b/lldb/test/API/functionalities/breakpoint/breakpoint_while_running/TestBreakpointWhileRunning.py
@@ -23,4 +23,3 @@ class BreakpointWhileRunning(TestBase):
         lldbutil.expect_state_changes(self, listener, process, [lldb.eStateRunning])
         self.runCmd("break set --source-pattern-regexp 'break here'")
         lldbutil.expect_state_changes(self, listener, process, [lldb.eStateStopped])
-        self.runCmd("process status")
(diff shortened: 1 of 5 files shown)


jimingham:
It is a bit odd to have a `bool do_it_now` parameter that will sometimes get ignored.  I think you are right that you don't want to allow callers to have to test this, and there aren't circumstances where you wouldn't want to directly insert sites when the program is running.  So this is probably the right decision, but you at least need to document the behavior as it is unexpected.

felipepiovezan:
> It is a bit odd to have a `bool do_it_now` parameter that will sometimes get ignored. I think you are right that you don't want to allow callers to have to test this, and there aren't circumstances where you wouldn't want to directly insert sites when the program is running. So this is probably the right decision, but you at least need to document the behavior as it is unexpected.

I've partially addressed this in the first PR by using a more self-documenting name, but also added some documentation here.

lldb/test/API/functionalities/breakpoint/breakpoint_while_running/TestBreakpointWhileRunning.py
```diff
@@ -0,0 +1,26 @@
...
+        )
+
+        self.dbg.SetAsync(True)
+        listener = self.dbg.GetListener()
+        self.runCmd("break delete --force")
+        self.runCmd("continue")
+        lldbutil.expect_state_changes(self, listener, process, [lldb.eStateRunning])
+        self.runCmd("break set --source-pattern-regexp 'break here'")
+        lldbutil.expect_state_changes(self, listener, process, [lldb.eStateStopped])
+        self.runCmd("process status")
```
jimingham: What is this line for?

jimingham:
LGTM except for the extraneous "process status" command at the end of the test.

lldb/test/API/functionalities/breakpoint/breakpoint_while_running/TestBreakpointWhileRunning.py
felipepiovezan: good catch! fixed

jimingham (approved):
LGTM

PR #198749, opened 2026-05-20
[AArch64][GlobalISel] Add support for scalar sqdmulh intrinsic

JoshdRod:
SQDMULH - Signed Saturated MUltiply, with result being Lower Half of the calculated value.

Previously, the scalar version of this intrinsic was failing to lower, because the RegBankSelect phase was attempting to put the source and destination values (i32s) on general purpose register banks. As NEON intrinsics aren't supported on GPR, change RegBankInfo to place all source and destination values onto FPR.


Lukacma:
Your description made me doubt myself  for a second. :) AArch64 supports 32-bit GPRs. I think what you meant to say is that SQDMULH isn't supported with GPR ?

JoshdRod:
Yes, my mistake :)

davemgreen (approved):
Other than fixing the summary, LGTM

PR #198750, opened 2026-05-20
[Offload] fix OffloadAPI unittests discovery

pbalcer:
Commit 3383f0d repointed LIBOMPTARGET_LIBRARY_DIR to a different runtimes lib dir, but the unit lit config still derived the unittest binary path from it. Pass the unittest directory explicitly instead.


pbalcer:
@jhuber6 fyi

jhuber6 (approved):

PR #198751, opened 2026-05-20
[lldb] Update TestDelayedBreakpoint test to use the right setting

felipepiovezan:
This test should regardless of which setting is the default for delayed breakpoints.


felipepiovezan:
This is just a test change, so I will go ahead and merge it.

PR #198754, opened 2026-05-20
[GlobalISel] Add `combine_or_of_and` from SelectionDAG

osmanyasar05:
This PR adds the pattern `// fold or (and x, y), x --> x` from SelectionDAG.


luisacicolini (commented):
looks good to me!

osmanyasar05:
hello @arsenm, @davemgreen I would appreciate your review on this PR, thank you!

davemgreen (commented):
Hi - The obvious question is can we add commutativity to the patterns so that we do not need to duplicate it manually 4 times. The sooner we do so the less we have to rewrite when it is added.

osmanyasar05:
> Hi - The obvious question is can we add commutativity to the patterns so that we do not need to duplicate it manually 4 times. The sooner we do so the less we have to rewrite when it is added.

hey @davemgreen, unfortunately GISEL does not support this now with the MIR patterns, it could be one of the next things I could work on. do you think this PR should wait until then, or is it good to go? thanks for the feedback!

davemgreen (approved):
> hey @davemgreen, unfortunately GISEL does not support this now with the MIR patterns, it could be one of the next things I could work on. do you think this PR should wait until then, or is it good to go? thanks for the feedback!

The sooner the better, to prevent us adding a lot of patterns that just need rewriting. Can you add a FIXME in the meantime?

Otherwise LGTM

llvm/test/CodeGen/AArch64/GlobalISel/combine-or-of-and.ll
```diff
@@ -0,0 +1,94 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
+; RUN: llc -mtriple=aarch64 -verify-machineinstrs %s -o - | FileCheck %s --check-prefixes=CHECK,CHECK-SD
+; RUN: llc -mtriple=aarch64 -global-isel -verify-machineinstrs %s -o - | FileCheck %s --check-prefixes=CHECK,CHECK-GI
```
arsenm:
```suggestion
; RUN: llc -global-isel=0 -mtriple=aarch64 < %s | FileCheck %s --check-prefixes=CHECK,CHECK-SD
; RUN: llc -global-isel=1 -mtriple=aarch64 < %s | FileCheck %s --check-prefixes=CHECK,CHECK-GI
```

PR #198757, opened 2026-05-20
[flang][Semantics] Warn on repeated do-variable in nested I/O implied DO

### LINKED ISSUE #198528: [flang][semantics] No diagnostic for repeated do-variable in nested io-implied-do
## Description

Flang silently accepts nested io-implied-do loops that reuse the same do-variable name, which violates Fortran 2023 section 12.6.3 paragraph 7:

> The do-variable of an io-implied-do that is in another io-implied-do shall not appear as, nor be associated with, the do-variable of the containing io-implied-do.

While this is not a constraint (and therefore 4.2(3) does not require a conforming processor to diagnose it), flang already diagnoses the analogous constraint C7128 for array constructors. Adding a similar check for io-implied-do would improve diagnostic quality and catch likely programmer errors.

## Example

```fortran
program test
  implicit none
  integer :: matrix(10,10), i, j

  do i = 1, 10
    do j = 1, 10
      matrix(i,j) = 10*(i - 1) + j
    end do
  end do

  ! OK: different do-variables
  write(*,'(10i5)') ((matrix(i,j), j=1,10), i=1,10)

  ! Violates 12.6.3 paragraph 7: j used in both inner and outer
  write(*,'(10i5)') ((matrix(i,j), j=1,10), j=1,10)

  ! Violates 12.6.3 paragraph 7: i used in both inner and outer
  write(*,'(10i5)') ((matrix(i,j), i=1,10), i=1,10)
end program test
```

## Current behavior

Flang compiles without any diagnostic.

## Expected behavior

A portability warning (enabled by `-pedantic`), analogous to the existing error for array constructors:

```
portability: I/O implied DO index 'j' is active in a surrounding I/O implied DO loop and should not have the same name
```

## Notes

- The array constructor analog (constraint C7128) is already diagnosed by flang as an error.
- Intel ifx and CCE reject this code with an error.
- nvfortran compiles with a warning.
- gfortran silently accepts it (same as flang).
- Since this is not a constraint, a warning (rather than an error) may be more appropriate.

cenewcombe:
Fixes #198528

Add a warning when an io-implied-do's do-variable appears as, or is associated with, the do-variable of a containing io-implied-do. This diagnoses violations of Fortran 2023 12.6.3p7:
>The do-variable of an io-implied-do that is in another io-implied-do shall not appear as, nor be associated with, the do-variable of the containing io-implied-do.

Since this is not a constraint, a warning is emitted rather than an error. As suggested in the associated issue, the warning is on by default and can be suppressed with `-Wno-io-implied-do-index-conflict`.

The check detects:

- Direct name reuse (same symbol in inner and outer implied DO)
- Association via EQUIVALENCE
- Association via USE-renaming
- Association via ASSOCIATE construct

A lit test (Semantics/io-implied-do01.f90) is added covering all of the above cases, plus valid cases (sibling implied-DOs, single implied-DO) that should not trigger the warning.

Example:
```
write(*,*) ((matrix(i,j), j=1,10), j=1,10)
! warning: I/O implied DO index 'j' appears in an enclosing I/O implied DO loop
!          and should not have the same name [-Wio-implied-do-index-conflict]
```

Assisted by: Claude

### DIFF
--- a/flang/lib/Semantics/check-do-forall.cpp
+++ b/flang/lib/Semantics/check-do-forall.cpp
@@ -1195,16 +1195,70 @@ static void CheckIoImpliedDoIndex(
   }
 }
 
+// F2023 12.6.3p7: The do-variable of an io-implied-do that is in another
+// io-implied-do shall not appear as, nor be associated with, the do-variable
+// of the containing io-implied-do.
+void DoForallChecker::CheckActiveIoImpliedDoVar(const parser::Name &name) {
+  if (!name.symbol) {
+    return;
+  }
+  const Symbol &resolved{ResolveAssociations(*name.symbol)};
+  const EquivalenceSet *equivSet{FindEquivalenceSet(resolved)};
+  for (const Symbol *active : activeIoImpliedDoVars_) {
+    const Symbol &activeResolved{ResolveAssociations(*active)};
+    bool associated{&resolved == &activeResolved};
+    if (!associated && equivSet) {
+      for (const EquivalenceObject &obj : *equivSet) {
+        if (obj.symbol == activeResolved) {
+          associated = true;
+          break;
+        }
+      }
+    }
+    if (associated) {
+      if (name.symbol == active) {
+        context_.Warn(common::UsageWarning::IoImpliedDoIndexConflict,
+            name.source,
+            "I/O implied DO index '%s' appears in an enclosing I/O implied DO loop and should not have the same name"_warn_en_US,
+            name.source);
+      } else {
+        context_.Warn(common::UsageWarning::IoImpliedDoIndexConflict,
+            name.source,
+            "I/O implied DO index '%s' should not be associated with do-variable '%s' of an enclosing I/O implied DO loop"_warn_en_US,
+            name.source, active->name());
+      }
+      break;
+    }
+  }
+  activeIoImpliedDoVars_.insert(name.symbol);
+}
+
+void DoForallChecker::Enter(const parser::OutputImpliedDo &outputImpliedDo) {
+  CheckActiveIoImpliedDoVar(parser::UnwrapRef<parser::Name>(
+      std::get<parser::IoImpliedDoControl>(outputImpliedDo.t).Name()));
+}
+
+void DoForallChecker::Enter(const parser::InputImpliedDo &inputImpliedDo) {
+  CheckActiveIoImpliedDoVar(parser::UnwrapRef<parser::Name>(
+      std::get<parser::IoImpliedDoControl>(inputImpliedDo.t).Name()));
+}
+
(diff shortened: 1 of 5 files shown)


flang/test/Semantics/io-implied-do01.f90
```diff
@@ -0,0 +1,64 @@
...
+! Section 12.6.3 paragraph 7: The do-variable of an io-implied-do that is in
+! another io-implied-do shall not appear as, nor be associated with, the
+! do-variable of the containing io-implied-do.
+
+module m
+  integer :: mk
+end module
+
+program test
+  use m, only: mk, mk_alias => mk
```
eugeneepshteyn, line 13: Since this is covering all sorts of associations, how about nested procedures and host association?

Also, please add a test without `implicit none`, with only implicitly declared vars.

flang/lib/Semantics/check-do-forall.cpp
```diff
@@ -1191,16 +1191,70 @@ static void CheckIoImpliedDoIndex(
...
+}
+
 void DoForallChecker::Leave(const parser::OutputImpliedDo &outputImpliedDo) {
-  CheckIoImpliedDoIndex(context_,
-      parser::UnwrapRef<parser::Name>(
-          std::get<parser::IoImpliedDoControl>(outputImpliedDo.t).Name()));
+  const auto &name{parser::UnwrapRef<parser::Name>(
+      std::get<parser::IoImpliedDoControl>(outputImpliedDo.t).Name())};
+  CheckIoImpliedDoIndex(context_, name);
+  if (name.symbol) {
```
eugeneepshteyn, line 1250: AI reviewer is concerned that this will break with sufficient nesting. Could you please check?
```
program p
  integer :: cube(5,5,5), i, j, k
  write(*,*) (((cube(i,j,k), k=1,5), (cube(i,j,k), k=1,5), j=1,5), k=1,5)
end
```

flang/test/Semantics/io-implied-do01.f90
cenewcombe: Test cases have been added ✅

flang/lib/Semantics/check-do-forall.cpp
cenewcombe: This is technically true, but has little practical impact. The set cannot track the same symbol at multiple nesting levels, so we produce only one warning in this example, instead of two. We could use a map instead of a set to properly track the count of repeated symbols and produce multiple warnings. However, I opted not to add this complexity to the implementation, since a single warning should be sufficient.

I've added a test case to demonstrate this case. Please let me know if you think further changes are warranted.

cenewcombe:
@eugeneepshteyn I am circling back to this PR -- is it okay to merge?

eugeneepshteyn (approved):
LGTM, thank you!

PR #198758, opened 2026-05-20
[lld][PAC] Print full version and platform values on core mismatch

jchlanda:
`toHex()` only prints a single byte of the integer value, which can hide the actual mismatch in AArch64 PAuth ABI core info diagnostics.

### DIFF
--- a/lld/ELF/Driver.cpp
+++ b/lld/ELF/Driver.cpp
@@ -3193,18 +3193,21 @@ static void readSecurityNotes(Ctx &ctx) {
     }
 
     if (ctx.aarch64PauthAbiCoreInfo != f->aarch64PauthAbiCoreInfo)
-      Err(ctx)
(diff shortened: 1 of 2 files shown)


smithp35 (approved):
LGTM.

One possible improvement to the test would be to alter it so that one or more of platform/version used a value greater than 255, but I expect the full printing of all the zeroes is sufficient to see that it would work for larger values.

Please leave some time for other LLD reviewers to comment.

MaskRay (approved):

jchlanda:
> LGTM.
> 
> One possible improvement to the test would be to alter it so that one or more of platform/version used a value greater than 255, but I expect the full printing of all the zeroes is sufficient to see that it would work for larger values.
> 
> Please leave some time for other LLD reviewers to comment.

Sure, I've flipped the 8th bit, so we get `0x142`.

kovdan01 (approved):
Thanks!

PR #198759, opened 2026-05-20
[SCEV] Refactor forgetLoop and forgetValue to improve compile time

Enna1:
- For `forgetLoop`
  Every cached SCEV varies with Loop `L` transitively contains an `AddRec`, and every `AddRec` for the loop is recorded in `LoopUsers[L]`. `forgetMemoizedResults` already closes this set transitively through `SCEVUsers` and `ExprValueMap`.
  Therefore `forgetLoop` does not need to walk the def-use chain starting from header PHIs, it only needs to initialize `ToForget` with `LoopUsers[L]` and explicitly remove each header PHI's entries from `ValueExprMap` and `ConstantEvolutionLoopExitValue`, push its cached SCEVs into `ToForget`.
  As a side effect, cached SCEVs that reside in the loop body but do not depend on any `AddRec` for the loop (e.g. a `SCEVUnknown` for an icmp) are no longer invalidated.

- For `forgetValue`
  The def-use walk is still required, but it is now pruned when pushing to the worklist, a user `U` is skipped when `!ValueExprMap.contains(U) && !isa<WithOverflowInst>(U)`. Because if a SCEVable instruction is not cached in `ValueExprMap`, none of its users can have a cached SCEV expressions that depends on it. The exception is `WithOverflowInst`, which is always pushed into the worklist so that its `extractvalue` users can be reached.
  The worklist now carries the cached SCEV with each `Instruction`, and the `eraseValueFromMap` call is dropped
  since `forgetMemoizedResults` already evicts those entries transitively.

Also remove the now-unused helpers `PushDefUseChildren`, `PushLoopPHIs`, and `visitAndClearUsers`.

Assisted-by: AI


Enna1:
Compile-time impact: https://llvm-compile-time-tracker.com/compare.php?from=64f70354e262e288e3787566645919230c05196d&to=3f19b490c4755d9a15291f5ae1e37202560c7f1d&stat=instructions%3Au

Enna1:
ping :)

nikic (commented):
Can you please split the forgetLoop() and the forgetValue() changes into separate PRs?

These are very high risk changes. We've had to revert or adjust attempted improvements to the invalidation many times in the past, because some new weird edge case shows up.

llvm/lib/Analysis/ScalarEvolution.cpp
```diff
@@ -8835,13 +8792,37 @@ void ScalarEvolution::forgetValue(Value *V) {
...
-  Visited.insert(I);
-  visitAndClearUsers(Worklist, Visited, ToForget);
+
+  auto PushIntoWorklist = [&](Instruction *UI, bool AllowMissingSCEV = false) {
+    bool IsSCEVable = isSCEVable(UI->getType());
+    bool IsWO = isa<WithOverflowInst>(UI);
+    if ((!IsSCEVable && !IsWO) || !Visited.insert(UI).second)
+      return;
+    if (IsSCEVable) {
+      if (const SCEV *S = getExistingSCEV(UI))
```
nikic, line 8805: Specifically, I don't think that this logic is correct. SCEV will generally not create intermediate SCEVs for all expressions. Consider if you have something like this:

```
%add1 = add i32 %x, 1
%add2 = add i32 %add1, 1
%add3 = add i32 %add2, 1
```
And then query SCEV for `%add3`. This will create SCEV for `%x` and `%add3`, but not the intermediate adds. So if you now invalidate `%add1` (e.g. because you changed the constant), then there will be no existing SCEV, and it will fail to invalidate `%add3`.

Enna1:
Thanks for your review!

Split into two PRs:
- `forgetLoop`: https://github.com/llvm/llvm-project/pull/201572

    https://llvm-compile-time-tracker.com/compare.php?from=8557b5714bca181906118a86d8152598f525fcc7&to=35f4b10f84f0c2a9f487599269b810801af44e0a&stat=instructions%3Au

- `forgetValue`: https://github.com/llvm/llvm-project/pull/201573

    Addressed the correctness issue in the forgetValue PR. After the fix, the change shows no measurable improvement on the compile-time tracker — I'm not sure this PR is still worth.
https://llvm-compile-time-tracker.com/compare.php?from=8557b5714bca181906118a86d8152598f525fcc7&to=21a5664f792522ad9944f76a75cc977ecb8eafb8&stat=instructions%3Au

For context: the following flamegraph is taken from one of our internal cases where forgetValue and forgetLoop are the main hotspots:
gperftools flamegraph 1

PR #198761, opened 2026-05-20
[RISCV] NFC: Use the new "let append" TableGen feature to reduce duplication

paulhuggett:
llvm#182382 introduced a language extension to accumulate field values: “append” concatenates the new value after the current value, whilst "prepend" concatenates before the existing value. This change uses that feature to eliminate repetition in the definition of some of the compressed instructions.

For example, line 267 of RISCVIntrInfoC.td establishes a scope for “`let Predicates = [HasStdExtZca] in {`”; this scope ends on line 515. Meanwhile, line 454 wants to add the `IsRV64` predicate for a single instruction but was forced to duplicate the previous condition as well: “`let Predicates = [HasStdExtZca, IsRV64] in`”. That’s no longer necessary since the addition can now be explicit: “`let append Predicates = [IsRV64] in `”

I‘ve verified that this change has no effect on the TableGen output.

It seems quite likely that this same change could be made in some of the other RISC-V TableGen source files…

### DIFF
--- a/llvm/lib/Target/RISCV/RISCVInstrInfoC.td
+++ b/llvm/lib/Target/RISCV/RISCVInstrInfoC.td
@@ -295,7 +295,7 @@ def C_LW_INX : CLoad_ri<0b010, "c.lw", GPRF32C, uimm7_lsb00>,
   let Inst{5} = imm{6};
 }
 
-let Predicates = [HasStdExtZca, IsRV64] in
+let append Predicates = [IsRV64] in
 def C_LD : CLoad_ri<0b011, "c.ld", GPRC, uimm8_lsb000>,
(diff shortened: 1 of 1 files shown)


paulhuggett:

See https://github.com/llvm/llvm-project/issues/198776

topperc (approved):
LGTM. Please put [RISCV] in the PR title.

arichardson (commented):
This new feature is great! Much easier to keep track of the scopes.

topperc:
Do you have commit access?

paulhuggett:
I don't (my contributions tend to be a bit sporadic). If you have access, would you mind committing it for me (assuming sufficiently happy reviewers)?

PR #78251, opened 2024-01-16
[AsmParser] Implicitly declare intrinsics

nikic:
We currently require that all referenced globals have an explicit declaration or definition in the IR. For intrinsics, this requirement is redundant, because they cannot be called indirectly (including "direct" calls with mismatched function type). The function type used in the call directly determines the function type of the intrinsic declaration.

Relax this requirement, and implicitly declare any intrinsics that do not have an explicit declaration. This will remove a common annoyance when writing tests and alive2 proofs.

(I also plan to introduce a mode where declarations for all missing symbols will be automatically added, to make working with incomplete IR easier -- but that will be behind a default-disabled flag.)

PR #198765, opened 2026-05-20
[AArch64][TTI][EarlyCSE] Add support for ld1xN and st1xN intrinsics

chenshanzhi:
Handle ld1x2, ld1x3, ld1x4, st1x2, st1x3, st1x4 in:
- AArch64TTIImpl::getTgtMemIntrinsic
- AArch64TTIImpl::getOrCreateResultFromMemIntrinsic

This enables EarlyCSE to optimize these NEON load/store intrinsics.

To test the changes, a new testcase (intrinsics-1xN.ll) derived from llvm/test/Transforms/EarlyCSE/AArch64/intrinsics.ll is added.


chenshanzhi:
This PR fixes a missed optimization in EarlyCSE. Here is an example: https://godbolt.org/z/hqqvf53a8
In the following code, ld1 in function `f` is not optimized away now, whereas ld2 in function `g` is successfully optimized away.
```c++
#include <arm_neon.h>
#include <cstdint>

uint32x4_t vaddq_s32(uint32x4_t a, uint32x4_t b);

uint32x4_t f(uint32_t *ptr, uint32x4x2_t a, int n, uint32x4_t dummy) {
    uint32x4_t res = dummy;
    for (int i = 0; i < n; ++i) {
        uint32x4x2_t vec;
        vst1q_u32_x2(ptr, a);
        vec = vld1q_u32_x2(ptr); // ld1 is not optimized away now
        res = vaddq_s32(vec.val[0], vec.val[0]);
    }
    return res;
}

uint32x4_t g(uint32_t *ptr, uint32x4x2_t a, int n, uint32x4_t dummy) {
    uint32x4_t res = dummy;
    for (int i = 0; i < n; ++i) {
        uint32x4x2_t vec;
        vst2q_u32(ptr, a);
        vec = vld2q_u32(ptr); // ld2 is optimized away
        res = vaddq_s32(vec.val[0], vec.val[0]);
    }
    return res;
}
```

For an immediate `load` from the same location after `store`:

- The call to `@llvm.aarch64.neon.ld1x2.v4i32.p0` is not optimized away:
Screenshot from 2026-05-21 14-44-45

- The call to `@llvm.aarch64.neon.ld2.v4i32.p0` is successfully optimized away:
Screenshot from 2026-05-21 14-45-04

llvm/test/Transforms/EarlyCSE/AArch64/intrinsics-1xN.ll
```diff
@@ -0,0 +1,365 @@
...
+; Function Attrs: nounwind
+declare void @llvm.aarch64.neon.st1x3.v4i32.p0(<4 x i32>, <4 x i32>, <4 x i32>, ptr nocapture)
+
+; Function Attrs: nounwind readonly
+declare { <4 x i32>, <4 x i32> } @llvm.aarch64.neon.ld1x2.v4i32.p0(ptr)
+
+; Function Attrs: nounwind readonly
+declare { <4 x i32>, <4 x i32>, <4 x i32> } @llvm.aarch64.neon.ld1x3.v4i32.p0(ptr)
+
+define internal fastcc <4 x i32> @vaddq_s32(<4 x i32> %__p0, <4 x i32> %__p1) {
```
davemgreen: Remove this as it isn't needed for the test.

llvm/test/Transforms/EarlyCSE/AArch64/intrinsics-1xN.ll
```diff
@@ -0,0 +1,365 @@
...
+; CHECK-NEXT:    ret void
+;
+entry:
+  %ld = tail call { <16 x i8>, <16 x i8>, <16 x i8>, <16 x i8> } @llvm.aarch64.neon.ld1x4.v16i8.p0(ptr %p)
+  %ext = extractvalue { <16 x i8>, <16 x i8>, <16 x i8>, <16 x i8> } %ld, 0
+  tail call void @llvm.aarch64.neon.st1x4.v16i8.p0(<16 x i8> %ext, <16 x i8> %A, <16 x i8> %B, <16 x i8> zeroinitializer, ptr %p)
+  ret void
+}
+
+; Function Attrs: nounwind
```
davemgreen: Remove `; Function Attrs: nounwind`
The intrinsics declarations might not be needed nowadays

llvm/test/Transforms/EarlyCSE/AArch64/intrinsics-1xN.ll
```diff
@@ -0,0 +1,365 @@
...
+  %vld1x3.fca.1.extract = extractvalue { <4 x i32>, <4 x i32>, <4 x i32> } %vld1x3, 1
+  %call = call <4 x i32> @vaddq_s32(<4 x i32> %vld1x3.fca.0.extract, <4 x i32> %vld1x3.fca.0.extract)
+  %inc = add nsw i32 %i.0, 1
+  br label %for.cond
+
+for.end:                                          ; preds = %for.cond
+  ret <4 x i32> %res.0
+}
+
+define void @test_ld1x4_st1x4_no_cse(ptr %p, <16 x i8> %A, <16 x i8> %B) {
```
davemgreen: Can you make all the tests like this one? Succinct and to the point, without any extra blocks.

Can you add tests for ld2 + st1x2 as well, to make sure it is OK. And make sure st1x4+ld1x4 (that does transform) has a test.

davemgreen (commented):
I think this sounds OK.

llvm/test/Transforms/EarlyCSE/AArch64/intrinsics-1xN.ll
chenshanzhi: Addressed.

llvm/test/Transforms/EarlyCSE/AArch64/intrinsics-1xN.ll
chenshanzhi: Addressed. You're right. Intrinsics are implicitly declared after #78251.
This is really good advice and I'll put it into my personal development & review guide. Thanks.

llvm/test/Transforms/EarlyCSE/AArch64/intrinsics-1xN.ll
chenshanzhi: Addressed. All tests are simplified and functions follow the naming convention `test_cse_for_xxx` or `test_no_cse_for_xxx`.

chenshanzhi:
And I observed something interesting: https://godbolt.org/z/Toeea9h5f
```llvm
define void @test1_st1x2_st1x2(ptr %p, <4 x i32> %A, <4 x i32> %B) {
entry:
  call void @llvm.aarch64.neon.st1x2.v4i32.p0(<4 x i32> %A, <4 x i32> %B, ptr %p)
  call void @llvm.aarch64.neon.st1x2.v4i32.p0(<4 x i32> %A, <4 x i32> %B, ptr %p)
  ret void
}

define void @test1_st2_st2(ptr %p, <4 x i32> %A, <4 x i32> %B) {
entry:
  call void @llvm.aarch64.neon.st2.v4i32.p0(<4 x i32> %A, <4 x i32> %B, ptr %p)
  call void @llvm.aarch64.neon.st2.v4i32.p0(<4 x i32> %A, <4 x i32> %B, ptr %p)
  ret void
}

define void @test2_st1x2_st1x2(ptr %p, <4 x i32> %A, <4 x i32> %B) {
entry:
  call void @llvm.aarch64.neon.st1x2.v4i32.p0(<4 x i32> %A, <4 x i32> %A, ptr %p)
  call void @llvm.aarch64.neon.st1x2.v4i32.p0(<4 x i32> %A, <4 x i32> %B, ptr %p)
  ret void
}

define void @test2_st2_st2(ptr %p, <4 x i32> %A, <4 x i32> %B) {
entry:
  call void @llvm.aarch64.neon.st2.v4i32.p0(<4 x i32> %A, <4 x i32> %A, ptr %p)
  call void @llvm.aarch64.neon.st2.v4i32.p0(<4 x i32> %A, <4 x i32> %B, ptr %p)
  ret void
}
```

**Observations:**
1. Currently, redundant stores in `test1_st1x2_st1x2` and `test1_st2_st2` are both eliminated under O3. 
  1.1. For `test1_st1x2_st1x2`, the elimination occurs in EarlyCSE. 
  1.2. For `test1_st2_st2`, the elimination occurs in DSE instead of EarlyCSE.
2. Currently, redundant stores in `test2_st1x2_st1x2` and `test2_st2_st2` are not eliminated under O3.

**Observation 2 is an obvious missed optimization.**

**Observation 1.1 is a bug caused by the program taking a wrong path, and can also be fixed by this PR.**

Before this PR, when running EarlyCSE on function `test1_st1x2_st1x2`,  the `MemInst` constructed in the follow line would have `MemInst.isValid() == false`, `MemInst.isStore() == false`, and `MemInst.isLoad() == false`.
https://github.com/llvm/llvm-project/blob/fc082e1a6ddd77dddc79c49285356fd9dfed7731/llvm/lib/Transforms/Scalar/EarlyCSE.cpp#L1579
This cause the program to incorrectly take the following `if` branch. However, the store operation should not be handled by this branch.
https://github.com/llvm/llvm-project/blob/fc082e1a6ddd77dddc79c49285356fd9dfed7731/llvm/lib/Transforms/Scalar/EarlyCSE.cpp#L1647-L1652

**Observation 1.2 might be another missed optimization.**
The elimination does not occur in EarlyCSE because currently `ParseMemoryInst::getValueType` does not handle target-specific intrinsics.
Relevant code in `EarlyCSE::processNode`, `EarlyCSE::overridingStores`, and `ParseMemoryInst::getValueType`:
https://github.com/llvm/llvm-project/blob/fc082e1a6ddd77dddc79c49285356fd9dfed7731/llvm/lib/Transforms/Scalar/EarlyCSE.cpp#L1749-L1761
https://github.com/llvm/llvm-project/blob/fc082e1a6ddd77dddc79c49285356fd9dfed7731/llvm/lib/Transforms/Scalar/EarlyCSE.cpp#L1335-L1345
`Earlier.getValueType()` returns `nullptr` and `overridingStores` returns `false` in line 1345.

https://github.com/llvm/llvm-project/blob/fc082e1a6ddd77dddc79c49285356fd9dfed7731/llvm/lib/Transforms/Scalar/EarlyCSE.cpp#L905-L908

For the missed optimizations in Observations 1.2 and 2, I'll take a closer look next week if time permits. Any potential changes will be addressed in follow-up PRs. I think this PR is ready for review.

llvm/test/Transforms/EarlyCSE/AArch64/intrinsics-1xN.ll
```diff
@@ -0,0 +1,176 @@
...
+entry:
+  %ld = call { <4 x i32>, <4 x i32> } @llvm.aarch64.neon.ld2.v4i32.p0(ptr %p)
+  %x0 = extractvalue { <4 x i32>, <4 x i32> } %ld, 0
+  %x1 = extractvalue { <4 x i32>, <4 x i32> } %ld, 1
+  call void @llvm.aarch64.neon.st1x2.v4i32.p0(<4 x i32> %x0, <4 x i32> %x1, ptr %p)
+  ret void
+}
+
+; Check that ld1x4 is optimized away by EarlyCSE.
+define <16 x i8> @test_cse_for_st1x4_ld1x4(ptr %p, <16 x i8> %A, <16 x i8> %B) {
```
davemgreen, line 178: Add a valid case for st1x3+ld1x3 too, to make sure it is tested.

davemgreen:
Interesting. I'm not sure where DSE is picking the info up from, but I would expect redundant stores to be eliminated there (so it can happen between stores of different types / sizes etc).

davemgreen (approved):
LGTM, thanks.

llvm/test/Transforms/EarlyCSE/AArch64/intrinsics-1xN.ll
chenshanzhi: > Add a valid case for st1x3+ld1x3 too, to make sure it is tested.

Addressed.

PR #198766, opened 2026-05-20
[NFC][mlir][SPIR-V] Test ceildivui/ceildivsi/floordivsi via test-convert-to-spirv

aobolensk:
SPIR-V has no direct opcode for these ops. Verify that using `--test-convert-to-spirv` pass is sufficient to convert these ops to SPIR-V

### DIFF
--- a/mlir/lib/Conversion/ArithToSPIRV/ArithToSPIRV.cpp
+++ b/mlir/lib/Conversion/ArithToSPIRV/ArithToSPIRV.cpp
@@ -10,7 +10,6 @@
 
 #include "../SPIRVCommon/Pattern.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
-#include "mlir/Dialect/Arith/Transforms/Passes.h"
 #include "mlir/Dialect/SPIRV/IR/SPIRVAttributes.h"
 #include "mlir/Dialect/SPIRV/IR/SPIRVDialect.h"
 #include "mlir/Dialect/SPIRV/IR/SPIRVOps.h"
@@ -1619,7 +1618,6 @@ struct ConvertArithToSPIRVPass
     target->addIllegalDialect<arith::ArithDialect>();
 
     RewritePatternSet patterns(&getContext());
-    arith::populateCeilFloorDivExpandOpsPatterns(patterns);
     arith::populateArithToSPIRVPatterns(typeConverter, patterns);
 
     if (failed(applyPartialConversion(op, *target, std::move(patterns))))
--- a/mlir/lib/Conversion/ArithToSPIRV/ArithToSPIRV.cpp
+++ b/mlir/lib/Conversion/ArithToSPIRV/ArithToSPIRV.cpp
@@ -10,6 +10,7 @@
 
 #include "../SPIRVCommon/Pattern.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Arith/Transforms/Passes.h"
 #include "mlir/Dialect/SPIRV/IR/SPIRVAttributes.h"
 #include "mlir/Dialect/SPIRV/IR/SPIRVDialect.h"
 #include "mlir/Dialect/SPIRV/IR/SPIRVOps.h"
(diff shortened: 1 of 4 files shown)


kuhar (commented):
I don't remember the exact discussion, but I think the reason this was not enabled by default is that there may be more efficient lowering paths via EIS ops. Can we double check that there is no better lowering by using those? Or alternatively, is there existing arith->arith expansion for these that we could use instead?

aobolensk:
> I don't remember the exact discussion, but I think the reason this was not enabled by default is that there may be more efficient lowering paths via EIS ops. Can we double check that there is no better lowering by using those? Or alternatively, is there existing arith->arith expansion for these that we could use instead?

As far as I can see, analogs there are working with FP only (according to the specs):
 - https://registry.khronos.org/SPIR-V/specs/unified1/GLSL.std.450.html (Ceil, Floor)
   * "The operand x must be a scalar or vector whose component type is floating-point."
 - https://registry.khronos.org/SPIR-V/specs/unified1/OpenCL.ExtendedInstructionSet.100.html (ceil, floor)
   * "Result Type and x must be floating-point or vector(2,3,4,8,16) of floating-point values."

If use them, there should be additional int -> float -> int casts inserted

kuhar:
> is there existing arith->arith expansion for these that we could use instead?

What about this?

aobolensk:
> > is there existing arith->arith expansion for these that we could use instead?
> 
> What about this?

These are already arith->arith, aren't they? I have not found alternatives to these patterns https://github.com/llvm/llvm-project/blob/57712b1eb7305d232ee5975990005cc8191ab262/mlir/lib/Dialect/Arith/Transforms/ExpandOps.cpp#L906-L911 in vanilla MLIR so far

aobolensk:
> > > is there existing arith->arith expansion for these that we could use instead?
> > 
> > 
> > What about this?
> 
> These are already arith->arith, aren't they? I have not found alternatives to these patterns
> 
> https://github.com/llvm/llvm-project/blob/57712b1eb7305d232ee5975990005cc8191ab262/mlir/lib/Dialect/Arith/Transforms/ExpandOps.cpp#L906-L911
> 
> in vanilla MLIR so far

Or do you suggest to use a bigger scope: `mlir::arith::populateArithExpandOpsPatterns`?

kuhar:
Yeah I'd rather use these for ceildiv and friends and drop the bespoke spirv path if there's nothing efficient we can do anyway

aobolensk:
> Yeah I'd rather use these for ceildiv and friends and drop the bespoke spirv path if there's nothing efficient we can do anyway

Then I guess it is already done, we are using `populateCeilFloorDivExpandOpsPatterns` in this PR. `populateArithExpandOpsPatterns` will decompose FP ops, which is not what we want.

kuhar:
> > Yeah I'd rather use these for ceildiv and friends and drop the bespoke spirv path if there's nothing efficient we can do anyway
> 
> Then I guess it is already done, we are using `populateCeilFloorDivExpandOpsPatterns` in this PR. `populateArithExpandOpsPatterns` will decompose FP ops, which is not what we want.

So concretely, what I'd suggest is for us to drop the spirv specific expansion code and instead wire up the arith rewrite to the `--test-convert-to-spirv` pass to show how this is supposed to be put together in downsteam projects.

aobolensk:
> > > Yeah I'd rather use these for ceildiv and friends and drop the bespoke spirv path if there's nothing efficient we can do anyway
> > 
> > 
> > Then I guess it is already done, we are using `populateCeilFloorDivExpandOpsPatterns` in this PR. `populateArithExpandOpsPatterns` will decompose FP ops, which is not what we want.
> 
> So concretely, what I'd suggest is for us to drop the spirv specific expansion code and instead wire up the arith rewrite to the `--test-convert-to-spirv` pass to show how this is supposed to be put together in downsteam projects.

Thanks, now I understood. Reverted the main change, added the test to the corresponding location

aobolensk:
@kuhar could you please take a look?

kuhar (approved):

PR #198773, opened 2026-05-20
[InterleavedAccessPass] Pass modifies its input and doesn't report it

sjoerdmeijer:
In a build with expensive checks enabled, it is very easy to trigger this error in the InterleavedAccess pass:

    Pass modifies its input and doesn't report it: Interleaved Access Pass
    Pass modifies its input and doesn't report it
    UNREACHABLE executed at /local/home/smeijer/llvm-project/llvm/lib/IR/LegacyPassManager.cpp:1404!

This patch removes (some) IR changes when the pass bails out.


llvm/lib/CodeGen/InterleavedAccessPass.cpp
```diff
@@ -395,9 +413,10 @@ bool InterleavedAccessImpl::lowerInterleavedLoad(
   // Try to create target specific intrinsics to replace the load and
   // shuffles.
   if (!TLI->lowerInterleavedLoad(cast<Instruction>(Load), Mask, Shuffles,
-                                 Indices, Factor, GapMask))
-    // If Extracts is not empty, tryReplaceExtracts made changes earlier.
-    return !Extracts.empty() || BinOpShuffleChanged;
+                                 Indices, Factor, GapMask)) {
+    eraseDeadMaskInstructions(Mask);
+    return hasMutatedIR();
```
davemgreen: Is it worth being conservative and returning true from here as we might have changed the IR by removing dead Masks. We could have all these functions just return true as "MayChanged" if they might have changed the IR.

llvm/test/Transforms/InterleavedAccess/AArch64/masked-store-mask-cleanup.ll
```diff
@@ -0,0 +1,29 @@
...
+;
+  %splat_v_in = insertelement <4 x i32> poison, i32 %x, i32 0
+  %splat_v = shufflevector <4 x i32> %splat_v_in, <4 x i32> poison, <4 x i32> zeroinitializer
+  %splat_m_in = insertelement <4 x i1> poison, i1 %m, i32 0
+  %splat_m = shufflevector <4 x i1> %splat_m_in, <4 x i1> poison, <4 x i32> zeroinitializer
+  tail call void @llvm.masked.store.v4i32.p0(<4 x i32> %splat_v, ptr %p, i32 8, <4 x i1> %splat_m)
+  ret void
+}
+
+declare void @llvm.masked.store.v4i32.p0(<4 x i32>, ptr, i32, <4 x i1>)
```
mshockwave: we no longer need to declare intrinsics

llvm/lib/CodeGen/InterleavedAccessPass.cpp
boomanaiden154: Yeah, that would definitely be simpler. This pass preserves the CFG so I'm not sure we would be invalidating anything too expensive by just returning true.

sjoerdmeijer: Ok, thanks, sounds good to me, will fix that.

sjoerdmeijer:
Ping

llvm/lib/CodeGen/InterleavedAccessPass.cpp
```diff
@@ -551,8 +565,10 @@ bool InterleavedAccessImpl::lowerInterleavedStore(
 
   // Try to create target specific intrinsics to replace the store and
   // shuffle.
-  if (!TLI->lowerInterleavedStore(Store, Mask, SVI, Factor, GapMask))
+  if (!TLI->lowerInterleavedStore(Store, Mask, SVI, Factor, GapMask)) {
+    eraseDeadMaskInstructions(Mask);
```
mshockwave, line 569: is the comment https://github.com/llvm/llvm-project/pull/198773#discussion_r3282546563 applicable to here as well?

llvm/lib/CodeGen/InterleavedAccessPass.cpp
```diff
@@ -725,8 +741,10 @@ bool InterleavedAccessImpl::lowerDeinterleaveIntrinsic(
   }
 
   // Try and match this with target specific intrinsics.
-  if (!TLI->lowerDeinterleaveIntrinsicToLoad(LoadedVal, Mask, DI, GapMask))
+  if (!TLI->lowerDeinterleaveIntrinsicToLoad(LoadedVal, Mask, DI, GapMask)) {
+    eraseDeadMaskInstructions(Mask);
```
mshockwave, line 745: ditto

llvm/lib/CodeGen/InterleavedAccessPass.cpp
```diff
@@ -777,8 +795,10 @@ bool InterleavedAccessImpl::lowerInterleaveIntrinsic(
   }
 
   // Try and match this with target specific intrinsics.
-  if (!TLI->lowerInterleaveIntrinsicToStore(StoredBy, Mask, InterleaveValues))
+  if (!TLI->lowerInterleaveIntrinsicToStore(StoredBy, Mask, InterleaveValues)) {
+    eraseDeadMaskInstructions(Mask);
```
mshockwave, line 799: ditto

PR #198778, opened 2026-05-20
Implement linux-specific sendmmsg(2)

labath:
Assisted by Gemini.


libc/include/sys/socket.yaml
```diff
@@ -121,6 +122,15 @@ functions:
       - type: const void*
       - type: size_t
       - type: int
+  - name: sendmmsg
+    standards:
+      - Linux
```
kaladron:
```suggestion
      - linux
```

I was looking at the YAML parser this morning, and it looks like it wants all of these to be lower case.  (That's why I sent the PR to you earlier with the changes for s,POSIX,posix,)

kaladron (approved):
One minor comment within.

PR #198780, opened 2026-05-20
[flang][OpenMP] Limit scope creation to constructs with data environment

kparzysz:
Identify specific constructs that require data envorinments, and only create scopes for them. This avoids scopes for loop-transformation constructs, for example.

This isn't a correctness fix, but a clarification and a simplification of the name-resolution code for OpenMP.


kparzysz:
cc: @dreachem

kparzysz:
The directives listed are the executable ones that accept a clause with a "data-environment attribute" or a "data-motion attribute" property.  They roughly correspond to constructs that have a "team-generating", "task-generating", or a "work-distribution" properties with a few tweaks.

"Taskgraph" is an odd one, with none of these properties, however it's explicitly named in the section on data-sharing rules:
> Loop-iteration variables inside parallel, teams, taskgraph, or task-generating constructs are private in the innermost such construct that encloses the loop.

dreachem:
> The directives listed are the executable ones that accept a clause with a "data-environment attribute" or a "data-motion attribute" property. They roughly correspond to constructs that have a "team-generating", "task-generating", or a "work-distribution" properties with a few tweaks.
> 
> "Taskgraph" is an odd one, with none of these properties, however it's explicitly named in the section on data-sharing rules:
> 
> > Loop-iteration variables inside parallel, teams, taskgraph, or task-generating constructs are private in the innermost such construct that encloses the loop.

Yes, `taskgraph` stuck out to me as well. I'm still not sure it belongs. It's interesting that taskgraph is mentioned in that sentence from the spec, but not (for example) taskgroup. 

When I look at the commit in the spec in which that change was added, it was at a time when the **taskgraph** construct was also accepting the private and shared clause. I suspect that we removed those data-sharing clauses from the construct but accidentally neglected to remove **taskgraph** from that particular sentence.

flang/test/Semantics/OpenMP/affected-loops.f90
luporl: It seems this comment needs updating.

luporl (commented):
I haven't thoroughly checked all the cases in HasDataEnvironment(), but overall, the changes look good to me.

flang/test/Semantics/OpenMP/affected-loops.f90
kparzysz: Done

kparzysz:
Deepak didn't object to the list of constructs, so I'll consider this accepted.

PR #198781, opened 2026-05-20
[libc] Refactor qsort code

vhscampos:
This patch makes the following changes:
 - Refactor the internal sorting functions to reduce code duplication.
 - Move the testing machinery done for the testing of `qsort_r` to a shared place.

These changes are done in anticipation to the introduction of Annex K's `qsort_s`. This function shares most of its semantics with `qsort_r`, therefore most of the testing logic can be shared between the two. Besides, `qsort`, `qsort_r` and `qsort_s` are all very similar, hence we can attempt to reduce duplication a bit more.


libc/src/stdlib/qsort_util.h
```diff
@@ -64,10 +64,29 @@ LIBC_INLINE void unstable_sort_impl(void *array, size_t array_len,
 }
 
 template <typename F>
+LIBC_INLINE void unstable_sort_dispatch(void *array, size_t array_len,
+                                        size_t elem_size, F is_less) {
+  constexpr bool use_quick_sort = (LIBC_QSORT_IMPL == LIBC_QSORT_QUICK_SORT);
```
kaladron: constexpr bool should be uppercase, so USE_QUICK_SORT

libc/test/src/stdlib/QsortReentrantTest.h
```diff
@@ -0,0 +1,150 @@
+
```
kaladron: Please remove the stray extra line at the top, and also please use the new-style LLVM headers per https://llvm.org/docs/CodingStandards.html#file-headers

libc/test/src/stdlib/QsortReentrantTest.h
```diff
@@ -0,0 +1,150 @@
...
+class QsortReentrantTest : public LIBC_NAMESPACE::testing::Test {
+private:
+  static int int_compare_count(const void *l, const void *r, void *count_arg) {
+    int li = *static_cast<const int *>(l);
+    int ri = *static_cast<const int *>(r);
+    SizeTy *count = static_cast<SizeTy *>(count_arg);
+    *count = *count + 1;
+    if (li == ri)
+      return 0;
+    else if (li > ri)
```
kaladron:
```suggestion
    if (li > ri)
```

The else is not needed after a return (here and below)

libc/test/src/stdlib/QsortReentrantTest.h
```diff
@@ -0,0 +1,150 @@
...
+  // comparison function that takes a specific type, which is passed to a
+  // library that then needs to sort an array of that type. The library can't
+  // safely pass the comparison function to qsort because a function that takes
+  // const T* being cast to a function that takes const void* is undefined
+  // behavior. The safer pattern is to pass a type erased comparator that calls
+  // into the typed comparator to qsort_r.
+  template <typename T>
+  static int type_erased_comp(const void *l, const void *r,
+                              void *erased_func_ptr) {
+    typedef int (*TypedComp)(const T *, const T *);
```
kaladron:
```suggestion
    using TypedComp = int (*)(const T *, const T *);
```

For idiomatic C++

libc/test/src/stdlib/qsort_r_test.cpp
```diff
@@ -6,139 +6,7 @@
...
-  EXPECT_EQ(array[1].priority, 1);
-  EXPECT_EQ(array[1].size, 10);
-  EXPECT_EQ(array[2].priority, 3);
-  EXPECT_EQ(array[2].size, 3);
-  EXPECT_EQ(array[3].priority, 10);
-  EXPECT_EQ(array[3].size, 0);
-  EXPECT_EQ(array[4].priority, 10);
-  EXPECT_EQ(array[4].size, 3);
-}
+QSORTREENTRANT_TEST(QsortR, LIBC_NAMESPACE::qsort_r, size_t)
```
kaladron, line 12: Are you planning on more QSORTREENTRANT_TEST tests?  If not, I'm wondering if it's worth just expanding the macro for the single use for simplicity.  Your commit message mentions sharing this with qsort_s, so "yes" is a reasonable answer. =)

kaladron (changes requested):

libc/test/src/stdlib/qsort_r_test.cpp
vhscampos: Yes! https://github.com/llvm/llvm-project/pull/198797/changes#diff-774d57f4cc4126f17aa5ad2bd0cf664a33e45d8d9003b8c0dd1260f6f5928ec3R18

kaladron (approved):
Thanks for addressing the comments!

PR #186166, opened 2026-03-12
[Flang][MLIR][OpenMP] Explicitly represent omp.target kernel types

skatrak:
Currently, the kernel type (i.e. `generic`, `spmd`, `spmd-no-loop` and `bare`) of an `omp.target` operation is not an explicit attribute of the operation. Rather, this is inferred based on the contents of its region and clauses.

The problems with this approach are that it can be a potentially resource intensive check for large kernels, and misidentifications are prone to happen based on the presence of arbitrary operations from other dialects.

Since the AST already contains the information needed to identify the kernel type in a more reliable manner, this patch moves that responsiblity to the Flang frontend. Other MLIR passes that create `omp.target` operations are updated as well.

One known limitation of this approach is that the MLIR op verifier for `omp.target` can't completely check that the contents of its region are compatible with the declared kernel type without being exposed to the same pattern-matching limitations that this patch is removing. Also, the `TargetOp::getInnermostCapturedOmpOp()` function is maintained but, ideally, a better solution should be implemented to remove its expensive and potentially flaky checks from MLIR.

PR #198782, opened 2026-05-20
[MLIR][OpenMP] Explicit tagging of combined constructs

skatrak:
Combined OpenMP constructs, such as `parallel do`, which represent nests of constructs where each one contains a single other construct without any other directives or statements in between, are currently not marked in any way in the MLIR representation.

This works because they don't usually require any specific handling other than what would be done for the included operations. However, the handling of `target` regions needs to know whether it was part of a combined construct in order to properly optimize for the SPMD case and detect when certain clauses must be inconditionally evaluated in the host.

So far, this has been achieved by having some MLIR pattern-matching logic to infer whether a nest of operations could have potentially been produced for a combined construct. This approach is error prone, computationally expensive and it can't really work in the general case. On the other hand, a compiler frontend can easily tell the difference and tag MLIR operations accordingly.

This patch extends the `ComposableOpInterface` of the OpenMP dialect to handle a new `omp.combined` attribute that must be set for all leafs (except for the innermost one) on a combined construct. Verification logic is added for this interface, which is added to all operations that can be used as part of a combined construct, and the previous `target`-related pattern-matching logic is removed.

This patch has to be followed up with Flang lowering changes to pass all unit tests.

### DIFF
--- a/mlir/lib/Dialect/OpenMP/IR/OpenMPDialect.cpp
+++ b/mlir/lib/Dialect/OpenMP/IR/OpenMPDialect.cpp
@@ -2571,7 +2571,8 @@ bool TargetOp::hasHostEvalTripCount() {
 
   // If it represents a `target teams distribute` construct, also evaluate the
   // `distribute` trip count on the host.
-  Operation *capturedOp = getInnermostCapturedOmpOp();
+  Operation *capturedOp =
+      cast<ComposableOpInterface>(getOperation()).findCapturedOp();
   if (auto loopNestOp = dyn_cast_if_present<LoopNestOp>(capturedOp)) {
     SmallVector<LoopWrapperInterface> loopWrappers;
     loopNestOp.gatherWrappers(loopWrappers);
@@ -2597,6 +2598,9 @@ bool TargetOp::hasHostEvalTripCount() {
 }
 
 LogicalResult TargetOp::verify() {
+  if (getKernelType() == TargetExecMode::bare && !isCombined())
+    return emitOpError() << "bare kernel requires 'omp.combined'";
+
   if (failed(verifyDependVarList(*this, getDependKinds(), getDependVars(),
                                  getDependIteratedKinds(),
                                  getDependIterated())))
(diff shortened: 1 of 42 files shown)


skatrak:
PR stack:
- #186166
- #198782 ◀️
- #198783

mlir/lib/Dialect/OpenMP/IR/OpenMPDialect.cpp
```diff
@@ -3160,6 +3065,98 @@ LogicalResult LoopWrapperInterface::verifyImpl() {
...
+  if (isCombined()) {
+    Operation *nestedOp = nullptr;
+    auto count = llvm::count_if(
+        op->getRegion(0).getOps(), [&nestedOp](mlir::Operation &op) {
+          if (isa<ComposableOpInterface, LoopWrapperInterface>(op)) {
+            nestedOp = &op;
+            return true;
+          }
+          return false;
+        });
```
tblah, line 3148: Maybe we should also reject any side effecting operation inside the nest which is not part of the composition?

skatrak: That's actually one of the big reasons I've gone through the process of making the frontend responsible for tagging combined constructs. The previous pattern-based approach actually did something like that: "a couple of nested OpenMP constructs are combined if there are no other OpenMP or side-effecting ops, or control flow in the parent op's region".

That regularly broke because of e.g. debug ops that had side effects just to avoid being optimized out or due to how certain clauses of the child op were lowered to MLIR, potentially adding memory allocations, etc. The current pattern (`TargetOp::getInnermostCapturedOmpOp` + `findCapturedOmpOp`) kind of works around those cases that we've detected so far, while also allowing operations that should probably break it.

Checks here after this change are more lenient on that sense on purpose. We just accept that we need to trust the frontend on this because it can't be reliably checked in MLIR. The thinking here is that a frontend adds the `omp.combined` attribute if it **knows** it can do so, but it would always (minus the `target teams ompx_bare` case) be correct to not put it there, though this could have a potential performance impact.

tblah: Okay sounds good. I was initially worried that allowing these could expose new bugs in MLIR->LLVM conversion but I didn't find any so this LGTM.

I think it is okay to rely on the frontend to produce something that makes sense in terms of OpenMP, but I want the MLIR dialect to either produce some LLVM-IR (fine to be undefined behaviour if the front end emitted undefined behaviour) or reject the IR as invalid: no crashes. It looks like that is achieved here.

tblah (approved):
LGTM

PR #198783, opened 2026-05-20
[Flang][OpenMP] Add combined construct information

skatrak:
This patch adds the `omp.combined` attribute to OpenMP dialect operations following changes to the `ComposableOpInterface`.

This attribute is added to operations representing non-innermost leaf constructs of a combined construct and to standalone block-associated constructs that can be combined with their parent construct.

Changes are made to the OpenMP lowering logic, as well as the do-concurrent, workshare and workdistribute transformation passes.

### DIFF
--- a/flang/lib/Lower/OpenMP/OpenMP.cpp
+++ b/flang/lib/Lower/OpenMP/OpenMP.cpp
@@ -79,6 +79,25 @@ makeObjects(llvm::ArrayRef<const semantics::Symbol *> syms) {
   return objects;
 }
 
+/// Return the directive that is immediately nested inside of the given
+/// \c parent evaluation, if it is its only non-end-statement nested evaluation
+/// and it represents an OpenMP construct.
+lower::pft::Evaluation *
+extractOnlyOmpNestedEval(lower::pft::Evaluation &parent) {
+  if (!parent.hasNestedEvaluations())
+    return nullptr;
+
+  auto &nested{parent.getFirstNestedEvaluation()};
+  if (!nested.isA<parser::OpenMPConstruct>())
+    return nullptr;
+
+  for (auto &sibling : parent.getNestedEvaluations())
+    if (&sibling != &nested && !sibling.isEndStmt())
+      return nullptr;
+
+  return &nested;
+}
+
 namespace {
 /// Structure holding information that is needed to pass host-evaluated
 /// information to later lowering stages.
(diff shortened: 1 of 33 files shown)


skatrak:
PR stack:
- #186166
- #198782
- #198783 ◀️

tblah (approved):
LGTM, thanks!

Meinersbur (approved):

flang/lib/Lower/OpenMP/OpenMP.cpp
```diff
@@ -79,6 +79,25 @@ makeObjects(llvm::ArrayRef<const semantics::Symbol *> syms) {
...
 
+/// Return the directive that is immediately nested inside of the given
+/// \c parent evaluation, if it is its only non-end-statement nested evaluation
+/// and it represents an OpenMP construct.
+lower::pft::Evaluation *
+extractOnlyOmpNestedEval(lower::pft::Evaluation &parent) {
+  if (!parent.hasNestedEvaluations())
+    return nullptr;
+
+  auto &nested{parent.getFirstNestedEvaluation()};
```
Meinersbur:
```suggestion
  auto &nested = parent.getFirstNestedEvaluation();
```
[not a change request] declaration initializer quite inconsistent in this file

flang/lib/Lower/OpenMP/OpenMP.cpp
```diff
@@ -4189,6 +4189,42 @@ static void genOMPDispatch(lower::AbstractConverter &converter,
   finalizeStmtCtx();
   if (loopLeaf)
     symTable.popScope();
+
+  // Add the omp.combined attribute to eligible ops. In this case, all
+  // composable ops that are not loop-associated, except for the ones that can
+  // only appear as the innermost leaf construct.
```
Meinersbur:
```suggestion
  // Add the omp.combined attribute to immediate nested construct that are equivalent to combined construct as an optimization. In this case, all
  // composable ops that are not loop-associated, except for the ones that can
  // only appear as the innermost leaf construct.
```
[suggestion] The interface docs if isCombined says "Check whether the operation is representing a non-innermost child leaf of a combined OpenMP construct or an equivalent immediate nesting of constructs." Having the info that the attribute is also for equivalent nestings (in contrast to it only describing source form) would have helped me here.

Feel free to leave out the "as an optimization" if you think it is semantically required.

Meinersbur (approved):
LGTM

flang/lib/Lower/OpenMP/OpenMP.cpp
skatrak: Done. We probably should make a clear exception for Flang lowering and just use MLIR formatting rules, rather than a mix.

flang/lib/Lower/OpenMP/OpenMP.cpp
skatrak: Thanks for pointing this out, I made some clarifications to the comments there. Hopefully it's easier to follow now, let me know.

PR #198785, opened 2026-05-20
[AMDGPU][GlobalISel] Combine rcp(sqrt(x)) into rsq for pseudo-scalar instructions

MadhurKumar004:
This fixes "TODO: GlobalISel should generate v_s_rsq." give in file "llvm/test/CodeGen/AMDGPU/pseudo-scalar-transcendental.ll"

### DIFF
--- a/llvm/test/CodeGen/AMDGPU/fdiv_flags.f32.ll
+++ b/llvm/test/CodeGen/AMDGPU/fdiv_flags.f32.ll
@@ -1620,12 +1620,18 @@ define float @v_recip_sqrt_f32_afn_ulp25_contract(float %x) {
 ; CODEGEN-IEEE-GISEL-NEXT:    v_rsq_f32_e32 v0, v0
 ; CODEGEN-IEEE-GISEL-NEXT:    s_setpc_b64 s[30:31]
 ;
-; IR-IEEE-LABEL: v_recip_sqrt_f32_afn_ulp25_contract:
-; IR-IEEE:       ; %bb.0:
-; IR-IEEE-NEXT:    s_waitcnt vmcnt(0) expcnt(0) lgkmcnt(0)
-; IR-IEEE-NEXT:    v_sqrt_f32_e32 v0, v0
-; IR-IEEE-NEXT:    v_rcp_f32_e32 v0, v0
-; IR-IEEE-NEXT:    s_setpc_b64 s[30:31]
+; IR-IEEE-SDAG-LABEL: v_recip_sqrt_f32_afn_ulp25_contract:
+; IR-IEEE-SDAG:       ; %bb.0:
+; IR-IEEE-SDAG-NEXT:    s_waitcnt vmcnt(0) expcnt(0) lgkmcnt(0)
+; IR-IEEE-SDAG-NEXT:    v_sqrt_f32_e32 v0, v0
+; IR-IEEE-SDAG-NEXT:    v_rcp_f32_e32 v0, v0
+; IR-IEEE-SDAG-NEXT:    s_setpc_b64 s[30:31]
+;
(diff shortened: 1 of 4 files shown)


llvm/test/CodeGen/AMDGPU/fdiv_flags.f32.ll
```diff
@@ -1620,25 +1620,37 @@ define float @v_recip_sqrt_f32_afn_ulp25_contract(float %x) {
...
+; IR-IEEE-SDAG:       ; %bb.0:
+; IR-IEEE-SDAG-NEXT:    s_waitcnt vmcnt(0) expcnt(0) lgkmcnt(0)
+; IR-IEEE-SDAG-NEXT:    v_sqrt_f32_e32 v0, v0
+; IR-IEEE-SDAG-NEXT:    v_rcp_f32_e32 v0, v0
+; IR-IEEE-SDAG-NEXT:    s_setpc_b64 s[30:31]
+;
+; IR-IEEE-GISEL-LABEL: v_recip_sqrt_f32_afn_ulp25_contract:
+; IR-IEEE-GISEL:       ; %bb.0:
+; IR-IEEE-GISEL-NEXT:    s_waitcnt vmcnt(0) expcnt(0) lgkmcnt(0)
+; IR-IEEE-GISEL-NEXT:    v_rsq_f32_e32 v0, v0
```
jayfoad, line 1633: Should probably investigate whether it's OK to use v_rsq here, and why SelectionDAG does not.

jayfoad (approved):
LGTM

llvm/test/CodeGen/AMDGPU/fdiv_flags.f32.ll
MadhurKumar004: I investigated why SDAG doesnt emit `v_rsq` in `IR-IEEE` + diabled codegenprepare path. So SDAG's `rcp(sqrt(x))` -> `rsq` combine depends on CodeGenPrepare having pre-expanded `fdiv(1.0, sqrt(x))` into an ir level reciprocal first, when `-amdgpu-codegenprepare-disable-fdiv-expansion=1` is used, that expansion doesnt happen, as a result sdag see a raw `fdiv` node causing its estimate combine to never fire. this is a pre-existing SDAG phase-ordering limitation specific to the disable-expansion path. Looking at `fdiv_flag.f32.ll` when CodeGenPrepare run normally (`CODEGEN-IEEE-SDAG`), SDAG correctly emits `v_rsq_f32` for contract-only cases proving that sdag itself consider contract sufficient justification for using `v_rsq`. The IR-IEEE-SDAG path failing to do so here i think is an optimization gap?

Gisel doesnt suffer from this limitation because its legalizer independently lowers `G_FSQRT` -> `amdgcn_sqrt` and `G_FDIV(1.0, x)` -> `amdgcn_rcp` regardless of what CodeGenPrepare did, so it naturally produces the `rcp(sqrt(x))` pattern that `matchRcpSqrtToRsq` needs. 

Regarding whether `v_rsq` is Ok to use: the IR has both `contract` and `afn` on both the sqrt and fdiv. `v_rsq_f32` is documented as of 1ULP accuracy , within that 2.5 ULP budget of `!fpmath !0`. afn permits approximate result and contract permits fusing. Gisel producing `v_rsq_f32` here is correct and matches what sdag does when CodeGenPrepare runs normally.

MadhurKumar004:
@jayfoad Friendly ping

MadhurKumar004:
@jayfoad

jayfoad:
What changed since my last approval?

jayfoad (approved):

MadhurKumar004:
No changes, i pushed again because -  i did a rebase and some MLIR test cases were giving false negative.

MadhurKumar004:
Can you push it as i dont have the access?

PR #197959, opened 2026-05-15
[LV] Support interleaving in order to reduce stalls

john-brawn-arm:
On an in-order CPU the CPU will stall after a load if the value is used immediately. Interleaving such that the combined reciprocal throughput of all of the load instructions is equal to the load latency means we will avoid this stall.

PR #198790, opened 2026-05-20
[CostModel] Move the arbitrary load latency into getMemoryOpCost

john-brawn-arm:
Currently TargetTransformInfoImpl returns an arbitrary cost of 4 for the latency of loads in getInstructionCost. This means even if a target correctly reports the latency for loads in getMemoryOpCost we still get this arbitrary cost in getInstructionCost. It also means the latency cost is inconsistent depending on whether you go through getInstructionCost or getMemoryOpCost.

Solve this by moving the current arbitrary cost into getMemoryOpCost. This has the side-effect of affecting the cost of masked loads if they aren't handled by the target, as in BasicTTIImpl the cost for these is calculated using getMemoryOpCost. This should mean the cost is more accurate though, and likely won't have any effect as in any transformation that could introduce masked loads (e.g. vectorization) the current cost is probably high enough that it's already not worth using.


john-brawn-arm:
This is related to #197959, where we need accurate load latency cost. I'm going to be adding this to the AArch64 backend, but as things currently are it's impossible to test in test/Analysis/CostModel because that uses getInstructionCost. Hence this PR moving the cost into getMemoryOpCost, so that when I update getMemoryOpCost to have a more accurate cost it'll be reflected in the test/Analysis/CostModel tests.

llvm/lib/Target/AArch64/AArch64TargetTransformInfo.cpp
```diff
@@ -5028,6 +5028,10 @@ InstructionCost AArch64TTIImpl::getMemoryOpCost(unsigned Opcode, Type *Ty,
                                                 TTI::TargetCostKind CostKind,
                                                 TTI::OperandValueInfo OpInfo,
                                                 const Instruction *I) const {
+  // FIXME: Arbitrary cost
+  if (Opcode == Instruction::Load && CostKind == TTI::TCK_Latency)
+    return 4;
```
sparker-arm: So, why not just call `BaseT::getMemoryOpCost`? Seems a bit odd to duplicate this arbitrary value into all the backends.

john-brawn-arm: Because I hadn't thought to do that. I've adjusted things to do that (though in many places BaseT::getMemoryOpCost gets called for other reasons, so there I've adjusted the condition).

sparker-arm (approved):
LGTM, cheers.

RKSimon (approved):
x86 change is ok for now

PR #198791, opened 2026-05-20
[libc] Fix inet_aton

labath:
The main (in terms of LOC) change is moving the implementation to an internal function in order to avoid the inet_addr->inet_aton dependency.

I also fix a bug where we (mistakenly) accepted whitespace and signs inside the address. I also match the glibc implementation in ignoring the data after the first whitespace.

### DIFF
--- a/libc/src/__support/net/address.cpp
+++ b/libc/src/__support/net/address.cpp
@@ -1,0 +1,65 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file implements helper functions for parsing network addresses.
+///
+//===----------------------------------------------------------------------===//
+
+#include "src/__support/net/address.h"
+#include "src/__support/common.h"
+#include "src/__support/ctype_utils.h"
+#include "src/__support/endian_internal.h"
+#include "src/__support/str_to_integer.h"
+
+namespace LIBC_NAMESPACE_DECL {
+namespace internal {
+
+cpp::optional<in_addr_t> inet_addr(const char *cp) {
+  constexpr int IPV4_MAX_DOT_NUM = 3;
+  in_addr_t parts[IPV4_MAX_DOT_NUM + 1] = {0};
+  int dot_num = 0;
+
+  for (; dot_num <= IPV4_MAX_DOT_NUM; ++dot_num) {
+    // strtointeger skips leading whitespace signs (1.+2.-3. 4), but we don't
+    // want that, so we explicitly check that the first character is a digit.
+    if (!isdigit(*cp))
+      return cpp::nullopt;
+
+    auto result = strtointeger<in_addr_t>(cp, 0);
+    parts[dot_num] = result;
+
+    if (result.has_error() || result.parsed_len == 0)
+      return cpp::nullopt;
+    cp += result.parsed_len;
+    if (*cp == '\0' || isspace(*cp))
+      break;
+    if (*cp != '.')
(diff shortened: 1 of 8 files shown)


libc/src/__support/net/address.cpp
labath, line 1: I wasn't sure where to put the new file. I created a new (and fairly generic) folder, so that it can also house things like the hypothetical address resolver, but I'm open to other suggestions.

libc/src/__support/net/address.cpp
```diff
@@ -0,0 +1,65 @@
...
+  in_addr_t result = 0;
+  for (int i = 0; i <= dot_num; ++i) {
+    in_addr_t max_part = i == dot_num ? (0xffffffffu >> (8 * dot_num)) : 0xffu;
+    if (parts[i] > max_part)
+      return cpp::nullopt;
+    int shift = i == dot_num ? 0 : 8 * (IPV4_MAX_DOT_NUM - i);
+    result |= parts[i] << shift;
+  }
+
+  return LIBC_NAMESPACE::Endian::to_big_endian(result);
```
labath: This actually has to be **LIBC_NAMESPACE**::Endian, because internal::Endian is an implementation detail of the endian class. This makes me a bit unsure if the internal namespace is meant to be an implementation detail of libc as a whole (i.e., pretty much everything should go there), or just an implementation detail of individual parts of libc.

libc/src/__support/net/address.cpp
kaladron: I believe this is correct.

libc/src/__support/net/address.cpp
kaladron: This one I'm less sure of.  @michaelrj-google ?

michaelrj-google: I don't think we've applied consistent standards on what goes in the `internal` namespace vs not. Generally `internal` gets used for code in `__support`, but beyond that it's whatever the author felt like. If I were to write a rule now I'd probably say "use the `internal` namespace for pieces you don't expect to be used by code outside of `__support`" but even that is probably too vague.

libc/src/__support/net/address.cpp
michaelrj-google: code in `__support` is usually header only, though I don't think this is a hard rule. It's more a consequence of frequent use of templates, or desire to share code via hand-in-hand.

michaelrj-google (approved):
LGTM with a few comments

libc/src/__support/net/address.cpp
labath: Is that a request to make this header only? If so, I can, though I'd rather not. :)

I'm also thinking about other network code that could go in here. getaddrinfo in particular is going to be fairly big. While there's no reason that couldn't be made header-only, doing that seems a bit wasteful.

OTOH, one of the problems we'll need to solve there is how to test that code, and *one* possibility is to make that a template (to allow mocking of syscalls). If we go down that path, pretty much all the code is going to have to be in a header anyway...

libc/src/__support/net/address.cpp
labath: That makes sense, and my only problem with that is figuring out how to call this function then, as it sort of really is the *internal* implementation of *inet_addr*. Looking around, I think that has been the problem for many other functions. Like I'm pretty sure that's how the functions in __support/wchar ended up in the internal namespace -- though maybe they're not the best example to follow as they also which file-local static functions, which [is not supposed to happen](https://libc.llvm.org/dev/code_style.html#inline-functions-and-variables-defined-in-header-files). Math functions solve this by creating their own (sub)namespace, so the new version of the patch does the same.

LMK is that's okay.

libc/src/__support/net/address.cpp
michaelrj-google: This code doesn't need to be header only. That's more of a fun fact. Jeff has been talking about making some syscall mocking system, but I don't think we're there yet. For now my policy has been that we need to test our code but not the kernel's code. We should try to make a syscall wrapper pass and fail in the test since that will exercise the branches, but we don't need to test all behaviors.

PR #198793, opened 2026-05-20
[flang-rt] Remove library dependency from flang-rt to offload

jhuber6:
Summary:
We need the offload project's RPC thread to handle the IO requests
originating from the GPU. Previously we did the 'easy' solution and just
linked this handler directly into the offload proejct. This is not ideal
because it prevents people's ability to build and configure libraries
separately.

This PR inverts the dependency, flang-rt now conditionally enables
support using the existing RPC callback mechanism. The cost is that
every flang-rt program now pays the cost of a boolean compare, the
benefit is the libraries are now independent of each-other.

### DIFF
--- a/flang-rt/lib/runtime/CMakeLists.txt
+++ b/flang-rt/lib/runtime/CMakeLists.txt
@@ -111,6 +111,8 @@ set(host_sources
 )
 if (TARGET llvm-libc-common-utilities)
   list(APPEND host_sources io-api-server.cpp)
+  set_property(SOURCE environment.cpp APPEND PROPERTY
+    COMPILE_DEFINITIONS FLANG_RT_HAS_RPC_SERVER)
 endif()
 
 # Sources that can be compiled directly for the GPU.
--- a/flang-rt/lib/runtime/CMakeLists.txt
+++ b/flang-rt/lib/runtime/CMakeLists.txt
@@ -111,7 +111,7 @@ set(host_sources
 )
 if (TARGET llvm-libc-common-utilities)
   list(APPEND host_sources io-api-server.cpp)
-  set_property(SOURCE environment.cpp APPEND PROPERTY
+  set_property(SOURCE main.cpp APPEND PROPERTY
     COMPILE_DEFINITIONS FLANG_RT_HAS_RPC_SERVER)
 endif()
 
--- a/flang-rt/lib/runtime/environment.cpp
+++ b/flang-rt/lib/runtime/environment.cpp
@@ -8,6 +8,7 @@
 
 #include "flang-rt/runtime/environment.h"
 #include "environment-default-list.h"
+#include "io-api-gpu.h"
 #include "flang-rt/runtime/memory.h"
 #include "flang-rt/runtime/tools.h"
 #include <cstdio>
@@ -267,6 +268,11 @@ void ExecutionEnvironment::Configure(int ac, const char *av[],
       PostConfigEnvCallback[i](ac, av, env, envDefaults);
     }
   }
+
+  // Registers the IO handlers with the offloading runtime only if present.
+#ifdef FLANG_RT_HAS_RPC_SERVER
+  Fortran::runtime::io::RegisterRPCHandlers();
+#endif
 }
 
 const char *ExecutionEnvironment::GetEnv(
--- a/flang-rt/lib/runtime/environment.cpp
+++ b/flang-rt/lib/runtime/environment.cpp
@@ -8,7 +8,6 @@
 
 #include "flang-rt/runtime/environment.h"
 #include "environment-default-list.h"
-#include "io-api-gpu.h"
 #include "flang-rt/runtime/memory.h"
 #include "flang-rt/runtime/tools.h"
 #include <cstdio>
@@ -268,11 +267,6 @@ void ExecutionEnvironment::Configure(int ac, const char *av[],
       PostConfigEnvCallback[i](ac, av, env, envDefaults);
     }
   }
-
-  // Registers the IO handlers with the offloading runtime only if present.
-#ifdef FLANG_RT_HAS_RPC_SERVER
-  Fortran::runtime::io::RegisterRPCHandlers();
-#endif
 }
 
 const char *ExecutionEnvironment::GetEnv(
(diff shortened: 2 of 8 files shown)


flang-rt/lib/runtime/environment.cpp
```diff
@@ -267,6 +268,11 @@ void ExecutionEnvironment::Configure(int ac, const char *av[],
       PostConfigEnvCallback[i](ac, av, env, envDefaults);
     }
   }
+
+  // Registers the IO handlers with the offloading runtime only if present.
+#ifdef FLANG_RT_HAS_RPC_SERVER
+  Fortran::runtime::io::RegisterRPCHandlers();
+#endif
```
Meinersbur: I think this fits better into `ProgramStart` in `main.cpp`. RPC handlers is not about env vars and there already is a comment about IO.

flang-rt/lib/runtime/io-api-server.cpp
```diff
@@ -292,10 +292,20 @@ rpc::RPCStatus HandleOpcodesImpl(rpc::Server::Port &port) {
...
+} // anonymous namespace
+
+using RPCCallbackTy = std::uint32_t (*)(void *, std::uint32_t);
+
+// COFF workaround for the lack of weak symbols, copied from the sanitizers.
+#if defined(_MSC_VER)
+extern "C" void register_rpc_callback_stub(RPCCallbackTy) {}
+#pragma comment(linker, \
+    "/alternatename:__tgt_register_rpc_callback=" \
+    "register_rpc_callback_stub")
```
Meinersbur, line 307: Where is `register_rpc_callback_stub` implemented?

flang-rt/lib/runtime/io-api-server.cpp
```diff
@@ -292,10 +292,20 @@ rpc::RPCStatus HandleOpcodesImpl(rpc::Server::Port &port) {
...
+#endif
+
+static std::uint32_t HandleRPCOpcodes(void *raw, std::uint32_t numLanes) {
```
Meinersbur, line 311: Could you add a comment what this function does?

You removed the line
```
// Used for I/O from the offloading device.
```
In addition to "used" to, it would be good what calls it, when it is called, and what it is doing.

flang-rt/lib/runtime/io-api-server.cpp
jhuber6: Right above, it's an empty function, but if the runtime symbol exists it will override it to provide a similar interface to extern weak. I believe that's how it works on Windows.

Meinersbur: OK, I learned to see one-line extern "C" as declaration, not definition. Sorry.

Could you add a comment into the implementation such as
```
extern "C" void register_rpc_callback_stub(RPCCallbackTy) {
  // This function is only called when liboffload with the actual __tgt_register_rpc_callback implementation is not present; do nothing if liboffload is not loaded as there are no RPC calls to receive.
}
```

flang-rt/lib/runtime/CMakeLists.txt
```diff
@@ -111,6 +111,8 @@ set(host_sources
 )
 if (TARGET llvm-libc-common-utilities)
```
Meinersbur, line 112: [unrelated to this PR] If I read this correctly, `include(FindLibcCommonUtils)` in `flang-rt/CMakeLists.txt` will ensure that `llvm-libc-common-utilities` is always created, making `if (TARGET llvm-libc-common-utilities)` unnecessary.

Meinersbur (approved):
LGTM, thanks

flang-rt/lib/runtime/CMakeLists.txt
jhuber6: This was defensive, I might remove it in a follow up. Basically I was just guarding in the case that anyone extracts `flang-rt` only without `libc/` but that's probably not common.

Meinersbur: It doesn't even compile without it:
```cpp
ld.lld: error: undefined symbol: Fortran::runtime::io::RegisterRPCHandlers()
>>> referenced by main.cpp
>>>               main.cpp.o:(_FortranAProgramStart) in archive /home/meinersbur/src/llvm/main/release/lib/clang/23/lib/x86_64-unknown-linux-gnu/libflang_rt.runtime.a
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

I don't think we need (or want to) support random changes to the repository checkout. The top-level directories have not been independent for some time now, and flang-rt has always cross-referenced flang/.

jhuber6: Fair enough, I'll land this and remove that guard. I think this PR made it not work anyway.

kkwli:
@jhuber6 The patch causes the undefined symbol error on MacOS.
```
Undefined symbols for architecture arm64:
  "___tgt_register_rpc_callback", referenced from:
      Fortran::runtime::io::RegisterRPCHandlers() in io-api-server.cpp.o
      Fortran::runtime::io::RegisterRPCHandlers() in io-api-server.cpp.o
ld: symbol(s) not found for architecture arm64

```

cc @luporl

jhuber6:
> @jhuber6 The patch causes the undefined symbol error on MacOS.
> 
> ```
> Undefined symbols for architecture arm64:
>   "___tgt_register_rpc_callback", referenced from:
>       Fortran::runtime::io::RegisterRPCHandlers() in io-api-server.cpp.o
>       Fortran::runtime::io::RegisterRPCHandlers() in io-api-server.cpp.o
> ld: symbol(s) not found for architecture arm64
> ```
> 
> cc @luporl

Honestly, we don't even support offloading on MacOS, but I would've expected this to work. Could you share what you get for `clang --print-target-triple`? Means that this is not working for some reason.
```c
#elif defined(__APPLE__)
extern "C" void __tgt_register_rpc_callback(RPCCallbackTy)
    __attribute__((weak_import));
#else
```

kkwli:
> 
> Honestly, we don't even support offloading on MacOS, but I would've expected this to work. Could you share what you get for `clang --print-target-triple`? Means that this is not working for some reason.
> 
> ```c
> #elif defined(__APPLE__)
> extern "C" void __tgt_register_rpc_callback(RPCCallbackTy)
>     __attribute__((weak_import));
> #else
> ```

```
% clang --print-target-triple
arm64-apple-darwin24.6.0
```

mmuetzel:
Could it matter that `___tgt_register_rpc_callback` in the undefined symbol error has three leading `_`, but the `extern "C"` function declaration has only two?

Issue #132800, opened 2025-03-24
[CI] Enable lldb on windows premerge


boomanaiden154:
We currently do not test lldb on windows premerge as it needs some custom environment setup that has not been done. We should figure out what needs to be done and then look at enabling it.

Nerixyz:
From https://github.com/llvm/llvm-project/blob/cc1eae6ea2de5ccadb81a182b9f0de2128b9fefc/.github/workflows/containers/github-action-ci-windows/Dockerfile, it looks like only GnuWin32 is missing ([buildbot reference](https://github.com/llvm/llvm-zorg/blob/9e8c5a43a4cf7004dd5bc84e1171f16da3408698/buildbot/google/docker/windows-base-vscode2019/Dockerfile#L49-L64)). I'm not sure if that conflicts with the `C:\llvm-mingw\bin`.

charles-zablit:
I made some progress on this and was able to get a green build (minus a test failure which I also reproduced at desk on a dap test). The 2 main issues:
- `make` is not installed in the Windows container.
- lldb skips resolving module paths if the debug event has `info.hFile == NULL`.

The following PRs fix all of those issues:
1. https://github.com/llvm/llvm-project/pull/198794
2. https://github.com/llvm/llvm-project/pull/198795
3. https://github.com/llvm/llvm-project/pull/198807
4. https://github.com/llvm/llvm-project/pull/198814

charles-zablit:
https://github.com/llvm/llvm-project/pull/198630 has an almost green run:

```
********************
Unresolved Tests (2):
  lldb-api :: tools/lldb-dap/launch/TestDAP_launch_stdio_redirection_and_console.py
  lldb-api :: tools/lldb-dap/runInTerminal/TestDAP_runInTerminal.py

********************
Failed Tests (1):
  lldb-api :: tools/lldb-dap/restart/TestDAP_restart_console.py
```

Those tests are failing due to a race condition in dap. I don't think this has anything to do with the fact they are running in a container. Once that issue is fixed in main, the run should be green.

charles-zablit:
A fully green build with pre merge checks for lldb on Windows: https://github.com/llvm/llvm-project/actions/runs/26762943126/job/78881288876?pr=198906

PR #198795, opened 2026-05-20
[lldb][windows] try to resolve module path if info.hFile is null

charles-zablit:
The Windows API can deliver `LOAD_DLL_DEBUG_EVENT` with `hFile` set to NULL. In practice, this happens in practice
inside Windows Server Core containers for system DLLs like `kernel32.dll`, `KernelBase.dll` and `ucrtbase.dll`. This is the issue described by @Nerixyz in https://github.com/llvm/llvm-project/issues/132800.
    
`DebuggerThread::HandleLoadDllEvent` previously dropped any such event.
    
This patch improves the module path resolution with the 2 following methods:
- `GetFileNameByLoadAddress`: resolve the module path given its base load address.
- `GetFileNameFromImageNameField`: resolve the module path given the `LOAD_DLL_DEBUG_INFO::lpImageName` field.

This requires:
- https://github.com/llvm/llvm-project/pull/198794

### DIFF
--- a/lldb/source/Plugins/Process/Windows/Common/DebuggerThread.cpp
+++ b/lldb/source/Plugins/Process/Windows/Common/DebuggerThread.cpp
@@ -540,16 +540,67 @@ static std::optional<std::string> GetFileNameFromHandleFallback(HANDLE hFile) {
   return ConvertNtDevicePathToDosPath(mapped_filename);
 }
 
+static std::optional<std::string> GetFileNameByLoadAddress(HANDLE hProcess,
+                                                           LPVOID base_addr) {
+  std::array<wchar_t, MAX_PATH + 1> module_filename;
+  DWORD len =
+      ::GetModuleFileNameExW(hProcess, reinterpret_cast<HMODULE>(base_addr),
+                             module_filename.data(), module_filename.size());
+  if (len > 0 && len < module_filename.size()) {
+    std::string path_utf8;
+    llvm::convertWideToUTF8(std::wstring(module_filename.data(), len),
+                            path_utf8);
+    return path_utf8;
+  }
+
+  // Fallback: ask the kernel for the file backing the mapping at this address.
+  std::array<wchar_t, MAX_PATH + 1> mapped_filename;
+  if (!::GetMappedFileNameW(hProcess, base_addr, mapped_filename.data(),
+                            mapped_filename.size()))
+    return std::nullopt;
+  return ConvertNtDevicePathToDosPath(mapped_filename);
+}
+
+// Resolve the LOAD_DLL_DEBUG_INFO::lpImageName field.
+static std::optional<std::string>
+GetFileNameFromImageNameField(HANDLE hProcess,
+                              const LOAD_DLL_DEBUG_INFO &info) {
+  if (info.lpImageName == nullptr)
+    return std::nullopt;
+
+  LPVOID name_addr = nullptr;
+  SIZE_T bytes_read = 0;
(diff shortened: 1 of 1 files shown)


lldb/source/Plugins/Process/Windows/Common/DebuggerThread.cpp
```diff
@@ -617,4 +681,4 @@ DebuggerThread::HandleRipEvent(const RIP_INFO &info, DWORD thread_id) {
   m_debug_delegate->OnDebuggerError(error, info.dwType);
 
   return DBG_CONTINUE;
-}
+}
```
Nerixyz:
```suggestion
}

```

lldb/source/Plugins/Process/Windows/Common/DebuggerThread.cpp
```diff
@@ -540,16 +540,67 @@ static std::optional<std::string> GetFileNameFromHandleFallback(HANDLE hFile) {
...
 
+static std::optional<std::string> GetFileNameByLoadAddress(HANDLE hProcess,
+                                                           LPVOID base_addr) {
+  std::array<wchar_t, MAX_PATH + 1> module_filename;
+  DWORD len =
+      ::GetModuleFileNameExW(hProcess, reinterpret_cast<HMODULE>(base_addr),
+                             module_filename.data(), module_filename.size());
+  if (len > 0 && len < module_filename.size()) {
+    std::string path_utf8;
+    llvm::convertWideToUTF8(std::wstring(module_filename.data(), len),
```
Nerixyz, line 551: Feels like `convertWideToUTF8` should take a `std::wstring_view` or `ArrayRef<wchar_t>` instead of a `const std::wstring&`.

lldb/source/Plugins/Process/Windows/Common/DebuggerThread.cpp
```diff
@@ -540,16 +540,67 @@ static std::optional<std::string> GetFileNameFromHandleFallback(HANDLE hFile) {
...
+
+  if (info.fUnicode) {
+    std::array<wchar_t, MAX_PATH + 1> wbuf{};
+    if (!::ReadProcessMemory(hProcess, name_addr, wbuf.data(),
+                             (wbuf.size() - 1) * sizeof(wchar_t), &bytes_read))
+      return std::nullopt;
+    std::string path_utf8;
+    llvm::convertWideToUTF8(wbuf.data(), path_utf8);
+    if (path_utf8.empty())
+      return std::nullopt;
```
Nerixyz, line 588: If the name is longer than MAX_PATH, it would get truncated, but I'm guessing that can't happen?

lldb/source/Plugins/Process/Windows/Common/DebuggerThread.cpp
charles-zablit: Agreed, should we do this in another PR?

Nerixyz: Yea, definitely. I recall encountering this before too. There should be a few other places in LLVM using this as well.

lldb/source/Plugins/Process/Windows/Common/DebuggerThread.cpp
charles-zablit: I added guards in both cases. For DLLs, they should be shorter than `MAX_PATH` in practice.

Nerixyz (approved):

PR #164091, opened 2025-10-18
[libc][annex_k] Add libc_constraint_handler.

bassiounix:
RFC https://discourse.llvm.org/t/rfc-bounds-checking-interfaces-for-llvm-libc/87685

Add internal `libc_constraint_handler` required by Annex K interface in LLVM libc.

PR #198797, opened 2026-05-20
[libc] Add Annex K qsort_s definition and tests

vhscampos:
This patch introduces Annex K's `qsort_s`.

Since its semantics are very similar to the ones for `qsort_r`, the two share the same testing logic, except that for `qsort_s` we also test the constraint violation semantics.


vhscampos:
Depends on: https://github.com/llvm/llvm-project/pull/164091, https://github.com/llvm/llvm-project/pull/164093, https://github.com/llvm/llvm-project/pull/197707, https://github.com/llvm/llvm-project/pull/198781

libc/src/stdlib/qsort_s.cpp
```diff
@@ -0,0 +1,47 @@
+//===-- Implementation of qsort -------------------------------------------===//
```
michaelrj-google: nit: please use the new style for file headers: https://llvm.org/docs/CodingStandards.html#file-headers

here and below

libc/src/stdlib/qsort_s.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "src/stdlib/qsort_s.h"
+#define __STDC_WANT_LIB_EXT1__ 1
+#include "hdr/stdint_proxy.h"
+#undef __STDC_WANT_LIB_EXT1__
+#include "src/__support/common.h"
+#include "src/__support/constraint_handler.h"
+#include "src/__support/macros/config.h"
+#include "src/stdlib/qsort_util.h"
+
+#define ERRNO_T_FAIL 1
```
michaelrj-google: is this macro necessary? If it's planned to be used in several files we should probably move it to a common area. Additionally, it's preferred to use `const` or `constexpr` variables for constants for improved type safety.

libc/test/src/stdlib/qsort_s_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+  const int *b_int = static_cast<const int *>(b);
+  return *a_int - *b_int;
+}
+
+TEST_F(LlvmLibcQsortSConstraintTest, ArraySizeGreaterThanRsizeMax) {
+  int array[] = {1, 3, 2};
+  errno_t retval =
+      LIBC_NAMESPACE::qsort_s(array, RSIZE_MAX + 1, sizeof(int), compare, 0);
+  EXPECT_NE(retval, 0);
+  EXPECT_STREQ(buffer, "qsort_s: array_size cannot be greater than RSIZE_MAX");
```
michaelrj-google: nit: testing the exact string results will make these tests brittle if we change those strings later. Is there a way we can test the behavior but not this specific implementation?

michaelrj-google (commented):
Overall LGTM with a few comments

libc/src/stdlib/qsort_s.cpp
vhscampos: For now I will remove the macro and use the value inline.

libc/test/src/stdlib/qsort_s_test.cpp
vhscampos: Good point. Unfortunately I believe there is no specified way to differentiate among the different error conditions. Both the error message and the errno_t code are implementation-defined.

I plan to change this to a simple boolean "error detected" flag. I don't think we can do better than that.

libc/test/src/stdlib/QsortReentrantTest.h
michaelrj-google: nit: formatting

michaelrj-google (approved):
Overall LGTM

PR #198799, opened 2026-05-20
[flang][OpenMP] Add structure checks for DECLARE VARIANT

abidh:
This PR adds declare-variant structure checking. Following checks are added:

- Validate [base:]variant arguments (including implicit base for single-name form).
- Require exactly one MATCH clause; reject a second MATCH on the same directive.
- Reject duplicate (base, variant) across multiple declare variant directives.
- Reject clauses not allowed on declare variant.
- Apply shared context-selector checks to MATCH (reuse metadirective logic).
- Require constant user conditions in MATCH for declare variant (dynamic selectors deferred).

Refactor metadirective support:

- Extract CheckContextSelectorSpecification for reuse.
- Reject SCORE on trait sets that do not allow it (also affects metadirective).

Co-authored-by: Cursor <cursoragent@cursor.com>


flang/lib/Semantics/check-omp-declare-variant.cpp
```diff
@@ -0,0 +1,183 @@
...
+      const parser::ScalarExpr &scalarExpr{
+          std::get<parser::ScalarExpr>(property.u)};
+      auto maybeType{GetDynamicType(scalarExpr.thing.value())};
+      if (!maybeType || maybeType->category() != TypeCategory::Logical) {
+        continue;
+      }
+      if (const auto *expr{GetExpr(scalarExpr)}) {
+        if (!IsConstantExpr(*expr, &context_.foldingContext())) {
+          context_.Say(property.source,
+              "USER condition in the MATCH clause must be a constant expression"_err_en_US);
```
kparzysz: It doesn't have to be constant.  The replacement function is selected dynamically, the expression is evaluated at the call site.

Edit: I saw in your commit message that you're only implementing constant conditions for now.  Please change the text of the diagnostic to something like "run-time condition is not implemented yet", the current wording suggests that run-time condition is simply not allowed.

flang/lib/Semantics/check-omp-structure.h
```diff
@@ -127,6 +127,9 @@ class OmpStructureChecker : public OmpStructureCheckerBase {
 
   void Enter(const parser::OmpDeclareVariantDirective &);
   void Leave(const parser::OmpDeclareVariantDirective &);
+  void CheckOmpDeclareVariantDirective(
+      const parser::OmpDeclareVariantDirective &);
+  void CheckDeclareVariantUserConditions(const parser::OmpContextSelector &);
```
kparzysz: The function headers (except Enter/Leave) and grouped by the .cpp file that contains the implementations.

flang/lib/Semantics/check-omp-declare-variant.cpp
```diff
@@ -0,0 +1,183 @@
...
+          std::get<std::list<parser::OmpTraitProperty>>(maybeProps->t)};
+      if (properties.size() != 1) {
+        continue;
+      }
+      const parser::OmpTraitProperty &property{properties.front()};
+      const parser::ScalarExpr &scalarExpr{
+          std::get<parser::ScalarExpr>(property.u)};
+      auto maybeType{GetDynamicType(scalarExpr.thing.value())};
+      if (!maybeType || maybeType->category() != TypeCategory::Logical) {
+        continue;
```
kparzysz: This is a good candidate for an error.  The expression should have type (I think BOZ literals are examples of typeless expressions, but everything logical that we accept should not be typeless), and the type should be logical.

kparzysz:
This file is relatively short, does it need to be a separate file?  If the file name is the issue, you could rename check-omp-metadirective.cpp to check-omp-variant.cpp, since both fall into the "variant directives" category.

flang/lib/Semantics/check-omp-declare-variant.cpp
abidh: Thanks. Yes, I meant it as stop gap until I implement the dynamic condition support. I have updated the error message.

flang/lib/Semantics/check-omp-structure.h
abidh: Done.

abidh:
> This file is relatively short, does it need to be a separate file? If the file name is the issue, you could rename check-omp-metadirective.cpp to check-omp-variant.cpp, since both fall into the "variant directives" category.

I agree and once the user condition check is removed, it will be even smaller. I have moved the checks into omp-metadirective.cpp. I did not rename the file as it will make reviewing the changes difficult. Once this goes in, I can rename it in a separate PR.

flang/lib/Semantics/check-omp-declare-variant.cpp
abidh: Before reaching here, we should have passed through `CheckTraitCondition` which already emits already emits `CONDITION trait requires a single LOGICAL expression`. There is also a test for this error in declare-variant-match.f90. Generating error here seems redundant.

kparzysz (approved):

PR #198802, opened 2026-05-20
[VPlan] Transfer nuw from CanIV -> WideCanIV

artagnon:
An nuw on a CanonicalIV recipe should transfer to a WideCanonicalIV recipe directly.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlan.h
+++ b/llvm/lib/Transforms/Vectorize/VPlan.h
@@ -1140,7 +1140,8 @@ struct VPRecipeWithIRFlags : public VPSingleDefRecipe, public VPIRFlags {
            R->getVPRecipeID() == VPRecipeBase::VPReductionEVLSC ||
            R->getVPRecipeID() == VPRecipeBase::VPReplicateSC ||
            R->getVPRecipeID() == VPRecipeBase::VPVectorEndPointerSC ||
-           R->getVPRecipeID() == VPRecipeBase::VPVectorPointerSC;
+           R->getVPRecipeID() == VPRecipeBase::VPVectorPointerSC ||
+           R->getVPRecipeID() == VPRecipeBase::VPWidenCanonicalIVSC;
   }
 
   static inline bool classof(const VPUser *U) {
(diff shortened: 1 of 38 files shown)


llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -4425,6 +4425,12 @@ void VPExpandSCEVRecipe::printRecipe(raw_ostream &O, const Twine &Indent,
 }
 #endif
 
+VPWidenCanonicalIVRecipe::VPWidenCanonicalIVRecipe(VPRegionValue *CanonicalIV)
+    : VPRecipeWithIRFlags(
+          VPRecipeBase::VPWidenCanonicalIVSC, {CanonicalIV},
+          VPIRFlags::WrapFlagsTy(
+              CanonicalIV->getDefiningRegion()->hasCanonicalIVNUW(), false)) {}
+
```
madhur13490: You will also want to handle `clone()` method. The `clone()` method, would call the constructor and resurrect the flag from the region but some transforms may have dropped using `dropPosionGeneratingFlags()`. The change here is resurrecting the flag. 

I think `clone()` should call a new constructor which explicitly takes wrap flags and `clone()` should call the new constructor.

madhur13490:
You should also update `classof()` method in `VPRecipeWithIRFlags` as now WidenIV inherits from IRFlags.

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -3968,16 +3969,18 @@ class VPCurrentIterationPHIRecipe : public VPHeaderPHIRecipe {
 /// A Recipe for widening the canonical induction variable of the vector loop.
 /// First operand is the canonical IV recipe, a second step operand  (VF * Part)
 /// is added during unrolling.
-class VPWidenCanonicalIVRecipe : public VPSingleDefRecipe {
+class VPWidenCanonicalIVRecipe : public VPRecipeWithIRFlags {
```
fhahn, line 3972: would be good to update print() to also include the flags + test

fhahn (approved):
LGTM, thanks

PR #198807, opened 2026-05-20
[lldb][test] Make dependent-modules-nodupe-windows case-insensitive

charles-zablit:
Running the test in containers returns `KERNEL32.DLL` instead of `kernel32.dll`, which causes the test to fail. Both names point to the same file. Windows path comparison is case-insensitive. The test is asserting an OS-level behavior, not a casing convention, so it should not be case-sensitive.


Nerixyz (approved):
In your PR description:
```diff
-instead of `kernell32.dll`,
+instead of `kernel32.dll`,
```

Issue #150515, opened 2025-07-24
[GlobalISel] Port known-bit and known-sign-bits for all operations


davemgreen:
gI was going through the SDAG version of computeKnownBits and computeSignBits to see what was missing from GISel.  This is a list of all the nodes that SDAG tries to handle and their current status is GISel. Some nodes have no equivalent in GISel. Missing nodes will presumably need to be ported over, under the assumption that they will eventually be useful somewhere. Some of the examples like https://github.com/llvm/llvm-project/pull/141933 and https://github.com/llvm/llvm-project/pull/142355 make a good example of what kind of thing needs to be added and how tests can be written.

|SDAG computeKnownBits | Status |
| --- | --- |
| MERGE_VALUES| Not relevant |
| SPLAT_VECTOR | [Done](https://github.com/llvm/llvm-project/pull/139501) |
| SPLAT_VECTOR_PARTS| Not relevant |
| STEP_VECTOR | #190598 |
| BUILD_VECTOR | Done |
| VECTOR_SHUFFLE | [Done](https://github.com/llvm/llvm-project/pull/139501) |
| VSCALE | #216025  |
| CONCAT_VECTORS | [Done](https://github.com/llvm/llvm-project/pull/141933) |
| INSERT_SUBVECTOR | #215048 |
| EXTRACT_SUBVECTOR | #214533 |
| SCALAR_TO_VECTOR| Not relevant |
| BITCAST | #216287  |
| AND | Done |
| OR | Done |
| XOR | Done |
| MUL | Done |
| MULHU | #158445 |
| MULHS | #158445 |
| ABDU | #186822 |
| ABDS | #186822 |
| UMUL_LOHI| Not relevant |
| SMUL_LOHI| Not relevant |
| AVGFLOORU | #209171 |
| AVGCEILU | #209171 |
| AVGFLOORS | #209171 |
| AVGCEILS | #209171 |
| SELECT | Done |
| VSELECT| Not relevant |
| SELECT_CC| Not relevant |
| SETCC | Done |
| SETCCCARRY| Not relevant |
| STRICT_FSETCC| Not relevant |
| STRICT_FSETCCS| Not relevant |
| SHL | Done |
| SRL | Done |
| SRA | Done |
| ROTL | #166365 |
| ROTR | #166365 |
| FSHL | #191260 |
| FSHR | #191260  |
| SHL_PARTS| Not relevant |
| SRA_PARTS| Not relevant |
| SRL_PARTS| Not relevant |
[...]

PR #198808, opened 2026-05-20
[GlobalISel] Port over `simplifyDemandedBits` to GlobalISel

lialan:
## Summary

* This patch  ports `SimplifyDemandedBits` to GlobalISel as a `CombinerHelper` walk. 
* It tries to structure it after InstCombine's `SimplifyDemandedBits` and SelectionDAG's `TargetLowering::SimplifyDemandedBits`. 
* GISel's `SimplifyDemandedBits` interface is kept the same as SDAG.
* As a first step, added `and`, `or`, `shl`, `shr`(constant only) bit simplifying mechanisms.
* MIR tests and unit tests are included in this PR.
* Tracks #150515 

## Interface Comparison

Here is the attempt to keep both APIs on GISel and SDAG/InstCombine's as similar as possible.

| Aspect | InstCombine | SDAG | This PR |
|---|---|---|---|
| Entry point | `SimplifyDemandedBits(I, OpNo, DemandedMask, Known, Depth)` | `SimplifyDemandedBits(Op, DemandedBits, DemandedElts, Known, TLO, Depth)` | `simplifyDemandedBits(MI, OpNo, DemandedBits, Known, Depth)` |
| Known bits | byproduct of the simplify walk | byproduct of the simplify walk | byproduct of the simplify walk |
| Unknown-opcode fallback | `computeKnownBits` | `computeKnownBits` | `getKnownBits` |
| Multi-use defs | `SimplifyMultipleUseDemandedBits` (look-through, reroute one use) | demand relaxed to all bits + separate `SimplifyMultipleUseDemandedBits` | same: pure `simplifyMultipleUseDemandedBits` look-through, then demand relaxed to all bits |
| `computeKnownBits` API | unchanged | unchanged | unchanged |


llvm/include/llvm/CodeGen/GlobalISel/CombinerHelper.h
```diff
@@ -1083,6 +1084,22 @@ class CombinerHelper {
...
+  /// Demanded-bits back-propagation driver. Attempts to rewrite \p MI under
+  /// the assumption that only the bits set in \p Mask.Bits and the elements
+  /// set in \p Mask.Elts of its result are consumed. Returns true if \p MI
+  /// (or one of its operand-producing instructions) was mutated or replaced.
+  ///
+  /// Mirrors SelectionDAG's TargetLowering::SimplifyDemandedBits but operates
+  /// directly on MachineInstrs through CombinerHelper. Target instructions
+  /// are dispatched to TargetLowering::simplifyDemandedBitsForTargetInstr.
+  bool trySimplifyDemandedBits(MachineInstr &MI,
+                               const GISelDemandedMask &Mask) const;
```
nhaehnle: Drop the part about SelectionDAG. This needs to be able to stand on its own.

This should have an `OperandNo` argument to support using it on instructions with multiple defs.

llvm/include/llvm/CodeGen/GlobalISel/CombinerHelper.h
```diff
@@ -1083,6 +1084,22 @@ class CombinerHelper {
...
+
+  // Match (G_TRUNC (G_LSHR/G_ASHR X, K-const)) when X's bits beyond DstBW
+  // (zero for LSHR, sign-bit replicated for ASHR) are provably idle. Rewrites
+  // to (G_LSHR/G_ASHR (G_TRUNC X), K-trunc), eliminating the outer trunc.
+  bool matchNarrowTruncShrConst(MachineInstr &MI, BuildFnTy &MatchInfo) const;
```
nhaehnle: This doesn't actually rewrite, does it?

nhaehnle (commented):
What's the justification of GISelDemandedMask?

lialan (commented):
(withdrawn)

llvm/include/llvm/CodeGen/GlobalISel/CombinerHelper.h
lialan: Updated

lialan: Restructured: `GISelDemandedMask` is removed entirely. The signature now matches InstCombine's `SimplifyDemandedBits(I, OpNo, DemandedMask, Known, Depth)` — a bits-only `APInt` on a `CombinerHelper` walk — and `GISelValueTracking` is untouched relative to main. The struct existed to thread a demand mask through `computeKnownBits`, which neither SDAG nor InstCombine does; with the analysis living inside the simplifier's own per-opcode switch (as in both of those), there is nothing left for it to carry. PR description has a parity table with the remaining (two, minimal) deviations.

llvm/include/llvm/CodeGen/GlobalISel/CombinerHelper.h
lialan: updated!

lialan:
@nhaehnle care to review it again?

PR #198809, opened 2026-05-20
[AArch64][GlobalISel] Add support for pmul intrinsic

JoshdRod:
Previously, pmul couldn't lower as there were neither any GlobalISel patterns for the intrinsic, nor a GlobalISel node to allow use of the SelectionDAG patterns.
Create a G_PMUL node linked to the SelectionDAG node 'clmul' (CarryLess MULtiplication). Now the ISel stage of GlobalISel has suitable patterns to lower the intrinsic.

### DIFF
--- a/llvm/docs/GlobalISel/GenericOpcode.rst
+++ b/llvm/docs/GlobalISel/GenericOpcode.rst
@@ -341,6 +341,15 @@ These each perform their respective integer arithmetic on a scalar.
 
 The above example adds %src1 to %src0 and stores the result in %dst.
 
+G_CLMUL
+^^^^^^
+
+Perform integer carry-less multiplication with saturation.
+
+.. code-block:: none
+
+  %dst:_(s32) = G_CLMUL %src_0:_(s32), %src1:_(32)
+
 G_SDIVREM, G_UDIVREM
 ^^^^^^^^^^^^^^^^^^^^
 
--- a/llvm/docs/GlobalISel/GenericOpcode.rst
+++ b/llvm/docs/GlobalISel/GenericOpcode.rst
@@ -342,9 +342,9 @@ These each perform their respective integer arithmetic on a scalar.
 The above example adds %src1 to %src0 and stores the result in %dst.
 
 G_CLMUL
-^^^^^^
+^^^^^^^
 
-Perform integer carry-less multiplication with saturation.
+Perform integer carry-less multiplication.
 
 .. code-block:: none
 
--- a/llvm/include/llvm/Support/TargetOpcodes.def
+++ b/llvm/include/llvm/Support/TargetOpcodes.def
@@ -271,9 +271,6 @@ HANDLE_TARGET_OPCODE(G_SUB)
 // Generic multiply instruction.
 HANDLE_TARGET_OPCODE(G_MUL)
 
-// Generic carry-less multiply instruction.
-HANDLE_TARGET_OPCODE(G_CLMUL)
-
 // Generic signed division instruction.
 HANDLE_TARGET_OPCODE(G_SDIV)
 
@@ -861,6 +858,9 @@ HANDLE_TARGET_OPCODE(G_BSWAP)
 
 /// Generic bit reverse.
 HANDLE_TARGET_OPCODE(G_BITREVERSE)
+
+// Generic carry-less multiply instruction.
+HANDLE_TARGET_OPCODE(G_CLMUL)
 
 /// Floating point ceil.
 HANDLE_TARGET_OPCODE(G_FCEIL)
--- a/llvm/include/llvm/Support/TargetOpcodes.def
+++ b/llvm/include/llvm/Support/TargetOpcodes.def
@@ -271,6 +271,9 @@ HANDLE_TARGET_OPCODE(G_SUB)
 // Generic multiply instruction.
 HANDLE_TARGET_OPCODE(G_MUL)
 
+// Generic carry-less multiply instruction.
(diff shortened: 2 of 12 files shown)


llvm/lib/Target/AArch64/AArch64InstrGISel.td
```diff
@@ -216,6 +216,12 @@ def G_SMULL : AArch64GenericInstruction {
   let hasSideEffects = 0;
 }
 
+def G_CLMUL : AArch64GenericInstruction {
```
davemgreen: This can be a target independent instruction

arsenm: Also should get the usual legalization treatment

llvm/docs/GlobalISel/GenericOpcode.rst
```diff
@@ -341,6 +341,15 @@ These each perform their respective integer arithmetic on a scalar.
 
 The above example adds %src1 to %src0 and stores the result in %dst.
 
+G_CLMUL
+^^^^^^
+
+Perform integer carry-less multiplication with saturation.
```
davemgreen: Remove "with saturation", unless it means something that I am unaware of.

llvm/include/llvm/Support/TargetOpcodes.def
```diff
@@ -271,6 +271,9 @@ HANDLE_TARGET_OPCODE(G_SUB)
 // Generic multiply instruction.
 HANDLE_TARGET_OPCODE(G_MUL)
 
+// Generic carry-less multiply instruction.
+HANDLE_TARGET_OPCODE(G_CLMUL)
```
davemgreen: I would put this at the bottom or next to G_BITREVERSE

llvm/include/llvm/Target/GenericOpcodes.td
```diff
@@ -328,6 +328,14 @@ def G_MUL : GenericInstruction {
   let isCommutable = true;
 }
 
+// Generic carry-less multiplication
+def G_CLMUL : GenericInstruction {
+  let OutOperandList = (outs type0:$dst);
+  let InOperandList = (ins type0:$src1, type0:$src2);
+  let hasSideEffects = false;
+  //let isCommutable = true;
```
davemgreen: Can uncomment this, it should be commutative. Again I would put it further down in the list.

llvm/lib/Target/AArch64/GISel/AArch64LegalizerInfo.cpp
```diff
@@ -207,7 +207,7 @@ AArch64LegalizerInfo::AArch64LegalizerInfo(const AArch64Subtarget &ST)
       .scalarizeIf(scalarOrEltWiderThan(0, 64), 0)
       .moreElementsToNextPow2(0);
 
-  getActionDefinitionsBuilder(G_MUL)
+  getActionDefinitionsBuilder({G_MUL, G_CLMUL})
```
davemgreen: Remove from here, add a new ActionDefinitionsBuilder that just makes it legalFor the correct types for the moment.

llvm/include/llvm/Support/TargetOpcodes.def
JoshdRod: Why specifically there?

llvm/lib/Target/AArch64/GISel/AArch64LegalizerInfo.cpp
JoshdRod: So, no resizing rules? Why not?
Surely the scalar and vector resizing rules are the same for G_CLMUL as they are for G_MUL

llvm/include/llvm/Support/TargetOpcodes.def
JoshdRod: Done

llvm/lib/Target/AArch64/GISel/AArch64LegalizerInfo.cpp
davemgreen: We can add them later when we add full legalization (it is a bigger task to add it all in). I would expect clmul to be quite different to mul as there are only certain operations available (pmul, pmull), and they are often only available with certain architecture features.

llvm/include/llvm/Support/TargetOpcodes.def
davemgreen: I would think of CLMUL less like a standard arithmetic operation (add, sub, mul, div), and more like a slightly oddball bitwise operation. But the order of a lot of these lists seems inconsistent at best.

llvm/lib/Target/AArch64/GISel/AArch64LegalizerInfo.cpp
```diff
@@ -424,6 +424,10 @@ AArch64LegalizerInfo::AArch64LegalizerInfo(const AArch64Subtarget &ST)
       .moreElementsToNextPow2(0)
       .lower();
 
+  getActionDefinitionsBuilder(G_CLMUL)
+      .legalFor({i32, i64, v8i8, v16i8, v4i16, v8i16, v2i32, v4i32, v2i64})
```
davemgreen:
```
getActionDefinitionsBuilder(G_CLMUL)
      .legalFor({v8i8, v16i8});
```
The other conditions are likely more complex and relate to which arch features are available. I believe this should be enough for the intrinsics in the tests.

davemgreen (approved):
Thanks. LGTM

PR #198810, opened 2026-05-20
[Offload] Make profiling support opt-in on events

jhuber6:
Summary:
Profiling requires getting time stamp information, which required
actually materializing barriers on the event. This caused performance
regressions in some OpenMP applications. Instead, we want to make this
opt in. This just adds a new argument to `olCreateEvent` which takes
some flags. Right now this is just none and profiling. Pretty mechanical
beyond that. Should revert this to the old behavior for OpenMP.

### DIFF
--- a/offload/liboffload/src/OffloadImpl.cpp
+++ b/offload/liboffload/src/OffloadImpl.cpp
@@ -172,9 +172,9 @@ std::atomic<size_t> ol_queue_impl_t::IdCounter(0);
 
 struct ol_event_impl_t {
   ol_event_impl_t(void *EventInfo, ol_device_handle_t Device,
-                  ol_queue_handle_t Queue)
-      : EventInfo(EventInfo), Device(Device), QueueId(Queue->Id), Queue(Queue) {
-  }
+                  ol_queue_handle_t Queue, bool ProfilingEnabled)
+      : EventInfo(EventInfo), Device(Device), QueueId(Queue->Id), Queue(Queue),
+        ProfilingEnabled(ProfilingEnabled) {}
   // Opaque backend-specific event state. This is expected to be non-null for
   // backends that materialize real events.
   void *EventInfo;
@@ -184,6 +184,7 @@ struct ol_event_impl_t {
   // It is provided only to implement OL_EVENT_INFO_QUEUE. Use QueueId to check
   // for queue equality instead.
   ol_queue_handle_t Queue;
+  bool ProfilingEnabled;
 };
 
 struct ol_program_impl_t {
@@ -859,6 +860,12 @@ Error olSyncEvent_impl(ol_event_handle_t Event) {
 Error olGetEventElapsedTime_impl(ol_event_handle_t StartEvent,
                                  ol_event_handle_t EndEvent,
                                  float *ElapsedTime) {
+  if (!StartEvent->ProfilingEnabled || !EndEvent->ProfilingEnabled)
+    return createOffloadError(
+        ErrorCode::INVALID_ARGUMENT,
+        "olGetEventElapsedTime requires both events to be created with "
+        "OL_EVENT_FLAGS_ENABLE_PROFILING");
+
   if (StartEvent->Device != EndEvent->Device)
     return createOffloadError(
         ErrorCode::INVALID_DEVICE,
@@ -875,7 +882,8 @@ Error olGetEventElapsedTime_impl(ol_event_handle_t StartEvent,
 
 Error olDestroyEvent_impl(ol_event_handle_t Event) {
   if (Event->EventInfo)
-    if (auto Res = Event->Device->Device->destroyEvent(Event->EventInfo))
+    if (auto Res = Event->Device->Device->destroyEvent(Event->EventInfo,
+                                                       Event->ProfilingEnabled))
       return Res;
 
   return olDestroy(Event);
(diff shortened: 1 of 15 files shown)


offload/liboffload/src/OffloadImpl.cpp
```diff
@@ -172,9 +172,9 @@ std::atomic<size_t> ol_queue_impl_t::IdCounter(0);
 
 struct ol_event_impl_t {
   ol_event_impl_t(void *EventInfo, ol_device_handle_t Device,
-                  ol_queue_handle_t Queue)
-      : EventInfo(EventInfo), Device(Device), QueueId(Queue->Id), Queue(Queue) {
-  }
+                  ol_queue_handle_t Queue, bool ProfilingEnabled)
+      : EventInfo(EventInfo), Device(Device), QueueId(Queue->Id), Queue(Queue),
+        ProfilingEnabled(ProfilingEnabled) {}
```
shiltian, line 177: I thought the flag is supposed to be global thing?

ronlieb (approved):

ronlieb:
verified performance recovered after applying PR.

offload/liboffload/src/OffloadImpl.cpp
jhuber6: Nah, it's a per-event thing, just like in CUDA and HIP you can opt-out per event. We're choosing opt-in here (which I like better.)

jhuber6: Nah, it's a per-event thing, just like in CUDA and HIP you can opt-out, this is opt-in.

offload/plugins-nextgen/amdgpu/src/rtl.cpp
```diff
@@ -1805,11 +1807,31 @@ struct AMDGPUEventTy {
...
+                                  bool EnableProfiling) {
   if (Queue == nullptr)
     return Plugin::error(ErrorCode::INVALID_NULL_POINTER,
                          "target queue was nullptr");
 
+  if (!EnableProfiling) {
+    std::lock_guard<std::mutex> StreamLock(Mutex);
+
+    if (size() > 0) {
+      Event.RecordedSyncCycle = SyncCycle;
```
kevinsala: Please recover the comment:

```suggestion
    if (size() > 0) {
      // Record the synchronize identifier (to detect stale recordings) and
      // the last valid stream's operation.
      Event.RecordedSyncCycle = SyncCycle;
```

offload/plugins-nextgen/amdgpu/src/rtl.cpp
```diff
@@ -1805,11 +1807,31 @@ struct AMDGPUEventTy {
...
+      Event.RecordedSlot = last();
```
kevinsala: Recover asserts:

```suggestion
      Event.RecordedSlot = last();
      
      assert(Event.RecordedSyncCycle >= 0 && "Invalid recorded sync cycle");
      assert(Event.RecordedSlot >= 0 && "Invalid recorded slot");
```

offload/plugins-nextgen/amdgpu/src/rtl.cpp
```diff
@@ -1805,11 +1807,31 @@ struct AMDGPUEventTy {
...
+    } else {
```
kevinsala: Recover comment:

```suggestion
    } else {
      // The stream is empty, everything already completed, record nothing.
```

offload/plugins-nextgen/amdgpu/src/rtl.cpp
```diff
@@ -1805,11 +1807,31 @@ struct AMDGPUEventTy {
...
+      Event.RecordedSyncCycle = -1;
+      Event.RecordedSlot = -1;
+    }
+
+    // Return the unused reusable signal back to the manager.
+    if (ReusedSignal) {
+      if (auto Err = SignalManager.returnResource(ReusedSignal))
+        return Err;
+    }
```
kevinsala:
```suggestion
    // Return the unused reusable signal back to the manager.
    if (ReusedSignal)
      return SignalManager.returnResource(ReusedSignal);
```

offload/plugins-nextgen/amdgpu/src/rtl.cpp
```diff
@@ -1805,11 +1807,31 @@ struct AMDGPUEventTy {
...
 Error AMDGPUStreamTy::recordEvent(AMDGPUEventTy &Event,
-                                  AMDGPUSignalTy *ReusedSignal) {
+                                  AMDGPUSignalTy *ReusedSignal,
+                                  bool EnableProfiling) {
   if (Queue == nullptr)
     return Plugin::error(ErrorCode::INVALID_NULL_POINTER,
                          "target queue was nullptr");
 
+  if (!EnableProfiling) {
+    std::lock_guard<std::mutex> StreamLock(Mutex);
```
kevinsala: Can you put these two paths into different private member functions within the AMDGPUStreamTy?

kevinsala (approved):
Apart from the comments, LGTM.

offload/plugins-nextgen/amdgpu/src/rtl.cpp
kevinsala: Looks good now. Thanks!

PR #198813, opened 2026-05-20
[LV] Fix dead early exit in test. NFC

lukel97:
The early exit in this test is dead because the predicated block's condition (< 0) means the exit cond (== 1) can't be met.


fhahn (approved):
LGTM, thanks

artagnon (approved):
LGTM, thanks!

PR #198814, opened 2026-05-20
[docker][windows] install make in Dockerfile

charles-zablit:
The lldb-api test suite builds each test's inferior via GNU make. The current Windows CI container (`ghcr.io/llvm/ci-windows-2022:latest`) does not install make, so LLDB's CMake configure step ends up with `LLDB_DEFAULT_TEST_MAKE-NOTFOUND` and every lldb-api test reports `UNRESOLVED`.
    
This patch adds a step in the Windows Dockerfile to install make with choco.


boomanaiden154 (approved):
LGTM.

Nerixyz (approved):

PR #198815, opened 2026-05-20
[libc][clang] Use __nvvm_nanosleep builtin over inline assembly

AlexMaclean:


jhuber6 (commented):
I used assembly because `nanosleep` requires PTX features which were dependent on compile flags. I don't remember if recent changes improved that. Can you verify that a `clang --target=nvptx64-nvidia-cuda -nogpulib -flto -c nanosleep.c` works without *any* CUDA installation detected?

michaelrj-google:
Please add a description of the change to the commit message.

clang/lib/Headers/nvptxintrin.h
```diff
@@ -199,7 +199,7 @@ _DEFAULT_FN_ATTRS [[noreturn]] static __inline__ void __gpu_exit(void) {
 // Suspend the thread briefly to assist the scheduler during busy loops.
 _DEFAULT_FN_ATTRS static __inline__ void __gpu_thread_suspend(void) {
   if (__nvvm_reflect("__CUDA_ARCH") >= 700)
-    asm("nanosleep.u32 64;" ::: "memory");
+    __nvvm_nanosleep(64);
```
Artem-B, line 202: This will fail on older GPUs:

```
def INT_NVVM_NANOSLEEP_I : BasicNVPTXInst<(outs), (ins i32imm:$i), "nanosleep.u32",
                             [(int_nvvm_nanosleep imm:$i)]>,
        Requires<[hasPTX<63>, hasSM<70>]>;
```

https://godbolt.org/z/4cr3z36GW

AlexMaclean:
@Artem-B / @jhuber6 You're both correct. This change won't work as written due to the PTX and SM version restrictions clang and NVPTX enforce on the builtin and intrinsic. It still seems like a bit of a hack to use inline assembly to circumvent these checks. Do either of you have ideas for how this could be made to work? maybe with `__has_builtin` instead of `__nvvm_reflect`?

jhuber6:
> maybe with `__has_builtin` instead of `__nvvm_reflect`?

That would have the effect of disabling it. The nvvm_reflect thing is a big hack and there's been a lot of discussion around it in the past, but basically it turns a compile time check into a runtime one, so one IR file can 'do the optimal thing' on a sufficiently new GPU. AMDGCN recently added some `__builtin_amdgcn_is_invocable` or something which does something similar, but it's only slightly less hacky.

Artem-B:
`__nvvm_reflect()` is a poor man's IR-level preprocessor. Using it with inline asm is probably the least bad choice if we need to emit IR with gpu-specific parts in it. That relies on the false branch to be physically removed before we get to the back-end which may not be able to lower IR intended for a different GPU. The only sound approach I can think of is to not use `_nvvm_reflect()` at all and stick with proper per-GPU IR without trying to make IR 'generic' while carrying very GPU-specific bits in it -- something the IR was never intended for. But given that `__nvvm_reflect`does exist, I can see the attractiveness of using it to tweak just small parts of otherwise generic code. It mostly works, but with caveats.

PR #122459, opened 2025-01-10
DAG: Fix vector bin op scalarize defining a partially undef vector

arsenm:
This avoids some of the pending regressions after AMDGPU implements
isExtractVecEltCheap.

In a case like shl <value, undef>, splat k, because the second operand
was fully defined, we would fall through and use the splat value for the
first operand, losing the undef high bits. This would result in an additional
instruction to handle the high bits. Add some reduced testcases for different
opcodes for one of the regressions.

PR #198825, opened 2026-05-20
[AMDGPU] Add regression test for extract of vector binop scalarization

carlobertolli:
Test that extracting both lanes from a binop of two build_vectors sharing a variable operand at different lane positions correctly folds per-lane constants.

Assisted-by: Cursor (Claude)


carlobertolli:
This bug was reported downstream on ROCM 7.2.3 but fixed by current main branch, nearest PR fix seems to be
https://github.com/llvm/llvm-project/pull/122459

shiltian (approved):

aobolensk (approved):

PR #198828, opened 2026-05-20
[DWARFLinker] Preserve module forward decls in parallel linker

JDevlieghere:
The parallel linker was dropping forward-declared types nested under DW_TAG_module skeleton CUs (clang -gmodules input). The classic linker keeps them when no definition exists and replaces them with the canonical definition when one does. collectRootsToKeep's default branch silently fell through, so the forward decl never made it to the worklist.

Route these through the type pool with MarkTypeEntryRec. Recursive marking keeps children of templated forward decls (e.g. DW_TAG_template_type_parameter under Foo<T> [decl]) intact, matching the classic linker. The existing decl-vs-def race resolution in allocateTypeDie and getFinalDie handles both ends without new synchronization.

- No definition anywhere. DeclarationDie is allocated and emitted.
- Definition in some other CU. allocateTypeDie sets Die via the body's spinlock. At emission, getFinalDie returns Die over DeclarationDie. Both clone orderings yield the same output.

For non-ODR languages getFinalPlacementForEntry forces PlainDwarf, so the forward decl stays in place under its module instead of moving to the artificial type unit.

drop-module-fwd-decl.test covers the dedup path on a C++ .ll input. submodules.m gains a --linker parallel RUN that exercises the preservation path.


llvm/test/tools/dsymutil/X86/drop-module-fwd-decl.test
```diff
@@ -0,0 +1,27 @@
+# A forward declaration of `X` nested under DW_TAG_module M, plus a real
+# definition of `X` in another CU. The parallel linker routes both
+# through the type pool, so allocateTypeDie's def-wins resolution plus
+# getFinalDie emit only the canonical definition under the module.
+# Regression test for the dedup of module forward declarations.
```
alx32: This test doesn't seem to fail without the code fix. Was that the intent ? "Regression test" would imply that it would fail.

llvm/lib/DWARFLinker/Parallel/DependencyTracker.cpp
```diff
@@ -214,7 +214,23 @@ void DependencyTracker::collectRootsToKeep(
...
+      // the definition and drops this declaration at emission time. For
+      // non-ODR languages getFinalPlacementForEntry forces PlainDwarf,
+      // so the forward decl is kept in place under its module.
+      if (Entry.DieEntry->getTag() == dwarf::DW_TAG_module &&
+          dwarf::isType(CurChild->getTag()) &&
+          dwarf::toUnsigned(Entry.CU->find(CurChild, dwarf::DW_AT_declaration),
+                            0)) {
+        ChildInfo.setIsModuleForwardDeclCandidate();
+        addActionToRootEntriesWorkList(
+            LiveRootWorklistActionTy::MarkSingleTypeEntry, ChildEntry,
```
alx32: Should this be `MarkTypeEntryRec` instead? A templated forward declaration under a module can carry a `DW_TAG_template_type_parameter` child ([example input](https://gist.github.com/alx32/a1c90d8622bdf65ae3124500b6e3cad3)), and since `MarkSingleTypeEntry` [returns before processing children](https://github.com/llvm/llvm-project/blob/b7531c04f000c8a1a0ebdf7e5416d71340377709/llvm/lib/DWARFLinker/Parallel/DependencyTracker.cpp#L508-L510) in `markDIEEntryAsKeptRec`, the no definition case ends up emitting `Foo<T>` with `DW_AT_declaration` but loses `T`. Classic linker keeps it on the same input.

JDevlieghere:
friendly ping

alx32:
@JDevlieghere - Are the two comments above addressed / not actual issues ? - they're still showing as Pending

JDevlieghere:
> @JDevlieghere - Are the two comments above addressed / not actual issues ? - they're still showing as Pending 

I don't see any comments. Could it be that they show up as pending because you haven't submitted the review yet?

alx32:
> Could it be that they show up as pending because you haven't submitted the review yet?

Oh no - that was exactly the case 😅

llvm/test/tools/dsymutil/X86/drop-module-fwd-decl.test
JDevlieghere: `submodules.m‎` is the regression test, which fails without this patch. This just checks the happy path, which was already working, but I realized didn't have test coverage yet.

llvm/lib/DWARFLinker/Parallel/DependencyTracker.cpp
JDevlieghere: Good catch. The failure mode was even broader than the lost template parameter. For the no-definition case, the whole subtree disappeared and left only a dangling DW_AT_import. MarkSingleTypeEntry was wrong even for the non-templated forward decls. The original test in this PR just happened to work because the definition in the other CU pulls everything through the type pool.

llvm/lib/DWARFLinker/Parallel/DWARFLinkerCompileUnit.h
```diff
@@ -260,6 +260,12 @@ class alignas(8) CompileUnit : public DwarfUnit {
     /// Track liveness for the DIE.
     SINGLE_FLAG_METHODS_SET(HasAnAddress, 0x800)
 
+    /// DIE is a forward-declared type child of a DW_TAG_module. Kept in
+    /// PlainDwarf placement so the module retains its name listing. The
+    /// cloner drops it when TypePool has a non-declaration DIE for the
+    /// same synthetic name (i.e. a real definition exists elsewhere).
+    SINGLE_FLAG_METHODS_SET(IsModuleForwardDeclCandidate, 0x1000)
```
alx32: Is this (`IsModuleForwardDeclCandidate`) actually used currenty or that is OK for it to be added but not used ?

llvm/test/tools/dsymutil/X86/drop-module-fwd-decl.test
alx32: I see - is it accurate to say `Regression test for the dedup of module forward declarations` - should we remove if not ?

llvm/lib/DWARFLinker/Parallel/DependencyTracker.cpp
alx32: Is PR description saying `Route these through the type pool with MarkSingleTypeEntry. ` still accurate after latest changes ?

llvm/lib/DWARFLinker/Parallel/DependencyTracker.cpp
```diff
@@ -214,7 +214,23 @@ void DependencyTracker::collectRootsToKeep(
...
-      // Nothing to do.
+      // A forward-declared type nested in a DW_TAG_module is the module's
+      // record that the name exists, even when no full definition has been
+      // emitted. Route it through the type pool: when another CU emits a
+      // real definition for the same synthetic name, the existing
+      // decl-vs-def race resolution in allocateTypeDie + getFinalDie keeps
+      // the definition and drops this declaration at emission time. For
+      // non-ODR languages getFinalPlacementForEntry forces PlainDwarf,
+      // so the forward decl is kept in place under its module.
+      if (Entry.DieEntry->getTag() == dwarf::DW_TAG_module &&
```
alx32, line 225: Non-blocking, pre-existing (identical output with and without this patch): the predicate keys off the **immediate** parent being `DW_TAG_module`, so namespace-nested forward decls are dropped even though classic keeps them — e.g. `module M → namespace N → class Foo<T> [decl]`. Repro (classic passes, parallel drops `N`/`Foo<T>`): https://gist.github.com/alx32/555cdaf9cc2f44f5d86b26de57df9ce9

Possible follow-up: key off module *scope* (`getIsInMouduleScope()`) rather than immediate parentage.

alx32 (approved):

llvm/lib/DWARFLinker/Parallel/DependencyTracker.cpp
JDevlieghere: Ack, will follow up on this one.

PR #198832, opened 2026-05-20
[Hexagon] Translate IEEE HVX intrinsics to QFloat on v79+

fhossein-quic:
On Hexagon v79 and later, the IEEE-floating-point HVX
instruction encodings (sf/hf operands) are no longer
present in the architecture.  Code that uses the IEEE HVX
intrinsics still needs to compile for those targets, so
this change implicitly lowers the intrinsic calls to
equivalent QFloat (qf16 / qf32) sequences during DAG-to-DAG
instruction selection.  v75 and earlier continue to emit
the original IEEE encodings unchanged.

The translation covers 22 intrinsics: arithmetic kernels
(vadd / vsub / vmpy / vmpy-acc) for both hf-hf and sf-hf
operand pairs, sign-bit manipulation (vabs / vfneg), the
non-IEEE min/max variants, the cross-domain conversions
(vcvt_hf_h, vcvt_h_hf, vcvt_sf_hf), and vassign_fp.

9 HVX IEEE intrinsics are not translated by this patch
and will be added incrementally: the byte/ubyte/uhalf
conversion variants (vcvt_b_hf, vcvt_hf_b, vcvt_ub_hf,
vcvt_hf_ub, vcvt_uh_hf, vcvt_hf_uh), vcvt_h_hf (v79
requires a runtime call), vcvt_hf_sf, and vdmpy_sf_hf.
isIEEEHVXIntrinsic excludes them, so they fall through
to the normal SelectCode path rather than hitting
llvm_unreachable (which would be UB in release builds).

Co-authored-by: Sumanth Gundapaneni <sgundapa@quicinc.com>
Co-authored-by: Santanu Das <quic_santdas@quicinc.com>


fhossein-quic:
Looks like the buildbot failure is unrelated to this patch (failing in AArch64). 
I’ll keep an eye on it, but please let me know if you think it could be connected.

llvm/lib/Target/Hexagon/HexagonISelDAGToDAGHVX.cpp
```diff
@@ -3009,3 +3009,453 @@ void HexagonDAGToDAGISel::SelectHVXDualOutput(SDNode *N) {
...
+         Opcode == Intrinsic::hexagon_V6_vfmin_sf_128B ||
+         Opcode == Intrinsic::hexagon_V6_vfmax_hf_128B ||
+         Opcode == Intrinsic::hexagon_V6_vfmax_sf_128B ||
+         Opcode == Intrinsic::hexagon_V6_vfneg_hf_128B ||
+         Opcode == Intrinsic::hexagon_V6_vfneg_sf_128B ||
+         Opcode == Intrinsic::hexagon_V6_vmpy_sf_hf_acc_128B ||
+         Opcode == Intrinsic::hexagon_V6_vmpy_hf_hf_acc_128B ||
+         Opcode == Intrinsic::hexagon_V6_vmpy_sf_hf_128B ||
+         Opcode == Intrinsic::hexagon_V6_vmpy_hf_hf_128B ||
+         Opcode == Intrinsic::hexagon_V6_vdmpy_sf_hf_128B ||
```
aankit-ca: I counted 31 intrinsics in this list and 23 cases in the translation function. There are 8 here that don't have the translation. Are they expected to be handled in a separate patch? I feel it would be better if we handle them within the same patch otherwise users don't know which intrinsics have the translation can be used on v79+ and which ones are not supported

aankit-ca: If the remaining 8 are not supported we should atleast change the llvm_unreachable to have a more descriptive error message

llvm/lib/Target/Hexagon/HexagonISelDAGToDAG.cpp
```diff
@@ -676,8 +676,17 @@ void HexagonDAGToDAGISel::SelectIntrinsicWChain(SDNode *N) {
 }
 
 void HexagonDAGToDAGISel::SelectIntrinsicWOChain(SDNode *N) {
-  unsigned IID = N->getConstantOperandVal(0);
+  unsigned IID = cast<ConstantSDNode>(N->getOperand(0))->getZExtValue();
   unsigned Bits;
+
+  // On v79 and above, IEEE HVX instructions are no longer present.
+  // The related intrinsics must be translated to QFloat implicitly in order
+  // to provide backward compatability.
```
androm3da: typo compatibility

llvm/test/CodeGen/Hexagon/autohvx/ieeetoqfloat.ll
```diff
@@ -0,0 +1,471 @@
...
+;v1:0.sf = vadd(v0.hf,v1.hf) is translated to
+;r2 = #15360; v2.h = vsplat(r2);
+;v5:4.qf32 = vmpy(v1.hf,v2.hf); v31:30.qf32 = vmpy(v0.hf,v2.hf)
+;v4.qf32 = vadd(v30.qf32,v4.qf32); v3.qf32 = vadd(v31.qf32,v5.qf32)
+;v0.sf = v4.qf32; v1.sf = v3.qf32
+;Widen the hf operands to sf by doing a widening multiply with 1.0f
+;and perform the add.
+
+; CHECK-LABEL: sfaddhfhf:
+; CHECK-NOT: v1:0.sf = vadd(v0.hf,v1.hf)
```
androm3da, line 112: Should this have  `75: v1:0.sf = vadd(v0.hf,v1.hf)`?  (same question for `sfsubhfhf`)

llvm/lib/Target/Hexagon/HexagonISelDAGToDAGHVX.cpp
```diff
@@ -3009,3 +3009,469 @@ void HexagonDAGToDAGISel::SelectHVXDualOutput(SDNode *N) {
...
+    SDNode *ConvNode = CurDAG->getMachineNode(Hexagon::V6_vconv_hf_qf16, DL,
+                                              ResTy, SDValue(AddNode, 0));
+    ReplaceUses(SDValue(N, 0), SDValue(ConvNode, 0));
+    CurDAG->RemoveDeadNode(N);
+    return;
+  }
+  // v0.hf = vsub(v0.hf,v1.hf) is translated to
+  // v2.qf16 = vsub(v0.hf,v1.hf); v0.hf = v2.qf16.
+  case Intrinsic::hexagon_V6_vsub_hf_hf_128B: {
+    SDNode *AddNode = CurDAG->getMachineNode(
```
androm3da: Should be `SubNode`?

llvm/lib/Target/Hexagon/HexagonISelDAGToDAGHVX.cpp
fhossein-quic: @aankit-ca isIEEEHVXIntrinsic now only returns true for the 22 intrinsics that have a corresponding translation case, any unrecognized intrinsic hit llvm_unreachable now. I will handle the 9 missing ones in the follow-up patches.

llvm/lib/Target/Hexagon/HexagonISelDAGToDAGHVX.cpp
fhossein-quic: Yes :), Done!

llvm/test/CodeGen/Hexagon/autohvx/ieeetoqfloat.ll
fhossein-quic: Done

llvm/lib/Target/Hexagon/HexagonISelDAGToDAG.cpp
fhossein-quic: fixed

androm3da (approved):

quic-santdas (approved):
LGTM.

PR #198833, opened 2026-05-20
[llvm][Instrumentor] Fix non-determinism in Instrumentor

ilovepi:
In InstrumentorStubPrinter the StringMap was being iterated over, which
broke the Instrumentation/Instrumentor/write_config.ll test under
LLVM_REVERSE_ITERATION builds. We can use a MapVector to ensure the
order is stable. We only need to update the test json ordering to match
the stable order for with and without LLVM_REVERSE_ITERATION.

### DIFF
--- a/llvm/lib/Transforms/IPO/InstrumentorConfigFile.cpp
+++ b/llvm/lib/Transforms/IPO/InstrumentorConfigFile.cpp
@@ -86,14 +86,14 @@ void writeConfigToJSON(InstrumentationConfig &IConf, StringRef OutputFile,
 
     J.attributeBegin(InstrumentationLocation::getKindStr(Kind));
     J.objectBegin();
-    for (auto &ChoiceIt : KindChoices) {
-      J.attributeBegin(ChoiceIt.getKey());
+    for (auto &[Name, Choice] : KindChoices) {
+      J.attributeBegin(Name);
       J.objectBegin();
-      J.attribute("enabled", ChoiceIt.second->Enabled);
-      J.attribute("filter", ChoiceIt.second->Filter);
+      J.attribute("enabled", Choice->Enabled);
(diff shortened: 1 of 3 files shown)


ilovepi:
* **#198833** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/198833?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/198833?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

llvm/lib/Transforms/IPO/InstrumentorConfigFile.cpp
```diff
@@ -200,7 +200,8 @@ bool readConfigFromJSON(InstrumentationConfig &IConf, StringRef InputFile,
             "malformed JSON configuration, expected an object", DS_Warning));
         continue;
       }
-      auto *IO = IChoiceMap.lookup(ObjIt.first);
+      auto *ChoiceIt = IChoiceMap.find(ObjIt.first);
```
kevinsala: I think you can still use lookup.

kevinsala (approved):
Thanks for the fix. Apart from the comment below, LGTM

llvm/lib/Transforms/IPO/InstrumentorConfigFile.cpp
ilovepi: huh, for some reason I thought MapVector didn't have that API. thanks that's much nicer.

kevinsala: Thanks!

ilovepi:
PowerPC failures are unrelated, so I'm going to land, since the other tests are passing.

PR #198834, opened 2026-05-20
[libc] Make FreeListHeap::free ignore null pointers.

simpal01:
Update `FreeListHeap::free` to return immediately when passed `nullptr`.

This matches the expected `free(nullptr)` behaviour as per the C standard and avoids running pointer validation on a null pointer allowing the hermetic malloc test libc/test/src/stdlib/malloc_test.cpp to pass.

As per standard:
The free function, paragraph:
```
  The free function causes the space pointed to by
  ptr to be deallocated, that is, made available for
  further allocation. If ptr is a null pointer,
  no action occurs.
```


mysterymath (approved):
Huh, wild that I missed this. LGTM!

michaelrj-google (approved):
LGTM, thanks for the fix!

PR #198838, opened 2026-05-20
[clang] Add -fcrash-diagnostics-tar for tarball of crash reproducer files

aeubanks:
Makes it easier to move around crash diagnostics.


aeubanks:
sorry, looks like this doesn't handle directories that we add in some crash dumps, let me fix that

clang/test/Driver/crash-diagnostics-modules.c
```diff
@@ -0,0 +1,17 @@
+// UNSUPPORTED: system-windows
```
efriedma-quic: Can we write some version of this test that works on Windows?

clang/lib/Driver/Driver.cpp
```diff
@@ -2271,9 +2273,71 @@ void Driver::generateCompilationDiagnostics(
...
+            Tar->append(llvm::sys::path::filename(TempFile),
+                        (*BufferOrErr)->getBuffer());
+          } else {
+            Diag(clang::diag::note_drv_command_failed_diag_msg)
+                << (std::string("Error reading file for tarball: ") + TempFile);
+          }
+        }
+      }
+      Diag(clang::diag::note_drv_command_failed_diag_msg)
+          << (std::string("Crash reproducer tarball created at: ") +
```
efriedma-quic: Do we want to modify the existing text, instead of adding an additional line?  Presumably if someone creates a tar-file, we want them to attach just the tar-file to a bug report, not all the other files.

AaronBallman (commented):
Thank you for this! Please be sure to add a release note to `clang/docs/ReleaseNotes.rst` so users know about the feature. Perhaps we should document it more loudly in the user's manual too?

One thing I'd like to be sure we have test coverage for (and maybe we do, I could have missed it): what happens when the source is read from, say, `stdin` as opposed to a file we can tar? Similarly, we should have test coverage for an invalid file path to the tar file.

aeubanks:
added tests for stdin/invalid tar path

added flag to release notes and user's manual

I think it'd be nice to turn this on by default and have it automatically dump a tarball in the same directory alongside the other crash reproducer files, but that should be a separate PR

AaronBallman (approved):
LGTM!

PR #198843, opened 2026-05-20
[ExpandVariadics] Fixup handeling of mismatched return types

AlexMaclean:
Fix an assert when RAUW tries to replace all uses with a different type.


JonChesterfield:
What if you have multiple calls with different mismatched return types, which should win?

Could you add the IR that asserts on master? Rauw failing is a bad thing, but the type on calls doesn't necessarily mean very much. It's not checked by the verifier and we have type punning nonsense leaking into the representation. The checking on variadic types return value shouldn't looser than on normal, so the assert in the commit message really shouldn't be happening.

If it is, the fix is probably a bitcast at the bad call sites, not changing the function return, but I may have misread the patch.

AlexMaclean:
> What if you have multiple calls with different mismatched return types, which should win?

I'm not sure I understand. Each call site would have the type consistent with the type of that call site prior to this pass running. 

> Could you add the IR that asserts on master? Rauw failing is a bad thing, but the type on calls doesn't necessarily mean very much. It's not checked by the verifier and we have type punning nonsense leaking into the representation. The checking on variadic types return value shouldn't looser than on normal, so the assert in the commit message really shouldn't be happening.

Here is one of the tests updated in this the PR:

https://godbolt.org/z/ec7GTaKeq

> If it is, the fix is probably a bitcast at the bad call sites, not changing the function return, but I may have misread the patch.

I don't think trying to coerce the type is going to be possible. For example in the case above, I'm not sure a bitcast with a struct type is valid. There can be other cases as well where the size does not match. The current solution implemented in this PR to use the callee return type seems like a preferable alternative.

AlexMaclean:
@JonChesterfield / @jhuber6 ping

jhuber6 (approved):

PR #198844, opened 2026-05-20
[NFC][AArch64][Cyclone] Model WriteSTP with a local SchedWriteRes

jcohen-apple:
Cyclone scheduling model uses SchedAlias between 2 SchedWriteRes definitions from AArch64Schedule.td.
This prevents other scheduling models from aliasing WriteSTP. This patch address the issue by defining a new CyWriteSTP and using that instead.


tomershafir (approved):
LGTM

PR #198845, opened 2026-05-20
Add a pre-filter to the breakpoint resolver overrides

jimingham:
That way if an override only wants to override, say, file and line breakpoints, it can say that when it registers itself, and we won't do the work of creating the resolver and checking with it if the breakpoint wasn't of the requested type.


jimingham:
gentle review ping...

lldb/include/lldb/API/SBTarget.h
```diff
@@ -639,9 +639,10 @@ class LLDB_API SBTarget {
                     lldb::SBError &error);
 
   /// Adds a breakpoint override implemented by class_name.  Returns the ID
-  /// of the new override or LLDB_INVALID_INDEX64 on error.
+  /// of the new override or LLDB_INVALID_INDEX64 on error.  The type_mask
+  /// is composed of elements from the lldb::BreakpointResolverType enum.
   uint64_t AddBreakpointOverride(const char *class_name,
-                                 const char *description,
+                                 const char *description, uint64_t type_mask,
```
JDevlieghere, line 645: Should we give `type_mask` a default of `eResolverFileAndLine` on the API itself (or explicitly reject a 0 mask with an error?), so out-of-tree callers get the same backward-compatible behavior as the command instead of a silently inert override?

lldb/include/lldb/Breakpoint/BreakpointResolver.h
```diff
@@ -154,12 +154,22 @@ class BreakpointResolver : public Searcher {
...
     if (SubclassID > ResolverTy::LastKnownResolverType)
       return ResolverTy::UnknownResolver;
     return (enum ResolverTy)SubclassID;
   }
 
+  uint64_t MaskForResolverTy();
+
+  /// Returns true if this resolver is in the mask (made of elements of
+  /// BreakpointResolverType.
+  bool ResolverInMask(uint64_t mask);
```
JDevlieghere: This is unimplemented? 
```suggestion
```

lldb/include/lldb/Breakpoint/BreakpointResolver.h
```diff
@@ -154,12 +154,22 @@ class BreakpointResolver : public Searcher {
   /// for any other purpose, as the values may change as LLDB evolves.
   unsigned getResolverID() const { return SubclassID; }
 
+  /// This checks whether the resolver's type matches the enum
+  /// lldb::eBreakpointResolverType.
```
JDevlieghere:
```suggestion
  /// lldb::BreakpointResolverType.
```

lldb/include/lldb/Target/Target.h
```diff
@@ -1008,16 +1008,20 @@ class Target : public std::enable_shared_from_this<Target>,
   // be the one we use.  If no overrides return an override resolver, we'll use
   // the original one.
 
-  // This is the abstract version of the override.  Particular implementations
-  // e.g. the scripted override will derive from this.
+  /// This is the abstract version of the override.  Particular implementations
+  /// e.g. the scripted override resolve instantiate actual versions of the
```
JDevlieghere: This seems to be missing a comma and needs to decide between resolve and instantiate?

lldb/source/Breakpoint/BreakpointResolver.cpp
```diff
@@ -63,6 +63,48 @@ BreakpointResolver::NameToResolverTy(llvm::StringRef name) {
   return UnknownResolver;
 }
 
+bool BreakpointResolver::ResolverTyInMask(uint64_t mask) {
+  if (mask == eResolverUnknown)
+    return false;
+
+  return !((mask & MaskForResolverTy()) == 0);
```
JDevlieghere:
```suggestion
  return (mask & MaskForResolverTy()) != 0;
```

lldb/test/API/functionalities/breakpoint/scripted_bkpt/overrides_resolver/bkpt_resolver.py
```diff
@@ -63,5 +109,21 @@ def set_breakpoint(self, bkpt: lldb.SBBreakpoint):
...
+        type = initial_resolver.GetValueForKey("Type").GetStringValue(1000)
+        if type != "SymbolName":
+            trivial_not_name += 1
+
         """Trivial - overrides nothing"""
         return False
+
+
+def __lldb_init_module(debugger, dict):
+    print(f"About to run: command script add -c {__name__}.CheckerCommand checker")
```
JDevlieghere:
```suggestion
```

lldb/test/API/functionalities/breakpoint/scripted_bkpt/overrides_resolver/bkpt_resolver.py
```diff
@@ -1,5 +1,43 @@
...
+override_not_file = 0
+trivial_not_name = 0
+
+
+class CheckerCommand:
+    def __init__(self, debugger, internal_dict):
+        self.debugger = debugger
+
+    def get_short_help(self):
+        return "A command the checks how many times the resolvers were called"
```
JDevlieghere:
```suggestion
        return "A command that checks how many times the resolvers were called"
```

lldb/include/lldb/API/SBTarget.h
```diff
@@ -639,9 +639,10 @@ class LLDB_API SBTarget {
                     lldb::SBError &error);
 
   /// Adds a breakpoint override implemented by class_name.  Returns the ID
-  /// of the new override or LLDB_INVALID_INDEX64 on error.
+  /// of the new override or LLDB_INVALID_INDEX64 on error.  The type_mask
+  /// is composed of elements from the lldb::BreakpointResolverType enum.
   uint64_t AddBreakpointOverride(const char *class_name,
```
bulbazord, line 644: This is an SBAPI... You'll need to add a new method instead of modifying this one, else we break ABI.

jimingham: That API has only been in place for a month or so now, and the only client I know of is expecting this change.  I think it should be possible to have "bake in" times for new API's so we don't end up unnecessarily uglifying the  API set.

bulbazord: Added on May 11th. Makes sense. This hasn't landed in any release yet, so I think this is fine.

lldb/include/lldb/API/SBTarget.h
jimingham: I thought about adding a default, but this is not a common API, and I think it's worth making users of it state what kinds of breakpoints they are overriding explicitly.

The way it would currently work is the invalid mask would get passed to the concrete implementation of the override, and its Validate call would fail if the mask is bogus.

But there's no reason not to check the mask explicitly up front, and that would centralize that bit of error handling.  I'll fix that.

lldb/include/lldb/API/SBTarget.h
jimingham: Thanks.  I actually prefer having just this interface, because the case it supports: "stop for every kind of breakpoint and ask me if I want to override it" shouldn't be the easiest way to register one of these things.  You really should have to think about which breakpoint types you are able to override.

lldb/source/Breakpoint/BreakpointResolver.cpp
jimingham: Yeah, no idea why I wrote it that awkward way...

lldb/test/API/functionalities/breakpoint/scripted_bkpt/overrides_resolver/bkpt_resolver.py
jimingham: Abandoned debug print, it's gone...

lldb/include/lldb/Target/Target.h
jimingham: That was actually a missing r on the end of "resolve", but I added the commas as well.

medismailben (approved):
LGTM! I think we could make the mask `uint32_t` or even smaller.

lldb/include/lldb/lldb-enumerations.h
```diff
@@ -1582,6 +1583,25 @@ enum BinaryInformationLevel {
   eBinaryInformationLevelFull
 };
 
+/// This reflects the BreakpointResolver::ResolverTy, but this is a convenient
+/// enum for making a mask to pass to RegisterOverrideResolver.  It has to be
+/// kept in sync with the ResolverTy.
+
```
JDevlieghere:
```suggestion
```

lldb/source/Breakpoint/BreakpointResolver.cpp
```diff
@@ -63,6 +63,55 @@ BreakpointResolver::NameToResolverTy(llvm::StringRef name) {
...
+    return eResolverAddress;
+  if (thisID == NameResolver)
+    return eResolverName;
+  if (thisID == FileRegexResolver)
+    return eResolverFileRegex;
+  if (thisID == PythonResolver)
+    return eResolverPython;
+  if (thisID == ExceptionResolver)
+    return eResolverException;
+  return eResolverUnknown;
```
JDevlieghere: Why not a switch?

lldb/source/Target/Target.cpp
```diff
@@ -1000,6 +1019,19 @@ bool Target::ProcessIsValid() {
...
 
+lldb::BreakpointResolverSP
+Target::CheckBreakpointOverrides(lldb::BreakpointResolverSP original_sp) {
+  for (auto const &elem : m_breakpoint_overrides) {
+    if (!original_sp->ResolverTyInMask(elem.second->GetTypeMask()))
+      continue;
+    lldb::BreakpointResolverSP overriden_sp;
+    overriden_sp = elem.second->CheckForOverride(*this, original_sp);
+    if (overriden_sp)
+      return overriden_sp;
```
JDevlieghere: This is 3 line for something that can be done in one. At least it should collapse the default construction with the copy assignment. But even better just do: 
```suggestion
    if (lldb::BreakpointResolverSP overriden_sp = elem.second->CheckForOverride(*this, original_sp))
      return overriden_sp;
```

JDevlieghere (approved):
LGTM besides the nits.

JDevlieghere:
> I think we could make the mask uint32_t or even smaller.

If we ever have to change our mind later, then that becomes an ABI break. I think the performance improvements are negligible that I would rather play it safe, even if it's overkill now (that's what they said about IPv4 ;-)

lldb/source/Breakpoint/BreakpointResolver.cpp
jimingham: Because, having written DescribeMask it was super easy to turn that into MaskForResolverTy as written...

jimingham:
> > I think we could make the mask uint32_t or even smaller.
> 
> If we ever have to change our mind later, then that becomes an ABI break. I think the performance improvements are negligible that I would rather play it safe, even if it's overkill now (that's what they said about IPv4 ;-)

I agree, we're never going to have enough of these that the size difference will matter, so we might as well give ourselves as much upfront space for resolvers as we can.

Issue #192311, opened 2026-04-15
[CIR] Upstream missing support for vector type operations


andykaylor:
### Overview

Upstream code from the incubator project to replace errorNYI calls in the
CIR codegen implementation for various vector type operations, including:
- Vec3 load/store with widening to vec4 when `PreserveVec3Type` is disabled
- ExtVector element store via read/modify/write shuffle
  (`emitStoreThroughExtVectorComponentLValue`)
- ExtVectorBoolType handling in ext vector element loads
- Nested ext vector element access (compositing indices for non-simple lvalues)
- Vector ternary operator (`cir::VecTernaryOp`) for OpenCL/SVE vector
  conditions in `?:` expressions
- `__builtin_astype` / `as_type` for bitcasting between vector types

Upstream NYI locations:
- `CIRGenExpr.cpp` — `emitStoreThroughLValue`: non-simple lvalue (ExtVectorElt)
- `CIRGenExpr.cpp` — `emitStoreOfScalar`: Vec3 widening to vec4 before store
- `CIRGenExpr.cpp` — `emitLoadOfScalar`: Vec3 load with shuffle back to
  original size
- `CIRGenExpr.cpp` — `emitLoadOfExtVectorElementLValue`: ExtVectorBoolType
  shuffle result
- `CIRGenExpr.cpp` — `emitExtVectorElementExpr`: non-simple nested ext vector
  element access
- `CIRGenExprScalar.cpp` — `VisitAbstractConditionalOperator`: vector ternary
  and SVE vector ternary
- `CIRGenExprScalar.cpp` — `VisitAsTypeExpr`: `__builtin_astype` bitcast

### Suggested minimal test case

```cpp
void test_ext_vector_element(float4 *p) {
  p->xy = p->zw;
}

float4 test_vec_ternary(int4 cond, float4 a, float4 b) {
  return cond ? a : b;
}
```

### Existing incubator tests

- `clang/test/CIR/CodeGen/vectype-ext.cpp`
- `clang/test/CIR/CodeGen/vector-ext-element.cpp`
- `clang/test/CIR/CodeGen/vectype.cpp`
- `clang/test/CIR/CodeGen/OpenCL/as_type.cl`

AmrDeveloper:
Unary ++, -- on Vector of floats depends on powerpc target

PR #198846, opened 2026-05-20
[CIR] Implement TernaryOp for Vec of float in OpenCL

AmrDeveloper:
Implement support for TernaryOp for Vec float in OpenCL

Issue https://github.com/llvm/llvm-project/issues/192311


andykaylor (approved):
lgtm

PR #198850, opened 2026-05-20
[X86] Change Gather and Scatter cost to 1024 for ZN4 and ZN5

Jason-Van-Beusekom:
Gather and scatter instructions are quite slow on ZN4 and ZN5 compared to other architectures  resulting in much worse performance compared when compiling with `-march=native`

Ie for:
VPSCATTERQQ (zmm)
Zen 4 μops = 48 	
Skylake-X μops ≤20

Port pressure is also worse — Zen 4's VPSCATTERQQ has 18 μops restricted to ports {4, 5}, vs. SKX's 8 μops on ports {2, 3} uops.info/html-ports/ZEN4/VPSCATTERQQ_VSIB_ZMM_K_ZMM-Measurements.html

This change also follows the AMD Zen 4 Software Optimization Guide (SOG) 57647 Rev. 1.00 January 2023

    https://amd.com/system/files/TechDocs/57647.zip

Section 2.11.2 Code recommendations:

    Avoid GATHER instructions when the indices are known ahead of time. Vector loads followed by shuffles result in a higher load bandwidth.

I also tested in the test suite and Found general performance improvements on a `AMD EPYC 9654`

```bash
Program                                       exec_time                             
                                              results-march-fix results-march diff  
SingleSour...lgebra/kernels/doitgen/doitgen                       0.51          inf%
MultiSourc...marks/VersaBench/ecbdes/ecbdes     1.31              3.49        165.5%
MultiSource/Benchmarks/nbench/nbench            0.63              0.76         19.8%
SingleSource/Benchmarks/Misc/oourafft           1.30              1.54         18.5%
SingleSour...h/linear-algebra/solvers/lu/lu    15.65             17.99         15.0%
SingleSour...olvers/gramschmidt/gramschmidt     2.40              2.67         11.1%
SingleSour...mining/correlation/correlation     1.21              1.34         10.5%
MultiSourc...chmarks/VersaBench/8b10b/8b10b     1.02              1.11          9.3%
MultiSourc...VC/Symbolics-dbl/Symbolics-dbl     0.82              0.89          8.5%
MultiSourc...OE-ProxyApps-C/XSBench/XSBench     1.22              1.32          8.2%
SingleSour...bench/medley/nussinov/nussinov     3.02              3.25          7.7%
Fortran/gf...cute-regression__matmul_15_f90     7.80              8.39          7.6%
MultiSourc...trolLoops-dbl/ControlLoops-dbl     0.58              0.62          6.6%
MultiSourc...SC_Sequoia/CrystalMk/CrystalMk     1.50              1.60          6.6%
MultiSourc...lds-flt/CrossingThresholds-flt     0.75              0.80          6.1%
                           Geomean difference                                   1.3%
/home/users/vanbeuse/.local/lib/python3.12/site-packages/pandas/core/nanops.py:1028: RuntimeWarning: invalid value encountered in subtract
  sqr = _ensure_numeric((avg - values) ** 2)
              exec_time                          
run   results-march-fix results-march        diff
count  121.000000        118.000000    122.000000
mean   3.043355          3.155764      inf       
std    8.874981          9.038026     NaN        
min    0.502000          0.506400     -1.000000  
25%    0.728600          0.763325     -0.010159  
50%    1.166000          1.185750      0.002239  
75%    2.093100          2.209225      0.020039  
max    92.808000         93.154500     inf       
```

For the reproducer:

``` c
#include <stdio.h>
#include <stdlib.h>
#include <time.h>

#define N 5

static double s[N][N];
static double a[N][N][N][N];
static double b[N][N];

int main(int argc, char **argv) {
    int niter = 30000000;
    if (argc > 1) niter = atoi(argv[1]);

    double w = 0.125;

    srand(42);
    for (int i = 0; i < N*N*N*N; i++) ((double*)a)[i] = (double)rand()/RAND_MAX * 1e4;
    for (int i = 0; i < N*N; i++) ((double*)b)[i] = (double)rand()/RAND_MAX * 1e-2;
    for (int i = 0; i < N*N; i++) ((double*)s)[i] = 0.0;

    clock_t t0 = clock();

    for (int iter = 0; iter < niter; iter++) {
        for (int i1 = 0; i1 < N; i1++)
            for (int i2 = 0; i2 < N; i2++)
                for (int i3 = 0; i3 < N; i3++)
                    for (int i4 = 0; i4 < N; i4++)
                        s[i1][i2] += a[i1][i3][i2][i4] * b[i3][i4] * w;
    }

    clock_t t1 = clock();

    double checksum = 0;
    for (int i = 0; i < N*N; i++) checksum += ((double*)s)[i];

    printf("Time:     %.4f s\n", (double)(t1-t0)/CLOCKS_PER_SEC);
    printf("Checksum: %.8e\n", checksum);
    return 0;
}
```

This fix results in a significant performance increase on a  AMD EPYC 9654:
``` bash
clang -O3 -ffast-math -march=native repro.c 
./a.out 
Time:     2.6538 s
Checksum: 5.96228077e+10
clang_nightly -O3 -ffast-math -march=native repro.c
 
./a.out 
Time:     10.3617 s
Checksum: 5.96228077e+10
```

This addresses the ZN4 and ZN5 side of [91370](https://github.com/llvm/llvm-project/issues/91370)

Assisted with Opus 4.6


Andarwinux:
```
[arch@wsl llvmbin]$ clang -O3 -march=tigerlake -mprefer-vector-width=512 -ffast-math repro.c -mno-gather
[arch@wsl llvmbin]$ time ./a.out
Time:     2.1080 s
Checksum: 5.96228077e+10

real    0m2.136s
user    0m2.109s
sys     0m0.001s
[arch@wsl llvmbin]$ clang -O3 -march=tigerlake -mprefer-vector-width=512 -ffast-math repro.c
[arch@wsl llvmbin]$ time ./a.out
Time:     5.0835 s
Checksum: 5.96228077e+10

real    0m5.147s
user    0m5.083s
sys     0m0.002s
```
gather on tigerlake is also very slow (GDS mitigation disabled).

Jason-Van-Beusekom:
> ```
> [arch@wsl llvmbin]$ clang -O3 -march=tigerlake -mprefer-vector-width=512 -ffast-math repro.c -mno-gather
> [arch@wsl llvmbin]$ time ./a.out
> Time:     2.1080 s
> Checksum: 5.96228077e+10
> 
> real    0m2.136s
> user    0m2.109s
> sys     0m0.001s
> [arch@wsl llvmbin]$ clang -O3 -march=tigerlake -mprefer-vector-width=512 -ffast-math repro.c
> [arch@wsl llvmbin]$ time ./a.out
> Time:     5.0835 s
> Checksum: 5.96228077e+10
> 
> real    0m5.147s
> user    0m5.083s
> sys     0m0.002s
> ```
> 
> gather on tigerlake is also very slow (GDS mitigation disabled).

I can add tigerlake to this patch, but I do not have access to a tigerlake system so I would not be able to test it.

MattPD (commented):
Mostly looks good with a suggestion: Consider adding an end-to-end (loop) vectorizer test with a loop matching the motivating high-level intent (we primarily care about avoiding AVX-512 gather and scatter for AMD Zen 4 & 5 whereas the cost model is the underlying implementation detail which we aren't necessarily attached to; e.g., use `opt -passes=loop-vectorize -mcpu=znver4` and verify with `CHECK-NOT: @llvm.masked.gather`).

RKSimon (commented):
We already have TuningPreferNoGather/TuningPreferNoScatter flags which might be a better fit?

Jason-Van-Beusekom:
> We already have TuningPreferNoGather/TuningPreferNoScatter flags which might be a better fit?

I left it alone as my impression of https://github.com/llvm/llvm-project/commit/993bdb047c90e9b85fb91578349a9faf4f6a853d#diff-68c9747a35ab1c8c3411f61ec8ba1aafee80c6dc9a9ea73828e5ade1de5ed2ce was that TuningPreferNoGather fully disables gather and scatter, while this case just makes the cost much higher.

Now there is an argument that the cost is so high it would never take it, i'll try to see if I can have a example where gather and scatter are still used.

ganeshgit:
If we set the cost in the compiler too low, the cost model will become optimistic. It will aggressively vectorize loops where we see the overhead of gather/scatter. If we set the cost too high, the compiler becomes too conservative. It will fail to vectorize loops that would have seen significant speedups.
I think we need costing which is more dynamic and context-aware. This might be a function of Vector Factor, Data Type, and the instruction (gather) sequence we would like to generate. We are working on a pre-tabulated cost considering this function which would help us in deciding whether to opt-in or opt-out with gathers. Hold on this patch. From AMD, Sumukh J Bharadwaj will submit a PR considering such a function with some perf numbers.

amd-subharad:
Thank you for the heads up @ganeshgit, here is the proposed gather scatter cost lowering patch
https://github.com/llvm/llvm-project/pull/199488

Jason-Van-Beusekom:
> Thank you for the heads up @ganeshgit, here is the proposed gather scatter cost lowering patch [https://github.com/llvm/llvm-project/pull/199488](url)

Just a heads up this links to an unrelated llvm PR and not https://github.com/llvm/llvm-project/pull/199488

amd-subharad:
> > Thank you for the heads up @ganeshgit, here is the proposed gather scatter cost lowering patch [https://github.com/llvm/llvm-project/pull/199488](url)
> 
> Just a heads up this links to an unrelated llvm PR and not #199488

Ohh sorry about that I have edited it to reflect the actual PR

PR #169213, opened 2025-11-23
[AMDGPU] Refactor GFX11 VALU Mask Hazard Waitcnt Merging

perlfu:
Move GFX11 SGPR VALU mask hazard waitcnt merging to the a forward scan within the AMDGPUWaitSGPRHazard pass.
This simplifies the hazard recognizer code and allows merging of waitcnt instructions in cases where SGPRs unaffected by pending writes are used.
In turn this greatly decreasing numbers of waits inserted in sequences of V_CMP instructions writing SGPRs improving VALU pipeline performance.

PR #198852, opened 2026-05-20
[AMDGPU] Skip processing some uses in fixVALUMaskWriteHazard (NFC)

IgWod:
This targets two areas in the code where defs and uses are checked in the same loop. This patch splits the loop into two so uses are only checked if required. This provides a modest 2% compile-time improvement on few tested pipelines where `fixVALUMaskWriteHazard` is a bottleneck.

Assisted-by: Claude Code

### DIFF
--- a/llvm/lib/Target/AMDGPU/GCNHazardRecognizer.cpp
+++ b/llvm/lib/Target/AMDGPU/GCNHazardRecognizer.cpp
@@ -3436,6 +3436,18 @@ bool GCNHazardRecognizer::fixVALUMaskWriteHazard(MachineInstr *MI) {
   auto IsVCC = [](const Register Reg) {
     return Reg == AMDGPU::VCC || Reg == AMDGPU::VCC_LO || Reg == AMDGPU::VCC_HI;
   };
+  auto IsRelevantSGPR = [&](const MachineOperand &Op) {
+    Register Reg = Op.getReg();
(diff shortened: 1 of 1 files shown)


IgWod:
Objectively considering a limited benefit, it may not be worth merging it, but since I already got a patch ready, I thought I may as well submit it and get some feedback.

llvm/lib/Target/AMDGPU/GCNHazardRecognizer.cpp
```diff
@@ -3452,36 +3464,27 @@ bool GCNHazardRecognizer::fixVALUMaskWriteHazard(MachineInstr *MI) {
...
     assert(!HazardDef);
     HazardDef = &Op;
+    break;
   }
 
   if (!HazardDef)
     return false;
 
+  // Then check for SGPR reads.
+  for (MachineOperand &Op : MI->operands()) {
```
jayfoad, line 3481: Use `MI->all_uses()`?

perlfu:
This would somewhat be fixed by #169213

IgWod:
> This would somewhat be fixed by https://github.com/llvm/llvm-project/pull/169213

Thanks, I wasn't aware of your PR. It looks like it'll have pretty much the same impact on those two loops as this PR. In that case I think we can close this patch and wait for yours to go in.

PR #198853, opened 2026-05-20
[RegAlloc][X86] Rematerialization of instructions with physical register

weiguozhi:
We can rematerialize instructions with physical register definition, with additional check if the physical register is dead at the use site.


llvm/lib/CodeGen/CalcSpillWeights.cpp
```diff
@@ -30,6 +30,12 @@ using namespace llvm;
 
 #define DEBUG_TYPE "calcspillweights"
 
+cl::opt<bool> AllowRematerializePhysReg(
+    "allow-rematerialize-phys-reg",
+    cl::desc("Allow rematerialize instructions with physical register "
+             "definition, like POPCNT on x86 can write EFLAGS register."),
+    cl::init(false), cl::Hidden);
```
arsenm, line 37: I don't want a new knob for this

llvm/test/CodeGen/X86/remat-flags.ll
```diff
@@ -0,0 +1,165 @@
...
+  %v4 = load volatile i32, i32* %p2, align 8
+  %v5 = add i32 %v1, %v4
+  %v6 = sub i32 %v2, %v5
+  store volatile i32 %v6, i32* %p2, align 8
+
+  %p64 = getelementptr i64, i64* %ptr, i32 6
+  %v7  = load volatile i64, i64* %p64, align 8
+  store volatile i64 %v7, i64* %p64, align 8
+
+  store volatile i32 %v6, i32* %p2, align 8
```
arsenm: Use opaque pointers

llvm/test/CodeGen/X86/remat-flags.ll
```diff
@@ -0,0 +1,165 @@
...
+
+; This file tests rematerialization of instructions with physical register
+; modified.
+
+; Instruction popcnt can be rematerialized.
+define void @test1(i32 %v1, i32 %v2, i32* %ptr) {
+; CHECK-LABEL: test1:
+; CHECK:       # %bb.0:
+; CHECK-NEXT:    pushl %ebp
+; CHECK-NEXT:    .cfi_def_cfa_offset 8
```
arsenm: Add nounwind to avoid CFI noise

llvm/test/CodeGen/X86/remat-flags.ll
```diff
@@ -0,0 +1,165 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 2
+; RUN: llc -mtriple=i686-unknown-unknown -mattr=+popcnt -mattr=cmov -frame-pointer=all -allow-rematerialize-phys-reg=true < %s | FileCheck %s
+
+; This file tests rematerialization of instructions with physical register
+; modified.
+
```
arsenm, line 6: This could also use a mir test

llvm/include/llvm/CodeGen/TargetInstrInfo.h
```diff
@@ -191,6 +191,14 @@ class LLVM_ABI TargetInstrInfo : public MCInstrInfo {
            (MI.getDesc().isRematerializable() && isReMaterializableImpl(MI));
   }
 
+  /// Return true if an instruction can be rematerialized as a different
+  /// instruction without clobbering the physical register.
+  virtual bool
+  canRematerializeIgnorePhysRegDef(const MachineInstr &MI,
+                                   const MachineOperand &MO) const {
+    return false;
+  }
```
arsenm, line 214: I'd rather avoid adding a new hook. The description and arguments aren't clear here. What is the physical register? What different instruction? The description mentions dead physical registers, so can't you just do this if you infer the physreg is dead unconditionally?

llvm/lib/Target/X86/X86InstrInfo.cpp
```diff
@@ -755,6 +755,26 @@ static bool regIsPICBase(Register BaseReg, const MachineRegisterInfo &MRI) {
   return isPICBase;
 }
 
+/// Return true if an instruction can be rematerialized as a different
+/// instruction without clobbering eflags register, or the definition of the
+/// physical register really doesn't matter.
+bool X86InstrInfo::canRematerializeIgnorePhysRegDef(
+    const MachineInstr &MI, const MachineOperand &MO) const {
+  assert(MO.isReg() && MO.getReg() && MO.getReg().isPhysical());
```
arsenm, line 763: I believe isPhyical is true for NoRegister. Should this be asserting it's a physreg?

llvm/lib/CodeGen/CalcSpillWeights.cpp
```diff
@@ -146,17 +152,44 @@ bool VirtRegAuxInfo::isRematerializable(const LiveInterval &LI,
...
+  auto MBB = LIS.getMBBFromIndex(UseIdx);
+  MachineInstr *UseMI = LIS.getInstructionFromIndex(UseIdx);
+  MachineBasicBlock::iterator UseIt = UseMI;
+  if (!UseMI) {
+    if (MBB->empty() ||
+        UseIdx < LIS.getInstructionIndex(*MBB->begin()).getBaseIndex())
+      UseIt = MBB->begin();
+    else {
+      UseIt = MBB->end();
+    }
```
arsenm, line 170: I'd assume there's a function to do this position check for you; you shouldn't need to worry about empty use blocks?

llvm/lib/CodeGen/CalcSpillWeights.cpp
weiguozhi: Me too.

When I enabled this optimization by default, I got 16 test failures in AMDGPU. I checked some of them, they are related to rematerialization of instructions with physical register definition, so they should be handled by canRematerializeIgnorePhysRegDef. 

I don't know anything about AMDGPU, so I added this command line option to disable the optimization by default.

llvm/test/CodeGen/X86/remat-flags.ll
weiguozhi: done.

llvm/test/CodeGen/X86/remat-flags.ll
weiguozhi: done.

llvm/test/CodeGen/X86/remat-flags.ll
weiguozhi: A new mir test is added.

llvm/include/llvm/CodeGen/TargetInstrInfo.h
weiguozhi: I added an example in the comment.

AArch64 and PPC can optionally set condition register, so an instruction setting condition register can be rematerialized as an instruction without setting condition register, there should be more opportunities of this optimization on those targets.

llvm/lib/Target/X86/X86InstrInfo.cpp
weiguozhi: MO.getReg() is used to check it isn't a NoRegister.
So both of !NoRegister and isPhysical are checked.

llvm/lib/CodeGen/CalcSpillWeights.cpp
weiguozhi: The purpose of this code snippet is to find a valid instruction iterator UseIt, it is used later to check if a physical register is dead at this position. For an empty basic block, the only valid instruction iterator is begin().

weiguozhi:
ping

llvm/include/llvm/CodeGen/TargetInstrInfo.h
```diff
@@ -191,6 +191,25 @@ class LLVM_ABI TargetInstrInfo : public MCInstrInfo {
            (MI.getDesc().isRematerializable() && isReMaterializableImpl(MI));
   }
 
+  /// Return true if an instruction can be rematerialized as a different
+  /// instruction without clobbering the physical register.
+  ///
+  /// For example in x86 we can rematerialize
+  ///     %100 = MOV32r0
+  /// it is actually translated to
+  ///     xor  %100, %100
```
qcolombet, line 200: Instead of doing a dance to teach an X86 quirk to the rest of the infrastructure, can we solve that problem solely in x86?
- Remove the EFLAGS def in the MOV32r0 target description
- Use the lowering that clobber eflags or not based on the liveness of eflags when we expand the pseudo instruction.

This would fix the rematerialization capabilities and keep the weirdness of dynamically changing the physical definitions to the x86 backend.

qcolombet (changes requested):

llvm/include/llvm/CodeGen/TargetInstrInfo.h
weiguozhi: It's not a X86 specific problem, it's even more common in some other architectures.

1. AArch64, many  ALU instructions can optionally update flag register, with the help of this function, we can rematerialize instructions with flag register definition as instructions without flag register update.
2. PowerPC, similar to AArch64, many ALU instructions can optionally update flag register.
3. AMDGPU, I observed some originally rematerializable instructions can define physical register, I don't know the reason of them, but I believe this function can help to solve the problem.

Compare to them, the impact to X86 is limited to some special patterns.

david-xl: Add a comment here saying this is currently only implemented for x86 for applicable to AArch64?

weiguozhi: Comment is added.

weiguozhi:
@arsenm @qcolombet do you have any other comments?

llvm/lib/CodeGen/CalcSpillWeights.cpp
RKSimon: regrettably I think you're going to have yak shave this first and avoid the new hook

weiguozhi: I will remove this command line option and the new hook.

But I think when we want to enable this kind of rematerialization on AArch64 or PowerPC, we still need something similar.

llvm/include/llvm/CodeGen/TargetInstrInfo.h
weiguozhi: > Instead of doing a dance to teach an X86 quirk to the rest of the infrastructure, can we solve that problem solely in x86?
> 
> * Remove the EFLAGS def in the MOV32r0 target description
> * Use the lowering that clobber eflags or not based on the liveness of eflags when we expand the pseudo instruction.
> 
> This would fix the rematerialization capabilities and keep the weirdness of dynamically changing the physical definitions to the x86 backend.

I'm trying to implement this method. But I got 200 test failures for the MOV32r0 change, all of them have similar pattern.
```
// Original instructions
; CHECK-NEXT:    xorl %ecx, %ecx    
; CHECK-NEXT:    subl {{[0-9]+}}(%esp), %eax    
; CHECK-NEXT:    cmovsl %ecx, %eax    
```
In the original instruction sequence, MOV32r0 clobbers EFLAGS, so it can't be put between the compare/conditional instruction pair.
```
        // New instructions
        subl    8(%esp), %ecx    
        movl    $0, %eax    
        cmovnsl %ecx, %eax    
```
In the new instruction sequence, MOV32r0 doesn't clobber EFLAGS, so it can be freely scheduled between the compare/conditional instruction pair, and the EFLAGS is not dead at this position, so it is expanded to "movl    $0, %eax". This instruction should have same performance, but with larger code size.

Is this behavior still OK?

weiguozhi:
I removed the target hook canRematerializeIgnorePhysRegDef.
I changed the x86 instruction MOV32r0 to not define EFLAGS.
It caused a lot of test failures. Most of them have the pattern
```
 ; CHECK-NEXT:    movzbl {{[0-9]+}}(%rsp), %esi
-; CHECK-NEXT:    xorl %r15d, %r15d
 ; CHECK-NEXT:    testb %sil, %sil
+; CHECK-NEXT:    movl $0, %r15d
 ; CHECK-NEXT:    je LBB0_14
 ; CHECK-NEXT:  LBB0_12: ## %cond_false.i
```
It is because now MOV32r0 doesn't change EFLAGS, then it can be scheduled between a pair of EFLAGS def/use instructions.

llvm/lib/CodeGen/CalcSpillWeights.cpp
```diff
@@ -146,17 +146,43 @@ bool VirtRegAuxInfo::isRematerializable(const LiveInterval &LI,
...
 
-bool VirtRegAuxInfo::allUsesAvailableAt(const MachineInstr *MI,
+bool VirtRegAuxInfo::allUsesAvailableAt(const MachineInstr *OrigMI,
                                         SlotIndex UseIdx,
                                         const LiveIntervals &LIS,
                                         const MachineRegisterInfo &MRI,
                                         const TargetInstrInfo &TII) {
-  SlotIndex OrigIdx = LIS.getInstructionIndex(*MI).getRegSlot(true);
+  SlotIndex OrigIdx = LIS.getInstructionIndex(*OrigMI).getRegSlot(true);
+  auto MBB = LIS.getMBBFromIndex(UseIdx);
```
arsenm, line 155: No auto

llvm/lib/CodeGen/CalcSpillWeights.cpp
```diff
@@ -146,17 +146,43 @@ bool VirtRegAuxInfo::isRematerializable(const LiveInterval &LI,
...
       continue;
 
+    if (!MO.readsReg()) {
+      if (MO.getReg().isVirtual())
+        continue;
+      // A physical register is defined here, and we allow rematerialization of
+      // an instruction with physical register definition, it must be dead at
+      // this position.
+      const TargetRegisterInfo *TRI = MRI.getTargetRegisterInfo();
+      if (MBB->computeRegisterLiveness(TRI, MO.getReg(), UseIt) ==
```
arsenm, line 179: You shouldn't be using computeRegisterLiveness in RA code, you have the full LiveIntervals available

llvm/lib/Target/X86/X86InstrInfo.cpp
```diff
@@ -6339,11 +6366,20 @@ static bool expandMOVSHP(MachineInstrBuilder &MIB, MachineInstr &MI,
...
   switch (MI.getOpcode()) {
-  case X86::MOV32r0:
-    return Expand2AddrUndef(MIB, get(X86::XOR32rr));
+  case X86::MOV32r0: {
+    if (MBB.computeRegisterLiveness(&TRI, X86::EFLAGS, MI) !=
+        MachineBasicBlock::LQR_Dead) {
+      MIB->setDesc(get(X86::MOV32ri));
+      MIB.addImm(0);
+      return true;
+    } else
```
arsenm, line 6379: No else after return

llvm/lib/Target/X86/X86InstrInfo.cpp
```diff
@@ -6339,11 +6366,20 @@ static bool expandMOVSHP(MachineInstrBuilder &MIB, MachineInstr &MI,
...
 
 bool X86InstrInfo::expandPostRAPseudo(MachineInstr &MI) const {
+  MachineBasicBlock &MBB = *MI.getParent();
   bool HasAVX = Subtarget.hasAVX();
   MachineInstrBuilder MIB(*MI.getParent()->getParent(), MI);
   switch (MI.getOpcode()) {
-  case X86::MOV32r0:
-    return Expand2AddrUndef(MIB, get(X86::XOR32rr));
+  case X86::MOV32r0: {
+    if (MBB.computeRegisterLiveness(&TRI, X86::EFLAGS, MI) !=
```
arsenm, line 6374: This isn't going to be that effective, and isn't really efficient

arsenm (commented):
If you're going to handle this in X86 code, doing in the post-RA pseudo expansion is probably the wrong choice. You'd probably want a pre-assignment optimization pass.

However I do think there is value in rematerializing the dead def case independently, but that doesn't require the target hook

PR #198854, opened 2026-05-20
[libc][NFC] Lowercase standard identifiers in YAML files

kaladron:
Update YAML files to use lowercase identifiers for standards.

In header.py, canonical identifiers for standards are explicitly defined in lowercase and mapped to their pretty names for display. This change ensures that all YAML files use the lowercase identifiers (posix, linux, bsd, gnu) expected by the header generation tool.

### DIFF
--- a/libc/include/arpa/inet.yaml
+++ b/libc/include/arpa/inet.yaml
@@ -10,19 +10,19 @@ objects: []
 functions:
   - name: htonl
     standards:
-      - POSIX
+      - posix
     return_type: uint32_t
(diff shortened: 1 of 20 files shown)


michaelrj-google (approved):
LGTM, we might want to add a check to headergen that detects incorrect standards.

vonosmas (approved):
LGTM, but I second what @michaelrj-google  is saying - I'm pretty sure we can enforce the correct spellings in header.py consumption of those files.

PR #198857, opened 2026-05-20
[bolt][test] Use non-GNU assembly directive

tambry:
`asm()` is a GNU extensions `__asm()` is supported in both standard-conforming and GNU mode.

### DIFF
--- a/bolt/test/AArch64/validate-secondary-entry-point.s
+++ b/bolt/test/AArch64/validate-secondary-entry-point.s
@@ -47,15 +47,16 @@ _bar:
 void _start() {}
 
 __attribute__((naked)) void foo() {
-  asm("ldr x16, .L_fnptr\n"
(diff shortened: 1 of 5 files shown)


paschalis-mpeis (approved):
Thanks @tambry.

While you are at it, can you apply the same in the below files too?

```
runtime/bolt-reserved.cpp
X86/Inputs/inlined.cpp
X86/Inputs/linenumber.cpp
X86/Inputs/double_jump.cpp
```

tambry:
> While you are at it, can you apply the same in the below files too?

Good point, also modified those.

paschalis-mpeis:
Thanks! Please apply the format change suggested on `bolt/test/X86/Inputs/double_jump.cpp` so clang-format stops complaining, and we should be good to go. :)

PR #198858, opened 2026-05-20
[lldb][test] Set explicit C standard for tests that require an older one

tambry:
These use identifiers which are keywords since C23. Once the default standard is changed we'd have to be explicit anyway.

### DIFF
--- a/lldb/test/API/lang/c/cpp_keyword_identifiers/Makefile
+++ b/lldb/test/API/lang/c/cpp_keyword_identifiers/Makefile
@@ -1,3 +1,4 @@
 C_SOURCES := main.c
+CFLAGS_EXTRAS := -std=c99
 
 include Makefile.rules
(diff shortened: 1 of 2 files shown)


DavidSpickett (approved):
If I understood correctly, the point of the tests is that lldb should not confuse the name of a variable in the program with a keyword in the expression language. Which I think is always c++ even for C programs. So you need to compile the test program for a version of C that allows these names as identifiers.

Given that, LGTM.

tambry:
@DavidSpickett Correct. These tests would break within a few years once the default C standard is bumped.

I've been using a local global Clang configurations with language-specific flags to track down such issues ahead of time, plus cases where the wrong frontend is used for compiling C/C++ code (e.g. #198855, #198856). There's only a few cases of these issues across the whole codebase, fortunately.

PR #198859, opened 2026-05-20
[lldb] Call SetSyntheticChildrenGenerated on GetSyntheticChildAtOffset

kastiglione:
Follow up to #197311 and based on the feedback on that PR.

One way a Python synthetic formatter can construct a child is with `SBValue::GetSyntheticChildAtOffset`. However those values were are not marked as having been generated by `SyntheticChildren`. This means the previous PR (#197311) did not apply to them as would be expected. The fix is to call `SetSyntheticChildrenGenerated(true)` on the child values created in `GetSyntheticChildAtOffset`.

Assisted-by: claude (tests and review)

### DIFF
--- a/lldb/test/API/python_api/sbvalue_get_parent/TestSBValueGetParent.py
+++ b/lldb/test/API/python_api/sbvalue_get_parent/TestSBValueGetParent.py
@@ -17,7 +17,18 @@ class TestCase(TestBase):
         frame = thread.GetFrameAtIndex(0)
 
         self._do_test(frame, "parent", "child")
-        self._do_test(frame, "vec", 0)
+        self._do_test(frame, "vec", 1)
+
+    def test_create_child_at_offset(self):
+        """Values returned by CreateChildAtOffset should return the synthetic parent from GetParent."""
+        self.build()
+        _, _, thread, _ = lldbutil.run_to_source_breakpoint(
+            self, "break here", lldb.SBFileSpec("main.cpp")
+        )
+        frame = thread.GetFrameAtIndex(0)
+
+        self.runCmd("command script import MyContainer_synthetic.py")
+        self._do_test(frame, "container", 1)
 
     def _do_test(
         self, frame: lldb.SBFrame, parent_name: str, child_key: Union[str, int]
--- a/lldb/test/API/python_api/sbvalue_get_parent/TestSBValueGetParent.py
+++ b/lldb/test/API/python_api/sbvalue_get_parent/TestSBValueGetParent.py
@@ -17,18 +17,11 @@ class TestCase(TestBase):
         frame = thread.GetFrameAtIndex(0)
 
         self._do_test(frame, "parent", "child")
-        self._do_test(frame, "vec", 1)
+        self._do_test(frame, "vec", 0)
 
-    def test_create_child_at_offset(self):
-        """Values returned by CreateChildAtOffset should return the synthetic parent from GetParent."""
-        self.build()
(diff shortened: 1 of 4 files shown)


jimingham:
SetSyntheticChildrenGenerated is a pretty confusingly named API.  If you took it literally it should mean something like "I have generated synthetic children", and then the code you added doesn't make sense.  After all, this is a Parent ValueObject that is generating a synthetic child to hand out, so it would make total sense to call SetSyntheticChildrenGenerated on the Parent VO that is generating the synthetic child.  But it seems odd to call it on the synthetic child you have just made because you don't know whether IT has or has not made any synthetic children.

But I am pretty sure that's not what this API (and the very awkward IsSyntheticChildrenGenerated equivalent) means.  Instead, what it really means is "there synthetic children somewhere in the VO path that ends up in me."  

For instance in the constructor for ValueObject that receives a parent, we do:

```
ValueObject::ValueObject(ValueObject &parent)
    : m_parent(&parent), m_update_point(parent.GetUpdatePoint()),
      m_manager(parent.GetManager()), m_id(++g_value_obj_uid) {
  m_flags.m_is_synthetic_children_generated =
      parent.m_flags.m_is_synthetic_children_generated;

```
so we're saying here "if your parent said there were synthetic children involved in this hierarchy, then you are part of that so you set the flag too."

Under that interpretation, it is correct to set the "SyntheticChildrenGenerated" on the child when realizing the synthetic child.

lldb/test/API/python_api/sbvalue_get_parent/TestSBValueGetParent.py
```diff
@@ -17,7 +17,18 @@ def test(self):
         frame = thread.GetFrameAtIndex(0)
 
         self._do_test(frame, "parent", "child")
-        self._do_test(frame, "vec", 0)
+        self._do_test(frame, "vec", 1)
```
DavidSpickett: Is there a specific reason you changed this from index 0 to index 1?

Does it make sure that the formatter had to do some work rather than taking a shortcut maybe?

lldb/test/API/python_api/sbvalue_get_parent/TestSBValueGetParent.py
```diff
@@ -17,7 +17,18 @@ def test(self):
...
+    def test_create_child_at_offset(self):
+        """Values returned by CreateChildAtOffset should return the synthetic parent from GetParent."""
+        self.build()
+        _, _, thread, _ = lldbutil.run_to_source_breakpoint(
+            self, "break here", lldb.SBFileSpec("main.cpp")
+        )
+        frame = thread.GetFrameAtIndex(0)
+
+        self.runCmd("command script import MyContainer_synthetic.py")
+        self._do_test(frame, "container", 1)
```
DavidSpickett: I'd merge this into the test above. Saves launching the debugee again.

The docstring can be a comment before the _do_test call.

DavidSpickett (approved):
I agree the name is strange. It's a thing that the parent calls on the newly created child, to tell the child that the parent created synthetic children? Anyway name aside I think I understand the intent.

LGTM

DavidSpickett:
Just removed my @ from the PR description because I'm pretty sure I would get mentioned a bunch when the commit ends up pushed somewhere else on Github.

lldb/test/API/python_api/sbvalue_get_parent/TestSBValueGetParent.py
kastiglione: no good reason, I reverted it.

PR #198861, opened 2026-05-20
[PowerPC][NFC] Change arguments of PPCPostRAExpPseudo/PseudoXFormMemOp

redstar:
The assembler string of the pseudo is almost always a # followed by the name of the pseudo. A good part of the pseudos does not have a pattern. Changing the order of arguments asmstr and pattern in PPCPostRAExpPseudo and PseudoXFormMemOp, and assigning default values, reduces repetitions.


redstar:
There are more pseudos to change. I find it too large to do it all in one PR.

lei137 (approved):

PR #193045, opened 2026-04-20
[libc++] Introduce implicit and explicit ABI annotations

philnik777:
This patch introduces `_LIBCPP_{BEGIN/END}_EXPLICIT_ABI_ANNOTATIONS` and marks everything within an `_LIBCPP_{BEGIN,END}_UNVERSIONED_NAMESPACE_STD` (and any derivatives like `_LIBCPP_{BEGIN,END}_NAMESPACE_STD`) implicitly by default. This allows us to drop `_LIBCPP_HIDE_FROM_ABI` in most of the code base, except for functions which shouldn't be `_LIBCPP_HIDE_FROM_ABI`.

This patch doesn't remove any `_LIBCPP_HIDE_FROM_ABI`s, since we have over 13k of them in the code base. Actually dropping them will happen over some time to avoid too many merge conflicts.

PR #198864, opened 2026-05-20
Add missing annotatinos for Apple platforms

ilovepi:
These seemed to be missed in #193045.


ilovepi:
* **#198864** <a href="https://app.graphite.com/github/pr/llvm/llvm-project/198864?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/llvm/llvm-project/198864?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/llvm?ref=merge-queue-instructions-in-stack-comment&prId=198864).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

libcxx/include/__functional/function.h
```diff
@@ -529,9 +529,15 @@ class __policy_func<_Rp(_ArgTypes...)> {
 
 #  if _LIBCPP_HAS_BLOCKS_RUNTIME
 
+_LIBCPP_DIAGNOSTIC_PUSH
```
philnik777: Why is this required?

philnik777 (commented):
Would it be possible to add a test to check that blocks at least compile? We have macOS builders, so the environment shouldn't be a problem.

libcxx/include/__functional/function.h
ilovepi: I think I just copied the pattern in the block above where the annotation was added for the other bits. I guess its probably not.

ilovepi:
> Would it be possible to add a test to check that blocks at least compile? We have macOS builders, so the environment shouldn't be a problem.

I'll take a stab at it, but since its coming from a macro we can't use a normal compiler reproducer and get the preprocessed file. The only place I've seen this is happening is in a pretty involved build to, so it may not be all that straightforward to come up with a test, and I don't have all that much time to devote to this beyond the quick fix I posted.

philnik777:
@ilovepi I'm pretty sure this should reproduce with any configuration that has a blocks runtime and includes `<functional>`.

ilovepi:
libcxx/utilities/function.objects/func.blocks.pass.cpp would probably catch it if Block.h was included first.

```cpp
// REQUIRES: has-blocks-runtime
// <functional>
// class function
// Test that including <Block.h> before <functional> compiles and runs.
// This is a regression test for an issue where redeclaring the Block runtime
// functions (_Block_copy, _Block_release) inside libc++'s explicit ABI
// annotations incorrectly added an 'abi_tag' attribute to them.

extern "C" void* _Block_copy(const void* aBlock);
// #include <Block.h>
#include <functional>

int main(int, char**) {
  std::function<void()> f = []() {};
  (void)f;
  return 0;
}
```

Should this be a new test file? if so, I guess next to func.blocks.pass.cpp is a good spot?

libcxx/test/libcxx/utilities/function.objects/block.func.compile.pass.cpp
```diff
@@ -0,0 +1,32 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// std::function support for the "blocks" extension
+
+// REQUIRES: has-blocks-runtime
```
philnik777: This seems wrong. I can't find any reference to such a flag.

philnik777 (changes requested):

libcxx/test/libcxx/utilities/function.objects/block.func.compile.pass.cpp
ilovepi: oh, I thought I removed that. I was looking for the feature and when I didn't find it, I thought I was going to have to add a new one to lit. But then I found the existing test and copied from there. When I copied I forgot to delete the wrong/stale bit here.

philnik777 (approved):

ilovepi:
hmm, not sure why this merged so quick w/ auto-merge? I thought that would wait for premerge tests. I'm pretty sure the failing test was unrelated to this pr, but this seems a bit odd. cc: @petrhosek @boomanaiden154

boomanaiden154:
> hmm, not sure why this merged so quick w/ auto-merge? I thought that would wait for premerge tests.

It waits for monorepo wide premerge checks (the `premerge.yaml` workflow), and specifically the x86 Linux and Windows jobs (AArch64 isn't required so it will merge without that finishing). Notably that does not include the libc++ workflows.

This is something I've wanted to fix for a while, but it's somewhat difficult.

ilovepi:
> > hmm, not sure why this merged so quick w/ auto-merge? I thought that would wait for premerge tests.
> 
> It waits for monorepo wide premerge checks (the `premerge.yaml` workflow), and specifically the x86 Linux and Windows jobs (AArch64 isn't required so it will merge without that finishing). Notably that does not include the libc++ workflows.
> 
> This is something I've wanted to fix for a while, but it's somewhat difficult.

TIL. I haven't seen anything reporting breakages, so I'm guessing its fine.

PR #198866, opened 2026-05-20
Silence "switch statement contains default but no case labels"; NFC

AaronBallman:
65f8a7c1c423d96597bdb14062a0d0eb27d7caf4 accidentally introduced some diagnostics from having a switch statement with a default label but no case labels. This removes the switch statements until we have more cases to add.


ilovepi (approved):
Thanks for catching this. I refactored this quite a bit trying to reland and it looks like I missed this.

PR #198867, opened 2026-05-20
[DWARFVerifier] Skip DW_TAG_LLVM_annotation in DWARFVerifier

augusto2112:
Annotations are not indexed, so we need to skip them on the verifier.

Assisted by: claude

### DIFF
--- a/llvm/test/tools/llvm-dwarfdump/X86/debug-names-verify-completeness-llvm-annotation.ll
+++ b/llvm/test/tools/llvm-dwarfdump/X86/debug-names-verify-completeness-llvm-annotation.ll
@@ -1,11 +1,20 @@
-; Verify that DW_TAG_LLVM_annotation DIEs are not required to appear in
-; .debug_names. 
+; Verify that DW_TAG_LLVM_annotation DIEs are not emitted into .debug_names,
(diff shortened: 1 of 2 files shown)


Michael137 (approved):

llvm/test/tools/llvm-dwarfdump/X86/debug-names-verify-completeness-llvm-annotation.ll
```diff
@@ -0,0 +1,28 @@
+; Verify that DW_TAG_LLVM_annotation DIEs are not required to appear in
+; .debug_names. 
+
```
Michael137, line 3: Do we have tests that assert that the annotations dont end up in the index?

adrian-prantl (approved):
Aside from Michael's question (which IIUC is orthogonal) this looks good.

llvm/test/tools/llvm-dwarfdump/X86/debug-names-verify-completeness-llvm-annotation.ll
augusto2112: I added it to this test.

PR #198869, opened 2026-05-20
[CIR] Fix bugs with array new in a ternary expression

andykaylor:
While attempting to compile Clang with CIR enabled, we ran into a problem with the cleanup handling for array new inside of a ternary operation. The reduced test case ended up showing different failure modes with and without exceptions enabled, but it failed in both cases. With exceptions enabled, we were failing to spill the value to be returned and reload it outside of the cleanup scope. With exceptions disabled, we were pushing an effectively empty entry on the EH stack and later asserting when trying to pop it.

This change fixes the spill and reload problem by inserting a RunCleanupsScope RAII object around the ternary branch generation code with a call to forceCleanups to generate the spill and reload.

The assertion failure is fixed by replacing the assert call with code to safely handle the condition of having an empty cleanup on the EH stack. We should probably modify the code that pushes EH-only cleanups to avoid doing so when exceptions are disabled, but that's a more extensive change and having the new handler as a safe backstop seems like a good idea.

Manual inspection revealed a second location where we're generating a cir.ternary operation that had the same spill and reload problem described above, so that is fixed here too.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenExpr.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExpr.cpp
@@ -2983,11 +2983,18 @@ CIRGenFunction::emitConditionalBlocks(const AbstractConditionalOperator *e,
     CIRGenFunction::LexicalScope lexScope{*this, loc, b.getInsertionBlock()};
     curLexScope->setAsTernary();
 
-    assert(!cir::MissingFeatures::incrementProfileCounter());
-    eval.beginEvaluation();
-    resultLV = branchGenFunc(*this, expr);
-    mlir::Value resultPtr = resultLV ? resultLV->getPointer() : mlir::Value();
-    eval.endEvaluation();
+    mlir::Value resultPtr;
+    {
+      // Emit any cleanups that were needed on this branch so we can spill
+      // and reload the return value.
+      CIRGenFunction::RunCleanupsScope branchCleanups(*this);
+      assert(!cir::MissingFeatures::incrementProfileCounter());
+      eval.beginEvaluation();
+      resultLV = branchGenFunc(*this, expr);
+      resultPtr = resultLV ? resultLV->getPointer() : mlir::Value();
+      eval.endEvaluation();
+      branchCleanups.forceCleanup({&resultPtr});
+    }
 
     if (resultPtr) {
       yieldTy = resultPtr.getType();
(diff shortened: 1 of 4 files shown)


clang/lib/CIR/CodeGen/CIRGenExpr.cpp
```diff
@@ -2983,11 +2983,18 @@ CIRGenFunction::emitConditionalBlocks(const AbstractConditionalOperator *e,
...
+    {
+      // Emit any cleanups that were needed on this branch so we can spill
+      // and reload the return value.
+      CIRGenFunction::RunCleanupsScope branchCleanups(*this);
+      assert(!cir::MissingFeatures::incrementProfileCounter());
+      eval.beginEvaluation();
+      resultLV = branchGenFunc(*this, expr);
+      resultPtr = resultLV ? resultLV->getPointer() : mlir::Value();
+      eval.endEvaluation();
+      branchCleanups.forceCleanup({&resultPtr});
```
erichkeane, line 2996: Does the DTor not do this?

andykaylor: The dtor would emit cleanups, but we need to call `forceCleanup` explicitly with a list of values to spill and reload to get the return value to escape the cleanup scope. This is a pattern we copied from classic codegen and we use it in a bunch of places to fix exactly the sort of problem that we're fixing here, but we need it in a few more places than classic codegen does because of our explicit cleanup scope regions vs. the branching that happens there.

The way this works is kind of surprising. The mlir::Value represented by `resultPtr` here is typically a live value defined inside the cleanup scope. When we pass it to `forceCleanup`, we create a new alloca outside the cleanup scope, spill the value of `resultPtr` to that alloca, then create load of the alloca outside of the cleanup scope and finally update resultPtr (which was passed by reference so we can do this) to be the load value instead of whatever it was before. This works because of MLIR's value semantics. Classic codegen accomplishes the same trick by passing a pointer-to-pointer-to-llvm::Value.

Before this PR, incidentally, we didn't get as far as this being reported as a dominance problem because the code on line 3001 was creating the YieldOp (which is supposed to yield the return value of the ternary op) inside the cleanup scope. The `RunCleanupsScope` object fixes that, but without the spill and reload it would have a dominance problem.

erichkeane: ah!  I see, thanks for the explanation.

erichkeane (approved):

PR #198871, opened 2026-05-20
[CIR] Handle atomic-fetch lowering of pointer types.

erichkeane:
We previously didn't properly handle the pointer type conversions, which resulted in a verifier error.  This patch does what classic codegen does for these: casts them to an integer type.  The rest of the logic around this in classic codegen isn't necessary as the llvm intrinsics now handle those types properly.


andykaylor (approved):
lgtm

PR #198873, opened 2026-05-20
Use existing encoding when writing to an existing file

perry-ca:
Clang lit test `Sema/warn-lifetime-safety-suggestions.cpp` is failing because the encoding for the new contents of the source file don't match the file tag on the file.

This change ensures the contents of a file are written in the same code page as the existing file.

- The copyFileTagAttributes() function will change the auto conversion to the code page of the file tag when writing a new version of a file.
- When writing to an existing file (eg. appending), use the file tag as the code page for auto conversion.  Don't assume 1047.

### DIFF
--- a/llvm/lib/Support/AutoConvert.cpp
+++ b/llvm/lib/Support/AutoConvert.cpp
@@ -82,12 +82,24 @@ int enablezOSAutoConversion(int FD) {
   return fcntl(FD, F_CONTROL_CVT, &Query);
 }
 
-std::error_code llvm::setzOSFileTag(int FD, int CCSID, bool Text) {
-  assert((!Text || (CCSID != FT_UNTAGGED && CCSID != FT_BINARY)) &&
+int enablezOSAutoConversionCcsid(int FD, int ccsid) {
+  struct f_cnvrt cvt = {
+      SETCVTALL,    // cvtcmd
+      CCSID_UTF_8,  // pccsid
+      0,            // fccsid
+  };
+  if (ccsid == FT_UNTAGGED)
+    return -1;
+  cvt.fccsid = ccsid;
+  return fcntl(FD, F_CONTROL_CVT, &cvt);
+}
+
+std::error_code llvm::setzOSFileTag(int FD, int CCSID, bool IsText) {
+  assert((!IsText || (CCSID != FT_UNTAGGED && CCSID != FT_BINARY)) &&
          "FT_UNTAGGED and FT_BINARY are not allowed for text files");
   struct file_tag Tag;
   Tag.ft_ccsid = CCSID;
-  Tag.ft_txtflag = Text;
+  Tag.ft_txtflag = IsText;
   Tag.ft_deferred = 0;
   Tag.ft_rsvflags = 0;
 
@@ -143,6 +155,10 @@ std::error_code llvm::copyFileTagAttributes(const std::string &Source,
   struct stat SourceAttributes;
   if (stat(Source.c_str(), &SourceAttributes) == -1)
     return std::error_code(errno, std::generic_category());
+
+  if (SourceAttributes.st_tag.ft_txtflag)
+    if (enablezOSAutoConversionCcsid(DestinationFD, SourceAttributes.st_tag.ft_ccsid) == -1)
+      return errnoAsErrorCode();
 
   return setzOSFileTag(DestinationFD, SourceAttributes.st_tag.ft_ccsid,
                        SourceAttributes.st_tag.ft_txtflag);
--- a/llvm/lib/Support/AutoConvert.cpp
+++ b/llvm/lib/Support/AutoConvert.cpp
@@ -84,9 +84,9 @@ int enablezOSAutoConversion(int FD) {
 
 int enablezOSAutoConversionCcsid(int FD, int ccsid) {
   struct f_cnvrt cvt = {
-      SETCVTALL,    // cvtcmd
-      CCSID_UTF_8,  // pccsid
-      0,            // fccsid
+      SETCVTALL,   // cvtcmd
+      CCSID_UTF_8, // pccsid
+      0,           // fccsid
   };
   if (ccsid == FT_UNTAGGED)
     return -1;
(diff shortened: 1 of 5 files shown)


llvm/lib/Support/Unix/Path.inc
cachemeifyoucan, line 1163: auto convert is no-op outside windows. Why do you need to update this file?

llvm/lib/Support/raw_ostream.cpp
cachemeifyoucan, line 1036: Not sure why this order matters?

llvm/lib/Support/AutoConvert.cpp
cachemeifyoucan, line 85: Can this function be unified into `enablezOSAutoConversion`?

cachemeifyoucan (commented):
I don't know the encoding specific on windows but other parts need some improvement.

also please add a unit-test.

llvm/lib/Support/Unix/Path.inc
perry-ca: Auto convert is used on z/OS.  This part of the function is used on z/OS to set the encoding for the file being read or written to.   The change here ensures the output is written in the same code page as the file being updated/appended.

llvm/lib/Support/raw_ostream.cpp
perry-ca: The z/OS docs state that the auto conversion can only be set before the first read/write.  Things worked as they were.  I swapped the order to ensure future changes can't accidentally insert a write between the creation of the ostream and setting the auto conversion.  I can add a comment if you like.

llvm/lib/Support/Unix/Path.inc
cachemeifyoucan: Sorry, I misread `__MVS__` as `__MSVC__`. Can this block under the same guard as well?

llvm/lib/Support/raw_ostream.cpp
cachemeifyoucan: Yes please

llvm/lib/Support/AutoConvert.cpp
perry-ca: I had thought of this.  I would need additional args to skip the query and adjustment of the settings.   That makes up the majority of the function.  I think it's cleaner as two separate functions.

llvm/lib/Support/Unix/Path.inc
perry-ca: Sorry, I'm not sure which guard you are referring to when you say "same guard".  The code I added is inside the same `#ifdef __MVS__` block as the existing code.  I don't think you're asking to use `__MSVC__` instead of `__MVS__`.  I think I'm overlooking something.

perry-ca:
>also please add a unit-test.

This fixes an existing clang lit test.  I should have included that in the description.  I've corrected that.

llvm/lib/Support/Unix/Path.inc
cachemeifyoucan: Ok. The guard is actually very far away. My bad. Just add a test and I don't have other comments about this.

cachemeifyoucan:
> > also please add a unit-test.
> 
> This fixes an existing clang lit test. I should have included that in the description. I've corrected that.

I think it is still good to have a basic unit test in LLVM in SupportTests.

llvm/unittests/Support/raw_ostream_test.cpp
```diff
@@ -508,14 +509,14 @@ TEST(raw_ostreamTest, raw_fd_ostream_mutual_ties) {
   SmallString<64> PathTiedTo;
   int FDTiedTo;
   ASSERT_FALSE(
-      sys::fs::createTemporaryFile("tietest1", "", FDTiedTo, PathTiedTo));
+      sys::fs::createTemporaryFile("tietest1", "", FDTiedTo, PathTiedTo, sys::fs::OF_Text));
```
cachemeifyoucan, line 512: Just for my understanding, why the OpenFlag need to be updated to be `OF_Text`? That seems to be unrelated to your commit message.

perry-ca: Thanks for seeing this and reviewing so quick.  You beat me to the explanation.  I was going to provide the answer in a comment in the PR.

This is a to match the change in checkFileData() to open the file as text.  All of the places that create a temporary file and then use checkFileData() need to create the file as a text file.

cachemeifyoucan: I will avoid changing existing test coverage that is `OF_None`.

perry-ca: >I will avoid changing existing test coverage that is OF_None.

I'm not sure I follow what you're saying/asking in this comment.  Could you expand on your comment.  Thanks

I did double check a few things.
- the place where OF_None is used to create raw_fd_ostream should probably use OF_Text, but no I/O is done so this doesn't really matter.
- if the temporary files are created with OF_Text they are created as binary files on z/OS.  The tests in the file are writing & reading text so they should be text.

We are in the process of upstreaming the changes to get the SystemTest unit test to pass on z/OS.  We need to upstream some changes to gtest first.  For this PR I want to make sure the new test I added works and I don't destablize any of the others on z/OS.  And making sure nothing files on other platforms.

abhina-sree: Just to clarify, the OF_Text flag vs OF_None flag only makes a difference on the z/OS platform AFAIK. I created a separate flag called OF_TextWithCRLF a while back for Windows platform to use because they did not want an extra control line feed for all text files. So there shouldn't be any other platforms affected by changing OF_None to OF_Text if it truly is a text file

cachemeifyoucan: I don't want to make assumptions about platforms since I don't know all the platforms we support. Is `OF_None` a problem on zOS? If it passes currently, I prefer not to change the test if not needed. If it currently fails on zOS and it needs to be `OF_Text`, then the change is inevitable, just document the difference on zOS.

perry-ca: I'm not sure what kind of change you are asking for.  The code in the unit test matches the way these llvm functions (eg. createTemporaryFile() & getFileOrSTDIN()) are used in clang/llvm.  

If you can elaborate, I'll make the change.  Thanks

perry-ca: @cachemeifyoucan, gentle ping.  Thanks

cachemeifyoucan (approved):
I don't have further comment. LGTM

PR #198874, opened 2026-05-20
Add InstCount Pass Before Optimization

InakiVA:
This way we can count instructions before the optimization pipeline for analysis sake


boomanaiden154 (commented):
What exactly about this has changed to ensure we do not hit compile timeouts?

It would also be nice if we can split out adding the pass option into a separate patch and then add it to the pipeline in a different patch so that we can revert the enablement more cleanly.

InakiVA:
> What exactly about this has changed to ensure we do not hit compile timeouts?

The pass creating the compilation timeouts was function properties analysis pass, not inscount. The previous iteration had both added as one. Right now this is just one and I'm working on fixing whatever issues func had to add it later

llvm/test/Analysis/InstCount/instcount.ll
```diff
@@ -72,4 +66,4 @@ declare void @g()
 declare void @h()
 declare void @k()
 declare void @l()
-declare void @m()
+declare void @m()
```
aeubanks: shouldn't remove newline

aeubanks (approved):
I'm not sure separating out the pre-opt change and the pipeline change will make anything easier/harder to revert, I think this is fine

lgtm

PR #198875, opened 2026-05-20
[mlir][ROCDL] Move ROCDL intrinsic enum immargs to enums

krzysz00:
In many cases, a "i32" `immarg` arguhment to an intrinsic in the AMDGPU backend actually corresponds directly to some enumerated set of values in the backend, which we have to smuggle through an I32. This makes the MLIR forms of intrinsics less readable and means that people either have to use the `amdgpu` dialect to get these enums or have to roll their own enums if they want to know what's going on.

This PR rips the band-aid off and breaks the world by swapping out those integer attributes for enum attributes.

Of special note is the handling of the aux/cachepolicy field on various intrinsics; in the backend, all the architectures share an enum and you've just got to use the right names in the right spots. Here, we've separated out the cases for pre-gfx942, gfx942+, and gfx12 enums as separate attributes (including separate casing for gfx12 atomics) and allowed any of them to be used. We also allow an I32Attr in those arguments for easy importing and to make the common case of "0" portably printable.

In all other cases, we've just moved to the new enums.

The WMMA ops, which have multiple immadiate arguments that could be confusing, have been given assembly formats than name the filed names.

This commit also moves the AMDGPU dialct to use the new enums.

In order to be able to use the current `EnumAttr` system, a method has been added to LLVM translation to allow such enum attributes to be converted to their underlying integer constants.

AI usage: this is a quasi-mechanical refactor that was pretty much farmed out to Codex


mlir/include/mlir/Dialect/LLVMIR/ROCDLEnums.td
```diff
@@ -0,0 +1,280 @@
...
+      ROCDL_WMMACModifierAbs,
+      ROCDL_WMMACModifierNegAbs
+    ]> {
+  let printBitEnumPrimaryGroups = 1;
+  let printBitEnumQuoted = 0;
+}
+
+def ROCDL_SchedGroupMaskNone : I32BitEnumCaseNone<"none">;
+def ROCDL_SchedGroupMaskNonMemNonSideEffect
+    : I32BitEnumCaseBit<"non_mem_non_sideffect", 0>;
```
Meinersbur:
```suggestion
    : I32BitEnumCaseBit<"non_mem_non_sideeffect", 0>;
```
the typo exited before already, can it be fixed?

Meinersbur (commented):
Typos in summary: arguhment, hhandling, dialct , immadiate , curernt

mlir/include/mlir/Dialect/LLVMIR/ROCDLAttrs.td
```diff
@@ -0,0 +1,84 @@
...
+      ROCDL_PreGfx12CachePolicyAttr, ROCDL_Gfx942CachePolicyAttr,
+      ROCDL_Gfx12CachePolicyAttr, ROCDL_Gfx12AtomicCachePolicyAttr, I32Attr],
+    "AMDGPU cache policy attribute"> {
+  let storageType = "::mlir::Attribute";
+  let returnType = "::mlir::Attribute";
+  let convertFromStorage = "$_self";
+  let constBuilderCall = "$0";
+}
+
+def ROCDL_DefaultCachePolicyAttr : DefaultValuedAttr<ROCDL_CachePolicyAttr,
```
efric: the use of this feels a bit too permissive right now; every relevant op is just using this, but doesn't that mean any op can use any cache policy family? why not just have each op use their respective/relevant policy family?

efric (commented):
I like the direction; just a concern:

mlir/include/mlir/Dialect/LLVMIR/ROCDLAttrs.td
krzysz00: A lot of these ops (mainly the buffer ones) are valid across architectures

We could maybe forbid gfx9 cache policies on some of the new ontrinsitcs?

efric: i sent the agents to look for issues with this

for prefetch, we could technically represent these https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/AMDGPU/IR/AMDGPUOps.cpp#L1521-L1523 right now which we shouldn't for consistency 

also for atomics; we should be careful with  

```
  gfx12<nt>

  means bit 0 in the normal GFX12 cache-policy enum.

  But for GFX12 atomic policy, bit 0 means:

  gfx12_atomic<return>
```

krzysz00: For prefetch, this is a low-level enough operation that I think some degree of "don't set the bit that doesn't do anything" is reasonable - or at least we know the compiler will drop that.

For atomics ... the `return` bit is handled by the backend, we don't actually need to set it, and I know at least one version of this PR has atomics and non-atomics on separate enums. I should definitely tighten up the verifiers here.

efric (approved):
LGTM, thanks!

PR #198877, opened 2026-05-20
[AMDGPU] Rewrite `-march` to `-mcpu` in the AMDGPU Toolchain

jhuber6:
Summary:
Pretty much every target uses either `-mcpu` or `-march` consistently.
AMDGPU has been accidentally using both for a while, mostly from some
fallout with the OpenMP Toolchain. This is too deep to pull out without
potentially disrupting users, but I want to at least contain it by
canonicalizing `-march` to `-mcpu` in the driver. This means we don't
need to bother checking both like every other target does.

### DIFF
--- a/clang/lib/Driver/ToolChains/AMDGPU.cpp
+++ b/clang/lib/Driver/ToolChains/AMDGPU.cpp
@@ -659,11 +659,7 @@ void amdgpu::getAMDGPUTargetFeatures(const Driver &D,
                                      std::vector<StringRef> &Features) {
   // Add target ID features to -target-feature options. No diagnostics should
   // be emitted here since invalid target ID is diagnosed at other places.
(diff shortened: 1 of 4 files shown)


yxsamliu:
do we need some negative lit tests?

jhuber6:
> do we need some negative lit tests?

Not much to negatively test, this begrudgingly accepts that you can use both `-march` and `-mcpu` for AMDGPU (I don't think this was ever done intentionally) but at least makes it to where we don't need to check both internally. There are existing tests that use `-march`, some OpenCL tests, and a lot of the OpenMP ones, so I think that's sufficient.

yxsamliu (commented):
LGTM. Thanks

yxsamliu (approved):

PR #198881, opened 2026-05-20
[clang] Allow clang only write module/pch when output is different

cachemeifyoucan:
Add an parameter to the output file creation function in clang to allow
it to configure output backend to only write output if the file is
different from the file on the disk. This is useful in a legacy build
system where the incremental build state is tracked by timestamp. For
example, producing the same clang module in the build will not update
the timestamp, thus downstream jobs do not need to be re-run.

The default value is not change so NFI is expected. This is only to
allow clients need such behavior to enable when needed.

Patched-by: hjyamauchi

### DIFF
--- a/clang/include/clang/Frontend/CompilerInstance.h
+++ b/clang/include/clang/Frontend/CompilerInstance.h
@@ -805,7 +805,7 @@ public:
   std::unique_ptr<raw_pwrite_stream> createDefaultOutputFile(
       bool Binary = true, StringRef BaseInput = "", StringRef Extension = "",
       bool RemoveFileOnSignal = true, bool CreateMissingDirectories = false,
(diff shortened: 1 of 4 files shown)


hjyamauchi (approved):

ojhunt (approved):
I'd be happier if this was an enum rather than another, especially due to the createDefaultOutputFile interface, but I know it's not clang style in general so I'm not going to force it.

cachemeifyoucan:
> I'd be happier if this was an enum rather than another, especially due to the createDefaultOutputFile interface, but I know it's not clang style in general so I'm not going to force it.

We can technically reuse the `vfs::OutputConfig` in clang interface but let's do that in the future. Keep this incremental. Merge current version for now.

PR #198884, opened 2026-05-20
[lldb] Ensure libunwind architecture matches test for TestLibUnwindRetInjection.py

bulbazord:
If the test is run arm64e while the just-built libunwind is arm64 only, the test will not function correctly.


medismailben (approved):
LGTM! Thanks!

PR #198886, opened 2026-05-20
[DAGCombiner] Redestribute NUW when both adds have it

AZero13:


jayfoad:
Can you write a test for this?

An Alive2 proof would be nice too.

PR #198887, opened 2026-05-20
[Matrix][HLSL] Add codegen support for Matrix Layout keywords

### LINKED ISSUE #192262: [HLSL][Matrix] Update all locations where we check Default Matrix Memory Layout to instead first check for a MatrixLayout Attribute.
Find all Invocation.getLangOpts().getDefaultMatrixMemoryLayout() in code base and add alternative lookup for memory layout meta data so we know if we should be doing a column major or row major  load or store.

The change itself looks to be easy, but has a very large test burden.

Add a helper that could look like this 
```cpp
// In a shared header or CodeGenFunction
bool isMatrixRowMajor(const Expr *E) {
  // Check if the underlying decl has an explicit layout attribute
  if (const auto *DRE = dyn_cast<DeclRefExpr>(E->IgnoreParenImpCasts())) {
    if (const auto *Attr = DRE->getDecl()->getAttr<HLSLMatrixLayoutAttr>())
      return Attr->getSemanticSpelling() == HLSLMatrixLayoutAttr::Keyword_row_major;
  }
  // Fall back to the global default
  return getLangOpts().getDefaultMatrixMemoryLayout() ==
         LangOptions::MatrixMemoryLayout::RowMajor;
}
```

 All 14+ sites in CGExpr.cpp, CGExprScalar.cpp, CGExprConstant.cpp, CGHLSLBuiltins.cpp, and CodeGenTypes.cpp that currently read getLangOpts().getDefaultMatrixMemoryLayout() need to be updated to first check for an HLSLMatrixLayoutAttr on the relevant declaration before falling back to the global default. A helper function like isMatrixRowMajor(const Expr *E) would centralize this logic.

Below are the areas in Codgen that needed to be updated.

| File:Line                | Purpose                         |
|--------------------------|---------------------------------|
| [CGExpr.cpp:2339]        | Matrix element l-value access   |
| [CGExpr.cpp:2596]        | Matrix subscript store          |
| [CGExpr.cpp:2904]        | Matrix subscript load           |
| [CGExpr.cpp:5191]        | Matrix element member expr      |
| [CGExpr.cpp:7401]        | Matrix element store            |
| [CGExprScalar.cpp:2242]  | Matrix cast                     |
| [CGExprScalar.cpp:2269]  | Matrix truncation cast          |
| [CGExprScalar.cpp:2355]  | Matrix multiply                 |
| [CGExprScalar.cpp:2595]  | Matrix binary ops               |
| [CGExprScalar.cpp:3175]  | Matrix compound assign          |
| [CGExprConstant.cpp:2648]| Constant matrix init list       |
| [CGHLSLBuiltins.cpp:1185]| HLSL mul builtin                |
| [CGHLSLBuiltins.cpp:1238]| HLSL transpose builtin          |
| [CodeGenTypes.cpp:115]   | Matrix type → LLVM IR type      |

[CGExpr.cpp:2339]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExpr.cpp#L2339
[CGExpr.cpp:2596]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExpr.cpp#L2596
[CGExpr.cpp:2904]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExpr.cpp#L2904
[CGExpr.cpp:5191]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExpr.cpp#L5191
[CGExpr.cpp:7401]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExpr.cpp#L7401

[CGExprScalar.cpp:2242]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExprScalar.cpp#L2242
[CGExprScalar.cpp:2269]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExprScalar.cpp#L2269
[CGExprScalar.cpp:2355]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExprScalar.cpp#L2355
[CGExprScalar.cpp:2595]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExprScalar.cpp#L2595
[CGExprScalar.cpp:3175]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExprScalar.cpp#L3175

[CGExprConstant.cpp:2648]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGExprConstant.cpp#L2648

[CGHLSLBuiltins.cpp:1185]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGHLSLBuiltins.cpp#L1185
[CGHLSLBuiltins.cpp:1238]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CGHLSLBuiltins.cpp#L1238

[CodeGenTypes.cpp:115]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/CodeGen/CodeGenTypes.cpp#L115

#### Sema Files (semantic checks)

Below are the areas in Sema that needed to be updated.

| File:Line                   | Purpose                          |
|-----------------------------|----------------------------------|
| [SemaChecking.cpp:16978]    | Init list element order check    |
| [SemaChecking.cpp:17102]    | Init list element order check    |

[SemaChecking.cpp:16978]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/Sema/SemaChecking.cpp#L16978
[SemaChecking.cpp:17102]: https://github.com/llvm/llvm-project/blob/5f62bae5666c3cad5439587fa0f330b92467241a/clang/lib/Sema/SemaChecking.cpp#L17102

farzonl:
fixes #192262
    
- Wrap Matrix Type in a row or column major layout attribute
- Add Helper to know which Matrix Layout to apply in codegen or check for in Sema
- Remove the Decl Atribute and only store on the type.

Assisted by Claud Opus 4.7

### DIFF
--- a/clang/include/clang/Basic/Attr.td
+++ b/clang/include/clang/Basic/Attr.td
@@ -5305,7 +5305,6 @@ def HLSLColumnMajor : TypeAttr {
   let LangOpts = [HLSL];
   let Documentation = [HLSLMatrixLayoutDocs];
 }
-def : MutualExclusions<[HLSLRowMajor, HLSLColumnMajor]>;
 
 def RandomizeLayout : InheritableAttr {
   let Spellings = [GCC<"randomize_layout">];
--- a/clang/include/clang/Basic/Attr.td
+++ b/clang/include/clang/Basic/Attr.td
@@ -5299,6 +5299,22 @@ def HLSLMatrixLayout : InheritableAttr {
   let Documentation = [HLSLMatrixLayoutDocs];
 }
 
+// Internal type-attribute markers attached to matrix QualTypes by Sema when
+// a decl carries `[[hlsl::row_major]]` / `[[hlsl::column_major]]`. They let
+// CodeGen / const-eval recover the layout from any matrix-typed expression
+// without re-walking back to the source decl.
+def HLSLRowMajor : TypeAttr {
+  let Spellings = [];
+  let LangOpts = [HLSL];
+  let Documentation = [InternalOnly];
+}
+
+def HLSLColumnMajor : TypeAttr {
+  let Spellings = [];
+  let LangOpts = [HLSL];
+  let Documentation = [InternalOnly];
+}
+
 def RandomizeLayout : InheritableAttr {
   let Spellings = [GCC<"randomize_layout">];
   let Subjects = SubjectList<[Record]>;
--- a/clang/include/clang/Basic/Attr.td
+++ b/clang/include/clang/Basic/Attr.td
@@ -5292,28 +5292,20 @@ def HLSLVkLocation : HLSLAnnotationAttr {
   let Documentation = [HLSLVkLocationDocs];
 }
 
-def HLSLMatrixLayout : InheritableAttr {
-  let Spellings = [CustomKeyword<"row_major">, CustomKeyword<"column_major">];
-  let Subjects = SubjectList<[Var, TypedefName, Field, Function]>;
+// `row_major` / `column_major` are HLSL keywords that select the in-memory
+// layout of a matrix-typed declaration. 
+def HLSLRowMajor : TypeAttr {
+  let Spellings = [CustomKeyword<"row_major">];
   let LangOpts = [HLSL];
   let Documentation = [HLSLMatrixLayoutDocs];
 }
 
-// Internal type-attribute markers attached to matrix QualTypes by Sema when
-// a decl carries `[[hlsl::row_major]]` / `[[hlsl::column_major]]`. They let
-// CodeGen / const-eval recover the layout from any matrix-typed expression
-// without re-walking back to the source decl.
-def HLSLRowMajor : TypeAttr {
-  let Spellings = [];
-  let LangOpts = [HLSL];
-  let Documentation = [InternalOnly];
-}
-
 def HLSLColumnMajor : TypeAttr {
-  let Spellings = [];
+  let Spellings = [CustomKeyword<"column_major">];
   let LangOpts = [HLSL];
-  let Documentation = [InternalOnly];
+  let Documentation = [HLSLMatrixLayoutDocs];
 }
+def : MutualExclusions<[HLSLRowMajor, HLSLColumnMajor]>;
 
 def RandomizeLayout : InheritableAttr {
   let Spellings = [GCC<"randomize_layout">];
--- a/clang/include/clang/Basic/Attr.td
+++ b/clang/include/clang/Basic/Attr.td
@@ -5303,6 +5303,22 @@ def HLSLMatrixLayout : InheritableAttr {
   let Documentation = [HLSLMatrixLayoutDocs];
 }
 
+// Internal type-attribute markers attached to matrix QualTypes by Sema when
+// a decl carries `[[hlsl::row_major]]` / `[[hlsl::column_major]]`. They let
+// CodeGen / const-eval recover the layout from any matrix-typed expression
+// without re-walking back to the source decl.
+def HLSLRowMajor : TypeAttr {
+  let Spellings = [];
+  let LangOpts = [HLSL];
+  let Documentation = [InternalOnly];
+}
+
+def HLSLColumnMajor : TypeAttr {
+  let Spellings = [];
+  let LangOpts = [HLSL];
+  let Documentation = [InternalOnly];
+}
+
 def RandomizeLayout : InheritableAttr {
   let Spellings = [GCC<"randomize_layout">];
   let Subjects = SubjectList<[Record]>;
--- a/clang/lib/CodeGen/CGExpr.cpp
+++ b/clang/lib/CodeGen/CGExpr.cpp
@@ -2361,8 +2361,8 @@ LValue CodeGenFunction::EmitMatrixElementExpr(const MatrixElementExpr *E) {
   // getEncodedElementAccess returns row-major linearized indices
   // If the matrix memory layout is column-major, convert indices
   // to column-major indices.
-  bool IsColMajor = isMatrixColumnMajor(getLangOpts(), E->getBase()->getType());
-  if (IsColMajor) {
+  bool IsRowlMajor = isMatrixRowMajor(getLangOpts(), E->getBase()->getType());
+  if (!IsRowlMajor) {
     const auto *MT = E->getBase()->getType()->castAs<ConstantMatrixType>();
     unsigned NumCols = MT->getNumColumns();
     for (uint32_t &Idx : Indices) {
--- a/clang/lib/CodeGen/CGExpr.cpp
+++ b/clang/lib/CodeGen/CGExpr.cpp
@@ -2361,8 +2361,8 @@ LValue CodeGenFunction::EmitMatrixElementExpr(const MatrixElementExpr *E) {
   // getEncodedElementAccess returns row-major linearized indices
   // If the matrix memory layout is column-major, convert indices
   // to column-major indices.
-  bool IsRowlMajor = isMatrixRowMajor(getLangOpts(), E->getBase()->getType());
-  if (!IsRowlMajor) {
+  bool IsRowMajor = isMatrixRowMajor(getLangOpts(), E->getBase()->getType());
+  if (!IsRowMajor) {
     const auto *MT = E->getBase()->getType()->castAs<ConstantMatrixType>();
     unsigned NumCols = MT->getNumColumns();
     for (uint32_t &Idx : Indices) {
--- a/clang/lib/CodeGen/CGExpr.cpp
+++ b/clang/lib/CodeGen/CGExpr.cpp
@@ -31,6 +31,7 @@
 #include "clang/AST/DeclObjC.h"
 #include "clang/AST/Expr.h"
 #include "clang/AST/InferAlloc.h"
+#include "clang/AST/MatrixUtils.h"
 #include "clang/AST/NSAPI.h"
 #include "clang/AST/ParentMapContext.h"
 #include "clang/AST/StmtVisitor.h"
@@ -2360,8 +2361,7 @@ LValue CodeGenFunction::EmitMatrixElementExpr(const MatrixElementExpr *E) {
   // getEncodedElementAccess returns row-major linearized indices
   // If the matrix memory layout is column-major, convert indices
   // to column-major indices.
-  bool IsColMajor = getLangOpts().getDefaultMatrixMemoryLayout() ==
-                    LangOptions::MatrixMemoryLayout::MatrixColMajor;
+  bool IsColMajor = isMatrixColumnMajor(getLangOpts(), E->getBase()->getType());
   if (IsColMajor) {
     const auto *MT = E->getBase()->getType()->castAs<ConstantMatrixType>();
     unsigned NumCols = MT->getNumColumns();
@@ -2617,8 +2617,7 @@ RValue CodeGenFunction::EmitLoadOfLValue(LValue LV, SourceLocation Loc) {
         ColIdx = ColConstsIndices->getAggregateElement(Col);
       else
         ColIdx = llvm::ConstantInt::get(Row->getType(), Col);
-      bool IsMatrixRowMajor = getLangOpts().getDefaultMatrixMemoryLayout() ==
-                              LangOptions::MatrixMemoryLayout::MatrixRowMajor;
+      bool IsMatrixRowMajor = isMatrixRowMajor(getLangOpts(), MatTy);
       llvm::Value *EltIndex =
           MB.CreateIndex(Row, ColIdx, NumRows, NumCols, IsMatrixRowMajor);
       llvm::Value *Elt = Builder.CreateExtractElement(MatrixVec, EltIndex);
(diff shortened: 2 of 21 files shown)


clang/lib/Sema/SemaTemplateInstantiateDecl.cpp
```diff
@@ -1044,13 +1044,6 @@ void Sema::InstantiateAttrs(const MultiLevelTemplateArgumentList &TemplateArgs,
       continue;
     }
 
-    if (auto *A = dyn_cast<HLSLMatrixLayoutAttr>(TmplAttr)) {
```
farzonl, line 1047: this was nice we get template instantiation for free now.

clang/lib/CodeGen/CGHLSLBuiltins.cpp
```diff
@@ -1221,13 +1223,14 @@ Value *CodeGenFunction::EmitHLSLBuiltinExpr(unsigned BuiltinID,
...
       Op0 = MB.CreateRowMajorToColumnMajorTransform(Op0, Rows0, Cols0);
+    if (IsRowMajorMat1)
       Op1 = MB.CreateRowMajorToColumnMajorTransform(Op1, Rows1, Cols1);
-    }
+
     Value *Result =
         MB.CreateMatrixMultiply(Op0, Op1, Rows0, Cols0, Cols1, "hlsl.mul");
-    if (IsRowMajor)
+    if (IsRowMajorMat0 || IsRowMajorMat1)
       Result = MB.CreateColumnMajorToRowMajorTransform(Result, Rows0, Cols1);
```
Icohedron: This is incorrect. The matrix multiply always returns a column-major matrix. We want to transform it to a row-major matrix if-and-only-if the result type is a row-major matrix (either due to the result being annotated with `row_major` or the default matrix layout is row-major).

The correct logic should be something like:
```suggestion
    bool IsResultRowMajor = isMatrixRowMajor(getLangOpts(), E->getType());
    if (IsResultRowMajor)
      Result = MB.CreateColumnMajorToRowMajorTransform(Result, Rows0, Cols1);
```

clang/lib/CodeGen/CGExprScalar.cpp
Icohedron: Can (the destination) `MatTy` have a different matrix layout than the `SrcMatTy` in matrix truncation? If so, this logic is incorrect. There should be separate `IsDstRowMajor` and `IsSrcRowMajor` booleans.

clang/lib/CodeGen/CGExpr.cpp
```diff
@@ -2363,9 +2364,8 @@ LValue CodeGenFunction::EmitMatrixElementExpr(const MatrixElementExpr *E) {
   // getEncodedElementAccess returns row-major linearized indices
   // If the matrix memory layout is column-major, convert indices
   // to column-major indices.
-  bool IsColMajor = getLangOpts().getDefaultMatrixMemoryLayout() ==
-                    LangOptions::MatrixMemoryLayout::MatrixColMajor;
-  if (IsColMajor) {
+  bool IsRowlMajor = isMatrixRowMajor(getLangOpts(), E->getBase()->getType());
+  if (!IsRowlMajor) {
```
Icohedron: typo:
```suggestion
  bool IsRowMajor = isMatrixRowMajor(getLangOpts(), E->getBase()->getType());
  if (!IsRowMajor) {
```

clang/lib/CodeGen/CGExprScalar.cpp
```diff
@@ -2247,8 +2248,8 @@ Value *ScalarExprEmitter::VisitMatrixSingleSubscriptExpr(
 
   for (unsigned Col = 0; Col != NumColumns; ++Col) {
     Value *ColVal = llvm::ConstantInt::get(RowIdx->getType(), Col);
-    bool IsMatrixRowMajor = CGF.getLangOpts().getDefaultMatrixMemoryLayout() ==
-                            LangOptions::MatrixMemoryLayout::MatrixRowMajor;
+    bool IsMatrixRowMajor =
+        isMatrixRowMajor(CGF.getLangOpts(), E->getBase()->getType());
```
Icohedron: nit: While you're here touching this code, maybe it's a good idea to hoist it out of the loop since it is loop-invariant.

clang/lib/CodeGen/CGExprScalar.cpp
```diff
@@ -3180,8 +3179,11 @@ Value *ScalarExprEmitter::VisitCastExpr(CastExpr *CE) {
       assert(SrcMatTy && "Source type must be a matrix type.");
       assert(NumRows <= SrcMatTy->getNumRows());
       assert(NumCols <= SrcMatTy->getNumColumns());
-      bool IsRowMajor = CGF.getLangOpts().getDefaultMatrixMemoryLayout() ==
-                        LangOptions::MatrixMemoryLayout::MatrixRowMajor;
+
+      // isMatrixRowMajor needs the full sugared QualType to find matrix layout
+      // attrs. So use Use E->getType() (the source QualType) rather than
+      // SrcMatTy b\c getAs<ConstantMatrixType>() strips the sugar.
```
Icohedron:
```suggestion
      // isMatrixRowMajor needs the full sugared QualType to find matrix layout
      // attrs. So use E->getType() (the source QualType) rather than
      // SrcMatTy b/c getAs<ConstantMatrixType>() strips the sugar.
```

Icohedron (changes requested):

clang/lib/CodeGen/CGExprScalar.cpp
farzonl: It sounds like you are suggesting something like  this:

```hlsl
row_major float3x4 m1;
column_major float3x3 m2 = (float3x3) m1;
```

Which Look to be valid to represent in dxc https://hlsl.godbolt.org/z/sGjoajM44

So Yeah we likely need this.

clang/test/SemaHLSL/matrix_layout_attr.hlsl
```diff
@@ -102,16 +102,13 @@ struct S2 {
 typedef row_major float ScalarRM;
 
 // Invalid: conflicting row_major and column_major on same decl.
-// expected-note@+2 {{conflicting attribute is here}}
```
Icohedron, line 105: I like these notes. It would be nice to keep them rather than remove them.
```
   error: 'column_major' and 'row_major' attributes are not compatible
     row_major column_major float3x3 conflict_mat;
               ^~~~~~~~~~~~
   note: conflicting attribute is here
     row_major column_major float3x3 conflict_mat;
     ^~~~~~~~~
```

clang/include/clang/Basic/Attr.td
```diff
@@ -5296,13 +5296,21 @@ def HLSLVkLocation : HLSLAnnotationAttr {
...
   let LangOpts = [HLSL];
   let Documentation = [HLSLMatrixLayoutDocs];
 }
 
+def HLSLColumnMajor : TypeAttr {
+  let Spellings = [CustomKeyword<"column_major">];
+  let LangOpts = [HLSL];
+  let Documentation = [HLSLMatrixLayoutDocs];
+}
+def : MutualExclusions<[HLSLRowMajor, HLSLColumnMajor]>;
```
Icohedron: `MutualExclusions` is dead code if the inputs are `TypeAttr`s

```suggestion
```

https://github.com/llvm/llvm-project/blob/243cb25fc005f0bee07eddd60f6900d4ac0d1b3b/clang/utils/TableGen/ClangAttrEmitter.cpp#L4553-L4563

clang/test/SemaHLSL/matrix_layout_attr.hlsl
farzonl: Removed because not emitted anymore.

Icohedron: Can you emit them again?

farzonl: Yes. I just have to put this line back `Diag(Existing->getLoc(), diag::note_conflicting_attribute);` I just didn't see much value  with the note. I still don't but don't feel strongly about it.

Icohedron (approved):

clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -2704,60 +2704,76 @@ static bool isMatrixOrArrayOfMatrix(const ASTContext &Ctx, QualType QT) {
...
+  ASTContext &Ctx = getASTContext();
+  attr::Kind AttrK = AL.getKind() == ParsedAttr::AT_HLSLRowMajor
+                         ? attr::HLSLRowMajor
+                         : attr::HLSLColumnMajor;
 
-  SemaRef.Diag(Loc, diag::err_hlsl_matrix_layout_non_matrix) << AttrName;
-  return true;
-}
+  if (T.isNull())
+    return nullptr;
```
bogner: Should we do this null check / early exist before bothering to work out AttrK?

clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -2704,60 +2704,76 @@ static bool isMatrixOrArrayOfMatrix(const ASTContext &Ctx, QualType QT) {
...
+    return ::new (Ctx) HLSLRowMajorAttr(Ctx, AL);
+  return ::new (Ctx) HLSLColumnMajorAttr(Ctx, AL);
 }
 
-bool SemaHLSL::diagnoseInstantiatedMatrixLayoutAttr(
-    Decl *D, const HLSLMatrixLayoutAttr *Attr) {
-  return diagnoseMatrixLayoutOnNonMatrix(SemaRef, D, Attr->getLoc(),
-                                         Attr->getAttrName());
+bool SemaHLSL::diagnoseMatrixLayoutInstantiation(attr::Kind K, QualType T,
+                                                 SourceLocation Loc) {
```
bogner, line 2772: Can we get a comment here explaining when this is needed? Between the name and looking at the use in TreeTransform I can see that this is handling template instantiation, but that isn't all that obvious when looking at the function itself.

clang/test/CodeGenHLSL/matrix-layout-attr-overrides-default.hlsl
```diff
@@ -0,0 +1,227 @@
+// RUN: %clang_cc1 -std=hlsl202x -finclude-default-header -x hlsl -triple \
+// RUN:   dxil-pc-shadermodel6.3-library %s -emit-llvm -disable-llvm-passes \
+// RUN:   -fmatrix-memory-layout=column-major -o - | FileCheck %s
+// RUN: %clang_cc1 -std=hlsl202x -finclude-default-header -x hlsl -triple \
+// RUN:   dxil-pc-shadermodel6.3-library %s -emit-llvm -disable-llvm-passes \
+// RUN:   -fmatrix-memory-layout=row-major -o - | FileCheck %s
```
bogner: I realize this matches a lot of existing practice in this directory, but less so across the whole LLVM repository. IMO splitting run lines like this does more harm than good, and just having lines longer than 80 col is fine here (so it's easy to copy-paste for example)

bogner (approved):
The PR title has a typo (s/codgen/codegen) and the PR description has some weird indentation (which shows up as a code block in the github UI) - Please fix these for the commit message.

Otherwise, LGTM with a few nitpicks.

clang/lib/Sema/SemaHLSL.cpp
farzonl: Something like this:
```
//  At parse time, row_major/column_major are validated to apply to a matrix.
//  But in a template the attribute is attached to a dependent type. This helper
//  catches the matrix layout at instantiation time.
```

farzonl: actually just fed in my comment to claude and it thinks this is better

```
// Re-validates an HLSL `row_major` / `column_major` attribute after template
// substitution. The parse-time check in `buildMatrixLayoutTypeAttr` is skipped
// for dependent types; `TransformAttributedType` calls this once the type is
// concrete. Returns `true` (and emits a diagnostic) if the substituted type is
// not a matrix or array of matrices, signaling the caller to abort the
// transform.
```

farzonl:
test failures are unrelated:
```
Failed Tests (2):
  Clang :: Driver/hip-toolchain-no-rdc.hip
  Clang :: Driver/hipspv-toolchain.hip
```

PR #198889, opened 2026-05-20
[SSAF][WPA] Bounds propagation graph is a supergraph of the pointer-flow graph

ziqingluo-90:
Background: The whole-program UnsafeBufferReachableAnalysis propagates bounds between pointers. It uses the pointer-flow graph extracted and linked from translation units.

This commit patches the gap between the semantics of bounds propagation and pointer-flow: the bounds propagation graph is a super graph of the pointer-flow graph in that a pointer-flow graph edge `(src, i) -> (dst, j)` is the projection of a finite set of bounds propagation graph edges `{(src, i+d) -> (dst, j+d) | 0 <= d < UB}` for a small constant upper bound UB.  See the following example for the idea:

```
void f(int ***p, int **q) {
   *p = q;
   (**p)[5] = 0;
}
```

There is one edge for the static pointer assignment: '(p, 2) -> (q, 1)'.  In terms of bounds propagation, this assignment implies that if 'p' at pointer level 2 requires bounds, 'q' at pointer level 1 must also have them. Furthermore, this relationship propagates to deeper indirection levels: if 'p' at level 3 requires bounds (due to '(**p)[5]'), so does 'q' at level 2.

### DIFF
--- a/clang/include/clang/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlow.h
+++ b/clang/include/clang/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlow.h
@@ -6,9 +6,9 @@
 //
 //===----------------------------------------------------------------------===//
 //
-//  This file defines an analysis that builds directed graphs where nodes
-//  are pointers and edges are assignment operations, each of which bridges two
-//  nodes.
+//  This file defines the per-translation-unit summary type for the
+//  pointer-flow analysis, which records static pointer assignment sites
+//  as directed graph edges (EdgeSet).
 //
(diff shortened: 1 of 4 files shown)


ziqingluo-90:
cc @dtarditi

jkorous-apple:
LGTM!
Basically, the pointer flow facts in the code imply constraints for all levels of the pointers involved.
We might consider making the abstractions easier to understand later but for now this is a great improvement.
FWIW I hit this difference between upstream and the old prototype over the weekend.

jkorous-apple (approved):

ziqingluo-90:
Thanks for the review!

PR #198894, opened 2026-05-20
[OFFLOAD][L0] Remove support for non-immediate command lists

adurang:
The code was in a dead path as immediate command lists are always used.


hansangbae (approved):

sarnex (approved):
not an expert in this code but no flags from me, nice!

adurang:
fyi, @pbalcer @lplewa

PR #195775, opened 2026-05-05
[lldb] Push ExpressionEvaluation policy and remove identity check fallbacks

medismailben:
Push `PolicyStack::Get().PushPublicStateRunningExpression()` at all three expression evaluation entry points (`LLVMUserExpression::DoExecute`, `FunctionCaller::ExecuteFunction`, `IRInterpreter`). This policy sets `can_run_breakpoint_actions=false`, preventing recursive breakpoint callback execution during expression eval.

Push `PolicyStack::Get().PushPrivateState()` unconditionally for all PSTs in `RunPrivateStateThread` (not just overrides), giving every PST the private view while keeping frame providers and recognizers enabled for normal stop processing. Override PSTs use `PushPrivateStateRunningExpression()` which additionally disables providers and recognizers.

With all PSTs and expression eval sites now covered by the policy, remove all host thread identity check fallbacks:
  - `CurrentThreadPosesAsPrivateStateThread()` in `Process::GetState()`
  - `CurrentThreadIsPrivateStateThread()` in `Target::GetAPIMutex()`
  - `IsOnThread()` in `PrivateStateThread::GetRunLock()`
  - `CurrentThreadPosesAsPrivateStateThread()` in `SelectMostRelevantFrame()`
  - `IsRunningExpression()` in `StopInfoBreakpoint::PerformAction()`

`SelectMostRelevantFrame` now checks `!can_run_frame_recognizers` instead of `View::Private`, so recognizers run during normal PST stop processing but are skipped during expression evaluation.

----

The following PRs are related to the Policy feature:
- #195762
- #195771
- #198897
- #195774
- #195775

rdar://176223894

PR #195774, opened 2026-05-05
[lldb] Harden PolicyStack against cross-thread Guard misuse

medismailben:
Two related correctness/clarity improvements to the `Policy` infrastructure introduced by 504a112:

`PolicyStack::Guard` now stores the std::thread::id of the thread that created it. Destruction and move operations call `llvm::report_fatal_error` when they happen on a different thread, since the `PolicyStack` is `thread_local`: popping from the wrong thread would silently corrupt that thread's stack.

`Push`/`Pop` on `PolicyStack` are now private. Callers go through named factories (`PushPrivateState`, `PushPublicStateRunningExpression`) that return RAII Guards. The transition factories on Policy (`CreatePrivateState`, `CreatePublicStateRunningExpression`) inherit from `PolicyStack::Get().Current()` and apply their named change on top, so pushed policies preserve existing stack state rather than resetting unrelated fields. `CreatePublicState` remains the baseline reference value (returns a default `Policy{}`); the stack returns to public state by popping the private guards, not by pushing a "public" policy on top.

----

The following PRs are related to the Policy feature:
- #195762
- #195771
- #198897
- #195774
- #195775

rdar://176223894

PR #195771, opened 2026-05-05
[lldb] Adopt Policy for private/public view and capability decisions

medismailben:
Push Policy::PrivateState() at all sites where the private state thread needs to see the private reality instead of the public illusion:
- StopInfoBreakpoint::ShouldStopSynchronous (sync callbacks, WasHit)
- StopInfoWatchpoint::ShouldStopSynchronous
- RunThreadPlan (original PST during expression evaluation)
- RunPrivateStateThread (override PSTs)

Consult the policy at all view/capability decision points:
- Process::GetState() -- private vs public state
- Target::GetAPIMutex() -- private vs public mutex
- PrivateStateThread::GetRunLock() -- private vs public run lock
- Thread::GetStackFrameList() -- parent vs provider-augmented frames
- StackFrameList::SelectMostRelevantFrame() -- skip frame recognizers
- StopInfoBreakpoint::PerformAction() -- skip breakpoint actions
- StopInfoWatchpoint::PerformAction() -- skip watchpoint actions

The existing host thread identity checks remain as fallbacks until Policy is pushed for all PSTs unconditionally.

This commit also add Stream::operator<<(bool) to allow streaming boolean values as "true"/"false".

----

The following PRs are related to the Policy feature:
- #195762
- #195771
- #195774
- #195775

rdar://176223894

PR #195762, opened 2026-05-05
[lldb] Add Policy infrastructure

medismailben:
Add a generic thread-local policy stack and a Policy struct that describes what view of the process a thread should see (private reality vs public illusion) and what operations it is allowed to perform.

This is the infrastructure for replacing ad-hoc host thread identity checks (CurrentThreadIsPrivateStateThread, IsOnThread, etc.) with a unified, composable mechanism. No behavioral changes yet -- adoption will follow in subsequent patches.

----

The following PRs are related to the Policy feature:
- #195762
- #195771
- #195774
- #195775

rdar://176223894

PR #198897, opened 2026-05-20
[lldb] Move Policy from Target to Utility (NFC)

medismailben:
Move the Policy struct and PolicyStack class from lldb/Target to lldb/Utility. This is a pure relocation -- no API or behavior changes.

This is needed so that lldbHost (which contains ProcessRunLock) can depend on Policy without introducing a layering violation, since lldbHost cannot depend on lldbTarget.

----

The following PRs are related to the Policy feature:
- #195762
- #195771
- #198897
- #195774
- #195775

rdar://176223894


bulbazord (approved):
layering is good

PR #198898, opened 2026-05-20
[RISCV][P-ext] Split v4i16/v8i8 vselect on RV32.

topperc:
We don't have a 64-bit merge instruction on RV32.

Assisted-by: Claude Sonnet 4.5


lenary:
Much better :) Didn't fully understand the shift changes, so no +1

topperc:
> Much better :) Didn't fully understand the shift changes, so no +1

SSHLSAT was being scalarized because vselect was set to Expand. https://github.com/llvm/llvm-project/blob/1df5015582835a38f512100aa984ec55350a61b7/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp#L12021

Now we expand using vector operations.

https://github.com/llvm/llvm-project/pull/198937 will replace it with a single instruction.

llvm/test/CodeGen/RISCV/rvp-simd-64.ll
```diff
@@ -1930,36 +1877,49 @@ define <4 x i16> @test_pssla_h(<4 x i16> %a, <4 x i16> %b) {
...
+; RV32-NEXT:    pmseq.dh a2, a0, t1
+; RV32-NEXT:    pmsltz.dh a0, a0
+; RV32-NEXT:    addi a7, a7, -1
+; RV32-NEXT:    pmv.hs a7, a7
+; RV32-NEXT:    merge a1, a7, t0
+; RV32-NEXT:    pack a5, a6, a5
+; RV32-NEXT:    merge a3, a1, a4
+; RV32-NEXT:    merge a0, a7, t0
+; RV32-NEXT:    merge a2, a0, a5
+; RV32-NEXT:    padd.dw a0, a2, zero
```
sihuan, line 1922: Heads-up: codegen here regresses from 33 to 46 ops (we lose the `ssha` match). Left the detailed context in the top-level comment.

sihuan (approved):
LGTM modulo two minor notes:

- Title typo: `v8i8`, not `v8i16`.
- `test_pssla_h` (vector-vector `sshl.sat.v4i16`) regresses 33 → 46 ops on RV32. Before this PR, VSELECT-Expand unrolled the whole `sshl.sat` chain into per-lane scalarization, and `slli + ssha + srli` matched per lane. With VSELECT now Custom, only SHL/SRA gets unrolled while SETCC + VSELECT stays vector-form. The new codegen ends up packing the per-lane scalar results back into v4i16 and computing saturation explicitly (`pmseq.dh` + `pmsltz.dh` + 4× `merge` + materializing INT16_MAX/MIN splats). What `ssha` did in one op now takes several.
  Since P-ext lacks a vector-vector saturating shift, it looks like recovering `ssha` here would require unrolling the entire chain at the SSHLSAT lowering stage. This feels like its own discussion, so I think we should leave it out of this PR.

PR #198900, opened 2026-05-20
[libc] Enable baremetal float printf using modular format

mysterymath:
The most best configuration for llvm-libc is for it to be used with a recent clang. Accordingly, this patch enables floating point type support in printf for baremetal without inflating code size by also enabling modular printf support. This is safe for older clang and other compilers, but it would increase code size for the default baremetal configuration on such targets.

### DIFF
--- a/libc/config/baremetal/config.json
+++ b/libc/config/baremetal/config.json
@@ -13,9 +13,6 @@
     "LIBC_CONF_PRINTF_DISABLE_FIXED_POINT": {
       "value": true
     },
-    "LIBC_CONF_PRINTF_DISABLE_FLOAT": {
-      "value": true
-    },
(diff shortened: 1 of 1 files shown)


libc/config/baremetal/config.json
```diff
@@ -33,6 +30,9 @@
     },
     "LIBC_COPT_PRINTF_DISABLE_BITINT": {
       "value": true
+    },
+    "LIBC_COPT_PRINTF_MODULAR": {
```
michaelrj-google: This is a cmake config option instead of the similarly named copt.
```suggestion
    "LIBC_CONF_PRINTF_MODULAR": {
```

michaelrj-google (commented):
LGTM from the libc side, after fixing the typo

petrhosek (approved):

PR #196818, opened 2026-05-10
[AMDGPU] Fine tune partial threshold for runtime unrolling.

carlobertolli:
Respond to reported issue:
https://github.com/llvm/llvm-project/issues/196372

PR #194924, opened 2026-04-29
[AMDGPU] Enable runtime loop unrolling

adelejjeh:
Enable auto runtime unrolling for AMDGPU by setting `UP.Runtime = true` in `getUnrollingPreferences`, with `PartialThreshold = Threshold / 4` to limit code-size growth.

Benchmarked on **MI350X (gfx950)** and **MI300X (gfx942)** using Composable Kernel, xpu-perf, and llama.cpp. Results showed some some improvements and no real regressions.

AI Disclaimer: Cursor was used to evaluate the change and run benchmarking experiments.

PR #198901, opened 2026-05-20
[AMDGPU] Remove explicit PartialThreshold setting in loop unrolling

### LINKED ISSUE #196372: [AMDGPU] Loop's auto runtime unrolling change partial threshold
Downstream saw some perf regressions after https://github.com/llvm/llvm-project/pull/194924

Investigation shows that the part:
```
PartialThreshold = Threshold / 4;
```
is changing global unrolling behaviors by gating compile-time partial unrolling, suppressing the partial-unroll-by-2 of small constant-trip-count reduction loops, causing unrolling regressions/behaviour changes.

We had to temporarily revert that in the pipeline to recover the lost: https://github.com/iree-org/iree/pull/24379/changes#diff-2d6e1348b856a2542889d67f2a91f0d42a9e7faca831e891449bded8870b056eR662-R664

adelejjeh:
Remove UP.PartialThreshold = UP.Threshold / 4 from AMDGPU TTI, restoring the default PartialThreshold of 150.

This was introduced in #194924 to limit code-size growth from runtime unrolling, but PartialThreshold also gates compile-time partial unrolling of constant-trip-count loops. This change will make the PartialThreshold back to the default value for both compile-time partial unrolling and runtime partial unrolling.

Benchmarked across CK, llama.cpp, and xpu-perf — no performance impact from restoring the default.

Fixes #196372, replaces #196818.

Assisted-by: Claude Code


shiltian:
Not ready for review yet?

adelejjeh:
@shiltian ready now!

lialan (approved):
+1

shiltian (approved):

ravil-mobile:
@adelejjeh

The change resulted in an aggressive transformation (i.e.,  loop unrolling in a software pipelined loop) that led to increased register pressure beyond the physical register file capacity.

The change also broke the numerical correctness of a Triton-GEMM kernel by allowing aggressive loop unrolling that:
1. Destroyed the MFMA accumulator chain
2. Used wrong base accumulators for iteration 1
3. Failed to properly chain spill/reload operations

shiltian:
The correctness issue can be fixed by https://github.com/llvm/llvm-project/pull/204958. Note that this change doesn't directly cause the issue. Rather, it exposes existing issues in other parts of the code. That said, the fix is probably not going to be trivial. I still need to figure out the best way to address it.

adelejjeh:
@ravil-mobile a temporary solution can be using `#pragma nounroll`.

ravil-mobile:
> @ravil-mobile a temporary solution can be using `#pragma nounroll`.

We are working with a Triton compiler; thus, we don't have pragma constructs. There is a way to add something similar to `nonunroll` but it is not desirable. The main issue that I personally concern is the regression in numerics.

shiltian:
> The change resulted in an aggressive transformation (i.e., loop unrolling in a software pipelined loop) that led to increased register pressure beyond the physical register file capacity.

That doesn't sound like an issue with this change. It seems more like a case of suboptimal instruction scheduling, or perhaps an interaction between instruction scheduling and register allocation. For example, the RA may not be able to stay within the budget assumed by the scheduler.

ravil-mobile:
> > The change resulted in an aggressive transformation (i.e., loop unrolling in a software pipelined loop) that led to increased register pressure beyond the physical register file capacity.
> 
> That doesn't sound like an issue with this change. It seems more like a case of suboptimal instruction scheduling, or perhaps an interaction between instruction scheduling and register allocation. For example, the RA may not be able to stay within the budget assumed by the scheduler.

Agree. I guess this PR just unlocked a latent bug in the backend

ravil-mobile:
@shiltian @adelejjeh I did a bisect of LLVM backend passes. The issue points to `machine-scheduler` pass. I created an internal JIRA ticket.

shiltian:
> @shiltian @adelejjeh I did a bisect of LLVM backend passes. The issue points to `machine-scheduler` pass. I created an internal JIRA ticket.

Can you share the JIRA ticket number? It seems like the issue exposed disappeared.

PR #198902, opened 2026-05-20
[Hexagon] Add XQFloat code generation and post-RA QFP handling

fhossein-quic:
Introduce two new passes for the Hexagon HVX floating-point pipeline, targeting v79+ where QFloat (qf16/qf32) is the native HVX FP format.

HexagonXQFloatGenerator lowers IEEE-754 HVX floating-point sequences (sf/hf) to native QFloat (qf16/qf32) operations. QFloat instructions are faster and more power-efficient than their IEEE counterparts, with optional accuracy trade-offs. The pass exposes four modes:

  * Strict IEEE-754 compliant
  * IEEE-754 compliant (extended dynamic range and subnormal precision, no IEEE-754 overflow handling)
  * Lossy subnormals
  * Legacy

HexagonPostRAHandleQFP runs after register allocation and corrects the spill/refill paths. QFloat operands carry four extra precision bits that are silently dropped if the value passes through a spill slot or a non-HVX instruction. The pass uses the Register DataFlow Graph (RDF) to walk use-def chains in non-SSA form, inserts qf->IEEE conversions before spills and non-HVX uses, and rewrites saturating instruction opcodes that can absorb IEEE operands directly.

Co-authored-by: Sumanth Gundapaneni <sgundapa@qti.qualcomm.com>
Co-authored-by: Santanu Das <santdas@qti.qualcomm.com>


clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -2235,6 +2236,115 @@ void Clang::AddX86TargetArgs(const ArgList &Args,
...
+                                 options::OPT_mhexagon_hvx_ieee_fp)) {
+      if (HvxVerNum >= 79) {
+        if (A->getOption().matches(options::OPT_mhexagon_hvx_qfloat_EQ)) {
+          const char *Mode =
+              llvm::StringSwitch<const char *>(StringRef(A->getValue()).lower())
+                  .Case("strict-ieee", "-hexagon-qfloat-mode=strict-ieee")
+                  .Case("ieee", "-hexagon-qfloat-mode=ieee")
+                  .Case("lossy", "-hexagon-qfloat-mode=lossy")
+                  .Case("legacy", "-hexagon-qfloat-mode=legacy")
+                  .Default("-hexagon-qfloat-mode=lossy");
```
androm3da: Does this `Default()` mean that if I did `-mhvx-qfloat=iee` I would get `lossy`?  If so I think that might be a mistake -- seems like it should be a fatal error instead.

llvm/include/llvm/CodeGen/RDFGraph.h
```diff
@@ -358,6 +358,8 @@ template <typename T> struct NodeAddr {
     return !operator==(NA);
   }
 
+  bool operator<(const NodeAddr<T> &NA) const { return Id < NA.Id; }
```
androm3da: Should this architecture independent change land in its own commit?

llvm/test/CodeGen/Hexagon/autohvx/xqf-warnings.ll
```diff
@@ -0,0 +1,143 @@
...
+!llvm.module.flags = !{!2, !3, !4, !5, !6}
+!llvm.ident = !{!7}
+
+!0 = distinct !DICompileUnit(language: DW_LANG_C11, file: !1, producer: "QuIC LLVM Hexagon Clang version 8.8-alpha2 Engineering Release: hexagon-clang-88-5172", isOptimized: true, runtimeVersion: 0, emissionKind: FullDebug, splitDebugInlining: false, debugInfoForProfiling: true, nameTableKind: None)
+!1 = !DIFile(filename: "test.c", directory: "/local/mnt/workspace/santdas/src/8_8/build")
+!2 = !{i32 7, !"Dwarf Version", i32 4}
+!3 = !{i32 2, !"Debug Info Version", i32 3}
+!4 = !{i32 1, !"wchar_size", i32 4}
+!5 = !{i32 7, !"frame-pointer", i32 2}
+!6 = !{i32 7, !"debug-info-assignment-tracking", i1 true}
```
androm3da: Is the debug info required for this test case?  If not, let's purge it.

androm3da (commented):
This very large patch would be easier to read if it were decomposed into multiple commits, each with their own motivation.  And once that decomposition is there it might make sense to land some of those commits as independent PRs.

llvm/lib/Target/Hexagon/HexagonXQFloatGenerator.cpp
```diff
@@ -0,0 +1,2177 @@
...
+
+// This vector contains the opcodes which generate qf32 from multiplication
+SmallVector<unsigned short, 5> XQFPMult32 = {
+    Hexagon::V6_vmpy_qf32, Hexagon::V6_vmpy_qf32_qf16, Hexagon::V6_vmpy_qf32_hf,
+    Hexagon::V6_vmpy_qf32_sf, Hexagon::V6_vmpy_qf32_mix_hf};
+// This vector contains the opcodes which generate qf16 from multiplication
+SmallVector<unsigned short, 3> XQFPMult16 = {Hexagon::V6_vmpy_qf16,
+                                             Hexagon::V6_vmpy_qf16_hf,
+                                             Hexagon::V6_vmpy_qf16_mix_hf};
+
```
androm3da, line 184: These `XQFP*` globals should each be `constexpr` ideally.

llvm/lib/Target/Hexagon/HexagonXQFloatGenerator.cpp
```diff
@@ -0,0 +1,2177 @@
...
+// qf and IEEE-754 value can be unnormal.
+// Additionally for v81, we have two new vsub instructions which are handled.
+
+#define HEXAGON_XQFLOAT_GENERATOR "XQFloat Generator pass"
+
+#include "Hexagon.h"
+#include "HexagonInstrInfo.h"
+#include "HexagonSubtarget.h"
+#include "HexagonTargetMachine.h"
+#include "vector"
```
androm3da: This refers to the C++ library's `std::vector`?  If so this should be `#include <vector>`

clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -2235,6 +2236,115 @@ void Clang::AddX86TargetArgs(const ArgList &Args,
...
+}
+
+static void addQFloatBackendArg(const Driver &D, const ArgList &Args,
+                                ArgStringList &CmdArgs) {
+  auto HvxVerOpt = toolchains::HexagonToolChain::GetHVXVersion(Args);
+  bool HasHVX = HvxVerOpt.has_value();
+  std::string HvxVer = HasHVX ? *HvxVerOpt : std::string();
+  if (Args.hasArg(options::OPT_mhexagon_hvx, options::OPT_mhexagon_hvx_EQ,
+                  options::OPT_mhexagon_hvx_ieee_fp) &&
+      HasHVX) {
```
androm3da: Instead of having this outermost `if` - do an early exit. https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -2235,6 +2236,115 @@ void Clang::AddX86TargetArgs(const ArgList &Args,
...
+        if (Arg *QFloatArg =
+                Args.getLastArg(options::OPT_mhexagon_hvx_qfloat,
+                                options::OPT_mhexagon_hvx_qfloat_EQ,
+                                options::OPT_mno_hexagon_hvx_qfloat);
+            QFloatArg && QFloatArg->getOption().matches(
+                             options::OPT_mhexagon_hvx_qfloat_EQ)) {
+          D.Diag(diag::warn_drv_unsupported_option_part_for_target)
+              << QFloatArg->getValue() << QFloatArg->getAsString(Args)
+              << (std::string("HVX ") + HvxVer +
+                  "; falling back to legacy qfloat mode");
```
androm3da: This fallback behavior (assuming it's desired) would probably be good to describe in the clang user manual `clang/docs/UsersManual.rst`.

llvm/lib/Target/Hexagon/HexagonXQFloatGenerator.cpp
```diff
@@ -0,0 +1,2177 @@
...
+      return checkIfInputFromAdder32(SrcReg);
+    return false;
+  } else if (Def->getOpcode() == TargetOpcode::REG_SEQUENCE) {
+    Register SrcReg1 = Def->getOperand(1).getReg();
+    Register SrcReg2 = Def->getOperand(2).getReg();
+    bool isTrue = false;
+    if (SrcReg1.isValid())
+      isTrue = checkIfInputFromAdder32(SrcReg1);
+    if (SrcReg2.isValid())
+      isTrue = checkIfInputFromAdder32(SrcReg2);
```
androm3da: If both `SrcReg1.isValid()` and `SrcReg2.isValid()` should `SrcReg2`'s `InputFromAdder32` take precedence for `isTrue`?  It can overwrite `isTrue`, is that by design?

clang/test/Driver/hexagon-hvx-ieee-qfloat.c
```diff
@@ -0,0 +1,25 @@
+// ---------------------------------------------------------------------------
```
androm3da, line 1: Can you also add test(s) to cover feature propagation for `-fno-integrated-as`?

```c
// RUN: %clang -### -fno-integrated-as -target hexagon-unknown-elf -mv79 \
// RUN:   -mhvx -mhvx-ieee-fp %s 2>&1 | FileCheck -check-prefix=AS-QFLOAT %s
// AS-QFLOAT: "-mattr=+hvx-qfloat"
```

llvm/lib/Target/Hexagon/HexagonTargetMachine.h
```diff
@@ -33,6 +35,7 @@ class HexagonTargetMachine : public CodeGenTargetMachineImpl {
                        std::optional<CodeModel::Model> CM, CodeGenOptLevel OL,
                        bool JIT);
   ~HexagonTargetMachine() override;
+  const HexagonSubtarget *getSubtargetImpl() const { return &Subtarget; }
```
androm3da: Several other targets deliberately omit this no-argument version of `getSubtargetImpl`.  This one would shadow the `getSubtargetImpl(Function)`?

llvm/lib/Target/Hexagon/HexagonXQFloatGenerator.cpp
```diff
@@ -0,0 +1,2177 @@
...
+      case Hexagon::V6_vmpy_qf32:
+        if (normalizeMultiplicationInputF32(MI, Reg1, Reg2, Dest, R_mpy,
+                                            PrologCreated))
+          OriginalMI.push_back(&MI);
+        Changed |= convertIfInputToNonHVX(MI, true);
+        break;
+
+      // qf16 = vmpy(hf, rt)
+      // Splat Rt to vector and then widening multiply
+      case Hexagon::V6_vmpy_rt_hf:
```
quic-santdas, line 1328: These opcodes (for different modes) were done for v79. Later architectures have probably introduced more qf types which need to be handled.

quic-santdas (commented):
For each of the modes in xqfgenerator pass (lossy, ieee,etc.) opcodes of the latest ISA need to be added too. This pass was targetting v79 after which a few new qf opcodes were added.

Some classes like the one triggered by EnablePostRAXqfCompliance in postRA pass (a verifier if there are some incorrect/missing transformations. It is not necessary for the functionality, but acts more like a debugger) and EnableConversionsRemoval in XQFgenerator (removal of extra conversions) can probably be removed from the scope of the patch. The latter is also not turned on by default.

llvm/lib/Target/Hexagon/HexagonXQFloatGenerator.cpp
quic-santdas: Ideally, these should be moved to HexagonInstrInfo and queries via instance of that class.

clang/lib/Driver/ToolChains/Clang.cpp
fhossein-quic: Done. We are erroring out now.

llvm/test/CodeGen/Hexagon/autohvx/xqf-warnings.ll
fhossein-quic: Done

llvm/lib/Target/Hexagon/HexagonXQFloatGenerator.cpp
fhossein-quic: I used  `constexpr` for now. The ideal refactoring could be applied in a separate PR to reduce CUI confusion.

llvm/lib/Target/Hexagon/HexagonXQFloatGenerator.cpp
fhossein-quic: Thanks for catching that. Fixed now by using `=|`

clang/test/Driver/hexagon-hvx-ieee-qfloat.c
fhossein-quic: Done.

llvm/lib/Target/Hexagon/HexagonXQFloatGenerator.cpp
fhossein-quic: Yes, after this patch is merged, we can update it with latest opcodes.

clang/lib/Driver/ToolChains/Clang.cpp
fhossein-quic: Done

llvm/lib/Target/Hexagon/HexagonXQFloatGenerator.cpp
fhossein-quic: Done

llvm/include/llvm/CodeGen/RDFGraph.h
fhossein-quic: I figured this change is not really needed. I relied on using NodeId instead, matching the pattern used by RDFLiveness.cpp.

clang/lib/Driver/ToolChains/Clang.cpp
fhossein-quic: Is it ok with you to add the documentation in a separate PR?

androm3da: sure

llvm/lib/Target/Hexagon/HexagonTargetMachine.h
fhossein-quic: Renamed to getHexagonSubtarget() to avoid the shadowing

fhossein-quic:
> For each of the modes in xqfgenerator pass (lossy, ieee,etc.) opcodes of the latest ISA need to be added too. This pass was targetting v79 after which a few new qf opcodes were added.
> 
> Some classes like the one triggered by EnablePostRAXqfCompliance in postRA pass (a verifier if there are some incorrect/missing transformations. It is not necessary for the functionality, but acts more like a debugger) and EnableConversionsRemoval in XQFgenerator (removal of extra conversions) can probably be removed from the scope of the patch. The latter is also not turned on by default.

After this patch is merged, we can update it with more opcodes needed to latest ISAs.▎

I removed VectorConvertRemove (EnableConversionsRemoval) and XqfPostRADiagnosis (EnablePostRAXqfCompliance) and I'll upstream them as follow up patches. This patch now contains only the core XQFloat generator and post-RA QFP handling.

clang/test/Driver/hexagon-hvx-qfloat-backend.c
```diff
@@ -0,0 +1,51 @@
...
+// RUN: %clang -c %s -### -target hexagon-unknown-elf -mv75 -mhvx \
+// RUN:   -mhvx-qfloat=ieee 2>&1 | FileCheck -check-prefix=CHECK-MODE-WARN %s
+// Test that an invalid qfloat mode is an error.
+// CHECK-INVALID: error: invalid value 'iee' in '-mhvx-qfloat=iee'
+// RUN: %clang -c %s -### -target hexagon-unknown-elf -mv79 -mhvx \
+// RUN:   -mhvx-qfloat=iee 2>&1 | FileCheck -check-prefix=CHECK-INVALID %s
+// Test that -fno-integrated-as suppresses the qfloat backend flag.
+// CHECK-NO-INTEG-AS-NOT: "-hexagon-qfloat-mode=
+// RUN: %clang -c %s -### -target hexagon-unknown-elf -mv79 -mhvx \
+// RUN:   -mhvx-qfloat=ieee -fno-integrated-as 2>&1 | FileCheck -check-prefix=CHECK-NO-INTEG-AS %s
```
androm3da: Hmm - sorry I think I was unclear about what I was asking for.

Correct me if I am wrong: `-mhvx-qfloat` has multiple features - (1) it permits the backend to emit opcodes for HVX QFloat and (2) based on the mode it impacts codegen.

I'm concerned primarily about (1) - we should propagate corresponding flags to the assembler when invoking the assembler in `-fno-integrated-as`.  So I wouldn't expect that we "suppress" flags in this case.  If `-mhvx-qfloat` doesn't do (1) then it's probably moot.

llvm/test/CodeGen/Hexagon/autohvx/xqf-assertion1.ll
```diff
@@ -0,0 +1,84 @@
+; On v79 and above, checks for Assertion `isImm() && "Wrong MachineOperand accessor"' failed
```
androm3da, line 1: This likely needs a `REQUIRES: asserts` to function as intended

llvm/test/CodeGen/Hexagon/autohvx/xqf-assertion1.ll
```diff
@@ -0,0 +1,84 @@
...
+!llvm.module.flags = !{!0, !1, !2}
+!llvm.ident = !{!6}
+
+!0 = !{i32 1, !"wchar_size", i32 4}
+!1 = !{i32 7, !"frame-pointer", i32 2}
+!2 = !{i32 5, !"CG MDInfo", !3}
+!3 = !{!4, !5}
+!4 = !{!"F", !"no_filename_available", !"", !"", i1 false, !""}
+!5 = !{!"C", !"set_double_vector_mode", !"(void)", !"(...)", i1 true, !""}
+!6 = !{!"QuIC LLVM Hexagon Clang version 8.8-alpha3 Engineering Release: hexagon-clang-88"}
```
androm3da: If we can strip some of these from the test case, let's please do so

fhossein-quic:
> ectorConvertRemove (EnableConversionsRemoval) and XqfPostRADiagnosis (EnablePostRAXqfCompliance)

You're absolutely right, and I appreciate the feedback. These two passes evolved together and are pretty tightly coupled. These two passes evolved together over many iterations and are pretty tightly coupled. Each pass itself also went through several rounds of fixes and corrections internally, but I realized those intermediate commits can't be landed individually since each one could produce incorrect codegen on its own.  I did pull out what I could and created multiple separate PRs for the part that could have taken out of this PR. I also pulled out VectorConvertRemove and XqfPostRADiagnosis and will upstream them as separate follow-up PRs per Santanu's suggestion. I'm open to any feedback that helps me in splitting this furthur.

llvm/include/llvm/CodeGen/RDFGraph.h
```diff
@@ -358,6 +358,7 @@ template <typename T> struct NodeAddr {
     return !operator==(NA);
   }
 
+
```
androm3da: stray newline

fhossein-quic: removed

llvm/test/CodeGen/Hexagon/autohvx/xqf-assertion1.ll
fhossein-quic: Done

llvm/test/CodeGen/Hexagon/autohvx/xqf-assertion1.ll
fhossein-quic: Done

clang/test/Driver/hexagon-hvx-qfloat-backend.c
fhossein-quic: My understanding is that this flag only affects (2). The guard I originally added was gating (2), which was a mistake, so I removed it. This patch doesn't add anything related to (1).

I also ran some tests with and without the integrated assembler. The non-integrated path does fail, but it appears to be a pre-existing issue unrelated to this patch. It seems that we don't pass -mhvx to the external assembler upstream, so all HVX instructions fail to assemble. That seems like a separate issue that should be fixed independently.

androm3da: Agreed, but please open an issue to track it.

androm3da (commented):
LGTM but someone on compiler team should approve

sgundapa (approved):
LGTM. Thanks for upstreaming this work

llvm/lib/Target/Hexagon/HexagonXQFloatGenerator.cpp
```diff
@@ -0,0 +1,1683 @@
...
+    // Normalize only second operand
+  } else if (secondconvert) {
+    input_mpy2 = MRI->createVirtualRegister(&Hexagon::HvxVRRegClass);
+
+    BuildMI(MBB, MI, DL, HII->get(Hexagon::V6_vadd_qf32_mix), input_mpy2)
+        .addReg(R_mpy)
+        .addReg(VR2);
+    BuildMI(MBB, MI, DL, HII->get(Hexagon::V6_vmpy_qf32), Dest)
+        .addReg(input_mpy2)
+        .addReg(Reg2);
```
nikic, line 927: Was this supposed to be Reg1?

fhossein-quic: Yes, thanks for catching that. Please check out the fix https://github.com/llvm/llvm-project/pull/208489

PR #198906, opened 2026-05-20
[lldb][windows] enable CI tests

### LINKED ISSUE #132800: [CI] Enable lldb on windows premerge

charles-zablit:
This patch enables CI testing for lldb on Windows. It skips dependent PR testing for now until we enable daemonized testing.

This patch also skips `TestDAP_restart_console` on Windows because it's failing on the bots and is unresolved in the docker container. See https://github.com/llvm/llvm-project/issues/200840 fore more details.

fix https://github.com/llvm/llvm-project/issues/132800.

### DIFF
--- a/.ci/compute_projects.py
+++ b/.ci/compute_projects.py
@@ -201,12 +201,7 @@ SKIP_PROJECTS = ["docs", "gn"]
 # where a project can be built but its tests are not yet stable on that
 # platform, so we still want a compile-time signal.
 PROJECT_CHECK_TARGETS_OVERRIDE = {
-    "Windows": {
-        # TODO(issues/132800): LLDB tests need environment setup on Windows.
-        # In the meantime, at least compile lldb and lldb-dap to catch
-        # breakage.
-        "lldb": "lldb lldb-dap",
-    },
+    "Windows": {},
 }
 
 
--- a/.ci/compute_projects.py
+++ b/.ci/compute_projects.py
@@ -123,6 +123,8 @@ EXCLUDE_WINDOWS = {
 # enabled on changes to dependencies.
 EXCLUDE_DEPENDENTS_WINDOWS = {
     "flang",
+    # TODO: Re-enable once Windows CI timings allow it (daemonized testing).
+    "lldb",
 }
 
 EXCLUDE_MAC = {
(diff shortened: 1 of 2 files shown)


.ci/compute_projects.py
```diff
@@ -201,12 +201,7 @@
 # where a project can be built but its tests are not yet stable on that
 # platform, so we still want a compile-time signal.
 PROJECT_CHECK_TARGETS_OVERRIDE = {
```
boomanaiden154: We can probably drop this completely now.

boomanaiden154 (commented):
Can we add a test case in `compute_projects_test.py`?

I'm also not sure that we want to run LLDB by default on clang/LLVM changes like we do on Linux given we're a bit more constrained timing wise on Windows, although that might change soon with daemonized testing.

charles-zablit:
Converting to draft until all the tests pass.

charles-zablit:
ping @boomanaiden154, all the tests are passing now.

lldb/test/API/tools/lldb-dap/restart/TestDAP_restart_console.py
```diff
@@ -12,7 +12,7 @@
 @skipIfBuildType(["debug"])
 class TestDAP_restart_console(lldbdap_testcase.DAPTestCaseBase):
     @skipIfAsan
-    @expectedFailureWindows
+    @skipIfWindows  # https://github.com/llvm/llvm-project/issues/200840
```
boomanaiden154, line 15: Can this be split into a separate PR for LLDB people to look at?

boomanaiden154 (commented):
Thanks, this largely looks good. If we can split off the LLDB test changes I think this should be good to go.

lldb/test/API/tools/lldb-dap/restart/TestDAP_restart_console.py
charles-zablit: Sure, I opened https://github.com/llvm/llvm-project/pull/201587 to skip it. I will rebase once it's merged.

.ci/compute_projects_test.py
```diff
@@ -191,16 +191,6 @@ def test_bolt(self):
         self.assertEqual(env_variables["runtimes_check_targets"], "")
         self.assertEqual(env_variables["runtimes_check_targets_needs_reconfig"], "")
 
-    def test_lldb(self):
```
boomanaiden154, line 194: We should keep this test unless there's a reason I'm missing for why we're dropping it.

charles-zablit: I was going to say that I removed them because `lldb` was added to `EXCLUDE_DEPENDENTS_WINDOWS`, but that impacts `test_lldb_windows`, not this test. I've reverted that change.

boomanaiden154 (approved):
LGTM.

PR #198907, opened 2026-05-20
[lldb-server] Add accelerator plugin infrastructure for debugging hardware accelerators like gpus

satyajanga:
This is the first patch of many related to https://discourse.llvm.org/t/upstreaming-basic-support-for-accelerators/89827/6

### What this patch adds

  - **`LLDBServerAcceleratorPlugin`** base class in
    `source/Plugins/Process/gdb-remote/`  so accelerator can implement the own plugin
  - **`accelerator-plugins+`** feature in `qSupported` response, only
    advertised when at least one plugin is installed
  -  **`jAcceleratorPluginInitialize`** GDB remote packet and its implementation in handlers, request and response.
  - **`AcceleratorActions`** struct so every plugin can return the actions that needs to be on the initilaize. in the future we will extend this install breakpoints etc.
  - **Mock accelerator plugin** for testing, gated by CMake option
    `LLDB_ENABLE_MOCK_ACCELERATOR_PLUGIN` (default OFF)
  - **Tests** that connect to a real lldb-server, verify
    `accelerator-plugins+` in `qSupported`, send
    `jAcceleratorPluginInitialize`, and validate the JSON response

  ### Design decisions
  - CMake option defaults to OFF so normal builds are unaffected
  - Tests skip automatically when the plugin is not compiled in

NOTE: Sorry for the big patch, We tried to make self contained as much as possible.


lldb/test/API/accelerator/mock/lit.local.cfg
```diff
@@ -0,0 +1,3 @@
+# Enable the mock accelerator plugin in lldb-server for all tests in this
+# directory. The plugin activates when this variable is present and non-empty.
+config.environment["LLDB_SERVER_ENABLE_MOCK_ACCELERATOR_PLUGIN"] = "1"
```
clayborg: remove. We should enable tests for the Mock GPU only when the cmake variable that enables it in the build is set.

lldb/include/lldb/Utility/GDBRemoteAcceleratorPackets.h
```diff
@@ -0,0 +1,35 @@
+//===-- GDBRemoteAcceleratorPackets.h --------------------------*- C++ -*-===//
```
JDevlieghere: Since https://discourse.llvm.org/t/is-c-in-header-files-still-relevant/83124
```suggestion
//===----------------------------------------------------------------------===//
```

lldb/source/Plugins/Process/gdb-remote/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ add_lldb_library(lldbPluginProcessGDBRemote PLUGIN
   GDBRemoteCommunicationServerCommon.cpp
   GDBRemoteCommunicationServerLLGS.cpp
   GDBRemoteCommunicationServerPlatform.cpp
+  LLDBServerAcceleratorPlugin.cpp
```
JDevlieghere, line 34: Should this live in a separate plugin? Related plugins *can* depend on each other and this seems pretty self contained.

clayborg: Are you asking if we should put this into a separate directory like `lldb/source/Plugins/Process/gdb-remote/Plugins`?

lldb/source/Utility/GDBRemoteAcceleratorPackets.cpp
```diff
@@ -0,0 +1,27 @@
+//===-- GDBRemoteAcceleratorPackets.cpp ------------------------*- C++ -*-===//
```
clayborg: Maybe rename to "GDBRemoteAccelerator.cpp"? These are just utilities for defining and encode/decode to/from JSON, nothing really packet related.

lldb/test/API/accelerator/mock/TestMockAcceleratorPlugin.py
```diff
@@ -0,0 +1,68 @@
...
+
+        raw_response = context.get("accel_init_response")
+        self.assertIsNotNone(raw_response)
+
+        decoded = self.decode_gdbremote_binary(raw_response)
+        actions = json.loads(decoded)
+        self.assertIsInstance(actions, list)
+        self.assertGreater(len(actions), 0)
+
+        mock_action = actions[0]
```
clayborg: Maybe we want a helper function that can extract a AccleratorAction dict for a specified plug-in name? Something like:
```
def get_acclerator_action(actions, plugin_name):
    for action in actions:
        if action["plugin_name"] == plugin_name:
            return action
    return None
```
Then this code becomes:
```
mock_action = get_acclerator_action(actions, "mock")
```
This will allow us to support multiple accelerator plug-ins in the future and easily get the actions for the right plug-in

lldb/test/API/accelerator/mock/TestMockAcceleratorPlugin.py
```diff
@@ -0,0 +1,68 @@
...
+        self.assertEqual(mock_action["plugin_name"], "mock")
```
clayborg: If we use the `get_acclerator_action` above, this would become:
```
self.assert(mock_action is not None);
```

clayborg (changes requested):
A few requested changes.

lldb/source/Utility/GDBRemoteAcceleratorPackets.cpp
satyajanga: > Maybe rename to "GDBRemoteAccelerator.cpp"? These are just utilities for defining and encode/decode to/from JSON, nothing really packet related.

@clayborg Renamed it to AcceleratorGDBRemotePackets.h, because in the subsequent PRs, I plan to add more structs (breakpoints, connection info, dynamic loader request/response etc.)  that are packet payloads.

lldb/source/Plugins/Process/gdb-remote/CMakeLists.txt
satyajanga: @JDevlieghere  I feel it is not self contained and LLDBServerAcceleratorPlugin is coupled with GDBRemoteCommunicationServerLLGS
for example 
- LLDBServerAcceleratorPlugin has a reference to GDBRemoteCommunicationServerLLGS
- GDBRemoteCommunicationServerLLGS maintains the m_accelerator_plugins.

if you strongly feel that I can try to move it.

lldb/include/lldb/Utility/GDBRemote.h
```diff
@@ -45,6 +46,28 @@ class StreamGDBRemote : public StreamString {
...
+    return PutEscapedBytes(json_string.c_str(), json_string.size());
+  }
+
+  template <class T> int PutAsJSONArray(const std::vector<T> &array) {
+    std::string json_string;
+    llvm::raw_string_ostream os(json_string);
+    os << "[";
+    for (size_t i = 0; i < array.size(); ++i) {
+      if (i > 0)
+        os << ",";
```
DavidSpickett: This feels like llvm::join with more steps, see if you could use that.

Also there is llvm::json::array and I assume you could construct one of those from the vector and it would be easier to format. I assume, I could be wrong but formatting as an array seems like a thing we'd have already written.

lldb/include/lldb/Utility/StringExtractorGDBRemote.h
```diff
@@ -173,6 +173,7 @@ class StringExtractorGDBRemote : public StringExtractor {
     eServerPacketType_jLLDBTraceGetState,
     eServerPacketType_jLLDBTraceGetBinaryData,
     eServerPacketType_jMultiBreakpoint,
+    eServerPacketType_jAcceleratorPluginInitialize,
```
DavidSpickett, line 176: New packets need documenting in https://lldb.llvm.org/resources/lldbgdbremote.html. The page is alphabetical so it'll go right at the start of the `j`s.

DavidSpickett: Include in the description the name of the qsupported feature that is required.

lldb/source/Utility/AcceleratorGDBRemotePackets.cpp
```diff
@@ -0,0 +1,27 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "lldb/Utility/AcceleratorGDBRemotePackets.h"
+
+using namespace lldb_private;
+
+llvm::json::Value lldb_private::toJSON(const AcceleratorActions &data) {
+  return llvm::json::Object{
+      {"plugin_name", data.plugin_name},
+      {"session_name", data.session_name},
```
DavidSpickett, line 16: Some things for you to document:
* plugin_name can presumably not be unique sometimes. If I had 2 GPUs there might be 2 NvidiaWhatever plugins, but they will have a different session name or identifier?
* session_name - I wonder about the scoping, is session_name the "outer scope" within which the identifiers live, or are identifiers global so you can just use the identifier to pick out a specific action on a specific accelerator

lldb/source/Utility/AcceleratorGDBRemotePackets.cpp
```diff
@@ -0,0 +1,27 @@
...
+      {"identifier", static_cast<int64_t>(data.identifier)},
```
DavidSpickett: I have not read the whole code yet, but - why static cast to int64_t can we not "just" store it in that form?

lldb/source/Utility/AcceleratorGDBRemotePackets.cpp
DavidSpickett: I see below that all 3 are required, so mention that too.

lldb/test/API/accelerator/mock/TestMockAcceleratorPlugin.py
```diff
@@ -0,0 +1,74 @@
...
+        context = self.expect_gdbremote_sequence()
+
+        raw_response = context.get("accel_init_response")
+        self.assertIsNotNone(raw_response)
+
+        decoded = self.decode_gdbremote_binary(raw_response)
+        actions = json.loads(decoded)
+        self.assertIsInstance(actions, list)
+
+        mock_action = get_accelerator_action(actions, "mock")
```
DavidSpickett, line 71: Just inline this function. You're only using it once at the moment and the absence of anything checking the "plugin_name" key in the current function is a bit suspicious.

It's not wrong as it is - it just takes more time than it should to realise what's being done here.

lldb/tools/lldb-server/CMakeLists.txt
```diff
@@ -34,6 +34,17 @@ else()
   list(APPEND LLDB_PLUGINS lldbPluginObjectFileELF)
 endif()
 
+option(LLDB_ENABLE_MOCK_ACCELERATOR_PLUGIN
+  "Enable the mock accelerator plugin for testing" OFF)
```
DavidSpickett: I'm not sure how to word this, but there's a part missing from this.

If you enable this, the built lldb-server will contain the mock plugin, right? So it's always going to say "I have an accelerator plugin". Meaning if you actually ship that, you've got 2 extra packets being sent to do nothing.

Unless I have that wrong, and nothing will ever cause the mock plugin to activate, and so there's no overhead in packets sent.

For builds that will also have a real plugin, I assume all it means is the list of plugins is one bigger which is fine, you're querying them anyway and you'll just ignore "mock".

For anyone else, this option's description should note that this is not zero overhead. It does result in more code in the lldb-server binary, however small that is.

lldb/source/Utility/AcceleratorGDBRemotePackets.cpp
DavidSpickett: Include whether there can be multiple accelerator plugins and whether those can have overlapping names / IDs / etc.

DavidSpickett (commented):
> NOTE: Sorry for the big patch, We tried to make self contained as much as possible.

I thought it was quite compact actually. Maybe because the changes are varied. The more similar the changes are to each other, the less tolerance I have for a large PR.

> jAcceleratorPluginInitialize GDB remote packet and its implementation in handlers, request and response.

Ultimately I don't care too much about the name, so this is more about understanding, but can you tell me what `jAcceleratorPluginInitialize` actually initializes and what work is involved in that?

At a glance it seems that it is more like `jAcceleratorPluginList`. The result may be used to initialize something in `lldb` but that's like calling memory read `memoryReadBufferInitialize`. The server doesn't care that the client is going to do whatever action with the result in that case.

If it looks this way just because the mock plugin is, a mock plugin, fair enough. If you could give some examples of work that needs to be done at this step for real plugins, that would be very helpful.

satyajanga:
> > NOTE: Sorry for the big patch, We tried to make self contained as much as possible.
> 
> I thought it was quite compact actually. Maybe because the changes are varied. The more similar the changes are to each other, the less tolerance I have for a large PR.
> 
> > jAcceleratorPluginInitialize GDB remote packet and its implementation in handlers, request and response.
> 
> Ultimately I don't care too much about the name, so this is more about understanding, but can you tell me what `jAcceleratorPluginInitialize` actually initializes and what work is involved in that?
> 
> At a glance it seems that it is more like `jAcceleratorPluginList`. The result may be used to initialize something in `lldb` but that's like calling memory read `memoryReadBufferInitialize`. The server doesn't care that the client is going to do whatever action with the result in that case.
> 
> If it looks this way just because the mock plugin is, a mock plugin, fair enough. If you could give some examples of work that needs to be done at this step for real plugins, that would be very helpful.

Good question — for real plugins (e.g., AMD GPU in the clayborg/llvm-server-plugins branch), the response includes breakpoints to set in the native process, connection info for creating a secondary GPU debug session, and synchronization flags. The client acts on this response by setting breakpoints on GPU driver functions, connecting to the GPU GDB server, and configuring how CPU/GPU debugging are coordinated. 
example code pointer: 
https://github.com/clayborg/llvm-project/blob/llvm-server-plugins/lldb/tools/lldb-server/Plugins/AMDGPU/LLDBServerPluginAMDGPU.cpp#L771C36-L771C56

Future patches will add these fields to AcceleratorActions as we upstream more of the protocol.

lldb/test/API/accelerator/mock/TestMockAcceleratorPlugin.py
satyajanga: > Just inline this function. You're only using it once at the moment and the absence of anything checking the "plugin_name" key in the current function is a bit suspicious.
> 
> It's not wrong as it is - it just takes more time than it should to realise what's being done here.

There was an earlier comment  from @clayborg suggesting to add this method.

lldb/source/Utility/AcceleratorGDBRemotePackets.cpp
satyajanga: The reason for the cast was the data.identifier is uint32_t. Changed identifier to int64_t to match the JSON representation directly — no cast needed now.

lldb/include/lldb/Utility/GDBRemote.h
satyajanga: thank you. Fixed it.

lldb/tools/lldb-server/CMakeLists.txt
satyajanga: > If you enable this, the built lldb-server will contain the mock plugin, right? So it's always going to say "I have an accelerator plugin"

Yes. if you built with -DLLDB_ENABLE_MOCK_ACCELERATOR_PLUGIN=ON then the mock plugin is enabled. 

I was not originally planning to enable the mock plugin by default, but If we do not enable it by default then it will be a gap in the CI because the tests will skip and regressions wont be caught. 

It points us towards a runtime gating, I will syncup with @clayborg and followup.

lldb/test/API/accelerator/mock/TestMockAcceleratorPlugin.py
DavidSpickett: Ok I can see the logic to that. The function gives you a chance to use the name to explain what it does, but putting it in a function means there's no visible use of the key in the surrounding lines. You both are more likely to have to debug this than I, so keep it as is.

I bring these things up often because if tests fail out of the blue, people don't have context and each non-obvious thing can cost time. But in this case, this stuff isn't going to be running widely anyway.

lldb/tools/lldb-server/CMakeLists.txt
DavidSpickett: Another way to put it is that normally options to enable more tests don't change the binary being tested.

ENABLE_X_TESTS - sounds like it might require some supporting package, but lldb-server wouldn't change
ENABLE_X - sounds like it could change lldb-server, and as a result, some tests might automatically start running (LLDB_ENABLE_PYTHON for example)

I don't want to make you do a lot of work to avoid what's going to be a tiny overhead anyway. Expanding the option description would be ok I think.

(maybe you want a runtime load option later to make distribution easier though)

DavidSpickett:
Ok so it's a mix then. The server might do some work (check that the GPU is in fact plugged in?), and the client gets a list of tasks that it must do to initialize its use of the plugin.

Understood, keep the name as it is.

lldb/docs/resources/lldbgdbremote.md
```diff
@@ -120,6 +120,47 @@ In any case, if we want the normal detach behavior we will just send:
...
+LLDB SENDS:    jAcceleratorPluginInitialize
+STUB REPLIES:  [{"plugin_name":"mock","session_name":"","identifier":0}]
+```
+
+If no accelerator plugins are installed, the server does not advertise the
+`accelerator-plugins+` feature and this packet should not be sent.
+
+In future patches, each `accelerator_action` will include additional fields
+such as breakpoints to set in the native process, connection info for
+secondary debug sessions, and synchronization options.
```
DavidSpickett, line 164: Please describe here / somewhere how the plugins act if the accelerator isn't present.

Perhaps that's not an issue for this packet and finding out that the GPU is unplugged will happen when lldb tries to execute one of the actions?

lldb/docs/resources/lldbgdbremote.md
```diff
@@ -120,6 +120,47 @@ In any case, if we want the normal detach behavior we will just send:
...
+LLDB SENDS:    jAcceleratorPluginInitialize
+STUB REPLIES:  [<accelerator_action>,...]
+```
+
+Each `accelerator_action` is a JSON object with the following required fields:
+
+| Key            | Type    | Description |
+|----------------|---------|-------------|
+| `plugin_name`  | string  | Unique name identifying the accelerator plugin (e.g. `"mock"`, `"amdgpu"`). Each installed plugin has a distinct name. |
+| `session_name` | string  | Name for the DAP debug session associated with this plugin. May be empty. |
```
DavidSpickett: Please explain why this needs its own DAP session ID. It seems like a thing lldb-dap should be taking care of, when the underlying lldb has multiple targets, but I don't know a lot about how DAP works.

Putting a DAP thing in the gdb-remote protocol seems strange on the surface.

lldb/docs/resources/lldbgdbremote.md
```diff
@@ -120,6 +120,47 @@ In any case, if we want the normal detach behavior we will just send:
...
+| `identifier`   | integer | Unique identifier for this action. Each action from a given plugin has a distinct identifier. |
```
DavidSpickett: Please make it more obvious how this nests. I think you mean that when I want to run an action I provide plugin name + identifier. So identifiers are unique within that plugin's "namespace".

> Each action from a given plugin has a distinct identifier.

Doesn't definitively say whether they're globally unique or per plugin. At least when I read it it doesn't.

lldb/docs/resources/lldbgdbremote.md
```diff
@@ -120,6 +120,47 @@ In any case, if we want the normal detach behavior we will just send:
 D
 ```
 
+## jAcceleratorPluginInitialize
+
+This packet requests initialization data from all accelerator plugins
+installed in lldb-server. Accelerator plugins allow lldb-server to support
+debugging of hardware accelerators (e.g. GPUs, FPGAs) alongside the native
+host process. This packet is only sent when the server reports the
+`accelerator-plugins+` feature in its `qSupported` response.
```
DavidSpickett: Would be good to add when you expect this to be sent. I assume as early as possible, even right after getting qSupported, but generally before creating an inferior process.

lldb/docs/resources/lldbgdbremote.md
DavidSpickett: Also is it possible that lldb has multiple cpu+accelerator targets being debugged and so the session name will clash?

Or do we know that there will be maximum 1 cpu+accelerator session in lldb.

lldb/docs/resources/lldbgdbremote.md
satyajanga: updated it.

lldb/docs/resources/lldbgdbremote.md
satyajanga: Thanks for raising this.
 It is not just for DAP.,any interface can use this. It's a human-readable label for distinguishing accelerator targets from the CPU target. Since  the plugin has this information we are sending this from the plugin.
Updated the description

lldb/docs/resources/lldbgdbremote.md
satyajanga: updated.

lldb/docs/resources/lldbgdbremote.md
satyajanga: updated it.

clayborg:
> > NOTE: Sorry for the big patch, We tried to make self contained as much as possible.
> 
> I thought it was quite compact actually. Maybe because the changes are varied. The more similar the changes are to each other, the less tolerance I have for a large PR.
> 
> > jAcceleratorPluginInitialize GDB remote packet and its implementation in handlers, request and response.
> 
> Ultimately I don't care too much about the name, so this is more about understanding, but can you tell me what `jAcceleratorPluginInitialize` actually initializes and what work is involved in that?
> 
> At a glance it seems that it is more like `jAcceleratorPluginList`. The result may be used to initialize something in `lldb` but that's like calling memory read `memoryReadBufferInitialize`. The server doesn't care that the client is going to do whatever action with the result in that case.

It allows any GPU plug-ins to initialize themselves. Right now they are just declaring themselves, but in the future the plug-ins will return GPUActions (will be renamed to AccleratorActions for upstream) which can set breakpoints and send back any LLDB GPU settings. See this branch for fully working AMD and Mock GPU debugging:

https://github.com/clayborg/llvm-project/tree/llvm-server-plugins

> 
> If it looks this way just because the mock plugin is, a mock plugin, fair enough. If you could give some examples of work that needs to be done at this step for real plugins, that would be very helpful.

Real GPU plug-ins will return breakpoints that need to be set in the native program that is running. This allows the native program to hit breakpoints in the GPU driver at important times in the lifetime of a code that will run on the GPU. The breakpoints are internal breakpoints with callbacks set on them that will eventually call into the GPU plug-in and let them know that their breakpoint was hit. This allows complete control over the GPU and avoids race conditions. These breakpoints are used to:
- find out if and when the GPU is enabled from the native program and that callback can return more GPUActions (more breakpoints and other things) in response to this. We don't want a GPU process to show up in LLDB unless the main program actually runs the GPU
- load/unload modules in the GPU process
- detect when the first kernel is launched as just because the GPU connection was initialized, it doesn't mean there is something to debug yet

So checkout the branch URL and have a look a complete working system with tests if you need to see what will be coming. There is a AMD plug-in and a more complete Mock GPU plug-in with full testing of all of the functionality in there with address space support, AMD debugging and more.

DavidSpickett:
Thanks, I get the picture now and I don't have any objections to it.

One thing I had been thinking about is how you knew which plugin to "select". For example if I see AMD and Nvidia plugins. But those breakpoints do that job I think. You have to hit one of those initial breakpoints to start using the plugin so if the program never hits it, or doesn't contain the symbol to break on in the first place, there's no overhead.

And if you have a program that uses multiple different types of accelerator, whether it picks one at runtime, or uses everything it can find, it sounds like this scheme would handle that too. Each accelerator would be an extra target.

DavidSpickett:
I defer to others for the actual code review.

And on the mock server thing - just make it clear in the description of the option that the mock server does end up in the lldb-server binary. At this time, this is all opt-in so if folks aren't reading what they're opting into, that's on them. In a future where there are real plugins upstream, it won't be a problem.

satyajanga:
@JDevlieghere could you please review this as well !

lldb/docs/resources/lldbgdbremote.md
```diff
@@ -168,7 +214,7 @@ fetch detailed information in batches, to keep the size of the
 packets from becoming too large.
 
 
-And the second form of jGetLoadedDynamicLibrariesInfos 
+And the second form of jGetLoadedDynamicLibrariesInfos
```
clayborg: revert unrelated changes

lldb/docs/resources/lldbgdbremote.md
```diff
@@ -1193,7 +1234,7 @@ In the example above, three lldb extensions are shown:
     * A list of compression types that the stub can use to compress packets
     when the QEnableCompression packet is used to request one of them.
   * `SupportedWatchpointTypes=<item,item,...>`
-    * A list of watchpoint types that this stub can manage. Currently defined 
+    * A list of watchpoint types that this stub can manage. Currently defined
```
clayborg: revert unrelated changes

lldb/docs/resources/lldbgdbremote.md
```diff
@@ -1523,7 +1564,7 @@ tuples to return are:
   region that are "dirty" -- they have been modified.
   Page addresses are in base 16. The size of a page can
   be found from the `qHostInfo`'s `page-size` key-value.
-  
+
```
clayborg: revert unrelated changes

lldb/docs/resources/lldbgdbremote.md
```diff
@@ -1603,7 +1644,7 @@ The request packet has the fields:
    1. mode bits in base 16
    2. file path in ascii-hex encoding
 
-Reply: 
+Reply:
```
clayborg: revert unrelated changes

lldb/include/lldb/Utility/GDBRemote.h
```diff
@@ -45,6 +46,25 @@ class StreamGDBRemote : public StreamString {
...
+  template <class T> int PutAsJSON(const T &obj, bool hex_ascii) {
+    std::string json_string;
+    llvm::raw_string_ostream os(json_string);
+    os << llvm::json::Value(toJSON(obj));
+    if (hex_ascii)
+      return PutStringAsRawHex8(json_string);
+    return PutEscapedBytes(json_string.c_str(), json_string.size());
+  }
+
+  template <class T> int PutAsJSONArray(const std::vector<T> &array) {
```
clayborg: Seems like we would want the "bool hex_ascii" here as well?

lldb/source/Plugins/Process/gdb-remote/LLDBServerAcceleratorPlugin.h
```diff
@@ -0,0 +1,43 @@
...
+class LLDBServerAcceleratorPlugin {
+public:
+  using GDBServer = process_gdb_remote::GDBRemoteCommunicationServerLLGS;
+
+  LLDBServerAcceleratorPlugin(GDBServer &gdb_server, MainLoop &main_loop);
+  virtual ~LLDBServerAcceleratorPlugin();
+
+  virtual llvm::StringRef GetPluginName() = 0;
+
+  virtual AcceleratorActions GetInitializeActions() = 0;
```
clayborg: We might want this to return a `std::optional<AcceleratorActions>` to allow GPU plug-ins to not return anything if the GPU they need to debug is not on available on the current machine. This will allow plug-ins to do some quick checks to see if hardware is available and return `std::nullopt` if hardware isn't available.

clayborg (changes requested):

lldb/tools/lldb-server/CMakeLists.txt
satyajanga: Thank you for the feedback. updated the description for now.

If needed I will come back to doing runtime loading in the future.

clayborg (approved):

PR #198909, opened 2026-05-20
[AArch64] Use GPR64commonRegClass for finding callee save regs, not GPR64RegClass

AZero13:
GPR64RegClass includes XZR/WZR and SP/WSP, which we cannot and should not use.

### DIFF
--- a/llvm/lib/CodeGen/DeadMachineInstructionElim.cpp
+++ b/llvm/lib/CodeGen/DeadMachineInstructionElim.cpp
@@ -32,7 +32,7 @@ namespace {
 class DeadMachineInstructionElimImpl {
   const MachineRegisterInfo *MRI = nullptr;
   const TargetInstrInfo *TII = nullptr;
-  LiveRegUnits LivePhysRegs;
+  LiveRegUnits LiveRegs;
 
 public:
   bool runImpl(MachineFunction &MF);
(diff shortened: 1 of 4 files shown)


AZero13:
@jayfoad I think this is what I should have done

jayfoad:
> I think this is what I should have done

I am not sure why you are tagging me. Is this referring back to some other PR? If so, I have forgotten it, sorry!

llvm/lib/Target/AArch64/AArch64FrameLowering.cpp
```diff
@@ -922,7 +922,9 @@ AArch64FrameLowering::findScratchNonCalleeSaveRegister(MachineBasicBlock *MBB,
   if (LiveRegs.available(MRI, AArch64::X9))
     return AArch64::X9;
 
-  for (unsigned Reg : AArch64::GPR64RegClass) {
+  // GPR64common excludes XZR/WZR and SP/WSP, which are not valid scratch regs
+  // for stack adjustment (GPR64sp) instructions.
+  for (MCPhysReg Reg : AArch64::GPR64commonRegClass) {
```
MacDue, line 927: Is this an NFC or have you encountered a case where XZR was selected?

llvm/lib/CodeGen/DeadMachineInstructionElim.cpp
```diff
@@ -32,7 +32,7 @@ namespace {
 class DeadMachineInstructionElimImpl {
   const MachineRegisterInfo *MRI = nullptr;
   const TargetInstrInfo *TII = nullptr;
-  LiveRegUnits LivePhysRegs;
+  LiveRegUnits LiveRegs;
```
MacDue, line 35: The other changes in this PR seem like unrelated NFCs.

PR #198910, opened 2026-05-20
[CIR] Lower member-pointer casts in constant emitter

adams381:
Depends on [#197555](https://github.com/llvm/llvm-project/pull/197555).

Base branch is `cir-data-member-offset-nfc` on the fork. The diff is only the two cast commits on top of that member_path work.

ConstExprEmitter and CIRGenItaniumCXXABI lower `CK_BaseToDerivedMemberPointer`, `CK_DerivedToBaseMemberPointer`, and `CK_ReinterpretMemberPointer` for data-member pointers using GEP-like `member_path` attributes. Runtime scalar casts use `emitMemberPointerConversion` on the CXXABI, matching classic CodeGen.

Supersedes closed [#197582](https://github.com/llvm/llvm-project/pull/197582) (`cir-data-member-cast` on `main`, offset-based).

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
@@ -933,11 +933,18 @@ public:
     switch (e->getCastKind()) {
     case CK_ToUnion:
     case CK_AddressSpaceConversion:
-    case CK_ReinterpretMemberPointer:
-    case CK_DerivedToBaseMemberPointer:
-    case CK_BaseToDerivedMemberPointer:
       cgm.errorNYI(e->getBeginLoc(), "ConstExprEmitter::VisitCastExpr");
       return {};
+
+    case CK_ReinterpretMemberPointer:
+    case CK_DerivedToBaseMemberPointer:
+    case CK_BaseToDerivedMemberPointer: {
+      mlir::Attribute srcAttr =
+          emitter.tryEmitPrivate(subExpr, subExpr->getType());
+      if (!srcAttr)
+        return {};
+      return cgm.getCXXABI().emitMemberPointerConversion(e, srcAttr);
+    }
 
     case CK_LValueToRValue:
     case CK_AtomicToNonAtomic:
--- a/clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
@@ -1926,11 +1926,6 @@ mlir::Attribute ConstantEmitter::tryEmitPrivate(const APValue &value,
 
     const auto *fieldDecl = cast<FieldDecl>(memberDecl);
     const auto *mpt = destType->castAs<MemberPointerType>();
-    const CXXRecordDecl *pointerClass = mpt->getMostRecentCXXRecordDecl();
-    const auto *fieldParent = cast<CXXRecordDecl>(fieldDecl->getParent());
-    if (fieldParent != pointerClass)
-      return {};
-
     return cgm.getDataMemberAttrForField(mpt, fieldDecl);
   }
   case APValue::LValue:
--- a/clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
@@ -1907,12 +1907,6 @@ mlir::Attribute ConstantEmitter::tryEmitPrivate(const APValue &value,
     if (!memberDecl)
       return builder.getZeroInitAttr(cgm.convertType(destType));
 
-    if (value.isMemberPointerToDerivedMember()) {
-      cgm.errorNYI(
-          "ConstExprEmitter::tryEmitPrivate member pointer to derived member");
-      return {};
-    }
-
     if (auto const *cxxDecl = dyn_cast<CXXMethodDecl>(memberDecl)) {
       auto ty = mlir::cast<cir::MethodType>(cgm.convertType(destType));
       if (cxxDecl->isVirtual())
@@ -1923,10 +1917,14 @@ mlir::Attribute ConstantEmitter::tryEmitPrivate(const APValue &value,
       return cgm.getBuilder().getMethodAttr(ty, methodFuncOp);
     }
 
-    auto cirTy = mlir::cast<cir::DataMemberType>(cgm.convertType(destType));
-
     const auto *fieldDecl = cast<FieldDecl>(memberDecl);
-    return builder.getDataMemberAttr(cirTy, fieldDecl->getFieldIndex());
+    const auto *mpt = destType->castAs<MemberPointerType>();
+    const CXXRecordDecl *pointerClass = mpt->getMostRecentCXXRecordDecl();
+    const auto *fieldParent = cast<CXXRecordDecl>(fieldDecl->getParent());
+    if (fieldParent != pointerClass)
+      return {};
+
+    return cgm.getDataMemberAttrForField(mpt, fieldDecl);
   }
   case APValue::LValue:
     return ConstantLValueEmitter(*this, value, destType).tryEmit();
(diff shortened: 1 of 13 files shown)


adams381:
Incremental diff for review (cast commits only, on top of #197555):

https://github.com/llvm/llvm-project/compare/adams381:cir-data-member-offset-nfc...adams381:cir-data-member-cast-stacked

GitHub shows the full branch vs `main` in the PR diff because the base branch lives on the fork. Use the compare link above for the stacked view.

clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
```diff
@@ -933,12 +933,24 @@ class ConstExprEmitter
...
       return {};
 
+    case CK_ReinterpretMemberPointer: {
+      mlir::Attribute src = emitter.tryEmitPrivate(subExpr, subExpr->getType());
+      if (!src)
+        return {};
+      return cgm.getCXXABI().emitMemberPointerConversion(e, src);
+    }
+
+    case CK_DerivedToBaseMemberPointer:
```
erichkeane, line 940: What is the difference between this and the ReinterpretMemberPointer version that we can't just use the same impl for both?

clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
```diff
@@ -1907,13 +1919,15 @@ mlir::Attribute ConstantEmitter::tryEmitPrivate(const APValue &value,
...
       return builder.getZeroInitAttr(cgm.convertType(destType));
 
-    if (value.isMemberPointerToDerivedMember()) {
-      cgm.errorNYI(
-          "ConstExprEmitter::tryEmitPrivate member pointer to derived member");
-      return {};
-    }
-
     if (auto const *cxxDecl = dyn_cast<CXXMethodDecl>(memberDecl)) {
+      // Member function pointers with a derived-member path adjustment use
```
erichkeane: Why did this move from above?

clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
```diff
@@ -1926,7 +1940,11 @@ mlir::Attribute ConstantEmitter::tryEmitPrivate(const APValue &value,
     auto cirTy = mlir::cast<cir::DataMemberType>(cgm.convertType(destType));
 
     const auto *fieldDecl = cast<FieldDecl>(memberDecl);
-    return builder.getDataMemberAttr(cirTy, fieldDecl->getFieldIndex());
+    const auto *mpt = destType->castAs<MemberPointerType>();
+    const CXXRecordDecl *pointerClass = mpt->getMostRecentCXXRecordDecl();
+    llvm::SmallVector<unsigned, 4> path;
+    cgm.buildDataMemberPath(pointerClass, fieldDecl, path);
```
erichkeane: We've now done this a pair of times, we should extract this to a function in CIRGenModule that takes the destType and fieldDecl, and does this.

clang/lib/CIR/CodeGen/CIRGenItaniumCXXABI.cpp
```diff
@@ -2398,6 +2405,127 @@ CIRGenItaniumCXXABI::buildVirtualMethodAttr(cir::MethodType methodTy,
...
+  e = e->IgnoreParenImpCasts();
+  if (const auto *declRef = dyn_cast<DeclRefExpr>(e))
+    return dyn_cast<FieldDecl>(declRef->getDecl());
+  if (const auto *memberExpr = dyn_cast<MemberExpr>(e))
+    return dyn_cast<FieldDecl>(memberExpr->getMemberDecl());
+  if (const auto *unaryOp = dyn_cast<UnaryOperator>(e)) {
+    if (unaryOp->getOpcode() == UO_AddrOf)
+      return getDataMemberFieldFromExpr(unaryOp->getSubExpr());
+  }
+  if (const auto *castExpr = dyn_cast<CastExpr>(e))
```
erichkeane: This actually needs to be an 'else-if' now.  Alternatively (and probably better), just move the unaryOp to the end.

clang/lib/CIR/CodeGen/CIRGenItaniumCXXABI.cpp
```diff
@@ -2398,6 +2405,127 @@ CIRGenItaniumCXXABI::buildVirtualMethodAttr(cir::MethodType methodTy,
...
+      return {};
+    }
+    return cgm.getBuilder().getDataMemberAttr(destTy, path);
+  }
+
+  if (const FieldDecl *field = getDataMemberFieldFromExpr(e->getSubExpr())) {
+    const CXXRecordDecl *destClass =
+        e->getType()->castAs<MemberPointerType>()->getMostRecentCXXRecordDecl();
+    llvm::SmallVector<unsigned, 4> path;
+    cgm.buildDataMemberPath(destClass, field, path);
```
erichkeane: Thrice now!

clang/lib/CIR/CodeGen/CIRGenModule.cpp
```diff
@@ -2235,8 +2285,12 @@ mlir::Value CIRGenModule::emitMemberPointerConstant(const UnaryOperator *e) {
   // Otherwise, a member data pointer.
   auto ty = mlir::cast<cir::DataMemberType>(convertType(e->getType()));
   const auto *fieldDecl = cast<FieldDecl>(decl);
-  return cir::ConstantOp::create(
-      builder, loc, builder.getDataMemberAttr(ty, fieldDecl->getFieldIndex()));
+  const auto *mpt = cast<MemberPointerType>(e->getType().getTypePtr());
+  const CXXRecordDecl *pointerClass = mpt->getMostRecentCXXRecordDecl();
+  llvm::SmallVector<unsigned, 4> path;
+  buildDataMemberPath(pointerClass, fieldDecl, path);
```
erichkeane: Four times, ah-ah-ah!

clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
adams381: All three cases (`CK_ReinterpretMemberPointer`, `CK_DerivedToBaseMemberPointer`, `CK_BaseToDerivedMemberPointer`) are now unified in `VisitCastExpr` — they all call `cgm.getCXXABI().emitMemberPointerConversion(e, srcAttr)`. The distinction is inside `emitMemberPointerConversion`: `CK_ReinterpretMemberPointer` locates the field by byte offset (needed when source and dest classes differ), while the base/derived casts peel the source expression to find the field decl directly.

clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
adams381: The  check was replaced rather than moved. It was a blanket NYI on all derived-member s. The new code handles the common sub-case (field declared directly in the destination class, e.g., `static_cast<B::*>(&D::f)` where `f` is in `B`) by doing the specific check `fieldParent != destClass` inside `getDataMemberAttrForField`, returning `{}` silently for the cross-class case so `tryEmitPrivate` falls back to non-constant evaluation. The wider derived-member path case (where the field really is in a different class from the pointer's target) still returns `{}` — it's just now decided at the field-lookup boundary instead of up front.

clang/lib/CIR/CodeGen/CIRGenExprConstant.cpp
adams381: Extracted — all four call sites now go through `cgm.getDataMemberAttrForField(mpt, fieldDecl)`, which handles the path building and the `fieldParent != destClass` guard internally.

clang/lib/CIR/CodeGen/CIRGenItaniumCXXABI.cpp
adams381: Moved `castExpr` before `unaryOp` in the else-if chain.

clang/lib/CIR/CodeGen/CIRGenItaniumCXXABI.cpp
adams381: Consolidated — same `getDataMemberAttrForField` helper.

clang/lib/CIR/CodeGen/CIRGenModule.cpp
adams381: Consolidated — same `getDataMemberAttrForField` helper.

adams381:
Superseded by [#200854](https://github.com/llvm/llvm-project/pull/200854), which covers CK_ReinterpretMemberPointer as part of the broader DataMemberAttr GEP-path redesign.

PR #198913, opened 2026-05-20
[Clang] [Docs] Remove stray release note

Sirraide:
The patch that added this release note was reverted (#198341), but then (#198167) accidentally added it back.


Sirraide:
CC @shafik

PR #198914, opened 2026-05-20
[RISCV][P-ext] Return SDValue() instead of calling DAG.UnrollVectorOp for non-splat shift amounts. NFCI

topperc:
The generic legalizer can call DAG.UnrollVectorOp itself.


lenary (approved):

ppenzin (approved):
LGTM

PR #193585, opened 2026-04-22
[rtsan] Allow for rtsan/ubsan runtime combos

cjappl:
All changes (and tests) necessary for `-fsanitize=realtime,undefined` to work as intended.

For @davidtrevelyan, one quirk is that if UBSan detects UB in a real-time thread, we a "nested failure" report from UBSan. This is due to UBSan allocating and printing (which isn't realtime safe). I think this behavior is OK, as I suspect users wanting this config want to squash all of their issues. I'm calling it out explicitly though

Many thanks to @gbMattN for the template in #169036

PR #198916, opened 2026-05-20
[ubsan] Introduce weak handle_error_begin and end hooks

cjappl:
Over on #193585 , we are trying to enable RTSan and UBSan as a combo sanitizer.

However, when UBSan hits an issue, it prints, which is real-time-unsafe and leads to an RTSan report. These "double reports" are noisy and don't provide the right info to the user. To mitigate this issue, we need a way to temporarily disable RTSan when UBSan is doing its thing.

This solution provides hooks that are called before and after UBSan running. This can then be used by us in RTSan to disable/re-enable our detection mechanism and not have cascading errors on a report. 

Completely open to other ways of solving this problem, this is the way I could think of that minimizes direct dependencies and is most flexible for other sanitizers in the future.


fmayer:
Please fix formatting and test. Otherwise looks fine to me

davidtrevelyan:
Nice idea to put the instantiation of the `ScopedHandleErrorHook` inside `GET_REPORT_OPTIONS`. This LGTM, but I'm not well-placed to approve changes in UBSan

fmayer (approved):

compiler-rt/lib/ubsan/ubsan_diag.h
```diff
@@ -211,9 +211,26 @@ struct ReportOptions {
...
-#define GET_REPORT_OPTIONS(unrecoverable_handler) \
-    GET_CALLER_PC_BP; \
-    ReportOptions Opts = {unrecoverable_handler, pc, bp}
+// Called when a UBSan handler begins processing an error.
+// Used internally for running sanitizers at the same time as ubsan
+extern "C" SANITIZER_INTERFACE_ATTRIBUTE void __ubsan_handle_error_begin(void);
+
+// Called when a UBSan handler finishes processing an error.
+extern "C" SANITIZER_INTERFACE_ATTRIBUTE void __ubsan_handle_error_end(void);
+
```
vitalybuka, line 220: Why ScopedReport was not enough?

davidtrevelyan: @vitalybuka I think putting it in `ScopedReport` is functionally enough at the moment, but I just had a (nonblocking) concern that this could prove a bit brittle, and proposed disentangling error hooks from error reporting in my comment [here](https://github.com/llvm/llvm-project/pull/193585#issuecomment-4497507902). The high-level summary of my concern was that `ScopedReport` is currently the only thing in UBSan's error handling that does "real-time unsafe" things (and therefore the only thing RTSan needs to be disabled for), and I don't think UBSan developers should have to be concerned by disabling RTSan for anything new they wish to add in future. If they'd like to do something else that allocates in a UBSan error handler, for example, I don't want failing RTSan tests to get in the way. By lifting RTSan's hooking to the top level of UBSan's error handling, I believe we can eliminate this concern.

To reiterate, this was a nonblocking concern and I don't feel strongly about it. If you prefer the `ScopedReport` hook placement, I trust your judgement and it's completely fine by me 👍

vitalybuka: If it works, I'd prefer to keep it in ScopedReport. To avoid mental effort from readers guessing why we need ScopedHandleErrorHook vs ScopedReport.

vitalybuka (approved):

compiler-rt/lib/ubsan/ubsan_diag.h
davidtrevelyan: Sure - seems reasonable to me - very happy to go with your preference on this. @cjappl @fmayer are you both happy if we close this PR and merge https://github.com/llvm/llvm-project/pull/193585 instead?

davidtrevelyan: Apologies, I misremembered the contents of https://github.com/llvm/llvm-project/pull/193585: I mistakenly thought it contained the patch shared by @cjappl earlier here: https://github.com/cjappl/llvm-project/compare/cjappl/rtsan_ubsan_combo_supported...cjappl:llvm-project:cjappl/ubsan_helper_method, and it's this latter patch that I was referring to in my linked comment with the nonblocking concern [here](https://github.com/llvm/llvm-project/pull/193585#issuecomment-4497507902).

To summarise, https://github.com/llvm/llvm-project/pull/193585 suffers from a nested report problem. We think solving the nested report problem needs entry/exit hooks in ubsan's error reporting - _somewhere_ - and I believe our choice is either:

1. hooks go in `ScopedReport` as `__ubsan_on_report_begin/end` (for which we need to consider how they live alongside, perhaps confusingly, the existing `__ubsan_on_report`), vs
2. hooks go in at a higher level (e.g. `GET_REPORT_OPTIONS`) as `__ubsan_handle_error_begin/end`.

Sorry for the confusion. I marginally prefer option 2. above, but don't have strong feelings and happy with 1. @cjappl does this match with your understanding?

cjappl: Yes, correct. The other option is if a ubsan report fires in a realtime block, RTSan also reports which seems a little messy. It would be nice to solve it with one of these two solutions but that does require tinkering with ubsan a bit.

@vitalybuka , can you see a way to enable/disable rtsan before and after a ubsan report using the existing ubsan code, or is this solution or https://github.com/cjappl/llvm-project/compare/cjappl/rtsan_ubsan_combo_supported...cjappl:llvm-project:cjappl/ubsan_helper_method necessary if we want this behavior?

cjappl: Friendly ping for @vitalybuka

PR #198917, opened 2026-05-20
IR: Move `simplify_type<(const) Use *>` to Use.h and fix a bug.

pcc:
These overloads live in User.h because the type of
`User::(const_)op_iterator` is `(const) Use *`, but they do not
necessarily need to be used together with `User`, so let's move them to
Use.h with the canonical type used in the overload for clarity.

There's also a bug where `dyn_cast_or_null` with a `Use *` argument
crashes if the argument is null. Fix it by adding a null check to
these overloads. The null check is expected to be optimized out of the
implementation of `isa`/`cast`/`dyn_cast` because these functions will
unconditionally load from the result of `getSimplifiedValue`.


efriedma-quic (approved):
LGTM

PR #198918, opened 2026-05-20
[CIR] Report NYI for defaulted union copy/move assignment

adams381:
A defaulted union copy/move assignment operator has an empty synthesized body -- Sema skips union fields (the FIXME in `SemaDeclCXX::buildSingleCopyAssign`), so there is no AST expression for the implied whole-object copy. CIRGen emitted that empty body and silently dropped the copy, miscompiling by-value union assignment such as MultiSource kimwitu++/kc's `*++yyvsp = yylval`.

This reports NYI for that case instead of emitting a body that does nothing. Struct and array memcpy-equivalent assignments are unaffected: their synthesized bodies carry the implicit memberwise copies (per-field assignment expressions, or a builtin memcpy call for array members) and lower through the existing body-emission loop.

Emitting the union copy properly means implementing the Sema FIXME so the synthesized body carries an AST node for the implied memcpy; that is a front-end change that also affects classic CodeGen and will be a separate PR.


clang/test/CIR/CodeGen/trivial-union-assign.cpp
```diff
@@ -0,0 +1,36 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -O3 -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck --check-prefix=CIR --input-file=%t.cir %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -O3 -fclangir -emit-llvm %s -o %t-cir.ll
+// RUN: FileCheck --check-prefix=LLVM --input-file=%t-cir.ll %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -O3 -emit-llvm %s -o %t.ll
+// RUN: FileCheck --check-prefix=OGCG --input-file=%t.ll %s
+
+union YYSTYPE {
```
erichkeane: Can we find somewhere else to put this test, in an already existing file?

Also, s/OGCG/LLVM, etc.

erichkeane (commented):
It appears we did this on purpose based on the comment.  What is the reasoning/justification for changing this?  Presumably we intended that the calls would be more 'obvious' to opt passes in the future, so a premature opt like this could defeat any sort of lifetime analysis/etc. 

@andykaylor wrote the comment initially, so interested to see what he has to say.

clang/test/CIR/CodeGen/trivial-union-assign.cpp
adams381: > Can we find somewhere else to put this test, in an already existing file?

Also, s/OGCG/LLVM, etc.

Moved the `-O3` union stack-assign repro into `cxx-special-member-attr.cpp` and removed the standalone file.  One `LLVM` check prefix covers both CIR-lowered and classic output.

adams381:
@erichkeane — on the design question: the prior comment assumed later LLVM opts could safely remove trivial `operator=` calls.  At `-O3` the inlined no-op assign dropped the store (`kimwitu++` `kc`, `*++yyvsp = yylval`).  Classic CodeGen already uses `EmitAggregateAssign` at the site; this matches that for trivial copy/move when `!mayInsertExtraPadding()`.

erichkeane:
> @erichkeane — on the design question: the prior comment assumed later LLVM opts could safely remove trivial `operator=` calls. At `-O3` the inlined no-op assign dropped the store (`kimwitu++` `kc`, `*++yyvsp = yylval`). Classic CodeGen already uses `EmitAggregateAssign` at the site; this matches that for trivial copy/move when `!mayInsertExtraPadding()`.

It isn't JUST llvm opts, we are also going to want this information around for CIR ops.  Classic-codegen does emission at a 'lower level' necessarily, but we have all of the CIR optimizer to do these sorts of modifications.  So I don't think that is good justification to do this already.

andykaylor:
The comment was just an explanation of behavior I was bringing over from the incubator. The comment there says:

```
      // From LLVM codegen:
      // We don't like to generate the trivial copy/move assignment operator
      // when it isn't necessary; just produce the proper effect here.
      // It's important that we use the result of EmitLValue here rather than
      // emitting call arguments, in order to preserve TBAA information from
      // the RHS.
      //
      // We don't early optimize like LLVM does:
      // LValue RHS = isa<CXXOperatorCallExpr>(CE) ? TrivialAssignmentRHS
      //                                           :
      //                                           emitLValue(*CE->arg_begin());
      // emitAggregateAssign(This, RHS, CE->getType());
      // return RValue::get(This.getPointer());
```

I looked into this a bit, and the problem is that while we're emitting the call to the trivial constructor, we're emitting an effectively empty body for the trivial constructor (it just saves and load a few temporaries), so LLVM optimizes it away even at -O1.

I'd say that we either need this change or we need to do something to perform the copy in the constructor body.

clang/test/CIR/CodeGen/trivial-union-assign.cpp
adams381: Done — `trivial-union-assign.cpp` is gone. The `-O3` bison-stack repro is now a `split-file` sub-file in `cxx-special-member-attr.cpp` with a shared `LLVM:` prefix covering both the CIR-lowered and OGCG outputs.

adams381:
The incubator comment was a deviation note documenting that CIR wasn't yet applying the classic CodeGen optimization, not a commitment to omit it permanently.  The optimization is necessary: the trivial copy/move operator body as currently emitted consists of a save/load sequence LLVM removes at -O1, so the store disappears entirely at -O3 on real code — the concrete regression was `*++yyvsp = yylval` in yacc/bison stacks (see the `-O3` repro in `cxx-special-member-attr.cpp`).

Classic CodeGen handles this identically in `EmitCXXMemberOrOperatorMemberCallExpr` — it never emits the call for trivial copy/move assignment, it goes straight to `EmitAggregateAssign` at the call site.  If a future CIR optimization pass should reason about trivial assignment, the prerequisite is fixing the operator body to emit a real copy rather than the save/load sequence LLVM eliminates; that's a separate change.

adams381:
@erichkeane — test is in `cxx-special-member-attr.cpp` as a split-file sub-test; standalone file removed. Single `LLVM` prefix on both emits. The inlined assign path matches classic CodeGen; the operator body as emitted today is save/load LLVM removes at `-O1`, so call-site inlining is what keeps the store at `-O3`. PR description updated to match the folded test.

erichkeane:
It seems to me the bug isn't this not being here, it is that the constructor generated is wrong, we should probably fix that instead.

adams381:
The current change mirrors classic CodeGen rather than diverging from it.  `CodeGenFunction::EmitCXXMemberOrOperatorMemberCallExpr` takes the same trivial copy/move-assignment path: it evaluates the RHS lvalue and calls `EmitAggregateAssign` at the call site, and never emits a call to the trivial `operator=`.  This PR does exactly that.

The empty body isn't something this PR introduces — classic emits the same thing.  Forcing classic to emit the union's `operator=` body (e.g. by taking `&U::operator=`) gives, at `-O0 -emit-llvm`:

```llvm
define linkonce_odr ptr @_ZN1UaSERKS_(ptr %this, ptr %0) {
entry:
  %this.addr = alloca ptr
  %.addr = alloca ptr
  store ptr %this, ptr %this.addr
  store ptr %0, ptr %.addr
  %this1 = load ptr, ptr %this.addr
  ret ptr %this1
}
```

No memcpy — classic's trivial union `operator=` body never performs the copy.  That's harmless precisely because the call-site shortcut means the body is never what does the assignment.  So making the body copy would be a departure from classic, not a correction toward it, and the `-O3` store-deletion this PR fixes comes from the call-site path, not the body.

If we instead want CIR to keep the `operator=` call so the CIR optimizer can reason about it, backed by a body that actually copies, that's a deliberate improvement over classic and I'm happy to take that direction.  It's a separate goal from matching classic and fixing the regression, which is what this PR currently does.

erichkeane:
> The current change mirrors classic CodeGen rather than diverging from it. `CodeGenFunction::EmitCXXMemberOrOperatorMemberCallExpr` takes the same trivial copy/move-assignment path: it evaluates the RHS lvalue and calls `EmitAggregateAssign` at the call site, and never emits a call to the trivial `operator=`. This PR does exactly that.
> 
> The empty body isn't something this PR introduces — classic emits the same thing. Forcing classic to emit the union's `operator=` body (e.g. by taking `&U::operator=`) gives, at `-O0 -emit-llvm`:
> 
> ```llvm
> define linkonce_odr ptr @_ZN1UaSERKS_(ptr %this, ptr %0) {
> entry:
>   %this.addr = alloca ptr
>   %.addr = alloca ptr
>   store ptr %this, ptr %this.addr
>   store ptr %0, ptr %.addr
>   %this1 = load ptr, ptr %this.addr
>   ret ptr %this1
> }
> ```
> 
> No memcpy — classic's trivial union `operator=` body never performs the copy. That's harmless precisely because the call-site shortcut means the body is never what does the assignment. So making the body copy would be a departure from classic, not a correction toward it, and the `-O3` store-deletion this PR fixes comes from the call-site path, not the body.
> 
> If we instead want CIR to keep the `operator=` call so the CIR optimizer can reason about it, backed by a body that actually copies, that's a deliberate improvement over classic and I'm happy to take that direction. It's a separate goal from matching classic and fixing the regression, which is what this PR currently does.

I understand classic codegen does this, but I'm saying that is the wrong choice for CIR, and a path we shouldn't go down here.  

As far as:
>It's a separate goal from matching classic and fixing the regression, which is what this PR currently does.

Our goal has always been to get a functional compiler that covers language semantics, and generates them correctly.  We often use 'IR matches OGCG' as a shortcut to that, but it is not, in fact, or goal.

This is a case where the 'premature optimization' that LLVM-IR is taking (and in fact, our lowering might very well choose to make the same choice) is actively harmful to tracking lifetimes/values, which is important to CIR's goals, so we should be going down that path.

adams381:
@erichkeane — reworked to fix the body instead of eliding the call. The defaulted trivial/union `operator=` now copies the whole object in its body (a `cir.copy`, the memcpy-equivalent case), so it is no longer a no-op LLVM deletes — the `-O3` store now survives. The call site is back to its upstream form, so the `operator=` call stays and the assignment is visible to later CIR passes. Replacing such calls, if we ever want that, belongs in a CIR optimization pass rather than CIRGen.

Non-trivial assignment operators keep their member-wise body emission, and the general field-coalescing memcpyizer is still not implemented. The test now checks the body performs the copy (CIR `cir.copy`, LLVM `memcpy`) with the call kept, and contrasts it with classic CodeGen, which inlines the copy at the call site and emits no `operator=`.

clang/test/CIR/CodeGen/cxx-special-member-attr.cpp
```diff
@@ -1,5 +1,9 @@
-// RUN: %clang_cc1 -std=c++11 -triple aarch64-none-linux-android21 -fclangir -emit-cir %s -o %t.cir
-// RUN: FileCheck --check-prefix=CIR --input-file=%t.cir %s
+// RUN: split-file %s %t
```
andykaylor: It's not clear to me why you're splitting the file.

clang/lib/CIR/CodeGen/CIRGenClass.cpp
```diff
@@ -901,6 +901,33 @@ void CIRGenFunction::emitImplicitAssignmentOperatorBody(FunctionArgList &args) {
...
+                                            getAddrOfLocalVar(args.back()));
+    LValue src = makeNaturalAlignAddrLValue(srcPtr, recordTy);
+    emitAggregateAssign(dest, src, recordTy);
+
+    for (Stmt *s : rootCS->body())
+      if (isa<ReturnStmt>(s))
+        if (emitStmt(s, /*useCurrentScope=*/true).failed())
+          cgm.errorNYI(s->getSourceRange(),
+                       std::string("emitImplicitAssignmentOperatorBody: ") +
+                           s->getStmtClassName());
```
andykaylor: This is really quite concerning. It looks like the AST is currently generating a compound statement containing nothing but a return statement (at least for the test case you're adding here), but this is just blindly looking for a return statement and throwing away anything else that happens to be here. I'd feel much better about this if it at least asserted on anything that wasn't a return statement.

clang/lib/CIR/CodeGen/CIRGenClass.cpp
```diff
@@ -901,6 +901,33 @@ void CIRGenFunction::emitImplicitAssignmentOperatorBody(FunctionArgList &args) {
   assert(!cir::MissingFeatures::incrementProfileCounter());
   assert(!cir::MissingFeatures::runCleanupsScope());
 
+  // For a memcpy-equivalent special member (a union, or a trivially-copyable
+  // record) the synthesized body either copies nothing -- a union body is just
+  // `return *this` -- or relies on a memcpy the AST does not spell out as
+  // field assignments.  Mirror classic CodeGen's AssignmentMemcpyizer: copy
+  // the whole object once, then fall through to emit the trailing
+  // `return *this`.  Emitting the copy but skipping the return would leave the
+  // result reference uninitialized.
```
andykaylor:
```suggestion
  // For a memcpy-equivalent assignment operator, copy the whole object,
  // then fall through to emit the trailing `return *this`.
```

clang/lib/CIR/CodeGen/CIRGenClass.cpp
```diff
@@ -901,6 +901,33 @@ void CIRGenFunction::emitImplicitAssignmentOperatorBody(FunctionArgList &args) {
...
+  if (assignOp->isMemcpyEquivalentSpecialMember(getContext())) {
+    CanQualType recordTy =
+        getContext().getCanonicalTagType(assignOp->getParent());
+    LValue dest = makeNaturalAlignAddrLValue(loadCXXThis(), recordTy);
+    // The source is the trailing reference parameter; load it to get the
+    // referent's address before copying (mirrors the copy-constructor path).
```
andykaylor: I find this comment to be more confusing than helpful.

clang/lib/CIR/CodeGen/CIRGenClass.cpp
```diff
@@ -901,6 +901,33 @@ void CIRGenFunction::emitImplicitAssignmentOperatorBody(FunctionArgList &args) {
...
+    mlir::Value srcPtr = builder.createLoad(getLoc(assignOp->getLocation()),
+                                            getAddrOfLocalVar(args.back()));
+    LValue src = makeNaturalAlignAddrLValue(srcPtr, recordTy);
+    emitAggregateAssign(dest, src, recordTy);
+
+    for (Stmt *s : rootCS->body())
+      if (isa<ReturnStmt>(s))
+        if (emitStmt(s, /*useCurrentScope=*/true).failed())
+          cgm.errorNYI(s->getSourceRange(),
```
andykaylor: This isn't actually something that's not implemented. It's just an error. In practice, it shouldn't happen, and if it does happen because of something that's NYI, we should have emitted an error earlier. You should just call `cgm.error()` here and give it a more specific message.

clang/lib/CIR/CodeGen/CIRGenClass.cpp
adams381: Added an assert, but kept it broader than return-only on purpose: a defaulted memcpy-equivalent body isn't always just `return *this`.  A trivially-copyable struct carries per-field assignment statements, and a struct with an array member carries a `__builtin_memcpy` `CallExpr` -- both subsumed by the whole-object copy.  The loop now emits the return and asserts every other statement is an `Expr`.  Added a struct-with-array-member case to the test to pin it.

clang/lib/CIR/CodeGen/CIRGenClass.cpp
adams381: Applied your wording.

clang/lib/CIR/CodeGen/CIRGenClass.cpp
adams381: Removed it.

clang/lib/CIR/CodeGen/CIRGenClass.cpp
adams381: Switched to cgm.error with a specific message.

clang/test/CIR/CodeGen/cxx-special-member-attr.cpp
adams381: Dropped split-file -- it's now one x86_64 module with CIR, `-fclangir -emit-llvm`, and classic `-emit-llvm` checks.  The split was only there because the attribute checks were written for aarch64 while the union codegen needs x86_64; the attribute and mangled names are target-independent, so the separate aarch64 run wasn't adding coverage.

clang/lib/CIR/CodeGen/CIRGenClass.cpp
andykaylor: I thought the whole point of this change was that the AST didn't contain the necessary expressions. For example, for the union test case, I'm seeing this:

```
| |-CXXMethodDecl <line:1:7> col:7 implicit used constexpr operator= 'YYSTYPE &(const YYSTYPE &) noexcept' inline default trivial external-linkage
| | |-ParmVarDecl <col:7> col:7 'const YYSTYPE &'
| | `-CompoundStmt <col:7>
| |   `-ReturnStmt <col:7>
| |     `-UnaryOperator <col:7> 'YYSTYPE' lvalue prefix '*' cannot overflow
| |       `-CXXThisExpr <col:7> 'YYSTYPE *' this
```
What I'm wondering is whether the real fix for this should be in the AST rather than in CIR codegen. I do see that the array case you added contains a builtin memcpy call, but that just reinforces my belief that the union case should have a more complete body in the AST. The fact that we're assuming that an aggregate copy is needed in all cases and ignoring something the AST provided in a subset of those cases just seems entirely wrong.

andykaylor: More evidence that the fix should be made in the AST construction:

https://github.com/llvm/llvm-project/blob/4f55655a932133fe49a8fb0e3e1f622f2ead66a3/clang/lib/Sema/SemaDeclCXX.cpp#L15474

```
    // FIXME: We should form some kind of AST representation for the implied
    // memcpy in a union copy operation.
    if (Field->isUnnamedBitField() || Field->getParent()->isUnion())
      continue;
```
This is why the copy assignment operator body is empty for unions. The move assignment operator contains the same code and FIXME.

adams381: This does belong in the AST.  The union body is empty only because Sema skips union fields at that FIXME, so no front end gets a copy expression to lower.  Classic gets away with it by eliding the memcpy-equivalent call at the call site and emitting the memcpy there; the `operator=` body it would generate is also just `return *this`.  This PR stopped eliding the call (per earlier review), which is exactly what exposed the empty body.

So instead of reconstructing the copy in CIRGen, I'll implement the FIXME: have Sema form an AST representation for the implied memcpy in union copy and move assignment, so the synthesized body contains it.  That fixes classic CodeGen and CIR together and lets this PR drop the whole-object-copy special case.

Since it changes shared AST and will shift classic codegen output (and some existing tests), I'd land it as a separate precursor PR and rebase this one on top.  Does that scope sound right before I start?

clang/lib/CIR/CodeGen/CIRGenClass.cpp
```diff
@@ -901,6 +901,36 @@ void CIRGenFunction::emitImplicitAssignmentOperatorBody(FunctionArgList &args) {
...
+  // For a memcpy-equivalent assignment operator, copy the whole object, then
+  // fall through to emit the trailing `return *this`.
+  if (assignOp->isMemcpyEquivalentSpecialMember(getContext())) {
+    CanQualType recordTy =
+        getContext().getCanonicalTagType(assignOp->getParent());
+    LValue dest = makeNaturalAlignAddrLValue(loadCXXThis(), recordTy);
+    mlir::Value srcPtr = builder.createLoad(getLoc(assignOp->getLocation()),
+                                            getAddrOfLocalVar(args.back()));
+    LValue src = makeNaturalAlignAddrLValue(srcPtr, recordTy);
+    emitAggregateAssign(dest, src, recordTy);
```
erichkeane: Hmm... this really seems wrong to me.  I think we should just leave the copies and count on some level of optimization to figure out "this is all the things, just optimize them all".

CIR should better reflect the AST/semantics.

adams381: This block is really doing two separate jobs, and only one of them needs the frontend.

For a trivially-copyable struct the AST already carries the work: per-field assignment expressions, plus a `__builtin_memcpy` call for array members. Dropping the whole-object copy and just emitting the body (which the non-memcpy path further down already does) is exactly the "reflect the AST and let optimization coalesce" behavior. The `assert(!MissingFeatures::assignMemcpyizer())` there is only the coalescing step; the per-field stores stand on their own and survive `-O3`. That case needs no frontend change.

The union is the one that can't go through that path. Its body is empty -- just `return *this` -- because Sema skips union fields at the FIXME in `SemaDeclCXX.cpp:15474` (copy) and `:15863` (move). There are no copies to leave, and a memcpyizer has nothing to coalesce on an empty body. That empty body is the original regression here: `*++yyvsp = yylval` in `kc`, where `YYSTYPE` is a union, whose no-op `operator=` LLVM deletes at `-O3`. Only a real AST node or this synthesis fills it.

So two ways forward, and I'd like your and @andykaylor's call before touching shared code (same scope question I raised on the other thread Jun 16):

1. Implement the FIXME -- form an AST node for the implied union memcpy in copy and move assignment. That fixes classic and CIR together and lets this PR drop the synthesis. It is a shared-AST change: it shifts classic union copy/move codegen, likely reaches constexpr handling, and updates existing CodeGen tests, and the FIXME has been open a long time, so I don't want to start it without a maintainer nod.
2. If that's more than this PR should carry, narrow it to the struct/array lowering and `errorNYI` the union body, then handle the union in its own PR once the AST question is settled.

Which do you prefer?

erichkeane: I think we should do both.  Lets leave Unions NYI here, and open a separate PR to fix the union case in the AST.  I think there is value to fixing this 'fixme', but it needs to be done with a much wider audience than this patch.

adams381: Dropped the whole-object synthesis. Struct and array assignments already carry the per-field copies in their defaulted bodies (a builtin memcpy for arrays), so those go through the normal body emission unchanged -- only the union needed handling, and it's NYI here. I'll do the union AST fix in a separate PR.

adams381:
@erichkeane @andykaylor  This now does the "do both" from the last round: the union case is `errorNYI` here, and struct/array assignment drops the whole-object synthesis and goes through normal body emission.  The union AST FIXME (`SemaDeclCXX`) will be its own PR.

andykaylor (approved):
lgtm

erichkeane (approved):

PR #198919, opened 2026-05-20
[WebKit Checkers] Treat a temporary which gets passed as a r-value reference as trivial

rniwa:
Allow a temporary object to be created which is passed to a function as a r-value reference since such a function argument is going to consume/absorb the argument.

### DIFF
--- a/clang/lib/StaticAnalyzer/Checkers/WebKit/PtrTypesSemantics.cpp
+++ b/clang/lib/StaticAnalyzer/Checkers/WebKit/PtrTypesSemantics.cpp
@@ -888,17 +888,59 @@ public:
   }
 
   bool checkArguments(const CallExpr *CE) {
+    auto *Callee = CE->getDirectCallee();
+    unsigned ArgOffset =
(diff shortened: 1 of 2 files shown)


clang/lib/StaticAnalyzer/Checkers/WebKit/PtrTypesSemantics.cpp
```diff
@@ -888,17 +888,59 @@ class TrivialFunctionAnalysisVisitor
   }
 
   bool checkArguments(const CallExpr *CE) {
+    auto *Callee = CE->getDirectCallee();
+    unsigned ArgOffset =
+        isa<CXXOperatorCallExpr>(CE) && isa_and_nonnull<CXXMethodDecl>(Callee);
+    unsigned ArgIndex = 0;
```
steakhal, line 894: Have you considered using `llvm::enumerate` and deriving this from the `llvm::enumerate(CE->arguments())`?

rniwa:
Not pursuing this approach anymore.

PR #198921, opened 2026-05-20
[CIR] Implement getUndefRValue for scalar, complex, and aggregate

adams381:
NYI builtin and call paths in CIR were calling getUndefRValue but the
helper only reported "unsupported type for undef rvalue" and returned
null.  libcxx then hit follow-on failures (including scalar-conversion
crashes) on the same translation units that already had other NYI
diagnostics.

Implement the same three cases as classic CodeGen: #cir.undef for
scalar and complex, and an undef.agg.tmp stack slot for aggregates.
Lower cir.const #cir.undef to llvm.undef in DirectToLLVM so emit-llvm
does not assert in the constant op lowering.  builtin-undef-rvalue.cpp
covers the __builtin_reduce_or NYI stub path with CIR/LLVM/OGCG checks.

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenExpr.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExpr.cpp
@@ -2244,8 +2244,30 @@ RValue CIRGenFunction::getUndefRValue(QualType ty) {
   if (ty->isVoidType())
     return RValue::get(nullptr);
 
-  cgm.errorNYI("unsupported type for undef rvalue");
-  return RValue::get(nullptr);
+  mlir::Location loc = builder.getUnknownLoc();
+
+  switch (getEvaluationKind(ty)) {
+  case cir::TEK_Complex: {
+    mlir::Type cirTy = convertType(ty);
+    mlir::Value v = builder.getConstant(loc, cir::UndefAttr::get(cirTy));
+    return RValue::getComplex(v);
+  }
+
+  // If this is a use of an undefined aggregate type, the aggregate must have
+  // an identifiable address.  Just because the contents of the value are
+  // undefined doesn't mean that the address can't be taken and compared.
+  case cir::TEK_Aggregate: {
+    Address destPtr = createMemTemp(ty, loc, "undef.agg.tmp");
+    return RValue::getAggregate(destPtr);
+  }
+
+  case cir::TEK_Scalar: {
+    mlir::Type cirTy = convertType(ty);
+    mlir::Value v = builder.getConstant(loc, cir::UndefAttr::get(cirTy));
+    return RValue::get(v);
+  }
+  }
+  llvm_unreachable("bad evaluation kind");
 }
 
 RValue CIRGenFunction::emitCall(clang::QualType calleeTy,
--- a/clang/lib/CIR/CodeGen/CIRGenExpr.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenExpr.cpp
@@ -2240,6 +2240,11 @@ CIRGenCallee CIRGenFunction::emitDirectCallee(const GlobalDecl &gd) {
   return CIRGenCallee::forDirect(callee, gd);
 }
 
+mlir::Value CIRGenFunction::getUndefConstant(mlir::Location loc,
+                                             mlir::Type cirTy) {
+  return builder.getConstant(loc, cir::UndefAttr::get(cirTy));
+}
+
 RValue CIRGenFunction::getUndefRValue(QualType ty) {
   if (ty->isVoidType())
     return RValue::get(nullptr);
@@ -2248,8 +2253,10 @@ RValue CIRGenFunction::getUndefRValue(QualType ty) {
 
   switch (getEvaluationKind(ty)) {
   case cir::TEK_Complex: {
-    mlir::Type cirTy = convertType(ty);
-    mlir::Value v = builder.getConstant(loc, cir::UndefAttr::get(cirTy));
+    QualType elemTy = ty->castAs<ComplexType>()->getElementType();
+    mlir::Type elemCirTy = convertType(elemTy);
+    mlir::Value undefElem = getUndefConstant(loc, elemCirTy);
+    mlir::Value v = builder.createComplexCreate(loc, undefElem, undefElem);
     return RValue::getComplex(v);
   }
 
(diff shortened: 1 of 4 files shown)


clang/test/CIR/CodeGenBuiltins/builtin-undef-rvalue.cpp
```diff
@@ -0,0 +1,25 @@
...
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-llvm -verify %s -o - \
+// RUN:     | FileCheck %s --check-prefix=LLVM
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -emit-llvm %s -o %t.ll
+// RUN: FileCheck --check-prefix=OGCG --input-file=%t.ll %s
+
+typedef int v4si __attribute__((vector_size(16)));
+
+int test_builtin_reduce_or_undef_rvalue(v4si x) {
+  // expected-error@+1 {{unimplemented X86 builtin call: __builtin_reduce_or}}
+  return __builtin_reduce_or(x);
```
erichkeane, line 10: This test seems unfortunately pretty fragile.  We're going to be implementing this eventually presumably, right?

clang/test/CIR/CodeGenBuiltins/builtin-undef-rvalue.cpp
```diff
@@ -0,0 +1,25 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-cir -verify %s -o - \
+// RUN:     | FileCheck %s --check-prefix=CIR
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -fclangir -emit-llvm -verify %s -o - \
+// RUN:     | FileCheck %s --check-prefix=LLVM
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -emit-llvm %s -o %t.ll
+// RUN: FileCheck --check-prefix=OGCG --input-file=%t.ll %s
```
erichkeane:
```suggestion
// RUN: FileCheck --check-prefix=LLVM --input-file=%t.ll %s
```

clang/lib/CIR/CodeGen/CIRGenExpr.cpp
```diff
@@ -2244,8 +2244,30 @@ RValue CIRGenFunction::getUndefRValue(QualType ty) {
   if (ty->isVoidType())
     return RValue::get(nullptr);
 
-  cgm.errorNYI("unsupported type for undef rvalue");
-  return RValue::get(nullptr);
+  mlir::Location loc = builder.getUnknownLoc();
+
+  switch (getEvaluationKind(ty)) {
+  case cir::TEK_Complex: {
+    mlir::Type cirTy = convertType(ty);
```
erichkeane: classic doesn't do this, it gets the element type.  Why the difference?

clang/lib/CIR/CodeGen/CIRGenExpr.cpp
```diff
@@ -2244,8 +2244,30 @@ RValue CIRGenFunction::getUndefRValue(QualType ty) {
...
+    mlir::Value v = builder.getConstant(loc, cir::UndefAttr::get(cirTy));
+    return RValue::getComplex(v);
+  }
+
+  // If this is a use of an undefined aggregate type, the aggregate must have
+  // an identifiable address.  Just because the contents of the value are
+  // undefined doesn't mean that the address can't be taken and compared.
+  case cir::TEK_Aggregate: {
+    Address destPtr = createMemTemp(ty, loc, "undef.agg.tmp");
```
erichkeane: classic does the 'without-cast' version of ths, why the difference?

clang/lib/CIR/CodeGen/CIRGenExpr.cpp
```diff
@@ -2244,8 +2244,30 @@ RValue CIRGenFunction::getUndefRValue(QualType ty) {
...
+    return RValue::getAggregate(destPtr);
+  }
+
+  case cir::TEK_Scalar: {
+    mlir::Type cirTy = convertType(ty);
+    mlir::Value v = builder.getConstant(loc, cir::UndefAttr::get(cirTy));
```
erichkeane: We should probably just have a 'getUndefConstant' function somewhere to handle this.

adams381: > We should probably just have a 'getUndefConstant' function somewhere to handle this.

Added `CIRGenFunction::getUndefConstant`; scalar and complex paths go through it.

clang/test/CIR/CodeGenBuiltins/builtin-undef-rvalue.cpp
adams381: > This test seems unfortunately pretty fragile. We're going to be implementing this eventually presumably, right?

Yes — when `__builtin_reduce_or` is implemented we will update this probe.  For now it is the smallest path that hits `getUndefRValue` on an NYI X86 builtin with stable CIR/LLVM/OGCG checks.

clang/test/CIR/CodeGenBuiltins/builtin-undef-rvalue.cpp
adams381: > ```suggestion
> // RUN: FileCheck --check-prefix=LLVM --input-file=%t.ll %s
> ```

Switched CIR and LLVM to `-o - > %t.{cir,ll}` plus `FileCheck --input-file=…`.  Clang does not write `-o` paths when `-verify` is active.

clang/lib/CIR/CodeGen/CIRGenExpr.cpp
adams381: > classic does the 'without-cast' version of ths, why the difference?

Added `createMemTempWithoutCast()` and use it for aggregate undef, same role as classic `CreateMemTempWithoutCast`.

clang/lib/CIR/CodeGen/CIRGenExpr.cpp
adams381: > classic doesn't do this, it gets the element type.  Why the difference?

Matched classic: undef on `ComplexType::getElementType()`, assembled with `createComplexCreate`.

clang/test/CIR/CodeGenBuiltins/builtin-undef-rvalue.cpp
erichkeane: I'm not sure what you're trying to say there?  Can you clarify?  All I asked was to stop doing the OGCG checks separately, but they are still there.

erichkeane (commented):
Test OGCG/etc stuff not updated.  otherwise this should be fine.

clang/test/CIR/CodeGenBuiltins/builtin-undef-rvalue.cpp
adams381: Dropped the separate OGCG RUN/CHECK block — this test only pins CIR undef under -verify; classic codegen still lowers the builtin.

erichkeane (approved):

PR #198923, opened 2026-05-20
[lldb] Make TypeSystem::m_sym_file atomic to fix data race

JDevlieghere:
SymbolFileCommon::GetTypeSystemForLanguage unconditionally writes this pointer with `ts->SetSymbolFile(this)` on every lookup, which races with concurrent reads from other threads.

The race is benign in practice: there is exactly one SymbolFile per Module, so every writer stores the same pointer, but it is still undefined behavior under the C++ memory model.

Make the field std::atomic<SymbolFile *> and turn SetSymbolFile into a compare-exchange that asserts a TypeSystem is never rebound to a different SymbolFile, documenting the invariant that lets us get away with this.

The alternative is to have the SymbolFile pointer passed in through the constructor, but that would require updating a bunch of call sites, including various plugin interfaces.

### DIFF
--- a/lldb/include/lldb/Symbol/TypeSystem.h
+++ b/lldb/include/lldb/Symbol/TypeSystem.h
@@ -566,10 +566,10 @@ public:
   /// back-pointer when it first hands out the TypeSystem so that lazy
   /// type-completion callbacks can route through it.
   ///
-  /// Each Module owns at most one SymbolFile and one TypeSystem per
+  /// Each Module owns at most one SymbolFile and one (Module-)TypeSystem per
   /// language, so the binding is expected to be established exactly once.
-  /// Repeated calls with the same pointer are no-ops. Rebinding to a
-  /// different non-null SymbolFile is a programming error.
+  /// Repeated calls with the same pointer are no-ops. Rebinding to a different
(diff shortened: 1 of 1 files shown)


lldb/include/lldb/Symbol/TypeSystem.h
```diff
@@ -559,13 +557,37 @@ class TypeSystem : public PluginInterface,
...
+  SymbolFile *GetSymbolFile() const { return m_sym_file.load(); }
+
+  /// Bind this TypeSystem to its owning SymbolFile.
+  ///
+  /// A TypeSystem is created by its Module's TypeSystemMap before that
+  /// Module has an associated SymbolFile. The SymbolFile installs this
+  /// back-pointer when it first hands out the TypeSystem so that lazy
+  /// type-completion callbacks can route through it.
+  ///
+  /// Each Module owns at most one SymbolFile and one TypeSystem per
```
adrian-prantl: Not necessary, but we could call this
```suggestion
  /// Each Module owns at most one SymbolFile and one (Module-)TypeSystem per
```
to avoid confusion with other TypeSystems like TypeSystemClangForExpressions.

adrian-prantl (approved):

PR #198924, opened 2026-05-20
[CIR] Fix Cmp-Xchg generation when casting to int

erichkeane:
We have logic in CIRGenAtomic to handle 'cast float to an int', which works for most of our atomics.  However, compare-exchange uses a second 'val' value, which was not getting cast, which caused a verification error.

This patch ensures that the correct cast is generated for 'val2' as well.


andykaylor (approved):
lgtm

PR #198925, opened 2026-05-20
[Github] Do not restrict branches for CI workflows

boomanaiden154:
This is covered in our CI best practices document in https://llvm.org/docs/CIBestPractices.html#ensuring-workflows-run-on-the-correct-events.

Otherwise we cannot run libc CI workflows on stacked pull requests.

### DIFF
--- a/.github/workflows/libc-freebsd-vm-tests.yml
+++ b/.github/workflows/libc-freebsd-vm-tests.yml
@@ -4,7 +4,6 @@ permissions:
   contents: read
 on:
   pull_request:
-    branches: [ "main" ]
     paths:
       - 'libc/**'
       - '.github/workflows/libc-freebsd-vm-tests.yml'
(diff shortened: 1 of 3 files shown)


michaelrj-google (approved):

SchrodingerZhu:
Notice this change will make global sccache files accessible in PR, which is a good thing though.

boomanaiden154:
> Notice this change will make global sccache files accessible in PR, which is a good thing though.

This change has no impact on sccache access?

SchrodingerZhu:
for each PR, GHA storage is accessible:

- if it is uploaded by the PR itself in previous runs;

- or it is uploaded on the base branch action.


Previously we don't have the second part.

SchrodingerZhu:
Ooops, I misread the change. The CI still runs in PR only. Then no behavior change is expected.

PR #198927, opened 2026-05-20
[SSAF] Let UnsafeBufferUsageExtractor & PointerFlowExtractor ignore templates

ziqingluo-90:
Templates are ignored for two reasons:

- Template instantiations are still handled. Template facts can be inferred from their instantiations.
- Templates are inherently difficult to reason about. Their ASTs can contain dependent expression types (such as ParenListExpr) that complicate analysis.


ziqingluo-90:
cc: @dtarditi

clang/lib/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlowExtractor.cpp
```diff
@@ -329,6 +329,12 @@ class PointerFlowTUSummaryExtractor : public TUSummaryExtractor {
 
     findContributors(Ctx, Contributors);
     for (auto *CD : Contributors) {
+      // Templates are skipped, but their instantiations are handled. The idea
+      // is that we can conclude facts about a template through all of its
+      // instantiations.
+      if (CD->isTemplated() || CD->getDeclContext()->isDependentContext())
```
steakhal: When I looked at the definition of `isTemplated`, the first thing it checked if it's a dependent decl context.
This suggests to me that the RHS part: `CD->getDeclContext()->isDependentContext()` is redundant.
Why did you add that?

Same question goes to the other hunk.

ojhunt (dismissed):
It would be nice if we just skipped dependent expressions rather than disregarding the entire thing, and I'd also like to include a test to ensure that we don't skip diagnostics when a function in a dependent context is not actually dependent, something like:

```cpp
template <class T> struct Foo {
void f(int *a) {
  
}
};
Foo<int>::f;
```

and the other tests being added should also verify instantiation produces diagnostics.

clang/lib/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlowExtractor.cpp
ziqingluo-90: I read the document of `isTemplated` and somehow interpreted _"Determine whether this declaration is a templated entity (whether it is within the scope of a template parameter)"_ as involving a template parameter.  So I decided that it needs to also cover the case that a Decl is inside a template but not involving a template parameter.

Anyway, you are right. The second part is not needed.

ziqingluo-90:
> It would be nice if we just skipped dependent expressions rather than disregarding the entire thing, and I'd also like to include a test to ensure that we don't skip diagnostics when a function in a dependent context is not actually dependent, 
> and the other tests being added should also verify instantiation produces diagnostics.

@ojhunt As background, these `Extractor` classes are not for a specific diagnostic or analysis. Their duty is to collect facts about specific aspects of the source code. Those facts may be used by various clients, e.g., analyzers or source-code editing tools. In addition, this is a part of the SSAF framework, which analyzes the entire project(s) as its "whole world".

Under this assumption, my opinion is that we do not reason about properties of the template itself but of all of its instantiations. A rough analogy is reasoning about `Forall T. P(T)` vs. `Forall T in {a, b, ...}. P(T)`, where `{a, b, ...}` is the finite set of concrete types with which `T` is instantiated.  Since the instantiations are collected from the "whole world", the latter is sufficient and simple to reason about.

WDYT?

clang/unittests/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlowTest.cpp
```diff
@@ -1126,4 +1126,37 @@ TEST_F(PointerFlowTest, NestedLambdaAssign) {
...
+//////////////////////////////////////////////////////////////
+TEST_F(PointerFlowTest, FunctionTemplate) {
+  ASSERT_EQ(setUpTest(R"cpp(
+    template <typename T>
+    T* f(T *p) {
+      int *q = p
+      return q;
+    }
+  )cpp"),
+            true);
```
steakhal: What is this parameter that is always set to `true`? Maybe that should be the default?

clang/unittests/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlowTest.cpp
```diff
@@ -1126,4 +1126,37 @@ TEST_F(PointerFlowTest, NestedLambdaAssign) {
...
+    struct Wrapper {
+      T *ptr;
+      void set(T *p) { ptr = p; }
+    };
+  )cpp"),
+            true);
+
+  auto *Sum = getEntitySummary<FunctionDecl>("set");
+
+  ASSERT_EQ(Sum, nullptr);
```
steakhal:
```suggestion
  ASSERT_FALSE(getEntitySummary<FunctionDecl>("set"));
```

clang/unittests/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlowTest.cpp
```diff
@@ -1126,4 +1126,37 @@ TEST_F(PointerFlowTest, NestedLambdaAssign) {
...
+    T* f(T *p) {
+      int *q = p
+      return q;
+    }
+  )cpp"),
+            true);
+
+  auto *Sum = getEntitySummary<FunctionDecl>("f");
+
+  ASSERT_EQ(Sum, nullptr);
```
steakhal: Let's keep this lean.

```suggestion
  ASSERT_FALSE(getEntitySummary<FunctionDecl>("f"));
```

clang/unittests/ScalableStaticAnalysisFramework/Analyses/UnsafeBufferUsage/UnsafeBufferUsageTest.cpp
```diff
@@ -570,4 +570,36 @@ TEST_F(UnsafeBufferUsageTest, NestedDefinitions2) {
...
+    template <typename T>
+    T* f(T *p) {
+      return &p[5];
+    }
+  )cpp"),
+            true);
+
+  auto *Sum = getEntitySummary<FunctionDecl>("f");
+
+  ASSERT_EQ(Sum, nullptr);
```
steakhal:
```suggestion
  ASSERT_FALSE(getEntitySummary<FunctionDecl>("f"));
```

clang/unittests/ScalableStaticAnalysisFramework/Analyses/UnsafeBufferUsage/UnsafeBufferUsageTest.cpp
```diff
@@ -570,4 +570,36 @@ TEST_F(UnsafeBufferUsageTest, NestedDefinitions2) {
...
+    struct Wrapper {
+      T *ptr;
+      void set(T *p) { ptr = p[5]; }
+    };
+  )cpp"),
+            true);
+
+  auto *Sum = getEntitySummary<FunctionDecl>("set");
+
+  ASSERT_EQ(Sum, nullptr);
```
steakhal:
```suggestion
  ASSERT_FALSE(getEntitySummary<FunctionDecl>("set"));
```

steakhal (commented):
This patch looks good to me.

IMO we should not rewrite unused templates - because they are unused anyway, so why bother.

Template instantiation will be represented in the current implementation, and assuming complete whole-program analyses, all users (instantiations) of the templates will be accounted for. Consequently, I have no correctness concerns.

I only had minor suggestions.

ziqingluo-90:
> Template instantiation will be represented in the current implementation, and assuming complete whole-program analyses, all users (instantiations) of the templates will be accounted for. 

You've explained the idea better than me!

clang/lib/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlowExtractor.cpp
steakhal: This wasnt addressed. Similarly in the other case.

steakhal (changes requested):
See the (probably overlooked) unresolved discussion:
https://github.com/llvm/llvm-project/pull/198927#discussion_r3376733893

ziqingluo-90:
> we do not reason about properties of the template itself but of all of its instantiations

With that said, I just realized that instantiations were not visited 😅
Fixed this in https://github.com/llvm/llvm-project/pull/198927/changes/236b9225d29f148c163fe25ade4feb4e7ca65e8d.

@steakhal sorry that I have to ask you to review this PR one more time.

clang/unittests/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlowTest.cpp
```diff
@@ -1266,6 +1270,72 @@ TEST_F(PointerFlowTest, CXXConstructExprArrayInit) {
...
+    }
+  )cpp"));
+
+  auto *Sum = getEntitySummary<FunctionDecl>("set");
+
+  ASSERT_TRUE(Sum);
+  ASSERT_EQ(*Sum, makeEdges(__LINE__, {{{"ptr", 1U}, {"p", 1U}}}));
+}
+
+//////////////////////////////////////////////////////////////
```
ziqingluo-90:
```suggestion
//////////////////////////////////////////////////////////////
```

clang/lib/ScalableStaticAnalysisFramework/Analyses/SSAFAnalysesCommon.cpp
```diff
@@ -34,6 +34,11 @@ class ContributorFinder : public DynamicRecursiveASTVisitor {
 public:
   std::set<const NamedDecl *> Contributors;
 
+  ContributorFinder() {
+    ShouldVisitTemplateInstantiations = true;
+    ShouldVisitImplicitCode = false;
```
steakhal, line 39: My personal stance is that if we took the time to evaluate what the default visitation options do then it's probably worths spelling them out explicitly here and not rely on the defaults.
However, this is just my personal preference and I don't see much risk of relying on the defaults.

EDIT: Same comment applies to `NamedDeclFinder`.

You can decide to act on this comment or disregard it if you wish.

clang/lib/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlowExtractor.cpp
```diff
@@ -336,6 +336,12 @@ class PointerFlowTUSummaryExtractor : public TUSummaryExtractor {
 
     findContributors(Ctx, Contributors);
     for (auto *CD : Contributors) {
+      // Templates are skipped, but their instantiations are handled. The idea
+      // is that we can conclude facts about a template through all of its
+      // instantiations.
+      if (CD->isTemplated())
+        continue;
```
steakhal, line 343: I've noticed that you have this comment for these two occurrences of `isTemplated` while in FindDecl you don't. This makes me think about why?

steakhal (approved):
The change looks correct to me.
I had some subjective opinions inline, but I don't have hard feelings of any of those.
I think we are good to go. Feel free to address the ones you agree with and merge it.

clang/lib/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlowExtractor.cpp
ziqingluo-90: The real reason is that I have a lower (or different) documentation standard for testing code than for functional code 😅. But I also don't feel like copying and pasting the lengthy explanation in multiple places. Ideally, we should factor out the common part and document it in one place.

clang/lib/ScalableStaticAnalysisFramework/Analyses/SSAFAnalysesCommon.cpp
ziqingluo-90: By 'spelling out explicitly', do you mean we should list and overwrite all the default options (and add some comments)? I guess another way to 'spell it out' would be in the name of the visitor itself, like 'InstantiationIncludedVisitor'.

steakhal: I see. Yes, listing these exhaustively was what i had in mind. Sorry for the confusion.

clang/lib/ScalableStaticAnalysisFramework/Analyses/PointerFlow/PointerFlowExtractor.cpp
steakhal: Got it. Consider this resolved.

clang/lib/ScalableStaticAnalysisFramework/Analyses/SSAFAnalysesCommon.cpp
ziqingluo-90: I will address this in a follow up patch.

ziqingluo-90:
@ojhunt 
Since this PR is currently blocking others from landing, I am going to go ahead and merge it for now.
Regarding the discussion https://github.com/llvm/llvm-project/pull/198927#pullrequestreview-4369558324,  we can address it in follow-up PRs if needed.

PR #198929, opened 2026-05-20
[llvm-readobj] Print all callgraph sections

Prabhuk:
Ensure that `llvm-readobj` / `llvm-readelf` processes and dumps all 
SHT_LLVM_CALL_GRAPH sections present in an object file, rather
than just the first one.

**Motivation:**
Consider this `foo.s` file
```
.pushsection .text, "axG", %progbits, foo
  .globl foo
  foo:
  ret
  .pushsection .llvm.callgraph, "?", %llvm_call_graph
    .byte 0, 0
    .dc.a foo
    .quad 0
  .popsection
.popsection

.pushsection .text, "axG", %progbits, bar
  .globl foo
  bar:
  ret
  .pushsection .llvm.callgraph, "?", %llvm_call_graph
    .byte 0, 0
    .dc.a bar
    .quad 0
  .popsection
.popsection
```
`clang -c foo.s && llvm-readelf --elf-output-style=JSON --pretty-print --call-graph-info foo.o` on that shows that there is only one record reported because the current implementation just looked at the first SHT_LLVM_CALL_GRAPH section and then ignored all others. This patch will address this gap by iterating over all the SHT_LLVM_CALL_GRAPH sections.


llvm/test/tools/llvm-readobj/ELF/call-graph-info-warn-malformed.test
```diff
@@ -271,7 +271,7 @@ Symbols:
   - Name:  baz
 ...
 
-## Check that the first call graph section is processed and the second call graph section is ignored.
+## Check that both the call graph sections are processed and printed.
 ## Note that we do not expect an object file to have more than one call graph section as we merge the
 ## call graph sections when there are more than one.
```
jh7370: Is this still relevant? If not, does this test case even belong in the "malformed" file any more?

llvm/test/tools/llvm-readobj/ELF/call-graph-info.test
```diff
@@ -574,3 +574,110 @@ Symbols:
     Type:     STT_FUNC
     Section:  .text
     Value:    0x1000
+
+## Check handling of multiple call graph sections in a relocatable object file.
+# RUN: yaml2obj --docnum=5 %s -o %t5
+# RUN: llvm-readobj --elf-output-style=LLVM --call-graph-info %t5 2>&1 | \
```
jh7370, line 583: GNU output too?

llvm/tools/llvm-readobj/ELFDumper.cpp
```diff
@@ -8338,101 +8338,108 @@ template <class ELFT> void LLVMELFDumper<ELFT>::printCallGraphInfo() {
...
-      Expected<const typename ELFT::Shdr *> SymtabOrErr =
-          this->Obj.getSection(CGRelSection->sh_link);
-      if (!SymtabOrErr) {
-        reportWarning(createError("invalid section linked to " +
-                                  this->describe(*CGRelSection) + ": " +
-                                  toString(SymtabOrErr.takeError())),
-                      this->FileName);
-        return;
+    FuncCGInfos.clear();
+    if (!this->processCallGraphSection(CGSection, FuncCGInfos) ||
```
jh7370: How about changing `processCallGraphSection` to return the `FuncCGInfos` instead? That would remove the need for a parameter to be passed in purely for output purposes and simplify this if (you'd just check the `FuncCGInfos.empty()` call in the if, after calling the function.

llvm/tools/llvm-readobj/ELFDumper.cpp
```diff
@@ -8338,101 +8338,108 @@ template <class ELFT> void LLVMELFDumper<ELFT>::printCallGraphInfo() {
     return;
   }
 
-  // Process and print the first SHT_LLVM_CALL_GRAPH type section found.
-  if (!this->processCallGraphSection(MapOrErr->begin()->first) ||
-      this->FuncCGInfos.empty())
-    return;
+  std::unique_ptr<ListScope> CGI;
+  SmallVector<FunctionCallgraphInfo, 16> FuncCGInfos;
```
jh7370: Can this be declared inside the loop? It's cleared every loop iteration, so there doesn't seem to be much point in it being declared outside the loop.

jh7370 (commented):
It would be helpful to have the description updated with the motivation, since this is a change from what was originally implemented.

llvm/test/tools/llvm-readobj/ELF/call-graph-info.test
Prabhuk: Thank you. Added the GNU output check (no output) to the tests which were missing this check.

llvm/test/tools/llvm-readobj/ELF/call-graph-info-warn-malformed.test
Prabhuk: It checks the ET_DYN case with multiple call graph sections. I moved it out of malformed file and into the file which tests valid cases. PTAL.

llvm/tools/llvm-readobj/ELFDumper.cpp
Prabhuk: I changed processCallGraphSection to return FuncCGInfos. Returns empty if there was a failure during parsing or if the list is actually empty.

llvm/tools/llvm-readobj/ELFDumper.cpp
Prabhuk: Done. PTAL.

llvm/tools/llvm-readobj/ELFDumper.cpp
jh7370: Couple of suggested tweaks to this comment:
"Read the SHT_LLVM_CALL_GRAPH type sections and process their contents to populate call graph related data structures which will be used to dump call graph info. Returns an empty vector if there are no such sections or if parsing fails."

llvm/test/tools/llvm-readobj/ELF/call-graph-info.test
jh7370: This one is still missing the GNU output check.

jh7370 (commented):
Mostly looks good, but missing one small thing.

llvm/tools/llvm-readobj/ELFDumper.cpp
Prabhuk: Thank you. Done. PTAL.

jh7370 (approved):
LGTM!

PR #198931, opened 2026-05-20
[LLDB] Add a progress event to xcrun invocations

adrian-prantl:
LLDB invokes xcrun to find SDKs on disk. This is usually very fast, but sometimes (after an Xcode update, or when the searched SDK does not exist) it can take very long (10s or more). The progress event provides user feedback to explain the hang.

### DIFF
--- a/lldb/source/Host/macosx/objcxx/HostInfoMacOSX.mm
+++ b/lldb/source/Host/macosx/objcxx/HostInfoMacOSX.mm
@@ -7,6 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "lldb/Host/macosx/HostInfoMacOSX.h"
+#include "lldb/Core/Progress.h"
 #include "lldb/Host/FileSystem.h"
 #include "lldb/Host/Host.h"
 #include "lldb/Host/HostInfo.h"
@@ -24,6 +25,7 @@
 #include "llvm/ADT/ScopeExit.h"
 #include "llvm/ADT/SmallString.h"
 #include "llvm/ADT/StringMap.h"
+#include "llvm/ADT/StringExtras.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/RWMutex.h"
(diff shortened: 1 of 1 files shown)


adrian-prantl:
Unfortunately I didn't find a _great_ way to test this, since this is very specific to macOS and the local Xcode installation.

jasonmolenda:
Ah good idea.

lldb/source/Host/macosx/objcxx/HostInfoMacOSX.mm
```diff
@@ -433,6 +434,13 @@ static bool ResolveAndVerifyCandidateSupportDir(FileSpec &path) {
 static llvm::Expected<std::string>
 xcrun(const std::string &sdk, llvm::ArrayRef<llvm::StringRef> arguments,
       llvm::StringRef developer_dir = "") {
+  std::string details;
+  llvm::raw_string_ostream os(details);
+  os << "--sdk " << sdk;
+  for (auto arg : arguments)
+    os << ' ' << arg;
```
JDevlieghere: [llvm::join](https://llvm.org/doxygen/namespacellvm.html#aa1fd1719afe5b07ebe55f1d105c98916)

lldb/source/Host/macosx/objcxx/HostInfoMacOSX.mm
```diff
@@ -433,6 +434,13 @@ static bool ResolveAndVerifyCandidateSupportDir(FileSpec &path) {
 static llvm::Expected<std::string>
 xcrun(const std::string &sdk, llvm::ArrayRef<llvm::StringRef> arguments,
       llvm::StringRef developer_dir = "") {
+  std::string details;
+  llvm::raw_string_ostream os(details);
+  os << "--sdk " << sdk;
```
JDevlieghere: We'll probably want to repeat `xcrun` here again.
```suggestion
  os << "xcrun --sdk " << sdk;
```

lldb/source/Host/macosx/objcxx/HostInfoMacOSX.mm
```diff
@@ -433,6 +434,13 @@ static bool ResolveAndVerifyCandidateSupportDir(FileSpec &path) {
...
+  for (auto arg : arguments)
+    os << ' ' << arg;
+  Progress progress("Running xcrun", details);
```
JDevlieghere, line 442: Maybe make this more generic:
```suggestion
  Progress progress("Looking for Xcode SDKs", details);
```
Then that will show up as:

```
Looking for Xcode SDKs: xcrun --sdk MacOSX 
```

adrian-prantl: This is not just used to find Xcode SDKs. The Swift plugin uses it to find tools in the SDK, too (such as swift-plugin-server).

adrian-prantl: It uses `HostInfoMacOSX::FindSDKTool`, which exists on main but is only unit-tested.

JDevlieghere: Ack, then let's keep "Running xcrun". I'd still duplicate it in the details, as different clients display this differently and it's not guaranteed to show up after the title (e.g. VS Code)

JDevlieghere (approved):
Nice

jasonmolenda:
reverted via https://github.com/llvm/llvm-project/pull/198945

PR #198933, opened 2026-05-20
[flang][FIRToMemRef] Fixed array_coor with box/shape/slice.

vzakhari:
A `fir.array_coor` with box input, a slice and a shape without
a shift, should also be converted using dimensions information
stored in the input box (case (c) in the updated code).

### DIFF
--- a/flang/test/Transforms/FIRToMemRef/array-coor-slice-shift.mlir
+++ b/flang/test/Transforms/FIRToMemRef/array-coor-slice-shift.mlir
@@ -75,10 +75,8 @@ func.func @array_coor_slice_shift_section() {
   return
 }
 
-// Descriptor-backed array_coor with shape_shift + slice. Even though the
-// memref is a fir.box, the explicit shape_shift carries lower bounds, so the
-// shapeVec path (synthesize row-major strides from extents) must be taken,
-// not the box_dims descriptor path.
+// Descriptor-backed array_coor with shape_shift + slice.
+// TODO: the strides must still be taken from the descriptor.
 // CHECK-LABEL: func.func @array_coor_box_shape_shift_slice
 // CHECK:       %[[C1:.*]] = arith.constant 1 : index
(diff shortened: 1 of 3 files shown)


flang/test/Transforms/FIRToMemRef/array-coor-slice-shift.mlir
```diff
@@ -74,3 +74,30 @@ func.func @array_coor_slice_shift_section() {
   fir.store %c1_i32 to %4 : !fir.ref<i32>
   return
 }
+
+// Descriptor-backed array_coor with shape_shift + slice. Even though the
+// memref is a fir.box, the explicit shape_shift carries lower bounds, so the
+// shapeVec path (synthesize row-major strides from extents) must be taken,
+// not the box_dims descriptor path.
+// CHECK-LABEL: func.func @array_coor_box_shape_shift_slice
```
jeanPerier, line 80: I do not really understand that case. I understand that you need to deal with the lower bounds, and that maybe the path that is using the box strides is not handling that, but I do not get how it is possible to ignore/not use the input box stride too.

Are you adding the `sliceInfo.shiftVec.empty()` condition because in practice such code is always produced with contiguous fir.box and that using the strides for that case would require fixing/modifying more code to deal with the lower bounds when using descriptor strides?

jeanPerier (approved):
Thanks, I have question about the case with shift, but I am OK with merging this as the handling of shift matches the existing behavior, so if there is an issue it is not a regression with this patch.

flang/test/Transforms/FIRToMemRef/array-coor-slice-shift.mlir
vzakhari: Great point! I was too focused on the particular issue at hand, and did not want to pessimize the other cases.

I will add a TODO here and will fix it in a follow-up PR, if that is okay. Thank you!

PR #198936, opened 2026-05-20
[MC][Support] Track macro definition locations for improved diagnostics

arichardson:
Track the exact location of macro definitions and instantiations inside the
SourceManager. By passing the instantiation location and the macro signature
location to a new AddMacroInstantiationBuffer method, the diagnostic handlers
can cleanly map virtual instantiation buffer locations back to their physical
source files. This avoids printing redundant include stacks and ensures macro
instantiation notes point directly to the macro definition and call sites.

Introduce getMacroInstantiationLoc and printIncludeStackForDiagnostic helper
methods to SourceMgr to share this nested unwinding logic between the asm parsers.

Assisted-By: Gemini


llvm/include/llvm/Support/SourceMgr.h
```diff
@@ -81,8 +82,15 @@ class SourceMgr {
...
 
     /// This is the location of the parent include, or null if at the top level.
+    /// For macro instantiation buffers, this is the macro call location.
     SMLoc IncludeLoc;
 
+    /// The parent buffer ID where this macro was expanded, or 0 if not a macro.
+    unsigned MacroParentBuf = 0;
+
+    /// The location in the parent buffer where this macro was defined.
+    SMLoc MacroDefLoc;
```
arichardson: Could reorder these to reduce padding, not sure if it's important

arichardson:
~I just noticed this won't actually work as is due to expansion changing the length of buffers and we can't cleanly map it back. Will update this to be a simpler change that mostly adds more test coverage.~

Done

arichardson:
ping?

arichardson:
@MaskRay Any suggestions who else could review this if you are too busy?

llvm/test/MC/AsmParser/macro-like-diagnostic.s
```diff
@@ -0,0 +1,19 @@
+# RUN: not llvm-mc -triple x86_64 %s -o /dev/null 2>&1 | FileCheck %s
```
MaskRay, line 1: This patch doesn't change the output for this test.

llvm/test/MC/AsmParser/macro-unknown-directive.s
```diff
@@ -1,19 +1,42 @@
...
-# CHECK-NEXT:^
+## This test verifies that:
+## * Macro instantiation notes point directly to the `.macro` signature line itself
+##   (e.g. `.macro .test1` instead of the first statement of the macro body).
+## * Nested active macro instantiation stacks unwind cleanly.
+## * Column numbers and caret alignments inside `<instantiation>` virtual buffers
+##   are reported correctly (e.g. matching `.macrobody0` at column 7 on line 2).
+
+
+#      CHECK:<instantiation>:2:7: error: unknown directive
```
MaskRay, line 11: Is the output ideal? It doesn't mention `.macro .test0`

llvm/include/llvm/Support/SourceMgr.h
```diff
@@ -158,6 +162,23 @@ class SourceMgr {
...
+      return FindBufferContainingLoc(getParentIncludeLoc(i));
+    return 0;
+  }
+
+  SMLoc getMacroDefLoc(unsigned i) const {
+    assert(isValidBufferID(i));
+    return Buffers[i - 1].MacroDefLoc;
+  }
+
+  unsigned getMacroDefBuf(unsigned i) const {
```
MaskRay, line 176: Dead?

MaskRay:
> @MaskRay Any suggestions who else could review this if you are too busy?

I like the direction of the patch but I'm worried about the AI slop.

llvm/test/MC/AsmParser/macro-like-diagnostic.s
arichardson: Yeah I just wanted to check that the macro like directives still work. Can pre commit that.

llvm/include/llvm/Support/SourceMgr.h
arichardson: Was used in a follow up change and I folded it into the wrong commit. Sorry

llvm/test/MC/AsmParser/macro-unknown-directive.s
arichardson: Yeah it's not ideal. How about printing `<.test0 instantiation>? Since the buffers just contain the textual replacement there might not be a good correlation with the actual source code.

arichardson: And maybe the location of the macro definition as the error location? Otherwise it's kinda hard to find where it came from?

MaskRay: I am not clear about the desired output; can you try ask agents to get inspiration from clang macro expansion?

llvm/test/MC/AsmParser/macro-like-diagnostic.s
MaskRay: Yeah - should just pre-commit this

PR #198937, opened 2026-05-20
[RISCV][P-ext] Support packed SSHLSAT with non-constant splat shift amount.

topperc:
We can use the pssha instructions. These instructions look at the lower 8-bits of the shift amount and treat positive values as a left shift and a negative value as a right shift.

Since out of bound shift amounts for SSHLSAT are poison, we can directly pass the shift amount to the instruction.


sihuan (approved):
LGTM.

PR #198940, opened 2026-05-20
Optimization on predecessor count checks for FunctionPropertiesAnalysis

InakiVA:
It used to need to get the whole bb predecessor count and then check with the size of it. That needed to go through the whole list only to check against 1, 2 or 3. By using its already existing basic block predecessor count function, these attributes can be counted much faster. Now it doesn't need to go through the whole list. The moment it gets to the 3rd element, no matter how many predecessors there might be, we have our answer for the metrics


boomanaiden154 (commented):
It's unclear to me how this is an optimization. We're doing the exact same logic but no looking at the distance more often. I'm pretty sure `pred_size` and `hasNPredecessors` both end up being O(n) on the number of predecessors given the predecessor iterator is not random access. Did this come up on the flame graph?

Based on https://github.com/llvm/llvm-project/blob/012abddca170c33425fef4dd9a26e77889a0dc91/llvm/include/llvm/IR/CFG.h#L123, it seems like `hasNPredecessors` is the preferred API, but we should still probably only be calling it once into a temporary. Theoretically the compiler with optimizations turned on should CSE the predecessor count, but the loops involved might make that difficult.

InakiVA:
> It's unclear to me how this is an optimization. We're doing the exact same logic but no looking at the distance more often. I'm pretty sure `pred_size` and `hasNPredecessors` both end up being O(n) on the number of predecessors given the predecessor iterator is not random access. Did this come up on the flame graph?

https://github.com/llvm/llvm-project/blob/2b460b36bdb4ab8f239f5efc38ad83ffde9b1b35/llvm/include/llvm/ADT/STLExtras.h#L2607-L2612


It is not in O(n), it is in O(N), with N referring to the number specified in has**N**Items (same for hasNItemsOrMore). Size is O(n), with n being the amount of predecessors of the basic block which can be much more than the maximum we need, 3

llvm/lib/Analysis/FunctionPropertiesAnalysis.cpp
```diff
@@ -104,12 +104,11 @@ void FunctionPropertiesInfo::updateForBB(const BasicBlock &BB,
     else if (SuccessorCount > 2)
       BasicBlocksWithMoreThanTwoSuccessors += Direction;
 
-    unsigned PredecessorCount = pred_size(&BB);
```
boomanaiden154, line 107: We should still just be updating this to call `BB.hasNPredecessors`

boomanaiden154 (commented):
Ah, missed that part about what the factors actually are.

llvm/lib/Analysis/FunctionPropertiesAnalysis.cpp
InakiVA: not sure if I'm following. What would be the purpose of assigning the boolean value and change it every time instead of just calling it during each check? This doesn't return the value bounded by N, it just returns the boolean of whether it has the exact N or not

boomanaiden154: Yeah, nevermind. For some reason I was still assuming we get the size.

boomanaiden154 (approved):
LGTM

PR #70151, opened 2023-10-25
Always use pthread_rwlock on Apple platforms

cooperp:
pthread_rwlock is faster than std::shared_mutex, especially in heavily threaded code such as the Swift compiler.

rdar://117445844

PR #198941, opened 2026-05-20
[lldb] Add Locked<T> and SharedLocked<T> RAII handles in Utility (NFC)

JDevlieghere:
Introduce two RAII handles that pair a pointer-like value with a lock on a caller-supplied mutex.

  - Locked<PtrT, Mutex>: holds std::unique_lock<Mutex>. Move-only; enforces exclusive access. Defaults to std::recursive_mutex to match LLDB's existing synchronization style.
  - SharedLocked<PtrT, Mutex>: holds a reference-counted std::shared_lock<Mutex>. Copyable — copies share the same reader lock and the lock releases when the last copy goes away. Defaults to llvm::sys::RWMutex, the LLDB convention.

PtrT may be a raw pointer, std::shared_ptr, or std::unique_ptr; convenience aliases LockedPtr/LockedSP/LockedUP and SharedLockedPtr/SharedLockedSP/SharedLockedUP cover the common combinations. SharedLocked's borrowed pointer is const-qualified so readers can't mutate the pointee.

### DIFF
--- a/lldb/include/lldb/Utility/Locked.h
+++ b/lldb/include/lldb/Utility/Locked.h
@@ -1,0 +1,142 @@
+//===-- Locked.h ------------------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLDB_UTILITY_LOCKED_H
+#define LLDB_UTILITY_LOCKED_H
+
+#include "llvm/Support/RWMutex.h"
+
+#include <memory>
+#include <mutex>
+#include <shared_mutex>
+#include <type_traits>
+#include <utility>
+
+namespace lldb_private {
+
+namespace detail {
+/// Common pointer-like accessors shared by `Locked` and `SharedLocked`.
+template <typename Derived, typename PtrT> class LockedAccessors {
+public:
+  auto operator->() const { return Raw(); }
+  decltype(auto) operator*() const { return *Self()->m_ptr; }
+  auto get() const { return Raw(); }
+  explicit operator bool() const { return Raw() != nullptr; }
+
+private:
+  const Derived *Self() const { return static_cast<const Derived *>(this); }
+
+  auto Raw() const {
+    if constexpr (std::is_pointer_v<PtrT>)
+      return Self()->m_ptr;
+    else
+      return Self()->m_ptr.get();
+  }
+};
+} // namespace detail
+
+/// A move-only RAII handle that pairs a pointer-like value with an exclusive
+/// lock on a caller-supplied mutex. While the handle is alive the borrowed
+/// pointer is serialized against other threads that go through the same
+/// mutex.
+///
+/// `PtrT` is the pointer-like value: a raw pointer (`T*`),
+/// `std::shared_ptr<T>`, or `std::unique_ptr<T>`. `Mutex` may be any type
+/// that satisfies `Lockable` — `std::mutex`, `std::recursive_mutex`,
+/// `std::shared_mutex`, or `llvm::sys::RWMutex` all work. Use the
+/// `LockedPtr`, `LockedSP`, `LockedUP` aliases for the common combinations.
+template <typename PtrT, typename Mutex = std::recursive_mutex>
+class Locked : public detail::LockedAccessors<Locked<PtrT, Mutex>, PtrT> {
+  friend class detail::LockedAccessors<Locked<PtrT, Mutex>, PtrT>;
+
+public:
+  using mutex_type = Mutex;
+  using lock_type = std::unique_lock<Mutex>;
+
+  Locked() = default;
+  Locked(mutex_type &m, PtrT p) : m_lock(m), m_ptr(std::move(p)) {}
+  Locked(lock_type lock, PtrT p)
+      : m_lock(std::move(lock)), m_ptr(std::move(p)) {}
+
+  Locked(Locked &&) noexcept = default;
+  Locked &operator=(Locked &&) noexcept = default;
+  Locked(const Locked &) = delete;
+  Locked &operator=(const Locked &) = delete;
+
+private:
+  lock_type m_lock;
+  PtrT m_ptr{};
+};
+
+/// A copyable RAII handle that pairs a pointer-like value with a shared
+/// (reader) lock on a caller-supplied mutex. Copies share the same
+/// underlying reader lock through reference counting; the lock is released
+/// when the last copy is destroyed. This makes a `SharedLocked` cheap to
+/// pass through code paths that branch or fan out without each leaf having
+/// to re-acquire.
+///
+/// The borrowed pointer is `const`-qualified so callers cannot mutate the
+/// pointee while holding only a reader's lock. `Mutex` must satisfy
+/// `SharedLockable` — `llvm::sys::RWMutex` (the LLDB convention) or
+/// `std::shared_mutex`. Use the `SharedLockedPtr`, `SharedLockedSP`,
+/// `SharedLockedUP` aliases for the common combinations.
+template <typename PtrT, typename Mutex = llvm::sys::RWMutex>
+class SharedLocked
+    : public detail::LockedAccessors<SharedLocked<PtrT, Mutex>, PtrT> {
+  friend class detail::LockedAccessors<SharedLocked<PtrT, Mutex>, PtrT>;
+
+public:
+  using mutex_type = Mutex;
+  using lock_type = std::shared_lock<Mutex>;
+
+  SharedLocked() = default;
+  SharedLocked(mutex_type &m, PtrT p)
+      : m_lock(std::make_shared<lock_type>(m)), m_ptr(std::move(p)) {}
+  SharedLocked(lock_type lock, PtrT p)
+      : m_lock(std::make_shared<lock_type>(std::move(lock))),
+        m_ptr(std::move(p)) {}
+
+  SharedLocked(const SharedLocked &) = default;
+  SharedLocked &operator=(const SharedLocked &) = default;
+  SharedLocked(SharedLocked &&) noexcept = default;
+  SharedLocked &operator=(SharedLocked &&) noexcept = default;
+
+private:
+  std::shared_ptr<lock_type> m_lock;
+  PtrT m_ptr{};
+};
+
+/// Exclusive (write) access aliases. The default mutex is
+/// `std::recursive_mutex` to match the existing LLDB synchronization style.
+/// @{
+template <typename T, typename Mutex = std::recursive_mutex>
+using LockedPtr = Locked<T *, Mutex>;
+
+template <typename T, typename Mutex = std::recursive_mutex>
+using LockedSP = Locked<std::shared_ptr<T>, Mutex>;
+
+template <typename T, typename Mutex = std::recursive_mutex>
+using LockedUP = Locked<std::unique_ptr<T>, Mutex>;
+/// @}
+
+/// Shared (read) access aliases. The default mutex is `llvm::sys::RWMutex`,
+/// the LLDB convention for read/write locks.
+/// @{
+template <typename T, typename Mutex = llvm::sys::RWMutex>
+using SharedLockedPtr = SharedLocked<const T *, Mutex>;
+
+template <typename T, typename Mutex = llvm::sys::RWMutex>
+using SharedLockedSP = SharedLocked<std::shared_ptr<const T>, Mutex>;
+
+template <typename T, typename Mutex = llvm::sys::RWMutex>
+using SharedLockedUP = SharedLocked<std::unique_ptr<const T>, Mutex>;
+/// @}
+
+} // namespace lldb_private
+
+#endif // LLDB_UTILITY_LOCKED_H
--- a/lldb/include/lldb/Utility/Locked.h
+++ b/lldb/include/lldb/Utility/Locked.h
@@ -11,6 +11,7 @@
 
 #include "llvm/Support/RWMutex.h"
 
+#include <cassert>
 #include <memory>
 #include <mutex>
 #include <shared_mutex>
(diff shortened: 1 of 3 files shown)


lldb/include/lldb/Utility/Locked.h
```diff
@@ -0,0 +1,43 @@
...
+/// caller-supplied mutex. The lock is held for as long as the handle is alive,
+/// so the borrowed pointer cannot outlive its synchronization scope.
+template <typename T, typename Mutex = std::recursive_mutex> class Locked {
+public:
+  Locked() = default;
+  Locked(Mutex &m, T *p) : m_lock(m), m_ptr(p) {}
+  Locked(std::unique_lock<Mutex> lock, T *p)
+      : m_lock(std::move(lock)), m_ptr(p) {}
+
+  Locked(Locked &&) noexcept = default;
```
augusto2112: Why noexcept? We never build with exceptions right?

augusto2112 (commented):
I like the change. Do we legitimately need access to the non-locked version of those functions too? If not, I think it would make sense to maybe move them to a different type or something like that, so new uses don't accidentally call the non-locked version? I think that would be an easy mistake to make.

lldb/include/lldb/Utility/Locked.h
JDevlieghere: Yeah this is redundant.

JDevlieghere:
> I like the change. Do we legitimately need access to the non-locked version of those functions too? If not, I think it would make sense to maybe move them to a different type or something like that, so new uses don't accidentally call the non-locked version? I think that would be an easy mistake to make.

Agreed. Before I put up  the PR, I did a quick grep and there are *a lot* (hundreds) of call sites. I'd have to audit them one-by-one to make sure it's safe to do the locking (most likely it's the right thing to do). So definitely we should aim to remove them, but I think it'll have to happen incrementally.

jasonmolenda:
I'm not the best at multi-threaded race issues, but all of these objects that we're returning pointers to within the Module use locking within themselves.  The only thing that would be thread unsafe is if a new Symtab/SectionList/ObjectFile could be created for the Module, invalidating the old pointer we handed out?  I'm probably wrong, but I can see why these API were done this way, it SEEMS like it would be OK.

jasonmolenda:
I can see where you describe the issue in your description

> The existing Module::GetObjectFile, GetSymbolFile, GetSymtab, and GetSectionList accessors take the Module mutex internally during lazy initialization but release it before returning a raw pointer. Callers that mutate the returned object run without synchronization afterwards, exposing those classes to data races.

So I have two threads that call GetSymtab() and there hasn't been a Symtab allocated in the Module yet.  Both lock, one wins and allocates a Symtab, releases the mutex and returns a pointer.  Second thread gets the lock and returns the same pointer to the newly created Symtab.  Is this a bug?  The Symtab object itself has a mutex to assure that the two threads don't corrupt it as the access/mutate it.  Unless Module can need to throw away one of these objects and create a new one, I don't see how the pointer returned becomes invalid.

augusto2112:
> > I like the change. Do we legitimately need access to the non-locked version of those functions too? If not, I think it would make sense to maybe move them to a different type or something like that, so new uses don't accidentally call the non-locked version? I think that would be an easy mistake to make.
> 
> Agreed. Before I put up the PR, I did a quick grep and there are _a lot_ (hundreds) of call sites. I'd have to audit them one-by-one to make sure it's safe to do the locking (most likely it's the right thing to do). So definitely we should aim to remove them, but I think it'll have to happen incrementally.

Could we rename the non locked case to reflect this? Otherwise I think it's too easy for anyone unaware of this change to reach out to that first.

JDevlieghere:
> Could we rename the non locked case to reflect this? Otherwise I think it's too easy for anyone unaware of this change to reach out to that first.

Yeah, I'd be supportive of that if we're okay with the churn.

JDevlieghere:
@jasonmolenda It's both issues. Not everything in those classes is protected, mostly lazy members that are computed later. Some examples include `m_type` and `m_strata` in `ObjectFile` and `m_calculated_abilities` and `m_type_list` in `SymbolFile`. You are correct that we can solve those at that layer (and I'd argue we should), but that still leaves the second issue you've correctly identified.

jasonmolenda:
Oh I see what you're doing, the Module's acquired-mutex gets returned along with the Symtab/ObjectFile/etc so other threads also trying to get a Symtab/ObjectFile will block until the first thread that won has completed whatever it was doing with the returned Symtab/ObjectFile/etc.  This does introduce more lock contention because now anyone operating on one of these objects within a Module precludes any other thread from accessing any of the other objects in the Module, until the first thread that gets the mutex has completed its work.  I doubt it's going to be a real performance issue, but it's extra lock holding unnecessarily in those cases.

medismailben:
Cool!

JDevlieghere:
> This does introduce more lock contention because now anyone operating on one of these objects within a Module precludes any other thread from accessing any of the other objects in the Module, until the first thread that gets the mutex has completed its work.

Yes, that's right. The alternative is to have separate mutexes for each of these, but then that means you need to be careful about the order in which you acquire the locks. We already synchronize a lot using the Module mutex, which is why I went this route. But your comment also highlights that you still need to be mindful/careful how long you hold onto this object.

EDIT: To elaborate on my statement that lots of things are using the Module mutex, take a look at SymbolFile. Everything that's synchronized there is already taking the Module mutex to do so. That keeps working because it's a recursive mutex, but it's likely no worse than what we had before. 

EDIT 2: https://github.com/llvm/llvm-project/pull/198963

Teemperor:
FWIW, I think me and Pavel were pondering a similar class a long time ago but never pushed to make it a thing (but that was maybe just us being lazy). So I think in principle everyone wants a class like this.

For the implementation, I hacked on more threaded server-side code and the one pattern I liked a lot was:

* Having a `Locked` class that is what we actually store in the object and that has data/mutex.
* An `Access` class that gives you read access to the resource. You can copy this access around. It will also only give you a const reference to the accessed object.
* A `WriteAccess` class that gives you unique write-access to the resource. You **can't** copy it to pass it around, but you get a non-const reference to the accessed object.

I attached some (not compiling) pseudo-code below. My only concern is that don't know how many read accesses we can have in LLDB given how much data is lazily loaded.

Another thing I would like to have in the far future is some kind of locking ordering  our code should adhere to (and we assert on). That is, we shouldn't allow a thread to access a Target after it already acquired a Module. That way we would solve at least some of our deadlocking issues deterministically. Unfortunately that pattern doesn't easily work with the way we get callbacks from dyld etc. (as those are on a different thread), but I guess we can't have it all.

Anyway, I don't think we should feature-creep this PR. The only thing I would nit about is to maybe rename `Locked` to `WriteAccess` or `WriteLock` to make it clear this is for write access, and that would allow us to add a `ReadAccess` later on without causing conflicts.

```cpp
template <typename T> struct LockedObject {
  friend struct Access<T>;
  friend struct WriteAccess<T>;

  ...
private:
  T data;
  mutable std::shared_mutex mutex;
};

template <typename T> struct Access {
  Access(const LockedObject<T> &target) : target(&target) {
    target.mutex.lock_shared();
  }
  ~Access() {
    target->mutex.unlock_shared();
  }

  const T *operator->() { return &target->data; }
  const T &operator*() const { return target->data; }

private:
   const LockedObject<T> *target = nullptr;
};

template <typename T> struct WriteAccess {
  // Can't copy.
  WriteAccess<T> &operator=(const WriteAccess<T> &o) = delete;
  WriteAccess(const WriteAccess<T> &target) = delete;

  WriteAccess(LockedObject<T> &target) : target(&target) {
    target.mutex.lock();
  }
  ~WriteAccess() {
    target->mutex.unlock();
  }

  T *operator->() { return &target->data; }
  T &operator*() const { return target->data; }

private:
  LockedObject<T> *target = nullptr;
};
```

JDevlieghere:
I've updated the PR to include only the new class, for which there seems to be general consensus. I made the following improvements based on feedback:

- Move-only exclusive access (e.g. write)
- Copyable shared access (e.g. read)
- Compatibility with shared and unique pointers

lldb/include/lldb/Utility/Locked.h
```diff
@@ -0,0 +1,142 @@
...
+public:
+  using mutex_type = Mutex;
+  using lock_type = std::unique_lock<Mutex>;
+
+  Locked() = default;
+  Locked(mutex_type &m, PtrT p) : m_lock(m), m_ptr(std::move(p)) {}
+  Locked(lock_type lock, PtrT p)
+      : m_lock(std::move(lock)), m_ptr(std::move(p)) {}
+
+  Locked(Locked &&) noexcept = default;
```
augusto2112: `noexcept` is back

lldb/include/lldb/Utility/Locked.h
```diff
@@ -0,0 +1,142 @@
...
+/// pass through code paths that branch or fan out without each leaf having
+/// to re-acquire.
+///
+/// The borrowed pointer is `const`-qualified so callers cannot mutate the
+/// pointee while holding only a reader's lock. `Mutex` must satisfy
+/// `SharedLockable` — `llvm::sys::RWMutex` (the LLDB convention) or
+/// `std::shared_mutex`. Use the `SharedLockedPtr`, `SharedLockedSP`,
+/// `SharedLockedUP` aliases for the common combinations.
+template <typename PtrT, typename Mutex = llvm::sys::RWMutex>
+class SharedLocked
```
augusto2112, line 98: Since the typedefs are what enforce that this is instantiated with a const pointer I think we should static assert that `PtrT` is always const-qualified in this type's constructor. Ideally we would want to hide these types and only allow access to the typedefs, but I don't think that's possible since these are templated.

lldb/include/lldb/Utility/Locked.h
augusto2112: A move of a raw pointer will leave the original pointer intact, this means that the:

```
  explicit operator bool() const { return Raw() != nullptr; }
```
check would return true for a moved locked. 

You probably want to implement this with std::exchange to null out the pointer.

lldb/include/lldb/Utility/Locked.h
```diff
@@ -0,0 +1,142 @@
...
+class Locked : public detail::LockedAccessors<Locked<PtrT, Mutex>, PtrT> {
+  friend class detail::LockedAccessors<Locked<PtrT, Mutex>, PtrT>;
+
+public:
+  using mutex_type = Mutex;
+  using lock_type = std::unique_lock<Mutex>;
+
+  Locked() = default;
+  Locked(mutex_type &m, PtrT p) : m_lock(m), m_ptr(std::move(p)) {}
+  Locked(lock_type lock, PtrT p)
```
augusto2112, line 64: You could add a 
```
    assert(m_lock.owns_lock() && "Locked requires an owning lock");
```
to make sure the lock that was passed in was actually locked.

lldb/include/lldb/Utility/Locked.h
```diff
@@ -0,0 +1,142 @@
...
+private:
+  std::shared_ptr<lock_type> m_lock;
+  PtrT m_ptr{};
+};
+
+/// Exclusive (write) access aliases. The default mutex is
+/// `std::recursive_mutex` to match the existing LLDB synchronization style.
+/// @{
+template <typename T, typename Mutex = std::recursive_mutex>
+using LockedPtr = Locked<T *, Mutex>;
```
augusto2112, line 149: Nit: personally I'd rename these to convey the fact that one is read only and the other is read/write.

lldb/include/lldb/Utility/Locked.h
```diff
@@ -0,0 +1,142 @@
...
+using LockedSP = Locked<std::shared_ptr<T>, Mutex>;
+
+template <typename T, typename Mutex = std::recursive_mutex>
+using LockedUP = Locked<std::unique_ptr<T>, Mutex>;
+/// @}
+
+/// Shared (read) access aliases. The default mutex is `llvm::sys::RWMutex`,
+/// the LLDB convention for read/write locks.
+/// @{
+template <typename T, typename Mutex = llvm::sys::RWMutex>
```
augusto2112, line 161: I see the comment says that the default for write is `std::recursive_mutex` and read-only is `llvm::sys::RWMutex` but why is that? I'm not super familiar with RWMutex, what does it buy us?

lldb/include/lldb/Utility/Locked.h
JDevlieghere: Claude really loves noexpect it seems. I even grepped for it but maybe I had a typo.

lldb/include/lldb/Utility/Locked.h
JDevlieghere: I don't really like the write nomenclature because it's not actually representative. A regular mutex synchronizes concurrent reads. I think Exclusive/Shared is better, which matches what the standard calls it. I omitted the Exclusive part because I'd argue it's the default.

lldb/include/lldb/Utility/Locked.h
JDevlieghere: The shared mutex implementation in libc++ is particularly slow on our system for back-deployment reasons. LLVM's RWMutex is a wrapper around pthread's rwmutex which is faster. On other platform it wraps shared_mutex.

See https://github.com/llvm/llvm-project/pull/197847 and https://github.com/llvm/llvm-project/pull/70151.

augusto2112 (approved):

lldb/unittests/Utility/LockedTest.cpp
```diff
@@ -0,0 +1,225 @@
...
+  Widget widget;
+
+  LockedPtr<Widget> handle(mutex, &widget);
+  ASSERT_TRUE(handle);
+  EXPECT_EQ(handle.get(), &widget);
+
+  // Recursive mutex lets us reacquire on the same thread, proving the lock
+  // really is held.
+  EXPECT_TRUE(mutex.try_lock());
+  mutex.unlock();
```
bulbazord: How about spawning another thread that calls `mutex.try_lock()` and asserts that it returns false? Right now you only check the same thread.

JDevlieghere: The tests are far from exhaustive. I intentionally kept them simple to act as smoke tests, rather than trying to cover the underlying locking implementation. But happy to add it if you think that's worth it.

bulbazord: I suppose it's about as worthy as testing re-acquiring the lock on the same thread? It's testing the semantics of `std::recursive_mutex` at that point. I'll leave it up to you.

bulbazord (approved):
lgtm

lldb/unittests/Utility/LockedTest.cpp
JDevlieghere: I was going to draw the line at one being a single line and the other not, but really it's simple enough that I've added it :-)

lldb/unittests/Utility/LockedTest.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+TEST(LockedTest, ExclusiveReleasesOnDestruction) {
+  std::mutex mutex;
+  Widget widget;
+  {
+    LockedPtr<Widget, std::mutex> handle(mutex, &widget);
+    EXPECT_FALSE(mutex.try_lock());
+  }
+  EXPECT_TRUE(mutex.try_lock());
+  mutex.unlock();
```
ingomueller-net, line 103: I am getting the following warning/error. Is that expected?

```
llvm-project/lldb/unittests/Utility/LockedTest.cpp:103:9: error: releasing mutex 'mutex' that was not held [-Werror,-Wthread-safety-analysis]
  103 |   mutex.unlock();
      |         ^
1 error generated.
```

ingomueller-net: Maybe #199218 is a solution?

PR #164638, opened 2025-10-22
[RFC] Extend FASTCC to use up to 22 registers under APX

phoebewang:
FASTCC calling convention is used for internal functions for better performance. X64 doesn't define a different calling convention for FASTCC. I think the reason is we have balanced caller/callee save registers defined in X64.

With APX feature, we have 16 more caller save registers. They can all be used for argument passing. So we extend FASTCC to use up to 22 registers in argument passing.

PR #198943, opened 2026-05-20
[coro] Use C calling convention for C++20 coroutines

### LINKED ISSUE #91123: coroutine frame from lambda
I think there are should be a way to create coroutine frame from lambda. There are use cases:

* when we sure dont need allocation here, just create a handle and pass into function which accepts coroutine handles
* when coroutine handle may be invoked concurrently from different threads ('when_any' implementation), but its not possible in standard C++ to handle this (calling .resume concurrently will lead to UB). With lambdas calling .resume concurrently may be handled with simple mutex
* to create 'always done' coroutine, just like std::noop_coroutine, but .done always returns true
* to create special coroutines, such as generator, which is always empty (without allocatons). This may be used in std::generator move operator (there are no default constructor for std::generator, but move constructor is present, so to create "default" state for such generator we can use some global generator object, which .resume does nothing and .destroy does nothing too)

I have "implementation", but for obvious reasons it is not good: it contains undefined behavior. And i dont know is it intentional, but clang ignores all noinline and std::launder hacks and just optimizes out all code every time (with O1/O2/O3):
https://godbolt.org/z/EP1b8Eb9x

```cpp
template <typename F>
struct alignas(__STDCPP_DEFAULT_NEW_ALIGNMENT__) coroutine_frame {
  void (*resume)(void*) = &do_resume<std::decay_t<coroutine_frame>>;
  void (*destroy)(void*) = &do_destroy<std::decay_t<coroutine_frame>>;
  F f;

  coroutine_frame(F value) : f(std::move(value)) {}

  [[gnu::noinline]] std::coroutine_handle<> handle() noexcept {
    return std::coroutine_handle<>::from_address(std::launder(this));
  }
};
```

vogelsgesang:
Change the calling convention for resume / destroy functions of C++ coroutines from `fastcc` to the C calling convention.

The resume / destroy functions are exposed as part of the coroutine ABI and must be compatible with other compilers and other versions of LLVM. fastcc is an LLVM-internal, unstable calling convention, though.

In practice, fastcc and the C calling convention are in sync for `void func(void*)` function signatures on almost all platforms. Therefore, I think we can still do this change without widespread ABI breakage.

`fastcc` and `ccc` do differ for i686 (x86-32), MIPS O32, PowerPC64 ELFv1 and Lanai. Afaik, those are all legacy ABIs and a recent feature like C++20 coroutines is unlikely to be used by projects still targeting legacy ABIs.

Historical context: I tried to figure out why `fastcc` was used. It is around since the original commit f93082e71a94 from 2016, predating the cross-vendor coroutine ABI work. I could not find any discussion of whether this should use `fastcc` on the review or on the mailing list.

As a side-effect, this also fixes #91123.

Co-authored-by: Cursor <cursoragent@cursor.com>

### DIFF
--- a/clang/docs/ReleaseNotes.rst
+++ b/clang/docs/ReleaseNotes.rst
@@ -107,6 +107,12 @@ ABI Changes in This Version
 - Clang now uses MSVC-compatible manglings for supported AArch64 SVE builtin
   types when targeting the Microsoft ABI. This changes symbol names for
   affected declarations compared to previous Clang releases. (#GH196170)
+- The resume and destroy functions of C++20 coroutines now use the platform C
+  calling convention instead of LLVM's internal ``fastcc``. This makes the
+  coroutine ABI stable across LLVM versions and interoperable with other
+  compilers. On most targets this is not a breaking change because ``fastcc``
+  and the platform C calling convention agree for ``void(ptr)``. It is an ABI
+  break on i686, MIPS O32, PowerPC64 ELFv1, and Lanai.
 
 AST Dumping Potentially Breaking Changes
 ----------------------------------------
--- a/clang/test/CodeGenCoroutines/coro-gro4.cpp
+++ b/clang/test/CodeGenCoroutines/coro-gro4.cpp
@@ -44,8 +44,8 @@ wrapper fn() { co_return; }
 // CHECK: call void @gro_destroy()
 // CHECK: ret void
 
-// CHECK: define internal fastcc void @_Z2fnv.resume
+// CHECK: define internal void @_Z2fnv.resume
 // CHECK-NOT: call void @gro_destroy()
 
-// CHECK: define internal fastcc void @_Z2fnv.destroy
+// CHECK: define internal void @_Z2fnv.destroy
 // CHECK-NOT: call void @gro_destroy()
--- a/clang/test/CodeGenCoroutines/coro-symmetric-transfer-03.cpp
+++ b/clang/test/CodeGenCoroutines/coro-symmetric-transfer-03.cpp
@@ -64,5 +64,5 @@ Task task0() {
 // CHECK-LABEL: define{{.*}} void @_Z5task0v.resume
 // This checks we are still in the scope of the current function.
 // CHECK-NOT: {{^}}}
-// CHECK: musttail call fastcc void
+// CHECK: musttail call void
 // CHECK-NEXT: ret void
--- a/clang/test/CodeGenCoroutines/coro-symmetric-transfer-04.cpp
+++ b/clang/test/CodeGenCoroutines/coro-symmetric-transfer-04.cpp
@@ -62,5 +62,5 @@ Task task0() {
 // CHECK-LABEL: define{{.*}} void @_Z5task0v.resume
 // This checks we are still in the scope of the current function.
 // CHECK-NOT: {{^}}}
-// CHECK: musttail call fastcc void
+// CHECK: musttail call void
 // CHECK-NEXT: ret void
--- a/clang/test/CodeGenCoroutines/pr65018.cpp
+++ b/clang/test/CodeGenCoroutines/pr65018.cpp
@@ -45,6 +45,6 @@ MyTask FooBar() {
 
 // CHECK: %[[RET:.+]] = {{.*}}call{{.*}}@_ZN7Awaiter13await_suspendESt16coroutine_handleIvE
 // CHECK: %[[RESUME_ADDR:.+]] = load ptr, ptr %[[RET]],
-// CHECK: musttail call fastcc void %[[RESUME_ADDR]]({{.*}}%[[RET]]
+// CHECK: musttail call void %[[RESUME_ADDR]]({{.*}}%[[RET]]
 // CHECK: ret
 
--- a/llvm/docs/Coroutines.rst
+++ b/llvm/docs/Coroutines.rst
@@ -382,7 +382,7 @@ Outlined resume part of the coroutine will reside in function `f.resume`:
 
 .. code-block:: llvm
 
-  define internal fastcc void @f.resume(ptr %frame.ptr.resume) {
+  define internal void @f.resume(ptr %frame.ptr.resume) {
   entry:
     %inc.spill.addr = getelementptr %f.frame, ptr %frame.ptr.resume, i64 0, i32 2
     %inc.spill = load i32, ptr %inc.spill.addr, align 4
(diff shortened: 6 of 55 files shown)


ChuanqiXu9:
I agree ABI compatiblity is important. But given fastcc is used for years, it may be a regression in fact to change that. Maybe we can add an opt-in option for people who don't care about ABI compatibility to still use fastcc.

vogelsgesang:
> Maybe we can add an opt-in option for people who don't care about ABI compatibility to still use fastcc.

I am not opposed, but would like to understand the motivation slightly better 🙂 

> may be a regression

Which type of regression are you concerned about?
Performance? ABI breakage? Something else?

Afaict, on all commonly used platforms (ARM, x64, many others), `fastcc` is de-facto identical with `ccc`, so it won't lead to a ABI breakage in practice for those ABIs.
Also performance-wise I think we should be fine.

The only case I can think of where this would break are LTO-links which might mix "old" with "new" post-split coroutines. In this code, the LTO-linker would see a disagreeing calling conventions and might miscompile. Not sure if LTO links across different LLVM versions are a supported scenario?

NewSigma:
I think `fastcc` is required for [TCO](https://llvm.org/docs/CodeGenerator.html#tail-call-optimization).

ChuanqiXu9:
> > Maybe we can add an opt-in option for people who don't care about ABI compatibility to still use fastcc.
> 
> I am not opposed, but would like to understand the motivation slightly better 🙂
> 
> > may be a regression
> 
> Which type of regression are you concerned about? Performance? ABI breakage? Something else?
> 
> Afaict, on all commonly used platforms (ARM, x64, many others), `fastcc` is de-facto identical with `ccc`, so it won't lead to a ABI breakage in practice for those ABIs. Also performance-wise I think we should be fine.
> 
> The only case I can think of where this would break are LTO-links which might mix "old" with "new" post-split coroutines. In this code, the LTO-linker would see a disagreeing calling conventions and might miscompile. Not sure if LTO links across different LLVM versions are a supported scenario?

Sorry, it seems like I misunderstand anything. If fastcc is the samething with ccc in all mainstream platform, why it is a compatibility problem and why why it solves https://github.com/llvm/llvm-project/issues/91123? I assume you're not solving a formal problem, right?

vogelsgesang:
> Sorry, it seems like I misunderstand anything. If fastcc is the samething with ccc in all mainstream platform, why it is a compatibility problem

**Today**, this is not a compatibility problem (for most platforms) because fastcc happens to be identical to ccc.
However, the fastcc calling convention is not considered stable, and LLVM can change it between releases.
That's at least how I understood the discussion in https://github.com/llvm/llvm-project/pull/164638 (@phoebewang please correct me if I misunderstood).
By exposing `fastcc` in a cross-module, cross-vendor way via coroutines, we are undermining the design goal that fastcc might be changed in the future.

For 32-bit i686, this is an actual ABI issue already today - and accordingly would be an ABI break. Not sure if we currently have any C++20 coroutine users which still compile for i686 and who would be impacted by this break.

> and why it solves https://github.com/llvm/llvm-project/issues/91123?

Currently, #91123 fails because the [InstCombineCalls](https://github.com/llvm/llvm-project/blob/577c2b6ccb1f358ed2948b6ae4cf443c4a842e75/llvm/lib/Transforms/InstCombine/InstCombineCalls.cpp#L4774) checks

```cplusplus
      // If the call and callee calling conventions don't match, and neither one
      // of the calling conventions is compatible with C calling convention
      // this call must be unreachable, as the call is undefined.
      if ((CalleeF->getCallingConv() != Call.getCallingConv() &&
           !(CalleeF->getCallingConv() == llvm::CallingConv::C &&
             TargetLibraryInfoImpl::isCallingConvCCompatible(&Call)) &&
           !(Call.getCallingConv() == llvm::CallingConv::C &&
             TargetLibraryInfoImpl::isCallingConvCCompatible(CalleeF))) &&
          // Only do this for calls to a function with a body.  A prototype may
          // not actually end up matching the implementation's calling conv for a
          // variety of reasons (e.g. it may be written in assembly).
          !CalleeF->isDeclaration()) {
        Instruction *OldCall = &Call;
        CreateNonTerminatorUnreachable(OldCall);
```

InstCombine sees two different calling conventions. It doesn't check if they happen to be compatible on the binary level. It decides that this is undefined behavior simply on the fact, that the calling conventions have different names

After this fix, the call-site also uses the C-calling convention, so the disagreement on calling conventions is gone

> I assume you're not solving a formal problem, right?

Re "ABI compatiblity with other vendors": yes, you might argue that this only fixes a "formal" problem
Re #91123: this actually fixes the problem

ChuanqiXu9:
Got it. Clear. LGTM

ChuanqiXu9 (approved):

ChuanqiXu9:
For tail call optimization, I think it is fine if we can be sure the symmetric transfer will be happen in X86 and AArch64, which I think is already checked by the test.
The TCO is an implementation for the symmetric transfer. However, on the one hand, it is not good to implement C++ feature in the middle end. This is false layering. On the other hand, the are already platforms not supporting symmetric transfer. So we're already not conforming in all platforms.

vogelsgesang:
> I think fastcc is required for [TCO](https://llvm.org/docs/CodeGenerator.html#tail-call-optimization).

Yep, that part of the documentation also confused me.
Afaict, the requirement for `fastcc` applies to the `tail` attribute, not the `musttail` attribute.

Looking at https://llvm.org/docs/LangRef.html#call-instruction, there are two attributes: `tail` and `musttail`.

For `tail` we have:

> Tail call optimization for calls marked tail is guaranteed to occur if the following conditions are met:
> * Caller and callee both have the calling convention fastcc or tailcc.
> * [...]
> * [Platform-specific constraints are met.](https://llvm.org/docs/CodeGenerator.html#tail-call-optimization)

I think the section you linked applies especially to `tail`, because:
1. For `musttail`, we don't have the constraint re `fastcc` documented in https://llvm.org/docs/LangRef.html#call-instruction
2. That https://llvm.org/docs/CodeGenerator.html#tail-call-optimization section is actually linked as a subbullet-point under the `tail` description, but not under the `musttail` description

vogelsgesang:
> Got it. Clear. LGTM

Thank you! 🙂 

I am planning to merge this in a week from now.
(I will wait a week to also give others an opportunity to chime in)

phoebewang:
I'm supportive of the change. Do not use fastcc for the sake of stability and backward compatibility.

> However, the fastcc calling convention is not considered stable, and LLVM can change it between releases.
> That's at least how I understood the discussion in #164638 (@phoebewang please correct me if I misunderstood).

Yes, it's a flexible calling convention that may change with target's capability. We have done similar thing in #164768 to limit the use of fastcc in LLVM.

PR #198944, opened 2026-05-21
[lldb][macos] Add dependency on Core in HostInfoMacOSX

jasonmolenda:


jasonmolenda:
https://github.com/llvm/llvm-project/pull/198931 broke the build on mac.
I made this PR adding a dependency on Core, but @JDevlieghere reminds me that we can only depend on Utility from Host, this PR is not acceptable.

PR #198948, opened 2026-05-21
[clang] Parse `__typeof_unqual__` consistently with `__typeof__`

Cons-Cat:
C23 `__typeof_unqual` and `__typeof_unqual__` are supported in all language modes as an extension.
However, existing tests missed this form:
```cpp
int main() {
   __typeof_unqual(int) x = 0;
}
```
That doesn't compile today.
```
<source>:2:4: error: expected expression
    2 |    __typeof_unqual(int) x = 0;
      |    ^
1 error generated.
Compiler returned: 1
```
I think the fix is to parse `tok::kw_typeof_unqual` everywhere that we currently parse `tok::kw_typeof`. It simply falls through `case`s beneath `tok::kw_typeof`, so this should handle them equivalently. No keyword semantics are changed by this PR.

While in here I noticed Clang Format ignores the token too, so I gave Clang Format the same parsing parity between these tokens. It now holds true under `isTypeName()`, just like `typeof`. `__typeof_unqual` kind of accidentally formatted correctly in C++ anyways, but not in C, so this is a meaningful fix. Previously it had no tests at all.

In the `clang-format` release notes, I'm not sure whether this should be considered a fix or a new feature. I've left the clang-format notes unchanged for now.


mydeveloperday (changes requested):
I would prefer for clang format changes to not touch Parser etc, when they support it we can

cor3ntin:
The clang changes looks good but i agree with @mydeveloperday  - can you split the PR in 2 parts (clang, clang format)?

cor3ntin (approved):
Thanks. Do you need me to merge that for you?

Cons-Cat:
> Thanks. Do you need me to merge that for you?

Hi, sorry for the late reply! Yes, if it looks mergeable that would be great.

PR #198949, opened 2026-05-21
   [lldb] Fix dummy target filtering regression in CommandInterpreter

itf:
In  #198429  (reland),  CommandObject::GetTarget()  was tightened to return  nullptr  instead of the dummy target when no real target exists, unless the command explicitly opts in via  eCommandAllowsDummyTarget  or standard  target requirements
 
 However in  CommandInterpreter::GetExecutionContext(bool adopt_dummy_target) :
  
  ```   
    ExecutionContext
    CommandInterpreter::GetExecutionContext(bool adopt_dummy_target) const {
      return !m_overriden_exe_contexts.empty()
                 ? m_overriden_exe_contexts.top()
                 : m_debugger.GetSelectedExecutionContext(adopt_dummy_target);
    }
  ```
  
  If  m_overriden_exe_contexts  is not empty, the method returned the top context immediately—completely ignoring the  adopt_dummy_target  argument requested by the command object.
  
  Because of this:
  
  1. During sourced script runs,  process attach  received the dummy target as its execution target (since  adopt_dummy_target = false  was ignored).
  2. It bypassed the  target == nullptr  check and proceeded to attach directly to the dummy target.
  3. As the dummy target was never registered in  m_target_list , the main target list remained empty ( No targets. ), causing all subsequent commands (e.g., setting breakpoints or continuing) to fail with  invalid target  errors.
  
  ### The Fix:
  
   lldb/source/Interpreter/CommandInterpreter.cpp :
  Respect  adopt_dummy_target = false  in  GetExecutionContext  when a dummy target is present in the overridden execution context stackm   so that if  adopt_dummy_target  is  false  and the overridden context on the stack contains the dummy target, we clear the context before returning it. This forces  GetTarget()  to return  nullptr  as originally intended.

### Test:
  
  •  lldb/test/Shell/Commands/process-attach-dummy.test :
  Add a new standalone Lit shell test to replicate this scenario. The test sources a command sequence executing  process attach  when no target exists, and verifies that  target list  successfully registers the newly created real
  target ( target #0: <none> ) instead of leaving the list empty.


lldb/source/Interpreter/CommandInterpreter.cpp
```diff
@@ -3284,9 +3284,15 @@ void CommandInterpreter::FindCommandsForApropos(llvm::StringRef search_word,
...
-             : m_debugger.GetSelectedExecutionContext(adopt_dummy_target);
+  if (!m_overriden_exe_contexts.empty()) {
+    ExecutionContext override_ctx = m_overriden_exe_contexts.top();
+    if (!adopt_dummy_target && override_ctx.GetTargetPtr() &&
+        override_ctx.GetTargetPtr()->IsDummyTarget()) {
+      override_ctx.Clear();
+    }
+    return override_ctx;
+  }
+  return m_debugger.GetSelectedExecutionContext(adopt_dummy_target);
```
cmtice: A cleaner alternative might be:

return (m_overriden_exe_contexts.empty() || !adopt_dummy_target)
                 ?  m_debugger.GetSelectedExecutionContext(adopt_dummy_target)
                  : m_overriden_exe_contexts.top();

(I think this basically does the same thing as your code...)

itf: Done! this also works

cmtice (approved):

JDevlieghere:
I don't believe this is correct. I think we should only bypass the override when its target is the dummy target. Otherwise breakpoint commands and stop hooks silently switch to the user's selected thread instead of the stop thread. The `adopt_dummy_target` means that we're going to adopt the dummy target if there's no real target. In other words, it signals that we *may* adopt the dummy target. That alone isn't sufficient reason to bypass the execution context stack.

Edit: I find it surprising if there would be no test failures with this patch. If that's because we don't test a breakpoint command or stop hook that operates on the non-selected thread, then we should certainly add that. If it's because this happens to still do the right thing indirectly, then let's figure if that's intentional and if we can make that more obvious (or document it).

PR #198951, opened 2026-05-21
[DAG] Convert select of constants to arithmetic operations

### LINKED ISSUE #196488: Miss optimization for ARM Code
Given an example:
```cpp
auto f(unsigned a, bool b)
{
    return a << (b ? 0 : 16);
}
```

Use `-march=armv7-m -O3` option to compile the code. The ARM GCC generates:
```asm
f(unsigned int, bool):
        eor     r1, r1, #1
        lsls    r1, r1, #4
        lsls    r0, r0, r1
        bx      lr
```

But clang generates:
```asm
f(unsigned int, bool):
        cmp     r1, #0
        mov.w   r1, #16
        it      ne
        movne   r1, #0
        lsls    r0, r1
        bx      lr
```

Here is [the link](https://gcc.godbolt.org/z/rYjzdE7fK) to godbolt.

dongjianqiang2:
Add the missing swapped case in DAGCombiner foldSelectOfConstants:
  select Cond, 0, Pow2 --> (zext (!Cond)) << log2(Pow2)
This fold is unconditional (like the existing 0/1, 1/0, 0/-1, -1/0 folds), so it fires for all targets without requiring a target hook. 

Fixes #196488

### DIFF
--- a/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
+++ b/llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
@@ -12753,6 +12753,15 @@ SDValue DAGCombiner::foldSelectOfConstants(SDNode *N) {
     return NotCond;
   }
 
+  // select Cond, 0, Pow2 --> (zext (!Cond)) << log2(Pow2)
+  if (C1->isZero() && C2->getAPIntValue().isPowerOf2()) {
+    SDValue NotCond = DAG.getNOT(DL, Cond, MVT::i1);
+    NotCond = DAG.getZExtOrTrunc(NotCond, DL, VT);
+    SDValue ShAmtC =
+        DAG.getShiftAmountConstant(C2->getAPIntValue().exactLogBase2(), VT, DL);
+    return DAG.getNode(ISD::SHL, DL, VT, NotCond, ShAmtC);
+  }
+
   // Use a target hook because some targets may prefer to transform in the
   // other direction.
   if (!shouldConvertSelectOfConstantsToMath(Cond, VT, TLI))
--- a/llvm/lib/Target/RISCV/RISCVISelLowering.cpp
+++ b/llvm/lib/Target/RISCV/RISCVISelLowering.cpp
@@ -1923,7 +1923,7 @@ RISCVTargetLowering::RISCVTargetLowering(const TargetMachine &TM,
   setTargetDAGCombine({ISD::INTRINSIC_VOID, ISD::INTRINSIC_W_CHAIN,
                        ISD::INTRINSIC_WO_CHAIN, ISD::ADD, ISD::SUB, ISD::MUL,
                        ISD::AND, ISD::OR, ISD::XOR, ISD::SETCC, ISD::SELECT,
-                       ISD::SRA});
+                       ISD::SRA, ISD::SHL});
   setTargetDAGCombine(ISD::SIGN_EXTEND_INREG);
 
   if (Subtarget.hasStdExtFOrZfinx())
@@ -1952,7 +1952,6 @@ RISCVTargetLowering::RISCVTargetLowering(const TargetMachine &TM,
                          ISD::VP_GATHER,
                          ISD::VP_SCATTER,
                          ISD::SRL,
-                         ISD::SHL,
                          ISD::STORE,
                          ISD::SPLAT_VECTOR,
                          ISD::BUILD_VECTOR,
@@ -21261,6 +21260,29 @@ static SDValue performSHLCombine(SDNode *N,
   if (SDValue V = combineOp_VLToVWOp_VL(N, DCI, Subtarget))
     return V;
 
+  SelectionDAG &DAG = DCI.DAG;
+
+  // Reverse the generic fold for select Cond, 0, Pow2: when Cond is an i1
+  // function argument promoted via AssertZext, prefer the select lowering
+  // (addi -1; andi Pow2) over the generic xori 1; slli C, because the former
+  // is two RVC-compressible instructions for small Pow2.
+  if (DCI.isAfterLegalizeDAG()) {
+    using namespace SDPatternMatch;
+    EVT VT = N->getValueType(0);
+    SDValue X;
+    uint64_t ShAmtVal;
+    if (!VT.isVector() &&
+        sd_match(N, m_Shl(m_OneUse(m_Xor(m_Value(X), m_One())),
+                          m_ConstInt(ShAmtVal))) &&
+        X.getOpcode() == ISD::AssertZext &&
+        cast<VTSDNode>(X.getOperand(1))->getVT() == MVT::i1) {
+      SDLoc DL(N);
+      APInt Pow2 = APInt::getOneBitSet(VT.getSizeInBits(), ShAmtVal);
+      return DAG.getNode(ISD::SELECT, DL, VT, X, DAG.getConstant(0, DL, VT),
+                         DAG.getConstant(Pow2, DL, VT));
+    }
+  }
+
   // (shl (sext x), C) -> (vwmulsu x, 1u << C)
   // (shl (zext x), C) -> (vwmulu  x, 1u << C)
 
@@ -21310,7 +21332,6 @@ static SDValue performSHLCombine(SDNode *N,
   if (NarrowBits * 2 != VT.getScalarSizeInBits())
     return SDValue();
 
-  SelectionDAG &DAG = DCI.DAG;
   SDLoc DL(N);
   SDValue Passthru, Mask, VL;
   switch (N->getOpcode()) {
--- a/llvm/test/CodeGen/RISCV/select-const.ll
+++ b/llvm/test/CodeGen/RISCV/select-const.ll
@@ -102,8 +102,8 @@ define signext i32 @select_const_int_harder(i1 zeroext %a) nounwind {
 ;
 ; RV32ZICOND-LABEL: select_const_int_harder:
 ; RV32ZICOND:       # %bb.0:
-; RV32ZICOND-NEXT:    xori a0, a0, 1
-; RV32ZICOND-NEXT:    slli a0, a0, 5
+; RV32ZICOND-NEXT:    addi a0, a0, -1
+; RV32ZICOND-NEXT:    andi a0, a0, 32
 ; RV32ZICOND-NEXT:    addi a0, a0, 6
 ; RV32ZICOND-NEXT:    ret
 ;
@@ -135,8 +135,8 @@ define signext i32 @select_const_int_harder(i1 zeroext %a) nounwind {
 ;
 ; RV64ZICOND-LABEL: select_const_int_harder:
 ; RV64ZICOND:       # %bb.0:
-; RV64ZICOND-NEXT:    xori a0, a0, 1
-; RV64ZICOND-NEXT:    slli a0, a0, 5
+; RV64ZICOND-NEXT:    addi a0, a0, -1
+; RV64ZICOND-NEXT:    andi a0, a0, 32
 ; RV64ZICOND-NEXT:    addiw a0, a0, 6
 ; RV64ZICOND-NEXT:    ret
   %1 = select i1 %a, i32 6, i32 38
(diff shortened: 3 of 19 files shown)


llvm/test/CodeGen/ARM/select-const.ll
```diff
@@ -0,0 +1,19 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+; RUN: llc -mtriple=armv7-linux-gnueabi < %s \
```
arsenm: Don't need the environment, and use canonical triple

llvm/test/CodeGen/ARM/select-const.ll
```diff
@@ -0,0 +1,19 @@
...
+; RUN:   | FileCheck %s --check-prefix=CHECK
```
arsenm: Don't use -check-prefix if it's the default

llvm/test/CodeGen/ARM/select-const.ll
```diff
@@ -0,0 +1,19 @@
...
+; CHECK-LABEL: select_zero_pow2:
+; CHECK:       @ %bb.0:
+; CHECK-NEXT:    eor r1, r1, #1
+; CHECK-NEXT:    lsl r1, r1, #4
+; CHECK-NEXT:    lsl r0, r0, r1
+; CHECK-NEXT:    bx lr
+  %cond = select i1 %b, i32 0, i32 16
+  %shl = shl i32 %a, %cond
+  ret i32 %shl
+}
```
arsenm: Test vector case? Can you add this to an existing test for this combine?

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
```diff
@@ -12766,15 +12766,25 @@ SDValue DAGCombiner::foldSelectOfConstants(SDNode *N) {
...
+  if (C1Val.isZero() && C2Val.isPowerOf2()) {
+    SDValue NotCond = DAG.getNOT(DL, Cond, MVT::i1);
+    NotCond = DAG.getZExtOrTrunc(NotCond, DL, VT);
+    SDValue ShAmtC = DAG.getShiftAmountConstant(C2Val.exactLogBase2(), VT, DL);
+    return DAG.getNode(ISD::SHL, DL, VT, NotCond, ShAmtC);
+  }
+
   // Use a target hook because some targets may prefer to transform in the
   // other direction.
   if (!shouldConvertSelectOfConstantsToMath(Cond, VT, TLI))
```
arsenm, line 12767: Putting this before shouldConvertSelectOfConstantsToMath is suspect

llvm/test/CodeGen/ARM/select-const.ll
dongjianqiang2: select-const.ll has been deleted, and its tests have been merged into the existing select-imm.ll (at the end of the file). Regarding vector tests — since foldSelectOfConstants only handles the case where scalar CondVT == MVT::i1, which does not apply to vectors, I’ve left out the vector tests for now. Please let me know if you’d prefer to have them included.

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
dongjianqiang2: select 0, 1 is already unconditional; select 0, Pow2 is its natural generalization (equivalent when Pow2=1). Placing this fold before the hook can avoid the OOM caused by enabling the ARM hook.

llvm/test/CodeGen/ARM/select-const.ll
dongjianqiang2: fixed

llvm/test/CodeGen/ARM/select-const.ll
dongjianqiang2: fixed

llvm/test/CodeGen/RISCV/select-zbb.ll
```diff
@@ -526,23 +526,23 @@ define i32 @select_umax_3(i1 zeroext %cond, i32 %a) {
...
+; RV64IMZBB-NEXT:    xori a0, a0, 1
+; RV64IMZBB-NEXT:    slli a0, a0, 5
 ; RV64IMZBB-NEXT:    maxu a0, a1, a0
 ; RV64IMZBB-NEXT:    ret
 ;
 ; RV32IMZICOND-LABEL: select_umax_3:
 ; RV32IMZICOND:       # %bb.0: # %entry
-; RV32IMZICOND-NEXT:    addi a0, a0, -1
-; RV32IMZICOND-NEXT:    andi a0, a0, 32
+; RV32IMZICOND-NEXT:    xori a0, a0, 1
```
svs-quic: Both `addi` and `andi` can be compressed to 2-byte instructions whereas `xori` can never be compressed. I don't think this transformation is always beneficial for RISCV.
@topperc thoughts?

dongjianqiang2: Thanks for catching the xori code size issue. I'm considering two options:  
1. Move (0, Pow2) behind the convertSelectOfConstantsToMath hook (opt-in), same as (Pow2, 0).
2. Keep it unconditional but add an opt-out hook for targets where it regresses. 
Thoughts on which direction you'd prefer?

dongjianqiang2: Moving it behind the hook doesn't help RISCV since the hook already returns true. The cleanest fix would be for RISCV to gate this in its own hook override (e.g., return false when the C extension is enabled), keeping the code-size tradeoff in the backend where it belongs.

topperc: I don't think we should add yet another target hook. We have too many already. Maybe we can add a DAG combine on the RISC-V side to put it back to the old form.

dongjianqiang2: Hi @topperc @svs-quic, I'm not very familiar with the RISC-V backend, so I tried asking an LLM to help reverse-select the `Cond,0,Pow2` fold for C/Zca code size ; )
Could you help sanity-check whether this looks reasonable?

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
svs-quic: Can you use `SDPatternMatch`  here to simplify the code? Something like:

```suggestion
// Reverse the generic fold: select Cond, 0, Pow2 -> (zext (!Cond)) <<
  // log2(Pow2) back to select Cond, 0, Pow2 when the C or Zca extension is
  // enabled, because the select lowering produces addi -1; andi Pow2 which
  // compresses to 2-byte instructions, whereas xori 1 never compresses.
  //
  // After legalization, the zext+xor pattern becomes:
  //   shl (xor (assertzext Cond, i1), 1), C
  if (DCI.isAfterLegalizeDAG() &&
      (Subtarget.hasStdExtC() || Subtarget.hasStdExtZca())) {
    using namespace SDPatternMatch;

    EVT VT = N->getValueType(0);
    SDValue Cond, AssertZextVT;
    uint64_t ShAmtVal;

    if (!VT.isVector() &&
        sd_match(N, m_Shl(m_OneUse(m_Xor(m_Node(ISD::AssertZext, m_Value(Cond),
                                                m_Value(AssertZextVT)),
                                          m_One())),
                          m_ConstInt(ShAmtVal))) &&
        cast<VTSDNode>(AssertZextVT)->getVT() == MVT::i1) {
      APInt Pow2 = APInt(VT.getSizeInBits(), 1) << ShAmtVal;
      SDLoc DL(N);
      return DAG.getNode(ISD::SELECT, DL, VT, Cond, DAG.getConstant(0, DL, VT),
                         DAG.getConstant(Pow2, DL, VT));
    }
  }
```

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
arsenm: I don't know what you mean by "Avoid the OOM". This shouldn't disrespect the hook

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
dongjianqiang2: Done, switched to sd_match. Thanks for the suggestion.

llvm/lib/CodeGen/SelectionDAG/DAGCombiner.cpp
dongjianqiang2: If the fold were behind the hook, the only way to fix it would be making ARM's convertSelectOfConstantsToMath return true, but that also enables the C1/C1±1 folds which trigger infinite DAGCombine loops on ARM — that's the OOM :(

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
topperc: You only need to check hasStdExtZca()

topperc: APInt::getOneBitSet(VT.getSizeInBits(), ShAmtVal)

topperc: assertzext is only present in the function argument case. This needs to use computeKnownBits to handle the general case of zero extended bool.

dongjianqiang2: Thanks, done. Added explicit ZERO_EXTEND from i1 matching alongside AssertZext — computeKnownBits at the post-legalize stage also matches and X, 1 patterns that aren't select→shift reversals, so the explicit opcode check is safer here.

dongjianqiang2: done

dongjianqiang2: done

llvm/test/CodeGen/ARM/select-const.ll
krzysz00: Is this applicable to splat vector constants or are is there already a `select(C, splat(V1), splat(V2))` somewhere - if it's up in IR, it's fine.

dongjianqiang2: Possibly, but the scalar case was straightforward. For vectors, the legalization behavior of `zext <N x i1> → <N x i32>` varies across targets, and there isn't a clear motivating case yet. Happy to revisit if someone runs into a concrete need.

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
topperc: After legalization, we will have setcc with an i32 or i64 result type that is either 0/1. There will be no zero extend. We need to use computeKnownBits or MaskedValueIsZero.

topperc: You can drop the Zca check. We can do this unconditionally. We try to minimize differences between Zca and no-Zca output if it isn't harmful.

llvm/lib/Target/RISCV/RISCVISelLowering.cpp
```diff
@@ -21238,6 +21237,38 @@ static SDValue performSHLCombine(SDNode *N,
...
+    using namespace SDPatternMatch;
+    EVT VT = N->getValueType(0);
+    SDValue X;
+    uint64_t ShAmtVal;
+    if (!VT.isVector() &&
+        sd_match(N, m_Shl(m_OneUse(m_Xor(m_Value(X), m_One())),
+                          m_ConstInt(ShAmtVal)))) {
+      SDLoc DL(N);
+      SDValue Cond;
+      if (X.getOpcode() == ISD::AssertZext &&
```
topperc: computeKnownBits can handle AssertZExt. We don't need to check it separately.

llvm/test/CodeGen/RISCV/select-const.ll
```diff
@@ -796,8 +796,9 @@ define i32 @diff_shl_addi(i32 signext %x) {
 ;
 ; RV32ZICOND-LABEL: diff_shl_addi:
 ; RV32ZICOND:       # %bb.0:
+; RV32ZICOND-NEXT:    lui a1, 4
 ; RV32ZICOND-NEXT:    srli a0, a0, 31
-; RV32ZICOND-NEXT:    slli a0, a0, 14
+; RV32ZICOND-NEXT:    czero.eqz a0, a1, a0
```
topperc: This is a regression

dongjianqiang2: Thanks, you're right. The computeKnownBits guard was too loose and caught (srli a, XLen-1) too. Tightening it to AssertZext-of-i1 (the actual case where select lowering produces compressible addi -1; andi Pow2), Posting an updated patch.

topperc:
please dont' force push to PRs during review. It makes it hard for reviewers to see changes.

PR #198953, opened 2026-05-21
[LLDB] Add a progress event to xcrun invocations

adrian-prantl:
LLDB invokes xcrun to find SDKs on disk. This is usually very fast, but sometimes (after an Xcode update, or when the searched SDK does not exist) it can take very long (10s or more). The progress event provides user feedback to explain the hang.

Previously this functionality was in HostInfo, but it may not depend on Core, so this patch is instrumenting the call sites instead.

cf: https://github.com/llvm/llvm-project/pull/198931

### DIFF
--- a/lldb/source/Core/Module.cpp
+++ b/lldb/source/Core/Module.cpp
@@ -14,6 +14,7 @@
 #include "lldb/Core/Debugger.h"
 #include "lldb/Core/Mangled.h"
 #include "lldb/Core/ModuleSpec.h"
+#include "lldb/Core/Progress.h"
 #include "lldb/Core/SearchFilter.h"
 #include "lldb/Core/Section.h"
 #include "lldb/Host/FileSystem.h"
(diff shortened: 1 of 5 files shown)


JDevlieghere (approved):
LGTM :(

felipepiovezan:
The three files in `Plugins/Platform/MacOSX`  are all implementing the same find-sdk-or-report-error multiple times, with inconsistent error printing in each of them.  I'm sure there is a common header in that directory (or it can be created) where a helper function could be declared, and defined in any of the `cpp` files of that directory.  
This way you also restore a good chunk of the original PR's conciseness.

adrian-prantl:
> The three files in `Plugins/Platform/MacOSX` are all implementing the same find-sdk-or-report-error multiple times, with inconsistent error printing in each of them. I'm sure there is a common header in that directory (or it can be created) where a helper function could be declared, and defined in any of the `cpp` files of that directory. This way you also restore a good chunk of the original PR's conciseness.

Good point!

https://github.com/llvm/llvm-project/pull/199141

PR #198955, opened 2026-05-21
[Offload] Fix LLVM_LINK_LLVM_DYLIB linking liboffload

jhuber6:
Summary:
When this is set you can only link against `LLVM`. The previous patch
did not respect this because I did not realize that internally in the
add_llvm_library that this was required.


mgorny (approved):
This fixes the issue for me. Thanks!

PR #198958, opened 2026-05-21
[CIR] Fix cast kind to support bool in builtin_*_overflow

erichkeane:
'bool' is an acceptable type to the __builtin_*_overflow functions, but we were unconditionally doing an integral cast. This adds some logic to do a bool-to-int cast if necessary.


andykaylor (approved):
lgtm

PR #198959, opened 2026-05-21
[libc][semaphore] Add post and wait operations for internal semaphore

Pengxiang-Huang:
This pr implements https://github.com/llvm/llvm-project/issues/190847, adding the post and wait operations in posix semaphore for linux using futex. 

Particularly, 
`sem_post`: https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_post.html
`sem_trywait`, `sem_wait`: https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_wait.html
`sem_timedwait`, `sem_clockwait`: https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_clockwait.html#

The implementation is assisted by Claude Code Opus 4.7, specifically for the unit tests generations.

### DIFF
--- a/libc/src/semaphore/linux/semaphore.h
+++ b/libc/src/semaphore/linux/semaphore.h
@@ -18,6 +18,7 @@
 #include "src/__support/CPP/limits.h"
 #include "src/__support/common.h"
 #include "src/__support/error_or.h"
+#include "src/__support/libc_assert.h"
 #include "src/__support/threads/futex_utils.h"
 #include "src/__support/time/abs_timeout.h"
 
@@ -31,6 +32,9 @@ class Semaphore {
   Futex value;
   unsigned int canary;
 
+  // Whether the semaphore is shared between processes.
+  bool is_shared;
+
   // A private constant canary used to detect use of uninitialized or
   // destroyed semaphores. Chose "SEM1" in ASCII (0x53='S', 0x45='E',
   // 0x4D='M', 0x31='1').
@@ -43,7 +47,7 @@ class Semaphore {
       if (trywait() == 0)
         return 0;
 
-      auto wait_or = value.wait(/*expected=*/0, timeout, /*is_shared=*/true);
+      auto wait_or = value.wait(/*expected=*/0, timeout, is_shared);
       if (!wait_or.has_value() && wait_or.error() == ETIMEDOUT) {
         // Final attempt in case a post() raced with the timeout.
         if (trywait() == 0)
@@ -54,8 +58,8 @@ class Semaphore {
   }
 
 public:
-  LIBC_INLINE constexpr Semaphore(unsigned int value)
-      : value(value), canary(SEM_CANARY) {}
+  LIBC_INLINE constexpr Semaphore(unsigned int value, bool shared)
+      : value(value), canary(SEM_CANARY), is_shared(shared) {}
 
   // Sanity check to detect use of uninitialized or destroyed semaphores.
   LIBC_INLINE bool is_valid() const { return canary == SEM_CANARY; }
@@ -92,9 +96,7 @@ public:
 
     // Wake one waiter if any,
     // waiter selection is left to linux futex implementation.
-    // Named semaphores live in MAP_SHARED memory and may be shared across
-    // processes, so use a shared wake.
-    value.notify_one(/*is_shared=*/true);
+    value.notify_one(is_shared);
     return 0;
   }
 
@@ -123,8 +125,7 @@ public:
       // Futex wait re-checks the value atomically,
       // so a racing post() before sleep is not lost.
       // Spurious wakeups just send back to trywait().
-      (void)value.wait(/*expected=*/0, /*timeout=*/cpp::nullopt,
-                       /*is_shared=*/true);
+      (void)value.wait(/*expected=*/0, /*timeout=*/cpp::nullopt, is_shared);
     }
   }
 
@@ -145,7 +146,10 @@ public:
     if (LIBC_LIKELY(timeout.has_value()))
       return wait_until(timeout.value());
 
-    switch (timeout.error()) {
+    internal::AbsTimeout::Error err = timeout.error();
+    LIBC_ASSERT(err == internal::AbsTimeout::Error::Invalid ||
+                err == internal::AbsTimeout::Error::BeforeEpoch);
+    switch (err) {
     case internal::AbsTimeout::Error::Invalid:
       return EINVAL;
     case internal::AbsTimeout::Error::BeforeEpoch:
--- a/libc/src/semaphore/linux/semaphore.h
+++ b/libc/src/semaphore/linux/semaphore.h
@@ -10,12 +10,16 @@
 #define LLVM_LIBC_SRC_SEMAPHORE_LINUX_SEMAPHORE_H
 
 #include "hdr/errno_macros.h"
+#include "hdr/time_macros.h"
+#include "hdr/types/clockid_t.h"
 #include "hdr/types/mode_t.h"
+#include "hdr/types/struct_timespec.h"
 #include "src/__support/CPP/atomic.h"
 #include "src/__support/CPP/limits.h"
 #include "src/__support/common.h"
 #include "src/__support/error_or.h"
 #include "src/__support/threads/futex_utils.h"
+#include "src/__support/time/abs_timeout.h"
 
 namespace LIBC_NAMESPACE_DECL {
 
@@ -32,11 +36,24 @@ class Semaphore {
   // 0x4D='M', 0x31='1').
   static constexpr unsigned int SEM_CANARY = 0x53454D31U;
 
-public:
-  // TODO:
-  // Add the blocking waiting operations: sem_wait, sem_timedwait,
-  //    sem_clockwait.
+  // Helper function for timed blocking wait, wait for a required
+  // absolute timeout.
+  LIBC_INLINE int wait_until(internal::AbsTimeout timeout) {
+    for (;;) {
+      if (trywait() == 0)
+        return 0;
 
+      auto wait_or = value.wait(/*expected=*/0, timeout, /*is_shared=*/true);
+      if (!wait_or.has_value() && wait_or.error() == ETIMEDOUT) {
+        // Final attempt in case a post() raced with the timeout.
+        if (trywait() == 0)
+          return 0;
+        return ETIMEDOUT;
+      }
+    }
+  }
+
+public:
   LIBC_INLINE constexpr Semaphore(unsigned int value)
       : value(value), canary(SEM_CANARY) {}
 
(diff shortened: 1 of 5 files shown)


SchrodingerZhu:
Sorry, I was busy with moving and training. I am taking a look at the patch now.

libc/src/semaphore/linux/semaphore.h
```diff
@@ -54,6 +77,83 @@ class Semaphore {
...
+    if (LIBC_LIKELY(timeout.has_value()))
+      return wait_until(timeout.value());
+
+    switch (timeout.error()) {
+    case internal::AbsTimeout::Error::Invalid:
+      return EINVAL;
+    case internal::AbsTimeout::Error::BeforeEpoch:
+      return ETIMEDOUT;
+    }
+    __builtin_unreachable();
```
SchrodingerZhu: Could you add sth like:
```
internal::AbsTimeout::Error err = timeout.error();
LIBC_ASSERT(err == internal::AbsTimeout::Error::Invalid || err == internal::AbsTimeout::Error::BeforeEpoch);
```
to make it easier to track if an unreachable state really happens?

libc/src/semaphore/linux/semaphore.h
```diff
@@ -26,11 +36,24 @@ class Semaphore {
...
-  // Add the posting and waiting operations: sem_post, sem_wait,
-  //    sem_trywait, sem_timedwait, sem_clockwait.
+  // Helper function for timed blocking wait, wait for a required
+  // absolute timeout.
+  LIBC_INLINE int wait_until(internal::AbsTimeout timeout) {
+    for (;;) {
+      if (trywait() == 0)
+        return 0;
+
+      auto wait_or = value.wait(/*expected=*/0, timeout, /*is_shared=*/true);
```
SchrodingerZhu: POSIX.1-2024 says:

> If the pshared argument has a non-zero value, then the semaphore is shared between processes; in this case, any process that can access the semaphore sem can use sem for performing [sem_clockwait()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_clockwait.html), [sem_destroy()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_destroy.html), [sem_post()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_post.html), [sem_timedwait()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_timedwait.html), [sem_trywait()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_trywait.html), and [sem_wait()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_wait.html) operations.

> If the pshared argument is zero, then the semaphore is shared between threads of the process; any thread in this process can use sem for performing [sem_clockwait()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_clockwait.html), [sem_destroy()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_destroy.html), [sem_post()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_post.html), [sem_timedwait()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_timedwait.html), [sem_trywait()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_trywait.html), and [sem_wait()](https://pubs.opengroup.org/onlinepubs/9799919799/functions/sem_wait.html) operations.

why do we always default to shared addressing mode here and below?

SchrodingerZhu (changes requested):

SchrodingerZhu:
Overall the code shape looks good to me. However, there are some details we need to discuss. Thank you for doing this!

libc/src/semaphore/linux/semaphore.h
Pengxiang-Huang: `is_shared` is used for the kernel to compute the futex key for identifying the sleepers/waker. The kernel can still compute the right key in the private mode, but with the unnecessary page table walking overhead. It's easier to avoid writing additional control flow handling, but I also agree it can be confusing to understand as you pointed out in the posix description.

SchrodingerZhu: I don’t think that’s entirely accurate.

1. Shared futexes do incur additional overhead because the kernel needs to resolve the futex key through the underlying shared mapping, whereas private futexes can be keyed directly from the process address space.
2. More importantly, private and shared futexes belong to different futex-key namespaces. A wake operation issued on a private futex will not match waiters sleeping on the corresponding shared futex, and vice versa.

Because of this, the POSIX pshared attribute should generally remain consistent with the futex flags used by the implementation.

SchrodingerZhu (approved):
LGTM. Thanks for doing this.

libc/src/semaphore/linux/semaphore.h
```diff
@@ -54,6 +81,83 @@ class Semaphore {
...
+  LIBC_INLINE int wait() {
+    for (;;) {
+      if (trywait() == 0)
+        return 0;
+
+      // Blocks when value was zero.
+      // Futex wait re-checks the value atomically,
+      // so a racing post() before sleep is not lost.
+      // Spurious wakeups just send back to trywait().
+      (void)value.wait(/*expected=*/0, /*timeout=*/cpp::nullopt, is_shared);
```
michaelrj-google: this should probably handle the error cases

libc/src/semaphore/linux/semaphore.h
```diff
@@ -54,6 +81,83 @@ class Semaphore {
...
+                                      cpp::MemoryOrder::RELAXED))
+        return 0;
+    }
+
+    return EAGAIN;
+  }
+
+  // Blocking wait, decrements the value when positive, or blocks until a
+  // post() makes it positive again.
+  LIBC_INLINE int wait() {
```
michaelrj-google, line 123: as currently written this function only ever returns 0. I think it should return an error code if the wait call fails, but if not it should probably be void since it only has one possible return.

libc/src/semaphore/linux/semaphore.h
```diff
@@ -54,6 +81,83 @@ class Semaphore {
...
+    for (;;) {
+      if (trywait() == 0)
+        return 0;
```
michaelrj-google: This could be simplified:
```suggestion
    while (trywait() == 0) {
```
Then put `return 0` after the end of the loop

libc/src/semaphore/linux/semaphore.h
```diff
@@ -9,30 +9,64 @@
...
+  // Helper function for timed blocking wait, wait for a required
+  // absolute timeout.
+  LIBC_INLINE int wait_until(internal::AbsTimeout timeout) {
+    while (trywait() != 0) {
+      auto wait_or = value.wait(/*expected=*/0, timeout, is_shared);
+      if (wait_or.has_value())
+        continue;
+
+      if (wait_or.error() == EAGAIN)
+        continue;
```
michaelrj-google: Combine
```suggestion
      if (wait_or.has_value() || (wait_or.error() == EAGAIN))
        continue;
```

libc/src/semaphore/linux/semaphore.h
```diff
@@ -9,30 +9,64 @@
...
-      : value(value), canary(SEM_CANARY) {}
+      if (wait_or.error() == ETIMEDOUT) {
+        // Final attempt in case a post() raced with the timeout.
+        if (trywait() == 0)
+          return 0;
+        return ETIMEDOUT;
+      }
+
+      // Any other error is unexpected (e.g. EINVAL/EFAULT).
+      return wait_or.error();
```
michaelrj-google: the unconditional return at the end of the loop is a sign that the loop condition is incorrect. Is there a way to reorder the loop such that there are no `continue`s?

Specifically it looks like `wait_or.has_value()` and `wait_or.error() == EAGAIN` are what cause the loop to continue, and if those fail then the loop ends.

michaelrj-google (commented):
Overall looks good, a few comments on organization

Pengxiang-Huang:
@michaelrj-google Can you help review the changes one more time, Thanks!

libc/src/semaphore/linux/semaphore.h
```diff
@@ -54,6 +85,87 @@ class Semaphore {
...
+    internal::AbsTimeout::Error err = timeout.error();
+    LIBC_ASSERT(err == internal::AbsTimeout::Error::Invalid ||
+                err == internal::AbsTimeout::Error::BeforeEpoch);
+    switch (err) {
+    case internal::AbsTimeout::Error::Invalid:
+      return EINVAL;
+    case internal::AbsTimeout::Error::BeforeEpoch:
+      return ETIMEDOUT;
+    }
+    __builtin_unreachable();
```
michaelrj-google: given that there's an assert above to check if this is a valid state it might make more sense for this to be `__builtin_trap()`. That way on a non-debug build that reaches this supposedly unreachable state it will properly stop instead of doing something undefined.

libc/test/src/semaphore/linux/semaphore_test.cpp
```diff
@@ -8,24 +8,129 @@
...
+  Semaphore sem(2, /*is_shared=*/false);
+
+  // value is positive: wait() should decrement without blocking.
+  ASSERT_EQ(sem.wait(), 0);
+  ASSERT_EQ(sem.getvalue(), 1);
+  ASSERT_EQ(sem.wait(), 0);
+  ASSERT_EQ(sem.getvalue(), 0);
+}
+
+TEST(LlvmLibcSemaphoreTest, TimedWaitNonBlocking) {
```
michaelrj-google, line 67: The standard says: "Under no circumstance shall the function fail with a timeout if the semaphore can be locked immediately. The validity of the abstime need not be checked if the semaphore can be locked immediately."

Add a test to cover this case. Specifically I want you to pass a null `abstime` to ensure it's never dereferenced.

libc/test/src/semaphore/linux/semaphore_test.cpp
```diff
@@ -8,24 +8,129 @@
...
 #include "hdr/fcntl_macros.h"
+#include "hdr/time_macros.h"
+#include "hdr/types/struct_timespec.h"
+#include "src/__support/time/clock_gettime.h"
 #include "src/semaphore/linux/semaphore.h"
 #include "test/UnitTest/Test.h"
 
 using LIBC_NAMESPACE::Semaphore;
 
 TEST(LlvmLibcSemaphoreTest, InitAndGetValue) {
```
michaelrj-google, line 19: add a test covering the case where the semaphore number is greater than `SEM_VALUE_MAX`.

michaelrj-google (commented):
Update the tests and ensure they still pass, then LGTM

libc/test/src/semaphore/linux/semaphore_test.cpp
Pengxiang-Huang: I added `if (trywait() == 0) return` to guard the null ptr dereference case.

michaelrj-google (approved):
LGTM

Pengxiang-Huang:
@michaelrj-google Could you help to land this when you can? Thanks!

michaelrj-google:
Merged, thanks for working on this!

PR #198962, opened 2026-05-21
[CIR] Add custom assembly format for alloca op to fix flag parsing

Andres-Salamanca:
The previously used assembly format was generating code like:
```cpp
  if (::mlir::succeeded(parser.parseOptionalComma())) {
    props.init = parser.getBuilder().getUnitAttr();
    if (parser.parseKeyword("init"))
      return ::mlir::failure();
  }
```
This means that upon seeing any comma, the parser would immediately set the `init` attribute and then expect the keyword "init" to follow. So a valid input like `["n", const]` would fail with:
```bash
  error: expected 'init'
```


xlauko:
What about changing the format non-ambiguous parse that does not need lookahead, i.e. dropping commas:

```
  let assemblyFormat = [{
    $allocaType `,` qualified(type($addr)) `,`
    ($dynAllocSize^ `:` type($dynAllocSize) `,`)?
    `[` $name
        oilist( `init`              $init
              | `const`             $constant
              | `cleanup_dest_slot` $cleanup_dest_slot )
    `]`
    ($annotations^)? attr-dict
  }];
```

or adding `comma-separated-oilist` to mlir tablegen?

Andres-Salamanca:
> What about changing the format non-ambiguous parse that does not need lookahead, i.e. dropping commas:
> 
> ```
>   let assemblyFormat = [{
>     $allocaType `,` qualified(type($addr)) `,`
>     ($dynAllocSize^ `:` type($dynAllocSize) `,`)?
>     `[` $name
>         oilist( `init`              $init
>               | `const`             $constant
>               | `cleanup_dest_slot` $cleanup_dest_slot )
>     `]`
>     ($annotations^)? attr-dict
>   }];
> ```
> 
> or adding `comma-separated-oilist` to mlir tablegen?

yes that works, i did it this way because i wanted to maintain the current format and avoid changing all the test cases 😅 since dropping the comma would break them, but if you prefer the oilist approach i can change it.

xlauko:
> > What about changing the format non-ambiguous parse that does not need lookahead, i.e. dropping commas:
> > ```
> >   let assemblyFormat = [{
> >     $allocaType `,` qualified(type($addr)) `,`
> >     ($dynAllocSize^ `:` type($dynAllocSize) `,`)?
> >     `[` $name
> >         oilist( `init`              $init
> >               | `const`             $constant
> >               | `cleanup_dest_slot` $cleanup_dest_slot )
> >     `]`
> >     ($annotations^)? attr-dict
> >   }];
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or adding `comma-separated-oilist` to mlir tablegen?
> 
> yes that works, i did it this way because i wanted to maintain the current format and avoid changing all the test cases 😅 since dropping the comma would break them, but if you prefer the oilist approach i can change it.

Yeah I got it, and if we are to change format we can make it a bit more uniform with other operations. It is a weird that alloca has some attributes in surrounded by `[` `]`. Not sure what would be the best here. @andykaylor @erichkeane would you have a preference or do we want to keep old format and custom printer/parser?

erichkeane:
> > > What about changing the format non-ambiguous parse that does not need lookahead, i.e. dropping commas:
> > > ```
> > >   let assemblyFormat = [{
> > >     $allocaType `,` qualified(type($addr)) `,`
> > >     ($dynAllocSize^ `:` type($dynAllocSize) `,`)?
> > >     `[` $name
> > >         oilist( `init`              $init
> > >               | `const`             $constant
> > >               | `cleanup_dest_slot` $cleanup_dest_slot )
> > >     `]`
> > >     ($annotations^)? attr-dict
> > >   }];
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > or adding `comma-separated-oilist` to mlir tablegen?
> > 
> > 
> > yes that works, i did it this way because i wanted to maintain the current format and avoid changing all the test cases 😅 since dropping the comma would break them, but if you prefer the oilist approach i can change it.
> 
> Yeah I got it, and if we are to change format we can make it a bit more uniform with other operations. It is a weird that alloca has some attributes in surrounded by `[` `]`. Not sure what would be the best here. @andykaylor @erichkeane would you have a preference or do we want to keep old format and custom printer/parser?

I am not attached to the format... I  kinda like the brackets as they stick out nicely, but am not particularly attached.

What I would like MORE is to not have custom parsing/printing ever.  Does MLIR printers NOT have something for a comma delimited list though?  I would expect this is such a common thing that it would be handled...

xlauko:
Agree we want to minimize any custom format, I would be probably for adding comma separated list support to mlir then. I am not that oposed to the format, just wanted to reconsider it if this was touching all alloca places.

Andres-Salamanca:
@erichkeane yeah MLIR doesn't have it natively, the closest is `oilist` but it doesn't support comma separators. the issue is the comma after the name e.g. `["n", init]`  dropping it would break all existing tests. i'm not a big fan of the custom parser either, i just created this PR because i was blocked on another one 😅 if dropping the commas is fine i can go ahead and update all the tests and use `oilist`, or we could also go with what @xlauko suggested and add `comma-separated-oilist` to MLIR.

andykaylor:
I'm not attached to the current format and wouldn't worry too much about having to update all the existing tests (we are probably overchecking details on allocas anyway, so maybe we can generalize them as they get updated). I would really like to see this move towards consistency with other operations, and maybe consistency with `llvm.alloca` too if we can, though I understand we have some CIR-specific properties that will never get passed on to the LLVM dialect.

Andres-Salamanca:
In that case we could keep the brackets as @erichkeane mentioned they stick out nicely, drop the commas between flags and use `oilist` as @xlauko suggested, and also move the alignment out of the attr-dict to make it look closer to `llvm.alloca` as @andykaylor suggested and consistent with `cir.store`:
```mlir
 %0 = cir.alloca !s32i, !cir.ptr<!s32i>, align(4) ["n" init const]
```
does that look good to all of you or do you have other suggestions?

xlauko:
> In that case we could keep the brackets as @erichkeane mentioned they stick out nicely, drop the commas between flags and use `oilist` as @xlauko suggested, and also move the alignment out of the attr-dict to make it look closer to `llvm.alloca` as @andykaylor suggested and consistent with `cir.store`:
> 
> ```mlir
>  %0 = cir.alloca !s32i, !cir.ptr<!s32i>, align(4) ["n" init const]
> ```
> 
> does that look good to all of you or do you have other suggestions?

So to be more uniform, we usually put attributes before operands (see e.g. load/store), types should be represented "funtional type", so you distinguish operand types and return types, i.e. to be uniform with the rest of the dialect:

```mlir
%0 = cir.alloca "n" align(4) init const %size : !s32i -> !cir.ptr<!s32i>
```

% maybe different order of unitattribute alignemnt and name?

this follows convention of other dialects too, e.g.

```mlir 
%0 = llvm.alloca inalloca %size x i32 : (i64) -> !llvm.ptr
```

I understand @erichkeane desire to make attributes more differentiable, but then question is why is alignment is at different place, why only this operation, why square brackets, which we use for "offset-like" information in other places? For newcomers and in general I find more valuable unsuprising choices.

Andres-Salamanca:
yeah that looks better! quick question why did you drop the type of `dynAllocSize`? i noticed `llvm.alloca` uses `x` to  separate it, in our case i think keeping it with a `,` looks good:
`%0 = cir.alloca "n" align(4) init %sz : !u64i, !s32i -> !cir.ptr<!s32i>`
what do you think?

erichkeane:
> I understand @erichkeane desire to make attributes more differentiable, but then question is why is alignment is at different place, why only this operation, why square brackets, which we use for "offset-like" information in other places? For newcomers and in general I find more valuable unsuprising choices.

That is definitely a valid argument (hah!).  TBH, I am probably just attached to the square brackets due to familiarity, but the examples you showed are also definitely more consistent with the rest of the dialect, so we should probably do that.

xlauko:
> yeah that looks better! quick question why did you drop the type of `dynAllocSize`? i noticed `llvm.alloca` uses `x` to separate it, in our case i think keeping it with a `,` looks good: `%0 = cir.alloca "n" align(4) init %sz : !u64i, !s32i -> !cir.ptr<!s32i>` what do you think?

I just forgot about it :D

Andres-Salamanca:
@andykaylor @xlauko @erichkeane

clang/include/clang/CIR/Dialect/IR/CIROps.td
```diff
@@ -653,13 +631,13 @@ def CIR_AllocaOp : CIR_Op<"alloca", [
...
+    oilist( `init`              $init
+          | `const`             $constant
+          | `cleanup_dest_slot` $cleanup_dest_slot)
     ($dynAllocSize^ `:` type($dynAllocSize) `,`)?
-    `[` $name
-       (`,` `init` $init^)?
-       (`,` `const` $constant^)?
-       (`,` `cleanup_dest_slot` $cleanup_dest_slot^)?
-    `]`
+    $allocaType `->` qualified(type($addr))
```
xlauko: hmm, sorry not catching this earlier, isn't allocaType always addr.getElementType? in such case we can infer it and never print it, maybe it is even unnecessery to store it 2 times? just add member like:
```
mlir::Type getAllocaType() { return getAddr().getElementType(): }
```
?

clang/test/CIR/CodeGen/agg-expr-lvalue.c
```diff
@@ -55,8 +55,8 @@ void test_member_arrow_in_array(void) {
 }
 
 // CIR-LABEL: cir.func{{.*}} @test_member_arrow_in_array
-// CIR:   %[[PTR:.*]] = cir.alloca !cir.ptr<!rec_Line>{{.*}}, ["line_ptr"]
-// CIR:   %[[ARR:.*]] = cir.alloca !cir.array<!rec_Point x 1>{{.*}}, ["arr", init]
+// CIR:   %[[PTR:.*]] = cir.alloca "line_ptr" {{.*}} !cir.ptr<!rec_Line>{{.*}}
```
xlauko:
```suggestion
// CIR:   %[[PTR:.*]] = cir.alloca "line_ptr" {{.*}} !cir.ptr<!rec_Line>
```

clang/test/CIR/CodeGen/agg-expr-lvalue.c
```diff
@@ -55,8 +55,8 @@ void test_member_arrow_in_array(void) {
...
+// CIR:   %[[ARR:.*]] = cir.alloca "arr" {{.*}} init !cir.array<!rec_Point x 1>{{.*}}
```
xlauko:
```suggestion
// CIR:   %[[ARR:.*]] = cir.alloca "arr" {{.*}} init !cir.array<!rec_Point x 1>
```

xlauko: this occurs on more places, you might want to regex it away

krzysz00:
(If a comma separated oilist in MLIR tablegen happens, tag me as reviewer)

clang/test/CIR/CodeGen/agg-expr-lvalue.c
Andres-Salamanca: Done

clang/include/clang/CIR/Dialect/IR/CIROps.td
Andres-Salamanca: I personally think it's good to keep it explicit in LLVM IR it works the other way around, the allocated type is shown but returning a pointer is implicit. i think the same applies here, printing the allocated type improves readability since you can immediately see what's being allocated without having to inspect the pointer type, especially with nested types like:
 ```mlir
%0 = cir.alloca "f" align(4) init const !cir.ptr<!cir.func<() -> !s32i>> -> !cir.ptr<!cir.ptr<!cir.func<() -> !s32i>>>
```
but happy to go either way if you and the others feel strongly about dropping it.
@erichkeane @andykaylor

**Comment by erichkeane** [comment]
I'm happy when @xlauko is btw.

File: clang/include/clang/CIR/Dialect/IR/CIROps.td (hunk 1 above)
Comment (xlauko) [reply clang/include/clang/CIR/Dialect/IR/CIROps.td]: But storing allocated type is wastefull, whis way we store the type two times on the operation (allocated type and wrapped in result type pointer).

Though I agree from the format perspective, printing just the allocated type is nicer. But I am not sure you can easily infer return type without storing the allocated type using just the assembly format and traits.
Can you explore this?

In any way, I would be against printing both allocated and return type as they are essentially the same.

Comment (Andres-Salamanca) [reply clang/include/clang/CIR/Dialect/IR/CIROps.td]: I explored this a bit more and found that `InferTypeOpInterface` could be used here. One advantage is that callers would no longer need to explicitly construct the pointer type. For example, instead of:

```cpp
mlir::Type itrTy = cgf.cgm.convertType(cgf.getContext().UnsignedLongLongTy);
auto itrPtrTy = cir::PointerType::get(itrTy);
cir::AllocaOp::create(builder, loc, itrPtrTy, itrTy, "itr", itrAlign);
```

we could just pass the allocated type and let the result type be inferred.

That said, it doesn't actually address the storage redundancy you mentioned. Even with `inferReturnTypes` building `cir::PointerType::get(ctx, adaptor.getAllocaType())`, both the allocated type and the pointer result type still end up stored in the operation.

I also noticed that `llvm.alloca` follows the same approach and stores both `elem_type` and the pointer result type separately.

Following your suggestion, the format would look something like:

`%0 = cir.alloca "n" align(4) init -> !cir.ptr<!s32i>`

and for dynamic allocations:

`%0 = cir.alloca "n" align(4) init %sz : !u64i -> !cir.ptr<!s32i>`

Does that look good to you?😊

xlauko: > I explored this a bit more and found that `InferTypeOpInterface` could be used here. One advantage is that callers would no longer need to explicitly construct the pointer type. For example, instead of:
> 
> ```c++
> mlir::Type itrTy = cgf.cgm.convertType(cgf.getContext().UnsignedLongLongTy);
> auto itrPtrTy = cir::PointerType::get(itrTy);
> cir::AllocaOp::create(builder, loc, itrPtrTy, itrTy, "itr", itrAlign);
> ```
> 
> we could just pass the allocated type and let the result type be inferred.
> 
> That said, it doesn't actually address the storage redundancy you mentioned. Even with `inferReturnTypes` building `cir::PointerType::get(ctx, adaptor.getAllocaType())`, both the allocated type and the pointer result type still end up stored in the operation.
> 
> I also noticed that `llvm.alloca` follows the same approach and stores both `elem_type` and the pointer result type separately.

The difference is that LLVM does not repeat the type, because return type is opaque pointer type, so you need separate element type. Whereas in our case we have the element type as part of the pointer type.

I would suggest:
- keep only return type?
- add extra class method `getElementType`?

> 
> Following your suggestion, the format would look something like:
> 
> `%0 = cir.alloca "n" align(4) init -> !cir.ptr<!s32i>`
> 
> and for dynamic allocations:
> 
> `%0 = cir.alloca "n" align(4) init %sz : !u64i -> !cir.ptr<!s32i>`
> 
> Does that look good to you?😊

I guess this would work, it is not really what we have at other places, as `->` always has some  preceeding arguments. When there are none, we would use only:
```
%0 = cir.alloca "n" align(4) init : !cir.ptr<!s32i>`
```

but this would collide with size in second case :/ 
does size need to be any int type? if it was deterministic i64, we don't need to print it's size anf get something like:

```
%0 = cir.alloca "n" align(4) init size(%sz) : !cir.ptr<!s32i>
```

Andres-Salamanca: i tried a few things to drop the type of `dynAllocSize`:

 - removing `type($dynAllocSize)` from the format gives: `error: type of operand #0, named 'dynAllocSize', is not buildable and a buildable type cannot be inferred`
 - changing to `Optional<UI64>` gives: `error: 'cir.alloca' op operand #0 must be 64-bit unsigned integer, but got '!cir.int<u, 64'`  since CIR uses `!cir.int<u, 64` which is different from the MLIR builtin `i64`
 - changing to `Optional<CIR_IntType>` gives the same buildable type error

 i also confirmed that in practice it's always `!u64i` from looking at the generated CIR. am i missing something here? if not i think we'd need to keep the type in the format:

 `%1 = cir.alloca "n" align(4) init size(%sz : !u64i) : !cir.ptr<!s32i`

xlauko: > i tried a few things to drop the type of `dynAllocSize`:
> 
> * removing `type($dynAllocSize)` from the format gives: `error: type of operand #0, named 'dynAllocSize', is not buildable and a buildable type cannot be inferred`
> * changing to `Optional<UI64>` gives: `error: 'cir.alloca' op operand #0 must be 64-bit unsigned integer, but got '!cir.int<u, 64'`  since CIR uses `!cir.int<u, 64` which is different from the MLIR builtin `i64`
> * changing to `Optional<CIR_IntType>` gives the same buildable type error
> 
> i also confirmed that in practice it's always `!u64i` from looking at the generated CIR. am i missing something here? if not i think we'd need to keep the type in the format:
> 
> `%1 = cir.alloca "n" align(4) init size(%sz : !u64i) : !cir.ptr<!s32i`

You can then use `CIR_UInt64`. For type to be buildable it needs ne be buildable just from context. See `CIR_UInt` in `CIRTypeConstraints.td`.

Andres-Salamanca:
@xlauko

clang/include/clang/CIR/Dialect/IR/CIROps.td
```diff
@@ -567,21 +567,7 @@ def CIR_SyncScopeKind : CIR_I32EnumAttr<"SyncScopeKind", "sync scope kind", [
...
-> : PredOpTrait<summary, CPred<comparator # "(" #
-      !subst("$_self", "$" # lhsArg # ".getType()", transform) #
-             ", $" # rhsArg # ")">
-> {
-  string lhs = lhsArg;
-  string rhs = rhsArg;
-  string transformer = transform;
-}
-
 def CIR_AllocaOp : CIR_Op<"alloca", [
```
xlauko, line 570: Please update examples to new format

Andres-Salamanca: Done.

xlauko: can you add also variants of other cases, otherwise lgtm

xlauko (approved):

PR #198963, opened 2026-05-21
[lldb] Unify ObjectFile and SymbolFile locking on the Module mutex

JDevlieghere:
SymbolFile already treats the Module mutex as its synchronization point: it exposes a GetModuleMutex() helper and takes that lock in its cache-on-first-use accessors. ObjectFile had no such convention, leaving lazy properties like m_type, m_strata, and m_cache_hash to read and write under no lock at all. A Module is shared across Target instances through the global ModuleList, so two threads can enter those accessors concurrently and observe a half-written value.

Mirror SymbolFile's pattern on ObjectFile. Add a GetModuleMutex() helper that asserts the Module is non-null, take it inside the cache-on-first-use accessors, and route the existing module_sp->GetMutex() call sites through it for consistency. SymbolFile's helper delegates to ObjectFile's so the Module dereference and its null check live in one place.

The exception is ObjectFile::m_sections_mutex, which nests the Module mutex inside it because CreateSections builds the local section list and merges into the Module-owned unified list in a single pass. Document its position relative to the Module mutex.

### DIFF
--- a/lldb/include/lldb/Symbol/ObjectFile.h
+++ b/lldb/include/lldb/Symbol/ObjectFile.h
@@ -24,6 +24,7 @@
 #include "lldb/lldb-private.h"
 #include "llvm/Support/Threading.h"
 #include "llvm/Support/VersionTuple.h"
+#include <mutex>
 #include <optional>
 
 namespace lldb_private {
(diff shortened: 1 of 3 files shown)


JDevlieghere:
Closing as this deadlocks on Linux while building the DWARF index.

PR #185094, opened 2026-03-06
[RISCV] Fix musttail with indirect arguments by forwarding incoming pointers

xroche:
When a `musttail` call passes arguments indirectly (fp128 on RV32, i128 on RV32), the backend allocates a stack temporary and hands the callee a pointer. The tail call deallocates the caller's frame, and the pointer dangles.

Fix by forwarding the incoming indirect pointers instead. They point to the caller's caller's frame, which stays valid after the tail call. Forwarded formal parameters reuse the pointer directly; computed values get stored into the incoming buffer first.

The pointers are saved in virtual registers (`CopyToReg`/`CopyFromReg`) rather than SDValues. The SelectionDAG is cleared between basic blocks and musttail calls can appear in non-entry blocks, so storing raw SDValues across BBs is unsound (this was the bug that led to the revert in 501417baa60f). The vreg save only fires when the function has musttail calls; other functions see no codegen change.

Non-musttail tail calls with indirect args are still rejected.

Fixes #185089.

## AI Disclosure

This patch was developed with assistance from Claude (Anthropic). All code has been reviewed and validated by the author.

PR #198965, opened 2026-05-21
[LoongArch] Fix musttail with indirect arguments by forwarding incoming pointers

heiher:
When a `musttail` call passes arguments indirectly (fp128 on LA32, i128 on LA32), the backend allocates a stack temporary and hands the callee a pointer. The tail call deallocates the caller's frame, and the pointer dangles.

Fix by forwarding the incoming indirect pointers instead. They point to the caller's caller's frame, which stays valid after the tail call. Forwarded formal parameters reuse the pointer directly; computed values get stored into the incoming buffer first.

The pointers are saved in virtual registers (`CopyToReg`/`CopyFromReg`) rather than SDValues. The SelectionDAG is cleared between basic blocks and musttail calls can appear in non-entry blocks, so storing raw SDValues across BBs is unsound (this was the bug that led to the revert in 501417baa60f). The vreg save only fires when the function has musttail calls; other functions see no codegen change.

Non-musttail tail calls with indirect args are still rejected.

This uses the same strategy as the corresponding RISC-V fix (#185094).


wangleiat (approved):
LGTM， thanks.

c-rhodes:
sorry but given this is based on the RISC-V fix (which isn't being backported) and there's already been a revert I think it's best to give this more time on main for testing. We're late in the 22 release cycle with 22.1.7 tomorrow so it's best this waits til 23.

c-rhodes:
apologies i meant to comment on the backport PR.

PR #198966, opened 2026-05-21
[CIR] Wide String literals in a larger variable lowering

erichkeane:
This showed up in a test suite: If the size of the variable is larger than the literal, we were asserting since we were only accounting for a single null terminator.  The FE will ensure that we have enough room for the variable + a null terminator, but the size must be larger!

This patch counds on 'trailing-zeroes' to make sure the null terminator is in place.


andykaylor (approved):
lgtm

PR #198968, opened 2026-05-21
[clang-tidy] Avoid strncmp FP in bugprone-not-null-terminated-result

zeyi2:
This commit changes the modeling of `strncmp` and `wcsncmp` arguments in `bugprone-not-null-terminated-result`. The check now allows the comparison length to include the null terminator when that character is still within
the bounds of the object being read. e.g. `strncmp(s, "foo", 4)` , where the fourth character is the terminator in the string literal.

This should be safe given the manpage:

> The strncmp() function compares not more than n characters.  Because strncmp() is designed for comparing strings rather than binary data, characters that appear after a ‘\0’ character are not compared.

AI Usage: Codex is used to write the test cases.
Closes https://github.com/llvm/llvm-project/issues/48144

### DIFF
--- a/clang-tools-extra/clang-tidy/bugprone/NotNullTerminatedResultCheck.cpp
+++ b/clang-tools-extra/clang-tidy/bugprone/NotNullTerminatedResultCheck.cpp
@@ -92,6 +92,40 @@ static unsigned getLength(const Expr *E,
   return 0;
 }
 
+// Returns the number of characters that may be read from \p E when it is used
+// as an argument of an 'ncmp' function.
+static unsigned getNcmpLength(const Expr *E,
+                              const MatchFinder::MatchResult &Result) {
+  if (!E)
+    return 0;
+
+  E = E->IgnoreParenImpCasts();
+
+  // String literals including the null terminator.
+  // e.g. "foo" -> 4.
+  if (const auto *SL = dyn_cast<StringLiteral>(E))
+    return SL->getLength() + 1;
+
+  if (const auto *DRE = dyn_cast<DeclRefExpr>(E))
+    if (const auto *VD = dyn_cast<VarDecl>(DRE->getDecl())) {
+      // Character arrays are readable up to their full array bound.
+      // e.g. char buf[8] = "foo" -> 8.
+      if (const auto *CAT =
+              Result.Context->getAsConstantArrayType(VD->getType()))
+        if (CAT->getElementType()->isAnyCharacterType())
+          return CAT->getSize().getZExtValue();
+
+      // Pointer variables initialized from literals inherit the literal size.
+      // e.g. const char *buf = "foo" -> 4.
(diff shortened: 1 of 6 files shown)


clang-tools-extra/clang-tidy/bugprone/NotNullTerminatedResultCheck.cpp
```diff
@@ -957,25 +991,27 @@ void NotNullTerminatedResultCheck::strerrorSFix(
 void NotNullTerminatedResultCheck::ncmpFix(
     StringRef Name, const MatchFinder::MatchResult &Result) {
   const auto *FunctionExpr = Result.Nodes.getNodeAs<CallExpr>(FunctionExprName);
-  const Expr *FirstArgExpr = FunctionExpr->getArg(0)->IgnoreImpCasts();
-  const Expr *SecondArgExpr = FunctionExpr->getArg(1)->IgnoreImpCasts();
+  const Expr *FirstArgExpr = FunctionExpr->getArg(0)->IgnoreParenImpCasts();
```
zeyi2, line 994: Drive-by modifications here, using `IgnoreParenImpCasts()` would be more robust.

The code below compares the source text of the `strncmp` arguments with the source text of the strlen argument, so both sides should ignore parentheses consistently.

e.g. `strncmp((s), t, strlen(s) + 1)` and `strncmp(s, t, strlen((s)) + 1)`.

clang-tools-extra/test/clang-tidy/checkers/bugprone/not-null-terminated-result-strlen.c
```diff
@@ -62,32 +62,63 @@ void good_strerror_s(int errno) {
...
+int bad_strncmp_4(void) {
+  char buf[8] = "blah";
+  return strncmp(buf, "1234567", 9);
+  // CHECK-MESSAGES: :[[@LINE-1]]:34: warning: comparison length is too long and might lead to a buffer overflow [bugprone-not-null-terminated-result]
+  // CHECK-FIXES: return strncmp(buf, "1234567", 8);
+}
+
+int bad_strncmp_5(char *str9) {
+  // FIXME: This should suggest 4. strncmp stops comparing at the embedded null.
+  return strncmp(str9, "foo\0bar", 9);
```
zeyi2, line 117: Personally, I think the bad fix-its for embedded null characters are acceptable for now (or low-priority). This seems uncommon in real-world code..

A quick regex search on GitHub repos: https://github.com/search?q=%2F%5Cb%28strncmp%7Cwcsncmp%29%5Cs*%5C%28%5B%5E%3B%5D*%28L%29%3F%22%5B%5E%22%5D*%5C%5C0%5B%5E%22%5D*%22%2F+%28language%3AC+OR+language%3AC%2B%2B%29&type=code&p=1

PR #198971, opened 2026-05-21
[libc] Enforce standard identifier validation in hdrgen

kaladron:
Implemented validation for standard identifiers in the hdrgen tool to catch typos and unknown standards. Aggregated standards from all entity types (macros, types, etc.) for validation, addressing a TODO in header.py.

Added llvm_libc_ext, stdc_ext, and llvm_libc_stdfix_ext to the canonical identifiers. Standardised invalid or inconsistently formatted standards in several YAML files.

This enforces the correctness of any provided standards field but does not yet require that every entity has one (many entities still inherit standards from the header level).

### DIFF
--- a/libc/include/math.yaml
+++ b/libc/include/math.yaml
@@ -153,7 +153,7 @@ functions:
     arguments:
       - type: float
   - name: atanhf16
-    standatds:
+    standards:
       - stdc
     return_type: _Float16
     arguments:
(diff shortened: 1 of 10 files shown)


kaladron:
This is a follow-up to #198854 and #198745 that brings enforcement to the YAML parser.

vonosmas (approved):

PR #198979, opened 2026-05-21
[GlobalISel] Add immediates and predicates to `LegalityQuery`

QuantumSegfault:
Expands `LegalityQuery` to include instruction immediates, and (comparison) predicate operands in the query.

This allows namely `G_SEXT_INREG` to be legalized conditionally based on sext width, and for `G_ICMP` and `G_FCMP` to do the same based on its predicate.

Also adds new `LegalityPredicates` helpers accordingly.

### DIFF
--- a/llvm/include/llvm/CodeGen/GlobalISel/LegalizationArtifactCombiner.h
+++ b/llvm/include/llvm/CodeGen/GlobalISel/LegalizationArtifactCombiner.h
@@ -205,11 +205,12 @@ public:
     Register TruncSrc;
     if (mi_match(SrcReg, MRI, m_GTrunc(m_Reg(TruncSrc)))) {
       LLT DstTy = MRI.getType(DstReg);
-      if (isInstUnsupported({TargetOpcode::G_SEXT_INREG, {DstTy}}))
-        return false;
-      LLVM_DEBUG(dbgs() << ".. Combine MI: " << MI);
       LLT SrcTy = MRI.getType(SrcReg);
       uint64_t SizeInBits = SrcTy.getScalarSizeInBits();
+      if (isInstUnsupported(
+              {TargetOpcode::G_SEXT_INREG, {DstTy}, {}, {(int64_t)SizeInBits}}))
+        return false;
+      LLVM_DEBUG(dbgs() << ".. Combine MI: " << MI);
       if (DstTy != MRI.getType(TruncSrc))
         TruncSrc = Builder.buildAnyExtOrTrunc(DstTy, TruncSrc).getReg(0);
       // Elide G_SEXT_INREG if possible. This is similar to eliding G_AND in
--- a/llvm/include/llvm/CodeGen/GlobalISel/LegalizerInfo.h
+++ b/llvm/include/llvm/CodeGen/GlobalISel/LegalizerInfo.h
@@ -20,6 +20,7 @@
 #include "llvm/CodeGen/MachineMemOperand.h"
 #include "llvm/CodeGen/TargetOpcodes.h"
 #include "llvm/CodeGenTypes/LowLevelType.h"
+#include "llvm/IR/InstrTypes.h"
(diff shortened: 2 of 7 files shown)


QuantumSegfault:
I'm not sure what the best way to go about testing `G_ICMP` and `G_FCMP` is. I haven't seen any current code using "custom" legalization for them. However I have seen a handful that opted to do something to the same effect during instruction select. Maybe those should be moved back to the legalizer to provide some good tests?

Or maybe hold off until later, when I add some support for lowering `G_ICMP` and `G_FCMP` (in terms of combinations of other predicates).

QuantumSegfault:
Ping 

@arsenm

nvjle:
> I'm not sure what the best way to go about testing `G_ICMP` and `G_FCMP` is. I haven't seen any current code using "custom" legalization for them. However I have seen a handful that opted to do something to the same effect during instruction select. Maybe those should be moved back to the legalizer to provide some good tests?
> 
> Or maybe hold off until later, when I add some support for lowering `G_ICMP` and `G_FCMP` (in terms of combinations of other predicates).

I'd recommend splitting this patch into two orthogonal parts:
1. `{immIs, immInSet, immIsNot}` support.
2.  `{predicateIs, predicateInSet, predicateIsNot}` support (probably after the first lands).

Although the first is tested by virtue of adding it to RISC-V, it might be good to add actual unit tests in `unittests/CodeGen/GlobalISel` (see `Legalizer*.cpp`). The second will definitely need unit tests. It might also be worth adding a unit test for the `LegalityQuery::print` changes too.

QuantumSegfault:
> I'd recommend splitting this patch into two orthogonal parts:
> 
>     1. `{immIs, immInSet, immIsNot}` support.
> 
>     2. `{predicateIs, predicateInSet, predicateIsNot}` support (probably after the first lands).
> 
> 
> Although the first is tested by virtue of adding it to RISC-V, it might be good to add actual unit tests in `unittests/CodeGen/GlobalISel` (see `Legalizer*.cpp`). The second will definitely need unit tests. It might also be worth adding a unit test for the `LegalityQuery::print` changes too.

Ah, okay. `unittests`...I haven't worked with those before. Thanks!

Yeah, might make sense to split. Depends on how involved predicate support becomes.

But beyond that, does the patch look okay? Thankfully the changes didn't end up being too widespread.

nvjle:
> > I'd recommend splitting this patch into two orthogonal parts:
> > ```
> > 1. `{immIs, immInSet, immIsNot}` support.
> > 
> > 2. `{predicateIs, predicateInSet, predicateIsNot}` support (probably after the first lands).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Although the first is tested by virtue of adding it to RISC-V, it might be good to add actual unit tests in `unittests/CodeGen/GlobalISel` (see `Legalizer*.cpp`). The second will definitely need unit tests. It might also be worth adding a unit test for the `LegalityQuery::print` changes too.
> 
> Ah, okay. `unittests`...I haven't worked with those before. Thanks!
> 
> Yeah, might make sense to split. Depends on how involved predicate support becomes.

As these really are two orthogonal issues, it is in line with the developer policy to split. At the same time, it isn't a huge patch either. 
 
> 
> But beyond that, does the patch look okay? Thankfully the changes didn't end up being too widespread.

I haven't done a detailed review yet, but at least the first part looks reasonable on a quick look-- and I like the idea.

cofibrant:
I'm interested in the effect of this change on memory usage in GlobalISel. Do you have any statistics?

QuantumSegfault:
> I'm interested in the effect of this change on memory usage in GlobalISel. Do you have any statistics?

No. I wouldn't even know a good way of determining that. I'm guessing I'd have to look into the benchmark suite and get it to report and compare max memory usage?

PR #198715, opened 2026-05-20
[DAG] UDIV/SDIV exact nodes support DemandedElts + add test

lbzzz1016:
Resolves https://github.com/llvm/llvm-project/issues/183047

This patch updates isKnownNeverZero to handle DemandedElts for UDIV and SDIV exact nodes.

PR #198980, opened 2026-05-21
[X86] Lower vector 8-bit rotr(x, 1) to pavgb pattern

### LINKED ISSUE #198060: [X86] Missed Optimization: Vector 8-bit `rotr(x, 1)` should be lowered as `pavgb(x, -(x & 1))`

michael1017:
Fixes #198060


llvm/lib/Target/X86/X86ISelLowering.cpp
RKSimon, line 32280: Move this down to here to avoid XOP regressions (and you should be able to remove the !GFNI check)

RKSimon (changes requested):

llvm/lib/Target/X86/X86ISelLowering.cpp
michael1017: Since the new location is behind the code block 
```
  if (!IsROTL) {
    // If the ISD::ROTR amount is constant, we're always better converting to
    // ISD::ROTL.
    if (SDValue NegAmt = DAG.FoldConstantArithmetic(ISD::SUB, DL, VT, {Z, Amt}))
      return DAG.getNode(ISD::ROTL, DL, VT, R, NegAmt);

    // XOP targets always prefers ISD::ROTL.
    if (Subtarget.hasXOP())
      return DAG.getNode(ISD::ROTL, DL, VT, R,
                         DAG.getNode(ISD::SUB, DL, VT, Z, Amt));
  }
```

ROTR with constant amount will be turned into ROTL. If rotate amount is `y`, the new rotate amount will become `EltSizeInBits - y`. 

Since the x we are lowering is 8-bit, we can get:
rotr(x,1) -> rotl(x,7) -> pavgb(x, 0 - (x & 1))

michael1017:
Hi @RKSimon thanks for you feedback, the code is modified accordingly and is ready to review now.

llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -32273,6 +32273,16 @@ static SDValue LowerRotate(SDValue Op, const X86Subtarget &Subtarget,
   if (VT.is256BitVector() && (Subtarget.hasXOP() || !Subtarget.hasAVX2()))
     return splitVectorIntBinary(Op, DAG, DL);
 
+  // rotl(x,7) -> pavgb(x, 0 - (x & 1))
+  if (EltSizeInBits == 8 && IsCstSplat &&
+      CstSplatValue.urem(EltSizeInBits) == 7 && IsROTL) {
+    SDValue One = DAG.getConstant(1, DL, VT);
+    SDValue LSB = DAG.getNode(ISD::AND, DL, VT, R, One);
+    SDValue Zero = DAG.getConstant(0, DL, VT);
+    SDValue Neg = DAG.getNode(ISD::SUB, DL, VT, Zero, LSB);
```
RKSimon:
```suggestion
    SDValue Neg = DAG.getNegative(LSB, DL, VT);
```
(and remove Zero)

llvm/test/CodeGen/X86/vector-fshr-rot-128.ll
```diff
@@ -1998,69 +1998,67 @@ define <16 x i8> @splatconstant_funnnel_v16i8(<16 x i8> %x) nounwind {
...
 ;
 ; AVX512VLBW-LABEL: splatconstant_funnnel_v16i8_by_one:
 ; AVX512VLBW:       # %bb.0:
-; AVX512VLBW-NEXT:    vpsllw $7, %xmm0, %xmm1
-; AVX512VLBW-NEXT:    vpsrlw $1, %xmm0, %xmm0
-; AVX512VLBW-NEXT:    vpternlogd {{.*#+}} xmm0 = xmm0 ^ (m32bcst & (xmm0 ^ xmm1))
+; AVX512VLBW-NEXT:    vpandd {{\.?LCPI[0-9]+_[0-9]+}}(%rip){1to4}, %xmm0, %xmm1
+; AVX512VLBW-NEXT:    vpxor %xmm2, %xmm2, %xmm2
+; AVX512VLBW-NEXT:    vpsubb %xmm1, %xmm2, %xmm1
+; AVX512VLBW-NEXT:    vpavgb %xmm1, %xmm0, %xmm0
```
RKSimon: regression?

llvm/test/CodeGen/X86/vector-fshr-rot-256.ll
```diff
@@ -1756,71 +1758,72 @@ define <32 x i8> @splatconstant_funnnel_v32i8(<32 x i8> %x) nounwind {
...
 ;
 ; AVX512VL-LABEL: splatconstant_funnnel_v32i8_by_one:
 ; AVX512VL:       # %bb.0:
-; AVX512VL-NEXT:    vpsllw $7, %ymm0, %ymm1
-; AVX512VL-NEXT:    vpsrlw $1, %ymm0, %ymm0
-; AVX512VL-NEXT:    vpternlogd {{.*#+}} ymm0 = ymm0 ^ (m32bcst & (ymm0 ^ ymm1))
+; AVX512VL-NEXT:    vpandd {{\.?LCPI[0-9]+_[0-9]+}}(%rip){1to8}, %ymm0, %ymm1
+; AVX512VL-NEXT:    vpxor %xmm2, %xmm2, %xmm2
+; AVX512VL-NEXT:    vpsubb %ymm1, %ymm2, %ymm1
+; AVX512VL-NEXT:    vpavgb %ymm1, %ymm0, %ymm0
```
RKSimon: regression?

RKSimon (changes requested):
Need a bit more investigation to what to do when VPTERNLOG is available

llvm/test/CodeGen/X86/vector-fshr-rot-128.ll
michael1017: Yeah you’re right, this looks like a regression. The old sequence used fewer instructions and fewer registers.

I think this happens because AVX512 can take advantage of `vpternlogd`. So we should probably lower to pavgb only when AVX512 isn’t available.

RKSimon (approved):
LGTM

Andarwinux:
@RKSimon For first-time contributors, workflows need to be approved first to make auto-merge actually work.

also https://github.com/llvm/llvm-project/pull/198715

PR #198982, opened 2026-05-21
[llvm,clang] Don't assume non-erased DenseMap entries remain valid after erase. NFC

MaskRay:
In preparation for switching DenseMap from tombstone deletion to
backward-shift deletion, update call sites that reuse an iterator or a
bucket reference after erasing another entry from the same map.

These work under tombstone deletion because unrelated buckets stay put,
but backward-shift deletion relocates entries to close the gap.

Add DenseMap::remove_if, similar to SmallPtrSet::remove_if, as
replacement for erase-while-iterating, and use it where applicable.

Aided by Claude Opus 4.7


aengelke:
Could an erase() that returns the next iterator be feasible? Similar to std::unordered_map. Or some remove_if()? When removing many elements, having to materialize a list of all of them is less efficient.

teresajohnson:
Is there some background on this change? I.e. runtime or memory improvements. Also, if this change is made https://llvm.org/docs/ProgrammersManual.html#llvm-adt-densemap-h needs to be updated to note that iterators are also invalidated on erasure.

MaskRay:
> Is there some background on this change? I.e. runtime or memory improvements. Also, if this change is made [llvm.org/docs/ProgrammersManual.html#llvm-adt-densemap-h](https://llvm.org/docs/ProgrammersManual.html#llvm-adt-densemap-h) needs to be updated to note that iterators are also invalidated on erasure.

Yes, significant runtime and memory improvement adopting Knuth TAOCP 6.4 Algorithm R for deletion:

[Compile-time tracker comparison](https://llvm-compile-time-tracker.com/compare.php?from=f7f7cd411e63d6ce99b7dd2cfb3e74875b2ed8a3&to=53631eec76eb04ff37f17ac6105bfd719d1085a5&stat=instructions:u)

### `instructions:u` geomean

| Benchmark | Geomean Δ |
|-----------|-----------|
| stage1-O3 | -1.32% |
| stage1-ReleaseThinLTO | -1.43% |
| stage1-ReleaseLTO-g | -1.63% |
| stage1-O0-g | -0.90% |
| stage1-aarch64-O3 | -1.27% |
| stage1-aarch64-O0-g | -0.74% |
| stage2-O3 | -1.25% |
| stage2-O0-g | -1.06% |

### `max-rss` geomean

| Benchmark | Geomean Δ |
|-----------|-----------|
| stage1-O3 | -1.46% |
| stage1-ReleaseThinLTO | -1.22% |
| stage1-ReleaseLTO-g | -1.13% |
| stage1-O0-g | -1.23% |
| stage1-aarch64-O3 | -1.11% |
| stage1-aarch64-O0-g | -1.27% |
| stage2-O3 | -0.01% |
| stage2-O0-g | +0.05% |

aengelke:
> Yes, significant runtime and memory improvement adopting Knuth TAOCP 6.4 Algorithm R for deletion:

The max-rss reduction is only for stage1 but not for stage2 builds. I think this is because of the smaller binary size from the GCC build and not actually related to the algorithmic change.

llvm/include/llvm/ADT/SetOperations.h
```diff
@@ -60,12 +61,12 @@ template <class S1Ty, class S2Ty> void set_intersect(S1Ty &S1, const S2Ty &S2) {
...
-      Next = std::next(I);
-      if (!S2.count(*I))
-        S1.erase(I); // Erase element if not in S2
-    }
+    SmallVector<typename S1Ty::value_type> ToRemove;
+    for (const auto &E : S1)
+      if (!S2.count(E))
+        ToRemove.push_back(E);
+    for (const auto &E : ToRemove)
+      S1.erase(E);
```
aengelke: I'm a bit worried that this might be used in cases where an extra materialization is not wanted.

llvm/lib/Analysis/ScalarEvolution.cpp
```diff
@@ -8749,8 +8749,9 @@ void ScalarEvolution::visitAndClearUsers(
     ValueExprMapType::iterator It =
         ValueExprMap.find_as(static_cast<Value *>(I));
     if (It != ValueExprMap.end()) {
+      SCEVUse Mapped = It->second;
       eraseValueFromMap(It->first);
-      ToForget.push_back(It->second);
+      ToForget.push_back(Mapped);
```
aengelke: Why? Add comment.

aengelke (commented):
LGTM otherwise.

The subsequent PR that bumps the epoch on erase should also update documentation and add a release note.

That said, I'm not particularly keen on the approach to *require* rematerialization+rehashing to erase while iterating... (see comment above). Looks like there's no real in-tree reason, though..

llvm/lib/Analysis/ScalarEvolution.cpp
MaskRay: `eraseValueFromMap(It->first);` invalidates `It`. We need to save the value first.
We should actually move ToForget before. Fixing...

llvm/include/llvm/ADT/SetOperations.h
MaskRay: `if constexpr (HasMemberRemoveIf) branch (L60-61)`  utilizes the added `remove_if`. This chunk has been reverted

aengelke (approved):
LGTM

nikic:
Looks like this causes a compile-time regression: https://llvm-compile-time-tracker.com/compare.php?from=69a5cf515fd317bcf918e48de9137dd8549870c5&to=6302439f5aaea6cb776d8ceb5c2ef9108fccf702&stat=instructions:u

I realize we'll more than recover this in a future change, but it still seems suspicious to me that this causes a regression. It doesn't seem like it should...

aengelke:
Huh, this is unexpected -- especially that this regression primarily hits O0g. Intuitively, I would suspect that the PMDataManager::removeNotPreservedAnalysis lambda is not inlined (legacy PM is comparatively more expensive in O0); but I haven't verified this.

MaskRay:
> Huh, this is unexpected -- especially that this regression primarily hits O0g. Intuitively, I would suspect that the PMDataManager::removeNotPreservedAnalysis lambda is not inlined (legacy PM is comparatively more expensive in O0); but I haven't verified this.

Good call.  In a CT experiment that reverts only `PMDataManager::removeNotPreservedAnalysis` back to the pre-#198982 erase-while-iterating loops, https://llvm-compile-time-tracker.com/compare.php?from=f6e4e71fcdbff98cb060f87695c75f00b43bf918&to=79e5f55c1c2aab9ffcd9d4665314c81daf38f709&stat=instructions:u 

```
 The deltas come back as a near-exact mirror image of the regression:

  ┌─────────────────────┬──────────────────────┬─────────────────────┐
  │       config        │ regression (#198982) │ revert just this fn │
  ├─────────────────────┼──────────────────────┼─────────────────────┤
  │ stage1-O0-g         │ +0.13%               │ −0.13%              │
  ├─────────────────────┼──────────────────────┼─────────────────────┤
  │ stage1-aarch64-O0-g │ +0.16%               │ −0.15%              │
  ├─────────────────────┼──────────────────────┼─────────────────────┤
  │ stage2-O0-g         │ +0.15%               │ −0.16%              │
  ├─────────────────────┼──────────────────────┼─────────────────────┤
  │ stage1-O3           │ +0.04%               │ −0.03%              │
  └─────────────────────┴──────────────────────┴─────────────────────┘
```

So the entire O0-g regression is this one function; the rest of the patch is neutral.

MaskRay:
Confirmed

`Two callers ⇒ the compiler emits DenseMapBase::remove_if<$_0> out-of-line rather than inlining it.`

Created a fix: https://github.com/llvm/llvm-project/pull/199571

PR #198988, opened 2026-05-21
[SPIR][SPIRV] Mark AS3 as having an unstable pointer representation

wenju-he:
Intel GPU backend uses a non-zero sentinel for null in OpenCL local address space. DataLayout::mustNotIntroducePtrToInt returns true when AS3 has unstable pointer representation, preventing optimization passes (e.g. SROA) from creating invalid ptrtoint/inttoptr round-trip:
```
  %1 = ptrtoint ptr addrspace(3) addrspacecast (ptr addrspace(4) null to ptr addrspace(3)) to i64
  %2 = inttoptr i64 %1 to ptr addrspace(3)
  %3 = addrspacecast ptr addrspace(3) %2 to ptr addrspace(4)
```
EarlyCSEPass further folds the sequence to `ptr addrspace(4) null`, which doesn't preserve null in local address space.

`spir-unknown-unknown`, `spir64-unknown-unknown`, `spirv32-unknown-unknown` and `spirv64-unknown-unknown`  are generic virtual triples. Marking AS3 as unstable pointer is a conservative approach to preserve local-address-space null semantics.

This fixes a SYCL test on Intel GPU.

### DIFF
--- a/clang/lib/Basic/Targets/SPIR.h
+++ b/clang/lib/Basic/Targets/SPIR.h
@@ -282,7 +282,8 @@ public:
     // Take the maximum because it's possible the Host supports wider types.
     MaxAtomicInlineWidth = std::max<unsigned char>(MaxAtomicInlineWidth, 64);
     resetDataLayout("e-p:32:32-i64:64-v16:16-v24:32-v32:32-v48:64-"
-                    "v96:128-v192:256-v256:256-v512:512-v1024:1024-G1");
+                    "v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"
+                    "-pu3:32:32");
   }
 
   void getTargetDefines(const LangOptions &Opts,
@@ -302,7 +303,8 @@ public:
     // Take the maximum because it's possible the Host supports wider types.
     MaxAtomicInlineWidth = std::max<unsigned char>(MaxAtomicInlineWidth, 64);
     resetDataLayout("e-i64:64-v16:16-v24:32-v32:32-v48:64-"
-                    "v96:128-v192:256-v256:256-v512:512-v1024:1024-G1");
+                    "v96:128-v192:256-v256:256-v512:512-v1024:1024-G1"
+                    "-pu3:64:64");
   }
 
   void getTargetDefines(const LangOptions &Opts,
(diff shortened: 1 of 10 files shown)


arsenm:
SPIR classic is still in use?

wenju-he:
> SPIR classic is still in use?

yes, our downstream still use it for SYCL: https://github.com/intel/llvm/blob/2d41460b781bcc384e81c731e2f807de5cd83ac2/clang/lib/Driver/Driver.cpp#L1140

Upstream SYCL uses spirv: https://github.com/llvm/llvm-project/blob/25f90289bb0e137d500827486ac19e8e0be781e1/clang/lib/Driver/Driver.cpp#L1041

I locally changed downstream to use spirv64 triple and the test have the same fail. So I applied the change to generic triples `spirv32-unknown-unknown` and `spirv64-unknown-unknown` as well in https://github.com/llvm/llvm-project/pull/198988/commits/b6fb0f1bc4e10de406c0d40558c1042a41ea399c

arsenm:
> Marking AS3 as unstable pointer is a conservative approach to preserve local-address-space null semantics.

I don't follow the problem description. You do not need unstable pointer representation to have non-0 null pointers in a non-0 address space. This has worked for a very long time

wenju-he:
> I don't follow the problem description. You do not need unstable pointer representation to have non-0 null pointers in a non-0 address space. This has worked for a very long time

Agreed, but the existing po/pz infrastructure doesn't help generic targets.

AMDGPU is currently protected by `AS3 is 32bit` and `DL.getTypeSizeInBits(OldTy) != DL.getTypeSizeInBits(NewTy)` in SROA's canConvertValue.
Proably other non-generic targets are protected in the same way.

spirv64 is a generic target. It essentiall has a `unknown null` for AS3, but currently there is no correct DataLayout mechanism today for `unknown null`.
Does it make sense to extend https://github.com/llvm/llvm-project/blob/0f3d6b9cac25c750f5fa477d92ba15b68b03e210/llvm/docs/LangRef.rst#L3435 to add a "unknown null" to prevent folding null across addrspacecast for spirv64?

PR #198992, opened 2026-05-21
[EarlyCSE] CSE with.overflow result extracts with binops

### LINKED ISSUE #197671: Missed optimization: unnecessary comparison with __builtin_add_overflow and duplicate addition on AArch64
When the sum of the first two arguments of __builtin_add_overflow is used instead of its result, Clang emits the same instructions for x64 but an additional comparison on AArch64: https://godbolt.org/z/E1bh1WGM9

jiangning26:
Teach EarlyCSE value numbering that index 0 of integer with.overflow                                                                                                                                                         
intrinsics is equivalent to the corresponding wrapping add/sub/mul                                                                                                                                                           
expression when a matching value is already available.                                                                                                                                                                       
                                                                                                                                                                                                                             
Do not apply the cross-kind equivalence to no-wrap binary operators,                                                                                                                                                         
because those may produce poison while with.overflow returns the                                                                                                                                                             
wrapped arithmetic result.                                                                                                                                                                                                   
                                                                                                                                                                                                                             
Fixes #197671


llvm/lib/Transforms/InstCombine/InstCombineAddSub.cpp
```diff
@@ -853,6 +855,107 @@ static Instruction *foldNoWrapAdd(BinaryOperator &Add,
...
+
+static Instruction *reuseOverflowIntrinsicResult(BinaryOperator &I,
+                                                 InstCombinerImpl &IC) {
+  if (I.use_empty())
+    return nullptr;
+
+  Value *Op0 = I.getOperand(0);
+  Value *Op1 = I.getOperand(1);
+  unsigned Op0Users = countUsersUpTo(Op0, MaxOverflowUsersToScan);
+  unsigned Op1Users = countUsersUpTo(Op1, MaxOverflowUsersToScan);
```
cardigan1008: Hi, there is a related crash case:

```llvm
define i32 @crash(i32 %a) {
entry:
  %sum = add i32 %a, 1
  ret i32 %sum
}
```

Crash:

```sh
opt: llvm/include/llvm/IR/Value.h:361: llvm::Value::const_use_iterator llvm::Value::materialized_use_begin() const: Assertion `hasUseList()' failed.
```

The fix assumed all operands of candidate add/sub instructions can be scanned through their use-lists. That is not true for all `Value` subclasses, especially constants without materialized use-lists. The scan was introduced as a profitability heuristic to locate `WithOverflowInst` users of one of the operands, but it is run before checking whether either operand is a suitable scan root. Existing InstCombine add/sub code commonly handles constants, so this new early fold broadens the crash surface to ordinary arithmetic such as `add i32 %a, 1`.

A robust fix should avoid calling `use_begin()` on values that do not have a materialized use-list / are constants, or otherwise guard the scan to instruction/argument-like values. For constant operands, there cannot be a matching `with.overflow` intrinsic using the same constant as a scanned operand in a useful bounded use-list way, so the helper can pick the non-constant operand or bail out safely.

For more details, please refer to https://archer.top/artifact/run/198992/run.review.md

> Found with [Archer](https://github.com/cuhk-s3/Archer). Please let me know if anything is wrong.

dtcxzyw:
The following correctness issue was found by [llvm-hackme](https://github.com/dtcxzyw/llvm-hackme).






This comment is generated by an automated correctness checking service designed to help identify critical correctness bugs (opt crashes or Alive2 miscompilations) and improve PR review efficiency under limited reviewer bandwidth.

## Status: Previously reported issue no longer reproduces

The correctness bug previously reported by [llvm-hackme](https://github.com/dtcxzyw/llvm-hackme) no longer reproduces with the current PR update, and fuzzing did not identify any new correctness bugs (opt crashes or Alive2 miscompilations).

**Baseline Revision**: `8b1c7ea6f6e91a9977177dd214aca0523f58064e`
**PR Head SHA**: `26bd7444bd6ed667888695a669e3150a72430073`
**Patch SHA256**: `50be66a193d5b017eb2c58d4db734f9f41c95facf893d6c0d870a51b7f83cc67`

dtcxzyw (changes requested):
A correctness issue was identified by [llvm-hackme](https://github.com/dtcxzyw/llvm-hackme). Please see the [issue comment](https://github.com/llvm/llvm-project/pull/198992#issuecomment-4507008928) for details.

llvm/test/Transforms/InstCombine/with-overflow-recompute.ll
```diff
@@ -0,0 +1,448 @@
...
+declare void @use.v4i1(<4 x i1>)
+
+declare { i32, i1 } @llvm.uadd.with.overflow.i32(i32, i32)
+declare { i32, i1 } @llvm.sadd.with.overflow.i32(i32, i32)
+declare { i32, i1 } @llvm.usub.with.overflow.i32(i32, i32)
+declare { i32, i1 } @llvm.ssub.with.overflow.i32(i32, i32)
+declare { i64, i1 } @llvm.uadd.with.overflow.i64(i64, i64)
+declare { <4 x i32>, <4 x i1> } @llvm.uadd.with.overflow.v4i32(<4 x i32>, <4 x i32>)
+
+define i32 @uadd_plain_recompute(i32 %a, i32 %b) {
```
dtcxzyw: It is covered by GVN: https://godbolt.org/z/aeP7vxGd8 https://github.com/llvm/llvm-project/blob/553b6099b20c9f4ebbb19197c5c8b7cfcda565c2/llvm/lib/Transforms/Scalar/GVN.cpp#L402-L411

I'd suggest extending GVN/EarlyCSE to perform the transformation.

jiangning26: Thanks, done. I moved the optimization to EarlyCSE

jiangning26:
Thanks @dtcxzyw and @cardigan1008 for the reviews.
    I have reworked the patch to implement this in EarlyCSE instead of InstCombine, as suggested. The earlier InstCombine crash report should no longer apply to the current version, and the automated correctness check reported that the previously found issue no longer reproduces.
    The current PR head is `7ca3694` and CI is green. I believe the earlier feedback has been addressed by the move to EarlyCSE. Could you please take another look when you have time?

jiangning26:
Hi @dtcxzyw, rebased onto latest main to fix the merge conflict (EarlyCSE sentinel refactor landed meanwhile). Implementation unchanged. CI green, benchmark #406 healthy (early-cse.NumCSE +62, no crashes/miscompiles). 
Could you take another look? Thanks!

Issue #131365, opened 2025-03-14
[libc++] Decouple the notion of providing a threading API and having multiple threads of execution


ldionne:
Today, `LIBCXX_ENABLE_THREADS` controls both:
- whether libc++ provides a threading API, and
- whether libc++ assumes the platform to allow multiple threads of execution

However, some systems exist where we can't (or don't want to) provide the libc++ threading API, yet the platform does support multiple threads of executions. In the current model, we are forced to pick between:
- `LIBCXX_ENABLE_THREADS=ON`, which causes libc++ to attempt to provide a threading API and fails to build
- `LIBCXX_ENABLE_THREADS=OFF`, which causes libc++ to assume that there can't be multiple threads of executions, and results in the code not using synchronization primitives, atomics and other thread-safe things. This can result in race conditions on such platforms.

Both options are clearly bad, and we should fix that.

efriedma-quic:
A list of impacted APIs from a quick look at various bits of code in libc++: std::call_once, locale::id::__get, tzdb_list_private.h, std::shared_ptr, std::random_shuffle, std::set_new_handler, std::set_terminate, std::set_unexpected.  And in libc++abi: cxa_guard_impl.h, cxa_handlers.cpp, cxa_exception.cpp, cxa_default_handlers.cpp, fallback_malloc.cpp.

Some of the checks are direct, some of them come indirectly from libcxx/src/include/atomic_support.h.

PR #198994, opened 2026-05-21
[libc++] Remove ios_base::__xindex_ from the ABI

philnik777:
`__xindex_` is only ever used from the dylib from a single function. We can simplify the code a bit by making the variable function-local and avoiding exposing it to the ABI at all. This also fixes a TODO about whether it's safe to use `atomic` with the GCC ABI: yes, since it's not actually part of our ABI.

### DIFF
--- a/libcxx/src/ios.cpp
+++ b/libcxx/src/ios.cpp
@@ -9,6 +9,7 @@
 #include <__config>
 #include <__locale>
 #include <algorithm>
+#include <atomic>
 #include <ios>
 #include <limits>
 #include <memory>
@@ -111,13 +112,6 @@ locale ios_base::imbue(const locale& newloc) {
 
 locale ios_base::getloc() const { return __loc_; }
 
-// xalloc
-#if _LIBCPP_HAS_C_ATOMIC_IMP && _LIBCPP_HAS_THREADS
-atomic<int> ios_base::__xindex_{0};
-#else
-int ios_base::__xindex_ = 0;
-#endif
-
 template <typename _Tp>
 static size_t __ios_new_cap(size_t __req_size, size_t __current_cap) { // Precondition: __req_size > __current_cap
   const size_t mx = std::numeric_limits<size_t>::max() / sizeof(_Tp);
@@ -127,7 +121,10 @@ static size_t __ios_new_cap(size_t __req_size, size_t __current_cap) { // Precon
     return mx;
 }
 
-int ios_base::xalloc() { return __xindex_++; }
+int ios_base::xalloc() {
+  constinit static atomic<int> xindex = 0;
+  return xindex++;
+}
 
 long& ios_base::iword(int index) {
   size_t req_size = static_cast<size_t>(index) + 1;
--- a/libcxx/include/ios
+++ b/libcxx/include/ios
@@ -235,10 +235,6 @@ storage-class-specifier const error_category& iostream_category() noexcept;
 #    include <__verbose_abort>
 #    include <version>
 
-#    if _LIBCPP_HAS_ATOMIC_HEADER
-#      include <__atomic/atomic.h> // for __xindex_
-#    endif
-
 #    if !defined(_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER)
 #      pragma GCC system_header
 #    endif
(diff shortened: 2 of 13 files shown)


libcxx/src/ios.cpp
```diff
@@ -9,6 +9,7 @@
 #include <__config>
```
ldionne, line 9: You're missing an entry in the ABI changelog.

ldionne (approved):
LGTM: I don't see how this can be depended on by a conforming program. This reduces our ABI surface and simplifies the code.

vrukesh:
@philnik777 

This patch potentially causes regression in bare-metal Arm-toolchain for Armv4T/Armv5TE/Armv6-M, by requiring unresolved __atomic_fetch_add_4.

Godbolt compiler explorer reproducer: https://godbolt.org/z/o1o677qKo

Github action: https://github.com/arm/arm-toolchain/actions/runs/27122992388

File path: test-results-build.sh-Ubuntu-22.04-x86_64/multilib-picolibc-builds/cxxlibs/armv5te_exn_rtti_size/build/libcxxabi/test/lit_results.junit.xml

Log snippet for your reference:
`
. - - -command stderr- - - - - -
| ld.lld: error: undefined symbol: __atomic_fetch_add_4
| >>> referenced by ios.cpp
| >>>               ios.cpp.obj:(std::__2::ios_base::xalloc()) in archive /home/runner/work/arm-toolchain/arm-toolchain/build/multilib-picolibc-builds/cxxlibs/armv5te_exn_rtti_size/build/libcxxabi/test-suite-install/lib/libc++.a
| clang++: error: linker command failed with exit code 1 (use -v to see invocation)

`

philnik777:
@vrukesh Are you saying your platform doesn't have native 4 byte atomics?

vrukesh:
@philnik777 

Yes ! Just for reference, we observe issue with library variants:

armv4t_exn_rtti_size
armv4t_size
armv5te_exn_rtti_size
armv5te_size
armv6m_soft_nofp_exn_rtti_size
armv6m_soft_nofp_size
armebv6m_soft_nofp_exn_rtti_size
armebv6m_soft_nofp_size

philnik777:
@vrukesh Do you possibly have downstream changes? I don't think using `atomic<int>` is new here, it's simply not guarded anymore behind a flag that was always true. FWIW I'd also suggest that maybe you should provide libatomic.

smithp35:
Just to give some context. On an armv6-m, like Cortex-M0 there are no atomic instructions so all calls will go out to libatomic. At least one issue where this has been mentioned https://github.com/llvm/llvm-project/issues/58603 

Unfortunately there isn't an easy way to implement libatomic on a v6-m in a way that would transparently satisfy all users. The most common implementation is to disable interrupts, do the operation, re-enable interrupts. However disabling interrupts requires privileged (handler mode I think) so it isn't universally applicable. The other possible implementation is to just do the operation without touching interrupts, but that may not be the desired behaviour.

The end result is an undefined link error for the missing libatomic functions.

A lot of embedded toolchains prefer to build without atomic<int> for v6-m to avoid running into this problem.

ldionne:
I think what @philnik777 's suggesting here is that something else must have changed too. Was `_LIBCPP_HAS_C_ATOMIC_IMP && _LIBCPP_HAS_THREADS` somehow false previously? Otherwise, you were already using `atomic<int>` in the code.

voltur01:
In ATfE we are building in single threaded mode (at least for now) so apparently `_LIBCPP_HAS_THREADS` was false, thus there was no use of atomic.

ldionne:
Ok, this actually brings up a really interesting question. First, just to reframe the discussion, `_LIBCPP_HAS_THREADS` doesn't mean "single threaded mode": it means that thread facilities are not available. That's different because "single threaded mode" implies that we could drop synchronization calls. This is actually not something that's very obvious from the documentation, and in fact it's so non-obvious that we don't even follow that consistently within libc++ itself. For example, libc++abi drops synchronization in `libcxxabi/src/cxa_guard_impl.h` when `_LIBCXXABI_HAS_NO_THREADS`, and that's incorrect because there could still be multiple threads of executions even if we don't ourselves provide the APIs to create them.

This is worth talking about because it seems that your use case is closer to wanting a "single threaded mode", where we assume there's only ever the current thread of execution. We don't provide that right now and I don't know if that even makes sense, but it's a possibility.

Another possibility, and that's perhaps the best path forward here since it doesn't require making a tough decision in libc++: we already provide an extension point in `libcxx/include/__atomic/support.h`. Perhaps we could have (or you could maintain downstream) a version that makes atomic non-atomic, basically. That would be like a no-op backend for libc++'s atomic facilities.

I think that would solve your immediate problem, but IDK if it would introduce new ones?

voltur01:
Thank you for the clarification! Yes, for now we build for single threaded, however support for multithreading is on the to-do list, so avoiding threading would not be a solution long term.

To solve the immediate problem, we are providing a fallback libatomic implementation for Armv4/5/6 which effectively ignores atomicity, but gives users a chance to override as it fits their needs.

efriedma-quic:
Opened #208356 to restore the old behavior with `LIBCXX_ENABLE_THREADS=OFF`.

Bug #131365 describes the issue that `LIBCXX_ENABLE_THREADS` doesn't mean what we want it to mean.

PR #198996, opened 2026-05-21
[lldb] Add missing includes.

da-viper:
Failed to build in CI because, ScriptedPythonInterface::CreatePlugingObject is a template function and the arguments are of incomplete types gotten from `lldb-forward.h` (typedef of lldb_private::XXXX = XXXXSP).

Introduced in commit 1b4a578a9f7760a00bf26525a603be1ec6e7d862


medismailben:
Thanks for fixing this @da-viper!

PR #198997, opened 2026-05-21
[flang][AddAliasTags] Fix segfault when type contains `fir.boxproc`

ergawy:
`fir.boxproc` currently has no LLVM representation (its converter returns `std::nullopt`). When `AddAliasTags` called `getTypeSizeAndAlignment` on a type containing `fir.boxproc` (e.g. a sequence of a derived type with procedure pointer components), `convertRecordType` and `convertSequenceType` would crash trying to mlir::cast a null type.

For any type that might recursively contain a non-convertible type (`fir.boxproc` in this case), `TypeConverter` would now propagate an empty optional `mlir::Type` and emit a debug warning that conversion failed. This helps us avoid seg faulting expecting that the type or some part of it were converted correctly.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/flang/lib/Optimizer/Transforms/AddAliasTags.cpp
+++ b/flang/lib/Optimizer/Transforms/AddAliasTags.cpp
@@ -65,10 +65,12 @@ extern llvm::cl::opt<bool> supportCrayPointers;
 
 namespace {
 
-// Return the size and alignment (in bytes) for the given type.
+// Return the size and alignment (in bytes) for the given type, or std::nullopt
+// if the type cannot be converted.
+//
 // TODO: this must be combined with DebugTypeGenerator::getFieldSizeAndAlign().
 // We'd better move fir::LLVMTypeConverter out of the FIRCodeGen component.
-static std::pair<std::uint64_t, unsigned short>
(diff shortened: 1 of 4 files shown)


flang/lib/Optimizer/Transforms/AddAliasTags.cpp
```diff
@@ -237,9 +245,10 @@ class PassState {
   // Return the byte size of the given declaration.
```
jeanPerier: I think I would rather return an optional here so that you can skip it with a AddAliasTag DBG message that says that TBAA is skipped because variable size could not be computed. It will make debugging of missing TBAA easier.

jeanPerier (approved):
Thanks, small nit/change, looks great otherwise.

flang/lib/Optimizer/Transforms/AddAliasTags.cpp
ergawy: Done, thanks for the suggestion.

tblah (approved):
LGTM, thanks

PR #198999, opened 2026-05-21
[AArch64] Lower scalable i64 CLMUL with SVE2/SME

MDevereau:
When AES or SSVE-AES are not available, but SVE2 or SME are, clmul.nxv2i64 can benefit from a cross-byte CLMUL of .S precision. This re-uses the functionality added for nxv8i16.


paulwalker-arm (approved):
I suspect we can shave an instruction off because the `eorbt(trn1())` could be just `eor`. That will complicate the lowering and because I think we might want to remove the multi-instruction isel for the existing legal cases I'm happy to not worry about it now.

PR #195658, opened 2026-05-04
[LangRef] Do not allow free via synchronization in nofree

nikic:
The nofree attribute is currently specified to only forbid direct free calls inside the function. A nofree function is still allowed to compel a pointer to be freed by a different thread through synchronization.

This is currently only spelled out for the function-level nofree attribute, but I assume the same semantics also hold for argument nofree (and this matches how the Attributor implementation infers it).

The original motivation for this definition was to keep the attributes orthogonal and independently inferable. However, the problem is that nosync is a too strong condition: It excludes *any* synchronization, not just synchronization that results in the free of a pointer.

Some frontends like Rust can guarantee that most pointer arguments cannot be freed for the duration of a function call, including via synchronization. However, they cannot guarantee that no synchronization takes place at all. The current definition of nofree makes this inexpressible in LLVM IR, which blocks further progress on dereferenceable-at-point semantics.

This PR makes two changes to the nofree spec:

 * Function-level nofree now disallows free via synchronization, specified in terms of happens-before.
 * Argument-level nofree is now specified in terms of provenance: The argument has derived provenance that makes it UB to free during the execution of the function (which includes freeing on another thread with appropriate synchronization). This also clarified that argument-level nofree only applies to the particular pointer (similar to `captures` etc). The underlying object may still be freed via a different argument to which the same pointer is passed. (I'm open to changing this, but that would essentially limit inference to noalias arguments.)

Note: This PR doesn't include the necessary inference changes to comply with the new semantics yet. I'd like to make sure we're aligned on the direction first.

PR #194391, opened 2026-04-27
[IR] Remove volatile from nosync

nikic:
Volatile operations are explicitly specified as not synchronizing...

> This is not Java’s “volatile” and has no cross-thread synchronization behavior.

... and LLVM does not model them as being synchronizing anywhere, except the definition of this attribute, which is largely unused outside the Attributor.

The ordering requirements of volatile operations are already fully encoded in their memory effects (unlike what is the case for stronger-than-monotonic atomics).

Clarify that "nosync" is specifically in the sense of "synchronizes-with" (rather than just any cross-thread communication) and remove volatile operations from the definition.

I checked the [original RFC](https://discourse.llvm.org/t/rfc-a-nofree-and-nosynch-function-attribute-mixing-dereferenceable-and-delete/49084) for this attribute, and the inclusion of volatile operations is just stated there, but never explicitly discussed.

PR #199002, opened 2026-05-21
[ValueTracking] Conservative NoSync check prevents vectorization

### LINKED ISSUE #180180: Vectorization fails as soon as memory is allocated in the same function
```
#include <algorithm>
#include <vector>

int* ptr;

auto test1(std::vector<int>& v, int n) {
  ptr = new int;
  return std::find_if(v.begin(), v.end(), [n](int v) { return v < n; });
}

auto test2(std::vector<int>& v, int n) {
  return std::find_if(v.begin(), v.end(), [n](int v) { return v < n; });
}
```

should generate the same code for both functions except for the additional allocation in `test1`. However, one seems to be unrolled, while the other gets vectorized.

https://godbolt.org/z/j47ehxef9

CC @fhahn

kshitijvp:
Conservative nosync check in ValueTracking.cpp returns false causing potentially
faulting load preventing vectorization. Instead check if any instructions between
Assume Instruction and Ctx Instruction are synchronizing.
Fixes #180180


kshitijvp:
I will add the cleaned IR Tests soon.
Sorry for the delay.

fhahn (commented):
Is there a reason to re-posting this as a new PR again instead of just updating the existing one? It is very hard to keep track of the comments

kshitijvp:
> Is there a reason to re-posting this as a new PR again instead of just updating the existing one? It is very hard to keep track of the comments

There was some problem with the previous PR and Nikic wasn't able to add reviews so had to open a new one.

kshitijvp:
Since Nikic pointed out, instead of isOrderedAtomic I have used I.maySynchronize(), and have also used the same name since synchronization is just another way of freeing.

kshitijvp:
I have added the cleaned IR Tests.
Sorry for the inconvenience caused.

llvm/lib/Analysis/ValueTracking.cpp
```diff
@@ -708,18 +708,23 @@ bool llvm::willNotFreeBetween(const Instruction *Assume,
     for (const auto &[Idx, I] : enumerate(Range)) {
       if (Idx > MaxInstrsToCheckForFree)
         return false;
-      if (const auto *CB = dyn_cast<CallBase>(&I))
-        if (!CB->hasFnAttr(Attribute::NoFree))
+      if (I.isVolatile())
```
nikic: This check can be dropped.

kshitijvp: If that is dropped the test @volatileNotVectorizingTest that I have added is getting vectorized too.

kshitijvp: Any reason why this check should be dropped?

nikic: We clarified that volatile is not synchronizing in https://github.com/llvm/llvm-project/pull/194391.

kshitijvp: Done.

llvm/include/llvm/Analysis/ValueTracking.h
```diff
@@ -639,7 +639,7 @@ inline bool isValidAssumeForContext(const Instruction *I,
 }
 
 /// Returns true, if no instruction between \p Assume and \p CtxI may free
-/// memory and the function is marked as NoSync. The latter ensures the current
+/// or synchronize memory. The latter ensures the current
```
nikic:
```suggestion
/// (including through synchronization). The latter ensures the current
```

llvm/lib/Analysis/ValueTracking.cpp
```diff
@@ -708,18 +708,21 @@ bool llvm::willNotFreeBetween(const Instruction *Assume,
...
+      if (I.maySynchronize())
+        return false;
+
+      auto *CB = dyn_cast<CallBase>(&I);
+      if (CB) {
+        // Non call site cases covered by the two checks above
+        if (!CB->hasFnAttr(Attribute::NoSync) ||
+            !CB->hasFnAttr(Attribute::NoFree))
           return false;
+      }
```
nikic:
```suggestion
      if (auto *CB = dyn_cast<CallBase>(&I))
        return CB->hasFnAttr(Attribute::NoFree));
      
      if (I.maySynchronize())
        return false;
```

Per the clarification in https://github.com/llvm/llvm-project/pull/195658, for calls, just checking nofree is sufficient (no need to check nosync as well).

nikic (commented):
Sorry, one more semantics change... otherwise this looks good to me.

llvm/lib/Analysis/ValueTracking.cpp
antoniofrighetto: Slightly confused, shouldn't we have something along the following, or a later non-nofree call would be ignored (and ignore nosync for calls)?
```cpp
if (auto *CB = dyn_cast<CallBase>(&I)) {
  if (!CB->hasFnAttr(Attribute::NoFree))
    return false;
} else if (I.maySynchronize()) {
  return false;
}
```

nikic: Yeah, you're right.

kshitijvp: Done.

llvm/include/llvm/Analysis/ValueTracking.h
kshitijvp: Done.

nikic (approved):
LGTM

llvm/include/llvm/Analysis/ValueTracking.h
```diff
@@ -639,7 +639,7 @@ inline bool isValidAssumeForContext(const Instruction *I,
 }
 
 /// Returns true, if no instruction between \p Assume and \p CtxI may free
-/// memory and the function is marked as NoSync. The latter ensures the current
+/// (including through synchronization). The latter ensures the current
 /// function cannot arrange for another thread to free on its behalf.
```
antoniofrighetto:
```suggestion
/// (including through synchronization).
```

llvm/lib/Analysis/ValueTracking.cpp
```diff
@@ -708,18 +708,16 @@ bool llvm::willNotFreeBetween(const Instruction *Assume,
     for (const auto &[Idx, I] : enumerate(Range)) {
       if (Idx > MaxInstrsToCheckForFree)
         return false;
-      if (const auto *CB = dyn_cast<CallBase>(&I))
+
+      if (auto *CB = dyn_cast<CallBase>(&I)) {
```
antoniofrighetto, line 712: Maybe s/hasNoFreeCalls/hasNoFreeInRange/ for the lambda function given we now check non-calls too?

antoniofrighetto (approved):
LGTM

llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-constant-size.ll
```diff
@@ -1456,4 +1456,364 @@ exit:
...
+entry:
+  %l.v = load ptr, ptr %v, align 8
+  call void @llvm.assume(i1 true) [ "align"(ptr %l.v, i64 4) ]
+  %pi.l.v = ptrtoint ptr %l.v to i64
+  %gep.v = getelementptr inbounds nuw i8, ptr %v, i64 8
+  %l.gep.v = load ptr, ptr %gep.v, align 8
+  call void @llvm.assume(i1 true) [ "align"(ptr %l.gep.v, i64 4) ]
+  %pi.l.gep.v = ptrtoint ptr %l.gep.v to i64
+  %sub = sub i64 %pi.l.gep.v, %pi.l.v
+  call void @llvm.assume(i1 true) [ "dereferenceable"(ptr %l.v, i64 %sub) ]
```
fhahn: The file focuses on testing deref assumptions with constant sizes. With non-constant expressions should go to  `llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-variable-size.ll`

But the interesting aspect of the new tests are the nosync etc checks. so I think you can just remove the `ptrtoint/sub` code to compute the deref size. Similarly, you can just pass `l.v` and `l.gep.v` as pointer arguments without loss of generality.

llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-constant-size.ll
```diff
@@ -1456,4 +1456,364 @@ exit:
...
+
+loop.header:
+  %merge = phi ptr [ %gep.merge, %loop.latch ], [ %l.v, %loop.preheader ]
+  %val = load i32, ptr %merge, align 4
+  %c.2 = icmp slt i32 %val, %n
+  br i1 %c.2, label %loop.early.exit, label %loop.latch
+
+loop.early.exit:
+  %iv = phi ptr [ %merge, %loop.header ]
+  %pi.iv = ptrtoint ptr %iv to i64
```
fhahn: can just return `ptr` from the function

llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-constant-size.ll
```diff
@@ -1456,4 +1456,364 @@ exit:
...
+  call void @llvm.assume(i1 true) [ "align"(ptr %l.v, i64 4) ]
+  call void @llvm.assume(i1 true) [ "align"(ptr %l.gep.v, i64 4) ]
+  %c.1 = icmp eq ptr %l.v, %l.gep.v
+  br i1 %c.1, label %exit, label %loop.preheader
+
+loop.preheader:
+  br label %loop.header
+
+loop.header:
+  %merge = phi ptr [ %gep.merge, %loop.latch ], [ %l.v, %loop.preheader ]
```
fhahn:
```suggestion
  %iv = phi ptr [ %gep.merge, %loop.latch ], [ %l.v, %loop.preheader ]
```

llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-constant-size.ll
```diff
@@ -1456,4 +1456,364 @@ exit:
...
+  call void @llvm.assume(i1 true) [ "align"(ptr %l.v, i64 4) ]
+  %pi.l.v = ptrtoint ptr %l.v to i64
+  %gep.v = getelementptr inbounds nuw i8, ptr %v, i64 8
+  %l.gep.v = load ptr, ptr %gep.v, align 8
+  call void @llvm.assume(i1 true) [ "align"(ptr %l.gep.v, i64 4) ]
+  %pi.l.gep.v = ptrtoint ptr %l.gep.v to i64
+  %sub = sub i64 %pi.l.gep.v, %pi.l.v
+  call void @llvm.assume(i1 true) [ "dereferenceable"(ptr %l.v, i64 %sub) ]
+  call void @llvm.assume(i1 true) [ "align"(ptr %l.v, i64 4) ]
+  call void @llvm.assume(i1 true) [ "align"(ptr %l.gep.v, i64 4) ]
```
fhahn: do we need the assumes twice?

llvm/include/llvm/Analysis/ValueTracking.h
kshitijvp: Done.

llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-constant-size.ll
kshitijvp: On removing those ptrtoint/sub the Tests which should be vectorized are not getting vectorized.
I will just move the tests to 
`llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-variable-size.ll`

llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-constant-size.ll
kshitijvp: Removed.

llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-constant-size.ll
kshitijvp: Done.

llvm/lib/Analysis/ValueTracking.cpp
kshitijvp: Done.

kshitijvp:
Why is this Test Failing?
I have updated the Test using update_test_checks.py after building the updated code.

nikic:
@kshitijvp Tests run against main, it's possible that something in the meantime affected the test output. You can try whether merging main reproduces the issue for you.

kshitijvp:
@fhahn Can I merge?

llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-constant-size.ll
fhahn: Right, but I think it should be possible to take one of the existing simpler tests that do not get vectorized from `llvm/test/Transforms/LoopVectorize/dereferenceable-info-from-assumption-constant-size.ll` like the changed `deref_assumption_in_header_constant_trip_count_nofree_via_context_b` and use that to test the main logic skipping with various instructions before the assumption? That way the entry block would be a single deref assumption  + the instruction we are testing.

The existing positive test with the early exit could go `llvm/test/Transforms/LoopVectorize/single-early-exit-deref-assumptions.ll`, I forgot that there was a dedicated file for reasoning about early exits

kshitijvp: @fhahn Isn't `deref_assumption_in_header_constant_trip_count_nofree_via_context_but_missing_nosync` getting vectorized too?

The Check-Lines contain the Vector-Preheader and Vector-Body which means it is getting vectorized.

kshitijvp: I attempted rewrote it to not get vectorized by creating a new point `%a.1` which loads from `%a`. 
And an atomic instruction that works on `%a` next to the dereferenceable instruction. The Loop Ends up using that %a.

However still that is getting vectorized.

kshitijvp: All of the tests within that file are getting vectorized.

fhahn: Yep, what I meant is take `deref_assumption_in_header_constant_trip_count_nofree_via_context_but_missing_nosync` and add the volatile loads etc

kshitijvp:
```
define void @atomicNotVectorizingTest(ptr noalias noundef %a, ptr noalias %b, ptr noalias %c) {
; CHECK-LABEL: define void @atomicNotVectorizingTest(
; CHECK-SAME: ptr noalias noundef [[A:%.*]], ptr noalias [[B:%.*]], ptr noalias [[C:%.*]]) {
; CHECK-NEXT:  [[ENTRY:.*:]]
; CHECK-NEXT:    [[A_1:%.*]] = load ptr, ptr [[A]], align 8
; CHECK-NEXT:    call void @llvm.assume(i1 true) [ "align"(ptr [[A_1]], i64 4), "dereferenceable"(ptr [[A_1]], i64 4000) ]
; CHECK-NEXT:    [[ATOMIC:%.*]] = load atomic ptr, ptr [[A]] seq_cst, align 8
; CHECK-NEXT:    br label %[[VECTOR_PH:.*]]
; CHECK:       [[VECTOR_PH]]:
; CHECK-NEXT:    br label %[[PRED_LOAD_IF:.*]]
; CHECK:       [[PRED_LOAD_IF]]:
; CHECK-NEXT:    [[INDEX:%.*]] = phi i64 [ 0, %[[VECTOR_PH]] ], [ [[INDEX_NEXT:%.*]], %[[PRED_LOAD_IF5:.*]] ]
; CHECK-NEXT:    [[TMP5:%.*]] = getelementptr inbounds i32, ptr [[B]], i64 [[INDEX]]
; CHECK-NEXT:    [[WIDE_LOAD:%.*]] = load <2 x i32>, ptr [[TMP5]], align 4
; CHECK-NEXT:    [[TMP4:%.*]] = icmp slt <2 x i32> [[WIDE_LOAD]], zeroinitializer
; CHECK-NEXT:    [[TMP10:%.*]] = extractelement <2 x i1> [[TMP4]], i64 0
; CHECK-NEXT:    br i1 [[TMP10]], label %[[PRED_LOAD_IF1:.*]], label %[[PRED_LOAD_CONTINUE:.*]]
; CHECK:       [[PRED_LOAD_IF1]]:
; CHECK-NEXT:    [[TMP6:%.*]] = getelementptr i32, ptr [[A_1]], i64 [[INDEX]]
; CHECK-NEXT:    [[TMP7:%.*]] = load i32, ptr [[TMP6]], align 4
; CHECK-NEXT:    [[TMP8:%.*]] = insertelement <2 x i32> poison, i32 [[TMP7]], i64 0
; CHECK-NEXT:    br label %[[PRED_LOAD_CONTINUE]]
; CHECK:       [[PRED_LOAD_CONTINUE]]:
; CHECK-NEXT:    [[TMP9:%.*]] = phi <2 x i32> [ poison, %[[PRED_LOAD_IF]] ], [ [[TMP8]], %[[PRED_LOAD_IF1]] ]
; CHECK-NEXT:    [[C_1:%.*]] = extractelement <2 x i1> [[TMP4]], i64 1
; CHECK-NEXT:    br i1 [[C_1]], label %[[PRED_LOAD_CONTINUE6:.*]], label %[[PRED_LOAD_IF5]]
; CHECK:       [[PRED_LOAD_CONTINUE6]]:
; CHECK-NEXT:    [[TMP13:%.*]] = add i64 [[INDEX]], 1
; CHECK-NEXT:    [[TMP11:%.*]] = getelementptr i32, ptr [[A_1]], i64 [[TMP13]]
; CHECK-NEXT:    [[TMP12:%.*]] = load i32, ptr [[TMP11]], align 4
; CHECK-NEXT:    [[TMP14:%.*]] = insertelement <2 x i32> [[TMP9]], i32 [[TMP12]], i64 1
; CHECK-NEXT:    br label %[[PRED_LOAD_IF5]]
; CHECK:       [[PRED_LOAD_IF5]]:
; CHECK-NEXT:    [[TMP16:%.*]] = phi <2 x i32> [ [[TMP9]], %[[PRED_LOAD_CONTINUE]] ], [ [[TMP14]], %[[PRED_LOAD_CONTINUE6]] ]
; CHECK-NEXT:    [[PREDPHI:%.*]] = select <2 x i1> [[TMP4]], <2 x i32> [[TMP16]], <2 x i32> [[WIDE_LOAD]]
; CHECK-NEXT:    [[TMP15:%.*]] = getelementptr inbounds i32, ptr [[C]], i64 [[INDEX]]
; CHECK-NEXT:    store <2 x i32> [[PREDPHI]], ptr [[TMP15]], align 4
; CHECK-NEXT:    [[INDEX_NEXT]] = add nuw i64 [[INDEX]], 2
; CHECK-NEXT:    [[TMP17:%.*]] = icmp eq i64 [[INDEX_NEXT]], 1000
; CHECK-NEXT:    br i1 [[TMP17]], label %[[MIDDLE_BLOCK:.*]], label %[[PRED_LOAD_IF]], !llvm.loop [[LOOP24:![0-9]+]]
; CHECK:       [[MIDDLE_BLOCK]]:
; CHECK-NEXT:    br label %[[EXIT:.*]]
; CHECK:       [[EXIT]]:
; CHECK-NEXT:    ret void
;
entry:
  %a.1 = load ptr, ptr %a, align 8
  call void @llvm.assume(i1 true) [ "align"(ptr %a.1, i64 4), "dereferenceable"(ptr %a.1, i64 4000) ]
  %atomic = load atomic ptr, ptr %a seq_cst, align 8                      ; Atomic Instruction
  br label %loop.header

loop.header:
  %iv = phi i64 [ 0, %entry ], [ %iv.next, %loop.latch ]
  %gep.a = getelementptr i32, ptr %a.1, i64 %iv
  %gep.b = getelementptr inbounds i32, ptr %b, i64 %iv
  %l.b = load i32, ptr %gep.b, align 4
  %c.1 = icmp sge i32 %l.b, 0
  br i1 %c.1, label %loop.latch, label %loop.then

loop.then:
  %l.a = load i32, ptr %gep.a, align 4
  br label %loop.latch

loop.latch:
  %merge = phi i32 [ %l.a, %loop.then ], [ %l.b, %loop.header ]
  %gep.c = getelementptr inbounds i32, ptr %c, i64 %iv
  store i32 %merge, ptr %gep.c, align 4
  %iv.next = add nuw nsw i64 %iv, 1
  %ec = icmp eq i64 %iv.next, 1000
  br i1 %ec, label %exit, label %loop.header

exit:
  ret void
}
```

Here is my modified test. I have added the Atomic Instruction after the Dereferenceable Asssume Instruction. However still this is getting Vectorized...

Putting the Atomic Instruction inside the Loop Header stops vectorization but that's not what we want to test right?

fhahn: yes it vectorizes, but does not execute the load in `loop.then` unconditionally. If the deref assumption can be used, it will execute the load unconditionally.

So with and without the `load atomic` it will get vectorized, but differently

kshitijvp: I have made the changes.

fhahn (approved):
LGTM, thanks

PR #199003, opened 2026-05-21
[AArch64][ISel] Prefer SVE over NEON for fixed-width ASL LSR

Harry-Ramsey:
Prefer SVE ASR/LSR for fixed-length vector SRA/SRL when SVE is available whilst retaining existing NEON lowering for immediate right shifts and negative shift-count splats.

### DIFF
--- a/llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
+++ b/llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
@@ -17281,6 +17281,20 @@ static bool canLowerSRLToRoundingShiftForVT(SDValue Shift, EVT ResVT,
   return true;
 }
 
+static bool isNegativeShift(SDValue Op) {
+  SDValue ShiftCount = Op;
+  if (auto *BVN = dyn_cast<BuildVectorSDNode>(Op.getNode()))
+    ShiftCount = BVN->getSplatValue();
+
+  if (!ShiftCount)
+    return false;
+  if (ShiftCount.getOpcode() != ISD::SUB)
+    return false;
+
+  SDValue ShiftValue = ShiftCount.getOperand(0);
+  return isNullConstant(ShiftValue) || isZerosVector(ShiftValue.getNode());
+}
+
 SDValue AArch64TargetLowering::LowerVectorSRA_SRL_SHL(SDValue Op,
                                                       SelectionDAG &DAG) const {
   EVT VT = Op.getValueType();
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -17333,6 +17347,13 @@ SDValue AArch64TargetLowering::LowerVectorSRA_SRL_SHL(SDValue Op,
                          Op->getFlags());
     }
 
+    if (useSVEForFixedLengthVectorVT(VT, /*OverrideNEON=*/true) &&
+        !isNegativeShift(Op.getOperand(1))) {
```
davemgreen: We can't have negative shifts for ASR/LSR nodes - it would produce poison which means we can (sort of) produce anything for.
Are you trying to detect things where the negate of the shift can be free? If so maybe name it something like isFreeToNegate?

davemgreen (commented):
The code looks OK to me but It looks like the neon ushl has more register freedom than the sve lsr. Unfortunately the SVE instructions can require an extra movprfx which either isnt always free or is at least more instructions in the loop.

Which means that whilst this is profitable outside of loops it might not be profitable inside one, depending on if there are other later uses of the value being shifted. We might need to move this later in the pipeline so that we can only use the new instruction when profitable.

davemgreen (commented):
Could this (the same logic) be added to the existing AArch64MIPeepholeOptPass?

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -17638,12 +17678,17 @@ SDValue AArch64TargetLowering::LowerVectorSRA_SRL_SHL(SDValue Op,
...
-    // instruction, but the shift left register instruction takes a signed
-    // value, where negative numbers specify a right shift.
+    // SVE preference for non-immediate shifts only. The SVE instruction is
+    // destructive, so prefer it only when the shifted value can be clobbered
+    // without needing a MOVPRFX. Keep NEON for splatted negated shift amounts
+    // because SSHL/USHL can consume the original value directly.
+    if (useSVEForFixedLengthVectorVT(VT, /*OverrideNEON=*/true) &&
+        canUseDestructiveSVERightShift(Op) &&
+        !isNegatedSplatShiftAmount(Op.getOperand(1), DAG))
+      return LowerToPredicatedOp(Op, DAG, PredOpc);
```
paulwalker-arm: We should not need this change. Instead you'll want to do `bool OverideNEON = !Subtarget->isNeonAvailable() || ....` and pass that into the existing `useSVEForFixedLengthVectorVT()` call.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -17586,6 +17586,47 @@ static bool canLowerSRLToRoundingShiftForVT(SDValue Shift, EVT ResVT,
...
+  if (ShiftedValue.hasOneUse())
+    return true;
+
+  bool IsLeftShift = ShiftedValue.getOpcode() == ISD::SHL;
+  if (ShiftedValue.getOpcode() == ISD::INTRINSIC_WO_CHAIN) {
+    auto *IID = dyn_cast<ConstantSDNode>(ShiftedValue.getOperand(0));
+    IsLeftShift = IID && (IID->getZExtValue() == Intrinsic::aarch64_neon_ushl ||
+                          IID->getZExtValue() == Intrinsic::aarch64_neon_sshl);
+  }
+
```
paulwalker-arm: I don't like trying to second guess register allocation during lowering. Add to this:
* SVE shift instructions have reversed forms so the only problematic case is when the result must be unique. 
* We generally treat MOVPRFX instructions as being free.

Given these points I would rather do away with this code and let benchmarks decide the next step.  If the second point turns out to be false in a way that affects performance then I feel we'll just introduce a Subtarget feature to block transformations like this.

paulwalker-arm (approved):

PR #192246, opened 2026-04-15
[AMDGPU][InstCombine] Optimize constant shuffle patterns

aleksandar-amd:
Detect llvm.amdgcn.wave.shuffle intrinsics where the lane index is a constant function of the lane ID and replace them with hardware-specific intrinsics.

PR #199004, opened 2026-05-21
[AMDGPU][InstCombine] Match ds_swizzle rotate mode for cyclic lane shuffles

barbara-amd:
Follow-up to 17cc4f77109d [AMDGPU][InstCombine] Optimize constant shuffle patterns (#192246).

Adds matchDsSwizzleRotatePattern to recognise shuffles of the form dst_lane = (src_lane + N) % 32 (N in [1, 31]) and lower them to a single ds_swizzle with rotate-mode encoding (ROTATE_MODE_ENC | N << 5), available on GFX9+. The bitmask mode cannot express such rotations since the carry between bit positions makes the per-bit mapping non-independent. On wave64 the pattern is accepted only when hasPeriodicLayout<32> confirms both 32-lane groups rotate by the same amount. Wave32-only ID forms (mbcnt.lo alone) are correctly rejected on wave64 targets.

### DIFF
--- a/llvm/lib/Target/AMDGPU/AMDGPUInstCombineIntrinsic.cpp
+++ b/llvm/lib/Target/AMDGPU/AMDGPUInstCombineIntrinsic.cpp
@@ -883,6 +883,29 @@ matchDsSwizzleBitmaskPattern(ArrayRef<uint8_t> Ids) {
          XorMask << AMDGPU::Swizzle::BITMASK_XOR_SHIFT;
 }
 
+/// Match a GFX9+ DS_SWIZZLE rotate-mode permutation: a cyclic left-rotation
+/// of all 32 lanes within each 32-lane group by a constant N in [0, 31],
+/// i.e. dst_lane = (src_lane + N) % 32. On wave64, hasPeriodicLayout<32>
+/// ensures both 32-lane groups rotate by the same amount.
+static std::optional<unsigned>
+matchDsSwizzleRotatePattern(ArrayRef<uint8_t> Ids) {
+  if (!hasPeriodicLayout<32>(Ids))
+    return std::nullopt;
+
+  // Determine the rotation amount from lane 0: every lane must read from
+  // lane (I + N) % 32 where N = Ids[0] and 0 <= N <= 31.
+  unsigned N = Ids[0];
+  if (N >= 32)
+    return std::nullopt;
+
+  for (unsigned I = 0; I < 32; ++I)
+    if (Ids[I] != (I + N) % 32)
+      return std::nullopt;
+
+  return AMDGPU::Swizzle::ROTATE_MODE_ENC |
+         (N << AMDGPU::Swizzle::ROTATE_SIZE_SHIFT);
+}
+
 /// Emit v_mov_b32_dpp with the given control word, row/bank masks 0xF, and
 /// bound_ctrl=1 so out-of-bounds lanes are well-defined and the DPP mov can
 /// be folded into a consuming VALU op by GCNDPPCombine.
@@ -1005,6 +1028,13 @@ static Value *matchShuffleToHWIntrinsic(IRBuilderBase &B, Value *Src,
   if (std::optional<unsigned> Imm = matchDsSwizzleBitmaskPattern(Ids))
     return createDsSwizzle(B, Src, *Imm, DL);
 
+  // DS_SWIZZLE rotate mode (GFX9+): handles cyclic 32-lane rotations that
+  // bitmask mode cannot express (e.g. +1 mod 32 requires inter-bit carry).
+  if (ST.hasDsSwizzleRotateMode()) {
+    if (std::optional<unsigned> Imm = matchDsSwizzleRotatePattern(Ids))
+      return createDsSwizzle(B, Src, *Imm, DL);
+  }
+
   if (ST.hasPermLane64() && matchHalfWaveSwapPattern(Ids))
     return createPermlane64(B, Src);
 
@@ -1048,7 +1078,6 @@ tryOptimizeShufflePattern(InstCombiner &IC, IntrinsicInst &II,
 
   return IC.replaceInstUsesWith(II, Result);
 }
-
 std::optional<Instruction *>
 GCNTTIImpl::instCombineIntrinsic(InstCombiner &IC, IntrinsicInst &II) const {
   Intrinsic::ID IID = II.getIntrinsicID();
--- a/llvm/test/Transforms/InstCombine/AMDGPU/wave-shuffle-patterns.ll
+++ b/llvm/test/Transforms/InstCombine/AMDGPU/wave-shuffle-patterns.ll
@@ -749,3 +749,285 @@ define i32 @test_broadcast_in_rows16_bitmask(i32 %val) {
   %result = call i32 @llvm.amdgcn.wave.shuffle.i32(i32 %val, i32 %idx)
   ret i32 %result
 }
+
+; Rotate-left-by-1 using mbcnt.lo only (wave32 ID form); folds to ds_swizzle
+; rotate mode (imm=0xC020) on GFX11 wave32, but not on wave64 targets (GFX11-W64,
+; GFX9) where mbcnt.lo alone is not the full lane ID.
+define i32 @test_rotate_add1_w32(i32 %val) {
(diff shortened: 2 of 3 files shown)


jayfoad:
Cc @aleksandar-amd

llvm/lib/Target/AMDGPU/AMDGPUInstCombineIntrinsic.cpp
```diff
@@ -883,6 +883,29 @@ matchDsSwizzleBitmaskPattern(ArrayRef<uint8_t> Ids) {
...
+/// ensures both 32-lane groups rotate by the same amount.
+static std::optional<unsigned>
+matchDsSwizzleRotatePattern(ArrayRef<uint8_t> Ids) {
+  if (!hasPeriodicLayout<32>(Ids))
+    return std::nullopt;
+
+  // Determine the rotation amount from lane 0: every lane must read from
+  // lane (I + N) % 32 where N = Ids[0] and 1 <= N <= 31.
+  unsigned N = Ids[0];
+  if (N == 0 || N >= 32)
```
jayfoad: Nit: I don't think you need to exclude 0 here. It should still work, even though it is a degenerate case.

llvm/test/Transforms/InstCombine/AMDGPU/wave-shuffle-patterns.ll
```diff
@@ -704,19 +704,19 @@ define i32 @test_xor5_bitmask(i32 %val) {
 }
 
 ; XOR-by-16 (cross-row). On GFX11 this matches permlanex16; on GFX9 neither
-; permlanex16 nor row_xmask exist, so the bitmask path catches it:
-; AND=0x1F, OR=0, XOR=16 -> 16415.
-define i32 @test_xor16_bitmask(i32 %val) {
-; GFX11-LABEL: @test_xor16_bitmask(
+; permlanex16 nor row_xmask exist. XOR-by-16 equals (lane+16)%32, so the
```
jayfoad: Nit: maybe call matchDsSwizzleRotatePattern after matchDsSwizzleBitmaskPattern, then you wouldn't need any changes here?

llvm/lib/Target/AMDGPU/AMDGPUInstCombineIntrinsic.cpp
```diff
@@ -999,6 +1022,13 @@ static Value *matchShuffleToHWIntrinsic(IRBuilderBase &B, Value *Src,
     }
   }
 
+  // DS_SWIZZLE rotate mode (GFX9+): handles cyclic 32-lane rotations that
+  // bitmask mode cannot express (e.g. +1 mod 32 requires inter-bit carry).
+  if (ST.getGeneration() >= AMDGPUSubtarget::GFX9) {
```
jayfoad: Should have a named helper function in GCNSubtarget for this specific feature.

llvm/test/Transforms/InstCombine/AMDGPU/wave-shuffle-patterns.ll
```diff
@@ -749,3 +749,180 @@ define i32 @test_broadcast_in_rows16_bitmask(i32 %val) {
...
+  %lane = call i32 @llvm.amdgcn.mbcnt.hi(i32 -1, i32 %lo)
+  %add = add i32 %lane, 1
+  %idx = and i32 %add, 63
+  %result = call i32 @llvm.amdgcn.wave.shuffle.i32(i32 %val, i32 %idx)
+  ret i32 %result
+}
+
+; Identity shuffle (N=0 rejected by rotate matcher; add %lane,0 folds to %lane).
+; On GFX11 wave32 this still folds to the identity quad-perm via DPP; on wave64
+; targets (GFX11-W64, GFX9) mbcnt.lo alone is not the full lane ID so no fold.
```
jayfoad: Rather than testing for these specific lowerings, why not add a special case that detects the identity shuffle and just removes the wave.shuffle call? But in a separate PR.

jayfoad (commented):
Looks good overall.

barbara-amd:
PR updated with the following changes:
- Added a `hasDsSwizzleRotateMode` helper in `GCNSubtarget`
- `matchDsSwizzleRotatePattern` is now called after `matchDsSwizzleBitmaskPattern`

llvm/test/Transforms/InstCombine/AMDGPU/wave-shuffle-patterns.ll
```diff
@@ -749,3 +749,180 @@ define i32 @test_broadcast_in_rows16_bitmask(i32 %val) {
...
+  %rot = and i32 %add, 31
+  %hi = and i32 %lane, 32
+  %idx = or i32 %rot, %hi
+  %result = call i32 @llvm.amdgcn.wave.shuffle.i32(i32 %val, i32 %idx)
+  ret i32 %result
+}
+
+; Negative: full-wave64 rotate crosses the 32-lane group boundary, so
+; hasPeriodicLayout<32> rejects it and no fold occurs.
+define i32 @test_rotate_add1_full_w64(i32 %val) {
```
PankajDwivedi-25, line 871: Can we have a test for both halves rotating, but by different amounts?

llvm/test/Transforms/InstCombine/AMDGPU/wave-shuffle-patterns.ll
```diff
@@ -749,3 +749,180 @@ define i32 @test_broadcast_in_rows16_bitmask(i32 %val) {
...
+  %idx = and i32 %add, 31
+  %result = call i32 @llvm.amdgcn.wave.shuffle.i32(i32 %val, i32 %idx)
+  ret i32 %result
+}
+
+; Same rotate-left-by-1 via ds_bpermute (byte-addressed); folds to ds_swizzle
+; rotate mode (imm=0xC020) on GFX11 wave32, but not on wave64 targets (GFX11-W64,
+; GFX9) where mbcnt.lo alone is not the full lane ID.
+define i32 @test_bpermute_rotate_add1_w32(i32 %val) {
+; GFX11-LABEL: @test_bpermute_rotate_add1_w32(
```
PankajDwivedi-25, line 815: Should we add the rotate-pattern test for f32 / pointer?

PankajDwivedi-25 (approved):
LGTM, with minor test cases addition, please wait for other reviewers approval too.

llvm/test/Transforms/InstCombine/AMDGPU/wave-shuffle-patterns.ll
PankajDwivedi-25: I think thats just one or two line change can we add in this patch itself?

In tryOptimizeShufflePattern, after tryBuildShuffleMap, all that is needed, I guess.

`if (llvm::all_of(seq(Ids.size()), [&](unsigned I) { return Ids[I] == I; }))
  return IC.replaceInstUsesWith(II, Src);`

barbara-amd:
Updated  added tests for asymmetric half-wave rotation, f32 and pointer type rotate.

llvm/test/Transforms/InstCombine/AMDGPU/wave-shuffle-patterns.ll
barbara-amd: I think it is better keep the identity shuffle detection as a separate PR - it doesn't really belong in a change titled "match ds_swizzle rotate mode".

saxlungs:
@arsenm You had mentioned on a PR of mine previously this PR: https://github.com/llvm/llvm-project/pull/182391

Do we want to be encouraging using this methodology for these kinds of "does this architecture/subtarget have this feature?" checks in the future, or was that specifically just for DPP-related checks?

arsenm:
> Do we want to be encouraging using this methodology for these kinds of "does this architecture/subtarget have this feature?" checks in the future, or was that specifically just for DPP-related checks?

Generation checks should always be avoided in all contexts. We don't have monotonically increasing ISA changes. It gets ugly to express the conditions on those kinds of changes, and it's a pain to untangle why the generation checks are what they are

barbara-amd:
Is the patch good to go as-is, or would you like any further changes? If it's good to go, could someone merge it on my behalf, as I don't have commit access yet?

arsenm (approved):

PR #199006, opened 2026-05-21
[Flang][OpenMP] Allow workdistribute inside 'target teams'

skatrak:
Currently, a `workdistribute` construct nested inside of a combined `target teams` is incorrectly reported as an error. This patch fixes that.


skc7 (approved):
LGTM

PR #199008, opened 2026-05-21
[RISCV] Ensure AVL dominates True in vmerge peephole

### LINKED ISSUE #198733: [LLVM][RISCV][RVV][Full LTO] MachineVerifier failure: Virtual register defs don't dominate all uses
Reproduces only with full LTO (-flto).
Does not reproduce with:
  - no -flto
  - -flto=thin

Target:
  -march=rv64gcv
  -mrvv-vector-bits=zvl

Optimization:
  -O3
  -O2

[red.c](https://github.com/user-attachments/files/28050721/red.c)

[verbose-logs.txt](https://github.com/user-attachments/files/28050723/verbose-logs.txt)

[compiler-opts.txt](https://github.com/user-attachments/files/28050733/compiler-opts.txt)

[compiler.txt](https://github.com/user-attachments/files/28050735/compiler.txt)

*** Bad machine code: Virtual register defs don't dominate all uses. ***
- function:    main
- v. register: %656
LLVM ERROR: Found 1 machine code errors.

This issue is Found via fuzzer.

lukel97:
When folding vmerge into its true operand, if vmerge has an AVL defined by a register and true has VLMAX, then the minimum AVL will be the register. In this case it's not guaranteed to dominate true, so we need to potentially sink true so it does.

This teaches ensureDominates to check multiple definitions at the same time, since we want the sinking to be atomic.

Fixes #198733

### DIFF
--- a/llvm/lib/Target/RISCV/RISCVVectorPeephole.cpp
+++ b/llvm/lib/Target/RISCV/RISCVVectorPeephole.cpp
@@ -71,7 +71,8 @@ private:
   bool hasSameEEW(const MachineInstr &User, const MachineInstr &Src) const;
   bool isAllOnesMask(const MachineInstr *MaskDef) const;
   std::optional<unsigned> getConstant(const MachineOperand &VL) const;
-  bool ensureDominates(const MachineOperand &Use, MachineInstr &Src) const;
+  bool ensureDominates(ArrayRef<const MachineOperand> Defs,
+                       MachineInstr &Use) const;
   Register
   lookThruCopies(Register Reg, bool OneUseOnly = false,
                  SmallVectorImpl<MachineInstr *> *Copies = nullptr) const;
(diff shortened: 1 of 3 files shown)


wangpc-pp (approved):
LGTM.

llvm/lib/Target/RISCV/RISCVVectorPeephole.cpp
```diff
@@ -462,22 +463,29 @@ static bool dominates(MachineBasicBlock::const_iterator A,
...
-/// If the register in \p MO doesn't dominate \p Src, try to move \p Src so it
+/// If a register in \p Uses doesn't dominate \p Src, try to move \p Src so it
 /// does. Returns false if doesn't dominate and we can't move. \p MO must be in
 /// the same basic block as \Src.
-bool RISCVVectorPeephole::ensureDominates(const MachineOperand &MO,
-                                          MachineInstr &Src) const {
-  assert(MO.getParent()->getParent() == Src.getParent());
-  if (!MO.isReg() || !MO.getReg().isValid())
-    return true;
+bool RISCVVectorPeephole::ensureDominates(ArrayRef<const MachineOperand> Defs,
```
topperc, line 469: Could this be ArrayRef<const *MachineOperand> instead of copying the MachineOperands? Not sure how expensive they are to copy.

llvm/lib/Target/RISCV/RISCVVectorPeephole.cpp
```diff
@@ -462,22 +463,29 @@ static bool dominates(MachineBasicBlock::const_iterator A,
   return &*I == A;
 }
 
-/// If the register in \p MO doesn't dominate \p Src, try to move \p Src so it
+/// If a register in \p Uses doesn't dominate \p Src, try to move \p Src so it
 /// does. Returns false if doesn't dominate and we can't move. \p MO must be in
 /// the same basic block as \Src.
```
bababuck, line 468: `Src` got replaced by `Use`/`Dest` in the function, didn't get updated in the comment.

llvm/lib/Target/RISCV/RISCVVectorPeephole.cpp
lukel97: Done in #199065

llvm/lib/Target/RISCV/RISCVVectorPeephole.cpp
lukel97: Woops thanks for catching, fixed in 866e9e03fe86eab363cc70526a59c8e48067a576

PR #199009, opened 2026-05-21
Add clang warning if fp exception functions are called without appropriate flags/pragmas

### LINKED ISSUE #128239: clang should warn if fp exception functions are called without appropriate flags/pragmas
We regularly get users complaining that fetestexcept()/feenableexcept()/etc. don't work properly.  We document that you need to use appropriate flags or `#pragma STDC FENV_ACCESS`.  But it's hard to find if you don't know what you're looking for.

clang should be able to detect calls to those functions and print a warning.  (This won't catch 100% of cases, but it's low-hanging fruit.)

maarcosrmz:
Fixes https://github.com/llvm/llvm-project/issues/128239

The implementation adds warnings for floating-point exception function calls (fenv.h) made without enabling strict floating-point semantics via `-ftrapping-math -frounding-math` or `#pragma STDC FENV_ACCESS ON`. It is only emitted on targets where strict floating-point behavior is supported. Recognizing all fenv.h builtins required adding `fexcept_t` and `fenv_t` as builtin types.

Files relevant for building flang and flang-rt that trigger the new warning were adapted to correctly use the pragma in cases where the compiler supports it.


maarcosrmz:
@efriedma-quic I created this as the follow-up PR to https://github.com/llvm/llvm-project/pull/187860. The notable changes are:
* Added a check for strict FP support before issuing the warning
* Added a test run for that case (see [fenv-access.c:4](https://github.com/llvm/llvm-project/pull/199009/changes#diff-9a3d067315c2151e2259f05e0465f1ae872eae2d482b0c4b7416c44b0581c81eR4), [fenv-access.c:56-58](https://github.com/llvm/llvm-project/pull/199009/changes#diff-9a3d067315c2151e2259f05e0465f1ae872eae2d482b0c4b7416c44b0581c81eR56-R58))

If there is anything else I should take into account based on the feedback from the original PR, please let me know.

efriedma-quic:
I think we need to address the other issues I mentioned before we merge this again.

For libc, maybe you can look into fixing [FEnvImpl.h](https://github.com/llvm/llvm-project/blob/f97789da65cc2e55d5ff77b83b6f779d1d11aa7f/libc/src/__support/FPUtil/FEnvImpl.h#L52)?  It looks like maintainers aren't actively investigating.

For flang, let's give the maintainers a little time to give an update.

libc/src/__support/FPUtil/FEnvImpl.h
```diff
@@ -37,26 +37,41 @@
 namespace LIBC_NAMESPACE_DECL {
 namespace fputil {
 
-LIBC_INLINE int clear_except(int excepts) { return feclearexcept(excepts); }
+LIBC_INLINE int clear_except(int excepts) {
+#pragma STDC FENV_ACCESS ON
```
lntue: Can you move libc changes to a separate PR?  Currently we disable this warning so your PR shouldn't break anything from us? https://github.com/llvm/llvm-project/blob/main/libc/cmake/modules/LLVMLibCCompileOptionRules.cmake#L327

The main reason is that it will be a lot more involved to add this pragma due to compatibility issues with other targets and compilers:
https://godbolt.org/z/dcYheosoY

maarcosrmz: I actually realised what you said just after pushing. If I'm not mistaken, the changes I just pushed should make it work, even for different targets (inspired by https://github.com/llvm/llvm-project/pull/198692/changes#diff-cbc6b94668e40294e8142de317bf19e528ab3fed9521f80d3ab11d7ddcfcce54).

Is this enough or should I still open a separate PR?

lntue: I don't think the fallback works:
```
// Portable fallback for GCC, MSVC, or older clang.
#define LIBC_FENV_ACCESS_ON _Pragma("STDC FENV_ACCESS ON")
```
https://godbolt.org/z/8bcdxojj9

We could add a compiler check to https://github.com/llvm/llvm-project/blob/main/libc/cmake/modules/CheckCompilerFeatures.cmake
and pass that definition to all libc's targets.

You also need to remove the (Wno-fenv-access) flag that we added.

Let's move it to a separate PR if it doesn't break this PR.

maarcosrmz: I'd actually be this: https://godbolt.org/z/7vqeznnxG.
The fallback refers to the case when `#pragma clang fp exceptions(maytrap)` is not supported but `#pragma STDC FENV_ACCESS ON` is. Those, however, are only set if the target itself supports strict fp behavior.

lntue: You forgot to expand those `LIBC_TARGET_*` macros, they are not defined in the standard headers.  This is the result of expanding them:
https://godbolt.org/z/Ezr1bxxr5

maarcosrmz: My bad, I see where this is going. I'll revert the commits here and create a separate PR. Thanks for your input!

efriedma-quic:
@eugeneepshteyn do you know if there's still more that needs to be done in flang?

maarcosrmz:
@efriedma-quic @eugeneepshteyn do you have any updates? If not it would be nice if we could merge this rather soon, before the PR goes stale or has too many merge conflicts.

maarcosrmz:
Apologies for the review-request noise on this PR. A bad rebase briefly ballooned the diff across many unrelated files and auto-requested a large set of code owners. The branch is now force-pushed back to just the clang fenv changes. @efriedma-quic  / @lntue, when you have a moment, could one of you please clear the stale reviewer requests? I don't have the access to remove them myself. Thanks, and sorry for the pings to everyone who got caught in it.

efriedma-quic:
Can you try a -Werror=fenv-access build of flang to check if there are still errors?  It's not clear to me from the previous discussion whether there are still issues on the flang side.

clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -1053,6 +1053,11 @@ def err_ptrauth_indirect_goto_addrlabel_arithmetic : Error<
   "%select{subtraction|addition}0 of address-of-label expressions is not "
   "supported with ptrauth indirect gotos">;
 
+def warn_fe_access_without_fenv_access : Warning<
+  "'%0' used without enabling floating-point exception behavior; use 'pragma STDC "
+  "FENV_ACCESS ON' or compile with '-ffp-exception-behavior=maytrap'">,
```
andykaylor: `maytrap` just prevents speculatively executing instructions that could raise an exception. It doesn't guarantee strict exception semantics. It also doesn't make any guarantees about rounding mode, though in practice you get that as a side effect. Should we be recommending '-ffp-model=strict' instead or at least mentioning it as an alternative?

efriedma-quic: I didn't really think carefully about the exact suggestion here.

It might be better to suggest something that also works with gcc, if possible.

andykaylor: For gcc, you'd need `-ftrapping-math` and `-frounding-math`. That would work for Clang too.

efriedma-quic:
@eugeneepshteyn Any update on flang?

@maarcosrmz were you able to try building flang like I suggested?

maarcosrmz:
@efriedma-quic Yes I have. There are two files that cause the build to fail with `-Werror=fenv-access`:
* [flang/lib/Evaluate/host.cpp](https://github.com/llvm/llvm-project/blob/main/flang/lib/Evaluate/host.cpp)
* [llvm/lib/Analysis/ConstantFolding.cpp](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Analysis/ConstantFolding.cpp)

Moreover, as some files include [libc/src/__support/FPUtil/FEnvImpl.h](https://github.com/llvm/llvm-project/blob/main/libc/src/__support/FPUtil/FEnvImpl.h), it also breaks the build. But this is fixable by adding `-DLIBC_COMPILER_HAS_STDC_FENV_ACCESS` when building flang (see #200268), although I'm not sure whether that is best-practice.

efriedma-quic:
It looks like ConstantFolding.cpp is missing the pragmas.

For host.cpp (and its caller ApplyHostFunctionHelper), it looks like there are also missing pragmas.

I think we need to fix those first.

Maybe @lntue can comment on how LIBC_COMPILER_HAS_STDC_FENV_ACCESS is supposed to work with hand-in-hand?

lntue:
> It looks like ConstantFolding.cpp is missing the pragmas.
> 
> For host.cpp (and its caller ApplyHostFunctionHelper), it looks like there are also missing pragmas.
> 
> I think we need to fix those first.
> 
> Maybe @lntue can comment on how LIBC_COMPILER_HAS_STDC_FENV_ACCESS is supposed to work with hand-in-hand?

@efriedma-quic : My plan for hand-in-hand to completely avoid touching floating point environments inside libc codes, like:
https://github.com/llvm/llvm-project/blob/main/llvm/lib/Support/APFloat.cpp#L41
and only do static rounding if needed, at least for clang/LLVM/compiler-rt from what I've seen so far.

maarcosrmz:
I created #224384 to fix the error triggered when building subprojects that include libc headers using FP environment functions, e.g. flang in our case. If @lntue's suggestion of avoiding any access to the FP environment inside libc when building these subprojects holds for all uses, this fix should prevent similar errors in the future. (@lntue, I would appreciate it if you could take a look at that PR.)

On a different note, the approach used [here](https://github.com/llvm/llvm-project/blob/main/flang/include/flang/Common/fp-control.h#L24-L32) breaks compilation for compilers that do not support the pragma (see [this](https://godbolt.org/z/svMoz4Kx6)). I therefore believe flang might need to follow libc (see https://github.com/llvm/llvm-project/pull/200268) in determining at build configuration time whether the pragma should be used. I suggest we do the same in flang and flang-rt, and in particular in [flang/lib/Evaluate/host.cpp](https://github.com/llvm/llvm-project/blob/main/flang/lib/Evaluate/host.cpp) and [llvm/lib/Analysis/ConstantFolding.cpp](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Analysis/ConstantFolding.cpp). @efriedma-quic, if there are no objections I will go ahead with this.

efriedma-quic:
> On a different note, the approach used [here](https://github.com/llvm/llvm-project/blob/main/flang/include/flang/Common/fp-control.h#L24-L32) breaks compilation for compilers that do not support the pragma (see [this](https://godbolt.org/z/svMoz4Kx6)). I therefore believe flang might need to follow libc (see #200268) in determining at build configuration time whether the pragma should be used. I suggest we do the same in flang and flang-rt, and in particular in [flang/lib/Evaluate/host.cpp](https://github.com/llvm/llvm-project/blob/main/flang/lib/Evaluate/host.cpp) and [llvm/lib/Analysis/ConstantFolding.cpp](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Analysis/ConstantFolding.cpp). @efriedma-quic, if there are no objections I will go ahead with this.

This seems fine.

clang/lib/Serialization/ASTWriter.cpp
```diff
@@ -6162,6 +6162,8 @@ ASTFileSignature ASTWriter::WriteASTCore(Sema *SemaPtr, StringRef isysroot,
     AddTypeRef(Context, Context.ObjCClassRedefinitionType, SpecialTypes);
     AddTypeRef(Context, Context.ObjCSelRedefinitionType, SpecialTypes);
     AddTypeRef(Context, Context.getucontext_tType(), SpecialTypes);
+    AddTypeRef(Context, Context.getfexcept_tType(), SpecialTypes);
+    AddTypeRef(Context, Context.getfenv_tType(), SpecialTypes);
```
ChuanqiXu9, line 6166: Sorry at first as I didn't look through all the patches. But we do want to avoid predefined types if possible. Why does the types are special enough that we have to write them unconditionally?

maarcosrmz: If I understood the recognition mechanism correctly, were we not to treat them as predefined types, we would not be able to recognize the functions from the fenv.h header (see [Builtins.td](https://github.com/llvm/llvm-project/pull/199009/changes#diff-987cbd809c993aa5915def17d1f8276fd08a12aa906f56335faee50044ac5537R4706-R4709)), which take arguments of these types.

ChuanqiXu9: We still need to figure out how to solve these issues. e.g., IgnoreSignature or other mechanism

PR #199011, opened 2026-05-21
[mlir][affine][NFC] Add visited set to hasDependencePath

chokobole:
Found this while building a zkVM. When my compiler lowers a fully-unrolled Poseidon2 permutation (Poseidon2 is a ZK hash) to MLIR the output is ~4 kLOC, and `affine-loop-fusion` hangs at 99% CPU on it.

`hasDependencePath` does a DFS with no visited set. The MDG is a DAG, so on inputs with lots of diamond merges it re-walks the same paths over and over — `O(2^V)` worst case. Adding a visited set fixes it (`O(V + E)`), same boolean returned since every reachable node is still visited at least once.

Used Claude to help draft the patch and this PR. I've read through the diff and stand behind it.


ftynse (changes requested):
Using AI tooling without disclosure is against our policy https://llvm.org/docs/AIToolPolicy.html.

ftynse:
If you want to proceed with this patch, _you_ must undertsand the problem and phrase it intelligibly in the PR description. You must also provide a test that is consistent with how MLIR tests its input.

ftynse:
We strongly prefer IR-based tests via LIT over unittests. Is there no way a LIT test can be constructed?

joker-eph:
This is NFC isn't it? Why would we even need a test?

chokobole:
NFC in the strict sense — happy to drop it. If we keep a test, I'll try `mlir-reduce` on the original Poseidon2 IR for a lit version.

mlir/lib/Dialect/Affine/Analysis/Utils.cpp
```diff
@@ -574,6 +574,14 @@ bool MemRefDependenceGraph::hasDependencePath(unsigned srcId,
   SmallVector<std::pair<unsigned, unsigned>, 4> worklist;
   worklist.push_back({srcId, 0});
   Operation *dstOp = getNode(dstId)->op;
+  // Track nodes already pushed onto the worklist. The MDG is a DAG (edges
+  // respect program order), but without a visited set this DFS enumerates
+  // every path from 'srcId' to every reachable node — exponential in the
+  // number of diamond merges. With many independent memref accesses on the
+  // same buffer (e.g. fully-unrolled SIMD-like bodies that load and store
+  // from one buffer between several affine.for loops), this is fatal.
```
joker-eph:
```suggestion
  // Track nodes already pushed onto the worklist to avoid redundant visit.
```

joker-eph:
Let's tag the PR as "NFC" (update the title) and remove the test: speeding up a worklist algorithm with a visited set seems like trivial enough to me.

joker-eph (approved):

joker-eph:
LGTM right now, @ftynse ?

PR #199012, opened 2026-05-21
[lldb][windows] Fix HostProcessWindows::Terminate clobbering error

charles-zablit:
When `m_process` is `nullptr` the function sets `ERROR_INVALID_HANDLE` then fell through to call `TerminateProcess(nullptr, 0)`, which clobbers the error.

Return early instead.


DavidSpickett (approved):
Makes sense. Also checked against https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminateprocess. LGTM.

PR #199013, opened 2026-05-21
[lldb][windows] Fix HostThreadWindows::Cancel inverted error reporting

charles-zablit:
QueueUserAPC returns non-zero on success and zero on failure. The old code stored the return value and unconditionally constructed a Status from it, so a successful call reported a phantom Win32 error code and a real failure reported success.

Check the return and read GetLastError() on failure.


DavidSpickett (approved):
Matches https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-queueuserapc, so LGTM.

PR #199014, opened 2026-05-21
[lldb][windows] Fix HostThreadWindows::Join dead-code on GetExitCodeThread failure

charles-zablit:
The old code unconditionally assigned `exit_code` to `*result` after the `GetExitCodeThread` check, so the failure branch assignment of 0 was overwritten. We never returned the `exit_code` value. `exit_code` happens to be zero-initialized so the bug was masked, but the structure was wrong.

### DIFF
--- a/lldb/source/Host/windows/HostThreadWindows.cpp
+++ b/lldb/source/Host/windows/HostThreadWindows.cpp
@@ -31,18 +31,22 @@ HostThreadWindows::~HostThreadWindows() { Reset(); }
 void HostThreadWindows::SetOwnsHandle(bool owns) { m_owns_handle = owns; }
 
 Status HostThreadWindows::Join(lldb::thread_result_t *result) {
+  if (!IsJoinable())
+    return Status(ERROR_INVALID_HANDLE, eErrorTypeWin32);
+
   Status error;
-  if (IsJoinable()) {
-    DWORD wait_result = ::WaitForSingleObject(m_thread, INFINITE);
-    if (WAIT_OBJECT_0 == wait_result && result) {
+  DWORD wait_result = ::WaitForSingleObject(m_thread, INFINITE);
+  if (wait_result == WAIT_OBJECT_0) {
+    if (result) {
       DWORD exit_code = 0;
-      if (!::GetExitCodeThread(m_thread, &exit_code))
+      if (::GetExitCodeThread(m_thread, &exit_code))
+        *result = exit_code;
+      else
         *result = 0;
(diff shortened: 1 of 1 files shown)


lldb/source/Host/windows/HostThreadWindows.cpp
```diff
@@ -31,18 +31,22 @@ HostThreadWindows::~HostThreadWindows() { Reset(); }
...
   Status error;
-  if (IsJoinable()) {
-    DWORD wait_result = ::WaitForSingleObject(m_thread, INFINITE);
-    if (WAIT_OBJECT_0 == wait_result && result) {
+  DWORD wait_result = ::WaitForSingleObject(m_thread, INFINITE);
+  if (wait_result == WAIT_OBJECT_0) {
+    if (result) {
       DWORD exit_code = 0;
-      if (!::GetExitCodeThread(m_thread, &exit_code))
+      if (::GetExitCodeThread(m_thread, &exit_code))
```
DavidSpickett, line 42: I was going to say you don't need to if here, because either it writes to exit_code or it doesn't.
But A: it looks weird to not check the result and B: we'd have to check it doesn't modify exit_code in case of failure too, it might be undefined.

lldb/source/Host/windows/HostThreadWindows.cpp
```diff
@@ -31,18 +31,22 @@ HostThreadWindows::~HostThreadWindows() { Reset(); }
...
+  if (!IsJoinable())
+    return Status(ERROR_INVALID_HANDLE, eErrorTypeWin32);
+
   Status error;
-  if (IsJoinable()) {
-    DWORD wait_result = ::WaitForSingleObject(m_thread, INFINITE);
-    if (WAIT_OBJECT_0 == wait_result && result) {
+  DWORD wait_result = ::WaitForSingleObject(m_thread, INFINITE);
+  if (wait_result == WAIT_OBJECT_0) {
+    if (result) {
```
DavidSpickett, line 40: Was going to ask for early returns, but you need to call Reset() at the end of each code path, so early returns creates more noise than it removes.

DavidSpickett (approved):
LGTM

> so the bug was masked

You don't explicitly say what the bug is. I think it is "we never returned the exit code value".

Can't have been that important though, to have been wrong this whole time. I guess we just report it somewhere but don't take actions based on the exit code value.

charles-zablit:
> You don't explicitly say what the bug is. I think it is "we never returned the exit code value".

Fixed the PR description.

PR #199015, opened 2026-05-21
[lldb][windows] Use GetEnvironmentVariableW instead of _wgetenv

charles-zablit:
`_wgetenv` is not thread-safe per Microsoft docs:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?view=msvc-170.

Use `GetEnvironmentVariableW` with a sized buffer instead.

### DIFF
--- a/lldb/source/Host/windows/HostInfoWindows.cpp
+++ b/lldb/source/Host/windows/HostInfoWindows.cpp
@@ -127,9 +127,15 @@ bool HostInfoWindows::GetEnvironmentVar(const std::string &var_name,
   if (!llvm::ConvertUTF8toWide(var_name, wvar_name))
     return false;
 
-  if (const wchar_t *wvar = _wgetenv(wvar_name.c_str()))
-    return llvm::convertWideToUTF8(wvar, var);
-  return false;
+  DWORD len = ::GetEnvironmentVariableW(wvar_name.c_str(), nullptr, 0);
+  if (len == 0)
+    return false;
+  std::vector<wchar_t> buffer(len);
+  DWORD written =
+      ::GetEnvironmentVariableW(wvar_name.c_str(), buffer.data(), len);
+  if (written == 0 || written >= len)
+    return false;
+  return llvm::convertWideToUTF8(buffer.data(), var);
 }
 
 static llvm::ManagedStatic<WindowsUserIDResolver> g_user_id_resolver;


DavidSpickett (approved):
The usage matches https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getenvironmentvariablew but I don't see anything on that page saying that it is thread safe. Whereas for _wgetenv it explicitly says it's not.

Is that just how Win32 docs work? Note the ones that are not thread safe, the rest are implicitly safe?

I see there is a SetEnvironmentVariable. Maybe it is safe as long as we aren't using that at the same time.

If either of those guesses is true, this LGTM.

charles-zablit:
> The usage matches [learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getenvironmentvariablew](https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getenvironmentvariablew) but I don't see anything on that page saying that it is thread safe. Whereas for _wgetenv it explicitly says it's not.
> 
> Is that just how Win32 docs work? Note the ones that are not thread safe, the rest are implicitly safe?
> 
> I see there is a SetEnvironmentVariable. Maybe it is safe as long as we aren't using that at the same time.
> 
> If either of those guesses is true, this LGTM.

I think I did not read the docs correctly when I wrote this. I got my source from the .net docs: https://learn.microsoft.com/en-us/dotnet/core/unmanaged-api/profiling/icorprofilerinfo11-getenvironmentvariable-method

The does do not guarantee that `GetEnvironmentVariable` is thread safe. Closing this.

DavidSpickett:
> The does do not guarantee that GetEnvironmentVariable is thread safe. Closing this.

If you can make an argument that it's a bit more safe than the other function, I'd accept that.

charles-zablit:
> > The does do not guarantee that GetEnvironmentVariable is thread safe. Closing this.
> 
> If you can make an argument that it's a bit more safe than the other function, I'd accept that.

I searched on stack overflow for empirical findings, but I couldn't find anything unfortunately.

DavidSpickett:
The current function:
> Returns a pointer to the environment table entry containing varname.

And I was going to say that the new function returning a copy makes it better, but without knowing what goes on in the Windows kernel, we could just be pushing the copy code into kernel space. Maybe the time frame is smaller but it's the same risk really.

PR #199016, opened 2026-05-21
[lldb][windows] Null-check m_session_data in ProcessDebugger::HaltProcess

charles-zablit:
Every other public method in this file checks `m_session_data` before dereferencing.


DavidSpickett (commented):
Makes me wonder if there's a pattern we could use to avoid the null checks, but that is not a topic for this PR.

LGTM.

Issue #109483, opened 2024-09-20
Meta issue to track changes for adding plugin and LLVM_BUILD_LLVM_DYLIB support for Windows


fsfod:
Some rational for these changes is explained in this old LLVM discourse post [Supporting LLVM_BUILD_LLVM_DYLIB on Windows](https://discourse.llvm.org/t/supporting-llvm-build-llvm-dylib-on-windows/58891).

## LLVM
- [x] #96630
- [x] #107873
- [ ] #108051
- [ ] #109024
- [x] #109156
- [ ] Switch to new visibility macros for X86 Target initialize functions definitions.
- [ ] Add visibility macros for Target initialize function declarations.
- [ ] Clang tooling generated visibility macros for LLVM.
- [ ] clang-cl template dllexport workaround.
- [ ] Default symbol visibility on non windows platforms set to hidden.
- [ ] MSVC deleted and defaulted class members changes.
- [ ] MSVC dllexport ilist fix or default template argument workaround.
- [ ] MSVC dllexport cl::opt fix or workaround.

## Clang
- [ ] #108276
- [ ] #109362 
- [ ] #109321
- [x] #109319
- [ ] Clang tooling generated visibility macros for Clang
- [ ] MSVC deleted and defaulted class members changes.

annamthomas:
hi @fsfod , I'm having trouble building my local downstream fork of LLVM which uses dynamic build and link (i.e. these options are present: -DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_LINK_LLVM_DYLIB=ON). 

If I revert back to before the first change landed in this series, it builds completely fine. 
The reason is that we have some dynamic dependencies between our downstream library and LLVM and when building those files, it complains of same issue as reported here: https://github.com/llvm/llvm-project/issues/107687

i.e. 
```
[02:35:10] #include "llvm/CodeGen/GenVT.inc"
[02:35:10]          ^~~~~~~~~~~~~~~~~~~~~~~~
[02:35:10] 1 error generated.
```

Is there any workarounds you can suggest for this? I've tried adding the `-DLLVM_BUILD_LLVM_DYLIB_VIS=ON` but it does not help.

fsfod:
I'm not sure why my changes would cause that file not to be generated. I did notice there was a merged PR thats looks similar to your issue https://github.com/llvm/llvm-project/pull/109306

annamthomas:
I'm on truck but that didn't fix it.

[...]

PR #199019, opened 2026-05-21
llvm: Fix most LLVM_ABI annotations in Analysis

Steelskin:
This updates most LLVM_ABI annotations in the Analysis headers to match expected usage:
* All public APIs should be properly annotated.
* Inlined functions should not be annotated.

These changes were done by a script fixing annotations on LLVM public headers and manually checked.

This effort is tracked in #109483.

### DIFF
--- a/llvm/include/llvm/Analysis/DominanceFrontier.h
+++ b/llvm/include/llvm/Analysis/DominanceFrontier.h
@@ -126,8 +126,9 @@ public:
   void dump() const;
 };
 
-extern template class DominanceFrontierBase<BasicBlock, false>;
-extern template class DominanceFrontierBase<BasicBlock, true>;
+extern template class LLVM_TEMPLATE_ABI
+    DominanceFrontierBase<BasicBlock, false>;
+extern template class LLVM_TEMPLATE_ABI DominanceFrontierBase<BasicBlock, true>;
 
 /// Analysis pass which computes a \c DominanceFrontier.
 class DominanceFrontierAnalysis
--- a/llvm/include/llvm/Analysis/DominanceFrontier.h
+++ b/llvm/include/llvm/Analysis/DominanceFrontier.h
@@ -100,11 +100,11 @@ public:
   using const_iterator = DominanceFrontier::const_iterator;
 
   /// Handle invalidation explicitly.
-  bool invalidate(Function &F, const PreservedAnalyses &PA,
-                  FunctionAnalysisManager::Invalidator &);
+  LLVM_ABI bool invalidate(Function &F, const PreservedAnalyses &PA,
+                           FunctionAnalysisManager::Invalidator &);
 };
 
-class DominanceFrontierWrapperPass : public FunctionPass {
+class LLVM_ABI DominanceFrontierWrapperPass : public FunctionPass {
   DominanceFrontier DF;
 
 public:
@@ -141,7 +141,7 @@ public:
   using Result = DominanceFrontier;
 
   /// Run the analysis pass over a function and produce a dominator tree.
-  DominanceFrontier run(Function &F, FunctionAnalysisManager &AM);
+  LLVM_ABI DominanceFrontier run(Function &F, FunctionAnalysisManager &AM);
 };
 
 /// Printer pass for the \c DominanceFrontier.
@@ -150,9 +150,9 @@ class DominanceFrontierPrinterPass
   raw_ostream &OS;
 
 public:
-  explicit DominanceFrontierPrinterPass(raw_ostream &OS);
+  LLVM_ABI explicit DominanceFrontierPrinterPass(raw_ostream &OS);
 
-  PreservedAnalyses run(Function &F, FunctionAnalysisManager &AM);
+  LLVM_ABI PreservedAnalyses run(Function &F, FunctionAnalysisManager &AM);
 };
 
 } // end namespace llvm
--- a/llvm/include/llvm/Analysis/HashRecognize.h
+++ b/llvm/include/llvm/Analysis/HashRecognize.h
@@ -31,7 +31,7 @@ struct CRCTable : public std::array<APInt, 256> {
   LLVM_ABI void print(raw_ostream &OS) const;
 
 #if !defined(NDEBUG) || defined(LLVM_ENABLE_DUMP)
-  LLVM_DUMP_METHOD void dump() const;
+  LLVM_ABI LLVM_DUMP_METHOD void dump() const;
 #endif
 };
 
--- a/llvm/include/llvm/Analysis/HashRecognize.h
+++ b/llvm/include/llvm/Analysis/HashRecognize.h
@@ -28,7 +28,7 @@ class LPMUpdater;
 
 /// A custom std::array with 256 entries, that also has a print function.
 struct CRCTable : public std::array<APInt, 256> {
-  void print(raw_ostream &OS) const;
+  LLVM_ABI void print(raw_ostream &OS) const;
 
 #if !defined(NDEBUG) || defined(LLVM_ENABLE_DUMP)
   LLVM_DUMP_METHOD void dump() const;
@@ -65,9 +65,9 @@ struct PolynomialInfo {
   // zero.
   Value *LHSAux;
 
-  PolynomialInfo(unsigned TripCount, Value *LHS, const APInt &RHS,
-                 Value *ComputedValue, bool ByteOrderSwapped,
-                 Value *LHSAux = nullptr);
+  LLVM_ABI PolynomialInfo(unsigned TripCount, Value *LHS, const APInt &RHS,
+                          Value *ComputedValue, bool ByteOrderSwapped,
+                          Value *LHSAux = nullptr);
 };
 
 /// The analysis.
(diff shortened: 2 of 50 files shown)


kasuga-fj:
I'm not sure why I was asked to review this, as I'm completely unfamiliar with the background and context of this work. That said, if nobody else is available, I'm okay to take on the double-check. In that case, could you please provide some more details about the expected usage of LLVM_ABI? For example, when should a class/struct be annotated, and what does it imply? (In that case, do we still need to annotate the member functions of that class, or not?)

Steelskin:
> I'm not sure why I was asked to review this, as I'm completely unfamiliar with the background and context of this work. That said, if nobody else is available, I'm okay to take on the double-check. In that case, could you please provide some more details about the expected usage of LLVM_ABI? For example, when should a class/struct be annotated, and what does it imply? (In that case, do we still need to annotate the member functions of that class, or not?)

My apologies, I was going by the suggested reviewers. I can ask someone else if you are not comfortable reviewing these changes.

Very basically, the `LLVM_ABI` annotations are used to mark the libLLVM dylib boundary. For Windows, we need to manually mark exported APIs with `__declspec(dllexport)` when they are exported and `__declspec(dllimport)` when they are imported. In general, we prefer marking a class as `dllexport`, in which case no individual methods get annotated - outside of static data members. However, marking a class as `dllexport` forces instantiation of all members, including the implicit copy ops, which break a number of classes. This is when we annotate the individual methods instead.

For context, we (Swift for Windows) want to build LLVM as a dynamic library and ship the toolchain this way. After discussions, we came to the conclusion that it was best to explicitly export every single LLVM API, you can check #109483 for details.

Please let me know how you would like to proceed from there.

fhahn (commented):
@Steelskin I think it was brought up previously. I think it would be helpful if there would be an automatic way to check if any annotations are missing. 

Ideally we would catch drift automatically, which would make fixing/updating/reviewing much easier

Steelskin:
> @Steelskin I think it was brought up previously. I think it would be helpful if there would be an automatic way to check if any annotations are missing.
> 
> Ideally we would catch drift automatically, which would make fixing/updating/reviewing much easier

@fhahn Yes, I am working on this. The PR message refers to a script I used to find these annotations and this is the result of that. One of the steps toward enabling the automatic workflow is to add all missing LLVM_ABI annotations on the existing headers, and this PR is a part of that. See this [comment](https://github.com/llvm/llvm-project/issues/109483#issuecomment-4398927211).

kasuga-fj:
> My apologies, I was going by the suggested reviewers. I can ask someone else if you are not comfortable reviewing these changes.

@Steelskin Sorry, I didn't mean to complain or anything. I was just a bit confused because I wasn't sure what I was supposed to do. It didn't seem like things would move forward on their own, so I left a comment yesterday.

And thanks for sharing the background. I don’t think I can comment on this topic from a technical perspective, but if it’s just simple work (e.g., manually checking whether everything is annotated properly), I can help with that. Just to confirm, would reading [this script](https://github.com/Steelskin/llvm-project/blob/f1462a75dcc09734d9c0efbe768c7d1fe06b3512/llvm/utils/git/ids-check-helper.py) help me to understand where the annotation should be attached?

Steelskin:
> > My apologies, I was going by the suggested reviewers. I can ask someone else if you are not comfortable reviewing these changes.
> 
> @Steelskin Sorry, I didn't mean to complain or anything. I was just a bit confused because I wasn't sure what I was supposed to do. It didn't seem like things would move forward on their own, so I left a comment yesterday.

No worries! Once again, if you don't feel comfortable, I can ask someone else who is more familiar with the dylib project to review. I just assumed people more familiar with this area of the codebase would want to check that I am not doing anything outright wrong.

> And thanks for sharing the background. I don’t think I can comment on this topic from a technical perspective, but if it’s just simple work (e.g., manually checking whether everything is annotated properly), I can help with that. Just to confirm, would reading [this script](https://github.com/Steelskin/llvm-project/blob/f1462a75dcc09734d9c0efbe768c7d1fe06b3512/llvm/utils/git/ids-check-helper.py) help me to understand where the annotation should be attached?

This is slightly out-of-date compared to the version I used but yes, it might help, as well as the corresponding [workflow](https://github.com/Steelskin/llvm-project/blob/f1462a75dcc09734d9c0efbe768c7d1fe06b3512/.github/workflows/ids-check.yml). But basically, this PR is what you'd get if you went over every header and added the annotations by hand.

kasuga-fj:
> I just assumed people more familiar with this area of the codebase would want to check that I am not doing anything outright wrong.

I see. Actually I'm also not an expert of this area (it just happened that I had a few more commits than usual recently), but I should at least be able to check whether you aren't doing anything obviously wrong (I believe...).

> This is slightly out-of-date compared to the version I used but yes, it might help, as well as the corresponding [workflow](https://github.com/Steelskin/llvm-project/blob/f1462a75dcc09734d9c0efbe768c7d1fe06b3512/.github/workflows/ids-check.yml). But basically, this PR is what you'd get if you went over every header and added the annotations by hand.

I see, thanks. If you're fine with this direction, I'll start going through all the files manually. As there are considerable amount of files, I think it will take some time. Are you okay with that? Also, if you find someone else who would be a more suitable reviewer, please feel free to forget about me.

Steelskin:
> I see, thanks. If you're fine with this direction, I'll start going through all the files manually. As there are considerable amount of files, I think it will take some time. Are you okay with that? Also, if you find someone else who would be a more suitable reviewer, please feel free to forget about me.

This is fine, I understand there is quite a bit to review, but these changes are mostly mechanical.

cc @nikic if you'd rather do the review yourself?

llvm/include/llvm/Analysis/DominanceFrontier.h
kasuga-fj: Don't they require `LLVM_TEMPLATE_ABI`?

llvm/include/llvm/Analysis/HashRecognize.h
```diff
@@ -28,7 +28,7 @@ class LPMUpdater;
 
 /// A custom std::array with 256 entries, that also has a print function.
 struct CRCTable : public std::array<APInt, 256> {
-  void print(raw_ostream &OS) const;
+  LLVM_ABI void print(raw_ostream &OS) const;
 
 #if !defined(NDEBUG) || defined(LLVM_ENABLE_DUMP)
   LLVM_DUMP_METHOD void dump() const;
```
kasuga-fj: What about `dump` methods annotated with `LLVM_DUMP_METHOD`?

llvm/include/llvm/Analysis/IR2Vec.h
kasuga-fj, line 547: Should this class be annotated with `LLVM_ABI` because it has virtual functions?

llvm/include/llvm/Analysis/RegionInfo.h
```diff
@@ -979,14 +983,14 @@ class RegionInfoPrinterPass
   raw_ostream &OS;
 
 public:
-  explicit RegionInfoPrinterPass(raw_ostream &OS);
+  LLVM_ABI explicit RegionInfoPrinterPass(raw_ostream &OS);
 
   PreservedAnalyses run(Function &F, FunctionAnalysisManager &AM);
```
kasuga-fj: Maybe missed?

llvm/include/llvm/Analysis/ScalarEvolution.h
kasuga-fj, line 383: Just to confirm: this class has two types of virtual methods: pure virtual function and one which has inlined default implementation. In such a case, this class isn't needed to be annotated with `LLVM_ABI` even though it has virtual functions, right?

kasuga-fj (commented):
I manually checked all the modified files while referring to https://llvm.org/docs/InterfaceExportAnnotations.html, and overall it looks good to me. I've left a few comments where an annotation might be missing, but I'm fairly sure I'm just misunderstanding something...

llvm/include/llvm/Analysis/ScalarEvolution.h
Steelskin: The rule of thumb for the class-level annotation is: annotate the class iff it has at least one out-of-line virtual method. In this case, the class has a vtable anchor that lives inside a TU, so the compiler emits the vtable in the producer TU and imports it anywhere else.

So yes, we don't need the class-level annotation here.

llvm/include/llvm/Analysis/DominanceFrontier.h
Steelskin: Yes, the script does not actually handle the template annotations so we're exporting these "as needed", meaning if anything outside of LLVM actually uses them. In an ideal world, these would also be automatically annotated but this is harder than the rest of the annotations.
Thanks for pointing it out, I added the annotation for these.

llvm/include/llvm/Analysis/HashRecognize.h
Steelskin: Yes, this was missed because the script runs on a release build.

llvm/include/llvm/Analysis/IR2Vec.h
Steelskin: All of the virtual functions for this class are either pure virtual or embedded inline so no.

llvm/include/llvm/Analysis/RegionInfo.h
Steelskin: Yes, my pass on the headers is 2 weeks old at this point and there were some changes made to the headers. I will do another full pass once I have landed this slate of changes and before re-enabling the workflow to automatically add missing annotations.

kasuga-fj (approved):
Thanks for all the explanation, LGTM.

PR #199022, opened 2026-05-21
[CI] Pass script args using env var in merged PRs workflow

DavidSpickett:
As was done in #198160, address the problem described in
https://docs.github.com/en/actions/concepts/security/script-injections using the solution recommended by
https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable.

Not all these inputs are untrusted, but I've applied it to all of them just to be consistent.

### DIFF
--- a/.github/workflows/merged-prs.yml
+++ b/.github/workflows/merged-prs.yml
@@ -40,7 +40,7 @@ jobs:
           PR_AUTHOR: ${{ github.event.pull_request.user.login }}
         run: |
           python3 ./github-automation.py \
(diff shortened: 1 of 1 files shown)


boomanaiden154 (approved):
LGTM.

boomanaiden154:
Thanks for the fix. When we enabled zizmor I tried to get the person driving it to ensure that the current repo state was clean before we enabled it by default, but apparently we did not do a very good job at that.

banach-space (approved):
LGTM, thanks!

PR #199024, opened 2026-05-21
[CI] Remove unused env var from commit-access-greeter

DavidSpickett:
LABEL_NAME has been there since the workflow's introduction in f8ef2699d860aea97750953f1b79db8ef7574e82, but has never been used.


boomanaiden154 (approved):
LGTM.

PR #199026, opened 2026-05-21
[CI] Pass label name to script using env var in pr-subscriber workflow

DavidSpickett:
As is done in issue-subscriber.yml already.

This addresses the problem described in
https://docs.github.com/en/actions/concepts/security/script-injections using the solution recommended by
https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable.


boomanaiden154 (approved):
LGTM.

boomanaiden154:
Still makes sense to fix, but I think people need commit access to create new labels, so not a big security concern.

PR #199028, opened 2026-05-21
[lldb][windows] Null-check m_session_data in ProcessDebugger::OnDebuggerError

charles-zablit:
Add a null check that logs and returns, matching the pattern in OnExitProcess.


DavidSpickett (approved):
LGTM

PR #128113, opened 2025-02-21
[mlir][affine] Use value bound inference to determine minimum/maximum trip counts in loop analysis

linuxlonelyeagle:
When ValueBoundOpInterface is utilized for both the lower and upper bounds of a loop, the loop cannot be precisely analyzed. To address this limitation, this PR introduces ValueBoundsConstraintSet analysis into the loop analysis infrastructure.

PR #199032, opened 2026-05-21
[MLIR][Affine] Simplify affine.for bounds by pruning redundant expressions via ValueBoundsConstraintSet

linuxlonelyeagle:
This PR introduces a new pattern `SimplifyAffineForBoundMap` to simplify multi-result lower and upper bounds of `affine.for` loops by pruning redundant expressions leveraging `ValueBoundsConstraintSet`. This PR addresses the limitation where the built-in folder of `affine.for` can only handle constant bounds.

### DIFF
--- a/mlir/lib/Dialect/Affine/Transforms/SimplifyAffineWithBounds.cpp
+++ b/mlir/lib/Dialect/Affine/Transforms/SimplifyAffineWithBounds.cpp
@@ -236,7 +236,6 @@ simplifyRedundantMapResults(AffineMap map, SmallVector<Value> operands,
 
   for (int i = mapResults.size() - 1; i >= 0; --i) {
     AffineExpr a = mapResults[i];
-
     AffineMap mapA =
         AffineMap::get(map.getNumDims(), map.getNumSymbols(), a, context);
     ValueBoundsConstraintSet::Variable varA(mapA, operands);
--- a/mlir/lib/Dialect/Affine/Transforms/SimplifyAffineWithBounds.cpp
+++ b/mlir/lib/Dialect/Affine/Transforms/SimplifyAffineWithBounds.cpp
@@ -218,11 +218,106 @@ struct SimplifyDelinearizeOfLinearizeDisjoint final
   }
 };
 
+/// Simplifies the affine map results by eliminating redundant expressions.
+///
+/// This function performs a pairwise comparison of all expressions in the map
+/// using the analysis from `ValueBoundsConstraintSet`. If an expression `a` is
+/// statically proven to be strictly bounded or covered by another expression
+/// `b` (based on the given comparison operator `cmp`), `a` is considered
+/// redundant and is safely pruned from the results. NOTE: We iterate backwards
+/// (from size-1 down to 0) to safely erase elements from the `SmallVector`
+/// without causing iterator invalidation or indexing shifts for upcoming
+/// elements.
+static SmallVector<AffineExpr>
+simplifyRedundantMapResults(AffineMap map, SmallVector<Value> operands,
+                            ValueBoundsConstraintSet::ComparisonOperator cmp) {
+  SmallVector<AffineExpr> mapResults(map.getResults());
+  auto *context = map.getContext();
+
+  for (int i = mapResults.size() - 1; i >= 0; --i) {
+    AffineExpr a = mapResults[i];
+
+    AffineMap mapA =
+        AffineMap::get(map.getNumDims(), map.getNumSymbols(), a, context);
+    ValueBoundsConstraintSet::Variable varA(mapA, operands);
+    bool shouldErase = false;
+
+    for (int j = 0, e = mapResults.size(); j < e; ++j) {
+      if (i == j)
+        continue;
+
+      AffineExpr b = mapResults[j];
+      AffineMap mapB =
+          AffineMap::get(map.getNumDims(), map.getNumSymbols(), b, context);
+      ValueBoundsConstraintSet::Variable varB(mapB, operands);
+
+      if (ValueBoundsConstraintSet::compare(varB, cmp, varA)) {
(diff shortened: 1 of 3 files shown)


linuxlonelyeagle:
The initial idea came from  https://github.com/llvm/llvm-project/pull/128113#issuecomment-2801043226. "So if you two expressions with ranges [4, 6] and [2, 3] (non-overlapping), the expressions with [4, 6] will never affect the bound. We can even remove it as an optimization."

krzysz00 (approved):
I think the pattern itself is correct, though I'd update the PR description to clarify that this isn't a canonicalization pattern

mlir/lib/Dialect/Affine/Transforms/SimplifyAffineWithBounds.cpp
```diff
@@ -218,11 +218,105 @@ struct SimplifyDelinearizeOfLinearizeDisjoint final
...
+/// This function performs a pairwise comparison of all expressions in the map
+/// using the analysis from `ValueBoundsConstraintSet`. If an expression `a` is
+/// statically proven to be strictly bounded or covered by another expression
+/// `b` (based on the given comparison operator `cmp`), `a` is considered
+/// redundant and is safely pruned from the results. NOTE: We iterate backwards
+/// (from size-1 down to 0) to safely erase elements from the `SmallVector`
+/// without causing iterator invalidation or indexing shifts for upcoming
+/// elements.
+static SmallVector<AffineExpr>
+simplifyRedundantMapResults(AffineMap map, SmallVector<Value> operands,
```
ftynse: Absoultely never have functions take a `SmallVector`. They should take a `SmallVectorImpl &`, a `MutableArrayRef` or an `ArrayRef`. In case of values, we have `ValueRange`.

mlir/lib/Dialect/Affine/Transforms/SimplifyAffineWithBounds.cpp
```diff
@@ -218,11 +218,105 @@ struct SimplifyDelinearizeOfLinearizeDisjoint final
...
+                            ValueBoundsConstraintSet::ComparisonOperator cmp) {
+  SmallVector<AffineExpr> mapResults(map.getResults());
+  auto *context = map.getContext();
```
ftynse: Almost always avoid auto. Acceptable uses are casts, constructors, iterators and lambdas.

mlir/lib/Dialect/Affine/Transforms/SimplifyAffineWithBounds.cpp
```diff
@@ -218,11 +218,105 @@ struct SimplifyDelinearizeOfLinearizeDisjoint final
...
+/// Simplifies the affine map results by eliminating redundant expressions.
+///
+/// This function performs a pairwise comparison of all expressions in the map
+/// using the analysis from `ValueBoundsConstraintSet`. If an expression `a` is
+/// statically proven to be strictly bounded or covered by another expression
+/// `b` (based on the given comparison operator `cmp`), `a` is considered
+/// redundant and is safely pruned from the results. NOTE: We iterate backwards
+/// (from size-1 down to 0) to safely erase elements from the `SmallVector`
+/// without causing iterator invalidation or indexing shifts for upcoming
+/// elements.
```
ftynse: The NOTE is an implementation detail and doesn't belong in the API documentation. It may be a simple in-line comment in the function.

mlir/lib/Dialect/Affine/Transforms/SimplifyAffineWithBounds.cpp
```diff
@@ -218,11 +218,105 @@ struct SimplifyDelinearizeOfLinearizeDisjoint final
...
+      ValueBoundsConstraintSet::Variable varB(mapB, operands);
+
+      if (ValueBoundsConstraintSet::compare(varB, cmp, varA)) {
+        shouldErase = true;
+        break;
+      }
+    }
+
+    if (shouldErase)
+      mapResults.erase(mapResults.begin() + i);
```
ftynse: A better approach is to iterate over the map (use getSubMap to produce single-expression maps) and collect indices at which to earse the expressions. Then create a new vector of expressions and copy those expressions whose index wasn't marked for erasure. This avoids additional copies of elements every time an element is earsed.

mlir/lib/Dialect/Affine/Transforms/SimplifyAffineWithBounds.cpp
```diff
@@ -218,11 +218,105 @@ struct SimplifyDelinearizeOfLinearizeDisjoint final
...
+  }
+
+  return mapResults;
+}
+
+/// A canonicalization pattern that simplifies multi-result lower and upper
```
ftynse: This is not a canonicalization pattern.

mlir/lib/Dialect/Affine/Transforms/SimplifyAffineWithBounds.cpp
```diff
@@ -218,11 +218,105 @@ struct SimplifyDelinearizeOfLinearizeDisjoint final
...
+    bool upperBoundUpdate =
+        upperMapExprs.size() < upperBoundMap.getNumResults();
+    if (!(lowerBoundUpdate || upperBoundUpdate))
+      return failure();
+
+    auto *context = forOp->getContext();
+    if (lowerBoundUpdate) {
+      AffineMap newMap =
+          AffineMap::get(lowerBoundMap.getNumDims(),
+                         lowerBoundMap.getNumSymbols(), lowerMapExprs, context);
```
ftynse: Just construct the new map in the helper function, no need to return a vector from it.

linuxlonelyeagle:
Ping for review

ftynse (approved):

PR #199034, opened 2026-05-21
[NFC][mlir][bufferization] Remove getMemRefType() helper

andrey-golubev:
Replace getMemRefType() helper with a direct usage of options.unknownTypeConverterFn() hook. This does not change any behaviour since all the existing call-sites would already implicitly call the hook. The major difference could have been in the handling of the memref layout but this does not seem to be properly handled anyhow.

As the `getMemRefType()` helper is removed, there are two cases to keep in mind for downstream users that need to adjust the code accordingly:
1. For creating memrefs without layouts, call `options.unknownTypeConverterFn` hook directly (the same way this patch is doing for upstream)
2. For creating memrefs with layouts (assuming, custom ones), please (re)implement the same helper. As only ranked memrefs can have a layout, a call to `mlir::MemRefType::get(shape, elementType, /*your custom layout=*/myLayout, memorySpace)` should be sufficient.

### DIFF
--- a/mlir/include/mlir/Dialect/Bufferization/IR/BufferizableOpInterface.h
+++ b/mlir/include/mlir/Dialect/Bufferization/IR/BufferizableOpInterface.h
@@ -660,23 +660,6 @@ OpTy replaceOpWithNewBufferizedOp(RewriterBase &rewriter, Operation *op,
   return newOp;
 }
 
-/// Return a MemRefType to which the TensorType can be bufferized.
-///
-/// If possible, op bufferization implementations should not use this function
-/// and instead infer precise memref types for tensor results by themselves.
-///
-/// Unless a layout map was specified, `options.unknownTypeConverterFn`
-/// determines what kind of layout map will be used. For best composability
-/// (without copies), the fully dynamic layout map is used by default.
-///
-/// Note: Canonicalization patterns could clean up layout maps and infer more
-/// precise layout maps after bufferization. However, many possible
-/// canonicalizations are currently not implemented.
-BaseMemRefType getMemRefType(TensorType tensorType,
-                             const BufferizationOptions &options,
-                             MemRefLayoutAttrInterface layout = {},
-                             Attribute memorySpace = nullptr);
-
 /// Return a MemRef type with fully dynamic layout. If the given tensor type
 /// is unranked, return an unranked MemRef type.
 BaseMemRefType
--- a/mlir/lib/Dialect/Bufferization/IR/BufferizableOpInterface.cpp
+++ b/mlir/lib/Dialect/Bufferization/IR/BufferizableOpInterface.cpp
@@ -812,29 +812,6 @@ LogicalResult BufferizationOptions::createMemCpy(OpBuilder &b, Location loc,
 // Bufferization-specific IRMapping support with debugging.
 //===----------------------------------------------------------------------===//
 
-BaseMemRefType bufferization::getMemRefType(TensorType tensorType,
-                                            const BufferizationOptions &options,
-                                            MemRefLayoutAttrInterface layout,
-                                            Attribute memorySpace) {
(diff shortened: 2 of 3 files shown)


andrey-golubev:
I guess we can start with this first and foremost.

mlir/include/mlir/Dialect/Bufferization/IR/BufferizableOpInterface.h
```diff
@@ -660,23 +660,6 @@ OpTy replaceOpWithNewBufferizedOp(RewriterBase &rewriter, Operation *op,
...
-/// and instead infer precise memref types for tensor results by themselves.
-///
-/// Unless a layout map was specified, `options.unknownTypeConverterFn`
-/// determines what kind of layout map will be used. For best composability
-/// (without copies), the fully dynamic layout map is used by default.
-///
-/// Note: Canonicalization patterns could clean up layout maps and infer more
-/// precise layout maps after bufferization. However, many possible
-/// canonicalizations are currently not implemented.
-BaseMemRefType getMemRefType(TensorType tensorType,
```
matthias-springer, line 675: Can we keep this function? Some downstream projects (e.g. IREE) are using it. Alternatively, if we remove it, mention in the commit message how to fix the code.

mlir/lib/Dialect/Bufferization/IR/BufferizableOpInterface.cpp
```diff
@@ -1006,7 +983,7 @@ FailureOr<BufferLikeType> bufferization::detail::defaultGetBufferType(
     return op->emitError("could not infer memory space");
 
   return cast<BufferLikeType>(
-      getMemRefType(tensorType, options, /*layout=*/{}, *memSpace));
+      options.unknownTypeConverterFn(tensorType, *memSpace, options));
```
matthias-springer, line 986: Let's update this function:
```
/// Default unknown type converter: Use a fully dynamic layout map.
BaseMemRefType
defaultUnknownTypeConverter(TensorType tensorType, Attribute memorySpace,
                            const BufferizationOptions &options) {
  return getMemRefTypeWithFullyDynamicLayout(tensorType, memorySpace);
}
```

It should return an unranked memref if the tensor is unranked.

Same for this in `Bufferize.cpp`:
```
      opt.unknownTypeConverterFn = [=](TensorType tensorType,
                                       Attribute memorySpace,
                                       const BufferizationOptions &options) {
```

andrey-golubev: what do you mean by "update"?

regarding,

> It should return an unranked memref if the tensor is unranked.

I believe this is already the case (see the first if-statement in `getMemRefTypeWithFullyDynamicLayout`):
https://github.com/llvm/llvm-project/blob/39658a8fa9d1900f90f218a5eeeb205c227e1262/mlir/lib/Dialect/Bufferization/IR/BufferizableOpInterface.cpp#L838-L858

I imagine the way it worked before was:
* getMemRefType() would return-early for unranked case
* unknown type converter had the same code path for unranked case

As I removed the getMemRefType() part, we still have the coverage for unranked within default unknown type conversion, as well as within the lambda defined in Bufferize.cpp (second case) -- `getMemRefTypeWithStaticIdentityLayout()` also handles unranked case.

matthias-springer: Oh, I missed that. Then no change is needed here.

mlir/include/mlir/Dialect/Bufferization/IR/BufferizableOpInterface.h
andrey-golubev: updated the PR description, would it be sufficient?

matthias-springer (approved):

PR #199039, opened 2026-05-21
[NFC][lldb][windows] Replace C-style casts with static_cast / reinterpret_cast

charles-zablit:


DavidSpickett (approved):
Thought about whether `0xf....fULL` would be clearer in some spots, but this casting of -1 is an established pattern anyway so readers will have to learn it one way or another.

LGTM.

lldb/include/lldb/Host/windows/PseudoConsole.h
```diff
@@ -128,12 +128,14 @@ class PseudoConsole {
...
+  HANDLE m_conpty_handle = reinterpret_cast<HANDLE>(static_cast<intptr_t>(-1));
+  HANDLE m_conpty_output = reinterpret_cast<HANDLE>(static_cast<intptr_t>(-1));
+  HANDLE m_conpty_input = reinterpret_cast<HANDLE>(static_cast<intptr_t>(-1));
   // Pipe mode: child-side handles passed to CreateProcessW, closed after launch
-  HANDLE m_pipe_child_stdin = ((HANDLE)(long long)-1);
-  HANDLE m_pipe_child_stdout = ((HANDLE)(long long)-1);
+  HANDLE m_pipe_child_stdin =
+      reinterpret_cast<HANDLE>(static_cast<intptr_t>(-1));
+  HANDLE m_pipe_child_stdout =
+      reinterpret_cast<HANDLE>(static_cast<intptr_t>(-1));
```
DavidSpickett, line 138: I'm assuming this only runs on the target machine, because intptr_t will change size based on that.

DavidSpickett: long long does too but I don't feel like remembering exactly how :)

Nerixyz: Yes, this only runs on the host. I think `intptr_t` is more correct than `long long` here, because it should be a typedef to `long long` on 64 bit and `long`/`int` on 32 bit.

Nerixyz (approved):

lldb/include/lldb/Host/windows/PseudoConsole.h
DavidSpickett: Understood, thanks. inptr_t is more clear, I agree.

JDevlieghere (approved):

PR #199040, opened 2026-05-21
[AArch64][GlobalISel] Skip RegBankSelect pass at -O0

c-rhodes:
Profiling of sqlite on aarch64-O0-g shows ~2.60% of compile-time is spent in RegBankSelect. Optimizing cross-register bank copies isn't necessary at O0, skipping this pass and selecting register banks before instruction selection is -a 1.20% geomean improvement on CTMark. sqlite is -1.72%.

https://llvm-compile-time-tracker.com/compare.php?from=376870209db2e3ad33b082f839a7899c2972cc93&to=8fb8909b7071bd7fb959ddfd8a8863736e83dd80&stat=instructions%3Au

Assisted-by: codex

### DIFF
--- a/llvm/include/llvm/CodeGen/GlobalISel/InstructionSelect.h
+++ b/llvm/include/llvm/CodeGen/GlobalISel/InstructionSelect.h
@@ -41,10 +41,11 @@ public:
   void getAnalysisUsage(AnalysisUsage &AU) const override;
 
   MachineFunctionProperties getRequiredProperties() const override {
-    return MachineFunctionProperties()
-        .setIsSSA()
-        .setLegalized()
-        .setRegBankSelected();
+    MachineFunctionProperties Properties;
+    Properties.setIsSSA().setLegalized();
+    if (RequireRegBankSelected)
+      Properties.setRegBankSelected();
+    return Properties;
   }
 
   MachineFunctionProperties getSetProperties() const override {
@@ -52,7 +53,7 @@ public:
   }
 
   InstructionSelect(CodeGenOptLevel OL = CodeGenOptLevel::Default,
-                    char &PassID = ID);
+                    bool RequireRegBankSelected = true, char &PassID = ID);
 
   bool runOnMachineFunction(MachineFunction &MF) override;
   bool selectMachineFunction(MachineFunction &MF);
@@ -67,6 +68,7 @@ protected:
   ProfileSummaryInfo *PSI = nullptr;
 
   CodeGenOptLevel OptLevel = CodeGenOptLevel::None;
+  bool RequireRegBankSelected = true;
 
   bool selectInstr(MachineInstr &MI);
 };
--- a/llvm/lib/CodeGen/GlobalISel/InstructionSelect.cpp
+++ b/llvm/lib/CodeGen/GlobalISel/InstructionSelect.cpp
@@ -63,8 +63,10 @@ INITIALIZE_PASS_END(InstructionSelect, DEBUG_TYPE,
                     "Select target instructions out of generic instructions",
                     false, false)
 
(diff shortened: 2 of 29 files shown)


arsenm (changes requested):
I think the correct replacement is to have a trivial regbankselect replacement pass rather than deferring the assignment to the selector, and introducing a new edge case where regbanks aren't set coming into selection. The trivial pass could amount to memset

aemerson:
> I think the correct replacement is to have a trivial regbankselect replacement pass rather than deferring the assignment to the selector, and introducing a new edge case where regbanks aren't set coming into selection. The trivial pass could amount to memset 

Agreed. This was a known cost that we paid willingly for the sake of our overall design.

davemgreen (commented):
1.2% is a fairly decent gain. I had something slightly different in mind but I'm not sure how much it matters. My though was that if you have a separate pass or iteration over all the instructions then by the time we get to the end of the function you have thrown away all the cache-locality that is important to make it truly efficient. I was instead thinking of performing the reg bank selection close to the actual selection of the instruction, so the accesses can remain close in memory.

I'm not sure how possible this is, and like I said I'm not sure how much it actually matters. It might help more with larger functions like sqlite, but in general for O0 functions might usually be quite small anyway. And cache misses are less likely to show up on the instruction count that is usually used for compile time tracking.

Without doing that it could probably just be a separate but simpler pass. But make sure it is still handling all the instructions it can, we should avoid falling back to sdag. The more we can share with the proper RBS the better and from my understanding there are a number of integer intrinsics we need to end up on fpr.

c-rhodes:
thanks for comments all, I will experiment with doing this as a standalone pass

c-rhodes:
I've tried a standalone "RegBankSelectFast" pass: https://github.com/llvm/llvm-project/compare/main...c-rhodes:llvm-project:perf/aarch64-gisel-regbankselectfast-pass

it seemed simpler to have a fallback to full regbankselect. Compile-time results are here: https://llvm-compile-time-tracker.com/compare.php?from=ea8f3dfe487e1c0df1fbd31dded17ec77567229e&to=7b8ae1044d71028e21e4f543dd0c91ed141887d3&stat=instructions%3Au

sqlite is about the same improvement as this, but geomean -0.69% vs 1.20% for this patch. The target specific `trySimpleRegBankSelect` hook is a significant part of that improvement (https://llvm-compile-time-tracker.com/compare.php?from=69361d3fd8c88f4dc40c7ecdaa2791663456e4ca&to=7b8ae1044d71028e21e4f543dd0c91ed141887d3&stat=instructions%3Au).

Although the same could be done with the existing RegBankSelect pass by implementing a fast `assignInstr`: https://github.com/c-rhodes/llvm-project/commit/35040a8aebc30213bd64494d41a1bd2541eaef6b

that's an even simpler change and geomean is -0.77%: https://llvm-compile-time-tracker.com/compare.php?from=ea8f3dfe487e1c0df1fbd31dded17ec77567229e&to=35040a8aebc30213bd64494d41a1bd2541eaef6b&stat=instructions%3Au

c-rhodes:
I prototyped a standalone pure type-based RBS pass: https://github.com/llvm/llvm-project/commit/a88bdb485bec5a92f217dcd49a7644b0da8d3839#diff-d68647a0638820a862a533c2f5d119ddea1e144b8b1dcc44af6b96e5479b741c

the compile-time results are the best so far: https://llvm-compile-time-tracker.com/compare.php?from=2de2edb943fe1b83d79bdffa03606eb8c5452e9b&to=a88bdb485bec5a92f217dcd49a7644b0da8d3839&stat=instructions%3Au

geomean -1.63% sqlite -3.59%

It does require extending instruction selection to handle for a few things like narrow GPR constants that RBS is currently handling, but I think the compile-time results are compelling enough this is worth pursuing.

PR #199041, opened 2026-05-21
[VPlan] Rename Expression::isSingleScalar (NFC)

artagnon:
The single-scalar terminology, as it is used in other places indicates that all operands are scalars and that the result is a scalar. VPExpressionRecipe::isSingleScalar is a misnomer, and is actually a vector-to-scalar, using the existing terminology. Rename it for clarity.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlan.h
+++ b/llvm/lib/Transforms/Vectorize/VPlan.h
@@ -3555,8 +3555,8 @@ public:
   /// effects.
   bool mayHaveSideEffects() const;
 
-  /// Returns true if the result of this VPExpressionRecipe is a single-scalar.
-  bool isSingleScalar() const;
+  /// Returns true if this VPExpressionRecipe produces a scalar.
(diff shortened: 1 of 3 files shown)


llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -3538,8 +3538,8 @@ class VPExpressionRecipe : public VPSingleDefRecipe {
   /// effects.
   bool mayHaveSideEffects() const;
 
-  /// Returns true if the result of this VPExpressionRecipe is a single-scalar.
-  bool isSingleScalar() const;
+  /// Returns true if this VPExpressionRecipe produces a scalar.
```
fhahn: nit: may be worth stressing it produces a single scalar value
```suggestion
  /// Returns true if this VPExpressionRecipe produces a single scalar.
```

fhahn (approved):
LGTM, much clearer thanks!

PR #199042, opened 2026-05-21
llvm: Fix most LLVM_ABI annotations in Transforms

Steelskin:
This updates most LLVM_ABI annotations in the Transforms headers to match expected usage:
* All public APIs should be properly annotated.
* Inlined functions should not be annotated.

These changes were done by a script fixing annotations on LLVM public headers and manually checked.

This effort is tracked in #109483.


nikic (approved):
LGTM

PR #199045, opened 2026-05-21
[CIR] Implement lowering of atomic-func ptrs

erichkeane:
This actually is a near-zero effort implementation, since other than an assertion, we had everything correct. It isn't clear where that assertion came from (other than an over abundance of caution?).  It doesn't exist in classic codegen, which always counts on the fallthrough.

However, we ARE missing some ptr-auth work in that area, so a MissingFeatures is added to be a placeholder for it.


clang/test/CIR/CodeGen/call.c
```diff
@@ -130,6 +130,26 @@ int f12(void) {
...
+
+// LLVM-LABEL: define{{.*}} void @f13()
+// LLVM: %[[FP_ALLOCA:.*]] = alloca ptr
+// LLVM: %[[LOAD_FP:.*]] = load atomic ptr, ptr %[[FP_ALLOCA]] seq_cst, align 8
+// LLVM: call void %[[LOAD_FP]]()
+
+// OGCG-LABEL: define{{.*}} void @f13() 
+// OGCG: %[[FP_ALLOCA:.*]] = alloca ptr
+// OGCG: %[[LOAD_FP:.*]] = load atomic ptr, ptr %[[FP_ALLOCA]] seq_cst, align 8
+// OGCG: call void %[[LOAD_FP]]()
```
xlauko, line 150: any reason to keep OGCG and LLVM apart here? or is it because rest of the file is kept apart?

xlauko (approved):
lgtm % nit

clang/test/CIR/CodeGen/call.c
erichkeane: Exactly, the rest of the file has them separate, so joining them ends up being a touch of a hassle?  if it helps, they were a copy and paste of eachother :) 

At one point, someone should go through all the OGCG tests and see how many we can just convert to not use OGCG. 

When we started, they were so dissimilar that it was necessary to keep them apart, but we're getting closer to the point where IMO it would help to convert whole files.

andykaylor (approved):

PR #199046, opened 2026-05-21
[lldb][windows] relax TestMissingDll.py assert

charles-zablit:
The lldb-server path wraps the underlying error with `Cannot launch '<path>': `, while the in-process ProcessWindows plugin returns the bare error string. Relax the assert so that the test passes with or without `LLDB_USE_LLDB_SERVER=1`.

rdar://177623653


DavidSpickett (approved):
LGTM

PR #199047, opened 2026-05-21
[VPlan] Introduce vputils::doesGeneratePerAllLanes

artagnon:
Introduce a query for recipes that produce scalar values for all VF lanes, returning true for non-single-scalar replicates, non-first-lane SIVSteps, and a couple of VPInstructions.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlan.h
+++ b/llvm/lib/Transforms/Vectorize/VPlan.h
@@ -1348,12 +1348,7 @@ public:
     OpsEnd = Intrinsic,
   };
 
-  /// Returns true if this VPInstruction generates scalar values for all lanes.
-  /// Most VPInstructions generate a single value per part, either vector or
-  /// scalar. VPReplicateRecipe takes care of generating multiple (scalar)
-  /// values per all lanes, stemming from an original ingredient. This method
-  /// identifies the (rare) cases of VPInstructions that do so as well, w/o an
-  /// underlying ingredient.
+  /// Returns true if this recipe produces scalar values for all VF lanes.
   bool doesGeneratePerAllLanes() const;
 
   /// Return the number of operands determined by the opcode of the
@@ -1510,8 +1505,7 @@ public:
   /// e.g. by performing a reduction or extracting a lane.
   bool isVectorToScalar() const;
 
-  /// Returns true if this VPInstruction's operands are single scalars and the
-  /// result is also a single scalar.
+  /// Returns true if the recipe produces a single scalar value.
   bool isSingleScalar() const;
 
   /// Returns the symbolic name assigned to the VPInstruction.
@@ -3432,7 +3426,11 @@ public:
                                          bool IsSingleScalar, ElementCount VF,
                                          VPCostContext &Ctx);
 
+  /// Returns true if the recipe produces a single scalar value.
   bool isSingleScalar() const { return IsSingleScalar; }
+
+  /// Returns true if the recipe produces scalar values for all VF lanes.
+  bool doesGeneratePerAllLanes() const { return !IsSingleScalar; }
 
   bool isPredicated() const { return IsPredicated; }
 
@@ -4266,6 +4264,9 @@ public:
     else
       addOperand(StartIndex);
   }
+
+  /// Returns true if this recipe produces scalar values for all VF lanes.
+  bool doesGeneratePerAllLanes() const;
 
   /// Returns true if the recipe only uses the first lane of operand \p Op.
   bool usesFirstLaneOnly(const VPValue *Op) const override {
--- a/llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
+++ b/llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
@@ -690,7 +690,8 @@ unsigned VPInstruction::getNumOperandsForOpcode() const {
 }
 
 bool VPInstruction::doesGeneratePerAllLanes() const {
-  return Opcode == VPInstruction::PtrAdd && !vputils::onlyFirstLaneUsed(this);
+  return Opcode == VPInstruction::Unpack ||
+         (Opcode == VPInstruction::PtrAdd && !vputils::onlyFirstLaneUsed(this));
 }
 
 bool VPInstruction::canGenerateScalarForFirstLane() const {
@@ -3046,6 +3047,10 @@ void VPDerivedIVRecipe::printRecipe(raw_ostream &O, const Twine &Indent,
   getStepValue()->printAsOperand(O, SlotTracker);
 }
 #endif
+
+bool VPScalarIVStepsRecipe::doesGeneratePerAllLanes() const {
+  return !vputils::onlyFirstLaneUsed(this);
+}
 
 InstructionCost VPScalarIVStepsRecipe::computeCost(ElementCount VF,
                                                    VPCostContext &Ctx) const {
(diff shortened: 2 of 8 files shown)


llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -4946,19 +4946,17 @@ void VPlanTransforms::materializePacksAndUnpacks(VPlan &Plan) {
   for (VPBasicBlock *VPBB :
        concat<VPBasicBlock *>(VPBBsOutsideLoopRegion, VPBBsInsideLoopRegion)) {
     for (VPRecipeBase &R : make_early_inc_range(*VPBB)) {
-      if (!isa<VPScalarIVStepsRecipe, VPReplicateRecipe, VPInstruction>(&R))
+      if (!TypeSwitch<VPRecipeBase *, bool>(&R)
+               .Case<VPScalarIVStepsRecipe, VPReplicateRecipe, VPInstruction>(
+                   [](auto *R) { return R->mustReplicate(); })
+               .Default([](auto *) { return false; }))
```
fhahn: Would be good to sink the logic into a `vputils::` helper, to avoid type switch in the callers? First step could also be independent of adding `mustReplicate`

fhahn (commented):
Is there a reason for closing? Having a utility function to consolidate the logic to query if a recipe is replicating seems like a good first step.

artagnon:
> Is there a reason for closing? Having a utility function to consolidate the logic to query if a recipe is replicating seems like a good first step.

Ah, I wasn't sure if this would be preferred or a more minimal https://github.com/llvm/llvm-project/pull/199750.

artagnon:
Gentle ping.

artagnon:
Gentle ping.

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -1350,13 +1350,10 @@ class LLVM_ABI_FOR_TEST VPInstruction : public VPRecipeWithIRFlags,
...
   };
 
-  /// Returns true if this VPInstruction generates scalar values for all lanes.
-  /// Most VPInstructions generate a single value per part, either vector or
-  /// scalar. VPReplicateRecipe takes care of generating multiple (scalar)
-  /// values per all lanes, stemming from an original ingredient. This method
-  /// identifies the (rare) cases of VPInstructions that do so as well, w/o an
-  /// underlying ingredient.
-  bool doesGeneratePerAllLanes() const;
+  /// Returns true if this VPInstruction replicates for each part, and generates
```
david-arm: By `for each part` do you mean `for each unrolled part`? If it's replication per lane, then it's probably better to say `for each lane`, because the loop vectoriser uses `part` to mean `unrolled part`.

llvm/lib/Transforms/Vectorize/VPlanRecipes.cpp
```diff
@@ -626,8 +626,9 @@ unsigned VPInstruction::getNumOperandsForOpcode() const {
   llvm_unreachable("all cases should be handled above");
 }
 
-bool VPInstruction::doesGeneratePerAllLanes() const {
-  return Opcode == VPInstruction::PtrAdd && !vputils::onlyFirstLaneUsed(this);
+bool VPInstruction::mustReplicate() const {
+  return Opcode == VPInstruction::Unpack ||
```
david-arm, line 693: To be honest, I'm a bit sceptical that this is really NFC, although no tests have changed. That's because `materializePacksAndUnpacks` wasn't checking for `VPInstruction::Unpack`, whereas `replicateByVF` was. By introducing the same function `mustReplicate` in both callsites we're extending the functionality of `materializePacksAndUnpacks`.

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -5196,19 +5196,14 @@ void VPlanTransforms::materializePacksAndUnpacks(VPlan &Plan) {
...
       auto *DefR = cast<VPSingleDefRecipe>(&R);
       auto UsesVectorOrInsideReplicateRegion = [DefR, LoopRegion](VPUser *U) {
         VPRegionBlock *ParentRegion = cast<VPRecipeBase>(U)->getRegion();
         return !U->usesScalars(DefR) || ParentRegion != LoopRegion;
       };
-      if ((isa<VPReplicateRecipe>(DefR) &&
-           cast<VPReplicateRecipe>(DefR)->isSingleScalar()) ||
-          (isa<VPInstruction>(DefR) &&
-           (vputils::onlyFirstLaneUsed(DefR) ||
-            !cast<VPInstruction>(DefR)->doesGeneratePerAllLanes())) ||
```
david-arm, line 5512: I think we now get here for `VPInstruction::Unpack` instructions too, where we didn't previously. Can we add a test for this? If it should be impossible to get here for Unpacks then it might be worth adding an assert.

david-arm (commented):
Seems like a nice clean up!

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
artagnon: Um, this function is materializePacksAndUnpacks, which creates the Unpacks in the first place? Not sure it's worth having an assert here?

david-arm: Oh ok. I wasn't sure where the Unpack instruction was being created. Now I can see it's created in the second for loop so it should be impossible to encounter it here.

david-arm (approved):
LGTM!

llvm/lib/Transforms/Vectorize/VPlanUtils.h
```diff
@@ -64,6 +64,10 @@ bool isHeaderMask(const VPValue *V, const VPlan &Plan);
 /// VPDerivedIV or the canonical IV).
 bool isUniformAcrossVFsAndUFs(const VPValue *V);
 
+/// Returns true if \p R must be replicated, to produce scalar values for all
```
fhahn: I am not sure about `must` here, as we use this for recipes that are already guaranteed to generate values for all VF lanes. Although it is slightly longer, `doesGeneratePerAllLanes` as in the original helper in VPInstruction may be slightly clearer/more generic .

It may also be good to stress to include that the recipes generates scalar values for all VF lanes.

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
```diff
@@ -490,6 +490,14 @@ bool vputils::isUniformAcrossVFsAndUFs(const VPValue *V) {
       });
 }
 
+bool vputils::doesGeneratePerAllLanes(const VPRecipeBase *R) {
+  if (auto *RepR = dyn_cast<VPReplicateRecipe>(R))
+    return RepR->doesGeneratePerAllLanes();
+  if (auto *VPI = dyn_cast<VPInstruction>(R))
+    return VPI->doesGeneratePerAllLanes();
+  return isa<VPScalarIVStepsRecipe>(R);
```
fhahn: I think whether `VPScalarIVStepsRecipe` generates all lanes or only the first depends on if only the first lane is used, so better to check here? Not sure if this can be tested.

llvm/lib/Transforms/Vectorize/VPlan.h
```diff
@@ -3453,8 +3449,13 @@ class LLVM_ABI_FOR_TEST VPReplicateRecipe : public VPRecipeWithIRFlags,
                                          bool IsSingleScalar, ElementCount VF,
                                          VPCostContext &Ctx);
 
+  /// Returns true if the recipe produces a single scalar value, with scalar
+  /// operands.
```
fhahn:
```suggestion
  /// Returns true if the recipe produces a single scalar value.
```

I think at least at earlier stages of VPlan, the operands may not be scalar (e.g. widen recipe)?

llvm/lib/Transforms/Vectorize/VPlanUtils.cpp
artagnon: There was actually a VPlan-printing test that changed, yay!

artagnon:
Gentle ping.

PR #199051, opened 2026-05-21
[mlir][tosa] Limit consecutive concat rewrite to MAX_TENSOR_LIST_SIZE

lhutton1:
Previously folding could produce an operation that would later be considered invalid in validation due to the number of operands it has. This change adds a check to prevent rewriting consecutive concat operations if the resulting operation has more than MAX_TENSOR_LIST_SIZE operands, based on the selected target environment level. If no level is specified, folding will proceed as before.

In addition, this change rewrites the concat folder as a canonicalization pattern, since it is not a fold of constant operands. The change also consolidates testing in
canonicalize.mlir.


lhutton1:
cc @sahas3 @Tai78641 @udaya-ranga

Tai78641:
LGTM

mlir/lib/Dialect/Tosa/IR/TosaCanonicalizations.cpp
```diff
@@ -367,9 +368,61 @@ struct ConcatOptimization : public OpRewritePattern<tosa::ConcatOp> {
...
 
+struct ConsecutiveConcatOptimization : public OpRewritePattern<tosa::ConcatOp> {
+  using OpRewritePattern<tosa::ConcatOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(tosa::ConcatOp op,
+                                PatternRewriter &rewriter) const override {
+    // Rewrite consecutive concats on the same axis into a single op.
+    // Keep track of the operands so we are able to construct a new concat
+    // later. Conservatively assume that we double the number of operands when
+    // folding
```
sahas3:
```suggestion
    // canonicalizing
```

sahas3 (approved):
LGTM

PR #199052, opened 2026-05-21
[AMDGPU] Fix ShiftAmt32Imm to use unsigned comparison

wooseoklee:
ShiftAmt32Imm used a signed 'Imm < 32' predicate, which incorrectly matched negative immediates such as -1. The scalar fshr fast path:
    
      def : GCNPat<(UniformTernaryFrag<fshr> i32:$src0, i32:$src1,
                                             (i32 ShiftAmt32Imm:$src2)),
        (i32 (EXTRACT_SUBREG (S_LSHR_B64 ..., $src2), sub0))>;
    
When fshl(scalar, X, Z) is lowered via expandFunnelShift for any constant Z in [0, 31], the generic code converts it to fshr(..., ~Z) or fshr(..., -Z), producing a negative shift amount. Because all such values satisfy Imm < 32 in a signed comparison, ShiftAmt32Imm matched and the pattern passed the negative immediate directly to S_LSHR_B64 without the S_AND_B32 masking. S_LSHR_B64 then shifted by the wrong amount, producing an incorrect result.
    
Fix by changing the predicate to an unsigned comparison so that only values in [0, 31] match, and negative values fall through to the general pattern which correctly applies S_AND_B32 before S_LSHR_B64.
    
Fixes miscompilations of scalar fshl with any constant shift amount at -O0 on AMDGPU targets, covering both fshl(scalar, const, Z) and fshl(scalar, scalar, Z) for all Z in [0, 31].


llvm/test/CodeGen/AMDGPU/fshl-scalar-shift-zero.ll
```diff
@@ -0,0 +1,89 @@
...
+;
+; Correct behaviour:
+;   fshl(X, Y, 0) == X
+;   fshl(X, Y, 1) == (X << 1) | (Y >> 31)
+
+declare i32 @llvm.amdgcn.readfirstlane(i32)
+declare i32 @llvm.fshl.i32(i32, i32, i32)
+
+; fshl(scalar, const, 0) must return scalar unchanged.
+define amdgpu_kernel void @fshl_scalar_const_shift0(ptr addrspace(1) %out, i32 %x) {
```
shiltian: no need to use `amdgpu_kernel` here

llvm/test/CodeGen/AMDGPU/fshl-scalar-shift-zero.ll
```diff
@@ -0,0 +1,89 @@
...
+;   S_AND_B32 masking and caused S_LSHR_B64 to shift by 63 instead of 0.
+;
+; - fshl(scalar, const, 1): same path computes -Z = -1 as the inverted shift;
+;   same buggy immediate pattern matched and skipped masking.
+;
+; Correct behaviour:
+;   fshl(X, Y, 0) == X
+;   fshl(X, Y, 1) == (X << 1) | (Y >> 31)
+
+declare i32 @llvm.amdgcn.readfirstlane(i32)
```
shiltian: nit: no need for declaration

llvm/test/CodeGen/AMDGPU/fshl-scalar-shift-zero.ll
```diff
@@ -0,0 +1,89 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
+; RUN: llc -mtriple=amdgcn-amd-amdhsa -mcpu=gfx950 -O0 < %s | FileCheck %s
```
shiltian, line 2: Is it possible to get a reproducer without the use of `-O0`?

llvm/test/CodeGen/AMDGPU/fshl-scalar-shift-zero.ll
```diff
@@ -0,0 +1,89 @@
...
+
+; Regression tests for scalar fshl/fshr with constant shift amounts that
+; caused miscompilation via the scalar S_LSHR_B64 fast path:
+;
+; - fshl(scalar, const, 0): the FSHL->FSHR conversion computed ~0 = -1 as the
```
LU-JOHN, line 7: Would this bug also be hit for fshl(scalar, const, [2-31])?
Would this bug also be hit for fshl(scalar, scalar, [0-31])?

llvm/test/CodeGen/AMDGPU/fshl-scalar-shift-zero.ll
wooseoklee: Unfortunately, the two bug reports from the fuzzer say they are only reproducible with -O0, so I think we should use -O0 here. I had to create another test file because of that. 

m015-scalar-fshl-zero `-O0` lowers scalar `fshl.i32(x, y, 0)` through a 64-bit shift-by-`-1` sequence that returns zero.
m016-scalar-fshl-one -O0` lowers scalar `fshl.i32(x, y, 1)` through a 64-bit shift-by-`-1` sequence that returns only bit 31.

llvm/test/CodeGen/AMDGPU/fshl-scalar-shift-zero.ll
wooseoklee: Thanks. Updated.

llvm/test/CodeGen/AMDGPU/fshl-scalar-shift-zero.ll
wooseoklee: Updated.

llvm/test/CodeGen/AMDGPU/fshl-scalar-shift-zero.ll
wooseoklee: Thanks. Yes. It covers the two cases, and they are added.

LU-JOHN:
> When fshl(scalar, const, 0) or fshl(scalar, const, 1) is lowered 

Update this comment so that it is clear that this bug does not require 0 or 1.

LU-JOHN (approved):
LGTM.

wooseoklee:
> > When fshl(scalar, const, 0) or fshl(scalar, const, 1) is lowered
> 
> Update this comment so that it is clear that this bug does not require 0 or 1.

Updated.

llvm/lib/Target/AMDGPU/SIInstrInfo.td
```diff
@@ -988,7 +988,7 @@ def NegSubInlineIntConst16 : ImmLeaf<i16, [{
 }], NegateImm>;
 
 def ShiftAmt32Imm : ImmLeaf <i32, [{
-  return Imm < 32;
+  return (uint32_t)Imm < 32;
```
arsenm:
```suggestion
  return static_cast<uint32_t>(Imm) < 32;
```

wooseoklee: Thanks. Fixed it.

llvm/test/CodeGen/AMDGPU/fshl-scalar-shift-zero.ll
wooseoklee: At -O2, the issue never reaches the backend at all — InstCombine/DAGCombine folds fshl(X, Y, 0) to just X (the identity for shift-by-zero) before any lowering happens. The entire expandFunnelShift → fshr(..., ~0=-1) chain is never triggered, so ShiftAmt32Imm is never consulted.

With -O0, expandFunnelShift() produces following: fshl(X, Y, 0) -> fshr( srl(X, 1),  fshr(X, Y, 1),  ~0 )

Then, before the patch, the ShiftMat32Imm is

def ShiftAmt32Imm : ImmLeaf <i32, [{
  return Imm < 32;
}]>;

causing this pattern matches because -1 (~0 : 0xffffffff) < 32 is true.

def : GCNPat<(UniformTernaryFrag<fshr> i32:$src0, i32:$src1, (i32 ShiftAmt32Imm:$src2)),
  (i32 (EXTRACT_SUBREG (S_LSHR_B64 (REG_SEQUENCE SReg_64, $src1, sub0, $src0, sub1), $src2), sub0))
>;

leading to LSHR applies the hardware mask of 6bits (-1 & 63 = 63), shifting by 63 and zeroing result.

After the patch, the ShfitMat32Imm

def ShiftAmt32Imm : ImmLeaf <i32, [{
  return static_cast<uint32_t>(Imm) < 32;
}]>;

causing the prior pattern fails to match because 0xffffffff < 32 is false. Instead, this pattern matches,

def : GCNPat<(UniformTernaryFrag<fshr> i32:$src0, i32:$src1, i32:$src2),
  (i32 (EXTRACT_SUBREG (S_LSHR_B64 (REG_SEQUENCE SReg_64, $src1, sub0, $src0, sub1), (S_AND_B32 $src2, (i32 31))), sub0))
>;

This filter ensures correct value: S_AND_B32 $src2, (i32 31)

  -1  =  0xFFFFFFFF  =  1111 1111 1111 1111 1111 1111 1111 1111
  31  =  0x0000001F  =  0000 0000 0000 0000 0000 0000 0001 1111
AND   =  0x0000001F  =  0000 0000 0000 0000 0000 0000 0001 1111  =  31

S_LSHR_B64 [X>>1 : fshr(X,Y,1)], 31
= bits [62:31] of the 64-bit value
= { (X>>1)[30:0], fshr(X,Y,1)[31] }
= { X[31:1],      X[0]            }
= X

PR #199055, opened 2026-05-21
[libc] Implement BSD extensions in <err.h>

kaladron:
Implemented the BSD-extension functions defined in <err.h>:

- err, errx, verr, verrx
- warn, warnx, vwarn, vwarnx

Added an internal reporting helper in src/__support/Err. The err family functions call internal::exit directly to satisfy [[noreturn]] requirements. Enabled these entrypoints for aarch64, riscv, and x86_64.


libc/test/src/err/err_test.cpp
```diff
@@ -0,0 +1,89 @@
+//===----------------------------------------------------------------------===//
```
michaelrj-google, line 1: ideally each function would have its own test file. That ensures that functions are still tested even if some of them are disabled on a target.

libc/src/err/report.cpp
```diff
@@ -0,0 +1,76 @@
...
+namespace {
+
+void write_to_stderr(cpp::string_view msg) {
+  LIBC_NAMESPACE::write_to_stderr(msg);
+}
+
+} // namespace
+
+void report(bool show_err, const char *fmt, internal::ArgList &args) {
+  int saved_errno = libc_errno;
```
michaelrj-google: saving errno isn't necessary, internal functions aren't allowed to write to errno.

libc/src/err/report.cpp
```diff
@@ -0,0 +1,76 @@
...
+namespace LIBC_NAMESPACE_DECL {
+namespace err_reporting {
+
+namespace {
+
+void write_to_stderr(cpp::string_view msg) {
+  LIBC_NAMESPACE::write_to_stderr(msg);
+}
+
+} // namespace
```
michaelrj-google: why is this here? Below you call `LIBC_NAMESPACE::write_to_stderr` directly. That explicit namespacing shouldn't be necessary either though, since this is all in the namespace anyways.

libc/src/err/report.cpp
```diff
@@ -0,0 +1,76 @@
...
+    char buffer[1024];
+    printf_core::FlushingBuffer wb(
+        buffer, sizeof(buffer),
+        [](cpp::string_view str, [[maybe_unused]] void *raw_stream) -> int {
+          LIBC_NAMESPACE::write_to_stderr(str);
+          return static_cast<int>(str.size());
+        },
+        nullptr);
+    printf_core::Writer writer(wb);
+    printf_core::printf_main(&writer, fmt, args);
```
michaelrj-google: `printf_main` returns the number of characters written on success and a negative error value on failure. You should check for errors here.

michaelrj-google (commented):
it looks like you're missing the `errc`/`warnc` families of functions (https://man.openbsd.org/err.3). Was that intentional?

kaladron:
> it looks like you're missing the `errc`/`warnc` families of functions (https://man.openbsd.org/err.3). Was that intentional?

Yes.  I'm only implementing the ones that glibc implemented here (we have code that uses these, but otherwise I don't want to take on all the other interfaces unless we actively decide to)

libc/test/src/err/err_test.cpp
michaelrj-google: tests should not be in `LIBC_NAMESPACE` so we are explicitly calling the LLVM-libc functions.

michaelrj-google (commented):
Mostly looks good

michaelrj-google (approved):

PR #199056, opened 2026-05-21
[lldb] Preserve FIFO order for equal time MainLoop callbacks

charles-zablit:
`MainLoopBase::AddCallback` / `AddPendingCallback` enqueue into a `std::priority_queue` keyed only on `TimePoint`. `AddPendingCallback` uses a default-constructed `TimePoint`, so all pending callbacks share the same key and the heap returns them in arbitrary order.

This patch fixes this by adding an int counter to sort entries with equal time points.

This fixes `commands/expression/dont_allow_jit/TestAllowJIT.py` when running with `LLDB_USE_LLDB_SERVER=1`.

rdar://177625246

### DIFF
--- a/lldb/include/lldb/Host/MainLoopBase.h
+++ b/lldb/include/lldb/Host/MainLoopBase.h
@@ -111,7 +111,7 @@ protected:
         : time_point(std::move(tp)), callback(std::move(cb)), sequence(seq) {}
 
     bool operator<(const CallbackEntry &other) const {
-      return std::tie(time_point, sequence) >
+      return std::tie(time_point, sequence) > // > for Min-priority queue
              std::tie(other.time_point, other.sequence);
     }
 
--- a/lldb/include/lldb/Host/MainLoopBase.h
+++ b/lldb/include/lldb/Host/MainLoopBase.h
@@ -14,9 +14,12 @@
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/Support/ErrorHandling.h"
 #include <chrono>
+#include <cstdint>
 #include <functional>
 #include <mutex>
 #include <queue>
+#include <tuple>
+#include <vector>
 
 namespace lldb_private {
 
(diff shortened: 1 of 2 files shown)


lldb/include/lldb/Host/MainLoopBase.h
```diff
@@ -99,11 +102,27 @@ class MainLoopBase {
...
+    TimePoint time_point;
+    Callback callback;
+
+    CallbackEntry(TimePoint tp, Callback cb, uint64_t seq)
+        : time_point(std::move(tp)), callback(std::move(cb)), sequence(seq) {}
+
+    bool operator<(const CallbackEntry &other) const {
+      if (time_point != other.time_point)
+        return time_point > other.time_point;
+      return sequence > other.sequence;
```
adrian-prantl: I think this is equivalent to the more idiomatic:
```
return std::tie(time_point, sequence) > std::tie(other.time_point, other.sequence);
```

adrian-prantl (commented):
Seems reasonable, one comment inside.

lldb/include/lldb/Host/MainLoopBase.h
charles-zablit: `first non-equal result is returned`. It's equivalent, switched to it. Thanks!

lldb/include/lldb/Host/MainLoopBase.h
```diff
@@ -99,11 +102,26 @@ class MainLoopBase {
...
+
+  struct CallbackEntry {
+    TimePoint time_point;
+    Callback callback;
+
+    CallbackEntry(TimePoint tp, Callback cb, uint64_t seq)
+        : time_point(std::move(tp)), callback(std::move(cb)), sequence(seq) {}
+
+    bool operator<(const CallbackEntry &other) const {
+      return std::tie(time_point, sequence) >
```
adrian-prantl:
```suggestion
      return std::tie(time_point, sequence) <
```
?

charles-zablit: We want a min-priority queue here. I added a comment to clarify.

lldb/include/lldb/Host/MainLoopBase.h
```diff
@@ -99,11 +102,26 @@ class MainLoopBase {
...
-  bool m_terminate_request : 1;
+
+  struct CallbackEntry {
+    TimePoint time_point;
+    Callback callback;
+
+    CallbackEntry(TimePoint tp, Callback cb, uint64_t seq)
+        : time_point(std::move(tp)), callback(std::move(cb)), sequence(seq) {}
+
+    bool operator<(const CallbackEntry &other) const {
```
adrian-prantl, line 114: ### 
```suggestion
    /// Sort using the `>`(!) operator to create a min-priority queue.
    bool operator<(const CallbackEntry &other) const {
```

adrian-prantl (commented):
LGTM with a suggestion for a comment.

adrian-prantl (approved):

PR #196900, opened 2026-05-11
[VPlan] Fold reverse(reverse(x)) -> x in simplifyRecipe()

madhur13490:
Eliminate redundant reverse(reverse(x)) recipes during VPlan simplification.

Issue #172961, opened 2025-12-19
Remove vp.reverse


Mel-Chen:
Currently, vp.reverse is only used for reverse accesses. Taking a reverse load as an example:
```
vp.reverse(
  vp.load(vec_end_ptr(addr, evl), true /* mask */, evl),
  true /* mask */,
  evl
)
```

Its semantics are equivalent to:
```
vector.reverse(
  vector.splice(
    poison,
    vp.load(vec_end_ptr(addr, evl), true /* mask */, evl),
    evl
  )
)
```

With this approach, we no longer need the vp.reverse intrinsic. Moreover, after the reverse operation is simplified away, optimizeMaskToEVL can still transform it into an reverse load/store with EVL tail folding.

lukel97:
Just making a note here, there's multiple ways to represent a vp.reverse with vector.splice.left/right + vector.reverse.

After trying out a few, I think `vector.splice.right(vector.reverse(V), poison, EVL)` is probably the best "canonical" choice. 

If you use `vector.splice.left(vector.reverse(V), poison, VF - EVL)` you have to calculate VF - EVL in LLVM IR.  If you use `vector.reverse(vector.splice.left(poison, V, VF))`, then you end up with extra vsetvli toggles on RISC-V:

```asm
	csrr	a2, vlenb
	vsetvli	a3, zero, e32, m1, ta, ma
	vid.v	v12
	srli	a3, a2, 1
	srli	a2, a2, 2
	sub	a3, a3, a1
	addi	a2, a2, -1
	vsetvli	a4, zero, e32, m2, ta, ma
	vslideup.vx	v10, v8, a3
	vsetvli	a3, zero, e32, m1, ta, ma
	vrsub.vx	v12, v12, a2
	vrgather.vv	v9, v10, v12
	vrgather.vv	v8, v11, v12
	vsetvli	zero, a1, e32, m2, ta, ma
	vse32.v	v8, (a0)
	ret
```

With vector.splice.right after the reverse we can reduce the VL and avoid the toggle

```asm
	csrr	a2, vlenb
	vsetvli	a3, zero, e32, m1, ta, ma
	vid.v	v10
	srli	a3, a2, 2
	srli	a2, a2, 1
	addi	a3, a3, -1
	vrsub.vx	v12, v10, a3
	vrgather.vv	v11, v8, v12
	vrgather.vv	v10, v9, v12
	sub	a2, a2, a1
[...]

PR #199057, opened 2026-05-21
[VPlan] Simplify reverse(reverse(x)) -> x

lukel97:
This is a version of #196900 that performs the simplification as a separate transform.

We need to add an additional `vp.splice.right(vp.splice.left(poison, x, evl), poison, evl) -> x` simplification to avoid left over splices whenever reverses are removed in an EVL tail folded loop.

Co-authored-by: Madhur Amilkanthwar <madhura@nvidia.com>


madhur13490:
I think I am missing something here. Does this mean I should abandon my patch? I don't see the following code in this patch.

```
  // reverse(reverse(x)) -> x. The two reverses cancel out; leave the inner
  // recipe for removeDeadRecipes to clean up if it becomes unused.
  if (match(Def, m_Reverse(m_Reverse(m_VPValue(X))))) 
    return Def->replaceAllUsesWith(X);
```

lukel97:
> I think I am missing something here. Does this mean I should abandon my patch? I don't see the following code in this patch.
> 
> ```
>   // reverse(reverse(x)) -> x. The two reverses cancel out; leave the inner
>   // recipe for removeDeadRecipes to clean up if it becomes unused.
>   if (match(Def, m_Reverse(m_Reverse(m_VPValue(X))))) 
>     return Def->replaceAllUsesWith(X);
> ```

That combine is included in this patch, see this line here: https://github.com/llvm/llvm-project/pull/199057/changes#diff-53267225b83e943ceae51c326c9941e323fd7aaf74a08b5e6998d6456f88d1ddR1812

I've included you as a co-author for this patch, but happy to take these changes into the other patch if you'd prefer. The main thing is that I don't think we can do the combine in `simplifyRecipes` without regressing RISC-V, so I've gone ahead and implemented what I had in mind in the review comment here: https://github.com/llvm/llvm-project/pull/196900#pullrequestreview-4280416928

lukel97:
@Mel-Chen in future I think we can probably extend such a pass like this to also handle more involved patterns like `reverse(binop(reverse(x))) -> binop(x)`

madhur13490:
> I think I am missing something here. Does this mean I should abandon my patch? I don't see the following code in this patch.
> 
> ```
>   // reverse(reverse(x)) -> x. The two reverses cancel out; leave the inner
>   // recipe for removeDeadRecipes to clean up if it becomes unused.
>   if (match(Def, m_Reverse(m_Reverse(m_VPValue(X))))) 
>     return Def->replaceAllUsesWith(X);
> ```

Oops. I spoke to soon. I see that in the big `m_CombineOr`

madhur13490:
> > I think I am missing something here. Does this mean I should abandon my patch? I don't see the following code in this patch.
> > ```
> >   // reverse(reverse(x)) -> x. The two reverses cancel out; leave the inner
> >   // recipe for removeDeadRecipes to clean up if it becomes unused.
> >   if (match(Def, m_Reverse(m_Reverse(m_VPValue(X))))) 
> >     return Def->replaceAllUsesWith(X);
> > ```
> 
> That combine is included in this patch, see this line here: https://github.com/llvm/llvm-project/pull/199057/changes#diff-53267225b83e943ceae51c326c9941e323fd7aaf74a08b5e6998d6456f88d1ddR1812
> 
> I've included you as a co-author for this patch, but happy to take these changes into the other patch if you'd prefer. The main thing is that I don't think we can do the combine in `simplifyRecipes` without regressing RISC-V, so I've gone ahead and implemented what I had in mind in the review comment here: [#196900 (review)](https://github.com/llvm/llvm-project/pull/196900#pullrequestreview-4280416928)

Thanks. No problem. I will take my patch back.

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -1802,6 +1802,25 @@ void VPlanTransforms::simplifyRecipes(VPlan &Plan) {
...
+void VPlanTransforms::simplifyReverses(VPlan &Plan) {
+  VPValue *X, *EVL;
+  PostOrderTraversal<VPBlockDeepTraversalWrapper<VPBlockBase *>> POT(
+      Plan.getEntry());
+  for (VPBasicBlock *VPBB : VPBlockUtils::blocksOnly<VPBasicBlock>(POT)) {
+    for (VPRecipeBase &R : make_early_inc_range(reverse(*VPBB))) {
+      if (match(&R,
+                m_CombineOr(m_Reverse(m_Reverse(m_VPValue(X))),
+                            m_Intrinsic<Intrinsic::experimental_vp_reverse>(
+                                m_Intrinsic<Intrinsic::experimental_vp_reverse>(
```
fhahn: Maybe it is time to add a `m_vp_Reverse` matcher?

llvm/test/Transforms/LoopVectorize/VPlan/simplify-reverse-reverse.ll
```diff
@@ -0,0 +1,78 @@
+; NOTE: Assertions have been autogenerated by utils/update_analyze_test_checks.py UTC_ARGS: --version 6
+; RUN: opt -p loop-vectorize -force-vector-width=4 -disable-output -vplan-print-after=printOptimizedVPlan %s 2>&1 | FileCheck %s
```
fhahn: can we check after `simplifyReverses`?

llvm/test/Transforms/LoopVectorize/VPlan/simplify-reverse-reverse.ll
```diff
@@ -0,0 +1,78 @@
...
+; CHECK-NEXT:      CLONE ir<%gep.dst> = getelementptr ir<%dst>, ir<%iv.next>
+; CHECK-NEXT:      vp<[[VP10:%[0-9]+]]> = vector-end-pointer ir<%gep.dst>, vp<[[VP0]]>
+; CHECK-NEXT:      WIDEN store vp<[[VP10]]>, ir<%x>
+; CHECK-NEXT:      EMIT vp<%index.next> = add nuw vp<[[VP5]]>, vp<[[VP1]]>
+; CHECK-NEXT:      EMIT branch-on-count vp<%index.next>, vp<[[VP2]]>
+; CHECK-NEXT:    No successors
+; CHECK-NEXT:  }
+; CHECK-NEXT:  Successor(s): middle.block
+; CHECK-EMPTY:
+; CHECK-NEXT:  middle.block:
```
fhahn, line 42: can filter-out-after middle block I think

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -1802,6 +1802,25 @@ void VPlanTransforms::simplifyRecipes(VPlan &Plan) {
   }
 }
 
+void VPlanTransforms::simplifyReverses(VPlan &Plan) {
+  VPValue *X, *EVL;
+  PostOrderTraversal<VPBlockDeepTraversalWrapper<VPBlockBase *>> POT(
```
fhahn: Do we need to traverse in post order?

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -1802,6 +1802,25 @@ void VPlanTransforms::simplifyRecipes(VPlan &Plan) {
   }
 }
 
+void VPlanTransforms::simplifyReverses(VPlan &Plan) {
```
fhahn, line 1856: may be worth to add a brief comment somewhere explaining why this his done separately instead of in `simplifyRecipes`

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -1802,6 +1802,25 @@ void VPlanTransforms::simplifyRecipes(VPlan &Plan) {
...
+                            m_Intrinsic<Intrinsic::experimental_vp_reverse>(
+                                m_Intrinsic<Intrinsic::experimental_vp_reverse>(
+                                    m_VPValue(X), m_VPValue(), m_VPValue(EVL)),
+                                m_VPValue(), m_Deferred(EVL))))) {
+        R.getVPSingleValue()->replaceAllUsesWith(X);
+        R.eraseFromParent();
+      }
+    }
+  }
+}
```
madhur13490, line 1863: The loop is not readable. Please outline the match part and use early-return pattern something like below:

```
VPValue *matchRevOfRev(VPRecipeBase ...) {
  if (match(&R, m_Reverse(m_Reverse(m_VPValue(X)))))
    return X;
  if (match(&R, m_Intrinsic<Intrinsic::experimental_vp_reverse>(
                    m_Intrinsic<Intrinsic::experimental_vp_reverse>(
                        m_VPValue(X), m_True(), m_VPValue(EVL)),
                    m_True(), m_Deferred(EVL))))
    return X;
  return nullptr;
}

void simplifyReverses(VPlan &Plan) {
  for (VPBasicBlock *VPBB : VPBlockUtils::blocksOnly<VPBasicBlock>(
           vp_depth_first_deep(Plan.getEntry()))) {
    for (VPRecipeBase &R : make_early_inc_range(*VPBB)) {
      VPValue *X = matchReverseOfReverse(R);
      if (!X)
        continue;
      cast<VPSingleDefRecipe>(&R)->replaceAllUsesWith(X);
      R.eraseFromParent();
    }
  }
}
```

It is better and manageable.

madhur13490 (changes requested):

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
lukel97: Done in ea078f02179b82cfd4ab5979842ee12231aaaf28

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
lukel97: I figured it's probably better to match from the bottom up in case there were multiple nested reverses. But looks like it doesn't make a difference in practice, removed in 04ad90cfd26d1b4d64e732a53d32a7a16f64f610

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
lukel97: Done in cecaa912354b07e60a8a81f16b88dfa4a53eb588

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
lukel97: Good idea, added a matcher lambda in cecaa912354b07e60a8a81f16b88dfa4a53eb588

llvm/test/Transforms/LoopVectorize/VPlan/simplify-reverse-reverse.ll
lukel97: Done in 9873d798f2f76316ceee4d9cc8896c03aed73f59

llvm/test/Transforms/LoopVectorize/VPlan/simplify-reverse-reverse.ll
lukel97: Done in de437f2af1825968f10421435e06e6ad3ac74b58

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -1802,6 +1802,27 @@ void VPlanTransforms::simplifyRecipes(VPlan &Plan) {
   }
 }
 
+void VPlanTransforms::simplifyReverses(VPlan &Plan) {
+  auto m_ReverseReverse = [](auto X) {
+    VPValue *EVL;
+    return m_CombineOr(
+        m_Reverse(m_Reverse(X)),
+        m_vp_Reverse(m_vp_Reverse(X, m_VPValue(), m_VPValue(EVL)), m_VPValue(),
```
fhahn: Just double checking, we can ignore the mask, as the extra masked off elements would be poison, and we may replace them with a non-poison value, right?

lukel97: Yup. And in the future we should be able to replace vp_reverse with regular reverse + splice https://github.com/llvm/llvm-project/issues/172961

Mel-Chen:
https://github.com/llvm/llvm-project/pull/199510
This POC is the approach I mentioned in https://github.com/llvm/llvm-project/pull/196900#issuecomment-4439787534. I hope we can complete the simplification of reverse before optimizeEVLMasks. This will make it much easier for us in the future when transforming reverse accesses into strided accesses with a stride of -1.

lukel97:
> #199510 This POC is the approach I mentioned in [#196900 (comment)](https://github.com/llvm/llvm-project/pull/196900#issuecomment-4439787534). I hope we can complete the simplification of reverse before optimizeEVLMasks. This will make it much easier for us in the future when transforming reverse accesses into strided accesses with a stride of -1.

From https://github.com/llvm/llvm-project/pull/200794#issuecomment-4611365817, it sounds like we now want to run permutation simplification after optimizeEVLMasks? 

Although I'm not sure why transforming reverse accesses into strided accesses with a stride of -1 would be more difficult after `simplifyReverses`. Wouldn't we just be matching `(vp.reverse (vp.load end-ptr, evl), evl)`?

Mel-Chen:
> > #199510 This POC is the approach I mentioned in [#196900 (comment)](https://github.com/llvm/llvm-project/pull/196900#issuecomment-4439787534). I hope we can complete the simplification of reverse before optimizeEVLMasks. This will make it much easier for us in the future when transforming reverse accesses into strided accesses with a stride of -1.
> 
> From [#200794 (comment)](https://github.com/llvm/llvm-project/pull/200794#issuecomment-4611365817), it sounds like we now want to run permutation simplification after optimizeEVLMasks?
> 
> Although I'm not sure why transforming reverse accesses into strided accesses with a stride of -1 would be more difficult after `simplifyReverses`. Wouldn't we just be matching `(vp.reverse (vp.load end-ptr, evl), evl)`?

I think there might be a misunderstanding here. Let me break down the workflow as I envision it:
1. A transformation to eliminate redundant `vector.reverse` operations (e.g., reverse(reverse(x)) -> x).
2. `VPlanTransforms::convertToStridedAccesses`: Based on the cost, it decides whether to convert reverse access patterns into strided accesses with a stride of -1, and clamps the VF of plan.
3. (For EVL tail folding) As proposed in #199510, `optimizeMaskToEVL` will no longer depend on the reverse operation of the loaded result or stored value. Therefore, even if reverse is eliminated in step 1, it won't affect EVL lowering. At this stage, splice operations will be generated, some of which might be redundant.
4. (For EVL tail folding) A transformation to eliminate redundant splice operations, which cleans up the extra splices introduced in step 3. (e.g. splice.right(splice.left(poison, val, evl), poison, evl) -> val)

The misunderstanding likely stems from the fact that I grouped both transformations (1 and 4) under `permutation elimination` and they can be implemented in a single transformation, and just need to call it two times.

Hopefully, this clarifies my perspective. :)

fhahn (approved):
> > > #199510 This POC is the approach I mentioned in [#196900 (comment)](https://github.com/llvm/llvm-project/pull/196900#issuecomment-4439787534). I hope we can complete the simplification of reverse before optimizeEVLMasks. This will make it much easier for us in the future when transforming reverse accesses into strided accesses with a stride of -1.
> > 
> > 
> > From [#200794 (comment)](https://github.com/llvm/llvm-project/pull/200794#issuecomment-4611365817), it sounds like we now want to run permutation simplification after optimizeEVLMasks?
> > Although I'm not sure why transforming reverse accesses into strided accesses with a stride of -1 would be more difficult after `simplifyReverses`. Wouldn't we just be matching `(vp.reverse (vp.load end-ptr, evl), evl)`?
> 
> I think there might be a misunderstanding here. Let me break down the workflow as I envision it:
> 
>     1. A transformation to eliminate redundant `vector.reverse` operations (e.g., reverse(reverse(x)) -> x).
> 
>  

This patch could serve as that starting point, yes? 

The logic here LGTM.

lukel97:
> 2\. `VPlanTransforms::convertToStridedAccesses`: Based on the cost, it decides whether to convert reverse access patterns into strided accesses with a stride of -1, and clamps the VF of plan.

@Mel-Chen  Ah my understanding was that converting reverse accesses to strides of -1 would be a separate transform from `VPlanTransforms::convertToStridedAccesses` that would run after optimizeMasksToEVL.

But yes I think we can also teach the convertToStridedAccesses to handle reverses too. So your proposed ordering sounds good to me.

So to check, are you happy with the following order of PRs?

1) Add simplifyReverses, after optimizeMasksToEVL #199057
2) Handle EVL reverses as splices in #199510
3) Move simplifyReverses up before convertToStridedAccesses, some future PR

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -1835,6 +1835,27 @@ void VPlanTransforms::simplifyRecipes(VPlan &Plan) {
...
+                     m_Deferred(EVL)));
+  };
+
+  for (VPBasicBlock *VPBB : VPBlockUtils::blocksOnly<VPBasicBlock>(
+           vp_depth_first_deep(Plan.getEntry()))) {
+    for (VPRecipeBase &R : make_early_inc_range(*VPBB)) {
+      VPValue *X;
+      if (match(&R, m_ReverseReverse(m_VPValue(X)))) {
```
david-arm: In general I have also seen IR generated the vectoriser like this:

%rev1 = reverse %a
%fadd = fadd fast %rev1, splat (float 1.0e+00)
%rev2 = reverse %fadd

where there is only one use of %rev1 and one use of %fadd. I think these pair-reverse simplifications could be used in more places to reduce the cost, since instcombine would fold away these reverses.

lukel97: Yeah the plan is to also simplify those away in VPlan after this PR by "pulling" reverses out of any elementwise operation, see #199234. So the `reverse(fadd(reverse x), 1.0)` becomes `reverse(reverse(fadd x, 1.0))`, and the reverses then cancel out. This is how InstCombine does the same simplification IIUC

Mel-Chen:
> > I think there might be a misunderstanding here. Let me break down the workflow as I envision it:
> > ```
> > 1. A transformation to eliminate redundant `vector.reverse` operations (e.g., reverse(reverse(x)) -> x).
> > ```
> 
> This patch could serve as that starting point, yes?
> 
> The logic here LGTM.

@fhahn There seems to be a misunderstanding here. To clarify, this is the execution order of the transformations, not the landing order of the patches.

Mel-Chen:
> > 2. `VPlanTransforms::convertToStridedAccesses`: Based on the cost, it decides whether to convert reverse access patterns into strided accesses with a stride of -1, and clamps the VF of plan.
> 
> @Mel-Chen Ah my understanding was that converting reverse accesses to strides of -1 would be a separate transform from `VPlanTransforms::convertToStridedAccesses` that would run after optimizeMasksToEVL.

This reverse-to-strided-access transformation is also valid even without EVL tail folding. Furthermore, if we perform this conversion before EVL lowering, we only need to match the standard reverse(load(...)) pattern and don't have to account for vp.reverse at all.

> 
> But yes I think we can also teach the convertToStridedAccesses to handle reverses too. So your proposed ordering sounds good to me.
> 
> So to check, are you happy with the following order of PRs?
> 
> 1. Add simplifyReverses, after optimizeMasksToEVL [[VPlan] Simplify reverse(reverse(x)) -> x #199057](https://github.com/llvm/llvm-project/pull/199057)
> 2. Handle EVL reverses as splices in [[VPlan][NFC] Simplify reverse access pattern detection in EVL vectorization #199510](https://github.com/llvm/llvm-project/pull/199510)
> 3. Move simplifyReverses up before convertToStridedAccesses, some future PR

While I feel we can always adjust the code to our plan regardless of which patch lands first, landing #199510 first as an NFC would allow us to adopt the `reverse(reverse(x)) -> x` rule sooner. We could even temporarily place it directly in simplifyRecipes as a simplification rule.
Additionally, this patch wouldn't have to deal with the vp.reverse case anymore. In other words, with #199510 in place, this patch should be able to do significantly less work.

lukel97:
@Mel-Chen glad to see https://github.com/llvm/llvm-project/pull/199510 got landed, rebased this PR now on top and moved the transform to VPlanTransforms::optimize. This also now requires a `splice.right(splice.left(x)) -> x` transform which I've added to optimizeMasksToEVL.

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -3138,11 +3148,22 @@ void VPlanTransforms::optimizeEVLMasks(VPlan &Plan) {
...
 
-  // Fold the following splice patterns into vp.reverse for reverse accesses:
+  // Fold the following splice patterns:
+  //   splice.right(splice.left(poison, x, evl), poison, evl) -> x
   //   vector.reverse(splice.left(poison, x, evl))  -> vp.reverse(x, true, evl)
   //   splice.right(vector.reverse(x), poison, evl) -> vp.reverse(x, true, evl)
   for (VPUser *U : collectUsersRecursively(EVL)) {
+    auto *Def = cast<VPRecipeBase>(U);
     VPValue *X;
+    if (match(U, m_Intrinsic<Intrinsic::vector_splice_right>(
```
david-arm, line 3177: Is this guaranteed to be safe? Is there something in the LangRef or intrinsic property of EVL that ensures EVL <= number of elements of x? If not, then you may end up permanently losing elements of x and so the transformation is unsafe.

Could this part of the PR be committed separately to the simplifyReverses bit?

lukel97: If EVL > the number of elements in x then per the langref the result is poison, so the transform should be safe: https://llvm.org/docs/LangRef.html#id981

We end up with stray vector.splices on RISC-V without this, and InstCombine doesn't fold the splices away. So unfortunately I think we will end up with regressions since the reverses that are currently emitted are folded away by InstCombine

david-arm: OK thanks for the explanation. In that case, is it worth separately also adding such a transform to InstCombine since the LangRef guarantees the safety of it? Or is the loop vectoriser the only possible source of such IR and therefore adds no value?

david-arm: I do agree from a cost model perspective it's much better to remove the splices here because otherwise they are too expensive

lukel97: I think the loop vectorizer is the only place within LLVM itself that emits splices, but perhaps there's downstream users too that could take advantage of an InstCombine transform. I don't think it's an unreasonable transform to have, I can create an issue for that

> I do agree from a cost model perspective it's much better to remove the splices here because otherwise they are too expensive

Agreed, and FWIW because reverses are so expensive I'm seeing a lot more loop vectorization across the board in combination with #199234. It removes a huge chunk of the cost in the VPlan when the loop just happens to be iterated backwards. I'll get some specific numbers on the increase in number of loops vectorized and include them in that PR.

Mel-Chen:
> @Mel-Chen in future I think we can probably extend such a pass like this to also handle more involved patterns like `reverse(binop(reverse(x))) -> binop(x)`

I’ve previously put together a list of planned simplification rules for reverse and wanted to share it https://github.com/llvm/llvm-project/issues/203206. Feel free to correct the issue if you spot any mistakes.

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
Mel-Chen: I think splice elimination is worth spinning off into a separate patch, provided there are no performance regressions.

Interestingly, I noticed that the cost of RISCV `vp.reverse` is actually higher than `vector.splice`. I'm not sure if there's an issue with the cost model here, but even if we only eliminate `vector.reverse` and leave `vector.splice`, the overall VPlan-based cost for EVL tail folding still decreases.

Could we check how bad the generated assembly looks with the leftover splice to help us decide whether to handle its elimination separately?

lukel97: The regression isn't with the cost model but with the codegen.  InstCombine currently cancels out the reverses when VPlan doesn't, but if we cancel out the reverses in VPlan but leave behind the splices, InstCombine won't remove them. 

E.g. this loop will regress:

```c
void f(int *x, int *y, int n) {
  for (int i = n - 1; i >= 0; i--)
    x[i] = y[i];
}
```

On ToT and with splice elimination:

```asm
.LBB0_5:                                # %vector.body
                                        # =>This Inner Loop Header: Depth=1
	vsetvli	a7, a5, e8, mf2, ta, ma
	not	a3, a4
	add	t1, a3, a2
	sh2add	t0, t1, a1
	sub	t2, a6, a7
	sh2add	a3, t2, t0
	vle32.v	v8, (a3)
	sh2add	a3, t1, a0
	sh2add	a3, t2, a3
	sub	a5, a5, a7
	vse32.v	v8, (a3)
	add	a4, a4, a7
	bnez	a5, .LBB0_5
```

Without

```asm
	vsetvli	t0, a4, e8, mf2, ta, ma
	not	a3, a5
	add	t2, a3, a2
	sh2add	t1, t2, a1
	sub	t3, a6, t0
	sh2add	a3, t3, t1
	vle32.v	v8, (a3)
	sub	t1, a7, t0
	vsetvli	a3, zero, e32, m2, ta, ma
	vslideup.vx	v10, v8, t1
	vsetvli	zero, a4, e32, m2, ta, ma
	vslidedown.vx	v8, v10, t1
	sh2add	a3, t2, a0
	sh2add	a3, t3, a3
	sub	a4, a4, t0
	vse32.v	v8, (a3)
	add	a5, a5, t0
	bnez	a4, .LBB0_5
```

Mel-Chen: Oh, I see! Since vp.reverse is folded by instcombine, not eliminating splice at the same time would cause a performance regression.

Mel-Chen (approved):
LGTM, thanks!

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
```diff
@@ -3170,12 +3191,11 @@ void VPlanTransforms::optimizeEVLMasks(VPlan &Plan) {
                        m_Reverse(m_VPValue(X)), m_Poison(), m_Specific(EVL)))))
       continue;
 
-    auto *Def = cast<VPSingleDefRecipe>(U);
```
fhahn, line 3173: Is there a reason we can no longer cast to VPSingleDefRecipe?

fhahn (approved):
LGTM, thanks

llvm/lib/Transforms/Vectorize/VPlanTransforms.cpp
lukel97: I moved Def further up before the other match so it could be shared, but this means we can only cast to VPRecipeBase. Since we'll iterate through VPWidenStoreEVL which isn't a single def. It should probably be renamed R for recipe actually.

PR #199058, opened 2026-05-21
[clang][CodeGen] Forward PseudoProbeForProfiling in IRInstr PGOOptions

fzakaria:
We build Clang with PGO at the moment but I want to transition us to CSSPGO.
In order to do so, I need add the pseudo probe instrumentation. Unfortunately, I can't seem to generate a profile and apply pseudo probes at the same time -- it looked to have been explicitly disabled.

The IRInstr branch in BackendUtil.cpp hardcodes PseudoProbeForProfiling to false, silently suppressing pseudo probe emission when -fprofile-generate and -fpseudo-probe-for-profiling are both passed. This change forwards the CodeGenOpts value instead so pseudo probes can coexist with instrumentation PGO.

I added it also to the `-fprofile-use` case as well, so that the code layout for PGO itself should match and there would be no profile statleness.

### DIFF
--- a/clang/lib/CodeGen/BackendUtil.cpp
+++ b/clang/lib/CodeGen/BackendUtil.cpp
@@ -837,17 +837,17 @@ void EmitAssemblyHelper::RunOptimizationPipeline(
                         CodeGenOpts.MemoryProfileUsePath, PGOOptions::IRInstr,
                         PGOOptions::NoCSAction, ClPGOColdFuncAttr,
                         CodeGenOpts.DebugInfoForProfiling,
-                        /*PseudoProbeForProfiling=*/false,
+                        CodeGenOpts.PseudoProbeForProfiling,
                         CodeGenOpts.AtomicProfileUpdate);
   else if (CodeGenOpts.hasProfileIRUse()) {
     // -fprofile-use.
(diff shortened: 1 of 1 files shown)


fzakaria:
CC @WenleiHe @apolloww

WenleiHe:
I can see that pseudo-probe for `-fprofile-use` can be valuable; but i don't see why one would want to generate probe for `-fprofile-generate`. CSSPGO more or less still depends on inlining of the profiling build, and `-fprofile-generate` is a special build so the CSSPGO profiling on that binary isn't that useful anyways.

fzakaria:
@WenleiHe we have this problem that we are generating PGO Clang now but I want to migrate to CSSPGO.
In order to migrate to CSSPGO, the binary must have the pseudo probe sections included.
These section additions however at the moment (prior to this PR) are mutually exclusive with the PGO profile-usage flag.

They don't really make sense together, except that we want to include the pseudo probes while still retaining the PGO profile performance in the meantime.

fzakaria:
I was led to believe the llvm.pseudoprobe intrinsics could subtly affect optimization decisions, causing the code layout to drift from what the profile expects. Is that incorrect?

fzakaria:
I will close this PR.
I got convinced it's not the right approach and I will try to migrate us to us to CSSPGO first or worst case an nternal patch.

PR #199064, opened 2026-05-21
[lldb] Consolidate ScriptedThreadPlan state into ScriptedMetadata

medismailben:
This is a follow-up to 1b4a578a9f77.

Replace ScriptedThreadPlan's separate `m_class_name` and `m_args_data` members with a single `m_scripted_metadata`, and update its constructor, Thread::QueueThreadPlanForStepScripted and the SB/CLI callers to take a `const ScriptedMetadata &` directly. Drop the now-redundant `args_sp` parameter from ScriptedThreadPlanInterface::CreatePluginObject; the Python override constructs a StructuredDataImpl from the metadata's args dict for the dispatched call.

ScriptedThreadPlan keeps its own `m_scripted_metadata` member because the interface's metadata is only set when `DidPush()` calls `CreatePluginObject`, so the plan needs to hold the metadata between construction and `DidPush()`.

### DIFF
--- a/lldb/include/lldb/Interpreter/Interfaces/ScriptedThreadPlanInterface.h
+++ b/lldb/include/lldb/Interpreter/Interfaces/ScriptedThreadPlanInterface.h
@@ -14,12 +14,11 @@
 #include "ScriptedInterface.h"
 
 namespace lldb_private {
-class ScriptedThreadPlanInterface : public ScriptedInterface {
+class ScriptedThreadPlanInterface : virtual public ScriptedInterface {
 public:
   virtual llvm::Expected<StructuredData::GenericSP>
   CreatePluginObject(const ScriptedMetadata &scripted_metadata,
-                     lldb::ThreadPlanSP thread_plan_sp,
-                     const StructuredDataImpl &args_sp) = 0;
+                     lldb::ThreadPlanSP thread_plan_sp) = 0;
 
   virtual llvm::Expected<bool> ExplainsStop(Event *event) { return true; }
 
--- a/lldb/source/API/SBThread.cpp
+++ b/lldb/source/API/SBThread.cpp
@@ -897,8 +897,13 @@ SBError SBThread::StepUsingScriptedThreadPlan(const char *script_class_name,
   Status new_plan_status;
   StructuredData::ObjectSP obj_sp = args_data.m_impl_up->GetObjectSP();
 
+  StructuredData::DictionarySP args_dict_sp;
+  if (obj_sp && obj_sp->GetType() == lldb::eStructuredDataTypeDictionary)
+    args_dict_sp = std::static_pointer_cast<StructuredData::Dictionary>(obj_sp);
+
+  ScriptedMetadata scripted_metadata(script_class_name, args_dict_sp);
   ThreadPlanSP new_plan_sp = thread->QueueThreadPlanForStepScripted(
-      false, script_class_name, obj_sp, false, new_plan_status);
+      false, scripted_metadata, false, new_plan_status);
 
   if (new_plan_status.Fail()) {
     error = Status::FromErrorString(new_plan_status.AsCString());
--- a/lldb/source/API/SBThreadPlan.cpp
+++ b/lldb/source/API/SBThreadPlan.cpp
@@ -65,9 +65,11 @@ SBThreadPlan::SBThreadPlan(lldb::SBThread &sb_thread, const char *class_name) {
   LLDB_INSTRUMENT_VA(this, sb_thread, class_name);
 
   Thread *thread = sb_thread.get();
-  if (thread)
-    m_opaque_wp = std::make_shared<ScriptedThreadPlan>(*thread, class_name,
-                                                       StructuredDataImpl());
+  if (thread) {
+    ScriptedMetadata scripted_metadata(class_name, {});
+    m_opaque_wp =
+        std::make_shared<ScriptedThreadPlan>(*thread, scripted_metadata);
+  }
 }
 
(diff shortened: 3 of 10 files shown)


lldb/source/API/SBThread.cpp
```diff
@@ -897,8 +897,13 @@ SBError SBThread::StepUsingScriptedThreadPlan(const char *script_class_name,
   Status new_plan_status;
   StructuredData::ObjectSP obj_sp = args_data.m_impl_up->GetObjectSP();
 
+  StructuredData::DictionarySP args_dict_sp;
+  if (obj_sp && obj_sp->GetType() == lldb::eStructuredDataTypeDictionary)
+    args_dict_sp = std::static_pointer_cast<StructuredData::Dictionary>(obj_sp);
```
bulbazord, line 902: Looks like you're tightening up the allowed kinds of StructuredData here too, like in the previous PR?

lldb/source/API/SBThreadPlan.cpp
```diff
@@ -65,19 +65,28 @@ SBThreadPlan::SBThreadPlan(lldb::SBThread &sb_thread, const char *class_name) {
   LLDB_INSTRUMENT_VA(this, sb_thread, class_name);
 
   Thread *thread = sb_thread.get();
-  if (thread)
-    m_opaque_wp = std::make_shared<ScriptedThreadPlan>(*thread, class_name,
-                                                       StructuredDataImpl());
+  if (thread) {
+    ScriptedMetadata scripted_metadata(class_name, {});
+    m_opaque_wp =
+        std::make_shared<ScriptedThreadPlan>(*thread, scripted_metadata);
```
bulbazord, line 71: I actually have a question about this... Why are we making a shared pointer and storing it into a weak pointer immediately? Weak pointers are non-owning, so who actually keeps this thing alive?

lldb/source/API/SBThreadPlan.cpp
```diff
@@ -65,19 +65,28 @@ SBThreadPlan::SBThreadPlan(lldb::SBThread &sb_thread, const char *class_name) {
...
-                                                       *args_data.m_impl_up);
+  if (thread) {
+    StructuredData::ObjectSP args_obj = args_data.m_impl_up->GetObjectSP();
+    StructuredData::DictionarySP args_dict_sp;
+    if (args_obj && args_obj->GetType() == lldb::eStructuredDataTypeDictionary)
+      args_dict_sp =
+          std::static_pointer_cast<StructuredData::Dictionary>(args_obj);
+    ScriptedMetadata scripted_metadata(class_name, args_dict_sp);
+    m_opaque_wp =
+        std::make_shared<ScriptedThreadPlan>(*thread, scripted_metadata);
```
bulbazord, line 88: Same here, doesn't m_opaque_wp expire almost immediately?

lldb/source/Target/ScriptedThreadPlan.cpp
```diff
@@ -92,8 +97,8 @@ void ScriptedThreadPlan::DidPush() {
 
 bool ScriptedThreadPlan::ShouldStop(Event *event_ptr) {
   Log *log = GetLog(LLDBLog::Thread);
-  LLDB_LOGF(log, "%s called on Scripted Thread Plan: %s )",
-            LLVM_PRETTY_FUNCTION, m_class_name.c_str());
+  LLDB_LOG(log, "{0} called on Scripted Thread Plan: {1} )",
+           LLVM_PRETTY_FUNCTION, GetScriptClassName());
```
bulbazord: Since you're already here, doesn't LLDB_LOG already handle function names in some capacity? Why do we need `LLVM_PRETTY_FUNCTION` here at all?

bulbazord (commented):
Conceptually fine, but there are a few things that raise some red flags in the existing code.

lldb/source/API/SBThreadPlan.cpp
medismailben: You're right to point this out but this was already here. @JDevlieghere changes the pimpl from sp to wp in 2ba7ce401e760990abb938c57e48c9d77d1b50de.

Also, I looked through the codebase and I couldn't find any test coverage or usage of this API which explains why we never hit this issue. At the same time, these constructors are part of the stable API so we can't really remove them.

lldb/source/API/SBThreadPlan.cpp
medismailben: Replied above

lldb/source/API/SBThread.cpp
medismailben: Yep

lldb/source/API/SBThreadPlan.cpp
bulbazord: Let's fix this in a follow-up

lldb/include/lldb/Interpreter/Interfaces/ScriptedThreadPlanInterface.h
```diff
@@ -14,12 +14,11 @@
 #include "ScriptedInterface.h"
 
 namespace lldb_private {
-class ScriptedThreadPlanInterface : public ScriptedInterface {
+class ScriptedThreadPlanInterface : virtual public ScriptedInterface {
```
bulbazord, line 17: diamonds are forever

bulbazord (approved):

lldb/include/lldb/Interpreter/Interfaces/ScriptedThreadPlanInterface.h
medismailben: 💎

PR #199065, opened 2026-05-21
[RISCV] Pass MachineOperand * to avoid copying. NFC

lukel97:
MachineOperand is 32 bytes according to clangd so passing a pointer will be cheaper


topperc (approved):
LGTM

PR #195537, opened 2026-05-03
[CIR] Atomic fetch-and-update operations on pointers

Lancern:
This patch adds support for atomic fetch-and-update operations on pointer inputs, including:

- C11 atomic operations: `__c11_atomic_fetch_add` and `__c11_atomic_fetch_sub` for pointer types;
- GNU atomic operations: `__atomic_fetch_add`, `__atomic_fetch_sub`, `__atomic_add_fetch`, and `__atomic_sub_fetch` for pointer types.

Assisted-by: Codex / GPT-5.5

PR #199066, opened 2026-05-21
[CIR] Fix atomic-fetch conflict:

erichkeane:
We had #195537 and #198871 both attempt to fix the same problem, in mildly different ways.  My patch (the latter) did so in a way that was rendered 'moot' by the former patch.

After looking into it further, I believe that the former is the 'correct' patch.

THIS patch is a partial revert of my patch, leaving the lambda (as I
    think it is a good 'readability' improvement), plus leaving/updating
the test.


erichkeane:
@Lancern FYI: No need to do anything here, but just so you know how this turned out.

andykaylor (approved):
lgtm

PR #199070, opened 2026-05-21
[llubi] Fix poison handling of shr exact with zero LHS

### LINKED ISSUE #198870: llubi: lshr/ashr exact poison test is incorrect
`lshr exact i1 false, true` should evaluate to poison, not 0

the test in `visitLShr()` isn't correct, `LHS.countr_zero() + 1` evaluates to 2, so the poison check is failing

```
~$ cat foo.ll
define i1 @main() {
  %x = lshr exact i1 false, true
  ret i1 %x
}
~$ alive-exec foo.ll

----------------------------------------
define i1 @main() {
#0:
  %x = lshr exact i1 0, 1
  ret i1 %x
}
Executing %#0
%x = poison
Returned #b0 / non-poison=false
ERROR: program returned poison
~$ llubi --verbose foo.ll
Entering function: main
  %x = lshr exact i1 false, true => F
  ret i1 %x
Exiting function: main
~$ 
```

dtcxzyw:
When the LHS is zero, and the exact flag is set on lshr/ashr, llubi cannot yield poison as  `LHS.countr_zero() + 1` is larger than the bit width.
Closes https://github.com/llvm/llvm-project/issues/198870.

Co-authored-by: John Regehr <regehr@cs.utah.edu>

### DIFF
--- a/llvm/test/tools/llubi/int_arith.ll
+++ b/llvm/test/tools/llubi/int_arith.ll
@@ -69,10 +69,12 @@ define void @main() {
   %lshr = lshr i8 127, 3
   %lshr_exact = lshr exact i8 126, 1
   %lshr_exact_poison = lshr exact i8 1, 1
+  %lshr_exact_poison_zero = lshr exact i8 0, 8
 
   %ashr = ashr i8 127, 3
   %ashr_exact = ashr exact i8 126, 1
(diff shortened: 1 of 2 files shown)


regehr (approved):
this looks good to me!

PR #199072, opened 2026-05-21
[AMDGPU] Use separate tables for VOPDX and VOPDY in `getCanBeVOPD` (NFC)

IgWod:
With this change a table for valid VOPDX (only for VOPD3X instructions) and VOPDY operations is generated through TableGen. This simplifies the look-up leading to a 2-3% compile-time speed-up for tested shaders where `getCanBeVOPD` is on a hot path.

Assisted-by: Claude Code

### DIFF
--- a/llvm/lib/Target/AMDGPU/SIInstrInfo.td
+++ b/llvm/lib/Target/AMDGPU/SIInstrInfo.td
@@ -3543,7 +3543,7 @@ def FP4FP8DstByteSelTable : GenericTable {
 def VOPDComponentTable : GenericTable {
   let FilterClass = "VOPD_Component";
   let CppTypeName = "VOPDComponentInfo";
-  let Fields = ["BaseVOP", "VOPDOp", "CanBeVOPDX", "CanBeVOPD3X"];
+  let Fields = ["BaseVOP", "VOPDOp", "CanBeVOPDX"];
   let PrimaryKey = ["BaseVOP"];
   let PrimaryKeyName = "getVOPDComponentHelper";
 }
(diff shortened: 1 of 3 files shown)


llvm/lib/Target/AMDGPU/SIInstrInfo.td
```diff
@@ -3572,6 +3572,22 @@ def lo16_to_vgpr32 : OutPatFrag<(ops node:$op),
 def hi16_to_vgpr32 : OutPatFrag<(ops node:$op),
   (REG_SEQUENCE VGPR_32, (i16 (IMPLICIT_DEF)), lo16, $op, hi16)>;
 
+def VOPDXTable : GenericTable {
```
shiltian, line 3578: Can you document what table X and Y mean here?

IgWod: Done, let me know if I need to expand it.

llvm/lib/Target/AMDGPU/SIInstrInfo.td
Sisyph: We have so many different VOPD tables. I would prefer to use an existing table like this rather than create the new ones (VOPDXTable..) Is it possible to use an existing one, like this? If not, it looks like CanBeVOPDX and CanBeVOPD3X would no longer be used. In that case could you remove them?

IgWod: You're right, I can re-use the existing table for the X instructions, it'll even save us one lookup. I'll clean it up and update the PR.

Sisyph (approved):
PR title and description may need an update since there is no separate table for VOPDX, but otherwise LGTM.

IgWod:
For newer targets VOP3DX is sub-target-dependent, so we cannot use `CanBeVOPD3X`. I have removed that field from the `VOPDComponentTable` and restored a VOPDX table that contains eligibility for VOP3DX instructions. VOPDX still uses `Info->CanBeVOPDX`. Let me know if there are any more comments based on that, otherwise I'm going to merge it.

llvm/lib/Target/AMDGPU/Utils/AMDGPUBaseInfo.cpp
```diff
@@ -657,28 +666,19 @@ unsigned getVOPDEncodingFamily(const MCSubtargetInfo &ST) {
 CanBeVOPD getCanBeVOPD(unsigned Opc, unsigned EncodingFamily, bool VOPD3) {
   bool IsConvertibleToBitOp = VOPD3 ? getBitOp2(Opc) : 0;
   Opc = IsConvertibleToBitOp ? (unsigned)AMDGPU::V_BITOP3_B32_e64 : Opc;
+  // Normalize through VOPDComponentTable so that e32, e64 and e64_dpp variants
```
rampitec, line 669: I think dpp versions shall be excluded from eligibility, though it is a separate change.

rampitec (approved):
LGTM

PR #188837, opened 2026-03-26
Changed stat passes to count instructions before and after optimizations

InakiVA:
Created this for instcount and func-properties-analysis to be able to see the change the optimization pipelines have on stats

PR #199073, opened 2026-05-21
Add Pre-Optimization FunctionPropertiesStatistics Pass

InakiVA:
Adding a PreOptimization pass for function properties that provides the relevant pre-opt pass information without leading to timeouts (also thanks to https://github.com/llvm/llvm-project/pull/198940 )

It previously led to timeouts so it got reverted
https://github.com/llvm/llvm-project/pull/188837


llvm/lib/Analysis/FunctionPropertiesAnalysis.cpp
```diff
@@ -282,6 +287,16 @@ FunctionPropertiesInfo FunctionPropertiesInfo::getFunctionPropertiesInfo(
   return FPI;
 }
 
+FunctionPropertiesInfo
+FunctionPropertiesInfo::getPreOptimizationFunctionPropertiesInfo(
```
aeubanks, line 291: probably want something like `getBasicFunctionPropertiesInfo()` or `getSimplifiedFunctionPropertiesInfo()`

llvm/lib/Analysis/FunctionPropertiesAnalysis.cpp
```diff
@@ -29,9 +29,14 @@ using namespace llvm;
 
 #define DEBUG_TYPE "func-properties-stats"
 
-#define FUNCTION_PROPERTY(Name, Description) STATISTIC(Num##Name, Description);
+#define FUNCTION_PROPERTY(Name, Description)                                   \
+  STATISTIC(Num##Name, Description);                                           \
+  STATISTIC(Num##Name##PreOptimization, Description " (before "                \
```
aeubanks, line 34: we shouldn't have pre-opt stats that are never updated, that's confusing if someone is looking through stats and sees 0 values

aeubanks (commented):
did you check that this resolves the compile time regression? adding this should basically be unnoticeable

did you ever figure out what was taking so long with the current version? it would be nice if there were just an easy fix

llvm/include/llvm/Analysis/FunctionPropertiesAnalysis.h
```diff
@@ -45,6 +45,11 @@ class FunctionPropertiesInfo {
   LLVM_ABI static FunctionPropertiesInfo
   getFunctionPropertiesInfo(Function &F, FunctionAnalysisManager &FAM);
 
+  // Does not consider domtree, loopinfo nor ir2vec because of bloat in
+  // nagivation that gets cleaned up anyways
```
boomanaiden154, line 49: s/navigation/preoptimization?

It would be good to understand exactly what the compile time issue was. Just constructing a dominator tree on pre-optimized IR should be fine. It will already be done by the pass pipeline. Same for all the other analyses.

PR #199076, opened 2026-05-21
[docs] Rewrite the ghlink Sphinx plugin to use doctree

rnk:
Previously we did a simple reST-based textual rewrite during the source-read Sphinx phase. This did the wrong thing for markdown sources, and would rewrite #GH123 in escaped code blocks.

The new implementations does a DFS walk over the doctree AST and does targetted replacements in text nodes matching the pattern. This makes it more accurate and portable between markup languages.

This does renumber the #idN anchors that refer back to the table of contents because inline reST links consume internal id numbers. However, the #id numbers aren't stable, they change any time we add a new heading, so I don't see this as a blocker.

An LLM did the first draft of the script, and I tuned it up from there.

### DIFF
--- a/clang/docs/ghlinks.py
+++ b/clang/docs/ghlinks.py
@@ -1,19 +1,86 @@
 #!/usr/bin/env python3
 
-from sphinx.application import Sphinx
+"""Script to link #GH123 to our github issue tracker.
+
+We use MyST and reST, so this Sphinx plugin operates on the shared doctree
+representation of the documentation, which is effectively a documentation-AST.
+
+Docutils doctree node reference:
+https://docutils.sourceforge.io/docs/ref/doctree.html
+"""
+
 import re
+from docutils import nodes
+from sphinx.application import Sphinx
 
 __version__ = "1.0"
 
+GH_LINK_RE = re.compile("#GH([0-9]+)")
+GH_LINK_URL = "https://github.com/llvm/llvm-project/issues/{}"
+SKIP_NODES: tuple[type[nodes.Node], ...] = (
+    nodes.FixedTextElement,
+    nodes.literal,
+    nodes.raw,
+    nodes.reference,
+)
 
-def subst_gh_links(app: Sphinx, docname, source):
-    regex = re.compile("#GH([0-9]+)")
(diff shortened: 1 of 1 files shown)


clang/docs/ghlinks.py
```diff
@@ -1,19 +1,86 @@
...
+"""
+
 import re
+from docutils import nodes
+from sphinx.application import Sphinx
 
 __version__ = "1.0"
 
+GH_LINK_RE = re.compile("#GH([0-9]+)")
+GH_LINK_URL = "https://github.com/llvm/llvm-project/issues/{}"
```
nigham: Nit: GH_LINK_TMPL?

clang/docs/ghlinks.py
```diff
@@ -1,19 +1,86 @@
...
+            replace_gh_links_in_subtree(child)
+
 
-def subst_gh_links(app: Sphinx, docname, source):
-    regex = re.compile("#GH([0-9]+)")
-    out_pattern = r"`#\1 <https://github.com/llvm/llvm-project/issues/\1>`_"
-    result = source[0]
-    result = regex.sub(out_pattern, result)
-    source[0] = result
+def subst_gh_links(_app: Sphinx, doctree: nodes.document) -> None:
```
nigham, line 121: Would be nice to add tests, even basic ones that LLM can add.

Also: how are we testing this works, is it via building the docs and verifying links are substituted correctly?

clang/docs/ghlinks.py
```diff
@@ -1,19 +1,86 @@
...
+https://docutils.sourceforge.io/docs/ref/doctree.html
+"""
+
 import re
+from docutils import nodes
+from sphinx.application import Sphinx
 
 __version__ = "1.0"
 
+GH_LINK_RE = re.compile("#GH([0-9]+)")
```
nigham: Maybe `#GH([1-9][0-9]+)` (since we don't want leading zeros)?

nigham (approved):
Broadly LGTM, couple nits.

clang/docs/ghlinks.py
rnk: Good point, I'll do that. Yes, I was just regenerating clang ReleaseNotes.html and spot checking a few URLs.

PR #199078, opened 2026-05-21
[GlobalISel] Implement parity reasoning via G_CTPOP

tejasgaikwad04:
This updates the implementation of PARITY in GlobalISel to use KnownBits::PARITY.

Progressing towards: #150515


tejasgaikwad04:
@arsenm Can you please review.

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
ZakyHermawan: Just use Op0

llvm/test/CodeGen/AArch64/GlobalISel/knownbits-ctpop-parity.mir
ZakyHermawan: Put new line at the end of file

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
ZakyHermawan: Just use Op1

pawan-nirpal-031: Remove redundant spaces.

pawan-nirpal-031, line 2184: Ditto.

tejasgaikwad04:
Hi @pawan-nirpal-031 , please review this

llvm/test/CodeGen/AArch64/GlobalISel/knownbits-ctpop-parity.mir
pawan-nirpal-031, line 1: Add tests for undef.

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
```diff
@@ -2173,15 +2173,12 @@ unsigned GISelValueTracking::computeNumSignBits(Register R,
       FirstAnswer = MRI.getType(R).getScalarSizeInBits();
       break;
     }
-
-    Register Src1 = MI.getOperand(1).getReg();
     unsigned Src1NumSignBits =
-        computeNumSignBits(Src1, DemandedElts, Depth + 1);
+        computeNumSignBits(Op0, DemandedElts, Depth + 1);
```
pawan-nirpal-031: clang-format.

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
arsenm: getVRegDef cannot fail for G_ operations. This should also use mi_match

arsenm: mi_match

jayfoad:
> This updates the implementation of PARITY in GlobalISel to use KnownBits::PARITY.

What?? Nothing in your patch calls KnownBits::parity or KnownBits::PARITY.

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
jayfoad, line 2161: This makes no sense to me.

jayfoad (changes requested):

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
tejasgaikwad04: Hi @jayfoad , GlobalISel doesn't have a `G_PARITY` opcode. In Generic MIR parity is represented as `(G_CTPOP x) & 1`, so this patch recognizes that pattern and applies the same reasoning as `KnownBits::parity()`. I'll update the description/comment to make that connection clearer.

arsenm: Why does SelectionDAG bother with a node if it's representable so simply? If it's really necessary should we introduce the G_ version? These should mostly mirror the DAG opcodes

arsenm: Just inline the lambda

arsenm: Use mi_match and no lambda

llvm/test/CodeGen/AArch64/GlobalISel/knownbits-ctpop-parity.mir
arsenm: Don't need IR section

llvm/lib/CodeGen/GlobalISel/GISelValueTracking.cpp
pawan-nirpal-031: @arsenm  I don't see a point of introducing a pseudo opcode, unless there are some pre-isel Target specific combines that it can be used for. @jayfoad can you help us understand why it might be useful to have such a generic pseudo op.

tejasgaikwad04: My understanding is that Generic MIR currently represents parity as (G_CTPOP x) & 1, which is why I matched that pattern here. This patch doesn't introduce a G_PARITY opcode. If parity should instead be modeled explicitly in GlobalISel to mirror SelectionDAG, I'm happy to investigate that approach.

tejasgaikwad04: Updated the requested chages

tejasgaikwad04: Updated the requested changes

tejasgaikwad04:
@jayfoad @arsenm, please review this

jayfoad (changes requested):
The whole patch still makes absolutely no sense. It looks like you're trying to implement `computeNumSignBits` for an expression like `ctpop(x) & 1`. Firstly, why would this need special handling? Generic handling for `&` should be able to do a good job. Secondly your implementation returns `getType(R).getScalarSizeInBits()` which would imply the value is 0 or -1 which is clearly not true.

Please take the time to try to understand what you are trying to do.

arsenm (changes requested):
Should this be closed? Special handling seems like it would only be necessary if / when G_PARITY is added

PR #199079, opened 2026-05-21
[AIX] Remove unsupported AIX native echo option -n

Himadhith:
AIX native echo doesn't support the `-n` flag.
Use printf instead to ensure the test works across all systems and making it portable.

The current test fails as follows:
```
FAIL: lit :: unit/Util.py (1 of 1)
******************** TEST 'lit :: unit/Util.py' FAILED ********************
Exit Code: 1

Command Output (stdout):
--
# RUN: at line 1
"/opt/freeware/bin/python3.12" /home/himadhit/llvm/community/build/utils/lit/tests/unit/Util.py
# executed command: /opt/freeware/bin/python3.12 /home/himadhit/llvm/community/build/utils/lit/tests/unit/Util.py
# | F..
# | ======================================================================
# | FAIL: test_basic (__main__.TestCommandCache.test_basic)
# | ----------------------------------------------------------------------
# | Traceback (most recent call last):
# |   File "/home/himadhit/llvm/community/build/utils/lit/tests/unit/Util.py", line 32, in test_basic
# |     self.assertEqual(lit_config.run_command_cached(["echo", "-n", "hi"]), b"hi")
# | AssertionError: b'-n hi\n' != b'hi'
# | 
# | ----------------------------------------------------------------------
# | Ran 3 tests in 2.092s
# | 
# | FAILED (failures=1)
```

### DIFF
--- a/llvm/utils/lit/tests/unit/Util.py
+++ b/llvm/utils/lit/tests/unit/Util.py
@@ -29,12 +29,12 @@ class TestCommandCache(unittest.TestCase):
     def test_basic(self):
         lit_config = self._lit_config()
 
-        self.assertEqual(lit_config.run_command_cached(["echo", "-n", "hi"]), b"hi")
+        self.assertEqual(lit_config.run_command_cached(["echo", "hi\\c"]), b"hi")
         self.assertNotEqual(lit_config.run_command_cached("ls"), None)
 
         # Test that arguments (e.g. text=True) get forwarded to subprocess.run
         self.assertEqual(
-            lit_config.run_command_cached(["echo", "-n", "hi"], text=True), "hi"
+            lit_config.run_command_cached(["echo", "hi\\c"], text=True), "hi"
         )
 
         # shell=True is not implied
--- a/llvm/utils/lit/tests/unit/Util.py
+++ b/llvm/utils/lit/tests/unit/Util.py
@@ -29,12 +29,12 @@ class TestCommandCache(unittest.TestCase):
     def test_basic(self):
         lit_config = self._lit_config()
 
-        self.assertEqual(lit_config.run_command_cached(["echo", "hi\\c"]), b"hi")
+        self.assertEqual(lit_config.run_command_cached(["printf", "hi"]), b"hi")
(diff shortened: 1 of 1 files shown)


ndrewh (approved):
Thanks! Appears to still work on macOS.

Himadhith:
For the test failure `LLVM.CodeGen/PowerPC/fast-isel-cmp-imm.ll`refer: https://github.com/llvm/llvm-project/pull/199053#issuecomment-4511354331

midhuncodes7 (approved):
LGTM

hubert-reinterpretcast:
@Himadhith, the PR description mentions something about using `\c`, but the PR changes don't seem to match?

Himadhith:
> @Himadhith, the PR description mentions something about using `\c`, but the PR changes don't seem to match?

First I tried to add `\c` but the test failed on linux build because linux does not recognise `\c` without the `-e` option passed to echo. But on AIX builtin echo `-e` was also not a supported option, so in the final code changes I had to modify it to use printf to have better portability. But the description remained to be explaining the previous changes, that's my bad. I will modify the description here, but do I need to revert the PR and make sure while merging in the PR the proper description is present?

hubert-reinterpretcast:
> but do I need to revert the PR and make sure while merging in the PR the proper description is present?

If it is not too much trouble, I would suggest that. TMK, you can post a PR with the revert and the the re-application (with corrected commit message) and push that to `main` from the command line.

Himadhith:
@hubert-reinterpretcast IIUC a PR to revert the commit with the title [Re-apply] [AIX] Remove unsupported AIX native echo option -n?

Himadhith:
If I revert the commit and also reapply it in the same PR, won't they cancel each other out and say no files are changed?

hubert-reinterpretcast:
> If I revert the commit and also reapply it in the same PR, won't they cancel each other out and say no files are changed?

Yes, it will. You can still rebase+push the commit to `main` from the command line after the PR is approved (i.e., **do not** use the GitHub UI to land the PR).

PR #199081, opened 2026-05-21
[AArch64] Lower cttz(bitcast <Nxi1> to iN) with shrn-based compressed movemask

as4230:
The existing lowering in vectorToScalarBitmask() creates a 1 bit per lane movemask using a powers of 2 reduction (and+addv with a constant pool entry).

This patch adds a DAG combine on ISD::CTTZ that recognizes cttz(bitcast <N x i1> to iN) and produces a compressed movemask with shrn (for i8 lanes) or xtn (for wider lanes) then runs scalar cttz on a 64- or 128-bit value. Dividing by bits per lane gives the lane index.

Supports lane counts {2, 4, 8, 16, 32} (one or two NEON registers)

For the example in the issue (`<16 x i8> -> i16`):

Before:
```asm
adrp x8, .LCPI0_0
cmlt v0.16b, v0.16b, #0
ldr  q1, [x8, :lo12:.LCPI0_0]
and  v0.16b, v0.16b, v1.16b
ext  v1.16b, v0.16b, v0.16b, #8
zip1 v0.16b, v0.16b, v1.16b
addv h0, v0.8h
fmov w8, s0
orr  w8, w8, #0x10000
rbit w8, w8
clz  w0, w8
ret
```
After:

```asm
cmlt v0.16b, v0.16b, #0
shrn v0.8b,  v0.8h,  #4
fmov x8, d0
rbit x8, x8
clz  x8, x8
lsr  x0, x8, #2
ret
```

Also created a new test file with 24 functions covering both endianness

- 7 basic widths (`<16 x i8>` down through `<2 x i32>`)
- 4 wider cases that span two NEON registers (`<32 x i8>` etc)
- 2 narrow cases that get sext'd up to fit one register
- 2  cases where the bitcast has two users 
- 4 alternative comparison operators (`eq`, `ne`, etc)
- 1 `cttz` with `is_zero_undef=true`
- 4 negative tests that bails for unsupported shapes 

Fixes #186295


llvm/test/CodeGen/AArch64/cttz-of-bool-vector-bitcast.ll
```diff
@@ -0,0 +1,853 @@
...
+  ret i16 %ctz
+}
+
+define i8 @cttz_v8i32(<8 x i32> %v) {
+; CHECK-LE-LABEL: cttz_v8i32:
+; CHECK-LE:       // %bb.0:
+; CHECK-LE-NEXT:    cmlt v1.4s, v1.4s, #0
+; CHECK-LE-NEXT:    cmlt v0.4s, v0.4s, #0
+; CHECK-LE-NEXT:    uzp1 v0.8h, v0.8h, v1.8h
+; CHECK-LE-NEXT:    mov x8, v0.d[1]
```
efriedma-quic: Does it make sense to perform another vector narrowing step here (for cttz_v8i32)?

I know I suggested it originally, but I wasn't considering using shrn like this patch does.

llvm/test/CodeGen/AArch64/cttz-of-bool-vector-bitcast.ll
```diff
@@ -0,0 +1,853 @@
...
+; CHECK-LE-NEXT:    ret
+;
+; CHECK-BE-LABEL: cttz_v16i8:
+; CHECK-BE:       // %bb.0:
+; CHECK-BE-NEXT:    rev64 v0.16b, v0.16b
+; CHECK-BE-NEXT:    ext v0.16b, v0.16b, v0.16b, #8
+; CHECK-BE-NEXT:    cmlt v0.16b, v0.16b, #0
+; CHECK-BE-NEXT:    rev16 v0.16b, v0.16b
+; CHECK-BE-NEXT:    shrn v0.8b, v0.8h, #4
+; CHECK-BE-NEXT:    rev64 v0.8b, v0.8b
```
efriedma-quic: For big-endian, maybe consider using AArch64ISD::NVCAST to avoid some of these REV operations?  Or add an explicit bailout for big-endian if you don't want to consider it.  The current result is pretty messy.

llvm/test/CodeGen/AArch64/cttz-of-bool-vector-bitcast.ll
as4230: Yep, extended the loop to keep narrowing to 64 bits when we can. v8i32, v16i16 and v4i64 now use a single i64 cttz.

llvm/test/CodeGen/AArch64/cttz-of-bool-vector-bitcast.ll
as4230: Switched to NVCAST as you suggested. The rev16/rev64 from bitcasts are gone and replaced with one rbit from CTTZ lowering.

as4230:
ping

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -26217,6 +26218,133 @@ static SDValue vectorToScalarBitmask(SDNode *N, SelectionDAG &DAG) {
...
+  auto CompressedCttz = [&](SDValue V) -> SDValue {
+    if (V.getValueType().getSizeInBits() == DRegBits) {
+      SDValue Nv = DAG.getNode(AArch64ISD::NVCAST, DL, MVT::v1i64, V);
+      SDValue Scalar = DAG.getNode(ISD::EXTRACT_VECTOR_ELT, DL, MVT::i64, Nv,
+                                   DAG.getConstant(0, DL, MVT::i64));
+      return DAG.getNode(ISD::CTTZ, DL, MVT::i64, Scalar);
+    }
+    SDValue Compressed = DAG.getBitcast(MVT::i128, V);
+    // BE puts lane 0 at the high end of the scalar after the bitcast, so we use
+    // clz instead of cttz to find the first match
```
efriedma-quic: "BE puts lane 0 at the high end of the scalar after the bitcast", sure... but why is this not relevant for 64-bit registers?

as4230: The asymmetry is intentional but the comment doesn't explain it well now that we have NVCAST. I updated the comment to clarify the BITCAST specific behavior.

Just to be clear...

NVCAST is a no op register cast. Lane 0 stays at the lowest significant bit regardless of endianness so the 64-bit path uses CTTZ uniformly. The 64-bit path works because i64 is a native scalar but i128 isnt so the 128-bit case has to be split into two i64s. BITCAST + i128 CTTZ lets the legalizer do that split. BITCAST inserts a byte-reverse on BE so it puts lane 0 at the high end of the i128 hence CLZ on BE for the 128-bit path.

I tried eliminating the rev64 left in the 128 bit case with manual NVCAST + two-halves CTTZ but it wouldn't help BE because switching from CTLZ to CTTZ (which lowers to rbit + clz) trades 1 rev64 for 2 rbits. The current pass still saves 7 instructions on v32i8 BE (24 -> 17) compared to no fold so not worth throwing out BE over.

efriedma-quic: Oh, I see, that makes sense.

efriedma-quic:
(The latest version looks like it's failing on buildbot.)

as4230:
> (The latest version looks like it's failing on buildbot.)

Build passes now. I had to rebase and regen tests. There was a addv optimization that affected some test regen.

as4230:
ping

as4230:
Ping. @efriedma-quic @davemgreen

llvm/test/CodeGen/AArch64/cttz-of-bool-vector-bitcast.ll
```diff
@@ -0,0 +1,818 @@
...
+; CHECK-BE-NEXT:    lsr x0, x8, #4
+; CHECK-BE-NEXT:    // kill: def $w0 killed $w0 killed $x0
+; CHECK-BE-NEXT:    ret
+  %cmp = icmp slt <4 x i64> %v, splat (i64 0)
+  %bm = bitcast <4 x i1> %cmp to i4
+  %ctz = call i4 @llvm.cttz.i4(i4 %bm, i1 false)
+  ret i4 %ctz
+}
+
+define i32 @cttz_v32i8(<32 x i8> %v) {
```
efriedma-quic, line 582: If I'm understanding the code correctly, it should also trigger on `<32 x i16>` etc.  Would it be useful to add tests for that?

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -26247,6 +26248,133 @@ static SDValue vectorToScalarBitmask(SDNode *N, SelectionDAG &DAG) {
...
+  // Narrow toward CompressedBits. shrn-pair is only used once since it
+  // leaves the i8 lanes already packed.
+  SDValue Cur = Mask;
+  bool UsedShrnPair = false;
+  while (Cur.getValueType().getSizeInBits() > CompressedBits) {
+    EVT CurVT = Cur.getValueType();
+    unsigned CurLaneBits = CurVT.getScalarSizeInBits();
+    if (CurLaneBits == MinLaneBits) {
+      if (UsedShrnPair)
+        break;
```
efriedma-quic: This "break" is confusing... if we don't reach the target number of bits, CompressedBits is the wrong value, I think?  I'm not sure if we can actually hit it, though; if we get to this point, something went wrong with the CompressedBits computation, I think.

llvm/test/CodeGen/AArch64/cttz-of-bool-vector-bitcast.ll
as4230: Yep, added tests for v32i16 and v16i32. I also pulled the original tests into a precommit commit so you can see the before and after codegen as a clean diff while reviewing.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
as4230: You're right. It should never run and if it did the result would be wrong. The reason it was there is that xtn needs the loop since narrowing wide lanes can take a few steps but the shrn only ever runs once. I had it inside the loop with that bool guard to stop a second shrn but since one shrn always lands exactly on CompressedBits the loop never comes back around to it anyways. I pulled the shrn out into an if after the loop. Now it can only ever run once so the bool guard and the break are both gone.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -26715,6 +26716,130 @@ static SDValue vectorToScalarBitmask(SDNode *N, SelectionDAG &DAG) {
...
+      SDValue Lo = DAG.getNode(ISD::EXTRACT_SUBVECTOR, DL, HalfVT, Cur,
+                               DAG.getConstant(0, DL, MVT::i64));
+      SDValue Hi = DAG.getNode(
+          ISD::EXTRACT_SUBVECTOR, DL, HalfVT, Cur,
+          DAG.getConstant(HalfVT.getVectorNumElements(), DL, MVT::i64));
+      Cur = DAG.getNode(ISD::CONCAT_VECTORS, DL, HalfVT, ShrnPair(Lo),
+                        ShrnPair(Hi));
+    }
+  }
+
```
efriedma-quic, line 26834: `assert(Cur.getValueType().getSizeInBits() == CompressedBits)`?

as4230: Done

efriedma-quic (approved):
LGTM

as4230:
Thanks so much for the reviews and the patience on this one @efriedma-quic . I don't have commit access, so would you be able to land it for me?

efriedma-quic:
It looks like your email address is set to private; see https://llvm.org/docs/DeveloperPolicy.html#email-addresses .

as4230:
> It looks like your email address is set to private; see https://llvm.org/docs/DeveloperPolicy.html#email-addresses .

Fixed. Local commits had my email but github account level email was private. Made public.

PR #199082, opened 2026-05-21
[PowerPC][NFC] Change arguments of PPCustomInserter

redstar:
Removes the `asmstring` argument, and gives the `pattern` argument an empty default value.
The `asmstring` is only accessible via `getMnemonic()`, which is only used in the `PPCInstPrinter`. Thus these strings are never emitted, and can be removed.

### DIFF
--- a/llvm/lib/Target/PowerPC/PPCInstrFormats.td
+++ b/llvm/lib/Target/PowerPC/PPCInstrFormats.td
@@ -2398,8 +2398,8 @@ class PPCEmitTimePseudo<dag OOL, dag IOL, string asmstr, list<dag> pattern>
 
 // Instruction that require custom insertion support
 // a.k.a. ISelPseudos, however, these won't have isPseudo set
-class PPCCustomInserterPseudo<dag OOL, dag IOL, string asmstr,
-                              list<dag> pattern>
+class PPCCustomInserterPseudo<dag OOL, dag IOL, list<dag> pattern = [],
+                         string asmstr = "#" # NAME>
     : PPCEmitTimePseudo<OOL, IOL, asmstr, pattern> {
   let usesCustomInserter = 1;
 }
--- a/llvm/lib/Target/PowerPC/PPCInstrFormats.td
+++ b/llvm/lib/Target/PowerPC/PPCInstrFormats.td
@@ -2398,9 +2398,8 @@ class PPCEmitTimePseudo<dag OOL, dag IOL, string asmstr, list<dag> pattern>
 
 // Instruction that require custom insertion support
 // a.k.a. ISelPseudos, however, these won't have isPseudo set
-class PPCCustomInserterPseudo<dag OOL, dag IOL, list<dag> pattern = [],
-                         string asmstr = "#" # NAME>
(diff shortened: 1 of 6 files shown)


llvm/lib/Target/PowerPC/PPCInstrInfo.td
```diff
@@ -3059,7 +3049,7 @@ def : InstAlias<"mtcr $rA", (MTCRF 255, gprc:$rA)>;
 let Predicates = [HasFPU] in {
 // Custom inserter instruction to perform FADD in round-to-zero mode.
 let Uses = [RM], mayRaiseFPException = 1 in {
-  def FADDrtz: PPCCustomInserterPseudo<(outs f8rc:$FRT), (ins f8rc:$FRA, f8rc:$FRB), "",
+  def FADDrtz: PPCCustomInserterPseudo<(outs f8rc:$FRT), (ins f8rc:$FRA, f8rc:$FRB),
                       [(set f64:$FRT, (PPCany_faddrtz f64:$FRA, f64:$FRB))]>;
```
lei137: previously this was set as a null string?  What does that mean for this pseudo?

redstar: The pseudo instructions have different lifetime. They "die" either in the custom inserter, the post-ra cleanup or in the asm printer. The name is what you see in MIR when you look at the output of a pass. An empty name prints, well, nothing. Not sure what this means for using such a MIR file as input for the another pass.

redstar: Changed, that should be NFC.

redstar: I changed the implementation again. See the updated description: the `asmstring` is never emitted. I removed it to save space.

llvm/lib/Target/PowerPC/PPCInstrFormats.td
```diff
@@ -2398,9 +2398,8 @@ class PPCEmitTimePseudo<dag OOL, dag IOL, string asmstr, list<dag> pattern>
 
 // Instruction that require custom insertion support
 // a.k.a. ISelPseudos, however, these won't have isPseudo set
-class PPCCustomInserterPseudo<dag OOL, dag IOL, string asmstr,
-                              list<dag> pattern>
-    : PPCEmitTimePseudo<OOL, IOL, asmstr, pattern> {
+class PPCCustomInserterPseudo<dag OOL, dag IOL, list<dag> pattern = []>
+    : PPCEmitTimePseudo<OOL, IOL, "", pattern> {
```
lei137, line 2402: You seem to have missed reverting this since emit time pseudos still require the asmstr param?

redstar: This class hierarchy or the name is unfortunate.
A real `PPCEmitTimePseudo` needs a name, that is true. But here a `PPCCustomInserterPseudo` is being defined, which does not need a name. This is realized by passing in `""`.

I was thinking about changing the hierarchy to avoid this confusion. I would do this in a follow-up PR.

PR #199083, opened 2026-05-21
workflows/require-release-manager: Refactor to check an arbitrary team

tstellar:
This will allow it to be used for checking that users are members of the llvm-committer team or possibly others.


rnk (approved):
Looks good to me.

boomanaiden154 (approved):
LGTM.

PR #199084, opened 2026-05-21
[GlobalISel][AArch64] Be conservative to fix incorrect MIFlags on instructions widened by the legalizer.

### LINKED ISSUE #192675: [GlobalISel] Incorrect MIFlags on instructions widened by the legalizer
Consider an arithmetic instruction that is too narrow for some target, for example:
```llvm
%2:_(s8) = nsw nuw G_ADD %0(s8), %1(s8)
```
assuming the smallest add legal for the target is a 32-bit one, this would be widened by the legalizer as:
```llvm
%3:_(s32) = G_ANYEXT %0(s8)
%4:_(s32) = G_ANYEXT %1(s8)
%5:_(s32) = nsw nuw G_ADD %3(s32), %4(s32)
%2:_(s8) = G_TRUNC %5(s32)
```
However, the legalizer doesn't drop the `nsw`/`nuw` flags from the instruction, even though it's not guaranteed that the arithmetic will not overflow with the newly inserted `G_ANYEXT` instructions as the upper 24 bits of the operands could be anything.

https://godbolt.org/z/4rr53sYbG

ZakyHermawan:
Be conservative on widen scalar by dropping `nsw` and `nuw` flags for `G_ADD`, `G_SUB`, and `G_MUL`.

Fixes: https://github.com/llvm/llvm-project/issues/192675


ZakyHermawan:
@davemgreen @arsenm 
Can I get a review on this PR ? Thanks

PR #199087, opened 2026-05-21
workflows/libcxx-run-benchmarks: Only run job for people with commit access

tstellar:
This job checks out untrusted code from a PR in a trusted context (issue_comment trigger), so we need to limit it to people with commit access to avoid possible privilege escalation.


tstellar:
This also includes changes from https://github.com/llvm/llvm-project/pull/199083

rnk (approved):
Looks good to me. 

As if anyone is truly qualified to review github action YAML and predict what it does until it is running in production. =/

boomanaiden154 (approved):
LGTM.

ldionne (commented):
Naive question, but is there no way to make this run against the version of the `built-at-commit` & friends script on the `main` branch instead, like we do for PRs?

boomanaiden154:
> Naive question, but is there no way to make this run against the version of the built-at-commit & friends script on the main branch instead, like we do for PRs?

Not exactly sure what you mean. You can check out whatever version of the scripts/code that you want.

We need a permissions check here because the workflow can access secrets.

ldionne:
Ah, I thought the issue was primarily that the action was running scripts checked out from the PR version of the code instead of the `main` version of the code, but seems like I misunderstood.

PR #193298, opened 2026-04-21
[clang][CodeGen][X86_64] Honor per-function AVX ABI in C/C++ call paths, maintain old psABI for PlayStation.

freaknbigpanda:
Fixes https://github.com/llvm/llvm-project/issues/64706

Wire per function x86 AVX ABI level into CodeGen arrangement methods so __attribute(__target("avx"))) / __attribute(__target("avx512f"))) on methods, ctors, and free-functions affects ABI lowering consistently.
    
Specifically:
- Added X86AVXABILevel member to CGFunctionInfo.
- Populated X86AVXABILevel member in CGFunctionInfo objects via arrangement methods declared in CodeGenTypes.h.
- Respect CGFunctionInfo AVX Level in X86_64ABIInfo::computeInfo.
- Add/extend regression tests for:
- free-function target-attribute AVX ABI lowering
- C++ method/ctor target-attribute AVX ABI lowering
- PS4/PS5 legacy ABI behavior (no per-function AVX ABI change)

Issue #128361, opened 2025-02-22
[Clang] -Wpsabi complains too eagerly


philnik777:
Clang currently complains when calling a function that the ABI changes if a feature is enabled, but ignores that the function is marked `[[gnu::always_inline]]`, which ensures that the type will never be part of the ABI. e.g.
```c++
using vec [[gnu::vector_size(32)]] = int;

[[gnu::always_inline]] vec test() {
  return {};
}

void call() {
  (void)test(); // complains that test() changes the ABI with avx enabled
}
```

Ironically, it doesn't complain if the address is taken, which is one of the few cases where this would actually matter, since LLVM might not be able to inline a call to the function pointer.

PR #199091, opened 2026-05-21
[clang][X86] Emit AVX level mismatch psABI warnings on function definitions

freaknbigpanda:
Emit -WpsABI for x86_64 function definitions whose return type or parameter type uses a vector wider than 128 bits without the required ABI feature enabled. 256-bit vectors require avx, and 512-bit vectors require avx512f.

Previously this diagnostic was only emitted at call sites, so definitions with wide vector signatures could be introduced without a warning until they were called. Use the function feature map so
attribute(target("avx/avx512f")) definitions are accepted, and emit no warnings for prototype-only declarations.


efriedma-quic:
Can you share examples of where this is triggering in libc/libc++?

freaknbigpanda:
> Can you share examples of where this is triggering in libc/libc++?

You can see an example of the libcxx test failures here: https://github.com/llvm/llvm-project/actions/runs/26239342166/job/77221171905

And an example of the libc build failure here: https://github.com/llvm/llvm-project/actions/runs/26183622799/job/77033061201

freaknbigpanda:
It seems like these failures have been mostly fixed with the application of -Wno-psabi but we should check to make sure there isn't something actually wrong with libc and libcxx before we merge this. The code should be fixed to avoid the warnings

jhuber6:
> It seems like these failures have been mostly fixed with the application of -Wno-psabi but we should check to make sure there isn't something actually wrong with libc and libcxx before we merge this. The code should be fixed to avoid the warnings

Why should this cause a warning? The original ABI warning was due to symbols escaping the TU and these are all static?

jhuber6:
I don't think adding `-Wno-psabi` is a valid solution here

efriedma-quic:
As discussed on #193298, gcc has adopted a model where the passing of AVX-size vectors depends on the caller's attributes: if the caller is target("avx"), use an AVX register, otherwise use a pair of SSE registers.  It maintains self-consistency using -Wps-abi: if it would use a pair of SSE registers, it always warns.

This isn't the only way we could handle AVX-size vectors.  We could instead make the AVX-ness part of the function type, which would allow code to use the non-AVX convention even if AVX is enabled.  But this introduces significant complexity into the type system, and isn't gcc-compatible.

Given we have a gcc-style model, we have to generate warnings on definition which are either externally visible, or address-taken; otherwise we can silently generate code using mismatched calling conventions.

We could possibly carve out an exception for `static inline` functions... but I don't think that covers the cases that are failing.  And it's not what gcc does.

jhuber6:
> We could possibly carve out an exception for `static inline` functions... but I don't think that covers the cases that are failing. And it's not what gcc does.

The fact that this required suppressing warnings globally in libc and libc++, which both have very intentional usage of these vectors, is a very big red flag to me. This seems to just turn on a "You used a wide vector!" warning for everyone, which isn't very helpful in my opinion.

For `libmvec` we intentionally upscale from a `f32x8` to a `f64x8` which will now always fire a warning if I understand correctly. The original intent of these warnings from what I understood was mixing ABIs accross TU boundaries, which is why we force all of these to be static.

I don't have the full context here, but clang vectors are much more powerful than GCC vectors so I don't think we necessarily need to be beholden to them.

jhuber6:
I think following GCC's warnings is too strict for how we want to use these vectors. Can't we just change the warnings to only emit on definitions that escape the TU?

boomanaiden154:
> For libmvec we intentionally upscale from a f32x8 to a f64x8 which will now always fire a warning if I understand correctly. The original intent of these warnings from what I understood was mixing ABIs accross TU boundaries, which is why we force all of these to be static.

You can run into ABI issues within a single translation unit too because the calling convention is dependent upon the caller's features. e.g. if `foo` is marked `avx512f` and calls `baz` (with a 512-bit vector type), it will use zmm registers. If in the same translation unit you have `bar` which doesn't have any `avx512f` features and it calls `baz`, it will pass the argument on the stack.

But given LLVM libc and libc++ happen to work, I would assume there aren't massive ABI issues like that and we shouldn't be warning in those cases.

jhuber6:
> You can run into ABI issues within a single translation unit too because the calling convention is dependent upon the caller's features. e.g. if `foo` is marked `avx512f` and calls `baz` (with a 512-bit vector type), it will use zmm registers. If in the same translation unit you have `bar` which doesn't have any `avx512f` features and it calls `baz`, it will pass the argument on the stack.
> 

Using `[[gnu::target]]` at all is already opening yourself up for massive ABI headaches. This use case of double-pumping the existing vector ABI in a single TU should be well defined and I don't think we should warn on it.

boomanaiden154:
> Using [[gnu::target]] at all is already opening yourself up for massive ABI headaches. This use case of double-pumping the existing vector ABI in a single TU should be well defined and I don't think we should warn on it.

Do you have other examples? I'm only aware of this one.

It might be well-defined, but there are certainly instances where we should be warning. If a function expects vectors to be passed on the stack and someone calls that function such that vectors are passed in registers, we should probably be warning in that case, and that's the intention of `-Wps-abi` in gcc as far as I'm aware.

libcxx/utils/libcxx/test/params.py
```diff
@@ -75,6 +75,8 @@
 
     # We're not annotating all the APIs, since that's a lot of annotations compared to how many we actually care about
     "-Wno-nullability-completeness",
+    # Adding this for now to get the build to pass but needs more detailed review
+    "-Wno-psabi",
```
philnik777: Why is this required?

jhuber6: From what I understand, the referenced PR changes the ABI warning to now trigger on any possible use of a vector width that does not strictly conform to the ABI. Both libc and libcxx seem to use these and compile with `-Werror` so this PR disables those warnings.

freaknbigpanda: I haven't had a chance to look into the failures in detail but I don't think that applying -Wno-psabi globally is the right thing to do. Ideally the code would be fixed so that the warning is not emitted. If it turns out that we can't avoid emitting the warning then we should be using pragmas to disable it just for the function definitions that need the warning suppressed.  

I think I should have made this a draft PR to make this more clear. I will convert it now.

Later this week I will take a deeper look at the failures to try to see exactly why libc and libcxx are emitting these warnings and suggest a path forward.

jhuber6: For `libc` I can say it's ilkely here, and there will be more cases like this as the vector math library is expanded https://github.com/llvm/llvm-project/blob/main/libc/src/__support/mathvec/expf.h#L75. The use-case is that we cast from f32 -> f64 for a higher precision operation and turn one f32 vector register into two f64 ones. The fact that we do not need to manually extract and shuffle to do this is the entire reason Clang's vector  types are useful and I'd prefer this doesn't get restricted.

freaknbigpanda: OK I think I found a pretty good solution, I am now guarding the emission of these warnings with: 
```
// psABI warnings & errors for function definitions that are only visible 
// in this translation unit are handled at call site by checkFunctionCallABI
if (!FD->isExternallyVisible())
  return;
```
Which fixes everything for libc, for libcxx there were already pragmas in the offending file so I expanded those to include the additional functions

jhuber6: I'm going to assume those pragmas were put there specifically because of https://github.com/llvm/llvm-project/issues/128361 which I fixed in https://github.com/llvm/llvm-project/pull/157570. Maybe @philnik777 remembers, but it's probably better for them to be gone if so.

freaknbigpanda: I tried removing the pragmas and even without any of the changes in this PR it still warns on the lambda call inside this function:

```
template <class _VecT>
[[__nodiscard__]] _LIBCPP_ALWAYS_INLINE _LIBCPP_HIDE_FROM_ABI _VecT
__broadcast(__simd_vector_underlying_type_t<_VecT> __val) {
  return [&]<std::size_t... _Indices>(index_sequence<_Indices...>) {
    return _VecT{((void)_Indices, __val)...};
  }(make_index_sequence<__simd_vector_size_v<_VecT>>());
}
```
```
/home/bluke/dev/llvm-project/build/include/c++/v1/__algorithm/simd_utils.h:132:10: warning: AVX vector return of type 'char __attribute__((ext_vector_type(32)))' (vector of 32 'char' values) without 'avx' enabled changes the ABI [-Wpsabi]
    132 |   return [&]<std::size_t... _Indices>(index_sequence<_Indices...>)
```

This lambda is returning a 256 bit vector without AVX enabled, and both `!Callee.isExternallyVisible()` and `Callee.hasAttr<AlwaysInlineAttr>()` return false for this lambda so https://github.com/llvm/llvm-project/pull/157570 wouldn't catch this case. Maybe this lambda declaration could be marked with _LIBCPP_ALWAYS_INLINE? That would prevent the call site warning. Anyways right now we still need the pragma. 

My change introduces 6 new warnings in addition to this one that was emitted before:

1. Warns on the same lambda above but this time the lambda definition, not call site
2. Warns on __broadcast definition since it returns 256bit vectors for non avx targets
3. Warns on __partial_load definition since it returns 256bit vectors for non avx target
4. Warns on lambda defined in __partial_load since it returns 256bit vector
5. Warns on __none_of because the first param is a 256bit vector
6. Warns on __find_first_set for the same reason

Guarding the warnings on `FD::hasAttr<AlwaysInlineAttr>()` would get rid of the first four warnings (__none_of and __find_first_set are not marked _LIBCPP_ALWAYS_INLINE)  but I believe I still need to emit warnings even if `FD::hasAttr<AlwaysInlineAttr>() == true` because AFAIK AlwaysInlineAttr doesn't mean that there can't be an externally visible out of line definition emitted somewhere.

freaknbigpanda: @philnik777 / @jhuber6  I asked ChatGPT Codex to come up with an example that uses public C++ STL APIs to create a situation where __find_first_set from __algorithm/simd_utils.h is instantiated in two different TUs with the same mangled name but different ABIs, thus creating potential runtime ABI mismatch problems, and it was able to come up with an example. So I think the new warnings that my PR introduces are highlighting a real potential issue that users may encounter in libcxx. I think it is rare enough that it probably hasn't been hit before but I think this is something that the libcxx maintainers should try to address. Let me know if you think I should raise an issue for this or contact someone else.

I think this is actually good motivation to merge this PR since library authors can avoid issues like this if they get prior warnings via diagnostics.

The example from codex is pasted below, basically both cpps instantiate __find_first_set versions with the same mangled names but different ABI and then the link order determines if the call to avx_public_mismatch() in main.cpp calls the right instantiation. If it calls the wrong one it doesn't return the expected result, 55.

```
bluke@bluke-ubuntu:~/dev/libcxx-abi-mismatch$ cat mismatch_avx.cpp 
#include <algorithm>

extern "C" int avx_public_mismatch() {
  alignas(32) char lhs[96] = {};
  alignas(32) char rhs[96] = {};
  rhs[55] = 7;

  auto result = std::mismatch(lhs, lhs + 96, rhs);
  return static_cast<int>(result.first - lhs);
}
bluke@bluke-ubuntu:~/dev/libcxx-abi-mismatch$ cat numget_noavx.cpp 
#include <ios>
#include <locale>

extern "C" const char* force_num_get_locale_instantiation(const char* first,
                                                          const char* last,
                                                          std::ios& io,
                                                          long& value) {
  std::locale loc(std::locale::classic(), new std::num_get<char, const char*>(1));
  const std::num_get<char, const char*>& facet =
      std::use_facet<std::num_get<char, const char*> >(loc);

  std::ios_base::iostate err = std::ios_base::goodbit;
  return facet.get(first, last, io, err, value);
}
bluke@bluke-ubuntu:~/dev/libcxx-abi-mismatch$ cat main.cpp
#include <ios>

extern "C" const char* force_num_get_locale_instantiation(const char*, const char*, std::ios&, long&);
extern "C" int avx_public_mismatch();

int main() {
  std::ios io(nullptr);
  long parsed = 0;
  const char text[] = "123";
  force_num_get_locale_instantiation(text, text + 3, io, parsed);

  int result = avx_public_mismatch();
  return result == 55 ? 0 : result + 1;
}
bluke@bluke-ubuntu:~/dev/libcxx-abi-mismatch$ cat build.sh
CXX=~/dev/llvm-project/build-ninja-release/bin/clang++
INC="-nostdinc++ -I /home/bluke/dev/llvm-project/build/include/c++/v1"
LIB="-L /home/bluke/dev/llvm-project/build/lib -Wl,-rpath,/home/bluke/dev/llvm-project/build/lib"

$CXX -std=c++20 -O0 $INC -c numget_noavx.cpp -o numget_noavx.o
$CXX -std=c++20 -O0 -mavx $INC -c mismatch_avx.cpp -o mismatch_avx.o
$CXX -std=c++20 -O0 $INC -c main.cpp -o main.o

$CXX -stdlib=libc++ $LIB \
  numget_noavx.o \
  mismatch_avx.o \
  main.o \
  -o repro_bad

./repro_bad
echo $?
bluke@bluke-ubuntu:~/dev/libcxx-abi-mismatch$ 
bluke@bluke-ubuntu:~/dev/libcxx-abi-mismatch$ ./build.sh
33
```

freaknbigpanda:
@philnik777 @jhuber6 Can either of you guys help to review this? If not can you recommend who I should poke?

jhuber6:
> @philnik777 @jhuber6 Can either of you guys help to review this? If not can you recommend who I should poke?

I can conclusively say that the `libc` portion I consider valid usage and I believe the changes have addressed my concerns. I think that the libcxx case should probably make these static similarly for this reason (It does bloat things, but I think SIMD ABIs justify it.) I will leave that to @philnik777 however.

freaknbigpanda:
@efriedma-quic @AaronBallman Can you guys help to review this or point me in the direction of someone who can?

philnik777 (approved):

philnik777 (commented):
(I'm only approving the libc++ part)

jhuber6 (approved):
I'm fine since this doesn't trigger libc's case anymore.

clang/lib/CodeGen/Targets/X86.cpp
```diff
@@ -1571,6 +1575,65 @@ static bool checkAVXParam(DiagnosticsEngine &Diag, ASTContext &Ctx,
...
+      CGM.getDiags().Report(Loc, diag::warn_avx_calling_convention)
+          << !IsReturn << Ty << Feature;
+      return true;
+    }
+
+    return false;
+  };
+
+  // psABI warnings & errors for function definitions that are only visible
+  // in this translation unit are handled at call site by checkFunctionCallABI
```
AaronBallman:
```suggestion
  // in this translation unit are handled at call site by checkFunctionCallABI.
```

clang/lib/CodeGen/Targets/X86.cpp
```diff
@@ -1571,6 +1575,65 @@ static bool checkAVXParam(DiagnosticsEngine &Diag, ASTContext &Ctx,
...
+  if (!FD->isExternallyVisible())
+    return;
+
+  // First check the return type and emit diagnostic if required
```
AaronBallman:
```suggestion
  // First check the return type and emit diagnostic if required.
```

clang/lib/CodeGen/Targets/X86.cpp
```diff
@@ -1571,6 +1575,65 @@ static bool checkAVXParam(DiagnosticsEngine &Diag, ASTContext &Ctx,
...
+  Check(FD->getReturnType(), GetReturnTypeLoc(FD), true);
+
+  // Go through the parameters and emit a warning for the first vector found
+  // without the matching function AVX level attribute
```
AaronBallman:
```suggestion
  // without the matching function AVX level attribute.
```

clang/test/CodeGenCXX/target-avx-abi-diag.cpp
```diff
@@ -0,0 +1,55 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -emit-llvm -disable-llvm-passes -verify=noavx,noavx512 -o - %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -target-feature +avx -emit-llvm -disable-llvm-passes -verify=noavx512 -o - %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -target-feature +avx512f -emit-llvm -disable-llvm-passes -verify=both -o - %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -std=c++17 -Wno-psabi -emit-llvm -disable-llvm-passes -verify=suppressed -o - %s
+// REQUIRES: x86-registered-target
+
+// both-no-diagnostics
+// suppressed-no-diagnostics
+
+typedef float v8f __attribute__((vector_size(32)));
```
AaronBallman, line 10: Can we also have test coverage for a K&R C function without a prototype?

clang/lib/CodeGen/Targets/X86.cpp
```diff
@@ -1571,6 +1575,65 @@ static bool checkAVXParam(DiagnosticsEngine &Diag, ASTContext &Ctx,
   return false;
 }
 
+void X86_64TargetCodeGenInfo::checkFunctionABI(CodeGenModule &CGM,
```
AaronBallman, line 1578: Do we need special handling for calls through function pointers where we might not have a `FunctionDecl`? What about things like Objective-C method declarations? Blocks?

AaronBallman (commented):
We should also add a release note to `clang/docs/ReleaseNotes.rst` so users know about the changes.

clang/test/CodeGenCXX/target-avx-abi-diag.cpp
freaknbigpanda: Added a test in CodeGen/target-avx-abi-diag-knr.c to cover this!

clang/lib/CodeGen/Targets/X86.cpp
freaknbigpanda: So the original implementation that only warned at callsites doesn't work if there is no FD. For example the following code doesn't warn:

```
typedef short avx256Type __attribute__((vector_size(32)));
typedef void (*fp_arg_t)(avx256Type);

__attribute__((used)) static void call_fp_arg(fp_arg_t fp) {
  avx256Type x = (avx256Type){0};
  fp(x); // Missing -Wpsabi warning.
}
```

GCC does warn however so I think we can expand this to work in cases where there is no FD but I think that is a followup. This PR is extending the current FD based functionality to definitions, not making the diagnostics work with no FD. Happy to create an issue or add a fixme if you would like.

freaknbigpanda:
Note that the current failures on my PR are due to libc++ issues on main, I'll rebase and make sure all tests are green before merging

freaknbigpanda:
@AaronBallman Can you help to finish this review?

clang/lib/CodeGen/Targets/X86.cpp
RKSimon: +1 for a new tracking issue

RKSimon (commented):
Please can you confirm the CI failures are irrelevant (do they go when you merge against trunk latest)?

AaronBallman (approved):
LGTM!

clang/lib/CodeGen/Targets/X86.cpp
freaknbigpanda: Created issue here: https://github.com/llvm/llvm-project/issues/208087

PR #199096, opened 2026-05-21
[AIX] Change default for llvm-nm and llvm-ar to Any

madanial0:
change default for llvm-nm and llvm-ar to be Any on aix, this is required to remove the dependancy on OBJECT_MODE when building llvm on AIX

### DIFF
--- a/llvm/tools/llvm-ar/llvm-ar.cpp
+++ b/llvm/tools/llvm-ar/llvm-ar.cpp
@@ -1290,7 +1290,7 @@ static BitModeTy getBitMode(const char *RawBitMode) {
       .Case("64", BitModeTy::Bit64)
       .Case("32_64", BitModeTy::Bit32_64)
       .Case("any", BitModeTy::Any)
-      .Default(BitModeTy::Unknown);
+      .Default(BitModeTy::Any);
 }
 
 static const char *matchFlagWithArg(StringRef Expected,
--- a/llvm/tools/llvm-ar/llvm-ar.cpp
+++ b/llvm/tools/llvm-ar/llvm-ar.cpp
@@ -1313,7 +1313,7 @@ static BitModeTy getBitMode(const char *RawBitMode) {
       .Case("64", BitModeTy::Bit64)
       .Case("32_64", BitModeTy::Bit32_64)
       .Case("any", BitModeTy::Any)
-      .Default(BitModeTy::Any);
+      .Default(BitModeTy::Unknown);
 }
 
 static const char *matchFlagWithArg(StringRef Expected,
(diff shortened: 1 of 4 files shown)


llvm/tools/llvm-nm/llvm-nm.cpp
```diff
@@ -2487,7 +2487,7 @@ int llvm_nm_main(int argc, char **argv, const llvm::ToolContext &) {
                   .Case("64", BitModeTy::Bit64)
                   .Case("32_64", BitModeTy::Bit32_64)
                   .Case("any", BitModeTy::Any)
-                  .Default(BitModeTy::Bit32);
+                  .Default(BitModeTy::Any);
```
DanielCChen, line 2490: May be I missed the broader conversation, but I thought we didn't want to change the default bitmode for `llvm-nm`?
Or `Any` is a NFC change?

llvm/tools/llvm-ar/llvm-ar.cpp
```diff
@@ -1313,7 +1313,7 @@ static BitModeTy getBitMode(const char *RawBitMode) {
       .Case("64", BitModeTy::Bit64)
       .Case("32_64", BitModeTy::Bit32_64)
       .Case("any", BitModeTy::Any)
-      .Default(BitModeTy::Unknown);
+      .Default(BitModeTy::Any);
```
kkwli: Does this change make `llvm-ar -X17 ...` to default to `any` instead of error?

kkwli:
The description does not match the changes. Please update it.

llvm/tools/llvm-ar/llvm-ar.cpp
madanial0: You are correct but I feel like that should be addressed at the option handling level

llvm/tools/llvm-nm/llvm-nm.cpp
madanial0: Any would work for both 32 and 64 bit, that way we don't need to rely on OBJECT_MODE

DanielCChen: I guess what I want to clarify is if `OBJECT_MODE` is not set, if users call `llvm-nm`, which version did we get?

madanial0: the same one as if `OBJECT_MODE` is set to any

llvm/tools/llvm-ar/llvm-ar.cpp
madanial0: fix posted. Thanks for pointing it out

jh7370 (changes requested):
This PR needs to update the testing, docs (I checked llvm-nm's command guide and it talked about the default being 32-bit, for example) and potentially the help text (I haven't checked this).

Also, it needs buy-in from the original stakeholders (e.g. @diggerlin or somebody else involved in the initial bring-up) or an RFC for better visibility, I feel, as the default was specifically chosen to match the IBM behaviour. I see @daltenty was involved in the original review for llvm-ar, so if they're already associated with @diggerlin and their organization, this would be fine with me, as an example.

See also the comments in https://github.com/llvm/llvm-project/pull/204281.

PR #195126, opened 2026-04-30
[clang] Complete fields in __builtin_offsetof designators

schuay:
Code completion was a no-op inside `__builtin_offsetof`: a cursor at ` __builtin_offsetof(T, ^)` or `__builtin_offsetof(T, a.^)` fell through to ordinary-name completion instead of suggesting fields. Route the code_completion token to a new SemaCodeCompletion entry point that walks the designator path so far, resolves the subobject's type, and enumerates its members. Methods are filtered out, inherited fields are included, indirect fields from anonymous unions and structs are peeled, and `using Base::field` resolves through its UsingShadowDecl. A code_completion token past a complete component (right after `]` or at the end of the chain) is dropped rather than offering fields the user can't paste without first typing `.`.

The offsetof and designated-initializer type walkers are folded into one helper parameterized by a field-lookup callback, which incidentally fixes reference-field and indirect-field traversal in designated-initializer completion too.

Tests: lit cases in offsetof.cpp covering empty/dot/array/inheritance/ reference/anonymous/using-shadow/macro forms, extended desig-init.cpp walker cases, and a clangd unit test exercising the IDE path.

Context: #195126 and #194407

This work was AI assisted but human-reviewed.

The followup for the added FIXME is at https://github.com/llvm/llvm-project/compare/main...schuay:llvm-project:refactor-offsetof-component-to-designation

PR #194407, opened 2026-04-27
[clangd] Resolve __builtin_offsetof designator components precisely

schuay:
Building on the new TraverseOffsetOfNode hook in RecursiveASTVisitor and
the OffsetOfNode DynTypedNode kind, teach SelectionTree, FindTarget, and
the explicit-references collector to address each designator component
individually. Cursor positions inside a nested designator (for example
the 'B' in __builtin_offsetof(A, B.c)) now resolve to the corresponding
field instead of always picking the innermost component.

- SelectionTree: wrap each OffsetOfNode visit in traverseNode so it
  becomes a selectable node alongside its enclosing OffsetOfExpr.
- FindTarget::allTargetDecls: resolve OffsetOfNode (Field kind) to its
  FieldDecl, and drop the OffsetOfExpr fallback so non-component
  selections do not guess a field target.
- ExplicitReferenceCollector: emit one ReferenceLoc per component via a
  new VisitOffsetOfNode hook, replacing the manual component loop in
  refInStmt.

Tests:
- TargetDeclTest.OffsetOf: precise resolution for outer/inner/base/
  identifier components, macro form, array index expressions, and
  non-component selections.
- Hover.All and Hover.OffsetOfBuiltin: field cases for nested offsetof,
  and expression hover on the builtin token.
- LocateSymbol.All: simple, outer, inner, macro, inherited-field, and
  builtin-token cases.
- AllRefsInFoo: existing offsetof cases continue to pass via the new
  per-component VisitOffsetOfNode path.

Context: 
https://github.com/llvm/llvm-project/pull/194122
https://github.com/llvm/llvm-project/pull/192953

PR #199097, opened 2026-05-21
[clang] Replace Sema::OffsetOfComponent with Designation

schuay:
The parser previously fed __builtin_offsetof through a small ad-hoc struct (Sema::OffsetOfComponent) that mirrors the same path data already modeled by Designation/Designator. With the recent code-completion work, the parser was building both shapes side by side.

Drop OffsetOfComponent. ActOnBuiltinOffsetOf, BuildBuiltinOffsetOf, and TreeTransform::RebuildOffsetOfExpr now take a const Designation &.

- Parser: stops maintaining the parallel Comps vector and feeds the Designation it already collected for completion straight to Sema.
- BuildBuiltinOffsetOf: iterates over the Designation.
- TreeTransform: rebuilds a Designation from each OffsetOfNode. The AST node already encodes "leading-dot collapsed to name location" via its Range constructor, so we mirror that by passing an empty DotLoc whenever the new Designation is still empty.
- OffsetOfNode: tighten the identifier constructor and getFieldName accessor to const IdentifierInfo*, matching Designator's accessor and letting BuildBuiltinOffsetOf drop the lingering const_cast.

No functional change intended; existing offsetof tests cover dependent types, dependent bases, anonymous unions, indirect fields, base-class indirections, and the non-POD diagnostic source range.

Context: https://github.com/llvm/llvm-project/pull/195126, https://github.com/llvm/llvm-project/pull/194407

This work was AI assisted, human-reviewed.

### DIFF
--- a/clang/lib/Sema/SemaExpr.cpp
+++ b/clang/lib/Sema/SemaExpr.cpp
@@ -16614,7 +16614,7 @@ ExprResult Sema::ActOnStmtExprResult(ExprResult ER) {
 
 ExprResult Sema::BuildBuiltinOffsetOf(SourceLocation BuiltinLoc,
                                       TypeSourceInfo *TInfo,
-                                      ArrayRef<OffsetOfComponent> Components,
+                                      const Designation &Desig,
                                       SourceLocation RParenLoc) {
   QualType ArgTy = TInfo->getType();
   bool Dependent = ArgTy->isDependentType();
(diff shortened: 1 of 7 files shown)


schuay:
cc @zwuis @ArcsinX , this is the cleanup pass on top of https://github.com/llvm/llvm-project/pull/195126.

schuay:
@zwuis friedly ping, could you ptal? Cleanup pass from the previous OffsetOf change that settles on the more common Designation/Designator.

schuay:
cc @zwuis @ArcsinX friendly ping, I am still looking for a reviewer.

ArcsinX (approved):
LGTM

clang/lib/Sema/SemaExpr.cpp
```diff
@@ -16638,19 +16638,30 @@ ExprResult Sema::BuildBuiltinOffsetOf(SourceLocation BuiltinLoc,
...
+  // Source-range helpers for a designator. The leading field has no '.', so
+  // its range collapses to the field-name location.
+  auto LocStart = [](const Designator &D) {
+    if (D.isFieldDesignator())
+      return D.getDotLoc().isValid() ? D.getDotLoc() : D.getFieldLoc();
+    return D.getLBracketLoc();
+  };
+  auto LocEnd = [](const Designator &D) {
+    return D.isFieldDesignator() ? D.getFieldLoc() : D.getRBracketLoc();
+  };
```
zwuis: It's up to you, but we can add member functions `Designator::getBeginLoc()` and `Designator::getEndLoc()` instead.

clang/lib/Sema/SemaExpr.cpp
```diff
@@ -16638,19 +16638,30 @@ ExprResult Sema::BuildBuiltinOffsetOf(SourceLocation BuiltinLoc,
...
+  for (unsigned I = 0, N = Desig.getNumDesignators(); I != N; ++I) {
+    const Designator &D = Desig.getDesignator(I);
```
zwuis, line 16642: `assert(!D.isArrayRangeDesignator());`

zwuis (approved):
> Context: https://github.com/llvm/llvm-project/pull/195126, https://github.com/llvm/llvm-project/pull/194407, https://github.com/llvm/llvm-project/pull/195126

Two 195126?

schuay:
> > Context: #195126, #194407, #195126
> 
> Two 195126?

Thank you folks. Ah yes, this snuck in in the previous commit already which appears to be self-referential :) Will fix in the updated commit message.

clang/lib/Sema/SemaExpr.cpp
schuay: Done

clang/lib/Sema/SemaExpr.cpp
schuay: Done

schuay:
All done. Please help me commit this. Note this was the last planned commit of the chain, ty again.

PR #199102, opened 2026-05-21
[mlir][spirv] Fix `fp8` and `bf16` leaking into unsupported ops

IgWod:
Including `SPIRV_AnyFloat` in the majority of types caused fp8 and bf16 to be allowed in ops that are not allowed by float8 and bfloat16 extensions. This patch tries to rectify to only allow fp8 and bf16 in ops allowed by the respective specs. Additional tests have been also added to increase the coverage with respect to those types.

Assisted-by: Codex + Claude Code

### DIFF
--- a/mlir/include/mlir/Dialect/SPIRV/IR/SPIRVArithmeticOps.td
+++ b/mlir/include/mlir/Dialect/SPIRV/IR/SPIRVArithmeticOps.td
@@ -464,12 +464,12 @@ def SPIRV_DotOp : SPIRV_Op<"Dot",
   }];
 
   let arguments = (ins
-    SPIRV_VectorOf<SPIRV_AnyFloat>:$vector1,
-    SPIRV_VectorOf<SPIRV_AnyFloat>:$vector2
+    AnyTypeOf<[SPIRV_FloatVector, SPIRV_VectorOf<SPIRV_BFloat16KHR>]>:$vector1,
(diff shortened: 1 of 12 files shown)


kuhar (approved):
thanks for catching this!

aobolensk (approved):

davidegrohmann (approved):

d-smirnov:
@IgWod @kuhar 
It seems the change is too restrictive. Following is not allowed now:

    spirv.func @group_non_uniform_ballot(%arg0: vector<2xbf16>, %arg1: bf16, %arg2: i32) "None" {
      %0 = spirv.VectorInsertDynamic %arg1, %arg0[%arg2] : vector<2xbf16>, i32
      spirv.Return
    }

However, it seems that for composite ops it should not be so strict (providing the capabilities enabled). Could you advise how to handle the example above, please?

IgWod:
It looks like `VectorInsertDynamic` slipped through cracks when this change went in. Indeed it's a composite op and should allow bf16 as per the extension spec. It looks like this op should use `SPIRV_ExtendedFloatVector` instead of `SPIRV_Vector`. There may be some other composite ops that also need updating.

@d-smirnov Would you be happy to submit this change? The change is fairly trivial, the only extra bit is that it needs tests adding (for both bf16 and fp8).

d-smirnov:
Sure. Thanks for clarifications

PR #149852, opened 2025-07-21
[SelectionDAG] Pass SDNodeFlags through getNode instead of setFlags.

topperc:
getNode updates flags correctly for CSE. Calling setFlags after getNode may set the flags where they don't apply.

I've added flags to getSelectCC and the signature of getNode that takes an ArrayRef of EVTs.

PR #199105, opened 2026-05-21
[SelectionDAG] this fixes fallout from #153238

Varnike:
Fix a regression after 4d010078.

This happens because `DAG.getSetCC()` now takes `SDNodeFlags` explicitly, so this combine no longer inherits fast-math flags from `SelectionDAG::FlagInserter`.

Reported in: https://github.com/llvm/llvm-project/pull/153238#issuecomment-4511030457


MacDue (commented):
Can you add the context of https://github.com/llvm/llvm-project/pull/149852 to the description?

Varnike:
I think the failure is due to the #153238 being merged after #169904 was merged upstream, even though it was written and approved before that change.

#169904 changed `getSetCC()` to pass `SDNodeFlags` explicitly, so the fast-math flags propagated by my original patch were no longer preserved in this path.

efriedma-quic (changes requested):
I reverted the underlying patch.  Please open a new PR to reapply with this fix.

PR #199106, opened 2026-05-21
[Instrumentor][FIX] Ensure we indicate changes properly.

jdoerfert:
We now indicate changes whenever we might have changed things even if the filter will rule out instrumentation. The issue was that we need to get the argument to check the filter and that process might create new IR, e.g., a global variable containing the function name.


kevinsala (approved):
LGTM

kevinsala:
Failing PowerPC tests are unrelated to this PR. Merging.

PR #199108, opened 2026-05-21
[Instrumentor] Improve the config wizard script

jdoerfert:
This makes the config wizard script more generic as we grow instrumentation opportunities. Better output, e.g., clear paths, are also displayed now.

Prepared with Claude (AI) and tested by me afterwards.


kevinsala (approved):
In general LGTM

PR #199110, opened 2026-05-21
[libc] Add a "Q" length modifier to support __float128 in printf

strel-12:
The spelling follows an existing pattern in e.g. libquadmath for format printing `__float128` values.

A flag is added to disable the modifier. This flag only affects whether the length modifier is parsed, because the same functionality will be used independently to implement `strfromf128`.


libc/src/__support/CPP/type_traits/disjunction.h
```diff
@@ -0,0 +1,25 @@
+#ifndef THIRD_PARTY_LLVM_LLVM_PROJECT_LIBC_SRC___SUPPORT_CPP_TYPE_TRAITS_DISJUNCTION_H_
```
michaelrj-google, line 1: nit: needs a license header

libc/src/stdio/printf_core/parser.h
```diff
@@ -254,12 +265,22 @@ template <typename ArgProvider> class Parser {
...
-        if (lm != LengthModifier::L) {
-          WRITE_ARG_VAL_SIMPLEST(section.conv_val_raw, double, conv_index);
-        } else {
+#if defined(LIBC_TYPES_HAS_FLOAT128)
+        if (lm == LengthModifier::Q) {
+#if !defined(LIBC_COPT_PRINTF_DISABLE_Q_LENGTH_MODIFIER)
+          WRITE_ARG_VAL_SIMPLEST(section.conv_val_raw, float128, conv_index);
+#else
+          section.has_conv = false;
+#endif // LIBC_COPT_PRINTF_DISABLE_Q_LENGTH_MODIFIER
```
michaelrj-google, line 274: it might be useful to simplify this by defining a single constant in `printf_config.h` that represents both "we have float128" and "Q is not disabled", similar to what we do with fixed point (see `LIBC_INTERNAL_PRINTF_HAS_FIXED_POINT`).

libc/test/src/stdio/sprintf_test.cpp
```diff
@@ -1521,9 +1635,26 @@ TEST(LlvmLibcSPrintfTest, FloatDecimalConv) {
...
+};
+
+#if defined(LIBC_SPRINTF_TEST_Q_MODIFIER)
+using LongDoubleFloat128TypesAndModifiers =
+    LIBC_NAMESPACE::testing::TypeList<TypeAndLengthModifier<long double, 'L'>,
+                                      TypeAndLengthModifier<float128, 'Q'>>;
+#else
+using LongDoubleFloat128TypesAndModifiers =
+    LIBC_NAMESPACE::testing::TypeList<TypeAndLengthModifier<long double, 'L'>>;
+#endif // LIBC_SPRINTF_TEST_Q_MODIFIER
```
michaelrj-google, line 1650: I don't think it's necessary to replace the `L` with a `Q`. It would be simpler to have separate tests for the `Q` cases.

PR #199116, opened 2026-05-21
[lldb] Fix vtable support on arm64e

bulbazord:
There were 2 small issues.
1. ValueObjectVTableChild was not fixing the addresses it was pulling from signed pointers. This broke things like `SBValue::GetLoadAddress` and identifying the function pointer type from debug info.
2. TestVTableValue.py made a lot of assumptions that did not hold on arm64e. a. GetValueAsUnsigned will return a raw pointer value. Most of the time, we needed GetValueAsAddress. b. The test was reading pointers out of memory without fixing them up. c. The summary for a function pointer on arm64e includes the load address. This isn't true on other platforms.

### DIFF
--- a/lldb/source/ValueObject/ValueObjectVTable.cpp
+++ b/lldb/source/ValueObject/ValueObjectVTable.cpp
@@ -73,6 +73,8 @@ protected:
       return false;
     }
 
+    parent_addr = process_sp->FixCodeAddress(parent_addr);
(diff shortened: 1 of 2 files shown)


jimingham (approved):
Looks right to me.

jasonmolenda (approved):
LGTM

PR #199117, opened 2026-05-21
[lldb] Remove ValueObjectRecognizerSynthesizedValue::IsSynthetic override

kastiglione:
Removes the `IsSynthetic` override on `ValueObjectRecognizerSynthesizedValue`. This class does not also override `GetNonSyntheticValue`.

There was a bug in which code assumed that when `IsSynthetic()` returned true, that `GetNonSyntheticValue` would produce a different value object. However the default behavior of `GetNonSyntheticValue` is to return itself.

It seems to me that either:
1. `ValueObjectSynthetic` should be the only class to override `IsSynthetic` to true
2. or, that classes which override `IsSynthetic` should also override `GetNonSyntheticValue`

In either case, I think it's best to remove this `IsSynthetic` on `ValueObjectRecognizerSynthesizedValue`.

### DIFF
--- a/lldb/include/lldb/Target/StackFrameRecognizer.h
+++ b/lldb/include/lldb/Target/StackFrameRecognizer.h
@@ -196,7 +196,6 @@ class ValueObjectRecognizerSynthesizedValue : public ValueObject {
   CompilerType GetCompilerTypeImpl() override {
     return m_parent->GetCompilerType();
   }
-  bool IsSynthetic() override { return true; }
 
  private:
   lldb::ValueType m_type;


kastiglione:
See also https://github.com/swiftlang/llvm-project/pull/13038

kastiglione:
@jimingham thoughts?

jimingham:
Even though it has Synthesize in the name, ValueObjectRecognizerSynthesizedValue is not a synthetic value object.  A Synthetic ValueObject is a ValueObject that stands in for the Value or Children of some other ValueObject.

But ValueObjectRecognizerSynthesizedValue is handed out as the first order result of providing values in a synthesized frame.  There is no ValueObject for it to be the synthetic representation of.  It would be like saying that a ValueObjectVariable is synthetic.  It doesn't make sense.

I have to think a little bit more about the general "IsSynthetic" question, however.  It seems to me like it matters more whether something has a synthetic child provider than the kind of a particular child returned by a synthetic child provider.  

For instance, if I have a struct and my synthetic child provider presents some of the original ValueObject's ValueObjectChildren as its synthetic children, and then adds some others.  Is it right to say that the original ValueObjectChildren are synthetic?

jimingham (approved):
I am less sure about the overall question, but there's no sense in which a ValueObjectRecognizerSynthesizedValue is a "Synthetic" value as lldb has been using the term synthetic.  It is on a level with ValueObjectVariable or other providers of root value objects.

kastiglione:
Thanks for sharing your thought process.

PR #196029, opened 2026-05-06
[AArch64][SVE] Use truncating stores whenever possible

sushgokh:
For fixed length SVE and fixed length vectors x/y, fold

```
store(concat_vector(truncate(x), truncate(y)))
-->  store(truncate(x))
     store(truncate(y))
```

PR #199122, opened 2026-05-21
[AArch64] Address issue reported in PR#196029

sushgokh:
For certain types of truncating stores, the lowering action is set to custom although no custom lowering exists for them.

Add  a guard to prevent such false entry.

Addresses issue reported in PR #196029

### DIFF
--- a/llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
+++ b/llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
@@ -1774,19 +1774,15 @@ AArch64TargetLowering::AArch64TargetLowering(const TargetMachine &TM,
     }
 
     // SVE supports truncating stores of 64 and 128-bit vectors
-    auto UseSVE = [&](MVT VT) {
-      return useSVEForFixedLengthVectorVT(
-          VT, /*OverrideNEON=*/Subtarget->useSVEForFixedLengthVectors());
-    };
-    if (UseSVE(MVT::v2i64)) {
-      setTruncStoreAction(MVT::v2i64, MVT::v2i8, Custom);
-      setTruncStoreAction(MVT::v2i64, MVT::v2i16, Custom);
-      setTruncStoreAction(MVT::v2i64, MVT::v2i32, Custom);
-    }
-    if (UseSVE(MVT::v2i32)) {
-      setTruncStoreAction(MVT::v2i32, MVT::v2i8, Custom);
-      setTruncStoreAction(MVT::v2i32, MVT::v2i16, Custom);
-    }
+    auto VT = MVT::v2i64;
+    if (useSVEForFixedLengthVectorVT(
+            VT, /*OverrideNEON=*/Subtarget->useSVEForFixedLengthVectors()))
+      setTruncStoreAction(VT, MVT::v2i32, Custom);
+
+    setTruncStoreAction(MVT::v2i64, MVT::v2i8, Custom);
+    setTruncStoreAction(MVT::v2i64, MVT::v2i16, Custom);
+    setTruncStoreAction(MVT::v2i32, MVT::v2i8, Custom);
+    setTruncStoreAction(MVT::v2i32, MVT::v2i16, Custom);
 
     for (auto VT : {MVT::nxv2f16, MVT::nxv4f16, MVT::nxv8f16, MVT::nxv2f32,
                     MVT::nxv4f32, MVT::nxv2f64}) {
--- a/llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
+++ b/llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
@@ -1774,11 +1774,19 @@ AArch64TargetLowering::AArch64TargetLowering(const TargetMachine &TM,
     }
 
     // SVE supports truncating stores of 64 and 128-bit vectors
-    setTruncStoreAction(MVT::v2i64, MVT::v2i8, Custom);
-    setTruncStoreAction(MVT::v2i64, MVT::v2i16, Custom);
-    setTruncStoreAction(MVT::v2i64, MVT::v2i32, Custom);
-    setTruncStoreAction(MVT::v2i32, MVT::v2i8, Custom);
-    setTruncStoreAction(MVT::v2i32, MVT::v2i16, Custom);
+    auto UseSVE = [&](MVT VT) {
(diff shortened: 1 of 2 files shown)


llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
```diff
@@ -1762,9 +1762,13 @@ AArch64TargetLowering::AArch64TargetLowering(const TargetMachine &TM,
     }
 
     // SVE supports truncating stores of 64 and 128-bit vectors
+    auto VT = MVT::v2i64;
+    if (useSVEForFixedLengthVectorVT(
+            VT, /*OverrideNEON=*/Subtarget->useSVEForFixedLengthVectors()))
+      setTruncStoreAction(VT, MVT::v2i32, Custom);
```
paulwalker-arm: This looks unnecessary? If the preference is to use NEON for v2i64->v2i32 then I would expect that to be a universal choice regardless of the value of vscale_range.

llvm/test/CodeGen/AArch64/sve-fixed-length-trunc-stores.ll
```diff
@@ -202,4 +202,31 @@ define void @store_trunc_v32i16i8(ptr %ap, ptr %dest) #0 {
...
+; CHECK:       // %bb.0:
+; CHECK-NEXT:    ldp q1, q0, [x0]
+; CHECK-NEXT:    uzp1 v0.4s, v1.4s, v0.4s
+; CHECK-NEXT:    str q0, [x1]
+; CHECK-NEXT:    ret
+  %ld = load <4 x i64>, ptr %src, align 4
+  %trunc = trunc <4 x i64> %ld to <4 x i32>
+  store <4 x i32> %trunc, ptr %dest, align 8
+  ret void
+}
```
paulwalker-arm: This test is identical to the one below because `+sve` implies `+neon`. I think you want these tests but with one using `vscale_range(1,0)` and the other `vscale_range(2,0)` to show how the generated code varies based on the minimum vector length. Ideally the new function would sit just after `store_trunc_v8i64i8` to maintain the current function ordering.

paulwalker-arm (commented):
I suspect we might need to better expose the multi-register load/store abilities of AArch64 to DAGCombiner but for today this change is small and well contained so works for me.

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
sushgokh: Its not about preference for NEON. 

In the isel logs for 'store_trunc_v4i64i32_neon_sve' test, we would see 
```
t1 = trunc ...
t2 = trunc ...
concat = concat_vectors t1,t2
```

PR #196029 would treat this as valid optimisation and would split and prefer individual truncating stores. See [this](https://godbolt.org/z/4qKbP4rq1)
Above check says that "if custom lowering actually exists, only then treat optimisation in PR #196029 as valid".

llvm/test/CodeGen/AArch64/sve-fixed-length-trunc-stores.ll
sushgokh: my bad. Mistook `+sve` to be `+sve,-neon`. Intended the latter part in `store_trunc_v4i64i32_sve_only` test. Will update the 1st and  2nd test, both.

Just curious:  why `+sve` also implies `+neon` if they appear as individual features ?

llvm/test/CodeGen/AArch64/sve-fixed-length-trunc-stores.ll
```diff
@@ -202,4 +202,32 @@ define void @store_trunc_v32i16i8(ptr %ap, ptr %dest) #0 {
...
+
+
+define void @store_trunc_v4i64i32_sve_no_neon(ptr %src, ptr %dest) vscale_range(1,0) "target-features"="-neon,+sve" {
```
sushgokh: @paulwalker-arm This test is crashing with assertions enabled but the crash is unrelated to this change. Can I remove this case from here and address the crash in a different patch?

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
paulwalker-arm: I'm not really following the logic.  The reason for guarding the custom lowering is because when SVE is not available the code generation is improved without it.  My point is that for 64-bit vectors I would not expect SVE to be able to improve on the code quality anyway and thus there is no reason to ever custom lower it.

To further this point. If I remove the above code I see no test changes?

llvm/test/CodeGen/AArch64/sve-fixed-length-trunc-stores.ll
paulwalker-arm: I see this as just a bogus test because SVE without NEON is an invalid implementation.  This is why I suggested varying the test by `vscale_range(1,0)` and `vscale_range(2,0)` because I believe this best shows the intent. Specially, for `vscale_range(1,0)` we use `ldp` but for `vscale_range(2,0)` we use SVE because the registers are big enough to only require a single `ld1`.

llvm/test/CodeGen/AArch64/sve-fixed-length-trunc-stores.ll
paulwalker-arm: It's a weird relationship whereby SVE is not really an extension of NEON but it's not fully independent of it either.  The closes you get to SVE being independent is Streaming-SVE, but even then there are a handful of NEON instructions that remain available.  At the C++ level this is why you'll see `Subtarget->isNeonAvailable()` used more often than `Subtarget->hasNEON()` to disambiguate the available NEON instructions from the feature itself.

llvm/test/CodeGen/AArch64/sve-fixed-length-trunc-stores.ll
sushgokh: ok thanks. Done

llvm/lib/Target/AArch64/AArch64ISelLowering.cpp
sushgokh: I was bit skeptical about removing the entire code. Removing the code makes no change to the test output. Thanks. Done.

paulwalker-arm (approved):

PR #199123, opened 2026-05-21
[CIR] Omit nsw/nuw on integer vector binops

adams381:
CIRGen was attaching `nsw` to `cir.add` on `!cir.vector` integer types
because the signed-overflow path keys off `compType` (the element type,
still `!s32i`), while the verifier only allows `nsw`/`nuw` on scalar
`!cir.int` results.  That mismatch showed up 144 times in the libcxx
CIR sweep on `std::experimental::simd` — `experimental/__simd/vec_ext.h`
increment/decrement (`__data + 1` / `__data - 1`).

Classic CodeGen never enters the signed-overflow block for vector
computation types; CIR now skips the scalar `nsw`/`nuw` path when the
MLIR operand is an integer vector, and uses `getAs<VectorType>()` for
`compType` so typedef-wrapped GCC vectors resolve the element type
correctly.

Regression test `vector-binop-overflow.cpp` checks scalar add still gets
`nsw` and vector add/sub/mul do not (CIR + LLVM + OGCG).

### DIFF
--- a/clang/test/CIR/CodeGen/vector-binop-overflow.cpp
+++ b/clang/test/CIR/CodeGen/vector-binop-overflow.cpp
@@ -1,0 +1,41 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -Wno-unused-value -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck --check-prefix=CIR --input-file=%t.cir %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -Wno-unused-value -fclangir -emit-llvm %s -o %t-cir.ll
+// RUN: FileCheck --check-prefix=LLVM --input-file=%t-cir.ll %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -Wno-unused-value -emit-llvm %s -o %t.ll
+// RUN: FileCheck --check-prefix=LLVM --input-file=%t.ll %s
+
+typedef int vi4 __attribute__((vector_size(16)));
+
+struct SimdStorage {
+  int __data __attribute__((vector_size(16)));
+};
+
+int scalar_add(int a, int b) { return a + b; }
+
+void vector_binops(vi4 a, vi4 b, SimdStorage &s) {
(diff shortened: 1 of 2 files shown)


clang/test/CIR/CodeGen/vector-binop-overflow.cpp
```diff
@@ -0,0 +1,49 @@
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -Wno-unused-value -fclangir -emit-cir %s -o %t.cir
+// RUN: FileCheck --check-prefix=CIR --input-file=%t.cir %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -Wno-unused-value -fclangir -emit-llvm %s -o %t-cir.ll
+// RUN: FileCheck --check-prefix=LLVM --input-file=%t-cir.ll %s
+// RUN: %clang_cc1 -triple x86_64-unknown-linux-gnu -Wno-unused-value -emit-llvm %s -o %t.ll
+// RUN: FileCheck --check-prefix=OGCG --input-file=%t.ll %s
```
erichkeane: Why is there separate check-lines for OGCG?

adams381: > Why is there separate check-lines for OGCG?

The `LLVM` lines check `-fclangir -emit-llvm` (`%t-cir.ll`); the `OGCG` lines check classic `-emit-llvm` (`%t.ll`). We duplicate the expectations so the test fails if CIR lowering and OGCG diverge — not because we expect different IR today.

erichkeane: No we don't?

adams381: You're right — fixed.  The last RUN line now uses `--check-prefix=LLVM` for both `%t-cir.ll` and `%t.ll`, and the duplicate `OGCG` check lines are gone.

adams381:
@erichkeane ready for re-review after the May 26 fix.

erichkeane (approved):

PR #199126, opened 2026-05-21
[clang] Use FileError in FileManager::getFileRef, getDirectoryRef

benlangmuir:
Most callers are unchanged, since they either ignore the specific error or have their own formatting of the error that includes both the path and the errorToErrorCode-unwrapped value. However, for clients that just forward the error it's helpful to ensure we do not lose track of the filename that the error is associated with, so use FileError.

Incidentally remove two uses of errorToErrorCode that were being used instead of consumeError; in both cases getOptionalFileRef was more appropriate.

### DIFF
--- a/clang/lib/Basic/FileManager.cpp
+++ b/clang/lib/Basic/FileManager.cpp
@@ -21,6 +21,7 @@
 #include "llvm/ADT/SmallString.h"
 #include "llvm/ADT/Statistic.h"
 #include "llvm/Config/llvm-config.h"
+#include "llvm/Support/Error.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/IOSandbox.h"
 #include "llvm/Support/MemoryBuffer.h"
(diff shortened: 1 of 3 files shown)


cachemeifyoucan (approved):

jansvoboda11 (approved):

MaskRay:
There is a small compile time regression https://llvm-compile-time-tracker.com/compare.php?from=d2e9300b72637f92aa0ebf4aa5b82f84e4023640&to=b081351c7e20a26a6e67b3e82c065f02e4cce8d8&stat=instructions:u

```
	59234M (+0.04%)	74056M (+0.04%)	85564M (+0.04%)	17471M (+0.09%)	66487M (+0.03%)	20224M (+0.08%)	51324M (+0.03%)	15478M (+0.12%)	21519515M (+0.01%)
```

benlangmuir:
@MaskRay Thanks for the heads up. I suspect specializing the `getOptional*Ref` functions to not construct Errors would win this back, but I'm going to revert until I have time to look at it carefully.

PR #199127, opened 2026-05-21
[CI] Unify handling of scripted PR workflows

ojhunt:
A few workflows already force the use of baseline scripting when testing PRs. This PR pulls that out into a separate action that we can just reuse in other workflows as well.

### DIFF
--- a/.github/actions/checkout-baseline-tooling/action.yml
+++ b/.github/actions/checkout-baseline-tooling/action.yml
@@ -6,7 +6,7 @@ description: >-
 runs:
   using: composite
   steps:
-    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+    - uses: llvm/llvm-project/.github/actions/checkout-baseline-tooling@main
       with:
         persist-credentials: false
         sparse-checkout: |
--- a/.github/actions/checkout-baseline-tooling/action.yml
+++ b/.github/actions/checkout-baseline-tooling/action.yml
@@ -1,0 +1,17 @@
+name: Checkout Baseline Tooling
+description: >-
+  Checkout the baseline tooling scripts prior to running any pre-commit hooks
+  or other GH action or workflow scripts.
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
+        sparse-checkout: |
+          llvm/utils/git/
+          .ci/
+          .github/
+        ref: main
+        path: baseline-scripts
--- a/.github/workflows/pr-code-format.yml
+++ b/.github/workflows/pr-code-format.yml
@@ -26,6 +26,9 @@ jobs:
           persist-credentials: false
           fetch-depth: 2
 
+      - name: Checkout Baseline Tooling
+        uses: llvm/llvm-project/.github/actions/checkout-baseline-tooling@main
+
       - name: Get changed files
         id: changed-files
         uses: tj-actions/changed-files@9426d40962ed5378910ee2e21d5f8c6fcbf2dd96 # v47.0.6
@@ -49,10 +52,10 @@ jobs:
         # Create an empty comments file so the pr-write job doesn't fail.
         run: |
           echo "[]" > comments &&
-          python ./llvm/utils/git/code-format-helper.py \
+          python ./baseline-scripts/llvm/utils/git/code-format-helper.py \
             --write-comment-to-file \
             --token ${{ secrets.GITHUB_TOKEN }} \
-            --issue-number $GITHUB_PR_NUMBER \
+            --issue-number "$GITHUB_PR_NUMBER" \
             --start-rev HEAD~1 \
             --end-rev HEAD \
             --changed-files "$CHANGED_FILES"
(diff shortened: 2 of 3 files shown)


ojhunt:
That commit actually does the "right" thing from the PoV of what I was trying to do, but of course it (in principle) still breaks the CI because of course these scripts already clobber the directory with the new actions. @tstellar Behold my skill :D

ojhunt:
Awww, I thought my clever "llvm/main" would work, but having read more I think it's a broken and/or dumb idea :(

I guess the real path is land change then update hashes :-/

boomanaiden154 (commented):
Can you link actions that are already doing this? I'm skeptical a new composite action actually makes things simpler.

I'm also not sure how much this actually helps. Premerge is going to have way more problems than code formatting/linting since it just runs the build, and has the same exposure as formatting/linting.

.github/actions/checkout-baseline-tooling/action.yml
```diff
@@ -0,0 +1,17 @@
+name: Checkout Baseline Tooling
+description: >-
+  Checkout the baseline tooling scripts prior to running any pre-commit hooks
+  or other GH action or workflow scripts.
+
+runs:
+  using: composite
+  steps:
+    - uses: llvm/llvm-project/.github/actions/checkout-baseline-tooling@main
```
ojhunt, line 9: Also - I think? even if this could work I need a repository parameter as well?

.github/actions/checkout-baseline-tooling/action.yml
```diff
@@ -0,0 +1,17 @@
...
+      with:
+        persist-credentials: false
+        sparse-checkout: |
+          llvm/utils/git/
+          .ci/
+          .github/
```
ojhunt, line 15: If this were meant to be a general action maybe it should cover all ci scripting directories? (e.g. libcxx, compiler-rt, etc?)

A consistent tool for all workflows seems like the sensible course of action

.github/actions/checkout-baseline-tooling/action.yml
boomanaiden154: This is just referencing itself currently which doesn't make a lot of sense?

.github/actions/checkout-baseline-tooling/action.yml
boomanaiden154: Most of the scripts are in `llvm/utils`. I'm not sure a single tool for all workflows makes a lot of sense as they will all want to check out different things. An additional checkout step I think is also about the same number of lines of code...

ojhunt: @boomanaiden154 my thinking is just having a baseline of "all the scripts should be the baseline versions" and a single action that we could have a trivial lint for seems like it would be a good idea? I also realize I did mean to include a pile of the libcxx testing utilities in here as well, and compiler-rt, and any others that just seemed used everywhere

.github/actions/checkout-baseline-tooling/action.yml
ojhunt: @boomanaiden154 I was trying to make it just pull from the main llvm repository on the assumption that would always be trusted and then avoid a specific SHA check :D

it looks like the actual correct way to do this would be land, and then enable with a pin, but I was so hopeful :D

ojhunt:
> Can you link actions that are already doing this? I'm skeptical a new composite action actually makes things simpler.
> 
> I'm also not sure how much this actually helps. Premerge is going to have way more problems than code formatting/linting since it just runs the build, and has the same exposure as formatting/linting.

do you mean the "Oliver read about composites and tried to write one?" or the hackery around getting the baseline? the latter was based on google and trying to get past the endless llm generated garbage that now fills the world.

I wasn't sure if it was correct to do the sparse checkout things like this, because they all seemed to occur in gated workflows.

```yaml
      - name: Checkout LLVM
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
        with:
          persist-credentials: false
          sparse-checkout: |
            .github/workflows/containers/github-action-ci-tooling/
            ...
```

The need for a separate checkout seemed plausible, but there's also a reason that I mass added CI folk to this PR :D

.github/actions/checkout-baseline-tooling/action.yml
boomanaiden154: If that's the intent, you need to SHA pin in the user workflows.

A recursive self reference I don't think makes sense in any case.

.github/actions/checkout-baseline-tooling/action.yml
boomanaiden154: Lumping a bunch of things makes the sparse checkout less effective. I don't think utilities from `compiler-rt` are used inside any of these sorts of workflows. All the libcxx utilities would be specific to libcxx workflows.

If we actually want to make this a policy, I think we're better off documenting it in https://llvm.org/docs/CIBestPractices.html and trying to enforce it during review. We would have to do that anyways to ensure that someone is using utilities from this action rather than elsewhere.

boomanaiden154:
> I wasn't sure if it was correct to do the sparse checkout things like this, because they all seemed to occur in gated workflows.

What do you mean by gated workflows?

ojhunt:
> > I wasn't sure if it was correct to do the sparse checkout things like this, because they all seemed to occur in gated workflows.
> 
> What do you mean by gated workflows?

Only automatically run by members of the project/people who can write to the repository

boomanaiden154:
Also, both of these actions are triggered by `pull_request` events. If someone wants to change what the workflow does, they can just modify the workflow file itself. I'm not sure doing checkouts of trusted tools actually make sense here given that.

PR #199129, opened 2026-05-21
[SLP] Propagate through FMulAdd instrinsics in BoUpSLP::getVectorElementSize()

bababuck:
We propagate through simple binary operations already, `fmuladd` is excluded since it happens to be an intrinsic.

Motivated by case exposed when removing vectorization from pre-LTO, see https://github.com/llvm/llvm-project/pull/195886#issuecomment-4486422243.


bababuck:
@lukel97 this is the first half of the fix for the `~12% regression on 538.imagick_r` with the pre-LTO vectorization disabling. Will push a cost fix that is also needed separately.

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -25835,6 +25835,12 @@ unsigned BoUpSLP::getVectorElementSize(Value *V) {
     if (Level > RecursionMaxDepth)
       continue;
 
+    // Can we propogate widths through a call instruction
+    // We can for simple intrinsics
+    auto IsCompatibleCall = [](const Instruction *I) -> bool {
+      const auto *CI = dyn_cast<CallInst>(I);
+      return CI && (CI->getIntrinsicID() == Intrinsic::fmuladd);
```
alexey-bataev: 1. Remove parens.
2. What about llvm.fma?

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -25835,6 +25835,13 @@ unsigned BoUpSLP::getVectorElementSize(Value *V) {
     if (Level > RecursionMaxDepth)
       continue;
 
+    // Can we propogate widths through a call instruction
+    // We can for simple intrinsics
+    auto IsCompatibleCall = [](const Instruction *I) -> bool {
+      const auto *CI = dyn_cast<CallInst>(I);
+      return CI && (CI->getIntrinsicID() == Intrinsic::fmuladd ||
+                    CI->getIntrinsicID() == Intrinsic::fma);
```
alexey-bataev: The test for fma?

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -25835,6 +25835,13 @@ unsigned BoUpSLP::getVectorElementSize(Value *V) {
     if (Level > RecursionMaxDepth)
       continue;
 
+    // Can we propogate widths through a call instruction
+    // We can for simple intrinsics
+    auto IsCompatibleCall = [](const Instruction *I) -> bool {
+      const auto *CI = dyn_cast<CallInst>(I);
```
alexey-bataev:
```suggestion
      const auto *CI = dyn_cast<IntrinsicInst>(I);
```

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
alexey-bataev: Can we extend it for some other intrinsics, like abs, constrined ops, etc.?

bababuck: Extended for abs and fabs, tests are in #199612. From a quick look it doesn't seem that we support vectorization of constrained ops.

alexey-bataev (approved):

llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp
```diff
@@ -26204,6 +26204,22 @@ unsigned BoUpSLP::getVectorElementSize(Value *V) {
     if (Level > RecursionMaxDepth)
       continue;
 
+    // Can we propogate widths through a call instruction
+    // We can for simple intrinsics
```
alexey-bataev: Nit: 
```suggestion
    // Can we propagate widths through a call instruction
    // We can for simple intrinsics
```

PR #199130, opened 2026-05-21
[CIR] Honor x86-64 high-eightbyte coerced return offset in CIRGen

adams381:
On x86_64 Linux SysV, a 16-byte struct whose low eightbyte is NoClass
and whose high eightbyte is Integer is returned in a single integer
register.  The value is only meaningful at offset 8 in the caller's
aggregate temporary, not at offset 0.  CIRGen treated every direct
return as offset 0, so `operator-(Index)` in Eigen's symbolic index
helpers stored the wrong eightbyte into `AddExpr` temporaries and
`is_same_symb` failed in `symbolic_index`.

The fix classifies that return shape in `CIRGenFunctionInfo`, emits
coerced `!s64i` on `cir.func`/`cir.call` where appropriate, stores call
results at the direct offset, and loads `__retval` from the same offset
before `cir.return`.  Scope is intentionally narrow (heuristic for the
NoClass/Integer high-eightbyte layout, not full `X86_64ABIInfo`).

Regression test `coerced-return-highpart.cpp` (CIR + LLVM + OGCG).
Eigen `symbolic_index` passes with `-fclangir` on the same pattern.

### DIFF
--- a/clang/include/clang/CIR/ABIArgInfo.h
+++ b/clang/include/clang/CIR/ABIArgInfo.h
@@ -53,10 +53,10 @@ private:
 public:
   ABIArgInfo(Kind k = Direct) : directAttr{0, 0}, theKind(k) {}
 
-  static ABIArgInfo getDirect(mlir::Type ty = nullptr) {
+  static ABIArgInfo getDirect(mlir::Type ty = nullptr, unsigned offset = 0) {
     ABIArgInfo info(Direct);
     info.setCoerceToType(ty);
-    assert(!cir::MissingFeatures::abiArgInfo());
+    info.directAttr.offset = offset;
     return info;
   }
 
@@ -82,13 +82,10 @@ public:
     return false;
   }
 
-  bool canHaveCoerceToType() const {
-    assert(!cir::MissingFeatures::abiArgInfo());
-    return isDirect();
-  }
+  bool canHaveCoerceToType() const { return isDirect(); }
 
   unsigned getDirectOffset() const {
-    assert(!cir::MissingFeatures::abiArgInfo());
+    assert(isDirect() && "ABIArgInfo offset is only valid for Direct");
     return directAttr.offset;
   }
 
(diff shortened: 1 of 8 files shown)


andykaylor (changes requested):
Is this just a temporary fix until the proper ABI handling is implemented? I don't think we should be doing this kind of handling in CIR codegen.

adams381:
Running a full Eigen sweep with this cherry-picked onto current upstream main shows 28 tests regressing from PASS to SIGSEGV that were passing without this PR. Bisect confirms this PR is the cause.

The root issue is that `classifyCIRReturnType` is too broad. It fires on **any** 16-byte struct whose second field at offset 8 is an 8-byte integer — not just the intended "low eightbyte is NoClass/padding" case.

Here is a minimal reproducer:

```cpp
struct DualLong { long a; long b; };
DualLong make(long a, long b) { return {a, b}; }
long sum(long a, long b) {
  DualLong r = make(a, b);
  return r.a + r.b;
}
```

OGCG emits `call { i64, i64 } @make`, extracts both fields, stores both. CIR emits `call i64 @make` (one register), stores at offset 8, and loads `r.a` from offset 0 which is uninitialized (zero from alloca). `sum(3, 4)` returns **4** instead of **7**.

The classifier also fires on `{double x; long n;}`, `{void *p; long n;}`, and any other 16-byte struct where field 1 happens to be an 8-byte integer. All have their field 0 silently dropped. For Eigen, matrix types like `{Index rows; Index cols;}` get misclassified — after a function returns such a struct, `rows` is always 0, and the matrix product loops are computed with wrong dimensions, producing out-of-bounds access and SIGSEGV.

This is exactly what andykaylor flagged — the classifier needs the full x86-64 two-eightbyte classification to know whether the low eightbyte is actually NoClass. A field-layout pattern match is not sufficient.

adams381:
The follow-up adds the missing guard: the classifier now requires field[0] to be a recursively-empty C++ class, which is the x86-64 SysV NoClass condition — not an approximation. That rules out `{long, long}`, `{double, long}`, and any struct with a scalar or array in the first eightbyte. If this moves to a proper ABI pass later, the guard migrates unchanged; the trigger condition is the spec criterion either way.

adams381:
Closing in favor of the calling convention lowering stack, which will provide the full x86-64 eightbyte classification and handle this case through the proper layering andykaylor described.

PR #199132, opened 2026-05-21
[offload] Use device memory for the multithreaded kernel lanuch test

kevinsala:
This commit modifies the multithreaded kernel launch test to use device memory instead of managed memory. The test is reported to be failing intermittently in systems where concurrent managed memory access is not supported. This is the case for NVIDIA devices that do not support CU_DEVICE_ATTRIBUTE_CONCURRENT_MANAGED_ACCESS.

The concept of concurrent and coherent managed memory access should be exposed to liboffload users somehow, e.g., adding it as device property, so it is clear what execution patterns are allowed with managed memory. However, this test is just testing concurrent kernel launches. This commit fixes it until we decide how to proceed with the guarantees on that type of allocations.

### DIFF
--- a/offload/unittests/OffloadAPI/kernel/olLaunchKernel.cpp
+++ b/offload/unittests/OffloadAPI/kernel/olLaunchKernel.cpp
@@ -79,24 +79,29 @@ TEST_P(olLaunchKernelFooTest, Success) {
 
 TEST_P(olLaunchKernelFooTest, SuccessThreaded) {
   threadify([&](size_t) {
-    void *Mem;
-    ASSERT_SUCCESS(olMemAlloc(Device, OL_ALLOC_TYPE_MANAGED,
-                              LaunchArgs.GroupSize.x * sizeof(uint32_t), &Mem));
+    void *DevAlloc, *HstAlloc;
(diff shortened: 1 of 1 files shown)


offload/unittests/OffloadAPI/kernel/olLaunchKernel.cpp
```diff
@@ -79,24 +79,29 @@ TEST_P(olLaunchKernelFooTest, Success) {
...
     ASSERT_SUCCESS(olLaunchKernel(Queue, Device, Kernel, &Args, sizeof(Args),
                                   &LaunchArgs, nullptr));
 
+    ASSERT_SUCCESS(olMemcpy(Queue, HstAlloc, Host, DevAlloc, Device, Size));
+
     ASSERT_SUCCESS(olSyncQueue(Queue));
 
-    uint32_t *Data = (uint32_t *)Mem;
-    for (uint32_t i = 0; i < 64; i++) {
+    uint32_t *Data = (uint32_t *)HstAlloc;
```
jhuber6: Can we do a C++ cast?

jhuber6 (approved):

PR #199133, opened 2026-05-21
[llvm][tools] Add debug map support llvm-dwarfdump

kastiglione:
Add debug map support to llvm-dwarfdump. This allows dwarfdump to be called on binaries which reference the object files from which it was linked. Those object files will usually contain the debug info. The debug map is only used when the binary itself has no compile units.

Assisted-by: claude

### DIFF
--- a/llvm/test/tools/llvm-dwarfdump/AArch64/Inputs/macho-oso-exe.yaml
+++ b/llvm/test/tools/llvm-dwarfdump/AArch64/Inputs/macho-oso-exe.yaml
@@ -1,0 +1,65 @@
+--- !mach-o
+FileHeader:
+  magic:           0xFEEDFACF
+  cputype:         0x0100000C
+  cpusubtype:      0x00000000
+  filetype:        0x00000002
+  ncmds:           3
+  sizeofcmds:      248
+  flags:           0x00200085
+  reserved:        0x00000000
+LoadCommands:
+  - cmd:             LC_SEGMENT_64
+    cmdsize:         152
+    segname:         __TEXT
+    vmaddr:          4294967296
+    vmsize:          4096
+    fileoff:         0
+    filesize:        4096
+    maxprot:         5
+    initprot:        5
+    nsects:          1
+    flags:           0
+    Sections:
+      - sectname:        __text
+        segname:         __TEXT
+        addr:            0x0000000100000000
+        size:            4
+        offset:          0x00000118
+        align:           2
+        reloff:          0x00000000
+        nreloc:          0
+        flags:           0x80000400
+        reserved1:       0x00000000
+        reserved2:       0x00000000
+        reserved3:       0x00000000
+  - cmd:             LC_SEGMENT_64
+    cmdsize:         72
+    segname:         __LINKEDIT
+    vmaddr:          4294971392
+    vmsize:          4096
+    fileoff:         4096
+    filesize:        64
+    maxprot:         1
+    initprot:        1
+    nsects:          0
+    flags:           0
+  - cmd:             LC_SYMTAB
+    cmdsize:         24
+    symoff:          4096
+    nsyms:           1
+    stroff:          4112
+    strsize:         16
+LinkEditData:
+  NameList:
+    - n_strx:          1
+      n_type:          0x66
+      n_sect:          0
+      n_desc:          1
+      n_value:         0
+  StringTable:
+    - ''
+    - 'test.o'
+    - ''
+    - ''
+...
--- a/llvm/test/tools/llvm-dwarfdump/AArch64/Inputs/macho-oso-obj.yaml
+++ b/llvm/test/tools/llvm-dwarfdump/AArch64/Inputs/macho-oso-obj.yaml
@@ -1,0 +1,106 @@
+--- !mach-o
+FileHeader:
+  magic:           0xFEEDFACF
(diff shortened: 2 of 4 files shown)


JDevlieghere (commented):
We already have code in dsymutil to parse and dump the debug map. Can we reuse that?

kastiglione:
ok, looking into it.

Michael137 (commented):
good idea!

kastiglione:
@JDevlieghere it's been a while since I've looked at dsymutil code, and I don't know if I've spent any time with its debug map functionality, but my initial opinion is that it isn't a good candidate for reuse for this change.

The `DebugMap` seems architected to dsymutil's needs which are of course different (more?) than dwarfdump. For example it provides mapping between the symbol addresses in the object file to their linked address in the linked binary. This seems unnecessary for dwarfdump. Looking at `MachODebugMapParser`, it contains a lot of state. I might even say that using the name "debug map" in this PR might be using a term which isn't quite right for these simpler needs.

Having said that, I know you've spent much more time with dsymutil, so I'm guessing you might have been prompting me to a direction you'd like to see taken.

kastiglione:
Updated to support OSO entries which reference archive members. In doing so, I added support for reference passing in archive members directly on the command line: `llvm-dwarfdump "path/to/archive.a(member.o)"` will now work.

JDevlieghere:
Maybe let's talk first about what you're trying to achieve. I didn't look at the PR in too much detail before my previous comment and I assumed you were implementing the equivalent of `dsymutil --dump-debug-map`. But it looks like you're going a step further and following the paths to dump their debug info.

As someone who's dealt with that a lot, I'm somewhat skeptical that that's all that useful. At least for anything but tiny test binaries. My concern is that it will just generate too much output to be useful, which is often already the case for a dSYM which is a lot smaller than the input.

A weaker counter-argument is that I also like that we're explicit about a linked binary *not* containing any debug info. If it were up to me, I would make dwarfump dump a simplified YAML debug map when you passed it a linked executable.

The other reason I care about repurposing the debug map is that for all intents and purposes, the YAML and the binary representation are equivalent. So if we wanted to make it possible to follow the object files, I would really want it to work on a debug map as well. Nick and have talked about this and he expressed interest in making the YAML the canonical debug map representation (and potentially separating it out in a side-car file). Anyway, the latter is just speculation and support in dwarfdump is at the bottom of the priority list for something like that, but if we can make it so that something like that would work out of the box, why not.

kastiglione:
> following the paths to dump their debug info.

yep

> As someone who's dealt with that a lot, I'm somewhat skeptical that that's all that useful.

can you elaborate on "dealt with that a lot"?

In the case of debug map style binaries, are users expected run dsymutil and then dwarfdump the dSYM? What about other platforms? Or are users expected to identify specific object files and then manually dwarfdump those?

> If it were up to me, I would make dwarfump dump a simplified YAML debug map when you passed it a linked executable.

and then what would users do with that? if they're calling dwarfdump, they'd presumably want to pipe this yaml into something that dumps the dwarf.

In my workflow, most files are built release, while a handful of "active" files (those I am editing or debugging), are built debug. There are other workflows which result in partial debug info too. I realize these are less common.

My belief is that many users running dwarfdump on a debug map style binary would rather get a firehose they can page or grep through, than see nothing at all.

JDevlieghere:
> can you elaborate on "dealt with that a lot"?

I meant that by working on dsymutil, I often go through this exact workflow of following the debug map to inspect/dump the input DWARF. 

> In the case of debug map style binaries, are users expected run dsymutil and then dwarfdump the dSYM? 

It depends, but no, I wouldn't tell someone to generate a dSYM to look at the input DWARF. I'd tell them to dump the debug map with `dsymutil`, identify the object they're interested in, and then dump that with dwarfdump. 

> What about other platforms? Or are users expected to identify specific object files and then manually dwarfdump those?

The debug map is strictly a Mach-O concept but dsymutil is cross platform. That said, I think it's a fair point that on other platforms, you likely have a dwarfdump but no dsymutil, and so I think the feature is useful. 

> and then what would users do with that? if they're calling dwarfdump, they'd presumably want to pipe this yaml into something that dumps the dwarf.

No, I'd expect them to identify the object file(s) they're interested in, and dump those. 

My whole spiel here is based on the assumption that going through all the input DWARF is impractical and that's what I'm optimizing for. That assumption is based on my personal experience, but it's possible that I'm biased toward larger C++ projects. But for something like clang, dumping the dSYM already takes minutes (I remember it taking ~20 min at some point) and with this approach I'd expect it take so long that people will just assume it never finishes.

The only useful aspect I see here (for me) is when I don't know which object file I'm interested in. Something like `--find=`. I guess `--lookup` wouldn't even work because the input DWARF is still unrelocated, so that's something that would need to be handled.

JDevlieghere:
Maybe to make this a bit more concrete:

- I like the idea of supporting the debug map in dwarfdump
- I am skeptical about dumping all the input DWARF but I see another valuable use case (i.e. `--find`)
- Besides easier access to the debug map on non-Darwin platforms, there's little value in duplicating what `dsymutil --dump-debug-map` already does, so may as well do something different for dwarfdump. 
- I still want to support a YAML debug map as input.
- If you want `--lookup` to work, you'll need to apply the relocations and you will need the data in the `DebugMap`.

PR #192205, opened 2026-04-15
[GitHub] Support issue_comment workflows in issue_write

lukel97:
This is split off from #190010. We want to add a new workflow triggered whenever a comment is added to an issue (workflow_run.event == 'issue_comment'), that also writes an comment back via the issue_write workflow.

However for issue_comment workflows, the head branch for the workflow won't be the head of the PR, but the default branch of the repository. So trying to fetch the PR based on the branch will fail.

GitHub docs seem to recommend that the PR number is explicitly passed via an artifact in these cases: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#using-data-from-the-triggering-workflow

This PR adds support for this so we can eventually leave comments from the test-suite.yml workflow

PR #190010, opened 2026-04-01
[Github][RFC] Add workflow to diff codegen on llvm-test-suite

lukel97:
A common task when reviewing PRs in the LLVM subproject is checking out the PR locally, building it, running it on some benchmarks e.g. llvm-test-suite, and comparing the codegen against some known version.

The process is fairly laborious so this PR adds a GitHub workflow to automate it. It's triggered by commenting "/test-suite" on a PR. The workflow will kick off, build clang with the head and base of the PR, build the benchmarks in llvm-test-suite for several configurations with each version of clang, compute the diff in the output assembly via the [tdiff.py script](https://github.com/llvm/llvm-test-suite/blob/main/utils/tdiff.py), and then report back with the diffs in a comment.

Here's an example on my fork where you can see the diff of a codegen change in the RISC-V backend: https://github.com/lukel97/llvm-project/pull/7#issuecomment-4283315532

At the moment it's very simple but could be fleshed out later. Currently it builds llvm-test-suite for handful of common configurations, some cross-compiled:

- `-target aarch64-linux-gnu -march=armv9-a -O3`
- `-target riscv64-linux-gnu -march=rva23u64 -O3`
- `-target x86_64-linux-gnu -O3`

We could eventually extend this to accept arbitrary targets and flags in the comment. It would also be nice to support LTO diffs in future but we will need to add some extra support in llvm-test-suite to extract the asm during the link step. 

It also just comments a link to download the codegen diffs, but could eventually also include some output from the ./utils/compare.py script about e.g. changes in code size or statistics. For now, those results are just uploaded as an artifact.

In terms of worker resources, running it on the free GitHub hosted workers is good enough. Building Clang takes a while, over an hour, but building the test-suite only takes around 10 minutes. But we could stick it on something beefier if we wanted the feedback to be faster.

This workflow requires the PR to be mergeable, as it wants to get the diff of the "mergeability" commit that GitHub generates for each PR. That way the diff is always between the latest version of the base branch and the PR, not the base branch at the time the PR was created.

PR #199134, opened 2026-05-21
[CI] Remove unnecessary PR validation

ojhunt:
There's no reason to try to verify the PR number here - if it's not correct we error out anyway, so just ignore it entirely and pull the information in directly from github.

### DIFF
--- a/.github/workflows/issue-write.yml
+++ b/.github/workflows/issue-write.yml
@@ -51,13 +51,9 @@ jobs:
           script: |
             var fs = require('fs');
             var comments = []
-            var pr_number = 0
             for (local_file of fs.readdirSync('.')) {
               if (local_file.startsWith("comments")) {
                 comments.push(...JSON.parse(fs.readFileSync(local_file)))
-              }
-              if (local_file.startsWith("pr_number")) {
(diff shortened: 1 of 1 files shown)


ojhunt:
I guess we could retain the original pr logic to report an error on mismatch? not sure if it's worth it though?

lukel97 (commented):
The pr_number file isn't for validation, it's needed for https://github.com/llvm/llvm-project/pull/190010. The issue_write event doesn't expose the PR number in the payload so it needs to be plumbed through manually

See https://github.com/llvm/llvm-project/pull/192205 for the rationale

ojhunt:
> The pr_number file isn't for validation, it's needed for #190010. The issue_write event doesn't expose the PR number in the payload so it needs to be plumbed through manually
> 
> See #192205 for the rationale

Well that would certainly do it - I saw the error(?) case and assumed it must always be available. Seems like closing this is the right thing to do?

boomanaiden154 (commented):
Yeah, closing this is probably the most reasonable course of action.

ojhunt:
Closing, reason: I am a muppet :D

PR #199136, opened 2026-05-21
[MLIR] Initial upstream of polygeist dialect

wsmoses:
Polygeist is an incubator project of LLVM (http://github.com/llvm/Polygeist), which contains many features, notably raising passes, including for backend retargettng, parallelism, and polyhedral analyses.
 
This PR is mostly just the creation of the dialect and upstreaming of first ops, pointer 2 memref and memref 2 pointer and relevant canonicalizations.

Passes and other operations to follow subsequently.

See https://discourse.llvm.org/t/rfc-add-scf-to-affine-conversion-pass-in-mlir/88036


mlir/include/mlir/Dialect/Polygeist/IR/PolygeistOps.td
```diff
@@ -0,0 +1,40 @@
+#ifndef POLYGEIST_OPS
+#define POLYGEIST_OPS
+
+include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/ViewLikeInterface.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+include "mlir/Dialect/LLVMIR/LLVMOpBase.td"
+include "mlir/Dialect/Polygeist/IR/PolygeistBase.td"
+
+def Memref2PointerOp
```
chelini, line 10: Have we considered targeting !ptr.ptr (the generic ptr dialect type) rather than !llvm.ptr? With that change the two conversion ops themselves would no longer be LLVM-related and could live directly in the memref dialect?

joker-eph (changes requested):
You're linking an RFC about adding a pass, it's not clear why a dialect is relevant to the RFC.

You should start by addressing this more clearly in the RFC, see questions: https://discourse.llvm.org/t/rfc-add-scf-to-affine-conversion-pass-in-mlir/88036/22


This looks like a wider scope contribution here and like deserve its own RFC IMO.

Issue #199135, opened 2026-05-21
LLVM.CodeGen/PowerPC/fast-isel-cmp-imm.ll does not pass despite my PR being unrelated


InakiVA:
https://github.com/llvm/llvm-project/pull/199073

InakiVA:
Appears to be same issue previously generated on https://github.com/llvm/llvm-project/issues/199099 , not sure if there is another from even earlier

efriedma-quic:
Should be fixed by https://github.com/llvm/llvm-project/pull/199140

PR #199137, opened 2026-05-22
[AArch64] Fix hasNearbyPairedStore to handle non-inbounds GEPs

kunalspathak:
Problem: `hasNearbyPairedStore` uses `stripAndAccumulateInBoundsConstantOffsets` to decompose store pointers into (base, offset) pairs and check whether two stores are 16 bytes apart. This fails when LSR has rewritten pointer arithmetic into non-inbounds GEPs because the function refuses to look through them. The two stores then appear to have different base pointers and the check returns false. When this happens, `lowerInterleavedStore` proceeds to emit `ST2` for a pattern that would be more profitable as `zip+stp`,  since the load-store optimizer can pair adjacent stores into `STP` but cannot merge `ST2` with anything. On a bf16-to-fp32 NEON conversion loop this causes a regression from 11 to 17 instructions per iteration. 
Note: Interleaved stores support was added for RISCV in https://github.com/llvm/llvm-project/pull/115354. Turning this off produces the desired STP instructions. 

https://godbolt.org/z/1afsjPd3e

Fix: Switch to `stripAndAccumulateConstantOffsets` with `AllowNonInbounds=true`. The function is a bail-out heuristic doing pure address arithmetic, so the inbounds semantic guarantee is not needed for correctness.


kunalspathak:
Test failures seems unrelated as reported in https://github.com/llvm/llvm-project/issues/199135

davemgreen (approved):
LGTM

PR #199138, opened 2026-05-22
[libc++] Reimplement `optional<T&>` and granularize `<optional>` header

smallp-o-p:
- Reimplement `optional<T&>` as a proper partial specialization
- Break up the `<optional>` header into different parts:
  - Common base class
  - Comparison operators
  - `nullopt_t`
  - `std::swap` overload
  - `hash<optional<T>>`
  - `optional<T>`
  - `optional<T&>`
- This also adjusts the tests that relied on error messages being in `optional`. 

Certain functions in new (unimplemented) headers return `optional<T&>`, and this should allow us to implement those without needing to include the entirety of `<optional>`

This leverages a similar granularization strategy as `<vector>`.


libcxx/include/__optional/common.h
```diff
@@ -0,0 +1,595 @@
...
+    __sfinae_assign_base< (is_copy_constructible_v<_Tp> && is_copy_assignable_v<_Tp>),
+                          (is_move_constructible_v<_Tp> && is_move_assignable_v<_Tp>)>;
+
+template <class _Tp, class = void>
+struct __optional_iterator_base : __optional_move_assign_base<_Tp> {
+  using __optional_move_assign_base<_Tp>::__optional_move_assign_base;
+};
+
+template <class _Tp>
+class optional;
```
frederick-vs-ja: I think it's better to move the forward declaration into `<__fwd/optional.h>`. We don't want to forward-declare classes arbitrarily.

libcxx/include/__optional/common.h
```diff
@@ -0,0 +1,595 @@
...
+  throw bad_optional_access();
+#  else
+  _LIBCPP_VERBOSE_ABORT("bad_optional_access was thrown in -fno-exceptions mode");
+#  endif
+}
+
+struct __optional_construct_from_invoke_tag {};
+
+template <class _Tp, bool = is_trivially_destructible<_Tp>::value>
+struct __optional_destruct_base;
```
frederick-vs-ja, line 94: Some helper base classes should be moved to `<__optional/optional.h>` IIUC.

libcxx/include/__optional/swap.h
```diff
@@ -0,0 +1,38 @@
+// -*- C++ -*-
```
frederick-vs-ja, line 1: IIUC, both `optional.h` and `optional_ref.h` should provide comparison operators, `swap`, and `hash` specialization to avoid ODR-violation in SFINAE results. I guess it would be simpler to provide comparison operators, `swap`, and `hash` specialization in the same internal header.

libcxx/include/__optional/common.h
smallp-o-p: I think the only base classes needed for `optional<T&>` is the reference overload of `__optional_storage_base` and the reference specialization of `__optional_iterator_base`. Would it be better to just cut those out directly to use inside `optional_ref.h`? We can then simplify the `optional<T>` base but I fear there may be concerns w.r.t ABI.

smallp-o-p:
Also CC @philnik777

PR #199139, opened 2026-05-22
[AArch64] Fold UMOV(lane 0) + GPR store in FPR store

### LINKED ISSUE #137086: AArch64: uses store from GP reg where vectorized reg would be better
Repro:
```
#include <string.h>
#include <arm_neon.h>
#include <arm_sve.h>

void foobar(uint16_t *p, uint8_t *p2, uint64x2_t vec, int x) {
  vec += (uint64x2_t){3, 4};
  if (x) {
    asm volatile(""); // side-effect so conditional move cannot be used.
    vec |= (uint64x2_t){0x12, 0x34};
    *p2 = vec[1];
    *p = vec[0];
  } else {
    // Commenting out the next two instructions will improve code in the "then" branch!
    vec ^= (uint64x2_t){0x34, 0x12};
    *p = vec[0];
  }
}

// clang++ -target aarch64-redhat-linux-gnu -O3 -S -o - test.cpp -march=armv9-a+sve2+fp16
```

The compiler ends up compiling the "then" part of the branch as:
```
...
       umov    w8, v0.h[0]
       st1     { v0.b }[8], [x1]
       strh    w8, [x0]
```
while this could be done without going throgh the w8 GP register:
```
       st1     { v0.b }[8], [x1]
       str     h0, [x0]
```

kunalspathak:
Problem: LLVM generates `umov w8, v0.h[0]` + `strh w8, [x0]` instead of `str h0, [x0]` when storing vector lane 0 to memory, specifically when SimplifyCFG merges stores across branches -- splitting the extractelement and store into different basic blocks and preventing the existing DAG combine from firing.

https://godbolt.org/z/v5G9ohMPa

Root cause: SimplifyCFG creates a PHI + merged store in a successor block. SelectionDAG ISel processes each block independently, so it lowers the extract to `UMOV` (GPR) in the predecessor and the store sees only a GPR value via the PHI. Late tail duplication puts the store back in the same block, but the `UMOV` is already baked in.

Fix: Added a post-RA peephole in `AArch64LoadStoreOptimizer` (step 6 in `optimizeBlock`) that recognizes `UMOVvi*_idx0` + GPR store patterns and replaces them with direct FPR sub-register stores. The peephole:
- Handles all element sizes: i8 (`bsub`), i16 (`hsub`), i32 (`ssub`), i64 (`dsub`)
- Correctly updates liveness by clearing intervening kill flags on the vector register
- Bails out if the GPR value has other uses, the vector register is clobbered, or the store doesn't kill the GPR

Assisted-by: Claude

Fixes: https://github.com/llvm/llvm-project/issues/137086

cc: @MatzeB @efriedma-quic


davemgreen (commented):
Do you have an llvm ir test case for this, to show where it comes up?

kunalspathak:
Test failures seems unrelated and are reported below:
- https://github.com/llvm/llvm-project/issues/199135
- https://github.com/llvm/llvm-project/issues/199225

kunalspathak:
> Do you have an llvm ir test case for this, to show where it comes up?

Sure, here's a minimal IR test case that shows where this comes up:

```llvm
define void @extract_store_post_simplifycfg(ptr %p, <8 x i16> %vec, i32 %cond) {
entry:
  %tobool = icmp eq i32 %cond, 0
  br i1 %tobool, label %if.else, label %if.then

if.then:
  call void asm sideeffect "", ""()
  %or = or <8 x i16> %vec, <i16 18, i16 0, i16 0, i16 0, i16 0, i16 0, i16 0, i16 0>
  %lane.then = extractelement <8 x i16> %or, i64 0
  br label %if.end

if.else:
  %lane.else = extractelement <8 x i16> %vec, i64 0
  %xor = xor i16 %lane.else, 52
  br label %if.end

if.end:
  %storemerge = phi i16 [ %xor, %if.else ], [ %lane.then, %if.then ]
  store i16 %storemerge, ptr %p, align 2
  ret void
}
```

This is what the IR looks like after SimplifyCFG merges the stores from both branches into a PHI + single store in `if.end`. The `extractelement` in `if.then` ends up in a different block from the store, so ISel processes them in separate DAGs and lowers the extract to `UMOV` (since it only sees a `CopyToReg` feeding the PHI, not a store). Tail duplication later sinks the store back into each branch, but by then the `UMOV` is already committed.

MIR after ISel shows the separation clearly:

```
bb.1.if.then:
  %6:fpr128 = ORRv8i16 %4:fpr128(tied-def 0), 18, 0
  %7:gpr32 = UMOVvi16_idx0 killed %6:fpr128, 0   ; lowered to GPR
  %0:gpr32all = COPY %7:gpr32                     ; feeds PHI
  B %bb.3

bb.3.if.end:
  %2:gpr32 = PHI %0:gpr32all, %bb.1, %1:gpr32all, %bb.2
  STRHHui %2:gpr32, %3:gpr64common, 0             ; store in separate block
```

After tail duplication sinks the store, we end up with `umov w8, v0.h[0]` + `strh w8, [x0]` instead of just `str h0, [x0]`. The existing DAG combine (`performSTORECombine`) can't help here because it needs both the extract and store in the same SelectionDAG.

I've added this as a `.ll` test at `llvm/test/CodeGen/AArch64/umov-fpr-store-from-phi.ll` alongside the MIR tests.

kunalspathak:
@davemgreen - do you need more information?

llvm/test/CodeGen/AArch64/ldst-opt-umov-fpr-store.mir
```diff
@@ -0,0 +1,169 @@
+# RUN: llc -mtriple=aarch64-linux-gnu -run-pass=aarch64-ldst-opt -o - %s | FileCheck %s
```
davemgreen, line 2: Use update_mir_test_checks.py can help keep the tests consistent and check all the output code.

llvm/test/CodeGen/AArch64/umov-fpr-store-from-phi.ll
```diff
@@ -0,0 +1,52 @@
...
+;   str   h0, [x0]           ; direct FPR store
+;
+; The post-RA peephole in aarch64-ldst-opt fixes this by folding
+; UMOVvi*_idx0 + GPR store --> FPR sub-register store.
+
+; With the fix applied, if.then should use a direct FPR store.
+; if.else legitimately needs the GPR (the XOR is a scalar GPR op),
+; so the umov there is expected.
+
+; CHECK-LABEL: extract_store_post_simplifycfg:
```
davemgreen: Use update_llc_test_check.py to show the whole thing. Someone will likely use it eventually anyway.

llvm/lib/Target/AArch64/AArch64LoadStoreOptimizer.cpp
```diff
@@ -3012,6 +3017,110 @@ bool AArch64LoadStoreOpt::tryToMergeIndexLdSt(MachineBasicBlock::iterator &MBBI,
...
+  MachineBasicBlock *MBB = StoreMI.getParent();
+  unsigned FPRStoreOpc = 0, SubRegIdx = 0;
+  MCPhysReg StoreValReg = StoreMI.getOperand(0).getReg();
+
+  if (!StoreMI.getOperand(0).isKill())
+    return false;
+
+  // Scan backward to find the UMOV that defines the store's value register.
+  MachineInstr *UMOVMI = nullptr;
+  for (auto It = MBBI; It != MBB->begin() && !UMOVMI;) {
```
davemgreen: Put a limit on the distance it searches, in case it gets quite big. We would expect most stores not to trigger this optimization.

davemgreen (commented):
> @davemgreen - do you need more information?

No, just thinking about it. I remember seeing this come up in #137086 now, the comments from @MatzeB and @efriedma-quic are still pertinent. It doesn't fit anywhere very nicely. I think GISel has the best chance of doing this well if it can eventually treat an xor on either regbank, but does not currently know how to do that properly. It can be better if we can do this kind of transform inside ISel so it can make use of all the other combines happening at the same time.

Doing it here might be the least-worst option. It is currently only handling certain types of stores, not all addressing modes. (And it only handles one-way I guess, not fpr->gpr too, but I'm not sure that is worth making it more complex for). Should it handle other store types?

kunalspathak:
> > @davemgreen - do you need more information?
> 
> No, just thinking about it. I remember seeing this come up in #137086 now, the comments from @MatzeB and @efriedma-quic are still pertinent. It doesn't fit anywhere very nicely. I think GISel has the best chance of doing this well if it can eventually treat an xor on either regbank, but does not currently know how to do that properly. It can be better if we can do this kind of transform inside ISel so it can make use of all the other combines happening at the same time.


I agree this doesn't fit anywhere very cleanly. ISel has already committed to GPR by the time tail duplication re-creates the store, and there's no mechanism to revisit that decision. GISel handling xor on either regbank would be the "right" fix long-term, but as you pointed it's not there yet.

> Doing it here might be the least-worst option. 

Yes that's what I concluded.

> It is currently only handling certain types of stores, not all addressing modes. (And it only handles one-way I guess, not fpr->gpr too, but I'm not sure that is worth making it more complex for). Should it handle other store types?

I've extended this to handle unscaled immediate and register-offset stores in addition to the unsigned-immediate ones, so it now covers the common non-pre/post addressing modes. Pre/post-indexed stores have a different operand layout (writeback tied def on operand 0) and are unlikely to appear with `UMOV` in practice, so I left those out for now. I also added a scan limit (`--aarch64-umov-fold-scan-limit`, default 16) to bound the backward search, since we expect most candidates to be within a few instructions.

For the FPR to GPR direction — I don't think it's worth the complexity right now. The GPR to FPR case is what
 we actually see from the ISel + tail-dup interaction, and I haven't come across the reverse in practice, but can add in future if we see such scenarios.

llvm/lib/Target/AArch64/AArch64LoadStoreOptimizer.cpp
```diff
@@ -3012,6 +3022,140 @@ bool AArch64LoadStoreOpt::tryToMergeIndexLdSt(MachineBasicBlock::iterator &MBBI,
...
+      continue;
+    if (++Count > UMOVFoldLimit)
+      return false;
+    if (MI.readsRegister(StoreValReg, TRI))
+      return false;
+    if (MI.definesRegister(StoreValReg, TRI)) {
+      SubRegIdx = getUMOVSubRegIdx(MI.getOpcode());
+      if (!SubRegIdx)
+        return false;
+      UMOVMI = &MI;
```
davemgreen, line 3127: break here?

llvm/lib/Target/AArch64/AArch64LoadStoreOptimizer.cpp
```diff
@@ -3012,6 +3022,140 @@ bool AArch64LoadStoreOpt::tryToMergeIndexLdSt(MachineBasicBlock::iterator &MBBI,
...
+  MachineBasicBlock::iterator B = MBB->begin();
+  unsigned SubRegIdx = 0;
+  unsigned Count = 0;
+  for (auto It = MBBI; It != B && !UMOVMI;) {
+    MachineInstr &MI = *--It;
+    if (MI.isDebugInstr())
+      continue;
+    if (++Count > UMOVFoldLimit)
+      return false;
+    if (MI.readsRegister(StoreValReg, TRI))
```
davemgreen, line 3121: Do we need to check that the store itself doesn't have another use of the register?

davemgreen (commented):
It looks like the extra opcodes helped this trigger more. Post and pre inc might help more cases too (but you don't need to do that here).

llvm/lib/Target/AArch64/AArch64LoadStoreOptimizer.cpp
kunalspathak: There is already a loop termination condition `!UMOVI`.

llvm/lib/Target/AArch64/AArch64LoadStoreOptimizer.cpp
kunalspathak: good point. fixed. I have also added some additional checks around volatile store, overlapping register.

llvm/lib/Target/AArch64/AArch64LoadStoreOptimizer.cpp
davemgreen: Oh I see. I wasnt expecting the condition to be there - that seems easy to miss. I would move the break here unless you have a reason to keep it where it is.

davemgreen (approved):
Otherwise, LGTM

llvm/lib/Target/AArch64/AArch64LoadStoreOptimizer.cpp
kunalspathak: sure, fixed.

davemgreen (approved):
Thanks. Are you happy for us to hit merge? I never know who has access.

kunalspathak:
> Thanks. Are you happy for us to hit merge? I never know who has access.

I don't have merge access. @MatzeB is this ok to merge?

PR #199143, opened 2026-05-22
[libc][freebsd] skip exhaustive math tests in FreeBSD CI

SchrodingerZhu:
Remove expensive tests to mitigate the long running time.


boomanaiden154 (commented):
Why are we going from full tests to build only?

What's the runtime like if we just disable the exhaustive math tests?

lntue:
> Why are we going from full tests to build only?
> 
> What's the runtime like if we just disable the exhaustive math tests?

I agree, let just disable the MPFR math tests first to see how much time it takes for the rest of them.

SchrodingerZhu:
Pushed. Let's check the runtime first.

lntue (approved):

SchrodingerZhu:
Discussed with @lntue, 26mins testing seems acceptable for now.

PR #198935, opened 2026-05-20
DebugInfo: Shrink-to-fit some containers to reduce peak memory usage

dwblaikie:
Changes originally by Tipp Moseley (tipp@google.com)

PR #199145, opened 2026-05-22
Revert "DebugInfo: Shrink-to-fit some containers to reduce peak memory usage"

efriedma-quic:
Reverts llvm/llvm-project#198935

I think this broke llvm/test/tools/llvm-gsymutil/X86/elf-dwo.yaml .


dwblaikie:
Thanks!

PR #199146, opened 2026-05-22
[X86] Fold XOR of two VGF2P8AFFINEQB instructions with same matrix

### LINKED ISSUE #196879: [X86] Missed Fold: `vgf2p8affineqb(x0, m) ^ vgf2p8affineqb(x1, m)` => `vgf2p8affineqb(x0 ^ x1, m)`
Given that `vgf2p8affineqb` performs an XOR permutation of the input, and that XORs are associated, XORing two values that where transformed in the same manner can be reassociated to happen before. Their immediates also need to be XORed together as well:

```asm
gfniXor_src:
        gf2p8affineqb   xmm0, xmm2, 1
        gf2p8affineqb   xmm1, xmm2, 3
        pxor    xmm0, xmm1
        ret
```

```asm
gfniXor_tgt:
        pxor    xmm0, xmm1
        gf2p8affineqb   xmm0, xmm2, 2
        ret
```

https://godbolt.org/z/16or45959

Most nodes that lower to `vgf2p8affineqb` already perform this fold so this mostly applies to intrinsics.

WalterKruger:
Adds an optimization to fold a XOR between two `vgf2p8affineqb` instructions that share the same matrix by XORing their sources beforehand. This patch:

- Can eliminate one `vgf2p8affineqb` instruction.
- Doesn't occur if either affine is multi use, preventing an increase in code size.
- Includes test coverage for both positive and negative cases.

Fixes #196879


WalterKruger:
cc @RKSimon

The Windows CI might be unrelated.

llvm/lib/Target/X86/X86ISelLowering.cpp
```diff
@@ -56318,29 +56352,46 @@ static SDValue combineXorWithGF2P8AFFINEQB(SDNode *N, const SDLoc &DL,
...
-// together. XORing two affine transformations of the same input can be folded
-// by XORing both their matrices and immediates together.
+// Given that vgf2p8affineqb performs a XOR permutation, two affines that share
+// a operand can be reassociated through a standalone XOR.
 static SDValue combineXorWithTwoGF2P8AFFINEQB(SDNode *N, const SDLoc &DL,
                                               SelectionDAG &DAG, EVT VT) {
   using namespace SDPatternMatch;
 
-  SDValue X0, Y0, Y1;
+  SDValue X0, X1, Y0, Y1;
```
phoebewang: Either use M0, M1 here, or y0, y1 in the comments to make them consistent

phoebewang (approved):
LGTM.

RKSimon (approved):
LGTM

PR #199147, opened 2026-05-22
[lit] Add --check to run only selected RUN lines from a test

jofrn:
Adds `llvm-lit --check=LIST <test>` which keeps only the listed RUN
directives in the test and discards the rest. `LIST` is a
comma-separated mix of 0-indexed integers and ranges (e.g.
`--check=0,2,4-6`). The selection is applied to the
`parseIntegratedTestScript` output.

Stacked above https://github.com/llvm/llvm-project/pull/199390; and below https://github.com/llvm/llvm-project/pull/200618.

### DIFF
--- a/llvm/utils/lit/lit/LitConfig.py
+++ b/llvm/utils/lit/lit/LitConfig.py
@@ -42,6 +42,7 @@ class LitConfig(object):
         per_test_coverage=False,
         gtest_sharding=True,
         update_tests=False,
+        checkFilter=None,
     ):
         # The name of the test runner.
         self.progname = progname
@@ -96,6 +97,7 @@ class LitConfig(object):
         self.gtest_sharding = bool(gtest_sharding)
         self.update_tests = update_tests
         self.test_updaters = [diff_test_updater]
+        self.checkFilter = checkFilter
 
     @property
     def maxIndividualTestTime(self):
--- a/llvm/utils/lit/lit/TestRunner.py
+++ b/llvm/utils/lit/lit/TestRunner.py
@@ -1921,6 +1921,23 @@ def _expandLateSubstitutionsExternal(commandLine):
         commandLine = "%s && test -e %s" % (commandLine, filePath)
     return commandLine
 
+def _applyCheckFilter(test, check_filter):
+    """Keep only CommandDirective entries whose 0-indexed integer(s)
+    is/are in check_filter.
+    """
+    if check_filter is None:
+        return test
+    out = []
+    n = -1
+    for d in test:
+        if isinstance(d, CommandDirective):
+            n += 1
+            if n not in check_filter:
+                continue
+        out.append(d)
+    return out
+
+
 def executeShTest(
     test, litConfig, useExternalSh, extra_substitutions=[], preamble_commands=[]
 ):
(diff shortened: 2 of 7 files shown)


boomanaiden154 (commented):
What exactly is the use case for this?

This seems like a lot of complexity for an extremely niche feature.

jofrn:
What do you mean? It isn't much code: `_check_filter` is the only real addition; the rest is boilerplate. It allows one to select a single RUN line from a test file to run, instead of requiring all to be run. This allows for more fine-grained development; rather than having to copy out tests or comment out run lines that one doesn't want to be currently testing.

Another feature I'm working on is to run based on function name also. Imagine a lit test file with hundreds of functions and tens of RUN lines; the dev may only want to run a subset.

boomanaiden154:
> What do you mean? It isn't much code. It allows one to select a single RUN line from a test file to run, instead of requiring all to be run. This allows for more fine-grained development; rather than having to copy out tests or comment out run lines that one doesn't want to be currently testing.

This to me doesn't actually seem more convenient to me than working in a separate file/commenting things out which is what I normally do when hacking on lit tests.

Also, just because something does not have a lot of code does not mean anything about the complexity of that code and/or the maintenance burden to keep it working. In my opinion, this patch doesn't meet the bar of utility to maintenance ratio to make sense to land.

> Another feature I'm working on is to run based on function name also. Imagine a lit test file with hundreds of functions and tens of RUN lines; the dev may only want to run a subset.

I would expect such a feature to have the same problem as this one. It's also not clear how that's different. What exactly is a function name in this context?

jofrn:
> This to me doesn't actually seem more convenient to me than working in a separate file/commenting things out which is what I normally do when hacking on lit tests.

Command line is important; not just copy-pasting from editor.

One won't have to comment out things, and can work from the command line. Better for scripting: `git bisect run sh -c 'llvm-lit --check=2 path/to/test.ll'` doesn't require pre-bisect file edits or post-bisect cleanup.

> Also, just because something does not have a lot of code does not mean anything about the complexity of that code and/or the maintenance burden to keep it working. In my opinion, this patch doesn't meet the bar of utility to maintenance ratio to make sense to land.

If the code regresses, it only affects users who passed `--check`. RUN-line parsing is already broadly covered by lit's existing tests; this change adds one more walk of `CommandDirective`.

> I would expect such a feature to have the same problem as this one. It's also not clear how that's different. What exactly is a function name in this context?

For example, `--fn=foo` narrows print-pass output (`-print-after=isel`, etc.) for function `@foo` in a `.ll` test, by injecting (for example) `--filter-print-funcs=foo` into llc/opt invocations on the RUN line. Name is whatever the function is called in the IR (`define ... @foo(...)`).

jofrn:
> > This to me doesn't actually seem more convenient to me than working in a separate file/commenting things out which is what I normally do when hacking on lit tests.
>
> Command line is important; not just copy pasting from editor.

I'd say this is a strong use-case for this. Why else with preference of Unix? : )

boomanaiden154 (changes requested):
https://github.com/llvm/llvm-project/pull/200351#pullrequestreview-4396534074

PR #199149, opened 2026-05-22
[LifetimeSafety] Add fix-it for misplaced lifetimebound attributes

### LINKED ISSUE #198634: [LifetimeSafety] Add fix-its for misplaced lifetimebound diagnostics
After #197753, we now warn when `lifetimebound` is written on the wrong declaration. This diagnostic would ideally have a fix-it that adds the attribute to the declaration that should carry it.

Currently we only warn that the user should add a `lifetimebound` attribute:

```cpp
int* get(int &x); // warning: 'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead

int* get(int &x [[clang::lifetimebound]]) {
  return &x;
}
```

It would be more user-friendly if we could also create a suggestion with a `lifetimebound` attribute:

```cpp
int* get(int &x); // fix-it: add [[clang::lifetimebound]] here

int* get(int &x [[clang::lifetimebound]]) {
  return &x;
}
```

After applying the fix-it:

```cpp
int* get(int &x [[clang::lifetimebound]]);

int* get(int &x [[clang::lifetimebound]]) {
  return &x;
}
```

davidmenggx:
This patch adds a fix-it hint for `warn_lifetime_safety_intra_tu_misplaced_lifetimebound` and `warn_lifetime_safety_cross_tu_misplaced_lifetimebound` to the appropriate declaration.

The fix-it attribute is emitted in the correct location, accounting for pure virtual functions, overrides, trailing return types, and default arguments. The message is suppressed for macros.

Resolves #198634


davidmenggx:
@usx95 I've given this an attempt, let me know if it seems reasonable or any suggestions

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
```diff
@@ -74,29 +90,63 @@ struct IntraSuppressedObj {
...
+  return obj;
+}
+
+struct Base {
+  virtual const MyObj& virtual_get(const MyObj& obj) const = 0;
+};
+
+struct Derived : Base {
+  auto virtual_get(const MyObj& obj) const -> const MyObj& override; // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
+  // CHECK: fix-it:"{{.*}}":{[[@LINE-1]]:36-[[@LINE-1]]:36}:" {{\[\[clang::lifetimebound\]\]}}"
```
davidmenggx: I hard coded the column number here and in line 126. Is there a better way to ensure that the placement of the attribute is syntactically valid?

clang/lib/Sema/SemaLifetimeSafety.h
```diff
@@ -297,9 +297,25 @@ class LifetimeSafetySemaHelperImpl : public LifetimeSafetySemaHelper {
...
+      if (const TypeSourceInfo *TSI = FDecl->getTypeSourceInfo())
+        FixItLoc =
+            Lexer::getLocForEndOfToken(TSI->getTypeLoc().getEndLoc(), 0,
+                                       S.getSourceManager(), S.getLangOpts());
+      else
+        FixItLoc = DiagLoc;
+
+      if (FixItLoc.isValid() && !FixItLoc.isMacroID())
+        DB << FixItHint::CreateInsertion(FixItLoc, " [[clang::lifetimebound]]");
+    }
```
NeKon69: We probably want the fix it to be placed after the diagnostic is emitted same as for the other suggestions.

clang/lib/Sema/SemaLifetimeSafety.h
```diff
@@ -297,9 +297,25 @@ class LifetimeSafetySemaHelperImpl : public LifetimeSafetySemaHelper {
...
             : diag::warn_lifetime_safety_intra_tu_misplaced_lifetimebound;
-    S.Diag(Lexer::getLocForEndOfToken(FDecl->getEndLoc(), 0,
-                                      S.getSourceManager(), S.getLangOpts()),
-           DiagID);
+
+    SourceLocation DiagLoc = Lexer::getLocForEndOfToken(
+        FDecl->getEndLoc(), 0, S.getSourceManager(), S.getLangOpts());
+
+    // Scope so diagnostic emits first.
+    {
```
NeKon69: I don't think we want this scope here.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-cross-tu.cpp
```diff
@@ -1,17 +1,20 @@
 // RUN: rm -rf %t
 // RUN: split-file %s %t
 // RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-cross-tu-misplaced-lifetimebound -Wno-dangling -I%t -verify %t/cross.cpp
+// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-cross-tu-misplaced-lifetimebound -Wno-dangling -I%t -fdiagnostics-parseable-fixits %t/cross.cpp 2>&1 | FileCheck %s
```
NeKon69, line 4: Please also add a validating run afterwards

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
```diff
@@ -1,19 +1,28 @@
 // RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-intra-tu-misplaced-lifetimebound -Wno-dangling -verify %s
+// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-intra-tu-misplaced-lifetimebound -Wno-dangling -fdiagnostics-parseable-fixits %s 2>&1 | FileCheck %s
```
NeKon69, line 2: Same here.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
NeKon69: We could add a validating run as per my suggestion above, but these hardcoded column numbers would still be nice to keep!

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
```diff
@@ -74,29 +90,63 @@ struct IntraSuppressedObj {
...
+}
+
+#define REF_PARAM MyObj &obj
+
+MyObj &macro_param(REF_PARAM); // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
+// Fix-it suppressed for macro.
+
+MyObj &macro_param(MyObj &obj [[clang::lifetimebound]]) { // expected-note {{'lifetimebound' attribute appears here on the definition}}
+  return obj;
+}
```
NeKon69: I am pretty sure we still want the suggestion here, since we do emit it currently when parameter escapes:

```cpp
#define REF_PARAM MyObj &obj

struct [[gsl::Owner]] MyObj {};

MyObj &macro_param(REF_PARAM);

MyObj &macro_param(MyObj &obj) { return obj; }
```

Warnings:
```
test_redecl_this.cpp:5:20: warning: parameter in intra-TU function should be marked [[clang::lifetimebound]] [-Wlifetime-safety-intra-tu-suggestions]
    5 | MyObj &macro_param(REF_PARAM);
      |                    ^~~~~~~~~
      |                              [[clang::lifetimebound]]
test_redecl_this.cpp:1:19: note: expanded from macro 'REF_PARAM'
    1 | #define REF_PARAM MyObj &obj
      |                   ^~~~~~~~~~
test_redecl_this.cpp:7:41: note: param returned here
    7 | MyObj &macro_param(MyObj &obj) { return obj; }
      |                                         ^~~
1 warning generated.

**Review by NeKon69 — COMMENTED** [review]
Hey! Thanks for the interest in LifetimeSafety! Here are a few suggestions for improving your PR:

1. Currently, you might not always get the correct attribute placement. Please see how we do that in `suggestLifetimeboundToParmVar` and `suggestLifetimeboundToImplicitThis`, and consider refactoring these functions to reuse that logic for your suggestions.

2. I also left a few inline code suggestions for you.

**Comment by davidmenggx** [comment]
@NeKon69 Thank you for the suggestions! Lmk how it looks now

**Review by NeKon69 — COMMENTED** [review]
Thanks!

One thing I noticed is that you haven’t reused the logic for getting the insertion point yet. I’m thinking it might make sense to create a helper function that takes either a `ParmVarDecl *` or a `CXXMethodDecl *` and returns both the fix-it text and the location. That way, the same logic could be reused in both your new code and the existing code without duplicating it.

Not something that needs to be addressed in this PR, but another thing I’m a bit unsure about is whether we should move these warnings to `LifetimeSafetySuggestions` group now that they have fix-its. My concern is that this would also cause these warnings to show up in `warn-lifetime-safety-fixits.cpp` and `warn-lifetime-safety-suggestions.cpp`, which probably wouldn’t be ideal.

We could maybe refactor these tests later, for example by splitting `-Wlifetime-safety-suggestions` into multiple subgroups and creating separate test files for each subgroup. However, I wouldn’t frame this as a final decision yet. We may want to wait for others’ opinions on that.

**Comment by davidmenggx** [comment]
@NeKon69 I see, for the first point something like this?

**Comment by NeKon69** [comment]
Yup, this looks good to me. Let’s wait for others to join the discussion as well, they might point out something I missed.

File: clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp (hunk 1 above)
Comment (usx95) [reply clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp]: Hardcoded numbers does not fail gracefully and are hard to debug and maintain.

There is a slightly inaccurate but still useful way to have the target loc on a separate line and validate the fixit appears on that line.

E.g. `auto virtual_get(const MyObj& obj) const -> const MyObj& override;` could look like
```cpp
auto virtual_get(const MyObj& 
                              obj // CHECK: ...
                 ) const -> const MyObj& override; 
```
Similar for `default_arg_param`.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
usx95: There could be a contention on the location here actually. E.g. In principle, fixit can also be attached inside the macro definition as well. But we may choose not to support that.

Can you try testing multiple params and see what we currently do:

```cpp
#define REF_PARAMS MyObj &obj1, MyObj &obj2

MyObj &macro_param(bool condition, REF_PARAMS);
​
MyObj &macro_params(bool condition,
                    MyObj &obj1 [[clang::lifetimebound]],
                    MyObj &obj2 [[clang::lifetimebound]]) {
  return condition ? obj1 : obj2;
}
```

That said, I am happy with the current state of having the same behaviour for suggestions and verification. For more functional changes, e.g. handling macros, we could followup separately.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
```diff
@@ -74,29 +93,63 @@ struct IntraSuppressedObj {
...
 }
 
 struct View {
   friend View friend_redecl(MyObj &obj); // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
+  // CHECK: fix-it:"{{.*}}":{[[@LINE-1]]:{{[0-9]+}}-[[@LINE-1]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
 };
 
 // FIXME: This diagnoses an attribute inherited from another redeclaration, not one written on the definition. Once we enforce that redeclarations agree on lifetimebound, handle this with a dedicated warning and note.
 View friend_redecl(MyObj &obj [[clang::lifetimebound]]); // expected-note {{'lifetimebound' attribute appears here on the definition}}
-
```
usx95, line 94: nit: unintentional line delete ?

clang/lib/Sema/SemaLifetimeSafety.h
```diff
@@ -397,6 +377,50 @@ class LifetimeSafetySemaHelperImpl : public LifetimeSafetySemaHelper {
 
 private:
   Sema &S;
```
usx95: nit: Move the new methods before the field members.

usx95 (commented):
Overall LG. Some comments.

NeKon69:
> Not something that needs to be addressed in this PR, but another thing I’m a bit unsure about is whether we should move these warnings to `LifetimeSafetySuggestions` group now that they have fix-its. My concern is that this would also cause these warnings to show up in `warn-lifetime-safety-fixits.cpp` and `warn-lifetime-safety-suggestions.cpp`, which probably wouldn’t be ideal.

@usx95 What would you say about that?

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
NeKon69: Currently we do not warn here at all. Which makes sense to me. We probably can suggest them on a macro, but I don't know if that's a good idea. Like, what if this macro is used somewhere else where this `lifetimebound` attribute suggestion doesn't make much sense?

usx95:
> > Not something that needs to be addressed in this PR, but another thing I’m a bit unsure about is whether we should move these warnings to `LifetimeSafetySuggestions` group now that they have fix-its. My concern is that this would also cause these warnings to show up in `warn-lifetime-safety-fixits.cpp` and `warn-lifetime-safety-suggestions.cpp`, which probably wouldn’t be ideal.
> 
> @usx95 What would you say about that?

​Suggestions is coupled a lot with inference/TU-analysis while this is surface-level placement policy ideally not even requiring the lifetime analysis. I would keep them separate and the latter simpler.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
davidmenggx: Yes, we currently do not warn in this case.

usx95: Thanks for trying. Could you please add that here to document as well.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
```diff
@@ -93,10 +115,43 @@ View friend_redecl(MyObj &obj) {
...
 MyObj &spec_func<MyObj>(MyObj &obj [[clang::lifetimebound]]); // expected-note {{'lifetimebound' attribute appears here on the definition}}
 
 template <>
 MyObj &spec_func<MyObj>(MyObj &obj) { return obj; }
+
+MyObj get_default_obj();
+const MyObj &default_arg_param(const MyObj&
+    obj  // CHECK: fix-it:"{{.*}}":{[[@LINE]]:{{[0-9]+}}-[[@LINE]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
+    = get_default_obj()); // expected-warning@-2 {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
+
```
usx95: Formatting suggestion:
```cpp
const MyObj &default_arg_param(const MyObj&  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
                               obj           // CHECK: fix-it:"{{.*}}":{[[@LINE]]:{{[0-9]+}}-[[@LINE]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
                               = get_default_obj());
```
Can you please do that at other places using `// expected-warning@-2` to attach directly at the correct line number.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
davidmenggx: Sorry I may have misunderstood. We do currently warn on both the single and multiple param macro cases. 

On the multiple param case you sent, the fix-it is applied like this:
```cpp
MyObj &macro_param(bool condition, REF_PARAMS [[clang::lifetimebound]]);
```
which incorrectly expands to this:
```cpp
MyObj &macro_param(bool condition, MyObj &obj1, MyObj &obj2 [[clang::lifetimebound]]);
```
https://godbolt.org/z/T9vTv9n4h

As NeKon said perhaps we should not apply fix-its to macros in general? I can update the `getLifetimeBoundFixIt`s to prevent this.

usx95: Yeah. Annotating macro definitions can introduce some edge cases. I would avoid attaching a fixit in such cases.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-macro.cpp
```diff
@@ -0,0 +1,26 @@
+// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-intra-tu-misplaced-lifetimebound -Wno-dangling -verify %s
```
davidmenggx, line 1: I've created a new test file to not interfere with the validating run in `warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp‎`

davidmenggx:
> # 🐧 Linux x64 Test Results
> * 95248 tests passed
> * 2884 tests skipped
> * 2 tests failed


I don't believe this failure is related?

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-macro.cpp
usx95: Makes sense. Can you remove the `-intra-tu` from the file name as it is not relevant here.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
usx95: I think this is still not fixed.

davidmenggx: I believe I fixed it in the last commit, but updated again to make sure?

NeKon69: I think Utkarsh meant that you should do this in other locations too, not only here.

usx95: Yes. Sorry for not clarifying.

davidmenggx: Sorry for the confusion. I've reformatted all of the fix-it checks to attach directly to the line. Some difficulty on the implicit `this` param, I can revert those back to the original format if needed.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
```diff
@@ -8,24 +8,30 @@ struct MyObj {
...
-  // CHECK: fix-it:"{{.*}}":{[[@LINE-1]]:{{[0-9]+}}-[[@LINE-1]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
+  const MyObj &param_only(const MyObj & // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
+      obj  // CHECK: fix-it:"{{.*}}":{[[@LINE]]:{{[0-9]+}}-[[@LINE]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
+  );
   
-  const MyObj &both(const MyObj &obj, bool); // expected-warning 2 {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
-  // CHECK: fix-it:"{{.*}}":{[[@LINE-1]]:{{[0-9]+}}-[[@LINE-1]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
-  // CHECK: fix-it:"{{.*}}":{[[@LINE-2]]:{{[0-9]+}}-[[@LINE-2]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
+  const MyObj &both(const MyObj &  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
+      // CHECK: fix-it:"{{.*}}":{[[@LINE+4]]:{{[0-9]+}}-[[@LINE+4]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
```
davidmenggx: The ordering of checks here and on line 56 is needed due to the order the diagnostics are emitted between the implicit `this` param and the other param.

If this is unclear, I can revert back to the original format

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-cross-tu.cpp
```diff
@@ -1,17 +1,24 @@
...
+HeaderObj &header_param(HeaderObj &  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers in other translation units; add it to the declaration instead}}
+	obj  // CHECK: fix-it:"{{.*}}cross.h":{[[#]]:{{[0-9]+}}-[[#]]:{{[0-9]+}}}:" {{\[\[}}clang::lifetimebound{{\]\]}}"
+);
 
 struct HeaderS {
   HeaderObj data;
-  HeaderObj &header_this(); // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers in other translation units; add it to the declaration instead}}
+  HeaderObj &header_this(
+  );  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers in other translation units; add it to the declaration instead}}
+      // CHECK: fix-it:"{{.*}}cross.h":{[[#]]:{{[0-9]+}}-[[#]]:{{[0-9]+}}}:" {{\[\[}}clang::lifetimebound{{\]\]}}"
```
NeKon69: I don't quite get this, does the suggestion show up on the next line?

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
NeKon69: You can use CHECK-DAG instead.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-cross-tu.cpp
davidmenggx: No, the fix-it would appear on the same line as the warning. 

To be more explicit this could also work:
```cpp
struct HeaderS {
  HeaderObj data;
  HeaderObj &header_this(
  );  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers in other translation units; add it to the declaration instead}}
      // CHECK: cross.h:[[#WARN_LINE:]]:{{[0-9]+}}: warning: 'lifetimebound' attribute on this definition is not visible
      // CHECK: fix-it:"{{.*}}cross.h":{[[#WARN_LINE]]:{{[0-9]+}}-[[#WARN_LINE]]:{{[0-9]+}}}:" {{\[\[}}clang::lifetimebound{{\]\]}}"
};
```

`@LINE-1` doesn't work here due to the mismatch between the test file's line numbers and `cross.h`'s line numbers

NeKon69: Oh, right, thanks! (I missed the `#`)

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
```diff
@@ -27,11 +27,10 @@ struct S {
   );
   
   const MyObj &both(const MyObj &  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
-      // CHECK: fix-it:"{{.*}}":{[[@LINE+4]]:{{[0-9]+}}-[[@LINE+4]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
-      obj,
-      // CHECK: fix-it:"{{.*}}":{[[@LINE-1]]:{{[0-9]+}}-[[@LINE-1]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
+      obj,  // CHECK-DAG: fix-it:"{{.*}}":{[[@LINE]]:{{[0-9]+}}-[[@LINE]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
       bool
   );  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
+      // CHECK-DAG: fix-it:"{{.*}}":{[[@LINE-1]]:{{[0-9]+}}-[[@LINE-1]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
```
NeKon69: You can do this instead

```
  );  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}} \
      // CHECK-DAG: fix-it:"{{.*}}":{[[@LINE]]:{{[0-9]+}}-[[@LINE]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
```

davidmenggx: Could be wrong but wouldn't `@LINE` be off by 1?

NeKon69: `\` at the end

davidmenggx: I don't believe this works because `@LINE` works on the absolute line number so the continuation `\` doesn't apply

At least this is what I see when testing locally

Source ? https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-pseudo-numeric-variables

NeKon69: That's interesting. What about reordering the checks?

davidmenggx: Something like
```cpp
);  // CHECK-DAG: fix-it:"{{.*}}":{[[@LINE]]:{{[0-9]+}}-[[@LINE]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}" \
    // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
```
also doesn't work here because now Filecheck is looking for the literal pattern `fix-it:"{{.*}}":{[[@LINE]]:{{[0-9]+}}-[[@LINE]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}" \    // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}`

(i believe this is the case from what i see when testing locally)

NeKon69: Thanks for trying! Let's leave it as is then.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
usx95: I would prefer if the parameters follow the indent of the opening`(`. Same for closing `)`.
Example:
```cpp
// Before
T &both(T &  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
      arg,  // CHECK-DAG: fix-it:"{{.*}}":{[[@LINE]]:{{[0-9]+}}-[[@LINE]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
      bool
  ); 

// After
T &both(T &  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
        arg,  // CHECK-DAG: fix-it:"{{.*}}":{[[@LINE]]:{{[0-9]+}}-[[@LINE]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
        bool
        ); 
```

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-cross-tu.cpp
usx95: This is quite confusing. I did not know about the #. TIL.
 I would appreciate if we use #WARN_LINE throught instead of #.

Also match the indent with the warning in such case.

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-cross-tu.cpp
```diff
@@ -1,17 +1,25 @@
...
+// RUN: %clang_cc1 -fsyntax-only -Wlifetime-safety-cross-tu-misplaced-lifetimebound -Wno-dangling -I%t -Werror %t/cross.cpp
 
 //--- cross.h
 struct HeaderObj {
   ~HeaderObj() {}
 };
 
-HeaderObj &header_param(HeaderObj &obj); // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers in other translation units; add it to the declaration instead}}
+HeaderObj &header_param(HeaderObj &  // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers in other translation units; add it to the declaration instead}}
+	obj  // CHECK: fix-it:"{{.*}}cross.h":{[[#]]:{{[0-9]+}}-[[#]]:{{[0-9]+}}}:" {{\[\[}}clang::lifetimebound{{\]\]}}"
```
usx95: If the check is for a separate line, can you leave the line empty, and move the `obj` to the next line (indent same as `(`)

clang/test/Sema/warn-lifetime-safety-misplaced-lifetimebound-intra-tu.cpp
```diff
@@ -34,7 +51,11 @@ const MyObj &S::both(
 template <class T>
 struct MixedSpecializations {
   T data;
-  T &both(T &arg, bool); // expected-warning 2 {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
+  T &both(T &   // expected-warning {{'lifetimebound' attribute on this definition is not visible to callers before the definition; add it to the declaration instead}}
+          arg,  // CHECK-DAG: fix-it:"{{.*}}":{[[@LINE]]:{{[0-9]+}}-[[@LINE]]:{{[0-9]+}}}:" {{\[\[clang::lifetimebound\]\]}}"
```
usx95, line 63: For a separate PR:
We highlight the complete param declaration but attach the warning to the beginning. It would make sense to attach the text to the place where lifetimebound will go here. This would match the fixit and warning location here.

https://godbolt.org/z/Pb9zjG6nx

usx95 (approved):
LGTM. Thanks.

NeKon69:
Thanks! You have conflicts on your branch, please resolve them so the CI can run.

usx95:
Thanks David for the contribution. I will land this for now.

davidmenggx:
Thank you both for the feedback and advice

PR #199150, opened 2026-05-22
[CoroSplit] Rename Suspend/End to AlwaysKill/NeverKill (NFC)

NewSigma:
Rename them so that we can generalize to more intrinsics, for example, `llvm.coro.is_in_ramp` in #198226


ChuanqiXu9 (approved):

NewSigma:
Thanks for the code review.

PR #199155, opened 2026-05-22
[InstCombine] Avoid propagating invalid metadata in FoldOpIntoSelect

Shiyas-N:
Fixes #186471

FoldOpIntoSelect may create a select with a different result type from
the original instruction. The existing implementation blindly copied all
metadata from the original select, which could propagate invalid
type-specific metadata to the transformed instruction.

In particular, folding an fcmp over a floating-point select could copy
!fpmath metadata onto a non-FP select, producing invalid IR and causing
verifier failures.

This change preserves only metadata that remains valid for the
transformed select and propagates !fpmath only for FP-typed selects.
Debug locations are also preserved explicitly.

### DIFF
--- a/llvm/lib/Transforms/InstCombine/InstructionCombining.cpp
+++ b/llvm/lib/Transforms/InstCombine/InstructionCombining.cpp
@@ -1820,7 +1820,24 @@ Instruction *InstCombinerImpl::FoldOpIntoSelect(Instruction &Op, SelectInst *SI,
     NewTV = foldOperationIntoSelectOperand(Op, SI, TV, *this);
   if (!NewFV)
     NewFV = foldOperationIntoSelectOperand(Op, SI, FV, *this);
-  return SelectInst::Create(SI->getCondition(), NewTV, NewFV, "", nullptr, SI);
+
+  SelectInst *NewSel = SelectInst::Create(SI->getCondition(), NewTV, NewFV);
+
+  // Preserve metadata that remains valid for the transformed select.
+  if (MDNode *Prof = SI->getMetadata(LLVMContext::MD_prof))
+    NewSel->setMetadata(LLVMContext::MD_prof, Prof);
+  if (MDNode *Unpred = SI->getMetadata(LLVMContext::MD_unpredictable))
+    NewSel->setMetadata(LLVMContext::MD_unpredictable, Unpred);
+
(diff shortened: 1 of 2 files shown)


Shiyas-N:
Gentle ping for review. Thanks!

llvm/lib/Transforms/InstCombine/InstructionCombining.cpp
```diff
@@ -1831,7 +1831,24 @@ Instruction *InstCombinerImpl::FoldOpIntoSelect(Instruction &Op, SelectInst *SI,
...
     NewFV = foldOperationIntoSelectOperand(Op, SI, FV, *this);
-  return SelectInst::Create(SI->getCondition(), NewTV, NewFV, "", nullptr, SI);
+
+  SelectInst *NewSel = SelectInst::Create(SI->getCondition(), NewTV, NewFV);
+
+  // Preserve metadata that remains valid for the transformed select.
+  if (MDNode *Prof = SI->getMetadata(LLVMContext::MD_prof))
+    NewSel->setMetadata(LLVMContext::MD_prof, Prof);
+  if (MDNode *Unpred = SI->getMetadata(LLVMContext::MD_unpredictable))
+    NewSel->setMetadata(LLVMContext::MD_unpredictable, Unpred);
```
nikic: Use copyMetadata with a list of metadata to copy.

llvm/lib/Transforms/InstCombine/InstructionCombining.cpp
```diff
@@ -1831,7 +1831,24 @@ Instruction *InstCombinerImpl::FoldOpIntoSelect(Instruction &Op, SelectInst *SI,
...
+
+  // Preserve !fpmath only for FP-typed selects.
+  if (isa<FPMathOperator>(NewSel))
+    if (MDNode *FPMath = SI->getMetadata(LLVMContext::MD_fpmath))
+      NewSel->setMetadata(LLVMContext::MD_fpmath, FPMath);
```
nikic: I don't think `!fpmath` on a select really makes sense, because they are precise bitwise copies, so talking about acceptable error for them is meaningless. Probably not worth trying to forbid it, but we don't need to go out of the way to copy it.

nikic (approved):
LGTM

PR #199157, opened 2026-05-22
[flang] Fixed FIR AA's getSource() for box loads inside acc.compute_region.

vzakhari:
This patch fixes a regression caused by #198635: when we call getSource()
for a `fir.load` of a box we have to handle the input value that might be
a `BlockArgument` and pass-through it.


razvanlupusoru (approved):
Thank you.

PR #199158, opened 2026-05-22
[flang][FIRToMemRef] Get strides from descriptor for some array_coor cases.

vzakhari:
This is a follow-up on Jean's comment https://github.com/llvm/llvm-project/pull/198933#discussion_r3279535746

This patch makes use of the descriptor strides when `fir.array_coor`'s
memref is a `fir.box` that is not a fir.embox result.


jeanPerier (approved):
Thanks

PR #199160, opened 2026-05-22
[lldb] Adopt LockedPtr in Module (NFC)

JDevlieghere:
Follow-up to #198941, which introduced Locked<T> and SharedLocked<T>. Add GetObjectFileLocked, GetSymbolFileLocked, GetSymtabLocked, and GetSectionListLocked alongside the existing accessors.

The locked variants cover two things:

1. They prevent the pointer from being swapped out from under the caller. The old getters take m_mutex only during lazy initialization and release it before returning. The unique_ptr or shared_ptr that owns the pointee can therefore be reassigned by another thread while the caller still holds the raw value. LockedPtr keeps the Module mutex held alongside the borrowed pointer, pinning the binding for the lifetime of the handle.

2. They serialize access to the pointee itself. This is not new, the classes in question were already relying on the Module mutex for synchronization.

Migrate the four call sites in Module where the existing patter maps to a single LockedPtr.

The legacy raw-pointer getters remain so call sites can migrate incrementally.

### DIFF
--- a/lldb/source/Core/Module.cpp
+++ b/lldb/source/Core/Module.cpp
@@ -1231,25 +1231,21 @@ SectionList *Module::GetSectionList() {
 }
 
 LockedPtr<ObjectFile> Module::GetObjectFileLocked() {
-  std::unique_lock<std::recursive_mutex> guard(m_mutex);
-  return LockedPtr<ObjectFile>(std::move(guard), GetObjectFile());
+  return LockedPtr<ObjectFile>(m_mutex, GetObjectFile());
 }
 
 LockedPtr<SectionList> Module::GetSectionListLocked() {
-  std::unique_lock<std::recursive_mutex> guard(m_mutex);
-  return LockedPtr<SectionList>(std::move(guard), GetSectionList());
+  return LockedPtr<SectionList>(m_mutex, GetSectionList());
 }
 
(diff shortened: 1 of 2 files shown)


lldb/source/Core/Module.cpp
```diff
@@ -1234,6 +1230,28 @@ SectionList *Module::GetSectionList() {
   return m_sections_up.get();
 }
 
+LockedPtr<ObjectFile> Module::GetObjectFileLocked() {
+  std::unique_lock<std::recursive_mutex> guard(m_mutex);
+  return LockedPtr<ObjectFile>(std::move(guard), GetObjectFile());
```
Teemperor: Maybe I'm missing something here, but these functions can just do:
```
LockedPtr<ObjectFile> Module::GetObjectFileLocked() {
  return LockedPtr<ObjectFile>(m_mutex, GetObjectFile());
}
```

Unless you really want to have unique_lock here to indicate this is an exclusive lock, but I think the LockedPtr type already makes that clear.

Teemperor (approved):
One small nit but otherwise this LGTR

augusto2112 (approved):

lldb/source/Core/Module.cpp
JDevlieghere: Thanks. This is a remnant of my original PR where I stored the lock rather than the mutex. You're right that this is no longer necessary.

PR #199161, opened 2026-05-22
[flang] Enabled pulling of rebox into array_coor.

vzakhari:
This patch enables pulling slicing `fir.rebox` operations
into `fir.array_coor`. This helps preserve information about
the original rank of the array being accessed.
`FIRToMemRef` and later passes may benefit from this.

Assisted by: Claude

### DIFF
--- a/flang/lib/Optimizer/Dialect/FIROps.cpp
+++ b/flang/lib/Optimizer/Dialect/FIROps.cpp
@@ -13,6 +13,7 @@
 #include "flang/Optimizer/Dialect/FIROps.h"
 #include "flang/Optimizer/Dialect/CUDAKernelOpInterface.h"
 #include "flang/Optimizer/Dialect/FIRAttr.h"
+#include "flang/Optimizer/Dialect/FIRBoxUtils.h"
 #include "flang/Optimizer/Dialect/FIRDialect.h"
 #include "flang/Optimizer/Dialect/FIROpsSupport.h"
 #include "flang/Optimizer/Dialect/FIRType.h"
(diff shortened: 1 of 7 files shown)


flang/lib/Optimizer/Dialect/FIROps.cpp
```diff
@@ -694,8 +695,18 @@ struct SimplifyArrayCoorOp : public mlir::OpRewritePattern<fir::ArrayCoorOp> {
       // ranks differ is out of scope.
       if (op.getSlice())
         return mlir::failure();
-      if (!boxedShape)
-        return mlir::failure();
+      if (!boxedShape) {
+        // fir.rebox with a rank-reducing slice often has no shape operand.
+        // Synthesize fir.shape from box_dims extents. Do not synthesize
+        // fir.shift: rebox descriptors already have (lb-1)*stride in base_addr
+        // and FIRToMemRef must not subtract lb again for rebox memrefs.
```
jeanPerier: > Do not synthesize fir.shift: rebox descriptors already have (lb-1)*stride in base_addr and FIRToMemRef must not subtract lb again for rebox memrefs.

I do not think that is an accurate comment of why no fir.shift must be generated. The lower bounds are not "already" applied in the base_addr, it is just that the output of a fir.rebox with not shift/shape_shift has not lower bounds (fir.rebox does not propagates the lower bounds from the input fir.box), so none should be extracted from the fir.box and added to the fir.array_coor.

jeanPerier (approved):
nit about a comment, LGTM otherwise.

flang/lib/Optimizer/Dialect/FIROps.cpp
vzakhari: Thank you for the review, Jean! I fixed the comment.

PR #199163, opened 2026-05-22
[InstCombine] Fold smax(smax(x<<C1, x+C2), C3) -> smax(x<<C1, C3)

FathimaHaris:
Fixes #152788

Folds a nested smax where the inner operand   x+C2 is always dominated  by either  x<<C1 or the 
outer constant bound C3,  making it redundant:
        smax(smax(x<<C1, x+C2), C3) --> smax(x<<C1, C3)

when both  x<<C1 and  x+C2  carry  nsw  and C2 <= (2^C1 - 1) / 2^C1 * C3.

Alive2 generic proof :  https://alive2.llvm.org/ce/z/c92WEM

### DIFF
--- a/llvm/lib/Transforms/InstCombine/InstCombineCalls.cpp
+++ b/llvm/lib/Transforms/InstCombine/InstCombineCalls.cpp
@@ -2352,6 +2352,32 @@ Instruction *InstCombinerImpl::visitCallInst(CallInst &CI) {
     if (Instruction *NewMinMax = factorizeMinMaxTree(II))
        return NewMinMax;
 
+    // smax(smax(x << C0, x + C1), C2) --> smax(x << C0, C2)
+    // when C1 <=s (C2 * (2^C0 - 1) / 2^C0)
+    if (IID == Intrinsic::smax) {
+      Value *X, *Shl;
+      const APInt *C0, *C1, *C2;
+      if (match(I0, m_OneUse(m_c_SMax(m_Value(Shl),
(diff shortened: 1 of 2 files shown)


FathimaHaris:
Gentle ping for review.

FathimaHaris:
Gentle Ping . Kindly requesting a review when you have  a moment . Thanks .

dtcxzyw (commented):
I don't have any idea how to generalize this pattern... It looks like we are trying to prove `X << C1` is always greater than `X + C2` if both of them are greater than `C3`.

FathimaHaris:
The intuition is that X + C2 can be removed only if it can never determine the result of :
smax(smax(x << C1, x + C2), C3).

The only interesting case is when X + C2 > X<<C1 , since otherwise the shift already dominates.

Assuming non-wrapping arithmetic,

	X+C2 > X * 2^C1 ==> X < C2 / (2^C1 - 1)
	
Within this region, X + C2 is maximized at the boundary X = C2 / (2^C1 - 1)

Substituting this into X + C2 gives:

	C2 / (2^C1 - 1) + C2 ==> C2 * 2^C1 / (2^C1 - 1)

Therefore, if

C2 * 2^C1 / (2^C1 - 1) <= C3 ,

or equivalently C2 <= C3 * (2^C1 - 1) / 2^C1,

then even in the worst case where X + C2 exceeds X << C1, it is still bounded by C3.

Hence under this condition , X + C2 can never affect the final result and may be removed.

FathimaHaris:
Gentle Ping.

PR #199166, opened 2026-05-22
Revert "[Driver][MSVC] Use LLD if DWARF is requested"

petrhosek:
Reverts llvm/llvm-project#198600 because the test doesn't correctly handle the case when Clang is configured `CLANG_DEFAULT_LINKER=lld`.


petrhosek:
I did a fix forward instead in https://github.com/llvm/llvm-project/pull/199167.

PR #199167, opened 2026-05-22
[Driver][MSVC] Correctly handle the -fuse-ld=(empty) case

petrhosek:
We need to distinguish the case where `-fuse-ld` is unspecified and when `-fuse-ld=(empty)` as is already done in the generic `ToolChain` implementation but wasn't done in the MSVC driver to correctly handle the `CLANG_DEFAULT_LINKER` CMake option.

### DIFF
--- a/clang/test/Driver/msvc-link.c
+++ b/clang/test/Driver/msvc-link.c
@@ -52,9 +52,11 @@
 // RUN:        FileCheck --check-prefix=HYBRID-WARN %s
 // HYBRID-WARN: warning: argument unused during compilation: '-marm64x' [-Wunused-command-line-argument]
 
+// RUN: %clang --target=i686-pc-windows-msvc -g -fuse-ld= -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LINK %s
 // RUN: %clang --target=i686-pc-windows-msvc -g -fuse-ld=link -### %s 2>&1 | FileCheck --check-prefix=DEBUG-LINK %s
 // DEBUG-LINK: link.exe"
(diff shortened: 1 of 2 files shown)


petrhosek:
This is a fix forward for an issue introduced in https://github.com/llvm/llvm-project/pull/198600.

clang/test/Driver/msvc-link.c
Nerixyz, line 63: This test should also be removed then.

Nerixyz (commented):
Sorry, i missed that the default value would be non-empty in this case, so the fallback wouldn't be used.

clang/test/Driver/msvc-link.c
petrhosek: I'm not sure if these test cases provide any value.

Nerixyz: Yes - or more general if `CLANG_DEFAULT_LINKER` has any other non-empty value.

It's a bit unfortunate that we can't test this.

petrhosek:
After looking closer at the MSVC driver implementation, I don't think this is the right solution, I think the issue is the handling of the `-fuse-ld=` (empty value) case in the MSVC driver. Given that, it might be better to revert https://github.com/llvm/llvm-project/pull/198600 for now until we can come up with a more robust implementation.

petrhosek:
@Nerixyz I have reworked the implementation to correctly handle the `-fuse-ld=(empty)` case, can you re-review the change and approve if it looks good to you?

Nerixyz (approved):
Thank you

PR #199169, opened 2026-05-22
[clang-format] Add a dependent build target to check-clang-format

owenca:
This makes check-clang-format automatically builds clang-format-check-format, which checks that the new clang-format doesn't break the existing format of the clang-format source.


HazardyKnusperkeks (approved):

Meinersbur:
This broke the LLVM build when using LLVM_ENABLE_IDE. I fixed it in #199638.

In any case, I don't think this is a good approch. Attaching the check to `check-clang-format` only will not run this check with `check-clang` (or `check`). So those commands don't check everything anymore. In particular, I think neither the premerge-CI nor the buildbots run `check-clang-format`, just `check-clang`.

PR #199170, opened 2026-05-22
[Verifier] Add missing null-check.

jlebar:
[Verifier] Add missing null-check.

visitProfMetadata's was using the result of dyn_extract without first
checkout that it's non-null.  Thus one could crash the verifier by
providing invalid IR.

This bug was found by a large run of Opus 4.7 looking for bugs in LLVM.


boomanaiden154 (approved):

jlebar:
Thank you for the review!

PR #196330, opened 2026-05-07
Don't use libcalls for i128<->fp conversion on 32-bit targets

nikic:
The libcalls are only available on 64-bit.

I wonder whether I should be dropping some of the test coverage affected here. The generated implementations are very large, and I'm not sure all of these tests really provide value. It's not like anyone will be able to determine correctness based on the generated assembly.

Fixes https://github.com/llvm/llvm-project/issues/196107.

PR #199174, opened 2026-05-22
[ExpandIRInsts] Support llvm.fpto{u,s}i.sat

jlebar:
Previously, running ExpandIRInsts on a program which needs to expand a
vector fptoui.sat would hit llvm_unreachable, because the `scalarize`
function didn't handle this intrinsic.

This bug was found by a large run of Opus 4.7 looking for bugs in LLVM.


llvm/lib/CodeGen/ExpandIRInsts.cpp
```diff
@@ -1133,6 +1133,11 @@ static void scalarize(Instruction *I,
     else if (auto *CastI = dyn_cast<CastInst>(I))
       NewOp = Builder.CreateCast(CastI->getOpcode(), Ext,
                                  I->getType()->getScalarType());
+    else if (auto *II = dyn_cast<IntrinsicInst>(I);
+             II && (II->getIntrinsicID() == Intrinsic::fptoui_sat ||
+                    II->getIntrinsicID() == Intrinsic::fptosi_sat))
```
nikic: Should assert this instead.

nikic (approved):
LGTM. This is already part of https://github.com/llvm/llvm-project/pull/196330 but fine to fix separately.

llvm/lib/CodeGen/ExpandIRInsts.cpp
jlebar: Done, thanks!

boomanaiden154:
I'll take care of the profcheck failure. I thought we had mostly fixed ExpandIRInsts by now, but apparently not.

jlebar:
> I'll take care of the profcheck failure.

Thank you!

PR #191246, opened 2026-04-09
[AMDGPU] A Vulkan-style memory model weaker than the LLVM model

ssahasra:
Add a new AMDGPU memory model specification that is weaker than the LLVM memory model using Vulkan-style availability/visibility semantics and scoped operations. The model allows more efficient implementations while maintaining a safe-by-default mapping to the standard LLVM model.

**RFC:** https://discourse.llvm.org/t/a-vulkan-style-memory-model-for-amdgpu-and-beyond/90498

Related changes:

- #191390
- #199489

PR #172090, opened 2025-12-12
AMDGPU: Add builtin/intrinsic global_(load|store)_b128

macurtis-amd:
Add clang builtins and associated llvm intrinsics for scoped load/store of 128bits

New builtins:
1. `__builtin_amdgcn_global_load_b128` ([documentation](https://github.com/macurtis-amd/llvm-project/blob/global-load-store-b128/clang/docs/LanguageExtensions.rst#__builtin_amdgcn_global_load_b128-and-__builtin_amdgcn_global_store_b128), [test/examples](https://github.com/macurtis-amd/llvm-project/blob/global-load-store-b128/clang/test/CodeGenOpenCL/builtins-amdgcn-global-load-store.cl))
2. `__builtin_amdgcn_global_store_b128` ([documentation](https://github.com/macurtis-amd/llvm-project/blob/global-load-store-b128/clang/docs/LanguageExtensions.rst#__builtin_amdgcn_global_load_b128-and-__builtin_amdgcn_global_store_b128), [test/examples](https://github.com/macurtis-amd/llvm-project/blob/global-load-store-b128/clang/test/CodeGenOpenCL/builtins-amdgcn-global-load-store.cl))

And corresponding intrinsics:
1. `llvm.amdgcn.global.load.b128` ([documentation](https://github.com/macurtis-amd/llvm-project/blob/global-load-store-b128/llvm/docs/AMDGPUUsage.rst) - search for intrinsic name, [test/examples](https://github.com/macurtis-amd/llvm-project/blob/global-load-store-b128/llvm/test/CodeGen/AMDGPU/llvm.amdgcn.global.load.b128.ll) )
2. `llvm.amdgcn.global.store.b128` ([documentation](https://github.com/macurtis-amd/llvm-project/blob/global-load-store-b128/llvm/docs/AMDGPUUsage.rst) - search for intrinsic name, [test/examples](https://github.com/macurtis-amd/llvm-project/blob/global-load-store-b128/llvm/test/CodeGen/AMDGPU/llvm.amdgcn.global.load.b128.ll) )

These will initially be used by [RCCL](https://github.com/ROCm/rccl) to address some low-level performance issues.

PR #199176, opened 2026-05-22
[AMDGPU][Clang] add __builtin_amdgcn_av_(load|store)_b128

ssahasra:
These builtins allow the program to request store-available and load-visible accesses as described in #191246. Each of them takes a __MEMORY_SCOPE_* operand that is then translated to target-specific cache policy bits.

This patch was extracted from #172090.

Co-authored-by: macurtis-amd <macurtis@amd.com>
Assisted-by: Claude Opus 4.6


clang/docs/LanguageExtensions.rst
```diff
@@ -5754,6 +5756,32 @@ returns the bit at the position of the current lane. It is almost equivalent to
 ``(mask & (1 << lane_id)) != 0``, except that its behavior is only defined if
 the given mask has the same value for all active lanes of the current wave.
 
+
+__builtin_amdgcn_av_{load,store}_b128
```
shiltian, line 5762: AMDGPU's builtins need to be documented in `clang/include/clang/Basic/BuiltinsAMDGPUDocs.td` at least.

https://llvm.org/docs/AMDGPU/DeveloperGuideline.html#document-new-builtins

clang/test/SemaOpenCL/builtins-amdgcn-global-load-store-target-error.cl
```diff
@@ -0,0 +1,26 @@
...
+
+// RUN: %clang_cc1 -cl-std=CL2.0 -triple amdgcn-unknown-unknown -target-cpu gfx602 -DTEST_STORE -S -verify -o - %s
+// RUN: %clang_cc1 -cl-std=CL2.0 -triple amdgcn-unknown-unknown -target-cpu gfx705 -DTEST_STORE -S -verify -o - %s
+// RUN: %clang_cc1 -cl-std=CL2.0 -triple amdgcn-unknown-unknown -target-cpu gfx810 -DTEST_STORE -S -verify -o - %s
+// REQUIRES: amdgpu-registered-target
+
+typedef __attribute__((__vector_size__(4 * sizeof(unsigned int)))) unsigned int v4u32;
+typedef v4u32 __global *global_ptr_to_v4u32;
+
+#ifdef TEST_LOAD
```
shiltian: Do we really need to use macro here? There are definitely some test cases that emit errors from different functions but w/o using macro.

clang/test/SemaOpenCL/builtins-amdgcn-global-load-store-error.cl
```diff
@@ -0,0 +1,22 @@
+// RUN: %clang_cc1 -cl-std=CL2.0 -triple amdgcn-unknown-unknown -target-cpu gfx950         -S -verify -o - %s
```
jmmartinez, line 1: Can you add a test in `SemaHIP` where we test that the proper usage of these builtins validate for host and device?

Something like:

```cpp
// RUN: %clang_cc1 -fsyntax-only -triple amdgcn -target-cpu gfx90a -verify %s -fcuda-is-device
// RUN: %clang_cc1 -fsyntax-only -triple x86_64 -aux-triple amdgcn -verify %s
```

In the past we had some builtins that failed when building for the host (even though the calls are on the device side).

clang/include/clang/Basic/BuiltinsAMDGPU.td
```diff
@@ -280,6 +280,19 @@ def __builtin_amdgcn_raw_ptr_buffer_load_async_lds : AMDGPUBuiltin<"void(__amdgp
...
+// Global Available/Visible memory accesses.
+//===----------------------------------------------------------------------===//
+
+def __builtin_amdgcn_av_load_b128
+    : AMDGPUBuiltin<"_ExtVector<4, unsigned int>(_ExtVector<4, unsigned int> *, int)", [], "gfx9-insts"> {
+  let Documentation = [DocAVLoadB128];
+}
+def __builtin_amdgcn_av_store_b128
+    : AMDGPUBuiltin<"void(_ExtVector<4, unsigned int> *, _ExtVector<4, unsigned int>, int)", [], "gfx9-insts"> {
+  let Documentation = [DocAVStoreB128];
```
shiltian, line 294: We can add `let Args = ...` here to make the document easier to follow, when we mentioned the argument names in the doc string.

clang/include/clang/Basic/BuiltinsAMDGPU.td
```diff
@@ -280,6 +280,19 @@ def __builtin_amdgcn_raw_ptr_buffer_load_async_lds : AMDGPUBuiltin<"void(__amdgp
 def __builtin_amdgcn_struct_ptr_buffer_load_lds : AMDGPUBuiltin<"void(__amdgpu_buffer_rsrc_t, void address_space<3> *, _Constant unsigned int, int, int, int, _Constant int, _Constant int)", [], "vmem-to-lds-load-insts">;
 def __builtin_amdgcn_struct_ptr_buffer_load_async_lds : AMDGPUBuiltin<"void(__amdgpu_buffer_rsrc_t, void address_space<3> *, _Constant unsigned int, int, int, int, _Constant int, _Constant int)", [], "vmem-to-lds-load-insts">;
 
+//===----------------------------------------------------------------------===//
+// Global Available/Visible memory accesses.
+//===----------------------------------------------------------------------===//
+
+def __builtin_amdgcn_av_load_b128
+    : AMDGPUBuiltin<"_ExtVector<4, unsigned int>(_ExtVector<4, unsigned int> *, int)", [], "gfx9-insts"> {
```
shiltian: I'd prefer to use a dedicated feature instead of `gfxXYZ-insts`. That would be more future-proof in case a future target doesn't fully support `gfx9-insts` (for example, not a full set to qualify for that feature) but still supports this specific functionality.

shiltian (approved):

clang/docs/LanguageExtensions.rst
ssahasra: Done.

clang/include/clang/Basic/BuiltinsAMDGPU.td
ssahasra: Done.

clang/include/clang/Basic/BuiltinsAMDGPU.td
ssahasra: Thanks for catching that! The intrinsics already use `FlatGlobalInsts`, so updated the builtins to check the same. This is a reasonable proxy feature when we want to say "support cache policy bits from gfx9 onwards for global and flat". Cache policy bits existed prior to gfx9 for flat insts, but there's no real reason to extend support earlier than gfx9, which introduced those bits for global insts.

clang/test/SemaOpenCL/builtins-amdgcn-global-load-store-target-error.cl
ssahasra: Replaced that with the use of `split-file`.

PR #199177, opened 2026-05-22
[SelectionDAG] Error on FLDEXP libcall with non-int-sized exponent

jlebar:
The `ldexp` libcall takes an int exponent, but llvm.ldexp accepts any
integer width. LegalizeDAG should error if its argument isn't an int,
just like we do for other libcalls like FPOWI.

This (minor) bug was found by a large run of Opus 4.7 looking for bugs in LLVM.


jlebar:
Closing this PR. On further investigation the fix needed to move from LegalizeDAG to SelectionDAGBuilder (since LegalizeDAG runs after type legalization, which can promote a user-written i32 exponent to i64 on RV64, masking the originally-narrow IR type). At that point the fix is more invasive than the bug warrants — the silent truncation matches C `ldexp`'s int parameter and is unlikely to bite real-world code (clang never generates `llvm.ldexp.fty.i64`).

PR #199182, opened 2026-05-22
[X86][APX] Extend original LI to the same range as DstReg

phoebewang:
The #189222 folds NDD+Load to non-NDD when NDD memory variant not preferred. However, this will changes DstReg from regular def to early-clobber def, which causes "corrupted sub-interval" in reMaterializeFor, because the OrigLI is not updated at the same time.

Fixes: https://godbolt.org/z/7n8ozz1EG


phoebewang:
Ping

phoebewang:
Ping

llvm/lib/CodeGen/InlineSpiller.cpp
```diff
@@ -1128,6 +1128,18 @@ foldMemoryOperand(ArrayRef<std::pair<MachineInstr *, unsigned>> Ops,
...
         Matrix->assign(LI, PhysReg);
+
+      Register OrigReg = VRM.getOriginal(CopyDstReg);
+      if (OrigReg != CopyDstReg && LIS.hasInterval(OrigReg)) {
+        // Extend the original LI to cover the same range so that the
+        // sub-interval invariant holds: the original must be live wherever
+        // any of its children are live.  Without this, reMaterializeFor()
+        // can query OrigLI at an early-clobber slot that falls inside
+        // [CopyIdx, FoldIdx) and get a null VNI, triggering an assertion.
+        LiveInterval &LI = LIS.getInterval(OrigReg);
```
fzou1: Better to rename LI as there is outer LI in line 1109.

llvm/lib/CodeGen/InlineSpiller.cpp
```diff
@@ -1128,6 +1128,18 @@ foldMemoryOperand(ArrayRef<std::pair<MachineInstr *, unsigned>> Ops,
...
+        if (VNInfo *VNI = LI.getVNInfoAt(FoldIdx.getRegSlot()))
```
fzou1: Same as for VNI.

fzou1 (approved):
LGTM with two nits

llvm/lib/CodeGen/InlineSpiller.cpp
```diff
@@ -1128,6 +1128,18 @@ foldMemoryOperand(ArrayRef<std::pair<MachineInstr *, unsigned>> Ops,
 
       if (NeedMatrixReassign)
         Matrix->assign(LI, PhysReg);
+
+      Register OrigReg = VRM.getOriginal(CopyDstReg);
+      if (OrigReg != CopyDstReg && LIS.hasInterval(OrigReg)) {
```
arsenm: can !hasInterval really happen here?

phoebewang: Not so far, turn into assert instead.

llvm/lib/CodeGen/InlineSpiller.cpp
phoebewang: Done.

llvm/lib/CodeGen/InlineSpiller.cpp
phoebewang: Done.

PR #197313, opened 2026-05-12
[libc++] Implement P3508R0: Wording for "constexpr for specialized memory algorithms"

smallp-o-p:
Resolves #118379

- Add `constexpr` to affected areas
- Add `constexpr` tests
- Update documentation

Tests were created with AI assistance.

PR #199183, opened 2026-05-22
[libc++][test] Make `copy_move_types.h` usable in old modes

frederick-vs-ja:
This PR makes types in `copy_move_types.h` usable in C++03/11 modes. Because it is discovered that some types in `copy_move_types.h` are useful for testing uninitialized memory algorithms in pre-C++20 modes.


frederick-vs-ja:
See https://github.com/llvm/llvm-project/pull/197313#discussion_r3287153276 for the background.

frederick-vs-ja:
Merging as CI failure was unrelated.

PR #199184, opened 2026-05-22
[clang][analyzer] Make CallAndMessage:ArgPointeeInitializedness released (NFC)

balazske:
The option was in `InAlpha` state but should be `Released` instead. It was improved in changes #164600 and #173854.


steakhal (approved):
Yep, we already have the docs - I've checked.

PR #199185, opened 2026-05-22
[libc++][test] Add backported `destroy({_at,_n})` for tests

frederick-vs-ja:
It would be convenient to use `destroy_at`, `destroy`, and `destroy_n` in tests for pre-C++17 uninitialized memory algorithms. So this PR add backported versions of them for tests.

### DIFF
--- a/libcxx/test/std/utilities/memory/specialized.algorithms/destroy.h
+++ b/libcxx/test/std/utilities/memory/specialized.algorithms/destroy.h
@@ -1,0 +1,47 @@
+// -*- C++ -*-
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LIBCPP_TEST_STD_UTILITIES_MEMORY_SPECIALIZED_ALGORITHMS_DESTROY_H
+#define LIBCPP_TEST_STD_UTILITIES_MEMORY_SPECIALIZED_ALGORITHMS_DESTROY_H
+
+#include <cstddef>
+#include <memory>
+#include <type_traits>
+
+#include "test_macros.h"
+
+namespace backport {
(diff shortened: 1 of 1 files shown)


frederick-vs-ja:
See https://github.com/llvm/llvm-project/pull/197313#discussion_r3287153276 for the background.

frederick-vs-ja:
Merging as CI failure was unrelated.

libcxx/test/std/utilities/memory/specialized.algorithms/destroy.h
```diff
@@ -0,0 +1,47 @@
...
+#ifndef LIBCPP_TEST_STD_UTILITIES_MEMORY_SPECIALIZED_ALGORITHMS_DESTROY_H
+#define LIBCPP_TEST_STD_UTILITIES_MEMORY_SPECIALIZED_ALGORITHMS_DESTROY_H
+
+#include <cstddef>
+#include <memory>
+#include <type_traits>
+
+#include "test_macros.h"
+
+namespace backport {
```
philnik777, line 19: Why are we introducing a new namespace here?

frederick-vs-ja: I think a namespace encourages users to use qualified function calls, which is consistent with our conventional use of `std::` in test files. Namespace `test` might be better, but `test` is often used as a function name.

philnik777: I mean why a new namespace instead of one we already use, like `support`, `utils` or `util`? I'd really like to condense all these into a single one, not increase the number.

frederick-vs-ja: Ah, just because I forgot looking files in `libcxx/test/support` directory. Also, namespace `support` etc. perhaps make more sense if the header is in `libcxx/test/support`. What do you think about move this header there?

philnik777: Sure, I wouldn't be against moving it.

frederick-vs-ja: Opened #199431.

PR #199192, opened 2026-05-22
[clangd] Fixed signature help for static call operator (C++23)

### LINKED ISSUE #196153: [clangd] [Signature help] [C++23] Parameter hints do not appear for static operator (and some obscure interaction with templates)
In the following example code I do not get any parameter hints for the operator, but if I manually type in `true` clangd recognizes that I am calling the operator.

<details>

```
struct Abc {
    static void operator()(bool a) {}
};

void test() {
    Abc abc;
    abc(^);
}
```

</details>

Image

If I remove the `static` from the operator definition I do get the paramtere hints.

<details>

```
struct Abc {
    void operator()(bool a) {}
};

void test() {
    Abc abc;
    abc(^);
}
```

</details>

Image

As straight forward as this bug seems to be, I stumbled upon some more odd behaviour. Consider now me using 2 parameters, this still does not work with static as you can see here.

<details>

```
struct Abc {
    static void operator()(bool a, bool b) {}
};

void test() {
    Abc abc;
    abc(^);
}
```

</details>

Image

But if I now template my operator it does work.

<details>

```
struct Abc {
    template <typename T>
    static void operator()(T a, bool b) {}
};

void test() {
    Abc abc;
    abc(^);
}
```

</details>

Image

The odd part here is that I **need** 2 parameters for my operator, because the templated version with only 1 parameter does **not** work.

<details>

```
struct Abc {
    template <typename T>
    static void operator()(T a) {}
};

void test() {
    Abc abc;
    abc(^);
}
```

</details>

Image

Also it is important that the template type is used in the parameters, otherwise it does **not** work.

<details>

```
struct Abc {
    template <typename T>
    static void operator()(bool a, bool b) { T c; }
};

void test() {
    Abc abc;
    abc(^);
}
```

</details>

Image

Ippo47:
This is a continuation of #197139 which was accidentally closed due to branch deletion. All reviewer feedback from that PR has been addressed.


cor3ntin:
You can reopen then push again to a branch of the same name

timon-ul (changes requested):
Well for next time just reopen the old one, whatever.

So overall this looks pretty good, everything seems to work as it should, which is nice. That being said I am still left a bit confused as to why the original code was not able to handle this. The call to `AddFunctionCandidates` looks to me like it is already doing exactly the same as this PR, iterating over all `NamedDecl`, looking at the underlying declaration, casting it to a `FunctionDecl` or a `FunctionTemplateDecl` and then calling the proper `Add[Template]OverloadCandidate`. The only things that are different is
1. a different `if` [branch](https://searchfox.org/llvm/source/clang/lib/Sema/SemaOverload.cpp#7837), but this should not be entered if it is `static` so I doubt this is the problem
2. some kind of argument slicing happening [here](https://searchfox.org/llvm/source/clang/lib/Sema/SemaOverload.cpp#7871) which seems more suspicious to me.

To me it seems like it makes more sense to adapt the code there and handle the case properly there instead of circumventing it altogether. So if it really is the slicing you should try to understand why and when the slicing is important and when we do not want it.

Ippo47:
I looked at the code and i think the slicing only happens when FirstArgumentIsBase is true. When called from signature help, FirstArgumentIsBase is false so the object argument never gets removed for static candidates, the fix is also doing similar thing just externally by intercepting static candidates before they reached AddFunctionCandidates. But you are correct i will make changes in that portion ans revert my chnages.

Ippo47:
@timon-ul hi, can you please review the code, i have made the necessary changes

clang/lib/Sema/SemaOverload.cpp
```diff
@@ -7869,9 +7869,10 @@ void Sema::AddFunctionCandidates(const UnresolvedSetImpl &Fns,
       // Slice the first argument (which is the base) when we access
       // static method as non-static.
       if (Args.size() > 0 &&
-          (!Args[0] || (FirstArgumentIsBase && isa<CXXMethodDecl>(FD) &&
-                        !isa<CXXConstructorDecl>(FD)))) {
-        assert(cast<CXXMethodDecl>(FD)->isStatic());
+          (!Args[0] ||
+           (isa<CXXMethodDecl>(FD) && !isa<CXXConstructorDecl>(FD) &&
+            (FirstArgumentIsBase || cast<CXXMethodDecl>(FD)->isStatic())))) {
+        assert(FirstArgumentIsBase || cast<CXXMethodDecl>(FD)->isStatic());
```
timon-ul: This assert is pretty pointless, you can only enter this branch if either `FirstArgumentIsBase || cast<CXXMethodDecl>(FD)->isStatic()` so of course this will hold. As far as I can tell the old assert logic was that if `FirstArgumentIsBase == true` it had to be a static function, but without explicitly testing for it this assert made sure that it is the case. Now with your new assert this could not be true and slip by this assert now.

clang/lib/Sema/SemaOverload.cpp
```diff
@@ -7869,9 +7869,10 @@ void Sema::AddFunctionCandidates(const UnresolvedSetImpl &Fns,
       // Slice the first argument (which is the base) when we access
       // static method as non-static.
       if (Args.size() > 0 &&
-          (!Args[0] || (FirstArgumentIsBase && isa<CXXMethodDecl>(FD) &&
-                        !isa<CXXConstructorDecl>(FD)))) {
-        assert(cast<CXXMethodDecl>(FD)->isStatic());
+          (!Args[0] ||
+           (isa<CXXMethodDecl>(FD) && !isa<CXXConstructorDecl>(FD) &&
+            (FirstArgumentIsBase || cast<CXXMethodDecl>(FD)->isStatic())))) {
```
timon-ul: I think if we can we do not want to let all static methods through here, but only for FirstArgumentIsBase or it is the function operator. For this I think the best option is to just do `FD->getOverloadedOperator() == OO_Call`.

timon-ul (changes requested):
I am a bit confused as to why we now have code changes in both places, `SemaCodeComplete.cpp` and `SemaOverload.cpp`, just the ones in `SemaOverload.cpp` should be enough.

zwuis:
It's up to you, but we can add tests to clang/test/CodeCompletion.

---

Could you please add tests for `static operator[]`, regardless of whether it is handled correctly?

clang/lib/Sema/SemaOverload.cpp
```diff
@@ -7867,11 +7867,13 @@ void Sema::AddFunctionCandidates(const UnresolvedSetImpl &Fns,
       // This branch handles both standalone functions and static methods.
 
       // Slice the first argument (which is the base) when we access
-      // static method as non-static.
+      // static operator() as non-static.
```
timon-ul, line 7870: I think I confused you now, sorry. To my understanding the original comment fits still to the new behaviour, basically whenever we access a static method as non-static we want to argument slice. Before this was checked by `FirstArgumentIsBase` and the idea was that this should hold true for these cases (quite frankly it is a bit messy, I think `FirstArgumentIsBase` is set whenever we have a member expr, no matter if it is static or not, but we only check for it in the branch where we know it to be static, which feels pointless, because we could have just asked if it was a static member expr instead? Not too sure on the logic here....). The problem we face is that operators are handled elsewhere and at that point we do not set `FirstArgumentIsBase` (and I think it also makes no sense for us to set it since some operator overloads might be non static? Unless they are handled individually, I actually do not know, you could test this out too I guess). This leaves us to having alternatively check for it being a static operator in each loop iteration. What this means though is that our changes now actually made this comment true (again).

clang/lib/Sema/SemaOverload.cpp
```diff
@@ -7867,11 +7867,13 @@ void Sema::AddFunctionCandidates(const UnresolvedSetImpl &Fns,
...
       if (Args.size() > 0 &&
-          (!Args[0] || (FirstArgumentIsBase && isa<CXXMethodDecl>(FD) &&
-                        !isa<CXXConstructorDecl>(FD)))) {
-        assert(cast<CXXMethodDecl>(FD)->isStatic());
```
timon-ul, line 7874: I think you also misunderstood me here, for the case of `FirstArgumentIsBase == true` we do nopt check for `static`, but we still expect it to be the case so the assert still makes sense to keep (in its original state).

timon-ul (changes requested):

timon-ul:
> It's up to you, but we can add tests to clang/test/CodeCompletion.

Probably a good call, I was also looking at that, but given we currently have all tests for this in `clangd` and also I am not even sure if anyone besides clangd is using it? Not sure why there was never a test for it there.

> Could you please add tests for `static operator[]`, regardless of whether it is handled correctly?

I also thought about this operator, but quite frankly what would this test look like? As far as I can tell current behaviour is that there is no signature help for even non static `operator[]`. I guess we could exactly test for there being none, so if we ever implement signature help for this we would get a test failure if we forget about this case, but given we do not have signature help for this to this day I am just wondering if this is something people would even care about. Please share your thoughts tho @zwuis .

zwuis:
> It's up to you, but we can add tests to clang/test/CodeCompletion.
> 
> ---
> 
> Could you please add tests for `static operator[]`, regardless of whether it is handled correctly?

Wait, I mixed up signature help and code completion. 🤦‍♂️ Never mind.

zwuis:
Could you please add tests with lambdas with `static` specifier?

```cpp
   []() static {};
//      ^^^^^^
```

Ippo47:
> Could you please add tests with lambdas with `static` specifier?
> 
> ```c++
>    []() static {};
> //      ^^^^^^
> ```

i will add it

PR #196924, opened 2026-05-11
[mlir][mem2reg] Promote memory slots through transparent view operations

jeanPerier:
This patch enables mem2reg to operate on load/store that are made through cast/views of the memory slot (not directly using the SSA value produced by the allocation op).

This is done by adding two APIs to `PromotableOpInterface` that must be implemented by operation that defines view of the slot and through which mem2reg should still happen. The first API is used to return a new slot (the "slot view") that will be used when visiting load/stores operation on its results. The second API allows inserting value casts (which will be bitcasts for instance in for the fir.convert use case) that the mem2reg framework will generate around load/stores so that the load/store APIs are provided with reaching definitions consistent with the element type of the slot view they are operating on.

Assisted by: Claude

RFC: https://discourse.llvm.org/t/mlir-mem2reg-rfc-promote-memory-slots-through-transparent-view-operations/90761

PR #199193, opened 2026-05-22
[mlir][mem2reg] fix assert for indirect blocking uses inside regions

jeanPerier:
When adding new blocking uses created by the interface of a previous blocking uses (typically forwarding  the blocking uses to the op result users), the mem2reg framework was assuming that the new blocking uses are in the same region as the original blocking use, which is not true in general and lead to the assert:

`Transforms/Mem2Reg.cpp:743: void {anonymous}::MemorySlotPromoter::removeBlockingUses(mlir::Region*): Assertion `op->getParentRegion() == region && "all operations must still be in the same region"' failed.`

This patch fixes this by adding the new uses into the userToBlockingUses for the region of the new blocking uses.

Note: this fix is extracted from https://github.com/llvm/llvm-project/pull/196924 for clarity as it is orthogonal.


gysit (approved):
LGTM

Moxinilian (approved):
Thanks!

PR #199198, opened 2026-05-22
Reland "[PowerPC] set libcall lowering for fp setcc ops on SPE boards"

Varnike:
This is a reland of 4d0100789dc9b4db7d77c033d8f53f8b7dc68437, which was reverted by c24ab4c814f680c5ac71600bf80d551123abd84d.

The functionality is unchanged from the original patch. This version only fixes the fast-math flag propagation issue by passing `SDNodeFlags` explicitly to `DAG.getSetCC()` (same as #199105 which was closed because the underlying patch had been reverted).


thurstond:
Drive-by comment: it's usually easier to review reland patches if the pull request contains separate commits for the original patch and the fix.

Varnike:
Done, thanks for the advice

Varnike:
ping

arsenm (approved):

PR #195485, opened 2026-05-02
[VPlan] Add type in VPRecipeValue, migrate first set of recipes. (NFC)

fhahn:
This patch adds a new Type field to hold the scalar type for the
VPRecipeValue. It starts the migration of an initial set of recipes to
set the scalar type and use it directly in VPTypeAnalysis, removing
special handling for those types.

ventually, VPTypeAnalysis can be retired when the migration is
complete, hopefully eliminating the need for type related caching, which
can lead to subtle invalidation errors.

To help the transition, a temporary getScalarTypeOrInfer has been added,
which returns the scalar type set in the VPValue for already migrated
recipes or falls back o type inferrence otherwise.

VPlanVerifier has been extended verify the set scalar type where
possible.

Tracking the type directly means a few places that change the type will
need to be updated to create a new replacement recipe with correct
types.

Depends on https://github.com/llvm/llvm-project/pull/195483

PR #199200, opened 2026-05-22
[VPlan] Clear VPTypeAnalysis cache in convertToStridedAccesses

lukel97:
This is an attempt to fix the crashes reported in https://github.com/llvm/llvm-project/pull/147297#issuecomment-4516349199

The VPTypeAnalysis that's reused in VPCostContext is live across a lot of various passes, including some that I'm fairly sure delete values e.g. VPlanTransforms::removeBranchOnConst. So this will run into the same issue https://github.com/llvm/llvm-project/issues/184650.

This isn't the prettiest solution and I don't think we can write a test case for this. But this should be temporary since we plan to eventually remove VPTypeAnalysis: https://github.com/llvm/llvm-project/pull/195485. It's really only to fix the crashes for now.

### DIFF
--- a/llvm/lib/Transforms/Vectorize/VPlanAnalysis.h
+++ b/llvm/lib/Transforms/Vectorize/VPlanAnalysis.h
@@ -61,6 +61,9 @@ public:
 
   /// Return the LLVMContext used by the analysis.
   LLVMContext &getContext() { return Ctx; }
+
+  /// Clears the type cache. Use if a value may have been deleted.
+  void invalidateCache() { CachedTypes.clear(); }
 };
 
 // Collect a VPlan's ephemeral recipes (those used only by an assume).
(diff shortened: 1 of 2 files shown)


Mel-Chen:
https://github.com/Mel-Chen/llvm-project/commit/f7c269256f891cfae9858fd78bb06766ed723c07
I actually have a local branch fixing this, but haven't open a PR yet. They are slightly different, but I think either works perfectly fine as a quick fix.

lukel97:
@Mel-Chen I think your approach is better, lets go with it instead

fhahn (commented):
> @Mel-Chen I think your approach is better, lets go with it instead

Sounds good, I think that was one of the alternatives suggested during he review (+ `makeCallWideningDecisions` also uses that I think)

fhahn:
But would be good to land a fix today to bring the bot back to green

Mel-Chen:
https://github.com/llvm/llvm-project/pull/199222

PR #199201, opened 2026-05-22
[bazel] Use `additional_compiler_inputs` to handle include scanning for TargetPassRegistry.inc

rupprecht:
This use of using `#include` with a macro breaks include scanning, for example:

* `GET_PASS_REGISTRY` defined here: https://github.com/llvm/llvm-project/blob/5c853423f4f9e7296b7596b7f3ccade481686bfd/llvm/lib/Target/AArch64/AArch64TargetMachine.cpp#L603
* `GET_PASS_REGISTRY` included here: https://github.com/llvm/llvm-project/blob/5c853423f4f9e7296b7596b7f3ccade481686bfd/llvm/include/llvm/Passes/TargetPassRegistry.inc#L60

When include scanning is enabled, the `PassRegistry.def` gets omitted because it the include scanner does not handle this case. Providing it via `additional_compiler_inputs` ensures it is included even in that case.

### DIFF
--- a/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel
+++ b/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel
@@ -3907,6 +3907,20 @@ gentbl_cc_library(
             allow_empty = True,
         ),
         hdrs = ["lib/Target/" + target["name"] + "/" + target["short_name"] + ".h"],
+        # The TargetMachine files use this pattern:
+        #
+        # // In FooTargetMachine.cpp
+        # #define GET_PASS_REGISTRY "FooPassRegistry.def"
+        # #include "llvm/Passes/TargetPassRegistry.inc"
+        #
+        # // In TargetPassRegistry.inc
+        # #include GET_PASS_REGISTRY
+        #
+        # This breaks when using include scanning, so include them explicitly here.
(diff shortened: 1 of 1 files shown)


keith (approved):

keith:
fwiw i heard google works around this in a different way by patching in a commented out `#include` in some cases, so that might be preferred for you? idk. this seems nicer to me, but i imagine also doesn't handle all the cases

rupprecht:
> fwiw i heard google works around this in a different way by patching in a commented out `#include` in some cases, so that might be preferred for you? idk. this seems nicer to me, but i imagine also doesn't handle all the cases

Yes, that is exactly what's motivating this :)

This is the silly thing I'd like to remove from ~10 files:
```
// include scanning hint
/*
#include "AArch64PassRegistry.def"
*/
```

I'd rather have this encoded in the build file instead of a local patch to sources, which can run into merge conflicts, and needs to be added to every `*TargetMachine.cpp`.

keith:
++, also nice if i can get anywhere with include scanning outside of google https://github.com/bazelbuild/bazel/issues/29354

rupprecht:
I'll ask around to see if I can figure out what the status of that is.

PR #199203, opened 2026-05-22
Fix Bazel build for 08124b7

akuegel:


akuegel:
I see it already got merged.

PR #199205, opened 2026-05-22
[CIR] Skip __retval for return types with deleted copy constructors

### LINKED ISSUE #198602: [CIR] Avoid illegal copies of return values
Because ABI calling convention lowering is deferred in CIR, we do not use `sret` parameters for functions that return a structure that can't be trivially copied. This might be OK except that we currently also generate a store to the `__retval` alloca and a reload from there to return the value. The following example shows this behavior:

```c++
struct S {
  S();
  S(const S&) = delete;
  ~S();
};

S foo();
S bar() {
  return foo();
}
```
That compiles to the following CIR:

```
  cir.func private @_Z3foov() -> !rec_S
  cir.func dso_local @_Z3barb() -> !rec_S {
    %0 = cir.alloca !rec_S, !cir.ptr<!rec_S>, ["__retval"] {alignment = 1 : i64}
    %1 = cir.call @_Z3foov() : () -> !rec_S
    cir.store align(1) %1, %0 : !rec_S, !cir.ptr<!rec_S>
    %2 = cir.load %0 : !cir.ptr<!rec_S>, !rec_S
    cir.return %2 : !rec_S
  }
```

We need to clean up the CIR generation to avoid such copies.

Note that this issue also concerns classes with non-trivial, but non deleted copy constructors.

While the store and reload shown in the example above will almost certainly be eliminated later in the compilation pipeline, the intent of this issue is to avoid having this semantically incorrect representation in any intermediate state.

TenHian:
Because ABI calling convention lowering is deferred in CIR, we do not
use sret parameters for functions that return an aggregate.  Instead,
CIR emits a store through the `__retval` alloca and a reload from it.
For types with a deleted copy constructor, this store+load constitutes
an illegal bitwise copy.

Approach
--------

Detect whether the return type's copy constructor is deleted (recursing
into members and bases).  When it is, skip the `__retval` alloca entirely
and forward the call result as an SSA value directly to `cir.return`.

Key changes:

- `emitAndUpdateRetAlloca`: skip `__retval` for non-copyable types
- `emitReturnStmt`: for non-copyable types, emit the call with
  `ReturnValueSlot::forNoAggregateStore()` and return the SSA value
- `emitReturnOfRValue`: when `returnValue` is absent, forward the RValue
  directly without going through a memory slot
- `ReturnValueSlot`: add `noAggregateStore` flag to prevent `emitCall`
  from storing to memory when the result will be returned directly
- `CIRGenVTables.cpp`: route aggregate returns through
  `emitReturnOfRValue` regardless of slot validity
- NRVO guard: skip NRVO allocation when `__retval` is absent

Impact
------

| Aspect                    | Description |
| ------------------------- | ----------- |
| Core fix                  | Non-copyable types no longer go through `__retval` store/load |
| Fallthrough               | Explicit trap (O0) / unreachable (O1+) when no `__retval` exists |
| Virtual dispatch          | Previously crashed, now correctly returns SSA value |
| Copyable types            | Unchanged — still use `__retval` store/load |
| Ternary / branches        | Still has redundant store/load through `agg.tmp` (not `__retval`; does not affect ABI lowering) |
| Lambda invoker / thunk    | Redundant store/load through a function-local temp when `returnValue` is absent (pre-existing from PR #197572; can be addressed in a follow-up) |

Fixes https://github.com/llvm/llvm-project/issues/198602

### DIFF
--- a/clang/lib/CIR/CodeGen/CIRGenCall.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenCall.cpp
@@ -1346,6 +1346,11 @@ RValue CIRGenFunction::emitCall(const CIRGenFunctionInfo &funcInfo,
     return getUndefRValue(retTy);
   switch (getEvaluationKind(retTy)) {
   case cir::TEK_Aggregate: {
+    if (returnValue.isNoAggregateStore()) {
+      mlir::ResultRange results = theCall->getOpResults();
+      assert(results.size() <= 1 && "multiple returns from a call");
+      return RValue::get(results[0]);
+    }
     Address destPtr = returnValue.getValue();
 
     if (!destPtr.isValid())
--- a/clang/lib/CIR/CodeGen/CIRGenCall.h
+++ b/clang/lib/CIR/CodeGen/CIRGenCall.h
@@ -259,12 +259,20 @@ public:
 /// whether the address is volatile or not.
 class ReturnValueSlot {
   Address addr = Address::invalid();
+  bool noAggregateStore = false;
 
 public:
   ReturnValueSlot() = default;
   ReturnValueSlot(Address addr) : addr(addr) {}
 
+  static ReturnValueSlot forNoAggregateStore() {
+    ReturnValueSlot slot;
+    slot.noAggregateStore = true;
+    return slot;
+  }
+
   bool isNull() const { return !addr.isValid(); }
+  bool isNoAggregateStore() const { return noAggregateStore; }
   Address getValue() const { return addr; }
 };
 
--- a/clang/lib/CIR/CodeGen/CIRGenDecl.cpp
+++ b/clang/lib/CIR/CodeGen/CIRGenDecl.cpp
@@ -94,7 +94,7 @@ CIRGenFunction::emitAutoVarAlloca(const VarDecl &d,
     // unless:
     // - it's an NRVO variable.
     // - we are compiling OpenMP and it's an OpenMP local variable.
-    if (nrvo) {
+    if (nrvo && returnValue.isValid()) {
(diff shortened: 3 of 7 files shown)


andykaylor:
Hi @TenHian. Thank you for contributing to the LLVM project!

I've looked over this PR, and while it is a reasonably-sized and focused change, it does appear to be at least partially AI-generated. Please review the LLVM AI policy linked in the automated response above.

More substantially, the PR only addresses classes with deleted copy constructors. That is indeed what is shown in the reproducer I provided for https://github.com/llvm/llvm-project/issues/198602, but I intended for the issue to also cover cases with non-trivial, but not deleted, copy constructors. Your PR does not attempt to address that case, so it does not entirely fix the issue.

If you've read the LLVM AI policy and are prepared to discuss the design and implementation choices involved in this PR. let me know and I'll be happy to continue with the review.

TenHian:
@andykaylor Thank you for the review, and for your guidance on the scope of Issue #198602. I've also seen your latest comment on the issue.

I spent time after your review expanding the check from `hasDeletedCopyConstructor` to `hasNonTrivialCopyConstructor` , and extending the fix to the lambda static invoker and vtable thunk paths (PR #197572). Those changes are on my local branch and pass the test suite.

But honestly, the end result has poor maintainability. Adding `forNoAggregateStore()` as an ad-hoc third slot state, pattern-matching `CallExpr` in `emitReturnStmt`, the NRVO fallback loading from the symbol table — can't imagine anyone would enjoy reviewing that thing. This issue touches a design decision that should be made by someone with real CIR expertise, not patched around by me.

I'd prefer to close this PR. Thank you for your time.

This PR was developed with AI assistance (Assisted-by: omp).

Issue #195400, opened 2026-05-02
[libc][math] Add missing isnan variants for float16, bfloat16, and float128


AnonMiraj:
The current implementations of `isnan` are limited to `float`, `double`, and `long double`. We should add support for the remaining types (`float16`, `bfloat16`, and `float128`).

Additionally, the existing implementations currently rely on `__builtin_isnan`. I think refactoring them to use `fputil` would be better, especially since `issignaling` already uses `fputil`.

AnonMiraj:
in #195598 
I added an fputil path if the builtin is not available/constexpr.
the only thing left is to implement
- isnanf16
- isnanbf16
- isnanf128

0bVdnt:
Hello, can I work on this issue?

AnonMiraj:
Sure feel free to make a pr for one of the three functions.

iamaayushrivastava:
Hi, I’d like to work on this issue if it’s open.

TarunAkash:
HI, can I work on this issue?

AnonMiraj:
@0bVdnt 
Which function are you working on

0bVdnt:
> @0bVdnt 
> Which function are you working on

I am working on `isnanf128`.
Is that fine?

AnonMiraj:
> I am working on `isnanf128`.

Great

@TarunAkash 
Feel free to take isnanbf16

@iamaayushrivastava 
You can work on isnanf16

TarunAkash:
> > I am working on `isnanf128`.
> 
> Great
> 
> [@TarunAkash](https://github.com/TarunAkash) Feel free to take isnanbf16
> 
> [@iamaayushrivastava](https://github.com/iamaayushrivastava) You can work on isnanf16

Yeah sure!

iamaayushrivastava:
I've started working on isnanf16. Thank you!

> > I am working on `isnanf128`.
> 
> Great
> 
> [@TarunAkash](https://github.com/TarunAkash) Feel free to take isnanbf16
> 
> [@iamaayushrivastava](https://github.com/iamaayushrivastava) You can work on isnanf16

iamaayushrivastava:
Hi @AnonMiraj, I've been working on this issue and have opened a PR implementing the the float16 variant of isnan i.e., `isnanf16`.

PR: [198115](https://github.com/llvm/llvm-project/pull/198115)

Please review and share your feedback.

Thanks!

SwarnaDharshiniS:
[...]

PR #199206, opened 2026-05-22
[libc][math] Add isnanf128

0bVdnt:
## Summary

- Add a fputil-based isnanf128 implementation and wire it into libc math entrypoints, public math header generation, and shared math wrappers.

- Add smoke test coverage for float128 NaN classification and shared math runtime/constexpr coverage.

Part of #195400


Sukumarsawant:
Do check YAML registration for the function in math.yaml

Sukumarsawant:
Also, these should be added to Bazel since we are making it header-only.

Sukumarsawant (commented):
LGTM after syncing your PR

libc/src/__support/math/isnanf128.h
```diff
@@ -0,0 +1,31 @@
...
+
+#ifdef LIBC_TYPES_HAS_FLOAT128
+
+#include "src/__support/FPUtil/FPBits.h"
+#include "src/__support/macros/config.h"
+
+namespace LIBC_NAMESPACE_DECL {
+namespace math {
+
+LIBC_INLINE LIBC_CONSTEXPR int isnanf128(float128 x) {
```
AnonMiraj, line 22:
```suggestion
LIBC_INLINE constexpr int isnanf128(float128 x) {
```
`LIBC_CONSTEXPR` is not needed here

AnonMiraj (commented):
LGTM with a small nit
\+ Have you tried building with gcc? 

CC @lntue

libc/src/__support/math/isnanf128.h
0bVdnt: Done, but it seems it causes CI failures for older gcc versions

0bVdnt:
> LGTM with a small nit + Have you tried building with gcc?
> 
> CC @lntue

Yes, I did build with a newer gcc, but it seems that for older versions of gcc, constexpr on `fputil::FPBits<float128>` isn't supported. Should I revert this back to `LIBC_CONSTEXPR`?

lntue:
> > LGTM with a small nit + Have you tried building with gcc?
> > CC @lntue
> 
> Yes, I did build with a newer gcc, but it seems that for older versions of gcc, constexpr on `fputil::FPBits<float128>` isn't supported. Should I revert this back to `LIBC_CONSTEXPR`?

Yes, you should revert it back to `LIBC_CONSTEXPR`.

lntue (approved):

lntue:
@0bVdnt do you mind syncing the PR to see if the CIs would pass?

PR #199208, opened 2026-05-22
[CostModel][X86] Update VECREDUCE MINMAX costs based off llvm-mca analysis

RKSimon:
Add full CostKinds, to improve a lot of reduction matching in vectorcombine/slp passes

These are based off SMIN/UMIN numbers, and a few SMAX/UMAX numbers don't always match, but are typically within +/-1


llvm/test/Transforms/PhaseOrdering/X86/horizontal-reduce-smax.ll
```diff
@@ -13,11 +13,15 @@
...
-; CHECK-NEXT:    ret i64 [[TMP4]]
+; SSE-LABEL: @test_reduce_v2i64(
+; SSE-NEXT:    [[TMP1:%.*]] = tail call i64 @llvm.vector.reduce.smax.v2i64(<2 x i64> [[A0:%.*]])
+; SSE-NEXT:    ret i64 [[TMP1]]
+;
+; AVX-LABEL: @test_reduce_v2i64(
+; AVX-NEXT:    [[TMP1:%.*]] = shufflevector <2 x i64> [[A0:%.*]], <2 x i64> poison, <2 x i32> <i32 1, i32 poison>
+; AVX-NEXT:    [[TMP2:%.*]] = tail call <2 x i64> @llvm.smax.v2i64(<2 x i64> [[A0]], <2 x i64> [[TMP1]])
+; AVX-NEXT:    [[TMP3:%.*]] = extractelement <2 x i64> [[TMP2]], i64 0
+; AVX-NEXT:    ret i64 [[TMP3]]
```
phoebewang: What's happening here?

llvm/test/Transforms/PhaseOrdering/X86/horizontal-reduce-smin.ll
```diff
@@ -13,11 +13,15 @@
...
-; CHECK-NEXT:    ret i64 [[TMP4]]
+; SSE-LABEL: @test_reduce_v2i64(
+; SSE-NEXT:    [[TMP1:%.*]] = tail call i64 @llvm.vector.reduce.smin.v2i64(<2 x i64> [[A0:%.*]])
+; SSE-NEXT:    ret i64 [[TMP1]]
+;
+; AVX-LABEL: @test_reduce_v2i64(
+; AVX-NEXT:    [[TMP1:%.*]] = shufflevector <2 x i64> [[A0:%.*]], <2 x i64> poison, <2 x i32> <i32 1, i32 poison>
+; AVX-NEXT:    [[TMP2:%.*]] = tail call <2 x i64> @llvm.smin.v2i64(<2 x i64> [[A0]], <2 x i64> [[TMP1]])
+; AVX-NEXT:    [[TMP3:%.*]] = extractelement <2 x i64> [[TMP2]], i64 0
+; AVX-NEXT:    ret i64 [[TMP3]]
```
phoebewang: ditto.

llvm/lib/Target/X86/X86TargetTransformInfo.cpp
```diff
@@ -5919,41 +5919,102 @@ X86TTIImpl::getMinMaxReductionCost(Intrinsic::ID IID, VectorType *ValTy,
   // We use the Intel Architecture Code Analyzer(IACA) to measure the throughput
```
phoebewang: Should this be changed to llvm-mca?

llvm/test/Transforms/PhaseOrdering/X86/horizontal-reduce-smax.ll
RKSimon: vectorcombine isn't replacing the shuffle chain with the reduction if the old/new costs match - I'll create a patch that accepts it as long as we reduce the instruction count (which is what we normally do in vectorcombine)

phoebewang (approved):
LGTM.

PR #199210, opened 2026-05-22
[Support] Support runtime override for LLVM_WINDOWS_PREFER_FORWARD_SLASH

Jwata:
Allow overriding the compile-time LLVM_WINDOWS_PREFER_FORWARD_SLASH setting at runtime using an environment variable of the same name.

This enables testing both path separator behaviors (forward slash vs. backslash on Windows) using a single build, which is useful for CI/Buildbots.

The environment variable is checked once and cached in a static variable for performance.

Also updated relevant tests in SupportTests (Path.cpp and CommandLineTest.cpp) to dynamically detect the preferred separator style at runtime instead of relying on the compile-time macro, making them compatible with the override.


llvm/lib/Support/Path.cpp
```diff
@@ -38,13 +38,29 @@ namespace {
   using llvm::sys::path::is_separator;
   using llvm::sys::path::Style;
 
+  inline bool prefer_forward_slash() {
+    static bool prefer = []() {
+      if (std::optional<std::string> Env =
+              sys::Process::GetEnv("LLVM_WINDOWS_PREFER_FORWARD_SLASH")) {
+        StringRef Ref(*Env);
+        if (Ref.empty() || Ref.equals_insensitive("0") ||
```
rnk: Return the negated Boolean, rather than the two return statements

rnk (commented):
@mstorsjo @mati865 @zmodem 
I think mingw was the main stakeholder for C:/ style paths, wdyt?
My main concern here is testability, not having to rebuild to test the functionality

mstorsjo:
> @mstorsjo @mati865 @zmodem I think mingw was the main stakeholder for C:/ style paths, wdyt? My main concern here is testability, not having to rebuild to test the functionality

No concerns from me; making it testable without rebuiling sounds good. (As mentioned somewhere in the original patch from many years ago, having it settable with a command line flag can be tricky, but this approach should be fine I think.)

mati865:
No concerns from me as well.

Jwata:
Thank you for your feedback! I think this is ready for review. PTAL

llvm/lib/Support/Path.cpp
```diff
@@ -38,13 +38,27 @@ namespace {
...
   using llvm::sys::path::Style;
 
+  inline bool prefer_forward_slash() {
+    static bool prefer = []() {
+      if (std::optional<std::string> Env =
+              sys::Process::GetEnv("LLVM_WINDOWS_PREFER_FORWARD_SLASH")) {
+        StringRef Ref(*Env);
+        return !(Ref.empty() || Ref.equals_insensitive("0") ||
+                 Ref.equals_insensitive("false") ||
+                 Ref.equals_insensitive("off") || Ref.equals_insensitive("no"));
```
zmodem: I wonder if we should reverse the logic here: default to false unless a specific value is used to turn it on? Maybe just checking for 1 would be enough?

llvm/unittests/Support/CommandLineTest.cpp
```diff
@@ -42,6 +42,14 @@ MATCHER(StringEquality, "Checks if two char* are equal as strings") {
   return std::string(std::get<0>(arg)) == std::string(std::get<1>(arg));
 }
 
+#ifdef _WIN32
+bool preferForwardSlash() {
+  SmallString<1> Path("/");
+  llvm::sys::path::native(Path);
+  return Path == "/";
+}
```
zmodem: This could be just one line I think:

```
return llvm::sys::path::native("/") == "/";
```

I wonder if we should expose `prefer_forward_slash()` from Path.cpp? Maybe not if it's just for these two tests.

llvm/lib/Support/Path.cpp
```diff
@@ -38,13 +38,23 @@ namespace {
   using llvm::sys::path::is_separator;
   using llvm::sys::path::Style;
 
+  inline bool prefer_forward_slash() {
+    static bool prefer = []() {
+      if (std::optional<std::string> Env =
+              sys::Process::GetEnv("LLVM_WINDOWS_PREFER_FORWARD_SLASH"))
+        return StringRef(*Env) == "1";
```
zmodem: How about just `*Env == "1"`?

zmodem (approved):
lgtm

PR #199211, opened 2026-05-22
[flang-rt] Remove fallback if RPC headers not found

jhuber6:
Summary:
These are stored in the libc/shared and have a unified CMake helper to
find them. Likely these will be a more core dependency as LLVM uses them
for constexpr math, libcxx uses it, and compiler-rt will probably use
bits of it.

The original intention was to allow building flang-rt with a partial
checkout, but i don't think this is a reasonable use-case and I do not
think this exists in practice.


Meinersbur (approved):
LGTM

PR #199213, opened 2026-05-22
[InstCombine] Narrow umin/umax/smin/smax through trunc.

fhahn:
Update EvaluateInDifferentType / canEvaluateTruncated to support
narrowing umin/umax/smin/smax intrinsics, when their result fits in the
narrow type: zero high bits for umin/umax, or enough sign
bits for smin/smax.

Alive2 Proofs:
 * umin/umax with high bits zero: https://alive2.llvm.org/ce/z/dJC_Fj
 * smin/smax with sign-bits set: https://alive2.llvm.org/ce/z/z7vM8Z

End-to-end examples from C workloads performing pixel math: https://llvm.godbolt.org/z/jK3bd3GfY


dtcxzyw (approved):
LGTM

PR #199217, opened 2026-05-22
[LLVM] Precise error message for intrinsic signature verification (2/n)

jurahul:
Print precise error message for overloaded types and `LLVMMatchType` when an intrinsic's type signature verfication fails.

### DIFF
--- a/llvm/lib/IR/Intrinsics.cpp
+++ b/llvm/lib/IR/Intrinsics.cpp
@@ -1039,8 +1039,8 @@ matchIntrinsicType(Type *Ty, ArrayRef<Intrinsic::IITDescriptor> &Infos,
 
       if (Ty == OverloadTys[OIdx])
         return false;
-      OS << Position << " type (MatchType<" << OIdx << ">) expected "
-         << *OverloadTys[OIdx] << ", but got " << *Ty;
+      OS << Position << " type (matching overload type " << OIdx
+         << ") expected " << *OverloadTys[OIdx] << ", but got " << *Ty;
       return true;
     }
 
--- a/llvm/lib/IR/Intrinsics.cpp
+++ b/llvm/lib/IR/Intrinsics.cpp
@@ -1032,8 +1032,8 @@ matchIntrinsicType(Type *Ty, ArrayRef<Intrinsic::IITDescriptor> &Infos,
   case IITDescriptor::Overloaded: {
     unsigned OIdx = D.getOverloadIndex();
     if (D.getOverloadKind() == IITDescriptor::AK_MatchType) {
-      // This is a dependent type instance, check it similar to other dependent
-      // types.
+      // This is a dependent type instance, check it similarly to other
+      // dependent types.
       if (OIdx >= OverloadTys.size())
         return IsDeferredCheck || DeferCheck(Ty);
 
--- a/llvm/lib/IR/Intrinsics.cpp
+++ b/llvm/lib/IR/Intrinsics.cpp
@@ -940,11 +940,15 @@ matchIntrinsicType(Type *Ty, ArrayRef<Intrinsic::IITDescriptor> &Infos,
 
   IITDescriptor D = Infos.consume_front();
 
-  auto PrintMsg = [&OS, &Position, Ty](bool IsValid,
-                                       const Twine &Expected) -> bool {
+  auto PrintMsg = [&OS, &Position,
+                   Ty](bool IsValid, const Twine &Expected,
+                       std::optional<unsigned> OIdx = std::nullopt) -> bool {
     if (IsValid)
       return false;
-    OS << Position << " type expected " << Expected << ", but got " << *Ty;
+    OS << Position << " type";
+    if (OIdx)
+      OS << " (overload type " << *OIdx << ")";
+    OS << " expected " << Expected << ", but got " << *Ty;
     return true;
   };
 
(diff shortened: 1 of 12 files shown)


jurahul:
@nikic gentle ping

llvm/lib/IR/Intrinsics.cpp
```diff
@@ -1025,35 +1029,42 @@ matchIntrinsicType(Type *Ty, ArrayRef<Intrinsic::IITDescriptor> &Infos,
...
-    if (D.getOverloadIndex() < OverloadTys.size())
-      return Ty != OverloadTys[D.getOverloadIndex()];
-
-    if (D.getOverloadIndex() > OverloadTys.size() ||
-        D.getOverloadKind() == IITDescriptor::AK_MatchType)
-      return IsDeferredCheck || DeferCheck(Ty);
+  case IITDescriptor::Overloaded: {
+    unsigned OIdx = D.getOverloadIndex();
+    if (D.getOverloadKind() == IITDescriptor::AK_MatchType) {
+      // This is a dependent type instance, check it similar to other dependent
```
nikic:
```suggestion
      // This is a dependent type instance, check it similarly to other dependent
```

llvm/test/Assembler/implicit-intrinsic-declaration-invalid.ll
```diff
@@ -3,7 +3,7 @@
 ; Use of intrinsic without mangling suffix and invalid signature should
 ; be rejected.
 
-; CHECK: error: intrinsic has incorrect argument type!
+; CHECK: error: intrinsic argument 1 type (MatchType<0>) expected i8, but got i16
```
nikic:
```suggestion
; CHECK: error: intrinsic argument 1 type expected i8 (overload type 0), but got i16
```
Possibly?

nikic (approved):
LGTM

llvm/test/Assembler/implicit-intrinsic-declaration-invalid.ll
jurahul: For future changes, I was thinking of repeating the tablegen syntax (without the LLVM suffix) for dependent types in the error message, so would like to keep that consistent here. For example, for error messages related to LLVMExtendType, I was planning to use the `ExtendType<n>` in the error message. With that scheme, maybe better to keep using the MatchType<> syntax here? WDYT?

llvm/lib/IR/Intrinsics.cpp
jurahul: Done.

llvm/test/Assembler/implicit-intrinsic-declaration-invalid.ll
jurahul: @nikic for viz.

nikic: Not sure, it seems a bit odd to leak TableGen details in verifier error messages. I won't insist though.

jurahul: That's fair, we just need to have proper English descriptions for the other dependent types to avoid repeating the tablegen syntax. I have sketched them out below just to make sure they seem feasible, which they do:

```
Extend/Truncate
intrinsic argument 0 is extended overload type 0, so overload type 0 expected to be int or vector of int, but got <4 x float>
intrinsic argument 0 (extended overload type 0) expected <4 x i64> (overload type 0 is <4 x i32>), but got <4 x i16>

OneNTh:
intrinsic argument 0 is 1/nth elements vector of overload type 3, so overload type 3 expected to be a vector, but got float
intrinsic argument 0 (1/nth elements vector of overload type 3) expected <4 x float> (overload type 3 is <12 x float>), but got float

SameVecWidth:
intrinsic argument 0 (same vector width of overload type 3) expected a scalar type (overload type 3 is float), but got <4 x float>
intrinsic argument 0 (same vector width of overload type 3) expected a vector type (overload type 3 is <4 x float>), but got float

VecElement:
intrinsic argument 0 is vector element type of overload type 3, so overloaded type 3 expected to be a vector type, but got float
intrinsic argument 0 (vector element type of overload type 3) expected float (overload type 3 is <4 x float>), but got <4 x float>

Subdivide:
intrinsic argument 0 is overload type 3 subdivided by 2, so overload type 3 expected to be a vector, but got float
intrinsic argument 0 (overload type 3 subdivided by 2) expected to be a <4 x i32> (overload type 3 is <2 x i64>), but got float

VecOfBitcastsToInt
integer argument 0 is integer bitcast vector of overload type 3, so overload type 3 expected to be vector, but got float
integer argument 0 (integer bitcast vector of overload type 3) expected <4 x i32> (overload type 3 is <4 x float>), but got i32

VecOfAnyPtrsToElt
intrinsic argument 0 is a vector of pointers to elements of overload type 3, so overload type 3 expected to be a vector, but got
intrinsic argument 0 (vector of pointers to elements of overload type 3) expected a vector of pointers with 6 elements (overload type 3 is <6 x i32>), but got <4 x float>
```

Based on this, let me remove the reference to the tablegen syntax here. I'll adopt these sketched error message in the next PR.

jurahul: One nit for this is I'd prefer to use (matching overloaded type 0) to distinguish between MatchType<> and the llvm_any* type in the intrinsic definition.

jurahul: Done, the error message now looks like: intrinsic argument 1 type (matching overload type 0) expected i8, but got i16

PR #199218, opened 2026-05-22
[lldb][test] Fix thread safety analysis warning in LockedTest

ingomueller-net:
Guard `mutex.unlock()` with `if (mutex.try_lock())` to satisfy thread safety analysis. Statically, the compiler cannot verify that `mutex.try_lock()` succeeded when it is only asserted by `EXPECT_TRUE`, leading to a "releasing mutex 'mutex' that was not held" compilation error.

This fixes a regression introduced in #198941.


itf (approved):

JDevlieghere:
Thanks!

PR #199220, opened 2026-05-22
[libc] implement mkstemp

un-pixelated:
Fixes #191266
Implements `mkstemp` as specified in POSIX
Currently Linux-only since it relies on the Linux syscall wrappers for `getrandom` and `open`

### DIFF
--- a/libc/config/linux/x86_64/entrypoints.txt
+++ b/libc/config/linux/x86_64/entrypoints.txt
@@ -198,6 +198,7 @@ set(TARGET_LIBC_ENTRYPOINTS
     libc.src.stdlib.llabs
     libc.src.stdlib.lldiv
     libc.src.stdlib.memalignment
+    libc.src.stdlib.mkstemp
     libc.src.stdlib.qsort
     libc.src.stdlib.qsort_r
     libc.src.stdlib.rand
--- a/libc/src/stdlib/mkstemp.cpp
+++ b/libc/src/stdlib/mkstemp.cpp
@@ -1,0 +1,87 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// Implementation of mkstemp, a POSIX function that creates a unique temporary
+/// file from a template string ending in at least six 'X' characters.
+///
+/// Replaces the trailing X's with random characters from the POSIX portable
+/// filename character set, opens the file exclusively, and returns an open
+/// file descriptor, retrying automatically on name collision. See:
+/// https://pubs.opengroup.org/onlinepubs/9799919799/functions/mkdtemp.html
+///
+//===----------------------------------------------------------------------===//
+
+#include "src/stdlib/mkstemp.h"
+#include "hdr/errno_macros.h"
+#include "hdr/fcntl_macros.h"
+#include "src/__support/CPP/string_view.h"
+#include "src/__support/OSUtil/linux/syscall_wrappers/getrandom.h"
+#include "src/__support/OSUtil/linux/syscall_wrappers/open.h"
+#include "src/__support/common.h"
+#include "src/__support/libc_errno.h"
+#include "src/__support/macros/config.h"
+#include "src/__support/macros/null_check.h"
+
+namespace LIBC_NAMESPACE_DECL {
+
+LLVM_LIBC_FUNCTION(int, mkstemp, (char *tmpl)) {
+  LIBC_CRASH_ON_NULLPTR(tmpl);
+
+  cpp::string_view str_view(tmpl);
+  size_t count = 0;
+  size_t len = str_view.size();
+
+  for (size_t i = len; i > 0; i--) {
+    if (str_view[i - 1] != 'X')
+      break;
+    count++;
+  }
+
+  if (count < 6) {
+    libc_errno = EINVAL;
+    return -1;
+  }
+
+  char *suffix = tmpl + len - count;
+
+  // POSIX portable filename character set, sorted by ASCII value.
+  // See
+  // https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap03.html#tag_03_265
+  const char charset[] = "-._0123456789"
+                         "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                         "abcdefghijklmnopqrstuvwxyz";
+
+  int result = -1;
+  bool file_created = false;
+  while (!file_created) {
+    for (size_t i = 0; i < count; i++) {
+      uint8_t rand_byte;
+      auto ret = linux_syscalls::getrandom(&rand_byte, 1, 0);
+      if (!ret.has_value()) {
+        libc_errno = ret.error();
+        return -1;
+      }
+      // sizeof(charset) - 1 to account for the null terminator
+      suffix[i] = charset[rand_byte % (sizeof(charset) - 1)];
+    }
+
+    auto fd = linux_syscalls::open(tmpl, O_RDWR | O_CREAT | O_EXCL, 0600);
+    if (!fd.has_value()) {
+      if (fd.error() == EEXIST)
+        continue;
+      libc_errno = fd.error();
+      return -1;
+    }
+    result = fd.value();
+    file_created = true;
+  }
+  return result;
+}
+
+} // namespace LIBC_NAMESPACE_DECL
--- a/libc/test/src/stdlib/mkstemp_test.cpp
+++ b/libc/test/src/stdlib/mkstemp_test.cpp
@@ -1,0 +1,207 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// Tests for mkstemp
+/// See: https://pubs.opengroup.org/onlinepubs/9799919799/functions/mkdtemp.html
+///
+//===----------------------------------------------------------------------===//
+
+#include "hdr/errno_macros.h"
+#include "hdr/fcntl_macros.h"
+#include "hdr/signal_macros.h"
+#include "src/stdlib/mkstemp.h"
+#include "src/string/strdup.h"
+#include "src/string/strlen.h"
+#include "src/unistd/access.h"
+#include "src/unistd/close.h"
+#include "src/unistd/read.h"
+#include "src/unistd/unlink.h"
+#include "src/unistd/write.h"
+#include "test/UnitTest/ErrnoCheckingTest.h"
+#include "test/UnitTest/Test.h"
+
+using LlvmLibcMkstempTest = LIBC_NAMESPACE::testing::ErrnoCheckingTest;
+
+TEST_F(LlvmLibcMkstempTest, ValidTemplate) {
+  char *tmpl = LIBC_NAMESPACE::strdup(libc_make_test_file_path("tmp_XXXXXX"));
+  int fd = LIBC_NAMESPACE::mkstemp(tmpl);
+  ASSERT_GE(fd, 0);
+  LIBC_NAMESPACE::close(fd);
+  LIBC_NAMESPACE::unlink(tmpl);
+  ::free(tmpl);
+}
+
+TEST_F(LlvmLibcMkstempTest, TemplateModifiedInPlace) {
+  char *tmpl = LIBC_NAMESPACE::strdup(libc_make_test_file_path("tmp_XXXXXX"));
+  size_t len = LIBC_NAMESPACE::strlen(tmpl);
+  size_t count = 0;
+  for (size_t i = len; i > 0 && tmpl[i - 1] == 'X'; i--)
+    count++;
+  int fd = LIBC_NAMESPACE::mkstemp(tmpl);
+  ASSERT_GE(fd, 0);
+  bool modified = false;
+  for (size_t i = len - count; i < len; i++)
+    if (tmpl[i] != 'X') {
+      modified = true;
+      break;
+    }
+  EXPECT_TRUE(modified);
+  LIBC_NAMESPACE::close(fd);
+  LIBC_NAMESPACE::unlink(tmpl);
+  ::free(tmpl);
+}
+
+TEST_F(LlvmLibcMkstempTest, AllCharactersInCharset) {
+  char *tmpl = LIBC_NAMESPACE::strdup(libc_make_test_file_path("tmp_XXXXXX"));
+  size_t len = LIBC_NAMESPACE::strlen(tmpl);
+  size_t count = 0;
+  for (size_t i = len; i > 0 && tmpl[i - 1] == 'X'; i--)
+    count++;
+  int fd = LIBC_NAMESPACE::mkstemp(tmpl);
+  ASSERT_GE(fd, 0);
+  // POSIX portable filename character set, sorted by ASCII value.
+  // See
+  // https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap03.html#tag_03_265
+  const char charset[] = "-._0123456789"
+                         "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                         "abcdefghijklmnopqrstuvwxyz";
+  for (size_t i = len - count; i < len; i++) {
+    bool found = false;
+    for (size_t j = 0; j < sizeof(charset) - 1; j++) {
+      if (tmpl[i] == charset[j]) {
+        found = true;
+        break;
+      }
+    }
+    EXPECT_TRUE(found);
+  }
+  LIBC_NAMESPACE::close(fd);
+  LIBC_NAMESPACE::unlink(tmpl);
+  ::free(tmpl);
+}
+
+TEST_F(LlvmLibcMkstempTest, FileOpenedCorrectly) {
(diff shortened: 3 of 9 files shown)


un-pixelated:
@michaelrj-google
Please review this

un-pixelated:
Okay I'm working on a fix for the builds

libc/src/stdlib/mkstemp.cpp
```diff
@@ -0,0 +1,86 @@
...
+  while (*walk != '\0') {
+    walk++;
+    len++;
+  }
+
+  // invalid string length check
+  if (len < 6) {
+    libc_errno = EINVAL;
+    return -1;
+  }
```
michaelrj-google: this could be simplified by creating a `cpp::string_view` on `tmpl`.

libc/src/stdlib/mkstemp.cpp
```diff
@@ -0,0 +1,86 @@
...
+  if (counter != 6) {
+    libc_errno = EINVAL;
+    return -1;
+  }
+
+  char *suffix = tmpl + len - 6;
+
+  const char charset[] = "qwertyuiopasdfghjklzxcvbnm"
+                         "QWERTYUIOPASDFGHJKLZXCVBNM"
+                         "1234567890";
```
michaelrj-google: The posix standard (here: https://pubs.opengroup.org/onlinepubs/9799919799/functions/mkdtemp.html) says this function should use the portable filename character set, defined here: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282

libc/src/stdlib/mkstemp.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+  uint8_t rand_bytes[6];
```
michaelrj-google: the standard says there can be more than 6 `X`s. Instead of rolling all of them at once you should roll them one at a time to avoid needing to buffer them. It won't be good for performance but that's unlikely to matter here.

libc/test/src/stdlib/mkstemp_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "src/stdlib/mkstemp.h"
+#include "src/unistd/access.h"
+#include "src/unistd/close.h"
+#include "src/unistd/read.h"
+#include "src/unistd/unlink.h"
+#include "src/unistd/write.h"
+#include "test/UnitTest/ErrnoCheckingTest.h"
+#include "test/UnitTest/Test.h"
+
+TEST(LlvmLibcMkstempTest, ValidTemplate) {
```
michaelrj-google: these tests should use `ErrnoCheckingTest`, see `mmap_test.cpp` for an example.

libc/test/src/stdlib/mkstemp_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+  char tmpl[] = "tmp_XXXXXX";
```
michaelrj-google: the filepath is going to be a bit complicated when we want to support bazel, since bazel is strict about where tests are allowed to write files. It provides a path to a file, which we normally use by calling `libc_make_test_file_path`. The fact that the paths need to be writeable here makes that a bit more complicated. Your best bet may be to call `strdup` on the result of `libc_make_test_file_path` and free the pointer at the end of the test.

See https://github.com/llvm/llvm-project/blob/main/libc/test/src/unistd/unlink_test.cpp for how `libc_make_test_file_path` is used.

libc/test/src/stdlib/mkstemp_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+  EXPECT_EQ(fd, -1);
+  ASSERT_ERRNO_EQ(EINVAL);
+}
+
+TEST(LlvmLibcMkstempTest, EmptyString) {
+  char tmpl[] = "";
+  int fd = LIBC_NAMESPACE::mkstemp(tmpl);
+  EXPECT_EQ(fd, -1);
+  ASSERT_ERRNO_EQ(EINVAL);
+}
```
michaelrj-google: nit: fix missing newline at EoF

libc/config/linux/x86_64/entrypoints.txt
```diff
@@ -206,6 +198,7 @@ set(TARGET_LIBC_ENTRYPOINTS
     libc.src.stdlib.llabs
     libc.src.stdlib.lldiv
     libc.src.stdlib.memalignment
+    libc.src.stdlib.mkstemp
```
michaelrj-google, line 201: you should add this to aarch64 and risc-v as well.

michaelrj-google (commented):
This looks like a good start, I've left a few comments

libc/src/stdlib/mkstemp.cpp
```diff
@@ -0,0 +1,70 @@
+//===-- Implementation of mkstemp -----------------------------------------===//
```
michaelrj-google: We're moving to the new LLVM license headers, described here: https://llvm.org/docs/CodingStandards.html#file-headers

Please update all the C++ files to use the new format.

libc/src/stdlib/mkstemp.cpp
```diff
@@ -0,0 +1,70 @@
...
+    libc_errno = EINVAL;
+    return -1;
+  }
+
+  char *suffix = tmpl + len - count;
+
+  const char charset[] = "QWERTYUIOPASDFGHJKLZXCVBNM"
+                         "qwertyuiopasdfghjklzxcvbnm"
+                         "1234567890"
+                         "._-";
```
michaelrj-google: for clarity, please sort these alphabetically. Also add a comment on where we got this character set from.

libc/src/stdlib/mkstemp.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+  while (true) {
+    for (size_t i = 0; i < count; i++) {
+      uint8_t rand_byte;
+      auto ret = linux_syscalls::getrandom(&rand_byte, 1, 0);
+      if (!ret.has_value()) {
+        libc_errno = ret.error();
+        return -1;
+      }
+      suffix[i] = charset[rand_byte % (26 + 26 + 10 + 3)];
```
michaelrj-google: this should use `sizeof(charset)` to avoid magic constants.
```suggestion
      suffix[i] = charset[rand_byte % (sizeof(charset))];
```

libc/src/stdlib/mkstemp.cpp
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  char *suffix = tmpl + len - count;
+
+  const char charset[] = "QWERTYUIOPASDFGHJKLZXCVBNM"
+                         "qwertyuiopasdfghjklzxcvbnm"
+                         "1234567890"
+                         "._-";
+
+  while (true) {
```
michaelrj-google: The actual loop condition is that the file has not yet been created. Instead of using `while(true)` I'd recommend creating a `bool file_created` variable which starts as `false`. Then when the file is detected as successfully created it can be set to `true`. That way you can move the `return fd.value()` outside the loop, which will make control flow easier to understand.

libc/test/src/stdlib/mkstemp_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+using LlvmLibcMkstempTest = LIBC_NAMESPACE::testing::ErrnoCheckingTest;
+
+TEST_F(LlvmLibcMkstempTest, ValidTemplate) {
+  char *tmpl = LIBC_NAMESPACE::strdup(libc_make_test_file_path("tmp_XXXXXX"));
+  int fd = LIBC_NAMESPACE::mkstemp(tmpl);
+  ASSERT_GE(fd, 0);
+  LIBC_NAMESPACE::close(fd);
+  LIBC_NAMESPACE::unlink(tmpl);
+  ::free(tmpl);
```
michaelrj-google, line 37: You should add an explicit dependency on `free` and call the namespaced version.
```suggestion
  LIBC_NAMESPACE::free(tmpl);
```

libc/test/src/stdlib/mkstemp_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+}
+
+TEST_F(LlvmLibcMkstempTest, TemplateModifiedInPlace) {
+  char *tmpl = LIBC_NAMESPACE::strdup(libc_make_test_file_path("tmp_XXXXXX"));
+  size_t len = 0;
+  while (tmpl[len] != '\0')
+    len++;
```
michaelrj-google: You can use strlen here to simplify.
```suggestion
  size_t len = LIBC_NAMESPACE::strlen(tmpl);
```

libc/test/src/stdlib/mkstemp_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+TEST_F(LlvmLibcMkstempTest, FdIsWritable) {
+  char *tmpl = LIBC_NAMESPACE::strdup(libc_make_test_file_path("tmp_XXXXXX"));
+  int fd = LIBC_NAMESPACE::mkstemp(tmpl);
+  ASSERT_GE(fd, 0);
+  const char msg[] = "hello";
+  EXPECT_EQ(LIBC_NAMESPACE::write(fd, msg, 5), static_cast<ssize_t>(5));
+  LIBC_NAMESPACE::close(fd);
+  LIBC_NAMESPACE::unlink(tmpl);
+  ::free(tmpl);
+}
```
michaelrj-google: these tests can be combined, since they're effectively testing the same thing: Was the file opened correctly.

libc/test/src/stdlib/mkstemp_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+TEST_F(LlvmLibcMkstempTest, DoesNotEndInXs) {
+  char tmpl[] = "tmp_XXXXXY";
+  int fd = LIBC_NAMESPACE::mkstemp(tmpl);
+  EXPECT_EQ(fd, -1);
+  ASSERT_ERRNO_EQ(EINVAL);
+}
+
+TEST_F(LlvmLibcMkstempTest, XsInMiddleNotEnd) {
+  char tmpl[] = "XXXXXXtmp";
```
michaelrj-google: nit: The Xs here are at the beginning, not the middle.

libc/test/src/stdlib/mkstemp_test.cpp
```diff
@@ -0,0 +1,182 @@
+//===-- Unittests for mkstemp ---------------------------------------------===//
```
michaelrj-google: it would be useful if there was a way to change the randomness of `mkstemp` to force it to give the same result. That way you could check that the loop works properly.

One way to do that would be to move the mkstemp logic into an internal function that takes its random function as a template argument. That way you could set it to a "random" function that will always follow the same pattern (e.g. "AAAAAA", "AAAAAB", etc.)

michaelrj-google (commented):
Making good progress, a couple more comments

libc/test/src/stdlib/mkstemp_test.cpp
un-pixelated: Looking at how other file such as `strdup_test.cpp` have implemented the free calls, I chose to follow that pattern and use the system `::free()`.

When I use the namespaced version of free (`LIBC_NAMESPACE::free()`), with the inclusion of `"src/stdlib/free.h"` in the header file and in the CMakeLists.txt DEPENDS `libc.src.stdlib.free` I get this build error:

```
root@bd736d3d8d9f:/llvm# ninja -j5 -C build libc.test.src.stdlib.mkstemp_test.__unit__
ninja: Entering directory `build'
[1/2] Linking CXX executable libc/test/src/stdlib/libc.test.src.stdlib.mkstemp_test.__unit__.__build__
FAILED: libc/test/src/stdlib/libc.test.src.stdlib.mkstemp_test.__unit__.__build__ 
: && /usr/bin/clang++ -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -Wimplicit-fallthrough -Wcovered-switch-default -Wno-noexcept-type -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wsuggest-override -Wstring-conversion -Wno-pass-failed -Wmisleading-indentation -Wctad-maybe-unsupported -fdiagnostics-color -Xclang -fno-pch-timestamp -g -fpie -ffreestanding -fno-exceptions -fno-unwind-tables -fno-asynchronous-unwind-tables -fno-rtti -Wall -Wextra -Wconversion -Wno-sign-conversion -Wimplicit-fallthrough -Wwrite-strings -Wno-c99-extensions -Wno-gnu-imaginary-constant -Wno-pedantic -Wstrict-prototypes -Wextra-semi -Wnewline-eof -Wnonportable-system-include-path -Wthread-safety libc/test/src/stdlib/CMakeFiles/libc.test.src.stdlib.mkstemp_test.__unit__.__build__.dir/mkstemp_test.cpp.o -o libc/test/src/stdlib/libc.test.src.stdlib.mkstemp_test.__unit__.__build__  libc/src/string/CMakeFiles/libc.src.string.strlen.__internal__.dir/./strlen.cpp.o  libc/src/__support/OSUtil/linux/CMakeFiles/libc.src.__support.OSUtil.linux.linux_util.dir/./exit.cpp.o  libc/src/__support/OSUtil/linux/CMakeFiles/libc.src.__support.OSUtil.linux.linux_util.dir/./fcntl.cpp.o  libc/src/stdlib/CMakeFiles/libc.src.stdlib.mkstemp.__internal__.dir/./mkstemp.cpp.o  libc/src/errno/CMakeFiles/libc.src.errno.errno.__internal__.dir/./libc_errno.cpp.o  libc/src/string/CMakeFiles/libc.src.string.strdup.__internal__.dir/./strdup.cpp.o  libc/src/unistd/linux/CMakeFiles/libc.src.unistd.linux.access.__internal__.dir/./access.cpp.o  libc/src/unistd/linux/CMakeFiles/libc.src.unistd.linux.close.__internal__.dir/./close.cpp.o  libc/src/unistd/linux/CMakeFiles/libc.src.unistd.linux.read.__internal__.dir/./read.cpp.o  libc/src/unistd/linux/CMakeFiles/libc.src.unistd.linux.unlink.__internal__.dir/./unlink.cpp.o  libc/src/unistd/linux/CMakeFiles/libc.src.unistd.linux.write.__internal__.dir/./write.cpp.o  libc/src/__support/StringUtil/CMakeFiles/libc.src.__support.StringUtil.error_to_string.dir/./error_to_string.cpp.o  libc/test/UnitTest/libLibcDeathTestExecutors.unit.a  libc/test/UnitTest/libLibcTest.unit.a && :
/usr/bin/ld: libc/test/src/stdlib/CMakeFiles/libc.test.src.stdlib.mkstemp_test.__unit__.__build__.dir/mkstemp_test.cpp.o: in function `LlvmLibcMkstempTest_ValidTemplate::Run()':
/llvm/libc/test/src/stdlib/mkstemp_test.cpp:38:(.text+0xd4): undefined reference to `__llvm_libc_23_0_0_git::free(void*)'
/usr/bin/ld: libc/test/src/stdlib/CMakeFiles/libc.test.src.stdlib.mkstemp_test.__unit__.__build__.dir/mkstemp_test.cpp.o: in function `LlvmLibcMkstempTest_TemplateModifiedInPlace::Run()':
/llvm/libc/test/src/stdlib/mkstemp_test.cpp:58:(.text+0x30c): undefined reference to `__llvm_libc_23_0_0_git::free(void*)'
/usr/bin/ld: libc/test/src/stdlib/CMakeFiles/libc.test.src.stdlib.mkstemp_test.__unit__.__build__.dir/mkstemp_test.cpp.o: in function `LlvmLibcMkstempTest_AllCharactersInCharset::Run()':
/llvm/libc/test/src/stdlib/mkstemp_test.cpp:86:(.text+0x5d4): undefined reference to `__llvm_libc_23_0_0_git::free(void*)'
/usr/bin/ld: libc/test/src/stdlib/CMakeFiles/libc.test.src.stdlib.mkstemp_test.__unit__.__build__.dir/mkstemp_test.cpp.o: in function `LlvmLibcMkstempTest_FileOpenedCorrectly::Run()':
/llvm/libc/test/src/stdlib/mkstemp_test.cpp:100:(.text+0x7a4): undefined reference to `__llvm_libc_23_0_0_git::free(void*)'
/usr/bin/ld: libc/test/src/stdlib/CMakeFiles/libc.test.src.stdlib.mkstemp_test.__unit__.__build__.dir/mkstemp_test.cpp.o: in function `LlvmLibcMkstempTest_Uniqueness::Run()':
/llvm/libc/test/src/stdlib/mkstemp_test.cpp:121:(.text+0xa34): undefined reference to `__llvm_libc_23_0_0_git::free(void*)'
/usr/bin/ld: libc/test/src/stdlib/CMakeFiles/libc.test.src.stdlib.mkstemp_test.__unit__.__build__.dir/mkstemp_test.cpp.o:/llvm/libc/test/src/stdlib/mkstemp_test.cpp:122: more undefined references to `__llvm_libc_23_0_0_git::free(void*)' follow
/usr/bin/ld: libc/test/src/stdlib/libc.test.src.stdlib.mkstemp_test.__unit__.__build__: hidden symbol `_ZN22__llvm_libc_23_0_0_git4freeEPv' isn't defined
/usr/bin/ld: final link failed: bad value
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

un-pixelated:
@michaelrj-google hey there
I've addressed the requested changes
Some things though:
1. turns out, not treating all Xs as a valid suffix breaks the POSIX spec, hence I've not implemented it. You said it was fine in our discord conversation but I'm still asking for confirmation here.
2. using the namespaced version of free causes a linker error, but using the system version of free does not. I've followed the pattern which other test files like strdup_test.cpp use ie using the system free rather than the libc one and my build succeeded. I have included the build error in the conversation above.
3. As discussed on discord, the change requested regarding testing the loop with 100% reliability is not included in this PR and a separate issue is to be made related to it. This is also relevant to point number 1.
4. I've added an extra test which checks whether the characters in the tmpfile name adhere to the posix portable filename charset or not

With that out of the way, I request you to review this pull request.

michaelrj-google (approved):
LGTM, thanks for working on this! I will merge it

PR #199221, opened 2026-05-22
[lldb][NativePDB] Use decl context from a method's class

Nerixyz:
When we look up the parent decl context for a method, we used to look for that by name. For "regular" types that works well.
Say, we have a `struct ns::Foo` and `ns::Foo::bar`. We'd search for `ns::Foo` and use its context.

For special types - mainly lambdas - this doesn't work. They can't be found by looking for the name we get from the function symbol. That name won't be in the TPI hash table we use.

I'm not fully sure under which name they're in that table. For example, if we get the name `Foo::fun::<lambda_1>::operator()` from the function symbol, we'd look for `Foo::fun::<lambda_1>`. This fails. A class with this name is actually present, but it has a different unique name. Its unique name demangles to `` `public: void __cdecl Foo::fun(void)'::`1'::<lambda_1> `` which doesn't match what we searched for. That might be the issue.

If we can't find a class/struct, we create namespaces. So before, we created namespaces for each segment and added a regular function inside. This was wrong for methods.

However, we can tell that a function is a method by looking at its function type. And we know the parent class from this type. With this PR that parent class is used if it's available.

### DIFF
--- a/lldb/source/Plugins/SymbolFile/NativePDB/PdbAstBuilderClang.cpp
+++ b/lldb/source/Plugins/SymbolFile/NativePDB/PdbAstBuilderClang.cpp
@@ -360,8 +360,9 @@ PdbAstBuilderClang::CreateDeclInfoForUndecoratedName(llvm::StringRef name) {
   return {context, std::string(uname)};
 }
 
-std::pair<clang::DeclContext *, std::string>
-PdbAstBuilderClang::CreateDeclInfoForCompilandSymbol(PdbCompilandSymId uid) {
+clang::DeclContext *
+PdbAstBuilderClang::GetOrCreateDeclContextForCompilandSymbol(
+    PdbCompilandSymId uid) {
   SymbolFileNativePDB *pdb = static_cast<SymbolFileNativePDB *>(
       m_clang.GetSymbolFile()->GetBackingSymbolFile());
   PdbIndex &index = pdb->GetIndex();
@@ -371,7 +372,7 @@ PdbAstBuilderClang::CreateDeclInfoForCompilandSymbol(PdbCompilandSymId uid) {
 
   std::optional<PdbTypeSymId> func_id = GetFunctionType(sym);
   if (!func_id || !symbol_name.contains("::"))
-    return CreateDeclInfoForUndecoratedName(symbol_name);
+    return CreateDeclInfoForUndecoratedName(symbol_name).first;
 
   // Try to get the context from class type of an LF_MFUNCTION.
   // For some types, we might not find a class type.
@@ -421,16 +422,9 @@ PdbAstBuilderClang::CreateDeclInfoForCompilandSymbol(PdbCompilandSymId uid) {
 
   clang::DeclContext *context = get_member_fn_context();
   if (!context)
-    return CreateDeclInfoForUndecoratedName(symbol_name);
+    return CreateDeclInfoForUndecoratedName(symbol_name).first;
 
-  MSVCUndecoratedNameParser parser(symbol_name);
-  llvm::ArrayRef<MSVCUndecoratedNameSpecifier> specifiers =
-      parser.GetSpecifiers();
-  if (specifiers.size() < 2) {
-    assert(false && "Member function name with less than two scopes");
-    return CreateDeclInfoForUndecoratedName(symbol_name);
-  }
-  return {context, std::string(specifiers.back().GetFullName())};
+  return context;
 }
 
 clang::DeclContext *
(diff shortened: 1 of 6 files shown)


lldb/test/Shell/SymbolFile/NativePDB/ast-lambdas-msvc.cpp
```diff
@@ -0,0 +1,46 @@
...
+int main() {
+  Foo f;
+  f.fun();
+  return 0;
+}
+
+// CHECK:      namespace `public: void __cdecl Foo::fun(void)'::`{{.*}}' {
+// CHECK-NEXT:     class <lambda_1> {
+// CHECK:              int `Foo::fun'::`{{.*}}'::<lambda_1>::operator()() const;
+// CHECK:              int operator()() const;
```
Nerixyz: We get a "real" `operator()`  here from the class type itself. The name above it comes from the function symbol we found.

On Clang, we don't get this method (in fact we don't get any methods declared on the type). It omits these on the type itself. From what I can tell, it stops here: https://github.com/llvm/llvm-project/blob/d7d60e1d69c0bf7850f52babdb22ba031aed0f87/clang/lib/CodeGen/CGDebugInfo.cpp#L2576-L2577

lldb/source/Plugins/SymbolFile/NativePDB/PdbAstBuilderClang.cpp
```diff
@@ -360,6 +360,79 @@ PdbAstBuilderClang::CreateDeclInfoForUndecoratedName(llvm::StringRef name) {
...
+    return CreateDeclInfoForUndecoratedName(symbol_name);
+
+  MSVCUndecoratedNameParser parser(symbol_name);
+  llvm::ArrayRef<MSVCUndecoratedNameSpecifier> specifiers =
+      parser.GetSpecifiers();
+  if (specifiers.size() < 2) {
+    assert(false && "Member function name with less than two scopes");
+    return CreateDeclInfoForUndecoratedName(symbol_name);
+  }
+  return {context, std::string(specifiers.back().GetFullName())};
```
ZequanWu: I think `std::string(specifiers.back().GetFullName())` is the reason why we get the full name of the method of a lambda. Can we use the base_name instead to just get the base name of the method?

lldb/test/Shell/SymbolFile/NativePDB/ast-lambdas-msvc.cpp
ZequanWu: Do you mean the second `operator()` is from the an extra function type record that MSVC generated while clang doesn't do it so we only see one `operator()` in the non-msvc test?

Can we deduplicate that in lldb?

lldb/source/Plugins/SymbolFile/NativePDB/PdbAstBuilderClang.cpp
Nerixyz: > I think `std::string(specifiers.back().GetFullName())` is the reason why we get the full name of the method of a lambda. Can we use the base_name instead to just get the base name of the method?

We didn't use this return value at here. Only `.first` was used. I removed that second value ([`22a6e5c` (this PR)](https://github.com/llvm/llvm-project/pull/199221/commits/22a6e5c2121932247fa60cba1688c69d4f2289a3)). 

[`PdbAstBuilderClang::GetOrCreateFunctionDecl`](https://github.com/llvm/llvm-project/blob/0513c9cba2a02fb853b0af385c0b47503e6678ff/lldb/source/Plugins/SymbolFile/NativePDB/PdbAstBuilderClang.cpp#L1057-L1065) is the reason we get the full name. It gets the context name from the decl context it retrieves. But there, we can be a bit smarter. If the function name doesn't start with the context, we can instead use the basename. I did that in [`26b0b74` (this PR)](https://github.com/llvm/llvm-project/pull/199221/commits/26b0b7406a15fcff210631e40d9871f0ddfe583b).

lldb/test/Shell/SymbolFile/NativePDB/ast-lambdas-msvc.cpp
Nerixyz: > Do you mean the second `operator()` is from the an extra function type record that MSVC generated while clang doesn't do it so we only see one `operator()` in the non-msvc test?

Yes.

> Can we deduplicate that in lldb?

In this case, we can do that. See https://github.com/llvm/llvm-project/pull/199221#discussion_r3316631576.

lldb/source/Plugins/SymbolFile/NativePDB/PdbAstBuilderClang.cpp
ZequanWu: Thanks for explanation.

ZequanWu (approved):

PR #199226, opened 2026-05-22
[mlir][LLVMIR] implement PromotableAliaserInterface

jeanPerier:
Implement the new PromotableAliaserInterface for LLVM IR dialect.

Note that since LLVM IR pointer are opaque, the type conversion logic is still done by the load/store as before, but this patch allows load/store after bitcast, launder.invariant.group, strip.invariant.group, zero offset getelementptr, and addrspacecast to be candidates for mem2reg promotion.

Without this patch, mem2reg does not succeeds on any of the newly added tests.

[Edit: tests written by Claude Opus 4.6].


mlir/lib/Dialect/LLVMIR/IR/LLVMMemorySlot.cpp
```diff
@@ -536,6 +536,18 @@ DeletionKind LLVM::BitcastOp::removeBlockingUses(
...
 
+void LLVM::BitcastOp::getPromotableSlotAliases(
+    OpOperand &aliasedSlotPointerOperand, const MemorySlot &parentSlot,
+    SmallVectorImpl<MemorySlot> &newMemorySlots) {
+  // Only pointer-to-pointer bitcasts alias a memory slot. With opaque
+  // pointers, the alias slot keeps the parent's element type since the
+  // pointer type itself does not carry it.
+  if (!isa<LLVM::LLVMPointerType>(getResult().getType()) ||
+      !isa<LLVM::LLVMPointerType>(getArg().getType()))
+    return;
```
gysit, line 547: I wonder if we should add this given that such a bitcast should fold and probably not be created in the first place (due to opaque pointers)?

gysit (approved):
Thanks for the follow up!

LGTM

The bitcast on opaque pointers should probably not show up in practice? Did you have an use case for that one? Otherwise, I would consider dropping the interface from that operation.

mlir/include/mlir/Dialect/LLVMIR/LLVMIntrinsicOps.td
```diff
@@ -412,6 +414,8 @@ def LLVM_LaunderInvariantGroupOp
 def LLVM_StripInvariantGroupOp
     : LLVM_OneResultIntrOp<"strip.invariant.group", [], [0],
         [DeclareOpInterfaceMethods<PromotableOpInterface>,
+         DeclareOpInterfaceMethods<PromotableAliaserInterface,
```
Moxinilian, line 417: getPromotableSlotAliases is not optional so it does not need to be mentioned explicitly.

mlir/lib/Dialect/LLVMIR/IR/LLVMMemorySlot.cpp
```diff
@@ -694,6 +727,17 @@ DeletionKind LLVM::GEPOp::removeBlockingUses(
   return DeletionKind::Delete;
 }
 
+void LLVM::GEPOp::getPromotableSlotAliases(
+    OpOperand &aliasedSlotPointerOperand, const MemorySlot &parentSlot,
+    SmallVectorImpl<MemorySlot> &newMemorySlots) {
+  // Only zero-index GEPs are no-op aliases of the slot pointer; non-zero
+  // indices step into the slot and cannot be projected back generically.
```
Moxinilian, line 734: I don't think it's impossible, it's just complicated, right? When it comes to aggregates, it should be fairly straight forward, and when it comes to arrays, you might need some data layout understanding and bitmap it into the value (unclear if useful), but it seems like it's doable in theory?

I'm not arguing in favor of doing it now, but it seems misleading to say it's not possible.

mlir/test/Dialect/LLVMIR/mem2reg.mlir
```diff
@@ -1180,3 +1180,314 @@ llvm.func @dead_direct_use(%arg0 : i1) {
...
+  %2 = llvm.getelementptr %1[0] : (!llvm.ptr) -> !llvm.ptr, i32
+  llvm.store %arg0, %2 : i32, !llvm.ptr
+  %3 = llvm.load %2 : !llvm.ptr -> i32
+  // CHECK: llvm.return %[[ARG0]] : i32
+  llvm.return %3 : i32
+}
+
+// -----
+
+// Non-zero GEPs are not aliases of the whole slot, so promotion must fail.
```
Moxinilian, line 1283: Same opinion here.

mlir/lib/Dialect/LLVMIR/IR/LLVMMemorySlot.cpp
Moxinilian: Also, those geps are likely to get folded too I expect, so I'm not sure how commonly this path would be triggered.

